PR #1904, opened 2019-04-02
Add dropdown to filter PR by build status success, failure and pending

### LINKED ISSUE #1806: Add an option to filter PR search by passing/failing/running CI checks
When a project has many open pull requests, just finding one which is ready for review can be tricky.  For example, [pytest](https://github.com/pytest-dev/pytest/pulls) and [matplotlib](https://github.com/matplotlib/matplotlib/pulls) would both find this useful, along with smaller projects.  Terrible mockup:

image

I know I'd love to be able to filter down to just the PRs with passing tests to review, or tests+approval to merge, or ...  You get the idea.  No idea why this doesn't already exist, so Refined Github seemed the place to suggest it 😄

HardikModha:
1. URLs on which PR can be tested.
- https://github.com/facebook/react/pulls

2. This PR closes #1806.  

ezgif com-video-to-gif (1)


HardikModha:
@bfred-it Hi, Can you please take a look at the PR and see if are there any changes needed? Thanks!

I have only one concern with the added extra filter and it's that if the number of PRs is huge then it might cause some styling issues. You can see in the attached image that the filter bar is almost full. 
download (3)

bfred-it:
It's definitely a lot. My solution would be to use icons for `sort`, `labels`, `milestones`, and `projects`, the rest of them don't have an recognizable icon. This might look weird though. Maybe grouping these icons to the right would clean it up, but still... 🤷‍♂️

HardikModha:
Yep, I was thinking the same of icons. But some icons and some text would be terrible. Apart from the icons you mentioned maybe we can use icons for the following.

1. Build status: Depending on the CI tool used by the project we can show the icon. e.g. CircleCI, Travis, Jenkins etc.
2. Reviews: Something like 👀? (all-contributors uses this) or 🔍?

Assignee and Author icons can be little confusing to the user.

bfred-it:
As I said, they are not recognizable and there aren't any icons for them on GitHub. I have no idea what 👀 and 🔍 mean if I see on that bar. Also emojis can't be used as they don't fit GitHub's UI.

source/features/filter-pr-by-build-status.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+	const [currentStatus = ''] = existingFilter.match(/(?<=status:)(success|failure|pending)/) || [];
+
+	if (currentStatus) {
+		existingFilter = existingFilter.replace(/status:(success|failure|pending)/g, '').trim();
+	}
+
+	const encodedUrl = existingFilter.split(' ').map(str => encodeURIComponent(str)).join('+');
+
+	const existingUrl = `${baseUrl}?q=${encodedUrl}`;
```
bfred-it: This whole thing needs to be rewritten using `URLSearchParams`. You can look at how this is used in other features. I tried writing some examples but it requires rewriting most of these lines so I gave up

source/features/filter-pr-by-build-status.tsx
```diff
@@ -0,0 +1,70 @@
...
+	const {ownerName, repoName} = getOwnerAndRepo();
+	const baseUrl = `https://github.com/${ownerName}/${repoName}/pulls`;
+
+	const statusList = ['Success', 'Failure', 'Pending'];
+
+	let existingFilter = select<HTMLInputElement>('#js-issues-search').value
+		.trim()
+		.replace(/\s+/, ' ');
+
+	const [currentStatus = ''] = existingFilter.match(/(?<=status:)(success|failure|pending)/) || [];
```
bfred-it:
```suggestion
	const [currentStatus] = existingFilter.match(/status:(success|failure|pending)/) || [undefined];
```

source/features/filter-pr-by-build-status.tsx
```diff
@@ -0,0 +1,70 @@
...
+				</div>
+			</details-menu>
+		</details>,
+		filterAssignee
+	);
+}
+
+const getStatusDropDown = () => {
+	const {ownerName, repoName} = getOwnerAndRepo();
+	const baseUrl = `https://github.com/${ownerName}/${repoName}/pulls`;
```
bfred-it: This line isn't needed (as described in another comment), but keep in mind that RGH also works on GitHub Enterprise. It's best to use relative URLs (starting with `/` or `?` where possible)

source/features/filter-pr-by-build-status.tsx
```diff
@@ -0,0 +1,70 @@
+import React from 'dom-chef';
+import select from 'select-dom';
+import {checkInline} from '../libs/icons';
+import features from '../libs/features';
+import {getOwnerAndRepo} from '../libs/utils';
+
+function init() {
+	const filterAssignee = select('.table-list-filters > .table-list-header-toggle:last-child > details:nth-child(7)');
+
+	filterAssignee.parentNode.insertBefore(
```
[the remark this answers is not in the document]
HardikModha: Nice one. Implemented.

source/features/filter-pr-by-build-status.tsx
```diff
@@ -0,0 +1,70 @@
+import React from 'dom-chef';
+import select from 'select-dom';
+import {checkInline} from '../libs/icons';
+import features from '../libs/features';
+import {getOwnerAndRepo} from '../libs/utils';
+
+function init() {
+	const filterAssignee = select('.table-list-filters > .table-list-header-toggle:last-child > details:nth-child(7)');
```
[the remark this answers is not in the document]
HardikModha: Done.

source/features/filter-pr-by-build-status.tsx
```diff
@@ -0,0 +1,70 @@
...
+			</summary>
+			<details-menu class="select-menu-modal position-absolute right-0" style={{zIndex: 99}} role="menu">
+				<div class="select-menu-header">
+					<span class="select-menu-title">Filter by Status</span>
+				</div>
+				<div class="select-menu-list">
+					{getStatusDropDown()}
+				</div>
+			</details-menu>
+		</details>,
```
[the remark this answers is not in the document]
HardikModha: Can you please explain a bit more on this? I'll have to deep clone the node. Right? Because passing false will only shallow clone the node, 
`<details class="details-reset details-overlay select-menu d-inline-block">` which I don't think would be any helpful.  

Also, Even after deep-cloning, I'll have to change the content based on some `if`s as after cloning I'm getting 5 children nodes among which three are just text-nodes, So it will be little dirty. 

```
    const statusFilterNode = filterAssignee.cloneNode(true);
	filterAssignee.before(statusFilterNode);

	const summary = statusFilterNode.childNodes;

	for (const childNode of statusFilterNode.childNodes) {
		if (childNode.constructor === HTMLElement) {
           if (childNode.tagName === 'SUMMARY') {
              // Manipulate summary
           }

           if (childNode.tagName === 'DETAILS-MENU') {
              // Manipulate details-menu
           }
		}
	}
``` 
or I just hard-code the index as `statusFilterNode[1]` and `statusFilterNode[3]`? Which would be clean and maintainable from your POV?

source/features/filter-pr-by-build-status.tsx
HardikModha: Here, I'm getting the value of the status filter. So the change you have suggested will return `status:success` whereas my regex will return `success` only.

source/features/filter-pr-by-build-status.tsx
HardikModha: Good to know. I wasn't aware of enterprise GitHub for RGH.

source/features/filter-pr-by-build-status.tsx
HardikModha: TIL. It's much better. Thanks :)

source/features/filter-pr-by-build-status.tsx
```diff
@@ -0,0 +1,70 @@
...
+		<details class="details-reset details-overlay float-left select-menu">
+			<summary class="btn-link select-menu-button" aria-haspopup="menu">
+				Status&nbsp;
+			</summary>
+			<details-menu class="select-menu-modal position-absolute right-0" style={{zIndex: 99}} role="menu">
+				<div class="select-menu-header">
+					<span class="select-menu-title">Filter by Status</span>
+				</div>
+				<div class="select-menu-list">
+					{getStatusDropDown()}
```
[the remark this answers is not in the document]
HardikModha: Done. 👍

HardikModha:
@bfred-it I've reduced the padding around filters and it looks better. Also, I have made few suggested changes except the ones which requires some input from your side. Please have a look. :)

Screen Shot 2019-04-15 at 11 33 27 PM

yakov116:
@HardikModha Thanks for your PR!
Can you try not to include the issue number in a commit it creates unnecessary noise

image

HardikModha:
@yakov116  Sorry, It was not my intention. Will keep it in mind 🙂

source/content.css
```diff
@@ -506,6 +506,12 @@ div.inline-comment-form .form-actions,
 	padding: 6px 9px;
 }
 
+/* Decrease the padding of filters dropdown in PR filters to accommodate new filter - Filter by build status */
+.table-list-filters .table-list-header-toggle .select-menu-button {
+	padding-left: 10px;
+	padding-right: 10px;
+}
```
bfred-it: This is changed somewhere else in the CSS already, make sure to reuse that one otherwise it’s duplicate.

However I don’t think it’s still enough since your screenshot is missing “Open all” which would fill all the remaining space anyway

source/features/filter-pr-by-build-status.tsx
```diff
@@ -0,0 +1,71 @@
+import React from 'dom-chef';
+import select from 'select-dom';
+import {checkInline} from '../libs/icons';
+import features from '../libs/features';
+
+function init() {
+	const filterAssignee = select('.table-list-header-toggle > details:nth-last-child(2)');
+
+	filterAssignee.before(
```
bfred-it: You can merge line 7 and 9 (avoid the variable)

source/features/filter-pr-by-build-status.tsx
```diff
@@ -0,0 +1,71 @@
...
+}
+
+function populateDropDown() {
+	const dropDownItems = getStatusDropDownItems();
+	const menu = select('.table-list-header-toggle > details:nth-last-child(3) .select-menu-list');
+
+	for (const item of dropDownItems) {
+		menu.append(item);
+	}
+}
```
bfred-it: This function can just be `select(…).append(...dropDownItems);` so actually you can merge this function with `getStatusDropDownItem`

source/features/filter-pr-by-build-status.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+const getStatusDropDownItems = () => {
+	const statusList = ['Success', 'Failure', 'Pending'];
+
+	const searchParam = new URLSearchParams(location.search);
+	let queryString = searchParam.get('q') || 'is:pr is:open';
```
bfred-it: If the query is empty, we shouldn’t change it. 

` || ''`

source/features/filter-pr-by-build-status.tsx
```diff
@@ -0,0 +1,71 @@
...
+	if (currentStatus) {
+		queryString = queryString.replace(/status:(success|failure|pending)/g, '').trim();
+	}
+
+	return statusList.map(status => {
+		const isSelected = currentStatus ? currentStatus.toLowerCase() === status.toLowerCase() : false;
+
+		return (
+			<a
+				href={`?${String(new URLSearchParams({q: queryString + ` status:${status.toLowerCase()}`}))}`}
```
bfred-it: Maybe this works without `String`

source/features/filter-pr-by-build-status.tsx
```diff
@@ -0,0 +1,71 @@
...
+				class={`select-menu-item ${isSelected ? 'selected' : ''}`}
+				aria-checked={isSelected ? 'true' : 'false'}
```
bfred-it: `String(isSelected`

source/features/filter-pr-by-build-status.tsx
```diff
@@ -0,0 +1,71 @@
...
+	}
+}
+
+const getStatusDropDownItems = () => {
+	const statusList = ['Success', 'Failure', 'Pending'];
+
+	const searchParam = new URLSearchParams(location.search);
+	let queryString = searchParam.get('q') || 'is:pr is:open';
+
+	const [currentStatus] = queryString.match(/(?<=status:)(success|failure|pending)/) || [undefined];
```
bfred-it: Actually if we use ` || ['']`

1. On line 43 you can drop the condition `if currentStatus`  (because you can run `.replace` on `''` anyway)
2. On line 48 you can drop the ternary operator

source/features/filter-pr-by-build-status.tsx
bfred-it: Ah right, then you’d also need to change the first part to `const [, currentStatus]` because all groups are also captured separately

source/features/filter-pr-by-build-status.tsx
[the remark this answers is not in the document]
bfred-it: Indeed, I meant `cloneNode(true)` to include all children. 

Why 5 places? I see only 2 custom strings and they could be replaced with:

```is
const filter = select('......').cloneNode(true);
select('summary', filter).textContent = 'Status\xA0';
select('.select-menu-title', filter).textContent = 'Filter by Status';
```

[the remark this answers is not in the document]
HardikModha: I should have been a little more clear. I was following the above index based access approach as it was returning me the `HTMLElement` type whereas the filter here is of type `Node` so `select-dom` was complaining.
```
Argument of type 'Node' is not assignable to parameter of type 'BaseElement'.
```

For now, avoided as problem by casting it as 
```
select('...').cloneNode(true) as HTMLElement;
```
Let me know if we want to avoid this.

source/features/filter-pr-by-build-status.tsx
HardikModha: Any specific reasons on why are we avoiding lookbehind? Compatibility? Performance?

HardikModha: Also, For the suggestion you provided We'll have to provide two default values. What do you say?

```
const [, currentStatus] = queryString.match(/status:(success|failure|pending)/) || ['', ''];
```

source/content.css
HardikModha: Thanks! I looked into it and found one place. When we have the `Open all` added before the filters, the padding is being reduced to `12px`. It is only for the case when `Open all` is added. So to prevent the overflow issue I have reduced that padding to `8px`. Which doesn't look bad, IMO. 

Screen Shot 2019-04-17 at 12 09 58 AM

So when the `Open all` is present the padding will be `8px` and when it's absent the padding will be `10px`.

source/features/filter-pr-by-build-status.tsx
HardikModha: Done.

source/features/filter-pr-by-build-status.tsx
HardikModha: Done.

source/features/filter-pr-by-build-status.tsx
HardikModha: Fixed.

source/features/filter-pr-by-build-status.tsx
HardikModha: Fixed.

source/features/filter-pr-by-build-status.tsx
HardikModha: Done.

source/features/filter-pr-by-build-status.tsx
bfred-it: Regexes aren't very readable, so the shorter they are the better, especially if that means avoiding "advanced" features like lookbehind

source/features/filter-pr-by-build-status.tsx
```diff
@@ -0,0 +1,53 @@
+import React from 'dom-chef';
+import select from 'select-dom';
+import {checkInline} from '../libs/icons';
+import features from '../libs/features';
+
+function init() {
+	const statusFilter = select('.table-list-header-toggle > details:nth-last-child(3)')!.cloneNode(true) as HTMLDetailsElement;
```
sindresorhus:
```suggestion
	const statusFilter = select<HTMLDetailsElement>('.table-list-header-toggle > details:nth-last-child(3)')!.cloneNode(true);
```

source/features/filter-pr-by-build-status.tsx
```diff
@@ -0,0 +1,53 @@
...
+	select('.table-list-header-toggle > details:nth-last-child(3)')!.addEventListener('toggle', populateDropDown, {once: true});
+}
+
+function populateDropDown() {
+	const statusList = ['Success', 'Failure', 'Pending'];
+
+	const searchParam = new URLSearchParams(location.search);
+	let queryString = searchParam.get('q') || '';
+
+	const [, currentStatus] = queryString.match(/status:(success|failure|pending)/) || ['', ''];
```
sindresorhus:
```suggestion
	const [, currentStatus = ''] = queryString.match(/status:(success|failure|pending)/);
```

Is this possible? Would be nicer.

sindresorhus: You should use `\b` to match on word boundaries to prevent false positives.

source/libs/icons.tsx
```diff
@@ -52,6 +52,8 @@ export const gear = (): JSX.Element => <svg aria-hidden="true" className="octico
 
 export const foldDown = (): JSX.Element => <svg aria-hidden="true" className="octicon octicon-fold-down" width="14" height="16" ><path fill-rule="evenodd" d="M4 11l3 3 3-3H8V5H6v6H4zm-4 0c0 .55.45 1 1 1h2.5l-1-1h-1l2-2H5V8H3.5l-2-2H5V5H1c-.55 0-1 .45-1 1l2.5 2.5L0 11zm10.5-2H9V8h1.5l2-2H9V5h4c.55 0 1 .45 1 1l-2.5 2.5L14 11c0 .55-.45 1-1 1h-2.5l1-1h1l-2-2z"/></svg>;
 
+export const checkInline = (): JSX.Element => <svg className="octicon octicon-check select-menu-item-icon" viewBox="0 0 12 16" version="1.1" width="12" height="16" aria-hidden="true"><path fill-rule="evenodd" d="M12 5l-8 8-4-4 1.5-1.5L4 10l6.5-6.5L12 5z"></path></svg>;
```
sindresorhus: Use the same attribute order as other icons.

sindresorhus:
```suggestion
export const checkInline = (): JSX.Element => <svg className="octicon octicon-check select-menu-item-icon" viewBox="0 0 12 16" width="12" height="16" aria-hidden="true"><path fill-rule="evenodd" d="M12 5l-8 8-4-4 1.5-1.5L4 10l6.5-6.5L12 5z"></path></svg>;
```

source/features/filter-pr-by-build-status.tsx
HardikModha: I have added `HTMLDetailsElement` in the generic but I'm not removing the `as HTMLDetailsElement`. `select` is complaining
```
Argument of type 'Node' is not assignable to parameter of type 'Element | DocumentFragment | Document | Window | undefined'.
```
I think we need to fix types for `select-dom` and allow `Node` as a possible type.

source/features/filter-pr-by-build-status.tsx
HardikModha: It's definitely better. Thanks!

source/libs/icons.tsx
HardikModha: Done.

HardikModha:
@sindresorhus Thank you for your review. I have made the suggested changes. Please have a look.

source/features/filter-pr-by-build-status.tsx
sindresorhus: It's actually just "yet another TS quirk": https://github.com/Microsoft/TypeScript/issues/283 ...

sindresorhus: But yes, `select` should support a `Node` type too. // @bfred-it

sindresorhus:
Can you add it to the readme and link to a screenshot? Like the other entries there.

source/features/filter-pr-by-build-status.tsx
bfred-it: Node doesn’t have querySelector, it would be like supporting `createTextNode().querySelector()` which would fail

source/features/filter-pr-by-build-status.tsx
bfred-it: Oops, sorry for pushing and pulling on this line, but the ` || []` at the end is required. 

Try merging with master to ensure that whatever we try is strict TypeScript

bfred-it:
As for the readme screenshot, we just need a small static screenshot around the dropdown, in “retina” quality, so if your screen isn’t @2x, zoom in your browser at 200%

HardikModha:
@sindresorhus @bfred-it I've added the screenshot and updated readme. Let me know if it works.

readme.md
```diff
@@ -94,6 +94,23 @@ GitHub Enterprise is also supported. More info in the options.
...
+	</tr>
+	<tr><!-- Prevent zebra stripes --></tr>
+	<tr>
+		<td width="50%">
+			<img src="https://user-images.githubusercontent.com/22439276/56336299-6c961880-61bd-11e9-8dd2-d8884a6f1479.png">
+		</td>
+		<td>
+		</td>
+	</tr>
+</table>
```
sindresorhus: Sorry for being unclear. I didn't mean to add it here. These are the highlights. This is not a hightlight'able feature.

readme.md
```diff
@@ -263,3 +281,5 @@ Want more? Here are some ideas you could develop!
 ## License
 
 MIT
+
+
```
sindresorhus: Don't do unrelated changes.

readme.md
HardikModha: Updated.

readme.md
HardikModha: My bad. Edited the file directly from GitHub so couldn't notice this.

HardikModha:
@sindresorhus I think the changes should be good now. Can you please take a look?

sindresorhus:
@HardikModha Can you commit to maintain this feature going forward? For example, in case GitHub changes something that breaks it. We already have way too many features to maintain and this one is not essential for us, so we need to find a way to make this sustainable.

HardikModha:
@sindresorhus Definitely, Yes. I really appreciate your work in the open-source community and you are like my role model. I also understand that open-source is sometimes (or most of the time) difficult, given the number of projects you handle. I will assure you that I'll maintain the features I add, also If I feel if there is something that I can contribute to, I'll do it and will do my best to give back to the community. 

Cheers :)

bfred-it:

```sh
```

HardikModha:
@bfred-it Done.

bfred-it:
Thanks for this feature! I think it's pretty useful

yakov116:
```
❌ filter-pr-by-build-status
content.js:1880 TypeError: Cannot read property 'cloneNode' of null
    at filter_pr_by_build_status_init (content.js:5481)
    at features_run (content.js:1875)
    at content.js:1927
    at HTMLDocument.<anonymous> (content.js:1832)
```
Will fail if there are no pull requests in the repo

HardikModha:
@bfred-it https://github.com/sindresorhus/refined-github/pull/1904/commits/11b3e96a8f6b79953d58e117b21d9e56bce4605c introduced a bug. 

Screen Shot 2019-04-25 at 8 26 48 PM

The first time filter will work correctly, but subsequent filters won't work.

ezgif com-video-to-gif (2)

bfred-it:
🤦‍♂️ sorry

Issue #1109, opened 2018-02-22
Consider moving Recently Pushed Branches elsewhere


bfred-it:
Following #1106 and #906 

How does this position look? 

<img width="686" alt="screen shot 2018-02-22 at 14 40 04" src="https://user-images.githubusercontent.com/1402241/36525858-572ff01c-17de-11e8-8cca-9ea6379d0c1c.png">

To achieve this, the element would have to be moved between the two elements and floated right:

```js
select('.pagehead-actions').after(select('.js-socket-channel'));
```

But probably it has to be move _inside_ and positioned absolutely in case the repo name or branch name are too long to fit in one line.

Alternatively there's this position, that expands on hover:

![](https://user-images.githubusercontent.com/1402241/36526234-fe0922e0-17df-11e8-9229-2bb12481e308.png)

silverwind:
The space between repo name and the buttons can be limited on small screens.

Maybe try with `position: fixed`, which would allow the user to scroll potentially hidden content into view.

bfred-it:
> Maybe try with position: fixed

That might make it more annoying (since it floats + it could cover _anything_ on the screen, on whatever border we fix it to)

hkdobrev:
I personally don't mind it if it overlaps with something and especially if that's the repo name (even though it may hide the commit status link we add there).

I like the first suggestion.

sindresorhus:
> Alternatively there's this position, that expands on hover:

I like this one.

tonglil:
Personally I liked it when it stretched across the top of the page.

Honestly, the green "PR" button needs to be bigger to click on, and would love to see this bar/notification on any page in the repo.

Very rough image:
<img width="1028" alt="screen shot 2018-03-08 at 1 22 32 pm" src="https://user-images.githubusercontent.com/3250776/37174507-f1976de2-22d3-11e8-836a-f3a5ef79c15c.png">

SuperSandro2000:
Did you think about moving out just under fork and star?

bfred-it:
> Did you think about moving out just under fork and star?

I don't think there's enough space, it will feel cramped if there are tabs under it.

jerone:
I hope we can fix this also for forks:

![image](https://user-images.githubusercontent.com/55841/48979226-ca638000-f0b7-11e8-961d-6e16b1612c5e.png)


**Steps to reproduce**
1. Fork project.
2. Create branch and commit.
3. Go to your forked project homepage.


**Expected**
[...]

PR #1909, opened 2019-04-03
Fix recently-pushed-branches once and for all

bfred-it:
Fixes https://github.com/sindresorhus/refined-github/issues/1109

- Repositions notification to black header bar
- Adds notification on all pages of repo (but it only ever "pops in" once)
- Makes the whole feature disableable (so if people want the regular full-width bar, they can)


source/features/recently-pushed-branches-enhancements.tsx
```diff
@@ -13,30 +13,30 @@ import {getRepoURL} from '../libs/utils';
...
-	const response = await fetch(codeTabURL);
-	const html = await response.text();
-
-	// https://github.com/sindresorhus/refined-github/issues/216
-	if (html.includes(fragmentURL)) {
-		select('.repository-content').prepend(<include-fragment src={fragmentURL}></include-fragment>);
-	}
+
+	return (
+		select(`[data-url='${fragmentURL}'], [src='${fragmentURL}']`) || // Already on page
```
notlmn: Explicit `if-else` using `select.exists()` would be a better choice here.

Took me a minute to understand what was happening, with all the combined logical OR and inline comments.

notlmn:
How about we also change the background colors to make it a bit less distracting?

source/features/recently-pushed-branches-enhancements.tsx
bfred-it: Indeed. Should be better now

bfred-it:
color

source/content.css
```diff
@@ -745,63 +745,70 @@ a.tabnav-extra[href$='mastering-markdown/'] {
...
 }
-.RecentBranches {
+.rgh-recently-pushed-branches .RecentBranches {
 	display: flex;
 	width: max-content;
-	margin-top: 10px; /* Add extra padding to make it look right on forked repos */
 	margin-left: auto;
 	flex-direction: column-reverse;
+	background: none;
+	border-color: #fff;
```
notlmn:
```suggestion
	border-color: currentColor;
```

notlmn:
Is the time any helpful here, just thinking 🤔?

bfred-it:
It helps you understand if it's something that was pushed just now or an old push. Not _really_ useful but it's not a huge distraction.

bfred-it:
@sindresorhus if the style looks ok I can finalize this

sindresorhus:
I found the button very useful. I always use it to create a new PR.

The style in https://github.com/sindresorhus/refined-github/pull/1909#issuecomment-479584096 looks very good. I would make the gradient gray button slightly darker.

sindresorhus:
Can you rebase from master?

bfred-it:
I don't know how dark you're envisioning it. Mind playing around with the filters in 3f0c55e?

Current look with multiple branches hidden

Screenshot 2019-04-18 at 19 18 09

sindresorhus:
Done

notlmn:
@bfred-it Still open to some changes?

### Visual changes

| Collapsed | Expanded |
|:-:|:-:|
| image | image |

### Source diff

``` diff
diff --git a/source/content.css b/source/content.css
index 5378d509e3b3..1d65ff42f8b8 100644
--- a/source/content.css
        text-indent: -999999px; /* Hide text */
-}
-.rgh-recently-pushed-branches .RecentBranches:not(:hover) .btn {
-       filter: grayscale(1) brightness(0.8) contrast(2);
+       background: #34d058;
 }
 .rgh-recently-pushed-branches .RecentBranches .btn + div {
        margin-right: 8px;
-       float: right;
        font-size: 11px;
 }
 .rgh-recently-pushed-branches .RecentBranches .btn .octicon {
@@ -797,9 +794,14 @@ a.tabnav-extra[href$='mastering-markdown/'] {
        border-bottom: 1px solid;
 }
 .rgh-recently-pushed-branches .RecentBranches-item {
+       display: flex;
+       flex-direction: row-reverse;
        color: inherit;
        padding: 4px !important;
 }
+.rgh-recently-pushed-branches .RecentBranches-item:not(:hover) .btn {
+       filter: grayscale(1) brightness(0.8) contrast(2);
+}
 .rgh-recently-pushed-branches .RecentBranches:not(:hover) .RecentBranches-item:not(:last-child) {
        /* Hide older branches */
        display: none;
```

### Changes

- Reduce `border-radius` in favor for non-retina display
- Change gradient background of button to solid background
- Use `flex`, drop `float`
- Move hover from items to individual item

bfred-it:
I'm ok with with making the gray flat to make it fit better up there, but not the green, as we still need to match GitHub's style.

bfred-it:
I applied the flat color on `:not(:hover)`, but:

- `float: right` is just one easy property
- The only reason we grayscale the button is to make it fit in the header, but once the user is "onto it", the whole widget should be enabled

busches:
This is broke on GHE of course :) I'll take a look at it tomorrow.

Issue #1914, opened 2019-04-09
Add Instructions for Edge Insiders Build to README


jorgegonzalez:
https://microsoftedge.microsoft.com/insider-addons/category/EdgeExtensions

The new Edge Insider release is built on Chromium; Refined GitHub can currently be installed to Edge from the Chrome Store (and it is working flawlessly). It would be nice if we could publish an official build in the Microsoft Store. No additional testing or effort will be required of the maintainers of Refined GitHub (besides publishing), and you would "officially" support all future releases of Microsoft Edge.

jorgegonzalez:
#1213

PR #1917, opened 2019-04-10
Add instructions for Edge

jorgegonzalez:
Closes https://github.com/sindresorhus/refined-github/issues/1914


bfred-it:
We can’t guarantee that it works on Edge, so unless actual instructions are needed, I wouldn’t say anything. I guess Edge users will know to just use the Chrome version.

PR #1783, opened 2019-02-15
Enable strict-mode for TypeScript

nickytonline:
Basically going with the non-null assertion operator in almost all places, because as mentioned in https://github.com/sindresorhus/refined-github/issues/1735#issuecomment-462027092, the assumption with the extension is that the DOM nodes referenced in most cases are expected to be there. If they aren't, the extension breaks, a bug is filed and it's fixed.

One thing to note @bfred-it, is currently dom-chef types need to handle JSX (Element and IntrinsicElements) or we import `@types/react`.

Closes https://github.com/sindresorhus/refined-github/issues/1735

PR #1924, opened 2019-04-13
Add `raw-file-link` feature

### LINKED ISSUE #1595: Add "View Raw" button next to all View buttons for files
Especially in pul request "Files changed" view, countless times I copied the link to "View" button and passed to curl, only to get a HTML page, because it wasn't the link to the raw file.

> image

Would it be cool to have a "View Raw" button via this extension?

idasbiste:
This PR closes #1595 
It adds a "View raw" button next to the usual "View file" one.

Without rich diff:
Screenshot 2019-04-13 at 01 22 20

With rich diff:
Screenshot 2019-04-13 at 01 22 14


source/content.ts
```diff
@@ -93,6 +93,7 @@ import './features/collapsible-content-button';
 import './features/hide-watch-and-fork-count';
 import './features/resolve-conflicts';
 import './features/follow-file-renames';
+import './features/view-raw-file';
```
bfred-it: The feature should probably be named `raw-file-link`

source/features/view-raw-file.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'dom-chef';
+import select from 'select-dom';
+import features from '../libs/features';
+
+const createRawUrl = (href: string) => href.replace(/blob\//, 'raw/');
```
bfred-it: This fails on user/repo containing blob. Use something like:

```js 
url = link.pathname.split('/')
url[3] = 'raw'; // Replaces 'blob'
return url.join('/');

File: source/features/view-raw-file.tsx [hunk 3]
```diff
@@ -0,0 +1,31 @@
...
+
+function init() {
+	const fileActionsGroups: HTMLElement[] = select.all(
+		'.file-header .file-actions > div.BtnGroup'
+	);
+
+	for (const group of fileActionsGroups) {
```
bfred-it: You can inline the select in here, on one line. 

Instead of selecting the group, select the link you’ll need on the next line with `[href*='/blob/']` and then use `link.after()` to append our link

source/features/view-raw-file.tsx
```diff
@@ -0,0 +1,31 @@
...
+		const rawUrl = createRawUrl(group.lastElementChild.getAttribute('href'));
+		group.append(
+			<a href={rawUrl} class="btn btn-sm BtnGroup-item" rel="nofollow">
+				View raw
```
bfred-it: GitHub just calls this button Raw, not View raw

source/features/view-raw-file.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+function init() {
+	const fileActionsGroups: HTMLElement[] = select.all(
+		'.file-header .file-actions > div.BtnGroup'
+	);
+
+	for (const group of fileActionsGroups) {
+		const rawUrl = createRawUrl(group.lastElementChild.getAttribute('href'));
+		group.append(
+			<a href={rawUrl} class="btn btn-sm BtnGroup-item" rel="nofollow">
```
bfred-it: We don’t need nofollow

source/features/view-raw-file.tsx
```diff
@@ -0,0 +1,31 @@
...
+			</a>
+		);
+	}
+}
+
+features.add({
+	id: 'view-raw-file',
+	include: [
+		features.isCommit,
+		features.isCommitList,
```
bfred-it: Commit list doesn’t have files

source/features/view-raw-file.tsx
```diff
@@ -0,0 +1,31 @@
...
+		features.isPRFiles
```
bfred-it: This page may load new files asynchronously. Look at some other features on this page to see how to watch the dom and update it

source/features/view-raw-file.tsx
idasbiste: 🤦‍♂️ Of course... I was worried about the path or the filename and forgot about the user/repo!

source/features/view-raw-file.tsx
idasbiste: You're correct 👍 
I've called it "View raw" because of the discussion on the issue. Happy to change it though for the sake of consistency.

source/features/view-raw-file.tsx
idasbiste: Good point, thanks! I'll have a look.

idasbiste:
@bfred-it I think I've addressed all your comments! Please have a look.
To deal with the asynchronous files load and the respective DOM updates, I've used the MutationObserver. I had to recursively look for elements with a specific class and observe it because the files are added in "cascade" mode inside the second `<div class="js-diff-progressive-container">`. If the PR is large enough you can get 3 or 4 levels of this div. Something like:
```
<div class="js-diff-progressive-container">
  ...
  <div class="js-diff-progressive-container">
    ...
    <div class="js-diff-progressive-container">
     ...
    </div>
  </div>
</div>
```

sindresorhus:
@idasbiste Can you rebase from master (we enabled TS strict mode) and fix any issues?

sindresorhus:
Seems to work fine for me otherwise.

source/features/raw-view-link.tsx
```diff
@@ -0,0 +1,54 @@
...
+			}
+		}
+	}
+}
+
+function init() {
+	addRawBtn(select('.js-diff-progressive-container:first-child'));
+	const lastProgressiveContainer = select('.js-diff-progressive-container:last-child');
+	if (lastProgressiveContainer) {
+		observer.observe(lastProgressiveContainer, {childList: true});
```
bfred-it: Maybe we should turn this into a utility function, like `on-new-pr-files.ts`

source/features/raw-view-link.tsx
```diff
@@ -0,0 +1,54 @@
...
+			}
+		}
+	}
+});
+
+function addRawBtn(node: Element | null) {
+	if (node) {
+		for (const link of select.all('.file-header .file-actions > div.BtnGroup [href*=\'/blob/\']', node)) {
+			const href = link.getAttribute('href');
+			if (href) {
```
bfred-it: We already select only links with `href`, so it always there

source/features/raw-view-link.tsx
```diff
@@ -0,0 +1,54 @@
...
+				link.after(
+					<a href={createRawUrl(href)} className="btn btn-sm BtnGroup-item">
```
bfred-it: You can use `link.pathname` instead of `(href)` here, skipping the variable altogether

source/features/raw-view-link.tsx
```diff
@@ -0,0 +1,54 @@
...
+				observer.observe(node, {childList: true});
+			} else {
+				addRawBtn(node);
+			}
+		}
+	}
+});
+
+function addRawBtn(node: Element | null) {
+	if (node) {
```
bfred-it: As in the other PR, node is always Element now

source/features/raw-view-link.tsx
```diff
@@ -0,0 +1,49 @@
...
+				addRawBtn(element);
+			}
+		}
+	}
+});
+
+function addRawBtn(node: Element) {
+	for (const link of select.all('.file-header .file-actions > div.BtnGroup [href*=\'/blob/\']', node)) {
+		link.after(
+			<a href={createRawUrl(link.getAttribute('href'))} class="btn btn-sm BtnGroup-item">
```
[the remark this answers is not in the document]
idasbiste: I wasn't able to get the `pathname` property from a raw Element. I need to cast it to an `HTMLAnchorElement`. Also, I didn't face any problem using the href because the path is always relative.
Does this look ok to you `createRawUrl((link as HTMLAnchorElement).pathname)`?

source/features/raw-view-link.tsx
```diff
@@ -0,0 +1,49 @@
...
+				Raw
+			</a>
+		);
+	}
+}
+
+function init() {
+	addRawBtn(select('.js-diff-progressive-container:first-child'));
+	observer.observe(select('.js-diff-progressive-container:last-child'), childList);
```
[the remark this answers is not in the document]
idasbiste: It is the last child. Please see my comment above https://github.com/sindresorhus/refined-github/pull/1924#issuecomment-482838249

source/features/raw-view-link.tsx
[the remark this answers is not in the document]
bfred-it: You can use `select.all<HTMLAnchorElement>` and `link` will automatically be one

source/features/raw-view-link.tsx
idasbiste: It is but after enabling TS strict mode, `getAttribute` returns string or null. This is solved by getting the pathname as mentioned above.

source/features/raw-view-link.tsx
idasbiste: Unfortunately, the `select` function returns an `HTMLElement` or null. By passing it as argument of this function when calling it in the `init` method, TS strict mode "forces" me to make this check.

source/features/raw-view-link.tsx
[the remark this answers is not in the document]
bfred-it: > It is the last child

In the HTML you posted I see two `:last-child`, and I think `querySelector` will match the first one (level 2), not the one you expect (level 3)

See https://github.com/sindresorhus/refined-github/pull/1932/commits/a6a6e15d1d0300789b4f548dffb7d5777ad10bc8 for an alternative implementation that avoids this altogether

Also do you know a PR that has more than one level? Because in the biggest PR I can think of (https://github.com/sindresorhus/refined-github/pull/1783) I'm seeing only one deferred level, _not_ nested:


```html
<div class="js-diff-progressive-container">already here</div>
<div class="js-diff-progressive-container">(content deferred)</div>
```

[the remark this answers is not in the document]
idasbiste: I like your approach on the other PR 👍 
In the HTML I've posted I was specifically talking about the container that gets updated -- which is the `:last-child` in the beginning (the one with deferred content). I.e. the HTML will, in the end, have a format of:
```
<div class="js-diff-progressive-container">already here</div>
<div class="js-diff-progressive-container">
  (nested containers)
</div>
```

source/features/raw-view-link.tsx
bfred-it: You can use `select(selector)!`, the `!` will become your best friend :D 

But https://github.com/sindresorhus/refined-github/commit/a6a6e15d1d0300789b4f548dffb7d5777ad10bc8 avoids this problem too.

source/features/raw-view-link.tsx
bfred-it: Like in the other comment, the null TS issue can be shut up with `!`, so

```ts
const href = link.getAttribute('href')!;
```

And `href` will always be `string` instead of `string | null`

source/features/raw-view-link.tsx
idasbiste: I didn't know about the `!`. Thanks!

bfred-it:

```sh
```

bfred-it:
I want to merge this but I'm seeing a couple of unrelated RGH buts on this, I want to use this PR as testing ground. I'll merge soon

notlmn:
> I want to merge this but I'm seeing a couple of unrelated RGH bugs on this (`show-names` and `co-authors`), I want to use this PR as testing ground. I'll merge soon

`show-names` is broken as the author of this PR has used an email to make commits that do not match with a GitHub account, in that case GH would display only the name used to create that commit. Querying the v4 API with this name (and especially which contain spaces) would fail, hence the broken feature.

#### Non-referencing commit
image

#### GraphQL exception
image

----------------------

Didn't want to go off topic but an it is an interesting bug that sometimes bites the GH team too, `add-co-authored-by` is probably broken for the same reason as above (didn't test that though).

bfred-it:
Thanks Eduardo! 🎯

PR #1925, opened 2019-04-13
chore: add badge license

Hongarc:
Use badge license instead of text `MIT`


bfred-it:
1. This would trigger an extra HTTP request and it wouldn’t add any info. 
2. All other Sindre’s repos follow the current style
3. Sindre is thinking about dropping the license section from the readme altogether https://github.com/sindresorhus/meta/issues/5

PR #1928, opened 2019-04-13
Apply `set-default-repositories-type-to-sources` to organizations too

### LINKED ISSUE #1761: Hide archived repos by default
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/refined-github/issues/1761)

I've just started using refined-github today and first off, thanks for making something very helpful!

I belong to a few github organizations and a couple of those have a huge list of repositories, but most of them are archived. I almost never want to see these and they really clutter up my view so I think it'd be nice to hide them by default. I am specifically talking about organization pages like https://github.com/rubocop-hq but it could also be a reasonable default for individual repository pages: https://github.com/sindresorhus?tab=repositories.

If this is something you don't think most users would prefer, is refined-github set up in such a way that it would be easy to have a disabled-by-default but easy-to-enable feature? I was surprised to see very few configuration options in the extension so I'm guessing this is probably not the case and I would need to make my own extension if it bothered me enough.

If this is something refined-github would support, I'd be willing to try implementing it.

> There is a $40.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/sindresorhus/refined-github/issues/1761).

janat08:
press cyclejs author link in breadcrumbs nav: https://github.com/cyclejs/cyclejs
select repositories: https://github.com/orgs/cyclejs/people
select org from orgs list: https://github.com/Frikki
I'm not sure that there're any more buttons related to this
Fixes #1761


bfred-it:
This should probably be merged into the profile version of it. I think only selectors are different

sindresorhus:
The PR needs a better title.

PR #1929, opened 2019-04-14
Add `remove-checks-tab` feature

### LINKED ISSUE #1304: Hide "Checks" tab on Pull request page
`Checks` tab is not related directly to the pull request workflow, would be helpful to have a feature that will disable this tab.
image

janat08:
checks not gone- non zero: https://github.com/probot/semantic-pull-requests/pull/40
 not gone- selected: https://github.com/sindresorhus/refined-github/pull/1928/checks
 this page demonstrates it being gone unless youre owner

closes #1304

### DIFF
--- a/source/features/remove-checks-tab.tsx
+++ b/source/features/remove-checks-tab.tsx
@@ -7,6 +7,14 @@ import features from '../libs/features';
 import {safeElementReady} from '../libs/dom-utils';
 
 async function init() {
+	// If there's a settings tab, the current user can enable checks,
+	// so the tab should not be hidden
+	if (select.exists([
+		'.js-repo-nav [data-selected-links^="repo_settings"]', // In repos
+		'.pagehead-tabs-item[href$="/settings/profile"]' // In organizations
+	].join())) {
+		return;
+	}
 	// Only remove the tab if it's not the current page and if it has 0 checks
 	const checksCounter = await safeElementReady('[data-hotkey="g 3"]:not(.selected) .Counter');
 
@@ -19,9 +27,6 @@ features.add({
 	id: 'remove-checks-tab',
 	include: [
 		features.isPR
-	],
-	exclude: [
-		features.isOwnRepo
 	],
 	load: features.onAjaxedPages,
 	init
--- a/source/features/remove-checks-tab.tsx
+++ b/source/features/remove-checks-tab.tsx
@@ -7,14 +7,19 @@ import features from '../libs/features';
 import {safeElementReady} from '../libs/dom-utils';
 
 async function init() {
-
 	const checksTab = await safeElementReady(`
 		[data-hotkey="g 3"]
 	`); // Wait for the tab to be loaded
 
-	if (!checksTab) return
-	const counter = select('.Counter', checksTab)
-	if (!counter || !counter.textContent) return
+	if (!checksTab) {
+		return;
+	}
+
+	const counter = select('.Counter', checksTab);
+	if (!counter || !counter.textContent) {
+		return;
+	}
+
 	// Only remove the tab if it's not the current page and if it has 0 checks
 	if (!checksTab.matches('.selected') && counter.textContent.trim() === '0') {
 		checksTab.remove();
--- a/source/features/remove-checks-tab.tsx
+++ b/source/features/remove-checks-tab.tsx
@@ -5,23 +5,18 @@ The `checks` tab in PR is removed unless you're owner
 import select from 'select-dom';
 import features from '../libs/features';
 import {safeElementReady} from '../libs/dom-utils';
-import {isOwnRepo} from '../libs/page-detect';
 
 async function init() {
-	if (isOwnRepo()) {
-		return;
-	}
 
-	await safeElementReady(`
+	const checksTab = await safeElementReady(`
 		[data-hotkey="g 3"]
 	`); // Wait for the tab to be loaded
 
-	const checksTab = select([
-		'[data-hotkey="g 3"]'
-	].join());
-
-	// Only remove the tab if it's not the current page and if it has 0 projects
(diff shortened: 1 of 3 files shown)


sindresorhus:
Can you fix the merge conflict and rebase from master (we enabled [TS strict mode](https://github.com/sindresorhus/refined-github/commit/7b3b275ad9a1b1409c56310c0b8ab286bef1f75f)) and fix any issues?

source/features/remove-checks-tab.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+	await safeElementReady(`
+		[data-hotkey="g 3"]
+	`); // Wait for the tab to be loaded
+
+	const checksTab = select([
+		'[data-hotkey="g 3"]'
+	].join());
+
+	// Only remove the tab if it's not the current page and if it has 0 projects
```
bfred-it: Here it still mentions Projects

source/features/remove-checks-tab.tsx
```diff
@@ -0,0 +1,36 @@
...
+		return;
+	}
+
+	await safeElementReady(`
+		[data-hotkey="g 3"]
+	`); // Wait for the tab to be loaded
+
+	const checksTab = select([
+		'[data-hotkey="g 3"]'
+	].join());
```
bfred-it: `join` don’t make sense here, don’t copy-paste code without reading it 😅 

You can add `const checkTabs =` before `await` and skip the `select` altogether

source/libs/page-detect.ts
```diff
@@ -77,6 +77,8 @@ export const isRepo = (): boolean => /^[^/]+\/[^/]+/.test(getCleanPathname()) &&
 	!isGist() &&
 	!isRepoSearch();
 
+export const isOwnRepo = (): boolean => isRepo() && getUsername() === getOwnerAndRepo().ownerName;
```
bfred-it: I think this can be tested in JSdom

source/features/remove-checks-tab.tsx
```diff
@@ -0,0 +1,36 @@
...
+	if (!checksTab.matches('.selected') && select('.Counter', checksTab).textContent.trim() === '0') {
+		checksTab.remove();
+	}
+}
+
+features.add({
+	id: 'remove-checks-tab',
+	include: [
+		features.isPR
+	],
```
bfred-it: `isOwnRepo` should be listed here inside `exclude`

janat08: ty

source/libs/page-detect.ts
janat08: getUsername depends on DOM

janat08: It's basically a combination of two base functions. I don't see any jsdom tests to learn from.

bfred-it: Indeed. The `is404` test in `tests/page-detect` technically also depends on the `<title>` tag, but it's more easily-accessible via `document.title`. Perhaps something like:

```js
test('isOwnRepo', t => {
	document.head.insertAdjacentHTML('<meta name="user-login" content="sindresorhus">');
	location.href = 'https://github.com/sindresorhus/refined-github/pull/1929'
	t.true(pageDetect.isOwnRepo());
});
```

The problem is that `isOwnRepo` is a one-time function, so unless @sindresorhus suggests a way to test it further, we can only do this once

bfred-it: Actually this function is incorrect and should be dropped. This is what was in `remove-projects-tab`

```js
	// If there's a settings tab, the current user can disable the projects,
	// so the tab should not be hidden
	if (select.exists([
		'.js-repo-nav [data-selected-links^="repo_settings"]', // In repos
		'.pagehead-tabs-item[href$="/settings/profile"]' // In organizations
	].join())) {
		return;
	}
```

Which is different from `currentUser === repoOwner`. This code should also work in organizations where the owner doesn't match the `repoOwner` part of the URL, like https://github.com/xojs/xo

source/libs/page-detect.ts
```diff
@@ -77,7 +77,9 @@ export const isRepo = (): boolean => /^[^/]+\/[^/]+/.test(getCleanPathname()) &&
 	!isGist() &&
 	!isRepoSearch();
 
-export const isRepoDiscussionList = (): boolean => /^(issues$|pulls$|labels\/)/.test(getRepoPath()!);
+export const isOwnRepo = (): boolean => isRepo() && getUsername() === getOwnerAndRepo().ownerName;
+
+export const isRepoDiscussionList = (): boolean => /^(issues$|pulls$|labels\/)/.test(getRepoPath());
```
bfred-it: Unrelated change?

bfred-it:

```sh
```

source/libs/page-detect.ts
janat08: basically, copy and pasted the selectors. I assume it applies exactly in the same way.

source/features/remove-checks-tab.tsx
```diff
@@ -0,0 +1,33 @@
...
+	// so the tab should not be hidden
+	if (select.exists([
+		'.js-repo-nav [data-selected-links^="repo_settings"]', // In repos
+		'.pagehead-tabs-item[href$="/settings/profile"]' // In organizations
+	].join())) {
+		return false;
+	}
+
+	// Only remove the tab if it's not the current page and if it has 0 checks
+	const checksCounter = select('.tabnav-tab[href$="/checks"]:not(.selected) .Counter');
```
bfred-it: I dropped `safeElementReady` here because `onAjaxedPages` always waits for `domReady` first.

We currently don't have a callback that is called before domready _and_ supports ajaxed pages. It'd be great to have that

bfred-it:
Thanks! ✅

PR #1930, opened 2019-04-14
web-ext integration

janat08:
Adds script for running web-ext and adding it to packages.

It might be worthwhile changing contribution.md that links to web-ext related api that is likely lower level, that I couldn't even be bothered to read, instead suggesting to use the script.


janat08:
I'm not sure if web-ext takes the time to make sure that the extension debugging is turned on, but it might be worthwhile empathizing that if it does.

janat08:
all of my other pull requests edit .gitignore. Perhaps you're looking to switch to pnpm?

bfred-it:
The gitignore entry was removed because:
- we already have package-lock, which may differ from pnpm’s own lock
- if you use pnpm you might get a different result from the default one
- use `pnpm install --no-lock` if you insist, so gitignore doesn’t need to be cluttered with all the possible installers from the past and future

package.json
```diff
@@ -14,7 +15,8 @@
 		"release": "run-s build update-version save-url release:*",
 		"save-url": "echo https://github.com/$TRAVIS_REPO_SLUG/tree/$TRAVIS_COMMIT > distribution/SOURCE_URL",
 		"update-version": "VERSION=$(utc-version); echo $VERSION; dot-json distribution/manifest.json version $VERSION",
-		"can-release": "if [ \"$TRAVIS_EVENT_TYPE\" = cron ] && [ $(git rev-list -n 1 --since=\"26 hours ago\" master) ]; then echo :ship-it:; else false; fi"
+		"can-release": "if [ \"$TRAVIS_EVENT_TYPE\" = cron ] && [ $(git rev-list -n 1 --since=\"26 hours ago\" master) ]; then echo :ship-it:; else false; fi",
+		"start:firefox": "web-ext run --source-dir ./distribution --firefox-profile=default"
```
sindresorhus:
```suggestion
		"start:firefox": "web-ext run --source-dir=distribution --firefox-profile=default"
```

contributing.md
```diff
@@ -113,4 +120,5 @@ Once built, load it in the browser of your choice:
 	</tr>
 </table>
 
-Firefox offers a more advanced loading technique that includes auto-reloading in the browser as well. Check-out [`web-ext run`](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/web-ext_command_reference#web-ext_run).
+Firefox offers a more advanced loading technique that includes auto-reloading in the browser as well. Check-out [`web-ext run`](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/web-ext_command_reference#web-ext_run). The `dev` script in package.json already take responsobility for
+using this package.
```
sindresorhus: Don't hard-wrap lines.

contributing.md
```diff
@@ -83,7 +83,14 @@ When working on the extension or checking out branches, use this to have it cons
 npm run watch # Listen for file changes and automatically rebuild
 ```
 
-Then load or reload it into the browser to see the changes (this does not happen automatically).
+Then load or reload it into the browser to see the changes (this does not happen automatically). If you'd like
+it to occur automatically than install firefox and:
```
sindresorhus:
```suggestion
it to occur automatically, install Firefox and run:
```

contributing.md
```diff
@@ -83,7 +83,14 @@ When working on the extension or checking out branches, use this to have it cons
...
+
+```sh
+npm run dev
+```
+
+This will load default profile (browser settings), but it will not keep the changes if made in session launched with the script.
```
sindresorhus:
```suggestion
This will load the default profile (browser settings), but it will not keep the changes if made in session launched with the script.
```

sindresorhus: This sentence needs to be better written.

sindresorhus:
This PR needs a more descriptive title.

bfred-it:
Does this actually work for you? It doesn't seem to reload the extension when it changes (it reloads the same version over and over) and the extension doesn't appear in `about:debugging` so I can't reload it manually either:

Screenshot 2019-04-18 at 15 53 21

janat08:
I see the packaged there. You might try removing the profile option if there're collisions.

bfred-it:
> You might try removing the profile option if there're collisions.

I did that and it works. For some reason it doesn't work in the default profile even if I completely remove Refined GitHub.

janat08:
lol, the profile option just crapped out on me too. I'm stuck with secure password I won't begin to remember, crap scroll speed, and crap text size/scaling. :D

PR #1931, opened 2019-04-14
sorting params issues/pr affecting top navbar

### LINKED ISSUE #1253: Don't clutter the global search bar in the global Pull Requests page
Hello, I think I encountered a bug. When I click on the "pull request" button on the navbar, the filters get into both input of the page instead of just being in the filter input.

image

janat08:
I did this by narrowing down selectors to links. the sorting still works. Open issues tab.

fixes #1253
it's strongly emphasised that the bug should be fixed, with the suggestion that the whole behaviour of feature could be refined in what it manifests.

### DIFF
--- a/source/features/sort-issues-by-update-time.tsx
+++ b/source/features/sort-issues-by-update-time.tsx
@@ -29,8 +29,8 @@ function getDefaultQuery(link, search) {
 function init() {
 	// Get issues links that don't already have a specific sorting applied
 	for (const link of select.all<HTMLAnchorElement>(`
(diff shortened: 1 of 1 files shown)


bfred-it:
I don’t think this fixes it, read the comments I left on that thread. 

The selector already selects items that have a `href` attribute… which are only `a`

janat08:
Why don't really know why it worked, as I recognized that input wouldn't have that attribute.

bfred-it:
I tested it and it didn't fix it, the global search field is still filled in, as I suggested earlier

PR #1932, opened 2019-04-14
Add `default-to-rich-diff` feature

### LINKED ISSUE #1599: Default to rich diffs for SVG files
When SVG files are edited, usually the rich diff (rendered image) is more useful/desirable than textual diff.

This is particularly useful when such files are edited in an editor rather than by hand, since the changes are usually extensive enough to go over the size limit for showing the diff automatically, so neither the code nor the rich diff are shown by default (the "Load diff" link is shown instead). See for example pretty much any PR at icon repos, such as https://github.com/elementary/icons/pulls.

idasbiste:
I've used a similar approach as the one in #1924. With `MutationObserver` we can watch changes in the DOM for files that are loaded asynchronously and added to the page _a posteriori_.

This functionality can be extended to other files (markdown ones, for example): just add the intended extension to the array defined in `default-to-rich-diff.tsx::7`.

Example of a PR with SVG files: https://github.com/elementary/icons/pull/792 (as the issue's OP said "pretty much any PR at icon repos, such as https://github.com/elementary/icons/pulls")

Closes #1599


source/features/default-to-rich-diff.tsx
```diff
@@ -0,0 +1,47 @@
...
+				setRichDiff(element);
+			}
+		}
+	}
+});
+
+function setRichDiff(node: Element) {
+	for (const fileHeader of select.all('.file-header', node)) {
+		if (checkFileIsRichDiffable(select('.file-info > a', fileHeader).title)) {
+			select('[aria-label="Display the rich diff"]', fileHeader).click();
```
bfred-it: You can select this in the for of and click it. If file `isRichDiffable` then it will have this button

source/features/default-to-rich-diff.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+const fileExtensionRegExp = new RegExp('\\.(' + richDiffExtensions.join('|') + ')$');
+
+const checkFileIsRichDiffable = (filename: string): boolean => fileExtensionRegExp.test(filename);
+
+const observer = new MutationObserver(([{addedNodes}]) => {
+	for (const node of addedNodes) {
+		const element = node as Element;
+		if (element.tagName === 'DIV') {
+			if (element.className === 'js-diff-progressive-container') {
```
bfred-it: You can merge these two conditions into `element.matches('.js-diff-progressive-container')`

source/features/default-to-rich-diff.tsx
```diff
@@ -0,0 +1,47 @@
+import select from 'select-dom';
+import features from '../libs/features';
+
+const childList = {childList: true};
```
bfred-it: No need to keep this as a variable

source/features/default-to-rich-diff.tsx
idasbiste: Good point and I can gladly change this but do we want to default to the rich diff mode regardless of the type of file? I've added the check function to filter by the desired extensions (SVG in this case).

source/features/default-to-rich-diff.tsx
idasbiste: Some of the added nodes can be text nodes and, in that case, `matches` is not supported, I believe.

bfred-it: You can skip the `as` and use `node instanceof Element && element.matches(…)`

source/features/default-to-rich-diff.tsx
bfred-it: Good question, we need to see what other files have rich diffs

bfred-it: Maybe not, Markdown files also have a rich diff but their code is usually handcrafted. 

In this case I’d like to see the selector specify SVG files if the info is available in an attribute or className somewhere. If not, the `isRichDiffable` function should just replaced by `/\.svg$/i.test(filename)`. Adding extensions in the future is easy

idasbiste: `.file-header` does have a `data-file-type` attribute with the extension of the file 👍 
I'll pick it from there.

source/features/default-to-rich-diff.tsx
```diff
@@ -0,0 +1,35 @@
+import select from 'select-dom';
+import features from '../libs/features';
+
+const observer = new MutationObserver(([{addedNodes}]) => {
+	for (const node of addedNodes) {
+		if (node instanceof Element) {
+			if (node.matches('.js-diff-progressive-container')) {
```
idasbiste: @bfred-it I wasn't able to use both conditions in the same if because we would get errors when using a non-Element node as BaseElement of `setRichDiff`.

sindresorhus:
Can you rebase from master (we enabled [TS strict mode](https://github.com/sindresorhus/refined-github/commit/7b3b275ad9a1b1409c56310c0b8ab286bef1f75f)) and fix any issues?

source/features/default-to-rich-diff.tsx
```diff
@@ -0,0 +1,40 @@
...
+function setRichDiff(node: Element | null) {
+	if (node) {
+		for (const fileHeader of select.all('.file-header[data-file-type=".svg"] [aria-label="Display the rich diff"]', node)) {
+			fileHeader.click();
+		}
+	}
+}
+
+function init() {
+	setRichDiff(select('.js-diff-progressive-container:first-child'));
```
sindresorhus: You could move the `.js-diff-progressive-container:first-child` to the `select.all` call in `setRichDiff`.

bfred-it:
I simplified the loading mechanism (I'd still like this moved to its own file), however I noticed a pretty big problem:

1. If GitHub's JS isn't loaded, the click will cause the form to be submitted to an empty page and navigate away
2. We can avoid this problem with `preventDefault`, but the Rich Preview button will remain unclickable and therefore the user can't see the preview at all

idasbiste:
@bfred-it 
I think we can't do much more than preventing the default behaviour. 
If for some reason one of the two GitHub's JS files is not loaded:
• framework: the files section is not loaded
• bootstrap: some actions (such as the rich diff preview) are unavailable and we should _respect_ it.
Don't you agree?

idasbiste:
@bfred-it 
We might be able to revert the button disable attribute after clicking:
```
// hacky...
button.addEventListener('click', event => setTimeout(() => {
	(event.target as Element).closest('button')!.removeAttribute('disabled')
}, 1000));
```
Or use `MutationObserver` on these elements looking for attribute changes and trigger an attribute removal if the element gets disabled.

bfred-it:
That restores the button and at least the user can click it, even though I'd rather have a solution that keeps trying to load it. From my tests it looks like no further requests are triggered if we keep clicking it and re-enabling it until it's `.selected`

This would let us drop the 1000ms timeout.

You can use something like this where we have `button.click()`

```js
const interval = setInterval(() => {
	if (button.classList.contains('selected')) {
		clearInterval(interval);
	} else {
		button.disabled = false;
		button.click();
	}
}, 300);
```

sindresorhus:
> @sindresorhus is the feature name ok?

👍

.github/CODEOWNERS
```diff
@@ -1,2 +1,3 @@
 source/features/filter-pr-by-build-status.tsx @HardikModha
 source/features/tags-dropdown.tsx @HardikModha
+source/features/default-to-rich-diff.tsx @idasbiste
```
bfred-it: `\n` before EOF

```suggestion
source/features/default-to-rich-diff.tsx @idasbiste

```

readme.md
```diff
@@ -123,6 +123,7 @@ GitHub Enterprise is also supported. More info in the options.
 - [Insert collapsible content when writing comments (via `<details>`).](https://user-images.githubusercontent.com/1402241/53678019-0c721280-3cf4-11e9-9c24-4d11a697f67c.png)
 - [Filter pull requests by their build status.](https://user-images.githubusercontent.com/22439276/56372372-7733ca80-621c-11e9-8b60-a0b95aa4cd4f.png)
 - [View raw version of files in PRs and commits.](https://user-images.githubusercontent.com/1402241/56484988-b99f2500-6504-11e9-9748-c944e1070cc8.png)
+- [SVG files in a PR default to rich-diff view.](https://user-images.githubusercontent.com/5243867/57125552-c08a2b00-6d81-11e9-9b84-cdb535baa98e.png)
```
bfred-it: This could probably be under "More info at a glance"

idasbiste: I was actually wondering where I should put it. Fine by me :)

bfred-it:
# 🌈

PR #1933, opened 2019-04-16
Don't clutter the global search bar in the global Pull Requests page

bfred-it:
Fixes https://github.com/sindresorhus/refined-github/issues/1253


sindresorhus:
Seems good to me. Can you rebase from master?

Issue #1735, opened 2019-01-18
Convert the whole codebase to TypeScript and make it strict


sindresorhus:
- [x] Convert the remaining files to TypeScript: 
  https://github.com/sindresorhus/refined-github/tree/master/source/libs
  https://github.com/sindresorhus/refined-github/tree/master/source
- [ ] Enable strict mode: https://github.com/sindresorhus/refined-github/blob/11b927189bab258a226e251c1bd87504feee51f4/tsconfig.json#L10
- [ ] Enable these rules again and fix the lint errors: https://github.com/sindresorhus/refined-github/blob/e90b58c16f58f97a0758de05ae7fe01bf7f65687/package.json#L77-L78

mohamedmansour:
@sindresorhus  why did you name them .tsx file? and not .ts ?

hansott:
@mohamedmansour Some files contain JSX, e.g. https://github.com/sindresorhus/refined-github/blob/master/source/features/trending-menu-item.tsx

mohamedmansour:
I Didn't see some files have it, majority do not. Okay :)

nickytonline:
I can grab this. I'm pretty comfortable in TypeScript land.

sindresorhus:
@nicksnyder Awesome! Go ahead 🎉

nicksnyder:
Wrong Nick :)

Raika:
I can help too.
where should I start?

nickytonline:
@sindresorhus @bfred-it, I've started the next step to make things strict. I just wanted to check with both of you how you see the strictness being implemented.

There's lots of examples in the codebase that always assumes that something exists, e.g.

```
const commentedNode = el.parentNode.nextSibling as Text;
```

There are a few ways to approach this:

```
// Short and sweet with the non-null assertion operator
const commentedNode = el!.parentNode!.nextSibling as Text;
```

or

```
// Can kill readbility
const commentedNode = (el as Element).(parentNode as Element).nextSibling as Text;
```

or

```
// Lots of boileplate for existence checks but better guarantees
if (!(el.parentNode && el.parentNode.nextsibling)) {
	return;
}

const commentedNode = el.parentNode.nextSibling as Text;
```

- The first way reads well and we just tell TS we know it exists
- Second way is potentially harder to read if properties are deeply nested and we just tell TS we know it exists.
- Third way can get more verbose with all the existence checks, but is more robust.

[...]

Issue #803, opened 2017-11-07
Parentheses in URLs break linkifying


hkdobrev:
See this line:
- [in a blob](https://github.com/hkdobrev/mdn-polyfills/blob/b1b5697bb4625e1758606f69ea80ad0ab81d68f6/src/Node.prototype.append/index.js#L1)
- [in a diff](https://github.com/msn0/mdn-polyfills/pull/35/files#diff-76215d6ba87ebfa4e77973aad1a13724R1)

The URL contains parentheses which make the anchor cut off the URL in both the `href` and the visual link producing a wrong URL.

fregante:
Context: https://github.com/sindresorhus/linkify-urls/commit/aca71b8855c0b13286308119e427184b8158cbed

TL;DR: support was removed because we don't want the final `)` inside markdown to be matched, .e.g `[github](https://github.com)` -> `https://github.com)`

hkdobrev:
I think this should be covered by a more sophisticated regex which would make the matching non-greedy for the last parenthesis if it detects a markdown link. This may prove difficult, though and we could work around it by detecting if you're looking at markdown or not.

sindresorhus:
See: https://github.com/kevva/url-regex/pull/35#issuecomment-342757205 I think this can be fixed now. We need to use the lookbehind operator in the `url-regex` module and then use it in `linkify-urls` again.

fregante:
Blocked by https://bugzilla.mozilla.org/show_bug.cgi?id=1225665

linkify-urls@3 implemented it but we're stuck on v2 because Firefox doesn't support it.

PR #1936, opened 2019-04-18
WIP - Remove some package types from global

nickytonline:
Relates to cleanup for #1735. See https://github.com/sindresorhus/refined-github/pull/1783#discussion_r276507475


source/features/linkify-urls-in-code.tsx
```diff
@@ -14,7 +14,7 @@ export const linkifiedURLClass = 'rgh-linkified-code';
 const currentRepo = getOwnerAndRepo();
 const options = {
 	user: currentRepo.ownerName || '/',
-	repo: currentRepo.repoName || '/',
+	repository: currentRepo.repoName || '/',
```
nickytonline: Just need to sort out the signature for the `fn` parameter as the options are different.

```bash
ERROR in /Users/nickytonline/dev/refined-github/source/features/linkify-urls-in-code.tsx
[tsl] ERROR in /Users/nickytonline/dev/refined-github/source/features/linkify-urls-in-code.tsx(35,21)
      TS2349: Cannot invoke an expression whose type lacks a call signature. Type '{ (string: string, options: TypeDomOptions): DocumentFragment; (string: string, options?: Options | undefined): string; } | { (string: string, options: TypeDomOptions): DocumentFragment; (string: string, options?: Options | undefined): string; }' has no compatible call signatures.
```

package.json
```diff
@@ -29,8 +29,8 @@
 		"indent-textarea": "^1.0.4",
 		"insert-text-textarea": "^1.0.3",
 		"intervalometer": "^1.0.4",
-		"linkify-issues": "^1.3.0",
-		"linkify-urls": "^2.2.0",
+		"linkify-issues": "^2.0.0",
+		"linkify-urls": "^3.1.0",
```
bfred-it: This can't be updated because Firefox doesn't support lookbehind.

https://github.com/sindresorhus/refined-github/issues/803#issuecomment-474172304

bfred-it:
By the way the types for `linkify-*` have juuuuust been added to their original repos, so we could drop them altogether... except for linkify-urls, unless you can port v3's types to v2 and let @sindresorhus publish that as 2.2.1

package.json
sindresorhus: We should at least update the comment about using the bundled type defs when it's implemented in Firefox and link to the Firefox issue: https://bugzilla.mozilla.org/show_bug.cgi?id=1225665

sindresorhus:
`linkify-issues` uses lookbehinds now too.

sindresorhus:
> unless you can port v3's types to v2 and let @sindresorhus publish that as 2.2.1

Nah. Too much trouble for it to be worth it. We can just wait it out.

package.json
bfred-it: Perhaps v3's types can be backported to v2 and you and you can publish v2.2.1

bfred-it:
> We can just wait it out.

I want to remind you that some bugs on bugzilla are older than some RGH contributors

sindresorhus:
> I want to remind you that some bugs on bugzilla are older than some RGH contributors

`¯\_(ツ)_/¯` They have pressure now that people are starting to use it. It will be implemented within the end of this year at the latest. I'm 100% sure.

bfred-it:
There are some type todo’s that would be good to complete as well

package.json
nickytonline: Just checking if we'll add the types to v.2.2.1. If so, I can do this tonight. If not we can leave them in RGH globals.d.ts with the TODO pointing to this comment as @bfred-it mentions, https://github.com/sindresorhus/refined-github/pull/1962#discussion_r277802529

sindresorhus: > Just checking if we'll add the types to v.2.2.1.

No, just leave them for now.

PR #1942, opened 2019-04-18
Fix `pr-branches` and `remove-projects-tab` features

### LINKED ISSUE #1935: console errors
- [ ] On pr page

```
❌ pr-branches
content.js:1865 Error: GraphQL:
Argument 'number' on Field 'NaN' has an invalid value. Expected type 'Int!'.
Argument 'number' on Field 'NaN' has an invalid value. Expected type 'Int!'.
Argument 'number' on Field 'NaN' has an invalid value. Expected type 'Int!'.
Argument 'number' on Field 'NaN' has an invalid value. Expected type 'Int!'.
Argument 'number' on Field 'NaN' has an invalid value. Expected type 'Int!'.
Argument 'number' on Field 'NaN' has an invalid value. Expected type 'Int!'.
Argument 'number' on Field 'NaN' has an invalid value. Expected type 'Int!'.
Argument 'number' on Field 'NaN' has an invalid value. Expected type 'Int!'.
Argument 'number' on Field 'NaN' has an invalid value. Expected type 'Int!'.
Argument 'number' on Field 'NaN' has an invalid value. Expected type 'Int!'.
Argument 'number' on Field 'NaN' has an invalid value. Expected type 'Int!'.
Argument 'number' on Field 'NaN' has an invalid value. Expected type 'Int!'.
Argument 'number' on Field 'NaN' has an invalid value. Expected type 'Int!'.
Argument 'number' on Field 'NaN' has an invalid value. Expected type 'Int!'.
Argument 'number' on Field 'NaN' has an invalid value. Expected type 'Int!'.
Argument 'number' on Field 'NaN' has an invalid value. Expected type 'Int!'.
Argument 'number' on Field 'NaN' has an invalid value. Expected type 'Int!'.
```

- [ ] Almost all pages

```
❌ remove-projects-tab
content.js:1865 TypeError: Cannot read property 'parentElement' of null
    at content.js:2490
    at content.js:478
    at remove_projects_tab_init (content.js:2501)
    at async features_run (content.js:1860)
```

- [x] Doesnt always give the error but it stopped working: https://github.com/sindresorhus/refined-github/issues/1940

```
❌ show-names
content.js:1865 TypeError: Cannot read property 'name' of undefined
    at show_names_init (content.js:4586)
    at async details.init (content.js:1903)
    at async features_run (content.js:1860)

```

HardikModha:
1. `pr-branches`: 
   - An alphanumeric string was being passed to `Number()` causing it to return `NaN` and eventually lead to failure of GraphQL query.
   - https://github.com/sindresorhus/refined-github/issues/1935#issuecomment-484849946

2. `remove-projects-tab`: Selector was renamed from `HeaderMenu` to `Header`.

3. `branch-buttons`: Feature scope was set to `isRepo` so it was breaking on some due to lack of the expected selector. Enabled the features on `isRepoRoot` and `isSingleFile`.

4. `releases-tab`
   - https://github.com/sindresorhus/refined-github/issues/1935#issuecomment-484849946

Fixes #1935
Closes #1935

### DIFF
--- a/source/features/pr-branches.tsx
+++ b/source/features/pr-branches.tsx
@@ -73,7 +73,7 @@ function buildQuery(issueIds: string[]) {
 	return `{
 		repository(owner: "${ownerName}", name: "${repoName}") {
 			${issueIds.map(id => `
-				${id}: pullRequest(number: ${String(id).replace('issue_', '')}) {
+				${id}: pullRequest(number: ${id.replace('issue_', '')}) {
 					baseRef {id}
 					headRef {id}
 					baseRefName
--- a/source/features/pr-branches.tsx
+++ b/source/features/pr-branches.tsx
@@ -67,13 +67,13 @@ function normalizeBranchInfo(data: BranchInfo): {
 	};
 }
 
-function buildQuery(numbers: number[]) {
+function buildQuery(issueIds: string[]) {
 	const {ownerName, repoName} = getOwnerAndRepo();
 
 	return `{
 		repository(owner: "${ownerName}", name: "${repoName}") {
-			${numbers.map(number => `
-				${number}: pullRequest(number: ${String(number).replace('issue_', '')}) {
+			${issueIds.map(id => `
+				${id}: pullRequest(number: ${String(id).replace('issue_', '')}) {
 					baseRef {id}
 					headRef {id}
 					baseRefName
@@ -109,7 +109,7 @@ async function init(): Promise<false | void> {
 	}
 
 	const {ownerName} = getOwnerAndRepo();
-	const query = buildQuery(elements.map(pr => Number(pr.id)));
+	const query = buildQuery(elements.map(pr => pr.id));
 	const [data, defaultBranch] = await Promise.all([
 		api.v4(query),
 		getDefaultBranch()
--- a/source/features/branch-buttons.tsx
+++ b/source/features/branch-buttons.tsx
@@ -97,7 +97,7 @@ async function init(): Promise<false | void> {
 		return false;
 	}
 
-	// If buttons are already added. Don't add again.
+	// If buttons are already added, don't add again.
 	// https://github.com/sindresorhus/refined-github/issues/1935#issuecomment-484849946
 	if (select.exists('.rgh-branch-buttons')) {
 		return false;
--- a/source/features/branch-buttons.tsx
+++ b/source/features/branch-buttons.tsx
@@ -121,7 +121,7 @@ async function init(): Promise<false | void> {
 features.add({
 	id: 'branch-buttons',
 	include: [
-		features.isRepoRoot,
+		features.isRepoTree,
 		features.isSingleFile
 	],
 	load: features.onAjaxedPages,
--- a/source/features/branch-buttons.tsx
+++ b/source/features/branch-buttons.tsx
@@ -128,7 +128,7 @@ features.add({
 	id: 'branch-buttons',
 	include: [
 		features.isRepoRoot,
-		features.isSingleFile,
+		features.isSingleFile
 	],
 	load: features.onAjaxedPages,
 	init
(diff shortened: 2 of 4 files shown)


source/features/pr-branches.tsx
```diff
@@ -67,13 +67,13 @@ function normalizeBranchInfo(data: BranchInfo): {
...
-function buildQuery(numbers: number[]) {
+function buildQuery(issueIds: string[]) {
 	const {ownerName, repoName} = getOwnerAndRepo();
 
 	return `{
 		repository(owner: "${ownerName}", name: "${repoName}") {
-			${numbers.map(number => `
-				${number}: pullRequest(number: ${String(number).replace('issue_', '')}) {
+			${issueIds.map(id => `
+				${id}: pullRequest(number: ${String(id).replace('issue_', '')}) {
```
bfred-it: It’s already a string now, so this can be dropped

source/features/pr-branches.tsx
```diff
@@ -67,13 +67,13 @@ function normalizeBranchInfo(data: BranchInfo): {
 	};
 }
 
-function buildQuery(numbers: number[]) {
+function buildQuery(issueIds: string[]) {
```
bfred-it: No need to rename the variables, while not of type `number` they are still issue “numbers”

source/features/pr-branches.tsx
HardikModha: Duh..Missed it. 😅

source/features/pr-branches.tsx
HardikModha: But IMHO, `issueIds` make more sense and are relevant to the context. They are "issue" ids.

HardikModha: It's actually `issue_1234` not just `1234` so I changed it to `issueIds`.

yakov116:
@HardikModha want to tackle this last one?

https://github.com/sindresorhus/refined-github/edit/master/.editorconfig
```
❌ branch-buttons
content.js:1880 TypeError: Cannot read property 'textContent' of null
    at getDefaultBranchLink (content.js:3760)
    at async branch_buttons_init (content.js:3783)
    at async features_run (content.js:1875)
```

HardikModha:
@yakov116 I'm not seeing this error. The feature is working fine for me.

source/features/pr-branches.tsx
bfred-it: 🤦‍♂️ indeed, I thought they were just `1234`

bfred-it:
> @yakov116 I'm not seeing this error. The feature is working fine for me.

It's there on the linked page

> content.js:1880 TypeError: Cannot read property 'textContent' of null

This is because the feature runs on all `isRepo` pages, it should be reduced to where the required elements appear (without adding more `if`s) like `isRepoTree`, `isSingleFile`, `isCommit`, etc

HardikModha:
Damn...and here I was thinking of why @yakov116 has linked `.editorconfig` here. :sweat_smile:

bfred-it:
Wrong direction, follow this:

> This is because the feature runs on all isRepo pages, it should be reduced to where the required elements appear (without adding more ifs) like isRepoTree, isSingleFile, isCommit, etc

It's not specifically about "Edit File" but it's `isRepo` was too broad

HardikModha:
@bfred-it  So we just add? 🤔  

```

include: [
    features.isRepoTree,
	features.isSingleCommit,
	features.isSingleFile,
],
```

bfred-it:
Yep, and maybe more that I don't remember

HardikModha:
@bfred-it  I went through all the pages and only pages having `.breadcrumb` are getting covered in the two features namely `features.isRepoRoot` and `features.isSingleFile`.

bfred-it:
I think this feature only makes sense when there’s a branch selector, because the feature is an extension of that, but we can probably detect this via page detects

bfred-it:
Thanks @HardikModha 🕊

I dropped the changes related to my latest comment in the other thread as that requires a proper solution in `view-markdown-source`

PR #1740, opened 2019-01-21
Use `eslint-config-xo-typescript`

sindresorhus:

PR #1943, opened 2019-04-18
Added missing explicit return types and turned eslint rule back on

nickytonline:
Here you go @bfred-it. This fixes missing explicit return types as per @sindresorhus comment, https://github.com/sindresorhus/refined-github/pull/1740#discussion_r251200144


source/features/branch-buttons.tsx
```diff
@@ -61,7 +61,7 @@ async function getTagLink() {
 	return link;
 }
 
-async function getDefaultBranchLink() {
+async function getDefaultBranchLink(): Promise<HTMLElement | SVGElement | undefined> {
```
bfred-it:
```suggestion
async function getDefaultBranchLink(): Promise<HTMLAnchorElement | undefined> {
```

source/features/faster-pr-diff-options.tsx
```diff
@@ -3,14 +3,14 @@ import select from 'select-dom';
 import * as icons from '../libs/icons';
 import features from '../libs/features';
 
-function createDiffStyleToggle() {
+function createDiffStyleToggle(): JSX.Element {
```
bfred-it: I'm confused, why does `HTMLElement` work in `getDefaultBranchLink` but not here? They both use JSX

source/features/follow-file-renames.tsx
```diff
@@ -13,7 +13,7 @@ async function findRename(
 	user: string,
 	repo: string,
 	lastCommitOnPage: string
-) {
+): Promise<Record<string, any>> {
```
bfred-it:
```suggestion
): Promise<AnyObject> {
```

source/features/releases-tab.tsx
```diff
@@ -20,7 +20,7 @@ const repoKey = `releases-count:${repoUrl}`;
 // Get as soon as possible, to have it ready before the first paint
 const cached = cache.get(repoKey);
 
-function updateReleasesCount() {
+function updateReleasesCount(): number | Promise<any> {
```
bfred-it: Perhaps?

```suggestion
function updateReleasesCount(): number | Promise<number> {
```

source/libs/cache.ts
```diff
@@ -20,7 +20,7 @@ export async function getSet(key: string, getter: CacheGetter, expiration?: numb
 	}
 }
 
-export async function get(key: string) {
+export async function get(key: string): Promise<any> {
```
bfred-it: Maybe
```suggestion
export async function get<T extends any>(key: string): Promise<T> {
```

source/libs/get-default-branch.ts
```diff
@@ -26,14 +26,14 @@ function parseBranchFromDom() {
 	return branchName; // `string` or undefined
 }
 
-async function fetchFromApi(user: string, repo: string) {
+async function fetchFromApi(user: string, repo: string): Promise<any> {
```
bfred-it:
```suggestion
// TODO: drop `response &&`
async function fetchFromApi(user: string, repo: string): Promise<any> {
```

source/features/releases-tab.tsx
nickytonline: I tried this, but it means you'd have to `await` `cached`, e.g. `return Number(await cached)` because cached values are always strings from what I can tell, which works, but it seems odd to await after as well where it's called. I don't think there is a performance issue because if the Promise is already resolved it will return right away.

If you do a `get<T>(key)` like you mentioned below, it'd probably work. The only thing is it's a string coerced into a number versus `Number(...)`

source/features/branch-buttons.tsx
nickytonline: This can't be HTMLAnchorElement. This kind of comes back to some of the JSX issues I mentioned way back, https://github.com/sindresorhus/refined-github/pull/1783#issuecomment-464412087.

`JSX.Element` is just an alias for `HTMLElement | SVGElement`, so I'll use that for now.

bfred-it: We can't use `HTMLElement | undefined` either?

source/features/faster-pr-diff-options.tsx
nickytonline: As mentioned above, `JSX.Element` is simply a type alias for `HTMLElement | SVGElement` which is why it works above.

source/features/follow-file-renames.tsx
nickytonline: Done.

source/features/releases-tab.tsx
bfred-it: I'm not sure, I just tried `number | Promise<number>` locally and neither TS nor XO complain

source/features/follow-file-renames.tsx
nickytonline: Didn't do this one as I just realized I can commit your suggestion, but did the one below.

source/features/releases-tab.tsx
nickytonline: Ahh it doesn't complain because the cache returns `Promise<any>` which is why `Promise<number>` works even though it's a string when returned from the cache.

source/features/branch-buttons.tsx
nickytonline: Doesn't seem to work. Like I said before, the JSX issue is still kind of around. We have a somewhat working solution with types, but it's so so. I gotta 😴 but I'll look at this some more tomorrow.

bfred-it:

```sh
```

source/features/branch-buttons.tsx
```diff
@@ -38,7 +38,7 @@ async function getTagLink(): Promise<'' | HTMLAnchorElement> {
 		latestRelease = tags[0];
 	}
 
-	const link = <a className="btn btn-sm btn-outline tooltipped tooltipped-ne">{icons.tag()}</a> as HTMLAnchorElement;
+	const link = <a className="btn btn-sm btn-outline tooltipped tooltipped-ne">{icons.tag()}</a> as unknown as HTMLAnchorElement;
```
bfred-it: This is the only drawback of this new approach. The element is still internally consistent (`<a>` accepts a `href` attribute but not a `value`) but once out it's a plain `HTMLElement` and we get this error without `as unknown`:


> source/features/branch-buttons.tsx:41:15 - error TS2352: Conversion of type '`Element`' to type '`HTMLAnchorElement`' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
>   Type '`Element`' is missing the following properties from type '`HTMLAnchorElement`': `charset`, `coords`, `download`, `hreflang`, and 20 more.
> 
> 41  `const link = <a className="btn btn-sm btn-outline tooltipped tooltipped-ne">{icons.tag()}</a>  as HTMLAnchorElement;`
>

source/globals.d.ts
```diff
@@ -64,8 +64,7 @@ interface GlobalEventHandlersEventMap {
 }
 
 declare namespace JSX {
-	type Element = Element | DocumentFragment;
```
bfred-it: `type Element = Element |` was actually a loop and `JSX.Element` had become `any`

I wish TS threw an error on this

source/features/branch-buttons.tsx
nickytonline: Yeah, I really want to nail down getting these types right for dom-chef. I think dom-chef is the first library to use JSX in a non-React/React like library which is why these issues are popping up.

source/globals.d.ts
nickytonline: I think here though, the `JSX.Element` type is a union type on the global `Element` (not itself) and `DocumentFragment` no?

source/libs/cache.ts
nickytonline: We can make this change, but from what I've seen in the code base `get` in JS land is always returning a `string` no? So we can type it as `Promise<T>` with the caveat that the string will be coerced to whatever type T is.

source/globals.d.ts
bfred-it: No, I think it was itself, hence the `any` that appeared. 

I saw React’s types create identical types outside the scope to avoid the problem, e.g.:

``` 
type NativeEvent = Event;
namespace React {
    type Event = BlahBlah<NativeEvent>
}

File: source/features/branch-buttons.tsx (hunk 7 above)
Comment (bfred-it) [reply source/features/branch-buttons.tsx]: There’s another one called `h` I think but it’s abandoned. I’m sure that there are more as I occasionally find comments of people having similar type issues

File: source/globals.d.ts (hunk 8 above)
Comment (nickytonline) [reply source/globals.d.ts]: It would be nice to get the types nailed down though for dom-chef. I'm just trying to avoid a rewrite of React types to accommodate our changes, but maybe that's what needs to be done. YOLO. 😉

File: source/features/branch-buttons.tsx (hunk 7 above)
Comment (nickytonline) [reply source/features/branch-buttons.tsx]: Yeah, `h` is what Preact uses for vdom.

File: source/libs/cache.ts (hunk 5 above)
Comment (bfred-it) [reply source/libs/cache.ts]: > `get` in JS land is always returning a `string` no?

Which `get`? Why would it only return a string? `Map#get` doesn't 🤔

File: source/features/branch-buttons.tsx (hunk 7 above)
Comment (bfred-it) [reply source/features/branch-buttons.tsx]: I'm referring to https://www.npmjs.com/package/h, it's dom-based. There's also:

https://github.com/glixlur/jsx-dom
https://github.com/egoist/dom-dom
https://github.com/treycordova/nativejsx (this was promising but it didn't support SVG)
https://github.com/oychao/jsx-dom-render
https://github.com/qwtel/create-element-x

and more: https://npms.io/search?q=dom+jsx

Maybe they have some good ideas we can learn from

Comment (nickytonline) [reply source/features/branch-buttons.tsx]: Yeah using the `unknown` was what I suggested at the beginning but wrapping it in a function to avoid `unknown` everywhere, but that means wrapping all JSX in a function whose main purpose is just to please the TypeScript compiler. https://github.com/sindresorhus/refined-github/issues/1735#issuecomment-462015923

Comment (bfred-it) [reply source/features/branch-buttons.tsx]: All of those new `as unknown` in 9a527dffb45f84742306d3032c110a6d36f7aa34 were not necessary, it was a wrong merge, probably. Only **this one** was necessary.

I sent a new merge

File: source/libs/cache.ts (hunk 5 above)
Comment (nickytonline) [reply source/libs/cache.ts]: Ahh, yeah, nevermind that comment. The only thing to alter is it should return `Promise<T | undefined>` because of 

```javascript
	// If it's not in the cache, it's best to return "undefined"
	if (value === null || value === undefined) {
		return undefined;
	}
```

nickytonline: I think we're good to go after that.

nickytonline: Actually, even if we type the get, it's possible undefined so you're not guaranteed a number in the case of the releases tab.

image

I'm going to wrap this up tonight (just starting my day 😉), but feel free to jump on this in the meantime if you want.

nickytonline: Ahh, I remember why I said it was always a string.

image

I'm guessinf the `CacheGetter` interface should be generic, `T | Promise<T>` instead of returning `string | Promise<string>`?

bfred-it: Ahh indeed. Values are stored in `document.cookie` so yes, they're always strings.

source/features/releases-tab.tsx
```diff
@@ -18,9 +18,9 @@ const repoUrl = getRepoURL();
 const repoKey = `releases-count:${repoUrl}`;
 
 // Get as soon as possible, to have it ready before the first paint
-const cached = cache.get(repoKey);
+const cached = cache.get<number>(repoKey);
```
nickytonline: So the cache is generic now and it's fine. I misread the code before when I said it was always returning a string from the cache.

It actually does a `JSON.stringify` before storing the value in the cookie and a `JSON.parse` when it pulls it out of the cookie, so the stored data has the correct type.

source/features/releases-tab.tsx
```diff
@@ -18,9 +18,9 @@ const repoUrl = getRepoURL();
...
 
-function updateReleasesCount(): number | Promise<number> {
+async function updateReleasesCount(): Promise<number | undefined> {
```
nickytonline: I made this `async` so that we could just return a Promise since non-Promise values get wrapped as a resolved Promise.

source/libs/cache.ts
```diff
@@ -1,26 +1,28 @@
-type CacheGetter = () => string|Promise<string>;
+ type CacheGetter<TValue extends any = any> = () => TValue | Promise<TValue>;
```
nickytonline: This was only type as string because it was only used in the releases tab feature. Now it's generic even if it's only used in one place at the moment.

source/libs/cache.ts
```diff
@@ -1,26 +1,28 @@
...
+	const cache = await get<TValue>(key);
 	if (cache !== undefined) {
 		return cache;
 	}
 
 	const value = await getter();
-	if (value !== undefined) {
-		await set(key, value, expiration);
-		return value;
+	if (value === undefined) {
```
nickytonline: TypeScript complains if I don't do an explicit return for undefined. We could turn the rule off for this function if you want to get rid of the explicit return.

nickytonline:
Now that the cache is generic, I think this is good to go. Let me know your thoughts on the cache changes @bfred-it.

PR #1945, opened 2019-04-19
Remove projects 

### LINKED ISSUE #977: Remove Projects from the sidebar if there are no Projects
Similar to the removal of the Projects tab if there are no projects, I think we should remove the Projects section in the issue/PR sidebar if there are no projects to select there.

There are both repository and organisation projects in there so we should check the contents of the selector instead of using the number of projects in the repo as we do for the repository tab.

You cannot create Projects from that selector so there's no reason you would need it unless you already have projects created.

janat08:
this page removes projects, or on a issue page.
closes #977

### DIFF
--- a/source/features/remove-projects-sidebar.tsx
+++ b/source/features/remove-projects-sidebar.tsx
@@ -4,19 +4,19 @@ The sidebar can't be used to create new projects.
 New projects can still be created via the [`Create new…` menu](https://user-images.githubusercontent.com/1402241/34909214-18b6fb2e-f8cf-11e7-8556-bed748596d3b.png).
 */
 
-import select from 'select-dom';
 import * as api from '../libs/api';
 import features from '../libs/features';
 import {getOwnerAndRepo} from '../libs/utils';
+import { safeElementReady } from '../libs/dom-utils';
 
-async function init(): Promise<false | void> {
-	const project = select('[aria-label="Select projects"]');
-	if (!project) {
-		return false;
-	}
+var loading = false
+var res: Promise<any>
 
-	const {ownerName, repoName} = getOwnerAndRepo();
-	const {repository} = await api.v4(`
+function query (){
+	const { ownerName, repoName } = getOwnerAndRepo();
+	// return await onetime((){}())
+	loading = true
+	res = api.v4(`
 		{
 			repository(owner: "${ownerName}", name: "${repoName}") {
 				projects(states: [OPEN, CLOSED]){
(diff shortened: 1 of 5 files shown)


janat08:
btw i'm not sure if api in question actually checks organization projects, since the issue description explains that organizations appear to have global/top-level projects that are included in repos projects menu.

source/features/remove-projects-sidebar.tsx
```diff
@@ -0,0 +1,43 @@
...
+    		}
+		}
+	}`;
+}
+
+async function init() {
+	const res = await api.v4(buildQuery());
+	const project = await safeElementReady(`
+		[aria-label="Select projects"]
+	`);
```
bfred-it: Since here you're using `load: features.onAjaxedPages`, `safeElementReady` is the same as `select`, so it's best to use that

source/features/remove-projects-sidebar.tsx
```diff
@@ -0,0 +1,42 @@
+/*
+The `Projects` tab is hidden from repositories and profiles when there are no projects
+
+New projects can still be created via the [`Create new…` menu](https://user-images.githubusercontent.com/1402241/34909214-18b6fb2e-f8cf-11e7-8556-bed748596d3b.png).
+*/
```
bfred-it: Also this header is a copy of `remove-project-tab`

source/features/remove-projects-sidebar.tsx
```diff
@@ -0,0 +1,42 @@
+/*
+The `Projects` sidebar is hidden from repositories and profiles when there are no projects.
+The sidebar can't be used to create new projects.
+New projects can still be created via the [`Create new…` menu](https://user-images.githubusercontent.com/1402241/34909214-18b6fb2e-f8cf-11e7-8556-bed748596d3b.png).
```
bfred-it: Unrelated. The first line is enough

janat08: I just rewrote the feature to get the results without flicker.

janat08: It's not linted, but I'm trying to make point of not putting on pretenses. If this still doesn't agree with you I'd rather be compensated all the same, and avoid making any PRs in future.

bfred-it: It's good to see that

bfred-it: > I'd rather be compensated all the same

For the record, bounties are just rewards, not compensation. PRs with bounties can get rejected just like any other.

sindresorhus: > It's not linted, but I'm trying to make point of not putting on pretenses.

What does this mean?

sindresorhus: > PRs with bounties can get rejected just like any other.

They are not exactly like any other PR in that we expect the PR submitter to put in extra effort in their work.

janat08:
we could probably remove all other sidebars in the same fashion. So arguably this could be more about the sidebar than about projects being cleaned up.

PR #1947, opened 2019-04-19
Meta: Remove most TypeScript `any` types

nickytonline:
Just some more cleanup post #1735.

### DIFF
--- a/source/features/follow-file-renames.tsx
+++ b/source/features/follow-file-renames.tsx
@@ -9,7 +9,7 @@ import * as api from '../libs/api';
 import {getCleanPathname} from '../libs/utils';
 
 interface File {
-	previous_filename: string;
+	previous_filename: string; // eslint-disable-line @typescript-eslint/camelcase
 	filename: string;
 	status: string;
 }
--- a/source/features/follow-file-renames.tsx
+++ b/source/features/follow-file-renames.tsx
@@ -8,14 +8,18 @@ import features from '../libs/features';
 import * as api from '../libs/api';
 import {getCleanPathname} from '../libs/utils';
 
+interface Content {
+	files: Iterable<Record<string, string>>;
+}
+
 // TODO: ensure that pages with a single commit aren't fetched twice (api.ts' cache should do it automatically, after #1783 is merged)
 async function findRename(
 	user: string,
 	repo: string,
 	lastCommitOnPage: string
-) {
+): Promise<Content> {
 	// API v4 doesn't support it: https://github.community/t5/GitHub-API-Development-and/What-is-the-corresponding-object-in-GraphQL-API-v4-for-patch/m-p/14502?collapse_discussion=true&filter=location&location=board:api&q=files%20changed%20commit&search_type=thread
-	return api.v3(`repos/${user}/${repo}/commits/${lastCommitOnPage}`);
+	return api.v3(`repos/${user}/${repo}/commits/${lastCommitOnPage}`) as Promise<Content>;
 }
 
 async function init(): Promise<false | void> {
--- a/source/features/follow-file-renames.tsx
+++ b/source/features/follow-file-renames.tsx
@@ -8,8 +8,10 @@ import features from '../libs/features';
 import * as api from '../libs/api';
 import {getCleanPathname} from '../libs/utils';
 
-interface Content {
-	files: Iterable<Record<string, string>>;
+interface File {
+	previous_filename: string;
+	filename: string;
+	status: string;
 }
 
 // TODO: ensure that pages with a single commit aren't fetched twice (api.ts' cache should do it automatically, after #1783 is merged)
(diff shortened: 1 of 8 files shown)


source/features/follow-file-renames.tsx
```diff
@@ -8,14 +8,18 @@ import features from '../libs/features';
...
 // TODO: ensure that pages with a single commit aren't fetched twice (api.ts' cache should do it automatically, after #1783 is merged)
 async function findRename(
 	user: string,
 	repo: string,
 	lastCommitOnPage: string
-) {
+): Promise<Content> {
 	// API v4 doesn't support it: https://github.community/t5/GitHub-API-Development-and/What-is-the-corresponding-object-in-GraphQL-API-v4-for-patch/m-p/14502?collapse_discussion=true&filter=location&location=board:api&q=files%20changed%20commit&search_type=thread
-	return api.v3(`repos/${user}/${repo}/commits/${lastCommitOnPage}`);
+	return api.v3(`repos/${user}/${repo}/commits/${lastCommitOnPage}`) as Promise<Content>;
```
nickytonline: I'm casting here, but in the future it might be nice to genericise the payload of `v3` and `v4` so we can do this instead:

```javascript
v3<Content>(`repos/${user}/${repo}/commits/${lastCommitOnPage}`) // returns Promise<Content>
```

sindresorhus: I think we should make it generic, but default the default generic parameter to `any` to make it flexible. Then we can use generics if we want or not, depending on the context.

source/resolve-conflicts.ts
```diff
@@ -10,8 +10,12 @@ declare module CodeMirror {
 	}
 }
 
-const select: typeof document.querySelector = document.querySelector.bind(document);
-const editor: CodeMirrorInstance = select<any>('.CodeMirror').CodeMirror;
+interface EditorLinkedDom extends HTMLElement {
+	CodeMirror: CodeMirrorInstance;
+}
```
sindresorhus: I don't really see the benefit of this. 

```ts
const editor = select<any>('.CodeMirror').CodeMirror as CodeMirrorInstance;
```

is good enough for one-off stuff like this.

@bfred-it ?

nickytonline: That's just me conditioned to remove `any`s whenever possible 😉

source/features/follow-file-renames.tsx
bfred-it: I don't think this solves anything, let's please leave it as `AnyObject`, I don't want infinitely-nested types like the ones I dropped here:

https://github.com/sindresorhus/refined-github/pull/1783/commits/9fbc1931169f5a5228bf86c1a8873c59c8aa0d98

Previous discussion: https://github.com/sindresorhus/refined-github/pull/1783#discussion_r259628430

source/resolve-conflicts.ts
bfred-it: `any` and `as` sound like bandaids but I'd rather have them in these cases, e.g. https://github.com/sindresorhus/refined-github/pull/1783/commits/399a0b16595f8b426d0121b03309112c2d8d0507 and https://github.com/sindresorhus/refined-github/pull/1783/commits/44b37ce1dd68793f58770e34600a7918e28b8657

nickytonline: I'll revert that change then.

source/resolve-conflicts.ts
```diff
@@ -10,12 +10,8 @@ declare module CodeMirror {
 	}
 }
 
-interface EditorLinkedDom extends HTMLElement {
-	CodeMirror: CodeMirrorInstance;
-}
-
-const select = document.querySelector.bind(document);
-const editor = select<EditorLinkedDom>('.CodeMirror')!.CodeMirror;
+const select: typeof document.querySelector = document.querySelector.bind(document);
```
bfred-it: I just realized `select` is no longer used anywhere else, so you can just drop this altogether 😅

nickytonline: Done.

nickytonline:
> @nickytonline regarding typing the API, how does this sound?
> 
> > Only type the parsed API output, not the received JSON
> 
> i.e. the API result in `fetchCoAuthoredData` should not get typed until it leaves the function, if ever, which is until we get the plain `Author` array.
> 
> This means that `Content` can stay, but the code should probably be adjusted so it returns `Promise<File[]>` where `File` has the fields `previous_filename`, `filename` and `status`

Happy to make the change, but reading through this again and I'm not following your comment. The `Content` interface is used on the `findRename` function in the follow-filerenames feature, not `fetchCoAuthoredData` function. Is that what you meant?

bfred-it:
Yes, I was talking about the API-related interfaces introduced by this PR

bfred-it:
🤖 🚀

bfred-it:
Yet another great TypeScript PR! We can add you as a collaborator to Refined GitHub if you want :)

Issue #1236, opened 2018-03-31
"Update Branch" button for stale branches


rarkins:
[<img alt="Issuehunt badges" src="https://issuehunt.io/r/sindresorhus/refined-github/issues/1236/badge.svg" />](https://issuehunt.io/r/sindresorhus/refined-github/issues/1236)

Similar to what was discussed by @vanniktech in #1105 but not exactly the same, because I have no interest in a "hot key" - just a button.

Summary:

If you enable branch protection for a repository and select "Require branches to be up-to-date before merging", *then* you get an Update Branch button whenever the branch goes stale.

![image](https://user-images.githubusercontent.com/6311784/38159810-cd0f73b8-34b0-11e8-928c-0ffad49b10f6.png)

However it would be convenient to have this for all my PRs, including ones in repositories that don't enforce this branch protection. 

As described by @hkdobrev in #1105:
> This button would merge the default branch in the PR branch with a merge commit if there are no conflicts.

It might be a *little* tricky to work out who has write access to that branch when it's coming from a fork. In theory this feature/button could initially work for branches within the same repository only though.

> There is a $80.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/sindresorhus/refined-github/issues/1236).
> - Checkout the [Issuehunt explorer](https://issuehunt.io/r/sindresorhus/refined-github/) to discover more funded issues.
> - Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on Issuehunt to raise funds.

bfred-it:
This is interesting, I needed it [recently](https://github.com/sindresorhus/refined-github/pull/1229#commits-pushed-b090b33) to avoid conflicts (merge + commit = ok; commit + merge = conflict) but I did it via CLI instead.

This might be possible [via API](https://developer.github.com/v3/repos/merging/) but we don't have API authentication yet in our code.

The second issue is that we'd have to make another API every time the thread updates to check if the PR is up to date and add the notice+button (avoidable if we just add the button somewhere).

I'm thinking this could be more generically implemented in the Compare page, to have an option to merge without a PR:

<img width="690" alt="" src="https://user-images.githubusercontent.com/1402241/38291666-f3f59982-380a-11e8-8b27-408ee3dd0184.png">

On top of this, to facilitate merging of the current PR, we could add a link to this page somewhere, if it makes sense. Example for the out-of-date https://github.com/sindresorhus/refined-github/tree/no-reaction-comments

[...]

PR #1950, opened 2019-04-19
Add `update-pr-from-base-branch` feature

bfred-it:
Closes https://github.com/sindresorhus/refined-github/issues/1236

- It works only when the branches are on the same repo, because of API limitations
- It works even if `master` was force-pushed
- The name is simplified to `master` because people _get it_ but it works on any branch

# Tasks

- [x] Merge with master, it needs a description field because of #2030

<details><summary>Edit: all resolved</summary>
- [x] Make less prominent, this shouldn't look like a necessary step.

    **HELP:** _Where/how should we show this button? Does it make sense there? This is what I got so far, which doesn't look right (warning icon, too prominent)_

- [x] The button might not appear sometimes; I'll fix this once we figure out where to place it 👆
- [x] Only show if necessary, like when a branch is stale or when GH already shows it 

    **HELP:** _how do we find out whether a branch is stale?_ Edit: maybe this https://developer.github.com/v3/repos/commits/#compare-two-commits

- [x] Handle errors. I'll probably change the API to just `ignoreHTTPStatus: false` instead of `accept404`. API reference: https://developer.github.com/v3/repos/merging/
- [x] Handle conflict (i.e. console error) with `add-co-authors` (fixed in #2019)
- [x] Handle possible caching issue after merging (the button reappears)
- [-] Update readme

cc @rarkins @sindresorhus 
</details>

# Test

Test it on any PR. It should depend on:

- User has permission to merge or not
- PR can be merged or not (e.g. conflicts or reviews required)
- PR type (draft or normal)

~~Outdated screenshot https://user-images.githubusercontent.com/1402241/56434750-842de800-6308-11e9-9092-75cd703d4414.png~~


bfred-it:
The commit above is done with this feature 🎉 https://github.com/sindresorhus/refined-github/pull/1950/commits/371960a3d7b7dfc3e12714e102c36342e5b61889

sindresorhus:
> Only show if necessary, like when a branch is stale or when GH already shows it

Yes, only show it if the PR branch behind `master`.

> Make less prominent, this shouldn't look like a necessary step.

Agreed. It's only a helpful utility, not a required operation. Not sure where to place it though. Maybe a small link here: 
Screen Shot 2019-04-26 at 13 09 45

source/features/update-pr-from-master.tsx
```diff
@@ -0,0 +1,62 @@
+import React from 'dom-chef';
+import select from 'select-dom';
+import features from '../libs/features';
+import * as api from '../libs/api';
+import * as icons from '../libs/icons';
+
+export async function mergeBranches(repoUrl: string, base: string, head: string) {
+	const result = await api.v3(`repos/${repoUrl}/merges`, {
```
sindresorhus: The `v4` API doesn't support this?

rarkins:
Re: default vs master terminology. Isn't it technically a third option - i.e. whatever is the "base" branch for the PR?

For example if my default branch is `master` but I have another branch called something like `v3` which is for the next major release, then I may have PRs targeting `v3` as the base branch and therefore to sync it with default/master would be incorrect. In such situations we should probably update it from `v3` or hide the option if it is not possible.

bfred-it:
I used `master` also because we have another feature named after it, but `base` is actually more correct because that's what this feature looks for.

source/features/update-pr-from-master.tsx
bfred-it: I think not, and even then the REST query is easier to build

bfred-it:
conflict

bfred-it:
Even better, link to `/conflicts` when there are conflicts in Draft PRs

sindresorhus:
Seems to work, however:

> Only show if necessary, like when a branch is stale or when GH already shows it

This doesn't work correctly. I made a PR (https://github.com/sindresorhus/sindre-playground/pull/4/), then updated a file in master, and reloaded the PR, got the "Update branch" button, clicked it, got "Updating branch...", then it disappeared, and then reappeared again, if I click it now, it just switches to "Updating branch...", but nothing happens. If I reload the page 1 minute later, the button no longer shows.

I have a feeling that this is caused by the API taking some time to update its status or something. Still, it's not a very good experience for users.

bfred-it:
> got the "Update branch" button, clicked it, got "Updating branch...", then it disappeared, and then reappeared again, if I click it now, it just switches to "Updating branch...", but nothing happens. If I reload the page 1 minute later, the button no longer shows.

Yeah that’s the second-last to-do on my list, I think I’ll just have to disable the feature after the merge because it’s unlikely to be needed right away

bfred-it:
I think this is good to go

Screenshot 2019-05-15 at 23 53 10

sindresorhus:
Looks good now. The only problem is that when you click "Update branch", then it updates, and you refresh the page, the button is there again, as the API is slow to update... I don't think we can do anything about that though.

rarkins:
Thanks all

PR #1963, opened 2019-04-22
Upgrade dependencies

sindresorhus:


source/libs/page-detect.ts
```diff
@@ -1,4 +1,4 @@
-/* eslint-disable unicorn/prefer-starts-ends-with */
+/* eslint-disable unicorn/prefer-starts-ends-with, @typescript-eslint/prefer-string-starts-ends-with */
```
sindresorhus: @bfred-it I didn't want to touch this as we already ignored a similar rule here, but I think we should fix this. Relying on `RegExp#test` coercing `undefined` to a string can be dangerous, as `undefined` might actually match something, causing a false-positive.

source/features/notification.d.ts
```diff
@@ -1,5 +1,8 @@
 export interface Notification {
-	participants: { username: string | null; avatar: string }[];
+	participants: Array<{
+		username: string | null;
+		avatar: string;
+	}>;
```
bfred-it: Why `Array`?

source/libs/page-detect.ts
bfred-it: Ugh, you're right. `/undefined/.test(undefined) => true`

source/features/notification.d.ts
bfred-it: Also, worth dropping that `null`?

source/features/mark-unread.tsx
```diff
@@ -270,7 +270,7 @@ function isCurrentSingleRepoPage({repository}: Notification): boolean {
 }
 
 function isParticipatingPage(): boolean {
-	return /\/notifications\/participating/.test(location.pathname);
+	return location.pathname.includes('/notifications/participating');
```
bfred-it: Probably

```suggestion
	return location.pathname.startsWith('/notifications/participating');
```

source/features/notification.d.ts
sindresorhus: It's more readable to only use `type[]` shorthand for simple types.

https://github.com/xojs/eslint-config-xo-typescript/commit/7b1f11891a4c2a3282792fcb139c7e134fe7f7dd

> Also, worth dropping that null?

Yes, but it's not really related to dependency updates.

source/features/mark-unread.tsx
sindresorhus: Yup, it's `https://github.com/notifications/participating`.

source/features/notification.d.ts
sindresorhus: Code compiles without the `null`, so I've removed it.

Issue #421, opened 2017-05-13
Remove Milestones filter if there are no milestones


hkdobrev:
![image](https://cloud.githubusercontent.com/assets/506129/26025465/29c6047c-37f1-11e7-8c20-32e3a3a6559c.png)

When there are no milestones the filter in issues and PRs listing page in a repo still have a filter for milestones which is not useful at all. There's just one option which just shows all issues/PRs with no milestones which are basically every issue/PR.

I think this would reduce clutter for projects not using milestones and not change anything if you're using milestones.

sindresorhus:
👍

busches:
I took a stab at this, but the milestone list isn't loaded until you attempt to filter on it. Is there another way we can determine if they're using milestones outside of what's in the list and if not, do we want to call the list all the time just to remove it?

sindresorhus:
@busches Use `MutationObserver`.

busches:
@sindresorhus that seems clunky though. I'd click on `Milestones` and then it'd load the list, it has no milestones, and we'd immediately remove it?

sindresorhus:
MutationObserver is pretty fast, so not sure it would even be visible before it was removed. Any other suggestions?

sindresorhus:
Ok, I misunderstood how it worked.

bfred-it:
1. Fetch and cache Milestones count at earlier convenience
2. Use it
3. Make it expire after a day (expiration just means storing the retrieval date, it doesn't need to be cleaned up since it's just a handful of bytes: `{'sindresorhus/refined-github': [0, 1501042347424]})

bfred-it:
This requires a fetch to the API to check if a project has any Milestones. Is it worth it?

PR #1965, opened 2019-04-22
Remove projects from sidebar if there are no projects

### LINKED ISSUE #977: Remove Projects from the sidebar if there are no Projects

notlmn:
Closes #977
Replaces and closes #1945

This features removes the projects section in the issue and PR sidebar if the user doesn't have any permissions or if there are no projects to select.


bfred-it:
- `safeElementReady` doesn't apply to `onAjaxedPages` because `onAjaxedPages` waits for dom ready
- only `isPRConversation` has the sidebar

notlmn:
> Great solution! It works already, however I'd like to see some caching, like in `releases-tab.tsx`:

Wouldn't that conflict with organization level projects. What I mean is that if we cache count for each repo and a project is added at organization level, the user might have to wait until the cache expires.

Not saying its a bad idea, but thinking if there would be any problems, because releases is for a single repo, but projects work at both repo and organization level!? 🤔🤷‍♂️

bfred-it:
That raises another question: is this piece of code correctly counting the projects in organizations as well?

And also: do closed projects appear in the dropdown?

> the user might have to wait until the cache expires

That's kinda the issue with all caches. The workaround would be to have a short cache (1 day)

bfred-it:
Alternatively, instead of hiding the Projects, we could condense the sidebar:

- hide all the _Empty_ messages from the sidebar:
  
    


- hide all the empty sections, the current user can't do anything about them


    

Related feature request to keep in mind: https://github.com/sindresorhus/refined-github/issues/421

notlmn:
> That raises another question: is this piece of code correctly counting the projects in organizations as well?

Yup, the partially loaded DOM contains item listing for both repository and organization projects combined in their own containers. Here we are counting these combined projects ([example](https://github.com/notlmn-dump/test/issues/1/show_partial?partial=issues%2Fsidebar%2Fprojects_menu_content)).

-------------

> And also: do closed projects appear in the dropdown?

Just tested it, no they don't. 🎉

--------------

> That's kinda the issue with all caches. The workaround would be to have a short cache (1 day)

I'll see what I can do with caching, might take time though.

notlmn:
> Alternatively, instead of hiding the Projects, we could condense the sidebar

Already thought of generalizing this feature hide all the empty sidebar stuff, but stopped after looking at "Assignees", that looks like a useful thing to have even when it doesn't have anything to show, or maybe that's just me.

Open to other thoughts too! 😉

bfred-it:
@sindresorhus agreed that this is the way to go, `clean-discussion-sidebar`: https://github.com/sindresorhus/refined-github/pull/1965#issuecomment-485492676

No fetches; just hide "no content" if the user is a collaborator or hide the whole section for all other users.

I would go as far as hiding the section header for `Labels` and `Notifications`:

clean-sidebar

bfred-it:
That's not what I meant 😅

> 2. Hide "None yet", "No milestone", etc for collaborators

The section should stay there for collaborators, it should only hide that extra text. This feature should not require any `fetch`es

notlmn:
> The section should stay there for collaborators, it should only hide that extra text. This feature should not require any `fetch`es

That does not actually align with what #977 and #421 sounds, but can be manageable. Will update in a bit.

notlmn:
<details>
<summary>Screenshots</summary>

## Repo with no permissions
### PR to some repo
| Before | After |
| :-: | :-: |
| 1-before | 1-after |

### Issue on some repo
| Before | After |
| :-: | :-: |
| 2-before | 2-after |

## Repo with permissions
### Assigned issue with projects
| Before | After |
| :-: | :-: |
| 3-before | 3-after |

### Non-assigned issue with projects
| Before | After |
| :-: | :-: |
| 4-before | 4-after |

</details>

-----------------

> 1. Hide empty sections for regular users

Done.

-----------------

> 3. Move "Assign yourself" next to "Assignees" header, if empty

That's not possible, the header itself is a `<details>` element, can't put `button` inside another interactive element. Can `position` it `absolute`ly, but I don't see the point in that.

notlmn:
> 4. I would go as far as hiding the section header for `Labels` and `Notifications`

I don't know good this looks like. Is this okay?
save

bfred-it:
> The section should stay there for collaborators, it **should only hide that extra text. This feature should not require any `fetch`es**

This

Issue #1908, opened 2019-04-03
Wrap code in comments


bfred-it:
[<img alt="Issuehunt badges" src="https://issuehunt.io/r/sindresorhus/refined-github/issues/1908/badge.svg" />](https://issuehunt.io/r/sindresorhus/refined-github/issues/1908)

GitHub is inconsistent. Code wraps in PR/commit diffs, but not in comments:

# PR (wraps)

<img width="825" alt="" src="https://user-images.githubusercontent.com/1402241/55483237-c42e7300-5658-11e9-94da-6799bf924ed4.png">

# Review comment (no wrap)
<img width="667" alt="" src="https://user-images.githubusercontent.com/1402241/55483231-c264af80-5658-11e9-9be9-302af0311763.png">

# Code block in comment (no wrap)

<img width="696" alt="" src="https://user-images.githubusercontent.com/1402241/55483349-fcce4c80-5658-11e9-9cad-733415c5b2af.png">

> There is a $80.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/sindresorhus/refined-github/issues/1908).
> - Checkout the [Issuehunt explorer](https://issuehunt.io/r/sindresorhus/refined-github/) to discover more funded issues.
> - Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on Issuehunt to raise funds.

notlmn:
> GitHub is inconsistent. Code wraps in PR/commit diffs, but not in comments:

To be frank, it is. They __do not wrap__ content if the code is shown in __unified view__ (like code comments in a PR "conversation" view, [example](https://github.com/sindresorhus/refined-github/pull/1896/files/65c298247d0197a94d4b199e911f1c1f5ecf140d?diff=unified#r271735560)), but __wrap__ content in the __split view__ ([example](https://github.com/sindresorhus/refined-github/pull/1896/files/65c298247d0197a94d4b199e911f1c1f5ecf140d?diff=split#r271735560)).

Because not wrapping content in the split view would cause a lot of problems, impossible actually, because both the left and right parts of the diff are contained in a single `tr` element, making it impossible to isolate scrollbars for left and right diff views.

Even if you did "unwrap" the lines, then there would be scrollbar for each line in the diff.

### Wrapped code (in split diff view, even for comments)
![image](https://user-images.githubusercontent.com/37769974/55496161-04dbbb80-565c-11e9-8d5b-c1d50df379d5.png)

[...]

PR #1967, opened 2019-04-22
Wrap code in comments

bfred-it:
Closes https://github.com/sindresorhus/refined-github/issues/1908

Anything else missing?

I avoided wrapping plain <code>```</code> blocks because it might get hard to read without line numbers.

### DIFF
--- a/source/content.css
+++ b/source/content.css
@@ -1119,6 +1119,7 @@ body > .footer li a {
 	text-align: center;
 }
 
+
 /* Improve dropdown expansion animation (e.g. top-right dropdown in comments) */
 /* It normally pops in from the center */
 .dropdown-menu-sw {
--- a/source/content.css
+++ b/source/content.css
@@ -1119,6 +1119,12 @@ body > .footer li a {
 	text-align: center;
 }
 
+/* Enable word wrap on code in comments (with line numbers only) */
+.discussion-item-body .blob-num ~ .blob-code-inner, /* Embedded code */
+.discussion-item-body .blob-num ~ .blob-code .blob-code-inner { /* Review diffs and suggestions */
+	white-space: pre-wrap;
+	word-wrap: break-word;
+}
 
 /* Improve dropdown expansion animation (e.g. top-right dropdown in comments) */
 /* It normally pops in from the center */


source/content.css
```diff
@@ -1119,6 +1119,13 @@ body > .footer li a {
 	text-align: center;
 }
 
+/* Enable word wrap on code in comments (with line numbers only) */
+.discussion-item-body .blob-num ~ .blob-code-inner, /* Embedded code */
+.discussion-item-body .blob-num ~ .blob-code .blob-code-inner { /* Review diffs and suggestions */
```
bfred-it: 🥇  Lines above should be wrapped.
```suggestion
.wrap-it-all .discussion-item-body .blob-num ~ .blob-code .blob-code-inner { /* Review diffs and suggestions */
```
🥈 Lines above should be wrapped.

https://github.com/sindresorhus/refined-github/blob/b48f289f835fa4447743fe0541dcc3d84c46b063/source/content.css#L358-L359

🥉 Lines above should be wrapped.

notlmn:
Uh @bfred-it, shouldn't soft-wrapped code lines be align with their parent lines?

#### This PR
image

#### Intended soft-wrapped code (?)
image

bfred-it:
That's not how GitHub wraps code natively though

notlmn:
> That's not how GitHub wraps code natively though

Yup agree, I think we should fix that with this PR. Don't ya think!?

bfred-it:
That's a whole 'nother can of expensive worms, especially on responsive views like `PR Files`, where we'd have to look through all the lines and add/remove padding

notlmn:
> That's a whole 'nother can of expensive worms, especially on responsive views like `PR Files`, where we'd have to look through all the lines and add/remove padding

I think you only need to move stuff once, and not whenever you resize window.

Not trying to be rude or anything but it looks like this feature only does **hard-wrapping**, not **soft-wrapping**!?

If you ask me, I'd rather have have a scrollbar than hard-wrapped code. IMHO, hard-wrapped code is way more work for the brain to process.

notlmn:
### Disclaimer
@bfred-it I've been giving suggestions but the both of us started working on this at the same time (😅). [Here](https://github.com/notlmn/refined-github/tree/soft-wrap)'s my implementation using JS, implementing soft-wrapping. You're open to use the code if you want to.

sindresorhus:
> I avoided wrapping plain ``` blocks because it might get hard to read without line numbers.

Would be nice to support this by adding line numbers there, but can be done in a separate PR.

sindresorhus:
Seems to work well, but would much prefer actual soft-wrapping, like @notlmn mentioned. I agree performance is a concern though.

bfred-it:
@notlmn please open a new PR to replace this one :)

bfred-it:
For the record, both are soft-wrapping. Hard-wrapping refers to adding `\n` in the code itself while soft wrapping is a feature of the editor/viewer

PR #1970, opened 2019-04-23
Fix filter-by-pr-status when no Pull Requests exist

busches:
This addresses the issue called out here: https://github.com/sindresorhus/refined-github/pull/1904#issuecomment-485615635

I didn't want to return on the first selector, as if GH changes the DOM, then we won't realize the feature broke as easily.

### DIFF
--- a/source/features/filter-pr-by-build-status.tsx
+++ b/source/features/filter-pr-by-build-status.tsx
@@ -41,6 +41,11 @@ function populateDropDown({currentTarget}: Event): void {
 }
 
 function init(): void {
+	// If there are no PRs then return
+	if (!select('.table-list-filters')) {
+		return;
+	}
+
 	const reviewsFilter = select('.table-list-header-toggle > details:nth-last-child(3)')!;
 
 	// Copy existing element and adapt its content
--- a/source/features/filter-pr-by-build-status.tsx
+++ b/source/features/filter-pr-by-build-status.tsx
@@ -40,13 +40,12 @@ function populateDropDown({currentTarget}: Event): void {
 	}
 }
 
-function init(): void {
-	// If there are no PRs then return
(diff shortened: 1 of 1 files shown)


source/features/filter-pr-by-build-status.tsx
```diff
@@ -41,6 +41,11 @@ function populateDropDown({currentTarget}: Event): void {
 }
 
 function init(): void {
+	// If there are no PRs then return
+	if (!select('.table-list-filters')) {
```
HardikModha: Prefer `select.exists`. We are following the same everywhere.

bfred-it:
The error is on `reviewFilter`, we can just add the if there. Also it should return false.

source/features/filter-pr-by-build-status.tsx
busches: Duh, I'll fix it.

busches:
@bfred-it I opted for the secondary check since if GH changes markup the feature will be disabled instead of failing. Same risk with my fix, but it's a lot less specific. Can it return false if we have it typed as void?

bfred-it:
> if GH changes markup the feature

We can't really do _anything_ about it. GH can change `details` just like it can change `.table-list-filters`, we can't say which one is more likely to happen. 

Lately I've been trying to avoid these `!exists` checks to let the errors through, but sometimes they're necessary like in this case.

busches:
> > if GH changes markup the feature
> 
> We can't really do _anything_ about it. GH can change `details` just like it can change `.table-list-filters`, we can't say which one is more likely to happen.
> 
> Lately I've been trying to avoid these `!exists` checks to let the errors through, but sometimes they're necessary like in this case.

Alright, you've convinced me. :)

PR #1973, opened 2019-04-23
Fix `recently-pushed-branches-enhancements` on GHE

busches:
GHE still uses the old fixed with header, along with different markup/classes.

### DIFF
--- a/source/features/recently-pushed-branches-enhancements.tsx
+++ b/source/features/recently-pushed-branches-enhancements.tsx
@@ -54,7 +54,7 @@ async function init(): Promise<false | void> {
 	document.body.classList.add('rgh-recently-pushed-branches');
 
 	// Move or add list next to the notifications bell
-	select('.Header-item--full')!.after(widget);
+	select('.Header-item--full,.HeaderMenu nav')!.after(widget);
 }
 
 features.add({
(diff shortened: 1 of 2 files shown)


busches:
And found it was duplicated on a page still, one more fix incoming.

source/features/recently-pushed-branches-enhancements.tsx
```diff
@@ -54,7 +54,7 @@ async function init(): Promise<false | void> {
 	document.body.classList.add('rgh-recently-pushed-branches');
 
 	// Move or add list next to the notifications bell
-	select('.Header-item--full')!.after(widget);
+	select('.Header-item--full,.HeaderMenu nav')!.after(widget);
```
bfred-it: `HeaderMenu` doesn't exist on regular GitHub, so no chances of conflict. If this works for you we can merge it

busches:
@bfred-it I'm running into this: 
```css
.rgh-recently-pushed-branches [data-url$='recently_touched_branches_list'] p {
	display: none;
}
```
not hiding the original box, how does this not hide both once we move it?

busches:
Actually, now I can't reproduce it, so 🤷‍♂️

PR #1974, opened 2019-04-23
Add Markdown support to `copy-file`

### LINKED ISSUE #1223: Missing copy button in Markdown files
Hey everyone,

I love this extension but I realized that the "Copy" button that is inserted for some file types isn't inserted in Markdown files? Is this by design? I store a bunch of Markdown files on GitHub that I regularly copy and it would be great to have the button there as well :)

If this isn't by design I'm happy to research and fix this issue.

Cheers,
Dominik

janat08:
fixes #1223

stills works:
code: https://github.com/sindresorhus/refined-github/blob/master/source/background.ts
will work once source is loaded:
binary: https://github.com/sindresorhus/refined-github/blob/master/media/icon.svg
markdown: https://github.com/sindresorhus/refined-github/blob/master/contributing.md


janat08:
couldn't load the raw file, don't remember why.

source/features/copy-file.tsx
```diff
@@ -2,25 +2,65 @@ import React from 'dom-chef';
...
+let loaded = false;
+let loading: Promise<any>;
+const isMarkDown = features.isMarkDown();
+
+async function loadFile(): Promise<void> {
+	loaded = true;
+	console.log('loading');
+	loading = new Promise((resolve => {
+		resolve(fetchSource());
+	}));
```
bfred-it: `fetchSource` already returns a promise.

Also this globals-based approach means that if you open `file1.md` and then `file2.md`, the globals will still point to `file1.md`

In `view-markdown-source` I avoided this problem with a WeakMap, so each source is connected to its file (actually, the file's button)

Thankfully we don't need to handle any of this: we can only show the `copy` button when the source is shown, just like it happens for [SVG files](https://github.com/sindresorhus/refined-github/blob/master/media/icon.svg). This means you don't have to call `fetchSource` at all, but rather you can call `.dispatchEvent(new CustomEvent('rgh:view-markdown-source'))` and have `copy-file` listen to it. It's much easier

source/features/view-markdown-source.tsx
```diff
@@ -50,6 +50,8 @@ async function showSource(): Promise<void> {
 	sourceButton.classList.add('selected');
 	renderedButton.classList.remove('selected');
 	blurButton(sourceButton);
+
+	document.dispatchEvent(new CustomEvent('rgh:view-markdown-source', {detail: await source}));
```
bfred-it:
```suggestion
	sourceButton.dispatchEvent(new CustomEvent('rgh:view-markdown-source', {bubbles: true}));
```

source/features/view-markdown-source.tsx
```diff
@@ -12,7 +12,7 @@ import {blurAccessibly} from './comment-fields-keyboard-shortcuts';
 
 const btnBodyMap = new WeakMap<Element, Element | Promise<Element>>();
 
-async function fetchSource(): Promise<Element> {
+export async function fetchSource(): Promise<Element> {
```
bfred-it:
```suggestion
async function fetchSource(): Promise<Element> {
```

source/features/copy-file.tsx
```diff
@@ -3,28 +3,49 @@ import select from 'select-dom';
 import copyToClipboard from 'copy-text-to-clipboard';
 import features from '../libs/features';
 
-function handleClick({currentTarget: button}: React.MouseEvent<HTMLButtonElement>): void {
-	const file = button.closest('.Box');
+function handleClickFactory(markdown?: Element): ({currentTarget: button}: React.MouseEvent<HTMLButtonElement>) => void {
```
bfred-it: You can get to the markdown source from `event.currentTarget`, you don't need the factory

source/features/copy-file.tsx
```diff
@@ -3,28 +3,49 @@ import select from 'select-dom';
...
 }
 
-function init(): void {
-	// This selector skips binaries + markdowns with code
-	for (const code of select.all('.blob-wrapper > .highlight:not(.rgh-copy-file)')) {
-		code.classList.add('rgh-copy-file');
+async function init(): Promise<void> {
+	// This won't show button on markdown and binary files until the code is loaded
+	if (features.isMarkDown() || select.all('.blob-wrapper > .highlight').length !== 0) {
+		document.addEventListener('rgh:view-markdown-source', e => {
```
bfred-it: You can use
```suggestion
		delegate('.rgh-md-source', 'rgh:view-markdown-source', event => {
			renderButton(event.delegateTarget);
```

source/features/copy-file.tsx
```diff
@@ -3,28 +3,49 @@ import select from 'select-dom';
 import copyToClipboard from 'copy-text-to-clipboard';
 import features from '../libs/features';
 
-function handleClick({currentTarget: button}: React.MouseEvent<HTMLButtonElement>): void {
-	const file = button.closest('.Box');
+function handleClickFactory(markdown?: Element): ({currentTarget: button}: React.MouseEvent<HTMLButtonElement>) => void {
+	return function ({currentTarget: button}: React.MouseEvent<HTMLButtonElement>): void {
+		if (features.isMarkDown()) {
```
bfred-it: You don't need this split; you can update the selector in `.blob-code-inner` on line 21 and the rest should work without copying the whole block

janat08: markdown in fact has blob-code elements containing several lines inside, to retain spacing different operations have to be made. Also two selectors are required.

source/features/copy-file.tsx
janat08: this results in type error since `rgh:...` isn't an event type. I'm not sure that it can be overridden.

janat08:
everything but markdown stopped working.

source/features/copy-file.tsx
bfred-it: Two selectors yes, but not two whole almost-identical pieces of code. At most you need:

```js
const content = select.all(isMarkDown() ? '.blame-hunk .blob-code' : '.blob-code-inner', file!) 
```

source/features/copy-file.tsx
bfred-it: You can find the events in `global.d.ts`, next to `details:toggle`

source/features/copy-file.tsx
janat08: I refractored everything out it.

bfred-it: I meant this 696b660

bfred-it: My bad, I meant https://github.com/sindresorhus/refined-github/pull/1974/commits/46464c8dcdaa9ba36d20848e78900edc6cc8854a

bfred-it:
I've been testing it on https://github.com/sindresorhus/refined-github/blob/master/readme.md

Two notes:

- The button should only appear when the code is shown (like for SVG files)
- The button keeps getting added every time you toggle between source and rendered

source/features/copy-file.tsx
janat08: I understood you correctly and refractored that, that split is required for spacing. In it's current state it will actually also miss lines of code. With your permission I will force push my refractors.

bfred-it: Can you point out which lines of code are missed exactly when clicking on Copy in https://github.com/sindresorhus/refined-github/blob/master/readme.md?

janat08: so apparently none, it used to be the case for contributing.

source/features/copy-file.tsx
```diff
@@ -1,26 +1,32 @@
...
-function init(): void {
-	// This selector skips binaries + markdowns with code
-	for (const code of select.all('.blob-wrapper > .highlight:not(.rgh-copy-file)')) {
-		code.classList.add('rgh-copy-file');
-		code
-			.closest('.Box')! // Closest common container
-			.querySelector('[data-hotkey="b"]')! // Easily-found `Blame` button
+async function init(): Promise<void> {
+	// This won't show button on markdown and binary files until the code is loaded
+	if (features.isMarkDown() || select.all('.blob-wrapper > .highlight').length === 0) {
```
bfred-it: Since we're listening to `rgh:view-markdown-source`, we need to match its condition:

```suggestion
	if (select.exists('.blob.instapaper_body')) {
```

bfred-it:
> The button keeps getting added every time you toggle between source and rendered

This happens because this was dropped:

`:not(.rgh-copy-file)`
`code.classList.add('rgh-copy-file')`

janat08:
It was, but I assumed that it was simply the case of button being added on events, so I run existence check. I assume this is equivalent and clearer?

janat08:
I imagine the loading behavior before the refractor around how features are added used to be different.

source/features/copy-file.tsx
```diff
@@ -17,8 +17,19 @@ async function init(): Promise<void> {
...
+			}
 			renderButton();
 		});
+		delegate('.rgh-md-source', 'rgh:view-markdown-rendered', () => {
+			// Prevents several buttons
+			const buttonSelector = '[aria-label="Copy file to clipboard"]'
+			if (select.exists(buttonSelector)){
+				select(buttonSelector)!.remove()
+			}
+		});
```
bfred-it: I think actually we only need this check here, because if we remove it every time it won't be there afterwards.

```js

		delegate('.rgh-md-source', 'rgh:view-markdown-source', renderButton);
		delegate('.rgh-md-source', 'rgh:view-markdown-rendered', () => {
			const button = select('.rgh-copy-file');
			if (button){
				button.remove()
			}
		});
```

I noticed we have a `copy-btn` class on the button, we should rename it to `rgh-copy-file`.

Also

source/features/copy-file.tsx
```diff
@@ -1,26 +1,44 @@
...
-function init(): void {
-	// This selector skips binaries + markdowns with code
-	for (const code of select.all('.blob-wrapper > .highlight:not(.rgh-copy-file)')) {
-		code.classList.add('rgh-copy-file');
-		code
-			.closest('.Box')! // Closest common container
-			.querySelector('[data-hotkey="b"]')! // Easily-found `Blame` button
+async function init(): Promise<void> {
+	// This won't show button on markdown and binary files until the code is loaded
+	if (select.exists('.blob.instapaper_body')) {
```
bfred-it: Now that I think about it, we don't need this `if` at all.

It's easier this way:

```js
renderButton();
delegate('.rgh-md-source', 'rgh:view-markdown-source', renderButton);
delegate('.rgh-md-source', 'rgh:view-markdown-rendered', () => {
	const button = select('.rgh-copy-file');
	if (button){
		button.remove()
	}
});

janat08: that should render the button on markdown files when they're rendered.

janat08: as apposed to being in source.

janat08: or I imagine you count on rendered firing and removing the button?

janat08: that doesn't work

janat08: changing import order didn't work either

bfred-it: Indeed, the selector is no longer `.blob-wrapper > .highlight` so `renderButton` was called even on rendered markdown. I updated it

bfred-it:
Thanks @janat08 🎳

janat08:
can I get the bounty

Issue #966, opened 2018-01-08
Link to "What changed since the previous release?"


bfred-it:
[<img alt="Issuehunt badges" src="https://issuehunt.io/r/sindresorhus/refined-github/issues/966/badge.svg" />](https://issuehunt.io/r/sindresorhus/refined-github/issues/966)

Example: https://github.com/parcel-bundler/parcel/releases

Can you tell what each release includes? GitHub links to the tagged commit but that's [rather useless](https://github.com/parcel-bundler/parcel/commit/68258b1e53d086f173d2888d9fa367fcba891807)

Wouldn't it make sense to list the commits between each tag/release (in chronological order).

This is the link that should appear on `v1.4.1`: https://github.com/parcel-bundler/parcel/compare/v1.4.0...v1.4.1

Without any fetches we can show the link on

- [lists](https://github.com/parcel-bundler/parcel/releases) but not each [tag/release pages](https://github.com/parcel-bundler/parcel/releases/tag/v1.4.1)
- every tag in a list, except the last one

> There is a $80.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/sindresorhus/refined-github/issues/966).
> - Checkout the [Issuehunt explorer](https://issuehunt.io/r/sindresorhus/refined-github/) to discover more funded issues.
> - Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on Issuehunt to raise funds.

hkdobrev:
I think this is similar to that old discussion: https://github.com/sindresorhus/refined-github/pull/395#issuecomment-301189407

👍

sindresorhus:
@tterranigma Please use the "reactions" feature for voting. Commenting without adding anything useful to the discussion is just noise ;)

sindresorhus:
> Without any fetches we can show the link on

We now use the API much more, so we can use the API for this feature too.

IssuehuntBot:
@issuehunt has funded $80.00 to this issue.

---
- Submit pull request via [IssueHunt](https://issuehunt.io/repos/51769689/issues/966) to receive this reward.
- Want to contribute? Chip in to this issue via [IssueHunt](https://issuehunt.io/repos/51769689/issues/966).
- Checkout the [IssueHunt Issue Explorer](https://issuehunt.io/issues) to see more funded issues.
- Need help from developers? [Add your repository](https://issuehunt.io/r/new) on IssueHunt to raise funds.

[...]

PR #1976, opened 2019-04-24
Publish a git tag for every Refined GitHub release

bfred-it:
This will:

- create a versioned tag on cronjob **and** tags called `hotfix`
- publish all other tags using the tag name as version

~~**WARNING:** As of a5aaa54 it already works **from this branch**, so any changes will be published to git releases and the old travis job will pick that up and publish it to the stores. This instant publishing is temporary and for testing only.~~

~~@sindresorhus I'd like to avoid giving my GH API token to travis, so I'd rather set this up with SSH, if possible: 6345bcf~~ _Done_

~~Perhaps this can be done in https://github.com/sindresorhus/refined-github/settings/keys~~ _Done_

Reference:

- https://github.com/alrra/travis-scripts/blob/master/docs/github-deploy-keys.md
- https://help.github.com/en/articles/which-remote-url-should-i-use#cloning-with-ssh-urls


yakov116:
Curious why?

yakov116:
Curious why?

bfred-it:
To see exactly what each version contains. Without it we have to open Travis, find the “cron” build, open it, look for the version in the log, and get the commit from the header. 

Plus I might also be able to attach the built extension to each tag/release, so people who might need it don’t have to download and build the source.

bfred-it:
Great! This seems to work correctly for `hotfix` tags. Now we wait for `cron`

bfred-it:
#966 would help here too

bfred-it:
`cron` also worked (and my regular commit was correctly skipped)



However I noticed that the same commit can be tagged twice. I opened a new issue

PR #1986, opened 2019-04-25
Meta: split CSS by feature /2

bfred-it:
Continues #1983

Basically done. Needs just a minor review.

All commits, except deletion ones, should touch exactly 3 files


yakov116:
@bfred-it I think you should rebase and merge, it may make things easier later. JMHO

bfred-it:
Indeed. How about a merge commit? Does that also work well in the History?

bfred-it:
Good idea @yakov116. I did that. Sadly merge commits [aren't highlighted anymore](https://github.com/sindresorhus/refined-github/issues/1988) so it wouldn't help.

PR #1987, opened 2019-04-25
Convert Webpack config to TypeScript

nickytonline:
As per comment https://github.com/sindresorhus/refined-github/pull/1962#issuecomment-486749235


webpack.config.ts
```diff
@@ -1,12 +1,14 @@
-'use strict';
-const path = require('path');
+import path from 'path';
+import CopyWebpackPlugin from 'copy-webpack-plugin';
+import TerserPlugin from 'terser-webpack-plugin';
+import MiniCssExtractPlugin from 'mini-css-extract-plugin';
+import webpack from 'webpack';
+
+// eslint-disable-next-line @typescript-eslint/no-require-imports,@typescript-eslint/no-var-requires
```
nickytonline: @bfred-it, I added the disabled lint rules here. Other options are adding a `declare module 'size-plugin'` in globals.d.ts or a PR to the https://github.com/GoogleChromeLabs/size-plugin. I can definitely send a PR over there at some point, but just have some other things on the go at the moment.

webpack.config.ts
```diff
@@ -1,12 +1,14 @@
...
 const SizePlugin = require('size-plugin');
-const CopyWebpackPlugin = require('copy-webpack-plugin');
-const TerserPlugin = require('terser-webpack-plugin');
-const MiniCssExtractPlugin = require('mini-css-extract-plugin');
 
-module.exports = (env, argv) => ({
-	devtool: 'sourcemap',
+module.exports = (_env: string, argv: Record<string, boolean|number|string>): webpack.Configuration => ({
```
nickytonline: The `_env` is more complex potentially in terms of it's type, see , but for now, I just kept it to string. Besides, we're not using it at the moment. I was actually surprised to find there was no type for the function -> configuration.

webpack.config.ts
```diff
@@ -1,12 +1,14 @@
...
+	devtool: 'source-map',
```
nickytonline: The types for webpack complained about this, so I put it to what it was expecting.

webpack.config.ts
sindresorhus: > I can definitely send a PR over there at some point, but just have some other things on the go at the moment.

👍

webpack.config.ts
```diff
@@ -1,12 +1,14 @@
...
+
+// eslint-disable-next-line @typescript-eslint/no-require-imports, @typescript-eslint/no-var-requires
 const SizePlugin = require('size-plugin');
-const CopyWebpackPlugin = require('copy-webpack-plugin');
-const TerserPlugin = require('terser-webpack-plugin');
-const MiniCssExtractPlugin = require('mini-css-extract-plugin');
 
-module.exports = (env, argv) => ({
-	devtool: 'sourcemap',
+module.exports = (_env: string, argv: Record<string, boolean | number | string>): webpack.Configuration => ({
```
sindresorhus: Are you sure `argv` is `boolean | number | string`? I'm pretty sure it's just `string`.

nickytonline: It's webpack doing something because the type for `argv.watch` is boolean. I'll double check this type tomorrow. Time for 🛏️

bfred-it: `webpack` parses all arguments to normalize them, so `webpack --production` is received as `{mode: 'production'}`. Maybe even `...{watch: false, etc}`

nickytonline: Yeah, I double checked this morning, they're manipulating `argv` like you said, because normally `process.argv`'s type is `string[]`. The do the same for env like I mention here, https://github.com/sindresorhus/refined-github/pull/1987#discussion_r278762904.

webpack.config.ts
bfred-it: I'd rather have `declare module 'size-plugin'` in globals.d.ts than a comment here + `require` but it doesn't seem to work

nickytonline: I'll make the change and try to get to creating the types for that package in the next week or so, so that we can remove the temp `declare module 'size-plugin'`.

bfred-it: We can do without the TODO, it’s not really something we **need** to type. Once the types are in it’ll just work

source/globals.d.ts
```diff
@@ -61,3 +61,6 @@ declare namespace JSX {
 interface NamedNodeMap {
 	[key: string]: Attr;
 }
+
+
+declare module 'size-plugin';
```
nickytonline: I'll check tonight why this isn't working,

nickytonline: So it turns out that ts-node, which webpack is using under the hood to process the webpack.config.ts, is not able to detect what is set in the globals.d.ts. So what you can do is add a `paths` setting in the tsconfig.json and create a d.ts file for the `size-plugin` package. However, [ts-node is unable to read the paths property in tsconfig.json](https://github.com/TypeStrong/ts-node/issues/138#issuecomment-349222070). To be able to read `paths`, you need to install the [tsconfig-paths-webpack-plugin](https://github.com/dividab/tsconfig-paths-webpack-plugin). This seems like a lot of work just to support missing types of one package. I'd suggest we stay with the disable eslint rules for now. Thoughts @sindresorhus @bfred-it?

sindresorhus: `// @tsignore`?

nickytonline: Yeah, I think that's a good middle ground until the `size-plugin` has types.

nickytonline:
For reference in regards to TypeScript types for the size-plugin package, https://github.com/GoogleChromeLabs/size-plugin/issues/20

source/globals.d.ts
blakeembrey: @nickytonline Just a heads up, `ts-node` does support types for the compiler. That issue, in particular, is more about resolving executable code based on the `paths` config (e.g. like what you can do in Webpack by aliases). If you wanted you could write a definition and it would work.

bfred-it: @blakeembrey how to do that in this case? `./source` is defined in `tsconfig` but why isn't it automatically picked up by ts-node?

blakeembrey: @bfred-it `ts-node` does not use `files`/`include`/`exclude` by default, but it does support `paths` and `typeRoots` by default. However, you can use them by including `--files`. This is mainly because `ts-node` is used in dozens of different ways and by different use-cases, so excluding by default resulted in the right trade-off between performance and flexibility.

blakeembrey: I mainly commented because of this:

> However, ts-node is unable to read the paths property in tsconfig.json. To be able to read paths, you need to install the tsconfig-paths-webpack-plugin.

This is a misunderstanding in that thread. `paths` is supported by `ts-node` in the same way TypeScript supports it. It does not, which most people want, automatically rewrite your imports to do things like `@myapp` to automatically come from `./src`.

PR #1881, opened 2019-03-23
Adding description field to features.add()

sunilkumarc:
@sindresorhus I've now added description field to features.add() and also description text for all of the existing features. Please review and let me know if any changes are required.

Fixes #1787

PR #1640, opened 2018-11-27
Reaction avatars bug fixes 

bfred-it:
Fixes #233 (oldest bug open)

And solves the two ugly problems you see here: 

wrapping + duplicate tooltip

PR #1989, opened 2019-04-26
Add `indented-code-wrapping` feature

### LINKED ISSUE #1908: Wrap code in comments

notlmn:
Closes #1908

Uses JS to detect number of spaces at the start of each line, and then uses CSS to move the block forward.

#### PR conversations and files (dynamically loaded comments, review, and files)
- https://github.com/sindresorhus/refined-github/pull/1881
- https://github.com/parcel-bundler/parcel/pull/2967

#### PRs with suggestions
- https://github.com/sindresorhus/refined-github/pull/1881


source/features/softwrap-code.tsx
```diff
@@ -0,0 +1,52 @@
...
+	for (const table of tables) {
+		if (table.classList.contains('rgh-softwrapped-code')) {
+			continue;
+		}
+
+		table.classList.add('rgh-softwrapped-code');
+
+		const lines = select.all('.blob-code-inner', table);
+		for (const line of lines) {
+			const leadingSpaceCharacters = line.innerHTML.match(/^\s+/);
```
bfred-it: Maybe `line.firstChild.textContent`

source/features/softwrap-code.tsx
```diff
@@ -0,0 +1,52 @@
+/*
+Soft-wrap code inside codeblocks
+*/
+import select from 'select-dom';
+import features from '../libs/features';
+
+const tabSize = 4;
```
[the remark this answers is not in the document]
notlmn: We leave out spaces to be spaces, as is. But tabs are used to compute how many spaces to match. I don't get what you mean by _detect indentation_, because RGH already sets size of tabs to be 4 using CSS.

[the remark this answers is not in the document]
bfred-it: > We leave out spaces to be spaces, as is

Spaces and tabs might be mixed

> RGH already sets size of tabs to be 4 using CSS.

People customize things https://github.com/lukechilds/github-custom-tab-size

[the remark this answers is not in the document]
bfred-it: Maybe you can select the indentation with Range and [get its width](https://stackoverflow.com/a/28457044/288906)

source/features/softwrap-code.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+async function init(): Promise<void> {
+	// Classes:
+	// `.diff-table` - Split and unified diffs
+	// `.d-table` - Suggested changes
+	// `.js-file-line-container` - Embedded code blocks
+	const tables = select.all('.diff-table, .d-table, .js-file-line-container');
```
bfred-it: You can skip this and just select `.blob-code-inner`

notlmn: Need this, as we target the classes added to these tables to change the layout of tables a bit.

source/features/softwrap-code.tsx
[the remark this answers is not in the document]
notlmn: Might've found a better way, please review 2523e31db0104f6988a00eea5269b94389c1cc5b

notlmn:
Temporarily blocked by https://github.com/lukechilds/github-custom-tab-size/pull/3.

bfred-it:
Interesting solutions. Twice on this PR I was like

g9954

source/features/softwrap-code.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+		table.classList.add('rgh-softwrapped-code');
+
+		const lines = select.all('.blob-code-inner', table);
+		for (const line of lines) {
+			if (!line.firstChild || !line.firstChild.textContent) {
+				continue;
+			}
+
+			const leadingSpaceCharacters = line.firstChild.textContent.match(/^\s+/);
```
bfred-it:
```suggestion
			const leadingSpaceCharacters = line.firstChild!.textContent!.match(/^\s+/);
```

This lets you skip the previous `if`

source/features/softwrap-code.tsx
```diff
@@ -0,0 +1,57 @@
...
+			if (leadingSpaceCharacters) {
```
bfred-it:
```suggestion
			if (!leadingSpaceCharacters) {
				continue;
			}
```

source/features/softwrap-code.tsx
```diff
@@ -0,0 +1,57 @@
...
+				let spaceCount = 0;
+				let tabCount = 0;
+				for (const char of leadingSpaceCharacters[0]) {
+					if (char === ' ') {
```
bfred-it: What about all other spaces? Like `&nbsp;`? Perhaps we should just use one condition `if isTab?, else`

source/features/softwrap-code.tsx
```diff
@@ -0,0 +1,57 @@
...
+						spaceCount += 1;
+					} else if (char === '\t') {
+						tabCount += 1;
+					}
+				}
+
+				// Move the whole line where it is supposed to be, then unindent the
+				// start of the line to compensate for indentation, preserving spaces
+				// We get `--tab-size` from RGH or compatible extensions like GCTS
```
bfred-it:
```suggestion
				// We might get `--tab-size` from compatible extensions like `github-custom-tab-size`
```

source/features/softwrap-code.tsx
```diff
@@ -0,0 +1,57 @@
...
+import './softwrap-code.css';
+import select from 'select-dom';
+import features from '../libs/features';
+
+async function init(): Promise<void> {
+	// Classes:
+	// `.diff-table` - Split and unified diffs
+	// `.d-table` - Suggested changes
+	// `.js-file-line-container` - Embedded code blocks
+	const tables = select.all('.diff-table, .d-table, .js-file-line-container');
```
bfred-it:
```suggestion
	const tables = select.all(`
		.diff-table:not(.rgh-softwrapped-code),
		.d-table:not(.rgh-softwrapped-code),
		.js-file-line-container:not(.rgh-softwrapped-code)
	`);
```

This lets you skip 

```js
		if (table.classList.contains('rgh-softwrapped-code')) {
			continue;
		}
```

source/features/softwrap-code.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+	for (const table of tables) {
+		if (table.classList.contains('rgh-softwrapped-code')) {
+			continue;
+		}
+
+		table.classList.add('rgh-softwrapped-code');
+
+		const lines = select.all('.blob-code-inner', table);
+		for (const line of lines) {
```
bfred-it:
```suggestion
		for (const line of select.all('.blob-code-inner', table)) {
```

notlmn:
Doesn't exactly show the change in tab size (Gamebar on Windows), but changing value looks something like this:

source/features/softwrap-code.tsx
```diff
@@ -0,0 +1,57 @@
...
+			}
+		}
+	}
+}
+
+features.add({
+	id: 'softwrap-code',
+	// include: [
+	// 	features.isPRConversation
+	// ],
```
bfred-it: ?

source/features/softwrap-code.tsx
notlmn: I rarely find any `&nbsp;`s in code, but yeah, that seems like a good compromise.

source/features/softwrap-code.tsx
notlmn: That's from an outdated commit, It's changed in the new commits.

notlmn:
Need to look into performance of the code, test it out more on ajaxed diff tables in PR files and reviews, decide on name of this feature, and can then only this PR should be good to go.

source/features/softwrap-code.tsx
```diff
@@ -0,0 +1,48 @@
...
+			// We might get `--tab-size` from compatible extensions like `github-custom-tab-size`
+			line.style.paddingLeft = `calc((var(--tab-size, 4) * ${tabCount}ch) + ${spaceCount}ch)`;
+			line.style.textIndent = `calc((var(--tab-size, 4) * -${tabCount}ch) - ${spaceCount}ch)`;
+		}
+	}
+}
+
+features.add({
+	id: 'softwrap-code',
+	load: features.onAjaxedPages,
```
bfred-it: Needs `include`. This feature only works on code with line numbers.

Also include all test URLs in your first post.

source/features/softwrap-code.tsx
```diff
@@ -0,0 +1,48 @@
...
+			// start of the line to compensate for indentation, preserving spaces
+			// We might get `--tab-size` from compatible extensions like `github-custom-tab-size`
+			line.style.paddingLeft = `calc((var(--tab-size, 4) * ${tabCount}ch) + ${spaceCount}ch)`;
+			line.style.textIndent = `calc((var(--tab-size, 4) * -${tabCount}ch) - ${spaceCount}ch)`;
+		}
+	}
+}
+
+features.add({
+	id: 'softwrap-code',
```
bfred-it: We need a better name because code is already soft-wrapped. Either something like `indentation-aware-code-wrapping`, `improved-code-wrapping`, etc

Ideas?

source/features/softwrap-code.tsx
```diff
@@ -0,0 +1,49 @@
+/*
+Soft-wrap code inside codeblocks
```
bfred-it: We need a better description because code is already soft-wrapped in most places. Same with the feature name

source/features/softwrap-code.css
```diff
@@ -0,0 +1,14 @@
+/* For soft-wrapped code inside tables */
+.rgh-softwrapped-code {
+	table-layout: fixed;
+}
+.rgh-softwrapped-code tr:not(.js-expandable-line) .blob-num {
+	width: 50px;
+}
+.rgh-softwrapped-code tr:not(.js-expandable-line) .blob-code-inner {
```
bfred-it: Only this seems to be needed (in PR Files). Either drop the rest of the code or add comments to explain it

source/features/softwrap-code.css
```diff
@@ -0,0 +1,14 @@
+/* For soft-wrapped code inside tables */
```
bfred-it:
```suggestion
```

bfred-it:
Can you commit to maintain this feature going forward? For example, in case GitHub changes something that breaks it. We already have way too many features to maintain, so we need to find a way to make this sustainable.

If so, can you add a line to the [CODEOWNERS](https://github.com/sindresorhus/refined-github/blob/master/.github/CODEOWNERS) file?

source/features/softwrap-code.tsx
```diff
@@ -15,7 +15,8 @@ async function init(): Promise<void> {
 	for (const table of tables) {
 		table.classList.add('rgh-softwrapped-code');
 
-		for (const line of select.all('.blob-code-inner', table)) {
+		// `span` required to exclude `td` in PRFiles
+		for (const line of select.all('span.blob-code-inner', table)) {
```
notlmn: Adding `span` wouldn't target code inside suggestions. For suggestions that class exists on the `td` itself.

source/features/softwrap-code.tsx
notlmn: This needed a different kind of workaround, it needed to be activated on PR conversation, commit, PR files, on ajaxed comments, ...

source/features/softwrap-code.tsx
notlmn: Moved forward with `indentation-aware-code-wrapping`, open to changes too!

source/features/softwrap-code.tsx
notlmn: Did my best to explain it, but it still looks a bit bad though!

notlmn:
> Can you commit to maintain this feature going forward? For example, in case GitHub changes something that breaks it. We already have way too many features to maintain, so we need to find a way to make this sustainable.
> 
> If so, can you add a line to [CODEOWNERS](https://github.com/sindresorhus/refined-github/blob/master/.github/CODEOWNERS) [](/sindresorhus/refined-github/blob/HEAD@{2019-05-02T15:16:39Z}/.github/CODEOWNERS)?

I would be glad to, and yes I added my name to the list. Thanks!

source/features/indentation-aware-code-wrapping.tsx
```diff
@@ -0,0 +1,63 @@
...
+		'.file .d-table:not(.rgh-softwrapped-code)', // "Suggested changes" in PRs
+		'.file .js-file-line-container:not(.rgh-softwrapped-code)' // Embedded code blocks
+	].join());
+
+	for (const table of tables) {
+		table.classList.add('rgh-softwrapped-code');
+
+		for (const line of select.all('.blob-code-inner', table)) {
+			// All lines may not have `firstChild`, set to `null`
+			if (!(line.firstChild && line.firstChild.textContent)) {
```
bfred-it:
```suggestion
			if (!line.firstChild || !line.firstChild.textContent) {
```

source/features/indentation-aware-code-wrapping.tsx
```diff
@@ -0,0 +1,63 @@
...
+		'.file .diff-table:not(.rgh-softwrapped-code)', // Split and unified diffs
+		'.file .d-table:not(.rgh-softwrapped-code)', // "Suggested changes" in PRs
+		'.file .js-file-line-container:not(.rgh-softwrapped-code)' // Embedded code blocks
+	].join());
+
+	for (const table of tables) {
+		table.classList.add('rgh-softwrapped-code');
+
+		for (const line of select.all('.blob-code-inner', table)) {
+			// All lines may not have `firstChild`, set to `null`
```
bfred-it:
```suggestion
			// Some matched elements don’t contain code
```

source/libs/on-pr-file-load.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+const fragmentsLoaded = debounce(() => {
+	// Safely run all callbacks
+	handlers.forEach(async cb => cb());
+}, {wait: 200});
+
+function setup(): void {
+	for (const fragment of select.all('include-fragment.diff-progressive-loader')) {
+		// Remove any existing event listeners
+		fragment.removeEventListener('load', fragmentsLoaded);
```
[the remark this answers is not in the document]
notlmn: > `addEventListener` doesn’t add the same handler two, so it doesn’t need to be removed.

Done

------------------

> Also this handler shouldn’t be too complicated, the whole file should just be a default export function with the loop that adds the callback as a listener.

Done

------------------

> Also it probably shouldn’t be part of `features`’ loaders because it’s never used on its own. You can leave plain call in your feature’s `init` function rather than using that new `load`

That piece of code was extracted out as you said, from `raw-file-link`, so that that part of code can be reused (with the intention of updating `raw-file-link` later).

This feature need to be triggered on both new comments and diff loads, but raw-file-link only depends on diff loads, hence the need for this new load mechanism.

notlmn:
Update: https://github.com/lukechilds/github-custom-tab-size/pull/3 has been merged 🎉

bfred-it:
I'll merge this after you verify if my changes introduced any regressions :D

notlmn:
> I'll merge this after you verify if my changes introduced any regressions :D

I was waiting for @sindresorhus' approval before making any more comments 😅.

The changes look good btw. Thanks for that.

sindresorhus:
I'm seeing something really weird on https://github.com/sindresorhus/refined-github/pull/1881/files:

Screen Shot 2019-05-07 at 10 41 12

bfred-it:
> I'm seeing something really weird

I'm not seeing that in Firefox and Chrome, but I did have to fix the code

notlmn:
Sorry it has taken a while for me to get back on this, because of me travelling a lot recently and my semester exams.

https://github.com/sindresorhus/refined-github/pull/1989/commits/a79de46a4eb593b7a6bc51f2aa6dca7c06b06073 should fix all of the issues that we have as of now. Tested it on unified and split diffs, and review suggestions.

I'm pretty sure we've missed this before, but we cannot wrap code inside embedded code blocks as the nodes are placed inside a ShadowDOM, so removed that selector in the last commit. 🤷‍♂️

notlmn:
The only quirk that I can find of is when the git-generated hunk is too large, which we intentionally don't wrap. I'm sure this can be left out as an exception.

image

bfred-it:
I don't get this, can you post a link? Why don't we wrap it

notlmn:
> I don't get this, can you post a link? Why don't we wrap it?

That's one of the review comments from the Parcel link above.

We don't wrap it because of the `'.blob-code-inner:not(.blob-code-hunk)'` selector used. I mean, we could wrap it too, but I don't think that's part of code. Let me know what you think. I'm open to wrap the hunk part too.

bfred-it:
That part should be wrapped too, but I think the JS shouldn't run on it because it's never indented: its first char is always `@`.

bfred-it:
Merge with master, it needs a description field because of #2030

notlmn:
Added the `description` field, but I'd leave out the exact wording to @bfred-it and @sindresorhus. 😉

bfred-it:
There's a jump we can probably avoid:



This is actually two parts:

- the line height changes
- the indentation changes (which is probably more problematic since we skip some short lines and thus the indentation will no longer align)

The first part wouldn't be noticeable if there was no delay. Sadly `on-new-comments` is debounced to 200ms (sadly I don't remember exactly [why](https://github.com/sindresorhus/refined-github/pull/1640/commits/4ce9d9e144e258bddf79a70897d82989156baac1) such a high value) so we have 3 options:

- add a `rgh-softwrapped-code` class on `body` to enable regular wrapping on all elements before the rest of the javascript runs;
- ensure that the height doesn't change;
- drop the 200ms delay, after ensuring that all the features that rely on it work correctly on: fresh comments (they're auto-loaded as people comment on issues), collapsed review comments, collapsed comments on long discussions.

Ideally all 3 changes would be made, but probably only one is needed.

notlmn:
> * ensure that the height doesn't change;

Done.

---------------

> * drop the 200ms delay, after ensuring that all the features that rely on it work correctly on: fresh comments (they're auto-loaded as people comment on issues), collapsed review comments, collapsed comments on long discussions.

I have no idea how changing this would affect other features. Ideally all layout changes should be done under 50ms (just a random value off the top of my head). How about lowering the delay instead of dropping it.

source/features/indented-code-wrapping.css
```diff
@@ -1,3 +1,8 @@
+.rgh-softwrapped-code .blob-code:not(.blob-code-hunk) {
+	line-height: 21px; /* Compensation for block-ifying table cells */
+}
+
+.rgh-code-wrapping-enabled .blob-code-hunk, /* Wrap code eagerly to avoid layout changes later */
```
notlmn: > * add a `rgh-softwrapped-code` class on `body` to enable regular wrapping on all elements before the rest of the javascript runs;

This should do the task, but I'm assuming this might affect fragment-targeted-link-loading (changing layout after page load might move the page over the desired view).

source/features/indented-code-wrapping.css
```diff
@@ -0,0 +1,13 @@
...
+.rgh-softwrapped-code .blob-code-hunk {
+	white-space: pre-wrap;
+}
+
+.rgh-softwrapped-code tr:not(.js-expandable-line) .blob-code-inner:not(.blob-code-hunk) {
+	display: block;
```
bfred-it: Since the height change is caused by this, shouldn’t `line-height` be here?

notlmn: Because that affects `td` inside suggestion blocks, changing their height from `26px` to `29px` if `line-height` is added.

We only need to add the CSS for `td.blob-code` inside diff tables, which already have `line-height` of `20px` (now changed to `21px`).

This is evident from the GIF above, the suggestion box is not affected before this change (or after for that matter).

notlmn:
Is GitHub testing soft-wrapping? I don't remember seeing these CSS classes before.

image

Test PR: https://github.com/sindresorhus/awesome-lint/pull/82

/cc @lukehefson @shayfrendt

bfred-it:
The layout change is smaller now but still there in Chrome and Firefox (macOS)

source/features/indented-code-wrapping.tsx
```diff
@@ -34,10 +34,12 @@ function run(): void {
...
 
-			// Move the whole line where it is supposed to be, then unindent the
-			// start of the line to compensate for indentation, preserving spaces
+			// Code suggestions already had some padding
+			const extraPadding = line.classList.contains('px-2') ? '8px + ' : '';
+
+			// Move the whole line where it is supposed to be, then unindent the start of the line to compensate for indentation, preserving spaces
 			// We might get `--tab-size` from compatible extensions like `github-custom-tab-size`
-			line.style.setProperty('padding-left', `calc((var(--tab-size, 4) * ${tabCount}ch) + ${spaceCount}ch)`, 'important');
+			line.style.setProperty('padding-left', `calc(${extraPadding}(var(--tab-size, 4) * ${tabCount}ch) + ${spaceCount}ch)`, 'important');
```
notlmn: Damn me for ignoring to add this, thought that it would be way much conditionals when we had to handle 3 table styles, forgot after we realized we were not handling embedded blocks anymore.

❤ this change.

bfred-it:
Last note: **mixed indentation will break** even after the suggested fix a few weeks ago.

If `tab-size: 4`, `   \t` should not be `3 + 4` characters long but still 4 characters.

Random indentation however should worry us, it _should_ look broken. I think the current implementation still works correctly for "accettable" mixed indentation, like:

```css
 -> -webkit-transform: none;
 -> ••••••••transform: none;
```

bfred-it:
Thanks @notlmn! I think will make the code more readable and is more performant than what I would have come up with 🏎

Taluu:
Not sure if it's just that I haven't properly looked around, but is there any possibility to disable the soft-wrapping introduced here in unified view ? It's annoying the hell out of me (I can understand though that some may prefer that).

So introducing a button or something to toggle on / off the soft-wrapping.

Can make a new issue if needed.

Thanks.

bfred-it:
All the JavaScript features can already be disabled in the options. This one is called `indented-code-wrapping`

Taluu:
Awesome, didn't see the field in the extension's config.

Is there a list somewhere of all the extensions names (Just being curious) ? :}

Thanks.

bfred-it:
In the options itself there’s a link to the list

bfred-it:
Can you add this to the readme? It needs simple before/after screenshot (copy-pasted arrows/test from https://user-images.githubusercontent.com/1402241/34438653-f66535a4-ecda-11e7-9406-2e1258050cfa.png)

PR #1994, opened 2019-04-28
Hide comment box upload message

### LINKED ISSUE #1992: Upload message showing in comment box
Not sure if it's just me, but the upload message is now showing in the comment box rather than being hidden like one would expect here:

```css
.rgh-has-upload-field .is-default .drag-and-drop {
    display: none;
}
```

Screen_Shot_2019-04-27_at_11_32_33_PM

It looks to me that the rule is being ignored (tested on both Chromium and Firefox)

Screen Shot 2019-04-27 at 11 06 12 PM

Adding `important!` does solve the issue, but wasn't sure if that was the preferred solution.. But if it is, I can quickly make the change.

doms:
## What

Closes #1992 

## Notes

@bfred-it mentioned [here](https://github.com/sindresorhus/refined-github/issues/1992#issuecomment-487345168) that

> the **!important** is required now because they have a utility class on it

### DIFF
--- a/source/features/upload-button.css
+++ b/source/features/upload-button.css
@@ -1,6 +1,6 @@
 /* Remove upload message on comment box */
 .rgh-has-upload-field .is-default .drag-and-drop {
-	display: none;
+	display: none !important;
 }
 
 .rgh-has-upload-field .is-default textarea {


doms:
Oops, just realized that I merged master into my branch, and that some of the css were moved into their own files. I'll revert and fix it 😄

PR #1996, opened 2019-04-29
Fix `show-names` features when bots are on the page

### LINKED ISSUE #1978: `show-names` feature  broken
```
❌ show-names
content.js:1880 Error: GraphQL:
Parse error on "[" (LBRACKET) at [1, 61]
    at content.js:1992
```

https://github.com/highlightjs/highlight.js/pull/2023

yakov116:
show-names fails when the link is to the marketplace.

~~fixes []() #1978~~

### DIFF
--- a/source/features/show-names.tsx
+++ b/source/features/show-names.tsx
@@ -7,7 +7,7 @@ import {getUsername} from '../libs/utils';
 
 async function init(): Promise<false | void> {
 	// `a` selector needed to skip commits by non-GitHub users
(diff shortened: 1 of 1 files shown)


bfred-it:
The broken user links to `https://github.com/greenkeeper%5Bbot%5D` so this does not fix that issue. However I noticed that the bots now point to `/marketplace/*` instead of `/apps/*`. I'll leave them both for now, but eventually we should skip commit authors altogether in `show-names`

PR #1997, opened 2019-04-29
Avoid duplicate releases

bfred-it:
Thanks to Google everyone can be a git/shell wizard 🧙‍♂️

Fixes #1995 

Documentation: 
- https://docs.travis-ci.com/user/deployment/#conditional-releases-with-on
- https://stackoverflow.com/a/33733020/288906
- https://stackoverflow.com/a/3826705/288906

Tested with:

```sh
TRAVIS_EVENT_TYPE=cron
if [[ ($TRAVIS_EVENT_TYPE = cron) && ($(git tag -l --points-at HEAD) = "") ]]
then 
	echo :ship-it:
fi
```

Further simplifications welcome

PR #1998, opened 2019-04-29
Add `tag-changelog-link` feature

### LINKED ISSUE #966: Link to "What changed since the previous release?"

HardikModha:
Should fix #966.

I tried to use the API but I couldn't find a way to fetch details of next and previous tags and fetching all tags at once wasn't feasible. (See discussion #966)

I've taken the following approach. 

1. First, I've extracted all tags, commit ids of those tags.

2. Now, iterating over each tag, I see if the next tag has the different commit id. (Given the fact that we can have multiple tags on the same commit id).  

3. If the next tag has different commit id, then I further check whether the tags are namespaced or not. If the namespace is the same then pick that tag for comparison.

4. If the next tag has same commit id then that tag is skipped and we check remaining tags.

5. For the last tag, I fetch the next page using `fetchDom` and perform the same steps as mentioned above. If a tag to compare is not found on the next page then we skip the tag. 

**Tested it on the following (and many other) repos.**

1. https://github.com/parcel-bundler/parcel/releases (It has multiple, namespaced and normal semver style tags)
2. https://github.com/tensorflow/tensorflow/releases (With Alpha, Beta kind of tags)
3. https://github.com/facebook/react/releases

**The feature works on the following pages:**

1. releases listing: e.g. https://github.com/parcel-bundler/parcel/releases
2. tags listing: e.g. https://github.com/parcel-bundler/parcel/tags
3. Single tag pages: https://github.com/parcel-bundler/parcel/releases/tag/v1.10.3
 
**Limitations:**

1. ~Namespaced tags are only checked within the single page to find it's previous tag~. **Update**: Now we check on the next page, too. 

2. ~For the last tag on the page, we only check on the next page to find the previous tag.~ **Update**: For all the tags on any page, we also check the next page for finding its previous tag. 

3. The tags are not semantically compared. But we can easily implement it.

**PS**: Please suggest a better way (if you have) to solve this issue. I would be more than happy to implement it. :)


HardikModha:
@bfred-it @sindresorhus Can you please take a look? Thanks!

source/content.ts
```diff
@@ -113,6 +114,7 @@ import './features/hide-tips.css';
 import './features/hide-readme-header.css';
 import './features/hide-obvious-tooltips.css';
 import './features/clean-discussions.css';
+import './features/what-changed-since-previous-release.css';
```
bfred-it: This should be inside the feature itself

source/features/what-changed-since-previous-release.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+	const {ownerName, repoName} = getOwnerAndRepo();
+
+	// To extract the tag "v16.8.6" from links like "/facebook/react/releases/tag/v16.8.6"
+	const tagRegExp = new RegExp(`/${ownerName}/${repoName}/releases/tag/(.*)`);
+
+	// To extract the commit hash from links like "/facebook/react/commit/92a1d8feac32d03ab5ea6ac13ae4941f6ae93b54"
+	const commitRegExp = new RegExp(`/${ownerName}/${repoName}/commit/([0-9a-f]{5,40})`);
+
+	const allTagsAnchor = [...select.all<HTMLAnchorElement>(getTagSelector())];
```
bfred-it: `select.all` already returns an array

source/features/what-changed-since-previous-release.tsx
```diff
@@ -0,0 +1,175 @@
...
+		return false;
+	}
+
+	const {ownerName, repoName} = getOwnerAndRepo();
+
+	// To extract the tag "v16.8.6" from links like "/facebook/react/releases/tag/v16.8.6"
+	const tagRegExp = new RegExp(`/${ownerName}/${repoName}/releases/tag/(.*)`);
+
+	// To extract the commit hash from links like "/facebook/react/commit/92a1d8feac32d03ab5ea6ac13ae4941f6ae93b54"
+	const commitRegExp = new RegExp(`/${ownerName}/${repoName}/commit/([0-9a-f]{5,40})`);
```
bfred-it: These links never point to a different repo, so you just need

```suggestion
	const tagRegExp = /releases\/tag\/(.*)$/;
	const commitRegExp = /commit\/([0-9a-f]{40})$/;
```

source/features/what-changed-since-previous-release.tsx
HardikModha: 🤦‍♂ There was some Typescript error and it wasn't going away but now it's gone. 💭 Removed the use of spread.

source/content.ts
HardikModha: Is it needed? The styles are getting applied irrespective of it. Also, none of the current features has these CSS files imported.

HardikModha:
> * `getCommitIdSelector` and `getTagSelector` can be skipped if you just use `commitAnchorSelector` and `tagAnchorSelector` respectively? This lets you skip a lot of code and even `isReleasesListPage` and `isTagsListPage` themselves

Done. Simplified the commit id and tag selectors. Also, removed the checks for `isReleaseListPage` and `isTagsListPage` by using the common classname as selector which was present on both the pages.

> * the code after `allTagsAnchor` and `nextPageAllTagsAnchor` is nearly identical and can probably be merged.

Now, Appending the next page tags into the same array and iterating only once. Check for the last tag is also removed.

> Also always use `for-of` and not a regular `for`

Here, I need an index of the item so that I can start looking for the tags to match just after that index. You can take a look from Line no 36 and Line no 62. 

> Position: I was expecting this link to be on the `commit | zip | tar.gz` line and something more clear than `...` (I didn't even see the link at first, that `...` is generally used to expand the content)

Initially, I had chosen the same. The same line as commit | zip | tar.gz and for the releases just below the commit hash. (As displayed in the screenshot)

1ec2a437-cc8e-45eb-97f7-4d15ac433700

But there was some miscommunication due to https://github.com/sindresorhus/refined-github/issues/966#issuecomment-487345368 and I put the link on `...` 😓.

HardikModha:
@bfred-it Please take a look at new changes and suggest me if anything still needs to be changed. :)

source/content.ts
bfred-it: Yes it’s needed, the JS features should “load” their own CSS. Those in content.ts are CSS-only features

Look better, there are quite a few feature that load their own CSS

source/features/what-changed-since-previous-release.tsx
```diff
@@ -0,0 +1,127 @@
...
+	if (nextPage) {
+		allTags.push(...extractValuesFromPathname(select.all<HTMLAnchorElement>(tagSelector, nextPage), tagRegExp));
+		allCommitIds.push(...extractValuesFromPathname(select.all<HTMLAnchorElement>(commitIdSelector, nextPage), commitRegExp));
+	}
+
+	for (let index = 0; index < allCommitIdsAnchor.length; index++) {
+		const previousTagIndex = getPreviousTagIndex(index + 1, allCommitIds[index], allCommitIds, allTags[index], allTags);
+
+		if (previousTagIndex !== -1) {
+			allCommitIdsAnchor[index].parentElement!.parentElement!.append(getTagComparisonLink(allTags[previousTagIndex], allTags[index]));
```
bfred-it: Use `.closest` instead of multiple `parentElement`

source/features/what-changed-since-previous-release.tsx
```diff
@@ -0,0 +1,127 @@
...
+};
+
+const isSameTagNameSpace = (tag1: string, tag2: string): boolean => {
+	const namespaceRegex = /(.*)@(.)*/;
+
+	const [, tagOneNameSpace = ''] = tag1.match(namespaceRegex) || [];
+	const [, tagTwoNameSpace = ''] = tag2.match(namespaceRegex) || [];
+
+	return tagOneNameSpace === tagTwoNameSpace;
+};
```
bfred-it: If I’m understanding this correctly, hierarchy and namespace are actually part of the same thing. What we’re looking for here is SOMETHING@version. It does not matter what the shape of SOMETHING is, you just need to make sure that SOMETHING is the same for both links, if it exists. Example:

```js
function doesNamespaceMatch() {
  const [namespace1] = tag1.split(/@[^@]+$|^/); // I think this regex will always keep the version at index 1 of the array
  const [namespace2] = tag2.split(/@[^@]+$|^/);
  return namespace1 === namespace2;
}
```

source/features/what-changed-since-previous-release.tsx
```diff
@@ -0,0 +1,127 @@
...
+	const nextPageLink = select<HTMLAnchorElement>('.pagination a:last-child');
+	const nextPage = nextPageLink ? await fetchDom(nextPageLink.href) : await getNextPageForSinglePageTag(allTags[0]);
+
+	if (nextPage) {
+		allTags.push(...extractValuesFromPathname(select.all<HTMLAnchorElement>(tagSelector, nextPage), tagRegExp));
+		allCommitIds.push(...extractValuesFromPathname(select.all<HTMLAnchorElement>(commitIdSelector, nextPage), commitRegExp));
+	}
+
+	for (let index = 0; index < allCommitIdsAnchor.length; index++) {
+		const previousTagIndex = getPreviousTagIndex(index + 1, allCommitIds[index], allCommitIds, allTags[index], allTags);
```
bfred-it: You’re passing the same information twice. 
```suggestion
		const previousTagIndex = getPreviousTagIndex(index + 1, allCommitIds, allTags);
```

You’re already passing `index` so create the missing variables in the function. This avoid having a 5-parameter function (too many parameters)

source/features/what-changed-since-previous-release.tsx
HardikModha: Much better. 🎉

bfred-it: I'm working on the regex, it doesn't what I had hoped

bfred-it: Ok, now it sort of does. It relies on the assumption that there's no such thing as `v1.2@v1.2` tag, so even if in a namespace-less tag the version is at the index `0`, it won't affect the comparison:

HardikModha: Hey @bfred-it, How did you know I was doing the same? 🤔 Simplifying the regex. I just changed regex from `/@[^@]+$|^/` to `/@[^@]+$/`. You spyin' on me? 😂

bfred-it: I had added `|^` thinking that it would split `v1.2` into `['', 'v1.2`]`, but that only broke the `@user/module@version`:

Screenshot 2019-05-01 at 14 26 35

source/features/what-changed-since-previous-release.css
```diff
@@ -0,0 +1,14 @@
...
+	margin-top: 4px;
+	margin-right: 8px;
+}
+
+.release > div:first-child > ul > li.rgh-what-changed {
+	margin-bottom: 4px;
+}
+
+.rgh-what-changed {
+	cursor: pointer;
```
bfred-it:
```suggestion
```

It's already a link

source/features/what-changed-since-previous-release.css
```diff
@@ -0,0 +1,14 @@
+.commit > ul > li.rgh-what-changed {
+	display: inline-block;
+	margin-top: 4px;
+	margin-right: 8px;
+}
+
+.release > div:first-child > ul > li.rgh-what-changed {
+	margin-bottom: 4px;
```
bfred-it: See if you can replace these styles with GitHub's own margin classes, like `mb-1` and `d-inline-block`. Maybe you can apply them without any conditions (i.e. just use `className="mb-1 mr-1 mt-1"` on all)

source/features/what-changed-since-previous-release.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+		if (prevTagIndex === -1) {
+			prevTagIndex = i;
+		}
+	}
+
+	return prevTagIndex;
+};
+
+const doesNamespaceMatch = (tag1: string, tag2: string): boolean => tag1.split(/@[^@]+$/)[0] === tag2.split(/@[^@]+$/)[0];
```
bfred-it:
```suggestion
const doesNamespaceMatch = (tag1: string, tag2: string): boolean =>
	tag1.split(/@[^@]+$/)[0] === tag2.split(/@[^@]+$/)[0];
```

source/features/what-changed-since-previous-release.tsx
```diff
@@ -0,0 +1,103 @@
...
+			return i;
+		}
+
+		if (prevTagIndex === -1) {
+			prevTagIndex = i;
+		}
+	}
+
+	return prevTagIndex;
+};
```
bfred-it: I think this defaults to the last tag on the page (or the next page), it never returns `-1`

If it cannot find a matching namespace, it should just use the next non-same-commit tag.

source/features/what-changed-since-previous-release.tsx
```diff
@@ -0,0 +1,103 @@
...
+	if (nextPage) {
+		allTags.push(...extractValuesFromPathname(select.all<HTMLAnchorElement>(tagSelector, nextPage), tagRegExp));
+		allCommitIds.push(...extractValuesFromPathname(select.all<HTMLAnchorElement>(commitIdSelector, nextPage), commitRegExp));
+	}
+
+	for (let index = 0; index < allCommitIdsAnchor.length; index++) {
+		const previousTagIndex = getPreviousTagIndex(index, allCommitIds, allTags);
+
+		if (previousTagIndex !== -1) {
+			allCommitIdsAnchor[index].closest('ul')!.append(getTagComparisonLink(allTags[previousTagIndex], allTags[index]));
```
bfred-it: You can inline `getTagComparisonLink` here (i.e. just create JSX inside `.append()`)

source/features/what-changed-since-previous-release.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+const doesNamespaceMatch = (tag1: string, tag2: string): boolean => tag1.split(/@[^@]+$/)[0] === tag2.split(/@[^@]+$/)[0];
+
+const getTagComparisonLink = (prevTag: string, nextTag: string): HTMLElement => {
+	const {ownerName, repoName} = getOwnerAndRepo();
+
+	return (
+		<li className="rgh-what-changed">
+			<a className="muted-link text-mono" href={`/${ownerName}/${repoName}/compare/${prevTag}...${nextTag}`}>
+				<div className="d-inline-block mr-1">{diff()}</div>compare
```
bfred-it:
```suggestion
				{diff()} Changelog
```

Other links on that line just rely on the space between the icon and the text

source/features/what-changed-since-previous-release.tsx
```diff
@@ -0,0 +1,111 @@
...
+			</a>
+		</li>
+	);
+};
+
+features.add({
+	id: 'what-changed-since-previous-release',
```
bfred-it:
```suggestion
	id: 'tag-changelog-link',
```

I think this is a clear name for this feature. Also the file should be renamed

source/features/what-changed-since-previous-release.css
HardikModha: Much better. Everything just reduced to `d-inline-block mb-1 mt-1 f6`. 
Got rid of the `css` file.

source/features/what-changed-since-previous-release.tsx
HardikModha: It can return `-1` when all the tags on the page are having the same commit id.

```
if (allCommitIds[i] === allCommitIds[index]) {
	continue;
}
```

source/features/tag-changelog-link.tsx
```diff
@@ -0,0 +1,111 @@
...
+};
+
+const doesNamespaceMatch = (tag1: string, tag2: string): boolean => tag1.split(/@[^@]+$/)[0] === tag2.split(/@[^@]+$/)[0];
+
+const getTagComparisonLink = (prevTag: string, nextTag: string): HTMLElement => {
+	const {ownerName, repoName} = getOwnerAndRepo();
+
+	return (
+		<li className="rgh-tag-changelog-link">
+			<a className="muted-link text-mono" href={`/${ownerName}/${repoName}/compare/${prevTag}...${nextTag}`}>
```
bfred-it:
```suggestion
			<a className="muted-link text-mono tooltipped tooltipped-n" aria-label={'See changes since ' + prevTag} href={`/${ownerName}/${repoName}/compare/${prevTag}...${nextTag}`}>
```

bfred-it:
The current version is buggy, it's skipping version 29 (look at the status bar):



https://github.com/sindresorhus/refined-github/releases

source/features/what-changed-since-previous-release.tsx
HardikModha: Can we add a lint rule to avoid this in the future? https://eslint.org/docs/rules/max-len

HardikModha:
> The current version is buggy, it's skipping version 29 (look at the status bar):
> 
> 
> 
> https://github.com/sindresorhus/refined-github/releases

Technically it isn't. The tag is on the same commit id so it looks for the next tag on different commit id.

bfred-it:
> Technically it isn't.

D'oh. True. #1997 coming soon

source/features/what-changed-since-previous-release.css
HardikModha: CSS file removed.

source/features/what-changed-since-previous-release.css
bfred-it: Don't forget to push your changes, I still see the file :)

source/features/tag-changelog-link.tsx
HardikModha: Love it, man! Thanks! Great suggestion!

bfred-it:
I'm not seeing the link on https://github.com/sindresorhus/refined-github/releases/tag/19.4.30.1910

HardikModha:
@bfred-it Seems like you missed screenshots from https://github.com/sindresorhus/refined-github/pull/1998#issuecomment-488116969.

For releases, the link is displayed just below the commit id on the left side.

changelog-link

bfred-it:
Seems like it's not working

Screenshot 2019-05-01 at 15 53 33

HardikModha:
@bfred-it Just confirming, You are on the latest changes. Right? Also, Can you check the console if there is anything? It seems to work fine on my end. Tried for multiple tags.

bfred-it:
Ah yeah, this is due to Firefox requiring `location.origin` on `fetch`

source/features/tag-changelog-link.tsx
```diff
@@ -0,0 +1,112 @@
...
+};
+
+const getNextPageForSinglePageTag = async (tag: string): Promise<void | DocumentFragment> => {
+	if (!isSingleTagPage()) {
+		return;
+	}
+
+	const {ownerName, repoName} = getOwnerAndRepo();
+
+	return fetchDom(`/${ownerName}/${repoName}/releases?after=${tag}`);
```
bfred-it:
```suggestion
	return fetchDom(`${location.origin}/${ownerName}/${repoName}/releases?after=${tag}`);
```

HardikModha:
> I'm not seeing the link on [`19.4.30.1910` (release)](https://github.com/sindresorhus/refined-github/releases/tag/19.4.30.1910) and [bfred-it/insert-text-textarea@`v1.0.3` (release)](https://github.com/bfred-it/insert-text-textarea/releases/tag/v1.0.3)

It's also working on https://github.com/bfred-it/insert-text-textarea/releases/tag/v1.0.3. Are you on Chrome right?

source/features/tag-changelog-link.tsx
```diff
@@ -0,0 +1,112 @@
...
+					</a>
+				</li>
+			);
+		}
+	}
+}
+
+const extractValuesFromPathname = (anchors: HTMLAnchorElement[], regexp: RegExp): string[] => {
+	return anchors.map((anchor: HTMLAnchorElement): string => {
+		const [, value] = anchor.pathname.match(regexp)!;
```
bfred-it:
```suggestion
		return anchor.textContent!.trim();
```

You might have to adjust the selector to select these links:

bfred-it:
You can use this:

HardikModha:
Seems like the changes you made has introduced a bug.

Screen Shot 2019-05-02 at 11 58 20 PM

It should compare with `@parcel/integration-tests@1.12.2` not `@parcel/integration-tests@1.12.1`.

source/features/tag-changelog-link.tsx
```diff
@@ -0,0 +1,93 @@
...
+		return false;
+	}
+
+	const documents = [document, await getNextPage()] as any; // TODO: fix select-dom types to accept mixed arrays
+	// These selectors need to work on:
+	// https://github.com/facebook/react/tags (tags list)
+	// https://github.com/facebook/react/releases (releases list)
+	// https://github.com/parcel-bundler/parcel/releases (releases list without release notes)
+	const tagElements = select.all('.f1 [href*="/releases/tag"], .commit-title [href*="/releases/tag"]', documents);
+	const commitElements = select.all('.muted-link[href*="/commit/"]', documents);
```
HardikModha: This selector is not reliable. It is selecting multiple links on the single tag release page causing the wrong tag to be displayed as the previous tag. We'll need to change it.

HardikModha: The same issue is also occurring on the release page where the tag is an actual release with the release notes. 

Apart from that, The feature has broken on such releases page due to index out of bounds. 

Screen Shot 2019-05-03 at 12 18 45 AM

sindresorhus:
Can you add this feature file to https://github.com/sindresorhus/refined-github/blob/master/.github/CODEOWNERS ?

sindresorhus:
It doesn't work for releases with a custom title. It should not use the release title, but rather the Git tag connected to the release. Example: https://github.com/avajs/ava/releases/tag/v2.0.0-beta.1 (Click "Changelog").

source/features/tag-changelog-link.tsx
```diff
@@ -0,0 +1,93 @@
+/*
+This feature adds a compare link on each releases/tags/single tag page so that you can see
+what has changed since the previous release. If the tags are namespaced then it tries to
+get the previous release of the same namespaced tag.
```
sindresorhus: Don't hard-wrap.

source/features/tag-changelog-link.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+const getPreviousTag = (index: number, commits: string[], tags: string[]): string | false => {
+	let previousTag: string | false = false;
+	for (let i = index + 1; i < commits.length; i++) {
+		if (commits[i] === commits[index]) {
+			continue;
+		}
+
+		// Ensure that they have the same namespace. e.g. `parcel@1.2.4` and `parcel@1.2.3`
+		if (tags[i].split(/@[^@]+$/)[0] === tags[index].split(/@[^@]+$/)[0]) {
```
sindresorhus: The regex should be moved to a variable with a descriptive name and reused for both of these.

bfred-it:
> It doesn't work for releases with a custom title. It should not use the release title, but rather the Git tag connected to the release. Example: [avajs/ava@`v2.0.0-beta.1` (release)](https://github.com/avajs/ava/releases/tag/v2.0.0-beta.1) (Click "Changelog").

Indeed, there are so many variations to the dom I ended up forgetting about this one.

One more thing missing is a link in the second meta box (it's duplicate in the dom for responsiveness). For this reason (and to avoid this awkward "double-select that has to match the same number of items"), I'd suggest another way to do this, something like:


```js
const tags = select.all('each tag’s main element').map(element => ({
	element,
	tag: element.querySelector('...'),
	commit: element.querySelector('...')
}));
```

This way there will be **one** array of tags/releases with content already parsed. Later, when adding the buttons, you can run something like:

```js
if (previousTag !== false) {
	const containers = tag.element.querySelectorAll('all the places that need the link');
	for (const container of containers) {
		container.append(
			...JSX...
		);
	}
}
```

HardikModha:
> For this reason (and to avoid this awkward "double-select that has to match the same number of items"), I'd suggest another way to do this, something like:

We are already avoiding the duplicate link right now by reverting back to the older selector. https://github.com/sindresorhus/refined-github/pull/1998/commits/67822230c90af73e2d4459d680028d53b82e4d05
Other than that, Do we also need to show the "Changelog" link for the smaller screens?

HardikModha:
Also, As Sindre pointed out a case, we'll need to revert back to the old regex-based approach to extract tag. 
https://github.com/sindresorhus/refined-github/blob/e38619ab55228937b8a5c42d21c965862eac68e7/source/features/what-changed-since-previous-release.tsx#L15

bfred-it:
> Do we also need to show the "Changelog" link for the smaller screens?

It would be ideal, also because it's not "smaller screens" but "smaller windows", e.g. side-by-side

> Also, As Sindre pointed out a case, we'll need to revert back to the old regex-based approach to extract tag.

Why? The tag is available in the dom, we just need to update the way it parses the dom as suggested. 

As we've seen, any changes to the selector could cause subtle errors that make it mismatch commits and tags, so fetching by "tag container" first would be preferred in every case.

HardikModha:
I just thought extracting from the link would be much safer. The tag name can be customized so.

e.g. If on the parcel release page they use the custom tag name, we don't have another way to extract actual tag name from the DOM. We'll have to use the regex to extract the name from the link. 

Screen Shot 2019-05-03 at 11 40 33 PM

For the releases with release notes, we can extract it from the container present on the left side. But then why follow two separate approaches? We can just stick to the regex.

bfred-it:
Alright, regex it is then :)

https://github.com/sindresorhus/refined-github/pull/1998#issuecomment-488970348 still applies though

HardikModha:
You mean displaying it on the smaller window or the approach that you suggested? or both? 😅

HardikModha:
Completely unrelated to the current topic, https://github.com/sindresorhus/refined-github/pull/1998#issuecomment-489183075 which I just made earlier (In the current thread), I cannot see the content when the scrollbar is present. There should be some minimum height enforced. What do you think?

bfred-it:
> You mean displaying it on the smaller window or the approach that you suggested? or both? 😅

The approach I suggested will make it appear on the smaller window

> Completely unrelated to the current topic, [#1998 (comment)](https://github.com/sindresorhus/refined-github/pull/1998#issuecomment-489183075) which I just made earlier (In the current thread), I cannot see the content when the scrollbar is present. There should be some minimum height enforced. What do you think?

Already a fix by @notlmn coming in #1989 😍

HardikModha:
So I followed the kind of same approach that @bfred-it suggested in (https://github.com/sindresorhus/refined-github/pull/1998/#issuecomment-488970348) with a little difference. I'm not maintaining an object of `element`, `tag` and `commit` instead it's just an array of subparts of the document which contains these tags. Then I'm iterating over each subpart to append the changelog link at all the places.

bfred-it:
> I'm not maintaining an object of `element`, `tag` and `commit`

Then we still have the same problem of 3 possibly-mismatching selectors. Please use the suggested solution.

> As we've seen, any changes to the selector could cause subtle errors that make it mismatch commits and tags, so fetching by "tag container" first would be preferred in every case.

HardikModha:
I'm already fetching the tag container first. All the searches for tag/commit hash are being performed on these containers. Can you please explain in which cases it will fail?

bfred-it:
It doesn't fail _now_ but it can **subtly** fail in the future like it did in https://github.com/sindresorhus/refined-github/pull/1998#issuecomment-488781643

I just don't like reusing the same index on now-3 separate arrays generated with 3 long selectors and then stitched together. In my suggested solution there's one source of truth that we can easily inspect in case it fails

HardikModha:
Please correct me If I am wrong, the issue https://github.com/sindresorhus/refined-github/pull/1998#issuecomment-488781643 was occurring because of the selector being too obvious (just `.muted-link[href*="/commit/"])` which resulted in false matches. That's why I have kept the selectors little more specific to avoid such false matches. 

Now regarding your last comment, I'm failing to understand how the solution I've implemented is different from what you are suggesting in the https://github.com/sindresorhus/refined-github/pull/1998#issuecomment-488970348. 
It, in fact, is doing what you are suggesting, just little differently. Again, correct me if I am wrong. 

You are suggesting to first select "all tags main elements" and then iterate over them to extract the tag and commit. I'm doing the same thing.

In my implementation, "each tag's main element" is `.release`, `.release-main-section:not(.commit)`, and `.Box-row .commit`. These three are the unique selectors which select the main container. 

1. `.release` => Releases with releases notes
2. `.release-main-section:not(.commit)` => Releases without release notes
3. `.Box-row .commit` => Tag on the tags listing. The `.Box-row` is important because the same `.commit` selector also exists as child in 1 & 2.

```
// Selecting all the main tag elements
const tagContainers = select.all('.release, .Box-row .commit, .release-main-section:not(.commit)', documents);
```

Now we have three different containers so in these containers tag and commit hash exists at different locations. So we need different selectors to extract the tag and commit hash. 

But as we know "releases without release notes" and "tags on listing page" have the same structure. We can use the same selector for them.

**Selectors for tag:**

1. `'h4.commit-title > a[href*="/releases/tag/"]` => Extracts the tag name from the title on tags listing and releases without release notes. 

2. `div.release-header .f1 > a[href*="/releases/tag/"]` => Extracts the tag name from the title from releases with release notes.

We just cannot use `a[href*="/releases/tag"]` as it _might_ return false matches.

**Selector for commit hash**

1. `.commit ul a[href*="/commit/"].muted-link` => In all three containers commit hash can be exracted using this single selector. 

```
// Extracting tag names from all (Passing tagContainers)
const tagElements = select.all<HTMLAnchorElement>('h4.commit-title > a[href*="/releases/tag/"], div.release-header .f1 > a[href*="/releases/tag/"]', tagContainers);

// Extracting commit hash from all (Passing tagContainers)
const commitElements = select.all<HTMLAnchorElement>('.commit ul a[href*="/commit/"].muted-link', tagContainers);
```

https://github.com/bfred-it/select-dom/blob/master/index.ts#L81 (Applies the passed selector in each individual container). 

Which is functionally same as, 

```
select.all('each tag’s main element').map(element => ({
	element,
	// These selectors can probably be very simple since they're already in the tag
	// and because they only select the first occurrence of the tag/commit link
	tag: element.querySelector('...'),
	commit: element.querySelector('...')
}));
```

The only difference I can see is that in my implementation we are doing the query for "tag" and "commit" separately.

And now at the last, we select all the places where we need the "Changelog link".

```
select.all('.commit > ul.f6, .commit > .release-header > ul, div:first-child > ul', tagContainer);
```

Don't get me wrong. I can easily convert my implementation to follow the _exact_ steps that you are suggesting but I just want to understand what am I missing here?

HardikModha:
Using your suggested approach as for edge case  (https://github.com/sindresorhus/refined-github/pull/1998/commits/5c0d57f11aabffde2c4b611dc4205538f5cfba18) my approach silently failed but your approach threw an error which I really liked. Also in your suggested approach, we only need to iterate only once to extract tag and commit details. (Removed the big comment which now I think is unnecessary.)

Let me know if any more changes are needed. Thanks :)

source/features/tag-changelog-link.tsx
```diff
@@ -0,0 +1,114 @@
...
+	// If tag is `@parcel/integration-tests@1.12.2` then namespace is `@parcel/integration-tests`
+	const namespaceRegex = /@[^@]+$/;
+
+	for (let i = index + 1; i < allTags.length; i++) {
+		if (allTags[i].commit === allTags[index].commit) {
+			continue;
+		}
+
+		// Ensure that they have the same namespace. e.g. `parcel@1.2.4` and `parcel@1.2.3`
+		if (allTags[i].tag.split(namespaceRegex)[0] === allTags[index].tag.split(namespaceRegex)[0]) {
```
bfred-it: Since `namespaceRegex` is a separate variable again, we might as well turn it into a global function, e.g.

```js
const getNameSpace = tag => tag.split(/@[^@]+$/)[0];
```

```suggestion
		if (getNameSpace(allTags[i].tag) === getNameSpace(allTags[index].tag)) {
```

source/features/tag-changelog-link.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+	// These selectors need to work on:
+	// https://github.com/facebook/react/tags (tags list)
+	// https://github.com/facebook/react/releases (releases list)
+	// https://github.com/parcel-bundler/parcel/releases (releases list without release notes)
+	const tagSelector = 'h4.commit-title > a[href*="/releases/tag/"], div.release-header .f1 > a[href*="/releases/tag/"]';
+
+	const allTags: TagDetails[] = select.all(tagContainerSelector, documents).map(element => ({
+		element,
+		tag: decodeURIComponent(select<HTMLAnchorElement>(tagSelector, element)!.pathname.match(tagRegex)![1]),
```
bfred-it: I think this isn't needed, also because the tag will end up back into a URL anyway
```suggestion
		tag: select<HTMLAnchorElement>(tagSelector, element)!.pathname.match(tagRegex)![1],
```

source/features/tag-changelog-link.tsx
```diff
@@ -0,0 +1,114 @@
...
+		commit: select('.commit ul a[href*="/commit/"].muted-link', element)!.textContent!.trim()
```
bfred-it: Can this be simplified now that it's limited to the `tagContainer`? `select` will only pick the only link anyway, we don't have a problem like before.

```suggestion
		commit: select('[href*="/commit/"]', element)!.textContent!.trim()
```

source/features/tag-changelog-link.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+	// not(.js-timeline-tags-expander) is needed as there can be some collapsed tags
+	// See https://github.com/facebook/react/releases?after=v16.7.0 for an example
+	const tagContainerSelector = '.release, .Box-row .commit, .release-entry .release-main-section:not(.commit):not(.js-timeline-tags-expander)';
+
+	// These selectors need to work on:
+	// https://github.com/facebook/react/tags (tags list)
+	// https://github.com/facebook/react/releases (releases list)
+	// https://github.com/parcel-bundler/parcel/releases (releases list without release notes)
+	const tagSelector = 'h4.commit-title > a[href*="/releases/tag/"], div.release-header .f1 > a[href*="/releases/tag/"]';
```
bfred-it: Can this be simplified now that it's limited to the `tagContainer`? `select` will only pick the only link anyway, we don't have a problem like before.
```suggestion
	const tagSelector = '[href*="/releases/tag/"]';
```

HardikModha: Yes, you are right the container is limited to `tagContainer` but this `tagContainer` can be either "release with notes" or "release without notes". So with the release with notes false matches can occur in the description.

source/features/tag-changelog-link.tsx
HardikModha: Same as https://github.com/sindresorhus/refined-github/pull/1998#discussion_r281033111

source/features/tag-changelog-link.tsx
bfred-it: The description comes after the header, and the `/releases/tag/` link is in the header, so it's matched first

source/features/tag-changelog-link.tsx
HardikModha: We are also showing the tooltip so if we don't `decodeURIComponent` it will display tooltip text in the encoded form.

bfred-it: Indeed. Then maybe we need to move `decodeURIComponent` to the tooltip because we need the tag to be urlencoded in the URL

source/features/tag-changelog-link.tsx
bfred-it: Yeah, this is what I was referring to. One long `tagContainer` selector and the rest can be more simple

HardikModha:
> Thank you and sorry for all the reviews

Haha...No need for sorry. I learned a lot through your and Sindre's hardcore reviews. Most of the time I (or any new person) is focused on making something to work and we forget some scenarios or make some silly mistakes. That's where you guys come into the picture. Just making something to work is not enough. I really appreciate your and Sindre's attitude towards perfectness. :)

source/features/tag-changelog-link.tsx
```diff
@@ -0,0 +1,113 @@
...
+If the tags are namespaced then it tries to get the previous release of the same namespaced tag.
+
+See it in action at: https://github.com/parcel-bundler/parcel/releases
+*/
+import './tag-changelog-link.css';
+import React from 'dom-chef';
+import select from 'select-dom';
+import features from '../libs/features';
+import fetchDom from '../libs/fetch-dom';
+import {diff} from '../libs/icons';
```
bfred-it:
```suggestion
import * as diff from '../libs/icons';
```

I think it's clearer to use this as `icons.diff()` later later than `diff()`

source/features/tag-changelog-link.css
```diff
@@ -0,0 +1,3 @@
+.release div.release-header > ul > li {
+	flex-grow: 0 !important;
```
bfred-it: I understand that this is to avoid this issue:




But this isn't right:

Screenshot 2019-05-09 at 15 50 10


The solution might be a bit ugly, but I think you should:

1. add the link after that `flex-auto`
2. remove the `flex-auto` class
3. add the `flex-auto` class to the `Changelog` link

Maybe you can do this like:

```js
if (position.is('.flex-auto')) {
  position.classList.remove('flex-auto');
  button.classList.add('flex-auto');
  position.after(button)
}
```

HardikModha:
> Also, perhaps this should use `tiny-version-compare` to ensure that the tag we're comparing to is actually lower than the current one.
> 
> For example on: https://github.com/tensorflow/tensorflow/releases
> The tag right after `v1.12.1` is `TensorFlow 2.0.0-alpha0` and the generated link shouldn't be:
> 
> [tensorflow/tensorflow@v2.0.0-alpha0...v1.12.1](https://github.com/tensorflow/tensorflow/compare/v2.0.0-alpha0...v1.12.1)
> 
> but it should compare it against `v1.12.0`, further down the page
> 
> [tensorflow/tensorflow@v1.12.0...v1.12.1](https://github.com/tensorflow/tensorflow/compare/v1.12.0...v1.12.1)

Yep, you are right. I am aware of this limitation and have already mentioned it in a description. I didn't look for any packages (pure out of laziness maybe) 😅. Thanks for the `tiny-version-compare` suggestion. This wekend, I'm out so won't be able to work on it. Next week I'll give it a try.

bfred-it:
Hallelujah! Finally we merged it. Thanks for this feature!

HardikModha:
Woahhh, Thanks for refining the feature. ❤️ I just returned today and saw this merged!!! 🎉 

But seems like there is a small issue. Tags are not being compared with a namespace.  

Screen Shot 2019-05-14 at 8 48 20 PM
It should compare `parcel-bundler` with it's matching the previous tag, not with the `parcel-integration`. The tag name is not being decoded using before extracting `version` and `namespace` using `parseTags`.

HardikModha:
Also, I noticed a behavior with `version` aware comparison. It is comparing the first previous tag and not the closest matching previous release. 

e.g. On https://github.com/flutter/flutter/releases?after=v1.5.1 releases page. 

Screen Shot 2019-05-14 at 8 59 04 PM 

It should compare `v1.4.17` with `v1.4.16` and not `v1.4.9-hotfix.1`. What do you think?

PR #2000, opened 2019-05-01
Add `link-to-file-in-file-history` feature

### LINKED ISSUE #1979: In file History pages, add links to the file itself
Example: https://github.com/sindresorhus/refined-github/commits/2190e1a6f68e483f43bfda3a03d82542cda6f05f/source/features/show-names.js

There's a `<>` icon that points to the whole repo at the time of the commit, but this lacks a link to the file itself.

I think GitHub doesn't have a link here because they reused a regular "commits list", so the link to the whole repo makes sense. e.g. https://github.com/sindresorhus/refined-github/commits/master

But if a file is selected, then this fundamental link is missing.

HardikModha:
Closes #1979 

The feature can be tested on
- [Refined GitHub Readme](https://github.com/sindresorhus/refined-github/commits/master/readme.md)
- [React Readme](https://github.com/facebook/react/commits/master/README.md) - Commits with verified badge and comments
- [Spring Project File](https://github.com/spring-projects/spring-framework/commits/master/spring-core/src/main/java/org/springframework/core/ConfigurableObjectInputStream.java) - Very long path
- [Spring Project Directory](https://github.com/spring-projects/spring-framework/commits/master/spring-core/src/main/java/org/springframework/core/serializer) - Directory


source/libs/page-detect.ts
```diff
@@ -97,6 +97,16 @@ export const isTrending = (): boolean => location.pathname === '/trending' || lo
 
 export const isUserProfile = (): boolean => select.exists('.user-profile-nav');
 
+export const isHistoryPage = (): boolean => {
```
bfred-it: These detections are ideally URL-based. It’s best you use `isCommitList` and then detect the required elements in the feature itself.

source/features/link-to-file-in-file-history.tsx
```diff
@@ -0,0 +1,45 @@
...
+import select from 'select-dom';
+import features from '../libs/features';
+import {getOwnerAndRepo} from '../libs/utils';
+import {file} from '../libs/icons';
+import {groupButtons} from '../libs/group-buttons';
+
+function init(): void {
+	const {repoName} = getOwnerAndRepo();
+
+	const pathToFile = select('.breadcrumb')!.textContent!.trim().replace(new RegExp(`^History for ${repoName}`), '');
```
bfred-it: Find the `path` from `getRepoPath()`

source/features/link-to-file-in-file-history.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+	for (const commitLinkCell of select.all<HTMLDivElement>('.commit-links-cell')) {
+		const fileAnchor = (
+			<a
+				href={`${select<HTMLAnchorElement>('.commit-links-group a.btn', commitLinkCell)!.pathname.replace('/commit/', '/blob/')}${pathToFile}`}
```
bfred-it: You can move this to a variable

source/features/link-to-file-in-file-history.tsx
```diff
@@ -0,0 +1,45 @@
...
+import {getOwnerAndRepo} from '../libs/utils';
+import {file} from '../libs/icons';
+import {groupButtons} from '../libs/group-buttons';
+
+function init(): void {
+	const {repoName} = getOwnerAndRepo();
+
+	const pathToFile = select('.breadcrumb')!.textContent!.trim().replace(new RegExp(`^History for ${repoName}`), '');
+
+	for (const commitLinkCell of select.all<HTMLDivElement>('.commit-links-cell')) {
```
bfred-it: What if you select the `tree` link that’s going to be next to this button? And then use `.closest` to get this cell. I think this enables some optimizations

source/features/link-to-file-in-file-history.tsx
```diff
@@ -0,0 +1,45 @@
+/**
+This feature adds direct link to file/directory when viewing the history.
+GitHub provides a link but it points to the root of the repo which is not that helpful.
```
bfred-it: You can drop “this feature” and the second line

source/libs/page-detect.ts
HardikModha: Sure. I will. Maybe we can move it back to `page-detect` if we have another feature for the history pages.

HardikModha: No? 😅 No worries. We won't :)

source/features/link-to-file-in-file-history.tsx
HardikModha: What kind of optimizations? Can you please tell me more? I selected the parent first and then applied the `tree` link selectors to reduces the chances of any false matches.

bfred-it: fe9629b

I always try to reuse elements as much as possible, for example in this case I just got one element to get:

- insertion position
- file path (which it turns out isn't needed)
- starting point to add `width: auto`

HardikModha: Simply great! 👍 I wasn't aware of this automatic redirection thing. 😄 TIL :)

bfred-it:
2000 ✨

PR #2002, opened 2019-05-02
Use and improve `fetch-dom`

bfred-it:


bfred-it:
Firefox doesn't let me override global `fetch` 😒

Issue #2005, opened 2019-05-02
Rename "Blame" to "Praise"


gaborcsardi:
To stay positive, how about renaming the "Blame" button in the source code view, to "Praise", and potentially also in the pop-up menu when you click on a line, and then on the "..."?

E.g. on 
https://github.com/sindresorhus/refined-github/blob/master/source/content.css

<img width="490" alt="Screen Shot 2019-05-02 at 9 59 58 PM" src="https://user-images.githubusercontent.com/660288/57106583-b20e2600-6d25-11e9-8ad9-adab4988a247.png">
<img width="590" alt="Screen Shot 2019-05-02 at 10 00 08 PM" src="https://user-images.githubusercontent.com/660288/57106584-b20e2600-6d25-11e9-8d99-dffb076d79e3.png">

lazarljubenovic:
To stay positive, never let a well-established command name affect you negatively.

bfred-it:
I understand the intent but this would confuse people rather than help them, so we’re not the right place to do this.

sindresorhus:
Instead, request it at support@github.com.

PR #2012, opened 2019-05-06
In discussions, always link the "Closed" label to the closing event

loilo:
Recap: Besides linking to a concrete PR/commit, the `extend-status-labels` feature also puts an anchor to the closing note on the status label (which I find really handy since the most valuable/helpful comments are often right before the closing note).

However, those anchors are currently not set when an issue was *not* closed by a PR/commit (e.g. #2005).

This PR changes that behavior by not bailing out prematurely when no commit/PR was found. Instead it always puts the anchor to the closing note and adds the `in #xxxx` label + link if needed.

### DIFF
--- a/source/features/extend-status-labels.tsx
+++ b/source/features/extend-status-labels.tsx
@@ -8,24 +8,33 @@ function init(): false | void {
 		return false;
 	}
 
+	// Get the latest status update
 	const lastActionRef = select.all(`
-		.discussion-item-closed [href*="/pull/"],
-		.discussion-item-closed code,
+		.discussion-item-closed,
 		.discussion-item-reopened,
-		.discussion-item-merged [href*="/commit/"]
+		.discussion-item-merged
 	`).pop();
 
-	// Leave if it was never closed or if it was reopened
+	// Leave if the issue/PR was never closed or if it was reopened
 	if (!lastActionRef || lastActionRef.matches('.discussion-item-reopened')) {
 		return false;
 	}
 
+	// Find reference to a commit/PR
+	const lastActionLink = select('[href*="/pull/"], [href*="/commit/"], code', lastActionRef);
+
 	const label = select('.gh-header-meta .State')!;
-	const isMerged = lastActionRef.closest('.discussion-item-merged');
-	label.append(isMerged ? ' as ' : ' in ', lastActionRef.cloneNode(true));
+
+	if (lastActionLink) {
+		const isMerged = lastActionLink.closest('.discussion-item-merged');
+		label.append(isMerged ? ' as ' : ' in ', lastActionLink.cloneNode(true));
+	}
+
+	// Select the label inside the action ref and from there search an ID upwards
+	const closestIdSelector = '#' + select('.discussion-item-header', lastActionRef)!.closest('[id]')!.id;
 
 	// Link label to event in timeline
-	wrap(label, <a href={'#' + lastActionRef.closest('[id]')!.id}/>);
+	wrap(label, <a href={closestIdSelector}/>);
 }
 
 features.add({
--- a/source/features/extend-status-labels.tsx
+++ b/source/features/extend-status-labels.tsx
@@ -26,7 +26,7 @@ function init(): false | void {
 	const label = select('.gh-header-meta .State')!;
 
 	if (lastActionLink) {
-		const isMerged = lastActionLink.closest('.discussion-item-merged');
+		const isMerged = lastActionRef.matches('.discussion-item-merged');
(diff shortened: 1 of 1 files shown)


source/features/extend-status-labels.tsx
```diff
@@ -8,24 +8,33 @@ function init(): false | void {
...
-		.discussion-item-closed [href*="/pull/"],
-		.discussion-item-closed code,
+		.discussion-item-closed,
 		.discussion-item-reopened,
-		.discussion-item-merged [href*="/commit/"]
+		.discussion-item-merged
 	`).pop();
 
-	// Leave if it was never closed or if it was reopened
+	// Leave if the issue/PR was never closed or if it was reopened
```
Loilo: Changed the wording here because the "it" may be confusing in regards to the comment I added in line 11.

source/features/extend-status-labels.tsx
```diff
@@ -8,24 +8,33 @@ function init(): false | void {
 		return false;
 	}
 
+	// Get the latest status update
 	const lastActionRef = select.all(`
```
bfred-it: This no longer a link to the `ref`, but just the last action. Perhaps `lastStatusChange` and you can drop the comment.

```suggestion
	const lastStatusChange = select.all(`
```

source/features/extend-status-labels.tsx
```diff
@@ -8,24 +8,33 @@ function init(): false | void {
...
 		return false;
 	}
 
+	// Find reference to a commit/PR
+	const lastActionLink = select('[href*="/pull/"], [href*="/commit/"], code', lastActionRef);
+
 	const label = select('.gh-header-meta .State')!;
-	const isMerged = lastActionRef.closest('.discussion-item-merged');
-	label.append(isMerged ? ' as ' : ' in ', lastActionRef.cloneNode(true));
+
```
bfred-it: This can be here, no comment necessary

```suggestion
 	const lastActionLink = select('[href*="/pull/"], [href*="/commit/"], code', lastActionRef); 
```

source/features/extend-status-labels.tsx
```diff
@@ -8,24 +8,33 @@ function init(): false | void {
...
+	if (lastActionLink) {
+		const isMerged = lastActionLink.closest('.discussion-item-merged');
+		label.append(isMerged ? ' as ' : ' in ', lastActionLink.cloneNode(true));
+	}
+
+	// Select the label inside the action ref and from there search an ID upwards
+	const closestIdSelector = '#' + select('.discussion-item-header', lastActionRef)!.closest('[id]')!.id;
```
bfred-it: Probably just this, the id is on the first child of `lastActionRef`

```suggestion
	const lastActionURI = '#' + select('[id]', lastActionRef)!.id;
```

Loilo: Yep, realized that as well but wasn't sure if we wanted to rely on it (although we probably have bigger problems when the markup changes in that way 😁).

source/features/extend-status-labels.tsx
Loilo: Oh okay, thought the "ref" was some React-related convention thingy. 👍

source/features/extend-status-labels.tsx
Loilo: Not exactly sure from the preview what you're suggesting, do you mean "drop the comment & move after `const label = ...`"?

source/features/extend-status-labels.tsx
Loilo: Resolved in 4da9b5e

source/features/extend-status-labels.tsx
```diff
@@ -8,24 +8,33 @@ function init(): false | void {
...
 
+	// Find reference to a commit/PR
+	const lastActionLink = select('[href*="/pull/"], [href*="/commit/"], code', lastActionRef);
+
 	const label = select('.gh-header-meta .State')!;
-	const isMerged = lastActionRef.closest('.discussion-item-merged');
-	label.append(isMerged ? ' as ' : ' in ', lastActionRef.cloneNode(true));
+
+	if (lastActionLink) {
+		const isMerged = lastActionLink.closest('.discussion-item-merged');
```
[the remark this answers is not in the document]
Loilo: Resolved in 4da9b5e

source/features/extend-status-labels.tsx
bfred-it: I think I had used `ref` in the git sense, like tags and commits are "refs"

Loilo:
The reviewed points have all been resolved, so feel free to look at the code again. 🙂

bfred-it:
Thanks @loilo! 🙋‍♂️

PR #1934, opened 2019-04-16
Add `sync-pr-commit-title` feature (replaces `fix-squash-and-merge-title`)

bfred-it:
Fixes https://github.com/sindresorhus/refined-github/issues/1790
Closes https://github.com/sindresorhus/refined-github/issues/1743

- When the commit title is manually updated, RGH will offer to also update the PR title
- Title/message features merged into one
- The title box is focused after it's opened
- The message box now expands to fit the content
- This now applies to Merge commits too

# Next

- [x] Readme
- [x] Add migration in background.js
- [x] Should we drop “fix-squash-and-merge-message”? I'm not entirely sure the new message is useful anyway and it contains a lot of noise ("fixes this and that", "I tried this", "What do you think", PR template comment, etc)
- [x] Does the "PR title update" part belong to the same feature? There's some overlap and both features kinda do the same thing, in two directions

# Test 

<details>
<summary>Commit info update</summary>

1. Visit PR with ONE commit (with 2+ commits, GitHub already updates **the title**)
2. Open the merge box (as Squash or Merge)
---
1. Visit PR
2. Change PR title
2. Open the merge box (as Squash or Merge)

</details>

<details>
<summary>PR title update</summary>

1. Visit PR
2. Open the merge box (as Squash or Merge)
3. Change title (the message will appear)
4. Undo changes (the message will disappear)
5. Merge (the PR title will be updated if the message was shown)
---
1. Visit PR
2. Open the merge box (as Squash or Merge)
3. Change title (the message will appear)
4. Press `cancel` (the message will disappear)
5. Merge (the PR title will **not** be updated)

</details>

**Untested**

Where admin permissions are required

Issue #276, opened 2016-07-30
Force "squash and merge" title to be the pull request title


sindresorhus:
It does if there are multiple commits, but if there is only one commit, if will use the commit title instead. I often rename pull requests because submitters do a very bad job of naming their pull requests. When I then "squash and merge" it will instead choose the commit title, which is the original bad title, and I have to rename yet again. Would be nice if we could just copy in the pull request title when "Merge pull request" is clicked and you're in the confirm view.

hkdobrev:
@sindresorhus Have you contacted GitHub about that? It seems like that should be the default.

sindresorhus:
@hkdobrev No. I'm tired of wasting time sending them feature requests (I've sent 100+ through the years) and getting back a generic fake-nice response and never seeing anything done. That's exactly why I made this extension. Feel free submit the request yourself though.

apapko:
Has it been reported that PR title is not used as commit message if pull request contains only one commit?
ex pr title: ```feat(test): squash test ```

alerque:
I've just found this extension and like many of the things it fixes. This is actually one "feature" I'd like to turn off though. I find that in PR's with one commit the author often has taken more care with the commit message than the PR title, and that's a better default to start with. Since the interface lets me adjust the commit message anyway, starting with the least intrusive option of not touching the commit message seems like a better place to start. Your adjustment isn't suited to all workflows, and I think the default Github behaviour is better for mine.

bfred-it:
> . Your adjustment isn't suited to all workflows, and I think the default Github behavior is better for mine.

That's why you can disable it as `fix-squash-and-merge-title`

Generally I found the opposite, where the commit is something like `Update README.md` and the title has any sense.

This very feature is actually being extended in https://github.com/sindresorhus/refined-github/pull/1934 to also fix the PR title as shown here: https://github.com/sindresorhus/refined-github/issues/1743

alerque:
Thanks for the feedback @bfred-it, I hadn't even discovered the extensions allowed selectively disabling features — only that it did something I didn't like out of the box. That's awesome, and just the Firefox extension just picked up five stars from me.

bfred-it:
That's good to hear :D

joeshaw:
FYI, this feature appears to apply to regular merge commits now as well.

bfred-it:
That was an intended change in https://github.com/sindresorhus/refined-github/pull/1934

bharathwaaj:
[...]

PR #2017, opened 2019-05-07
Bring back fix-squash-and-merge-message

zachwhaley:
The `fix-squash-and-merge-message` feature was recently removed from refined-github.
I asked about this and the reply was that it would not return.
See https://github.com/sindresorhus/refined-github/commit/9be9440699d2b8040af0c7c26ae524169b3cd0b8

But I am not the only one that uses this very helpful feature, @bharathwaaj https://github.com/sindresorhus/refined-github/pull/1934#issuecomment-490053460, and I would assume there are more that would like to see this come back.

So I am opening this PR to start a discussion about why the `fix-squash-and-merge-message` was removed and if we can bring it back in some form.

Thank you!

### DIFF
--- a/source/background.ts
+++ b/source/background.ts
@@ -14,12 +14,16 @@ new OptionsSync().define({
 	migrations: [
 		options => {
 			options.disabledFeatures = (options.disabledFeatures as string)
+				.replace('milestone-navigation', '') // #1767
+				.replace('op-labels', '') // #1776
+				.replace('delete-fork-link', '') // #1791
+				.replace('exclude-filter-shortcut', '') // #1831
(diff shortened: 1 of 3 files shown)


bfred-it:
I already said why it would not return, it causes noise in the commit message and all the linked issues.

vtintillier:
Hello @bfred-it and @sindresorhus ,

I can understand that sometimes the PR description is not suitable for the commit message.
However as you say in https://github.com/sindresorhus/refined-github/issues/276#issuecomment-484792346 features can be individually enabled/disabled, so why can't we add this feature, disabled by default, and let people with a different workflow (i.e. where PR description is meant to be the final commit message) enable it?

Thanks a lot.

PR #2019, opened 2019-05-09
Reliable merge box handlers

bfred-it:
Depends on https://github.com/sindresorhus/mem/pull/35

Maybe these could benefit from this as well, once they're merged:
- [ ] [~~#1950~~](https://github.com/sindresorhus/refined-github/pull/1950) (edit: no, the button appears outside the form, not inside) 
- [x] #2023 (✅ 966e14b) 

Reason of the changes:

- `details:toggled` was unreliable (small fix)
- `fit-textareas` was being called from an unrelated feature
- `add-co-authored-by` wasn't being added when the session was restored (i.e. type a new PR title and reload the page)
- `add-co-authored-by` was listening to a click on a button, which conflicted with the future #1950


source/libs/on-pr-merge-panel-open.ts
```diff
@@ -0,0 +1,39 @@
+// Memoization here is used to let onPrMergePanelOpen() be called multiple times without risking multiple attached handlers
+import mem from 'mem';
+import delegate, {DelegateSubscription, DelegateEvent} from 'delegate-it';
+
+const delegateHandler = mem((callback: EventListener) => (event: DelegateEvent) => {
+	if (event.delegateTarget.matches('.open')) {
+		callback(event);
+	}
+});
+
```
bfred-it: I love this pattern.

Since `addEventListener` doesn't add the same handler twice, it's important that the same exact handler is passed every time. To do this I always use a `WeakMap` and an ugly sequence of `get or create, then set and return`, but I think I finally found a way to hide this complexity with `mem`.

sindresorhus: I like this pattern a lot 👍

bfred-it: Also used here https://github.com/bfred-it/delegate-it/pull/10

Screenshot 2019-05-17 at 21 58 06

PR #2020, opened 2019-05-10
Delete contributing.md

mohamademampour:


yakov116:
@mohamademampour what in world are you doing?

PR #2021, opened 2019-05-10
Checkbox option to use PR description as merge commit message

zachwhaley:
A couple days ago I responded to a comment in regards to the squash and merge message feature that was removed asking if refined-github might consider returning it as an optional feature, see https://github.com/sindresorhus/refined-github/issues/276#issuecomment-490123715

This PR is a final attempt to return the squash and merge message feature.  In this change I've added a checkbox to allow the user to decide if they want to use the PR's description as their merge commit message before merging.  Note, I am very new to any kind of browser extension development and this is not anywhere near complete so feedback is very welcome.

merge-message-option


bfred-it:
I don’t think I’m clear enough. Stop opening PRs. It was rejected and it’s not coming back. It creates noise and I’m not adding UI for a feature that creates noise.

Issue #1792, opened 2019-02-15
`wait-for-build` currently broken


bfred-it:
[<img alt="Issuehunt badges" src="https://issuehunt.io/r/sindresorhus/refined-github/issues/1792/badge.svg" />](https://issuehunt.io/r/sindresorhus/refined-github/issues/1792)

"Wait for successful checks" isn't waiting 😰 

https://github.com/sindresorhus/refined-github/blob/master/source/features/wait-for-build.tsx

> There is a $60.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/sindresorhus/refined-github/issues/1792).
> - Checkout the [Issuehunt explorer](https://issuehunt.io/r/sindresorhus/refined-github/) to discover more funded issues.
> - Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on Issuehunt to raise funds.

IssuehuntBot:
@issuehunt has funded $40.00 to this issue.

---
- Submit pull request via [IssueHunt](https://issuehunt.io/repos/51769689/issues/1792) to receive this reward.
- Want to contribute? Chip in to this issue via [IssueHunt](https://issuehunt.io/repos/51769689/issues/1792).
- Checkout the [IssueHunt Issue Explorer](https://issuehunt.io/issues) to see more funded issues.
- Need help from developers? [Add your repository](https://issuehunt.io/users/44827199/repositories) on IssueHunt to raise funds.

gr2m:
I assume this is the feature that should show a checkbox that would wait until all checks are green? I just installed Refined GitHub for the first time (latest Chrome on Mac) and don’t see it, here is a screencast

![Kapture 2019-04-05 at 9 45 17](https://user-images.githubusercontent.com/39992/55643356-c7744c80-5787-11e9-89dc-a269abd1a860.gif)

IssuehuntBot:
@gr2m has funded $20.00 to this issue.

---
- Submit pull request via [IssueHunt](https://issuehunt.io/repos/51769689/issues/1792) to receive this reward.
- Want to contribute? Chip in to this issue via [IssueHunt](https://issuehunt.io/repos/51769689/issues/1792).
- Checkout the [IssueHunt Issue Explorer](https://issuehunt.io/issues) to see more funded issues.
- Need help from developers? [Add your repository](https://issuehunt.io/users/39992/repositories) on IssueHunt to raise funds.

Loilo:
[...]

PR #2023, opened 2019-05-11
Fix `wait-for-build`

bfred-it:
Fixes https://github.com/sindresorhus/refined-github/issues/1792

Tested in Chrome and Firefox


bfred-it:
Tested again in Chrome for:

- new commit appears (submission is cancelled)
- ci fails (submission is cancelled)
- ci succeeds (PR is merged)
- cancel and click "merge" multiple times (it keeps waiting and then merges)

PR #2026, opened 2019-05-12
Add `add-tag-to-commits` feature

### LINKED ISSUE #807: Show tags in commit history
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/refined-github/issues/807)

Add a link to a tag that is set on a commit when looking at the commit history. This makes it easier to identify new commits that haven't been released yet, or identify change ranges.

I realize you can use the pull request/diffing UI for that, but that's way more complicated than looking at the history, especially when a tag only includes a few commits. Plus you need to know the tag name in to compare, since those are not auto completed by the UI.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $65.00)

- [issuehunt issuehunt](https://issuehunt.io/u/issuehunt) ($60.00)

### Tips

---
</details>

dotconnor:
Note: For some reason, the API returns empty data on some tags

Closes #807


source/features/add-tag-to-commits.tsx
```diff
@@ -0,0 +1,67 @@
...
+							}
+						}
+					}
+				}
+			}
+		}
+	}`);
+	let tags: Tag[] = repository.refs.edges.map((edge: any) => edge.node.target).filter((tag: Tag) => tag.name && tag.commitResourcePath).map((tag: Tag) => ({...tag, commit: tag.commitResourcePath.split('/')[4]}));
+	if (repository.refs.pageInfo.hasNextPage) {
+		tags = tags.concat(await getTags(repository.refs.pageInfo.endCursor));
```
bfred-it: This is doing exactly something that I didn’t want to do: fetch multiple pages of tags on every Commits page 😕

dotconnor: From what I've seen 100 tags (1 request) is enough for most repos, the only encounter I had with more than that was [facebook/react](https://github.com/facebook/react)

bfred-it: It's true, but also there are repos like https://github.com/nodejs/node

Are 6 sequential requests acceptable? Are these heavy? How difficult would it be to only query the next page if no matching tags are found on the current one?

dotconnor: [nodejs/node](https://github.com/nodejs/node) doesn't tag their releases on master so even if you did, you would still pull all 6 requests each time.

sindresorhus: > fetch multiple pages of tags on every Commits page 

I assume we can cache the tags? At least the older ones.

> Are these heavy?

@dotconnor ⬆️

source/features/add-tag-to-commits.tsx
```diff
@@ -0,0 +1,65 @@
...
+					{icons.tag()}
+					<span className="ml-1">{targetTags.map(tags => tags.name).join(', ')}</span>
+				</div>
+			);
+		}
+	}
+}
+
+features.add({
+	id: 'add-tag-to-commits',
```
sindresorhus: Can you pull from `master` and then add a `description` field here?

source/features/add-tag-to-commits.tsx
bfred-it: > I assume we can cache the tags? At least the older ones.

Cache where? Our cache is rather small and if we're gonna store them in a regular `browser.storage.local` we also need a cleaning mechanism so it doesn't stay there forever.

> Are these heavy?

I think the definition of heavy is "what's the [API cost](https://developer.github.com/v4/guides/resource-limitations/#returning-a-calls-rate-limit-status)?" and "how long does it take for an answer?"

If the tags appear 5 seconds after dom ready then I probably won't see them.

dotconnor:
> Note: For some reason, the API returns empty data on some tags

The API only returns the Tag type on annotated tags and not lightweight ones for example: https://github.com/sindresorhus/refined-github/commit/123425df7037f6453122930f7e6146b0ae70a4dc, and from what I've can tell you can't retrieve the tag name for the Github API.

source/features/add-tag-to-commits.tsx
dotconnor: > I assume we can cache the tags? At least the older ones.

Sure we could cache the tags, but you would still have to make the API calls to validate that the remaining commits don't have tags. So caching based on the commit (key) and its associated tags (value), and only make a network request if a commit isn't found in the cache, would work best.

> Cache where? Our cache is rather small and if we're gonna store them in a regular browser.storage.local we also need a cleaning mechanism so it doesn't stay there forever.

I think a simple LRU implementation on top of `browser.storage.local` with lazy saving/cleaning should suffice. 

> Are these heavy?

Speed? Really depends on Github, from my testing its a little all over the place from ~200ms to sometimes a full second.

Size? A full page of tags (100) is about 4KB

bfred-it: > I think a simple LRU implementation on top of `browser.storage.local` with lazy saving/cleaning should suffice.

If you want to implement that it would be great, you can keep it as a separate PR and replace the existing cookie cache.

bfred-it:
> The API only returns the Tag type on annotated tags and not lightweight ones

This works: https://stackoverflow.com/a/53923870/288906


Screenshot 2019-05-25 at 16 26 03


Compared to the query in this PR:


Screenshot 2019-05-25 at 16 26 11

source/features/add-tag-to-commits.tsx
```diff
@@ -0,0 +1,73 @@
...
+interface Tag {
+	name: string;
+	commit: string;
+}
+
+async function getTags(after?: string): Promise<Tag[]> {
+	const {ownerName, repoName} = getOwnerAndRepo();
+	const {repository} = await api.v4(`{
+		repository(owner: "${ownerName}", name: "${repoName}") {
+			refs(first: 100, refPrefix: "refs/tags/", orderBy: {field: TAG_COMMIT_DATE, direction: DESC}${after ? `, after:"${after}"` : ''}) {
```
bfred-it:
```suggestion
			refs(
				first: 100,
				refPrefix: "refs/tags/",
				orderBy: {
					field: TAG_COMMIT_DATE,
					direction: DESC
				}
				${after ? `, after: "${after}"` : ''}
			) {
```

source/features/add-tag-to-commits.tsx
```diff
@@ -0,0 +1,73 @@
...
+					hasNextPage
+					endCursor
+				}
+				nodes {
+					name
+					target {
+						... on Tag {
+							commitResourcePath
+						}
+						... on Commit {
```
bfred-it: Can these two `... on` be avoided or merged?

source/features/add-tag-to-commits.tsx
```diff
@@ -0,0 +1,73 @@
...
+							commitResourcePath
+						}
+					}
+				}
+			}
+		}
+	}`);
+	let tags: Tag[] = repository.refs.nodes.map((node: any) => ({name: node.name, commit: node.target.commitResourcePath.split('/')[4]}));
```
bfred-it:
```suggestion
	let tags: Tag[] = repository.refs.nodes.map((node: any) => ({
		name: node.name,
		commit: node.target.commitResourcePath.split('/')[4]
	}));
```

dotconnor:
I hate the concept of pages because there is no way to guarantee you get the same response in each request. Therefore caching based on pages would also be problematic, they would have to be very short-lived because the first page will always change anytime a new tag is created, otherwise the feature will become unreliable. Heres what I think:

1. Keep all the commits and their tags in a single object for each repo, being kept in the cache for a longer amount of time.
```js
commitsOnPage = select.all()
let tags = cache.get()
while (not all commitsOnPage in tags) {
	tags = tags + getMoreTags()
}
cache.set(tags)
for (commit of commitsOnPage) {
	// add tags[commit] to page
}
```

Cons:
- If another tag is added to commit, it may not show for a while


2. Keep a different cache item for each commit, and only fetch the remaining commits:
```js
commitsOnPage = select.all()
for (commit of commitsOnPage) {
	tag = cache.get(commit) // if the commit was found while fetching the previous commit it will be in the cache now
	if (!tag) {
      tag = getTagForCommit(commit)
    }
    // add tag to page
}
function getTagForCommit(commit) {
	const tags = {};
	while (!commitFound) {
		tags = {...tags, apiCall()}
        for (tag of tags) {
        	cache.set(tag)
		}
        if (tags[commit]) {
        	return tags[commit]
        }
	}
	return []
}
```

bfred-it:
The cache would not be per page, but per commit. That doesn’t change. For every fetched page, you can extend the cache. 

However… I just realized that neither of our solutions actually limits the pages fetched because both keep fetching pages until all commits on the page are tagged… which is wrong. Not all commits will have a tag so the code will always fetch all pages looking for a match. 

I guess this search would have to be even smarter and complicated because the search has to stop when it finds a tag that is on a commit older than what’s on the page. 

🤷‍♂️

bfred-it:
Unless you have a better idea, let’s keep it easy for now:

- Fetch pages until the fetched page has some overlap with the cache. 

The cache should already have all the tags, except the very latest. When you fetch a page with tag already in the cache (i.e. overlap), it can stop fetching more pages.

dotconnor:
We can get the timestamp from tag/commit and also ask for the time of the last commit on the page:

```graphql
{
    refs(first: 100, refPrefix: "refs/tags/", orderBy: {field: TAG_COMMIT_DATE, direction: DESC}) {
      pageInfo {
        hasNextPage
        endCursor
      }
      nodes {
        name
        target {
          commitResourcePath
          ... on Tag {
            tagger {
              date
            }
          }
          ... on Commit {
            committedDate
          }
        }
      }
    }
    object(expression: "9b2acbbc0023c5df62c55b2595febd8277d652eb") {
      ... on Commit {
        committedDate
      }
    }
  }
}

```

If the date of the last commit is younger than the last tag then stop looking for tags?

bfred-it:
I didn’t think about the date, that works too. However I’m trying to find the least complex solution and maybe having to keep track of dates _as well_ might be more complicated than the last solution I proposed. 

Thoughts?

dotconnor:
I'll try both and see which one works better.

dotconnor:
@bfred-it Since we are keeping each commit as a separate cache entry, it would be nice to have a cache function to get/set multiple at the same time:

```
cache.getAll([...commits])
```

Maybe convert our current API to mimic the native one while still adding expiration

bfred-it:
> Found a possible problem in your solution, using [nodejs/node](https://github.com/nodejs/node/commits/master) as a test page, using the cache overlap solution you still will fetch all the pages. Comparing dates you will only fetch the first one.

Good point, my solution will fetch all the pages at first to fill the cache, dates can allow us to be more selective.

> Both of them also share a problem but I don't know how major they will be: If we stumble upon a commit not in the cache, no matter what "page" the commit appears on, we will still have to fetch all pages leading up to that page. This should be minor unless you go back a few pages in the UI (I don't know why you would do this though.)

Hmmm no, because as I suggested [here](https://github.com/sindresorhus/refined-github/pull/2026#issuecomment-500097926) we stop as soon as there's any overlap, we don't look for _every_ commit.

> Also, if there are two tags on the same commit separated by a page, there is a chance that only one of these tags will appear.

My [later solution](https://github.com/sindresorhus/refined-github/pull/2026#issuecomment-500097926) fetches _all_ the tags at first, it doesn't have this problem.

A date-based solution would not suffer this problem either, unless this commit is the last on the current page (github.com) as well as the last tag on the current tag page (api). It's a rare occurrence, I'd say, especially considering how rarely a commit has multiple tags.

> @bfred-it Since we are keeping each commit as a separate cache entry, it would be nice to have a cache function to get/set multiple at the same time:

The cache should be grouped by repo:

```js
cache.set('tags:sindresorhus/refined-github', {
	'eb9feb9': ['v1'],
	'a0ce6e6': ['v2', v2-beta]
});
```

dotconnor:
Okay so we're back to the two different solutions

 - [ ] fetch all the tags on the first run
   - large upfront cost
 - [ ] fetch based on the dates
   - large cost down the road if you wander too far back into the past

bfred-it:
Dates sound good, if the resulting code isn't too complex. Do you think you'll also have to store the commit date in the cache for this to work?

```js
cache.set('tags:sindresorhus/refined-github', {
	'eb9feb9': {tags: ['v1'], date},
	'a0ce6e6': {tags: ['v2', 'v2-beta'], date}
});
```

dotconnor:
I have it working without storing the dates, but I also store when a commit has no tags:

```js
cache.set('tags:sindresorhus/refined-github', {
	'eb9feb9': ['v1'],
	'a0ce6e6': ['v2', 'v2-beta'],
    'cc2c967': []
});
```

bfred-it:
> I also store when a commit has no tags:

I thought about that, but wouldn't that get too heavy? Some repos have maaany commits.

dotconnor:
> I also store when a commit has no tags

I did this to avoid an API call if every commit is cached, the same effect can be achieved other ways I believe.

> I thought about that, but wouldn't that get too heavy? Some repos have maaany commits.

Yeah storing it in a single key would make it a little heavy, since when you add a key to it, it will basically never get expired.

That's why I assumed we would keep a different key for each commit that way you can expire the commits that you no longer see.

> Let's see it, if you already wrote it

I did 😃, cleaning up lint things

source/features/add-tag-to-commits.tsx
```diff
@@ -0,0 +1,142 @@
...
+			commitsWithNoTags.push(targetCommit);
+		}
+
+		if (targetTags && targetTags.length > 0) {
+			select('.commit-meta', commit)!.append(
+				<div className="ml-2">
+					{icons.tag()}
+					<span className="ml-1">{targetTags.map((tags, i) => (
+						<>
+							<a href={`/${getRepoURL()}/releases/${tags}`}>{tags}</a>
```
jerone:
```suggestion
							<a href={`/${getRepoURL()}/releases/tag/${tags}`}>{tags}</a>
```

sindresorhus:
@dotconnor Can you fix the merge conflict?

sindresorhus:
Thanks for working on this, @dotconnor 👌

fregante:
This is pretty useful even to Refined GitHub, when looking at the [commit log](https://github.com/sindresorhus/refined-github/commits/master):

PR #2038, opened 2019-05-13
Fix `hide-empty-meta`

### LINKED ISSUE #2037: Feature `hide-empty-meta` is broken
> I think it's broken as it's supposed to hide this italics text
> 
> Screenshot 2019-05-13 at 23 30 11
>
> Live demo: https://github.com/bfred-it/temp

_Originally posted by @bfred-it in https://github.com/sindresorhus/refined-github/pull/2030/files/5ac58f64b8e7d4479b84acc56b1fe1723ae042e2_

notlmn:
Fixes #2037

# Test

https://github.com/Witold101/wpw

### DIFF
--- a/source/features/hide-empty-meta.tsx
+++ b/source/features/hide-empty-meta.tsx
@@ -2,10 +2,8 @@ import select from 'select-dom';
 import features from '../libs/features';
 
 function init(): void {
-	const emptyDescription = select('.repository-content > :not(#readme):not(.commit-tease) em');
-
-	if (emptyDescription && !select.exists('#repo-meta-edit')) {
-		emptyDescription.style.setProperty('display', 'none', 'important');
+	if (select.exists('.repository-content em') && !select.exists('.js-edit-repo-meta-button')) {
+		select('.repository-content').firstElementChild!.remove();
 	}
 }
 
--- a/source/features/hide-empty-meta.tsx
+++ b/source/features/hide-empty-meta.tsx
@@ -2,8 +2,10 @@ import select from 'select-dom';
 import features from '../libs/features';
 
 function init(): void {
-	if (select.exists('.repository-meta em') && !select.exists('.js-edit-repo-meta-button')) {
-		select('.repository-meta')!.style.display = 'none';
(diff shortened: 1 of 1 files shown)


source/features/hide-empty-meta.tsx
```diff
@@ -2,8 +2,10 @@ import select from 'select-dom';
 import features from '../libs/features';
 
 function init(): void {
-	if (select.exists('.repository-meta em') && !select.exists('.js-edit-repo-meta-button')) {
-		select('.repository-meta')!.style.display = 'none';
+	const emptyDescription = select('.repository-content > :not(#readme):not(.commit-tease) em');
```
bfred-it: There are a lot of other children in `.repository-content`, we can't go by exclusion

Screenshot 2019-05-14 at 00 22 06

It should probably start with `#repo-meta-edit` or maybe just be `#repo-meta-edit em`. Double-check

yakov116: I also think the selector is different if you have rights to edit

notlmn: > There are a lot of other children in `.repository-content`, we can't go by exclusion

But usually commit list an the the readme are the only likely ones to contain an `em`.

-----------------

> It should probably start with `#repo-meta-edit` or maybe just be `#repo-meta-edit em`. Double-check

Targeting `#repo-meta-edit em` fails for repos with no permissions, the existence of `#repo-meta-edit` is the condition to bail.

notlmn: > I also think the selector is different if you have rights to edit

`#repo-meta-edit` only exists if you have permission, else its just `em` inside a "`div`-tree-with-utility-class" hell.

bfred-it: True, actually because it's so high in the dom we might get away with just `.repository-content em` because `select` only gets the first one

bfred-it: Screenshot 2019-05-14 at 00 38 15

Yeah it's definitely the first one

bfred-it: But it should drop `d-flex`, not just em, otherwise there's some space left behind

PR #2042, opened 2019-05-14
Decode tag before extracting version and namespace

HardikModha:
Fixes https://github.com/sindresorhus/refined-github/pull/1998#issuecomment-492285094

Tested it on 

1. https://github.com/parcel-bundler/parcel/releases (Namespaced tags)
2. https://github.com/facebook/react/releases (Tags with the beta, alpha versions)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10827,6 +10827,11 @@
 				"semver": "^5.0.3"
 			}
 		},
+		"semver-regex": {
+			"version": "3.1.0",
+			"resolved": "https://registry.npmjs.org/semver-regex/-/semver-regex-3.1.0.tgz",
+			"integrity": "sha512-xJqZonbAohJmpogzq1Mx7DB4DT3LO+sG5J4i/rcRyZ527dcUqsTVb5T0vM5gwOBf3KO0v7i94EpbdvwSukJyAQ=="
+		},
 		"serialize-error": {
 			"version": "2.1.0",
 			"resolved": "https://registry.npmjs.org/serialize-error/-/serialize-error-2.1.0.tgz",
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
 		"onetime": "^5.0.0",
 		"p-memoize": "^3.0.0",
 		"select-dom": "^5.0.2",
+		"semver-regex": "^3.1.0",
 		"shorten-repo-url": "^1.5.1",
 		"tiny-version-compare": "^1.0.0",
 		"type-fest": "^0.4.1",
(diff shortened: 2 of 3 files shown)


bfred-it:
> It is comparing the first previous tag and not the closest matching previous release.

We could sort the tags with `tiny-version-compare`, using something like `tags.sort((a, b) => versionCompare(a.version, b.version))` this however complicates things a lot when the tags can't be properly sorted (e.g. no version or non-standard version)

jerone:
You could use the use [to-semver](https://www.npmjs.com/package/to-semver) to extract the real semver tags.

bfred-it:
Not all tags use semver, that’s why we stopped using that a while ago and I published my looser version comparator

HardikModha:
> this, however, might complicate things a lot when the tags can't be properly sorted (e.g. no version or non-standard version)

Yes, I agree with this. If non-standard versions are there then it would become difficult to correctly know which version is greater and which is not. However, If the library correctly follows semver then it would be great to have this. 

How about we only follow this approach when the versions strictly adhere to semver guidelines? For the rest, we may follow the current approach. Thoughts?

bfred-it:
Maybe. Perhaps you can use `semver-regex` to validate all the tags and `tiny-version-compare` (which we already use elsewhere) to sort. 

E.G.

```js 
if tags.every(tag => semver.test(tag.version))
    tags.sort(...)
```

HardikModha:
Yep, I already took a look at https://github.com/sindresorhus/semver-regex and also tried out the sorting with `tiny-version-compare`. In the initial implementation, I found, some hotfixes tags are not correctly sorted. i.e `v1.0.1-hotfix should be > v1.0.1`. Do we want `tiny-version-compare` to support `hotfix` versions?

HardikModha:
SO I tested the above approach little more and it seems to work fine as we only sort when we are sure that the versions are semver. Unpredictable comparisons are being avoided.

bfred-it:
I think that generally the dash means “pre” and I thought I add support for that a while ago, but I guess not. Can you open an issue on `tiny-version-compare`?

bfred-it:
I think it's also missing this

HardikModha:
> I think that generally, the dash means “pre” and I thought I add support for that a while ago, but I guess not. Can you open an issue on tiny-version-compare?

Opened an issue https://github.com/bfred-it/tiny-version-compare/issues/8.

> I think it's also missing this.

Yes, Encoding part is needed but `decodeURIComponent` part should not be removed. If we want to remove it from here, we can just add it at a single place where we are extracting tag from the `href`.

HardikModha:
I made the changes to sort tags wherever possible by checking whether all the versions strictly adhere to semver regex. I have tested it on the following and it seems to work great.

https://github.com/nodejs/node/tags
https://github.com/flutter/flutter/releases?after=v1.5.1 
https://github.com/avajs/ava/tags

bfred-it:
> Yes, Encoding part is needed but `decodeURIComponent` part should not be removed

Ah, yeah, it was not needed at all, the previous version was more readable

PR #2046, opened 2019-05-16
Show always the language stats

muescha:
Show always the language stats

old:
Bildschirmfoto 2019-05-16 um 22 21 23

new:
Bildschirmfoto 2019-05-16 um 22 07 53

Example URL:


sindresorhus:
Thanks for the suggestion, but we're going to pass on this. It's not essential information and you should be able to remember the colors for the most important languages anyway.

PR #2053, opened 2019-05-19
Add `sort-issues-by-update-time` to `filter-comments-by-you`

### LINKED ISSUE #2052: `filter-comments-by-you` does not add sort:updated-desc
Should be simple to add

yakov116:
Fixes #2052

### DIFF
--- a/source/content.ts
+++ b/source/content.ts
@@ -40,7 +40,7 @@ import './features/hide-useless-comments';
 import './features/navigate-pages-with-arrow-keys';
 import './features/global-discussion-list-filters';
 import './features/filter-comments-by-you';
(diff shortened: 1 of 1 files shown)


bfred-it:
This is from another feature and shouldn’t be in here, ideally

It should work automatically if the sorting feature runs after this feature

yakov116:
This
https://github.com/sindresorhus/refined-github/blob/dfee7e87714c07941119809f6674fd742524c52b/source/content.ts#L61

Should be moved here?
https://github.com/sindresorhus/refined-github/blob/dfee7e87714c07941119809f6674fd742524c52b/source/content.ts#L41-L42

bfred-it:
Probably, yes, before that

yakov116:
@bfred-it done

PR #2058, opened 2019-05-19
Fix clean-sidebar feature name

### LINKED ISSUE #2057: Disabling clean-sidebar does not disable clean-sidebar
Here are the settings I have configured:

---
image

---

This is [the PR](https://github.com/rust-lang/rustup.rs/pull/1860) sidebar with refined-github enabled:

---

image

---

This is the PR with refined-github disabled:

---

image

---

As you can see, disabling the clean-sidebar feature does not bring back anything in the sidebar. Am I disabling the wrong feature?

I can't find clean-sidebar listed in the console whether I have it in my disabled features list or not. That makes me think that maybe it is actually called something else? It's in the files list, so I'm not sure why this wouldn't be working.

busches:
Closes #2057

### DIFF
--- a/source/features/clean-sidebar.tsx
+++ b/source/features/clean-sidebar.tsx
@@ -83,7 +83,7 @@ function init(): void {
 }
 
 features.add({
-	id: 'link-to-file-in-file-history',
+	id: 'clean-sidebar',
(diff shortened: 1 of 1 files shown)


bfred-it:
Thanks!

PR #2059, opened 2019-05-19
 Add `responsive-layout` feature 

Systemcluster:
A lot of pages have different `min-width` set on the body tag (e.g. the profile page with `1280px`), this feature reduces the overall `min-width` to `768px` and adjusts some layouts below `1080px` to allow this.

`768px` has been chosen because in a lot of cases everything "just works". Going further below that would require a lot more additional layouting.

I opted to namespace the styles with the `rgh-responsive-layout` class which is added to the body via JS, so the feature can be easily disabled.

It's disabled by default on pages that are already responsive (containing the `page-responsive` class). It should therefore be future-proof in case of GitHub updating more pages to be responsive by default.

# Examples

<details>
<summary>Profile page of other user</summary>

Screenshot_2019-05-20 sindresorhus - Overview

</details>

<details>
<summary>Profile page of own user</summary>

Screenshot_2019-05-20 Systemcluster - Overview

</details>

<details>
<summary>Issue page</summary>

Screenshot_2019-05-19 mods settings - invalid load order file when using script merger · Issue #1 · Systemcluster The-Witch

</details>

<details>
<summary>Repo page</summary>

Screenshot_2019-05-19 sindresorhus refined-github

</details>


bfred-it:
Sorry but this doesn’t belong to Refined GitHub. Making a site responsive is a huge endeavor and GitHub is already working on it.

There’s probably already some userstyle for it

PR #2060, opened 2019-05-20
Fix recently-pushed-branches div position in GHE

### LINKED ISSUE #2044: Bug with pull request helper and Github Entreprise
Hello,

Better than words, here's a screenshot of the problem occuring on a Github Entreprise server resulting in breaking the header:

Yproximite_store-factory__StoreFactory_reborn

The difference with Public Github is that there is no text "Entreprise" in the header.

simonkotwicz:
Closes #2044


busches:
Looks good to me, thanks!

PR #2062, opened 2019-05-21
Improve clickability of reaction buttons in Firefox

bfred-it:
Clicking on the avatars in Firefox brings you to their profile. So I dropped the link on those avatars. Also the `a` is used in the styles so I just left it there to avoid having to change the styles as well.

### DIFF
--- a/source/features/reactions-avatars.tsx
+++ b/source/features/reactions-avatars.tsx
@@ -45,7 +45,7 @@ function add(): void {
 
 			// Without this, Firefox will follow the link instead of submitting the reaction button
 			if (!navigator.userAgent.includes('Firefox/')) {
-				(avatar as any).href = `/${participant.username}`
+				(avatar as any).href = `/${participant.username}`;
 			}
 
 			participant.container.append(avatar);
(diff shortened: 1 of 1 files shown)


source/features/reactions-avatars.tsx
```diff
@@ -37,11 +37,18 @@ function add(): void {
...
-					<img src={`/${participant.username}.png?size=${window.devicePixelRatio * 20}`}/>
+			const avatar = (
+				<a>
+					<img src={`/${participant.username}.png?size=${window.devicePixelRatio * 20}`} />
 				</a>
 			);
+
+			// Without this, Firefox will follow the link instead of submitting the reaction button
+			if (!navigator.userAgent.includes('Firefox/')) {
+				(avatar as any).href = `/${participant.username}`
```
bfred-it: Sadly `as HTMLAnchorElement` causes:

```ts
      TS2352: Conversion of type 'Element' to type 'HTMLAnchorElement' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Type 'Element' is missing the following properties from type 'HTMLAnchorElement': charset, coords, download, hreflang, and 20 more.
```

And I don't remember if we came up with a better solution for this

PR #2067, opened 2019-05-24
Show only GH Enterprise notifications on GH Enterprise

### LINKED ISSUE #2065: Unread notifications from GHE appear on Github and vice versa
Can we display notifications marked as unread from the same domain only?

simonkotwicz:
Closes #2065

### DIFF
--- a/source/features/mark-unread.tsx
+++ b/source/features/mark-unread.tsx
@@ -49,7 +49,7 @@ async function getNotifications(): Promise<Notification[]> {
 	const {unreadNotifications} = await browser.storage.local.get({
 		unreadNotifications: []
 	});
-	return unreadNotifications.filter(({url}) => url.startsWith(location.origin));
+	return unreadNotifications.filter(({url}: AnyObject) => location.hostname.endsWith(new URL(url).hostname));
 }
 
 async function setNotifications(unreadNotifications: Notification[]): Promise<void> {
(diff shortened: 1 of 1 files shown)


simonkotwicz:
Actually, I might need to modify this PR, I forgot about gist.github.com.

bfred-it:
>  I forgot about `gist.github.com`

That's fine, there are no notifications on that domain

simonkotwicz:
There is a notification icon though, do we not care about seeing unread notifications there?

bfred-it:
Good point. You can use the `URL` constructor to parse the `url` variable and match the main domain only, e.g. `github.com` instead of the whole origin.

bfred-it:
Actually that's a bad idea because `url` could be `hello.co.uk` and currentDomain` might be `wow.co.uk`, which we have to turn into `co.uk` to make my logic work.

This needs to explicitly account for `gist.github.com` instead

source/features/mark-unread.tsx
```diff
@@ -49,7 +49,7 @@ async function getNotifications(): Promise<Notification[]> {
 	const {unreadNotifications} = await browser.storage.local.get({
 		unreadNotifications: []
 	});
-	return unreadNotifications;
+	return unreadNotifications.filter(({url}: AnyObject) => location.hostname.endsWith(new URL(url).hostname));
```
bfred-it: That's a good solution

```suggestion
	return unreadNotifications.filter(({url}: Notification) => location.hostname.endsWith(new URL(url).hostname));
```

bfred-it:
This doesn't break github.com and gist.github.com, so it's a-ok for me

busches:
I never have notifications on GHE. I'm not sure if they're not fully baked yet or our instance isn't configured to use them. Either way, this looks fine to me.

PR #2070, opened 2019-05-24
Lint CSS files recursively in all sub-directories

notlmn:
`stylelint` wasn't linting CSS files under `source/features` sub-directory all along because the build script included a glob pattern to target files only under single level with `source/*.css`.

This PR modifies the glob pattern to `source/**/*.css` to target all CSS files recursively, and also lints the files (meanwhile disabling rules at some points).

## Test
First observed when https://github.com/sindresorhus/refined-github/pull/1989/commits/51a5be1e3bdbcf2bf7da7d0528187bd3df4dc217 (with indirect changes from https://github.com/sindresorhus/refined-github/pull/1989/commits/3666b19ecc2b33f195100eff405868ffd9588610) wasn't failing on Travis, but my editor was complaining as intended. Manually checked if the versions were the same everywhere, later found the bug in `package.json`.


source/features/ci-link.css
```diff
@@ -1,7 +1,7 @@
 .rgh-ci-link {
 	position: relative;
 	top: -0.1em;
-	animation: fade-in 0.2s;
+	animation: fade-in 0.2s; /* stylelint-disable-line no-unknown-animations */
```
bfred-it: Any way to have stylelint understand this is in a different file?

notlmn: image

It looks like `stylelint` cannot access references from another file and also there's no way to specify that one exits.

But it also looks like `stylelint` ignores values if placed inside a CSS custom property like:

``` css
.follower-badge {
	--animation-name: fade-in;
	animation: var(--animation-name) 0.2s; /* No errors */
}
```

What do you think about this?

bfred-it: I think it’s easier if we disable the rule altogether since we could technically also use GitHub’s animations

sindresorhus: I've removed the rule from `stylelint-config-xo` now. Just upgrade.

Also see https://github.com/stylelint/stylelint/issues/4088

package.json
```diff
@@ -31,7 +31,7 @@
 		"insert-text-textarea": "^1.0.3",
 		"intervalometer": "^1.0.5",
 		"linkify-issues": "^1.3.0",
-		"linkify-urls": "^2.2.0",
+		"linkify-urls": "^3.1.0",
```
bfred-it: This breaks Firefox, needs to be restored:

```suggestion
		"linkify-urls": "^2.2.0",
```

package.json
```diff
@@ -41,7 +41,6 @@
 		"type-fest": "^0.4.1",
 		"webext-domain-permission-toggle": "^0.1.0",
 		"webext-dynamic-content-scripts": "^5.0.2-0",
-		"webext-options-sync": "^0.15.5",
```
bfred-it: Uhm?

notlmn: I'm having a bad day manually editing `package-lock.json` file to not include updates to other packages. Because now we have new JS linting error because `eslint-config-xo-typescript` auto updates inside lock file even when I'm only updating `stylelint-config-xo`.

notlmn:
The build fails because of inconsistencies between `eslint-config-xo-typescript` and `select-dom`s Typescript definitions.

```
Argument of type 'HTMLElement | null' is not assignable to parameter of type 'ParentNode | Window | undefined'.
  Type 'null' is not assignable to type 'ParentNode | Window | undefined'.ts(2345)
```

TS is thinking that the assertions are not necessary because it thinks that `null` and `undefined` are equal. If I remove the assertions, I get the error above.

I have no intentions of digging into `package-lock.json` to manually reset the config dependency to reset back its version. Sorry.

bfred-it:
> have no intentions of digging into `package-lock.json`

Nor do you have to. You just have to undo any changes you brought to `package-lock` and just run `npm i stylelint-config-xo@latest`.

bfred-it:
Honestly it's maddening that restoring sanity still requires deleting `node_modules`.

Issue #1137, opened 2018-03-05
Refined GitHub updates


bfred-it:
# Welcome to Refined GitHub! <img src="https://raw.githubusercontent.com/sindresorhus/refined-github/master/source/icon.png" width="32">

You can subscribe to this thread to receive updates regarding new features added to the extension. Click the `Subscribe` button in the sidebar!

## Contributing

Suggestions and pull requests are highly encouraged! Have a look at the [open issues](https://github.com/sindresorhus/refined-github/issues?q=is%3Aissue+is%3Aopen+label%3A%22help+wanted%22+sort%3Areactions-%2B1-desc), especially [the easy ones](https://github.com/sindresorhus/refined-github/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22+sort%3Areactions-%2B1-desc). Check out [contributing.md](https://github.com/sindresorhus/refined-github/edit/master/contributing.md) for more info!

bfred-it:
https://github.com/sindresorhus/refined-github/issues/818

bfred-it:
- [Automatically closes modals when they’re no longer visible](https://user-images.githubusercontent.com/1402241/37022353-531c676e-2155-11e8-96cc-80d934bb22e0.gif)
- [Shows all keyboard shortcuts in the help modal](https://user-images.githubusercontent.com/29176678/36999174-9f07d33e-20bf-11e8-83e3-b3a9908a4b5f.png)  *(<kbd>?</kbd> hotkey)*
- [Adds links to the default branch and to the latest semver release](https://user-images.githubusercontent.com/1402241/36961980-389bd2b0-2080-11e8-8bff-bb8af52e7128.png)
- [Adds `Commented` button to Issues/Pull Requests page](https://user-images.githubusercontent.com/8295888/36827126-8bfc79c4-1d37-11e8-8754-992968b082be.png)
- [Highlight closing PRs in open issues and PRs](https://user-images.githubusercontent.com/1402241/37037746-8b8eac8a-2185-11e8-94f6-4d50a9c8a152.png)
- [Show latest tag in `latest release button`](https://user-images.githubusercontent.com/1402241/36968396-f9115420-2094-11e8-8bd6-ddc1748a891d.png)

bfred-it:
- [Open multiple issues at once in your repos.](https://user-images.githubusercontent.com/1402241/38084752-4820b0d8-3378-11e8-868c-a1582b16f915.gif)
[...]

PR #2073, opened 2019-05-25
Add `show-whitespaces` feature

### LINKED ISSUE #588: Option to show whitespace characters
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/refined-github/issues/588)

It'd be nice if we could see whitespace characters like in an IDE/code editor so it's easier to see things like tabs vs spaces, trailing spaces, etc.

> There is a $80.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/sindresorhus/refined-github/issues/588).

notlmn:
## Epilogue

Before explaining how this feature works, I'd like to outline all the things I tried to implement this feature.

### 1. Custom Font

Just like @bfred-it in https://github.com/sindresorhus/refined-github/issues/588#issuecomment-324617316, I tried to create a custom font using FontForge adding glyps to the codepoints "Space" (`U+0020`), "New Line" (`U+000A`), and "Horizontal Tabulation" (`U+0009`).

As as expected it failed, the tab glyph was substitued by space glyph, and new line glyph was outright ignored. This looks like that the visual for control characters is handled by the application parsing the text and not by the font being used. That may be the reason no font in the world contains any glyph information placed in these codepoints, so the font-fallback ends up at the application handling that code point.

### 2. Font Ligatures

I have a feeling that this can be solved using font ligatures, providing glyphs information for how these whitespace characters interact with codepoints in the Basic Multilingual Plane (BMP).

I have no way to test this, I don't even know if this is possible, end of story.

### 3. Replacing Characters

The characters space and tab can be easily replaced with any of the codepoint in Unicode's [Private Use Area](https://unicode-table.com/en/blocks/private-use-area) plane. But as mentioned above, we loose semantics for control characters.

## This PR

Closes #588.

With the above tests in consideration, I don't know if this feature can be implemented in any other way except for doing DOM manipulations. If there is I'm all ears.

This PR is a lot similar to #1989, doing a heck of lot DOM manipulations. At the start I had concerns about performance, but as it seems this feature does look a bit light that #1989.

Below are the list of characters and their substitution characters for reference.

### White-space Characters

| Entity | Replacement character | Unicode name | Unicode number | HTML-code |
| :-: | :-: | :-: | :-: | :-: |
| Space | `•` | "Bullet" | `U+2022` | `&#8226;` |
| Tab | `→` | "Rightwards Arrow" | `U+2192` | `&#8594;` |
| New Line (Disabled) | `¬` | "Not Sign" | `U+00AC` | `&#172;` |

> **Note**: The part to enable "New Line" is disabled for now, as it has unusual behavior while displayed on empty lines (where it is the only child). And also because how it interacts when there is no new-line at the end of file. Yes, there is commented code committed to Git.

## Testing this PR

- **Blob file**: https://github.com/sindresorhus/refined-github/blob/master/source/features/indented-code-wrapping.tsx.
- **PR files**: https://github.com/sindresorhus/refined-github/pull/1989/files.
- **PR conversation**: https://github.com/sindresorhus/refined-github/pull/1881, https://github.com/parcel-bundler/parcel/pull/2967.
- **Single commit**: https://github.com/sindresorhus/refined-github/commit/3666b19ecc2b33f195100eff405868ffd9588610.
- **Compare view**: https://github.com/sindresorhus/refined-github/compare/master...notlmn:show-whitespaces.
- **Gist**: https://gist.github.com/notlmn/fe7f722971dea345379850bd64516fd2 (with mixed indentation and trailing white-spaces).
- **Embedded Code**: https://github.com/sindresorhus/refined-github/blob/2c4e7a851784d28c179d474c5a459496b8942163/source/features/clean-discussions.css#L10-L15

Also works on all lazy-loaded and demand-loaded diff and file views.

## Possible Future Additions

- Show white-space character in code blocks inside comments.
  This is not a part of the current PR, as code blocks in comments are not separated into lines. Will see what I can do later.

- Display every "invisible" (non-printable) character outside BMP using some method.
  Generalizing this feature to `show-invisibles`. Can be used especially to detect file corruptions. But this might be heavy on the browser (way too many DOM nodes).

- To show or hide white-space characters per file.

## Screenshots

image


source/features/show-whitespaces.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+// Process a single line for each frame loop
+function loop(): void {
+	const line = queue.shift();
+
+	if (line === undefined) {
+		return;
+	}
+
+	const iterator = document.createNodeIterator(line, NodeFilter.SHOW_TEXT, {
```
bfred-it: I think we already have `get-text-nodes` for this

source/features/show-whitespaces.tsx
```diff
@@ -0,0 +1,99 @@
...
+features.add({
+	id: 'show-whitespaces',
+	description: 'Show whitespace characters in diffs',
+	include: [
+		features.isSingleFile,
+		features.isPRFiles,
+		features.isCommit,
+		features.isPRConversation,
+		features.isGist,
+		features.isCompare
```
bfred-it: Try the new `features.hasCode` instead of all of these.

notlmn:
One problem I see is that embedded code has empty spaces add to the start and end of each line.

image

A problem from GitHub's end, messes with copy and pasting code.

/cc @lukehefson @shayfrendt

source/features/show-whitespaces.tsx
```diff
@@ -0,0 +1,99 @@
...
+		'table.js-file-line-container:not(.rgh-showing-whitespace)', // Single blob file, and gist
+		'.file table.diff-table:not(.rgh-showing-whitespace)', // Split and unified diffs
+		'.file table.d-table:not(.rgh-showing-whitespace)' // "Suggested changes" in PRs
+	].join());
+
+	for (const table of tables) {
+		table.classList.add('rgh-showing-whitespace');
+
+		for (const line of select.all('.blob-code-inner', table)) {
+			queue.push(line);
```
bfred-it: Instead of creating a queue, you can just parse each line directly for now. e.g. `loop(line)`

source/features/show-whitespaces.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+			(textNode as Element).replaceWith(fragment);
+		}
+	}
+
+	// Add a new-line character at the end (optional)
+	// const br = line.querySelector('br');
+	// line.insertBefore(<span className="pl-ws pl-nl">&nbsp;</span>, br);
+
+	requestAnimationFrame(loop);
```
bfred-it: This is processing one line per frame, not very efficient. If you want to split up the job without blocking the rendering, process X lines at a time. I think that if you use many `setTimeout`s the rendered will do as much as possible in the current frame and then continue in the next one.

But for now this sounds like premature optimization, let's see how bad it performs first.

source/features/show-whitespaces.tsx
```diff
@@ -0,0 +1,99 @@
...
+				if (char === '\t') {
+					fragment.append(<span className="pl-ws pl-tab">{char}</span>);
+				} else if (char === ' ') {
+					fragment.append(<span className="pl-ws pl-space">{char}</span>);
+				} else {
+					fragment.append(char);
+				}
+			}
+
+			(textNode as Element).replaceWith(fragment);
```
bfred-it: What if instead of changing the dom, you just add a `data-rgh-spaces="••••"` attribute and then display them on all `.blob-code-inner` with `content: attr(data-rgh-spaces)`

source/features/show-whitespaces.tsx
notlmn: I thought we had a `hasCode` but I can't find it anymore. Even tried [searching](https://github.com/sindresorhus/refined-github/search?q=hasCode&type=Code) for it.

source/features/show-whitespaces.tsx
notlmn: How about processing one table for each frame, that should be a reasonable compromise.

source/features/show-whitespaces.tsx
bfred-it: I think it’s still in an open PR. Let’s wait for that then 😅

source/features/show-whitespaces.tsx
bfred-it: This is not heavy processing, plus if we wait for frames the code won’t run until the page is visible again

notlmn: It did become a heavy task causing the page to jank on load on very large PR like the Parcel example above. After which I moved to `rAF`.

If only we had a way to execute a function repeatedly for specific duration and schedule the next call when the main thread is free (chunking of work).

I know that `requestIdleCallback` exists but it does work a lot similar to `rAF` in this case, and also has no chunking of work.

bfred-it: If the attribute-based suggestion works it will become much lighter though

bfred-it:
Your fork has an obsolete tag, do you mind deleting it?

```sh
git tag -d hotfix
git push --delete origin hotfix
```

source/features/show-whitespaces.tsx
notlmn: Done, please review at your convenience.

notlmn:
With the latest commit, I managed to enable showing new-line characters. I can't decide this yet, but is this a bit distracting?

image

bfred-it:
On hold. Let's solve #2085 first

sindresorhus:
Is this ready? If not what's left to do?

@notlmn Can you fix the merge conflict?

notlmn:
One final thing to consider: https://github.com/sindresorhus/refined-github/pull/2073#issuecomment-495856632. Otherwise, everything is fine.

bfred-it:
> A problem from GitHub's end, messes with copy and pasting code.

Is it? I don't have trouble copy-pasting code on GitHub. Can you link to a specific troublesome code?

notlmn:
> Can you link to code with that issue?

The embedded code example in this PR's [description](https://github.com/sindresorhus/refined-github/pull/2073#issue-282199326), under the test section.

source/features/show-whitespaces.tsx
```diff
@@ -0,0 +1,96 @@
...
+import onPrFileLoad from '../libs/on-pr-file-load';
+import onNewComments from '../libs/on-new-comments';
+import getTextNodes from '../libs/get-text-nodes';
+
+function showWhiteSpacesOn(line: Element): void {
+	const textNodes = getTextNodes(line);
+
+	for (const textNode of textNodes) {
+		const nodeValue = textNode.textContent!;
+		if (nodeValue.length === 0) {
```
bfred-it: I'm looking for way to speed up parsing. Can this skip nodes without whitespace? **Perhaps** `.contains` is faster than `/[ \t]/.test(nodeValue)`

```suggestion
		if (nodeValue.length === 0 || !(nodeValue.contains(' ') || nodeValue.contains('\t'))) {
```

source/features/show-whitespaces.css
```diff
@@ -0,0 +1,23 @@
+.pl-ws {
```
bfred-it: If these are our own classes, they should start with `rgh` to make it clear.

If GitHub also uses these, ignore this comment

source/features/show-whitespaces.tsx
```diff
@@ -0,0 +1,96 @@
+import './show-whitespaces.css';
```
bfred-it: Feature name:
```suggestion
import './show-whitespace.css';
```

source/features/show-whitespaces.tsx
```diff
@@ -0,0 +1,96 @@
...
+import features from '../libs/features';
+import onPrFileLoad from '../libs/on-pr-file-load';
+import onNewComments from '../libs/on-new-comments';
+import getTextNodes from '../libs/get-text-nodes';
+
+function showWhiteSpacesOn(line: Element): void {
+	const textNodes = getTextNodes(line);
+
+	for (const textNode of textNodes) {
+		const nodeValue = textNode.textContent!;
```
bfred-it: Does this work? `nodeValue` isn't particularly helpful, so perhaps we can just use `textContent` this way:
```suggestion
		const {textContent} = textNode;
```

source/features/show-whitespaces.tsx
```diff
@@ -0,0 +1,96 @@
...
+		if (nodeValue.length === 0) {
+			continue;
+		}
+
+		const fragment = document.createDocumentFragment();
+
+		let lastEncounteredCharType;
+		let charType: 'space' | 'tab' | 'other';
+		let node;
```
bfred-it:
```suggestion
		let node = 'other';
```

This lets you shorten a following condition, I think

source/features/show-whitespaces.tsx
```diff
@@ -0,0 +1,96 @@
...
+		for (const char of nodeValue) {
+			if (char === ' ') {
+				charType = 'space';
+			} else if (char === '\t') {
+				charType = 'tab';
+			} else {
+				charType = 'other';
+			}
+
+			if ((lastEncounteredCharType && lastEncounteredCharType === charType) && node) {
```
bfred-it: As a consequence of the previous suggestion, I think you can now shorten this to:

```suggestion
			if (node && lastEncounteredCharType === charType) {
```

source/features/show-whitespaces.tsx
```diff
@@ -0,0 +1,96 @@
+import './show-whitespaces.css';
+import React from 'dom-chef';
+import select from 'select-dom';
+import features from '../libs/features';
+import onPrFileLoad from '../libs/on-pr-file-load';
+import onNewComments from '../libs/on-new-comments';
+import getTextNodes from '../libs/get-text-nodes';
+
+function showWhiteSpacesOn(line: Element): void {
```
bfred-it: Could you add a test for this function? Nothing too fancy, just ensure that it works on 4 lines like:


```
noIndent(' ' + ' ');
 spaceIndent('   double   ');
  doubleSpace;
	// tab
		// double tab
	 	// tab space tab
```

bfred-it:
> The embedded code example in this PR's [description](https://github.com/sindresorhus/refined-github/pull/2073#issue-282199326)

Indeed. This might make the bug more visible and some might open an issue, so we'll have to be ready to defer bug reports to support@github.com

source/features/show-whitespaces.tsx
```diff
@@ -0,0 +1,96 @@
...
+					fragment.append(node);
+				}
+
+				if (charType === 'space') {
+					node = <span className="pl-ws pl-space" data-rgh-spaces="·">{char}</span>;
+				} else if (charType === 'tab') {
+					node = <span className="pl-ws pl-tab" data-rgh-tabs="→">{char}</span>;
+				} else {
+					node = <>{char}</>;
+				}
```
bfred-it: I think that if you put this here, you can drop it from both places (line 42 and line 58):

```js
if (node) {
	fragment.append(node);
}
```

source/features/show-whitespaces.tsx
```diff
@@ -0,0 +1,96 @@
...
+		'table.js-file-line-container:not(.rgh-showing-whitespace)', // Single blob file, and gist
+		'.file table.diff-table:not(.rgh-showing-whitespace)', // Split and unified diffs
+		'.file table.d-table:not(.rgh-showing-whitespace)' // "Suggested changes" in PRs
+	].join());
+
+	for (const table of tables) {
+		table.classList.add('rgh-showing-whitespace');
+
+		for (const line of select.all('.blob-code-inner', table)) {
+			showWhiteSpacesOn(line);
```
bfred-it: If you add `await true // disable-eslint` here it should give the browser some breathing room to paint frames whenever needed.

This is particularly a problem on the occasional huge page like: https://github.com/Microsoft/TypeScript/blob/master/lib/lib.dom.d.ts

source/features/show-whitespaces.tsx
notlmn: Tried implementing tests (with various hacks for types inconsistency, and handling CSS imports), but stuck at this error.

```
  Uncaught exception in test/show-whitespace.ts

  /mnt/d/dev/git/sindresorhus/refined-github/source/options-storage.ts:20

   19: const featureOptions: Partial<RGHOptions> = {};
   20: for (const feature of __featuresList__) {
   21:   featureOptions[`feature:${feature}`] = true;

  ReferenceError: __featuresList__ is not defined
```

source/features/show-whitespaces.tsx
notlmn: `<>{char}</>` creates a `#document-fragment` that cannot be modified once appended to the DOM. The append operation has to happen only on the next iteration when the current fragment is filled with required characters (hence the additional check after the end of loop).

source/features/show-whitespaces.tsx
notlmn: Can you test https://github.com/sindresorhus/refined-github/pull/2073/commits/78be55684163bf1b23457814ee35f7a927afa424 for me and see if it still blocks main thread.

bfred-it: Strangely it doesn't work. I think I had tried `while(true){await Promise.resolve()}` in the console and it seemed to work but it doesn't anymore.

Adding this in the loop works but it's crazy slow, of course:

```js
await new Promise(resolve => setTimeout(resolve));
```

Perhaps we should only await it every 100 lines or so

```js
if (i % 100 === 0) {
  await new Promise(resolve => setTimeout(resolve));
}
```

bfred-it:
(tested https://github.com/sindresorhus/refined-github/pull/2185)

source/features/show-whitespaces.tsx
bfred-it: In that case, I think you can just use `else { fragment.append(char) }`, right?

notlmn: Using `.append()` creates a new `#text` node for each appended character, that was the reason I had to shift to using `.textContent += ` in https://github.com/sindresorhus/refined-github/commit/5d7c3d31d8c593ad98ace45f4047a70af52e6a9b#diff-0f8cf70b012a85b8fddc8744630b8e96R34.

source/features/show-whitespace.tsx
```diff
@@ -74,12 +74,45 @@ async function run(): Promise<void> {
 		table.classList.add('rgh-showing-whitespace');
 
 		for (const line of select.all('.blob-code-inner', table)) {
-			showWhiteSpacesOn(line);
-			await true;
+			yield line;
```
bfred-it: Why this? A regular await should work just as well

```js
		for (const [i, line] of select.all('.blob-code-inner', table).entries()) {
			showWhiteSpacesOn(line);
			if (i % 100) {
				await Promise(setTimeout);
			}
		}
```

notlmn: Was testing something else here 😅.

source/features/show-whitespace.tsx
```diff
@@ -0,0 +1,109 @@
...
+	for (const table of tables) {
+		table.classList.add('rgh-showing-whitespace');
+
+		for (const [i, line] of select.all('.blob-code-inner', table).entries()) {
+			showWhiteSpacesOn(line);
+
+			processedLineCount++;
+
+			if (processedLineCount > 10000) {
+				console.warn('Stopped showing whitespaces, too many lines to process!');
```
bfred-it: I don't think we need this limit

source/features/show-whitespaces.tsx
bfred-it: When you resolve this I miss it

bfred-it: If you want to test it, it'd have to be extracted into its own file because regular feature files depend on a lot of dangerous files like `features.tsx` and that webpack-replaced variable. Perhaps this function can be moved to `libs/show-whitespace.tsx`

bfred-it:
Merging this because I want to see it in the wild and it's received enough reviews for now. These 2 suggestions would still be good though:

- https://github.com/sindresorhus/refined-github/pull/2087#issuecomment-511219223
- https://github.com/sindresorhus/refined-github/pull/2073#discussion_r299819540

bfred-it:
Thank you for this feature ^_^ 
I hope it's received well and doesn't need a _disable_ toggle

jesstelford:
This is a great feature, but it's frustrating that it's enabled by default.

I'm very used to reading code on GitHub, and right now don't have time to go hunting for what suddenly changed (but I did, and ended up finding it here after scrolling through the long list of GHR options one by one) to restore things to how they were.

Perhaps this could be disabled by default, or provide a toggle at the top of code views to "show whitespace characters" (like the "show whitespace" option in PRs)?

jerone:
I wanted this feature too, but I expected to see an toggle button. 🤔

waldyrious:
I'm personally fine with the whitespaces being made visible by default, but I'd rather retain the ability to select text including whitespace. With this feature, that's only possible if I start the selection in the non-whitespace part.

Even if I'm not particularly interested in selecting the whitespace _per se_, the disabling of selections makes it hard to select text that is surrounded by whitespace, since I need to place the cursor exactly at the word border (e.g the start of the non-whitespace part of a line) for the text selection to be enabled, which is definitely cumbersome.

wearhere:
I don't necessarily mind this feature being enabled by default and yet more of a heads-up would have been nice. For instance—should https://github.com/sindresorhus/refined-github/issues/1137 have been bumped?

connorjclark:
I _do_ mind this being enabled by default ... was very unexpected and imo it's just noise.

bfred-it:
@wearhere we don’t update that page that often because, once a month usually

henrikno:
I also came here because I found this distracting while reviewing PRs. I first disabled Refined Github to check whether it was Github or Refined that added it. Then I spent 10 minutes scrolling through the options to find out which one I should disable.

Having a button and defaulting to off would be my choice https://github.com/sindresorhus/refined-github/issues/2240

michaelficarra:
I'm surprised to see internal whitespace made visible here. Usually I only see leading/trailing whitespace. Because of this, I had to disable the option. It made it too difficult to read.

notlmn:
The author of this PR is surprised to see how many people find this feature distracting. The option to view whitespace per file was already mentioned in the original post of this PR, but was merged anyway because this PR had way too many reviews already and was already almost 2 months old, and assuming there would be future possible additions, and yes the author agrees of that.

Meanwhile the author of this PR is
- tired after attending a very long hackathon
- going through a slight identity crisis, kind of
- feeling shitty everyday after returning from college recently

The author would like to mention the users that they are always welcome to disable this feature in the extension settings at any point, which is why this opiniated extension has added the ability to disable features in the first place because too many people had specific requirements.

The author would like to get back to this issue as soon as they can and would also like to link to this [xkcd comic](https://xkcd.com/1172/) hoping that the users would calm down.

Feedback is always appreciated.

Peace 🖖.

jerone:
Appreciate [all](https://github.com/sindresorhus/refined-github/pulls?q=is%3Apr+author%3Anotlmn) your work @notlmn 👍

**Let's divert everyone to https://github.com/sindresorhus/refined-github/issues/2240.** 

When I find the time (hopefully this weekend); I'll write a toggle button. Because I love this feature, but only want it enabled when I need it (and don't want to enable/disable the whole feature and then reload the page again).

PR #2076, opened 2019-05-25
Move Security to `more-dropdown` and cleanup code

### LINKED ISSUE #2072: List of repository tabs is too long
In my case the bar has so many tabs that the last one appears on second line.

Screenshot_2019-05-24 ifm24 ifm24_7 cz

bfred-it:
Closes #2072, among other cleanups


source/features/more-dropdown.tsx
```diff
@@ -30,34 +30,38 @@ async function init(): Promise<void> {
...
 	select('.reponav-dropdown .dropdown-menu')!.append(
-		<a href={`/${repoUrl}/compare`} className="rgh-reponav-more dropdown-item" data-skip-pjax>
+		<a href={`/${repoUrl}/compare`} className="rgh-reponav-more dropdown-item">
 			{icons.darkCompare()} Compare
 		</a>,
 
 		isEnterprise() ? '' :
-			<a href={`/${repoUrl}/network/dependencies`} className="rgh-reponav-more dropdown-item rgh-dependency-graph" data-skip-pjax>
+			<a href={`/${repoUrl}/network/dependencies`} className="rgh-reponav-more dropdown-item">
 				{icons.dependency()} Dependencies
```
bfred-it: I almost dropped this altogether because of the new "Used by" counter. For actual dependencies I have https://github.com/npmhub/npmhub

source/features/more-dropdown.tsx
```diff
@@ -29,35 +29,38 @@ async function init(): Promise<void> {
...
+		<a href={`/${repoUrl}/commits`} className="rgh-reponav-more dropdown-item">
 			{icons.history()} Commits
 		</a>,
 
-		<a href={`/${repoUrl}/branches`} className="rgh-reponav-more dropdown-item" data-skip-pjax>
+		<a href={`/${repoUrl}/branches`} className="rgh-reponav-more dropdown-item">
 			{icons.branch()} Branches
 		</a>,
+
+		<a href={insightsTab.href} className="rgh-reponav-more dropdown-item">
```
busches: Can we only add the link if it exists? Confident GHE doesn't have this or it's configuration based on allowing it.

source/features/more-dropdown.tsx
```diff
@@ -30,34 +30,38 @@ async function init(): Promise<void> {
...
+			{securityTab.firstElementChild} Security
+		</a>,
 	);
 
-	// Remove native Insights tab
-	const insightsTab = select('[data-selected-links~="pulse"]');
-	if (insightsTab) {
-		insightsTab.remove();
-	}
+	insightsTab.remove();
```
busches: Won't this throw if the insights/security tab doesn't exist?

source/features/more-dropdown.tsx
bfred-it: We've been showing it without conditions for the longest time, but now I guess if it's not there the whole feature will fail.

source/features/more-dropdown.tsx
bfred-it: I didn't think that'd ever happen. On GitHub.com it's always there I think?

sindresorhus:
When I go to https://github.com/avajs/ava/pulse and click the "More", a "Pulse" menu item is shown without an icon.

bfred-it:
Good catch, it was selecting the links from the mobile navigation, which is only available on mobile pages.

PR #2079, opened 2019-05-25
Remove special styling for unread notifications indicator

simonkotwicz:
Now that Github supports marking notifications as unread, that bit of CSS only changes the colour of the notifications indicator for notifications that are marked as unread by refined-github, not Github. I think it probably makes sense to be consistent with Github's own unread notifications styling.


bfred-it:
They’re still not the same, GitHub can mark notifications as unread but RGH can mark any issue/PR as such before ever entering the notification box. 

Plus the color tells you that the notifications are displayed only on your computer rather than being part of GitHub itself.

PR #2086, opened 2019-05-27
Show drag and drop area on release page

lautis:
The visible drag and drop area is the only method of attaching files to a release. It looks like the selector meant to hide the message on discussions accidentally applies to also releases.

The drag and drop area for releases was hidden in #2068.

# Before

Screenshot 2019-05-27 at 20 45 50

# After

Screenshot 2019-05-27 at 20 44 47

# Test

As an admin of refined-github, you can check the behaviour on https://github.com/sindresorhus/refined-github/releases/new?tag=19.5.26.1924

### DIFF
--- a/source/features/clean-rich-text-editor.css
+++ b/source/features/clean-rich-text-editor.css
@@ -6,11 +6,12 @@ markdown-toolbar > :nth-last-child(4) { /* H1, B, I */
 }
 
 /* Remove upload message on comment box */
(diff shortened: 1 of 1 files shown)


bfred-it:
👍

bfred-it:
This is great, thanks! Bugs like these go unnoticed and just cause pain. Thanks for the PR 🍦

PR #1896, opened 2019-03-30
Add `tags-dropdown` feature

HardikModha:
1. Here is the list of URLs that can be used to test the functionality.

- https://github.com/facebook/react/releases (Releases page)
- https://github.com/facebook/react/releases/tag/v16.8.1 (Release page of a particular tag)
- https://github.com/sindresorhus/refined-github/releases (Releases page with no releases)

2. It Fixes #1895 and closes #1895.

Note: The feature uses the GitHub API to fetch the releases for the project. If the personal token is available then it uses api v4 (GraphQL) otherwise it uses the api v3 (REST).

Demo:

ezgif com-video-to-gif

PR #1694, opened 2019-01-04
Move loading logic into each feature

bfred-it:
- Move loading logic (`when`, `where`) near the feature's code instead of being listed in 3 separate functions (`init`, `onDomReady`, `ajaxedPagesHandler`) in a different file (`content.js`)
- Avoids the messy content.js list of imports/uselessFunctionNames by halving the file length
- Enables us to actually list all the features in Options (not part of this PR) since they are _all_ added in `feature.add('feature-name')` on load

~~This is just a proof of concept so only a feature has been implemented.~~

PR #2087, opened 2019-05-28
Restore `toggle-everything-with-alt` feature

### LINKED ISSUE #825: Add support for loading all large diffs at once
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/refined-github/issues/825)

by default larges diffs are not shown. If a PR contains many large diffs, one has to go over them one by one and there is a risk to miss something in a code review.

It would be great to have a button at the top "load all diffs".

> There is a $40.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/sindresorhus/refined-github/issues/825).

### LINKED ISSUE #1729: Restore `toggle-all-things-with-alt` feature
It was removed in https://github.com/sindresorhus/refined-github/pull/1524 as requested in https://github.com/sindresorhus/refined-github/issues/1486, but our implementation touched more than just diffs.

Some other things that can be toggled in batch:

- resolved comments (not just reviews) e.g.: https://github.com/sindresorhus/refined-github/pull/1694#issuecomment-451926782
- review comments per-file ("Show comments" checkbox): https://github.com/sindresorhus/refined-github/pull/1896/files
- large collapsed diffs: https://github.com/sindresorhus/refined-github/issues/825

notlmn:
Closes #1729.
Implements and closes #825.

The feature now allows collapsing/expanding:
- Hidden comments: https://github.com/sindresorhus/refined-github/pull/1694.
- Review comments in PR files ("Show/Hide conversation"): https://github.com/sindresorhus/refined-github/pull/1896/files.
~~- Review comments per-file ("Show/Hide comments"): https://github.com/sindresorhus/refined-github/pull/1896/files.~~
- Loading large diffs at once: https://github.com/parcel-bundler/parcel/pull/2967/files.


source/features/toggle-all-things-with-alt.tsx
```diff
@@ -0,0 +1,70 @@
+import select from 'select-dom';
+import delegate, {DelegateEvent} from 'delegate-it';
+import features from '../libs/features';
+
+function getSimilarItems(item: Element): HTMLButtonElement[] | HTMLLabelElement[] | HTMLElement[] {
+	// Collapsed comments in PR conversations and files
+	if (item.matches('summary')) {
+		if ((item.parentElement! as HTMLDetailsElement).open === true) {
```
bfred-it:
```suggestion
		if ((item.parentElement as HTMLDetailsElement).open) {
```

source/features/toggle-all-things-with-alt.tsx
```diff
@@ -0,0 +1,70 @@
...
+			return inputs.filter(input => !input.checked).map(input => input.labels![0]);
+		}
+
+		return inputs.filter(input => input.checked).map(input => input.labels![0]);
+	}
+
+	return [];
+}
+
+function handleEvent(event: DelegateEvent<MouseEvent, HTMLButtonElement | HTMLLabelElement | HTMLElement>): void {
```
bfred-it:
```suggestion
function handleEvent(event: DelegateEvent<MouseEvent, HTMLElement>): void {
```

source/features/toggle-all-things-with-alt.tsx
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	// "Load diff" buttons in PR files
+	if (item.matches('.js-diff-load')) {
+		return select.all('.js-file .js-diff-load');
+	}
+
+	// "Show comments" checkboxes
+	if (item instanceof HTMLLabelElement) {
+		const inputs: HTMLInputElement[] = select.all('.js-file .dropdown-item .js-toggle-file-notes');
```
bfred-it:
```suggestion
		const inputs = select.all<HTMLInputElement>('.js-file .dropdown-item .js-toggle-file-notes');
```

source/features/toggle-all-things-with-alt.tsx
```diff
@@ -0,0 +1,70 @@
...
+		}
+
+		return inputs.filter(input => input.checked).map(input => input.labels![0]);
+	}
+
+	return [];
+}
+
+function handleEvent(event: DelegateEvent<MouseEvent, HTMLElement>): void {
+	if (event.altKey) {
```
bfred-it: Clearer intent:

```suggestion
	if (!event.altKey) {
		return;
	}
```

bfred-it:
In Firefox, clicking any "Show comment" in https://github.com/sindresorhus/refined-github/pull/1694 freezes the browser for 4 seconds in my case, but not in Chrome. Can you look into it?

notlmn:
> 3\. The suggested link for large diffs only has one large diff on the page: [parcel-bundler/parcel/pull/2967/files](https://github.com/parcel-bundler/parcel/pull/2967/files)

There are multiple of them on that page. It is the same as loading diffs for deleted files: https://github.com/sindresorhus/refined-github/pull/1524/files.

image

---------------------

> 1. In Firefox, clicking any "Show comment" in #1694 freezes the browser for 4 seconds in my case, but not in Chrome. Can you look into it?

Happens to me on Chrome too. Not a bug, just main thread jank for clicking on 90 elements all at once. (Any solution?)

---------------------

> 2\. Also review comments don't work in Chrome and Firefox, here [sindresorhus/refined-github/pull/1896/files](https://github.com/sindresorhus/refined-github/pull/1896/files)

Missed that, added it in.

bfred-it:
> Happens to me on Chrome too. Not a bug, just main thread jank for clicking on 90 elements all at once. (Any solution?)

It's worth investigating if it's the click or the selection being slow

notlmn:
It's the clicks, events being triggered recursively. But the functions early return anyway.

image

bfred-it:
Should we space them out / delay them a bit? We can't cause multi-second freezes. In this case I'd probably click on them no more than 3 per second, also because it's gonna take longer than that to open + see them anyway.

Additionally, perhaps, it could be limited to 10 files at a time, but this code might be a bit ugly and maybe not worth it.

bfred-it:
Also can you rename this to `toggle-everything-with-alt`?

source/features/toggle-everything-with-alt.tsx
```diff
@@ -0,0 +1,83 @@
...
+	delegate('.repository-content', '.minimized-comment details summary', 'click', handleEvent);
+	delegate('.repository-content', '.js-file .js-diff-load', 'click', handleEvent);
+	delegate('.repository-content', '.js-file .dropdown-menu label.dropdown-item:first-child', 'click', handleEvent);
+	delegate('.repository-content', '.js-file .js-resolvable-thread-toggler', 'click', handleEvent);
+}
+
+features.add({
+	id: 'toggle-everything-with-alt',
+	load: features.onAjaxedPages,
+	description: 'Toggle all similar items while holding `alt`',
```
bfred-it: I think we can list all supported "items" both here and in the readme, e.g. `Toggle all items when alt-clicking minimized commments, deferred diffs, hidden comments in the PR Files tab"

notlmn: How about
> Toggle all similar items (minimized comments, deferred diffs, etc) while holding `alt`

bfred-it:
Can you add the same mechanism here? https://github.com/sindresorhus/refined-github/pull/2073/files#diff-bfeb5de166cf585c43d9150166877a54R89

Except spaced every 3 and using requestAnimationFrame instead

notlmn:
> Can you add the same mechanism here? [sindresorhus/refined-github/pull/2073/files#diff-bfeb5de166cf585c43d9150166877a54R89](https://github.com/sindresorhus/refined-github/pull/2073/files#diff-bfeb5de166cf585c43d9150166877a54R89)

The handlers are delegated over to `.repository-content`, so this works even for ajaxed file diffs in a PR even without using that util.

---------------

> Except spaced every 3 and using requestAnimationFrame instead

I don't seem to follow you on this.

If you are talking about https://github.com/sindresorhus/refined-github/pull/2087#issuecomment-498793479, then that's not possible, as the possible solution in #2073 [mentioned](https://github.com/sindresorhus/refined-github/pull/2073#discussion_r300480974) would not work here. Spacing out callback to let the user scroll will cause the last scroll position to be invalidated and when the callback is eventually fired in rAF, the page scrolls to random positions for each callback.

Having a jank is pretty reasonable IMHO (instead of handling scroll position changes for random file load jumps).

source/features/toggle-everything-with-alt.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+	const clickedItem = event.delegateTarget;
+	const viewportOffset = clickedItem.parentElement!.getBoundingClientRect().top;
+	const similarItems = getSimilarItems(clickedItem);
+
+	for (const item of similarItems) {
+		if (item !== clickedItem) {
+			item.click();
+		}
+	}
```
bfred-it: I don't think we should ever cause multi-second locks, it's just bad, especially because it will look like the page is unresponsive and the user might alt-click it again and... now the queue is longer.

---

It's a little more complicated because each batch would have to be surrounded by a `get scroll` and `set scroll`, but the `anchorScroll` we have in #2146 should help. 

```js
const chunked = chunkArray(similarItems.filter(item => item !== clickedItem), 3);
for (const chunk of chunked) {
	await anchorScroll(() => {
		for (const item of chunked) {
			item.click();
		}
	});
}
```

Or maybe an easier way to just process 3 at a time with a regular for loop

bfred-it:
There might have been some miscommunication. My message was referring to the `await promise setTimeout` mechanism (https://github.com/sindresorhus/refined-github/commit/652bb9e38c8607cf41e5b1934e3e225a02e4f844#diff-bfeb5de166cf585c43d9150166877a54R90) but the link I had posted pointed to the wrong line (probably after a commit its line changed)

source/features/toggle-everything-with-alt.tsx
notlmn: > It's a little more complicated because each batch would have to be surrounded by a `get scroll` and `set scroll`

Yup, that's the problem. But its more like guarding each call to `.click()` and its respective layout change, while handling scroll position changes.

> but the `anchorScroll` we have in #2146 should help.

That doesn't seem to help much too, the scroll position ends up being at some random positions after triggering the clicks.

I've tried handling duplicate events, any events emitted on repeated clicks, chunking callbacks, even plain promises, and all of them have the same symptom, the scroll position ends up being at random positions (especially for chunked triggers and allowing user scrolling without jank).

bfred-it: > the scroll position ends up being at some random positions after triggering the clicks

Indeed, our function only works with sync changes but in this case the fragments might be loaded asynchronously so our "reset position on rAF" was **before** the fragment loaded. I dropped it in 867752c2

Anyway it looks like the browser's scroll anchoring already takes care of it if the elements are completely outside the viewport.

bfred-it:
I implemented the mechanism I mentioned and but now it only _awaits_ if the operation is taking a while. It might be worth implementing the same thing in #2073 as well instead of having the arbitrary 100 lines limit.

bfred-it:
In the meanwhile I think GH changed the selectors because this doesn't work anymore:

> Hidden comments (test: #1694

And this one is not in your test list at all:

> review comments per-file ("Show comments" checkbox): [sindresorhus/refined-github/pull/1896/files](https://github.com/sindresorhus/refined-github/pull/1896/files)

Also feel free to restore c2c44b1 if necessary anywhere (but the re-set needs to happen before/inside the `await` **and** the new `viewportOffset` needs to be _gotten_ after it)

notlmn:
With the latest commit, everything should be fixed now.

I think GitHub changed how they handle clicks on dropdown items, when a click is called on them, the `summary` element is focused, so I had to restore the focus back to the original dropdown.

This brings us to a situation where we don't see multi-second freezes but the focus keeps changes to new elements, scrolling the page and then restoring focus and scroll position takes us back to the original position. This is **very weird** behavior, but hey that's how GH implemented it.

And yes, I restored https://github.com/sindresorhus/refined-github/commit/c2c44b10e42ff2ccbbf94cf66fef5bf7ef9f6f2e, we need it to make sure the scroll position never changes.

--------------

> And this one is not in your test list at all:
> 
> > review comments per-file ("Show comments" checkbox): [sindresorhus/refined-github/pull/1896/files](https://github.com/sindresorhus/refined-github/pull/1896/files)

It _is_ listed in the OP, not the same wording, but the same test link.

bfred-it:
> It _is_ listed in the OP, not the same wording, but the same test link.

It's different: `✓ Show comments` completely hides the comment. You're only referring to the `Show conversation` on resolved conversations.

Alt+clicking `Show comments` does not work in the latest version.

source/features/toggle-everything-with-alt.tsx
```diff
@@ -0,0 +1,104 @@
...
+		}
+
+		return inputs.filter(input => !input.checked).map(input => input.labels![0]);
+	}
+
+	return [];
+}
+
+async function handleEvent(event: DelegateEvent<MouseEvent, HTMLElement>): Promise<void> {
+	if (!event.altKey || !event.isTrusted || event.srcElement instanceof HTMLInputElement) {
```
bfred-it: `srcElement` is the same as `target` 
https://developer.mozilla.org/en-US/docs/Web/API/Event/srcElement


```suggestion
	if (!event.altKey || !event.isTrusted || event.target instanceof HTMLInputElement) {
```

source/features/toggle-everything-with-alt.tsx
```diff
@@ -0,0 +1,104 @@
...
+		anchorElement = clickedItem.parentElement!;
+	}
+
+	const viewportOffset = anchorElement.getBoundingClientRect().top;
+
+	let timeKeeper = Date.now();
+	for (const item of getSimilarItems(clickedItem)) {
+		if (item !== clickedItem) {
+			// Avoid holding the thread for much longer than 50ms
+			if (Date.now() - timeKeeper > 50) {
```
bfred-it: We can probably bring this number to 300ms.

1/3 second is still manageable and should overall speed up the operation

source/features/toggle-everything-with-alt.tsx
```diff
@@ -0,0 +1,104 @@
...
+				await new Promise(resolve => setTimeout(resolve));
+				timeKeeper = Date.now();
+			}
+
+			item.click();
+		}
+	}
+
+	// Scroll to original position where the click occurred after the rendering of all click events is done
+	requestAnimationFrame(() => {
```
bfred-it: This code should not run if the loop was held via the `await` (easy solution: set the `isAsync` variable in the loop)

notlmn:
@fregante, the current JS profile as compared to https://github.com/sindresorhus/refined-github/pull/2087#issuecomment-498045425:

image

---------

Everything in https://github.com/parcel-bundler/parcel/pull/2967/files now toggles in less than 400ms.

source/features/toggle-everything-with-alt.tsx
```diff
@@ -0,0 +1,107 @@
...
+	delegate('.repository-content', '.minimized-comment details summary', 'click', handleEvent);
+	delegate('.repository-content', '.js-file .js-diff-load', 'click', handleEvent);
+	delegate('.repository-content', '.js-file .dropdown-menu label.dropdown-item:first-child', 'click', handleEvent);
+	delegate('.repository-content', '.js-file .js-resolvable-thread-toggler', 'click', handleEvent);
+}
+
+features.add({
+	id: __featureName__,
+	load: features.onAjaxedPages,
+	description: 'Toggle all similar items (minimized comments, deferred diffs, etc) while holding `alt`',
```
fregante:
```suggestion
	description: 'Adds a shortcut to toggle all similar items (minimized comments, deferred diffs, etc) at once: `alt` `click` on each button or checkbox.',
```

fregante:
I'm sorry, I think this feature can no longer be made of just 2 functions. In its first iteration last year, all toggles behaved similarly, now it seems that each one needs to behave differently because we keep fixing one side and breaking the other.

Example in Chrome (you can see the menus flashing):



In Firefox it's way worse as it takes longer and the scroll goes nuts.

----

How about this: let's focus on **one** feature and get that one in first, for example https://github.com/sindresorhus/refined-github/issues/825

Ensure that it works in both browsers and without long stalls, we'll merge it.

fregante:
You can leave the name as `toggle-everything-with-alt` at the moment and try to use `anchorScroll` if possible

fregante:
I have good news for this:

> * Review comments per-file ("Show/Hide comments"): [sindresorhus/refined-github/pull/1896/files](https://github.com/sindresorhus/refined-github/pull/1896/files).

Toggle can be simulated via:

```js
document.dispatchEvent(new KeyboardEvent('keydown', {code: 'KeyI', key: 'i'}))
```

The `i` key will toggle all File comments, it's a native GitHub shortcut. Hopefully this can be integrated easily (in the next PR). Goes to show how each part has its own way to do things.

fregante:
Ping on this if you have time

notlmn:
> Ping on this if you have time

Sorry I couldn't reply on this, I've been away from keyboard lately because of a lot of distractions. I'll try to get back to handling things as fast as I can.

notlmn:
I thought of closing this PR and opening a new one as it was getting a bit complex with way too many reviews, but I think I found solutions to the above issues, so left it as is.

--------------

> Example in Chrome (you can see the menus flashing):
> ...
> In Firefox it's way worse as it takes longer and the scroll goes nuts.

Found the source of that bug, I was targeting the summary element deep within the hidden comment, which is also that dropdown.

---------------

> I have good news for this:
> ...
> Toggle can be simulated via:
> ...
> The `i` key will toggle all File comments, it's a native GitHub shortcut. Hopefully this can be integrated easily (in the next PR). Goes to show how each part has its own way to do things.

That still has the issues mentioned above in https://github.com/sindresorhus/refined-github/pull/2087#issuecomment-513573393, the focus keeps shifting to the last file in the list, adding/removing is still the better way to go.

fregante:
Tested in Firefox.


>     * Hidden comments: #1694

This still freezes the browser for 1s+ to toggle all of those... but that's passable.



> Review comments in PR files ("Show/Hide conversation"): [sindresorhus/refined-github/pull/1896/files](https://github.com/sindresorhus/refined-github/pull/1896/files)

Seems to work correctly.

>     * Review comments per-file ("Show/Hide comments"): [sindresorhus/refined-github/pull/1896/files](https://github.com/sindresorhus/refined-github/pull/1896/files)

**Does not work.** The menu stays open as if the click was not registered at all. Previously reported in https://github.com/sindresorhus/refined-github/pull/2087#issuecomment-514136981

>     * Loading large diffs at once: [parcel-bundler/parcel/pull/2967/files](https://github.com/parcel-bundler/parcel/pull/2967/files)

Seems to work correctly.

fregante:
As said before, I was no a fan of all the `if` in those two functions, they handled too much.

I rewrote the feature trying to split it where it made sense and it's a bit shorter and readable.

However I dropped the part that wasn't working (as per my previous comment) and given how complex it was, I'd rather see it replaced by something like


```js
anchorScroll(() => {
    document.dispatchEvent(new KeyboardEvent('keydown', {code: 'KeyI', key: 'i'}))
}, clickedElement.closest('.js-file'));
```

But that can come in a secondary PR. I'll test this further in both browser and I'll merge it soon.

PR #2088, opened 2019-05-28
Fix `indented-code-wrapping` for Firefox

### LINKED ISSUE #2085: Misaligned diff view with tabs
When mixing tabs and spaces within a GitHub diff view Refined GitHub seems to screw up the indentation to make it look like there is a formatting issue when there is actually not. The issue can be seen in [1] and the raw diff in [2] where we can see there is no formatting issue. I've attached screenshots with Refined GitHub enabled and disabled.

[1] https://github.com/eclipse/omr/commit/75c09985f5d5dd6cfd7c9d1e4f4d56e9406a45c4#diff-3990e65f3237ed46d278a1a8aefbaeeaR389
[2] https://raw.githubusercontent.com/eclipse/omr/75c09985f5d5dd6cfd7c9d1e4f4d56e9406a45c4/port/common/j9nls.c

## Refined GitHub enabled
image

## Refined GitHub disabled
image

notlmn:
Quick fix that probably closes #2085. 🤞

Surprisingly, I found this fix while glazing trough the CSS spec for `text-indent` property. Thanks @bfred-it.
image

## Test
https://github.com/eclipse/omr/commit/75c09985f5d5dd6cfd7c9d1e4f4d56e9406a45c4#diff-3990e65f3237ed46d278a1a8aefbaeeaR389

### DIFF
--- a/source/features/indented-code-wrapping.css
+++ b/source/features/indented-code-wrapping.css
@@ -6,3 +6,12 @@
 	display: block;
 	line-height: 21px; /* `display: block` reduces the line height; this undoes the change */
 }
+
+/*
+ * Inline-blockify all child nodes and reset text-indentation, a fix of Firefox
+ * https://github.com/sindresorhus/refined-github/issues/2085
+ */
+.rgh-code-wrapping-enabled .blob-code-inner * {
+	display: inline-block;
+	text-indent: 0;
+}
--- a/source/features/indented-code-wrapping.css
+++ b/source/features/indented-code-wrapping.css
@@ -4,6 +4,7 @@
 
 .rgh-code-wrapping-enabled span.blob-code-inner {
 	display: block;
+	text-indent: 0; /* Default indentation for all code blocks, see Firefox fix below */
 	line-height: 21px; /* `display: block` reduces the line height; this undoes the change */
 }
 


bfred-it:
Does it also fix this? https://github.com/vatesfr/xen-orchestra/pull/3155/files#diff-37458fb9e51d193ef360403ba893e2e2R47

notlmn:
#2089 is a bug (depending on how you see it), the latest commit fixes that too.

But we might have some noticeable shifting around of code.

bfred-it:
This breaks the indentation feature for me:

notlmn:
Over the last few days, I've tried all possible ways to fix this without any side effects. Adding pseudo-elements, wrapping inside a container, and other layout changes. But as the bug seems to be in the indentation itself, we cannot stop the property from being inherited and then possibly being miscalculated.

Will close this in a day if I still can't find any solution, and in favor of https://bugzilla.mozilla.org/show_bug.cgi?id=1555903.

bfred-it:
Thanks for trying!

notlmn:
Closing this in favor of [bugzilla.mozilla.org/show_bug.cgi?id=1555903](https://bugzilla.mozilla.org/show_bug.cgi?id=1555903).

As this is a Firefox bug, doing something to fix this would result in a bug in this repo when Firefox actually fixes this.

---------------

@bfred-it I think we can continue with #2073, which at least partially solves this on Firefox. WDYT!?

bfred-it:
If #2073 completely fixes it then ok, but if it doesn't we probably have to disable `indented-code-wrapping`

notlmn:
> If #2073 completely fixes it then ok, but if it doesn't we probably have to disable `indented-code-wrapping`

Yes it does.

image

At least for the case of #2085 as far as I can verify.

notlmn:
> > If #2073 completely fixes it then ok, but if it doesn't we probably have to disable `indented-code-wrapping`
> 
> Yes it does.

Scratch that, #2073 doesn't fix #2085. Realized that that's not how tab characters are supposed to work.

#2073 now works as an independent feature.

PR #2090, opened 2019-05-28
Make sure release count is always shown

### LINKED ISSUE #2077: Releases counter is missing
Regression:

It's detected on the repo root but it's not cached from there like it should. We don't need to fetch it with the API but just remember it after visiting the root.

gavinhenderson:
Gif of Fix

Closes #2077 

The cache lib was defining its own `browser.runtime.onMessage.addListener` and there can only be one of these. The solution to this was to make the background script delegate the message to the cache lib function if necessary.

# Test
You can view this fix in action at https://github.com/mulberrysymbols/mulberry-symbols

### DIFF
--- a/source/libs/cache.ts
+++ b/source/libs/cache.ts
@@ -56,20 +56,18 @@ if (location.pathname === '/_generated_background_page.html') {
 			const [cached] = document.cookie.split('; ')
 				.filter(item => item.startsWith(key + '='));
 			if (cached) {
-        const [, value] = cached.split('=');
+				const [, value] = cached.split('=');
 				console.log('CACHE: found', key, value);
-        return JSON.parse(value)
-			} else {
-				console.log('CACHE: not found', key);
-        return
+				return JSON.parse(value);
 			}
+
+			console.log('CACHE: not found', key);
 		} else if (code === 'set-cache') {
 			console.log('CACHE: setting', key, value);
 
 			// Store as JSON to preserve data type
 			// otherwise Booleans and Numbers become strings
 			document.cookie = `${key}=${JSON.stringify(value)}; max-age=${expiration ? expiration * 3600 * 24 : ''}`;
-    }
-    return
+		}
 	});
 }
--- a/source/libs/cache.ts
+++ b/source/libs/cache.ts
@@ -45,30 +45,35 @@ export function set<TValue extends any = any>(key: string, value: TValue, expira
 }
 
 /* Accept messages in background page */
-if (!browser.runtime.getBackgroundPage) {
-	browser.runtime.onMessage.addListener((request: CacheRequest, _sender, sendResponse) => {
-		if (!request) {
-			return;
+export const handleCacheCalls = async (
+	request: CacheRequest,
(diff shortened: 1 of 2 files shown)


bfred-it:
> The cache lib was defining its own `browser.runtime.onMessage.addListener` and there can only be one of these

Yeah that's as expected, there's a condition that enables the listener only in the background page, so the file can be included both as a background page and as a content page, without modification.

The same concept is used in the simple https://github.com/bfred-it/webext-content-script-ping module

bfred-it:
> The cache lib was defining its own `browser.runtime.onMessage.addListener` and there can only be one of these

That's not right, you can attach as many listeners as you want.

source/features/raw-file-link.tsx
```diff
@@ -10,17 +10,16 @@ const createRawUrl = (href: string): string => {
 };
 
 function addRawButtons(): void {
-	for (const anchor of select.all<HTMLAnchorElement>('.file-header .file-actions > div.BtnGroup [href*=\'/blob/\']:last-child')) {
```
bfred-it: This should be a separate PR

gavinhenderson:
Sorry that extra commit was an accident, meant to open at a second PR, was hoping I could remove it before you notice. I've removed it now.

Hmmm, I assumed that you could only have one listener and that was the problem. When I was debugging the listener in the `cache.ts` wasn't being fired at all so I assumed that it was being registered first then the second listener was being registered over the top of the first one

source/libs/cache.ts
```diff
@@ -45,30 +45,35 @@ export function set<TValue extends any = any>(key: string, value: TValue, expira
 }
 
 /* Accept messages in background page */
-if (!browser.runtime.getBackgroundPage) {
```
bfred-it: I think the change required here is:

```js
if (location.pathname === '/_generated_background_page.html') {
```

because `browser.runtime.getBackgroundPage` was only truthy in the background page and this condition makes it exactly clear of what we're expecting.

source/libs/cache.ts
```diff
@@ -45,30 +45,35 @@ export function set<TValue extends any = any>(key: string, value: TValue, expira
...
-	browser.runtime.onMessage.addListener((request: CacheRequest, _sender, sendResponse) => {
```
bfred-it: Here also it would be good to replace `sendResponse` with a Promise, because what I think broke the feature was the polyfill dropping the `sendResponse` callback


Docs here: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/onMessage

> return a `Promise` from the event listener, and resolve when you have the response (or reject it in case of an error).

gavinhenderson:
@bfred-it That is those requested changes made, all works now 👍

bfred-it:
The cache keeps not working in in Firefox, but that's due to the cookie-based cache. I opened https://github.com/sindresorhus/refined-github/issues/2100 for that

gavinhenderson:
@bfred-it Thanks for walking me through this contribution 💯

Issue #2093, opened 2019-05-28
Cannot see Marketplace link in default location


drewswaycool:
I have `move-marketplace-link-to-profile-dropdown` turned off, however the link still does not appear in the header links.

<img width="976" alt="Screen Shot 2019-05-28 at 3 12 25 PM" src="https://user-images.githubusercontent.com/3419661/58512817-4029d080-815b-11e9-8c3e-6db1d792ab82.png">

yakov116:
https://github.com/sindresorhus/refined-github/blob/1ec1b32ec19d54d8c7b063ca5f5a36a29a7906d5/source/content.css#L38-L41

Needs to be move to the feature

busches:
For now you can override the case rule from above.

bfred-it:
Indeed the CSS is the way to go. I'm leaving this open for a bit if anyone wants to:

1. remove the CSS
2. rename the feature to `move-marketplace-link-to-profile-dropdown`
3. implement `elementReady` like in `trending-menu-item` to detect the header link, remove it, and add the link to the dropdown.

PR #2094, opened 2019-05-29
Let users disable `move-marketplace-link-to-profile-dropdown`

bfred-it:
Fixes https://github.com/sindresorhus/refined-github/issues/2093 and actually adds the link to the dropdown, it has been broken.


source/features/move-marketplace-link-to-profile-dropdown.tsx
```diff
@@ -1,25 +1,25 @@
...
+import domLoaded from 'dom-loaded';
 import features from '../libs/features';
 
-function init(): false | void {
-	const lastDivider = select.last('.user-nav .dropdown-divider');
-	if (!lastDivider) {
-		return false;
-	}
+async function init(): Promise<void> {
+	(await elementReady('.Header-link[href="/marketplace"]'))!.remove();
```
dshoreman: Wouldn't it be better to do this *after* adding the replacement link in case it fails? That way, if Github were to change the class (again), the Marketplace link would at least remain accessible in its "normal" location.

dshoreman:
Thanks for this PR - came here to fix the same issue but I would've ended up going a far more hacky route filtering by `.dropdown-menu.mt-2` since they removed the `.user-nav` class, so I'm glad I searched first.

To expand on the review comment above - returning `false` if/when the `.before()` fails would probably be enough to work as a fallback (assuming original link is removed *after* vs before)

Though I guess it's a bit of a moot point, since it can just be disabled entirely... That said, I forgot the Marketplace even *existed* until I started checking the Features individually in the extension settings. :man_shrugging:

source/features/move-marketplace-link-to-profile-dropdown.tsx
bfred-it: That's a good point, but we want to remove it before the first paint to avoid, the layout change on every page in the header.

dshoreman: That's fair. Would restoring the constant with the `if / return` be a way to work around that? If `select.last(...)` is stored/checked first, theoretically the rest can remain the same with the main navbar link still being removed before `domLoaded`.

bfred-it: There's a way to do what you're saying, but the way I set it up now will show an error in the console and it'll push people to fix the selector sooner. Before this PR, the error was swallowed and nobody saw it

Issue #2098, opened 2019-05-30
Webpack: generate a file containing the list of features


bfred-it:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%2440%20Funded-%2300A156.svg" />](https://issuehunt.io/r/sindresorhus/refined-github/issues/2098)

Recently we [added a UI](https://github.com/sindresorhus/refined-github/pull/1780) that lets you select which features to disable, but it requires the whole extension's code to run as a background page to collect the 100+ features' metadata.

As a background page, we only need the feature names, so it would be perfect if webpack just included or generated a file containing the list of features.

I already wrote [some code to generate the list](https://github.com/sindresorhus/refined-github/pull/1780/commits/1fcd383538367b59186a4d145bdaaf60f0366cfb), but **I need your help to generate a file with it,** like:

```js
// distribution/features-list.js
window.featuresList = ${JSON.stringify(features)}
```

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Sponsors (Total: $40.00)

- [<img src='https://avatars3.githubusercontent.com/u/44827199?v=4' alt='issuehunt' width=24 height=24> issuehunt](https://issuehunt.io/u/issuehunt) ($40.00)

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/sindresorhus/refined-github/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

IssueHuntBot:
@issuehunt has funded $40.00 to this issue.

---
- Submit pull request via [IssueHunt](https://issuehunt.io/repos/51769689/issues/2098) to receive this reward.
- Want to contribute? Chip in to this issue via [IssueHunt](https://issuehunt.io/repos/51769689/issues/2098).
- Checkout the [IssueHunt Issue Explorer](https://issuehunt.io/issues) to see more funded issues.
- Need help from developers? [Add your repository](https://issuehunt.io/r/new) on IssueHunt to raise funds.

bfred-it:
Sindre suggested https://github.com/sindresorhus/add-asset-webpack-plugin, I think I got it

bfred-it:
If anyone knows a dynamic replacer (i.e. `replace('process.env.LIST', () => listString)`)

IssueHuntBot:
[...]

PR #2099, opened 2019-05-30
Use auto-generated feature list to set defaults

bfred-it:
Closes https://github.com/sindresorhus/refined-github/issues/2098


webpack.config.ts
```diff
@@ -49,6 +51,16 @@ module.exports = (_env: string, argv: Record<string, boolean | number | string>)
...
 	plugins: [
+		new AddAssetPlugin('features-list.js', (compilation: any) => {
+			const features: string[] = [];
+			for (const file of compilation.fileDependencies) {
+				const [, feature = false] = file.match(/source\/features\/([^/]+)\.tsx$/) || []
+				if (feature) {
+					features.push(feature);
+				}
+			}
+			return `window.featuresList = ${JSON.stringify(features)}`
```
notlmn:
```suggestion
			return `window.featuresList = ${JSON.stringify(features)};`
```

bfred-it:
If anyone knows a dynamic replacer (i.e. `replace('process.env.LIST', () => listString)`) I can avoid the extra file altogether and inline the list in `distribution/options-storage.js`

webpack.config.ts
```diff
@@ -21,6 +22,25 @@ module.exports = (_env: string, argv: Record<string, boolean | number | string>)
 	},
 	module: {
 		rules: [
+			{
+				test: /options-storage/,
+				loader: 'string-replace-loader',
+				options: {
+					search: '__featuresList__',
+					replace: (() => {
```
bfred-it: This could be a function that runs every time the files change, but it's alright for now. If we find a better replacer webpack plugin the change will be _deleting 4 characters_

PR #2101, opened 2019-05-30
Restore `raw-file-link` feature

### LINKED ISSUE #2006: `raw-file-link` broken in PRs due to new dropdown
1. Visit https://github.com/sindresorhus/refined-github/pull/1783/files
2. No "Raw" link because they moved everything in the dropdown:

Last PR: https://github.com/sindresorhus/refined-github/pull/1924
cc @idasbiste

gavinhenderson:
Closes #2006  

View raw screenshot

# Test

https://github.com/sindresorhus/refined-github/pull/1783/files

### DIFF
--- a/source/features/raw-file-link.tsx
+++ b/source/features/raw-file-link.tsx
@@ -10,17 +10,16 @@ const createRawUrl = (href: string): string => {
 };
 
 function addRawButtons(): void {
-	for (const anchor of select.all<HTMLAnchorElement>('.file-header .file-actions > div.BtnGroup [href*=\'/blob/\']:last-child')) {
+	for (const anchor of select.all<HTMLAnchorElement>('.file-header .file-actions > div > details > details-menu > a:nth-child(3)')) {
 		anchor.after(
-			<a href={createRawUrl(anchor.pathname)} className="btn btn-sm BtnGroup-item">
-				Raw
+			<a href={createRawUrl(anchor.pathname)} className="pl-5 dropdown-item btn-link">
+				View raw
 			</a>
 		);
 	}
 }
 
 function init(): void {
-	addRawButtons();
 	onPrFileLoad(addRawButtons);
 }
 
--- a/source/features/raw-file-link.tsx
+++ b/source/features/raw-file-link.tsx
@@ -10,17 +10,14 @@ const createRawUrl = (href: string): string => {
 };
 
 function addRawButtons(): void {
-	for (const dropdown of select.all<HTMLElement>('.js-file-header-dropdown')) {
-		const anchor: HTMLAnchorElement | null = select('[data-ga-click^="View file"]', dropdown);
-
-		// Skip if the link has already been added
-		if (!select.exists('.raw-file-link', dropdown) && anchor) {
-			anchor.after(
-				<a href={createRawUrl(anchor.pathname)} className="pl-5 dropdown-item btn-link raw-file-link" role="menuitem">
-          View raw
-				</a>
-			);
-		}
+	const links = select.all<HTMLAnchorElement>('.js-file-header-dropdown [data-ga-click^="View file"]:not(.rgh-has-raw-file-link)');
+	for (const fileLink of links) {
+		fileLink.classList.add('rgh-has-raw-file-link');
+		fileLink.after(
+			<a href={createRawUrl(fileLink.pathname)} className="pl-5 dropdown-item btn-link" role="menuitem">
+				View raw
+			</a>
+		);
 	}
 }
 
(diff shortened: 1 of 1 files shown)


source/features/raw-file-link.tsx
```diff
@@ -10,17 +10,16 @@ const createRawUrl = (href: string): string => {
 };
 
 function addRawButtons(): void {
-	for (const anchor of select.all<HTMLAnchorElement>('.file-header .file-actions > div.BtnGroup [href*=\'/blob/\']:last-child')) {
+	for (const anchor of select.all<HTMLAnchorElement>('.file-header .file-actions > div > details > details-menu > a:nth-child(3)')) {
```
bfred-it: This selector seems overly specific, it should still look for `[href*=\'/blob/\']` instead

gavinhenderson:
@bfred-it  Thanks for the feedback! I've made it less specific, using the href selector as suggested. Hopefully this makes it more robust to any future UI changes from upstream

source/features/raw-file-link.tsx
```diff
@@ -10,17 +10,16 @@ const createRawUrl = (href: string): string => {
 };
 
 function addRawButtons(): void {
-	for (const anchor of select.all<HTMLAnchorElement>('.file-header .file-actions > div.BtnGroup [href*=\'/blob/\']:last-child')) {
+	for (const anchor of select.all<HTMLAnchorElement>('.file-header .file-actions > div > details [href*=\'/blob/\']')) {
```
notlmn: This can be `.file .js-file-header-dropdown [href*=\'/blob/\']`. You can even drop the `.file` part at the start too.

source/features/raw-file-link.tsx
```diff
@@ -10,17 +10,16 @@ const createRawUrl = (href: string): string => {
...
 		anchor.after(
-			<a href={createRawUrl(anchor.pathname)} className="btn btn-sm BtnGroup-item">
-				Raw
+			<a href={createRawUrl(anchor.pathname)} className="pl-5 dropdown-item btn-link">
```
notlmn:
```suggestion
			<a href={createRawUrl(anchor.pathname)} className="pl-5 dropdown-item btn-link" role="menuitem">
```

source/features/raw-file-link.tsx
gavinhenderson: Hey @notlmn , what's the benefit of that selector over others?

bfred-it: > what's the benefit of that selector over others?

The shorter and less specific the better. Many `>` will make it too dependent on small variations

bfred-it: Also perhaps we should avoid that `/blob/` check because that would match any link in a `/blob/` repo or `/blob/` directory. Let's try something like `[data-ga-click^="View file"]`

source/features/raw-file-link.tsx
```diff
@@ -10,17 +10,16 @@ const createRawUrl = (href: string): string => {
...
-				Raw
+			<a href={createRawUrl(anchor.pathname)} className="pl-5 dropdown-item btn-link" role="menuitem">
+				View raw
 			</a>
 		);
 	}
 }
 
 function init(): void {
-	addRawButtons();
```
bfred-it: This needs to be there, or else it will only run if there are many files in a PR

gavinhenderson:
@bfred-it I initially removed it as it was duplicating as you said. Latest commit restores the call before the PR is loaded and makes sure not to duplicate the link.

source/features/raw-file-link.tsx
```diff
@@ -10,16 +10,22 @@ const createRawUrl = (href: string): string => {
...
 
 function addRawButtons(): void {
-	for (const anchor of select.all<HTMLAnchorElement>('.js-file-header-dropdown [data-ga-click^="View file"]')) {
-		anchor.after(
-			<a href={createRawUrl(anchor.pathname)} className="pl-5 dropdown-item btn-link" role="menuitem">
-				View raw
-			</a>
-		);
+	for (const dropdown of select.all<HTMLElement>('.js-file-header-dropdown')) {
+		const anchor: HTMLAnchorElement | null = select('[data-ga-click^="View file"]', dropdown);
```
bfred-it: In this case you can add a `!` at the end instead of specifying `|null`, and can specify the expected type with `<...>`

```
const anchor = select<HTMLAnchorElement>('[data-ga-click^="View file"]', dropdown)!;
```

bfred-it:
Thanks for the fix 🐛 , @gavinhenderson

gavinhenderson:
@bfred-it Thanks for the help!

PR #2108, opened 2019-06-01
Add `highest-rated-comment` feature

### LINKED ISSUE #1898: 'Top/Most Helpful Comments' on issues
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/refined-github/issues/1898)

Hey!

Often, when debugging problems we run into when using open source tools, we end up on a relevant GitHub issue of the repository in question. These issues sometimes contain solutions to the problems in the comments, and users react to the most helpful comments using the 👍 and 🎉 reactions.

I think a feature which would highlight the comments with a large amount of 👍 and 🎉 would be really useful for quickly finding the most useful comments out of an issue. So instead of having to scroll through all of the comments, a user who ends up on the issue-page could instantly see the most 'upvoted' comments which most likely contain the information they are looking for.

You could think of this rather similarly to Stack Overflow answers. Stack Overflow sorts the answers by having the most helpful ones at the top. But we probably don't want to sort the comments, since that could be very confusing for the user.

Thus, I think a good option might be to have a small list of 'top comments' somewhere at the top of an issue-page, which would contain a small snippet of the comment (e.g. first 50 characters), and by clicking the comment listed there it would scroll the user to the full comment.

If people agree that this would be a good feature, I could try to sketch a UI for it and make a PR :)

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Sponsors (Total: $40.00)

### Tips

</details>

lubien:
Closes #1898

# Test

https://github.com/leafo/sticky-kit/issues/201
https://github.com/facebook/react/issues/13991
https://github.com/flutter/flutter/issues/11609 has controversial comments
https://github.com/MrMEEE/bumblebee-Old-and-abbandoned/issues/123

### DIFF
--- a/source/features/highest-rated-comment.css
+++ b/source/features/highest-rated-comment.css
@@ -1,9 +1,10 @@
 .rgh-highest-rated-comment {
-	border: 1px solid #ffa500 !important;
+	border: 2px solid #ffa500 !important;
 }
 
 a.rgh-highest-rated-comment .timeline-comment-header-text {
 	text-overflow: ellipsis;
 	white-space: nowrap;
 	overflow: hidden;
+	max-width: auto;
 }
--- a/source/features/highest-rated-comment.css
+++ b/source/features/highest-rated-comment.css
@@ -1,12 +1,9 @@
 .rgh-highest-rated-comment {
-	border-width: 1px;
-	border-style: solid;
-	border-color: #ffa500 !important;
+	border: 1px solid #ffa500 !important;
 }
 
 .rgh-highest-rated-comment .rgh-highest-rated-comment-text {
 	text-overflow: ellipsis;
 	white-space: nowrap;
 	overflow: hidden;
-	flex-grow: 1;
 }
--- a/source/features/highest-rated-comment.css
+++ b/source/features/highest-rated-comment.css
@@ -1,16 +1,16 @@
 .rgh-highest-rated-comment.comment {
-	border-color: orange;
-    border-width: 2px;
+	border-color: #ffa500;
+	border-width: 2px;
 }
 
 .rgh-highest-rated-comment .rgh-highest-rated-comment-summary {
-	border: solid 1px orange;
-	border-bottom: solid 1px orange !important;
+	border: solid 1px #ffa500;
+	border-bottom: solid 1px #ffa500 !important;
 }
 
 .rgh-highest-rated-comment .rgh-highest-rated-comment-btn {
 	margin-right: 10px;
-	padding: 3px 3px;
+	padding: 3px;
 }
 
 .rgh-highest-rated-comment .rgh-highest-rated-comment-text {
@@ -22,5 +22,5 @@
 
 .rgh-highest-rated-comment a {
 	/* reset default styles */
-	text-decoration: none!important;
+	text-decoration: none !important;
 }
--- a/source/features/highest-rated-comment.css
+++ b/source/features/highest-rated-comment.css
@@ -6,5 +6,5 @@ a.rgh-highest-rated-comment .timeline-comment-header-text {
 	text-overflow: ellipsis;
 	white-space: nowrap;
 	overflow: hidden;
-	max-width: auto;
+	max-width: none;
 }
--- a/source/features/highest-rated-comment.css
+++ b/source/features/highest-rated-comment.css
@@ -1,16 +1,5 @@
-.rgh-highest-rated-comment.comment {
-	border-color: #ffa500;
-	border-width: 2px;
-}
-
-.rgh-highest-rated-comment .rgh-highest-rated-comment-summary {
-	border: solid 1px #ffa500;
-	border-bottom: solid 1px #ffa500 !important;
-}
-
-.rgh-highest-rated-comment .rgh-highest-rated-comment-btn {
-	margin-right: 10px;
-	padding: 3px;
+.rgh-highest-rated-comment {
+	border: solid 1px #ffa500 !important;
 }
 
 .rgh-highest-rated-comment .rgh-highest-rated-comment-text {
@@ -18,9 +7,4 @@
 	white-space: nowrap;
 	overflow: hidden;
 	flex-grow: 1;
-}
-
-.rgh-highest-rated-comment a {
-	/* reset default styles */
-	text-decoration: none !important;
 }
--- a/source/features/highest-rated-comment.css
+++ b/source/features/highest-rated-comment.css
@@ -1,5 +1,5 @@
 .rgh-highest-rated-comment {
-	border: solid 1px #ffa500 !important;
+	border: solid 1px #ffa500;
 }
 
 .rgh-highest-rated-comment .rgh-highest-rated-comment-text {
--- a/source/features/highest-rated-comment.css
+++ b/source/features/highest-rated-comment.css
@@ -1,5 +1,7 @@
 .rgh-highest-rated-comment {
-	border: solid 1px #ffa500;
+	border-width: 1px;
+	border-style: solid;
+	border-color: #ffa500 !important;
 }
 
 .rgh-highest-rated-comment .rgh-highest-rated-comment-text {
--- a/source/features/highest-rated-comment.css
+++ b/source/features/highest-rated-comment.css
@@ -2,7 +2,7 @@
 	border: 1px solid #ffa500 !important;
 }
 
-.rgh-highest-rated-comment .rgh-highest-rated-comment-text {
+a.rgh-highest-rated-comment .timeline-comment-header-text {
 	text-overflow: ellipsis;
 	white-space: nowrap;
 	overflow: hidden;
--- a/source/features/highest-rated-comment.css
+++ b/source/features/highest-rated-comment.css
@@ -1,0 +1,26 @@
+.rgh-highest-rated-comment.comment {
+	border-color: orange;
+    border-width: 2px;
+}
+
+.rgh-highest-rated-comment .rgh-highest-rated-comment-summary {
+	border: solid 1px orange;
+	border-bottom: solid 1px orange !important;
+}
+
+.rgh-highest-rated-comment .rgh-highest-rated-comment-btn {
+	margin-right: 10px;
+	padding: 3px 3px;
+}
+
+.rgh-highest-rated-comment .rgh-highest-rated-comment-text {
+	text-overflow: ellipsis;
+	white-space: nowrap;
+	overflow: hidden;
+	flex-grow: 1;
+}
+
+.rgh-highest-rated-comment a {
+	/* reset default styles */
+	text-decoration: none!important;
+}
--- a/source/features/highest-rated-comment.tsx
+++ b/source/features/highest-rated-comment.tsx
@@ -10,8 +10,7 @@ function getCount(reaction: HTMLElement): number {
 
 function init(): false | void {
 	let highest;
-	const likes = select.all('.js-discussion .js-timeline-item:nth-child(n+6) [aria-label*="reacted with thumbs up"]');
-	for (const like of likes) {
+	for (const like of select.all('[aria-label*="reacted with thumbs up"]')) {
 		const count = getCount(like);
 		const dislike = select('[aria-label*="reacted with thumbs down"]', like.parentElement!);
 
--- a/source/features/highest-rated-comment.tsx
+++ b/source/features/highest-rated-comment.tsx
@@ -53,8 +53,8 @@ function init(): false | void {
 
 features.add({
 	id: 'highest-rated-comment',
-	description: 'Highlight and make a shortcut to most useful comments in issues.',
-	screenshot: 'https://i.imgur.com/vXmv0R6.png',
+	description: 'The most useful comment in issues is highlighted.',
+	screenshot: 'https://user-images.githubusercontent.com/1402241/58757449-5b238880-853f-11e9-9526-e86c41a32f00.png',
 	include: [
 		features.isIssue
 	],
--- a/source/features/highest-rated-comment.tsx
+++ b/source/features/highest-rated-comment.tsx
@@ -28,13 +28,13 @@ function init(): false | void {
 		return false;
 	}
 
-	const parent = highest.like.closest('.js-timeline-item')!;
-	const {hash} = select<HTMLAnchorElement>('.timestamp', parent)!;
-	const text = select('.comment-body', parent)!.textContent!.substring(0, 100);
-	const avatar = select('.avatar-parent-child.timeline-comment-avatar', parent)!.cloneNode(true);
+	const event = highest.like.closest('.js-timeline-item')!;
+	const text = select('.comment-body', event)!.textContent!.substring(0, 100);
+	const avatar = select('.timeline-comment-avatar', event)!.cloneNode(true);
+	const {hash} = select<HTMLAnchorElement>('.timestamp', event)!;
 
-	select('.unminimized-comment', parent)!.classList.add('rgh-highest-rated-comment');
-	parent.parentElement!.firstElementChild!.after((
+	select('.unminimized-comment', event)!.classList.add('rgh-highest-rated-comment');
+	event.parentElement!.firstElementChild!.after((
 		<div className="timeline-comment-wrapper">
 			{avatar}
 
--- a/source/features/highest-rated-comment.tsx
+++ b/source/features/highest-rated-comment.tsx
@@ -34,21 +34,25 @@ function init(): false | void {
 	const {hash} = select<HTMLAnchorElement>('.timestamp', event)!;
 
 	select('.unminimized-comment', event)!.classList.add('rgh-highest-rated-comment');
-	event.parentElement!.firstElementChild!.after((
-		<div className="timeline-comment-wrapper">
-			{avatar}
 
-			<a href={hash} className="no-underline rounded-1 rgh-highest-rated-comment bg-gray px-2 d-flex flex-items-center">
-				<span className="btn btn-sm mr-2 pr-1">
-					{icons.arrowDown()}
-				</span>
+	const position = select.all('.js-comment').indexOf(highest.like.closest('.js-comment') as HTMLElement);
+	if (position >= 4) {
+		event.parentElement!.firstElementChild!.after((
+			<div className="timeline-comment-wrapper">
+				{avatar}
 
-				<span className="text-gray timeline-comment-header-text">
-					Highest-rated comment: <em>{text}</em>
-				</span>
-			</a>
-		</div>
-	));
+				<a href={hash} className="no-underline rounded-1 rgh-highest-rated-comment bg-gray px-2 d-flex flex-items-center">
+					<span className="btn btn-sm mr-2 pr-1">
+						{icons.arrowDown()}
+					</span>
+
+					<span className="text-gray timeline-comment-header-text">
+						Highest-rated comment: <em>{text}</em>
+					</span>
+				</a>
+			</div>
+		));
+	}
 }
 
 features.add({
--- a/source/features/highest-rated-comment.tsx
+++ b/source/features/highest-rated-comment.tsx
@@ -4,30 +4,6 @@ import select from 'select-dom';
 import features from '../libs/features';
 import * as icons from '../libs/icons';
 
-type Props = {
-	hash: string;
-	text: string;
-	avatar: HTMLImageElement;
-}
-
-const element = ({hash, text, avatar}: Props): Node => (
-	<div className="timeline-comment-wrapper">
-		{avatar}
-
-		<a href={hash} className="no-underline rounded-1 rgh-highest-rated-comment d-block">
-			<div className="bg-gray px-2 d-flex flex-items-center">
-				<span className="btn btn-sm mr-2 pr-1">
-					{icons.arrowDown()}
-				</span>
-
-				<div className="text-gray timeline-comment-header-text rgh-highest-rated-comment-text">
-					Highest-rated comment: <em>{text}</em>
-				</div>
-			</div>
-		</a>
-	</div>
-);
-
 function getCount(reaction: HTMLElement): number {
 	return Number(reaction.textContent!.match(/\d+/)![0]);
 }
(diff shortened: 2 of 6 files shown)


source/features/highest-rated-comment.tsx
```diff
@@ -0,0 +1,125 @@
...
+		</div>
+
+		<a href={`#${id}`}>
+			<div className="bg-white border details-reset rounded-1">
+				<div className="bg-gray border-bottom-0 px-2 py-0 rgh-highest-rated-comment-summary">
+					<div className="d-flex flex-items-center">
+						<a href={`#${id}`} className="btn btn-sm rgh-highest-rated-comment-btn">
+							<svg height="16" className="octicon octicon-arrow-down" viewBox="0 0 10 16" version="1.1" width="20" aria-hidden="true">
+								<path fill-rule="evenodd" d="M7 7V3H3v4H0l5 6 5-6H7z"></path>
+							</svg>
```
bfred-it: This icon should be part of `icons.tsx`. Also `viewBox` and `version` attributes should be dropped, and the other attributes sorted like the other icons are in there.

source/features/highest-rated-comment.tsx
```diff
@@ -0,0 +1,125 @@
...
+						</a>
+
+						<div className="text-gray timeline-comment-header-text rgh-highest-rated-comment-text">
+							<span>Highest-Rated Comment: {text}</span>
```
bfred-it:
```suggestion
							Highest-rated comment: <em>{text}</em>
```

source/features/highest-rated-comment.tsx
```diff
@@ -0,0 +1,125 @@
...
+						</div>
+					</div>
+				</div>
+			</div>
+		</a>
+	</div>
+);
+
+function init(): void {
```
bfred-it: In this function you can simplify things a lot, but you'll have to fully rewrite it.

Select the comments that have 👍 reactions, so we skip a lot of them already. You can do so by selecting the specific button, and use a simple `for-of` to find the highest non-controversial comment:


```js
function getCount(reaction): number {
	return Number(reaction.textContent!.match(/\d+/)![0]!);
}

let highest;
for (const like of select.all('.reaction-summary-item[value="THUMBS_UP react"]')) {
	const count = getCount(like);
	const dislike = select('[value="THUMBS_DOWN react"]', like.parentElement);
	if (dislike && getCount(dislike) >= count / 2) {
		continue; // Controversial comment
	}
	if (!highest) {
		highest = {like, count};
	} else if (count > highest.count) {
		highest = {like, count};
	}
}
```

at the end of it, you'll have exactly the highest-rated non-controversial comment. Then just make sure its likes are enough (>10, > comments * 0.8) and add it.

```js
if (highest && highest.count > 10) {
	append(element(highest.like.closest('.comment')))
}
```

source/features/highest-rated-comment.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+features.add({
+	id: 'highest-rated-comment',
+	description: 'Highlight and make a shortcut to most useful comments in issues.',
+	screenshot: 'https://i.imgur.com/vXmv0R6.png',
+	include: [
+		features.isIssue
+	],
+	exclude: [
+	],
```
bfred-it: You can drop this

source/features/highest-rated-comment.tsx
```diff
@@ -0,0 +1,125 @@
...
+	load: features.onDomReady, // Wait for dom-ready
```
bfred-it:
```suggestion
	load: features.onAjaxedPages
```

source/features/highest-rated-comment.tsx
```diff
@@ -0,0 +1,125 @@
...
+		return notClose && mostLikes && minimum && !controversial;
+	}
+
+	const comment = options.find(candidate);
+
+	if (comment && comment.$el) {
+		const $parent = comment.$el.closest('.timeline-comment-group')!;
+		const {id} = $parent;
+		const username = select('.author', comment.$el)!.innerText;
+		const text = select('.comment-body', comment.$el)!.innerText.substring(0, 100);
```
bfred-it: Always prefer `.textContent`, it's much lighter

source/features/highest-rated-comment.css
```diff
@@ -0,0 +1,26 @@
...
+.rgh-highest-rated-comment .rgh-highest-rated-comment-text {
+	text-overflow: ellipsis;
+	white-space: nowrap;
+	overflow: hidden;
+	flex-grow: 1;
+}
+
+.rgh-highest-rated-comment a {
+	/* reset default styles */
+	text-decoration: none !important;
```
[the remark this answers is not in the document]
lubien: This is for the preview on the top. Somewhat links get the underline decoration upon hover there. I'll look for another approach here.

lubien:
I'm on it

source/features/highest-rated-comment.css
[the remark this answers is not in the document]
lubien: For reference:

image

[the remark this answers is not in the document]
bfred-it: There are special classes for this, like `no-underline`

I found it with

1. Open Chrome's developer tools, `sources` tab
2. cmd+alt+f (search all the code)
3. `text-decoration:none`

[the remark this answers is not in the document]
lubien: TIL!

source/features/highest-rated-comment.tsx
lubien: I'm on this right now. Found out `THUMBS_UP react` becomes `THUMBS_UP unreact` upon reaction so I'm using `.reaction-summary-item[value="THUMBS_UP react"],.reaction-summary-item[value="THUMBS_UP unreact"]` if that's okay.

lubien: Should I still verify if it's at least the 5th comment?

bfred-it: We can do that later in the selector. I’ll send a patch

bfred-it: > I'm on this right now. Found out `THUMBS_UP react` becomes `THUMBS_UP unreact` upon reaction so I'm using `.reaction-summary-item[value="THUMBS_UP react"],.reaction-summary-item[value="THUMBS_UP unreact"]` if that's okay.

Try `[aria-label*="reacted with thumbs up"]` and `[aria-label*="reacted with thumbs down"]` for the dislike button

lubien: Works just fine.

lubien: So this would be the selector: `.js-discussion .js-timeline-item:nth-child(n+6) [aria-label*="reacted with thumbs up"]`

n+6 because there's a hidden item just after the first comment so to get the actual 5th elemente we need +6.

What do you think?

lubien:
Done

source/features/highest-rated-comment.tsx
bfred-it: Yes! The only issue is that `.js-timeline-item` includes labels and title change events, so it's not completely reliable, but maybe it's "good enough" for now.

source/libs/icons.tsx
```diff
@@ -67,3 +67,5 @@ export const privateLockFilled = (): SVGElement => (
 		<path d="M5.05 9.46h-1v-1h1zm0 2.93h-1v1h1zm0-2h-1v1h1z" />
 	</svg>
 );
+
+export const arrowDown = (): SVGElement => <svg aria-hidden="true" className="octicon octicon-arrow-down" width="16" height="16"><path fill-rule="evenodd" d="M7 7V3H3v4H0l5 6 5-6H7z"></path></svg>;
```
bfred-it:
```suggestion
export const arrowDown = (): SVGElement => <svg aria-hidden="true" className="octicon octicon-arrow-down" width="16" height="16"><path d="M7 7V3H3v4H0l5 6 5-6H7z" /></svg>;
```

source/features/highest-rated-comment.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+		if (!highest) {
+			highest = {$like, count};
+		} else if (count > highest.count) {
+			highest = {$like, count};
+		}
+	}
+
+	if (!highest || highest.count < 10) {
+		return;
```
bfred-it:
```suggestion
		return false;
```

source/features/highest-rated-comment.tsx
```diff
@@ -0,0 +1,82 @@
...
+			</div>
+		</a>
+	</div>
+);
+
+function getCount(reaction: HTMLElement): number {
+	return Number(reaction.textContent!.match(/\d+/)![0]);
+}
+
+function init(): void {
```
bfred-it:
```suggestion
function init(): false | void {
```

source/features/highest-rated-comment.tsx
```diff
@@ -0,0 +1,82 @@
...
+}
+
+const element = ({id, text, avatar}: Props): Node => (
+	<div className="timeline-comment-wrapper">
+		{avatar}
+
+		<a href={`#${id}`} className="no-underline">
+			<div className="bg-white details-reset rounded-1 rgh-highest-rated-comment">
+				<div className="bg-gray border-bottom-0 px-2 py-0">
+					<div className="d-flex flex-items-center">
```
bfred-it: Can any of these elements be merged or dropped? At the very least, the initial `a` should have the classes `bg-white details-reset rounded-1 rgh-highest-rated-comment` and that div can be dropped already.

source/features/highest-rated-comment.tsx
```diff
@@ -0,0 +1,82 @@
...
+	</div>
+);
+
+function getCount(reaction: HTMLElement): number {
+	return Number(reaction.textContent!.match(/\d+/)![0]);
+}
+
+function init(): void {
+	let highest;
+	const $likes = select.all('.js-discussion .js-timeline-item:nth-child(n+6) [aria-label*="reacted with thumbs up"]');
```
bfred-it: No `$` in the variable names, that makes it feel like it's jQuery

source/features/highest-rated-comment.tsx
```diff
@@ -0,0 +1,82 @@
...
+		const $dislike = select('[aria-label*="reacted with thumbs down"]', $like.parentElement!);
+
+		if ($dislike && getCount($dislike) >= count / 2) {
+			continue; // Controversial comment
+		}
+
+		if (!highest) {
+			highest = {$like, count};
+		} else if (count > highest.count) {
+			highest = {$like, count};
```
bfred-it: Actually we only need one condition
```js
if (!highest || count > highest.count) {
```

source/features/highest-rated-comment.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+	if (!highest || highest.count < 10) {
+		return;
+	}
+
+	const $parent = highest.$like.closest('.js-timeline-item')!;
+	const $comment = select('.comment', $parent)!;
+	const {id} = select('.timeline-comment-group', $parent)!;
+	const text = select('.comment-body', $parent)!.textContent!.substring(0, 100);
+	const $avatar = select('.avatar-parent-child.timeline-comment-avatar', $parent)! as HTMLImageElement;
```
bfred-it:
```suggestion
	const avatar = select<HTMLImageElement>('.avatar-parent-child.timeline-comment-avatar', $parent)!
		.cloneNode(true) as HTMLImageElement;
```

source/features/highest-rated-comment.tsx
```diff
@@ -0,0 +1,82 @@
...
+		}
+	}
+
+	if (!highest || highest.count < 10) {
+		return;
+	}
+
+	const $parent = highest.$like.closest('.js-timeline-item')!;
+	const $comment = select('.comment', $parent)!;
+	const {id} = select('.timeline-comment-group', $parent)!;
```
bfred-it: Let's try getting the permalink exactly:

```suggestion
	const {hash} = select<HTMLAnchorElement>('.timestamp', $parent)!;
```

source/features/highest-rated-comment.tsx
```diff
@@ -0,0 +1,82 @@
...
+	const text = select('.comment-body', $parent)!.textContent!.substring(0, 100);
+	const $avatar = select('.avatar-parent-child.timeline-comment-avatar', $parent)! as HTMLImageElement;
+	const avatar = $avatar.cloneNode(true) as HTMLImageElement;
+	const props: Props = {id, text, avatar};
+
+	select('.js-discussion')!.prepend(element(props));
+	$comment.classList.add('rgh-highest-rated-comment');
```
bfred-it: `$comment` isn't used anywhere else, so I'd rather see
```suggestion
	select('.comment', $parent)!.classList.add('rgh-highest-rated-comment');
```

source/features/highest-rated-comment.tsx
bfred-it: You can use this button to merge my suggestions as well

lubien: Somewhat this button doesn't work for me (even on Files page) so I'll manually add it.

lubien:
Implemented and added to CODEOWNERS ;)

source/features/highest-rated-comment.css
```diff
@@ -0,0 +1,12 @@
+.rgh-highest-rated-comment {
+	border-width: 1px;
+	border-style: solid;
+	border-color: #ffa500 !important;
```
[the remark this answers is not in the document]
lubien: I used separated cuz the original comment had default styles so doing this way would leave less chances of getting over other stuff but it's just me overthinking, if you want I can make it a oneliner again :D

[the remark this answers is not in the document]
bfred-it: Yeah we want it to be this way anyway, so one line is preferred

source/features/highest-rated-comment.tsx
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	const parent = highest.like.closest('.js-timeline-item')!;
+	const {hash} = select<HTMLAnchorElement>('.timestamp', parent)!;
+	const text = select('.comment-body', parent)!.textContent!.substring(0, 100);
+	const avatar = select('.avatar-parent-child.timeline-comment-avatar', parent)!
+		.cloneNode(true) as HTMLImageElement;
+	const props: Props = {hash, text, avatar};
+
+	select('.js-discussion')!.prepend(element(props));
```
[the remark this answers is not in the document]
lubien: Okay. I'm gonna get rid of Props type too then.

source/features/highest-rated-comment.tsx
```diff
@@ -0,0 +1,75 @@
...
+	text: string;
+	avatar: HTMLImageElement;
+}
+
+const element = ({hash, text, avatar}: Props): Node => (
+	<div className="timeline-comment-wrapper">
+		{avatar}
+
+		<a href={hash} className="no-underline rounded-1 rgh-highest-rated-comment d-block">
+			<div className="bg-gray px-2 d-flex flex-items-center">
```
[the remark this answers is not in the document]
lubien: Actually worked. I remember I tried this one but perhaps I told chrome to reload extension too soon before the compilation.

lubien:
Done ;)

source/features/highest-rated-comment.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+	if (!highest || highest.count < 10) {
+		return false;
+	}
+
+	const parent = highest.like.closest('.js-timeline-item')!;
+	const {hash} = select<HTMLAnchorElement>('.timestamp', parent)!;
+	const text = select('.comment-body', parent)!.textContent!.substring(0, 100);
+	const avatar = select('.avatar-parent-child.timeline-comment-avatar', parent)!
+		.cloneNode(true) as HTMLImageElement;
```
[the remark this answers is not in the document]
bfred-it: Sorry I meant “on one line”

source/features/highest-rated-comment.tsx
```diff
@@ -0,0 +1,75 @@
...
+	<div className="timeline-comment-wrapper">
+		{avatar}
+
+		<a href={hash} className="no-underline rounded-1 rgh-highest-rated-comment d-block">
+			<div className="bg-gray px-2 d-flex flex-items-center">
+				<span className="btn btn-sm mr-2 pr-1">
+					{icons.arrowDown()}
+				</span>
+
+				<div className="text-gray timeline-comment-header-text rgh-highest-rated-comment-text">
```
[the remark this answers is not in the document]
bfred-it: Is flex still needed here?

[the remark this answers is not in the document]
lubien: I dropped the flex from CSS already but if you mean it's parent, yes.

source/features/highest-rated-comment.tsx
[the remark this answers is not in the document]
lubien: Done

bfred-it:
New screenshot:

bfred-it:
Can you find more threads where this can be tested?

source/features/highest-rated-comment.tsx
```diff
@@ -0,0 +1,63 @@
...
+		return false;
+	}
+
+	const event = highest.like.closest('.js-timeline-item')!;
+	const text = select('.comment-body', event)!.textContent!.substring(0, 100);
+	const avatar = select('.timeline-comment-avatar', event)!.cloneNode(true);
+	const {hash} = select<HTMLAnchorElement>('.timestamp', event)!;
+
+	select('.unminimized-comment', event)!.classList.add('rgh-highest-rated-comment');
+	event.parentElement!.firstElementChild!.after((
```
bfred-it: The highest-rate comment might still be within those first 5 comments, so we should still highlight that one _but_ we should not add the link.

Example: https://github.com/facebook/react/issues/12014

To do, perhaps you can use:
```suggestion
	const position = select.all('.js-comment').indexOf(select('.js-comment', event));
	if (position > 5) {
		event.parentElement!.firstElementChild!.after((
```

lubien:
I tried to search for cases on the 5th+ comment but had no luck over some huge repos I went through. Most would be the first one.

About the changes I can tackle monday if it's okay to you. Today I have to deal with some things first :)

source/features/highest-rated-comment.tsx
lubien: Used >=4 to hit from the 5th, hope that's fine.

source/features/highest-rated-comment.tsx
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	const event = highest.like.closest('.js-timeline-item')!;
+	const text = select('.comment-body', event)!.textContent!.substring(0, 100);
+	const avatar = select('.timeline-comment-avatar', event)!.cloneNode(true);
+	const {hash} = select<HTMLAnchorElement>('.timestamp', event)!;
+
+	select('.unminimized-comment', event)!.classList.add('rgh-highest-rated-comment');
+
+	const position = select.all('.js-comment').indexOf(highest.like.closest('.js-comment') as HTMLElement);
```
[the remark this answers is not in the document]
lubien: image

Tried that one.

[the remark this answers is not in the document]
bfred-it: This should work
```suggestion
	const position = select.all<Element>('.js-comment').indexOf(highest.like.closest('.js-comment'));
```

Maybe it also needs a `!`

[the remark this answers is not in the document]
lubien: I tried some positions for ! and couldn't manage to git this way working. Any clue?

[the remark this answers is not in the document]
bfred-it: This should work
```suggestion
	const position = select.all<Element>('.js-comment').indexOf(highest.like.closest('.js-comment')!);
```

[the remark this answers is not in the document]
lubien: Yep, it's one of those I tried.

image

Still a no go.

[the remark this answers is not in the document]
bfred-it: Indeed, the problem moved to `select.all` because the generic (`<...>`) needs to "extend" `HTMLElement`, but `Element` is extended _by_ `HTMLElement`

Your solution was fine. 😅

source/features/highest-rated-comment.tsx
```diff
@@ -0,0 +1,67 @@
...
+import features from '../libs/features';
+import * as icons from '../libs/icons';
+
+function getCount(reaction: HTMLElement): number {
+	return Number(reaction.textContent!.match(/\d+/)![0]);
+}
+
+function init(): false | void {
+	let highest;
+	const likes = select.all('.js-discussion .js-timeline-item:nth-child(n+6) [aria-label*="reacted with thumbs up"]');
```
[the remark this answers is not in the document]
bfred-it: Yeah, the point of https://github.com/sindresorhus/refined-github/pull/2108/commits/f94a89d8409501abfbfd6b0be0d8aec0c562199e was to _replace_ the selector, or else it never highlights anything before the 5th

bfred-it:
This is great! Thank you @lubien! 🥇

lubien:
Oh. I was gonna do it when I came back home. Thanks for your help 

Glad to work with you and see you next time

Lemmmy:
Really like this feature, makes reading long discussion pages significantly easier. However, sometimes the comment isn't minimised, and so the first time a user of the extension sees the gold border, it isn't immediately obvious what the colour means:

image

I had to look through the DOM to see what was adding the `border-color` to find out what it meant, especially as it is not yet in the changelog:

image

Perhaps adding a badge on the unminimised comment (similar to the 'Contributor', 'Collaborator' and 'Author' badges) saying "Highest rated comment" would improve accessibility for new users? Something like this:

image

This suggestion might be more appropriate in a separate issue or PR - happy to create one if so.

lubien:
Agreed @Lemmmy.

What's your take on this @bfred-it? Should we go for another issue?

bfred-it:
You can send a PR and let’s see how it looks. I don’t think it’s enough to make that connection “Orange = highest rated comment”

We don’t add the linking banner when the comment appears near the top (within the first 5 comments)

PR #2109, opened 2019-06-01
New cache implementation

### LINKED ISSUE #2100: Replace our cookie-based cache with self-clearing browser.storage
> > Cache where? Our cache is rather small and if we're gonna store them in a regular `browser.storage.local` we also need a cleaning mechanism so it doesn't stay there forever.
>
> I think a simple LRU implementation on top of `browser.storage.local` with lazy saving/cleaning should suffice. 

_Originally posted by @dotconnor in https://github.com/sindresorhus/refined-github/pull/2026_

Whatever the method used, it's important that unused items don't linger in the cache forever.

dotconnor:
What methods should I implement besides get/set?

 - [x] set
 - [x] get
 - [x] has
 - [x] purge (clean up old/unused cached data)

Closes #2100

# Test

https://github.com/sindresorhus/refined-github (Open devtools on the background page to see the console, repo pages use `release-count` and `default-branch` caches)

### DIFF
--- a/source/libs/cache.ts
+++ b/source/libs/cache.ts
@@ -1,17 +1,18 @@
 const storage: browser.storage.StorageArea = browser.storage.local;
 
-export async function has<TValue extends any = any>(key: string): Promise<boolean> {
+export async function has(key: string): Promise<boolean> {
 	const values = await storage.get(key);
 	return values[key] !== undefined;
 }
 
 export async function get<TValue extends any = any>(key: string): Promise<TValue | undefined> {
 	const values = await storage.get(key);
-	let value = values[key];
+	const value = values[key];
 	// If it's not in the cache, it's best to return "undefined"
 	if (value === null || value === undefined) {
 		return undefined;
 	}
+
 	console.log('CACHE: found', key, value.data);
 	return value.data;
 }
--- a/source/libs/cache.ts
+++ b/source/libs/cache.ts
@@ -44,6 +44,26 @@ export function set<TValue extends any = any>(key: string, value: TValue, expira
 	});
 }
 
+/* global chrome */
+const chromeStorage = {
+	get(keys: string | object | string[] | undefined) {
+		return new Promise(resolve => {
+			chrome.storage.local.get(keys, resolve);
+		});
+	},
+	set(items: any) {
+		return new Promise(resolve => {
+			chrome.storage.local.set(items, resolve);
+		});
+	}
+};
+
+const storage: browser.storage.StorageArea | undefined = (typeof browser === 'object' ? browser.storage.local : typeof chrome === 'object' ? (chromeStorage as unknown as browser.storage.StorageArea) : undefined);
+
+export function purge(): void {
+
+}
+
 /* Accept messages in background page */
 if (location.pathname === '/_generated_background_page.html') {
 	browser.runtime.onMessage.addListener(async (request: CacheRequest) => {
@@ -53,6 +73,17 @@ if (location.pathname === '/_generated_background_page.html') {
 
 		const {code, key, value, expiration} = request;
 		if (code === 'get-cache') {
+			if (storage) {
+				const values = await storage.get(key);
+				let value = values[key];
+				if (value) {
+					value = JSON.parse(value);
+					console.log('CACHE: found', key, value.data);
+					return value.data;
+				}
+			}
+
+			// Fallback to cookie based cache
 			const [cached] = document.cookie.split('; ')
 				.filter(item => item.startsWith(key + '='));
 			if (cached) {
@@ -67,7 +98,17 @@ if (location.pathname === '/_generated_background_page.html') {
 
 			// Store as JSON to preserve data type
 			// otherwise Booleans and Numbers become strings
-			document.cookie = `${key}=${JSON.stringify(value)}; max-age=${expiration ? expiration * 3600 * 24 : ''}`;
+			if (storage) {
+				storage.set({
+					[key]: JSON.stringify({
+						data: value,
+						lastUse: Date.now(),
+						expiration: Date.now() + (1000 * 3600 * 24)
+					})
+				});
+			} else {
+				document.cookie = `${key}=${JSON.stringify(value)}; max-age=${expiration ? expiration * 3600 * 24 : ''}`;
+			}
 		}
 	});
 }
--- a/source/libs/cache.ts
+++ b/source/libs/cache.ts
@@ -44,21 +44,7 @@ export function set<TValue extends any = any>(key: string, value: TValue, expira
 	});
 }
 
-/* global chrome */
-const chromeStorage = {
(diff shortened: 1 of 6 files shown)


source/libs/cache.ts
```diff
@@ -53,6 +59,17 @@ if (location.pathname === '/_generated_background_page.html') {
 
 		const {code, key, value, expiration} = request;
 		if (code === 'get-cache') {
+			if (storage) {
```
bfred-it: No fallbacks, it should just use the storage.

dotconnor: Purging? Heres what I was thinking:
On every get and/or set operation the purge function is called and does the following:

 - Gets all the data in storage currently
 - Filters out items not created the cache (probably by prefixing the keys)
 - Checks each item and see when it expires, if it has remove it.

We could also use the `Storage​Area​.get​Bytes​InUse` API to retrieve the size of the cache and remove items that haven't been used recently using the `lastUse` key.

source/libs/cache.ts
```diff
@@ -1,73 +1,33 @@
...
-		const {code, key, value, expiration} = request;
-		if (code === 'get-cache') {
-			const [cached] = document.cookie.split('; ')
-				.filter(item => item.startsWith(key + '='));
-			if (cached) {
-				const [, value] = cached.split('=');
-				console.log('CACHE: found', key, value);
-				return JSON.parse(value);
-			}
+export function purge(): void {
```
bfred-it: We don't really need a purge method, it should take care of itself

source/libs/cache.ts
```diff
@@ -1,73 +1,33 @@
...
-	return browser.runtime.sendMessage({
-		key,
-		value,
-		expiration,
-		code: 'set-cache'
+	console.log('CACHE: setting', key, value);
+	return storage.set({
+		[key]: {
+			data: value,
+			lastUse: Date.now(),
```
bfred-it: I think we should make the expiration time a required parameter.
And since we always have that we don't need the `lastUse`

source/libs/cache.ts
bfred-it: I don't think we should depend on the available space. When it's expired it should just go regardless.

If we check the entire cache on every get/set that will make each operation heavy. I'd suggest:

- when getting an item, ensure it's not expired. If it is, delete it and return undefined.
- automatically run the purge with a setInterval of 24h

source/libs/get-default-branch.ts
```diff
@@ -33,9 +33,14 @@ async function fetchFromApi(user: string, repo: string): Promise<any> {
 	}
 }
 
-export default function (): Promise<any> {
+export default async function (): Promise<any> {
```
bfred-it: I know it's unrelated, but this should probably be
```suggestion
export default async function (): Promise<string> {
```

source/libs/cache.ts
```diff
@@ -1,73 +1,66 @@
...
-		await set(key, value, expiration);
-	}
-
-	return value;
+export async function has(key: string): Promise<boolean> {
+	const values = await storage.get(key);
+	return values[key] !== undefined;
 }
 
 export async function get<TValue extends any = any>(key: string): Promise<TValue | undefined> {
```
bfred-it: I'm not sure, but can you try replacing all `any`s with `unknown` in this file?

bfred-it:
Apart from the review, is this PR still a WIP?

source/libs/cache.ts
```diff
@@ -59,8 +57,8 @@ async function purge(): Promise<void> {
 }
 
 // Automatically clear cache every day
-if (location.pathname === '/_generated_background_page.html') {
+if (isBackgroundPage()) {
 	// Call immediately to ensure the cache gets purged at some point.
```
bfred-it: I preferred my comment, it's no longer "immediate" https://github.com/sindresorhus/refined-github/pull/2109#discussion_r290444434

source/libs/cache.ts
```diff
@@ -1,73 +1,64 @@
...
-			// otherwise Booleans and Numbers become strings
-			document.cookie = `${key}=${JSON.stringify(value)}; max-age=${expiration ? expiration * 3600 * 24 : ''}`;
-		}
-	});
+// Automatically clear cache every day
+if (isBackgroundPage()) {
+	// Call immediately to ensure the cache gets purged at some point.
+	setTimeout(purge, 60000);
+	setInterval(purge, 1000 * 3600 * 24);
 }
```
bfred-it: I'd add a `export default {has, get, set}` instead of exporting each method separately, so it can be used as 

```js
import cache from './libs/cache'
```

source/libs/cache.ts
dotconnor: Sorry I missed that one.

dotconnor:
@bfred-it I'm good with this when you are.

bfred-it:
The suggested test isn't working in Chrome nor Firefox. They appear in the cache but the cached releases number isn't picked up and shown elsewhere

Screenshot 2019-06-05 at 03 58 58

dotconnor:
@bfred-it Looks like the default branch one is working (for me at least), still trying to figure out why the release count one isn't working.

bfred-it:
Thanks for this! Are you interested in publishing it as a module? I’ve been wanting to publish a module like this as `webext-storage-cache`

dotconnor:
@bfred-it https://github.com/dotconnor/webext-cache

bfred-it:
I'd have preferred `webext-storage-cache` because there's also the regular browser cache which might be confused with it

bfred-it:
I got the repo transfer, thanks! Can you `npm unpublish webext-cache --force` to avoid duplicates? I just published a temporary `webext-storage-cache`

```yaml
Storage Cache for Web Extensions
https://github.com/dotconnor/webext-storage-cache#readme

dist
.tarball: https://registry.npmjs.org/webext-storage-cache/-/webext-storage-cache-0.0.0.tgz
.shasum: ad847845ce2cb56ce53c89b68a991d46635e2c1c
.integrity: sha512-ZwBQLPHrETO5uf81Cw4iTr75KaQj5vGs8vW3LXfm1pl3FitflGMB2h/x1zEs8c5nkGJWseD6sF3dftx4XJ2ZgA==
.unpackedSize: 4.7 kB

dependencies:
webext-detect-page: ^0.9.1 

maintainers:
- bfred-it <npm@bfred.it>

dist-tags:
latest: 0.0.0  

published 2 minutes ago by bfred-it <npm@bfred.it>
```

notlmn:
I'd suggest to open a new issue to track the migration process, updating across all features that use cache.

bfred-it:
You can use the current version for now. When I'm ready with the npm module I'll just implement it in RGH

notlmn:
Weird, after this change I'm getting linting errors from VSCode but **not** when running `npm run lint:js`. For comparison, this PR should've errored out.

image

PR #2111, opened 2019-06-02
Make Enterprise injection more reliable

bfred-it:
I changed the way content scripts are injected. They are no longer injected manually with `tabs.executeScript` but they are declaratively registered via `contentScript.register` (which is polyfilled to `declarativeContent.RequestContentScript`).

2 polyfills and the glue for both (which registers the scripts on new origins) will be extracted into their own modules once verified that this all works.

The only minor bug right now is in Firefox: even if you remove the permission, the scripts will still be injected until the next restart. It's Firefox' bug, nothing asynchronous works after `request`/`remove` for some reason.

cc @busches 

# Test

1. Visit github.com, no "double injection" errors should ever appear, even if you try "Enable Refined GitHub on this domain" 
---
1. Visit google.com, no injection happening, so no errors in the console or background.js
1. Visit google.com/images, "Enable Refined GitHub on this domain"
2. Reload the page and see the errors in the console (obviously, because it's not GitHub)
3. Search for images
4. Open a photo
5. You shouldn't see the same errors repeated, because it's an ajaxed page and it the content scripts shouldn't be injected multiple times.

In no case should there ever be any errors in the console. I dropped a try/catch so the previously-silenced errors should no longer ever happen.


source/manifest.json
```diff
@@ -15,6 +15,7 @@
 		"storage",
 		"contextMenus",
 		"activeTab",
+		"https://github.com/*",
```
webknjaz:
```suggestion
		"x",
```

source/manifest.json
```diff
@@ -15,6 +15,7 @@
...
 		"https://api.github.com/*"
```
webknjaz:
```suggestion
 		"y"
```

source/libs/permission-events-polyfill.ts
```diff
@@ -0,0 +1,36 @@
+/* global chrome */
```
notlmn: @bfred-it Can you extract this file out as another npm module?

bfred-it:
Yes after this has been tested and works correctly. There are 3 modules to be extracted

busches:
FYI I changed jobs and don't currently have access to GHE to test currently. 😐

PR #2112, opened 2019-06-02
Fix typo in `fix-hidden-comments` feature's ID

dshoreman:
Fixes an issue where clicking the source link in Features list (in about:addons) would result in a 404 due to id being different to the actual source file.

When I loaded my dev copy as a temporary extension in Firefox, changing the ID seemed to result in the feature being disabled. Not sure if there's some way to fix that?


bfred-it:
Thank you @dshoreman

PR #2114, opened 2019-06-03
Add `clean-issue-filters` feature

### LINKED ISSUE #2016: Hide irrelevant issue/PR filters
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/refined-github/issues/2016)

In a repo's issue/PR list, some filters are irrelevant for specific repos:

* **Labels** if there are no labels in the repo
* **Projects** if there are no projects in the repo
* **Milestones** if there are no milestones in the repo
* **Status** if there is no CI in the repo

These filter buttons can be hidden to unclutter the UI (and maybe make space for more useful features).

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Sponsors (Total: $40.00)

### Tips

</details>

notlmn:
Closes #2016.

## Test
- With no projects or milestones: https://github.com/ramlmn/serv/pulls.
- With repo level projects and no milestones: https://github.com/notlmn-dump/test/pulls.
- With milestone and no projects: https://github.com/notlmn-dump/now-express-test/pulls.
- With org level projects and not milestones: https://github.com/notlmn-dump/opev-gen-googio/pulls (needs org access for oAuth token).


notlmn:
Also thought of adding support for repo deployment status, but that requires another API request (one to get latest commit, and one to get its deployment status). And targeting status dropdown is a bit of problem too.

source/libs/utils.ts
```diff
@@ -86,3 +86,7 @@ export function getOP(): string {
 
 	return select('.timeline-comment-header-text .author')!.textContent!;
 }
+
+export function idx(target: any, path: string[]): any | null | undefined {
```
sindresorhus: `idx` is not a good name

Use `unknown` instead of `any` whenever possible.
```suggestion
export function idx(target: unknown, path: string[]): unknown {
```

notlmn: > `idx` is not a good name

Apparently that's what [Facebook uses](https://github.com/facebook/Docusaurus/pull/917#discussion_r213177032), every library that does this has a weird name anyway.

source/features/clean-issue-filters.tsx
```diff
@@ -0,0 +1,43 @@
+import select from 'select-dom';
+import features from '../libs/features';
+import {v4} from '../libs/api';
```
bfred-it: I prefer `* as api` because using it as `api.v4` is more readable

source/features/clean-issue-filters.tsx
```diff
@@ -0,0 +1,43 @@
...
+	const result = await v4(`
+		query {
+			repository(owner: "${ownerName}", name: "${repoName}") {
+				projects { totalCount }
+				milestones { totalCount }
+				labels { totalCount }
+			},
+			organization(login: "${ownerName}") {
+				projects { totalCount }
+			}
```
bfred-it: Users have their own projects too

source/features/clean-issue-filters.tsx
```diff
@@ -0,0 +1,43 @@
...
+}
+
+features.add({
+	id: 'clean-issue-filters',
+	description: 'Hide irrelevant issue/PR filters',
+	init,
+	load: features.onAjaxedPages,
+	include: [
+		features.isPRList,
+		features.isIssueList
```
bfred-it: There’s already `isDiscussionList` that covers both

source/features/clean-issue-filters.tsx
```diff
@@ -0,0 +1,43 @@
...
+			organization(login: "${ownerName}") {
+				projects { totalCount }
+			}
+		}
+	`, {
+		allowErrors: true
+	});
+
+	// Hide projects filter only if there are no repo and organization level projects
+	if (result.repository.projects.totalCount === 0 && !idx(result, ['organization', 'projects', 'totalCount'])) {
```
bfred-it: Is idx required here? Maybe it’s just `results.organization && results.organization.projects.totalCount` because if organization exists, projects and their count exist too, I’d think.

source/features/clean-issue-filters.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+	// Hide milestones filter if there are none
+	if (result.repository.milestones.totalCount === 0) {
+		select('[data-hotkey="m"')!.parentElement!.remove();
+	}
+}
+
+features.add({
+	id: 'clean-issue-filters',
+	description: 'Hide irrelevant issue/PR filters',
```
bfred-it: Maybe “empty” instead of irrelevant

source/features/clean-issue-filters.tsx
notlmn: `isDiscussionList` contains `isGlobalDiscussionList` which we don't need, and `isRepoDiscussionList` which targets labels page too.

All of which not required.

source/libs/utils.ts
notlmn: Adding `AnyObject` as type to `target` gives type-not-found error while running ava, it looks like types from `globals.d.ts` are not included while running tests.

But fixed that using `ts-node/register/transpile-only` option in `package.json`, which makes the tests run way faster too. Maybe worth looking into adding this!?

source/features/clean-issue-filters.tsx
bfred-it: > `isRepoDiscussionList` which targets labels page too

labels page is still a regular discussion list, so it's indeed required: https://github.com/sindresorhus/refined-github/labels/bug

source/features/clean-issue-filters.tsx
```diff
@@ -0,0 +1,49 @@
...
+			}
+		}
+	`, {
+		allowErrors: true
+	});
+
+	// Hide projects filter only if there are no repo, user, and organization level projects
+	if (
+		result.repository.projects.totalCount === 0 &&
+		(result.organization && result.organization.projects.totalCount === 0) &&
```
bfred-it: If organization === undefined, the link won't be removed.

I think the check should be:

```js
!result.organization || result.organization.projects.totalCount === 0
```

source/features/clean-issue-filters.tsx
notlmn: Oh, thought that was targeting just `/labels`.

bfred-it: This is still unresolved. `isIssueList` can be dropped from `page-detect` as well

source/features/clean-issue-filters.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+	// Hide milestones filter if there are none
+	if (result.repository.milestones.totalCount === 0) {
+		select('[data-hotkey="m"')!.parentElement!.remove();
+	}
+}
+
+features.add({
+	id: 'clean-issue-filters',
+	description: 'Hide empty issue/PR filters',
```
bfred-it:
```suggestion
	description: 'Hide empty issue/PR filters in lists',
```

notlmn:
> If not impossible, it would be nice to also drop the Checks selector. Does the API include that?

Took me a while to debug, but that's part of `filter-pr-by-build-status`. The dropdown doesn't exist until that features adds it, maybe it's better to handle this in that feature.

----------------------

> Also it would be good to cache the result of this, per-repo, especially with #2109 coming :)

Let's deal with that in a separate PR. Or will have to wait until #2109 is merged.

source/libs/utils.ts
bfred-it: Transpile-only doesn’t actually check the types, so it defeats the purpose of using typescript

bfred-it:
> The dropdown doesn't exist until that features adds it, maybe it's better to handle this in that feature.

D’oh, I forgot about that

source/features/clean-issue-filters.tsx
```diff
@@ -0,0 +1,48 @@
...
+			repository(owner: "${ownerName}", name: "${repoName}") {
+				projects { totalCount }
+				milestones { totalCount }
+			},
+			organization(login: "${ownerName}") {
+				projects { totalCount }
+			}
+			user(login: "${ownerName}") {
+				projects { totalCount }
+			}
```
bfred-it: Apparently even if I have a personal project, public, with an issue from the current repo inside, the dropdown is still empty for me.

I dropped this

FloEdelmann:
Thanks @notlmn! :blush:

bfred-it:
It does look much better now!

notlmn:
> > If not impossible, it would be nice to also drop the Checks selector. Does the API include that?

> Took me a while to debug, but that's part of filter-pr-by-build-status. The dropdown doesn't exist until that features adds it, maybe it's better to handle this in that feature.

Opened #2127 to track this.

yakov116:
@notlmn 	nothing major but
```
❌ clean-issue-filters
content.js:2111 TypeError: Cannot read property 'parentElement' of null
    at clean_issue_filters_init (content.js:6423)
    at async features_run (content.js:2106)
```
When a repo has not Issues or Pull requests

PR #2115, opened 2019-06-03
Add `limit-commit-title-length` feature

### LINKED ISSUE #1475: Warn about long `Squash and Merge` commit message
When merging pull requests with `Squash and Merge` button it's possible to write too long header message, resulting trauncated line in commit log.

It would be useful to add a warning, similarly to what GitHub shows for file edit interface
image

Not sure if the limit should be 50 characters, like github has, or 72, like in [commitlint/config-conventional](https://github.com/marionebl/commitlint/blob/d636811b4537904231c483a05e91aaf96deb7f6e/%40commitlint/config-conventional/index.js#L5) and vscode's `git.inputValidation` setting.

It may be also useful to add a per-line limit to commit body and add countdown, like in vscode with `"git.inputValidation": "always"`:
image

notlmn:
Closes #1475.

- [x] Changed font for all commit message fields to monospace.
- [x] Hard-coded input fields' `maxlength` property to 70.
- [x] Adds a warning of field somehow gets filled with message longer than 70 chars.
- [ ] Possibly also handle issue/PR title changes too?

## Test
While creating a PR or merging it, the input fields do not accept characters greater than 70 characters in length.


source/features/limit-commit-title-length.tsx
```diff
@@ -0,0 +1,46 @@
...
+import features from '../libs/features';
+
+const getError = onetime<[], HTMLElement>((): HTMLElement =>
+	<p className="note text-red">Title of commit should be less than 70 characters.</p>
+);
+
+function init(): void {
+	const inputs = select.all<HTMLInputElement>([
+		'#commit-summary-input', // Commit title on edit file page
+		'#pull_request_title', // PR title while creating PR
```
bfred-it: This is unrelated. If the user wants the PR title to match the commit they’ll do so when merging the PR 🙂

source/features/limit-commit-title-length.tsx
```diff
@@ -0,0 +1,46 @@
...
+				getError().remove();
+				return;
+			}
+
+			inputField.after(getError());
+		};
+
+		inputField.setAttribute('maxlength', '70');
+
+		inputField.addEventListener('input', validityCallback);
```
bfred-it: I don’t think we need a custom UI for this. Can you ensure that the browser triggers the regular form validation? It’ll say “the field is longer than allowed” without any additional JS

Actually this would be even better than limiting with maxlength because maxlength makes it hard to edit the field.

source/features/limit-commit-title-length.tsx
```diff
@@ -0,0 +1,46 @@
...
+features.add({
+	id: 'limit-commit-title-length',
+	description: 'Limits the length of commit fields to 70 characters',
+	init,
+	load: features.onAjaxedPages,
+	include: [
+		features.isPRConversation,
+		features.isPRFiles,
+		features.isCompare,
+		features.isEditingFile
```
bfred-it: I think we already have a detection for this, but I may be wrong.

source/features/limit-commit-title-length.tsx
```diff
@@ -0,0 +1,46 @@
...
+}
+
+features.add({
+	id: 'limit-commit-title-length',
+	description: 'Limits the length of commit fields to 70 characters',
+	init,
+	load: features.onAjaxedPages,
+	include: [
+		features.isPRConversation,
+		features.isPRFiles,
```
bfred-it: I don’t think you can merge from the PR Files tab

source/features/limit-commit-title-length.tsx
notlmn: When creating a PR from a branch with a very long commit title, GitHub already separates out pull request title to 70 characters in title field and the rest into the message field.

So this should be a good practice for both the contributor (while opening PR) and for the owner (while merging PR) right!?

source/features/limit-commit-title-length.tsx
notlmn: Yes, GitHub validates all inputs before submitting using the forms `validate` attribute. But I think there is no way for us to trigger the browsers popup by our self.

The popup appears only while trying to submit the form, not while editing the field. For comparison, GitHub shows it's own warning message while typing, but that's not a strict enforcement anyway.

source/features/limit-commit-title-length.tsx
notlmn: No we don't (search for `edit` in `page-detect.ts` 😉).

source/features/limit-commit-title-length.tsx
bfred-it: >  think there is no way for us to trigger the browsers popup by our self.

The popup naturally comes up when validation is required, when the form is submitted, but I don't know if GitHub is using `novalidate` or something

notlmn: Yes, we can set custom error message using `setCustomValidity()`, but I wanted a warning while the user was typing. The custom validation message only appears when form is being submitted.

If that's too much of code to have for too little use, I'd be happy to drop it.

bfred-it: > but I wanted a warning while the user was typing

Not a bad idea, but I wouldn't want to make the area too busy, we already have some small text there from `sync-pr-commit-title`

How about using some `:invalid` style like in `options.css`?

notlmn: image

😅

test/page-detect.ts
```diff
@@ -391,3 +391,11 @@ test('isSingleTagPage', urlMatcherMacro, pageDetect.isSingleTagPage, [
 	'https://github.com/sindresorhus/refined-github',
 	'https://github.com/sindresorhus/refined-github/graphs'
 ]);
+
+test('isEditingFile', urlMatcherMacro, pageDetect.isEditingFile, [
+	'https://github.com/sindresorhus/refined-github/edit/master/readme.md',
+	'https://github.com/sindresorhus/refined-github/edit/ghe-injection/source/background.ts'
+], [
+	'https://github.com/sindresorhus/refined-github',
+	'https://github.com/sindresorhus/refined-github/pulls'
```
bfred-it: Tests need to include some tricky urls like a user called `edit`, a repo called `edit`, a file called `edit` and so on

notlmn: Done! 😊

notlmn:
> validation does not apply to the initial value (or values set programmatically...

The one thing I still can't figure out. But from what I've tested it looks like for the browser to validate autofilled inputs, the constraints have to be already present when the DOM is being created.

To bypass this I've tried `setCustomValidity(message)`, which does trigger validation on form, but only until the message is reset back to normal for some reason. I don't even know if the HTML Forms spec defines this kind of behavior.

In the example screenshot below, the input wasn't touched while running the following snippets.

image

But by using `setCustomValidity()`, we loose a lot of default error messages provided by the user agents. And again, with resetting validation message "for some reason", we loose validation.

To be frank, @bfred-it's avatar sums up this whole situation for triggering validation for autofilled input fields.

notlmn:
> The actual limit seems to be 72 characters, after which GitHub clips back to 70 characters.

I'm thinking of sticking to GitHub's convention of 70 character limit, to be consistent with GH. I can't say if this actually happens, but we might loose data when moving between 70 and 72.

bfred-it:
> I'm thinking of sticking to GitHub's convention of 70 character limit

GitHub's limit is 72, I just tested it

> But by using `setCustomValidity()`, we loose a lot of default error messages provided by the user agents. And again, with resetting validation message "for some reason", we loose validation.

Definitely, I like that the browsers say "it should be 72, but now there are 76", which we wouldn't get with `pattern` for example.

`setCustomValidity` lets the developer tell the browser whether it's valid or not, so if you pass a truthy string it will interpret it as _"the developer said it's invalid, and the string is the error to show the user."_ It's the opposite of what we want

> To be frank, @bfred-it's avatar sums up this whole situation for triggering validation for autofilled input fields.

Hehe yeah, other than trimming `sync-pr-title-commit` I don't know what else to do, setting the value programmatically breaks the validation altogether.

notlmn:
> `setCustomValidity` lets the developer tell the browser whether it's valid or not, so if you pass a truthy string it will interpret it as _"the developer said it's invalid, and the string is the error to show the user."_ It's the opposite of what we want

I should've read the spec for that one, turned out to be completely different from what I thought it would be.

---------------------

@bfred-it take a look at https://github.com/sindresorhus/refined-github/pull/2115/commits/739fcd52c491c34f06cca457407204c945dab077, as we are already limiting PR title length to 72 characters when the PR is being created there should not be many issues here. But we might also need to extend this to limit to editing PR and issue titles too.

bfred-it:
Ok, we should restore the JS validation, but it can be very simple:

On input, setValidity(field.length > limit && `The title should not be longer than 72 chars, but is currently ${field.length - limit}`)

Almost one line

notlmn:
Output for the last commit, while editing a file (tested with disabling `maxlength`):

image

bfred-it:
Uhm, the suggestion I gave you was literally one line of code inside the listener. No need to make it more complex I think

bfred-it:
> The title should be maximum 72 characters, but is 139

notlmn:
> Perhaps the same function needs to be run when the form is submitted as well, as long as it blocks the submission when it fails.
>
> Also we can drop the maxlength attribute altogether since we now have JS validation. It makes it easier to edit the text in it
>
> Likewise we can drop the PR title limit altogether. This feature will error on merge, if the merger has RGH and the feature is enabled, rather than by the PR sender

All done!

-----------------

> The title should be maximum 72 characters, but is 139

image

bfred-it:
Doesn't work correctly in Firefox: the submission is stopped but the button still changes to "Merging". We probably need to catch the click on "Merge" instead.

source/features/limit-commit-title-length.tsx
```diff
@@ -11,9 +11,14 @@ function init(): void {
...
 		const validityCallback = (event: Event): void => {
 			if (inputField.value.length > 72) {
-				inputField.setCustomValidity(`The title should be maximum 72 characters, but is ${inputField.value.length}`);
 				event.preventDefault();
 				event.stopImmediatePropagation();
+
+				inputField.setCustomValidity(`The title should be maximum 72 characters, but is ${inputField.value.length}`);
+
+				if (event.type === 'submit') {
+					inputField.reportValidity();
```
notlmn: This still fails to stop "Merge" button from being clicked only on "auto-filled values" (like when editing text-field and then the retained value on reloading the page). If the user had edited text and then clicked merge, this works as intended.

`reportValidity()` is place behind `submit` event check as doing this for each input event looks like constant annoyance.

What if we trigger an `input` event on the fields on page load!?

notlmn: 67ea2fcc20b42767ed8c34ed103247727ed29943 should fix this!

source/features/limit-commit-title-length.tsx
```diff
@@ -0,0 +1,50 @@
...
+		inputField.form!.addEventListener('submit', validityCallback);
+	}
+
+	// For PR merges, GitHub restores any saved commit messages on page load
+	// Triggering input event for these fields immediately validates the form
+	onPrMergePanelOpen(() => {
+		for (const inputField of inputs) {
+			inputField.dispatchEvent(new Event('input'));
+		}
+	});
```
bfred-it: Good idea, I wonder if this makes the `submit` handler unnecessary. I don't think the title field is ever programmatically changed when the PR merge panel is open; even GitHub specifically avoids this. 

In that case, validityCallback can be simplified as I had first suggested, likely.

Also this event only needs to be triggered on the title field because that's what this feature cares about.

notlmn: > In that case, validityCallback can be simplified as I had first suggested, likely.

Yup!

-------------

> Also this event only needs to be triggered on the title field because that's what this feature cares about.

The merge input field can be on index 0 or 1 of the `inputFields` variable when querying DOM. Instead of searching for it, I just figured it'd be easy to trigger input events on all of them. Doesn't matter that much anyway.

bfred-it: I had assumed that `fields` was `[title, message]` but I think it's always just one; the double selector made me thing it was two. I think you just need a `select` and you can drop the loop

bfred-it:
:neato: 😎

bfred-it:
Can you add this to the readme?

PR #2110, opened 2019-06-02
Format titles on single issue / PR pages

HardikModha:
Closes #1984

### Test
1. For issues: All issues present on https://github.com/HardikModha/github-playground/issues

2. For PR pages: All PRs present on https://github.com/HardikModha/github-playground/pulls

3. Also tested the feature by updating the issue / PR title. (https://github.com/HardikModha/github-playground/issues/9)

PR #2123, opened 2019-06-05
Fix branch name alignment in recently-pushed-branches

HardikModha:
Fixes https://github.com/sindresorhus/refined-github/pull/2110#issuecomment-498922194.

Issues:
- It looked misaligned.
- The branch icon was not in alignment with the text.

Before:
Screen Shot 2019-06-05 at 10 53 35 PM

After:

Screen Shot 2019-06-05 at 10 56 05 PM

### DIFF
--- a/source/features/recently-pushed-branches-enhancements.css
+++ b/source/features/recently-pushed-branches-enhancements.css
@@ -32,6 +32,10 @@
 	border-radius: 5px;
 	background-color: #24292e;
 }
+.rgh-recently-pushed-branches .RecentBranches .octicon-git-branch {
+	vertical-align: middle;
+	margin-right: 2px;
+}
 .rgh-recently-pushed-branches .RecentBranches .btn {
 	width: 32px;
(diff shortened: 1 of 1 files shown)


bfred-it:
I don't know if this is any better, now that I look at it. The branch name now is farther away from the button

HardikModha:
Hmm...Not sure but I'm also little inclined towards the right aligned text now. I was reading a bit about UX and it says, 
> The big advantage to right-aligned labels is the strong visual connection between label and input. Because items near each other appear related.

Maybe we can just drop https://github.com/sindresorhus/refined-github/pull/2123/commits/75948316a47287b05077d748f196e55ec554acea and only add https://github.com/sindresorhus/refined-github/pull/2123/commits/86883e44efd333d07ec99e901925cf746eb9537e? What do you say? The icon alignment was slightly off before.

bfred-it:
Sounds good

PR #2133, opened 2019-06-07
Only show `filter-pr-by-build-status` if the repo has CI

### LINKED ISSUE #2127: Hide status filter for PR list if there are no checks
From https://github.com/sindresorhus/refined-github/pull/2114#issuecomment-499763367.

Part of `filter-pr-by-build-status` feature, if the repository doesn't have any checks then there is no need to create and add the filter dropdown.

HardikModha:
Closes #2127

# Test
1. https://github.com/sindresorhus/refined-github/pulls (Filter should be present)
2. https://github.com/HardikModha/github-playground/pulls (Filters should not be present)

### DIFF
--- a/source/features/filter-pr-by-build-status.tsx
+++ b/source/features/filter-pr-by-build-status.tsx
@@ -36,9 +36,9 @@ function populateDropDown({currentTarget}: Event): void {
 }
 
 async function init(): Promise<void | false> {
-	const ciStatusIcon = await fetchCIStatus();
+	const hasCI = await fetchCIStatus();
 
-	if (!ciStatusIcon) {
+	if (!hasCI) {
 		return false;
 	}
 
--- a/source/features/filter-pr-by-build-status.tsx
+++ b/source/features/filter-pr-by-build-status.tsx
@@ -2,6 +2,7 @@ import React from 'dom-chef';
 import select from 'select-dom';
 import {checkInline} from '../libs/icons';
 import features from '../libs/features';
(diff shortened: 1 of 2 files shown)


source/features/filter-pr-by-build-status.tsx
```diff
@@ -34,7 +35,13 @@ function populateDropDown({currentTarget}: Event): void {
...
 
-function init(): void | false {
+async function init(): Promise<void | false> {
+	const ciStatusIcon = await fetchCIStatus();
+
+	if (!ciStatusIcon) {
+		return false;
+	}
+
 	const reviewsFilter = select('.table-list-header-toggle > details:nth-last-child(3)')!;
```
bfred-it: Since https://github.com/sindresorhus/refined-github/pull/2114 removed some filters, we might need to change the way we find the position.

It may work before this PR because this check is added before `clean-issue-filters` runs

HardikModha: Given the structure, I don't think it will be possible to detect "Reviews" filter by just using CSS. We can detect the reviews filter using JavaScript and add the build status filter just after that? What do you say?

HardikModha: Hey, "Assignee" and "Sort" filters are always there at the last and they are not getting removed. So this approach won't break.

bfred-it:
Thanks for the quick fix @HardikModha ⚡️

PR #2136, opened 2019-06-08
Align `add-asset-download-count` with asset size

### LINKED ISSUE #2126: Vertically align the asset download counts in releases
https://github.com/electron/electron/releases

Vertically align them on the icon. And with some more right margin. Would make it look cleaner. I think we could use CSS grid for this.

Screen Shot 2019-06-07 at 12 00 36

dotconnor:
Closes #2126 
image
# Test
[electron/electron/releases](https://github.com/electron/electron/releases)

### DIFF
--- a/source/content.ts
+++ b/source/content.ts
@@ -121,6 +121,6 @@ import './features/hide-readme-header.css';
 import './features/hide-obvious-tooltips.css';
 import './features/clean-discussions.css';
 import './features/clean-rich-text-editor.css';
-
+import './features/show-asset-download-count.css';
 // Add global for easier debugging
 (window as any).select = select;
--- a/source/features/show-asset-download-count.css
+++ b/source/features/show-asset-download-count.css
@@ -1,0 +1,5 @@
+.release .Box-body small:nth-child(2) {
+	min-width: 60px;
+	display: inline-block;
+	text-align: right;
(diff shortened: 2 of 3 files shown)


sindresorhus:
I think it’s should be vertically aligned by the icons, not the number.

dotconnor:
ezgif-1-2208b23d610e


1. Moves icon, download count and size all to the same parent with `align-items: center`
2. Applied `align-items: center` to all download container and parent container (this pr)
3. Moves icon, download count and size all to the same parent and removes flex, applies `vertical-align: middle` to the icon
4. Moves icon, download count and size all to the same parent and removes flex, applies `line-height: 16; height: 16px` to text

I'd go with 1 or 4

bfred-it:
See #2139

source/content.ts
```diff
@@ -121,6 +121,6 @@ import './features/hide-readme-header.css';
 import './features/hide-obvious-tooltips.css';
 import './features/clean-discussions.css';
 import './features/clean-rich-text-editor.css';
-
+import './features/show-asset-download-count.css';
```
bfred-it: By the way only CSS-only features should be added here; JS features should import their own CSS file. Refer to my implementation of this issue

PR #2105, opened 2019-05-30
Add link to forked repo below the original

jerone:
### Description
Add links to the forked repos below the original repo. For users and organizations.

### Closes
Closes #1129
Supersedes  #1108

### Screenshots
**Original repo:**
image

**User forked repo:**
image

**Organization forked repo:**
image

**Hover card:**
image

### Test
1. Open any repo and click on the <kbd>Fork</kbd> button.
2. Fork the repo.
3. Go back to the original repo. 
4. Below the repo name, a link to your fork will presented.

PR #1108, opened 2018-02-22
Add link to your fork under a repo's title

sangeeth96:
_Note by @bfred-it: I updated this comment to add screenshots and paragraphs._

Fixes: #1129

screen shot 2018-02-26 at 15 28 48

GitHub links upstream repo right below the title of our fork's title when we visit the fork's page. But it doesn't do the same thing on an upstream repo's page, i.e, link to the fork that we created. 

IMHO, it is odd that you have to go to the top-right corner and click on the big "Fork" button to switch to your fork. The intention of that button is to create a fork. If I own multiple organizations, clicking on the fork button asks me whether I want to fork under an organization or switch to the fork under my account. 

This is an added step to get to where we want and slows things down. I find it natural to expect the link to be below the title given that's how GitHub does it for a fork. This patch will add a link to our fork on upstream's page as soon as the dynamic pages are loaded.

I believe this can be extended in various ways (like a dropdown that shows forks under your orgs) but I want to know if the current implementation is fine and whether you guys think this is a good feature. The implementation currently has one main issue:

- Default pagination only returns 30 entries and I don't know if GitHub returns the details for logged-in user's fork in the first 30 entries itself.

PR #2141, opened 2019-06-09
Add link to forked repo below the original 

### LINKED ISSUE #1129: Add link to forked repo below the original
To resolve this issue, jump to: https://github.com/sindresorhus/refined-github/issues/1129#issuecomment-371044209

[Issuehunt badges](https://issuehunt.io/r/sindresorhus/refined-github/issues/1129)

Hey guys! Amazing addon with lot of awesome features 💯 

Github displays the original repo path when visiting a fork. But, the original don't have link to your fork. Right now we have to access the forks from here... 

My proposal is to add a `forked to cg-cnu/refined-github` below the original repo.

Let me know your comments and ideas. Will be happy to do a pr if you guys like it. 
Cheers 😄

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Sponsors (Total: $40.00)

### Tips

</details>

jerone:
### Description
Add links to the forked repos below the original repo. For users and organizations.

### Closes
Closes #1129
Supersedes attempt 1 #1108
Supersedes attempt 2 #2105

### Screenshots
**Original repo:**
image

**User forked repo:**
image

**Organization forked repo:**
image

~~**Hover card:**~~
image
_Feature removed_

### Test
1. Open any repo and click on the <kbd>Fork</kbd> button.
2. Fork the repo.
3. Go back to the original repo. 
4. Below the repo name, a link to your fork will presented.

### DIFF
--- a/source/features/forked-to.tsx
+++ b/source/features/forked-to.tsx
@@ -7,14 +7,6 @@ import {isOwnRepo} from '../libs/page-detect';
 
 const currentRepo = getRepoURL();
 
-async function init(): Promise<void> {
-	onForkDialogOpened();
-
-	onForkedPage();
-
-	await checkForks();
-}
-
 // Check for cached forks.
 async function checkForks(): Promise<void> {
 	const repo = getOriginalRepo();
@@ -128,6 +120,14 @@ function appendLink(fork: string): void {
 // Create the cache key.
 function key(repo: string): string {
 	return `forked-to:${repo}`;
+}
+
+async function init(): Promise<void> {
+	onForkDialogOpened();
+
+	onForkedPage();
+
+	await checkForks();
 }
 
 features.add({
--- a/source/features/forked-to.tsx
+++ b/source/features/forked-to.tsx
@@ -14,7 +14,16 @@ async function showForks(): Promise<void> {
 	const repo = getSourceRepo();
 	const cached = await getCache(repo);
 	const forks = cached.filter(fork => fork !== currentRepo);
-	appendLink(...forks);
+	const pageHeader = select<HTMLElement>('.pagehead h1.public')!;
+	for (const fork of forks) {
+		pageHeader.append(
+			<span className="fork-flag rgh-forked">
+				<span className="text">forked to&nbsp;
+					<a href={`/${fork}`}>{fork}</a>
+				</span>
+			</span>
+		);
+	}
 }
 
 // Store fork if we are on a forked page.
@@ -92,20 +101,6 @@ async function storeCache(repo: string, ...forks: string[]): Promise<void> {
 	}
 
 	await cache.set<string[]>(repoKey, cached, 10);
-}
-
-// Create a fork link.
-function appendLink(...forks: string[]): void {
-	const pageHeader = select<HTMLElement>('.pagehead h1.public')!;
-	for (const fork of forks) {
-		pageHeader.append(
-			<span className="fork-flag rgh-forked">
-				<span className="text">forked to&nbsp;
-					<a href={`/${fork}`}>{fork}</a>
-				</span>
-			</span>
-		);
-	}
 }
 
 async function init(): Promise<void> {
--- a/source/features/forked-to.tsx
+++ b/source/features/forked-to.tsx
@@ -1,8 +1,8 @@
 import './more-dropdown.css';
 import React from 'dom-chef';
-import cache from '../libs/cache';
 import select from 'select-dom';
 import features from '../libs/features';
+import cache from '../libs/cache';
 import {getRepoURL} from '../libs/utils';
 import {isOwnRepo} from '../libs/page-detect';
 
--- a/source/features/forked-to.tsx
+++ b/source/features/forked-to.tsx
@@ -1,10 +1,10 @@
 import './more-dropdown.css';
 import React from 'dom-chef';
+import cache from '../libs/cache';
 import select from 'select-dom';
 import features from '../libs/features';
 import {getRepoURL} from '../libs/utils';
 import {isOwnRepo} from '../libs/page-detect';
-import cache from '../libs/cache';
 
 const currentRepo = getRepoURL();
 
@@ -18,21 +18,14 @@ async function init(): Promise<void> {
 
 // Check for cached forks.
 async function checkForks(): Promise<void> {
-	let repoKey = key(currentRepo);
-
-	// Check if we are already on a fork.
-	const forkedFromElm = select<HTMLElement>('.fork-flag:not(.ghr-forked) a');
-	if (forkedFromElm) {
-		const forkedRepo = forkedFromElm.getAttribute('href')!.substring(1);
-		repoKey = key(forkedRepo);
-	}
-
+	const repo = getOriginalRepo();
+	const repoKey = key(repo);
 	const cached = await cache.get<string[]>(repoKey) || [];
 	const validForks = cached.filter(validateFork);
 	for (const fork of await Promise.all(validForks)) {
 		if (fork !== currentRepo) {
 			appendHtml(fork);
-			storeCache(currentRepo, fork);
+			storeCache(repo, fork);
 		}
 	}
 }
@@ -69,15 +62,29 @@ function onForkDialogOpened(): void {
 
 // Event called when fork dialog is opened.
 function onFragmentLoaded(parent: HTMLElement): void {
+	const repo = getOriginalRepo();
 	for (const forkElm of select.all<HTMLElement>('.octicon-repo-forked', parent)) {
 		const fork = forkElm.parentNode!.textContent!.trim();
 		appendHtml(fork);
-		storeCache(currentRepo, fork);
+		storeCache(repo, fork);
 	}
+}
+
+// Get the original repo, by checking if we are already on a fork.
+function getOriginalRepo(): string {
+	let repo = currentRepo;
+
+	const forkedFromElm = select<HTMLElement>('.fork-flag:not(.ghr-forked) a');
+	if (forkedFromElm) {
+		repo = forkedFromElm.getAttribute('href')!.substring(1);
+	}
+
+	return repo;
 }
 
 // Save forks to cache.
 async function storeCache(repo: string, fork: string): Promise<void> {
+	console.log('storeCache', arguments);
 	const repoKey = key(repo);
 	const cached = await cache.get<string[]>(repoKey) || [];
 	if (!cached.includes(fork)) {
@@ -103,7 +110,7 @@ function appendHtml(fork: string): void {
 
 // Create the cache key.
 function key(repo: string): string {
-	return `forked-to:${repo}`;
+	return `forked-to5:${repo}`;
 }
 
 features.add({
--- a/source/features/forked-to.tsx
+++ b/source/features/forked-to.tsx
@@ -3,6 +3,7 @@ import React from 'dom-chef';
 import select from 'select-dom';
 import features from '../libs/features';
 import {getRepoURL} from '../libs/utils';
+import {isOwnRepo} from '../libs/page-detect';
 import cache from '../libs/cache';
 
 const currentRepo = getRepoURL();
@@ -10,7 +11,9 @@ const currentRepo = getRepoURL();
 async function init(): Promise<void> {
 	checkForks();
 	onForkDialogOpened();
-	onForkedPage();
+	if (isOwnRepo()) {
+		onForkedPage();
+	}
 }
 
 // Check for cached forks.
--- a/source/features/forked-to.tsx
+++ b/source/features/forked-to.tsx
@@ -5,17 +5,12 @@ import cache from '../libs/cache';
 import {getRepoURL} from '../libs/utils';
 import {isOwnRepo} from '../libs/page-detect';
 
-const id = 'forked-to';
+const getCacheKey = (repo: string): string => `forked-to:${repo}`;
 
-const currentRepo = getRepoURL();
-
-// Show cached forks.
 async function showForks(): Promise<void> {
-	const repo = getSourceRepo();
-	const cached = await getCache(repo);
-	const forks = cached.filter(fork => fork !== currentRepo);
-	const pageHeader = select<HTMLElement>('.pagehead h1.public')!;
-	for (const fork of forks) {
+	const cached = await getValidatedCache(getSourceRepo());
+	const pageHeader = select('.pagehead h1.public')!;
+	for (const fork of cached.filter(fork => fork !== getRepoURL())) {
 		pageHeader.append(
 			<span className="fork-flag rgh-forked">
 				<span className="text">forked to&nbsp;
@@ -26,20 +21,18 @@ async function showForks(): Promise<void> {
 	}
 }
 
-// Store fork if we are on a forked page.
-function checkForForkedPage(): void {
+function rememberCurrentFork(): void {
 	if (!isOwnRepo()) {
 		return;
 	}
 
-	const forkSourceElement = select<HTMLElement>('.fork-flag:not(.rgh-forked) a');
+	const forkSourceElement = select<HTMLAnchorElement>('.fork-flag:not(.rgh-forked) a');
 	if (forkSourceElement) {
-		const forkedRepo = forkSourceElement.getAttribute('href')!.substring(1);
-		storeCache(forkedRepo, currentRepo);
+		const forkedRepo = forkSourceElement.pathname.substring(1);
+		addAndStoreCache(forkedRepo, getRepoURL());
 	}
 }
 
-// Watch for opening the fork dialog.
 function watchForkDialog(): void {
 	const forkDialog = select<HTMLElement>('details-dialog[src*="/fork"]')!;
 	const forkFragment = select<HTMLElement>('include-fragment', forkDialog)!;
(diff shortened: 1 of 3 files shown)


jerone:
@bfred-it Before I going to fix every linting error, can you review if this satisfies the way you like to see this implemented.

jerone:
* When you visit a repo
  1. ...checks for cached forks.
  2. ...validates if forks still exists.
  3. ...add fork HTML.
* When you visit a fork
  1. ... it caches the fork.
* When you open the fork dialog
  1. ...it caches all forks.

bfred-it:
Looks about right, the only issue is ensuring that the forked page you visit is actually **your** fork, so you might need to look for the `Settings` tab or something like that to double check.

jerone:
@bfred-it Ok, this is ready for review. 👀

source/features/forked-to.tsx
```diff
@@ -0,0 +1,154 @@
...
+		repo = forkedFromElm.getAttribute('href')!.substring(1);
+	}
+
+	return repo;
+}
+
+// Get cache and sort it.
+async function getCache(repo: string): Promise<string[]> {
+	const currentUser = getUsername();
+	const repoKey = key(repo);
```
bfred-it: key() is only used once. Let’s inline it

source/features/forked-to.tsx
```diff
@@ -0,0 +1,154 @@
...
+	const cached = await cache.get<string[]>(repoKey) || [];
+	cached.sort((a, b) => {
+		let order = a.localeCompare(b);
+		if (a.startsWith(`${currentUser}/`)) {
+			order -= 100;
```
bfred-it: I don’t think that’s how it works, it should only return -1, 0, or 1. This only works because the function always returns less than 0. 

Anyway: we probably shouldn’t assume that the user’s fork has priority, we can probably avoid this check altogether.

source/features/forked-to.tsx
```diff
@@ -0,0 +1,154 @@
...
+		}
+
+		return order;
+	});
+	return Promise.resolve(cached);
```
bfred-it: Async functions always return promises, no need to wrap `cached`

source/features/forked-to.tsx
```diff
@@ -0,0 +1,154 @@
...
+	const forks = select.all<HTMLElement>('.rgh-forked');
+	for (const fork of forks) {
+		fork.remove();
+	}
+}
+
+// Create the HTML.
+function appendHtml(fork: string): void {
+	const pageHeader = select<HTMLElement>('.pagehead h1.public')!;
+	pageHeader.append(
```
bfred-it: Let’s avoid this variable. Prefer `select().append(...)`

source/features/forked-to.tsx
```diff
@@ -0,0 +1,154 @@
...
+import React from 'dom-chef';
+import select from 'select-dom';
+import features from '../libs/features';
+import cache from '../libs/cache';
+import {getRepoURL, getUsername} from '../libs/utils';
+import {isOwnRepo} from '../libs/page-detect';
+
+const currentRepo = getRepoURL();
+
+async function init(): Promise<void> {
```
bfred-it: Can you move init before features.add? All features are like that. I tend to follow this rule even if it’s not enforced automatically

https://eslint.org/docs/2.0.0/rules/no-use-before-define

bfred-it:
This can be cleaned up further

source/features/forked-to.tsx
```diff
@@ -0,0 +1,154 @@
...
+	}
+}
+
+// Create the HTML.
+function appendHtml(fork: string): void {
+	const pageHeader = select<HTMLElement>('.pagehead h1.public')!;
+	pageHeader.append(
+		<span className={'fork-flag rgh-forked'} data-repository-hovercards-enabled>
+			<span className={'text'}>forked to&nbsp;
+				<a data-hovercard-type="repository" data-hovercard-url={`/${fork}/hovercard`} href={`/${fork}`}>
```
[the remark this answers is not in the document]
jerone: Probably true, but it doesn't hurt either. I can remove.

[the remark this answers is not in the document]
bfred-it: Visual noise and HTTP requests, I’d drop it

source/features/forked-to.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+// Check for cached forks.
+async function checkForks(): Promise<void> {
+	const repo = getOriginalRepo();
+	const cached = await getCache(repo);
+	const validForks = cached.filter(validateFork);
+	for (const fork of await Promise.all(validForks)) {
+		if (fork !== currentRepo) {
+			appendHtml(fork);
+			storeCache(repo, fork);
```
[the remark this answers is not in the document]
jerone: It's a key per original repo. The forks are in array as value.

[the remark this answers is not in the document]
bfred-it: Indeed, but each loop is getting the whole array, adding one key and saving it. 

And: you’re not awaiting storeCache so all of these requests might happen not in the order you want.

And: I don’t think invalidated forks are removed from the cache.

I think it’s best to validate the whole cache and just store the whole array as-is, instead of merging it with the existing one. It’s safer.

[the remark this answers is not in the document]
jerone: I'm on it.

jerone:
@bfred-it Feedback round 1 done.

source/features/forked-to.tsx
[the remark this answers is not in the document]
bfred-it: This wasn’t resolved. You’re still merging the array with the cached one. You probably don’t need the helper function at all, just store the validated cache into the cached

source/features/forked-to.tsx
```diff
@@ -0,0 +1,154 @@
...
+// Check for opening the fork dialog.
+function onForkDialogOpened(): void {
+	const forkDialog = select<HTMLElement>('details-dialog[src*="/fork"]')!;
+	const forkFragment = select<HTMLElement>('include-fragment', forkDialog)!;
+	forkFragment.addEventListener('load', () => onFragmentLoaded(forkDialog));
+}
+
+// Event called when fork dialog is opened.
+function onFragmentLoaded(parent: HTMLElement): void {
+	removeAllHtml();
```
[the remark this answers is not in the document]
bfred-it: The second part wasn’t done

source/features/forked-to.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+	storeCache(repo, ...forks);
+}
+
+// Get the original repo, by checking if we are already on a fork.
+function getOriginalRepo(): string {
+	let repo = currentRepo;
+
+	const forkedFromElm = select<HTMLElement>('.fork-flag:not(.rgh-forked) a');
+	if (forkedFromElm) {
```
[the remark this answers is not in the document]
bfred-it: You’re still setting the variable. Just return the value instead

source/features/forked-to.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+	const forkSourceElement = select<HTMLElement>('.fork-flag:not(.rgh-forked) a');
+	if (forkSourceElement) {
+		const forkedRepo = forkSourceElement.getAttribute('href')!.substring(1);
+		storeCache(forkedRepo, currentRepo);
+	}
+}
+
+// Check for opening the fork dialog.
+function onForkDialogOpened(): void {
```
bfred-it: This isn’t a listener, it should be a verb like `watchForkDialog`

source/features/forked-to.tsx
```diff
@@ -0,0 +1,154 @@
...
+				method: 'HEAD'
+			});
+		return response.ok;
+	} catch (error) {
+		return false;
+	}
+}
+
+// Check if we are on a forked page.
+function onForkedPage(): void {
```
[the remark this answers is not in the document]
bfred-it: This isn’t a listener and doesn’t just “check”; it stores the fork, name the function accordingly

source/features/forked-to.tsx
```diff
@@ -0,0 +1,145 @@
...
+		});
+	return response.ok;
+}
+
+// Get cache and sort it.
+async function getCache(repo: string): Promise<string[]> {
+	const repoKey = key(repo);
+	const cached = await cache.get<string[]>(repoKey) || [];
+	const validForks = cached.filter(validateFork);
+	validForks.sort(undefined);
```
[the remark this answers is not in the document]
jerone: Because of `@typescript-eslint/require-array-sort-compare`.

`undefined` is the [explicit default behavior](https://github.com/typescript-eslint/typescript-eslint/issues/247).

source/features/forked-to.tsx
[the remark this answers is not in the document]
jerone: Not every repo has this element.

source/features/forked-to.tsx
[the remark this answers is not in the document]
jerone: Merging the array after filtering is done here: https://github.com/sindresorhus/refined-github/pull/2141/files#diff-ff16301f232100bd87e910eb98614f66R78

jerone:
Brought it back to 122 LOC.

jerone:
Found one issue, that I'm not sure is related to this new feature...
1. Go to the [global PR's](https://github.com/pulls).
2. Find any PR repo that you know has forks.
3. See no forks being loaded!
4. Navigate to **issues** or any other page on the repo (or refresh page) and it works.

I can only reproduce the issue via above route. Any other route, the forks are loaded correctly.

source/features/forked-to.tsx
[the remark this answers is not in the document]
bfred-it: Sort returns the array, leaving it on one line after filter makes it more clear

source/features/forked-to.tsx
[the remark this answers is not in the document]
bfred-it: You’re misunderstanding me. You can replace `repo = ` with `return` and it will work.

jerone:
@bfred-it All feedback resolved.

source/features/forked-to.tsx
```diff
@@ -0,0 +1,117 @@
...
+		return forkSourceElement.getAttribute('href')!.substring(1);
+	}
+
+	return currentRepo;
+}
+
+// Check if the fork still exists.
+async function validateFork(repo: string): Promise<boolean> {
+	const url = new URL(`/${repo}`, location.href);
+	const response = await fetch(String(url),
```
bfred-it:
```suggestion
	const response = await fetch(location.origin + '/' + repo, {
	    method: 'HEAD'
	});
```

source/features/forked-to.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+// Store fork if we are on a forked page.
+function checkForForkedPage(): void {
+	if (!isOwnRepo()) {
+		return;
+	}
+
+	const forkSourceElement = select<HTMLElement>('.fork-flag:not(.rgh-forked) a');
+	if (forkSourceElement) {
+		const forkedRepo = forkSourceElement.getAttribute('href')!.substring(1);
```
bfred-it:
```suggestion
		const forkedRepo = forkSourceElement.pathname.substring(1);
```

Maybe. 

Also getSourceRepo is nearly identical. The code could be deduplicated.

source/features/forked-to.tsx
```diff
@@ -0,0 +1,117 @@
...
+	}
+}
+
+// Store fork if we are on a forked page.
+function checkForForkedPage(): void {
+	if (!isOwnRepo()) {
+		return;
+	}
+
+	const forkSourceElement = select<HTMLElement>('.fork-flag:not(.rgh-forked) a');
```
bfred-it: HTMLElement is the default. No need to set it. This however enables my next comment. 
```suggestion
	const forkSourceElement = select<HTMLAnchorElement>('.fork-flag:not(.rgh-forked) a');
```

source/features/forked-to.tsx
```diff
@@ -0,0 +1,117 @@
...
+// Show cached forks.
+async function showForks(): Promise<void> {
+	const repo = getSourceRepo();
+	const cached = await getCache(repo);
+	const forks = cached.filter(fork => fork !== currentRepo);
+	const pageHeader = select<HTMLElement>('.pagehead h1.public')!;
+	for (const fork of forks) {
+		pageHeader.append(
+			<span className="fork-flag rgh-forked">
+				<span className="text">forked to&nbsp;
```
bfred-it: Are these two spans required or can any be dropped? What’s this class?

source/features/forked-to.tsx
```diff
@@ -0,0 +1,117 @@
...
+const id = 'forked-to';
+
+const currentRepo = getRepoURL();
+
+// Show cached forks.
+async function showForks(): Promise<void> {
+	const repo = getSourceRepo();
+	const cached = await getCache(repo);
+	const forks = cached.filter(fork => fork !== currentRepo);
+	const pageHeader = select<HTMLElement>('.pagehead h1.public')!;
```
bfred-it:
```suggestion
	const pageHeader = select('.pagehead h1.public')!;
```

It’s the default

source/features/forked-to.tsx
```diff
@@ -0,0 +1,117 @@
+import React from 'dom-chef';
+import select from 'select-dom';
+import features from '../libs/features';
+import cache from '../libs/cache';
+import {getRepoURL} from '../libs/utils';
+import {isOwnRepo} from '../libs/page-detect';
+
+const id = 'forked-to';
```
bfred-it: If you need a global you might as well turn it into a function to get the full ID, I made a mistake earlier suggesting that it was only used in one place. 

Also use proper var names, I don’t know what `id` means

```suggestion
const getCacheKey = repo => 'forked-to:' + repo;
```

source/features/forked-to.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+const currentRepo = getRepoURL();
```
bfred-it: I think this function is already cached. You can use it online without further variables.

source/features/forked-to.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+// Show cached forks.
+async function showForks(): Promise<void> {
+	const repo = getSourceRepo();
+	const cached = await getCache(repo);
```
bfred-it: You can avoid the extra const
```suggestion
	const cached = await getCache(getSourceRepo());
```

There’s suggestions probably apply to other parts of the PR as well

source/features/forked-to.tsx
```diff
@@ -0,0 +1,117 @@
...
+				<span className="text">forked to&nbsp;
+					<a href={`/${fork}`}>{fork}</a>
+				</span>
+			</span>
+		);
+	}
+}
+
+// Store fork if we are on a forked page.
+function checkForForkedPage(): void {
```
bfred-it: Instead of adding a comment, name the function appropriately, like
```suggestion
function rememberCurrentRepoIfFork(): void {
```

Or

```suggestion
function rememberCurrentFork(): void {
```

source/features/forked-to.tsx
```diff
@@ -0,0 +1,117 @@
...
+	forkFragment.addEventListener('load', () => {
+		const repo = getSourceRepo();
+		const forks = select.all<HTMLElement>('.octicon-repo-forked', forkDialog).map(forkElement => {
+			return forkElement.parentNode!.textContent!.trim();
+		});
+		storeCache(repo, ...forks);
+	});
+}
+
+// Get the source repo, by checking if we are already on a fork.
```
bfred-it: Redundant comment, the function name is clear enough

source/features/forked-to.tsx
```diff
@@ -0,0 +1,117 @@
...
+	const validForks = cached.filter(validateFork).sort(undefined);
+
+	if (cached.length !== validForks.length) {
+		await cache.set<string[]>(repoKey, validForks, 10);
+	}
+
+	return validForks;
+}
+
+// Save forks to cache.
```
bfred-it: Redundant comment. Also function name can be improved since this doesn’t simply store it, it *adds* the forks to the existing cache.

source/features/forked-to.tsx
jerone: In `utils` its not cached, but then again it's not an intensive function. 👍

jerone:
@bfred-it Feedback processed again.

source/features/forked-to.tsx
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	const forkedRepo = findForkedRepo();
+	if (forkedRepo) {
+		addAndStoreCache(forkedRepo, getRepoURL());
+	}
+}
+
+function watchForkDialog(): void {
+	const forkDialog = select<HTMLElement>('details-dialog[src*="/fork"]')!;
```
bfred-it: All the feedback I sent before applies to other parts of the file as well. I can’t review every single line. This PR has been sloppy. Please clean it up and open a new PR

jerone: Webpack compiles, lint succeeds, the PR does exactly what it should do, in the way you specified it should be done, without code style guide. I try to follow other features syntax, but you have a very opinionated way of writing code. I'm not saying that that is bad, and the feedback is great.

See https://github.com/sindresorhus/refined-github/pull/2153

sindresorhus: I haven't followed the review here, but what @bfred-it means is that when he reviews some change, and you fix it, you should look for other places where the same fix or similar fix could be applied. You should also look if there's any way to improve/simplify the code after the changes.

PR #2145, opened 2019-06-10
Format selected text in textareas with keyboard shortcuts

### LINKED ISSUE #2032: Shortcut to format selected text as code
There are already shortcuts for making text bold, italicized and linked in comment boxes. One formatting style that I use all the time (perhaps even more than bold and italics) is code, using backticks (``` `...` ```).

As someone who uses a keyboard layout where <kbd>\`</kbd> is a dead key (so it allows typing accented letters like `à`, etc.), I find myself wanting a shortcut pretty often.

Even for those using keyboards where it works as a normal key, I can imagine that selecting some text and pressing a shortcut would be more convenient than moving the cursor to the start of the intended selection, typing <kbd>\`</kbd>, then moving it to the end of the selection and typing <kbd>\`</kbd> again.

(The same could be done for quoting selections with `'` or `"`, by the way, but that's a separate request.)

notlmn:
Closes #2032.

If any text is selected while using these keys, the text is wrapped with each of the pressed characters. Text selection is still preserved so that the user can chain these key presses.

## Demo

format-textarea

## Test

On any comment textarea for commits or comments.


jerone:
I saw characters `[` and `(` in the list. Does that mean if I press `[`, it will add two `[[`, instead of `[]`?

notlmn:
> I saw characters `[` and `(` in the list. Does that mean if I press `[`, it will add two `[[`, instead of `[]`?

Aah, that's right. Forgot to push some changes last night. Thanks for pointing that out.

source/features/comment-fields-keyboard-shortcuts.tsx
```diff
@@ -75,6 +80,24 @@ function init(): void {
 					select<HTMLTextAreaElement>('.js-comment-field', lastOwnComment)!.selectionStart = Number.MAX_SAFE_INTEGER;
 				});
 			}
+		} else if (formattingCharacters.includes(event.key)) {
+			const formattingChar = event.key;
+			const [start, end] = [field.selectionStart, field.selectionEnd];
+
+			// If `start` and `end` of selection are the same, then no text is selected
+			// For keys that require selection, if no text is selected, bail immediately
```
bfred-it: I don’t want to actually make a text editor in the comment field, I think this feature should just take care of selected text, *if any*, not automatically insert two characters every time you press a key without selection

notlmn: I see it as a small addition to what we can do in a textarea with little to no effort, like not having to press `"`, `` ` `` two times when I can press it only once.

If you say that it might be bit confusing or might looks like we are overdoing things, then I wouldn't mind dropping that specific change.

bfred-it: Yeah because in the editor if you delete the character you entered it will also delete the nearby matching character, for example.

Wrapping selected text saves a lot of keystrokes, so it’s worth adding it. Adding a matching character without selection saves one stroke, we can avoid it.

notlmn: Fixed!

notlmn:
> Are any shortcuts in this file mentioned there?

This PR doesn't exactly add keyboard shortcuts, I don't know if we can call these actions "keyboard shortcuts".

------------------

> Should this be mentioned in the readme?

Even if we added it to the readme, what should we call it? ¯\\\_(ツ)\_/¯

sindresorhus:
Yes, should be mentioned in the readme as it's not something most would just discover.

bfred-it:
If you want to extract it into its own feature it could be `easy-formatting` or `one-key-formatting` maybe. This file has a few unrelated features, like “esc to blur” and “esc to cancel” I think.

notlmn:
> This file has a few unrelated features, like “esc to blur” and “esc to cancel” I think.

Last time I checked, some of these sub-features weren't working, but that's another issue anyway.

There's still the question of what should this be called when including in readme, keyboard shortcut? Or if this should be extracted into its own feature.

bfred-it:
“Format selected text with a key press when writing comments”

Perhaps. Not sure about “key press”. Is there an equivalent to “in a click”?

PatricNox:
"Faster text formatting"

surround your word or phrase with *asterisks.* To italicize text, place _underscores_ around a section.

This can also be achieved through highlighting the specific word, then pressing the underscore symbol on your keyboard to automatically format the text.




________

Something like this, maybe?

bfred-it:
The formatting is part of GitHub already, so we don’t need to explain that part.

PR #2146, opened 2019-06-10
Add `minimize-user-comments` feature

### LINKED ISSUE #1629: Add ability to mute comments from certain users
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/refined-github/issues/1629)

There are certain users that do no harm, so I don't want to block them, but they generally produce low-quality comments and I just don't want to see it. Would be nice to be able to hide comments from them. This is different from blocking a user in that they won't know that I'm muting them and it would work on other people's repo too.

It would be a text field in the settings that accepts a list of usernames.

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Sponsors (Total: $50.00)

- [issuehunt issuehunt](https://issuehunt.io/u/issuehunt) ($50.00)

### Tips

</details>

notlmn:
Closes #1629.

- [x] Let's you mute or unmute users from the dropdown available for comments. Comments from these muted users will be minimized on page load.
- [x] As mentioned in https://github.com/sindresorhus/refined-github/issues/1629#issuecomment-441348474, the original post, issues, and PRs from muted users will still be visible, but any of their comments will be minimized.
- [x] Shows list of users who are muted in the options.
- [x] Disable self-muting.
- [ ] Possibly show on a users profile if you are muting them.

### A minimized comment
image

### Unmute button
image

### Mute button
image

### Options menu
image

# Test
On any issue or PR.


source/features/hide-useless-comments.tsx
```diff
@@ -55,11 +57,112 @@ function unhide(event: React.MouseEvent<HTMLButtonElement>): void {
...
+					className="dropdown-item btn-link rgh-mute-unmute-button"
+					role="menuitem"
+					title="Browse repository like it appeared on this day"
+					onClick={onMuteUnmuteClick}>
+					{isMutedUser ? 'Unmute' : 'Mute'} user
+				</button>
+			);
+		}
+
+		// If the use _is_ muted, minimize their comment
```
jerone: `use` -> `user`

bfred-it:
Should the comment be hidden by `hide-useless-comments` instead of being minimized?

That’s how muting generally works.

notlmn:
> Should the comment be hidden by `hide-useless-comments` instead of being minimized inline?

Looking at the feature name, it should be hidden and not minimized. But if a user were having a conversation with another user muted by the viewer then it would not make any sense on what is happening. But if there were a minimized comment then that would mean the they are having a conversation.

And if the comments were hidden inline, then the user can individually expand that specific comment, unmute the user then and there without having to go look in the options section and try to see who commented on the issue/PR by individually unmuting each user.

-------------

> If not, then this is a completely different feature from `hide-useless-comments` and should be separate.

That's what I initially thought when the code grew a bit large than what it was before. WDYT? Should I move it to something like `minimize-muted-user-comments`?

bfred-it:
> And if the comments were hidden inline, then the user can individually expand that specific comment, unmute the user then and there without having to go look in the options section and try to see who commented on the issue/PR by individually unmuting each user.

To see the hidden comment you’d click the regular link at the bottom: “3 unhelpful comments were hidden”

I never needed this feature so let’s see what @sindresorhus expects from it

sindresorhus:
> Should the comment be hidden by hide-useless-comments instead of being minimized inline?

It should be minimized.

bfred-it:
Then should be extracts into `mute-users` because there’s no overlap.

source/options.html
```diff
@@ -45,6 +45,16 @@ <h2 class="only-firefox">Options</h2>
 
 	<hr>
 
+	<p>
+			<label>
+				<strong>Muted users</strong>
```
bfred-it: The feature that does this needs to be clear; perhaps:

> Users muted by `muted-users` feature 

Ideally the field would appear in the feature list, but I think that’s a little more complicated. Perhaps the field can be moved from here to the list via JS, right after the list is generated. 

The label should then just be “Users”

source/features/hide-useless-comments.tsx
```diff
@@ -55,11 +57,112 @@ function unhide(event: React.MouseEvent<HTMLButtonElement>): void {
...
+		const isMutedUser = mutedUsers.includes(user);
+
+		const dropdowns = select.all('.show-more-popover', comment);
+		for (const dropdown of dropdowns) {
+			// Add option to mute or unmute user
+			appendBefore(dropdown, 'a[data-ga-click^="Report"]',
+				<button
+					className="dropdown-item btn-link rgh-mute-unmute-button"
+					role="menuitem"
+					title="Browse repository like it appeared on this day"
```
bfred-it: 👆

source/features/hide-useless-comments.tsx
```diff
@@ -55,11 +57,112 @@ function unhide(event: React.MouseEvent<HTMLButtonElement>): void {
 	event.currentTarget.parentElement!.remove();
 }
 
+async function getMutedUsers(): Promise<string[]> {
+	return ((await optionsStorage.getAll() as Options).mutedUsers).split(' ');
```
bfred-it: 👆

source/features/hide-useless-comments.tsx
```diff
@@ -55,11 +57,112 @@ function unhide(event: React.MouseEvent<HTMLButtonElement>): void {
...
+	select('.unminimized-comment', comment)!.classList.add('d-none');
+}
+
+function unminimizeComment(comment: HTMLElement): void {
+	if (select.exists('.js-targetable-comment[id^="issue-"]', comment)) {
+		return;
+	}
+
+	select('.minimized-comment', comment)!.classList.add('d-none');
+	select('.minimized-comment .Details-element', comment)!.removeAttribute('open');
```
bfred-it: This line is identical in `minimizeComment`

Also is it possible to deduplicate all minimize/unminimize code? e.g. “toggleComment(comment, boolean)”

source/features/hide-useless-comments.tsx
```diff
@@ -55,11 +57,112 @@ function unhide(event: React.MouseEvent<HTMLButtonElement>): void {
...
+		const newOffset = comment.getBoundingClientRect().top;
+		window.scrollBy(0, newOffset - viewportOffset);
+	});
+}
+
+async function minimizeMutedUserComments(): Promise<void> {
+	const mutedUsers = await getMutedUsers();
+
+	const comments = select.all('.js-discussion .js-comment-container');
+	for (const comment of comments) {
```
bfred-it: When the selector isn’t super long, you can skip the extra variable and just write “for const comment of select.all”

source/features/hide-useless-comments.tsx
```diff
@@ -55,11 +57,112 @@ function unhide(event: React.MouseEvent<HTMLButtonElement>): void {
...
+	const viewportOffset = comment.getBoundingClientRect().top;
+	const user = select('.author', comment)!.textContent!;
+
+	if (mutedUsers.includes(user)) {
+		mutedUsers = mutedUsers.filter(mutedUser => mutedUser !== user);
+	} else {
+		mutedUsers.push(user);
+	}
+
+	await setMutedUsers(mutedUsers);
```
bfred-it: We don’t need to await this right here, it can be done later.

source/features/hide-useless-comments.tsx
```diff
@@ -55,11 +57,112 @@ function unhide(event: React.MouseEvent<HTMLButtonElement>): void {
...
+			for (const button of select.all('.rgh-mute-unmute-button', comment)) {
+				button.textContent = 'Unmute user';
+			}
+		} else {
+			unminimizeComment(comment);
+			for (const button of select.all('.rgh-mute-unmute-button', comment)) {
+				button.textContent = 'Mute user';
+			}
+		}
+	}
```
bfred-it: `minimizeMutedUserComments` can probably *add or update* the “mute user” menu item and this code can be dropped.

source/features/hide-useless-comments.tsx
```diff
@@ -55,11 +57,112 @@ function unhide(event: React.MouseEvent<HTMLButtonElement>): void {
...
+
+	const comments = select.all('.js-discussion .js-comment-container');
+	for (const comment of comments) {
+		const user = select('.author', comment)!.textContent!;
+		const isMutedUser = mutedUsers.includes(user);
+
+		const dropdowns = select.all('.show-more-popover', comment);
+		for (const dropdown of dropdowns) {
+			// Add option to mute or unmute user
+			appendBefore(dropdown, 'a[data-ga-click^="Report"]',
```
bfred-it: Like we do in `faster-comment-minimize` (I don’t remember the actual feature name) we might be able to add the menu item **only when the user clicks on dropdown**, on demand, to avoid changing the DOM of every comment for a feature that’s very rarely used. 

This means would split this function in two parts:

- hide muted comments on load
- add menu item on dropdown open

source/features/hide-useless-comments.tsx
```diff
@@ -55,11 +57,112 @@ function unhide(event: React.MouseEvent<HTMLButtonElement>): void {
...
+			for (const button of select.all('.rgh-mute-unmute-button', comment)) {
+				button.textContent = 'Mute user';
+			}
+		}
+	}
+
+	requestAnimationFrame(() => {
+		const newOffset = comment.getBoundingClientRect().top;
+		window.scrollBy(0, newOffset - viewportOffset);
+	});
```
bfred-it: All hiding can probably benefit from this. Some threads are long and sometimes I have already scrolled before the DOM ready event.

source/features/hide-useless-comments.tsx
```diff
@@ -55,11 +57,112 @@ function unhide(event: React.MouseEvent<HTMLButtonElement>): void {
...
+
+	select('.minimized-comment', comment)!.classList.add('d-none');
+	select('.minimized-comment .Details-element', comment)!.removeAttribute('open');
+	select('.unminimized-comment', comment)!.classList.remove('d-none');
+}
+
+async function onMuteUnmuteClick(event: React.MouseEvent<HTMLButtonElement>): Promise<void> {
+	let mutedUsers = await getMutedUsers();
+
+	const comment = (event.target as HTMLElement).closest('.js-comment-container')!;
```
bfred-it: Prefer currentTarget, I think it’s already an element

source/options.html
notlmn: image

???

source/features/hide-useless-comments.tsx
notlmn: > Also is it possible to deduplicate all minimize/unminimize code? e.g. “toggleComment(comment, boolean)”

We explicitly require `minimizeComment` at feature's initial run.

source/options.html
bfred-it: Seems redundant. Either Users or “User list” I think would suffice

source/features/hide-useless-comments.tsx
```diff
@@ -55,11 +55,16 @@ function unhide(event: React.MouseEvent<HTMLButtonElement>): void {
...
+	hideUselessComments();
+}
+
 features.add({
 	id: 'hide-useless-comments',
 	description: 'Hide useless comments like "+1"',
 	include: [
-		features.isIssue
+		features.isIssue,
+		features.isPRConversation
```
bfred-it: All the changes to this file should be undone. PRs were explicitly avoided because “👍” might be an acceptable response to PRs

source/features/mute-users.tsx
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	requestAnimationFrame(() => {
+		const newOffset = comment.getBoundingClientRect().top;
+		window.scrollBy(0, newOffset - viewportOffset);
+	});
+}
+
+async function handleMenuOpening(event: DelegateEvent): Promise<void> {
+	const dropdown = select('.show-more-popover', event.delegateTarget.parentElement!)!;
```
bfred-it: I think ! isn’t needed in the second parameter of select. Same thing on the next line

source/features/mute-users.tsx
```diff
@@ -0,0 +1,134 @@
...
+			minimizeComment(comment);
+		} else {
+			unminimizeComment(comment);
+		}
+	}
+
+	requestAnimationFrame(() => {
+		const newOffset = comment.getBoundingClientRect().top;
+		window.scrollBy(0, newOffset - viewportOffset);
+	});
```
bfred-it: Previous review 👆 

Everything after setMutedComments should not be here. Just call minimizeMutedUserComments

source/features/mute-users.tsx
```diff
@@ -0,0 +1,134 @@
...
+	const viewportOffset = comment.getBoundingClientRect().top;
+	const user = select('.author', comment)!.textContent!;
+
+	if (mutedUsers.includes(user)) {
+		mutedUsers = mutedUsers.filter(mutedUser => mutedUser !== user);
+	} else {
+		mutedUsers.push(user);
+	}
+
+	await setMutedUsers(mutedUsers);
```
bfred-it: This is only used once. You can drop the function and use `optionsStorage.set` directly.

source/features/mute-users.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+function unminimizeComment(comment: HTMLElement): void {
+	if (select.exists('.js-targetable-comment[id^="issue-"]', comment)) {
+		return;
+	}
+
+	select('.minimized-comment', comment)!.classList.add('d-none');
+	select('.minimized-comment .Details-element', comment)!.setAttribute('open', 'true');
+	select('.unminimized-comment', comment)!.classList.remove('d-none');
+}
```
bfred-it: Previous review. 

These two functions can be merged into `toggleComment`. The methods can be called as (example):

```is
select('.minimized-comment', comment)!.classList.toggle('d-none', toggle);
select('.minimized-comment .Details-element', comment)![toggle ? 'setAttribute' : 'remove']('open', 'true');
``` 

If TS gives you grief on the second line, use a regular IF

source/features/mute-users.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+async function onMuteUnmuteClick(event: React.MouseEvent<HTMLButtonElement>): Promise<void> {
+	let mutedUsers = await getMutedUsers();
+
+	const comment = (event.target as HTMLElement).closest('.js-comment-container')!;
```
bfred-it: 👆 previous review

source/features/mute-users.tsx
```diff
@@ -0,0 +1,134 @@
...
+		const newOffset = comment.getBoundingClientRect().top;
+		window.scrollBy(0, newOffset - viewportOffset);
+	});
+}
+
+async function handleMenuOpening(event: DelegateEvent): Promise<void> {
+	const dropdown = select('.show-more-popover', event.delegateTarget.parentElement!)!;
+	const user = select('.author', dropdown.closest('.js-comment-container')!)!.textContent!;
+
+	if (user === currentUser) {
```
bfred-it: I think the current user has a specific class name on each comment. Perhaps you can exclude them in the delegation selector

source/features/mute-users.tsx
```diff
@@ -0,0 +1,134 @@
...
+		return;
+	}
+
+	select('.minimized-comment', comment)!.classList.add('d-none');
+	select('.minimized-comment .Details-element', comment)!.setAttribute('open', 'true');
+	select('.unminimized-comment', comment)!.classList.remove('d-none');
+}
+
+async function onMuteUnmuteClick(event: React.MouseEvent<HTMLButtonElement>): Promise<void> {
+	let mutedUsers = await getMutedUsers();
```
bfred-it: This can be moved to the line before it’s used

source/features/mute-users.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+		// If the user _is_ muted, minimize their comment
+		if (mutedUsers.includes(user) && user !== currentUser) {
+			minimizeComment(comment);
+		}
+	}
+}
+
+function init(): void {
+	currentUser = getUsername();
```
bfred-it: No need to cache this, the function is already called once. Just use `getUsername` 👍

source/options.css
```diff
@@ -69,15 +69,26 @@ h2 ~ h2 {
 	text-decoration: line-through;
 }
 
+.js-features :not(:checked) + .info .extended-options {
+	display: none;
+}
+
 .js-features .description {
-	font-style: 0.8em;
```
bfred-it: 👎

source/options.tsx
```diff
@@ -33,18 +33,20 @@ function buildFeatureCheckbox([name, {description, screenshot, disabled}]: [stri
...
-				<span className="feature-name">{name}</span>
-				{' '}
-				<a href={`https://github.com/sindresorhus/refined-github/blob/master/source/features/${name}.tsx`}>
-					source
-				</a>
-				{screenshot ? <>, <a href={screenshot}>screenshot</a></> : ''}
-				<br/>
-				<span className="description">{parsedDescription}</span>
-			</label>
+			<div className="info">
```
bfred-it: Div can’t be a child of P

source/options.css
```diff
@@ -69,15 +69,26 @@ h2 ~ h2 {
...
 
 .js-features a {
 	font-size: 0.9em;
 }
 
+.js-features .info {
+	flex: 1;
+}
+
+.js-features .info p {
```
bfred-it: There’s no P in info

source/options.css
```diff
@@ -69,15 +69,26 @@ h2 ~ h2 {
...
-	font-style: 0.8em;
 	opacity: 0.8;
 }
 
 .js-features a {
 	font-size: 0.9em;
 }
 
+.js-features .info {
+	flex: 1;
```
bfred-it: Does this have any effect? If so, can you avoid the shorthand and use the plain property? e.g. `flex-shrink: 1`

source/features/mute-users.tsx
```diff
@@ -0,0 +1,134 @@
...
+	minimizeMutedUserComments();
+
+	delegate('.timeline-comment-action', 'click', handleMenuOpening);
+}
+
+features.add({
+	id: 'mute-users',
+	description: 'Mute users to minimize comments from them',
```
bfred-it: This might need improvement, but I can’t think of anything more obvious than just “mute users”

source/features/hide-useless-comments.tsx
bfred-it: ?

`minimizeComment(comment)` would become `toggleComment(comment, true)`

source/features/mute-users.tsx
jerone: Agree, because "mute" means to me that I'll not get any notifications from them anymore, which is not the case.

Maybe "**ignore**". It doesn't sound as bad as "blocking".

source/features/mute-users.tsx
notlmn: `parentElement` can return `HTMLElement | null`, but second argument of `select` only accepts `ParentNode | undefined`.

source/options.css
notlmn: My editor was bugging me 😅.

And it's supposed to be `font-size`, not `font-style`.

source/options.css
notlmn: Yes, the `textarea` doesn't fill entire width if this is not expanded.

source/options.html
notlmn: image

source/features/hide-useless-comments.tsx
notlmn: Now we have to as we are immediately querying it again in `minimizeMutedUserComments` on the next line.

source/features/hide-useless-comments.tsx
notlmn: It was so annoying with the DOM contents jumping around on muting/unmuting users without this, so left it out.

source/features/mute-users.tsx
notlmn: From https://github.com/sindresorhus/refined-github/pull/2146#discussion_r295721161.

source/features/mute-users.tsx
notlmn: It's not about micro-optimizing all the time, I wanted to keep it readable.

source/options.tsx
notlmn: Done!

source/options.css
notlmn: Now there is! 😛

source/features/mute-users.tsx
notlmn: > This might need improvement, but I can’t think of anything more obvious than just “mute users”

"Mute users" seems obvious but that doesn't tell us what the feature actually does.

source/options.css
bfred-it: Oh wow, it wasn’t being caught by stylelint either. 😳 

Can you restore it as font-size? I tested the value in the developer tools and then wrote it this way in the repo 🤦‍♂️

source/features/mute-users.tsx
bfred-it: So should we turn any code longer than 5 words into its own one-line function? 😅

source/features/mute-users.tsx
bfred-it: Yeah that’s why I was asking for a better one since I couldn’t think of a better description. 

@jerone Perhaps the feature should be `minimize-user` or something like that? Mute is self-explanatory but it indeed it doesn’t work like the general Mute feature found on socials.

source/options.css
notlmn: On Chrome the `font-size` for this is already at `12px`, if we add this rule, then it goes down to `9.6px`, way too small for anyone to read.

The current options UI font size if fine, IMHO.

source/features/mute-users.tsx
bfred-it: classList has its specific `toggle` method

source/features/mute-users.tsx
bfred-it: I understand the need for the function and I’m saying that it should be part of `minimizeMutedUserComments`

source/features/mute-users.tsx
yakov116: What about `minimize-a-users-comments`

source/features/mute-users.tsx
```diff
@@ -0,0 +1,114 @@
...
+}
+
+async function minimizeMutedUserComments(): Promise<void> {
+	const mutedUsers = await getMutedUsers();
+
+	for (const comment of select.all('.js-discussion .js-minimizable-comment-group')) {
+		const user = select('.author', comment)!.textContent!;
+
+		// If the user _is_ muted, minimize their comment
+		if (mutedUsers.includes(user) && user !== getUsername()) {
```
bfred-it: You can use this as a Boolean as the second parameter of toggleComment, without needing a IF. If it looks too long as a second parameter it can be a const. e.g. 
```suggestion
		const muted = mutedUsers.includes(user) && user !== getUsername());
```

source/features/mute-users.tsx
jerone: `minimize-user` sounds like the right description here. 👍

source/features/mute-users.tsx
bfred-it: Alright

source/features/mute-users.tsx
bfred-it: 👍 

“Ignore” is used by GitHub the same way “Mute” is, actually. See the repo “Watch” dropdown

notlmn: Yes, `minimize-user-comments` sounds fine. But this means we'd have to change the label used in user dropdowns to something other than "Mute/Unmute user".

Maybe "Minimize user comments" and "_Maximize_ user comments"?

source/features/mute-users.tsx
notlmn: In `minimizeMutedUserComments`, there is no concept of a `comment` on which the original event was triggered. That exists only inside `onMuteUnmuteClick`.

source/features/mute-users.tsx
notlmn: `toggle` doesn't work on initial page loads, actually we are not toggling attributes here, we set specific classes and attributes based on the value of `minimize`.

Consider that case where `toggleComment` was called for a comment with the same argument values, `toggle` fails in that case.

bfred-it: `classList.toggle(name, bool)` works exactly the same as `classList[bool ? 'add' : 'remove'](name)`, so I don't know where you're seeing this difference.

`toggleAttribute` however doesn't exist so that line can stay as is.

bfred-it: I just found out that [`toggleAttribute`](https://developer.mozilla.org/en-US/docs/Web/API/Element/toggleAttribute) does exist so all the ternary operators can be dropped

source/features/mute-users.tsx
bfred-it: Yeah. "Comment" would be "whatever is centered on screen right now", e.g. `document.elementFromPoint(innerWidth/2, innerHeight/2)`

There's a library (apparently written for GitHub itself) that gets [the element the mouse is hovering](https://github.com/josh/scroll-anchoring/blob/master/scroll-anchoring.js#L61), but I think that doesn't work well with fixed elements (sidebar, header)

Browsers have a native concept of "scroll anchoring" recently added by both Chrome and Firefox, but I don't think it works in all cases, so probably we can't rely on it.

bfred-it: If you wrap it into its own `scroll-anchoring.ts` function it can be used by other features as well: https://github.com/sindresorhus/refined-github/pull/2087/files#diff-7259b235bc8829b04689968edcc74f94R58

Maybe `anchorScroll(fn: function, anchor?: Node)`

source/features/mute-users.tsx
sindresorhus: > minimize-user-comments

👍 

> Maybe "Minimize user comments" and "Maximize user comments"?

- `Minimize user's comment`
- `Maximize user's comment`

source/features/minimize-user-comments.tsx
```diff
@@ -0,0 +1,105 @@
...
+}
+
+async function setMinimizedUsers(minimizedUsers: string[]): Promise<void> {
+	return optionsStorage.set({minimizedUsers: minimizedUsers.join(' ')});
+}
+
+function toggleComment(comment: HTMLElement, minimize: boolean): void {
+	if (comment.id.startsWith('issue-')) {
+		return;
+	}
```
bfred-it: Hopefully the last change. It may be a good idea to _not_ hide comments by the OP, but currently:

- nothing happens when you _Minimize_ the user
- the button changes to ~~_Maximize_~~ _Restore_ but it still does nothing and makes no sense, because the comment isn't minimized.

So either we hide all comments anyway, except the first one (this can be done in `minimizeMutedUserComments`'s selector) or still apply some visual changes to the author's comments, e.g. `opacity: 0.5`

notlmn:
Blocked by https://github.com/bfred-it/webext-options-sync/commit/fc77d661742e95ecefce66d0b738c1e3a7d7aae3#r34067329.

bfred-it:
(tested https://github.com/sindresorhus/refined-github/pull/2185)

bfred-it:
Is this still blocked?

notlmn:
> Is this still blocked?

Need a new release of https://github.com/bfred-it/webext-options-sync.

bfred-it:
oops, done. 0.21.2

source/features/minimize-user-comments.tsx
```diff
@@ -0,0 +1,102 @@
...
+import {appendBefore} from '../libs/dom-utils';
+import optionsStorage from '../options-storage';
+import {getUsername} from '../libs/utils';
+import onNewComments from '../libs/on-new-comments';
+import anchorScroll from '../libs/anchor-scroll';
+
+const getLabel = (restore: boolean): string => `${restore ? 'Restore' : 'Minimize'} user’s comments`;
+
+async function getMinimizedUsers(): Promise<string[]> {
+	return (await optionsStorage.getAll()).minimizedUsers.split(/\s+/).filter(Boolean);
```
bfred-it:
```suggestion
	return (await optionsStorage.getAll()).minimizedUsers.trim().split(/\s+/);
```

bfred-it:
Weird, I minimized @notlmn first response here:

- it was minimized
- following page loads also kept it minimized

but... your second comment here was never minimized, even if the menu correctly says "restore user"

Screenshot 2019-07-15 at 02 05 58

It might have something to do with him being the OP, but it's still a bug

source/features/minimize-user-comments.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+function init(): void {
+	minimizeMutedUserComments();
+	onNewComments(minimizeMutedUserComments);
+	delegate('.timeline-comment-action', 'click', handleMenuOpening);
+}
+
+features.add({
+	id: __featureName__,
+	description: 'Minimize comments for users you don’t want to see',
```
fregante:
```suggestion
	description: 'Adds ability to minimize comments of certain users.',
```

readme.md
```diff
@@ -179,6 +179,7 @@ Thanks for contributing! 🦋🙌
 - [](# "scrollable-code-and-blockquote") [Limits the height of tall code blocks and quotes.](https://github.com/sindresorhus/refined-github/issues/1123)
 - [](# "hide-comments-faster") [Simplifies the UI to hide comments.](https://user-images.githubusercontent.com/1402241/43039221-1ddc91f6-8d29-11e8-9ed4-93459191a510.gif)
 - [](# "open-issue-to-latest-comment") [Links the comments icon to the latest comment.](https://user-images.githubusercontent.com/14323370/57962709-7019de00-78e8-11e9-8398-7e617ba7a96f.png)
+- [](# "minimize-user-comments") [Minimize comments for users you don’t want to see.](https://user-images.githubusercontent.com/37769974/61595681-d6d4b400-ac17-11e9-98b9-03f27b004a94.gif)
```
fregante:
```suggestion
- [](# "minimize-user-comments") [Adds ability to minimize comments of certain users.](https://user-images.githubusercontent.com/37769974/61595681-d6d4b400-ac17-11e9-98b9-03f27b004a94.gif)
```

fregante:
# 🙌

fregante:
2 months in the making

fregante:
Oh no

fregante:
Also the `.dropdown-menu` should be set to `width: 205px !important` because the text is being clipped...

sharkykh:
How's this?
https://github.com/sindresorhus/refined-github/compare/master...sharkykh:simple-fix-mute-users

fregante:
> Or maybe it should be more like `comments-time-machine-links`?

No, the fewer DOM operations the better. `comments-time-machine-links` used to be 100% outside the dropdown and now it's 50/50



> How's this?
> [master...sharkykh:simple-fix-mute-users](https://github.com/sindresorhus/refined-github/compare/master...sharkykh:simple-fix-mute-users)

I think we just need a `:not([aria-label="Add your reaction"])` selector in the delegate

sharkykh:
> I think we just need a `:not([aria-label="Add your reaction"])` selector in the delegate

The edit buttons also get selected by that. So, `summary.timeline-comment-action:not([aria-label="Add your reaction"])`?

PR #2156, opened 2019-06-13
`Shift+O` shouldn't work when inside an editable element

### LINKED ISSUE #2152: `Shift+O` shouldn't work when writing inside an editable element
GitHub ~recently added a list of similar issues when opening a new issue:

---

Screen Shot 2019-06-13 at 01 58 35

---

The problem is that whenever I type an uppercased O, i.e. `Shift+O` is pressed, the first of those listed issues is opened in a new tab.

`Shift+O` should be ignored when an editable element has the focus.

jerone:
Fixes #2152

### Test
1. Go to https://github.com/sindresorhus/refined-github/issues
2. Give focus to the first issue
3. Put cursor in "Search all issues" searchbox
4. Press `Shift+O`
5. It should not open any issues

### DIFF
--- a/source/features/esc-to-deselect-line.tsx
+++ b/source/features/esc-to-deselect-line.tsx
@@ -1,4 +1,5 @@
 import features from '../libs/features';
+import {isEditable} from '../libs/dom-utils';
 
 function isLineSelected(): boolean {
 	// Example hashes:
@@ -12,7 +13,7 @@ function listener(event: KeyboardEvent): void {
 	if (
 		event.key === 'Escape' && // Catch `Esc` key
 		isLineSelected() &&
-		!(event.target instanceof HTMLTextAreaElement || event.target instanceof HTMLInputElement) // If a field isn’t focused
+		(event.target && !isEditable(event.target)) // If a field isn’t focused
 	) {
 		location.hash = '#no-line'; // Update UI, without `scroll-to-top` behavior
 		history.replaceState({}, document.title, location.pathname); // Drop remaining # from url
--- a/source/features/esc-to-deselect-line.tsx
+++ b/source/features/esc-to-deselect-line.tsx
@@ -13,7 +13,7 @@ function listener(event: KeyboardEvent): void {
 	if (
 		event.key === 'Escape' && // Catch `Esc` key
 		isLineSelected() &&
-		(event.target && !isEditable(event.target)) // If a field isn’t focused
+		!isEditable(event.target) // If a field isn’t focused
 	) {
 		location.hash = '#no-line'; // Update UI, without `scroll-to-top` behavior
 		history.replaceState({}, document.title, location.pathname); // Drop remaining # from url
--- a/source/libs/dom-utils.ts
+++ b/source/libs/dom-utils.ts
@@ -46,7 +46,7 @@ export const wrapAll = (targets: Element[], wrapper: Element): void => {
 	wrapper.append(...targets);
 };
 
-export const isEditable = (target: Element | EventTarget): boolean => {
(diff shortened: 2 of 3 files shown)


source/features/esc-to-deselect-line.tsx
```diff
@@ -12,7 +13,7 @@ function listener(event: KeyboardEvent): void {
 	if (
 		event.key === 'Escape' && // Catch `Esc` key
 		isLineSelected() &&
-		!(event.target instanceof HTMLTextAreaElement || event.target instanceof HTMLInputElement) // If a field isn’t focused
+		(event.target && !isEditable(event.target)) // If a field isn’t focused
```
jerone: `event.target` could be null

source/libs/dom-utils.ts
```diff
@@ -45,3 +45,9 @@ export const wrapAll = (targets: Element[], wrapper: Element): void => {
 	targets[0].before(wrapper);
 	wrapper.append(...targets);
 };
+
+export const isEditable = (target: Element | EventTarget): boolean => {
+	return target instanceof HTMLTextAreaElement ||
```
jerone: Can change these lines also to 
```typescript
(target instanceof Element && (target as any as Element).matches('input,textarea,[contenteditable=true]'));
```
But I found this cleaner.

source/libs/dom-utils.ts
```diff
@@ -45,3 +45,9 @@ export const wrapAll = (targets: Element[], wrapper: Element): void => {
...
+		target instanceof HTMLInputElement ||
+		(target instanceof Element && (target as any as Element).matches('[contenteditable=true]'));
```
jerone: Needs to be `[contenteditable=true]`, because `false` is also an acceptable value.

source/features/esc-to-deselect-line.tsx
bfred-it: Target _can_ be null. You can change the function to accept a null parameter and drop the checks outside the function

source/libs/dom-utils.ts
```diff
@@ -45,3 +45,9 @@ export const wrapAll = (targets: Element[], wrapper: Element): void => {
 	targets[0].before(wrapper);
 	wrapper.append(...targets);
 };
+
+export const isEditable = (target: Element | EventTarget | null): boolean => {
```
bfred-it:
```suggestion
export const isEditable = (node?: node): boolean => {
```

jerone: @bfred-it It doesn't work: https://travis-ci.org/sindresorhus/refined-github/builds/547586732

```
ERROR in /home/travis/build/sindresorhus/refined-github/source/features/esc-to-deselect-line.tsx
[tsl] ERROR in /home/travis/build/sindresorhus/refined-github/source/features/esc-to-deselect-line.tsx(16,15)
      TS2345: Argument of type 'EventTarget | null' is not assignable to parameter of type 'Node | undefined'.
  Type 'null' is not assignable to type 'Node | undefined'.
ERROR in /home/travis/build/sindresorhus/refined-github/source/features/selection-in-new-tab.tsx
[tsl] ERROR in /home/travis/build/sindresorhus/refined-github/source/features/selection-in-new-tab.tsx(8,52)
      TS2345: Argument of type 'EventTarget | null' is not assignable to parameter of type 'Node | undefined'.
  Type 'null' is not assignable to type 'Node | undefined'.
```

bfred-it: `node?: any`, since it’s a type guard function

Maybe the question mark isn’t necessary

jerone: Without the `?` it works.

Issue #2119, opened 2019-06-04
Show colors on hover


sumanthratna:
For hex colors and RBG codes, it would be nice to show a box above the cursor when the user hovers over the code.

sindresorhus:
Would be useful, but not sure it's worth the implementation complexity. You should definitely request the feature at support@github.com.

Tip: You can see colors `#FF6347` in comments.

jerone:
There is a great well-maintained userscript from Mottie which does exactly this, "[GitHub code colors](https://github.com/Mottie/GitHub-userscripts/wiki/GitHub-code-colors)".

jerone:
Maybe we can use above mention userscript as a reference, because I think this would be great addition to this extension.

PR #2158, opened 2019-06-15
Add view color feature

sumanthratna:
Implements #2119. 

NOTE: the types are all wrong, I'd appreciate it if someone else fixes the types (but the Chrome extension still works).

### DIFF
--- a/source/features/view-color.tsx
+++ b/source/features/view-color.tsx
@@ -1,0 +1,166 @@
+import features from '../libs/features';
+
+function init(): void {
+	addColors();
+}
+const namedColors: string = [
+	"aliceblue", "antiquewhite", "aqua", "aquamarine", "azure", "beige",
+	"bisque", "black", "blanchedalmond", "blue", "blueviolet", "brown",
+	"burlywood", "cadetblue", "chartreuse", "chocolate", "coral",
+	"cornflowerblue", "cornsilk", "crimson", "cyan", "darkblue", "darkcyan",
+	"darkgoldenrod", "darkgray", "darkgrey", "darkgreen", "darkkhaki",
+	"darkmagenta", "darkolivegreen", "darkorange", "darkorchid", "darkred",
+	"darksalmon", "darkseagreen", "darkslateblue", "darkslategray",
+	"darkslategrey", "darkturquoise", "darkviolet", "deeppink", "deepskyblue",
+	"dimgray", "dimgrey", "dodgerblue", "firebrick", "floralwhite",
+	"forestgreen", "fuchsia", "gainsboro", "ghostwhite", "gold", "goldenrod",
+	"gray", "grey", "green", "greenyellow", "honeydew", "hotpink",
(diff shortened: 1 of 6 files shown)


source/features/view-color.tsx
```diff
@@ -0,0 +1,166 @@
+import features from '../libs/features';
+
+function init(): void {
+	addColors();
+}
+const namedColors: string = [
```
jerone: Can probably be replaced by some npm package, like [color-name-list](https://www.npmjs.com/package/color-name-list)...

jerone:
If you copy the whole source of the [mentioned UserScript](https://github.com/sindresorhus/refined-github/issues/2119#issuecomment-498982592), you probably have to ask permission from the author.

sumanthratna:
> If you copy the whole source of the [mentioned UserScript](https://github.com/sindresorhus/refined-github/issues/2119#issuecomment-498982592), you probably have to ask permission from the author.

He's okay with it!

bfred-it:
I did ask to integrate his user scripts, but instead of copy-pasting the code and then having to support it separately, I’d rather `require` the script, add the globals needed (`GM.addStyle` and the events we need) and build a regular feature around it. 

This would enable us to just follow any updates Mottie has in the future, rather than forking his code, and add more features userscripts the same way in the future.

To do so, we need to keep the feature as two files:

- `color-preview.user.js`: 99% copy-pasted file, excluded from XO, only changes to export a function and disable the listeners
- `color-preview.tsx`: the actual feature wrapper that has the listeners and feature initialization. It will import the user.js file and run it as necessary.

source/features/view-color.tsx
```diff
@@ -0,0 +1,137 @@
...
+			if (!status.done) {
+				requestAnimationFrame(loop);
+			}
+		};
+		loop();
+	}
+}
+features.add({
+	id: 'view-color',
+	description: 'View various color codes',
```
bfred-it: “All color previews next to colors everywhere”

jerone:
> I did ask to integrate his user scripts, but instead of forking the code and then having to support it separately, I’d rather `require` the script, add the globals needed (`GM.addStyle` and the events we need) and build a regular feature around it.

Interesting idea. That will open up using other UserScripts too.



> Alternatively we can probably `npm install` and require the file directly from GitHub and then use the existing webpack replace loader do make these changes on the fly. So no copy-pasting required.

That will work:
```cli
npm install --save git+https://github.com/Mottie/GitHub-userscripts.git#v1.0.241
```

bfred-it:
npm natively supports installing from GitHub using `npm install user/repo`

sumanthratna:
I think this PR should just be reimplemented from scratch, using the suggestions mentioned above.

fregante:
> I did ask to integrate his user scripts,

Change of direction, for posterity: **if a userscript/extension already exist, just use that.**

Refined GitHub does not have to include everything, we already have a lot of stuff to take care of.

PR #2166, opened 2019-06-22
Dynamically add feature names to files

bfred-it:
This will avoid bugs like https://github.com/sindresorhus/refined-github/pull/2112/files and https://github.com/sindresorhus/refined-github/issues/2057

### DIFF
--- a/source/globals.d.ts
+++ b/source/globals.d.ts
@@ -1,6 +1,7 @@
 type AnyObject = Record<string, any>;
 
 declare const __featuresList__: string[];
+declare const __featureName__: 'webpack will replace this string';
 
 interface Window {
 	collectFeatures: Map<string, FeatureDetails>;
--- a/webpack.config.ts
+++ b/webpack.config.ts
@@ -63,6 +63,10 @@ module.exports = (_env: string, argv: Record<string, boolean | number | string>)
 				}
 
 				return JSON.stringify(features);
+			}),
+			// @ts-ignore
+			__featureName__: webpack.DefinePlugin.runtimeValue(({module}) => {
+				return JSON.stringify(path.basename(module.resource, '.tsx'));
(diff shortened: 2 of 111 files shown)


source/globals.d.ts
```diff
@@ -1,6 +1,7 @@
 type AnyObject = Record<string, any>;
 
 declare const __featuresList__: string[];
+declare const __featureName__: 'webpack will replace this string';
```
bfred-it: This is to avoid contributors trying to replace the `__featureName__` with the actual feature name.

webpack.config.ts
```diff
@@ -22,25 +22,6 @@ module.exports = (_env: string, argv: Record<string, boolean | number | string>)
...
-						for (const filename of readdirSync(directoryPath)) {
-							if (filename.endsWith('.tsx')) {
-								features.push(filename.replace('.tsx', ''));
-							}
-						}
-
-						return JSON.stringify(features);
-					}
-				}
-			},
```
bfred-it: I'm also moving this to `webpack.DefinePlugin` but I'm leaving `string-replace-loader` in the dependencies to enable #2119

webpack.config.ts
```diff
@@ -69,6 +50,25 @@ module.exports = (_env: string, argv: Record<string, boolean | number | string>)
...
+				const directoryPath = path.join(__dirname, 'source/features');
+				for (const filename of readdirSync(directoryPath)) {
+					if (filename.endsWith('.tsx')) {
+						features.push(filename.replace('.tsx', ''));
+					}
+				}
+
+				return JSON.stringify(features);
+			}),
+			// @ts-ignore
```
bfred-it: Waiting for https://github.com/webpack/webpack/issues/9316

PR #2167, opened 2019-06-22
Meta: Parse feature details at compile time

bfred-it:
Follows https://github.com/sindresorhus/refined-github/pull/2099

This avoids any issues/annoyances caused by loading `content.js` in the `options.html` page.


webpack.config.ts
```diff
@@ -1,11 +1,35 @@
+/// <reference types="./source/globals" />
```
bfred-it: This also solves the `// @ts-ignore` issue below for `size-plugin`

webpack.config.ts
```diff
@@ -1,11 +1,27 @@
...
+
+function parseFeatureDetails(name: string): FeatureInfo {
+	const content = readFileSync(`source/features/${name}.tsx`, {encoding: 'utf-8'});
+	const rawRegex = ['description', 'screenshot', 'disabled']
+		.map(field => `\n\t${field}: '(?<${field}>[^\\n]+)'`) // Named group regex
+		.join('|');
+
+	const feature = new RegExp(rawRegex).exec(content)!.groups!;
+	feature.name = name;
+	return feature as unknown as FeatureInfo;
```
bfred-it: The description field might have some `\'` that will be shown like that in the options page.

Technically we prefer `’` so this isn't a problem, but we don't enforce it automatically so some might slip through.

Help wanted to create a local `eslint` plugin to enforce and autofix it (replace `\'` with `’`)

sindresorhus: I don't have time to write a rule, but this is what you need:

```js
'use strict';

const create = context => {
	return {
		Literal: node => {
          	const {value} = node;
     
			if (typeof value !== 'string') {
				return;
			}
          
        	const fixedValue = value.replace(/'/, '‘');

			if (fixedValue === value) {
				return;
			}

			context.report({
				node,
				message: 'Use other quote kind',
				fix: fixer => fixer.replaceTextRange([node.range[0] + 1, node.range[1] - 1], fixedValue)
			});
		}
	};
};

module.exports = {
	create,
	meta: {
		type: 'problem',
		fixable: 'code'
	}
};
```

https://astexplorer.net/#/gist/3a67df351a57608ae38a6b55102376d6/c8425e326ddce75bf7242295c8818134ef2d2e16

sindresorhus: Maybe you should propose it on https://github.com/sindresorhus/eslint-plugin-unicorn

bfred-it:
I'm gonna merge this because I'm having the exact problems that this change is meant to solve.

... but please review this code and maybe #2099 anyway just in case I'm doing something stupid.

PR #2181, opened 2019-06-27
Make Enterprise injection more reliable /2

bfred-it:
Follows #2111
Fixes #2174 #2173 #2171 https://github.com/sindresorhus/refined-github/pull/2184

cc'ing GHE users @ahatzz11 @gibfahn @rastersize @SivanMehta @NickCraver

Issue #2174, opened 2019-06-24
`content.css` not injected on GHE


ahatzz11:
This was initially discussed in https://github.com/sindresorhus/refined-github/issues/2170

Here is how the feature should work:
![](https://user-images.githubusercontent.com/1402241/60038476-08c12c00-96e7-11e9-8af2-99fe6d5e1e6f.gif)

Here is how it actually looks on GHE for me (in incognito with no other extensions)

![image](https://user-images.githubusercontent.com/6256032/60051439-a0e3f480-9698-11e9-9b0e-1c8229f6ba92.png)

bfred-it:
This is a consequence of https://github.com/sindresorhus/refined-github/pull/2111

I’ll have to fix it

bfred-it:
Great! The API I used to implement it in Chrome is broken for CSS: https://bugs.chromium.org/p/chromium/issues/detail?id=708115

Found on https://bugzilla.mozilla.org/show_bug.cgi?id=1323433 (Mozilla decided not to implement Chrome's API because it's broken and unused and they have a better one)

In my test, the CSS is injected only on the first load until you visit a different domain.

I'll see if I can work around this issue or if I have to go back to https://github.com/bfred-it/webext-dynamic-content-scripts/ in Chrome 🙄

tknickman:
Confirmed this is fixed on GHE `v2.16.10`. Thanks so much!

PR #2184, opened 2019-06-27
Add display: none; to remove-diff-signs.css

StephenBrown2:
I have found that running v19.6.26 in Chrome 75.0.3770.100 on our GitHub Enterprise instance running 2.16, `content: none;` does not work to remove the diff marker, but `display: none;` does. Also, it appears to throw off the indentation wildly and inconsistently, so for now I've disabled that feature.

An addition line looks like this on GHE 2.16:
```html
<td class="blob-code blob-code-addition blob-code-marker-cell" data-code-marker="+">
  </td>
```
Whereas it's a span on Github.com:
```html
<span class="blob-code-inner blob-code-marker" data-code-marker="+"><br></span>
```

### DIFF
--- a/source/features/remove-diff-signs.css
+++ b/source/features/remove-diff-signs.css
@@ -1,4 +1,5 @@
 .rgh-remove-diff-signs .blob-code-inner::before, /* Inline review blocks in PRs */
 .rgh-remove-diff-signs .blob-code-marker-cell::before { /* Commit page */
 	content: none;
+	display: none;
 }


notlmn:
The element being `span` or `td` doesn't matter, and this PR specifically fixes just diff markers issue.

What you are experiencing on GitHub Enterprise is #2174. It is currently being fixed in #2181.

busches:
Please test #2181 and provide feedback there.

bfred-it:
Sorry for the trouble and thanks for the PR, but indeed there has a been a bug on GHE. It'll be fixed in the next release

StephenBrown2:
Thanks! I'll wait for the next release, as I noticed the other issues in #2174 as well.

PR #2191, opened 2019-06-28
Update `webext-storage-cache`

bfred-it:
Follows #2109

Finalized the module. I'll add tests at some point too: https://github.com/bfred-it/webext-storage-cache

cc @dotconnor @notlmn

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13769,9 +13769,9 @@
 			}
 		},
 		"webext-storage-cache": {
-			"version": "0.0.0",
-			"resolved": "https://registry.npmjs.org/webext-storage-cache/-/webext-storage-cache-0.0.0.tgz",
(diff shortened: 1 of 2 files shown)


notlmn:
Weird, after this change I'm getting linting errors from VSCode but **not** when running `npm run lint:js`. For comparison, this PR should've errored out.

image

bfred-it:
This may be because you can configure a different TS version in the bottom status bar of VSCode and the latest version is stricter. I had that same error when using an index signature but it went away when I switched to Record (almost the same, but tighter)

dotconnor:
Heres the typescript issue on this for reference: https://github.com/microsoft/TypeScript/issues/15300

Simple fix it just turn `CacheEntry` into a `type` alias:

```
type CacheEntry = {
	repoProjectCount: number;
	orgProjectCount: number;
	milestoneCount: number;
}
```

PR #2185, opened 2019-06-27
Enable `update-pr-from-base-branch` on forks

bfred-it:
Closes #2097

Also on draft PRs, thanks to the "out of date" message:

- API-less out-date detection
- better placement

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

This pull request has been submitted to:
- [#2097: Update feature `update-pr-from-base-branch` to use new API](https://issuehunt.io/repos/51769689/issues/2097)
---

IssueHunt has been backed by the following sponsors. [Become a sponsor](https://issuehunt.io/membership/members)
</details>

PR #2193, opened 2019-06-29
Replace username with realname if they’re the same: `show-names`

### LINKED ISSUE #2165: Avoid adding real names when the user is using a real name
This isn't useful:

I'd rather see this:

--- 

Feature: `show-names`

```
If name matches username
	replace username with name (to keep capitalization)
```

- Either we drop the space altogether (so in this case there won't be any changes for these users, but there would be for `sindresorhus`, which would become `SindreSorhus`), or 
- we replace it with `<span style="margin-left:0.2em"/>` so it won't break copy-pasting.

Example URL:
https://github.com/microsoft/TypeScript/issues/30442#ref-pullrequest-433925264

dotconnor:
Closes #2165

## After

# Test

https://github.com/microsoft/TypeScript/issues/30442#ref-pullrequest-433925264
https://github.com/bfred-it/webext-options-sync/issues/13#issuecomment-497941181

Requires #2146 to use `select<HTMLAnchorElement>('.author')!.pathname.slice(1)`


source/features/show-names.tsx
```diff
@@ -38,22 +45,27 @@ async function init(): Promise<false | void> {
...
-				' (',
-				<bdo className="css-truncate">
-					<span className="css-truncate-target" style={{maxWidth: '200px'}}>
-						{names[userKey].name}
-					</span>
-				</bdo>,
-				') '
-			);
+			if (compareNames(username, names[userKey].name)) {
+				usernameEl.textContent = names[userKey].name;
```
bfred-it: This breaks "copy-pastability" of users. The alternative is to add the spaces and punctuation via non-selectable elements, as suggested in the original issue.

@sindresorhus thoughts? Is copy-pastability important?

bfred-it:
https://github.com/sindresorhus/refined-github/issues/2165#issuecomment-504649392

>  to use `(select('.author')! as HTMLAnchorElement).pathname.slice(1)`

It might be worth looking for other places that would break because of this.

TS note: the code would be `select<HTMLAnchorElement>('.author')!.pathname.slice(1)`

source/features/show-names.tsx
```diff
@@ -5,6 +5,13 @@ import * as api from '../libs/api';
 import features from '../libs/features';
 import {getUsername} from '../libs/utils';
 
+// Returns true if username is basiclly the same as realname
+function compareNames(username: string, realname: string): boolean {
+	const reg = /[ .]/g;
+
+	return username.replace(reg, '').toLowerCase() === realname.replace(reg, '').toLowerCase();
```
jerone: Checking the username for spaces is not necessary, as per GitHub self:
> Username may only contain alphanumeric characters or single hyphens, and cannot begin or end with a hyphen.

jerone:
I would check for the following:
* username: (long and short) hyphen and dot.
* real name: (long and short) hyphen, dot and space.

source/features/show-names.tsx
bfred-it:
```suggestion
	return username.replace(/[.-]/g, '').toLowerCase() === realname.replace(/[^W]/gi, '').toLowerCase();
```

But the comparison needs to be even better: `@nicolo` won't match `Nicolò` without unicode normalization.

dotconnor: `String.prototype.noramlize` is available for unicode normalization:

```
realname.normalize(‘NFD’).replace(/[\u0300-\u036f]/g, '').replace(/[^W]/giu, '').toLowerCase();
```

source/features/show-names.tsx
bfred-it: Maybe it's ok, the real username appears in the hovercard

source/features/show-names.tsx
bfred-it: When you implement this, it would be good to have tests to match usernames to their real names, to ensure it works correctly without having to find them in the real world.

source/features/show-names.tsx
yakov116: Not so easy to copy though
Video_2019-06-30_095646

bfred-it: Uhm? The regular username is the same, it's a link so you have to hold `alt` to select it easily

bfred-it:
(tested #2185)

bfred-it:
Can you pick this up again before #2229?

fregante:
Oops, didn't notice you fixed this a while ago.

Tested and it works great both in discussions and newsfeed.

PR #2204, opened 2019-07-03
Fix text formatting error with non-Roman inputs

### LINKED ISSUE #2203: Pressing ' conflicts with Korean keyboard composition
in korea some use not typical keyboard layouts like me.
common korean word made by automatas that state seleted mode
in this case, some key use `'` for typo, not a quote.

see this image

auto-quotes

we want to disable this automatically quite feature.
but can't see disable option.

Example URL:

bfred-it:
Fixes #2203 

@soomtong can you test this? I can't replicate the original error


soomtong:
yes I tested this branch. nice works.
it solve the problem and not remove auto quotes with normal selected text block.

auto-quotes-solved

bfred-it:
Excellent! Thank you for testing :)

PR #2211, opened 2019-07-07
Restore lock icons on private orgs

### LINKED ISSUE #2210: Private orgs no longer showing lock icons
Hiya 👋 

I noticed this morning that private orgs no longer showed the lock icon on them. There was also a relevant error message saying that `publicOrgs.map is not a function`. After some digging, it seems as though the response from GitHub's API changed a bit(?), and some of the keys in the response aren't iterable.

<p align="center">
	Screen_Shot_2019-07-07_at_3_21_53_PM
</p>

I have a PR incoming with a suggested fix 😄

doms:
## What

Fixes #2210 

## Notes

After some digging, it seems as though the response from GitHub's API changed a bit(?), and some of the keys in the response aren't iterable. 🤔 

This just uses an explicit `Object.values` around the `publicOrgs` response object. 😄

### Before

Screen Shot 2019-07-07 at 3 39 56 PM

### After

Screen Shot 2019-07-07 at 3 40 08 PM

### DIFF
--- a/source/features/mark-private-orgs.tsx
+++ b/source/features/mark-private-orgs.tsx
@@ -12,7 +12,7 @@ async function init(): Promise<false | void> {
 	}
 
 	let publicOrgs = await api.v3(`users/${getUsername()}/orgs`);
-	publicOrgs = publicOrgs.map((orgData: AnyObject) => `/${orgData.login}`);
+	publicOrgs = Object.values(publicOrgs).map((orgData: AnyObject) => `/${orgData.login}`);
 
 	for (const org of orgs) {
 		if (!publicOrgs.includes(org.pathname)) {
--- a/source/features/mark-private-orgs.tsx
+++ b/source/features/mark-private-orgs.tsx
@@ -12,7 +12,7 @@ async function init(): Promise<false | void> {
 	}
 
 	let publicOrgs = await api.v3(`users/${getUsername()}/orgs`);
-	publicOrgs = Object.values(publicOrgs).map((orgData: AnyObject) => `/${orgData.login}`);
+	publicOrgs = publicOrgs.map((orgData: AnyObject) => `/${orgData.login}`);
 
 	for (const org of orgs) {
 		if (!publicOrgs.includes(org.pathname)) {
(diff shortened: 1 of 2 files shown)


bfred-it:
They should be iterable; if this happens there must be a problem in the API or in the cache module

doms:
Perhaps “not iterable” was a poor choice of words, but I wasn’t sure other words to use to describe it when I was debugging.

When I was `console.log`’ing the response of fetching the public orgs, the response would look like as shown:

Screen Shot 2019-07-08 at 10 21 10 PM

And so I figured I would iterate through the keys, and it seemed that sometimes I was getting `0` as a key, so I figured that’s why it was error’ing out with the `publicOrgs.map is not a function` 🤔

```js
let publicOrgs = await api.v3(`users/${getUsername()}/orgs`);
	
// console.log's to see the contents of `publicOrgs` response
console.log(publicOrgs)	
for (let key in publicOrgs) {
	console.log('key: ', key)
}
	
publicOrgs = Object.values(publicOrgs).map((orgData: AnyObject) => `/${orgData.login}`);
```
Screen Shot 2019-07-08 at 10 22 53 PM

As you see in the screenshot, we are getting the keys of:

* `0` (the actual response as shown in the first image above)
* `status`
* `ok`

It seems that `status: 200` and `ok: true` are key-value pairs, and the `0` key has the user orgs as the values 🤔 

So I figured I would just use `Object.values(publicOrgs)` to make sure I was getting a type that did have the `map` functionality (Array), and go from there. It _seems_ to have solved the issue, but the weird part is that the original code works too in a different environment (https://repl.it/repls/DefiantCraftyInterface)

I'm getting similar behavior in both Brave (Chromium) and Firefox.

So yea, not entirely sure what’s happening, but I figured I would attempt to fix it maybe get some feedback in the code review on _why_ `mark-private-orgs` might be broken on my end. :smile:

jerone:
I can confirm this issue:
image

It is because the API itself returns an `array`, which is [because of recent change](https://github.com/sindresorhus/refined-github/pull/2185/files#diff-3b4e3f99bb4c228c70878e87f3e7808cL103) now [spread to an `object`](https://github.com/sindresorhus/refined-github/blob/master/source/libs/api.ts#L107) in #2185. Resulting in using the `map` function on an `object` instead of the previous `array`.

bfred-it:
@jerone that’s funny 😅 good catch

The solution is to use Object.assign(array, {.......})

Issue #2194, opened 2019-06-30
Add link to http://forked.yannick.io/ next to the fork button to enable finding most recent fork


koppor:
When searching GitHub for interesting projects, some of them are abandoned. Sometimes, there are forks existing. GitHub presents the forks just as plain list. There is no indication which one is the most active, etc. Now, there is the service http://forked.yannick.io/ which list all forks an metrics on the forks. Example:

![image](https://user-images.githubusercontent.com/1366654/60397454-d03daa00-9b4d-11e9-92ce-363daaadd782.png)

It would be very helpful if one could jump from a repository directly to the respective page in there. For instance from this extension to http://forked.yannick.io/sindresorhus/refined-github.

sindresorhus:
I think it could be useful, but can you illustrate how you image it would look? I don't think putting it next to the Fork button would look good.

bfred-it:
Prior art: https://github.com/musically-ut/lovely-forks

I never actually found this useful.

GitHub offers a "network" page showing that information as well: https://github.com/sindresorhus/refined-github/network

1138-4EB:
It seems that `yannick.io` is no longer available. I'd say that the corresponding repo is https://github.com/ys/forked.

I think that a feature such as this should use the multiple metrics available in the 'Pulse' view of each repo/fork. Precisely, I submitted the following suggestion through https://www.research.net/r/J7BJ7XF:

> Allow to 'zoom out' (to have a better view of the 'big picture').
>
> It is hard to tell apart forks which just add custom details (such as changing titles/names/assets for a company/project), from actual forks that contribute useful features for other users (probably because the 'original' repo is stalled). I think that adding statistics about the forks (stars, contributors, opened/closed PRs, etc.) would really help here. There is enough info already available in the 'Pulse' view of each repo/fork that could be used for this task.

However, this might be difficult to implement outside of GitHub, due to API rate limits: https://github.com/ys/forked/issues/9#issuecomment-304769045.

koppor:
@1138-4EB  It used to work when I opened the issue. Thank you for the pointers.

@bfred-it Quite useful tool.

koppor:
> I think it could be useful, but can you illustrate how you image it would look? I don't think putting it next to the Fork button would look good.

I would indeed put it right of the fork button. Keeping the number of forks, but replacing the link from the normal GitHub fork network to a more useful list.

Since http://forked.yannick.io/ is down, I will close this request. If it comes back, I will reopen it.

[...]

PR #2214, opened 2019-07-08
Fix UI jumping when repo forked

karlhorky:
The UI jumps when a repo has been forked:

Kapture 2019-07-08 at 13 58 01

Setting `position: absolute` on the `.rgh-forked` element removes it from the document flow and prevents pushing the other content further down the page.

The design seems still acceptable:

**After:**

**Before:**

# Test

Visit a URL of a repo that you forked and watch the UI jump.

### DIFF
--- a/source/features/forked-to.css
+++ b/source/features/forked-to.css
@@ -1,0 +1,3 @@
+.rgh-forked {
+	position: absolute;
+}
--- a/source/features/forked-to.tsx
+++ b/source/features/forked-to.tsx
@@ -1,3 +1,4 @@
+import './forked-to.css';
 import React from 'dom-chef';
 import select from 'select-dom';
 import cache from 'webext-storage-cache';


jerone:
This won't work with multiple forks, or when you are already on a fork:
image

bfred-it:
cc @jerone 

If accepted, it should just use the existing `position-absolute` class instead of having a CSS file.

bfred-it:
If you want to fix this, you'll have to use `elementReady` (to add the link as soon as the repo title is loaded). However `onAjaxedPages` always waits for `onDomReady` so a new solution needs to be found.

karlhorky:
> This won't work with multiple forks, or when you are already on a fork

Oh oops, sorry about that! Not sure exactly what the solution is that you're proposing... but happy to help once I understand that.

Wonder if it would be possible to instead rewrite the existing text for cases where this DOM element already exists...

bfred-it:
On long pages, parts of the page may be displayed a few seconds **before**  the "DOM ready" event, so we use [`elementReady`](https://github.com/sindresorhus/element-ready) to look for elements as soon as they are loaded. This helps exactly this issue: the jumping.

However, `elementReady` can't be used together with the loading method used by this feature: `load: features.onAjaxedPages` because it calls `init`:
- on the first "DOM ready" event
- every time the page is loaded via ajax 

We need a new loading method to call `init`:

- instantly (so we can use [`elementReady`](https://github.com/sindresorhus/element-ready))
- every time the page is loaded via ajax

karlhorky:
Ah ok, I understand, working on the loading/ready internals a bit.

Would my other suggestion be an alternative option?

> Wonder if it would be possible to instead append to the existing innerHTML for cases where this DOM element already exists...

I was thinking of something like:

forked to __user-a/repo-name__, __user-b/repo-name__

...for more complex cases:

forked from __user-z/repo-name__, forked to __user-c/repo-name__

A problem that I could imagine would be that the vertical space is not long enough for longer lists. With a lot it looks like the screenshot below. I guess this would be an edge case though.

**GitHub's normal styles with long text**

Screen Shot 2019-07-08 at 18 32 42

**Refined GitHub's element with `position: absolute` and long text**

Screen Shot 2019-07-08 at 18 32 53

bfred-it:
I'd rather have the proper loading mechanism. I think you can just add one more loader exactly as:

```ts
features.add({
	id: __featureName__,
	description: '',
	include: [
		features.isRepo
	],
	load: elementReady('.pagehead h1.public'),
	init
});
```

The only thing you'd have to make sure is that the links aren't added twice since now `init` will be called on `elementReady` **and** on `domReady`

jerone:
@karlhorky If you like and can't find a solution, I can take over.

bfred-it:
I'm not sure I like the position of this feature either, it makes me feel I'm always on the fork:

Screenshot 2019-07-10 at 14 04 15

Screenshot 2019-07-10 at 14 04 48
 

---

cc @yakov116 I moved my comment here because I had posted it in the wrong thread

jerone:
> I'm not sure I like the position of this feature either, it makes me feel I'm always on the fork:

Now that I've used this feature for a while, I agree. Still, I do find it very useful to instantly see if I (or my organization) already forked a repo.

What position would be better? 

Thinking about this more, I do like the [suggestion](https://github.com/sindresorhus/refined-github/pull/2214#issuecomment-509299455) of **combining the "forked-to" with "forked-from" into one sentence**. I doubt that people will have more then two or three "forked-to/from"'s.

karlhorky:
> I'm not sure I like the position of this feature either, it makes me feel I'm always on the fork

At the beginning I also had this issue: over time I got used to knowing I was on the fork if my eye caught a glimpse of something up there (didn't even need to read it).

But I have gotten used to the new behavior of Refined GitHub since too...

bfred-it:
> Thinking about this more, I do like the [suggestion](https://github.com/sindresorhus/refined-github/pull/2214#issuecomment-509299455) of **combining the "forked-to" with "forked-from" into one sentence**. I doubt that people will have more then two or three "forked-to/from"'s. That way we also get ride of the UI jumping, because we stay on one line

This doesn't the most common  case: viewing the origin. The origin _does not_ have a line there so there's no line to extend.

I'd probably hide it in here: (old mockup)



The middle button would be a link to the fork or a dropdown to other existing forks


Or... we just change the feature to **preload** the forks box instead, so clicking there will be faster.

karlhorky:
> This doesn't the most common case: viewing the origin. The origin does not have a line there so there's no line to extend.

Yep, my solution was just for the other cases.

The behavior on the origin could be the current behavior with `position: absolute` to fix the jumping (like in this PR).

karlhorky:
But yeah, if it's better to move it elsewhere, then I can close this PR and we can go ahead with that instead.

jerone:
> This doesn't the most common case: viewing the origin.

Oh yes 😊

> I'd probably hide it in here: (old mockup)

This has no advantage over just clicking the <kbd>Fork</kbd> button; it has the same list.

1138-4EB:
> > I'd probably hide it in here: (old mockup)
> 
> This has no advantage over just clicking the Fork button; it has the same list.

I think that the advantage would be that it would be faster. When a single fork exists, it will jump directly with a single click. When more than a fork exist, the list will be shown right next to the button; unlike the actual list that is shown in the middle of the page. This makes it faster to move the mouse/touchpad.

I think this option is preferred to any other alternative that tries to fit everything in a single (short) line. As commented in:

> I doubt that people will have more then two or three "forked-to/from"'s. That way we also get ride of the UI jumping, because we stay on one line

I have a case where three forks of a repo are shown. The same repo is forked by a user and by two organization to which the user belongs.

I is also possible to mix both solutions:

- In the origin:
  - If a single fork is found, 'forked to' is shown.
  - If more than a single fork are found, just a symbol is shown, and the list is provided as a dropdown of the 'Fork' button.
- In a fork:
  - 'forked from' is shown.
  - If other forks are found, the list is provided as a dropdown of the 'Fork' button.

This approach would allow to show 'forked to' and 'forked from' on the right of the name of the repo, instead of trying to fit it below. I believe that this would avoid this issue.

Moreover, having a dropdown in 'Fork' can be used in the future to show not only the users forks, but also other 'notable forks' (see #2194).

bfred-it:
I was thinking of something easier:

- one fork (or first fork, always): adds extra button (↗️) next to `Fork`, as in "Visit your fork"
- other forks (or more than one): just preload the modal so it's faster to visit it.

1138-4EB:
>     * one fork (or first fork, always): adds extra button (↗️) next to `Fork`, as in "Visit your fork"

I think it is interesting to keep this close to the repo name and to explicitly show the target. If the user belongs to an organization and the fork is in the organization, not in the user's account, 'Visit your fork' might be misleading.

> **sindresorhus/refined-github** forked to [1138-4EB/refined-github]()
> **1138-4EB/refined-github** forked from [sindresorhus/refined-github]()

looks quite natural to me.

>     * other forks (or more than one): just preload the regular `Fork` button modal so it's faster to visit it.

Apart from the comment above, I think it is important to keep 'Make a fork' and 'Visit/view existing forks' in separate buttons/modals/menus.

> I don't know how many forks you keep, but generally most people only keep one fork and that "Visit your fork" button should be good enough for most.

I have quite a few cases with two forks. There is the original repo, one in a organization I belong to, and the last one in my userspace. In these cases, having quick access to two forks is really useful.

jerone:
For me, the main advantage of this feature, is to (1) **quickly identify** that I have **no forks**, or **# amount of forks**. The second advantage is to (2) quickly go to one (or more) of my forks.

As stated [above](https://github.com/sindresorhus/refined-github/pull/2214#issuecomment-509954071), I do agree that the current place (below the repo name) is a little bit confusing, because we are so used to identifying that space as knowing that the current repo is already a fork.

Coming back to:

>> I'd probably hide it in here: (old mockup)
>
> This has no advantage over just clicking the Fork button; it has the same list.

...and your reply:

> I was thinking of something easier:
> 
> * one fork (or first fork, always): adds extra button (↗️) next to `Fork`, as in "Visit your fork"
> 
> * other forks (or more than one): just preload the regular `Fork` button modal so it's faster to visit it.

I see your suggestion. Let me suggest an alternative:

* **No forks**: we show no pull-down button.
* **One fork**: let's you directly visit your fork.
* **2 or more forks**: show a pull-down menu (not the `fork-fragment`) with all your forks.

This way you can identify having **no forks** and **# amount of forks**. Conforming to (1).
I deliberately say "not the `fork-fragment`", because that's already accessible when pressing the normal <kbd>Fork</kbd> button. My suggestion would be to just create a pull-down menu listing all the forks when having 2 or more forks. Conforming to (2).

bfred-it:
> I think it is interesting to keep this close to the repo name and to explicitly show the target. If the user belongs to an organization and the fork is in the organization, not in the user's account, 'Visit your fork' might be misleading.

Indeed, therefore such button should only point to the user's own fork. Any other forks in their organizations will just be listed in the modal.

> Apart from the comment above, I think it is important to keep 'Make a fork' and 'Visit/view existing forks' in separate buttons/modals/menus.

My suggestion would still keep them as two buttons:
- Fork (open modal, standard github behavior)
- Open your fork (right next to it, as a shortcut)



> For me, the main advantage of this feature, is to (1) **quickly identify** that I have **no forks**, or **# amount of forks**. The second advantage is to (2) quickly go to one (or more) of my forks.

In my suggested solution, both of these advantages apply to your own fork, however it won't help for organizations' forks.

> I see your suggestion. Let me suggest an alternative:
> 
>     * **No forks**: we show no pull-down button.
> 
>     * **One fork**: let's you directly visit your fork.
> 
>     * **2 or more forks**: show a pull-down menu (not the `fork-fragment`) with all your forks.

That's alright too, it just requires a bit more code. This is what you're suggesting, right?

No forks: 

User fork:  (with icon from octicons)

Any other fork:  (opens dropdown)

Shortcut for this view:

```js
forkSummary.classList.add('BtnGroup-item');
forkSummary.closest('li').classList.add('BtnGroup');
forkSummary.closest('li').append(<a href="" class="btn btn-sm BtnGroup-item">•</a>);
```

karlhorky:
Interesting solutions. Feels to me like it's getting too complicated, but I defer to your wisdom as maintainers.

However, if I can make one suggestion, maybe a more descriptive button than <kbd>•</kbd>? Text is best, but if that doesn't work due to space constraints a more descriptive icon would be an alternative.

bfred-it:
`•` was a placeholder, the icon can be chosen from https://octicons.github.com/

jerone:
> This is what you're suggesting, right?

Yes, exactly. If others agree ofc. 😄

jerone:
I'll send a PR soon, have something working now.

bfred-it:
Closing this in favor of #2230

PR #2153, opened 2019-06-13
Add link to forked repo below the source

jerone:
### Description
Add links to the forked repos below the original repo. For users and organizations.

### Closes
Closes #1129
Supersedes attempt 1 #1108
Supersedes attempt 2 #2105
Supersedes attempt 3 #2141

### Screenshots
**Original repo:**
forked-to

**User forked repo:**
image

**Organization forked repo:**
image

### Test
1. Open any repo and click on the <kbd>Fork</kbd> button.
2. Fork the repo.
3. Go back to the original repo. 
4. Below the repo name, a link to your fork will presented.

PR #2215, opened 2019-07-08
Make cache key unique per username

jerone:
### Description
Make cache key unique per username. 
This will fetch all forks again.
Already cached forks will invalidate after 10 days anyways.

### Closes 
Closes https://github.com/sindresorhus/refined-github/pull/2153#issuecomment-509045202

/cc @1138-4EB & @bfred-it


bfred-it:
This by the way will lose the current cache, which will be forgotten within 10 days

PR #2216, opened 2019-07-08
Various fixes to `forked-to`

bfred-it:
Follows #2153 

Changes:

- Actually validate fork via HTTP request
- Don't validate current page
- Delete cache entry completely when fork is gone
- Avoid merging with existing cache when unnecessary (i.e. when opening the fork dialog)
- Get/validate/update cache in one step before showing the links
- Reduce number of functions

Currently untested

cc @jerone

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10671,6 +10671,14 @@
 			"resolved": "https://registry.npmjs.org/p-defer/-/p-defer-2.1.0.tgz",
 			"integrity": "sha512-xMwL9id1bHn/UfNGFEMFwlULOprQUEOg6vhqSfr6oKxPFB0oSh0zhGq/9/tPSE+cyij2+RW6H8+0Ke4xsPdZ7Q=="
 		},
+		"p-filter": {
+			"version": "2.1.0",
+			"resolved": "https://registry.npmjs.org/p-filter/-/p-filter-2.1.0.tgz",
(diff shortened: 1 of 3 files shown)


jerone:
I was trying something like the following to show the forks sooner:

```js
async function load(callback: () => void): Promise<void> {
	await elementReady('.pagehead h1.public');
	callback();
	await features.onAjaxedPages(callback);
}
```
But this code will ofc show double forks on some pages.

jerone:
I tested this and aldo it works, it is noticeable slower. Maybe the `validateFork` check should be executed **after** adding the forks HTML, on the cache, so next page load shows the validated forks.

bfred-it:
That works. Also I’ll have to avoid revalidation if it was done “recently”

I think a better strategy is to just fetch the fork box ***if*** there’s a fork in the cache:

- it will validate all forks and replenish the cache in one HTTP request

jerone:
> I think a better strategy is to just fetch the fork box [..]

The fork fragment? If yes, see https://github.com/jerone/refined-github/blob/9df50080dfddee5f7a2a6a1dc4465166339fedfe/source/features/forked-to.tsx#L7-L12 how to load the fragment.

bfred-it:
Now forks will be validated after being shown.

If you want to speed it up further with multiple forks, you can fetch the fork fragment, but for now let's see what happens in https://github.com/sindresorhus/refined-github/pull/2214

PR #2220, opened 2019-07-10
Fix missing `options.css`

sharkykh:
Options page was missing its `.css` file.

# Test
Open extension options.

### DIFF
--- a/source/options.tsx
+++ b/source/options.tsx
@@ -1,3 +1,4 @@
+import './options.css';
 import React from 'dom-chef';
 import select from 'select-dom';
 import linkifyUrls from 'linkify-urls';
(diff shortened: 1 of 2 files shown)


bfred-it:
Ouch, indeed. I think that this will affect other CSS files as well. I'd rather re-include `options.css` in `webpack.config.js` under `CopyFilesPlugin`

sharkykh:
@bfred-it 
I verified that it doesn't change anything in `content.css`.
Because these are split into separate entries (`source/content.ts` and `source/options.ts`) there shouldn't be any issue. Unless I'm missing something?

bfred-it:
Ah probably. I was thinking `[name]` matched whatever file the original imported CSS, not the name of the entry

sharkykh:
Yes, IIRC it uses the same template as [`output.filename`](https://github.com/sindresorhus/refined-github/blob/40b12d34bd5ed89a97274aaa127ca8b7d0a3ceff/webpack.config.ts#L42)

bfred-it:
This works, thanks!

PR #2218, opened 2019-07-09
Add mark files as viewed feature

lukaszklis:
Whenever I approve a PR I’d love to have all of the changed files marked as viewed.

# Test

https://github.com/lukaszklis/refined-github-mark-files-as-viewed/pull/1/files (can grant collaborators privileges upon request)

PR #2223, opened 2019-07-11
Fix checkbox selector for mark files as viewed

lukaszklis:
This morning when approving a PR I spotted a problem with the ”mark all files as viewed” feature I added in #2218. The selector I used to find ”Viewed” checkboxes did not take into the account that some of them might have already been checked.


notlmn:
I forgot to ask in the original PR, does this feature work for ajaxed file diffs?

lukaszklis:
@notlmn I’ve just checked and in fact, it will not work with ajaxed file diffs. :( I will add a bug ticket since I won’t have time for the next 1,5 weeks to take a closer look at it. Perhaps someone else could take care of it. :)

PR #2226, opened 2019-07-12
Fix branch link broken with `/` in the branch name

### LINKED ISSUE #2049: Master branch links are broken for branches with a `/` in the branch name
Master branch links are broken for branches with a `/` in the branch name.

eg. branch `sk/dev` links to `[repo]/tree/master/dev`

bug is here: https://github.com/sindresorhus/refined-github/blob/master/source/features/branch-buttons.tsx#L78 
and here: https://github.com/sindresorhus/refined-github/blob/master/source/libs/page-detect.ts#L84

kevin940726:
Closes #2049 

I took the closet approach I can think of via a `<link>` tag in head which looks like follow.

```html
<link href="https://github.com/sindresorhus/refined-github/commits/master.atom" rel="alternate" title="Recent Commits to refined-github:master" type="application/atom+xml">
```

Looks like every pages have it, and we can get the current branch from it stably. Please let me know if there is any concern.

I tried getting the current branch from the dropdown menu, but they are lazily loaded. Then I tried to get it from the `title` attribute of the dropdown button, it works for branch, but not for tag. I want to have a more general solution so I took this approach.

# Test

https://github.com/kevin940726/refined-github/blob/fix/branch-buttons/.travis.yml

### DIFF
--- a/source/features/branch-buttons.tsx
+++ b/source/features/branch-buttons.tsx
@@ -7,17 +7,7 @@ import features from '../libs/features';
 import {isRepoRoot} from '../libs/page-detect';
 import {groupSiblings} from '../libs/group-buttons';
 import getDefaultBranch from '../libs/get-default-branch';
-import {getRepoURL, getRepoPath, getOwnerAndRepo, getCurrentBranch} from '../libs/utils';
-
-function replaceBranch(currentBranch: string, newBranch: string): string {
-	// Should be either `/tree/<branch>` or `/blob/<branch>/<filePath>`
-	const repoPath = getRepoPath()!;
-	// Replace the first matched branch name to the new one
-	const newBranchRepoPath = repoPath.replace(currentBranch, newBranch);
-
-	// Join the repo url with new branch repo path
-	return `/${getRepoURL()}/${newBranchRepoPath}`;
-}
+import {getRepoURL, getOwnerAndRepo, getCurrentBranch, replaceBranch} from '../libs/utils';
 
 async function getTagLink(): Promise<'' | HTMLAnchorElement> {
 	const {ownerName, repoName} = getOwnerAndRepo();
--- a/source/features/branch-buttons.tsx
+++ b/source/features/branch-buttons.tsx
@@ -7,15 +7,7 @@ import features from '../libs/features';
 import {isRepoRoot} from '../libs/page-detect';
 import {groupSiblings} from '../libs/group-buttons';
 import getDefaultBranch from '../libs/get-default-branch';
-import {getRepoURL, getOwnerAndRepo} from '../libs/utils';
-
-function getCurrentBranch(): string {
-	const commitLink = select('link[rel="alternate"]')!;
-	// Will be something like https://github.com/sindresorhus/refined-github/commits/master.atom
-	const href = commitLink.getAttribute('href')!;
-	const group = /\/commits\/(.+)\.atom$/.exec(href)!;
-	return group[1];
-}
+import {getRepoURL, getOwnerAndRepo, getCurrentBranch} from '../libs/utils';
 
 async function getTagLink(): Promise<'' | HTMLAnchorElement> {
 	const {ownerName, repoName} = getOwnerAndRepo();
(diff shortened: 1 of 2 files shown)


source/libs/utils.ts
```diff
@@ -30,6 +30,15 @@ export const getRepoBranch = (): string | false => {
 	return false;
 };
 
+export const getCurrentBranch = (): string | undefined => {
+	const recentCommitsLink = select('link[title^="Recent Commits to"]')!;
+	// Will be something like https://github.com/sindresorhus/refined-github/commits/master.atom
+	const href = recentCommitsLink.getAttribute('href')!;
+	const group = /\/commits\/(.+)\.atom$/.exec(href);
+
+	return group ? group[1] : undefined;
```
bfred-it:
```ts
return select('[aria-current="page"]')!
	.pathname // <user>/<repo>/tree/<branch/slashed-branch>
	.split('/')
	.slice(4) // Drops everything and including `tree`
	.join('/');
```

source/features/branch-buttons.tsx
```diff
@@ -74,9 +72,7 @@ async function getDefaultBranchLink(): Promise<HTMLElement | undefined> {
 	if (isRepoRoot()) {
 		url = `/${getRepoURL()}`;
 	} else {
-		const urlParts = location.pathname.split('/');
-		urlParts[4] = defaultBranch; // Change branch of current blob/tree
-		url = urlParts.join('/');
+		url = location.pathname.replace(currentBranch, defaultBranch);
```
bfred-it: This isn't safe, it will fail on `/sindresorhus/ava/blob/ava`

kevin940726: You're right! I forgot about that, thx!

source/libs/utils.ts
kevin940726: LGTM, it seems to be more resilient than my approach.

kevin940726: The only problem I can think of is that when we are in the repo root (`https://github.com/sindresorhus/refined-github`), the above code will yield an empty string. It might works just fine with this feature since we have another check about whether we are in repo root, but might not be appropriate for a more general utility function. Should we fallback to default branch in this case? Is it possible?

source/features/branch-buttons.tsx
```diff
@@ -40,7 +40,7 @@ async function getTagLink(): Promise<'' | HTMLAnchorElement> {
 
 	const link = <a className="btn btn-sm btn-outline tooltipped tooltipped-ne">{icons.tag()}</a> as unknown as HTMLAnchorElement;
 
-	const currentBranch = select('.branch-select-menu .css-truncate-target')!.textContent;
```
kevin940726: Actually I'm curious about why we chose to select the current branch like this before. AFAIK, it would only selected the truncated branch name but not the full name. Is there any concerns I don't know about?

bfred-it: The name is only truncated via CSS so in the DOM it's still full

source/libs/utils.ts
bfred-it: Let's revert to `select('.branch-select-menu .css-truncate-target').textContent`, it works all the time.

kevin940726: I'm still getting `branch/still-b…` instead of `branch/still-branch` by using that [here](https://github.com/simonkotwicz/refined-github/blob/branch/still-branch/.travis.yml#L6). I might have did something wrong?

bfred-it: Indeed, they must have changed it. This seems to work:

```js
select('.branch-select-menu .css-truncate').title
```

kevin940726: Like I stated in the PR description, I think this would only work for branch but not for tag? (I don’t have my laptop to verify now)

sharkykh:
@bfred-it An empty file made it in:
https://github.com/sindresorhus/refined-github/blob/master/source/libs/get-current-branch.ts

bfred-it:
Thanks! Fixed

bfred-it:
And thanks @kevin940726 for your contribution 🥇

Issue #1065, opened 2018-02-05
Show actor real name in dashboard events


hkdobrev:
Similar to comments where we show the user's name next to their username, that would be useful to show the full name of the actor performing the action.

![image](https://user-images.githubusercontent.com/506129/35797953-01bbd00a-0a6a-11e8-9de5-1a46ae42369d.png)

In this example, I didn't realise `gaeron` is Dan Abramov, but this would have been useful to know.

bfred-it:
Do we really want to add the full name everywhere? 

#922 went the opposite way: leave only the avatar

hkdobrev:
If you work with someone on a regular basis the avatar is definitely the thing you can recognise them by quickly while scanning content - that's just how the brain works - images are quicker to recognise as we process them on a very basic level very fast. This is what I constantly use in conversations to recognise who said what and who participated (I have another suggestion about that).

However, in the dashboard, I would have more people who I've followed, but don't see their avatars on a day-to-day basis. Here, I've recognised his avatar only because I was referred to some of his Tweets the other day. But I remember quite more names, so having the famous name there would bring my attention to that post if I read the name. I can decide to look further into the starred repo.

busches:
I'd like the name, as 99% of people in my enterprise appears to be OK using the default avatar.

PR #2229, opened 2019-07-12
Add full names to events on for dashboard (extends `show-name`)

jerone:
# Description
Show full names in dashboard.

image

# Closes
Closes https://github.com/sindresorhus/refined-github/issues/1065

# Test
* Go to your dashboard on https://github.com
* See full names being loaded on the dashboard feed.
* Scroll to the bottom, to load more.
* See full names being loaded on the dashboard feed again.

# Conflicting
This will conflict with https://github.com/sindresorhus/refined-github/pull/2193 but everything can be solved 😄


source/features/show-names.tsx
```diff
@@ -1,20 +1,42 @@
...
-	// `a` selector needed to skip commits by non-GitHub users
-	const usernameElements = select.all('.js-discussion a.author:not(.rgh-fullname):not([href*="/apps/"]):not([href*="/marketplace/"]):not([data-hovercard-type="organization"])');
+	if (features.isDashboard()) {
+		observer.observe((await elementReady('#dashboard .news'))!, {childList: true});
+	}
+
+	await usernames();
+}
+
+async function usernames(): Promise<false | void> {
```
bfred-it: `usernames()` is not descriptive, use a verb

source/features/show-names.tsx
```diff
@@ -1,20 +1,42 @@
...
+			observer.observe(node, {childList: true});
+		}
+	}
+});
+
 async function init(): Promise<false | void> {
-	// `a` selector needed to skip commits by non-GitHub users
-	const usernameElements = select.all('.js-discussion a.author:not(.rgh-fullname):not([href*="/apps/"]):not([href*="/marketplace/"]):not([data-hovercard-type="organization"])');
+	if (features.isDashboard()) {
+		observer.observe((await elementReady('#dashboard .news'))!, {childList: true});
```
bfred-it: `onNewComments` waits for `domReady` and `elementReady` stops checking after `domReady`, i.e. this works exactly like a regular `select`

bfred-it: Also you can probably use `observeEl` like:

```js
async updateAndWatch([{addedNodes}], observer) {
	await usernames();

	// Observe the new ajaxed-in containers
	for (const node of addedNodes) {
		if (node instanceof HTMLDivElement) {
			observer.observe(node, {childList: true});
		}
	}
}

	if (features.isDashboard()) {
		observeEl('.news', updateAndWatch)
	} else {
		await usernames(); // This is called by observeEl, so it needs to be inside `else`
	}
}
```

source/features/show-names.tsx
```diff
@@ -1,20 +1,42 @@
...
+	}
+
+	await usernames();
+}
+
+async function usernames(): Promise<false | void> {
+	const usernameElements = select.all([
+		'.js-discussion a.author:not(.rgh-fullname):not([href*="/apps/"]):not([href*="/marketplace/"]):not([data-hovercard-type="organization"])', // `a` selector needed to skip commits by non-GitHub users.
+		'#dashboard a.text-bold[data-hovercard-type="user"]:not(.rgh-fullname)' // On dashboard `.text-bold` is required to not fetch avatars.
```
bfred-it: Thanks for the comment! They help a lot on selectors, they tend to be "simplificable" while they may contain important exclusion logic.

source/features/show-names.tsx
```diff
@@ -1,20 +1,42 @@
 import './show-names.css';
 import React from 'dom-chef';
 import select from 'select-dom';
+import elementReady from 'element-ready';
 import * as api from '../libs/api';
 import features from '../libs/features';
 import {getUsername} from '../libs/utils';
 
+const observer = new MutationObserver(async ([{addedNodes}]) => {
+	await usernames();
```
bfred-it: `await` doesn't actually block the observer, so we don't need it

source/features/show-names.tsx
jerone: Running above (slightly altered) code results in an error, which doesn't makes sense to me:

image

jerone: See https://github.com/sindresorhus/refined-github/pull/2229/commits/0e1288059512f1db11b4d3e6e799b2ad4c957d56

bfred-it: Indeed, that's an ugly error. It's because the function parameter is being destructured as `[{addedNodes}]` but `observeEl` only calls that callback with `[]`, so it fails when it tries to destructure `parameter[0]` (`undefined`) into `{addedNodes}`

Fix:

```ts
async function updateAndWatch([firstChange]: MutationRecord[], observer: MutationObserver): Promise<void> {
...
if (firstChange) {
	// use firstChange.addedNodes
```

jerone: Implemented. Thanks.

PR #2230, opened 2019-07-12
Change `forked-to` links to button

jerone:
# Description
Change the `forked-to` links below the repo name, to a button next to the <kbd>Fork</kbd> button.

The design changes depending on the amount of forks:
* **0** = no extra button.
* **1** = extra button is link to your fork.
* **> 1** = extra button opens menu, which contains list of all your forks.

# Screenshots
image

<details><summary>Design 1</summary>

_0 forks:_
image

_1 fork:_
image

_2 forks (https://github.com/sindresorhus/refined-github/pull/2230/commits/706d2760fc7be3ff8dfc866067a74366e1078c5b):_
image

_Play nice with `hide-watch-and-fork-count` feature:_
image
</details>

# Closes
Discussion https://github.com/sindresorhus/refined-github/pull/2214 & https://github.com/sindresorhus/refined-github/pull/2216
Follows https://github.com/sindresorhus/refined-github/pull/2153 & https://github.com/sindresorhus/refined-github/pull/2216
Supersedes https://github.com/sindresorhus/refined-github/pull/2214

---

/cc @bfred-it, @karlhorky, @1138-4EB


bfred-it:
> _forks:_
> image

I just realized that this is the same exact content that the regular Fork button, minus the change to create more forks. It doesn't make much sense to have two buttons that do the same thing.

Suggestions:

- 1 fork is fine like you're showing
- when there are multiple forks:
	- add a ▼ to the Fork button (as a visual indicator)
	- preload the lightbox
- always:
	- listen to the lightbox loading and adjust its DOM to display as a dropdown instead; maybe its content should just change from `details-dialog` to `details-menu`

Thoughts?

jerone:
Not sure I like this better, but here it is:

_2 forks button:_
image
_2 forks menu:_
image

1138-4EB:
> Not sure I like this better, but here it is:

I definitely think that the first one is better than this. I believe the way to go would be to add the 'Create new fork' item as an 'additional fork item' in https://user-images.githubusercontent.com/55841/61155136-6bd8fe00-a4f0-11e9-941c-e55647a8eadf.png. Hence, in practice, this PR will 'convert the modal into a dropdown menu'.

jerone:
I definitely like the first custom dropdown menu, but I do agree with @bfred-it that the custom menu and build-in dialog are almost the same.

What if we just replace the build-in fork dialog with the custom dropdown menu. And append (like @1138-4EB said) an menu-item to [create a fork](https://github.com/sindresorhus/refined-github/fork) (an action that is only done once per user/organization, accepting the one extra click).

jerone:
Design 3; this replaces the build-in fork dialog with an custom fork dropdown menu. It includes 2 or more fork links, plus a link to create a new fork.

_0 forks:_
image

_1 fork:_
image

_2 forks (https://github.com/sindresorhus/refined-github/pull/2230/commits/e9522f170ab1e3d14ceaab1cdaa057052687f1f8), button:_
image

_2 forks (https://github.com/sindresorhus/refined-github/pull/2230/commits/e9522f170ab1e3d14ceaab1cdaa057052687f1f8), menu:_
image

_P.s. hiding the build-in dialog should be improved._

bfred-it:
This would then slow down the creation of more forks. My suggestion was to reuse the same light box content (all of it, but adjusted with some classes) to look like a dropdown. The dark shadow can also be dropped

jerone:
[**@bfred-it**](https://github.com/bfred-it) commented on [Jul 14, 2019, 4:09 AM GMT+2](https://github.com/sindresorhus/refined-github/pull/2230#issuecomment-511167112 "2019-07-14T02:09:02Z - Replied by Github Reply Comments"):
> This would then slow down the creation of more forks. My suggestion was to reuse the same light box content (all of it, but adjusted with some classes) to look like a dropdown. The dark shadow can also be dropped

Design 2 thus: https://github.com/sindresorhus/refined-github/pull/2230#issuecomment-511146632 (without the dark shadow) ?

bfred-it:
Design 3 of course is better-looking but it's best to also include the other links in that lightbox:

```
# Your forks
% user/repo
% org/repo

# Create fork
% org2
% org3
% org4 (no permission) `disabled`
```

As for our cache, we only need to store the first fork, for 2 reasons:

- a filled cache means we have to validate it, i.e. we'll have to preload the dropdown
- we don't actually need the list of forks in the cache because we'll only ever show it after the dropdown is preloaded

jerone:
Just an heads-up; I'm still working on this. The button is done, and the forks list is created, I just need to read the fork dialog to get the missing organizations. To be continued...

fregante:
Something to keep in mind: users without organizations don't have a popup:


```html
<form class="btn-with-count" action="/chartjs/Chart.js/fork" accept-charset="UTF-8" method="post">
            <button class="btn btn-sm btn-with-count" data-hydro-click="{&quot;event_type&quot;:&quot;repository.click&quot;,&quot;payload&quot;:{&quot;target&quot;:&quot;FORK_BUTTON&quot;,&quot;repository_id&quot;:8843683,&quot;client_id&quot;:&quot;1512404804.1558770078&quot;,&quot;originating_request_id&quot;:&quot;F8AF:14C9:984A6D:DCBCD6:5D3EC08B&quot;,&quot;originating_url&quot;:&quot;https://github.com/chartjs/Chart.js/tree/6632b8ba8475e020fd0533e2630117376a9cbc73&quot;,&quot;referrer&quot;:null,&quot;user_id&quot;:53252526}}" data-hydro-click-hmac="4fc7804f0046d6c8bdca3e655e7d8efee5bf3569d3f8ce0683b5f8515fa86bc8" data-ga-click="Repository, show fork modal, action:files#disambiguate; text:Fork" type="submit" title="Fork your own copy of chartjs/Chart.js to your account" aria-label="Fork your own copy of chartjs/Chart.js to your account">              {SVG icon}
              Fork
</button></form>
```

This code seems unchanged whether I already have the fork or not

jerone:
_x forks:_
image


> % org4 (no permission) `disabled`

There [appears](https://primer.style/css/components/select-menu-deprecated) to be no disabled state for select-menu.

fregante:
Looks good in the screenshot! Is this done?

source/features/forked-to.tsx
```diff
@@ -51,25 +52,144 @@ async function updateForks(forks: string[]): Promise<void> {
...
+			</div>
+		);
+		detailsDialog.append(...toBeForked);
+	}
+
+
+	const forkButton = select('summary[title^="Fork your own copy of"]')!;
+	forkButton.after(detailsDialog);
+
+	forkDialog!.remove();
```
fregante: This function does a lot, maybe we can't afford to maintain all this code for this feature. We should find shortcuts.

This shorter code gets us 80% there:


```js
// Code written in console
const dialog = $('[src$="/fork?fragment=1"]');
const menu = document.createElement('details-menu');

// Copies the attributes from `dialog` and from the `watch` `details-menu` item
[...dialog.attributes, ...$('.select-menu-modal').attributes].forEach(attribute => menu.setAttribute(attribute.name, attribute.value));

// Moves the content
menu.append(...dialog.childNodes);
dialog.replaceWith(menu);

// Final style fixes
menu.parentElement.classList.remove('details-overlay-dark');
menu.classList.add('right-0');

menu.querySelector('include-fragment').addEventListener('load', () => {
	menu.querySelectorAll('.p-4').forEach(el => el.classList.replace('p-4', 'p-2'));
	menu.querySelectorAll('.text-normal').forEach(el => el.remove());
})
```

jerone: I much rather go back to [design 3](https://github.com/sindresorhus/refined-github/pull/2230#issuecomment-511156020) and for 2 or more forks let the "Create fork" menuitem go to the [fork page](https://github.com/sindresorhus/refined-github/fork).

fregante: > I much rather go back to [design 3](https://github.com/sindresorhus/refined-github/pull/2230#issuecomment-511156020)

Design 3 is only better than 1 because of this comment on 1:

> I just realized that this is the same exact content that the regular Fork button, minus the chance to create more forks. It doesn't make much sense to have two buttons that do the same thing.

All that design 3 does, compared to 1, is move the "regular fork button" to the dropdown... only to go back to the same comment (2 buttons, similar comment) + adding an extra step to create forks.

---

So if my last suggestion isn't liked, I'd go to the first commit https://github.com/sindresorhus/refined-github/pull/2230/commits/706d2760fc7be3ff8dfc866067a74366e1078c5b

jerone: I completely agree about the maintainability of [design 4](https://github.com/sindresorhus/refined-github/pull/2230#issuecomment-520178324); even creating it was a lot of hassle. 
Using a [lesser design](https://github.com/sindresorhus/refined-github/pull/2230#discussion_r313044618) to just get the dropdown menu and one step less, is not satisfying.

I'm more then happy to go back to [design 1](https://github.com/sindresorhus/refined-github/pull/2230#issue-297197052); no extra steps, easy visible to see the amount of forks, easy access to your fork(s) and maintainable code.

fregante: # 🙌

jerone:
@fregante This PR is restored to [design 1](https://github.com/sindresorhus/refined-github/pull/2230#issue-297197052). Feel free to make changes if required.

source/features/forked-to.tsx
```diff
@@ -57,19 +59,48 @@ async function init(): Promise<void> {
...
+					className="select-menu-button float-left btn btn-sm btn-with-count rgh-forked-button"
+					title="Open any of your forks"></summary>
+				<details-menu
+					style={{zIndex: 99}}
+					className="select-menu-modal position-absolute right-0 mt-5">
+					<div className="select-menu-header">
+						<span className="select-menu-title">Your forks</span>
+					</div>
+					{...forks.map(fork =>
+						<a href={`/${fork}`}
```
sindresorhus:
```suggestion
						<a
							href={`/${fork}`}
```

source/features/forked-to.tsx
```diff
@@ -57,19 +59,49 @@ async function init(): Promise<void> {
...
+			</details>
+		);
 	}
 
 	// Validate cache after showing links once, to make it faster
-	await updateForks(forks || []);
+	await updateForks(forks);
 }
 
 features.add({
```
fregante: Before merging:

- Update screenshot here and in readme
- Update the description here and in readme (and include trailing period). Probably "Adds a shortcut to your forks next to the `Fork` button on the current repo."

Possible future changes:

- `await updateForks(forks)` should probably be replaced by that code that triggers the lightbox to load, so our existing `load` handler can check all the forks at once. `updateForks` and `validateFork` can then be dropped altogether.
- I think #2334 hasn't been by this PR

fregante:
Thanks! This makes the feature more _refined!_

I'll keep this as the last comment as a to-do

>  - [ ] `await updateForks(forks)` should probably be replaced by that code that triggers the lightbox to load, so our existing `load` handler can check all the forks at once. `updateForks` and `validateFork` can then be dropped altogether.

PR #1873, opened 2019-03-22
Add `view-markdown-source` feature

bfred-it:
Closes #946 

# Test

https://github.com/sindresorhus/refined-github/blob/master/readme.md

PR #2232, opened 2019-07-13
Highlight permalinked line in `view-markdown-source`

karlhorky:
It seems like f7e0d05b0f09c46833c2411ab12997c73963958d from https://github.com/sindresorhus/refined-github/pull/1873 does not highlight the selected lines currently:

Screen Shot 2019-07-13 at 14 39 14

Switching out the hash once to an empty string and back to the original value causes the highlight to be applied.

cc @bfred-it

### DIFF
--- a/source/features/view-markdown-source.tsx
+++ b/source/features/view-markdown-source.tsx
@@ -93,7 +93,7 @@ async function init(): Promise<false | void> {
 
 	// Add support for permalinks to the code
 	if (location.hash.startsWith('#L')) {
-		showSource();
+		await showSource();
 		// Enable highlighting
 		window.dispatchEvent(new HashChangeEvent('hashchange', {
 			oldURL: location.href,
--- a/source/features/view-markdown-source.tsx
+++ b/source/features/view-markdown-source.tsx
@@ -94,6 +94,10 @@ async function init(): Promise<false | void> {
 	// Add support for permalinks to the code
 	if (location.hash.startsWith('#L')) {
 		showSource();
+		// Enable highlighting
+		const originalHash = location.hash;
+		location.hash = '';
+		location.hash = originalHash;
 	}
 }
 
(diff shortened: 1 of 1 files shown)


source/features/view-markdown-source.tsx
```diff
@@ -94,6 +94,10 @@ async function init(): Promise<false | void> {
 	// Add support for permalinks to the code
 	if (location.hash.startsWith('#L')) {
 		showSource();
+		// Enable highlighting
+		const originalHash = location.hash;
+		location.hash = '';
+		location.hash = originalHash;
```
bfred-it: This will cause 2 history changes (i.e. to go back in history, you'll have to press `back` 2 times without visual change). Any other way to trigger `:target`?

At the very least, line 99 should use `replaceState` to remove the hash without creating an entry, so it only causes 1 history change.

karlhorky: Good point! Taking a look.

karlhorky:
@bfred-it ok I've switched to dispatching a HashChangeEvent, which seems to work (at least when I fire it manually from the browser console).

karlhorky:
Wonder if there's any possibility of race condition if `showSource()` is not finished yet... If we see this is the case during testing, we could wait for application of necessary DOM modifications.

bfred-it:
> @bfred-it ok I've switched to dispatching a HashChangeEvent, which seems to work (at least when I fire it manually from the browser console).

Neat! Can you verify that it works in both Chrome and Firefox?

Also, `showSource` is a promise so you probably just need to prepend `await` on line 96

karlhorky:
Ok, verified that it works in Firefox Devtools too. And I'm awaiting the promise now.

bfred-it:
Amazing, this works perfectly. Thank you for your contribution! 💡

Example https://github.com/sindresorhus/refined-github/blob/master/readme.md#L1

karlhorky:
Great, updated extension works well! Thanks for the guidance and merge!

PR #2235, opened 2019-07-13
Fix `minimize-upload-bar` on new issue

### LINKED ISSUE #2234: Can I have a Attach Files button in the issue creation description form?
Can I have a Attach Files button in the issue creation description form?

jerone:
# Description
Can't use `:last-child` selector, because the last child inside the markdown-toolbar on the new issue page is an hidden element.

# Closes 
Fixes #2234.

# Test
1. https://github.com/sindresorhus/refined-github/issues/new?labels=bug&template=bug_report.md

### DIFF
--- a/source/features/minimize-upload-bar.tsx
+++ b/source/features/minimize-upload-bar.tsx
@@ -6,15 +6,13 @@ import features from '../libs/features';
 import * as icons from '../libs/icons';
 
 async function addButton(): Promise<void> {
-	for (const toolbar of select.all('form:not(.rgh-minimize-upload-bar) markdown-toolbar')!) {
-		// Can't use `:last-child` selector, because new issue contains an hidden element.
-		const toolbarGroup = select.all(':scope > :not(.Details-content--hidden)', toolbar).slice(-1)[0]!;
-		toolbarGroup.prepend(
+	for (const toolbarButton of select.all('md-ref')) {
+		toolbarButton.after(
 			<button type="button" className="toolbar-item tooltipped tooltipped-n rgh-upload-btn" aria-label="Attach files">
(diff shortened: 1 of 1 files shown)


bfred-it:
The `collapsible-content-button` is there; use this

https://github.com/sindresorhus/refined-github/blob/11b8ab4ef75b824cbbd2ed2dd1e3a80f24070b46/source/features/collapsible-content-button.tsx#L31

source/features/minimize-upload-bar.tsx
```diff
@@ -6,13 +6,13 @@ import features from '../libs/features';
...
-	for (const toolbarGroup of select.all('form:not(.rgh-minimize-upload-bar) markdown-toolbar > :last-child')) {
-		toolbarGroup.prepend(
+	for (const toolbarButton of select.all('md-ref')) {
+		toolbarButton.after(
 			<button type="button" className="toolbar-item tooltipped tooltipped-n rgh-upload-btn" aria-label="Attach files">
 				{icons.cloudUpload()}
 			</button>
 		);
-		toolbarGroup.closest('form')!.classList.add('rgh-minimize-upload-bar');
+		toolbarButton.closest('form')!.classList.add('rgh-minimize-upload-bar');
```
bfred-it: This class is no longer used. Is it necessary? Will the current code cause duplicates when new comments are loaded?

jerone: I couldn't get any duplicate buttons. That's why I removed the `form:not(.rgh-minimize-upload-bar)` selector above. 
But this class is still used and necessary in the attached stylesheet.

bfred-it:
Test

Issue #1900, opened 2019-04-01
Filter by "Draft PRs only" in PR list


FloEdelmann:
In a repo's PR list, it would sometimes be useful to see only draft pull requests (e.g. to close long stale drafts or look which can be made "ready to review"). This can be done by adding `is:draft` to the search term (see [GitHub help article](https://help.github.com/en/articles/searching-issues-and-pull-requests#search-for-draft-pull-requests); unfortunately `is:pr is:open -is:draft` doesn't work as expected :confused:).

I see two possibilities where this could be added:

1. On the left "Open/Close" switcher: Add a "Draft" state. Problems:
	- Takes up much space, which could reintroduce #1830.
	- A draft PR is also "open".
    - The number of open PRs would likely need to be fetched via the API.
2. In the "Reviews" filter menu: Add a "Not ready for review (Draft PR)" item. Problem:
	- Not very discoverable.

bfred-it:
> In the "Reviews" filter menu: Add a "Not ready for review (Draft PR)" item. 

I think this is good enough

FloEdelmann:
Should it be a new feature (like `filter-draft-prs`) or part of `filter-pr-by-build-status`? It's the only feature that adding it to might be considered. Maybe then rename the feature to simply `filter-prs`?

bfred-it:
Yeah it might make sense to make a `pr-filters` feature

sharkykh:
> unfortunately `is:pr is:open -is:draft` doesn't work as expected 😕

Oddly enough, it works as expected on the global GitHub search... 🤔
[query: `repo:sindresorhus/refined-github is:pr is:open -is:draft`](https://github.com/search?q=repo%3Asindresorhus%2Frefined-github+is%3Apr+is%3Aopen+-is%3Adraft)
([currently there is one draft PR open](https://github.com/search?q=repo%3Asindresorhus%2Frefined-github+is%3Apr+is%3Aopen+is%3Adraft))

FloEdelmann:
I've researched a bit more:

Combining `is:draft`/`-is:draft` with `is:open` / `is:closed` is not respected at all in PR lists:

[...]

PR #2236, opened 2019-07-14
Fix: Unwanted scroll bar in quotes with an Emoji

sharkykh:
## Comparison
Might be a little hard to tell, but the Emoji is clipped on the bottom.
[Comparison with a slider](https://cdn.knightlab.com/libs/juxtapose/latest/embed/index.html?uid=6d42935a-a612-11e9-b9b8-0edaf8f81e27)

#### Before:
Before
#### After:
After

## Test
https://github.com/sindresorhus/refined-github/issues/1900#issuecomment-511180163

### DIFF
--- a/source/features/scrollable-code-and-blockquote.css
+++ b/source/features/scrollable-code-and-blockquote.css
@@ -5,6 +5,13 @@
 	overflow-y: auto;
 }
 
+/*
+A tiny scroll bar appears when the last paragraph of a quote contains an Emoji.
+*/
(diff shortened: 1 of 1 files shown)


bfred-it:
Indeed, also happens on macOS Firefox but I never noticed because my scrollbars are hidden.

bfred-it:
Thank you for the bugfixes! 🎇

sharkykh:
Apparently this is an issue too...

> A quote `with some code in it`...

Issue #1374, opened 2018-06-14
API calls need Personal Token to support frequent calls and private repos


bfred-it:
Refined GitHub needs

- [x] [Personal token](https://github.com/settings/tokens) field in the options
- [x] Automatic token usage by https://github.com/sindresorhus/refined-github/blob/master/source/libs/api.js, if present

notlmn:
Features that can use API:
* Embedded gists (https://developer.github.com/v3/gists/#get-a-single-gist)
* CI/CD status links (https://developer.github.com/v3/repos/statuses/#get-the-combined-status-for-a-specific-ref)
* Get users full name (https://developer.github.com/v3/users/#get-a-single-user)

Can also use GraphQL API to request only required data.

yakov116:
Maybe its just me, I feel very unconformable with the token being displayed with one click. In your issue linker its 2 steps (1. click the ext. 2. Click options).

bfred-it:
> In your issue linker its 2 steps (1. click the ext. 2. Click options).

How is it different here? The token appears in the options

yakov116:
Its 2 steps

1.
![image](https://user-images.githubusercontent.com/16872793/41508166-2a22dcee-720e-11e8-8008-977e34dec904.png)
Then Step 2 opens the options

bfred-it:
Right, so is Refined GitHub

yakov116:
@bfred-it clicking on the ext shows the token.

bfred-it:
Why is your button not hidden in the overflow? It's pretty much useless

yakov116:
>Why is your button not hidden in the overflow? It's pretty much useless

What?

![image](https://user-images.githubusercontent.com/16872793/41510506-ca2fbdea-7233-11e8-8c06-38fcc3680135.png)

bfred-it:
This overflow:

<img width="332" alt="extensions overflow" src="https://user-images.githubusercontent.com/1402241/41510511-2f8bad3e-7266-11e8-9cc5-fdbe662a66ec.png">

yakov116:
@bfred-it 👍, ok...

Would rather not do that.

bfred-it:
@sindresorhus thoughts?

sindresorhus:
If someone has access/control of your computer, all bets are off anyway, so I don't fully agree with your reasoning. That being said, I don't mind making it a password type input. GitHub hides it too: https://github.com/settings/tokens

yakov116:
[...]

PR #2257, opened 2019-07-19
used input type="password" for personalToken

trivikr:
# Test
* Confirmed that `npm build` is successful and tested unpacked extension in browser
* `npm test` is successful

# Screenshots
* Before 
input-type-text
* After 
input-type-password


yakov116:
@trivikr 	see #1374 and https://github.com/sindresorhus/refined-github/commit/a60638ad250c067e9f5753e3e82906a44eb6099b
It gives a false sense of security (since its plain text)

However the size show be made to 40% now

bfred-it:
As @yakov116 said, but thanks for the PR 🙂

bfred-it:
@trivikr if that's your real token in the screenshot I suggest changing it because the missing characters are easily guessable

trivikr:
@bfred-it I'd created and destroyed a token just for the screenshot

PR #2259, opened 2019-07-19
Shorten personal token field in options for privacy

yakov116:
Do to layout changes the box was made wider so 50% is still too big

Closes #2257


bfred-it:
Let’s just use `em`  instead of %

Issue #1096, opened 2018-02-17
Add button to preview HTML files


benwaffle:
very useful for .html files

sindresorhus:
Can you please elaborate a bit more about what you're looking for?

benwaffle:
When i'm browsing a github repo and I want to see the webpage of an html file, i will paste it into rawgit.com

e.g. I see this file https://github.com/CodingTrain/website/blob/master/CodingChallenges/CC_11_PerlinNoiseTerrain_p5.js/index.html
so i go to http://rawgit.com, paste the URL, and get this: https://rawgit.com/CodingTrain/website/master/CodingChallenges/CC_11_PerlinNoiseTerrain_p5.js/index.html

bfred-it:
The difference is subtle: 

```
RawGit serves raw files directly from GitHub with proper Content-Type headers.
```

I wouldn't add another Raw button that is _only sometimes_ different.

I probably wouldn't replace the current Raw button with one linked to rawgit (maybe? because it breaks expectations)

Not sure how discoverable it would be to add a _modifier_+click on the Raw button to open rawgit instead

bfred-it:
Perhaps we can add a `View`/`Preview` button for HTML files, I think that's the only case where rawgit links are **visibly** different.

benwaffle:
That would be useful

theel0ja:
Something like this in Gists when the file is HTML:

![image](https://user-images.githubusercontent.com/5832930/39673027-75c8e3b0-513d-11e8-94e0-e0e8052e041b.png)

sgabler:
I think this issue can be closed, as RawGit will soon shut down: https://rawgit.com/

> RawGit is now in a sunset phase and will soon shut down. It's been a fun five years, but all things must end.
> 
> GitHub repositories that served content through RawGit within the last month will continue to be served until at least October of 2019. URLs for other repositories are no longer being served.
> 
> If you're currently using RawGit, please stop using it as soon as you can.

bfred-it:
Are there any alternative services?

bfred-it:
No way! They’re closing it because of spam, no one here is gonna clone the service (and its issues) just to preview HTML files 😃

bfred-it:
Staticaly.com seems to be an alternative service and it works correctly with HTML files:

[...]

PR #2265, opened 2019-07-22
Add: `jsdelivr-cdn`

Rawnly:
# Features
- Adds a new button near the `Copy` button that lets you copy a cdn url from jsdelivr, the pattern for the url is: `https://cdn.jsdelivr.net/gh/<user>/<repo>/<path-to-file>`

# Test
[This one](https://github.com/twbs/bootstrap/blob/master/dist/css/bootstrap.min.css) `https://github.com/twbs/bootstrap/blob/master/dist/css/bootstrap.min.css`

Should generate the [following](https://cdn.jsdelivr.net/gh/twbs/bootstrap/dist/css/bootstrap.min.css)
`https://cdn.jsdelivr.net/gh/twbs/bootstrap/dist/css/bootstrap.min.css`

### DIFF
--- a/source/content.ts
+++ b/source/content.ts
@@ -43,6 +43,7 @@ import './features/improve-shortcut-help';
 import './features/deprioritize-marketplace-link';
 import './features/view-markdown-source';
 import './features/copy-file';
+import './features/jsdelivr-cdn';
 import './features/hide-own-stars';
 import './features/infinite-scroll';
 import './features/hide-empty-meta';
--- a/source/features/jsdelivr-cdn.tsx
+++ b/source/features/jsdelivr-cdn.tsx
@@ -1,0 +1,66 @@
+import React from 'dom-chef';
+import select from 'select-dom';
+import copyToClipboard from 'copy-text-to-clipboard';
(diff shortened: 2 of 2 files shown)


bfred-it:
- This only applies to JavaScript files
- jsdelivr is only one of the possible CDNs people might be interested in and we can’t add them all

so it doesn’t belong to RGH. It’s best you keep this as a bookmarklet or make another extension/userscript

Rawnly:
> * jsdelivr is only one of the possible CDNs people might be interested in and we can’t add them all

I understand this and I agree with you. However the URL works with every type of file, there's no restriction to js files. ✌️

bfred-it:
Yeah but nobody is going to use jsdeliver for rust files. What I mean is that it only applies to web developers. This button will appear on every file but it’d only make sense on some. Plus I suppose most use a bundler and not CDNs like that, so this further reduces its utility.

bfred-it:
This is the closest feature we’d be interested in, as far as CDNs go: https://github.com/sindresorhus/refined-github/issues/1096

PR #2272, opened 2019-07-26
Use updated `webext-options-sync` (testing needed)

fregante:
I'm looking for testers of this [new version of webext-options-sync](https://github.com/fregante/webext-options-sync/pull/28#issuecomment-515147948)

You can [download a pre-built extension](https://github.com/sindresorhus/refined-github/files/3436334/refined_github-0.0.0.zip) — or fetch, `npm install`, build, and load this PR as usual.

This mainly affects:

- how the options form is saved and restored
- how the options are stored (default values are now excluded, less data is stored to disk)

Things to test:

- new installation: does the extension work correctly without opening the Options page?
- updated of existing installation: does the extension work correctly without opening the Options page? What about after opening it?
- does the Option page show the previous values?
- does the Option page save/restore the changes in all the fields? (excluding an invalid token)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4442,6 +4442,14 @@
 				}
 			}
 		},
+		"dom-form-serializer": {
+			"version": "1.0.7",
+			"resolved": "https://registry.npmjs.org/dom-form-serializer/-/dom-form-serializer-1.0.7.tgz",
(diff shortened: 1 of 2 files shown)


fregante:
More changes will be coming to this module, I should probably add them and _then_ ask for testing of everything.

PR #2294, opened 2019-08-01
Fix `sticky-discussion-sidebar`

sharkykh:
Feature did not work because `.discussion-sidebar`'s height was almost the whole page, and it never got "tagged" with `.rgh-sticky-sidebar`.
Also recorded [a corrected screenshot](https://user-images.githubusercontent.com/10238474/62276723-5a2eaa80-b44d-11e9-810b-ff598d1c5c6a.gif), because [the old one](https://user-images.githubusercontent.com/380914/39878141-7632e61a-542c-11e8-9c66-74fcd3a134aa.gif) was for the "Sticky Issues Header" feature.

### Test
Visit any issue or PR with a sidebar that fits within your screen height and scroll up and down.


readme.md
```diff
@@ -186,7 +186,7 @@ Thanks for contributing! 🦋🙌
 
 - [](# "batch-open-issues") 🔥 [Adds a button to open multiple discussions at once in your repos.](https://user-images.githubusercontent.com/1402241/38084752-4820b0d8-3378-11e8-868c-a1582b16f915.gif)
 - [](# "split-issue-pr-search-results") [Separates issues from PRs in the global search.](https://user-images.githubusercontent.com/1402241/52181103-35a09f80-2829-11e9-9c6f-57f2e08fc5b2.png)
-- [](# "make-discussion-sidebar-sticky") [Makes the discussion filters bar sticky.](https://user-images.githubusercontent.com/380914/39878141-7632e61a-542c-11e8-9c66-74fcd3a134aa.gif)
+- [](# "make-discussion-sidebar-sticky") [Makes the discussion sidebar sticky.](https://user-images.githubusercontent.com/10238474/62276723-5a2eaa80-b44d-11e9-810b-ff598d1c5c6a.gif)
```
fregante: I'd keep the two features nearby and I'd rename them to `sticky-discussion-sidebar` and `sticky-discussion-list-toolbar`

```md
- [](# "sticky-discussion-sidebar") [Makes the discussion sidebar sticky.](https://user-images.githubusercontent.com/10238474/62276723-5a2eaa80-b44d-11e9-810b-ff598d1c5c6a.gif)
- [](# "sticky-discussion-list-toolbar") [Makes the discussion list’s filters toolbar sticky.](https://user-images.githubusercontent.com/380914/39878141-7632e61a-542c-11e8-9c66-74fcd3a134aa.gif)
```

sharkykh: I will rename the features, but I address the sticky toolbar in #2295,
so I will rename and move it there after this PR is merged (to prevent a merge conflict)

sharkykh:
There's a small visual bug on PRs.
Notice the horizontal line at the bottom of the sidebar.
nS24XMIXGj

fregante:
This also needs a `featureWasRenamed` in `options-storage`

fregante:
> There's a small visual bug on PRs.

What if you remove the element? It doesn't seem necessary. Just ensure it's empty (`childNodes.length === 0` or something)

The alternative fix to this selector change would be to solve the issue at the root:
```css
.rgh-sticky-sidebar {
    align-self: flex-start;
}
```

But the above bar still stretches the element further than necessary, making the height ~30px longer for no reason.

sharkykh:
> What if you remove the element?

We could, although personally I prefer to not remove or hide elements.

> The alternative fix to this selector change would be to solve the issue at the root

I couldn't get that to work.

I have another solution, but it feels a little bit hacky...
Tested with Chrome and Firefox and seems to work fine.
I pushed it as a commit, because it's easier than explaining.

There's one more thing I noticed (unrelated to this bug):
When the page is loaded the calculated sidebar height is the unclean one (before `clean-sidebar` executes), which is usually rather big for no reason.
A possible solution for this is to observe `#partial-discussion-sidebar` and run the `updateStickiness` again after `clean-sidebar` is executed (assuming it's enabled).
We can observe the element classes for the when `.rgh-clean-sidebar` is added, maybe?

sharkykh:
We could also set `height: max-content` for `.discussion-sidebar` (the container element)  regardless of whether this feature is enabled or not, which will fix the height issue and allow us to use one less selector

<details>
<summary>Code</summary>

```js
function updateStickiness(): void {
const sidebar = select<HTMLElement>('.discussion-sidebar')!;
const sidebarHeight = sidebar.offsetHeight + 25 + 60; // 60 matches sticky header's height
sidebar.classList.toggle('rgh-sticky-sidebar', sidebarHeight < window.innerHeight);
}
```

```css
/*
Set the sidebar container's height to make it fit its contents
(instead of being the full height of the discussion)
*/
.discussion-sidebar {
height: max-content;
}

.rgh-sticky-sidebar {
position: sticky;
top: 60px !important; /* Matches sticky header's height */
z-index: 26 !important;
}

/* Remove bottom padding from last divider (on pull requests) */
.rgh-sticky-sidebar > .py-3:last-child {
padding-bottom: inherit !important;
}
```

</details>

fregante:
Thanks for looking into this!

PR #2295, opened 2019-08-01
Restore `sticky-discussion-list-toolbar` feature

### LINKED ISSUE #2280: Restore "Sticky Issues Header" feature
It was added in https://github.com/sindresorhus/refined-github/pull/1311 but I can't find it anymore.

Anyone remember if it was removed for any reason? cc @tommoor @hkdobrev @yakov116 

Anyway this is a quick fix if anyone wants to pick it up.

sharkykh:
Closes #2280

Couldn't use the original `.issues-listing .table-list-header` CSS selector because GitHub changed their markup (see https://github.com/sindresorhus/refined-github/commit/dd8760bf324c665088790df8be37a647882a0d88).
Did not need to set `max-height` on the dropdown lists because GitHub set their height to 400px (see 4f43ef673281ce0dec28a88d7893043fcf1a6ca2).
Also added the option to disable this feature from the options page.

### Test
(There aren't enough PRs to scroll currently, but you get the point)


source/features/make-discussion-list-toolbar-sticky.css
```diff
@@ -0,0 +1,7 @@
+/* Sticky table header on issues list */
+.rgh-sticky-discussion-list-toolbar #js-issues-toolbar {
+	position: sticky;
+	top: 0;
+	/* Must be above .modal-backdrop (z-index 20) #1317 */
+	z-index: 25;
+}
```
fregante: I'd just keep the CSS instead. This feature is minor enough and does not have the logic that the sidebar one has

PR #2314, opened 2019-08-08
Fix `sticky-discussion-sidebar`

### LINKED ISSUE #2313: `sticky-discussion-sidebar` feature broken
The `.discussion-sidebar` class doesn't exist anymore, will need to use `#partial-discussion-sidebar` to target the sidebar element instead.

image

sharkykh:
Fixes #2313 

> Actually, it looks like they made our job easier with this clean-up.

Turns out that not really, but it's still doable.

### Test
Any discussion.

### DIFF
--- a/source/features/sticky-discussion-sidebar.css
+++ b/source/features/sticky-discussion-sidebar.css
@@ -2,7 +2,7 @@
 Set the sidebar container's height to make it fit its contents
 (instead of being the full height of the discussion)
 */
-.discussion-sidebar {
+#discussion_bucket > :last-child {
 	height: max-content;
 }
 
--- a/source/features/sticky-discussion-sidebar.css
+++ b/source/features/sticky-discussion-sidebar.css
@@ -1,6 +1,7 @@
 /*
 Set the sidebar container's height to make it fit its contents
 (instead of being the full height of the discussion)
+`#discussion_bucket` holds two columns: the conversation container (1st) and the sidebar container (2nd)
 */
 #discussion_bucket > :last-child {
 	height: max-content;
--- a/source/features/sticky-discussion-sidebar.css
+++ b/source/features/sticky-discussion-sidebar.css
@@ -1,19 +1,10 @@
-/*
-Set the sidebar container's height to make it fit its contents
-(instead of being the full height of the discussion)
-`#discussion_bucket` holds two columns: the conversation container (1st) and the sidebar container (2nd)
-*/
-#discussion_bucket > :last-child {
-	height: max-content;
-}
-
 .rgh-sticky-sidebar {
 	position: sticky;
 	top: 60px !important; /* Matches sticky header's height */
 	z-index: 26 !important;
 }
 
-/* Remove padding from last divider (on pull requests) */
-.rgh-sticky-sidebar > .py-3:last-child {
-	padding: 0 !important;
+/* Hides the last divider (on pull requests) to avoid https://user-images.githubusercontent.com/10238474/62282128-af6fb980-b457-11e9-8b18-c29687b88da1.gif */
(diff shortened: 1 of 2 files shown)


source/features/sticky-discussion-sidebar.css
```diff
@@ -2,7 +2,7 @@
 Set the sidebar container's height to make it fit its contents
 (instead of being the full height of the discussion)
 */
-.discussion-sidebar {
+#discussion_bucket > :last-child {
```
sharkykh: This points at the same element as the JavaScript file does. Since there is no way to get the parent element in CSS, I used this.

notlmn: I think this was the layout change that made them remove the class that existed before.

This works for now, but we need to mention what the selector targets, preferably in the comments above the selector.

sharkykh: It does say it targets the "sidebar container" above it, if it's not good enough I'll add a sentence similar to my comment above.

notlmn: That comment does make sense, but I was referring to maybe adding a comment on why that specific CSS selector was used. Doesn't make much of a difference though, but I thought it would a little bit more clear.

sharkykh: Eh, it doesn't hurt to explain a little more. Thanks you :)

source/features/sticky-discussion-sidebar.css
```diff
@@ -1,8 +1,9 @@
 /*
 Set the sidebar container's height to make it fit its contents
 (instead of being the full height of the discussion)
+`#discussion_bucket` holds two columns: the conversation container (1st) and the sidebar container (2nd)
 */
-.discussion-sidebar {
+#discussion_bucket > :last-child {
```
fregante: Why this change? `#partial-discussion-sidebar` should be enough

source/features/sticky-discussion-sidebar.tsx
```diff
@@ -4,7 +4,7 @@ import debounce from 'debounce-fn';
 import features from '../libs/features';
 
 function updateStickiness(): void {
-	const sidebar = select<HTMLElement>('.discussion-sidebar')!;
+	const sidebar = select<HTMLElement>('#partial-discussion-sidebar')!.parentElement!;
```
fregante: Same here

sharkykh: Again, that's what I thought at first too, unfortunately no,
as it leads to this issue again:
https://github.com/sindresorhus/refined-github/pull/2294#issuecomment-517205362

nS24XMIXGj

source/features/sticky-discussion-sidebar.css
sharkykh: That's what I thought at first, but it seems GitHub didn't really change anything other than removing the `.discussion-sidebar` element that last week used to contain the `#partial-discussion-sidebar`.

`#partial-discussion-siderbar`` is at the correct height,
it's the container element (the column element) that needs its height adjusted.

sharkykh: image

fregante: Right, I see the issue. I think we should just drop it, it doesn't improve the design much and it looks like an afterthought for them too, since it's out of place.

The whole CSS would be:

```css
.rgh-sticky-sidebar {
	position: sticky;
	top: 60px !important; /* Matches sticky header's height */
	z-index: 26 !important;
}

/* Hides the last divider (on pull requests) */
.rgh-sticky-sidebar + .border-top:empty {
	display: none;
}
```

And the code below would be
```js
	const sidebar = select<HTMLElement>('#partial-discussion-sidebar')!;
```

PR #2315, opened 2019-08-08
Fix branch button alignment

notlmn:
The markup for navigation bar on repo tree is different from that of a single file page. This causes the insertion of a `div` before breadcrumbs whose parent is not a flex container to push breadcrumbs element to the bottom, even when there is enough space, Leading to a slightly broken view on repo tree pages.

## Before view for file and repo tree
branch-buttons-before

## After view for file and repo tree
branch-buttons-after

### DIFF
--- a/source/features/branch-buttons.css
+++ b/source/features/branch-buttons.css
@@ -1,0 +1,11 @@
+.rgh-branch-buttons {
+	margin-left: 0.5rem;
+}
+
+.file-navigation .rgh-branch-buttons {
+	margin: 0 3px !important;
+}
+
+.rgh-branch-buttons .BtnGroup {
+	margin: 0 !important;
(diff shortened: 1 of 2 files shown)


source/features/branch-buttons.css
```diff
@@ -0,0 +1,11 @@
+.rgh-branch-buttons {
+	margin-left: 0.5rem;
```
fregante: We should use the same units in both. Let’s go with pixels since the rest of the site does

notlmn: https://github.com/sindresorhus/refined-github/pull/2315/commits/b7496a6ed4673fddf31a07ad67a41c12958b804f should fix most of the styling inconsistencies.

PR #2320, opened 2019-08-11
Add `hidden-review-comments-indicator` feature

### LINKED ISSUE #1999: Comment indicators when comments are hidden on PR review
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/refined-github/issues/1999)

When I'm reading through changes on a PR I generally like to do so with the comments hidden. Comments can take up large parts of the screen. 

image

I find myself toggling comments on and off though. When comments are hidden (especially on a long PR) I don't know where comments are that I might care about. So there's this all or nothing approach... Enable comments and try to read the changes around them or disable comments and miss context. 

I'd like to be able to hide comments, but to have positional placeholders where comments appear. It'd also be useful to be able to expand/collapse comments on those placeholders. 

I'm not great at design, but it could be something like the below: 

Frame

Clicking on the convo icon would expand the comments.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $45.00)

### Submitted pull Requests
- [#2320 Add `review-comments-hidden-indicator` feature](https://issuehunt.io/r/sindresorhus/refined-github/pull/2320)
---

### Tips

---
</details>

sidvishnoi:
- Adds an indicator besides line number gutter to show number of review comments at that line, if any.
- Clicking on that indicator toggles hide/show all review comments in PR.

| comments hidden | comments visible |
| --- | --- |
| image | image |

Closes #1999

# Test
https://github.com/sidvishnoi/refined-github/pull/1/files

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#1999: Comment indicators when comments are hidden on PR review](https://issuehunt.io/repos/51769689/issues/1999)
---

</details>


source/features/review-comments-hidden-indicator.tsx
```diff
@@ -0,0 +1,131 @@
...
+		}
+
+		break;
+	}
+};
+
+const commentToggle = (count: number): JSX.Element => (
+	<tr className="refined-toggle-comments">
+		<td className="blob-num" colSpan={2}>
+			<button onClick={ev => toggleComments(ev.target as HTMLElement)}>
```
fregante: Create a generic handler function instead of creating one inline here.

```suggestion
			<button type="button" onClick={showCommentsHandler}>
```

Basically just renamed `toggleComments` above to `showCommentsHandler` and just get the `target` in there

source/features/review-comments-hidden-indicator.tsx
```diff
@@ -0,0 +1,131 @@
...
+const getOffsetY = (el: Element): number =>
+	el.getBoundingClientRect().top + window.pageYOffset;
+
+// Toggle comments while maintaining scroll position
+const toggleComments = (target: HTMLElement): void => {
+	const container = target.closest(SELECTOR_CUSTOM_TOGGLE)!;
+	const scrollOld = window.scrollY;
+	const offsetOld = getOffsetY(container);
+
+	select.all<HTMLInputElement>('input.js-toggle-file-notes').forEach(input => {
```
fregante: `toggleComments` looks for _all_ the comments. This shouldn't happen. If you click one of the indicators, only the comments on that file should be restored.

Also there's a utility function called `anchorScroll` that you can use to avoid the scroll. Example usage:

```js
const showCommentsHandler = (event): void => {
	anchorScroll(() => {
		(event.target as HTMLElement)
			.closest(SELECTOR_CUSTOM_TOGGLE)!
			.nextElementSibling!
			.querySelector('.js-toggle-file-notes')!
			.click();
	});
};
```

⬆️ This is probably what the whole function will look like.

source/features/review-comments-hidden-indicator.tsx
```diff
@@ -0,0 +1,131 @@
+import './review-comments-hidden-indicator.css';
+import React from 'dom-chef';
+import select from 'select-dom';
+import features from '../libs/features';
+import observeEl from '../libs/simplified-element-observer';
+import {comment as commentIcon} from '../libs/icons';
```
fregante:
```suggestion
import * as icons from '../libs/icons';
```

The use it as 

```js
icons.comment()
```

fregante:
Thank you for your PR, @sidvishnoi :)
I hope you can make these changes, the resulting code will be much shorter and more efficient

source/features/review-comments-hidden-indicator.tsx
```diff
@@ -0,0 +1,131 @@
...
+			if (comments.length === 0) {
+				removeToggles();
+			} else {
+				addToggle(container);
+			}
+		}
+	};
+
+	select.all('.diff-table tbody').forEach(tbody => {
+		observeEl(tbody, onTableMutation, {childList: true, subtree: true});
```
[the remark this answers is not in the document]
sidvishnoi: Thanks for the detailed feedback! 🙂

> The count might get out of date but it's not a big deal. That can be patched later, if necessary, by listening to specific "comment added" events.

Alright

source/features/review-comments-hidden-indicator.tsx
sidvishnoi: TypeScript was shouting over MouseEvent, will try again :)

source/features/review-comments-hidden-indicator.tsx
sidvishnoi: > toggleComments looks for all the comments. This shouldn't happen. If you click one of the indicators, only the comments on that file should be restored.

toggleComments actually looks for all files (not comments) using `input.js-toggle-file-notes` (default GitHub checkbox to toggle comments) and toggles using it only.

> If you click one of the indicators, only the comments on that file should be restored.

Sure we want that behaviour? I would've preferred hiding all comments. Fine with what you say.

> Also there's a utility function called `anchorScroll`

Thanks. Will look into it.

source/features/review-comments-hidden-indicator.tsx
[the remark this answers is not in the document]
sidvishnoi: > 1. Use observeEl to observe the class attribute:
> ```
> observeEl('.has-inline-notes', onTableMutation, {
> ```

How would we know if a new comment container is added? 
Also, the above will just observe first diff with `.has-inline-notes`. I think we would need to observe attribute change (`'.has-inline-notes'`) with `childNodes` on all diff/files.


> When `.show-inline-notes` is added, generate the toggles unless they're already there.

`.show-inline-notes` is added by default. So all toggles will be generated on first load anyway.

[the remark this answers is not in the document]
fregante: > How would we know if a new comment container is added?

Like I said: *No new threads are created when the comments are hidden and... the indicators are added the next time the comments are hidden again.*

This means that when the comments are added, we don’t need to take any action. Only when the user hides the comment, the new comment threads (if any) will be used to create new indicators. 

> .show-inline-notes is added by default. So all toggles will be generated on first load anyway.

My bad, I meant that when that class is **removed**, i.e. when the comments are hidden

source/features/review-comments-hidden-indicator.tsx
fregante: Search in other features, there’s some weird type called React.MouseEvent

source/features/review-comments-hidden-indicator.tsx
fregante: Yes, that’s what I meant, all the comments on the page will be restored and that’s not necessary. Only the comments on the current file should (because that’s how the default “Show comments” works

source/features/review-comments-hidden-indicator.tsx
sidvishnoi: Thanks. It works. Just fails in my VSCode 🎉

source/features/review-comments-hidden-indicator.tsx
```diff
@@ -3,33 +3,22 @@ import React from 'dom-chef';
...
-	const offsetNew = getOffsetY(container);
-	const scrollY = offsetNew - offsetOld + scrollOld;
-	window.scrollTo(0, scrollY);
+// TODO: remove `any`, use Event
+const toggleComments = (event: any): void => {
+	const target = event.target as Element;
+	anchorScroll(() => {
+		const file = target.closest('.file.js-file')!;
+		const inputToggle = select('input.js-toggle-file-notes', file)!;
+		inputToggle.click();
```
fregante: My suggested code with a chain is easier to follow, I don’t like to mix `closest` with `select` because one goes from left to right and the other from right to left. It’s hard to follow

In the next line you can use `}, event.target);` instead of creating the other constant

sidvishnoi:
Thanks for reviewing. Will continue with rest tomorrow. It's late night here.

sidvishnoi:
@fregante Notable changes since yesterday:
1. Using `observeEl` to observe each file for removal of `.show-inline-notes`, to add new indicators on that file, if needed. 
2. Removed `countListener`. Counts will get outdated if comments are added or removed. This also means, if comment count reaches 0 at some indicator, that indicator won't be removed until refresh.
3. The feature will now also run on `features.isPRCommit`

source/features/review-comments-hidden-indicator.css
```diff
@@ -0,0 +1,41 @@
...
+	align-items: center;
+	line-height: 1;
+}
+
+tr.rgh-comments-indicator button:hover {
+	color: #586069;
+}
+
+.show-inline-notes .rgh-comments-indicator {
+	display: none;
```
sidvishnoi: This will remove the ability to toggle comment through indicators. I don't think we should do that. The original feature request also suggested so:
> It'd also be useful to be able to expand/collapse comments on those placeholders.

fregante: I missed that part, I was working on the assumption that they only appear when the comments are hidden.

The idea isn't bad, but I think it makes the UI more busy.

sidvishnoi: Perhaps making the underline a bit lighter will improve the UI.

sidvishnoi:
> OctoLinker seems to add a overflow-x property which causes the icon to be clipped off:

It's not OctoLinker that's adding the `overflow-x`. And it's working fine with that overflow actually. The image shows `overflow-x` being added without Octolinker.
image

But OctoLinker is definitely clashing somewhere else. Will update if I find out something.

fregante:
> It's not OctoLinker that's adding the `overflow-x`.

Indeed. I found the culprit:

https://github.com/OctoLinker/OctoLinker/blob/0174f49ed52e936dadcacc81f8cdf428d86bc950/packages/helper-insert-link/style.css#L45-L47

`overflow-x` is not a problem until all children become relative to that exact element.

sidvishnoi:
The `.highlight` in Octolinker seems more purposeful than the `overflow-x` here. Removing either doesn't cause an visible issue on PR files pages (at least I didn't find any)
What do you recommend?

source/features/review-comments-hidden-indicator.css
sidvishnoi: Revert cd5abfc8ba1fb74e446b7ebb33826c39ec099fd4?

fregante:
>  What do you recommend?

I think removing either could cause issues and conflicts. The best solution is to ask about it on OctoLinker and suggest its removal there. Perhaps the selector can be improved to avoid files.

source/features/review-comments-hidden-indicator.css
sidvishnoi: @fregante 	We may want to resolve this first?

source/features/hidden-review-comments-indicator.tsx
```diff
@@ -0,0 +1,82 @@
...
+function init(): void {
+	observeFiles();
+	onPrFileLoad(observeFiles);
+}
+
+features.add({
+	id: __featureName__,
+	description: 'Adds comment indicators when comments are hidden in PR review',
+	screenshot:
+		'https://user-images.githubusercontent.com/1402241/35480123-68b9af1a-043a-11e8-8934-3ead3cff8328.gif', // TODO
```
sidvishnoi: Also, need an actual screenshot first. 
Can you create a GIF? I would have to install a screen recorder and video converter on Windows 😅

sidvishnoi:
@fregante 	Made indicators toggle-able, not just hide-able. Feel free to revert if you do not want to support that feature.
Also, added a gif: https://user-images.githubusercontent.com/8426945/63024694-59722b80-bec5-11e9-812e-8c1df69a4243.gif

sindresorhus:
What's the purpose of hiding a visible comment? That's what the "Resolve" button is for, right?

sidvishnoi:
@sindresorhus 	While reviewing code, one may want to reduce the distractions in order to have a better picture of diff - so we hide comments, as requested in #1999. Resolve button is to tell comment has been resolved. So, entirely different purpose.

sindresorhus:
But there's already a button to hide *all* the comments. I don't really see the point in hiding a single comment.

sidvishnoi:
> But there's already a button to hide all the comments.

Yes. It extends that functionality. The default one hides all comments, and one can't know if there are any comments unless they click that button again. This feature adds an indicator when comments are hidden.

> I don't really see the point in hiding a single comment.

This feature hides all the comments, not a single comment.

fregante:
> This feature hides all the comments, not a single comment.

It hides all the comments in a single file, which is the same feature already offered by GitHub. We can do without this

fregante:
Thanks for this feature @sidvishnoi! ^____^

sidvishnoi:
Oh, I totally missed split-diff view! 

Thanks a lot @fregante for the insightful reviews and edits! 🎉
I learned a lot from this PR 🙂

zephraph:
Thanks for this! I was doing a code review and it started working. Very excited 😄

PR #2323, opened 2019-08-12
Make footer text readable when user hovers over it

### LINKED ISSUE #2212: The footer is incredibly hard to read in Explore page
When navigating to https://github.com/explore, the refined github plugin causes the footer to be extra light and rendering any text unreadable. I'm kinda OK with this in the github repository view, but on the rest of the site, I think the footer should be more visible.

With Refined Github:
image

Without Refined Github:
image

kumarharsh:
Closes #2212

# Test
* https://github.com/about

### DIFF
--- a/source/features/hide-readme-header.css
+++ b/source/features/hide-readme-header.css
@@ -22,7 +22,7 @@
 	right: 0;
 	top: 0;
 	opacity: 0.2;
-	transition: opacity 250ms;
+	transition: opacity 0.25s;
 }
 
 .repository-content #readme:not(.blob) .Box-header .btn-octicon:hover,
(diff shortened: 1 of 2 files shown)


source/features/hide-readme-header.css
```diff
@@ -22,7 +22,7 @@
 	right: 0;
 	top: 0;
 	opacity: 0.2;
-	transition: opacity 250ms;
+	transition: opacity 0.25s;
```
fregante: y tho 😅

fregante:
Thank you ☺️

source/features/hide-readme-header.css
kumarharsh: Felt the irresistible itch to make the durations uniform 😅

PR #2344, opened 2019-08-15
Fix `clean-sidebar` on sidebar updates

### LINKED ISSUE #2333: `clean-discussion-sidebar` isn't re-applied after changes to the sidebar
1. Visit an issue
2. Post a comment or add a label
3. Notice that the sidebar is _dirty_ again

asd

There should either be an even listener on the ajaxed element (it's probably called `socket:message`) or a `MutationObserver`

sharkykh:
Using the new `onUpdatableContentUpdate` by @fregante.

Closes #2333.

### Test
On a discussion where you have access to assign things on the sidebar (label, milestone, etc.), make a change - after the "partial" content is updated, the sidebar is cleaned.

### DIFF
--- a/source/features/clean-sidebar.tsx
+++ b/source/features/clean-sidebar.tsx
@@ -2,7 +2,7 @@ import './clean-sidebar.css';
 import React from 'dom-chef';
 import select from 'select-dom';
 import features from '../libs/features';
-import observeEl from '../libs/simplified-element-observer';
+import onUpdatableContentUpdate from '../libs/on-updatable-content-update';
 import {isPR} from '../libs/page-detect';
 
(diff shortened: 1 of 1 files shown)


fregante:
Excellent! Now I wonder if `onUpdatableContentUpdate` should accept a selector instead. Perhaps `HTMLElement | string` like many other of our helper functions do (e.g. `observeEl`)

Also will you look into `parse-backticks` in the header or should I send a PR?

sharkykh:
Not today.
I think you mean https://github.com/sindresorhus/refined-github/blob/master/source/features/format-discussion-titles.tsx?

PR #2346, opened 2019-08-17
Make `copy-file` feature work on gists

kidonng:
I don't know if this works before, but from `features.isGist` I guess it is expected to, so here it is.

# Test

https://gist.github.com/KomaliVuddanti/f062a78b4fff3ee001529efe2a468375

### DIFF
--- a/source/features/copy-file.tsx
+++ b/source/features/copy-file.tsx
@@ -5,7 +5,7 @@ import copyToClipboard from 'copy-text-to-clipboard';
 import features from '../libs/features';
 
 function handleClick({currentTarget: button}: React.MouseEvent<HTMLButtonElement>): void {
-	const selector = features.isGist() ? '.js-gist-file-update-container': '.Box';
+	const selector = features.isGist() ? '.js-gist-file-update-container' : '.Box';
 	const file = button.closest(selector);
 	const content = select.all('.blob-code-inner', file!)
 		.map(({innerText: line}) => line === '\n' ? '' : line) // Must be `.innerText`
@@ -17,7 +17,7 @@ function renderButton(): void {
 	const selector = features.isGist() ? '.file-actions .btn' : '[data-hotkey="b"]';
 	for (const blameButton of select.all(selector)) {
 		if (features.isGist()) {
-			blameButton.classList.add('BtnGroup-item')
+			blameButton.classList.add('BtnGroup-item');
 		}
 
 		blameButton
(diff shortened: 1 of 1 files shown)


source/features/copy-file.tsx
```diff
@@ -5,16 +5,22 @@ import copyToClipboard from 'copy-text-to-clipboard';
 import features from '../libs/features';
 
 function handleClick({currentTarget: button}: React.MouseEvent<HTMLButtonElement>): void {
-	const file = button.closest('.Box');
+	const selector = features.isGist() ? '.js-gist-file-update-container' : '.Box';
+	const file = button.closest(selector);
```
fregante: No need to is isGist, just merge the selectors

source/features/copy-file.tsx
```diff
@@ -5,16 +5,22 @@ import copyToClipboard from 'copy-text-to-clipboard';
...
 	copyToClipboard(content);
 }
 
 function renderButton(): void {
-	for (const blameButton of select.all('[data-hotkey="b"]')) {
-		blameButton
+	const selector = features.isGist() ? '.file-actions .btn' : '[data-hotkey="b"]';
+	for (const button of select.all(selector)) {
+		if (features.isGist()) {
+			button.classList.add('BtnGroup-item');
```
fregante: You can use the groupButtons function to do this automatically, without conditions. Call it after the prepend

kidonng: Have I done it right?

fregante: You can leave the old code (without the extra const) and use `groupSiblings(button)` instead, I think.

kidonng: I thought you were literally talking `groupButtons` function.

source/features/copy-file.tsx
```diff
@@ -3,18 +3,20 @@ import select from 'select-dom';
...
 	const content = select.all('.blob-code-inner', file!)
 		.map(({innerText: line}) => line === '\n' ? '' : line) // Must be `.innerText`
 		.join('\n');
 	copyToClipboard(content);
 }
 
 function renderButton(): void {
-	for (const blameButton of select.all('[data-hotkey="b"]')) {
-		blameButton
+	const selector = features.isGist() ? '.file-actions .btn' : '[data-hotkey="b"]';
```
jerone: Can these selectors also be merged?

source/features/copy-file.tsx
fregante: Sorry about the confusion 😁

source/features/copy-file.tsx
fregante: This isn’t resolved yet. By “merging” he means what you did with the previous selector where I suggested to drop isGist

kidonng: Sorry I always forget its parameter can be multiple selectors

PR #2347, opened 2019-08-17
Add `embed-gist-via-iframe` feature

kidonng:
Before:

After:
image

GitHub has a listener for the menu items so we just need to copy an existing item to make it work.

The description change is to distinguish the two embed items and is also consistent with the existing "Clone via HTTPS" and "Clone via SSH" items.

# Test

https://gist.github.com/KomaliVuddanti/f062a78b4fff3ee001529efe2a468375


sindresorhus:
> My code isn't too elegant so I'll be all ears

If you have to write this, you should probably put some work into it ;)

source/features/embed-gist-via-iframe.tsx
```diff
@@ -0,0 +1,37 @@
...
+import features from '../libs/features';
+
+function init(): void {
+	const embedMenuItem = select('.select-menu-item')!;
+	select('.select-menu-button')!.textContent = 'Embed via <script>';
+	select('.select-menu-item-heading', embedMenuItem)!.innerHTML = 'Embed via <code>&lt;script&gt;</code>';
+	select('.description', embedMenuItem)!.innerHTML = 'Embed this gist in your website via <code>&lt;script&gt;</code>.';
+
+	const iframe = `<iframe src="https://gist.github.com${location.pathname}.pibb"></iframe>`;
+	embedMenuItem.insertAdjacentElement(
```
jerone: This can be replaced by `.after()`. See https://github.com/sindresorhus/refined-github/pull/760

kidonng: Oh thanks I'm rebuilding this via `cloneNode`

source/features/embed-gist-via-iframe.tsx
```diff
@@ -0,0 +1,37 @@
...
+		'afterend',
+		<button name="button" type="button" className="select-menu-item width-full" aria-checked="false" role="menuitemradio" value={iframe}>
+			<svg className="octicon octicon-check select-menu-item-icon" viewBox="0 0 12 16" version="1.1" width="12" height="16" aria-hidden="true"><path fill-rule="evenodd" d="M12 5l-8 8-4-4 1.5-1.5L4 10l6.5-6.5L12 5z" /></svg>
```
jerone: Use the icon: https://github.com/sindresorhus/refined-github/blob/master/source/libs/icons.tsx#L3

kidonng:
@jerone How does it look now :)

kidonng:
> If you have to write this, you should probably put some work into it ;)

Oh sorry I should have taken it seriously 😋

sindresorhus:
Also don't forget to send the feature request to GitHub: support@github.com It feels like something that should be available natively.

source/features/embed-gist-via-iframe.tsx
```diff
@@ -0,0 +1,37 @@
+import select from 'select-dom';
+import features from '../libs/features';
+
+function init(): void {
+	const embedViaScript = select('.select-menu-item')!;
```
fregante: The selector is too broad

source/features/embed-gist-via-iframe.tsx
```diff
@@ -0,0 +1,37 @@
...
+	const embedViaIframe = embedViaScript.cloneNode(true) as HTMLElement;
+
+	// Modify description to distinguish the two items
+	// This is also consistent with the existing "Clone via HTTPS" and "Clone via SSH" items
+	const viaScript = 'Embed via <script>';
+	select('[data-menu-button]')!.textContent = viaScript;
```
fregante: Broad selector or this needs to be 
```suggestion
	select('[data-menu-button]', embedViaScript)!.textContent = viaScript;
```

source/features/embed-gist-via-iframe.tsx
```diff
@@ -0,0 +1,37 @@
...
+	select('.select-menu-item-heading', embedViaScript)!.textContent = viaScript;
+	select('.description', embedViaScript)!.textContent = 'Embed this gist in your website via <script>.';
+
+	// Remove analytics attributes
+	embedViaIframe.removeAttribute('data-hydro-click');
+	embedViaIframe.removeAttribute('data-hydro-click-hmac');
+
+	// Set required content
+	embedViaIframe.setAttribute('aria-checked', 'false');
+	embedViaIframe.setAttribute('value', embedViaIframe.getAttribute('value')!.replace(/script/g, 'iframe').replace('.js', '.pibb'));
```
fregante: Properties are more readable than attributes here

```suggestion
	embedViaIframe.value = embedViaIframe.value!.replace(...)
```

Also this fails if the username or domain (for GH Enterprise) contains `script` or `.js` anywhere. It needs to match the HTML tag more precisely

kidonng: > Properties are more readable than attributes here

That will fail the type check as it is marked as `HTMLElement` 😅 well I should have changed it to `HTMLButtonElement`

> Also this fails if the username or domain (for GH Enterprise) contains `script` or `.js` anywhere. It needs to match the HTML tag more precisely

Thanks for reminding me, should adopt the previously used method: extract URL from `location`

source/features/embed-gist-via-iframe.tsx
kidonng: That's actually wrong but never mind

source/features/embed-gist-via-iframe.tsx
```diff
@@ -0,0 +1,37 @@
+import select from 'select-dom';
+import features from '../libs/features';
+
+function init(): void {
+	const embedViaScript = select('.file-navigation-option .select-menu-item')!;
```
fregante: Still too broad; how about:

```suggestion
	const embedViaScript = select('.file-navigation-option [value$="<script"]')!;
```

source/features/embed-gist-via-iframe.tsx
```diff
@@ -0,0 +1,37 @@
...
+	const embedViaIframe = embedViaScript.cloneNode(true) as HTMLButtonElement;
+
+	// Modify description to distinguish the two items
+	// This is also consistent with the existing "Clone via HTTPS" and "Clone via SSH" items
+	const viaScript = 'Embed via <script>';
+	select('.file-navigation-option [data-menu-button]')!.textContent = viaScript;
```
fregante: I saw this now, I don't think we need to update the label here. 

The regular label is fine, `<script` is visible right next to it:



It can be dropped

```suggestion
```

source/features/embed-gist-via-iframe.tsx
```diff
@@ -0,0 +1,37 @@
...
+	select('.select-menu-item-heading', embedViaScript)!.textContent = viaScript;
```
fregante:
```suggestion
	select('.select-menu-item-heading', embedViaScript)!.textContent = 'Embed via <script>';
```

source/features/embed-gist-via-iframe.tsx
kidonng: If we need to drop this, we should drop the line below as well (i.e. https://github.com/sindresorhus/refined-github/pull/2347/files#diff-202a2bef15a311f406c80dbc2c6e74bfR10-R12 together), because when you click another button and click the "Embed" button again, the label will sync with our modified title and become "Embed via <script>", which can be confusing.

source/features/embed-gist-via-iframe.tsx
kidonng: I guess you are writing `$('.file-navigation-option [value^="<script"]')` 😂

source/features/embed-gist-via-iframe.tsx
fregante: The next line fine as is. As a starter the it’s better as simply “embed”

fregante:
I'm seeing this really weird behavior in Firefox. The buttons are swapped and incorrect.



I'm about to push a change that sorts the code and fixes the `script` label, but the result is similar.

It works correctly in Chrome though.

PR #2310, opened 2019-08-07
Add `conflict-marker` feature

Esemesek:
Not sure if the placement or the style of the label is the best, but this is how I imagined it. I am open for suggestions C:

Closes #1593 

Screenshot:

Screenshot 2019-08-09 at 14 00 46

# Test

I've tested this feature on [react-native](https://github.com/facebook/react-native/pulls) repository.

PR #2351, opened 2019-08-18
Add `highlight-collaborators-in-lists` feature

### LINKED ISSUE #438: Highlight issues and PRs from org members
Context: for example, in the [React](https://github.com/facebook/react/pulls) team, much of the work happens via PRs, even by the core team. However, these core team PRs are lost in the long, long list of contributions.

Proposal: highlight PRs that are from members of the repo owning org, Facebook in the example of React.

jerone:
# Description
Highlights PRs opened by organization collaborators.

# Screenshot
screenshot

# Closes
Closes #438

# Test
1. Visit an issue/PR list like: https://github.com/sindresorhus/refined-github/issues
2. Look for collaborators of the current repo or organization


source/features/highlight-collaborators-prs.tsx
```diff
@@ -0,0 +1,46 @@
...
+		return false;
+	}
+
+	const authors = select.all('.js-issue-row [data-hovercard-type="user"]');
+	if (authors.length === 0) {
+		return false;
+	}
+
+	const data = await api.v4(buildQuery());
+	const collaborators = data.repository.collaborators.nodes.map((node: AnyObject) => node.login);
```
fregante: I'd rather see a `getCollaborators` function instead of a random `buildQuery`


Also the feature should work on the global PR list as well (`/pulls`) as suggested for this other feature https://github.com/sindresorhus/refined-github/pull/2310#discussion_r311644086

source/features/highlight-collaborators-prs.tsx
```diff
@@ -0,0 +1,46 @@
...
+		repository(${getRepoGQL()}) {
+			collaborators(first: 100) {
+				nodes { login }
+			}
+		}
+	`;
+}
+
+async function init(): Promise<false | void> {
+	if (!features.isRepoWithAccess()) {
```
fregante: There's an `exclude` array in `features.add`

fregante:
Also this should not be limited to PRs. The filter should be `isDiscussionList`

source/features/highlight-collaborators-prs.tsx
jerone: > There's an exclude array in features.add

:+1:

> But is this needed?

Yeah, this is needed because it only works on repos which you have settings rights to. On this repo I get the following error:
```
{
  "data": {
    "repository": {
      "collaborators": null
    }
  },
  "errors": [
    {
      "type": "FORBIDDEN",
      "path": [
        "repository",
        "collaborators"
      ],
      "locations": [
        {
          "line": 3,
          "column": 5
        }
      ],
      "message": "Must have push access to view repository collaborators."
    }
  ]
}
```

> Also it should probably be renamed to canUserCollaborate or something more specific.

That can be in another PR.

source/features/highlight-collaborators-prs.tsx
jerone: > I'd rather see a `getCollaborators` function instead of a random `buildQuery`

:+1:, was just copied from other feature. Will fix.



> Also the feature should work on the global PR list as well (`/pulls`) as suggested for this other feature

That will make it an magnitude harder. Running this on an repo discussion list binds it to just one repo (one fetch). The global discussion list can be any repo; possible extending the query by 10 (page size) repo fetches.

source/features/highlight-collaborators-prs.tsx
fregante: > Yeah, this is needed because it only works on repos which you have settings rights to. On this repo I get the following error:

Wow, what an ugly limitation. I think this limits its usefulness a bit. Do you think it's worth using this hack? https://stackoverflow.com/a/40987268/288906
Parsing the DOM here is not different from what we do across the extension (e.g. `forked-to`)

source/features/highlight-collaborators-prs.tsx
fregante: > That will make it an magnitude harder

It _is_ a bit harder, repos would have to be de-duplicated so they only appear once in each fetch. 

Perhaps this info is not as useful in those pages in the first place.

source/features/highlight-collaborators-prs.tsx
jerone: Interesting. The _hack_ is already available for almost 3 year, it might be _safe_ to use.

But I'm not sure if we should go that way. This [feature request](https://github.com/sindresorhus/refined-github/issues/438#issue-231104704) was only about viewing PRs from other collaborators (who should have push access to the repo). And as owner I'm most interested in this information.

Aldo this information might be interesting for other 'normal' users too. 🤔

I could add the url hack as a fallback in case the API v4 throws...

jerone: I also noticed that the url hack included my own name, which is false in the case of the [example](https://github.com/Leaflet/Leaflet/issues/show_menu_content?partial=issues/filters/authors_content).

fregante: > I could add the url hack as a fallback in case the API v4 throws...

No, there should only be one thing. We can't maintain 2.

fregante: > I also noticed that the url hack included my own name, which is false in the case of the [example](https://github.com/Leaflet/Leaflet/issues/show_menu_content?partial=issues/filters/authors_content).

Indeed, but that's your own name, it doesn't need this highlight (and it already is)

jerone: So, you want me to change the implementation to the **url hack**, instead of the **limited API v4** ?

fregante: I think it's best to go with the URL hack so it becomes more useful.

jerone:
@fregante I implemented the URL hack solution and enabled this feature on all discussion lists.

source/content.ts
```diff
@@ -138,6 +138,7 @@ import './features/revert-file';
 import './features/hidden-review-comments-indicator';
 import './features/reload-failed-proxied-images';
 import './features/clean-rich-text-editor';
+import './features/highlight-collaborators-prs';
```
fregante:
```suggestion
import './features/highlight-collaborators-in-lists';
```

source/features/highlight-collaborators-prs.tsx
```diff
@@ -0,0 +1,34 @@
...
+		if (collaborators.includes(author.textContent!.trim())) {
+			author.classList.add('rgh-collaborator');
+		}
+	}
+}
+
+features.add({
+	id: __featureName__,
+	description: 'Highlights discussions opened by organization collaborators.',
+	screenshot: '',
```
fregante: Needs a "retina" screenshot as always

source/content.ts
```diff
@@ -138,6 +138,7 @@ import './features/revert-file';
 import './features/hidden-review-comments-indicator';
 import './features/reload-failed-proxied-images';
 import './features/clean-rich-text-editor';
```
fregante: test

fregante:
Thanks for tackling the oldest open issue on the repo 🙀

source/features/highlight-collaborators-prs.tsx
jerone: Will do

fregante:
🍍 🎂

PR #2353, opened 2019-08-19
fix: marketplace is opt-in for Enterprise, handle it being turned off

busches:
Marketplace is an opt-in feature for GHE, if you're on an instance where it's not enabled, `marketPlaceLink` will be undefined and return an error in the console when `.closest` is invoked.

### DIFF
--- a/source/features/deprioritize-marketplace-link.tsx
+++ b/source/features/deprioritize-marketplace-link.tsx
@@ -5,7 +5,7 @@ import domLoaded from 'dom-loaded';
 import features from '../libs/features';
 
 async function init(): Promise<void> {
-	const marketPlaceLink = (await elementReady('.Header-link[href="/marketplace"]'))!;
+	const marketPlaceLink = (await elementReady('.Header-link[href="/marketplace"]'));
 	if (marketPlaceLink) {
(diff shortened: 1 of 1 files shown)


source/features/deprioritize-marketplace-link.tsx
```diff
@@ -5,9 +5,11 @@ import domLoaded from 'dom-loaded';
 import features from '../libs/features';
 
 async function init(): Promise<void> {
-	(await elementReady('.Header-link[href="/marketplace"]'))!
+	const marketPlaceLink = (await elementReady('.Header-link[href="/marketplace"]'))!;
```
fregante: I think there’s an extra `()!` here. There’s an if that avoids the need for `!` already

fregante: After this, it can be merged if it works. I don’t have GHE

busches: It works on GHE without it and on GH with it. I don't have access to an instance of GHE that does have it opted in, but assuming the markup is like GH, it'll work. If the markup isn't like GH, well it never would have worked. :)

PR #2354, opened 2019-08-19
fix: only add mark unread on PR Conversation view

busches:
Adding the mark unread button on all pages of the PR, e.g. commits, results in a console error:

```
content.js:2568 TypeError: Cannot read property 'after' of null
    at addMarkUnreadButton (content.js:3769)
    at mark_unread_init (content.js:4011)
    at async features_run (content.js:2562)
```

### DIFF
--- a/source/features/mark-unread.tsx
+++ b/source/features/mark-unread.tsx
@@ -409,7 +409,7 @@ async function init(): Promise<void> {
 			delegate('.js-delete-notification button', 'click', updateUnreadIndicator),
 			delegate('.js-mark-visible-as-read', 'submit', markVisibleNotificationsRead)
 		);
-	} else if (pageDetect.isPR() || pageDetect.isIssue()) {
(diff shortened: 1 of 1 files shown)


source/features/mark-unread.tsx
```diff
@@ -409,7 +409,7 @@ async function init(): Promise<void> {
 			delegate('.js-delete-notification button', 'click', updateUnreadIndicator),
 			delegate('.js-mark-visible-as-read', 'submit', markVisibleNotificationsRead)
 		);
-	} else if (pageDetect.isPR() || pageDetect.isIssue()) {
+	} else if (pageDetect.isPRConversation() || pageDetect.isIssue()) {
```
fregante: `markRead` needs to be called on all pages of PRs, not just the conversation view. Only the button needs to have this condition change

busches: Thanks, I don't pay attention to notifications at all, but this error pops up all the time. :)

PR #2359, opened 2019-08-21
Fix always expanded "More" dropdown menu

kidonng:
Closes #2357
Closes #2358

For some reason using `<button>` won't open the dropdown, I have read GitHub's code for some time but can't figure out why, so changed it to `<span>` and everything looks fine.

# Test

This repo

Issue #2357, opened 2019-08-21
Bug: "More" dropdown always expanded


corysimmons:
Refined GitHub enabled on Chrome:

![image](https://user-images.githubusercontent.com/224636/63393581-b2403900-c389-11e9-9bd1-c88c2dd48360.png)

Disabled:

![image](https://user-images.githubusercontent.com/224636/63393594-bff5be80-c389-11e9-8d5a-a7d6a1b95692.png)

It just started today, and I've only noticed it does this when viewing a Repo (not other views afaik).

Clicking the dropdown, or anything, doesn't hide the dropdown. Seems like it's just stuck in this state.

joscha:
> It just started today, and I've only noticed it does this when viewing a Repo (not other views afaik).

Also happens in pull requests and is overlaying the file tab and branch name, which is quite unfortunate:

<img width="352" alt="fix__some_dependencies_by_joscha_·_Pull_Request__64857_·_Canva_canva" src="https://user-images.githubusercontent.com/188038/63393924-a8a5e800-c400-11e9-91b1-f3267f830995.png">

Nightfirecat:
Not a fix for the issue, but disabling `more-dropdown` removes this while it is being fixed.

rohitkrishna094:
For anyone having trouble, you can find `more-dropdown` in the extension's settings. For chrome, just right click on the extension(refined github in this case) and click on `options`. Once there, just scroll down and you will find `more-dropdown` as Nightfirecat mentioned.

busches:
Looks like we're on the legacy markup and they removed support for it today. We'll need to convert it to the `<details><summary>` markup you can find in the other drop downs. Most likely GHE will still need the current markup though.

casz:
As a temporary solution, I created #2358 till we get a proper fix

coreysnyder:
Seeing the same thing just now.

kidonng:
@busches Nice catch! Created a fix PR.

alerque:
Just a confirmation ... the original report states this happens on Chrome. I'm on Firefox and seeing the same thing today. The menu is stuck open whether on old or new tabs and cannot be closed.

cocoonkid:
Same on Brave

rajatkantinandi:
Same issue here
[...]

PR #2363, opened 2019-08-21
Fix `releases-tab` when `more-dropdown` is disabled

### LINKED ISSUE #2361: `releases-tab` does not work without `more-dropdown`
When `more-dropdown` is disabled or have failed to initialize due to an error,
`releases-tab` fails to inject.
It's not really a bug, but rather an enhancement request - to make `releases-tab` isolated.

`more-dropdown` enabled + `releases-tab` enabled:
image

`more-dropdown` disabled + `releases-tab` enabled:
image

Error in console:
```
❌ releases-tab
content.js:2709 TypeError: Cannot read property 'before' of null
    at releases_tab_init (content.js:3509)
    at async features_run (content.js:2704)
```

kidonng:
Closes #2361 

# Test

- [x] Repo without admin access
- [x] Repo with admin access

### DIFF
--- a/source/features/releases-tab.tsx
+++ b/source/features/releases-tab.tsx
@@ -6,6 +6,7 @@ import features from '../libs/features';
 import * as icons from '../libs/icons';
 import {getRepoURL} from '../libs/utils';
 import {isRepoRoot, isReleasesOrTags} from '../libs/page-detect';
+import {appendBefore} from '../libs/dom-utils';
 
 const repoUrl = getRepoURL();
 const repoKey = `releases-count:${repoUrl}`;
@@ -38,11 +39,8 @@ async function init(): Promise<false | void> {
 		</a>
 	);
 
-	if(select('.reponav-dropdown')) {
-		select('.reponav-dropdown')!.before(releasesTab);
-	} else {
-		select.last('.reponav-item')!.after(releasesTab);
-	}
+	const selector = select('.reponav-dropdown') ? '.reponav-dropdown' : '[href$="settings"]';
+	appendBefore('.reponav', selector, releasesTab);
 
 	if (isReleasesOrTags()) {
 		const selected = select('.reponav-item.selected');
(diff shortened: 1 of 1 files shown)


sharkykh:
Test this on a repo with admin access, I believe the releases tab should be before "Settings".
Right now "Settings" is always the last nav item...

kidonng:
I forgot the settings button. Solved that.

sharkykh:
Well, now this won't work for repositories where you don't have admin access 😅

Also need to consider that this won't work as expected if the repository name ends in `settings`, like:
https://github.com/probot/settings
https://github.com/ledermann/rails-settings
(The selector would select the `Code` nav link)

kidonng:
> Well, now this won't work for repositories where you don't have admin access 😅

Does it? Works fine on my browser...

sharkykh:
> > Well, now this won't work for repositories where you don't have admin access 😅
> 
> Does it? Works fine on my browser...

I haven't actually tried, just looked at the code - How could this work if there is no link that ends in `settings` when you don't have admin access?

kidonng:
It's handled by the util:

https://github.com/sindresorhus/refined-github/blob/0b2400511c28c9e22edbfc18770f113eafcecfc4/source/libs/dom-utils.ts#L25-L37

kidonng:
> Also need to consider that this won't work as expected if the repository name ends in `settings`

Note that the code will only select direct children, so it's fine.

P.S. I spent some time figure out why it doesn't work https://github.com/probot/settings you give, and eventually realized it has no release 🤣

tunnckoCore:
Just to mention, it's probably not exactly related to this PR but today I noticed that `more-dropdown` is kind of bugged and was always showing the dropdown, so I disabled it.

deadpyxel:
@tunnckoCore That was reported in #2357 

@kidonng made #2359 to solve it.

tunnckoCore:
Ah, cool. Sorry for messing here, but didn't find it earlier :)

fregante:
We’re not gonna support all combinations of supported features.

PR #2375, opened 2019-08-23
Hide unnecessary tip on comment field

### LINKED ISSUE #566: Don't hide new comment toolbar
screen shot 2017-07-04 at 11 16 18

This is an anti-feature. It does not remove UI bloat. Instead it removes the ability to easily quote an entire block of text, to easily convert a few lines in an ordered list, etc.
The toolbar is not just about typing Markdown. It has editing capabilities.

Since you're against adding options, I suggest to just remove that feature.

### LINKED ISSUE #2311: Readd URL format
Hi :hugs: 

I looked into the settings, while I was unable to find any option to restore the URL formating (that one which you use to first mark a certain text and then press it, so it becomes [text]() automatically.

### LINKED ISSUE #2349: Empty comment field displays unnecessary "please fill out this field" tooltip when hovered over in Firefox
As stated in the title, this tooltip appears when mousing over the comment field while it is empty. I know how issue comments work, I don't need this appearing and obscuring my text every time I go to write a comment. This is something which happens in normal GitHub, so I was hoping Refined GitHub would be able to fix it. I guess that makes this a feature request?

This only happens in Firefox as far as I can tell, which is possibly why it seems to have gone unnoticed. It's definitely a part of the site itself though, and not the browser: this message *will* appear in Chrome, but only when hovering over the issue title field, and not the comment field. In that case it's not really a problem though, since that's the default focused text box and the tooltip appears low enough that it doesn't cover the text anyway.

Here's a screenshot of the tooltip in question.
image

kidonng:
Closes #2349
Closes #566 
Closes #2311

# Test

This PR

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -163,7 +163,7 @@ Thanks for contributing! 🦋🙌
 - [](# "comment-fields-keyboard-shortcuts") [Adds a shortcut to edit your last comment: <kbd>↑</kbd>.](https://github.com/sindresorhus/refined-github/pull/961) (Only works in the following comment field, if it’s empty.)
 - [](# "comment-fields-keyboard-shortcuts") [Wraps selected text when pressing one of Markdown symbols instead of replacing it:](https://user-images.githubusercontent.com/37769974/59958878-39c51500-94cb-11e9-910a-061bf8ca6575.gif) `[` <code>\`</code> `'` `"` `*` `~` `_`
 - [](# "minimize-upload-bar") [Reduces the upload bar to a small button.](https://user-images.githubusercontent.com/55841/59802383-3d994180-92e9-11e9-835d-60de67611c30.png)
-- [](# "clean-rich-text-editor") [Hides unnecessary comment field tooltips and toolbar items.](https://user-images.githubusercontent.com/1402241/53629083-a4fe8900-3c47-11e9-8211-bfe2d254ffcb.png) (each one has a keyboard shortcut)
+- [](# "clean-rich-text-editor") [Hides unnecessary comment field tooltips and toolbar items](https://user-images.githubusercontent.com/1402241/53629083-a4fe8900-3c47-11e9-8211-bfe2d254ffcb.png) (each one has a keyboard shortcut.)
 - [](# "monospace-textareas") Use a monospace font for all textareas.
 
 <!-- Refer to style guide above. Keep this message between sections. -->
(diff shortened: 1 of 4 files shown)


fregante:
If you reuse that name, it will make it appear as a disableable feature but disabling it won’t make the buttons reappear.

kidonng:
Then should this single line goes to a feature? Do you have any suggestion for the name?

kidonng:
Maybe this should go to `hide-obvious-tooltips` feature instead.

fregante:
Good call about that feature, but my suggestion still stands: if you reuse a CSS feature and make it JS, you have to make sure that disabling the JS feature will **also** avoid applying the CSS. 

Look at the social count hider feature (forgot the name) to see how that works.

jerone:
> Look at the social count hider feature (forgot the name) to see how that works.

https://github.com/sindresorhus/refined-github/blob/master/source/features/hide-watch-and-fork-count.tsx#L5

jerone:
This will also fix https://github.com/sindresorhus/refined-github/issues/2311#issuecomment-519409302

fregante:
Only issue: the class is added a bit late. You can duplicate `feature.add` and don’t set the `load` method: this feature’s init will only add a class, once, as soon as possible. 

Look at how some other features have two `feature.add` calls and follow that style.

jerone:
I can't seem to find this feature `clean-rich-text-editor` in the options list...

image

_Running latest source code._

busches:
@jerone the feature addition logic is duplicated for some reason: https://github.com/sindresorhus/refined-github/blob/986c49ac811a06837571878f3a861e12de67295d/source/features/clean-rich-text-editor.ts#L15-L34 I'd wager removing the second set would fix the issue.

jerone:
@busches Double features logic is intentional and should not break the add-on. Other features have it too, and those are visible in the Options. Like `recently-pushed-branches-enhancements`, `fit-textareas`, etc.

Can you confirm you don't have this feature `clean-rich-text-editor` option either?

busches:
I totally don't.

jerone:
I think that the values of [`description` and `screenshot`](https://github.com/sindresorhus/refined-github/blob/986c49ac811a06837571878f3a861e12de67295d/source/features/clean-rich-text-editor.ts#L28-L29) the second feature logic are wrong and need to be `false`:

```
features.add({
	id: __featureName__,
	description: false,
	screenshot: false,
	...
});
```

jerone:
> I totally don't.

Meaning that you can see or cannot see this feature in the Options? (sorry for asking for confirmation).

busches:
I don't see it in options.

fregante:
The file should work, but it doesn’t because it doesn’t have a `tsx` extension. All features have a `tsx` extension’s regardless of whether they use JSX or not, just to avoid future renames, etc. 

The fix is to rename the file to  `*.tsx`

jerone:
Oh wow, I totally missed that.

PR #2379, opened 2019-08-24
Fix can't open dropdowns on Gists

### LINKED ISSUE #2378: Can't click buttons on `All gists` page.
I was able to fix this problem by disabling this extension, so I can confirm that it's the cause.

I can't click on the buttons as seen on the screenshot, on the `All gists` page. (Go to [Gists](https://gist.github.com) then click on `See all of your gists`).

All gists - Buttons

kidonng:
Closes #2378

# Test

https://gist.github.com/defunkt


kidonng:
`(embed failed)` ? Maybe we should fix this too

fregante:
Thank you!

PR #2381, opened 2019-08-25
Cleanup mess around formatting/linkifying code

### LINKED ISSUE #2263: Cleanup: there's some duplicate code in the linkifier
[libs/linkify-text-nodes.ts](https://github.com/sindresorhus/refined-github/blob/master/source/libs/linkify-text-nodes.ts) was extracted from [features/linkify-urls-in-code.tsx](https://github.com/sindresorhus/refined-github/blob/master/source/features/linkify-urls-in-code.tsx) in #1780 but its code was left behind, likely duplicate.

It'd be good to delete the duplicate code from `linkify-urls-in-code.tsx`

SirWindfield:
Closes #2263.

# Test

https://github.com/SirWindfield/refined-github/blob/master/.travis.yml


source/features/linkify-urls-in-code.tsx
```diff
@@ -61,17 +26,12 @@ function init(): false | void {
 			continue;
 		}
 
-		const linkified = linkifyUrls(element.textContent!, options);
-		if (linkified.children.length === 0) { // Children are <a>
-			continue;
-		}
-
-		zipTextNodes(element, linkified);
+		linkify(linkifyUrls, element, element.textContent!);
```
fregante: zipTextNodes is probably still needed  to support linkification of URLs in diffs (see previous related PRs in the blame view) and I wonder if it should be made part of `linkify` itself (maybe not, but worth a double-check)

SirWindfield: I wasn't actually sure what to do with `zipTextNodes`. I had no element array to pass to it so I left it out. I thought you would bring it up if it is needed :)

What would be the best call here? As mentioned above, the previous method won't work anymore. I could add it to `linkify` though...

fregante: Not sure, either way it probably needs to be added somewhere (haven’t tested this, but I guess it was there for that reason)

SirWindfield: Just checked, it basically puts HTMLElements together. Example taken from the Github page:

```js
const base = <>I live in <a href="it">Italy</a></>;
const grammar = highlightVerb(base);
// <>I <em>live</em> in Italy</> // e.g. the link was lost

zipTextNodes(base, grammar);
// <>I <em>live</em> in <a href="it">Italy</a></>
// The new `em` is copied from `grammar` to `base`
```

I guess this is really needed. I will add it to the library file so that calling `linkify` will automatically zip them.

source/libs/linkify-text-nodes.ts
```diff
@@ -60,5 +61,9 @@ export default (
 
 			textNode.replaceWith(linkified);
 		}
+
+		if (el instanceof Element) {
```
SirWindfield: A call to this utility function is made from within `options.ts`. The passed type is `DocumentFragment` but `zipTextNodes` requires `Element`. Sadly, `DocumentFragment` is not of type element. So I added this check. It won't change the behaviour or signatures of other functions that rely on this method.

source/options-storage.ts
```diff
@@ -32,6 +32,7 @@ for (const feature of __featuresList__) {
 
 const migrations = [
 	featureWasRenamed('make-discussion-sidebar-sticky', 'sticky-discussion-sidebar'), // Merged on August 1st
+	featureWasRenamed('linkify-code', 'linkify-code'), // Merged on August 1st
```
jerone: `linkify-urls-in-code`

fregante:
```suggestion
	featureWasRenamed('linkify-code', 'linkify-urls-in-code'), // Merged on September 1st
```

PR #2387, opened 2019-08-28
Consider more positive reactions in `highest-rated-comment`

### LINKED ISSUE #2190: Count more reactions when considering the highest rated comment
Because of #2108 we can see the most helpful comment (thanks @lubien!). At the moment that only counts `👍` as a positive and `👎` as a negative. But there are more reactions that can be used to indicate a post is helpful or not. How about we also count those?

| Positive | Negative |
| :------: | :------: |
| 👍        | 👎        |
| 😄        | 😕        |
| 🎉        |          |
| ❤️        |          |

I've left the ambiguous `👀` and (the slightly less ambiguous) `🚀` out.

CTrando:
Closes #2190

# Test
From original to make sure it still works:
leafo/sticky-kit#201
facebook/react#13991
flutter/flutter#11609 
MrMEEE/bumblebee-Old-and-abbandoned#123
Example of likes vs hearts + likes (also my project :smiley:)
https://github.com/microsoft/python-language-server/issues/19

### DIFF
--- a/source/features/highest-rated-comment.tsx
+++ b/source/features/highest-rated-comment.tsx
@@ -96,15 +96,16 @@ function linkBestComment(bestComment: HTMLElement): void {
 }
 
 function getWatchedReactions(): HTMLElement[] {
-	return positiveReactions.flatMap(reaction => select.all(`${COMMENT_SELECTOR} ${reaction}`));
+	// Map reaction [aria-label*=...] to .js-item [aria-label*=...], then join and pass to select.all
+	return select.all(positiveReactions.map(reaction => `${COMMENT_SELECTOR} ${reaction}`).join(','));
 }
 
 function getNegativeReactions(reactionBox: HTMLElement): HTMLElement[] {
-	return negativeReactions.flatMap(reaction => select.all(`${reaction}`, reactionBox));
+	return select.all(negativeReactions.join(','), reactionBox);
 }
 
 function getPositiveReactions(reactionBox: HTMLElement): HTMLElement[] {
-	return positiveReactions.flatMap(reaction => select.all(`${reaction}`, reactionBox));
+	return select.all(positiveReactions.join(','), reactionBox);
 }
 
 function getCount(reactions: HTMLElement[]): number {
--- a/source/features/highest-rated-comment.tsx
+++ b/source/features/highest-rated-comment.tsx
@@ -29,11 +29,9 @@ function init(): false | void {
 
 function getBestComment(): HTMLElement | null {
 	let highest;
-	for (const posReaction of getWatchedReactions()) {
-		const nearestComment = posReaction.closest(COMMENT_SELECTOR) as HTMLElement;
-
-		const likes = getPositiveReactions(nearestComment);
-		const dislikes = getNegativeReactions(nearestComment);
+	for (const comment of getCommentsWithReactions()) {
+		const likes = getPositiveReactions(comment);
+		const dislikes = getNegativeReactions(comment);
 
 		const likeCount = getCount(likes);
 		const dislikeCount = getCount(dislikes);
@@ -44,16 +42,16 @@ function getBestComment(): HTMLElement | null {
 		}
 
 		if (!highest || likeCount > highest.count) {
-			highest = {nearestComment, count: likeCount};
+			highest = {comment, count: likeCount};
 		}
 	}
 
 	// If count is not high enough don't bother telling user
-	if (!highest || highest.count < 10 || !highest.nearestComment) {
+	if (!highest || highest.count < 10 || !highest.comment) {
 		return null;
 	}
 
-	return highest.nearestComment;
+	return highest.comment;
 }
 
 function highlightBestComment(bestComment: HTMLElement): void {
@@ -71,7 +69,7 @@ function highlightBestComment(bestComment: HTMLElement): void {
 function linkBestComment(bestComment: HTMLElement): void {
 	// Find position of comment in thread
 	const position = select.all('.js-timeline-item').indexOf(bestComment);
-	// Only insert element if there are enough comments in the thread to warrant it
+	// Only link to it if it doesn't already appear at the top of the conversation
 	if (position >= 3) {
 		const text = select('.comment-body', bestComment)!.textContent!.substring(0, 100);
 		const avatar = select('.timeline-comment-avatar', bestComment)!.cloneNode(true);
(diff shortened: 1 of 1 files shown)


source/features/highest-rated-comment.tsx
```diff
@@ -63,6 +95,22 @@ function init(): false | void {
 	}
 }
 
+function getWatchedReactions(): HTMLElement[] {
+	return positiveReactions.flatMap(reaction => select.all(`${COMMENT_SELECTOR} ${reaction}`));
```
jerone: Nice use of `flatMap` :+1:

fregante: What's "watched reactions"?

source/features/highest-rated-comment.tsx
```diff
@@ -63,6 +95,22 @@ function init(): false | void {
...
+}
+
+function getNegativeReactions(reactionBox: HTMLElement): HTMLElement[] {
+	return negativeReactions.flatMap(reaction => select.all(`${reaction}`, reactionBox));
+}
+
+function getPositiveReactions(reactionBox: HTMLElement): HTMLElement[] {
+	return positiveReactions.flatMap(reaction => select.all(`${reaction}`, reactionBox));
+}
```
fregante: You can avoid these helper functions and just query the selector once:

```js
select.all(positiveReactions.join(','), reactionBox)
```

source/features/highest-rated-comment.tsx
```diff
@@ -4,48 +4,80 @@ import select from 'select-dom';
...
+	linkBestComment(bestComment);
+}
+
+function getBestComment(): HTMLElement | null {
 	let highest;
-	// `.js-timeline-item` excludes the very first comment
-	for (const like of select.all('.js-timeline-item [aria-label*="reacted with thumbs up"]')) {
-		const count = getCount(like);
-		const dislike = select('[aria-label*="reacted with thumbs down"]', like.parentElement!);
+	for (const posReaction of getWatchedReactions()) {
```
fregante: I think here you're getting the same comment up to 3 times. I'd rather see:

```
for each '.js-timeline-item'
	count positive reactions
	if count < 10
		continue
	count negative reactions
	... ignore controversial, etc
endfor
```

CTrando: Yup that is intentional. I actually tried iterating over all the comments but I found there was a significant slow down compared to going over only the positive reactions.

source/features/highest-rated-comment.tsx
CTrando: Thanks, although I think the method below is better. Not used to select-dom api so didn't know that shortcut.

source/features/highest-rated-comment.tsx
CTrando: ok

CTrando: I prefer to leave the helper functions in though for future refactoring and readability.

source/features/highest-rated-comment.tsx
fregante: Then I'd still suggest this

```rb
for each commentsWithPositiveReactions
	count positive reactions
	if count < 10
		continue
	count negative reactions
	... ignore controversial, etc
end
```

`commentsWithPositiveReactions` is something like:

```js
new Set(getPositiveReactions().map(reaction => reaction.closest('comment')))
```

This Set will include each unique comment, not the reactions.

This way it will skip any counting if the comment doesn't pass the base requirement of 10 votes, which most comments don't.

source/features/highest-rated-comment.tsx
```diff
@@ -4,48 +4,80 @@ import select from 'select-dom';
...
 		>
 			Highest-rated comment
 		</span>
 	);
+}
+
+function linkBestComment(bestComment: HTMLElement): void {
+	// Find position of comment in thread
+	const position = select.all('.js-timeline-item').indexOf(bestComment);
+	// Only insert element if there are enough comments in the thread to warrant it
```
fregante:
```suggestion
	// Only link to it if it doesn't already appear at the top of the conversation
```

source/features/highest-rated-comment.tsx
CTrando: Sure, I'll test it out. I was initially worried because I thought that since Sets use reference equality then perhaps the results from two different calls might have different references and fail the equality check but maybe that's not the case.

source/features/highest-rated-comment.tsx
CTrando: ok

source/features/highest-rated-comment.tsx
```diff
@@ -63,6 +93,27 @@ function init(): false | void {
 	}
 }
 
+function getCommentsWithReactions(): Set<HTMLElement> {
+	// Map reaction [aria-label*=...] to .js-item [aria-label*=...], then join and pass to select.all
+	const reactions = select.all(positiveReactions.map(reaction => `${COMMENT_SELECTOR} ${reaction}`).join(','));
+	// Find closest comment to each reaction, cast to HTMLElement, then filter out null values
+	const comments = reactions.map(reaction => reaction.closest(COMMENT_SELECTOR))
+		.map(comment => comment as HTMLElement).filter(comment => comment);
```
fregante: It's impossible that there are null values since the initial selector has `COMMENT_SELECTOR`

To fix the types, this just needs to be

```ts
function getCommentsWithReactions(): Set<HTMLElement> {
	const comments = getPositiveReactions().map(reaction => reaction.closest<HTMLElement>(COMMENT_SELECTOR)!);
	return new Set(comments);
}
```

And you can include `COMMENT_SELECTOR` in the original `positiveReactions` array.

Also probably you can avoid the `HTMLElement` altogether:



```ts
function getCommentsWithReactions(): Set<Element> {
	const comments = getPositiveReactions().map(reaction => reaction.closest(COMMENT_SELECTOR)!);
	return new Set(comments);
}
```

source/features/highest-rated-comment.tsx
```diff
@@ -4,48 +4,78 @@ import select from 'select-dom';
...
 			aria-label="This comment has the most positive reactions on this issue."
 		>
 			Highest-rated comment
 		</span>
 	);
+}
+
+function linkBestComment(bestComment: HTMLElement): void {
+	// Find position of comment in thread
+	const position = select.all('.js-timeline-item').indexOf(bestComment);
```
fregante:
```suggestion
	const position = select.all(COMMENT_SELECTOR).indexOf(bestComment);
```

There's a button to apply these suggestions in a click right here on GitHub ⬆️. It will also resolve this comment.

CTrando: Oh that's cool I never saw that before. I just did it manually anyways though thanks!

source/features/highest-rated-comment.tsx
CTrando: I got a syntax error with the first way, I looked at the source and I think it only allows for types that extend certain bases to be passed as type parameters. 
I went with the second way instead.

fregante: The code still doesn't match my suggestion

CTrando: The suggested code did not compile without some default parameters and I didn't know that document could be a default parameter because in C# default parameters must be compile time constant.

CTrando:
Question, how come you swapped the default parameter for the null parameter and that or construct? I'm not too experienced with typescript but is there a difference?

fregante:
`(reactionBox = document)` will set the type of `reactionBox` to `Document` but we need it to stay `Element` so the other calls will keep working.

Instead of changing the type of the parameter, I thought it would be easier to just make it optional and then handle its absence later in the code (`reactionBox || document`)

CTrando:
Ah I see, sorry about that, I didn't see the compile errors but since it still ran I thought it was fine. Too used to static languages 😄 I'll test this out today or tomorrow and merge, thanks for the explanation.

CTrando:
I'll wait for another day just to be safe if you want to do anything, but I tested it out on the example cases and it works fine.

fregante:
I tested it as well and it seems to work correctly

nikoladev:
Thanks @CTrando!

PR #2391, opened 2019-08-29
Update dependencies and lint

fregante:


package.json
```diff
@@ -105,13 +105,17 @@
 		"rules": {
 			"no-alert": "off",
 			"import/no-unassigned-import": "off",
-			"import/no-unresolved": "warn"
+			"import/no-unresolved": "warn",
+			"import/named": "warn"
```
fregante: https://github.com/benmosher/eslint-plugin-import/issues/1341

package.json
```diff
@@ -105,13 +105,17 @@
...
 		},
 		"overrides": [
 			{
 				"files": "**/*.+(ts|tsx)",
 				"extends": "xo-typescript",
 				"rules": {
+					"@typescript-eslint/no-misused-promises": "off",
```
fregante: Annoying: It doesn't let us use `async` functions as event handlers.

package.json
```diff
@@ -105,13 +105,17 @@
...
+					"@typescript-eslint/require-await": "off",
```
fregante: Buggy: it doesn't recognize returning a promise. It's already on top of some arguably annoying rules and it conflicts with them:

e.g.

```js
() => Promise.resolve()
  ✖  173:17  Functions that return promises must be async. @typescript-eslint/promise-function-async
```

```js
async () => Promise.resolve()
  ✖  178:17  Async arrow function has no await expression. @typescript-eslint/require-await
```

```js
async () => await Promise.resolve()
  ✖  173:29  Redundant use of await on a return value. no-return-await
```

The first one was perfectly fine.

package.json
```diff
@@ -105,13 +105,17 @@
...
+					"@typescript-eslint/no-explicit-any": "off",
```
fregante: Good in theory, but makes us have to type on-off things just for the heck of it:

- Code examples:


	```ts
	(window as any).select = select;
	```


	```ts
	const editor: CodeMirrorInstance = document.querySelector<any>('.CodeMirror').CodeMirror;
	```

	```ts
	(global as any).navigator = window.navigator;
	(global as any).document = window.document;
	(global as any).location = new URL('https://github.com');
	// Which becomes at least, with no advantage
	(global as unknown as typeof window).navigator = window.navigator;
	(global as unknown as typeof window).document = window.document;
	(global as unknown as typeof window).location = new URL('https://github.com');
	```

- It'd force us to type API calls
- Also kills `AnyObject` which is pretty much required for the API handler

source/globals.d.ts
```diff
@@ -1,3 +1,5 @@
+// TODO: Drop some definitions when their related bugs are resolved
```
fregante: I merged all of these TODOs into one. They're just noise since we can only wait.

package.json
fregante: https://github.com/typescript-eslint/typescript-eslint/pull/674#issuecomment-526128992

package.json
```diff
@@ -51,41 +51,41 @@
...
-		"@types/mini-css-extract-plugin": "^0.2.1",
-		"@types/react": "^16.8.23",
+		"@types/mini-css-extract-plugin": "^0.8.0",
+		"@types/react": "^16.9.2",
 		"@types/terser-webpack-plugin": "^1.2.1",
-		"@typescript-eslint/eslint-plugin": "^1.11.0",
-		"@typescript-eslint/parser": "^1.11.0",
-		"ava": "^2.2.0",
+		"@typescript-eslint/eslint-plugin": "^1.13.0",
+		"@typescript-eslint/parser": "^1.13.0",
```
fregante: I'm temporarily avoiding v2 because of a bug I don't know how to work around: https://github.com/typescript-eslint/typescript-eslint/issues/890

>   webpack.config.ts:undefined:undefined
>   ✖   0:0   Parsing error: If "parserOptions.project" has been set for @typescript-eslint/parser, ./webpack.config.ts must be included in at least one of the projects provided.         
> 
>   test/page-detect.ts:undefined:undefined
>   ✖   0:0   Parsing error: If "parserOptions.project" has been set for @typescript-eslint/parser, ./test/page-detect.ts must be included in at least one of the projects provided.       
> 
>   test/utils.ts:undefined:undefined
>   ✖   0:0   Parsing error: If "parserOptions.project" has been set for @typescript-eslint/parser, ./test/utils.ts must be included in at least one of the projects provided.             
> 
>   test/fixtures/globals.ts:undefined:undefined
>   ✖   0:0   Parsing error: If "parserOptions.project" has been set for @typescript-eslint/parser, ./test/fixtures/globals.ts must be included in at least one of the projects provided. 

They should go away by adding `test` in tsconfig's `includes`, but that file appears to be ignored.

Also reported in https://github.com/xojs/eslint-config-xo-typescript/issues/15

package.json
sindresorhus: While it's convenient, my reasoning is that using async there makes it hard to see that the event handler function will cause an uncaught rejection if the `await` rejects. For example, using async in a `new Promise` can cause hard to track down issues.

And when people see:

```js
[1, 2, 3].forEach(async value => {
  await doSomething(value);
});
```

they might be fooled into thinking that it awaits for each iteration, which it does not.

I'm happy to discuss this though. Maybe there some additional config that could be added to the rule?

package.json
sindresorhus: That rule has other problems too. Disabled: https://github.com/xojs/xo/commit/ee145cbe719073aeccd349af4bb85ddc542a2818

package.json
sindresorhus: The idea was that you explicitly disable the rule where you want `any` to make it inconvenient enough to use that it's not misused, but I encountered some annoyance with it too, so I'll disable it for now: https://github.com/xojs/eslint-config-xo-typescript/commit/bbbf88972520e8af3439ba2ab3f1a1c40e68041c

package.json
fregante: I understand the value in that case, but I don't have a solution for our use-case: our handlers are `async` just because `Promise` handling is great, not because anyone expects `addEventListener` to `await` them.

If we enable this rule, we have to write listeners as:

```js
function handler() {
	(async () => {
		document.body.append(await fetchDom(...));
	})();
}
```

fregante: A more useful related rule would be the opposite: always await async functions. Does this exist already?


```js
async function set(){...}
async function get(){...}
set(1); // Forgot `await`
await get();
```

package.json
fregante: > The idea was that you explicitly disable the rule where you want `any` to make it inconvenient enough to use that it's not misused

Maybe that is worth it, I missed this one just recently: https://github.com/sindresorhus/refined-github/pull/2393/files#diff-f8fed5d366f31fce028a9c8ceca38c3eR6

The lines in my examples would just have to use `eslint-disable-next-line` or be wrapped by `disabled/enable`

sindresorhus: I think it will be easier to enable in a year or so when TS has matured more and we need less `any` in general.

package.json
sindresorhus: > A more useful related rule would be the opposite: always await async functions. Does this exist already?

This? https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-floating-promises.md

package.json
fregante: What other problems have you had?

Do you think there will ever be a solution to my examples? They basically are:

- setting a global (without having to create a property on Global/Window)
- setting/getting one-off properties. It could probably already be fixed with:
	```ts
	type CodeMirrorElement = CodeMirrorInstance & {CodeMirror: CodeMirrorInstance}
	const editor = document.querySelector<CodeMirrorElement>('.CodeMirror').CodeMirror;
	```
- we just want to avoid typing of certain parts like APIs because it's extra work that **has to** be updated when the query changes, otherwise the API response won't actually match the types (in essence, manually typing the API is as bad as an `as`sertion). Also the API response ALWAYS has to start from `AnyObject`, there's no workaround for that.

package.json
fregante: Almost, `returnsPromise()` is not awaited even though it returns a promise. It seems that it requires `await` only on variables

package.json
sindresorhus: I had problems with some bad types from Electron and React. Basically out of my control. I also had to use `any` for some things like this: https://github.com/sindresorhus/caprine/blob/7965b36c11ae1f7ccb2185bebdc85e9fc3d01078/source/menu.ts#L92

PR #2393, opened 2019-08-30
Add `reload-failed-proxied-images` feature

GramThanos:
# Description
Many times the badges images fail to load. This feature detects such failed to load images and tries to reload them so that the user don't have to manually reload the hole page.

# Test
On the issue https://github.com/badges/shields/issues/1568 there is a 4 seconds delayed badge that will always fail to load (the GitHub's proxy timeouts) you can see on the console tab that the feature will try to reload that image up to 6 times.

image

Test image with 40% fail rate (no cache):
reload-failed-proxied-images

# Preview
reload-failed-proxied-images

reload-failed-proxied-images


source/features/reload-failed-proxied-images.tsx
```diff
@@ -0,0 +1,79 @@
...
+		}
+	});
+
+	// Handle failure of loading the image
+	// (maybe, the original image source failed to provide the image to the proxy in the 4 seconds timeframe)
+	tester.addEventListener('error', () => {
+		let tries = parseInt(image.dataset.brokenProxiedImage || '0', 10);
+
+		// Flag image (to limit max tries)
+		if (tries === 0) {
```
busches: The logic this statement is doing and the one in the else look equivalent and this one can be removed.

source/features/reload-failed-proxied-images.tsx
```diff
@@ -0,0 +1,79 @@
...
+			// Tested loaded the image so now the visual broken image can be updated
+			// Update image
+			image.src = String(image.src);
+		}
+	});
+
+	// Handle failure of loading the image
+	// (maybe, the original image source failed to provide the image to the proxy in the 4 seconds timeframe)
+	tester.addEventListener('error', () => {
+		let tries = parseInt(image.dataset.brokenProxiedImage || '0', 10);
```
busches: We can pass `tries` into the `test` function and avoid updating the DOM.

GramThanos:
> 👎 on this feature.
> 
> The reason why badges might not be loading is due to traffic issues. We should not make the problem worse.

I get your point but I think the problem is both traffic and cache related. So since you already made the 1st request, the badges server has already started creating and caching your image, an additional request will not cause that much trouble assuming that the image is cached and ready to be served. Thus we can for example allow only 1 or 2 tries with a bigger delay (maybe 5 seconds).

kidonng:
@fregante This problem is not limited to badges (based on my experiences).
Though I agree this "is due to traffic issue" to some degree, if @GramThanos can handle it right I would like to have this useful feature. There are lengthy READMEs (and other documents) you just don't want to refresh and wait.

source/features/reload-failed-proxied-images.tsx
GramThanos: I am not sure if there is a case where `init` will be fired (by the `onAjaxedPages`) while the page will still have the same images.

source/features/reload-failed-proxied-images.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+async function init(): Promise<false | void> {
+	// Get each image on the page
+	select.all('img').forEach(image => {
+		// If it is a proxied image
+		if (isGithubProxiedImg(image)) {
+			test(image);
+		}
+	});
+}
```
fregante: >  an additional request will not cause that much trouble assuming that the image is cached and ready to be served.

Let's retry once, then.

If you use https://github.com/fregante/image-promise this whole feature can just be:

```ts
async function handleErroredImage({target}): Promise<void> {
	await delay(5000);
	try {
		// A clone image retries downloading
		// `loadImage` awaits it
		// If successfully loaded, the failed image will be replaced.
		target.replaceWith(await loadImage(target.cloneNode()));
	} catch {}
}

function init(): void {
	delegate('img[src^="https://camo.githubusercontent.com"]', 'error', handleErroredImage);
}

features.add({
	id: __featureName__,
	description: 'Retries downloading images that failed downloading due to GitHub limited proxying.',
	screenshot: false,
	init
});
```

GramThanos: Based on my tests/debugging `delegate` does not work in this case. I think GitHub gabs the errors for statistics and the event does not bubble up to the body element.

fregante: It works if you pass a `true` as the last parameter of `delegate` to make it a passive listener.

fregante:
Thank you ^_^

paulmelnikow:
Hi from Shields! I noticed this was mentioned from badges/shields#1568.

I'll give my belated 👍on this as the general alternative is reloading the entire readme which typically forces a reload of all the images. It's both a better experience and overall less traffic to reload just the broken images.

Can I ask whether this reloads the images just once, or if it repeatedly attempts to reload the images? I'd favor the former, or else it'd be a good idea to include some kind of backoff.

By the way, we're tracking the current capacity problem at badges/shields#3874 and hope to have some resolution shortly!

GramThanos:
@paulmelnikow Due to various traffic concerns, it reloads the failed images 5 seconds after the error event, only once.

Issue #2399, opened 2019-09-03
`wait-for-build` throws error on PR page


janpio:
(Firefox 68.0.2 (64-bit) on Windows 10, plugin version 19.9.2)

Happens for me on for example this PR: https://github.com/fastlane/fastlane/pull/15255

![image](https://user-images.githubusercontent.com/183673/64195618-357d8680-ce82-11e9-9cd4-eaff34fd1b2f.png)

```
19:36:52.676 ❌ wait-for-build content.js:2837:15
19:36:52.702
TypeError: "select_dom_default.a.last(...) is undefined"
    getLastCommit moz-extension://57b5be31-7fbe-4b7d-bae1-b2f5de44e6b4/content.js:5401
    addEventListener moz-extension://57b5be31-7fbe-4b7d-bae1-b2f5de44e6b4/content.js:5436
    wait_for_build_init moz-extension://57b5be31-7fbe-4b7d-bae1-b2f5de44e6b4/content.js:5519
    features_run moz-extension://57b5be31-7fbe-4b7d-bae1-b2f5de44e6b4/content.js:2833
    add moz-extension://57b5be31-7fbe-4b7d-bae1-b2f5de44e6b4/content.js:2880
    onAjaxedPages moz-extension://57b5be31-7fbe-4b7d-bae1-b2f5de44e6b4/content.js:2795
    onAjaxedPagesRaw moz-extension://57b5be31-7fbe-4b7d-bae1-b2f5de44e6b4/content.js:2790
content.js:2838:15
    features_run moz-extension://57b5be31-7fbe-4b7d-bae1-b2f5de44e6b4/content.js:2838
    add moz-extension://57b5be31-7fbe-4b7d-bae1-b2f5de44e6b4/content.js:2880
    onAjaxedPages moz-extension://57b5be31-7fbe-4b7d-bae1-b2f5de44e6b4/content.js:2795
    onAjaxedPagesRaw moz-extension://57b5be31-7fbe-4b7d-bae1-b2f5de44e6b4/content.js:2790
```

I don't know if this actually breaks anything, but I guess it should be fixed anyway.

---

Same on Chrome:
![image](https://user-images.githubusercontent.com/183673/64196251-be48f200-ce83-11e9-9cee-5e9de5f91841.png)

jerone:
I cannot confirm this. Is there any other extension interfering?

janpio:
No, I tested with only Refined Github enabled.

[...]

PR #2402, opened 2019-09-04
Fix `sync-pr-commit-title` with new issue option

razh:
GitHub added a **Reference in new issue** option to comments that also
contain a `[name="issue[title]"]` element.

```html
<dl class="form-group">
	<dt><label for="convert-to-issue-title==">Title</label></dt>
	<dd><input id="convert-to-issue-title==" class="form-control" type="text" name="issue[title]" value="Example issue title" aria-label="Issue title" autofocus="" required=""></dd>
</dl>
```

Reference in new issue dialog

Resulting in this error:

```js
content.js:5341 Uncaught Error: Refined GitHub: `sync-pr-commit-title` can’t update the PR title
    at maybeShowNote (content.js:5341)
    at onMergePanelOpen (content.js:5358)
    at HTMLDivElement.<anonymous> (content.js:5293)
    at HTMLDivElement.listenerFn (content.js:364)
```

Because `maybeShowNote()` only expects two elements:

https://github.com/sindresorhus/refined-github/blob/84d0df7ef8a38afa0ab0011c1b674a8db45f0a8d/source/features/sync-pr-commit-title.tsx#L47-L50

# Test
The **Reference in new issue** option is available for comments in **Issues** & **Pull requests** in repos where you are an author or collaborator.

### DIFF
--- a/source/features/sync-pr-commit-title.tsx
+++ b/source/features/sync-pr-commit-title.tsx
@@ -8,7 +8,7 @@ import features from '../libs/features';
 import onPrMergePanelOpen from '../libs/on-pr-merge-panel-open';
 
 const commitTitleLimit = 72;
-const prTitleFieldSelector = '[name="issue[title]"]';
+const prTitleFieldSelector = '.js-issue-update [name="issue[title]"]';
 const prTitleSubmitSelector = '.js-issue-update [type="submit"]';
 
 const createCommitTitle = debounce<[], string>((): string => {


fregante:
If this is in reference to an existing issue, please use "Fixes #000issuenumber00" so it will be automatically closed (and will give additional context)

razh:
@fregante There's no existing issue. Should I create one?

fregante:
Indeed, I got confused with this other issue https://github.com/sindresorhus/refined-github/issues/2399

fregante:
Thank you for the fix, @razh 🌺

PR #2403, opened 2019-09-05
Fix `profile-hotkey`

### LINKED ISSUE #2400: Go to Profile is broken
The `g m` shortcut is not working, after a bit of digging into the code it is because of the selector `#user-links a.dropdown-item[href="/${getUsername()}"]` does not work because `#user-links` is not used anymore.

I would create a PR, but I am not sure how to best target the button as they don't use any ID for that specific dropdown. Any ideas? Maybe since it is not needed since there is very strict part of the selector: `[href="/${getUsername()}"]`? That should lead always the element which is needed?

AuHau:
Closes #2400 

# Test
any page and press `g m` to go to profile.

### DIFF
--- a/source/features/profile-hotkey.tsx
+++ b/source/features/profile-hotkey.tsx
@@ -3,7 +3,7 @@ import features from '../libs/features';
 import {getUsername} from '../libs/utils';
 
 function init(): false | void {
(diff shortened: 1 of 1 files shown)


fregante:
Thank you ☺️

PR #2410, opened 2019-09-07
Improve linkification in Chrome (dependency updates)

fregante:
Thanks to https://github.com/sindresorhus/linkify-issues/pull/6 and https://github.com/sindresorhus/linkify-urls/pull/29

Replaces https://github.com/sindresorhus/refined-github/pull/1936
Follows the cleanup in https://github.com/sindresorhus/refined-github/pull/2381

This change:

- lets us use the proper types for those packages
- improves the linkification: https://github.com/sindresorhus/issue-regex/issues/3  https://github.com/sindresorhus/linkify-urls/issues/7
- preserves Firefox compatibility, which users will keep using the previous version's regexes due to https://bugzilla.mozilla.org/show_bug.cgi?id=1225665


yakov116:
From v78 its has been added
https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/78#JavaScript

fregante:
Neat! In July we can update this then

```json
		"linkify-issues": "^2.0.0-nolookbehind",
		"linkify-urls": "^3.1.0-nolookbehind",
```

And the minimum Firefox version

PR #2420, opened 2019-09-09
Soften PR review comment requirement

### LINKED ISSUE #2418: Reviews shouldn't always require a comment
I made a small mistake in https://github.com/sindresorhus/refined-github/pull/2412. The comment isn't required if there are some review comments already. It should look for them before adding the `required` attribute.

PR welcome

dertieran:
This fixes reviews always requiring a comment even if other review comments are pending

Closes #2418

Not sure if this is a *"to simple"* solution, but this issue prevented me from reviewing so I wanted a *quick* fix :see_no_evil:


fregante:
How's this?

fregante:
It was annoying me too, thank you for the fix!

PR #2424, opened 2019-09-10
Feature/add copy code block

### LINKED ISSUE #2423: Permit to copy the single line code block with a click
Might be interesting to have a feature permitting to copy a code block via a click on a clipboard icon

Example URL:

image

oktapodia:
Adding a feature permitting to copy a code block via a click on a clipboard icon

Closes #2423 

# Test

image
Can be tested in every repository which includes a README with a codeblock, this feature add the copy icon only on the single line codeblocks, another PR will come soon to handle the multi-line code blocks (I preferred to split them because some people might wants to enable/disable these each by each)


source/features/copy-code-block.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+function handleClick({currentTarget: button}: React.MouseEvent<HTMLButtonElement>): void {
+	if (button.previousSibling && button.previousSibling.textContent) {
+		copyToClipboard(button.previousSibling.textContent);
+	}
+}
+
+function renderButton(): void {
+	for (const codeBlock of select.all('code')) {
+		console.log(codeBlock);
```
jerone: 🔥

source/features/copy-code-block.tsx
```diff
@@ -0,0 +1,38 @@
...
+				<button className="copy-code-block btn-link" aria-label="Copy the code block" onClick={handleClick}>
+					{clippy()}
+				</button>
+			);
+	}
+}
+
+function init(): void {
+	if (select.exists('code')) {
+		renderButton();
```
jerone: The whole `renderButton` function code can be placed inside the `init`.
And the extra `if (select.exists('code'))` check is redundant with the for-loop.

source/features/copy-code-block.tsx
```diff
@@ -0,0 +1,31 @@
+import './copy-code-block.css';
+import React from 'dom-chef';
+import select from 'select-dom';
+import copyToClipboard from 'copy-text-to-clipboard';
+import features from '../libs/features';
+import {clippy} from '../libs/icons';
```
jerone: The [convention](https://github.com/sindresorhus/refined-github/blob/2e4a24d3957c205c7bc49525c69e1fa08593d2ca/source/features/edit-readme.tsx#L4) is:

```suggestion
import * as icons from '../libs/icons';
```

And then use `{icons.clippy()}` later.

oktapodia:
Updated, thank you for your comments

fregante:
https://github.com/sindresorhus/refined-github/issues/2423#issuecomment-529876338

PR #2429, opened 2019-09-10
Stricter usage of `any`

fregante:
Follow-up to #2391 

Perhaps we can just disable `any` for these exceptions. In some cases it can be replaced with `AnyObject` for a cheap workaround. I don't want to risk passing what was in `reload-failed-proxied-images` anymore

### DIFF
--- a/source/content.ts
+++ b/source/content.ts
@@ -143,5 +143,5 @@ import './features/clean-rich-text-editor';
 import './features/highlight-collaborators-in-lists';
 
 // Add global for easier debugging
-/* eslint-disable-next-line @typescript-eslint/no-explicit-any */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
 (window as any).select = select;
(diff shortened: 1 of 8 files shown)


source/globals.d.ts
```diff
@@ -1,5 +1,7 @@
 // TODO: Drop some definitions when their related bugs are resolved
+// TODO: Improve JSX types for event listeners so we can use `MouseEvent` instead of `React.MouseEvent`, which is incompatible with regular `addEventListeners` calls
 
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
KubaJastrz:
```suggestion
// eslint-disable-next-line @typescript-eslint/no-explicit-any
```

source/content.ts
```diff
@@ -143,4 +143,5 @@ import './features/clean-rich-text-editor';
 import './features/highlight-collaborators-in-lists';
 
 // Add global for easier debugging
+/* eslint-disable-next-line @typescript-eslint/no-explicit-any */
```
sindresorhus:
```suggestion
// eslint-disable-next-line @typescript-eslint/no-explicit-any
```

PR #2436, opened 2019-09-13
Temporary fix for `widen-search-field`

### LINKED ISSUE #2435: Big input box on issue page, for search box
Screenshot attached, for Refined GitHub disabled and enabled. The search box is huge on issues pages.

Refined GitHub disabled:

1 - Refined GitHub - Disabled

Refined GitHub enabled:

2 - Refined GitHub - Enabled

KubaJastrz:
Technically fixes #2435

GitHub made some weird changes to their css regarding `.subnav` selector on Issues and Pull Requests pages. It looks like they are migrating from the old `.float-left/right` model to flexbox, but at the same time they ignore existing "semantic" classes.

You can also see it on the Releases page where they are using the old `.subnav` and `.d-flex.flex-justify-between`.

I would call this PR a temporary workaround until things settle down with GitHub changes. Also, we should probably expect some more broken features soon.

# Test

Open [issues](https://github.com/sindresorhus/refined-github/issues) or [pull request](https://github.com/sindresorhus/refined-github/pulls) page and see.

### DIFF
--- a/source/features/widen-search-field.css
+++ b/source/features/widen-search-field.css
@@ -7,6 +7,7 @@
 	flex-grow: 1;
 	display: flex;
 	font: none !important;
+	align-items: flex-start;
 }
 
 /* Move `New issue` button back to the right */
--- a/source/features/widen-search-field.css
+++ b/source/features/widen-search-field.css
@@ -7,7 +7,7 @@
 	font: none !important;
 }
 
-.d-flex[role="search"] {
+.d-flex[role='search'] {
 	flex-grow: 1;
 	margin-bottom: 20px;
 }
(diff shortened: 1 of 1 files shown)


fregante:
The `.subnav` is used across the site so I’m not keen on merging temporary fixes that aren’t tested on all pages where that bar appears. 

Either it’s tested and not temporary or we can’t risk merging this and breaking another part 😅

KubaJastrz:
Makes sense, in that case it would be best to disable this feature for the time being.

joak46:
I use Stylus extension to add custom css and fix this. unset keyword works fine too.
It will be the same as just delete these two selectors, i think ?



## Issue page


## Pull requests page


## Labels page

fregante:
Needs further testing, but this fixes the height. `widen-search-field` however remains broken.

It can be merged before the next release (in 13 hours) unless bugs are found.

PR #2439, opened 2019-09-15
Move testing to GitHub Actions

fregante:
I think GitHub Actions can replace Travis for testing. Configuration is verbose but it's only done once and the speed increase seems worth it.

Deployment can stay there until https://github.com/notlmn/browser-extension-template/issues/19, it should be unaffected by this PR because Travis still runs the daily cronjob.

PR and branch testing can later be disabled on https://travis-ci.org/sindresorhus/refined-github/settings
Screenshot 2019-09-16 at 03 58 02

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,7 +1,7 @@
-name: Run tests
+name: Test
 on: push
 jobs:
-  test:
+  run:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v1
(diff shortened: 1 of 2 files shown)


fregante:
This is neat:

Screenshot 2019-09-17 at 13 31 21

Every build from now on will have a built extension next to it. Artifacts likely won't stay there forever but it's a good addition when wanting users to test your PRs.

busches:
@fregante this doesn't appear to run on PRs from forks, was that intended?

sindresorhus:
GitHub Actions only run on PRs from users in the GitHub Actions beta.

busches:
And you're in the beta so it runs for the repo? As I have three PRs open, one from my fork and two from branches on the repo and the latter built and the former did not. Why it's running isn't super intuitive. 😀

PR #2440, opened 2019-09-16
Update dependencies and lint /2 

fregante:
I was able to solve the issues I had in #2391, so this updates the remaining linting dependencies to their latest version.


package.json
```diff
@@ -117,6 +117,7 @@
 				"rules": {
 					"@typescript-eslint/no-misused-promises": "off",
 					"@typescript-eslint/promise-function-async": "off",
+					"@typescript-eslint/no-explicit-any": "error",
```
fregante: - It was disabled upstream https://github.com/xojs/eslint-config-xo-typescript/commit/bbbf88972520e8af3439ba2ab3f1a1c40e68041c
- But here its support was fixed: https://github.com/sindresorhus/refined-github/pull/2429

package.json
```diff
@@ -107,8 +107,8 @@
 		"rules": {
 			"no-alert": "off",
 			"import/no-unassigned-import": "off",
-			"import/no-unresolved": "warn",
-			"import/named": "warn"
+			"import/no-unresolved": "off",
+			"import/named": "off"
```
fregante: Reportedly disabled upstream: https://github.com/xojs/xo/commit/ee145cbe719073aeccd349af4bb85ddc542a2818

...but it appears to still be enabled here, so I'm disabling it manually.

tsconfig.json
```diff
@@ -6,6 +6,8 @@
 		"esModuleInterop": true
 	},
 	"include": [
-		"source"
+		"source",
+		"test",
+		"webpack.config.ts"
```
fregante: This fixed the issue I was having in the previous PR.

package.json
sindresorhus: Not released yet.

PR #2442, opened 2019-09-17
Merge 2 features into `highlight-collaborators-and-own-discussions`

### LINKED ISSUE #2419: Merge/match `highlight-own-issues-and-prs` and `highlight-collaborators-in-lists`
I think that the style of `highlight-own-issues-and-prs` (https://github.com/sindresorhus/refined-github/pull/1782) wasn't very effective, I always have to squint to find my own discussions.

`highlight-collaborators-in-lists` (https://github.com/sindresorhus/refined-github/pull/2351) made this more obvious, its style it's much more visible:

I'm wondering... since `highlight-collaborators-in-lists` never highlights your own issues, why not just use the same style for both? Perhaps just with a slightly different color.

The features can be merged into one (`highlight-collaborators-and-own-discussions`) because they affect the same exact area, even if they're applied at different times.

cc @jerone

fregante:
Closes #2419 

## Test

https://github.com/pulls


source/options-storage.ts
```diff
@@ -31,8 +31,8 @@ for (const feature of __featuresList__) {
 }
 
 const migrations = [
-	featureWasRenamed('make-discussion-sidebar-sticky', 'sticky-discussion-sidebar'), // Merged on August 1st
 	featureWasRenamed('linkify-code', 'linkify-urls-in-code'), // Merged on September 1st
+	featureWasRenamed('highlight-collaborators-in-lists', 'highlight-collaborators-and-own-discussions'), // Merged on September 20th
```
jerone: And `highlight-own-issues-and-prs` ?

fregante: It will just be dropped by `removeUnused`

jerone: Ah yes, so `highlight-collaborators-in-lists` will be leading.

fregante: Yes. Migration are just to preserve disabled features, and `highlight-own-issues-and-prs` is such a minor feature to disable it just does not matter.

Issue #1676, opened 2018-12-19
Hide/collapse bots in commits


mike-north:
With the rise in popularity of dependency management bots and other non-human commit-generating entities, commit logs can get [very noisy](https://github.com/typed-ember/ember-cli-typescript/compare/v2.0.0-beta.3...v2.0.0-rc.1) relative to the typical experience in the pre "rise of the bots" era.

It would be nice to have an option for filtering these types of commits out of the list, particularly in the `/compare/X...Y` UI.

bfred-it:
There’s a problem with including this feature: it’s a very minor feature and might cause problems (i.e. hiding commits some people might not want to hide), including false positive and thus constant tweaking of the selector.

sindresorhus:
It's nice to remove noise, but not sure if this is a good idea. What if a rogue bot commits something bad to your repo and we hide that commit.

sindresorhus:
> including false positive and thus constant tweaking of the selector.

Don't all bot usernames end with `[bot]`, like `sindresorhus[bot]`?

bfred-it:
> Don't all bot usernames end with `[bot]`, like `sindresorhus[bot]`?

> false positives

https://github.com/Sobot

sindresorhus:
@bfred-it That's just a bot using a normal user account. I don't think we should care about that. It's good enough if it just handles proper `[bot]` accounts.

bfred-it:
> That's just a bot

Where are you seeing that? What I'm saying is that there might be usernames ending in -bot and a blanket `bot$` selector would ghost all those commits.

That, plus what you said in your first comment.

sindresorhus:
> Where are you seeing that?

Look for `renovate[bot]` here: https://github.com/typed-ember/ember-cli-typescript/compare/v2.0.0-beta.3...v2.0.0-rc.1 And `greenkeeper[bot]` here: https://github.com/lukechilds/cacheable-request/commits/master It seems GitHub adds `[bot]` to bot usernames.

bfred-it:
I was referring to https://github.com/Sobot

mike-north:
> There’s a problem with including this feature: it’s a very minor feature and might cause problems (i.e. hiding commits some people might not want to hide), including false positive and thus constant tweaking of the selector.

My thinking was that this filter would default to "off". I definitely agree that starting to filter stuff without telling anyone is a bad idea.

RE: concerns around false positives, there's a knowable and relatively small set of bots that produce this kind of noise

[...]

PR #2447, opened 2019-09-19
Add `dim-bots` feature

fregante:
Closes https://github.com/sindresorhus/refined-github/issues/1676 

# Test

https://github.com/typed-ember/ember-cli-typescript/commits/master
https://github.com/OctoLinker/OctoLinker/commits/master
https://github.com/OctoLinker/OctoLinker/pulls?q=is%3Apr+is%3Aclosed (this list may still be painful to your eyes)

commits

prs


fregante:
prs

source/features/dim-bots.css
```diff
@@ -0,0 +1,51 @@
...
+In PR lists we can hide the meta just by acting on `max-height` + `overflow: hidden`
+In Commit lists however this is not possible because they are tables, so we need to hack around `margin-bottom`
+
+Transitions are used to delay the "hover" status so it's not too annoying when moving the mouse over a list of dimmed items.
+*/
+.rgh-dim-bot.commit .commit-title,
+.rgh-dim-bot.commit .commit-meta,
+.rgh-dim-bot.commit .commit-links-cell summary,
+.rgh-dim-bot.commit .commit-links-cell .btn {
+	transition: 0.1s 0.3s;
```
fregante: Titles are dimmed and meta info/buttons are hidden. This effect is undone with a slight delay on hover/focus, currently set to `0.3s`

source/features/dim-bots.css
```diff
@@ -0,0 +1,51 @@
...
+	transition-property: opacity, margin-bottom;
+}
+
+.rgh-dim-bot.commit:not(.navigation-focus) .commit-title {
+	opacity: 0.2;
```
fregante: Is this value too much?

kidonng: Maybe `0.3` like we did to footer?

sindresorhus:
> Should this feature be disabled for repo owners?

👍

FloEdelmann:
I don't think this feature should be disabled for repo owners. When I look at a commit list or closed PR list in my own repo, I want to focus on the real changes, just like a visitor (or contributor) does.

The only place where it maybe would make sense to disable would be the open PR list, but also there I'd opt for dimming.

sindresorhus:
> I don't think this feature should be disabled for repo owners. When I look at a commit list or closed PR list in my own repo, I want to focus on the real changes, just like a visitor (or contributor) does.

So don't use spammy bots in the first place?

FloEdelmann:
When I don't need spammy bots, I don't need the feature to be disabled in my own repo :wink:

fregante:
That’s irrelevant, when there are no bots, the feature doesn’t apply. The argument is whether to show something to the people who enabled *something.* 

For every optional feature we hide, we always keep showing it to the repo owner, because if they’re don’t want to see it, then they should disable it. The same concept applies here.

Issue #410, opened 2017-05-09
Move the "Close issue" button to the left


Loilo:
It's not like this is a problem you see each and every day, but when it happens, it's kind of annoying: From time to time people fat-finger on the "Close issue" button when they actually meant to submit a comment.

![image](https://cloud.githubusercontent.com/assets/1922624/25845165/3b8dce98-34ad-11e7-81c0-5728ca591013.png)

It would be nice to create a little more space between those buttons, maybe even to completely move the "Close issue" button to the left.

![image](https://cloud.githubusercontent.com/assets/1922624/25845261/91226382-34ad-11e7-872d-bc5c265986f5.png)

What do you think?

sindresorhus:
👍 Has happened to me in the past.

paulmolluzzo:
Yeah, I don't think we need a heavy discussion, this has happened to me a few times. 👍

sindresorhus:
@Loilo Would you be interested in doing a pull request?

Loilo:
Sure. :)

sindresorhus:
@Loilo Make sure you also contact GitHub support about doing this natively ;)

> Our hope is that GitHub will notice and implement some of these much needed improvements. So if you like any of these improvements, please email [GitHub support](mailto:support@github.com) about doing it.

yakov116:
Did this stop working?

loilo:
It did indeed. GitHub seems to have wrapped the button inside a div and overall use Flexbox. Going to attach a PR fixing this.

PR #2449, opened 2019-09-19
Fix `safer-destructive-actions` on comments

loilo:
Re-implements the suggestion from #410 which recently broke since GitHub seems to have changed their markup/CSS a bit.

# Test
Test in any of your own issues, so I can't link any specific one here. 🙂


loilo:
Oh, sure — didn't have that in mind at all. 😅

fregante:
Thank you! :)

loilo:
My pleasure. 🙂

PR #2466, opened 2019-09-28
Restore "Delete branch" button in responsive branch list

### LINKED ISSUE #2452: Restore "Delete branch" button in responsive branch list
1. Visit https://github.com/sindresorhus/refined-github/branches
2. Resize the window to less than 1000px
3. See "Delete branch" icon disappear

This makes it impossible to delete a branch unless you make your window bigger or zoom out.

The trash icon should always be visible. The `New PR` button can be hidden (leave as is). There still are alternative ways to create PRs, but there's no other way to delete a branch.

Button gone

tanmayrajani:
Closes #2452 

(Assuming we don't want to show the delete branch button after it hits the 768px width, if we do, I'll change it accordingly)

### DIFF
--- a/source/content.css
+++ b/source/content.css
@@ -105,23 +105,3 @@
 .dropdown-menu-sw {
 	transform-origin: 90% top;
 }
-
-/* Restore "Delete branch" button in responsive branch list */
-@media (max-width: 1011px) {
-	.js-branch-row .Details-content--shown {
-		display: block !important;
-	}
-	form.js-branch-destroy {
-		display: block !important;
-	}
-}
-
-@media (max-width: 767px) {
-	.js-branch-row .col-12 {
-		width: auto;
-		flex-grow: 1;
-	}
-	:root .js-branch-row .col-12 .css-truncate-target {
-		white-space: initial;
-	}
-}
--- a/source/content.css
+++ b/source/content.css
@@ -111,10 +111,17 @@
 	.js-branch-row .Details-content--shown {
 		display: block !important;
 	}
-	.js-branch-row .col-12 {
-		width: 91.667%;
-	}
 	form.js-branch-destroy {
 		display: block !important;
+	}
+}
+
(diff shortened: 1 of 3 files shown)


fregante:
> Assuming we don't want to show the delete branch button after it hits the 768px width

I'd show it always.

Also `display: block` is enough, they are flex items anyway, so as long as it's not `none` it works.

fregante:
sdfsdf

`flex` causes spacing issues

fregante:
Screenshot 2019-09-28 at 19 14 42

Also the icon is pushed out when it's too small. Not a huge deal but if it's an easy fix it's better to avoid this bug.

tanmayrajani:
> the icon is pushed out when it's too small

For this, they'd only kept enough space for the check. So, I used `width: 91.667%` to make `col-12` as `col-11`. Let me know if there's a better approach I'm missing

fregante:
Around 768px, it stops looking like a table (with columns) so we can drop the `width` altogether, which was only used to keep columns the same width.

We should add something like this instead:

```css
@media (max-width: 767px) { 
	.js-branch-row .col-12 {
		width: auto;
		flex-grow: 1;
	}
	:root .js-branch-row .col-12 .css-truncate-target {
		white-space: initial;
	}
}
```

source/content.css
```diff
@@ -105,3 +105,23 @@
 .dropdown-menu-sw {
 	transform-origin: 90% top;
 }
+
+/* Restore "Delete branch" button in responsive branch list */
+@media (max-width: 1011px) {
+	.js-branch-row .Details-content--shown {
+		display: block !important;
+	}
+	form.js-branch-destroy {
```
fregante: These two selectors can be merged.

Also, as a (probably) last change, this CSS could be moved to its own `always-show-branch-delete-buttons.css` file

tanmayrajani:
Can we merge this if we are good with the changes? :)

PR #2475, opened 2019-10-03
Fix `comments-time-machine-links` feature for pending review comments

### LINKED ISSUE #2187: `comments-time-machine-links` errors on review page with pending reviews
... and cause this incorrect dropdown:

```js
❌ comments-time-machine-links
content.js:2596 TypeError: Cannot read property 'attributes' of null
    at comments_time_machine_links_init (content.js:5907)
    at details.init (content.js:2630)
    at features_run (content.js:2590)
    at content.js:2634
    at content.js:2542
    at onAjaxedPagesRaw (content.js:2537)
```

1. Start review
2. Reload page
3. Notice bug

arnaudvalle:
Closes #2187 

# Test

- Start a review on a PR by adding a comment (--> it's now "pending")
- Reload the `Files changed` page of the PR
- On the comment that was just added, open the dropdown menu (... button)
--> It shouldn't be broken anymore (unlike in that screenshot, below)

image

- Go to the `Conversation` tab of the PR and see the pending comment here as well.
- Again, open the dropdown menu (... button)
--> It shouldn't be broken anymore (unlike in that screenshot, above)

Note: I'm not 100% sure where the `View repo at this time` link is meant to be (or if there's some extra condition to see it) as I couldn't see it myself but I could see the error in my console and my fix seems to be solving that... So, please feel free to edit the test scenario with something a bit more accurate if needs be.

### DIFF
--- a/source/features/comments-time-machine-links.tsx
+++ b/source/features/comments-time-machine-links.tsx
@@ -57,7 +57,7 @@ function init(): void {
 	// PR reviews' main content has nested `.timeline-comment`, but the deepest one doesn't have `relative-time`. These are filtered out with `:not([id^="pullrequestreview"])`
 	const comments = select.all(`
 		:not(.js-new-comment-form):not([id^="pullrequestreview"]) > .timeline-comment:not(.rgh-time-machine-links),
-		.review-comment:not(.rgh-time-machine-links)
+		.review-comment:not(.rgh-time-machine-links) > .previewable-edit:not(.is-pending)
 	`);
 
 	for (const comment of comments) {
--- a/source/features/comments-time-machine-links.tsx
+++ b/source/features/comments-time-machine-links.tsx
@@ -61,15 +61,11 @@ function init(): void {
 	`);
 
 	for (const comment of comments) {
-		const timestampEl = select('relative-time', comment)!;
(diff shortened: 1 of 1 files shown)


fregante:
Thank you for the PR

This only hides the error though, it doesn’t resolve the underlying issue. The feature should keep working there too

arnaudvalle:
Thanks for clearing that up. So what should the expected behaviour be? Would a a link to the `HEAD` of the repo for those pending comments (since there's no `<relative-time>` or any other timestamp I don't think in the HTML of that comment block) be ok?

fregante:
Ahhh my bad. They’re pending comments. The fix is correct.

fregante:
Is it possible to exclude pending review comments in the main `select.all` instead of adding an `if` inside the loop?

arnaudvalle:
Would make sense, yes. I'll have a look to see if there's a decent css selector to do this.

arnaudvalle:
I think I got this to work and I have reverted my initial fix (the `if` in the `for` loop).
I had to change the `.review-comment` selector to target an element further down the DOM tree - `.previewable-edit` - which is the element that has the `is-pending` class as well. Hopefully that doesn't introduce other issues elsewhere...

PR #2478, opened 2019-10-05
Update personal token link based off domain selected

### LINKED ISSUE #2408: per domain setting personal Token Found Here  Link is not getting updated according to selected domain
image

as seen on bottom right corner of image link still points to github.com instead of selected domain

busches:
This updates the `personal token` link in the settings to use the domain from the dropdown to construct the link. When there is no link, it defaults to GitHub. We now show the protocol in the domain dropdown, otherwise I had to infer it for the link.

image

Closes #2408

# Test

I tested it on GH, a random site I gave permission to, and then a GHE instance (2.17) to confirm the settings URL was the same as GH, it is.

### DIFF
--- a/source/options-storage.ts
+++ b/source/options-storage.ts
@@ -56,12 +56,12 @@ export default getOptions(location.protocol.startsWith('http') ? location.host :
 
 export async function getAllOptions(): Promise<Map<string, OptionsSync<RGHOptions>>> {
 	const optionsByDomain = new Map<string, OptionsSync<RGHOptions>>();
-	optionsByDomain.set('github.com', getOptions('github.com'));
+	optionsByDomain.set('https://github.com', getOptions('github.com'));
 
 	const {origins} = await getAdditionalPermissions();
 	for (const origin of origins) {
-		const {host} = new URL(origin);
-		optionsByDomain.set(host, getOptions(host));
+		const {protocol, host} = new URL(origin);
+		optionsByDomain.set(`${protocol}//${host}`, getOptions(host));
 	}
 
 	return optionsByDomain;
--- a/source/options-storage.ts
+++ b/source/options-storage.ts
@@ -56,12 +56,12 @@ export default getOptions(location.protocol.startsWith('http') ? location.host :
 
 export async function getAllOptions(): Promise<Map<string, OptionsSync<RGHOptions>>> {
 	const optionsByDomain = new Map<string, OptionsSync<RGHOptions>>();
-	optionsByDomain.set('https://github.com', getOptions('github.com'));
+	optionsByDomain.set('github.com', getOptions('github.com'));
 
 	const {origins} = await getAdditionalPermissions();
 	for (const origin of origins) {
-		const {protocol, host} = new URL(origin);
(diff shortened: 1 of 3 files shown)


source/options.tsx
```diff
@@ -66,14 +66,16 @@ async function init(): Promise<void> {
...
+		dropdown.addEventListener('change', (event) => {
 			for (const [domain, options] of optionsByDomain) {
 				if (dropdown.value === domain) {
 					options.syncForm(form);
 				} else {
 					options.stopSyncForm();
 				}
 			}
+			const newDomain = (event.target as HTMLInputElement).value;
+			select('#personal-token-link')!.setAttribute('href', `${newDomain}/settings/tokens/new?description=Refined%20GitHub&scopes=repo`);
```
fregante: Can you just change the `.host` property as I had suggested instead? This avoids having to repeat the URL here and in the HTML

source/options-storage.ts
```diff
@@ -56,12 +56,12 @@ export default getOptions(location.protocol.startsWith('http') ? location.host :
...
 	const optionsByDomain = new Map<string, OptionsSync<RGHOptions>>();
-	optionsByDomain.set('github.com', getOptions('github.com'));
+	optionsByDomain.set('https://github.com', getOptions('github.com'));
 
 	const {origins} = await getAdditionalPermissions();
 	for (const origin of origins) {
-		const {host} = new URL(origin);
-		optionsByDomain.set(host, getOptions(host));
+		const {protocol, host} = new URL(origin);
+		optionsByDomain.set(`${protocol}//${host}`, getOptions(host));
```
fregante: We don’t need to consider protocol changes. GitHub should only be available on https

source/options.tsx
```diff
@@ -66,14 +66,16 @@ async function init(): Promise<void> {
 			</select>
 		) as unknown as HTMLSelectElement;
 		form.before(<p>Domain selector: {dropdown}</p>, <hr/>);
-		dropdown.addEventListener('change', () => {
+		dropdown.addEventListener('change', (event) => {
```
fregante: Is the linter on GitHub Actions not working anymore? I think this should be without parens

busches: CI isn't running on forks. I'll double check it.

fregante: Ah yes, it’s only set to run on push but not on PRs.

Can you make that change too? And then it can be merged if it runs on this PR

source/options.tsx
busches: I did not realize you could swap it like that, TIL. :)

source/options.tsx
```diff
@@ -66,14 +66,17 @@ async function init(): Promise<void> {
...
 			for (const [domain, options] of optionsByDomain) {
 				if (dropdown.value === domain) {
 					options.syncForm(form);
 				} else {
 					options.stopSyncForm();
 				}
 			}
+
+			const newHost = (event.target as HTMLInputElement).value;
+			(select('#personal-token-link') as HTMLAnchorElement)!.host = newHost;
```
fregante: You can move the type to the generic parameter: `select<HTMLAnchorElement>(…)`

PR #2484, opened 2019-10-09
Add draft PR filter in `pr-filters`

### LINKED ISSUE #1900: Filter by "Draft PRs only" in PR list

FloEdelmann:
Closes #1900. As noted there, I extended the existing `filter-pr-by-build-status` feature and renamed it to `pr-filters`.

It uses the undocumented `draft:true`/`draft:false` filters because they actually work as intended (opposed to `is:draft`/`-is:draft`). They have been working long enough now that I think we can assume they will stay.

Screenshot_2019-10-09_23-10-14


source/features/pr-filters.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+	const search = new URLSearchParams({
+		q: query + (isSelected ? '' : ` ${filterQuery}`)
+	});
+
+	dropdown.append(
+		<a
+			href={`?${search}`}
+			className={`select-menu-item ${isSelected ? 'selected' : ''}`}
+			aria-checked={isSelected}
```
fregante: This should be stringified first. `attr={false}` means “no attr” in JSX

source/features/pr-filters.tsx
```diff
@@ -0,0 +1,101 @@
...
+	statusFilter.addEventListener('toggle', () => {
+		for (const status of ['Success', 'Failure', 'Pending']) {
+			addDropdownItem(dropdown, status, 'status', status.toLowerCase());
+		}
+	}, {once: true});
+	reviewsFilter.after(statusFilter);
+}
+
+async function init(): Promise<void | false> {
+	const reviewsFilter = select('.table-list-header-toggle > details:nth-last-child(3)')!;
```
fregante: This exclamation mark is unnecessary

source/features/pr-filters.tsx
```diff
@@ -0,0 +1,101 @@
...
+			role="menuitemradio"
+		>
+			{checkInline()}
+			<div className="select-menu-item-text">{title}</div>
+		</a>
+	);
+}
+
+function addDraftFilter(reviewsFilter: HTMLElement): void {
+	reviewsFilter.addEventListener('toggle', () => {
```
fregante: Can you move `addEventListener` to `init` then the listener function can stay here in the top level.

source/features/pr-filters.tsx
```diff
@@ -0,0 +1,101 @@
...
+		const dropdown = select('.select-menu-list', reviewsFilter)!;
+
+		dropdown.append(
+			<div
+				className="SelectMenu-divider"
+			>
```
fregante: This **opening** tag can be on one line

source/features/pr-filters.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+	select('summary', statusFilter)!.textContent = 'Status\u00A0';
+	select('.select-menu-title', statusFilter)!.textContent = 'Filter by build status';
+	dropdown.textContent = ''; // Drop previous filters
+
+	statusFilter.addEventListener('toggle', () => {
+		for (const status of ['Success', 'Failure', 'Pending']) {
+			addDropdownItem(dropdown, status, 'status', status.toLowerCase());
+		}
+	}, {once: true});
```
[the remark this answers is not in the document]
FloEdelmann: The toggle listener for the status dropdown already existed before my refactoring. Should I remove the toggle listener only for the status dropdown, only for the review dropdown, or both?

FloEdelmann:
@fregante Is there anything left that prevents this from being merged?

source/features/pr-filters.tsx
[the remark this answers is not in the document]
fregante: Only *in this case*, since we’re already adding DOM for it. The previous function did nothing at all until the menu was opened.

fregante:
# 💯

PR #2486, opened 2019-10-10
Add `separate-draft-pr-button` feature

### LINKED ISSUE #2473: Separate "Create Draft PR" from "Create PR" dropdown
A PR created as a "full" (non-draft) PR is impossible to convert to a draft PR.

Half the PRs I make, meaning for them to be a draft, end up accidentally created as a full PR just because I hit the button rather than the drop-down. That requires deleting that PR and creating a new one - this time as a draft.

Judging by [this thread on the Github community forum](https://github.community/t5/How-to-use-Git-and-GitHub/Feature-Request-Switch-from-ready-to-draft-in-pull-requests/td-p/19107), I'm not the only one.

I would suggest for refined-github to pull the "Create Draft PR" button from the dropdown to instead be besides the main "Create PR" button.

Mockup:
2019-10-02 at 17 33

FloEdelmann:
Closes #2473.

separate-draft-pr-button

~The screenshot should probably be a "before and after" comparison.~ Also, I didn't include an option to make drafts the primary action because I first wanted to gather some feedback.

By the way: This PR was submitted via this feature :wink:


readme.md
```diff
@@ -228,6 +228,7 @@ Thanks for contributing! 🦋🙌
 - [](# "update-pr-from-base-branch") [Adds button to update a PR from the base branch to ensure it builds correctly before merging the PR itself.](https://user-images.githubusercontent.com/1402241/57941992-f2170080-7902-11e9-8f8a-594aad983559.png) GitHub only adds it when the base branch is "protected".
 - [](# "quick-review-buttons") [Simplifies the PR review form: Approve or reject reviews faster with one-click review-type buttons.](https://user-images.githubusercontent.com/1402241/34326942-529cb7c0-e8f3-11e7-9bee-98b667e18a90.png)
 - [](# "warn-pr-from-master") [Warns you when creating a pull request from the default branch, as it’s an anti-pattern.](https://user-images.githubusercontent.com/1402241/52543516-3ca94e00-2de5-11e9-9f80-ff8f9fe8bdc4.png)
+- [](# "separate-draft-pr-button") [Separates `Create draft pull request` and `Create pull request` buttons.](https://user-images.githubusercontent.com/202916/66609826-4a081980-ebba-11e9-80d8-cdef503a8a80.png)
```
busches: I think this should go up one, instead of splitting the two that start with Warns.

fregante: They’re not in a particular order. Let’s just spend the features to each list instead of adding them in the middle.

source/features/separate-draft-pr-button.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+		const classList = ['btn', 'ml-2', 'tooltipped', 'tooltipped-s'];
+		if (!title.toLowerCase().includes('draft')) {
+			classList.push('btn-primary');
+		}
+
+		const submitForm = (): void => {
+			radioButton.checked = true;
+			previewForm.submit();
+		};
```
fregante: I don’t think all of this is necessary. Probably you can set the `name` and `value` to the button and that will be submitted instead of the radio. 

Also generally use more descriptive variable names because “menuItems” doesn’t tell me what they are.

source/features/separate-draft-pr-button.tsx
```diff
@@ -0,0 +1,58 @@
...
+import features from '../libs/features';
+
+function init(): void {
+	const previewForm = select('.new-pr-form')! as HTMLFormElement;
+
+	// Move the "Allow edits from maintainers" to its own line if it exists
+	const allowEditsCheckbox = select('.float-left', previewForm);
+	if (allowEditsCheckbox) {
+		allowEditsCheckbox.classList.remove('float-left');
+	}
```
fregante: If you shorten the label to just “create draft” you probably won’t have to mess with this checkbox at all.

readme.md
busches: I get that they're not in the order, but it seems wrong to disrupt the flow that's already there. Appending will solve this too.

busches:
Works fine on GHE 2.17 :)

source/features/separate-draft-pr-button.tsx
```diff
@@ -0,0 +1,58 @@
...
+		);
+	}
+}
+
+features.add({
+	id: __featureName__,
+	description: 'Separates `Create draft pull request` and `Create pull request` buttons.',
+	screenshot: 'https://user-images.githubusercontent.com/202916/66609826-4a081980-ebba-11e9-80d8-cdef503a8a80.png',
+	include: [
+		features.isCompare
```
busches: This is too wide and errors on other comparison pages. e.g. https://github.com/sindresorhus/refined-github/compare/19.9.27...19.10.10 It looks like they add a class of `is-pr-composer-expanded` to the body we could use to detect if you're composing a PR.

FloEdelmann:
@fregante @busches Thanks for your quick reviews! I don't have much today but I'll work on this tomorrow.

source/features/separate-draft-pr-button.tsx
fregante: You can compose from any compare page, right?

busches: Not to my knowledge, see the above example. Comparing branches with a PR already also didn't work and didn't  have that class.

source/features/separate-draft-pr-button.tsx
FloEdelmann: True, but I think it might then be confusing for users not familiar with draft pull requests that they actually *are* pull requests:

Unbenannt

Can we assume that all users of Refined GitHub already know how draft PRs work?

Also, we then deviate from the official GitHub wording (meaning that we also have to implement a special case in the button adding code below).

source/features/separate-draft-pr-button.tsx
FloEdelmann: I just added a check if the PR compose form exists: https://github.com/sindresorhus/refined-github/pull/2486/commits/0d78e59881b9dda064044476eecad4ced264c578

source/features/separate-draft-pr-button.tsx
fregante: The H1 on the page is “Open a pull request” so it would contextually make sense. 

If we want to keep the name in the button it could be a shorter “Create draft PR”

source/features/separate-draft-pr-button.tsx
```diff
@@ -0,0 +1,59 @@
...
+	const createPrButtonGroup = select('.BtnGroup', buttonBar)!;
+	const createPrDropdownItems = select.all('.select-menu-item', createPrButtonGroup);
+
+	for (const dropdownItem of createPrDropdownItems) {
+		const title = select('.select-menu-item-heading', dropdownItem)!.textContent!.trim();
+		const description = select('.description', dropdownItem)!.textContent!.trim();
+		const radioButton = select('[type=radio]', dropdownItem)! as HTMLInputElement;
+
+		const classList = ['btn', 'ml-2', 'tooltipped', 'tooltipped-s'];
+		if (!title.toLowerCase().includes('draft')) {
```
fregante: Can you use a `regex.test()` for this?

source/features/separate-draft-pr-button.tsx
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	const buttonBar = select('.flex-justify-end', previewForm)!;
+	const createPrButtonGroup = select('.BtnGroup', buttonBar)!;
+	const createPrDropdownItems = select.all('.select-menu-item', createPrButtonGroup);
+
+	for (const dropdownItem of createPrDropdownItems) {
+		const title = select('.select-menu-item-heading', dropdownItem)!.textContent!.trim();
+		const description = select('.description', dropdownItem)!.textContent!.trim();
+		const radioButton = select('[type=radio]', dropdownItem)! as HTMLInputElement;
```
fregante: You can use `select<…type…>(…)`

source/features/separate-draft-pr-button.tsx
```diff
@@ -0,0 +1,59 @@
...
+			</button>
+		);
+
+		createPrButtonGroup.remove();
+	}
+}
+
+features.add({
+	id: __featureName__,
+	description: 'Separates `Create draft pull request` and `Create pull request` buttons.',
```
fregante: How about

> Lets you create draft pull requests in one click.

source/features/separate-draft-pr-button.tsx
```diff
@@ -0,0 +1,59 @@
...
+		return;
+	}
+
+	// Move the "Allow edits from maintainers" to its own line if it exists
+	const allowEditsCheckbox = select('.float-left', previewForm);
+	if (allowEditsCheckbox) {
+		allowEditsCheckbox.classList.remove('float-left');
+	}
+
+	const buttonBar = select('.flex-justify-end', previewForm)!;
```
fregante: This selector is a little generic, is there anything more specific?

source/features/separate-draft-pr-button.tsx
```diff
@@ -0,0 +1,55 @@
+import React from 'dom-chef';
+import select from 'select-dom';
+import features from '../libs/features';
+
+function init(): void {
+	const previewForm = select<HTMLFormElement>('.new-pr-form');
```
fregante: I think that the type is not necessary here

source/features/separate-draft-pr-button.tsx
```diff
@@ -0,0 +1,55 @@
...
+	if (!previewForm) {
+		return;
+	}
+
+	const buttonBar = select('.timeline-comment > :last-child', previewForm)!;
+	const createPrButtonGroup = select('.BtnGroup', buttonBar)!;
+	const createPrDropdownItems = select.all('.select-menu-item', createPrButtonGroup);
+
+	for (const dropdownItem of createPrDropdownItems) {
+		let title = select('.select-menu-item-heading', dropdownItem)!.textContent!.trim();
```
fregante: I think you can skip the `select` altogether and just set the `title` in the following condition block

FloEdelmann: I can't. The regex in the condition is tested against this title.

fregante: D’oh

fregante:
Still a WIP?

FloEdelmann:
I am done with the code, if you don't have any more remarks.

As mentioned in the initial PR description, the screenshot should be a "before and after comparison" (and needs to be updated anyway with the new wording and placement). @fregante Could you do that to match the visual style of the other screenshots?

fregante:
You can copy the before/after from existing comparisons, that’s how all screenshots are done since Sindre’s original screenshot 😅

FloEdelmann:
@fregante Is there anything left that prevents this from being merged?

busches:
Looks good, if @fregante doesn't say anything in a day, I'll merge it in. Thanks for this, I wanted to use it yesterday. :)

FloEdelmann:
Thanks for merging! :)

paulvanliew:
How is this usable in Github Enterprise Cloud? Since the ability to make draft PRs default configurable was nixed by people, many of us have to feel the pain (and watch CI systems constant build junk commits because they're not draft PRs, and so on). At least showing the button would be a bad, but at least something compromise for those of us not self-hosting for users.

PR #2488, opened 2019-10-11
Trigger test workflow on PRs

busches:
Trigger workflows on PRs in addition to push. @fregante do we need push still?

[e] Looks like this does work on forks and draft PRs.


fregante:
Can you put it on two lines?

And then it can be merged

PR #2499, opened 2019-10-23
Fix issue links in options

### LINKED ISSUE #2460: settings: link to issues are broken
where:

- open settings
- hoover over issue link

what i see:
- a link `chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/options.html///issues/2041`

what i expected:
- a link to `https://github.com/sindresorhus/refined-github/issues/2041`

screenshot:

see link in left bottom statusline:

Bildschirmfoto 2019-09-26 um 23 52 49

dertieran:
Fixes #2460

I decided that the easiest way to fix this was by allowing overwriting the options for `linkify-issues`.

### DIFF
--- a/source/libs/dom-formatters.ts
+++ b/source/libs/dom-formatters.ts
@@ -13,15 +13,17 @@ export const linkifiedURLClass = 'rgh-linkified-code';
 // https://github.com/sindresorhus/refined-github/issues/1305
 const currentRepo = getOwnerAndRepo();
 
-export function linkifyIssues(element: Element): void {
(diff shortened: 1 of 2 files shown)


fregante:
The previous version just had those as defaults. For example instead of `/` it should be sindresorhus, then refined-github, then the same host check we have for the options, something like `if location.protocol === https, baseUrl = location.origin, else github.com

dertieran:
As mentioned in the issue, this still wouldn't work because `currentRepo.ownerName === 'options.html'`.

It's also possible to handle this instead, just thought making this more configurable would make handling edgecases like this easier.

sindresorhus:
Thanks for fixing :)

PR #2505, opened 2019-10-27
Add `html-preview-link` feature

### LINKED ISSUE #1096: Add button to preview HTML files

kidonng:
Closes #1096 

# Test

https://github.com/CodingTrain/website/blob/4f90eedb9618257d9166241e92e51a7f3f00a08e/code_challenges/PerlinNoiseTerrain_p5.js/index.html

### DIFF
--- a/source/features/html-preview-link.tsx
+++ b/source/features/html-preview-link.tsx
@@ -6,15 +6,15 @@ import {groupSiblings} from '../libs/group-buttons';
 function init(): void {
 	const rawButton = select('#raw-url') as HTMLAnchorElement;
 	if (rawButton && /\.html?/.test(rawButton.pathname)) {
-		const link = rawButton.pathname.split('/')
+		const link = rawButton.pathname.split('/');
 		// Remove /raw/
-		link.splice(3, 1)
+		link.splice(3, 1);
 		rawButton
 			.parentElement! // `BtnGroup`
 			.prepend(
 				<a
 					className="btn btn-sm BtnGroup-item"
-				  href={`https://cdn.statically.io/gh${link.join('/')}`}>
+					href={`https://cdn.statically.io/gh${link.join('/')}`}>
 					Preview
 				</a>
 			);
@@ -34,4 +34,4 @@ features.add({
 	],
 	load: features.onAjaxedPages,
 	init
-})
+});
--- a/source/features/html-preview-link.tsx
+++ b/source/features/html-preview-link.tsx
@@ -1,25 +1,22 @@
 import React from 'dom-chef';
 import select from 'select-dom';
 import features from '../libs/features';
-import {groupSiblings} from '../libs/group-buttons';
+import {isSingleHTMLFile} from '../libs/utils';
 
 function init(): void {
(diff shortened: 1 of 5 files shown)


source/features/html-preview-link.tsx
```diff
@@ -0,0 +1,37 @@
+import React from 'dom-chef';
+import select from 'select-dom';
+import features from '../libs/features';
+import {groupSiblings} from '../libs/group-buttons';
+
+function init(): void {
+	const rawButton = select('#raw-url') as HTMLAnchorElement;
+	if (rawButton && /\.html?/.test(rawButton.pathname)) {
```
fregante: I think you can move this to the feature’s `include`, by creating a `isHTMLSingleFile` test function (not in the current file, not `page-detect.ts`

source/features/html-preview-link.tsx
```diff
@@ -0,0 +1,37 @@
...
+		rawButton
+			.parentElement! // `BtnGroup`
+			.prepend(
+				<a
+					className="btn btn-sm BtnGroup-item"
+					href={`https://cdn.statically.io/gh${link.join('/')}`}>
+					Preview
+				</a>
+			);
+		groupSiblings(rawButton);
```
fregante: This probably isn’t necessary in this feature. Maybe it’s not necessary even in the feature you copied it from, but I don’t remember.

source/features/html-preview-link.tsx
```diff
@@ -0,0 +1,37 @@
+import React from 'dom-chef';
+import select from 'select-dom';
+import features from '../libs/features';
+import {groupSiblings} from '../libs/group-buttons';
+
+function init(): void {
+	const rawButton = select('#raw-url') as HTMLAnchorElement;
```
fregante: You can use `select<HTMLAnchorElement>(...)!;

kidonng: 😅 `rawButton.pathname` is used below where `pathname` is from `HTMLAnchorElement`

source/features/html-preview-link.tsx
kidonng: You mean `groupSiblings`? It's from `copy-file` feature. Will check whether it is needed later 😊

source/features/html-preview-link.tsx
kidonng: > I’m gonna guess that if (rawbutton) isn’t necessary in that case.

Yeah I doubted that when I first read it but didn't check, come to think it now, there will always be a raw button for files.

kidonng: > I think you can move the regex test to the feature’s include, by creating a isSingleHTMLFile test function (not in the current file, not page-detect.ts

😅 Honestly, I don't know where to put it then, so I put it into `utils.ts`.

source/features/html-preview-link.tsx
kidonng: You're right #2506

fregante:
Thank you! Finally! That issue was opened almost 2 years ago

PR #2507, opened 2019-10-28
Support dark mode in the options view

sindresorhus:
Screenshot 2019-10-28 at 15 36 44

### DIFF
--- a/source/options.css
+++ b/source/options.css
@@ -6,8 +6,36 @@ html {
 }
 
 :root {
+	--background-color: #fff;
+	--text-color: #202124;
+	--secondary-text-color: #5f6368;
+	--link-color: #3367d6;
+	--input-border-color: rgba(0, 0, 0, 0.2);
 	--github-green: #28a745;
 	--github-red: #cb2431;
+}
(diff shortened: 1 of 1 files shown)


fregante:
What about Firefox? Does it break anything there?

sindresorhus:
Looks fine to me:

Screenshot 2019-10-29 at 10 20 09

fregante:
Perfect!

fregante:
I noticed (in my other extension) that `--background-color: white` breaks Firefox’s regular background (it appears as a white border around the gray background)



We can probably drop the variables from `:root` and the lack of it will make the browser use the default values, essentially _skipping_ all of the new rules on non-dark browsers (ideal, Firefox’ default styles are better than this)

fregante:
Never mind. I can't reproduce consistently.

debajyoti-xyz:
On windows 10 with dark mode enabled the text color of feature filter input are a bit hard to read.

image

fregante:
@debajyoti-xyz can you open a new issue?

PR #2510, opened 2019-11-01
Fix vulnerabilities for https-proxy-agent

sotayamashita:


fregante:
Thank you, but this is a sub dependency of part of a devDependency that we don't use (`addon-linter`) so I'm not worried about it. Let's just wait for `web-ext` to update.

PR #2511, opened 2019-11-02
Add explanatory tooltips to `global-discussion-list-filters`

kevinjalbert:
This change adds some accessibility to the filter links using the aria-label attribute for the filter link.

Adding the title attribute will show more of an explanation if you mouse hover over (which GitHub has for some of the other filter links).

01-09-skqpo-4rpi4

# Test

This change adds the extra attributes on the added 'Commented' and 'Yours' filters in the following pages: 

- https://github.com/pulls
- https://github.com/issues


kevinjalbert:
@fregante 👍 on the changes you've added.

PR #2512, opened 2019-11-02
Tests all URLs in `page-detect` against each other

fregante:
- Enforces tests on all new page detections
- Keeps expected URLs near the function, as examples
- Tests each function with ALL the available URLs in the file, so any new URL added is an automatic _new test_ for all the rest.

They'll be tree-shaken by webpack.

### DIFF
--- a/test/page-detect.ts
+++ b/test/page-detect.ts
@@ -1,21 +1,53 @@
-import test, {ExecutionContext} from 'ava';
+import test from 'ava';
 import './fixtures/globals';
 import * as pageDetect from '../source/libs/page-detect';
 
-function urlMatcherMacro(
-	t: ExecutionContext,
-	detectFn: (url: string) => boolean,
-	shouldMatch: string[] = [],
-	shouldNotMatch: string[] = []
-): void {
-	for (const url of shouldMatch) {
-		location.href = url;
-		t.true(detectFn(url));
+const allUrls = new Set<string>();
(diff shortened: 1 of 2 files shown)


test/page-detect.ts
```diff
@@ -1,21 +1,53 @@
...
+for (const imported of Object.values(pageDetect)) {
+	if (Array.isArray(imported)) {
+		for (const url of imported) {
+			allUrls.add(url);
+		}
+	}
+}
+
+for (const [key, detect] of Object.entries(pageDetect)) {
+	if (key.endsWith('Test') || typeof detect !== 'function') {
```
jerone:
```suggestion
	if (key.endsWith('Test') || key.endsWith('TestSkipNegatives') || typeof detect !== 'function') {
```

fregante: It already matches `typeof detect !== 'function'` because in this case `detect` will be `true`.

If this wasn't the case, it would error later anyway with:

```
The function `isRepoTestSkipNegatives` doesn’t have any tests. Export an array of valid URLs as `isRepoTestSkipNegativesTest`
```

jerone: Ah, yes. Ignore my suggestion.

jerone:
This will increase the deployed extension with ±250 loc.

fregante:
@jerone yeah it does feel dirty but I was annoyed that some functions were added without tests. This is now no longer possible

PR #2513, opened 2019-11-02
Add `linkify-labels-on-dashboard` feature

### LINKED ISSUE #2509: Linkable labels in dashboard
https://github.com/dashboard

There is 'Recent Activity' list in the dashboard.
Unfortunately labels are not linkable.
Expected behavior: Click on label opens issues filtered by label.

Workaround: open the issue and then click on the label in the issue.

dertieran:
Just ran into this issue a few days back when I wanted to click on one of the labels and nothing happened. :smile: 

I included the `sort:updated-desc` query from the `sort-issues-by-update-time` feature.
Not sure if this should be included here or if there is another way for both to work together.

Also I wasn't sure what the cleanest way is to detect the activities being loaded, so I copied this part from `infinite-scroll`. 

Closes #2509


source/features/add-dashboard-label-links.tsx
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	const repository = select('a[data-hovercard-type="repository"]', activity) as HTMLAnchorElement;
+	if (!repository) {
+		return;
+	}
+
+	for (const label of labels) {
+		const search = new URLSearchParams();
+		const text = (label.textContent || '').trim();
```
fregante:
```suggestion
		const text = label.textContent!.trim();
```

source/features/add-dashboard-label-links.tsx
```diff
@@ -0,0 +1,49 @@
...
+import {wrap} from '../libs/dom-utils';
+import observeEl from '../libs/simplified-element-observer';
+
+function linkifyLabels(activity: HTMLElement): void {
+	const labels = select.all('.IssueLabel', activity);
+	if (labels.length === 0) {
+		return;
+	}
+
+	const repository = select('a[data-hovercard-type="repository"]', activity) as HTMLAnchorElement;
```
fregante:
```suggestion
	const repository = select<HTMLAnchorElement>('a[data-hovercard-type="repository"]', activity)!;
```

source/features/add-dashboard-label-links.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+	for (const label of labels) {
+		const search = new URLSearchParams();
+		const text = (label.textContent || '').trim();
+		search.set('q', `is:issue is:open sort:updated-desc label:"${text}"`);
+		wrap(label, <a href={`${repository.href}/issues?${search}`} />);
+	}
+}
+
+function findActivities(): void {
```
fregante: `findActivities` receives the `observer` as an argument; you can use it to disable it altogether instead of using `rgh-linkified-labels`:

```js
observer.disconnect();
```

source/features/add-dashboard-label-links.tsx
```diff
@@ -0,0 +1,49 @@
...
+import observeEl from '../libs/simplified-element-observer';
+
+function linkifyLabels(activity: HTMLElement): void {
+	const labels = select.all('.IssueLabel', activity);
+	if (labels.length === 0) {
+		return;
+	}
+
+	const repository = select('a[data-hovercard-type="repository"]', activity) as HTMLAnchorElement;
+	if (!repository) {
```
fregante: Are you sure this check is necessary? Maybe you just need a `!` after `select()` to shut TS up. I expect the repo link to always be there

source/features/add-dashboard-label-links.tsx
```diff
@@ -0,0 +1,49 @@
+import React from 'dom-chef';
+import select from 'select-dom';
+import features from '../libs/features';
+import {wrap} from '../libs/dom-utils';
+import observeEl from '../libs/simplified-element-observer';
+
+function linkifyLabels(activity: HTMLElement): void {
+	const labels = select.all('.IssueLabel', activity);
+	if (labels.length === 0) {
```
fregante: This check isn't necessary, the loop just won't run if length is 0. I'd rather see

```ts
 	for (const label of select.all('.IssueLabel', activity)) {
```

source/features/add-dashboard-label-links.tsx
dertieran: No I also think it should always be there, mainly did it to handle it if it should ever happen.
But I guess it's better to error out then instead of failing silently.

source/features/add-dashboard-label-links.tsx
dertieran: Mainly did this to avoid querying for the `repository` if it isn't needed.
Guess it's not to important, so I can also change it if you like that style better.

fregante:
Can you take a screenshot of that box and add a simple arrow to a label? This would make it clear

dertieran:
Thanks for the improvements :+1:

Just noticed that I missed handling PR labels :smile:
Fixing this right now and then I will add the screenshot.

sindresorhus:
Can you add the feature to the readme?

dertieran:
@sindresorhus will do, I'm just not on a PC till tomorrow (~9h)

dertieran:
@fregante I don't feel too strongly about that, so I'm also okay changing it.

But I still think handling issue and pr activities is the better default.
Right now it has the same behavior as clicking on a label in the issues and pr tab. 
*(If the default query wasn't changed)* 

I also think it is clear to the user whey they sees only issues or prs, because they clicked on the respective activity.
They can then still change the query and remove the `is:pr` part.

fregante:
@sindresorhus thoughts?

sindresorhus:
> But I still think handling issue and pr activities is the better default.

👍

sindresorhus:
I noticed a bug. It doesn't linkify labels shown when you click "Show more".

fregante:
> I noticed a bug. It doesn't linkify labels shown when you click "Show more".

If that's the case, only the selector needs to be tweaked because all the activity is loaded at once. This is the fragment: https://github.com/dashboard/recent-activity

dertieran:
@sindresorhus Good catch it seems I don't have enough activities :smile:

Changed the selector to a simple `li`, they would also share a `.d-flex` but I think that is also very generic.

fregante:
Thank you @dertieran! 🗿

PR #2514, opened 2019-11-03
Meta: Add feature option search

### LINKED ISSUE #2147: Filter-as-you-type searchbox in Options features list
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/refined-github/issues/2147)

It's great that you have recently added the ability to enable/disable features with a click in the extension's Options.

The features list of checkboxes however is a quite large one. 
How about adding a **filter-as-you-type searchbox**, to only display/filter the matching ones as you type? (hoping that it's possible of course)

PS. Using Refined Github 19.6.9 in Chrome 75/Firefox 67 in win10.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $5.00)

### Tips

---
</details>

Mottie:
## Description

This PR adds a feature option search to Refined GitHub's option popup.

Searching for `open batch` will only return one result ("batch-open-issues"); the order of the words doesn't matter. Under the hood, this makes sure that every word in the search pattern matches some part of the feature name and description.

Notes:

- This only searches the feature name & description.
- All matches are case-insensitive.
- The following characters are ignored: <code>,.:;'"’`</code>.
- The following character is replaced with a space: `-`; this allows `"open-batch` to still match "batch-open-issues"

## Screenshots

search

## Closes
Closes #2147

## Test

Open Refined-GitHub's options

### DIFF
--- a/source/options.css
+++ b/source/options.css
@@ -38,11 +38,6 @@ button {
 	border: 1px solid var(--input-border-color);
 }
 
-input:not([type='checkbox']):not(:placeholder-shown):valid {
-	color: var(--github-green) !important;
-	border: 1px solid;
-}
-
 input:not([type='checkbox']):invalid {
 	color: var(--github-red) !important;
 	border: 1px solid;
--- a/source/options.css
+++ b/source/options.css
@@ -110,7 +110,8 @@ h2 ~ h2 {
 	text-decoration: line-through;
 }
 
-.js-features :not(:checked) + .info .extended-options {
+.js-features :not(:checked) + .info .extended-options,
+.feature[hidden] {
 	display: none;
 }
 
--- a/source/options.css
+++ b/source/options.css
@@ -110,8 +110,7 @@ h2 ~ h2 {
 	text-decoration: line-through;
 }
 
-.js-features :not(:checked) + .info .extended-options,
-.feature[hidden] {
+.js-features :not(:checked) + .info .extended-options {
 	display: none;
 }
 
--- a/source/options.css
+++ b/source/options.css
@@ -95,7 +95,7 @@ h2 ~ h2 {
 	margin-top: 2em;
 }
 
-.js-features .feature {
+.js-features .feature:not([hidden]) {
 	display: flex;
 	margin-top: 1em;
 }
@@ -110,8 +110,7 @@ h2 ~ h2 {
 	text-decoration: line-through;
 }
 
-.js-features :not(:checked) + .info .extended-options,
-.feature[hidden] {
+.js-features :not(:checked) + .info .extended-options {
 	display: none;
 }
 
--- a/source/options.tsx
+++ b/source/options.tsx
@@ -6,6 +6,7 @@ import {applyToLink} from 'shorten-repo-url';
 import indentTextarea from 'indent-textarea';
 import {getAllOptions} from './options-storage';
 import * as domFormatters from './libs/dom-formatters';
+import debounce from 'debounce-fn';
 
 function parseDescription(description: string): DocumentFragment {
 	const descriptionElement = <span>{description}</span>;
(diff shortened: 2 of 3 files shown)


source/options.tsx
```diff
@@ -48,6 +49,53 @@ function buildFeatureCheckbox({name, description, screenshot, disabled}: Feature
...
+		matchAll ?
+			pattern.every((str: string) => text.includes(str)) :
+			pattern.some((str: string) => text.includes(str))
+	);
+
+	const searchHandler = (event: Event): void => {
+		const pattern = (event.target as HTMLInputElement).value || '';
+		const patternArray = replaceCharacters(pattern)
+			.split(separators)
+			.filter(s => s); // Remove empty strings
```
yakov116:
```suggestion
			.filter(Boolean); // Remove empty strings
```

fregante:
Great to see this!

I don't think one ever needs the "non-exact" search, which seems to me it's just an "keyword OR keyword" search. If I enter more words, I want all of them to appear.

For the implementation, have a look at this, it's shorter:

https://github.com/HangYang/one-click-extensions-manager/blob/1ac4566f7d28329c98ce8f0edc7a380a8ce9ecf7/src/popup.js#L94-L102

Mottie:
I removed the basic match and rebased from the master. This made my screenshot dark 😺

Oh, you've updated your comment with code... I replaced some of the punctuation in the search because it's not so easy to type `’` on some computers.

Mottie:
Ok, I've updated the code. The search from the screenshot doesn't work the same because there are no longer any string replacements.

source/options.tsx
```diff
@@ -62,6 +62,17 @@ async function init(): Promise<void> {
...
 	indentTextarea.watch('textarea');
 
+	// Search feature options
+	select<HTMLInputElement>('#search-features')!.addEventListener('input', event => {
+		const keywords = (event.target as HTMLInputElement)!.value.toLowerCase().split(' ').filter((s: string) => s.length);
+		const hasKeywords = keywords.length > 0;
+		for (const feature of select.all('.feature')) {
+			feature.hidden = hasKeywords ?
+				!keywords.every(word => feature.dataset.text.includes(word)) :
+				false;
```
fregante: This shouldn't be necessary. I _think_ this is equal:

```js
feature.hidden = !keywords.every(word => feature.dataset.text.includes(word));
```

source/options.css
```diff
@@ -110,7 +110,8 @@ h2 ~ h2 {
 	text-decoration: line-through;
 }
 
-.js-features :not(:checked) + .info .extended-options {
+.js-features :not(:checked) + .info .extended-options,
+.feature[hidden] {
```
fregante: This shouldn't be needed, `[hidden] {display:none}` is offered by the browser

fregante:
Also everything should be renamed to `filter` rather than `search` https://ux.stackexchange.com/a/4756/35104

Mottie:
Umm, please go ahead and make these changes... I'm having some issues with my computer.

Apparently running `npm install` now makes my computer reboot.

Then I saw this:

2019-11-03 22_46_20-MINGW64__x_Repos_+Misc_refined-github

And after a repair:

2019-11-03 22_51_20-Settings

I think I'm gonna call it a night LOL

source/options.tsx
Mottie: You're right... like I said my computer was doing some crazy stuff. I cleared the input and all the features stayed hidden. I thought I was going crazy. That and the times the input would enter text in backwards.

source/options.css
Mottie: It doesn't seem to be set... having the hidden attribute was not hiding the entries.

Mottie:
So... `npm install` is still having issues, but apparently `yarn install` worked.

Then YAY, I made the changes.

`npm test` failed and was showing some weird error message. `yarn test` caused a GSOD (apparently Windows insider now shows a green-screen of death vs the older, outdated blue-screen of death).

Tests seem to be passing here (finally) LOL... I still don't know how to deal with the `feature.dataset.text` TS error other than wrapping it in a `String()`.

_Now_, I'm going to bed.

Mottie:
And oops... I removed the `[hidden] { display: none; }` so now the features aren't getting hidden.

fregante:
> And oops... I removed the `[hidden] { display: none; }` so now the features aren't getting hidden.

Uhm, what browser are you using? It works correctly in both Firefox and Chrome for me

Mottie:
👍 Thanks for the fixes! Typescript is still weird to me.

jerone:
Great work @Mottie 👍

PR #2516, opened 2019-11-04
Move disabled features to the top of the list (Options page)

fregante:
This makes it easier to find the features that you previously disabled. This only happens on load and _not_ as the user checks/unchecks the features; you don't want things to disappear from the view.

---

### DIFF
--- a/source/options.tsx
+++ b/source/options.tsx
@@ -55,7 +55,7 @@ function buildFeatureCheckbox({name, description, screenshot, disabled}: Feature
 async function init(): Promise<void> {
 	// Generate list
 	const container = select('.js-features')!;
-	container.append(...__featuresInfo__.sort().map(buildFeatureCheckbox));
+	container.append(...__featuresInfo__.map(buildFeatureCheckbox));
 
 	// Update list from saved options
 	const form = select('form')!;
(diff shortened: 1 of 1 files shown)


jerone:
I'm leaning to -1 right now. Some buttons that allow sorting the list would be more useful I think; sort alphabetical & sort by state.

yakov116:
I think a button showing disabled would be better since #2514 would enable searching.

fregante:
The only difference between this and the full alphabetical order is that feature `hide-own-stars` for example is not in the middle of the list anymore. 

Now, why would you expect it to be there? Are you looking for that feature specifically? Then search it with the new filter bar. If anything, you’ll see it without searching because it’s already on top. 

Having your disabled feature *always* on top, **even when you’re not specifically looking for them,** makes them a whole lot more noticeable. For example in that very screenshot I noticed that I had disabled a feature I didn’t mean to disable.

fregante:
Yeah I don’t feel strongly about keeping broken feature up there too. I can restore those

mikhaelkh:
Maybe also move broken features to the bottom? Reason is pretty much the same — to have all broken features in one place. Also if one the first few features become broken the order will become less obvious.

PR #2521, opened 2019-11-06
Updates and lint

fregante:
I was hoping to use TypeScript 3.7’s new `nullish coalesce operator` and `optional chaining operator` but neither one is supported by eslint and ts-node, so I had to put them off. Eslint will support them once they reach Stage 4


package.json
```diff
@@ -118,6 +118,7 @@
 				"rules": {
 					"@typescript-eslint/no-misused-promises": "off",
 					"@typescript-eslint/promise-function-async": "off",
+					"@typescript-eslint/strict-boolean-expressions": "off",
```
fregante: cc @sindresorhus 

Have a look at these errors and where they appear. It doesn't let me use `if (!element) {return}`. I don't think that converting that to `typeof element === 'undefined'` makes it any better. 😬

If anything, `!element` and `element` are easier to tell apart than `typeof element === 'undefined' and `typeof element !== 'undefined'

https://github.com/sindresorhus/refined-github/runs/290686882

sindresorhus: I came to the same conclusion after having used it in the wild. Fixed: https://github.com/xojs/eslint-config-xo-typescript/releases/tag/v0.20.0

sindresorhus:
> I was hoping to use TypeScript 3.7’s new nullish coalesce operator and optional chaining operator but neither one is supported by eslint and ts-node, so I had to put them off. Eslint will support them once they reach Stage 4

I'm using it now: https://github.com/sindresorhus/got/commit/9f4fe33585bf89de37f61ff5614435105be28f64

Just upgrade to latest version of eslint-config-xo-typescript:
- https://github.com/xojs/eslint-config-xo-typescript/releases/tag/v0.20.0
- https://github.com/xojs/eslint-config-xo-typescript/releases/tag/v0.21.0

source/libs/get-text-nodes.ts
```diff
@@ -1,12 +1,12 @@
 export default (el: Node): Text[] => {
 	const walker = document.createTreeWalker(el, NodeFilter.SHOW_TEXT);
 	const nodes: Text[] = [];
-	let node: Text;
+	let node;
```
sindresorhus: This would be more correct:

```suggestion
	let node: Text | null;
```

I try not to use `as` when there are there are other solutions, as `as` can easily hide type errors as it force-casts.

source/libs/get-text-nodes.ts
```diff
@@ -1,12 +1,12 @@
 export default (el: Node): Text[] => {
```
sindresorhus: While you're at it:
```suggestion
export default (element: Node): Text[] => {
```

source/libs/on-new-comments.ts
```diff
@@ -7,7 +7,6 @@ const observed = new WeakSet();
 
 const run = debounce(() => {
 	// Safely run all callbacks
-	// eslint-disable-next-line @typescript-eslint/require-await
 	handlers.forEach(async cb => cb());
```
sindresorhus:
```suggestion
	for (const handler of handlers) {
		handler();
	}
```

source/libs/simplified-element-observer.ts
```diff
@@ -1,12 +1,12 @@
-export default function observeEl(el: Node|string, listener: MutationCallback, options: MutationObserverInit = {childList: true}): MutationObserver | undefined {
+export default function observeEl(
+	el: Node | string,
```
sindresorhus:
```suggestion
	element: Node | string,
```

test/utils.ts
```diff
@@ -150,6 +150,7 @@ test('getOwnerAndRepo', t => {
 
 	Object.keys(ownerAndRepo).forEach(url => {
```
sindresorhus:
```suggestion
	for (const [url, object] of ownerAndRepo) {
```

test/utils.ts
```diff
@@ -150,6 +150,7 @@ test('getOwnerAndRepo', t => {
...
 		location.href = url;
+		// @ts-ignore
```
sindresorhus: Add a comment why it's ignored.
```suggestion
		// @ts-ignore Because of unicorns.
```

source/libs/page-detect.ts
```diff
@@ -75,7 +75,7 @@ export const isReleasesOrTags = (): boolean => /^(releases|tags)/.test(getRepoPa
 export const isEditingFile = (): boolean => /^edit/.test(getRepoPath()!);
```
sindresorhus:
```suggestion
export const isEditingFile = (): boolean => getRepoPath()?.startsWith('edit');
```

Applies to many others in this file.

source/libs/fetch-dom.ts
```diff
@@ -2,12 +2,12 @@ import domify from 'doma';
 import mem from 'mem';
 
 async function fetchDom(url: string): Promise<DocumentFragment>;
-async function fetchDom<TElement extends Element>(url: string, selector: string): Promise<TElement>;
-async function fetchDom(url: string, selector?: string): Promise<Node> {
+async function fetchDom<TElement extends Element>(url: string, selector: string): Promise<TElement | null>;
+async function fetchDom(url: string, selector?: string): Promise<Node | null> {
```
sindresorhus: Would you be willing to change `fetchDom` to return `undefined` instead of `null`?

Re: https://github.com/sindresorhus/meta/issues/7

We cannot change the browser APIs, but at least we can improve our own APIs.

source/features/mark-private-orgs.tsx
```diff
@@ -11,8 +11,8 @@ async function init(): Promise<false | void> {
 		return false;
 	}
 
-	let publicOrgs = await api.v3(`users/${getUsername()}/orgs`);
-	publicOrgs = publicOrgs.map((orgData: AnyObject) => `/${orgData.login}`);
+	const response = await api.v3(`users/${getUsername()}/orgs`);
+	const publicOrgs = response.map((orgData: AnyObject) => `/${orgData.login}`) as string[];
```
sindresorhus: Do you really need the type assertion here? I would assume TS would be able to infer the string.

Even this would be better:

```suggestion
	const publicOrgs: string[] = response.map((orgData: AnyObject) => `/${orgData.login}`);
```

source/features/infinite-scroll.tsx
```diff
@@ -3,16 +3,16 @@ import debounce from 'debounce-fn';
 import features from '../libs/features';
 import observeEl from '../libs/simplified-element-observer';
 
-let btn: HTMLButtonElement;
+let btn: HTMLButtonElement | undefined;
```
sindresorhus:
```suggestion
let button: HTMLButtonElement | undefined;
```

source/libs/on-new-comments.ts
fregante: Not all handlers are async. If `handler` errors it will crash RG. Sadly there's no `for async of` to replicate `.foreach(async)`

source/features/mark-private-orgs.tsx
fregante: > I would assume TS would be able to infer the string.

Sadly not. I think `publicOrgs` was an `any` here. I typed it because `no-unnecessary-condition` gave me an error on the `.includes()` condition

source/features/infinite-scroll.tsx
fregante: Didn't you add a rule to detect all of these shortened variables? e.g. `evt` and `err`

fregante:
> I'm using it now: [sindresorhus/got@9f4fe33](https://github.com/sindresorhus/got/commit/9f4fe33585bf89de37f61ff5614435105be28f64)

`ts-node` (we use it for AVA) was giving me syntax errors. How are you able to use it?

source/libs/on-new-comments.ts
sindresorhus: Ok, how about this then? The intent is clearer.

```js
for (const handler of handlers) {
	(async () => {
		try {
			await handler();
		} catch {}
	})();
}
```

source/features/infinite-scroll.tsx
sindresorhus: Yes, but I disabled it in XO until the rule stabilizes and I have time to deal with the repercussions (it will be unpopular with many users). We could enable it manually in our config for now though?

sindresorhus:
> ts-node (we use it for AVA) was giving me syntax errors. How are you able to use it?

https://github.com/sindresorhus/got/blob/9f4fe33585bf89de37f61ff5614435105be28f64/package.json#L91-L105

source/libs/on-new-comments.ts
fregante: Sorry Sindre but I really don't see why you insist on making things complicate around async functions. It turns a 5-keyword single line and comment into 10-keyword 7-line that isn't any better to read nor faster to execute. Just why.

sindresorhus: 🤷‍♂ I just prefer explicicity. Feel free to ignore.

fregante: I like it too, but not in this case. I will extend the comment 😁

fregante:
It still errors, even with a brand-new lock file and `node_modules`

```ts
ERROR in ./source/features/ci-link.tsx 8:61
Module parse failed: Unexpected token (8:61)
File was processed with these loaders:
 * ./node_modules/ts-loader/index.js
You may need an additional loader to handle the result of these loaders.
| import fetchDom from '../libs/fetch-dom';
| export const fetchCIStatus = onetime(async () => {
>     const url = `/${getRepoURL()}/commits/${getRepoBranch() ?? ''}`;
|     const icon = await fetchDom(url, '.commit-build-statuses');
|     if (icon) {
```

source/libs/get-text-nodes.ts
fregante: I agree and that was my first solution, but this can't be avoided, I think.

`walker.nextNode()` returns a `Node` which doesn't fit `Text | null`

We can change `get-text-nodes` to return `Node[]` (even though we're sure that `createTreeWalker` returns `Text`) but then we'll still have to cast those nodes to `Text` in other places that use APIs not available on `Node`:

`textNode.replaceWith(fragment)` in `show-whitespace` would have to become `(textNode as Text).replaceWith(fragment)`

source/libs/on-new-comments.ts
sindresorhus: > I will extend the comment

Instead of letting the code do the explanation...

sindresorhus:
Works for me locally:

```
~/dev/oss/refined-github pr/2521
❯ nt

> @ test /Users/sindresorhus/dev/oss/refined-github
> run-p --silent test:js lint:* build


⠙
  source/globals.d.ts:1:1
  ⚠   1:1  Unexpected todo comment.  no-warning-comments
  ⚠   2:1  Unexpected todo comment.  no-warning-comments

  source/libs/post-form.ts:9:3
  ⚠   9:3  Unexpected todo comment.  no-warning-comments

  source/features/hidden-review-comments-indicator.tsx:37:2
  ⚠  37:2  Unexpected todo comment.  no-warning-comments

  source/features/revert-file.tsx:89:4
  ⚠  89:4  Unexpected todo comment.  no-warning-comments

  5 warnings
⠇ Starting type checking service...
Using 1 worker with 2048MB memory limit

  42 tests passed


           background.js ⏤  10.6 kB
 browser-polyfill.min.js ⏤  0 B (-3.14 kB)
             content.css ⏤  10.1 kB
              content.js ⏤  85.3 kB
            features.css ⏤  0 B (-7.3 kB)
             options.css ⏤  1.12 kB
            options.html ⏤  0 B (-964 B)
              options.js ⏤  29.2 kB
    resolve-conflicts.js ⏤  1.62 kB

Built at: 11/07/2019 6:14:35 PM
```

source/features/infinite-scroll.tsx
fregante: Done. Love it. However `--fix` drops TypeScript types

source/libs/get-text-nodes.ts
sindresorhus: Ugh, I assumed the TS type would be good enough to return the correct type because we specify `SHOW_TEXT`...

source/features/infinite-scroll.tsx
sindresorhus: > However --fix drops TypeScript types

That's not good...

test/utils.ts
fregante: Fixed error altogether by changing data structure to a Map

fregante:
What node version are you on? Could that be it? I tried on both Node 10 and 12

> Works for me locally:

Did you make that change locally? It was not pushed to the PR. CI fails for 
95fd3b5  https://github.com/sindresorhus/refined-github/runs/292536684

test/utils.ts
fregante: `ownerAndRepo` isn't an iterable object. But I made it one and also avoided the `ts-ignore` 
fc5d463

source/libs/on-new-comments.ts
fregante: I like explicity too but I don't think it helps in this case.

The semi-equivalent would be this, except the error would technically be considered "caught":

```js
	for (const handler of handlers) {
		try {
			handler();
		} catch (error) {
			console.log(error);
		}
	}
```

But then again I see no advantage over 1-line + comment, which I extended: 
9a1c22b

test/utils.ts
sindresorhus: > ownerAndRepo isn't an iterable object. 

Objects aren't iterable, yes, but you could have used `Object.entries()` for that. Doesn't matter now though.

sindresorhus:
Oh, I assumed I could just test with the PR.

sindresorhus:
I finally figured it out. You need to set the `target` in tsconfig to `es2018` instead of `esnext`. See: https://github.com/TypeStrong/ts-loader/issues/1031

source/libs/get-text-nodes.ts
fregante: DOM types aren’t great. Once I suggested a change to add a generic to cloneNode, I think, and they refused because of performance concerns. So… assertions it is

source/libs/page-detect.ts
fregante: undefined

This results in `boolean | undefined`. I'd probably have to change it to:

```suggestion
export const isEditingFile = (): boolean => getRepoPath()?.startsWith('edit') || false;
```

Worth it?

fregante:
I added some `?.` and `??`, I'm not entirely sure I'm using them right yet.

I noticed that TypeScript doesn't yet use `if (foo?.bar) {}` as a signal that `foo` isn't `undefined` inside the following block, so you'll have to use `!` in the block: `if (foo?.bar) {foo!.bar}`

Likewise you can't use `unknown?.bar` because it complains that it's `unknown`. But "type checks by property-check" have always been a pain in TS, requiring `(foo as any).bar` even if you're just testing the property specifically to determine `foo`'s type.

sindresorhus:
> Likewise you can't use unknown?.bar because it complains that it's unknown. 

Yeah, that's annoying. You could try to use `Record<string, unknown>` instead of `unknown` if you know the type is some kind of object.

sindresorhus:
Let's get this merged and we can look into using `??` and `?.` more later on.

source/features/preview-hidden-comments.tsx
```diff
@@ -21,7 +21,7 @@ const init = (): void => {
 			summary .discussion-item-copy
 		`, details)!;
 
-		const [, reason = ''] = /was marked as ([^.]+)/.exec(header.textContent!) || [];
+		const reason = /was marked as ([^.]+)/.exec(header.textContent!)?.[1] || '';
```
sindresorhus:
```suggestion
		const reason = /was marked as ([^.]+)/.exec(header.textContent!)?.[1] ?? '';
```

PR #2522, opened 2019-11-06
Pull Request details spacing fix

gaillota:
# Issue
Small design glitch where the PR details from Refined-Github overflow the PR informations (see screenshots below)

# Fix
Using Github CSS class `mt-1`, it adds a tiny spacing above the PR details injected by Refined Github

# Screenshots
## Before
Screenshot 2019-11-06 at 15 02 24

## After
Screenshot 2019-11-06 at 15 03 04

### DIFF
--- a/source/features/pr-branches.tsx
+++ b/source/features/pr-branches.tsx
@@ -129,7 +129,7 @@ async function init(): Promise<false | void> {
 		}
 
 		select('.text-small.text-gray', PR)!.append(
-			<span className="issue-meta-section d-inline-block">
+			<span className="mt-1 issue-meta-section d-inline-block">
 				{openPullRequest()}
 				{' '}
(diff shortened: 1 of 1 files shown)


gaillota:
Edit: It appears that the details are supposed to be inlined, and my screenshots happens when the content is too long and the element goes to the next line. Hence not working

fregante:
I think this PR is fine

gaillota:
Won't this affect the inline version of the PR details by adding an unnecessary `margin-top` ?

fregante:
I think the distance would be minimal

gaillota:
What do you mean ?

fregante:
I mean that if this PR is merged, the extra spacing would not be noticeable when it doesn't wrap, so it's ok.

Issue #2264, opened 2019-07-20
`shorten-links` should apply also after you edit your comment


fregante:
1. Edit one of your comments containing a URL
2. Save it
3. Notice that the links are no longer shortened by the `shorten-links` feature.

### How to fix this

- My other extension https://github.com/bfred-it/github-issue-link-status correctly re-applies the changes when you edit your comment, so the code can probably be copied from there
- This should probably be part of `on-new-comments.ts` so all related features benefit from it.

getlife2206:
I have no freakin clue what im supposed to say! But I'll say sorry. I can do better. But ive got truck broke down and start job tomm. Have to ride bike to get part. I make mistakes. I'm human.

fregante:
Duplicate of https://github.com/sindresorhus/refined-github/issues/2572

fregante:
This should follow https://github.com/sindresorhus/refined-github/pull/2992

arjunvegda:
This issue seems very interesting! 

I did a little bit of digging, and it looks like the `shorten-links` feature is not being applied to the newly added comments as well. Based on the description of this issue, it sounds like it should, right?

And should `shorten-links` feature include `onNewComments` listener or indicate that the page has comments?

Adding this to the `shorten-links` seems to "fix" the issue for links in the comments. However, it may not work for links outside of the comment/discussion box.

```
include: [
 pageDetect.hasComments
]
``` 

I am pretty new to repo so I might not be heading in the right direction, right away, haha.

fregante:
If `include` is missing, it will be applied to every page. Adding it might fix it because explicitly specifying that [automatically includes `onNewComments`](https://github.com/sindresorhus/refined-github/blob/f0305d035ea72573bc73d89620713b29448d3775/source/features/index.tsx#L159).

You can try just adding `onNewComments` like in: https://github.com/sindresorhus/refined-github/blob/0fa514dbeeab6395ef095cff80beac41443a6574/source/features/hide-inactive-deployments.tsx#L28

PR #2533, opened 2019-11-12
Various small improvements to ajaxed widgets updates handler

fregante:
I was trying to tackle https://github.com/sindresorhus/refined-github/issues/2264 but I don't know if I will get there.


fregante:
Merging this as is, as it's already an improvement over the previous polling code

PR #2549, opened 2019-11-18
Add `user-local-time` feature

### LINKED ISSUE #1233: Add local time to hover cards based on user location
When collaborating on open-source with people from around the world, it's good to know if the local time with the person you're communicating is in the morning/evening.

image

dertieran:
This is the first implementation of adding the user's time to their hovercard.
This is done through checking their last commit author date.

<details>
<summary>Screenshot</summary>

image
</details>

It is still a bit rough implementation and might not be final.

Todos:
- [x] Should the custom `api` be handled here or extracted out
- [ ] How should errors be handled?
	Right now most of them just result into an empty time (that is also chached).

Closes #1233

### DIFF
--- a/source/features/show-local-user-time.tsx
+++ b/source/features/show-local-user-time.tsx
@@ -18,38 +18,28 @@ const timeFormatter = new Intl.DateTimeFormat('en-US', {
 	hour12: false
 });
 
-function normalizeApiUrl(url: string): string {
-	return new URL(url).pathname.slice(1).replace('api/v3/', '');
-}
+async function loadLastCommit(pathname: string): Promise<Commit | void> {
+	for await (const page of api.v3paginated(pathname)) {
+		for (const event of page as any) { // eslint-disable-line @typescript-eslint/no-explicit-any
+			if (event.type !== 'PushEvent') {
+				continue;
+			}
 
-async function loadLastCommit(pathname: string): Promise<Commit | undefined> {
-	const events = await api.v3(pathname) as any; // eslint-disable-line @typescript-eslint/no-explicit-any
-	for (const event of events) {
-		if (event.type !== 'PushEvent') {
-			continue;
-		}
+			// Start from the latest commit, which is the last one in the list
+			for (const commit of event.payload.commits.reverse()) {
+				const response = await api.v3(commit.url, {ignoreHTTPStatus: true});
 
-		// NOTE: We want to iterate the commits in reverse to start with the latest commitjust
-		for (const commit of event.payload.commits.reverse()) {
-			const response = await api.v3(normalizeApiUrl(commit.url), {ignoreHTTPStatus: true});
-
-			// `response.author` only appears if GitHub can match the email to a GitHub user
-			if (response.author?.id === event.actor.id) {
-				return commit;
+				// `response.author` only appears if GitHub can match the email to a GitHub user
+				if (response.author?.id === event.actor.id) {
+					return commit;
+				}
 			}
 		}
 	}
-
-	const [, next] = /<([^>]+)>; rel="next"/.exec(events.headers.get('link')!) ?? [];
-	if (!next) {
-		return;
-	}
-
-	return loadLastCommit(normalizeApiUrl(next));
 }
 
 async function getCommitPatch(commitUrl: string): Promise<string> {
-	const {textContent} = await api.v3(normalizeApiUrl(commitUrl), {
+	const {textContent} = await api.v3(commitUrl, {
 		json: false,
 		headers: {
 			Accept: 'application/vnd.github.v3.patch'
--- a/source/features/show-local-user-time.tsx
+++ b/source/features/show-local-user-time.tsx
@@ -79,13 +79,10 @@ function init(): void {
 			return;
 		}
 
-		// NOTE: Adding the time element might changes the height of the hovercard and thuse break the positioning.
-		// So we store the container height and adjust the positioning if needed after adding the placeholder.
+		// Adding the time element might change the height of the hovercard and thus break its positioning
 		const containerHeight = container.offsetHeight;
-		const classNames = ['Popover-message--bottom-right', 'Popover-message--bottom-left'];
-		const needsAdjustment = classNames.some(name => container.classList.contains(name));
 
-		const placeholder = <span>Loading…</span>;
+		const placeholder = <span>Loading timezone…</span>;
 
 		select('div.d-flex.mt-3 > div.overflow-hidden.ml-3', container)!.append(
 			<div className="rgh-local-user-time mt-2 text-gray text-small">
@@ -93,11 +90,11 @@ function init(): void {
 			</div>
 		);
 
-		if (needsAdjustment) {
+		if (container.matches('.Popover-message--bottom-right, .Popover-message--bottom-left')) {
 			const diff = container.offsetHeight - containerHeight;
 			if (diff > 0) {
 				const parent = container.parentElement!;
-				const top = parseInt(parent.style.top ?? '0', 10);
+				const top = parseInt(parent.style.top, 10);
 				parent.style.top = `${top - diff}px`;
 			}
 		}
--- a/source/features/show-local-user-time.tsx
+++ b/source/features/show-local-user-time.tsx
@@ -12,48 +12,35 @@ interface Commit {
 	url: string;
 }
 
-interface Event {
-	type: string;
-	payload: {
-		commits: Commit[];
-	};
-	actor: {
-		id: number;
-	};
+const timeFormatter = new Intl.DateTimeFormat('en-US', {
+	hour: 'numeric',
+	minute: 'numeric',
+	hour12: false
+});
+
+function normalizeApiUrl(url: string): string {
+	return new URL(url).pathname.slice(1).replace('api/v3/', '');
 }
 
-const normalizeApiUrl = (url: string): string => {
-	const {pathname} = new URL(url);
-	return pathname.replace(/^\/(api\/v3\/)?/, '');
-};
-
-async function loadLastCommit(pathname: string): Promise<Commit|undefined> {
-	const response = await api.v3(pathname);
-
-	// NOTE: We iterate in series here to reduce requests
-	// The first commit should already work in most cases
-	const events = response.filter((event: Event) => event.type === 'PushEvent') as Event[];
+async function loadLastCommit(pathname: string): Promise<Commit | undefined> {
+	const events = await api.v3(pathname) as any; // eslint-disable-line @typescript-eslint/no-explicit-any
 	for (const event of events) {
+		if (event.type !== 'PushEvent') {
+			continue;
+		}
+
 		// NOTE: We want to iterate the commits in reverse to start with the latest commitjust
 		for (const commit of event.payload.commits.reverse()) {
 			const response = await api.v3(normalizeApiUrl(commit.url), {ignoreHTTPStatus: true});
-			if (!response.ok) {
-				// NOTE: Ignore 404 errors and check the next commit
-				if (response.httpStatus === 404) {
-					continue;
-				}
 
-				return;
-			}
-
-			// NOTE: Some commits don't have an author when there is no github user for the autor email.
+			// `response.author` only appears if GitHub can match the email to a GitHub user
 			if (response.author?.id === event.actor.id) {
 				return commit;
 			}
 		}
 	}
 
-	const [, next] = /<([^>]+)>; rel="next"/.exec(response.headers.get('link')!) ?? [];
+	const [, next] = /<([^>]+)>; rel="next"/.exec(events.headers.get('link')!) ?? [];
 	if (!next) {
 		return;
 	}
@@ -61,8 +48,8 @@ async function loadLastCommit(pathname: string): Promise<Commit|undefined> {
 	return loadLastCommit(normalizeApiUrl(next));
 }
 
-async function getCommitPatch(commit: Commit): Promise<string> {
-	const {textContent} = await api.v3(normalizeApiUrl(commit.url), {
+async function getCommitPatch(commitUrl: string): Promise<string> {
+	const {textContent} = await api.v3(normalizeApiUrl(commitUrl), {
 		json: false,
 		headers: {
 			Accept: 'application/vnd.github.v3.patch'
@@ -80,15 +67,13 @@ const loadTimezoneOffset = mem(async (login: string): Promise<number|undefined> 
 		return;
 	}
 
-	const patch = await getCommitPatch(commit);
+	const patch = await getCommitPatch(commit.url);
 	const [, hourString, minuteString] = (/^Date: .* ([-+]\d\d)(\d\d)$/m).exec(patch) ?? [];
 
 	const hours = parseInt(hourString, 10);
 	const minutes = parseInt(minuteString, 10);
 	return (hours * 60) + (hours < 0 ? -minutes : minutes);
 });
-
-const format = (number: number): string => number.toString().padStart(2, '0');
 
 function init(): void {
 	const container = select('.js-hovercard-content > .Popover-message')!;
@@ -98,8 +83,8 @@ function init(): void {
 			return;
 		}
 
-		const profile = select<HTMLAnchorElement>('a[data-octo-dimensions="link_type:profile"]', container);
-		if (!profile) {
+		const login = select<HTMLAnchorElement>('a[data-octo-dimensions="link_type:profile"]', container)?.pathname.slice(1);
+		if (!login) {
 			return;
 		}
 
@@ -126,16 +111,15 @@ function init(): void {
 			}
 		}
 
-		const login = profile.pathname.replace(/^\//, '');
 		const offset = await loadTimezoneOffset(login);
-		if (!Number.isFinite(offset!)) {
+		if (typeof offset === 'undefined') {
 			placeholder.textContent = '-';
 			return;
 		}
 
 		const date = new Date();
-		date.setMinutes(offset! + date.getTimezoneOffset() + date.getMinutes());
-		placeholder.textContent = `${format(date.getHours())}:${format(date.getMinutes())}`;
+		date.setMinutes(offset + date.getTimezoneOffset() + date.getMinutes());
+		placeholder.textContent = timeFormatter.format(date);
 	});
 }
 
--- a/source/features/show-local-user-time.tsx
+++ b/source/features/show-local-user-time.tsx
@@ -8,10 +8,13 @@ import {clock} from '../libs/icons';
 interface Commit {
 	url: string;
 }
+
 interface Event {
 	type: string;
 	payload: AnyObject;
-	actor: {id: number};
+	actor: {
+		id: number;
+	};
 }
 
 // NOTE: This is basically copied from api.v3() but we can't use it because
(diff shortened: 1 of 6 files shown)


fregante:
Why is the custom API implementation needed? Is it just to get the headers? You can include them in the original API like we include `httpStatus`

dertieran:
@fregante I actually need it for all three usages 😄 

1. For the `users/:login/events` request I need to get the `link` header.
https://github.com/sindresorhus/refined-github/blob/4cc6cda099ab953ca80a4f3b4780080c485ed995/source/features/show-local-user-time.tsx#L79-L81

2. For the commit request I need to be able to pass an absolute url.
https://github.com/sindresorhus/refined-github/blob/4cc6cda099ab953ca80a4f3b4780080c485ed995/source/features/show-local-user-time.tsx#L60

3. For the patch commit I need the raw `.text()` response.
https://github.com/sindresorhus/refined-github/blob/4cc6cda099ab953ca80a4f3b4780080c485ed995/source/features/show-local-user-time.tsx#L92-L95

I wasn't sure how to integrate this into the `api.v3` properly without bloating it 🙈

fregante:
1. You can pass `headers` in the final object
2. `await api(new URL(commit.url).pathname)` should work
3. This is a little more complicated, but it should only mean adding one more option: `json: false` (`json: true` being the default). And then something like
	```js
	const apiResponse: JsonObject = json ? JSON.parse(textContent) : {textContent};

	if (response.ok || ignoreHTTPStatus) {
		return Object.assign(apiResponse, {
			httpStatus: response.status,
			ok: response.ok
		});
	}
	```

This would be a replacement for the implicit `textContent.length > 0` check we have now to avoid `JSON.parse`. The features that rely on that would also have to be changed to explicitly avoid it, by using `json: false`

source/features/show-local-user-time.tsx
```diff
@@ -0,0 +1,184 @@
...
+// IDEA: We could also return the date from the patch
+// This could help to identify "wrong" offsets e.g. daylight saving
+async function loadTimezoneOffset(login: string): Promise<number> {
+	const commit = await getLastCommit(login);
+	if (!commit) {
+		return NaN;
+	}
+
+	const patch = await getCommitPatch(commit);
+	const match = (/^Date: .*? (\+|-)(\d\d)(\d\d)$/m).exec(patch);
```
fregante:
```suggestion
	const match = (/^Date: .*? ([-+]\d\d)(\d\d)$/m).exec(patch);
```

The sign can be parsed with the hour, so later you don't have to handle it explicitly.

Also we should always expect a date, let's not handle the error at all.

```suggestion
	const [, hours, minutes] = (/^Date: .* ([-+]\d\d)(\d\d)$/m).exec(patch) || [];
	return (parseInt(hours, 10) * 60) + parseInt(minutes, 10);
```

source/features/show-local-user-time.tsx
```diff
@@ -0,0 +1,184 @@
...
+		if (!container.childElementCount || select.exists('.rgh-local-user-time')) {
+			return;
+		}
+
+		const profile = select<HTMLAnchorElement>('a[data-octo-dimensions="link_type:profile"', container);
+		if (!profile) {
+			return;
+		}
+
+		const content = select('div.d-flex.mt-3 > div.overflow-hidden.ml-3', container);
```
fregante:
```suggestion
		const content = select('div.d-flex.mt-3 > div.overflow-hidden.ml-3', container)!;
```

We always expect the element, right? If it's not there, the DOM has changed and the error should be thrown to the console. Let's not fill the code with `if !aaa return` unless these elements might really be missing in some situations (like `!profile`, that's fine)

source/features/show-local-user-time.tsx
```diff
@@ -0,0 +1,184 @@
...
+			// NOTE: Some commits don't have an author when there is no github user for the autor email.
+			if (author && author.id === event.actor.id) {
+				return commit;
+			}
+		}
+	}
+
+	if (!response.headers.has('link')) {
+		return null;
+	}
```
fregante: I think you can skip this check. The `exec` will just skip the `undefined`

source/features/show-local-user-time.tsx
```diff
@@ -0,0 +1,184 @@
...
+
+const getLastCommit = (login: string): Promise<Commit|null> => loadLastCommit(`users/${login}/events`);
+async function getCommitPatch(commit: Commit): Promise<string> {
+	const response = await api(commit.url, {headers: {Accept: 'application/vnd.github.v3.patch'}});
+	return response.ok ? response.text() : '';
+}
+
+// IDEA: We could also return the date from the patch
+// This could help to identify "wrong" offsets e.g. daylight saving
+async function loadTimezoneOffset(login: string): Promise<number> {
```
fregante:
```suggestion
const loadTimezoneOffset = mem(async (login: string): Promise<number> => {
```

`mem` will take care of the caching and you can drop `getTimezoneOffset` altogether

source/features/show-local-user-time.tsx
```diff
@@ -0,0 +1,184 @@
...
+	return fetch(url, {
+		...options,
+		headers
+	});
+};
+
+async function loadLastCommit(url: string): Promise<Commit|null> {
+	const response = await api(url);
+	if (!response.ok) {
+		return null;
```
sindresorhus: Prefer `undefined` over `null`

source/features/show-local-user-time.tsx
```diff
@@ -0,0 +1,184 @@
...
+				return commit;
+			}
+		}
+	}
+
+	if (!response.headers.has('link')) {
+		return null;
+	}
+
+	const [, next] = (/<([^>]+)>; rel="next"/).exec(response.headers.get('link')!) || [];
```
sindresorhus:
```suggestion
	const [, next] = /<([^>]+)>; rel="next"/.exec(response.headers.get('link')!) ?? [];
```

source/features/show-local-user-time.tsx
```diff
@@ -0,0 +1,184 @@
...
+	const response = await api(commit.url, {headers: {Accept: 'application/vnd.github.v3.patch'}});
+	return response.ok ? response.text() : '';
+}
+
+// IDEA: We could also return the date from the patch
+// This could help to identify "wrong" offsets e.g. daylight saving
+async function loadTimezoneOffset(login: string): Promise<number> {
+	const commit = await getLastCommit(login);
+	if (!commit) {
+		return NaN;
```
sindresorhus: Don't use `NaN`. Just return undefined.

```suggestion
		return;
```

source/features/show-local-user-time.tsx
```diff
@@ -0,0 +1,184 @@
...
+			return;
+		}
+
+		// NOTE: Adding the time element might changes the height of the hovercard and thuse break the positioning.
+		// So we store the container height and adjust the positioning if needed after adding the placeholder.
+		const containerHeight = container.offsetHeight;
+		const classNames = ['Popover-message--bottom-right', 'Popover-message--bottom-left'];
+		const needsAdjustment = classNames.some(name => container.classList.contains(name));
+
+		const placeholder = <span>Loading...</span>;
```
sindresorhus:
```suggestion
		const placeholder = <span>Loading…</span>;
```

source/features/show-local-user-time.tsx
```diff
@@ -0,0 +1,184 @@
...
+			<div className="rgh-local-user-time mt-2 text-gray text-small">
+				{clock()} {placeholder}
+			</div>
+		);
+
+		if (needsAdjustment) {
+			const diff = container.offsetHeight - containerHeight;
+			if (diff > 0) {
+				const parent = container.parentElement!;
+				const top = parseInt(parent.style.top || '0', 10);
```
sindresorhus:
```suggestion
				const top = parseInt(parent.style.top ?? '0', 10);
```

source/features/show-local-user-time.tsx
sindresorhus: `||` => `??`

source/features/show-local-user-time.tsx
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+	return cache.get(login)!;
+}
+
+const format = (number: number): string => number.toString().padStart(2, '0');
+function init(): void {
+	const container = select('.js-hovercard-content > .Popover-message')!;
+	observeEl(container, async () => {
+		if (!container.childElementCount || select.exists('.rgh-local-user-time')) {
```
sindresorhus:
```suggestion
		if (container.childElementCount === 0 || select.exists('.rgh-local-user-time')) {
```

source/features/show-local-user-time.tsx
```diff
@@ -0,0 +1,184 @@
...
+			return;
+		}
+
+		const profile = select<HTMLAnchorElement>('a[data-octo-dimensions="link_type:profile"', container);
```
sindresorhus:
```suggestion
		const profile = select<HTMLAnchorElement>('a[data-octo-dimensions="link_type:profile"]', container);
```

source/features/show-local-user-time.tsx
dertieran: Sure that's possible it's just important to then also apply the sign to the minutes.

source/features/show-local-user-time.tsx
```diff
@@ -0,0 +1,184 @@
...
+	// The first commit should already work in most cases
+	const events = await response.json() as Event[];
+	for (const event of events) {
+		if (event.type !== 'PushEvent') {
+			continue;
+		}
+
+		// NOTE: We want to iterate the commits in reverse to start with the latest commit
+		const commits = event.payload.commits as Commit[];
+		for (let i = commits.length - 1; i >= 0; i--) {
```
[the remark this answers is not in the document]
dertieran: Had it like this first 😄
Changed it to not mutate the commits but I guess thats not that important here.

source/features/show-local-user-time.tsx
fregante: Errrr, you're right. Discard that suggestion then, you can just drop `if (!match)`

dertieran:
Thanks for the review, will go through it :+1:

Just noticed another problem:
@fregante latest push event commit is a [merge commit](https://api.github.com/repos/maxmclau/refined-github/commits/5f5516995da03d2cb333600f405d0208e141403b).
But the [patch file](https://github.com/maxmclau/refined-github/commit/5f5516995da03d2cb333600f405d0208e141403b.patch) for that commit doesn't contain the merge commit and so the wrong date is parsed.
So another special case that needs to be hanled :smile:

source/features/show-local-user-time.tsx
dertieran: Already changed it :smile:
But I think the new solution might be clearer.

source/features/show-local-user-time.tsx
dertieran: Really like that solution :+1:

fregante:
>  But the [patch file](https://github.com/maxmclau/refined-github/commit/5f5516995da03d2cb333600f405d0208e141403b.patch) for that commit doesn't contain the merge commit and so the wrong date is parsed.

This means that `fetch patch` needs to be part of the loop, right? `if fetch(patch) === mergeCommit, continue`

dertieran:
> This means that `fetch patch` needs to be part of the loop, right? `if fetch(patch) === mergeCommit, continue`

Yes I think that's the solution.

dertieran:
@fregante I moved everything into `libs/api` in ad21c0d919db6865169b2be331be8e7bc9572b82.

I wasn't sure how to make the iterator work without TS yelling so I did this instead.
https://github.com/dertieran/refined-github/blob/ad21c0d919db6865169b2be331be8e7bc9572b82/source/features/show-local-user-time.tsx#L35-L36

Happy for any suggestions to make this work with TS :smile:

source/features/show-local-user-time.tsx
```diff
@@ -0,0 +1,148 @@
...
+		headers: {
+			Accept: 'application/vnd.github.v3.patch'
+		}
+	});
+
+	return textContent;
+}
+
+// IDEA: We could also return the date from the patch
+// This could help to identify "wrong" offsets e.g. daylight saving
```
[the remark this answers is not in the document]
dertieran: Maybe this was bad wording, with identify I didn't mean programmatically.
But maybe that information would be helpful to a user how *old* this information is.

*I thought about this when checking the time for somebody I knew had the same timezone as I. But the time was off by an hour, so I checked the patch and noticed that it was from summer.*
I thouhad this issue when checking the time of a colleague of mine, hi

fregante:
I think now this is left:

- https://github.com/sindresorhus/refined-github/pull/2549#issuecomment-557791905
- caching via `webext-storage-cache` (key: user, value: timezone, time: 10 days?)

I left a `fetch` logging in there to have an idea of how many requests will happen in practice.

source/features/show-local-user-time.tsx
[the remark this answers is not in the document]
fregante: Indeed, perhaps we should return/cache the whole unparsed date string that we get from the `patch` and then we can display it in a `title` attribute, like: `Last commit found: Sat, 2 Nov 2019 22:01:35 +0700`

source/libs/api.ts
```diff
@@ -85,9 +88,15 @@ export const v3 = mem(async (
...
+	const {ignoreHTTPStatus, method, body, headers, json} = {...v3defaults, ...options};
 	const {personalToken} = await settings;
 
+	// Handle full URLs we might have received from the API itself
+	if (query.startsWith('https://')) {
+		query = new URL(query).pathname.slice(1).replace('api/v3/', '');
+	}
+
+	console.log('Will fetch', api3 + query); // TODO: remove temporary logging
 	const response = await fetch(api3 + query, {
```
dertieran: Just realized I made an error in my code before, this will strip the `search` from the url 🙈 
So it should have been:
```js
const { pathname, search } = new URL(url);
return pathname.slice(1).replace('api/v3/', '') + search;
```

But now that this is in the lib it might be easier to just do:
```js
const url = new URL(query, api3);
await fetch(url)
```

source/libs/api.ts
```diff
@@ -101,18 +110,35 @@ export const v3 = mem(async (
 	const textContent = await response.text();
 
 	// The response might just be a 200 or 404, it's the REST equivalent of `boolean`
-	const apiResponse: JsonObject = textContent.length > 0 ? JSON.parse(textContent) : {};
+	const apiResponse: JsonObject = json ? JSON.parse(textContent.length > 0 ? textContent : '') : {textContent};
```
dertieran: I think that doesn't fix the issue here, because it still tries to do `JSON.parse('')` 👀 

Maybe something like this would be enough:
```suggestion
	const apiResponse: JsonObject = (json && textContent.length > 0) ? JSON.parse(textContent) : {textContent};
```

source/features/show-local-user-time.tsx
```diff
@@ -0,0 +1,148 @@
...
+		for (const commit of event.payload.commits.reverse()) {
+			const response = await api.v3(normalizeApiUrl(commit.url), {ignoreHTTPStatus: true});
+			if (!response.ok) {
+				// NOTE: Ignore 404 errors and check the next commit
+				if (response.httpStatus === 404) {
+					continue;
+				}
+
+				return;
+			}
```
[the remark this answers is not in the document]
dertieran: Yes 😄  
I ran into this on the [contributors page](https://github.com/sindresorhus/refined-github/graphs/contributors), but I don't have the commit anymore. 🙈 
My guess would that this could happen through a force push. 🤷‍♂

source/libs/api.ts
fregante: I like your second solution! However it needs to verify that query doesn’t start with `/` because it would skip `api/v3` in that case.

source/features/show-local-user-time.tsx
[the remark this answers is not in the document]
fregante: Force pushes don’t deleted commits, they’re still on GitHub even if they’re no longer directly accessible. 

The only situation where this could happen is if a PushEvent is still listed in the events even if the repo is deleted. My assumption is that after a repo deletion, all of its events are dropped from the list. If my assumption is wrong, then yes 404s are possible in this case.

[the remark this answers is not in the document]
yakov116: I dont think it is.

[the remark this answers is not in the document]
fregante: Indeed. Verified



```yml
$ http 'https://api.github.com/repos/fregante/temp/commits/2b9bc9d637176d6522659c16c15aaa43a41ad047' Accept:application/vnd.github.v3.patch

HTTP/1.1 404 Not Found
Access-Control-Allow-Origin: *
Access-Control-Expose-Headers: ETag, Link, Location, Retry-After, X-GitHub-OTP, X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Reset, X-OAuth-Scopes, X-Accepted-OAuth-Scopes, X-Poll-Interval, X-GitHub-Media-Type
Content-Encoding: gzip
Content-Security-Policy: default-src 'none'
Content-Type: application/json; charset=utf-8
Date: Wed, 27 Nov 2019 00:19:19 GMT
Referrer-Policy: origin-when-cross-origin, strict-origin-when-cross-origin
Server: GitHub.com
Status: 404 Not Found
Strict-Transport-Security: max-age=31536000; includeSubdomains; preload
Transfer-Encoding: chunked
X-Content-Type-Options: nosniff
X-Frame-Options: deny
X-GitHub-Media-Type: github.v3; param=patch
X-GitHub-Request-Id: CC8F:04BD:2253F8:2C58D2:5DDDC107
X-RateLimit-Limit: 60
X-RateLimit-Remaining: 54
X-RateLimit-Reset: 1574817518
X-XSS-Protection: 1; mode=block

{
    "documentation_url": "https://developer.github.com/v3/repos/commits/#get-a-single-commit",
    "message": "Not Found"
}
```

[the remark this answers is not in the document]
fregante: I think only `response.httpStatus === 404` needs to be there though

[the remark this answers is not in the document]
dertieran: Okay I handled this now :+1:

I also exposed the `getError` function from `libs/api` to throw an error when `response.httpStatus !== 404`.

I also assume that any other commit of that `PushEvent` would be gone and `break` out of the loop instead of `continue`.

source/libs/api.ts
dertieran: Added the check :+1: 
*Wasn't sure what the message should be though :smile:*

source/features/show-local-user-time.tsx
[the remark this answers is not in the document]
dertieran: Also added a message when no commit/date was found.

dertieran:
Added the `webext-storage-cache` :+1:
Right now I only cache the `date` string from the patch *(if one was found)*.

fregante:
Thank you for this PR @dertieran! 🕙 

As expected in some cases this triggers a looot of requests, for example `@pento` triggers 44 requests: https://github.com/sindresorhus/refined-github/pulls?utf8=%E2%9C%93&q=is%3Apr+author%3Apento

For this reason caching is vital and there are further improvements possible, if anyone wants to improve it.

- Display progress (for example add a dot for every fetch, `.` `..` `...` `.` `..` etc)
- Pause the loop when the hovercard isn't visible (this might be tricky when coupled with the caching)
- Also display it on profiles like `linkify-user-location`

yakov116:
@dertieran 
```js
content.js:9484 Uncaught (in promise) TypeError: Cannot read property 'append' of null
    at MutationObserver.<anonymous> (content.js:9484)
```
When hovering over the contributors on the main repo page

Looks like [this](https://github.com/sindresorhus/refined-github/blob/master/source/features/user-local-time.tsx#L102)selector needs to be more specific

PR #2551, opened 2019-11-19
Add `emphasize-draft-pr-label` feature

### LINKED ISSUE #2450: Change button colour background to the `Draft` PR label
It's nice to have "labels" on our pull requests, though with a `Draft` pull request this isn't always practical, though when you view a list of PRs the `Draft` PR label/status is easily missed, adding a light grey background colour to the `Draft` text would help make these stand out a little clearer IMHO.

Example URL: https://github.com/WordPress/gutenberg/pulls
Screen Shot 2019-09-20 at 12 47 58 pm

hikerpig:
Closes #2450

# Screenshot

emphasize-draft-pr-label


fregante:
This feature has to be CSS-only, not JS

busches:
The JS is only for easy toggling of the feature, we have others that do this. What's wrong with this one?

fregante:
> The JS is only for easy toggling of the feature, we have others that do this. What's wrong with this one?

This is a very non-breaking feature. We don't have toggles for all the little CSS tweaks and this one isn't one that needs it. JS is just overhead.

hikerpig:
Just thought someone might want to turn this off.
But since it's only a subtle visual change, I agree that we should remove the overhead.

PR #2554, opened 2019-11-19
Add reviewer thumbnails with status

### LINKED ISSUE #2495: add a condensed list of reviewers and status on pulls list
Consider this example at https://github.com/sindresorhus/refined-github/pulls:

image

If we visit the PR, we see:

image

Perhaps instead of saying "Changes Requested" we could show the reviewer name and the a symbol (approved, requested changes, awaiting review) for each reviewer.

Stepping back to the use case, I visit the pulls page to find out the status of PRs I have created.  I have a browser bookmark of pulls created by me.  What I'm typically looking for when I visit this page is information that will help me move my PR towards the merged state.  Typically this means I have to ask myself "do I need to add another reviewer?" or "do I need to nag an existing reviewer?".  A list of pending reviewers and their states would help me answer these questions.  I'm not sure if the same could be said for the list of assignees.  I don't use that feature but I imagine it has the same concern.  Thanks for your consideration.

Example URL:

zadjii-msft:
Adds a simple list of reviewers to each item on the pull requests list, to enable viewing a user's review status at a glance.

Closes #2495

image

# Test
I've been using the following to test:
https://github.com/microsoft/terminal/pulls
https://github.com/microsoft/terminal/issues

### DIFF
--- a/source/content.ts
+++ b/source/content.ts
@@ -153,6 +153,7 @@ import './features/conflict-marker';
 import './features/html-preview-link';
 import './features/linkify-labels-on-dashboard';
 import './features/linkify-user-location';
+import './features/pr-reviewers-thumbnails';
 
 // Add global for easier debugging
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
--- a/source/features/pr-reviewers-thumbnails.tsx
+++ b/source/features/pr-reviewers-thumbnails.tsx
@@ -4,23 +4,17 @@ import features from '../libs/features';
 import {getOwnerAndRepo} from '../libs/utils';
 import * as api from '../libs/api';
 import {appendBefore} from '../libs/dom-utils';
-// import {getUsername} from '../libs/utils';
 
-// function log() {
-//     console.log('✨', <div className="rgh-jsx-element"/>);
-// }
+interface UserInfo {
+    id: number,
+    name: string,
+    state: string
+};
+interface UserStateMap {
+    [id: number]: UserInfo
+};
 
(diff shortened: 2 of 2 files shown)


busches:
For the UI, it's confusing to me now to have the reviewers in one spot and the assigned in the other, especially in your example of `Add supressApplicationTitle as boolean` where the same person is in both. Can we have them in the same spot and highlight assigned reviewers? Maybe denote assigned people without reviews another way?

zadjii-msft:
@busches That's a really good point. Lemme try and work that out, thanks!

fregante:
#2495 was rejected

fregante:
You can publish this as its own extension

zadjii-msft:
@fregante I was hoping to continue the discussion of #2495 with a PR (since discussion in the original thread seemed to die). I believe this is a feature that would be valuable to users, and I'm open to discussing the right way of implementing it to not unnecessarily clutter the UI.

fregante:
> would this be acceptable as an off-by-default feature?

If it's off by default, most users won't see it. We don't need features that most users don't need/see. For that, they can just implemented as their own extension.

Again, not all features need to be part of Refined GitHub, anyone can publish their own features.

zadjii-msft:
Fair point. Would there be any form in which this would be acceptable to add to Refined Github? I think one of the big pros of this extension is that there's a _lot_ of github improvements in here, without needing to chase down extensions for each and every little feature. It's certainly the reason most of our team has adopted it. What about @busches's suggestion of combining it with the assigned-to thumbnails?

PR #2556, opened 2019-11-20
Add `quick-mention` feature

### LINKED ISSUE #1003: Fix the mention autocomplete
See: https://twitter.com/jaffathecake/status/953247205200154624

Not sure what we could do, but we could at least put users that have commented on the issue first, with the author of the last comment at the top.

### LINKED ISSUE #2532: Button to `@` mention in discussion
We could add a button to quickly mention someone. It can either go by the "Edit comment" button or near/below the user/avatar, as such:

Clicking it would instantly `@mention` the user and focus the comment field, like the <kbd>r</kbd> key does when quoting the selection.

This would also be a solution/replacement for https://github.com/sindresorhus/refined-github/issues/1003 and fix the _annoyance_ caused by [`show-names`](https://github.com/sindresorhus/refined-github/issues/2165) (it makes it harder to copy-paste usernames)

fregante:
Closes #2532
Closes #1003

It appears only on hover

# Screenshot

Only one problem: when the comment is near the top of the window, the hovercard appears _below_ the avatar, covering our button.

Suggestions?


sindresorhus:
Clicking the reply button adds a space in front of the username. Can you trim it away?

sindresorhus:
Other than that, it works well 👌

fregante:
What about the overlap with the hovercard?

sindresorhus:
Put the reply button on the left side of the avatar instead?

jerone:
Does the reply button open the hovercard?

fregante:
> Put the reply button on the left side of the avatar instead?

Then it doesn't fit when the window is too small

> Does hovering over the reply button also open the hovercard?

Nope! So I guess that'd make it kinda ok as long as the user doesn't mistakenly hover the avatar on the way there.

source/features/quick-mention.tsx
```diff
@@ -0,0 +1,57 @@
...
+		return false; // Discussion is locked
+	}
+
+	for (const avatar of select.all(`.TimelineItem-avatar > :not([href="/${getUsername()}"]):not(.rgh-quick-mention)`)) {
+		avatar.classList.add('rgh-quick-mention');
+		avatar.after(
+			<button
+				type="button"
+				className="rgh-quick-mention tooltipped tooltipped-e btn-link"
+				aria-label="Mention user in a new comment"
```
yakov116: I think this should say the user name ex: `Mention @fregante in a new comment`

yakov116: @fregante should I open a new issue?

yakov116: @fregante should I open a new issue?

PR #2558, opened 2019-11-20
Missing file extension in screenshot

mobilutz:
I don't think there needs to be a big description for this 😉


fregante:
Thank you ☺️

PR #2559, opened 2019-11-20
Avoid duplicate view switcher in `faster-pr-diff-options`

### LINKED ISSUE #2217: Duplicated unified/split diff buttons
On PRs, it looks like Refined GitHub adds icon buttons for unified/split diffs, but doesn't remove the existing textual buttons:

image

With the extension disabled, the icon buttons don't appear.

arnaudvalle:
The text for data-ga-load is actually slightly different on the two buttons so we need a selector that covers both values

Closes #2217

# Test

- Go to a single commit view (default) (https://github.com/sindresorhus/refined-github/commit/75624f2acf21abf14fa75ea512cfcab42eeb6a2c)
--> the original split/unified buttons should be hidden and the 'icon-only' buttons should be displayed
- Go to a single commit view (split) (https://github.com/sindresorhus/refined-github/commit/75624f2acf21abf14fa75ea512cfcab42eeb6a2c?diff=split)
--> the original split/unified buttons should be hidden and the 'icon-only' buttons should be displayed
- Go to a single commit view (unified) (https://github.com/sindresorhus/refined-github/commit/75624f2acf21abf14fa75ea512cfcab42eeb6a2c?diff=unified)
--> the original split/unified buttons should be hidden and the 'icon-only' buttons should be displayed

### DIFF
--- a/source/features/faster-pr-diff-options.tsx
+++ b/source/features/faster-pr-diff-options.tsx
@@ -80,7 +80,7 @@ function init(): false | void {
 	}
 
 	// Remove previous options UI
-	const singleCommitUI = select('[data-ga-load="Diff, view, Viewed Split Diff"],[data-ga-load="Diff, view, Viewed Unified Diff"]');
+	const singleCommitUI = select('[data-ga-load^="Diff, view"]');
 
 	if (singleCommitUI) {
 		singleCommitUI.remove();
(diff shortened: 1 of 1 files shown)


arnaudvalle:
Right ok I wasn't sure about any other element that may match this as well. I'll change that.

PR #2560, opened 2019-11-21
Add `repo-age` feature

### LINKED ISSUE #2526: Repository creation date
It would be nice if we could have a quick access to the repository creation date (to know if it's a recent or an old project).

Example URL:

github.com/user/repo

maxmclau:
Adds the age of the repository to the end of the stats bar

Closes #2526 

# Screenshot
Screen Shot 2019-11-21 at 3 18 11 AM


source/features/repo-age.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+			<span className="num text-emphasized">{interval}</span> {unit} old
+		</li>
+	);
+
+	if (select.all('.numbers-summary li').length > 4) {
+		select('.numbers-summary li:last-child')!.before(element);
+	} else {
+		select('.numbers-summary li:last-child')!.after(element);
+	}
```
fregante: I think here you can use `appendBefore`, it’s a helper function we wrote. It behaves like “append to number summary, but before the license field if it exists”

source/features/repo-age.tsx
```diff
@@ -0,0 +1,44 @@
...
+}
+
+features.add({
+	id: __featureName__,
+	description: 'Adds the age of the repository to the stats/numbers bar',
+	screenshot: 'https://user-images.githubusercontent.com/3848317/69256318-95e6af00-0bb9-11ea-84c8-c6996d39da80.png',
+	include: [
+		features.isRepo
```
fregante: Too generic, I think it should be `isRepoRoot`

source/features/repo-age.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+	return new Date(repository.createdAt);
+}
+
+async function init(): Promise<void> {
+	const date = await getRepoCreationDate();
+	const {interval, unit} = timeAgo(date);
+	const element = (
+		<li title={`Repository created ${date.toDateString()}`}>
+			{icons.calendar()}
```
fregante: I’m not a fan of this icon, it’s very generic about dates rather than the birth of something. Can we think of an alternative and more specific start/creation icon? 

For example would this make sense? https://octicons.github.com/icon/light-bulb/

source/features/repo-age.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+			<span className="num text-emphasized">{interval}</span> {unit} old
```
fregante: GitHub has their own relative time element and we should use that one instead of having our own “time ago” function

maxmclau: I didn't realize that, sweet. I'm not super familiar with web components - I added the ```time-ago``` time element to the IntrinsicElements in _globals.d.ts_ interface and have it working, but how would you recommend I separate the value into an emphasized <span> so it's 

**1** year ago or **2** hours ago

Should I extend the component?

source/features/repo-age.tsx
maxmclau: Screen Shot 2019-11-21 at 11 50 24 AM

I agree the calendar icon is very generic - I'm not sure which other octicon would be better though. The lightbulb here looks a little off to me.

source/features/repo-age.tsx
fregante: Alright, maybe not. But you can replace the current code with `Intl`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RelativeTimeFormat

It would probably become:

```Js
const [interval, unit] = rtf.format(date).split(' ');
```

maxmclau: As far as I can tell, ```Intl.RelativeTimeFormat()``` is only for language localization of the relative time units. You can't pass it a date to compute difference, instead you pass it an already decided interval and unit of time so I'll still need some kind of interval difference logic.

fregante: D'oh, indeed. I just wonder if there's a way to shorten that code (`timeAgo`)

I'll look into it

source/libs/utils.ts
```diff
@@ -138,3 +138,26 @@ export function reportBug(featureName: string, bugName: string): void {
...
+		}
+	}
+
+	const interval = Math.round(diff / timespan);
+	if (interval > 1) {
+		unit += 's';
+	}
+
+	return {interval, unit};
+}
```
fregante: How about this?

```js
function timeAgo (date: Date): [number | string, string?] {
    const today = new Date();
    let difference = (today.getTime() - date.getTime()) / 1000 / 3600 / 24; // In days
    let unit = 'day';
    if (difference > 360) {
        difference /= 365;
        unit = 'year';
    } else if (difference > 30) {
        difference /= 30;
        unit = 'month';
    } else if (date.getDate() === today.getDate()) {
        return ['today'];
    } else if (date.getDate() === today.getDate() - 1) {
        return ['yesterday'];
    }
    
    return [
        Math.floor(difference),
        difference < 2 ? unit : unit + 's'
    ];
}
```

Test the output here: https://runkit.com/embed/b7khixb3zrzu

Almost the same function length but accounts for today/yesterday/months and drops seconds/minutes/hours

source/libs/icons.tsx
```diff
@@ -56,6 +56,8 @@ export const gear = (): SVGElement => <svg aria-hidden="true" className="octicon
 
 export const foldDown = (): SVGElement => <svg aria-hidden="true" className="octicon octicon-fold-down" width="14" height="16" ><path fill-rule="evenodd" d="M4 11l3 3 3-3H8V5H6v6H4zm-4 0c0 .55.45 1 1 1h2.5l-1-1h-1l2-2H5V8H3.5l-2-2H5V5H1c-.55 0-1 .45-1 1l2.5 2.5L0 11zm10.5-2H9V8h1.5l2-2H9V5h4c.55 0 1 .45 1 1l-2.5 2.5L14 11c0 .55-.45 1-1 1h-2.5l1-1h1l-2-2z"/></svg>;
 
+export const calendar = (): SVGElement => <svg aria-hidden="true" width="18" height="16" className="octicon octicon-calendar"><path fill-rule="evenodd" d="M13 2h-1v1.5c0 .28-.22.5-.5.5h-2c-.28 0-.5-.22-.5-.5V2H6v1.5c0 .28-.22.5-.5.5h-2c-.28 0-.5-.22-.5-.5V2H2c-.55 0-1 .45-1 1v11c0 .55.45 1 1 1h11c.55 0 1-.45 1-1V3c0-.55-.45-1-1-1zm0 12H2V5h11v9zM5 3H4V1h1v2zm6 0h-1V1h1v2zM6 7H5V6h1v1zm2 0H7V6h1v1zm2 0H9V6h1v1zm2 0h-1V6h1v1zM4 9H3V8h1v1zm2 0H5V8h1v1zm2 0H7V8h1v1zm2 0H9V8h1v1zm2 0h-1V8h1v1zm-8 2H3v-1h1v1zm2 0H5v-1h1v1zm2 0H7v-1h1v1zm2 0H9v-1h1v1zm2 0h-1v-1h1v1zm-8 2H3v-1h1v1zm2 0H5v-1h1v1zm2 0H7v-1h1v1zm2 0H9v-1h1v1z"></path></svg>;
```
fregante: This should work too

```suggestion
export const calendar = (): SVGElement => <svg aria-hidden="true" width="18" height="16" className="octicon octicon-calendar"><path d="M13 2h-1v1.5c0 .28-.22.5-.5.5h-2c-.28 0-.5-.22-.5-.5V2H6v1.5c0 .28-.22.5-.5.5h-2c-.28 0-.5-.22-.5-.5V2H2c-.55 0-1 .45-1 1v11c0 .55.45 1 1 1h11c.55 0 1-.45 1-1V3c0-.55-.45-1-1-1zm0 12H2V5h11v9zM5 3H4V1h1v2zm6 0h-1V1h1v2zM6 7H5V6h1v1zm2 0H7V6h1v1zm2 0H9V6h1v1zm2 0h-1V6h1v1zM4 9H3V8h1v1zm2 0H5V8h1v1zm2 0H7V8h1v1zm2 0H9V8h1v1zm2 0h-1V8h1v1zm-8 2H3v-1h1v1zm2 0H5v-1h1v1zm2 0H7v-1h1v1zm2 0H9v-1h1v1zm2 0h-1v-
1h1v1zm-8 2H3v-1h1v1zm2 0H5v-1h1v1zm2 0H7v-1h1v1zm2 0H9v-1h1v1z"/></svg>;
```

source/libs/utils.ts
maxmclau: My original timeAgo utility accounted for those descriptions of the time like 'just now/yesterday/last year', because I wanted to be consistent with how Github handled it. I ended up going back to it always being ```[interval] [unit] old``` after playing around with it because I thought it looked more like a metric in the bar that changes and less like the License.

It's another minor point though and the extension styling is up to you, so let me know

Screen Shot 2019-11-22 at 12 28 23 AM
Screen Shot 2019-11-22 at 12 29 29 AM
Screen Shot 2019-11-22 at 12 33 24 AM

yakov116:
I think there should be tooltip since it's not obvious what was ' 21hours ago '

source/libs/utils.ts
sindresorhus: Why not just the native browser API? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RelativeTimeFormat

yakov116: https://github.com/sindresorhus/refined-github/pull/2560#discussion_r349017077

maxmclau: > [#2560 (comment)](https://github.com/sindresorhus/refined-github/pull/2560#discussion_r349017077)

That link points to the icon thread, here's the one about ```Intl.RelativeTimeFormat```.

https://github.com/sindresorhus/refined-github/pull/2560#discussion_r349205427

fregante: If our intent is to save some code, I think this is the shortest we can do:

```tsx
// https://github.com/github/time-elements#time-ago
function timeAgo (date: Date): string {
	return (<time-ago datetime={date.toISOString()}/>).textContent;
}

// Replace 'ago' with 'old' if present
// Display time in bold
const [time, ago] = timeAgo(date).split(/(?=ago)/);
<span><b>{time}<b> {ago ?? 'old'}<span>
```

fregante:
> I think there should be tooltip since it's not obvious what was ' 21hours ago '

We don't show `ago` but `old`, which would be self-explanatory in _the repo’s header._ As long as `old` is always visible, the tooltip isn't necessary.

If there's a chance that our string doesn't include `old` ("Just now", "Today", etc), then yes we need `tooltipped` + `aria-label`

source/features/repo-age.tsx
fregante: Would the regular repo icon be helpful?

```
[repo icon] 2 years old
```

source/features/repo-age.tsx
fregante: Or if you're targeting the license field, perhaps it's better to look for it specifically. Other extensions might add information to this bar and could throw our count-based solution off.

source/libs/utils.ts
maxmclau: I played around with it a little more and I think forcing the output to always be ```interval``` ```unit``` **```old```** is pretty key to preventing confusion about what the metric is telling you. Another option I see is to prepend a word like **created** for ```<time-ago>``` returns now/today/last year...

Alternatively here's an implementation that will always force the ```interval``` ```unit``` **```old```** format using the [micro](https://github.com/github/time-elements#micro-format) return (1m, 1h, 1d, 1y), though it's maybe a little complicated.

```typescript
function timeAgo(date: Date): {interval: number; timespan: string} {
	const units = ['minute', 'hour', 'day', 'year']

	const ago = (<time-ago datetime={date.toISOString()} format="micro"/>).textContent; // 1m, 1h, 1d, 1y.
	const [interval, short] = ago!.match(/[a-z]+|[^a-z]+/gi)! as [number, string];

	let timespan = units.find(unit => unit.slice(0, 1) == short)!;
	timespan = interval > 1 ? timespan + 's' : timespan

	return {interval, timespan};
}
```

Really didn't expect there to be so many ways to skin this cat, gee. :see_no_evil:

source/features/repo-age.tsx
maxmclau: Screen Shot 2019-11-23 at 2 56 31 AM

I don't mind it myself and I don't see many other options unless Github adds a 🎂 octicon tomorrow.

source/libs/utils.ts
fregante: Not sure I like the short time and why are you adding s to the short time? This turns 2m into 2ms (milliseconds)

fregante: Doesn’t `time-ago` **always** include `ago` already?? From the docs it appears that it only doesn’t when it’s `just now`, which I don’t really care for since nobody will ever see it.

source/features/repo-age.tsx
fregante: Works for me

source/libs/utils.ts
maxmclau: I though that was the case from the docs, but looking at how it's [actually implemented](https://github.com/github/time-elements/blob/master/src/relative-time.js#L207) it can return things like last week/today/yesterday. It's also important to handle the ```just now``` possibility since you run into this right after you fork a repo. Side note, the docs say it'll return ```just now```, but it only returns ```now```.

sindresorhus:
Looks good.

@maxmclau Can you add it to the readme?

fregante:
Thanks for this feature @maxmclau! 🎂

source/features/repo-age.tsx
```diff
@@ -0,0 +1,49 @@
...
+import {getRepoGQL} from '../libs/utils';
+
+async function getRepoCreationDate(): Promise<Date> {
+	const {repository} = await api.v4(`
+		repository(${getRepoGQL()}) {
+			createdAt
+		}
+	`);
+
+	return new Date(repository.createdAt);
```
fregante: @maxmclau could you add long-term caching to this value?
Something like

```js
const key = __featureName__ + ':' + getRepoURL();
if (await cache.has(key)) {
	return cache.get(key);
}

	repository(${getRepoGQL()}) {
		createdAt
	}
`);

cache.set(key, repository.createdAt);
return repository.createdAt;
```

fregante: #2578 merged

maxmclau: @fregante this memoization implementation is pretty neat - I opened a PR that should close this.

PR #2584, opened 2019-12-02
Disable `linkify-user-location` on own editable profile

### LINKED ISSUE #2581: `linkify-user-location` breaks location edits in profile
1. Yes, works correctly without extension (or disabled `linkify-user-location` feature)
2. "Your profile" page. Click "Edit Profile" button.
3. 
image

As `linkify-user-location` wraps underlying `<input>` inside `<a>` it's impossible to click that input field and instead you get navigated to google maps.

Relates: #2544

romansp:
Closes #2581.

I chose slightly different approach: instead of `except: [features.isUserOwnProfile]` it will skip editable nodes by using existing `isEditable` helper. That's because there are two sets of DOM nodes (display and form) on user profile and "Edit profile" button just toggles between those two.

This way displayed location will remain linkified, but editable field will be kept intact.

### DIFF
--- a/source/features/linkify-user-location.tsx
+++ b/source/features/linkify-user-location.tsx
@@ -1,11 +1,15 @@
 import React from 'dom-chef';
 import select from 'select-dom';
-import {wrap} from '../libs/dom-utils';
+import {wrap, isEditable} from '../libs/dom-utils';
 import features from '../libs/features';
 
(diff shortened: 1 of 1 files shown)


fregante:
Thank you for contributing this bugfix and justifying your choice Roman!

Issue #2530, opened 2019-11-11
Console Errors


yakov116:
- [x]  https://github.com/sponsors/sindresorhus

```js
❌ forked-to
content.js:2856 TypeError: Cannot read property 'addEventListener' of null
    at forked_to_init (content.js:7820)
    at features_run (content.js:2851)
    at content.js:2898
    at content.js:2813
    at onAjaxedPagesRaw (content.js:2808)
```
```js
❌ more-dropdown
content.js:2856 TypeError: Cannot read property 'append' of null
    at appendBefore (content.js:3537)
    at createDropdown (content.js:3565)
    at more_dropdown_init (content.js:3578)
    at async features_run (content.js:2851)
```
```js
❌ edit-files-faster
content.js:2856 Error: Unable to fetch.
Ensure that your token has access to this repo.
{
	"message": "Not Found",
	"documentation_url": "https://developer.github.com/v3/repos/#get"
}
    at getError (content.js:3063)
    at async chrome-extension:/hlepfoohegkhhmjieoechaddaejaokhf/content.js:3025
    at async fetchFromApi (content.js:3078)
    at async get_default_branch (content.js:3086)
    at async edit_files_faster_init (content.js:6973)
    at async features_run (content.js:2851)
```
- [x]  Click on the sponsor button in a repo or the watch dropdown
```js
content.js:4502 Uncaught TypeError: Failed to execute 'observe' on 'IntersectionObserver': parameter 1 is not of type 'Element'.
    at HTMLDocument.<anonymous> (content.js:4502)
    at HTMLDocument.listenerFn (content.js:369)
```

- [ ] https://gist.github.com/yakov116
```js
❌ embed-gist-via-iframe
content.js:2856 TypeError: Cannot read property 'cloneNode' of null
    at embed_gist_via_iframe_init (content.js:8398)
    at features_run (content.js:2851)
    at Object.add (content.js:2896)
```

busches:
Once this is merged, we can update our dependency: https://github.com/Mottie/github-reserved-names/pull/8

yakov116:
@busches it was

fregante:
Fix for the last one: `embed-gist-via-iframe` needs to be limited to actual gists:

Ok: https://gist.github.com/yakov116/2d81580f7907839f9065e670f0aba64c
Not ok: https://gist.github.com/yakov116

cc @kidonng

fregante:
[...]

PR #2592, opened 2019-12-05
Bump dependencies

busches:
This bumps almost all the dependencies, except for Stylelint and the two XO configs.

Bumping `github-reserved-names` to 1.1.7 fixes one of the errors called out here: https://github.com/sindresorhus/refined-github/issues/2530 when you visit a sponsor page, e.g. https://github.com/sponsors/sindresorhus


fregante:
Can you update a few more dependencies if minor so this can be a routine “Update dependencies”?

source/libs/utils.ts
```diff
@@ -138,3 +138,5 @@ export function reportBug(featureName: string, bugName: string): void {
 	console.log('Find existing issues:\n' + String(issuesUrl));
 	console.log('Open new issue:\n' + String(newIssueUrl));
 }
+
+export const noOp = (): void => {}; // eslint-disable-line @typescript-eslint/no-empty-function
```
sindresorhus:
```suggestion
export const noop = (): void => {}; // eslint-disable-line @typescript-eslint/no-empty-function
```

It's more common to write it like this.

https://api.jquery.com/jQuery.noop/

fregante: This is the smallest non-optimization ever. Can we just leave it inlined?

busches:
Lol, I didn't do it as an optimization, but to only have to do the eslint
rule suppression once.

fregante:
Yeah I didn’t mean it for you, but in general I haven’t seen a `noop` function declared in so long since arrow functions made it so short. The rule can be disabled in package.json

Issue #2604, opened 2019-12-07
Make Refined GitHub act sooner with `element-ready`


fregante:
The browser may display content on screen long before `dom-ready`, so sometimes content appears for a few seconds before Refined GitHub changes it. This is a bit annoying.

https://github.com/sindresorhus/element-ready can avoid this by acting on elements as soon as they appear in the DOM while the page is still loading/streaming.

GitHub sometimes loads new pages via ajax so we have to **listen to both the real `dom-ready` event and the ajaxed one**. We handle this with `features.onAjaxedPages`

We use `element-ready` in some features but there's currently no loading mechanism that combines `element-ready` and ajaxed loads.

## Possible solutions

Here's some implementations I thought of.

### A new single loading method

```js
feature.init({
	...,
	load: features.onAjaxedPagesOrWait('selector')
});
```

Better yet, have a `onAjaxedPages` version that doesn't wait for `dom-ready` on the first run and `element-ready` can be used in the feature itself.

**✅ Preferred**
```js
feature.init({
	...,
	load: features.onAjaxedPagesOrInstant
});
```

### A completely new loading API

This would split each event and allow us to combine them in any way, but it's a bit verbose

```js
feature.init({
	...,
	loadOnAjaxed: true,
	loadAfterElement: 'selector'
});
```

this means regular features (`dom-ready` + ajaxed) would have to be declared explicitly:

```js
feature.init({
	...,
	loadOnDomReady: true,
	loadOnAjaxed: true
});
```

### ... other suggestions welcome

PR #2605, opened 2019-12-07
Avoid FOUC on some features

fregante:
Fixes https://github.com/sindresorhus/refined-github/issues/2604

Naming suggestions welcome. Perhaps `features.onAjaxedPagesInstant` (like the existing `features.onAjaxedPagesRaw`)

# Before

This shows a page refresh. Notice how `releases-tab`, `more-dropdown` and `forked-to` are not there for a moment before `dom-ready` happens.

> before

# After

Notice how `parse-backticks` still has a FUOC (Flash Of Unstyled Content) but it no longer happens for the 3 features above.

> after


liow788:
Can directly to my bank account

PR #2610, opened 2019-12-09
Use `@primer/octicons` package instead of curating icons.tsx

### LINKED ISSUE #1752: Use package octicons instead of including icons.tsx
https://www.npmjs.com/package/octicons

Import as plain text:

```js
import doma from 'doma';
import alertIcon from 'octicons/build/svg/alert.svg';

document.body.append(doma.one(alertIcon))
```

Or as JSX with https://github.com/janjakubnanista/svg-jsx-loader

```js
import alert from 'octicons/build/svg/alert.svg';

document.body.append(alert());
```

There's also the official `@githubprimer/octicons-react` package but I don't think it's being built+published correctly yet.

Also possibly: https://github.com/siddharthkp/jsxo but it needs Babel

fregante:
Closes #1752 

The only issue I think is that this throws off the sourcemap (untested) and I'm not sure of how to fix that. Perhaps https://github.com/Rich-Harris/magic-string


fregante:
The sourcemaps seem to continue working, as long as you don't step into a `*Icon()` function, which is mostly useless anyway since they're plain `dom-chef` calls.

PR #2626, opened 2019-12-12
Small meta changes

fregante:
I will also disable sourcemaps (in the published bundles) because Store reviewers don't care about them and because we don't include the source, so sourcemaps are not useful there.

Let me know if I'm wrong


fregante:
I ended up leaving the sourcemaps because they do work in the browser

fregante:
Notice: once you check out the latest version and run it in the browser, the options will be compressed. This means you can't try an older version of the extension without losing the options.

Info: https://github.com/fregante/webext-options-sync/pull/40#issuecomment-565834213

Issue #2490, opened 2019-10-15
Ability to out-dent text selection.


itsjamie:
Original GIF: https://user-images.githubusercontent.com/1402241/33802977-beb8497c-ddbf-11e7-899c-698d89298de4.gif

I thought there was a bug where I assumed folks were originally Shift+Tab to outdent the content here and it wasn't working for me, but I assume they were now just Ctrl-Zing after reading other reports. Looked at source, and there is no handling for Shift+Tab. 

It would be really nice honestly to have a paste option standpoint of pasting in code from an editor and you end up with code with leading spaces, to be able to "match" the indentation of the first line for all other lines.

<img width="355" alt="Screenshot 2019-10-14 23 43 55" src="https://user-images.githubusercontent.com/1956521/66795958-9023ec80-eedc-11e9-9922-eef22c7908c3.png">

However, I can see where that would be difficult to "do the right thing", and so having the ability to manually correct this, would be awesome!

```
const tracksList = this.tracks.map((track) => {
          return {
            _id: track.id.toString(),
            label: track.name,
            kind: track.type.toLowerCase(),
            default: track.default
          };
        });
```

fregante:
This is has been haphazardly implemented a while ago but I never cleaned it up https://github.com/fregante/indent-textarea/pull/6

fregante:
Closing as duplicate of https://github.com/fregante/indent-textarea/issues/1

PR #2627, opened 2019-12-12
Enable `shift-tab` to unindent comment fields

fregante:
Closes https://github.com/sindresorhus/refined-github/issues/2490

Finally I completed https://github.com/fregante/indent-textarea/pull/6 so `tab-to-indent` now also includes `shift-tab` support (as well as a `tab` behavior that matches editors more closely)

Have a look and see if it behaves correctly. I modeled it after VS Code. Sublime Text seems to **add** a `tab` with `shift-tab` if no text is selected. Weird.


sindresorhus:
The behavior when just with whitespace is weird:


It skips going back to the start of the second line.

fregante:
Good catch! This is one of those edge cases I was hoping to find before merging. 

```ts
// `|` is the cursor
undindent('\n|\t')

	expected: '\n|'
	actual: '|\n'
```

fregante:
This is harder than I thought.

Anyway.

`unindent` or `outdent`? VSCode seems to use [`outdent`](https://github.com/microsoft/vscode/blob/4ebf8ee0b153e0420eed6950cdb9c33f95120162/src/vs/editor/common/commands/shiftCommand.ts) but elsewhere I found `unindent`

fregante:
This should be fixed now.

PR #2628, opened 2019-12-12
Restore `view-markdown-source` buttons

### LINKED ISSUE #2619: `view-markdown-source` is not being loaded on `*.md` pages
When I load https://github.com/fastify/fastify/blob/master/docs/Getting-Started.md with `view-markdown-source` and "Show the features enabled on each page in the console" enabled, I can see that the option is not loaded. This is happening on any markdown document in that repository, as well as on other repositories.

Extension version: 19.11.30

arnaudvalle:
Closes #2619 

# Test

- Go to some markdown file (i.e. https://github.com/sindresorhus/refined-github/blob/master/contributing.md)
--> Notice the two buttons to display the source and rendered markdown are displayed (1st set of buttons on the screenshot just below)
Screenshot 2019-12-12 22 39 00
- Click on the first button
--> This should show the source of the markdown file
- Click on the second button
--> This should show the rendered markdown file


arnaudvalle:
> It looks like the class for the replacement isn't right, there's some padding around the code

Well spotted! Should be fixed now.

fregante:
Thank you for the fix, Arnaud!

Issue #1247, opened 2018-04-05
Old events don't work on history.back()


fregante:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%2480%20Funded-%2300A156.svg" />](https://issuehunt.io/r/sindresorhus/refined-github/issues/1247)

Apparently GitHub caches but regenerates the DOM between ajax loads. This means that our markup is still there, but no events are attached to it.

I think in many cases they still work because we use global delegated events, but it doesn't where we use [`onClick`](https://github.com/sindresorhus/refined-github/blob/4ca173cbd93862a518577acd58796c9d221802ed/source/features/open-all-selected.js#L35) and [`addEventListener`](https://github.com/sindresorhus/refined-github/blob/4ca173cbd93862a518577acd58796c9d221802ed/source/features/display-issue-suggestions.js#L81)

I don't think it's a consequence of #1175 because before it, the old button would still be there and inactive.

---

To reproduce the bug:
1. Visit https://github.com/sindresorhus/refined-github/issues
2. Visit the "Code" tab
5. Press the "Back" browser button
6. Click the "Open all (issue)" link above the issue list. It doesn't work anymore

---

This is rather annoying, the only solution I can think of is to only use delegated events. Not great.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $80.00)

- [<img src='https://avatars3.githubusercontent.com/u/44827199?v=4' alt='issuehunt' width=24 height=24> issuehunt](https://issuehunt.io/u/issuehunt) ($80.00)

### Submitted pull Requests
- [#2639 Fix back button](https://issuehunt.io/r/sindresorhus/refined-github/pull/2639)
---

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/sindresorhus/refined-github/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
---
IssueHunt has been backed by the following sponsors. [Become a sponsor](https://issuehunt.io/membership/members)
</details>

IssuehuntBot:
@issuehunt has funded $80.00 to this issue.

---
- Submit pull request via [IssueHunt](https://issuehunt.io/repos/51769689/issues/1247) to receive this reward.
- Want to contribute? Chip in to this issue via [IssueHunt](https://issuehunt.io/repos/51769689/issues/1247).
[...]

PR #2639, opened 2019-12-18
Keep features working after clicking the browser’s Back button

fregante:
Fixes https://github.com/sindresorhus/refined-github/issues/1247

Untested. If it works thank TypeScript and XO.

GitHub uses delegated events across the board and we have to do the same because PJAX either flattens the page to plain HTML or uses `cloneNode`, both of which preserve changes we make (and buttons we add) but discard attached events.

# Test

All the related features 😅

- Test normal function
- Open page, then navigate away (to another ajaxed page), click back

It would be great of anyone could help test this out and list which features you tried

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#1247: Old events don't work on history.back()](https://issuehunt.io/repos/51769689/issues/1247)
---

</details>


yakov116:
@fregante if you upload a CRX I can test it

fregante:
The Build workflow has an artifact. Look in the dropdown on the top right of Actions

fregante:
Current version: https://github.com/sindresorhus/refined-github/suites/365152039/artifacts/751370

fregante:
Tested and working, both normally and after clicking the browser's "back" button

source/features/hide-comments-faster.tsx
source/features/batch-open-issues.tsx

fregante:
I manually tested/fixed every feature with the exception of a couple rare instances for `on-new-comments` and `on-pr-merge-panel-open`

Time to merge this and fix this behavior once for all!

PR #2653, opened 2019-12-23
Enable `parse-backticks` on commit titles on more pages

samuelmeuli:
Currently, text between backticks is only highlighted in commit messages in the commit list. This PR applies the same styles to the latest commit on the repository start page:

screenshot 2019-12-23 at 17 12 59


source/features/parse-backticks.tsx
```diff
@@ -18,7 +18,8 @@ features.add({
 	screenshot: 'https://user-images.githubusercontent.com/170270/55060505-31179b00-50a4-11e9-99a9-c3691ba38d66.png',
 	include: [
 		features.isDiscussionList,
-		features.isCommitList
+		features.isCommitList,
```
fregante: I think isCommitList is already included in isRepo, but I wonder if isRepo is too broad. `include` should be as specific as possible

source/features/parse-backticks.tsx
```diff
@@ -6,7 +6,7 @@ import {parseBackticks} from '../libs/dom-formatters';
 function init(): void {
 	for (const title of select.all(`
 		[aria-label="Issues"][role="group"] .js-navigation-open,
-		.commit-title .js-navigation-open
+		.message
```
fregante: This seems rather generic. Needs testing to make sure it doesn’t target unrelated elements

source/features/parse-backticks.tsx
samuelmeuli: You're right, I'll update this :)

source/features/parse-backticks.tsx
samuelmeuli: I suspect the `message` class is the one we should use. It will e.g. also get the code styling to work in the file list:

image

source/features/parse-backticks.tsx
```diff
@@ -6,7 +6,7 @@ import {parseBackticks} from '../libs/dom-formatters';
 function init(): void {
 	for (const title of select.all(`
 		[aria-label="Issues"][role="group"] .js-navigation-open,
-		.commit-title .js-navigation-open
```
yakov116: Why did you remove this line? This is still needed

source/features/parse-backticks.tsx
samuelmeuli: See 51fdb91

source/features/parse-backticks.tsx
samuelmeuli: It should be covered by the `.message` selector

yakov116: Nope. This is for
image

Which does not have such a class

samuelmeuli: It doesn't have the `.commit-title` class either. Isn't this covered by `source/features/format-discussion-titles.tsx`?

yakov116: You are correct

samuelmeuli:
Code formatting could be added in a few more places:

- Latest commit on repo start page (https://github.com/sindresorhus/refined-github) – `commit-tease` or `message` class:

image

- File list on repo start page (https://github.com/sindresorhus/refined-github) – `message` class:

image

- Single commit page (e.g. https://github.com/sindresorhus/refined-github/commit/3bf451f88360b5f1b898f636542a443ea2b687d2) – `commit-title` class:

image

source/features/parse-backticks.tsx
fregante: I think you’re looking for isRepoTree

source/features/parse-backticks.tsx
fregante: Given the `js-navigation-open` class id guess that was for isCommitList

samuelmeuli:
Updated to cover the aforementioned use cases.

source/features/parse-backticks.tsx
```diff
@@ -6,19 +6,36 @@ import {parseBackticks} from '../libs/dom-formatters';
...
+		features.isRepoTree,
+
+		// Commit list: ".message" and ".commit-desc" selectors
+		features.isCommitList,
+
+		// Single commit view: ".commit-title" and ".commit-desc" selectors
+		features.isCommit,
+
+		// Single file view: '.Box--condensed .link-gray[href*="/commit/"]' selector
+		features.isSingleFile
```
fregante: If you want to comment on each selector, use this an array as selector:

```js
select.all([
  '.commit-title', // `isCommit`
  '.commit-desc', // `isCommit` and `isCommitList`
  '.message' // etc
]);
```
```

samuelmeuli: Nice, didn't know this feature 👍

fregante:
This looks great, thank you!

fregante:
There's one more place: https://github.com/sindresorhus/refined-github/issues/2671#ref-pullrequest-546231073

PR #2659, opened 2019-12-29
Restore `tags-dropdown` link to open tag pages

### LINKED ISSUE #2658: Tags dropdown is broken on releases page
It redirects to the repo's code tab rather than redirecting to the tag page. 

**Steps to reproduce**:
1. Open any repo's releases page. (e.g. https://github.com/sindresorhus/refined-github/releases)
2. Open the "Select tag" dropdown and click on any tag. (e.g. Selecting 19.12.22)
3. Observer the behavior, it redirects to the repo's code tab (https://github.com/sindresorhus/refined-github/tree/19.12.22), but it should redirect to https://github.com/sindresorhus/refined-github/releases/tag/19.12.22 page.

hardikmodha:
Closes #2658 

# Test
1. Open any repo's releases page. (e.g. https://github.com/sindresorhus/refined-github/releases)
2. Open the "Select tag" dropdown and click on any tag. (e.g. Selecting 19.12.22)
3. Observer the behavior, it should redirect to the tag page. (https://github.com/sindresorhus/refined-github/releases/tag/19.12.22)


source/features/tags-dropdown.tsx
```diff
@@ -37,12 +37,16 @@ function onFragmentLoaded(): void {
...
 
-	// Change links, which point to the content of each tag, to open the tag page instead
-	for (const anchorElement of select.all<HTMLAnchorElement>('.rgh-tags-dropdown [href*="/tree/"]')) {
-		const pathnameParts = anchorElement.pathname.split('/');
-		pathnameParts[3] = 'releases/tag'; // Replace `tree`
-		anchorElement.pathname = pathnameParts.join('/');
-	}
+	// https://github.com/github/remote-input-element#events
+	// Wait until the network request is finished and HTML body is updated
+	select('.rgh-tags-dropdown remote-input')!.addEventListener('remote-input-success', () => {
```
fregante: Can you use `one-event.ts` instead?

hardikmodha: Good suggestion but `remote-input` also fetches new tags as we type in the input box so this callback should be invoked every time tags are fetched. Making to run it only one time will fail to modify links on the newly fetched data.

jenniferbillings: Hfc

fregante:
This feature used to be handled by `onFragmentLoaded`. If that doesn't work, it needs to be removed; perhaps you can listen to the event before the fragment loads, so you don't have to listen to `load`

hardikmodha:
> This feature used to be handled by onFragmentLoaded. If that doesn't work, it needs to be removed; perhaps you can listen to the event before the fragment loads, so you don't have to listen to load
 
Sorry, but I didn't get this. It is being handled by `onFragmentLoaded` right now and it is working.

fregante:
I mean the `load` event handler was used to detect the updated list. Now the list update is handled by `remote-input-success`, so `load` is no longer necessary: they're there for the same purpose.

hardikmodha:
I believe, we still need this `load` event handler because the `remote-input`, tags list is only available after when this load event handler is fired.

fregante:
If the event bubbles, we don't have to listen to it specifically on `remote-input`

hardikmodha:
Got that. The `remote-input-success` event being bubbled https://github.com/github/remote-input-element/blob/master/src/index.ts#L122 but still, we'll need `load` event to change the tab from "Branch" to "Tags". https://github.com/sindresorhus/refined-github/blob/master/source/features/tags-dropdown.tsx#L38. 

A relevant question, Is there any disadvantage of listening it specifically on `remote-input` element?

fregante:
>  Is there any disadvantage of listening it specifically on `remote-input` element?

That's irrelevant, my suggestion was to drop code that was no longer necessary (i.e. listening to the `load` event)

> we'll need `load` event to change the tab from "Branch" to "Tags".

If that's the case, `onFragmentLoaded` can stay, but I'd still like the handlers to be separated and storing the `remote-input-success` handler function on the top level.

Something like:

```ts
function init(): void {
	if (select.exists('.blankslate')) {
		return false;
	}

	select('.subnav')!.append(
		// ...
	);

	// https://github.com/github/remote-input-element#events
	// Wait until the network request is finished and HTML body is updated
	select('.rgh-tags-dropdown')!.addEventListener('remote-input-success', onRemoteInput);
}

function onFragmentLoaded(): void {
	// Change the tab to "Tags"
	select('.rgh-tags-dropdown .select-menu-tab:last-child button')!.click();
}

function onRemoteInput(): void {
	// Change links, which point to the content of each tag, to open the tag page instead
	for (const anchorElement of select.all<HTMLAnchorElement>('.rgh-tags-dropdown .select-menu-list:last-child [href*="/tree/"]')) {
		const pathnameParts = anchorElement.pathname.split('/');
		pathnameParts[3] = 'releases/tag'; // Replace `tree`
		anchorElement.pathname = pathnameParts.join('/');
	}
}
```

fregante:
The two handlers could also be renamed to what they actually do since they don't actually handle the event; e.g. `useTagsTab` and `updateLinksToTags` respectively

fregante:
Thanks Hardik 🌶

source/features/tags-dropdown.tsx
```diff
@@ -23,22 +23,29 @@ function init(): false | void {
...
 					</include-fragment>
 				</details-menu>
 			</details>
 		</div>
 	);
+
+	// https://github.com/github/remote-input-element#events
+	// Wait until the network request is finished and HTML body is updated
+	// "remote-input-success" event is bubbled
+	select('.rgh-tags-dropdown')!.addEventListener('remote-input-success', updateLinksToTag);
```
fregante, line 37: Support for `on-remote-input-success` JSX attributes [has been added](https://github.com/vadimdemedes/dom-chef/pull/71), in case you want to move this.

PR #2662, opened 2020-01-02
Infinite scroll in commit lists

Adk9p:
Just going to preface this with I don't know a lot about javascript and this is my first time touching typescript. This is less of a contribution and more a an outline for an idea.

The code isn't test it's just a mvp to show off the feature. 

Just wanted to see if it was a feature you guys would consider adding.

### Test

https://github.com/sindresorhus/refined-github/commits/master


yakov116:
@Adk9p you can do xo --fix

Adk9p:
@yakov116 I don't know what that means... :cry:

Adk9p:
Playing around with it I noticed two bugs
1. `Commits on %Date%` show up twice
2. Sometimes it just fails, and once you reach the end of the page nothing happens

fregante:
`fetch-dom.ts` is a module we have. Look under `source/libs`

Avoid `innerHTML`: move the DOM from the fetched page and use JSX instead of the inline HTML string. 

Use modern DOM APIs like `.before` instead of `parentNode.insertBefore`

Adk9p:
@fregante I communicated the bug poorly, it's decided on page load. That was what I was trying to say by
> This only happens once, If you reach the end of the page and it loads it will continue to load

I don't know what causes it but an easy way to trigger is it to go from
Commit-list > Code > Commit-list

it might have something to do with constants in the code? I have no clue.

Adk9p:
Code now breaks once you get down to `Commits on Apr 25, 2019`
As well, `Commit-list > Code > Commit-list` still breaks it

fregante:
Ah yes, it’s because you need `load: features.onAjaxedPages` not `onDomReady`

source/features/infinite-commit-list-scroll.tsx
```diff
@@ -0,0 +1,92 @@
...
+const githubLoadingButton = <button className="btn mn-3" disabled><span>Loading</span><span className="AnimatedEllipsis"></span></button>;
+
+let container: HTMLElement;
+let link: HTMLAnchorElement | undefined;
+
+let lastDate: string;
+
+function getLastDate(document_?: HTMLDivElement): string {
+	if (!document_) {
+		document_ = select<HTMLDivElement>('.commits-listing')!;
```
fregante: Are you sure you need this? It think it should just default to `document`

source/features/infinite-commit-list-scroll.tsx
```diff
@@ -0,0 +1,92 @@
...
+async function appendOlder(): Promise<void> {
+	// Don't load if we don't have the link
+
+	if (!link?.href) {
+		return;
+	}
+
+	// Replace buttons with loading button
+	select('*', container)!.remove();
+	container.append(githubLoadingButton);
```
fregante: Rename `container` to `pagination`

Use `pagination.replaceWith(githubLoadingButton)`

Later you can use the loading button as insertion point:

```is
githubLoadingButton.before(…)
githubLoadingButton.replaceWith(…)
```

source/features/infinite-commit-list-scroll.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+	// Fetch older content
+	const olderContent = await fetchDom(link.href, '.repository-content')!;
+	if (!olderContent) {
+		console.log('Failed to fetch next page');
+		return;
+	}
```
fregante: No need for this. If it fails, it will just show an error in the console

source/features/infinite-commit-list-scroll.tsx
```diff
@@ -0,0 +1,92 @@
...
+		appendOlder();
+	}
+}, {
+	rootMargin: '500px'
+});
+
+function init(): void {
+	lastDate = getLastDate();
+
+	container = select<HTMLDivElement>('.paginate-container')!;
```
fregante: HTMLDivElement is never necessary. Let it default to HTMLElement

source/features/infinite-commit-list-scroll.tsx
```diff
@@ -0,0 +1,92 @@
+import select from 'select-dom';
+import React from 'dom-chef';
+import fetchDom from '../libs/fetch-dom';
+import features from '../libs/features';
+
+// Used feature: infinite-scroll as template
```
fregante: Unnecessary

source/features/infinite-commit-list-scroll.tsx
```diff
@@ -0,0 +1,92 @@
...
+	const lastDivInCommitListing = select('.commits-listing > div:last-of-type', document_);
+	return lastDivInCommitListing!.textContent!;
+}
+
+async function appendOlder(): Promise<void> {
+	// Don't load if we don't have the link
+
+	if (!link?.href) {
+		return;
+	}
```
fregante: This never happens or should never happen. This function is called by IntersectionObserver so `link` is being watched and it exists. If not, the observed should be disconnected earlier.

source/features/infinite-commit-list-scroll.tsx
Adk9p: The reason I removed the children and placed the loading button inside `container` is it solves 2 problems, 
1. I don't have to `.disconnect()` and `.observe()` with `inView` when I change something
   * since I don't full understand `IntersectionObserver` don't know what to `.observe()` after calling `githubLoadingButton.replaceWith(…)`
1. It's keeps it style of `container`, so I don't have to add style to the loading button, it aligns it's self

source/features/infinite-commit-list-scroll.tsx
Adk9p: Defaulting to `document` gives an 
> Type 'HTMLElement' is missing the following properties from type 'Document': URL, activeElement, alinkColor, all, and 86 more.

And I can't change `_document` to type `Document` because what fetchDom returns `Element`

Adk9p:
I removed the duplicate date remover code because it was the source of the bug were it stops working when it gets to `Commits on Apr 25, 2019` in https://github.com/sindresorhus/refined-github/commits/master

Im looking into how to fix it.

source/features/infinite-commit-list-scroll.tsx
```diff
@@ -0,0 +1,79 @@
...
+	link = select<HTMLAnchorElement>('div > a:last-child', container)!;
+
+	if (link) {
+		inView.observe(container);
+	}
+}
+
+features.add({
+	id: __featureName__,
+	description: 'Infinite scroll on commit list',
```
sindresorhus:
```suggestion
	description: 'Infinite scroll on commit lists.',
```

source/features/infinite-commit-list-scroll.tsx
```diff
@@ -0,0 +1,79 @@
...
+import fetchDom from '../libs/fetch-dom';
+import features from '../libs/features';
+
+// Button from: https://primer.style/css/components/loaders
+const githubLoadingButton = <button className="btn mn-3" disabled><span>Loading</span><span className="AnimatedEllipsis"></span></button>;
+
+let container: HTMLElement;
+let link: HTMLAnchorElement | undefined;
+
+let lastDate: string;
```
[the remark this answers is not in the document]
Adk9p: Sorry I'm bad at naming, `lastDate` is indented to hold the last date label in a commit-list ([img](https://imgur.com/a/qMC91D2))
It's used to check if the first label in the older content is the same and removed it if true

i.e. Removed duplicate labels

I'm bad at creating verbose names, so if you can think of better one I can change it to that.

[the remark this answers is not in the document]
sindresorhus: > Sorry I'm bad at naming

You're not going to improve if you don't try 😉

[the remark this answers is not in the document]
Adk9p: Pff, who needs to try when github already has a perfectly verbose class name for the Element!

Adk9p:
beautiful

Adk9p:
@fregante would adding inf scroll upwards fix that?

fregante:
I think implementing preload is the safer and easier alternative. A proper infinite scroll takes a lot of work.

PR #2667, opened 2020-01-05
Use first commit for repo age

### LINKED ISSUE #2660: `repo-age` should display the date of the first commit
PR https://github.com/sindresorhus/refined-github/pull/2560 uses the repo creation time to show the age. This may be confusing and misleading by tricking people into thinking it's a project age.

Some projects may migrate between code hostings over time. For example, for https://github.com/cherrypy/cherrypy Refined GitHub shows the age of 3 years while it's way older: the first commit is dated 2004 (I think the project started ~2002 but the version control doesn't have that info).

So it would be nice to have some switch to enable using the oldest commit date instead of the repo creation time.

hardikmodha:
Closes #2660

Using the latest commit available on the repo root and the total number of commits to fetch the oldest commit. 

## Test
1. https://github.com/sindresorhus/refined-github
2. https://github.com/nodejs/node
3. https://github.com/python/cpython


source/features/repo-age.tsx
```diff
@@ -15,15 +15,21 @@ const dateFormatter = new Intl.DateTimeFormat('en-US', {
 });
 
 const getRepoCreationDate = cache.function(async (): Promise<string> => {
-	const {repository} = await api.v4(`
-		repository(${getRepoGQL()}) {
-			createdAt
-		}
-	`);
+	const commitsCount = Number(select('li.commits .num')!.textContent!.trim()!.replace(',', ''));
```
fregante: Works without trimming. Also the selectos is more specific

```suggestion
	const commitsCount = Number(select('.octicon-history + span')!.textContent!.replace(',', ''));
```

source/features/repo-age.tsx
```diff
@@ -15,15 +15,21 @@ const dateFormatter = new Intl.DateTimeFormat('en-US', {
...
+	const lastCommitHash = select('.commit-tease .commit-tease-sha')!.getAttribute('href')!.split('/').pop();
```
fregante: Shorter selector; use props instead of attributes where available

```suggestion
	const lastCommitHash = select('.commit-tease-sha')!.href.split('/').pop();
```

source/features/repo-age.tsx
```diff
@@ -15,15 +15,21 @@ const dateFormatter = new Intl.DateTimeFormat('en-US', {
...
 
-	return repository.createdAt;
+	if (!commitsCount || !lastCommitHash) {
+		return Promise.reject(new Error('Cannot get commit count or last commit hash'));
+	}
```
fregante: If these values aren't available, those `select` will fail. 

Also since this is an async function you can just throw an error and it becomes a rejected promise.

source/features/repo-age.tsx
```diff
@@ -15,15 +15,21 @@ const dateFormatter = new Intl.DateTimeFormat('en-US', {
...
+
+	const relativeTime = await fetchDom(
+		`${getCleanPathname()}/commits?after=${lastCommitHash}+${commitsCount - 2}`, 'ol:last-child > li.commits-list-item relative-time'
+	);
+
+	return relativeTime!.getAttribute('datetime')!;
 }, {
-	cacheKey: () => __featureName__ + ':' + getRepoURL()
+	expiration: Number.MAX_SAFE_INTEGER,
```
fregante: Storage isn't infinite. The default 30 days is enough.
```suggestion
```

source/features/repo-age.tsx
```diff
@@ -15,15 +15,21 @@ const dateFormatter = new Intl.DateTimeFormat('en-US', {
...
+	cacheKey: () => __featureName__ + 'updated:' + getRepoURL() // Don't know how to update the existing cache so.
```
fregante: Why do you need to update the key in the first place? Let it expire for now

yakov116:
@fregante should we fall back to the old way if there are no commits?

fregante:
> should we fall back to the old way if there are no commits?

I don't think anybody cares about repos without commits 😄 

At most we can prevent the error by using `if count === 0, return false`

source/features/repo-age.tsx
fregante: My bad, I was wrong about the selector. Yours was fine

source/features/repo-age.tsx
hardikmodha: For 2nd point, I tried just throwing an error but ts compiler was complaining 🤔 but now I tried again and it isn't complaining! 🤦‍♂ Not throwing the error now just returning if needed information is not available.

hardikmodha:
I think just returning makes sense, saves us an unnecessary request.

source/features/repo-age.tsx
hardikmodha: Yeah, That's what I thought. Current selector is contextual and less error-prone.

hardikmodha:
> **Just a thought for now,** but perhaps it would be nice to also link to that first commit since we have the information on the same page _and_ every piece in the summary is also a link.

Makes sense. All the items is a link in summary we can do that.

hardikmodha:
@fregante I have incorporated the changes to point to the oldest commit in the `repo-age` but this introduces a change that will break for existing cached values. I'm now expecting cached values to be a valid JSON string but they are date strings. 
Do we want to handle it for the existing values?

yakov116:
@hardikmodha looks great. One this. With this change there is a noticeable jump on every load.

hardikmodha:
@yakov116 Yes, I noticed that as well. I'm working on adding some sort of loader icon as we discussed.

fregante:
No need to get fancy with loaders. The user timezone feature just uses a static `Guessing timezone…` placeholder

fregante:
> Do we want to handle it for the existing values?

I think it’s nicer to handle that now and drop the code later rather than add extra text to our clear key name 😁

source/features/repo-age.tsx
fregante: The only check here should be `if count === 0, return false`

I think `undefined` won’t be cached and we always expect the hash to be there. If it’s not, it’s an error and it should appear in the console automatically.

source/features/repo-age.tsx
```diff
@@ -37,7 +53,9 @@ async function init(): Promise<void> {
 
 	const element = (
 		<li className="text-gray" title={`Repository created on ${dateFormatter.format(date)}`}>
```
fregante: Should we update the tooltip as well? `First commit dated {…}`

source/features/repo-age.tsx
```diff
@@ -4,30 +4,46 @@ import cache from 'webext-storage-cache';
...
+	if (!commitsCount || !lastCommitHash) {
+		return;
+	}
+
+	const oldestCommit = await fetchDom(
+		`${getCleanPathname()}/commits?after=${lastCommitHash}+${commitsCount - 2}`, 'ol:last-child > li.commits-list-item'
+	);
+
+	return JSON.stringify({
+		link: select<HTMLAnchorElement>('.commit-title [href*="/commit/"]', oldestCommit)!.href,
```
fregante: `url` instead of link

yakov116:
```console
❌ repo-age
content.js:3474 RangeError: Invalid time value
    at repo_age_init (content.js:9396)
    at async features_run (content.js:3469)
```

hardikmodha:
@yakov116 I was not handling the existing cached values previously. Now, I am. Can you check again?

yakov116:
@hardikmodha  nope try
https://github.com/sindresorhus/got

hardikmodha:
@yakov116 It's working for me. Are you sure on the latest commit?

yakov116:
Yup

yakov116:
Maybe the cache got messed up when I was testing?

hardikmodha:
Maybe but not sure. Cause I tested it with the repos which I had previously visited and they had cached values, it is working fine.

yakov116:
@hardikmodha same error; Can you add a console.log before? (I dont have git installed on my work computer)

yakov116:
@hardikmodha the first time you go any repo it crashes
```console
❌ repo-age
content.js:3474 TypeError: Cannot read property 'href' of null
    at content.js:9349
    at async chrome-extension:/chfocjcjdofobbicimhpfdlojinofdhd/content.js:3602
    at async repo_age_init (content.js:9377)
    at async features_run (content.js:3469)
```

hardikmodha:
@yakov116 Yes, I noticed that too. That's why I added https://github.com/sindresorhus/refined-github/commit/e1a95ba4f6f752ecf352e21c2d9c5142b4d9d2e8

yakov116:
@hardikmodha public repo's I see the issue is fixed. I'm seeing it on private ones

source/features/repo-age.tsx
```diff
@@ -4,50 +4,108 @@ import cache from 'webext-storage-cache';
...
 	// `twas` could also return `an hour ago` or `just now`
 	const [value, unit] = twas(date.getTime())
 		.replace('just now', '1 second')
 		.replace(/^an?/, '1')
 		.split(' ');
 
-	const element = (
+	const repoAge = <>{repoIcon()} <span className="num text-emphasized">{value}</span> {unit} old</>;
+
+	// Commit url won't be present for the existing cached values
```
webknjaz: Why not cache both date and the commit hash? Just in case it needs to be invalidated after the force-push.

webknjaz:
> I don't think anybody cares about repos without commits

There's also a corner case when the repo's main branch doesn't have commits but there's commits in a separate orphaned tree.

yakov116:
@hardikmodha I also noticed it will fail on any link that the path has `tree` in it (open a file from this pr and go to the main page)

source/features/repo-age.tsx
```diff
@@ -4,50 +4,108 @@ import cache from 'webext-storage-cache';
...
+		<li className="text-gray">
+			{repoIcon()} <span className="text-emphasized js_repo_age_placeholder">Loading...</span>
+		</li>
+	);
+
+	await elementReady('.overall-summary + *');
+
+	addRepoAgeItem(placeholder);
+
+	const details = await getOldestCommitDetails();
```
fregante: Start fetching this first so we're not waiting for the DOM

source/features/repo-age.tsx
```diff
@@ -4,50 +4,108 @@ import cache from 'webext-storage-cache';
...
 	const [value, unit] = twas(date.getTime())
 		.replace('just now', '1 second')
 		.replace(/^an?/, '1')
 		.split(' ');
 
-	const element = (
+	const repoAge = <>{repoIcon()} <span className="num text-emphasized">{value}</span> {unit} old</>;
+
+	// Commit url won't be present for the existing cached values
+	const element = oldestCommitUrl ? (
```
fregante: Sorry I wasn't clear, this isn't what I meant. By "handling old data" I meant "if cached value is old, call `cache.delete(...)` and call `getOldestCommitDetails` again" (https://github.com/fregante/webext-storage-cache/issues/9 would make it easier)

source/features/repo-age.tsx
```diff
@@ -4,50 +4,108 @@ import cache from 'webext-storage-cache';
...
+	} else {
+		select('.numbers-summary')!.append(element);
+	}
+};
+
 async function init(): Promise<void> {
-	const date = new Date(await getRepoCreationDate());
+	const placeholder = (
+		<li className="text-gray">
+			{repoIcon()} <span className="text-emphasized js_repo_age_placeholder">Loading...</span>
```
fregante: No snake case
Refined GitHub classes start with `rgh-`
Probably you don't need a class at all; just select `span`

source/features/repo-age.tsx
```diff
@@ -4,50 +4,108 @@ import cache from 'webext-storage-cache';
...
+	const date = new Date(repoCreationDate);
 
 	// `twas` could also return `an hour ago` or `just now`
 	const [value, unit] = twas(date.getTime())
 		.replace('just now', '1 second')
 		.replace(/^an?/, '1')
 		.split(' ');
 
-	const element = (
+	const repoAge = <>{repoIcon()} <span className="num text-emphasized">{value}</span> {unit} old</>;
```
fregante: Instead of regenerating the whole item, you can use the same technique you used to set `Cannot determine repo age from the oldest commit`

In this case, you'll need

```js
select('span', placeholder)!.replaceWith(
	<><span className="num text-emphasized">{value}</span> {unit} old</>
);
```

fregante:
If this is starting to look like a lot of moving parts, you can also split the PR into 2 pieces:

- add loader
- use updated data source
- or even a 3rd piece: add link (which means changing the data format)

source/features/repo-age.tsx
fregante: That PR is in. Here's an usage example: https://github.com/sindresorhus/refined-github/commit/e172c1687ab22f6c7b5636ba386926683ab5b47d#diff-6fb950c4dc62c32202ad39058c423bf1R36

`repoProjectCount` was the old key, so its presence meant that the cache was in the old format and therefore considered "expired"

source/features/repo-age.tsx
hardikmodha: Older code. We are caching the last commit URL and date. Also, We are not invalidating on hash change currently.

source/features/repo-age.tsx
```diff
@@ -4,50 +4,87 @@ import cache from 'webext-storage-cache';
...
+type OldestCommitDetails = {
+	url: string;
+	datetime: string;
+};
 
-	return repository.createdAt;
+const getOldestCommitDetails = cache.function(async (): Promise<OldestCommitDetails | false> => {
+	await elementReady<HTMLAnchorElement>('.commit-tease-sha', {target: select('.commit-tease')!});
+	const commitsCount = Number(select('li.commits .num')!.textContent!.replace(',', ''));
+	const lastCommitHash = select<HTMLAnchorElement>('.commit-tease-sha')!.href.split('/').pop()!;
```
hardikmodha: @fregante Even though I had added `await elementReady<HTMLAnchorElement>('.commit-tease-sha')` on Line 23, this was failing with error `Cannot read property href of null`. I added `target: select('.commit-tease')` and after that I did not see that error. Is this a bug?  

@yakov116 Can you also check if you are still getting an error?

yakov116: I am not, but... i cleared the cache so...

fregante: `elementReady` may also return `null` because by default it stops at domready.

You can use `lastCommit = await elementReady(...)` to just get the item on the spot instead of using `select` again.

Also if that `target` is available on the page when `elementReady` runs, then you don't need `elementReady` because `commit-tease` and `commit-tease-sha` loaded together, I think.

source/features/repo-age.tsx
```diff
@@ -4,50 +4,87 @@ import cache from 'webext-storage-cache';
...
+
+	if (commitsCount === 0) {
+		return false;
+	}
+
+	if (commitsCount === 1) {
```
fregante: I think repos with a single commit are rare enough that we can drop this condition.

source/features/repo-age.tsx
```diff
@@ -4,50 +4,87 @@ import cache from 'webext-storage-cache';
...
+	const oldestCommit = await fetchDom(
+		`${getRepoURL()}/commits?after=${lastCommitHash}+${commitsCount - 2}`, 'ol:last-child > li.commits-list-item'
+	);
+
+	return {
+		url: select<HTMLAnchorElement>('.commit-title [href*="/commit/"]', oldestCommit)!.href,
+		datetime: select('relative-time', oldestCommit)!.getAttribute('datetime')!
+	};
 }, {
+	isExpired: (cachedValue: OldestCommitDetails | false) => !cachedValue || typeof cachedValue.url === 'undefined',
```
fregante: `isExpired` is only called if the age was cached, it can't be false. Does this type work?

```suggestion
 	// eslint-disable-next-line @typescript-eslint/no-explicit-any
	isExpired: (cachedValue: any) => typeof cachedValue === 'string',
```

source/features/repo-age.tsx
```diff
@@ -4,50 +4,87 @@ import cache from 'webext-storage-cache';
...
+
+	// `twas` could also return `an hour ago` or `just now`
+	const [value, unit] = twas(date.getTime())
+		.replace('just now', '1 second')
+		.replace(/^an?/, '1')
+		.split(' ');
+
+	placeholder.title = `First commit dated ${dateFormatter.format(date)}`;
+	placeholder.innerHTML = '';
+	placeholder.append(<a href={details.url}>{repoIcon()} <span className="num text-emphasized">{value}</span> {unit} old</a>);
```
fregante:
```suggestion
	placeholder.textContent = '';
	placeholder.append(
		<a href={details.url} title={`First commit dated ${dateFormatter.format(date)}`}>
			{repoIcon()} <span className="num text-emphasized">{value}</span> {unit} old
		</a>
	);
```

source/features/repo-age.tsx
hardikmodha: It will work but I was trying to avoid `any`.

source/features/repo-age.tsx
hardikmodha: Then, What do we display for 1 commit repo?

source/features/repo-age.tsx
fregante: Yeah the correct type would be `string | OldestCommitDetails` (because the cache is either in the old format or the new format) but `webext-storage-cache` doesn't allow that type, I have to fix it.

source/features/repo-age.tsx
fregante: Without this condition, it will just continue like it would, to `fetchDom` etc.

hardikmodha: But the request would not return anything, the page returned by `fetchDom` would be empty and we won't find any commits there....

fregante: Ouch. Didn’t think it would be. I guess it can stay then.

source/features/repo-age.tsx
hardikmodha: > `isExpired` is only called if the age was cached, it can't be false. Does this type work

I was going through the `webext-storage-cache` code and found that it will cache the `false` value also. We are returning `false` when no of commits are 0. Also, I found that at a couple of places we are returning false from the getter functions but in that case, the false values will be cached. Am I making sense?

```
const freshValue = await getter(...args);
await set<TValue>(key, freshValue, options.expiration);
return freshValue;
```

hardikmodha: So I can change the type to 
```
isExpired: (cachedValue: OldestCommitDetails | false | string) => !cachedValue || typeof cachedValue === 'string'
```
and it will work as well as reflect the correct types.

source/features/repo-age.tsx
hardikmodha: Got it. I passed `stopOnDomReady: false` and it's no longer causing an error.

source/features/repo-age.tsx
```diff
@@ -4,50 +4,86 @@ import cache from 'webext-storage-cache';
...
+	const oldestCommit = await fetchDom(
+		`${getRepoURL()}/commits?after=${lastCommitHash}+${commitsCount - 2}`, 'ol:last-child > li.commits-list-item'
+	);
+
+	return {
+		url: select<HTMLAnchorElement>('.commit-title [href*="/commit/"]', oldestCommit)!.href,
+		datetime: select('relative-time', oldestCommit)!.getAttribute('datetime')!
+	};
 }, {
+	isExpired: (cachedValue: OldestCommitDetails | false | string) => !cachedValue || typeof cachedValue === 'string',
```
fregante: If `isExpired` returns true, the cache is discarded/ignored.

If you include `!cachedValue`, you're saying that a `false` value is expired, i.e. "discard any cached `false` value`.


```suggestion
	isExpired: (cachedValue: OldestCommitDetails | false | string) => typeof cachedValue === 'string',
```

hardikmodha: `false` will only be cached when there were no commits previously. We again need to check if commits are added. If yes, we try to get the first commit and date otherwise we return `false`. Correct?

fregante: > We again need to check if commits are added

If you want to check _again_, then it shouldn't be cached in the first place the function should return `undefined`

If you store `false`, every time you open the Repo tab you get 3 `storage` calls:

1. get (returns `false`)
2. delete (due to `isExpiring`)
3. set (sets `false`)

If you don't store it at all, you get:

1. get

fregante: Thanks for discussing this with me btw, I know it can get frustrating

hardikmodha: Returning `undefined` would be a better way but seems like types are incompatible. `cache.function` doesn't allow `undefined` to be returned. I think, It should be allowed. The types for both can be made different. I mean returned values from the function and cached values. What do you think?

fregante: v2.3.0 should do it: https://github.com/fregante/webext-storage-cache/pull/14

fregante:
Hardik can you split this into multiple PRs as suggested before? https://github.com/sindresorhus/refined-github/pull/2667#issuecomment-571003482

The first 2 PRs should be pretty straightforward and can be merged quickly. The 3rd apparently is taking a while to discuss

PR #2668, opened 2020-01-05
Highlight new features

### LINKED ISSUE #2664: Ability to disable new options by default
How hard would it be to have the ability to disable any new options by default? There have been a number of changes recently that I don't care for, and I find myself struggling to find the name of the preference that was added.

jerone:
# Description
Highlight new features in the options page for 10 days after opening the settings window, per https://github.com/sindresorhus/refined-github/issues/2664#issuecomment-570794975.

# Screenshot
image

# Closes
Closes #2664


yakov116:
@jerone would it make sense to allow searching for the word new?

jerone:
[**@yakov116**](https://github.com/yakov116) commented on [Jan 5, 2020, 8:59 PM GMT+1](https://github.com/sindresorhus/refined-github/pull/2668#issuecomment-570942274 "2020-01-05T19:59:01Z - Replied by Github Reply Comments"):
> [@jerone](https://github.com/jerone) would it make sense to allow searching for the word new?

I rather see some sorting/filtering being introduced.

jerone:
Good call. I included an timespan of 10 days. This will start after opening the settings window.

source/options.css
```diff
@@ -13,6 +13,7 @@ html {
 	--input-border-color: rgba(0, 0, 0, 0.2);
 	--github-green: #28a745;
 	--github-red: #cb2431;
+	--feature-new: #ff4f4f;
```
fregante: Can we use `--github-red` instead?

source/options.css
```diff
@@ -130,6 +131,17 @@ h2 ~ h2 {
 	margin-left: 0.6em;
 }
 
+.feature-new .feature-name::after {
+	content: 'New';
+	color: var(--feature-new);
+	font-size: 14px;
```
fregante: Use `em` everywhere (except `border-width`), this text could be changed by the browser, we inherit their style.

source/options.css
```diff
@@ -130,6 +131,17 @@ h2 ~ h2 {
...
+	border: 1px solid var(--feature-new);
```
fregante: `border` inherits from `color`

```suggestion
	border: 1px solid;
```

source/options.tsx
```diff
@@ -71,6 +72,25 @@ async function init(): Promise<void> {
...
+		}
+	}
+
+	for (const feature of __featuresInfo__) {
+		if (!(feature.name in newFeatures)) {
+			newFeatures[feature.name] = Date.now();
+		}
+	}
+
+	await cache.set('features-new', newFeatures, 1000);
```
fregante: I think you should just store them with `browser.storage.local.setItem`. 1000 days is almost 3 years

source/options.tsx
```diff
@@ -71,6 +72,25 @@ async function init(): Promise<void> {
 		container.prepend(unchecked.closest('.feature')!);
 	}
 
+	// Highlight new features
+	const newFeatures: Record<string, number> = await cache.get<Record<string, number>>('features-new') ?? {};
```
fregante:
```suggestion
	const newFeatures = await cache.get<Record<string, number>>('features-new') ?? {};
```

Also perhaps we need a better key. `features-new` seems to suggest that it's a list of new features while it's more like `features-first-seen-date`

source/options.tsx
```diff
@@ -71,6 +72,25 @@ async function init(): Promise<void> {
...
+	if (browser.runtime.getManifest().version !== '19.12.22') { // First version introducing highlighting
+		const timespan = 10 * 24 * 60 * 60 * 1000; // Highlighting will be shown for 10 days after opening the options
```
fregante: What if we make it clearer?
```suggestion
		const tenDaysAgo = Date.now() - (10 * 24 * 60 * 60 * 1000);
```

And then `newFeatures[feature.id] > tenDaysAgo, add class`

fregante:
Also I just loaded the extension and all of them were marked as NEW (Firefox)

fregante:
> @jerone would it make sense to allow searching for the word new?

`new` already shows about 10 features. `is:new` works with this change in the filter:

```diff
- .replace(/\W/g, ' ')
+ .replace(/[^\w:]/g, ' ')
```

source/options.tsx
jerone: `1000` was just an random high number, because it should never expire.

jerone:
[**@fregante**](https://github.com/fregante) commented on [Jan 7, 2020, 8:21 AM GMT+1](https://github.com/sindresorhus/refined-github/pull/2668#issuecomment-571468059 "2020-01-07T07:21:43Z - Replied by Github Reply Comments"):
> Also I just loaded the extension and all of them were marked as NEW (Firefox)

Did you load the extension as Developer? Because then the extension version is 0.0.0.

fregante:
What version do you expect this to work correctly on?

source/options.tsx
fregante: Right, so it’s not cache. Cache is meant to expire

jerone:
[**@fregante**](https://github.com/fregante) commented on [Jan 7, 2020, 9:48 AM GMT+1](https://github.com/sindresorhus/refined-github/pull/2668#issuecomment-571492705 "2020-01-07T08:48:08Z - Replied by Github Reply Comments"):
> What version do you expect this to work correctly on?

Before merging this PR, the version number [here](https://github.com/sindresorhus/refined-github/pull/2668/files#diff-041820b02ac3d6d7c4e862285d60afb5R77) needs to be altered to the next extension version. This is the first version introducing highlighting. Otherwise the same thing happens like you experienced; all features highlighted.

Thinking about it, I can probably work-around that.

fregante:
> the version number [here](https://github.com/sindresorhus/refined-github/pull/2668/files#diff-041820b02ac3d6d7c4e862285d60afb5R77)

That's not a good strategy; this expects the user to open the Options page exactly on that version and not a digit later

jerone:
Agree. Will implement another strategy.

jerone:
Implemented `browser.storage.local`. It will detect if the key `featuresAlreadySeen` is present. If not, this means the user visits the options page for the first time. This solves https://github.com/sindresorhus/refined-github/pull/2668#issuecomment-571468059.

If you want to test this feature, add the following code on [line 79](https://github.com/sindresorhus/refined-github/pull/2668/files#diff-041820b02ac3d6d7c4e862285d60afb5R79):
``` ts
delete featuresAlreadySeen['add-tag-to-commits'];
```

fregante:
Do you think you can merge the two loops using only `select.all`?

```
for feature in checkboxes
  if not in list
    add it 

  if not first visit and not within 10 days
    mark as new

jerone:
> Do you think you can merge the two loops using only `select.all`?

Done ✅

fregante:
Looking good, thank you!

This is still welcome https://github.com/sindresorhus/refined-github/pull/2668#issuecomment-571469995

PR #2670, opened 2020-01-06
Fix `cloneNode` type

fregante:
Due to https://github.com/microsoft/TypeScript/issues/283

However it's not entirely correct, so if you know a solution to this comment, speak up https://github.com/microsoft/TypeScript/issues/283#issuecomment-571024679 😄


fregante:
I'm merging this. If you know a better solution, comment on that issue

PR #2678, opened 2020-01-07
Meta: Enable JSX linting

fregante:
Finally JSX is linted now, thanks to https://github.com/xojs/eslint-config-xo-react and https://github.com/yannickcr/eslint-plugin-react


jerone:
Cool :shipit: 

_Offtopic_: I get the feeling that `lint:css` is not run on the PR. Can you confirm?

fregante:
It’s being run:

https://github.com/sindresorhus/refined-github/blob/1d180ef1005d91199419fd9fdc771fee62a235b5/.github/workflows/test.yml#L25

https://github.com/sindresorhus/refined-github/blob/1d180ef1005d91199419fd9fdc771fee62a235b5/package.json#L4

jerone:
> It should be:

The weird thing is, that on my development environment it gives the following errors if I run `lint:css`:
```
source/options.css
  98:1  ×  Expected selector ".js-features input" to come before selector "input:not([type='checkbox']):invalid"   no-descending-specificity
 166:2  ×  Expected selector "body > :not(hr)" to come before selector ":root"                                     no-descending-specificity
```

But I'm not seeing those errors on GitHub Actions. I didn't touch that file. Weird 🙃

fregante:
You may be right; This shouldn't be valid CSS: https://github.com/sindresorhus/refined-github/commit/2a56ca13bb65b7983a7d6aa786d4e5055258bfb5

But lint passes: https://github.com/sindresorhus/refined-github/runs/381524191

jerone:
> `stylelint source/*.css` includes files directly under `source` but `stylelint source/**/*.css` doesn't.

`**` should match zero or more directories. Meaning that `source/**/*.css` matches CSS files under `source` and all sub-directories. I can confirm that when I run `lint:css` and get the errors on the file `source/options.css`. You also confirmed that in your test, that it was able to run on file `source/options.css`: https://github.com/sindresorhus/refined-github/runs/381528802#step:4:12

This all doesn't explain **why I** get these errors, and GitHub Actions not. Do you get those errors too when you run `lint:css`?

I'm on Windows. Also cleaned my whole `node_modules` folder. Also tried `stylelint@12.0.1`, without success.

I do see some talks about `no-descending-specificity` in the [latest PR](https://github.com/stylelint/stylelint/pull/4491#issuecomment-569019760), but it's [not yet released](https://github.com/stylelint/stylelint/issues/4489#issuecomment-572481272).

fregante:
> I'm on Windows

That's the difference.

With `stylelint source/**/*.css` the glob expansion is up to the shell. Windows does it, bash apparently not.

With `stylelint 'source/**/*.css'`, stylelint takes care of the glob expansion in a cross-platform way instead of letting the shell do it.

PR #2680, opened 2020-01-09
Enable `parse-backticks` in more places on discussion pages

samuelmeuli:
Follow-up for https://github.com/sindresorhus/refined-github/pull/2653#issuecomment-571549229

**Adds code highlighting in the following places:**

- Issue/PR references:

image

- Commmits pushed to a PR:

image

- Issue/PR title changes:

image

The last one unfortunately doesn't have a dedicated CSS class. I used `.TimelineItem-body` which targets all timeline items on PRs/issues. It should probably be discussed whether this inclusion is worth it (considering site performance).

### DIFF
--- a/source/features/parse-backticks.tsx
+++ b/source/features/parse-backticks.tsx
@@ -1,9 +1,10 @@
+import elementReady from 'element-ready';
 import './parse-backticks.css';
 import select from 'select-dom';
 import features from '../libs/features';
 import {parseBackticks} from '../libs/dom-formatters';
 
-function init(): void {
+function initGeneral(): void {
 	for (const title of select.all([
 		'.commit-title', // `isCommit`
 		'.commit-desc', // `isCommit`, `isCommitList`, `isRepoTree`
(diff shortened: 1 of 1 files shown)


fregante:
Can the selector be made more specific? Perhaps with `:only-child` or anything related to that strike through

samuelmeuli:
Maybe `.TimelineItem-body del, .TimelineItem-body ins`? There probably aren't many `del`/`ins` elements on a PR/issue page, so performance shouldn't be too bad

samuelmeuli:
Added highlighting on the Dashboard page:

image

fregante:
Wow, this is starting to look like a lot of places 😯

samuelmeuli:
Yeah, I hope that's finally it 😅

samuelmeuli:
If these features are loaded dynamically based on the page URL, it might be worth considering splitting this up into chunks (to avoid all selectors being evaluated on each page)

fregante:
I don’t think the selector is a huge performance problem. A single querySelectorAll should be way more efficient than multiple “features” and their overhead

fregante:
Thanks Samuel! 🐰

PR #2683, opened 2020-01-09
Add `watch:chrome` NPM script

samuelmeuli:
`web-ext` added support for Chromium browsers in v3.2. This PR adds an NPM script for debugging the extension in Chrome.

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -112,6 +112,8 @@ Once built, load it in the browser of your choice:
 				<li>Click on the <strong>Load unpacked extension</strong> button;
 				<li>Select the folder <code>refined-github/distribution</code>.
 			</ol>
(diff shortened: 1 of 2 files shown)


fregante:
Honestly I’d rather drop both of them. I don’t think anyone uses them and if you do you have to set your key every time. It’s not a good experience

samuelmeuli:
Fair point about the key, I've been testing without it so far.

jerone:
> Honestly I’d rather drop both of them. I don’t think anyone uses them and if you do you have to set your key every time. It’s not a good experience

I'm using it. It's much simpler then remembering the whole command. In addition I'm running the command with an custom profile (`-p=Development`).
```diff
-		"watch:firefox": "web-ext run --source-dir=distribution",
+		"watch:firefox": "web-ext run --source-dir=distribution -p=Development",
```

PR #2687, opened 2020-01-11
Meta: Restore stylelint on all files, cross-platform

jerone:
See discussion [here](https://github.com/sindresorhus/refined-github/pull/2678#issuecomment-572209852). Stylelint was not running correctly on any other OS then Windows with the glob value.

This PR fixes:
1. the **glob issue**. See proof here: https://github.com/sindresorhus/refined-github/pull/2687/checks?check_run_id=384780372#step:4:9. I couldn't use single quotes `'`, because that resulted in `Error: No files matching the pattern "'source/**/*.css'" were found.`. Using double quotes fixed that.
2. fixes the **errors from stylelint**.

### DIFF
--- a/source/options.css
+++ b/source/options.css
@@ -95,7 +95,7 @@ h2 ~ h2 {
 	margin-top: 1em;
 }
 
-.js-features input { /* stylelint-disable-line no-descending-specificity */
+.js-features input[type='checkbox'] {
 	flex-shrink: 0;
 	margin-inline-end: 0.6em; /* Spacing for Chrome */
 	margin-top: 4px; /* Alignment for Chrome */
--- a/source/options.css
+++ b/source/options.css
@@ -95,7 +95,7 @@ h2 ~ h2 {
 	margin-top: 1em;
 }
 
-.js-features input {
+.js-features input { /* stylelint-disable-line no-descending-specificity */
 	flex-shrink: 0;
 	margin-inline-end: 0.6em; /* Spacing for Chrome */
 	margin-top: 4px; /* Alignment for Chrome */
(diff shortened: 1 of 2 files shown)


source/options.css
```diff
@@ -95,7 +95,7 @@ h2 ~ h2 {
 	margin-top: 1em;
 }
 
-.js-features input {
+.js-features input { /* stylelint-disable-line no-descending-specificity */
```
fregante: Can you fix it instead?

jerone: I found an solution for this one. Not for the other one.

fregante: The solution for both is to move the whole rule up or down in the file. The selector doesn't need to be solved

fregante: I think `stylelint` is awkwardly picking up the `input` selector inside the `@-moz-document` rule. Perhaps the selector fix is a better solution 👌

fregante:
🎉

jerone:
@fregante Your last commit blocks Windows! See my description above:
> I couldn't use single quotes `'`, because that resulted in `Error: No files matching the pattern "'source/**/*.css'" were found.`. Using double quotes fixed that.

source/options.css
jerone: I would expect stylelint to not trigger this error when the source is different, e.g. no properties matches.

jerone:
> @fregante Your last commit blocks Windows! See my description above:
> > I couldn't use single quotes `'`, because that resulted in `Error: No files matching the pattern "'source/**/*.css'" were found.`. Using double quotes fixed that.

See https://github.com/stylelint/stylelint/issues/3624 and https://github.com/stylelint/stylelint.io/blob/master/package.json#L14

Issue #2693, opened 2020-01-12
[REQUEST] Document Required Token Access Scope per Feature


LesterCovax:
Currently, the link to create a PAT for Refined Twitter encompasses the entire `Repo` scope.  There is [currently a link provided](https://github.com/sindresorhus/refined-github/search?q=libs/api) within the extension's options to the functions accessing the API, but it's not immediately clear which access scope each one requires (i.e. `repo:status`, `repo_deployment`, `public_repo`, `repo:invite`).

I think it would be ideal to document these specifics (and even the need for a Personal Access Token) in the [README.md](https://github.com/sindresorhus/refined-github/blob/master/readme.md), instead of users finding out once they open the extension options for the first time.  Adding the required scope after each feature requiring it would be nice for those that like to limit PAT auth access, without them having to dig through the source code.  I realize that the `Repo` scope isn't very invasive, but it would have been nice to know and have documented up front.

Cheers!

yakov116:
Duplicate of #2679

LesterCovax:
@yakov116 How can an **issue** be a duplicate of a **_comment_** within an unrelated issue? That's not how that works.  
(I did search for similar issues before posting FWIW)

fregante:
@LesterCovax It's duplicate in the sense that both are about documentation around PAT.

I've been hesitant to add more content on the options page, but since I had to explain the PAT in the issues before, perhaps we can add a short text near the field, either via expandable `details` or linked to a more explanatory document on the repo or issues.

When installing the extension, https://github.com/sindresorhus/refined-github/issues/1137 opens. Recently I added a small paragraph suggesting to add the token.

>  There is [currently a link provided](https://github.com/sindresorhus/refined-github/search?q=libs/api) within the extension's options to the functions accessing the API, but it's not immediately clear which access scope each one requires.

None of those extra permissions are required, but the all-inclusive `repo` permission is the only way to gain access to private repos, so we have to select it.

fregante:
Continue discussion in https://github.com/sindresorhus/refined-github/pull/2694

Issue #2679, opened 2020-01-08
"Update Branch" button: show warning if PAT is undefined


mmoayyed:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%2410%20Funded-%2300A156.svg" />](https://issuehunt.io/r/sindresorhus/refined-github/issues/2679)

Thank you for this useful extension!

If no personal access token is defined, clicking on the "Update Branch" button changes the button caption to "Not Found", and it's not immediately obvious why. Alternatively, the button can be displayed, hidden or perhaps a warning message should/could be displayed in its place instead, when no access token is defined/available.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $10.00)

- [<img src='https://avatars0.githubusercontent.com/u/183673?v=4' alt='janpio' width=24 height=24> janpio](https://issuehunt.io/u/janpio) ($10.00)

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/sindresorhus/refined-github/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

sindresorhus:
Would be nice if features in the settings indicated whether or not they required a PAT.

fregante:
> Would be nice if features in the settings indicated whether or not they required a PAT.

We do have a [link](https://github.com/sindresorhus/refined-github/search?q=libs/api) for that in the options:

<img width="409" alt="" src="https://user-images.githubusercontent.com/1402241/71966709-ab309f00-3234-11ea-886b-6a13dfae8abd.png">

I don't think it would solve this issue specifically since the button would still be shown.

---

Like with any feature, the way to find out why a feature isn't working, is to open the console. Such error can show up in the future regardless of the PAT and the console should have more information.

How can we teach the users to open the console when something doesn't work?

fregante:
**I extended the copy in** https://github.com/sindresorhus/refined-github/issues/1137 (the first thing new users see) with info about:

- where to look for errors;
- the PAT: until now, the user had to open the options to even know that a PAT may be needed.

fregante:
I'll add: Even if there's a token saved, it doesn't mean that it's valid and that it has access to the current repo, so:

- **"user action" features like this should show a warning**: "XXX failed because YYY"
- "background" features that just show info will silently fail (other than the console error)

[...]

PR #2694, opened 2020-01-13
Meta: Improve copy around personal access token in the options

fregante:
Related to https://github.com/sindresorhus/refined-github/issues/2679
Closes https://github.com/sindresorhus/refined-github/issues/2693

Please review and suggest further copy or explanation you think it's needed here.

### DIFF
--- a/source/options.html
+++ b/source/options.html
@@ -6,8 +6,9 @@
 <form id="options-form" class="detail-view-container">
 	<p>
 		<label>
-			<strong>Personal token</strong><br>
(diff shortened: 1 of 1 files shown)


jerone:
For the forms I create, I always assume all fields are required and put an <sup>`(optional)`</sup> remark behind the ones that are optional.

fregante:
Indeed, but I can't come up with anything better than the current wording: "Not required _but_ ..."

Perhaps:

```
Personal token (optional)
[Some features]() may not work without it. [Generate one]().
```

jerone:
I like that last one ✔🤝

PR #2706, opened 2020-01-15
fix notifications beta auto completetion overlap

### LINKED ISSUE #2703: `display: flex` doesn't play well with Notifications Beta search
Without the extension enabled:

img-2020-01-14-171320

With the extension enabled (quick search overlaps the search box):

img-2020-01-14-171339

I looked at the CSS used by refined-github, and it changes the `display` of the search elements from `block` and `inline-block` to `flex`. The autocomplete is a `block` and ends up interacting poorly with other elements there.

I'm using this custom CSS as a workaround:

```css
.js-notifications-container .subnav-search-input {
	display: inline-block;
}

.js-notifications-container .subnav-search {
	display: block;
}
```

but there's probably a better way to do it (using `flex` for the autocomplete, for example).

### LINKED ISSUE #2705: Notification beta filter drop-down auto completion is overlapping
- I disabled both Refined and GitHub Dark turning them on and off. This is definitely located in Refined GitHub.
- https://github.com/notifications/beta?query=is%3Aunread\
With Refined GitHub enabled.
image
With Refined GitHub disabled
image

I located the issue, it's related to the form parent element being set to a `display: flex` instead of how GitHub intended it to be `display: block`

This is the CSS in question.

https://github.com/sindresorhus/refined-github/blob/cfbc2d4c7166eea2bd4d653ff93a746b87e53d93/source/features/widen-search-field.css#L16-L24

casz:
Closes #2705
Closes #2703

# Test

https://github.com/notifications/beta

### DIFF
--- a/source/features/widen-search-field.css
+++ b/source/features/widen-search-field.css
@@ -16,7 +16,7 @@ https://github.com/sindresorhus/refined-github/releases/tag/19.9.16
 /* Expand issue search field width */
 .repository-content [role='search'], /* Repo issue/PR lists */
 .subnav-links + .float-right, /* Global search pages */
-.subnav-search,
+.subnav-search:not(form.js-notification-search-form), /* Ignore notifications (beta) search form */
 .subnav-search-input {
 	flex-grow: 1;
 	display: flex;


source/features/widen-search-field.css
```diff
@@ -16,7 +16,7 @@ https://github.com/sindresorhus/refined-github/releases/tag/19.9.16
 /* Expand issue search field width */
 .repository-content [role='search'], /* Repo issue/PR lists */
 .subnav-links + .float-right, /* Global search pages */
-.subnav-search,
+.subnav-search:not(form.js-notification-search-form), /* Ignore notifications (beta) search form */
```
casz: not sure a `:not` is the best way to exclude the form.

fregante:
I don’t have the beta. If it works for you, I’ll merge. The selector is specific enough that it shouldn’t affect other parts of the site.

casz:
Works for me, Let me take a screenshot

casz:
image

fregante:
Thank you!

casz:
I found another one that you can test: https://github.com/sindresorhus/refined-github/actions

I think we could add some CSS that targets the `form.subnav > div.card-filter-autocomplete-dropdown` directly

image

PR #2708, opened 2020-01-16
Meta: Add error logger that helps user search/report issues

fregante:
I felt that the errors could be more detailed and helpful


fregante:
Also, can some take a look at that `<reference>` tag? I don’t know why AVA errors without it

yakov116:
@fregante are those items links or plane text?

fregante:
If you’re referring to the URLs in the console, they’re clickable

PR #2714, opened 2020-01-20
Restore link to `highest-rated-comment`

### LINKED ISSUE #2623: `highest-rated-comment` no longer adds link to comment
The comment is highlighted but no link is added at the top.

```
❌ highest-rated-comment content.js:2996:15
TypeError: "select_dom_default(...)(...) is null"
    linkBestComment moz-extension://67f5e0a6-5eef-6b42-923e-0d235f11edd2/content.js:7709
    highest_rated_comment_init moz-extension://67f5e0a6-5eef-6b42-923e-0d235f11edd2/content.js:7673
    features_run moz-extension://67f5e0a6-5eef-6b42-923e-0d235f11edd2/content.js:2992
    add moz-extension://67f5e0a6-5eef-6b42-923e-0d235f11edd2/content.js:3039
    onAjaxedPages moz-extension://67f5e0a6-5eef-6b42-923e-0d235f11edd2/content.js:2949
```

## Test

https://github.com/zeit/next.js/issues/9524

dertieran:
Fixes #2623 

Changed the selectors and adjusted the styling a bit so it fits into the timeline again.

image

### DIFF
--- a/source/features/highest-rated-comment.tsx
+++ b/source/features/highest-rated-comment.tsx
@@ -82,8 +82,8 @@ function linkBestComment(bestComment: Element): void {
 		link.removeAttribute('data-hovercard-url');
 		link.href = hash;
 
-		// We don't copy the exact timeline item structure,
-		// so we need to align the avatar with the other avatars in the timeline.
+		// We don't copy the exact timeline item structure, so we need to align the avatar with the other avatars in the timeline.
+		// TODO: update DOM to match other comments, instead of applying this CSS
 		avatar.style.left = '-55px';
 
 		bestComment.parentElement!.firstElementChild!.after((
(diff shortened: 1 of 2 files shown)


fregante:
Spacing isn't right:

Screen Shot 2020-01-21 at 03 10 05

But I'm surprised this fixed most of it! When I tried fixing it it looked like I had to rewrite the whole dom because they changed it a lot.

dertieran:
@fregante Yea also thought that first and then tried to just fix the styling as good as possible. 

But I somehow didn't notice the avatar :see_no_evil: 
That could be fixed by adjusting `left: -55px`, not sure if that should be the solution or if we should instead copy the new structure.

I don't think that the avatar also linked to the comment, also I wouldn't have expected that, but I can add it.

fregante:
I think that won’t work if the view is responsive (e.g. reload the page when the window is less than 700px wide)

PR #2715, opened 2020-01-20
Avoid console error caused by `forked-to`

### LINKED ISSUE #2334: `forked-to`: element not found
https://github.com/sindresorhus/refined-github/blob/f8c2fab4cfd8f85f19552a897f1a6df49ea441d9/source/features/forked-to.tsx#L54-L56

Browser: Chrome 78.0.3880.4
Extension: 19.8.14
Example URL: all repos 💥

I don't know if this only happens to me, but there seems to be no `details-dialog[src*="/fork"]` element it is looking for.
I've tried only enabling Refined GitHub and still the same.

Well, this bug has been there quite a while, but I thought it was network error caused by uBlock (at first glance) and didn't mind 😅

dertieran:
Fixes #2334 

Not sure if that is what you expected to fix the issue :smile:

### DIFF
--- a/source/features/forked-to.tsx
+++ b/source/features/forked-to.tsx
@@ -56,8 +56,12 @@ async function updateForks(forks: string[]): Promise<void> {
 }
 
 async function init(): Promise<void> {
-	select('details-dialog[src*="/fork"] include-fragment')!
-		.addEventListener('load', saveAllForks);
+	const fragment = select('details-dialog[src*="/fork"] include-fragment');
+	if (!fragment) {
+		return;
+	}
+
+	fragment.addEventListener('load', saveAllForks);
 
 	const forks = await cache.get<string[]>(getCacheKey());
 
--- a/source/features/forked-to.tsx
+++ b/source/features/forked-to.tsx
@@ -55,11 +55,15 @@ async function updateForks(forks: string[]): Promise<void> {
 	}
 }
 
-async function init(): Promise<void> {
-	select('details-dialog[src*="/fork"] include-fragment')?.addEventListener('load', saveAllForks);
+async function init(): Promise<void | false> {
+	const fragment = select('details-dialog[src*="/fork"] include-fragment');
+	if (!fragment) {
+		return false;
+	}
+
+	fragment.addEventListener('load', saveAllForks);
 
 	const forks = await cache.get<string[]>(getCacheKey());
-
 	if (!forks) {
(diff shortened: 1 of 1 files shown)


fregante:
I think a full solution would be to still updated the cache somehow, but I'll take this bugfix for now.

jerone:
I was never able to reproduce this error. That element should always be present. Being part of an organization is not required, as you can always fork an repo to your own account.

It could be that that element is just not ready yet. `elementReady` could be an solution.

dertieran:
Hmmm I can't find a `details-dialog[src*="/fork"]`.
*(Tested in this PR)*

For me the fork button looks something like this:
```html
<form action="/sindresorhus/refined-github/fork" method="post">
	<button type="submit">Fork</button>
</form>
<a href="/sindresorhus/refined-github/network/members" class="social-count">
	813
</a>
```

jerone:
Reload the page, open dev console, execute `$('details-dialog[src*="/fork"] include-fragment`)

This element will be replaced as soon as you hover over the fork button.

jerone:
Oke, I was finally able to reproduce this error; it happens when the user is 1) not part of an organization and 2) has not forked the repo.

The suggested solution by @fregante [above](https://github.com/sindresorhus/refined-github/pull/2715#issuecomment-576414209) should be sufficient.

jerone:
It looks like @fregante was correct after all; this element exists only if you are part of an organization. Having forked the repo (or not) doesn't matter.

That means there is no detection possible to know if you have forked the repo yourself or not.

For now, current fix fixes the error. But we might need another detection for people without organization. And now I go to bed 😄

fregante:
I think I solved it. Follow me:

1. `include-fragment` lets you choose which organization to fork the current repo into
2. If you’re not part of an organization, it’s missing
3. If you’re not part of an organization, clicking the fork button will just create or open an existing fork

Look at point 3: that’s what this feature is about. If you don’t have organizations, you don’t need this feature because forking/opening an existing fork is already one-click. 

So the original solution was correct.

jerone:
Agree.

Maybe we should rename this feature to `forked-to-organization` 😄

fregante:
Kinda, but not really. I can still fork to my own user. I think perhaps the description should be updated to include “if you are part of any organizations”

fregante:
Thanks for the bugfix Philipp!

PR #2598, opened 2019-12-06
Fix `z-index` bug caused by `sticky-discussion-sidebar`

yohanneshabte:
Closes #2576

# Test

https://github.com/sindresorhus/refined-github/issues/1137
**Check by opening the customize notifications menu on the right side**
#### Before
before
#### After
after

PR #2716, opened 2020-01-20
Fix `z-index` bug caused by `sticky-discussion-sidebar`

### LINKED ISSUE #2576: `z-index` problem when modal is opened in issue detail page
Hi,

I found a bug in issue detail page. Some elements appeared on the modal backdrop when the modal is opened.

I tested on Chrome(v78.0.3904.108) and Microsoft Edge Dev(v79.0.287.3).

**Sample Url**
`https://github.com/sindresorhus/refined-github/issues/1137`

**Screenshot**
image

dertieran:
Fixes #2576 

This fixes the `z-index` issue with modals opened from the sticky sidebar by increasing the `z-index` of the sidebar.

I set the `z-index` to 110 so the backdrop also overlays the sticky header.
We could also set it to 1 to only overlay the timeline.

### DIFF
--- a/source/features/sticky-discussion-sidebar.css
+++ b/source/features/sticky-discussion-sidebar.css
@@ -5,6 +5,7 @@
 .rgh-sticky-sidebar {
 	position: sticky;
 	top: 60px !important; /* Matches sticky header's height */
+	z-index: 110; /* Needed so the backdrop of notification modal is over the other elements in the discussion */
 }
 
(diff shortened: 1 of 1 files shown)


fregante:
Seems to work without side effects. Maybe! 🤞

fregante:
Kidding. 😞 

bug


I think the solution in https://github.com/sindresorhus/refined-github/pull/2598 was good, with an adjusted selector

dertieran:
@fregante weird it works for me, the sidebar always stays below the sticky header.
image
*(I'm on Chromium 79)*

PR #2682, opened 2020-01-09
Add space between `ci-link` and "Private" label

jonheslop:
This gives the build status icon some breathing space

Closes #2681

## Before

Screenshot 2020-01-09 at 10 58 20

## After

Screenshot 2020-01-09 at 11 03 59

PR #2717, opened 2020-01-20
Fix ci-link's spacing for private repositories

lukaszklis:
**BEFORE:**

before

**AFTER:**

after

**DOM structure for private repos:**

image


fregante:
I think this was fixed by https://github.com/sindresorhus/refined-github/pull/2682

lukaszklis:
My bad. Thanks for spotting this.

PR #2736, opened 2020-01-26
Avoid duplicate `comments-time-machine-links` in comment dropdown

### LINKED ISSUE #2637: `comments-time-machine-links` piling up when hiding comments
1. Hide a comment
1. Open the comment's dropdown again and notice the duplicates

bblommers:
Closes #2637  

# Test

Reproduceable steps:
1. Add a comment (to any PR/issue)
2. Hide the comment
3. Add a second comment
4. Check the hidden comment - there should be only one 'View repo at this time' link

# Underlying issue
The feature marks comments with a special class to indicate whether the button is there.  
Everytime you add a new comment it will select comments that do not have that class, but the select was looking for that class in the wrong place


fregante:
Test

fregante:
Looks great! Thanks for another bugfix Bert! 🍰

Issue #2732, opened 2020-01-23
High CPU and RAM usage on large files (due to `show-whitespace`)


pdonias:
Opening the link below with Refined GitHub uses a lot of CPU and eats all the RAM very quickly.

<details>
<summary>Be ready to have your computer freeze before clicking the link</summary>
<a href="https://github.com/teropa/to-sting/blob/master/index.js">https://github.com/teropa/to-sting/blob/master/index.js</a>
</details>

![Capture_2020-01-23_15:59:25](https://user-images.githubusercontent.com/10992860/72995676-98ec5e80-3df9-11ea-9bd7-84d38c92a4fd.png)

Refined GitHub 20.1.22
Firefox 72.0.1 (I couldn't reproduce the issue on Chrome)
Ubuntu 18.04.3 LTS

fregante:
Confirmed, this is due to `show-whitespace`. The solution would be "easy": detect the slow loop and stop:

https://github.com/sindresorhus/refined-github/pull/2087/files/c2c44b10e42ff2ccbbf94cf66fef5bf7ef9f6f2e..e3474a401792d5dab19b391208798fd8d5d3e77a

PR welcome. Also perhaps there are improvements in the algorithm but this would be a quick fix.

AleksandrHovhannisyan:
> // eslint-disable-next-line no-await-in-loop

What if instead of using `await` we use old-fashioned promise handlers?

fregante:
I don't understand. How would you rewrite that loop with "promise handlers"?

AleksandrHovhannisyan:
What I probably meant is promise chaining. I don't know the terminology too well.

But `await` is gonna slow down that loop for sure.

fregante:
Await isn’t slower than promise chains. The lint rule is because await stops the loop and some users don’t realize that. 

I found the problem of this issue: the feature is currently wrapping each individual character into its own DocumentFragment and then appending the whole thing back. The line in the middle is 200k characters long so you can see how that might take forever to finish. 

I’m rewriting the feature to be a lot more efficient and I’ll probably also add a line length limit for these cases

AleksandrHovhannisyan:
> The lint rule is because await stops the loop and some users don’t realize that.

Right, I just figured stopping the loop is what was slowing the whole thing down.

Sorry for sticking my nose in this, though; I'm trying to get more involved in OSS but haven't found something that works for me.

fregante:
[...]

PR #2737, opened 2020-01-26
Improve `show-whitespace` performance on large files

fregante:
Closes https://github.com/sindresorhus/refined-github/issues/2732

I rewrote the feature to limit DOM manipulation. The previous one created a new DocumentFragment for each character, whether it was whitespace or not.

Even this version was slow until I cached `textNode.textContent`; Firefox really doesn't like reading that many times in a loop. Now it's only read once per replacement, which makes the test URL instant. A related bug was opened here but closed as `worksforme`: https://bugzilla.mozilla.org/show_bug.cgi?id=1330475

~~I wonder if manipulation could be further improved with [`Range.surroundContents`](https://developer.mozilla.org/en-US/docs/Web/API/range/surroundContents) but the current solution seems fast enough.~~ Edit: tried, 5x slower: https://github.com/sindresorhus/refined-github/pull/2737/commits/066402aa8a083228d6f875cb9d255bbca65d05b3

Current bugs: 
- [x] for some reason it's also trying to replace other characters like [`=` and `{`](https://user-images.githubusercontent.com/1402241/73139102-fd7b1800-409c-11ea-89f7-1a2620702b3f.png)
- [x] alignment seems off

# Test
Careful, this URL will freeze your Firefox window if you open it with the `master` version of Refined GitHub: https://github.com/teropa/to-sting/blob/master/index.js

### DIFF
--- a/source/features/show-whitespace.css
+++ b/source/features/show-whitespace.css
@@ -14,6 +14,6 @@
 	opacity: 0.25;
 }
 
-[data-rgh-whitespace^='→'] {
+[data-rgh-whitespace^='→']::before {
 	letter-spacing: calc((var(--tab-size, 4) * 1ch) - 1ch);
 }
--- a/source/features/show-whitespace.css
+++ b/source/features/show-whitespace.css
@@ -1,10 +1,11 @@
-.rgh-ws-char {
+[data-rgh-whitespace] {
 	position: relative;
 	line-height: 1em;
 	text-indent: 0; /* Reset any indentation added by `indented-code-wrapping` feature */
 }
 
-.rgh-ws-char::before {
+[data-rgh-whitespace]::before {
+	content: attr(data-rgh-whitespace);
 	pointer-events: none;
 	user-select: none;
 	position: absolute;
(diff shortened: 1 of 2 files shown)


fregante:
On my computer, the long https://github.com/microsoft/TypeScript/blob/master/lib/lib.dom.d.ts file performs terribly, with or without Refined GitHub. Now that I dropped the `await setTimeout` I should make sure that performance on the file isn't impacted.

notlmn:
Will take a look at this once I'm back at a keyboard!

fregante:
I tested `lib.dom.d.ts` and it still took between 5s and 25s to complete. I initially added a timer (https://github.com/sindresorhus/refined-github/pull/2737/commits/f6f9f8b1c5de9ec9947469876fc156904930bb46) but I think we can skip the work altogether: **only parse visible code** (7cccc48)

This will make the whitespace appear a couple of frames later but I think it's worth it. I thought that `observ`ing thousands of elements would slow the browser down but I don't see any difference on that page.

bonbon-713:
Thanks

xplendour:
Than

fregante:
@notlmn let me know today if you want to look at this because I’d like to merge it and publish a new version with this

PR #2742, opened 2020-01-29
Change style of `emphasize-draft-pr-label`

fregante:
Closes https://github.com/sindresorhus/refined-github/issues/2450

The original problem was that the contrast between green and gray was poor. I think this solution increases the contrast without making Draft PRs look bigger.

## Normally

normal

## Before

before

## After

after

# Test

https://github.com/WordPress/gutenberg/pulls

### DIFF
--- a/source/features/emphasize-draft-pr-label.css
+++ b/source/features/emphasize-draft-pr-label.css
@@ -1,4 +1,4 @@
 .js-issue-row [aria-label='Open draft pull request'] svg {
 	color: #fff !important;
-	filter: drop-shadow(#6a737d 0 0 0.5px);
+	filter: drop-shadow(#000 0 0 0.7px);
 }
--- a/source/features/emphasize-draft-pr-label.css
+++ b/source/features/emphasize-draft-pr-label.css
@@ -1,4 +1,4 @@
 .js-issue-row [aria-label='Open draft pull request'] svg {
 	color: #fff !important;
-	filter: drop-shadow(#6a737d 0px 0px 0.5px);
+	filter: drop-shadow(#6a737d 0 0 0.5px);
 }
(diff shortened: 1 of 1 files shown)


tooomm:
I like the outlined style a lot. 👍 
Maybe little less shadow if it's still readable as it also stresses things a bit?

Idea: Did you also try the usual green color from PR's over grey?

fregante:
Yeah my initial idea was to add a solid outline instead but this looked ok. 

You mean a green shadow/outline? Doesn’t sound bad

sindresorhus:
Good idea, but the shadow is too weak. Hard to see the icon.

tooomm:
>Yeah my initial idea was to add a solid outline instead but this looked ok.

Solid outline could also work. And deals with @sindresorhus's concern.
Could you post a picture, please?

>You mean a green shadow/outline? Doesn’t sound bad

Exactly.

<br>

>Good idea, but the shadow is too weak. Hard to see the icon.

I even feel like the shadow stresses the icon in some way as it pops out from the background in comparison to the solid ones.
Would need to see a comparison to judge if grey/green and outlined with/without shadow work better.

fregante:
Outline would be best, but there's no way to do that in CSS. I'd have to recreate the icon with the outline and embed the whole new SVG. Anyone wants to take a crack at creating the style in Illustrator/Sketch/etc?

sindresorhus:
Stroke:

Screenshot 2020-02-01 at 01 05 47

```css
the-icon {
	stroke: rgba(0, 0, 0, 0.8);
	stroke-width: 0.2px;
}
```

sindresorhus:
Yours with stronger shadow looks pretty good though:

Screenshot 2020-02-01 at 01 11 59

```css
filter: drop-shadow(rgb(0, 0, 0) 0px 0px 0.7px);
```

fregante:
> Looks pretty bad...

Yeah, because the stroke is applied _on_ the border (half inside, half outside).

PR #2744, opened 2020-01-31
Fix `hide-inactive-deployments`

FloEdelmann:
GitHub changed the selector.

# Test
* https://github.com/OpenLightingProject/open-fixture-library/pull/425


fregante:
It might be worth updating this selector too https://github.com/sindresorhus/refined-github/blob/7ba1fe2af7ca3b0b7e1a8661bfcfd428dbae2618/source/features/clean-discussions.css

PR #2745, opened 2020-01-31
Add `extend-discussion-status-filters` feature

### LINKED ISSUE #2574: Remove is:open/is:closed issue search query with a click
<table><td>
@fregante said:
<tr><td>
For the bounty:

1. Use this https://github.com/sindresorhus/refined-github/issues/2574#issuecomment-559944187
2. Remove icons as suggested in https://github.com/sindresorhus/refined-github/issues/2574#issuecomment-564284650
3. Add `is:merged` filter in PRs

</table>

[Issuehunt badges](https://issuehunt.io/r/sindresorhus/refined-github/issues/2574)

To see all issues (open and closed) you need to manually remove is:open/is:closed search query. Maybe add "All" next to "Open", "Closed", with or without counter, to remove this query with a click?

Example URL:

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $15.00)

- [fregante fregante](https://issuehunt.io/u/fregante) ($15.00)

### Tips

---
</details>

KarlHeitmann:
Closes #2574 feature request.

# Test

You can test it on any **issues** page. For example: [here](https://github.com/sindresorhus/refined-github/issues)


KarlHeitmann:
I've passed the linter to fix the previous errors, however there are some errors that right now I couldn't figure out how to solve. Can someone give me a hint, or if I need to find another approach to solve these errors.

yakov116:
Did you try xo --fix?

source/features/remove-is-open-is-closed.tsx
```diff
@@ -0,0 +1,80 @@
+import React from 'dom-chef';
+import select from 'select-dom';
+import features from '../libs/features';
+
+const count = (string: string, re: RegExp): number => {
+	return ((string || '').match(re) ?? []).length;
+	// return (re.exec(string || '') ?? []).length;
```
sindresorhus: Don't leave commented out code.

source/features/remove-is-open-is-closed.tsx
```diff
@@ -0,0 +1,80 @@
...
+	id: __featureName__,
+	description: 'Remove is:open/is:closed issue search query with a click, add Merged link button next to them.',
+	screenshot: 'https://user-images.githubusercontent.com/3003032/73557979-02d7ba00-4431-11ea-90da-5e9e37688d61.png',
+	include: [
+		features.isRepo
+	],
+	exclude: [
+		features.isOwnUserProfile
+	],
+	load: features.onDomReady, // Wait for DOM ready
```
sindresorhus: I think the code is already clear enough. No need for this comment.

source/features/remove-is-open-is-closed.tsx
```diff
@@ -0,0 +1,80 @@
+import React from 'dom-chef';
+import select from 'select-dom';
+import features from '../libs/features';
+
+const count = (string: string, re: RegExp): number => {
```
sindresorhus:
```suggestion
const countMatches = (string: string, regex: RegExp): number => {
```

KarlHeitmann:
> Did you try `xo --fix`?

@yakov116 , there are some comments where I've run `npm run lint-fix`, however, the lint errors that remains, the auto fixer is unable to fix. I believe it is because there must be added some type of variable on TS that I don't know which should be. I've tried again with `npx xo --fix`, but this command was also unable to auto fix the problems.

source/features/remove-is-open-is-closed.tsx
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	if (countMatches(linkMergedSearchString, /is:merged/) === 1) {
+		linkMergedSearchParameters.set('q', linkMergedSearchString.replace(/is:merged/, 'is:issue').trim());
+		linkIsMerged.classList.add('selected');
+	} else if (countMatches(linkMergedSearchString, regexpQueryTotal) === 1) {
+		linkMergedSearchParameters.set('q', linkMergedSearchString.replace(regexpQuery, 'is:merged').trim());
+	}
+
+	linkIsMerged.search = String(linkMergedSearchParameters);
```
fregante: To generate this link, try:
- cloning the “closed” link: cloneNode(true)
- replacing the `textContent` to Merged
- running this on its query: `.replace('is:closed', 'is:merged')`

source/features/remove-is-open-is-closed.tsx
```diff
@@ -0,0 +1,78 @@
...
+	if ((divTableListFiltersParent === null) || (inputJsIssuesSearch === null)) {
+		return;
+	}
+
+	const mergeLink = createMergeLink();
+
+	const containerTargetButtons = divTableListFiltersParent.children[0].children[0];
+	const targetButtons: HTMLCollectionOf<HTMLAnchorElement> = containerTargetButtons.children;
+	for (const link of targetButtons) {
+		link.firstElementChild!.remove();
```
fregante: If you’re trying to remove the icon, it’s safer to run `select('.octicon', link)!.remove()` because the first element could change and we’d remove the wrong one.

source/features/remove-is-open-is-closed.tsx
```diff
@@ -0,0 +1,78 @@
...
+		if (link.classList.contains('selected')) {
+			link.prepend(
+				<svg className="octicon octicon-check" viewBox="0 0 12 16" version="1.1" width="12" height="16" aria-hidden="true">
+					<path fill-rule="evenodd" d="M12 5l-8 8-4-4 1.5-1.5L4 10l6.5-6.5L12 5z"/>
+				</svg>
+			);
```
fregante: Look at how other features handle icons; I think you’ll have to `import checkIcon from 'octicons/check.svg'`

source/features/remove-is-open-is-closed.tsx
```diff
@@ -0,0 +1,78 @@
...
+			const linkSearchParameters = new URLSearchParams(link.search);
+			const linkQuery = linkSearchParameters.get('q');
+			linkSearchParameters.set('q', linkQuery!.replace(/is:open|is:closed/, '').trim());
+			link.search = String(linkSearchParameters);
+		}
+	}
+
+	containerTargetButtons.append(mergeLink);
+}
+
```
fregante: Split the feature in two functions (same file): `togglableFilters` and `addMergedLink`

I think togglableFilters also has to act on the merged link, so it should run after it

source/features/remove-is-open-is-closed.tsx
```diff
@@ -0,0 +1,78 @@
...
+	const divTableListFiltersParent = select('div.table-list-filters');
+	const inputJsIssuesSearch = select<HTMLInputElement>('#js-issues-search');
+	if ((divTableListFiltersParent === null) || (inputJsIssuesSearch === null)) {
+		return;
+	}
+
+	const mergeLink = createMergeLink();
+
+	const containerTargetButtons = divTableListFiltersParent.children[0].children[0];
+	const targetButtons: HTMLCollectionOf<HTMLAnchorElement> = containerTargetButtons.children;
```
fregante: Use selectors for these two variables

source/features/remove-is-open-is-closed.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+	return linkIsMerged;
+}
+
+function init(): void {
+	const divTableListFiltersParent = select('div.table-list-filters');
+	const inputJsIssuesSearch = select<HTMLInputElement>('#js-issues-search');
+	if ((divTableListFiltersParent === null) || (inputJsIssuesSearch === null)) {
+		return;
+	}
```
fregante: Only use this condition if there’s some situation where these elements are absent. If you always expect them there, don’t use this condition. 

This is connected to the `include` filter below. `isRepo` is too wide. Change the filter to this specific view, I think called `isDiscussionList`

source/features/remove-is-open-is-closed.tsx
KarlHeitmann: I don't have certainty that on the page there would be another link (anchor) with class .btn-link. That's why I wanted to grab first the container and then manually get to the anchor elements.

After reading your second comment (the one I put the eyes on) I have the idea that I can grab the divTableListFiltersParent, and then using `select('.btn-link', divTableListFiltersParent)` I can get to the anchor elements I need to take.

What do you think? It's the right approach to get the elements? Or it's better to directly grab them without touching the container element?

source/features/remove-is-open-is-closed.tsx
KarlHeitmann: Didn't know about it. Thanks. It is done on my local repo.

source/features/remove-is-open-is-closed.tsx
KarlHeitmann: Wow, on `page-detect.ts` there is this whole list full of options to send to the include filter. Thanks.

KarlHeitmann: Would it be better if the page is isRepoIssueList? isDiscussionList has:
- isGlobalDiscussionList
  - https://github.com/issues
  - https://github.com/issues?q=is%3Apr+is%3Aopen
  - https://github.com/issues/assigned
  - https://github.com/issues/mentioned
  - https://github.com/pulls
  - https://github.com/pulls?q=issues
  - https://github.com/pulls/assigned
  - https://github.com/pulls/mentioned
  - https://github.com/pulls/review-requested
- isRepoDiscussionList
  - isRepoPRList
  - isRepoIssueList

isRepoIssueList is narrower

fregante: All those pages have the same filters, so why exclude them? 😉

source/features/remove-is-open-is-closed.tsx
fregante: Generally speaking, if you have a container ready, yes you can just look into it. 

Alternatively, maybe you can skip creating a variable for that container and just use the full selector where needed. 

In this case (I didn’t look at the actual DOM) you can achieve the same result with this selector `.table-list-filters > :first-child > :first-child` but surely there are better classes to target.

source/features/remove-is-open-is-closed.tsx
KarlHeitmann: Nice, this way I would not get anymore the typescript error telling me that there are some attributes missing.

However, I doubt that your third point may work, because if the **closed** filter is ON, it's query would not have the `is:closed` filter, so the `.replace('is:closed', 'is:merged')` will not work. That's why I prefer the way it is.

Or maybe I must choose the **open** link to clone if the **closed** link is selected, and vice versa. This way, i will always have the `/is:closed|is:open/` string to replace by is:merged... I'm gonna give it a try!

KarlHeitmann: I've done it. Before calling the function to create the merge link, I check which link I have available on the current page: open, closed or total, and which one is selected. After this check, I clone the appropiate link using cloneNode and send it to the function so it can make the right search query. The fix is uploaded.

source/features/remove-is-open-is-closed.tsx
KarlHeitmann: I didn't make any changes, I understand that it is better to not create a new variable. However, I need the container to iterate over it's elements and change it's behaviours (open & close links), and secondly to add the new mergeLink anchor element.

I would be glad to hear more feedback if you could see the code, and tell me if it is better to create the variable with container, or if it's better to search twice the dom using `.table-list-filters > :first-child > :first-child` under the actual circumstances.

Best regards

source/features/remove-is-open-is-closed.tsx
fregante: What I meant here is that this feature should apply to all of them, hence `isDiscussionList` is appropriate.

07fb61b7

source/features/remove-is-open-is-closed.tsx
fregante: Rather than discussing about variable reuse etc, think more in terms of:
- does this _need_ to exist? Is the code more readable as a consequence of it existing?
- is this code split in reasonable parts? Don't add a piece here and another in a different function, if they're working for the same purpose. Don't look for an element here and then pass it to another function which will just work **around** it.

As you designed them, both `addMergeLink` and `togglableFilters` accept a container _inside which_ you'll be doing work. That's the opposite of what functions normally do. If you pass an element to a function, generally I expect it to work on the element specifically, _not_ to be a "close enough" reference to the work you're about to do.

Refer to 872cd8a7 and 81e41219

source/features/remove-is-open-is-closed.tsx
fregante: Always prefer less code, that's the point of using `cloneNode` instead of creating a new link. All those conditions aren't readable.

Also I suggested the function name `addMergeLink` because it should literally add the merge link, on its own, without external help. It's easier to reason about functions when they're self-contained.

Refer to 872cd8a7

source/features/remove-is-open-is-closed.tsx
fregante: What I meant here is that this feature should apply to all of them, therefore `isDiscussionList` is appropriate.

07fb61b7

fregante:
I left a couple of review comments to help improve the code for future PRs ^_^

source/features/remove-is-open-is-closed.tsx
KarlHeitmann: Thanks, you were clear on the comment. I did understood that isDiscussionList is appropriate.

KarlHeitmann:
> I left a couple of review comments to help improve the code for future PRs ^_^

Thank you Fregante, very kind. I'm gonna review the new commits and comments.

source/features/remove-is-open-is-closed.tsx
KarlHeitmann: Ahhh, I saw again the code and the line you replaced by `isDiscussionList` I put it before reading your comment XD my mistake.

fregante:
Thanks [Karl](https://www.youtube.com/watch?v=SJixW2u4IvQ) 🦙

KarlHeitmann:
Happy to help! See you on the next commit :smile:

yakov116:
Looks Like this stopped working :(

PR #2748, opened 2020-02-02
Fix style of deep-linked `highest-rated-comment`

loilo:
**tl;dr:**

Comparison of a targeted highest rated comment before and after the PR

---

This PR fixes a minor visual bug I didn't bother to open an issue for: The green border of a highest rated comment carries an `!important` marker only on the main box, not on the triangle on the upper left. This leads to the triangle being incorrectly colored when the comment is focused via a URL anchor (because then there's a `:target` pseudo class in GitHub's CSS which overrules our selector's specificity).

### DIFF
--- a/source/features/highest-rated-comment.css
+++ b/source/features/highest-rated-comment.css
@@ -3,7 +3,7 @@
 }
 
 .rgh-highest-rated-comment.timeline-comment--caret::before {
-	border-right-color: #2cbe4e;
+	border-right-color: #2cbe4e !important;
 }
 
 .rgh-highest-rated-comment.timeline-comment--caret::after {


fregante:
Can you add a test link?

loilo:
Sure, here's the one from the screenshot: [microsoft/typescript#23182](https://github.com/microsoft/TypeScript/issues/23182#issuecomment-379091887).

fregante:
Strangely it works in Firefox (left; Chrome on the right)

Screen Shot 2020-02-03 at 14 18 35

loilo:
The specificities of their/our selectors were even. Maybe in Firefox, the extension's stylesheet is loaded *after* GitHub's internal one, which would result in this behavior.

PR #2754, opened 2020-02-04
Add `expand-all-hidden-comments` feature

### LINKED ISSUE #1892: Add way to load *all* comments when some are hidden due to number of comments

kevin940726:
Closes #1892 

Add `expand-all-hidden-comments` feature. Doing a <kbd>Alt</kbd> + <kbd>Click</kbd> on the `Load more` button will load all the hidden comments in the same page. ~Currently we just recursively click on the button over and over again until there is no more button.~ We use an undocumented API from #2625 to load all the comments. Should we add a maximum limit on how many pages a click should load?

# Test

https://github.com/rust-lang/rfcs/pull/2544

GIF: (pay attention to the scroll bar on the right, it takes a long time...)

Kapture 2020-02-05 at 17 42 43

### DIFF
--- a/source/features/expand-all-hidden-comments.ts
+++ b/source/features/expand-all-hidden-comments.ts
@@ -6,7 +6,7 @@ const expanderSelector = '.ajax-pagination-form';
 const expanderButtonSelector = `${expanderSelector} .ajax-pagination-btn`;
 const containerSelector = '#js-progressive-timeline-item-container';
 
-function observe(expanderElement: HTMLElement) {
+function observe(expanderElement: HTMLElement): void {
 	// Waits for the next loaded comments and clicks on any additional "Load more..." buttons it finds
 	const expandingCodeObserver = new MutationObserver(([mutation]) => {
 		const expandButton = select(expanderButtonSelector, mutation.target as Element);
--- a/source/features/expand-all-hidden-comments.ts
+++ b/source/features/expand-all-hidden-comments.ts
@@ -2,9 +2,6 @@ import select from 'select-dom';
 import delegate, {DelegateEvent} from 'delegate-it';
 import features from '../libs/features';
 
-const EXPANDER_SELECTOR = '.ajax-pagination-form';
-const REMAIN_COMMENTS_REGEX = /\s*(\d+) hidden items\s*/;
-
 function handleAltClick(event: DelegateEvent<MouseEvent, HTMLFormElement>): void {
 	if (!event.altKey) {
 		return;
@@ -13,8 +10,8 @@ function handleAltClick(event: DelegateEvent<MouseEvent, HTMLFormElement>): void
 	const form = event.delegateTarget;
 
 	const buttons = select.all('button', form);
-	const remainCommentsButton = buttons.find(button => REMAIN_COMMENTS_REGEX.test(button.textContent!))!;
-	const [remainComments] = REMAIN_COMMENTS_REGEX.exec(remainCommentsButton.textContent!)!;
+	const remainCommentsButton = buttons.find(button => /\s*(\d+) hidden items\s*/.test(button.textContent!))!;
+	const remainComments = remainCommentsButton.textContent!.replace(/\D+g/, '');
 
 	/**
 	As per #2625:
@@ -32,7 +29,7 @@ function handleAltClick(event: DelegateEvent<MouseEvent, HTMLFormElement>): void
 }
 
 function init(): void {
-	delegate(EXPANDER_SELECTOR, 'click', handleAltClick);
+	delegate('.ajax-pagination-form', 'click', handleAltClick);
 }
 
 features.add({
--- a/source/features/expand-all-hidden-comments.ts
+++ b/source/features/expand-all-hidden-comments.ts
@@ -1,14 +1,10 @@
 import delegate, {DelegateEvent} from 'delegate-it';
 import features from '../libs/features';
 
-/**
-As per #2625:
-Construct the new URL. This is undocumented, and may break at any time.
-As of 12/08/2019, URLs look like this:
-`/_render_node/MDExOlB1bGxSZXF1ZXN0MjE2MDA0MzU5/timeline/more_items?variables%5Bafter%5D=Y3Vyc29yOnYyOpPPAAABZemjg2AAqTQyMjE5MTk1MQ%3D%3D&variables%5Bbefore%5D=Y3Vyc29yOnYyOpPPAAABaENrVHAAqTQ1Mzc3MjMzNg%3D%3D&variables%5Bfirst%5D=60&variables%5BhasFocusedReviewComment%5D=false&variables%5BhasFocusedReviewThread%5D=false`
-The key "variables[first]" in the URL query parameters appears to control how many additional comments are fetched.
-Github appears to always set this to 60, but it can be increased up to the total number of hidden items.
-By setting "variables[first]" to total number of hidden items (extracted from the button tex), we can fetch all hidden comments at once.
+/*
+The ajaxed form that loads the new comments points to a URL like:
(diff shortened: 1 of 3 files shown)


fregante:
Thank you for the PR! https://github.com/sindresorhus/refined-github/pull/2625 implemented this in a single HTTP request, could you copy it and adjust it to work on `alt+click`?

fregante:
Also, **maybe** it would make sense to add this as an extra feature of `toggle-everything-with-alt`

kevin940726:
@fregante I'm a little bit hesitate of implementing it with an official API. Even though everything we did here would potentially break at any time, at least we simulate a regular user interaction as much as possible. However, undocumented API is not meant to be call by the users in any way, especially when we don't know what's the limit of that API. Is there a maximum count on `variables[first]`? Is there timeouts for that API? Would it be considered malicious in Github logs? Maybe we could ping Github folks for permission though 🤔?

Anyways, that's just my concern. Maybe I worry too much 😂 

> Also, maybe it would make sense to add this as an extra feature of toggle-everything-with-alt

Ahh, I didn't know we have that! I'll take a look, thx!

fregante:
If it breaks in the future, not a huge deal. This feature is enable via alt+click so it doesn’t break regular usage. I’d rather use that than making multiple HTTP calls.

kevin940726:
@fregante Cool, sure! Let me do that later 👍

kevin940726:
@fregante All done! Not sure we should include this in `toggle-everything-with-alt` though. If it should then probably `expand-all-collapsed-code` should too 🤔 ?

fregante:
> @fregante All done! Not sure we should include this in `toggle-everything-with-alt` though. If it should then probably `expand-all-collapsed-code` should too 🤔 ?

Agree. Let's merge everything.

Kidding! heheh it's good as is.

source/features/expand-all-hidden-comments.ts
```diff
@@ -0,0 +1,48 @@
...
+	By setting "variables[first]" to total number of hidden items (extracted from the button tex), we can fetch all hidden comments at once.
+	 */
+	const actionPath = form.getAttribute('action')!;
+	const url = new URL(actionPath, location.origin);
+	url.searchParams.set('variables[first]', remainComments);
+	form.setAttribute('action', url.pathname + url.search);
+}
+
+function init(): void {
+	delegate(EXPANDER_SELECTOR, 'click', handleAltClick);
```
fregante: You can inline it
```suggestion
	delegate('.ajax-pagination-form', 'click', handleAltClick);
```

source/features/expand-all-hidden-comments.ts
```diff
@@ -0,0 +1,48 @@
...
+	As per #2625:
+	Construct the new URL. This is undocumented, and may break at any time.
+	As of 12/08/2019, URLs look like this:
+	`/_render_node/MDExOlB1bGxSZXF1ZXN0MjE2MDA0MzU5/timeline/more_items?variables%5Bafter%5D=Y3Vyc29yOnYyOpPPAAABZemjg2AAqTQyMjE5MTk1MQ%3D%3D&variables%5Bbefore%5D=Y3Vyc29yOnYyOpPPAAABaENrVHAAqTQ1Mzc3MjMzNg%3D%3D&variables%5Bfirst%5D=60&variables%5BhasFocusedReviewComment%5D=false&variables%5BhasFocusedReviewThread%5D=false`
+	The key "variables[first]" in the URL query parameters appears to control how many additional comments are fetched.
+	Github appears to always set this to 60, but it can be increased up to the total number of hidden items.
+	By setting "variables[first]" to total number of hidden items (extracted from the button tex), we can fetch all hidden comments at once.
+	 */
+	const actionPath = form.getAttribute('action')!;
+	const url = new URL(actionPath, location.origin);
```
fregante:
```suggestion
	const url = new URL(form.action);
```

source/features/expand-all-hidden-comments.ts
```diff
@@ -0,0 +1,48 @@
...
+	url.searchParams.set('variables[first]', remainComments);
+	form.setAttribute('action', url.pathname + url.search);
```
fregante:
```suggestion
	form.action = url.href;
```

source/features/expand-all-hidden-comments.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+function init(): void {
+	delegate(EXPANDER_SELECTOR, 'click', handleAltClick);
+}
+
+features.add({
+	id: __featureName__,
+	description: 'Expands all the hidden comments when you alt-click on any "Load more..." button in issues.',
```
fregante:
```suggestion
	description: 'On long discussions where GitHub hides comments under a "Load more...", alt-clicking "Load more..." will load *all* the comments at once instead of part of them.',
```

source/features/expand-all-hidden-comments.ts
```diff
@@ -0,0 +1,48 @@
...
+	screenshot: 'https://user-images.githubusercontent.com/7753001/73830151-1f179480-483f-11ea-97b5-635fd9d3f0bd.gif',
```
fregante: I think a screenshot is enough here, the gif can't show that **all** of them are loaded as opposed to just part of them.

```suggestion
	screenshot: 'https://user-images.githubusercontent.com/1402241/73838332-0c548e00-4846-11ea-935f-28d728b30ae9.png',
```

Screen Shot 2020-02-05 at 18 33 11

source/features/expand-all-hidden-comments.ts
```diff
@@ -0,0 +1,48 @@
...
+function handleAltClick(event: DelegateEvent<MouseEvent, HTMLFormElement>): void {
+	if (!event.altKey) {
+		return;
+	}
+
+	const form = event.delegateTarget;
+
+	const buttons = select.all('button', form);
+	const remainCommentsButton = buttons.find(button => REMAIN_COMMENTS_REGEX.test(button.textContent!))!;
+	const [remainComments] = REMAIN_COMMENTS_REGEX.exec(remainCommentsButton.textContent!)!;
```
fregante:
```js
const hiddenItemsCount = Number(form.textContent.replace(/\D+/g, ''));
```

kevin940726: `URLSearchParams` requires the value to be type of `string`, so I left it as is.

fregante: Good point, however you missed the `form` and `hiddenItemsCount` parts. Finding the button is not needed. This is the whole function; you can move the description to the header of the feature, below the imports.

```ts
function handleAltClick(event: DelegateEvent<MouseEvent, HTMLButtonElement>): void {
	if (!event.altKey) {
		return;
	}

	const {form} = event.delegateTarget;
	const hiddenItemsCount = form.textContent!.replace(/\D+g/, '');
	const url = new URL(form.action);
	url.searchParams.set('variables[first]', hiddenItemsCount);
	form.action = url.href;
}

function init(): void {
	delegate('.ajax-pagination-form [type="submit"]', 'click', handleAltClick);
}
```

Also notice that I (just now) changed it to listen to clicks on the button rather than the form, because the form includes the squiggly line next to it and does nothing.

kevin940726: Ahh right! This is much cleaner now! 👍

fregante: You missed the variable name again: `hiddenItemsCount`, not `remainComments`

kevin940726: 😱 Sorry 😞

fregante:
謝謝凱

PR #2757, opened 2020-02-04
Restore `pr-filters` and adjust spacing

FloEdelmann:
Currently, the Draft PR filters are not added to the *Reviews* filter, but to the *Assignee* filter in [some repos](https://github.com/OpenLightingProject/open-fixture-library/pulls) and not at all in [others](https://github.com/sindresorhus/refined-github/pulls). The Status filter is not added in either one.

This PR does the following:

* Fix selectors (to show both filters again)
* Use new dropdown syntax (to fix the layout)
* Fix spacing between dropdown buttons
* Cache CI status for faster Status filter (checks another item from #2671)

# Test

* https://github.com/OpenLightingProject/open-fixture-library/pulls
* https://github.com/sindresorhus/refined-github/pulls
* https://github.com/sindresorhus/refined-github/issues (spacing is fixed there, too)

### DIFF
--- a/source/features/pr-filters.tsx
+++ b/source/features/pr-filters.tsx
@@ -86,7 +86,7 @@ async function addStatusFilter(): Promise<void> {
 	const statusFilter = reviewsFilter.cloneNode(true);
 	statusFilter.id = '';
 
-	select('summary', statusFilter)!.firstChild!.textContent = 'Status '; // Only replace text node, keep caret
+	select('summary', statusFilter)!.firstChild!.textContent = 'Status\u00A0'; // Only replace text node, keep caret
 	select('.SelectMenu-title', statusFilter)!.textContent = 'Filter by build status';
 
 	const dropdown = select('.SelectMenu-list', statusFilter)!;
--- a/source/features/pr-filters.tsx
+++ b/source/features/pr-filters.tsx
@@ -5,8 +5,7 @@ import checkIcon from 'octicon/check.svg';
 import features from '../libs/features';
 import {getIcon as fetchCIStatus} from './ci-link';
 
-// The Reviews dropdown doesn't have a specific class, so we expect this sequence (span contains Projects and Milestones)
-const reviewsFilterSelector = '#label-select-menu + span + details';
+const reviewsFilterSelector = '#reviews-select-menu';
 
 function addDropdownItem(dropdown: HTMLElement, title: string, filterCategory: string, filterValue: string): void {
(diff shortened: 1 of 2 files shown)


source/features/pr-filters.tsx
```diff
@@ -64,25 +68,35 @@ async function addStatusFilter(): Promise<void> {
...
-	// TODO: replace this with an API call
-	const hasCI = await fetchCIStatus();
+	const cachedCIStatus = cache.function(async () => {
+		// TODO: replace this with an API call
+		const statusIcon = await fetchCIStatus();
+		return statusIcon !== undefined;
+	}, {
+		expiration: 3,
+		cacheKey: () => __featureName__ + ':' + getRepoURL()
+	});
```
fregante: This function needs to be moved to the top-level scope. Best yet if we update it to use the API, example:

```gql
{
    head: object(expression: "HEAD") {
      ... on Commit {
        history(first: 10) {
          nodes {
            status {
              state
            }
          }
        }
      }
    }
  }
}

```

```js
return repository.head.history.nodes.some((commit: AnyObject) => commit.status);
```

source/features/pr-filters.tsx
```diff
@@ -64,25 +68,35 @@ async function addStatusFilter(): Promise<void> {
...
 	if (!hasCI) {
 		return;
 	}
 
 	// Copy existing element and adapt its content
 	const statusFilter = reviewsFilter.cloneNode(true);
-	const dropdown = select('.select-menu-list', statusFilter)!;
+	statusFilter.id = '';
+
+	select('summary', statusFilter)!.firstChild!.textContent = 'Status '; // Only replace text node, keep caret
```
fregante: I think `\u00A0` avoided line breaks in that place. Refer to the git blame

FloEdelmann: Thanks, good catch!

source/features/pr-filters.tsx
FloEdelmann: I'll have a look at this, so I reverted the cache commit to get this PR ready to merge.

fregante:
Looks great! Hope to see the speed improvements too

fregante:
I noticed that `status:pending` is kinda broken:

Screen Shot 2020-02-07 at 14 47 39


(not our fault though)

fregante:
If you get a chance to work on this again, let’s change every mention of “Status” to “Checks”

PR #2760, opened 2020-02-04
Reword tooltip in `forked-to` links

Kinrany:
"Open your fork to username/repo" -> "Open your fork **at** username/repo"

English is not my first language, but it seems to make more sense if the address is used as a location?

"Clicking this link will open your fork that can be found at https://github.com/username/repo"

# Test


fregante:
I think that's right! "to" would make sense if the action was "to fork to X" or if the verb was "to go to X", but "to visit to X" doesn't

Kinrany:
😄

PR #2771, opened 2020-02-11
Tell me whether master is the latest published version

### LINKED ISSUE #2601: Tell me whether master is the latest published version
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/refined-github/issues/2601)

I think that the objective of https://github.com/sindresorhus/refined-github/issues/288 was to avoid using the wrong version of a readme.

We should have display cue that the user is looking at something that hasn't been released. Such cue would also be useful for contributors that it may be time to tag a new release.

GitHub already includes this information for branches:

We probably cannot display it the same way because it would cause the content to jump down. Ideally we can merge this info with the current tag button in the `branch-buttons` feature.

### Current branch is latest

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Submitted pull Requests
- [#2771 Tell me whether master is the latest published version](https://issuehunt.io/r/sindresorhus/refined-github/pull/2771)
---

### Tips

---
</details>

bblommers:
Fixes #2601

If understood correctly, there are three use cases here:
1. The user is viewing an old release
2. The user is viewing the latest release
3. The user is viewing the latest release, and master has new commits post-release

# Test
1. https://github.com/spulec/uncurl
2. https://github.com/spulec/uncurl/tree/0.0.10
3. https://github.com/spulec/moto
3. https://github.com/spulec/moto/tree/1.3.14

The last test case now has a new icon:

bleeding_edge

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#2601: Tell me whether master is the latest published version](https://issuehunt.io/repos/51769689/issues/2601)
---

</details>


fregante:
Thank you for this! Just a note: by "bleeding edge" I meant "on master but not tagged", so a tag cannot be marked as "bleeding edge"

Also perhaps we need better naming. cc @sindresorhus

source/features/latest-tag-button.tsx
```diff
@@ -18,45 +30,83 @@ const getLatestTag = cache.function(async (): Promise<string | false> => {
...
 	expiration: 1,
-	cacheKey: () => __featureName__ + ':' + getRepoURL()
+	cacheKey: () => __featureName__ + '_tags:' + getRepoURL()
+});
+
+const masterHasCommitsAfter = cache.function(async (commitDate: string): Promise<boolean> => {
+	// Find out whether we have commits after a certain date
+	const repository = await api.v4(`
+		repository(${getRepoGQL()}) {
+			object(expression: "master") {
```
fregante: It’s not `master` but the default branch. Use `defaultBranchRef` if possible, or our own `getDefaultBranch()` function

source/features/latest-tag-button.tsx
```diff
@@ -18,45 +30,83 @@ const getLatestTag = cache.function(async (): Promise<string | false> => {
 			}) {
 				nodes {
 					name
+					target {
+						oid
+						... on Commit {
+							committedDate
```
fregante: Don’t compare dates, git can compare commits directly. Once you have the latest tag, you just compare “default branch” to “latest tag”, and if there’s a difference then it’s “bleeding edge”

You can get both of these information in the same GraphQL query: `{latestTag, isBleedingEdge}`

source/features/latest-tag-button.tsx
```diff
@@ -18,45 +30,83 @@ const getLatestTag = cache.function(async (): Promise<string | false> => {
...
+						}
+					}
 				}
 			}
 		}
 	`);
 
-	const tags: string[] = repository.refs.nodes.map((tag: {name: string}) => tag.name);
+	const tags: TagInfo[] = repository.refs.nodes.map((tag: {name: string; target: {oid: string; committedDate: string}}) => new TagInfo(tag));
```
fregante: Use regular objects, not classes, they can’t be saved in cache. But in this case this change isn’t needed at all because dates shouldn’t be compared.

source/features/latest-tag-button.tsx
```diff
@@ -18,45 +30,83 @@ const getLatestTag = cache.function(async (): Promise<string | false> => {
...
-
-	if (currentBranch === latestRelease) {
-		link.classList.add('disabled');
-		link.setAttribute('aria-label', 'You’re on the latest release');
+	if (currentBranch === latestRelease.name) {
+		const [isBleedingEdge] = await Promise.all([
+			masterHasCommitsAfter(latestRelease.date)
+		]);
+		if (isBleedingEdge) {
+			link.setAttribute('aria-label', 'Current branch is bleeding edge');
```
fregante:
```suggestion
			link.setAttribute('aria-label', 'The current branch is ahead of the latest release');
```

cc @sindresorhus suggestions?

source/features/latest-tag-button.tsx
fregante: Correction, maybe getting both piece of information in a single request isn't possible, but this is the API to see if the current branch is the latest:

```
https://api.github.com/repos/sindresorhus/refined-github/compare/master...20.2.3
```

Where `master` is the branch shown on the page and `20.2.3` is the `latestTag` info we have.

The API returns an object that has `behind_by` and `ahead_by` properties we can store and use to display exactly 

```
${branch} is ${ahead_by} commits ahead of the latest release
```

bblommers: @fregante One downside to this approach that I just discovered: The V3 API returns a lot of unnecessary information, which slows down the call. (Google tells me it's not possible to only return a few fields.. )
One of my example calls has 320 commits after the last release - retrieving that takes 2.2 seconds:
bleeding_edge_slow_response
Which means that the latest-tag box only shows up way after the page has finished loading.
Is that acceptable?

One other approach I took this morning, before I saw your suggestion, was to simply get the latest commit for each branch in a single call:
```
{
  repository(name: "moto", owner: "spulec") {
    latestTag:object(expression: "1.3.14") {
      ... on Commit {
        oid
      }
    },
    defaultBranch:object(expression: "master") {
      ... on Commit {
        oid
      }
    }
  }
}
```
It doesn't give us the full picture, but it's enough for the simple 'Current branch is ahead of the latest release' message

bblommers: Just freestyling here - we can always do both?
1. Use GraphQL to check whether master is bleeding edge
2. Display the icon, so the user immediately knows
3. Make the slow V3 call to get the specifics
4. Update the icon label accordingly

source/features/latest-tag-button.tsx
sindresorhus: > ${branch} is ${ahead_by} commits ahead of the latest release

👍

source/features/latest-tag-button.tsx
fregante: I didn’t think it would be so slow. 
I think your single-call solution is best, at least until there’s an equivalent GraphQL `compare` API. I can’t believe some things are so difficult with GraphQL

bblommers: Done @fregante. I also changed the logic so that only the default branch can be bleeding edge

> Just a note: by "bleeding edge" I meant "on master but not tagged", so a tag cannot be marked as "bleeding edge"

source/features/latest-tag-button.tsx
```diff
@@ -37,26 +39,59 @@ const getLatestTag = cache.function(async (): Promise<string | false> => {
...
+			  defaultBranch:object(expression: "${defaultBranch}") {
+				... on Commit {
+					oid
+				}
+			}
+		}
+	`);
+	// If the default branch has a different (later) commit, it's considered bleeding edge
+	return repository.repository.defaultBranch!.oid !== repository.repository.latestTag!.oid;
+}
```
fregante: The `getLatestTag` function now is duplicate and should be removed/merged. That query can get both the tag and the commit

This function should take its place, name and also also be cached with the same key.

source/features/latest-tag-button.tsx
```diff
@@ -37,26 +39,59 @@ const getLatestTag = cache.function(async (): Promise<string | false> => {
...
+
+	// Find the last commits from the default branch and latest tag
+	const repository = await api.v4(`
+		repository(${getRepoGQL()}) {
+			latestTag:object(expression: "${latestTag}") {
+				... on Commit {
+					oid
+				}
+			  },
+			  defaultBranch:object(expression: "${defaultBranch}") {
```
fregante: Is there any other way to just get the latest commit or `HEAD` of a repo? I think this works `defaultBranchRef.target.oid`

source/features/latest-tag-button.tsx
```diff
@@ -37,26 +39,59 @@ const getLatestTag = cache.function(async (): Promise<string | false> => {
...
+}
+
+async function getTagLink(latestRelease: string): Promise<HTMLAnchorElement> {
 	const link = <a className="btn btn-sm btn-outline tooltipped tooltipped-ne ml-2">{tagIcon()}</a> as unknown as HTMLAnchorElement;
 
 	const currentBranch = getCurrentBranch();
-
+	const defaultBranch = await getDefaultBranch();
+	const [isBleedingEdge] = await Promise.all([
+		currentBranchIsBleedingEdge(latestRelease, defaultBranch, currentBranch)
```
fregante: Let’s drop **every** reference to “bleeding edge” and just use “ahead” or “behind”

“$branch is ahead of the latest release”

source/features/latest-tag-button.tsx
```diff
@@ -72,7 +107,9 @@ async function init(): Promise<false | void> {
 		return false;
 	}
 
-	breadcrumbs.before(getTagLink(latestTag));
+	const [tagLink] = await Promise.all([getTagLink(latestTag)]);
```
fregante: Why is there a `Promise.all` here and above? They don’t do anything

source/features/latest-tag-button.tsx
```diff
@@ -37,26 +39,59 @@ const getLatestTag = cache.function(async (): Promise<string | false> => {
...
 
-function getTagLink(latestRelease: string): HTMLAnchorElement {
+async function currentBranchIsBleedingEdge(latestTag: string, defaultBranch: string, currentBranch: string): Promise<boolean> {
+	// Only default branch can be cutting edge (i.e., ahead of latest release)
+	if (currentBranch !== defaultBranch) {
+		return false;
+	}
+
+	// Find the last commits from the default branch and latest tag
+	const repository = await api.v4(`
```
fregante:
```suggestion
	const {repository} = await api.v4(`
```

This lets your skip the unsightly `repository.repository...` later

source/features/latest-tag-button.tsx
```diff
@@ -72,7 +107,9 @@ async function init(): Promise<false | void> {
 		return false;
 	}
 
-	breadcrumbs.before(getTagLink(latestTag));
+	const [tagLink] = await Promise.all([getTagLink(latestTag)]);
+
+	breadcrumbs.before(tagLink);
```
fregante: At the end, after all the suggestions I made, you should be able to revert this change and just use the previous one:

```js
breadcrumbs.before(getTagLink(latestTag)); 
```

`getTagLink(latestTag, isAhead)` should only generate the link, it doesn't need to be `async`

source/features/latest-tag-button.tsx
bblommers: Done

source/features/latest-tag-button.tsx
bblommers: Done

source/features/latest-tag-button.tsx
bblommers: Cheers, didn't know that

source/features/latest-tag-button.tsx
bblommers: Not sure what you mean by this one - is there already a util/function that knows about the defaultBranch oid?
The only thing I could find in the codebase was the getDefaultBranch(), but that retrieves the  name only

fregante: It’s the API itself: use

```GQL
repository(…) {
  defaultBranchRef {
    target {
      oid 
    }
  }
}
```

bblommers: Ah, I understand. Cheers

fregante:
There are now 4 states:

### Latest tag

tag

### Default branch, tagged

default tagged

### Default branch, not tagged

default branch not tagged

### Other branches or other tags

other branch

fregante:
I think this feature is 🔥

fregante:
Possible improvements:

- smaller icon
- lower opacity
- different icon

While I think this piece of information is important, maybe it shouldn't look like such a big problem.



The "info" icon doesn't seem right either




Other options: https://octicons.github.com/

yakov116:
https://octicons.github.com/icon/dashboard/ ? It looks like a speedometer to me 😀

bblommers:
Completely agree on the icon @fregante - it does make it look like there's a problem. The info-icon is better, IMO, as it simply denotes an informatory message.

Is the eye-icon an alternative? Basically saying "Hey, have a look here, there's something you might want to pay attention to"
https://octicons.github.com/icon/eye/

fregante:
If we had a `behind_by` count we could add a small <sup>+2</sup>

source/features/latest-tag-button.tsx
```diff
@@ -18,67 +30,101 @@ const getLatestTag = cache.function(async (): Promise<string | false> => {
...
-		return false;
+	if (currentBranch === latestTag || isUpToDate) {
+		link.setAttribute('aria-label', 'You’re on the latest release');
+		link.classList.add('disabled');
+		return;
 	}
 
-	breadcrumbs.before(getTagLink(latestTag));
+	const defaultBranch = await getDefaultBranch();
+	if (currentBranch === defaultBranch) {
```
fregante: In here we could trigger an extra request to find out exactly how many commits HEAD is behind.

```js
const getDifference = cache.function(async (): Promise<string> => {
	const tagPage = await fetchDom(`/${getRepoURL()}/releases/tag/${latestTag}`);
	// If it exists, this text is "4 commits to master since this tag"
	return select('.release-header relative-time + a[href*="/compare/"]')!.textContent.replace(/\D/g, '');
}
```

```js
 	if (currentBranch === defaultBranch) { 
		const aheadBy = await getDifference();
		// set label and <sup>+${aheadBy}</sup>
	}
```

bblommers: That looks (and behaves) quite nicely @fregante - thanks!
It now looks like this:
xCommitsAhead


Note: I had to change the query to get the tag OID, as tags don't always have explicit commits attached.  
See for example this query:
```
{
  repository(name: "caprine", owner: "sindresorhus") {
    refs(first: 5, refPrefix: "refs/tags/", orderBy: {field: TAG_COMMIT_DATE, direction: DESC}) {
      nodes {
        name
        tag: target {
          oid
          ... on Tag {
            commit: target {
              abbreviatedOid
              oid
              id
            }
          }
        }
      }
    }
  }
}

```
Tag 2.42.1 doesn't have a commit attached to it, so the tag-OID is the same as the commit-oid.
Other tags do have a separate tag-oid and commit-oid.

Not sure what the difference is, why some tags don't have an explicit commit? 
I changed the logic to accept both, depending on whether the commit is available

fregante: Indeed! Thanks for the change

https://mijingo.com/blog/two-types-of-git-tags

source/features/latest-tag-button.tsx
```diff
@@ -18,67 +31,125 @@ const getLatestTag = cache.function(async (): Promise<string | false> => {
...
+			isUpToDate: false
+		};
 	}
 
-	// If all tags are plain versions, parse them
-	if (tags.every(tag => /^[vr]?\d/.test(tag))) {
-		return tags.sort(compareVersions).pop()!;
+	const tags: Tag[] = repository.refs.nodes.map((node: AnyObject) => ({
+		name: node.name,
+		commit: getTagOid(node)
```
fregante: I think something like this should be enough:

```suggestion
		commit: node.tag.commit?.oid ?? node.tag.oid
```

source/features/latest-tag-button.tsx
```diff
@@ -76,6 +78,28 @@ const getRepoPublishState = cache.function(async (): Promise<RepoPublishState> =
...
+
+	return node.tag.oid;
+}
+
+const getDifference = cache.function(async (latestTag: string): Promise<string> => {
+	const tagPage = await fetchDom(`/${getRepoURL()}/releases/tag/${latestTag}`);
+	// This text is "4 commits to master since this tag"
+	const commitsToMaster = select('.release-header relative-time + a[href*="/compare/"]', tagPage)!.textContent;
+	if (commitsToMaster === null) {
+		return '';
```
fregante: This never happens, you can always use `.textContent!`

In this case 

```ts
return select('.release-header relative-time + a[href*="/compare/"]', tagPage)!.textContent!.replace(/\D/g, '');
```

source/features/latest-tag-button.tsx
```diff
@@ -76,6 +78,28 @@ const getRepoPublishState = cache.function(async (): Promise<RepoPublishState> =
...
 
+function getTagOid(node: AnyObject): string {
+	if (node.tag === {}) {
+		return node.tag.commit.oid;
+	}
+
+	return node.tag.oid;
+}
+
+const getDifference = cache.function(async (latestTag: string): Promise<string> => {
```
fregante: Maybe `getAheadByCount`. Paging @sindresorhus for better naming

source/features/latest-tag-button.tsx
bblommers: Cheers! Wasn't familiar with that notation

source/features/latest-tag-button.tsx
bblommers: I've used getAheadByCount for now - happy to update again if necessary

bblommers:
Anything I can do to help move this along @fregante @sindresorhus?

fregante:
Updating screenshots:


### Latest tag

tag

### Default branch, tagged

default tagged

### Default branch, not tagged

default branch not tagged

### Other branches or other tags

other branch

fregante:
Looks great!

fregante:
Strangely the second call is failing on https://github.com/fregante/webext-storage-cache

```
 TypeError: "Object(...)(...) is null"
    getAheadByCount moz-extension://b3d02b04-e4fa-0c45-9406-305c7b1b784a/content.js:5820
```

yakov116:
@bblommers thanks this feature is awesome!

PR #2777, opened 2020-02-13
Fix AVA-related CI testing issues

fregante:
Travis started failing at some point, probably because tests create thousands of parallel threads and at some point we reached the limit.

https://travis-ci.org/sindresorhus/refined-github/builds/647752010


fregante:
Now AVA seems to be failing 🤷‍♂️ 
cc @sindresorhus

sindresorhus:
Maybe because of https://github.com/avajs/ava/releases/tag/v3.3.0 ?

fregante:
But we have a lockfile

https://github.com/sindresorhus/refined-github/blob/a55748c84ad87c56a813dbd0e281665c1df0a7cf/package-lock.json#L2040-L2042

sindresorhus:
Right, forgot we have that. Must be either that Travis changed something then, or more likely, the Node.js version got a patch/minor release which changed something.

fregante:
Possible. I remember seeing Node 10 on GH Actions but now I see v12.13.1

https://help.github.com/en/actions/reference/software-installed-on-github-hosted-runners#ubuntu-1804-lts

fregante:
Post-mortem: AVA testing needs to be upgraded so that Node 13 is supported.
And I don't know what's up with Travis, but serial tests and Node 11 work.

.github/workflows/test.yml
```diff
@@ -14,8 +14,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v1
```
sindresorhus: You should also upgrade to `checkout@v2` ;)

https://github.com/actions/checkout/releases/tag/v2.0.0

fregante: pr welcome

PR #2784, opened 2020-02-15
Update cached values earlier (stale while revalidate)

fregante:
I'm adding a `keepFresh` option, which implements a behavior commonly known as `stale-while-revalidate` (naming suggestions welcome)

Normally our cached data is displayed until it expires; only after its expiration, new data is fetched and displayed (which introduces a delay because there's no cache at that point)

This new `keepFresh` option lets us display "stale" data but it will still trigger whatever HTTP request is behind it. This will silently update the cache without it ever expiring (as long as the user visits a page between the stale and expiration dates)

Instead of just making it a boolean (i.e. _always_ keep the data fresh after displaying stale data), I decided to make it another `number`

```js
{
	expiration: 30, // After 30 days, the cache is deleted
	keepFresh: 10 // The data is still served from cache, but after 10 days it will trigger another request
}
```

Is this clear? Should the option in https://github.com/fregante/webext-storage-cache/pull/15 be named anything different?


package.json
```diff
@@ -94,6 +93,7 @@
 		"typescript": "^3.7.5",
 		"web-ext": "^4.0.0",
 		"web-ext-submit": "^4.0.0",
+		"webext-storage-cache": "^2.4.0-0",
```
jdreesen: This should remain in `dependencies`, right?

jerone:
> Is this clear? Should the option in [fregante/webext-storage-cache#15](https://github.com/fregante/webext-storage-cache/pull/15) be named anything different?

`staleAfter` maybe...

fregante:
`staleAfter` doesn’t sound bad. Perhaps `stale: 10` would also work since there’s `expiration: 30`

sindresorhus:
I would go even further: `staleAfterDays`. It's not clear what unit `{staleAfter: 10}` at the callsite.

fregante:
What about `expiration: 30`? Is that clear enough? `expireAfterDays`?

Also, maybe: `staleAfterDays` doesn't explain what happens then, should this be `retryAfterDays`?

sindresorhus:
I would use `invalidateAfterDays`.

fregante:
You mean:

```js
{
	invalidateAfterDays: 10,
	expiration: 30
}
```

Or maybe

```js
{
	invalidateAfterDays: 10,
	deleteAfterDays: 30
}
```

fregante:
I think we should skip `*afterDays`:  API don't necessarily specify the unit: sometimes they're milliseconds, sometimes they're seconds, you just have to read the docs (or JSDocs) to find out.

Also maybe match the [HTTP Cache-Control](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control) terminology.

https://web.dev/stale-while-revalidate/

```http
Cache-Control: max-age=1, stale-while-revalidate=59
```


```js
cache.function(fn, {
	maxAge: 10,
	staleWhileRevalidate: 20
});
/* Equivalent to {
	keepFresh: 10,
	expiration: 30
} */
```

sindresorhus:
> API don't necessarily specify the unit: sometimes they're milliseconds, sometimes they're seconds, you just have to read the docs (or JSDocs) to find out.

Even though many existing APIs doesn't do it, doesn't mean they shouldn't. Usually, the expectation in the JS world is that if no unit is specified, it's milliseconds. 🤷‍♂️

---

> cache.function(fn, {
	maxAge: 10,
	staleWhileRevalidate: 20
});

Sounds good. That should be familiar to people.

fregante:
This needs a little testing before I publish it. Also (separately) I'm thinking about a mechanism to force getting/refreshing the cached value. Example:

```js
if (userJustClosedAnIssue) {
	getBugsCount.refresh()
}
```

Issue #2755, opened 2020-02-04
Draft PR icon looks weird


Jamesking56:
The Draft PR icon (changed in #2742) looks really strange and doesn't resemble anything.

![image](https://user-images.githubusercontent.com/253237/73739862-f4461700-473e-11ea-9b4a-6e722db2d717.png)

I'm using Brave Version 1.2.43 Chromium: 79.0.3945.130 (Official Build) (64-bit) on Ubuntu 18.04 LTS

fregante:
I think you should report that to Brave, it's a plain `filter: drop-shadow()`

fregante:
If you want a quick fix to undo this change, use this CSS:

```css
:root .js-issue-row [aria-label="Open draft pull request"] svg {
	color: #6a737d !important;
	filter: none !important;
}
```

jdreesen:
It looks strange in Firefox, too: 
![image](https://user-images.githubusercontent.com/424602/73743877-d7fea600-474f-11ea-8789-36b9debcab2f.png)

I'm using Firefox 72.0.2 (64-Bit) on Ubuntu 18.04 LTS

Jamesking56:
Would it not be better (and more supported) to use `box-shadow`?

fregante:
`box-shadow` only works on a square, not on the "visible pixels" like `filter`

fregante:
Reopening to discuss alternatives and ask for help to create this SVG:

> [A solid] outline would be best, but there's no way to do that in CSS. I'd have to recreate the icon with the outline and embed the whole new SVG. Anyone wants to take a crack at creating the style in Illustrator/Sketch/etc?

Jamesking56:
Makes more sense to do a solid outline directly in the SVG :+1:

FloEdelmann:
What about this CSS?

```css
.js-issue-row [aria-label='Open draft pull request'] svg {
  stroke: #6a737d; /* GitHub's default draft PR color */
  stroke-width: 2px;
  color: #fff !important;
  paint-order: stroke;
  overflow: visible !important;
}
```

![grafik](https://user-images.githubusercontent.com/202916/73761806-a1d11e80-476f-11ea-8ea3-0fb49d37f90b.png)

fregante:
That's really interesting, if lighter, but there's a small artifact near the arrow that kinds ruins it.

[...]

PR #2788, opened 2020-02-15
Change icon for `mark-private-orgs`

fregante:
I think the "invisible" icon makes more sense than the previous "locked"/"secure" icon. Also this lets us drop the custom inlined icon


sindresorhus:
Confirmed working, but the white background is not round and nice in Chrome:

Screenshot 2020-02-16 at 18 08 05

fregante:
Maybe it needs `important` on the `overflow` in Chrome. I copied the CSS from #2755

PR #2789, opened 2020-02-15
Faster `mark-private-orgs`

fregante:
It didn't make sense to follow #2671 since this only adds some overlays on a minor part of the page, but at least this adds caching and changes the API to the lighter v4


fregante:
This is ready to merge. To test it, clear your cache and visit your profile

PR #2790, opened 2020-02-15
Faster `profile-gists-link`

fregante:
Context: #2671 

- Adds cache
- Switches from API v3 to v4
- Uses `element-ready`

---

Discussion: should the tab be added at all if there are no gists? I don't think so, but this might make the feature "slower" since it would _have to_ wait for the API to respond.

I probably never used this link in the first place, so 🤷‍♂️ . cc @sindresorhus

### DIFF
--- a/source/features/profile-gists-link.tsx
+++ b/source/features/profile-gists-link.tsx
@@ -1,13 +1,27 @@
 import './profile-gists-link.css';
 import React from 'dom-chef';
-import select from 'select-dom';
+import elementReady from 'element-ready';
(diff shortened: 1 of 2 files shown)


yakov116:
@fregante I have used this multiple times!!

sindresorhus:
> should the tab be added at all if there are no gists?

Yes, I think it's better for it to be predictable. It can just show `0`.

> I probably never used this link in the first place, so 

I use it all the time. I even forgot it is part of RG and not native.

PR #2794, opened 2020-02-16
Improve settings in dark mode in Firefox

### LINKED ISSUE #2791: Improve dark mode style in Firefox (Options page)
Our options page supports dark mode, but its colors are modeled after Chrome and not Firefox. You can see the awkward "different gray" background:

---

PR welcome to fix this elegantly

golota60:
Closes #2791 

Changing body background color based on the browser you're using.
Tested on Win10 using Firefox and Chrome.

# Test

image

### DIFF
--- a/source/options.css
+++ b/source/options.css
@@ -157,6 +157,10 @@ h2 ~ h2 {
 		margin-right: 6px;
 	}
 
+	:root {
+		background: none !important;
+	}
+
 	@media (prefers-color-scheme: dark) {
 		:root {
 			--background-color: #201f24;
--- a/source/options.css
+++ b/source/options.css
@@ -161,8 +161,9 @@ h2 ~ h2 {
 		background: none !important;
 	}
 
-	html, body {
-		--background-color: #201f24;;
+	html,
+	body {
+		--background-color: #201f24;
 	}
 
 	:root h2 {
--- a/source/options.css
+++ b/source/options.css
@@ -24,7 +24,7 @@ html {
 		--input-border-color: rgba(255, 255, 255, 0.3);
 	}
 	.is-firefox {
-		--background-color: #201F24;
+		--background-color: #201f24;
 	}
 }
 
--- a/source/options.css
+++ b/source/options.css
@@ -23,10 +23,6 @@ html {
 		--link-color: #8ab4f8;
 		--input-border-color: rgba(255, 255, 255, 0.3);
 	}
-
-	.is-firefox {
-		--background-color: #201f24;
-	}
 }
 
 body {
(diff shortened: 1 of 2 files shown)


source/options.tsx
```diff
@@ -130,6 +130,10 @@ async function init(): Promise<void> {
 			select<HTMLAnchorElement>('#personal-token-link')!.host = dropdown.value;
 		});
 	}
+
+	if(navigator.userAgent.indexOf('Firefox') > -1) {
```
yakov116: use `isFirefox`
see https://github.com/sindresorhus/refined-github/blob/4c502f50b45d339f9eb6fb3964b4eabc7ba7a36f/source/features/embed-gist-inline.tsx#L5

source/options.tsx
```diff
@@ -130,6 +130,10 @@ async function init(): Promise<void> {
...
+		select<HTMLBodyElement>('body')!.style.backgroundColor = '#201F24';
```
sindresorhus: I think it would be better to inject a `.is-firefox` class into the body so the CSS can stay in CSS and it's more reusable for the future.


Also: 
```suggestion
		document.body.style.backgroundColor = '#201F24';
```

source/options.css
```diff
@@ -23,6 +23,10 @@ html {
 		--link-color: #8ab4f8;
 		--input-border-color: rgba(255, 255, 255, 0.3);
 	}
+
+	.is-firefox {
+		--background-color: #201f24;
+	}
```
fregante: This doesn’t need JavaScript, there’s a media query just for firefox just below. 

Also, do you think that `html, body {background: none !important}` works in either browser? I’d prefer that if you can test it

golota60:
```html, body {background: none !important}``` - definitely does not work in firefox, here's the result

image

Same thing with Chrome
image

#201f24 - we use the same color for fields as firefox does(here's a screenshot from proxy settings in firefox), although their options pop-ups have lighter backgrounds

image

**Comment by golota60** [comment]
I've left the fields color as it is. Let me know if you want a lighter shade for fields

**Comment by fregante** [comment]
> Let me know if you want a lighter shade for fields anyway

Yes, we still need some contrast. They should either be lighter or darker

File: source/options.css [hunk 4]
```diff
@@ -161,6 +161,11 @@ h2 ~ h2 {
 		background: none !important;
 	}
```
fregante: This is now duplicate

source/options.css
```diff
@@ -161,6 +161,11 @@ h2 ~ h2 {
...
 
+	html,
+	body {
```
fregante: This should just be `:root` like the rest of the variables

Issue #10, opened 2016-02-16
Disable certain refinements


jonashaag:
I'd like to disable the "hide comment box toolbar" refinement. Would be nice if there was a checkbox of each refinement to trigger whether it's enabled or disabled.

sindresorhus:
See: https://github.com/sindresorhus/refined-github#contribute

I'm not interested in adding options. It's a lot more work, not including all the dumb boilerplate and hassle Chrome forces you to go through to add options. I would personally recommend just learning Markdown. It's a super simple syntax. Anyone is free to adapt the code here to do so, or you could just fork and modify it yourself ;)

PR #2805, opened 2020-02-19
Fix issue count for repos with over 1000 issues

### LINKED ISSUE #2780: Issue count appears as "NaN" for repositories with more than 5,000 open issues
- **Browser:** Firefox 72
- **Refined GitHub version:** 20.2.14

Since GitHub displays issue counts above 5,000 as "5,000+", it seems Refined GitHub fails to parse this and displays "NaN" instead. It should probably be changed to subtract the number of bugs from 5,000 and display a trailing `+` to indicate the number of issues is approximate. This happens on https://github.com/godotengine/godot, specifically.

image

cajames:
@fregante This solution goes down the path of querying all open issues and bugs in the one query, then using those values for the bugs and issues tabs. Let me know what you think. :)

Closes #2780

Test URLs:

- https://github.com/typeorm/typeorm
- https://github.com/godotengine/godot
- https://github.com/golang/go (repo with over 5000 open issues, this has none tagged `bug`, so won't render the tab)

Screenshots:

Screen Shot 2020-02-19 at 3 59 01 pm
Screen Shot 2020-02-19 at 3 40 00 pm
Screen Shot 2020-02-19 at 4 29 35 pm

<!-- 

Thanks for contributing! 🍄

1. LINKED ISSUES:

2. TEST URLS:
   Add some test URLs

3. SCREENSHOT:

### DIFF
--- a/source/features/bugs-tab.tsx
+++ b/source/features/bugs-tab.tsx
@@ -7,16 +7,18 @@ import * as api from '../libs/api';
 import SearchQuery from '../libs/search-query';
 import {getRepoGQL, getRepoURL} from '../libs/utils';
 
-const countBugs = cache.function(async (): Promise<number> => {
+const countIssuesAndBugs = cache.function(async (): Promise<{totalCount: number; bugsCount: number}> => {
 	const {repository} = await api.v4(`
 		repository(${getRepoGQL()}) {
 			bugs: issues(labels: ["bug"], states: [OPEN]) {
 				totalCount
 			}
+			open: issues(states: [OPEN]) {
+				totalCount
+			}
 		}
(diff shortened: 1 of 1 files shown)


fregante:
Thank you for the PR! Can you test the query performance on repos that have 5000+ issues? I wouldn’t like it if it takes seconds to complete, I’d rather just show 4000+ (With 1000 bugs)

fregante:
In the meanwhile, I'd propose an easier solution: #2813

fregante:
Closing due to change of direction https://github.com/sindresorhus/refined-github/issues/2844#issuecomment-592662373

PR #2809, opened 2020-02-19
Restore `filter-comments-by-you` on GHE

paschdan:
in github enterprise the old html template is still in use

at least for Enterprise v2.18.8

hint, i am using `getUserName()` instead of `@me`, in enterprise `@me` is not available.

### DIFF
--- a/source/features/filter-comments-by-you.tsx
+++ b/source/features/filter-comments-by-you.tsx
@@ -5,7 +5,7 @@ import SearchQuery from '../libs/search-query';
 
 function init(): void {
 	// Use an existing dropdown item to preserve its DOM structure (supports old GHE versions)
-	const sourceItem = select([
+	const sourceItem = select<HTMLAnchorElement>([
 		'#filters-select-menu a:last-child', // GHE
 		'.subnav-search-context li:last-child'
 	])!;
--- a/source/features/filter-comments-by-you.tsx
+++ b/source/features/filter-comments-by-you.tsx
@@ -1,18 +1,22 @@
 import select from 'select-dom';
 import features from '../libs/features';
-import {getUsername, getRepoURL} from '../libs/utils';
-
-const repoUrl = getRepoURL();
+import {getUsername} from '../libs/utils';
+import SearchQuery from '../libs/search-query';
 
 function init(): void {
-	const dropDownElement = select('#filters-select-menu a:last-child, .subnav-search-context li:last-child')!;
-	const newDropDownElement = dropDownElement.cloneNode(true);
-	const href = `/${repoUrl}/issues?q=is%3Aopen+commenter:${getUsername()}`;
-	const newDropDownElementLink = newDropDownElement.matches('a') ? newDropDownElement : select('a', newDropDownElement)!;
-	newDropDownElementLink.textContent = 'Everything commented by you';
-	newDropDownElementLink.setAttribute('href', href);
-	newDropDownElementLink.removeAttribute('target');
-	dropDownElement.before(newDropDownElement);
+	// Use an existing dropdown item to preserve its DOM structure (supports old GHE versions)
+	const sourceItem = select([
+		'#filters-select-menu a:last-child', // GHE
+		'.subnav-search-context li:last-child'
+	])!;
+
+	const menuItem = sourceItem.cloneNode(true);
+	const link = select('a', menuItem) ?? menuItem;
+	link.textContent = 'Everything commented by you';
+	link.removeAttribute('target');
+	new SearchQuery(link).set(`is:open commenter:${getUsername()}`);
+
+	sourceItem.before(menuItem);
 }
 
 features.add({
--- a/source/features/filter-comments-by-you.tsx
+++ b/source/features/filter-comments-by-you.tsx
@@ -1,21 +1,23 @@
-import React from 'dom-chef';
 import select from 'select-dom';
 import features from '../libs/features';
-import {getRepoURL} from '../libs/utils';
+import {getUsername, getRepoURL} from '../libs/utils';
 
 const repoUrl = getRepoURL();
 
 function init(): void {
-	select('.subnav-search-context .SelectMenu-list a:last-child')!
-		.before(
-			<a
-				href={`/${repoUrl}/issues?q=is%3Aopen+commenter:@me`}
-				className="SelectMenu-item"
-				role="menuitem"
-			>
-				Everything commented by you
-			</a>
-		);
+	const dropDownElement = select('#filters-select-menu a:last-child, .subnav-search-context li:last-child');
+	// Exit if not logged in or selector does not match
+	if (!dropDownElement) {
+		return;
+	}
+
+	const newDropDownElement = dropDownElement.cloneNode(true);
+	const href = `/${repoUrl}/issues?q=is%3Aopen+commenter:${getUsername()}`;
+	const newDropDownElementLink = newDropDownElement.matches('a') ? newDropDownElement : select('a', newDropDownElement)!;
+	newDropDownElementLink.textContent = 'Everything commented by you';
+	newDropDownElementLink.setAttribute('href', href);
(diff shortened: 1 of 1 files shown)


source/features/filter-comments-by-you.tsx
```diff
@@ -1,21 +1,38 @@
...
-		.before(
-			<a
-				href={`/${repoUrl}/issues?q=is%3Aopen+commenter:@me`}
-				className="SelectMenu-item"
-				role="menuitem"
-			>
-				Everything commented by you
-			</a>
-		);
+	if (isEnterprise()) {
```
busches: There is a lot of duplication here. The key differences here are:
- the selector: this can be joined together
- the `<li>` wrapper: this isn't needed on 2.19
- The classname of `SelectMenu-Item`: this can be applied to both if needed and again, this markup is different on 2.19

This needs to be rewritten in a more generic way to apply to GH and differing versions of GHE.

paschdan: Yeah I thought about writing GE version retriever, and do it more generic then

busches: We don't want to go down the route of coding to specific versions of GHE. Code so it works for both and eventually GHE will catch up and that code can be removed.

paschdan: the question is, do we want to support old versions of GHE. Because some enterprises may keep a specific GHE version for a longer time although there are maybe like 5 newer versions available, so the catch up will not happen

busches: Supporting older is fine, but it needs to be done in a better way than using `isEnterprise`.

paschdan: but the classname `SelectMenu-Item` is unfortunatly not usable for both, `SelectMenu-Item` (new) vs `select-menu-item` (old)

i am trying to find a possibility

busches: You can pass two selectors to the select, this is better than two branching if statements with 99% of the code the same. 😀

source/features/filter-comments-by-you.tsx
```diff
@@ -1,21 +1,38 @@
...
+import {isEnterprise} from '../libs/page-detect';
 
 const repoUrl = getRepoURL();
 
 function init(): void {
-	select('.subnav-search-context .SelectMenu-list a:last-child')!
-		.before(
-			<a
-				href={`/${repoUrl}/issues?q=is%3Aopen+commenter:@me`}
-				className="SelectMenu-item"
```
fregante:
```js
className="SelectMenu-item select-menu-item" 
```

source/features/filter-comments-by-you.tsx
```diff
@@ -1,21 +1,38 @@
...
 import select from 'select-dom';
 import features from '../libs/features';
-import {getRepoURL} from '../libs/utils';
+import {getUsername, getRepoURL} from '../libs/utils';
+import {isEnterprise} from '../libs/page-detect';
 
 const repoUrl = getRepoURL();
 
 function init(): void {
-	select('.subnav-search-context .SelectMenu-list a:last-child')!
```
fregante:
```js
select([
	'.subnav-search-context .SelectMenu-list a:last-child',
	'.subnav-search-context details-menu li:last-child' // Old selector, for GHE
])! 
```

source/features/filter-comments-by-you.tsx
```diff
@@ -1,21 +1,38 @@
...
-			>
-				Everything commented by you
-			</a>
-		);
+	if (isEnterprise()) {
+		select('.subnav-search-context details-menu li:last-child')!
+			.before(
+				<li>
+					<a
+						href={`/${repoUrl}/issues?q=is%3Aopen+commenter:${getUsername()}`}
```
fregante: And this one, as you wrote it

source/features/filter-comments-by-you.tsx
paschdan: not quite, as we would miss the `<li>` in case of the old version, i will provide a solution, that works for both, and we can continue the discussion on the new solution :)

paschdan:
@fregante @busches I decided to rewrite this, so it copies an existing list element, overwrites the href and text and then prepends it.

this way it works for a list of`<a/>` tags and `<li/>` tags.

source/features/filter-comments-by-you.tsx
```diff
@@ -1,21 +1,23 @@
...
-			<a
-				href={`/${repoUrl}/issues?q=is%3Aopen+commenter:@me`}
-				className="SelectMenu-item"
-				role="menuitem"
-			>
-				Everything commented by you
-			</a>
-		);
+	const dropDownElement = select('#filters-select-menu a:last-child, .subnav-search-context li:last-child');
+	// Exit if not logged in or selector does not match
```
busches: RGH doesn't run when you're not logged in and we'd want an error in the console so it's easier to spot when the selector changes to fix the feature.

busches:
I think this approach is fine, but I don't have access to an older GHE to test.

source/features/filter-comments-by-you.tsx
```diff
@@ -1,21 +1,18 @@
...
-				Everything commented by you
-			</a>
-		);
+	const dropDownElement = select('#filters-select-menu a:last-child, .subnav-search-context li:last-child')!;
+	const newDropDownElement = dropDownElement.cloneNode(true);
+	const href = `/${repoUrl}/issues?q=is%3Aopen+commenter:${getUsername()}`;
+	const newDropDownElementLink = newDropDownElement.matches('a') ? newDropDownElement : select('a', newDropDownElement)!;
+	newDropDownElementLink.textContent = 'Everything commented by you';
+	newDropDownElementLink.setAttribute('href', href);
+	newDropDownElementLink.removeAttribute('target');
```
fregante: Is it better to just copy the first link since it doesn’t have this attribute in the first place?

source/features/filter-comments-by-you.tsx
```diff
@@ -1,21 +1,18 @@
...
-		.before(
-			<a
-				href={`/${repoUrl}/issues?q=is%3Aopen+commenter:@me`}
-				className="SelectMenu-item"
-				role="menuitem"
-			>
-				Everything commented by you
-			</a>
-		);
+	const dropDownElement = select('#filters-select-menu a:last-child, .subnav-search-context li:last-child')!;
```
fregante:
```suggestion
	const dropDownElement = select([
		'#filters-select-menu a:last-child', // For GHE compatibility
		'.subnav-search-context li:last-child'
	])!;
```

source/features/filter-comments-by-you.tsx
```diff
@@ -1,21 +1,18 @@
...
+	const newDropDownElement = dropDownElement.cloneNode(true);
+	const href = `/${repoUrl}/issues?q=is%3Aopen+commenter:${getUsername()}`;
```
fregante: It's already cached, no need for another variable

```suggestion
	const href = `/${getRepoURL()}/issues?q=is%3Aopen+commenter:${getUsername()}`;
```

fregante:
I made some changes directly instead of writing further comments. Untested though

source/features/filter-comments-by-you.tsx
paschdan: and yes, we need to copy the first link, otherwise we just add the `is:open commenter:${getUsername()}` to a link targeting the external url `https://help.github.com/enterprise/2.18/user/articles/searching-issues`

paschdan:
@fregante now it is working even though the type from the first select is not really always a `HTMLAnchorElement`

fregante:
>  the type from the first select is not really always a HTMLAnchorElement

Indeed, I wasn’t sure if `menuItem as HTMLAnchorElement` would work right away (I’m editing from my phone) and just used my solution. Feel free to use that assertion if it works. Otherwise I prefer a white lie to “more code”

PR #2819, opened 2020-02-20
Revert usage of `@me` search keyword to restore GHE support

paschdan:
The `@me` filter is not available at enterprise github at least not for v2.18.8

i left out commented-by-you feature as there is #2809 

<!-- 

Thanks for contributing! 🍄

1. LINKED ISSUES:

2. TEST URLS:
   Add some test URLs

3. SCREENSHOT:

### DIFF
--- a/source/features/global-discussion-list-filters.tsx
+++ b/source/features/global-discussion-list-filters.tsx
@@ -2,6 +2,7 @@ import './global-discussion-list-filters.css';
 import React from 'dom-chef';
 import select from 'select-dom';
 import features from '../libs/features';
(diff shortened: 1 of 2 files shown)


busches:
I tested `@me` yesterday against 2.19, it's not there.

fregante:
There’s no point in keeping `@me` if we want to keep supporting GHE. The regular username works in both

paschdan:
so @fregante you suggest to revert the `@me` to `getUsername()` then?

fregante:
@paschdan correct

fregante:
Thank you @paschdan ^_^

PR #2825, opened 2020-02-22
Add 3 `file-finder`-related features

### LINKED ISSUE #2467: "File Finder" improvements
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/refined-github/issues/2467)

1. Visit https://github.com/sindresorhus/refined-github
2. Press `t`
3. A file finder will load

There are 3 improvements that can be made:

- `t` only works while you're on `/tree/` and `/blob/` pages; it should work on every page of the repo, including issues.
- When **navigating back** to the File Finder (e.g. open finder, open file from list, go back to finder), the previous input is lost; this should instead be preserved.
- After pressing `t`, you have to wait for the page to load before typing a file name; usually software will buffer your keyboard input even if the UI takes a while to load; we should simulate the same behavior so you can type `t` and start typing your query right away.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $15.16)

- [fregante fregante](https://issuehunt.io/u/fregante) ($15.16)

### Submitted pull Requests
- [#2825 File Finder improvements](https://issuehunt.io/r/sindresorhus/refined-github/pull/2825)
---

### Tips

---
</details>

danedavid:
Closes #2467. Adds 2 new features, which are improvements to the `t` file finder:

1. **Repo-wide file finder shortcut** (`repo-wide-file-finder`): Currently, GH supports the `t` shortcut only in paths with `/blob` or `/tree` in them. Added a hidden button with `data-hotkey` so that the `t` works throughout other tabs in the repo, (eg: Issues, Releases, etc.)

2. **Preserves search term in file finder** (`preserve-file-finder-term.tsx`): The search term in file finder is persisted ( in local storage ) across file finder invocations. Also, the keys pressed immediately following the `t` are persisted in a buffer and added to the search term. If buffer is present, the buffer is used, otherwise the value from previous search is used.

Kindly review and let me know if the features need to be added elsewhere (`README`, etc.). Thanks!

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#2467: "File Finder" improvements](https://issuehunt.io/repos/51769689/issues/2467)
---

</details>


source/features/repo-wide-file-finder.tsx
```diff
@@ -0,0 +1,52 @@
...
+				data-hotkey="t"
+				data-pjax="true"
+				href={`/${getRepoURL()}/find/master`}
+			>
+				Hidden Find
+			</a>
+		</li>
+	);
+	if (!fileFinderButtonExists) {
+    pageHeaderElement?.appendChild(hiddenButton);
```
fregante: This link doesn’t have to be public. We have two options:
- add it to `body` with the `hidden` attribute
- add it to `head` as a `link` rather than an anchor, but I don’t know if that’ll work

source/features/repo-wide-file-finder.tsx
```diff
@@ -0,0 +1,52 @@
...
+function init(): void {
+	const pageHeaderElement = select<HTMLUListElement>('.pagehead-actions');
+	const fileFinderButtonExists = select.exists('[data-hotkey="t"]');
+	const hiddenButton: HTMLElement = (
+		<li className="rgh-file-finder-hidden-btn">
+			<a
+				className="btn btn-sm empty-icon float-right BtnGroup-item"
+				data-hotkey="t"
+				data-pjax="true"
+				href={`/${getRepoURL()}/find/master`}
```
fregante: You can use `getDefaultBranch()` because not all repos use master

source/features/preserve-file-finder-term.tsx
```diff
@@ -0,0 +1,120 @@
...
+	id: __featureName__,
+	description: 'Preserve search term in file finder',
+	screenshot: false,
+	include: [
+		features.isRepo
+	],
+	load: features.onAjaxedPages,
+	init,
+	deinit
+});
```
fregante: # Pre-load buffer

The buffer only makes sense between the time the user starts navigating to FF (file finder) and the time it loads. This period can be detected by listening to GitHub’s PJAX events. I’m not sure, but maybe their “pjax start” event also includes which page is about to load, so this period can be detected *exactly.*

Once this buffer is added to the field, it should be discarded immediately.

This feature can only work if FF only loads via ajax, every time. If it doesn’t, this feature isn’t feasible: we’d lose key presses between the time the page unloads and our content scripts load on the new page. 

Because of this, the buffer can be stored in a regular `let` variable.

source/features/preserve-file-finder-term.tsx
```diff
@@ -0,0 +1,120 @@
...
+features.add({
+	id: __featureName__,
+	description: 'Preserve search term in file finder',
+	screenshot: false,
+	include: [
+		features.isRepo
+	],
+	load: features.onAjaxedPages,
+	init,
+	deinit
```
fregante: Read the buffer’s note first. I think this file should be split in two features because they’re completely different. 

# Text restoration after going `back`

This feature only needs two handlers:
- User navigates to a file in FF (maybe detectable via `pjax:start` event)
- FF loads

In the first event, you can just get the value of the field and store it in sessionStorage (which avoids race conditions since its synchronous). IF the location.href hasn’t changed at this point, you can use `history.replaceState` to store this value. This is the perfect place to do so. 

In the second event, you can just check the value and restore it. 

## Bonus points 😃 

GitHub has their own “resumable field” feature that they use on comments and issue titles. Maybe you can use it and GitHub will handle all of this automatically. Look at the exactly class in their code. 

There’s a chance that this doesn’t work if GitHub’s “resumable field” code runs before we add the class.

source/features/repo-wide-file-finder.tsx
danedavid: Sorry, but I don't understand by what you mean by "This link doesn't have to be public". I'm creating a button and appending to only those pages where github doesn't already have the button. Could you kindly explain what advantage is expected by taking any of the approaches above?

source/features/preserve-file-finder-term.tsx
danedavid: In my tests, the FF is loaded via PJAX only when navigated from Code, Issues & Pull Requests tab. Initially I was using a simple `let` variable just like you said. Then I went with persisting the buffer in storage to store key presses when the file finder is not loaded via PJAX.  
Another issue I stumbled across was key presses coinciding with other shortcuts, and the FF not opening up.

TL;DR: FF is not loaded via ajax every time. Should we drop this feature, and just continue with text restoration in FF?

fregante: Code, issues and PRs covers most situations. I think it’s worth implementing it in these cases. This implementation should be easier than before since you can just read write synchronously

source/features/repo-wide-file-finder.tsx
fregante: I meant: You’re appending it to a specific element and adding text inside it. If the link isn’t visible, you can just add it with 

```js
if (!select.exists('[data-hotkey="t"]') {
  document.body.append(<a hidden …/>);
}
```

The visual classes (`btn`), the text content and the `li` around it aren’t necessary. The user is not meant to see any of this.

danedavid:
@fregante 
PR updated after review. Couple of notes:

- Repo-wide `t` works only from Issues & Pull Requests tab. This is because only those 2 are inside the `pjax-container` element given by Github. Since now we can be sure that FF is always loaded via pjax, I'm using a normal variable as buffer like you said
- For text restoration, tried using Github's inbuilt `js-session-resumable` class. Unfortunately, adding the class didn't work out; probably the code that handles it runs before that, like you had hinted. So went with storing in sessionStorage.
- Should have been 3 PRs, but initially didn't have the idea they were gonna be 3 different features. Hence single PR 😄

source/features/repo-wide-file-finder.tsx
fregante: The code I showed is exactly all you need inside `init`. You can use `await getDefaultBranch()` directly inside the JSX tag, no need to create **any** variables. 

`deinit` isn’t necessary.

source/features/repo-wide-file-finder.tsx
```diff
@@ -0,0 +1,52 @@
...
+	include: [
+		features.isRepoDiscussionList,
+		features.isPR,
+		features.isIssue
+	],
+	exclude: [
+		features.isRepoRoot,
+		features.isRepoTree,
+		features.isSingleFile,
+		features.isFileFinder
```
fregante: What you listed in `include` doesn’t include these pages, so you don’t need to specify anything here.

source/features/file-finder-buffer.tsx
```diff
@@ -0,0 +1,72 @@
...
+  char <= 'Z'
+) || (
+	char >= '0' &&
+  char <= '9'
+));
+
+const keyDownHandler = ({key, target}: KeyboardEvent): void => {
+	const nodeName = (target as Element).nodeName;
+	if (nodeName !== 'INPUT' && nodeName !== 'TEXTAREA' && isValidCharacter(key)) {
+		fileFinderBuffer = `${fileFinderBuffer}${key}`;
```
fregante: This method won’t work as well as an actual input field, for example the backspace and non-latin characters don’t work.

What if you create a hidden `type=text` field (with `visibility:hidden`), focus it and have it receive all the input natively? When the actual FF’s field is available, either copy the value or swap the fields (`ghField.replaceWith(ourField)`, ideal, so it works without interruptions)

The advantage of this is that ****later**** we can make it visible as soon as the user presses `t`. But the styling for this is harder, so keep it hidden for now. I’ll design it next month

source/features/file-finder-buffer.tsx
```diff
@@ -0,0 +1,72 @@
...
+	id: __featureName__,
+	description: 'Buffering for search term pressed after `t`',
+	screenshot: false,
+	include: [
+		features.isRepoDiscussionList,
+		features.isPR,
+		features.isIssue,
+		features.isRepoRoot,
+		features.isRepoTree,
+		features.isSingleFile
```
fregante: You can just use `isRepo` here. The `/find/` test in the pjax handler will exclude everything else

source/features/preserve-file-finder-term.tsx
```diff
@@ -0,0 +1,48 @@
+import features from '../libs/features';
+import select from 'select-dom';
+import elementReady from 'element-ready';
+
+const FILE_FINDER_INPUT_SELECTOR = '.js-tree-finder > .breadcrumb > #tree-finder-field';
+
+const unloadHandler = (): void => {
+	const inputElement = select<HTMLInputElement>(FILE_FINDER_INPUT_SELECTOR);
```
fregante: The selector can be inlined here, it doesn’t need to be a constant. In the other feature as well.

source/features/file-finder-buffer.tsx
```diff
@@ -0,0 +1,53 @@
+import './file-finder-buffer.css';
+import React from 'dom-chef';
+import features from '../libs/features';
+import select from 'select-dom';
+
+const FILE_FINDER_INPUT_SELECTOR = '.js-tree-finder > .breadcrumb > #tree-finder-field';
+
+const pjaxStartHandler = ((event: CustomEvent): void => {
+	const destinationURL: string = event.detail?.url || '';
+	if (destinationURL.includes('/find/')) {
```
fregante: Use destinationURL.split() and test the 3rd item specifically; otherwise this simple `includes` will return true on the `sindresorhus/find` repo

source/features/preserve-file-finder-term.tsx
```diff
@@ -0,0 +1,48 @@
...
+		sessionStorage.setItem('rgh-file-finder-term', inputElement.value);
+	}
+};
+
+// Set the input field value & trigger event
+const setValueInField = async (): Promise<void> => {
+	const value: string = sessionStorage.getItem('rgh-file-finder-term') ?? '';
+	const inputElement = select<HTMLInputElement>(FILE_FINDER_INPUT_SELECTOR);
+	if (inputElement) {
+		if (!inputElement.value && value) {
```
fregante: Nested conditions unnecessary

source/features/repo-wide-file-finder.tsx
```diff
@@ -0,0 +1,31 @@
...
+async function init(): Promise<void> {
+	if (!select.exists('[data-hotkey="t"]')) {
+    select('#js-repo-pjax-container')?.appendChild(
+	<a
+		hidden
+		data-hotkey="t"
+		data-pjax="true"
+		href={`/${getRepoURL()}/find/${await getDefaultBranch()}`}
+	/>
+		);
```
fregante: Also ensure that the indentation is correct everywhere. Normally the linter should fix that but I don’t know why it’s not. I’m seeing it misaligned it here

source/features/file-finder-buffer.tsx
```diff
@@ -0,0 +1,53 @@
...
+		const hiddenInput = select<HTMLInputElement>('#rgh-file-finder-buffer') ?? document.body.insertBefore(
```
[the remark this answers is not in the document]
danedavid: This was intentionally written so that instead of adding/removing it every time, we could continue using the input field once it is created. The same reason why I chose to empty it in pjax:complete instead of removing it.

source/features/repo-wide-file-finder.tsx
danedavid: This happened after running `npm run lint-fix`. I thought it was some issue with my editor. Don't know how to fix this as whatever I do, linting puts it back at this.

source/features/file-finder-buffer.css
```diff
@@ -0,0 +1,13 @@
...
+*/
+.visuallyhidden {
+	position: absolute;
+	overflow: hidden;
+	clip: rect(0 0 0 0);
+	height: 1px;
+	width: 1px;
+	margin: -1px;
+	padding: 0;
+}
```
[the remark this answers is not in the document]
danedavid: Worked like a charm! Thanks!

source/features/preserve-file-finder-term.tsx
fregante: I still see `FILE_FINDER_INPUT_SELECTOR` in both files

danedavid: Ok, I'd thought you meant the constant `inputElement` was not required. My bad.

source/features/file-finder-buffer.tsx
```diff
@@ -0,0 +1,43 @@
...
+	}
+}) as EventListener; // Explicit type cast. See https://github.com/microsoft/TypeScript/issues/28357#issuecomment-436484705
+
+const pjaxCompleteHandler = (): void => {
+	const fileFinderInput = select<HTMLInputElement>('.js-tree-finder > .breadcrumb > #tree-finder-field');
+	const hiddenInput = select<HTMLInputElement>('#rgh-file-finder-buffer');
+	if (hiddenInput && fileFinderInput) {
+		fileFinderInput.value = hiddenInput.value;
+		fileFinderInput.dispatchEvent(new Event('input')); // Manually trigger event to trigger search
+		document.body.removeChild(hiddenInput);
```
fregante: `hiddenInput.remove()`

source/features/file-finder-buffer.tsx
```diff
@@ -0,0 +1,43 @@
+import React from 'dom-chef';
+import features from '../libs/features';
+import select from 'select-dom';
+
+const pjaxStartHandler = ((event: CustomEvent): void => {
+	const destinationURL: string = event.detail?.url || '';
```
fregante: The type isn’t required here

source/features/preserve-file-finder-term.tsx
```diff
@@ -0,0 +1,44 @@
...
+const unloadHandler = (): void => {
+	const inputElement = select<HTMLInputElement>('.js-tree-finder > .breadcrumb > #tree-finder-field');
+	if (inputElement) {
+		sessionStorage.setItem('rgh-file-finder-term', inputElement.value);
+	}
+};
+
+// Set the input field value & trigger event
+const setValueInField = async (): Promise<void> => {
+	const value: string = sessionStorage.getItem('rgh-file-finder-term') ?? '';
```
fregante: `??` isn’t needed, let it be `undefined`, we’re checking for it in the condition anyway. The explicit type isn’t required here either.

source/features/preserve-file-finder-term.tsx
```diff
@@ -0,0 +1,43 @@
...
+	const inputElement = select<HTMLInputElement>(FILE_FINDER_INPUT_SELECTOR);
+	if (inputElement && !inputElement.value && value) {
+		await elementReady('.js-tree-browser-results > li', {stopOnDomReady: false});	// For search to work
+		inputElement.value = value;
+		inputElement.dispatchEvent(new Event('input')); // Manually trigger event to trigger search
+	}
+};
+
+async function init(): Promise<void> {
+	await setValueInField();
```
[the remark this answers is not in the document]
danedavid: It's needed for the `elementReady` function inside `setValueinField`. I'm waiting for the element tree to be ready because we need a list of files when the input is populated with the value, for the filtering to take place. Otherwise the value is filled in, but files not filtered.

source/features/file-finder-buffer.tsx
```diff
@@ -0,0 +1,43 @@
...
+			<input
+				type="text"
+				className="sr-only"
+				id="rgh-file-finder-buffer"
+			/>
+		);
+		document.body.prepend(hiddenInput);
+		hiddenInput.focus();
+	}
+}) as EventListener; // Explicit type cast. See https://github.com/microsoft/TypeScript/issues/28357#issuecomment-436484705
```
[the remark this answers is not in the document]
danedavid: This is because `pjax:start` is of type `CustomEvent`, and adding a listener for a custom event makes `window.addEventListener('pjax:start', pjaxStartHandler)` throw the following error: `No overload matches this call.` Searched it up and found the linked issue. Another workaround is to cast type of `event` as `any`.

source/features/preserve-file-finder-term.tsx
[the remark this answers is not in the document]
fregante: That’s irrelevant; all that this await is doing is delaying the listeners that follow. That happens inside `setValueInField` isn’t touched by this await

fregante:
Looks good so far! I’m not sure when I’ll get around testing the code though

source/features/file-finder-buffer.tsx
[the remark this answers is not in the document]
fregante: I think you just need to add both new events to https://github.com/sindresorhus/refined-github/blob/825f4fd51efaef62a5f131a83b2c7a2cc14116ee/source/globals.d.ts#L33

And it will work thanks to this generic type https://github.com/microsoft/TypeScript/blob/177713ef45f7714c91af79246b8b5e2b6bd59128/lib/lib.dom.d.ts#L5546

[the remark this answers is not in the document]
fregante: Or maybe add this to `global.d.ts`

```ts
interface EventSourceEventMap {
	'pjax:start': CustomEvent
	etc
}
```

fregante:
The interesting part is that GitHub already has a `js-pjax-restore-captured-input` class on the field but apparently it doesn't work. I guess/hope `preserve-file-finder-term` will have a short life.

fregante:
I hope to see this merged soon, it's only missing descriptions and screenshots

danedavid:
@fregante Yes, I'll look into the description & screenshots over the weekend.

readme.md
```diff
@@ -141,6 +141,9 @@ Thanks for contributing! 🦋🙌
 - [](# "reload-failed-proxied-images") [Retries downloading images that failed downloading due to GitHub limited proxying.](https://user-images.githubusercontent.com/14858959/64068746-21991100-cc45-11e9-844e-827f5ac9b51e.png)
 - [](# "forked-to") [Adds a shortcut to your forks next to the `Fork` button on the current repo.](https://user-images.githubusercontent.com/55841/64077281-17bbf000-cccf-11e9-9123-092063f65357.png)
 - [](# "repo-age") [Adds the age of the repository to the statistics bar.](https://user-images.githubusercontent.com/3848317/69494069-7d2b1180-0eb7-11ea-9aa1-d4194e566340.png)
+- [](# "repo-wide-file-finder") The <kbd>t</kbd> file finder works on the `Issues` and `Pull Requests` tabs as well.
+- [](# "file-finder-buffer") While invoking the file finder, keystrokes pressed immediately following the <kbd>t</kbd> are captured and populated in the search field.
+- [](# "preserve-file-finder-term") [The search term is preserved across multiple invocations of the <kbd>t</kbd> file finder](https://user-images.githubusercontent.com/1402241/75542106-1c811700-5a5a-11ea-8aa5-bea0472c59e2.gif)
```
fregante: This is the wrong screenshot

```suggestion
- [](# "preserve-file-finder-term") Preserves the search terms when navigating back and forth between the File Finder and the files.
```

readme.md
```diff
@@ -141,6 +141,9 @@ Thanks for contributing! 🦋🙌
 - [](# "reload-failed-proxied-images") [Retries downloading images that failed downloading due to GitHub limited proxying.](https://user-images.githubusercontent.com/14858959/64068746-21991100-cc45-11e9-844e-827f5ac9b51e.png)
 - [](# "forked-to") [Adds a shortcut to your forks next to the `Fork` button on the current repo.](https://user-images.githubusercontent.com/55841/64077281-17bbf000-cccf-11e9-9123-092063f65357.png)
 - [](# "repo-age") [Adds the age of the repository to the statistics bar.](https://user-images.githubusercontent.com/3848317/69494069-7d2b1180-0eb7-11ea-9aa1-d4194e566340.png)
+- [](# "repo-wide-file-finder") The <kbd>t</kbd> file finder works on the `Issues` and `Pull Requests` tabs as well.
+- [](# "file-finder-buffer") While invoking the file finder, keystrokes pressed immediately following the <kbd>t</kbd> are captured and populated in the search field.
```
fregante:
```suggestion
- [](# "file-finder-buffer") [Lets you start typing your search immediately after invoking the File Finder (<kbd>t</kbd>), instead of having you wait for it to load first.](https://user-images.githubusercontent.com/1402241/75542106-1c811700-5a5a-11ea-8aa5-bea0472c59e2.gif)
```

readme.md
```diff
@@ -141,6 +141,9 @@ Thanks for contributing! 🦋🙌
 - [](# "reload-failed-proxied-images") [Retries downloading images that failed downloading due to GitHub limited proxying.](https://user-images.githubusercontent.com/14858959/64068746-21991100-cc45-11e9-844e-827f5ac9b51e.png)
 - [](# "forked-to") [Adds a shortcut to your forks next to the `Fork` button on the current repo.](https://user-images.githubusercontent.com/55841/64077281-17bbf000-cccf-11e9-9123-092063f65357.png)
 - [](# "repo-age") [Adds the age of the repository to the statistics bar.](https://user-images.githubusercontent.com/3848317/69494069-7d2b1180-0eb7-11ea-9aa1-d4194e566340.png)
+- [](# "repo-wide-file-finder") The <kbd>t</kbd> file finder works on the `Issues` and `Pull Requests` tabs as well.
```
fregante:
```suggestion
- [](# "repo-wide-file-finder") Enables the the File Finder keyboard shortcut (<kbd>t</kbd>) on Issues and Pull Request pages as well.
```

source/features/file-finder-buffer.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+function init(): void {
+	window.addEventListener('pjax:start', pjaxStartHandler);
+	window.addEventListener('pjax:complete', pjaxCompleteHandler);
+}
+
+features.add({
+	id: __featureName__,
+	description: 'Buffering for search term pressed after `t`',
+	screenshot: false,
```
fregante:
```suggestion
	description: 'Lets you start typing your search immediately after invoking the File Finder (`t`), instead of having you wait for it to load first.',
	screenshot: 'https://user-images.githubusercontent.com/1402241/75542106-1c811700-5a5a-11ea-8aa5-bea0472c59e2.gif',
```

source/features/repo-wide-file-finder.tsx
```diff
@@ -0,0 +1,31 @@
...
+				data-pjax="true"
+				href={`/${getRepoURL()}/find/${await getDefaultBranch()}`}
+			/>
+		);
+	}
+}
+
+features.add({
+	id: __featureName__,
+	description: 'Enables file finder on `t` on Issues & Pull Requests tab',
```
fregante:
```suggestion
	description: 'Enables the the File Finder keyboard shortcut (`t`) on Issues and Pull Request pages as well.',
```

source/features/preserve-file-finder-term.tsx
```diff
@@ -0,0 +1,47 @@
...
+}
+
+function deinit(): void {
+	window.removeEventListener('beforeunload', unloadHandler);
+	window.removeEventListener('pjax:start', unloadHandler);
+}
+
+features.add({
+	id: __featureName__,
+	description: 'Preserve search term in file finder',
```
fregante:
```suggestion
	description: 'Preserves the search terms when navigating back and forth between the File Finder and the files.',
```

fregante:
I never thought this'd see the day!

PR #2828, opened 2020-02-23
Don't apply `shorten-links` on search results

kidonng:
I don't know if this is a GitHub bug or to highlight the matched keywords, URLs in repo description on the search page aren't completely linkified, and `shorten-links` makes it even more inaccessible:

image

Sure it will be better if we can fix this up, but it may be difficult to handle various situations. I think this should be implemented by GitHub itself first.

So this PR ~~add `linkifiedURLClass` to these URLs to make `shorten-links` skip them~~ disables `shorten-links` on search page:

image

Test URL: https://github.com/search?q=nodejs

### DIFF
--- a/source/features/shorten-links.tsx
+++ b/source/features/shorten-links.tsx
@@ -2,8 +2,15 @@ import select from 'select-dom';
 import {applyToLink} from 'shorten-repo-url';
 import features from '../libs/features';
 import {linkifiedURLClass} from '../libs/dom-formatters';
+import {isGlobalSearchResults} from '../libs/page-detect'
 
 function init(): void {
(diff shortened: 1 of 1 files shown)


fregante:
I see what’s happening: the search highlights the searched keyword, in this case `node`, breaking the link before their linkifier is applied.

Can you report it? Don’t mention Refined GitHub or else they might think it’s a compatibility issue rather than an actual linkification bug.

kidonng:
> To exclude these pages, use the `exclude` property in `features.add` instead

Makes sense 👍 I didn't notice there are no other links that could be shortened,.

fregante:
# :woot:

PR #2831, opened 2020-02-25
Fix `mark-private-orgs` showing on all organizations

kytwb:
Fixes https://github.com/sindresorhus/refined-github/issues/2830.
Details: https://github.com/sindresorhus/refined-github/issues/2830#issuecomment-591111897.

### DIFF
--- a/source/features/mark-private-orgs.tsx
+++ b/source/features/mark-private-orgs.tsx
@@ -24,7 +24,8 @@ async function init(): Promise<false | void> {
 
 	const publicOrganizations = await getPublicOrganizations(getUsername());
 	for (const org of orgs) {
-		if (!publicOrganizations.includes(org.pathname.slice(1))) {
+		// https://github.com/sindresorhus/refined-github/issues/2830#issuecomment-591111897
+		if (!publicOrganizations.includes(org.pathname.slice(15))) {
 			org.classList.add('rgh-private-org');
 			org.append(eyeClosedIcon());
 		}
--- a/source/features/mark-private-orgs.tsx
+++ b/source/features/mark-private-orgs.tsx
@@ -25,7 +25,7 @@ async function init(): Promise<false | void> {
 	const publicOrganizationsNames = await getPublicOrganizationsNames(getUsername());
 	for (const org of orgs) {
 		// https://github.com/sindresorhus/refined-github/issues/2830#issuecomment-591111897
-		if (!publicOrganizationsNames.includes(org.pathname.slice('/organizations/'.length))) {
+		if (!publicOrganizationsNames.includes(org.pathname.slice(1).replace('organizations/', ''))) {
 			org.classList.add('rgh-private-org');
 			org.append(eyeClosedIcon());
 		}
--- a/source/features/mark-private-orgs.tsx
+++ b/source/features/mark-private-orgs.tsx
@@ -24,8 +24,7 @@ async function init(): Promise<false | void> {
 
 	const publicOrganizationsNames = await getPublicOrganizationsNames(getUsername());
 	for (const org of orgs) {
-		// https://github.com/sindresorhus/refined-github/issues/2830#issuecomment-591111897
-		if (!publicOrganizationsNames.includes(org.pathname.slice(1).replace('organizations/', ''))) {
+		if (!publicOrganizationsNames.includes(org.pathname.replace(/^[/]organizations[/]/, ''))) {
 			org.classList.add('rgh-private-org');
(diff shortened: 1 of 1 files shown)


kytwb:
I am not sure if this change to the organizations URLs is globally available on GitHub or if we should handle both the old/new URLs?

source/features/mark-private-orgs.tsx
```diff
@@ -22,9 +22,10 @@ async function init(): Promise<false | void> {
 		return false;
 	}
 
-	const publicOrganizations = await getPublicOrganizations(getUsername());
+	const publicOrganizationsNames = await getPublicOrganizationsNames(getUsername());
 	for (const org of orgs) {
-		if (!publicOrganizations.includes(org.pathname.slice(1))) {
+		// https://github.com/sindresorhus/refined-github/issues/2830#issuecomment-591111897
+		if (!publicOrganizationsNames.includes(org.pathname.slice('/organizations/'.length))) {
```
busches: Looks like a typo here, you're missing a ) after organizations.

kytwb: It's to slice the value of `'/organizations/'.length`, equivalent to `.slice(15)`. Did that to enhance readability (not to have to guess where the 15 comes from, but now it leads to a syntax error assumption. The irony 😄

busches: I see, I was thinking it was an array at a quick glance, but it's a string. This approach won't be backwards compatible with GHE, I'm assuming it's rolled out for all of GH, and ideally we remove `/organizations/` in another way that would work even if it exists or not.

kytwb: Addressed via ff49bdb.

source/features/mark-private-orgs.tsx
```diff
@@ -22,9 +22,10 @@ async function init(): Promise<false | void> {
 		return false;
 	}
 
-	const publicOrganizations = await getPublicOrganizations(getUsername());
+	const publicOrganizationsNames = await getPublicOrganizationsNames(getUsername());
 	for (const org of orgs) {
-		if (!publicOrganizations.includes(org.pathname.slice(1))) {
+		// https://github.com/sindresorhus/refined-github/issues/2830#issuecomment-591111897
+		if (!publicOrganizationsNames.includes(org.pathname.slice(1).replace('organizations/', ''))) {
```
fregante: Either drop slice or replace. They can do the same thing. `replace(/^[/]organizations[/]/, '')` has clear intent

fregante:
Thank you for the fast PR!

source/features/mark-private-orgs.tsx
kytwb: Good idea! I have incorporated your feedback via 6e6af3a 🙏

fregante:
I'm not seeing these URLs yet; this change would break the feature for me.

kytwb:
@fregante The change has been reverted on my GitHub as well. I have updated the replace regex to work in both cases if it were to be restored.

source/features/mark-private-orgs.tsx
```diff
@@ -22,9 +22,9 @@ async function init(): Promise<false | void> {
 		return false;
 	}
 
-	const publicOrganizations = await getPublicOrganizations(getUsername());
+	const publicOrganizationsNames = await getPublicOrganizationsNames(getUsername());
 	for (const org of orgs) {
-		if (!publicOrganizations.includes(org.pathname.slice(1))) {
+		if (!publicOrganizationsNames.includes(org.pathname.replace(/^\/(organizations\/)?|\/$/, ''))) {
```
fregante: I think this is equivalent
```suggestion
		if (!publicOrganizationsNames.includes(org.pathname.replace(/^\/(organizations\/)?$/, ''))) {
```

kytwb: I thought so as well when I first read it (`/` followed by optional `organizations/`), but:

```
> '/organizations/getferdi'.replace(/^\/(organizations\/)?$/, '')
'/organizations/getferdi'
> '/getferdi'.replace(/^\/(organizations\/)?$/, '')
'/getferdi'
```

Then I noticed I mistakenly added the `$`:

```
> '/organizations/getferdi'.replace(/\/(organizations\/)?/, '')
'getferdi'
> '/getferdi'.replace(/\/(organizations\/)?/, '')
'getferdi'
```

kytwb: Addressed via bf1d679. Thanks for bearing with me! 😄

PR #1121, opened 2018-02-28
Update style for merge commits

bfred-it:
Fixes #1052

Test on https://github.com/babel/babel/commits/master?after=ddd40bf5c7ad8565fc990f26142f85613958a329+104

Notice: GitHub doesn't show the commit directly below the merge. The merged commit in this case is the third one, **Andarist**'s

merge commit screenshot

PR #2832, opened 2020-02-26
Add back `mark-merge-commits-in-list` feature

### LINKED ISSUE #2827: Add `dim-merge-commits` feature
This would solve the "noise" concern mentioned in #2768

It would have to use a combined GraphQL query to determine which commits on the current page are merge commits, and then just apply `opacity: 0.6` on their whole line. Or something like that.

"Combined GraphQL query" example: https://github.com/sindresorhus/refined-github/pull/2795/files#diff-04c19a6c246911e079b45f999eab501cR14-R22

### Test URL

https://github.com/babel/babel/commits/master?after=ddd40bf5c7ad8565fc990f26142f85613958a329+104

_Originally posted by @fregante in https://github.com/sindresorhus/refined-github/issues/2768#issuecomment-590036267_

yakov116:
closes #2827 

I hate typescript

I did not add the icon since typescript would not let me.
~wasted~ used about an hour because if it :confused:

image
https://github.com/babel/babel/commits/master?after=ddd40bf5c7ad8565fc990f26142f85613958a329+104

### DIFF
--- a/source/features/mark-merge-commits-in-list.tsx
+++ b/source/features/mark-merge-commits-in-list.tsx
@@ -4,13 +4,7 @@ import * as api from '../libs/api';
 import features from '../libs/features';
 import {getRepoGQL} from '../libs/utils';
 
-interface CommitInfo {
-	parents: {
-		totalCount: number | string;
-	};
-}
-
-const getCommitParentCount = async (commits: string[]): Promise<Record<string, CommitInfo>> => {
+const getMergeCommits = async (commits: string[]): Promise<string[]> => {
 	const {repository} = await api.v4(`
 		repository(${getRepoGQL()}) {
 			${commits.map((commit: string) => `
@@ -25,8 +19,19 @@ const getCommitParentCount = async (commits: string[]): Promise<Record<string, C
 		}
 	`);
 
-	return repository;
+	return filterCommits(repository);
 };
+
+function filterCommits(commits: object): string[] {
+	const mergeCommits = [];
+	for (const commit of Object.entries(commits)) {
+		if (commit[1].parents.totalCount === 2) {
+			mergeCommits.push(commit[0]);
+		}
+	}
+
+	return mergeCommits;
+}
 
 function getCommitHash(commit: HTMLElement): string {
 	return (commit.dataset.channel as string).split(':')[3];
@@ -34,11 +39,11 @@ function getCommitHash(commit: HTMLElement): string {
 
 async function init(): Promise<void | false> {
 	const pageCommits = select.all('li.commit');
-	const parentCommitCount = await getCommitParentCount(pageCommits.map(getCommitHash));
+	const parentCommitCount = await getMergeCommits(pageCommits.map(getCommitHash));
 	for (const commit of pageCommits) {
-		const commitHash = getCommitHash(commit);
-		const {totalCount} = parentCommitCount[api.escapeKey(commitHash)].parents;
-		if (totalCount === 2) {
+		const commitHash = api.escapeKey(getCommitHash(commit));
+
+		if (parentCommitCount.includes(commitHash)) {
 			commit.classList.add('refined-github-merge-commit');
 		}
 	}
--- a/source/features/mark-merge-commits-in-list.tsx
+++ b/source/features/mark-merge-commits-in-list.tsx
@@ -5,7 +5,7 @@ import * as api from '../libs/api';
 import features from '../libs/features';
 import {getRepoGQL} from '../libs/utils';
 
-const getMergeCommits = async (commits: string[]): Promise<string[]> => {
+const filterMergeCommits = async (commits: string[]): Promise<string[]> => {
 	const {repository} = await api.v4(`
 		repository(${getRepoGQL()}) {
 			${commits.map((commit: string) => `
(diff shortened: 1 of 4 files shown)


source/features/mark-merge-commits-in-list.tsx
```diff
@@ -0,0 +1,56 @@
...
+							parents{
+								totalCount
+							}
+						}
+					}
+			`).join('\n')}
+		}
+	`);
+
+	return repository;
```
fregante: Parse the API response and just return the data you want: a list of commit hashes that are merge commits. This way you won’t have to type CommitInfo. Also the function name would become `filterMergeCommits`

source/features/mark-merge-commits-in-list.tsx
```diff
@@ -0,0 +1,56 @@
...
+}
+
+async function init(): Promise<void | false> {
+	const pageCommits = select.all('li.commit');
+	const parentCommitCount = await getCommitParentCount(pageCommits.map(getCommitHash));
+	for (const commit of pageCommits) {
+		const commitHash = getCommitHash(commit);
+		const {totalCount} = parentCommitCount[api.escapeKey(commitHash)].parents;
+		if (totalCount === 2) {
+			commit.classList.add('refined-github-merge-commit');
```
fregante: Our classes start with `rgh-`

yakov116: @fregante np I just revered the removal commit and started from there. The classes are exactly how they were before.

yakov116:
@fregante this is what you meant?

source/features/mark-merge-commits-in-list.tsx
```diff
@@ -0,0 +1,63 @@
...
+import features from '../libs/features';
+import {getRepoGQL} from '../libs/utils';
+
+const getMergeCommits = async (commits: string[]): Promise<string[]> => {
+	const {repository} = await api.v4(`
+		repository(${getRepoGQL()}) {
+			${commits.map((commit: string) => `
+				${api.escapeKey(commit)}: object(expression: "${commit}") {
+					... on Commit {
+							parents{
```
fregante: Too much indentation here. Also there’s no space between parents and the {

source/features/mark-merge-commits-in-list.tsx
```diff
@@ -0,0 +1,63 @@
...
+								totalCount
+							}
+						}
+					}
+			`).join('\n')}
+		}
+	`);
+
+	return filterCommits(repository);
```
fregante: You can inline this function in here. `getMergeCommits` should become `filterMergeCommits`

source/features/mark-merge-commits-in-list.tsx
```diff
@@ -0,0 +1,63 @@
...
+};
+
+function filterCommits(commits: object): string[] {
+	const mergeCommits = [];
+	for (const commit of Object.entries(commits)) {
```
fregante: Destructure, so you don’t have to use `commit[1]` later 
```suggestion
	for (const [key, commit] of Object.entries(commits)) {
```

source/features/mark-merge-commits-in-list.tsx
```diff
@@ -0,0 +1,63 @@
...
+		if (commit[1].parents.totalCount === 2) {
+			mergeCommits.push(commit[0]);
```
fregante: The current behavior is “given a list of commit hashes, return a list of commit hashes *as escaped by the API.*” The second part feels out of place.

Functions should try to be self-contained. Here it’s returning the escaped API key instead of a regular commit. Use this to unescape it

```suggestion
			mergeCommits.push(commit.slice(1));
```

As mentioned earlier, `filterMergeCommits` receives a list of commit hashes and returns a list of merge commit hashes. Such function would make the code more understandable.

source/features/mark-merge-commits-in-list.tsx
fregante: Please follow my reviews fully, especially when I provide exact code suggestions. I used `key` because the value will be the API's escaped key, not the commit. It becomes `commit` only after you apply `.slice(1)`

fregante:
Can you always include a screenshot in the first comment when un-drafting a PR? This way one doesn't have to check out the PR to make sure it looks right

yakov116:
> Can you always include a screenshot in the first comment when un-drafting a PR? This way one doesn't have to check out the PR to make sure it looks right

Sure

source/features/mark-merge-commits-in-list.tsx
yakov116: No problem.

Just go easy on me. I am working hard and I am also fighting with typescript too.

yakov116: Is this the only issue?

yakov116: @fregante 
the return value is 
```javascript
{_0341d299c80dbfc8f4aabbb4d36829db8cf7f2a9:
parents:
totalCount: 1}
```
so key would be the commit and the second value is the parents. Can You give me a drop more detail where i'm going wrong?

fregante:
It looks good. I suggest dimming the buttons on the right too and simplify the selector

source/features/mark-merge-commits-in-list.tsx
fregante: Copy-paste my code suggestion, that’s all that went wrong

yakov116:
@fregante the old style was like this
image
Do you want to follow that or the new style is fine?

fregante:
That wasn't the style that was approved in that PR: https://github.com/sindresorhus/refined-github/pull/1121#issuecomment-370018467

The current format is fine, but I'd dim the whole line, including the buttons.

yakov116:
image
Is this good?

fregante:
Correct. The highlight is still there so it makes sense to show the whole thing. `dim-bots` works similarly

PR #2835, opened 2020-02-26
Don't show unnecessary `default-branch` after AJAX navigation

### LINKED ISSUE #2799: `default-branch-button` appears on the default branch
Replaces #2728

1. Visit https://github.com/sindresorhus/refined-github/commit/bf5dce817b31cea8c4d1535759adc81cc194d305
1. Refresh the page
2. Click the "Code" tab

This unnecessary "default branch button" might appear:

unnecessary

arnaudvalle:
On ajaxed pages, an extra <link> tag is added so we need to get the last one to figure out what the current branch is.

Closes #2799 

1. Visit https://github.com/sindresorhus/refined-github/commit/bf5dce817b31cea8c4d1535759adc81cc194d305
2. Refresh the page
3. Click the "Code" tab
4. Note that there is NO 'left arrow' button next to the branch dropdown, unlike here:
image
5. Repeat steps 1 to 4 with a single file url, i.e. https://github.com/sindresorhus/refined-github/blob/bf5dce817b31cea8c4d1535759adc81cc194d305/source/features/profile-gists-link.tsx

<!-- 

Thanks for contributing! 🍄

1. LINKED ISSUES:

2. TEST URLS:
   Add some test URLs

3. SCREENSHOT:

### DIFF
--- a/source/libs/utils.ts
+++ b/source/libs/utils.ts
@@ -52,7 +52,7 @@ export const replaceBranch = (currentBranch: string, newBranch: string): string 
 };
 
 export const getCurrentBranch = (): string => {
-	return select<HTMLLinkElement>('link[rel="alternate"]')!
+	return select.last<HTMLLinkElement>('link[rel="alternate"]')!
 		.href
 		.split('/')
 		.slice(6)


fregante:
That’s weird, the default branch is cached, this code should not be called there anyway. Can you double-check this by adding a console.log in this function? Last time I checked it was correctly cached

fregante:
My bad, I didn't realize this function gets the current branch, _not_ the default branch.

cc @lukehefson this is happening because the commit page's `link[rel="alternate"]` isn't being removed by GitHub after the ajaxed navigation. Probably because it doesn't have a `data-pjax-transient` attribute.

arnaudvalle:
> My bad, I didn't realize this function gets the current branch, _not_ the default branch.

Sorry about that, I think the wording of my commit / PR was a bit misleading as well... Thanks for merging it so quickly ;)

Issue #2770, opened 2020-02-10
Sidebar is not sticky


tooomm:
![no-widget](https://user-images.githubusercontent.com/9874850/74180045-0512f300-4c3f-11ea-8b1e-c36c8127b241.png)

Zooming out makes it appear (Zoom in as well!) - then it sticks, also when returning to 100% zoom...

My laptops resolution is 1366x768
Firefox 72.0.2 on Windows 10
Addon version 20.2.3

fregante:
Can you take a screenshot of the whole sidebar at 100%?

When the sidebar refreshes the listener forgets about it (because it’s a new element) so we need to use or helper (I think called onUpdatableContent…) to make sure it’s listened to again.

tooomm:
Which screenshot do you need me to take exactly? Can you explain a bit further, please?
Same as above, but after zooming to make the sidebar appear?
Or the normal view at the very top?

fregante:
1. Zoom at 100%
2. Refresh the page
3. Scroll just enough that the whole sidebar is visible

tooomm:
>Can you take a screenshot of the whole sidebar at 100%?

![widget2](https://user-images.githubusercontent.com/9874850/74248168-75754f00-4ce7-11ea-864c-c769a557ea6d.png)

<br>

>And also, is the sidebar sticky after doing this?

Making the sidebar appear via zooming in or out work, going back to 100% after that keeps it. Refreshing the page then makes the sidebar disappear.

<br>

I also realized that moving out of fullscreen makes the sidebar sticky again. Going back to keep it. Refreshing the page in fullscreen removes it again.

fregante:
Avoid using the zoom/fullscreen to test it, it's irrelevant: the stickiness is determined on `resize`, so just resize the window.

Anyway, it's not working for me at all now

tooomm:
Resizing has the same result.
Moving down leads to the sidebar disappearing. Making the window narrower suddenly makes it appear. Going back to normal width keeps it until hitting refresh.

How does the sidebar update in between?
It **never** works for me

fregante:
Please record a screencast because I have no idea

fregante:
Also try in chrome and see if there are any errors in the console

tooomm:
>Please record a screencast because I have no idea

https://screencast-o-matic.com/watch/cYn2oPwyaf

fregante:
[...]

PR #2853, opened 2020-03-03
Show `00:mm` instead of `24:mm` in `user-local-time`

hugovk:
# Before

When hovering over usernames, sometimes it showed a local time in the hour after midnight:

image

That looks strange to me, normally I see those times beginning "00:xx", not "24:xx". I guess I'm not used the US locale.

# After

Use the browser's default locale, with fallback to `en-US`:

image

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DateTimeFormat#Examples

### DIFF
--- a/source/features/user-local-time.tsx
+++ b/source/features/user-local-time.tsx
@@ -14,7 +14,7 @@ interface Commit {
 	sha: string;
 }
 
-const timeFormatter = new Intl.DateTimeFormat(['default', 'en-US'], {
+const timeFormatter = new Intl.DateTimeFormat(['en-GB'], {
 	hour: 'numeric',
 	minute: 'numeric',
 	hour12: false
--- a/source/features/user-local-time.tsx
+++ b/source/features/user-local-time.tsx
@@ -14,7 +14,7 @@ interface Commit {
 	sha: string;
 }
 
-const timeFormatter = new Intl.DateTimeFormat('en-US', {
+const timeFormatter = new Intl.DateTimeFormat('default', {
 	hour: 'numeric',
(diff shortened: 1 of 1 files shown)


fregante:
That super weird, I’ve never seen `24:10` either. Perhaps the locale should be changed to one that makes more sense. Does en-UK format the time correctly? 

Does GitHub show times like this anywhere on the site? We should copy their format

hugovk:
Yes, `en-UK` shows `00:10` format.

Here's a GitHub [popup](https://github.com/sindresorhus/refined-github/issues/2770#issuecomment-586762838) (with no Refined GitHub): 

image

fregante:
> Yes, en-UK shows 00:10 format.

I’d use that then. 

> Here's a GitHub popup (with no Refined GitHub):

Good! We’re safe as long as we match the styles and formats that GitHub uses.

hugovk:
Updated!

source/features/user-local-time.tsx
```diff
@@ -14,7 +14,7 @@ interface Commit {
 	sha: string;
 }
 
-const timeFormatter = new Intl.DateTimeFormat('en-US', {
+const timeFormatter = new Intl.DateTimeFormat(['en-GB'], {
```
fregante: No array necessary

hugovk: Thanks, updated!

fregante:
Thank you for the PR!

hugovk:
You're welcome, and thank you for Refined GitHub, it makes things so much easier!

PR #2858, opened 2020-03-04
Add `pr-commit-lines-changed` feature

### LINKED ISSUE #2326: Show lines changed for current commit
When I'm on a specific commit of a pull request, I'd like to see the number of lines changed in this commit, only.

Currently, the total number of lines changed in the whole pull request is shown: Screenshot from 2019-08-13 09-52-59

Example URL: [sindresorhus/refined-github#2182/commits/7e8bb264c](https://github.com/sindresorhus/refined-github/pull/2182/commits/7e8bb264cda190a5ce2e09e51196cb8f4b230696)

This number of lines changed could also appear on [the commit page](https://github.com/sindresorhus/refined-github/commit/7e8bb264cda190a5ce2e09e51196cb8f4b230696) (outside of the pull request).

yakov116:
closes #2326 

image


source/features/lines-changed-per-commit.tsx
```diff
@@ -0,0 +1,53 @@
...
+			</span>);
+	}
+}
+
+features.add({
+	id: __featureName__,
+	description: 'Add line changes per commit',
+	screenshot: 'https://user-images.githubusercontent.com/16872793/75835025-cb3f9300-5d8b-11ea-8126-d8778c71e467.png',
+	include: [
+		features.isCommitList
```
fregante: This should be `isCommit` and `isPRCommit`. This information isn’t useful on lists

source/features/lines-changed-per-commit.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+function getCommitHash(commit: HTMLElement): string {
+	return (commit.dataset.channel as string).split(':')[3];
+}
+
+async function init(): Promise<void | false> {
+	const pageCommits = select.all('li.commit:not(.rgh-merge-commit)');
+	const commitChanges = await getCommitChanges(pageCommits.map(getCommitHash));
+	for (const commit of pageCommits) {
+		// @ts-ignore
```
fregante: No need to add this, the build completes even if there are errors. They will likely be fixed before the PR is merged

source/features/lines-changed-per-commit.tsx
```diff
@@ -0,0 +1,53 @@
...
+				... on Commit {
+                    additions
+                    deletions
+					}
+				}
+			`).join('\n')}
+		}
+	`);
+
+	return repository;
```
fregante: Again this function should return a raw API response, I repeatedly explained this in previous PRs.

yakov116: @fregante explain what you mean? I need 2 items from the response.  The commit and the additions / deletions. What do you expect the response to be?

fregante: https://github.com/sindresorhus/refined-github/pull/2832#discussion_r384950532

fregante: It returns an object like: 

```js
{
  _1a2b3c: {
    additions: 1,
    deletions: 1
  }
}
``` 
but should be

```js
{
  1a2b3c: {
    additions: 1,
    deletions: 1
  }
}
``` 

but yeah I guess in this case the difference isn’t much to justify the extra code. Plus what changes required for this PR skips the problem altogether

source/features/lines-changed-per-commit.tsx
yakov116: is commit has the correct _info_ 
image

source/features/lines-changed-per-commit.tsx
```diff
@@ -0,0 +1,52 @@
...
+		}
+    `);
+	// @ts-ignore
+	const {additions, deletions} = repository.object;
+	return [additions, deletions];
+};
+
+async function init(): Promise<void | false> {
+	const pageCommit = (await elementReady('.sha.user-select-contain'))!;
+	const commitSha = String(pageCommit.textContent);
```
fregante: `String()` isn’t necessary here.  The type of `textContent` is already `string | null`. If you want to avoid the error, just append `!` and it will ignore the `null` part.

source/features/lines-changed-per-commit.tsx
```diff
@@ -0,0 +1,52 @@
...
+                }
+		}
+    `);
+	// @ts-ignore
+	const {additions, deletions} = repository.object;
+	return [additions, deletions];
+};
+
+async function init(): Promise<void | false> {
+	const pageCommit = (await elementReady('.sha.user-select-contain'))!;
```
fregante:
```suggestion
	const commitSha = (await elementReady('.sha.user-select-contain'))!.textContent!;`
```

You can merge these 2 lines.

source/features/lines-changed-per-commit.tsx
```diff
@@ -0,0 +1,52 @@
...
+	const commitSha = String(pageCommit.textContent);
+	const [additions, deletions] = await getCommitChanges(commitSha);
```
fregante: API calls should be made as early as possible. In this case this can be the first line of the init function

source/features/lines-changed-per-commit.tsx
```diff
@@ -0,0 +1,52 @@
...
+			<span className="diffstat-block-neutral"/>
+			<span className="diffstat-block-neutral"/>
+			<span className="diffstat-block-neutral"/>
+
+		</span>);
+}
+
+features.add({
+	id: __featureName__,
+	description: 'Add line changes on pr commits',
```
fregante: Less sloppy

source/features/lines-changed-per-commit.tsx
```diff
@@ -0,0 +1,52 @@
...
+                    }
+                }
+		}
+    `);
+	// @ts-ignore
+	const {additions, deletions} = repository.object;
+	return [additions, deletions];
+};
+
+async function init(): Promise<void | false> {
```
fregante: This function never returns false

source/features/lines-changed-per-commit.tsx
```diff
@@ -0,0 +1,52 @@
...
+                ... on Commit {
+                    additions
+                    deletions
+                    }
+                }
+		}
+    `);
+	// @ts-ignore
+	const {additions, deletions} = repository.object;
+	return [additions, deletions];
```
fregante: Maybe this shortens it a bit
```suggestion
	return [repository.object.additions, repository.object.deletions];
```

source/features/lines-changed-per-commit.tsx
```diff
@@ -0,0 +1,52 @@
...
+const getCommitChanges = async (commit: string): Promise<number[]> => {
+	const {repository} = await api.v4(`
+		repository(${getRepoGQL()}) {
+            object(expression: "${commit}") {
+                ... on Commit {
+                    additions
+                    deletions
+                    }
+                }
+		}
```
fregante: Indentation. Don’t wait for my review to apply these changes, make sure your PR is “finalized” before asking for a review, I’m not your autocorrect

source/features/lines-changed-per-commit.tsx
```diff
@@ -0,0 +1,52 @@
+import select from 'select-dom';
+import * as api from '../libs/api';
+import features from '../libs/features';
+import {getRepoGQL} from '../libs/utils';
+import React from 'dom-chef';
+import elementReady from 'element-ready';
+
+const getCommitChanges = async (commit: string): Promise<number[]> => {
```
fregante: Can you add `cache.function`?

Also this type would be more correct because there are only 2 values: 
```suggestion
const getCommitChanges = async (commit: string): Promise<[number, number]> => {
```

yakov116:
I don't like the implementation I did. Since it removes the diffstat entirely. (Unless you know how to calculate it)

Maybe we add this.
image
But without the expandable link to the files changed.

source/features/lines-changed-per-commit.tsx
```diff
@@ -0,0 +1,52 @@
...
+		repository(${getRepoGQL()}) {
+            object(expression: "${commit}") {
+                ... on Commit {
+                    additions
+                    deletions
+                    }
+                }
+		}
+    `);
+	// @ts-ignore
```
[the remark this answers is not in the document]
yakov116: Sorry when doing now watch typescript has to pass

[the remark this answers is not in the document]
fregante: > Sorry when doing npm watch typescript has to pass

No it doesn't, the typecheck specifically happens in parallel and completes after webpack outputs the files.

fregante:
> Maybe we add this.

Where would you add it? I prefer the current way instead of making the content jump down because of the extra line

yakov116:
> > Maybe we add this.
> 
> Where would you add it? I prefer the current way instead of making the content jump down because of the extra line

Maybe next to the branch info? If your ok with the way it is currently then i'm fime too.

source/features/lines-changed-per-commit.tsx
yakov116: The way it is now is good?

fregante: > In this case this can be the first line of the init function

source/content.ts
```diff
@@ -162,6 +162,7 @@ import './features/quick-mention';
 import './features/extend-discussion-status-filters';
 import './features/expand-all-hidden-comments';
 import './features/bugs-tab';
+import './features/lines-changed-per-commit';
```
fregante:
```suggestion
import './features/pr-commit-lines-changed';
```

source/features/lines-changed-per-commit.tsx
```diff
@@ -0,0 +1,53 @@
...
+	const {repository} = await api.v4(`
+		repository(${getRepoGQL()}) {
+            object(expression: "${commit}") {
+                ... on Commit {
+                    additions
+                    deletions
+                    }
+                }
+		}
+    `);
```
fregante: Whitespace...

fregante:
After this you can add the mention in the readme

source/features/lines-changed-per-commit.tsx
yakov116: Sorry saw it and forgot to fix that.

yakov116:
Should we detect if the number of additions and subtraction are the same then leave it alone? Or it is not necessary?

yakov116:
What is the screenshot of?

fregante:
> What is the screenshot of?

What you have in the first comment looks good

source/features/pr-commit-lines-changed.tsx
```diff
@@ -43,6 +43,7 @@ async function init(): Promise<void> {
 features.add({
 	id: __featureName__,
 	description: 'Add line changes on PR commits.',
```
fregante: I wonder if this should be
```suggestion
	description: 'Adds diff stats on PR commits.',
```

yakov116: Agree

yakov116:
@fregante anything else I need to do?

fregante:
Take car. Go to mum's. Kill Phil, grab Liz, go to the Winchester, have a nice cold pint, and wait for all of this to blow over.

yakov116:
> Take car. Go to mum's. Kill Phil, grab Liz, go to the Winchester, have a nice cold pint, and wait for all of this to blow over.
> 

Done! :laughing:

PR #2860, opened 2020-03-05
Drop `friendlier-requested-changes` feature

fregante:
Reverts sindresorhus/refined-github#2611

I don't think this is an improvement and many others in that PR also think that.

There's nothing wrong with red, these colors match traffic lights and are well understood everywhere.


sindresorhus:
I'm fine with this. I don't feel strongly either way.

PR #2864, opened 2020-03-06
Fix `dim-bots` styles

kidonng:
I have this strange feeling when I open GitHub today... it seems they have changed their mockup somehow. When I saw commit history page, I knew it was true.

**Chrome**
image

**Firefox**
image

This PR kindly stops the performance of these gorgeous buttons and fix `dim-bots` up.

## Test URLs

- https://github.com/renovatebot/renovate/commits/master
- https://github.com/kidonng/readhub/commits/master

### DIFF
--- a/source/features/dim-bots.css
+++ b/source/features/dim-bots.css
@@ -8,8 +8,9 @@ Transitions are used to delay the "hover" status so it's not too annoying when m
 */
 .rgh-dim-bot.commit .commit-title,
 .rgh-dim-bot.commit .commit-meta,
-.rgh-dim-bot.commit .commit-links-cell summary,
-.rgh-dim-bot.commit .commit-links-cell .btn {
+.rgh-dim-bot.commit summary,
+.rgh-dim-bot.commit .commit-links-group,
+.rgh-dim-bot.commit .commit-links-group + .btn {
 	transition: 0.1s 0.3s;
 	transition-property: opacity, margin-bottom;
 }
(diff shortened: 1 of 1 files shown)


source/features/dim-bots.css
```diff
@@ -8,8 +8,9 @@ Transitions are used to delay the "hover" status so it's not too annoying when m
 */
 .rgh-dim-bot.commit .commit-title,
 .rgh-dim-bot.commit .commit-meta,
-.rgh-dim-bot.commit .commit-links-cell summary,
-.rgh-dim-bot.commit .commit-links-cell .btn {
+.rgh-dim-bot.commit summary,
```
kidonng: If `summary` seems too broad, we can use `.signed-commit-badge` or something alike instead.

source/features/dim-bots.css
```diff
@@ -8,8 +8,9 @@ Transitions are used to delay the "hover" status so it's not too annoying when m
...
+.rgh-dim-bot.commit .commit-links-group,
+.rgh-dim-bot.commit .commit-links-group + .btn {
```
kidonng: I admit `+ .btn` looks awful, but I have no better idea ¯\_(ツ)_/¯

fregante:
The selectors seem fine as long as they work as expected. Does the result still match this gif?

kidonng:
Dimmed commits appear to be a little shorter than in your gif... that's fine I think?

Here's my gif (in Chrome)

xQTxBbDGsE

fregante:
> Dimmed commits appear to be a little shorter than in your gif... that's fine I think?

I think they just reduced the padding, it looks fine/better

PR #2866, opened 2020-03-06
Fix `releases-tab` position on repo with "settings" in their name

### LINKED ISSUE #2763: Wrong `releases-tab`  position if repo name contains "settings"
image

# Test
* https://github.com/atom-community/sync-settings

# Fix
https://github.com/sindresorhus/refined-github/blob/master/source/features/releases-tab.tsx#L61 should be 
```diff
-appendBefore('.reponav', '.reponav-dropdown, [href$="settings"]', releasesTab);
+appendBefore('.reponav', '.reponav-dropdown, [href$="/settings"]', releasesTab);
```
Same as https://github.com/sindresorhus/refined-github/blob/master/source/features/releases-tab.tsx#L17

arnaudvalle:
Closes #2763 

- Go to https://github.com/sindresorhus/refined-github/
- Go to https://github.com/atom-community/sync-settings
- Go to https://github.com/aritchie/settings
--> The `releases` tab should be right before the `More` tab

Screenshot 2020-03-06 at 15 27 58

<!-- 

Thanks for contributing! 🍄

1. LINKED ISSUES:

2. TEST URLS:
   Add some test URLs

3. SCREENSHOT:

### DIFF
--- a/source/libs/dom-utils.ts
+++ b/source/libs/dom-utils.ts
@@ -28,10 +28,8 @@ export const appendBefore = (parent: string|Element, before: string, child: Elem
 		parent = select(parent)!;
 	}
 
-	const scopedBefore = getScopedSelector(before);
-
 	// Select direct children only
-	const beforeElement = select(scopedBefore, parent);
+	const beforeElement = select(getScopedSelector(before), parent);
 	if (beforeElement) {
 		beforeElement.before(child);
 	} else {
--- a/source/libs/dom-utils.ts
+++ b/source/libs/dom-utils.ts
@@ -1,5 +1,5 @@
 import select from 'select-dom';
-import { getScopedSelector } from './utils';
+import {getScopedSelector} from './utils';
 
 /**
  * Append to an element, but before a element that might not exist.
--- a/source/libs/dom-utils.ts
+++ b/source/libs/dom-utils.ts
@@ -1,4 +1,5 @@
 import select from 'select-dom';
+import { getScopedSelector } from './utils';
 
 /**
  * Append to an element, but before a element that might not exist.
@@ -27,8 +28,10 @@ export const appendBefore = (parent: string|Element, before: string, child: Elem
 		parent = select(parent)!;
 	}
 
+	const scopedBefore = getScopedSelector(before);
+
 	// Select direct children only
-	const beforeElement = select(`:scope > ${before}`, parent);
+	const beforeElement = select(scopedBefore, parent);
 	if (beforeElement) {
 		beforeElement.before(child);
 	} else {
--- a/source/libs/utils.ts
+++ b/source/libs/utils.ts
@@ -149,3 +149,11 @@ export function reportBug(featureName: string, bugName: string): void {
 	console.log('Find existing issues:\n' + String(issuesUrl));
 	console.log('Open new issue:\n' + String(newIssueUrl));
 }
+
+/**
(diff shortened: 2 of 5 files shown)


fregante:
How come `more-dropdown` isn't affected? It uses the same selector

https://github.com/sindresorhus/refined-github/blob/5d15c14f36dd20ee72ba3fe933b8d0756e8f9892/source/features/more-dropdown.tsx#L18

arnaudvalle:
> How come `more-dropdown` isn't affected? It uses the same selector

Could it be because there's extra async stuff going on for the releases tab (to get the count for ex) which delays it slightly more than in the other case?

fregante:
But `.reponav [href$="settings"]` matches the same Code tab regardless of timing. They have the same code (including the same elementReady):

https://github.com/sindresorhus/refined-github/blob/5d15c14f36dd20ee72ba3fe933b8d0756e8f9892/source/features/releases-tab.tsx#L61-L62

fregante:
Yep, found the bug. It's worth understanding _why_ something doesn't work as expected.

In `appendBefore` we use `:scope > ` to ensure that the item is a direct child of `.reponav`. The `.reponav-dropdown, [href$="settings"]` selector is simply appended to `:scope > `, making it:

```css
:scope > .reponav-dropdown, [href$="settings"]
```

Or written more clearly:


```css
:scope > .reponav-dropdown,
[href$="settings"]
```

The selector no longer ensures that `[href$="settings"]` is a direct descendant of `:scope`.

---

The fix should **only** be inside `dom-utils`, something like 

```js
before.split(',').map(selector => `:scope > ${before}`).join(',')
```

arnaudvalle:
Ah nice one, thanks for looking into it - I'm still not familiar enough with the insides of the project. I'll have a look at a fix as soon as I can.

arnaudvalle:
> (because this breaks selectors like `[title=","]`)

Regarding this, and without trying to reinvent the wheel, should we be using some external package to handle those edge cases? A quick search and I found https://github.com/joakimbeng/split-css-selector or https://github.com/goto-bus-stop/split-selector that would do the job for example. Let me know if you have a preference or if you'd prefer to do a home-made solution (or do nothing since that edge isnt happening in the app yet)

source/libs/dom-utils.ts
```diff
@@ -27,8 +28,10 @@ export const appendBefore = (parent: string|Element, before: string, child: Elem
 		parent = select(parent)!;
 	}
 
+	const scopedBefore = getScopedSelector(before);
+
 	// Select direct children only
-	const beforeElement = select(`:scope > ${before}`, parent);
+	const beforeElement = select(scopedBefore, parent);
```
fregante: You inline the call here, no need for an extra constant

source/libs/utils.ts
```diff
@@ -149,3 +149,11 @@ export function reportBug(featureName: string, bugName: string): void {
...
 	console.log('Open new issue:\n' + String(newIssueUrl));
 }
+
+/**
+ * Prepend `:scope >` to a single or group of css selectors.
+ * @param {string} selector A css selector.
+ */
+export const getScopedSelector = (selector: string): string => {
+	return selector.split(',').map(sub => `:scope > ${sub.trim()}`).join(',');
+};
```
fregante:
```suggestion
export function getScopedSelector(selector: string): string {
	return selector.split().map(sub => `:scope > ${sub.trim()}`).join();
}
```

PR #2877, opened 2020-03-07
Fix position of `user-profile-follower-badge`

### LINKED ISSUE #2874: `user-profile-follower-badge` is in wrong position
1. Go to your profile
1. Open Followers tab
1. Open first follower
4. Notice wrong position of badge

I'd also rename this feature to `user-follows-you`

<table>
<tr><th>Current (above avatar)<th> Expected (below username)
<tr><td><td>

kidonng:
Fix #2874

Cause for this: `.float-left` name card
image

~~Also remove the margin since it looks pretty fine.~~ I'm using custom styles in the following image, please ignore it

image

### DIFF
--- a/source/features/user-profile-follower-badge.css
+++ b/source/features/user-profile-follower-badge.css
@@ -4,6 +4,7 @@
 	animation: fade-in 0.2s;
 	margin-top: -15px;
 	margin-bottom: 5px;
+	clear: both;
 }
 
 .user-profile-bio {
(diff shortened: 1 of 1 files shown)


fregante:
I think the margin was added to avoid causing a jump. I don’t remember. Alternatively this can be done via position absolute (without adding top/left/bottom/right properties)

kidonng:
The margin doesn't help, I still see a jump, which doesn't seem to matter, we even have an animation for the badge.

PR #2881, opened 2020-03-11
Add `deep-blame` feature

### LINKED ISSUE #2876: "Deep blame" for squashed commits
1. Visit https://github.com/sindresorhus/refined-github/blame/0ca864af6da9cac9f503e4abd0c58b60fbe1abc6/source/features/link-to-file-in-file-history.tsx#L29

There's no blame specifically for that line, there's only a big blame block. However if you follow the PR's link, (5 clicks and 4 page loads later) you'll end up on this page:

https://github.com/sindresorhus/refined-github/blame/a2f6e93ca789d2b0677e152bc28d35d4a4180175/source/features/link-to-file-in-file-history.tsx#L34

---

This feature would need I think:

1. On blame page, add "deep blame this commit" button on each commit that has a PR link in its title
2. When the button is clicked, it has to query the API to get the PR's latest commit (user will be shown a loading icon)
3. Redirect to that commit

Step 2 should (if possible) verify that the linked PR is effectively the one that was squashed, and not just a randomly-linked PR.

yakov116:
Closes #2876

image

# Test

Related PR in commit title: it should "enter" it https://github.com/sindresorhus/refined-github/blame/d1e27cb34f18b9d5d57fdbf853e5b15c3a660d85/source/features/deep-reblame.tsx#L67

Unrelated PR in commit title: it should `alert()` https://github.com/sindresorhus/refined-github/blame/d1e27cb34f18b9d5d57fdbf853e5b15c3a660d85/source/features/deep-reblame.tsx#L67


fregante:
> I need to know how I should link it.

I think

- add `tooltipped` class to tell the user they can alt-click
- add alt-click handler

yakov116:
What if the is no link there to go back? (There is no icon since there is no direct change for that line)
Just make a regular click?

fregante:
> What if the is no link there to go back?

I think we have to add the icon and yes, make it a regular click

source/features/deep-blame.tsx
```diff
@@ -0,0 +1,104 @@
...
+		const currentParentElement = pullRequest.closest('.blame-hunk');
+		const versionsParent = select('.blob-reblame', currentParentElement!);
+		const currentLineNumber = select('.js-line-number', currentParentElement!)!.textContent!;
+
+		const href = new URL(window.location.href.replace(String(getReference()), prBlameCommit));
+		href.hash = 'L' + currentLineNumber;
+
+		if (select.exists('a', versionsParent!)) {
+			const versionsElement = select('a', versionsParent!);
+			versionsElement!.setAttribute('aria-label', 'View blame prior to this change. Press Alt to see `something....`');
```
yakov116: @fregante I need help with the wording

source/features/deep-blame.tsx
```diff
@@ -0,0 +1,104 @@
...
+				> {versionIcon()}
+				</a>
+			);
+		}
+	}
+}
+
+features.add({
+	id: __featureName__,
+	description: 'Hello',
```
yakov116: Help with the wording

source/features/deep-blame.tsx
```diff
@@ -0,0 +1,104 @@
...
+					location.href = String(href);
+				}
+			});
+		} else {
+			versionsParent!.append(
+				<a
+					href={String(href)}
+					aria-label="View blame prior to this change"
+					className="reblame-link link-hover-blue no-underline tooltipped tooltipped-e d-inline-block pr-1"
+				> {versionIcon()}
```
yakov116: I think we should make the icon purple. Since there is no way to tell that this is only a `deep-blame`

source/features/deep-blame.tsx
```diff
@@ -0,0 +1,106 @@
...
+	return (commit.nextElementSibling! as HTMLAnchorElement).href.split('/').slice(-1)[0];
+}
+
+async function init(): Promise<void | false> {
+	const pullRequests = select.all('[data-hovercard-type="pull_request"]');
+	if (pullRequests.length === 0) {
+		return;
+	}
+
+	const commits = await filterMergeCommits([...new Set(pullRequests.map(getCommitHash))]);
```
fregante: I thought we agreed to make this action only appear AFTER alt clicking, no API calls should be made before. It's in my first post,

> 2. When the button is clicked, it has to query the API to get the PR's latest commit (user will be shown a loading icon)

yakov116: Understand.

However most times the version icon is not there even that you can go back?! Where should the use click?

yakov116:
Messy code...
@fregante something like this

yakov116:
I will add the loading later

source/features/deep-blame.tsx
fregante: > Unless you mean don't run anything unless the user clicks a button and then load everything I wrote.

That’s exactly what I meant/wrote. 

My reasoning is that blame pages could have dozens of commits so the query could be very large. Also because this feature isn’t meant to be used that frequently, so even if the query is light, it would be useless.

fregante:
> @fregante something like this

How would that work?

source/features/deep-blame.tsx
fregante: You asked for confirmation here, I confirmed, and then you changed the whole thing 🤷‍♂️ 

https://github.com/sindresorhus/refined-github/issues/2876#issuecomment-596212371

yakov116: First we are only showing for pull request commits

Second 
>However most times the version icon is not there even that you can go back?! Where should the use click?

yakov116:
> > @fregante something like this
> 
> How would that work?

Click the button and it loads the commits

yakov116:
@fregante can you explain to me where the user is clicking??

yakov116:
For example
image


Where is the user clicking alt??? If there is a link there I can add an alt, if not then.....

fregante:
> 1. On blame page, add "deep blame this commit" button on each commit that has a PR link in its title

If there’s already a “prior blame” button, use alt-click on it
If there’s no button, add it

yakov116:
@fregante I should add a button even if there is not linked pull request? (so the link will do nothing)??

yakov116:
That is why I added a button and then it will load all and add the links. 
>My reasoning is that blame pages could have dozens of commits so the query could be very large. Also because this feature isn’t meant to be used that frequently, so even if the query is light, it would be useless.

That is why I added a button so only then will it load.

Sorry for getting so frustrated, I would love to see my hard work go through and I am not understanding what is wrong.

fregante:
> That is why I added a button so only then will it load.

But in the user’s eyes that button does nothing and maybe it really does nothing (if no PR commits exist on the page). 

> @fregante I should add a button even if there is not linked pull request? (so the link will do nothing)??

We only add the button if a PR number is in the title, so we have some degree of certainty that the commit has an associated PR. If that doesn’t work, we use an `alert` saying “No PRs associated with this commit” or something like that

yakov116:
Understand. I am changing the implementation anyways but that button would have not shown if there were no pr's

yakov116:
@fregante can you give me some feedback? I tried my best (I promise).

I should cache.function the commit right?

source/features/deep-blame.tsx
```diff
@@ -0,0 +1,111 @@
...
+	if (event.delegateTarget.href && event.altKey) {
+		event.preventDefault();
+	}
+
+	event.delegateTarget.classList.add('anim-pulse');
+	event.delegateTarget.replaceChild(octofaceIcon(), event.delegateTarget.firstElementChild!);
+
+	const prBlameCommit = await getPullRequestBlameCommit(String(pullRequestCommit));
+	if (!prBlameCommit) {
+		event.delegateTarget.removeChild(event.delegateTarget.firstElementChild!);
```
fregante: Use .remove and .replaceWith, never “*Child” APIs. 

Also there should be an `alert` in here

source/features/deep-blame.tsx
```diff
@@ -0,0 +1,111 @@
...
+	return undefined;
+};
+
+function getCommitHash(commit: HTMLElement): string {
+	return (commit.nextElementSibling! as HTMLAnchorElement).href.split('/').slice(-1)[0];
+}
+
+async function redirectToBlameCommit(event: DelegateEvent<MouseEvent, HTMLLinkElement>): Promise<void | false> {
+	const {lineNumber, commit: pullRequestCommit} = event.delegateTarget.dataset;
+	if (event.delegateTarget.href && event.altKey) {
```
fregante: `event.delegateTarget` is used a lot, it’s better to create a (`blameLink`) variable for it

source/features/deep-blame.tsx
```diff
@@ -0,0 +1,111 @@
...
+	event.delegateTarget.classList.add('anim-pulse');
+	event.delegateTarget.replaceChild(octofaceIcon(), event.delegateTarget.firstElementChild!);
+
+	const prBlameCommit = await getPullRequestBlameCommit(String(pullRequestCommit));
+	if (!prBlameCommit) {
+		event.delegateTarget.removeChild(event.delegateTarget.firstElementChild!);
+		return;
+	}
+
+	const href = new URL(window.location.href.replace(String(getReference()), prBlameCommit));
```
fregante: You can drop `window` and `String`. Use `getReference()!`

source/features/deep-blame.tsx
```diff
@@ -0,0 +1,111 @@
...
+	href.hash = 'L' + String(lineNumber);
+	location.href = String(href);
```
fregante: Are other links loaded via AJAX? Can we trigger an AJAX navigation? You could create a util function called `navigateTo`

source/features/deep-blame.tsx
```diff
@@ -0,0 +1,111 @@
...
+}
+
+async function init(): Promise<void | false> {
+	const pullRequests = select.all('[data-hovercard-type="pull_request"]');
+	if (pullRequests.length === 0) {
+		return;
```
fregante: return false

source/features/deep-blame.tsx
```diff
@@ -0,0 +1,111 @@
+import './deep-blame.css';
+import select from 'select-dom';
+import * as api from '../libs/api';
+import React from 'dom-chef';
+import features from '../libs/features';
+import {getRepoGQL, getReference} from '../libs/utils';
+import delegate, {DelegateEvent} from 'delegate-it';
+import versionIcon from 'octicon/versions.svg';
+import octofaceIcon from 'octicon/octoface.svg';
```
fregante: Sort these so they look like a tree 😃 refer to the look in my other features. This makes it slightly easier to follow because they’re closer to being “aligned”

And always put the npm modules above the local `./` and `../` files

source/features/deep-blame.tsx
yakov116: Where do you have an example that I should follow.
This is the only place we navigate

source/features/deep-blame.tsx
yakov116: I knew you were going to comment on that LOL

source/features/deep-blame.tsx
yakov116: I am working on this now. It should be a window alert?

yakov116: @fregante I just realized how is this case even possible.

Are we supposed to check the linked PR number? Since The merge commit will always = a merge commit but it may not be the correct PR.

yakov116: Ohhh Now i remember.... Ouch when I rewrote the code I deleted the commit to check against LOL

yakov116:
Test on https://github.com/yakov116/TestR/blame/master/asf

source/features/deep-blame.tsx
fregante: Why did you let me comment then 🤷‍♂️

source/features/deep-blame.tsx
fregante: I tested and other links are not loaded via AJAX, so we're good here

source/features/deep-blame.tsx
yakov116: I forgot to change it. I was saying when I was writing the code....

source/features/deep-blame.tsx
fregante: Commit titles are custom:

```
Fix dropdown (replaces PR #1234)
```

Definitely an edge case. An alternative would be to just skip the check and just link to the PR; this would also simplify the code and wouldn't leave the user hanging... even if technically it's wrong... but that's what users get for mentioning unrelated PRs.

source/features/deep-blame.tsx
```diff
@@ -0,0 +1,140 @@
...
+			}
+		}
+	`);
+
+	const {nodes} = repository.object.associatedPullRequests;
+	const mergeCommit = nodes[0].mergeCommit.oid;
+	const pullRequestNumber = nodes[0].number;
+
+	if (pullRequestNumber === prNumber && mergeCommit === commit) {
+		return nodes[0].commits.nodes[0].commit.oid;
```
fregante:
```suggestion
	const {[associatedPullRequest]} = repository.object.associatedPullRequests;
	const mergeCommit = associatedPullRequest.mergeCommit.oid;
	const pullRequestNumber = associatedPullRequest.number;

	if (pullRequestNumber === prNumber && mergeCommit === commit) {
		return associatedPullRequest.commits.nodes[0].commit.oid;
```

source/features/deep-blame.tsx
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	return false;
+}, {
+	cacheKey: ([commit]) => __featureName__ + commit
+});
+
+function getCommitHash(commit: HTMLElement): string {
+	return (commit.nextElementSibling! as HTMLAnchorElement).href.split('/').slice(-1)[0];
```
fregante:
```suggestion
	return (commit.nextElementSibling as HTMLAnchorElement).href.split('/').pop();
```

Let's fix it in other features as well (in this PR, if you want)

source/features/deep-blame.tsx
```diff
@@ -0,0 +1,140 @@
...
+	if (pullRequestNumber === prNumber && mergeCommit === commit) {
+		return nodes[0].commits.nodes[0].commit.oid;
+	}
+
+	return false;
+}, {
+	cacheKey: ([commit]) => __featureName__ + commit
+});
+
+function getCommitHash(commit: HTMLElement): string {
```
fregante:
```suggestion
export function getCommitHash(commit: HTMLElement): string {
```

Lets reuse this function in other features. No need to move to `utils` because it's so specific

source/features/deep-blame.tsx
```diff
@@ -0,0 +1,140 @@
...
+	return (commit.nextElementSibling! as HTMLAnchorElement).href.split('/').slice(-1)[0];
+}
+
+const githubSpinner = (
+	<img
+		src="https://github.githubassets.com/images/spinners/octocat-spinner-128.gif"
+		alt="Octocat Spinner Icon"
```
fregante:
```suggestion
		alt="Loading"
```

or 
```suggestion
```

source/features/deep-blame.tsx
```diff
@@ -0,0 +1,140 @@
...
+		className="mr-2"
+		width="18"
+	/>
+);
+
+async function redirectToBlameCommit(event: DelegateEvent<MouseEvent, HTMLLinkElement>): Promise<void | false> {
```
fregante:
```suggestion
async function redirectToBlameCommit(event: DelegateEvent<MouseEvent, HTMLAnchorElement>): Promise<void> {
```

source/features/deep-blame.tsx
```diff
@@ -0,0 +1,140 @@
...
+	return false;
+}, {
+	cacheKey: ([commit]) => __featureName__ + commit
+});
+
+function getCommitHash(commit: HTMLElement): string {
+	return (commit.nextElementSibling! as HTMLAnchorElement).href.split('/').slice(-1)[0];
+}
+
+const githubSpinner = (
```
fregante: This should not be a global, or else this element is generated at every pageload, and the image will be loaded too

source/features/deep-blame.tsx
```diff
@@ -0,0 +1,140 @@
...
+			blameLink.firstElementChild!.replaceWith(versionIcon());
+		} else {
+			blameLink.firstElementChild!.remove();
+		}
+
+		alert('Unable to find linked Pull Request');
+		return;
+	}
+
+	const href = new URL(location.href.replace(getReference()!, prBlameCommit));
```
fregante: Perhaps this is more straightforward
```suggestion
	const href = new URL(`/${getRepoURL()}/blame/${prBlameCommit}`, location.origin);
```

source/features/deep-blame.tsx
yakov116: I am not sure. I am so lost. LOL
I think that leading the user to the wrong place is worse. 
There is another case, where there was force push and the Pr Ccommit was replaced like [this](https://github.community/t5/GitHub-API-Development-and/assiociatedPullRequests-are-lost-after-history-rewrite/m-p/26950#M2187) idiot

source/features/deep-blame.tsx
```diff
@@ -0,0 +1,140 @@
...
+		const prNumber = select('.issue-link', currentParentElement)!.textContent!.replace('#', '');
+		const commitHash = getCommitHash(pullRequest);
+
+		if (select.exists('a', versionsParent!)) {
+			const versionsElement = select('a', versionsParent!)!;
+			versionsElement.setAttribute('aria-label', 'View blame prior to this change. Press Alt to view the `deep-blame`');
+			versionsElement.classList.add('rgh-deep-blame');
+			versionsElement.dataset.commit = commitHash;
+			versionsElement.dataset.lineNumber = currentLineNumber;
+			versionsElement.dataset.prNumber = prNumber;
```
fregante: Why add these here (and in the link below)? They can just be found inside `redirectToBlameCommit`

In this function you just need to change the label (or add the icon)

source/features/deep-blame.tsx
```diff
@@ -0,0 +1,140 @@
...
+		} else {
+			versionsParent!.append(
+				<a
+					data-commit={commitHash}
+					data-line-number={currentLineNumber}
+					data-pr-number={prNumber}
+					aria-label="View `deep-blame` prior to this change"
+					className="reblame-link link-hover-blue no-underline tooltipped tooltipped-e d-inline-block pr-1 rgh-deep-blame"
+				> {versionIcon()}
```
fregante:
```suggestion
				>
					{versionIcon()}
```

source/features/deep-blame.tsx
fregante: > like [this](https://github.community/t5/GitHub-API-Development-and/assiociatedPullRequests-are-lost-after-history-rewrite/m-p/26950#M2187)

In that case, a more lax version would actually work for them, especially if we just query the PR's last commit instead of using `associatedPullRequests`

If we use `associatedPullRequests`, that won't work regardless of this suggested change.

fregante: We can leave it as is, with the `alert`

fregante: > I am not sure. I am so lost. LOL

You can slow down

yakov116:
@fregante I will be away from my computer until tomorrow night.

I will try to update then

source/features/deep-blame.tsx
yakov116: But I need the file name

fregante: D'oh, yeah

source/features/deep-blame.tsx
yakov116: When I am doing the loop I am looking at the pull request number. The version element is not within the same div. So to get the info I would need to do 2x closest. Which is why I saved everything as a dataset

source/features/deep-blame.tsx
yakov116: The function is used differently. Its taking the next element. By the tags and the merge commits we took it from the dataset .

https://github.com/sindresorhus/refined-github/blob/4f25a67d0688afe55c85f95f1ca720c172553150/source/features/mark-merge-commits-in-list.tsx#L33-L35

source/features/deep-blame.tsx
yakov116: I reverted the this suggestion. Yours did not work.

yakov116:
@fregante is there a way to load all the commits from cashe and if the commit is there and the commit does not exist then dont show the icon?

fregante:
That’s what the first version was, we’ve been through that already. 

I think we shouldn’t consider cache that much, other than for the current session. These actions are very rare (they are per file and per commit)

source/features/deep-blame.tsx
fregante: There’s no way you have to do closest twice. Closest is towards the closest common ancestor, by definition once you found one, the others are also common.

source/features/deep-blame.tsx
fregante: ✅

yakov116:
> That’s what the first version was, we’ve been through that already.
> 
> I think we shouldn’t consider cache that much, other than for the current session. These actions are very rare (they are per file and per commit)

How do I add cache for only that session?

yakov116:
image

I forgot to attach this...
I switched the alert to this. Is this good? Or should I go back to alert?

source/features/deep-blame.tsx
yakov116:
```js
const {[associatedPR]} = repository.object.associatedPullRequests;
```
I am assuming this was a mistake.

I changed it to 
```js
const associatedPR = repository.object.associatedPullRequests.nodes[0];
```

source/features/deep-blame.tsx
yakov116: You are correct. I did not word it correctly.

The pull request popup ( this is the one were scanning for in the beginning) has all the elements we need right next to it. `('.blame-hunk')` 
Otherwise I would need to go up to the parentElement and then to `('.blame-hunk')` and get all the info.

I you feel that is better, I will change it.

fregante:
That only appears at the top; the user might have scrolled down already.

And no, don't scroll up to it, the user will lose their position on the page.

fregante:
> How do I add cache for only that session?

Use `mem` instead of `cache.function`. My point was that the action is so infrequent that cache is nearly useless.

source/features/deep-blame.tsx
fregante: > I you feel that is better, I will change it.

Yes

yakov116:
image

source/features/deep-blame.tsx
```diff
@@ -44,13 +44,16 @@ const getPullRequestBlameCommit = mem(async (commit: string, prNumber: number):
 	return false;
 });
 
-function getCommitHash(commit: HTMLElement): string {
-	return (commit.nextElementSibling as HTMLAnchorElement).href.split('/').pop()!;
+function getCommitHash(commit: HTMLLinkElement): string {
```
fregante:
```suggestion
function getCommitHash(commit: HTMLAnchorElement): string {
```

`HTMLLinkElement` is `<link>`

source/features/deep-blame.tsx
```diff
@@ -44,13 +44,16 @@ const getPullRequestBlameCommit = mem(async (commit: string, prNumber: number):
...
+	return commit.href.split('/').pop()!;
 }
 
 async function redirectToBlameCommit(event: DelegateEvent<MouseEvent, HTMLAnchorElement>): Promise<void> {
 	const blameLink = event.delegateTarget;
-	const {lineNumber, commit: pullRequestCommit, prNumber} = blameLink.dataset;
+	const currentParentElement = blameLink.closest('.blame-hunk')!;
+	const lineNumber = select('.js-line-number', currentParentElement)!.textContent!;
+	const prNumber = looseParseInt(select('.issue-link', currentParentElement)!.textContent!);
+	const pullRequestCommit = getCommitHash(select<HTMLLinkElement>('a.message', currentParentElement)!);
```
fregante: It's used once, you can inline it
```suggestion
	const pullRequestCommit = select<HTMLAnchorElement>('a.message', currentParentElement)!.href.split('/').pop()!;
```

source/features/deep-blame.tsx
```diff
@@ -44,13 +44,16 @@ const getPullRequestBlameCommit = mem(async (commit: string, prNumber: number):
...
-function getCommitHash(commit: HTMLElement): string {
-	return (commit.nextElementSibling as HTMLAnchorElement).href.split('/').pop()!;
+function getCommitHash(commit: HTMLLinkElement): string {
+	return commit.href.split('/').pop()!;
 }
 
 async function redirectToBlameCommit(event: DelegateEvent<MouseEvent, HTMLAnchorElement>): Promise<void> {
 	const blameLink = event.delegateTarget;
-	const {lineNumber, commit: pullRequestCommit, prNumber} = blameLink.dataset;
+	const currentParentElement = blameLink.closest('.blame-hunk')!;
```
fregante: You could rename this to `blameHunk`. "Hunk" is the git term for "group of lines"
```suggestion
	const blameHunk = blameLink.closest('.blame-hunk')!;
```

source/features/deep-blame.tsx
yakov116: >HTMLLinkElement is <link> 
oh. What is an A?

source/features/deep-blame.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+	blameLink.firstElementChild!.replaceWith(githubSpinner);
+	// Hide tooltip after click, it’s shown on :focus
+	blurAccessibly(blameLink);
+
+	const prBlameCommit = await getPullRequestBlameCommit(prCommit, Number(prNumber));
+	if (prBlameCommit) {
+		const href = new URL(location.href.replace(getReference()!, prBlameCommit));
+		href.hash = 'L' + lineNumber;
+		location.href = String(href);
```
fregante: We have a link, let's use it to our advantage
```suggestion
		blameLink.pathname = location.pathname.replace(getReference()!, prBlameCommit);
		blameLink.hash = 'L' + lineNumber;
		blameLink.click();
```

source/features/deep-blame.tsx
```diff
@@ -0,0 +1,127 @@
...
+	const mergeCommit = associatedPR.mergeCommit.oid;
+
+	if (associatedPR.number === prNumber && mergeCommit === commit) {
+		return associatedPR.commits.nodes[0].commit.oid;
+	}
+
+	return false;
+});
+
+async function redirectToBlameCommit(event: DelegateEvent<MouseEvent, HTMLAnchorElement>): Promise<void> {
```
fregante:
```suggestion
async function redirectToBlameCommit(event: DelegateEvent<MouseEvent, HTMLAnchorElement>): Promise<void> {
	if (blameLink.href) {
		if (event.altKey) {
			event.preventDefault();
		} else {
			return; // Unmodified click on regular link: let it proceed
		}
	}
```

source/features/deep-blame.tsx
```diff
@@ -0,0 +1,127 @@
...
+	const blameLink = event.delegateTarget;
+	const blameHunk = blameLink.closest('.blame-hunk')!;
+	const lineNumber = select('.js-line-number', blameHunk)!.textContent!;
+	const prNumber = looseParseInt(select('.issue-link', blameHunk)!.textContent!);
+	const prCommit = select<HTMLAnchorElement>('a.message', blameHunk)!.href.split('/').pop()!;
+	if (blameLink.href && event.altKey) {
+		event.preventDefault();
+	}
```
fregante:
```suggestion
```

source/features/deep-blame.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+function init(): void|false {
+	const pullRequests = select.all('[data-hovercard-type="pull_request"]');
+	if (pullRequests.length === 0) {
+		return false;
+	}
+
+	for (const pullRequest of pullRequests) {
+		const currentParentElement = pullRequest.closest('.blame-hunk')!;
+		const versionsParent = select('.blob-reblame', currentParentElement);
```
fregante: Use `!` as early as possible so you don't have to use it every time later
```suggestion
		const versionsParent = select('.blob-reblame', currentParentElement)!;
```

source/features/deep-blame.tsx
```diff
@@ -0,0 +1,132 @@
+import './deep-blame.css';
+import blurAccessibly from 'blur-accessibly';
+import delegate, {DelegateEvent} from 'delegate-it';
+import mem from 'mem';
+import React from 'dom-chef';
+import versionIcon from 'octicon/versions.svg';
+import select from 'select-dom';
+import * as api from '../libs/api';
+import features from '../libs/features';
+import {getRepoGQL, getReference, looseParseInt} from '../libs/utils';
```
fregante: Can these imports be sorted? And for future PRs as well

source/features/deep-blame.tsx
```diff
@@ -0,0 +1,132 @@
...
+	// Restore the regular version link if there was one
+	if (blameLink.href) {
+		blameLink.setAttribute('aria-label', 'View blame prior to this change.');
+		blameLink.classList.remove('rgh-deep-blame');
+		githubSpinner.replaceWith(versionIcon());
+	} else {
+		githubSpinner.remove();
+	}
+
+	alert('Unable to find linked Pull Request.');
```
fregante:
```suggestion
	alert('The PR linked in the title didn’t create this commit');
```

source/features/deep-blame.tsx
```diff
@@ -0,0 +1,132 @@
...
+}
+
+function init(): void|false {
```
fregante:
```suggestion
function init(): void | false {
```

Manually "lint" your PRs please, don't expect me to bug you in every single PR

source/features/deep-blame.tsx
```diff
@@ -0,0 +1,132 @@
...
+	for (const pullRequest of pullRequests) {
+		const currentParentElement = pullRequest.closest('.blame-hunk')!;
+		const versionsParent = select('.blob-reblame', currentParentElement)!;
+
+		if (select.exists('a', versionsParent)) {
+			const versionsElement = select('a', versionsParent)!;
+			versionsElement.setAttribute('aria-label', 'View blame prior to this change. Press Alt to view the `deep-blame`');
+			versionsElement.classList.add('rgh-deep-blame');
+		} else {
+			versionsParent.append(
```
fregante:
```suggestion
	for (const pullRequest of pullRequests) {
		const hunk = pullRequest.closest('.blame-hunk')!;

		const reblameLink = select('.reblame-link', hunk)!;
		if (reblameLink) {
			reblameLink.setAttribute('aria-label', 'View blame prior to this change. Hold Alt to extract commits from this PR first');
			reblameLink.classList.add('rgh-deep-blame');
		} else {
			select('.blob-reblame', hunk)!.append(
```

source/features/deep-blame.tsx
```diff
@@ -0,0 +1,132 @@
...
+				<a
+					aria-label="View `deep-blame` prior to this change"
```
fregante:
```suggestion
					aria-label="View blame prior to this change (extracts commits from this PR first)"
```

source/features/deep-blame.tsx
```diff
@@ -0,0 +1,132 @@
...
+			);
+		}
+	}
+
+	delegate('.rgh-deep-blame', 'click', redirectToBlameCommit);
+}
+
+features.add({
+	id: __featureName__,
+	description: 'Add Deep Blame to commits',
```
fregante:
```suggestion
	description: 'When exploring blames, `alt`-clicking the "Reblame" buttons will extract the associated PR’s commits first, instead of treating the commit a single change.',
```

source/features/deep-blame.tsx
```diff
@@ -0,0 +1,132 @@
...
+	screenshot: 'https://user-images.githubusercontent.com/16872793/77025598-e9022000-6967-11ea-92dd-6f49875a225e.png',
```
fregante: This needs to be updated once the copy is confirmed

source/features/deep-blame.tsx
```diff
@@ -0,0 +1,132 @@
...
+	const prNumber = looseParseInt(select('.issue-link', blameHunk)!.textContent!);
+	const prCommit = select<HTMLAnchorElement>('a.message', blameHunk)!.href.split('/').pop()!;
+
+	const githubSpinner = (
+		<img
+			src="https://github.githubassets.com/images/spinners/octocat-spinner-128.gif"
+			className="mr-2"
+			width="18"
+		/>
+	);
```
fregante: Can you extract this into a file exactly as asked in the other PR?

```js
export default function loadingIcon (): HTMLElement {
	return (
		<img
			src="https://github.githubassets.com/images/spinners/octocat-spinner-128.gif"
			width="18"
		/>
	);
}
```

Then use it as:

```js
const spinner = loadingIcon();
spinner.classList.add('mr-2');
```

https://github.com/vadimdemedes/dom-chef/issues/55 will eventually let us use it as 

```jsx
const spinner = <LoadingIcon className="mr-2"/>;

File: source/features/deep-blame.tsx [hunk 35]
```diff
@@ -0,0 +1,132 @@
...
+
+	blameLink.firstElementChild!.replaceWith(githubSpinner);
+	// Hide tooltip after click, it’s shown on :focus
+	blurAccessibly(blameLink);
+
+	const prBlameCommit = await getPullRequestBlameCommit(prCommit, Number(prNumber));
+	if (prBlameCommit) {
```
fregante:
```suggestion
	if (prBlameCommit) {
	 	const lineNumber = select('.js-line-number', blameHunk)!.textContent!;
```

source/features/deep-blame.tsx
```diff
@@ -0,0 +1,132 @@
...
+	if (blameLink.href) {
+		if (event.altKey) {
+			event.preventDefault();
+		} else {
+			return; // Unmodified click on regular link: let it proceed
+		}
+	}
+
+	const blameHunk = blameLink.closest('.blame-hunk')!;
+	const lineNumber = select('.js-line-number', blameHunk)!.textContent!;
```
fregante:
```suggestion
```

source/features/deep-blame.tsx
```diff
@@ -0,0 +1,132 @@
...
+		return false;
+	}
+
+	for (const pullRequest of pullRequests) {
+		const currentParentElement = pullRequest.closest('.blame-hunk')!;
+		const versionsParent = select('.blob-reblame', currentParentElement)!;
+
+		if (select.exists('a', versionsParent)) {
+			const versionsElement = select('a', versionsParent)!;
+			versionsElement.setAttribute('aria-label', 'View blame prior to this change. Press Alt to view the `deep-blame`');
```
fregante:
```suggestion
			versionsElement.setAttribute('aria-label', 'View blame prior to this change. Hold Alt to extract commits from this PR first');
```

source/features/deep-blame.tsx
```diff
@@ -0,0 +1,132 @@
...
+	for (const pullRequest of pullRequests) {
+		const currentParentElement = pullRequest.closest('.blame-hunk')!;
+		const versionsParent = select('.blob-reblame', currentParentElement)!;
+
+		if (select.exists('a', versionsParent)) {
+			const versionsElement = select('a', versionsParent)!;
+			versionsElement.setAttribute('aria-label', 'View blame prior to this change. Hold Alt to extract commits from this PR first');
+			versionsElement.classList.add('rgh-deep-blame');
+		} else {
+			versionsParent.append(
```
fregante:
```suggestion
		const hunk = pullRequest.closest('.blame-hunk')!;

		const reblameLink = select('.reblame-link', hunk)!;
		if (reblameLink) {
			reblameLink.setAttribute('aria-label', 'View blame prior to this change. Hold Alt to extract commits from this PR first');
			reblameLink.classList.add('rgh-deep-blame');
		} else {
			select('.blob-reblame', hunk)!.append(
```

source/content.ts
```diff
@@ -122,6 +122,7 @@ import './features/update-pr-from-base-branch';
 import './features/hide-disabled-milestone-sorter';
 import './features/tag-changelog-link';
 import './features/link-to-file-in-file-history';
+import './features/deep-blame';
```
fregante: Maybe this. It's what GitHub calls the button
```suggestion
import './features/deep-reblame';
```

fregante:
asd

It freezes the browser in Firefox, crashing all extensions

source/features/deep-reblame.tsx
```diff
@@ -0,0 +1,125 @@
...
+	for (const pullRequest of pullRequests) {
+		const hunk = pullRequest.closest('.blame-hunk')!;
+
+		const reblameLink = select('.reblame-link', hunk)!;
+		if (reblameLink) {
+			reblameLink.setAttribute('aria-label', 'View blame prior to this change. Hold `Alt` to extract commits from this PR first');
+			reblameLink.classList.add('rgh-deep-blame');
+		} else {
+			select('.blob-reblame', hunk)!.append(
+				<a
```
fregante: Plain links aren't focusable and the cursor doesn't change to `pointer`

Since we're using this link to navigate later (so we can't change it to the more appropriate `<buton type=button>`) I suggest
```suggestion
				<a
					href="#deep-reblame"
```

And then the condition on line 55 and 77 needs to be updated

yakov116:
> asd
> 
> It freezes the browser in Firefox, crashing all extensions

See Im talented.
What could be causing it? Github sometimes takes its time to respond

source/features/deep-reblame.tsx
yakov116: Can I do `href="#"` and then
	if href.length > 1

fregante: No, `#` go to the top

yakov116:
@fregante what did you do to get it to freeze? I tried it.

yakov116:
@fregante looks like if I add an href github replaces the ref with a direct link to the same line. breaking the entire feature

fregante:
> @fregante what did you do to get it to freeze? I tried it.


1. Visit /source/features/link-to-file-in-file-history.tsx@0ca864a#L29 (blame)
2. Click deep reblame

source/features/deep-reblame.css
```diff
@@ -0,0 +1,7 @@
+a:not([href]).rgh-deep-blame .octicon-versions {
```
fregante: ⬆️

yakov116: I didnt fix it since 

> @fregante looks like if I add an href github replaces the ref with a direct link to the same line. breaking the entire feature

try it you will see

fregante:
Same happens in Chrome, here's where it's looping:

Screen Shot 2020-03-23 at 14 56 15

source/features/deep-reblame.css
fregante: > @fregante looks like if I add an href github replaces the ref with a direct link to the same line. breaking the entire feature

It's us: `link-to-prior-blame-line`

yakov116:
> Same happens in Chrome, here's where it's looping:
> 
> Screen Shot 2020-03-23 at 14 56 15

What would be the issue?

fregante:
It's an infinite loop

source/features/deep-reblame.css
yakov116: ahh so we need to run that after

yakov116:
I need help how to debug it

fregante:
I think I got it. The `.click()` is triggering the delegate event, which tries the redirect via `.click()`, which triggers the redirect, etc. 

Let’s undo my earlier suggestion and instead change page via `location.href`

fregante:
This will also let you use `<button type=button>`

yakov116:
> This will also let you use `<button type=button>`

And Remove the href right?

source/features/deep-reblame.css
```diff
@@ -1,7 +1,7 @@
-a:not([href='#deep-reblame']).rgh-deep-blame .octicon-versions {
+button.rgh-deep-reblame .octicon-versions {
 	color: #e40de48c;
 }
 
-a.rgh-deep-blame .octicon-versions:hover {
+a.rgh-deep-reblame .octicon-versions:hover, button.rgh-deep-reblame .octicon-versions:hover {
```
fregante:
```suggestion
.rgh-deep-reblame .octicon-versions:hover {
```

source/features/deep-reblame.tsx
```diff
@@ -47,7 +47,7 @@ const getPullRequestBlameCommit = mem(async (commit: string, prNumber: number):
 
 async function redirectToBlameCommit(event: DelegateEvent<MouseEvent, HTMLAnchorElement>): Promise<void> {
 	const blameLink = event.delegateTarget;
-	if (blameLink.href !== '#deep-reblame') {
+	if (blameLink.href) {
```
fregante:
```suggestion
	if (blameButton instanceof HTMLButtonElement) {
```

source/features/deep-reblame.tsx
```diff
@@ -47,7 +47,7 @@ const getPullRequestBlameCommit = mem(async (commit: string, prNumber: number):
 
 async function redirectToBlameCommit(event: DelegateEvent<MouseEvent, HTMLAnchorElement>): Promise<void> {
 	const blameLink = event.delegateTarget;
```
fregante: This is no longer just an `<a>`

```suggestion
async function redirectToBlameCommit(event: DelegateEvent<MouseEvent, HTMLAnchorElement | HTMLButtonElement>): Promise<void> {
	const blameElement = event.delegateTarget;
```

source/features/deep-reblame.tsx
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	const blameHunk = blameLink.closest('.blame-hunk')!;
+	const prNumber = looseParseInt(select('.issue-link', blameHunk)!.textContent!);
+	const prCommit = select<HTMLAnchorElement>('a.message', blameHunk)!.href.split('/').pop()!;
+	const spinner = loadingIcon();
+	spinner.classList.add('mr-2');
+	blameLink.firstElementChild!.replaceWith(spinner);
+	// Hide tooltip after click, it’s shown on :focus
+	blurAccessibly(blameLink);
```
fregante:
```suggestion

	blurAccessibly(blameLink); // Hide tooltip after click, it’s shown on :focus
```

source/features/deep-reblame.tsx
yakov116: I need the other way around (Not a button element) What is that called (an A element)

source/features/deep-reblame.tsx
```diff
@@ -0,0 +1,126 @@
...
+	const blameHunk = blameElement.closest('.blame-hunk')!;
+	const prNumber = looseParseInt(select('.issue-link', blameHunk)!.textContent!);
+	const prCommit = select<HTMLAnchorElement>('a.message', blameHunk)!.href.split('/').pop()!;
+	const spinner = loadingIcon();
+	spinner.classList.add('mr-2');
+	blameElement.firstElementChild!.replaceWith(spinner);
+
+	blurAccessibly(blameElement); // Hide tooltip after click, it’s shown on :focus
+
+	const prBlameCommit = await getPullRequestBlameCommit(prCommit, Number(prNumber));
```
fregante:
```suggestion
	const prBlameCommit = await getPullRequestBlameCommit(prCommit, prNumber);
```

source/features/deep-reblame.tsx
```diff
@@ -0,0 +1,126 @@
...
+					}
+				}
+			}
+		}
+	`);
+
+	const associatedPR = repository.object.associatedPullRequests.nodes[0];
+	const mergeCommit = associatedPR.mergeCommit.oid;
+
+	if (associatedPR.number === prNumber && mergeCommit === commit) {
```
yakov116: This will always fail now `associatedPR.number === prNumber `

yakov116: Never mind I forgot we used `looseParseInt`

source/features/deep-reblame.tsx
```diff
@@ -0,0 +1,126 @@
...
+							}
+						}
+					}
+				}
+			}
+		}
+	`);
+
+	const associatedPR = repository.object.associatedPullRequests.nodes[0];
+	const mergeCommit = associatedPR.mergeCommit.oid;
```
fregante: Errors: https://github.com/sindresorhus/refined-github/blame/d1e27cb34f18b9d5d57fdbf853e5b15c3a660d85/source/features/deep-reblame.tsx#L67

```
Uncaught (in promise) TypeError: Cannot read property 'oid' of null
```

Maybe this is all that's needed, test it

```suggestion
	const mergeCommit = associatedPR.mergeCommit?.oid;
```

yakov116: Good catch

yakov116: Needed a full is statement. since associatedPR returns undefined

fregante:
# 🙌

yakov116:
Wow never thought this would happen. Thanks for keeping with me! (while I have to work from home and I **very** busy as you can see)

fregante:
https://github.com/sindresorhus/refined-github/blame/fad67fb8d8fcf608601c9b8ae1b5f12337cbccce/source/options.html#L1

This button points to a 404

Issue #2889, opened 2020-03-14
Incorrect amount of bugs shown (due to cache)


szmarczak:
![image](https://user-images.githubusercontent.com/36894700/76682971-ac999180-6600-11ea-866d-25a62b2c0309.png)

https://github.com/szmarczak/cacheable-lookup

fregante:
Cache. It expires within an hour and is refreshed in the next pageload

szmarczak:
Well then it should update when somebody clicks on the tab, right?

fregante:
There’s currently no mechanism to force a refresh manually. I’m thinking of changing the cache API so that you could set the cache manually in these cases. For example:

```js
const countBugs = cache.function(…, {options});

…

if (isBugsPage) {
  countBugs.set(getCountFromDom()); // this sets the cache using the options already defined for cache.function
}

const bugs = await countBugs()
``` 

cc @sindresorhus a new API isn’t strictly necessary, but it helps reuse the logic/options without having to manually create additional constants. 

API suggestions welcome

szmarczak:
> Cache. It expires within an hour and is refreshed in the next pageload

It never expires for me. It has been days and I still see `1`. Reload doesn't do anything.

fregante:
Chrome or Firefox?

Try this in the background script inspector

```
await browser.storage.get('bugs-tab: szmarczak/cacheable-lookup')

**Comment by fregante** [comment]
The cache should be there for 4 days, but if you reload the page twice after half an hour it should be expired/updated

**Comment by szmarczak** [comment]
> Chrome or Firefox?

Chrome.

![image](https://user-images.githubusercontent.com/36894700/76700644-cf3bb100-66b9-11ea-908d-24b5de976f7f.png)

> So if you load the page now it should definitely show 0, otherwise it’s not working as expected

It still shows `1`.

**Comment by fregante** [comment]
My bad,

```
await browser.storage.local.get('bugs-tab:szmarczak/cacheable-lookup')
```

szmarczak:
![image](https://user-images.githubusercontent.com/36894700/76868913-da344400-6867-11ea-9920-e25822bde51d.png)

fregante:
I suggesting emptying the cache, it's at the bottom of the options page

szmarczak:
Yep, that worked!

szmarczak:
Why has this been closed? Is it solved already?

PR #2898, opened 2020-03-16
`bugs-tab`: Use information from page

fregante:
Fixes https://github.com/sindresorhus/refined-github/issues/2889#issuecomment-599179639

This code is a mess though, every time I try to get data from the DOM _and_ the API it's more complex than I thought.

I'm not sure I'll merge this, the cache is set to expire after 30 minutes already, it's not the end of the world.

### DIFF
--- a/source/features/bugs-tab.tsx
+++ b/source/features/bugs-tab.tsx
@@ -8,6 +8,11 @@ import SearchQuery from '../libs/search-query';
 import {getRepoGQL, getRepoURL} from '../libs/utils';
 
 const numberFormatter = new Intl.NumberFormat();
+const cacheOptions = {
+	maxAge: 1 / 24 / 2, // Stale after half an hour
+	staleWhileRevalidate: 4,
+	cacheKey: (): string => __featureName__ + ':' + getRepoURL()
+};
 const countBugs = cache.function(async (): Promise<number> => {
 	const {repository} = await api.v4(`
 		repository(${getRepoGQL()}) {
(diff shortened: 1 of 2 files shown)


source/features/bugs-tab.tsx
```diff
@@ -18,11 +23,13 @@ const countBugs = cache.function(async (): Promise<number> => {
...
-}, {
-	maxAge: 1 / 24 / 2, // Stale after half an hour
-	staleWhileRevalidate: 4,
-	cacheKey: (): string => __featureName__ + ':' + getRepoURL()
-});
+}, cacheOptions);
+
+function cleanForComparison(searchQuery: SearchQuery): SearchQuery {
+	const newQuery = searchQuery.clone();
+	newQuery.remove('sort:updated-desc', 'is:open', 'is:closed');
```
busches: Do we need to remove @me and author:

fregante:
No, we need to match the exact query because anything else will change the number. 

`newsfeed label:bug` will show different results and result counts, that’s why this comparison is so complex

source/features/bugs-tab.tsx
fregante: Only these queries don’t change the numbers on the page. `author` does

fregante:
I think I'll just make an API to force an refresh from the source, this method isn't sustainable.

Example:

```js
const isBugsPage = new SearchQuery(location).includes('label:bug');
const countPromise = isBugsPage ? countBugs.skipCache() : countBugs();
```

PR #2905, opened 2020-03-20
Exclude pre-releases from `latest-tag-button`

### LINKED ISSUE #2239: Differentiate between releases and pre-releases
Releases can be created from tags, which can then be further classified as pre-release or the default of the latest release.
Screen Shot 2019-07-15 at 6 47 35 PM

The latest release widget displays a release marked as pre-release as the latest release.
Screen Shot 2019-07-15 at 6 46 31 PM

The API supports the ability to differentiate between the two.
https://developer.github.com/v3/repos/releases/#list-releases-for-a-repository

fregante:
Fixes #2239 

## Tests

I can't find a repo that has pre-releases right now, but I also **included some automated tests** to make sure the algorithm works.

### DIFF
--- a/source/features/latest-tag-button.tsx
+++ b/source/features/latest-tag-button.tsx
@@ -4,18 +4,12 @@ import cache from 'webext-storage-cache';
 import select from 'select-dom';
 import tagIcon from 'octicon/tag.svg';
 import elementReady from 'element-ready';
-import compareVersions from 'tiny-version-compare';
 import * as api from '../libs/api';
 import features from '../libs/features';
 import fetchDom from '../libs/fetch-dom';
 import {isRepoRoot} from '../libs/page-detect';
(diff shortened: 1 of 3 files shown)


staabm:
https://github.com/desktop/desktop Has some recent pre-releases

fregante:
And?

staabm:
Nope, I just wanted to let you know a example repo where the feature can be tested

fregante:
The point of this feature is to tell visitors **“the version you’re using is newer than the one shown here”**

Usually people don’t use pre-releases, especially as dependencies, so it would make sense to exclude them

fregante:
Ah you answered my request. Thank you ☺️

fregante:
My bad, I totally misread that comment 🤦‍♂️

PR #2915, opened 2020-03-21
Avoid losing options when removing the github.com permission

### LINKED ISSUE #2908: Extension options are reset when changing site access
When I change this to "When you click the extension", all my extension options get reset:
image

jack1142:
Closes #2908

### DIFF
--- a/source/options-storage.ts
+++ b/source/options-storage.ts
@@ -74,7 +74,8 @@ async function initializeAllOptions(): Promise<void> {
 	// Remove old domains
 	browser.permissions.onRemoved!.addListener(({origins}) => {
 		if (origins) {
-			const optionKeysToRemove = origins.map(origin => getStorageName(new URL(origin).host));
+			const optionKeysToRemove = origins.map(origin => getStorageName(new URL(origin).host))
+				.filter(key => key !== 'options');
 			browser.storage.sync.remove(optionKeysToRemove);
 		}
 	});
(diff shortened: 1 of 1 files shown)


source/options-storage.ts
```diff
@@ -74,7 +74,8 @@ async function initializeAllOptions(): Promise<void> {
 	// Remove old domains
 	browser.permissions.onRemoved!.addListener(({origins}) => {
 		if (origins) {
-			const optionKeysToRemove = origins.map(origin => getStorageName(new URL(origin).host));
+			const optionKeysToRemove = origins.map(origin => getStorageName(new URL(origin).host))
+				.filter(key => key !== 'options');
```
fregante:
```suggestion
			const optionKeysToRemove = origins
				.filter(origin => origin !== 'https://github.com')
				.map(origin => getStorageName(new URL(origin).host));
```

jack1142: It doesn't - the actual origin would be `https://github.com/*` but still, it's not the only domain for which `getStorageName` can return `options` (`https://api.github.com/*`, `https://gist.github.com/*`)

fregante:
```suggestion
			const optionKeysToRemove = origins
				.map(origin => getStorageName(new URL(origin).host))
				.filter(key => key !== 'options');
```

fregante:
Thank you! 🙌

PR #2919, opened 2020-03-22
Improve error UX

fregante:
1. `content.js` was renamed because it makes console messages impossible to tell apart; `refined-github.js` makes it super clear.

2. I had written the same logging/reporting function twice

3. console


fregante:
Honestly this is great and only makes me hate the "default" `content.*` naming more. You can instantly find the origin of a script or style:

css

PR #2921, opened 2020-03-23
Make `forked-to`'s cache case-insensitive

### LINKED ISSUE #2909: `forked-to` only works on some repositories
Out of 31 source repos of the forks I have on my account, only on 3 of them (funnily, refined-github is one of them) `forked-to` seems to work.
Example repo where it doesn't work: https://github.com/Cog-Creators/Red-DiscordBot
Screenshot:
image

jack1142:
Fixes #2909

### DIFF
--- a/source/features/forked-to.tsx
+++ b/source/features/forked-to.tsx
@@ -12,7 +12,7 @@ import {isForkedRepo} from '../libs/page-detect';
 import {getRepoURL, getUsername, getForkedRepo} from '../libs/utils';
 
 const getForkSourceRepo = (): string => getForkedRepo() ?? getRepoURL();
-const getCacheKey = (): string => `forked-to:${getUsername()}@${getForkSourceRepo()}`;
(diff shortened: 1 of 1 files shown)


fregante:
I think we should change the original function, so it matches `getRepoURL()`

https://github.com/sindresorhus/refined-github/blob/4b8056ab2d53a9a7ba20130a38f06dec0813fc98/source/libs/utils.ts#L93

jack1142:
I was thinking about that but I noticed that one of functions using it is user-facing:
https://github.com/sindresorhus/refined-github/blob/6c46766237ccbf8a8c23ba07aca1e48a11a71b4b/source/features/show-open-prs-of-forks.tsx#L77

I can change it but what should I do about this occurrence?

fregante:
You're right, it's not always safe

fregante:
Thank you for the fix!

PR #2923, opened 2020-03-23
Make `show-open-prs-of-forks`'s cache case-insensitive

jack1142:
Something I spotted when commenting on #2921, it doesn't cause any issues as far as I'm aware but it's there if you want it ;)


fregante:
Good point. I think this specific case isn't an issue though. `getRepoURL()` is end-user-dependant, because it's just a URL they entered, but `getForkedRepo()` comes straight from GitHub's DOM.

Unless it's being compared to `getRepoURL()` anywhere, I don't think it needs to be normalized.

PR #2925, opened 2020-03-23
Allow multiple loaders in a single `features.add`

fregante:
Not yet tested

1. Allow multiple loaders in a single `features.add` (avoids having to set useless `description: ''` properties)
2. Allow any loading method combination (avoids weird functions like `nowAndOnAjaxedPages`)

I'd also like to remove description/screenshot from `features.add`, ideally as `export description = ''` so they will be tree-shaken.

**Suggestions?** I need something that can be validated/required like TypeScript does


yakov116:
@fregante  you marked this as help wanted. How can I help?

fregante:
> I'd also like to remove description/screenshot from `features.add`, ideally as `export description = ''` so they will be tree-shaken.
> 
> **Suggestions?** I need something that can be validated/required like TypeScript does

This part

fregante:
The idea of adding multiple load events has to be rethought from scratch because it can't work. There are 2 very different events:

1. page change
1. something on the current page changed

And this representation alone is wrong too, it should be:

1. page change
   a. something on the current page changed

What happens on 1 that has to happen _before_ 1a:
  - deinit
  - page-detect

Even the previous mechanism is probably broken (e.g. `features.onNewComments`)

PR #2930, opened 2020-03-25
Add watch:chrome

yakov116:
web-ext added this option.

@fregante is there a simple way to set a secret-key to add the api key. Otherwise its a pain since every time you relaunch you need to set api key again
<!--

Thanks for contributing! 🍄

1. LINKED ISSUES:

2. TEST URLS:
   Add some test URLs

3. SCREENSHOT:

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
 		"build": "webpack --mode=production",
 		"watch": "webpack --mode=development --watch",
 		"watch:firefox": "web-ext run --source-dir=distribution",
+		"watch:chrome": "web-ext run --source-dir=distribution  --target chromium",
 		"prerelease:version": "dot-json distribution/manifest.json version $VER",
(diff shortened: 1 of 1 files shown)


fregante:
> Otherwise its a pain

Yep, that's why I don't use these scripts. The alternative is to use the flag to reuse your current profile instead of creating a new one, but it doesn't work reliably

yakov116:
>but it doesn't work reliably

Only works if you shut chrome entirely. any way to set the api key via a script?

yakov116:
>but it doesn't work reliably

Only works if you shut chrome entirely. any way to set the api key via a script?

fregante:
> any way to set the api key via a script?

Not that I know of

yakov116:
Going to close this then. I see no value.

PR #2933, opened 2020-03-25
Fix `close-out-of-view-modals` on more pages

yakov116:
Did some more debugging.  This should be correct.

Also fixed the error when setting your status
<!--

Thanks for contributing! 🍄

1. LINKED ISSUES:

2. TEST URLS:
   Add some test URLs

3. SCREENSHOT:


fregante:
I'm not confident that `details-menu` is used everywhere, especially in older `.dropdown-menu` items

List where you tested this or else we're doing whack-a-mole.
Perhaps include each tested dropdown in the code, e.g.:

```js
// This should work on:
// "Labels" dropdown in issues
// "Watch" dropdown on repos’ headers
```
and so on.

source/features/close-out-of-view-modals.tsx
```diff
@@ -11,10 +11,10 @@ const observer = new IntersectionObserver(([{intersectionRatio, target}]) => {
 
 function init(): void {
 	// The `open` attribute is added after this handler is run, so the selector is inverted
-	delegate('.details-overlay:not([open]) > summary', 'click', ({delegateTarget: summary}) => {
+	delegate('.details-overlay:not([open]) > summary[aria-haspopup="menu"]', 'click', ({delegateTarget: summary}) => {
```
fregante: I'd avoid this extra check, there's only ever one `summary`

yakov116: Nope, Let me get you a few links

yakov116:
> ```js
> // This should work on:
> // "Labels" dropdown in issues
> // "Watch" dropdown on repos’ headers
> ```


Side note: `"Labels" dropdown in issues,`

 With sticky side bar off....

source/features/close-out-of-view-modals.tsx
fregante: Indeed, hopefully GitHub settled on `[aria-haspopup="menu"]' + details-menu` for all. Maybe we should also make sure all of our dropdown match this as well.

fregante:
>  Did github add stick-disscussopn-sidebar? i turned it off and its still sticky

I disabled Refined GitHub, zoomed out to 25% and the sidebar isn't sticky. Did you load RG twice?

yakov116:
> > Did github add stick-disscussopn-sidebar? i turned it off and its still sticky
> 
> I disabled Refined GitHub, zoomed out to 25% and the sidebar isn't sticky. Did you load RG twice?

Ahh now I see it

source/features/close-out-of-view-modals.tsx
yakov116: In this PR?

yakov116: @fregante Done

fregante:
Excellent! This is the kind of solutions I like

PR #2945, opened 2020-03-27
Meta: Move deploy from Travis to GHA

fregante:
From https://github.com/notlmn/browser-extension-template/pull/21 and https://github.com/npmhub/npmhub/pull/101


.github/workflows/deploy-auto.yml
```diff
@@ -0,0 +1,25 @@
+# Automatically create a new version/tag, if new commits are available since the last tag
+name: Auto-tagger
+
+on:
+  push:
+    tags:
+      - deploy
```
fregante: I'm thinking this could be avoided altogether; if the tag isn't deleted in time by GHA, someone could fork it and keep the tag forever, which could cause trouble.

.github/workflows/deploy-auto.yml
```diff
@@ -0,0 +1,25 @@
...
+  schedule:
+    - cron: '51 15 */4 * *' # Every 4 days at 15:15 https://crontab.guru/#51_15_*/4_*_*
+
+jobs:
+  create-version:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
```
fregante: If we don't drop the `deploy` tag, we have to set this so it never gets stuck on "some random old commit that was tagged as `deploy` months ago"
```suggestion
      - uses: actions/checkout@v2
        with:
          ref: master
```

.github/workflows/deploy-auto.yml
fregante: Yeah, dropping it, it's not a big deal to manually creating the rare hotfix `20.5.10.1` tag

.github/workflows/deploy-auto.yml
```diff
@@ -0,0 +1,21 @@
...
+      - uses: fregante/setup-git-token@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+      - run: |
+          if [[ $(git tag -l --points-at HEAD) = "" ]]; then
+            export VER=$(npx daily-version)
```
notlmn: Versioning here is different from https://github.com/sindresorhus/refined-github/pull/2945/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R15. Should use either `20.x.x` or `2020.x.x`.

.github/workflows/deploy-auto.yml
```diff
@@ -0,0 +1,21 @@
+# Automatically create a new version/tag, if new commits are available since the last tag
+name: Auto-tagger
+
+on:
+  schedule:
+    - cron: '51 15 */4 * *' # Every 4 days at 15:15 https://crontab.guru/#51_15_*/4_*_*
```
notlmn:
```suggestion
    - cron: '51 15 */4 * *' # Every 4 days at 15:51 https://crontab.guru/#51_15_*/4_*_*
```

package.json
```diff
@@ -6,16 +6,13 @@
...
-		"prerelease:version": "dot-json distribution/manifest.json version $VER",
-		"prerelease:source-url": "echo https://github.com/sindresorhus/refined-github/archive/\"${TRAVIS_COMMIT:-$VER}\".zip > distribution/SOURCE_URL",
-		"release:amo": "cd distribution && web-ext-submit",
-		"release:cws": "cd distribution && webstore upload --auto-publish",
-		"release": "VER=$(daily-version) npm-run-all build prerelease:* create-git-tag release:*",
-		"create-git-tag": "git tag $VER -m $VER && git push origin $VER"
+		"version": "dot-json distribution/manifest.json version $VER",
+		"release:cws": "webstore upload --source=distribution --auto-publish",
+		"release:amo": "web-ext-submit --source-dir distribution",
+		"release": "VER=20$(daily-version) run-s build version release:*"
```
fregante:
```suggestion
		"release": "VER=$(daily-version) run-s build version release:*"
```

.github/workflows/deploy-auto.yml
fregante: Good catch, it was a mistake. We have `20.x.x` versions but npmhub has `2020.x.x` versions. I had copied it from there.

fregante:
Live testing on https://github.com/npmhub/npmhub/runs/540747579

There may be some issues.

notlmn:
You could also remove `.travis` directory at this point. 🤷‍♂️

fregante:
Indeed! I missed it

.github/workflows/deploy-tag.yml
```diff
@@ -0,0 +1,56 @@
...
+
+  deploy-to-chrome:
+    needs: build
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/download-artifact@v1
+      with:
+        name: built-extension
+        path: distribution
+    - run: cd distribution && npx chrome-webstore-upload-cli upload --auto-publish
```
fregante: This may be possible

> Basic (download to current working directory):

```suggestion
    - run: npx chrome-webstore-upload-cli upload --auto-publish
```

https://github.com/actions/download-artifact#usage

Also it's best to make the version sticky

```suggestion
    - run: npx chrome-webstore-upload-cli@1 upload --auto-publish
```

.github/workflows/deploy-tag.yml
```diff
@@ -0,0 +1,56 @@
...
+
+  deploy-to-firefox:
+    needs: build
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/download-artifact@v1
+      with:
+        name: built-extension
+        path: distribution
+    - run: cd distribution && npx web-ext-submit
```
fregante:
```suggestion
    - run: npx web-ext-submit@4
```

.github/workflows/deploy-auto.yml
```diff
@@ -0,0 +1,21 @@
...
+      - uses: actions/checkout@v2
+      - uses: fregante/setup-git-token@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+      - run: |
+          if [[ $(git tag -l --points-at HEAD) = "" ]]; then
+            export VER=$(npx daily-version)
+            git tag $VER -m $VER
+            git push origin $VER
+          fi
```
fregante: I hope someone at some point figures out a way to move this condition to the workflow itself, like: 
```suggestion
      - if: is HEAD not tagged?
        run: |
          VER=$(npx daily-version)
          git tag $VER -m $VER
          git push origin $VER
```

Using: https://help.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions

fregante:
I just disabled Travis and its cronjob. It had sneaked its last cronjob just an hour ago...



_Thanks, [Obama]()._

.github/workflows/deploy-tag.yml
notlmn: Looking at https://github.com/DrewML/chrome-webstore-upload-cli/blob/master/wrapper.js#L28-L30, wouldn't the CRX now contain other non-essential files like `package.json` and potentially `node_modules`? Or am I missing something?

package.json
```diff
@@ -9,13 +9,10 @@
 		"test:js": "TS_NODE_FILES=true ava --serial # Due to Travis",
 		"build": "webpack --mode=production",
 		"watch": "webpack --mode=development --watch",
-		"watch:firefox": "web-ext run --source-dir=distribution",
```
FloEdelmann, line 12: @fregante Was it a mistake that you deleted this line? I always use it for development.

fregante: I removed it because of this https://github.com/sindresorhus/refined-github/pull/2683#issuecomment-572716508

PR #2947, opened 2020-03-27
Add `parse-backticks`  to more places

yakov116:
Test on dashboard
image

And test on https://github.com/sindresorhus/refined-github/blame/14565644541a4ac720d9f6db27236496fbe075c1/source/features/default-branch-button.tsx

and test on https://github.com/sindresorhus/refined-github/commits?author=fregante

and test on https://github.com/sindresorhus/refined-github/compare/WIP-parcel...master

and test on https://github.com/sindresorhus/refined-github/releases/tag/20.4.1 and https://github.com/yakov116/TestR/releases/tag/1
and test on https://github.com/sindresorhus/refined-github/actions
and test on https://github.com/sindresorhus/refined-github/actions/runs/68625874
and test on https://github.com/yakov116/TestR/wiki/%60parse-me%60

@fregante don't grill me please

<!--

Thanks for contributing! 🍄

1. LINKED ISSUES:

2. TEST URLS:
   Add some test URLs

3. SCREENSHOT:


source/features/parse-backticks.tsx
```diff
@@ -15,18 +15,28 @@ function initGeneral(): void {
...
-		'.js-pinned-issue-list-item > span' // Pinned Issues
+		'.js-pinned-issue-list-item > span', // Pinned Issues
+		'blame-commit-message a' // `isBlame`
 	])) {
 		parseBackticks(title);
 	}
 }
 
+export function parseBackticksOnNewsFeedItems(): void {
+	for (const title of select.all('.issues_labeled .text-gray-dark > a:not(.rgh-parse-backticks), .commits blockquote:not(.rgh-parse-backticks)')) {
```
yakov116: `.commits blockquote:not(.rgh-parse-backticks)`
This is the push commits if clean dashbaord is off

fregante: I'd
1. Merge this info `initGeneral`'s array
2. Rename `initGeneral` to `parseAll` or something
3. Add `.map(selector => selector + ':not(.rgh-parse-backticks)') to that array
4. Actually add the class inside the loop

More items in that list are dynamic like this and may benefit from being called `onNewComments`

Alternatively you could have 2 separate functions:

- `parseStatic`: no need for `.rgh-parse-backticks` class
- `parseDynamic`: called by the newsfeed listener and onNewComments (not necessarily in this PR)

source/features/parse-backticks.tsx
```diff
@@ -15,18 +15,28 @@ function initGeneral(): void {
...
+
 // Highlight code in issue/PR titles on Dashboard page ("Recent activity" container)
 async function initDashboard(): Promise<void> {
 	await elementReady('.js-recent-activity-container', {stopOnDomReady: false});
 	for (const title of select.all('.js-recent-activity-container .text-bold')) {
 		parseBackticks(title);
 	}
+
+	await elementReady('.issues_labeled', {stopOnDomReady: false});
+	parseBackticksOnNewsFeedItems();
```
fregante: Instead of this and the merge with `infinite-scroll`, I'd use a MutationObserver to call `parseDynamic`.

Also keep in mind that `{stopOnDomReady: false}` is unsafe: in this case for example it will run `document.querySelector` 60 times a second, forever, if my feed doesn't have that `.issues_labeled`. Currently, _my feed doesn't have it._

source/libs/page-detect.ts
```diff
@@ -37,13 +37,14 @@ export const _isCommitList = [
 	'https://github.com/sindresorhus/refined-github/pull/148/commits'
 ];
 
-export const isRepoCommitList = (): boolean => /^commits\//.test(getRepoPath()!);
+export const isRepoCommitList = (): boolean => /^commits\//.test(getRepoPath()!) || (/^commits/.test(getRepoPath()!) && location.search.startsWith('?author='));
```
fregante: I think

```suggestion
export const isRepoCommitList = (): boolean => getRepoPath()?.startsWith('commit') ?? false
```

yakov116: The issue is the this will catch pull request comits

source/features/parse-backticks.tsx
yakov116: The issue is that we need to wait for it to load just like the other dashboard events

yakov116: I started doing this and I realized that this may be over my head. (Unless you want to guide me)

source/libs/page-detect.ts
fregante: Aren’t they under pull/?

source/features/parse-backticks.tsx
```diff
@@ -15,18 +15,28 @@ function initGeneral(): void {
 		'[id^=ref-issue-]', // Issue references in `isIssue`, `isPRConversation`
 		'[id^=ref-pullrequest-]', // PR references in `isIssue`, `isPRConversation`
 		'.TimelineItem-body > del, .TimelineItem-body > ins', // Title edits in `isIssue`, `isPRConversation`
-		'.js-pinned-issue-list-item > span' // Pinned Issues
+		'.js-pinned-issue-list-item > span', // Pinned Issues
+		'blame-commit-message a' // `isBlame`
 	])) {
```
fregante: I meant this
```suggestion
	].map(selector => selector + ':not(.rgh-parse-backticks)'))) {
		title.classList.add('rgh-parse-backticks');
```



- [ ] Merge this info initGeneral's array
- [ ] Rename initGeneral to parseAll or something
- [x] Add `.map(selector => selector + ':not(.rgh-parse-backticks)')` to that array
- [x] Actually add the class inside the loop

source/libs/page-detect.ts
yakov116: AVA Failed.

source/features/parse-backticks.tsx
yakov116: >Actually add the class inside the loop

What did you mean by that? Adding that will add a class to everything! did you mean t add `rgh-parse**d**-backticks`

source/libs/page-detect.ts
fregante: Darn it I meant `startsWith('commits')`, not `'commit'`

source/features/parse-backticks.tsx
fregante: [Here](https://github.com/sindresorhus/refined-github/pull/2947/commits/79e1dbbdeab966df45e3efbc5d4b5e4c8f1255d0#diff-7b0a2c8650978a85cebddd08a46e40ba) you have a selector with `a:not(.rgh-parse-commits)`

That never happens.

`rgh-parse-commits` exists exclusively on `code`, so if you want a loop like this ("select unparsed items, then parse them") you must add the "parsed" marker. _That's_ what I meant by _"Actually add the class"_

In this case I didn't notice that `rgh-parse-backticks` was already used by `code`, so I added it on the element that was being looped, as only **that** works.

---

Since `rgh-parse-backticks` is already used by `code`, yes, let's use a different class (just just parse**d**-backticks, it's too similar), but you still need to add it in this loop.

source/features/parse-backticks.tsx
```diff
@@ -15,8 +15,11 @@ function initGeneral(): void {
 		'[id^=ref-issue-]', // Issue references in `isIssue`, `isPRConversation`
 		'[id^=ref-pullrequest-]', // PR references in `isIssue`, `isPRConversation`
 		'.TimelineItem-body > del, .TimelineItem-body > ins', // Title edits in `isIssue`, `isPRConversation`
-		'.js-pinned-issue-list-item > span' // Pinned Issues
-	])) {
+		'.js-pinned-issue-list-item > span', // Pinned Issues
+		'blame-commit-message a', // `isBlame`
+		'.issues_labeled .text-gray-dark > a',
+		'.commits blockquote'
```
fregante: `//` missing

yakov116: I know I was waiting for you comment above

source/features/parse-backticks.tsx
yakov116: Got you. Will try later tonight

source/features/parse-backticks.tsx
yakov116: MutationObserver what am I observing? Why would we not have the same issue? It will never disconnect if there are no issues?

yakov116: MutationObserver what am I observing? Why would we not have the same issue? It will never disconnect if there are no issues?

yakov116:
@fregante i copied almost everything from hide-own-stars

source/features/parse-backticks.tsx
yakov116: Should be done.

Good night

source/features/parse-backticks.tsx
```diff
@@ -4,46 +4,68 @@ import select from 'select-dom';
...
 		features.isCommit,
 		features.isCommitList,
+		features.isCompare,
 		features.isDiscussionList,
 		features.isIssue,
 		features.isPRConversation,
+		features.isPulse,
+		features.isReleasesOrTags,
 		features.isRepoTree,
 		features.isSingleFile
```
fregante, line 65: It looks like a lot of pages could have this. As long as selectors are specific enough we can just run this on `isRepo`

```suggestion
		features.isRepo
```

yakov116: It wont break anything

yakov116:
@fregante should I remove the isPulse?

Also I added isActions but there is no such a thing as isActions yet. Is that ok?

fregante:
You might as well leave them

yakov116:
@fregante I am done (I think)

source/features/parse-backticks.tsx
```diff
@@ -4,46 +4,63 @@ import select from 'select-dom';
...
-		features.isRepoTree,
-		features.isSingleFile
+		features.isRepo
 	],
 	load: features.onAjaxedPages,
-	init: initGeneral
+	init: parseAll
 });
 
 features.add({
```
fregante, line 66: The listener already exists:

```js
init: () => onNewsfeedLoad(parseAll)
```

In a separate PR you might want to drop it from `hide-own-stars`

yakov116: Any other changes needws?

fregante: It doesn't look like it at the moment

fregante: The suggested code actually might not be enough since maybe it only covers the feed, not the "Recent Activity" box

yakov116: >The suggested code actually might not be enough since maybe it only covers the feed, not the "Recent Activity" box

But it would take away that MutationObserver
So make a new includes for it correct?

source/libs/page-detect.ts
```diff
@@ -33,16 +33,18 @@ export const _isCommitList = [
 	'https://github.com/sindresorhus/refined-github/commits/0.13.0',
 	'https://github.com/sindresorhus/refined-github/commits/230c2',
 	'https://github.com/sindresorhus/refined-github/commits/230c2935fc5aea9a681174ddbeba6255ca040d63',
-	'https://github.com/sindresorhus/refined-github/pull/148/commits'
+	'https://github.com/sindresorhus/refined-github/pull/148/commits',
+	'https://github.com/sindresorhus/refined-github/commits?author=fregante'
 ];
 
-export const isRepoCommitList = (): boolean => String(getRepoPath()).startsWith('commits/');
+export const isRepoCommitList = (): boolean => String(getRepoPath()).startsWith('commits');
```
yakov116, line 40: Take note I changed this

source/features/parse-backticks.tsx
fregante: And?

yakov116:
We still need this 	
'https://github.com/sindresorhus/refined-github/commits?author=fregante'

Add a commit for that?

source/features/parse-backticks.tsx
yakov116: Yes it does

fregante:
It’s unrelated. Send a new PR with page detect changes

PR #2948, opened 2020-03-27
Fix `hide-inactive-deployments` and "Load more..." detection

FloEdelmann:
GitHub changed the selector.

# Test
* https://github.com/OpenLightingProject/open-fixture-library/pull/425

### DIFF
--- a/source/features/hide-inactive-deployments.tsx
+++ b/source/features/hide-inactive-deployments.tsx
@@ -6,7 +6,7 @@ function init(): void {
 	deployments.pop(); // Don't hide the last deployment, even if it is inactive
 
 	for (const deployment of deployments) {
-		if (select.exists('.is-inactive', deployment)) {
+		if (select.exists('[title="Deployment Status Label: Inactive"]', deployment)) {
 			deployment.closest<HTMLElement>('.TimelineItem')!.hidden = true;
 		}
 	}
--- a/source/features/hide-inactive-deployments.tsx
+++ b/source/features/hide-inactive-deployments.tsx
@@ -8,7 +8,7 @@ function init(): void {
 
 	for (const deployment of deployments) {
 		if (select.exists('[title="Deployment Status Label: Inactive"]', deployment)) {
-			deployment.hidden = true;
+			deployment.remove();
 		}
 	}
 }
--- a/source/features/hide-inactive-deployments.tsx
+++ b/source/features/hide-inactive-deployments.tsx
@@ -2,15 +2,13 @@ import select from 'select-dom';
 import features from '../libs/features';
 
 function init(): void {
-	const deployments = select.all(`
-		.TimelineItem [title^="Deployment Status Label:"],
-		.js-socket-channel[data-url$="/pull_requests/events/deployed"] .TimelineItem a.btn
-	`);
+	// Selects all the deployments first so that we can leave the last one on the page
+	const deployments = select.all('.js-socket-channel[data-url$="/pull_requests/events/deployed"]');
 	deployments.pop(); // Don't hide the last deployment, even if it is inactive
 
 	for (const deployment of deployments) {
-		if (deployment.title === 'Deployment Status Label: Inactive') {
-			deployment.closest<HTMLElement>('.TimelineItem')!.hidden = true;
+		if (select.exists('[title="Deployment Status Label: Inactive"]', deployment)) {
+			deployment.hidden = true;
 		}
 	}
 }
--- a/source/features/hide-inactive-deployments.tsx
+++ b/source/features/hide-inactive-deployments.tsx
@@ -2,11 +2,14 @@ import select from 'select-dom';
 import features from '../libs/features';
 
 function init(): void {
-	const deployments = select.all('.TimelineItem .deployment-meta');
+	const deployments = select.all(`
+		.TimelineItem [title^="Deployment Status Label:"],
+		.js-socket-channel[data-url$="/pull_requests/events/deployed"] .TimelineItem a.btn
+	`);
 	deployments.pop(); // Don't hide the last deployment, even if it is inactive
 
 	for (const deployment of deployments) {
-		if (select.exists('[title="Deployment Status Label: Inactive"]', deployment)) {
+		if (deployment.title === 'Deployment Status Label: Inactive') {
 			deployment.closest<HTMLElement>('.TimelineItem')!.hidden = true;
 		}
 	}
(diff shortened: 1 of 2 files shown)


FloEdelmann:
Thanks, you're right.

Unfortunately, there is no longer one selector that matches inactive, pending, failed, and successful deployment events. Actually, the new selector for successful deployments is quite ugly, but we need it to hide the last inactive deployment before the active one.

fregante:
I’d put your new label selector in the `select.all` and then just use `closest` to reach the “timeline item”, instead of having a `select.exists` condition

FloEdelmann:
I'm confused. Isn't that exactly what I did in the last commit? There is no `select.exists` anymore.

source/features/hide-inactive-deployments.tsx
```diff
@@ -2,11 +2,14 @@ import select from 'select-dom';
 import features from '../libs/features';
 
 function init(): void {
-	const deployments = select.all('.TimelineItem .deployment-meta');
+	const deployments = select.all(`
+		.TimelineItem [title^="Deployment Status Label:"],
+		.js-socket-channel[data-url$="/pull_requests/events/deployed"] .TimelineItem a.btn
+	`);
```
fregante:
```suggestion
	const deployments = select.all(`.TimelineItem [title^="Deployment Status Label: Inactive"]`);
```

source/features/hide-inactive-deployments.tsx
```diff
@@ -2,11 +2,14 @@ import select from 'select-dom';
...
 	deployments.pop(); // Don't hide the last deployment, even if it is inactive
 
 	for (const deployment of deployments) {
-		if (select.exists('.is-inactive', deployment)) {
+		if (deployment.title === 'Deployment Status Label: Inactive') {
 			deployment.closest<HTMLElement>('.TimelineItem')!.hidden = true;
 		}
```
fregante:
```suggestion
		deployment.closest<HTMLElement>('.TimelineItem')!.hidden = true;
```

source/features/hide-inactive-deployments.tsx
fregante: I meant like this. Skip the condition altogether, just select what you need to delete.

fregante: Actually I think this is the only change required in this file, I forgot that we always select _all_ the deployments so that we can leave the last one on the page:

```suggestion
	// Selects all the deployments first so that we can leave the last one on the page
	const deployments = select.all('.js-socket-channel[data-url$="/pull_requests/events/deployed"']');
```

source/features/hide-inactive-deployments.tsx
```diff
@@ -2,11 +2,14 @@ import select from 'select-dom';
...
-	const deployments = select.all('.TimelineItem .deployment-meta');
+	const deployments = select.all(`
+		.TimelineItem [title^="Deployment Status Label:"],
+		.js-socket-channel[data-url$="/pull_requests/events/deployed"] .TimelineItem a.btn
+	`);
 	deployments.pop(); // Don't hide the last deployment, even if it is inactive
 
 	for (const deployment of deployments) {
-		if (select.exists('.is-inactive', deployment)) {
+		if (deployment.title === 'Deployment Status Label: Inactive') {
```
fregante: This should be reset to what it was
```suggestion
 		if (select.exists('.is-inactive', deployment)) {
```

FloEdelmann: It's actually a different selector (`[title="Deployment Status Label: Inactive"]`), as that was the change that triggered this PR :wink:

I get the idea though.

source/features/hide-inactive-deployments.tsx
```diff
@@ -2,12 +2,13 @@ import select from 'select-dom';
...
-	const deployments = select.all('.TimelineItem .deployment-meta');
+	// Selects all the deployments first so that we can leave the last one on the page
+	const deployments = select.all('.js-socket-channel[data-url$="/pull_requests/events/deployed"]');
 	deployments.pop(); // Don't hide the last deployment, even if it is inactive
 
 	for (const deployment of deployments) {
-		if (select.exists('.is-inactive', deployment)) {
-			deployment.closest<HTMLElement>('.TimelineItem')!.hidden = true;
+		if (select.exists('[title="Deployment Status Label: Inactive"]', deployment)) {
+			deployment.hidden = true;
```
fregante: Let's take this opportunity to remove it entirely

```suggestion
			deployment.remove();
```

source/libs/on-new-comments.ts
```diff
@@ -38,8 +38,10 @@ function addListeners(): void {
 		childList: true
 	});
 
-	// When hidden comments are loaded by clicking "Load more..."
-	delegates.add(delegate('.js-ajax-pagination', 'page:loaded', run));
+	// When hidden comments are loaded by clicking "Load more…"
+	delegates.add(delegate('.js-ajax-pagination', 'submit', event => {
+		event.target?.addEventListener('page:loaded', run, {once: true});
+	}));
```
fregante: Never generate a new function here because `delegate` won't deduplicate it.

Also, what does this mean? `event.target` is still `.js-ajax-pagination`, so why listen to `submit` first and then add a separate `page:loaded` listener later?

FloEdelmann: When clicking the "Load more…" button, the form is submitted and the XHR/pjax request is started. When the response arrives, the form is replaced with the new contents from the server (possibly containing another "Load more…" form. After that, the `page:loaded` event is triggered, but at that time the form is not part of the DOM tree anymore, i.e. it doesn't have a `parentNode`, so the event can't bubble up to the `document` for `delegate` to catch it.

source/libs/on-new-comments.ts
```diff
@@ -11,6 +11,13 @@ function run(): void {
 	handlers.forEach(async callback => callback());
 }
 
+function paginationSubmitHandler(event: DelegateEvent): void {
+	// When the 'page:loaded' event is triggered, the .js-ajax-pagination form won't have
+	// a parent, so the event doesn't bubble up. We delegate the 'submit' event instead
+	// and listen once for the 'page:loaded' event directly.
+	event.target?.addEventListener('page:loaded', run, {once: true});
+}
```
fregante, line 19:
```suggestion
// The form is detached just before the `page:loaded` event is triggered so the event won’t bubble up and `delegate` won’t catch it.
// This intermediate handler is required to catch the `page:loaded` event on the detached element.
function paginationSubmitHandler({delegateTarget: form}: DelegateEvent): void {
	form.addEventListener('page:loaded', run, {once: true});
}
```

FloEdelmann:
> I hope this works

Well, at least for me it does, while the current `master` branch doesn't.

PR #2950, opened 2020-03-28
Use `webext-base-css` to improve Option page’s style

fregante:
I'm extracting some common base styles into its own reusable module (https://github.com/fregante/webext-base-css) so I can easily import it into my other extensions and support Dark Mode everywhere.

I took the opportunity to also clean up the rest of the Options style

chrome
firefox


source/options.css
```diff
@@ -1,189 +1,69 @@
...
+	margin-top: 0;
 }
 
-input:not([type='checkbox']),
-textarea {
-	display: block;
-	box-sizing: border-box;
-	width: 100%;
+:root [name='customCSS'],
+:root [name='personalToken'] {
```
sindresorhus: Could you add a `.monotext` class or something to your package and then we could just mark the inputs in the HTML with such class?

source/options.css
```diff
@@ -1,189 +1,69 @@
...
-	color: var(--text-color);
-	border: 1px solid var(--input-border-color);
-}
-
-input:not([type='checkbox']):invalid {
-	color: var(--github-red) !important;
-	border: 1px solid;
+p {
+	margin-top: 0;
 }
```
sindresorhus: Shouldn't this be in your package too?

fregante:
Answering all your comments: I think I want `webext-base-css` to be an extension of the default browser stylesheet, not its own framework and opinionated reset. 

The `img vertical-align: middle` may belong there, custom margins maybe not, classes likely not (as tempting as `monotext` is)

PR #2963, opened 2020-04-01
Remove `copy-file` from Markdown view

### LINKED ISSUE #2957: `copy-file` appears on rendered markdown files (and fails)
There used to be a condition/events that made the Copy button appear only when showing the source code. 

1. Visit https://github.com/DrewML/GifHub/blob/master/LICENSE.md
2. Notice Copy button (it shouldn't be there)
3. Switch to see the source (it works now)

yakov116:
Fixes #2957 
Test on https://github.com/DrewML/GifHub/blob/master/LICENSE.md 
<!--

Thanks for contributing! 🍄

1. LINKED ISSUES:

2. TEST URLS:
   Add some test URLs

3. SCREENSHOT:

### DIFF
--- a/source/features/copy-file.tsx
+++ b/source/features/copy-file.tsx
@@ -42,7 +42,7 @@ function removeButton(): void {
 function init(): void {
 	delegate('.rgh-copy-file', 'click', handleClick);
 
-	if (select.exists('.blob.js-code-block-container')) {
+	if (select.exists('.blob > .markdown-body')) {
 		delegate('.rgh-md-source', 'rgh:view-markdown-source', renderButton);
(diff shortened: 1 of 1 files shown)


source/features/copy-file.tsx
```diff
@@ -42,7 +42,7 @@ function removeButton(): void {
 function init(): void {
 	delegate('.rgh-copy-file', 'click', handleClick);
 
-	if (select.exists('.blob.instapaper_body')) {
+	if (select.exists('.blob.js-code-block-container')) {
```
fregante, line 45: I'd prefer something more self-explanatory
```suggestion
	if (select.exists('.blob > .markdown-body')) {
```

PR #2964, opened 2020-04-01
Fix deployment cronjob

fregante:
Due to https://github.com/notlmn/browser-extension-template/issues/25

@sindresorhus can you generate and set a token in the secrets? E.g. `TAGGER_TOKEN`


sindresorhus:
Done

fregante:
Let's hope this works

Issue #2907, opened 2020-03-20
Console Errors


yakov116:
- [x] When clicking on almost any dropdown

```javascript
content.js:4477 Uncaught TypeError: Cannot read property 'getBoundingClientRect' of null
    at content.js:4477
```

Go to any PR start commenting and refresh the page
```javascript
content.js:808 Uncaught (in promise) TypeError: elements is not iterable
    at Function.watchAndFit [as watch] (content.js:808)
    at enable (content.js:7545)
    at fitPrCommitMessageBox (content.js:7553)
    at HTMLDocument.<anonymous> (content.js:5818)
watchAndFit @ content.js:808
enable @ content.js:7545
fitPrCommitMessageBox @ content.js:7553
(anonymous) @ content.js:5818
async function (async)
(anonymous) @ content.js:5817
so @ session-resume.esm.js:144
(anonymous) @ session-resume.ts:33
```
```javascript
content.js:8660 Uncaught (in promise) TypeError: Cannot read property 'dispatchEvent' of null
    at triggerValidation (content.js:8660)
    at HTMLDocument.<anonymous> (content.js:5818)
```

The first time you go to any repo

```javascript
content.js:12768 ❌ Refined GitHub: feature `repo-age`

 TypeError: Cannot read property 'src' of undefined
    at content.js:10313
    at async getSet (content.js:2412) 
```

- [x] Press <kbd>T</kbd> on any repo issue page.

```javascript
content.js:10791 Uncaught (in promise) TypeError: Cannot read property 'rghFileFinderTerm' of null
    at setValueInField (content.js:10791)
    at preserve_file_finder_term_init (content.js:10800)
    at features_run (content.js:2798)
    at content.js:2853
    at content.js:2742
```
- [x] When editing and creating a new milestone
```js
❌ Refined GitHub → filter-comments-by-you → TypeError: Cannot read property 'cloneNode' of null
    at filter_comments_by_you_init (refined-github.js:8380)
    at features_run (refined-github.js:5917)
    at refined-github.js:5972
    at refined-github.js:5861 
Search issue: https://github.com/sindresorhus/refined-github/issues?q=is%3Aissue+Cannot%20read%20property%20'cloneNode'%20of%20null

refined-github.js:4113 ❌ Refined GitHub → hide-issue-list-autocomplete → TypeError: Cannot read property 'setAttribute' of null
    at hide_issue_list_autocomplete_init (refined-github.js:9478)
[...]

PR #2965, opened 2020-04-02
Fix some page detection

fregante:
Should fix part of #2907, specifically https://github.com/sindresorhus/refined-github/issues/2907#issuecomment-607861547

Plus small related improvements

### DIFF
--- a/source/libs/page-detect.ts
+++ b/source/libs/page-detect.ts
@@ -248,6 +248,7 @@ export const _isRepo = [
 	'https://github.com/sindresorhus/notifications/',
 	'https://github.com/sindresorhus/refined-github/pull/148',
 	'https://github.com/sindresorhus/refined-github/milestones/new', // Gotcha for isRepoTaxonomyDiscussionList
+	'https://github.com/sindresorhus/refined-github/milestones/1/edit', // Gotcha for isRepoTaxonomyDiscussionList
 	'https://github.com/sindresorhus/refined-github/issues/new/choose', // Gotcha for isRepoIssueList
 	'https://github.com/sindresorhus/refined-github/issues/templates/edit' // Gotcha for isRepoIssueList
 ];
--- a/source/libs/page-detect.ts
+++ b/source/libs/page-detect.ts
@@ -222,7 +222,7 @@ export const _isQuickPR = [
 	'https://github.com/sindresorhus/refined-github/compare/test-branch?quick_pull=1'
 ];
 
-export const isReleasesOrTags = (): boolean /^tags$|^releases($|\/tag)/.test(getRepoPath()!);
+export const isReleasesOrTags = (): boolean => /^tags$|^releases($|\/tag)/.test(getRepoPath()!);
 export const _isReleasesOrTags = [
 	'https://github.com/sindresorhus/refined-github/releases',
 	'https://github.com/sindresorhus/refined-github/tags',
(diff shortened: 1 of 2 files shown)


source/libs/page-detect.ts
```diff
@@ -222,11 +222,7 @@ export const _isQuickPR = [
 	'https://github.com/sindresorhus/refined-github/compare/test-branch?quick_pull=1'
 ];
 
-export const isReleasesOrTags = (): boolean => {
-	const parts = (getRepoPath() ?? '').split('/');
-	return /^(releases|tags)$/.test(parts[0]) && parts[1] !== 'new';
-};
-
+export const isReleasesOrTags = (): boolean => /^tags$|^releases($|\/tag)/.test(getRepoPath()!);
```
sindresorhus, line 225:
```suggestion
export const isReleasesOrTags = (): boolean => ['tags', 'releases', 'releases/tag'].includes(getRepoPath()!);
```

fregante: `includes` is a direct match so it won't catch `releases/tags/1.2.3`

`.startsWith` also can't be used because `releases` should match but `releases/new` shouldn't

fregante:
@yakov116 seeing any issues with this PR on the milestones/templates pages?

source/libs/page-detect.ts
```diff
@@ -247,49 +242,47 @@ export const isRepo = (): boolean => /^[^/]+\/[^/]+/.test(getCleanPathname()) &&
...
 
-export const isRepoIssueList = (): boolean => {
-	const parts = (getRepoPath() ?? '').split('/');
-	return parts[0] === 'issues' && parts[1] !== 'new' && !/\d/.test(parts[1]); // `issues/fregante` is a list but `issues/1` isn't
-};
-
+// `issues/fregante` is a list but `issues/1`, `issues/new`, `issues/new/choose`, `issues/templates/edit` aren’t
+export const isRepoIssueList = (): boolean =>
+	String(getRepoPath()).startsWith('issues') &&
+	!/^issues\/(\d+|new|templates)($|\/)/.test(getRepoPath()!);
```
fregante, line 279: This line specifically should fix https://github.com/sindresorhus/refined-github/issues/2907#issuecomment-607861547

yakov116: It fixed new but not edit

source/libs/page-detect.ts
```diff
@@ -14,7 +13,7 @@ export const _is404 = domBased; // They're specified in page-detect.ts
 export const is500 = (): boolean => document.title === 'Server Error · GitHub' || document.title === 'Unicorn! · GitHub';
```
yakov116, line 13:
```suggestion
export const is500 = (): boolean => document.title === 'Server Error · GitHub' || document.title === 'Unicorn! · GitHub' || document.title === '504 Gateway Time-out';
```

Github just went down and that was the error

source/libs/page-detect.ts
fregante: What's edit? Always post a full link with bug reports

fregante: Do you mean `/sindresorhus/refined-github/issues/templates/edit`?

What do you mean by not fixed? `isRepoIssueList` doesn't match that page, so any feature that's supposed to run on `isRepoIssueList` won't run on that page.

This is tested (via _negative_ tests)

https://github.com/sindresorhus/refined-github/blob/641dbcbcccc0a0a04ed9ec1660a593ed23ff6749/source/libs/page-detect.ts#L252

yakov116: milestone edit

PR #2973, opened 2020-04-03
Fix examples from contributing.md

iamandrewluca:


contributing.md
```diff
@@ -40,16 +41,16 @@ Here's an example using all of the possible `feature.add` options:
...
 
-let delegate;
+let delegate: DelegateSubscription | undefined;
 function log() {
 	console.log('✨', <div className="rgh-jsx-element"/>);
 }
 function init(): void {
 	// Events must be set via delegate, unless shortlived
-	delegate = delegate('.btn', 'click', log);
+	delegate = delegateIt('.btn', 'click', log);
```
fregante, line 53: This was fine as `delegate`

iamandrewluca: it conflicted with `let delegate;`

fregante: Indeed. Better change that variable than this one then. It was just a random example

iamandrewluca: @fregante done!

fregante:
Mersi Andrew ☺️

PR #2976, opened 2020-04-05
Fix `deep-reblame` css selector

yakov116:
Is needs to be stronger.

When we made it originally we used `a.` but then we changed it to a button so we removed the `a`. Github has `link-hover-blue` we have 2 options, remove Githubs or change our selector. 

<!--

Thanks for contributing! 🍄

1. LINKED ISSUES:

2. TEST URLS:
   Add some test URLs

3. SCREENSHOT:


source/features/deep-reblame.css
```diff
@@ -2,6 +2,6 @@ button.rgh-deep-reblame .octicon-versions {
 	color: #e40de48c;
 }
 
-.rgh-deep-reblame:hover {
-	color: #8000808c;
+[aria-label*='PR first'].rgh-deep-reblame:hover {
+	color: #8000808c !important;
```
fregante, line 6: The rule above applies it to the icon itself. A parent can never override the children.

```suggestion
.rgh-deep-reblame:hover .octicon-versions {
	color: #8000808c;
```

Now we can also drop `!important` since the icon's color doesn't have have `!important`

source/features/deep-reblame.css
```diff
@@ -2,6 +2,6 @@ button.rgh-deep-reblame .octicon-versions {
 	color: #e40de48c;
```
fregante, line 2: Test

fregante, line 2: Test review

source/features/deep-reblame.css
fregante: Sorry I'm testing something. Duplicate review follows:

The rule above applies it to the icon itself. A parent can never override the children.

```suggestion
.rgh-deep-reblame:hover .octicon-versions {
	color: #8000808c;
```

Now we can also drop `!important` since the icon's color doesn't have have `!important`

yakov116: >The rule above applies it to the icon itself. A parent can never override the children.
👍

fregante:
Missing testing links as always...

PR #2980, opened 2020-04-06
Adds back New Pull Request button.

kenhkelly:
The New Pull Request button was removed in #1876 in favor of the buttons. Because the buttons no longer exist anymore, this PR adds the New Pull Request button back. 

Before: 
image

After:
image

### DIFF
--- a/source/refined-github.css
+++ b/source/refined-github.css
@@ -43,11 +43,6 @@
 	margin-left: 0;
 }
 
-/* Hide `New pull request button` near file list */
-.file-navigation .new-pull-request-btn {
-	display: none;
-}
-
 @keyframes fade-in {
 	from {
 		opacity: 0;


fregante:
>  Because the buttons no longer exist anymore

They sure exist, they're called `default-branch-button` and `latest-tag-button`

The link still exists as More -> Compare

kenhkelly:
Hm, I've been using this for a little bit now and I'm not seeing anything in that spot, or anything matched those setting screenshots, even on this project.

This change came about because some developers on my team have also adopted it and wondered where the button went. Since I noticed the files in the PR I referenced were gone, I thought it would be ok to put back.

Under the More -> Compare is fine, but it's still more clicks than we like, and the spirit of this project is to remove the annoyances of Github.

image

fregante:
I think the issue wasn't "lack of space" as much as it being duplicate or "useless":

- On master, it points to an empty page
- On other branches, there's already a "Pull request" button on the right, making it a duplicate button.

duplicate

kenhkelly:
Ok, that makes sense. I will close the PR then.

PR #2987, opened 2020-04-09
Improve feature loading mechanism

fregante:
New attempt at solving this: https://github.com/sindresorhus/refined-github/pull/2925#issuecomment-608541758

WIP: I'll have to review this code again and do a couple of new TODOs

## Why

- Lets you add any number of listeners for AJAXed widget, like `onNewComments`, `onNewsfeedLoad`, etc.
- The listeners themselves can be simplified to avoid having to manage `onAjaxedPages` internally.
- Easier to understand (_what did `onAjaxedPages` mean anyway?_)

## Changes

- By default, run on DOM ready and AJAX pages.
- You can customize this behavior with 3 separate options:

	- `waitForDomReady: true`: ["fast features"](https://github.com/sindresorhus/refined-github/issues/2671) will have to set this to false.
	- `repeatOnAjax: true`:  It's the most common configuration anyway. It only **_has to_** be set to `false` in rare occasions.
	- `repeatOnAjaxEvenOnBackButton: false`: this the equivalent of the old `onAjaxedPagesRaw`. Only used by 3 features.

- Pass additional listeners via `additionalListeners: []` property. These listeners will be run _after_ each page load, whether this is "document start", DOM ready or AJAX.


fregante:
@sindresorhus @yakov116 it'd be great if you could use this for a few days. It looks like it's working for me, but I haven't change the features that were "updated" the most.

yakov116:
@fregante I have very limited access to a computer over the next week but I will try it and let you know how it goes.

yakov116:
>I haven't specifically tested the features that were "updated" the most.

@fregante what things specifically should i test?

fregante:
> what things specifically should i test?

Any feature that was updated beyond just the drop of the `load` property. 

Example of a feature to test:


test


Example of a feature to skip:


skip

yakov116:
I am having issues with #2992 and this. I am not sure is #2992 is the issue or this PR is. Testing and will let you know.

yakov116:
#2992 is the issue

Issue #2564, opened 2019-11-22
Clear PR merge commit description field


fregante:
On PRs, when squashing, the commit description is filled with every commit's title.

I think that if you squash, probably aren't interested in _every commit's title_ and should therefore be dropped. If they're useful, maybe you should merge or rebase instead, IMHO.

Thoughts?

Perhaps we should just preserve a deduplicated list of `Co-authored-by`

# Currently

<img width="757" alt="" src="https://user-images.githubusercontent.com/1402241/69406614-0a922880-0d35-11ea-9262-1e21e731881c.png">

# Expected

<img width="756" alt="" src="https://user-images.githubusercontent.com/1402241/69406863-acb21080-0d35-11ea-8449-33c1c9c49081.png">

sindresorhus:
I sometimes find it useful to be able to quickly scan what commits are included in the squash and I do think we should preserve a deduplicated list of co-authors. How about we present a list of commit titles underneath or something? So I can still scan it, but it's not included by default.

Issue #2579, opened 2019-11-30
PR review "final report box" to show before merging


fregante:
When about to merge PRs:

<img width="756" alt="Screenshot 2019-11-30 at 16 44 43" src="https://user-images.githubusercontent.com/1402241/69898814-bddfca80-1390-11ea-885d-322c479f4336.png">

we could show a final "report" box listing what's about to happen:

- full list of commits that will be merged/squashed/rebased, as suggested in https://github.com/sindresorhus/refined-github/issues/2564#issuecomment-557433237
- issues/PRs that will automatically be closed (via `closes #11` keyword) or if none will be closed
- any potential issues to be aware of:
  - unrebased `!squash`/`!fixup` commits: https://github.com/sindresorhus/refined-github/issues/2385
  - reviews that haven't been resolved
  - CI that hasn't completed or passed
- TODOs added by the PR https://github.com/sindresorhus/refined-github/issues/2634

What else could this include?

fregante:
> full list of commits that will be merged/squashed/rebased

This would be a good start for this feature, coupled with the easier https://github.com/sindresorhus/refined-github/issues/2564

fregante:
> * issues/PRs that will automatically be closed (via `closes #11` keyword) or if none will be closed

Actually I'd like to see this more. Sometimes new users forget to mark "closes X" correctly so it'd be good to see "No issues will be closed when merging this PR" as well as "X Y and Z will be closed after merging.

rkrisztian:
I need the warnings or even merge checks for `!squash` and `!fixup` commits the most. I'm concerned that https://github.com/13rac1/block-fixup-merge-action still [does not support merge queues](https://github.com/13rac1/block-fixup-merge-action/issues/31), already has [deprecation warnings](https://github.com/13rac1/block-fixup-merge-action/issues/30) and it has not received updates since 2021.

cooljeanius:
> I'm concerned that [13rac1/block-fixup-merge-action](https://github.com/13rac1/block-fixup-merge-action) still [does not support merge queues](https://github.com/13rac1/block-fixup-merge-action/issues/31)

I mean, support for merge queues isn't really that widespread in general; see also #6850

fregante:
Closing in favor of:

- https://github.com/refined-github/refined-github/issues/7894

After 5 years, this issue isn't any closer to being done. Consider other requests rejected for now.

PR #3003, opened 2020-04-14
Add `clear-pr-merge-commit-message` feature

fregante:
Closes https://github.com/sindresorhus/refined-github/issues/2564

screen

## Test

Find a PR:
- you can merge
- has some Co-Authors (or doesn't)


yakov116:
>How about we present a list of commit titles underneath or something? So I can still scan it, but it's not included by default.

I am nervous that not having the commits at all will cause some backlash

fregante:
https://github.com/sindresorhus/refined-github/issues/2579 would help

PR #3011, opened 2020-04-16
Add `go-to-action-from-file` feature

### LINKED ISSUE #3008: From a GitHub Action workflow file add a button to see the workflow in action
Thanks for refined-github!

Could a button help going from e.g. https://github.com/r-hub/docs/blob/master/.github/workflows/pr.yml to https://github.com/r-hub/docs/actions?query=workflow%3ACommands using the [workflow name](https://github.com/r-hub/docs/blob/b57f8a2001ef48270ae5ce3d039ea049013a982f/.github/workflows/pr.yml#L5)?

This way when trying to assess the use of an action, one would use GitHub search, then find workflow files, then from the files easily see logs of the action.

maelle:
Fix #3008 

https://github.com/r-hub/docs/blob/master/.github/workflows/pr.yml
https://github.com/r-hub/docs/blob/master/.github/workflows/urlchecks.yml (not working yet, not sure why the URL encoding fails)

image

### DIFF
--- a/source/features/gha-runs-from-workflow-file.tsx
+++ b/source/features/gha-runs-from-workflow-file.tsx
@@ -1,32 +1,17 @@
-import React from 'dom-chef';
-import select from 'select-dom';
+//import React from 'dom-chef';
+//import select from 'select-dom';
 import features from '../libs/features';
 
 const isWorkflowFile = (): boolean => (location.pathname.endsWith('.yaml') && location.pathname.split('/', 7)[5] == ".github"  && location.pathname.split('/', 7)[6] == "workflows");
 
-	console.log(location.pathname)
 function init(): void {
-  const actionName =
+  const actionName = 
   	document.body('.blob-code-inner')
   		.find(line => line.textContent.startsWith('name'))
   		.textContent
   		.replace(/^name:\s+/, '')
   		.trim();
-	const runButton = select<HTMLAnchorElement>('#workflow-in-action')!;
-	const link = runButton.pathname.split('/');
-	// Remove /raw/
-	link.splice(3, 1);
 	console.log(actionName)
-	runButton
-		.parentElement! // `BtnGroup`
-		.prepend(
-			<a
-				className="btn btn-sm BtnGroup-item"
-				href={`https://github.com/r-hub/docs/actions?query={actionName}`}
-			>
-				Workflow in Action
-			</a>
-		);
 }
 
 features.add({
@@ -35,7 +20,8 @@ features.add({
 	screenshot: 'https://user-images.githubusercontent.com/44045911/67634792-48995980-f8fb-11e9-8b6a-7b57d5b12a2f.png'
 }, {
 	include: [
-//		isWorkflowFile
+	  features.isSingleFile,
+	  isWorkflowFile
 	],
 	exclude: [
 		features.isEnterprise
--- a/source/features/gha-runs-from-workflow-file.tsx
+++ b/source/features/gha-runs-from-workflow-file.tsx
@@ -19,8 +19,6 @@ function init(): void {
 	
 	const actionURL = `https://github.com/${link[1]}/${link[2]}/actions?query=workflow%3A${encodeURIComponent(actionName)}`;
 	
-	console.log(actionURL);
-	
 	rawButton
 		.parentElement! // `BtnGroup`
 		.prepend(
--- a/source/features/gha-runs-from-workflow-file.tsx
+++ b/source/features/gha-runs-from-workflow-file.tsx
@@ -17,7 +17,8 @@ function init(): void {
 	const rawButton = select<HTMLAnchorElement>('#raw-url')!;
 	const link = location.pathname.split('/', 7);
 	
-	const actionURL = `https://github.com/${link[1]}/${link[2]}/actions?query=workflow%3A"${actionName}"`;
+	const actionURL = new URL(`${location.origin}/${getRepoURL()}/actions`);
+	actionURL.searchParams.set('query', `workflow:"${actionName}"`)
 
 	rawButton
 		.parentElement! // `BtnGroup`
--- a/source/features/gha-runs-from-workflow-file.tsx
+++ b/source/features/gha-runs-from-workflow-file.tsx
@@ -37,7 +37,6 @@ features.add({
 	screenshot: 'https://user-images.githubusercontent.com/44045911/67634792-48995980-f8fb-11e9-8b6a-7b57d5b12a2f.png'
 }, {
 	include: [
-	  features.isSingleFile,
 	  isWorkflowFile
 	],
 	exclude: [
--- a/source/features/gha-runs-from-workflow-file.tsx
+++ b/source/features/gha-runs-from-workflow-file.tsx
@@ -2,7 +2,7 @@ import React from 'dom-chef';
 import select from 'select-dom';
 import features from '../libs/features';
 
-const isWorkflowFile = (): boolean => ( ((location.pathname.endsWith('.yaml')) || (location.pathname.endsWith('.yml'))) && location.pathname.split('/', 7)[5] == ".github"  && location.pathname.split('/', 7)[6] == "workflows");
+const isWorkflowFile = (): boolean => /\/\.github\/workflows\/.+\.ya?ml$/.test(getRepoPath()!);
 
 function init(): void {
   
--- a/source/features/gha-runs-from-workflow-file.tsx
+++ b/source/features/gha-runs-from-workflow-file.tsx
@@ -1,16 +1,17 @@
 //import React from 'dom-chef';
-//import select from 'select-dom';
+import select from 'select-dom';
 import features from '../libs/features';
 
 const isWorkflowFile = (): boolean => (location.pathname.endsWith('.yaml') && location.pathname.split('/', 7)[5] == ".github"  && location.pathname.split('/', 7)[6] == "workflows");
 
 function init(): void {
+  
   const actionName = 
-  	document.body('.blob-code-inner')
-  		.find(line => line.textContent.startsWith('name'))
-  		.textContent
-  		.replace(/^name:\s+/, '')
-  		.trim();
+	select.all('.blob-code-inner')
+		.find(line => line.textContent.startsWith('name'))
+		.textContent
+		.replace(/^name:\s+/, '')
+		.trim();
 	console.log(actionName)
 }
 
--- a/source/features/gha-runs-from-workflow-file.tsx
+++ b/source/features/gha-runs-from-workflow-file.tsx
@@ -1,4 +1,4 @@
-//import React from 'dom-chef';
+import React from 'dom-chef';
 import select from 'select-dom';
 import features from '../libs/features';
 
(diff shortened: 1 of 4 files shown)


maelle:
Actually it fails on other pages, that shouldn't include it, on it, sorry.

maelle:
Now I think it's ok.

source/features/gha-runs-from-workflow-file.tsx
```diff
@@ -0,0 +1,46 @@
...
+			>
+				See runs
+			</a>
+		);
+}
+
+features.add({
+	id: __featureName__,
+	description: 'Adds a link to view Action workflow in action.',
+	screenshot: 'https://user-images.githubusercontent.com/44045911/67634792-48995980-f8fb-11e9-8b6a-7b57d5b12a2f.png'
```
maelle, line 37: I haven't added the screenshot yet, will do after your review, since I suspect the button name can change.

source/features/gha-runs-from-workflow-file.tsx
```diff
@@ -0,0 +1,46 @@
+import React from 'dom-chef';
+import select from 'select-dom';
+import features from '../libs/features';
+
+const isWorkflowFile = (): boolean => ( ((location.pathname.endsWith('.yaml')) || (location.pathname.endsWith('.yml'))) && location.pathname.split('/', 7)[5] == ".github"  && location.pathname.split('/', 7)[6] == "workflows");
```
sindresorhus, line 5: This check looks very fragile. It should also be moved to https://github.com/sindresorhus/refined-github/blob/master/source/libs/page-detect.ts and tested.

source/features/gha-runs-from-workflow-file.tsx
```diff
@@ -0,0 +1,46 @@
...
+		.find(line => line.textContent.startsWith('name'))
+		.textContent
+		.replace(/^name:\s+/, '')
+		.replace('"', '')
+		.replace("'", '')
+		.trim();
+	const rawButton = select<HTMLAnchorElement>('#raw-url')!;
+	const link = location.pathname.split('/', 7);
+	
+	const actionURL = `https://github.com/${link[1]}/${link[2]}/actions?query=workflow%3A"${actionName}"`;
```
sindresorhus, line 20: Use `URLSearchParams` to construct query string.

source/features/gha-runs-from-workflow-file.tsx
```diff
@@ -0,0 +1,46 @@
...
+features.add({
+	id: __featureName__,
+	description: 'Adds a link to view Action workflow in action.',
+	screenshot: 'https://user-images.githubusercontent.com/44045911/67634792-48995980-f8fb-11e9-8b6a-7b57d5b12a2f.png'
+}, {
+	include: [
+	  isWorkflowFile
+	],
+	exclude: [
+		features.isEnterprise
```
sindresorhus, line 43: Why?

source/features/gha-runs-from-workflow-file.tsx
```diff
@@ -0,0 +1,46 @@
...
+				href={actionURL}
+			>
+				See runs
+			</a>
+		);
+}
+
+features.add({
+	id: __featureName__,
+	description: 'Adds a link to view Action workflow in action.',
```
sindresorhus, line 36: This description could be improved.

sindresorhus:
Make sure to run `npm test` locally and that is passes.

maelle:
Short question (sorry!): `select.all('.blob-code-inner')` is apparently not a good idea (error in tests) but I'm not sure what it should actually be? I didn't understand `$$('.blob-code-inner')`. I tried finding similar examples but failed at that.

source/features/gha-runs-from-workflow-file.tsx
fregante: Previously I didn't add tests to page-detect when they seemed too specific to be used anywhere else, like `isSingleHTMLFile`.

As for this test, probably something like:
```suggestion
const isWorkflowFile = (): boolean => /\/\.github\/workflows\/.+\.ya?ml$/.test(getRepoPath()!);
```

The problem is that URL-based tests aren't safe in this case (because of branches with slashes), so if you want it to be safe (i.e. DOM-based) it can't be tested `page-detect` either.

maelle:
https://github.com/sindresorhus/refined-github/blob/bf2a893741883572e7eb20c5f6d57d4ad43d6991/source/features/show-whitespace.tsx#L67

Uses async 🤔

source/features/gha-runs-from-workflow-file.tsx
maelle: I forgot to remove this, will do.

source/features/gha-runs-from-workflow-file.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+function init(): void {
+  
+  const actionName = 
+	select.all('.blob-code-inner')
+		.find(line => line.textContent.startsWith('name'))
+		.textContent
+		.replace(/^name:\s+/, '')
+		.replace('"', '')
+		.replace("'", '')
```
fregante, line 15: `.replace` with a string will only replace the first occurrence; you'll need a regex with the `g` flag
```suggestion
		.replace(/["']/g, '')
```

source/features/gha-runs-from-workflow-file.tsx
```diff
@@ -0,0 +1,46 @@
+import React from 'dom-chef';
+import select from 'select-dom';
+import features from '../libs/features';
+
+const isWorkflowFile = (): boolean => ( ((location.pathname.endsWith('.yaml')) || (location.pathname.endsWith('.yml'))) && location.pathname.split('/', 7)[5] == ".github"  && location.pathname.split('/', 7)[6] == "workflows");
+
+function init(): void {
+  
+  const actionName =
```
fregante, line 9: Run `npx xo --fix` locally to fix errors like on this line.

source/features/gha-runs-from-workflow-file.tsx
maelle: Should it be made DOM-based and if so what would be a good existing example in the project?

source/features/gha-runs-from-workflow-file.tsx
fregante: Use relative URLs or `location.origin` so it works on GitHub Enterprise:

```suggestion
	const actionURL = new URL(`${location.origin}/${getRepoURL()}/actions`);
	actionURL.searchParams.set('query', `workflow:"${actionName}"`)
```

I think this is the right query? Not sure

source/refined-github.ts
```diff
@@ -171,6 +171,7 @@ import './features/show-open-prs-of-forks';
 import './features/clone-branch';
 import './features/deep-reblame';
 import './features/clear-pr-merge-commit-message';
+import './features/gha-runs-from-workflow-file';
```
fregante, line 174: Maybe:

```suggestion
import './features/go-to-action-from-file';
```

source/features/gha-runs-from-workflow-file.tsx
```diff
@@ -0,0 +1,46 @@
...
+		.replace(/^name:\s+/, '')
+		.replace('"', '')
+		.replace("'", '')
+		.trim();
+	const rawButton = select<HTMLAnchorElement>('#raw-url')!;
+	const link = location.pathname.split('/', 7);
+	
+	const actionURL = `https://github.com/${link[1]}/${link[2]}/actions?query=workflow%3A"${actionName}"`;
+
+	rawButton
```
fregante, line 22: The variable isn't needed since it's only used once. Same for the `<HTMLAnchorElement>`type
```suggestion
	select('#raw-url')!
```

fregante:
> `select.all('.blob-code-inner')` is apparently not a good idea (error in tests) but I'm not sure what it should actually be? I didn't understand `$$('.blob-code-inner')`. I tried finding similar examples but failed at that.

`$$` is equivalent to `select.all`, it's a [console helper](https://developer.mozilla.org/en-US/docs/Tools/Web_Console/Helpers)

The linter is probably just complaining about the whitespace, to be fixed with `npx xo --fix`

source/features/gha-runs-from-workflow-file.tsx
```diff
@@ -0,0 +1,46 @@
...
+import select from 'select-dom';
+import features from '../libs/features';
+
+const isWorkflowFile = (): boolean => ( ((location.pathname.endsWith('.yaml')) || (location.pathname.endsWith('.yml'))) && location.pathname.split('/', 7)[5] == ".github"  && location.pathname.split('/', 7)[6] == "workflows");
+
+function init(): void {
+  
+  const actionName = 
+	select.all('.blob-code-inner')
+		.find(line => line.textContent.startsWith('name'))
```
fregante, line 11:
```suggestion
		.find(line => line.textContent!.startsWith('name'))
```

maelle:
Thanks both for your reviews!

I can explore more tomorrow, at the moment the feature works but there are still errors I need to fix

```js
ERROR in /home/maelle/Documents/misc_projects/refined-github/source/features/go-to-action-from-file.tsx(10,2):
TS2531: Object is possibly 'null'.
ERROR in /home/maelle/Documents/misc_projects/refined-github/source/features/go-to-action-from-file.tsx(10,2):
TS2532: Object is possibly 'undefined'.
ERROR in /home/maelle/Documents/misc_projects/refined-github/source/features/go-to-action-from-file.tsx(25,5):
TS2322: Type 'URL' is not assignable to type 'string'.

```

source/features/gha-runs-from-workflow-file.tsx
maelle: I removed it.

source/features/gha-runs-from-workflow-file.tsx
maelle: I tried to make it clearer.

source/features/go-to-action-from-file.tsx
```diff
@@ -0,0 +1,37 @@
+import React from 'dom-chef';
+import select from 'select-dom';
+import features from '../libs/features';
+import {getRepoURL, getRepoPath} from '../libs/utils';
+
+const isWorkflowFile = (): boolean => /\/\.github\/workflows\/.+\.ya?ml$/.test(getRepoPath()!);
+
+function init(): void {
+	const actionName = select.all('.blob-code-inner')
+		.find(line => line.textContent!.startsWith('name'))!
```
fregante, line 10: This might be missing, so you’ll need `return false`

maelle: With `if`?

maelle:
Yay, thanks!

I like the idea of using an icon but the "play" one would make me think I can _launch_ a run. 🤔

maelle:
Keywords that might help or not: "logs", "previous runs", "past runs".

fregante:
I like Past runs, even if not technically correct (some runs might still be in progress)

maelle:
Did we try updating it at the same time? 🤔

fregante:
That’s not possible with git

maelle:
Right, just me being useless from a mobile device. I'll edit from my laptop once "Past runs" is the confirmed choice. :-)

maelle:
I changed the name, are further changes needed?

fregante:
1. Use "Past runs"
2. Update screenshot
2. Update readme (add it to Repositories, I think)

maelle:
I'm getting an error, I'm guessing I'll need to rebase my PR.

maelle:
(on it)

maelle:
Done. :slightly_smiling_face:

PR #3019, opened 2020-04-19
Use Rollup instead of Webpack

fregante:
Our Webpack config is huge and the output is ugly (and not optimized: https://github.com/webpack/webpack/issues/10720), so I've been looking at alternatives.

Sadly Rollup is too slow (https://github.com/rollup/rollup/issues/3506) and it requires extra code/config to match Webpack-as-an-app-builder capabilities (e.g. copying asset files effortlessly).

This is always the problem. Rollup output is always 💯 but it's limited/limiting.

This can be revisited once there's a way to parallelize the build and sync assets effortlessly.


fregante:
Related: Parcel PR https://github.com/sindresorhus/refined-github/pull/2983

PR #3021, opened 2020-04-20
Add `open-and-closed-issues-sort-by-update-time`

maricn:
This is fork of `sort-issues-by-update-time` which displays both open and closed issues.

The reason I made this is because I often search for an issue in some repo, and I like it being sorted by `Recently updated`. I am not interested whether it's open or closed, I just want more info on the issue.
However the `sort-issues-by-update-time` feature keeps the search query like: `is:issue is:open sort:updated-desc` and it makes it inconvenient to update - I have to select the innermost filter is:open, remove it, and then type my search term. In case I want to change the term, I need to repeat the process, or move the search term to be the last parameter.

I'm open for suggestions regarding naming. Also, this could just be an update on existing `sort-issues-by-update-time` if people think that's better. And in case you think this functionality is not needed, I'll keep it for myself.

1. LINKED ISSUES:
   No existing issues.

2. TEST URLS:
   https://github.com/sindresorhus/refined-github (check the link to Issues)

3. SCREENSHOT:
   
desktop-2020-04-20T17:47:26+0200

### DIFF
--- a/source/features/open-and-closed-issues-sort-by-update-time.tsx
+++ b/source/features/open-and-closed-issues-sort-by-update-time.tsx
@@ -1,0 +1,38 @@
+import select from 'select-dom';
+import elementReady from 'element-ready';
+import features from '../libs/features';
+import SearchQuery from '../libs/search-query';
+
+function init(): void {
+	// Get issues links that don't already have a specific sorting applied
+	for (const link of select.all<HTMLAnchorElement>(`
+		[href*="/issues"]:not([href*="sort%3A"]):not(.issues-reset-query),
+		[href*="/pulls" ]:not([href*="sort%3A"]):not(.issues-reset-query)
+	`)) {
+		// Pick only links to lists, not single issues
+		// + skip pagination links
+		// + skip pr/issue filter dropdowns (some are lazyloaded)
+		if (/(issues|pulls)\/?$/.test(link.pathname) && !link.closest('.pagination, .table-list-filters')) {
+			new SearchQuery(link).add('is:closed sort:updated-desc');
+		}
+	}
+}
+
+async function cleanBar(): Promise<void> {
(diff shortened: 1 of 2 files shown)


fregante:
These are two unrelated features merged into one: 

- one sorts (already exists)
- one shows both open and closed issues at the same time

If you want the second part, you can use this extension: https://github.com/sindresorhus/show-all-github-issues

fregante:
You might also like `extend-discussion-status-filters`

maricn:
@fregante

> If you want the second part, you can use this extension: https://github.com/sindresorhus/show-all-github-issues

Thanks for sharing that other extension, but I just noticed that is not exactly the behavior I'm looking for.

What I'm interested in is both open and closed issues at the same time. I am already using `extend-discussion-status-filters`, but that requires me an additional click+page load to get to the state where I can browse open+closed issues together.

And the extension you shared is Chrome only, I'm on Firefox.


> These are two unrelated features merged into one:
> 
>     * one sorts (already exists)
> 
>     * one shows both open and closed issues at the same time

True, this is two different features merged into one, and I could refactor them to play together nicely as well as in isolation.

fregante:
> I just noticed that is not exactly the behavior I'm looking for.

Indeed, my bad.

However this is not a change I'd add to Refined GitHub. Closed issues are considered "resolved" and uninteresting for most, they shouldn't clutter the view _by default_

maricn:
> > I just noticed that is not exactly the behavior I'm looking for.
> 
> Indeed, my bad.
> 
> However this is not a change I'd add to Refined GitHub. Closed issues are considered "resolved" and uninteresting for most, they shouldn't clutter the view _by default_

Closed issues are not interesting for project developers, but for project users who want to find solutions to their problems perhaps. That is my use case anyways.

And also, in that case, the feature should exclude applying itself to PRs.

PR #3034, opened 2020-04-25
Meta: Use JSX when generating elements with functions

fregante:
Testing/implementing https://github.com/vadimdemedes/dom-chef/pull/59

### DIFF
--- a/source/features/bugs-tab.tsx
+++ b/source/features/bugs-tab.tsx
@@ -1,6 +1,7 @@
+import React from 'dom-chef';
 import cache from 'webext-storage-cache';
 import select from 'select-dom';
-import bugIcon from '@primer/octicons/build/svg/bug.svg';
+import BugIcon from '@primer/octicons/build/svg/bug.svg';
 import elementReady from 'element-ready';
 import features from '../libs/features';
 import * as pageDetect from '../libs/page-detect';
@@ -53,7 +54,7 @@ async function init(): Promise<void | false> {
 	const bugsTab = issuesTab.cloneNode(true);
 
 	// Update its appearance
-	select('.octicon', bugsTab)!.replaceWith(bugIcon());
+	select('.octicon', bugsTab)!.replaceWith(<BugIcon/>);
 	select('[itemprop="name"]', bugsTab)!.textContent = 'Bugs';
 
 	// Set temporary counter
--- a/source/features/clone-branch.tsx
+++ b/source/features/clone-branch.tsx
@@ -6,7 +6,7 @@ import * as textFieldEdit from 'text-field-edit';
 import * as api from '../libs/api';
 import features from '../libs/features';
 import * as pageDetect from '../libs/page-detect';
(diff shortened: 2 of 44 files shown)


source/features/clone-branch.tsx
```diff
@@ -44,8 +44,8 @@ async function cloneBranch(event: delegate.Event<MouseEvent, HTMLButtonElement>)
 	const currentBranch = getBranchBaseSha(branchName);
 	let newBranchName = prompt('Enter the new branch name')?.trim();
 
-	const spinner = loadingIcon();
-	spinner.classList.add('ml-2');
+	const spinner = <LoadingIcon className="ml-2"/>;
```
fregante, line 47: Advantage: get icon and add classes in one go.

source/features/pr-filters.tsx
```diff
@@ -27,17 +27,14 @@ function addDropdownItem(dropdown: HTMLElement, title: string, filterCategory: s
...
-
 	dropdown.append(
 		<a
 			href={`?${String(search)}`}
 			className="SelectMenu-item"
 			aria-checked={isSelected ? 'true' : 'false'}
 			role="menuitemradio"
 		>
-			{icon}
+			<CheckIcon className="SelectMenu-icon SelectMenu-icon--check"/>
```
fregante, line 37: Advantage: get icon and add classes in one go (bonus: skip `icon` variable altogether)

source/globals.d.ts
```diff
@@ -70,11 +72,10 @@ interface HTMLFormControlsCollection {
 }
 
 declare module '*.svg' {
-	export default (): SVGElement => SVGElement;
+	export default (): JSX.Element => JSX.Element;
```
fregante, line 75: Sad point: you can't use regular DOM-returning functions as `<Fun>`, they must return `JSX.Element`, which only exists the DOM is already generated with JSX (or its overridden this way)

source/features/bugs-tab.tsx
```diff
@@ -53,7 +54,7 @@ async function init(): Promise<void | false> {
 	const bugsTab = issuesTab.cloneNode(true);
 
 	// Update its appearance
-	select('.octicon', bugsTab)!.replaceWith(bugIcon());
+	select('.octicon', bugsTab)!.replaceWith(<BugIcon/>);
```
fregante, line 57: Downside: if you want consistency (always `BugIcon`, never `bugIcon`), your DOM-returning functions have to use JSX whether you need it or not, or else you'll get:


>  ✖   44:17  A function with a name starting with an uppercase letter should only be used as a constructor.  new-cap

Using JSX silences this error.

Also this means you'll have to `import React from 'dom-chef'` where you normally don't need it.

fregante: I made changes in https://github.com/vadimdemedes/dom-chef/pull/60 so that plain `<Function/>` calls are as optimized as possible.

yakov116:
Before this PR
image

After

image

Issue #2960, opened 2020-03-31
"Can you add this feature?"


fregante:
I'll use this issue to write down what I think belongs to Refined GitHub and what doesn't

> Can you add this feature?

Maybe, if:

1. It doesn't already exist as a [userscript](https://github.com/Mottie/GitHub-userscripts) or extension.
1. It makes sense as a default for everyone (so no "disabled by default").
1. It doesn't require options.
1. It doesn't require a lot of view/UI, code and effort to maintain.
1. It doesn't require an external server; it can be done via API or GitHub.com itself.
1. It doesn't require too much HTTP traffic.
1. It's a safe operation and not bad practice.

### Features from other extensions

If it already exists as an extension, use it, support it, extend it. Not everything needs to be part of Refined GitHub.

### "Disabled by default" features

"Disabled by default" means most people don't need it, so it doesn't belong to Refined GitHub. In that case, it's best to keep it as a standalone userscript or extension.

### Configurable features

Anything that requires configuration most likely does not belong to Refined GitHub. It’s just extra complexity and we already have 130+ features to attend to.

Imagine you want a feature to be "per branch", someone else wants it "per organization", etc. The options’ UI would just balloon because of different needs.

fregante:
## 1. It doesn't already exist as a [userscript](https://github.com/Mottie/GitHub-userscripts) or extension.

Use it, support it, extend it. Not everything needs to be part of Refined GitHub.

fregante:
## 2. It makes sense as a default for everyone.

"Disabled by default" means most people don't need it, so it doesn't belong to Refined GitHub. In that case, it's best to keep it as a standalone userscript or extension.

fregante:
## 3. It doesn't require options.

Anything that requires configuration most likely does not belong to Refined GitHub. It’s just extra complexity and we already have 130+ features to attend to.

Imagine you want a feature to be "per branch", someone else wants it "per organization", etc. The options’ UI would just balloon because of different needs.

fregante:
## 4. It doesn't require a lot of view/UI, code and effort to maintain.

Refined GitHub is about small refinements and small features. If a feature requires a considerable amount of UI controls or especially its own page/URL, we can't add it, it would take too much effort.

fregante:
## 5. It doesn't require an external server.

Most of us are privacy conscious and don't want to add 3rd party servers in the mix, with all of that entails, especially maintenance, privacy and security.

[...]

PR #3036, opened 2020-04-26
Add `stop-redirecting-in-notification-bar` feature

### LINKED ISSUE #2952: Stay on the page after clicking "Done" in the Notifications bar
After opening a notification, I get to the page that triggered it (e.g. a PR or issue I subscribed to). At the top of the page, there is a light blue bar with buttons that act on the notification:

Screenshot_2020-03-29_14-07-14

I'd like to have a quick way of marking the notification as "Done" (and maybe also the other actions) on that page without being redirected back to the Notifications page.

Two possible solutions I can imagine:

1. When clicking on the buttons while pressing the "Alt" keyboard button, the button action is executed, the notification bar disappears, but I stay on the page. When "Alt" is not pressed, the behavior doesn't change.
2. When clicking on the buttons, the button action is executed, the notification buttons change (e.g. after clicking "Done", the only valid action is now "Move to inbox"), but I stay on the page. To go back to notifications or close the notification bar, I have to click the respective button.

Similar to #2546.

FloEdelmann:
Closes #2952. I implemented ~version 2 from the issue since it was easier and more people will discover it~ *a mix of version 1 and 2*.

refined-github-notifications-bar

Test this on this page if you came here from a notification :wink:
Otherwise, go to any other page via your [notifications list](https://github.com/notifications) ([done notifications](https://github.com/notifications?query=is%3Adone) work fine, too).


fregante:
I don’t want this to be a default. If you’re “Done” with it then why do you need to stay on the page?

As you suggested in the issue, it’s best to do this when alt-clicking it

source/features/stop-redirecting-in-notifications-bar.tsx
```diff
@@ -0,0 +1,25 @@
...
+	delegate(document, actionButtonSelector, 'click', onNotificationActionClick);
+}
+
+features.add({
+	id: __filebasename,
+	description: 'Stops redirecting to notification inbox from notification bar actions while holding `Alt`.',
+	screenshot: 'https://user-images.githubusercontent.com/202916/80318782-c38cef80-880c-11ea-9226-72c585f42a51.png'
+}, {
+	include: [
+		pageDetect.isRepo
```
fregante, line 22: I think
```suggestion
		pageDetect.isPR,
		pageDetect.isIssue
```

source/features/stop-redirecting-in-notifications-bar.tsx
```diff
@@ -0,0 +1,25 @@
+import delegate from 'delegate-it';
+import features from '../libs/features';
+import * as pageDetect from '../libs/page-detect';
+
+function onNotificationActionClick(event: delegate.Event<MouseEvent, HTMLButtonElement>): void {
```
fregante, line 5: I like that the way this function is written doesn't break successive non-alt clicks. Can you mention this fact in a comment?

source/features/stop-redirecting-in-notifications-bar.tsx
```diff
@@ -0,0 +1,25 @@
...
+	const shouldRedirectToInbox = !event.altKey;
+	const actionForm = event.delegateTarget.closest('form')!;
+	actionForm.toggleAttribute('data-redirect-to-inbox-on-submit', shouldRedirectToInbox);
+}
+
+async function init(): Promise<void> {
+	const actionButtonSelector = '.notifications-v2.notification-shelf .js-notification-action button';
```
fregante, line 12: I think only this is needed:

```suggestion
	const actionButtonSelector = '.notification-shelf .js-notification-action button';
```

source/features/stop-redirecting-in-notifications-bar.tsx
FloEdelmann: At least commits and releases can be notification targets, too.

fregante: Indeed. I then we need a new detection function in the current file. Maybe:

```js
const hasNotification = () => location.search.startsWith('?notification_referrer_id=');
```

And then:

```js
{
	include: [
		hasNotification
	],
	waitForDomReady: false,
	repeatOnAjax: false,
	init
}
```

FloEdelmann: This won't work if you reload on a page with notification bar, as the URL query parameter is removed.

E.g. `https://github.com/sindresorhus/refined-github/pull/3036?notification_referrer_id=XXXXXXXX#pullrequestreview-401122761` will be detected. That turns to `https://github.com/sindresorhus/refined-github/pull/3036#pullrequestreview-401122761` immediately, so after a reload, that page will not be detected, but the notifications bar is still there.

fregante: > so after a reload [...] the notifications bar is still there.

That doesn't match my experience:

demo2

From what I see, GitHub uses exactly the same `location.search` test

FloEdelmann: Weird, when I reload this page, the notifications bar re-appears after a short delay. I also tried disabling all GitHub-related browser extensions (Refined GitHub, OctoLinker and Notifications Preview for GitHub), but my experience doesn't change.

fregante: > It apparently only re-appears on PRs, not issues or releases.

Indeed. Weirdly enough, PRs store this state in `sessionStorage`. Maybe:

```js
const hasNotificationBar = () =>
	location.search.startsWith('?notification_referrer_id=') ||
	JSON.parse(sessionStorage.notification_shelf ?? '{}').pathname === location.pathname;
```

FloEdelmann: That works fine!

source/libs/page-detect.ts
```diff
@@ -414,3 +414,8 @@ export const _isActionPage = [
 	'https://github.com/marketplace/actions/github-action-for-assignee-to-reviewer',
 	'https://github.com/marketplace/actions/hugo-actions'
 ];
+
+export const _hasNotificationBar = skip;
+export const hasNotificationBar = (): boolean =>
+	location.search.startsWith('?notification_referrer_id=') ||
+	JSON.parse(sessionStorage.notification_shelf ?? '{}').pathname === location.pathname;
```
fregante, line 421: This is too specific to be part of `page-detect`, can you move it to the feature’s file?

FloEdelmann: Done.

fregante:
Thanks Flo!

FloEdelmann:
@fregante Thanks for reviewing and merging :blush:

tonglil:
My reason for wanting it to stop redirecting is that I often open many tabs at once, either through 
image
or ctrl-clicking the ones I need to check "as a batch".

The redirection is weird because I don't know how long to wait before it has marked it as done so it is safe to close the tab, but the redirect is slow.

Ideal workflow: press "done" + ctrl+w to move on to the next tab
Not ideal: press "done", wait for redirect to start, ctrl+w -> this just feels slow and jarring (two browser paints in quick succession)

karlhorky:
Thanks for this feature!

I would like to voice my support for having an option to turn this on **without** holding `alt`. I do agree with @fregante that it should be an option, not turned on by default.

fregante:
1. This extension only tries to improve the experience, but this seems just _different_ than the expected/standard behavior.
2. We don't have ["off by default" features](https://github.com/sindresorhus/refined-github/issues/2960#issuecomment-606765382)

Let's discuss on #3173

PR #3037, opened 2020-04-26
Update `web-ext run` command in contributing guide

FloEdelmann:
The npm script was removed in https://github.com/sindresorhus/refined-github/pull/2945#pullrequestreview-400512900 because of #2683.


contributing.md
```diff
@@ -135,7 +135,11 @@ Once built, load it in the browser of your choice:
 			<p>Or you can use <a href="https://developer.mozilla.org/en-US/Add-ons/WebExtensions/web-ext_command_reference#web-ext_run"><code>web-ext run</code></a> to have Firefox automatically load and reload it:</p>
 			<pre>npx web-ext run --source-dir=distribution</pre>
 			<p>Recommended arguments:</p>
-			<pre>npx web-ext run --source-dir=distribution --firefox-profile=refined-github --keep-profile-changes --start-url=github.com</pre>
+			<ul>
+				<li><code>--firefox-profile=refined-github</code></li>
+				<li><code>--keep-profile-changes</code></li>
+				<li><code>--start-url=github.com</code></li>
+			</ul>
```
fregante, line 142: I liked it better in a single command. You can make it multiline as:

```suggestion

			```
			npx web-ext run \
			--source-dir=distribution \
			--firefox-profile=refined-github \
			--keep-profile-changes
			--start-url=github.com
			```
```

Also I'd like to see this for Chrome as well. Perhaps it should be a separate table. Like:

## Loading into the browser

Once built, load it in the browser of your choice.

### Load with [`web-ext`](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/web-ext_command_reference#web-ext_run)

< table with web-ext commands>

### Load manually

< current table >

fregante:
Could you double-check both commands? They don't work

```
❯ npx web-ext run --source-dir=distribution \
  --target=firefox \
  --firefox-profile=refined-github \
  --keep-profile-changes \
  --start-url=github.com

Invalid values:
  Argument: target, Given: "firefox", Choices: "firefox-desktop", "firefox-android", "chromium"
```

FloEdelmann:
Indeed, the `firefox` target (although mentioned in the `web-ext` docs) is invalid. I replaced it with `firefox-desktop`. The Chromium command is working fine though.

fregante:
```
❯ npx web-ext run --source-dir=distribution \
    --target=firefox-desktop \
    --firefox-profile=refined-github \
    --keep-profile-changes \
    --start-url=github.com
Applying config file: ./package.json
Running web extension from ./distribution

s: The request "refined-github" profile name cannot be resolved to a profile path
```

FloEdelmann:
You will have to create a new profile called `refined-github` first: https://support.mozilla.org/en-US/kb/profile-manager-create-remove-switch-firefox-profiles

fregante:
> You will have to create a new profile called `refined-github` first:

You tell me, but you should tell the people who read the Contributing guide. If we suggest something it should work, we shouldn't send them on a Google ride

FloEdelmann:
Fair point. I added a note for both browsers.

fregante:
I don't think it's worth suggesting flags without giving the user complete instructions. 

Please make sure both commands run without having to customize them or reduce them or drop them.

We're including them to improve the developer experience, not make it harder to contribute.

FloEdelmann:
I added the `web-ext` parameters that apply to all users to the `package.json` (like explained [here](https://extensionworkshop.com/documentation/develop/getting-started-with-web-ext/#setting-option-defaults-in-a-configuration-file)). This makes the default command easy to use.

Chrome's profile directory doesn't need to be in the [user data directory](https://chromium.googlesource.com/chromium/src/+/master/docs/user_data_dir.md), so I hard-coded the path to be inside the repo folder and added it to `.gitignore`.

This way, users can use either the simple command (no custom profile) without any modification, or the one with a custom profile with no (Chrome) or minimal (Firefox) setup.

fregante:
This seems to work perfectly. Firefox even updates the CSS on the fly so it's super easy to make CSS changes now.

I'll leave this open a little longer to let more people see it, possibly.

FloEdelmann:
Nice! The profile directory options are badly documented for both Firefox and Chrome, so I didn't think about specifying a directory for Firefox, too.

One thing that might be a problem is that we use the same directory for Chrome's and Firefox's profiles, so they might overwrite each other's files. To be safe, we could specify one directory for each browser.

However, we can also leave it like it is, and if a user actually uses both commands and runs into problems, they can just remove the directory and start over, or we can have another look at this then.

fregante:
> One thing that might be a problem is that we use the same directory for Chrome's and Firefox's profiles, so they might overwrite each other's files. To be safe, we could specify one directory for each browser.

Nah, it doesn't seem to happen at the moment and these are testing profiles, no important data is lost. Let's keep it simple

Issue #2911, opened 2020-03-21
Notifications Beta: `open-all-notifications` doesn't work


jack1142:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%2410%20Funded-%2300A156.svg" />](https://issuehunt.io/r/sindresorhus/refined-github/issues/2911)

Not entirely sure if you are currently interested in fixing issues related to new notifications (still in beta I think?) but the [`open-all-notifications`](https://github.com/sindresorhus/refined-github/blob/master/source/features/open-all-notifications.tsx) doesn't work anymore - there's no "Open all in tabs" button anywhere.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $10.00)

- [<img src='https://avatars0.githubusercontent.com/u/183673?v=4' alt='janpio' width=24 height=24> janpio](https://issuehunt.io/u/janpio) ($10.00)

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/sindresorhus/refined-github/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

gibfahn:
I no longer see the option to go back to the old version, so can no longer work around this.

I'll take a look if I get the time, but I'm unikely to get the time 😁 .

SuperSandro2000:
I got also forced into the new notification system and already missing the new tab button.

riyathawani1:
yes, sometime it does not work for me as well. 

https://babatax.com/gstin/

Jolg42:
I just created the PR #3046 that fixes it 😉 (First commit to Refined GitHub 🎊)

Jolg42:
@fregante I didn't want to be mean in PR comments 🙃
From my point of view your comments were confusing sometimes 🤷‍♂️ 

Now it looks like I don't need to do anything anymore in PR #3046 as you copied what I did into your own PR #3047 👀 

Thanks for the speedy feedback anyway.

fregante:
I closed your PR because I can't ask the same thing 3 times and then be told that screenshots are confusing. I didn't copy what you did, almost nothing is the same.

PR #3046, opened 2020-05-01
Restore `open-all-notifications` feature

Jolg42:
1. LINKED ISSUES:
   Closes #2911

2. TEST URLS:
   https://github.com/notifications

3. SCREENSHOT:
Screen Shot 2020-05-01 at 16 06 35

I fixed the "Open all unread in tabs" for the new notification page that is now per default.
I removed the `addOpenReposButton` as it's broken anyway.

It will make @janpio really happy 😉

PR #3047, opened 2020-05-01
Restore `open-all-notifications` feature

fregante:
Fixes https://github.com/sindresorhus/refined-github/issues/2911
Replaces #3046

## Test

Try both the "Grouped by date" and "Grouped by repository" representations


fenollp:
This doesn't seem to work as intended when there are filters in the textbox: the "open all in new tabs" button then only shows up in the "Save" filters on the right to the filter textbox.
Or maybe I have another plugin colliding with this one?

fregante:
Screenshot and link please. In my tests it works correctly.

Exception: https://github.com/notifications?query=is%3Asaved but it can't work there because
- no header
- no concept of "unread" (and no class to select them either)

Issue #2561, opened 2019-11-21
Make the dropdown of GitHub Actions a single button, if there's only one action


tunnckoCore:
[Here](https://github.com/sindresorhus/refined-github/commit/0ae4e39ec1d10e88cf8605a6a0536cbdc5c50fe0/checks?check_suite_id=321434835) you should see a dropdwon "Re-run checks", which has only one item "Re-run all checks". I propose to make the dropdown a button "Re-run all checks" directly. At least for now, until we get more items in that dropdown.

More: https://twitter.com/tunnckoCore/status/1197456822891896833

fregante:
This feature should be something like:

```js
if dropdown has 1 action
	add 'btn' class to action
	dropdown.replaceWith(action)
```

mehrdadn:
Just wanted to request that there also be a button for canceling the run :)

tunnckoCore:
@fregante true.

tunnckoCore:
<3

PR #3050, opened 2020-05-02
Add `unwrap-useless-dropdowns` feature

fregante:
# Test


fregante:
The same feature (but a separate `init()`) should also include a fix for https://github.com/sindresorhus/refined-github/issues/2561, but the button doesn't appear all the time anymore, they may be working on it now.

karlhorky:
Awesome @fregante, thanks!! So much better UX.

fregante:
I'm glad you like it! I'm sure these menus will be filled with more options at some point, but at the moment they're just _one extra click_

PR #3051, opened 2020-05-02
Add `clean-notifications` feature

### LINKED ISSUE #3030: Make new notifications behave more like old notifications UI
Some people like the new UI, some people hate it.

GitHub (I don't know when) added "group by repository" which kinda makes the UI look like the old one (I personally like this view), but there are some things missing from this and some things that can be removed.

1. Linkify `user/repo` when grouping by notifications is applied, allows you to quickly go to a repo (there is no link on notifications UI that lets you do this).
2. Remove redundant `user/repo` on each notification when grouping by notifications is applied, also move issue/PR id down.
3. Move notification count and notification reasons on top of each other, thay should make more room for notification title
4. Convert blue dots on the left to blue bars, similar to how it looks like on issue/PR lists

Why am I opening this issue: reading this https://twitter.com/ljharb/status/1253018009091731456 (/cc @ljharb)
Why don't I have any screenshots included: I'm opening this issue from mobile.

Some personal preferences in here, open to discussion and be proven wrong.

Related: #2891, #2961.

notlmn:
Closes #3030.

## Test

https://github.com/notifications (only on inbox with group by repositories applied)


notlmn:
From https://github.com/sindresorhus/refined-github/issues/3030#issuecomment-622628004.

1. Have't included `.js-notifications-group` as the dropdown check fails anyway expect only on "Inbox".
2. I tried doing this, but that messed up all the UI, maybe because the elements are under different parents, or having a stray `#text` element. So ignored that as font difference for id (`12px`) and title (`14px`) is only `2px` (`1px` on top and bottom).

source/features/clean-notifications.tsx
```diff
@@ -0,0 +1,23 @@
...
+features.add({
+	id: __filebasename,
+	description: 'Removes redundant user/repo from notifications when being grouped by repository',
+	screenshot: 'https://user-images.githubusercontent.com/37769974/80862883-0419bc80-8c96-11ea-84a7-15baf1eac0e6.png'
+}, {
+	include: [
+		pageDetect.isNotifications
+	],
+	init
+});
```
fregante, line 23: Why add a JS file when adding the extra selector as suggested already works correctly? This feature should be CSS only. This is the whole CSS:

```css
/* Place discussion number before the title */
.js-notifications-group [id^="notification"] {
    flex-direction: row !important;
}

/* Hide duplicate repository URL */
.js-notifications-group [id^="notification"] > .d-flex > .f6 {
    font-size: 0 !important;
}

/* Size and position the discussion number */
.js-notifications-group [id^="notification"] > .d-flex > .f6 > span {
	display: block;
    font-size: 12px;
    margin-top: 2px; /* Align baseline with discussion title */
    margin-right: 8px;
    word-break: keep-all;
}
```

notlmn: > Why add a JS file

So that this feature can be disabled.

fregante: > in case anything actually breaks.

That applies to any CSS feature, but we don't make every CSS feature disableable "just in case it breaks."

If it breaks _we_ should just disable it for everyone.

source/refined-github.css
```diff
@@ -114,3 +114,22 @@ pr-branches
...
+}
+/* Hide duplicate repository URL */
+.js-notifications-group [id^='notification'] > .d-flex > .f6 {
+	font-size: 0 !important;
+}
+/* Size and position the discussion number */
+.js-notifications-group [id^='notification'] > .d-flex > .f6 > span {
+	display: block;
+	font-size: 12px;
+	font-variant-numeric: tabular-nums;
```
notlmn, line 131: Added to be careful about some fonts with non-tabular numbers.

@fregante WDYT? Leave this here or remove it?

source/refined-github.css
```diff
@@ -114,3 +114,22 @@ pr-branches
...
+	margin-top: 2px; /* Align baseline with discussion title */
+	margin-right: 8px;
+	word-break: keep-all;
+}
```
fregante, line 135: Can you keep this CSS to its own `clean-notifications.css` file?

FloEdelmann:
@fregante @notlmn Now, the styles are restricted to a `rgh-clean-notifications` class which is not added anywhere (because there is no `tsx` feature file). I guess it was removed by accident in [`c272ba1`](https://github.com/sindresorhus/refined-github/pull/3051/commits/c272ba147fa56b13dff729853c2ef24c1d59e8ee).

notlmn:
@fregante might've copied the CSS from the wrong commit.

Fixed in https://github.com/sindresorhus/refined-github/commit/09919eb5ae018246c30b8f4510d606c86303255f.

fregante:
D'oh! Thankfully this doesn't break anything but it means it will be published in the next release.

PR #3056, opened 2020-05-03
Fix console errors in `clean-issue-filters`

FloEdelmann:
Test on https://github.com/fregante/slash-tag-test/issues

The errors shows up only on Chrome, apparently because they happen inside Promises:

```
Uncaught (in promise) TypeError: Cannot read property 'parentElement' of undefined
    at hideMilestones (refined-github.js:11936)
Uncaught (in promise) TypeError: Cannot read property 'parentElement' of undefined
    at hideProjects (refined-github.js:11954)
```

### DIFF
--- a/source/features/clean-issue-filters.tsx
+++ b/source/features/clean-issue-filters.tsx
@@ -58,10 +58,10 @@ async function hideProjects(): Promise<void> {
 	}
 }
 
-async function init(): Promise<void> {
+async function init(): Promise<void | false> {
 	if (!await elementReady('#js-issues-toolbar')) {
 		// Repo has no issues, so no toolbar is shown
-		return;
+		return false;
 	}
 
 	hideMilestones();
--- a/source/features/clean-issue-filters.tsx
+++ b/source/features/clean-issue-filters.tsx
@@ -59,6 +59,11 @@ async function hideProjects(): Promise<void> {
 }
 
 function init(): void {
+	if (!select.exists('#js-issues-toolbar')) {
+		// Repo has no issues, so no toolbar is shown
+		return;
+	}
+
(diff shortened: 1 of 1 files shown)


source/features/clean-issue-filters.tsx
```diff
@@ -59,6 +59,11 @@ async function hideProjects(): Promise<void> {
 }
 
 function init(): void {
+	if (!select.exists('#js-issues-toolbar')) {
+		// Repo has no issues, so no toolbar is shown
+		return;
+	}
```
fregante, line 65: We have `waitForDomReady: false` so we should use `elementReady`.

```suggestion
async function init(): Promise<void | false> {
	if (!await elementReady('#js-issues-toolbar')) {
		// Repo has no issues, so no toolbar is shown
		return;
	}
```

This however means we can drop the other 2 `elementReady` from the page

source/features/clean-issue-filters.tsx
```diff
@@ -54,11 +54,16 @@ async function hasProjects(): Promise<boolean> {
...
-		(await elementReady('[data-hotkey="p"]'))!.parentElement!.remove();
+		select('[data-hotkey="p"]')!.parentElement!.remove();
 	}
 }
 
-function init(): void {
+async function init(): Promise<void> {
+	if (!await elementReady('#js-issues-toolbar')) {
+		// Repo has no issues, so no toolbar is shown
+		return;
```
fregante, line 64:
```suggestion
async function init(): Promise<void | false> {
	if (!await elementReady('#js-issues-toolbar')) {
		// Repo has no issues, so no toolbar is shown
		return false;
```

Issue #3055, opened 2020-05-03
Incorrect tab size in GitHub diff UI


notlmn:
## Explanation

This is a restriction of tab characters being collapsed to nearest tab point because of that extract character at the start of the line.

Here is an example:

``` diff
$ git diff master
diff --git a/source/refined-github.css b/source/refined-github.css
index 8d7a559..3d8cdd5 100644
--- a/source/refined-github.css
+++ b/source/refined-github.css
@@ -114,3 +114,22 @@ pr-branches
 .js-issue-row .text-small.text-gray {
        line-height: 1.8;
 }
+
+/* Removes redundant user/repo from notifications when being grouped by repository */
+/* Place discussion number before the title */
+.js-notifications-group [id^='notification'] {
+       flex-direction: row !important;
+}
+/* Hide duplicate repository URL */
+.js-notifications-group [id^='notification'] > .d-flex > .f6 {
+       font-size: 0 !important;
+}
+/* Size and position the discussion number */
+.js-notifications-group [id^='notification'] > .d-flex > .f6 > span {
+       display: block;
+       font-size: 12px;
+       font-variant-numeric: tabular-nums;
+       margin-top: 2px; /* Align baseline with discussion title */
+       margin-right: 8px;
+       word-break: keep-all;
+}
```

Here Git renders a tab character with 7 spaces, 1 for the marker at the front. This is a limitation (if you could say that) on terminals, but ideally should not be on the web where we have much more control.

This is especially noticeable when we have `tab-size` set to `4` with this extension, now we have the first tab character of each line being rendered to only 2 physical spaces.

We kinda solved this with `remove-diff-signs`, where we added the ability to remove markers completely from the DOM flow, restoring the tab translation to 4 as intended. But after #2970, this issue got introduced again, with the diff markers interfering with the tab characters at the start of the line.

## Reproduction

- If having extension **_disabled_**
	- Navigate to https://github.com/sindresorhus/refined-github/pull/3051/files (or any GitHub page containing diffs)
	- The first tab space should 8 characters wide, but it is only 6 characters wide.
- If having extension **_enabled_**, the first tab space should 4 characters wide, but
	- On https://github.com/sindresorhus/refined-github/pull/2925/files#diff-cc4aac3e9be04e0413c9520f223b493c, it is 3 characters wide (for Markdown files).
	  ![image](https://user-images.githubusercontent.com/37769974/80908637-17349700-8d3f-11ea-98d2-66105629cf59.png)
[...]

PR #3058, opened 2020-05-03
Add `fix-first-tab-length` feature

fregante:
Fixes https://github.com/sindresorhus/refined-github/issues/3055

**Note:** this works regardless of `remove-diff-signs`’ status.

# Test

Needs further testing, but this is the starting point:

https://github.com/sindresorhus/refined-github/commit/a741080862a6a5c2f6f85a331feabd8f81684403#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R161

Before (short first tab) and after (all tabs are 4-characters long)
demo2

### DIFF
--- a/source/features/fix-first-tab-length.css
+++ b/source/features/fix-first-tab-length.css
@@ -1,0 +1,13 @@
+/* Fixed a GitHub bug: In diffs, the first tab on each line appears 2 characters shorter than it should be. #3055 */
+
+/* Remove diff signs the layout flow */
+.blob-code-marker::before {
+	position: absolute;
+	transform: translate(-100%, 2px);
+}
+
+/* Restore the space that collapsed as a consequence of the previous rule */
+.code-review.blob-code {
+	padding-left: 26px; /* GitHub sets it to 18px */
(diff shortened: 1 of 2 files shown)


notlmn:
@fregante, what specifically made you drop this PR?

fregante:
https://github.com/sindresorhus/refined-github/issues/3055#issuecomment-624923363

notlmn:
Yes, but I took a stab at it today, and it seems to work as expected both in unified and diff views.

``` css
/* Remove diff signs from the layout flow */
.blob-code-inner::before,
.blob-code-marker-cell::before {
	position: absolute;
	transform: translateX(-100%);
}

/* Restore the space that collapsed as a consequence of the previous rule */
.blob-code {
	padding-left: 25px !important; /* GitHub sets it to 18px */
	text-indent: 0 !important; /* GitHub sets it to -7px */
}

/* Adjust add comment button position, because of extra padding above */
.add-line-comment {
	margin-left: -37px !important; /* GitHub sets this at -28px */
}

```

fregante:
Place it in the CSS box in the options and test it out for a few days. If it works open a PR

PR #3062, opened 2020-05-05
Extract page-detect.ts into `github-url-detection`

fregante:
As discussed a while back, I'm going to extract the page detection functions into a standalone module: https://github.com/fregante/github-url-detection

## Pros

- They can be reused by others
- This frees us from testing them here on every commit (the tests often time out)

## Con

- Function updates and additions need to be made to a separate repo (but can temporarily also be in the feature’s file)

### Con alleviator

- [ ] Set up automatic deployment and maybe automatic protected PR merging

## TODO

- [x] Pick name **_(help wanted)_**. Does `github-page-detection` sound good?
- [x] Publish package
- [x] Install dependency here


fregante:
cc @Mottie would these be useful to you? https://github.com/sindresorhus/refined-github/blob/2e5c19a4d8fa7db94e775beeac9b8bd610a2fa2b/source/libs/page-detect.ts

They already depend on https://github.com/Mottie/github-reserved-names

sindresorhus:
> Does github-page-detection sound good?

👍

yakov116:
If you want I resorted the imports for you https://github.com/yakov116/refined-github/commit/12ab361736e0b61ac2635db76cd7eb2fbdd3ffae and removed the duplicate imports

fregante:
> > Does github-page-detection sound good?
> 
> 👍

Eventually I think I'd use `github-url-detection` since GitHub Pages are something completely unrelated and might cause confusion.

fregante:
This is ready to be merged

fregante:
## https://github.com/fregante/github-url-detection 🎉 

Here's the clean output that a non-webpack-built module can achieve outside Refined GitHub: https://www.unpkg.com/browse/github-url-detection@1.0.0/esm/index.js

This will also speed up our testing and avoid CI testing timeouts.

PR #3070, opened 2020-05-06
Update `highest-rated-comment` style to match a Discussion "Answer"

fregante:
[GitHub Discussions](https://github.com/zeit/next.js/discussions/12559#discussioncomment-9555) have a "Selected answer" style we can reuse.

# Test

https://github.com/facebook/create-react-app/issues/1333#issuecomment-327776666

# Comment

a

# Link (unchanged)


yakov116:
Shouldn't the big check also have a tooltip?

Issue #3084, opened 2020-05-12
Explore using `selector-observer` to run features


fregante:
GitHub uses [selector-observer](https://github.com/josh/selector-observer) to enable logic on simple classes, which is why adding or removing classes like `.js-fit-content` will enable or disable their features.

It's built around `MutationObserver` with various optimizations on top of it. Its performance might not be _super_ (https://bugzilla.mozilla.org/show_bug.cgi?id=1422522) but it seems to work well enough for GitHub.

It's worth investigating if anyone has time, it would elegantly solve issues around dynamic content and all the specific listeners we have. 

I think `observe` calls would go inside each feature’s `init`, which could only be called once.

fregante:
Many of the bugs currently open can be fixed with this, so I really hope that someone can figure it out.

fregante:
Let’s avoid any more conversions to selector-observer for the time being. There are some outstanding drawbacks we need to solve first. For example:

- it always automatically waits for domready, so it’s not as fast as one using element-ready.
- sometimes it should be disabled on `deinit`, like for that recent feature to remove labels. 
- some features should only be applied when the elements are visible (I have some code about this in the first related PR)

We should find a common, efficient way to cover all of these issues. Either via code or via documentation (e.g. “if feature requires X, then you must use deinit”)

fregante:
cc @yakov116 @kidonng

kidonng:
> * it always automatically waits for domready, so it’s not as fast as one using element-ready.

I don't have a strong feeling about this so I can't say anything.

> * sometimes it should be disabled on `deinit`, like for that recent feature to remove labels.

I think most of the time this could be solved with carefully chosen selectors.

> * some features should only be applied when the elements are visible (I have some code about this in the first related PR)

Are you referring to [this one](https://github.com/sindresorhus/refined-github/pull/3091/files#diff-97fdf073abb60110f8e6719ed5280bd0)?

---

I'm not familiar with all existing and pending usage of selector-observer, but I would say it is definitely useful for dealing with ajaxed components. For sure we don't want to convert every feature into using it, but I think our original intention is to fix bugs which will be hard/verbose without it. Take my two open PR converting into selector-obsever as an example, I think there's a good reason to use it.
Existing usage like `linkify-user-edit-history-popup` and `remove-label-faster` may be rewritten to reduce the usage of selector-observer (judge from a glance, correct me if I'm wrong).

fregante:
[...]

PR #2066, opened 2019-05-23
Restore and extend url linkifier

bfred-it:
Fixes https://github.com/sindresorhus/refined-github/issues/695 (oldest bug open)
Fixes https://github.com/sindresorhus/refined-github/issues/1499

Thanks to https://github.com/bfred-it/zip-text-nodes. Comments welcome on this package and its name too

# Test

URLs in diffs
https://github.com/Homebrew/homebrew-core/pull/18015/files

URLs in regular files
[/source/features/copy-markdown.js@`3f5fc48`#L45](
https://github.com/sindresorhus/refined-github/blob/3f5fc489e417d4f4a14da5ea423775e9ca9246fd/source/features/copy-markdown.js#L45)

URLs in PR files
https://github.com/sindresorhus/refined-github/pull/546/files#diff-36280fbf56b6409696e3d46d6b30118fR51

URLs in embedded code
https://github.com/sindresorhus/refined-github/blob/3f5fc489e417d4f4a14da5ea423775e9ca9246fd/source/features/copy-markdown.js#L45

Issue reference in file

[/source/features/sort-issues-by-update-time.tsx@`2ade9a8`#L18](
https://github.com/sindresorhus/refined-github/blob/2ade9a84b1894c7879fab9d3e2d045e876f941a6/source/features/sort-issues-by-update-time.tsx#L18)

Issue reference in embedded code

https://github.com/sindresorhus/refined-github/blob/2ade9a84b1894c7879fab9d3e2d045e876f941a6/source/features/sort-issues-by-update-time.tsx#L18

PR #636, opened 2017-07-19
Don't break 'copy' if copy-markdown fails

bfred-it:
Fixes #635 

Untested. Will be running this for a while but I think we can merge it preventively if it doesn't break anything itself.

PR #3091, opened 2020-05-13
`linkify-code` in un-collapsed diffs

### LINKED ISSUE #2091: Apply auto-linking after unfolding diffs
This is a follow-up from https://github.com/sindresorhus/refined-github/issues/1499. Of the repro cases I provided there

>  URLs appear to not be linked even if the URL _as a whole_ was added ([first example](https://github.com/sindresorhus/refined-github/pull/546/files#diff-36280fbf56b6409696e3d46d6b30118fR51)), or even if they're just in comments alongside changes ([second example](https://github.com/sindresorhus/refined-github/pull/636/files#diff-36280fbf56b6409696e3d46d6b30118fR51)—note that I unfolded the top of the diff to get there, not sure if that matters).

> URLs do appear to be linked if they're in non-PR code blocks ([third example](https://github.com/sindresorhus/refined-github/blob/3f5fc489e417d4f4a14da5ea423775e9ca9246fd/source/features/copy-markdown.js#L45)).

The first is fixed with https://github.com/sindresorhus/refined-github/pull/2066, and the third remains fixed, but the second is still broken.

I can repro both if I visit the unfolded part directly ([link](https://github.com/sindresorhus/refined-github/pull/636/files#diff-36280fbf56b6409696e3d46d6b30118fR51)) and if I go to the folded diff ([link](https://github.com/sindresorhus/refined-github/pull/636/files)), then unfold the top (one unfold should be sufficient to see the link, at line 51).

Not sure if this is a regression or no.

fregante:
Fixes #2091 using #3084 

## Test

Additional test URLs to be found on https://github.com/sindresorhus/refined-github/pull/2066#issue-281711810

1. Open https://github.com/sindresorhus/refined-github/pull/636/files
1. Click the "Expand up" button in the diff
1. URL should be linkified:

### DIFF
--- a/source/libs/once-visible-observer.ts
+++ b/source/libs/once-visible-observer.ts
@@ -1,6 +1,15 @@
+import {observe, Observer} from 'selector-observer';
+
+export type ElementCallback = (element: Element) => void;
+
 /** Like `IntersectionObserver`, but call callback ONCE when the observed element becomes visible */
 export default class OnceVisibleObserver extends IntersectionObserver {
-	constructor(callback: (element: Element) => void) {
+	// `selector-observe` compatibility
+	// so it can be used as `observe(selector, new OnceVisibleObserver(callback))`
+	add = this.observe.bind(this);
+	remove = this.unobserve.bind(this);
+
+	constructor(callback: ElementCallback) {
 		super(changes => {
 			for (const change of changes) {
 				if (change.isIntersecting) {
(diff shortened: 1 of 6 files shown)


source/libs/once-visible-observer.ts
```diff
@@ -0,0 +1,35 @@
+import {observe, Observer} from 'selector-observer';
+
+export type ElementCallback = (element: Element) => void;
+
+/** Like `IntersectionObserver`, but call callback ONCE when the observed element becomes visible */
+export default class OnceVisibleObserver extends IntersectionObserver {
+	// `selector-observe` compatibility
```
staabm, line 7:
```suggestion
	// `selector-observer` compatibility
```

fregante:
This is reapplied multiple times, I'll have to restore `linkifiedURLClass`

PR #3092, opened 2020-05-13
Make `show-whitespace` use `selector-observer`

fregante:
Follows #3091 using #3084

`selector-observer` really makes writing features easier


fregante:
The technique needs further work. Refer to https://github.com/sindresorhus/refined-github/pull/3091#issuecomment-628533931

PR #3094, opened 2020-05-14
Use `selector-observe` on more features

fregante:
Follows #3091 using #3084


fregante:
The technique needs further work. Refer to https://github.com/sindresorhus/refined-github/pull/3091#issuecomment-628533931

PR #3096, opened 2020-05-15
Simplify `remove-checks-tab`

yakov116:
1. LINKED ISSUES:
   None
2. TEST URLS:
   https://github.com/RandomEngy/VidCoder/pull/490

3. SCREENSHOT:
   None

~~One thing to note `pageDetect.isRepoWithAccess`~~ 

~~https://github.com/fregante/github-url-detection/blob/f88d029d45d29a31ec6c91de6b1588ea30874986/index.ts#L318~~

~~does not include~~
```js
'.pagehead-tabs-item[href$="/settings/profile"]' // In organizations
```
~~@busches can you check?~~ sorry I mixed up organizations and enterprises 

Edit: There is no need for that check here its only on `isOrganizationProfile` so it was never needed

### DIFF
--- a/source/features/remove-checks-tab.tsx
+++ b/source/features/remove-checks-tab.tsx
@@ -3,7 +3,7 @@ import * as pageDetect from 'github-url-detection';
 
 import features from '../libs/features';
 
-function init(): false | void {
+function init(): void {
 	// Only remove the tab if it's not the current page and if it has 0 checks
 	const checksCounter = select('.tabnav-tab[href$="/checks"]:not(.selected) .Counter');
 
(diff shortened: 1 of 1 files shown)


fregante:
> There is no need for that check here

Indeed, it was copied verbatim from `hide-projects-tab`

yakov116:
Its not related per say to this PR. But I want to know if you thing like [This](https://github.com/yakov116/refined-github/commit/8440849c0d9a1c411662cc2233b8bcd1c5e9530f) you would be interested in?

source/features/remove-checks-tab.tsx
```diff
@@ -1,26 +1,13 @@
...
-	])) {
-		return false;
-	}
-
+function init(): void {
 	// Only remove the tab if it's not the current page and if it has 0 checks
 	const checksCounter = select('.tabnav-tab[href$="/checks"]:not(.selected) .Counter');
 
-	if (checksCounter && checksCounter.textContent!.trim() === '0') {
+	if (checksCounter?.textContent?.trim() === '0') {
```
fregante, line 10:
```suggestion
	if (checksCounter?.textContent!.trim() === '0') {
```

PR #3100, opened 2020-05-17
Use Optional chaining where possible

yakov116:
1. LINKED ISSUES:
   None

2. TEST URLS:
   There should be no changed to functionality, I tested them all.

3. SCREENSHOT:
   None

### DIFF
--- a/source/features/show-names.tsx
+++ b/source/features/show-names.tsx
@@ -25,7 +25,7 @@ async function init(): Promise<false | void> {
 		// Drop 'commented' label to shorten the copy
 		const commentedNode = element.parentNode!.nextSibling;
 		if (commentedNode?.textContent!.includes('commented')) {
-			commentedNode.remove();
+			commentedNode!.remove();
 		}
 	}
 
--- a/source/features/show-names.tsx
+++ b/source/features/show-names.tsx
@@ -24,7 +24,7 @@ async function init(): Promise<false | void> {
 
 		// Drop 'commented' label to shorten the copy
 		const commentedNode = element.parentNode!.nextSibling;
-		if (commentedNode && commentedNode.textContent!.includes('commented')) {
+		if (commentedNode?.textContent?.includes('commented')) {
 			commentedNode.remove();
 		}
 	}
@@ -44,7 +44,7 @@ async function init(): Promise<false | void> {
 		const userKey = api.escapeKey(username);
 
 		// For the currently logged in user, `names[userKey]` would not be present.
-		if (names[userKey] && names[userKey].name) {
+		if (names[userKey]?.name) {
 			// If it's a regular comment author, add it outside <strong>
 			// otherwise it's something like "User added some commits"
 			if (compareNames(username, names[userKey].name)) {
--- a/source/features/show-names.tsx
+++ b/source/features/show-names.tsx
@@ -24,7 +24,7 @@ async function init(): Promise<false | void> {
 
 		// Drop 'commented' label to shorten the copy
 		const commentedNode = element.parentNode!.nextSibling;
-		if (commentedNode?.textContent?.includes('commented')) {
+		if (commentedNode?.textContent!.includes('commented')) {
 			commentedNode.remove();
 		}
 	}
--- a/source/features/show-user-top-repositories.tsx
+++ b/source/features/show-user-top-repositories.tsx
@@ -15,8 +15,8 @@ function buildUrl(queryField: string): URL {
 }
 
 function init(): void {
-	// Show title case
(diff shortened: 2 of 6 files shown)


source/features/show-user-top-repositories.tsx
```diff
@@ -14,13 +14,9 @@ function buildUrl(queryField: string): URL {
...
 
-function init(): false | void {
-	const showcaseTitle = select('.js-pinned-items-reorder-container .text-normal');
-	if (!showcaseTitle) {
-		return false;
-	}
-
-	showcaseTitle.firstChild!.after(
+function init(): void {
+	// Show title case
```
fregante, line 18:
```suggestion
	// Showcase title
```

source/features/show-user-top-repositories.tsx
```diff
@@ -14,13 +14,9 @@ function buildUrl(queryField: string): URL {
...
+	select('.js-pinned-items-reorder-container .text-normal')?.firstChild?.after(
```
fregante, line 19:
```suggestion
	select('.js-pinned-items-reorder-container .text-normal')?.firstChild!.after(
```

You can disable the related ESLint rule **globally** if it complains.

source/features/show-names.tsx
```diff
@@ -24,7 +24,7 @@ async function init(): Promise<false | void> {
 
 		// Drop 'commented' label to shorten the copy
 		const commentedNode = element.parentNode!.nextSibling;
-		if (commentedNode && commentedNode.textContent!.includes('commented')) {
+		if (commentedNode?.textContent?.includes('commented')) {
```
fregante, line 27:
```suggestion
		if (commentedNode?.textContent!.includes('commented')) {
```

source/features/show-user-top-repositories.tsx
yakov116: Should I change the 2 other places I found that i could do this as part of this pr?

fregante: Yes please

fregante:
In this kind of refactorings (small, no feature change, no feature add), no need to follow the PR template.

source/features/show-user-top-repositories.tsx
yakov116: I made a mistake. never mind

source/features/show-user-top-repositories.tsx
```diff
@@ -14,13 +14,9 @@ function buildUrl(queryField: string): URL {
...
-function init(): false | void {
-	const showcaseTitle = select('.js-pinned-items-reorder-container .text-normal');
-	if (!showcaseTitle) {
-		return false;
-	}
-
-	showcaseTitle.firstChild!.after(
+function init(): void {
+	// Showcase title
+	select('.js-pinned-items-reorder-container .text-normal')?.firstChild!.after(
```
yakov116, line 19: After all the `?` was needed

fregante: What do you mean? Make sure you’re using the latest typescript. They changed this behavior in 3.9

yakov116: the `.firstChild!` you cant do after on null
But I solved it another way there in the other PR. (I like that way since with a `?` it will not error if github changes the selector)

fregante: > (I like that way since with a `?` it will not error if github changes the selector)

You forgot that we **want** RGH to throw errors when GitHub changes the selector...

> But I solved it another way in the other PR.

Undo it

fregante: > But I solved it another way in the other PR

That's the right way, **not** silencing errors.

fregante: >  They changed this behavior in 3.9 and now it’s not needed

👆👆👆👆

RGH is still on 3.8, _that_ is the issue.

fregante: Resolved now

fregante:
If you want to do further refactoring:

- files in `libs` can be split between GitHub-related and generic functions 
	- `libs/pr-ci-status.ts` -> `github-helpers/pr-ci-status.ts`
	- `libs/post-form.ts` stays

**_Separately,_** the same could be done for the functions in `utils.ts` and `dom-utils.ts`. I hate that "utils" files become a dumping ground.

- Some larger functions in `utils.ts` and `dom-utils.ts` can be extracted to their own files — or grouped separately:
	- `getCleanPathname`, `parseRoute`, etc could go in `github-helpers/metadata.ts`

fregante:
I'm working on this ⬆️

PR #2941, opened 2020-03-26
Stop `expand-all-collapsed-code` with `esc`

yakov116:
Closes #2939
https://github.com/microsoft/TypeScript/commit/55218d64a58a9964229daa0a38bdfb0a6be717e5#diff-46fd87925e4552c166ec188712741c3f

Not sure if this is what you had in mind.
<!--

Thanks for contributing! 🍄

1. LINKED ISSUES:
   Does this PR close/fix an existing issue? Write something like `Closes #10`

2. TEST URLS:
   Add some test URLs

3. SCREENSHOT:
   Add a screenshot here if your PR makes visual changes

PR #2795, opened 2020-02-17
Add `pinned-issues-update-time` feature

yakov116:
Closes #2747

@fregante I tried my best :) it works.

Typescript made me do funny things lol

image

PR #3101, opened 2020-05-17
Lint and console errors

yakov116:
Follows #3083
This PR fixes 2 console errors and lints another few

1. https://github.com/sindresorhus/refined-github/commit/3c9f3e4dbd997e1e86adf834b44b8d71e0727738 There is no default branch for `default-branch-button` to apply on history lists 
Test https://github.com/sindresorhus/refined-github/commits/master/.editorconfig
2. https://github.com/sindresorhus/refined-github/pull/3101/commits/096bbd1705195c9e2807fa45d3540795ca654218 If you change a pinned repo before the expiration of the cache if will throw.
3. https://github.com/sindresorhus/refined-github/commit/a4b46800e0b0f544ed93f54e3fbb2baf0518c1f7 Removes an outdated selector 
4. https://github.com/sindresorhus/refined-github/commit/32adbc23c268a7a51874b2bac64f3ca6b46af6f0 Follow up to #2941 I forgot to add it that when it disconnects it should remove the listener
5. https://github.com/sindresorhus/refined-github/commit/03686d5f3daf038410f851b878d16e218dd71cb3 **Does the following comment**.

>I usually expect `init` to be at the bottom of the file, just before `features.add`
>
>_Originally posted by @fregante in https://github.com/sindresorhus/refined-github/pull/2795#discussion_r380456476_

Let me know if I should split the PR


source/features/default-branch-button.tsx
```diff
@@ -24,7 +24,11 @@ async function init(): Promise<false | void> {
 		url = replaceBranch(currentBranch, defaultBranch);
 	}
 
-	const branchSelector = (await elementReady('#branch-select-menu'))!;
+	const branchSelector = (await elementReady('#branch-select-menu'));
+	if (!branchSelector) { // File history page
+		return;
+	}
+
```
fregante, line 31: If it doesn't work there, `exclude` it

Folders don't work either: https://github.com/sindresorhus/refined-github/commits/master/source

I'm not sure why `pageDetect.isRepoCommitList` is listed here. Where is this supposed to work? Can you double check?

source/features/pinned-issues-update-time.tsx
```diff
@@ -42,7 +42,12 @@ async function init(): Promise<void | false> {
 	const lastUpdated: Record<string, IssueInfo> = await getLastUpdated(pinnedIssues.map(getPinnedIssueNumber));
 	for (const pinnedIssue of pinnedIssues) {
 		const issueNumber = getPinnedIssueNumber(pinnedIssue);
-		const {updatedAt} = lastUpdated[api.escapeKey(issueNumber)];
+		const {updatedAt} = lastUpdated[api.escapeKey(issueNumber)] || {};
+		// There was a pinned issue that was changed
```
fregante, line 46: This should never happen. This is why I suggest that `cache.function` depends on the arguments received.

Suggested change instead of this one:

```ts
	cacheKey: ([issues]) => __filebasename + ':' + getRepoURL() + ':' + String(issues) // <- I think
```

source/features/default-branch-button.tsx
yakov116: #2946

source/features/pinned-issues-update-time.tsx
yakov116: Worked

source/features/default-branch-button.tsx
yakov116: The reason I did it this way was if I do it in exclude it may not catch it since the waitForDomReady is set to false

fregante: I see, this only works on `isRepoCommitList` **unless** you're in a folder/file

https://github.com/sindresorhus/refined-github/commits/6c46766237ccbf8a8c23ba07aca1e48a11a71b4b

So instead of stopping it here, stop it at the very beginning of `init`:

```js
if parseRoute() filepath == undefined
  return false
```

yakov116: Awesome idea!

yakov116:
Waitng for #3083

source/features/default-branch-button.tsx
```diff
@@ -25,6 +30,7 @@ async function init(): Promise<false | void> {
 	}
 
 	const branchSelector = (await elementReady('#branch-select-menu'))!;
+
```
yakov116, line 33:
```suggestion

```

source/features/default-branch-button.tsx
```diff
@@ -6,9 +6,14 @@ import ChevronLeftIcon from 'octicon/chevron-left.svg';
...
 import {groupButtons} from '../libs/group-buttons';
 import getDefaultBranch from '../libs/get-default-branch';
-import {getRepoURL, getCurrentBranch, replaceBranch} from '../libs/utils';
+import {getRepoURL, getCurrentBranch, replaceBranch, parseRoute} from '../libs/utils';
 
 async function init(): Promise<false | void> {
+	// The branch selector will be on `isRepoCommitList()` **unless** you're in a folder/file
+	if (pageDetect.isRepoCommitList() && parseRoute(location.pathname)[6]) {
+		return false;
+	}
```
fregante, line 15: You can move this to `exclude` i think

yakov116: I put it here since I felt it was not so _simple_, will move it.

source/libs/utils.ts
```diff
@@ -204,3 +205,22 @@ export function getLatestVersionTag(tags: string[]): string {
 
 	return latestVersion;
 }
+
+export function parseRoute(pathname: string) {
```
fregante, line 209: This is weird. Normally TS complains if this type is missing
```suggestion
export function parseRoute(pathname: string): string {
```

yakov116: Really this should not be part of this PR. I Cherry picked it until #3083 is merged. I will fix it there

fregante: Let's merge this one first. It's done

fregante:
I broke it again before merging.............................

PR #3103, opened 2020-05-18
Improve and fix `remove-projects-tab`

yakov116:
Split `remove-projects-tab` into 2 parts.

1. addNewProjectLink()
2. removeProjectsTab()

Test on:
https://github.com/microsoft
https://github.com/twbs/bootstrap
https://github.com/babel


fregante:
> should isOrgRepoWithAccess be moved to the other module

Yes, but this is not an **organization repo** check, it's an **organization** check. It applies to: https://github.com/babel

I suggest renaming them to `canUserEditRepo` (from `isRepoWithAccess`) and `canUserEditOrganization`

`isRepoWithAccess` should still stay, but as a [deprecated](https://github.com/microsoft/TypeScript/issues/390) alias

source/features/set-default-repositories-type-to-sources.tsx
```diff
@@ -6,7 +6,7 @@ function init(): void {
 	const links = select.all<HTMLAnchorElement>([
 		'.user-profile-nav [href$="tab=repositories"]', // "Your repositories" in header dropdown
 		'#user-links [href$="tab=repositories"]', // "Repositories" tab on user profile
-		'.orgnav .pagehead-tabs-item:first-child', // "Repositories" tab on organization profile
+		'.orgnav .pagehead-tabs-item:first-child:not([href$="/settings/profile"])', // "Repositories" tab on organization profile and exclude the settings tab
```
fregante, line 9: Example of [organization page](https://github.com/babel) without a Repositories tab?

yakov116: @fregante no it catches both the repo tab and the settings tab

fregante: `.orgnav` does not appear on repo pages.

fregante: I don't get it. `.orgnav .pagehead-tabs-item:first-child` must always be the repositories tab. 

Example URL.

yakov116: @fregante I'll be by a computer in about 20 min

fregante:
This whole feature can be rewritten in 2 init functions:

1. `addNewProjectLink`
1. `removeProjectsTab`

They have very different conditions, however both depend on `!projectsTab`, I think:

```ts
function getProjectsTab() {
	return elementReady([
		'[data-hotkey="g b"]', // In organizations and repos
		'.user-profile-nav [href$="?tab=projects"]' // In user profiles
	].join());
}
```

source/features/set-default-repositories-type-to-sources.tsx
yakov116: Go to a org that you have access to
image

The second one is the settings tab, I tested on https://github.com/talknsmile

yakov116:
Blocked by https://github.com/fregante/github-url-detection/pull/3

source/features/remove-projects-tab.tsx
```diff
@@ -6,18 +6,21 @@ import * as pageDetect from 'github-url-detection';
...
 
-const addNewProjectLink = onetime(() => {
-	if (pageDetect.isUserProfile()) {
-		// The link already exists on our profile,
-		// and we can't create projects on others' profiles
-		return;
-	}
+export const canUserEditOrganization = (): boolean => pageDetect.isOrganizationProfile() && select.exists('.pagehead-tabs-item[href$="/settings/profile"]');
+
+export const canUserEditRepo = (): boolean => pageDetect.isRepo() && select.exists('.reponav-item[href$="/settings"]');
```
yakov116, line 11:
```suggestion
```
After https://github.com/fregante/github-url-detection/pull/3

yakov116:
This needs more work. Dont review this PR yet (But I can still use https://github.com/fregante/github-url-detection/pull/3)

source/features/remove-projects-tab.tsx
```diff
@@ -31,38 +34,18 @@ const addNewProjectLink = onetime(() => {
...
-		'.js-repo-nav [data-selected-links^="repo_settings"]', // In repos
-		'.pagehead-tabs-item[href$="/settings/profile"]' // In organizations
-	])) {
-		return;
-	}
-
-	// Only remove the tab if it's not the current page and if it has 0 projects
-	if (!projectsTab.matches('.selected') && select('.Counter', projectsTab)!.textContent!.trim() === '0') {
+	// Only remove the tab if it's not the current page and if it has 0 projects or on `isOrganizationProfile` it will be blank
+	if (!projectsTab.matches('.selected') && ['0', ''].includes(select('.Counter', projectsTab)!.textContent!.trim().charAt(0))) {
```
yakov116, line 48: the `charAt(0)` make sure we dont include 10 or 100 etc

fregante: Why this change? Doesn’t seem necessary and it’s not a good solution

source/features/remove-projects-tab.tsx
```diff
@@ -77,5 +60,16 @@ features.add({
 		pageDetect.isUserProfile,
 		pageDetect.isOrganizationProfile
 	],
-	init
+	exclude: [
+		// If there's a settings tab, the current user can disable the projects, so the tab should not be hidden
```
fregante, line 64: The comment is irrelevant now. The function names are self explanatory

source/features/remove-projects-tab.tsx
yakov116: On an org page if there are any projects it will display a number in the counter. If not the counter textcontent will be blank. 
On a repo page the counter will be 0 if there are no projects.

source/features/remove-projects-tab.tsx
fregante: Done

yakov116: Released?

source/features/remove-projects-tab.tsx
fregante: > If not the counter textcontent will be blank.

Not true. Does this work at all on orgs? There's no `.Counter` element there, this code should error.

https://github.com/babel



https://github.com/microsoft



Maybe:

```js
const counter = select('.Counter', projectsTab);
if (!projectsTab.matches('.selected') && (!counter || counter.textContent!.trim() === '0')) {
  return false;
}
```

source/features/set-default-repositories-type-to-sources.tsx
fregante:
```suggestion
		'.orgnav > .pagehead-tabs-item:first-child', // "Repositories" tab on organization profile
```

source/features/remove-projects-tab.tsx
fregante: Uhhh `npm publish` was still waiting for me OTP 😅

Done now

source/features/remove-projects-tab.tsx
```diff
@@ -77,5 +60,15 @@ features.add({
...
+		canUserEditRepo,
+		canUserEditOrganization
+	],
+	init: removeProjectsTab
+}, {
+	include: [
+		pageDetect.isRepo,
+		pageDetect.isOrganizationProfile
+	],
+	init: onetime(addNewProjectLink)
```
fregante, line 73: I think, but not totally sure, that you can just use this since it's outside the ajaxed area
```suggestion
	repeatOnAjax: false
	init: addNewProjectLink
```

source/features/remove-projects-tab.tsx
```diff
@@ -31,40 +30,23 @@ const addNewProjectLink = onetime(() => {
...
-	// If there's a settings tab, the current user can disable the projects,
-	// so the tab should not be hidden
-	if (select.exists([
-		'.js-repo-nav [data-selected-links^="repo_settings"]', // In repos
-		'.pagehead-tabs-item[href$="/settings/profile"]' // In organizations
-	])) {
-		return;
+	const counter = select('.Counter', projectsTab);
+	// Dont run if the tab is selected, it does not exists or the counter is 0
+	if (projectsTab.matches('.selected') || !counter || looseParseInt(counter.textContent!) !== 0) {
```
yakov116, line 45: Took me an hour... to get this straight. I think I got it right

source/features/remove-projects-tab.tsx
```diff
@@ -77,5 +59,20 @@ features.add({
...
+	],
+	init: removeProjectsTab
+}, {
+	include: [
+		pageDetect.isRepo,
+		pageDetect.isOrganizationProfile
+	],
+	exclude: [
+		() => !pageDetect.canUserEditRepo(),
+		() => !pageDetect.canUserEditOrganization()
```
yakov116, line 74: The only time the new projects should be added is if they have access? 
Am I wrong?

source/features/remove-projects-tab.tsx
```diff
@@ -31,40 +30,23 @@ const addNewProjectLink = onetime(() => {
...
-	// If there's a settings tab, the current user can disable the projects,
-	// so the tab should not be hidden
-	if (select.exists([
-		'.js-repo-nav [data-selected-links^="repo_settings"]', // In repos
-		'.pagehead-tabs-item[href$="/settings/profile"]' // In organizations
-	])) {
-		return;
+	const counter = select('.Counter', projectsTab);
+	// Dont run if it does not exists or the counter is 0
+	if (!counter || looseParseInt(counter.textContent!) !== 0) {
```
yakov116, line 45: Took me an hour... to get this straight. I think I got it right

source/features/remove-projects-tab.tsx
```diff
@@ -31,40 +29,22 @@ const addNewProjectLink = onetime(() => {
...
-
-	// If there's a settings tab, the current user can disable the projects,
-	// so the tab should not be hidden
-	if (select.exists([
-		'.js-repo-nav [data-selected-links^="repo_settings"]', // In repos
-		'.pagehead-tabs-item[href$="/settings/profile"]' // In organizations
-	])) {
-		return;
+	// Dont run if it does not exists or the counter is 0
+	if (Number(select('.Counter', projectsTab)?.textContent!) !== 0) {
```
yakov116, line 43: Took me an hour... to get this straight. I think I got it right

source/features/remove-projects-tab.tsx
fregante: Check

yakov116: I did. Just double checking that I am not missing something.

source/features/remove-projects-tab.tsx
fregante: Looks right. You can drop the `!`

source/features/remove-projects-tab.tsx
```diff
@@ -31,40 +29,22 @@ const addNewProjectLink = onetime(() => {
...
-
-	// If there's a settings tab, the current user can disable the projects,
-	// so the tab should not be hidden
-	if (select.exists([
-		'.js-repo-nav [data-selected-links^="repo_settings"]', // In repos
-		'.pagehead-tabs-item[href$="/settings/profile"]' // In organizations
-	])) {
-		return;
+	// Dont run if the counter is 0. (When it does not exists it will converted to a 0)
+	if (Number(select('.Counter', projectsTab)?.textContent) !== 0) {
```
fregante, line 43: Actually, wrong comment and wrong code. "Nothing is converted to zero" and if there's no counter, it will `return`

source/features/remove-projects-tab.tsx
```diff
@@ -31,40 +28,22 @@ const addNewProjectLink = onetime(() => {
...
```
fregante, line 42: Nothing is converted to zero and if there's no counter, it will return.

It should return if:

- has counter and it's > 0

yakov116:
Updated logic and first post.

Dont have a org with projects to test on.

yakov116:
One second! `source\features\hide-zero-packages` is the second part!

source/features/remove-projects-tab.tsx
```diff
@@ -1,25 +1,22 @@
...
 
-const addNewProjectLink = onetime(() => {
-	if (pageDetect.isUserProfile()) {
-		// The link already exists on our profile,
-		// and we can't create projects on others' profiles
-		return;
-	}
+function getProjectsTab() {
+	return elementReady([
+		'[data-hotkey="g b"]:not(.selected)', // In organizations and repos
```
fregante, line 10: I like this, but this also means that if you're on the Projects tab, `addNewProjectLink` also won't run

yakov116: Done

fregante:
This works, but there's a little big problem: counters are loaded after dom-ready so sometimes the tab is removed even on https://github.com/microsoft

fregante:
This seems to work. Thoughts?

yakov116:
Awesome! Very clean

fregante:
🙌

PR #3105, opened 2020-05-18
Move simple early returns to exclude

yakov116:
The commit titles explain each one. 

- Simplify `hide-disabled-milestone-sorter`
- Move early return `preserve-whitespace-option-in-nav`
- Move early return `quick-mention`
- Move early return `repo-wide-file-finder`
- Move early return `view-markdown-source`
- Move early return `wait-for-build`
- Add missing return false to `warning-for-disallow-edits`
- Move early return `tag-changelog-link`
- Move early return `tags-dropdown`


source/features/profile-hotkey.tsx
```diff
@@ -3,13 +3,11 @@ import select from 'select-dom';
...
 import {getUsername} from '../libs/utils';
 
-function init(): false | void {
+function init(): void {
 	const menuItem = select(`a[href="/${getUsername()}"]`);
 
 	if (menuItem) {
 		menuItem.dataset.hotkey = 'g m';
-	} else {
-		return false;
```
fregante, line 12: Did you mean to use `?.`?

```js
select(`a[href="/${getUsername()}"]`)?.dataset.hotkey = 'g m';
```

yakov116: You cant since your assigning something

fregante: Indeed, not valid at all: 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#Optional_chaining_not_valid_on_the_left-hand_side_of_an_assignment

So why drop `return false`?

yakov116: @fregante this is the only place in the code that I found a simple return false that follows an if.

In our words why here do we do return false and in all other features that get swallowed by the first `if` we do not.

fregante: > all other features that get swallowed by the first `if` we do not.

That's a problem with those features, rather than this one. It's not a huge problem given it only changes the debug logger, but in this case I think it also affects the `shortcuts` field of `features.add` (and if it doesn't, it should)

I don't think this feature needs any change

yakov116:
>Do you want to push these to page detect as well? “Is repo empty”

What are you referring to?

source/features/tag-changelog-link.tsx
```diff
@@ -124,5 +120,8 @@ features.add({
 	include: [
 		pageDetect.isReleasesOrTags
 	],
+	exclude: [
+		() => select.exists('.blankslate') // The repo has no tags
```
fregante, line 124: I meant this, sorry

yakov116: Not sure. Its not an empty repo, it anywhere where its a blank state. 
Ex. No issues, No pull requests.

fregante: Right, it becomes “is empty repo” if you prepend it with `isRepo() &&`

yakov116: What about `isBlankRepoState`?

fregante:
🦄

PR #3106, opened 2020-05-18
Fix position of `more-dropdown` and `releases-tab`

### LINKED ISSUE #3102: Project tabs don't fit on their line
Page url: https://github.com/sindresorhus/refined-github

Screenshot with refined-github enabled:
image

Screenshot with refined-github disabled:
image

notlmn:
Fixes #3102.

GitHub likely changed their DOM and moved all the items inside `.reponav` into an unordered list. This change makes sure that we add elements inside that list, and not as its siblings.

Bug explained at https://github.com/sindresorhus/refined-github/issues/3102#issuecomment-630102823 and https://github.com/sindresorhus/refined-github/issues/3102#issuecomment-630209900.

### Test

On any repo that you are admin of, take a look at navigation tab order.

### DIFF
--- a/source/features/more-dropdown.tsx
+++ b/source/features/more-dropdown.tsx
@@ -17,7 +17,7 @@ const repoUrl = getRepoURL();
 
 function createDropdown(): void {
 	// Markup copied from native GHE dropdown
-	appendBefore('.reponav', '[data-selected-links^="repo_settings"]',
+	appendBefore('.reponav > ul', '[data-selected-links^="repo_settings"]',
 		<details className="reponav-dropdown details-overlay details-reset">
 			<summary className="btn-link reponav-item" aria-haspopup="menu">
 				{'More '}
--- a/source/features/more-dropdown.tsx
+++ b/source/features/more-dropdown.tsx
@@ -17,7 +17,10 @@ const repoUrl = getRepoURL();
 
 function createDropdown(): void {
 	// Markup copied from native GHE dropdown
-	appendBefore('.reponav > ul', '[data-selected-links^="repo_settings"]',
+	appendBefore(
+		// GHE doesn't have `reponav > ul`
+		select('.reponav > ul') ?? select('.reponav')!,
+		'[data-selected-links^="repo_settings"]',
 		<details className="reponav-dropdown details-overlay details-reset">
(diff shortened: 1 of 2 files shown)


notlmn:
Also the `<a>` tag that we use for "releases tab", and "more dropdown" should be wrapped inside `li`, for consistency with GitHub's DOM.

@fregante WDYT?

source/features/more-dropdown.tsx
```diff
@@ -17,7 +17,7 @@ const repoUrl = getRepoURL();
 
 function createDropdown(): void {
 	// Markup copied from native GHE dropdown
-	appendBefore('.reponav', '[data-selected-links^="repo_settings"]',
+	appendBefore('.reponav > ul', '[data-selected-links^="repo_settings"]',
```
scottbusche, line 20: GHE (2.19) still doesn't have the `ul` types and this will break the feature. I know we don't care most of the time about GHE, but these two features are used a lot and will break a lot of people and this is why we currently don't have the `ul` check in there. Can we check if the `ul` exists first or use `isEnterprise()` instead?

I think we're now at 2+ years of GitHub having swapped to the `ul` type on GH, but not ported it over to GHE yet. 😭

busches, line 20: GHE (2.19) still doesn't have the `ul` types and this will break the feature. I know we don't care most of the time about GHE, but these two features are used a lot and will break a lot of people and this is why we currently don't have the `ul` check in there. Can we check if the `ul` exists first or use `isEnterprise()` instead?

I think we're now at 2+ years of GitHub having swapped to the `ul` type on GH, but not ported it over to GHE yet. 😭

fregante: I think you just need `ul.reponav, .reponav ul`

busches: No `ul` either way on GHE.

notlmn: That would be `.reponav ul, .reponav` in that case.

fregante: .reponav would be selected first before reaching the UL. As far as I know the selector order doesn’t matter; just DOM order does.

busches: We moved to 2.20 today and it still has the old markup, this is the latest release or GHE 😥

PR #3109, opened 2020-05-19
Drop `replacebranch` in favor of `parseRoute`

yakov116:
1. LINKED ISSUES:
   Follow up to #3083

2. TEST URLS:
	### Default branch button
	https://github.com/sindresorhus/refined-github/commits/incremental-tag-changelog-link
	With pagnation
   https://github.com/sindresorhus/refined-github/commits/incremental-tag-changelog-link?after=51975964dd72a947a51b9a7be118d75e971549d1+34
	#### On a file
	Default branch: https://github.com/sindresorhus/refined-github/blob/master/.editorconfig
	https://github.com/sindresorhus/refined-github/blob/incremental-tag-changelog-link/.editorconfig

	### Latest tag button
	https://github.com/sindresorhus/refined-github/tree/incremental-tag-changelog-link
	#### On a file
	Default branch: https://github.com/sindresorhus/refined-github/blob/master/.editorconfig
	https://github.com/sindresorhus/refined-github/blob/incremental-tag-changelog-link/.editorconfig

3. SCREENSHOT:
   None


fregante:
parseRoute should not change. The operation is identical to what other features do.

yakov116:
@fregante OK
All the other features needed to change the route or something more. Here its the same function just replacing the current branch with something else.

I will change it.

source/libs/utils.ts
```diff
@@ -173,7 +164,7 @@ export function parseRoute(pathname: string): string[] {
 		throw new Error('The branch of the current page must match the branch in the `pathname` parameter');
 	}
 
-	const filePath = parts.replace(currentBranch + '/', '');
+	const filePath = parts.replace(currentBranch, '').replace(/^\//, '');
```
yakov116, line 167: @fregante this change is still needed, there may not be a leading slash.

source/features/default-branch-button.tsx
```diff
@@ -17,11 +17,18 @@ async function init(): Promise<false | void> {
...
+		return false;
+	}
+
 	let url;
 	if (pageDetect.isRepoRoot()) {
 		url = `/${getRepoURL()}`;
 	} else {
-		url = replaceBranch(currentBranch, defaultBranch);
+		pathParts[5] = defaultBranch;
+		url = pathParts.join('/') + location.search;
```
fregante, line 31: No need to preserve the search

source/features/latest-tag-button.tsx
```diff
@@ -95,7 +95,9 @@ async function init(): Promise<false | void> {
 	if (pageDetect.isRepoRoot()) {
 		href = `/${getRepoURL()}/tree/${latestTag}`;
 	} else {
-		href = replaceBranch(currentBranch, latestTag);
+		const pathParts = parseRoute(location.pathname);
+		pathParts[5] = latestTag;
```
fregante, line 99: 5?

yakov116: Branch Name

source/features/default-branch-button.tsx
yakov116: Ok thought it would not hurt

yakov116:
```suggestion
		url = pathParts.join('/')
```

source/features/default-branch-button.tsx
```diff
@@ -17,11 +17,18 @@ async function init(): Promise<false | void> {
...
+		return false;
+	}
+
 	let url;
 	if (pageDetect.isRepoRoot()) {
 		url = `/${getRepoURL()}`;
 	} else {
-		url = replaceBranch(currentBranch, defaultBranch);
+		pathParts[5] = defaultBranch;
+		url = pathParts.join('/')
```
yakov116, line 31:
```suggestion
		url = pathParts.join('/');
```

source/features/latest-tag-button.tsx
yakov116: @fregante good catch. Not sure how I managed that!

source/features/default-branch-button.tsx
```diff
@@ -17,11 +17,18 @@ async function init(): Promise<false | void> {
 		return false;
 	}
 
+	const pathParts = parseRoute(location.pathname);
+	// The branch selector will be on `isRepoCommitList()` **unless** you're in a folder/file
+	if (pageDetect.isRepoCommitList() && Boolean(pathParts[6])) {
```
fregante, line 22: Boolean can be dropped now

fregante:
Test it before asking for a review

source/features/default-branch-button.tsx
```diff
@@ -17,11 +17,18 @@ async function init(): Promise<false | void> {
 		return false;
 	}
 
+	const pathParts = parseRoute(location.pathname);
+	// The branch selector will be on `isRepoCommitList()` **unless** you're in a folder/file
+	if (pageDetect.isRepoCommitList() && !pathParts[6]) {
```
fregante, line 22: Isn’t this supposed to be 5?

fregante:
There’s a change you can make to parseRoute to avoid these mistakes: have it export an object instead of an array. 

Bonus: add a `toString` function to that object to allow stringification without having the user call `join('/')`

yakov116:
How does the , toString()} work I get a syntax error when I try

fregante:
```js
{
	toString() {
		return `/${user}/${repo}/.............`
	}
}
```

yakov116:
...Looks at the branch https://github.com/sindresorhus/refined-github/tree/parser...

@fregante I am in middle of testing. I want to just make sure that I did what you wanted before I continue.

source/libs/utils.ts
```diff
@@ -156,23 +156,33 @@ export function getLatestVersionTag(tags: string[]): string {
...
 		user,
 		repository,
 		route,
-		currentBranch,
-		filePath
-	];
+		branch,
+		filePath,
+		toString() {
+			return `/${this.user}/${this.repository}/${this.route}/${this.branch}/${this.filePath}`.replace(/\/$/, '');
```
fregante, line 183:
```suggestion
			return `/${user}/${repository}/${route}/${branch}/${filePath}`.replace(/\/$/, '');
```

fregante:
That's right

source/libs/utils.ts
yakov116: When I left out the `this` any replacement I made it did not change it in the toString()

fregante:
You can take this opportunity to move this function to its own file

```
source/github-helpers/parse-route.ts
```

yakov116:
What a messy merge

source/features/deep-reblame.tsx
```diff
@@ -66,17 +67,17 @@ async function redirectToBlameCommit(event: delegate.Event<MouseEvent, HTMLAncho
...
-	const pathnameParts = parseRoute(location.pathname);
-	const currentFilename = pathnameParts[5];
+	const path = parseRoute(location.pathname);
+	const {filePath: currentFilename} = path;
 
 	const spinner = <LoadingIcon className="mr-2"/>;
 	blameElement.firstElementChild!.replaceWith(spinner);
 
 	try {
 		const prBlameCommit = await getPullRequestBlameCommit(prCommit, prNumber, currentFilename);
```
fregante, line 77:
```suggestion
		const prBlameCommit = await getPullRequestBlameCommit(prCommit, prNumber, path.filePath);
```

source/features/deep-reblame.tsx
```diff
@@ -66,17 +67,17 @@ async function redirectToBlameCommit(event: delegate.Event<MouseEvent, HTMLAncho
 	const blameHunk = blameElement.closest('.blame-hunk')!;
 	const prNumber = looseParseInt(select('.issue-link', blameHunk)!.textContent!);
 	const prCommit = select<HTMLAnchorElement>('a.message', blameHunk)!.pathname.split('/').pop()!;
-	const pathnameParts = parseRoute(location.pathname);
-	const currentFilename = pathnameParts[5];
+	const path = parseRoute(location.pathname);
+	const {filePath: currentFilename} = path;
```
fregante, line 71:
```suggestion
```

source/libs/utils.ts
fregante: [](https://knowyourmeme.com/memes/la-noire-doubt-press-x-to-doubt)

source/features/latest-tag-button.tsx
```diff
@@ -95,7 +95,9 @@ async function init(): Promise<false | void> {
 	if (pageDetect.isRepoRoot()) {
 		href = `/${getRepoURL()}/tree/${latestTag}`;
 	} else {
-		href = replaceBranch(currentBranch, latestTag);
+		const pathnameParts = parseRoute(location.pathname);
+		pathnameParts.branch = latestTag;
+		href = pathnameParts.toString();
```
fregante: What about this pattern for assignment? (Only if it's done all in one place; don't refactor other code that has conditions, etc)
```suggestion
		href = {
			...parseRoute(location.pathname),
			branch: latestTag
		}.toString();
```

yakov116:
@fregante that review came in very funny.

I cant apply any suggestions

source/libs/utils.ts
yakov116: I tried it again.
Without the this it does not replace anything when calling toString()

fregante: Heh you’re right, the object is changed but the constants aren’t. I wasn’t thinking about the replacements

yakov116: > Heh you’re right

It can happen... sometimes :smile:

yakov116:
@fregante I tested this out seems to be fully working.

Side note: Deep reblame will sometimes fail if there is an open PR. I am not sure fully why, but there can be more than one PR associated with the commit.

See all the blame cases in https://github.com/sindresorhus/refined-github/pull/3083.

I have a way to fix it. But I will leave it for another pr.

PR #3115, opened 2020-05-20
`hide-zero-packages` on organization profiles 

yakov116:
Follows #3103 
1. LINKED ISSUES:
   None

2. TEST URLS:
   https://github.com/microsoft (Should **not** get removed)
	https://github.com/babel (Should get removed)

3. SCREENSHOT:
   None

### DIFF
--- a/source/features/hide-zero-packages.tsx
+++ b/source/features/hide-zero-packages.tsx
@@ -11,7 +11,7 @@ async function init(): Promise<void | false> {
 		'.orgnav > .pagehead-tabs-item[href$="/packages"]:not(.selected)' // `isOrganizationProfile`
 	].join());
 
-	if (!packagesCounter || await getTabCount(packagesCounter, '.Counter, .num') > 0) {
+	if (!packagesCounter || await getTabCount(packagesCounter) > 0) {
 		return false;
 	}
 
--- a/source/features/hide-zero-packages.tsx
+++ b/source/features/hide-zero-packages.tsx
@@ -2,16 +2,20 @@ import elementReady from 'element-ready';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
+import getTabCount from './remove-projects-tab';
 
-async function init(): Promise<void> {
+async function init(): Promise<void | false> {
 	const packagesCounter = await elementReady([
-		'.numbers-summary [href$="/packages"] .num', // `isRepoRoot`
-		'.UnderlineNav-item[href$="?tab=packages"] .Counter' // `isUserProfile`
-	].join(','));
+		'.numbers-summary [href$="/packages"]', // `isRepoRoot`
+		'.UnderlineNav-item[href$="?tab=packages"]', // `isUserProfile`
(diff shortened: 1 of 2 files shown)


yakov116:
That second commit needs to be reverted. Will do it soon.

source/features/remove-projects-tab.tsx
```diff
@@ -31,8 +31,8 @@ async function addNewProjectLink(): Promise<void |false> {
 	);
 }
 
-async function getTabCount(tab: Element): Promise<number> {
-	const counter = select('.Counter', tab);
+export default async function getTabCount(tab: Element, selectors: string): Promise<number> {
+	const counter = select(selectors, tab);
```
fregante, line 35:
```suggestion
	const counter = select('.Counter, .num', tab);
```

source/features/hide-zero-packages.tsx
```diff
@@ -2,16 +2,20 @@ import elementReady from 'element-ready';
...
 
-async function init(): Promise<void> {
+async function init(): Promise<void | false> {
 	const packagesCounter = await elementReady([
-		'.numbers-summary [href$="/packages"] .num', // `isRepoRoot`
-		'.UnderlineNav-item[href$="?tab=packages"] .Counter' // `isUserProfile`
-	].join(','));
+		'.numbers-summary [href$="/packages"]', // `isRepoRoot`
+		'.UnderlineNav-item[href$="?tab=packages"]', // `isUserProfile`
+		'.orgnav > .pagehead-tabs-item[href$="/packages"]:not(.selected)' // `isOrganizationProfile`
```
fregante, line 11: Shouldn't all of them have `:not(.selected)`?

yakov116: Correct the `isUserProfile` since `isRepoRoot` is not a tab

yakov116:
@fregante should we exclude orgRepoWithAccess?

fregante:
What's orgRepoWithAccess?

source/features/hide-zero-packages.tsx
```diff
@@ -2,16 +2,20 @@ import elementReady from 'element-ready';
...
+	].join());
 
-	if (packagesCounter?.textContent?.trim() === '0') {
-		packagesCounter.closest('li, .UnderlineNav-item')!.remove();
+	if (!packagesCounter || await getTabCount(packagesCounter) > 0) {
+		return false;
 	}
+
+	(packagesCounter.closest('li, .UnderlineNav-item') ?? packagesCounter).remove();
```
fregante, line 18: Why did you add `?? packagesCounter`? Doesn't seem to ever be needed

yakov116: Organisations the closest will be null

yakov116:
> What's orgRepoWithAccess?

We should not hide the packages

source/features/hide-zero-packages.tsx
yakov116: Your right, I tested it now, its working just fine

PR #3117, opened 2020-05-20
Clean Up `parseRoute`

yakov116:
Follows this comment 🌮 
>What a messy merge
>
>_Originally posted by @yakov116 in https://github.com/sindresorhus/refined-github/pull/3109#issuecomment-631089893_

Also fix's `follow-file-renames` to not run when there is no path

TEST URLS:
https://github.com/sindresorhus/refined-github/commits?author=fregante

PR #3118, opened 2020-05-20
`parse-route` should not error when no branch is passed

yakov116:
Fix's `follow-file-renames` when there is no branch in the path`

1. LINKED ISSUES:
   None

2. TEST URLS:

3. SCREENSHOT:
	None

The branch will be blank, not sure if this is the correct way of handling it.


fregante:
Merge this into a single cleanup PR with #3117

yakov116:
I did that.. then backed out

Issue #2940, opened 2020-03-26
Fix `recently-pushed-branches-enhancements` positioning in responsive views


fregante:
1. Push new branch to your repo
1. Visit the Issues page, e.g. https://github.com/sindresorhus/refined-github/issues

	<img width="1054" alt="regular" src="https://user-images.githubusercontent.com/1402241/77656223-d85b3680-6f73-11ea-80af-2640be061d71.png">

3. Resize window to ~500px
4. Notice wrong position of widget

	<img width="500" alt="wrong" src="https://user-images.githubusercontent.com/1402241/77656233-dd1fea80-6f73-11ea-98a9-eeea55da4192.png">

I'd expect:
- appear next to the notifications bell
- appear as a small square that expands on hover, maybe just the green button

fregante:
This is good enough since #3130

PR #3119, opened 2020-05-20
Code cleanup

yakov116:
Continues #3083 Replaces #3117

- Moves `link-to-file-in-file-history` to `parseRoute` and run only on history pages
- Run `follow-file-renames` only on history pages
- Moves a few more init's to the bottom of the file.
- Cleaned up the typescript return falses and added a few return falses
- Finished sorting the imports
- Makes `recently-pushed-branches-enhancements` not fall off the screen #2940 (Does not fix it, just makes it not as bad)
- Moves listenToCommentFields to its own file `on-comment-field-keydown`

- Follows this comment 🌮 And cleans up the mess.

>What a messy merge
>
>_Originally posted by @yakov116 in https://github.com/sindresorhus/refined-github/pull/3109#issuecomment-631089893_

TEST URLS:
##### `follow-file-renames` should not run on

##### `link-to-file-in-file-history`
https://github.com/sindresorhus/refined-github/commits/master/source/background.ts

### DIFF
--- a/source/features/comment-fields-keyboard-shortcuts.tsx
+++ b/source/features/comment-fields-keyboard-shortcuts.tsx
@@ -2,9 +2,9 @@ import select from 'select-dom';
 import delegate from 'delegate-it';
 
 import features from '.';
-import listenToCommentFields from '../github-events/on-comment-field-edit';
+import onCommentFieldKeydown from '../github-events/on-comment-field-keydown';
 
-function init(event: delegate.Event<KeyboardEvent, HTMLTextAreaElement>): void {
+function eventHandler(event: delegate.Event<KeyboardEvent, HTMLTextAreaElement>): void {
 	const field = event.delegateTarget;
 
 	if (event.key === 'Escape') {
@@ -55,6 +55,10 @@ function init(event: delegate.Event<KeyboardEvent, HTMLTextAreaElement>): void {
 	}
 }
 
+function init(): void {
+	onCommentFieldKeydown(eventHandler);
+}
+
 features.add({
 	id: __filebasename,
 	description: 'Adds shortcuts to comment fields: `↑` to edit your previous comment; `esc` to blur field or cancel comment.',
@@ -66,7 +70,5 @@ features.add({
 }, {
 	waitForDomReady: false,
 	repeatOnAjax: false,
-	init: () => {
-		listenToCommentFields(init);
-	}
+	init
 });
--- a/source/features/comment-fields-keyboard-shortcuts.tsx
+++ b/source/features/comment-fields-keyboard-shortcuts.tsx
@@ -2,25 +2,9 @@ import select from 'select-dom';
 import delegate from 'delegate-it';
 
 import features from '.';
+import listenToCommentFields from '../github-events/on-comment-field-edit';
 
-// eslint-disable-next-line import/prefer-default-export
-export function listenToCommentFields(callback: delegate.EventHandler<KeyboardEvent, HTMLTextAreaElement>): void {
-	delegate<HTMLTextAreaElement, KeyboardEvent>(document, '.js-comment-field, #commit-description-textarea', 'keydown', event => {
-		const field = event.delegateTarget;
-
-		// Don't do anything if the autocomplete helper is shown or if non-Roman input is being used
-		if (select.exists('.suggester', field.form!) || event.isComposing) {
-			return;
-		}
-
-		callback(event);
-	}, {
-		// Adds support for `esc` key; GitHub seems to use `stopPropagation` on it
-		capture: true
-	});
-}
-
-function handler(event: delegate.Event<KeyboardEvent, HTMLTextAreaElement>): void {
+function init(event: delegate.Event<KeyboardEvent, HTMLTextAreaElement>): void {
 	const field = event.delegateTarget;
 
 	if (event.key === 'Escape') {
@@ -71,10 +55,6 @@ function handler(event: delegate.Event<KeyboardEvent, HTMLTextAreaElement>): voi
 	}
 }
 
-function init(): void {
-	listenToCommentFields(handler);
-}
-
 features.add({
 	id: __filebasename,
 	description: 'Adds shortcuts to comment fields: `↑` to edit your previous comment; `esc` to blur field or cancel comment.',
@@ -86,5 +66,7 @@ features.add({
 }, {
 	waitForDomReady: false,
 	repeatOnAjax: false,
-	init
+	init: () => {
+		listenToCommentFields(init);
+	}
 });
--- a/source/features/extend-diff-expander.tsx
+++ b/source/features/extend-diff-expander.tsx
@@ -5,11 +5,15 @@ import * as pageDetect from 'github-url-detection';
 
 import features from '.';
 
-function init(event: delegate.Event): void {
+function expandDiff(event: delegate.Event): void {
 	// Skip if the user clicked directly on the icon
 	if (!(event.target as Element).closest('.js-expand')) {
 		select<HTMLAnchorElement>('.js-expand', event.delegateTarget)!.click();
 	}
+}
+
+function init(): void {
+	delegate(document, '.diff-view .js-expandable-line', 'click', expandDiff);
 }
 
 features.add({
@@ -21,7 +25,5 @@ features.add({
 		pageDetect.isPRFiles,
 		pageDetect.isCommit
 	],
-	init: () => {
-		delegate(document, '.diff-view .js-expandable-line', 'click', init);
-	}
+	init
 });
--- a/source/features/extend-diff-expander.tsx
+++ b/source/features/extend-diff-expander.tsx
@@ -5,15 +5,11 @@ import * as pageDetect from 'github-url-detection';
 
 import features from '.';
 
-function expandDiff(event: delegate.Event): void {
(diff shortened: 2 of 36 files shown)


source/features/follow-file-renames.tsx
```diff
@@ -69,5 +69,9 @@ features.add({
 	include: [
 		pageDetect.isRepoCommitList
 	],
+	exclude: [
+		// Probably looking at the base /commits/<branch> page, not a subfolder or file.
+		() => !select.exists('.breadcrumb')
```
yakov116, line 74: Add to page-detect?
`isCommitHistoryPage` => isRepoCommitList && exists('.breadcrumb')

fregante: Too specific IMHO

fregante:
PR title doesn't match content. If you want to do further small related cleanups I can leave this PR open for a few more days.

yakov116:
@fregante OK I will push the rest to this branch I have 2 more things (I think)

yakov116:
@fregante can you look at the last 2 commits and tell me if this is something you would want?

yakov116:
Shouldn't `listenToCommentFields` be in its own file? its used in 3 places

source/features/forked-to.tsx
```diff
@@ -2,11 +2,10 @@ import './forked-to.css';
 import React from 'dom-chef';
 import cache from 'webext-storage-cache';
 import select from 'select-dom';
-import elementReady from 'element-ready';
-import * as pageDetect from 'github-url-detection';
-
 import ForkIcon from 'octicon/repo-forked.svg';
 import CheckIcon from 'octicon/check.svg';
+import elementReady from 'element-ready';
+import * as pageDetect from 'github-url-detection';
```
fregante, line 8: These were specifically split in 2 groups because this feature uses 3 icons. Reverting

source/features/swap-branches-on-compare.tsx
```diff
@@ -2,8 +2,8 @@ import React from 'dom-chef';
 import select from 'select-dom';
 import * as pageDetect from 'github-url-detection';
 
-import {wrap} from '../helpers/dom-utils';
 import features from '.';
+import {wrap} from '../helpers/dom-utils';
```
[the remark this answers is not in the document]
yakov116: Ok I thought that `.` comes before `..`

yakov116:
> Shouldn't `listenToCommentFields` be in its own file? its used in 3 places

put in file `on-comment-field-focus`
Is that a good name?

source/features/extend-diff-expander.tsx
```diff
@@ -25,5 +21,7 @@ features.add({
 		pageDetect.isPRFiles,
 		pageDetect.isCommit
 	],
-	init
+	init: () => {
+		delegate(document, '.diff-view .js-expandable-line', 'click', init);
+	}
```
yakov116, line 26: What do you say to init's that have just a simple delegate calling another function? Move them here or leave them alone?

fregante: I'm conflicted on this one. We do have other "events" in the `features.add` object, so maybe it makes sense to also have `delegate` here.

But maybe not: `delegate` is used for user events (`click`, etc) so it's part of the feature rather than the loading logic.

You can revert this.

source/features/swap-branches-on-compare.tsx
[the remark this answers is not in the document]
fregante: Don't worry about that, that kind of order is now enforced by `eslint` and it only splits `node_modules` and local modules.

fregante:
> put in file `on-comment-field-focus`
> Is that a good name?

Does the event listen to focus? no

yakov116:
> > put in file `on-comment-field-focus`
> > Is that a good name?
> 
> Does the event listen to focus? no

Sorry I wrote one name in the comment and used another name.

`on-comment-field-edit` is what I used

yakov116:
OK I am done. The only thing that I think should be in its own PR is the switching `link-to-file-in-file-history` to parse route.
(I will update the first post)

Side Note: Since Covid is just about over I am back to regular work schedule next week, so I will be much less active. It was great being able to contribute! 
Thanks @fregante!!

source/github-events/on-comment-field-edit.tsx
```diff
@@ -0,0 +1,18 @@
+import select from 'select-dom';
+import delegate from 'delegate-it';
+
+export default function listenToCommentFields(callback: delegate.EventHandler<KeyboardEvent, HTMLTextAreaElement>): void {
+	delegate<HTMLTextAreaElement, KeyboardEvent>(document, '.js-comment-field, #commit-description-textarea', 'keydown', event => {
+		const field = event.delegateTarget;
+
+		// Don't do anything if the autocomplete helper is shown or if non-Roman input is being used
```
notlmn, line 8: By **_"non-roman"_** don't you mean **_"non-latin"_**? Or am I missing something?

source/features/comment-fields-keyboard-shortcuts.tsx
```diff
@@ -2,25 +2,9 @@ import select from 'select-dom';
 import delegate from 'delegate-it';
 
 import features from '.';
+import listenToCommentFields from '../github-events/on-comment-field-edit';
```
fregante, line 5: Can you rename the variable to match the filename?

source/features/comment-fields-keyboard-shortcuts.tsx
```diff
@@ -86,5 +66,7 @@ features.add({
 }, {
 	waitForDomReady: false,
 	repeatOnAjax: false,
-	init
+	init: () => {
+		listenToCommentFields(init);
+	}
```
fregante, line 71: Avoid doing this for any user action, as suggested here https://github.com/sindresorhus/refined-github/pull/3119#discussion_r428759134

Key/click listeners should not be named `init`. Revert here and other related places (including `keyup`)

fregante:
> The only thing that I think should be in its own PR is the switching link-to-file-in-file-history to parse route.

No need

source/features/link-to-file-in-file-history.tsx
```diff
@@ -40,5 +32,9 @@ features.add({
 	include: [
 		pageDetect.isRepoCommitList
 	],
+	exclude: [
+		// Probably looking at the base /commits/<branch> page, not a subfolder or file.
+		() => !select.exists('.breadcrumb')
```
fregante, line 37: I think this can now be determined via parseRoute. “If filePath === '', then this feature, like others, shouldn’t run”

source/github-events/on-comment-field-edit.tsx
fregante: [Good catch!](https://en.wikipedia.org/wiki/Keyboard_layout)

yakov116:
@fregante Done.

What do you say to [https://github.com/sindresorhus/refined-github/pull/3119/commits/f6c10140b57ac8156c2580341122042da30ed142 ](https://github.com/sindresorhus/refined-github/pull/3119/commits/f6c10140b57ac8156c2580341122042da30ed142)

yakov116:
Dont merge yet may have one more.

yakov116:
I had fun with https://github.com/sindresorhus/refined-github/pull/3119/commits/e3e40ee59d4d8eb620c524b8d07dce1c3225c5f5 just revert if its no good.

I'm really done now:)

fregante:
toString is the standard method to convert objects to strings. It also works as `String(parseRoute(…))`

yakov116:
I'll revert that then.

source/features/follow-file-renames.tsx
```diff
@@ -67,7 +71,8 @@ features.add({
 	screenshot: 'https://user-images.githubusercontent.com/1402241/54799957-7306a280-4c9a-11e9-86de-b9764ed93397.png'
 }, {
 	include: [
-		pageDetect.isRepoCommitList
+		// Only run on history pages
+		() => pageDetect.isRepoCommitList() && parseRoute(location.pathname).filePath.length > 0
```
fregante, line 75: This parseRoute is duplicate, it’s already part of the init

source/features/link-to-file-in-file-history.tsx
```diff
@@ -38,7 +30,8 @@ features.add({
 	screenshot: 'https://user-images.githubusercontent.com/22439276/57195061-b88ddf00-6f6b-11e9-8ad9-13225d09266d.png'
 }, {
 	include: [
-		pageDetect.isRepoCommitList
+		// Only run on history pages
+		() => pageDetect.isRepoCommitList() && parseRoute(location.pathname).filePath.length > 0
```
fregante, line 34: Same

source/features/follow-file-renames.tsx
yakov116: What do you mean I should do a return false instead of the inside the include?

fregante: Right

yakov116: Done

fregante:
It's best to keep bugfixes separate from "no-op" code style changes. You can send the reverted commits in other PRs:

- One for slashes in branchname support
- One for "minor bug fixes" (recently-pushed-branches-enhancements, show-names)

Issue #3001, opened 2020-04-13
Bugs and Issues tabs are selected at the same time after the Back button


fregante:
1. You're on this issue (it works with any issue, not just bugs)
1. Click the repository’s Code tab
1. Go back in the browser (`history.back()`)

<img width="336" alt="bugs-tab-bug" src="https://user-images.githubusercontent.com/1402241/79167277-68f49c00-7de7-11ea-854a-e26210c0438e.png">

Semi-related possible enhancement: the bugs tab should be selected when opening a "bug" issue like this one

yakov116:
And here https://github.com/sindresorhus/refined-github/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc+label%3Abug

PR #3130, opened 2020-05-24
Minor bug fixes

yakov116:
Continues #3119

- Reuse filePath var in `revert-file`
- Makes recently-pushed-branches-enhancements not fall off the screen #2940 (Does not fix it, just makes it not as bad)
- Don't run `show-names` on app bots (Revert's the change I made) _Fixes https://github.com/sindresorhus/refined-github/issues/3133_
- Use `getRepoURL` instead of `getOwnerandRepo` for `add-tag-to-commits`
- Don't highlight bugs tab when on the PR tab (Continues fixing #3001)
- Run `esc-to-deselect-line` only on code pages
- Don't error `selection-in-new-tab` if there is no element to mark unread (in RepoRoot for example)

### DIFF
--- a/source/features/bugs-tab.tsx
+++ b/source/features/bugs-tab.tsx
@@ -71,7 +71,8 @@ async function init(): Promise<void | false> {
 	new SearchQuery(bugsLink).add('label:bug');
 
 	// Change the Selected tab if necessary
-	bugsLink.classList.toggle('selected', isBugsPage);
+	const isPRTab = new SearchQuery(location).includes('is:pr');
+	bugsLink.classList.toggle('selected', isBugsPage && !isPRTab);
 	select('.selected', issuesTab)?.classList.toggle('selected', !isBugsPage);
 
 	issuesTab.after(bugsTab);
--- a/source/features/bugs-tab.tsx
+++ b/source/features/bugs-tab.tsx
@@ -71,8 +71,7 @@ async function init(): Promise<void | false> {
 	new SearchQuery(bugsLink).add('label:bug');
 
 	// Change the Selected tab if necessary
-	const isPRTab = new SearchQuery(location).includes('is:pr');
-	bugsLink.classList.toggle('selected', isBugsPage && !isPRTab);
+	bugsLink.classList.toggle('selected', isBugsPage && !pageDetect.isPRList());
 	select('.selected', issuesTab)?.classList.toggle('selected', !isBugsPage);
 
 	issuesTab.after(bugsTab);
--- a/source/features/revert-file.tsx
+++ b/source/features/revert-file.tsx
@@ -31,8 +31,7 @@ const getBaseReference = onetime(async (): Promise<string> => {
 	return repository.pullRequest.baseRefOid;
 });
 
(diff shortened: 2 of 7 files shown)


yakov116:
@fregante this is ready.

source/features/bugs-tab.tsx
```diff
@@ -71,7 +71,8 @@ async function init(): Promise<void | false> {
 	new SearchQuery(bugsLink).add('label:bug');
 
 	// Change the Selected tab if necessary
-	bugsLink.classList.toggle('selected', isBugsPage);
+	const isPRTab = new SearchQuery(location).includes('is:pr');
+	bugsLink.classList.toggle('selected', isBugsPage && !isPRTab);
```
fregante, line 75: There's already a `isPRList`

yakov116: Changed

source/features/revert-file.tsx
```diff
@@ -54,12 +53,11 @@ async function deleteFile(menuItem: Element): Promise<void> {
 	await postForm(form!);
 }
 
-async function commitFileContent(menuItem: Element, content: string): Promise<void> {
+async function commitFileContent(menuItem: Element, content: string, filePath: string): Promise<void> {
 	let {pathname} = menuItem.previousElementSibling as HTMLAnchorElement;
 	// Check if file was deleted by PR
 	if (menuItem.closest('[data-file-deleted="true"]')) {
 		menuItem.textContent = 'Undeleting…';
-		const filePath = pathname.split('/')[5]; // The URL was something like /$user/$repo/blob/$startingCommit/$path
```
fregante, line 62: Why not just change this to

```ts
const filePath = menuItem.closest<HTMLDivElement>('[data-path]')!.dataset.path!;
```

yakov116: You have it already on line 91. its used 2 times, why not use the same var again?

yakov116:
@fregante should't features that use onCommentFieldKeydown be limited to `pageDetect.hasRichTextEditor`?

source/features/revert-file.tsx
fregante: What I’m suggesting is that you’re creating the variable in another function and then passing it around and just using it here. Right? So why not avoid that?

fregante: Oh I noticed it’s used in 2 places. Looks good.

fregante:
The rich field editor appears on pages other than comments, the two are unrelated.

yakov116:
I just checked all the features that have onCommentFieldKeydown are correctly running on `hasRichTextEditor` (It add's new issue and isCompare).
Never mind lets leave it out of this PR for now

source/features/selection-in-new-tab.tsx
```diff
@@ -12,7 +12,7 @@ function init(): void {
 			});
 
 			// Get the list element that contains the unread class and mark it as read.
-			selected.closest('li')!.classList.replace('unread', 'read');
+			selected.closest('li')?.classList.replace('unread', 'read');
```
fregante, line 15:
```suggestion
			selected.closest('.unread')?.classList.replace('unread', 'read');
```

PR #3131, opened 2020-05-24
Ensure features work on branches/tags with slashes /2

yakov116:
Continues #3083 

- Moves `link-to-file-in-file-history` to `parseRoute` and run only on history pages
- Run `follow-file-renames` only on history pages

TEST URLS:
##### `follow-file-renames` should not run on

##### `link-to-file-in-file-history`
https://github.com/sindresorhus/refined-github/commits/master/source/background.ts

### DIFF
--- a/source/features/follow-file-renames.tsx
+++ b/source/features/follow-file-renames.tsx
@@ -24,12 +24,14 @@ async function findRename(
 
 function init(): false | void {
 	const disabledPagination = select.all('.paginate-container [disabled], .paginate-container .disabled');
+	const parts = parseRoute(location.pathname);
 
-	if (disabledPagination.length === 0) {
+	if (
+		disabledPagination.length === 0 ||
+		parts.filePath.length === 0 // Not a history page
+	) {
 		return false;
 	}
-
-	const {user, repository, branch: reference, filePath: currentFilename} = parseRoute(location.pathname);
 
 	disabledPagination.forEach(async button => {
 		const isNewer = button.textContent === 'Newer';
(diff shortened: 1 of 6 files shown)


source/features/follow-file-renames.tsx
```diff
@@ -5,45 +5,48 @@ import * as pageDetect from 'github-url-detection';
...
 }
 
 function init(): false | void {
 	const disabledPagination = select.all('.paginate-container [disabled], .paginate-container .disabled');
+	const parts = parseRoute(location.pathname);
 
-	if (disabledPagination.length === 0) {
+	if (
+		disabledPagination.length === 0 ||
+		parts.filePath.length === 0 // Not a history page
```
fregante, line 28: It still is a history page.
```suggestion
		!parts.filePath
```

source/features/follow-file-renames.tsx
```diff
@@ -5,45 +5,48 @@ import * as pageDetect from 'github-url-detection';
...
+async function findRename(lastCommitOnPage: string): Promise<File[]> {
 	// API v4 doesn't support it: https://github.community/t5/GitHub-API-Development-and/What-is-the-corresponding-object-in-GraphQL-API-v4-for-patch/m-p/14502?collapse_discussion=true&filter=location&location=board:api&q=files%20changed%20commit&search_type=thread
-	const {files} = await api.v3(`repos/${user}/${repo}/commits/${lastCommitOnPage}`);
+	const {files} = await api.v3(`repos/${getRepoURL()}/commits/${lastCommitOnPage}`);
 	return files;
 }
 
 function init(): false | void {
 	const disabledPagination = select.all('.paginate-container [disabled], .paginate-container .disabled');
+	const parts = parseRoute(location.pathname);
```
fregante, line 24: Let's use consistent naming everywhere parseRoute is used
```suggestion
	const pathnameParts = parseRoute(location.pathname);
```

fregante: > Everywhere

yakov116: >That's right! I'd name the variables just `path` or now instead of `pathnameParts`
>
>So you can use them as `path.branch`, `path.user`, etc
>
>_Originally posted by @fregante in https://github.com/sindresorhus/refined-github/pull/3109#issuecomment-631038563_

fregante:
# 🖐️

fregante:
Adding a link to 8a9462f66a19eac17bf0543b5c91c1abc574d236 for a possible future API improvement, but it might not be necessary anymore

source/features/follow-file-renames.tsx
```diff
@@ -5,45 +5,45 @@ import * as pageDetect from 'github-url-detection';
...
 	filename: string;
 	status: string;
 }
 
-async function findRename(
-	user: string,
-	repo: string,
-	lastCommitOnPage: string
-): Promise<File[]> {
+async function findRename(lastCommitOnPage: string): Promise<File[]> {
```
yakov116, line 16: Why is there no cache here?

fregante: Because it's so specific to the current file and commit that seems too rare to cache.

PR #3135, opened 2020-05-25
Only show `comments-time-machine-links` after visiting the link

fregante:
Fixes https://github.com/sindresorhus/refined-github/issues/1953
Closes https://github.com/sindresorhus/refined-github/issues/2618
Related to https://github.com/sindresorhus/refined-github/issues/2131
Ancestor of https://github.com/sindresorhus/refined-github/pull/1863
Prophecy of https://github.com/sindresorhus/refined-github/issues/1448 🪁

# Test

To test this feature, click the test links in https://github.com/sindresorhus/refined-github/issues/2618 (the feature depends on comment date)

Also here are two pre-edited links to see what they look like:

```
Still exist: https://github.com/sindresorhus/refined-github/blob/master/source/features/comments-time-machine-links.tsx?rgh-link-date=2019-12-11T10%3A12%3A11Z
File deleted: https://github.com/sindresorhus/refined-github/blob/master/source/features/default-to-rich-diff.tsx?rgh-link-date=2019-12-11T10%3A12%3A11Z
```

# Screenshots

If the file still exists, we can also support https://github.com/sindresorhus/refined-github/issues/1953 by getting the branch name from the page

If the file is deleted, the feature will link to HEAD and hope that the file exists there.

PR #3137, opened 2020-05-25
Support timestamps in `parseRoute`

yakov116:
See https://github.com/sindresorhus/refined-github/pull/3135#issuecomment-633622883
Test URL's
https://github.com/sindresorhus/refined-github/blob/master@%7B2019-04-20T03:27:17Z%7D/license
https://github.com/sindresorhus/refined-github/blob/master@%7B2019-12-11T10:12:11Z%7D/source/features/comments-time-machine-links.tsx
https://github.com/sindresorhus/eslint-plugin-unicorn/tree/HEAD@%7B2020-05-24T16:33:08Z%7D

@fregante I dont have much time. (I did this very quickly)

### DIFF
--- a/source/github-helpers/parse-route.ts
+++ b/source/github-helpers/parse-route.ts
@@ -11,7 +11,7 @@ interface Pathname {
 
 const timeStampRegex = /.*@{(\d{4}([.\-/ ])\d{2}\2\d{2}T\d\d:\d\d:\d\dZ)}/;
 function isTimeStamp(part1: string): boolean {
-	return Boolean(timeStampRegex.exec(decodeURIComponent(part1)));
+	return timeStampRegex.test(decodeURIComponent(part1));
 }
 
 export default function parseRoute(pathname: string): Pathname {
--- a/source/github-helpers/parse-route.ts
+++ b/source/github-helpers/parse-route.ts
@@ -8,10 +8,17 @@ interface Pathname {
 	filePath: string;
 	toString: () => string;
 }
+
+const timeStampRegex = /.*@{(\d{4}([.\-/ ])\d{2}\2\d{2}T\d\d:\d\d:\d\dZ)}/;
+function isTimeStamp(part1: string): boolean {
+	return Boolean(timeStampRegex.exec(decodeURIComponent(part1)));
+}
+
 export default function parseRoute(pathname: string): Pathname {
 	const [user, repository, route, ...next] = pathname.replace(/^\/|\/$/g, '').split('/');
-	const parts = next.join('/');
 	const branch = getCurrentBranch();
+	next[0] = isTimeStamp(next[0]) ? branch + '/' : next[0];
+	const parts = next.join('/');
 	if (parts !== branch && !parts.startsWith(branch + '/')) {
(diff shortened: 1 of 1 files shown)


yakov116:
@fregante I didn't parse (pun intended) what you are saying.

fregante:
There are many more formats supported by @: https://linux.die.net/man/1/git-rev-parse

What’s the issue in this case specifically?

yakov116:
The see on default branch and latest tag will both fail on a time link.

Maybe just let it error out?

fregante:
Neither. The problem is that the URL doesn't match the meta tag (`getCurrentBranch`) so the solution is to just skip this check and assume that the pathname passed matches the page.

source/github-helpers/parse-route.ts
```diff
@@ -8,10 +8,17 @@ interface Pathname {
...
+	return timeStampRegex.test(decodeURIComponent(part1));
+}
+
 export default function parseRoute(pathname: string): Pathname {
 	const [user, repository, route, ...next] = pathname.replace(/^\/|\/$/g, '').split('/');
-	const parts = next.join('/');
 	const branch = getCurrentBranch();
+	next[0] = isTimeStamp(next[0]) ? branch + '/' : next[0];
+	const parts = next.join('/');
 	if (parts !== branch && !parts.startsWith(branch + '/')) {
```
fregante, line 22: Maybe 
```suggestion
	if (!isTimeStamp(next[0]) && parts !== branch && !parts.startsWith(branch + '/')) {
```

fregante:
More test URLs:

https://github.com/sindresorhus/refined-github/blob/master@%7B1500%7D/package.json
https://github.com/sindresorhus/refined-github/blob/master@%7Byesterday%7D/package.json

Really there are a lot more situations where this might not work:

https://github.com/yakov116/refined-github/blob/master%5E/package.json
https://github.com/yakov116/refined-github/blob/master~1500/package.json

source/github-helpers/parse-route.ts
yakov116: That will only work if there is no file name.
https://github.com/sindresorhus/eslint-plugin-unicorn/tree/HEAD@%7B2020-05-24T16:33:08Z%7D

But the following will not
https://github.com/sindresorhus/refined-github/blob/master@%7B2019-04-20T03:27:17Z%7D/license
https://github.com/sindresorhus/refined-github/blob/master@%7B2019-12-11T10:12:11Z%7D/source/features/comments-time-machine-links.tsx

yakov116:
100% correct

source/github-helpers/parse-route.ts
fregante: Just this

```suggestion
	// Exclude extended revisions from being tested #3137 https://git-scm.com/docs/git-rev-parse#_specifying_revisions
	if (parts !== branch && !parts.startsWith(branch + '/') && !/\^|~|@{/.test(next[0])) {
```

fregante:
Wait I'll rewrite this function to be less strict

yakov116:
Thanks. ❤️

PR #3138, opened 2020-05-25
Meta: relax `parseRoute` and turn it into a more-easily usable class

fregante:
Follows https://github.com/sindresorhus/refined-github/pull/3131
Replaces and fixes #3137


yakov116:
@fregante your awesome!

Have bad new.

https://github.com/yakov116/refined-github/commits/f23b687b3b89aa95a76193722cdfeff740646670?after=f23b687b3b89aa95a76193722cdfeff740646670+34&path%5B%5D=source&path%5B%5D=features&path%5B%5D=release-download-count.tsx

On a history page once you go to the next page, the path is in the search :(

fregante:
Unrelated again xD can you open a new issue or PR?

yakov116:
So sorry. (lesson learned dont work and test RG at the same time)
Yes. I just did not want you to change everything all over again.

fregante:
Can you find the commit that broke this? I feel like this was working and shouldn't necessarily be part of `ObjectPath`

yakov116:
https://github.com/sindresorhus/refined-github/pull/3131

We used to get the filename from the dom
But keep in mind that `follow-file-renames` never worked

yakov116:
@fregante if you want me to open a new issue let me know and we can discuss it there.

I thought this over we really have 4 options.

1. Go back to using the dom. 
	- Cant use `waitForDomReady:false` for `default-branch button`
	- Requires much more code per feature

2. Keep it the way it is and in each feature do
	- `path.filePath = path.filePath ||  filePathFromSearch()`
	- Requires more code per feature and is not so clean

3. Allow `ObjectPath` to take an option custom filePath and pass it in each feature
	- Does not require much code per feature and is clean

4. `ObjectPath` should use the search directly -> https://github.com/sindresorhus/refined-github/commit/6407380a37212a20c218ecc0b1885866514bde70
	- All existing code stays the same

fregante:
How about this: `ObjectPath` is meant to parse the specified URL with the help of the DOM exclusively to extract the current branch, **if** it has slashes. It does not and should not depend on global state (DOM or `location.search` as suggested in https://github.com/sindresorhus/refined-github/commit/6407380a37212a20c218ecc0b1885866514bde70)

Even then, I'm still conflicted about using the global state _sometimes_

Also perhaps it should parse the whole URL rather than just the `pathname`.

fregante:
@yakov116 Regarding issues specific to `follow-file-renames`, open an issue and/or a PR for them after this PR is merged.

PR #3147, opened 2020-05-27
Use Octicons v2

fregante:
GitHub.com just switched to the [new icons](https://primer.style/octicons-v2/) even though they're not "stable" on npm for use to use. However I want to update RGH asap to _preserve the illusion_ and make any improvements later

There are 2 issues:

- bug.svg no longer exists: https://github.com/primer/octicons/issues/430
  I avoided this issue by keeping the older v1 around just for this icon.
- every filename now ends with `-16.svg`.
  I avoided this issue by using the last "v2" version without this requirement. This will have to be fixed once they're "stable"

### DIFF
--- a/octicon-svg-loader.ts
+++ b/octicon-svg-loader.ts
@@ -9,5 +9,5 @@ export default function svgLoader(this: webpack.loader.LoaderContext, source: st
 	);
 	return `
 	import doma from 'doma';
-	export default () => doma.one('${svgWithClass.replace('\'', '\\\'')}')`;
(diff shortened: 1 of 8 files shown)


fregante:
# Before / After

yakov116:
@fregante https://github.com/primer/octicons/releases/tag/v10.0.0 🎉

fregante:
It might be worth trying to use the React package now, it looks like they dropped a lot of cruft and is likely treeshakeable https://www.unpkg.com/browse/@primer/octicons-react@10.0.0/dist/index.esm.js

yakov116:
Should I open an issue here to see if someone wants to take it? I am not able to do it.

PR #3150, opened 2020-05-28
Reset vertical-align for new octicons in js-issue-row

fspoettel:
Hi,
fixes a misaligned icon I noticed after the recent `octicons` update. Not sure if this is the best way to go about it, feel free to close in case there is a better way to achieve this.

Before:

Screenshot from 2020-05-28 10-07-48

After:

Screenshot from 2020-05-28 10-31-13

Thanks for maintaining this extension! :sunny:

### DIFF
--- a/source/refined-github.css
+++ b/source/refined-github.css
@@ -116,3 +116,8 @@ pr-branches
 .js-issue-row .text-small.text-gray {
 	line-height: 1.8;
 }
+
+/* Reset `vertical-align` for .octicons to work with above increased line-height */
+.js-issue-row .text-small.text-gray .octicon {
+	vertical-align: middle !important;
+}
--- a/source/refined-github.css
+++ b/source/refined-github.css
@@ -118,6 +118,6 @@ pr-branches
 }
 
 /* Reset `vertical-align` for .octicons to work with above increased line-height */
-.issue-meta-section .octicon {
+.js-issue-row .text-small.text-gray .octicon {
 	vertical-align: middle;
 }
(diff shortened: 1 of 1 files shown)


fregante:
Thanks for the PR, but if this happens without Refined GitHub (which apparently does) we shouldn't fix it since they're probably still working on it.

fspoettel:
@fregante Thanks for taking a look.

It does not look off without Refined Github for me, this is what it looks like in that case:

Screenshot from 2020-05-28 11-09-35

The `line-height: 1.8` rule in combination with Github `vertical-align: bottom` causes the misalignment

fregante:
You're absolutely right!

source/refined-github.css
```diff
@@ -116,3 +116,8 @@ pr-branches
 .js-issue-row .text-small.text-gray {
 	line-height: 1.8;
 }
+
+/* Reset `vertical-align` for .octicons to work with above increased line-height */
+.js-issue-row .text-small.text-gray .octicon {
+	vertical-align: middle !important;
+}
```
fregante, line 123: Let's match GitHub’s own selector for the alignment. Also `!important` doesn't seem to be needed. Can you double-check? 

```suggestion
.issue-meta-section .octicon {
	vertical-align: middle;
}
```

fspoettel: Your suggestion makes sense in my books, `stylelint` disagrees :stuck_out_tongue:. OK if I disable it for the line?

fregante: I just saw why you chose that selector in the first place. Feel free to restore `.js-issue-row .text-small.text-gray .octicon` to fix the issue :D

fspoettel: Ok, reverted it but removed `important` as it was indeed not needed. Should be all good now

PR #3111, opened 2020-05-19
Fix `fix-view-file-link-in-pr` and ignore deleted files

yakov116:
1. LINKED ISSUES:
   Fixes #3108 

2. TEST URLS:
	Deleted File
   https://github.com/yakov116/TestR/pull/21/files
	Local File (Not deleted)
https://github.com/sindresorhus/refined-github/pull/3106/files
	None Local Files (Not deleted)
https://github.com/sindresorhus/refined-github/pull/3103/files

3. SCREENSHOT:
	None

I broke the feature (when tested it, I was testing on my own repo)

PR #3152, opened 2020-05-28
Let users view/edit/delete files when comparing branches

### LINKED ISSUE #2704: Make "Edit/delete file" clickable on compare page
Example URL:
https://github.com/Cog-Creators/Red-DiscordBot/compare/V3/develop...jack1142:pr/ghactions?expand=1

"Edit/delete file" links are already there, but they're not active, IMO it would be useful to be able to quickly edit files from there if you notice some error in files:
image

Of course that link should only be active when you can edit the head branch.

Edit: I also noticed that "View file" link there doesn't link you to branch but to commit so that makes it take even longer to edit a file.

### LINKED ISSUE #3123: "View file" is broken when using RefinedGithub
As a reproduction case, click `View file` in the dropdown here:
https://github.com/yuzu-emu/yuzu/pull/3976/files

This causes me to end up on 
https://github.com/yuzu-emu/yuzu/blob/qdarkstyle_fix_prop/dist/qt_themes/qdarkstyle/style.qss whereas it should lead me to 
https://github.com/yuzu-emu/yuzu/blob/ea14af2164934faa0761a88c6236b3652d8cdaea/dist/qt_themes/qdarkstyle/style.qss.

This issue doesn't occur when the plugin is not installed.

yakov116:
1. LINKED ISSUES:
   Closes #2704 
	Fixes #3123

2. TEST URLS:
	The following links are links I tested on. I dont know what forks you have write access to.
	##### Compare (The edit and delete will not work since you dont have rights)
	https://github.com/yakov116/refined-github/compare/123...sindresorhus:improve-build
	##### Permalink and random forks (The edit and delete will not work since you dont have rights or its a Permalink)
	Permalink: https://github.com/yakov116/refined-github/compare/123...sindresorhus:improve-build
Same repo Permalink: https://github.com/sindresorhus/refined-github/compare/a@b...20.5.27

	Random fork: https://github.com/yakov116/refined-github/compare/upstream...FloEdelmann:faster-reviews
	

3. SCREENSHOT:
Before:
image
After
image

When reviewing please look at each commit, the rename makes it look like the whole thing changed


fregante:
This probably closed because I deleted my branch. 

Change the base so I can delete it again

source/features/enable-file-links-in-compare-view.tsx
```diff
@@ -0,0 +1,120 @@
...
+			href={String(url)}
+			role="menuitem"
+			className="pl-5 dropdown-item btn-link"
+			rel="nofollow"
+			aria-label="Change this file using the online editor"
+			data-ga-click="Edit file, click, location:files_changed_dropdown"
+		>
+			Edit file
+		</a>
+	);
```
fregante, line 83: Could this be replaced with:

```suggestion
	const editFile = viewFile.cloneNode(true);
	editFile.textContent = 'Edit file';
	editFile.removeAttribute('data-ga-click');
	editFile.href = url.assign({route: 'edit'}).toString();
```

Also no fake `ga` events please. It's not our business

source/features/enable-file-links-in-compare-view.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+	const viewFile = select<HTMLAnchorElement>('[data-ga-click^="View file"]', dropdown)!;
+	const url = new GitHubURL(viewFile.href);
+	url.assign({
+		branch: location.pathname.replace(/.*:/, '')
+	});
+	viewFile.href = String(url);
+	viewFile.classList.add('rgh-actionable-link'); // Mark this as processed
+
+	// If the allow maintainers to edit is not available your either comparing a Permalink and the file can't be edited or your comparing a fork you can't edit the file on
```
fregante, line 62: Reword or fix errors when you have a chance

source/features/enable-file-links-in-compare-view.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+	// Only if it's not already there
+	if (select.exists('.rgh-actionable-link', dropdown)) {
+		return;
+	}
+
+	const viewFile = select<HTMLAnchorElement>('[data-ga-click^="View file"]', dropdown)!;
+	const url = new GitHubURL(viewFile.href);
+	url.assign({
+		branch: location.pathname.replace(/.*:/, '')
```
fregante, line 57:
```suggestion
		branch: location.pathname.replace(/.+:/, '')
```

source/features/enable-file-links-in-compare-view.tsx
```diff
@@ -0,0 +1,120 @@
...
+	// - PRs opened from the default branch
+	const viewFile = select<HTMLAnchorElement>('[data-ga-click^="View file"]', dropdown)!;
+	const url = new GitHubURL(viewFile.href);
+	const [user, repository, ...branchParts] = select<HTMLAnchorElement>('.commit-ref.head-ref a')!.title.replace(':', '/').split('/');
+	const branch = branchParts.join('/');
+	url.assign({
+		user,
+		repository,
+		branch
+	});
```
fregante, line 39: This code doesn't seem shorter than the previous one. It can restored

source/features/enable-file-links-in-compare-view.tsx
yakov116: Ok I will try to explain.
When I am doing here is making the edit and delete buttons work. However if you are comparing a Permalink you cant edit the file on that nor delete it. 
Another case where you cant edit the file would be if you dont have right to edit the repo.

I looked through the dom and I saw that when creating a PR the `allow maintainers to edit` is not there in those cases.
However my solution will fail when creating a PR on your own repo.

source/features/enable-file-links-in-compare-view.tsx
yakov116: Is this better?

source/features/enable-file-links-in-compare-view.tsx
fregante: In the code, not as a comment.

source/features/enable-file-links-in-compare-view.tsx
```diff
@@ -0,0 +1,98 @@
...
+		repository,
+		branch: getCurrentBranch() // Ensures that the branch name is attached even when it links to the default branch
+	});
+
+	viewFile.href = String(url);
+
+	viewFile.classList.add('rgh-actionable-link'); // Mark this as processed
+}
+
+function isCompareMenuOpening(event: delegate.Event): void {
```
fregante, line 45: This is not a Boolean function, it should start with `is`

source/features/enable-file-links-in-compare-view.tsx
```diff
@@ -0,0 +1,98 @@
...
+	if (!select.exists('[name="collab_privs"], .js-issue-sidebar-form')) {
+		return;
+	}
+
+	// Fix the edit link
+	const editFile = viewFile.cloneNode(true);
+	editFile.textContent = 'Edit file';
+	editFile.removeAttribute('data-ga-click');
+	editFile.href = url.assign({route: 'edit'}).toString();
+	viewFile.nextElementSibling!.replaceWith(editFile);
```
fregante, line 71: Not super safe. It’s best to use
```suggestion
	select('edit button', dropdown)!.replaceWith(editFile);
```
Also for the delete link

source/features/enable-file-links-in-compare-view.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+	// Only if it's not already there
+	if (select.exists('.rgh-actionable-link', dropdown)) {
+		return;
+	}
+
+	const viewFile = select<HTMLAnchorElement>('[data-ga-click^="View file"]', dropdown)!;
+	const url = new GitHubURL(viewFile.href);
+	url.assign({
+		branch: location.pathname.replace(/.+:/, '')
```
fregante, line 56: How does this work on this page? There's no `:`

```
https://github.com/sindresorhus/refined-github/compare/a@b...20.5.27
```

source/features/enable-file-links-in-compare-view.tsx
```diff
@@ -0,0 +1,98 @@
...
+	});
+	viewFile.href = String(url);
+	viewFile.classList.add('rgh-actionable-link'); // Mark this as processed
```
fregante, line 59: Let's move this class to `event.delegateTarget` itself, so `delegate` will skip it altogether with 

```js
delegate(document, '.file-header:not([data-file-deleted="true"]) .js-file-header-dropdown > summary:not(.rgh-actionable-link)', 'click', handleMenuOpening);
```

source/features/enable-file-links-in-compare-view.tsx
yakov116: I am going to revert the whole fix for this here. I will open a new PR for it.

source/features/enable-file-links-in-compare-view.tsx
yakov116: The problem is the there is nothing unique to select, the only difference between the edit and the delete is the text content

source/features/enable-file-links-in-compare-view.tsx
```diff
@@ -0,0 +1,82 @@
+import select from 'select-dom';
+import delegate from 'delegate-it';
+import * as pageDetect from 'github-url-detection';
+
+import features from '.';
+import GitHubURL from '../github-helpers/github-url';
+
+function handleIsPRMenuOpening(event: delegate.Event): void {
```
fregante, line 8: Why `is`? It does not belong here

yakov116: Woops, That was a mistake.

source/features/enable-file-links-in-compare-view.tsx
```diff
@@ -0,0 +1,82 @@
...
+	});
+	viewFile.href = String(url);
+
+	// Dont replace the edit and delete buttons if its not possible
+	if (!select.exists([
+		'[name="collab_privs"]', // Excludes repo's that you dont have rights to edit and Permalink's (the whole create pull request will not exist on a Permalink)
+		'.js-issue-sidebar-form' // Dont exclude same repo compare pages you have rights to by detecting if the review's/labels exists
+	])) {
+		return;
+	}
```
fregante, line 49: Doesn't this apply to View links as well?
The feature "fixes" the View link by making Edit/Delete possible on the View page.
If Edit/Delete isn't possible here, it won't be possible on the next page either.

I think this condition should be moved to `features.add`’s `exclude`. I think this also applies to the 2 conditions in `handleIsPRMenuOpening`

Also please improve your comments, there are a lot of typos (It's `repos` or `repositories`, not `repo's`). The lack of quotation marks, capitalization and consistency also make them hard to read.

source/features/enable-file-links-in-compare-view.tsx
```diff
@@ -0,0 +1,83 @@
...
+	init
+}, {
+	include: [
+		pageDetect.isCompare
+	],
+	exclude: [
+		// If you cant create a PR you are probably looking at a permalink.
+		() => !select.exists('#new_pull_request'),
+		// Exclude repositories you don't have access to.
+		() => !select.exists([
```
fregante, line 76: Can be merged

source/features/enable-file-links-in-compare-view.tsx
```diff
@@ -0,0 +1,83 @@
...
+}
+
+function handleCompareMenuOpening(event: delegate.Event): void {
+	event.delegateTarget.classList.add('rgh-actionable-link'); // Mark this as processed
+
+	const dropdown = event.delegateTarget.nextElementSibling!;
+	const viewFile = select<HTMLAnchorElement>('[data-ga-click^="View file"]', dropdown)!;
+	const url = new GitHubURL(viewFile.href);
+	url.assign({
+		branch: location.pathname.replace(/.+:|.+\.{3}/, '')
```
fregante, line 28: Can this be safely replaced with 
```suggestion
		branch: select('[title^="compare"]')!.title.split(' ').pop()
```

It feels like the regex is hard to understand without a comment

source/features/enable-file-links-in-compare-view.tsx
yakov116: The truth is that the first one is not needed at all I just did not know how to word it correctly without putting it into 2.
If `[name="collab_privs"]` does not exist then `'#new_pull_request'` does not exist.

I will give it a shot, I will try to be as clear as can be.

source/features/enable-file-links-in-compare-view.tsx
yakov116: Is there a reason we cant just take the textContent?

fregante: Yep, works too! I didn't check.

source/features/enable-file-links-in-compare-view.tsx
fregante: What if it just looks for the "Create PR" button if that's the logic you're following? Could that work?

e.g. if `.existing-pull-button, [data-ga-click*="text:Create pull request"]` exists, then the links can be _fixed_

- ✅ FIX: PR exists: https://github.com/sindresorhus/refined-github/compare/master...yakov116:use-filePathFromSearch
- ✅ FIX: Can create PR: https://github.com/sindresorhus/refined-github/compare/a@b...yakov116:use-filePathFromSearch
- ❌ LEAVE: Cannot create PR: https://github.com/sindresorhus/refined-github/compare/a@b...yakov116:123

yakov116: That's fine with me. I though you wanted more of a detection if you are able to edit it too.

yakov116:
Keep note that
https://github.com/sindresorhus/refined-github/compare/master...FloEdelmann:faster-reviews will also have the edit, since there is a PR that exists.

fregante:
Why are you picking that? Of course it has and edit button, it’s a branch.

source/features/enable-file-links-in-compare-view.tsx
fregante: Edit buttons appear everywhere on the site whether you can edit it directly or you have to fork it first.

yakov116: Great I never took me one of that

yakov116:
image

source/features/enable-file-links-in-compare-view.tsx
```diff
@@ -0,0 +1,81 @@
...
+}
+
+function handleCompareMenuOpening(event: delegate.Event): void {
+	event.delegateTarget.classList.add('rgh-actionable-link'); // Mark this as processed
+	const branch = select('[title^="compare"]')!.textContent!;
+
+	const dropdown = event.delegateTarget.nextElementSibling!;
+	const viewFile = select<HTMLAnchorElement>('[data-ga-click^="View file"]', dropdown)!;
+	viewFile.before(
+		<div className="dropdown-header">Branch: {branch}</div>
```
notlmn, line 29:
```suggestion
		<div className="dropdown-header pl-5 d-block">
			<span class="position-absolute ml-n3"><!-- octicon-branch --></span>
			{branch}
		</div>
```

image

yakov116: :heart:

source/features/enable-file-links-in-compare-view.tsx
```diff
@@ -0,0 +1,84 @@
...
+}
+
+function handleCompareMenuOpening(event: delegate.Event): void {
+	event.delegateTarget.classList.add('rgh-actionable-link'); // Mark this as processed
+	const branch = select('[title^="compare"]')!.textContent!;
+
+	const dropdown = event.delegateTarget.nextElementSibling!;
+	const viewFile = select<HTMLAnchorElement>('[data-ga-click^="View file"]', dropdown)!;
+	viewFile.before(
+		<div className="dropdown-header pl-5 d-block">
```
fregante, line 29: Try matching the existing elements instead. They don't have `pl-5 d-block`. Does `dropdown-header dropdown-item` work?

yakov116: The dropdown-item causes the text to wrap 
image

fregante: True. This should work though
```suggestion
		<div className="dropdown-header pl-5">
```

yakov116:
What do you feel about having the head user? (yakov116:branchname)

yakov116:
Long branch
image

Short Branch
image

source/features/enable-file-links-in-compare-view.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+function handlePRMenuOpening(event: delegate.Event): void {
+	event.delegateTarget.classList.add('rgh-actionable-link'); // Mark this as processed
+
+	// This solution accounts for:
+	// - Branches with slashes in it
+	// - PRs opened from the default branch
+	const dropdown = event.delegateTarget.nextElementSibling!;
+	const viewFile = select<HTMLAnchorElement>('[data-ga-click^="View file"]', dropdown)!;
+	const headBranchUrl = select<HTMLAnchorElement>('.commit-ref.head-ref a')!.pathname;
```
fregante, line 18:
```suggestion
	const headBranchUrl = select<HTMLLinkElement>('[type="application/atom+xml"]')!.href.replace(/\.atom$/, '');
```

Should fix https://github.com/sindresorhus/refined-github/issues/3123

Side note: it looks like `getCurrentBranch` works on PRs as well.

yakov116: I made the same mistake before https://github.com/sindresorhus/refined-github/pull/3111, that gives you the local repo url  `sindresorhus/refined-github/` vs `yakov116/refined-github/`

fregante: Ugh PRs are awkward. GitHub treats your branch as a branch of this repo, which is what threw me off

```
https://github.com/sindresorhus/refined-github/commits/enable-file-links-in-compare-view.atom
```

Therefore, for PRs between default branches, the result of `getCurrentBranch` is completely wrong. Perhaps `getCurrentBranch` _cannot_ be used in PRs.

Example: in https://github.com/yuzu-emu/yuzu/pull/3953/files

```
https://github.com/yuzu-emu/yuzu/commits/master.atom
```
should be

```
https://github.com/GodKratos/yuzu/commits/master.atom
```

yakov116: Which is what let me to [this](https://github.com/sindresorhus/refined-github/pull/3152/commits/650bde6e8356e89d57b6778edb782f3e05ce4272)

>Therefore, for PRs between default branches, the result of getCurrentBranch is completely wrong. Perhaps getCurrentBranch cannot be used in PRs.

The branch is correct but the headrepo is not

source/features/enable-file-links-in-compare-view.tsx
```diff
@@ -0,0 +1,93 @@
...
+	],
+	init
+}, {
+	include: [
+		pageDetect.isCompare
+	],
+	exclude: [
+		// Only enable if you can create a PR or view an existing PR, if you cant you are probably looking at a permalink.
+		() => !select.exists('.existing-pull-button, [data-ga-click*="text:Create pull request"]')
+
```
fregante, line 90:
```suggestion
```

source/features/enable-file-links-in-compare-view.tsx
```diff
@@ -0,0 +1,92 @@
...
+	// You can only get the user and repository, since PR's open from the default branch will not have the branch in the pathname.
+	const [user, repository] = select<HTMLAnchorElement>('.commit-ref.head-ref a')!.pathname.split('/');
+	const url = new GitHubURL(viewFile.href);
+	url.assign({
+		user,
+		repository,
+		branch: getCurrentBranch()
+	});
+
+	viewFile.href = String(url);
```
fregante, line 28: As the code comment mention, this used to work. https://github.com/sindresorhus/refined-github/pull/3111 broke it and the previous version of the code is cleaner than the current one. Does this still work?

```suggestion
	const viewFile = select<HTMLAnchorElement>('[data-ga-click^="View file"]', dropdown)!;
	const pathParts = viewFile.pathname.split('/'); // Example pathname: $owner/$repository/blob/$sha/$path_to_file.tsx
	pathParts[4] = getCurrentBranch();
	viewFile.pathname = pathParts.join('/');
```

yakov116: No since the view file link is the local repo not the head repo. Hence my mistake that was testing on my own repo

The last fully working version is this.
https://github.com/yakov116/refined-github/blob/a24d9460d36e5147f5a9f389923f2eb75f34e34f/source/features/fix-view-file-link-in-pr.tsx#L25-L34

fregante: I think the difference is that the link will still point to the base repo rather than the head repo, but it's not a big deal as long as it workks.

source/features/enable-file-links-in-compare-view.tsx
```diff
@@ -0,0 +1,92 @@
+import React from 'dom-chef';
+import select from 'select-dom';
+import delegate from 'delegate-it';
+import GitBranchIcon from 'octicon/git-branch.svg';
+import * as pageDetect from 'github-url-detection';
+
+import features from '.';
+import GitHubURL from '../github-helpers/github-url';
+import {getCurrentBranch} from '../github-helpers';
+
```
fregante, line 10:
```suggestion

/** Rebuilds the "View file" link because it points to the base repo and to the commit, instead of the head repo and its branch */
```

source/features/enable-file-links-in-compare-view.tsx
fregante, line 28:
```suggestion
	const [user, repository] = select<HTMLAnchorElement>('.commit-ref.head-ref a')!.pathname.split('/', 3);
	const filePath = dropdown.closest('[data-path]')!.getAttribute('data-path')!;

	const viewFile = select<HTMLAnchorElement>('[data-ga-click^="View file"]', dropdown)!;
	viewFile.path = [, user, repository, 'blob', getCurrentBranch(), filePath].join('/'); // Do not replace with `GitHubURL`  #3152 #3111 #2595
```

qaisjp:
Looks great!

PR #3153, opened 2020-05-28
Fix URL parsing in history views

### LINKED ISSUE #3165: `link-to-file-in-file-history` doesn't appear on some URLs
These 2 pages are equivalent: 

- https://github.com/yakov116/refined-github/commits/f23b687b3b89aa95a76193722cdfeff740646670/source/features/release-download-count.tsx?after=f23b687b3b89aa95a76193722cdfeff740646670+34
- https://github.com/yakov116/refined-github/commits/f23b687b3b89aa95a76193722cdfeff740646670?after=f23b687b3b89aa95a76193722cdfeff740646670+34&path%5B%5D=source&path%5B%5D=features&path%5B%5D=release-download-count.tsx

But the latter doesn't show `link-to-file-in-file-history`

yakov116:
Follows all unrelated comments (sorry @fregante)
Fixes #3165 

Test on https://github.com/yakov116/refined-github/commits/f23b687b3b89aa95a76193722cdfeff740646670?after=f23b687b3b89aa95a76193722cdfeff740646670+34&path%5B%5D=source&path%5B%5D=features&path%5B%5D=release-download-count.tsx

`default-branch-button` Should not error
`follow-file-renames` Should work
`link-to-file-in-file-history` Should work


fregante:
I made [the change to use the whole `href`](https://github.com/sindresorhus/refined-github/pull/3138#issuecomment-634646217) instead of just the `pathname` specifically to have `GitHubURL` handle the whole URL internally, so this should be moved inside it.

source/features/default-branch-button.tsx
```diff
@@ -19,6 +19,7 @@ async function init(): Promise<false | void> {
 	}
 
 	const url = new GitHubURL(location.href);
+	url.filePath = url.filePath || filePathFromSearch();
 	// The branch selector will be on `isRepoCommitList()` **unless** you're in a folder/file
 	if (pageDetect.isRepoCommitList() && url.filePath.length > 0) {
 		return false;
```
fregante: I wonder if in this case we should just test the branch selector and drop this condition.

```suggestion
	const branchSelector = await elementReady('#branch-select-menu');
	// The branch selector is missing from History pages of files and folders (it only appears on the root)
	if (!branchSelector) {
		return false;
```

source/github-helpers/github-url.ts
```diff
@@ -1,8 +1,9 @@
 import {getCurrentBranch} from '.';
 
-function disambiguateReference(ambiguousReference: string[]): {branch: string; filePath: string} {
+function disambiguateReference(ambiguousReference: string[], search: string): {branch: string; filePath: string} {
```
fregante, line 3: Let's make `disambiguateReference` a private method of the class so it has access to `this` and don't have to pass `search`

source/github-helpers/github-url.ts
```diff
@@ -1,8 +1,9 @@
...
 	const branch = ambiguousReference[0];
-	const filePath = ambiguousReference.slice(1).join('/');
+	const filePathFromSearch = decodeURI(search).split(/&path\[]=/).splice(1).join('/');
```
fregante, line 5: Use `this.searchParams` to parse the file path safely:
```suggestion
	const filePathFromSearch = this.searchParams.getAll('path[]').join('/');
```
If the above is not falsey, return instantly
```suggestion
	const filePathFromSearch = this.searchParams.getAll('path[]').join('/');
	if (filePathFromSearch) {
		return {branch, filePath: filePathFromSearch};
	}
```

source/github-helpers/github-url.ts
```diff
@@ -44,6 +16,39 @@ export default class GitHubURL extends URL {
...
+		}
+
+		const filePath = ambiguousReference.slice(1).join('/');
+
+		const currentBranch = getCurrentBranch();
+		const currentBranchSections = currentBranch.split('/');
+		if (
+			ambiguousReference.length === 1 || // Ref has no slashes
+		currentBranchSections.length === 1 || // Current branch has no slashes
+		/\^|~|@{/.test(branch) // Ref is an extended revision #3137 https://git-scm.com/docs/git-rev-parse#_specifying_revisions
```
fregante, line 33: Indentation

source/features/follow-file-renames.tsx
```diff
@@ -22,7 +22,8 @@ async function findRename(lastCommitOnPage: string): Promise<File[]> {
 function init(): false | void {
 	const disabledPagination = select.all('.paginate-container [disabled], .paginate-container .disabled');
 	const url = new GitHubURL(location.href);
-
+	// Clear the search from the url, so it does not get passed to the rename link.
+	url.search = '';
```
fregante, line 26: Good point, but then again this should be part of `GitHubURL` and be more specific to `path[]`, probably.

source/features/default-branch-button.tsx
```diff
@@ -32,7 +27,12 @@ async function init(): Promise<false | void> {
 		url.branch = defaultBranch;
 	}
 
-	const branchSelector = (await elementReady('#branch-select-menu'))!;
+	const branchSelector = await elementReady('#branch-select-menu');
+	// The branch selector is missing from History pages of files and folders (it only appears on the root)
+	if (!branchSelector) {
+		return false;
+	}
```
fregante, line 34: Should be moved way higher. If the element doesn't exist, why do anything else?

source/features/follow-file-renames.tsx
fregante: Also make sure to add tests

yakov116: >  and be more specific to path[], probably.
I just noticed this. You mean to say only clear the path ones?

test/github-url.ts
```diff
@@ -66,3 +66,16 @@ test('change filePath', t => {
...
+	const url = new GitHubURL('https://github.com/yakov116/refined-github/commits/f23b687b3b89aa95a76193722cdfeff740646670?after=f23b687b3b89aa95a76193722cdfeff740646670+34&path%5B%5D=source&path%5B%5D=features&path%5B%5D=release-download-count.tsx');
+	t.is(url.user, 'yakov116');
+	t.is(url.repository, 'refined-github');
+	t.is(url.route, 'commits');
+	t.is(url.branch, 'f23b687b3b89aa95a76193722cdfeff740646670');
+	t.is(url.filePath, 'source/features/release-download-count.tsx');
+	t.is(url.pathname, '/yakov116/refined-github/commits/f23b687b3b89aa95a76193722cdfeff740646670/source/features/release-download-count.tsx');
+	t.is(url.href, 'https://github.com/yakov116/refined-github/commits/f23b687b3b89aa95a76193722cdfeff740646670/source/features/release-download-count.tsx');
+	t.is(url.search, '');
+	t.is(String(url), 'https://github.com/yakov116/refined-github/commits/f23b687b3b89aa95a76193722cdfeff740646670/source/features/release-download-count.tsx');
```
fregante, line 80: > > > > > be more specific to `path[]`, probably.

You're changing the URL. They're not equivalent:

```js
// https://github.com/yakov116/refined-github/commits/f23b687b3b89aa95a76193722cdfeff740646670?after=f23b687b3b89aa95a76193722cdfeff740646670+34&path%5B%5D=source&path%5B%5D=features&path%5B%5D=release-download-count.tsx
// https://github.com/yakov116/refined-github/commits/f23b687b3b89aa95a76193722cdfeff740646670/source/features/release-download-count.tsx
```

yakov116: And I should be not? Or the test is wrong?

fregante: I think this is the closest equivalent URL

```suggestion
	t.is(String(url), 'https://github.com/yakov116/refined-github/commits/f23b687b3b89aa95a76193722cdfeff740646670/source/features/release-download-count.tsx?after=f23b687b3b89aa95a76193722cdfeff740646670+34');
```

yakov116: https://github.com/yakov116/refined-github/blob/aefc57224678e5019380cb68d5742895dddfbea7/source/github-helpers/github-url.ts#L23-L25

```
			// Clear the search from the url, so it does not get passed to the new url.
			this.search = '';
			return {branch, filePath: filePathFromSearch};
```
So this is wrong correct?

yakov116: https://github.com/yakov116/refined-github/blob/aefc57224678e5019380cb68d5742895dddfbea7/source/github-helpers/github-url.ts#L23-L25

```
			// Clear the search from the url, so it does not get passed to the new url.
			this.search = '';
			return {branch, filePath: filePathFromSearch};
```
So this is wrong correct?

source/features/follow-file-renames.tsx
fregante: Not sure, but the Older link doesn't work on https://github.com/yakov116/refined-github/commits/f23b687b3b89aa95a76193722cdfeff740646670/source/features/release-download-count.tsx?after=f23b687b3b89aa95a76193722cdfeff740646670+34

source/github-helpers/github-url.ts
```diff
@@ -44,6 +16,40 @@ export default class GitHubURL extends URL {
 		this.pathname = super.pathname;
 	}
 
+	disambiguateReference(ambiguousReference: string[]): {branch: string; filePath: string} {
```
fregante, line 19:
```suggestion
	private disambiguateReference(ambiguousReference: string[]): {branch: string; filePath: string} {
```

test/github-url.ts
fregante: The tests are correct now, so that is wrong since it drops the `after` parameter.

source/features/follow-file-renames.tsx
fregante: Apparently that also happens without this PR. I'll open a new issue.

fregante: Nope. It works correctly. The build wasn't updating 👌

source/github-helpers/github-url.ts
```diff
@@ -49,13 +21,51 @@ export default class GitHubURL extends URL {
...
 
+	toString() {
+		return this.href;
+	}
+
+	private disambiguateReference(ambiguousReference: string[]): {branch: string; filePath: string} {
+		const branch = ambiguousReference[0];
+		const filePathFromSearch = this.searchParams.getAll('path[]').join('/');
+		if (filePathFromSearch) {
+			// Clear the search from the url, so it does not get passed to the new url.
```
yakov116, line 32:
```suggestion
			// Clear the search from the `path[]` param, so it does not get passed to the new url.
```

I dont think that is even needed any more WFYT?

source/github-helpers/github-url.ts
```diff
@@ -16,7 +16,16 @@ export default class GitHubURL extends URL {
 		this.pathname = super.pathname;
 	}
 
-	disambiguateReference(ambiguousReference: string[]): {branch: string; filePath: string} {
+	assign(replacements: Partial<GitHubURL>): this {
+		Object.assign(this, replacements);
+		return this;
+	}
```
fregante: It looks like this works too, it lets us assign multiple times if necessary, e.g. url.assign(opts, {route: 'blob'})
```suggestion
	assign = Object.assign.bind(this);
```

But it needs to be moved above the constructor.

source/github-helpers/github-url.ts
fregante: Yeah I think it can be dropped, it's self-explanatory: _we used it, so we're dropping it._

source/github-helpers/github-url.ts
yakov116: it will affect the way assign works.
I will commit so you can see

fregante:
So long and thanks for all the fixes!

source/github-helpers/github-url.ts
```diff
@@ -39,14 +11,49 @@ export default class GitHubURL extends URL {
...
-	assign(replacements: Partial<GitHubURL>): this {
-		Object.assign(this, replacements);
-		return this;
+	toString() {
+		return this.href;
+	}
+
+	private disambiguateReference(ambiguousReference: string[]): {branch: string; filePath: string} {
+		const branch = ambiguousReference[0];
+		const filePathFromSearch = this.searchParams.getAll('path[]').join('/');
```
fregante, line 28: Needs a comment explaining when this happens. e.g. `// History pages might use search parameters`

yakov116: Will add

PR #3154, opened 2020-05-28
Deemphasize `update-pr-from-base-branch` button

### LINKED ISSUE #2723: Make `update-pr-from-base-branch`'s button smaller
Currently the "Update branch" is rather large and it looks like an action that should be taken. It's not really something the user should do that often, so we should probably de-emphasize it.

<table><td>

</table>

This is an alternative, it just replaces `.btn` with `.btn-link`:

<table><td>

</table>

Also:

- [ ] it should probably use a `window.prompt()` window to ask for confirmation
- [ ] it could have an "info" icon with a link to https://github.com/sindresorhus/refined-github/issues/1236 or any other post with a clearer explanation of when this action may be useful

### LINKED ISSUE #3017: `Update branch` button is misaligned on a draft PR
Noticed a misaligned `Update branch` button on my draft PR. The button is not there at all with the extension disabled. I am about to update the PR, hence no link, but here's a screenshot:

image

FloEdelmann:
Closes #2723, closes #3017.

I've taken a little different approach than outlined in #2723: Always add the button as a `.btn-link` to the `.status-meta`, and also show `.status-meta` for success status messages if necessary. That way, we don't have to deal with different locations for the same button.

I hope I have tested all cases, please everyone try this in other PRs with different status messages!

Screenshot_2020-05-28_20-40-27
Screenshot_2020-05-29_17-26-54
Screenshot_2020-05-29_17-27-21
Screenshot_2020-05-29_17-30-12
Screenshot_2020-05-28_20-42-02
Screenshot_2020-05-28_20-44-03


fregante:
Excellent! However I think this undoes `clean-mergeability-box` at least partially. Can you just add an extra line and only unhide _that_ (with `:not()`) rather than the whole line that we normally hide?

source/features/update-pr-from-base-branch.tsx
```diff
@@ -88,8 +94,9 @@ async function addButton(): Promise<void> {
 		return;
 	}
 
-	for (const heading of select.all('.mergeability-details > :not(.js-details-container) .status-heading')) {
-		heading.append(createButton(base, head));
+	for (const meta of select.all('.mergeability-details > :not(.js-details-container) .status-meta')) {
+		meta.classList.add('rgh-has-update-pr-from-master-button');
+		meta.append(createButton(base, head));
```
fregante, line 99: Basically something like this (but optimized to avoid repetition and unnecessary wrappers)
```suggestion
	for (const meta of select.all('.mergeability-details > :not(.js-details-container) .status-meta')) {
		meta.after(
			<span class="status-meta rgh-has-update-pr-from-master-button">
				{createButton(base, head))}
			</span>
		)
        
```

FloEdelmann:
I've put it inside a separate `.status-meta`, and it even stays on the same line because both are `display: inline;` :blush:

fregante:
I see it on this PR 



but not on #3153:

draft


Is this due to #3155?

FloEdelmann:
@fregante Probably, since I didn't change the logic *when* to show the button in this PR, just *where* and *how*.

Could you check if the button is maybe added in a hidden section?

In a similar case (https://github.com/OpenLightingProject/open-fixture-library/pull/359: I have write access to the repo, someone else opened a draft PR that is now out-of-date), I see it's added for me (both in the current version and the updated look).

fregante:
It doesn't seem related to this PR. The button doesn't appear on `master` either.

I'll test this further and merge it

source/features/update-pr-from-base-branch.tsx
```diff
@@ -28,38 +28,43 @@ async function mergeBranches(): Promise<AnyObject> {
...
 		throw new api.RefinedGitHubAPIError('update-pr-from-base-branch: ' + JSON.stringify(response));
 	}
 }
 
 function createButton(base: string, head: string): HTMLElement {
-	return (
-		<button type="button" className="btn float-right rgh-update-pr-from-master tooltipped tooltipped-n" aria-label={`Merge the ${base} branch into ${head}`}>
-			Update branch
+	const button = (
+		<button type="button" className="btn-link tooltipped tooltipped-n" aria-label={`Merge the ${base} branch into ${head}`}>
```
fregante, line 58: Since we ask for confirmation now, this seems duplicate/unnecessary
```suggestion
		<button type="button" className="btn-link">
```

FloEdelmann: Done.

fregante:
🙌

PR #3157, opened 2020-05-30
Meta: config and build cleanup

fregante:
- XO’s configuration grew a lot, so I extracted it from package.json
- Instead of using a webpack plugin for asynchronous type checking, I just run `tsc` in parallel and get native error messages
- Webpack’s config file could be improved slightly
- I had also tried avoiding duplicated code with [SplitChunks](https://webpack.js.org/plugins/split-chunks-plugin/) but for some reason the content scripts weren't executing (maybe due to https://github.com/webpack/webpack/issues/7230). Also Chrome doesn't like `~` in filenames, but that was an [easy fix](https://webpack.js.org/plugins/split-chunks-plugin/#splitchunksautomaticnamedelimiter)


.xo-config.js
```diff
@@ -0,0 +1,60 @@
...
+				svg: 'always'
+			}
+		],
+		'unicorn/no-fn-reference-in-iterator': 'off'
+	},
+	settings: {
+		react: {
+			version: '16.3'
+		}
+	}
```
sindresorhus, line 59: https://github.com/xojs/eslint-config-xo-react/blob/ab2d7bb616c59b48d8b2f9aa1a89723deceaccaf/index.js#L12-L16

.xo-config.js
```diff
@@ -0,0 +1,60 @@
...
+		'import/no-anonymous-default-export': 'error',
+		'import/no-commonjs': 'error',
+		'import/no-named-default': 'error',
+		'import/no-unassigned-import': 'off',
+		'import/prefer-default-export': 'error',
+		'import/extensions': [
+			'error',
+			'never',
+			{
+				svg: 'always'
```
sindresorhus, line 50: Happy to have this one added to XO.

.xo-config.js
```diff
@@ -0,0 +1,60 @@
...
+		'@typescript-eslint/prefer-readonly-parameter-types': 'off',
+		'no-alert': 'off',
+		'no-void': 'off',
+		'react/function-component-definition': [
+			'error',
+			{
+				namedComponents: 'function-declaration'
+			}
+		],
+		'react/jsx-first-prop-new-line': 'error',
```
sindresorhus, line 25: https://github.com/xojs/eslint-config-xo-react/blob/ab2d7bb616c59b48d8b2f9aa1a89723deceaccaf/index.js#L101

.xo-config.js
fregante: We don't have react so the detection fails

Issue #2852, opened 2020-03-03
`show-whitespace` doesn’t work on newly sent suggestions


szmarczak:
![Peek 2020-03-03 16-01](https://user-images.githubusercontent.com/36894700/75788387-74749200-5d68-11ea-946d-807455b8c309.gif)

szmarczak:
No, I checked that for PRs.

yakov116:
WHat?

yakov116:
Ok the issue is (From what I see) OnNewComments Does not run on the PR File tab. So any new comments on that tab will not work.

The only issue I have is that it was my own PR so I was only able to comment (not apply a real review)

If someone wants they can open a PR on https://github.com/yakov116/TestR/ so I could test

szmarczak:
Checked again, the issue still persists.

yakov116:
@szmarczak can you open a pr there and turn off subscribe so I could test a real review?

yakov116:
@szmarczak I am only seeing it on the PR Files Tab.

szmarczak:
> I am only seeing it on the PR Files Tab.

That's what I'm talking about... `Files changed`... I guess I could've said that at the beginning :P

yakov116:
>Ok the issue is (From what I see) OnNewComments Does not run on the PR File tab. So any new comments on that tab will not work.

I tried to figure it out... No luck.

fregante:
While looking for this I figured out an easy way to see GitHub's event names:

```js
d = Node.prototype.dispatchEvent;
Node.prototype.dispatchEvent = function (...a) {
	console.log(...a);
	// debugger; // Uncomment when necessary
	d.apply(this, a);
}
```

fregante:
Anyway, no special events seem to be available. The only solution I can think of:

```js
o = new MutationObserver(runUpdates);

for commentHolders of select.all('.diff-table tbody, .js-comments-holder')
	o.observe(commentHolders)
```

This should probably be part of `on-new-comments` somehow. After #2987

fregante:
I really think that https://github.com/sindresorhus/refined-github/issues/3084 should fix all of our problems, it should be revisited: https://github.com/sindresorhus/refined-github/pull/3092

That PR probably just needs an efficient mechanism to avoid applying changes to the same element.

notlmn:
@fregante any reason we can't listen on `socket:message` events and handle everything from there?

fregante:
If it works, sure.

PR #3158, opened 2020-05-30
Fix handling lazy-loaded comments

notlmn:
This is similar to how lazy handlers are added in `on-pr-file-load`, and as `include-fragment` is not a direct descendant of `details` element.

Related to #2852, but doesn't exactly fix that. Stilling figuring out on how to handle that.


FloEdelmann:
For reference, this is similar to https://github.com/sindresorhus/refined-github/pull/2948.

PR #3159, opened 2020-05-31
Be consistent with the rest of the notif buttons

qaisjp:
Instead of a custom tooltip, the other GitHub buttons on this page use `title`.

**GitHub button**

image

**Existing open unread button**

image

**New open unread button**

image

### DIFF
--- a/source/features/open-all-notifications.tsx
+++ b/source/features/open-all-notifications.tsx
@@ -47,7 +47,7 @@ function addOpenReposButton(): void {
 		}
 
 		select('.js-grouped-notifications-mark-all-read-button', repository)!.before(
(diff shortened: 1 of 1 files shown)


notlmn:
Title attributes almost always goes against accessibility

- https://mrwweb.com/the-title-attribute-and-why-its-almost-useless
- https://silktide.com/blog/2013/i-thought-title-text-improved-accessibility-i-was-wrong

qaisjp:
Hmm, well TIL.

fregante:
TIL too :o

I guess nowadays it's more about SEO than anything.

Also `.tooltipped` looks nicer and works more consistently on GitHub

PR #3160, opened 2020-06-01
Lint and add missing screenshots

yakov116:
Fixes one error and lint's a few more

- Wait for milestone and project dropdowns to be ready
	Looks like Github does not **always** load them right away. ( I dont get the error every time)

```javascript
refined-github.js:12163 Uncaught (in promise) TypeError: Cannot read property 'parentElement' of null
    at hideMilestones (refined-github.js:12163)
    at clean_issue_filters_init (refined-github.js:12189)
    at async runFeature (refined-github.js:5637)
    at async setupPageLoad (refined-github.js:5651)
hideMilestones @ refined-github.js:12163
clean_issue_filters_init @ refined-github.js:12189
async function (async)
clean_issue_filters_init @ refined-github.js:12185
runFeature @ refined-github.js:5637
setupPageLoad @ refined-github.js:5651
add @ refined-github.js:5708
async function (async)
add @ refined-github.js:5682
./source/refined-github.ts @ refined-github.js:12192
__webpack_require__ @ refined-github.js:20
(anonymous) @ refined-github.js:84
(anonymous) @ refined-github.js:87
refined-github.js:12181 Uncaught (in promise) TypeError: Cannot read property 'parentElement' of null
    at hideProjects (refined-github.js:12181)
```

- Simple linting https://github.com/sindresorhus/refined-github/commit/6ce246a52fede2e51fe2b2292975addf8e321bbb
- Add screenshot links to features that had them in the readme but not in the feature itself https://github.com/sindresorhus/refined-github/commit/2c4ea713787b50fe27cb7915c422418ccb82e503

-  Switch shortcut map to Object.entries (TODO comment) https://github.com/sindresorhus/refined-github/commit/dfc42b9ed21d4794b3b1fc22878e8c23de05be5c

PR #3163, opened 2020-06-02
Meta: Change format of shortcuts map

yakov116:
Follows
>I think this is the format I was referring to
>
>```suggestion
>		shortcutMap.set(hotkey, description);
>```
>
>_Originally posted by @fregante in https://github.com/sindresorhus/refined-github/pull/3160_

_Was almost related to this PR_ but https://github.com/fregante/webext-additional-permissions/blob/master/index.ts#L15 uses flatMap and that is only supported from [firefox 62](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap#Browser_compatibility) currently were set to 61

### DIFF
--- a/source/features/improve-shortcut-help.tsx
+++ b/source/features/improve-shortcut-help.tsx
@@ -16,7 +16,7 @@ function improveShortcutHelp(dialog: Element): void {
 			</div>
 
 			<ul>
-				{features.getShortcuts().map(({hotkey, description}) => (
+				{[...features.shortcutMap].map(([hotkey, description]) => (
 					<li className="Box-row d-flex flex-row">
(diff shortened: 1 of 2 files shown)


yakov116:
I think you called it shortcutMap because it was clashing with the types

fregante:
The types are right, it's just that there's already a constant called `shortcuts` in one of the functions:

https://github.com/sindresorhus/refined-github/blob/347c7d6df3ef5313c339514180ef719daba16f20/source/features/index.tsx#L187-L199

Alright it can be changed back

fregante:
👯

PR #3166, opened 2020-06-02
Restore private repo in some features

yakov116:
Follows https://github.com/sindresorhus/refined-github/commit/9ddcd95d766d26af8fe6362c65f32370464c5bd2#r39608686

### DIFF
--- a/source/github-helpers/index.ts
+++ b/source/github-helpers/index.ts
@@ -17,7 +17,7 @@ export const getDiscussionNumber = (): string | undefined => {
 };
 
 /**
-Tested on isRepoTree, isBlame, isSingleFile, isEditFile, isCommit, isCommitList, isCompare, isPR
+Tested on isRepoTree, isBlame, isSingleFile, isEditFile, isSingleCommit, isCommitList, isCompare. Subtly incompatible with isPR
 Example tag content on public repositories: https://github.com/sindresorhus/refined-github/commits/branch-or-commit/even/with/slashes.atom
 Example tag content on private repositories https://github.com/private/private/commits/master.atom?token=AEAXKWNRHXA2XJ2ZWCMGUUN44LM62
 */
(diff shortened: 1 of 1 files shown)


fregante:
Can you add a comment with the two examples? Otherwise this is bound to get “fixed” again

fregante:
Better yet, make the regex more specific since it currently breaks if a branch or repo name contains `.atom`

yakov116:
@fregante done.
I force-pushed since I reopened a very old branch (from my last pr)

fregante:
> Better yet, make the regex more specific since it currently breaks if a branch or repo name contains `.atom`

I prefer this

source/github-helpers/index.ts
```diff
@@ -18,15 +18,16 @@ export const getDiscussionNumber = (): string | undefined => {
...
+Example tag content on private repositories https://github.com/private/private/commits/master.atom?token=AEAXKWNRHXA2XJ2ZWCMGUUN44LM62
 */
 export const getCurrentBranch = (): string => {
 	return select<HTMLLinkElement>('[type="application/atom+xml"]')!
 		.href
 		.split('/')
 		.slice(6) // Drops the initial https://host/user/repo/route/ part
 		.join('/')
-		.replace(/\.atom$/, '');
+		.replace(/\.atom[^.]*$/, '');
```
fregante, line 30: I meant something like this
```suggestion
		.replace(/\.atom(\?token=.+)?$/, '');
```

sindresorhus:
Feels like it would be more future-proof to use `URL` to clear the query string:

```js
const x = new URL('https://github.com/private/private/commits/master.atom?token=AEAXKWNRHX');
x.search = '';
console.log(x.pathname);
```

yakov116:
Or even
```
const x = new URL('https://github.com/private/private/commits/master.atom?token=AEAXKWNRHX').pathname.split('/')....;
```

Then there is no need to even clear the search

PR #3167, opened 2020-06-03
Meta: update code comment

FloEdelmann:
Follows #3152.


fregante:
This is why I prefer an estimated date there and not `???` so we don't forget to update it.

PR #3168, opened 2020-06-03
Upgrade dependencies

sindresorhus:


.xo-config.js
```diff
@@ -50,7 +43,14 @@ module.exports = {
 				svg: 'always'
 			}
 		],
-		'unicorn/no-fn-reference-in-iterator': 'off'
+
+		// Temporarily disabled
+		'@typescript-eslint/no-invalid-void-type': 'off',
+		'@typescript-eslint/no-extra-non-null-assertion': 'off',
```
sindresorhus, line 49: This one wanted to fix something that TypeScript didn't agree with. I suggest keeping it disabled for a while and hope it will be improved.

.xo-config.js
```diff
@@ -50,7 +43,14 @@ module.exports = {
...
+		'@typescript-eslint/no-unnecessary-condition': 'off',
```
sindresorhus, line 50: This has a lot of false-positives. Mostly because array access doesn't result in `string | undefined`, but rather `string`. I guess I'll have to disable it in XO again...

.xo-config.js
```diff
@@ -50,7 +43,14 @@ module.exports = {
 				svg: 'always'
 			}
 		],
-		'unicorn/no-fn-reference-in-iterator': 'off'
+
+		// Temporarily disabled
+		'@typescript-eslint/no-invalid-void-type': 'off',
```
sindresorhus, line 48: The rule recommends not using `void` in a union type, but `undefined | false` didn't seem to work, so not entirely sure what the correct solution is.

.xo-config.js
sindresorhus: https://github.com/typescript-eslint/typescript-eslint/issues/1798

fregante: Yeah that rule is problematic especially for libraries since it's true that TS users can't pass the wrong type, but regular JS users can. So this might still sense:

```js
function foo(bar: string) {
	if (typeof bar !== 'string') { // no-unnecessary-condition!! … or is it?
		throw new TypeError('Not a duck')
	}
}
```

.xo-config.js
fregante: > but `undefined | false` didn't seem to work

That might be a bug then, because their documentation specifically talks about returning undefined. 🤔

fregante: Either way I don't particularly want to clutter each `init` function with `return undefined`

.xo-config.js
```diff
@@ -22,7 +18,6 @@ module.exports = {
 				namedComponents: 'function-declaration'
 			}
 		],
-		'react/jsx-first-prop-new-line': 'error',
```
fregante, line 25: Did you disable this because you enabled it in xo?

.xo-config.js
```diff
@@ -10,10 +10,6 @@ module.exports = {
 		'__filebasename'
 	],
 	rules: {
-		'@typescript-eslint/no-empty-function': 'off',
-		'@typescript-eslint/no-floating-promises': 'off',
-		'@typescript-eslint/no-non-null-asserted-optional-chain': 'off',
-		'@typescript-eslint/prefer-readonly-parameter-types': 'off',
```
fregante, line 16: # 🎉

.xo-config.js
```diff
@@ -50,7 +43,14 @@ module.exports = {
 				svg: 'always'
 			}
 		],
-		'unicorn/no-fn-reference-in-iterator': 'off'
+
+		// Temporarily disabled
+		'@typescript-eslint/no-invalid-void-type': 'off',
+		'@typescript-eslint/no-extra-non-null-assertion': 'off',
+		'@typescript-eslint/no-unnecessary-condition': 'off',
+		'@typescript-eslint/no-floating-promises': 'off',
```
fregante, line 51: Will you allow `void` on this instead?

.xo-config.js
sindresorhus: Yup: https://github.com/xojs/eslint-config-xo-react/blob/93f78865088aa5deab246e1cc4f40959fc5ff5be/index.js#L101

package.json
```diff
@@ -59,11 +57,11 @@
 		"flat-zip": "^1.0.1",
 		"github-reserved-names": "^1.1.8",
 		"github-url-detection": "^1.1.5",
-		"image-promise": "^7.0.0",
+		"image-promise": "^7.0.1",
 		"indent-textarea": "^2.0.1",
 		"linkify-issues": "2.0.0-nolookbehind",
 		"linkify-urls": "3.1.0-nolookbehind",
-		"mem": "^6.0.1",
+		"mem": "^6.1.0",
```
fregante, line 64: I actually had left this on 6.0.1 because it's technically a better version since we don't use `mem.clear`

.xo-config.js
sindresorhus: The rule wants us to convert:

```js
if (checksCounter?.textContent!.trim() === '0') {
	checksCounter?.parentElement!.remove();
}
```

to

```js
if (checksCounter!.textContent!.trim() === '0') {
	checksCounter!.parentElement!.remove();
}
```

Are we ok with that?

source/features/center-reactions-popup.css
```diff
@@ -53,6 +53,6 @@ td.js-line-comments:last-child .timeline-comment-actions .add-reaction-popover::
 
 /* Keep colored button when reaction popup is open */
 .dropdown-details[open] .timeline-comment-action {
-	opacity: 1;
+	opacity: 100%;
```
fregante, line 56: I've never seen this used anywhere before. Seems unnecessar(il)y (longer)

source/features/first-published-tag-for-merged-pr.tsx
```diff
@@ -15,7 +15,7 @@ const getFirstTag = cache.function(async (commit: string): Promise<string | unde
 		'ul.branches-tag-list li:last-child a'
 	);
 
-	return firstTag?.textContent!;
+	return firstTag?.textContent ?? undefined;
```
fregante, line 18: ?

This already meant:

```js
firstTag === null || firstTag === void 0 ? void 0 : firstTag.textContent;
```

`textContent` is never null in practice.

That lengthens it to this, without change in behavior

```js
var _a;
(_a = firstTag === null || firstTag === void 0 ? void 0 : firstTag.textContent) !== null && _a !== void 0 ? _a : undefined;
```

source/features/parse-backticks.css
```diff
@@ -1,6 +1,6 @@
 /* Style copied from GitHub's <code> */
 .rgh-parse-backticks {
-	background-color: rgba(27, 31, 35, 0.05);
+	background-color: rgb(27 31 35 / 5%);
```
fregante, line 3: If we do this we have to raise the minimum compatible version in manifest.json

source/github-helpers/github-url.ts
```diff
@@ -29,15 +29,15 @@ function disambiguateReference(ambiguousReference: string[]): {branch: string; f
...
+	// @ts-expect-error
 	repository: string;
-	// @ts-ignore
+	// @ts-expect-error
 	route: string;
-	// @ts-ignore
+	// @ts-expect-error
 	branch: string;
-	// @ts-ignore
+	// @ts-expect-error
```
fregante, line 40: # 🙌

.xo-config.js
fregante: Not ok, `checksCounter` might actually be missing, so this would cause an error.

The condition should read as "if it exists and it's equal to 0, remove it"

fregante: The only unnecessary part in that is the second `checksCounter?` since if the condition is true, `checksCounter` is definitely not `undefined`. But TypeScript doesn't recognize that yet.

source/features/center-reactions-popup.css
sindresorhus: It's the modern syntax and more readable as it makes it clear what the range is.

https://twitter.com/mathias/status/1253242715304857601

source/features/parse-backticks.css
sindresorhus: https://github.com/sindresorhus/refined-github/pull/3168/commits/854e51b524e9d4607d8325b51e7270e263123a79

source/features/center-reactions-popup.css
fregante: hqdefault

package.json
sindresorhus: https://github.com/sindresorhus/refined-github/pull/3168/commits/3b998202f2cb574e196d275839062aa342b381a5

source/features/first-published-tag-for-merged-pr.tsx
sindresorhus: It was reported as an error:

> source/features/first-published-tag-for-merged-pr.tsx:18:9
  ✖   18:9  Optional chain expressions can return undefined by design - using a non-null assertion is unsafe and wrong.  @typescript-eslint/no-non-null-asserted-optional-chain

.xo-config.js
sindresorhus: Do you think it makes sense to explicitly annotate promise-returning functions with `void` if they don't need to be await'ed?

.xo-config.js
sindresorhus: I agree. We can leave it off for now, and we can think about whether it should be disabled in XO or just here. I haven't made up my mind yet.

.xo-config.js
fregante: It'd force this everywhere 😬

```diff
- features.add({
+ void features.add({
```

But I think it's worth it everywhere else. It's good that the error message also suggests that you can just prepend `void` to fix the issue.

source/features/first-published-tag-for-merged-pr.tsx
fregante: Siiiiigh so `@typescript-eslint/no-non-null-asserted-optional-chain` is still broken.

sindresorhus:
`"target": "es2018",` => `"target": "es2020",`?

fregante:
I'm not entirely sure though, it's worth trying out the extension in both browsers. I think it's as easy as loading it: if it works without errors (content, options, background), then the browser supports the syntax

.xo-config.js
```diff
@@ -50,7 +43,14 @@ module.exports = {
...
 		],
-		'unicorn/no-fn-reference-in-iterator': 'off'
+
+		// Temporarily disabled
+		'@typescript-eslint/no-invalid-void-type': 'off', // Reason: https://github.com/sindresorhus/refined-github/pull/3168#discussion_r434479434
+		'@typescript-eslint/no-extra-non-null-assertion': 'off', // Reason: https://github.com/sindresorhus/refined-github/pull/3168#discussion_r434478709
+		'@typescript-eslint/no-unnecessary-condition': 'off', // Reason: https://github.com/sindresorhus/refined-github/pull/3168#discussion_r434479136
+		'@typescript-eslint/no-floating-promises': 'off',
+		'unicorn/no-fn-reference-in-iterator': 'off',
+		'unicorn/no-reduce': 'off' // TODO: Remove `reduce` usage.
```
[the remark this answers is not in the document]
sindresorhus: Can be done separately.

source/options.tsx
```diff
@@ -89,7 +89,7 @@ async function init(): Promise<void> {
 		}
 	}
 
-	browser.storage.local.set({featuresAlreadySeen});
+	await browser.storage.local.set({featuresAlreadySeen});
```
sindresorhus, line 92: @fregante I assume this is correct?

.xo-config.js
sindresorhus: Yeah, I'm disabling this rule for now.

.xo-config.js
sindresorhus: Done

sindresorhus:
Webpack errors a lot when I change the output target:

```
ERROR in ./source/features/linkify-symbolic-links.tsx 7:29
Module parse failed: Unexpected token (7:29)
File was processed with these loaders:
 * ./node_modules/ts-loader/index.js
You may need an additional loader to handle the result of these loaders.
| import features from '.';
| function init() {
>     if (select('.file-mode')?.textContent === 'symbolic link') {
|         const line = select('.js-file-line');
|         wrap(line.firstChild, React.createElement("a", { href: line.textContent }));
```

I think that can be done later.

source/features/index.tsx
```diff
@@ -222,15 +222,18 @@ const add = async (meta?: FeatureMeta, ...loaders: FeatureLoader[]): Promise<voi
 
 		const details = {include, exclude, init, deinit, additionalListeners, onlyAdditionalListeners};
 		if (waitForDomReady) {
-			domLoaded.then(async () => setupPageLoad(id, details));
+			(async () => {
+				await domLoaded;
+				await setupPageLoad(id, details);
+			})();
```
fregante, line 228: sneaky sindre

source/features/first-published-tag-for-merged-pr.tsx
sindresorhus: I'v disabled that rule for now and reverted this change.

source/options.tsx
fregante: No since this delays unrelated features.
```suggestion
	void browser.storage.local.set({featuresAlreadySeen});
```

I actually split this `init` function into multiple parts before, but then I undid it because it was unrelated to the PR. I'll open a new PR for that later

fregante:
> Webpack errors a lot when I change the output target:

Apparently due to https://github.com/acornjs/acorn/pull/891

yakov116:
> > Webpack errors a lot when I change the output target:
> 
> Apparently due to [acornjs/acorn#891](https://github.com/acornjs/acorn/pull/891)



https://github.com/webpack/webpack/releases/tag/v5.0.0-beta.18 includes the acorn update

PR #3172, opened 2020-06-03
Split `options.tsx`’ `init` function into multiple parts

fregante:
As mentioned in #3168 

The `init` function is [73-lines long](https://github.com/sindresorhus/refined-github/blob/8b86ff97aa652475792ee9c2e1c86f92cfe35120/source/options.tsx#L54-L127). This PR brings it down to a more reasonable 33-lines.

Suggestions welcome, but know that almost everything depends on these 2 lines

```js
// Generate list
select('.js-features')!.append(...__featuresMeta__.map(buildFeatureCheckbox));

// Update list from saved options
await perDomainOptions.syncForm('form');
```

### DIFF
--- a/source/options.tsx
+++ b/source/options.tsx
@@ -99,7 +99,7 @@ async function highlightNewFeatures(): Promise<void> {
 	void browser.storage.local.set({featuresAlreadySeen});
 }
 
-async function init(): Promise<void> {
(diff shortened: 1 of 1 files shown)


fregante:
Boom, 2 lines!

PR #3174, opened 2020-06-04
Enable `open-issue-to-latest-comment` on the dashboard

yakov116:
Small enhancement. Need some more code since the icon is not a link at all.

I am not sure if this should be here or part of `linkify-labels-on-dashboard`

and have `open-issue-to-latest-comment` run after to add the '#partial-timeline'

Test on https://github.com/


source/features/open-issue-to-latest-comment.tsx
```diff
@@ -9,6 +11,14 @@ function init(): void {
 	}
 }
 
+function initDashboard(): void {
+	for (const icon of select.all('.js-recent-activity-container :not(a) > .octicon-comment')) {
+		const url = new URL(select<HTMLAnchorElement>('a', icon.closest('li')!)!.href);
```
fregante, line 16: Don't use `select` and `closest` in the same line, it's unreadable. Also tag-only selectors automatically set the type

```suggestion
		const url = new URL(icon.closest('li')!.querySelector('a')!.href);
```

yakov116: > Also tag-only selectors automatically set the type

Good to know thanks

source/features/open-issue-to-latest-comment.tsx
```diff
@@ -9,6 +11,14 @@ function init(): void {
 	}
 }
 
+function initDashboard(): void {
+	for (const icon of select.all('.js-recent-activity-container :not(a) > .octicon-comment')) {
+		const url = new URL(icon.closest('li')!.querySelector('a')!.href);
+		url.hash = '#partial-timeline';
+		wrap(icon, <a className="muted-link" href={String(url)}/>);
```
fregante, line 18:
```suggestion
		const url = icon.closest('li')!.querySelector('a')!.pathname + '#partial-timeline';
		wrap(icon, <a className="muted-link" href={url}/>);
```

source/features/open-issue-to-latest-comment.tsx
```diff
@@ -9,6 +11,13 @@ function init(): void {
 	}
 }
 
+function initDashboard(): void {
+	for (const icon of select.all('.js-recent-activity-container :not(a) > .octicon-comment')) {
```
fregante, line 15: Not just the icon. The count needs to be clickable too

yakov116: done

source/features/open-issue-to-latest-comment.tsx
```diff
@@ -12,9 +12,9 @@ function init(): void {
 }
 
 function initDashboard(): void {
-	for (const icon of select.all('.js-recent-activity-container :not(a) > .octicon-comment')) {
+	for (const icon of select.all('.js-recent-activity-container :not(.rgh-latest-comment) .octicon-comment')) {
```
fregante, line 15: This is a common confusion about how `:not()` works

This selector does **not** mean
> select `.octicon-comment` as long as it's not inside `.rgh-latest-comment`.
It means: 
> select `.octicon-comment` as long as one of its ancestors doesn't have that class.

There's a number of elements between `.js-recent-activity-container` and `.octicon-comment` so this will always match it. 

`>` limited the search just to `.octicon-comment` so it worked before this last commit.

source/features/open-issue-to-latest-comment.tsx
```diff
@@ -9,6 +11,13 @@ function init(): void {
 	}
 }
 
+function initDashboard(): void {
+	for (const icon of select.all('.js-recent-activity-container :not(a) > div > .octicon-comment')) {
+		const url = icon.closest('li')!.querySelector('a')!.pathname + '#partial-timeline';
+		wrap(icon.parentElement!, <a className="muted-link rgh-latest-comment" href={url}/>);
```
fregante, line 17: Does this work?

```suggestion
		wrapAll(icon.parentElement!.childNodes, <a className="muted-link rgh-latest-comment" href={url}/>);
```

The parentElement has a `text-gray` class that overrides our `muted-link`. It's best to have the `a` right around the icon and text node only.

fregante: If that doesn't work, something like:
```suggestion
		const container = icon.parentElement!;
		const link = <a className="muted-link rgh-latest-comment" href={url}/>;
		icon.parentElement!.append(link);
		link.append(icon, icon.nextSibling);
```

yakov116: I didnt wrap (pun intended) exactly what you said. I will test it now

yakov116:
Thanks
V2 worked.
While I was at it I fixed githubs bug

image

source/features/open-issue-to-latest-comment.tsx
```diff
@@ -9,6 +10,16 @@ function init(): void {
...
+		const link = <a className="muted-link" href={url}/>;
+		icon.parentElement!.classList.remove('col-1'); // Also fix extra space added by GitHub #3174
+		icon.parentElement!.append(link);
+		link.append(icon, icon.nextSibling!);
+	}
+}
+
 void features.add({
 	id: __filebasename,
 	description: 'Links the comments icon to the latest comment.',
```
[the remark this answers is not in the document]
fregante:
```suggestion
	description: 'Makes the "comment" icon in issue lists link to the latest comment of the issue.',
```

Issue #2927, opened 2020-03-25
`latest-tag-button` sometimes fails getting the "ahead-by count"


fregante:
https://github.com/sindresorhus/refined-github/blob/c84162f1dc7ad898c7a3e303da285004c8c996da/source/features/latest-tag-button.tsx#L73-L74

The expected text is missing from that page: https://github.com/fregante/webext-storage-cache/releases/tag/v2.3.1)

Sadly the count was quite difficult to get. Here was the previous attempt via API: https://github.com/sindresorhus/refined-github/pull/2771#discussion_r378109419

## Example

https://github.com/fregante/webext-storage-cache

<img width="240" alt="example" src="https://user-images.githubusercontent.com/1402241/77490088-3421a480-6e3a-11ea-8289-7e313acdeb15.png">

yakov116:
Instead of using the api lets use fetchDom

```ts
const getAheadByCount = cache.function(async (latestTag: string, defaultBranch: string): Promise<string> => {
	const tagPage = fetchDom(`/${getRepoURL()}/releases/tag/${latestTag}`, '.release-header relative-time + a[href*="/compare/"]');
	const commitsPage = fetchDom(`/${getRepoURL()}/compare/${latestTag}...${defaultBranch}`, '#js-repo-pjax-container div.overall-summary > ul span');
	const aheadCount = await Promise.race([tagPage, commitsPage]);
	// This text is "4 commits to master since this tag"
	return aheadCount!.textContent!.replace(/\D/g, '');
}, {
	maxAge: 1 / 24, // One hour
	staleWhileRevalidate: 2,
	cacheKey: ([latestTag]) => `tag-ahead-by:${getRepoURL()}/${latestTag}`
});
```
From my testing `commitsPage` is almost always wins on speed and always has the info.

yakov116:
What about something like this?
```ts
const getAheadByCount = cache.function(async (latestTag: string, defaultBranch: string): Promise<number> => {
	let aheadCount = await fetchDom(`/${getRepoURL()}/releases/tag/${latestTag}`, '.release-header relative-time + a[href*="/compare/"]');
	if (!aheadCount) {
		aheadCount = await fetchDom(`/${getRepoURL()}/compare/${latestTag}...${defaultBranch}`, '#js-repo-pjax-container div.overall-summary > ul span');
	}

	// This text is "4 commits to master since this tag or 4 Commits"
	return looseParseInt(aheadCount!.textContent!);
}, {
[...]

PR #3180, opened 2020-06-05
Avoid errors in `latest-tag-button` if count can’t be determined

yakov116:
Follows https://github.com/sindresorhus/refined-github/issues/2927#issuecomment-638420454

### DIFF
--- a/source/features/latest-tag-button.tsx
+++ b/source/features/latest-tag-button.tsx
@@ -112,9 +112,8 @@ async function init(): Promise<false | void> {
 	}
 
 	if (currentBranch === latestTag || isUpToDate) {
-		link.classList.add('tooltipped', 'tooltipped-ne');
 		link.setAttribute('aria-label', 'You’re on the latest release');
-		link.classList.add('disabled');
+		link.classList.add('disabled', 'tooltipped', 'tooltipped-ne');
 		return;
 	}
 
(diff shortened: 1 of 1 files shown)


source/features/latest-tag-button.tsx
```diff
@@ -70,10 +69,10 @@ const getRepoPublishState = cache.function(async (): Promise<RepoPublishState> =
 	cacheKey: () => __filebasename + ':' + getRepoURL()
 });
 
-const getAheadByCount = cache.function(async (latestTag: string): Promise<string> => {
-	const tagPage = await fetchDom(`/${getRepoURL()}/releases/tag/${latestTag}`);
+const getAheadByCount = cache.function(async (latestTag: string): Promise<string | undefined> => {
+	const tagPage = await fetchDom(`/${getRepoURL()}/releases/tag/${latestTag}`, '.release-header relative-time + a[href*="/compare/"]');
```
fregante, line 73: Changing the selector without changing the variable name

yakov116: Sorry, I did not even think about it. I wrote this code a few days ago before you made a comment to me yesterday.
I will make sure this does not happen in the future

yakov116: Sorry, I did not even think about it. I wrote this code a few days ago before you made a comment to me yesterday.
I will make sure this does not happen in the future

PR #3181, opened 2020-06-05
Chrome: Open the options source/screenshot links in a new tab

yakov116:
Chrome attempts to open them in the options popup which leads to

image

Firefox does not have this issue

**To be clear:** I am talking about the screenshot and source links


fregante:
Perhaps it's safer to add a listener instead so this doesn't happen to any future links by mistake.

```js
delegate(document, '[href^="http"]', /* preventDefault, window.open()*/)

Issue #3081, opened 2020-05-12
Repo page completely broken when new feature preview enabled


SUPERCILEX:
Enable this feature preview: https://gist.github.com/broccolini/2245234ac3a4936049e8ffc13f376986. Go to a [repo front page](https://github.com/sindresorhus/refined-github) and you'll see a brief flash of content before it's all wiped out.

isaacrlevin:
The New GitHub UI causes tons of errors in console, here is the UI update

more-dropdown → TypeError: Cannot read property 'append' of null
edit-readme → TypeError: Cannot read property 'textContent' of null
repo-age → TypeError: Cannot read property 'append' of null
edit-files-faster → TypeError: Cannot read property 'textContent' of null
latest-tag-button → TypeError: Cannot read property 'textContent' of null

auscompgeek:
Interestingly, I haven't seen RGH wipe out the page completely. (Firefox 76, Refined GitHub 20.5.27)

The console errors I get are different too:

- more-dropdown → TypeError: "parent is null"
- edit-readme → TypeError: "The element wasn’t found, the selector needs to be updated."
- edit-files-faster → TypeError: "The element wasn’t found, the selector needs to be updated."
- repo-age → TypeError: "commitInfo is undefined"

latest-tag-button seems to work for me.

fregante:
`more-dropdown` is kind of a problem now since the new design has an automatic "overflow" hamburger menu:

<img width="588" alt="" src="https://user-images.githubusercontent.com/1402241/84262936-d50d2b00-ab1e-11ea-82da-85002d1e62f1.png">

The obvious solution is to add all those elements to the navigation and let them naturally flow into the overflow, but that's noisy. Until now we added items to the dropdown without much trouble, since it's "out of the way", but this change forces us to reconsider it.

aminya:
I get similar errors in the new design:
![image](https://user-images.githubusercontent.com/16418197/84381779-1fd68380-abaf-11ea-8dd3-49bcfff96519.png)

fregante:
Sizes need to be updated too:

<img width="338" alt="" src="https://user-images.githubusercontent.com/1402241/84260284-13541b80-ab1a-11ea-925f-0db5f75b20d2.png">

I think it's only a matter of adding `if branchselector.is('.btn-sm'), mybutton.classList.add('btn-sm')`

_Originally posted by @fregante in https://github.com/sindresorhus/refined-github/pull/3205#issuecomment-641926371_

[...]

PR #3186, opened 2020-06-06
Restore `releases-tab` in "Repository refresh" beta

FloEdelmann:
Relates to #3081.

With feature preview enabled:

Screenshot_2020-06-06_13-59-03

With feature preview disabled:

Screenshot_2020-06-06_13-59-31

### DIFF
--- a/source/features/releases-tab.tsx
+++ b/source/features/releases-tab.tsx
@@ -62,28 +62,46 @@ async function init(): Promise<false | void> {
 	const releasesTab = (
 		<a href={`/${repoUrl}/releases`} className="reponav-item" data-hotkey="g r">
 			<TagIcon/>
-			<span> Releases </span>
+			<span data-content="Releases"> Releases </span>
 			{count === undefined ? '' : <span className="Counter">{count}</span>}
 		</a>
 	);
 
 	await elementReady('.pagehead + *'); // Wait for the tab bar to be loaded
+
+	const repoNavigationBar = select('.js-repo-nav')!;
+
+	if (repoNavigationBar.classList.contains('UnderlineNav')) {
+		// Adapt releases tab to GitHub's repository redesign
+		releasesTab.classList.remove('reponav-item');
+		releasesTab.classList.add(
+			'js-selected-navigation-item',
+			'UnderlineNav-item',
+			'hx_underlinenav-item',
+			'no-wrap',
+			'js-responsive-underlinenav-item'
+		);
+		select('svg', releasesTab)!.classList.add('UnderlineNav-octicon');
+	}
+
 	appendBefore(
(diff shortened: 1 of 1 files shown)


source/features/releases-tab.tsx
```diff
@@ -16,7 +16,15 @@ const cacheKey = `releases-count:${repoUrl}`;
...
-		return Number(releasesCountElement ? looseParseInt(releasesCountElement.textContent!) : 0);
+		if (releasesCountElement) {
+			return looseParseInt(releasesCountElement.textContent!);
+		}
+
+		// In GitHub's repository redesign, look for the "+ XXX releases" link in the sidebar
+		const moreReleasesCountElement = select('.BorderGrid .text-small[href$="/releases"]');
+		if (moreReleasesCountElement) {
+			return looseParseInt(moreReleasesCountElement.textContent!) + 1;
+		}
```
fregante, line 27: Since we're in `isRepoRoot`, it should return `0` if it's not found.

source/features/releases-tab.tsx
```diff
@@ -52,31 +60,58 @@ async function init(): Promise<false | void> {
...
+			'js-responsive-underlinenav-item'
+		);
+		select('svg', releasesTab)!.classList.add('UnderlineNav-octicon');
+	}
+
 	appendBefore(
-		// GHE doesn't have `reponav > ul`
-		select('.reponav > ul') ?? select('.reponav')!,
+		// GHE doesn't have `.js-repo-nav > ul`
+		select(':scope > ul', repoNavigationBar) ?? repoNavigationBar,
```
fregante, line 89: Are you sure `GHE` currently has `js-repo-nav`?

FloEdelmann: No, I am not.

@busches I think you have access to a GHE instance, right? Can you confirm this?

fregante: Probably isn't necessary since the suggested solution is to completely fork the code, so this section would remain untouched.

yakov116:
This PR broke the releases tab when not on isRepoRoot
```javascript
❌ Refined GitHub → releases-tab → TypeError: Cannot read property 'append' of null
    at init (refined-github.js:13858)
    at async runFeature (refined-github.js:10346)
    at async setupPageLoad (refined-github.js:10360) 
Search issue: https://github.com/sindresorhus/refined-github/issues?q=is%3Aissue+Cannot%20read%20property%20'append'%20of%20null

Open an issue: https://github.com/sindresorhus/refined-github/issues/new?labels=bug&template=bug_report.md&title=%60releases-tab%60%3A%20Cannot%20read%20property%20'append'%20of%20null
```

PR #3187, opened 2020-06-06
Restore `toggle-files-button` in "Repository refresh" beta

FloEdelmann:
Relates to #3081.

With feature preview enabled:

Screenshot_2020-06-06_14-12-13

With feature preview disabled:

Screenshot_2020-06-06_14-12-38

### DIFF
--- a/source/features/toggle-files-button.css
+++ b/source/features/toggle-files-button.css
@@ -12,7 +12,7 @@
 	transform: rotate(0.5turn);
 }
 
-.rgh-files-hidden .files,
+.rgh-files-hidden .files, /* Pre "Repository refresh" layout */
 .rgh-files-hidden #files ~ .Details {
 	display: none;
 }
--- a/source/features/toggle-files-button.css
+++ b/source/features/toggle-files-button.css
@@ -12,7 +12,8 @@
 	transform: rotate(0.5turn);
 }
 
-.rgh-files-hidden .files {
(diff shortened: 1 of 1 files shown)


source/features/toggle-files-button.css
```diff
@@ -12,7 +12,8 @@
 	transform: rotate(0.5turn);
 }
 
-.rgh-files-hidden .files {
+.rgh-files-hidden .files,
```
fregante, line 15: Mark the old style with this code so in the future we know what to drop
```suggestion
.rgh-files-hidden .files, /* Pre "Repository refresh" layout */
```

source/features/toggle-files-button.css
```diff
@@ -12,7 +12,8 @@
 	transform: rotate(0.5turn);
```
fregante, line 12: I think we got this wrong. "Down" means "open" while "up" means closed. You can see this in this very page for each collapsable file.

Do you mind dropping `:not()` from this selector in this PR? It's a minor change

FloEdelmann: In PR Files tabs, "down" means *open* while **"right"** means *closed*. So I think this for this feature, it should be "down" for open and "left" for *closed*.

fregante: You're right. Perhaps we should just use octicon-fold/octicon-unfold instead. In another PR

PR #3189, opened 2020-06-06
`stop-redirecting-in-notification-bar` automatically when possible

### LINKED ISSUE #3173: `stop-redirecting-in-notification-bar` automatically when possible
As discussed in https://github.com/sindresorhus/refined-github/issues/2952 and  https://github.com/sindresorhus/refined-github/pull/3036

I think `stop-redirecting-in-notification-bar` can be done automatically **if the user opened the notification in a new tab.**

For example:

1. disable automatic redirect for all buttons, always
2. try calling `history.back()` (unless `alt` is pressed)

If you opened the notification **in a new tab,** `history.back()` won't have any effect.

IMHO this makes sense since in that case, the notifications list is still open in its own tab.

PR welcome if you test that this works as expected in both browsers.

cc @FloEdelmann @karlhorky @tonglil

karlhorky:
Closes #3173

Feature originally implemented in #2952 and #3036.

### DIFF
--- a/source/features/stop-redirecting-in-notification-bar.tsx
+++ b/source/features/stop-redirecting-in-notification-bar.tsx
@@ -15,7 +15,9 @@ function handleClick(event: delegate.Event<MouseEvent, HTMLButtonElement>): void
 }
 
 async function init(): Promise<void> {
-	// If the history length is `1` on init, the notification has been opened in a new tab
+	// If the history length is `1` on init, that means there are no
+	// older history states (this means the notification has been opened
+	// in a new browser tab)
 	sessionStorage.rghIsNewTab = history.length === 1;
 	delegate(document, '.notification-shelf .js-notification-action button', 'click', handleClick);
 }
--- a/source/features/stop-redirecting-in-notification-bar.tsx
+++ b/source/features/stop-redirecting-in-notification-bar.tsx
@@ -8,16 +8,13 @@ const hasNotificationBar = (): boolean =>
 
 function handleClick(event: delegate.Event<MouseEvent, HTMLButtonElement>): void {
 	// Disable the redirect to the Notifications inbox if either:
-	// 1. The alt key was held down during click
-	// 2. The notification has been opened in a new tab
+	// 1. The alt key was held down during click (user choice)
+	// 2. The notification has been opened in a new tab (the inbox is still open in the previous tab)
 	const redirectDisabled = event.altKey || sessionStorage.rghIsNewTab === 'true';
 	event.delegateTarget.form!.toggleAttribute('data-redirect-to-inbox-on-submit', !redirectDisabled);
 }
 
 async function init(): Promise<void> {
-	// If the history length is `1` on init, that means there are no
-	// older history states (this means the notification has been opened
-	// in a new browser tab)
 	sessionStorage.rghIsNewTab = history.length === 1;
 	delegate(document, '.notification-shelf .js-notification-action button', 'click', handleClick);
 }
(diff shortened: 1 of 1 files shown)


source/features/stop-redirecting-in-notification-bar.tsx
```diff
@@ -7,11 +7,16 @@ const hasNotificationBar = (): boolean =>
 	JSON.parse(sessionStorage.notification_shelf ?? '{}').pathname === location.pathname;
 
 function handleClick(event: delegate.Event<MouseEvent, HTMLButtonElement>): void {
-	// Also restores the attribute on successive non-alt clicks
-	event.delegateTarget.form!.toggleAttribute('data-redirect-to-inbox-on-submit', !event.altKey);
+	// Disable the redirect to the Notifications inbox if either:
+	// 1. The alt key was held down during click
+	// 2. The notification has been opened in a new tab
+	const redirectDisabled = event.alt || sessionStorage.rghIsNewTab === "true";
```
FloEdelmann, line 13:
```suggestion
	const redirectDisabled = event.altKey || sessionStorage.rghIsNewTab === 'true';
```

karlhorky: Oops, thanks @FloEdelmann !

source/features/stop-redirecting-in-notification-bar.tsx
```diff
@@ -7,11 +7,16 @@ const hasNotificationBar = (): boolean =>
...
-	event.delegateTarget.form!.toggleAttribute('data-redirect-to-inbox-on-submit', !event.altKey);
+	// Disable the redirect to the Notifications inbox if either:
+	// 1. The alt key was held down during click
+	// 2. The notification has been opened in a new tab
+	const redirectDisabled = event.altKey || sessionStorage.rghIsNewTab === 'true';
+	event.delegateTarget.form!.toggleAttribute('data-redirect-to-inbox-on-submit', !redirectDisabled);
 }
 
 async function init(): Promise<void> {
+	// If the history length is `1` on init, the notification has been opened in a new tab
```
fregante, line 18: I think the variable name is self-explanatory, no comment needed.

karlhorky: Hmm, disagree here... The name in the line below is super hard to parse and understand.


```js
sessionStorage.rghIsNewTab = history.length === 1;
```

Also - I would like to explain my intent, in case our assumptions are wrong and this leads to a bug in the future.

So I have updated the comment to add more background / motivation information: b935c605660ac00809930533ff15aea9b0ffe83b

karlhorky: Although I would suggest keeping the comment, If for whatever reason this is not acceptable to have guiding comments in the codebase, then I can remove.

fregante: The line says:

- we want to know whether the current tab is a new tab
- we expect to determine this by measuring `history.length`

The comment says the same exact thing, in different words.

Comments are supposed to explain why you do something, not what you're doing bit by bit (unless it's a super complex line — which isn't the case)

fregante:
This works great! Thanks [Kaaaaarl](https://www.youtube.com/watch?v=ZpjyH-LkEAg)

karlhorky:
Thanks for the merge :) 

hehe ah Llamas with Hats, good one! Different languages can be funny too - eg. the German version is where I saw it first

FloEdelmann:
Thanks a lot @karlhorky, I end up using this all the time now :blush:

karlhorky:
@FloEdelmann great, glad you're happy, thanks for the input in the issue!

PR #3191, opened 2020-06-06
Use fold/unfold octicons for `toggle-files-button`

FloEdelmann:
As discussed in <https://github.com/sindresorhus/refined-github/pull/3187#discussion_r436266144>.

### DIFF
--- a/source/features/toggle-files-button.css
+++ b/source/features/toggle-files-button.css
@@ -8,8 +8,9 @@
 	border-radius: 3px;
 }
 
-.repository-content:not(.rgh-files-hidden) .commit-tease .octicon-chevron-down {
-	transform: rotate(0.5turn);
+.repository-content:not(.rgh-files-hidden) .commit-tease .octicon-unfold,
+.repository-content.rgh-files-hidden .commit-tease .octicon-fold {
+	display: none;
 }
 
 .rgh-files-hidden .files, /* Pre "Repository refresh" layout */
--- a/source/features/toggle-files-button.tsx
+++ b/source/features/toggle-files-button.tsx
@@ -3,7 +3,8 @@ import React from 'dom-chef';
 import select from 'select-dom';
 import delegate from 'delegate-it';
 import * as pageDetect from 'github-url-detection';
-import ChevronDownIcon from 'octicon/chevron-down.svg';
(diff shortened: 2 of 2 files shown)


source/features/toggle-files-button.css
```diff
@@ -8,8 +8,9 @@
 	border-radius: 3px;
 }
 
-.repository-content:not(.rgh-files-hidden) .commit-tease .octicon-chevron-down {
-	transform: rotate(0.5turn);
+.repository-content:not(.rgh-files-hidden) .commit-tease .octicon-unfold,
+.repository-content.rgh-files-hidden .commit-tease .octicon-fold {
```
fregante, line 12: Could you check how GitHub does this on comments? Maybe all we need is a class, without extra CSS

FloEdelmann: It seems to be specific to the `<details>` element:

Screenshot_2020-06-06_18-39-06

fregante: To be fair, this whole feature can probably be handled by GitHub, meaning technically we only need to add those `.Detail` classes (which aren't specific to `detail` element) and skip the delegation too. But if you're not interested, no problem, it's just a possible improvement

FloEdelmann: I'm not sure if this will work because I think the functionality is coupled to the actual `<details>` / `<summary>` elements and the classes are only there for styling.

Either way, I think this should be part of a separate PR.

fregante: > I'm not sure if this will work because I think the functionality is coupled to the actual `<details>` / `<summary>` elements and the classes are only there for styling.

Nope, GitHub has this:


```css
.Details--on .Details-content--shown,
.Details:not(.Details--on) .Details-content--hidden {
 display:none!important
}
.Details:not(.Details--on) .Details-content--hidden-not-important {
 display:none
}
```

```js
const n = null != e ? e : !o.classList.contains("open");
o.classList.toggle("open", n),
o.classList.toggle("Details--on", n),
```

> Either way, I think this should be part of a separate PR.

👍

fregante:
I hate the fold/unfold icons in Octicons v2, the arrows are too small. I think we should revert this PR

PR #1939, opened 2019-04-18
Add `edit-files-faster` feature

bfred-it:
Closes https://github.com/sindresorhus/refined-github/issues/758

# Test

https://github.com/sindresorhus/refined-github

Open any repo's root, try changing branches, go back and forth in history, navigate in folders

# Preview

Screenshot 2019-04-18 at 22 57 45

PR #3192, opened 2020-06-06
Restore `edit-files-faster` in "Repository refresh" beta

FloEdelmann:
References #3081.

### DIFF
--- a/source/features/edit-files-faster.tsx
+++ b/source/features/edit-files-faster.tsx
@@ -11,9 +11,9 @@ import getDefaultBranch from '../github-helpers/get-default-branch';
 import onFileListUpdate from '../github-events/on-file-list-update';
 
 async function init(): Promise<void> {
-	const isPermalink = /Tag|Tree/.test(select('.branch-select-menu i')!.textContent!);
-	for (const fileIcon of select.all('.files :not(a) > .octicon-file')) {
-		const fileLink = fileIcon.closest('tr')!.querySelector<HTMLAnchorElement>('.js-navigation-open')!;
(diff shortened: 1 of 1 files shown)


source/features/edit-files-faster.tsx
```diff
@@ -11,9 +11,9 @@ import getDefaultBranch from '../github-helpers/get-default-branch';
 import onFileListUpdate from '../github-events/on-file-list-update';
 
 async function init(): Promise<void> {
-	const isPermalink = /Tag|Tree/.test(select('.branch-select-menu i')!.textContent!);
-	for (const fileIcon of select.all('.files :not(a) > .octicon-file')) {
-		const fileLink = fileIcon.closest('tr')!.querySelector<HTMLAnchorElement>('.js-navigation-open')!;
+	const isPermalink = /Tag|Tree/.test(select('[data-hotkey="w"] i')!.textContent!);
+	for (const fileIcon of select.all('.js-navigation-container .octicon-file')) {
```
fregante, line 15: The `not` avoided wrapping the icon twice. It may not be necessary but it’s worth double-checking, perhaps `blam`ing the line to find discussions around it.

FloEdelmann: It was there since the very first commit: [https://github.com/sindresorhus/refined-github/pull/1939/commits/5accac93731f530fe132ef30654a65afcbae51eb ](https://github.com/sindresorhus/refined-github/pull/1939/commits/5accac93731f530fe132ef30654a65afcbae51eb)

But I don't think it's necessary, because the file list will only be replaced as a whole (triggering the `onFileListUpdate` listener).

PR #3194, opened 2020-06-06
Disable `remove-upload-files-button` in "Repository refresh" beta

FloEdelmann:
It's no longer needed there:

### Before

Screenshot_2020-06-06_19-56-31

### After

Screenshot_2020-06-06_19-56-04


FloEdelmann:
> You can use the previous code and just add “not dropdown-item” to the selector.

Done.

> rather change the dropdown to a single button as part of “ unwrap-useless-dropdowns” instead

Why? I like the new style, it doesn't take much space and still provides both options.

PR #3197, opened 2020-06-07
Use api to get the latest tag ahead count

### LINKED ISSUE #2927: `latest-tag-button` sometimes fails getting the "ahead-by count"

yakov116:
Fixes #2927

@fregante would love your input.

A bit of history of what i thought of doing.

First attempt was to get all the tag release dates in the original call. But github (as usual) does not give it 100% of the time.
So the next thing I did was if there was no date get it from the dom.
But then I realized that you can have commit dates not in order. (Very common by bootstrap on a rebase and merge).

So then I went with the following.

Try to get the ahead count from the dom. If the dom does not have it, take the release date from the dom  and get the aheadcount from the API. With the issue of commit dates not in order, we should not be _that off_.

One issue, if the tag was created locally together with a commit the count will be ahead by one since they both happened the same time. I could fix it, not sure if its worth it.

Let me know what you think.

Test On https://github.com/fregante/delegate-it

### DIFF
--- a/source/features/latest-tag-button.tsx
+++ b/source/features/latest-tag-button.tsx
@@ -1,13 +1,14 @@
 import './latest-tag-button.css';
 import React from 'dom-chef';
 import cache from 'webext-storage-cache';
+import select from 'select-dom';
 import TagIcon from 'octicon/tag.svg';
 import elementReady from 'element-ready';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
-import fetchDom from '../helpers/fetch-dom';
 import * as api from '../github-helpers/api';
+import fetchDom from '../helpers/fetch-dom';
 import GitHubURL from '../github-helpers/github-url';
 import getDefaultBranch from '../github-helpers/get-default-branch';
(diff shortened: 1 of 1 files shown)


yakov116:
Nice try but commits can be our of order 😢

fregante:
Two problems:

- I can tag an old commit **today**. The commit’s date is useless
- So much code

I think we should go with my suggestion: https://github.com/sindresorhus/refined-github/issues/2927#issuecomment-640050034

This way we get the exact match if it's within the last 10/20/whatever commits, and if not, we don't have to guess, we just display "+".

The best part: it's just one API call. Total. This PR has 2 API calls and a `fetchDom`

yakov116:
Agree

PR #3201, opened 2020-06-08
Improve `repo-age`’s reliability

### LINKED ISSUE #3185: AngularJS repository marked as being 51 years old
I know that this is probably caused due to broken unix timestamps in the repository, but it should be possible to exclude those broken timestamps.

[angular/angular.js](https://github.com/angular/angular.js)

yakov116:
1. LINKED ISSUES:
   Fixes #3185

2. TEST URLS:
 
https://github.com/angular/angular.js
https://github.com/torvalds/linux

3. SCREENSHOT:
image


source/features/repo-age.tsx
```diff
@@ -7,15 +7,42 @@ import elementReady from 'element-ready';
...
+				}
+			}
+		}
+	`);
+
+	const {committedDate, resourcePath} = repository.defaultBranchRef.target.history.nodes
+		.reverse()
+		.find(({committedDate}: {committedDate: string}) => {
+			return new Date(committedDate) > new Date('1970-01-01T00:00:01');
+		});
```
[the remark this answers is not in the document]
yakov116: Cool!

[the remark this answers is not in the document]
yakov116: @fregante new Date(0) = 'Dec 31 1969 19:00:00' which is too early. The problematic commits are 1970-01-01T00:00:01Z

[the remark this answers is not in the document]
fregante: > new Date(0) = 'Dec 31 1969 19:00:00'
Nope. Timezones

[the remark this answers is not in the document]
yakov116: Hmm
But when I used your code by me it did not work. How would I set it to use ISO?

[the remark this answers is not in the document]
fregante: The full time for me is

```
Thu Jan 01 1970 01:00:00 GMT+0100 (Central European Standard Time) {}
```

source/features/repo-age.tsx
```diff
@@ -7,15 +7,41 @@ import elementReady from 'element-ready';
...
+					}
+				}
+			}
+		}
+	`);
+
+	const {committedDate, resourcePath} = repository.defaultBranchRef.target.history.nodes
+		.reverse()
+		// Filter out any invalid commit dates #3185
+		.find((commit: AnyObject) => new Date(commit.committedDate) > new Date('1970-01-01T00:00:01Z'));
```
fregante, line 40: Wait, lol. Just use
```suggestion
		.find((commit: AnyObject) => new Date(commit.committedDate) > 0);
```

yakov116: 💥 💥 
Works but typescript does not like it

source/features/repo-age.tsx
```diff
@@ -7,15 +7,41 @@ import elementReady from 'element-ready';
...
+					}
+				}
+			}
+		}
+	`);
+
+	const {committedDate, resourcePath} = repository.defaultBranchRef.target.history.nodes
+		.reverse()
+		// Filter out any invalid commit dates #3185
+		.find((commit: AnyObject) => new Date(commit.committedDate) > 0);
```
FloEdelmann, line 40: What about this then?

```suggestion
		.find((commit: AnyObject) => new Date(commit.committedDate).getTime() > 0);
```

yakov116: Yup!

source/features/repo-age.tsx
yakov116: i was wrong it did not work

fregante: Looks clear as 1970

source/features/repo-age.tsx
```diff
@@ -33,13 +59,7 @@ const getFirstCommit = cache.function(async (): Promise<[string, string] | undef
...
 	}
 
-	const commit = await fetchDom(
-		`${getRepoURL()}/commits?after=${commitSha}+${commitsCount - 2}`,
-		'.commit'
-	);
-	const timeStamp = select('relative-time', commit)!.attributes.datetime.value;
-	const {pathname} = select<HTMLAnchorElement>('a.message', commit)!;
-	return [timeStamp, pathname];
+	return getRepoAge(commitSha, commitsCount - Math.min(6, commitsCount));
```
fregante, line 62: What does this mean? That's 6?

yakov116: We was all the last commit - 6 (Which is 5 its complicated but I thnk 0 and 1 are the same)
I will double check give me a min

yakov116: Yes 0 and 1 are the same

Try it move the 1886 to 1887 and 1888
```
{
    defaultBranchRef {
      target {
        ... on Commit {
          history(first: 5, after: "208835085ee2a78b6a9d4f30b1407c7eac1b88b3 1886") {
            nodes {
              authoredDate
              committedDate
              resourcePath
            }
          }
        }
      }
    }
  }
}
```

fregante: Why 6?

yakov116: We want 5, 1 =0 (I think)

yakov116: OK the math is like this. We are at commit 10 and we want AFTER commit 10 so we need to start from number 9 = (10 - (1 + 5) )

fregante: So the math belongs to `getRepoAge`, not outside.
5 is inside and 6 is outside; doesn't make sense.

```suggestion
	return getRepoAge(commitSha, commitsCount);
```

yakov116: We have commit number 1 we want the last 5. There are 1000 commits. If we do 1 + 1000 (The commit count) we get commit number 1001 which does not exist.
So we need to go back 1 and then another 5 to get the last 5.

fregante: The latest version looks right

fregante:
I think in the Repository refresh, this information should be moved to the sidebar:

yakov116:
I'll leave that for @FloEdelmann. I promise he will do a better job 🥇.

PR #3205, opened 2020-06-08
Restore `bugs-tab` in "Repository refresh" beta

FloEdelmann:
References #3081.

Screenshot_2020-06-08_22-32-22

### DIFF
--- a/source/features/bugs-tab.tsx
+++ b/source/features/bugs-tab.tsx
@@ -58,6 +58,7 @@ async function init(): Promise<void | false> {
 	const bugsLink = select('a', bugsTab)!;
 	bugsLink.removeAttribute('data-hotkey');
 	bugsLink.removeAttribute('data-selected-links');
+	select('a', issuesTab)!.removeAttribute('data-selected-links');
 
 	// Update its appearance
 	const bugsTabTitle = select('[data-content]', bugsTab);
--- a/source/features/bugs-tab.tsx
+++ b/source/features/bugs-tab.tsx
@@ -39,7 +39,7 @@ async function init(): Promise<void | false> {
 		return false;
 	}
 
-	const issuesTab = (await elementReady('.pagehead [data-hotkey="g i"]'))?.parentElement;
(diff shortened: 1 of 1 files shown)


yakov116:
Your on 🔥

fregante:
@FloEdelmann if you want to take care of this issue, I think it's only a matter of adding `if branchselector.is('.btn-sm'), mybutton.classList.add('btn-sm')`

source/features/bugs-tab.tsx
```diff
@@ -39,7 +39,7 @@ async function init(): Promise<void | false> {
 		return false;
 	}
 
-	const issuesTab = (await elementReady('.pagehead [data-hotkey="g i"]'))?.parentElement;
+	const issuesTab = (await elementReady('[data-hotkey="g i"]'))?.parentElement;
```
FloEdelmann, line 42: Unfortunately, they are not unique:

Screenshot_2020-06-10_13-43-27

FloEdelmann:
The `init` function is not called at all when clicking on the browser's back button. That means that the Issues tab is automatically re-selected when navigating back. How can we catch that case?

fregante:
Refer to the previous fix: https://github.com/sindresorhus/refined-github/commit/af843eb243083154adc979663280d98ac14c7c75

In that case our code was fine, but it GitHub was **adding** the class due to `data-selected-links` attribute.

FloEdelmann:
The `data-selected-links` is actually still removed from the Bugs tab: https://github.com/sindresorhus/refined-github/pull/3205/commits/1da152b3f8c51dd72d8ae6ad9cc96e0c2d8716f5#diff-932095cc3c0dff00495b4c392d78f0afR60

I suspect we can't remove it from the Issues tab, right?

If the Issues tab still has it, however, GitHub will select it when going back to the Bugs tab (which is actually the Issues page).

FloEdelmann:
I've now tested removing `data-selected-links` and it indeed fixes the #3001 bug. It also looks like everything is still working as expected, only the Issues tab no longer is rendered via PJAX, but triggers a full page load.

Issue #2900, opened 2020-03-16
AVA tests incompatible with Node 13


fregante:
Node 13 supports ES Modules natively, but the `esm` package monkey patches ES Modules to CommonJS and causes Node to choke when it finds `type: module` in some dependencies.

This issue may depend on any of these:
- https://github.com/avajs/ava/issues/2293
- https://github.com/TypeStrong/ts-node/issues/935
- https://github.com/standard-things/esm/issues/868

PR #3206, opened 2020-06-09
Meta: Fix tests to run (only) on newer Node versions

fregante:
Closes https://github.com/sindresorhus/refined-github/issues/2900

I tried transpiling just the test files, but then they still imported/required TypeScript files and the solution would be to transpile the whole `source` folder as well… or a whole lot of config


notlmn:
IMHO, the divided Node community f'ed up ES Modules support.

fregante:
The Node ESM problem is made worse here by:

- having to support webpack and Node at the same time
- TypeScript

I think one of my least favorite part of this mess is `createRequire`

.github/workflows/test.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
         with:
-          node-version: 12.15 # https://github.com/sindresorhus/refined-github/issues/2900
+          node-version: 14
```
fregante, line 25: AVA refuses to run `.mjs` tests on Node 12. Related: https://github.com/avajs/ava/issues/2432

package.json
```diff
@@ -1,7 +1,7 @@
 {
 	"private": true,
 	"scripts": {
-		"ava": "TS_NODE_FILES=true ava",
+		"ava": "rollup test/*.ts --plugin typescript --dir .built/test --chunkFileNames _.mjs --entryFileNames [name].mjs && ava",
```
fregante, line 4: I tried using webpack (too much config) and Parcel (less config) but for one reason or another I couldn't make the tests pass or the build complete successfully.

Thoughts on using `.built/test` as the output folder?

source/github-helpers/index.ts
```diff
@@ -1,15 +1,14 @@
 import select from 'select-dom';
 import oneTime from 'onetime';
 import compareVersions from 'tiny-version-compare';
-import {isPR, isIssue, utils} from 'github-url-detection';
+import detection from 'github-url-detection';
```
fregante, line 4: For whatever reason, this works even if 'github-url-detection' is not a Node ES Module, but it does require using the default export instead of named exports.

sindresorhus, line 4:
```suggestion
import urlDetection from 'github-url-detection';
```

Just `detection` is too general.

package.json
sindresorhus: > Thoughts on using .built/test as the output folder?

👍

sindresorhus:
Actually, I noticed an error when running this now:

```
refined-github.js:436 Uncaught TypeError: Cannot read property 'utils' of undefined
    at Object.<anonymous> (refined-github.js:436)
    at __webpack_require__ (refined-github.js:10)
    at Module.<anonymous> (refined-github.js:2888)
    at __webpack_require__ (refined-github.js:10)
    at refined-github.js:58
    at refined-github.js:59
```

Which is coming from:

```js
onetime__WEBPACK_IMPORTED_MODULE_1___default()(github_url_detection__WEBPACK_IMPORTED_MODULE_3__["default"].utils.getUsername);
```

fregante:
Indeed, the import was technically incorrect, even if somehow it passed under Rollup’s, TypeScript’s, and Node’s radar.

[#3206@`ca645ced`](https://github.com/sindresorhus/refined-github/pull/3206/commits/ca645cedaa9b45cebe0815ef4536d5ee15a01a3c) fixed the problem, but really it could be avoided by just making the imported package a Node ES module: https://github.com/fregante/github-url-detection/commit/ffb779a60986e98536165bec0eb59b282763ca6f

If the issue reappears for another package, [#3206@`ca645ced`](https://github.com/sindresorhus/refined-github/pull/3206/commits/ca645cedaa9b45cebe0815ef4536d5ee15a01a3c) can be restored.

PR #3208, opened 2020-06-09
Fix URL parsing in Firefox

### LINKED ISSUE #3193: `GitHubURL`: 'pathname' called on an object that does not implement interface URL
This line causes the error in Firefox 77:

https://github.com/sindresorhus/refined-github/blob/8b135715bbd73609a67ffd3ab104b1dbbe39edc1/source/github-helpers/github-url.ts#L19

It appears to happen only for some links, not all. It doesn't matter whether I have the "Repository refresh" beta enabled or disabled. Places where I have spotted it:

### ```edit-files-faster```

https://github.com/sindresorhus/refined-github/blob/8b135715bbd73609a67ffd3ab104b1dbbe39edc1/source/features/edit-files-faster.tsx#L17

<details>
<summary>Sometimes only the last link or even more links are affected, but they seem to always keep working up to some point.</summary>

Screenshot_2020-06-06_18-20-58

</details>

### `edit-readme`

https://github.com/sindresorhus/refined-github/blob/8b135715bbd73609a67ffd3ab104b1dbbe39edc1/source/features/edit-readme.tsx#L20

Sometimes it works just fine, sometimes it errors (in the same repo).

### `latest-tag-button`

https://github.com/sindresorhus/refined-github/blob/8b135715bbd73609a67ffd3ab104b1dbbe39edc1/source/features/latest-tag-button.tsx#L97

Sometimes it works just fine, sometimes it errors (in the same repo).

FloEdelmann:
Closes #3193.

In Firefox 77, inheriting from `URL` seems to cause the errors. This PR works around it by having an internal `URL` object ~~that most fields are proxied to (["Facade pattern"](https://en.wikipedia.org/wiki/Facade_pattern))~~ (["Adapter pattern"](https://en.wikipedia.org/wiki/Adapter_pattern)).

This does indeed fix both errors from #3193: `assign` works again, and there is no console error anymore.

### DIFF
--- a/source/features/deep-reblame.tsx
+++ b/source/features/deep-reblame.tsx
@@ -74,7 +74,7 @@ async function redirectToBlameCommit(event: delegate.Event<MouseEvent, HTMLAncho
 
 	try {
 		blameUrl.branch = await getPullRequestBlameCommit(prCommit, prNumber, blameUrl.filePath);
-		blameUrl.url.hash = 'L' + select('.js-line-number', blameHunk)!.textContent!;
+		blameUrl.hash = 'L' + select('.js-line-number', blameHunk)!.textContent!;
 		location.href = String(blameUrl);
 	} catch (error) {
 		spinner.replaceWith(<VersionIcon/>);
--- a/source/features/deep-reblame.tsx
+++ b/source/features/deep-reblame.tsx
@@ -74,7 +74,7 @@ async function redirectToBlameCommit(event: delegate.Event<MouseEvent, HTMLAncho
 
 	try {
 		blameUrl.branch = await getPullRequestBlameCommit(prCommit, prNumber, blameUrl.filePath);
-		blameUrl.hash = 'L' + select('.js-line-number', blameHunk)!.textContent!;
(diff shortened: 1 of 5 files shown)


fregante:
Too verbose. If `extend` isn't possible, I'd rather just have a read-only `url` property rather than all that.

And really I don't understand why the current solution works... sometimes. Either it works or it doesn't. Is it a browser bug? Should it be reported?

fregante:
Could this be the problem? 

[https://github.com/sindresorhus/refined-github/pull/3153/commits/c01ed3d29773c3f5d1b08004b0c83549393fe1d2 ](https://github.com/sindresorhus/refined-github/pull/3153/commits/c01ed3d29773c3f5d1b08004b0c83549393fe1d2)

or the following 
[https://github.com/sindresorhus/refined-github/pull/3153/commits/effebd55f026752ca6d254577a7cbadd79559734 ](https://github.com/sindresorhus/refined-github/pull/3153/commits/effebd55f026752ca6d254577a7cbadd79559734)

fregante:
Until someone finally figures this out, perhaps your original solution was cleaner. I dropped some unused method

PR #3210, opened 2020-06-10
Docs: Mention Node requirement in contributing.md

### LINKED ISSUE #3202: Node 12+ required to run the extension locally
Hello there,

just noticed that a recent change from https://github.com/sindresorhus/refined-github/pull/3171 is making the extension runable only on Node 12+.

I reckon this is coming from the use of [Object.fromEntries](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries) in `webpack.config.ts` (i.e. https://github.com/sindresorhus/refined-github/blame/fe806803b0a6cc6e473e5e10d00a00fb5571a178/webpack.config.ts#L51 or https://github.com/sindresorhus/refined-github/blame/fe806803b0a6cc6e473e5e10d00a00fb5571a178/webpack.config.ts#L98).

I couldn't see any mention of that requirement anywhere, so I wondered if it had been done on purpose? If it wasn't, maybe it's worth a note somewhere (or we could add an `engines` property in the `package.json` for example)?

arnaudvalle:
Closes #3202 

Add a note about the node version required to run the extension locally.

I ended up adding it just before the workflow section since that's where we start  mentioning commands and scripts to run. I thought it made more sense to keep it together.

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -92,11 +92,7 @@ features.add({
 
 ## Requirements
 
-[Node](https://nodejs.org/en/download/) version 12 or up is required. You can check your current version with:
-
-```sh
-node -v
-```
+[Node.js](https://nodejs.org/en/download/) version 12 or later is required.
 
 ## Workflow
 
--- a/contributing.md
+++ b/contributing.md
@@ -92,7 +92,7 @@ features.add({
 
 ## Requirements
 
-[Node.js](https://nodejs.org/en/download/) version 12 or later is required.
+[Node.js](https://nodejs.org/en/download/) version 13 or later is required.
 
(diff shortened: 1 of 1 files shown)


contributing.md
```diff
@@ -90,6 +90,13 @@ features.add({
 });
 ```
 
+## Requirements
+
+[Node](https://nodejs.org/en/download/) version 12 or up is required. You can check your current version with:
```
sindresorhus, line 95: Node => Node.js

or up => or later


I would drop the instructions on how to check the version. That’s expected knowledge.

PR #3219, opened 2020-06-11
Fix console errors and make `releases-tab` more reliable

yakov116:
Fix releases tab not ever running from api. It returned false and were using ??.

I think it may make more sense to do `pageDetect.isRepoRoot() ? parseCountFromDom() : fetchFromApi()` Not sure

Also fixes attempting to apply a shortcut in `faster-reviews` on locked pr's
Test on https://github.com/yakov116/TestR/pull/6

### DIFF
--- a/source/github-helpers/index.ts
+++ b/source/github-helpers/index.ts
@@ -22,7 +22,7 @@ Example tag content on public repositories: https://github.com/sindresorhus/refi
 Example tag content on private repositories https://github.com/private/private/commits/master.atom?token=AEAXKWNRHXA2XJ2ZWCMGUUN44LM62
 */
 export const getCurrentBranch = (): string => {
-	return new URL(select<HTMLLinkElement>('[type="application/atom+xml"]')!.href)
+	return new URL(select.last<HTMLLinkElement>('[type="application/atom+xml"]')!.href)
 		.pathname
 		.split('/')
 		.slice(4) // Drops the initial /user/repo/route/ part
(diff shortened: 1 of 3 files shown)


yakov116:
I restored the select.last since sometimes there are 2. Watch the console, and the default branch buttons
Current Branch

source/github-helpers/index.ts
```diff
@@ -22,7 +22,7 @@ Example tag content on public repositories: https://github.com/sindresorhus/refi
 Example tag content on private repositories https://github.com/private/private/commits/master.atom?token=AEAXKWNRHXA2XJ2ZWCMGUUN44LM62
 */
 export const getCurrentBranch = (): string => {
-	return new URL(select<HTMLLinkElement>('[type="application/atom+xml"]')!.href)
+	return new URL(select.last<HTMLLinkElement>('[type="application/atom+xml"]')!.href)
```
fregante, line 25: Where do you see 2 `[type="application/atom+xml"]` elements? This doesn't change the result in my case

fregante:
> Also fixes attempting to apply a shortcut in `faster-reviews` on locked pr's

This is the right URL for testing: https://github.com/yakov116/TestR/pull/6/files#submit-review

fregante:
> I think it may make more sense to do `pageDetect.isRepoRoot() ? parseCountFromDom() : fetchFromApi()` Not sure

Yes. Go for it

source/github-helpers/index.ts
yakov116: You saw the gif?

fregante: I totally glossed over it 🤦‍♂️

yakov116:
Did you leave me with anything to do?

fregante:
I was rushing to merge this before the cronjob, but it failed: https://github.com/sindresorhus/refined-github/actions/runs/132226170

Will be fixed by https://github.com/sindresorhus/refined-github/pull/3206 and I'll tag it manually later.

PR #3227, opened 2020-06-12
Use `@primer/octicons-react` icons

FloEdelmann:
I've tried to use the official `@primer/octicons-react` package, as suggested in <https://github.com/sindresorhus/refined-github/pull/3147#issuecomment-643161211>.

I think this is not possible, at least not for now. (Or I am not skilled enough in TypeScript / JSX / React / `dom-chef`.)

1. The default props are not applied.

	<details>
		<summary>In particular the <code>size</code> prop is missing, so all the icons are huge.</summary>

	Screenshot_2020-06-12_21-14-51

	</details>

	<details>
		<summary>Relevant code snippets in <a href="https://www.unpkg.com/browse/@primer/octicons-react@10.0.0/dist/index.esm.js">octicons-react/dist/index.esm.js</a>:</summary>

	Screenshot_2020-06-12_21-13-05

	---

	Screenshot_2020-06-12_21-13-31

	</details>

2. There is no nice way to pass props to an icon function in `dom-chef`, as `<SomeIcon some-prop="foo"/>` is compiled to `createElement(SomeIcon(), {someProp: 'foo'})` instead of `createElement(SomeIcon({someProp: 'foo'}))`.

3. The icons don't have specific classes, i.e. we have to add the classes manually where we need it. E.g. this selector will break:

	https://github.com/sindresorhus/refined-github/blob/4dabb8ce0539222101d39c5329857dafb88327b6/source/features/toggle-files-button.css#L11-L12

Nevertheless, I wanted to share my progress, maybe someone else can figure out how to fix those issues :crossed_fingers:


fregante:
> 3\. The icons don't have specific classes, i.e. we have to add the classes manually where we need it. E.g. this selector will break:

Yeah not worth it then. If we make this change, it should make things easier during development, not harder/more verbose.

FloEdelmann:
I created two PRs in upstream repos:

1. https://github.com/vadimdemedes/dom-chef/pull/63 to support `defaultProps` in `dom-chef` → resolves problem 1
2. https://github.com/primer/octicons/pull/453 to add icon-specific classes to the components' `defaultProps` → resolves problem 3

That would only leave us with problem 2, but we could still override the resulting attributes manually instead of passing custom props:

```jsx
<SomeIcon width="32" height="32" />
<SomeIcon size={32} /> // will wrongly add a `size` attribute
```

Now let's see if those PRs get merged :smile:

fregante:
Thank you for working on this! I hope https://github.com/primer/octicons/pull/453 can be resolved soon

fregante:
Let's get this over with:

Screen Shot 2020-11-27 at 18 50 50

I added a webpack loader to take care of this without having to wait forever for Octicons.

Please check, speak now or forever hold your peace

webpack.config.ts
```diff
@@ -79,6 +80,16 @@ const config: Configuration = {
 	},
 	module: {
 		rules: [
+			{
+				test: /octicons-react\//,
+				loader: 'string-replace-loader',
+				options: {
+					search: /\n(.*)Icon.defaultProps = {\n\s+className: octicon/g,
+					replace: (match: string, name: string) => {
+						return match.replace('octicon', 'octicon octicon-' + decamelize(name, '-'));
```
FloEdelmann, line 89: Maybe add a comment with a link to https://github.com/primer/octicons/pull/453?

webpack.config.ts
```diff
@@ -79,6 +80,16 @@ const config: Configuration = {
 	},
 	module: {
 		rules: [
+			{
+				test: /octicons-react\//,
+				loader: 'string-replace-loader',
+				options: {
+					search: /\n(.*)Icon.defaultProps = {\n\s+className: octicon/g,
```
FloEdelmann, line 87: Doesn't `octicon` needs to quoted? Also, some other small suggestions:

```suggestion
					search: /(\w+)Icon\.defaultProps = {\n\s+className: 'octicon'/g,
```

fregante: Heh you’re right. The linter suggested this change 🤷‍♂️

webpack.config.ts
fregante: We’re all notified when that PR closes. If merged one of us will drop the code. If rejected the comment will become unnecessary.

yakov116 (changes requested):
image
image

fregante:
@yakov116 that was due to the broken regex suggested by the linter

fregante:
```
❯ nr build

        background.js ⏤  18 kB
  browser-polyfill.js ⏤  3.93 kB
          options.css ⏤  1.78 kB
           options.js ⏤  32.5 kB
   refined-github.css ⏤  11.6 kB (+4 B)
    refined-github.js ⏤  168 kB (+44.2 kB)
 resolve-conflicts.js ⏤  1.03 kB

[webpack-cli] Compilation finished
```

Given the file increase, the icons are definitely not being tree-shaken 😬

fregante:
```

❯ nr build

        background.js ⏤  17.2 kB (-404 B)
  browser-polyfill.js ⏤  3.9 kB (+11 B)
          options.css ⏤  1.78 kB
           options.js ⏤  31.7 kB (-340 B)
   refined-github.css ⏤  11.6 kB
    refined-github.js ⏤  110 kB (-55.1 kB)
 resolve-conflicts.js ⏤  1.04 kB (+9 B)

[webpack-cli] Compilation finished
```

fregante (approved):

source/features/forked-to.tsx
```diff
@@ -5,9 +5,11 @@ import select from 'select-dom';
...
 import * as pageDetect from 'github-url-detection';
 
-import ForkIcon from 'octicon/repo-forked.svg';
-import CheckIcon from 'octicon/check.svg';
-import LinkExternalIcon from 'octicon/link-external.svg';
+import {
+	RepoForkedIcon,
+	CheckIcon,
+	LinkExternalIcon
+} from '@primer/octicons-react';
```
fregante, line 12: @yakov116 this or 
```suggestion
import {RepoForkedIcon,	CheckIcon, LinkExternalIcon} from '@primer/octicons-react';
```

yakov116: Single

source/globals.d.ts
```diff
@@ -70,9 +67,10 @@ interface HTMLFormControlsCollection {
 	[key: string]: HTMLInputElement | HTMLTextAreaElement | HTMLButtonElement;
 }
 
-declare module '*.svg' {
-	const Icon = (): JSX.Element => JSX.Element;
-	export default Icon;
+declare module 'react' {
+	const FC = (): JSX.Element => JSX.Element;
+	const React = {FC};
+	export default React;
```
fregante, line 73: I think this isn't needed, right? Doesn't the module already export the correct types?

FloEdelmann: In the `react` module, the return type of functional components is `ReactElement`, not `JSX.Element`. So when I remove the type declaration, I get errors like this:

```
source/github-widgets/notice-bar.tsx:17:6 - error TS2786: 'XIcon' cannot be used as a JSX component.
  Its return type 'ReactElement<any, any> | null' is not a valid JSX element.
    Type 'ReactElement<any, any>' is not assignable to type 'Element'.

17                      <XIcon/>
```

fregante:
Yessss 🙌

yakov116:
>image
image

I still have this when building local.  Any idea why?

FloEdelmann:
@yakov116 `npm i` maybe?

yakov116:
I tried everything I delete node modules, npm update etc

PR #3233, opened 2020-06-14
Improve PR commit 404 page

FloEdelmann:
Add a second `init` function in `useful-not-found-page` that acts on not found PR commits. These can happen if you link to the PR commit (`/repo/pull/number/commits/commit_ref`) instead of the commit itself (`/repo/commit/commit_ref`)<sup>1</sup> and the PR gets force-pushed.

Test: [`32c8a88` (#3227)](https://github.com/sindresorhus/refined-github/pull/3227/commits/32c8a88360a85739f151566eae0225d530ce6a15) (should link to https://github.com/sindresorhus/refined-github/commit/32c8a88360a85739f151566eae0225d530ce6a15)

Unbenannt

<sup>1</sup>: We actually encourage to do so in `prevent-pr-commit-link-loss`, maybe that's the reason why GitHub doesn't use PR commit links by default.

### DIFF
--- a/source/features/patch-diff-links.tsx
+++ b/source/features/patch-diff-links.tsx
@@ -30,7 +30,7 @@ void features.add({
 		pageDetect.isCommit
 	],
 	exclude: [
-		() => document.title.startsWith('Commit range not found · Pull Request')
+		pageDetect.isPRCommit404
 	],
 	init
 });
--- a/source/features/patch-diff-links.tsx
+++ b/source/features/patch-diff-links.tsx
@@ -12,7 +12,8 @@ function init(): void {
 		commitUrl = commitUrl.replace(/\/pull\/\d+\/commits/, '/commit');
 	}
 
-	select('.commit-meta > :last-child')!.append(
+	// Do nothing on not found PR commits (404 pages)
+	select('.commit-meta > :last-child')?.append(
 		<span className="sha-block patch-diff-links">
 			<a href={`${commitUrl}.patch`} className="sha">patch</a>
 			{ ' ' /* Workaround for: JSX eats whitespace between elements */ }
--- a/source/features/patch-diff-links.tsx
+++ b/source/features/patch-diff-links.tsx
@@ -12,8 +12,7 @@ function init(): void {
 		commitUrl = commitUrl.replace(/\/pull\/\d+\/commits/, '/commit');
 	}
 
-	// Do nothing on not found PR commits (404 pages)
-	select('.commit-meta > :last-child')?.append(
+	select('.commit-meta > :last-child')!.append(
 		<span className="sha-block patch-diff-links">
 			<a href={`${commitUrl}.patch`} className="sha">patch</a>
 			{ ' ' /* Workaround for: JSX eats whitespace between elements */ }
@@ -29,6 +28,9 @@ void features.add({
 }, {
 	include: [
 		pageDetect.isCommit
+	],
+	exclude: [
+		() => document.title.startsWith('Commit range not found · Pull Request')
 	],
 	init
 });
--- a/source/features/useful-not-found-page.tsx
+++ b/source/features/useful-not-found-page.tsx
@@ -130,7 +130,10 @@ void features.add({
 	init
 }, {
 	include: [
-		() => document.title.startsWith('Commit range not found · Pull Request')
+		pageDetect.isPRCommit
+	],
+	exclude: [
+		() => !document.title.startsWith('Commit range not found · Pull Request')
 	],
 	waitForDomReady: false,
 	repeatOnAjax: false,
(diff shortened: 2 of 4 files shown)


yakov116:
Me think it should be its own feature. (JMHO)

source/features/useful-not-found-page.tsx
```diff
@@ -118,4 +132,11 @@ void features.add({
 	],
 	repeatOnAjax: false,
 	init
+}, {
+	include: [
+		pageDetect.isPRCommit
+	],
```
yakov116, line 138:
```suggestion
	],
	exclude: [
		() => !document.title.startsWith('Commit range not found · Pull Request')
	],
```

Then there is no need for elementready

FloEdelmann: @yakov116 It's a good idea to detect the page via `document.title`. I changed the `include` instead, though. Do you think we do have to combine that with `isPRCommit` to prevent false positives?

I didn't drop `elementReady`, because it prevents the page from changing after it has loaded. I don't know if that's actually necessary here, but it also doesn't hurt.

yakov116: Its good to have it as an includes and then exclude if it does not match. Since there is no way to know that `document.title.startsWith('Commit range not found etc` = isPRCoomit

FloEdelmann: It is, since it says *PR* and *commit* :wink: 

But there could theoretically be an issue, PR or commit that causes the title to start with that exact string.

yakov116: Could be. But that how we did all excludes until now.

FloEdelmann: I guess you're right, it doesn't hurt to be explicit :)

source/features/patch-diff-links.tsx
```diff
@@ -12,7 +12,8 @@ function init(): void {
 		commitUrl = commitUrl.replace(/\/pull\/\d+\/commits/, '/commit');
 	}
 
-	select('.commit-meta > :last-child')!.append(
+	// Do nothing on not found PR commits (404 pages)
+	select('.commit-meta > :last-child')?.append(
```
yakov116, line 16: You could use the same exclude here instead of the `?`

fregante:
> Me think it should be its own feature. (JMHO)

That's what I thought too, but it feels like it matches the `useful-not-found-page` name, even if technically it's not a 404.

source/features/useful-not-found-page.tsx
```diff
@@ -108,6 +109,15 @@ function init(): false | void {
 	}
 }
 
+async function initPRCommit(): Promise<void | false> {
+	const commitUrl = location.pathname.replace(/\/$/, '').replace(/\/pull\/\d+\/commits/, '/commit');
```
fregante, line 113: I think the "strip slash" can be avoided and the replacement can be slightly simplified
```suggestion
	const commitUrl = location.pathname.replace(/pull\/\d+\/commits/, 'commit');
```

FloEdelmann:
> even if technically it's not a 404

It is, just a different one.

Screenshot_2020-06-15_14-49-49

fregante:
Neat! Now imagine if GitHub showed a proper 404 page like that for everything… 😔

source/features/patch-diff-links.tsx
```diff
@@ -29,5 +29,8 @@ void features.add({
 	include: [
 		pageDetect.isCommit
 	],
+	exclude: [
+		() => document.title.startsWith('Commit range not found · Pull Request')
```
fregante, line 33: Can you also send a PR for this in github-url-detection?

As `isPRCommit404`

source/features/useful-not-found-page.tsx
```diff
@@ -108,6 +109,15 @@ function init(): false | void {
 	}
 }
 
+async function initPRCommit(): Promise<void | false> {
```
fregante, line 112: It never returns `false`

```suggestion
async function initPRCommit(): Promise<void> {
```

source/features/patch-diff-links.tsx
FloEdelmann: Done: https://github.com/fregante/github-url-detection/pull/12

source/features/useful-not-found-page.tsx
```diff
@@ -108,6 +109,15 @@ function init(): false | void {
 	}
 }
 
+async function initPRCommit(): Promise<void> {
+	const commitUrl = location.pathname.replace(/pull\/\d+\/commits/, 'commit');
```
fregante, line 113: Feels pretty straight-forward since we already have `is404` in the same file

```suggestion
async function initPRCommit(): Promise<void | false> {
	const commitUrl = location.href.replace(/pull\/\d+\/commits/, 'commit');
	if (await is404(commitUrl)) {
		return false;
	}
```

How does this sound? Does it work?

FloEdelmann: Nice!

Issue #853, opened 2017-12-08
Keyboard shortcuts to navigate and collapse files in diffs


hkdobrev:
It would be lovely if we can use <kbd>j</kbd>/<kbd>k</kbd> to go through files in the "Files changed" tab and use <kbd>Spacebar</kbd> to toggle collapsing.

![image](https://user-images.githubusercontent.com/506129/33761728-bbb9349c-dc12-11e7-94f6-e8e850d10dad.png)

bfred-it:
For reference, it's there but it doesn't work:

<img width="557" alt="shortcuts in pr view" src="https://user-images.githubusercontent.com/1402241/33779617-9a0789a4-dc88-11e7-84a3-c600e1e3f8d5.png">

sindresorhus:
*For anyone that wants this, please also send this feature request to GitHub: support@github.com*

yakov116:
@fregante can we combine this issue with #3319

fregante:
That issue is for bugs, this is an enhancement

fregante:
4 years are enough, sorry peeps 😅

yannxaver:
Are there any intentions to implement this? I think this would be an immensely useful feature :)

cheap-glitch:
> Are there any intentions to implement this? I think this would be an immensely useful feature :)

Moving through files in diffs with <kbd>j</kbd> / <kbd>k</kbd> is already implemented in `keyboard-navigation`. It just needs to support <kbd>Space</kbd> for toggling.

@yannxaver Please open a new issue if you request this so we can track it. It should be relatively easy to implement, feel free to take a stab at it :)

fregante:
@cheap-glitch the issue already exists, it's the PR that's missing. We don't need to *track* this

PR #3235, opened 2020-06-15
Enable `ctrl+enter` to publish the current release notes

yakov116:
1. LINKED ISSUES:
   Closes #2696

2. TEST URLS:
   sindresorhus/refined-github/releases/new

3. SCREENSHOT:
   None

Note `pageDetect.isReleasesOrTags` is needed since if there is an error publishing the release it does not stay on the `new` page

PR #3242, opened 2020-06-17
Add `keyboard-navigation` feature

### LINKED ISSUE #1270: Navigating Issues/PRs with keyboard shortcuts
- Up/down comments - j/k (not up and down arrows for a11y reasons)
- New comment - n
- When comment focused:
  - Resolve comment - r
  - Delete comment - d
  - Edit comment - e
  - Reactions?
    - 👍 `+` (=)
    - 👎  `-`

yakov116:
1. LINKED ISSUES:
   Closes #1270

2. TEST URLS:
   https://github.com/sindresorhus/refined-github/issues/1270

3. SCREENSHOT:
   None

Adds the following shortcuts:

		j: 'Move up a comment',
		k: 'Move down a comment',
		e: 'Edit the focused comment',
		d: 'Delete the focused comment'


source/features/comment-shortcuts.tsx
```diff
@@ -0,0 +1,92 @@
...
+	delegate(document, '.timeline-comment.unminimized-comment, .review-comment', 'click', focusComment);
+	document.addEventListener('keypress', event => {
+		if (!shortcutKeys.has(event.key) || event.target instanceof HTMLTextAreaElement || event.target instanceof HTMLInputElement) {
+			return;
+		}
+
+		event.preventDefault();
+
+		if (['j', 'k'].includes(event.key)) {
+			const items = select.all<HTMLAnchorElement>('.timeline-comment-group[id^="issue"]:not([href])');
```
notlmn, line 39: This could be changed to `.js-targetable-element`, that way, you can target review comments too.

PS: You may have to filter out some items that you do not want to target.

/cc @fregante

source/features/comment-shortcuts.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+			return;
+		}
+
+		if (event.key === 'n') {
+			select<HTMLTextAreaElement>('#new_comment_field')?.focus();
+			return;
+		}
+
+		if (/^#issue|^#discussion_/.test(location.hash)) {
```
notlmn, line 67:
```suggestion
		if (/^#issue-|^pullrequestreview-|^#discussion_/.test(location.hash)) {
```

If we were to target review comments.

source/features/comment-shortcuts.tsx
yakov116: My issue was that how do I stay in the same review? (I would not want it to go to the next review as that could be totally unrelated)

yakov116:
I added support for these https://github.com/sindresorhus/refined-github/pull/3235#pullrequestreview-430614941-body-html

image

yakov116:
Watch the URL hash. As you can see when there is a target and then you **click** on another comment the hash goes blank (if you click the comment again then the hash is filled in)

Video_2020-06-17_092031

source/features/comment-shortcuts.tsx
notlmn: We do get all comments in order when querying DOM, we can treat everything as a "comment".

How exactly is this an issue, can you please explain any unwanted behavior you are seeing here.

fregante: > My issue was that how do I stay in the same review?

The answer is that you don't. j/k only mean previous/next on the current page. We're not going to implement boundaries.

If it's a comment, we should focus it.

source/features/comment-shortcuts.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+import features from '.';
+
+const shortcutKeys = new Set(['e', 'd', 'h', 'j', 'k', '+', '-', 'n']);
+const shortcutClass: Record<string, string> = {
+	e: '.rgh-edit-comment, [aria-label="Edit comment"]',
+	d: '[aria-label="Delete comment"]',
+	h: '[aria-label="Hide comment"]',
+	'+': '[data-reaction-label="+1"]',
+	'-': '[data-reaction-label="-1"]'
```
fregante, line 13: I'd drop these honestly

source/features/comment-shortcuts.tsx
```diff
@@ -0,0 +1,100 @@
+import select from 'select-dom';
+import delegate from 'delegate-it';
+import * as pageDetect from 'github-url-detection';
+
+import features from '.';
+
+const shortcutKeys = new Set(['e', 'd', 'h', 'j', 'k', '+', '-', 'n']);
+const shortcutClass: Record<string, string> = {
```
fregante, line 8: Should probably be a `Map`
```suggestion
const shortcutClass = new Map<string, string>([
```

source/features/comment-shortcuts.tsx
```diff
@@ -0,0 +1,100 @@
...
+function focusComment({delegateTarget: comment}: delegate.Event<MouseEvent, HTMLButtonElement>): void {
+	if (select.exists(':target')) {
+		console.log('I dont work correctly');
+	}
+
+	history.replaceState({}, document.title, commentHash(comment));
+}
+
+function init(): void {
+	delegate(document, '.timeline-comment.unminimized-comment, .review-comment', 'click', focusComment);
```
fregante, line 38: If I understand this correctly, there's no need for this. Just click the permalink if you want to focus a comment

source/features/comment-shortcuts.tsx
```diff
@@ -0,0 +1,100 @@
...
+	document.addEventListener('keypress', event => {
```
fregante, line 39: No inline event handlers

source/features/comment-shortcuts.tsx
```diff
@@ -0,0 +1,100 @@
...
+		if (!shortcutKeys.has(event.key) || event.target instanceof HTMLTextAreaElement || event.target instanceof HTMLInputElement) {
```
fregante, line 40: `shortcutKeys` is duplicate. Just use `shortcutClass`

```suggestion
		if (!shortcutClass.has(event.key) || event.target instanceof HTMLTextAreaElement || event.target instanceof HTMLInputElement) {
```

source/features/comment-shortcuts.tsx
```diff
@@ -0,0 +1,100 @@
...
+				block: 'start'
+			});
+
+			return;
+		}
+
+		if (event.key === 'n') {
+			select<HTMLTextAreaElement>('#new_comment_field')?.focus();
+			return;
+		}
```
fregante, line 73: No need, really. `r` already works, even if it copies the selected text if any.

source/features/comment-shortcuts.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+		event.preventDefault();
+
+		if (['j', 'k'].includes(event.key)) {
+			const items = select.all<HTMLAnchorElement>('.timeline-comment-group[id^="issue"]:not([href]), .timeline-comment-group[id^="pullrequestreview-"]:not([href])');
+			// `j` goes to the next comment `k` goes back a comment
+			const direction = event.key === 'j' ? 1 : -1;
+			// Find current
+			const currentComment = /^#issue|^#pullrequestreview-/.test(location.hash) ? select<HTMLAnchorElement>(location.hash)! : select<HTMLAnchorElement>(':target')!;
```
fregante, line 51: Why is this necessary instead of the suggested
```suggestion
			const currentComment = select<HTMLAnchorElement>(':target');
```

source/features/comment-shortcuts.tsx
```diff
@@ -0,0 +1,100 @@
...
+			}
+
+			// Find chosen
+			const chosen = items[Math.min(currentIndex + direction, items.length - 1)]; // Clamp it so it cant go past the last one
+
+			// Focus comment and dont put into history
+			location.replace(commentHash(chosen));
+			chosen.scrollIntoView({
+				block: 'start'
+			});
```
fregante, line 65: It's the default? https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView

```suggestion
			chosen.scrollIntoView();
```

source/features/comment-shortcuts.tsx
```diff
@@ -0,0 +1,100 @@
...
+		}
+
+		if (event.key === 'n') {
+			select<HTMLTextAreaElement>('#new_comment_field')?.focus();
+			return;
+		}
+
+		if (/^#issue|^#pullrequestreview-|^#discussion_/.test(location.hash)) {
+			triggerShortcut(event.key);
+		}
```
fregante, line 77: Just call it. There's already a `?.` in there.

Actually just inline it here
```suggestion
	select(':target')?.querySelector<HTMLButtonElement>(shortcutClass[shortcut])?.click();
```

source/features/comment-shortcuts.tsx
yakov116: If I drop the focus comment then your suggestion is correct

source/features/comment-shortcuts.tsx
yakov116: But skip I wil skip the minimized comments/reviews

source/features/comment-shortcuts.tsx
yakov116: I thought it was cool, but I dropped it.

yakov116:
I tested it looks like its working. Will work on in tomorrow morning some more

source/features/comment-shortcuts.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+const shortcutClass = new Map<string, string>([
+	['e', '.rgh-edit-comment, [aria-label="Edit comment"]'],
+	['d', '[aria-label="Delete comment"]'],
+	['h', '[aria-label="Hide comment"]'],
+	['j', ''],
+	['k', '']
+]);
+
+function commentHash(comment: HTMLElement): string {
```
fregante, line 15: This function feels like it should not be necessary.

`items` should only includes directly-focusable elements, not its parents/ancestors, so the right hash should always be just `chosen.hash`

source/features/comment-shortcuts.tsx
```diff
@@ -0,0 +1,79 @@
...
+	if (['j', 'k'].includes(event.key)) {
+		const items = select.all<HTMLAnchorElement>([
+			'.timeline-comment-group[id^="issue"]:not([href])', // Regular comments
+			'.timeline-comment-group[id^="pullrequestreview-"]:not([href])', // Base review comments (Approved/ChangesRequested)
+			'.review-comment.js-minimizable-comment-group' // Review comments
+		]);
+		// `j` goes to the next comment `k` goes back a comment
+		const direction = event.key === 'j' ? 1 : -1;
+		// Find current
+		const currentComment = select<HTMLAnchorElement>(':target')!;
```
fregante, line 39: Does this work? If not, it's ok.
```suggestion
		const currentComment = select(':target')!;
```

source/features/comment-shortcuts.tsx
```diff
@@ -0,0 +1,79 @@
...
+import * as pageDetect from 'github-url-detection';
+
+import features from '.';
+
+const shortcutClass = new Map<string, string>([
+	['e', '.rgh-edit-comment, [aria-label="Edit comment"]'],
+	['d', '[aria-label="Delete comment"]'],
+	['h', '[aria-label="Hide comment"]'],
+	['j', ''],
+	['k', '']
```
fregante, line 12: These can be dropped and moved exactly where needed (1/4)

```suggestion
	['h', '[aria-label="Hide comment"]']
```

source/features/comment-shortcuts.tsx
```diff
@@ -0,0 +1,79 @@
...
+	const reviewComment = comment.closest('.js-comment[id^="pullrequestreview-"]');
+	const {hash} = reviewComment ? reviewComment.querySelector<HTMLAnchorElement>('a.js-timestamp')! : comment.querySelector<HTMLAnchorElement>('a.js-timestamp')!;
+
+	return reviewComment ? hash + '-body-html' : hash;
+}
+
+function runShortcuts(event: delegate.Event<KeyboardEvent>): void {
+	if (!shortcutClass.has(event.key) || event.target instanceof HTMLTextAreaElement || event.target instanceof HTMLInputElement) {
+		return;
+	}
```
fregante, line 26: 2/4

```suggestion
	if (isEditable(event.target)) {
		return;
	}
```

source/features/comment-shortcuts.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+	event.preventDefault();
+
+	if (['j', 'k'].includes(event.key)) {
```
fregante, line 30: 3/4
```suggestion
	if (['j', 'k'].includes(event.key)) {
		event.preventDefault();
```

source/features/comment-shortcuts.tsx
```diff
@@ -0,0 +1,79 @@
...
+		// Find chosen
+		const chosen = items[Math.min(currentIndex + direction, items.length - 1)]; // Clamp it so it cant go past the last one
+
+		// Focus comment and dont put into history
+		location.replace(commentHash(chosen));
+		chosen.scrollIntoView();
+		return;
+	}
+
+	select(':target')?.querySelector<HTMLButtonElement>(shortcutClass.get(event.key)!)?.click();
```
fregante, line 55: 4/4
```suggestion
	const actionClass = shortcutClass.get(event.key);
	if (actionClass) {
		event.preventDefault();
		select(':target')?.querySelector<HTMLButtonElement>(actionClass)?.click();
	}
```

Also, is `HTMLButtonElement` needed?

source/features/comment-shortcuts.tsx
```diff
@@ -0,0 +1,79 @@
...
+}
+
+function init(): void {
+	delegate(document, '*', 'keypress', runShortcuts);
```
fregante, line 59: If you don't use the selector, this is equivalent and lighter:

```suggestion
	document.addEventListener('keypress', runShortcuts);
```

source/features/comment-shortcuts.tsx
yakov116: image

The issue is we need number 1 but need to focus number 2. This is the only case we need it for

source/features/comment-shortcuts.tsx
yakov116: Unrelated I took the code from here.
https://github.com/sindresorhus/refined-github/blob/0fa514dbeeab6395ef095cff80beac41443a6574/source/features/improve-shortcut-help.tsx#L50

We should do the same there?

source/features/comment-shortcuts.tsx
yakov116: We need to prevent default. Otherwise when doing <kbd>e</kbd> for edit the `e` will be placed in the comment

source/features/comment-shortcuts.tsx
yakov116: Doesnt work

source/features/comment-shortcuts.tsx
yakov116: > Also, is `HTMLButtonElement` needed?

Yes for the `.click()`

yakov116:
I need to figure out how to skip the minimized comments.

Will try again soon

yakov116:
@fregante are you able to leave a `requested changes` and minimize the comment? (Like this I have all 3 kinds of minimized comments)

source/features/comment-shortcuts.tsx
```diff
@@ -0,0 +1,81 @@
+import select from 'select-dom';
+import * as pageDetect from 'github-url-detection';
+
+import features from '.';
+import {isEditable} from '../helpers/dom-utils';
+
+const shortcutClass = new Map<string, string>([
```
fregante, line 7: Minimized comment

fregante: This, however it's unrelated to the type of review

yakov116: Perfect working great

yakov116:
@fregante I spent about 2 hours trying to figure out another solution.

I need the popup to open for review comments and the hide to work.

source/features/comment-shortcuts.tsx
fregante: Sure

source/features/comment-shortcuts.tsx
fregante: Why do you _need_ number 1?

I think you only need exactly what you're about to focus.

yakov116: That is where the hash of the comment is

yakov116: Regular comments and review comments work one way requested changes and approve work a drop different

image

fregante: I checked, your 1/2 comment was incorrect. 

1. Visit https://github.com/sindresorhus/refined-github/pull/3242#pullrequestreview-433511748
2. Run `$(':target')` in the console.

Neither the 1 nor the 2 are the `:target`, so you're selecting the wrong `items` below

fregante: `a.js-timestamp` is a completely unrelated element and should not be selected.

What this needs to select is the focusable element itself, **not** a link that points to it.

There can be 10 `a.js-timestamp` that point to the same element and they're irrelevant to navigation.

source/features/comment-shortcuts.tsx
```diff
@@ -0,0 +1,97 @@
...
+		location.replace(commentHash(chosen));
+		chosen.scrollIntoView();
+		return;
+	}
+
+	const actionClass = shortcutClass.get(event.key);
+	if (actionClass) {
+		event.preventDefault();
+		const target = select(':target');
+		if (!target || location.hash === '#partial-timeline') {
```
fregante, line 57: This fails for any other `#hash`

The proper check is to verify that `:target` is a supported one like we do with the `indexOf` above. You might have to extract a `getItems` function so it it can be reused

source/features/comment-shortcuts.tsx
yakov116: Let me try to explain. I spent a really long time trying to find unique thing to each comment and it should be easy to go to it.

From what I found it would be the comment time, since that has a hash. That hash takes us to the comment. **except for** requested changes and approve. those take us to the top part not the comment it self. We want to focus the comment.

I hope I am clear enough

source/features/comment-shortcuts.tsx
```diff
@@ -0,0 +1,97 @@
...
+			return;
+		}
+
```
fregante, line 60: - Edit and Delete don't need the popup to be open. Skip the popup in that case
- **Let's avoid "Resolve" altogether for now** so that this feature can be merged

source/features/comment-shortcuts.tsx
yakov116: Let me try to explain. I spent a really long time trying to find unique thing to each comment and it should be easy to go to it.

However the comment hash on a requested changes and approve is above the comment

source/features/comment-shortcuts.tsx
yakov116: ok

yakov116:
I am down to one thing.


If I dont use a popup then when trying to edit review comments I get the following
```
index.esm.js:3 Uncaught QueryError: Element not found: <HTMLElement> .js-comment
    at c (https://github.githubassets.com/assets/vendor-8df47ee1.js:1:371)
    at HTMLButtonElement.<anonymous> (https://github.githubassets.com/assets/github-bootstrap-46ed5ea7.js:1:24641)
    at HTMLDocument.st (https://github.githubassets.com/assets/vendor-8df47ee1.js:120:5054)
```

source/features/comment-shortcuts.tsx
yakov116: i am dead wrong here. i had one long day. Sorry

source/features/comment-shortcuts.tsx
yakov116: I removed this crazy hash thing. What a long day...

fregante:
`edit-comments-faster` moves the edit button outside the popup, (and therefore you definitely **not** need the popup to be open) so your selector might be looking in the wrong place

yakov116:
> `edit-comments-faster` moves the edit button outside the popup, (and therefore you definitely **not** need the popup to be open) so your selector might be looking in the wrong place

My fix is working. I dont under why I even need the fix!!

I will give it a go tomorrow.

Side note: This has to work even with `edit comments faster` disabled.

yakov116:
@fregante the following would close https://github.com/sindresorhus/refined-github/issues/853

```jsx
		let items;
		if (location.hash.startsWith('#diff-')) {
			// Include the target, Since after a comment is minimized it will not have the `Details--on` class
			items = select.all(':target, .Details--on[id^="diff-"].open');
		} else {
			items = select.all('.js-minimizable-comment-group')
				.filter(comment => !comment.querySelector('.minimized-comment:not(.d-none)'));
		}
```

we can even add `x` with `$(':target').querySelector('[aria-label="Toggle diff contents"]').click()`

New feature?

yakov116:
@FloEdelmann want to help review this?

source/features/comment-shortcuts.tsx
```diff
@@ -0,0 +1,69 @@
...
+	if (isEditable(event.target) || !focusedComment) {
+		return;
+	}
+
+	if (['j', 'k'].includes(event.key)) {
+		event.preventDefault();
+
+		const items = select.all('.js-minimizable-comment-group')
+			.filter(comment => !comment.querySelector('.minimized-comment:not(.d-none)'));
+		// `j` goes to the next comment `k` goes back a comment
```
FloEdelmann, line 23:
```suggestion
		// `j` goes to the next comment, `k` goes back a comment
```

source/features/comment-shortcuts.tsx
```diff
@@ -0,0 +1,69 @@
...
+		const direction = event.key === 'j' ? 1 : -1;
+
+		const currentIndex = items.indexOf(focusedComment);
+		// Nothing selected or we are on the first comment
+		if (currentIndex + direction < 0 || currentIndex === -1) {
```
FloEdelmann, line 28: I'd swap these to match the comment.

```suggestion
		if (currentIndex === -1 || currentIndex + direction < 0) {
```

source/features/comment-shortcuts.tsx
```diff
@@ -0,0 +1,69 @@
...
+			return;
+		}
+
+		// Find chosen and clamp it so it cant go past the last one
+		const chosen = items[Math.min(currentIndex + direction, items.length - 1)];
+
+		// Focus comment and dont put into history
+		location.replace('#' + chosen.id);
+		// Avoid the extra jump
+		chosen.scrollIntoView();
```
FloEdelmann, line 38:
```suggestion
		const chosenCommentIndex = Math.min(currentIndex + direction, items.length - 1);

		// Focus comment without pushing to history
		location.replace('#' + items[chosenCommentIndex].id);
		// Avoid the extra jump
		items[chosenCommentIndex].scrollIntoView();
```

yakov116:
@FloEdelmann Thanks!

source/features/comment-shortcuts.tsx
```diff
@@ -0,0 +1,69 @@
...
+		const direction = event.key === 'j' ? 1 : -1;
+
+		const currentIndex = items.indexOf(focusedComment);
+		// Nothing selected or we are on the first comment
+		if (currentIndex === -1 || currentIndex + direction < 0) {
+			return;
+		}
+
+		// Find chosen and clamp it so it cant go past the last one
+		const chosenCommentIndex = Math.min(currentIndex + direction, items.length - 1);
```
fregante, line 33:
```suggestion
		const currentIndex = items.indexOf(focusedComment);

		// Start at 0 if nothing is; clamp index
		const chosenCommentIndex = Math.min(
			Math.max(0, currentIndex + direction),
			items.length - 1
		);
```

source/features/comment-shortcuts.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+		// Focus comment without pushing to history
+		location.replace('#' + items[chosenCommentIndex].id);
+		// Avoid the extra jump
+		items[chosenCommentIndex].scrollIntoView();
```
fregante, line 38: The first comment works for both

```suggestion
		// Focus comment without pushing to history
		location.replace('#' + items[chosenCommentIndex].id);
		items[chosenCommentIndex].scrollIntoView();
```

source/features/comment-shortcuts.tsx
```diff
@@ -0,0 +1,69 @@
...
+const shortcutClass = new Map<string, string>([
+	['e', '.unminimized-comment [aria-label="Edit comment"]:not([hidden])'],
+	['d', '[aria-label="Delete comment"]']
+]);
+
+function runShortcuts(event: KeyboardEvent): void {
+	const focusedComment = select(':target');
+	if (isEditable(event.target) || !focusedComment) {
+		return;
+	}
```
fregante, line 16: If you exit here, then `currentIndex` will never be `-1`

```suggestion
	if (isEditable(event.target)) {
		return;
	}
	
	const focusedComment = select(':target');
```

If a comment isn't selected, j/k will select one.

source/features/comment-shortcuts.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+	if (['j', 'k'].includes(event.key)) {
+		event.preventDefault();
+
+		const items = select.all('.js-minimizable-comment-group')
+			.filter(comment => !comment.querySelector('.minimized-comment:not(.d-none)'));
```
fregante, line 22: Note to self: verify these selectors

yakov116:
Before I commit, double checking from what I see.

You want that if no comment is selected it should start from the first comment on the page?

source/features/comment-shortcuts.tsx
yakov116: https://github.com/sindresorhus/refined-github/blob/0fa514dbeeab6395ef095cff80beac41443a6574/source/features/preview-hidden-comments.tsx#L13

fregante:
> You want that if no comment is selected it should start from the first comment on the page?

Just merge my suggested change and it should work without anything else.

source/features/comment-shortcuts.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+		// Start at 0 if nothing is; clamp index
+		const chosenCommentIndex = Math.min(
+			Math.max(0, currentIndex + direction),
+			items.length - 1
+		);
+
+		// Focus comment without pushing to history
+		location.replace('#' + items[chosenCommentIndex].id);
+		items[chosenCommentIndex].scrollIntoView();
```
fregante, line 37: Can you make a gif showing the difference between having this and not having this? `location.replace` seems to already to it.

readme.md
```diff
@@ -180,6 +180,7 @@ Thanks for contributing! 🦋🙌
 - [](# "minimize-upload-bar") [Reduces the upload bar to a small button.](https://user-images.githubusercontent.com/55841/59802383-3d994180-92e9-11e9-835d-60de67611c30.png)
 - [](# "clean-rich-text-editor") [Hides unnecessary comment field tooltips and toolbar items](https://user-images.githubusercontent.com/1402241/53629083-a4fe8900-3c47-11e9-8211-bfe2d254ffcb.png) (each one has a keyboard shortcut.)
 - [](# "monospace-textareas") Use a monospace font for all textareas.
+- [](# "comment-shortcuts") Adds shortcuts to comments: <kbd>j</kbd> to move down a comment; <kbd>k<kbd> to move up a comment.; <kbd>e</kbd> to edit a comment; <kbd>d</kbd> to delete a comment.
```
fregante, line 183: I'd call this feature `navigate-comments` and in the future we can have another feature called `navigate-pr-files` (or if possible they can be merged)

source/features/comment-shortcuts.tsx
```diff
@@ -0,0 +1,68 @@
...
+		return;
+	}
+
+	const focusedComment = select(':target')!;
+
+	if (['j', 'k'].includes(event.key)) {
+		event.preventDefault();
+
+		const items = select.all('.js-minimizable-comment-group')
+			.filter(comment => !comment.querySelector('.minimized-comment:not(.d-none)'));
```
fregante, line 23: How about this?
```suggestion
		const items = select.all('.js-targetable-element');
```

source/features/comment-shortcuts.tsx
fregante: How about this?
```suggestion
		const items = select.all('.js-targetable-element');
```

source/features/comment-shortcuts.tsx
fregante: Related: GitHub adds an annoying and buggy double-scroll that happens every time a comment is focused; this is unrelated to us and doesn't seem to be solvable (but I wish it could be)

source/features/comment-shortcuts.tsx
```diff
@@ -0,0 +1,68 @@
...
+	}
+}
+
+function init(): void {
+	document.addEventListener('keypress', runShortcuts);
+}
+
+void features.add({
+	id: __filebasename,
+	description: 'Adds shortcuts to comments: `j` to move down a comment; `k` to move up a comment.; `e` to edit a comment; `d` to delete a comment',
```
fregante, line 54:
```suggestion
	description: 'Adds shortcuts to comments: `j` focuses the comment below; `k` focuses comment above; `e` edits the focused comment; `d` deletes the focused comment.',
```

source/features/comment-shortcuts.tsx
fregante: This selector also works on on PR files, so I'd rename the feature to `keyboard-navigation`

yakov116: But we still need to filter out the minimized comments. 
IMHO opening a minimized comment is counter productive

What do you think of my idea https://github.com/sindresorhus/refined-github/pull/3242#issuecomment-647141375?

source/features/comment-shortcuts.tsx
yakov116: I am not good at these things hopefully you can tell from the gif.

the scroll to void avoids a second jump.

Video_2020-06-22_081805

readme.md
```diff
@@ -180,7 +180,7 @@ Thanks for contributing! 🦋🙌
 - [](# "minimize-upload-bar") [Reduces the upload bar to a small button.](https://user-images.githubusercontent.com/55841/59802383-3d994180-92e9-11e9-835d-60de67611c30.png)
 - [](# "clean-rich-text-editor") [Hides unnecessary comment field tooltips and toolbar items](https://user-images.githubusercontent.com/1402241/53629083-a4fe8900-3c47-11e9-8211-bfe2d254ffcb.png) (each one has a keyboard shortcut.)
 - [](# "monospace-textareas") Use a monospace font for all textareas.
-- [](# "comment-shortcuts") Adds shortcuts to comments: <kbd>j</kbd> to move down a comment; <kbd>k<kbd> to move up a comment.; <kbd>e</kbd> to edit a comment; <kbd>d</kbd> to delete a comment.
+- [](# "keyboard-navigation") Adds shortcuts to comments: <kbd>j</kbd> focuses the comment below; <kbd>k</kbd> focuses comment above; <kbd>e</kbd> edits the focused comment; <kbd>d</kbd> deletes the focused comment.
```
fregante, line 183:
```suggestion
- [](# "keyboard-navigation") Adds shortcuts to comments: <kbd>j</kbd> focuses the comment below; <kbd>k</kbd> focuses the comment above; <kbd>e</kbd> edits the focused comment; <kbd>d</kbd> deletes the focused comment.
```

source/features/keyboard-navigation.tsx
```diff
@@ -54,8 +54,8 @@ void features.add({
 	description: 'Adds shortcuts to comments: `j` focuses the comment below; `k` focuses comment above; `e` edits the focused comment; `d` deletes the focused comment.',
```
fregante, line 54:
```suggestion
	description: 'Adds shortcuts to comments: `j` focuses the comment below; `k` focuses the comment above; `e` edits the focused comment; `d` deletes the focused comment.',
```

source/features/comment-shortcuts.tsx
fregante: That issue appears regardless of that line in my experience (Chrome)

yakov116: Me too but there a second jump (That I noticed) without the scroll to view.

I have no issue to removing it

source/features/comment-shortcuts.tsx
fregante: I think feature should (eventually, if not now) be able to unhide those comments as well, or at least expand them.

Perhaps you should still use that “targetable” selector and run the filter to exclude what we don’t want (e.g. hidden comments, at the moment)

yakov116: When using the tagetable selector all the pr files are included in "items".

From my understanding of what you want is: `j` `k` should not select files only `o` and `v`

source/features/comment-shortcuts.tsx
fregante: Can be removed

source/features/comment-shortcuts.tsx
fregante: I specifically said that I want the files to be selected as well. O and V will be additional shortcuts to toggle and mark files as viewed. That part can come later, but even just navigation is useful.

source/features/keyboard-navigation.tsx
```diff
@@ -0,0 +1,67 @@
...
+		return;
+	}
+
+	const focusedComment = select(':target')!;
+
+	if (['j', 'k'].includes(event.key)) {
+		event.preventDefault();
+
+		const items = select.all('.js-targetable-element')
+			.filter(comment => !comment.querySelector('.minimized-comment:not(.d-none)'));
```
yakov116, line 23: This filter does not work correctly it needs to check also for `'.js-minimizable-comment-group'`.
Can I do the following?
 
```suggestion
		const itemsSelectors = pageDetect.isPRFiles ? '.js-targetable-element' : '.js-minimizable-comment-group'
		const items = select.all(itemsSelectors)
			.filter(comment => !comment.querySelector('.minimized-comment:not(.d-none)'));
```

fregante: If by **"does not work correctly"** you mean **"it highlights reviews but they don't have a highlight outline"**, then the solution could be:

`.js-targetable-element:not([id^="pullrequestreview"])`

source/features/keyboard-navigation.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+void features.add({
+	id: __filebasename,
+	description: 'Adds shortcuts to comments: `j` focuses the comment below; `k` focuses the comment above; `e` edits the focused comment; `d` deletes the focused comment.',
+	screenshot: false,
+	shortcuts: {
+		j: 'Focus the comment below',
+		k: 'Focus the comment above',
+		e: 'Edit the focused comment',
+		d: 'Delete the focused comment'
```
fregante, line 59:
```suggestion
		j: 'Focus the comment/file below',
		k: 'Focus the comment/file above'
```

source/features/keyboard-navigation.tsx
```diff
@@ -0,0 +1,67 @@
...
+	}
+}
+
+function init(): void {
+	document.addEventListener('keypress', runShortcuts);
+}
+
+void features.add({
+	id: __filebasename,
+	description: 'Adds shortcuts to comments: `j` focuses the comment below; `k` focuses the comment above; `e` edits the focused comment; `d` deletes the focused comment.',
```
fregante, line 53:
```suggestion
	description: 'Adds shortcuts to conversations and PR file lists: `j` focuses the comment/file below; `k` focuses the comment/file above.',
```

source/features/keyboard-navigation.tsx
```diff
@@ -0,0 +1,67 @@
...
+		// Focus comment without pushing to history
+		location.replace('#' + items[chosenCommentIndex].id);
+		return;
+	}
+
+	const actionClass = shortcutClass.get(event.key);
+	if (actionClass) {
+		event.preventDefault();
+		select(actionClass, focusedComment)?.click();
+	}
```
fregante, line 44:
```suggestion
	}
```

source/features/keyboard-navigation.tsx
```diff
@@ -0,0 +1,67 @@
...
+import * as pageDetect from 'github-url-detection';
+
+import features from '.';
+import {isEditable} from '../helpers/dom-utils';
+
+const shortcutClass = new Map<string, string>([
+	['e', '.unminimized-comment [aria-label="Edit comment"]:not([hidden])'],
+	['d', '[aria-label="Delete comment"]']
+]);
+
```
fregante, line 11:
```suggestion
```

readme.md
```diff
@@ -180,6 +180,7 @@ Thanks for contributing! 🦋🙌
 - [](# "minimize-upload-bar") [Reduces the upload bar to a small button.](https://user-images.githubusercontent.com/55841/59802383-3d994180-92e9-11e9-835d-60de67611c30.png)
 - [](# "clean-rich-text-editor") [Hides unnecessary comment field tooltips and toolbar items](https://user-images.githubusercontent.com/1402241/53629083-a4fe8900-3c47-11e9-8211-bfe2d254ffcb.png) (each one has a keyboard shortcut.)
 - [](# "monospace-textareas") Use a monospace font for all textareas.
+- [](# "keyboard-navigation") Adds shortcuts to comments: <kbd>j</kbd> focuses the comment below; <kbd>k</kbd> focuses the comment above; <kbd>e</kbd> edits the focused comment; <kbd>d</kbd> deletes the focused comment.
```
fregante, line 183:
```suggestion
- [](# "keyboard-navigation") Adds shortcuts to comments and PR files: <kbd>j</kbd> focuses the comment below..
```

fregante, line 183:
```suggestion
- [](# "keyboard-navigation") Adds shortcuts to conversations and PR file lists: <kbd>j</kbd> focuses the comment/file below; <kbd>k</kbd> focuses the comment/file above.
```

source/features/keyboard-navigation.tsx
yakov116: No mean that minimized comments are highlighted

yakov116:
Applied will test later tonight

yakov116:
Tested. Its good with me except for https://github.com/sindresorhus/refined-github/pull/3242#discussion_r448540801

source/features/keyboard-navigation.tsx
```diff
@@ -0,0 +1,53 @@
...
+	const focusedComment = select(':target')!;
+
+	if (['j', 'k'].includes(event.key)) {
+		event.preventDefault();
+
+		const items = select.all('.js-targetable-element')
+			.filter(comment => !comment.querySelector('.minimized-comment:not(.d-none)'));
+		// `j` goes to the next comment, `k` goes back a comment
+		const direction = event.key === 'j' ? 1 : -1;
```
notlmn, line 20: This should be the opposite. `j` should be previous and `k` should be next.

```suggestion
		const direction = event.key === 'j' ? -1 : 1;
```


Also this looks very slow, especially on large PRs (not ou  r fault though, its GH).

And pressing the same key repeatedly (on Firefox) fails to take us to the right comment (GHs `window.scrollTo()` call for some reason gets delayed and adds up eventually scrolling to random positions).

source/features/keyboard-navigation.tsx
```diff
@@ -0,0 +1,53 @@
...
+		const currentIndex = items.indexOf(focusedComment);
+
+		// Start at 0 if nothing is; clamp index
+		const chosenCommentIndex = Math.min(
+			Math.max(0, currentIndex + direction),
+			items.length - 1
+		);
+
+		// Focus comment without pushing to history
+		location.replace('#' + items[chosenCommentIndex].id);
```
notlmn, line 31: I'd avoid doing this if `location.hash` is already the same as comment ID. I'm seeing some weird behavior as mentioned in https://github.com/sindresorhus/refined-github/issues/3263#issuecomment-647483078 and https://github.com/sindresorhus/refined-github/issues/3263#issuecomment-647498245.

source/features/keyboard-navigation.tsx
yakov116: I am following GitHub

image

source/features/keyboard-navigation.tsx
fregante: > if `location.hash` is already the same as comment ID

That shouldn't happen... unless there's a single file or comment on the page.

source/features/keyboard-navigation.tsx
fregante: > adds up eventually scrolling to random positions

This is probably GitHub screwing it up: https://github.com/sindresorhus/refined-github/issues/3263

source/features/keyboard-navigation.tsx
fregante: That's fine. Minimized comments are still comments

source/features/keyboard-navigation.tsx
notlmn: If you are already on the first comment and press <kbd>k</kbd> again, this happens. The same goes for pressing <kbd>j</kbd> when already on the last comment.

fregante: Indeed. I was thinking that this cycles the list. 

Let’s add a `if chosenCommentIndex !== currentIndex`

yakov116:
FTR this is what I am talking about. That it randomly opens minimized reviews when using only the `.js-targetable-element` selector.
Video_2020-07-03_094716

But lets deal with this issue another time.

source/features/keyboard-navigation.tsx
```diff
@@ -0,0 +1,55 @@
...
+	screenshot: false,
+	shortcuts: {
+		j: 'Focus the comment/file below',
+		k: 'Focus the comment/file above'
+	}
+}, {
+	include: [
+		pageDetect.hasComments
+	],
+	repeatOnAjax: false,
```
fregante, line 53:
```suggestion
```

This can be dropped. Going from `isConversationList` to `isPR` sometimes happens via ajax, so `hasComments` won't be re-tested, so the feature won't work.

Since `init` is just `addEventListener`, it's not a big deal

readme.md
```diff
@@ -179,6 +179,7 @@ Thanks for contributing! 🦋🙌
 - [](# "minimize-upload-bar") [Reduces the upload bar to a small button.](https://user-images.githubusercontent.com/55841/59802383-3d994180-92e9-11e9-835d-60de67611c30.png)
 - [](# "clean-rich-text-editor") [Hides unnecessary comment field tooltips and toolbar items](https://user-images.githubusercontent.com/1402241/53629083-a4fe8900-3c47-11e9-8211-bfe2d254ffcb.png) (each one has a keyboard shortcut.)
 - [](# "monospace-textareas") Use a monospace font for all textareas.
+- [](# "keyboard-navigation") Adds shortcuts to conversations and PR file lists: <kbd>j</kbd> focuses the comment/file below; <kbd>k</kbd> focuses the comment/file above.
```
fregante, line 182: Gif missing here

source/features/keyboard-navigation.tsx
fregante: > `.js-targetable-element:not([id^="pullrequestreview"])`

Let's also use this

readme.md
fregante: aasad

```suggestion
- [](# "keyboard-navigation") [Adds shortcuts to conversations and PR file lists: <kbd>j</kbd> focuses the comment/file below; <kbd>k</kbd> focuses the comment/file above.](https://user-images.githubusercontent.com/1402241/86573176-48665900-bf74-11ea-8996-a5c46cb7bdfd.gif)
```

source/features/keyboard-navigation.tsx
```diff
@@ -0,0 +1,54 @@
...
+}
+
+function init(): void {
+	document.addEventListener('keypress', runShortcuts);
+}
+
+void features.add({
+	id: __filebasename,
+	description: 'Adds shortcuts to conversations and PR file lists: `j` focuses the comment/file below; `k` focuses the comment/file above.',
+	screenshot: false,
```
fregante, line 44:
```suggestion
	screenshot: 'https://user-images.githubusercontent.com/1402241/86573176-48665900-bf74-11ea-8996-a5c46cb7bdfd.gif',
```

fregante:
You made it! I opened a new issue for the remaining concerns: https://github.com/sindresorhus/refined-github/issues/3319

PR #3221, opened 2020-06-12
Fix button styles in "Repository refresh" beta

FloEdelmann:
Design fix for <https://github.com/sindresorhus/refined-github/issues/3081#issuecomment-643198654> / <https://github.com/sindresorhus/refined-github/pull/3205#issuecomment-641926371>.

### After

Screenshot_2020-06-12_14-23-11
Screenshot_2020-06-12_14-22-42

Note that I didn't change the button size for `latest-tag-button` and `list-prs-for-file`; I just centered them, as that looks better IMO.

<details>
<summary>See how it looks with changed button sizes instead:</summary>

Screenshot_2020-06-12_14-15-01
Screenshot_2020-06-12_14-15-34

</details>

PR #3214, opened 2020-06-10
Add “Compare to HEAD” link next to `latest-tag-button`

yakov116:
Follows #3199
Closes #2861

Waiting for
>Also we need to wait for https://github.com/sindresorhus/refined-github/pull/3205#issuecomment-641926371
>_Originally posted by @fregante in https://github.com/sindresorhus/refined-github/pull/3199_

PR #3243, opened 2020-06-17
Fix `default-branch-button` in "old design"

FloEdelmann:
Follows #3221. Sorry for not testing with "Repository refresh" beta disabled :/

Thanks @yakov116 for catching this in https://github.com/sindresorhus/refined-github/pull/3214#issuecomment-645008022

### DIFF
--- a/source/features/default-branch-button.tsx
+++ b/source/features/default-branch-button.tsx
@@ -1,5 +1,4 @@
 import React from 'dom-chef';
-import select from 'select-dom';
 import elementReady from 'element-ready';
 import * as pageDetect from 'github-url-detection';
 import ChevronLeftIcon from 'octicon/chevron-left.svg';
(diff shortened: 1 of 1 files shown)


source/features/default-branch-button.tsx
```diff
@@ -43,7 +44,7 @@ async function init(): Promise<false | void> {
 		</a>
 	);
 
-	if (branchSelector.classList.contains('btn-sm')) {
+	if (select.exists('[data-hotkey="w"].btn-sm', branchSelector)) {
```
yakov116, line 47: Does this have any negative effect since domReady is set false?

FloEdelmann:
Thanks again @yakov116, that may have caused a race condition. I've updated the code.

Also, I've fixed the dropdown position:

### Before

Screenshot_2020-06-17_18-28-20

### After

"Repository refresh" beta enabled:  
Screenshot_2020-06-17_18-27-56

"Repository refresh" beta disabled:  
Screenshot_2020-06-17_18-28-45

fregante:
Thanks @FloEdelmann and @yakov116 

Please include the 3 pieces of information asked in the PR template when opening PRs

PR #3250, opened 2020-06-18
Add `repo-contained-header` feature

adamlacombe:
This feature only applies to those that have enabled the repository refresh feature preview.

1. LINKED ISSUES:
	- fixes https://github.com/sindresorhus/refined-github/issues/3280
   > Great until you’re on a big screen and then there’s just weird black spaces and loads of mouse travel to the left hand nav bar and back to the central content. -[@kurucu83](https://twitter.com/kurucu83/status/1273476984765480961)

2. TEST URLS:
   https://github.com/sindresorhus/Gifski

3. SCREENSHOT:

- Before
before

- After
after


fregante:
I feel this too, but let's way for this new style to go live first before trying to fix it. Also this is likely targetable with CSS alone, without any JS.

Let's review this later

source/features/repo-contained-header.css
```diff
@@ -0,0 +1,25 @@
+.pagehead .UnderlineNav-item:focus,
+.pagehead .UnderlineNav-item:hover {
+	border-bottom-color: var(--repo-nav-border-bottom-color-hover, #2ea44f80) !important;
+}
```
kidonng, line 4: I don't think this matches GitHub's style very much.

adamlacombe: I agree, I was looking for something with more contrast against the light gray border. How about https://primer.style/css/support/variables#$border-gray-darker? I would be okay with removing this declaration if need be.

kidonng: > I would be okay with removing this declaration if need be.

I'm fine with the `$border-gray-darker`, but let's see how maintainers think about this.

fregante:
How come all that code is required? Is it possible to contain the nav without affecting the border? Ideally the CSS should just be one rule with “max with” and “margin auto”

adamlacombe:
I was able to reduce it down to two declarations. I had to add `box-shadow: inset 0 -1px 0 #e1e4e8;` to `.pagehead` so the gray rule wasn't lost.

fregante:
Lets wait for the discussion to resolve in https://github.com/sindresorhus/refined-github/issues/3280

PR #3251, opened 2020-06-18
Don't run `comments-time-machine-links` if comment is after latest commit

### LINKED ISSUE #3244: `comments-time-machine-links` shouldn't show banner when latest commit is before comment
Clicking a non-perma link to a file from `comments-time-machine-links` adds a query parameter with the comment's date. Then adds a banner at the top of that file allowing the user to go to that file whenever the comment was posted.

If the comment was posted after the latest commit for that file then the banner shouldn't be shown.

For example: https://github.com/sindresorhus/refined-github/blob/master/source/features/comments-time-machine-links.tsx

The latest commit timestamp can be found on the dom element: `.repository-content .Box.Box--condensed relative-time`.

yakov116:
1. LINKED ISSUES:
   Closes #3244

2. TEST URLS:
Should not show
	https://github.com/sindresorhus/refined-github/blob/master/source/features/comments-time-machine-links.tsx

Should show
```
https://github.com/sindresorhus/refined-github/blob/master/source/features/comments-time-machine-links.tsx?rgh-link-date=2019-12-11T10%3A12%3A11Z
```

3. SCREENSHOT:

### DIFF
--- a/source/features/comments-time-machine-links.tsx
+++ b/source/features/comments-time-machine-links.tsx
@@ -71,8 +71,10 @@ async function showTimemachineBar(): Promise<void | false> {
 			return false;
 		}
 
-		// "Repository refresh" layout uses the itemprop="dateModified"
-		const lastCommitDate = await elementReady('.repository-content .Box.Box--condensed relative-time, [itemprop="dateModified"] relative-time');
+		const lastCommitDate = await elementReady([
+			'.repository-content .Box.Box--condensed relative-time',
+			'[itemprop="dateModified"] relative-time' // "Repository refresh" beta
+		].join());
 		if (date > lastCommitDate?.attributes.datetime.value!) {
 			return false;
 		}
--- a/source/features/comments-time-machine-links.tsx
+++ b/source/features/comments-time-machine-links.tsx
@@ -71,6 +71,11 @@ async function showTimemachineBar(): Promise<void | false> {
 			return false;
 		}
 
+		const lastCommitDate = await elementReady('.repository-content .Box.Box--condensed relative-time');
+		if (date > lastCommitDate?.attributes.datetime.value!) {
+			return false;
+		}
+
 		const parsedUrl = new GitHubURL(location.href);
 		parsedUrl.branch = `${parsedUrl.branch}@{${date}}`;
 		url.pathname = parsedUrl.pathname;
--- a/source/features/comments-time-machine-links.tsx
+++ b/source/features/comments-time-machine-links.tsx
@@ -65,13 +65,14 @@ async function showTimemachineBar(): Promise<void | false> {
 		url.pathname = pathnameParts.join('/');
 	} else {
 		// This feature only makes sense if the URL points to a non-permalink
-		const branchSelector = await elementReady('.branch-select-menu i');
+		const branchSelector = await elementReady('[data-hotkey="w"] i');
 		const isPermalink = /Tag|Tree/.test(branchSelector!.textContent!);
 		if (isPermalink) {
 			return false;
 		}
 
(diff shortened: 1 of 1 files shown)


source/features/comments-time-machine-links.tsx
```diff
@@ -71,6 +71,11 @@ async function showTimemachineBar(): Promise<void | false> {
 			return false;
 		}
 
+		const lastCommitDate = await elementReady('.repository-content .Box.Box--condensed relative-time');
```
fregante, line 74: Not compatible with the repository refresh. You can probably **add** `[itemprop="dateModified"] relative-time`. Add a JS comment exactly like in the other "Repository refresh" commits.

yakov116: Hope you like the way I did the comment. It's a bit different here since it's just another selector.

source/features/comments-time-machine-links.tsx
```diff
@@ -65,12 +65,18 @@ async function showTimemachineBar(): Promise<void | false> {
...
 		// This feature only makes sense if the URL points to a non-permalink
-		const branchSelector = await elementReady('.branch-select-menu i');
+		const branchSelector = await elementReady('[data-hotkey="w"] i');
 		const isPermalink = /Tag|Tree/.test(branchSelector!.textContent!);
 		if (isPermalink) {
 			return false;
 		}
 
+		// "Repository refresh" layout uses the itemprop="dateModified"
+		const lastCommitDate = await elementReady('.repository-content .Box.Box--condensed relative-time, [itemprop="dateModified"] relative-time');
```
FloEdelmann, line 75: Should we do it like this to keep the comment near the actual code?

```suggestion
		const lastCommitDate = await elementReady([
			'.repository-content .Box.Box--condensed relative-time',
			'[itemprop="dateModified"] relative-time' // "Repository refresh" beta
		].join(','));
```

source/features/comments-time-machine-links.tsx
```diff
@@ -65,12 +65,20 @@ async function showTimemachineBar(): Promise<void | false> {
...
-		const branchSelector = await elementReady('.branch-select-menu i');
+		const branchSelector = await elementReady('[data-hotkey="w"] i');
 		const isPermalink = /Tag|Tree/.test(branchSelector!.textContent!);
 		if (isPermalink) {
 			return false;
 		}
 
+		const lastCommitDate = await elementReady([
+			'.repository-content .Box.Box--condensed relative-time',
+			'[itemprop="dateModified"] relative-time' // "Repository refresh" beta
```
FloEdelmann, line 76: Sorry, we use this everywhere else, too :see_no_evil: 

```suggestion
			'[itemprop="dateModified"] relative-time' // "Repository refresh" layout
```

source/features/comments-time-machine-links.tsx
```diff
@@ -65,12 +65,20 @@ async function showTimemachineBar(): Promise<void | false> {
 		url.pathname = pathnameParts.join('/');
 	} else {
 		// This feature only makes sense if the URL points to a non-permalink
-		const branchSelector = await elementReady('.branch-select-menu i');
+		const branchSelector = await elementReady('[data-hotkey="w"] i');
```
fregante, line 68: Since you're editing this, can you check whether we use this exact selector everywhere in RGH?

yakov116: We do, I checked. (Should it be its own global function?)

fregante: > Should it be its own global function

This is one of those situations that are _too specific_ yet common. Creating a function for it won't save us once we need a _slightly different_ selector and definitely won't save us from selectors in CSS files. We just need to learn to search and replace every time a selector is changed anywhere.

Issue #2729, opened 2020-01-22
Restore `indented-code-wrapping` feature


fregante:
Originally implemented in https://github.com/sindresorhus/refined-github/pull/1989, the feature is currently disabled due to https://github.com/sindresorhus/refined-github/issues/2421

If you're interested in this feature, either rewrite it or restore the file deleted in https://github.com/sindresorhus/refined-github/pull/3253 and send a PR to fix this issue.

---

It also has a minor bug to look into after it's restored: https://github.com/sindresorhus/refined-github/issues/2205

fregante:
January 2020 was a different era. Closing

PR #3253, opened 2020-06-18
Remove broken `indented-code-wrapping`

fregante:
`indented-code-wrapping` has been disabled for 9 months and probably needs a rewrite/rethink.

No need to keep it around in the readme while we wait.

https://github.com/sindresorhus/refined-github/issues/2729 will stay open.


fregante:
I hope to see it back. It was useful

PR #3195, opened 2020-06-06
Restore `download-folder-button` in "Repository refresh" beta

FloEdelmann:
References #3081.

Screenshot_2020-06-06_21-15-56

PR #3254, opened 2020-06-18
Fix `download-folder-button` and `latest-tag-button`

FloEdelmann:
1. Fixes <https://github.com/sindresorhus/refined-github/pull/3195#issuecomment-646323153>.

   `download-folder-button` doesn't need to be shown when "Clone or download" button exists.

   Problem: `isRepoRoot` seems to not work on branches with slashes. It caused the big button to appear with "Repository refresh" beta disabled and a console error with "Repository refresh" beta enabled.

   Test: https://github.com/yakov116/TestR/tree/this/branch/has/many/slashes

2. Fixes <https://github.com/sindresorhus/refined-github/issues/3081#issuecomment-643198654> again after #3214 was merged.

   Test: https://github.com/OpenLightingProject/open-fixture-library

   Before:  
   Screenshot_2020-06-19_00-05-34

   After:  
   Screenshot_2020-06-19_00-07-56

### DIFF
--- a/source/features/download-folder-button.tsx
+++ b/source/features/download-folder-button.tsx
@@ -42,7 +42,7 @@ void features.add({
 		pageDetect.isRepoTree
 	],
 	exclude: [
-		() => select.exists('get-repo-controller') // Already has an native download ZIP button
+		pageDetect.isRepoRoot // Already has an native download ZIP button
 	],
 	init
 });
--- a/source/features/download-folder-button.tsx
+++ b/source/features/download-folder-button.tsx
@@ -6,12 +6,15 @@ import DownloadIcon from 'octicon/download.svg';
 import features from '.';
 
 function init(): void {
+	const downloadUrl = new URL('https://download-directory.github.io/')
+	downloadUrl.searchParams.set('url', location.href);
+
 	const folderButtonGroup = select('.file-navigation .BtnGroup.float-right');
 	if (folderButtonGroup) {
 		folderButtonGroup.prepend(
 			<a
 				className="btn btn-sm BtnGroup-item"
-				href={`https://download-directory.github.io/?url=${location.href}`}
+				href={downloadUrl.href}
 			>
 				Download
 			</a>
@@ -21,7 +24,7 @@ function init(): void {
 		select('.file-navigation > .d-flex')!.append(
 			<a
 				className="btn ml-2"
-				href={`https://download-directory.github.io/?url=${location.href}`}
+				href={downloadUrl.href}
 			>
 				<DownloadIcon className="mr-1"/>
 				Download
--- a/source/features/download-folder-button.tsx
+++ b/source/features/download-folder-button.tsx
@@ -39,7 +39,7 @@ void features.add({
 		pageDetect.isRepoTree
 	],
 	exclude: [
-		pageDetect.isRepoRoot // Already has an native download ZIP button
+		() => select.exists('get-repo-controller') // Already has an native download ZIP button
 	],
 	init
 });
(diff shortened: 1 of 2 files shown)


yakov116:
@FloEdelmann maybe throw https://github.com/sindresorhus/refined-github/pull/3195#discussion_r438047898 in?

source/features/download-folder-button.tsx
```diff
@@ -39,7 +39,7 @@ void features.add({
 		pageDetect.isRepoTree
 	],
 	exclude: [
-		pageDetect.isRepoRoot // Already has an native download ZIP button
+		() => select.exists('get-repo-controller') // Already has an native download ZIP button
```
fregante, line 42: Might be worth keeping them both
```suggestion
		pageDetect.isRepoRoot, // Already has an native download ZIP button
		() => select.exists('get-repo-controller') // Same as `isRepoRoot` but supports branches with slashes
```

Better yet:
- **add** the `exists()` to `isRepoRoot` in `github-url-detection` ("if no URL is supplied, also check the dom")
- look for a meta tag in the `head` instead; if it exists, this is generally safer (because it can be run even on `waitForDom: false`)

yakov116: 🎉 `<meta name="selected-link" value="repo_source" data-pjax-transient>`

yakov116: @fregante $('meta[name="selected-link"]') is an awesome selector! It works on almost all pages I tested and it gives back what page your on!

yakov116: @fregante $('meta[name="selected-link"]') is an awesome selector! It works on almost all pages I tested and it gives back what page your on!

fregante: You mean `[name="analytics-location"]`?

yakov116: We may have to use both

fregante: Neither one seems to apply to this case though.

This is true on any page of the Code tab. It might be worth adding as a new `isRepoSource` or `isRepoCodeTab` which is probably useful in many cases.
```js
$('meta[name="selected-link"]').getAttribute('value') === 'repo_source'
```

FloEdelmann: I've opened a PR to improve `isRepoRoot`: https://github.com/fregante/github-url-detection/pull/16

source/features/download-folder-button.tsx
```diff
@@ -39,7 +42,7 @@ void features.add({
...
```
busches, line 45: Can we disable this for enterprise too, while we're making changes?

fregante: Enterprise support should come to `download-directory` soon, there’s a PR open

busches: 😮

FloEdelmann: For reference, this is https://github.com/download-directory/download-directory.github.io/pull/29

busches: Interesting, I have yet to see an enterprise install that would be publicly accessible, even with a token. Normally they're behind a VPN or a third party login, e.g. Okta.

fregante: @busches indeed, but I don't think that's a hard requirement.

fregante, line 45: I think this specific change can be dropped if you use the [v2](https://github.com/fregante/github-url-detection/releases/tag/v2.0.0) thanks to https://github.com/fregante/github-url-detection/pull/16

FloEdelmann: I've reverted the commit; the package will be updated in #3308.

Issue #3232, opened 2020-06-14
Poor performance on collapsed lists


dakshshah96:
The transitions are very sluggish while hovering over PRs. Would love to see them become more smooth and help if required.

![Kapture 2020-06-14 at 22 09 52](https://user-images.githubusercontent.com/7896438/84598939-f9a13400-ae8b-11ea-8a23-5e8cd37a7b8d.gif)

Also, sorry if this is not the right place to ask this, but is there a way to disable this collapse behaviour? I couldn't find the name of this feature in the list of features.

FloEdelmann:
The feature is called `dim-bots`:

https://github.com/sindresorhus/refined-github/blob/master/source/features/dim-bots.tsx
https://github.com/sindresorhus/refined-github/blob/master/source/features/dim-bots.css

fregante:
Confirmed. This was broken by the last “fix” we applied to that feature. PR welcome to revert that and/or fix it again, but preserving the way it was working before.

PR #3128, opened 2020-05-23
Avoid clipping descriptions and CI status messages in `dim-bots`

tanettrimas:
Fixes #3124

Solution:
- The overflow-hidden property is moved to not:navigation-focus
- Implemented a scroller for text that exceeds the container, which was not the case before.

Skjermbilde 2020-05-23 kl  19 39 57
Skjermbilde 2020-05-23 kl  19 39 21

**TEST URL**
https://github.com/kulshekhar/ts-jest/commits/master

PR #3257, opened 2020-06-19
Improve `dim-bots`’s behavior

fregante:
Fixes https://github.com/sindresorhus/refined-github/issues/3232

Follows https://github.com/sindresorhus/refined-github/pull/3128 
Follows https://github.com/sindresorhus/refined-github/pull/2864

Instead of clipping via `height` and `overflow: hidden`, I'm just using negative margins to collapse the meta information.

- No more visible overflow
- No more clipped titles (because there's no fixed height)

Also some selectors needed to be updated because lists no longer use `display: table` :tada:

# Test

## PR List

https://github.com/facebook/react/pulls?q=bump

## Commit list

https://github.com/kulshekhar/ts-jest/commits/master?after=3247ae409af4d9e47acf1ac31707b970f829e990+104

11s

### DIFF
--- a/source/features/pr-branches.tsx
+++ b/source/features/pr-branches.tsx
@@ -97,7 +97,7 @@ function createLink(reference: RepositoryReference): HTMLSpanElement {
 async function init(): Promise<false | void> {
 	const prLinks = select.all('.js-issue-row .js-navigation-open[data-hovercard-type="pull_request"]')
 		// Exclude bots
-		.filter(link => !link.parentElement?.querySelector(botSelectors.join(',')));
+		.filter(link => !link.parentElement?.querySelector(botSelectors.join()));
 	if (prLinks.length === 0) {
 		return false;
 	}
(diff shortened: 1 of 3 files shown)


source/features/pr-branches.tsx
```diff
@@ -94,7 +95,9 @@ function createLink(reference: RepositoryReference): HTMLSpanElement {
 }
 
 async function init(): Promise<false | void> {
-	const prLinks = select.all('.js-issue-row .js-navigation-open[data-hovercard-type="pull_request"]');
+	const prLinks = select.all('.js-issue-row .js-navigation-open[data-hovercard-type="pull_request"]')
+		// Exclude bots
+		.filter(link => !link.parentElement?.querySelector(botSelectors.join(',')));
```
yakov116, line 100:
```suggestion
		.filter(link => !link.parentElement?.querySelector(botSelectors.join()));
```
Isn't `.join` by default `','`?

Issue #3239, opened 2020-06-16
default-branch-button doesn't actually use default branch


DianaNites:
The default branch button normally only appears when you're not currently on the default branch, but github allows repositories to change the default branch, which causes the button to *always* appear.

Button appears, despite being on the (different) default branch

![Screenshot_20200615_214643](https://user-images.githubusercontent.com/5275194/84722622-b7dac100-af51-11ea-805f-f620f0e43d78.png)

No button when on the default branch

![Screenshot_20200615_214245](https://user-images.githubusercontent.com/5275194/84722416-29fed600-af51-11ea-86ab-6c77cfa8c11d.png)

Button when not on the default branch

![Screenshot_20200615_214604](https://user-images.githubusercontent.com/5275194/84722635-c6c17380-af51-11ea-803e-7124e01dee14.png)

yakov116:
We dont check every time you visit a repo if the default branch changed. Since changing a default branch is not a common thing, we check once every 10 days.

Either you can wait for the 10 days or you can clear the cache from the extension options.

yakov116:
@fregante I think we should add a detection if the branch is “main” and the default is “master” then re validate. Due to many people changing the master branch to main.

DianaNites:
Oh, there is a check? Is that documented anywhere?

Clearing the cache did correct the issue, so not a bug I guess.

fregante:
> I think we should add a detection if the branch is “main” and the default is “master” then re validate.

That means "always fetch unless you're on the default branch"

Changing default branch is not a common action so we shouldn't slow down the other hundred thousand "get default branch" requests for that.

What we _could_ do is speed up this getter since it's one of the most common information we use.

For example, when you're on `isRepoRoot`, you're _definitely_ on the default branch. This means:

- we can use the `current branch` value, already on the page
- we can skip/refresh the cache https://github.com/fregante/webext-storage-cache/issues/18

If anyone wants to add the first part: https://github.com/sindresorhus/refined-github/commit/1de71bb0a7b1b60d6dd123dd6699f346029399f8

yakov116:
Another detection can be when your on the branches page, the default branch is there.

yakov116:
@fregante wasn't this done?

fregante:
> * we can use the `current branch` value, already on the page
> 
> * we can skip/refresh the cache [fregante/webext-storage-cache#18](https://github.com/fregante/webext-storage-cache/issues/18)

[...]

PR #3261, opened 2020-06-21
Make `getDefaultBranch` quicker on `isRepoHome`

yakov116:
Follows https://github.com/sindresorhus/refined-github/issues/3239#issuecomment-646072282

**Note:** `isRepoTree`cannot be used instead of `!String(getRepoPath()).startsWith('tree/')` since `isRepoTree` includes `isRepoRoot`

### DIFF
--- a/source/github-helpers/get-default-branch.ts
+++ b/source/github-helpers/get-default-branch.ts
@@ -1,9 +1,9 @@
 import cache from 'webext-storage-cache';
 import select from 'select-dom';
-import {isForkedRepo} from 'github-url-detection/esm';
+import * as pageDetect from 'github-url-detection';
 
 import * as api from './api';
-import {RepositoryInfo, getRepositoryInfo, getRepoURL} from '.';
+import {RepositoryInfo, getRepositoryInfo, getRepoURL, getCurrentBranch, getRepoPath} from '.';
 
 // This regex should match all of these combinations:
 // "This branch is even with master."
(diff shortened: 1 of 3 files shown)


source/github-helpers/get-default-branch.ts
```diff
@@ -14,7 +14,11 @@ const branchInfoRegex = /([^ ]+)\.$/;
 
 export default cache.function(async (repository: Partial<RepositoryInfo> = getRepositoryInfo()): Promise<string> => {
 	if (JSON.stringify(repository) === JSON.stringify(getRepositoryInfo())) {
-		if (!isForkedRepo()) {
+		if (pageDetect.isRepoRoot() && !String(getRepoPath()).startsWith('tree/')) {
```
fregante, line 17: Isn't this equivalent?
```suggestion
		if (pageDetect.isRepoRoot() && !pageDetect.isRepoTree()) {
```

yakov116: >Note: isRepoTreecannot be used instead of !String(getRepoPath()).startsWith('tree/') since isRepoTree includes isRepoRoot

fregante: > No its not. I left a note in the first post

I will have to read your first post more carefully.

fregante: Can you leave those comments inline instead of in the first post? So they're linked exactly to the comment

fregante: Example: https://github.com/sindresorhus/refined-github/pull/3206#pullrequestreview-427166121

yakov116: Sure

fregante, line 17: If you want the absolute repo root, then just look for it. This should be it:
```suggestion
		if (getRepoPath() === '') {
```

fregante: I'd call that `repo home`

yakov116: Will send a PR now

PR #3269, opened 2020-06-23
Fix `Add-tag-to-commits` and `mark-merge-commits-in-list`

### LINKED ISSUE #3294: Marks merge commits in commit lists not working
Since the GitHub redesign, the _Marks merge commits in commit lists_ is not working anymore. 😭 
example from README

yakov116:
Fixes #3294

This fixes both features that relied on the same dataset and the dom changed.

Test On:
https://github.com/sindresorhus/refined-github/commits/master

### DIFF
--- a/source/features/mark-merge-commits-in-list.tsx
+++ b/source/features/mark-merge-commits-in-list.tsx
@@ -36,7 +36,7 @@ const filterMergeCommits = async (commits: string[]): Promise<string[]> => {
 // eslint-disable-next-line import/prefer-default-export
 export function getCommitHash(commit: HTMLElement): string {
 	// Pre "Repository refresh" layout
-	return 	commit.dataset.channel!.split(':')[3] ||
+	return 	commit.dataset.channel!.split(':')[3] ??
 	commit.querySelector<HTMLAnchorElement>('a[href]')!.href.split('/').pop()!;
 }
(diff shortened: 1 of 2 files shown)


source/features/mark-merge-commits-in-list.tsx
```diff
@@ -33,8 +33,11 @@ const filterMergeCommits = async (commits: string[]): Promise<string[]> => {
 	return mergeCommits;
 };
 
-function getCommitHash(commit: HTMLElement): string {
-	return commit.dataset.channel!.split(':')[3];
+// eslint-disable-next-line import/prefer-default-export
+export function getCommitHash(commit: HTMLElement): string {
+	// Pre "Repository refresh" layout
+	return 	commit.dataset.channel!.split(':')[3] ||
```
fregante, line 39: I think this should be `??`

yakov116: Channel still exists

yakov116: Your right. It returns undefined - I responded half in my sleep.
```suggestion
	return 	commit.dataset.channel!.split(':')[3] ??
```

fregante:
Thanks Yakow ^_^

PR #3270, opened 2020-06-24
Drop outdated style

kidonng:
- Drop `widen-search-field`
 
  Now the search box is full-width by default, we may be able to drop some outdated `widen-search-field` style.

  ~~Also adjusted the icon position.~~

  <details><summary><s>Icon position adjustment</s></summary>

  **Before:**
  
  **After:**
  image
  </details>
- Drop `improve-shortcut-help` style
- Drop a declartion in `toggle-files-button`


fregante:
Dropping code is always welcome 🎉

busches:
This is a minor change, but it'll likely be a while before GHE is updated, do we want to move these styles to GHE only or drop them all?

source/features/improve-shortcut-help.css
```diff
@@ -2,21 +2,3 @@
 	width: 908px;
 	cursor: default;
 }
-
-.rgh-inactive-shortcut-group {
```
kidonng, line 6: These classes only appear here, so I assume they can be safely removed.

source/features/toggle-files-button.css
```diff
@@ -4,15 +4,6 @@
 	padding: 10px 13px;
 }
 
-.rgh-files-hidden .commit-tease {
```
kidonng, line 7: `.commit-tease` seems to disappear in the new design.

This file is also touched by some other open PR, but I can't find it.

source/features/widen-search-field.css
```diff
@@ -13,26 +13,6 @@ https://github.com/sindresorhus/refined-github/milestones
...
-.subnav-search-input {
-	flex-grow: 1;
-	display: flex;
-	align-items: flex-start;
-}
-
-/* Clear .subnav-search input field when using display: flex */
-form.subnav-search > div.card-filter-autocomplete-dropdown {
-	margin-top: 40px !important;
-}
```
kidonng, line 29: The above two declarations are no longer needed since they are already full width of the container.

source/features/widen-search-field.css
```diff
@@ -44,11 +24,6 @@ form.subnav-search > div.card-filter-autocomplete-dropdown {
 	float: left;
 }
 
-/* Reset width of search field in global lists */
-.page-content .subnav-search-input-wide {
-	width: auto;
-}
-
```
kidonng, line 51: The above two removed declarations are already presented in GitHub's own style.

kidonng:
There are still 5 declarations left in `widen-search-field`:

1 & 2: should move to `tags-dropdown` probably
3: Hide "Filters" text
4: Center the icon
5: Hide label and milestone icons

source/features/improve-shortcut-help.css
```diff
@@ -1,22 +0,0 @@
-:root .facebox .shortcuts {
```
kidonng, line 1: `.facebox` can't be found in GitHub's code now.

PR #3247, opened 2020-06-18
Match border radius in GitHub new design

kidonng:
Related to #3081
Fixes #3248

Match border radius with GitHub's new design.

Features and test URLs:
- `reactions-avatars`: here

  <details>

  "Design updates" feature preview **disabled**:
  image
  "Design updates" feature preview **enabled**:
  image

  </details>
- `parse-backticks`: https://github.com/sindresorhus/refined-github/pulls
- `clean-dashboard`: https://github.com/
- `clean-pinned-issues`: https://github.com/sindresorhus/refined-github/issues
- `hide-watch-and-fork-count`: here
- `highlight-collaborators-and-own-discussions`: https://github.com/sindresorhus/refined-github/pulls
- `minimize-upload-bar`: here
- `recently-pushed-branches-enhancements`: 🤷

PR #3276, opened 2020-06-24
Update `parse-backticks` style

kidonng:
Match `.markdown-body code`'s `border-radius`.

I'm testing other `border-radius` style to see if they need updating too.


kidonng:
Move this to #3247.

PR #3282, opened 2020-06-25
Update `view-markdown-source` selector

kidonng:
- Fix selector
- Match GitHub's tooltip style (position)

Test: https://github.com/sindresorhus/refined-github/blob/master/contributing.md

**Before**
image
**After**
image

### DIFF
--- a/source/features/view-markdown-source.tsx
+++ b/source/features/view-markdown-source.tsx
@@ -78,12 +78,12 @@ async function init(): Promise<void> {
 	delegate(document, '.rgh-md-source:not(.selected)', 'click', showSource);
 	delegate(document, '.rgh-md-rendered:not(.selected)', 'click', showRendered);
 
-	select('.repository-content .Box-header .d-flex')!.prepend(
+	select('.repository-content > .Box:nth-of-type(3) .Box-header .d-flex')!.prepend(
 		<div className="BtnGroup">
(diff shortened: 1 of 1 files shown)


source/features/view-markdown-source.tsx
```diff
@@ -78,12 +78,12 @@ async function init(): Promise<void> {
 	delegate(document, '.rgh-md-source:not(.selected)', 'click', showSource);
 	delegate(document, '.rgh-md-rendered:not(.selected)', 'click', showRendered);
 
-	select('.repository-content .Box-header .d-flex')!.prepend(
+	select('.repository-content > .Box:nth-of-type(3) .Box-header .d-flex')!.prepend(
```
fregante, line 81: We should try to preserve compatibility with the old design to enable GHE support. 

In this case you can use:

``` js 
const toolbar = select(new selector) ?? selector(old selector)!;
toolbar.prepend( JSX )

fregante: The same applies to most of your other PRs

kidonng: I see, thanks for the reminder!

kidonng:
Well, I should've converted it to draft because I just found a page where the problem still exists: https://github.com/kidonng/cherry/blob/master/Bookmarklets.md

Issue #2649, opened 2019-12-21
Highlight **created** repository count 


Androz2091:
I would like it to be possible to highlight the number of repositories a user **has created**, not the ones he has forked. Indeed, some users fork hundreds of repositories without making any changes to them, and this distorts a little bit the number of repositories. 
Actually, the number is the forks + the created repositories. It would be nice if the number of created repositories were highlighted and the number of forks and if the number of forks was put a little bit in the background.

Example URL:

https://github.com/user

AleksandrHovhannisyan:
I kinda like this idea, but you can already determine those counts using the `Type` dropdown:

![image](https://user-images.githubusercontent.com/19352442/71426525-cc13c480-2678-11ea-8f65-0312effd5009.png)

Androz2091:
Oh, you're right. Couldn't we do something more visible? Like, for example, when you hover over the span, it displays something like this? (and when you're now hovering it, it displays the fork + owned repository count)
<img width="404" alt="Capture d’écran 2019-12-25 à 11 14 20" src="https://user-images.githubusercontent.com/42497995/71442846-f3fe3900-2707-11ea-989a-e1b96060e0d0.png">

fregante:
Since we already show the “Source” repos by default, it would make sense to update that number in the sidebar to only count those. 

The global search has a counter widget, we might be able to use that so that this feature can be implemented as easily as 

```js
tabNumber.replaceWith(<counter query="...."/>) // I don’t remember the widget name actually 
```

AleksandrHovhannisyan:
That's a good point! I didn't notice that :)

fregante:
No longer applies since there are no counters anymore, but refer to https://github.com/sindresorhus/refined-github/pull/3286

PR #898, opened 2017-12-20
Add user gists link to profile

leggsimon:
Closes #863 
screenshot 2017-12-20 15 55 13

~I haven't implemented getting the count of gists yet, if anyone has any suggestions on how to get that, I'd be happy to add it.~ Implemented in https://github.com/sindresorhus/refined-github/pull/898/commits/cbe26d75cb09cdb446cd38065bffdcd1fc770afe

PR #3286, opened 2020-06-26
Extend `profile-gists-link` feature

kidonng:
This is a continuation of https://github.com/sindresorhus/refined-github/pull/3272#issuecomment-648803294. ~~Since we are improving the profile nav, we can move non-working parts of features like `remove-projects-tab` and `hide-zero-packages` to it.~~

This PR ~~includes #3272 and also~~:

- Add counters to profile nav
- Hide empty (`0`) items

Test:

- https://github.com/kidonng (Repository/Package/Gist counts, hide project)
- https://github.com/coee (No counts)
- https://github.com/github (Organization package count)
- https://github.com/airbnb (Organization hide package)


fregante:
Definitely not. Both the feature you mentioned also work on repos, not just profiles.

kidonng:
Oh, I thought that packages are not shown on repos now.

fregante:
Also to keep in mind: https://github.com/sindresorhus/refined-github/issues/2649

kidonng:
Are you saying we should display "Source" (public + not fork + not archived) repo count instead? Unfortunately there is no archive state in the retuned data 😕 So I just exclude forks.

fregante:
I think that's the problem with the API you're using. `totalCount` includes items that shouldn't be included. Use `search()` instead

source/features/extend-profile-nav.tsx
```diff
@@ -4,7 +4,7 @@ import cache from 'webext-storage-cache';
 import select from 'select-dom';
 import elementReady from 'element-ready';
 import * as pageDetect from 'github-url-detection';
-import CodeSquareIcon from '@primer/octicons-v2/build/svg/code-square.svg';
+import CodeSquareIcon from 'octicon/code-square.svg';
```
yakov116: Can you sort the from 
```
import CodeSquareIcon from 'octicon/code-square.svg';
```

kidonng: Shortened and sorted right before your comment

fregante:
> I think that's the problem with the API you're using. `totalCount` includes items that shouldn't be included. Use `search()` instead

This applies to all. AFAIK closed projects are part of the `totalCount` but shouldnn't.

source/features/extend-profile-nav.tsx
```diff
@@ -0,0 +1,131 @@
...
+		packages: user.packages.totalCount,
+		gists: user.gists.totalCount
+	};
+}, {
+	cacheKey: ([username]) => 'user-counts:' + username
+});
+
+const getOrganizationPackageCount = cache.function(async (organizationName: string): Promise<number> => {
+	const {organization} = await api.v4(`
+	  organization(login: "${organizationName}") {
```
fregante, line 49: Can you merge this into the other query? `remove-projects-tab` used to query both users and organizations in one v4 call; if one failed, it was just `undefined` but it kept working.

source/features/extend-profile-nav.tsx
```diff
@@ -0,0 +1,131 @@
...
+}, {
+	cacheKey: ([organizationName]) => 'organization-package-count:' + organizationName
+});
+
+async function extendUserNav(): Promise<void> {
+	await elementReady('.UnderlineNav-body + *');
+
+	const username = getCleanPathname();
+	const href = pageDetect.isEnterprise() ? `/gist/${username}` : `https://gist.github.com/${username}`;
+	const link = (
```
fregante, line 65:
```suggestion
	const gistsLink = (
```

source/features/extend-profile-nav.tsx
```diff
@@ -0,0 +1,131 @@
...
+			packageElement.append(<span className="Counter">{packageCount}</span>);
+		} else {
+			packageElement.remove();
+		}
+	}
+}
+
+void features.add({
+	id: __filebasename,
+	description: 'Extend profile navigation with counts, gist and hide empty items.',
```
fregante, line 117:
```suggestion
	description: 'Adds counters and a link to Gist to user profiles, or hides empty sections.',
```

source/features/extend-profile-nav.tsx
```diff
@@ -0,0 +1,131 @@
...
+	projects: number;
+	packages: number;
+	gists: number;
+}
+
+const getUserCounts = cache.function(async (username: string): Promise<UserCounts> => {
+	const {search, user} = await api.v4(`
+		search(type: REPOSITORY query: "user:${username} archived:false") {
+			repositoryCount
+		}
```
fregante, line 24: Can you merge this into a single API query? 

```suggestion
		}

		organization(login: "${organizationName}") {
			packages {
				totalCount
			}
		}
```

`remove-projects-tab` used to query both users and organizations in one v4 call; if one failed, it was just `undefined` but it kept working.

kidonng:
GitHub added repositories and packages count back:
image

fregante:
I think it might be as is

kidonng:
For the gist counter, I agree with what you said in https://github.com/sindresorhus/refined-github/pull/898#issuecomment-353496843, sometimes I found people prolific in gists and get many useful snippets/info from them, however without the counter it's not likely I'm going to check every user's gist page to find out that.
If you do not mind I can open a new PR to add the gist counter back.

fregante:
>  If you do not mind I can open a new PR to add the gist counter back.

hmm why? This feature also hides empty tabs

kidonng:
It only removes two tabs: projects on user profiles and packages on organization profiles (which needs to make a request). I don't think it's worth it. If you are interested in hiding projects on user profiles, I can probably add that in the new PR, this one has been messy.

kidonng:
I forgot there is already `remove-projects-tab`, but the selector of the user profile is broken now, let's fix that instead.

PR #3287, opened 2020-06-26
Fix `remove-projects-tab` selector

kidonng:
https://github.com/sindresorhus/refined-github/pull/3286#issuecomment-662357039

This PR was initially solving the same problem so I didn't open a new one, I hope this won't cause trouble.


kidonng:
Merged into #3286

PR #3288, opened 2020-06-26
Add `fix-first-tab-length` feature

### LINKED ISSUE #3055: Incorrect tab size in GitHub diff UI

notlmn:
_Shamelessly copied PR description over from #3058_

Replaces #3058
Fixes #3055

**Note**: this works regardless of `remove-diff-signs`’ status.

# Test

Have been using this placed in CSS box in options page for more than 20 days, and haven't see any issues.

[a741080#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R161](https://github.com/sindresorhus/refined-github/commit/a741080862a6a5c2f6f85a331feabd8f81684403#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R161)

Before (short first tab) and after (all tabs are 4-characters long)

tab


fregante:
Please make a new gif to show that the current CSS is working, not the one from the previous PR

notlmn:
@fregante updated prev/new tab change GIF. PTAL.

fregante:
Help wanted testing this. I copy-pasted this into the CSS field for now, but these days I won't be using GitHub much

source/features/fix-first-tab-length.css
```diff
@@ -0,0 +1,18 @@
...
+	position: absolute;
+	transform: translateX(-100%);
+}
+
+/* Restore the space that collapsed as a consequence of the previous rule */
+.file :not(.js-suggested-changes-blob) tr:not(.js-expandable-line) .blob-code,
+.file .js-suggested-changes-blob .blob-code-inner {
+	padding-left: 25px !important; /* GitHub sets it to 18px */
+	text-indent: 0 !important; /* GitHub sets it to -7px */
+}
```
fregante, line 13: Testing this
```suggestion
.file,
	.file :not(.js-suggested-changes-blob) tr:not(.js-expandable-line) .blob-code,
	.file .js-suggested-changes-blob .blob-code-inner {
		padding-left: 25px !important; /* GitHub sets it to 18px */
		text-indent: 0 !important; /* GitHub sets it to -7px */
}
```

source/features/fix-first-tab-length.css
```diff
@@ -0,0 +1,18 @@
...
+
+/* Adjust add comment button position, because of extra padding above */
+.file .add-line-comment {
+	margin-left: -37px !important; /* GitHub sets this at -28px */
```
fregante, line 17:
```suggestion
/* Adjust the "Add comment" button’s position, because of the extra padding above */
.file .add-line-comment {
	margin-left: -37px !important; /* GitHub sets this to -28px */
```

source/features/fix-first-tab-length.css
```diff
@@ -0,0 +1,18 @@
+/* Remove diff signs from the layout flow */
+.file .blob-code-inner::before,
+.file .blob-code-marker-cell::before {
+	position: absolute;
+	transform: translateX(-100%);
+}
+
+/* Restore the space that collapsed as a consequence of the previous rule */
+.file :not(.js-suggested-changes-blob) tr:not(.js-expandable-line) .blob-code,
```
fregante, line 9: Can you double-check this selector? I don't see this anywhere

source/features/fix-first-tab-length.css
```diff
@@ -0,0 +1,18 @@
...
+.file .js-suggested-changes-blob .blob-code-inner {
```
fregante, line 10: The first selector doesn't exclude `.js-suggested-changes-blob` as expected because there are many elements between `.file` and `tr`. This means it can be dropped.


```suggestion
.file tr:not(.js-expandable-line) .blob-code, /* Each line of code except `@@ ... @@` headers */
.file .js-suggested-changes-blob .blob-code-inner { /* Each line of code in PR review comments */
```

fregante:
The gif/screenshot needs to be added to the readme in the "github’s bugs" section

However perhaps it needs some explanation in the image, like "GitHub bug: the first tab appears 2 spaces shorter than it should be". Even a plain screenshot of the "before" state would be enough

readme.md
```diff
@@ -319,6 +319,7 @@ Thanks for contributing! 🦋🙌
 
 ### Fixes for GitHub shortcomings
 
+- [](# "fix-first-tab-length") [The first tab space in file diff appears 2 spaces shorter than it should be.](https://user-images.githubusercontent.com/37769974/86880618-91e8ac80-c10a-11ea-8679-e7b61b253f54.png)
```
fregante, line 322: This looks good, but can you add a red arrow (any style) pointing to the last 2-space tab in the before state? This would make it super clear about what the problem is

Issue #2991, opened 2020-04-10
`show-whitespace` isn't wrapped to next line properly


jack1142:
https://github.com/discord/discord-api-docs/pull/1396/files#diff-99b76a48d014c043f2f8804f6653ed82
![image](https://user-images.githubusercontent.com/6032823/79006779-2cf8d700-7b5a-11ea-9fe5-4a8c6dbce886.png)

There's also a little different problem happening when GitHub Dark is enabled but I'm not sure if this is something you want to fix as well:
![image](https://user-images.githubusercontent.com/6032823/79006869-56b1fe00-7b5a-11ea-8b5c-9ea388f87713.png)

fregante:
This one is though. There's no easy way to show the whitespace without compromising the copy-paste.

Currently we just overlay the •→ symbols via `:before` pseudo-element with `position: absolute`; this can't support the regular `inline` line breaking.

fregante:
A solution could be wrapping each whitespace in its own element. This feature is already set up to only act on the visible lines so it might handle the extra operations without slowing down (except when the user scrolls past a lot of files very quickly)

fregante:
I'm trying this out now, I can't replicate that wrapping behavior at all in both Chrome and Firefox. The spaces never wrap:

## With RG

<img width="882" alt="" src="https://user-images.githubusercontent.com/1402241/87654642-83784180-c757-11ea-8804-3152d0a23eda.png">

## Without RG

<img width="892" alt="" src="https://user-images.githubusercontent.com/1402241/87654754-a99de180-c757-11ea-9d08-dc30219b64fd.png">


cc @notlmn

Issue #2150, opened 2019-06-11
Add option to display invisible/unprintable characters


waldyrious:
This is already mentioned as "Possible Future Additions" in #2073, but I thought it would be helpful to have a dedicated issue to track the implementation.

Ideally, it should be possible to visualize non-printable characters, control characters, zero-width spaces, etc. both when viewing the source code of a file, in code blocks in comments, and on diffs. Especially on diffs.

See for example [how Diffchecker does it](https://www.diffchecker.com/Q4ng5hqs) (note the red dot after "Lorem" on the left text box):

<img width="1125" alt="Screenshot 2019-06-11 at 17 52 42" src="https://user-images.githubusercontent.com/478237/59291013-cffa6f00-8c71-11e9-8a95-99ff974768d2.png">

bfred-it:
I think should would belong to the same feature, which also has its own “tracking issue” here https://github.com/sindresorhus/refined-github/issues/588

bfred-it:
Perhaps I’ll leave this open as a possible extension to that feature

notlmn:
I left that out in the "possible future extensions" section because:
1. Way too many DOM nodes
1. No clear idea on what should be displayed in place of the character
   1. Like what Sublime shows!?
      ![image](https://user-images.githubusercontent.com/37769974/59326377-054f9d00-8d04-11e9-976f-ae1c25f0f857.png)
   1. Or use the [replacement character](https://en.wikipedia.org/wiki/Specials_(Unicode_block)#Replacement_character) like VSCode does
     ![image](https://user-images.githubusercontent.com/37769974/59326782-4300f580-8d05-11e9-95b5-4fd29b4ccfd7.png)
   1. Or [leave with what GitHub already shows](https://github.com/google/WebFundamentals/pull/6755/files#diff-76e41ac4fea56c88fec254a0edb06ce3L119) for **some** non-printable characters
     ![image](https://user-images.githubusercontent.com/37769974/59326917-b0148b00-8d05-11e9-9608-314b1659969b.png)
1. And finally what specific Unicode ranges to handle

tonglil:
I actually like this feature, however maybe there is a way to break it into several options?
- [ ] show all whitespaces
- [x] show whitespaces other than spaces (the dots) `<- my preference`
- [ ] disable (uncheck option in prefs)

allejo:
[...]

PR #3292, opened 2020-06-26
Fix display of wrapped whitespace in `show-whitespace`

notlmn:
Fixes: #2991

Uses clipped SVG background images for spaces and tabs, this makes it more flexible for text wrapping, and any other future characters (#2150).

## Test

* [text wrapping](https://github.com/discord/discord-api-docs/pull/1396/files#diff-99b76a48d014c043f2f8804f6653ed82R515) issue from #2991

* [standalone file](https://github.com/sindresorhus/refined-github/blob/master/source/features/show-whitespace.tsx)

  standalone

* [diff](https://github.com/sindresorhus/refined-github/pull/3168/files)

  diff

* [gist](https://gist.github.com/notlmn/fe7f722971dea345379850bd64516fd2)

  gist

-----

This PR is still in the works, as SVG content used are static paths, which should ideally be text nodes that inherit font from DOM (Figma still flattens all text nodes while exporting).

I tried to match the existing styles (and size) as much as possible, but we might have to reconsider that anyway (i.e. using actual text).

------

**_Edit:_** Updated GIFs

### DIFF
--- a/source/features/show-whitespace.css
+++ b/source/features/show-whitespace.css
@@ -7,11 +7,11 @@
 }
 
 [data-rgh-whitespace='tab'] {
-	background-image: url("data:image/svg+xml,%3Csvg fill='none' viewBox='0 0 12 24' xmlns='http://www.w3.org/2000/svg' preserveAspectRatio='xMinYMid meet' %3E%3Cpath d='M9.5 10.44L6.62 8.12L7.32 7.26L12.04 11V11.44L7.28 14.9L6.62 13.9L9.48 11.78H0V10.44H9.5Z' fill='%23545E5D' fill-opacity='0.5'/%3E%3C/svg%3E");
+	background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' preserveAspectRatio='xMinYMid meet' viewBox='0 0 12 14'%3E%3Cstyle%3E %23text %7B font-family: SFMono-Regular, Consolas, Liberation Mono, Menlo, monospace; line-height: 100%25; %7D %3C/style%3E%3Ctext id='text' x='4.13' y='-2.07' font-size='14' fill='%2324292e' fill-opacity='0.5' alignment-baseline='before-edge' dominant-baseline='text-before-edge'%3E%3Ctspan x='5.88' dy='0em' alignment-baseline='before-edge' dominant-baseline='text-before-edge' text-anchor='middle'%3E→%3C/tspan%3E%3C/text%3E%3C/svg%3E%0A");
 	background-size: calc(var(--tab-size, 4) * 1ch) 1em;
 }
 
 [data-rgh-whitespace='space'] {
-	background-image: url("data:image/svg+xml,%3Csvg fill='none' viewBox='0 0 12 24' xmlns='http://www.w3.org/2000/svg preserveAspectRatio='xMidYMid meet' %3E%3Crect x='4.5' y='9.5' width='3' height='3' rx='1.5' fill='%2324292E' fill-opacity='0.5'/%3E%3C/svg%3E");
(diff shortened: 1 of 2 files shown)


fregante:
Great idea if it works reliably. However I’d preserve the previous opacity, this one isn’t very visible.

notlmn:
That's just the very thin tab character from Inconsolata flattened to `<path>` in SVG.

I'm still working on using `<text>` nodes here to get proper font rendering (if possible). Wanted to put this PR out for sharing the idea.

(As mentioned Figma flattens all text nodes to `<path>`s in SVG, and I'm still afraid to use Inkscape 😅, will have to find some other vector edit tool)

fregante:
You can make the SVG from scratch, it’s just 2 tags (`text` tag inside `svg` tag)

notlmn:
Looks like there is no way for SVG backgrounds (either using `data:` or `http:`) would inherit properties like font and line-height from parent DOM. Even if that were to happen, it would be very complex to handle, especially baseline was very tricky with `background-size`.

Instead I'd propose the use of static `path` elements, this way the backgrounds says consistent on any platform regardless of the font.

Open to other ideas as well.

fregante:
That’s right 👍 let’s match the current color though

notlmn:
> That’s right +1 let’s match the current color though

Done!

source/features/show-whitespace.css
```diff
@@ -1,18 +1,17 @@
...
-[data-rgh-whitespace]::before {
-	content: attr(data-rgh-whitespace);
-	pointer-events: none;
-	user-select: none;
-	position: absolute;
-	left: 0;
-	top: 0;
-	opacity: 25%;
+[data-rgh-whitespace='tab'] {
+	background-image: url('data:image/svg+xml,%3Csvg preserveAspectRatio="xMinYMid meet" viewBox="0 0 12 24" xmlns="http://www.w3.org/2000/svg"%3E%3Cpath d="M9.5 10.44L6.62 8.12L7.32 7.26L12.04 11V11.44L7.28 14.9L6.62 13.9L9.48 11.78H0V10.44H9.5Z" fill="%2324292E" fill-opacity="0.5"/%3E%3C/svg%3E"');
```
fregante, line 10:
```suggestion
	background-image: url('data:image/svg+xml,%3Csvg preserveAspectRatio="xMinYMid meet" viewBox="0 0 12 24" d="M9.5 10.44L6.62 8.12L7.32 7.26L12.04 11V11.44L7.28 14.9L6.62 13.9L9.48 11.78H0V10.44H9.5Z" fill="%2324292E" fill-opacity="0.5"/%3E%3C/svg%3E"');
```

Also, does `fill` support `rgba()`? Let's use that instead of `fill` + `fill-opacity`

source/features/show-whitespace.css
```diff
@@ -1,18 +1,17 @@
...
+	background-size: calc(var(--tab-size, 4) * 1ch) 1.25em;
 }
 
-[data-rgh-whitespace^='→']::before {
-	letter-spacing: calc((var(--tab-size, 4) * 1ch) - 1ch);
+[data-rgh-whitespace='space'] {
+	background-image: url('data:image/svg+xml,%3Csvg preserveAspectRatio="xMinYMid meet" viewBox="0 0 12 24" xmlns="http://www.w3.org/2000/svg"%3E%3Cpath d="M4.5 11C4.5 10.1716 5.17157 9.5 6 9.5C6.82843 9.5 7.5 10.1716 7.5 11C7.5 11.8284 6.82843 12.5 6 12.5C5.17157 12.5 4.5 11.8284 4.5 11Z" fill="%2324292E" fill-opacity="0.5"/%3E%3C/svg%3E');
```
fregante, line 15:
```suggestion
	background-image: url('data:image/svg+xml,%3Csvg preserveAspectRatio="xMinYMid meet" viewBox="0 0 12 24" d="M4.5 11C4.5 10.1716 5.17157 9.5 6 9.5C6.82843 9.5 7.5 10.1716 7.5 11C7.5 11.8284 6.82843 12.5 6 12.5C5.17157 12.5 4.5 11.8284 4.5 11Z" fill="%2324292E" fill-opacity="0.5"/%3E%3C/svg%3E');
```

source/features/show-whitespace.css
notlmn: `fill` does support `rgba()`, but removing `xmlns` doesn't work as this image essentially renders outside of main HTML context and needs its own context.

notlmn: Also let's reduce the opacity compensating thickness. WDYT?

fregante:
This is a clever solution and it appears to work. Let's see if it causes any issues in the real world, I'll publish this today.

PR #3295, opened 2020-06-26
Replace switched off rawgit html preview

SuperSandro2000:
Rawgit announced to shutdown a while ago and now https://ghcdn.rawgit.org/ stopped working and only shows a cloudflare cached page. I have found https://htmlpreview.github.io online and it works quiet good but I am not settled on it. If someone knows a better alternative it should be used.

Test URLs:
https://htmlpreview.github.io/?https://github.com//sindresorhus/refined-github/master/source/options.html ( thanks @kidonng. The rawgit alternative does not work and returns a 502 https://ghcdn.rawgit.org/sindresorhus/refined-github/master/source/options.html.)

### DIFF
--- a/source/features/html-preview-link.tsx
+++ b/source/features/html-preview-link.tsx
@@ -15,7 +15,7 @@ function init(): void {
 		.prepend(
 			<a
 				className="btn btn-sm BtnGroup-item"
-				href={`https://ghcdn.rawgit.org${link.join('/')}`}
+				href={`https://htmlpreview.github.io/?https://github.com/${link.join('/')}`}
 			>
 				Preview
 			</a>


kidonng:
https://htmlpreview.github.io/ seems an optimal choice because other services may shutdown HTML preview feature due to possible blockage, but `github.io` is less likely to be blocked.

fregante:
Include some test URLs like the PR template requests please

kidonng:
https://ghcdn.rawgit.org/ is accessible now.

Test: https://htmlpreview.github.io/?https://github.com//sindresorhus/refined-github/master/source/options.html

SuperSandro2000:
> ghcdn.rawgit.org is accessible now.

But still does not work with new files.

fregante:
This service can probably be implemented via CloudFlare workers by the way, if anyone wants to implement it. All it needs is a fetch and a filter towards html files hosted on GitHub. I’m not sure if it follows the terms of service though. Might be worth contacting CF

SuperSandro2000:
I think we should fix this feature first and then try to develop a new one.

kidonng:
> This service can probably be implemented via CloudFlare workers

2025-09-24 Update: moved to https://github.com/refined-github/refined-github/wiki/Extended-feature-descriptions#html-preview-link

Test: https://refined-github-html-preview.kidonng.workers.dev/ajtoo/Demo-Single-Page/raw/master/index.html

fregante:
Looks perfect! I think it should also force the HTML mime header right? How come it already works without that? 🤔

kidonng:
I think it's content type sniffing. I don't really feel like dealing with content type, but I guess setting the most common ones would be fine. I will update the script.

fregante:
If it already works, no need to add any further code :)

PR welcome to use your worker. Just call it `refined-github-html-preview` or something like that

fregante:
Also, since it's only supposed to work on github.com, I suggest changing it from

```
https://gh-proxy.kidonng.workers.dev/https://github.com/ajtoo/Demo-Single-Page/raw/master/index.html
```

to
```
https://gh-proxy.kidonng.workers.dev/ajtoo/Demo-Single-Page/raw/master/index.html
```

PR #3085, opened 2020-05-12
Add `prevent-pr-commit-link-loss` feature

max-arias:
Closes: https://github.com/sindresorhus/refined-github/issues/2327

---

Note: I had to rename the URL to something other than the URL, otherwise Github replaces the URL anyways with the commit URL in the markdown 🤷‍♂️ .

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

This pull request has been submitted to:
- [#2327: Fix PR commit links on submission (GitHub’s bug)](https://issuehunt.io/repos/51769689/issues/2327)
---
</details>

PR #3296, opened 2020-06-26
Fix `prevent-pr-commit-link-loss` for new Issue / PR

### LINKED ISSUE #3234: Enable `prevent-pr-commit-link-loss` for issue creation
The feature is currently only enabled for comments, but is also needed for new issues.

yakov116:
1. LINKED ISSUES:
   Fixes #3234

2. TEST URLS:
	Test URL `https://github.com/sindresorhus/refined-github/pull/3085/commits/a7b1a8bd85c230ff9ff8e407ca8f6d3646c90eb6`
   https://github.com/sindresorhus/refined-github/issues/new
https://github.com/sindresorhus/refined-github/compare/master...yakov116:pr-parent-commit?expand=1 (hint check ot out)
	To test out the maintainers, try something from your own fork

3. SCREENSHOT:
   None


source/features/prevent-pr-commit-link-loss.tsx
```diff
@@ -9,6 +9,10 @@ import * as textFieldEdit from 'text-field-edit';
 import features from '.';
 import {prCommitUrlRegex, preventPrCommitLinkLoss} from '../github-helpers';
 
+function toggleAllowEditsVisibility({visible}: {visible: boolean}) {
+	select<HTMLInputElement>('[name="collab_privs"]')?.closest('.float-left')?.classList.toggle('d-none', !visible);
```
yakov116, line 13: I could have used `#new_pull_request .float-left` but I felt it was not specific enough =.

yakov116:
@FloEdelmann can you help review?

source/features/prevent-pr-commit-link-loss.tsx
fregante: Why do this at all?

yakov116: Before
image

After

image

fregante: Can the conflict be resolved with CSS instead? Doesn’t sound like JS should be needed

yakov116: @kidonng Can you try to help me find a CSS solution? (Revert the last commit to see the issue)

source/features/prevent-pr-commit-link-loss.tsx
```diff
@@ -46,7 +46,19 @@ void features.add({
...
+	include: [
+		pageDetect.isNewIssue,
+		pageDetect.isCompare
+	],
+	exclude: [
+		() => select.exists('.form-actions')
+	],
+	init: () => {
+		// New issues and Pull Requests are missing the form-actions selector, this avoids having a separate logic just for them.
+		select('#new_issue tab-container, #new_pull_request tab-container')?.after(<div className="form-actions"/>);
```
FloEdelmann: @yakov116 If you reset to your first commit only, and then apply this suggestion, it should work :)

The culprit is `float: left;` of the "Allow edits" checkbox, so we need a `clear: left;` before our warning, which is added by the `clearfix` class.
The utility classes `pb-0` and `pt-2` fix the spacing around the warning.

```suggestion
		const formTabContainer = select('#new_issue tab-container, #new_pull_request tab-container');
		formTabContainer?.classList.add('clearfix');
		formTabContainer?.after(<div className="form-actions pb-0 pt-2"/>);
```

source/features/prevent-pr-commit-link-loss.tsx
FloEdelmann: See https://github.com/sindresorhus/refined-github/pull/3296#discussion_r448624228 for a CSS-only solution.

source/features/prevent-pr-commit-link-loss.tsx
yakov116: Thanks!! will try it in a few

yakov116:
Awesome!! Thanks @FloEdelmann! Works perfectly.

source/features/prevent-pr-commit-link-loss.tsx
```diff
@@ -46,7 +46,22 @@ void features.add({
...
+}, {
+	include: [
+		pageDetect.isNewIssue,
+		pageDetect.isCompare
+	],
+	exclude: [
+		() => select.exists('#new_issue .form-actions, #new_pull_request .form-actions')
+	],
+	init: () => {
+		// New issues and Pull Requests are missing the form-actions selector, this avoids having a separate logic just for them.
```
fregante, line 61: You mention that this code is “to avoid separate logic just for them,” but this code is exactly “separate logic just for them,” so let’s integrate this in the original code if necessary

yakov116: Let me explain what I mean by changing the logic.

The entire code  (hiding and showing the bar) is based off the fact that `.form-actions` exists. My idea is to just add the `div` and keep the rest of the code intact.

> so let’s integrate this in the original code if necessary, instead of adding an init function

The **current code** would not make sense to be part of the original code. We would need to check if form actions exists and if its a new pull request/issue. IMHO an init is simpler and cleaner.

> This is still not CSS-only

The only js that could be moved to css is the `formTabContainer?.classList.add('clearfix');`. Do you want to add a file just for that?

@fregante if you feel that there is a better solution then the one I implemented. Then please feel free to close.

fregante: > is based off the fact that `.form-actions` exists

Is it? Why can't that selector be changed instead? Why add a dummy element instead of just changing a selector?

fregante: > The only js that could be moved to css is the `formTabContainer?.classList.add('clearfix');`. Do you want to add a file just for that?

You can add the class/rule to `rgh-warning-for-disallow-edits` instead of here

yakov116: I tried to look for another one but nothing looked right. (correct position and not messing anything else up)

I will give it another shot later tonight.

yakov116: I tried again and did not see any selector to append to

PR #3297, opened 2020-06-27
Drop `hide-zero-packages` feature

kidonng:
In the new design, GitHub changed how they show "Packages" tab:

- repos: only shown to collaborators (~~Is there a need to hide them as well?~~)
  <details>

  image
  
  </details>
- users: only show if they have packages, without counter (**and `hide-zero-packages` will always activate, this PR fixes that**)
- organization: always show without counter

~~Test: https://github.com/kidonng~~

Drop it according to https://github.com/sindresorhus/refined-github/pull/3297/files#issuecomment-652513782


fregante:
Just to confirm:

- repos: only shown to collaborators 👍 
- users: not shown 🎉 
- organization: still shown (and should still be hidden) 😕

source/features/hide-zero-packages.tsx
```diff
@@ -7,7 +7,7 @@ import getTabCount from './remove-projects-tab';
 async function init(): Promise<void | false> {
 	const packagesTab = await elementReady([
 		'.numbers-summary [href$="/packages"]', // `isRepoRoot`
-		'.UnderlineNav-item[href$="?tab=packages"]:not(.selected)', // `isUserProfile`
+		'.user-profile-nav .UnderlineNav-item[href$="?tab=packages"]:not(.selected)', // `isUserProfile`
```
fregante, line 10: It’s worth adding the “pre-repository refresh” comment

kidonng:
> Just to confirm:
> 
> * repos: only shown to collaborators 👍
> * users: not shown 🎉
> * organization: still shown (and should still be hidden) 😕

Great summary! I updated my description to make it clear.

Just to mention that if we want to hide “Packages” tab on organization, we need to query the API.

fregante:
So, is it correct to say that this feature currently does not apply to GitHub at all?

Perhaps it should be dropped and moved to the "added by github" section in the readme. I expect the organization profile to follow the user profile in hiding the packages.

kidonng:
No, the third rule (organization) still works.

fregante:
> No, the third rule (organization) still works.

I still see the tab, so I think it doesn't work

kidonng:
😅 My mistake, I wrote "always show without counter" and forget what does it mean. You are right, it's only for old GitHub now.

readme.md
```diff
@@ -338,7 +337,7 @@ Thanks for contributing! 🦋🙌
 
 ### Previously part of Refined GitHub
 
-GitHub implemented 24 features that used to be part of Refined GitHub 🎉
+GitHub implemented 25 features that used to be part of Refined GitHub 🎉
```
fregante, line 340:
```suggestion
GitHub implemented dozens of features that used to be part of Refined GitHub 🎉
```

PR #3304, opened 2020-07-01
Update `view-markdown-source` selector again

kidonng:
Continue of https://github.com/sindresorhus/refined-github/pull/3282#issuecomment-652526963

I forgot that `Element.querySelector` uses depth-first pre-order traversal, so it will match the old selector sometimes.

Test:
- https://github.com/sindresorhus/refined-github/blob/master/readme.md
- https://github.com/sindresorhus/refined-github/blob/master/contributing.md
- https://github.com/kidonng/cherry/blob/master/Bookmarklets.md


kidonng:
The `nth-of-type` is not stable as well because there may be other boxes:
image

PR #3305, opened 2020-07-01
Use CloudFlare Workers for `html-preview-link`

kidonng:
Close #3295

Tests

- https://github.com//sindresorhus/refined-github/blob/master/source/options.html -> https://refined-github-html-preview.kidonng.workers.dev/sindresorhus/refined-github/raw/master/source/options.html
- https://github.com/ajtoo/Demo-Single-Page/blob/master/index.html -> https://refined-github-html-preview.kidonng.workers.dev/ajtoo/Demo-Single-Page/raw/master/index.html

### DIFF
--- a/source/features/html-preview-link.tsx
+++ b/source/features/html-preview-link.tsx
@@ -4,18 +4,16 @@ import * as pageDetect from 'github-url-detection';
 
 import features from '.';
 
-const isSingleHTMLFile = (): boolean => pageDetect.isSingleFile() && (location.pathname.endsWith('.html') || location.pathname.endsWith('.htm'));
+const isSingleHTMLFile = (): boolean => pageDetect.isSingleFile() && /[.]html?$/.test(location.pathname);
 
 function init(): void {
 	const rawButton = select<HTMLAnchorElement>('#raw-url')!;
-	const link = rawButton.pathname.split('/');
-	link.splice(3, 1); // Remove /raw/
 	rawButton
 		.parentElement! // `BtnGroup`
 		.prepend(
 			<a
 				className="btn btn-sm BtnGroup-item"
-				href={`https://ghcdn.rawgit.org${link.join('/')}`}
+				href={`https://refined-github-html-preview.kidonng.workers.dev${rawButton.pathname}`}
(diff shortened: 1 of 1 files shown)


fregante:
Excellent! Might be worth redirecting the root (https://refined-github-html-preview.kidonng.workers.dev/) to this PR.

kidonng:
Great idea, now the root is redirecting to this PR.

fregante:
Some pages are showing on Google: https://www.google.com/search?q=site:refined-github-html-preview.kidonng.workers.dev

Can you add an `X-Robots-Tag: noindex, nofollow, noarchive` header? https://developers.google.com/search/reference/robots_meta_tag

kidonng:
Oops, that's a big oversight. Added `x-robots-tag` and `robots.txt`, you can refer to the [updated code](https://github.com/sindresorhus/refined-github/pull/3295#issuecomment-651500710).

fregante:
`none` doesn't include `noarchive`, which I think means that those 3 links won't be removed.

Also probably we should drop robots.txt because that prevents Google from seeing the robots tag. If I remember correctly, if Google isn't allowed to fetch the links it finds, it will still list them on the SERPs as _"Google couldn't fetch the description due to the robots.txt file. more info"_

kidonng:
Just noticed [how this works](https://developers.google.com/search/docs/advanced/crawling/block-indexing):

> For the noindex directive to be effective, the page must not be blocked by a robots.txt file. If the page is blocked by a robots.txt file, the crawler will never see the noindex directive, and the page can still appear in search results, for example if other pages link to it.

So I made some changes accordingly. Thanks for the reminder!

I have also submitted removal requests for existing results:

image

nativeOS-Project:
Hi

mvy-siteimprove:
FYI: this feature doesn't work well with private repos - in fact, it _looks_ like a phishing attack, since it renders github 404 screen with the `refined-github-html-preview.kidonng.workers.dev` in the address bar (and the login form is focused 🙈).

fregante:
Good point, private repos will be excluded

PR #3306, opened 2020-07-01
Use webextension-manifest-loader

jsmnbom:
I was asked by @fregante to try to implement my [webextension-manifest-loader](https://github.com/jsmnbom/webextension-manifest-loader) for the build system.
https://github.com/fregante/Awesome-WebExtensions/pull/53#issuecomment-652426849

The targetVendor of the webextension-manifest-loader is set to "chrome" always, as currently your build system doesn't have separate commands for vendors - and doesn't really need it for the manifest.json (like [this](https://github.com/jsmnbom/ao3-enhancements/blob/main/src/manifest.json) - see the \_\_firefox\_XXX\_\_ and \_\_chrome\_XXX\_\_)

This PR is a draft due to a couple of potential shortcomings:

- The content_scrips needs to be run at "document_end" instead of "document_start". This is due to using style-loader (which needs at least the DOM to be loaded) instead of MiniCssExtractPlugin which has incompatibilities that I've so far been unable to fix. (extract-loader could maybe be used here, but that would create many many seperate css files which isn't optimal either)
- I had multiple linting errors even before i began work, so there might be stuff that I've missed in the setup...
- It's a bit of hacky solution to be honest. Webpack so far is only made to use js files as entry points (https://github.com/webpack/webpack/issues/536).

I should theoretically be possible to clean up the manifest.json up even further (remove the ~file-loader?name=[name].[ext]! and ?entry parts), but I've as of yet been unsuccessful in getting webpack to load js files from node_modules using file-loader by configuring it in the webpack config rules.


fregante:
>  content_scrips needs to be run at "document_end" instead of "document_start"

I think this is a deal breaker. The functionality should not change; this change would make Refined GitHub feel much slower

jsmnbom:
That's totally fair. I'll see if I can find a solution for that, and if I can't I'll go ahead and close this PR.

fregante:
Perhaps you can exclude the polyfill file entirely and let the Copy plugin handle that. That should simplify the situation and maybe avoid the limitation

jsmnbom:
The issue isn't actually with the polyfill file though. It's that mini-css-extract-plugin runs into an issue like https://github.com/webpack-contrib/mini-css-extract-plugin/issues/489 meaning we can't use it (the error happens somewhere deep inside webpack and I cannot figure it out at all).
The issue only happens with the css that's imported directly in `refined-github.ts` and the imported features.

Therefore a possible solution would be to use css `@import`s in the `refined-github.css` file. Then we can use our faithfull file-loader+extract-loader combo directly. An upside to this solution is that then the script and css will be clearly seperate. That upside is also a bit of a downside as now eah feature isn't as neatly contained. For me, the main reason I wrote webextension-manifest-loader was so one wouldn't have to do stuff like importing css within js/ts files, but could instead let the loader handle the css loading.

Would this be an alright solution?

fregante:
I think that having the JS file load its own style sheet makes it easier to keep track of it (“does this feature have a css file? Is the css file being loaded?”) so the current situation is preferable. 

It’s too bad that Webpack still isn’t as flexible as it purports to be. Years ago they said they’d make CSS a first class citizen yet we still have to “extract” it with a plugin  🤷‍♂️ 

Hopefully Parcel 2 will support web extensions natively soon

jsmnbom:
Yea I suppose you're right that for this use case having it together makes more sense.

Okay so in terms of how it was done before, the stylesheet was included in the content_scripts css property, and it was then injected at document_start. But as far as I understand how browsers work and what my observations have led me to believe (and do correct me if I'm wrong :)), the browser then proceeds to load the entire html document before then showing anything. This means that we inject the stylesheet at Document.readyState == 'interactive', it should show immediately when the page loads.

I've pushed a change where this is the case and I can't tell a difference in how quickly Refined Github features visually load now. (and since it's how using "document_start", the script will load just as quickly as it did before).

fregante:
> the browser then proceeds to load the entire html document before then showing anything.

No, on long pages/slow connections the browser starts rendering _before_ it that `readyState`.

The loading mechanism **cannot** be changed _"just because the build doesn't support it"_

jsmnbom:
Yea that's totally fair. I personally think that the benefit of using a manifest.json as entry point is nice and makes a lot of sense. But you're totally right that it shouldn't come at the cost of performance. I'll close this PR and if I ever https://github.com/webpack-contrib/mini-css-extract-plugin/issues/489 I'll consider opening it up again if I can get it work properly. Thanks for your time :)

fregante:
Thanks for sending a PR!

PR #3308, opened 2020-07-01
Rename `discussion` → `conversation`

FloEdelmann:
Follows https://github.com/fregante/github-url-detection/pull/17.


fregante:
Awesome!

fregante:
It might be worth deleting `getRepoURL` and adding it to this line https://github.com/sindresorhus/refined-github/blob/daf633137ec08e6f5647a6f8bb0b02d5dce58ef9/source/github-helpers/index.ts#L8

FloEdelmann:
> It might be worth deleting `getRepoURL` and adding it to this line https://github.com/sindresorhus/refined-github/blob/daf633137ec08e6f5647a6f8bb0b02d5dce58ef9/source/github-helpers/index.ts#L8

Oh, indeed, I wanted to do that but forgot in this PR. We have to make it lowercase though, to not undo #2892.

PR #3015, opened 2020-04-19
Drop `deemphasize-unrelated-commit-references`

fregante:
The feature has been added by GitHub

PR #3310, opened 2020-07-02
README updates

kidonng:
- Use `github.blog` domain for blog posts
- Add an extra (missing?) item in "Previously part of Refined GitHub"
- Fix the wrong number of GitHub implemented features (possibly due to #3015, which drops a feature without increasing the count).


fregante:
I mistyped when merging... anyway. It looks good!

Issue #3280, opened 2020-06-25
Properly align tab navigation


szmarczak:
Example: https://github.com/sindresorhus/got

![image](https://user-images.githubusercontent.com/36894700/85687756-72489180-b6d1-11ea-81bd-220abec757da.png)

kidonng:
#3250

PR #3313, opened 2020-07-03
Add `align-repository-header` feature

fregante:
Fixes https://github.com/sindresorhus/refined-github/issues/3280
Replaces and closes #3250

Instead of aligning both sides and limiting the available space for the tabs, this only aligns the left side

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,8 @@
 			"selector-type-no-unknown": null,
 			"declaration-no-important": null,
 			"selector-class-pattern": null,
-			"selector-id-pattern": null
+			"selector-id-pattern": null,
+			"selector-max-universal": null
 		}
 	},
 	"ava": {
--- a/source/features/align-repository-header.css
+++ b/source/features/align-repository-header.css
@@ -1,0 +1,9 @@
+/* Align repo header to the content, just on the left side*/
+.repohead > * {
+    padding-left: calc(50% - 1280px / 2) !important;
+}
+
+/* Restore the original padding on those 2 elements */
+.repohead > * > :first-child {
+    margin-left: 32px;
+}
--- a/source/features/align-repository-header.css
+++ b/source/features/align-repository-header.css
@@ -1,9 +1,19 @@
-/* Align repo header to the content, just on the left side*/
+/* Align repo header to the content, just on the left side */
 .repohead > * {
 	padding-left: calc(50% - 1280px / 2) !important;
 }
 
-/* Restore the original padding on those 2 elements */
-.repohead > * > :first-child {
-	margin-left: 32px;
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -26,7 +26,8 @@
 			"selector-type-no-unknown": null,
 			"declaration-no-important": null,
 			"selector-class-pattern": null,
-			"selector-id-pattern": null
+			"selector-id-pattern": null,
+			"selector-max-universal": null
```
fregante, line 30: @sindresorhus this rule seems rather arbitrary

sindresorhus: `*` is very inefficient and far-reaching. I think it makes sense to limit it in normal code. RG is an expectation since it has to do a lot of overrides, that you wouldn't do in a normal codebase.

source/features/align-repository-header.css
```diff
@@ -0,0 +1,21 @@
+/* Align repo header to the content, just on the left side */
+.repohead > * {
+	padding-left: calc(50% - 1280px / 2) !important;
```
kidonng, line 3: This is overridden by GitHub:

image

fregante: How’s that possible? The selector is less specific than ours

FloEdelmann: @fregante `*` has a specificity of zero, so they are equally specific and the latter overrides the former.

fregante: That’s weird. I’ve always thought that `>` increased specificity. 

I guess I can replace a star with a tag name and solve both issues reported in this PR

fregante: Can't, there are 2 tag types in each level (div/nav > div/ul) and `:is()` isn't supported by Chrome yet

kidonng: It's still overridden. Can we use `.repohead > div, .repohead > nav`?

yakov116:
Looks like this has an adverse effect on gists
Before
image
After
image

yakov116:
Looks like this has an adverse effect on gists
Before
image
After
image

kidonng:
`.repohead` -> `.repohead:not(.gisthead)`

epmatsw:
Looks like this breaks enterprise users: https://github.com/sindresorhus/refined-github/issues/3330

fregante:
PR welcome to fix gists and #3332

szmarczak:
How can I use this? I'm having `20.7.16` and I still see no difference...

kidonng:
@szmarczak Because it's broken now https://github.com/sindresorhus/refined-github/pull/3362#issuecomment-659955345

PR #3314, opened 2020-07-03
Lint and avoid errors

### LINKED ISSUE #3134: `clean-sidebar`: can't access property "nextElementSibling", header is null
When having a repository with "projects" turned off, this `header` is null https://github.com/sindresorhus/refined-github/blob/173e39faf0fadc97e134d06d61911ebd86b418fb/source/features/clean-sidebar.tsx#L37-L43
but there is only a null check for `nextElementSibling` and not `header`

A link to a repo is not so valuable as you have to be an owner of the repo in order to see this bug. 😛

### LINKED ISSUE #3307: Clone or Download > Open in Desktop button is missing
Open in Desktop button is missing in the "Clone or Download" dialog.

With extension:

Screenshot 2020-07-01 at 23 00 24

Without extension:

Screenshot 2020-07-01 at 23 00 35

This seems to be caused by https://github.com/sindresorhus/refined-github/pull/3211. It used to be that in Chrome (but not Firefox) the CSS for `js-remove-unless-platform` was not used and GitHub's own CSS took higher precedence for the element. However, after latest changes to HTML this isn't the case.

### LINKED ISSUE #3309: Review requires comment even with pending comment(s)
Run into an issue with submitting a review with pending comments.
This happens because the pending comment isn't visible in the current page.
*(e.g. when stepping through the commits one by one)*

I think this was added in 52991fea170973356c3fd54edd14b8ff046c053a, happy to provide a PR to fix this, but I wasn't sure what edgecase that commit fixes.

image

### LINKED ISSUE #3312: Options page in Edge Chromium has fixed width
Not really a *bug*, but I would like to open it here to track the issue.

This is how options page looks in Edge Chromium (85.0.552.1, Windows 10):

image

The width is defined by the following line, which is initially added by #934:

https://github.com/sindresorhus/refined-github/blob/4e0245d3faf0518e3e22358cdc5315e4e0c2f13c/source/options.css#L3

It seems Edge has a fixed width for options popup and won't adjust it responsively:

image

yakov116:
Yeah that time of week LOL
	

- Fix `releases-tab` again. Not all repositories have it on the side bar, yet they all have it on the main bar.
- Exclude pr file 404 from `pr-jump-to-first-non-viewed-file`
- Fix projects not available of `clean-sidebar` fixes #3134
- Fix min width Options page in Edge Chromium fixes #3312
- Fix missing open on desktop fixes #3307
- Fix "comment required" detection in `quick-review-buttons` fixes #3309 
- Restore `hide-watch-and-fork-count` border https://github.com/sindresorhus/refined-github/pull/3247#discussion_r449607417
- `reaction-avatars` Use css class instead of inline style https://github.com/sindresorhus/refined-github/pull/3247#pullrequestreview-442442869
- Run `align-issue-labels` only on `isConversationList`


source/features/pr-jump-to-first-non-viewed-file.tsx
```diff
@@ -28,5 +28,8 @@ void features.add({
 	include: [
 		pageDetect.isPRFiles
 	],
+	exclude: [
+		pageDetect.isPRCommit404
```
yakov116, line 32: @fregante we need another detection for 
https://github.com/sindresorhus/eslint-plugin-unicorn/pull/789/files/e01aeec9777d9bb12d418bffaf37221d5f268976..a58b37845f1b2660221de019e4ae6c736feedc26

is that a new detection `isPRFiles404` or join it as one?

fregante: I wonder if `isPRFiles` should automatically exclude 404 pages.

I'm a fan of potentially having to add this `exclude 404` on every feature that requires it. I almost don't care that the lack of it causes errors in the console (it's a literal 404 error page after all)

yakov116:
Was going to do 

>It might be worth deleting `getRepoURL` and adding it to this line https://github.com/sindresorhus/refined-github/blob/daf633137ec08e6f5647a6f8bb0b02d5dce58ef9/source/github-helpers/index.ts#L8

_Originally posted by @fregante in https://github.com/sindresorhus/refined-github/pull/3308#issuecomment-653246212_

But dont this we gain anything since

>We have to make it lowercase though, to not undo #2892.

_Originally posted by @FloEdelmann in https://github.com/sindresorhus/refined-github/pull/3308#issuecomment-653373118_

source/features/pr-jump-to-first-non-viewed-file.tsx
yakov116: > I'm a fan of potentially having to add this `exclude 404` on every feature that requires it.

lost you on this one.

yakov116: Nm. I just used exclude

yakov116:
https://github.com/sindresorhus/refined-github/blob/0fa514dbeeab6395ef095cff80beac41443a6574/source/features/copy-on-y.tsx#L8

@fregante can this be changed to `isEditable`?

fregante:
Yes

source/features/quick-review-buttons.tsx
```diff
@@ -73,7 +74,7 @@ function init(): false | void {
 
 	// This will prevent submission when clicking "Comment" and "Request changes" without entering a comment and no other review comments are pending
 	delegate<HTMLButtonElement>(form, 'button', 'click', ({delegateTarget: {value}}) => {
-		const submissionRequiresComment = !select.exists('.review-comment-contents > .is-pending') && (value === 'reject' || value === 'comment');
+		const submissionRequiresComment = looseParseInt(select('.js-reviews-toggle .js-pending-review-comment-count')!.textContent!) === 0 && (value === 'reject' || value === 'comment');
```
fregante, line 77: Let’s extract this count into `const pendingComments`

source/options.css
```diff
@@ -1,6 +1,12 @@
+@media not screen and (width: 400px) {
+	/* Excludes Edge */
```
fregante, line 2: You moved the comment. It should be next to the query because that’s what it’s about

kidonng:
@yakov116 Could you also add https://github.com/sindresorhus/refined-github/pull/3247#discussion_r449607417?

source/options.css
yakov116: Sorry vs code formatter messed it up. Will fix

source/features/esc-to-deselect-line.tsx
```diff
@@ -11,11 +11,11 @@ function isLineSelected(): boolean {
...
 }
 
-function listener(event: KeyboardEvent): void {
+function listener({key, target}: KeyboardEvent): void {
 	if (
-		event.key === 'Escape' && // Catch `Esc` key
+		key === 'Escape' && // Catch `Esc` key
 		isLineSelected() &&
-		!isEditable(event.target) // If a field isn’t focused
+		!isEditable(target) // If a field isn’t focused
```
yakov116, line 18: Can these comments be deleted?

fregante:
Regarding `reaction-avatars`: GitHub has a native component for that, it’s called “avatar stack” and we could use it to drop some of our custom CSS. It can be found on the dashboard. 

However that should probably be its own PR because I don’t know if it works as well as ours

yakov116:
I think I'm done here. 😄

source/features/reactions-avatars.tsx
```diff
@@ -62,7 +62,7 @@ async function showAvatarsOn(commentReactions: Element): Promise<void> {
 		container.append(
 			// Without this, Firefox will follow the link instead of submitting the reaction button
 			<a href={isFirefox ? undefined : `/${username}`} className="rounded-1 avatar-user">
-				<img src={imageUrl} style={{borderRadius: 'inherit'}}/>
+				<img src={imageUrl} className="rounded-1"/>
```
kidonng, line 65: You should use `className="avatar-user rounded-1"` becasue `rounded-1` is for old design and has smaller radius:
image

yakov116:
```suggestion
				<img src={imageUrl} className="avatar-user rounded-1"/>
```

fregante: @yakov116 Why?

yakov116: Followed @kidonng https://github.com/sindresorhus/refined-github/pull/3314#discussion_r449885938
> You should use `className="avatar-user rounded-1"` because `rounded-1` is for old design and has smaller radius (`6px`):

source/features/esc-to-deselect-line.tsx
```diff
@@ -11,12 +11,8 @@ function isLineSelected(): boolean {
...
 }
 
-function listener(event: KeyboardEvent): void {
-	if (
-		event.key === 'Escape' && // Catch `Esc` key
-		isLineSelected() &&
-		!isEditable(event.target) // If a field isn’t focused
-	) {
+function listener({key, target}: KeyboardEvent): void {
+	if (key === 'Escape' && isLineSelected() &&	!isEditable(target)) {
```
fregante, line 15:
```suggestion
	if (key === 'Escape' && isLineSelected() && !isEditable(target)) {
```

source/options.css
```diff
@@ -1,6 +1,11 @@
+@media not screen and (width: 400px) { /* Excludes Edge */
```
fregante, line 1: Was this query tested in Edge?

source/features/pr-jump-to-first-non-viewed-file.tsx
```diff
@@ -28,5 +28,8 @@ void features.add({
 	include: [
 		pageDetect.isPRFiles
 	],
+	exclude: [
+		() => document.title.startsWith('Commit range not found · Pull Request')
+	],
```
fregante, line 33: Can you move this to `github-url-detection`?

source/options.css
yakov116: >It works in both Edge and Chrome 😄

_Originally posted by @kidonng in https://github.com/sindresorhus/refined-github/issues/3312#issuecomment-653560583_

source/features/pr-jump-to-first-non-viewed-file.tsx
yakov116: `isPr404`?

fregante: `isPR404Commit`

source/options.css
yakov116: Tested

kidonng:
Could you please revert another mistake in https://github.com/sindresorhus/refined-github/pull/3247#discussion_r450173111 again 😅? (and you should use `6px` now)

source/features/minimize-upload-bar.css
```diff
@@ -5,4 +5,5 @@
 
 .rgh-minimize-upload-bar .js-upload-markdown-image.is-default textarea {
 	border-bottom-style: solid;
+	border-radius: 3px;
```
kidonng, line 8: Well, `6px`
image

yakov116:
```suggestion
	border-radius: 6px;
```

source/features/pr-jump-to-first-non-viewed-file.tsx
```diff
@@ -28,5 +28,8 @@ void features.add({
 	include: [
 		pageDetect.isPRFiles
 	],
+	exclude: [
+		() => document.title.startsWith('Commit range not found · Pull Request')
```
yakov116, line 32:
```suggestion
		pageDetect.isPRFile404
```

yakov116:
@fregante anything wait on me?

yakov116:
@fregante did you get the new preview? If yes from what i see we now have an issue detecting permalinks from the branch selector. It does not say if its a tag/tree anymore.

fregante:
Let's handle that in a separate PR once that design leaves beta.

fregante:
In your next round you could change this line to use `github-url-detection` https://github.com/sindresorhus/refined-github/blob/da8313019d31ab4bdf3fd4a5f75be961d95a6a9a/source/features/file-finder-buffer.tsx#L18

yakov116:
> In your next round you could change this line to use `github-url-detection`

Next round?? Why would there be that? Lol

source/features/align-issue-labels.tsx
```diff
@@ -12,5 +14,8 @@ void features.add({
 }, {
 	waitForDomReady: false,
 	repeatOnAjax: false,
```
yakov116, line 16: any reason this is not working correctly with `repeatOnAjax: false`

fregante: Because sometimes it goes from `isPR` to `isConversationList` with an ajax load, but `repeatOnAjax: false` prevents the page detection

PR #3316, opened 2020-07-05
Update dependencies; make Firefox 78 the minimum version

yakov116:
Follows https://github.com/sindresorhus/refined-github/pull/2410#issuecomment-643135144

I updated all the dependencies too. Let me know if I should revert or if you want to do this entirely on your own.


fregante:
Finally!

fregante:
Is it possible to change the target ES version in tsconfig? `??` doesn't need to be transpiled anymore, but `acorn` support in `webpack` made it impossible to change the `target`

yakov116:
I can try later today. I dont **think** its possible without updating to webpack v5 (where acorn was updated).

yakov116:
> Is it possible to change the target ES version in tsconfig?

No

Issue #3265, opened 2020-06-22
`restore-file` error


MaxLeiter:
When I try to 'restore' a file I deleted this error occurs

yakov116:
That error does not sound related.
The more-dropdown error is https://github.com/sindresorhus/refined-github/issues/3215

Do you get any other error when using the restore file?

MaxLeiter:
Originally I got an error about no token, but then I added one. When I try to restore the file it says "Committing..." then to check in console.

yakov116:
Can you paste the **full** error from the console after you got that message?

MaxLeiter:
The only error in my console:

```

yakov116:
That error is https://github.com/sindresorhus/refined-github/issues/3081 and its followup #3215

fregante:
The error reported is unrelated, but there might still be an error to debug, especially if is says to check the console and there’s nothing there.

fregante:
However, we need a link to the page where your error appears and a screenshot of it, like the issue template suggests

fregante:
Closing as unreproducible/incomplete

Issue #3215, opened 2020-06-10
`more-dropdown` compatibility with "Repository refresh" design


fregante:
`more-dropdown` is kind of a problem now since the new design has an automatic "overflow" hamburger menu:

<img width="588" alt="" src="https://user-images.githubusercontent.com/1402241/84262936-d50d2b00-ab1e-11ea-82da-85002d1e62f1.png">

The obvious solution is to add all those elements to the navigation and let them naturally flow into the overflow, but that's noisy. 

Until now we added items to the dropdown without much trouble, since it was "out of the way", but this change forces us to reconsider it.

_Originally posted by @fregante in https://github.com/sindresorhus/refined-github/issues/3081#issuecomment-641942885_

cc @sindresorhus

aminya:
Not sure if these errors are related to this, but I get them in the new design:
![image](https://user-images.githubusercontent.com/16418197/84377496-4513c380-aba8-11ea-8cc4-f12b5b6d5487.png)

yakov116:
@aminya thats https://github.com/sindresorhus/refined-github/issues/3081

fregante:
Is it possible to have the dropdown on the bar, but when it "overflows" its contents end up in the native dropdown?

E.g.


```
Wide viewport
 _________________________________________________________________________
|                                                                         |
| Code - Issues - Pull requests - Actions - [ More ▼ ]                    |
|                                            Commits                      |
|                                            Branches                     |
|                                            Compare                      |
|_________________________________________________________________________|
```

Becomes


```
Small viewport
 ___________________________________________
|                                           |
| Code - Issues - Pull requests - [ ... ]   |
|                                 Actions   |
|                                 Commits   |
|                                 Branches  |
|                                 Compare   |
|___________________________________________|
```

kidonng:
> Is it possible to have the dropdown on the bar, but when it "overflows" its contents end up in the native dropdown?

Isn't this what we have in #3186?

notlmn:
We can start by adding a `max-width` to `.js-repo-nav`, but that still keeps the dropdown trigger to the right.

There's some layout issues using `position: relative` on nav, and there's also figuring out `max-width` in the first place.

Just some thoughts.

fregante:
I like the max-width solution, but it’s a little risky because other extensions add tabs too. 

@kidonng yes, but that’s just 1 item out, 1 item in. I’m talking about 1 item out (the More dropdown) and 5 items in (once it overflows): could that work?

[...]

PR #3321, opened 2020-07-06
Disable `more-dropdown`

FloEdelmann:
While we haven't figured out a solution to #3215, the feature will throw an error on every repo page.

By disabling the feature, we can prevent issues like #3265.

### DIFF
--- a/source/features/more-dropdown.tsx
+++ b/source/features/more-dropdown.tsx
@@ -78,6 +78,7 @@ async function init(): Promise<void> {
 }
 
 void features.add({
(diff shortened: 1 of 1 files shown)


yakov116:
Why should GHE users not be able to use it.

If anything maybe exclude where location.origin = github.com

fregante:
Nah, the error message is super clear that it’s about `more-dropdown` and not `restore-file`

Sure it’s annoying to see it there, but also I want it to be fixed soon, so maybe it *should* be annoying

PR #3322, opened 2020-07-07
Limit `align-repository-header` to non-full-width pages

fregante:
Some pages are full-width, so they don't need to be "aligned".

- Centered: https://github.com/sindresorhus/refined-github
- Centered: this very page
- Centered: https://github.com/sindresorhus/refined-github/pull/3322/files?diff=unified
- Full width: https://github.com/sindresorhus/refined-github/pull/3322/files?diff=split
- Full width: https://github.com/sindresorhus/refined-github/blame/master/.gitattributes
- Full width: https://github.com/sindresorhus/refined-github/commit/f87c46617028be1882047bc3cf3166de94c86e14?diff=split

### DIFF
--- a/source/features/align-repository-header.css
+++ b/source/features/align-repository-header.css
@@ -1,21 +1,21 @@
 /* Align repo header to the content, just on the left side */
-.page-responsive .repohead > * {
+body:not(.full-width) .repohead > * {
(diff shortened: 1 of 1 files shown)


fregante:
Example of what full-width pages look like before this PR

PR #3334, opened 2020-07-10
Fix links in `tags-dropdown`

### LINKED ISSUE #3331: `tags-dropdown` doesn't point to tags pages, but to /tree/
1. https://github.com/sindresorhus/refined-github/releases
2. Open the tags dropdown on the top right
3. Nice that the links point to /tree/ pages instead of /releases/tag

This is not working because now tags are lazy-loaded:

https://github.com/sindresorhus/refined-github/blob/120de1b0ebaba019d9775705907c652bbdd770a1/source/features/tags-dropdown.tsx#L20

dertieran:
Fixes #3331

It seems the layout structure changed, so I fixed the selector.
There is also an `id="tags-menu"` on the `[role="tabpanel"]:last-child` element so we could also use that instead.

### DIFF
--- a/source/features/tags-dropdown.tsx
+++ b/source/features/tags-dropdown.tsx
@@ -15,7 +15,7 @@ function changeTabToTags(): void {
 
 function updateLinksToTag(): void {
 	// Change links, which point to the content of each tag, to open the tag page instead
-	for (const anchorElement of select.all<HTMLAnchorElement>('.rgh-tags-dropdown .SelectMenu-list:last-child [href*="/tree/"]')) {
+	for (const anchorElement of select.all<HTMLAnchorElement>('.rgh-tags-dropdown [role="tabpanel"]:last-child [href*="/tree/"]')) {
 		const pathnameParts = anchorElement.pathname.split('/');
 		pathnameParts[3] = 'releases/tag'; // Replace `tree`
 		anchorElement.pathname = pathnameParts.join('/');
--- a/source/features/tags-dropdown.tsx
+++ b/source/features/tags-dropdown.tsx
@@ -15,7 +15,7 @@ function changeTabToTags(): void {
 
 function updateLinksToTag(): void {
 	// Change links, which point to the content of each tag, to open the tag page instead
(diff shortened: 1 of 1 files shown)


fregante:
I can’t check right now but, are we sure that the list has already loaded when this code runs? Maybe it also needs a MutationObserver

dertieran:
It worked for me, but with all the changes in Github lately might be good to confirm this.

But I think the lazy loading was already handled through [`remote-input-success`](https://github.com/github/remote-input-element#events)
https://github.com/sindresorhus/refined-github/blob/75deae9d56cb5ecf7d3045c25b8b6131f2871fc0/source/features/tags-dropdown.tsx#L47-L50

source/features/tags-dropdown.tsx
```diff
@@ -15,7 +15,7 @@ function changeTabToTags(): void {
 
 function updateLinksToTag(): void {
 	// Change links, which point to the content of each tag, to open the tag page instead
-	for (const anchorElement of select.all<HTMLAnchorElement>('.rgh-tags-dropdown .SelectMenu-list:last-child [href*="/tree/"]')) {
+	for (const anchorElement of select.all<HTMLAnchorElement>('.rgh-tags-dropdown [role="tabpanel"]:last-child [href*="/tree/"]')) {
```
fregante, line 18: The problem was just that `SelectMenu-list` isn't the last child anymore
```suggestion
	for (const anchorElement of select.all<HTMLAnchorElement>('.rgh-tags-dropdown .SelectMenu-list [href*="/tree/"]')) {
```

However we can take this opportunity to make the selector more specific to exclude the branches that also appear in the DOM
```suggestion
	for (const anchorElement of select.all<HTMLAnchorElement>('.rgh-tags-dropdown #tags-menu .SelectMenu-item[href*="/tree/"]')) {
```

dertieran: Sure that also works 👍 
I thought the `:last-child` selector was meant to only select the tags tab. 😄

PR #3346, opened 2020-07-14
Refresh `more-dropdown` feature

### LINKED ISSUE #3215: `more-dropdown` compatibility with "Repository refresh" design

fregante:
Fixes #3215 

<details>
<summary>First implementation</summary>

I'm not entirely sure that this is the right approach to this and the code is hacky and repetitive. 

Thoughts?

</details>

Edit: this might be better and less repetitive:

## Next

I noticed a problem with `releases-tab` not always disappearing: https://github.com/sindresorhus/refined-github/issues/3347

This will have to be added separately:

- [ ] Also move Security and Insights tabs

### DIFF
--- a/source/features/more-dropdown.css
+++ b/source/features/more-dropdown.css
@@ -10,3 +10,7 @@
 .rgh-reponav-more svg {
 	width: 16px;
 }
+
+:root .rgh-has-more-dropdown {
+	overflow: visible !important;
+}
--- a/source/features/more-dropdown.css
+++ b/source/features/more-dropdown.css
@@ -11,10 +11,12 @@
 	width: 16px;
 }
 
+/* Always show the overflow menu button */
 .js-responsive-underlinenav-overflow {
 	visibility: visible !important;
 }
 
+/* Only show the divider if there are other items above it */
 .js-responsive-underlinenav-overflow [hidden] + .dropdown-divider {
 	display: none;
 }
--- a/source/features/more-dropdown.css
+++ b/source/features/more-dropdown.css
@@ -11,6 +11,10 @@
 	width: 16px;
 }
 
-:root .rgh-has-more-dropdown {
-	overflow: visible !important;
+.js-responsive-underlinenav-overflow {
+	visibility: visible !important;
+}
+
+.js-responsive-underlinenav-overflow [hidden] + .dropdown-divider {
+	display: none;
 }
--- a/source/features/more-dropdown.tsx
+++ b/source/features/more-dropdown.tsx
@@ -8,7 +8,6 @@ import DiffIcon from 'octicon/diff.svg';
 import BranchIcon from 'octicon/git-branch.svg';
 import HistoryIcon from 'octicon/history.svg';
 import PackageIcon from 'octicon/package.svg';
-import KebabHorizontalIcon from 'octicon/kebab-horizontal.svg';
 
 import features from '.';
 import {appendBefore} from '../helpers/dom-utils';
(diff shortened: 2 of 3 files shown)


source/features/releases-tab.tsx
```diff
@@ -64,14 +65,20 @@ async function init(): Promise<false | void> {
 		// "Repository refresh" layout
 
 		const releasesTab = (
-			<a href={`/${repoUrl}/releases`} className="js-selected-navigation-item UnderlineNav-item hx_underlinenav-item no-wrap js-responsive-underlinenav-item" data-hotkey="g r" data-selected-links="repo_releases" data-tab-item="releases-tab">
+			<a href={`/${repoUrl}/releases`} className="js-selected-navigation-item UnderlineNav-item hx_underlinenav-item no-wrap js-responsive-underlinenav-item" data-hotkey="g r" data-selected-links="repo_releases" data-tab-item="rgh-releases-item">
 				<TagIcon className="UnderlineNav-octicon"/>
 				<span data-content="Releases">Releases</span>
-				{count === undefined ? '' : <span className="Counter">{count}</span>}
+				{count ?? <span className="Counter">{count}</span>}
```
kidonng, line 71: I think this should be `&&`

source/features/more-dropdown.tsx
```diff
@@ -31,18 +31,45 @@ function createDropdown(): void {
...
 
-	const menu = select('.reponav-dropdown .dropdown-menu')!;
+	const nav = select('.js-responsive-underlinenav .UnderlineNav-body');
+	if (nav) {
+		// "Repository refresh" layout
+		nav.parentElement!.classList.add('rgh-has-more-dropdown');
+		select('.js-responsive-underlinenav-overflow ul')!.append(
+			<li className="dropdown-divider" role="separator"/>,
+			createDropdownItem('Compare', compareUrl),
+			pageDetect.isEnterprise() ? '' : createDropdownItem('Dependencies', dependenciesUrl),
```
kidonng, line 60: can this also use `&&`?

fregante: No because `.append` will stringify the Boolean, unlike dom-chef

source/features/more-dropdown.css
```diff
@@ -10,3 +10,11 @@
 .rgh-reponav-more svg {
 	width: 16px;
 }
+
+.js-responsive-underlinenav-overflow {
+	visibility: visible !important;
+}
+
+.js-responsive-underlinenav-overflow [hidden] + .dropdown-divider {
```
fregante, line 18:
```suggestion
/* Only show the divider if there are other items above it */
.js-responsive-underlinenav-overflow [hidden] + .dropdown-divider {
```

source/features/more-dropdown.css
```diff
@@ -10,3 +10,11 @@
 .rgh-reponav-more svg {
 	width: 16px;
 }
+
+.js-responsive-underlinenav-overflow {
```
fregante, line 14:
```suggestion
/* Always show the overflow menu button */
.js-responsive-underlinenav-overflow {
```

PR #3353, opened 2020-07-16
Fix `latest-tag-button`

kidonng:
Fix https://github.com/sindresorhus/refined-github/issues/3277#issuecomment-655173202

This is a placeholder, once the feature goes into production (likely) we can fix it (or make other changes).

Screenshot with the fix:
image

### DIFF
--- a/source/features/latest-tag-button.tsx
+++ b/source/features/latest-tag-button.tsx
@@ -84,7 +84,7 @@ async function init(): Promise<false | void> {
 		return false;
 	}
 
-	const breadcrumb = await elementReady('.breadcrumb');
+	const breadcrumb = await elementReady('#branch-select-menu');
 	if (!breadcrumb) {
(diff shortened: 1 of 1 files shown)


fregante:
I don't think we need to wait for that. The branch selector already exists on both versions.

I `blame`d the line and there doesn't seem to be a specific reason why it used `.breadcrumb` in the first place

kidonng:
Well, let's remove `btn-sm` at least before merging? 😂

image

kidonng:
> I `blame`d the line and there doesn't seem to be a specific reason why it used `.breadcrumb` in the first place

Possibly because of this

https://github.com/sindresorhus/refined-github/blob/79a8166955f56ca9185ec9f8db2d56da0d2b304f/source/features/list-prs-for-file.tsx#L133

PR #1049, opened 2018-01-29
Move release link from readme to branch selector

bfred-it:
Addresses https://github.com/sindresorhus/refined-github/pull/395#issuecomment-301186914

Related: https://github.com/sindresorhus/refined-github/issues/1029

cc @hkdobrev who suggested the change

screen shot

PR #3356, opened 2020-07-16
Adjust `latest-tag-button`'s size

kidonng:
From https://github.com/sindresorhus/refined-github/pull/3353#issuecomment-659260733

Before: 

After:

### DIFF
--- a/source/features/latest-tag-button.tsx
+++ b/source/features/latest-tag-button.tsx
@@ -97,7 +97,7 @@ async function init(): Promise<false | void> {
 	});
 
 	const link = (
-		<a className="btn btn-sm btn-outline ml-2 flex-self-center" href={String(url)}>
+		<a className="btn btn-outline ml-2 flex-self-center" href={String(url)}>
 			<TagIcon/>
 		</a>
 	);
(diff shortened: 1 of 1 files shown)


fregante:
The smaller size was on purpose

kidonng:
This class exists (or more precisely, is added) in the feature's first commit (https://github.com/sindresorhus/refined-github/pull/1049/files#diff-041e2c7f28122df0efbf5dbb6ee3c2adR22), in which I don't see any reason except matching GitHub's style.

fregante:
https://github.com/sindresorhus/refined-github/pull/3221 by @FloEdelmann, in the very first post

kidonng:
Missed that one. But that PR is made when design update is in progress, do you still think it looks fine with different button size?

image
image

image
image

fregante:
> do you still think it looks fine with different button size?

I do; you can see how the buttons also become wider and the space on that line might be limited already

kidonng:
I will override it with the following style:

```css
/* https://github.com/sindresorhus/refined-github/pull/3356 */
#branch-select-menu ~ .btn,
#branch-select-menu + .BtnGroup .btn,
#branch-select-menu ~ .flex-self-center .btn {
  padding: 5px 16px;
  font-size: 14px;
}

#branch-select-menu ~ .btn .octicon,
#branch-select-menu + .BtnGroup .octicon,
#branch-select-menu ~ .flex-self-center .octicon {
  vertical-align: text-bottom;
}
```

PR #3361, opened 2020-07-17
Restore `bugs-tab`

### LINKED ISSUE #3360: Bug Tab no longer working
Did not have time to de"bug" it yet

yakov116:
Fixes #3360

### DIFF
--- a/source/features/bugs-tab.tsx
+++ b/source/features/bugs-tab.tsx
@@ -39,10 +39,7 @@ async function init(): Promise<void | false> {
 		return false;
 	}
 
-	const issuesTab = (await elementReady([
-		'.pagehead [data-hotkey="g i"]',
-		'.js-repo-nav [data-hotkey="g i"]' // "Repository refresh" layout
-	].join()))?.parentElement;
+	const issuesTab = (await elementReady('.js-repo-nav [data-hotkey="g i"]'))?.parentElement;
 	if (!issuesTab) {
 		// Repo is archived
 		return false;
--- a/source/features/bugs-tab.tsx
+++ b/source/features/bugs-tab.tsx
@@ -39,7 +39,10 @@ async function init(): Promise<void | false> {
 		return false;
 	}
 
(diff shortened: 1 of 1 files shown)


fregante:
I think it used `.js-repo-nav` before too, so we can probably **replace** most `.pagehead` selectors with it.

yakov116:
I will try it now.

yakov116:
We used '[data-hotkey="g i"]' but now there are 2 of them. so we need to specify which one. 

I dont know how the old layout was

fregante:
> We used '[data-hotkey="g i"]' but now there are 2 of them. so we need to specify which one.

Right, but `.js-repo-nav` **is** the right one; the first of those link is in the page header

source/features/bugs-tab.tsx
```diff
@@ -39,7 +39,10 @@ async function init(): Promise<void | false> {
 		return false;
 	}
 
-	const issuesTab = (await elementReady('.pagehead [data-hotkey="g i"]'))?.parentElement;
+	const issuesTab = (await elementReady([
+		'.pagehead [data-hotkey="g i"]',
+		'.js-repo-nav [data-hotkey="g i"]' // "Repository refresh" layout
+	].join()))?.parentElement;
```
fregante, line 45: Even if it doesn't work on GHE, it's fine
```suggestion
	const issuesTab = (await elementReady('.js-repo-nav [data-hotkey="g i"]'))?.parentElement;
```

fregante: If you feel like it, you can drop `.parentElement` here
```suggestion
	const issuesLink = await elementReady('.js-repo-nav [data-hotkey="g i"]'));
```

And move it to:

```ts
	const bugsTab = issuesTab.parentElement!.cloneNode(true);
```

busches: This exists in GHE with the new selector. If there are specific concerns, @ me and I can check.

fregante:
🙌

yakov116:
My 100th commit to refined github!!

PR #3362, opened 2020-07-17
Fix features relying on `.pagehead` 

### LINKED ISSUE #3367: `ci-link`: Cannot read property 'append' of null
`ci-link` is not visible on any page where it's supposed to be.

Screenshot 2020-07-18 20 35 50

Also getting a bunch of other errors along with this error:

Screenshot 2020-07-18 20 36 05

yakov116:
Selector Update

Fixes #3367

### DIFF
--- a/source/features/ci-link.tsx
+++ b/source/features/ci-link.tsx
@@ -25,7 +25,7 @@ async function init(): Promise<false | void> {
 	}
 
 	// Append to title (aware of forks and private repos)
-	appendBefore('.pagehead h1, .hide-full-screen h1', '.fork-flag', icon);
+	appendBefore('.pagehead h1, .hide-full-screen h1', '.fork-flag, .hide-full-screen span.text-small', icon);
 }
 
 void features.add({
--- a/source/features/ci-link.tsx
+++ b/source/features/ci-link.tsx
@@ -1,11 +1,11 @@
 import './ci-link.css';
+import select from 'select-dom';
 import oneTime from 'onetime';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
 import fetchDom from '../helpers/fetch-dom';
 import {getRepoURL} from '../github-helpers';
-import {appendBefore} from '../helpers/dom-utils';
 
 // Look for the CI icon in the latest 2 days of commits #2990
 const getIcon = oneTime(fetchDom.bind(null,
@@ -25,7 +25,7 @@ async function init(): Promise<false | void> {
 	}
 
 	// Append to title (aware of forks and private repos)
-	appendBefore('.pagehead h1, .hide-full-screen h1', '.fork-flag, .hide-full-screen span.text-small', icon);
+	select('[itemprop="name"]')!.after(icon);
 }
 
 void features.add({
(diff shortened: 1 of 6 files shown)


fregante:
Wait, did they remove the pagehead class? If so, it should be replaced in a bunch of features

yakov116:
@fregante  yes and it's not so simple

kidonng:
> Wait, did they remove the pagehead class? If so, it should be replaced in a bunch of features

`.gisthead` and `.pagehead` still exists on Gits; `.pagehead` and `.repohead` is gone on main site.

fregante:
That class is not unique enough, it’s a utility class

yakov116:
I dont see anything else 😞. Its all generic classes

yakov116:
This is not going to be fun. There are many more features that are broken.

yakov116:
I fixed some more, I did not add any comments (yet). I really dont see anything we can work with

source/features/ci-link.tsx
```diff
@@ -25,7 +25,7 @@ async function init(): Promise<false | void> {
 	}
 
 	// Append to title (aware of forks and private repos)
-	appendBefore('.pagehead h1', '.fork-flag', icon);
+	appendBefore('.pagehead h1, .hide-full-screen h1', '.fork-flag, .hide-full-screen span.text-small', icon);
```
fregante, line 28: The DOM completely changed, `.fork-flat` is not inside `h1`.

You can probably just use `select([itemprop="name"])!.after(icon)`

source/features/releases-tab.tsx
```diff
@@ -59,7 +59,7 @@ async function init(): Promise<false | void> {
 		return false;
 	}
 
-	await elementReady('.pagehead + *'); // Wait for the tab bar to be loaded
+	await elementReady('.pagehead + *, .js-repo-nav + *'); // Wait for the tab bar to be loaded
```
fregante, line 62: Nothing exists after `.js-repo-nav`, but this should work
```suggestion
	await elementReady('.pagehead + *, .UnderlineNav-body + *'); // Wait for the tab bar to be loaded
```

source/features/show-open-prs-of-forks.tsx
```diff
@@ -67,7 +67,7 @@ async function initHeadHint(): Promise<void | false> {
 		return false;
 	}
 
-	select('.fork-flag .text')!.append(
+	select('.fork-flag .text, .hide-full-screen span.text-small')!.append(
```
fregante, line 70: Same as before I think:

```suggestion
	select<HTMLAnchorElement>(`[data-hovercard-type="repository"][href="/${getForkSourceRepo()}"]`)!.append(
```

fregante:
If supporting the old version gets messy, let's just drop it. GHE users will have to upgrade eventually. Such is life.

source/features/ci-link.tsx
```diff
@@ -25,7 +25,7 @@ async function init(): Promise<false | void> {
 	}
 
 	// Append to title (aware of forks and private repos)
-	appendBefore('.pagehead h1', '.fork-flag', icon);
+	select('[itemprop="name"]')!.after(icon);
```
fregante, line 28: This works, but it's slightly misaligned

Screen Shot 2020-07-17 at 19 13 03

PR #3369, opened 2020-07-19
Fix permalink detection in 3 features

fregante:
They removed the text we were using:
Screen Shot 2020-07-19 at 12 15 46 ➡️ Screen Shot 2020-07-19 at 12 15 36

Causing

```
❌ Refined GitHub → edit-files-faster → TypeError: The element wasn’t found, the selector needs to be updated.
```
```
❌ Refined GitHub → edit-readme → TypeError: The element wasn’t found, the selector needs to be updated.
```

cc @yakov116 I think you mentioned this first

# Test

https://github.com/sindresorhus/refined-github/tree/20.7.16 (`edit-readme` and `edit-files-faster`)

```
`comments-time-machine-links` bar at the top: https://github.com/sindresorhus/refined-github/blob/master/source/features/comments-time-machine-links.tsx?rgh-link-date=2019-12-11T10%3A12%3A11Z
```

### DIFF
--- a/source/github-helpers/index.ts
+++ b/source/github-helpers/index.ts
@@ -2,6 +2,7 @@ import select from 'select-dom';
 import oneTime from 'onetime';
 import compareVersions from 'tiny-version-compare';
 import * as pageDetect from 'github-url-detection/esm/index.js'; // eslint-disable-line import/extensions -- Required for Node tests compatibility
+import elementReady from 'element-ready';
 
 // This never changes, so it can be cached here
 export const getUsername = oneTime(pageDetect.utils.getUsername);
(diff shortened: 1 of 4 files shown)


source/github-helpers/index.ts
```diff
@@ -110,3 +111,21 @@ export function preventPrCommitLinkLoss(url: string, pr: string, commit: string,
...
+
+/** Is tag or commit */
+export function isPermalink(): boolean {
+	return (
+		// Pre "Latest commit design updates"
+		/Tag|Tree/.test(select('[data-hotkey="w"] i')?.textContent!) || // Text appears in the branch selector
+
+		// "Latest commit design updates"
+		select.exists('[data-hotkey="w"] .octicon-tag') || // Tags have an icon
+		!select.exists('[data-menu-item="code-tab"] [href*="/tree/"]') // Commits don't change the "Code" tab address
```
fregante, line 123: Doesn't work on the default branch

PR #3377, opened 2020-07-21
Fix `keyboard-navigation` issues

### LINKED ISSUE #3319: `keyboard-navigation` issues
## Hidden conversations in Files tabs are "focused"

1. Visit https://github.com/sindresorhus/refined-github/pull/3242/files
2. Press the `i` key
3. Press `j` 3 times 

You'll see that the focus sometimes is _lost_ because it's focusing the hidden comments

## PR reviews are skipped

[Messageless reviews](https://github.com/sindresorhus/refined-github/pull/3242#pullrequestreview-441470147) don't have a focus ring so I asked to exclude reviews, but I didn't realize that this also excludes [reviews with a message](https://github.com/sindresorhus/refined-github/pull/3242#pullrequestreview-442052008)

Is it possible to focus the latter without focusing the former?

## Others

I might have forgotten other issues. Please write them below

dertieran:
Fixes #3319 

Instead of making complex query selectors this approach just filters out elements that shouldn't be *targetable*.

I mainly tested this code with use cases described in the [issue](https://github.com/sindresorhus/refined-github/issues/3319#issue-651357203), so if there are any other edge case I didn't consider please let me know.

> Hidden conversations in Files tabs are "focused"

For this now any `.js-minimizable-comment-group` is ignored if if it is *minimized*.
I tested this with in [#3242 (files)](https://github.com/sindresorhus/refined-github/pull/3242/files), now only the files are targeted.
For me the comments also where hidden *(and not visible when targeted)* without pressing `i` first, so I handled both scenarios.

> PR reviews are skipped

Targetable `#pullrequestreview-{id}` elements are now only *targetable* if they contain a message.
PR reviews with a message also contain a targetable element with `#pullrequestreview-{id}-body-html`.

So now pressing `j` on [this issue comment](https://github.com/sindresorhus/refined-github/pull/3242#issuecomment-652709634) skips the two messageless review comments and goes to the next [review with a message](https://github.com/sindresorhus/refined-github/pull/3242#pullrequestreview-442052008) below.

---

I noticed some problems with the `.filter()` approach because we might filtered out the current targeted element.
So e.g. if a [messageless review](https://github.com/sindresorhus/refined-github/pull/3242#pullrequestreview-441470147) is targeted, pressing `j` will jump to the first *targetable* element.
A better approach might be to target the next/previous *targetable* element instead.

Happy for any feedback on this, or if there are any other issues/use cases where this code doesn't work 😄

### DIFF
--- a/source/features/keyboard-navigation.tsx
+++ b/source/features/keyboard-navigation.tsx
@@ -4,6 +4,22 @@ import * as pageDetect from 'github-url-detection';
 import features from '.';
 import {isEditable} from '../helpers/dom-utils';
 
+function isCommentGroupMinimized(comment: HTMLElement): boolean {
+	if (select.exists('.minimized-comment:not(.d-none)', comment)) {
+		return true;
+	}
+
+	if (comment.closest('.js-resolvable-thread-contents.d-none')) {
+		return true;
+	}
+
+	if (comment.closest('.js-resolvable-timeline-thread-container:not([open])')) {
+		return true;
+	}
+
+	return false;
+}
+
 function runShortcuts(event: KeyboardEvent): void {
 	if (isEditable(event.target)) {
 		return;
@@ -15,7 +31,14 @@ function runShortcuts(event: KeyboardEvent): void {
 		event.preventDefault();
 
 		const items = select.all('.js-targetable-element:not([id^="pullrequestreview"]')
-			.filter(comment => !comment.querySelector('.minimized-comment:not(.d-none)'));
+			.filter(element => {
+				if (element.classList.contains('js-minimizable-comment-group')) {
+					return !isCommentGroupMinimized(element);
+				}
+
+				return true;
+			});
+
 		// `j` goes to the next comment, `k` goes back a comment
 		const direction = event.key === 'j' ? 1 : -1;
 
--- a/source/features/keyboard-navigation.tsx
+++ b/source/features/keyboard-navigation.tsx
@@ -29,8 +29,9 @@ function runShortcuts(event: KeyboardEvent): void {
 				'.js-minimizable-comment-group'
 			])
 			.filter(element =>
-				!element.classList.contains('js-minimizable-comment-group') ||
-				!isCommentGroupMinimized(element)
+				element.classList.contains('js-minimizable-comment-group') ?
+					!isCommentGroupMinimized(element) :
+					true
 			);
 
 		// `j` goes to the next comment, `k` goes back a comment
--- a/source/features/keyboard-navigation.tsx
+++ b/source/features/keyboard-navigation.tsx
@@ -4,29 +4,12 @@ import * as pageDetect from 'github-url-detection';
 import features from '.';
 import {isEditable} from '../helpers/dom-utils';
 
-function isCommentGroupMinimized(comment: HTMLElement): boolean {
-	if (select.exists('.minimized-comment:not(.d-none)', comment)) {
-		return true;
-	}
-
-	if (comment.closest('.js-resolvable-thread-contents.d-none')) {
-		return true;
-	}
-
-	if (comment.closest('.js-resolvable-timeline-thread-container:not([open])')) {
-		return true;
-	}
-
-	return false;
-}
-
-function pullRequestReviewHasComment(review: HTMLElement): boolean {
-	if (review.id.endsWith('-body-html')) {
-		return false;
-	}
-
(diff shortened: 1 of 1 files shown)


fregante:
It looks like event are also being targeted: https://github.com/sindresorhus/refined-github/pull/3242#event-3515544984

I'll suggest a selector for that

source/features/keyboard-navigation.tsx
```diff
@@ -14,8 +38,19 @@ function runShortcuts(event: KeyboardEvent): void {
...
+				if (element.id.startsWith('pullrequestreview-')) {
+					return pullRequestReviewHasComment(element);
+				}
+
+				if (element.classList.contains('js-minimizable-comment-group')) {
+					return !isCommentGroupMinimized(element);
+				}
+
+				return true;
+			});
```
fregante, line 52: Maybe a longer selector is still faster than looping via JS
```suggestion
		const items = select.all([
			// Excludes reviews without a main comment
			'.js-targetable-element[id^="pullrequestreview"]:not(.js-comment)',
			
			// All targetable elements, except reviews and unnecessary events
			'.js-targetable-element:not([id^="pullrequestreview"]):not([id^="event-"])'
		]).filter(isCommentGroupMinimized);
```

fregante: But maybe at this point we should _select_ exactly what we want, rather than select everything and then _filter_ it. Example

```suggestion
		const items = select.all([
			// Reviews with a main comment
			'[id^="pullrequestreview"][id$="-body-html"]',
			
			// Files in diffs
			'[id^="#diff-"]',
			
			// Comments (to be `.filter()`ed)
			'.js-minimizable-comment-group',
			
			// PR comments?
			'etc',
		])
		.filter(comment => !comment.querySelector('.minimized-comment:not(.d-none)'));
```

This is easier, even if the filter doesn't really apply to most of them

source/features/keyboard-navigation.tsx
```diff
@@ -4,6 +4,30 @@ import * as pageDetect from 'github-url-detection';
...
+function isCommentGroupMinimized(comment: HTMLElement): boolean {
+	if (select.exists('.minimized-comment:not(.d-none)', comment)) {
+		return true;
+	}
+
+	if (comment.closest('.js-resolvable-thread-contents.d-none')) {
+		return true;
+	}
+
+	if (comment.closest('.js-resolvable-timeline-thread-container:not([open])')) {
```
kidonng, line 16: Merge into `comment.closest` above?

fregante: True, the function can look something like:

```js
return (
	exists('0') ||
	closest(`
		1,
		2
	`)
);
```

source/features/keyboard-navigation.tsx
dertieran: I like that approach it still wouldn't solve *losing* the current target if it isn't in `items`.
But I noticed there are a lot of possible targets that we don't select anyways, so it might not be the best idea anyways.

The only thing I changed was
1. `#pullrequestreview-<id>-body-html` also have `.js-minimizable-comment-group` so I combined that.
2.  `[id^="diff-"]` was to general and also selected [lines in diffs](https://github.com/sindresorhus/refined-github/pull/3242/files#diff-0730bb7c2e8f9ea2438b52e419dd86c9R179) so I added `.js-targetable-element`.

I couldn't find any `.js-minimizable-comment-group:not(.js-targetable-element)` so I kept the simpler selector.

dertieran:
I think I broke the test by force pushing mb :sweat_smile:

source/features/keyboard-navigation.tsx
```diff
@@ -38,18 +21,17 @@ function runShortcuts(event: KeyboardEvent): void {
...
-			});
+		const items = select
+			.all([
+				// Files in diffs
+				'.js-targetable-element[id^="diff-"]',
+				// Comments (to be `.filter()`ed)
+				'.js-minimizable-comment-group'
+			])
+			.filter(element =>
+				!element.classList.contains('js-minimizable-comment-group') ||
```
fregante: It can't select `.js-minimizable-comment-group` and then exclude it, it must be wrong.

Also, if necessary, such simple check should just be part of the selector: `.original-selector:not(.something-to-exclude)`

fregante:
Thank you! It appears to have fixed both issues

PR #3384, opened 2020-07-22
Fix `repo-age` where there is only one commit

yakov116:
Test on https://github.com/sindresorhus/css-in-readme-like-wat

Selector update


fregante:
Thanks!

PR #3396, opened 2020-07-24
Fix `comment-fields-keyboard-shortcuts`

### LINKED ISSUE #3388: Up arrow to edit own comment is broken
When trying to quickly edit your own PR comment with Up arrow, it now results in the following error:

```
refined-github.js:3466 Uncaught TypeError: Cannot read property 'click' of null
    at eventHandler (refined-github.js:3466)
    at HTMLDocument.Object.capture (refined-github.js:3446)
    at HTMLDocument.listenerFn (refined-github.js:568)
```

refined-github-edit-error

Happens on all PR urls where you're trying to amend a comment

kidonng:
Fix #3388

GitHub has some kind of lazy load logic for the edit comment button, ~~so we need to dispatch a `mouseover` event to make the button "appear" (though not visible, unless `edit-comments-faster` is also enabled).~~

~~The first trigger will be slow (about 2-3 seconds on my machine) but following triggers will be instant.~~

I think the same method can be used to make `edit-comments-faster` more reliable (or can it?)

### DIFF
--- a/source/features/comment-fields-keyboard-shortcuts.tsx
+++ b/source/features/comment-fields-keyboard-shortcuts.tsx
@@ -6,7 +6,7 @@ import * as pageDetect from 'github-url-detection';
 import features from '.';
 import onCommentFieldKeydown from '../github-events/on-comment-field-keydown';
 
-async function eventHandler(event: delegate.Event<KeyboardEvent, HTMLTextAreaElement>): Promise<void> {
+function eventHandler(event: delegate.Event<KeyboardEvent, HTMLTextAreaElement>): void {
 	const field = event.delegateTarget;
 
 	if (event.key === 'Escape') {
--- a/source/features/comment-fields-keyboard-shortcuts.tsx
+++ b/source/features/comment-fields-keyboard-shortcuts.tsx
@@ -4,8 +4,9 @@ import * as pageDetect from 'github-url-detection';
 
 import features from '.';
 import onCommentFieldKeydown from '../github-events/on-comment-field-keydown';
+import elementReady from 'element-ready';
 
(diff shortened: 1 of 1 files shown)


source/features/comment-fields-keyboard-shortcuts.tsx
```diff
@@ -39,7 +40,9 @@ function eventHandler(event: delegate.Event<KeyboardEvent, HTMLTextAreaElement>)
 			});
 
 		if (lastOwnComment) {
-			select<HTMLButtonElement>('.js-comment-edit-button', lastOwnComment)!.click();
+			// Make edit comment button appear
+			select('details[id]', lastOwnComment)!.dispatchEvent(new MouseEvent('mouseover'));
+			(await elementReady('.js-comment-edit-button', {target: lastOwnComment, stopOnDomReady: false}) as HTMLButtonElement).click();
```
fregante, line 45: No need for that specific button. GitHub will react to any click on `<button class="js-comment-edit-button"/>`
```suggestion
			// Make the comment editable (the native edit button might not be available yet)
			const editButton = <button class="js-comment-edit-button" hidden/>;
			lastOwnComment.append(editButton);
			editButton.click();
			editButton.remove();
```

fregante: The same way, you can create a whole new button for `edit-comments-faster`

fregante: Is this confirmed to work? I didn't directly test my code

kidonng: Your code works very well (instant repsonse) 👍

PR #3290, opened 2020-06-26
Apply `parse-backticks` to more pages

kidonng:
Add/update the following selectors in the new layout:

- Commit title and description on files: https://github.com/kidonng/readhub/blob/test/README.md
- Pinned Issues: https://github.com/sindresorhus/refined-github/issues
- Removed `.Box-row.js-navigation-item a.link-gray-dark`: https://github.com/sindresorhus/refined-github/issues & https://github.com/sindresorhus/refined-github/pulls
- Actions list: https://github.com/sindresorhus/refined-github/actions
- Commit title and description on repo homepage: https://github.com/kidonng/readhub/tree/test
- ~~Issue title: test on this page~~
- Existing PR title: https://github.com/sindresorhus/refined-github/compare/master...FloEdelmann:octicons-react?expand=1
- Linked issues: `https://github.com/sindresorhus/refined-github/issues/3234` & `https://github.com/sindresorhus/refined-github/pull/3290` (prevent referencing)
- Wiki: https://github.com/kidonng/readhub/wiki/%60backtick%60-tets
- Notifications: https://github.com/notifications
- [More](https://github.com/sindresorhus/refined-github/pull/3290#issuecomment-655505621)

PR #3397, opened 2020-07-24
Make `parse-backticks` more reliable

kidonng:
On request by https://github.com/sindresorhus/refined-github/pull/3290#issuecomment-659262450

Part of #3084

~~Edited the title and it stil runs well!!! 😋 (though there's about 2 seconds' delay)~~ (Never mind, that is `format-conversation-titles`)

### DIFF
--- a/source/features/parse-backticks.tsx
+++ b/source/features/parse-backticks.tsx
@@ -4,17 +4,8 @@ import {observe} from 'selector-observer';
 import features from '.';
 import {parseBackticks} from '../github-helpers/dom-formatters';
 
-function parse(selectors: string[]): void {
-	observe(selectors.map(selector => selector + ':not(.rgh-backticks-already-parsed)').join(), {
-		add(element) {
-			element.classList.add('rgh-backticks-already-parsed');
-			parseBackticks(element);
-		}
-	});
-}
-
 function init(): void {
-	parse([
+	const selectors = [
 		'.BorderGrid--spacious .f4.mt-3', // `isRepoHome` repository description
 		'.js-commits-list-item .mb-1, .js-commits-list-item pre', // `isCommitList` commit message and description
 		'.Details[data-issue-and-pr-hovercards-enabled] .d-none a.link-gray-dark', // `isRepoRoot` commit message
@@ -41,7 +32,14 @@ function init(): void {
 		'.profile-timeline-card .text-gray-dark', // `isUserProfileMainTab` issue and PR title
 		'#user-repositories-list [itemprop="description"]', // `isUserProfileRepoTab` repository description
 		'.js-hovercard-content > .Popover-message .link-gray-dark' // Hovercard
-	]);
+	].map(selector => selector + ':not(.rgh-backticks-already-parsed)').join();
+
+	observe(selectors, {
+		add(element) {
+			element.classList.add('rgh-backticks-already-parsed');
+			parseBackticks(element);
+		}
+	});
 }
 
 void features.add({
--- a/source/features/parse-backticks.tsx
+++ b/source/features/parse-backticks.tsx
@@ -6,13 +6,12 @@ import features from '.';
 import {parseBackticks} from '../github-helpers/dom-formatters';
 
 function parse(selectors: string[]): void {
-	for (const selector of selectors) {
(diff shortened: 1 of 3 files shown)


source/features/parse-backticks.tsx
```diff
@@ -1,15 +1,17 @@
...
 import features from '.';
-import observeElement from '../helpers/simplified-element-observer';
 import {parseBackticks} from '../github-helpers/dom-formatters';
 
 function parse(selectors: string[]): void {
-	for (const element of select.all(selectors.map(selector => selector + ':not(.rgh-backticks-already-parsed)'))) {
-		element.classList.add('rgh-backticks-already-parsed');
-		parseBackticks(element);
+	for (const selector of selectors) {
+		observe(selector, {
```
fregante, line 10:
```suggestion
	observe(selectors.join(','), {
```

I'm pretty sure `rgh-backticks-already-parsed` is still needed; last time I tried that was exactly the issue. Try for example:
1. load this PR
2. visit another PR tab (to trigger AJAX navigation)
3. click the browser's back button

You might see that the feature run twice on the same piece of DOM

fregante: Also you can merge all the `init`s into one so there's a single `observe` for the whole feature

kidonng: > 3. click the browser's back button

Usually there need to be a 4th step to trigger that: then click the forward button. But whatever.

kidonng:
We have to test the selectors and narrow scope down since they are now global

kidonng:
- Grouped/merged related selectors
- Replace some generic selectors with more specific selectors
- Add detailed comments

source/features/parse-backticks.tsx
```diff
@@ -1,113 +1,54 @@
...
-	]);
-}
-
-function initUserProfile(): void {
-	parse([
+		'.commits blockquote', // `isDashboard` newsfeed commits
+		'.notifications-list-item p.text-normal', // `isNotifications` issue and PR title
 		'.profile-timeline-card .text-gray-dark', // `isUserProfileMainTab` issue and PR title
-		'[itemprop="description"]' // `isUserProfileRepoTab` repository description
+		'[itemprop="description"]', // `isUserProfileRepoTab` repository description
```
fregante, line 42: Indeed this is probably too broad now

yakov116:
Not sure if you fixed this but 

image

kidonng:
Yeah it's parsed

image

Isn't the selector for that place is already there before this PR?

yakov116:
Funny it was not working for me before

source/features/parse-backticks.tsx
```diff
@@ -1,113 +1,54 @@
...
+		add(element) {
+			element.classList.add('rgh-backticks-already-parsed');
+			parseBackticks(element);
+		}
+	});
 }
 
-function initRepo(): void {
+function init(): void {
 	parse([
```
fregante, line 17: You can avoid the extra `parse` function as well, just merge it into `init`

yakov116:
Should we make a line break in between each unique page type? (Ex: `isWiki` `isDashboard`) I think it will make it a drop easier to read.

fregante:
I don't think there's any way to make it any more readable. It's a list of wildly different selectors, it's just what it is

fregante:
Looks great! Possible additions:

- use `selector-observer` in https://github.com/sindresorhus/refined-github/blob/master/source/features/linkify-code.tsx (this would fix https://github.com/sindresorhus/refined-github/issues/2091)
- remove parse backticks from https://github.com/sindresorhus/refined-github/blob/master/source/features/format-conversation-titles.tsx
- either rename `format-conversation-titles` back to `linkify-titles` or merge it into `linkify-code`

yakov116:
This is awesome!! https://github.com/tanmayrajani/notifications-preview-github is now parsed too!!

kidonng:
Is it https://github.com/sindresorhus/refined-github/pull/3397/files#diff-7b0a2c8650978a85cebddd08a46e40baR31? I don't have the extension

yakov116:
Yes.

You should get it, its an awesome extension.

PR #3406, opened 2020-07-27
Restore `ci-link`

yakov116:
Selector update


kidonng:
Is it supposed to appear before the "Private" label?

image

yakov116:
@kidonng I don't know. I don't remember how it was before

kidonng:
https://github.com/sindresorhus/refined-github/blob/6842f86e937502476997206d0003e17d703dd513/source/features/ci-link.tsx#L29

So the location hasn't changed, it just lacks an appropriate margin.

PR #3405, opened 2020-07-27
Add `cleanup-repo-filelist-actions` feature

kidonng:
Resolves #3403

Current screenshot:

image

<details>
<summary>Prototypes</summary>

image

image

</details>

I don't know how to name and describe this feature so suggestions welcome.

I feel like the dropdown caret on the "Add" button should not be dropped. And there's this download directory button added by us:

image

Wonder whether should take the text away.

PR #2992, opened 2020-04-11
Improve feature application on AJAXed widgets

fregante:
# Tests

1. Visit an issue or PR (like this one)
2. Add a reaction

Or:

1. Visit an issue or PR (like this one)
2. Edit a comment

You should still see the features being applied correctly:

- `sticky-discussion-sidebar`
- `clean-sidebar`
- `reaction-avatars`
- `edit-comments-faster`’s icon (fixes https://github.com/sindresorhus/refined-github/issues/2572)
- `mark-unread`’s button

Follows https://github.com/sindresorhus/refined-github/pull/2533 https://github.com/sindresorhus/refined-github/pull/2344

PR #3407, opened 2020-07-27
Restore `edit-comments-faster` on review comments

### LINKED ISSUE #3279: "Edit comment" button injection is not reliable
This is in Brave 1.10.93 Chromium: 83.0.4103.106 (Official Build) (64-bit).
All shields are turned off (gray icon).

Oftentimes when I edit the comment, the button is gone after submitting:

image

After a full page refresh it is back again:

image

I can't reproduce this bug in Chrome 83.0.4103.116. When I submit the edit there, the icon is hidden for a split second, but then is always back.

In both browsers I see the console error during initial load ([example page](https://github.com/ua-books/ua-books/issues/33)):

```
❌ Refined GitHub → more-dropdown → TypeError: Cannot read property 'append' of null
    at appendBefore (refined-github.js:3157)
    at createDropdown (refined-github.js:3171)
    at more_dropdown_init (refined-github.js:3184)
    at async runFeature (refined-github.js:2837)
    at async setupPageLoad (refined-github.js:2845)
```

however, I doubt it's related, because it does not appear during the submitting.

kidonng:
Maybe a continuation of #2992

<details><summary>Obsolete information</summary>

## Explanation of the lazy load logic mentioned in #3396

```
.js-minimizable-comment-group
	.minimized-comment // Invisible
		...
			details-menu
				.js-comment-edit-button // A
    .js-comment
		...
			details-menu[src]
				// Contents are lazy loaded when you hover the three dots menu
				.js-comment-edit-button // B
```

1. Why #3388 happens: `comment-fields-keyboard-shortcuts` tries to find B which isn't loaded yet
2. Why `edit-comments-faster` in review comments doesn't work: only A triggers the function and the icon is added into an invisible container
3. Why https://github.com/sindresorhus/refined-github/issues/3388#issuecomment-662886260 happens: B loads and triggers the function and the icon is added into an invisible container

</details>

## This PR

- ~~Determine if it's A triggering the function, and **add the edit icon to the correct location in review comments** (fix 2)~~
- ~~Skip if it's B triggering the function, to **prevent duplicate icon addition** (fix 3)~~
- Upgrade to `selector-observer`, fix #3279
- Rewrite the feature to fix existing issues

Test: A PR with review comments (like `https://github.com/sindresorhus/refined-github/pull/3405#discussion_r460863841`)


source/features/edit-comments-faster.tsx
```diff
@@ -14,7 +14,16 @@ function init(): void {
 		const button = item.cloneNode();
 		button.append(<PencilIcon/>);
 		button.classList.replace('dropdown-item', 'timeline-comment-action');
-		item.closest('details')!.before(button);
+
+		// Prevent duplicate icon in review comments
+		if (item.closest('details-menu[src]')) {
+			return false;
```
fregante, line 20: If you return false, you interrupt the loop. Use `continue`

fregante:
This seems unnecessary. As I suggested earlier, the solution is to create a brand-new button instead of cloning the existing one. This way you just need to select the right place to place the button and avoid all this document traversing/filtering.

kidonng:
Filtering is needed because there *can be* two edit buttons in a single review comment.

Check out the latest code, maybe the simplest one.

source/features/edit-comments-faster.tsx
fregante: test

fregante:
Please always provide a link where this is testable, as requested in the PR template.

kidonng:
Sorry, added test link and make what this PR does more clear.

kidonng:
Seems like GitHub removed the aforementioned lazy load logic, so this bug doesn't exist anymore. Closing for the moment.

kidonng:
After using `selector-observer` it works for unresolved conversations 🎉

fregante:
I'm still seeing the lazy-loaded context-menus. For the 3rd time, do not use `cloneNode`. Just generate the link via JSX.

kidonng:
🤷‍♀️ That `cloneNode` is not written by me

source/features/edit-comments-faster.tsx
```diff
@@ -1,21 +1,22 @@
...
-
-		const button = item.cloneNode();
-		button.append(<PencilIcon/>);
-		button.classList.replace('dropdown-item', 'timeline-comment-action');
-		item.closest('details')!.before(button);
-	}
+			const button = item.cloneNode();
+			button.append(<PencilIcon/>);
+			button.classList.replace('dropdown-item', 'timeline-comment-action');
+			item.closest('.js-minimizable-comment-group')!.querySelector('.js-comment-header-reaction-button')!.after(button);
```
fregante, line 17: Indeed. But the feature could now be shorter, if you want to try this:

```js
observe('js-comment-header-reaction-button:not(.rgh-edit-comment)', {
	add(reactions) {
		reactions.after(
			<button className="timeline-comment-action" etc etc>
				<PencilIcon/>
			</button>
		);
	}
});

Comment (kidonng) [reply source/features/edit-comments-faster.tsx:17]: > ```js
> reactions.classList.add('rgh-edit-comment');
> ```

Is this needed? No code use it and I don't think it needs to customized

Comment (kidonng) [reply source/features/edit-comments-faster.tsx]: Instinctive and works well 👍

File: source/features/edit-comments-faster.tsx [hunk 3]
```diff
@@ -26,5 +28,6 @@ void features.add({
 	include: [
 		pageDetect.hasComments
 	],
-	init
+	waitForDomReady: false,
```
fregante, line 31: Can be dropped because it can't be avoided
```suggestion
```

fregante (approved):
Thank you!

Note: `hasComment` automatically [implies a comment event listener though](https://github.com/sindresorhus/refined-github/blob/f527259f797d18964483d2e065d13d4dfbe36200/source/features/index.tsx#L156). This has no effect since we use `onetime`, but we should keep it in mind.

Perhaps we can remove that implication once we refactor [every/most `hasComments` feature](https://github.com/sindresorhus/refined-github/search?q=hasComments&type=Code) to use `selector-observer`

source/features/edit-comments-faster.tsx
```diff
@@ -1,21 +1,23 @@
...
+import onetime from 'onetime';
+import {observe} from 'selector-observer';
 import PencilIcon from 'octicon/pencil.svg';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
 
 function init(): void {
-	const menuItems = select.all('details .js-comment-edit-button:not(.rgh-edit-comment)');
+	observe('.current-user .js-comment-header-reaction-button:not(.rgh-edit-comment)', {
```
fregante, line 10: > ```
> .current-user
> ```

This is a substantial change that I didn't notice. It might make sense, but also requires a change of the description.

I think it's still worth keeping it for all the comments for consistency though

fregante: This needs to be undone first

kidonng: > I think it's still worth keeping it for all the comments for consistency though

I get what you mean know 🤦‍♂️ I chose this class by mistake because I was testing in this repo

kidonng:
I reverted [`9c147cd`](https://github.com/sindresorhus/refined-github/pull/3407/commits/9c147cd0c56b8e3e1393741a1856d9d8b79ed45d) with new selectors

source/features/edit-comments-faster.tsx
```diff
@@ -1,21 +1,23 @@
...
 
-	for (const item of menuItems) {
-		item.classList.add('rgh-edit-comment');
-
-		const button = item.cloneNode();
-		button.append(<PencilIcon/>);
-		button.classList.replace('dropdown-item', 'timeline-comment-action');
-		item.closest('details')!.before(button);
-	}
+			comment.closest('.js-comment')?.querySelector('.js-comment-header-reaction-button')?.after(
```
fregante, line 14: Why are you selecting js-comment-update and then navigating to this one? Just select this element in the observer.

kidonng: Does `this one` refer to `.js-comment-header-reaction-button`? I failed to find a selector meeting the following requirements:

- Comments in review (`.review-comment`)
- Made by others (not `.current-user`)
- Can be edited (the key issue, comment's classes didn't tell whether you can edit or not)

So I reverted how this PR was initially made ([`3e5fb20`](https://github.com/sindresorhus/refined-github/pull/3407/commits/3e5fb20d78b7965cd71db4c03fcf06a21626e159))

fregante: > * Made by others (not `.current-user`)

Why?

> * Can be edited (the key issue, there is no difference in editable and not-editable review comments' classes)

It can be edited if:

- it's the current user, or
- the current user can edit the sidebar, for example.

> So I reverted how this PR was initially made ([`3e5fb20`](https://github.com/sindresorhus/refined-github/pull/3407/commits/3e5fb20d78b7965cd71db4c03fcf06a21626e159))

Let's not go reverting things, this can definitely be done, including selector-observer

fregante: > * Comments in review (`.review-comment`)

That's completely irrelevant. Every comment can have this button.

fregante: Those restrictions are not present in your current code except `can be edited`.

The code is fine except `?.` and the lack of explanation why the DOM traversal is needed. I fixed/added both.

fregante (approved):

fregante:
3 months in the making

# 🥳

source/features/edit-comments-faster.tsx
kidonng: Though the PR has been merged, I'm answering these questions here:

> > * Comments in review (`.review-comment`)
> 
> That's completely irrelevant. Every type of comment can be edited.

For comments in the timeline (`.timeline-comment`) comments have `.editable-comment`, but this is not the case for comments in reviews.

> > * Made by others (not `.current-user`)
> 
> Why?

Comments made by users themselves have `.current-user`, but now I remember this doesn't always mean it can be edited.

> > * Can be edited (the key issue, there is no difference in editable and not-editable review comments' classes)
> 
> It can be edited if:
> 
> * it's the current user, or
> * the current user is a maintainer (you can determine this by checking whether the user can edit the sidebar, for example)

That's what this PR needs 👍 I limited my thoughts to the comments.

yakov116:
Not sure what broke here. After this PR on https://github.com/sindresorhus/refined-github/pull/3621 only the first comment has a edit icon. the rest throw a whole bunch of console errors.

fregante:
Hidden comments break the feature because they have no reaction button. The `observe` selector needs to be updated

PR #3409, opened 2020-07-28
Apply `linkify-code` to unfolded diffs

kidonng:
Part of #3084 (if I should mention it)
Resolves #2091

Test: https://github.com/sindresorhus/refined-github/pull/636/files (unfold the diff above)

This PR simply warp the old code with `selector-observer` and add an additional selector (`.blob-expanded`) to make it work on unfolded diffs.

~~Merge `format-conversation-titles` in this PR as well (on request by https://github.com/sindresorhus/refined-github/pull/3397#issuecomment-664580685)~~

PR #3410, opened 2020-07-28
Move backtick parsing in `format-conversation-titles`  test

kidonng:
Note `format-conversation-titles` will be removed in #3409


fregante:
This doesn't seem to work like the rest of ajaxed GH pieces. The title has its specific update function that just changes the `textContent`, making which is invisible to our new observer
```js
u(n, '.js-issue-title').textContent = o.issue_title;
```

fregante:
`format-conversation-titles` should probably be excluded from https://github.com/sindresorhus/refined-github/pull/3409 as well.

If anything, `format-conversation-titles` should be fixed because the title isn't re-parsed when the title is changed

kidonng:
#3458

Issue #3414, opened 2020-07-30
Support editor indent/deintent in new-issue editor


dwelle:
The improvements to textarea (around indenting at least) added to replies editor does not extend to the editor when creating a new issue.

In other words, I can't use `tab`/`Shift+tab` in editor while creating new issues.

Was this intended or an oversight?

fregante:
It was mistakenly limited by https://github.com/sindresorhus/refined-github/pull/3256

That code just need an extra line now: `pageDetect.isNewIssue` I think

yakov116:
@fregante I will fix it

PR #3420, opened 2020-07-31
Preserve current branch in `forked-to` link when possible

### LINKED ISSUE #3415: `forked-to` feature always redirects to `HEAD` branch on the fork
The enhancement that was made to `forked-to` that makes the forked-to link go to the currently open file is really useful, but it's a bit limited - no matter on what branch the file I'm browsing is currently on, the fork link will go to that file on `HEAD` commit. This has two issues:
- to edit file, you need to switch to an actual branch as `HEAD` is just using HEAD commit and you can't use edit button from there
- this feature only works properly for default branch; otherwise the link will go to location that doesn't exist

So to illustrate with an example, it would be great if clicking the fork button on:
`https://github.com/Cog-Creators/Red-DiscordBot/blob/V3/develop/redbot/__main__.py`
would make me go to:
`https://github.com/jack1142/Red-DiscordBot/blob/V3/develop/redbot/__main__.py`
rather than:
`https://github.com/Cog-Creators/Red-DiscordBot/blob/HEAD/redbot/__main__.py`

Similarly, going from non-default branch:
`https://github.com/Cog-Creators/Red-DiscordBot/blob/develop/red.py`
should go to:
`https://github.com/jack1142/Red-DiscordBot/blob/develop/red.py`
rather than non-existent:
`https://github.com/jack1142/Red-DiscordBot/blob/HEAD/red.py`

About the second situation, I'm not sure what should happen if the branch doesn't exist on fork (or the file you're looking at isn't in it), but imo just showing 404 isn't all that bad in such cases as they're probably not going to be as common.

### LINKED ISSUE #3416: Make `forked-to` link to edit page when you're on edit page
Basically, I would want the `forked-to` feature to link to edit page when I'm already on edit page, i.e:
When I'm on:
`https://github.com/Cog-Creators/Red-DiscordBot/edit/V3/develop/redbot/logging.py`
Clicking the button should make it go to:
`https://github.com/jack1142/Red-DiscordBot/edit/V3/develop/redbot/logging.py`
Rather than the "home page" of repository.

yakov116:
1. LINKED ISSUES:
	Resolves #3416
	Resolves #3415

2. TEST URLS:
Have a fork of this repo
   https://github.com/sindresorhus/refined-github/edit/master/.editorconfig

3. SCREENSHOT:
  	None

### DIFF
--- a/source/features/forked-to.tsx
+++ b/source/features/forked-to.tsx
@@ -31,16 +31,16 @@ const updateCache = cache.function(async (): Promise<string[] | undefined> => {
 });
 
 function createLink(baseRepo: string): string {
-	if (pageDetect.isRepoRoot() || !(pageDetect.isSingleFile() || pageDetect.isRepoTree() || pageDetect.isEditingFile())) {
-		return '/' + baseRepo;
+	if (pageDetect.isSingleFile() || pageDetect.isRepoTree() || pageDetect.isEditingFile()) {
+		const [user, repository] = baseRepo.split('/', 2);
+		const url = new GitHubURL(location.href).assign({
+			user,
+			repository
+		});
+		return url.pathname;
 	}
 
-	const [user, repository] = baseRepo.split('/', 2);
-	const url = new GitHubURL(location.href).assign({
-		user,
-		repository
-	});
-	return url.pathname;
+	return '/' + baseRepo;
 }
 
 async function updateUI(forks: string[]): Promise<void> {
--- a/source/features/forked-to.tsx
+++ b/source/features/forked-to.tsx
@@ -84,9 +84,7 @@ function forkDropdown(fork: string): HTMLAnchorElement {
 		</a>
 	) as unknown as HTMLAnchorElement;
 
-	if (fork === getRepoURL()) {
-		link.href = location.href;
-	} else {
+	if (fork !== getRepoURL()) {
 		setURL(link, fork);
 	}
 
--- a/source/features/forked-to.tsx
+++ b/source/features/forked-to.tsx
@@ -56,7 +56,6 @@ function createLink(baseRepo: string): string {
 	});
 
 	if (pageDetect.isEditingFile()) {
-		url.route = 'blob'; // Replace the 'edit'
 		void updateForkLinkToDefaultBranch(url); // Don't await it, since the link will work without the update
 	}
 
--- a/source/features/forked-to.tsx
+++ b/source/features/forked-to.tsx
@@ -31,6 +31,10 @@ const updateCache = cache.function(async (): Promise<string[] | undefined> => {
 });
 
 function createLink(baseRepo: string): string {
+	if (pageDetect.isRepoRoot() || !(pageDetect.isSingleFile() || pageDetect.isRepoTree() || pageDetect.isEditingFile())) {
+		return '/' + baseRepo;
+	}
+
 	const [user, repository] = baseRepo.split('/', 2);
 	const url = new GitHubURL(location.href).assign({
 		user,
--- a/source/features/forked-to.tsx
+++ b/source/features/forked-to.tsx
@@ -32,13 +32,13 @@ const updateCache = cache.function(async (): Promise<string[] | undefined> => {
 	staleWhileRevalidate: 5
 });
 
-async function updateForkBranch(url: GitHubURL, fork: HTMLAnchorElement): Promise<void> {
+async function updateForkBranch(url: GitHubURL, link: HTMLAnchorElement): Promise<void> {
 	const currentBranch = getCurrentBranch();
 	const existsOnDefaultBranch = doesFileExist(url);
 	const existsOnCurrentBranch = doesFileExist(url.assign({branch: currentBranch}));
 
 	if (await existsOnCurrentBranch) {
-		fork.pathname = url.pathname;
+		link.pathname = url.pathname;
 		return;
 	}
 
@@ -47,16 +47,21 @@ async function updateForkBranch(url: GitHubURL, fork: HTMLAnchorElement): Promis
 		url.assign({
 			branch: defaultBranch
 		});
-		fork.pathname = url.pathname;
+		link.pathname = url.pathname;
 		return;
 	}
 
-	fork.pathname = `/${url.user}/${url.repository}`;
+	link.pathname = `/${url.user}/${url.repository}`;
 }
 
-function setURL(fork: HTMLAnchorElement, baseRepo: string): void {
+function setURL(link: HTMLAnchorElement, baseRepo: string): void {
+	if (baseRepo === getRepoURL()) {
+		link.href = location.href;
+		return;
+	}
+
 	if (pageDetect.isRepoRoot() || !(pageDetect.isSingleFile() || pageDetect.isRepoTree() || pageDetect.isEditingFile())) {
-		fork.href = '/' + baseRepo;
+		link.href = '/' + baseRepo;
 		return;
 	}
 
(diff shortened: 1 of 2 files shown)


yakov116:
I am confused between #3415 and #3414 are we using default branch or are we using the current branch?

source/features/forked-to.tsx
```diff
@@ -42,6 +55,11 @@ function createLink(baseRepo: string): string {
 		branch: 'HEAD'
 	});
 
+	if (pageDetect.isEditingFile()) {
+		url.route = 'blob'; // Replace the 'edit'
```
fregante, line 59: Why?

source/features/forked-to.tsx
```diff
@@ -30,8 +32,19 @@ const updateCache = cache.function(async (): Promise<string[] | undefined> => {
...
 });
 
+async function updateForkLinkToDefaultBranch(url: GitHubURL): Promise<void> {
+	if (await doesFileExist(url)) {
+		const defaultBranch = await getDefaultBranch(getRepositoryInfo(url.pathname.slice(1)));
+
+		select<HTMLAnchorElement>(`.pagehead-actions [href*="${url.pathname}"]`)!.pathname = url.assign({
+			route: 'edit',
+			branch: defaultBranch
+		}).pathname;
```
fregante, line 42: `createLink` is called multiple times in a `map` below, which means that there are multiple links. Use `select.all`

source/features/forked-to.tsx
```diff
@@ -30,8 +32,19 @@ const updateCache = cache.function(async (): Promise<string[] | undefined> => {
 	staleWhileRevalidate: 5
 });
 
+async function updateForkLinkToDefaultBranch(url: GitHubURL): Promise<void> {
+	if (await doesFileExist(url)) {
+		const defaultBranch = await getDefaultBranch(getRepositoryInfo(url.pathname.slice(1)));
+
+		select<HTMLAnchorElement>(`.pagehead-actions [href*="${url.pathname}"]`)!.pathname = url.assign({
+			route: 'edit',
```
fregante, line 40: This is not restricted to `isEditingFile` at all. It applies to every type of link

source/features/forked-to.tsx
yakov116: Can't check if the file exists with a route of edit??

yakov116:
@fregante can you see my comment about and clear me what we want to do? I am super confused

source/features/forked-to.tsx
yakov116: The way it's written now it gets called for each link separately.

source/features/forked-to.tsx
fregante: `doesFileExist` doesn't even use `url.route`, what are you talking about? 
https://github.com/sindresorhus/refined-github/blob/6f4ac06f08b5f2aea453dc4be5c50824b7cda2af/source/github-helpers/does-file-exist.tsx

yakov116: It takes the full url. If the route is Edit it will fail. 

I am not sure what you are not understanding.

yakov116: I remember why I did it. If the file does not exist the link will be a 404. Since the route is `edit`

source/features/forked-to.tsx
fregante: You're right. However this needs to be made more readable though:
```suggestion
		url.assign({
			branch: defaultBranch
		});
		select<HTMLAnchorElement>(`.pagehead-actions [href*="${url.pathname}"]`)!.pathname = url.pathname;
```

But really this way of changing the URL is complete spaghetti. `createLink` is a getter, it shouldn't also trigger an HTTP request and query the DOM independently. This also applies to the previous similar PR, but it made it a little awkward there with the JSX.

A more reasonable way would be something like:

```js
const link = (
	<a
		className="btn btn-sm float-left rgh-forked-button"
		title={`Open your fork at ${forks[0]}`}
	>
		<LinkExternalIcon/>
	</a>
);
forkCounter.before(link);
setUrl(link);
```

`setUrl` would then first set the `HEAD` URL and then update it if necessary.

fregante: If you have any p/react experience this might be a good place to use Preact instead of `dom-chef`, it would _probably_ make these kind of updates more sensible.

source/features/forked-to.tsx
fregante: If the file doesn't exist it will be a 404 regardless of the route

source/features/forked-to.tsx
yakov116: I have none

fregante: Then the `setUrl` suggestion still applies

fregante:
I think you're also trying to support https://github.com/sindresorhus/refined-github/issues/3415, right? Not just #3416

fregante:
To answer https://github.com/sindresorhus/refined-github/pull/3420#issuecomment-667110407

#3416 is unrelated to branches. It’s just about enabling it on isEditingFile
#3415 means that it should point to the current branch if it exists, otherwise point to the default branch.

yakov116:
@fregante I fully test my code (I have 2 forks of RH).

2 Things.

1. Am I on the right track here? (I think the code can be improved, for now I made it work)
2. how do you want to handle editing until the network request finishes. Sometimes github takes a second or 2 to respond. If the url route is `edit` and the branch is still `HEAD` it will land on a 404. Should I change it to blob and later change it back to edit?

source/features/forked-to.tsx
```diff
@@ -30,9 +32,32 @@ const updateCache = cache.function(async (): Promise<string[] | undefined> => {
 	staleWhileRevalidate: 5
 });
 
-function createLink(baseRepo: string): string {
-	if (pageDetect.isRepoRoot() || !(pageDetect.isSingleFile() || pageDetect.isRepoTree())) {
```
fregante, line 34: Add `pageDetect.isEditingFile`  here

source/features/forked-to.tsx
```diff
@@ -42,7 +67,28 @@ function createLink(baseRepo: string): string {
 		branch: 'HEAD'
```
fregante, line 67: Remove this line here
```suggestion
```

fregante: Or, actually, even better! This means that we can skip a lot of code.

I'll push something for you to have a look

fregante:
Toughts?

yakov116:
Very Good!

I am not in from of my computer, I will test it later.

yakov116:
The only issue is if it does not exist on the default branch it will lead to a 404 :(

fregante:
> The only issue is if it does not exist on the default branch it will lead to a 404 :(

It's a risk we can run, instead of having to do 2 HTTP requests for every page visited on forks. Once on the 404 the user can still open the repo.

yakov116:
> Before, this wasn't possible because `isRepoTree` includes `isRepoRoot` and setting the branch would break that page, but now we're not setting the branch, so we can't skip it.

But if your on `isRepoRoot` <repo>/<branch>/undefined/undefined

Because of

https://github.com/sindresorhus/refined-github/blob/a918f9a3eb6827dd67efdd1bcf362df2e7152406/source/github-helpers/github-url.ts#L70

The last 2 return undefined.

Should we add a `.replace('/undefined','')` ?

fregante:
> [this.user, this.repository, this.route, this.branch, this.filePath].filter(Boolean).join('/')

This looks good, but I'm afraid that it would make it hard to debug a missing `route`, because it would just become `user/repo/develop` instead of `user/repo/undefined/develop`

This is inelegant but it works correctly `.replace(/(undefined[/])+$/g, '')`

yakov116:
> This looks good, but I'm afraid that it would make it hard to debug a missing `route`, because it would just become `user/repo/develop` instead of `user/repo/undefined/develop`

Me too.

Or we can just go back to the "less readable version" code

yakov116:
> This is inelegant but it works correctly `.replace(/(undefined[/])+$/g, '')`

Make the change here or in GithubURL?

fregante:
In GithubURL

yakov116:
I did not check your code but That other one is/was for a trailing slash.

yakov116:
@fregante I need your feedback on how you want to handle `isEditingFile`

fregante:
> @fregante I need your feedback on how you want to handle `isEditingFile`

What about it?

PR #800, opened 2017-11-06
Add build status and link to CI next to the repo's title

bfred-it:
CI build status

multiple CI status

Notes:

* disregard tooltip position, it'll be removed by #799 
* needs mention in the readme once accepted

Testable pages:

* https://github.com/sindresorhus/refined-github (Travis only)
* https://github.com/Microsoft/vscode (multiple)
* https://github.com/bfred-it-obsolete/empty (no ci)

Fixes #726

PR #3429, opened 2020-08-03
Append `ci-link` icon after private label

kidonng:
Test URL: a private repo

Before: 
image

After:
image

<details>
<summary>No longer apply <s>The extra margin also pairs nicely with other extensions like Octotree.</s>
</summary>

Before:
image

After:
image

Before:
image

After:
image

</details>

### DIFF
--- a/source/features/ci-link.css
+++ b/source/features/ci-link.css
@@ -5,3 +5,7 @@
 .rgh-ci-link summary .octicon {
 	vertical-align: 0;
 }
+
+.Label + .rgh-ci-link {
+	margin-left: 8px;
+}
--- a/source/features/ci-link.css
+++ b/source/features/ci-link.css
@@ -6,6 +6,7 @@
 	vertical-align: 0;
 }
 
+/* Adds space between `ci-link` and "Private" label, if present */
 .Label + .rgh-ci-link {
 	margin-left: 8px;
 }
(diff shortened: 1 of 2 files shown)


fregante:
The icon should actually go at the end of the line so it doesn't move anything to the right when it's added.

https://github.com/sindresorhus/refined-github/pull/800#issuecomment-342894687

Can you move it after the label?

kidonng:
I asked the same thing in https://github.com/sindresorhus/refined-github/pull/3406#issuecomment-666265758, I don't remember where it was before so I just kept the location. Of course would like to move it if it is placed at the wrong place.

source/features/ci-link.tsx
```diff
@@ -27,7 +27,11 @@ async function init(): Promise<false | void> {
 	}
 
 	// Append to title (aware of forks and private repos)
-	select('[itemprop="name"]')!.after(icon);
+	const privateLabel = select('[itemprop="name"] + .Label');
+	(privateLabel ?? select('[itemprop="name"]'))!.after(icon);
+	if (privateLabel) {
+		icon.classList.add('ml-2');
+	}
```
fregante, line 34: Can you move the spacing to the CSS? It would be easier to specify it there and it also simplifies this code to:

```suggestion
	select('[itemprop="name"]')!.parentElement!.append(icon);
```

source/features/ci-link.css
```diff
@@ -5,3 +5,7 @@
 .rgh-ci-link summary .octicon {
 	vertical-align: 0;
 }
+
```
fregante, line 8:
```suggestion

// Adds space between `ci-link` and "Private" label, if present
```

PR #1558, opened 2018-10-03
Make 404 pages more useful

bfred-it:
Examples:

- 404 file: https://github.com/sindresorhus/refined-github/blob/master/wrong-file

	

- 410 file: https://github.com/sindresorhus/refined-github/blob/master/extension/content.js

	

- 404 branch + file: https://github.com/sindresorhus/refined-github/blob/404/wrong-file

	

- 410 branch + file: https://github.com/sindresorhus/refined-github/blob/404/package.json

	

Closes #485
Replaces, includes and closes #1531
Depends on and includes #1557

PR #3430, opened 2020-08-03
Fixes and improvements to `useful-not-found-page`

fregante:
GitHub seems to have finally started automatically redirecting URLs to old branches:

Screen Shot 2020-08-03 at 19 24 49

Demo:

- 404 branch, file: https://github.com/sindresorhus/refined-github/blob/404/package.json
- 404 branch, folder: https://github.com/sindresorhus/refined-github/blob/404/source

So I'm removing this behavior from `blob` and `tree`

---

However, they don't apply this behavior to `edit` URLs, so I'm leaving the logic just for that:

- 404 branch, **edit** file: https://github.com/sindresorhus/refined-github/edit/404/package.json

---

Bugfixes:

- `/user/repo/edit/` URLs should not be queried in the loop:
	https://github.com/sindresorhus/refined-github/edit/master/wrong-file
	

---

Original feature tests in https://github.com/sindresorhus/refined-github/pull/1558

### DIFF
--- a/source/features/useful-not-found-page.tsx
+++ b/source/features/useful-not-found-page.tsx
@@ -68,7 +68,7 @@ async function addDefaultBranchLink(bar: Element): Promise<void> {
 
 	bar.after(
 		<p className="container mt-4 text-center">
-			See also the file on the <a href={url}>default branch</a>
+			See also the object on the <a href={url}>default branch</a>
 		</p>
 	);
 }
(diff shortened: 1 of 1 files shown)


fregante:
@yakov116 how come this feature was not switched over to `GitHubURL`? Did it not work?

yakov116:
CurrentBranch does not exist

fregante:
Does that make it crash altogether?

PR #3412, opened 2020-07-29
Add `linkify-user-edit-history-popup` feature

yakov116:
1. LINKED ISSUES:
   Closes #2161

2. TEST URLS:
	https://github.com/Mottie/GitHub-userscripts/issues/88#issuecomment-502933879 (with bot and should not be linkified)
	https://github.com/sindresorhus/refined-github/pull/1766#issue-251741293
	This pull request

3. SCREENSHOT:
image

PR #3442, opened 2020-08-08
Replace `repeatOnAjax` with explicit `onetime`

fregante:
Waiting for https://github.com/sindresorhus/onetime/issues/12

Can be reviewed. Doesn't need testing but it's worth a look just in case features _need_ that old specific behavior (unlikely)


contributing.md
```diff
@@ -68,11 +68,8 @@ features.add({
 	/** Whether to wait for DOM ready before runnin `init`. `false` makes `init` run right as soon as `body` is found. @default true */
 	waitForDomReady: false,
 
-	/** Whether to re-run `init` on pages loaded via AJAX. @default true */
-	repeatOnAjax: false,
```
yakov116, line 72: Want to give an example of using oneTime here?

fregante: Thought about it, but it's not necessary

sindresorhus:
What's the benefit of this change?

fregante:
Sorry, I didn't add context:
https://github.com/sindresorhus/refined-github/pull/3314#discussion_r453770549
https://github.com/sindresorhus/refined-github/pull/3412#pullrequestreview-457547068

In short, pages can change via ajax, but if `repeatOnAjax === false` RG won't call their `init` even if now their `include` conditions would match. 

That logic was never correct; our real intent was only to prevent duplicate execution on the same document, hence the `onetime`

sindresorhus:
Makes sense. I would probably have added a `initOnce` option or something, since it's used so much.

Issue #3340, opened 2020-07-12
Repository Creation date is incorrect.


JesterOrNot:
This project was created a few months ago (I made the repo and I'm not even 19)

URL: https://github.com/JesterOrNot/MathExpression

![image](https://user-images.githubusercontent.com/50386709/87236755-89b29a80-c3b2-11ea-9405-e343a75325c1.png)

yakov116:
Dont know what happened there but I get 2 months

![image](https://user-images.githubusercontent.com/16872793/87239231-c7c3b480-c3da-11ea-9c1c-24aa524a1307.png)

fregante:
<img width="354" alt="Screen Shot 2020-07-12 at 15 13 28" src="https://user-images.githubusercontent.com/1402241/87247143-427be680-c452-11ea-9238-9dadc7aa277d.png">


Also can't replicate. Firefox

fregante:
@yakov116 how do you think we should handle this? Why is the item shown even if the API throws an error?

yakov116:
What I **think** is happening they had a very old cache version and when it re-ran without an api key, it went haywire. does that make any sence.

I can try it out later tonight and see if that happens.

I tried it without a token and the feature never runs.

yakov116:
@fregante confirmed. I changed it to fetch dom. The issue is that we removed the `shouldRevalidate`. Any one the just updated their refined-github (they did not use it in a while) will have this issue.

![image](https://user-images.githubusercontent.com/16872793/88246007-299bde00-cc67-11ea-8e13-81771a244b11.png)


Add the `should revalidate` back? or just ignore?

fregante:
No, we're good, thanks for checking. We merged that update 43 days ago and the cache was set to expire after 30 days anyway, not many will see this unless they just updated RGH.

fregante:
The timestamp reported in the cache of https://github.com/sindresorhus/refined-github/issues/3394 is actually in the past (June 18th) but `cache.function` still returns the expired value.
 
Here's what's happening:

1. Expired items are not automatically cleared daily (https://github.com/fregante/webext-storage-cache/issues/7), but if you call `cache.get` you will receive `undefined` and the item will be deleted (✅ )
2. `cache.function` doesn't use this "smart" `cache.get` because the intent is to immediately write a new value if the past one is expired. This skips an unnecessary deletion (✅)
[...]

PR #3445, opened 2020-08-10
Only use the default branch to determine `repo-age`

yakov116:
Follows https://github.com/sindresorhus/refined-github/issues/3340#issuecomment-671049624

### DIFF
--- a/source/features/repo-age.tsx
+++ b/source/features/repo-age.tsx
@@ -16,27 +16,6 @@ const dateFormatter = new Intl.DateTimeFormat('en-US', {
 	day: 'numeric'
 });
 
-const getCommitCount = async (): Promise<AnyObject> => {
-	const {repository} = await api.v4(`
-		repository(${getRepoGQL()}) {
-			defaultBranchRef {
-				target {
-					... on Commit {
-						oid
-						committedDate
-						resourcePath
-						history {
-							totalCount
-						}
-					}
-				}
-			}
-		}
-	`);
-
-	return repository.defaultBranchRef.target;
-};
-
 const getRepoAge = async (commitSha: string, commitsCount: number): Promise<[string, string]> => {
 	const {repository} = await api.v4(`
 		repository(${getRepoGQL()}) {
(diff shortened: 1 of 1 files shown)


source/features/repo-age.tsx
```diff
@@ -17,6 +16,27 @@ const dateFormatter = new Intl.DateTimeFormat('en-US', {
 	day: 'numeric'
 });
 
+const getCommitCount = async (): Promise<AnyObject> => {
```
fregante, line 19: No need to extract this into its own function though, especially considering the AnyObject typing. Just merge it back into getFirstCommit

yakov116: Besides that everything else is ok?

yakov116: I remember why I split it.  I was having issues with constants they both are returning the same names.

fregante: I don't know what you're talking about

fregante:
Thanks for this fix!

source/features/repo-age.tsx
yakov116: That's what happens when I reply half in my sleep. Sorry I misunderstood you I thought you wanted both api requests in one function.

Anyways I'm up now ☕

source/features/repo-age.tsx
```diff
@@ -43,34 +42,28 @@ const getRepoAge = async (commitSha: string, commitsCount: number): Promise<[str
...
-	const commitUrl = commitInfo instanceof HTMLAnchorElement ? commitInfo.href : commitInfo!.src;
-	const commitSha = commitUrl.split('/').pop()!;
-
-	// In "Repository refresh" layout, the number of commits may not be rendered yet
-	const commitsCountElement = select('li.commits .num') ?? (await elementReady('.Box-header--blue [aria-label^="Commits on "]'))!.parentElement;
-	const commitsCount = looseParseInt(commitsCountElement!.textContent!);
-
-	// Returning undefined will make sure that it is not cached. It will check again for commits on the next load.
-	// Reference: https://github.com/fregante/webext-storage-cache/#getter
-	if (commitsCount === 0) {
```
fregante, line 65: I just realized that this was removed. It shouldn't be

yakov116: Since until now the element could have not been ready on time so it would be undefined. Now the only case would be on a brand new repo.

yakov116: https://github.com/sindresorhus/refined-github/pull/2783/files/f976d2c0fe70d52b9adfecec5a349526d6b6a2b5#diff-9d82f672e2199b6b2b455fa80d3373f2

fregante: > Now the only case would be on a brand new repo.

Yes. Brand new repos still exist and now I think this would throw in those cases because `repository.defaultBranchRef.target` is null

fregante: The easiest way is to add `isEmptyRepoRoot` to `excludes`

TODO lint <-- search this next week :D

yakov116: Lol

PR #3451, opened 2020-08-10
Meta: Update devDependencies

fregante:
Comments below


package.json
```diff
@@ -73,41 +73,41 @@
...
+		"jsdom": "^16.4.0",
 		"mini-css-extract-plugin": "^0.9.0",
 		"npm-run-all": "^4.1.5",
-		"rollup": "^2.19.0",
+		"rollup": "^2.23.1",
 		"size-plugin": "^2.0.1",
 		"string-replace-loader": "^2.3.0",
 		"stylelint": "^13.6.1",
 		"stylelint-config-xo": "^0.19.0",
 		"terser-webpack-plugin": "^3.0.6",
```
fregante, line 103: I'm not updating this one yet because in the next version (https://github.com/webpack-contrib/terser-webpack-plugin/commit/100e38e0db035f6eb21bbec2bbc1f8b12aee9f60) they start affecting `copy-webpack-plugin` files too, which means that our neat [webextensions.min.js](https://www.unpkg.com/browse/webextension-polyfill@0.6.0/dist/browser-polyfill.min.js) is stripped of license information and unnecessarily "beautified"

I initially extracted this file because:

- I want `browser` to be a global
- I thought AMO [preferred](https://extensionworkshop.com/documentation/publish/third-party-library-usage/) unmodified libraries
- It's minified by default!
- It can be used by all entry points without being bundled in each one of them

There are 2 possible solutions to this:

- find/add a way for `terser-webpack-plugin` to exclude this file
- have webpack create a vendor file

I tried the second one in https://github.com/sindresorhus/refined-github/pull/3157 already but that failed

package.json
```diff
@@ -73,41 +73,41 @@
...
-		"ts-loader": "^7.0.5",
+		"ts-loader": "^8.0.2",
 		"ts-node": "^8.10.2",
 		"type-fest": "^0.16.0",
-		"typescript": "^3.9.6",
-		"web-ext": "^4.3.0",
-		"web-ext-submit": "^4.3.0",
-		"webpack": "^4.43.0",
+		"typescript": "^4.1.0-dev.20200810",
```
fregante, line 107: On my computer, since https://github.com/sindresorhus/refined-github/pull/3157 I started getting incorrect "file was updated" logs by TypeScript every few seconds.

This version seems to fix it

package.json
```diff
@@ -73,41 +73,41 @@
...
 		"size-plugin": "^2.0.1",
 		"string-replace-loader": "^2.3.0",
 		"stylelint": "^13.6.1",
 		"stylelint-config-xo": "^0.19.0",
 		"terser-webpack-plugin": "^3.0.6",
-		"ts-loader": "^7.0.5",
+		"ts-loader": "^8.0.2",
 		"ts-node": "^8.10.2",
 		"type-fest": "^0.16.0",
-		"typescript": "^3.9.6",
```
fregante, line 107: TypeScript 4 enables labeled Tuple Types, we can apply these to existing RG code
https://devblogs.microsoft.com/typescript/announcing-typescript-4-0-beta/#labeled-tuple-elements

cc @yakov116

yakov116: Remind me next week 👍

package.json
fregante: May be possible with https://github.com/webpack-contrib/terser-webpack-plugin#exclude

fregante: Fixed!

package.json
yakov116: @fregante can you give me an example? I don't know typescript that well and i am having a hard time understanding how this works.

yakov116: https://github.com/Microsoft/TypeScript/issues/28259#issuecomment-452912594 Great explanation. 🚀

PR #3454, opened 2020-08-12
Add `remove-label-faster` feature

### LINKED ISSUE #3440: Easy remove of labels
I would like to propose a simple enhancement that would make removing labels from Issues very easy.

When would you hover over label simple crosshair would appear at the edge of the label which opon clicking on it, it would remove the label from the Issue. This would be restricted only to those who have rights to change labels in the repo.

Currently, you have to open the labels popup, find the right issue and then remove it. I find this flow very cumbersome while managing my Issues.

Happy to write PR for it if it would be accepted.

artusm:
1. LINKED ISSUES:
Closes #3440 

2. TEST URLS:
[sindresorhus/refined-github/issues/3440](https://github.com/sindresorhus/refined-github/issues/3440)
[sindresorhus/refined-github/pull/3454](https://github.com/sindresorhus/refined-github/pull/3454)

3. SCREENSHOT:
screenshoot

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -222,6 +222,7 @@ Thanks for contributing! 🦋🙌
 - [](# "pinned-issue-update-time") [Adds the updated time to pinned issues.](https://user-images.githubusercontent.com/1402241/75525936-bb524700-5a4b-11ea-9225-466bda58b7de.png)
 - [](# "clean-pinned-issues") [Changes the layout of pinned issues from side-by-side to a standard list.](https://user-images.githubusercontent.com/1402241/84509958-c82a3c00-acc4-11ea-8399-eaf06a59e9e4.png)
 - [](# "remove-label-faster") [Adds one-click buttons to remove labels in conversations.](https://user-images.githubusercontent.com/36174850/89980178-0bc80480-dc7a-11ea-8ded-9e25f5f13d1a.gif)
+- [](# "clean-conversation-headers") [Remove duplicate information in conversation headers.](https://user-images.githubusercontent.com/44045911/89736767-686ec800-da9e-11ea-81c3-252e9813140b.png)
 
 <!-- Refer to style guide above. Keep this message between sections. -->
 
--- a/readme.md
+++ b/readme.md
@@ -221,7 +221,7 @@ Thanks for contributing! 🦋🙌
 - [](# "bugs-tab") [Adds a "Bugs" tab to repos, if there are any open issues with the "bug" label.](https://user-images.githubusercontent.com/1402241/73720910-a688d900-4755-11ea-9c8d-70e5ddb3bfe5.png)
 - [](# "pinned-issue-update-time") [Adds the updated time to pinned issues.](https://user-images.githubusercontent.com/1402241/75525936-bb524700-5a4b-11ea-9225-466bda58b7de.png)
 - [](# "clean-pinned-issues") [Changes the layout of pinned issues from side-by-side to a standard list.](https://user-images.githubusercontent.com/1402241/84509958-c82a3c00-acc4-11ea-8399-eaf06a59e9e4.png)
-- [](# "remove-label-with-a-single-click") [Adds one-click buttons to remove labels in conversations.](https://user-images.githubusercontent.com/36174850/89980178-0bc80480-dc7a-11ea-8ded-9e25f5f13d1a.gif)
+- [](# "remove-label-button") [Adds one-click buttons to remove labels in conversations.](https://user-images.githubusercontent.com/36174850/89980178-0bc80480-dc7a-11ea-8ded-9e25f5f13d1a.gif)
 
 <!-- Refer to style guide above. Keep this message between sections. -->
 
--- a/readme.md
+++ b/readme.md
@@ -221,6 +221,7 @@ Thanks for contributing! 🦋🙌
 - [](# "bugs-tab") [Adds a "Bugs" tab to repos, if there are any open issues with the "bug" label.](https://user-images.githubusercontent.com/1402241/73720910-a688d900-4755-11ea-9c8d-70e5ddb3bfe5.png)
 - [](# "pinned-issue-update-time") [Adds the updated time to pinned issues.](https://user-images.githubusercontent.com/1402241/75525936-bb524700-5a4b-11ea-9225-466bda58b7de.png)
 - [](# "clean-pinned-issues") [Changes the layout of pinned issues from side-by-side to a standard list.](https://user-images.githubusercontent.com/1402241/84509958-c82a3c00-acc4-11ea-8399-eaf06a59e9e4.png)
+- [](# "remove-label-with-a-single-click") [Adds one-click buttons to remove labels in conversations.](https://user-images.githubusercontent.com/36174850/89980178-0bc80480-dc7a-11ea-8ded-9e25f5f13d1a.gif)
 
 <!-- Refer to style guide above. Keep this message between sections. -->
 
--- a/readme.md
+++ b/readme.md
@@ -221,7 +221,7 @@ Thanks for contributing! 🦋🙌
 - [](# "bugs-tab") [Adds a "Bugs" tab to repos, if there are any open issues with the "bug" label.](https://user-images.githubusercontent.com/1402241/73720910-a688d900-4755-11ea-9c8d-70e5ddb3bfe5.png)
 - [](# "pinned-issue-update-time") [Adds the updated time to pinned issues.](https://user-images.githubusercontent.com/1402241/75525936-bb524700-5a4b-11ea-9225-466bda58b7de.png)
 - [](# "clean-pinned-issues") [Changes the layout of pinned issues from side-by-side to a standard list.](https://user-images.githubusercontent.com/1402241/84509958-c82a3c00-acc4-11ea-8399-eaf06a59e9e4.png)
-- [](# "remove-label-button") [Adds one-click buttons to remove labels in conversations.](https://user-images.githubusercontent.com/36174850/89980178-0bc80480-dc7a-11ea-8ded-9e25f5f13d1a.gif)
+- [](# "remove-label-faster") [Adds one-click buttons to remove labels in conversations.](https://user-images.githubusercontent.com/36174850/89980178-0bc80480-dc7a-11ea-8ded-9e25f5f13d1a.gif)
 
 <!-- Refer to style guide above. Keep this message between sections. -->
 
--- a/source/features/remove-label-button.css
+++ b/source/features/remove-label-button.css
@@ -12,12 +12,12 @@
 	border-radius: 50%;
 	padding-top: 0.125rem;
 	padding-bottom: 0.125rem;
-	fill: currentColor;
 	width: 14px !important;
 	height: 14px !important;
 }
 
-.rgh-remove-label-button svg:hover {
+.rgh-remove-label-button:focus svg,
+.rgh-remove-label-button:hover svg {
 	background-color: var(--rgh-remove-label-bg);
 	color: var(--rgh-remove-label-color);
 }
--- a/source/features/remove-label-button.css
+++ b/source/features/remove-label-button.css
@@ -6,12 +6,10 @@
 }
 
 .rgh-remove-label-button svg {
-	margin-top: 2px;
-	margin-bottom: 2px;
-	margin-right: 2px;
+	margin: 2px;
+	margin-left: 0;
 	border-radius: 50%;
-	padding-top: 2px;
-	padding-bottom: 2px;
+	padding: 2px 0;
 	width: 14px !important;
 	height: 14px !important;
 }
--- a/source/features/remove-label-button.css
+++ b/source/features/remove-label-button.css
@@ -1,11 +1,11 @@
-.rgh-remove-label-with-a-single-click {
+.rgh-remove-label-button {
 	display: flex;
 	color: inherit;
 	margin-left: 2px;
 	margin-right: -7px;
 }
 
-.rgh-remove-label-with-a-single-click svg {
+.rgh-remove-label-button svg {
 	margin-top: 0.125rem;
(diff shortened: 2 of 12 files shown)


readme.md
```diff
@@ -221,6 +221,7 @@ Thanks for contributing! 🦋🙌
 - [](# "bugs-tab") [Adds a "Bugs" tab to repos, if there are any open issues with the "bug" label.](https://user-images.githubusercontent.com/1402241/73720910-a688d900-4755-11ea-9c8d-70e5ddb3bfe5.png)
 - [](# "pinned-issue-update-time") [Adds the updated time to pinned issues.](https://user-images.githubusercontent.com/1402241/75525936-bb524700-5a4b-11ea-9225-466bda58b7de.png)
 - [](# "clean-pinned-issues") [Changes the layout of pinned issues from side-by-side to a standard list.](https://user-images.githubusercontent.com/1402241/84509958-c82a3c00-acc4-11ea-8399-eaf06a59e9e4.png)
+- [](# "remove-label-with-a-single-click") [Adds one-click buttons to remove labels in conversations.](https://user-images.githubusercontent.com/36174850/89980178-0bc80480-dc7a-11ea-8ded-9e25f5f13d1a.gif)
```
fregante, line 224: I'd shorten the name (to be used in the classname as well)

```suggestion
- [](# "remove-label-button") [Adds one-click buttons to remove labels in conversations.](https://user-images.githubusercontent.com/36174850/89980178-0bc80480-dc7a-11ea-8ded-9e25f5f13d1a.gif)
```

source/features/remove-label-with-a-single-click.tsx
```diff
@@ -0,0 +1,84 @@
...
+	);
+
+	closeButton.style.setProperty('--rgh-remove-label-bg', color);
+	closeButton.style.setProperty('--rgh-remove-label-color', backgroundColor);
+
+	return closeButton;
+}
+
+function init(): void {
+	for (const label of select.all('.labels > a')) {
```
fregante, line 59:
```suggestion
	await api.expectToken();

	for (const label of select.all('.labels > a')) {
```

source/features/remove-label-with-a-single-click.tsx
```diff
@@ -0,0 +1,84 @@
...
+			data: {},
+			cached: false
+		}
+	}));
+}
+
+async function removeButtonClickHandler(event: delegate.Event<MouseEvent, HTMLSpanElement>): Promise<void> {
+	event.preventDefault();
+
+	const removeButton = event.delegateTarget;
```
fregante, line 31: Is it a `closeButton` or `removeButton`? Let's be consistent everywhere in this file

source/features/remove-label-with-a-single-click.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+function updateSidebar() {
+	select('#partial-discussion-sidebar')!.dispatchEvent(new CustomEvent('socket:message', {
+		bubbles: false,
+		cancelable: false,
+		detail: {
+			name: '',
+			data: {},
+			cached: false
+		}
```
fregante, line 24: Can this config be reduced? What's the shortest it can be?

source/features/remove-label-with-a-single-click.css
```diff
@@ -0,0 +1,23 @@
+.rgh-remove-label-with-a-single-click {
+	display: flex;
+	color: inherit;
+	margin-left: 2px;
+	margin-right: -7px;
+}
+
+.rgh-remove-label-with-a-single-click svg {
+	margin-top: 0.125rem;
```
fregante, line 9: Let's use `px` everywhere. That's what GitHub uses

source/features/remove-label-with-a-single-click.tsx
```diff
@@ -0,0 +1,84 @@
...
+	await api.v3(`repos/${getRepoURL()}/issues/${getConversationNumber()!}/labels/${removeButton.dataset.name!}`, {
+		method: 'DELETE'
+	});
+
+	updateSidebar();
+}
+
+function makeRemoveButton(labelName: string, color: string, backgroundColor: string) {
+	const closeButton = (
+		<span
```
fregante, line 43: This should be a button, not a span (if it works in both Chrome and Firefox)

source/features/remove-label-with-a-single-click.tsx
```diff
@@ -0,0 +1,84 @@
...
+import React from 'dom-chef';
+import XIcon from 'octicon/x.svg';
+import select from 'select-dom';
+import oneTime from 'onetime';
+import delegate from 'delegate-it';
+import * as pageDetect from 'github-url-detection';
+
+import features from '.';
+import * as api from '../github-helpers/api';
+import onReplacedElement from '../helpers/on-replaced-element';
```
fregante, line 11: Could you look into using `selector-observer` like in some other features? It's more reliable than this

source/features/remove-label-with-a-single-click.css
```diff
@@ -0,0 +1,23 @@
...
+	margin-bottom: 0.125rem;
+	margin-right: 0.125rem;
+	border-radius: 50%;
+	padding-top: 0.125rem;
+	padding-bottom: 0.125rem;
+	fill: currentColor;
```
fregante, line 15: I think this is the default
```suggestion
```

source/features/remove-label-with-a-single-click.css
```diff
@@ -0,0 +1,23 @@
...
+	width: 14px !important;
+	height: 14px !important;
+}
+
+.rgh-remove-label-with-a-single-click svg:hover {
```
fregante, line 20: `:hover, :focus` should apply to the clickable element
```suggestion
.rgh-remove-label-with-a-single-click:hover svg {
```

source/features/remove-label-button.tsx
```diff
@@ -0,0 +1,86 @@
...
+}
+
+void features.add({
+	id: __filebasename,
+	description: 'Adds one-click buttons to remove labels in conversations.',
+	screenshot: 'https://user-images.githubusercontent.com/36174850/89980178-0bc80480-dc7a-11ea-8ded-9e25f5f13d1a.gif'
+}, {
+	include: [
+		pageDetect.isIssue,
+		pageDetect.isPR
```
fregante, line 80:
```suggestion
		pageDetect.isIssue,
		pageDetect.isPRConversation
```

source/features/remove-label-button.tsx
```diff
@@ -0,0 +1,86 @@
...
+	removeLabelButton.style.setProperty('--rgh-remove-label-button-bg-color', backgroundColor);
+	removeLabelButton.style.setProperty('--rgh-remove-label-button-color', color);
+
+	return removeLabelButton;
+}
+
+async function init(): Promise<void> {
+	await api.expectToken();
+
+	observe('.labels > a', {
```
fregante, line 61: I think this will need a `:not(.rgh-something)` because we need to ensure it's not applied when the user pressed the browser’s back button.

source/features/remove-label-button.tsx
```diff
@@ -0,0 +1,86 @@
...
+		constructor: HTMLElement,
+		add(label) {
+			// Override !important rule
+			label.style.setProperty('display', 'inline-flex', 'important');
```
fregante, line 65: Probably: 
```suggestion
			label.classList.add('d-inline-flex', 'rgh-something');
```

source/features/remove-label-button.tsx
```diff
@@ -0,0 +1,86 @@
...
+			label.append(makeRemoveLabelButton(label.dataset.name!, label.style.color, label.style.backgroundColor));
```
fregante, line 66: You can inline it here and use the `style` attribute
```suggestion
			label.append(
				<button
					type="button"
					aria-label="Remove this label"
					className="btn-link tooltipped tooltipped-nw rgh-remove-label-button"
					data-name={labelName},
					style={{
						'--rgh-remove-label-button-bg-color': 'label.style.backgroundColor'
					}}
				>
					<XIcon/>
				</button>
			);
```

source/features/remove-label-button.tsx
```diff
@@ -0,0 +1,86 @@
+import './remove-label-button.css';
+import React from 'dom-chef';
+import XIcon from 'octicon/x.svg';
+import select from 'select-dom';
+import oneTime from 'onetime';
+import delegate from 'delegate-it';
+import * as pageDetect from 'github-url-detection';
+import {observe} from 'selector-observer';
```
fregante, line 8:
```suggestion
import {observe} from 'selector-observer';
import * as pageDetect from 'github-url-detection';
```

source/features/remove-label-button.tsx
artusm: It doesn't work.
Screenshot_20200813_023350

source/features/remove-label-button.tsx
artusm: It doesn't work too. 
`dom-chef` doesn't support css vars in the `style` attribute, because `dom-chef` [uses](https://github.com/vadimdemedes/dom-chef/blob/master/index.ts#L28) `element.style[name as any] = value;` for setting css properties.
css

source/features/remove-label-button.tsx
fregante: I don't like setting CSS via javascript when there's already a class and/or a CSS file we can use. You can add it via CSS with:

```css
:root .rgh-remove-label-button-already-added {
  display: inline-flex !important;
}
```

Maybe `:root` has to be doubled: `:root:root`. We use this format often, sometimes just to override the native selector without using `!important` (this is not the case)

source/features/remove-label-button.css
```diff
@@ -0,0 +1,21 @@
...
+	color: inherit !important;
+}
+
+.rgh-remove-label-button svg {
+	margin: 2px;
+	margin-left: 0;
+	border-radius: 50%;
+	padding: 2px 0;
+	width: 14px !important;
+	height: 14px !important;
```
fregante, line 14: I'm not sure that `!important` is required here. I think CSS properties override SVG attributes

source/features/remove-label-button.tsx
fregante: Ouch. I opened an issue for that. Let's add a comment:

```js
// TODO: set variable via JSX and inline function in `init` after https://github.com/vadimdemedes/dom-chef/issues/66
```

source/features/remove-label-button.tsx
```diff
@@ -0,0 +1,85 @@
...
+	event.preventDefault();
+
+	const removeLabelButton = event.delegateTarget;
+
+	removeLabelButton.disabled = true;
+	await api.v3(`repos/${getRepoURL()}/issues/${getConversationNumber()!}/labels/${removeLabelButton.dataset.name!}`, {
+		method: 'DELETE'
+	});
+
+	updateSidebar();
```
fregante, line 37: This doesn't seem to work reliably, especially in Firefox.

Let's just remove the label, it's faster anyway
```suggestion
	removeLabelButton.closest('a')!.remove()
```

source/features/remove-label-faster.tsx
```diff
@@ -0,0 +1,85 @@
...
+	event.preventDefault();
+
+	const removeLabelButton = event.delegateTarget;
+
+	removeLabelButton.disabled = true;
+	await api.v3(`repos/${getRepoURL()}/issues/${getConversationNumber()!}/labels/${removeLabelButton.dataset.name!}`, {
+		method: 'DELETE'
+	});
+
+	updateSidebar();
```
fregante, line 37: This doesn't seem to work reliably, especially in Firefox.

Let's just remove the label, it's faster anyway
```suggestion
	removeLabelButton.closest('a')!.remove()
```

fregante:
What's the bug?

source/features/remove-label-faster.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+	removeLabelButton.closest('a')!.remove();
+
+	const addLabelMenu = select('.sidebar-labels details-menu')!;
+	const addLabelMenuContent = addLabelMenu.querySelector('.hx_rsm-content');
+
+	if (addLabelMenuContent?.querySelector('include-fragment') === null) {
+		addLabelMenuContent.innerHTML = '';
+		addLabelMenuContent.append(<include-fragment src={addLabelMenu.getAttribute('src')!}/>);
+	}
```
fregante, line 35: I saw the issue. Here are some minor improvements:

- adds explanatory comment
- only 1 query instead of 3, if `include-fragment` is still there
- avoids setting `src` (normally it's not there)
```suggestion
	// Force update of label selector if necessary
	if (!select.exists('.sidebar-labels include-fragment')) {
		const deferredContentWrapper = select('.sidebar-labels .hx_rsm-content');
		deferredContentWrapper.textContent = '';
		deferredContentWrapper.append(<include-fragment/>);
	}
```

artusm: > * avoids setting `src` (normally it's not there)

Peek 2020-08-13 22-06

fregante: Love the reverse engineering!

However the second point still applies, your code does 3 queries before it’s necessary. Let’s use my code and add `const menu = deferredContentWrapper.closest('[src]')!`

AuHau:
Wow thanks @artusm !

fregante:
Oops! The drawback of using `selector-observer`: this feature shouldn't be enabled on lists, but there's currently no mechanism to disable `selector-observer` between ajaxed page changes, so a feature will stay active even when not necessary




1. Visit issue https://github.com/sindresorhus/refined-github/issues/3084
1. Click on Pull Requests tab, sometimes the list is loaded via ajax

I think we need to use `deinit` to call [`observer.abort()`](https://github.com/josh/selector-observer/blob/b811bb5d13cba9940026430f92a6760d5f9e365c/lib/selector-observer.js#L73)

cc @kidonng

PR #3467, opened 2020-08-16
Add `conversation-navigation` feature

### LINKED ISSUE #343: Left and right navigation to quickly go through issues and commits
I'm wondering if something like this has already been added, if not myself or someone else who has the time should try and add this functionality. If you've ever used a system such as Taiga, there's nice left and right navigation buttons to quickly go through issues, especially if you have a lot of them.

Currently with Github, every time you read an issue and comment / close it, you then have to go back to the entire list of issues, remember where you left off and somehow keep track of the next issue to open.

This lack of functionality has bothered me so much that I actually sent Github staff a long email about it over a year ago. They replied but since then have not added this basic functionality, for a reference this is precisely what I'm describing, if you've ever used Taiga they already have this.

github-annoyances

As well, a similar type of functionality when you want to quickly go through commits, review changes, pull requests, etc. would also be extremely helpful.

github-annoyances2

iamandrewluca:
image

Closes #343


iamandrewluca:
The only edge case IMO left is when user closes a conversation, but cache still exists for the page where that conversation existed.

iamandrewluca:
@fregante I made some changes, now buttons in disabled state are added before making any requests.

Can we move cursor functions into helpers?

fregante:
I still see large chunks of repeated code all over the file, where the only difference is `-1` vs `+1`. Most of the code can be streamlined into something like

```js
function doSomething(){}
await fetchData()
doSomething(-1)
doSomething(1)
```

Probably you can merge a lot of the functions into a single "doSomething" that returns a URL or false (`string | false`). If you simply that over and over you'll find it that this feature really around 120 lines at most.

An easy way to think about it is: 
1. write the simplest function possible to create one button
2. call the function twice


> we keep them in this feature until other feature will need them?

This.

source/features/conversation-navigation.tsx
```diff
@@ -0,0 +1,258 @@
...
+function getNextCursor(cursor: string) {
+	return getCursor(getOffset(cursor) + 1);
+}
+
+function getPageCursors(page: number) {
+	return {
+		after: getCursor((page - 1) * ITEMS_PER_PAGE),
+		before: getCursor((page * ITEMS_PER_PAGE) + 1)
+	};
+}
```
fregante, line 243: All of this for example should be at most:

```js
function getCursor(offset: number): string {
  	return btoa(`cursor:${offset}`);
  }
 
  function alterCursor(cursor: string, diff: number): string {
  	const index = Number.parseInt(atob(cursor).replace('cursor:', ''), 10);
  	return getCursor(index + diff);
  }
```

Then just call `alterCursor(cursor, diff)`, where diff is -1 or +1

Also I think almost everywhere you return an object, it's making things unnecessarily lengthy

source/features/conversation-navigation.tsx
```diff
@@ -0,0 +1,258 @@
...
+	type: ISSUE,
+	first: ${first},
+	after: "${after}",
+	query: "${query}"
+) {
+	edges {
+		cursor
+		node {
+			... on Issue { number, url }
+			... on PullRequest { number, url }
```
fregante, line 195: I think:
```suggestion
			... on IssueOrPullRequest { number, url }
```

fregante: Also I think that if you write it like this, you won't have to write `edges.node.map()` later, but just `node.map()`:

```gql
edges {
  cursor
}
node {
  ... on IssueOrPullRequest { number, url }
}
```

source/features/conversation-navigation.tsx
```diff
@@ -0,0 +1,258 @@
...
+function getListQuery() {
+	const referrerUrl = new URL(document.referrer);
+
+	// Coming from another issue with query in URL
+	if (!pageDetect.isConversationList(referrerUrl)) {
+		return parseConversationListURL(new URL(location.href));
+	}
+
+	// Default conversation lists, even global, don't have query in URL
+	// But it seems there is some refined-github feature that adds a default query in URL based on query input
```
fregante, line 154: GitHub does have default queries, for example `/pulls` implies `is:open is:pr author:@me archived:false`. This is by default.

You'll be happy to know that this issue is already handled by `search-query.ts`

https://github.com/sindresorhus/refined-github/blob/2863dddde6fa72d9dd85fea242e69ed650b6cf88/source/features/sort-conversations-by-update-time.tsx#L23

This means that `getListQuery` could be replaced with

```ts
function getConversationList(): URL | location {
	const referrerUrl = new URL(document.referrer);
	return pageDetect.isConversationList(referrerUrl) ? referrerURL : location;
}
```

And then in `getConversationList`:

```diff
  function getConversationList() {
-	const listQuery = getListQuery();
+   const listUrl = getConversationList();
+   const query = new SearchQuery(listUrl).get();
+   const page = listUrl.searchParams.get('page') ?? '1';
```

source/features/conversation-navigation.tsx
iamandrewluca: It seems this is not valid. Also I need cursors to be near data.

image

fregante: No, that part is actually valid; it's the `IssueOrPullRequest` that's missing in this case sadly. This works:


```gql
{
  search(type: ISSUE, first: 10, query: "is:issue repo:sindresorhus/refined-github") {
    edges {
      cursor
    }
    nodes {
      ... on Issue {
        number
      }
      ... on PullRequest {
        number
      }
    }
  }
}
```

fregante:
If you force-push you make it difficult to see the changes. Force-pushing is never necessary for this repository ✌️

iamandrewluca:
@fregante Ok. I don't quite understand how to use `SearchQuery` in the context of this feature. Now I;m doing some experiments.

source/features/conversation-navigation.tsx
iamandrewluca: In this case I get cursors separated from nodes in a `edges` array

fregante:
I think all the use you need is in the code I suggested earlier. Copy-paste it and it should be done, I think.

iamandrewluca:
Think I should handle the case when `pageDetect.isRepoConversationList` and add `repo:${getRepoURL()}`

iamandrewluca:
Kind of works :man_shrugging: Midnight over here. Good night.

source/features/conversation-navigation.tsx
```diff
@@ -0,0 +1,195 @@
...
+		};
+	}
+
+	// `currentConversation` is from previous page, we need to fetch previous page items
+	listQuery.page = conversation!.cursor === after ?
+		listQuery.page - 1 :
+		// `currentConversation` is from next page, we need to fetch next page items
+		(conversation!.cursor === before ?
+			listQuery.page + 1 :
+			listQuery.page);
```
fregante, line 106:
```suggestion
	if(conversation!.cursor === after) {
		// `currentConversation` is from previous page, we need to fetch previous page items
		listQuery.page -= 1
	} else if (conversation!.cursor === before) {
		// `currentConversation` is from next page, we need to fetch next page items
		listQuery.page += 1;
	}
```

source/features/conversation-navigation.tsx
```diff
@@ -0,0 +1,195 @@
...
+	url: string;
+};
+
+const fetchConversationList = cache.function(
+	async (listQuery: ReturnType<typeof getListQuery>): Promise<Conversation[]> => {
+		const searchQuery = getConversationSearchQuery(listQuery);
+
+		return api.v4(searchQuery).then(({search}) => search.edges.map((edge: any) => ({
+			...edge.node,
+			cursor: edge.cursor
```
fregante, line 76: From what I understand, the cursor is just a "stringified" number. Let's just store the number and only convert it the string when querying the API.

```suggestion
			cursor: getOffset(edge.cursor)
```

This makes operations easier

iamandrewluca: Makes sense, so we can get rid of all cursor functions and do basic maths.
I'll take a look tommorow.

iamandrewluca:
Something is broke. Investigating.

iamandrewluca:
Single problem left is when a page is cached, and after some, time less then the cache, a conversation from that page gets closed.
User will still navigate through that closed conversation.

source/features/conversation-navigation.tsx
```diff
@@ -0,0 +1,179 @@
...
+			... on PullRequest { number, url }
+		}
+	}
+}
+	`;
+}
+
+function getConversationNumber() {
+	const conversationHashtag = select('.gh-header-number')?.textContent ?? '#';
+	return Number.parseInt(conversationHashtag.replace('#', ''), 10);
```
yakov116, line 163:
```suggestion
	return looseParseInt(conversationHashtag);
```

I think this should work

yakov116, line 163:
```suggestion
	return looseParseInt(conversationHashtag);
```

I think this should work

source/features/conversation-navigation.tsx
```diff
@@ -0,0 +1,164 @@
...
+	select('.gh-header-meta')!.append(
+		<div className="BtnGroup ml-2">
+			{previousButton}
+			{nextButton}
+		</div>
+	);
+
+	const conversationNumber = getConversationNumber();
+	const {list, query, page} = await getConversationList(conversationNumber);
+	const conversation = list.find(item => item.number === conversationNumber);
```
fregante, line 40: You pass `conversationNumber` to a function, why does that function return a list instead of the specific conversation you specifically requested? It should return `conversation` itself.

source/features/conversation-navigation.tsx
```diff
@@ -0,0 +1,164 @@
...
+			className="btn btn-lg BtnGroup-item"
+			aria-label={`Navigate to ${direction} Conversation`}
+			data-hotkey={direction.charAt(0)}
+		>
+			{direction === 'next' ? <ChevronRightIcon/> : <ChevronLeftIcon/>}
+		</a>
+	) as unknown as HTMLAnchorElement;
+}
+
+async function init() {
```
fregante, line 27: `Init` should be at the bottom of the file, right above `features.add`

source/features/conversation-navigation.tsx
```diff
@@ -0,0 +1,164 @@
...
+				cursor
+				node {
+					... on Issue { number, url }
+					... on PullRequest { number, url }
+				}
+			}
+		}
+	`);
+
+	return search.edges.map((edge: any) => ({
```
fregante, line 94: Would it be even easier if, instead of returning an array of objects, you return a array-like object?

```js
{
  3: [number, url],
  4: [number, url],
  5: [number, url],
  6: [number, url], etc
}
```

Basically

```js
const list = {};
for (const edge of search.edges) {
  list[looseParseInt(atob(edge.cursor))] = edge.node
}
return list;
```

This would save you from having to use `.find` over and over, because you can just always use `list[cursor]`

source/features/conversation-navigation.tsx
```diff
@@ -0,0 +1,164 @@
...
+		page,
+		list.find(item => item.cursor === conversation!.cursor + 1)
+	);
+}
+
+function setButtonHref(button: HTMLAnchorElement, query: string, page: number, conversation?: Conversation) {
+	if (conversation) {
+		const url = new URL(conversation.url);
+		url.searchParams.set('q', query);
+		url.searchParams.set('page', Number(page));
```
fregante: I think these should not be public. Can you try to move them to `sessionStorage`?

source/features/conversation-navigation.tsx
```diff
@@ -0,0 +1,164 @@
...
+	const nextButton = getButton('next');
+
+	select('.gh-header-meta')!.append(
+		<div className="BtnGroup ml-2">
+			{previousButton}
+			{nextButton}
+		</div>
+	);
+
+	const conversationNumber = getConversationNumber();
```
fregante, line 38: This function is returning a string for some reason, could you fix that with `Number()` in the function itself?

fregante (changes requested):

source/features/conversation-navigation.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+	if (pageDetect.isRepoConversationList(url)) {
+		return new SearchQuery(url).add(`repo:${getRepoURL()}`);
+	}
+
+	return new SearchQuery(url);
+}
+
+void features.add({
+	id: __filebasename
```
fregante, line 153: The point of using strict types here is to make it painfully obvious that they need to be set. If we use junk data (`description: '', screenshot: false`) we defeat its purpose, that's why I removed them in https://github.com/sindresorhus/refined-github/pull/3467/commits/be54b915ae65e829c2db121d13645113c15512fb

source/features/conversation-navigation.tsx
iamandrewluca: Is this still valid?

source/features/conversation-navigation.tsx
iamandrewluca: Fixed

source/features/conversation-navigation.tsx
iamandrewluca: Done

source/features/conversation-navigation.tsx
iamandrewluca: Renamed `getConversationList` to `getConversationMeta` and moved all that logic for finding next/previous inside

source/features/conversation-navigation.tsx
iamandrewluca: I can do this, but then searching for current conversation looks ugly. Something like

```typescript
const conversation = Object.entries(list).find(([key, value]) => value[0] === conversationNumber)![1]
```

source/features/conversation-navigation.tsx
iamandrewluca: Added a more descriptive description, and screenshot to false

iamandrewluca:
from 263 to 152 lines, nice code shrinking :slightly_smiling_face:

source/features/conversation-navigation.tsx
fregante: Why did you set false again? This feature needs a screenshot

source/features/conversation-navigation.tsx
fregante: Yes. It changes the feature a bit because the query is either in document.referrer searchParams OR in the sessionsStorage

fregante:
What problems do they add? If the `search()` ignores them, that’s a big problem; the order needs to match

source/features/conversation-navigation.tsx
```diff
@@ -0,0 +1,151 @@
...
+		</a>
+	) as unknown as HTMLAnchorElement;
+}
+
+function setButtonHref(button: HTMLAnchorElement, query: string, page: number, conversation?: Conversation) {
+	if (conversation) {
+		const url = new URL(conversation.url);
+		url.searchParams.set('q', query);
+		url.searchParams.set('page', String(page));
+		button.removeAttribute('aria-disabled');
```
yakov116, line 32: Added the following will add the hovercard.
@fregante WDYT? I love it ❤️ 
```suggestion
		button.removeAttribute('aria-disabled');
		button.dataset.hovercardUrl = url.pathname + '/hovercard';
```

yakov116, line 32: Added the following will add the hovercard.
@fregante WDYT? I love it ❤️ 
```suggestion
		button.removeAttribute('aria-disabled');
		button.dataset.hovercardUrl = url.pathname + '/hovercard';
```

iamandrewluca:
@fregante I mean this use case

1. User goes to  issues list
2. User opens an issue
3. First page of items is fetched and cached
4. User closes the issue
5. User goes back to issue list
6. User opens a issue
7. Issue list is loaded from cache (with closed issue  also)
8. User navigates forward and will bump in closed issue that is not in list anymore

Right now the only solution I see is to decrease caching time as much as possible.

fregante:
I see what you mean. I see 2 big issues:

1. The order is dynamic: if we leave a comment, the current issue will end up first. 

This is ok if the list is cached. Once the cache expires, the order is reset and meaningless. Also once you reach the boundaries, the new fetched list might have skipped some issues. 

2. The current issue might leave the list altogether

If you close the issue or remove a tag or assignee, the list of issue might change altogether, so if we fetch it again, the current issue might not be on it and the feature won’t work. This issue is relatively minor because the links are generated on load, before any changes/closure. 

—

Especially for 1, this feature has potential to be confusing, making you skip issues and find yourself somewhere you don’t expect to be.

The only solution is see, is to exclusively preserve the list **as seen** when opening the issue. When I open an issue, store the list. That’s it. That’s the only list we’re navigating. We can’t fetch more. 

We could avoid the API call altogether and instead:

```
onConversationList
  store current list in session storage

onConversation
  is there a list in session storage?
    is current conversation in list?
      show buttons
```

iamandrewluca:
We can make the API call when we reach the boundaries, otherwise we can't navigate through pages.

fregante:
Right, but as I explained we can’t, because the user might be in a place that doesn’t make sense. Example:

1. Go to first page of issues
2. Open last issue
2. Mark as closed
3. Click “next issue”, you’ll see the first issue of the second page and the API fetches again
4. Click “next issue”, you’ll see the THIRD issue of the second page, because it became the second one

When the user reaches the end of the preloaded list, it’s done, they have to go back to the list

iamandrewluca:
> When the user reaches the end of the preloaded list, it’s done, they have to go back to the list

So I should start refactoring?

- Navigation only for previous discussion list page
- Store that list in session storage

fregante:
What you can do however, is to keep the API call, but request ~25 before and ~25 after the current conversation, so the user can still move a lot in the current list, until they go back to the `/issues` page.

iamandrewluca:
> ~25 before and ~25 after

Ok, but still we have to fetch one page backwards and one forwards. We can't query exactly from a conversation.

fregante:
> We can't query exactly from a conversation.

Why can't we? We know the exact position of the conversation on the current page, so use that.

```js
Page: 2
Position: 5
---
RealPosition: ((Page - 1) * MaxPerPage + Position)


Start: Math.max(0, RealPosition - MaxPerPage)
Count: 50
```

fregante:
Refer to the full suggestion earlier. The code would start in the list now, that's why we know the position already:

```
onConversationList
  store current list in session storage

onConversation
  is there a list in session storage?
    is current conversation in list?
      show buttons
```

iamandrewluca:
I'm on a little vacation, I'll come back next week and resume the work.

yakov116:
@iamandrewluca I merged master and fixed the build. Hope to see this merged!

iamandrewluca:
Looking forward to finish this. At the moment I have busy schedule at work :rocket:

source/features/conversation-navigation.tsx
```diff
@@ -71,13 +71,15 @@ const fetchConversationList = cache.function(async (query: string, page: number)
...
+	maxAge: {
+		hours: 1
+	}
 });
 
 // Current default number of items in a conversation list
 const ITEMS_PER_PAGE = 25;
 
-async function getConversationMeta() {
+async function getConversationMeta(): Promise<AnyObject> {
```
fregante, line 82: AnyObject is a meaningless type and it’s wrong here. Only use it while parsing API results.

yakov116: 👍 I did it just to make the build pass after the merge with master

fregante:
Refer to https://github.com/sindresorhus/refined-github/issues/343#issuecomment-698673577

source/features/conversation-navigation.tsx
fregante: The build is strict to enforce good coding standards; if we just bypass them it forces me to become a linter. Let the build complain until it's actually fixed.

yakov116: Noted

PR #3476, opened 2020-08-17
Fix `remove-label-faster`

artusm:
1. LINKED ISSUES:
   [comment](https://github.com/sindresorhus/refined-github/pull/3454#issuecomment-674854302)

2. TEST URLS:
[sindresorhus/refined-github/issues/3440](https://github.com/sindresorhus/refined-github/issues/3440)
[sindresorhus/refined-github/pull/3454](https://github.com/sindresorhus/refined-github/pull/3454)

3. SCREENSHOT:

### DIFF
--- a/source/features/remove-label-faster.tsx
+++ b/source/features/remove-label-faster.tsx
@@ -84,6 +84,6 @@ void features.add({
 	exclude: [
 		canNotEditLabels
 	],
-	init: oneTime(init),
+	init,
 	deinit
 });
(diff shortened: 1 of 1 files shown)


source/features/remove-label-faster.tsx
```diff
@@ -79,5 +84,6 @@ void features.add({
 	exclude: [
 		canNotEditLabels
 	],
-	init: oneTime(init)
+	init: oneTime(init),
```
fregante, line 87: I think we also need to drop `oneTime` since the observer is removed now

kidonng:
I know I'm late😥 but is prepending `.discussion-sidebar-item` to the selector enough?

fregante:
> I know I'm late😥 but is prepending `.discussion-sidebar-item` to the selector enough? That way we don't need `deinit`

Likely. PR welcome

PR #3408, opened 2020-07-27
Add `clean-conversation-headers` feature

kidonng:
Resolves #3089

Open issue:

image

Note: if someone just created a PR to close the issue or someone edited currently displayed PR, GitHub will update the header and this feature won't reapply

Open issue with pending PR:

image

Closed issue:

image

Open PR (same author, to default branch):

image

Merged PR (same author, to default branch):

image

Merged PR (different author, to default branch):

image

Open PR (same author, to non-default branch):

image

PR #3478, opened 2020-08-18
Lint

yakov116:
See [this](https://github.com/sindresorhus/refined-github/pull/3442/commits/3aa8dd8e42a672ef0d9cb37f22ee0c899228ca86) for word definition

- Exclude isEmptyRepoRoot from `repo-age`
- Expect token for `clone-branch`
- Use `features.error` instead of throwing (firefox)
- Tuple Types
- Add a deinit for `highlight-deleted-and-added-files-in-diffs`. When switching pages, the list needs a reload.
- Fix `clean-conversation-headers` error see https://github.com/sindresorhus/refined-github/pull/3408#issuecomment-675449244 (not sure if this fixes both errors)

### DIFF
--- a/source/features/clean-conversation-headers.tsx
+++ b/source/features/clean-conversation-headers.tsx
@@ -23,7 +23,7 @@ function initPR(): void {
 	observe('.gh-header-meta .flex-auto', {
 		add(byline) {
 			const isMerged = select.exists('#partial-discussion-header [title="Status: Merged"]');
-			const isSameAuthor = select('.js-discussion > .TimelineItem:first-child .author')?.textContent === select('.author', byline)!.textContent;
+			const isSameAuthor = select('.js-discussion > .TimelineItem:first-child .author')?.textContent === select('.author', byline)?.textContent;
 			const baseBranch = select('.commit-ref:not(.head-ref)', byline)!;
 			const isDefaultBranch = (baseBranch.firstElementChild as HTMLAnchorElement).pathname.split('/').length === 3;
 
--- a/source/features/clean-conversation-headers.tsx
+++ b/source/features/clean-conversation-headers.tsx
@@ -20,10 +20,11 @@ function initIssue(): void {
 }
 
 function initPR(): void {
-	observe('.gh-header-meta .flex-auto', {
+	observe('.gh-header-meta .flex-auto:not(.rgh-clean-conversation-header)', {
 		add(byline) {
+			byline.classList.add('rgh-clean-conversation-header');
 			const isMerged = select.exists('#partial-discussion-header [title="Status: Merged"]');
-			const isSameAuthor = select('.js-discussion > .TimelineItem:first-child .author')?.textContent === select('.author', byline)?.textContent;
+			const isSameAuthor = select('.js-discussion > .TimelineItem:first-child .author')?.textContent === select('.author', byline)!.textContent;
 			const baseBranch = select('.commit-ref:not(.head-ref)', byline)!;
 			const isDefaultBranch = (baseBranch.firstElementChild as HTMLAnchorElement).pathname.split('/').length === 3;
 
--- a/source/features/highlight-deleted-and-added-files-in-diffs.tsx
+++ b/source/features/highlight-deleted-and-added-files-in-diffs.tsx
@@ -1,15 +1,19 @@
 import React from 'dom-chef';
 import select from 'select-dom';
-import oneTime from 'onetime';
-import {observe} from 'selector-observer';
 import elementReady from 'element-ready';
 import * as pageDetect from 'github-url-detection';
+import {observe, Observer} from 'selector-observer';
 
 import features from '.';
 import {observeOneMutation} from '../helpers/simplified-element-observer';
 
+let observer: Observer;
+
 async function loadDeferred(jumpList: Element): Promise<void> {
 	const loadJumpList = (jumpList: Element) => jumpList.parentElement!.dispatchEvent(new MouseEvent('mouseover'));
+	// Between ajax pages the load will be instant
+	loadJumpList(jumpList);
+	// The shortest time to load in 700 ms
 	setTimeout(loadJumpList, 700, jumpList);
 	// The event listener might not have been attached yet, so we can try twice
 	setTimeout(loadJumpList, 1200, jumpList);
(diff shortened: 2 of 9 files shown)


source/features/repo-age.tsx
```diff
@@ -122,6 +137,9 @@ void features.add({
 	include: [
 		pageDetect.isRepoRoot
 	],
+	exclude: [
+		pageDetect.isEmptyRepoRoot
+	],
```
yakov116, line 142: @fregante I dont this will work since we have `waitForDomReady: false`

source/features/highlight-deleted-and-added-files-in-diffs.tsx
```diff
@@ -65,6 +69,7 @@ void features.add({
 		pageDetect.isPRFile404,
 		pageDetect.isPRCommit404
 	],
-	init: oneTime(init),
+	init,
+	deinit: () => observer.abort(),
```
yakov116, line 73: Needed since when you change from a commit page to the file changes tab, we need to reload the list.

source/features/remove-label-faster.tsx
```diff
@@ -85,5 +80,5 @@ void features.add({
 		canNotEditLabels
 	],
 	init,
-	deinit
+	deinit: () => observer.abort()
```
yakov116, line 83: Good?

source/features/repo-age.tsx
```diff
@@ -42,16 +49,17 @@ const getRepoAge = async (commitSha: string, commitsCount: number): Promise<[str
 	return [committedDate, resourcePath];
 };
 
-const getFirstCommit = cache.function(async (): Promise<[string, string]> => {
+// eslint-disable-next-line import/prefer-default-export
+export const getLatestCommitInfo = cache.function(async (): Promise<LatestCommitInfo> => {
```
yakov116, line 53: Should this be in its own file?

fregante: Nah, queries like this grow and disappear as needed, they’re not general-purpose enough. Extracting them would make it seem like it’s the official way of getting the commit info, which it isn’t.

source/features/repo-age.tsx
fregante: It might, or it might not. I’m not particularly concerned about console errors on the rare empty repo. 

Anyway most of the changes to this file should be in their own PR, this is not a lint

yakov116: ok will send later tonight

source/features/clean-conversation-headers.tsx
```diff
@@ -23,7 +23,7 @@ function initPR(): void {
 	observe('.gh-header-meta .flex-auto', {
 		add(byline) {
 			const isMerged = select.exists('#partial-discussion-header [title="Status: Merged"]');
-			const isSameAuthor = select('.js-discussion > .TimelineItem:first-child .author')?.textContent === select('.author', byline)!.textContent;
+			const isSameAuthor = select('.js-discussion > .TimelineItem:first-child .author')?.textContent === select('.author', byline)?.textContent;
```
fregante, line 26: What’s causing this issue exactly? How can I replicate it?

yakov116: Go to the file changes tab then go to the main tab

yakov116: Sorry commits tab

fregante: I see. I don't think this is the right way to fix it though. Separately, this feature works on every tab, so something is changing only when switching tab and _that_ should probably be accounted for

fregante: What I'm seeing is that this feature doesn't have `:not(.rgh-clean-conversation-header)` in the selector, so that might be it.

source/features/highlight-deleted-and-added-files-in-diffs.tsx
```diff
@@ -1,18 +1,20 @@
...
+
 async function loadDeferred(jumpList: Element): Promise<void> {
+	// This event will trigger the loading, but if run too early, GitHub might not have attached the listener yet, so we try multiple times.
 	const loadJumpList = (jumpList: Element) => jumpList.parentElement!.dispatchEvent(new MouseEvent('mouseover'));
 	loadJumpList(jumpList);
-	// The event listener might not have been attached yet, so we can try twice
-	setTimeout(loadJumpList, 1000, jumpList);
+	setTimeout(loadJumpList, 700, jumpList);
+	setTimeout(loadJumpList, 1200, jumpList);
 	await observeOneMutation(jumpList);
```
fregante, line 18: Let's use this

```js
const retrier = setInterval(() => {
	jumpList.parentElement!.dispatchEvent(new MouseEvent('mouseover'))
}, 200);
await observeOneMutation(jumpList);
clearInterval(retrier);
```

source/features/clean-conversation-headers.tsx
yakov116: Without looking I think the issue is that the author only exists in the conversation view. I will check later tonight

yakov116: > What I'm seeing is that this feature doesn't have `:not(.rgh-clean-conversation-header)` in the selector, so that might be it.

Yup

source/features/highlight-deleted-and-added-files-in-diffs.tsx
```diff
@@ -1,19 +1,21 @@
...
+
 async function loadDeferred(jumpList: Element): Promise<void> {
-	const loadJumpList = (jumpList: Element) => jumpList.parentElement!.dispatchEvent(new MouseEvent('mouseover'));
-	loadJumpList(jumpList);
-	// The event listener might not have been attached yet, so we can try twice
-	setTimeout(loadJumpList, 1000, jumpList);
+	// This event will trigger the loading, but if run too early, GitHub might not have attached the listener yet, so we try multiple times.
+	const retrier = setInterval(() => {
+		jumpList.parentElement!.dispatchEvent(new MouseEvent('mouseover'));
+	}, 100);
```
yakov116, line 16: I did 100 since on ajax it really instant

fregante (approved):

fregante:
If you’re looking for more linting (separate PR maybe) there are a few rules here that were disabled, but are helpful and probably are not a problem to us:

https://github.com/xojs/eslint-config-xo-typescript/blob/master/index.js

Have a look, perhaps you can try adding them to our config and seeing if they’re reasonable. 

If they’re too difficult to add or seem to require a rewrite (or many eslint-ignore), let me know and don’t lose your mind over it 😃

yakov116:
I will _git_ it a ~lint~ shot

PR #3479, opened 2020-08-19
Fix `format-conversation-titles`

### LINKED ISSUE #3458: `format-conversation-titles` is not very reliable
1. Visit an issue like the current one, but that you can edit
2. Change the issue title (or add/remove a tag)

The formatting is sometimes lost. Sometimes it appears a second later, sometimes it doesn't

GitHub only replaces the text node, so we can't use `selector-observer`

artusm:
Thanks for contributing! 🍄

1. LINKED ISSUES:
   Closes #3458

2. TEST URLS:

3. SCREENSHOT:
   BeforeBefore

   AfterAfter

### DIFF
--- a/source/features/format-conversation-titles.tsx
+++ b/source/features/format-conversation-titles.tsx
@@ -3,34 +3,15 @@ import * as pageDetect from 'github-url-detection';
 
 import features from '.';
 import * as domFormatters from '../github-helpers/dom-formatters';
-import observeElement, {observeOneMutation} from '../helpers/simplified-element-observer';
-
-let headerWrapper: HTMLElement;
+import observeElement from '../helpers/simplified-element-observer';
 
 function init(): void {
-	for (const title of select.all('.js-issue-title:not(.rgh-formatted-title)')) {
-		title.classList.add('rgh-formatted-title');
-		domFormatters.linkifyIssues(title);
-		domFormatters.parseBackticks(title);
+	for (const title of select.all('.js-issue-title')) {
+		if (!select.exists('a, code', title)) {
+			domFormatters.linkifyIssues(title);
+			domFormatters.parseBackticks(title);
+		}
 	}
-}
-
-async function submitHandler(event: Event) {
-	const form = event.target as HTMLFormElement;
-
-	if (!form.id.startsWith('edit_header')) {
-		return;
-	}
-
-	const title = select('.js-issue-title')!;
-
-	await observeOneMutation(title, {
-		subtree: false,
-		childList: true
-	});
-
-	domFormatters.linkifyIssues(title);
-	domFormatters.parseBackticks(title);
 }
 
 void features.add({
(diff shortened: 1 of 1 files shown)


source/features/format-conversation-titles.tsx
```diff
@@ -23,6 +43,10 @@ void features.add({
 		pageDetect.isIssue
 	],
 	init() {
-		observeElement(select('#partial-discussion-header')!.parentElement!, init);
```
fregante, line 26: What if we just change this call to include subtree: true? I don’t like to use it on whole documents, but here the base element is just the header. Would that work?

artusm: > What if we just change this call to include subtree: true? I don’t like to use it on whole documents, but here the base element is just the header.

`
observeElement(select('#partial-discussion-header')!, init, {});
`

artusm: > What if we just change this call to include subtree: true? I don’t like to use it on whole documents, but here the base element is just the header.

``` javascript
observeElement(select('#partial-discussion-header')!, init, {
	subtree: true
});
```
> Would that work?

No, because GitHub replaces whole header element on new events (a new message, title, etc).
Peek 2020-08-19 12-25

artusm: We can use `onReplacedElement`

``` javascript
		init();
		headerWrapper = select('#partial-discussion-header')!.parentElement!;

		headerWrapper.addEventListener('submit', submitHandler);
	},
	additionalListeners: [
		() => void onReplacedElement('#partial-discussion-header', init)
	],
```

fregante: > because GitHub replaces whole header element on new events 

That doesn’t sound right, I just pushed a comment and changed the title (in another tab) and the feature kept working after the updates.

fregante: I think you mistakenly dropped `parentElement`

> ```js
> observeElement(select('#partial-discussion-header')!, init, {
> 	subtree: true
> });
> ```

The code should be

```diff
- observeElement(select('#partial-discussion-header')!.parentElement!, init);
+ observeElement(select('#partial-discussion-header')!.parentElement!, init, {
+ 	subtree: true
+ });
```

artusm: > I think you mistakenly dropped `parentElement`

No, it wasn't a mistake. I just thought you were talking
> I don’t like to use it on whole documents

about

``` javascript 
observeElement(select('#partial-discussion-header')!.parentElement!
```

Screenshot_20200819_142700

artusm: > The code should be

Peek 2020-08-19 14-59

artusm: > The code should be

It works without

`
:not(.rgh-formatted-title)
`

artusm: with `:not(.rgh-formatted-title)`
Peek 2020-08-19 15-11
without `:not(.rgh-formatted-title)`
Peek 2020-08-19 15-10

fregante: Indeed, that should be dropped. In this case:

```js
for (const title of select.all('.js-issue-title')) {
	if (!select.exists('a, code', title)) {
		domFormatters.linkifyIssues(title);
		domFormatters.parseBackticks(title);
	}
}

fregante (approved):

PR #3488, opened 2020-08-19
Enable `batch-open-conversations` on repos without write access

### LINKED ISSUE #3487: `batch-open-conversations` add checkboxes to repo without read access
### Summary

- Add checkboxes to issues for `batch-open-conversations` when i don't have write access to the repo.
- see PoC below

### Where

- when i go to Issues or Pull Requests (or bugs), for example https://github.com/sindresorhus/refined-github/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc

### What i see

- missing checkboxes (note: the `Open all` checkbox is visible)
Missing Checkboxes

### What i Expect

- see checkboxes like this image:

batch open

### PoC

when i add manually the missing HTML components then this function works also on repo without write access - and i can open only the selected items.

*Note: i just tested i only on this url https://github.com/sindresorhus/notifier-for-github/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc and not checked side effects and/or for other problems*

#### Screenshots

Bildschirmfoto 2020-08-19 um 19 49 52

Bildschirmfoto 2020-08-19 um 19 50 04

#### Modifications

it was done just with copy over HTML from a repo with write access

##### a) add checkboxes to each issue row:

add as first child of `div.js-issue-row > div.d-flex`:

```html
<label class="flex-shrink-0 py-2 pl-3  d-none d-md-block">
  <input 
    type="checkbox" 
    data-check-all-item="" 
    class="js-issues-list-check" 
    name="issues[]" 
    value="${issueId}" 
    aria-labelledby="issue_${issueId}_link" 
    autocomplete="off">
</label>
```

and fill in the `${issueId}`

##### b) add the triage menu (without the triage actions)

add this code after `div.table-list-filters`:

```html
<div class="table-list-triage flex-auto js-issues-toolbar-triage">
  <span class="text-gray table-list-header-toggle">
    <span data-check-all-count="">1</span> selected
    <button type="button" class="btn-link rgh-batch-open-issues pl-3">Open selected</button>
  </span>
</div>
```

##### c) add the "select all checkbox"

add this code before `div.table-list-filters`
```html
<div class="mr-3 d-none d-md-block">
  <input type="checkbox" data-check-all="" aria-label="Select all issues" autocomplete="off">
</div>
```

### Environment

Chrome: Version 84.0.4147.135 (Offizieller Build) (64-Bit)
Refined GitHub: 20.8.16

muescha:
### Description

Add checkboxes to issues for batch-open-conversations when i don't have write access to the repo.

### 1. LINKED ISSUES:

closes #3487

### 2. TEST URLS:

*Note: an repo i don't have write access:*
https://github.com/gatsbyjs/gatsby/pulls?page=2&q=is%3Apr+sort%3Aupdated-desc+is%3Aclosed+is%3Aunread

### 3. SCREENSHOT:

*Note: an repo i don't have write access:*
Bildschirmfoto 2020-08-20 um 00 50 52
Bildschirmfoto 2020-08-20 um 00 53 08

### DIFF
--- a/source/features/batch-open-conversations.tsx
+++ b/source/features/batch-open-conversations.tsx
@@ -68,7 +68,7 @@ async function init(): Promise<void | false> {
 		const issuesToolbar = select('#js-issues-toolbar')!;
 		issuesToolbar.prepend(
 			<div className="mr-3 d-none d-md-block">
-				<input type="checkbox" data-check-all="" aria-label="Select all issues" autoComplete="off"/>
+				<input data-check-all type="checkbox" aria-label="Select all issues" autoComplete="off"/>
 			</div>
 		);
 		issuesToolbar.append(
@@ -86,8 +86,8 @@ async function init(): Promise<void | false> {
 			conversation.firstElementChild!.prepend(
 				<label className="flex-shrink-0 py-2 pl-3  d-none d-md-block">
 					<input
+						data-check-all-item
 						type="checkbox"
-						data-check-all-item=""
 						className="js-issues-list-check"
 						name="issues[]"
 						value={number}
--- a/source/features/batch-open-conversations.tsx
+++ b/source/features/batch-open-conversations.tsx
@@ -95,9 +95,8 @@ async function init(): Promise<void | false> {
 		);
 
 		const prRows = select.all('.js-issue-row');
-		const prs = prRows.map(getIssueConfig);
-
-		for (const pr of prs) {
+		for (const prRow of prRows) {
+			const pr = getIssueConfig(prRow);
 			pr.prependAt.prepend(
 				<label className="flex-shrink-0 py-2 pl-3  d-none d-md-block">
 					<input
--- a/source/features/batch-open-conversations.tsx
+++ b/source/features/batch-open-conversations.tsx
@@ -34,6 +34,21 @@ function openIssues(): void {
 	});
 }
 
+interface IssueConfig {
+	number: string;
+	prependAt: Element;
+}
+
+function getIssueConfig(prRow: Element): IssueConfig {
+	let match = prRow.id.match(/^issue_(\d*)$/);
+	const number: string = match?.[1] ?? '';
+	return {
+		number,
+		prependAt: prRow.children[0]!,
+	};
+}
+
+
 async function init(): Promise<void | false> {
 	if (!await elementReady('.js-issue-row + .js-issue-row')) {
 		return false;
@@ -63,12 +78,44 @@ async function init(): Promise<void | false> {
 				Open selected
 			</button>
 		);
+	} else {
+		select('#js-issues-toolbar')?.prepend(
+			<div className="mr-3 d-none d-md-block">
+  			<input type="checkbox" data-check-all="" aria-label="Select all issues" autoComplete="off"/>
+		  </div>
+		);
+		select('#js-issues-toolbar')?.append(
+			<div className="table-list-triage flex-auto js-issues-toolbar-triage">
+				<span className="text-gray table-list-header-toggle">
+					<span data-check-all-count="">1</span> selected
+					<button type="button" className="btn-link rgh-batch-open-issues pl-3">Open selected</button>
+				</span>
+			</div>
+		);
+
+		const prElements = select.all('.js-issue-row');
+		const prs = prElements.map(getIssueConfig);
+
+		for (const pr of prs) {
+			pr.prependAt.prepend(
+				<label class="flex-shrink-0 py-2 pl-3  d-none d-md-block">
+					<input
+						type="checkbox"
+						data-check-all-item=""
+						className="js-issues-list-check"
+						name="issues[]"
+						value={pr.number}
+						aria-labelledby={`issue_${pr.number}_link`}
+						autoComplete="off" />
+				</label>
+			);
+		}
 	}
 }
 
 void features.add({
 	id: __filebasename,
-	description: 'Adds a button to open multiple conversations at once in your repos.',
+	description: 'Adds a button to open multiple conversations at once.',
 	screenshot: 'https://user-images.githubusercontent.com/1402241/38084752-4820b0d8-3378-11e8-868c-a1582b16f915.gif'
 }, {
 	include: [
--- a/source/features/batch-open-conversations.tsx
+++ b/source/features/batch-open-conversations.tsx
@@ -40,14 +40,13 @@ interface IssueConfig {
 }
 
 function getIssueConfig(prRow: Element): IssueConfig {
-	let match = prRow.id.match(/^issue_(\d*)$/);
+	const match = /^issue_(\d*)$/.exec(prRow.id);
 	const number: string = match?.[1] ?? '';
 	return {
 		number,
-		prependAt: prRow.children[0]!,
+		prependAt: prRow.children[0]!
 	};
 }
-
 
 async function init(): Promise<void | false> {
 	if (!await elementReady('.js-issue-row + .js-issue-row')) {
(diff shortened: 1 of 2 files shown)


muescha:
🤦 I forget linting. I will fix it tomorrow

source/features/batch-open-conversations.tsx
```diff
@@ -63,12 +78,44 @@ async function init(): Promise<void | false> {
 				Open selected
 			</button>
 		);
+	} else {
+		select('#js-issues-toolbar')?.prepend(
```
fregante, line 82: This is being selected twice. Also don’t use `?.` because it silenced runtime errors, which we need to see when features break. 

I think this needs to be behind a condition, we shouldn’t add this UI when there are no visible issues

source/features/batch-open-conversations.tsx
```diff
@@ -63,12 +78,44 @@ async function init(): Promise<void | false> {
...
+					<span data-check-all-count="">1</span> selected
+					<button type="button" className="btn-link rgh-batch-open-issues pl-3">Open selected</button>
+				</span>
+			</div>
+		);
+
+		const prRows = select.all('.js-issue-row');
+		const prs = prRows.map(getIssueConfig);
+
+		for (const pr of prs) {
```
fregante, line 99: You can inline getIssueConfig in here, no need for 2 separate loops

source/features/batch-open-conversations.tsx
muescha: ### `?.` 

i used the same like in line 60 🤔 
```shell
select('.table-list-header-toggle:not(.states)')?.prepend(
```

### selected twice

i make a variable for it

i guess you suggest the `!`? because if i check the variable for null then it is again a "silent error"

### behind a condition

there is an early exit in line 53 which show the batch function only if there are 2 issues

```javascript
	if (!await elementReady('.js-issue-row + .js-issue-row')) {
		return false;
	}
```

source/features/batch-open-conversations.tsx
muescha: i was adopting the code like this `getPRConfig`:

https://github.com/sindresorhus/refined-github/blob/daf633137ec08e6f5647a6f8bb0b02d5dce58ef9/source/features/conflict-marker.tsx#L50


but less code is always better to maintain 👍 
i do inline it

fregante: That one has to loop because it’s preparing the data for the API call. Here it’s just 2 consecutive loops. Thanks for merging them

source/features/batch-open-conversations.tsx
fregante: I think it’s good now. Thanks

source/features/batch-open-conversations.tsx
muescha: ok - makes sense 👍 
i will inline the function code as well

source/features/batch-open-conversations.tsx
```diff
@@ -63,12 +63,49 @@ async function init(): Promise<void | false> {
...
+			</div>
+		);
+
+		const regExpPrNumber = /^issue_(\d*)$/;
+
+		const prRows = select.all('.js-issue-row');
+		for (const prRow of prRows) {
+			const prNumber = regExpPrNumber.exec(prRow.id)?.[1] ?? '';
+			const prependAt = prRow.children[0]!;
+			prependAt.prepend(
```
fregante, line 89:
```suggestion
			const prependAt = prRow.firstElementChild!.prepend(
```

source/features/batch-open-conversations.tsx
```diff
@@ -63,12 +63,49 @@ async function init(): Promise<void | false> {
...
+					<button type="button" className="btn-link rgh-batch-open-issues pl-3">Open selected</button>
+				</span>
+			</div>
+		);
+
+		const regExpPrNumber = /^issue_(\d*)$/;
+
+		const prRows = select.all('.js-issue-row');
+		for (const prRow of prRows) {
+			const prNumber = regExpPrNumber.exec(prRow.id)?.[1] ?? '';
```
fregante, line 87:
```suggestion
			const number = looseParseInt(prRow.id);
```

source/features/batch-open-conversations.tsx
```diff
@@ -63,12 +63,49 @@ async function init(): Promise<void | false> {
...
+					{' '}
+					<button type="button" className="btn-link rgh-batch-open-issues pl-3">Open selected</button>
+				</span>
+			</div>
+		);
+
+		const regExpPrNumber = /^issue_(\d*)$/;
+
+		const prRows = select.all('.js-issue-row');
+		for (const prRow of prRows) {
```
fregante, line 86: None of these are "PRs", we've started calling them Conversations
```suggestion
		for (const conversation of select.all('.js-issue-row')) {
```

source/features/batch-open-conversations.tsx
```diff
@@ -63,12 +63,49 @@ async function init(): Promise<void | false> {
...
+			const prNumber = regExpPrNumber.exec(prRow.id)?.[1] ?? '';
+			const prependAt = prRow.children[0]!;
+			prependAt.prepend(
+				<label className="flex-shrink-0 py-2 pl-3  d-none d-md-block">
+					<input
+						type="checkbox"
+						data-check-all-item=""
```
fregante, line 93: Does this work?
```suggestion
						data-check-all-item
```

source/features/batch-open-conversations.tsx
```diff
@@ -63,12 +63,49 @@ async function init(): Promise<void | false> {
...
+				</label>
+			);
+		}
 	}
 }
 
 void features.add({
 	id: __filebasename,
-	description: 'Adds a button to open multiple conversations at once in your repos.',
+	description: 'Adds a button to open multiple conversations at once.',
```
fregante, line 108: I'm not a fan of _"Adds a button to..."_ descriptions, let's take this opportunity to improve it
```suggestion
	description: 'Lets you open multiple conversations at once via checkboxes.',
```

source/features/batch-open-conversations.tsx
```diff
@@ -63,12 +63,49 @@ async function init(): Promise<void | false> {
...
+		issuesToolbar.prepend(
+			<div className="mr-3 d-none d-md-block">
+				<input type="checkbox" data-check-all="" aria-label="Select all issues" autoComplete="off"/>
+			</div>
+		);
+		issuesToolbar.append(
+			<div className="table-list-triage flex-auto js-issues-toolbar-triage">
+				<span className="text-gray table-list-header-toggle">
+					<span data-check-all-count="">1</span> selected
+					{' '}
```
fregante, line 77:
```suggestion
					<span data-check-all-count="">1</span>
					{' selected '}
```

source/features/batch-open-conversations.tsx
```diff
@@ -63,12 +63,49 @@ async function init(): Promise<void | false> {
 				Open selected
 			</button>
 		);
+	} else {
+		const issuesToolbar = select('#js-issues-toolbar')!;
+		issuesToolbar.prepend(
+			<div className="mr-3 d-none d-md-block">
+				<input type="checkbox" data-check-all="" aria-label="Select all issues" autoComplete="off"/>
```
fregante, line 70: It it works
```suggestion
				<input type="checkbox" data-check-all aria-label="Select all issues" autoComplete="off"/>
```

source/features/batch-open-conversations.tsx
muescha: ok - i had this version before, but the eslint example suggested more this empty string and in codebase it was already one `{' '}`

muescha: i will change it

source/features/batch-open-conversations.tsx
muescha: with this change i get from normal linter in WebStorm: `Wrong attribute value` 

> HtmlUnknownBooleanAttribute
> Unknown HTML boolean tag attribute
> This inspection highlights HTML non-boolean tag attributes without value as invalid, and lets mark such attributes as Custom to avoid highlighting them as invalid.

the XO linter not complain

muescha:
i marked comments as resolved.
still decition need for the `data-*` tags

source/features/batch-open-conversations.tsx
fregante: Yeah, WebStorm is stricter, but dom-chef should support it, which is what matters.

muescha: and similar is already in codebase:

https://github.com/sindresorhus/refined-github/blob/4248fcf8a430093077c5109e6e5f675c6a068045/source/features/show-associated-branch-prs-on-fork.tsx#L80-L82

i change it

fregante (approved):
Untested, but looks fine so far

fregante:
Works great!

kidonng:
Sometimes the checkboxes didn't show for all issues, but I can't always reproduce it🤷‍♀️
image

yakov116:
Was fixed on master already

PR #3489, opened 2020-08-20
Meta: enable more lint rules

yakov116:
@fregante I went through them all only `@typescript-eslint/require-await': 'error'` am I not sure if they are correct.

### DIFF
--- a/.xo-config.js
+++ b/.xo-config.js
@@ -44,7 +44,27 @@ module.exports = {
 		// Temporarily disabled
 		// TODO: Need to decide what to do with this rule. Either keep it disabled here or remove it from XO.
 		'@typescript-eslint/no-invalid-void-type': 'off', // Reason: https://github.com/sindresorhus/refined-github/pull/3168#discussion_r434479434
-		'unicorn/no-fn-reference-in-iterator': 'off'
+		'unicorn/no-fn-reference-in-iterator': 'off',
+		'@typescript-eslint/no-extra-non-null-assertion': 'error',
+		'@typescript-eslint/consistent-type-definitions':'error',
+		'@typescript-eslint/explicit-function-return-type': [
+			'error',
+			{
+				allowExpressions: true,
+				allowTypedFunctionExpressions: true,
+				allowHigherOrderFunctions: true,
+				allowConciseArrowFunctionExpressionsStartingWithVoid: false
+			}
+		],
+		'@typescript-eslint/explicit-module-boundary-types': [
+			'error',
+			{
+				allowTypedFunctionExpressions: true,
+				allowHigherOrderFunctions: true,
+				allowDirectConstAssertionInArrowFunctions: true,
+				shouldTrackReferences: true
+			}
+		]
 	},
 	// We don't use React, so it can't be automatically detected by the linter
 	settings: {
--- a/.xo-config.js
+++ b/.xo-config.js
@@ -20,7 +20,6 @@ module.exports = {
 
 		'@typescript-eslint/no-extra-non-null-assertion': 'error',
 		'@typescript-eslint/consistent-type-definitions':'error',
-		'@typescript-eslint/require-await': 'error',
 		'@typescript-eslint/explicit-function-return-type': [
 			'error',
 			{
--- a/.xo-config.js
+++ b/.xo-config.js
@@ -47,6 +47,7 @@ module.exports = {
 		'unicorn/no-fn-reference-in-iterator': 'off',
 		'@typescript-eslint/no-extra-non-null-assertion': 'error',
 		'@typescript-eslint/consistent-type-definitions':'error',
+		'@typescript-eslint/require-await': 'error',
 		'@typescript-eslint/explicit-function-return-type': [
 			'error',
 			{
--- a/.xo-config.js
+++ b/.xo-config.js
@@ -10,19 +10,29 @@ module.exports = {
 		'__filebasename'
 	],
 	rules: {
+		'import/no-unassigned-import': 'off',
 		'no-alert': 'off',
 		'no-void': 'off',
-
+		'react/jsx-key': 'off',
+		'unicorn/no-fn-reference-in-iterator': 'off',
 		// Test files are pre-compiled and bundled by rollup
 		'ava/no-ignored-test-files': 'off',
 
-		'react/function-component-definition': [
+		'@typescript-eslint/no-extra-non-null-assertion': 'error',
+		'@typescript-eslint/consistent-type-definitions':'error',
+		'@typescript-eslint/require-await': 'error',
+		'@typescript-eslint/explicit-function-return-type': [
 			'error',
 			{
-				namedComponents: 'function-declaration'
+				allowExpressions: true,
+				allowTypedFunctionExpressions: true,
+				allowHigherOrderFunctions: true,
+				allowConciseArrowFunctionExpressionsStartingWithVoid: false
 			}
 		],
-		'react/jsx-key': 'off',
+		'import/first': 'error',
+		'import/newline-after-import': 'error',
+		'import/no-commonjs': 'error',
 		'import/order': [
 			'error',
 			{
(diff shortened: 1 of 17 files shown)


source/features/add-tag-to-commits.tsx
```diff
@@ -134,9 +134,6 @@ async function init(): Promise<void | false> {
 			// No tags for this commit found in the cache, check in github
 			cached = mergeTags(cached, await getTags(lastCommitOnPage)); // eslint-disable-line no-await-in-loop
 			targetTags = cached[targetCommit];
-		}
-
-		if (!targetTags) {
```
yakov116, line 139: Please Double check that my change is correct and for some reason it was designed like this.

fregante: No, this should be undone, as you can see it tries to set targetTags in 2 ways, so it has to check twice.

source/features/remove-label-faster.tsx
```diff
@@ -36,7 +36,7 @@ async function removeLabelButtonClickHandler(event: delegate.Event<MouseEvent, H
 }
 
 // TODO: set variable via JSX and inline function in `init` after https://github.com/vadimdemedes/dom-chef/issues/66
-function makeRemoveLabelButton(labelName: string, backgroundColor: string) {
+function makeRemoveLabelButton(labelName: string, backgroundColor: string): HTMLButtonElement {
```
fregante, line 39: Unless we need a specific tag (e.g. to access href), just use `JSX.Element` as the type, so you don’t have to use the assertions below

source/features/remove-projects-tab.tsx
```diff
@@ -7,7 +7,7 @@ import * as pageDetect from 'github-url-detection';
 import features from '.';
 import {observeOneMutation} from '../helpers/simplified-element-observer';
 
-function getProjectsTab() {
+async function getProjectsTab(): Promise<HTMLElement | undefined> {
```
fregante, line 10: I’m not particularly fond of the rule that enforces this (“promise functions must be async). If it’s the only case where this happens, fine, but if it requires a lot of similar unnecessary changes let’s remove it

fregante: Ok, good as is.

fregante:
Everything else looks good so far. Let me know when you’re done making changes, I might have other rule suggestions

yakov116:
I am good except for my first comment

fregante:
Ok, can you sort them?

1. Group by state (2 groups: either disabled from the shared config, or new rule)
2. Alphabetically in each group

```js
rules: {
	/* Disable rules */
	a: "off",
	b: "off",

	/* Add rules */
	c: "error",
	d: "error",
}
```

.xo-config.js
```diff
@@ -44,7 +44,28 @@ module.exports = {
 		// Temporarily disabled
 		// TODO: Need to decide what to do with this rule. Either keep it disabled here or remove it from XO.
 		'@typescript-eslint/no-invalid-void-type': 'off', // Reason: https://github.com/sindresorhus/refined-github/pull/3168#discussion_r434479434
-		'unicorn/no-fn-reference-in-iterator': 'off'
+		'unicorn/no-fn-reference-in-iterator': 'off',
+		'@typescript-eslint/no-extra-non-null-assertion': 'error',
+		'@typescript-eslint/consistent-type-definitions':'error',
+		'@typescript-eslint/require-await': 'error',
```
sindresorhus, line 50: FYI: https://github.com/xojs/eslint-config-xo-typescript/blob/7842d05deaadff1ced9a46f3e56d786e7b2922d2/index.js#L443

.xo-config.js
```diff
@@ -44,7 +44,28 @@ module.exports = {
 		// Temporarily disabled
 		// TODO: Need to decide what to do with this rule. Either keep it disabled here or remove it from XO.
 		'@typescript-eslint/no-invalid-void-type': 'off', // Reason: https://github.com/sindresorhus/refined-github/pull/3168#discussion_r434479434
-		'unicorn/no-fn-reference-in-iterator': 'off'
+		'unicorn/no-fn-reference-in-iterator': 'off',
+		'@typescript-eslint/no-extra-non-null-assertion': 'error',
+		'@typescript-eslint/consistent-type-definitions':'error',
```
sindresorhus, line 49: While it might default to `interface`, this is not documented and also not clear.

Also, any reasoning to defaulting to `interface`?

https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/consistent-type-definitions.md

The only benefit of `interface` is declaration merging, but `type` has other benefits too: https://stackoverflow.com/a/52682220

.xo-config.js
```diff
@@ -44,7 +44,28 @@ module.exports = {
 		// Temporarily disabled
 		// TODO: Need to decide what to do with this rule. Either keep it disabled here or remove it from XO.
 		'@typescript-eslint/no-invalid-void-type': 'off', // Reason: https://github.com/sindresorhus/refined-github/pull/3168#discussion_r434479434
```
sindresorhus, line 46: This was disabled in the shared config.

.xo-config.js
```diff
@@ -44,7 +44,28 @@ module.exports = {
...
+		'@typescript-eslint/explicit-function-return-type': [
+			'error',
+			{
+				allowExpressions: true,
+				allowTypedFunctionExpressions: true,
+				allowHigherOrderFunctions: true,
+				allowConciseArrowFunctionExpressionsStartingWithVoid: false
+			}
+		],
+		'@typescript-eslint/explicit-module-boundary-types': [
```
sindresorhus, line 60: Just make sure you're ok with this: https://github.com/xojs/eslint-config-xo-typescript/blob/7842d05deaadff1ced9a46f3e56d786e7b2922d2/index.js#L136-L138

sindresorhus (changes requested):

.xo-config.js
fregante: According to [its docs](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/explicit-module-boundary-types.md#when-not-to-use-it), this rule is only a "less strict" version of `explicit-function-return-type`, which we already use. Right?

If so, it should be dropped from here and from XO’s comments

.xo-config.js
fregante: > but `type` has other benefits too

The only benefit I see is that you can use primitives, like `type S = string`, but this rule doesn't seem to prevent us from using that.

From the replacements I see in this PR, it only affects `type`s that can as well be `interface`s.

In my eyes, the difference between the two:

- `type x = {};` looks like a declaration
- `interface x {}` looks like a definition

Minimal difference, really, it's just about consistency

.xo-config.js
fregante: I remember this. It looks like the lint is failing because of this one.

If it works painlessly on the current code, we can enable it.

.xo-config.js
sindresorhus: > The only benefit I see is that you can use primitives, like type S = string, but this rule doesn't seem to prevent us from using that.

Sure, but if you ever need that, you're back to inconsistency. Which is why I slightly prefer `type`, but it doesn't matter much. 🤷‍♂️

Kinda annoying that TS has such similar duplicate things.

fregante: > any reasoning to defaulting to `interface`

I like definitions over declarations. The former looks cleaner to me:

```ts
interface X {}
type X = {};
```

Kinda like I prefer `class X {}` over `const X = class {}`

.xo-config.js
yakov116: This was the only one I was having issues with (hence my first comment)

fregante (approved):

fregante:
If you want to update the dependencies, this would be useful: https://github.com/fregante/webext-storage-cache/pull/20

PR #3491, opened 2020-08-20
Make `clean-pinned-issues` slightly more visible

fregante:
I feel that with `clean-pinned-issues`, the pinned issues have become _less visible_, which was not the intent. Here I just make the border thicker. Alternative suggestions welcome.

### DIFF
--- a/source/features/clean-pinned-issues.css
+++ b/source/features/clean-pinned-issues.css
@@ -11,7 +11,7 @@
 		margin-bottom: 20px;
 
 		/* Rounded table border https://stackoverflow.com/a/2586780/288906 */
-		box-shadow: 0 0 0 1px #d1d5da;
+		box-shadow: 0 0 0 2px #d1d5da;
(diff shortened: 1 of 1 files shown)


fregante:
Or...

## Alternative before / after

notlmn:
How about yellow?

kidonng:
I can see color is already used for active tabs, but in my opinion, it doesn't apply to pinned issues because they are not "active" or something that should be "highlighted".

fregante:
> something that should be "highlighted".

Isn't that the point of "pinned" issues? It's to highlight them. But I suppose not _that_ highlighted

> How about yellow?

Yellow on white either is invisible or looks "in progress" like for PRs

kidonng:
> But I suppose not _that_ highlighted

👍

fregante:
> If color is needed, how about `$border-gray-darker` `#444d56`

That's a little ugly TBH :D

PR #3497, opened 2020-08-24
Fix `latest-tag-button` position

### LINKED ISSUE #3493: Wrong position of `latest-tag-button`
Extension version: 20.8.20
Link: Any page with `latest-tag-button` (like https://github.com/sindresorhus/refined-github)
Screenshot: 
image

kidonng:
Fix #3493

Test: https://github.com/sindresorhus/refined-github

### DIFF
--- a/source/features/latest-tag-button.tsx
+++ b/source/features/latest-tag-button.tsx
@@ -3,7 +3,7 @@ import React from 'dom-chef';
 import cache from 'webext-storage-cache';
 import TagIcon from 'octicon/tag.svg';
 import DiffIcon from 'octicon/diff.svg';
-import elementReady from 'element-ready';
+import select from 'select-dom';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
@@ -84,11 +84,6 @@ async function init(): Promise<false | void> {
 		return false;
 	}
 
-	const breadcrumb = await elementReady('#branch-select-menu');
-	if (!breadcrumb) {
-		return;
-	}
-
 	const currentBranch = getCurrentBranch();
 	const url = new GitHubURL(location.href);
 	url.assign({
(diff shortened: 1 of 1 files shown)


source/features/latest-tag-button.tsx
```diff
@@ -102,7 +97,7 @@ async function init(): Promise<false | void> {
 		</a>
 	);
 
-	breadcrumb.after(link);
+	select('#branch-select-menu')!.parentElement!.after(link);
```
kidonng, line 100: This element always exists, no need to check

source/features/latest-tag-button.tsx
```diff
@@ -139,7 +134,7 @@ async function init(): Promise<false | void> {
 		link.setAttribute('aria-label', 'Visit the latest release');
 	}
 
-	link.classList.add('tooltipped', 'tooltipped-ne');
+	link.classList.add('tooltipped', 'tooltipped-ne', 'rgh-latest-tag-button');
```
kidonng, line 137: This change isn't related to bugfix but rather #3356. With current changes it is really inconvenient to find a succinct selector to customize the button's size, so I think adding a class won't hurt.

source/features/latest-tag-button.tsx
fregante: Looks right. We used to use an element that sometimes didn't exist; the selector was changed but not the logic: https://github.com/sindresorhus/refined-github/pull/3353/files

fregante (approved):

source/features/latest-tag-button.tsx
kidonng: I should have added it in this line as well 😅 https://github.com/sindresorhus/refined-github/pull/3497/files#diff-5664f505bec7f184c9588cc203936104R107

PR #3504, opened 2020-08-26
Add `render-conversation-templates` feature

### LINKED ISSUE #3365: Render issue/PR templates
EDIT: this has been partially implemented, but your help is needed to improve it as described in: https://github.com/sindresorhus/refined-github/pull/3504

You can pick up where I left off

---

Issue/PR templates are easy to dismiss and the links in it are not clickable.

What if we rendered them? (via GitHub’s "preview comment" feature)

<table><th>Before<th>After
<tr><td valign=top>

<td>
</table>

To keep in mind:

- this won't work for all templates (like https://github.com/babel/babel/issues/new/choose), but it may be worth trying
	- it could be opt-in by each template, for example _"Only if it starts with **Instructions:** or similar keywords"_
- we might have to display the render **under** the field to avoid moving the field down
- we probably can't delete the field’s content because some templates have to be filled by the user

fregante:
Closes #3365 (?)

I need help on this feature. There's currently a major issue with the current implementation:

**It expects a very specific format that means it can only be opt-in by maintainers, hugely limiting its usefulness.**

This is because:

- The template has to look and work well **without** Refined GitHub. This is fundamental. 
- The user can no longer manually delete content from the template, so any text that isn't an HTML comment will be shown.
- The rendered form has to make sense, i.e. ideally the field’s description should be **above** the field, not half above half below (colons in the middle of a line complicates things). The current version actually includes the description as the `placeholder`.

Currently I'm using this format, which follows every constraint:

```
<!-- Please follow the template -->
<!-- Thanks for contributing! 🍄 -->

**LINKED ISSUES:**
<!-- Does this PR fix an existing issue? Write something like `Closes #10` -->

**TEST URLS:**
<!-- Add some REAL test URLs -->

**SCREENSHOT**
<!-- Include a screenshot here if your PR makes visual changes -->
```

Notice the lack of `:` after SCREENSHOT. That's because users can't upload images directly into textareas, so this will look like it's pointing to the main Markdown editor.

Possible solutions/additions:

- [ ] An "exit"/"something's wrong" button that disables the feature ("forever" in the cache, for the current repo) and just restores the original plain template in the Markdown editor.
- [ ] An "apply" button that the user has to press before creating the issue/PR, so they have a chance to review.
- [ ] Every field could accept file uploads to avoid the aforementioned SCREENSHOTS awkwardness.

Also currently missing:

- [ ] Support for the Preview tab
- [ ] Support for checkboxes. An "apply" button could make them editable, but users will likely forget them.

Next simple changes that remove the **need** for the `<!-- Please follow the template -->` opener:
- [ ] When opening a PR, GitHub will automatically include the commit’s body in the textarea, making it difficult for RGH to know where user content ends and the template starts. The opener could be replaced by just looking for the commit’s body into the DOM and removing it from the field.

I don't think I'll have time to finish this PR very soon and it'd be great if anyone else could take it for a spin. **Feel free to open test/discussion PRs against this branch.**


source/manifest.json
```diff
@@ -4,7 +4,7 @@
 	"description": "Simplifies the GitHub interface and adds useful features",
 	"homepage_url": "https://github.com/sindresorhus/refined-github",
 	"manifest_version": 2,
-	"minimum_chrome_version": "80",
+	"minimum_chrome_version": "85",
```
fregante, line 7: Due to String#replaceAll

source/features/render-conversation-template.tsx
```diff
@@ -0,0 +1,92 @@
...
+}
+
+void features.add({
+	id: __filebasename
+}, {
+	include: [
+		pageDetect.isNewIssue,
+		pageDetect.isQuickPR,
+		// TODO: Merge with above
+		() => pageDetect.isCompare() && location.search.includes('expand=1')
```
fregante, line 89: cc @yakov116 I think we need a unified `isNewPR`

fregante:
To be clear, I'm looking to safely enable this feature on every repository.

source/features/render-conversation-template.tsx
yakov116: :+1: I can send a pr tomorrow. 

Does expand always happen on new pr's? I always used a dom based detection.

fregante: Not sure. Here's an example URL:

```
https://github.com/sindresorhus/refined-github/compare/incremental-tag-changelog-link?expand=1
```

yakov116: https://github.com/sindresorhus/refined-github/compare/master...master?expand=1

This does not mean that I can create a PR. I think this is why we called it isCompare since we cant tell if its a new PR.

fregante: > This does not mean that I can create a PR

True, but that's also true for `isQuickPR` if a PR already exists.

Should I just use `isCompare`?

fregante:
Feel free to discuss or finishing up this PR

fregante:
I’ll pick this up again and limit it to this repo. The code exists and seems to work. At least we can get something out of it

yakov116:
If up your doing this then on bug it should include the extention version number.

fregante:
One day…

PR #3505, opened 2020-08-27
Refactor `highlight-collaborators-and-own-conversations` to use `selector-observer`

yakov116:
#2671 

Test on

### DIFF
--- a/source/features/highlight-collaborators-and-own-conversations.tsx
+++ b/source/features/highlight-collaborators-and-own-conversations.tsx
@@ -1,6 +1,7 @@
 import './highlight-collaborators-and-own-conversations.css';
 import cache from 'webext-storage-cache';
 import select from 'select-dom';
+import {observe} from 'selector-observer';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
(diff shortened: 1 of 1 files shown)


source/features/highlight-collaborators-and-own-conversations.tsx
```diff
@@ -53,6 +50,10 @@ void features.add({
 	include: [
 		pageDetect.isRepoConversationList
 	],
+	exclude: [
+		() => select.exists('.blankslate')
```
fregante, line 54: We have this on github-url-detection

yakov116: Only isEmptyRepoRoot

fregante: We should probably drop that one and just use this function everywhere

fregante (approved):

source/features/highlight-collaborators-and-own-conversations.tsx
yakov116: https://github.com/fregante/github-url-detection/commit/0c36c4c08636174cc039b077680ff90acb3b5d37

yakov116: @fregante how to proceed?

fregante: `npm deprecate *`

fregante (changes requested):
👎  This doesn't seem to have benefits, according to https://github.com/sindresorhus/refined-github/issues/3084#issuecomment-683345700 — that is, this is not a "ajaxed hot-area"

Thoughts?

yakov116:
There is one benefit here that the request is sent before dom ready and then applied as soon as dom ready. I could do await dom ready, that would accomplish the same thing.

fregante:
Indeed, but that's unrelated to `selector-observer`. You can send a PR just for that

yakov116:
In other words you mean don't wait for dom ready and then after the request do await dom loaded?

fregante:
Correct

PR #3509, opened 2020-08-27
Add `isRepoSearch` into `parse-backticks`

kidonng:
This is tougher than I previously thought (*just add a selector is fine*) because existing functions won't handle elements inside, while the keyword should remain highlighted on the search page. I end up with:

- ~~Adding a workaround (`child.textContent = child.textContent`)~~
- ~~Rehighlight the search keywords~~

~~The code may not look pretty but it works, though only for **issues/PRs**. **Commits** has a deeper level of elements to couple with, which I may come up with an idea later.~~

Suggestions/testing welcome.

1. LINKED ISSUES:
   None

2. TEST URLS:
- Commit search: https://github.com/sindresorhus/refined-github/search?q=latest+button&type=Commits
- Issue search: https://github.com/sindresorhus/refined-github/search?q=latest+button+is%3Aissue&type=Issues
- PR search: https://github.com/sindresorhus/refined-github/search?q=latest+button+is%3Aissue+is%3Apr&type=Issues

3. SCREENSHOT:
Before:
image
After:
image

### DIFF
--- a/source/features/latest-tag-button.tsx
+++ b/source/features/latest-tag-button.tsx
@@ -96,7 +96,7 @@ async function init(): Promise<false | void> {
 	});
 
 	const link = (
-		<a className="btn btn-sm btn-outline ml-2 flex-self-center" href={String(url)}>
+		<a className="btn btn-sm btn-outline ml-2 flex-self-center rgh-latest-tag-button" href={String(url)}>
 			<TagIcon/>
 		</a>
 	);
@@ -138,7 +138,7 @@ async function init(): Promise<false | void> {
 		link.setAttribute('aria-label', 'Visit the latest release');
 	}
 
-	link.classList.add('tooltipped', 'tooltipped-ne', 'rgh-latest-tag-button');
+	link.classList.add('tooltipped', 'tooltipped-ne');
 }
 
 void features.add({
--- a/source/features/parse-backticks.tsx
+++ b/source/features/parse-backticks.tsx
@@ -47,7 +47,6 @@ function init(): void {
 	// `isRepoSearch`
 	observe('#issue_search_results .f4:not(.rgh-backticks-already-parsed)', {
 		add(element) {
-			// Prepare
 			const child = element.firstElementChild!;
 			const keywords = [...child.querySelectorAll('em')].map(element => element.textContent);
 			// Combine text content to enable backticks parsing
--- a/source/features/parse-backticks.tsx
+++ b/source/features/parse-backticks.tsx
@@ -5,6 +5,7 @@ import zipTextNodes from 'zip-text-nodes';
 
 import features from '.';
 import {parseBackticks} from '../github-helpers/dom-formatters';
+import parseBackticksCore from '../github-helpers/parse-backticks';
 
 function init(): void {
 	const selectors = [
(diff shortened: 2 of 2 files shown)


source/features/latest-tag-button.tsx
```diff
@@ -96,7 +96,7 @@ async function init(): Promise<false | void> {
 	});
 
 	const link = (
-		<a className="btn btn-sm btn-outline ml-2 flex-self-center" href={String(url)}>
+		<a className="btn btn-sm btn-outline ml-2 flex-self-center rgh-latest-tag-button" href={String(url)}>
```
kidonng, line 99: Again, changes in this file are not really related to this PR (sorry if it bothers😅). I added the class into the wrong position in the previous PR (https://github.com/sindresorhus/refined-github/pull/3497/files#r478513136), and I don't feel like make a separate PR for this trivial change.

source/features/parse-backticks.tsx
```diff
@@ -41,6 +43,40 @@ function init(): void {
...
 		}
 	});
+
+	// `isRepoSearch`
+	observe('#issue_search_results .f4:not(.rgh-backticks-already-parsed)', {
+		add(element) {
+			const child = element.firstElementChild!;
+			const keywords = [...child.querySelectorAll('em')].map(element => element.textContent);
+			// Combine text content to enable backticks parsing
+			child.textContent = `${child.textContent!}`;
```
kidonng, line 53: Can't just do `child.textContent = child.textContent` because the linter like it go complicated

source/features/parse-backticks.tsx
```diff
@@ -41,6 +43,40 @@ function init(): void {
...
+
+			// Parse backticks
+			element.classList.add('rgh-backticks-already-parsed');
+			parseBackticks(element);
+
+			// Rehighlight search keywords
+			for (const keyword of keywords) {
```
kidonng, line 60: This part is obviously borrowed from the two backtick parsing functions, just want to say they are clean and effective ❤

fregante:
> because existing functions won't handle elements inside

Try again, because `zip-text-nodes` does exactly that, and it's part of the "dom formatters" file. What you're doing here is rewriting most of the parseBackticks function that already exists.

kidonng:
> because `zip-text-nodes` does exactly that

Didn't noticed it exist 👍 exactly what this feature needs

fregante:
Almost! What I don’t understand is why the regular dom-formatter doesn’t work. Doesn’t it already include zip-text-nodes?

kidonng:
Not in the `parseBackticks` function

https://github.com/sindresorhus/refined-github/blob/de6dfe894ecfa2fd41c144f76385ce651b5456f6/source/github-helpers/dom-formatters.ts#L67-L75

fregante:
I see why. `parseBackticks` took the more lightweight approach of just parsing each textNode separately. Perhaps it should be changed to match the style of the 2 other DOM formatters (so it always includes zipTextNodes). But it needs to be tested in every function that uses it

fregante:
Alternatively… this could just stay the way it is, without affecting the performance of other features. It's kinda dirty because that dom-formatter is different from the other 2… but it's also faster. Sigh. 

I's fine as is

fregante (approved):
:woot: :woot: for 4 lines of code!

source/features/parse-backticks.tsx
```diff
@@ -41,6 +43,14 @@ function init(): void {
 			parseBackticks(element);
 		}
 	});
+
+	// `isRepoSearch`
```
fregante, line 47: Perhaps it needs to specify _why_ this needs its own version
```suggestion
	// `isRepoSearch` might highlight keywords inside backticks, breaking the regular dom-formatter #3509
```

PR #3513, opened 2020-08-28
Refactor `clone-branch` to use `selector-observer`

yakov116:
1. LINKED ISSUES:
   #2671

2. TEST URLS:
   https://github.com/sindresorhus/refined-github/branches

3. SCREENSHOT:
   None

### DIFF
--- a/source/features/clone-branch.tsx
+++ b/source/features/clone-branch.tsx
@@ -103,6 +103,5 @@ void features.add({
 	include: [
 		pageDetect.isBranches
 	],
-	waitForDomReady: false,
 	init: onetime(init)
 });
--- a/source/features/clone-branch.tsx
+++ b/source/features/clone-branch.tsx
@@ -1,6 +1,9 @@
 import React from 'dom-chef';
 import select from 'select-dom';
+import onetime from 'onetime';
 import delegate from 'delegate-it';
+import {observe} from 'selector-observer';
+import elementReady from 'element-ready';
 import GitBranchIcon from 'octicon/git-branch.svg';
 import * as pageDetect from 'github-url-detection';
 import * as textFieldEdit from 'text-field-edit';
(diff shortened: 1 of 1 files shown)


source/features/clone-branch.tsx
```diff
@@ -71,27 +73,29 @@ async function cloneBranch({delegateTarget: cloneButton}: delegate.Event<MouseEv
...
 	await api.expectToken();
-	const deleteIcons = select.all([
+	const deleteIconClass = [
 		'branch-filter-item-controller .octicon-trashcan', // Pre "Repository refresh" layout
 		'branch-filter-item .octicon-trashcan'
-	]);
+	].join();
 	// If the user does not have rights to delete a branch, they can’t create one either
-	if (deleteIcons.length === 0) {
+	if (!await elementReady(deleteIconClass)) {
```
fregante, line 81: We probably don’t need both elementReady and observe. Let’s simplify

yakov116: I did it for the return false. Drop it?

fregante: > I did it for the return false. Drop it?

Yeah.

`return false` generally nice to have, but not worth it if we have to keep otherwise-completely-useless code.

source/features/clone-branch.tsx
```diff
@@ -104,10 +103,6 @@ void features.add({
 	include: [
 		pageDetect.isBranches
 	],
-	init() {
-		observeElement([
-			'[data-target="branch-filter-controller.results"]', // Pre "Repository refresh" layout
-			'[data-target="branch-filter.result"]'
-		].join(), init);
-	}
+	waitForDomReady: false,
```
fregante, line 106: Since we can't avoid this, we should drop it
```suggestion
```

fregante (approved):

PR #3514, opened 2020-08-28
Refactor `conversation-links-on-repo-list` to use `selector-observer` 

yakov116:
1. LINKED ISSUES:
   #2671

2. TEST URLS:
   https://github.com/search?q=user%3Asindresorhus+sindre
https://github.com/sindresorhus?tab=repositories&type=source

3. SCREENSHOT:
   None


fregante (approved):
Looks right

fregante:
Actually… it’s not faster. I found out that selector-observer always waits for dom-ready. This sort of limits the usefulness of this package.

This is kind of a problem for some features. I wonder if we need to find another package. 

It makes sense that it waits (for performance reasons) but also this means that features a no longer instant.

yakov116:
Your right. https://github.com/josh/selector-observer/blob/master/lib/ready.js 

I have however found that on Ajax requests or when we were doing observelement on init this was faster. But otherwise it only makes it simpler.

fregante:
I'm failing to see the purpose of this change. I don't think that the page is particularly ajaxed

fregante (changes requested):
👎  This doesn't seem to have benefits, according to https://github.com/sindresorhus/refined-github/issues/3084#issuecomment-683345700 — that is, this is not a "ajaxed hot-area"

Thoughts?

fregante:
I guess this _does_ have `observeElement('#user-repositories-list', init)` and the initialization is slightly shorter. It could be ok

source/features/conversation-links-on-repo-lists.tsx
```diff
@@ -1,42 +1,43 @@
...
 import PullRequestIcon from 'octicon/git-pull-request.svg';
 
 import features from '.';
-import observeElement from '../helpers/simplified-element-observer';
 
 function init(): void {
-	const repositories = select.all<HTMLAnchorElement>([
+	observe([
 		'[itemprop="name codeRepository"]:not(.rgh-discussion-links)', // `isUserProfileRepoTab`
 		'[data-hydro-click*=\'"model_name":"Repository"\']' // `isGlobalSearchResults`
```
fregante, line 14: Also needs `:not(.rgh-discussion-links)`

fregante (approved):

fregante:
Congrats on your first merged PR 😸

yakov116:
Thanks!

source/features/conversation-links-on-repo-lists.tsx
```diff
@@ -46,15 +47,8 @@ void features.add({
 }, {
 	include: [
 		pageDetect.isUserProfileRepoTab,
-		pageDetect.isGlobalSearchResults
-	],
-	init
-}, {
-	include: [
+		pageDetect.isGlobalSearchResults,
 		pageDetect.isUserProfileRepoTab
```
jerone, line 51: Duplicate page detect.

fregante: I feel a Lint PR coming

yakov116: @fregante I am going to send a lint pr but I only have 2 small things. Have any lint in mind??

fregante: Can you cleanup some feature names?

- `cleanup` -> `clean`
- drop `add-`

Also if you want to:
- Ensure descriptions always end with `.` (in source or readme)
- Double-check if TODOs have been resolved or can easily be resolved
- Check whether features still work, picking some at random or in any order you'd like

yakov116: > * Change all cache durations to be on one line, i.e. `maxAge: {hours: 1}`

Haha I made then on new lines since I though it was easier to read

PR #3225, opened 2020-06-12
Restore `repo-age` in "Repository refresh" beta

FloEdelmann:
References #3081, follows <https://github.com/sindresorhus/refined-github/pull/3201#issuecomment-641948774>.

Test this on any repo, e.g. <https://github.com/sindresorhus/refined-github>. Note that you should clear the cache to test if the commit link can be extracted correctly.

Unbenannt

PR #3516, opened 2020-08-28
Prevent `linkify-urls` from running multiple times

### LINKED ISSUE #3226: `linkify-code` runs multiple times on the same URL
This became visible when running [the `github-issue-link-status` extension](https://github.com/fregante/github-issue-link-status):

Test: https://github.com/sindresorhus/refined-github/pull/3225#discussion_r439535538

kidonng:
Both `.blob-wrapper` and `.comment-body` matches a code block in review comments, so the linkify functions got run for multiple times.

1. LINKED ISSUES:
   Fix #3226

2. TEST URLS:
   https://github.com/sindresorhus/refined-github/pull/3225#discussion_r439535538

3. SCREENSHOT:
   Before: image
   After: 
image

### DIFF
--- a/source/features/linkify-code.tsx
+++ b/source/features/linkify-code.tsx
@@ -9,8 +9,8 @@ import {linkifiedURLClass, linkifyURLs, linkifyIssues} from '../github-helpers/d
 function init(): void {
 	const selectors = [
 		'.js-blob-wrapper',
-		':not(.js-suggested-changes-blob) > .blob-wrapper',
-		'.comment-body',
+		'.blob-wrapper',
+		'.comment-body.d-block',
 		'.blob-expanded'
 	].map(selector => selector + `:not(.${linkifiedURLClass})`).join();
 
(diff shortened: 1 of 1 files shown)


fregante:
I think we should limit the comment-body selector instead. It should only work on `code` (I think)

kidonng:
Do you mean it should apply to standalone comments (i.e. not review comments)?

fregante:
I think this feature needs to be slightly rewritten to avoid this duplicate `select.all` usage

Can you try picking up https://github.com/sindresorhus/refined-github/pull/3091 instead? It should be better because:

- it only affects visible code, also used in [#2737](https://github.com/sindresorhus/refined-github/pull/2737/commits/f6f9f8b1c5de9ec9947469876fc156904930bb46)
- it selects the specific elements instead of "observe their parent + select.all inside"
- it should also solve https://github.com/sindresorhus/refined-github/issues/2091, which was for `linkify-code` but also affects `linkify-urls` afaik

PR #3389, opened 2020-07-23
Fix `dim-bots` selectors

kidonng:
Fix #3386 

I'm wondering should I preserve old selectors (for enterprise)

cygvoCHJtO

PR #3519, opened 2020-08-28
Correctly hide more information in `dim-bots` 

kidonng:
- The selectors added in #2864 are a mess now
- They didn't hide the comment link
- They didn't correctly hide the meta if there is commit message (added in #3389)

1. LINKED ISSUES:
   See above

2. TEST URLS:
   - https://github.com/kidonng/readhub/commits/master
   - https://github.com/bors-ng/bors-ng/commits/master

3. SCREENSHOT:
   Before: 
image
image

   After: 
image
  (Also note the height difference)
image

### DIFF
--- a/source/features/dim-bots.css
+++ b/source/features/dim-bots.css
@@ -7,11 +7,9 @@ Commits that have a long commit message will not be dimmed after the user uncoll
 .rgh-dim-bot .commit-meta, /* Pre "Repository refresh" layout */
 .rgh-dim-bot .mb-1,
 .rgh-dim-bot .mb-1 + .d-flex,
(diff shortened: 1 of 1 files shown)


fregante:
Aren’t comments potentially important though? If a leave a comment on a bot commit, this will also “fade out”

kidonng:
> Aren’t comments potentially important though?

I was thinking the same thing... but usually commenting on bot's commits is uncommon (in my case it's the Vercel integration)

> If a leave a comment on a bot commit, this will also “fade out”

We can make it fade out if there are comments, but the current style is confusing.

fregante (approved):

PR #3530, opened 2020-09-05
Meta: Ability to remotely disable broken features

### LINKED ISSUE #3529: Ability to remotely disable broken features
Crazy idea, but occasionally we have very breaking bugs that cause frustration: https://github.com/sindresorhus/refined-github/issues/3527 https://github.com/sindresorhus/refined-github/issues/2357 https://github.com/sindresorhus/refined-github/issues/1805 https://github.com/sindresorhus/refined-github/pull/1148 https://github.com/sindresorhus/refined-github/issues/2700 [etc](https://github.com/sindresorhus/refined-github/issues?q=label%3Abug+is%3Aclosed+sort%3Areactions-%2B1-desc)

These are often fixed quickly, but users have to suffer it until the browser updates the extension, which could be a day later.

We could have a mechanism that pings this repo to check whether a feature should be temporarily disabled. For example inside [`globalReady`](https://github.com/sindresorhus/refined-github/blob/e1d836a2f542bb77e9510e5c5aaed2af8c572c91/source/features/index.tsx#L105)

```js
const options = await optionsStorage.getAll();
const hotfix = await cache.get('hotfix');
fetchHotfixesAsynchronously(); // `cache.function` to be run up to ~4 times a day?
Object.assign(options, hotfix);
```

`fetchHotfixesAsynchronously` could fetch the content of a dedicated, locked issue with a content like:

```js
false
```

or

```js
{"cleanup-repo-filelist-actions": false, "unaffected": "20.9.4"}
```

`fetchHotfixesAsynchronously` would use `looseVersionCompare` to compare `unaffected` with the current version and either store `{'cleanup-repo-filelist-actions': false}` or `{}` accordingly.

## Notes

Performance: up to 4 fetches a day should make it lightweight enough
Privacy: it would only fetch public content on GitHub, only if you open GitHub
Security: it would only `GET` JSON content from api.github.com, it doesn't `POST` anything

shinenelson:
Closes #3529.

`hotfix.json` has been added to the repository tree to
indicate the existence of the file in the repository. I have
set the value to `{ false }` for now, but if that might
introduce confusion to contributors, we could delete the file
altogether.

Note : since this is a background feature, I couldn't figure
out a way to test these changes.


source/features/index.tsx
```diff
@@ -150,6 +157,29 @@ const setupPageLoad = async (id: FeatureID, config: InternalRunConfig): Promise<
 	}
 };
 
+// Fetch hotfix releases asynchronously
+// every 6 hours ( 4 times a day )
+// this is achieved by GETting a JSON file
+// from the hotfix branch and comparing
+// with the current version of extension
```
fregante, line 164: These comments are rephrasing the whole function line by line, that's not useful. Comments should either explain what's going on if code isn't immediately clear or _why_ something is being done.

source/features/index.tsx
```diff
@@ -104,6 +106,11 @@ const globalReady: Promise<RGHOptions> = new Promise(async resolve => {
 	// Options defaults
 	const options = await optionsStorage.getAll();
 
+	// Fetch hotfixes asynchronously
```
fregante, line 109:
```suggestion
	// If features are remotely marked as "seriously breaking" by the maintainers, disable them without having to wait for proper updates to propagate #3529
```

source/features/index.tsx
```diff
@@ -150,6 +157,29 @@ const setupPageLoad = async (id: FeatureID, config: InternalRunConfig): Promise<
...
+	const response = await fetch('https://raw.githubusercontent.com/sindresorhus/refined-github/hotfix/hotfix.json');
+	const hotfixes = await response.json();
+
+	if (!hotfixes) {
+		return {};
+	}
+
+	if (hotfixes.unaffected) {
+		const currentVersion = browser.runtime.getManifest().version;
+		if (looseVersionCompare(hotfixes.unaffected, currentVersion) >= 0) {
```
fregante, line 175: I think this is the correct condition
```suggestion
		if (looseVersionCompare(hotfixes.unaffected, currentVersion) < 1) {
```

source/features/index.tsx
```diff
@@ -150,6 +157,29 @@ const setupPageLoad = async (id: FeatureID, config: InternalRunConfig): Promise<
...
+			return {};
+		}
+	}
+
+	return hotfixes;
+}, {maxAge: {hours: 6}});
```
fregante, line 181: This needs the key I specified, or else it won't be cached as `hotfix`
```suggestion
}, {
	maxAge: {hours: 6},
	cacheKey: 'hotfix'
});
```

source/features/index.tsx
```diff
@@ -150,6 +157,29 @@ const setupPageLoad = async (id: FeatureID, config: InternalRunConfig): Promise<
 	}
 };
 
+// Fetch hotfix releases asynchronously
+// every 6 hours ( 4 times a day )
+// this is achieved by GETting a JSON file
+// from the hotfix branch and comparing
+// with the current version of extension
+const fetchHotfixesAsynchronously = cache.function(async () => {
```
fregante, line 165:
```suggestion
const checkForHotfixes = cache.function(async () => {
```

source/features/index.tsx
```diff
@@ -1,9 +1,11 @@
 import React from 'dom-chef';
 import select from 'select-dom';
 import domLoaded from 'dom-loaded';
+import cache from 'webext-storage-cache';
```
fregante, line 4: I try to align the "from" so the module names are easier to follow/read
```suggestion
import React from 'dom-chef';
import cache from 'webext-storage-cache';
import select from 'select-dom';
import domLoaded from 'dom-loaded';
```

source/features/index.tsx
```diff
@@ -1,9 +1,11 @@
...
 import stripIndent from 'strip-indent';
 import {Promisable} from 'type-fest';
 import elementReady from 'element-ready';
+import looseVersionCompare from 'tiny-version-compare';
```
fregante, line 8:
```suggestion
import compareVersions from 'tiny-version-compare';
```

This matches https://github.com/sindresorhus/refined-github/blob/0a58069c118954556d667c661d8cee31ff826400/source/github-helpers/index.ts#L3

source/background.ts
```diff
@@ -34,6 +35,11 @@ browser.runtime.onInstalled.addListener(async ({reason}) => {
 			active: false
 		});
 	}
+
+	// Clear hotfix cache after update
+	if (reason === 'update') {
+		await cache.delete('hotfix');
+	}
```
fregante, line 42:
```suggestion
	// Hope that the feature was fixed in this version
	await cache.delete('hotfix');
```

source/features/index.tsx
```diff
@@ -104,6 +106,11 @@ const globalReady: Promise<RGHOptions> = new Promise(async resolve => {
 	// Options defaults
 	const options = await optionsStorage.getAll();
 
+	// Fetch hotfixes asynchronously
+	const hotfix = await cache.get('hotfix');
+	await fetchHotfixesAsynchronously();
```
fregante, line 111: The function is not meant to be awaited. 


```suggestion
	checkForHotfixes();
```

[cache.function](https://github.com/fregante/webext-storage-cache#cachefunctiongetter-options) is a shortcut that either gets the value from the cache (instant) or calls the passed function to generate the value (might take longer than 500ms), cache it, and return it.

Normally, you would use it as:

```js
const hotfix = await checkForHotfixes();
```

However, in this case we don't want the whole extension to have to await that function, so we just check the cache manually first (`cache.get` is nearly instant) and then we call `checkForHotfixes` so that the next time `cache.get` is called, the value will probably be cached.

source/features/index.tsx
```diff
@@ -104,6 +106,11 @@ const globalReady: Promise<RGHOptions> = new Promise(async resolve => {
 	// Options defaults
 	const options = await optionsStorage.getAll();
 
+	// Fetch hotfixes asynchronously
+	const hotfix = await cache.get('hotfix');
```
fregante, line 110: Missing:
https://github.com/sindresorhus/refined-github/issues/3529#issuecomment-687417913

Also it would need a comment, e.g. `// Ignores the cache when loaded locally`

fregante:
> Note : since this is a background feature, I couldn't figure
> out a way to test these changes.

To test this:

1. Know that there's a "clear cache" button in the extension’s option page
2. Try changing the version in `manifest.json` to `20.9.3`
3. Open https://github.com/sindresorhus/refined-github

The feature `cleanup-repo-filelist-actions` should be disabled on that page because of this content: https://github.com/sindresorhus/refined-github/blob/hotfix/hotfix.json

source/features/index.tsx
shinenelson: there is also an `import tinyVersionCompare from 'tiny-version-compare';` in `source/features/tag-changelog-link.tsx` but I agree `compareVersions` is better than `{tiny,loose}VersionCompare`

source/features/index.tsx
shinenelson: I pondered over this for a while before deciding it was accurate. I believe we are skipping the application of the hotfix with `return {}` here.

Hence, it would make sense that when the `versionCompare` check is `-1` ( second one is greater ), it should pass-through.

shinenelson: wait, I just realized that my assumption was wrong. I was looking at the code example in the [Usage section](https://github.com/fregante/tiny-version-compare#usage) of the `README` for `tiny-version-compare`; but the [API section](https://github.com/fregante/tiny-version-compare#api) suggests the opposite as in the review here.

I'll put up a pull request for that change to that repository as well :)

source/features/index.tsx
shinenelson: I'm sorry I missed this. :confused: :pray:

source/features/index.tsx
fregante: Nope.

```js
tinyVersionCompare('20.9.4', '20.9.3') === 1 // Is affected, don't return {}
tinyVersionCompare('20.9.4', '20.9.4') === 0 // Is NOT affected, exit
tinyVersionCompare('20.9.4', '20.9.5') === -1 // Is NOT affected, exit
```

The examples are correct. Run them yourself on RunKit:

https://npm.runkit.com/tiny-version-compare

Screen Shot 2020-09-05 at 09 37 41

fregante: > but the [API section](https://github.com/fregante/tiny-version-compare#api) suggests the opposite as in the review here

Uh, yeah. In my mind -1 meant left one is greater, 1 meant right is greater, hence the confusion 🤦

fregante: I fixed the readme

shinenelson: > In my mind -1 meant left one is greater, 1 meant right is greater

yes, that makes logical sense :
```
   -1		 0		1
  left	  equal  right
```

this means that this suggestion is incorrect and the original version was accurate right?

source/features/index.tsx
shinenelson: ( bringing over the comment from the issue )

> ```
> const hotfix = chrome.manifest.version === '0.0.0' || await cache.get('hotfix');
> ```
> 
> Edit: the code should be exactly this. Firefox reuses the same options when adding the development version, so we cannot rely on or affect the cache.

are you sure it should be `chrome.manifest.version` and not `browser.runtime.getManifest().version`?

I'm getting a TypeScript error that reads `TS2339: Property 'manifest' does not exist on type 'typeof chrome'.`

or did you mean 
```
const hotfix = browser.runtime.getManifest().version === '0.0.0' || await cache.get('hotfix');
```

source/features/index.tsx
fregante: My mind was incorrect, the suggestion is incorrect. Test it.

yakov116:
@fregante should this go into its own file? The index.ts it's getting very large.

source/features/index.tsx
shinenelson: @fregante, ping :point_up: some pointers needed

the reason why I was confused was because you mentioned that the code should be **exactly** that. And also, _Firefox uses the same options_ indicates that this should a chrome-specific value hence it should be `chrome.*`

source/features/index.tsx
```diff
@@ -164,13 +164,9 @@ const setupPageLoad = async (id: FeatureID, config: InternalRunConfig): Promise<
 // with the current version of extension
 const checkForHotfixes = cache.function(async () => {
 	const response = await fetch('https://raw.githubusercontent.com/sindresorhus/refined-github/hotfix/hotfix.json');
-	const hotfixes = await response.json();
+	const hotfixes: AnyObject | false = await response.json();
```
shinenelson, line 167: should this have the `| false`?

trying to resolve a `@typescript-eslint/prefer-optional-chain` error on L169 threw me : 
```
source/features/index.tsx:167:16 - error TS2339: Property 'unaffected' does not exist on type 'false | Record<string, any>'.
  Property 'unaffected' does not exist on type 'false'.

169  if (hotfixes?.unaffected) {
                   ~~~~~~~~~~

source/features/index.tsx:169:32 - error TS2339: Property 'unaffected' does not exist on type 'false | Record<string, any>'.
  Property 'unaffected' does not exist on type 'false'.

171  	if (compareVersions(hotfixes.unaffected, currentVersion) >= 0) {
     	                             ~~~~~~~~~~
```

I'm not an expert with TypeScript, but I believe we should always expect `hotfixes` to be a JSON object right? Even if we do not have any breaking features, we'll probably have `{ false }` in the file. I don't know what would happen if there was a network error though.

I was able to avoid the lint error locally by removing the `| false`. Hence the question.

fregante: If eslint complains about `hotfix && hotfix.unaffected`, that rule is wrong in this case; disable it with `eslint-disable-next-line`

fregante: `{false}` is not a valid JSON object (or even valid js). `false` is. The file will literally only contain the word `false`, hence the type `AnyObject | false`

source/features/index.tsx
fregante: Yes I meant that. That code was just a pointer, like the initial code.

source/features/index.tsx
fregante:
```suggestion
```

shinenelson:
after multiple roadblocks and workarounds, I've finally managed to test the feature.

my first issue was with Firefox not able to `fetch` the `hotfix.json` file from remote because of a [sourceMap error](https://developer.mozilla.org/en-US/docs/Tools/Debugger/Source_map_errors#NetworkError_when_attempting_to_fetch_resource) with https://github.com/webpack/webpack/issues/1194 on Firefox which took me down a rabbit hole. I finally gave up and went to Chromium to test the remote fetch.

I realized this early ( during my Firefox testing itself when I over-rode the remote `fetch` with a hard-coded object ), but my second issue was with the content of `hotfix.json` itself. The feature name is supposed to be prepended with `feature:` for it to actually override the value in `options`.

```json
{"feature:cleanup-repo-filelist-actions": false, "unaffected": "20.9.4"}
```
**Disclaimer** : I was only able to confirm that the JSON that is pulled from the remote file is appended to `options`. Separately, I also tested, provided that the hot-fix value is the right format, then the feature is disabled ( I tested this by hard-coding the value, so I disabled multiple features and all of them were disabled upon reloading the page ).

Also, I noticed that if `logging` is enabled on the extension, then the disabled feature is logged to the console as `Skipping cleanup-repo-filelist-actions <empty string>`. It would great if we could somehow trigger the reason in place of `<empty string>`. I spent some time thinking and looked through the code where we could put this logic, but I couldn't find a place where it would really work. Would it be possible to over-ride the `features.add` function call or something?

PS : The lint check failing against my last commit 37646fd looks like a false positive. There is no double `import` in `source/background.ts`. I ran `npm run test` and `npm run lint` multiple times locally ( without errors ) before `push`ing my last commit to my remote.

source/features/index.tsx
```diff
@@ -105,9 +105,11 @@ const globalReady: Promise<RGHOptions> = new Promise(async resolve => {
...
 	// Options defaults
 	const options = await optionsStorage.getAll();
-
-	// If features are remotely marked as "seriously breaking" by the maintainers, disable them without having to wait for proper updates to propagate #3529
-	const hotfix = await cache.get('hotfix');
+	const hotfix = browser.runtime.getManifest().version === '0.0.0' || await cache.get('hotfix'); // Ignores the cache when loaded locally
+	// if features are marked as "seriously breaking" by maintainers
+	// disable them without having to wait for fixed version updates
+	// to propogate to browsers
+	// refer GitHub issue #3529 for more information
```
fregante, line 112: Why did you badly rewrite this comment? I had already written it.

fregante:
> The lint check failing against my last commit [37646fd](https://github.com/sindresorhus/refined-github/commit/37646fde157cfe3af3045fada5a301a280f2a867) looks like a false positive

PR checks are tested against [the "merge commit"](https://stackoverflow.com/a/57697936/288906). This means that even if not visible, the actual file being tested was https://github.com/sindresorhus/refined-github/blob/817871850908ffa687b0b087d03199f5eed3c40e/source/background.ts#L1-L3

You can see that there are two imports, because one was recently added by 2dc5f263b97730b82ceee8cb64edc5da76fe5893

Thankfully we have this feature to update the PR with those changes and fix them: https://github.com/sindresorhus/refined-github/blob/master/source/features/update-pr-from-base-branch.tsx

source/features/index.tsx
shinenelson: I'll revert my changes.

shinenelson: done. and all checks are passing. thanks for updating the branch with the default branch. I should've thought of that before calling the test failure a 'false positive'.

fregante:
I tested this and it works in Chrome, but not in Firefox: neither with `window.fetch` nor `window.content.fetch`, apparently due to lack of permission in the former and due to CSP in the latter.

We can't use `https://raw.githubusercontent.com` because the browsers will block the extension and ask the users to re-enable it. The solution is to get the content of the file via API.

You can use `api.v3` on this endpoint: https://developer.github.com/v3/repos/contents/#get-repository-content

source/features/index.tsx
```diff
@@ -150,6 +157,24 @@ const setupPageLoad = async (id: FeatureID, config: InternalRunConfig): Promise<
 	}
 };
 
+const checkForHotfixes = cache.function(async () => {
+	const response = await fetch('https://raw.githubusercontent.com/sindresorhus/refined-github/hotfix/hotfix.json');
+	const hotfixes: AnyObject | false = await response.json();
+
+	// eslint-disable-next-line @typescript-eslint/prefer-optional-chain
+	if (hotfixes && hotfixes.unaffected) {
```
yakov116, line 165:
```suggestion
	if (hotfixes?.unaffected) {
```

No?

fregante: No, it's `false` sometimes

shinenelson:
> I tested this and it works in Chrome, but not in Firefox: neither with window.fetch nor window.content.fetch, apparently due to lack of permission in the former and due to CSP in the latter.

were you testing with `web-ext`? that didn't work for me ( as mentioned in my [previous comment](https://github.com/sindresorhus/refined-github/pull/3530#issuecomment-691583804) ), but I was hoping that it would work if it came via an update through the proper channels.

> We can't add https://raw.githubusercontent.com to the permission list because the browsers will block the extension and ask the users to re-enable it. The solution is to get the content of the file via API.

Good catch. I'll use the `repos/contents` API for the fetch.

> I just updated the contents of hotfix.json now
thanks. it'll make testing easier for me.

fregante:
Don’t call the API directly, we have the file `api.ts` for that. Call `api.v3(…)`

yakov116:
@fregante would that cause issues when run on enterprise?

shinenelson:
I am always skeptical about `atob` whenever I have used it. So, to be on the safer side, I use `Buffer.from` everywhere.

I think it makes sense here since this is a web extension and it is only about DOM manipulation and Web APIs. Honestly, I did not see the edited message and made the changes before I saw the edited comment. If you feel strongly about `atob`, I can make that change as well but c8ba275 works with `Buffer.from` though. I've tested it with Chromium.

shinenelson:
@yakov116 : I noticed this too when I was exploring `source/github-helpers/api.ts` and pondered over this. I might be wrong here, but I believe L67-L69 creates the domain part of the URL i.e., `api.github.com` or `github.enterprise.com` and L109 adds the actual path of the URL.

with normal GitHub : 
```
api3 = https://api.github.com/
query = repos/sindresorhus/refined-github/contents/hotfix.json?ref=hotfix
url = https://api.github.com/repos/sindresorhus/refined-github/contents/hotfix.json?ref=hotfix
```

with GitHub Enterprise :
```
location.origin = https://github.enterpise.com
api3 = https://github.enteripse.com/api/v3/
query = repos/sindresorhus/refined-github/contents/hotfix.json?ref=hotfix
url = https://github.enterprise.com/repos/sindresorhus/refined-github/contents/hotfix.json?ref=hotfix
```

is this what you are talking about? or did I miss the point altogether?

yakov116:
> is this what you are talking about?

Yes

fregante:
> @fregante would that cause issues when run on enterprise?

Good catch. I wanted to use `api.v3` specifically to support GHE, but didn't think that it would actually break it

yakov116:
@fregante I think having an enterprise token will make an issue. Since it includes the token if it exists.

I tried to put in an invalid token and I got 

```js
refined-github.js:19790 Uncaught (in promise) Error: The token seems to be incorrect or expired. Update it in the options.
    at getError (refined-github.js:19892)
    at async Module../source/github-helpers/api.ts.mem__WEBPACK_IMPORTED_MODULE_0___default.cacheKey (refined-github.js:19842)
    at async ./source/features/index.tsx.webext_storage_cache__WEBPACK_IMPORTED_MODULE_1__.default.function.maxAge.seconds (refined-github.js:12131)
    at async getSet (refined-github.js:6490)
```

fregante:
I tried using api.ts to possible use the token when making the call, but every “automatic” part comes back to bite us in this case. 

Let’s revert to fetch

yakov116:
But in the lint pr?

shinenelson:
please let me know if I can to help here. does this only involve changing the `response` variable to use fetch as in 5324689?

yakov116:
> please let me know if I can to help here. does this only involve changing the `response` variable to use fetch as in [5324689](https://github.com/sindresorhus/refined-github/commit/53246890b4ae85258bd10ea590af7e55bf7ad0bd)?

```js
	const request = await fetch('https://api.github.com/repos/sindresorhus/refined-github/contents/hotfix.json?ref=hotfix');
	const response = JSON.parse(await request.text());
	const hotfixes: AnyObject | false = JSON.parse(atob(response.content));
```

You can send a pr. This should be the code.

PR #3532, opened 2020-09-05
Correctly dim `bot` label in `dim-bots`

wilau2:
Fix: #3520 
@fregante I went a bit fast by approving the change you propose. 
This will work a lot better, since there is not `/` in the URL.

### DIFF
--- a/source/features/dim-bots.tsx
+++ b/source/features/dim-bots.tsx
@@ -13,7 +13,7 @@ export const botSelectors = [
 
 	/* Issues/PRs */
 	'.opened-by [href*="author%3Aapp%2F"]',
-	'.labels [href$="/bot"]'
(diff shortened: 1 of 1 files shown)


fregante:
Always provide a link where the PR can be tested

wilau2:
Here is a link with a repository with the `bot` label:
https://github.com/wilau2/harvest-balance-calculator/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc
Here is the inspector open on the label:
image

fregante:
I think then it should be a more clear `label%3Abot`

wilau2:
@fregante updated and tested.

fregante:
Thanks! It’s a selector by the way, not a regex 👍

PR #3540, opened 2020-09-11
Lint

yakov116:
- Faster `show-associated-branch-prs-on-fork`, Request before dom ready
- Sort From
- **`isEmptyRepoRoot` does not work**
- Lint

@fregante 
> https://github.com/fregante/github-url-detection/commit/0c36c4c08636174cc039b077680ff90acb3b5d37

I followed https://github.com/sindresorhus/refined-github/pull/3505/files/92330f5ca0627ce23adb5ef216c7007008e844dc#diff-7988466746fb3fcac77295b4dd4038d1 (npm deprecate *) but that gave me an error. (I can provide a stack trace if needed 😉 )

'.blankState' does not exist on repo root. 

Direction needed.

### DIFF
--- a/source/features/ci-link.tsx
+++ b/source/features/ci-link.tsx
@@ -39,7 +39,7 @@ void features.add({
 		pageDetect.isRepo
 	],
 	exclude: [
-		() => select.exists('[aria-label="Cannot fork because repository is empty."]')
+		pageDetect.isEmptyRepo
 	],
 	awaitDomReady: false,
 	init
--- a/source/features/ci-link.tsx
+++ b/source/features/ci-link.tsx
@@ -38,6 +38,9 @@ void features.add({
 	include: [
 		pageDetect.isRepo
 	],
+	exclude: [
+		() => select.exists('[aria-label="Cannot fork because repository is empty."]')
+	],
 	waitForDomReady: false,
 	init
 });
--- a/source/features/ci-link.tsx
+++ b/source/features/ci-link.tsx
@@ -41,6 +41,6 @@ void features.add({
 	exclude: [
 		() => select.exists('[aria-label="Cannot fork because repository is empty."]')
 	],
-	waitForDomReady: false,
+	awaitDomReady: false,
 	init
 });
--- a/source/features/repo-age.tsx
+++ b/source/features/repo-age.tsx
@@ -123,7 +123,7 @@ void features.add({
 		pageDetect.isRepoRoot
 	],
 	exclude: [
-		pageDetect.isEmptyRepoRoot
+		() => select.exists('[aria-label="Cannot fork because repository is empty."]')
 	],
 	waitForDomReady: false,
 	init
--- a/source/features/repo-age.tsx
+++ b/source/features/repo-age.tsx
@@ -123,9 +123,9 @@ void features.add({
 		pageDetect.isRepoRoot
 	],
 	exclude: [
-		// Excludes empty repos, compatible with `waitForDomReady: false`
+		// Excludes empty repos, compatible with `awaitDomReady: false`
 		() => !select.exists('link[rel="canonical"]')
 	],
-	waitForDomReady: false,
+	awaitDomReady: false,
 	init
 });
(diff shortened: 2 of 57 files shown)


yakov116 (commented):
.

source/features/repo-age.tsx
```diff
@@ -123,7 +123,7 @@ void features.add({
 		pageDetect.isRepoRoot
 	],
 	exclude: [
-		pageDetect.isEmptyRepoRoot
+		() => !select.exists('link[rel="canonical"]')
```
yakov116, line 126: I think isEmptyRepo should be isRepoHome and :arrow_up:

source/features/ci-link.tsx
```diff
@@ -38,6 +38,9 @@ void features.add({
 	include: [
 		pageDetect.isRepo
 	],
+	exclude: [
+		() => select.exists('[aria-label="Cannot fork because repository is empty."]')
```
fregante, line 42: Why this here and the rel=canonical below?

yakov116: This runs on pages other then repo home. Only repo home will always has the rel.

fregante: Empty repos don't have the canonical on any page

yakov116:
Sorry let me explain slowly. 

The only time repo-age will error is on a brand new repo without any commits. 

>Empty repos don't have the canonical on any page

Correct but ci-link runs on other pages and canonical does not exists on all pages even on non-empty repos.

fregante:
> There will only be repoHome

I see what you mean, but that's unnecessarily specific. The feature is already limited to `isRepoRoot`, which includes `isRepoHome`, and any other page that `isRepoRoo` but not `isRepoHome` already can't happen, so it's pointless to explicitly exclude it.

> the fastest way we can detect that its empty is using rel

Good thinking here. Needs a JS comment though.



> canonical does not exists on all pages even on non-empty repos

🙌

source/features/repo-age.tsx
fregante:
```suggestion
		// Excludes empty repos, compatible with `waitForDomReady: false`
		() => !select.exists('link[rel="canonical"]')
```

fregante:
Can you also rename `waitForDomReady` to `awaitDomReady` everywhere?

yakov116:
> Can you also rename `waitForDomReady` to `awaitDomReady` everywhere?

Like! Will do.

source/features/ci-link.tsx
```diff
@@ -38,6 +38,9 @@ void features.add({
 	include: [
 		pageDetect.isRepo
 	],
-	waitForDomReady: false,
+	exclude: [
+		() => select.exists('[aria-label="Cannot fork because repository is empty."]')
```
fregante, line 42: I think this can be added as `isEmptyRepo`, what do you think?

source/features/repo-age.tsx
```diff
@@ -123,8 +123,9 @@ void features.add({
 		pageDetect.isRepoRoot
 	],
 	exclude: [
-		pageDetect.isEmptyRepoRoot
+		// Excludes empty repos, compatible with `awaitDomReady: false`
+		() => !select.exists('link[rel="canonical"]')
```
fregante, line 127: I think this should be merged back as `isEmptyRepoRoot`

```js
// On empty repos, there's only isRepoHome
const isEmptyRepoRoot = () => isRepoHome() && !select.exists('link[rel="canonical"]');
```

fregante: In the future, changes like this one should be made on `github-url-detection`, not here.

source/features/ci-link.tsx
yakov116: Yes

source/features/repo-age.tsx
yakov116: That was my intention. I brought it up here to explain the issue.

I will make the 2 pr's now.


In the future would you like me to create an issue there first. From my experience it usually works out best, when we discuss it here and then send the pr.

fregante: Ok

fregante:
Yes. Lint PRs shouldn’t really change the behavior, especially if like in that case it’s an improvement.

PR #3482, opened 2020-08-19
Move "Security" and "Insights" repository tab to `more-dropdown`

artusm:
1. LINKED ISSUES:
   Closes #3354

2. TEST URLS:
   [https://github.com/sindresorhus/refined-github](https://github.com/sindresorhus/refined-github)

3. SCREENSHOT:
  
Screenshot_20200819_083833
Screenshot_20200819_083918

Issue #3354, opened 2020-07-16
Move "Security" and "Insights" repository tab to `more-dropdown`


fregante:
https://github.com/sindresorhus/refined-github/pull/3346 restored `more-dropdown` but now we also need to move the Security and Insights tabs to it, like it was before the "Repository refresh"

Possible way to do it:

- remove tabs from dom and show/enable the items in the dropdown

Gotchas:

- the DOM order of the menu **matters**, so you probably need to also move them at the end of the dropdown, _after_ "Settings"

yakov116:
Currently the . . . is always shown (even if nothing is inside) This would also fix that.

fregante:
The dropdown can never be empty, regardless of circumstances. I don’t know if #3347 is related, but I think not. If it’s empty, the creation is failing, **not** the native resize listener.

yakov116:
What is a repo does not have releases such as https://github.com/yakov116/pulse

fregante:
Open a new issue. It’s a bug.

Xottab-DUTY:
This is ridiculously inconvenient! We need some mechanism to be able to configure what should be in the `more-dropdown` and what should be on top.

I just spent 15+ minutes to find out what the heck and where is my Insights button...

busches:
@Xottab-DUTY you can disable the feature and keep the tabs where they started. We won't be adding a configuration option for this, at most we could split each tab into it's own feature.

fregante:
It's good as is. If you need those tabs, you can disable `more-dropdown`, it can't be split

Issue #3347, opened 2020-07-14
`releases-tab` sometimes doesn't flow into overflow hamburger menu


fregante:
This is due to a race condition

1. Be on any (responsive) page
2. Resize the window

Sometimes it disappears, sometimes it doesn't:

![](https://user-images.githubusercontent.com/1402241/87455271-777c6a80-c605-11ea-92dc-f9cb428d30c7.gif)

This might be due to the tab being added **after** GitHub looks for the tabs, so the DOM element should probably be added _before_ the API call, hidden, and then unhidden/removed as necessary.

fregante:
This might be the code responsible for this:
```js
p(".js-responsive-underlinenav", {
	constructor: HTMLElement,
	subscribe: e=>{
		const t = e.querySelectorAll(".js-responsive-underlinenav-item");
		return async function(e, t) {
			await ge,
			Os(e, t)
		}(e, t),
		ae(window, "resize", ()=>Os(e, t))
	}
})
```

`p` is probably `selector-observer` and as soon as it finds the menu, it selects the items in it. `releases-tab` might not be in there yet.

PR #3547, opened 2020-09-13
Move Security and Insights tabs to `more-dropdown`

fregante:
Closes https://github.com/sindresorhus/refined-github/issues/3354 (this issue wasn't a "good first issue" after all 😅)
Replaces and closes https://github.com/sindresorhus/refined-github/pull/3482

This seems to work, but https://github.com/sindresorhus/refined-github/issues/3347 still exists, complicating things a bit.

### DIFF
--- a/source/features/releases-tab.tsx
+++ b/source/features/releases-tab.tsx
@@ -104,7 +104,7 @@ async function init(): Promise<false | void> {
 			releasesTab.setAttribute('aria-current', 'page');
 		}
 
-		select('[data-menu-item="insights-tab"]', repoNavigationBar)!.after(
+		select('.dropdown-divider', repoNavigationBar)!.before(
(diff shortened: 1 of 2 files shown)


source/features/releases-tab.tsx
```diff
@@ -104,7 +104,7 @@ async function init(): Promise<false | void> {
 			releasesTab.setAttribute('aria-current', 'page');
 		}
 
-		select('[data-menu-item="insights-tab"]', repoNavigationBar)!.after(
+		select('.dropdown-divider', repoNavigationBar)!.before(
```
fregante, line 107: For the record, this code seems to be executed **after** `more-dropdown`, so `[data-menu-item="insights-tab"]` no longer exists

fregante:
This works in Chrome but seems to break the overflow logic in Firefox.

PR #3552, opened 2020-09-14
Faster `highlight-collaborators-and-own-conversations`

yakov116:
https://github.com/sindresorhus/refined-github/pull/3505#issuecomment-687302600

### DIFF
--- a/source/features/highlight-collaborators-and-own-conversations.tsx
+++ b/source/features/highlight-collaborators-and-own-conversations.tsx
@@ -1,6 +1,7 @@
 import './highlight-collaborators-and-own-conversations.css';
 import cache from 'webext-storage-cache';
 import select from 'select-dom';
+import domLoaded from 'dom-loaded';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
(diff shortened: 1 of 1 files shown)


source/features/highlight-collaborators-and-own-conversations.tsx
```diff
@@ -53,6 +49,10 @@ void features.add({
 	include: [
 		pageDetect.isRepoConversationList
 	],
+	exclude: [
+		() => select.exists('.blankslate')
```
fregante, line 53: This won't be found if we don't await dom ready, so it can be removed

yakov116: Tested, It works before dom ready

yakov116:
>Merge branch 'master' into **hilighcolab**

Oh my that branch name 🤦

source/features/highlight-collaborators-and-own-conversations.tsx
fregante: Race condition warning‼️

If you're ok with it working sometimes and don't come back in 3 days with another PR that's fine. Of course this expects half the page to be loaded, so it's not gonna work 100% of the time

fregante: Either way, there's no error.

Best case scenario, it avoids an API call.
Worst case scenario, there's an unnecessary API call.

Issue #3551, opened 2020-09-14
Don't see button for "Adds one-click buttons to change diff style and to ignore the whitespace."


SethDG:
Refined GitHub 20.9.11
Firefox 80.0.1
Chrome 85.0.4183.102

I have added the same personal access token to both Chrome and Firefox

I do not see the button for "Adds one-click buttons to change diff style and to ignore the whitespace." as shown here: https://user-images.githubusercontent.com/1402241/54178764-d1c96080-44d1-11e9-889c-734ffd2a602d.png

For example, on this page here I would expect to see it: https://github.com/seismic/engagement-service/pull/551/files

In Chrome, it's not working and I see the following error, and I do not see the error if I disable Refined GitHub:

`refined-github.js:9838 
Uncaught TypeError: Cannot read property 'cloneNode' of null
    at add (refined-github.js:9838)
    at runAdd (refined-github.js:4524)
    at applyChanges (refined-github.js:4476)
    at SelectorObserver.addRootNodes (refined-github.js:4742)
    at Array.processBatchQueue (refined-github.js:4433)
    at MutationObserver.handleMutations (refined-github.js:4457)`

Interestingly, if I go to https://github.com/sindresorhus/refined-github/pull/3540/files in Chrome, I don't see the error but it's still not working.

In Firefox, it's not working and I only see the following error, but I see that error whether Refined GitHub is enabled or not:

`Content Security Policy: The page’s settings blocked the loading of a resource at inline (“script-src”). onloadwff.js:71:798666`

fregante:
Duplicate of https://github.com/sindresorhus/refined-github/issues/2597

SethDG:
Awwwww, so sad.  Oh well.  Thanks for the response!  And the awesome add-on!

SethDG:
@fregante can i make a request that you update the readme.md to remove "Adds one-click buttons to change diff style and to ignore the whitespace."?

fregante:
Yeah the whole feature should be dropped from the repo because it needs to be rewritten anyway

PR #3553, opened 2020-09-14
Drop `faster-pr-diff-options` feature

yakov116:
https://github.com/sindresorhus/refined-github/issues/3551#issuecomment-692278619


busches:
As someone who uses this multiple times a day, why?

yakov116:
> As someone who uses this multiple times a day, why?

ohhh its still there on enterprise... did not know that

fregante:
> ohhh its still there on enterprise... did not know that

It's not just on Enterprise, but on any commit. The feature is only disabled on PRs, so the readme is still up to date.

PR #1738, opened 2019-01-20
Split PRs and issues in the top search

jerone:
Repository search results are currently categorized in 4 types: Code, Commits, Issues and Wikis (if available).
Where issues is a combination of issues and PR. This PR will make another type for all PR's and separate them out of the issues type.

### Test
https://github.com/sindresorhus/refined-github/search?q=test+is%3Apr&type=Issues

### Screenshot
image

### Ref
* [x] Closes #1635

PR #3556, opened 2020-09-15
Restore `split-issue-pr-search-results`

yakov116:
1. LINKED ISSUES:
   https://github.com/sindresorhus/refined-github/pull/1738

2. TEST URLS:
   https://github.com/sindresorhus/refined-github/search?q=batch-mark-files-as-viewed&unscoped_q=batch-mark-files-as-viewed

3. SCREENSHOT:
   None

Small selector update

### DIFF
--- a/source/features/split-issue-pr-search-results.tsx
+++ b/source/features/split-issue-pr-search-results.tsx
@@ -28,7 +28,7 @@ function init(): void {
 	pageSearchQuery = new SearchQuery(location);
 
 	const issueLink = select<HTMLAnchorElement>([
-		'nav.menu a[href*="&type=Issues"]', // Pre "Repository refresh" layout
+		'nav.menu a[href*="&type=Issues"]', // Todo remove when GHE selector is updated - (Maybe October 2020)
 		'.menu-item[href*="&type=issues"]'
 	])!;
 	issueLink.textContent = 'Issues'; // Drops any possible counter
--- a/source/features/split-issue-pr-search-results.tsx
+++ b/source/features/split-issue-pr-search-results.tsx
@@ -27,7 +27,10 @@ function init(): void {
 	cleanLinks();
 	pageSearchQuery = new SearchQuery(location);
 
-	const issueLink = select<HTMLAnchorElement>('nav.menu a[href*="&type=Issues"]')!;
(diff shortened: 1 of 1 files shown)


source/features/split-issue-pr-search-results.tsx
```diff
@@ -27,7 +27,10 @@ function init(): void {
 	cleanLinks();
 	pageSearchQuery = new SearchQuery(location);
 
-	const issueLink = select<HTMLAnchorElement>('nav.menu a[href*="&type=Issues"]')!;
+	const issueLink = select<HTMLAnchorElement>([
+		'nav.menu a[href*="&type=Issues"]', // Pre "Repository refresh" layout
```
fregante, line 31: Is it? This page is not technically a repo page. I think this should just be updated

yakov116: Your right but enterprise needs it. I wanted to keep the same wording that when we drop the support for it we can drop em all.

fregante: We can’t support 2 versions of everything. The repo refresh was an exception since it affected a lot of central features. GHE will just have to miss this minor feature for the time being.

fregante: Since it’s so minor we *could* keep the double selector, but it needs to be marked as GHE, not the current comment.

yakov116:
```suggestion
		'nav.menu a[href*="&type=Issues"]', // Todo remove when GHE selector is updated - (October 2020)
```

source/features/split-issue-pr-search-results.tsx
```diff
@@ -27,7 +27,10 @@ function init(): void {
 	cleanLinks();
 	pageSearchQuery = new SearchQuery(location);
 
-	const issueLink = select<HTMLAnchorElement>('nav.menu a[href*="&type=Issues"]')!;
+	const issueLink = select<HTMLAnchorElement>([
+		'nav.menu a[href*="&type=Issues"]', // Todo remove when GHE selector is updated - (Maybe October 2020)
```
fregante, line 31:
```suggestion
		'nav.menu a[href*="&type=Issues"]', // Only for GHE
```

Issue #3349, opened 2020-07-15
Scroll sidebar on repositories 


piranna:
With new projects page layout, project info has been moved from page top to a column on the right side. When scrolling down, this info gets out of view, getting an empty space. My suggestion is to don't scroll it beyond its bottom, so when scrolling down the page, it gets fixed an visible all the time.

fregante:
We can do that if it fits the viewport vertically like we do for the conversation sidebar

piranna:
Yes, it's a something similar behaviour :-)

shinenelson:
This could be as easy as duplicating `source/features/sticky-conversation-list-toolbar.css` and changing the CSS selector to `.BorderGrid`. However, most repositories have many sections that usually do not fit in the viewport height. What do we do about that?

Personally, I like to see the Languages section, but most often, I have to scroll to find it. Making the sidebar sticky would mean that I have to get the bottom of the page before I can see the Languages section.

piranna:
What about allowing to scroll just only itself, like a subsection in the page?

fregante:
@piranna scrollable frames on web pages don’t belong to this decade 😃

shinenelson:
@fregante I did try hooking into `sticky-conversation-sidebar` but I hit 2 problems : 
1. the `.rgh-sticky-sidebar` class is toggled with the condition `sidebarHeight < window.innerHeight`. As mentioned previously, most repositories have sidebar sections that will invalidate the expression, hence not enabling stickiness on the sidebar.
2. the value of `top` is hard-coded in the CSS of `.rgh-sticky-sidebar`.

For 2, we could create an extra class for the `top` value specifically for the different kinds of elements ( for example, `rgh-conversation-sidebar`, `rgh-repo-sidebar`, `rgh-issues-list-toolbar` ). Even then, it still does not solve 1 though.

For 1, if we were to over-engineer, we could hide / remove a few sections that could be un-necessary ( Sponsors, Contributors and empty sections like Packages ) and then make the sidebar sticky. But then, that could be a new feature altogether though.

It was when I was digging around the CSS that I found `sticky-conversation-list-toolbar` and tested with that and it worked ( at least for making the sidebar sticky ).

fregante:
> 1. most repositories have sidebar sections that will invalidate the expression, hence not enabling stickiness on the sidebar.

Right. That's what we want. If it doesn't fit, it can't be sticky. That's the right behavior.

[...]

PR #3566, opened 2020-09-16
Hide redundant 'Releases' section from repo sidebar

shinenelson:
The 'Releases' section in the 'Refreshed Repository' design sidebar is redundant when there is already the releases-tab.

This change removes the 'Releases' section from the sidebar.

I'm not sure whether this has been discussed before ( I did dig through some old issues, but could not find any ), but there is one direct link to the latest release that goes away with the section.

I usually find myself either navigating directly to `/releases/latest` or simply going to the main Releases page and referring to the latest release. I do not know if people use the latest release link as often. But the remaining information is already on the releases-tab.

1. LINKED ISSUES:
   None

2. TEST URLS:
   Any repository root page
	* https://github.com/xojs/xo
	* https://github.com/avajs/ava

3. SCREENSHOT:

	Before : 
	Screenshot_2020-09-17 sindresorhus refined-github

	After : 
	Screenshot_2020-09-17 sindresorhus refined-github (1)

### DIFF
--- a/source/features/releases-tab.tsx
+++ b/source/features/releases-tab.tsx
@@ -112,7 +112,8 @@ async function init(): Promise<false | void> {
 
 		// Hide redundant 'Releases' section from repo sidebar
 		if (pageDetect.isRepoRoot()) {
-			select('.BorderGrid-cell a[href$="/releases"]')?.closest('.BorderGrid-row')!.setAttribute('hidden', '');
+			const sidebarReleases = await elementReady('.BorderGrid-cell a[href$="/releases"]');
+			sidebarReleases!.closest('.BorderGrid-row')!.hidden = true;
 		}
 
 		return;
--- a/source/features/releases-tab.tsx
+++ b/source/features/releases-tab.tsx
@@ -112,7 +112,7 @@ async function init(): Promise<false | void> {
 
 		// Hide redundant 'Releases' section from repo sidebar
 		if (pageDetect.isRepoRoot()) {
-			select('.BorderGrid-cell a[href$="/releases"]')?.closest('.BorderGrid-row')!.removeAttribute('hidden');
+			select('.BorderGrid-cell a[href$="/releases"]')?.closest('.BorderGrid-row')!.setAttribute('hidden', '');
 		}
 
 		return;
--- a/source/features/releases-tab.tsx
+++ b/source/features/releases-tab.tsx
@@ -113,7 +113,7 @@ async function init(): Promise<false | void> {
 		// Hide redundant 'Releases' section from repo sidebar
 		if (pageDetect.isRepoRoot()) {
 			const sidebarReleases = await elementReady('.BorderGrid-cell a[href$="/releases"]');
-			sidebarReleases!.closest('.BorderGrid-row')!.hidden = true;
+			sidebarReleases!.closest('.BorderGrid-row')!.setAttribute('hidden', '');
 		}
 
 		return;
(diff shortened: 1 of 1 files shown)


fregante:
> The 'Releases' section in the 'Refreshed Repository' design sidebar is redundant

Is it? The Releases tab points to /releases while that section can point straight to the latest release at /releases/v2.3.4

However… /releases will most likely show the latest release first, so this may be right.

fregante (approved):

shinenelson:
> The Releases tab points to /releases while that section can point straight to the latest release at /releases/v2.3.4

The section header and the '+ x releases' link at the bottom of the section, both point to `/releases` 
whereas the tag that lists the latest release - that is the only one that points directly to the latest release. That is why I mentioned that that information might be lost. But I am doubtful how many people actually use that link to go the latest release. I think most people go to the release page and like you said
>  /releases will most likely show the latest release first

which is my case for the argument that the section is redundant

fregante:
Why hide it instead of removing it?

shinenelson:
I was thinking in terms of when the sidebar becomes sticky ( #3349 ) and the view has scrolled down quite a bit. There, one would not have quick access to the Releases Tab and would prefer to see the Releases section in the sticky sidebar. At that point, if we wanted to re-activate the section, we would not be able to if the section was completely removed.

I was thinking for the future. Also, there maybe other features or ( even external extensions ) that query the DOM for the Releases section and not finding it would break them. After all, we are not removing an empty section without any content, we are removing a section that is seemingly redundant when the Releases Tab is in view. That information could still be useful when the Releases Tab is not in view.

fregante:
> maybe other features or ( even external extensions ) that query the DOM for the Releases section and not finding it would break them

This is a good reason

> would prefer to see the Releases section in the sticky sidebar

This is not. The sticky sidebar depends on the height, so we can't make it taller once it becomes sticky, because it would invalidate the calculation.

source/features/releases-tab.tsx
```diff
@@ -110,6 +110,11 @@ async function init(): Promise<false | void> {
 			})
 		);
 
+		// Hide redundant 'Releases' section from repo sidebar
+		if (pageDetect.isRepoRoot()) {
+			select('.BorderGrid-cell a[href$="/releases"]')?.closest('.BorderGrid-row')!.setAttribute('hidden', '');
```
fregante, line 115:
```suggestion
			select('.BorderGrid-cell a[href$="/releases"]')?.closest('.BorderGrid-row')!.removeAttribute('hidden');
```

fregante, line 115: This appears to always exist (so no `?.`) but it might not be ready yet, the previous code only awaits the header. I think you need:
```suggestion
			const sidebarReleases = await elementReady('.BorderGrid-cell a[href$="/releases"]');
			sidebarReleases!.closest('.BorderGrid-row')!.hidden = true;
```

fregante:
# 🎉 🙌

Geeyun-JY3:
I think this should be mentioned in the description of `releases-tab` option.

aried3r:
> I think this should be mentioned in the description of `releases-tab` feature.

I agree, I wish this was behind a feature flag itself. I did use this more than I thought it would, to directly go to the latest release which often contains the release notes/change log (like this repo).

Is there any possibility this would get its own feature flag? I understand this is a heavily opinionated set of refinements so of course if there's no chance for that, I'll just disable `releases-tab` as a compromise. I'd also be happy to contribute :)

yakov116:
What if we merged https://github.com/sindresorhus/refined-github/blob/master/source/features/hide-empty-meta.tsx
With ⬆️ and https://github.com/sindresorhus/refined-github/commit/47a284946834f92f747fdaecdfcdc8a8e8e69e77 into one feature of `clean-repo-sidebar`?

PR #3585, opened 2020-09-21
Avoid repository settings bug caused by `stop-pjax-loading-with-esc`

### LINKED ISSUE #3584: Edit button doesn't work on repo settings->Webhooks page when using RGH
I ran into this weird bug where clicking on the edit button in repo's Webhooks settings doesn't do anything when I'm using Refined GitHub and I have to manually go to the site the link goes to so that the sudo mode prompt shows up.

Here's the page I'm referring to, the "Edit" button (as well as the blured webhook link that also goes to same page) doesn't do anything:
image
Example link: `https://github.com/REPO_OWNER/REPO_YOU_HAVE_ADMIN_PERMS_ON/settings/hooks` (I can't give you my link since you obviously don't have admin access on the repo I have this issue on)

When clicking the edit button, the URL in the browser's URL bar gets updated (`https://github.com/REPO_OWNER/REPO_YOU_HAVE_ADMIN_PERMS_ON/settings/hooks/WEBHOOK_ID`) and if I refresh the site, it properly shows the sudo mode prompt, but that's really just a workaround for this issue.

Here's sudo mode prompt if you don't know what I'm referring to:
image

artusm:
1. LINKED ISSUES:
   Closes #3584

2. TEST URLS:
	https://github.com/sindresorhus/refined-github/settings/hooks
   `https://github.com/REPO_OWNER/REPO_YOU_HAVE_ADMIN_PERMS_ON/settings/hooks`

3. SCREENSHOT:

### DIFF
--- a/source/features/stop-pjax-loading-with-esc.tsx
+++ b/source/features/stop-pjax-loading-with-esc.tsx
@@ -4,7 +4,6 @@ import * as pageDetect from 'github-url-detection';
 import features from '.';
 
 let progressLoader: HTMLElement;
-let pjaxErrorShouldBePrevented = false;
 const progressLoaderLoadingClass = 'is-loading';
 
 function fixProfileNavAndTimeline(): void {
@@ -34,17 +33,16 @@ function keydownHandler(event: KeyboardEvent): void {
 		}
 	}
 
-	pjaxErrorShouldBePrevented = true;
+	window.addEventListener('pjax:error', pjaxErrorHandler, {once: true});
 
 	history.back();
 	progressLoader.classList.remove(progressLoaderLoadingClass);
 }
 
 function pjaxErrorHandler(event: CustomEvent): void {
-	if (pjaxErrorShouldBePrevented && event.cancelable) {
+	if (event.cancelable) {
 		// Avoid location.replace() when AbortController.abort() throw an error
 		event.preventDefault();
-		pjaxErrorShouldBePrevented = false;
 	}
 }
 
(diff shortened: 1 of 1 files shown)


fregante:
It feels like this shouldn't `pjaxErrorHandler` shouldn't be called at all if the user didn't press `esc`

I don't remember, why is that needed? What triggers it?

I think the solution is to only add the `pjax:error` if necessary, once, for example inside `keydownHandler` instead of in `init`:

```js
window.addEventListener('pjax:error', pjaxErrorHandler, {once: true});
```

artusm:
> I don't remember, why is that needed? What triggers it?

https://github.com/sindresorhus/refined-github/blob/a50690c4f5e53820d4c84bd2ecfabe49aab98da7/source/features/stop-pjax-loading-with-esc.tsx#L42

fregante:
Indeed, but why would it throw? We're just calling `history.back()` 
Does that cause the error?

artusm:
> We're just calling `history.back()`

which triggers `popstate` event.

```javascript
window.addEventListener("popstate", function (t) {
        ut || null == currentPjaxAbortController || currentPjaxAbortController.abort();
```
which in turn will trigger `AbortController.abort()`.



```javascript
try {
const { signal } = (currentPjaxAbortController = new AbortController());
fetch(url, { signal } );
} catch (k) {           
    const isPrevented = at(r, "pjax:error");

    if (!isPrevented) {         
       rt && K(null, "", rt.url), window.location.replace(t);
	}
}
```

fregante:
excellent - YEe4Y10PuOTYI

jack1142:
Thanks for the quick fix and review ❤️

arturoiam (commented):
Sorry

PR #3589, opened 2020-09-22
Persist `toggle-files-button` state

### LINKED ISSUE #3588: Make `toggle-files-button` feature persistent across pages
Currently the `toggle-file-button` feature is not extremely useful, as it basically gets you to the readme quicker, and that's it. Ever since a new link popped up on the right side of the page, it might make sense to remove this feature as mentioned in #3580 and #3587.

But I think it would be very nice to keep this feature, and to give more usefulness by making it persist during navigation -- meaning if the file list is toggled off, it will be off when reloading or navigating to a different repo page.

I seldom browse the repo files directly on GitHub, so having the list hidden by default would save quite a bit of scrolling (and yes I do know and use "Hide Files on GitHub", but it doesn't really help with [this kind of repo](https://github.com/nvie/git-toolbelt) for example).

**Feature checklist**:
 1. _It doesn't already exist as a userscript or extension._
 2. _It makes sense as a default for everyone_: if you hide the files on a repo page, there's a good chance you might want it hidden on the other pages as well. Otherwise if you just want to jump to the readme, there's a link for that.
 3. _It doesn't require options_: if the list is open, it stays open ; if it's closed it stays closed.
 4. _It doesn't require a lot of view/UI, code and effort to maintain_: no new UI, but the business logic might be a tad tricky (I tried a naive implementation by adding a check in [`init()`](https://github.com/sindresorhus/refined-github/blob/835b0a0d6d1b5cbb60e7a93cf43b3535602e599e/source/features/toggle-files-button.tsx#L36) but it creates some annoying jitter on page load)
 5. _It doesn't require an external server._
 6. _It doesn't require too much HTTP traffic._
 7. _It doesn't enhance "Projects"_

I'm willing to try implementing this if you accept it as a valid feature request.

Thank you for maintaining and improving this indispensable extension! 👍

cheap-glitch:
Thanks for contributing! 🍄

1. LINKED ISSUES: Closes #3588 

3. SCREENSHOT:
Peek 2020-09-22 20-00

As you can see in the GIF above, I couldn't totally get rid of the content flash. I don't find it too distracting though, so I could live with it. But it's probably possible to do better, given that "Hide Files on GitHub" modifies the file list on load without any noticeable jitters.

### DIFF
--- a/source/features/toggle-files-button.tsx
+++ b/source/features/toggle-files-button.tsx
@@ -12,9 +12,9 @@ import UnfoldIcon from 'octicon/unfold.svg';
 import features from '.';
 import observeElement from '../helpers/simplified-element-observer';
 
-function addButton(): void {
-	console.log('will add');
+const cacheKey = 'files-hidden';
 
+function addButton(): void {
 	// `div` excludes `include-fragment`, which means the list is still loading. #2160
 	const filesHeader = select([
 		'div.commit-tease',
@@ -36,19 +36,21 @@ function addButton(): void {
 	);
 }
 
+async function toggleHandler(): Promise<void> {
+	const isHidden = select('.repository-content')!.classList.toggle('rgh-files-hidden');
+	if (isHidden) {
+		await cache.set(cacheKey, true);
+	} else {
+		await cache.delete(cacheKey);
+	}
+}
+
 async function init(): Promise<void> {
-	const cacheKey = 'hide-file-list';
 	const repoContent = (await elementReady('.repository-content'))!;
 	observeElement(repoContent, addButton);
+	delegate(document, '.rgh-toggle-files', 'click', toggleHandler);
 
-	delegate(document, '.rgh-toggle-files', 'click', async () => {
-		console.log('will toggle');
-		await cache.set(cacheKey, repoContent.classList.toggle('rgh-files-hidden'));
-	});
-
-	console.log('check closed');
 	if (await cache.get<boolean>(cacheKey)) {
-		console.log('was closed');
 		repoContent.classList.add('rgh-files-hidden');
 	}
 }
--- a/source/features/toggle-files-button.tsx
+++ b/source/features/toggle-files-button.tsx
@@ -42,8 +42,9 @@ async function init(): Promise<void> {
 	delegate(document, '.rgh-toggle-files', 'click', async () => {
 		await cache.set(cacheKey, repoContent.classList.toggle('rgh-files-hidden'));
 	});
+
 	if (await cache.get<boolean>(cacheKey)) {
-		select('.rgh-toggle-files')!.click();
+		repoContent.classList.toggle('rgh-files-hidden');
 	}
 }
 
--- a/source/features/toggle-files-button.tsx
+++ b/source/features/toggle-files-button.tsx
@@ -40,16 +40,14 @@ async function init(): Promise<void> {
 	observeElement(repoContent, addButton);
 
 	const cacheKey = 'files-list-toggled-off';
-	if (await cache.get<boolean>(cacheKey)) {
-		repoContent.classList.toggle('rgh-files-hidden', true);
-		select('.rgh-toggle-files')!.setAttribute('aria-expanded', 'false');
-	}
-
 	delegate(document, '.rgh-toggle-files', 'click', async ({delegateTarget}) => {
 		const toggleState = repoContent.classList.toggle('rgh-files-hidden');
 		delegateTarget.setAttribute('aria-expanded', String(!toggleState));
 		await cache.set(cacheKey, toggleState);
 	});
+	if (await cache.get<boolean>(cacheKey)) {
+		select('.rgh-toggle-files')!.click();
+	}
 }
 
 void features.add({
--- a/source/features/toggle-files-button.tsx
+++ b/source/features/toggle-files-button.tsx
@@ -60,5 +60,6 @@ void features.add({
 	include: [
 		pageDetect.isRepoTree
 	],
+	awaitDomReady: false,
 	init
 });
--- a/source/features/toggle-files-button.tsx
+++ b/source/features/toggle-files-button.tsx
@@ -40,12 +40,11 @@ async function init(): Promise<void> {
 	observeElement(repoContent, addButton);
 
 	delegate(document, '.rgh-toggle-files', 'click', async () => {
-		// Persist a closed list or remove from cache
-		await cache.set(cacheKey, (repoContent.classList.toggle('rgh-files-hidden') || undefined)!);
+		await cache.set(cacheKey, repoContent.classList.toggle('rgh-files-hidden'));
 	});
 
 	if (await cache.get<boolean>(cacheKey)) {
-		repoContent.classList.toggle('rgh-files-hidden');
+		select('.rgh-toggle-files')!.click();
 	}
 }
 
--- a/source/features/toggle-files-button.tsx
+++ b/source/features/toggle-files-button.tsx
@@ -58,5 +58,6 @@ void features.add({
 		pageDetect.isRepoTree
 	],
 	awaitDomReady: false,
+	repeatOnBackButton: true,
 	init
 });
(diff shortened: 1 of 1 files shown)


yakov116:
https://github.com/sindresorhus/refined-github/pull/3589/files#diff-cea4665bd7b702cb017621898603e791R62

Set awaitDomReady:false

source/features/toggle-files-button.tsx
```diff
@@ -33,11 +35,20 @@ function addButton(): void {
...
+	const repoContent = (await elementReady('.repository-content'))!;
 	observeElement(repoContent, addButton);
-	delegate(document, '.rgh-toggle-files', 'click', ({delegateTarget}) => {
-		delegateTarget.setAttribute('aria-expanded', String(!repoContent.classList.toggle('rgh-files-hidden')));
+
+	const cacheKey = 'files-list-toggled-off';
+	if (await cache.get<boolean>(cacheKey)) {
+		repoContent.classList.toggle('rgh-files-hidden', true);
+		select('.rgh-toggle-files')!.setAttribute('aria-expanded', 'false');
+	}
```
fregante, line 46: If you move this condition after `delegate`, you can change it to:

```tsx

  	if (await cache.get<boolean>(cacheKey)) {
 		select('.rgh-toggle-files')!.click();
  	}
```

source/features/toggle-files-button.tsx
```diff
@@ -33,11 +35,20 @@ function addButton(): void {
...
+
+	delegate(document, '.rgh-toggle-files', 'click', async ({delegateTarget}) => {
+		const toggleState = repoContent.classList.toggle('rgh-files-hidden');
+		delegateTarget.setAttribute('aria-expanded', String(!toggleState));
+		await cache.set(cacheKey, toggleState);
 	});
 }
```
fregante, line 53: You'll need to add `awaitDomReady: false` to the `features.add`

source/features/toggle-files-button.tsx
```diff
@@ -33,11 +35,20 @@ function addButton(): void {
...
+
+	const cacheKey = 'files-list-toggled-off';
+	if (await cache.get<boolean>(cacheKey)) {
+		repoContent.classList.toggle('rgh-files-hidden', true);
+		select('.rgh-toggle-files')!.setAttribute('aria-expanded', 'false');
+	}
+
+	delegate(document, '.rgh-toggle-files', 'click', async ({delegateTarget}) => {
+		const toggleState = repoContent.classList.toggle('rgh-files-hidden');
+		delegateTarget.setAttribute('aria-expanded', String(!toggleState));
```
fregante, line 50: I think this attribute was not correct/useful. Can you drop it from everywhere in this function?

cheap-glitch: Done!

source/features/toggle-files-button.tsx
```diff
@@ -25,20 +27,24 @@ function addButton(): void {
...
+	const repoContent = (await elementReady('.repository-content'))!;
 	observeElement(repoContent, addButton);
-	delegate(document, '.rgh-toggle-files', 'click', ({delegateTarget}) => {
-		delegateTarget.setAttribute('aria-expanded', String(!repoContent.classList.toggle('rgh-files-hidden')));
+
+	delegate(document, '.rgh-toggle-files', 'click', async () => {
+		await cache.set(cacheKey, repoContent.classList.toggle('rgh-files-hidden'));
 	});
+	if (await cache.get<boolean>(cacheKey)) {
+		select('.rgh-toggle-files')!.click();
```
fregante, line 46: We can return to the plain class toggle now that it's just one instruction anyway
```suggestion

	if (await cache.get<boolean>(cacheKey)) {
		repoContent.classList.toggle('rgh-files-hidden');
```

fregante (approved):
# 🍩

source/features/toggle-files-button.tsx
```diff
@@ -25,20 +27,25 @@ function addButton(): void {
...
-	const repoContent = select('.repository-content')!;
+async function init(): Promise<void> {
+	const cacheKey = 'files-list-toggled-off';
+	const repoContent = (await elementReady('.repository-content'))!;
 	observeElement(repoContent, addButton);
-	delegate(document, '.rgh-toggle-files', 'click', ({delegateTarget}) => {
-		delegateTarget.setAttribute('aria-expanded', String(!repoContent.classList.toggle('rgh-files-hidden')));
+
+	delegate(document, '.rgh-toggle-files', 'click', async () => {
+		await cache.set(cacheKey, repoContent.classList.toggle('rgh-files-hidden'));
```
fregante, line 43: `undefined` will simply drop it from the cache if unnecessary

```suggestion
		// Persist a closed list or remove from cache
		await cache.set(cacheKey, repoContent.classList.toggle('rgh-files-hidden') || undefined);
```

source/features/toggle-files-button.tsx
```diff
@@ -25,20 +27,25 @@ function addButton(): void {
...
 			<FoldIcon/>
 			<UnfoldIcon/>
 		</button>
 	);
 }
 
-function init(): void {
-	const repoContent = select('.repository-content')!;
+async function init(): Promise<void> {
+	const cacheKey = 'files-list-toggled-off';
```
fregante, line 38:
```suggestion
	const cacheKey = 'hide-file-list';
```

source/features/toggle-files-button.tsx
```diff
@@ -49,5 +57,6 @@ void features.add({
 	include: [
 		pageDetect.isRepoTree
 	],
+	awaitDomReady: false,
```
yakov116, line 60:
```suggestion
	awaitDomReady: false,
	repeatOnBackButton: true,
```

This is needed if not the toggle does not work once the back is pressed

yakov116 (approved):
Tested works great
Just needs that one change

source/features/toggle-files-button.tsx
cheap-glitch: Done!

yakov116:
I am having an issue that it always stays closed.

Maybe because the delegate is inline?

cheap-glitch:
> I am having an issue that it always stays closed.
> 
> Maybe because the delegate is inline?

Turns out passing `undefined` to `cache.set()` per @fregante's suggestion did not actually remove the cached value. From the readme of https://github.com/fregante/webext-storage-cache:
> undefined won't be cached, it's considered "no value" in the Storage API.

So `hide-file-list` stayed `true`.

Also a line had been reversed by an auto-commit from suggested changes (probably my bad, should have been more attentive).

fregante:
> Turns out passing `undefined` to `cache.set()` per @fregante's suggestion did not actually remove the cached value

Indeed, I opened https://github.com/fregante/webext-storage-cache/issues/25

fregante (commented):
I checked the latest version and noticed that the listeners pile up after changing pages as you can see with the logs in my latest test commit. I'm working on it.

fregante:
It should work now, sorry for the churn

source/features/toggle-files-button.tsx
```diff
@@ -25,20 +29,30 @@ function addButton(): void {
...
 			<FoldIcon/>
 			<UnfoldIcon/>
 		</button>
 	);
 }
 
-function init(): void {
-	const repoContent = select('.repository-content')!;
+async function toggleHandler(): Promise<void> {
+	const isHidden = select('.repository-content')!.classList.toggle('rgh-files-hidden');
```
yakov116, line 40:
```suggestion
	// Todo use `undefined` after https://github.com/fregante/webext-storage-cache/issues/25
	const isHidden = select('.repository-content')!.classList.toggle('rgh-files-hidden');
```

I think we should add a todo so we don't forget about it

fregante: I already added the condition, there’s no need. 

Also use the same style as we always use: `TODO:`

fregante:
Thanks @cheap-glitch! 👨‍💻

PR #3594, opened 2020-09-25
Enhance `view-markdown-source` with line actions

### LINKED ISSUE #2542: Enhance `view-markdown-source` with line actions
Currently within the [`source/features/view-markdown-source.tsx`](https://github.com/sindresorhus/refined-github/blob/3d517d6cd00e83a51973808c7ff03cae2b692d2b/source/features/view-markdown-source.tsx) the plugin does not add the `...` menu on line highlight. 

Before:
Screen Shot 2019-11-14 at 3 11 01 PM

After:
Screen Shot 2019-11-14 at 3 10 26 PM

Example URL:
https://github.com/sindresorhus/refined-github/blob/3d517d6cd00e83a51973808c7ff03cae2b692d2b/contributing.md

yakov116:
1. LINKED ISSUES:
   Closes #2542 

2. TEST URLS:
   https://github.com/sindresorhus/refined-github/blob/3d517d6cd00e83a51973808c7ff03cae2b692d2b/contributing.md#L7

3. SCREENSHOT:
Video_2020-09-25_100744

### DIFF
--- a/source/features/view-markdown-source.tsx
+++ b/source/features/view-markdown-source.tsx
@@ -14,69 +14,66 @@ import fetchDom from '../helpers/fetch-dom';
 import GitHubURL from '../github-helpers/github-url';
 import {getRepoURL} from '../github-helpers';
 
-const lineActions = onetime((): JSX.Element => {
-	const blame = new GitHubURL(location.href).assign({route: 'blame'}).toString();
-	return (
-		<details
-			className="details-reset details-overlay BlobToolbar position-absolute js-file-line-actions dropdown d-none"
-			aria-hidden="true"
+const lineActions = onetime(() => (
+	<details
+		className="details-reset details-overlay BlobToolbar position-absolute js-file-line-actions dropdown d-none"
+		aria-hidden="true"
+	>
+		<summary
+			className="btn-octicon ml-0 px-2 p-0 bg-white border border-gray-dark rounded-1"
+			aria-label="Inline file action toolbar"
+			aria-haspopup="menu"
+			role="button"
 		>
-			<summary
-				className="btn-octicon ml-0 px-2 p-0 bg-white border border-gray-dark rounded-1"
-				aria-label="Inline file action toolbar"
-				aria-haspopup="menu"
-				role="button"
+			<KebabHorizontalIcon/>
+		</summary>
+		<details-menu role="menu">
+			<ul
+				className="BlobToolbar-dropdown dropdown-menu dropdown-menu-se mt-2"
+				style={{width: '185px'}}
 			>
-				<KebabHorizontalIcon/>
-			</summary>
-			<details-menu role="menu">
-				<ul
-					className="BlobToolbar-dropdown dropdown-menu dropdown-menu-se mt-2"
-					style={{width: '185px'}}
-				>
-					<li>
-						<clipboard-copy
-							role="menuitem"
-							className="dropdown-item zeroclipboard-link"
-							id="js-copy-lines"
-						>
-							Copy line
-						</clipboard-copy>
-					</li>
-					<li>
-						<clipboard-copy
-							role="menuitem"
-							className="dropdown-item zeroclipboard-link"
-							id="js-copy-permalink"
-						>
-							Copy permalink
-						</clipboard-copy>
-					</li>
-					<li>
-						<a
-							className="dropdown-item js-update-url-with-hash"
-							id="js-view-git-blame"
-							role="menuitem"
-							href={blame}
-						>
-							View git blame
-						</a>
-					</li>
-					<li>
-						<a
-							className="dropdown-item"
-							id="js-new-issue"
-							role="menuitem"
-							href={`/${getRepoURL()}/issues/new`}
-						>
-							Reference in new issue
-						</a>
-					</li>
-				</ul>
-			</details-menu>
-		</details>
-	);
-});
+				<li>
+					<clipboard-copy
+						role="menuitem"
+						className="dropdown-item zeroclipboard-link"
+						id="js-copy-lines"
+					>
+						Copy line
+					</clipboard-copy>
+				</li>
+				<li>
+					<clipboard-copy
+						role="menuitem"
+						className="dropdown-item zeroclipboard-link"
+						id="js-copy-permalink"
+					>
+						Copy permalink
+					</clipboard-copy>
+				</li>
+				<li>
+					<a
+						className="dropdown-item js-update-url-with-hash"
+						id="js-view-git-blame"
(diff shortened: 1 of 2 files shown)


source/features/view-markdown-source.tsx
```diff
@@ -5,9 +5,75 @@ import delegate from 'delegate-it';
...
+						</a>
+					</li>
+					<li>
+						<a
+							className="dropdown-item"
+							id="js-new-issue"
+							role="menuitem"
+							href={`https://github.com/${getRepoURL()}/issues/new`}
+						>Reference in new issue
+						</a>
```
fregante, line 70: To all:
```suggestion
						>
							Reference in new issue
						</a>
```

source/features/view-markdown-source.tsx
```diff
@@ -5,9 +5,75 @@ import delegate from 'delegate-it';
...
+							href={blame}
+						>View git blame
+						</a>
+					</li>
+					<li>
+						<a
+							className="dropdown-item"
+							id="js-new-issue"
+							role="menuitem"
+							href={`https://github.com/${getRepoURL()}/issues/new`}
```
fregante, line 68:
```suggestion
							href={`/${getRepoURL()}/issues/new`}
```

source/features/view-markdown-source.tsx
```diff
@@ -5,9 +5,75 @@ import delegate from 'delegate-it';
...
+						>
+							Copy permalink
+						</clipboard-copy>
+					</li>
+					<li>
+						<a
+							className="dropdown-item js-update-url-with-hash"
+							id="js-view-git-blame"
+							role="menuitem"
+							href={blame}
```
fregante, line 59: Doesn't GitHub automatically set this?

source/features/view-markdown-source.tsx
```diff
@@ -5,9 +5,75 @@ import delegate from 'delegate-it';
...
+							style={{cursor: 'pointer'}}
+						>Copy line
+						</clipboard-copy>
+					</li>
+					<li>
+						<clipboard-copy
+							role="menuitem"
+							className="dropdown-item"
+							id="js-copy-permalink"
+							style={{cursor: 'pointer'}}
```
fregante, line 49: If this is required, just add `clipboard-copy {cursor: pointer}` to `refined-github.css`
```suggestion
```

source/features/view-markdown-source.tsx
```diff
@@ -53,6 +119,7 @@ async function showSource(): Promise<void> {
 	sourceButton.classList.add('selected');
 	renderedButton.classList.remove('selected');
 	blurButton(sourceButton);
+	select('.blob-wrapper')!.before(lineActions());
```
fregante, line 122: We already have this element, I think it's
```suggestion
	(await source).before(lineActions());
```

Also, what happens if I toggle between source and rendered multiple times? It looks like it's just going to add it multiple times.

If you wrap `lineActions` with `onetime`, you can add this in `showRendered` I think:

```js
lineActions().remove();
```

You can find this pattern in other parts of RG

fregante (changes requested):

source/features/view-markdown-source.tsx
yakov116: No they only change the hash

source/features/view-markdown-source.tsx
```diff
@@ -1,13 +1,79 @@
...
+					>
+						Copy permalink
+					</clipboard-copy>
+				</li>
+				<li>
+					<a
+						className="dropdown-item js-update-url-with-hash"
+						id="js-view-git-blame"
+						role="menuitem"
+						href={new GitHubURL(location.href).assign({route: 'blame'}).toString()}
```
fregante, line 58: I'm not a fan of `.toString()`
```suggestion
						href={new GitHubURL(location.href).assign({route: 'blame'}).href}
```

source/globals.d.ts
```diff
@@ -57,6 +57,7 @@ declare namespace JSX {
 		'details-menu': BaseIntrinsicElement & {src?: string; preload?: boolean};
 		'time-ago': BaseIntrinsicElement & {datetime: string; format?: string};
 		'relative-time': BaseIntrinsicElement & {datetime: string};
+		'clipboard-copy': BaseIntrinsicElement & {role?: string};
```
fregante, line 60: `role` is a native attribute though, perhaps it should extend `HTMLButtonElement` instead of `div`

fregante (approved):
Other than that, LGTM, merge at will.

source/globals.d.ts
yakov116: hmmm
```
'clipboard-copy' components don't accept text as child elements. Text in JSX has the type 'string', but the expected type of 'children' is 'HTMLCollection'.
```

fregante: You can just set `HTMLButtonElement` there, it's probably `IntrinsicElements['button']`

yakov116:
@fregante great job! (It's all your code 😀)

fregante:
I'd like to _blame_ my code because this is a bad idea. Don't we already have access to this code via the page we're loading?

Currently all unrelated code is being stripped by the second parameter of 

```js
const dom = await fetchDom(path, '.blob-wrapper');
```

but I think that we can avoid having to maintain these 70 lines by using something like:


```js
const page = await fetchDom(path);
lineActions /*global*/ = select('blah blah', page);
const source = select('.blob-wrapper', page);
```

I don't think we even need to customize anything in `lineActions` since it comes straight from GitHub

yakov116:
I can give it a shot. This should be part of November lint 🙄

yakov116:
> Don't we already have access to this code via the page we're loading?

Did some debugging. We dont looks like its loaded loader on

fregante:
Indeed. We fetch the blame page but only the regular source page is supposed to have it, like https://github.com/sindresorhus/refined-github/blob/master/.editorconfig#L3

I've noticed this problem because this feature is currently the longest file and to be fair it's a nightmare just looking at it imagining what class will be dropped next by GitHub.

I wonder if we could just fetch it from a random file:

```js
const lineActions = onetime(() => {
	const bar = fetchDom('https://github.com/sindresorhus/refined-github/blob/b1229bbaeb8cf071f0711bc2ed1b40dd96cd7a05/.editorconfig', 'BlobToolbar');
	select('#js-view-git-blame', bar)!.href = new GitHubURL(location.href).assign({route: 'blame'}).href;
	select('#js-new-issue', bar)!.href = `/${getRepoURL()}/issues/new`;
	return bar;
});
```

We'd need a `!isEnterprise` but the maintainability improvement is worth it.

PR #3598, opened 2020-09-26
Add `table-input` feature

### LINKED ISSUE #2875: Improve <table> input in the comment field
There are many ways to simplify creation of tables. This is the simplest UI I've [seen]((https://www.tablesgenerator.com/html_tables)) but I'm open to suggestions. It's only 2 clicks to add a table.

1. Click on "add table" button (dropdown opens)
1. Click the desired table size

UI

The grid is too big for us, but a 5x5 would be plenty. No need to get fancy with alignment options and headers; those can be more easily added by the user if desired.

Example table usage: https://github.com/sindresorhus/refined-github/issues/2874

Output would be a simplified HTML5 table markup:

```html
<table>
	<tr>
		<td> A1
		<td> B1
	<tr>
		<td> A2
		<td> B2
</table>
```
<table>
	<tr>
		<td> A1
		<td> B1
	<tr>
		<td> A2
		<td> B2
</table>

cheap-glitch:
Thanks for contributing! 🍄

1. LINKED ISSUES: Closes #2875 

2. TEST URLS:
 * https://github.com/sindresorhus/refined-github/issues/new
 * https://github.com/sindresorhus/refined-github/issues/2875

3. SCREENSHOT:
Peek 2020-09-26 19-53

Managed to get this done in just under 80 lines, don't know how it could be reduced but it's probably possible. A few notes/questions:
 * I couldn't find a "table" icon in the octokit, so I used the next best thing (although it's pretty close and with some slight modifications it could make a passable icon).
 * You can't get rid of the table header or the parser won't recognize the table. I chose to add the header on top of whatever size the user selected. Thoughts?
 * The cells have whitespace in them, it makes the table more visually appealing when it's empty but could prove annoying during edition?
 * I didn't use `observeElement` (I got started by copying and hacking [`collapsible-content-button`](https://github.com/sindresorhus/refined-github/blob/master/source/features/collapsible-content-button.tsx) which doesn't use it either). Does it even makes sense to use it here?


source/features/table-input.tsx
```diff
@@ -0,0 +1,79 @@
...
+	});
+}
+
+function getSquarePos(button: HTMLElement): [number, number] {
+	return button.dataset.position!.split(',', 2).map(Number) as [number, number];
+}
+
+function init(): void {
+	delegate(document, '.rgh-add-table-popup__square', 'click', addTable);
+	delegate(document, '.rgh-add-table-popup__square', 'mouseover', highlightSquares);
```
fregante, line 43:
```suggestion
	delegate(document, '.rgh-add-table-popup__square', 'mouseenter', highlightSquares);
```

source/features/table-input.css
```diff
@@ -0,0 +1,27 @@
+.rgh-add-table-popup {
+	width: 108px;
+	height: 108px;
+	padding: 4px;
+	z-index: 99;
+	display: grid;
+	grid-template: repeat(5, 1fr) / repeat(5, 1fr);
+	place-content: center;
```
fregante, line 8: You're already setting width and height of both the container and the children, there's no need to specify another constraint
```suggestion
	display: flex;
	flex-wrap: wrap;
```

source/features/table-input.css
```diff
@@ -0,0 +1,27 @@
...
+}
+
+.rgh-add-table-popup__square {
+	appearance: none;
+	background: none;
+	border: none;
+	padding: 2px;
+}
+
+.rgh-add-table-popup__square > div {
```
fregante, line 18: There's no need for a `div`
```suggestion
	margin: 2px;
```

fregante (changes requested):

source/features/table-input.tsx
```diff
@@ -0,0 +1,79 @@
...
+					role="button" aria-label="Add a table" aria-haspopup="menu"
+				>
+					<TableIcon/>
+				</summary>
+				<details-menu className="select-menu-modal position-absolute left-0 hx_rsm-modal rgh-add-table-popup" role="menu">
+					{Array.from({length: 5}).map((_, y) => Array.from({length: 5}).map((_, x) => (
+						<button
+							type="button"
+							className="rgh-add-table-popup__square"
+							data-position={`${x + 1},${y + 1}`}
```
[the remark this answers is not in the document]
fregante: Better yet, store them separately so you don't need `getSquarePos` at all. TypeScript will still let you use `<=` with strings
```suggestion
							data-x={x + 1}
							data-y={y + 1}
```

source/features/table-input.tsx
```diff
@@ -0,0 +1,79 @@
...
+	field.setSelectionRange(cursorPos, cursorPos);
+}
+
+function highlightSquares(event: delegate.Event<MouseEvent, HTMLButtonElement>): void {
+	const [x, y] = getSquarePos(event.delegateTarget);
+	[...event.delegateTarget.parentElement!.children].forEach(child => {
+		const [cx, cy] = getSquarePos(child as HTMLElement);
+		child.classList.toggle('rgh-add-table-popup__square--selected', cx <= x && cy <= y);
+	});
+}
```
fregante, line 35: If you follow all of my suggestions, this should look like:
```suggestion
function highlightSquares(event: delegate.Event<MouseEvent, HTMLButtonElement>): void {
	const hover = event.delegateTarget;
	for (const cell of event.delegateTarget.parentElement!.children) {
		cell.classList.toggle('selected', cell.dataset.x <= hover.dataset.x && cell.dataset.y <= hover.dataset.y)
	}
}
```

source/features/table-input.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+import features from '.';
+
+function generateGfmTable([w, h]: [number, number]): string {
+	return `${'|     '.repeat(w)}|\n${'|-----'.repeat(w)}|\n${`${'|     '.repeat(w)}|\n`.repeat(h)}`.slice(0, -1);
+}
+
+function addTable(event: delegate.Event<MouseEvent, HTMLButtonElement>): void {
+	event.preventDefault();
+	(event.delegateTarget.parentElement!.parentElement as HTMLDetailsElement)!.open = false;
```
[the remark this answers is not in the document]
cheap-glitch: I tried all of those but it didn't work. The textarea gets focused but the popup stays open, and `field.click()` does nothing.

source/features/table-input.tsx
```diff
@@ -0,0 +1,79 @@
...
+import * as textFieldEdit from 'text-field-edit';
+
+import features from '.';
+
+function generateGfmTable([w, h]: [number, number]): string {
+	return `${'|     '.repeat(w)}|\n${'|-----'.repeat(w)}|\n${`${'|     '.repeat(w)}|\n`.repeat(h)}`.slice(0, -1);
+}
+
+function addTable(event: delegate.Event<MouseEvent, HTMLButtonElement>): void {
+	event.preventDefault();
```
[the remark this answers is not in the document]
cheap-glitch: If this is removed, the form is submitted when a grid cell is clicked. I think this is because the grid cells are `<button>`s.

But now that I think about it there's no reason for them to be buttons in the first place, other than pure semantic correctness...

source/features/table-input.tsx
```diff
@@ -0,0 +1,79 @@
...
+	(event.delegateTarget.parentElement!.parentElement as HTMLDetailsElement)!.open = false;
+
+	const field = event.delegateTarget.form!.querySelector('textarea')!;
+	const endsWithNl = field.value.length && !field.value.endsWith('\n');
+	const cursorPos = field.value.length + (endsWithNl ? 4 : 3);
+
+	textFieldEdit.insert(field, (endsWithNl ? '\n' : '') + generateGfmTable(getSquarePos(event.delegateTarget)));
+
+	field.focus();
+	field.setSelectionRange(cursorPos, cursorPos);
```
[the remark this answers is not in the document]
cheap-glitch: But the first/last `<td>` might not be the one inserted.

Actually right now the table is always appended to the end of the text field. Instead it should go right after the cursor (and then we can use `indexOf('<td>')`).

source/features/table-input.tsx
cheap-glitch: I think I tried `mouseenter` first, but for some reason it didn't work.

source/features/table-input.css
cheap-glitch: If you set a margin directly on the buttons, the user might click between the grid cells and nothing will happen, which is annoying. Also the table size feedback will feel slightly choppy.

By having buttons stuck together with no gaps + decorative divs, we can get a smooth UI while keeping the visual clarity and aesthetic of separate cells. Sure it's a detail but I think it matters :smile:

source/features/table-input.tsx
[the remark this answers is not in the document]
fregante: This feels like something that's handled by the widget automatically. Perhaps we need to add `role=menuitem` or `role=menuitemradio` to each button

I see this code when I select an item from the Canned responses widget:

source/features/table-input.tsx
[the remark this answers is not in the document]
fregante: Try ;)

You're thinking of the native `type=submit` behavior, but we're using `type=button`

source/features/table-input.tsx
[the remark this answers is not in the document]
fregante: This piece of code uses `field.selectionStart` to limit the search near the one inserted.



> Actually right now the table is always appended to the end of the text field

No it isn't for me, that's taken care by `textFieldEdit.insert`:

source/features/table-input.tsx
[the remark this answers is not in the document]
cheap-glitch: > You're thinking of the native `type=submit` behavior, but we're using `type=button`

I know plain buttons aren't supposed to submit forms, but that's what happens nonetheless... 🤷

source/features/table-input.css
fregante: Good point, it does matter!

source/features/table-input.tsx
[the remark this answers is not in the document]
fregante: I just tried removing that and the form is not submitted (Chrome). What browser are you in?

source/features/table-input.tsx
```diff
@@ -0,0 +1,79 @@
...
+function generateGfmTable([w, h]: [number, number]): string {
+	return `${'|     '.repeat(w)}|\n${'|-----'.repeat(w)}|\n${`${'|     '.repeat(w)}|\n`.repeat(h)}`.slice(0, -1);
+}
+
+function addTable(event: delegate.Event<MouseEvent, HTMLButtonElement>): void {
+	event.preventDefault();
+	(event.delegateTarget.parentElement!.parentElement as HTMLDetailsElement)!.open = false;
+
+	const field = event.delegateTarget.form!.querySelector('textarea')!;
+	const endsWithNl = field.value.length && !field.value.endsWith('\n');
```
sindresorhus, line 20:
```suggestion
	const endsWithNewline = field.value.length && !field.value.endsWith('\n');
```

source/features/table-input.tsx
```diff
@@ -0,0 +1,79 @@
...
+	const cursorPos = field.value.length + (endsWithNl ? 4 : 3);
```
sindresorhus, line 21:
```suggestion
	const cursorPosition = field.value.length + (endsWithNl ? 4 : 3);
```

Don't use abbreviations. Applies to all identifiers.

source/features/table-input.tsx
```diff
@@ -0,0 +1,79 @@
...
+	textFieldEdit.insert(field, (endsWithNl ? '\n' : '') + generateGfmTable(getSquarePos(event.delegateTarget)));
+
+	field.focus();
+	field.setSelectionRange(cursorPos, cursorPos);
+}
+
+function highlightSquares(event: delegate.Event<MouseEvent, HTMLButtonElement>): void {
+	const [x, y] = getSquarePos(event.delegateTarget);
+	[...event.delegateTarget.parentElement!.children].forEach(child => {
+		const [cx, cy] = getSquarePos(child as HTMLElement);
```
sindresorhus, line 32: `cx` is not very readable. Use a proper name.

source/features/table-input.tsx
```diff
@@ -0,0 +1,79 @@
...
+		child.classList.toggle('rgh-add-table-popup__square--selected', cx <= x && cy <= y);
+	});
+}
+
+function getSquarePos(button: HTMLElement): [number, number] {
```
sindresorhus, line 37: TypeScript supports named tuples now. Would make this more readable. Same with other places tuples are used.

source/features/table-input.tsx
```diff
@@ -0,0 +1,79 @@
...
+					)))}
+				</details-menu>
+			</details>
+		);
+	}
+}
+
+void features.add({
+	id: __filebasename,
+	description: 'Adds a popup in the toolbar of the enriched text editor to insert a GFM table.',
```
sindresorhus, line 72: I think the description here could be better written.

sindresorhus (changes requested):

source/features/table-input.tsx
fregante: I think the code covered by your first 3 comments will be replaced by: https://github.com/sindresorhus/refined-github/pull/3598#discussion_r495484541

But I agree with the sentiment

source/features/table-input.tsx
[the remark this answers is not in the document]
cheap-glitch: Latest stable Firefox (v81) on Arch Linux.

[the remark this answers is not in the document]
cheap-glitch: Tested again, it is indeed not needed. Maybe it's because I was using `form.click()` to try to close the popup.

source/features/table-input.tsx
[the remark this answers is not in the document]
cheap-glitch: > Perhaps we need to add `role=menuitem`

It works!

source/features/table-input.tsx
cheap-glitch: > `mouseover` is fired every time the mouse moves

Actually that's [`mousemove`](https://developer.mozilla.org/en-US/docs/Web/API/Element/mousemove_event). Tested again with `mouseenter`, doesn't work (because of the embedded `<div>` "covering" each button), so I'll keep `mouseover`.

cheap-glitch:
* Applied most suggestions by @fregante and @sindresorhus 
 * Rewrote the CSS to use Grid (getting Flexbox to make a perfect grid is a pita) but removed duplicated constraints
 * Tables are now closed to prevent successive tables from "swallowing" each other

Peek 2020-09-27 13-21

source/features/table-input.tsx
```diff
@@ -0,0 +1,76 @@
...
+import delegate from 'delegate-it';
+import TableIcon from 'octicon/diff-added.svg';
+import * as pageDetect from 'github-url-detection';
+import * as textFieldEdit from 'text-field-edit';
+
+import features from '.';
+import smartBlockWrap from '../helpers/smart-block-wrap';
+
+type TableDimensions = [number, number];
+function generateHtmlTable([w, h]: TableDimensions): string {
```
sindresorhus, line 13:
```suggestion
function generateHtmlTable([width, height]: TableDimensions): string {
```

source/features/table-input.tsx
```diff
@@ -0,0 +1,76 @@
...
+import select from 'select-dom';
+import delegate from 'delegate-it';
+import TableIcon from 'octicon/diff-added.svg';
+import * as pageDetect from 'github-url-detection';
+import * as textFieldEdit from 'text-field-edit';
+
+import features from '.';
+import smartBlockWrap from '../helpers/smart-block-wrap';
+
+type TableDimensions = [number, number];
```
sindresorhus, line 12: Use named tuples.

source/features/table-input.tsx
```diff
@@ -0,0 +1,76 @@
...
+function generateHtmlTable([w, h]: TableDimensions): string {
+	return '<table>\n' + ('<tr>\n' + '\t<td>\n'.repeat(w)).repeat(h) + '</table>\n';
+}
+
+function addTable(event: delegate.Event<MouseEvent, HTMLButtonElement>): void {
+	const field = event.delegateTarget.form!.querySelector('textarea')!;
+	const cursorPosition = field.selectionStart;
+
+	textFieldEdit.insert(field, smartBlockWrap(generateHtmlTable(['x', 'y'].map(dim => event.delegateTarget.dataset[dim]!).map(Number) as TableDimensions), field));
```
sindresorhus, line 21:
```suggestion
	textFieldEdit.insert(field, smartBlockWrap(generateHtmlTable(['x', 'y'].map(dimension => event.delegateTarget.dataset[dimension]!).map(Number) as TableDimensions), field));
```

source/features/table-input.tsx
```diff
@@ -0,0 +1,76 @@
...
+				<summary
+					className="text-center menu-target py-2 p-md-1 hx_rsm-trigger ml-1"
+					role="button"
+					aria-label="Add a table"
+					aria-haspopup="menu"
+				>
+					<TableIcon/>
+				</summary>
+				<details-menu className="select-menu-modal position-absolute left-0 hx_rsm-modal rgh-add-table-popup" role="menu">
+					{Array.from({length: 25}).map((_, i) => (
```
sindresorhus, line 50:
```suggestion
					{Array.from({length: 25}).map((_, index) => (
```

source/helpers/smart-block-wrap.ts
```diff
@@ -0,0 +1,20 @@
+// Wraps string in at least 2 \n on each side,
+// as long as the field doesn't already have them.
+// Code adopted/adapted from GitHub.
```
sindresorhus, line 3: Pick one.

source/helpers/smart-block-wrap.ts
```diff
@@ -0,0 +1,20 @@
+// Wraps string in at least 2 \n on each side,
```
sindresorhus, line 1:
```suggestion
// Wraps string in at least two newlines on each side,
```

sindresorhus (changes requested):

source/features/table-input.tsx
```diff
@@ -0,0 +1,76 @@
...
+					))}
+				</details-menu>
+			</details>
+		);
+	}
+}
+
+void features.add({
+	id: __filebasename,
+	description: 'Adds a special input in the toolbar of the enriched text editor to quickly insert a simplified HTML table.',
```
sindresorhus, line 69:
```suggestion
	description: 'Adds a button in the text editor to quickly insert a simplified HTML table.',
```

source/features/table-input.tsx
```diff
@@ -0,0 +1,76 @@
...
+import TableIcon from 'octicon/diff-added.svg';
+import * as pageDetect from 'github-url-detection';
+import * as textFieldEdit from 'text-field-edit';
+
+import features from '.';
+import smartBlockWrap from '../helpers/smart-block-wrap';
+
+type TableDimensions = [number, number];
+function generateHtmlTable([w, h]: TableDimensions): string {
+	return '<table>\n' + ('<tr>\n' + '\t<td>\n'.repeat(w)).repeat(h) + '</table>\n';
```
cheap-glitch, line 14:
```suggestion
	return '<table>\n' + ('<tr>\n' + '\t<td>\n'.repeat(width)).repeat(height) + '</table>\n';
```

source/helpers/smart-block-wrap.ts
cheap-glitch, line 3:
```suggestion
// Code adapted from GitHub.
```

source/features/table-input.tsx
cheap-glitch, line 12:
```suggestion
type TableDimensions = [width: number, height: number];
```

cheap-glitch:
I'm still bugged by the icon. Is there any easy way to modify octokit SVGs?

fregante:
You can customize the SVG directly and inline it instead of importing it. Just pass it through SVGO if you edit it with Illustrator

source/features/table-input.tsx
fregante: IMO it a tuple is named and exported as its own type, it might as well just be an object. 

So either change it to an object or just use it inline where needed without creating a `type`

source/features/table-input.tsx
```diff
@@ -0,0 +1,80 @@
...
+		anchor.after(
+			<details className="details-reset details-overlay flex-auto toolbar-item select-menu select-menu-modal-right hx_rsm">
+				<summary
+					className="text-center menu-target py-2 p-md-1 hx_rsm-trigger ml-1"
+					role="button"
+					aria-label="Add a table"
+					aria-haspopup="menu"
+				>
+					{/* eslint-disable react/no-unknown-property */}
+					<svg width="16" height="16" xmlns="http://www.w3.org/2000/svg" className="octicon octicon-diff-added">
```
fregante, line 46:
```suggestion
					<svg width="16" height="16" xmlns="http://www.w3.org/2000/svg" className="octicon">
```

source/features/table-input.tsx
```diff
@@ -0,0 +1,80 @@
...
+	for (const anchor of select.all('md-task-list')) {
+		anchor.after(
+			<details className="details-reset details-overlay flex-auto toolbar-item select-menu select-menu-modal-right hx_rsm">
+				<summary
+					className="text-center menu-target py-2 p-md-1 hx_rsm-trigger ml-1"
+					role="button"
+					aria-label="Add a table"
+					aria-haspopup="menu"
+				>
+					{/* eslint-disable react/no-unknown-property */}
```
fregante, line 45: The error specifically tells you what to do, don't silence it

```
  ✖  48:713  Unknown property clip-rule found, use clipRule instead        react/no-unknown-property
  ✖  48:733  Unknown property fill-rule found, use fillRule instead        react/no-unknown-property
  ✖  49:64   Unknown property stroke-width found, use strokeWidth instead  react/no-unknown-property
  ✖  50:66   Unknown property stroke-width found, use strokeWidth instead  react/no-unknown-property
```

source/features/table-input.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+function highlightSquares({delegateTarget: hover}: delegate.Event<MouseEvent, HTMLElement>): void {
+	for (const cell of Array.from(hover.parentElement!.children) as HTMLElement[]) {
+		cell.classList.toggle('selected', cell.dataset.x! <= hover.dataset.x! && cell.dataset.y! <= hover.dataset.y!);
+	}
+}
+
+function init(): void {
+	delegate(document, '.rgh-table-input-cell', 'click', addTable);
+	delegate(document, '.rgh-table-input-cell', 'mouseover', highlightSquares);
```
fregante, line 34:
```suggestion
	delegate(document, '.rgh-table-input-cell', 'mouseenter', highlightSquares, {capture: true});
```

source/features/table-input.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+	textFieldEdit.insert(field, smartBlockWrap(generateHtmlTable(Number.parseInt(event.delegateTarget.dataset.x!, 10), Number.parseInt(event.delegateTarget.dataset.y!, 10)), field));
+
+	field.focus();
+	const firstRowPosition = field.value.indexOf('<td>', cursorPosition) + '<td>'.length;
+	field.setSelectionRange(firstRowPosition, firstRowPosition);
+}
+
+function highlightSquares({delegateTarget: hover}: delegate.Event<MouseEvent, HTMLElement>): void {
+	for (const cell of Array.from(hover.parentElement!.children) as HTMLElement[]) {
```
fregante, line 27:
```suggestion
	for (const cell of hover.parentElement!.children as HTMLCollectionOf<HTMLButtonElement>) {
```

source/features/table-input.tsx
fregante: Don't "resolve" suggestions when they aren't resolved because then I have to go find it again and re-open it.

`mouseenter` works, you just need to use `{capture: true}` as an additional parameter: https://stackoverflow.com/questions/50177348/mouseenter-delegation-using-vanilla-javascript

```js
	delegate(document, '.rgh-add-table-popup__square', 'mouseenter', highlightSquares, {capture: true});
```

source/features/table-input.css
fregante: Same for this. Flex isn't pita, it requires less code than before: c54d1e12

Also in the width wasn't applied in Chome:






That commit also fixed it

fregante (changes requested):

source/features/table-input.tsx
cheap-glitch: Sorry about that. But what does it matter if we use `mouseover` instead of `mouseenter` here? (not being dismissive, this is a real question)

source/features/table-input.tsx
cheap-glitch: I silenced it because `strokeWidth` and al. weren't converted in their snake-case equivalents during build. Is there another way to circumvent this?

source/features/table-input.tsx
```diff
@@ -0,0 +1,80 @@
...
+					aria-haspopup="menu"
+				>
+					{/* eslint-disable react/no-unknown-property */}
+					<svg width="16" height="16" xmlns="http://www.w3.org/2000/svg" className="octicon octicon-diff-added">
+						<path d="m13.25,2.5l-10.5,0c-0.13807,0 -0.25,0.11193 -0.25,0.25l0,10.5c0,0.1381 0.11193,0.25 0.25,0.25l10.5,0c0.1381,0 0.25,-0.1119 0.25,-0.25l0,-10.5c0,-0.13807 -0.1119,-0.25 -0.25,-0.25zm-10.5,-1.5l10.5,0c0.9665,0 1.75,0.7835 1.75,1.75l0,10.5c0,0.9665 -0.7835,1.75 -1.75,1.75l-10.5,0c-0.9665,0 -1.75,-0.7835 -1.75,-1.75l0,-10.5c0,-0.9665 0.7835,-1.75 1.75,-1.75zm5.25,3c0.41421,0 0.75,0.33579 0.75,0.75l0,2.5l2.5,0c0.4142,0 0.75,0.33579 0.75,0.75c0,0.41421 -0.3358,0.75 -0.75,0.75l-2.5,0l0,2.5c0,0.4142 -0.33579,0.75 -0.75,0.75c-0.41421,0 -0.75,-0.3358 -0.75,-0.75l0,-2.5l-2.5,0c-0.41421,0 -0.75,-0.33579 -0.75,-0.75c0,-0.41421 0.33579,-0.75 0.75,-0.75l2.5,0l0,-2.5c0,-0.41421 0.33579,-0.75 0.75,-0.75z" clip-rule="evenodd" fill-rule="evenodd"/>
+						<line y2="14.168" x2="8.00791" y1="1.80263" x1="7.99209" stroke-width="1.6" stroke="currentColor"/>
+						<line y2="8.00113" x2="14.16224" y1="8.00113" x1="1.76197" stroke-width="1.6" stroke="currentColor"/>
+					</svg>
```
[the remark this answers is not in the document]
cheap-glitch: Yes it's a pretty terrible attempt. Bear with me on this 😄

source/features/table-input.tsx
fregante: Indeed, that's a dom-chef bug. Let's mention it:
```suggestion
					{/* eslint-disable react/no-unknown-property -- TODO: until https://github.com/vadimdemedes/dom-chef/issues/72*/}
```

source/features/table-input.tsx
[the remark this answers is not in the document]
cheap-glitch: Fixed the icon to be a single path. It looks better and as a bonus we can drop the eslint comments since the `no-unknown-property` rule isn't triggered anymore.

source/features/table-input.tsx
cheap-glitch: Applied the change to move things along, but I'm still curious why `mouseenter` was a better choice here than `mouseover`.

cheap-glitch:
Uploading a GIF for the feature demo

Peek 2020-09-29 14-34

source/features/table-input.tsx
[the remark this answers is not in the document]
fregante: Neato!

source/features/table-input.tsx
fregante: Mousemove is fired every time the mouse moves. Delegate still has to filter all those events through the selector. 

Mouseenter is the one that fires when you enter an element, so *it’s the right tool for the job* in the first place 😁

fregante (approved):
Merci!

source/features/table-input.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+function highlightSquares({delegateTarget: hover}: delegate.Event<MouseEvent, HTMLElement>): void {
+	for (const cell of hover.parentElement!.children as HTMLCollectionOf<HTMLButtonElement>) {
+		cell.classList.toggle('selected', cell.dataset.x! <= hover.dataset.x! && cell.dataset.y! <= hover.dataset.y!);
+	}
+}
+
+function init(): void {
+	delegate(document, '.rgh-table-input-cell', 'click', addTable);
+	delegate(document, '.rgh-table-input-cell', 'mouseenter', highlightSquares, {capture: true});
```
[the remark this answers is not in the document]
cheap-glitch: Tested in latest Firefox on Arch Linux, didn't see any errors.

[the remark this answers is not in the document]
fregante: https://github.com/fregante/delegate-it/issues/20

KaKi87:
Hello,
Why HTML instead of Markdown ?
Thanks

fregante:
https://github.com/refined-github/refined-github/issues/2875#issuecomment-596369002

KaKi87:
How about letting the user choose, then ?
Thanks

fregante:
> How about letting the user choose, then ?

No options allowed:

- https://github.com/refined-github/refined-github/wiki/%22Can-you-add-this-feature%3F%22#3-it-doesnt-require-options

Also:

- #5934 


### Good news, though

Markdown table are now available natively on GitHub: https://github.com/orgs/community/discussions/40299

ko

`table-input` will stay here however as it's vastly superior in every way (except keyboard accessibility):

PR #3600, opened 2020-09-29
Focus field after using `collapsible-content-button`

cheap-glitch:
Thanks for contributing! 🍄

1. LINKED ISSUES: none

2. TEST URLS: https://github.com/sindresorhus/refined-github/issues/new

While working on #3598 I noticed a small bug in `collapsible-content-button.tsx`: there was a few lines of code to restore the selection after wrapping it in a `<details>` tag, but it had no effect (at least on latest Firefox):
without-focus

This can be fixed by focusing the textarea before restoring the selection:
with-focus

### DIFF
--- a/source/features/collapsible-content-button.tsx
+++ b/source/features/collapsible-content-button.tsx
@@ -46,6 +46,7 @@ function addContentToDetails(event: delegate.Event<MouseEvent, HTMLButtonElement
 
 	// Restore selection.
 	// `selectionStart` will be right after the newly-inserted text
+	field.focus();
 	field.setSelectionRange(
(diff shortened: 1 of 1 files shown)


busches:
Am I crazy or are the two gifs the same?

cheap-glitch:
> Am I crazy or are the two gifs the same?

No, in the second one the text stays selected.

fregante:
Thanks! Can you always .focus before the .insert? If a field is not focused, that .insert will focus it and then restore the focus back to where it was.

cheap-glitch:
> Thanks! Can you always .focus before the .insert? If a field is not focused, that .insert will focus it and then restore the focus back to where it was.

Done!

fregante (approved):

PR #3602, opened 2020-09-29
Lint

yakov116:
- Remove duplicate page detect
- Sort types A-Z
- `add-tags-to-commits` => `tags-on-commits-list`
- Ensure descriptions end with a period
- Move `maxAge` to be on one line 
- Update `github-url-detection`

PR #3603, opened 2020-09-30
Drop `remove-upload-files-button` feature

yakov116:
As per lint pr https://github.com/sindresorhus/refined-github/pull/3602#discussion_r497275803

I can't seem to find it in the Readme 🧐


fregante:
Yeah this is one of those ancient features that just tagged along and wasn’t documented properly.

fregante (approved):

PR #3604, opened 2020-09-30
Add `next-scheduled-github-action` feature

### LINKED ISSUE #3104: Show next scheduled time of a GitHub Action
GitHub shows a list of historical runs for a specific deployment in order. Some of these Actions could have been triggered from a scheduled run, with cron. But we do not know when the next scheduled run is going to be.

We could add an information bar here on the top if there is going to be a scheduled job for that Action and when.

image

This feature request is similar to what Travis shows in their cron job settings, like "**Scheduled in 20 hours from now**".

image

This can be computed based on the last _scheduled_ run of an Action added with the schedule values for that Action.

Example: https://github.com/notlmn/copy-as-markdown/actions?query=workflow%3ADeployment

cheap-glitch:
Thanks for contributing! 🍄

1. LINKED ISSUES: Closes #3104 

2. TEST URLS: https://github.com/sindresorhus/refined-github/actions
Couldn't find other repos with scheduled GAs...

3. SCREENSHOT:
image

Probably still far from perfect, but it's working. Notes/questions:
 * I chose 1 day for the caching of the YAML config files, no idea if that's appropriate
 * ~~not sure if `getWorkflows` should return directly the yaml strings instead?~~
 * should the regexes be moved out of the `map()`, or is TS smart enough to optimize this? (or maybe JS engines know this?)
 * should something be displayed next to "All workflows"? Like maybe the next closest action?


source/features/show-github-action-next-time.tsx
```diff
@@ -0,0 +1,71 @@
...
+				... on Tree {
+					entries {
+						object { ... on Blob { text } }
+					}
+				}
+			}
+		}
+	`);
+
+	return workflows;
```
fregante, line 26: Always parse the API responses before returning them so you don’t have to type the entire API (perhaps giving false expectations). You can inline parseYamls in here to do that.

source/features/show-github-action-next-time.tsx
```diff
@@ -0,0 +1,71 @@
...
+	maxAge: {
+		days: 1
+	},
+	cacheKey: () => __filebasename + ':' + getRepoURL()
+});
+
+function parseYamls(yamls: string[]): {[index: string]: string} {
+	return Object.fromEntries(yamls
+		.map(yaml => [(/^name:\s+(\S+)$/m.exec(yaml) ?? [])[1], (/^\s*-\scron:\s+['"](.+)['"]/m.exec(yaml) ?? [])[1]])
+		.filter(([name, cron]) => name !== undefined && cron !== undefined)
```
fregante, line 37: This double loop with long lines is unreadable, it’s best to use a regular “for of”

source/features/show-github-action-next-time.tsx
```diff
@@ -0,0 +1,71 @@
...
+	);
+}
+
+async function init(): Promise<void> {
+	const actionsSidebar = (await elementReady('.hx_actions-sidebar'))!;
+	if (actionsSidebar) {
```
fregante, line 43: Use `if !something, return false` instead of infinitely nesting conditions

source/features/show-github-action-next-time.tsx
```diff
@@ -0,0 +1,71 @@
...
+}, {
+	maxAge: {
+		days: 1
+	},
+	cacheKey: () => __filebasename + ':' + getRepoURL()
+});
+
+function parseYamls(yamls: string[]): {[index: string]: string} {
+	return Object.fromEntries(yamls
+		.map(yaml => [(/^name:\s+(\S+)$/m.exec(yaml) ?? [])[1], (/^\s*-\scron:\s+['"](.+)['"]/m.exec(yaml) ?? [])[1]])
```
fregante, line 36: The value of name can also be quoted; use the same regex as cron’s

source/features/show-github-action-next-time.tsx
```diff
@@ -0,0 +1,71 @@
...
+}
+
+async function init(): Promise<void> {
+	const actionsSidebar = (await elementReady('.hx_actions-sidebar'))!;
+	if (actionsSidebar) {
+		const currentDate = new Date();
+		const workflows = await getWorkflows();
+		if (workflows) {
+			const actionsSchedules = parseYamls(workflows.map(workflow => workflow.object.text));
+			for (const actionListItem of select.all('li > a', actionsSidebar).slice(1)) {
```
fregante, line 48: slice(1) can be expressed via CSS selector if necessary

cheap-glitch:
Thanks for the review! I fixed the code but still have some uncertainties, mainly:
> * I chose 1 day for the caching of the YAML config files, no idea if that's appropriate
> * should something be displayed next to "All workflows"? Like maybe the next closest action?

source/features/show-github-action-next-time.tsx
```diff
@@ -0,0 +1,83 @@
...
+	}
+}
+
+void features.add({
+	id: __filebasename,
+	description: 'Shows the next scheduled time of relevant GitHub Actions in the workflows sidebar.',
+	screenshot: 'https://user-images.githubusercontent.com/46634000/94690232-2476a180-0330-11eb-99d7-e174bb762cea.png'
+}, {
+	include: [
+		pageDetect.isRepo
```
yakov116, line 79:
```suggestion
		pageDetect.isActions
```

cheap-glitch: I used `isRepo()` because pasting `https://github.com/sindresorhus/refined-github/actions` in https://fregante.github.io/github-url-detection/ gives `false` for both `isActions()` and `isActionPage()`.

source/features/show-github-action-next-time.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+import features from '.';
+import * as api from '../github-helpers/api';
+import {getRepoURL, getRepoGQL, getCurrentBranch} from '../github-helpers';
+
+const getActionsSchedules = cache.function(async (): Promise<{[index: string]: string} | undefined> => {
+	const {repository: {object: {entries: actions}}} = await api.v4(
+		`repository(${getRepoGQL()}) {
+			object(expression: "${getCurrentBranch()}:.github/workflows") {
+				... on Tree {
```
fregante, line 17: Does this support using “on Blob” directly? Perhaps “.github/workflows” will return both the folder and the objects in that folder

source/features/show-github-action-next-time.tsx
yakov116: Can you open a issue there?

source/features/show-github-action-next-time.tsx
```diff
@@ -0,0 +1,83 @@
...
+	const currentDate = new Date();
+	for (const actionListItem of select.all('li:not(:first-child) > a', actionsSidebar)) {
+		const actionName = actionListItem.textContent!.trim();
+		if (!(actionName in actionsSchedules)) {
+			continue;
+		}
+
+		const nextTime = parseCron.nextDate(actionsSchedules[actionName], currentDate);
+		if (nextTime) {
+			actionListItem.append(<span className="rgh-github-action-next-time">(next <relative-time datetime={nextTime.toString()}/>)</span>);
```
fregante, line 68: You can split this line into 5 lines for readability 
```suggestion
			actionListItem.append(
				<span className="rgh-github-action-next-time">
					(next <relative-time datetime={nextTime.toString()}/>)
				</span>
			);
```

source/features/show-github-action-next-time.tsx
```diff
@@ -0,0 +1,83 @@
...
+	return schedules;
+}, {
+	maxAge: {
+		days: 1
+	},
+	cacheKey: () => __filebasename + ':' + getRepoURL()
+});
+
+async function init(): Promise<void> {
+	const actionsSidebar = (await elementReady('.hx_actions-sidebar'))!;
```
fregante, line 49: There’s a condition after this, no need to also wrap the type with `()!`

source/features/show-github-action-next-time.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+void features.add({
+	id: __filebasename,
+	description: 'Shows the next scheduled time of relevant GitHub Actions in the workflows sidebar.',
+	screenshot: 'https://user-images.githubusercontent.com/46634000/94690232-2476a180-0330-11eb-99d7-e174bb762cea.png'
+}, {
+	include: [
+		pageDetect.isRepo
+	],
+	awaitDomReady: false,
```
yakov116, line 81:
```suggestion
	exclude: [
		location.pathname.endsWith('actions/new')
	],
	awaitDomReady: false,
```

Don't apply the suggestion, wait for https://github.com/fregante/github-url-detection/issues/40. (I plan on doing it but need to confirm on the name)

With this you should not have to check if `actionsSidebar` exists or not. Since if actions is not setup it redirects to the new page

source/features/show-github-action-next-time.tsx
```diff
@@ -0,0 +1,83 @@
...
+	`);
+
+	if (!actions) {
+		return undefined;
+	}
+
+	const schedules: {[index: string]: string} = {};
+	for (const actionYaml of actions.map((action: {[index: string]: {[index: string]: string}}) => action.object.text)) {
+		const name = /^name:\s+['"]?(.+)['"]?/m.exec(actionYaml);
+		const cron = /^\s*-\scron:\s+['"](.+)['"]/m.exec(actionYaml);
```
fregante, line 33: I don't think it's safe to use the `m` flag or else when will `.+` stop on an unquoted `name`?

```suggestion
		const name = /^name[:\s'"]+(.+)['"]?/.exec(actionYaml);
		const cron = /^\s*-\scron[:\s'"]+(.+)['"]/.exec(actionYaml);
```

This also covers `name : "name"`

source/features/show-github-action-next-time.tsx
```diff
@@ -0,0 +1,83 @@
...
+					entries {
+						object { ... on Blob { text } }
+					}
+				}
+			}
+		}
+	`);
+
+	if (!actions) {
+		return undefined;
```
fregante, line 27: `undefined` means "don't cache it"

Use `false`

source/refined-github.ts
```diff
@@ -198,6 +198,7 @@ import './features/linkify-full-profile-readme-title';
 import './features/same-page-definition-jump';
 import './features/new-repo-disable-projects-and-wikis';
 import './features/link-to-gihub-io';
+import './features/show-github-action-next-time';
```
fregante, line 201:
```suggestion
import './features/next-scheduled-github-action';
```

Every feature "Shows" or "adds" something, so it's implied

source/features/show-github-action-next-time.tsx
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	const currentDate = new Date();
+	for (const actionListItem of select.all('li:not(:first-child) > a', actionsSidebar)) {
+		const actionName = actionListItem.textContent!.trim();
+		if (!(actionName in actionsSchedules)) {
+			continue;
+		}
+
+		const nextTime = parseCron.nextDate(actionsSchedules[actionName], currentDate);
```
fregante, line 66: No need to supply default values. `new Date()` is not that processor-intensive
```suggestion
		const nextTime = parseCron.nextDate(actionsSchedules[actionName]);
```

source/features/show-github-action-next-time.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+async function init(): Promise<void> {
+	const actionsSidebar = (await elementReady('.hx_actions-sidebar'))!;
+	if (!actionsSidebar) {
+		return;
+	}
+
+	const actionsSchedules = await getActionsSchedules();
+	if (!actionsSchedules) {
+		return;
```
fregante, line 56: Return false in `init` because that marks the feature as “not enabled on this page”

fregante (changes requested):

fregante:
> * I chose 1 day for the caching of the YAML config files, no idea if that's appropriate

That's ok, you can also use `staleWhileRevalidate: {days: 10}`

source/features/show-github-action-next-time.tsx
cheap-glitch, line 27:
```suggestion
		return false;
```

source/features/show-github-action-next-time.tsx
```diff
@@ -0,0 +1,83 @@
...
+import select from 'select-dom';
+import elementReady from 'element-ready';
+import * as pageDetect from 'github-url-detection';
+import {parseCron} from '@cheap-glitch/mi-cron';
+
+import features from '.';
+import * as api from '../github-helpers/api';
+import {getRepoURL, getRepoGQL, getCurrentBranch} from '../github-helpers';
+
+const getActionsSchedules = cache.function(async (): Promise<{[index: string]: string} | undefined> => {
```
cheap-glitch, line 13:
```suggestion
const getActionsSchedules = cache.function(async (): Promise<{[index: string]: string} | boolean> => {
```

source/features/show-github-action-next-time.tsx
cheap-glitch, line 49:
```suggestion
	const actionsSidebar = await elementReady('.hx_actions-sidebar');
```

source/features/show-github-action-next-time.tsx
cheap-glitch, line 56:
```suggestion
		return false;
```

source/features/show-github-action-next-time.tsx
```diff
@@ -0,0 +1,83 @@
...
+	if (!actionsSidebar) {
+		return;
+	}
+
+	const actionsSchedules = await getActionsSchedules();
+	if (!actionsSchedules) {
+		return;
+	}
+
+	const currentDate = new Date();
```
cheap-glitch, line 59:
```suggestion
```

fregante:
> * should the regexes be moved out of the `map()`, or is TS smart enough to optimize this?

That's a micro-optimization that we normally don't do



> * should something be displayed next to "All workflows"? Like maybe the next closest action?

Doesn't seem necessary, there's the list right below it already

fregante:
If you're gonna make a lot of changes (in this case I see 5 review comments), just edit the file; there are already plenty of reviews on the page :smile:

source/features/show-github-action-next-time.tsx
```diff
@@ -0,0 +1,86 @@
...
+import select from 'select-dom';
+import elementReady from 'element-ready';
+import * as pageDetect from 'github-url-detection';
+import {parseCron} from '@cheap-glitch/mi-cron';
+
+import features from '.';
+import * as api from '../github-helpers/api';
+import {getRepoURL, getRepoGQL, getCurrentBranch} from '../github-helpers';
+
+const getActionsSchedules = cache.function(async (): Promise<{[index: string]: string} | boolean> => {
```
fregante: In both places
```suggestion
const getActionsSchedules = cache.function(async (): Promise<Record<string, string> | boolean> => {
```

source/features/show-github-action-next-time.tsx
cheap-glitch: Tested in the GraphQL explorer, couldn't get it to work without `...on Tree`. Then again, I'm a newbie at GraphQL.

source/features/next-scheduled-github-action.tsx
```diff
@@ -0,0 +1,85 @@
...
+	for (const actionListItem of select.all('li:not(:first-child) > a', actionsSidebar)) {
+		const actionName = actionListItem.textContent!.trim();
+		if (!(actionName in actionsSchedules)) {
+			continue;
+		}
+
+		const nextTime = parseCron.nextDate(actionsSchedules[actionName]);
+		if (nextTime) {
+			actionListItem.append(
+				<span className="rgh-github-action-next-time">
```
fregante, line 67:
```suggestion
				<em>
```

So you can drop the CSS

source/features/next-scheduled-github-action.tsx
```diff
@@ -0,0 +1,85 @@
...
+	if (!actionsSidebar) {
+		return false;
+	}
+
+	const actionsSchedules = await getActionsSchedules();
+	if (!actionsSchedules) {
+		return false;
+	}
+
+	for (const actionListItem of select.all('li:not(:first-child) > a', actionsSidebar)) {
```
fregante, line 58: This selector can be more specific. And also since the filter will be more specific with https://github.com/fregante/github-url-detection/issues/40 we can also move the elementReady here and expect it to exist

```suggestion
	for (const actionListItem of select.all('[href*="?query"]', await elementReady('.hx_actions-sidebar'))) {
```

source/features/next-scheduled-github-action.tsx
```diff
@@ -0,0 +1,85 @@
...
+import * as pageDetect from 'github-url-detection';
+import {parseCron} from '@cheap-glitch/mi-cron';
+
+import features from '.';
+import * as api from '../github-helpers/api';
+import {getRepoURL, getRepoGQL} from '../github-helpers';
+
+const getActionsSchedules = cache.function(async (): Promise<{[index: string]: string} | false> => {
+	const {repository: {object: {entries: actions}}} = await api.v4(
+		`repository(${getRepoGQL()}) {
```
fregante, line 15:
```suggestion
	const {repository: {object: {entries: actions}}} = await api.v4(`
		repository(${getRepoGQL()}) {
```

fregante (commented):
Looking good so far!

fregante:
Converting to draft while we wait for https://github.com/fregante/github-url-detection/issues/40

source/features/next-scheduled-github-action.tsx
```diff
@@ -0,0 +1,75 @@
...
+	`);
+	if (!actions) {
+		return false;
+	}
+
+	const schedules: Record<string, string> = {};
+	for (const action of actions) {
+		const actionYaml = action.object.text;
+		const name = /^name:\s+['"]?(.+)['"]?/m.exec(actionYaml);
+		const cron = /^\s*-\scron:\s+['"](.+)['"]/m.exec(actionYaml);
```
fregante, line 32: As previously noted, does hit work:
```suggestion
		const name = /^name[:\s'"]+(.+)['"]?/.exec(actionYaml);
		const cron = /^\s*-\s*cron[:\s'"]+(.+)['"]/.exec(actionYaml);
```

Also

source/features/next-scheduled-github-action.tsx
```diff
@@ -0,0 +1,75 @@
...
+import elementReady from 'element-ready';
+import * as pageDetect from 'github-url-detection';
+import {parseCron} from '@cheap-glitch/mi-cron';
+
+import features from '.';
+import * as api from '../github-helpers/api';
+import {getRepoURL, getRepoGQL} from '../github-helpers';
+
+const getActionsSchedules = cache.function(async (): Promise<Record<string, string> | false> => {
+	const {repository: {object: {entries: actions}}} = await api.v4(`
```
fregante, line 13: These are not actions but workflows. This applies to every place where we call them `actions` (except in the name because that's the official name of "GitHub Actions")
```suggestion
	const {repository: {object: {entries: workflow}}} = await api.v4(`
```

source/features/next-scheduled-github-action.tsx
```diff
@@ -0,0 +1,75 @@
...
+import select from 'select-dom';
+import elementReady from 'element-ready';
+import * as pageDetect from 'github-url-detection';
+import {parseCron} from '@cheap-glitch/mi-cron';
+
+import features from '.';
+import * as api from '../github-helpers/api';
+import {getRepoURL, getRepoGQL} from '../github-helpers';
+
+const getActionsSchedules = cache.function(async (): Promise<Record<string, string> | false> => {
```
fregante, line 12:
```suggestion
const getScheduledWorkflows = cache.function(async (): Promise<Record<string, string> | false> => {
```

readme.md
```diff
@@ -132,7 +132,8 @@ Thanks for contributing! 🦋🙌
...
 - [](# "new-repo-disable-projects-and-wikis") [Automatically disables projects and wikis when creating a repository.](https://user-images.githubusercontent.com/16872793/92803886-dc460e00-f385-11ea-8af6-d6b7a0d3bf91.png)
 - [](# "sticky-sidebar") [Makes conversation sidebars and repository sidebars sticky, if they fit the viewport.](https://user-images.githubusercontent.com/10238474/62276723-5a2eaa80-b44d-11e9-810b-ff598d1c5c6a.gif)
-- [](# "link-to-github-io") [Add a link to visit the user’s github.io website from its repo.](https://user-images.githubusercontent.com/31387795/94045261-dbcd5e80-fdec-11ea-83fa-30bb673cc26e.jpg)
+- [](# "link-to-github-io") [Adds a link to visit the user’s github.io website from its repo.](https://user-images.githubusercontent.com/31387795/94045261-dbcd5e80-fdec-11ea-83fa-30bb673cc26e.jpg)
+- [](# "show-github-action-next-time") [Shows the next scheduled time of relevant GitHub Actions in the workflows sidebar.](https://user-images.githubusercontent.com/46634000/94690232-2476a180-0330-11eb-99d7-e174bb762cea.png)
```
yakov116, line 136: The feature name needs to be updated here

source/features/next-scheduled-github-action.tsx
```diff
@@ -0,0 +1,79 @@
...
+}
+
+void features.add({
+	id: __filebasename,
+	description: 'Shows the next scheduled time of relevant GitHub Actions in the workflows sidebar.',
+	screenshot: 'https://user-images.githubusercontent.com/46634000/94690232-2476a180-0330-11eb-99d7-e174bb762cea.png'
+}, {
+	include: [
+		pageDetect.isRepo
+	],
```
yakov116, line 76:
```suggestion
	include: [
		pageDetect.isRepositoryActions
	],
	exclude: [
		pageDetect.isNewAction
	],
```

source/features/next-scheduled-github-action.tsx
cheap-glitch: I used the `m` flag because I wanted to anchor the regex to the beginning of the line. It works if you remove both `^` and `m`, but that means it could technically match "name: xxx" in a comment or something before the actual name declaration (same for `cron: xxx`).

fregante: > I used the `m` flag because I wanted to anchor the regexes to the beginning of the line. It works if you remove both `^` and `m`

Oh wow I think I misunderstood the `m` flag, I thought it was the opposite. I was expecting `^` to be the beginning of the line already like it is in my editor.



> Is it valid YAML to have a space before `:`?

Yep, even a tab

fregante: Updated suggestion: 
```suggestion
		const name = /^name[:\s'"]+(.+)['"]?/m.exec(actionYaml);
		const cron = /^[\s-]+cron[:\s'"]+(.+)['"]/m.exec(actionYaml);
```

fregante: Perhaps the `cron` regex should be updated to include `schedule` in it, like `/schedule[:\n\s-]+cron`

cheap-glitch: I have to admit I'm not a fan of `[:\s'"]+`. It works fine but feels a little "hackish". Something like `\s*:\s+['"]?` may be a little more verbose but it makes sense at first glance. Plus the first one matches stuff like `name::::: Action" which is a little weird. But in the end it's a matter of preference so I'll leave it up to you.

> Perhaps the cron regex should be updated to include schedule in it

I agree, imo the more specific the better in this case

fregante: > Plus the first one matches stuff like `name::::: Action" which is a little weird

We're not here to validate YML, we just need to get the value and get out. `\s*:\s+['"]?` is way too verbose and it's wrong because it won't catch `name:Test`. Less noisy regexes are preferred. If we had to be _exact_ we'd use a YML parser, but we don't do that.

cheap-glitch: Applied the suggestions and added `schedule[:\s-]` to the cron regex (removed the `\n` as it's already included in `\s`).

fregante:
## :woot:

fregante:
FYI workflows can now be disabled, so this feature should not add the next schedule on disabled jobs

PR #3608, opened 2020-10-01
Fix toolbar padding on milestone page

### LINKED ISSUE #3606: On milestones page, the bulk issue-update toolbar has incorrect padding that breaks vertical centering
Here's an example of this kind of page https://github.com/learningequality/kolibri/milestone/60.

image

If you click any checkbox, the menu a the top right of this div will change the bulk-update menu, and as you can see in the screenshot, the menu of dropdowns is missing some padding.

I tried to fix this by adding this custom CSS (copying the selector from the DOM inspector)

```css
#js-issues-toolbar > div.table-list-triage.js-issues-toolbar-triage > div {
	padding: 13px;
}
```

But this messes with the same menu on the main issues page:

Next screenshot is from https://github.com/learningequality/kolibri/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc (click any checkbox to see this; without the above custom CSS, you won't see this excessive padding).

image

So I added a second rule to override the first one

```css
#js-issues-toolbar > div.table-list-triage.flex-auto.js-issues-toolbar-triage > div {
	padding: 0 !important;
}
```

And this combination of custom CSS seems to fix the layout issues for me.

cheap-glitch:
Thanks for contributing! 🍄

1. LINKED ISSUES: Closes #3606 

2. TEST URLS:
- https://github.com/learningequality/kolibri/milestone/60
- https://github.com/cheap-glitch/fretboarder/milestone/1

3. SCREENSHOT:

Before:
Screenshot_2020-10-01 cheap-glitch fretboarder

After:
Screenshot_2020-10-01 cheap-glitch fretboarder(1)

### DIFF
--- a/source/refined-github.css
+++ b/source/refined-github.css
@@ -22,8 +22,9 @@
 	padding: 0 !important;
 }
 
+/* Modify only the left and right padding (#3606) */
 #js-issues-toolbar .table-list-header-toggle summary {
-	padding: 0 8px !important;
+	padding: inherit 8px !important;
 }
 
 #js-issues-toolbar .table-list-header-toggle > :last-child > summary {
--- a/source/refined-github.css
+++ b/source/refined-github.css
@@ -22,8 +22,11 @@
 	padding: 0 !important;
 }
 
-/* Modify only the left and right padding #3606 */
-#js-issues-toolbar .table-list-header-toggle summary {
+#js-issues-toolbar .table-list-filters .table-list-header-toggle summary {
+	padding: 0 8px !important;
+}
+
+#js-issues-toolbar .table-list-triage .table-list-header-toggle summary {
 	padding: inherit 8px !important;
 }
 
--- a/source/refined-github.css
+++ b/source/refined-github.css
@@ -22,7 +22,7 @@
 	padding: 0 !important;
 }
 
-/* Modify only the left and right padding (#3606) */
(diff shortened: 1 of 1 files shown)


source/refined-github.css
```diff
@@ -22,8 +22,9 @@
 	padding: 0 !important;
 }
 
+/* Modify only the left and right padding (#3606) */
```
yakov116, line 25:
```suggestion
/* Modify only the left and right padding #3606 */
```

👍 Thanks for putting a comment!

fregante:
Regular issues list (https://github.com/sindresorhus/refined-github/issues) on Firefox:

fregante:
> It should just change `padding-left` and `padding-right`

No comment necessary

cheap-glitch:
Fixed the wrong padding on the issues toolbar by creating two separate and more specific selectors. Although this might break something else if the previous selector was targeting another toolbar somewhere.

From what I can gather from the git blame, it was already made more specific to target only the issues toolbar, so it should be fine.

fregante:
If you ignore reviews, you're wasting everyone's time.

cheap-glitch:
I'm very sorry about that, I completely misunderstood one of you comments. Fixed the PR if you still want to use it.
Again, sorry for wasting your time with dumb mistakes 🙇

PR #509, opened 2017-06-19
Move Compare into More dropdown

bfred-it:
Fixes #440

Closes #501

screen shot 2017-06-19 at 11 53 37

Issue #440, opened 2017-05-25
Repo tabs doesn't fit when Wiki and Projects tabs are enabled


sindresorhus:
<img width="939" alt="screen shot 2017-05-25 at 14 43 49" src="https://cloud.githubusercontent.com/assets/170270/26440906/9512c24a-4158-11e7-9b3a-db1e9bbf7156.png">

Maybe we should renamed `Insights` to `More` and move `Settings` there too?

hkdobrev:
To clarify they don't fit when you have Issues, Projects and Wiki enabled and Settings available.

hkdobrev:
I'd prefer to remove the Compare tab in these cases. If we do the More drop down suggestion, I'd move Compare in there instead of Settings.

sindresorhus:
How often do you really need Settings though?

hkdobrev:
Well, I use the tab 2-3 times per week on various repositories. However, I am yet to use the Compare tab.

To clarify we'd always have the Compare tab except for cases when you have admin access to a repo with a project.

hkdobrev:
We can consider reducing spacing or using just icons with a shortcut letter just in this case.

sindresorhus:
Ok, let's put Compare in the More dropdown if it doesn't fit.

hkdobrev:
Steps to reproduce for anyone who want to pick this up:
1. Create a repo
2. Go to settings
3. Create a project

bfred-it:
Possible alternatives:

- Shorten Pull Requests to PRs
- Use the "Settings" icon only, no text
- Use the "Graphs" icon for "Insights", no text

PR #3616, opened 2020-10-03
Remove Security and Insights from more-dropdown 

galagora:
"Hide", "Insights", and "Releases" were the first things I searched for in the options page.

### DIFF
--- a/source/features/more-dropdown.tsx
+++ b/source/features/more-dropdown.tsx
@@ -75,9 +75,6 @@ async function init(): Promise<void> {
 			createDropdownItem('Commits', commitsUrl),
 			createDropdownItem('Branches', `/${repoUrl}/branches`)
 		);
-
-		onlyShowInDropdown('security-tab');
-		onlyShowInDropdown('insights-tab');
 		return;
 	}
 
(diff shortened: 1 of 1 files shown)


busches:
This now lists all the things we add/move to the dropdown, but we don't move everything, compare for example is added.

fregante:
It hides 2 tabs, but it still **adds** 4 links

galagora:
Wow, I completely glossed over that. The original issue that prompted moving them under the dropdown, #2072, is no longer a problem. So I think it's best to just restore the location of the two tabs.

fregante:
If you don't like this feature as is, it can be disabled.

This feature is much older than #2072. Insights used to be its own dropdown with Pulse, Community, and Graphs inside it as seen here: https://github.com/sindresorhus/refined-github/issues/440

In https://github.com/sindresorhus/refined-github/pull/509 we moved one of our tabs inside it and renamed it to "More". The reason for this feature to still exist is that one of Refined GitHub’s intention is to reduce clutter. I don't think many use the Security tab daily.

galagora:
Well, I, for one, use the Insights tab quite often. Nevertheless, the extra options in the dropdown seem useful; I guess I'll just have to write a userscript.

PR #3620, opened 2020-10-05
Parcel test (needs extra work)

fregante:
Follows #2983 and #3618 

Based on #3227 (most of the _changed files_ can be ignored)

Parcel is 3 times slower than Webpack.
Fully-minified filesizes are comparable.

Given the speed drop, I'm not sure if it's worth spending time on the remaining issues.

## Before

```js
❯ nr build:webpack
Time: 4544ms
                   Asset      Size
       refined-github.js   291 KiB
      refined-github.css  40.7 KiB
           background.js  33.7 KiB
            options.html  2.19 KiB
             options.css   3.5 KiB
              options.js  99.2 KiB
    resolve-conflicts.js  2.43 KiB
                icon.png  1.89 KiB
           manifest.json   1.1 KiB
 browser-polyfill.min.js  9.87 KiB
```

## After

```js
❯ nr build:parcel
✨ Built in 15.78s

distribution/refined-github.js         266.81 KB    6.85s
distribution/refined-github.css          25.1 KB    4.65s
distribution/background.js              34.84 KB    6.57s
distribution/options.html                   2 KB    6.85s
distribution/options.4fc4ce54.css        2.41 KB    2.70s
distribution/options.9f319f47.js        55.07 KB    2.86s
distribution/resolve-conflicts.js        1.53 KB    2.86s
distribution/icon.1f68ed73.png           1.89 KB     20ms
distribution/manifest.0c1d4e4f.json      1.04 KB     37ms
```

## Next steps

- [ ] Port the [dynamic variable replacer](https://github.com/sindresorhus/refined-github/blob/0e50ebcf0632680d2e654a719214280a5aa3bc29/webpack.config.ts#L94-L106) to Parcel or Babel
- [ ] Filenames of PNG and JSON files need to be preserved (they're currently hashed)
- [ ] Possibly use the Parcel-native TypeScript type-checker (called Validator)
- [ ] Find ways to speed up?


fregante:
Perhaps we should just wait for ES Modules support in Firefox (https://bugzilla.mozilla.org/show_bug.cgi?id=1451545) so we can just use Snowpack.

notlmn:
Options we have right now:

- I hate `webpack` as much as I love it (simple to use, but doesn't have first-class HTML entrypoints)
- `parcel` being parcel is bloated as always and hence the 1/3rd performance compared to webpack
- `rollup` is awesome, but requires us to write custom config the size of a linux kernel
- `snowpack` sounds promising, would be worth a try if it lives up to its reputation

The reason I would like to keep sticking to webpack (for now) is the ease of use, __it just works__ (most of the time though).

Also relevant to the discussion at https://github.com/notlmn/browser-extension-template/issues/34

---------

> ## Next steps
> - Port the dynamic variable replacer to Parcel or Babel
> - Filenames of PNG and JSON files need to be preserved (they're currently hashed)
> - Possibly use the Parcel-native TypeScript type-checker (called Validator)
> - Find ways to speed up?

These do require writing a config file anyway (I assume) which needs to be maintained, so we're right back to the question if we want to keep maintaining the `webpack` config we have.

fregante:
> if we want to keep maintaining the `webpack` config we have.

That's what I thought. For Refined GitHub, changing builder means porting/rewriting the feature list injection so it takes effort that the change of builder is supposed to prevent.

I had also tried Rollup (https://github.com/sindresorhus/refined-github/pull/3019) but, yeah, it needed just as much configuration and is as slow as Parcel (https://github.com/rollup/rollup/issues/3506). The only advantage is better code output.

https://github.com/sindresorhus/refined-github/pull/3227 itself will drop the need for an SVG loader, so the config will be reduced slightly.

---

However, if you remove SVG import (which is supported via SVGR in Parcel 2 and [supposedly supported](https://github.com/sindresorhus/refined-github/pull/3618/files) in Parcel 2) and feature list injection (which is a very custom solution anyway), Parcel 1 would be good enough for most extensions thanks to [parcel-plugin-web-extension](https://github.com/kevincharm/parcel-plugin-web-extension), as described in https://github.com/notlmn/browser-extension-template/issues/34

PR #3623, opened 2020-10-07
Don't show next scheduled time of disabled workflows

cheap-glitch:
Thanks for contributing! 🍄

1. LINKED ISSUES: None (PR #3604)

2. TEST URLS: https://github.com/sindresorhus/refined-github/actions

3. SCREENSHOT:

Before | After
--- | ---
image | image

This is a pretty flaky fix since it depends on the SVG icon next to the workflow name, but I couldn't find another way to do this.

### DIFF
--- a/source/features/next-scheduled-github-action.tsx
+++ b/source/features/next-scheduled-github-action.tsx
@@ -54,7 +54,7 @@ async function init(): Promise<false | void> {
 	}
 
 	for (const workflowListItem of select.all('[href*="?query"]', await elementReady('.hx_actions-sidebar'))) {
-		if (select.exists('.octicon-stop')) {
+		if (select.exists('.octicon-stop', workflowListItem)) {
 			continue;
 		}
 
--- a/source/features/next-scheduled-github-action.tsx
+++ b/source/features/next-scheduled-github-action.tsx
@@ -54,6 +54,10 @@ async function init(): Promise<false | void> {
 	}
 
 	for (const workflowListItem of select.all('[href*="?query"]', await elementReady('.hx_actions-sidebar'))) {
+		if (select('svg', workflowListItem)?.classList.contains('octicon-stop')) {
+			continue;
+		}
+
 		const workflowName = workflowListItem.textContent!.trim();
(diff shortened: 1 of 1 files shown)


source/features/next-scheduled-github-action.tsx
```diff
@@ -54,6 +54,10 @@ async function init(): Promise<false | void> {
 	}
 
 	for (const workflowListItem of select.all('[href*="?query"]', await elementReady('.hx_actions-sidebar'))) {
+		if (select('svg', workflowListItem)?.classList.contains('octicon-stop')) {
```
yakov116, line 57: I am not in front of a computer not but would a css idea like this one work here?

https://github.com/sindresorhus/refined-github/blob/6651e05638b1b6609048dcc8799a043448834eb0/source/features/bypass-checks.tsx#L20-L24

cheap-glitch: We could loop directly over the SVGs in the workflows list (excluding the ones with `octicon-stop` class), use `nextSibling()` instead of `textContent` and `insertAdjacentElement('afterend', x)` instead of `append()`.

Thoughts @fregante ?

yakov116: Can you look to see if there is any css solution?

cheap-glitch: By "css solution", do you mean hide the inserted info using a selector like `.hx_actions-sidebar svg:not(.octicon-stop) ~ em`?

If we're going to rely on the SVGs anyway, wouldn't it be better to ignore the disabled workflows directly in the loop instead of computing the info and then hiding it?

fregante: He means in a single CSS selector. Use `select.exists('.octicon-stop', item)`

fregante: You’re missing part of the code

cheap-glitch: Sorry about that, fixed it.

fregante (approved):

PR #3624, opened 2020-10-07
Add an auto assign button for reviewers

rockwotj:
The main issue here is that github has both a reviewers section and an assignees section, which are very similar. Some people prefer using one or the other. The flow for reviewing is nice, as it automatically toggles when you leave a review and then you can be re-requested.

However, then updating the assignees field based on that is super clunky. This add a button to update the assignees in a single click. If you have reviews requested then those reviewers are assigned - everyone else is removed. Otherwise if there are no reviewers, then the pull request author is assigned.

1. LINKED ISSUES:
   None.

2. TEST URLS:
   https://github.com/sindresorhus/refined-github/pull/3407
   https://github.com/sindresorhus/refined-github/pull/3623

3. SCREENSHOT:
   
Screen Shot 2020-10-07 at 9 41 18 AM

auto-assign

### DIFF
--- a/source/features/assign-reviewer-shortcut.css
+++ b/source/features/assign-reviewer-shortcut.css
@@ -1,0 +1,3 @@
+.rgh-auto-assign-link {
+	cursor: pointer;
+}
--- a/source/features/assign-reviewer-shortcut.tsx
+++ b/source/features/assign-reviewer-shortcut.tsx
@@ -72,6 +72,11 @@ async function addSidebarAutoAssignButton(): Promise<void | false> {
 	if (select.exists(".rgh-auto-assign-link", sidebarReviewsSection)) {
 		return false;
 	}
+        // Only add the button if you have the ability to request reviews.
+	if (!select.exists("#reviewers-select-menu", sidebarReviewsSection)) {
+		return false;
+	}
+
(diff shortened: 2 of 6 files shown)


fregante:
It’s best to discuss features before sending a PR. How are reviewers selected and how are these different from the ones that GitHub already suggests?

rockwotj:
@fregante sorry you're too fast for me :)

Basically there are lots of people that just use the assignee field to see if they need to take action on a PR. I've updated the PR description here to help.



> It’s best to discuss features before sending a PR. How are reviewers selected and how are these different from the ones that GitHub already suggests?

I understand - and I know this project is opinionated about features, so no hard feelings if you feel like this isn't a good fit. I'm going to use my fork with this feature (as well as the right of my company) either way, hence me just sending a PR instead of opening an issue first - hope that's OK!

fregante:
I think this is unnecessary as it duplicates an existing feature: request review. You prefer Assignees, but I don’t see why one can’t just use Review Requests (since it even used GitHub’s own suggested reviewers to be assigned, from what I understand)

rockwotj:
This doesn't use githubs suggestions, it's more about making them the same when you have co-workers with different styles (some use reviewers and some use assignees).

No worries - I'll close this. Thanks for your time!

PR #3625, opened 2020-10-07
Extend `linkify-labels-on-dashboard` to newsfeed items

yakov116:
1. LINKED ISSUES:
   None

2. TEST URLS:
   https://github.com/

3. SCREENSHOT:
  None

@fregante I added help-wanted so you should see this in your feed


source/features/linkify-labels-on-dashboard.tsx
```diff
@@ -9,11 +9,12 @@ import features from '.';
 
 function init(): void {
 	const labelClass = [
-		'.js-recent-activity-container :not(a) > .IssueLabel', // Recent activity
-		'.js-all-activity-header + div :not(a) > .IssueLabel' // Newsfeed
+		'.js-recent-activity-container :not(a) > .IssueLabel:not(.rgh-linkfied-label)', // Recent activity
+		'.js-all-activity-header + div :not(a) > .IssueLabel:not(.rgh-linkfied-label)' // Newsfeed
```
yakov116, line 13: is this needed since we do `:not(a)`

source/features/linkify-labels-on-dashboard.tsx
```diff
@@ -9,11 +9,12 @@ import features from '.';
...
 	].join();
 	observe(labelClass, {
 		add(label) {
+			label.classList.add('rgh-linkfied-label');
```
yakov116, line 17: is this needed since we do :not(a)

fregante: That's right. You can drop that and leave a comment so we avoid having this conversation again.

source/features/linkify-labels-on-dashboard.tsx
```diff
@@ -8,9 +8,14 @@ import {wrap} from '../helpers/dom-utils';
...
-	observe('.js-recent-activity-container :not(a) > .IssueLabel', {
+	const labelClass = [
+		'.js-recent-activity-container :not(a) > .IssueLabel:not(.rgh-linkfied-label)', // Recent activity
+		'.js-all-activity-header + div :not(a) > .IssueLabel:not(.rgh-linkfied-label)' // Newsfeed
+	].join();
+	observe(labelClass, {
 		add(label) {
-			const activity = label.closest('li')!;
+			label.classList.add('rgh-linkfied-label');
+			const activity = label.closest('li')! ?? label.parentElement; // Newsfeed
```
fregante, line 18: Can this be merged? Random example:
```suggestion
			const activity = label.closest('li, div')!;
```

Also, never use `! ??`, because `??` already handles and null value.

yakov116: Need to test it (give me a few min working on anther PR) but I _think_ there is a `div` there that will get caught.

yakov116: > Can this be merged

No. There is a `div` in the non-newsfeed there that will get caught.

fregante: That was _random example_ to ask you to find a common selector

yakov116: [`63d9957` (#3625)](https://github.com/sindresorhus/refined-github/pull/3625/commits/63d99571f38f527cf05cbf04b5c652684280d9d1) How's this?

source/features/linkify-labels-on-dashboard.tsx
```diff
@@ -8,9 +8,14 @@ import {wrap} from '../helpers/dom-utils';
...
 function init(): void {
-	observe('.js-recent-activity-container :not(a) > .IssueLabel', {
+	const labelClass = [
+		'.js-recent-activity-container :not(a) > .IssueLabel', // Recent activity
+		'.js-all-activity-header + div :not(a) > .IssueLabel' // Newsfeed
+	].join();
+	observe(labelClass, {
 		add(label) {
-			const activity = label.closest('li')!;
+			// A rgh class is not needed since we already check for `not(a)` https://github.com/sindresorhus/refined-github/pull/3625#discussion_r501309853
```
fregante, line 17:
```suggestion
			// A rgh class is not needed since we already check for `not(a)` #3625
```

source/features/linkify-labels-on-dashboard.tsx
```diff
@@ -8,9 +8,14 @@ import {wrap} from '../helpers/dom-utils';
...
+			const activity = label.closest('li, div:not([class])')!;
```
fregante, line 18: It's best to explain what it's for since Newsfeed events are much more rare.

```suggestion
			const activity = label.closest('li, div:not([class])')!; // `div` is for the Newsfeed
```

yakov116: 👍 Forgot to add the comment back

fregante (approved):

PR #3627, opened 2020-10-07
Use system setting for `user-local-time` time formatting

### LINKED ISSUE #3597: Ability to preview `user-local-time` in 12h format
I do not see any option in the extensions settings to change the `user-local-time` format to 12h, it would be great if it were possible to change it.
image

yakov116:
1. LINKED ISSUES:
   Resolves #3597 

2. TEST URLS:
   My Avatar

3. SCREENSHOT:
image

### DIFF
--- a/source/features/user-local-time.tsx
+++ b/source/features/user-local-time.tsx
@@ -17,10 +17,9 @@ interface Commit {
 	sha: string;
 }
 
-const timeFormatter = new Intl.DateTimeFormat('en-GB', {
+const timeFormatter = new Intl.DateTimeFormat(undefined, {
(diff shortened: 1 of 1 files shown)


sindresorhus (approved):
I disagree with this in principle, as 12-hour time is dumb and ambiguous. But I also recognize RG is not the place to fix world problems.

fregante:
I also argue for 24h, but I saw that GitHub uses 12h on my en-US computer, so it’d kinda make sense to follow the same format

fregante (approved):
Begrudgingly

yakov116:
Title is good?

PR #3629, opened 2020-10-07
Use API for `bypass-checks`

yakov116:
1. LINKED ISSUES:
   Dont remember which issue

2. TEST URLS:
   https://github.com/xojs/xo/pull/497
https://github.com/sindresorhus/got/pull/1491
https://github.com/webpack/webpack/pull/11438

3. SCREENSHOT:
   None

### DIFF
--- a/source/features/bypass-checks.tsx
+++ b/source/features/bypass-checks.tsx
@@ -1,19 +1,20 @@
+import mem from 'mem';
 import onetime from 'onetime';
 import {observe} from 'selector-observer';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
-import fetchDom from '../helpers/fetch-dom';
+import * as api from '../github-helpers/api';
+import {getRepositoryInfo} from '../github-helpers';
+
+const getDirectLink = mem(async (runNumber: number): Promise<string> => {
+	const directLink = await api.v3(`https://api.github.com/repos/${getRepositoryInfo()!.owner!}/${getRepositoryInfo()!.name!}/check-runs/${runNumber}`);
+	return directLink.details_url;
+});
 
 async function bypass(detailsLink: HTMLAnchorElement): Promise<void> {
-	const directLink = await fetchDom<HTMLAnchorElement>(
-		detailsLink.href,
-		'[data-hydro-click*="check_suite.external_click"]'
-	);
-
-	if (directLink) {
-		detailsLink.href = directLink.href;
-	}
+	const runNumber = detailsLink.href.split(/\/|=/).pop();
+	detailsLink.href = await getDirectLink(Number(runNumber));
 }
 
 function init(): void {
--- a/source/features/bypass-checks.tsx
+++ b/source/features/bypass-checks.tsx
@@ -7,8 +7,8 @@ import features from '.';
 import * as api from '../github-helpers/api';
 import {getRepoURL} from '../github-helpers';
 
-const getDirectLink = mem(async (runNumber: number): Promise<string> => {
-	const directLink = await api.v3(`repos/${getRepoURL()}/check-runs/${runNumber}`);
+const getDirectLink = mem(async (runId: number): Promise<string> => {
+	const directLink = await api.v3(`repos/${getRepoURL()}/check-runs/${runId}`);
 	return directLink.details_url;
 });
 
--- a/source/features/bypass-checks.tsx
+++ b/source/features/bypass-checks.tsx
@@ -5,10 +5,10 @@ import * as pageDetect from 'github-url-detection';
 
 import features from '.';
 import * as api from '../github-helpers/api';
-import {getRepositoryInfo} from '../github-helpers';
+import {getRepoURL} from '../github-helpers';
 
 const getDirectLink = mem(async (runNumber: number): Promise<string> => {
-	const directLink = await api.v3(`https://api.github.com/repos/${getRepositoryInfo()!.owner!}/${getRepositoryInfo()!.name!}/check-runs/${runNumber}`);
+	const directLink = await api.v3(`https://api.github.com/repos/${getRepoURL()}/check-runs/${runNumber}`);
 	return directLink.details_url;
 });
 
--- a/source/features/bypass-checks.tsx
+++ b/source/features/bypass-checks.tsx
@@ -14,7 +14,7 @@ const getDirectLink = mem(async (runNumber: number): Promise<string> => {
 
 async function bypass(detailsLink: HTMLAnchorElement): Promise<void> {
 	const runId = new URLSearchParams(detailsLink).get('check_run_id') ?? detailsLink.pathname.split('/').pop();
-	detailsLink.href = await getDirectLink(Number(runNumber));
+	detailsLink.href = await getDirectLink(Number(runId));
 }
 
 function init(): void {
--- a/source/features/bypass-checks.tsx
+++ b/source/features/bypass-checks.tsx
@@ -8,7 +8,7 @@ import * as api from '../github-helpers/api';
 import {getRepoURL} from '../github-helpers';
 
 const getDirectLink = mem(async (runNumber: number): Promise<string> => {
-	const directLink = await api.v3(`https://api.github.com/repos/${getRepoURL()}/check-runs/${runNumber}`);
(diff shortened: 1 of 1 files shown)


fregante:
🔥

source/features/bypass-checks.tsx
```diff
@@ -1,19 +1,20 @@
...
+import * as api from '../github-helpers/api';
+import {getRepositoryInfo} from '../github-helpers';
 
-async function bypass(detailsLink: HTMLAnchorElement): Promise<void> {
-	const directLink = await fetchDom<HTMLAnchorElement>(
-		detailsLink.href,
-		'[data-hydro-click*="check_suite.external_click"]'
-	);
+const getDirectLink = mem(async (runNumber: number): Promise<string> => {
+	const directLink = await api.v3(`https://api.github.com/repos/${getRepositoryInfo()!.owner!}/${getRepositoryInfo()!.name!}/check-runs/${runNumber}`);
```
fregante, line 11: I think it's worth using v4 so that we can also limit the amount of requests: https://developer.github.com/v4/object/checkrun/#detailsurl

fregante: I know however that this complicates it considerably with `selector-observer` since you'd have to build a list in `add`, then in the next tick, make the joint v4 query.

yakov116: Unless you can figure out how to use that thing, I gave up!

It was a nest of 7 and was only per commit. In other words I could not search for a specific run I was only able to load a commit and then load all its _runs_.

V3 we can search for a specific run

fregante: @sindresorhus if you respond in time: what's a good pattern for this? It's like the `event loop` but instead of callbacks it's a list of elements

1. Add to array in loop
2. Wait for the next tick (likely via `await Promise.resolve()`?)
3. Use the array

source/features/bypass-checks.tsx
```diff
@@ -1,19 +1,20 @@
...
+const getDirectLink = mem(async (runNumber: number): Promise<string> => {
+	const directLink = await api.v3(`repos/${getRepoURL()}/check-runs/${runNumber}`);
+	return directLink.details_url;
+});
 
-	if (directLink) {
-		detailsLink.href = directLink.href;
-	}
+async function bypass(detailsLink: HTMLAnchorElement): Promise<void> {
+	const runNumber = detailsLink.href.split(/\/|=/).pop();
```
fregante, line 16:
```suggestion
	const runNumber = detailsLink.href.split(/[/=]/).pop();
```

fregante: I'm only seeing links with `/` however, where did you see the `=`?

yakov116: Hovercards

fregante: I don't see it there, but in the PR Mergeability box the URL is toward a PR page:

```
https://github.com/sindresorhus/refined-github/pull/3629/checks?check_run_id=1223857819
```

fregante: Let's make this explicit:

```suggestion
	const runId = new URLSearchParams(detailsLink).get('check_run_id') ?? detailsLink.pathname.split('/').pop();
```

yakov116: Sorry I mixed up what was what. Explicit is much better!

source/features/bypass-checks.tsx
```diff
@@ -1,19 +1,20 @@
...
+	const directLink = await api.v3(`repos/${getRepoURL()}/check-runs/${runNumber}`);
+	return directLink.details_url;
+});
 
-	if (directLink) {
-		detailsLink.href = directLink.href;
-	}
+async function bypass(detailsLink: HTMLAnchorElement): Promise<void> {
+	const runId = new URLSearchParams(detailsLink).get('check_run_id') ?? detailsLink.pathname.split('/').pop();
+	detailsLink.href = await getDirectLink(Number(runNumber));
```
yakov116, line 17:
```suggestion
	detailsLink.href = await getDirectLink(Number(runId));
```

source/features/bypass-checks.tsx
```diff
@@ -1,19 +1,20 @@
...
-import fetchDom from '../helpers/fetch-dom';
+import * as api from '../github-helpers/api';
+import {getRepoURL} from '../github-helpers';
 
-async function bypass(detailsLink: HTMLAnchorElement): Promise<void> {
-	const directLink = await fetchDom<HTMLAnchorElement>(
-		detailsLink.href,
-		'[data-hydro-click*="check_suite.external_click"]'
-	);
+const getDirectLink = mem(async (runNumber: number): Promise<string> => {
```
yakov116, line 10:
```suggestion
const getDirectLink = mem(async (runId: number): Promise<string> => {
```

source/features/bypass-checks.tsx
```diff
@@ -1,19 +1,20 @@
...
+	const directLink = await api.v3(`repos/${getRepoURL()}/check-runs/${runNumber}`);
```
yakov116, line 11:
```suggestion
	const directLink = await api.v3(`repos/${getRepoURL()}/check-runs/${runId}`);
```

yakov116:
Will fix the types when I get to a computer

source/features/bypass-checks.tsx
```diff
@@ -1,19 +1,20 @@
...
-import fetchDom from '../helpers/fetch-dom';
+import * as api from '../github-helpers/api';
+import {getRepoURL} from '../github-helpers';
 
-async function bypass(detailsLink: HTMLAnchorElement): Promise<void> {
-	const directLink = await fetchDom<HTMLAnchorElement>(
-		detailsLink.href,
-		'[data-hydro-click*="check_suite.external_click"]'
-	);
+const getDirectLink = mem(async (runId: number): Promise<string> => {
```
fregante, line 10: The API is already memoize, no need to do it again

source/features/bypass-checks.tsx
```diff
@@ -1,19 +1,20 @@
...
+	const directLink = await api.v3(`repos/${getRepoURL()}/check-runs/${runId}`);
+	return directLink.details_url;
+});
 
-	if (directLink) {
-		detailsLink.href = directLink.href;
-	}
+async function bypass(detailsLink: HTMLAnchorElement): Promise<void> {
+	const runId = new URLSearchParams(detailsLink.href).get('check_run_id') ?? detailsLink.pathname.split('/').pop();
```
fregante, line 16:
```suggestion
	const runId = new URLSearchParams(detailsLink.search).get('check_run_id') ?? detailsLink.pathname.split('/').pop();
```

source/features/bypass-checks.tsx
yakov116: https://github.com/sindresorhus/refined-github/blob/022e418940b98f7dfa7f4e2c532b2b83782ca45a/source/features/deep-reblame.tsx#L16

I keep doing it since I remember from here

fregante: It can be removed from there too. It would only make sense if it saved heavy operations (e.g. HTTP requests), but what follows is just a few conditions.

source/features/bypass-checks.tsx
fregante: I had missed your last comment. Thanks for trying. It sucks that the GraphQL API is so limited compared to REST.

source/features/bypass-checks.tsx
```diff
@@ -3,17 +3,13 @@ import {observe} from 'selector-observer';
...
-	const directLink = await fetchDom<HTMLAnchorElement>(
-		detailsLink.href,
-		'[data-hydro-click*="check_suite.external_click"]'
-	);
-
-	if (directLink) {
-		detailsLink.href = directLink.href;
-	}
+	const runId = new URLSearchParams(detailsLink.search).get('check_run_id') ?? detailsLink.pathname.split('/').pop();
+	const directLink = await api.v3(`repos/${getRepoURL()}/check-runs/${Number(runId)}`);
```
fregante, line 11: Why number? It goes into a string.

yakov116: Sorry I have no idea why I did that!

fregante (approved):

PR #3640, opened 2020-10-12
Restore `embed-gist-inline` feature

### LINKED ISSUE #2022: `embed-gist-inline` fails
Gists fail to load due to CSP

# Live example

https://gist.github.com/sompylasar/99b5d307da3168b833c1119fb95caf11

# Screenshot

yakov116:
1. LINKED ISSUES:
   Fixes #2022 

2. TEST URLS:

https://gist.github.com/sompylasar/99b5d307da3168b833c1119fb95caf11

3. SCREENSHOT:
   None


source/background.ts
```diff
@@ -42,3 +42,16 @@ browser.runtime.onInstalled.addListener(async ({reason}) => {
...
+
+// `background` fetch required to avoid avoid CORB introduced in Chrome 73 https://chromestatus.com/feature/5629709824032768
+browser.runtime.onMessage.addListener(({action, payload}, _sender, sendResponse): void | boolean => {
+	if (action === 'fetch') {
+		void fetch(payload)
+			.then(response => sendResponse(response.json()))
+			.catch(error => ({
+				error: error.message // Make error JSON.stringify-able
+			}))
+			.then(sendResponse);
```
fregante, line 54: Thanks to the promised `browser.*` API, you can just return the promise instead of all of this. Drop as much code as possible. 

Example: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/onMessage

yakov116:
Test

https://gist.github.com/sompylasar/99b5d307da3168b833c1119fb95caf11

source/background.ts
```diff
@@ -42,3 +42,10 @@ browser.runtime.onInstalled.addListener(async ({reason}) => {
 
 // GitHub Enterprise support
 addDomainPermissionToggle();
+
+// `background` fetch required to avoid avoid CORB introduced in Chrome 73 https://chromestatus.com/feature/5629709824032768
+browser.runtime.onMessage.addListener(({action, payload}): Promise<Response> | void => {
+	if (action === 'fetch') {
+		return Promise.resolve(fetch(payload).then(async response => response.json()));
```
fregante, line 49: This is already a promise. Also generally avoid `.then`
```suggestion
		const response = await fetch(payload);
		return response.json();
```

source/background.ts
```diff
@@ -42,3 +42,10 @@ browser.runtime.onInstalled.addListener(async ({reason}) => {
 
 // GitHub Enterprise support
 addDomainPermissionToggle();
+
+// `background` fetch required to avoid avoid CORB introduced in Chrome 73 https://chromestatus.com/feature/5629709824032768
+browser.runtime.onMessage.addListener(({action, payload}): Promise<Response> | void => {
```
fregante, line 47: Does this cause trouble in the polyfill?
```suggestion
browser.runtime.onMessage.addListener(async ({action, payload}): Promise<Response | void> => {
```

source/background.ts
fregante: Could also test what happens in case of errors in Chrome? Are they correctly passed back to the content script? Code to test:

```js
const response = await fetch('http://doesn.t.exist');
```

and 

```js
const response = await fetch('https://google.com');
```

which exists, but should fail in `.json()`

source/background.ts
fregante: I just read the docs. Let's add this as a comment
```suggestion
// Don’t turn this into an `async` function https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/onMessage#addListener_syntax
browser.runtime.onMessage.addListener(({action, payload}): Promise<Response> | void => {
```

yakov116:
One sec I am getting lost in the comments

notlmn:
1. indent
	https://gist.github.com/sompylasar/99b5d307da3168b833c1119fb95caf11

source/background.ts
fregante: This

yakov116: I cant do `await` without `async` and that we cant do as we said below

fregante: Darn it. Alright 😃

yakov116: 5 Points for trying 😉

source/features/embed-gist-inline.tsx
```diff
@@ -20,8 +19,11 @@ async function embedGist(link: HTMLAnchorElement): Promise<void> {
 	link.after(info);
 
 	try {
-		const response = await fetch(`${link.href}.json`);
-		const gistData = await response.json();
+		// Get the gist via background.js due to CORB policies introduced in Chrome 73
+		const gistData = await browser.runtime.sendMessage({
+			action: 'fetch',
+			payload: `${link.href}.json`
```
fregante, line 25: Can we shorten this object to just being `{fetch: 'blah blah json'}`. Splitting payload from the action seems unnecessary (even though that’s how I’ve done it for years)

yakov116: I tried doing this and typescript did not like me. Want me to commit?

yakov116: Also the offical **google** docs say to do a check first so that and sendmessage that is passed does not end up by mistake fetched

fregante: What kind of check?

What does TypeScript say? It’s the same type, except that instead of calling the property `payload` it calls it `fetch`. That’s it.

yakov116: The problem was that `fetch` is a reserved word.

kidonng: Then [`got`](https://github.com/sindresorhus/got)

fregante: > The problem was that `fetch` is a reserved word.

Ah the problem is in the destructuring part I suppose

fregante (approved):

fregante:
Code looks good. Hope you tested the latest version in both browsers

yakov116:
I did

PR #3639, opened 2020-10-11
Meta: Update dependencies and simplify build

fregante:
- Update Webpack to v5 (+ loaders)
- Replace `ts-loader` with [`esbuild-loader`](https://github.com/privatenumber/esbuild-loader) to save a couple of seconds of build time
- Keep static files to `./distribution` instead of using Webpack to copy them (reduces config)
- Drop old unused dependencies
- Lint `DefinePlugin` 

Overall, the config file is 46 lines shorter (-27%) :cake:

Next, maybe:

- [ ] Update dependencies again later to avoid some deprecation notices
- [x] Find out why it still takes 11 seconds to build. Edit: it appears to wait idly for a few seconds for no reason:

👆 made with https://webpack.js.org/plugins/profiling-plugin/

PR #3644, opened 2020-10-13
Don’t run `bypass-checks` on non "run" actions

yakov116:
1. LINKED ISSUES:
   None

2. TEST URLS:
	https://github.com/eslint/eslint/pull/13748
3. SCREENSHOT:
   None


yakov116:
I dont like what I wrote. maybe will gibve it a shot in the morning. @fregante I am open to idea's.

fregante:
> 2\. TEST URLS:
>     [eslint/eslint#13748](https://github.com/eslint/eslint/pull/13748)

Which one? I don't see any check that needs to be bypassed there.

source/features/bypass-checks.tsx
```diff
@@ -7,7 +7,13 @@ import * as api from '../github-helpers/api';
 import {getRepoURL} from '../github-helpers';
 
 async function bypass(detailsLink: HTMLAnchorElement): Promise<void> {
-	const runId = new URLSearchParams(detailsLink.search).get('check_run_id') ?? detailsLink.pathname.split('/').pop()!;
+	const [actionType, splitRunid] = detailsLink.pathname.split('/').splice(-2);
+	const runId = new URLSearchParams(detailsLink.search).get('check_run_id') ?? splitRunid;
+
+	if (!Number.parseInt(runId, 10) || actionType !== 'run') {
```
fregante, line 13: A bit verbose. Can pageDetect help?

yakov116: I am 100% at fault here. I wrote this half in my sleep, I should have waited for the morning...

I will update it later with more details

yakov116:
Let me explain the issue.
https://github.com/eslint/eslint/pull/13748 has a check for https://github.com/eslint/eslint-github-bot/blob/master/docs/commit-message-check.md and https://github.com/eslint/eslint/issues/13719 (hover over the first commit status to see the second one)

These 2 are not action runs and they already link to their direct page.

fregante:
Use GitHub URL Detection with an argument, no need for custom reflexes unless you’re *extracting* data. In this case `isActionRun(detailsLink)` should work, except that the types don’t accept a raw HTMLAnchorElement. PR welcome there to change that

fregante:
Also include the 2 types of URLs in comments

source/features/bypass-checks.tsx
```diff
@@ -7,7 +7,12 @@ import * as api from '../github-helpers/api';
 import {getRepoURL} from '../github-helpers';
 
 async function bypass(detailsLink: HTMLAnchorElement): Promise<void> {
-	const runId = new URLSearchParams(detailsLink.search).get('check_run_id') ?? detailsLink.pathname.split('/').pop()!;
+	const runId = pageDetect.isActionJobRun(detailsLink) ? 	detailsLink.pathname.split('/').pop() :// https://github.com/xojs/xo/runs/1104625522
```
fregante, line 10: Rough whitespace around here. There's a tab after `?`

This is the expected indentation for the ternary operator
```suggestion
	const runId = pageDetect.isActionJobRun(detailsLink) ?
		detailsLink.pathname.split('/').pop() : // https://github.com/xojs/xo/runs/1104625522
```

source/features/bypass-checks.tsx
```diff
@@ -7,7 +7,12 @@ import * as api from '../github-helpers/api';
...
+		new URLSearchParams(detailsLink.search).get('check_run_id'); // https://github.com/sindresorhus/refined-github/pull/3629/checks?check_run_id=1223857819
+	if (!runId) {
+		return;
```
fregante, line 13:
```suggestion
		// Sometimes the URL doesn't point to Checks at all
		return;
```

fregante (approved):

fregante:
You can drop the package files, I updated the dependency in #3639

PR #3649, opened 2020-10-14
Preserve capitalization of repository URL in links

### LINKED ISSUE #3648: Release links are lowercased
https://github.com/BtbN/FFmpeg-Builds has Releases link https://github.com/btbn/ffmpeg-builds/releases instead of https://github.com/BtbN/FFmpeg-Builds/releases
All other tabs respect letter case, `releases-tab` feature should respect it too. Sidebar Releases block which is removed by this feature also respects letter case.

fregante:
Fixes #3648


source/features/go-to-action-from-file.tsx
```diff
@@ -15,7 +15,7 @@ function init(): void {
 		.replace(/["']/g, '')
 		.trim();
 
-	const actionURL = new URL(`${location.origin}/${getRepoURL()}/actions`);
+	const actionURL = new URL(`/${getRepoURL(true)}/actions`);
```
fregante, line 18: Actually this could be an opportunity make it even more useful:
```suggestion
	const actionURL = new URL(getRepoURL('actions'));
```

- `getRepoURL()` without argument would be used for API calls, selectors and cache keys
- `getRepoURL(path)` would be use to generate URLs to user pages.

It could go one step further and accept an extra argument, these would be equivalent
```js
	const actionURL = new URL(getRepoURL('actions'));
	const actionURL = getRepoURL('actions', URL);
```

fregante: Done. URL creation is now less noisy too.

```diff
- const url = `/${getRepoURL()}/actions`
+ const url = getRepoURL('actions')
```

fregante:
Last thought: should this be its own function?

```ts
function getRepoURL(): string;
function buildRepoURL(...parts: string[]): string;
```

yakov116:
At first I thought no, but then I saw your last commit. So :+1:

Issue #3599, opened 2020-09-28
Explore using `@octokit/rest`


fregante:
GitHub offers their own API module: https://github.com/octokit/rest.js

I've always shied away from using it due to its [size](https://bundlephobia.com/result?p=@octokit/rest@18.0.6), but it's worth seeing what our code would look like if we replaced our [api.ts](https://github.com/sindresorhus/refined-github/blob/master/source/github-helpers/api.ts) with the 1st-party module.

Would there be advantages? 
Would there be disadvantages other than weight?
Would it still need a lot of boilerplate?

PR welcome to convert one or two `api.v3` features to `octokit/rest.js` to evaluate it.

There's also https://github.com/octokit/graphql.js to consider and it would probably be more advantageous since they offer proper [GraphQL variables](https://graphql.org/learn/queries/#variables)

PR #3650, opened 2020-10-14
Use `@octokit/rest`

dertieran:
This PR converts some `api.v3` features to `@octokit/rest` as mentioned in #3599.

The main thing was to get octokit working with an async `token`.
For that I implemented something similar to [`@octokit/auth-token`](https://github.com/octokit/auth-token.js).
_(The main challenge for me was typescript here :sweat_smile:)_

Regarding size of the built `refined-github.js`
Before: `98.9 kB`
After: `113 kB`

The [documentation](https://octokit.github.io/rest.js) is okay and always links to the actual rest api.
Found it hard to find the right function first, but then I saw that the search on the top is actually quite powerful, that made it way easier.

### DIFF
--- a/source/features/bypass-checks.tsx
+++ b/source/features/bypass-checks.tsx
@@ -3,8 +3,8 @@ import {observe} from 'selector-observer';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
-import * as api from '../github-helpers/api';
-import {getRepoURL} from '../github-helpers';
+import octokit from '../github-helpers/octokit';
+import {getRepositoryInfo} from '../github-helpers';
 
 async function bypass(detailsLink: HTMLAnchorElement): Promise<void> {
 	const runId = pageDetect.isActionJobRun(detailsLink) ?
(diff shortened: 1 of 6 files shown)


source/features/bypass-checks.tsx
```diff
@@ -15,8 +15,14 @@ async function bypass(detailsLink: HTMLAnchorElement): Promise<void> {
 		return;
 	}
 
-	const directLink = await api.v3(`repos/${getRepoURL()}/check-runs/${runId}`);
-	detailsLink.href = directLink.details_url;
+	const {owner, name: repo} = getRepositoryInfo();
+	const check_run_id = Number.parseInt(runId, 10);
+	if (!owner || !repo || !check_run_id) {
```
yakov116, line 20: Why the check?

yakov116 (changes requested):
🎉

source/features/bypass-checks.tsx
dertieran: The `check_run_id` one?
Just for good measure if for some reason the `runId` could not be parsed into an int.

I guess it would also be possible to pass the `runId` directly as a `Number`.

_(I'm also not an expert in Typescript, so there might be better ways in general :smile:)_

fregante:
Thank you for trying this out!

So it’s a 15% increase (plus possibly another 15% for the GraphQL one). Do you reckon this brings something useful? So far it just appears to lengthen the code

dertieran:
I guess the main benefit here would be to have type support for all request/responses.
Otherwise I don't directly see a benefit, maybe that it is a bit more declarative.
But then again it also requires more boilerplate to achieve the same thing.

Just FYI `@octokit/graphql` might be [smaller](https://bundlephobia.com/result?p=@octokit/graphql@4.5.6) because it doesn't define all possible routes.

But I just saw that it is also included [`@octokit/core`](https://github.com/octokit/core.js), so it should be possible to just do.
```js
octokit.graphql('query { viewer { login }}');
```
_(Not tested, just saw that it is defined.)_

yakov116:
@fregante what your take on this? I don't think its worth it.

yakov116:
@dertieran I want to thank you for your work on this however I don't think its worth it.

Thanks for contributing!

PR #3621, opened 2020-10-06
Meta: Validate feature descriptions and functionality

yakov116:
<table>
	<thead>
		<tr>
			<th>Feature Name</th>
			<th>Readme Link</th>
			<th>Testing Link</th>
			<th>Tested/Working</th>
		</tr>
	</thead>
	<tbody>
		<tr>
			<td><a href="https://github.com/sindresorhus/refined-github/blob/master/source/features/action-used-by-link.tsx">action-used-by-link</a></td>
			<td><a href="https://github.com/sindresorhus/refined-github/blob/bef4b211ea364bf6cc5f5a24bf7651b6ffb2fc1c/readme.md#L328">#L328</a></td>
			<td><a href="https://github.com/marketplace/actions/urlchecker-action">Link</a></td>
			<td>✔️</td>
		</tr>
		<tr>
			<td><a href="https://github.com/sindresorhus/refined-github/blob/master/source/features/align-issue-labels.tsx">align-issue-labels</a></td>
			<td><a href="https://github.com/sindresorhus/refined-github/blob/bef4b211ea364bf6cc5f5a24bf7651b6ffb2fc1c/readme.md#L207">#L207</a></td>
			<td><a href="https://github.com/sindresorhus/refined-github/pulls">Link</a></td>
			<td>✔️</td>
		</tr>
		<tr>
			<td><a href="https://github.com/sindresorhus/refined-github/blob/master/source/features/batch-mark-files-as-viewed.tsx">batch-mark-files-as-viewed</a></td>
			<td><a href="https://github.com/sindresorhus/refined-github/blob/bef4b211ea364bf6cc5f5a24bf7651b6ffb2fc1c/readme.md#L240">#L240</a></td>
			<td></td>
			<td>✔️</td>
		</tr>
		<tr>
			<td><a href="https://github.com/sindresorhus/refined-github/blob/master/source/features/batch-open-conversations.tsx">batch-open-conversations</a></td>
			<td><a href="https://github.com/sindresorhus/refined-github/blob/bef4b211ea364bf6cc5f5a24bf7651b6ffb2fc1c/readme.md#L203">#L2013</a></td>
			<td></td>
			<td>✔️</td>
		</tr>
		<tr>
			<td><a href="https://github.com/sindresorhus/refined-github/blob/master/source/features/bugs-tab.tsx">bugs-tab</a></td>
[...]

PR #3651, opened 2020-10-15
Disable `edit-comments-faster` on hidden comments

kidonng:
Thanks for contributing! 🍄

1. LINKED ISSUES:
   https://github.com/sindresorhus/refined-github/pull/3407#issuecomment-707088160

2. TEST URLS:
   https://github.com/sindresorhus/refined-github/pull/3621

3. SCREENSHOT:
   None


kidonng:
I [have mentioned](https://github.com/sindresorhus/refined-github/pull/3407#issuecomment-665801560) that `edit-comments-faster` doesn't work if you unhide a comment. I've tried to find a simple way to fix that but failed. So here comes an alternative way to this issue: always append the edit icon before the three dots menu.

https://github.com/sindresorhus/refined-github/blob/149cddf538f1a53d716a45a9b4eb78ca49c30cdc/source/features/edit-comments-faster.tsx#L15

```diff
- comment.closest('.js-comment')!.querySelector('.js-comment-header-reaction-button')!.after(
+ comment.closest('.js-comment')!.querySelector('.timeline-comment-actions')!.lastElementChild!.before(
```

It may look like this:

image
image

Not sure if it is useful to edit hidden comments though, so thoughts welcome.

fregante (approved):
I think the current solution is perfect. As you mentioned, probably it’s not that useful to edit minimized comments.

yakov116 (approved):

yakov116:
Thanks @kidonng!

You helped me confirm that `wait-for-build` is working too!

PR #3653, opened 2020-10-15
Avoid errors caused by `more-dropdown` on issues

### LINKED ISSUE #3641: `GetCurrentBranch` Stopped working on issues
This is a major problem

Test on https://github.com/sindresorhus/refined-github/issues/3638 or even this issue.

yakov116:
1. LINKED ISSUES:
   Fixes #3641 

2. TEST URLS:
   #3641 

3. SCREENSHOT:
   None

### DIFF
--- a/source/github-helpers/github-url.ts
+++ b/source/github-helpers/github-url.ts
@@ -43,7 +43,7 @@ export default class GitHubURL {
 
 		const filePath = ambiguousReference.slice(1).join('/');
 
-		const currentBranch = getCurrentBranch();
+		const currentBranch = getCurrentBranch()!;
 		const currentBranchSections = currentBranch.split('/');
 		if (
 			ambiguousReference.length === 1 || // Ref has no slashes
--- a/source/github-helpers/github-url.ts
+++ b/source/github-helpers/github-url.ts
@@ -44,6 +44,9 @@ export default class GitHubURL {
 		const filePath = ambiguousReference.slice(1).join('/');
 
 		const currentBranch = getCurrentBranch();
+		if (!currentBranch) {
+			throw new Error('GitHubURL can only be used on pages with a branch/reference.')
+		}
 		const currentBranchSections = currentBranch?.split('/');
 		if (
 			!currentBranchSections || // Current page does not have a branch; GitHubURL probably shouldn’t be used here
--- a/source/github-helpers/github-url.ts
+++ b/source/github-helpers/github-url.ts
@@ -46,7 +46,7 @@ export default class GitHubURL {
 		const currentBranch = getCurrentBranch();
 		const currentBranchSections = currentBranch?.split('/');
 		if (
-			!currentBranch || // Current page does not have a branch; GitHubURL probably shouldn’t be used here
+			!currentBranchSections || // Current page does not have a branch; GitHubURL probably shouldn’t be used here
 			ambiguousReference.length === 1 || // Ref has no slashes
 			currentBranchSections.length === 1 // Current branch has no slashes
 		) {
--- a/source/github-helpers/github-url.ts
+++ b/source/github-helpers/github-url.ts
@@ -35,6 +35,7 @@ export default class GitHubURL {
 
 	private disambiguateReference(ambiguousReference: string[]): {branch: string; filePath: string} {
 		const branch = ambiguousReference[0];
+		// History pages might use search parameters
 		const filePathFromSearch = this.searchParams.getAll('path[]').join('/');
 		if (filePathFromSearch) {
 			this.searchParams.delete('path[]');
@@ -45,11 +46,11 @@ export default class GitHubURL {
 
 		const currentBranch = getCurrentBranch();
 		if (!currentBranch) {
-			throw new Error('GitHubURL can only be used on pages with a branch/reference.')
+			throw new Error('GitHubURL can only be used on pages with a branch/reference.');
 		}
-		const currentBranchSections = currentBranch?.split('/');
+
+		const currentBranchSections = currentBranch.split('/');
 		if (
-			!currentBranchSections || // Current page does not have a branch; GitHubURL probably shouldn’t be used here
 			ambiguousReference.length === 1 || // Ref has no slashes
 			currentBranchSections.length === 1 // Current branch has no slashes
 		) {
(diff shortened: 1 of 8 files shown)


source/github-helpers/index.ts
```diff
@@ -21,9 +21,15 @@ Tested on isRepoTree, isBlame, isSingleFile, isEditFile, isSingleCommit, isCommi
...
 Example tag content on private repositories https://github.com/private/private/commits/master.atom?token=AEAXKWNRHXA2XJ2ZWCMGUUN44LM62
 */
-export const getCurrentBranch = (): string => {
+export const getCurrentBranch = (): string | undefined => {
 	// .last needed for #2799
-	return new URL(select.last<HTMLLinkElement>('[type="application/atom+xml"]')!.href)
+	const atom = select.last<HTMLLinkElement>('[type="application/atom+xml"]');
+	// The atom URL is not available on `isIssue` #3641
+	if (!atom) {
+		return undefined;
```
fregante, line 29:
```suggestion
		return;
```

Doesn’t work? Or just `void` as the type?

fregante: I’m on mobile now but it sounds strange that there’s no atom feed on issues at all

yakov116: I agree! This was working all along.

source/github-helpers/index.ts
```diff
@@ -21,9 +21,15 @@ Tested on isRepoTree, isBlame, isSingleFile, isEditFile, isSingleCommit, isCommi
 Example tag content on public repositories: https://github.com/sindresorhus/refined-github/commits/branch-or-commit/even/with/slashes.atom
 Example tag content on private repositories https://github.com/private/private/commits/master.atom?token=AEAXKWNRHXA2XJ2ZWCMGUUN44LM62
 */
-export const getCurrentBranch = (): string => {
+export const getCurrentBranch = (): string | undefined => {
 	// .last needed for #2799
-	return new URL(select.last<HTMLLinkElement>('[type="application/atom+xml"]')!.href)
+	const atom = select.last<HTMLLinkElement>('[type="application/atom+xml"]');
```
notlmn, line 26: Could be more descriptive as

```suggestion
	const feedLink = select.last<HTMLLinkElement>('[type="application/atom+xml"]');
```

Or `repoFeedLink` to be a bit more verbose?

source/github-helpers/index.ts
```diff
@@ -21,9 +21,15 @@ Tested on isRepoTree, isBlame, isSingleFile, isEditFile, isSingleCommit, isCommi
...
+export const getCurrentBranch = (): string | undefined => {
 	// .last needed for #2799
-	return new URL(select.last<HTMLLinkElement>('[type="application/atom+xml"]')!.href)
+	const feedLink = select.last<HTMLLinkElement>('[type="application/atom+xml"]');
+	// The feedLink is not available on `isIssue` #3641
+	if (!feedLink) {
+		return;
+	}
+
+	return new URL(atom.href)
```
fregante, line 32:
```suggestion
	return new URL(feedLink.href)
```

fregante (approved):

yakov116:
@fregante when you approve can I assume the title is good?

source/github-helpers/github-url.ts
```diff
@@ -43,7 +43,7 @@ export default class GitHubURL {
 
 		const filePath = ambiguousReference.slice(1).join('/');
 
-		const currentBranch = getCurrentBranch();
+		const currentBranch = getCurrentBranch()!;
 		const currentBranchSections = currentBranch.split('/');
 		if (
 			ambiguousReference.length === 1 || // Ref has no slashes
```
fregante, line 49:
```suggestion
		const currentBranch = getCurrentBranch();
		const currentBranchSections = currentBranch?.split('/');
		if (
			!currentBranch || // Current page does not have a branch; GitHubURL shouldn’t really be used here
			ambiguousReference.length === 1 || // Ref has no slashes
```

source/github-helpers/github-url.ts
```diff
@@ -44,8 +44,9 @@ export default class GitHubURL {
 		const filePath = ambiguousReference.slice(1).join('/');
 
 		const currentBranch = getCurrentBranch();
-		const currentBranchSections = currentBranch.split('/');
+		const currentBranchSections = currentBranch?.split('/');
 		if (
+			!currentBranch || // Current page does not have a branch; GitHubURL probably shouldn’t be used here
```
fregante, line 49: This should fix the types
```suggestion
			!currentBranchSections || // Current page does not have a branch; GitHubURL probably shouldn’t be used here
```

yakov116:
https://github.com/sindresorhus/refined-github/blob/10dc4ac56c299c011733483a20725f3618a05175/source/github-helpers/github-url.ts#L65

Here is the issue

fregante:
Just fix the type issue with `!` or something.

I checked all the features where we use GitHubURL and all of them have a branch. In `forked-to` we specifically avoid errors by limiting the usage of GitHubURL to pages with a branch, even if the feature is enabled on `isRepo` https://github.com/sindresorhus/refined-github/blob/d7d700a20befb552f9f64e15504b845dbe04e699/source/features/forked-to.tsx#L34-L36

In another feature we also specifically mention the incompatibility so we don't use GitHubURL:
 https://github.com/sindresorhus/refined-github/blob/b7d31c6ae88ce48f4f5fa2716716a40e32eeea1b/source/features/comments-time-machine-links.tsx#L77

fregante:
Actually the ideal solution would be to throw an error when the page doesn't have a current branch to avoid silent errors.

source/github-helpers/github-url.ts
```diff
@@ -44,8 +44,9 @@ export default class GitHubURL {
 		const filePath = ambiguousReference.slice(1).join('/');
 
 		const currentBranch = getCurrentBranch();
```
fregante, line 46: How about this?
```suggestion
		const currentBranch = getCurrentBranch();
		if (!currentBranch) {
			throw new Error('GitHubURL can only be used on pages with a branch/reference.')
		}
```

fregante:
> @fregante when you approve can I assume the title is good?

Not always. Generally try to make the commit titles readable by users. "returns undefined" doesn't explain how this helps the user. In this case the specific user-facing bug being solved is ``Avoid errors caused by `more-dropdown` on issues`` or something along those lines

fregante (approved):

PR #3654, opened 2020-10-15
Make `releases-tab` non dependent of `more-dropdown`

### LINKED ISSUE #3652: `Release-tab` throws an error if `more-dropdown` is turned off
```
refined-github.js:5841 ❌ Refined GitHub → releases-tab → TypeError: Cannot read property 'before' of null
    at init (refined-github.js:9463)
    at async runFeature (refined-github.js:5881)
    at async setupPageLoad (refined-github.js:5894) 
Search issue: https://github.com/sindresorhus/refined-github/issues?q=is%3Aissue+Cannot%20read%20property%20'before'%20of%20null

Open an issue: https://github.com/sindresorhus/refined-github/issues/new?labels=bug&template=bug_report.md&title=%60releases-tab%60%3A%20Cannot%20read%20property%20'before'%20of%20null
```

yakov116:
1. LINKED ISSUES:
   Resolves #3652 

2. TEST URLS:
   None

3. SCREENSHOT:
   None

### DIFF
--- a/source/features/releases-tab.tsx
+++ b/source/features/releases-tab.tsx
@@ -71,7 +71,7 @@ async function init(): Promise<false | void> {
 		const releasesTab = (
 			<a
 				href={`/${repoUrl}/releases`}
-				className="js-selected-navigation-item UnderlineNav-item hx_underlinenav-item no-wrap js-responsive-underlinenav-item"
+				className="js-selected-navigation-item UnderlineNav-item no-wrap"
 				data-hotkey="g r"
 				data-selected-links="repo_releases"
 				data-tab-item="rgh-releases-item"
@@ -103,11 +103,16 @@ async function init(): Promise<false | void> {
 			releasesTab.setAttribute('aria-current', 'page');
 		}
 
-		select('.dropdown-divider', repoNavigationBar)!.before(
-			createDropdownItem('Releases', `/${repoUrl}/releases`, {
-				'data-menu-item': 'rgh-releases-item'
-			})
-		);
+		const moreDropdown = select('.dropdown-divider', repoNavigationBar);
+
+		if (moreDropdown) {
+			releasesTab.classList.add('hx_underlinenav-item', 'js-responsive-underlinenav-item');
+			moreDropdown.before(
+				createDropdownItem('Releases', `/${repoUrl}/releases`, {
+					'data-menu-item': 'rgh-releases-item'
+				})
+			);
+		}
 
 		// Hide redundant 'Releases' section from repo sidebar
 		if (pageDetect.isRepoRoot()) {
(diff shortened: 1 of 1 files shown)


fregante:
The dropdown is native, we’re not creating it. Perhaps the only problem is that there’s no native separator in there.

source/features/releases-tab.tsx
```diff
@@ -103,11 +103,16 @@ async function init(): Promise<false | void> {
 			releasesTab.setAttribute('aria-current', 'page');
 		}
 
-		select('.dropdown-divider', repoNavigationBar)!.before(
```
fregante, line 106: The only change I think should be like 
```js
		const position = select('.dropdown-divider', repoNavigationBar) ?? select('something else', repoNavigationBar);
		position.before(
```

source/features/releases-tab.tsx
```diff
@@ -103,11 +103,16 @@ async function init(): Promise<false | void> {
...
-			createDropdownItem('Releases', `/${repoUrl}/releases`, {
-				'data-menu-item': 'rgh-releases-item'
-			})
-		);
+		const moreDropdown = select('.dropdown-divider', repoNavigationBar);
+
+		if (moreDropdown) {
+			releasesTab.classList.add('hx_underlinenav-item', 'js-responsive-underlinenav-item');
```
yakov116, line 109: If this change is not done it errors out since github is looking to hide but it cant since it does not exist

source/features/releases-tab.tsx
```diff
@@ -103,11 +103,16 @@ async function init(): Promise<false | void> {
 			releasesTab.setAttribute('aria-current', 'page');
 		}
 
-		select('.dropdown-divider', repoNavigationBar)!.before(
-			createDropdownItem('Releases', `/${repoUrl}/releases`, {
-				'data-menu-item': 'rgh-releases-item'
-			})
-		);
+		const moreDropdown = select('.dropdown-divider', repoNavigationBar);
```
fregante, line 106: Actually we should just match the logic on line 85, which simply appends it. We only need to slightly modify because we have a divider. Thankfully we have an helper function that does this exactly. I think the only change needed is:

```suggestion
		appendBefore(
			select('.js-responsive-underlinenav .dropdown-menu ul'),
			'.dropdown-divider',

			createDropdownItem('Releases', `/${repoUrl}/releases`, {
				'data-menu-item': 'rgh-releases-item'
			})
		)
```

yakov116: Off topic alert!  Give me a min powering up the computer...

yakov116: Works perfectly!

fregante:
PR title needs update than it’s ok if it works

yakov116:
How is it now?

PR #3444, opened 2020-08-09
Minor style fixes for GitHub bugs

kidonng:
<details><summary>Edit: fix removed</summary>

Test: https://github.com/sindresorhus/refined-github/pull/3442

Before:
image

After:
image

</details>
---

Test: [`429a03a` (#3442)](https://github.com/sindresorhus/refined-github/pull/3442/commits/429a03a3dc22a7ad723e95b85d7b7f218b811b30)

Before:
image

After:
image

---

Test: https://github.com/sindresorhus/refined-github/commit/62c35395125d18d7ad53ce5688ef8603c735888b

Before:
image

After:
image

PR #3660, opened 2020-10-16
Meta: remove no-longer-necessary alignment CSS

kidonng:
- Add first fix from #3444 again (details on https://github.com/sindresorhus/refined-github/pull/3444#discussion_r468017909)
- Remove second fix from #3444 since it has been fixed by GitHub
- Added a new fix

<table>
<tr>
	<td> Test
	<td> Before
	<td> After
<tr>
	<td> This PR
	<td> 
	<td> 
<tr>
	<td> <a href="https://github.com/sindresorhus/refined-github/pull/3660/commits/ad3b3d20d36c017dcbc06a46277aa0ed595b309a">Link</a>
	<td> 
	<td> 

<tr>
	<td> <a href="https://github.com/sindresorhus/refined-github/blob/master/package.json">Link</a>
	<td> 
	<td> 
</table>

### DIFF
--- a/source/features/github-bugs.css
+++ b/source/features/github-bugs.css
@@ -17,11 +17,6 @@
 	border-bottom-right-radius: 5px;
 }
 
-/* Align author name on PR sticky header #3660 */
-.sticky-content .author.css-truncate-target {
-	vertical-align: -4px;
-}
-
 /* Align dropdown button on file page #3660 */
 #blob-more-options-details {
 	margin-top: 5.6px;
--- a/source/features/github-bugs.css
+++ b/source/features/github-bugs.css
@@ -10,16 +10,19 @@
 	margin: -5px 0;
 }
 
-/* Reposition committer's avatar on PR commits #3444 */
-.pull-request-tab-content .commit .AvatarStack {
-	display: inline-block;
-	margin-bottom: -4px;
-	margin-right: -4px;
-}
-
 /* Add missing border radius to collapsed PR files #3444 */
 .js-blob-wrapper,
 .file:not(.open) .file-header.file-header--expandable {
 	border-bottom-left-radius: 5px;
 	border-bottom-right-radius: 5px;
+}
+
+/* Align author name on PR sticky header */
+.sticky-content .author.css-truncate-target {
+	vertical-align: -4px;
+}
+
+/* Align dropdown button on file page */
+#blob-more-options-details {
+	margin-top: 5.6px;
 }
--- a/source/features/github-bugs.css
+++ b/source/features/github-bugs.css
@@ -16,8 +16,3 @@
 	border-bottom-left-radius: 5px;
 	border-bottom-right-radius: 5px;
 }
-
-/* Align dropdown button on file page #3660 */
-#blob-more-options-details {
-	margin-top: 5.6px;
-}
--- a/source/features/github-bugs.css
+++ b/source/features/github-bugs.css
@@ -17,12 +17,12 @@
 	border-bottom-right-radius: 5px;
 }
 
-/* Align author name on PR sticky header */
+/* Align author name on PR sticky header #3660 */
 .sticky-content .author.css-truncate-target {
 	vertical-align: -4px;
 }
 
-/* Align dropdown button on file page */
+/* Align dropdown button on file page #3660 */
 #blob-more-options-details {
 	margin-top: 5.6px;
 }


source/features/github-bugs.css
```diff
@@ -10,16 +10,19 @@
...
 }
+
+/* Align author name on PR sticky header */
+.sticky-content .author.css-truncate-target {
+	vertical-align: -4px;
+}
+
+/* Align dropdown button on file page */
+#blob-more-options-details {
+	margin-top: 5.6px;
```
kidonng, line 27: This is to align the dropdown button (whose height is `20.8px`) with the find file button (whose height is `32px`)

(32 - 20.8) / 2 = 5.6

source/features/github-bugs.css
```diff
@@ -10,16 +10,19 @@
...
-}
-
 /* Add missing border radius to collapsed PR files #3444 */
 .js-blob-wrapper,
 .file:not(.open) .file-header.file-header--expandable {
 	border-bottom-left-radius: 5px;
 	border-bottom-right-radius: 5px;
 }
+
+/* Align author name on PR sticky header */
```
kidonng, line 20: This issue has been confirmed on the latest Chrome 86 (table build)

source/features/github-bugs.css
```diff
@@ -10,16 +10,19 @@
...
 .file:not(.open) .file-header.file-header--expandable {
 	border-bottom-left-radius: 5px;
 	border-bottom-right-radius: 5px;
 }
+
+/* Align author name on PR sticky header #3660 */
+.sticky-content .author.css-truncate-target {
+	vertical-align: -4px;
+}
+
```
fregante, line 24: We don’t need to fix this since GitHub will fix it if/when necessary.

source/features/github-bugs.css
fregante: Give it time. I don’t want more CSS code to fix something they probably will fix soon

kidonng: This issue has been there for a while, so does the misaligned dropdown (though the former one is only noticeable in newer versions of Chrome).
But I agree these fixes should not be hurried. Do we have some staging area for this kind of improvements (like an issue)?

andi73-bit (approved):

source/features/github-bugs.css
fregante: 1. It's just on your computer



2. This CSS affects everyone, not just Chrome users, so if this fixes it for you, it breaks everyone else’s

3. We don't even show the username anymore thanks to `clean-conversation-headers`, so this "fix" shouldn't be part of RG anyway in the first place.

source/features/github-bugs.css
fregante: There's `display: flex` on that line, just add `align-items: center !important;` on `.d-flex` to align the whole line since it's not aligned:

source/features/github-bugs.css
fregante: Is it possible that you have corrupted font files?

notlmn:
Hey @kidonng from the first set to screenshots you shared, looks like you have some RGH feature disabled.
This is how it looks like for PRs from external source branches

image

and this for internal source branches

image

I don't see the entire _"__fregante__ wants to merge 3 commits into __`master`__ from __`remove-repeat-on-ajax`__"_, I remember this being a feature, don't quite remember which one.

yakov116:
It's from the sticky bar.

fregante:
> It's from the sticky bar.

It confused me too at first, that's why screenshots should include a little more context sometimes.

source/features/github-bugs.css
kidonng: > 1. It's just on your computer
> 

I just asked a fellow and he can also reproduce it. We are both on Windows, is that the problem?

I will test it on Linux later

busches: On Chrome 86 on Windows as well and it looks fine.

fregante: Are you and your fellow on zh-CN locales? I don’t have experience with that, but I would guess that it might affect alignment.

kidonng: > Are you and your fellow on zh-CN locales?

You are right! The issue disappeared right after I changed my locale.

Then I have no problem with removing the fix. :+1:  I will keep it in my custom style.

source/features/github-bugs.css
fregante: I hope this is fixed across the board, the whole bar is probably the most broken-looking part of GitHub/Refined GitHub. Example: https://github.com/sindresorhus/refined-github/issues/3671 (not totally RG’s fault, but every element has its own alignment)

fregante:
I’ll merge the part of this PR that is ready since the rest needs more work.

PR #3661, opened 2020-10-16
Fix `profile-hotkey` on Gist pages

yakov116:
1. LINKED ISSUES:
   None

2. TEST URLS:
   https://gist.github.com/yakov116

3. SCREENSHOT:
   None

It did not work there since `/${getUsername()` leads to the `Your gists` button

### DIFF
--- a/source/features/profile-hotkey.tsx
+++ b/source/features/profile-hotkey.tsx
@@ -2,15 +2,9 @@ import select from 'select-dom';
 import onetime from 'onetime';
 
 import features from '.';
-import {getUsername} from '../github-helpers';
 
-function init(): false | void {
+function init(): void {
 	select('a[data-ga-click$="text:your profile"]')!.dataset.hotkey = 'g m';
-	if (!menuItem) {
-		return false;
-	}
-
-	menuItem.dataset.hotkey = 'g m';
 }
 
 void features.add({
--- a/source/features/profile-hotkey.tsx
+++ b/source/features/profile-hotkey.tsx
@@ -5,7 +5,7 @@ import features from '.';
 import {getUsername} from '../github-helpers';
 
 function init(): false | void {
-	const menuItem = select.last(`[aria-label="View profile and more"] ~ details-menu a[href$="/${getUsername()}"]`);
+	select('a[data-ga-click$="text:your profile"]')!.dataset.hotkey = 'g m';
 	if (!menuItem) {
 		return false;
 	}
--- a/source/features/profile-hotkey.tsx
+++ b/source/features/profile-hotkey.tsx
@@ -5,7 +5,7 @@ import features from '.';
 import {getUsername} from '../github-helpers';
 
 function init(): false | void {
-	const menuItem = select(`a[href="/${getUsername()}"]`);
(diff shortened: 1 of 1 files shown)


source/features/profile-hotkey.tsx
```diff
@@ -5,7 +5,7 @@ import features from '.';
 import {getUsername} from '../github-helpers';
 
 function init(): false | void {
-	const menuItem = select(`a[href="/${getUsername()}"]`);
+	const menuItem = select.last(`[aria-label="View profile and more"] ~ details-menu a[href$="/${getUsername()}"]`);
```
notlmn, line 8: Shouldn't this be?

```suggestion
	const menuItem = select.last(`[aria-label="View profile and more"] ~ details-menu a[href^="/${getUsername()}"]`);
```

I mean, both work, but I'm a bit inclined towards `^` for no specific reason.

yakov116: No. The url on the gist page is `https://github.com/yakov116`
`/${getUsername()` = `https://gist.github.com/yakov116`

yakov116: Updated the PR description

fregante: Let's cut to the chase and just use something like:

```js
const url = '/' + getUsername();
document.body.append(<a href={isGist() && isEnterprise() ? location.origin + url : url } hidden data-hotkey="g m"/>)
```

or `document.head`/`<link>` if that works

fregante: But in reality this was pretty simple:

```suggestion
	const menuItem = select('[data-ga-click="Header, go to profile, text:your profile"]');
```

notlmn: > RG is only expected to work on logged in users.

Yes, but at least we can fail gracefully. Would help us as well, if all of a sudden GitHub decides to change class names and other stuff, we can still fail gracefully.

BUT that would beg the question of how we would be able to detect a failure. Errors is how we get most of the reports from users, raising issues.

yakov116: >But in reality this was pretty simple:

OMG how did I not catch that?!

yakov116:
do ga's work on enterprise?

fregante:
That’s a good question. We use them in a few features and nobody complained though

source/features/profile-hotkey.tsx
fregante: > Yes, but at least we can fail gracefully. 

Nah. Do you really wanna check all the features on logged out pages? I don’t. That’s why we display a warning in that case. 

> if all of a sudden GitHub decides to change class names and other stuff, we can still fail gracefully.

Unless a feature has the potential to lose data or break native functionality, we don’t need that. At most we display an alert if the feature is a complex action like `restore-files`

In this case, the previous code silently failed until Yakov went testing all the features; it’s better it let the error appear in the console so someone fixes it.

fregante:
Let's merge this in the meanwhile. We can decide later what we can do about GHE support (if anything)

PR #3670, opened 2020-10-19
Fetch widget for `view-markdown-source` instead of maintaining a copy of it

yakov116:
1. LINKED ISSUES:
   https://github.com/sindresorhus/refined-github/pull/3594#issuecomment-711839565

2. TEST URLS:
   https://github.com/sindresorhus/refined-github/blob/3d517d6cd00e83a51973808c7ff03cae2b692d2b/contributing.md#L3-L5

3. SCREENSHOT:
   None

@fregante why do you say we need to exclude enterprise?

### DIFF
--- a/source/background.ts
+++ b/source/background.ts
@@ -45,9 +45,8 @@ addDomainPermissionToggle();
 
 // `background` fetch required to avoid avoid CORB introduced in Chrome 73 https://chromestatus.com/feature/5629709824032768
 // Don’t turn this into an `async` function https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/onMessage#addListener_syntax
-browser.runtime.onMessage.addListener(({request, json}): Promise<Response> | void => {
-	return fetch(request).then(async response => {
-		const textContent = await response.text();
-		return json ? JSON.parse(textContent) : textContent;
-	});
+browser.runtime.onMessage.addListener((message): Promise<string> | void => {
+	if (message?.request) {
+		return fetch(request).then(async response => response.text());
+	}
 });
--- a/source/background.ts
+++ b/source/background.ts
@@ -47,6 +47,6 @@ addDomainPermissionToggle();
 // Don’t turn this into an `async` function https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/onMessage#addListener_syntax
 browser.runtime.onMessage.addListener((message): Promise<string> | void => {
 	if (message?.request) {
-		return fetch(request).then(async response => response.text());
+		return fetch(message.request).then(async response => response.text());
 	}
 });
--- a/source/background.ts
+++ b/source/background.ts
@@ -45,6 +45,9 @@ addDomainPermissionToggle();
 
 // `background` fetch required to avoid avoid CORB introduced in Chrome 73 https://chromestatus.com/feature/5629709824032768
 // Don’t turn this into an `async` function https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/onMessage#addListener_syntax
-browser.runtime.onMessage.addListener(({request}): Promise<Response> | void => {
-	return fetch(request).then(async response => response.json());
+browser.runtime.onMessage.addListener(({request, json}): Promise<Response> | void => {
+	return fetch(request).then(async response => {
+		const textContent = await response.text();
+		return json ? JSON.parse(textContent) : textContent;
+	});
 });
--- a/source/features/embed-gist-inline.tsx
+++ b/source/features/embed-gist-inline.tsx
@@ -20,7 +20,7 @@ async function embedGist(link: HTMLAnchorElement): Promise<void> {
 
 	try {
 		// Get the gist via background.js due to CORB policies introduced in Chrome 73
-		const gistData = await browser.runtime.sendMessage({request: `${link.href}.json`});
+		const gistData = await browser.runtime.sendMessage({request: `${link.href}.json`, json: true});
 
 		const files = domify.one(gistData.div)!;
 		const fileCount = files.children.length;
--- a/source/features/embed-gist-inline.tsx
+++ b/source/features/embed-gist-inline.tsx
@@ -20,9 +20,9 @@ async function embedGist(link: HTMLAnchorElement): Promise<void> {
 
 	try {
 		// Get the gist via background.js due to CORB policies introduced in Chrome 73
-		const gistData = await browser.runtime.sendMessage({request: `${link.href}.json`, json: true});
+		const gistData = await browser.runtime.sendMessage({request: `${link.href}.json`});
 
-		const files = domify.one(gistData.div)!;
+		const files = domify.one(JSON.parse(gistData).div)!;
 		const fileCount = files.children.length;
 
 		if (fileCount > 1) {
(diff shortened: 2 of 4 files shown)


source/features/view-markdown-source.tsx
```diff
@@ -119,7 +65,7 @@ async function showSource(): Promise<void> {
 	sourceButton.classList.add('selected');
 	renderedButton.classList.remove('selected');
 	blurButton(sourceButton);
-	(await source).before(lineActions());
+	(await source).before(await lineActions() as unknown as HTMLDetailsElement);
 
 	dispatchEvent(sourceButton, 'rgh:view-markdown-source');
```
fregante, line 70: We can dispatch this event **before** the line actions are added

source/features/view-markdown-source.tsx
```diff
@@ -119,7 +65,7 @@ async function showSource(): Promise<void> {
 	sourceButton.classList.add('selected');
 	renderedButton.classList.remove('selected');
 	blurButton(sourceButton);
-	(await source).before(lineActions());
+	(await source).before(await lineActions() as unknown as HTMLDetailsElement);
```
fregante, line 68: The assertions should not be needed, just have the function return the right type

source/features/view-markdown-source.tsx
```diff
@@ -7,73 +7,19 @@ import * as pageDetect from 'github-url-detection';
...
-					>
-						Reference in new issue
-					</a>
-				</li>
-			</ul>
-		</details-menu>
-	</details>
-));
+const lineActions = onetime(async () => {
+	// Avoid having to create the entire 70 lines of JSX. The URL is hardcoded to support enterprise
```
fregante, line 17: CORB will likely block this request on enterprise, but this needs to be confirmed. Either way the reason why this is hardcoded is that we just need the DOM from a file we know exists.

fregante (changes requested):

source/features/view-markdown-source.tsx
yakov116: background request?

source/helpers/fetch-dom.ts
```diff
@@ -2,11 +2,17 @@ import mem from 'mem';
...
 	const absoluteURL = new URL(url, location.origin).toString(); // Firefox `fetch`es from the content script, so relative URLs fail
-	const response = await fetch(absoluteURL);
-	const dom = domify(await response.text());
+	let dom;
+	if (backgroundRequest) {
+		dom = domify(await browser.runtime.sendMessage({request: absoluteURL}));
+	} else {
+		const response = await fetch(absoluteURL);
+		dom = domify(await response.text());
+	}
```
yakov116, line 14: I don't like what I did here

source/background.ts
```diff
@@ -45,6 +45,9 @@ addDomainPermissionToggle();
 
 // `background` fetch required to avoid avoid CORB introduced in Chrome 73 https://chromestatus.com/feature/5629709824032768
 // Don’t turn this into an `async` function https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/onMessage#addListener_syntax
-browser.runtime.onMessage.addListener(({request}): Promise<Response> | void => {
-	return fetch(request).then(async response => response.json());
+browser.runtime.onMessage.addListener(({request, json}): Promise<Response> | void => {
+	return fetch(request).then(async response => {
+		const textContent = await response.text();
+		return json ? JSON.parse(textContent) : textContent;
+	});
```
fregante, line 52: Just have this always return `.text()`. Leave JSON out

source/features/embed-gist-inline.tsx
```diff
@@ -20,7 +20,7 @@ async function embedGist(link: HTMLAnchorElement): Promise<void> {
 
 	try {
 		// Get the gist via background.js due to CORB policies introduced in Chrome 73
-		const gistData = await browser.runtime.sendMessage({request: `${link.href}.json`});
+		const gistData = await browser.runtime.sendMessage({request: `${link.href}.json`, json: true});
 
 		const files = domify.one(gistData.div)!;
```
fregante, line 25:
```suggestion
		const gistData = await browser.runtime.sendMessage({request: `${link.href}.json`});

		const files = domify.one(JSON.parse(gistData).div)!;
```

source/features/view-markdown-source.tsx
```diff
@@ -7,73 +7,19 @@ import * as pageDetect from 'github-url-detection';
...
-						Reference in new issue
-					</a>
-				</li>
-			</ul>
-		</details-menu>
-	</details>
-));
+const lineActions = onetime(async () => {
+	// Avoid having to create the entire 60 lines of JSX. The URL is hardcoded to a file we know the DOM exists.
+	const blobToolbar = await fetchDom('https://github.com/sindresorhus/refined-github/blob/b1229bbaeb8cf071f0711bc2ed1b40dd96cd7a05/.editorconfig', '.BlobToolbar', true);
```
fregante, line 18: Don't taint `fetchDom` with this, it's unlikely we ever need it again.

```suggestion
	const randomKnownFile = 'https://github.com/sindresorhus/refined-github/blob/b1229bbaeb8cf071f0711bc2ed1b40dd96cd7a05/.editorconfig';
	const html = await browser.runtime.sendMessage({request: randomKnownFile});
	const blobToolbar = domify(html).querySelector('.BlobToolbar');
```

fregante (changes requested):

source/background.ts
fregante, line 52: Also, I just realized that this lacked the existence check. It never returned `void`
```suggestion
browser.runtime.onMessage.addListener(({request}): Promise<Response> | void => {
	if (request) {
		return fetch(request).then(async response => response.text());
	}
```

fregante: Also it receives an `any` which can also causes errors in the destructuring:

```suggestion
browser.runtime.onMessage.addListener((message): Promise<string> | void => {
	if (message?.request) {
		return fetch(request).then(async response => response.text());
	}
```

And the type is wrong, it's does not return a response.

yakov116:
Working 🎉 

Thanks for confirming my thoughts
 > Don't taint `fetchDom` with this, it's unlikely we ever need it again.

source/background.ts
```diff
@@ -45,6 +45,8 @@ addDomainPermissionToggle();
 
 // `background` fetch required to avoid avoid CORB introduced in Chrome 73 https://chromestatus.com/feature/5629709824032768
 // Don’t turn this into an `async` function https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/onMessage#addListener_syntax
-browser.runtime.onMessage.addListener(({request}): Promise<Response> | void => {
-	return fetch(request).then(async response => response.json());
+browser.runtime.onMessage.addListener((message): Promise<string> | void => {
+	if (message?.request) {
+		return fetch(message.request).then(async response => response.text());
```
fregante, line 50: Does this fail lint? I don't remember if we disabled that rule
```suggestion
		return fetch(message.request).then(response => response.text());
```

source/features/view-markdown-source.tsx
```diff
@@ -137,7 +86,7 @@ async function showRendered(): Promise<void> {
 	sourceButton.classList.remove('selected');
 	renderedButton.classList.add('selected');
 	blurButton(renderedButton);
-	lineActions().remove();
+	(await lineActions()).remove();
 
 	dispatchEvent(sourceButton, 'rgh:view-markdown-rendered');
```
fregante, line 91: Same goes here
```suggestion
	dispatchEvent(sourceButton, 'rgh:view-markdown-rendered');

	(await lineActions()).remove();
```

fregante (approved):

source/background.ts
yakov116: No we kept is since it was only one place

yakov116:
Is the title ok?

Issue #14, opened 2016-02-17
Safari version: help us setup the build and deployment


fantattitude:
Would be great to have a Safari version as well 😁

sindresorhus:
Shouldn't be hard to adapt the code. Feel free to steal the code and make a Safari extension. I'm not going to, though.

fantattitude:
I'll probably PR later today then ;P

sindresorhus:
I'm not interested in maintaining it. Just fork the repo and do your thing.

fantattitude:
What if I propose you a PR making the build process automatic like what does Octotree (builds for every navigator are automated through Gulp) ? https://github.com/buunguyen/octotree

sindresorhus:
@fantattitude No. Sorry, not trying to be difficult, but I've gone through this before and made extensions for all browsers, and it was such pain. Even if you automate the build-step, I will still receive issues regarding the Safari extension. Can't you just maintain it on your own, and I'd be happy to link to your repo.

fantattitude:
Oh yeah don't worry 👌🏻
I was just asking I can totally understand your point. Supporting with several browsers can be painful. 

I'll do that then, thanks for your time and nice work on the extension !

fantattitude:
Hi @sindresorhus, I made a repo with the Safari extension at https://github.com/fantattitude/refined-github-safari, feel free to tell me if everything's okay for you (regarding credits for example) and link to it on your README 😉

sindresorhus:
@fantattitude Cool. Linked to it from the readme :)

nesl247:
I suppose this could be reopened then?

sindresorhus:
Yes! https://developer.apple.com/documentation/safariservices/safari_web_extensions/converting_a_web_extension_for_safari

ngyikp:
Running `xcrun safari-web-extension-converter` on the Chrome Web Store version without any modifications seem to work at first glance:

<img width="1440" alt="refined github" src="https://user-images.githubusercontent.com/6135313/85346258-d9592100-b4ec-11ea-81b9-aaff66e94815.png">

nesl247:
@ngyikp Can you post some instructions on how you did this?

tbodt:
Can you distribute a safari extension outside of the Mac App Store? If so do you need to notarize it (and thus pay for a $100 developer account)?

sindresorhus:
[...]

PR #3677, opened 2020-10-21
Add unpublished Safari build

fregante:
Related to #14 

This PR ~~currently doesn't include any custom code,~~ it's ~~just~~ the direct result of:

```
xcrun safari-web-extension-converter distribution/ --project-location safari --app-name 'Refined GitHub' --bundle-identifier 'com.sindresorhus.refinedGitHub'
```

Next:

- [x] Find solution the lack of [negative lookbehinds support](https://bugs.webkit.org/show_bug.cgi?id=174931) 🤦‍♂️ 
	<details>
		
	
	
	I saw that @lautis’ wrapper forces compatible versions of those modules for Safari: 
	
	https://github.com/lautis/refined-github-safari/commit/8eebe28037ea18d63ab4d129c25dbcc380df00ba#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L10
	
	We can just undo https://github.com/sindresorhus/refined-github/pull/3316

	</details>

- [x] Cleanup options CSS
- [x] Fix options’s filter and fit-textarea
- [x] Find out why `permission` hosts aren't automatically allowed


safari/Refined GitHub/Refined GitHub/ViewController.swift
```diff
@@ -0,0 +1,51 @@
...
+//
+//  Created by Rico on 21/10/20.
+//
+
+import Cocoa
+import SafariServices.SFSafariApplication
+import SafariServices.SFSafariExtensionManager
+
+let appName = "Refined GitHub"
+let extensionBundleIdentifier = "com.sindresorhus.Refined-GitHub-Extension"
```
sindresorhus, line 13:
```suggestion
let extensionBundleIdentifier = "com.sindresorhus.Refined-GitHub"
```

Since that's the bundle ID I registered.

sindresorhus:
You need to gitignore:

```
xcuserdata
project.xcworkspace
```

fregante:
> You need to gitignore:

That's what I was hoping to find out with this PR 🎉 

---

Also, initial Safari compatibility achieved 🎉 


Safari compatibility achieved:

fregante:
Firefox vs Safari vs Chrome

fregante:
## How to test

1. Install XCode 12 and Safari 14
2. You probably need to enable **Allow Unsigned Extensions** in Safari’s **Develop** menu.
3. Run:

	```sh
	npm run build
	npm run pack:safari
	npm run start:safari
	```

4. I also had to click **Edit websites** under Safari **Preferences -> Extensions -> Refined GitHub** and manually allow GitHub.com.

## Questions

4. About the number 4 above: Is this because it's unsigned? Does it happen for you as well? How can we avoid this?

## Next

I think this PR won't need substantial more work, but the next steps will be:

- [ ] @sindresorhus sign the extension (?)
- [ ] Set up Fastlane

fregante:
**Regarding point 4:** I think there's no way around it. `permissions` is essentially ignored and always acts as `optional_permissions`, the user has to enable the extension. More information from ~10:00 to ~15:00 in the video: https://developer.apple.com/videos/play/wwdc2020/10665/


Enterprise support in Safari appears easier, but requires changes:

- Safari offers a UI to enable the extension on any domain, so [webext-domain-permission-toggle](https://github.com/fregante/webext-domain-permission-toggle) needs to be dropped
- [webext-dynamic-content-scripts](https://github.com/fregante/webext-dynamic-content-scripts) doesn't seem to work, but I think this can be replaced by `<all_urls>` under `content_scripts`. This is perfect, but it requires a custom `manifest.json` for Safari
- **The domain selector currently appears for everyone,** so it needs to be hidden in Safari. The cause appears to be https://github.com/fregante/webext-additional-permissions/issues/1

lautis:
You already use the nolookbehind variants, so that's fixed :+1:

Another patch I have for Safari is the following CSS:

```css
.rgh-parse-backticks {
  display: inline-block;
}
```

This fixes line-wrapping of code blocks in issues list.

fregante:
> Another patch I have for Safari is the following CSS:

Thanks for making me notice that. It looks like a Safari bug. `inline-block` will prevent wrapping though, but it looks like moving the `sr-only` elements outside `rgh-parse-backticks` fixes it:

https://github.com/sindresorhus/refined-github/blob/a38244e63e108b141745a80f0264ebd1ff6740b0/source/github-helpers/parse-backticks.tsx#L11-L15

PR #3681, opened 2020-10-22
Restore `faster-reviews` autofocus

FloEdelmann:
As mentioned in #3621. You can test it right here in this PR :wink:


fregante:
I think this is a GitHub but we could wait to be fixed.

This feature merely mirrors the native "Add your review" button that appears when someone requests your review:

FloEdelmann:
I use it all the time, that's why I wanted to fix it. For quite some time now we are adding the "review now" button for no good reason, it leads to exactly the same page/state as the "Files" tab does.

I didn't know about that other place where the link is used as well. So maybe GitHub has actually an incentive to fix it themselves.

fregante (approved):

fregante:
Alternatively, this could be implemented as an additional init in the same file so that it fixes the native `#submit-review` hash as well.

FloEdelmann:
I've changed it back to `#submit-review`.

> could be implemented as an additional init in the same file

Why an additional init?

PR #3657, opened 2020-10-15
Use a consistent border color

ocarreterom:
- move border color to CSS variable
- normalize gray border color to `#e1e4e8`

Thanks for contributing! 🍄

1. LINKED ISSUES:
   No issue

2. TEST URLS:
   https://github.com/ (dashboard)

3. SCREENSHOT:
The border after "All activity" was darker.
<table>
<tr>
	<td>Before
	<td>After
<tr>
	<td>Captura de pantalla 2020-10-16 a las 1 07 09
	<td>Captura de pantalla 2020-10-16 a las 1 06 46
</table>

PR #3684, opened 2020-10-23
Lint

yakov116:
- Update XO and Regenerate package-lock
- Enable new XO Rules
- Make `api.v3` default to `repos/${getRepoURL}`
- Add `getPRRepositoryInfo`
- November Lint
- Change `getRepoURL` to use `getPRRepositoryInfo`


yakov116:
@fregante did you regenerate the `package-lock.json` after #3639?

I was having issues so I regenerated.

source/globals.d.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable @typescript-eslint/consistent-indexed-object-style */
```
yakov116, line 1: Had to do this because all the types would get messed up

source/features/show-open-prs-of-forks.tsx
```diff
@@ -32,7 +32,7 @@ const countPRs = cache.function(async (forkedRepo: string): Promise<[prCount: nu
 	`);
 
 	// Only show PRs originated from the current repo
-	const prs = search.nodes.filter((pr: AnyObject) => pr.headRepository.nameWithOwner.toLowerCase() === getRepoURL());
+	const prs = search.nodes.filter((pr: AnyObject) => pr.headRepository.nameWithOwner.toLowerCase() === getRepositoryInfo().url!);
```
yakov116, line 35: This may be an issue since we need to do `.toLowerCase()`

source/github-helpers/index.ts
```diff
@@ -39,8 +39,6 @@ export const getCurrentBranch = (): string | undefined => {
 
 export const isFirefox = navigator.userAgent.includes('Firefox/');
 
-export const getRepoURL = (): string => location.pathname.slice(1).split('/', 2).join('/').toLowerCase();
```
yakov116, line 42: `.toLowerCase()` was dropped. We may need to rethink.

source/globals.d.ts
yakov116: I think its https://github.com/typescript-eslint/typescript-eslint/pull/2678 which is not released yet

fregante:
What timezone are you in?

source/globals.d.ts
```diff
@@ -85,6 +87,5 @@ declare module '*.svg' {
 // Make `element.cloneNode()` preserve its type instead of returning Node
 interface Node extends EventTarget {
 	// Not equivalent
```
fregante, line 89: This line was the explanation

source/globals.d.ts
fregante: Let's wait for it

source/github-helpers/parse-backticks.tsx
```diff
@@ -5,7 +5,7 @@ const splittingRegex = /`(.*?)`/g;
 export default function parseBackticks(description: string): DocumentFragment {
 	const fragment = new DocumentFragment();
 	for (const [index, text] of description.split(splittingRegex).entries()) {
-		if (index % 2 && text.length >= 1) {
+		if (index % 2 && text.length > 0) {
```
fregante, line 8: ?

source/github-helpers/api.ts
```diff
@@ -103,7 +105,7 @@ export const v3 = mem(async (
 	const {personalToken} = await settings;
 
 	if (query.startsWith('/')) {
-		throw new TypeError('The query parameter must not start with a slash.');
+		query = `repos/${getRepoURL()}${query}`;
```
fregante, line 108: It should be the opposite. `/repos` marks the root while `${query}` would be relative to the current repo.

```js
	if (query.startsWith('/')) {
		query = query.slice(1);
	} else {
 		query = `repos/${getRepoURL()}${query}`;
```

Also please please avoid using template literals unless there are multiple strings being interspersed. Look how much more understandable this is without having to figure out if there's any significant character in the sea of `${}${}`

```js
 		query = 'repos/' + getRepoURL() + query;
```

yakov116:
Eastern

source/features/index.tsx
```diff
@@ -52,12 +52,12 @@ interface InternalRunConfig {
...
 let log: typeof console.log;
 
-function logError(id: FeatureID, error: Error | string, ...extras: unknown[]): void {
+function logError(id: FeatureID, error: Error | string | unknown, ...extras: unknown[]): void {
 	if (error instanceof TypeError && error.message === 'Object(...)(...) is null') {
 		error.message = 'The element wasn’t found, the selector needs to be updated.';
 	}
 
-	const message = typeof error === 'string' ? error : error.message;
+	const message = typeof error === 'string' ? error : (error as Error).message;
```
fregante, line 60: Avoid assertions when possible. The line already has a typecheck.
```suggestion
	const message = error instanceof Error ? error.message : error; 
```

source/features/deep-reblame.tsx
```diff
@@ -75,9 +75,9 @@ async function redirectToBlameCommit(event: delegate.Event<MouseEvent, HTMLAncho
 		blameUrl.branch = await getPullRequestBlameCommit(prCommit, prNumber, blameUrl.filePath);
 		blameUrl.hash = 'L' + select('.js-line-number', blameHunk)!.textContent!;
 		location.href = String(blameUrl);
-	} catch (error) {
+	} catch (error: unknown) {
```
fregante, line 78: Judging by the alert, we expect an `Error`

fregante (changes requested):

source/github-helpers/parse-backticks.tsx
yakov116: Xo rule

source/features/deep-reblame.tsx
yakov116: But the rule only allows `any` or `unknown`

fregante: Indeed. Sounds like a useless rule. Let's disable it instead

source/github-helpers/parse-backticks.tsx
fregante: Heh didn't realize it does the same thing

source/features/deep-reblame.tsx
yakov116: What is the point of the rule? Sindre just enabled in that last update.

source/features/user-local-time.tsx
```diff
@@ -34,15 +34,15 @@ async function loadCommitPatch(commitUrl: string): Promise<string> {
...
+	for await (const page of api.v3paginated(`/users/${login}/events`)) {
 		for (const event of page as any) {
 			if (event.type !== 'PushEvent') {
 				continue;
 			}
 
 			// Start from the latest commit, which is the last one in the list
 			for (const commit of event.payload.commits.reverse() as Commit[]) {
-				const response = await api.v3(commit.url, {ignoreHTTPStatus: true});
+				const response = await api.v3('/' + commit.url, {ignoreHTTPStatus: true});
```
fregante, line 45: I'm not entirely sure, but I'd think that `commit.url` has the WHOLE URL already, including `https:`. Please verify

source/github-helpers/api.ts
```diff
@@ -6,7 +6,8 @@ next to the name of the feature that caused them.
 Usage:
 
 import * as api from '../github-helpers/api';
-const user = await api.v3(`users/${username}`);
+If the query starts with `repos/getRepoURL` start the query without a `/`
+const user = await api.v3(`commits`);
```
fregante, line 10:
```suggestion
const user = await api.v3(`/users/${username}`);
const repositoryCommits = await api.v3('commits'); // Without a leading `/`, this is equivalent to `/repo/$current-repository/commits`
```

source/features/index.tsx
```diff
@@ -43,12 +43,12 @@ interface InternalRunConfig {
...
 let log: typeof console.log;
 
-function logError(id: FeatureID, error: Error | string, ...extras: unknown[]): void {
+function logError(id: FeatureID, error: Error | string | unknown, ...extras: unknown[]): void {
 	if (error instanceof TypeError && error.message === 'Object(...)(...) is null') {
 		error.message = 'The element wasn’t found, the selector needs to be updated.';
 	}
 
-	const message = typeof error === 'string' ? error : error.message;
+	const message = error instanceof Error ? error.message : error as string;
```
fregante, line 51: If the code expects a string, make it a string
```suggestion
	const message = error instanceof Error ? error.message : String(error);
```

source/features/deep-reblame.tsx
fregante: `any` lets you use variable however you want to use them while `unknown` forces you to type them before use. That's my understanding.

`unknown` forces us to check the type of the error before using it, but if we're just going to use assertions everywhere this isn't nearly a helpful.

We can leave it enabled

fregante (approved):
This can be merged once the few last few comments I left are resolved

source/features/user-local-time.tsx
yakov116: It does and with the current logic the query end up being  repos/username + the link. Maybe we should rather add another option of `absoluteURL`.

yakov116: @fregante mind to give this a quick look?

fregante: Ah true. You can just detect wether it starts with http and skip any transforms. No options necessary

fregante:
Likewise, we can change the v4 to automatically inject the current owner/name when it encounters an empty `'repository()'` string and we can **delete** the awkward `getRepoGQL` function

yakov116:
Agree but I was not sure how to implement it. Will try on the next round.

fregante:
I think it'd be something like:

```js
query = query.replace('repository() {', `repository(owner: ${…} name: ${…}) {`)

fregante:
Wait though because https://github.com/sindresorhus/refined-github/pull/3695 changes the same lines

yakov116:
Its not November yet... (at least in my time zone) 😃

fregante:
I'll start adding ideas to "the last lint pr":

- https://github.com/sindresorhus/refined-github/pull/3657#issuecomment-720115096

PR #3596, opened 2020-09-25
Add `git-checkout-pr` feature

yakov116:
1. LINKED ISSUES:
   Closes #2767 

2. TEST URLS:
Remote URL: https://github.com/twbs/bootstrap/pull/30329
Upstream URL: https://github.com/twbs/bootstrap/pull/30740
Own PR: `undefined`
Local PR: I and I am collaborator, I removed the exclude for merged PR from it. You can probably find a PR since you are a collaborator in more repositories then me :). 

3. SCREENSHOT:
image

PR #3694, opened 2020-10-31
Add `unfinished-comments` feature

### LINKED ISSUE #3680: Notify the user of unfinished comments in a tab
How many times do you start writing a comment in a tab, the scroll up, change tab, and forget about it?

We could display this in the title/tab:

cheap-glitch:
Thanks for contributing! 🍄

1. LINKED ISSUES: Closes #3680 

2. TEST URLS:
- https://github.com/sindresorhus/refined-github/issues/3680
- https://github.com/sindresorhus/refined-github/pull/3596 (hidden textareas)

3. SCREENSHOT:
Peek 2020-10-31 18-24

Notes:
- I used the `visibilitychange` event instead of `pageshow`/`pagehide` as the latter weren't working for me in Firefox. As an added bonus, it also works when hiding/minifying the whole browser window.
- I followed the suggestion of the issue for the notification text, but imho something shorter would be better, e.g. `(+)` or `[*]` (akin to what editors show when a file has been modified but not saved).

### DIFF
--- a/source/features/unfinished-comments.tsx
+++ b/source/features/unfinished-comments.tsx
@@ -3,28 +3,23 @@ import * as pageDetect from 'github-url-detection';
 
 import features from '.';
 
-let documentTitle = document.title;
+let documentTitle = '';
+
+async function updateDocumentTitle(): Promise<void> {
+	if (document.visibilityState === 'hidden') {
+		if (select.all('textarea').some(textarea => textarea.value.length > 0 && (textarea.offsetWidth > 0 || textarea.offsetHeight > 0))) {
+			console.log('changing title');
+			documentTitle = document.title;
+			document.title = '(Draft comment) ' + document.title;
+		}
+	} else {
+		document.title = documentTitle;
+	}
+}
 
 function init(): void {
-	if (document.body.dataset.visibilityListenerAdded) {
-		return;
-	}
-
-	const textareas = select.all('textarea');
-
-	document.addEventListener('visibilitychange', async () => {
-		if (document.visibilityState === 'hidden') {
-			if (textareas.some(textarea => (textarea.offsetWidth > 0 || textarea.offsetHeight > 0) && textarea.value.length > 0)) {
-				documentTitle = document.title;
-				document.title = '(Draft comment) ' + document.title;
-			}
-
-			return;
-		}
-
-		document.title = documentTitle;
-	});
-	document.body.dataset.visibilityListenerAdded = String(true);
+	documentTitle = document.title;
+	document.addEventListener('visibilitychange', updateDocumentTitle);
 }
 
 void features.add(__filebasename, {
--- a/source/features/unfinished-comments.tsx
+++ b/source/features/unfinished-comments.tsx
@@ -1,10 +1,9 @@
 import select from 'select-dom';
-import cache from 'webext-storage-cache';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
 
-const cacheKey = __filebasename + ':' + window.location.href;
+let documentTitle = document.title;
 
 function init(): void {
 	if (document.body.dataset.visibilityListenerAdded) {
@@ -16,17 +15,14 @@ function init(): void {
 	document.addEventListener('visibilitychange', async () => {
 		if (document.visibilityState === 'hidden') {
 			if (textareas.some(textarea => (textarea.offsetWidth > 0 || textarea.offsetHeight > 0) && textarea.value.length > 0)) {
-				await cache.set(cacheKey, document.title);
+				documentTitle = document.title;
 				document.title = '(Draft comment) ' + document.title;
 			}
 
 			return;
 		}
 
-		if (await cache.has(cacheKey)) {
-			document.title = String(await cache.get(cacheKey));
-			await cache.delete(cacheKey);
-		}
+		document.title = documentTitle;
 	});
 	document.body.dataset.visibilityListenerAdded = String(true);
 }
--- a/source/features/unfinished-comments.tsx
+++ b/source/features/unfinished-comments.tsx
@@ -3,17 +3,19 @@ import * as pageDetect from 'github-url-detection';
 
 import features from '.';
 
-let documentTitle: string | null = null;
+let documentTitle: string | undefined = undefined;
+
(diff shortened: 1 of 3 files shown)


source/features/unfinished-comments.tsx
```diff
@@ -0,0 +1,39 @@
...
+	if (document.body.dataset.visibilityListenerAdded) {
+		return;
+	}
+
+	const textareas = select.all('textarea');
+
+	document.addEventListener('visibilitychange', async () => {
+		if (document.visibilityState === 'hidden') {
+			if (textareas.some(textarea => (textarea.offsetWidth > 0 || textarea.offsetHeight > 0) && textarea.value.length > 0)) {
+				await cache.set(cacheKey, document.title);
```
fregante, line 19: A global is plenty, no need to literally store it in the cache 😃

source/features/unfinished-comments.tsx
```diff
@@ -0,0 +1,39 @@
...
+const cacheKey = __filebasename + ':' + window.location.href;
+
+function init(): void {
+	if (document.body.dataset.visibilityListenerAdded) {
+		return;
+	}
+
+	const textareas = select.all('textarea');
+
+	document.addEventListener('visibilitychange', async () => {
```
fregante, line 16: This function needs to be moved to the top level so you don’t need to use dataset. There’s no problem with just selecting the textareas every time.

source/features/unfinished-comments.tsx
```diff
@@ -0,0 +1,39 @@
...
+		if (document.visibilityState === 'hidden') {
+			if (textareas.some(textarea => (textarea.offsetWidth > 0 || textarea.offsetHeight > 0) && textarea.value.length > 0)) {
```
fregante, line 18: I’d use the length check first since it’s the lighter part

fregante (changes requested):
Thank you!

cheap-glitch:
What do you think about using a shorter text for the indicator (e.g. `[+]`)? It hides less of the original tab title — but on the other hand it's also less noticeable.

source/features/unfinished-comments.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+async function updateDocumentTitle(): Promise<void> {
+	if (document.visibilityState === 'hidden') {
+		if (select.all('textarea').some(textarea => textarea.value.length > 0 && (textarea.offsetWidth > 0 || textarea.offsetHeight > 0))) {
+			console.log('changing title');
+			documentTitle = document.title;
+			document.title = '(Draft comment) ' + document.title;
+		}
+	} else {
+		document.title = documentTitle;
```
fregante, line 16: We should only change the title if we changed it, and then we should reset it. The title might be changed by GitHub as well, so we don’t want to cache it for long.

source/features/unfinished-comments.tsx
```diff
@@ -0,0 +1,30 @@
...
+import * as pageDetect from 'github-url-detection';
+
+import features from '.';
+
+let documentTitle = '';
+
+async function updateDocumentTitle(): Promise<void> {
+	if (document.visibilityState === 'hidden') {
+		if (select.all('textarea').some(textarea => textarea.value.length > 0 && (textarea.offsetWidth > 0 || textarea.offsetHeight > 0))) {
+			console.log('changing title');
```
fregante, line 11: ?

fregante:
I suggested the extended text because anything else doesn’t give enough information about what’s going on, especially since this isn’t a GitHub feature. Ideally you want to go back to the tab as soon as possible and finish the comment anyway.

source/features/unfinished-comments.tsx
cheap-glitch: This should be fixed now, I think.

source/features/unfinished-comments.tsx
```diff
@@ -0,0 +1,30 @@
...
+			document.title = '(Draft comment) ' + document.title;
+		}
+	} else if (documentTitle) {
+		document.title = documentTitle;
+		documentTitle = null;
+	}
+}
+
+function init(): void {
+	documentTitle = document.title;
```
fregante, line 21:
```suggestion
```

source/features/unfinished-comments.tsx
```diff
@@ -0,0 +1,30 @@
+import select from 'select-dom';
+import * as pageDetect from 'github-url-detection';
+
+import features from '.';
+
+let documentTitle: string | null = null;
+
+async function updateDocumentTitle(): Promise<void> {
+	if (document.visibilityState === 'hidden') {
```
fregante, line 9: If you flip this condition, you can avoid nesting the conditions:
```js
if (documentTitle) { // PR review-only comment: if this exists, we're 'visible'
	document.title = documentTitle;
	documentTitle = undefined; // PR review-only comment: never null https://github.com/sindresorhus/meta/issues/7
} else if (document.visibilityState === 'hidden' && hasDraftComments()) {
	document.title = '(Draft comment) ' + document.title;
}
```

source/features/unfinished-comments.tsx
```diff
@@ -0,0 +1,30 @@
...
+		if (select.all('textarea').some(textarea => textarea.value.length > 0 && (textarea.offsetWidth > 0 || textarea.offsetHeight > 0))) {
```
fregante, line 10: We only ever need to check one dimension:
```suggestion
		if (select.all('textarea').some(textarea => textarea.value.length > 0 && textarea.offsetWidth > 0) {
```

fregante:
It works!

fregante (approved):

fregante:
Test

fregante:
This unfortunately also catches comments that are being sent, for example:

1. Write comment
2. Click to send it
3. Immediately switch tab

When switching the tab, the comment will be detected as "draft" but will be sent not long after. I think that there should be an event we can listen to so that when the comment is submitted, the label is removed.

I wonder how this can be solved easily and comprehensively (`wait-for-ci` for example also will cause this)

cheap-glitch:
> When switching the tab, the comment will be detected as "draft" but will be sent not long after. I think that there should be an event we can listen to so that when the comment is submitted, the label is removed.

We could try listening to the `submit` event of every form on the page.

PR #3231, opened 2020-06-14
Meta: Minor improvements and lint

fregante:
Small fixes

PR #3695, opened 2020-11-01
Meta: Extract and extend `getRepositoryInfo`

fregante:
Follows https://github.com/fregante/github-url-detection/pull/47


source/features/ci-link.tsx
```diff
@@ -5,11 +5,11 @@ import * as pageDetect from 'github-url-detection';
 
 import features from '.';
 import fetchDom from '../helpers/fetch-dom';
-import {getRepoURL} from '../github-helpers';
+import {buildRepoURL} from '../github-helpers';
 
 // Look for the CI icon in the latest 2 days of commits #2990
 const getIcon = onetime(fetchDom.bind(null,
-	`/${getRepoURL()}/commits`, [
+	buildRepoURL('commits'), [
```
fregante, line 12: I'm super annoyed that I missed a bunch of these in my previous PR 🤷‍♂️

source/features/releases-tab.tsx
```diff
@@ -10,9 +10,9 @@ import * as api from '../github-helpers/api';
 import looseParseInt from '../helpers/loose-parse-int';
 import {appendBefore} from '../helpers/dom-utils';
 import {createDropdownItem} from './more-dropdown';
-import {buildRepoURL, getRepoURL, getRepoGQL} from '../github-helpers';
+import {buildRepoURL, getRepoGQL, getRepositoryInfo} from '../github-helpers';
 
-const cacheKey = `releases-count:${getRepoURL()}`;
+const cacheKey = `releases-count:${getRepositoryInfo()!.url}`;
```
fregante, line 15: Most of the changes look like this. Given that `getRepositoryInfo` is now normalized via meta tag, capitalization should not be a problem.

source/features/show-open-prs-of-forks.tsx
```diff
@@ -32,7 +32,7 @@ const countPRs = cache.function(async (forkedRepo: string): Promise<[prCount: nu
 	`);
 
 	// Only show PRs originated from the current repo
-	const prs = search.nodes.filter((pr: AnyObject) => pr.headRepository.nameWithOwner.toLowerCase() === getRepoURL());
+	const prs = search.nodes.filter((pr: AnyObject) => pr.headRepository.nameWithOwner === getRepositoryInfo()!.url);
```
fregante, line 35: Since `getRepositoryInfo` normalizes it, we also don't need to use `toLowerCase` on the source.

Also this makes me think that I should use `.nameWithOwner` instead of `.url`, making it even longer 😅

```diff
- getRepoURL()
+ getRepositoryInfo()!.nameWithOwner
```

source/features/forked-to.tsx
```diff
@@ -12,10 +12,10 @@ import LinkExternalIcon from 'octicon/link-external.svg';
 import features from '.';
 import fetchDom from '../helpers/fetch-dom';
 import GitHubURL from '../github-helpers/github-url';
-import {getRepoURL, getUsername, getForkedRepo} from '../github-helpers';
+import {getUsername, getForkedRepo, getRepositoryInfo} from '../github-helpers';
 
-const getForkSourceRepo = (): string => getForkedRepo() ?? getRepoURL();
-const getCacheKey = (): string => `forked-to:${getUsername()}@${getForkSourceRepo().toLowerCase()}`;
+const getForkSourceRepo = (): string => getForkedRepo() ?? getRepositoryInfo()!.url;
+const getCacheKey = (): string => `forked-to:${getUsername()}@${getForkSourceRepo()}`;
```
fregante, line 18: This feature was one of the affected by the capitalization:

- `getForkedRepo` already reported the canonical capitalization
- `getForkSourceRepo` now _always_ reports the canonical capitalization, so it doesn't need to be lowercased anywhere (it wasn't necessary before either since `getRepoURL` was lowercased already)

source/features/forked-to.tsx
```diff
@@ -12,16 +12,16 @@ import LinkExternalIcon from 'octicon/link-external.svg';
 import features from '.';
 import fetchDom from '../helpers/fetch-dom';
 import GitHubURL from '../github-helpers/github-url';
-import {getRepoURL, getUsername, getForkedRepo} from '../github-helpers';
+import {getUsername, getForkedRepo, getRepositoryInfo} from '../github-helpers';
 
-const getForkSourceRepo = (): string => getForkedRepo() ?? getRepoURL();
-const getCacheKey = (): string => `forked-to:${getUsername()}@${getForkSourceRepo().toLowerCase()}`;
+const getForkSourceRepo = (): string => getForkedRepo() ?? getRepositoryInfo()!.url;
+const getCacheKey = (): string => `forked-to:${getForkSourceRepo()}@${getUsername()}`;
```
fregante, line 18: This feature was one of the affected by the capitalization:

- `getForkedRepo` already reported the canonical capitalization
- `getForkSourceRepo` now _always_ reports the canonical capitalization, so it doesn't need to be lowercased anywhere (it wasn't necessary before either since `getRepoURL` was lowercased already)
- `getCacheKey` has also been updated because the fork list now is not being stored as lowercase anymore, so the cache needs to be invalidated

source/features/show-open-prs-of-forks.tsx
yakov116: If I remember correctly we had that and you did not like it. So you split it.... ( now I cant seem to find that PR 😠 )

source/github-helpers/index.ts
```diff
@@ -39,8 +39,6 @@ export const getCurrentBranch = (): string | undefined => {
 
 export const isFirefox = navigator.userAgent.includes('Firefox/');
 
-export const getRepoURL = (): string => getRepositoryInfo().url!.toLowerCase();
```
fregante, line 42: I'm tempted to preserve this function so that we have a shortcut, but this would then obfuscate the fact that it calls `getRepositoryInfo` and it might therefore be misused in functions that need both `url` and `path`, for example

yakov116: Agree But I like how short and clean it was.

source/features/show-open-prs-of-forks.tsx
yakov116: https://github.com/sindresorhus/refined-github/pull/3231/files

fregante: Where does it say `nameWithOwner`?

source/github-helpers/index.ts
fregante: Went with the shortcut, now this isn't too bad:

```diff
- getRepoURL()
+ getRepo()!.url
```

yakov116: Looks good

PR #3707, opened 2020-11-07
Improve `unfinished-comments` detection logic

cheap-glitch:
Thanks for contributing! 🍄

1. LINKED ISSUES: #3694 (cf. [this comment](https://github.com/sindresorhus/refined-github/pull/3694#issuecomment-722707088))

2. TEST URLS: #3707

3. SCREENSHOT: n/a

### DIFF
--- a/source/features/unfinished-comments.tsx
+++ b/source/features/unfinished-comments.tsx
@@ -8,8 +8,10 @@ let documentTitle: string | undefined;
 
 function hasDraftComments(): boolean {
 	// `[disabled]` excludes the PR description field that `wait-for-build` disables while it waits
-	return select.all('textarea:not([disabled])').some(textarea =>
-		textarea.value.length > 0 && textarea.offsetWidth > 0 && !textarea.closest('.rgh-is-sending-comment')
+	return select.all<HTMLTextAreaElement>('textarea:not([disabled])').some(textarea =>
+		textarea.value !== textarea.textContent && // Exclude comments being edited but not yet changed (and empty comment fields)
+		textarea.offsetWidth > 0 && // Exclude invisible fields
+		!textarea.closest('.rgh-is-sending-comment') // Exclude forms being submitted
 	);
 }
 
@@ -27,7 +29,7 @@ function tagSendingForm({delegateTarget: form}: delegate.Event<Event, HTMLFormEl
 	form.classList.add('rgh-is-sending-comment');
 	form.addEventListener('reset', () => {
 		form.classList.remove('rgh-is-sending-comment');
-	}, {once: true})
+	}, {once: true});
 }
 
 function init(): void {
--- a/source/features/unfinished-comments.tsx
+++ b/source/features/unfinished-comments.tsx
@@ -20,11 +20,13 @@ function updateDocumentTitle(): void {
 		documentTitle = undefined;
 	}
 
-	select('.rgh-is-sending-comment')?.classList.remove('rgh-is-sending-comment');
 }
 
-function tagSendingForm({delegateTarget}: delegate.Event<Event, HTMLElement>): void {
-	delegateTarget.classList.add('rgh-is-sending-comment');
+function tagSendingForm({delegateTarget: form}: delegate.Event<Event, HTMLFormElement>): void {
+	form.classList.add('rgh-is-sending-comment');
+	form.addEventListener('reset', () => {
+			form.classList.remove('rgh-is-sending-comment');
+	}, {once: true})
 }
 
 function init(): void {
--- a/source/features/unfinished-comments.tsx
+++ b/source/features/unfinished-comments.tsx
@@ -18,6 +18,7 @@ function updateDocumentTitle(): void {
 		document.title = documentTitle;
 		documentTitle = undefined;
 	}
+
 	select('.rgh-is-sending-comment')?.classList.remove('rgh-is-sending-comment');
 }
 
--- a/source/features/unfinished-comments.tsx
+++ b/source/features/unfinished-comments.tsx
@@ -7,7 +7,8 @@ import features from '.';
 let documentTitle: string | undefined;
 
 function hasDraftComments(): boolean {
-	return select.all('textarea').some(textarea => textarea.value.length > 0 && textarea.offsetWidth > 0 && !textarea.closest('.rgh-is-sending-comment'));
+	// `[disabled]` excludes the PR description field that `wait-for-build` disables while it waits
+	return select.all('textarea:not([disabled])').some(textarea => textarea.value.length > 0 && textarea.offsetWidth > 0 && !textarea.closest('.rgh-is-sending-comment'));
 }
 
 function updateDocumentTitle(): void {
--- a/source/features/unfinished-comments.tsx
+++ b/source/features/unfinished-comments.tsx
@@ -30,7 +30,7 @@ function init(): void {
 
 void features.add(__filebasename, {
 	include: [
-		pageDetect.hasComments
+		pageDetect.hasRichTextEditor
 	],
 	init
 });
--- a/source/features/unfinished-comments.tsx
+++ b/source/features/unfinished-comments.tsx
@@ -12,12 +12,17 @@ function hasDraftComments(): boolean {
 }
 
 function updateDocumentTitle(): void {
+	resetDocumentTitle();
+	if (document.visibilityState === 'hidden' && hasDraftComments()) {
+		documentTitle = document.title;
+		document.title = '(Draft comment) ' + document.title;
+	}
+}
+
+function resetDocumentTitle(): void {
 	if (documentTitle) {
 		document.title = documentTitle;
 		documentTitle = undefined;
-	} else if (document.visibilityState === 'hidden' && hasDraftComments()) {
-		documentTitle = document.title;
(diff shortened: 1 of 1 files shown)


cheap-glitch:
The feature should now work correctly in both of these cases:
- when a comment is submitted from a hidden tab
- when a comment is submitted from a hidden tab but there are still some drafts in the page

source/features/unfinished-comments.tsx
```diff
@@ -9,18 +10,19 @@ function hasDraftComments(): boolean {
 	return select.all('textarea').some(textarea => textarea.value.length > 0 && textarea.offsetWidth > 0);
 }
 
-async function updateDocumentTitle(): Promise<void> {
-	if (documentTitle) {
-		document.title = documentTitle;
-		documentTitle = undefined;
-	} else if (document.visibilityState === 'hidden' && hasDraftComments()) {
+function updateDocumentTitle(): void {
+	if (document.visibilityState === 'hidden' && hasDraftComments() && !documentTitle) {
```
cheap-glitch, line 14: I know it's ugly, but the nested `if` was here to account for the case where a comment is sent from a hidden tab but there are still drafts somewhere on the page (and thus the title shouldn't be reset).
But maybe this possibility doesn't exist in the first place?

fregante:
I tested both your last commit and my commit, and neither one actually catch the situation described by https://github.com/sindresorhus/refined-github/pull/3694#issuecomment-722707088

`submit` might happen when the user is still on the page, but the page can be updated when it's hidden.

If possible, perhaps, the easier way is to exclude "sending" textareas in `hasDraftComments`, example:

```js
	return select.all('textarea').some(textarea => textarea.value.length > 0 && textarea.offsetWidth > 0 && !textarea.closest('.is-sending')); // made-up class
```

source/features/unfinished-comments.tsx
fregante: You're right, the 2 conditions aren't equivalent when "is hidden, has draft and has a documentTitle". I don't know if it matters at the moment because the change doesn't fix the issue, as mentioned in the other comment.

cheap-glitch:
> If possible, perhaps, the easier way is to exclude "sending" textareas in hasDraftComments

So the solution would be to mark the submitted forms with a class/attribute in the `submit` listener?

fregante:
I expect GitHub to already be doing that, so look for such a class in the dom first.

cheap-glitch:
> I expect GitHub to already be doing that, so look for such a class in the dom first.

Maybe I didn't look in the right place, but I found no such class.

fregante:
> Maybe I didn't look in the right place, but I found no such class.

The fields and forms don't appear to have the class, but what do you think about looking for non-disabled **Comment** buttons instead? This slightly complicates the selector to cover the various types of comments and merge commits descriptions, but we wouldn't have to check the content nor the visibility.

Example: 

a212

cheap-glitch:
> The fields and forms don't appear to have the class, but what do you think about looking for non-disabled Comment buttons instead?

After a few quick tests, I see two problems with this approach:
1. There are a lot of buttons with the same (or very similar) attributes as the `Comment` buttons, and those are not disabled but simply hidden. I'm afraid we would still have to rely on checking visibility, parent elements or text content and that would complicate things quickly.

2. The `Comment` buttons associated with replies to code comments never get disabled (at least on Firefox) even when the corresponding textareas are hidden, blurred and/or empty — except for a very brief moment when the comment is posted.

source/features/unfinished-comments.tsx
```diff
@@ -1,26 +1,34 @@
...
+	select('.rgh-is-sending-comment')?.classList.remove('rgh-is-sending-comment');
+}
+
+function tagSendingForm({delegateTarget}: delegate.Event<Event, HTMLElement>): void {
+	delegateTarget.classList.add('rgh-is-sending-comment');
 }
 
 function init(): void {
 	document.addEventListener('visibilitychange', updateDocumentTitle);
+	delegate(document.body, 'form', 'submit', tagSendingForm);
```
fregante, line 31: This alone won't cover `wait-for-ci`

source/features/unfinished-comments.tsx
```diff
@@ -1,26 +1,34 @@
...
+function updateDocumentTitle(): void {
+	if (document.visibilityState === 'hidden' && hasDraftComments()) {
 		documentTitle = document.title;
 		document.title = '(Draft comment) ' + document.title;
+	} else if (documentTitle) {
+		document.title = documentTitle;
+		documentTitle = undefined;
 	}
+
+	select('.rgh-is-sending-comment')?.classList.remove('rgh-is-sending-comment');
```
fregante, line 22: Unfortunately this also isn't compatible with `wait-for-ci` if the user switches to the tab twice (because the first switch will remove the class)

source/features/unfinished-comments.tsx
```diff
@@ -1,26 +1,34 @@
...
+}
+
+function tagSendingForm({delegateTarget}: delegate.Event<Event, HTMLElement>): void {
+	delegateTarget.classList.add('rgh-is-sending-comment');
```
fregante, line 26: This is the right time to remove the class:

```suggestion
function tagSendingForm({delegateTarget: form}: delegate.Event<Event, HTMLFormElement>): void {
	form.classList.add('rgh-is-sending-comment');
	form.addEventListener('reset', () => {
			form.classList.remove('rgh-is-sending-comment');
	}, {once: true})
```
Untested, likely not supported:

- review comments
- editing comments
- PR submissions

source/features/unfinished-comments.tsx
fregante:
```suggestion
```

fregante (changes requested):
Ok I found a solution for the class removal.

source/features/unfinished-comments.tsx
```diff
@@ -1,26 +1,34 @@
...
+import delegate from 'delegate-it';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
 
 let documentTitle: string | undefined;
 
 function hasDraftComments(): boolean {
-	return select.all('textarea').some(textarea => textarea.value.length > 0 && textarea.offsetWidth > 0);
+	return select.all('textarea').some(textarea => textarea.value.length > 0 && textarea.offsetWidth > 0 && !textarea.closest('.rgh-is-sending-comment'));
```
fregante, line 10:
```suggestion
	return select.all('textarea:not([disabled])').some(textarea => textarea.value.length > 0 && textarea.offsetWidth > 0 && !textarea.closest('.rgh-is-sending-comment'));
```

source/features/unfinished-comments.tsx
fregante: Found a solution for this too. Check the other comment in this review

fregante (changes requested):

source/features/unfinished-comments.tsx
```diff
@@ -1,26 +1,36 @@
...
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
 
 let documentTitle: string | undefined;
 
 function hasDraftComments(): boolean {
-	return select.all('textarea').some(textarea => textarea.value.length > 0 && textarea.offsetWidth > 0);
+	// `[disabled]` excludes the PR description field that `wait-for-build` disables while it waits
+	return select.all('textarea:not([disabled])').some(textarea => textarea.value.length > 0 && textarea.offsetWidth > 0 && !textarea.closest('.rgh-is-sending-comment'));
```
fregante, line 11:
```suggestion
	return select.all('textarea:not([disabled])').some(textarea =>
		textarea.value.length > 0 && textarea.offsetWidth > 0 && !textarea.closest('.rgh-is-sending-comment')
	);
```

source/features/unfinished-comments.tsx
```diff
@@ -1,26 +1,40 @@
...
 
 let documentTitle: string | undefined;
 
 function hasDraftComments(): boolean {
-	return select.all('textarea').some(textarea => textarea.value.length > 0 && textarea.offsetWidth > 0);
+	// `[disabled]` excludes the PR description field that `wait-for-build` disables while it waits
+	return select.all<HTMLTextAreaElement>('textarea:not([disabled])').some(textarea =>
+		textarea.value !== textarea.textContent && // Exclude comments being edited but not yet changed (and empty comment fields)
+		textarea.offsetWidth > 0 && // Exclude invisible fields
+		!textarea.closest('.rgh-is-sending-comment') // Exclude forms being submitted
```
fregante, line 14: I think this would cover:

- regular comments
- review comments
- PR merges

```suggestion
		select.exists('.btn-primary[disabled]', textarea.form!);
```

This reads as: if the form of the current textarea (which is filled, changed, and visible) contains a disabled "submit" button, then the textarea is being sent.


I'm not sure if the selector needs to be tweaked.

cheap-glitch: Tested all three cases and they work great! Although the feature doesn't work when creating a new issue/PR — but I don't think it was the goal in the first place?

fregante: I think that's just due to `pageDetect.hasComments`. Perhaps the correct one is `hasRichTextEditor`

fregante:
Thanks @cheap-glitch!

PR #3708, opened 2020-11-08
Meta: Update release workflow with `ghat`

fregante:
1. Updates the workflow to the latest version as described in https://github.com/notlmn/browser-extension-template/pull/39
2. Starts using [ghat](https://github.com/fregante/ghat) to simplify the sharing/updating of the workflow.

New source: https://github.com/fregante/ghatemplates/blob/master/webext/release.yml

Before merging, ideally:

- [x] https://github.com/fregante/ghat/issues/12


.github/workflows/release.yml
```diff
@@ -0,0 +1,55 @@
...
+name: Release
+on:
+  workflow_dispatch: null
+  schedule:
+    - cron: 42 17 * * 4
+jobs:
+  Version:
+    outputs:
+      created: '${{ steps.daily-version.outputs.created }}'
+      version: '${{ steps.daily-version.outputs.version }}'
```
fregante, line 14: `js-yaml` messes with the template’s quotations, I hope that it doesn't affect GitHub

fregante: Auto-deployment in less than 12 hours. Let's see how it goes.

Issue #2712, opened 2020-01-20
Link to review from the overview at the end of a PR


tooomm:
I miss the possibility to see a link in the summary at the end of a PR to each review and requested changes etc.
I want to click the text `requested changes` or `approved` next to a reviewers name to jump to that review and read it again without searching for it in between all the comments.

Example URL:
Any PR which has a review added. E.g. https://github.com/sindresorhus/refined-github/pull/2667

fregante:
Sounds good, but this is already possible by clicking on the sidebar (which is always visible on the right)

<img width="1024" alt="click to visit" src="https://user-images.githubusercontent.com/1402241/72798224-2ce0ed80-3c75-11ea-8e0a-a0aa93a33498.png">

tooomm:
- 1. The link on the right is not visible if you scroll down for me, as it stick to the top and the widget(s) doesn't follow the page scrolling down.
That is an inconsistent behavior as it works in this issue right now. I always see the tags and notification settings as they keep scrolling down with the content.
But in the linked PR this does not work. I'm using Firefox 72.0.1, no other GitHub addons installed.
- 2. That link brings you to the changes tab. I requested a link to the review in comments itself, e.g: https://github.com/sindresorhus/refined-github/pull/2667#pullrequestreview-341590546

fregante:
> The link on the right is not visible if you scroll down for me,

The sidebar is sticky in Refined GitHub https://github.com/sindresorhus/refined-github/blob/master/source/features/sticky-discussion-sidebar.tsx

>     1. That link brings you to the changes in files tab. I requested a link to the review in comments itself,

👍

tooomm:
>The sidebar is sticky in Refined GitHub

As I said, it does work in this issue for example...
But it does NOT work for whatever reason in the PR I linked and others as well. Seems to not work for any PRs for me. Some other issue posts I tried don't work too.

Refined Github 20.1.14.1820, Firefox 72.0.1

Any idea for troubleshooting?

fregante:
@tooomm the sidebar isn't sticky if it doesn't fit your screen vertically. Try pressing `ctrl -` to _zoom out_ the page until it definitely fits, and then try scrolling. If it still doesn't work, please record a gif and open a new issue

tooomm:
>If it still doesn't work, please record a gif and open a new issue

Done. #2770

<br>

Back to topic then:
**This issue request a direct link to the review comment at the bottom of a PR in the list of reviews.**

[...]

PR #3710, opened 2020-11-09
Add `jump-to-change-requested-comment` feature

xtianproject:
Thanks for contributing! 🍄

1. LINKED ISSUES:
	Closes [#2712](https://github.com/sindresorhus/refined-github/issues/2712)

2. TEST URLS:
	[#17968](https://github.com/facebook/react/pull/17968#partial-pull-merging)

3. SCREENSHOT:
    
link-to-requested-change-comment

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -158,6 +158,8 @@ Thanks for contributing! 🦋🙌
 - [](# "open-issue-to-latest-comment") [Makes the "comment" icon in issue lists link to the latest comment of the issue.](https://user-images.githubusercontent.com/14323370/57962709-7019de00-78e8-11e9-8398-7e617ba7a96f.png)
 - [](# "expand-all-hidden-comments") [On long conversations where GitHub hides comments under a "Load more...", alt-clicking it will load up to 200 comments at once instead of 60.](https://user-images.githubusercontent.com/1402241/73838332-0c548e00-4846-11ea-935f-28d728b30ae9.png)
 - [](# "keyboard-navigation") [Adds shortcuts to conversations and PR file lists: <kbd>j</kbd> focuses the comment/file below; <kbd>k</kbd> focuses the comment/file above.](https://user-images.githubusercontent.com/1402241/86573176-48665900-bf74-11ea-8996-a5c46cb7bdfd.gif)
+- [](# "keyboard-navigation") [Adds shortcuts to conversations and PR file lists: <kbd>j</kbd> focuses the comment/file below; <kbd>k</kbd> focuses the comment/file above.](https://user-images.githubusercontent.com/1402241/86573176-48665900-bf74-11ea-8996-a5c46cb7bdfd.gif)
+- [](# "link-to-requested-change-comment") [Adds a link to the requested change comment.](https://user-images.githubusercontent.com/19198931/98581317-45052480-22fc-11eb-8a79-0c0fe812a81a.gif)
 
 <!-- Refer to style guide above. Keep this message between sections. -->
 
--- a/readme.md
+++ b/readme.md
@@ -158,7 +158,7 @@ Thanks for contributing! 🦋🙌
 - [](# "open-issue-to-latest-comment") [Makes the "comment" icon in issue lists link to the latest comment of the issue.](https://user-images.githubusercontent.com/14323370/57962709-7019de00-78e8-11e9-8398-7e617ba7a96f.png)
 - [](# "expand-all-hidden-comments") [On long conversations where GitHub hides comments under a "Load more...", alt-clicking it will load up to 200 comments at once instead of 60.](https://user-images.githubusercontent.com/1402241/73838332-0c548e00-4846-11ea-935f-28d728b30ae9.png)
 - [](# "keyboard-navigation") [Adds shortcuts to conversations and PR file lists: <kbd>j</kbd> focuses the comment/file below; <kbd>k</kbd> focuses the comment/file above.](https://user-images.githubusercontent.com/1402241/86573176-48665900-bf74-11ea-8996-a5c46cb7bdfd.gif)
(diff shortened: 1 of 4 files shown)


fregante:
Thanks for the PR! That area of the page is updated via ajax, so it needs to use `selector-observer`. Search this in other features to see how it’s used

xtianproject:
I updated the feature to use `selector-observer`, let me know if it is correct.

source/features/link-to-requested-change-comment.tsx
```diff
@@ -0,0 +1,27 @@
...
+import * as pageDetect from 'github-url-detection';
+
+import {wrap} from '../helpers/dom-utils';
+import features from '.';
+
+function init(): void {
+	observe('div.js-merge-message-container', {
+		constructor: HTMLDivElement,
+		add(messageContainer) {
+			for (const reviewItem of select.all('.merge-status-item.review-item', messageContainer)) {
```
fregante, line 14: `observe` should select the final elements and should almost never contain a loop, since it can just call the `add` method on each element found.

The `constructor` is not necessary unless it adds require type information (e.g. `HTMLAnchorElement`)

```suggestion
	observe('merge-status-item.review-item', {
		add(reviewItem) {
```

fregante: But this selection is too broad: `.review-item` even includes the whole dropdown and other links. I suggest:

1. changing this selector to select the dropdown link (the selector should end with `[href^="#pullrequestreview-"]`)
2. finding the container with something like `.closest('.review-status-item').querySelector('[title^="these changes"]')`
3. wrapping the specific "requested changes" text node, with `container.lastChild`

readme.md
```diff
@@ -158,6 +158,8 @@ Thanks for contributing! 🦋🙌
 - [](# "open-issue-to-latest-comment") [Makes the "comment" icon in issue lists link to the latest comment of the issue.](https://user-images.githubusercontent.com/14323370/57962709-7019de00-78e8-11e9-8398-7e617ba7a96f.png)
 - [](# "expand-all-hidden-comments") [On long conversations where GitHub hides comments under a "Load more...", alt-clicking it will load up to 200 comments at once instead of 60.](https://user-images.githubusercontent.com/1402241/73838332-0c548e00-4846-11ea-935f-28d728b30ae9.png)
 - [](# "keyboard-navigation") [Adds shortcuts to conversations and PR file lists: <kbd>j</kbd> focuses the comment/file below; <kbd>k</kbd> focuses the comment/file above.](https://user-images.githubusercontent.com/1402241/86573176-48665900-bf74-11ea-8996-a5c46cb7bdfd.gif)
+- [](# "keyboard-navigation") [Adds shortcuts to conversations and PR file lists: <kbd>j</kbd> focuses the comment/file below; <kbd>k</kbd> focuses the comment/file above.](https://user-images.githubusercontent.com/1402241/86573176-48665900-bf74-11ea-8996-a5c46cb7bdfd.gif)
```
fregante, line 161: Duplicate

fregante (changes requested):

source/features/link-to-requested-change-comment.tsx
xtianproject: I've updated my implementation similar to your suggestion. Please review it again.

source/features/link-to-requested-change-comment.tsx
```diff
@@ -0,0 +1,28 @@
...
+import select from 'select-dom';
+import onetime from 'onetime';
+import {observe} from 'selector-observer';
+import * as pageDetect from 'github-url-detection';
+
+import {wrap} from '../helpers/dom-utils';
+import features from '.';
+
+function init(): void {
+	observe('.review-item .dropdown-item[href^="#pullrequestreview-"]', {
```
fregante, line 11: We also need to add a `.rgh` class to mark the element as "processed". Refer to the other features using selector-observer

source/features/link-to-requested-change-comment.tsx
```diff
@@ -0,0 +1,28 @@
...
+		constructor: HTMLAnchorElement,
+		add(messageContainer) {
+			const element = select('.review-status-item div[title*="requested changes"]')?.lastChild;
+
+			if (element) {
+				wrap(element, <a href={messageContainer.href}/>);
+			}
```
fregante, line 18: Conditions like this makes any future errors silent, however this element is expected to be "not found" when GitHub starts hiding some comments on long discussions.

It's not ideal, but we don't have a reasonable solution here.

source/features/link-to-requested-change-comment.tsx
fregante: Seems good! Thanks

readme.md
```diff
@@ -158,6 +158,7 @@ Thanks for contributing! 🦋🙌
 - [](# "open-issue-to-latest-comment") [Makes the "comment" icon in issue lists link to the latest comment of the issue.](https://user-images.githubusercontent.com/14323370/57962709-7019de00-78e8-11e9-8398-7e617ba7a96f.png)
 - [](# "expand-all-hidden-comments") [On long conversations where GitHub hides comments under a "Load more...", alt-clicking it will load up to 200 comments at once instead of 60.](https://user-images.githubusercontent.com/1402241/73838332-0c548e00-4846-11ea-935f-28d728b30ae9.png)
 - [](# "keyboard-navigation") [Adds shortcuts to conversations and PR file lists: <kbd>j</kbd> focuses the comment/file below; <kbd>k</kbd> focuses the comment/file above.](https://user-images.githubusercontent.com/1402241/86573176-48665900-bf74-11ea-8996-a5c46cb7bdfd.gif)
+- [](# "link-to-requested-change-comment") [Adds a link to the requested change comment.](https://user-images.githubusercontent.com/19198931/98718312-418b9f00-23c9-11eb-8da2-dfb616e95eb6.gif)
```
yakov116, line 161: What about `jump-to-change-requested-comment`?

fregante (approved):
# 🥳

xtianproject:
> # 🥳

Thank you! 🥳

Issue #3702, opened 2020-11-06
`recently-pushed-branches-enhancements` stopped working


yakov116:
![image](https://user-images.githubusercontent.com/16872793/98310614-6e8c1c00-1f9b-11eb-8c03-fa0a354612e0.png)

PR #3721, opened 2020-11-13
Fix `recently-pushed-branches-enhancements`

yakov116:
1. LINKED ISSUES:
   Fixes https://github.com/sindresorhus/refined-github/issues/3702

Reverts the change made in #3596 that was incorrect since it does not start with `https://`


fregante (approved):
🙌

PR #3723, opened 2020-11-13
Lint

yakov116:
- Drop `getRepoGQL`
- November Lint
- Sort From
- Drop `textContent` when calling `loosePartIn`


fregante (approved):
As usual let’s keep this open a little longer

yakov116:
Have been under the weather the past 2 weeks so I haven't been linting. But yeah leave this open until December lint.

yakov116:
@fregante I think `pageDetect.isRepo` is not working sincew `recently-pushed-branches` trued running on github.com

will debug tomorrow

source/github-helpers/api.ts
```diff
@@ -161,6 +161,8 @@ export const v4 = mem(async (
 		throw new TypeError('`query` should only be what’s inside \'query {...}\', like \'user(login: "foo") { name }\', but is \n' + query);
 	}
 
+	query = query.replace('repository() {', `repository(owner: "${getRepo()!.owner}", name: "${getRepo()!.name}") {`);
```
fregante, line 164: This will break because getRepo will be called on non-repo pages

fregante: Perhaps use:

``` 
repository = getRepo
if repository
  replace
```

yakov116: I went with checking if the string exists since the only place those calls can be made is where it exists.

fregante: Or you can use `query.replace(string, () => 'blah blah')` and the function will only be called if necessary

fregante:
`lintideas`

- do we have other features that need this update? Maybe other shortcut-only features. https://github.com/sindresorhus/refined-github/pull/3725#discussion_r523828791
- since they're dropping more specific classes, we'll have to start using `select()!.closest()` instead of fragile selectors (unless they're for `delegate` and `delegate` can't be changed to the above)

yakov116:
> * do we have other features that need this update? Maybe other shortcut-only features.

I'll do some looking

source/features/recently-pushed-branches-enhancements.tsx
```diff
@@ -7,26 +7,26 @@ import * as pageDetect from 'github-url-detection';
 import features from '.';
 import {getRepo} from '../github-helpers';
 
-const fragmentURL = `/${getRepo()!.nameWithOwner}/show_partial?partial=tree%2Frecently_touched_branches_list`;
-const selector = `[data-url='${fragmentURL}' i], [src='${fragmentURL}' i]`;
+const fragmentURL = (): string => `/${getRepo()!.nameWithOwner}/show_partial?partial=tree%2Frecently_touched_branches_list`;
+const selector = (): string => `[data-url='${fragmentURL()}' i], [src='${fragmentURL()}' i]`;
```
yakov116, line 11: Should I put this in its own PR? its breaking all non repo features. Or Do you think we will merge by Thursday?

fregante: Yeah you/I can merge it by Wednesday

fregante: But actually, if you have time, there are 2 possible improvements:

- if only one of `data-src` and `src` are necessary, drop the other one
- merge the 2 init functions and avoid these globals altogether. You can replicate the `onetime` part with a plain global boolean.

yakov116: Will git it a shot

yakov116: > * if only one of `data-src` and `src` are necessary, drop the other one

f40da70ed947dfdde79475bdee9b93f6848e9202

I am not sure what you were fixing here. Remeber?

fregante: The selector change was not introduced by that commit, but the commit did separate the comment from the code it commented. 

That comment explains why there are 2 selectors; it should be reinstated and probably the selector won’t have to be changed.

fregante:
`lintidea`

`looseParseInt` now automatically uses `.textContent` if you pass an element. We can change all of its uses as:

```diff
- looseParseInt(element.textContent!)
+ looseParseInt(element)
```

fregante:
> I think `pageDetect.isRepo` is not working sincew `recently-pushed-branches` trued running on github.com

That's because that code was running outside `init`. 

Screen Shot 2020-11-16 at 18 38 50

`github-url-detection` is thoroughly cross-tested so the URL-based detections can't fail

yakov116:
Sorry for not updating my comment, I figured that out later on, you are correct.

fregante:
No worries, that was my thought too. I was just clarifying that it would be very unlikely.

source/features/highlight-non-default-base-branch.tsx
```diff
@@ -6,7 +6,7 @@ import PullRequestIcon from 'octicon/git-pull-request.svg';
 import features from '.';
 import * as api from '../github-helpers/api';
 import getDefaultBranch from '../github-helpers/get-default-branch';
-import {getRepoGQL, buildRepoURL} from '../github-helpers';
+import {buildRepoURL} from '../github-helpers';
```
yakov116, line 9: Note to self fix sort

source/features/pinned-issues-update-time.tsx
```diff
@@ -6,15 +6,15 @@ import * as pageDetect from 'github-url-detection';
 import features from '.';
 import * as api from '../github-helpers/api';
 import looseParseInt from '../helpers/loose-parse-int';
-import {getRepoGQL, getRepo} from '../github-helpers';
+import {getRepo} from '../github-helpers';
```
yakov116, line 9: And here

source/features/recently-pushed-branches-enhancements.tsx
```diff
@@ -7,7 +7,7 @@ import * as pageDetect from 'github-url-detection';
 import features from '.';
 import {getRepo} from '../github-helpers';
 
-const fragmentURL = `/${getRepo()!.nameWithOwner}/show_partial?partial=tree%2Frecently_touched_branches_list`;
+const fragmentURL = `/${getRepo()?.nameWithOwner}/show_partial?partial=tree%2Frecently_touched_branches_list`;
```
fregante, line 10: This should hide the error for the time being. I had thought about functions as well but they looked like a bad patch. Let's merge the 2 `init`s in the next lint/PR

fregante (approved):

fregante:
lintideas

Feel free to check these boxes if/when you do them:

- [ ] Replace hardcoded colors and variables with GitHub’s official variables

PR #3724, opened 2020-11-14
Avoid duplicate upload button in 'rgh-minimize-upload-bar'

### LINKED ISSUE #3714: Reuse GitHub’s native upload button in `minimize-upload-bar`
GitHub added an upload button, but only on mobile.

We should drop ours from `minimize-upload-bar` and resurface the native one instead, if possible.

KennethTsang9286:
1. LINKED ISSUES:
   Closes #3714 

2. TEST URLS:
   https://github.com/sindresorhus/refined-github/pull/3724

3. SCREENSHOT:
edit: update screenshots
image for readme:
Screen Shot 2020-11-14 at 4 03 56 pm


fregante:
Thanks for the PR! 

Both buttons do the same thing, so our button should also be removed. 

Additionally, it should be part of the same feature, not a new one

KennethTsang9286:
@fregante sorry for misunderstanding. updated.

fregante:
Much better, thanks!

From what I see, this feature can be simplified further:
1. The js file should only add a class on body in the init function
2. The CSS file might need to be updated to expect the class on the body element and then include `form` in each selector (but ideally a more specific class than that, considering that this applies to a specific widget and not just any form)

fregante:
About 2: the previous selectors probably don’t need to be updated at all

KennethTsang9286:
1. thx. I missed that.
2. I find a class name candidate, `js-previewable-comment-form`. I have checked "new PR comments", "edit PR comments", "inline-code comments" and "review comments". They seem to be ok. That being said, I don't know all the forms affected by the previous selector so I may have missed a few. So, I cant ensure the new behavior will match to old one. Is it ok?

fregante:
Our class needs to be added on body. 

Then that form class needs to be used to select the form in your new selector. 

The number of forms changes on the page, so the loop will not include all of them.

KennethTsang9286:
@fregante Thank you for reviewing. I have tried to address the problem. Can you check if I have understood correctly?

fregante:
Looks about right!

Issue #3720, opened 2020-11-13
`profile-hotkey` is broken since the dropdown is lazy loaded


yakov116:
Should we not attach it to the button and just make our own shortcut?

yakov116:
Also breaking `deprioritize-marketplace-link`

Maybe we should force load it?

fregante:
The second one can just wait for it to load, maybe on opening as usual. No need to force it, either way

PR #3725, opened 2020-11-15
Fix `deprioritize-marketplace-link`

yakov116:
1. LINKED ISSUES:
   https://github.com/sindresorhus/refined-github/issues/3720#issuecomment-726810747

2. TEST URLS:
  all over

3. SCREENSHOT:
  None

### DIFF
--- a/source/features/deprioritize-marketplace-link.tsx
+++ b/source/features/deprioritize-marketplace-link.tsx
@@ -16,9 +16,9 @@ function handleMenuOpening({delegateTarget: dropdown}: delegate.Event): void {
 }
 
 async function init(): Promise<void> {
-	const marketPlaceLink = await elementReady('.Header-link[href="/marketplace"]');
+	(await elementReady('.Header-link[href="/marketplace"]'))
 	// The Marketplace link seems to have an additional wrapper that other links don't have https://i.imgur.com/KV9rtSq.png
-	marketPlaceLink?.closest('.border-top, .mr-3')!.remove();
+		?.closest('.border-top, .mr-3')!.remove();
 
 	delegate(document, 'header div.Header-item [data-feature-preview-indicator-src]:not(.rgh-marketplace-link)', 'toggle', handleMenuOpening, true);
 }
--- a/source/features/deprioritize-marketplace-link.tsx
+++ b/source/features/deprioritize-marketplace-link.tsx
@@ -1,25 +1,26 @@
 import React from 'dom-chef';
 import select from 'select-dom';
 import onetime from 'onetime';
-import domLoaded from 'dom-loaded';
+import delegate from 'delegate-it';
 import elementReady from 'element-ready';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
 
-async function init(): Promise<void> {
-	const marketPlaceLink = (await elementReady('.Header-link[href="/marketplace"]'));
-	if (marketPlaceLink) {
-		// The Marketplace link seems to have an additional wrapper that other links don't have https://i.imgur.com/KV9rtSq.png
-		marketPlaceLink.closest('.border-top, .mr-3')!.remove();
-	}
-
-	await domLoaded;
-
-	select.last('.header-nav-current-user ~ .dropdown-divider')!.before(
+function handleMenuOpening({delegateTarget: dropdown}: delegate.Event): void {
+	dropdown.classList.add('rgh-marketplace-link'); // Mark this as processed
+	select.last('.dropdown-divider', dropdown)!.before(
 		<div className="dropdown-divider"/>,
 		<a className="dropdown-item" href="/marketplace">Marketplace</a>
 	);
+}
+
+async function init(): Promise<void> {
+	const marketPlaceLink = await elementReady('.Header-link[href="/marketplace"]');
+	// The Marketplace link seems to have an additional wrapper that other links don't have https://i.imgur.com/KV9rtSq.png
+	marketPlaceLink?.closest('.border-top, .mr-3')!.remove();
+
+	delegate(document, 'header div.Header-item [data-feature-preview-indicator-src]:not(.rgh-marketplace-link)', 'toggle', handleMenuOpening, true);
 }
 
 void features.add(__filebasename, {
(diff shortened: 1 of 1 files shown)


source/features/deprioritize-marketplace-link.tsx
```diff
@@ -1,27 +1,28 @@
...
+	dropdown.classList.add('rgh-marketplace-link'); // Mark this as processed
+	select.last('.dropdown-divider', dropdown)!.before(
 		<div className="dropdown-divider"/>,
 		<a className="dropdown-item" href="/marketplace">Marketplace</a>
 	);
 }
 
+async function init(): Promise<void> {
+	(await elementReady('.Header-link[href="/marketplace"]'))
+	// The Marketplace link seems to have an additional wrapper that other links don't have https://i.imgur.com/KV9rtSq.png
```
fregante, line 20: The comment indentation should match the following line. Other than this, it looks ok. Untested

yakov116:
@busches can you double check this on enterprise?

source/features/deprioritize-marketplace-link.tsx
```diff
@@ -1,27 +1,28 @@
...
 		<a className="dropdown-item" href="/marketplace">Marketplace</a>
 	);
 }
 
+async function init(): Promise<void> {
+	(await elementReady('.Header-link[href="/marketplace"]'))
+		// The Marketplace link seems to have an additional wrapper that other links don't have https://i.imgur.com/KV9rtSq.png
+		?.closest('.border-top, .mr-3')!.remove();
+
+	delegate(document, 'header div.Header-item [data-feature-preview-indicator-src]:not(.rgh-marketplace-link)', 'toggle', handleMenuOpening, true);
```
fregante, line 23: Since the header doesn't change on ajax, we can target it directly with
```suggestion
	(await elementReady('[aria-label="View profile and more"]'))
		?.closest('details')!
		.addEventListener('toggle', handleMenuOpening, {once: true});
```

And no `rgh-marketplace-link` class necessary

yakov116: Nice!

source/features/deprioritize-marketplace-link.tsx
```diff
@@ -1,27 +1,28 @@
 import React from 'dom-chef';
 import select from 'select-dom';
 import onetime from 'onetime';
-import domLoaded from 'dom-loaded';
+import delegate from 'delegate-it';
```
yakov116, line 4:
```suggestion
```

yakov116:
Oh my... that applied bad. 🤣

source/features/deprioritize-marketplace-link.tsx
```diff
@@ -1,27 +1,29 @@
...
 	);
 }
 
+async function init(): Promise<void> {
+	(await elementReady('.Header-link[href="/marketplace"]'))
+		// The Marketplace link seems to have an additional wrapper that other links don't have https://i.imgur.com/KV9rtSq.png
+		?.closest('.border-top, .mr-3')!.remove();
+
+	(await elementReady('[aria-label="View profile and more"]'))
+		?.closest('details')!
```
fregante, line 23: Actually, is the `?.` needed in either one? If Enterprise doesn't have the link, it should just be excluded

yakov116: I am not sure.

I think it was added because some pages, like the enter password dont have the links. I dont know if its still relevant

fregante: > like the enter password dont have the links

Let's find out because I don't care about password pages. They have nothing so, if anything, RGH should not run there at all.

yakov116: 51c8d6699afb8540304412df194ec262c503ab23

fregante: So the code should be updated to use a regular if:

If there’s a marketplace link, move it to the dropdown.

fregante: With a comment

busches:
@yakov116 This is not a thing on Enterprise, keeping the `if` like was there before will be fine. I would suggest that we disable this on Enterprise or only add the link if we actual move it. Right now we always add the link even if one didn't exist before.

source/features/deprioritize-marketplace-link.tsx
```diff
@@ -1,27 +1,28 @@
...
+		.addEventListener('toggle', handleMenuOpening, {once: true});
```
fregante, line 23:
```suggestion
	const marketplaceLink = await elementReady('.Header-link[href="/marketplace"]');
	if (marketplaceLink)
		// The link seems to have an additional wrapper that other links don't have https://i.imgur.com/KV9rtSq.png
		marketplaceLink.closest('.border-top, .mr-3')!.remove();

		(await elementReady('[aria-label="View profile and more"]'))!
			.closest('details')!
			.addEventListener('toggle', handleMenuOpening, {once: true});
	}
```

source/features/deprioritize-marketplace-link.tsx
```diff
@@ -1,27 +1,28 @@
...
-		// The Marketplace link seems to have an additional wrapper that other links don't have https://i.imgur.com/KV9rtSq.png
-		marketPlaceLink.closest('.border-top, .mr-3')!.remove();
-	}
-
-	await domLoaded;
-
+function handleMenuOpening(): void {
 	select.last('.header-nav-current-user ~ .dropdown-divider')!.before(
 		<div className="dropdown-divider"/>,
 		<a className="dropdown-item" href="/marketplace">Marketplace</a>
```
fregante, line 12: Does GHE link here or always to github.com? If github.com, let's use this:
```suggestion
		<a className="dropdown-item" href="https://github.com/marketplace">Marketplace</a>
```

busches: Not sure, I've not seen it activated in the wild, just sounds possible from the release notes.

fregante: Who knows then. The selector only specifically selects local links (`.Header-link[href="/marketplace"]`) so if this change is required, the selector also needs to be changed.

Since nobody complained, I'll leave it as is.

fregante (approved):
If the current implementation works for you, it can be merged

PR #3747, opened 2020-11-20
Fix `close-out-of-view-modals` bug that breaks Watch dropdown

### LINKED ISSUE #3742: `close-out-of-view-modals` Breaks the new watch settings
>  Make sure the bug is caused by Refined GitHub. Try disabling the extension first.

It is caused by Refined GitHub and disabling the extension fixes the issue.

> Include a full URL where the bug appears.

On any repository

> Include a screenshot/gif

ezgif com-gif-maker

fregante:
Fix #3742

This feature is so fickle because GitHub keeps changing how dropdown works. This time there are 2 elements and both need to be tracked continously.

If it breaks the site again, we drop it.


source/features/close-out-of-view-modals.tsx
```diff
@@ -1,29 +1,58 @@
...
-		target.closest('details')!.open = false;
+const visible = new Set();
+const observer = new IntersectionObserver(entries => {
+	let lastModal: Element;
+	for (const {intersectionRatio, target: modal} of entries) {
+		if (intersectionRatio > 0) {
+			visible.add(modal);
+		} else {
+			visible.delete(modal);
+		}
```
fregante, line 15: Too bad that `IntersectionObserver` doesn't keep a list of entries available at all times, so we have to track what's visible manually.

source/features/close-out-of-view-modals.tsx
```diff
@@ -1,29 +1,58 @@
...
+		features.error(__filebasename, 'Modal was closed too quickly. Disabling feature');
+		return;
+	}
+
+	lastOpen = Date.now();
+
+	const modals = select.all([
+		':scope > details-menu', // "Watch repo" dropdown
+		':scope > details-dialog', // "Watch repo" dropdown
+		':scope > div > .dropdown-menu' // "Clone or download" and "Repo nav overflow"
```
fregante, line 43: This PR also reintroduces the feature for other dropdowns. Wish me luck

source/features/close-out-of-view-modals.tsx
```diff
@@ -1,29 +1,58 @@
...
-	if (modalBox.getBoundingClientRect().width === 0) {
-		features.error(__filebasename, 'Modal element was not correctly detected for', details);
-	} else {
-		observer.observe(modalBox);
+	const details = event.target as HTMLDetailsElement;
+
+	// Safety check #3742
+	if (!details.open && lastOpen > Date.now() - 1000) {
+		delegation!.destroy();
+		features.error(__filebasename, 'Modal was closed too quickly. Disabling feature');
```
fregante, line 34: The drawback is that if the user is _Flash_, they will see this error in the console. 🤷‍♂️

PR #3749, opened 2020-11-20
Fully restore `set-default-repositories-type-to-sources`

### LINKED ISSUE #3733: `set-default-repositories-type-to-sources` not working
<s>

1. Click my username
2. Visit the "Repositories" tab on my profile

or
</s>

1. Click your "user dropdown" (top-right corner of every page)
2. Click "Your repositories"

The URL should be

```
https://github.com/fregante?tab=projects&type=source
```

But is

```
https://github.com/fregante?tab=projects
```

cheap-glitch:
Thanks for contributing! 🍄

1. LINKED ISSUES: Fixes #3733 

2. TEST URLS:
    1. Click your "user dropdown" (top-right corner of every page)
    2. Click "Your repositories"

Should we also take the opportunity to drop the selectors used for the old layout?


source/features/set-default-repositories-type-to-sources.tsx
```diff
@@ -1,25 +1,35 @@
...
 	]);
 
 	for (const link of links) {
-		const search = new URLSearchParams(link.search);
-		search.set('type', 'source');
-		link.search = String(search);
+		addSourceTypeToLink(link);
 	}
+
+	observe('.header-nav-current-user ~ [href$="tab=repositories"]', { // "Your repositories" in header dropdown
```
fregante, line 27: We should copy this instead:

https://github.com/sindresorhus/refined-github/blob/1bedcd71886ac9c20bf91a6d48e428622ab6d1b1/source/features/deprioritize-marketplace-link.tsx#L22-L24

`observe` is too heavy for a single change that might never happen

cheap-glitch:
Replaced `observe` with `elementReady` as requested. Also, to reiterate:
> Should we also take the opportunity to drop the selectors used for the old layout?

(or maybe it should be done in a separate PR?)

fregante:
> Should we also take the opportunity to drop the selectors used for the old layout?

The current PR code is fine. We only need to drop the old code when it requires active support, which usually doesn't since we kept them separate (or it's just a selector like in this case).

fregante:
Remember to use `Fixes #123456789` or else the original issue you're fixing will remain open. I updated your post

source/features/set-default-repositories-type-to-sources.tsx
```diff
@@ -1,25 +1,36 @@
...
 	for (const link of links) {
-		const search = new URLSearchParams(link.search);
-		search.set('type', 'source');
-		link.search = String(search);
+		addSourceTypeToLink(link);
 	}
+
+	(await elementReady('[aria-label="View profile and more"]'))! // "Your repositories" in header dropdown
+		.closest('details')!
+		.addEventListener('toggle', headerDropdownListener, {once: true});
```
fregante, line 33: Apparently this doesn't work sometimes, for both the features that use it. This happens when the user clicks the dropdown _before_ GitHub loads the dropdown content, so we need to listen for the content load instead.

This is easy-ish since we do in in other places: watch for the `load` event on its `include-fragment`.

If it's not already, it might be worth standardizing this "add listener function". Maybe something like:

```ts
onFragmentLoad(element: Element /*or whatever include-fragment is*/, handler: EventListener): void
```
or

```ts
onFragmentLoad(element: Element /*or whatever include-fragment is*/): Promise<EventListener>
```
cc @yakov116

yakov116: observeOneMutation but yeah this should be maid as its own function since not always will we need to wait for it to load since it may be there already.

fregante: > observeOneMutation

No, it's all event-based: https://github.com/sindresorhus/refined-github/blob/4a064058d5889a5052de0abdfab089c6894d7266/source/github-events/on-new-comments.ts#L55

cheap-glitch: > This file is almost exactly what we need, except it's specific to PR files rather than generic to `include-fragment`

So the idea would be to make `onPrFileLoad()` more generic and use it in this feature?

cheap-glitch:
I went ahead and made a more generic version of https://github.com/sindresorhus/refined-github/blob/f9d8b7a5a402e19b0cb16fd937e0f19c43a6e699/source/github-events/on-pr-file-load.ts named `on-fragment-load.ts` to test in the feature. It works great, although I'm not too happy with the selector (the targeted `include-fragment` element has no classes attached to it afaik).

Next step I guess would be to remove `on-pr-file-load` and use the more generic version in `show-whitespace` and `hidden-review-comments-indicator`?

source/features/set-default-repositories-type-to-sources.tsx
```diff
@@ -1,25 +1,34 @@
...
 	]);
 
 	for (const link of links) {
-		const search = new URLSearchParams(link.search);
-		search.set('type', 'source');
-		link.search = String(search);
+		addSourceTypeToLink(link);
 	}
+
+	onFragmentLoad('.Header-item > details include-fragment', headerDropdownListener); // "Your repositories" in header dropdown
```
fregante, line 31: I made the wrong suggestion in this case since this fragment only ever loads **once**, unlike the "PR file load" listener.

We only need a promise in this case:

```ts
// If the dropdown exists, await it; If not, it probably already loaded
async function onProfileDropdownLoad(): Promise<void> {
	const dropdown = select('.Header details-menu[src^="/users/"] include-fragment');
	if (dropdown) {
		await oneEvent(dropdown, 'load');
	}
}
```
```suggestion
	// "Your repositories" in header dropdown
	await onProfileDropdownLoad();
	addSourceTypeToLink(select<HTMLAnchorElement>('.header-nav-current-user ~ [href$="tab=repositories"]')!);
```

The good part in this case is that if the code runs _after_ the fragment has been loaded, it still works.

Possible improvements **in a new PR**

- add `on-fragment-load.ts` (you can remove it from this PR)
- import it inside `on-pr-file-load` and use it inside
- look for other features that expect/await `include-fragment` and see whether they can use either `on-fragment-load` or `onProfileDropdownLoad`

fregante (approved):
Untested but LGTM

source/features/set-default-repositories-type-to-sources.tsx
```diff
@@ -1,25 +1,40 @@
...
+	const search = new URLSearchParams(link.search);
+	search.set('type', 'source');
+	link.search = String(search);
+}
+
+// If the dropdown exists, await it; If not, it probably already loaded
+async function onProfileDropdownLoad(): Promise<void> {
+	const dropdown = select('.Header details-menu[src^="/users/"] include-fragment');
+	if (dropdown) {
+		await oneEvent(dropdown, 'load');
```
fregante, line 16: It doesn't seem to be enough for me. Maybe `load` is triggered a little before the content is added

```
refined-github.js:2891 Uncaught TypeError: Cannot read property 'before' of undefined
    at HTMLDetailsElement.handleMenuOpening (refined-github.js:2891)
```

Screen Shot 2020-12-14 at 20 41 01

fregante (changes requested):

source/features/set-default-repositories-type-to-sources.tsx
cheap-glitch: Should we use a `MutationObserver`?

fregante: You can try `one-mutation` with a filter if there’s no other way. 

Does `await Promise.resolve()` fix it? It helps using the debugger with dom breakpoints to see when the elements are appended

cheap-glitch: `one-mutation` works fine, we don't even need a filter or to watch the subtree since all the content is loaded at once. Tested on Firefox & Chromium.

fregante:
I’d still like it to be its own separate function so it can be reused

fregante (approved):
Untested but LGTM 😃

fregante:
Now it also fixes `deprioritize-marketplace-link`

fregante:
I opened a couple of issues related to this PR: https://github.com/sindresorhus/refined-github/issues/3908 https://github.com/sindresorhus/refined-github/issues/3898

PR #3750, opened 2020-11-20
Add `pr-easy-toggle-files` feature

### LINKED ISSUE #3738: Pull request / Tab Files - Toggle files by clicking on their header bar
When viewing the files list of a pull request,
it would be nice to be allowed to toggle any file by clicking on their "header bar" (bar that contains the filename, the checkbox `[ ] Viewed` and the 3-dots menu).

For instance on this page:
https://github.com/microsoft/TypeScript/pull/12114/files

This feature should appear on this kind of URL:
`https://github.com/<owner>/<repository>/pull/<pr_id>/files`

MrChocolatine:
1. LINKED ISSUES:
  Fixes #3738.

2. TEST URLS:
   https://github.com/microsoft/TypeScript/pull/12114/files
   https://github.com/sindresorhus/refined-github/pull/3750/files

3. SCREENSHOT:

refined-github--pr-easy-toggle-files

### DIFF
--- a/source/features/pr-easy-toggle-files.tsx
+++ b/source/features/pr-easy-toggle-files.tsx
@@ -1,5 +1,4 @@
 import select from 'select-dom';
-import delegate from 'delegate-it';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
--- a/source/features/pr-easy-toggle-files.tsx
+++ b/source/features/pr-easy-toggle-files.tsx
@@ -9,7 +9,7 @@ function toggleFile(event: Event): void {
 
 	// The clicked element is either the bar itself or one of its 2 children
 	if (elementClicked === headerBar || elementClicked.parentElement === headerBar) {
-		select('.file-info > button', headerBar)!.click();
+		select('[aria-label="Toggle diff contents"]', headerBar)!.click();
 	}
 }
 
--- a/source/features/pr-easy-toggle-files.tsx
+++ b/source/features/pr-easy-toggle-files.tsx
@@ -4,23 +4,18 @@ import * as pageDetect from 'github-url-detection';
 
 import features from '.';
 
-const selectorHeaderBar = '.file-header.file-header--expandable';
-
-function toggleFile(event: delegate.Event<Event, HTMLElement>): void {
+function toggleFile(event: Event): void {
 	const elementClicked = event.target as HTMLElement;
-	const headerBar = elementClicked.closest(selectorHeaderBar);
+	const headerBar = elementClicked.closest<HTMLElement>('.file-header.file-header--expandable')!;
 
 	// The clicked element is either the bar itself or one of its 2 children
-	if (
-		elementClicked.isSameNode(headerBar) ||
-		elementClicked.parentElement!.isSameNode(headerBar)
-	) {
-		select('.file-info > button', headerBar as HTMLElement)!.click();
+	if (elementClicked === headerBar || elementClicked.parentElement === headerBar) {
+		select('.file-info > button', headerBar)!.click();
 	}
 }
 
 async function init(): Promise<void> {
-	delegate('#files', selectorHeaderBar, 'click', toggleFile);
+	document.body.addEventListener('click', toggleFile);
 }
 
 void features.add(__filebasename, {
--- a/source/features/pr-easy-toggle-files.tsx
+++ b/source/features/pr-easy-toggle-files.tsx
@@ -1,0 +1,32 @@
+import select from 'select-dom';
+import delegate from 'delegate-it';
+import * as pageDetect from 'github-url-detection';
+
+import features from '.';
+
+const selectorHeaderBar = '.file-header.file-header--expandable';
+
+function toggleFile(event: delegate.Event<Event, HTMLElement>): void {
+	const elementClicked = event.target as HTMLElement;
+	const headerBar = elementClicked.closest(selectorHeaderBar);
+
+	// The clicked element is either the bar itself or one of its 2 children
+	if (
+		elementClicked.isSameNode(headerBar) ||
+		elementClicked.parentElement!.isSameNode(headerBar)
+	) {
+		select('.file-info > button', headerBar as HTMLElement)!.click();
+	}
+}
+
+async function init(): Promise<void> {
+	delegate('#files', selectorHeaderBar, 'click', toggleFile);
+}
+
+void features.add(__filebasename, {
+	include: [
+		pageDetect.isPRFiles
+	],
+	awaitDomReady: false,
+	init
+});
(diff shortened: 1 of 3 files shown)


source/features/pr-easy-toggle-files.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+const selectorHeaderBar = '.file-header.file-header--expandable';
+
+function toggleFile(event: delegate.Event<Event, HTMLElement>): void {
+	const elementClicked = event.target as HTMLElement;
+	const {parentElement} = elementClicked;
+
+	if (parentElement === null) {
+		return;
+	}
```
fregante, line 15: This never happens. If you're trying to fix the TypeScript error use `!` after the variable name.

source/features/pr-easy-toggle-files.tsx
```diff
@@ -0,0 +1,48 @@
...
+	 * A header bar is displayed with flexbox and 2 cells:
+	 *
+	 * 1. Cell on the left
+	 * The largest one, contains the toggle button, the filename...
+	 * Has a large empty area (`.file-header.file-header--expandable > .file-info`) so there is
+	 * a high probability we click on it since it is the easiest.
+	 *
+	 * 2. Cell on the right
+	 * Contains the checkbox "[ ] Viewed" and the 3-dots menu
+	 */
```
fregante, line 27: This is unrelated to what the code does. It could be replaced with "The clicked element is either the bar itself or its 2 children" or more specifically "Only capture clicks on the bar itself or its 2 flexbox children"

source/features/pr-easy-toggle-files.tsx
```diff
@@ -0,0 +1,48 @@
...
+	if (
+		parentElement.matches(selectorHeaderBar) ||
+		elementClicked.matches(selectorHeaderBar)
+	) {
+		const button = select('button', elementClicked) ?? select('button', parentElement);
```
fregante, line 32: Instead of `delegate`, `.matches` and `select`, you can use this:
```suggestion
	const elementClicked = event.target as HTMLElement;
	const bar = clickedElement.closest(selectorHeaderBar);
	if (bar === clickedElement || bar.parentElement === clickedElement) {
		select(this needs a better selector, bar)!.click();
```

fregante (commented):
Thanks for the PR! Here's a few suggestions

source/features/pr-easy-toggle-files.tsx
MrChocolatine: Indeed, I used it for other parts for forgot for this one.

source/features/pr-easy-toggle-files.tsx
MrChocolatine: I decided to use [`isSameNode()`](https://developer.mozilla.org/en-US/docs/Web/API/Node/isSameNode) instead of `===`.

MrChocolatine:
Updated according to your suggestions @fregante.

source/features/pr-easy-toggle-files.tsx
fregante: That is the same but longer. No need

fregante:
Thanks! Generally here force pushing is not necessary and it makes reviewing only the changes harder

source/features/pr-easy-toggle-files.tsx
```diff
@@ -4,23 +4,18 @@ import * as pageDetect from 'github-url-detection';
...
-	) {
-		select('.file-info > button', headerBar as HTMLElement)!.click();
+	if (elementClicked === headerBar || elementClicked.parentElement === headerBar) {
+		select('.file-info > button', headerBar)!.click();
 	}
 }
 
 async function init(): Promise<void> {
-	delegate('#files', selectorHeaderBar, 'click', toggleFile);
+	document.body.addEventListener('click', toggleFile);
```
MrChocolatine, line 17: I used that before, but by seeing at other files I thought you would prefer the use of `delegate()`. Care to explain the reason?

source/features/pr-easy-toggle-files.tsx
MrChocolatine: I wasn't aware of that, good to know thanks!

MrChocolatine:
> Thanks! Generally here force pushing is not necessary and it makes reviewing only the changes harder

You are welcome! Your extension is great and brings a lot of convenience.

@fregante no offense at all but for the "it makes reviewing only the changes harder" when I see some commits simply titled "shorten" and "lint" 😅 ...

source/features/pr-easy-toggle-files.tsx
fregante: Thanks for checking the other features too

`delegate` uses `.closest` internally to filter the events, but since we're our listener does it explicitly, we're doing the same thing twice.

Alternatively, if we were to use `delegate`, we'd just need:

```diff
	const headerBar = elementClicked.closest<HTMLElement>('.file-header.file-header--expandable')!;
	const headerBar = element.delegateTarget;
```

They're pretty much equivalent, except that `delegate` does a little more under the hood and therefore since it's easier to avoid it in this case, I removed it.

fregante:
> when I see some commits simply titled "shorten" and "lint" 😅 ...

I use "lint" generously to describe "obvious" changes that are only meant to make the build, types or linter pass.

As for the "shorter" I just followed what my previous review suggested + I left a comment there explaining why `===` is enough.

We squash commits in the end, so my comment was more about _being able to tell the difference between a review and the next,_ so that I don't need to review the same piece of code twice.

source/features/pr-easy-toggle-files.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+import features from '.';
+
+function toggleFile(event: Event): void {
+	const elementClicked = event.target as HTMLElement;
+	const headerBar = elementClicked.closest<HTMLElement>('.file-header.file-header--expandable')!;
+
+	// The clicked element is either the bar itself or one of its 2 children
+	if (elementClicked === headerBar || elementClicked.parentElement === headerBar) {
+		select('.file-info > button', headerBar)!.click();
```
fregante, line 12:
```suggestion
		select('[aria-label="Toggle diff contents"]', headerBar)!.click();
```

MrChocolatine: Are you sure? I thought relying on an attribute so "unstable"/volatile (I mean his value could change more easily and more quickly than the others) would not be a good idea.

I also thought about using `[aria-expanded]` as in the left part of the bar, only the button would have this attribute.

source/features/pr-easy-toggle-files.tsx
MrChocolatine: I see, thanks.

source/features/pr-easy-toggle-files.tsx
fregante: > Are you sure? I thought relying on an attribute so "unstable"/volatile (I mean his value could change more easily and more quickly than the others) would not be a good idea.

heheh welcome to the world of "hacking on content you don't own."

**Anything** we use can change with **any** frequency, so it's best to use the most unique identifiers rather than loose selectors, because GitHub might change the order of elements and suddenly `.file-info > button` does something you don't expect at all.

If `[aria-label="Toggle diff contents"]` changes, the click will do nothing (with a console error) and that's still preferable to "doing the unexpected."

GitHub has been removing many unique classes so we're left with fickle position-based selectors. Where possible, even if it requires more code, it's best to still select the most unique part of the page and then traverse the DOM from there, for example:

https://github.com/sindresorhus/refined-github/blob/38fb7b001222421c6e3d3e17bfddb3d373a37739/source/features/deprioritize-marketplace-link.tsx#L22-L24

`data-ga-click` and `aria-label` are the attributes that are less likely to change, whereas even `.file-info` could be gone one day in place of `.box.position-relative.px-2`

MrChocolatine: And I completely understand, and in this direction I still thought `[aria-expanded]` would be a better choice. But I will update the selector like your suggestion then.

fregante (approved):

yakov116 (approved):

fregante:
Thanks Mr Pan au Chocolate

MrChocolatine:
> Thanks Mr Pan au Chocolate

Noooo, damned! How is that possible 😄 ?!

PR #3756, opened 2020-11-22
Meta: Extract modules `one-mutation` `push-form` `regex-join`

fregante:
I'm extracting some useful modules:

- https://github.com/fregante/one-mutation
- https://github.com/fregante/push-form
- https://github.com/fregante/regex-join


source/features/highlight-deleted-and-added-files-in-diffs.tsx
```diff
@@ -14,7 +14,7 @@ async function loadDeferred(jumpList: Element): Promise<void> {
 	const retrier = setInterval(() => {
 		jumpList.parentElement!.dispatchEvent(new MouseEvent('mouseover'));
 	}, 100);
-	await observeOneMutation(jumpList);
+	await oneMutation(jumpList, {childList: true, subTree: true});
```
fregante, line 17: I made the tough decision to avoid having this as a default because the options object now also includes a `filter` function.

I guess the alternative would be to support an _optional_ argument between the 2, but it seems awkward nowadays:

```js
// Both would work:
oneMutation(jumpList, {childList: true});
oneMutation(jumpList, filterFunction, {childList: true});
```

---

Anyway, `subTree` might not be required anywhere in this PR, but dropping it requires testing a few already-complex features.

sindresorhus: > Anyway, subTree might not be required anywhere in this PR, but dropping it requires testing a few already-complex features.

Maybe add a todo comment so we don't forget?

fregante (commented):
I think it's just those 2 modules for now. I don't see much else

PR #3767, opened 2020-11-26
Enable `unfinished-comments` on temporarily-hidden text fields

### LINKED ISSUE #3765: `unfinished-comments` doesn't work after toggling Preview comment tab.
`unfinished-comments` doesn't work after toggling Preview comment tab.
Screenshot: 
unfinished

cheap-glitch:
Thanks for contributing! 🍄

1. LINKED ISSUES: Closes #3765 

2. TEST URLS:
- issue: https://github.com/sindresorhus/refined-github/issues/3765
- PR: https://github.com/sindresorhus/refined-github/pull/3767

Tested on Firefox and Chromium in all possible cases (issue comment, PR comment, PR review comment). Side note: it also works on the "New issue"/"New PR" pages now, but we get the same "switching tabs right after posting" problem as we had for the comments (at least on Firefox).

### DIFF
--- a/source/features/unfinished-comments.tsx
+++ b/source/features/unfinished-comments.tsx
@@ -9,7 +9,7 @@ function hasDraftComments(): boolean {
 	// `[disabled]` excludes the PR description field that `wait-for-build` disables while it waits
 	return select.all<HTMLTextAreaElement>('textarea:not([disabled])').some(textarea =>
 		textarea.value !== textarea.textContent && // Exclude comments being edited but not yet changed (and empty comment fields)
-		textarea.offsetWidth > 0 && // Exclude invisible fields
+		(textarea.offsetWidth > 0 || select('.js-preview-body', textarea.form!)!.textContent !== 'Nothing to preview') && // Exclude invisible fields that aren't being previewed
 		!select.exists('.btn-primary[disabled]', textarea.form!) // Exclude forms being submitted
 	);
 }
--- a/source/features/unfinished-comments.tsx
+++ b/source/features/unfinished-comments.tsx
@@ -9,7 +9,6 @@ function hasDraftComments(): boolean {
 	// `[disabled]` excludes the PR description field that `wait-for-build` disables while it waits
 	return select.all<HTMLTextAreaElement>('textarea:not([disabled])').some(textarea =>
 		textarea.value !== textarea.textContent && // Exclude comments being edited but not yet changed (and empty comment fields)
-		(textarea.offsetWidth > 0 || select('.js-preview-body', textarea.form!)!.textContent !== 'Nothing to preview') && // Exclude invisible fields that aren't being previewed
 		!select.exists('.btn-primary[disabled]', textarea.form!) // Exclude forms being submitted
 	);
 }


source/features/unfinished-comments.tsx
```diff
@@ -9,7 +9,7 @@ function hasDraftComments(): boolean {
 	// `[disabled]` excludes the PR description field that `wait-for-build` disables while it waits
 	return select.all<HTMLTextAreaElement>('textarea:not([disabled])').some(textarea =>
 		textarea.value !== textarea.textContent && // Exclude comments being edited but not yet changed (and empty comment fields)
-		textarea.offsetWidth > 0 && // Exclude invisible fields
+		(textarea.offsetWidth > 0 || select('.js-preview-body', textarea.form!)!.textContent !== 'Nothing to preview') && // Exclude invisible fields that aren't being previewed
```
cheap-glitch, line 12: Maybe this should be
```suggestion
		(textarea.offsetWidth > 0 || select('.js-preview-body', textarea.form!)?.textContent !== 'Nothing to preview') && // Exclude invisible fields that aren't being previewed
```
?

fregante: I think so because `textarea` isn't necessarily a comment.

However I think we should drop this check altogether since there are too many exceptions. For examples when you type a final review comment (in the "Review changes" dropdown) and then you close it.

Can you find a list all situations where there are hidden textareas that should be ignored?

cheap-glitch: After a bit of testing, I agree we can drop this check altogether. The only reason it was introduced in the first place is because I relied on just `textarea.value.length` to check for draft comments. But afaik the improved condition we're using now makes checking for hidden textareas redundant.

> For examples when you type a final review comment (in the "Review changes" dropdown) and then you close it.

There's actually a case where it doesn't work regardless: when you review your own PR. It's caused by the disabled `Approve` and `Request change` buttons. I don't know how common that situation is though.

> Can you find a list all situations where there are hidden textareas that should be ignored?

After testing the most common cases I haven't found any. I'll try to think of possible edge cases.

cheap-glitch:
After a bit more testing, I found no cases where invisible textareas caused a problem. There are probably some situations I couldn't/didn't check, but at least the main use cases should be covered.

I did find a few bugs, maybe for another PR:
 * the feature doesn't work for the final review comment when you review your own PR (because of the disabled `Approve` and `Request change` buttons)
 * there are still a few cases where quickly switching to another tab can leave a dangling `Draft comment` in the tab title even though the comment has been sent:
   * when creating a new PR
   * when marking a PR discussion as "resolved" while writing a comment

fregante:
> * the feature doesn't work for the final review comment when you review your own PR (because of the disabled `Approve` and `Request change` buttons)

Works for me!

nook

fregante:
> * when marking a PR discussion as "resolved" while writing a comment

That's fine for me since you didn't actually send the comment. It technically is still there

cheap-glitch:
> > * the feature doesn't work for the final review comment when you review your own PR (because of the disabled `Approve` and `Request change` buttons)
> 
> Works for me!
> 

Did you try on a PR you opened yourself? Because in your example it doesn't seem like the buttons were disabled.

Geeyun-JY3:
It seems that this feature only works on the "New issue"/"New PR" pages when the title textarea is not empty.

cheap-glitch:
> It seems that this feature only works on the "New issue"/"New PR" pages when the title textarea is not empty.

Yes, this is because GitHub disables the "Submit" button.

PR #3768, opened 2020-11-26
Meta: Update dependencies

fregante:
And also use module instead of local file

- https://github.com/fregante/one-event


package.json
```diff
@@ -79,12 +80,12 @@
 		"@rollup/plugin-typescript": "^6.1.0",
 		"@sindresorhus/tsconfig": "^0.7.0",
```
fregante, line 81: I reverted the update to this because it requires 109 changes in the code. It should be its own PR, maybe not one that just adds `!` after each

PR #3774, opened 2020-11-27
Meta: Validate token in options page

### LINKED ISSUE #3752: Verify token in options
When the user pastes a token, we could verify that the token:

1. Is active/valid
2. Has the expected scope(s)

#3734 will add a function that helps with this validation.

yakov116:
1. LINKED ISSUES:
   Closes #3752 

2. TEST URLS:
   The options Page

3. SCREENSHOT:
image

I also fixed the dropdown listener since it was broken


yakov116:
Hold off on the review I want to make changes

source/options.tsx
```diff
@@ -10,6 +10,59 @@ import * as indentTextarea from 'indent-textarea';
...
+			Accept: 'application/vnd.github.v3+json',
+			Authorization: `token ${personalToken}`
+		}
+	});
+
+	if (response.ok) {
+		setValidationStatus('✔️ Validated');
+	} else {
+		const statusText = response.status === 404 ?
+			'Invalid Domain' :
```
yakov116, line 37: Not sure if we need this. I tested on a random site so I added this check.

distribution/options.html
```diff
@@ -6,15 +6,18 @@
...
+			</div>
 		</label>
 	</p>
-	<ul>
-		<li>The token enables <a href="https://github.com/sindresorhus/refined-github/search?q=github-helpers/api" target="_blank">some features</a> to read data from public repositories
-		<li>The <code>public_repo</code> scope lets them edit your public repositories
-		<li>The <code>repo</code> scope lets them edit private repositories as well
-		<li>The <code>delete_repo</code> scope is only used by the <code>quick-fork-deletion</code> feature
+	<ul class="invalid">
+		<li style="list-style: none;">The token enables <a href="https://github.com/sindresorhus/refined-github/search?q=github-helpers/api" target="_blank">some features</a> to read data from public repositories
```
fregante, line 17: If it's not a list item, then it shouldn't be a list item.

I had initially included it here to make it one of the items to _check_, but I saw you put it right next to the field. That's ok too

source/options.tsx
```diff
@@ -10,6 +10,59 @@ import * as indentTextarea from 'indent-textarea';
 
 import {perDomainOptions} from './options-storage';
 
+function setValidationStatus(text: string): void {
+	select('#validation')!.textContent = text;
+	select('form ul')!.classList.toggle('invalid', !text.includes('✔️'));
+}
+
+async function getHeaders(personalToken: string): Promise<string> {
```
fregante, line 18: Make function names reflect what they do. This isn't a plain getter. This function should be refactored

source/options.tsx
```diff
@@ -10,6 +10,59 @@ import * as indentTextarea from 'indent-textarea';
...
+	if (response.ok) {
+		setValidationStatus('✔️ Validated');
+	} else {
+		const statusText = response.status === 404 ?
+			'Invalid Domain' :
+			String((await response.json()).message);
+		setValidationStatus('❌ ' + statusText);
+	}
+
+	return response.headers.get('X-OAuth-Scopes')! ?? '';
```
fregante, line 42: `! ??` never makes sense. `??` already takes care of the null

```suggestion
	return response.headers.get('X-OAuth-Scopes')! ?? '';
```

source/options.tsx
```diff
@@ -117,6 +176,9 @@ function addEventListeners(): void {
 	// Add cache clearer
 	select('#clear-cache')!.addEventListener('click', clearCacheHandler);
 
+	// Add token validation
+	select('[name="personalToken"]')!.addEventListener('change', validateToken);
```
fregante, line 180:
```suggestion
	select('[name="personalToken"]')!.addEventListener('input', validateToken);
```

source/options.tsx
```diff
@@ -99,8 +153,13 @@ async function generateDom(): Promise<void> {
...
 	// Update domain-dependent page content when the domain is changed
-	select('.js-options-sync-selector')?.addEventListener('change', ({currentTarget: dropdown}) => {
-		select<HTMLAnchorElement>('#personal-token-link')!.host = (dropdown as HTMLSelectElement).value;
+	select('.OptionsSyncPerDomain-picker select')?.addEventListener('change', ({currentTarget: dropdown}) => {
+		const host = (dropdown as HTMLSelectElement).value === 'default' ? 'github.com' : (dropdown as HTMLSelectElement).value;
+		select<HTMLAnchorElement>('#personal-token-link')!.host = host;
+		// Delay validating to let options load first
+		setTimeout(() => {
+			void validateToken();
+		}, 100);
```
fregante, line 162:
```suggestion
		setTimeout(validateToken, 100);
```

But better yet if you dispatch an event at the end of https://github.com/fregante/webext-options-sync-per-domain/blob/dcc61fe58ab67d0982b2948cf3c1264375ef4096/index.ts#L129

Example: https://github.com/fregante/webext-options-sync/blob/master/index.ts#L253

… and then listen to that event specifically, instead of changing this listener

source/options.tsx
```diff
@@ -99,8 +153,13 @@ async function generateDom(): Promise<void> {
 
 function addEventListeners(): void {
 	// Update domain-dependent page content when the domain is changed
-	select('.js-options-sync-selector')?.addEventListener('change', ({currentTarget: dropdown}) => {
-		select<HTMLAnchorElement>('#personal-token-link')!.host = (dropdown as HTMLSelectElement).value;
+	select('.OptionsSyncPerDomain-picker select')?.addEventListener('change', ({currentTarget: dropdown}) => {
+		const host = (dropdown as HTMLSelectElement).value === 'default' ? 'github.com' : (dropdown as HTMLSelectElement).value;
+		select<HTMLAnchorElement>('#personal-token-link')!.host = host;
```
fregante, line 158: 🙌

source/options.tsx
```diff
@@ -10,6 +10,59 @@ import * as indentTextarea from 'indent-textarea';
...
+
+async function validateToken(): Promise<void> {
+	const personalToken = select<HTMLInputElement>('[name="personalToken"]')!.value;
+	setValidationStatus('');
+
+	if (!/[\da-f]{40}/.exec(personalToken)) {
+		return;
+	}
+
+	setValidationStatus('Validating...');
```
fregante, line 53:
```suggestion
	setValidationStatus('Validating…');
```

source/options.tsx
```diff
@@ -10,6 +10,59 @@ import * as indentTextarea from 'indent-textarea';
...
+	}
+
+	return response.headers.get('X-OAuth-Scopes')! ?? '';
+}
+
+async function validateToken(): Promise<void> {
+	const personalToken = select<HTMLInputElement>('[name="personalToken"]')!.value;
+	setValidationStatus('');
+
+	if (!/[\da-f]{40}/.exec(personalToken)) {
```
fregante, line 49: The form already validates it, you can use
```suggestion
	if (!tokenField.validity.valid) {
```

source/options.tsx
fregante: Nah, it can be part of a catch-all error handler:

```ts
} else {
	try {
		setValidationStatus('❌ ' + (await response.json()).message);
	} catch (error: unknown) {
		setValidationStatus('❌ ' + (error as Error).message);
	}
}

File: source/options.css [hunk 10]
```diff
@@ -16,15 +16,29 @@ ul {
 	padding-left: 1.5em;
 }
 
+[data-scope] {
+	float: right;
+	margin-right: 50px;
+}
```
fregante, line 22: My idea was that the bullet points themselves can be replaced without causing any layout shifts. This is what I currently see: (ignore the token)

source/options.css
```diff
@@ -16,15 +16,29 @@ ul {
...
 	line-height: 1.5;
 }
 
+#validation {
+	display: inline-block;
+}
+
 [name='personalToken'] {
 	width: 20em !important; /* https://github.com/sindresorhus/refined-github/issues/1374#issuecomment-397906701 */
+	display: inline-block !important;
```
fregante, line 41: Why is this important?

distribution/options.html
```diff
@@ -6,15 +6,18 @@
 	<p>
 		<label>
 			<strong>Personal token</strong> (optional, <a id="personal-token-link" href="https://github.com/settings/tokens/new?description=Refined%20GitHub&scopes=repo" target="_blank">generate one</a>)<br>
-			<!-- placeholder is set to enable use of :placeholder-shown CSS selector -->
-			<input type="text" name="personalToken" spellcheck="false" autocomplete="off" pattern="[\da-f]{40}" placeholder=" ">
+			<div>
```
fregante, line 9: This wrapper doesn't seem needed

source/options.css
```diff
@@ -16,15 +16,29 @@ ul {
...
+
 :root [name='customCSS'],
 :root [name='personalToken'] {
 	font-family: SFMono-Regular, Consolas, 'Liberation Mono', Menlo, Courier, monospace;
 	font-size: 11px;
 	line-height: 1.5;
 }
 
+#validation {
+	display: inline-block;
```
fregante, line 36: You might want to just use a `span` instead of a `div`

fregante (changes requested):

source/options.css
yakov116: What browser? I tried in both Firefox and chrome.

fregante: Chrome

source/options.tsx
yakov116: Is it good now?

Not sure how to spit it any more since I need to get the response if it errors

source/options.css
yakov116: Needed so that the span displays next to it

source/options.tsx
yakov116: I tried to figure this out. But I could not get it to work by listening to the event.

fregante (changes requested):

distribution/options.html
fregante: > If it's not a list item, then it shouldn't be a list item.

_and shouldn't be aligned like a list item, and shouldn't be part of `ul`_

source/options.tsx
fregante: Nothing changed, you just moved some code into a separate function, but the function is still being called from here.

- This doesn't return multiple headers, so it shouldn't be called headerS
- This is returning scopes, so return scopes, pre-parsed into an array


```suggestion
async function getTokenScopes(personalToken: string): Promise<string[]> {
```

mohdjuwaiser (approved):

source/options.css
```diff
@@ -16,6 +16,14 @@ ul {
 	padding-left: 1.5em;
 }
 
+ul:not(.invalid) li.valid {
+	list-style: '✔️ ';
```
fregante, line 20: Unfortunately Safari doesn't support custom strings in `list-style`, so this needs to be reworked with `::before` instead

source/options.tsx
```diff
@@ -10,6 +10,61 @@ import * as indentTextarea from 'indent-textarea';
...
+	if (!tokenField.validity.valid || tokenField.value.length === 0) {
+		return;
+	}
+
+	displayValidationStatus('Validating…');
+
+	let tokenScopes;
+	try {
+		tokenScopes = await getTokenScopes(tokenField.value);
+		displayValidationStatus('✔️ Token Validated');
```
fregante, line 53:
```suggestion
		displayValidationStatus('✔️ Valid');
```

source/options.tsx
```diff
@@ -10,6 +10,63 @@ import * as indentTextarea from 'indent-textarea';
...
+	const response = await fetch(url, {
+		cache: 'no-store',
+		headers: {
+			'User-Agent': 'Refined GitHub',
+			Accept: 'application/vnd.github.v3+json',
+			Authorization: `token ${personalToken}`
+		}
+	});
+
+	void setValidationStatus(response);
```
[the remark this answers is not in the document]
fregante: Part of this code is missing

source/options.css
yakov116: ok.

I dont have safari so I will do it and have you test it

fregante: There's no need. `::before` works everywhere. Just replace it and it's fine

source/options.tsx
```diff
@@ -10,6 +10,61 @@ import * as indentTextarea from 'indent-textarea';
...
+		cache: 'no-store',
+		headers: {
+			'User-Agent': 'Refined GitHub',
+			Accept: 'application/vnd.github.v3+json',
+			Authorization: `token ${personalToken}`
+		}
+	});
+
+	if (!response.ok) {
+		const message = (await response.json())?.message ?? response.status;
```
yakov116, line 34:
```suggestion
		const message = (await response.json())?.message
```

An I right that this is not even possible?

fregante: I'd expect that a non-ok response would include an error, so yeah, we can drop `?.` too
```suggestion
		const {message} = (await response.json())!;
```

yakov116:
Should all be good now. (I was looking over the code and I was wondering where my updated code was. Turns out I forgot to push :hehe)

fregante:
Full demo in Chrome (field size changed only in demo):



Dark Chrome:

Screen Shot
Screen Shot 1

Dark Firefox: 

Screen Shot 2


Dark Safari:

Screen Shot 4

fregante (approved):

source/options.tsx
```diff
@@ -10,6 +10,76 @@ import * as indentTextarea from 'indent-textarea';
...
+	if (scopes.includes('repo')) {
+		scopes.push('public_repo');
+	}
+
+	return scopes;
+}
+
+async function validateToken(): Promise<void> {
+	const tokenField = select('input[name="personalToken"]')!;
+	reportStatus({});
```
fregante, line 67:
```suggestion
	reportStatus({});
	const tokenField = select('input[name="personalToken"]')!;
```

source/options.tsx
```diff
@@ -10,6 +10,76 @@ import * as indentTextarea from 'indent-textarea';
...
+	if (!tokenField.validity.valid || tokenField.value.length === 0) {
+		return;
+	}
+
+	reportStatus({text: 'Validating…'});
+
+	try {
+		const scopes = await getTokenScopes(tokenField.value);
+		reportStatus({scopes});
```
fregante, line 76:
```suggestion
		reportStatus({
			scopes: await getTokenScopes(tokenField.value)
		});
```

source/options.tsx
```diff
@@ -99,8 +170,11 @@ async function generateDom(): Promise<void> {
 
 function addEventListeners(): void {
 	// Update domain-dependent page content when the domain is changed
-	select('.js-options-sync-selector')?.addEventListener('change', ({currentTarget: dropdown}) => {
-		select('a#personal-token-link')!.host = (dropdown as HTMLSelectElement).value;
+	select('.OptionsSyncPerDomain-picker select')?.addEventListener('change', ({currentTarget: dropdown}) => {
+		const host = (dropdown as HTMLSelectElement).value.replace(/^default$/, 'github.com');
+		select('a#personal-token-link')!.host = host;
```
fregante, line 175:
```suggestion
		const host = (dropdown as HTMLSelectElement).value;
		select('a#personal-token-link')!.host = host === 'default' ? 'github.com' : host;
```

yakov116:
image

PR #3783, opened 2020-12-01
Add `clean-repo-sidebar` feature

cheap-glitch:
Thanks for contributing! 🍄

1. LINKED ISSUES: Closes #3668 

2. TEST URLS:
* [Description, releases & no packages](https://github.com/sindresorhus/refined-github)
* [Empty meta](https://github.com/Witold101/wpw)
* [Repo with packages](https://github.com/Jaboo9/github-packages)

3. SCREENSHOTS:

<details>
<summary><code>clean-repo-sidebar</code></summary>

clean-repo-sidebar

</details>

<details>
<summary><code>hide-zero-packages</code></summary>

hide-zero-packages

</details>

PR #3784, opened 2020-12-02
Fix `hide-zero-packages` on repo roots

cheap-glitch:
Thanks for contributing! 🍄

1. LINKED ISSUES: None

This feature broke because the URL of the "Packages" links now ends with some query parameters (e.g. `https://github.com/users/sindresorhus/packages?repo_name=refined-github`)

### DIFF
--- a/source/features/hide-zero-packages.tsx
+++ b/source/features/hide-zero-packages.tsx
@@ -6,7 +6,7 @@ import getTabCount from './remove-projects-tab';
 
 async function init(): Promise<void | false> {
 	const packagesTab = await elementReady([
-		'.BorderGrid-cell a[href$="/packages"]', // `isRepoRoot`
+		'.BorderGrid-cell a[href*="/packages"]', // `isRepoRoot`
 		'.UnderlineNav-item[href$="?tab=packages"]:not(.selected)' // `isUserProfile`
 	].join());
 


source/features/hide-zero-packages.tsx
```diff
@@ -6,7 +6,7 @@ import getTabCount from './remove-projects-tab';
 
 async function init(): Promise<void | false> {
 	const packagesTab = await elementReady([
-		'.BorderGrid-cell a[href$="/packages"]', // `isRepoRoot`
+		'.BorderGrid-cell a[href*="/packages"]', // `isRepoRoot`
```
fregante, line 9: This would also catch any repo that has the word “packages” in it, doesn’t it?

fregante, line 9: This would also catch any repo that has the word “packages” in it, doesn’t it?

cheap-glitch: You're right, didn't think about that. Then how about
```suggestion
		'.BorderGrid-cell a[href*="/packages?"]', // `isRepoRoot`
```

cheap-glitch:
Closing because of https://github.com/sindresorhus/refined-github/pull/3783#issuecomment-738695463.

PR #3787, opened 2020-12-03
Meta: Update Sourcegraph Readme link

tjkandala:
The old link is now broken. Thanks for including us in your list of related extensions!


yakov116:
❤️

PR #3794, opened 2020-12-07
Fix `new-repo-disable-projects-and-wiki` feature

cheap-glitch:
Thanks for contributing! 🍄

1. LINKED ISSUES: none

2. TEST URLS: https://github.com/new

The `new-repo-disable-projects-and-wiki` feature was the only one passing an empty query string to `api.v3()`, which caused a trailing slash to appear at the end of the URL and resulted in a 404 error.

### DIFF
--- a/source/github-helpers/api.ts
+++ b/source/github-helpers/api.ts
@@ -114,7 +114,7 @@ export const v3 = mem(async (
 	const {personalToken} = await settings;
 
 	if (!query.startsWith('https')) {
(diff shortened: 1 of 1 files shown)


yakov116:
Just pass a slash instead of an empty string

cheap-glitch:
> Just pass a slash instead of an empty string in the feature

This won't work. We want `'repos/' + getRepo()!.nameWithOwner`, not an empty URL.

yakov116:
Oh your right misunderstood the issue

yakov116:
Thanks @cheap-glitch

PR #3795, opened 2020-12-07
Expand `quick-fork-deletion` to empty repos

cheap-glitch:
Thanks for contributing! 🍄

1. LINKED ISSUES: Closes #3779 

2. TEST URLS: see https://github.com/sindresorhus/refined-github/pull/3734#issue-523513983

3. SCREENSHOT: **TODO**

* Renamed the feature to `quick-repo-deletion`
* Fixed an incorrect selector that broke the feature
* Enabled the feature on every repo and added an additional confirmation step for non-forks

PR #3796, opened 2020-12-07
Fix `more-dropdown`

### LINKED ISSUE #3811: `more-dropdown`: Cannot read property 'parentElement' of null
- [x] 1. Checked.
- [x] 2. [I noticed it here](https://github.com/warrensbox/terraform-switcher), but it appears in all the repositories pages.
- [x] 3. A screenshot: 

```
refined-github.js:2098 ❌ Refined GitHub → more-dropdown → TypeError: Cannot read property 'parentElement' of null
    at onlyShowInDropdown (refined-github.js:3113)
    at init (refined-github.js:3130)
    at async runFeature (refined-github.js:2114)
    at async setupPageLoad (refined-github.js:2122) 
Search issue: https://github.com/sindresorhus/refined-github/issues?q=is%3Aissue+Cannot%20read%20property%20'parentElement'%20of%20null

Open an issue: https://github.com/sindresorhus/refined-github/issues/new?labels=bug&template=bug_report.md&title=%60more-dropdown%60%3A%20Cannot%20read%20property%20'parentElement'%20of%20null
		
logError @ refined-github.js:2098
runFeature @ refined-github.js:2116
async function (async)
runFeature @ refined-github.js:2114
setupPageLoad @ refined-github.js:2122
add @ refined-github.js:2156
async function (async)
add @ refined-github.js:2141
(anonymous) @ refined-github.js:3120
(anonymous) @ refined-github.js:8132
(anonymous) @ refined-github.js:8133
```

cheap-glitch:
Thanks for contributing! 🍄

1. LINKED ISSUES:  Fixes #3811 

3. SCREENSHOT: -

### DIFF
--- a/source/features/more-dropdown.tsx
+++ b/source/features/more-dropdown.tsx
@@ -72,8 +72,8 @@ async function init(): Promise<void> {
 			createDropdownItem('Branches', branchesUrl)
 		);
 
-		onlyShowInDropdown('security-tab');
-		onlyShowInDropdown('insights-tab');
+		onlyShowInDropdown('i4security-tab');
+		onlyShowInDropdown('i5insights-tab');
 		return;
 	}
 


yakov116:
@fregante Maybe we should use `[data-ga-click]` ?

yakov116:
@fregante maybe we should use `[data-ga-click]`?

fregante:
If it works with a simple id change in this case I don’t see why complicate it with ga attributes.

yakov116:
I was thinking of enterprise when they change it we may have to start supporting 2 selectors

cheap-glitch:
> @fregante Maybe we should use `[data-ga-click]` ?

We could use `nav [data-content="Security"]` like I did here https://github.com/sindresorhus/refined-github/pull/3795/files#r537577913, but in this case it would imply rewriting part of the feature — which is why I chose the simpler option of just updating the classes.

I think we can wait and see, if it breaks again then a partial rewrite will be warranted.

fregante:
Good point, I think however that GHE uses the older part of this feature, which is unaffected. 

I think, even if it breaks, it will just result in the 2 tabs *not* being hidden, which isn’t a big deal.

fregante:
What I’m interested in at the moment is understanding why they made this change and how this affects us. Are they using the number for ordering? Should we use it too? Did they solve our old https://github.com/sindresorhus/refined-github/issues/3347 issue?

yakov116:
@fregante can we merge this today so it can ship. However we should open an issue to investigate the ordering.

PR #3804, opened 2020-12-09
Add `theme-switch` feature

### LINKED ISSUE #3802: Toggle for Github Theme
Since github is supporting theme natively now, Could you please add a toggle on home screen for toggling the theme right away from the Homepage of github instead of sifting through appearance settings?

kidonng:
Thanks for contributing! 🍄

1. LINKED ISSUES:
   Resolves #3802

2. TEST URLS:
   Anywhere except Gists (which doesn't have a dark theme yet)

3. SCREENSHOT:

  image

P.S. I was expecting some headache before I accidentally found that we can just put the form into the dropdown directly. I love this implementation 😄 

---

I just realized GitHub has a toggle on your own profile page:

image

But it doesn't support "auto" (follow system) so sticking to the settings one.


yakov116:
And it's also on the dashboard, not sure this is worth it.

kidonng:
> And it's also on the dashboard

I clicked the `X` and it is removed 😅 I guess it won't be there forever

fregante:
The feature hasn’t been accepted yet

PR #3806, opened 2020-12-09
Further improvements to Dark mode

mcornella:
I believe this PR fixes all Dark mode issues, so close issue #3798 once this is merged if you agree.

-----
1. LINKED ISSUES:
   Fix border issues in #3798:
   - [Pinned issues border](https://github.com/sindresorhus/refined-github/issues/3798#issuecomment-741568995).
   - [Divider line in grouped events in Activity list](https://github.com/sindresorhus/refined-github/issues/3798#issuecomment-741604005).

2. TEST URLS:
   - [Pinned issues example](https://github.com/sindresorhus/refined-github/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc).
   - [Activity list example](https://github.com/) (see while logged in).

3. SCREENSHOTS:

<table>
<tr>
	<td>Before
	<td>After
<tr>
	<td>pinned issues - before
	<td>pinned issues - after
</table>

<table>
<tr>
	<td>Before
	<td>After
<tr>
	<td>grouped events divider line - before
	<td>grouped events divider line - after
</table>

Since this PR uses the `--github-border-color` CSS variable already present in the codebase, this automatically fixes other border colors that use this variable. This includes table-input cell borders in comments:

<table>
<tr>
	<td>Before
	<td>After
<tr>
	<td>table-input cells - before
	<td>table-input cells - after
</table>

This is a better approach than making `--github-border-color` a fallback like I did in the previous PR since it requires less changes.

---

## More fixes

1. Feature: `parse-backticks`. Use the `--color-markdown-code-bg` CSS variable to color the background in backticks parsed by the `parse-backticks` feature. To find the appropriate CSS variable I inspected the CSS of code elements in Markdown content (you can do it yourself with this very comment). [First reported here](https://github.com/sindresorhus/refined-github/issues/3798#issuecomment-742444029).

	<table>
	<tr>
		<td>Before
		<td>After
	<tr>
		<td>before
		<td>after
	</table>

2. Feature: `split-issue-pr-search-results`. In a [GitHub search](https://github.com/search?q=asdfasdf&type=Issues), remove the background color in split Issues and PRs, then fix the color when hovering by using the `--color-menu-bg-active` CSS variable. I had to use `.menu-item` to add specificity to the selector, otherwise it could be done with `!important`.

	<table>
	<tr>
		<td>Before
		<td>After
	<tr>
		<td>before
		<td>after
	</table>

[...]

Issue #3798, opened 2020-12-08
Support for Github Darkmode


fionera:
When using Refined Github with the new Github Darkmode the Activity List is rendered with a white background.

https://imgur.com/a/uy8K50q

kubukoz:
Also, this:

![image](https://user-images.githubusercontent.com/894884/101520485-dcee4080-3984-11eb-97c5-e0997dffa513.png)

yakov116:
OH MY OMG OMG. So cool

@fregante this is going to be real fun.....

mcornella:
This custom CSS makes it work in both light and dark mode:
```css
.dashboard .js-all-activity-header + div {
    background-color: var(--color-bg-canvas);
    border-color: var(--color-border-primary);
}
```
Go to the extension settings to apply that.

jonrohan:
The best way to go is to use the CSS variables GitHub is now using, here's a list.

https://primer.style/primitives/

kamal:
Also this 
![image](https://user-images.githubusercontent.com/269/101529578-99281680-3945-11eb-841d-99744d4009fb.png)

busches:
> The best way to go is to use the CSS variables GitHub is now using, here's a list.
> 
> https://primer.style/primitives/

Best way shouldn't break every enterprise user out there.

fregante:
@busches `var(--property, fallback)`

mcornella:
Would this work?
```css
.dashboard .js-all-activity-header + div {
    background-color: var(--color-bg-canvas, #fff);
    border-color: var(--color-border-primary, var(--github-border-color));
}
```

mcornella:
[Also this for the reaction border colors](https://github.com/sindresorhus/refined-github/issues/3798#issuecomment-740790219):
```css
button.reaction-summary-item {
	border-top-color: var(--color-border-primary, rgb(225, 228, 232));
    border-bottom-color: var(--color-border-primary, rgb(225, 228, 232));
}
```

aliglelo:
> This custom CSS makes it work in both light and dark mode:
> 
> ```css
> .dashboard .js-all-activity-header + div {
>     background-color: var(--color-bg-canvas);
>     border-color: var(--color-border-primary);
> }
> ```
> 
> Go to the extension settings to apply that.
> 
> EDIT: submitted #3801 which fixes this particular issue. There may be other problems though.

Works like a charm.Kudos :heart:

fregante:
[...]

PR #3812, opened 2020-12-11
Fix `parse-backticks` in PR commit list and Compare page

mcornella:
1. LINKED ISSUES:
   None specifically, but these two comments: https://github.com/sindresorhus/refined-github/issues/3798#issuecomment-742561887, https://github.com/sindresorhus/refined-github/pull/3806#issuecomment-742667247.

2. TEST URLS:
   [This PR with many commits](https://github.com/sindresorhus/refined-github/pull/1986).

3. SCREENSHOT:
	<table>
	<tr>
		<td>Before
		<td>After
	<tr>
		<td>backticks in PR commit list - before
		<td>backticks in PR commit list - after
	</table>

-----

I also fixed the way the `:not(.rgh-backticks-already-parsed)` pseudo-class is added to the selectors, in the case where multiple selectors are grouped in a single array item (see commit 2).

### DIFF
--- a/source/features/parse-backticks.tsx
+++ b/source/features/parse-backticks.tsx
@@ -15,7 +15,9 @@ function init(): void {
 		'.Details[data-issue-and-pr-hovercards-enabled] .d-none a.link-gray-dark', // `isRepoRoot` commit message
 		'.commit-title', // `isCommit` commit message
 		'.commit-desc', // `isCommit` commit description
-		'.js-commit .pr-1 > code', // `isPRConversation`, `isCompare`, `isReleasesOrTags` pushed commits
+		'.js-commit .pr-1 > code', // `isPRConversation` pushed commits
+		'.js-details-container .pr-1 > code', // `isCompare` pushed commits
+		'.Box-row .mb-1 a', // `isCompare` open Pull Request title
 		'.blame-commit-message', // `isBlame` commit message
 		'a[id^="issue_"]', // `isConversationList` issue and PR title
 		'.TimelineItem-body > del', // `isIssue`, `isPRConversation` title edits
--- a/source/features/parse-backticks.tsx
+++ b/source/features/parse-backticks.tsx
@@ -13,7 +13,7 @@ function init(): void {
 		'.js-commits-list-item .mb-1, .js-commits-list-item pre', // `isCommitList` commit message and description
 		'.Details[data-issue-and-pr-hovercards-enabled] .d-none a.link-gray-dark', // `isRepoRoot` commit message
 		'.commit-title, .commit-desc', // `isCommit` commit message and description
-		'.commit-message', // `isPRConversation`, `isCompare`, `isReleasesOrTags` pushed commits
+		'.js-commit .pr-1 > code', // `isPRConversation`, `isCompare`, `isReleasesOrTags` pushed commits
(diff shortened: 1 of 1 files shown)


fregante:
Can you react with 🚀 on each comment that this PR fixes in the main issue? So I mark them as resolved once I merge it

source/features/parse-backticks.tsx
```diff
@@ -35,7 +35,7 @@ function init(): void {
 		'.profile-timeline-card .text-gray-dark', // `isUserProfileMainTab` issue and PR title
 		'#user-repositories-list [itemprop="description"]', // `isUserProfileRepoTab` repository description
 		'.js-hovercard-content > .Popover-message .link-gray-dark' // Hovercard
-	].map(selector => selector + ':not(.rgh-backticks-already-parsed)').join();
+	].map(s => s.split(',').map(selector => selector + ':not(.rgh-backticks-already-parsed)').join()).join();
```
fregante, line 38: Good catch, but this situation is ridiculous. Let's just split the selectors

source/features/parse-backticks.tsx
```diff
@@ -13,7 +13,7 @@ function init(): void {
 		'.js-commits-list-item .mb-1, .js-commits-list-item pre', // `isCommitList` commit message and description
 		'.Details[data-issue-and-pr-hovercards-enabled] .d-none a.link-gray-dark', // `isRepoRoot` commit message
 		'.commit-title, .commit-desc', // `isCommit` commit message and description
-		'.commit-message', // `isPRConversation`, `isCompare`, `isReleasesOrTags` pushed commits
+		'.commit-message, .js-commit .pr-1 > code', // `isPRConversation`, `isCompare`, `isReleasesOrTags` pushed commits
```
fregante, line 16: If `.commit-message` no longer appears in any of those pages, let's drop it. I don't want to keep old, unmarked selectors around forever.

mcornella: I assume older GHE versions still have this, that's why I didn't remove it.

yakov116: These selectors are always changing we are not going to keep every selector since maybe GHE is still using it.

mcornella: Can't find pushed commits in `isReleasesOrTags` pages to verify the selector. Is that no longer in GitHub or should I leave it in the comment?

yakov116: https://github.com/sindresorhus/refined-github/releases all the bullets

mcornella: That is markdown, it's highlighted already without the extension.

yakov116: Taken from https://github.com/sindresorhus/refined-github/pull/2947. I blamed the file to find the PR
https://github.com/sindresorhus/refined-github/releases/tag/20.4.1 (I think)

mcornella: https://github.com/sindresorhus/refined-github/releases/tag/20.4.1 is targeted (now at least) by `.commit-desc` and https://github.com/yakov116/TestR/releases/tag/1 is targeted by `.release-header`, so I'm going to assume that the `.commit-message` selector is definitively defunct and remove the `isReleasesOrTags` part of the comment.

fregante (approved):
The PR title could be updated

PR #3817, opened 2020-12-13
Allow `clean-dashboard` to be disabled

### LINKED ISSUE #3816: Let users disable `clean-dashboard`
I'm talking about the homepage: https://github.com/

I prefer the original GitHub homepage activities section over what this plugin does. Is there an easy way to toggle it off? I've searched through the options and haven't found anything.

I have tried using these CSS rules to disable some of the work done in this area, but I wish there was a global CSS rule I can toggle to disable all modifications done in that section.

```css
:root > * {
  --github-border-color: var(--color-border-primary);
}

.dashboard .flex-items-baseline .f6.mt-2 {
}

/* leave the box, I like it */
.dashboard .js-all-activity-header + div .Box {
}

/* Remove the top border */
.dashboard-rollup-items .body {
}

/* Remove the border between collapsible items */
.dashboard .body > .border-bottom {
}

.dashboard .flex-items-baseline.py-3 {
}
```

It just feels too condensed for my liking, and I prefer the way GitHub spaces all the items out.

cheap-glitch:
Thanks for contributing! 🍄

1. LINKED ISSUES: Closes #3816 

2. TEST URLS: [Dashboard](https://github.com)

3. SCREENSHOT:

<table>
<tr>
	<td>Before
	<td>After
<tr>
	<td>
	<td>
</table>

### DIFF
--- a/source/features/clean-dashboard.css
+++ b/source/features/clean-dashboard.css
@@ -8,14 +8,14 @@
 */
 
 /* Hide duplicate repo name in some non-grouped events */
-.dashboard .watch_started .border-gray.flex-items-baseline .text-bold.text-gray-dark,
-.dashboard .repo .border-gray.flex-items-baseline .text-bold.text-gray-dark,
-.dashboard .public .border-gray.flex-items-baseline .text-bold.text-gray-dark {
+.rgh-clean-dashboard .dashboard .watch_started .border-gray.flex-items-baseline .text-bold.text-gray-dark,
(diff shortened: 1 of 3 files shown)


yakov116:
I don't think we should do this. We can provide the css to disable the feature.

fregante:
I don’t feel super strongly about that anymore, especially if the feature can’t be disabled with a couple of lines. Here for example it makes sense because the CSS applied is substantial so it can’t easily be undone.

fregante (approved):

yakov116 (approved):

PR #3826, opened 2020-12-17
Meta: Standardize registration of CSS-only features

yakov116:
@fregante `clean-pinned-issues` was missing `awaitDomReady` so I we need to fix that before the next release.


fregante:
Good, but further repetition can be avoided with a new `features.addCssFeature(FeatureID, BooleanFunction[])`

Inside it, it just calls features.add. 

This would be a shortcut for actual CSS-only features, and not for features that have multiple loaders.

fregante:
No rush for the release by the way, I disabled the cronjob

source/features/index.tsx
```diff
@@ -244,6 +248,19 @@ const add = async (id: FeatureID, ...loaders: FeatureLoader[]): Promise<void> =>
...
+const addCssFeature = async (id: FeatureID, loaders: CssInternalRunConfig): Promise<void> => {
+	async function cssOnlyFeature(featureName: string): Promise<void> {
+		await Promise.resolve(); // The event fires a bit before the document body loaded
+		document.body.classList.add('rgh-' + featureName);
+	}
+
+	void add(id, {
+		...loaders,
+		awaitDomReady: false,
+		init: () => void cssOnlyFeature(id)
```
fregante, line 260: Just inline it, don’t call a secondary function.

yakov116: Forgot to make it a draft. I did not have time to finish it yet.

yakov116:
@fregante if you want to release sooner I have no issue if you want finish this up. If not I will do it a drop later.

yakov116:
Need typescript help

fregante:
The whole implementation is wrong, it’s not just the types. The function accepts an array but you’re using the rest operator onto an object. And also you’re passing single functions instead of arrays of functions

yakov116:
> And also you’re passing single functions instead of arrays of functions

I am not sure what you mean here

fregante:
The parameter type was `BooleanFunction[]`
you passed `BooleanFunction`

You fixed it

source/features/align-issue-labels.tsx
```diff
@@ -3,14 +3,4 @@ import * as pageDetect from 'github-url-detection';
...
-}
-
-void features.add(__filebasename, {
-	awaitDomReady: false,
-	include: [
-		pageDetect.isConversationList
-	],
-	init
-});
+void features.addCssFeature(__filebasename,	[pageDetect.isConversationList]);
```
fregante, line 6: There’s a tab instead of a space on this line in every file

source/features/index.tsx
```diff
@@ -244,6 +244,18 @@ const add = async (id: FeatureID, ...loaders: FeatureLoader[]): Promise<void> =>
 	}
 };
 
+const addCssFeature = async (id: FeatureID, pageDetections: BooleanFunction[]): Promise<void> => {
+	void add(id, {
+		include: [
+			...pageDetections
```
fregante, line 250: This variable should be called `include` and since it’s already an array you don’t need to use the rest parameter

fregante (changes requested):

source/refined-github.ts
```diff
@@ -24,18 +24,25 @@ import './features/align-repository-header.css';
 // DO NOT add CSS files here if they are part of a JavaScript feature.
 // Import the `.css` file from the `.tsx` instead.
 
+// CSS ONLY features
```
fregante, line 27:
```suggestion
// CSS-only disableable features
```

fregante:
Excellent!

PR #3827, opened 2020-12-17
Improve detection in `hide-useless-comments`

### LINKED ISSUE #3822: Add more keywords to `hide-useless-comments`
You can find great variety of new useless keywords on this issue: https://github.com/justinfrankel/licecap/issues/97

- ditto
- same
- same here
- me too
- ⬆️

With a period or exclamation mark at the end

cheap-glitch:
Thanks for contributing! 🍄

1. LINKED ISSUES: Closes #3822 

2. TEST URLS:
  - ["Please update" and variations](https://github.com/justinfrankel/licecap/issues/97)
  - ["+1" and variations](https://github.com/hayaku/hayaku/issues/303)
  - ["+1" and thumbs-ups](https://github.com/isaacs/github/issues/215)

This PR:
 * removes spaces and basic punctuation from the comments before testing them — it makes it easier to match small sentences like "Please update, thanks" and all their possible variations
 * adds several keywords and their variations
 * adds support for skin tone modifiers on hand-related emojis (it's relatively rare but I've seen them used a few times)

TODO:
- [x] Rewrite using `text.replace()`
- [x] Add tests

### DIFF
--- a/test/helpers.ts
+++ b/test/helpers.ts
@@ -15,6 +15,7 @@ import {
 	prCompareUrlRegex
 } from '../source/github-helpers';
 import {getParsedBackticksParts} from '../source/github-helpers/parse-backticks';
+import isUselessComment from '../source/helpers/useless-comments';
 
 test('getConversationNumber', t => {
 	const pairs = new Map<string, string | undefined>([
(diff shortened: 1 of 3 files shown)


fregante:
Thanks! It’s worth adding a comment about the skin modifiers. 

Also what do you think about:

```js
if (text.replace(everykeyword, '').trim() === '') (
  return
}
```

This should cover repetition without complicating the regex, and also avoid multiple regexes. 

It’d be also great to add some tests for this function

cheap-glitch:
@fregante: Building on your idea, how about:
```javascript
if (text.replace(everykeyword, '').trim().length < 20) {
  return
}
```
This would catch a lot of small variations (e.g. https://github.com/justinfrankel/licecap/issues/97#issuecomment-744489315, https://github.com/justinfrankel/licecap/issues/97#issuecomment-735165103) very easily. Ofc `20` is just off the top of my head, it should probably be adjusted.

fregante:
It's best to leave some comments visible than risk hiding useful comments; the comment needs to be a "perfect match" for us to be sure that it can be hidden. For this reason I'd adjust the regex to only specifically hide `me too` and not just `me`

test/helpers.ts
```diff
@@ -259,3 +260,21 @@ test('parseBackticks', t => {
...
+	t.true(isUselessComment('👍'));
+	t.true(isUselessComment('👍🏾'));
+	t.true(isUselessComment('me too'));
+	t.true(isUselessComment('please update!'));
+	t.true(isUselessComment('please update 🙏🏻'));
+	t.true(isUselessComment('Same here, please update, thanks'));
+	t.true(isUselessComment('Same here! Please update, thank you.'));
+
+	t.false(isUselessComment('+1\n<some useful information>'));
+	t.false(isUselessComment('Same here. <some useul information>'));
```
fregante, line 279:
```suggestion
	t.false(isUselessComment('Same here. <some useful information>'));
```

fregante (approved):
+1

yakov116:
@cheap-glitch keep up your great work. Wanted to give you a shout out on your super clear first posts!

PR #3837, opened 2020-12-22
Fix and improve `split-issue-pr-search-results`

### LINKED ISSUE #3824: The UI doesn't update when only showing issues (`split-issue-pr-search-results`)
1. Visit https://github.com/search?q=refined+github&type=Issues
2. Click Pull requests (and wait for load)
2. Click Issues

The UI isn't updated to reflect that now **only issues** are being shown. As you can see in the gif, the bug only appears _after_ seeing the Pull Requests:

gif

dertieran:
Fixes #3824 

Steps to test as mentioned in the issue:
1. Visit https://github.com/search?q=refined+github&type=Issues
2. Click Pull requests (and wait for load)
3. Click Issues

The problem was that `createUrl` always appended the next `is:<type>` query part.
This is also used by Github for the filter and every previous `is:<type>` is ignored.
But the detection did just check for the first occurence of `is:<type>` so the ui wasn't updated correctly.

I changed the code to first clear all `is:<type>` and then add the new filter.
I also improved the detection if multiple `is:<types>` are in the search.
As a bonus I also added that the language and state search parameters are also copied for each url

### DIFF
--- a/source/features/split-issue-pr-search-results.tsx
+++ b/source/features/split-issue-pr-search-results.tsx
@@ -25,22 +25,6 @@ function updateLinkElement(link: HTMLAnchorElement, type: GitHubConversationType
 	);
 }
 
-// Only the last is:pr/issue query part is used by github for the filter results so we find the one with the higher index
-function searchQueryIs(searchQuery: SearchQuery): GitHubConversationType | null {
-	const queryParts = searchQuery.getQueryParts();
-	const pr = queryParts.lastIndexOf('is:pr');
-	const issue = queryParts.lastIndexOf('is:issue');
-	if (pr > issue) {
-		return 'pr';
-	}
-
-	if (issue !== -1) {
-		return 'issue';
-	}
-
-	return null;
-}
-
 function init(): void {
 	cleanLinks();
 
@@ -58,12 +42,11 @@ function init(): void {
 	const title = select('.codesearch-results h3')!.firstChild!;
 
 	const searchQuery = new SearchQuery(location.search);
-	const is = searchQueryIs(searchQuery);
-	if (is === 'pr') {
+	if (searchQuery.includes('is:pr')) {
 		// Update UI in PR searches
 		issueLink.classList.remove('selected');
 		title.textContent = title.textContent!.replace('issue', 'pull request');
-	} else if (is !== 'issue' && searchQuery.searchParams.get('type') === 'Issues') {
+	} else if (!searchQuery.includes('is:issue') && searchQuery.searchParams.get('type') === 'Issues') {
 		// Update UI in combined searches (where there's no `is:<type>` query)
 		title.textContent = title.textContent!
 			.replace(/issue\b/, 'issue or pull request')
--- a/source/features/split-issue-pr-search-results.tsx
+++ b/source/features/split-issue-pr-search-results.tsx
@@ -12,30 +12,18 @@ function cleanLinks(): void {
 	}
 }
 
-function getPageSearchQuery(): SearchQuery {
-	const searchParameters = new URLSearchParams(location.search);
-	return new SearchQuery(searchParameters);
-}
-
 type GitHubConversationType = 'pr' | 'issue';
 
-function createUrl(type: GitHubConversationType, pathname = location.pathname): string {
-	const url = new URL(pathname, location.origin);
-	const searchQuery = getPageSearchQuery();
-	// Remove all existing is:pr/issue query parts, only the last is used anyways
+function updateLinkElement(link: HTMLAnchorElement, type: GitHubConversationType): void {
+	link.textContent = type === 'pr' ? 'Pull requests' : 'Issues'; // Drops any possible counter
+
+	const searchQuery = new SearchQuery(link);
 	searchQuery.remove('is:pr', 'is:issue');
 	searchQuery.add(`is:${type}`);
-	url.searchParams.set('q', searchQuery.get());
-	url.searchParams.set('type', 'Issues');
 
-	// Copy the language (l) and state (closed/open) search parameters if they are set
-	for (const name of ['l', 'state']) {
-		if (searchQuery.searchParams.has(name)) {
-			url.searchParams.set(name, searchQuery.searchParams.get(name)!);
-		}
-	}
-
-	return String(url);
+	link.append(
+		<include-fragment src={`${link.pathname}/count${link.search}`}/>
+	);
 }
 
 // Only the last is:pr/issue query part is used by github for the filter results so we find the one with the higher index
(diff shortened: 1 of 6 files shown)


fregante:
Good catch. The code is a little verbose but I haven’t looked exactly at how it can be improved

source/features/split-issue-pr-search-results.tsx
```diff
@@ -12,20 +12,51 @@ function cleanLinks(): void {
 	}
 }
 
+function getPageSearchQuery(): SearchQuery {
+	const searchParameters = new URLSearchParams(location.search);
+	return new SearchQuery(searchParameters);
```
fregante, line 17: `getPageSearchQuery` be dropped, SearchQuery accepts location
```suggestion
	return new SearchQuery(location);
```

source/features/split-issue-pr-search-results.tsx
```diff
@@ -12,20 +12,51 @@ function cleanLinks(): void {
...
+	searchQuery.add(`is:${type}`);
+	url.searchParams.set('q', searchQuery.get());
 	url.searchParams.set('type', 'Issues');
+
+	// Copy the language (l) and state (closed/open) search parameters if they are set
+	for (const name of ['l', 'state']) {
+		if (searchQuery.searchParams.has(name)) {
+			url.searchParams.set(name, searchQuery.searchParams.get(name)!);
+		}
+	}
```
fregante, line 36: Instead of copying them, you can set the whole search at once.

source/features/split-issue-pr-search-results.tsx
```diff
@@ -12,20 +12,51 @@ function cleanLinks(): void {
...
+
 type GitHubConversationType = 'pr' | 'issue';
 
 function createUrl(type: GitHubConversationType, pathname = location.pathname): string {
 	const url = new URL(pathname, location.origin);
-	url.searchParams.set('q', pageSearchQuery.get() + ` is:${type}`);
+	const searchQuery = getPageSearchQuery();
+	// Remove all existing is:pr/issue query parts, only the last is used anyways
+	searchQuery.remove('is:pr', 'is:issue');
+	searchQuery.add(`is:${type}`);
```
fregante, line 27: There’s a replace method that accepts a regex

source/features/split-issue-pr-search-results.tsx
```diff
@@ -47,11 +78,13 @@ function init(): void {
 	issueLink.after(prLink);
 
 	const title = select('.codesearch-results h3')!.firstChild!;
-	if (pageSearchQuery.includes('is:pr')) {
+
+	const is = searchQueryIs(searchQuery);
+	if (is === 'pr') {
```
fregante, line 83: Instead of this change, perhaps SearchQuery can handle deduplicating opposite flags (issue/pr for this bugfix; later for open/closed, merged/unmerged, etc)

source/features/split-issue-pr-search-results.tsx
dertieran: This only works if the `searchQuery` isn't modified otherwise `location.search` is also changed.
So it would not work for `createUrl`, didn't personally like that side effect from `SearchQuery` so explicitly created a new copy 😅

source/features/split-issue-pr-search-results.tsx
dertieran: My main thought here was to only copy things that are actually available/used.
E.g. this would avoid copying something like [`?package_type=npm`](https://github.com/search?package_type=npm&q=refined+github&type=RegistryPackages) which has no effect.
But if it is more readable I can also just copy all search params instead.

source/features/split-issue-pr-search-results.tsx
dertieran: Good idea, didn't think about using a regex 😅

source/features/split-issue-pr-search-results.tsx
dertieran: Sure why not 👍
I think `SearchQuery` just looked a bit scary to touch with it's side effects 😄

source/features/split-issue-pr-search-results.tsx
fregante: Indeed, I forgot how it worked. I think the issue here was that we're rebuilding both links from scratch (`createUrl`) starting from `location` instead of just changing the existing `Issues` link. Would you like to try that? This would also avoid the other comment (https://github.com/sindresorhus/refined-github/pull/3837/files#r547508824) since they'll already be in the link

dertieran: Done this :+1:

source/features/split-issue-pr-search-results.tsx
dertieran: When should the query be cleaned?
Right now it is only cleaned when it is `set`, so I added the deduplicate code there.
_(This allows to just use `add` [here](https://github.com/sindresorhus/refined-github/pull/3837#pullrequestreview-557362968))_

Should it also be cleaned on `get` then we could use the `includes` here.
Alternative would also be to add something linke `last(...searchStrings: string[]): string | null {}`.

fregante: As for the reloading, I think it should stop accepting a raw Location, perhaps it should be changed to use the `location.search` string, this way we eliminate the side effect. 

I’m conflicted about where this should go, because it should be a .clean()/.deduplicate() method to make it explicit, but also we don’t want to have to manually call it every time. I think they should be deduplicated:

- in the constructor, for existing values
- in set, for any new values 

So maybe you can implement it as a private method and call it from both places

dertieran: Okay also implemented this now, I just did `this.set(this.get())` to clean the initial query string.
This also means it will update the search for all passed *"sources"*.

fregante (approved):
Untested but it looks right!

Can you add some tests to ensure that the query is cleaned correctly? Also add some _tricky_ queries to ensure that they're not broken by mistake.

dertieran:
@fregante Sure, just not sure what a *"tricky"* query would be :sweat_smile: 
Right now I added a test that uses the query from the issue `?q=refined+github` and the `.add('is:issue')`.
I guess that's fairly simple, would tricky be something with multiple `is:pr/issue`?

Also I noticed that before the passed `URLSearchParams` where just assigned, I changed it that it now creates a new copy, should I change that back?

fregante:
I just realized that the regex is acting on the raw string like it did at the beginning of the PR. I’d rather getQueryParts as previously suggested

> SearchQuery was added specifically to avoid having to reparse the query with different code in every feature

dertieran:
> I just realized that the regex is acting on the raw string like it did at the beginning of the PR

So you mean something like this?
```js
set(query) {
	const parts = query.split(/\s+/).filter(Boolean);
	this.searchParams.set('q', deduplicate(parts).join(' '));
}
```

dertieran:
Okay this now are changes then I initially intended :sweat_smile:  
But I now also removed to `TODO` for supporting spaces in quoted query strings.

test/search-query.ts
```diff
@@ -4,34 +4,34 @@ import './fixtures/globals';
...
 
 test('.get', t => {
-	const query = new SearchQuery(new URLSearchParams('?q=wow'));
+	const query = new SearchQuery('?q=wow');
 	t.is(query.get(), 'wow');
 });
 
 test('.getQueryParts', t => {
-	const query = new SearchQuery(new URLSearchParams('?q=cool+is%3Aissue'));
+	const query = new SearchQuery('?q=cool+is%3Aissue');
```
fregante, line 12: I'd rather see a more readable
```suggestion
	const query = new SearchQuery(['q', 'cool is:issue']);
```

dertieran: Sure, so just accept all `URLSearchParams` options?
```js
URLSearchParams(init?: string | string[][] | Record<string, string> | URLSearchParams)
```

*Would be this then*
```js
new SearchQuery([['q', 'cool is:issue']]);
```

But it would also work to just pass the not escaped string if it is just for readability in the tests
```js
new SearchQuery('q=cool is:issue');
```

dertieran: I just changed `SearchQuery` to accept all `URLSearchParams` options.
And then used the `{q: 'query'}` parameter *(thought it was more readable than the entries option)*.

Hope that is okay otherwise I can also change it back or to something else.

source/github-helpers/search-query.ts
```diff
@@ -1,17 +1,43 @@
...
+		if (lastIndex > indexToKeep) {
+			indexToKeep = lastIndex;
+		}
+	}
+
+	if (indexToKeep === -1) {
+		return array;
+	}
+
+	return array.filter((value, index) => index >= indexToKeep || !values.includes(value));
```
fregante, line 24:
```suggestion
	let alreadyFound = false;
	const newArray = []
	for (const item of array.reverse()) {
		const found = values.includes(item);
		if (!alreadyFound || !found) {
			newArray.unshift(item)
		}
		
		alreadyFound = alreadyFound || found;
	}

	return newArray;
```

dertieran: I think this might be a bit *"slower"* in some edgcases but it's way more readable :+1:
Thanks :heart_eyes:

fregante:
Thanks for fixing this!

PR #3841, opened 2020-12-24
Reposition `tag-changelog-link`

### LINKED ISSUE #2699: Reposition `tag-changelog-link`
GitHub [implemented](https://github.blog/changelog/2020-01-13-shortcut-to-compare-across-two-releases/) a more generic (i.e. less efficient) version of `tag-changelog-link`

Their feature is not a strict replacement, but now `tag-changelog-link` needs to be moved above it (or... alternative solutions welcome)

## Test

https://github.com/sindresorhus/refined-github/releases/tag/20.1.6

cheap-glitch:
Thanks for contributing! 🍄

1. LINKED ISSUES: Fixes #2699 

2. TEST URLS:
  * [Single release](https://github.com/sindresorhus/refined-github/releases/tag/20.1.6)
  * [Multiple releases](https://github.com/facebook/react/releases)
  * [Tags in release list](https://github.com/facebook/react/releases?after=v16.7.0)
  * [Tags list](https://github.com/facebook/react/tags)
  * [Release with "Verified" badge](https://github.com/tensorflow/tensorflow/releases?after=v1.12.0-rc1)

3. SCREENSHOT:
<table>
<tr>
	<td><strong>Before</strong>
	<td><strong>After</strong>
<tr>
	<td>
	<td>
</table>

This PR:
  * Moves the "Changelog" link above the GH dropdown. It relies on some GH class names so it might not be the best approach, but it keeps things simple.
  * Mitigates the content jump by reserving some space with CSS (there is still a jump but it happens much sooner in the page load). The magic number is not great though.
  * cleans up some outdated code regarding the "Verified" label (it's now properly wrapped inside a `<li>` element)

### DIFF
--- a/source/features/tag-changelog-link.css
+++ b/source/features/tag-changelog-link.css
@@ -1,5 +1,6 @@
 /* Reserve some space above the "Compare" dropdown (the selector will be disabled as soon as the "Changelog" link is inserted) */
 .rgh-tag-changelog-link .release:not(.label-draft) .list-style-none > .d-block:nth-child(3):not(.rgh-changelog-link) {
-	/* Use padding to avoid overwriting the margin already present on the element */
+	/* Use the padding to avoid overwriting the margin already present on the element
+	 * The height in pixels is taken from the commit link element (https://user-images.githubusercontent.com/1402241/103394704-bf2d7a80-4aef-11eb-8377-5c59b46cc2b3.png) */
 	padding-top: 21px !important;
 }
--- a/source/features/tag-changelog-link.css
+++ b/source/features/tag-changelog-link.css
@@ -1,0 +1,3 @@
+.rgh-tag-changelog-link .release:not(.label-draft) .list-style-none > .d-block:nth-child(3):not(.rgh-changelog-link) {
+	padding-top: 21px !important;
+}
--- a/source/features/tag-changelog-link.css
+++ b/source/features/tag-changelog-link.css
@@ -1,6 +1,5 @@
 /* Reserve some space above the "Compare" dropdown (the selector will be disabled as soon as the "Changelog" link is inserted) */
 .rgh-tag-changelog-link .release:not(.label-draft) .list-style-none > .d-block:nth-child(3):not(.rgh-changelog-link) {
-	/* Use the padding to avoid overwriting the margin already present on the element
-	 * The height in pixels is taken from the commit link element (https://user-images.githubusercontent.com/1402241/103394704-bf2d7a80-4aef-11eb-8377-5c59b46cc2b3.png) */
+	/* 21px is the height of the link that is being inserted https://user-images.githubusercontent.com/1402241/103394704-bf2d7a80-4aef-11eb-8377-5c59b46cc2b3.png */
 	padding-top: 21px !important;
 }
--- a/source/features/tag-changelog-link.css
+++ b/source/features/tag-changelog-link.css
@@ -1,3 +1,4 @@
+/* Reserve some space above the "Compare" dropdown (the selector will be disabled as soon as the "Changelog" link is inserted) */
 .rgh-tag-changelog-link .release:not(.label-draft) .list-style-none > .d-block:nth-child(3):not(.rgh-changelog-link) {
 	padding-top: 21px !important;
 }
--- a/source/features/tag-changelog-link.css
+++ b/source/features/tag-changelog-link.css
@@ -1,4 +1,5 @@
 /* Reserve some space above the "Compare" dropdown (the selector will be disabled as soon as the "Changelog" link is inserted) */
 .rgh-tag-changelog-link .release:not(.label-draft) .list-style-none > .d-block:nth-child(3):not(.rgh-changelog-link) {
+	/* Use padding to avoid overwriting the margin already present on the element */
 	padding-top: 21px !important;
 }
--- a/source/features/tag-changelog-link.tsx
+++ b/source/features/tag-changelog-link.tsx
@@ -110,6 +110,8 @@ async function init(): Promise<void> {
 					</a>
 				</li>
 			);
+			/* Fix spacing issue when the window is < 700px wide https://github.com/sindresorhus/refined-github/pull/3841#issuecomment-754325056 */
+			lastLink.classList.remove('flex-auto');
 		}
 	}
 }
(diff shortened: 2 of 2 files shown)


source/features/tag-changelog-link.css
```diff
@@ -0,0 +1,3 @@
+.rgh-tag-changelog-link .release:not(.label-draft) .list-style-none > .d-block:nth-child(3):not(.rgh-changelog-link) {
```
fregante, line 1: What is this targeting? Why 21 exactly? Can we add it via class with JS instead?

cheap-glitch: > What is this targeting?

The third list element in the little sidebar portrayed in the screenshots (so either the "Changelog" dropdown or the "Verified" label). When inserted, the Changelog link will become the third element and the padding will disappear because of `:not(.rgh-changelog-link)`

> Why 21 exactly?

It's the computed height of the link element. Basically a magic number that should be replaced with something better.

> Can we add it via class with JS instead?

Probably, but I thought a pure CSS solution would apply itself faster during page load, and thus alleviate the content jump the most.

Another solution could be to systematically inject an empty list element (or some dummy text) before computing the tags, and then either remove it or replace its content with the link.

Or we could just ignore the content jump (imo it's not that bad) and drop this part altogether.

fregante: Ohhh I see what you’re doing. You’re avoiding the content jump, so a CSS-only solution is indeed better. However it needs a comment, use margin, and probably a better selector (for example you can target the second element and user margin-bottom instead of using :not)

fregante: Ohhh I see what you’re doing. You’re avoiding the content jump, so a CSS-only solution is indeed better. However it needs a comment, use margin, and probably a better selector (for example you can target the second element and user margin-bottom instead of using :not)

cheap-glitch: The idea with the `:not()` is to have the selector immediately be disabled when the "Changelog" link is inserted. There's probably a better way to do this, but I haven't found one.
Also I used `padding` because the dropdown already has some margin and we want to add to it, not override it.

fregante: So it just needs a comment explaining what's going on :D

cheap-glitch: I'd like to get rid of the "magic" `21px`, but I can't find a pure CSS replacement for it.
Any ideas @yakov116?

yakov116: I wont have time until friday to look. But don't count on me finding anything ☕

fregante: If GitHub had size variables you could use them in a `calc`, but I don't think it does. With JavaScript you could place a whole hidden element while the changelog link loads, but there will be a FOUC.

The hard-coded value is fine as long as you show how you got it. For example include this screenshot
Screen Shot

fregante: By the way what you're doing is called "Avoiding layout shift" https://web.dev/cls/

source/features/tag-changelog-link.tsx
```diff
@@ -87,27 +90,22 @@ async function init(): Promise<void> {
...
-							<DiffIcon/> Changelog
-						</a>
-					</li>
-				);
-
-				// `lastLink` is no longer the last link, so it shouldn't push our new link away.
-				// Same page as before: https://github.com/tensorflow/tensorflow/releases?after=v1.12.0-rc1
-				lastLink.classList.remove('flex-auto');
-			}
+		for (const lastLink of select.all('.list-style-none > .d-block:nth-child(2), .list-style-none > .d-inline-block:last-child', container.element)) {
```
fregante, line 97: Can you comment each selector?
```suggestion
		const lastLinks = select.all([
			'.list-style-none > .d-block:nth-child(2)', // XXX
			'.list-style-none > .d-inline-block:last-child' // YYY
		], container.element);
		for (const lastLink of lastLinks) {
```

source/features/tag-changelog-link.tsx
```diff
@@ -87,27 +90,22 @@ async function init(): Promise<void> {
 
 	for (const [index, container] of allTags.entries()) {
 		const previousTag = getPreviousTag(index, allTags);
+		if (!previousTag) {
+			return;
```
fregante, line 94:
```suggestion
			continue;
```

source/features/tag-changelog-link.css
```diff
@@ -0,0 +1,6 @@
+/* Reserve some space above the "Compare" dropdown (the selector will be disabled as soon as the "Changelog" link is inserted) */
+.rgh-tag-changelog-link .release:not(.label-draft) .list-style-none > .d-block:nth-child(3):not(.rgh-changelog-link) {
+	/* Use the padding to avoid overwriting the margin already present on the element
+	 * The height in pixels is taken from the commit link element (https://user-images.githubusercontent.com/1402241/103394704-bf2d7a80-4aef-11eb-8377-5c59b46cc2b3.png) */
```
fregante, line 4:
```suggestion
	/* 21px is the height of the link that is being inserted https://user-images.githubusercontent.com/1402241/103394704-bf2d7a80-4aef-11eb-8377-5c59b46cc2b3.png */
```

fregante:
The feature needs `awaitDomReady: false` or else the body class it added too late. This **might** need `elementReady` somewhere as well.

fregante:
Screen Shot

The link appears centered in when the window is <700px wide

fregante:
Thank you! That issue was about to turn 366 days old 💣

fregante:
It's probably not caused by this change, but the feature seems broken. I also just cleared the cache in the options as well.

https://github.com/sindresorhus/refined-github/releases

```
 Refined GitHub → tag-changelog-link → (2)
TypeError: null is not an object (evaluating '/\/releases\/tag\/(.*)/.exec(tagUrl)[1]')
parseTags — refined-github.js:5843:127
map
(anonymous function) — refined-github.js:5923:138
asyncFunctionResume
[native code]
promiseReactionJobWithoutPromise
```

 
Screen Shot 2

yakov116:
I am not seeing the error

cheap-glitch:
I just updated the extension and it's working fine for me too (tested on Firefox and Chromium).

Issue #3814, opened 2020-12-11
Show entire conversation by default


hyperfekt:
For very long conversations, GitHub hides everything but the beginning and end by default. Not only is it very annoying to keep expanding new items to read through a conversation, it also makes it very easy to miss comments when using the browser's search function.
Refined GitHub should automatically display all items in a conversation.

An example of the problem can be found [here](https://github.com/isaacs/github/issues/50#:~:text=hidden%20items) (search for 'hidden items').

yakov116:
You can press ALT and click on unhide and it will unhide 200 items

hyperfekt:
For me it's only 60?

fregante:
Showing long discussions is relatively heavy and more often than not I’m not interested in “middle” messages in conversations with hundreds of comments. If I am, I’ll expand the discussion. 

As Yakov mentioned, we have a feature that expands the entire conversation on alt-click, but it’s possible that it’s not working at the moment.

hyperfekt:
Hm, I presume there isn't space for disabled-by-default features in this extension, given that the preference for our against this is variable. In that case a userscript is probably more appropriate.

Issue #615, opened 2017-07-12
Collapse large sequences of label, milestone, etc changes


aaronbushnell:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%2450%20Funded-%2300A156.svg" />](https://issuehunt.io/r/sindresorhus/refined-github/issues/615)

I apologize if this has already been mentioned, but I did not see an issue for it.

Sometimes issues get very noisy when labels and milestones are added/removed. They push comments down and make it difficult to see updates to the item that aren't related to the project management side.

It would be nice if these large sequences (excluding comments) could "collapse". Here's an example of what I'm describing.

![screenflow](https://user-images.githubusercontent.com/315202/28134893-677e8e60-6712-11e7-9228-f9c8da1b26a5.gif)

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $50.00)

- [<img src='https://avatars3.githubusercontent.com/u/44827199?v=4' alt='issuehunt' width=24 height=24> issuehunt](https://issuehunt.io/u/issuehunt) ($50.00)

### Submitted pull Requests
- [#4029 Add `conversation-activity-filter` feature](https://issuehunt.io/r/sindresorhus/refined-github/pull/4029)
---

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/sindresorhus/refined-github/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

bfred-it:
The name changes are definitely annoying so I'd put a vote to hide _those_, but references from other issues are useful.

jgierer12:
I also think references and commits should stay expanded. But hiding title changes, deploys and label changes would be great. Some things like assignments don't occur very often per issue, so I think they should stay expanded too.

---

Hidden events should be as unobstrusive as possible IMO, perhaps something like this would work:
![Hidden events mockup](https://user-images.githubusercontent.com/4331946/28176188-c989f696-67f6-11e7-8062-46e4af70d80d.png)
Clicking the "+" would expand the hidden events

hkdobrev:
I'd prefer collapsing title changes with title changes, label changes with label changes, milestone changes with milestone changes, project changes with project changes, reviewers changes with reviewers changes and assignees changes with assignees changes.

jgierer12:
[...]

PR #3847, opened 2020-12-27
Add `conversation-timeline-filter` feature

### LINKED ISSUE #3764: Toggle/Hide TimelineItem elements (e.g. added label etc)
I have searched the issues and installed the extension but don't see functionality for hiding `TimelineItem`s. 

My request is to be able to hide/toggle all Timelineitems from an issue or pull request view/page. 

image

GitLab has this built-in and I think GitHub should too, but next best thing would be for Refined GitHub to have this feature:
image

shoter:
1. LINKED ISSUES:
   Closes #3764

2. TEST URLS:
   https://github.com/dotnet/aspnetcore/pull/13040 - Contains all cases that I already found on github
   https://github.com/dotnet/aspnetcore/pull/28939 ( https://github.com/dotnet/aspnetcore/pull/28939#discussion_r551490178 shows comment attached to review that is unresolvable as it is attached to comment from previous review)

3. SCREENSHOT:
	(I already fixed top-bar display issue above unresolved comments)
	filter

4. Description

Creates new functionality that let user filter not needed things on pull request timeline.
Can filter following things:
- Unresolved comments (default: Show)
- Resolved comments (default: Hide)
- Normal comments (comment that cannot be resolved) (default: Show)
- Commits (default: Show)
- Other (default: Show)

Additionally to quickly see all things in given timeline it has option to auto load hidden items - by default off.
I tested it on Github and Github Enterprise.

It is my first time contributing to this repo so if there is code duplication or I do not adhere to some standard then please correct me to do things right :).


source/features/pull-request-timeline-filter.tsx
```diff
@@ -0,0 +1,261 @@
...
+import select from 'select-dom';
+import elementReady from 'element-ready';
+import {observe} from 'selector-observer';
+import * as pageDetect from 'github-url-detection';
+
+import features from '.';
+
+async function sleep(s: number): Promise<any> {
+	await new Promise(resolve => setTimeout(resolve, s * 1000));
+}
```
shoter: How should I fix lint issue here?

fregante: We already use a package called `delay`

fregante:
I’m hesitant to accept large UIs, but this appears relatively compact for what it does and it could be a solution to https://github.com/sindresorhus/refined-github/issues/615 (but not by default). 

Do you have time for the lengthy reviews that I expect on this bad boy? Some code can be dropped because we already have handlers/patterns for GitHub UI changes, some can be improved, and I think this could work better if implemented to look exactly like the label selector (a dropdown instead of a modal)

source/features/pull-request-timeline-filter.tsx
fregante: But this part of the feature should be dropped as it was already rejected: https://github.com/sindresorhus/refined-github/issues/3814

readme.md
```diff
@@ -206,6 +206,7 @@ Thanks for contributing! 🦋🙌
 - [](# "pr-jump-to-first-non-viewed-file") [Jumps to first non-viewed file in a pull request when clicking on the progress bar.](https://user-images.githubusercontent.com/16872793/85226580-3bf3d500-b3a6-11ea-8494-3d9b6280d033.gif)
 - [](# "git-checkout-pr") [Adds copy-pastable git commands to checkout a PR.](https://user-images.githubusercontent.com/16872793/96938908-8e024f80-1499-11eb-8976-0caf95175dd6.png)
 - [](# "jump-to-change-requested-comment") [Adds a link to jump to the latest changed requested comment.](https://user-images.githubusercontent.com/19198931/98718312-418b9f00-23c9-11eb-8da2-dfb616e95eb6.gif)
+- [](# "pull-request-timeline-filter") [Adds filters to hide unwanted things on pull request and adds ability to load all hidden items on PR page.](https://user-images.githubusercontent.com/3819892/103170243-2d014a00-4843-11eb-8a32-7e28eb755f00.gif)
```
fregante, line 209: I think this feature can also apply to issues, so it should be named `conversation-timeline-filter`. Any PR-related checkbox would have to be hidden in issues.

source/features/pull-request-timeline-filter.tsx
```diff
@@ -0,0 +1,260 @@
...
+
+const CurrentSettings: FilterSettings =
+{
+	HideResolved: true,
+	HideUnresolved: false,
+	hideNormalComment: false,
+	HideOthers: false,
+	HideCommits: false,
+	AutoLoadHidden: false
+};
```
fregante, line 30: You should be able to read the values of each checkbox by referring to their `form` element and using FormData, I think, to read the values. For example, instead of having this global, have a function that returns `new FormData(select(the form))` and then access its values with `.get(id)`

source/features/pull-request-timeline-filter.tsx
```diff
@@ -0,0 +1,260 @@
...
+
+function regenerateFilterSummary(): void {
+	const timelineFilter = select(`#${timelineFiltersSelectorId}`)!;
+	const newSummary = (
+		<p className="reason text-small text-gray">
+			{CurrentSettings.HideUnresolved ? 'Hide' : 'Show'} unresolved comments. <br/>
+			{CurrentSettings.HideResolved ? 'Hide' : 'Show'} resolved comments. <br/>
+			{CurrentSettings.hideNormalComment ? 'Hide' : 'Show'} normal comments. <br/>
+			{CurrentSettings.HideOthers ? 'Hide' : 'Show'} other items. <br/>
+			{CurrentSettings.HideCommits ? 'Hide' : 'Show'} commits. <br/>
```
fregante, line 53: Like for labels, this section should only show non-default options, that is: only show these lines if they are “Hide”

source/features/pull-request-timeline-filter.tsx
```diff
@@ -0,0 +1,260 @@
...
+	const normalComment = select('.js-comment-container', item);
+
+	if (pr) {
+		processPR(item);
+	} else if (commitGroup) {
+		applyDisplay(item, CurrentSettings.HideCommits);
+	} else if (normalComment) {
+		applyDisplay(item, CurrentSettings.hideNormalComment);
+	} else {
+		applyDisplay(item, CurrentSettings.HideOthers);
```
fregante, line 202: What do you think about this:

- process each item once, only to add a class (e.g. `rgh-is-review-comment`)
- when the settings change, only change a single class on the timeline container (`.rgh-hide-reviews .rgh-is-review-comment {display: none}`
- avoid processing items at all if they’re already targetable without the custom class: `.rgh-hide-reviews .js-comment[id^=pullrequestreview] {display: none}`

source/features/pull-request-timeline-filter.tsx
```diff
@@ -0,0 +1,260 @@
...
+	applyDisplay(item, !hasVisibleElement);
+}
+
+async function init(): Promise<any> {
+	await addTimelineItemsFilter();
+
+	// There are some cases when github will remove this filter. In that case we need to add it again.
+	// Example: Editing comment will make timeline filter to disappear.
+	observe(`#${timelineFiltersSelectorId}`, {
+		async remove() {
```
fregante, line 236: If you use `add: addTimelineItemsFilter` instead you can avoid having to call the function at all, even line 231 can be dropped. I *think* that this is how we manage the sidebar updates, you can double-check other sidebar features

source/features/pull-request-timeline-filter.tsx
```diff
@@ -0,0 +1,260 @@
...
+			await addTimelineItemsFilter();
+		}
+	});
+
+	observe(timelineItemSelector, {
+		add(element) {
+			const htmlElement = element as HTMLElement;
+			processTimelineItem(htmlElement);
+		}
+	});
```
fregante, line 246: You can use `on-new-comments` instead

source/features/pull-request-timeline-filter.tsx
```diff
@@ -0,0 +1,260 @@
...
+	// Close button should restore previous settings.
+	select('div.Box-header button', detailsDialog)!.addEventListener('click', restoreSettings);
+
+	const form = <div/>;
+
+	createItem(form, hideResolvedSelectorId, 'Hide resolved comments', '', CurrentSettings.HideResolved, false);
+	createItem(form, hideCommitsSelectorId, 'Hide commits', '', CurrentSettings.HideCommits, true);
+	createItem(form, hideUnresolvedSelectorId, 'Hide unresolved comments', '', CurrentSettings.HideUnresolved, true);
+	createItem(form, hideNormalCommentsSelectorId, 'Hide normal comments', 'Hides comments that does not contain unresolved/resolved state.', CurrentSettings.hideNormalComment, true);
+	createItem(form, hideOthersSelectorId, 'Hide other', 'Hides any other kind of activity that was not specified above.', CurrentSettings.HideOthers, true);
```
fregante, line 161: Instead of passing `form`, you could build the whole form in a single JSX by returning the generated element:

``` JSX
const form = (
  <form>
    {createItem(etc)}
    {createItem(etc)}
  
   </form>
)
```

fregante (changes requested):

source/features/pull-request-timeline-filter.tsx
fregante: Alternatively they can also be read with `form.fields[name].checked`

source/features/pull-request-timeline-filter.tsx
fregante: This change can come later, let’s fix the UI and rest of the code first.

source/features/pull-request-timeline-filter.tsx
```diff
@@ -0,0 +1,260 @@
...
+
+	// Close button should restore previous settings.
+	select('div.Box-header button', detailsDialog)!.addEventListener('click', restoreSettings);
+
+	const form = <div/>;
+
+	createItem(form, hideResolvedSelectorId, 'Hide resolved comments', '', CurrentSettings.HideResolved, false);
+	createItem(form, hideCommitsSelectorId, 'Hide commits', '', CurrentSettings.HideCommits, true);
+	createItem(form, hideUnresolvedSelectorId, 'Hide unresolved comments', '', CurrentSettings.HideUnresolved, true);
+	createItem(form, hideNormalCommentsSelectorId, 'Hide normal comments', 'Hides comments that does not contain unresolved/resolved state.', CurrentSettings.hideNormalComment, true);
```
fregante, line 160: I think we can consider all comments as "unresolved" until they're hidden or resolved.

- Resolved/hidden comments
- Unresolved comments
- Commits (only on PRs)
- Meta events (instead of "other")
I think that by default all of these should be shown because hiding user content might cause confusion

shoter: In my opinion it's very useful to have this special distinction for 3rd type of comments. It's very helpful in PRs with a lot of comments where you only want to see what unresolved problems you have left.

Look at PR that is a test url in this PR. Trying to find unresolved comments is pretty hard there, if there would be no option to just hide normal comments.

Changing name to meta events and not hiding anything by default is ok.

I will try to make those changes today/tomorrow.
Thanks for Review.

fregante: Makes sense. Then maybe:
- Hidden comments (on Issues) | Resolved reviews (on PRs, even if it also targets hidden comments)
- Unresolved reviews (PR only)
- Regular comments
- Commits (PR only)
- Meta events

fregante: Also perhaps commits should be rolled into the other events:

- Showing _only_ commits makes no sense, there's a dedicated view for that.
- Hiding commits but leaving other events also makes little sense IMHO

fregante: Sorry, thinking out loud, but I had another idea 😅

What if was a single option (same display, except only one can be selected, like a `<select>`) among:

- [x] **Show all**
- [ ] Only show comments
  _Hides commits and events_
- [ ] Only show unresolved comments
  _Also hides resolved reviews and hidden comments_
- [ ] Only show unresolved reviews
  _Also hides regular comments (PR only)_

shoter: Good ideas. I will implement them.

fregante:
I think this feature can be useful, ping if you have time

readme.md
shoter: Done

source/features/pull-request-timeline-filter.tsx
shoter: Done

source/features/pull-request-timeline-filter.tsx
shoter: Done

source/features/pull-request-timeline-filter.tsx
shoter: `on-new-comments` does not return newly added element in callback. I think it's better to stay with observer so I am not iterating whole page over and over again.

source/features/pull-request-timeline-filter.tsx
shoter: Done

source/features/pull-request-timeline-filter.tsx
shoter: I removed this feature.

source/features/pull-request-timeline-filter.tsx
shoter: I do not know what you mean. Could you write short example or show me file where it is used?

source/features/pull-request-timeline-filter.tsx
shoter: I implemented the idea. The only thing left is hidden comments.

shoter:
Sorry for delay. I was super busy lately. I hope that we will get this review to the end together.

From what I see I've got only hiding hidden comments to be implemented.

I would not like to implement labels like UI. It's quite troublesome to understand how this UI work and at the same time I think current approach looks nice. Also labels functionality is used to select multiple items whereas notification like UI is used to select 1 item from the list. Therefore I think it's better to leave notification popup like UI because we are also selecting only 1 thing from the list.

source/features/conversation-timeline-filter.tsx
```diff
@@ -0,0 +1,225 @@
...
+async function init(): Promise<any> {
+	await addTimelineItemsFilter();
+
+	// There are some cases when github will remove this filter. In that case we need to add it again.
+	// Example: Editing comment will make timeline filter to disappear.
+	observe(`#${timelineFiltersSelectorId}`, {
+		async remove() {
+			await addTimelineItemsFilter();
+		}
+	});
```
fregante, line 210: There must be a way to merge this. Probably you can just observe the closest sidebar element instead
```suggestion
	observe(`#whatever sidebar`, {
		add() {
			void addTimelineItemsFilter();
		}
	});
```

source/features/conversation-timeline-filter.tsx
```diff
@@ -0,0 +1,225 @@
...
+	}
+
+	const newSummary = (
+		<p className="reason text-small text-gray">
+			{text}
+		</p>
+	);
+
+	select('p.reason', timelineFilter)!.replaceWith(newSummary);
+}
```
fregante, line 56: There must be a better way to handle this, switch-less and enum-less. For example:


```js
const copy = {
	ShowAll: 'Show all',
	ShowOnlyUnresolvedComments: 'Show unresolved comments',
};

function regenerateFilterSummary(value) {
  	select('a single selector here .reason')!.textContent = copy[value];
}
```

fregante:
> I think it's better to leave notification popup like UI because we are also selecting only 1 thing from the list.

I think not. There's no reason for this to be a lightbox. Instead of cloning and adapting the labels dropdown however you can just generate the whole block, it's probably more efficient: https://primer.style/css/components/select-menu#list-items

source/features/conversation-timeline-filter.tsx
```diff
@@ -0,0 +1,225 @@
...
+
+	detailsDialog.setAttribute('aria-label', 'Timeline filter settings');
+	select('div.Box-header h3', detailsDialog)!.textContent = 'Timeline filter settings';
+
+	let showUnresolvedReviews = null;
+	if (pageDetect.isPRConversation()) {
+		showUnresolvedReviews = createRadio(showFilterName, 'Show only unresolved reviews', 'Also hides regular comments (PR only)', FilterSettings.ShowOnlyUnresolvedReviews, true);
+	}
+
+	const form = (
```
fregante, line 131:
```suggestion
	detailsDialog.append(
```

source/features/conversation-timeline-filter.tsx
```diff
@@ -0,0 +1,225 @@
...
+
+		// We need to hide whole thread group if we have hidden all comments inside.
+		hasVisibleElement = hasVisibleElement || threadContainer.style.display === '';
+	}
+
+	if (hasVisibleElement) {
+		item.style.display = '';
+	} else {
+		item.style.display = 'none';
+	}
```
fregante, line 190: You can replace every usage of the `display` style with the `hidden` property, without if/else

```suggestion
	item.hidden = !hasVisibleElement
```

source/features/conversation-timeline-filter.tsx
```diff
@@ -0,0 +1,225 @@
...
+function processTimelineItem(item: HTMLElement): void {
+	const pr = select('.js-comment[id^=pullrequestreview]', item);
+	const normalComment = select('.js-comment-container', item);
+
+	if (pr) {
+		processPR(item);
+	} else if (normalComment) {
+		applyDisplay(item, FilterSettings.ShowOnlyComments, FilterSettings.ShowOnlyUnresolvedComments);
+	} else {
+		applyDisplay(item, FilterSettings.ShowAll);
```
fregante, line 161:
```suggestion
	if (select.exists('.js-comment[id^=pullrequestreview]', item)) { // PR review thread
		processPR(item);
	} else if (select.exists('.js-comment-container', item)) { // Regular comments
		applyDisplay(item, FilterSettings.ShowOnlyComments, FilterSettings.ShowOnlyUnresolvedComments);
	} else { // Other events
		applyDisplay(item, FilterSettings.ShowAll);
```

source/features/conversation-timeline-filter.tsx
```diff
@@ -0,0 +1,225 @@
...
+
+function processPR(item: HTMLElement): void {
+	let hasVisibleElement = false;
+
+	for (const threadContainer of select.all('.js-resolvable-timeline-thread-container', item)) {
+		const commentContainer = select('.inline-comment-form-container', threadContainer);
+
+		if (threadContainer.getAttribute('data-resolved') === 'true') {
+			applyDisplay(threadContainer, FilterSettings.ShowOnlyComments);
+		} else if (commentContainer === null) {
```
fregante, line 173:
```suggestion
		if (threadContainer.getAttribute('data-resolved') === 'true') {
			applyDisplay(threadContainer, FilterSettings.ShowOnlyComments);
		} else if (!select.exists('.inline-comment-form-container', threadContainer)) {
```

source/features/conversation-timeline-filter.tsx
```diff
@@ -0,0 +1,225 @@
...
+			// There is 1 special case here when github shows you a comment that was added to previous comment thread but it does not show whether it is resolved or not resolved comment.
+			// It's kinda tricky to know what to do with this so it is marked as normal comment for meantime.
+			// We are just checking here if user is able to comment inside that timeline thread, if not then it means we have this special situation that was just described.
+			applyDisplay(threadContainer, FilterSettings.ShowOnlyComments, FilterSettings.ShowOnlyUnresolvedComments);
```
fregante, line 177: Can you post a link to such case? This way it can be also found and tested in the future

fregante (changes requested):

source/features/pull-request-timeline-filter.tsx
fregante: I left a new review for this

fregante:
I implemented it as a select and the code is already shorter (probably some elements and classes can still be dropped/merged).

Screen Shot 3

However saveSettings/restoreSettings need to be updated.

source/features/conversation-timeline-filter.tsx
shoter: https://github.com/dotnet/aspnetcore/pull/28939#discussion_r551490178
I also posted it on the top.

source/features/conversation-timeline-filter.tsx
shoter: Done

source/features/conversation-timeline-filter.tsx
```diff
@@ -0,0 +1,225 @@
...
+let currentSettings: FilterSettings = FilterSettings.ShowAll;
+
+const showFilterName = 'rgh-show-filter';
+const filterFormId = 'show-filter-form';
+
+// Every element on the timeline that is recognizable by this feature will be marked with tis class.
+const timelineFiltersSelectorId = 'timeline-filters';
+const detailsSelector = `#${timelineFiltersSelectorId} details`;
+const notiticationsSelector = '.discussion-sidebar-item.sidebar-notifications';
+const timelineItemSelector = '.js-timeline-item';
```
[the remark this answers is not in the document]
shoter: Done

source/features/conversation-timeline-filter.tsx
```diff
@@ -0,0 +1,225 @@
...
+	}
+
+	// Copy existing element and adapt its content
+	const timelineFilter = notifications.cloneNode(true);
+	timelineFilter.id = 'timeline-filters';
+
+	select('form.thread-subscribe-form', timelineFilter)!.remove();
+	const summary = select('summary', timelineFilter)!;
+	summary.setAttribute('aria-label', 'Customize timeline filters');
+	summary.addEventListener('click', restoreSettings);
```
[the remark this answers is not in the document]
shoter: No longer applies.

source/features/conversation-timeline-filter.tsx
```diff
@@ -0,0 +1,225 @@
...
+	regenerateFilterSummary();
+	reapplySettings();
+}
+
+function reapplySettings(): void {
+	select
+		.all(timelineItemSelector)
+		.forEach(element => {
+			processTimelineItem(element);
+		});
```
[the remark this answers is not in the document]
shoter: Done

source/features/conversation-timeline-filter.tsx
shoter: I tried to do that but this was not working. It was not hiding/showing things.

source/features/conversation-timeline-filter.tsx
shoter: Done. It looks neat.

source/features/conversation-timeline-filter.tsx
shoter: Done

source/features/conversation-timeline-filter.tsx
```diff
@@ -0,0 +1,225 @@
...
+const notiticationsSelector = '.discussion-sidebar-item.sidebar-notifications';
+const timelineItemSelector = '.js-timeline-item';
+
+function regenerateFilterSummary(): void {
+	const timelineFilter = select(`#${timelineFiltersSelectorId}`)!;
+
+	let text = '';
+	switch (currentSettings) {
+		case FilterSettings.ShowAll:
+			text = 'Show All';
```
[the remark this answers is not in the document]
shoter: Done

source/features/conversation-timeline-filter.tsx
shoter: Done

source/features/conversation-timeline-filter.tsx
```diff
@@ -0,0 +1,174 @@
...
+}
+
+async function saveSettings(filterSettings: FilterSettings, test: string): Promise<any> {
+	currentSettings = filterSettings;
+	regenerateFilterSummary();
+	reapplySettings();
+
+	select.all('.' + menuItemCheckbox)
+		.forEach(element => {
+			element.style.display = 'none';
```
fregante, line 42: Use the hidden property everywhere instead of this 
```suggestion
			element.hidden = true
```

source/features/conversation-timeline-filter.tsx
```diff
@@ -0,0 +1,174 @@
...
+	select(`#${timelineFiltersSelectorId}`)!.textContent = summaries[currentSettings];
+}
+
+async function saveSettings(filterSettings: FilterSettings, test: string): Promise<any> {
+	currentSettings = filterSettings;
+	regenerateFilterSummary();
+	reapplySettings();
+
+	select.all('.' + menuItemCheckbox)
+		.forEach(element => {
```
fregante, line 41: Basically never use forEach, we prefer for-of

source/features/conversation-timeline-filter.tsx
```diff
@@ -0,0 +1,174 @@
...
+const menuItemCheckbox = 'rgh-filter-menu-item-checkbox';
+
+// Every element on the timeline that is recognizable by this feature will be marked with tis class.
+const timelineFiltersSelectorId = 'timeline-filters';
+
+const summaries = {
+	[FilterSettings.ShowAll]: '',
+	[FilterSettings.ShowOnlyComments]: 'Show only comments',
+	[FilterSettings.ShowOnlyUnresolvedComments]: 'Show only unresolved reviews',
+	[FilterSettings.ShowOnlyUnresolvedReviews]: 'Show unresolved comments'
```
fregante, line 28: I previously suggested this, we don’t need enums, you can use these keys as string directly instead of 1, 2, 3, 4
```suggestion
	showOnlyUnresolvedReviews: 'Show unresolved comments'
```

If it works but typescript complains I’ll fix it

source/features/conversation-timeline-filter.tsx
```diff
@@ -0,0 +1,174 @@
...
+	select(test)!.style.display = '';
+}
+
+function reapplySettings(): void {
+	for (const element of select.all('.js-timeline-item')) {
+		processTimelineItem(element);
+	}
+}
+
+// @ts-expect-error TODO: adjust to new select menu
```
fregante, line 54: Can be dropped now
```suggestion
// @ts-expect-error TODO: adjust to new select menu
```

source/features/conversation-timeline-filter.tsx
```diff
@@ -0,0 +1,225 @@
...
+function reapplySettings(): void {
+	select
+		.all(timelineItemSelector)
+		.forEach(element => {
+			processTimelineItem(element);
+		});
+}
+
+function restoreSettings(): void {
+	select.all<HTMLInputElement>(`#${showFilterName}`).forEach(element => {
```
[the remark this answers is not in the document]
fregante: This stil applies, but I don’t know what’s selecting, there are no inputs on the page.

source/features/conversation-timeline-filter.tsx
```diff
@@ -0,0 +1,174 @@
...
+	select('#partial-users-participants')!.before(
+		<div className="discussion-sidebar-item js-discussion-sidebar-item">
+			<details className="details-reset details-overlay select-menu hx_rsm">
+				<summary className="text-bold discussion-sidebar-heading discussion-sidebar-toggle hx_rsm-trigger" aria-haspopup="menu" data-hotkey="x" role="button">
+					<GearIcon/>
+					<p>Filters</p>
+					<div id={timelineFiltersSelectorId}/>
+				</summary>
+
+				<details-menu className="select-menu-modal position-absolute right-0 hx_rsm-modal js-discussion-sidebar-menu" style={{zIndex: 99}}>
```
fregante, line 84: Long JSX lines like this should be split. You can run prettier on this file if you want help with the wrapping and then use `xo --fix` again. We don’t agree with everything Prettier does but it can help

source/features/conversation-timeline-filter.tsx
```diff
@@ -0,0 +1,174 @@
...
+			</div>
+		</label>
+	);
+}
+
+async function addTimelineItemsFilter(): Promise<void> {
+	select('#partial-users-participants')!.before(
+		<div className="discussion-sidebar-item js-discussion-sidebar-item">
+			<details className="details-reset details-overlay select-menu hx_rsm">
+				<summary className="text-bold discussion-sidebar-heading discussion-sidebar-toggle hx_rsm-trigger" aria-haspopup="menu" data-hotkey="x" role="button">
```
fregante, line 78: Let’s discuss hotkeys in a different issue or PR
```suggestion
				<summary className="text-bold discussion-sidebar-heading discussion-sidebar-toggle hx_rsm-trigger" aria-haspopup="menu" role="button">
```

source/features/conversation-timeline-filter.tsx
```diff
@@ -0,0 +1,174 @@
...
+	select.all<HTMLInputElement>(`#${showFilterName}`).forEach(element => {
+		element.checked = false;
+	});
+	select<HTMLInputElement>(`input[name=${showFilterName}][value="${currentSettings.valueOf()}"]`)!.checked = true;
+}
+
+function createRadio(title: string, summary: string, filterSettings: FilterSettings, checked: boolean): JSX.Element {
+	return (
+		<label className="select-menu-item d-flex" aria-checked={String(checked)} role="menuitemradio" tabIndex={0} onClick={async () => saveSettings(filterSettings, `#rgh-filter-menu-item-checkbox-${filterSettings}`)}>
+			<CheckIcon id={`rgh-filter-menu-item-checkbox-${filterSettings}`} style={{display: checked ? 'inherit' : 'none'}} className={`${menuItemCheckbox} select-menu-item-icon`} aria-hidden="true"/>
```
fregante, line 65:
```suggestion
			<CheckIcon id={`rgh-filter-menu-item-checkbox-${filterSettings}`} hidden={!checked} className={`${menuItemCheckbox} select-menu-item-icon`} aria-hidden="true"/>
```

source/features/conversation-timeline-filter.tsx
```diff
@@ -0,0 +1,174 @@
...
+					<div className="hx_rsm-content" role="menu">
+						{createRadio('Show all', '', FilterSettings.ShowAll, true)}
+						{createRadio('Show only comments', 'Hides commits and events', FilterSettings.ShowOnlyComments, false)}
+						{createRadio('Show only unresolved comments', 'Also hides resolved reviews and hidden comments', FilterSettings.ShowOnlyUnresolvedComments, false)}
+						{pageDetect.isPRConversation() && createRadio('Show only unresolved reviews', 'Also hides regular comments', FilterSettings.ShowOnlyUnresolvedReviews, false)}
+					</div>
+				</details-menu>
+			</details>
+
+			<p className="reason text-small text-gray"/>
```
fregante, line 100: I don’t think this used anymore, there’s a div.description above, right?

fregante (changes requested):

fregante:
There's something that's causing extra space here:

Screen Shot 1


It should look liekelike:

Screen Shot 2

fregante:
This isn't working:

Screen Shot

fregante:
Screen Shot

source/features/conversation-timeline-filter.tsx
fregante: I just did, it works fine

source/features/conversation-timeline-filter.tsx
fregante: In fact one of these functions is causing this 

Screen Shot

fregante:
This is currently hiding entire reviews by default: https://github.com/sindresorhus/refined-github/pull/3926, which made me review twice:

`master` | This PR
Screen Shot 6 Screen Shot

source/features/conversation-timeline-filter.tsx
shoter: Done

source/features/conversation-timeline-filter.tsx
shoter: Done

source/features/conversation-timeline-filter.tsx
shoter: Done

source/features/conversation-timeline-filter.tsx
shoter: Hmm. Yeah it is working now. Weird it was not working for me.

shoter:
> 
> 
> This is currently hiding entire reviews by default, which made me review #3926 twice:
> `master` 	This PR
> Screen Shot 6 	Screen Shot

Fixed in a055fb48f448be1324cbc35b0ac566a3b8b5f0b6 and d3a62f2d8ea12b8cf8f6e24f62b26372901aa349

shoter:
I also finally found a case how hidden comment looks like and it is in this PR.
image

Is `.rgh-preview-hidden-comments` class good enough to find hidden comments? I will implement hiding them for 3rd and 4th option in the filter.

source/features/conversation-timeline-filter.tsx
shoter: I could not figure it out how to write it in a way that you would like.
I had 2-3 attempts and I failed miserably every time.

shoter: test

shoter:
test

shoter:
test

shoter:
I also observed js error while selecting filters:
image

Probably some conflict with original github code though I did not observe any side-effects of that error. I do not know how I would deal with that.

yakov116:
@shoter can you update to node 15/npm 7 since you are creating a new package.lock

Thanks

fregante:
> Is `.rgh-preview-hidden-comments` class good enough to find hidden comments

No, we try not to make features inter-dependent. However if the `preview-hidden-comments` has good selectors, you can use something like `export const hiddenCommentSelector = '……'` in it and import it here.

> since you are creating a new package.lock

Where?

fregante:
> I also observed js error while selecting filters:

That error means that whatever you're calling `.closest()` on does not exist. Its selector may be wrong

yakov116:
> > since you are creating a new package.lock
> 
> Where?

My bad it was already reverted [here](https://github.com/sindresorhus/refined-github/pull/3847/commits/00aa77f6d11ed1156f3462b698070b7b677056ac)

fregante:
I shortened the labels a bit, all that text was hard to parse and understand.

Screen Shot

fregante:
I shortened the labels a bit, all that text was hard to parse and understand.

Screen Shot 3

source/features/pull-request-timeline-filter.tsx
fregante: Adding a comment to test this feature

fregante (commented):
I'm adding a new comment to test part of the feature

fregante:
@shoter as you can see above, the "special case" is actually well-marked by GitHub. In this case, I added a review with comment to a resolved conversation.

Please update the code accordingly (i.e. drop the "special case" and just detect it as a regular resolved/unresolved review comment)

fregante:
There are a few more improvements to be made:

- [ ] We should add a note at the bottom of the conversation to show that something is being hidden. You can refer to `hide-useless-comments` for the style/code
- [ ] It'd be good to also show how many events are being hidden in such comment, e.g. "3 events have been hidden. [Show all](#js)"
- [ ] The feature should actually not run at all until the user makes a choice, i.e. don't call `observe('.js-timeline-item'` until the user makes the first selection. I think the feature is _relatively_ heavy and should not run on every single page load even if it's almost never going to be used.

_I think_ after this we're good, but we'll see how that looks after, hoping there are not more bugs that I haven't seen. 😅

source/features/conversation-timeline-filter.tsx
```diff
@@ -0,0 +1,162 @@
...
+			// We are just checking here if user is able to comment inside that timeline thread, if not then it means we have this special situation that was just described.
+			applyDisplay(
+				threadContainer,
+				'showOnlyComments',
+				'showOnlyUnresolvedComments'
+			);
+		}
+
+		// We need to hide whole thread group if we have hidden all comments inside.
+		hasVisibleElement = hasVisibleElement || !threadContainer.hidden;
```
fregante, line 129: I think checking whether a whole review is resolved is easier than keeping track this way. 

Before this loop you can probably just check “select.exists(unresolved review threads, item)”, and if not, just hide the whole item and return; No need to loop at that point.

fregante:
I'll rewrite the missing parts in a new PR. No worries, the idea is good, but it required/s a lot of work.

shoter:
@fregante Thanks for that. Lately I could not find some time to have more look into it. I hope it makes it into repo :).

helmer:
Looks extremely useful, how do I enable this, 3 months in?

PR #3849, opened 2020-12-28
Meta: Update `element-ready`

fregante:
Includes https://github.com/sindresorhus/element-ready/issues/29

Findings about the new default:

- ✅ safer, fixes bugs we never knew we had
- ✅ avoids awkward/lengthy workarounds
- ⚠️ slows down operations that don't depend on the content
- ❌ seriously slowed down only one selector: `body`

You can see the 3 commits below highlighting the 3 levels above (✅⚠️❌)

I added `{waitForChildren: false}` to most `elementReady` calls, but actually this only **saves a handful of frames/milliseconds**. The time saved depends on the size of the target element’s HTML string and the connection speed. 

Hoping that we agree that safety is a good default, I have a couple of questions:

- Refined GitHub: Are these savings worth the extra configuration?
- element-ready: Does the amount of usage justify a shortcut?
  ```diff
  - elementReady(selector, {waitForChildren: false})
  + elementReady.fast(selector)


sindresorhus:
> Hoping that we agree that safety is a good default

👍 

> Refined GitHub: Are these savings worth the extra configuration?

👍 

> element-ready: Does the amount of usage justify a shortcut?

👎

PR #3850, opened 2020-12-28
Update dependencies

fregante:
Except #3849


tsconfig.json
```diff
@@ -3,7 +3,8 @@
 	"compilerOptions": {
 		"target": "es2020",
 		"declaration": false,
-		"esModuleInterop": true
+		"esModuleInterop": true,
+		"noUncheckedIndexedAccess": false
```
fregante, line 7: `true` brings a slew of mindless `!` across the code; 94 errors to be exact. In the vast majority of cases it doesn't really highlight anything except the inexactness of some types (`''.split(' ')[0]` is never `undefined`). It also breaks a few patterns like parameter destructuring because inevitably you'll have the `no X property in `{X} | undefined`

Maybe it can be enabled separately by someone willing to refactor those instances by actually making them safer (if ever possible)

source/features/release-download-count.tsx
```diff
@@ -3,6 +3,7 @@ import React from 'dom-chef';
 import select from 'select-dom';
 import {DownloadIcon} from '@primer/octicons-react';
 import * as pageDetect from 'github-url-detection';
+import {abbreviateNumber} from 'js-abbreviation-number';
```
fregante, line 6: I found this library not a whole lot bigger than our own code

PR #3693, opened 2020-10-30
Add `feature-descriptions` feature

yakov116:
1. LINKED ISSUES:
   Closes #3630 
2. TEST URLS:
   https://github.com/sindresorhus/refined-github/blob/master/source/features/batch-mark-files-as-viewed.tsx

3. SCREENSHOT:
image

PR #3852, opened 2020-12-28
Add `rgh-feature-descriptions` feature

### LINKED ISSUE #3630: Meta-feature: Add information box to each `feature` file
On this same repo, when visiting a feature file (https://github.com/sindresorhus/refined-github/blob/master/source/features/action-used-by-link.tsx) we could display a box with parsed information about the feature, both as documentation and as a debug helper.

Content the box could include:

- the screenshot as a ``
- the formatted description
- links to PRs that implemented it, like the [history page](https://github.com/sindresorhus/refined-github/commits/master/source/features/action-used-by-link.tsx) without clutter (skips any meta PRs) — discussed [here first](https://github.com/sindresorhus/refined-github/pull/3621#issuecomment-704596791)
- an issue ego-search, like https://github.com/sindresorhus/refined-github/issues?q=%22show-whitespace%22, displaying the text: _34 conversations, 3 open_

FloEdelmann:
I picked up @yakov116's work from #3693, and added a few bits. I hope this is enough to close #3630.

Test URLs:

https://github.com/sindresorhus/refined-github/blob/master/source/features/batch-mark-files-as-viewed.tsx

<details>
<summary>Screenshot</summary>

image

</details>

https://github.com/sindresorhus/refined-github/blob/20.12.27/source/features/align-issue-labels.css

<details>
<summary>Screenshot</summary>

grafik

</details>

https://github.com/sindresorhus/refined-github/blob/20.12.27/source/features/clean-dashboard.ts

<details>
<summary>Screenshot</summary>

image

</details>


source/features/feature-descriptions.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+function getConversationsLink(featureName: string): Element {
+	const searchParams = new URLSearchParams({
+		q: `"${featureName}" sort:updated-desc`
+	});
+
+	const searchUrl = new URL('https://github.com/sindresorhus/refined-github/issues');
+	searchUrl.search = String(searchParams);
+
+	return <a className="ml-3" href={String(searchUrl)}>Conversations</a>;
```
FloEdelmann, line 102: Unfortunately, there seems to be no simple way to filter repository issues via the API.

source/features/feature-descriptions.tsx
```diff
@@ -0,0 +1,149 @@
...
+	return repository.defaultBranchRef.target.history.nodes;
+};
+
+function getCommitUrl(commit: Commit): string {
+	const [, pullRequestUrl] = /<a[^>]+href="([^"]+)">/.exec(commit.messageHeadlineHTML) ?? [];
+	return pullRequestUrl ?? commit.commitUrl;
+}
+
+async function getHistoryDropdown(featureName: string): Promise<Element> {
+	let history = await getFeatureHistory(featureName + '.tsx');
```
yakov116, line 46: I think we should add a detection to featureMeta if it's a CSS feature. Avoiding 2 calls.

yakov116: Also it needs a cache.function...

FloEdelmann: Problem is: when visiting a `.css` file, there is no way to tell (without an API call) whether it belongs to a regular `.tsx` feature, or a CSS-only `.ts` feature.

yakov116: > I think we should add a detection to featureMeta

Add it here

FloEdelmann: > add a detection to featureMeta

How? `__featuresMeta__` is provided by `webpack.config.ts`, which only loads `source/refined-github.ts` and `readme.md`.

yakov116: Can we detect if the file has `features.addCssFeature`?

fregante: That only works on CSS features that can be disabled. If you’re visiting a real CSS-only feature it won’t appear in __featuresMeta__ at all. 

Does `tsx` have to be specified in the API call?

fregante: That only works on CSS features that can be disabled. If you’re visiting a real CSS-only feature it won’t appear in __featuresMeta__ at all. 

Does `tsx` have to be specified in the API call?

yakov116: Duhh

source/features/feature-descriptions.tsx
```diff
@@ -0,0 +1,149 @@
...
+	descriptionElement.classList.add('text-bold');
+
+	const commitInfoBox = (await elementReady('.Box-header--blue.Details'))!.parentElement!;
+	commitInfoBox.classList.add('width-fit', 'flex-auto');
+	commitInfoBox.classList.remove('mb-3');
+
+	const featureInfoBox = (
+		<div className="Box" style={{flex: '1 0 360px'}}>
+			<div className="Box-row d-flex">
+				{screenshot && id !== __filebasename &&
```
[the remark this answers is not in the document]
FloEdelmann: I've not modified this logic. @yakov116 could you please explain your reasoning behind this?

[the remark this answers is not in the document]
yakov116: When I made the feature it had a picture so I thought it did not make sense to show a picture of itself

source/features/feature-descriptions.tsx
FloEdelmann: > Does `.tsx` have to be specified in the API call? Can it specify both tsx and css?

It expects exactly one full file path including file extension, or none at all.

> Maybe this information box just can’t appear on CSS features.

I fear you might be right :frowning_face:

source/features/feature-descriptions.tsx
```diff
@@ -0,0 +1,149 @@
...
+					<a href={screenshot}>
+						<img
+							src={screenshot}
+							className="d-block border"
+							height="100"
+							width="100"
+							style={{objectFit: 'cover'}}/>
```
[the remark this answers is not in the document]
FloEdelmann: I've taken a different approach in [`ab553eb`](https://github.com/sindresorhus/refined-github/pull/3852/commits/ab553eb548489ccb0c024da11112d012d6a96850): Restrict the max-height and max-width, but keep the aspect ratio of the image:

grafik

grafik

grafik

source/features/feature-descriptions.tsx
yakov116: I feel that this feature is replacing the description and picture that we removed. The CSS ones never had it until now.

source/features/feature-descriptions.tsx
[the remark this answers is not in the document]
fregante: Looks great! I'd allow even more horizontal space, it'd be useful for landscape screenshots

source/features/feature-descriptions.tsx
```diff
@@ -0,0 +1,152 @@
...
+					{descriptionElement}
+					{await getHistoryDropdown(id!)}
+					{getConversationsLink(id!)}
+				</div>
+			</div>
+		</div>
+	);
+
+	wrapAll([commitInfoBox, featureInfoBox], <div className="d-flex flex-wrap" style={{gap: 16}}/>);
+}
```
fregante, line 144: Instead of having the whole thing `await` the API call, can you just append the dropdown later?


Example here:
```suggestion
	select('.rgh-feature-box').append(await getHistoryDropdown(id))
}
```

source/features/feature-descriptions.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+	return <a className="ml-3" href={String(searchUrl)}>Conversations</a>;
+}
+
+async function init(): Promise<void | false> {
+	const [, currentFeature] = /features\/([^.]+)/.exec(location.pathname)!;
+	const {id, description, screenshot} = __featuresMeta__.find(feature => feature.id === currentFeature) ?? {};
+	if (!description) {
+		return false;
+	}
```
fregante, line 111:
```suggestion
	const feature = __featuresMeta__.find(feature => feature.id === currentFeature);
	if (!feature) {
		return false;
	}

	const {id, description, screenshot} = feature;
```

This way `id` will never be `string | undefined`

readme.md
```diff
@@ -110,6 +110,7 @@ Thanks for contributing! 🦋🙌
 - [](# "file-finder-buffer") [Lets you start typing your search immediately after invoking the File Finder (<kbd>t</kbd>), instead of having you wait for it to load first.](https://user-images.githubusercontent.com/1402241/75542106-1c811700-5a5a-11ea-8aa5-bea0472c59e2.gif)
 - [](# "preserve-file-finder-term") Preserves the search terms when navigating back and forth between the File Finder and the files.
 - [](# "show-associated-branch-prs-on-fork") [Shows the associated pull requests on branches for forked repositories.](https://user-images.githubusercontent.com/16872793/81504659-7e5ec800-92b8-11ea-9ee6-924110e8cca1.png)
+- [](# "feature-descriptions") [Shows the description and screenshot of each Refined GitHub feature when visiting the features’ file.](https://user-images.githubusercontent.com/16872793/97794235-b8f14f80-1bcd-11eb-9e66-de7c6880a162.gif)
```
fregante, line 113: Since this feature is enabled exclusively on very specific files on our repo, it neither needs to be documented nor should it be disableable. Can you add an exception in `webpack.config.ts` so there are no errors and it doesn't appear in `options.html`?

source/features/feature-descriptions.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+	return (
+		<details className="dropdown details-reset details-overlay d-inline-block">
+			<summary className="text-gray d-inline" aria-haspopup="true">
+				Feature history
+				<div className="dropdown-caret ml-1"/>
+			</summary>
+
+			<div className="dropdown-menu dropdown-menu-s" style={{width: 400}}>
+				<ul className="overflow-y-auto" style={{maxHeight: '60vh'}}>
```
fregante, line 72: Is the height necessary? The whole page can scroll instead

source/features/feature-descriptions.tsx
```diff
@@ -0,0 +1,152 @@
...
+								</div>
+							</a>
+						</li>
+					))}
+					{skippedCommitsCount > 0 && (
+						<>
+							<li className="dropdown-divider" role="separator"/>
+							<li>
+								<a className="dropdown-item" href={String(historyUrl)}>
+									View full history (+{skippedCommitsCount} commits)
```
fregante, line 88: We never need this link, there's already a History link on the page

source/features/feature-descriptions.tsx
```diff
@@ -0,0 +1,152 @@
...
+			</div>
+		</details>
+	);
+}
+
+function getConversationsLink(featureName: string): Element {
+	const searchUrl = new URL('https://github.com/sindresorhus/refined-github/issues');
+	searchUrl.searchParams.set('q', `"${featureName}" sort:updated-desc`);
+
+	return <a className="ml-3" href={String(searchUrl)}>Conversations</a>;
```
fregante, line 103: Feels like an overkill just to avoid manually escaping 2 characters. I'm sure Sindre disagrees, but 😜
```suggestion
	const url = '/sindresorhus/refined-github/issues?q=${featureName}+sort%3Aupdated-desc';
	return <a className="ml-3" href={url}>Conversations</a>;
```

source/features/feature-descriptions.tsx
FloEdelmann: So should I also drop the `skippedCommitsCount`?

readme.md
FloEdelmann: That also means that the feature won't be enabled on its own `.tsx` file :sweat_smile:

source/features/feature-descriptions.tsx
FloEdelmann: It doesn't need to, but it looks weird on very small features with a long history, where the dropdown menu is longer than the page itself.

readme.md
yakov116: Can't you just exclude it In the index.tsx and that would exclude it from the options?

fregante: > That also means that the feature won't be enabled on its own `.tsx` file 😅

Sounds like something that can be fixed with… code. 👌

fregante:
> [@fregante](/fregante) [fregante](/fregante) deleted a comment from [FloEdelmann](/FloEdelmann) [7 minutes ago](#event-4160501132)

oops my bad

source/features/rgh-feature-descriptions.tsx
```diff
@@ -0,0 +1,140 @@
...
+							}
+						}
+					}
+				}
+			}
+		}
+	`);
+
+	return repository.defaultBranchRef.target.history.nodes.filter(
+		(commit: Commit) => !/^Meta|^Document|^Readme|^Lint|^Update.+dependencies/.test(commit.messageHeadline)
```
fregante, line 41: Let's use https://github.com/notlmn/release-with-changelog/blob/5c804153ef8227047cf9bf373a704791aa3ee755/generate-release-notes.js#L6

readme.md
fregante: > Since this feature is enabled exclusively on very specific files on our repo, it neither needs to be documented nor should it be disableable.

fregante (changes requested):

readme.md
```diff
@@ -110,6 +110,7 @@ Thanks for contributing! 🦋🙌
 - [](# "file-finder-buffer") [Lets you start typing your search immediately after invoking the File Finder (<kbd>t</kbd>), instead of having you wait for it to load first.](https://user-images.githubusercontent.com/1402241/75542106-1c811700-5a5a-11ea-8aa5-bea0472c59e2.gif)
 - [](# "preserve-file-finder-term") Preserves the search terms when navigating back and forth between the File Finder and the files.
 - [](# "show-associated-branch-prs-on-fork") [Shows the associated pull requests on branches for forked repositories.](https://user-images.githubusercontent.com/16872793/81504659-7e5ec800-92b8-11ea-9ee6-924110e8cca1.png)
+- [](# "rgh-feature-descriptions") [Shows the description and screenshot of each Refined GitHub feature when visiting the features’ file.](https://user-images.githubusercontent.com/202916/103462909-e4f19400-4d28-11eb-86d3-5b97d75b4781.png)
```
fregante, line 113: > Since this feature is enabled exclusively on very specific files on our repo, it neither needs to be documented nor should it be disableable. 
```suggestion
```

source/options.tsx
```diff
@@ -82,7 +82,8 @@ async function highlightNewFeatures(): Promise<void> {
 
 async function generateDom(): Promise<void> {
 	// Generate list
-	select('.js-features')!.append(...__featuresMeta__.map(buildFeatureCheckbox));
+	const features = __featuresMeta__.filter(feature => !feature.id.startsWith('rgh-'));
+	select('.js-features')!.append(...features.map(buildFeatureCheckbox));
```
fregante, line 86:
```suggestion
	const checkboxes = __featuresMeta__
		.filter(feature => !feature.id.startsWith('rgh-'))
		.map(buildFeatureCheckbox);
	select('.js-features')!.append(...checkboxes);
```

readme.md
fregante: The feature needs to be excluded in `webpack.config.ts` so it doesn't error

webpack.config.ts
```diff
@@ -42,6 +42,13 @@ function parseFeatureDetails(readmeContent: string, id: FeatureID): FeatureMeta
 		};
 	}
 
+	if (id.startsWith('rgh-')) {
+		return {
+			id,
+			description: '<p>Refined GitHub Meta feature</p>'
+		};
+	}
+
```
fregante, line 51: This is happening too late and it leaves the feature in every list, with junk content.

Fixed in

fregante (approved):

PR #3855, opened 2020-12-29
Drop `select-dom` in favor of extending the prototype

fregante:
It's not April 1st yet, but I wondered what the code would look like if:

- we used `$()`/`$$()` (like the console API) instead of `select()`/`select.all()`
- we used `parent.$(str)` instead of `select(str, parent)`

### Advantages

- More linear DOM traversing:

	```diff
	- select('.child', parent.closest('body'))
	+ parent.closest('body').$('.child')
	```
- `select.all` _always_ converts NodeList to Array, but that's rarely needed thanks to for-of, so it can be skipped
- More code can be copied from/to the console when testing selectors
- No more imports

### Drawbacks

- Magic globals (however TypeScript still recognizes them and shows its types)
- `tab.$('.btn')` might appear strange at first

### Alternative 

Would it be better if the method was `.query()` instead of `.$()`? query/queryAll were supposed to be DOM4 extensions but were lost along the way. The reason why we used `select-dom` was because `"document.querySelectorAll()"` was too long, but query/queryAll is actually shorter than select/select.all. The globals could still be $/$$


yakov116:
> It's not April 1st yet, but I wondered what the code would look like if:

What's April 1st? 

> tab.$('.btn') might appear strange at first

It does but if that is the only trade off 👍

FloEdelmann:
Another drawback: `$` looks like jQuery, so I expect `$(selector)` to return a jQuery object instead of a DOM element. I'd be fine with `select`/`selectAll` though.

fregante:
> Another drawback: $ looks like jQuery

True, but it also looks like the console API. If it’s annoying to others we can just use query/all everywhere

fregante:
Closing this as it was a proof of concept only. The intent still exists

yakov116:
Reopening to test the `update-pr-from-base-branch` issue

PR #3873, opened 2021-01-09
Fix `separate-draft-pr-button` position 

### LINKED ISSUE #3871: `separate-draft-pr-button` is in the wrong position
1. Try opening a PR, e.g. https://github.com/sindresorhus/refined-github/compare/unknow-feature-name?expand=1
2. See the buttons above the field rather than below

Screen Shot

cheap-glitch:
Thanks for contributing! 🍄

1. LINKED ISSUES: Fix #3871 

2. TEST URLS:
 * https://github.com/sindresorhus/refined-github/compare/unknow-feature-name?expand=1

3. SCREENSHOT:
<table>
<tr>
	<th>Before
	<th>After
<tr>
	<td>

Screenshot_2021-01-09 sindresorhus refined-github
	</td>
	<td>

Screenshot_2021-01-09 sindresorhus refined-github(1)
	</td>
</table>

Bug was caused by a new wrapper around the PR form.

### DIFF
--- a/source/features/separate-draft-pr-button.tsx
+++ b/source/features/separate-draft-pr-button.tsx
@@ -13,7 +13,7 @@ function init(): void | false {
 		return false;
 	}
 
-	const buttonBar = select('.timeline-comment > :last-child', previewForm)!;
+	const buttonBar = select('.js-slash-command-surface > :last-child', previewForm)!;
 	const createPrButtonGroup = select('.BtnGroup', buttonBar);
 	if (!createPrButtonGroup) {
 		// Free accounts can't open Draft PRs in private repos, so this element is missing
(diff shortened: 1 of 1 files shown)


source/features/separate-draft-pr-button.tsx
```diff
@@ -13,7 +13,7 @@ function init(): void | false {
 		return false;
 	}
 
-	const buttonBar = select('.timeline-comment > :last-child', previewForm)!;
+	const buttonBar = select('.js-slash-command-surface > :last-child', previewForm)!;
 	const createPrButtonGroup = select('.BtnGroup', buttonBar);
```
fregante, line 17: There's a better class now:
```suggestion
	const createPrButtonGroup = select('.hx_create-pr-button');
```

And then use `createPrButtonGroup.before()` instead of `buttonBar.prepend()`

PR #3839, opened 2020-12-23
Expand but hide "1 similar comment" in `hide-useless-comments` #3830

PauloCarneiro99:
1. LINKED ISSUES:
   Closes #3830 

2. TEST URLS:
   https://github.com/justinfrankel/licecap/issues/97#issuecomment-737063337

PR #3879, opened 2021-01-10
Don't show `update-pr-from-base-branch` when there are conflicts

### LINKED ISSUE #3861: `update-pr-from-base-branch` shouldn't appear when there's a "Resolve conflicts" button
"Resolve conflicts" is the native button to update the branch, so `update-pr-from-base-branch` should not run in that case.

Example URL: https://github.com/sindresorhus/refined-github/pull/3855

yakov116:
1. LINKED ISSUES:
   Fixes #3861 

2. TEST URLS:
   #3855 
3. SCREENSHOT:
   None

### DIFF
--- a/source/features/update-pr-from-base-branch.tsx
+++ b/source/features/update-pr-from-base-branch.tsx
@@ -94,7 +94,7 @@ async function init(): Promise<void | false> {
 
 	// Button exists when the current user can merge the PR.
 	// Button is disabled when there are conflicts (there's already a native "Resolve conflicts" button)
-	const canMerge = select.exists('[data-details-container=".js-merge-pr"]:not(:disabled)');
+	const canMerge = select.exists('.merge-pr > .text-gray:first-child');
 	const hasConflicts = select.exists('.js-merge-pr a[href$="/conflicts"]');
 
 	if (hasConflicts || !canMerge) {
--- a/source/features/update-pr-from-base-branch.tsx
+++ b/source/features/update-pr-from-base-branch.tsx
@@ -41,12 +41,6 @@ async function handler({delegateTarget}: delegate.Event): Promise<void> {
 	const response = await mergeBranches();
 	if (response.ok) {
 		statusMeta.remove();
-	} else if (response.message?.toLowerCase().startsWith('merge conflict')) {
-		// Only shown on Draft PRs
-		statusMeta.textContent = '';
-		statusMeta.append(
-			<a href={location.pathname + '/conflicts'} className="btn float-right"><AlertIcon/> Resolve conflicts</a>
-		);
 	} else {
 		statusMeta.textContent = response.message ?? 'Error';
 		statusMeta.prepend(<AlertIcon/>, ' ');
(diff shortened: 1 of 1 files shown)


source/features/update-pr-from-base-branch.tsx
```diff
@@ -104,7 +98,9 @@ async function init(): Promise<void | false> {
 	// - Draft PR (show the button anyway)
```
fregante, line 98: This shouldn't exist anymore. Maybe the condition is `hasConflicts || !canMerge`

source/features/update-pr-from-base-branch.tsx
```diff
@@ -104,7 +98,9 @@ async function init(): Promise<void | false> {
...
 	const canMerge = select.exists('[data-details-container=".js-merge-pr"]:not(:disabled)');
 	const isDraftPR = select.exists('[action$="ready_for_review"]');
-	if (!canMerge && !isDraftPR) {
+	const hasConflicts = select.exists('.js-merge-pr a[href$="/conflicts"]');
+
+	if ((!canMerge && !isDraftPR) || hasConflicts) {
```
fregante, line 103: The logic overlaps, maybe the whole block should be be rewritten because the core logic is "if has conflicts or can not merge, return"

We already have `hasConflicts` but can we determine whether "it can merge" in a single selector? Without also checking specifically for draft PRs.

There are 8 situations to be tested, combinations of:

- PR state: regular or draft
- user: collaborator or third party
- conflicts: yes or no

Note: these are situations to be tested, not conditions that should appear in the code

yakov116: Why are we checking if its a draft PR?

Don't we only need to check if the user can merge?

fregante: That's what I'm saying. We don't.

> the core logic is "if has conflicts or can not merge, return"

> Without also checking specifically for draft PRs.

fregante:
I think our `canMerge` logic is incorrect. Even if you cannot merge, if you can push to the branch you can certainly update it.

We should replace it with `canPush`. I think that this text should be a good indicator, it appears on:

- repo owner: me, PR opener: me
- repo owner: other, PR opener: me
- repo owner: other, PR opener: other, but I'm a collaborator 

Screen Shot 3


Maybe the selector is this? `.merge-pr > .text-gray:first-child`

fregante:
I don't see "Update branch" here: https://github.com/sindresorhus/refined-github/pull/3839

I think the logic suggested above would fix it

source/features/update-pr-from-base-branch.tsx
fregante: This doesn't work on draft PRs at all now because the button is always `.disabled` there

fregante:
> We should replace it with `canPush`

source/features/update-pr-from-base-branch.tsx
```diff
@@ -94,10 +94,10 @@ async function init(): Promise<void | false> {
 
 	// Button exists when the current user can merge the PR.
 	// Button is disabled when there are conflicts (there's already a native "Resolve conflicts" button)
```
fregante: ?

fregante (changes requested):
Please send a complete PR

yakov116:
Not sure if this PR did it.

But https://github.com/sindresorhus/refined-github/pull/3904 has update base branch 3 times

PR #3886, opened 2021-01-11
Add `wait-for-attachments` feature

### LINKED ISSUE #3881: Wait for attachments to be uploaded before allowing comment submission
## Problem

1. Write a comment
2. Attach file
3. Click the "Comment" button before it's done uploading

This is the result: (image will be missing)

## Solution

When the user tries to submit a comment, the form should be disabled. Interestingly enough, the "New issue" form has this feature already, so we might be able to reuse it somehow.

Notice the green button fading:

gif

cheap-glitch:
Thanks for contributing! 🍄

1. LINKED ISSUES: Closes #3881 

2. TEST URLS:
- writing a new comment
- editing a comment
- [opening a new PR](https://github.com/sindresorhus/refined-github/compare/unknow-feature-name?expand=1)

3. SCREENSHOT:

demo


fregante:
I was hoping this was a literal single class to be added since GitHub already has code taking care of the whole operation. I noticed that on upload they call `textarea.setCustomValidity("uploading")` and the invalidity of the form disables the Comment buttons on the `isNewIssue` page

fregante:
Alternatively I see `upload:setup` and `upload:complete` being fired, which could make it simple enough

cheap-glitch:
I rewrote the feature using the suggestions of https://github.com/sindresorhus/refined-github/issues/3881#issuecomment-758372304 and it works great!

The "Close with comment" buttons are unaffected though, do we extend the feature to them too?

n012try (approved):

source/features/wait-for-attachments.ts
```diff
@@ -0,0 +1,43 @@
...
+	if (disabled) {
+		textarea.removeAttribute('data-required-trimmed');
+	} else {
+		textarea.dataset.requiredTrimmed = 'Text field is empty';
+	}
+
+	// Force the "Update comment"/"Create pull request" buttons to be disabled
+	for (const button of select.all('button[type="submit"]', target.closest('form')!)) {
+		button.toggleAttribute('disabled', disabled);
+	}
```
fregante, line 17: One of these is unnecessary. Either you toggle the `data-required-trimmed` attribute _or_ manually disable the buttons. The whole point of removing `data-required-trimmed` would be so that we don't have to manually disable/enable the buttons since that's handled by the upload events

cheap-glitch: That's what I thought too, but for some reason the "Comment" button doesn't get disabled if the `data-required-trimmed` attribute is not removed on the textarea. My guess is that the `disabled` attribute on the button is silently overwritten by GitHub at some point before or during the upload.

So the attribute removal is for the "Comment" button on new comment forms, and the `for` loop for all the others (including the ones added by RGH).

fregante: > the `for` loop for all the others

I'd rather address the source of this problem directly, which incidentally is the same reason why "Close with comment" is also unaffected:

`data-disable-invalid` isn't set on either button.

I made some changes but I did not address this issue specifically yet.

fregante: I think `data-disable-invalid` can't be set on "Close with comment" because the same button has to work as a plain "Close pull request" when the field is empty.

For this button we do indeed need the `for` loop, so we might as well restore it for both comment editing and "close with comment" buttons.

fregante: > need the `for` loop, so we might as well restore it for both comment editing and "close with comment" buttons.

This is still missing

cheap-glitch: > This is still missing

Done. I also had to change the selector for the textarea because the previous one didn't work when editing a comment or on the new PR page.

source/features/wait-for-attachments.ts
```diff
@@ -17,6 +17,11 @@ function toggleSubmitButtons({target, type}: Event): void {
 		textarea.setAttribute(attribute, textarea.getAttribute(attributeBackup)!);
 		textarea.removeAttribute(attributeBackup);
 	}
+
+	// Needed for "Update comment" and "Close with comment" buttons
+	for (const button of select.all('button[type="submit"]', form.closest('form')!)) {
```
fregante, line 22: Is `form` not a `<form>` already?

source/features/wait-for-attachments.ts
```diff
@@ -0,0 +1,42 @@
...
+
+import features from '.';
+
+// `data-required-trimmed` overrides the `disabled` state of the form
+const attribute = 'data-required-trimmed';
+const attributeBackup = 'data-rgh-required-trimmed';
+
+function toggleSubmitButtons({target, type}: Event): void {
+	const form = target as HTMLFormElement;
+	const textarea = select('textarea', form)!;
```
fregante, line 12: The selector can’t be simplified 019556fa9b269d0895ef071aa57cd3656f4e532f

Maybe it’s worth adding a comment to explain why the selector is complex.

cheap-glitch: I restored the selector, and added a check to make the feature work on PR pages and when editing a comment.

fregante: I added a comment to explain what exactly is happening here.

I haven't re-verified our code based on this explanation, but I think this feature should:

- temporarily remove `data-disable-invalid`
- permanently add `data-disable-invalid` to the buttons that don't have it

Does that sound correct? Is it true that we can (oops, once again) drop the loop and instead just permanently add `data-disable-invalid` to those buttons.

The only issue I see is finding the right time to add `data-disable-invalid`, maybe that should only happen in `upload:setup`

cheap-glitch: > Does that sound correct? Is it true that we can (oops, once again) drop the loop and instead just permanently add data-disable-invalid to those buttons.

Indeed, we don't need to disable the buttons, adding the attribute works. Tested in the three cases (new comment, edit and new PR).
The loop could probably be optimized though, we don't need it to run every time.

source/features/wait-for-attachments.ts
```diff
@@ -0,0 +1,50 @@
...
+
+import features from '.';
+
+const attribute = 'data-required-trimmed';
+const attributeBackup = 'data-rgh-required-trimmed';
+
+function toggleSubmitButtons({target, type}: Event): void {
+	const fileAttachment = target as HTMLElement;
+
+	// Ensures every submit button has the `data-disable-invalid` attribute
```
fregante, line 16: Doesn't this affect the "Close and comment" button? That should never be disabled. I think the selector needs to be more specific to the buttons we explicitly support, or else we risk breaking submission in these edge cases

cheap-glitch: Yes, the "Close with comment" is affected since it's a submit button.

> the buttons we explicitly support

What are those exactly? Just the green ones?

fregante: Apparently: 

https://github.com/sindresorhus/refined-github/pull/3886/commits/4837311169006bdf54131e2c4a5f6c6887add4fc#diff-de5c204d3ff1280d1b01a540e2b88a77fc7441a87265233b04915a1e9c2eccb0L29

Therefore it's just the "Update comment", that we know of, at this time.

cheap-glitch: I changed the selector to target only primary submit buttons. What about the "Create draft PR" button? Should we leave it unaffected too?

fregante: It could be ok for now

fregante:
Can you fix the conflicts?

PR #3890, opened 2021-01-12
Avoid embedding large files in `embed-gist-inline`

### LINKED ISSUE #3889: Page freezes after refinded-github Chrome ext runs
Page that seems to freeze in Chrome when using this ext is: https://github.com/hashicorp/vagrant/issues/11987
Disabling the Refined Github plugin resolves the issue.

I also noticed in the Chrome task manager, the tab jumps to 100% CPU and uses at ~455 MB of memory.

fregante:
Fixes #3889 

Test on https://github.com/hashicorp/vagrant/issues/11987 (warning, may cause browser to hang before this PR is merged)

## Screenshot
Screen Shot 4

### DIFF
--- a/source/background.ts
+++ b/source/background.ts
@@ -23,8 +23,13 @@ const messageHandlers = {
 		const response = await fetch(url);
 		return response.text();
 	},
-	async fetchJSON(url: string) {
+	async fetchJSON(url: string, _: any, message: any) {
 		const response = await fetch(url);
+		if (response.headers.get('Content-Length')! > message.limit) {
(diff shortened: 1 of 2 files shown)


yakov116 (approved):
Tested!

fregante:
I just noticed that `node-fetch` has a convenient [`size`](https://github.com/node-fetch/node-fetch#options) limit but [the spec doesn't](https://github.com/node-fetch/node-fetch/issues/879#issuecomment-643450359)

PR #3893, opened 2021-01-13
Don't apply `wait-for-build` when auto-merge is enabled

### LINKED ISSUE #3887: Disable wait-for-build when Github auto-merge is enabled
Github has a new (beta) feature that solves the same problem as wait-for-build but doesn't require the tab to stay open, it also respects other merge rules. See here: https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/automatically-merging-a-pull-request

The Refined Github wait-for-build UI should be hidden when this feature is enabled. The current UI (below) is a little confusing.

image

jack1142:
1. LINKED ISSUES:
   Fixes #3887

2. TEST URLS:
   Not possible to provide, the tester will need to make a repo with proper conditions met themselves.
   See SCREENSHOT section below for possible conditions.

3. SCREENSHOT:
   - [x] Auto-merge (GH only shows it when there are pending required checks):
   image
   - [x] Regular merge, auto-merge enabled, pending optional checks:
   image
   - [x] Regular merge, auto-merge enabled, all optional and required checks have passed already:
   image
   - [x] Admin merge (GH only shows it when there are pending required checks), auto-merge enabled:
   image
   image
   - [x] Regular merge, auto-merge disabled, pending optional checks:
   image
   image
   - [x] Admin merge (GH only shows it when there are pending required checks), auto-merge disabled:
   image
   image

### DIFF
--- a/source/features/wait-for-build.tsx
+++ b/source/features/wait-for-build.tsx
@@ -32,9 +32,14 @@ function showCheckboxIfNecessary(): void {
 	const checkbox = getCheckbox();
 	const isNecessary = prCiStatus.get() === prCiStatus.PENDING;
 	if (!checkbox && isNecessary) {
-		const container = select('.commit-form-actions .select-menu');
-		if (container) {
-			container.append(generateCheckbox());
+		const containers = select.all('.commit-form-actions .select-menu');
+		for (const container of containers) {
+			const container_form = container.closest('form');
+			// Only show checkbox for merge form, not for auto merge form
+			if (container_form?.classList.contains('js-merge-form')) {
+				container.append(generateCheckbox());
+				break;
+			}
 		}
 	} else if (checkbox && !isNecessary) {
 		checkbox.parentElement!.remove();
--- a/source/features/wait-for-build.tsx
+++ b/source/features/wait-for-build.tsx
@@ -32,14 +32,7 @@ function showCheckboxIfNecessary(): void {
 	const checkbox = getCheckbox();
 	const isNecessary = prCiStatus.get() === prCiStatus.PENDING;
 	if (!checkbox && isNecessary) {
-		const containers = select.all('.commit-form-actions .select-menu');
-		for (const container of containers) {
-			const container_form = container.closest('form');
-			// Only show checkbox for merge form, not for auto merge form
-			if (container_form?.classList.contains('js-merge-form')) {
-				container.append(generateCheckbox());
-				break;
-			}
+		select('.js-merge-form .select-menu')?.append(generateCheckbox());
 		}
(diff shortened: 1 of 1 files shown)


jack1142:
Should be ready for review now.

source/features/wait-for-build.tsx
```diff
@@ -32,9 +32,14 @@ function showCheckboxIfNecessary(): void {
...
-		if (container) {
-			container.append(generateCheckbox());
+		const containers = select.all('.commit-form-actions .select-menu');
+		for (const container of containers) {
+			const container_form = container.closest('form');
+			// Only show checkbox for merge form, not for auto merge form
+			if (container_form?.classList.contains('js-merge-form')) {
+				container.append(generateCheckbox());
+				break;
+			}
```
fregante, line 42: It looks to me this code is just looking for `.js-merge-form`. Is this equivalent?
```suggestion
		for (const container of select.all('.js-merge-form')) {
			select('.select-menu', container)!.append(generateCheckbox());
```

fregante:
Thanks for the PR and the many screenshots!

source/features/wait-for-build.tsx
jack1142: The only difference would be that this could possibly find multiple forms with `.js-merge-form` class and add the checkbox to all of them but there's only one present on GH's site so I guess that's not really a concern. But we could alternatively do this instead:
```suggestion
		const container = select('.js-merge-form');
		if (container)
			select('.select-menu', container)!.append(generateCheckbox());
```

fregante: If it works in every case, go for it!

Also that's the same as 
```suggestion
		select('.js-merge-form .select-menu')?.append(generateCheckbox());
```

😄

fregante (approved):
The behavior is unchanged for non-auto-merge users, but I assume that you tested the latest version and that it works for you for auto-merge PRs

jack1142:
I have tested the first version, but I have not tested the version with changes after your review comment, but I don't have much reason to believe it changes anything as it really just simplified the code that was there before without changing what it actually does.

fregante:
> I have not tested the version with changes after your review comment

It might be worth checking it out since that's the only change this PR is expected to make 😄 or else we're back to square 1 tomorrow.

jack1142:
Alright, just checked all the conditions from the PR description and they still work. Not gonna post screenshots again since it all looks the same as before.

fregante:
Perfect, thanks!

tom-sherman:
Amazing work, thanks so much!

PR #3904, opened 2021-01-19
Update to next dom-chef

fregante:
Follows https://github.com/vadimdemedes/dom-chef/pull/62 and https://github.com/vadimdemedes/dom-chef/pull/75


source/features/action-used-by-link.tsx
```diff
@@ -19,7 +19,7 @@ function init(): void {
 
 	select('.d-block.mb-2[href^="/contact"]')!.after(
 		<a href={String(actionURL)} className="d-block mb-2">
-			<SearchIcon width={14} className="text-gray-dark mr-2"/>Usage examples
+			<SearchIcon width={14} className={SearchIcon.defaultProps.className + ' text-gray-dark mr-2'}/>Usage examples
```
fregante, line 22: Ugly.

Also the types seem to be broken
 - because `SearchIcon` is now `any`
	 - because `React.FC` is `any` in octagon’s types.

Did https://github.com/vadimdemedes/dom-chef/pull/62 break anything here or has it always been like this? We also have this In our `globals.d.ts`:

```ts
declare module 'react' {
	const FC = (): JSX.Element => JSX.Element;
	const React = {FC};
	export default React;
}
```

fregante: @TheRealSyler have you been using version 5 of dom-chef since your PR? Do you know whether this block needs to be dropped or included in dom-chef?

source/features/download-folder-button.tsx
```diff
@@ -26,7 +26,7 @@ function init(): void {
 				className="btn ml-2"
 				href={downloadUrl.href}
 			>
-				<DownloadIcon className="mr-1"/>
+				<DownloadIcon className={DownloadIcon.defaultProps.className + ' mr-1'}/>
```
fregante, line 29: Maybe it should just use
```suggestion
				<DownloadIcon className="octicon octicon-download mr-1"/>
```

yakov116: 👍

fregante:
Closing because that PR is inactive

PR #3903, opened 2021-01-18
Improve automatic dark mode support

shinenelson:
for `prefers-color-scheme: dark` @media query
and `[data-color-mode: auto]`

1. LINKED ISSUES:
   https://github.com/sindresorhus/refined-github/pull/3806#issuecomment-761849635

2. TEST ENVIRONMENT:
   - Dark Mode is enabled for GitHub
   - System-wide dark mode is enabled
      - some way where the `<html>` tag has `data-color-mode: auto`

3. STEPS TO TEST
   1. Show Whitespace
      - open any source file
      - confirm whether whitespace indicators are displayed
   2. Conflict Markers
      - open a pull request with merge conflicts
      - confirm whether the :warning: sign is darker
   3. Reaction Avatars
      - open any issue page
      - react on any comment
      - confirm whether there are incorrect white borders on the reaction
        emoji or the reaction avatars

PR #3909, opened 2021-01-24
Improve automatic dark mode support

shinenelson:
re-opening #3903


shinenelson:
oops, I just realized that the changes from #3903 were already committed.

PR #3913, opened 2021-01-25
Fix heisenbug in `set-default-repositories-type-to-sources`

### LINKED ISSUE #3908: `set-default-repositories-type-to-sources`: undefined is not an object (evaluating 'link.search')
Screen Shot 1

1. Open a PR like https://github.com/sindresorhus/refined-github/pull/3904
2. Visit the PR Commits tab
3. Open the profile dropdown
	Screen Shot 3

This happens because:
1. `init` is called on every ajax load
2. `init` attaches a mutation handler
3. when the user opens the dropdown, the handler fires twice, but the second time the selector no longer finds anything

cheap-glitch:
Thanks for contributing! 🍄

1. LINKED ISSUES: Fixes #3908

2. TEST URLS: https://github.com/sindresorhus/refined-github/pull/3904 > Go to "Commits" tab > Hover on your avatar

3. SCREENSHOT: -

Changing the selector fixes this, but maybe it's better to check the return value of `select()` instead of applying the modification twice?


fregante:
It’s best to fix the logic by splitting the init function in 2

source/features/set-default-repositories-type-to-sources.tsx
```diff
@@ -26,7 +26,7 @@ async function init(): Promise<void> {
 
 	// "Your repositories" in header dropdown
 	await onProfileDropdownLoad();
-	addSourceTypeToLink(select('.header-nav-current-user ~ a[href$="tab=repositories"]')!);
+	addSourceTypeToLink(select('.header-nav-current-user ~ a[href*="tab=repositories"]')!);
 }
 
 void features.add(__filebasename, {
```
fregante, line 32: I think it would be something like
```suggestion
function profileDropdown () {
	await onProfileDropdownLoad();
	addSourceTypeToLink(select('.header-nav-current-user ~ a[href$="tab=repositories"]')!);		 
}

void features.add(__filebasename, {
 	init: once(profileDropdown)
 }, {
```

source/features/set-default-repositories-type-to-sources.tsx
```diff
@@ -1,4 +1,5 @@
 import select from 'select-dom';
+import once from 'once';
```
yakov116, line 2: Look at other features. We use a package oneTime

fregante (approved):

Issue #3848, opened 2020-12-27
Releases panel is removed (without `releases-tab`)


altrisi:
Edit: This (not being able to access releases at all with only one of the features enabled) has been confirmed as WAI. There is a workaround userscript here: https://github.com/sindresorhus/refined-github/issues/3848#issuecomment-751515243.

---

The releases panel on the right of every repo is removed (since one of the latter updates of the extension), even when `releases-tab` is disabled.

The panel is present when disabling the extension.

This happens in any repo (with releases, at least).

Example URL where the bug appears: https://github.com/sindresorhus/refined-github

Screenshot:
Extension disabled:
![imagen](https://user-images.githubusercontent.com/17107132/103175313-559c3a80-4869-11eb-9ad8-75f5ee1c0b67.png) 
Extension enabled:
![imagen](https://user-images.githubusercontent.com/17107132/103175319-65b41a00-4869-11eb-93c7-64252b147642.png)

yakov116:
A few features were merged. To unhide that disable `clean-repo-sidebar` https://github.com/sindresorhus/refined-github/pull/3783

altrisi:
But, shouldn't that feature hide only empty panels?

yakov116:
Well we are showing the release tab as part of the other tabs.

altrisi:
Except when `releases-tab` is disabled, like in this issue, or is it present somewhere else?

busches:
@altrisi we don't have a rule to not hide that section when the releases tab feature is disabled. It's possible to add it, but this type of interaction occurs with plenty of features and we don't code around them either. I'm not in favor of coupling them together and instead letting the user disable both.

fregante:
That part of `releases-tab` was consolidated into `clean-repo-sidebar` as requested in https://github.com/sindresorhus/refined-github/issues/3706

altrisi:
Then, there is no way except from manually changing the url of getting into the releases page without enabling `releases-tab` or disabling `clean-repo-sidebar`.

fregante:
Correct. Either leave both enabled or disable both.

altrisi:
That makes no sense to me, but well, it's your extension. I suppose it's ok to make things not work at all if that's what you want.

I can't find where it was requested, though.

fregante:
Yesterday `clean-repo-sidebar` didn’t even exist, now you’re complaining that you want to keep it enabled. 

And why do you want to disable `releases-tab`?

[...]

PR #3918, opened 2021-01-28
Fix error in `clean-repo-sidebar`

cheap-glitch:
Thanks for contributing! 🍄

1. LINKED ISSUES: -

2. TEST URLS: [Repo with releases disabled in the sidebar](https://github.com/co1ncidence/mountaineer.vim)

This PR prevents an error in the console, when the "Releases" section in the repo sidebar has been deactivated by the owner.

### DIFF
--- a/source/features/clean-repo-sidebar.tsx
+++ b/source/features/clean-repo-sidebar.tsx
@@ -10,7 +10,9 @@ async function init(): Promise<void> {
 
 	// Hide redundant "Releases" section
 	const sidebarReleases = await elementReady('.BorderGrid-cell a[href$="/releases"]', {waitForChildren: false});
-	sidebarReleases!.closest<HTMLElement>('.BorderGrid-row')!.hidden = true;
+	if (sidebarReleases) {
+		sidebarReleases.closest<HTMLElement>('.BorderGrid-row')!.hidden = true;
+	}
 
 	// Hide empty "Packages" section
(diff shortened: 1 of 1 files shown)


fregante:
I'm leaning towards dropping this part of the feature altogether. Re: https://github.com/sindresorhus/refined-github/issues/3848

cheap-glitch:
> I'm leaning towards dropping this part of the feature altogether. Re: #3848

I don't feel strongly about this either way, but I do find the release link in the sidebar slightly more informative.

> Maybe we can just drop the "Releases" header instead, like we do for "Labels" in `clean-conversation-sidebar`

Then I assume this would be in a separate feature?

fregante:
> Then I assume this would be in a separate feature?

We don't need to fix this if we're removing it in a next PR. Either we delete the whole block or edit it here so that it drops just the title.

fregante:
> Then I assume this would be in a separate feature?

We don't need to fix this if we're removing it in a next PR. This PR should either:

- delete the `sidebarReleases` code entirely, or
- edit this piece of code so that it just hides the "Releases" title

cheap-glitch:
I've tried hiding the "Releases" title only:
<table>
<tr>
	<th> Vanilla
	<th> RGH
<tr>
	<td>

Screenshot_2021-01-30 sindresorhus refined-github(1)
	</td>
	<td>

Screenshot_2021-01-30 sindresorhus refined-github
	</td>
</tr>
</table>

fregante (approved):

fregante:
I hope this is gonna make people in #3848 happy

PR #3919, opened 2021-01-28
Fix `deprioritize-marketplace-link` error in some instances

### LINKED ISSUE #3898: `deprioritize-marketplace-link`: Argument 1 ('target') to MutationObserver.observe must be an instance of Node
On every page (I see it on this issue’s page), probably since #3749, but it's not consistent.

Screen Shot 8

cheap-glitch:
Thanks for contributing! 🍄

1. LINKED ISSUES: Fixes #3898 

2. TEST URLS: Go to [the PR list](https://github.com/sindresorhus/refined-github/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc) and do a few "hard refresh" (Ctrl-F5). This triggers the error almost every time for me.

I think #3898 was caused by `awaitDomReady: false` combined with the plain `document.querySelector()` in `onProfileDropdownLoad()`. Using `elementReady` seems to fix it.

Also `set-default-repositories-type-to-sources` is unaffected.

### DIFF
--- a/source/github-events/on-profile-dropdown-load.ts
+++ b/source/github-events/on-profile-dropdown-load.ts
@@ -1,10 +1,15 @@
+import select from 'select-dom';
 import oneMutation from 'one-mutation';
 import elementReady from 'element-ready';
 
 export default async function onProfileDropdownLoad(): Promise<Element> {
 	const dropdown = await elementReady('.Header-item:last-child .dropdown-menu');
-	if (!exists()) {
-		await oneMutation(dropdown!, {childList: true, filter: exists});		
+	if (!dropdownContentExists()) {
+		await oneMutation(dropdown!, {childList: true, filter: dropdownContentExists});
 	}
 	return dropdown!;
+}
+
+function dropdownContentExists() {
+	return select.exists('.Header-item:last-child .dropdown-menu .dropdown-item')
 }
--- a/source/github-events/on-profile-dropdown-load.ts
+++ b/source/github-events/on-profile-dropdown-load.ts
@@ -2,7 +2,9 @@ import oneMutation from 'one-mutation';
 import elementReady from 'element-ready';
 
 export default async function onProfileDropdownLoad(): Promise<Element> {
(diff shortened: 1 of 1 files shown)


source/github-events/on-profile-dropdown-load.ts
```diff
@@ -1,7 +1,8 @@
 import oneMutation from 'one-mutation';
+import elementReady from 'element-ready';
 
 export default async function onProfileDropdownLoad(): Promise<Element> {
-	const dropdown = document.querySelector('.Header-item:last-child .dropdown-menu')!;
-	await oneMutation(dropdown, {childList: true});
-	return dropdown;
+	const dropdown = await elementReady('.Header-item:last-child .dropdown-menu', {waitForChildren: false});
+	await oneMutation(dropdown!, {childList: true});
```
fregante, line 6: I think at least one of these must be changed:

- `waitForChildren: false` means that the promise might resolve before the `include-fragment` inside has been parsed, so `oneMutation` might just catch that when it appears in the DOM, but that's not the element we want.
- `oneMutation` probably needs a `filter` function to ensure that the mutation actually contains the element we expect.
- `oneMutation` might never be called if the code is executed too late and the mutation already happened.

What if we take a shortcut and use `selector-observer` instead of both? Maybe something like:


```suggestion
	const dropdown = await new Promise(resolve => {
		const observer = observe('the final dropdown content', {
			add(element) {
				observer.disconnect();
				resolve(element);
			}
		})
	})
```

fregante: Actually it might be an overkill. How about
```suggestion
	const dropdown = await elementReady('.Header-item:last-child .dropdown-menu');
	if (!exists()) {
		await oneMutation(dropdown!, {childList: true, filter: exists});		
	}
```

and then:

```js
function exists() {
	return select.exists('the final dropdown content')
}
```

fregante (approved):
LGTM if it works

fregante:
woot woot for bugfixes!

PR #3922, opened 2021-01-29
Shorten and fix `separate-draft-pr-button`

fregante:
Fixes #3914

Tested in Chrome, but I did not have the problem in the first place.

@mischnic @MrChocolatine can you test this?

```sh
# git clone repo
npm install
npm run build
# manually load the directory "distribution" in Chrome
```

@busches do you see any issues on GHE?

Issue #3699, opened 2020-11-04
`update-pr-from-base-branch` doesn't add buttons for failing checks


natelaws:
Continued from #3155, I'm running into two new scenarios where the button isn't showing up:

- If you don't have the required number of approvers
- If required status checks are failing

<img width="860" alt="Screen Shot 2020-11-04 at 2 59 45 PM" src="https://user-images.githubusercontent.com/26171/98161942-8fc30e80-1eae-11eb-9011-e53a34a369d2.png">
<img width="858" alt="Screen Shot 2020-11-04 at 3 00 07 PM" src="https://user-images.githubusercontent.com/26171/98161943-905ba500-1eae-11eb-9169-6c0fb021cab4.png">

A partial workaround is to switch it back to a draft PR, but thats a little awkward. 

(Thanks for the great extension!)

PR #3929, opened 2021-01-31
Fix `update-pr-from-base-branch` duplication

### LINKED ISSUE #3927: Text from `update-pr-from-base-branch` is duplicated
image
image

I can't really send links since only maintainers and PR authors can update the base branch, but it seems to happen on any PR I can update the base branch on.

yakov116:
1. LINKED ISSUES:
   Fixes #3927 

2. TEST URLS:
   https://github.com/sindresorhus/refined-github/pull/3847 (this needed 800)
https://github.com/sindresorhus/refined-github/pull/3922

3. SCREENSHOT:
   None

I went with debounce over select observer. My thinking was that this is the only issue.

### DIFF
--- a/source/features/update-pr-from-base-branch.tsx
+++ b/source/features/update-pr-from-base-branch.tsx
@@ -1,7 +1,7 @@
 import React from 'dom-chef';
 import select from 'select-dom';
-import debounce from 'debounce-fn';
 import delegate from 'delegate-it';
+import {observe} from 'selector-observer';
 import {AlertIcon} from '@primer/octicons-react';
 import * as pageDetect from 'github-url-detection';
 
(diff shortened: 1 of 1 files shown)


fregante:
Debounce delays the first API call by 800ms. Can you look into why the API call isn’t being cached? What event(s) is calling this function 3 times?

yakov116:
Looks like there are a few events that are happening. I could not figure them all out... So select-observer it is.

fregante:
Can the observer’s selector be made more specific so we don’t have to use stillLoading? It should observe the final element that we need, it’s called once it’s found. No other listener should call addButton

yakov116:
> Can the observer’s selector be made more specific so we don’t have to use stillLoading?

I think it could will check later tonight

fregante:
Might be worth looking for a selector that also works on failing PRs https://github.com/sindresorhus/refined-github/issues/3699

fregante:
I got an idea

PR #3939, opened 2021-02-03
Fix `table-input` at 90% zoom

### LINKED ISSUE #3936: Feature `table-input` seems broken
1. The bug is cause by Refined GitHub

2. URL where the bug appears

Whenever you have a text editor, where you can insert a table like:
- https://github.com/sindresorhus/refined-github/issues/new
- Or simply when replying to this current issue

3. Video

https://user-images.githubusercontent.com/47531779/106594332-dc6bc380-6549-11eb-9b7b-4d3190d95f04.mov

cheap-glitch:
Thanks for contributing! 🍄

1. LINKED ISSUES: Fixes #3936 

2. TEST URLS: any rich text editor

The bug was caused by the fixed width of `108px`. Setting it to `auto` and using CSS Grid for the layout solves the problem.

Tested in Firefox & Chromium only.

### DIFF
--- a/source/features/table-input.css
+++ b/source/features/table-input.css
@@ -1,9 +1,9 @@
 :root .rgh-table-input {
-	width: 108px;
+	width: auto;
 	padding: 3px;
 	z-index: 99;
-	display: flex;
-	flex-wrap: wrap;
+	display: grid;
+	grid-template: repeat(5, 20px) / repeat(5, 20px);
 }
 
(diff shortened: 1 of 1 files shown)


MrChocolatine:
Tested locally, with several negative and positive zooms, and it works correctly.
Thanks @cheap-glitch.

fregante:
Thanks! It turns out grid was the safest option after all 😁

PR #3944, opened 2021-02-05
Fix merged button text in `extend-conversation-status-filters`

### LINKED ISSUE #3781: Filtering for `Merged` in the PR view hides `Open` and `Closed` options
When clicking on the `Merged` filter in the head of the PR list (same for sindresorhus/refined-github), the `Open` and `Closed` filters are gone and it requires an additional step to go back to see only open ones. The only option given is to click on `Total` and get a mixed view which is rarely useful/wanted.
I think there is an issue with the count for `Total` (suggests all PR's) as well, looks more like the number of all merged PR's.

Running version 20.11.26 with FF 83.0 on Windows 10.

### Screenshots
- `Open` selected
open

- `Closed` selected
closed

- `Merged` selected
merged

### What should be shown when clicking on `Merged`:
<kbd>10 Open</kbd> <kbd>1699 Closed</kbd> <kbd>**✓ 1472 Merged**</kbd>

yakov116:
1. LINKED ISSUES:
   Fixes #3781 

2. TEST URLS:
   https://github.com/sindresorhus/refined-github/pulls?q=is%3Apr+sort%3Aupdated-desc+is%3Amerged

3. SCREENSHOT:
   Before
image

After

image

The issue was that there are 2 buttons but returning breaks the loop.

### DIFF
--- a/source/features/extend-conversation-status-filters.tsx
+++ b/source/features/extend-conversation-status-filters.tsx
@@ -21,13 +21,13 @@ function addMergeLink(): void {
 		if (lastLinkQuery.includes('is:merged')) {
 			// It's a "Total" link for "is:merged"
 			lastLink.lastChild!.textContent = lastLink.lastChild!.textContent!.replace('Total', 'Merged');
-			return;
+			continue;
 		}
 
 		if (lastLinkQuery.includes('is:unmerged')) {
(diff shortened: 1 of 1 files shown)


fregante (approved):
Neat!

PR #3946, opened 2021-02-05
Fix console error in `list-prs-for-file`

yakov116:
Looks like I missed testing when there are no prs

### DIFF
--- a/source/features/comments-time-machine-links.tsx
+++ b/source/features/comments-time-machine-links.tsx
@@ -94,7 +94,7 @@ async function showTimemachineBar(): Promise<void | false> {
 		const lastCommitDate = await elementReady([
 			'.repository-content .Box.Box--condensed relative-time',
 			'[itemprop="dateModified"] relative-time' // "Repository refresh" layout
-		].join());
+		].join(), {waitForChildren: false});
 		if (date > lastCommitDate?.attributes.datetime.value!) {
 			return false;
 		}
--- a/source/features/comments-time-machine-links.tsx
+++ b/source/features/comments-time-machine-links.tsx
@@ -94,7 +94,7 @@ async function showTimemachineBar(): Promise<void | false> {
 		const lastCommitDate = await elementReady([
 			'.repository-content .Box.Box--condensed relative-time',
 			'[itemprop="dateModified"] relative-time' // "Repository refresh" layout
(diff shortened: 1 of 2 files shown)


source/features/list-prs-for-file.tsx
```diff
@@ -103,7 +103,7 @@ async function init(): Promise<void> {
 		prs = prs.filter(pr => pr !== Number(editingPRNumber));
 	}
 
-	if (prs.length === 0) {
+	if (!prs || prs.length === 0) {
```
fregante, line 106: Let’s have a single if `!pr, return` on line 100, so we don’t have to test pr twice

source/features/comments-time-machine-links.tsx
```diff
@@ -94,7 +94,7 @@ async function showTimemachineBar(): Promise<void | false> {
 		const lastCommitDate = await elementReady([
 			'.repository-content .Box.Box--condensed relative-time',
 			'[itemprop="dateModified"] relative-time' // "Repository refresh" layout
-		].join(), {waitForChildren: false});
+		].join());
```
fregante, line 97: I don’t think this is correct, we only use the attribute, not the children

fregante: Also if you want to drop all pre-refresh code you can do so in a new PR, as long as no other changes are done in the same PR, because they’d be lost.

source/features/list-prs-for-file.tsx
yakov116: But prs can be empty after the filter (len of 0)

fregante: Yeah, `if !prs` before and `prs.length === 0` after. It's still 2 separate `if`s but we don't have to use `pr ||` in 2 conditions

yakov116: Got it

source/features/comments-time-machine-links.tsx
yakov116: > I don’t think this is correct, we only use the attribute, not the children

Going to skip this for now as I don't have time to debug.

> Also if you want to drop all pre-refresh code you can do so in a new PR, as long as no other changes are done in the same PR, because they’d be lost.

Sure

fregante (commented):
🚀

Issue #3673, opened 2020-10-20
`latest-tag-button`, `default-branch-button` and `list-prs-for-file` are not fast/reliable enough


fregante:
1. Visit https://github.com/sindresorhus/refined-github/blob/42dfa16cd5366de26b475de94f8225003df07364/source/features/list-prs-for-file.tsx (or any file with an open PR) 
2. Refresh the page
3. Wait for each button to appear, if ever

![a212](https://user-images.githubusercontent.com/1402241/96651315-3679ad80-12fa-11eb-9c96-6436d9fa40b1.gif)

These are 3 separate features but they affect the same area. I think they should be reviewed to understand why they're so slow and why `latest-tag-button` sometimes doesn't appear at all. 

`default-branch-button` is the most annoying one because it's a simple feature with data that's easily available, it should be instant.

yakov116:
Debugged `default-branch-button`
If we move this
https://github.com/sindresorhus/refined-github/blob/017c17b36565186d87edb7618d781b0e87b206f3/source/features/default-branch-button.tsx#L19-L19
Before this
https://github.com/sindresorhus/refined-github/blob/017c17b36565186d87edb7618d781b0e87b206f3/source/features/default-branch-button.tsx#L13-L13

There is almost no jump at all. (not perfect but better then before)

fregante:
Makes sense! Any API calls should appear before `elementReady`, or else we _definitely_ wait too long.

fregante:
Here's an error:

```
refined-github.js:5657 ❌ Refined GitHub → latest-tag-button → TypeError: Cannot read property 'parentElement' of null
    at init (refined-github.js:6078)
    at async runFeature (refined-github.js:5697)
    at async setupPageLoad (refined-github.js:5710)
```

It can't find `#branch-select-menu`. It happened on the root of a fork:

https://github.com/fregante/refined-github

PR #3947, opened 2021-02-05
Make `latest-tag-button` more reliable

cheap-glitch:
Thanks for contributing! 🍄

1. LINKED ISSUES: -

2. TEST URLS: Go to a repo homepage (e.g. https://github.com/sindresorhus/refined-github) and hard-refresh a few times.

I noticed this feature failing with an error once in a while (especially when my connection is sluggish). Waiting for the `#branch-select-menu` element seems to fix it.

### DIFF
--- a/source/features/latest-tag-button.tsx
+++ b/source/features/latest-tag-button.tsx
@@ -1,7 +1,7 @@
 import './latest-tag-button.css';
 import React from 'dom-chef';
 import cache from 'webext-storage-cache';
-import select from 'select-dom';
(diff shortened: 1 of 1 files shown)


fregante:
Makes sense, thank you!

fregante:
It looks like this was mentioned in https://github.com/sindresorhus/refined-github/issues/3673

Does the other 2 feature need a similar treatment?

cheap-glitch:
> Does the other 2 feature need a similar treatment?

Not from what I can see.

Issue #3945, opened 2021-02-05
Copy button doesn't show (and other file-related features)


munael:
E.g.: https://github.com/TRSasasusu/SpikingMoYF/blob/master/network.py
![image](https://user-images.githubusercontent.com/5769148/107047735-d20f2b00-67d0-11eb-85e7-11f67024b288.png)

fregante:
I found this too. It appears after a refresh. It probably has something to do with Ajax navigation

yakov116:
This issue is very complex. Basically any time you navigate from a file selection any feature that needs to run does not.
This includes not just features that are within the file box. Even `fork-to` is affected.

fregante:
> Even `fork-to` is affected.

I don’t think so, that area is not updated when navigating the files, so the RG changes aren't lost nor do they need to be re-applied.

![gif](https://user-images.githubusercontent.com/1402241/108640914-0f76e800-7462-11eb-8930-a2758c369400.gif)


> This issue is very complex

This issue just means that we can't rely on `has-rgh` to deduplicate our features. Fixing this would mean making `features.add` aware of which features need to be run on filelists as well and having their own `filelist-has-rgh` file.

I think it's easier to just use per-feature checks on filelist features, as described in https://github.com/sindresorhus/refined-github/pull/3968#discussion_r574831236

fregante:
The comment on https://github.com/sindresorhus/refined-github/blob/61bfebbb2be4007c4b9a89b1e4c6cacea5123c31/source/features/download-folder-button.tsx#L25:L25 will also have to be updated as part of this PR

yakov116:
> I don’t think so, that area is not updated when navigating the files, so the RG changes aren't lost nor do they need to be re-applied.

When going to a single file fork-to is supposed to link to the file on the fork. (Maybe I am mixing up feature nanes)

fregante:
Ah crap, you may be right. That part of the feature isn't update.

It appears that the issue affects every feature on files and filelists. Another example:

1. Visit https://github.com/sindresorhus/refined-github/find/main
2. Type forkedto
3. Open the feature
4. No `rgh-feature-descriptions` enabled 😕 
5. Refresh the page
6. There it is…

kidonng:
Just FYI, navigation through Octotree does pair nicely with Refined GitHub (sadly they are closed source now, so it doesn't help much).

fregante:
[...]

PR #3950, opened 2021-02-06
Restore features in file lists

fregante:
Fixes https://github.com/sindresorhus/refined-github/issues/3945 and maybe more

It seems that there's a new nested ajax container. The `pjax:end` listener was being called, but `has-rgh` was not being removed by GitHub because GitHub changed a niece of `has-rgh` instead of its parent.


fregante:
As I feared, this actually broke other features. Merge undone, issue reopened.

Screen Shot 1

PR #3952, opened 2021-02-08
Move `download-folder-button` to the dropdown

### LINKED ISSUE #3949: Move `download-folder-button` to the dropdown
It looks like GitHub added a "Delete directory" button in a dropdown:

Screen Shot 1

Now there are too many buttons, so we should move `download-folder-button` to the dropdown. We can drop the icon and name use "Download directory" as label

yakov116:
1. LINKED ISSUES:
   Resolves #3949 

2. TEST URLS:
   https://github.com/sindresorhus/refined-github/tree/main/media

3. SCREENSHOT:

	image

I added `repeatOnBackButton: true,` since I found that when going from dir to dir it was not applying unless I refreshed (nothing to do with the current PR)

### DIFF
--- a/source/features/download-folder-button.tsx
+++ b/source/features/download-folder-button.tsx
@@ -41,6 +41,6 @@ void features.add(__filebasename, {
 	exclude: [
 		pageDetect.isRepoRoot // Already has an native download ZIP button
 	],
-	repeatOnBackButton: true,
+	repeatOnBackButton: true, // Due to anti-duplication mechanism #3945
 	init
 });
--- a/source/features/download-folder-button.tsx
+++ b/source/features/download-folder-button.tsx
@@ -22,12 +22,9 @@ function init(): void {
 	} else {
 		// "Repository refresh" layout
 		for (const deleteButton of select.all(`form[action^="/${getRepo()!.nameWithOwner}/tree/delete"]:not(.rgh-download-folder)`)) {
-			deleteButton.classList.add('rgh-download-folder');
+			deleteButton.classList.add('rgh-download-folder'); // TODO: Drop when #3945 is completely fixed
 			deleteButton.before(
-				<a
-					className="dropdown-item btn-link"
-					href={downloadUrl.href}
-				>
+				<a className="dropdown-item" href={downloadUrl.href}>
 					Download directory
 				</a>
 			);
--- a/source/features/download-folder-button.tsx
+++ b/source/features/download-folder-button.tsx
@@ -1,9 +1,9 @@
 import React from 'dom-chef';
 import select from 'select-dom';
-import {DownloadIcon} from '@primer/octicons-react';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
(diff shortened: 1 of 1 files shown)


fregante:
> I found that when going from dir to dir it was not applying 

That’s the same as #3945. The global anti-duplication mechanism is preventing the call of `init`s on file lists. It’s a tough problem 😕 it’s worth adding a comment pointing to #3945 on the `repeatOnBackButton` line

yakov116:
@fregante is this good?

source/features/download-folder-button.tsx
```diff
@@ -40,5 +42,6 @@ void features.add(__filebasename, {
 	exclude: [
 		pageDetect.isRepoRoot // Already has an native download ZIP button
 	],
+	repeatOnBackButton: true, // Due to anti-duplication mechanism #3945
```
fregante, line 45:
```suggestion
	repeatOnBackButton: true, // TODO: Drop when #3945 is completely fixed
```

source/features/download-folder-button.tsx
```diff
@@ -21,15 +21,17 @@ function init(): void {
...
-				Download
-			</a>
-		);
+		for (const deleteButton of select.all(`form[action^="/${getRepo()!.nameWithOwner}/tree/delete"]:not(.rgh-download-folder)`)) {
+			deleteButton.classList.add('rgh-download-folder');
+			deleteButton.before(
+				<a
+					className="dropdown-item btn-link"
+					href={downloadUrl.href}
+				>
```
fregante, line 30:
```suggestion
				<a className="dropdown-item btn-link" href={downloadUrl.href}>
```

fregante, line 30:
```suggestion
				<a className="dropdown-item btn-link" href={downloadUrl.href}>
```

Is `btn-link` required here? It's not a `<button>` to begin with

source/features/download-folder-button.tsx
fregante, line 45:
```suggestion
	repeatOnBackButton: true, // TODO: Drop when #3945 is completely fixed
```

source/features/download-folder-button.tsx
yakov116: I will double check

source/features/download-folder-button.tsx
```diff
@@ -21,15 +21,17 @@ function init(): void {
...
-			<a
-				className="btn ml-2"
-				href={downloadUrl.href}
-			>
-				<DownloadIcon className="mr-1"/>
-				Download
-			</a>
-		);
+		for (const deleteButton of select.all(`form[action^="/${getRepo()!.nameWithOwner}/tree/delete"]:not(.rgh-download-folder)`)) {
+			deleteButton.classList.add('rgh-download-folder');
```
yakov116, line 25: Add another TODO here?

fregante:
After the outstanding comments, feel free to merge this, #3962 and release a new version via the Actions repo tab 🚢

PR #3955, opened 2021-02-08
Add `useful-forks` feature

### LINKED ISSUE #3682: Add a link to `useful-forks`
The forks list page should have a button/link to a service that lists useful forks for the current project.

Feature location: https://github.com/sindresorhus/refined-github/network/members
Button URL: https://useful-forks.github.io?repository=sindresorhus/refined-github

This is a super easy feature, it just needs to create a link and position it somewhere on that page.

The button URL hasn't been implemented by https://github.com/useful-forks yet.

---

_Simplified request from https://github.com/sindresorhus/refined-github/issues/3638_

payne911:
1. LINKED ISSUES:
   fixes #3682

2. TEST URLS:
   * https://github.com/payne911/PieMenu/network
   * https://github.com/payne911/PieMenu/network/members

3. SCREENSHOT:
   image

   image


payne911:
The CI fails due to an unrelated file:

image

image

yakov116:
Can you run `npm run lint --fix`?

payne911:
I indeed had a few things to fix in my file.

Thanks!

yakov116:
@fregante I should we be checking if there are any forks? Also should private repo's be excluded?

readme.md
```diff
@@ -106,6 +106,7 @@ Thanks for contributing! 🦋🙌
 - [](# "link-to-github-io") [Adds a link to visit the user’s github.io website from its repo.](https://user-images.githubusercontent.com/31387795/94045261-dbcd5e80-fdec-11ea-83fa-30bb673cc26e.jpg)
 - [](# "next-scheduled-github-action") [Shows the next scheduled time of relevant GitHub Actions in the workflows sidebar.](https://user-images.githubusercontent.com/46634000/94690232-2476a180-0330-11eb-99d7-e174bb762cea.png)
 - [](# "quick-repo-deletion") [Lets you delete your repos in a click, if they have no stars, issues, or PRs.](https://user-images.githubusercontent.com/1402241/99716945-54a80a00-2a6e-11eb-9107-f3517a6ab1bc.gif)
+- [](# "useful-forks") [Adds a button to find relevant forks quickly.](https://user-images.githubusercontent.com/38117856/107256825-fe60bc80-6a07-11eb-93d5-7c3e8e20e922.png)
```
fregante, line 109:
```suggestion
- [](# "useful-forks") [Helps you find the most active forks.](https://user-images.githubusercontent.com/38117856/107256825-fe60bc80-6a07-11eb-93d5-7c3e8e20e922.png), via https://useful-forks.github.io.
```

Also I would include more of the UI in the screenshot so what's being pictured is clearer:

Screen Shot 1

An arrow may be necessary to point to the new button. Any style works, you can see other screenshots for reference.

If possible, zoom the page first so that the screenshot is in higher resolution.

source/features/useful-forks.tsx
```diff
@@ -0,0 +1,32 @@
...
+function init(): void {
+	const downloadUrl = new URL('https://useful-forks.github.io/?repo=' + pageDetect.utils.getRepoURL());
+
+	select('#network')!.prepend(
+		<a
+			className="btn mt-2 mb-2"
+			href={downloadUrl.href}
+			target="_blank"
+			rel="noreferrer"
+		>
```
fregante, line 17: No need for `target`
```suggestion
		<a className="btn mt-2 mb-2" href={downloadUrl.href}>
```

fregante: Also I'd align it on the right of the title, in both cases, with a `.float-right` class. Example:

Screen Shot 2

source/features/useful-forks.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'dom-chef';
+import select from 'select-dom';
+import {LinkExternalIcon} from '@primer/octicons-react';
+import * as pageDetect from 'github-url-detection';
+
+import features from '.';
+
+function init(): void {
+	const downloadUrl = new URL('https://useful-forks.github.io/?repo=' + pageDetect.utils.getRepoURL());
```
fregante, line 9:
```suggestion
	const downloadUrl = new URL('https://useful-forks.github.io');
	downloadUrl.searchParams.set('repo', pageDetect.utils.getRepoURL());
```

source/features/useful-forks.tsx
payne911: My personal preference would be to have it opened on a new tab. Is that generally discouraged by ``refined-github`` ?

source/features/useful-forks.tsx
payne911: Okay, I'm now using `getRepo()`.

However, I was aware that `downloadUrl.searchParams.set` existed, but it causes a bug, so I went for manual concatenation.

The thing encodes the special characters in HTML format, likeso:

image

I would expect `payne911/PieMenu`, but get `payne911%2FPieMenu` instead.

source/features/useful-forks.tsx
payne911: The `float-right` looks nice on the `Network` page, but on the `Forks` it looks like it's just randomly floating in a corner:

image

image

fregante: > have it opened on a new tab. Is that generally discouraged by `refined-github` ?

It's discouraged everywhere. Generally links don't need to open in a new tab unless they break the current flow (e.g. the user is about to buy something or fields haven't been saved). The user can choose where to open the links

fregante: > but on the `Forks` it looks like it's just randomly floating in a corner

Yeah that page is rather empty, but generally we try to avoid content shift (i.e. adding a button and pushing links lower, which can cause misclicks).

source/features/useful-forks.tsx
fregante: > I would expect `payne911/PieMenu`, but get `payne911%2FPieMenu` instead.

That's because you should use the same thing in useful-forks as well. The URL parameters should be correctly encoded/unencoded. In this case:

```js
const repo = new URLSearchParams(location.search).get('repo');

Comment (payne911) [reply source/features/useful-forks.tsx:9]: All fixed! Thanks.

**Comment by payne911** [comment]
image

Not sure what is wrong, to be honest.

**Comment by yakov116** [comment]
> About the lint: not sure what is wrong, to be honest

> Can you run `npm run lint --fix`?

File: source/features/useful-forks.tsx [hunk 4]
```diff
@@ -0,0 +1,29 @@
+import React from 'dom-chef';
+import select from 'select-dom';
+import {RepoForkedIcon} from '@primer/octicons-react';
+import * as pageDetect from 'github-url-detection';
+import {getRepo} from '../github-helpers';
+
+import features from '.';
```
yakov116, line 7:
```suggestion
import React from 'dom-chef';
import select from 'select-dom';
import * as pageDetect from 'github-url-detection';
import {RepoForkedIcon} from '@primer/octicons-react';

import features from '.';
import {getRepo} from '../github-helpers';
```

This should also fix the lint error

yakov116:
> For that second point, I'm not exactly sure how I should go about it. If I was displaying the button only in the `Forks` page, then I could count the amount of repos listed, but since we're trying to support the `Network` page too, I don't think I can easily achieve this goal without requiring an API call ?

You can check if the fork icon is disabled.

source/features/useful-forks.tsx
payne911: I did run the `npm run lint --fix`, but I found the message very cryptic. I don't see what is wrong with the way the imports were before and why what you did actually fixes anything.

payne911:
> You can check if the fork icon is disabled.

I just looked at one of my public repositories which has no forks and I see no difference between the fork icons. Would you mind expanding on your idea a bit more, please?

yakov116:
> I just looked at one of my public repositories which has no forks and I see no difference between the fork icons. Would you mind expanding on your idea a bit more, please?

I was talking about private repo's

payne911:
> I was talking about private repo's

I just looked at one of my private repo which has no forks and the icon looks the same.

source/features/useful-forks.tsx
payne911: Okay, I think I get it: it's separating imports of external dependencies and imports from within the same source repository.

yakov116:
$(".social-count[href$='/network/members']") that will give you how many forks there are.

```js
if (looseParseInt(select(".social-count[href$='/network/members']"))=== 0) {
	return false
}
```

payne911:
@yakov116 Okay, thanks! I didn't realize that number was there because it's visually hidden by `refined-github`.

payne911:
Okay, all that's left is pure aesthetics, and to be brutally honest, I'm not really a fan of messing for hours with CSS selectors.

Would you mind picking up from here, and updating the `readme` with an appropriate screenshot once that's done ?

fregante:
If anyone is willing to send a full PR, they can open a new one an pick up from here.

payne911:
> If anyone is willing to send a full PR, they can open a new one an pick up from here.

@fregante would you mind expressing yourself on where you think the button should be placed? It's literally the only thing left to be done. If it was fine where it is right now, then this PR is pretty much done.

fregante:
Leaving a PR unfinished defeats the purpose of sending a PR, especially when it's trivial to implement. If you don't have time to finish it, that's fine.

The position has been discussed and you already marked that discussion as Resolved. What's missing is, at this point:

- [ ] Making it work. It errors right now: `❌ Refined GitHub → useful-forks → DOMException: Failed to execute 'querySelector' on 'Document': '.social-count[href$="/network/members"!]' is not a valid selector.`
- [ ] Sending a PR to https://github.com/fregante/github-url-detection with the new page detections, including tests
- [ ] Using `elementReady` so that we don't have to wait for the whole page to load (especially a problem when there are lots of forks, e.g. https://github.com/sindresorhus/refined-github/network/members)
- [ ] Updating the screenshot once it's finished

fregante:
Also it causes the whole content to shift, so its selector needs to be updated

gif

payne911:
> Making it work. It errors right now

My bad, that's fixed.

> Sending a PR to https://github.com/fregante/github-url-detection with the new page detections, including tests

Done: https://github.com/fregante/github-url-detection/pull/58/

> Using `elementReady` so that we don't have to wait for the whole page to load

Not really sure how to use that properly. But I tried the current version on [the URL you showcased](https://github.com/sindresorhus/refined-github/network/members) and it seems like it's alright now? If not, I would appreciate a bit of guidance on how I should use `elementReady` properly.

My domain is Java backend, hence my lack of knowledge in terms of TypeScript and CSS, sorry.

> Updating the screenshot once it's finished

What do you think of those? (I use `kryonet` because it's a good example.)

Option 1:

image

Option 2:

image

Option 3:

useful-forks

I decided to use option 1 for now.

payne911:
(Since the PR is closed, I think it didn't pick up on the latest commits I push to my branch.)

yakov116:
Try to push now

yakov116:
@payne911 are you ok if I push some code?

payne911:
@yakov116 please do! I take absolutely no offense, and in fact rejoice at your care. :)

payne911:
That looks like solid changes. Thanks!

payne911:
Let me guess: since the `package` file is VCS's, it needs to be updated and pushed so that those "new" imports can be recognized ?

source/features/useful-forks.tsx
```diff
@@ -0,0 +1,35 @@
+import React from 'dom-chef';
+import elementReady from 'element-ready';
+import {isRepoForksList, isRepoNetworkGraph} from 'github-url-detection';
```
yakov116, line 3:
```suggestion
import * as pageDetect from 'github-url-detection';
```

source/features/useful-forks.tsx
```diff
@@ -0,0 +1,35 @@
...
+async function init(): Promise<void | false> {
+	const forkCount = looseParseInt((await elementReady('.social-count[href$="/network/members"]', {waitForChildren: false}))!);
+	if (forkCount === 0) {
+		return false;
+	}
+
+	const downloadUrl = new URL('https://useful-forks.github.io');
+	downloadUrl.searchParams.set('repo', getRepo()!.nameWithOwner);
+
+	const selector = isRepoForksList() ? '#network' : '#repo-content-pjax-container h2';
```
yakov116, line 19:
```suggestion
	const selector = pageDetect.isRepoForksList() ? '#network' : '#repo-content-pjax-container h2';
```

source/features/useful-forks.tsx
```diff
@@ -0,0 +1,35 @@
...
+			<RepoForkedIcon className="mr-2"/>
+			Find useful forks
+		</a>
+	);
+}
+
+void features.add(__filebasename, {
+	include: [
+		isRepoForksList,
+		isRepoNetworkGraph
```
yakov116, line 31: pageDetect.isRepoForksList,
pageDetect.isRepoNetworkGraph

yakov116 (changes requested):

payne911:
@yakov116 Sorry, I pushed as you made your suggestions. I looked at how the other features were doing it and decided to try doing it like them.

yakov116:
You need to update the `github-url-detection` package too. Commit the package.json and lock file. It should only have that change.

payne911:
> It should only have that change.

True for `package.json`, but the lock file has 21 differences, including a downgrade of `lockfileVersion`.

source/features/useful-forks.tsx
```diff
@@ -0,0 +1,35 @@
...
+async function init(): Promise<void | false> {
+	const forkCount = looseParseInt((await elementReady('.social-count[href$="/network/members"]', {waitForChildren: false}))!);
+	if (forkCount === 0) {
+		return false;
+	}
+
+	const downloadUrl = new URL('https://useful-forks.github.io');
+	downloadUrl.searchParams.set('repo', getRepo()!.nameWithOwner);
+
+	const selector = pageDetect.isRepoForksList() ? '#network' : '#repo-content-pjax-container h2';
```
fregante, line 19: If you just use the merged selector, does it work? 
```suggestion
	const selector = '#network, #repo-content-pjax-container h2';
```

source/features/useful-forks.tsx
```diff
@@ -0,0 +1,35 @@
...
+	(await elementReady(selector, {waitForChildren: false}))!.prepend(
```
fregante, line 20: Too many punctuation characters on one line make it a little hard to parse. Also if the suggestion above applies, the selector can be inlined in here
```suggestion
	const container = await elementReady(selector, {waitForChildren: false});
	container!.prepend(
```

readme.md
```diff
@@ -106,6 +106,7 @@ Thanks for contributing! 🦋🙌
 - [](# "link-to-github-io") [Adds a link to visit the user’s github.io website from its repo.](https://user-images.githubusercontent.com/31387795/94045261-dbcd5e80-fdec-11ea-83fa-30bb673cc26e.jpg)
 - [](# "next-scheduled-github-action") [Shows the next scheduled time of relevant GitHub Actions in the workflows sidebar.](https://user-images.githubusercontent.com/46634000/94690232-2476a180-0330-11eb-99d7-e174bb762cea.png)
 - [](# "quick-repo-deletion") [Lets you delete your repos in a click, if they have no stars, issues, or PRs.](https://user-images.githubusercontent.com/1402241/99716945-54a80a00-2a6e-11eb-9107-f3517a6ab1bc.gif)
+- [](# "useful-forks") [Helps you find the most active forks.](https://user-images.githubusercontent.com/38117856/107463541-542e8500-6b2c-11eb-8b25-082f344c1587.png), via https://useful-forks.github.io.
```
fregante, line 109: Option 1 or 3 work for me. @yakov116?

source/features/useful-forks.tsx
yakov116: @payne911 please check.

I would want 
```suggestion
	const selector = [
		'#network', //isRepoForksList
		'#repo-content-pjax-container h2' //isRepoNetworkGraph
	].join()
```

readme.md
yakov116: Option 1

payne911: Option 1 has already been committed, so we're good. :)

yakov116:
@payne911 Please check this https://github.com/sindresorhus/refined-github/pull/3955#pullrequestreview-588262057

source/features/useful-forks.tsx
payne911: I'm not sure I understand those suggestions. `#network` appears in both pages: this seems unsafe to me.

I just tested it, and somehow `isRepoNetworkGraph` still ends up ignoring the `#network` and going for the `#repo-content-pjax-container h2` appropriately, but I've no clue why that is happening. If it used `#network`, the button would not be placed at the same position as it is now.

payne911: (As an unrelated side note, I don't like that linting rule which doesn't allow me to adopt the cleaner ternary operator syntax that I prefer because it seems more readable)

```typescript
cond()
  ? whenTrue()
  : whenFalse();
```

yakov116: > I'm not sure I understand those suggestions. `#network` appears in both pages: this seems unsafe to me.

So then this cannot be done. Its good the way it is

yakov116 (approved):
LGTM

source/features/useful-forks.tsx
payne911: I had tried to explore a CSS selector which would be valid in both cases. I found one, but it was really ugly, hacky, and susceptible to breaking rapidly in terms of UI changes GitHub might make.

One possible other CSS selector idea I had in mind, I just didn't know if it was even possible. It was something along the lines of
> "from `#network`, take the parent `div`, and then select the first child `div` of that parent"

But that's also ugly, hacky, and britle.

fregante: > somehow `isRepoNetworkGraph` still ends up ignoring the `#network` and going for the `#repo-content-pjax-container h2` appropriately

To explain what's happening: `querySelector` (the API under `select`) just returns the first element that matches the selector. In this case, even if both are found on the page, probably `#repo-content-pjax-container h2` appears _earlier_ in the dom, therefore "#network" appears to be _ignored_

source/features/useful-forks.tsx
```diff
@@ -0,0 +1,36 @@
...
+import elementReady from 'element-ready';
+import * as pageDetect from 'github-url-detection';
+import {RepoForkedIcon} from '@primer/octicons-react';
+
+import features from '.';
+import {getRepo} from '../github-helpers';
+import looseParseInt from '../helpers/loose-parse-int';
+
+async function init(): Promise<void | false> {
+	const forkCount = looseParseInt((await elementReady('.social-count[href$="/network/members"]', {waitForChildren: false}))!);
```
fregante, line 11: The children in this case is a 2-character text node. Not important. Also, too many parens
```suggestion
	const forkCount = looseParseInt(await elementReady('.social-count[href$="/network/members"]')!);
```

fregante (approved):
Untested but LGTM. Just make sure it doesn't cause layout shift like before

source/features/useful-forks.tsx
yakov116: I think it needs the parens for lint

payne911: Fixed.

source/features/useful-forks.tsx
```diff
@@ -0,0 +1,37 @@
...
+import {RepoForkedIcon} from '@primer/octicons-react';
+
+import features from '.';
+import {getRepo} from '../github-helpers';
+import looseParseInt from '../helpers/loose-parse-int';
+
+async function init(): Promise<void | false> {
+	const forkCountBadgeValue = await elementReady('.social-count[href$="/network/members"]');
+	const forkCount = looseParseInt(forkCountBadgeValue!);
+	if (forkCount === 0) {
```
yakov116, line 13:
```suggestion
	const forkCount = await elementReady('.social-count[href$="/network/members"]')!;
	if (looseParseInt(forkCount) === 0) {
```

yakov116 (approved):

payne911:
Just tested this latest commit on the provided URLs, plus on `https://github.com/sindresorhus/refined-github/network/members` which was provided as an example of when it might displace stuff.

I personally didn't see any issue.

You could test it yourself too, just to be sure. :)

In any case: *ready to merge* !

source/features/useful-forks.tsx
payne911: Would've caused an error. I pushed a fix with aligns with this suggestion.

yakov116:
@payne911 Thanks for contributing!

First time contributing can be hard, but you made it through! ☕

payne911:
I even wrote a CSS selector. As a backend dev, that's a big deal! ;)

fregante:
Thanks @payne911! `useful-forks` is a useful tool!

payne911:
My pleasure!

I wish I could get to scan massive fork lists entirely, but unfortunately GitHub's rate limit is a bit low. :(

payne911:
@fregante @yakov116 Well, `useful-forks` has really gained in popularity ever since it was integrated: really cool to watch.

Anyways, I know this is none of your concerns, but I thought maybe you could help me with [a little problem](https://github.com/useful-forks/useful-forks.github.io/issues/13#issuecomment-787011933) since you guys have probably worked a lot with GitHub's API.

I thought my tool was heavily restricted by the Rate-Limit, but it turns out I barely reach 10% of it before seeing my requests getting blocked due to an "[abuse-detection mechanism](https://docs.github.com/en/rest/overview/resources-in-the-rest-api#abuse-rate-limits)" `403 error code`.

Have you guys ever run into such limitations? Any suggestions or ideas?

fregante:
The tool makes several API requests at the same time, I wouldn’t expect anything else. We expect the API key in most of our calls.

Anyway, I was thinking that this feature doesn’t apply to GitHub Enterprise, can you send a PR to exclude it from there?

Also, how does it handle private repositories? Does it just show a 404? In that case we should also exclude private repos.

fregante:
You could however use the octokit API library instead of manually doing fetches, it has automatic rate limit handling.

payne911:
> We expect the API key in most of our calls.

Yes, but do you have any feature which sends massive amounts of calls like this tool does?

> Anyway, I was thinking that this feature doesn’t apply to GitHub Enterprise, can you send a PR to exclude it from there?

Why exactly do you think it does not apply? (Excuse my ignorance.)

> Also, how does it handle private repositories? Does it just show a 404? In that case we should also exclude private repos.

I just looked up the [documentation for the `LIST forks` endpoint](https://docs.github.com/en/rest/reference/repos#list-forks) and it seems like it somehow also does list the private ones (since there is a `private` boolean attribute returned in the response), although I wouldn't know which repository to use to test this out.

But yes, that would be an easy fix.

> You could however use the octokit API library instead of manually doing fetches, it has automatic rate limit handling.

Yeah, I *just* discovered that thing. Which I knew about it earlier. Thanks for that!

fregante:
I recently implemented that octokit in a library that scans and removed one’s fork in batch. It’s pretty fast, it doesn’t just add a second. 

If it works with GraphQL, sure, but keep in mind that GraphQL **requires** a key. 

GH’s API only operates on GitHub.com. GHE has a whole different dataset and requires its own API address and key. GHE support doesn’t make sense for your tool anyway.

payne911:
> I recently implemented that octokit in a library that scans and removed one’s fork in batch. It’s pretty fast, it doesn’t just add a second.

Alright. I think that will be my next step then: update code to use `octokit`.

> If it works with GraphQL, sure, but keep in mind that GraphQL requires a key.

I didn't know that, and it's a pretty big deal. I think for now I'll set that aside, then. Thanks for the input.

> GH’s API only operates on GitHub.com. GHE has a whole different dataset and requires its own API address and key. GHE support doesn’t make sense for your tool anyway.

Another thing I didn't know. This will indeed have to be fixed. Is there a utility function in `refined-github` which detects if a user is on GHE ?

fregante:
> Is there a utility function in `refined-github` which detects if a user is on GHE ?

```diff
+ exclude: [
+   pageDetect.isEnterprise
+ ]
```

payne911:
```
+ exclude: [
+   pageDetect.isEnterprise
+ ]
```

Added! Thanks.

> It doesn't look like it, maybe there's a function somewhere in our code. It's worth a search and if not it should be written. Ideally page detections should be found in the <head>, but that doesn't always exist.

I couldn't quickly find anything, but the code-base is large, and I'm not certain about which keywords to look for.

In any case, it *is* possible that a large team works on a private repo and end up having different forks which could be useful to scan, no? All they'd have to do would be to provide an Access Token with the right permissions to allow the scan to work properly.

fregante:
> which keywords to look for

"private"

> having different forks which could be useful to scan, no

I mean, why. Even if there are 10 forks, they're not "useful". If they're looking for stray commits, the Network view already does that and is helpful for a few forks.

payne911:
> "private"

That yields "100+" results in IntelliJ:

image

which means I can't fully explore that avenue.

`isPrivate` yields 0 results, so I'm *guessing* that would've been the function name if it was in there and thus that it doesn't exist?

PR #3958, opened 2021-02-09
Add `link-to-changelog-file` feature

### LINKED ISSUE #3930: Link to CHANGELOG.md on Releases page
1. Visit https://github.com/nodeca/js-yaml/releases
2. See nothing useful
3. Navigate to https://github.com/nodeca/js-yaml/blob/master/CHANGELOG.md because thats there the changelog is kept

We can send a HEAD request to `$default/CHANGELOG.md` and link to it if it exists when visiting a releases page, perhaps just if the page doesn't have releases.

The unfortunate part is that the HTTP request is case-sensitive, so with this method we can only check and link to that specific but common format.

yakov116:
1. LINKED ISSUES:
   Closes #3930 

2. TEST URLS: (will add more later)
   https://github.com/nodeca/js-yaml/releases/tag/4.0.0
https://github.com/pyca/cryptography/releases

3. SCREENSHOT: (Will update when layout confirmed)
   
image

### DIFF
--- a/source/features/link-to-changelog-file.tsx
+++ b/source/features/link-to-changelog-file.tsx
@@ -28,10 +28,16 @@ async function fetchFromApi(): Promise<string | false > {
 		}
 	`);
 
-	return repository.object.entries.find((file: {name: string}) => file.name.toLowerCase().startsWith('changelog.'))?.name ?? false;
+	for (const file of repository.object.entries) {
+		if (file.name.toLowerCase().startsWith('changelog')) {
+			return file.name;
+		}
+	}
+
+	return false;
 }
 
-const doesChangelogExist = cache.function(async () => pageDetect.isRepoHome() ? parseFromDom() : fetchFromApi(), {
+const getChangelogName = cache.function(async () => pageDetect.isRepoHome() ? parseFromDom() : fetchFromApi(), {
 	cacheKey: getCacheKey
 });
 
@@ -48,10 +54,14 @@ async function init(): Promise<void | false> {
 		return false;
 	}
 
-	const url = buildRepoURL('blob', 'HEAD', changelog);
-	const tooltip = 'View the ' + changelog + ' file';
 	(await elementReady('.subnav div', {waitForChildren: false}))!.after(
-		<a className="btn ml-3 tooltipped tooltipped-n" aria-label={tooltip} href={url} style={{padding: '6px 16px'}} role="button">
+		<a
+			className="btn ml-3 tooltipped tooltipped-n"
+			aria-label={`View the ${changelog} file`}
+			href={buildRepoURL('blob', 'HEAD', changelog)}
+			style={{padding: '6px 16px'}}
+			role="button"
+		>
 			<BookIcon className="text-blue mr-2"/>
 			<span>Changelog</span>
 		</a>
--- a/source/features/link-to-changelog-file.tsx
+++ b/source/features/link-to-changelog-file.tsx
@@ -13,7 +13,7 @@ const getCacheKey = (): string => `changelog:${getRepo()!.nameWithOwner}`;
 
 async function parseFromDom(): Promise<false> {
 	await cache.delete(getCacheKey());
-	void cache.set(getCacheKey(), select('.js-navigation-item [title^="changelog." i]')?.textContent ?? false);
+	void cache.set(getCacheKey(), select('.js-navigation-item [title^="changelog" i]')?.textContent ?? false);
 	return false;
 }
 
--- a/source/features/link-to-changelog-file.tsx
+++ b/source/features/link-to-changelog-file.tsx
@@ -11,11 +11,13 @@ import {buildRepoURL, getRepo} from '../github-helpers';
 
 const getCacheKey = (): string => `changelog:${getRepo()!.nameWithOwner}`;
 
-function parseFromDom(): string | false {
-	return select('.js-navigation-item [title^="changelog." i]')?.textContent ?? false;
+async function parseFromDom(): Promise<false> {
+	await cache.delete(getCacheKey());
+	void cache.set(getCacheKey(), select('.js-navigation-item [title^="changelog." i]')?.textContent ?? false);
+	return false;
 }
 
-async function fetchFromApi(): Promise<string | false > {
+const getChangelogName = cache.function(async (): Promise<string | false > => {
 	const {repository} = await api.v4(`
 		repository() {
 			object(expression: "HEAD:") {
@@ -35,21 +37,12 @@ async function fetchFromApi(): Promise<string | false > {
 	}
 
 	return false;
-}
-
-const getChangelogName = cache.function(async () => pageDetect.isRepoHome() ? parseFromDom() : fetchFromApi(), {
+}, {
 	cacheKey: getCacheKey
 });
 
 async function init(): Promise<void | false> {
-	// Always prefer the information in the DOM
-	if (pageDetect.isRepoHome()) {
-		await cache.delete(getCacheKey());
-		void doesChangelogExist();
-		return;
-	}
-
-	const changelog = await doesChangelogExist();
+	const changelog = await getChangelogName();
 	if (!changelog) {
 		return false;
 	}
@@ -81,5 +74,5 @@ void features.add(__filebasename, {
 	exclude: [
 		pageDetect.isEmptyRepoRoot
 	],
-	init
+	init: parseFromDom
 });
--- a/source/features/link-to-changelog-file.tsx
+++ b/source/features/link-to-changelog-file.tsx
@@ -12,7 +12,7 @@ import {buildRepoURL, getRepo} from '../github-helpers';
 const getCacheKey = (): string => `changelog:${getRepo()!.nameWithOwner}`;
 
 function parseFromDom(): string | false {
-	return select('.js-navigation-item [title^="changelog" i]')?.textContent ?? false;
+	return select('.js-navigation-item [title^="changelog." i]')?.textContent ?? false;
 }
 
 async function fetchFromApi(): Promise<string | false > {
@@ -28,7 +28,7 @@ async function fetchFromApi(): Promise<string | false > {
 		}
 	`);
 
-	return repository.object.entries.find((file: {name: string}) => file.name.toLowerCase().startsWith('changelog'))?.name ?? false;
+	return repository.object.entries.find((file: {name: string}) => file.name.toLowerCase().startsWith('changelog.'))?.name ?? false;
 }
 
 const doesChangelogExist = cache.function(async () => pageDetect.isRepoHome() ? parseFromDom() : fetchFromApi(), {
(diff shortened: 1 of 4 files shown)


source/features/link-to-changelog.md.tsx
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	const changelog = await doesChangelogExist();
+	if (!changelog) {
+		return false;
+	}
+
+	const url = buildRepoURL('blob', 'HEAD', changelog);
+	select('.details-reset.Details-element')!.before(
+		<a className="btn btn-sm btn-invisible mt-2 p-0" href={url} type="button">
```
fregante, line 52: Wrong type argument

source/features/link-to-changelog.md.tsx
```diff
@@ -0,0 +1,75 @@
...
+		return;
+	}
+
+	const changelog = await doesChangelogExist();
+	if (!changelog) {
+		return false;
+	}
+
+	const url = buildRepoURL('blob', 'HEAD', changelog);
+	select('.details-reset.Details-element')!.before(
```
fregante, line 51: The changelog is global, so it shouldn’t be added next to a tag. Maybe it should be next to the page title if it exists or on the right (floating)

source/features/link-to-changelog.md.tsx
```diff
@@ -0,0 +1,75 @@
...
+		<a className="btn btn-sm btn-invisible mt-2 p-0" href={url} type="button">
+			<FileIcon/>
+			<span>View {changelog}</span>
+		</a>
+	);
+}
+
+void features.add(__filebasename, {
+	include: [
+		pageDetect.isSingleTag
```
fregante, line 61: I was expecting this to be on the main Releases page, not (just) on the single tag. Changelog.md  includes all the versions, not just one.

source/features/link-to-changelog.md.tsx
```diff
@@ -0,0 +1,75 @@
...
+				...on Tree {
+					entries {
+						name
+					}
+				}
+			}
+		}
+	`);
+
+	return repository.object.entries.find((file: {name: string}) => /changelog\.md/i.exec(file.name))?.name ?? false;
```
fregante, line 30: I’d use .toLowerCase().startsWith('changelog'). The extension could be different or missing altogether

source/features/link-to-changelog.md.tsx
yakov116: Can you give me a visual of what your thinking.

Do you want to put this together with `tag-changelog` (forgot the name) feature?

fregante: The example URL I used in the feature request was https://github.com/nodeca/js-yaml/releases, not https://github.com/nodeca/js-yaml/releases/tag/4.0.0 (but I think it can still appear in every `/releases/* page)

I'm expecting this link to appear above/outside the tags, 2 examples:

Screen Shot 1

yakov116: > The example URL I used in the feature request was [nodeca/js-yaml/releases](https://github.com/nodeca/js-yaml/releases), not [nodeca/js-yaml@`4.0.0` (release)](https://github.com/nodeca/js-yaml/releases/tag/4.0.0) (but I think it can still appear in every `/releases/*` page)

Got you. I was dealing with the _pain point_ that when going to a release you can't see the changelog.



> I'm expecting this link to appear above/outside the tags, 2 examples (I'm not super happy about either one):

Let me give it a shot and see what I can come up with.

yakov116:
@fregante 
image

Thoughts

fregante:
I like it!

source/features/link-to-changelog-file.tsx
```diff
@@ -0,0 +1,75 @@
...
+				...on Tree {
+					entries {
+						name
+					}
+				}
+			}
+		}
+	`);
+
+	return repository.object.entries.find((file: {name: string}) => file.name.toLowerCase().startsWith('changelog.'))?.name ?? false;
```
fregante, line 31: A bit clearer 

```suggestion
	for (const file of repository.object.entries) {
		if (file.name.toLowerCase().startsWith('changelog.')) {
			return file.name;
		}
	}

	return false;
```

fregante: Found a nice extension-less changelog https://github.com/genesi/gtk2/blob/master/ChangeLog

source/features/link-to-changelog-file.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+	const changelog = await doesChangelogExist();
+	if (!changelog) {
+		return false;
+	}
+
+	const url = buildRepoURL('blob', 'HEAD', changelog);
+	const tooltip = 'View the ' + changelog + ' file';
+	(await elementReady('.subnav div', {waitForChildren: false}))!.after(
+		<a className="btn ml-3 tooltipped tooltipped-n" aria-label={tooltip} href={url} style={{padding: '6px 16px'}} role="button">
```
fregante, line 54:
```suggestion
	(await elementReady('.subnav div', {waitForChildren: false}))!.after(
		<a
			className="btn ml-3 tooltipped tooltipped-n"
			aria-label={`View the ${changelog} file`}
			href={buildRepoURL('blob', 'HEAD', changelog)}
			style={{padding: '6px 16px'}}
			role="button"
		>
```

Also the padding seems arbitrary. Can this be solved with vertical-align or some flex property?

source/features/link-to-changelog-file.tsx
```diff
@@ -0,0 +1,75 @@
...
+	cacheKey: getCacheKey
+});
+
+async function init(): Promise<void | false> {
+	// Always prefer the information in the DOM
+	if (pageDetect.isRepoHome()) {
+		await cache.delete(getCacheKey());
+		void doesChangelogExist();
+		return;
+	}
```
fregante, line 44: I'd move this to its own init function and always return `false`

source/features/link-to-changelog-file.tsx
```diff
@@ -0,0 +1,75 @@
...
+					}
+				}
+			}
+		}
+	`);
+
+	return repository.object.entries.find((file: {name: string}) => file.name.toLowerCase().startsWith('changelog.'))?.name ?? false;
+}
+
+const doesChangelogExist = cache.function(async () => pageDetect.isRepoHome() ? parseFromDom() : fetchFromApi(), {
```
fregante, line 34:
```suggestion
const getChangelogName = cache.function(async () => pageDetect.isRepoHome() ? parseFromDom() : fetchFromApi(), {
```

source/features/link-to-changelog-file.tsx
```diff
@@ -0,0 +1,75 @@
...
+const getCacheKey = (): string => `changelog:${getRepo()!.nameWithOwner}`;
+
+function parseFromDom(): string | false {
+	return select('.js-navigation-item [title^="changelog." i]')?.textContent ?? false;
+}
+
+async function fetchFromApi(): Promise<string | false > {
+	const {repository} = await api.v4(`
+		repository() {
+			object(expression: "HEAD:") {
```
fregante, line 21: There's a `defaultBranchRef` I think

source/features/link-to-changelog-file.tsx
yakov116: The issue is what if someone has a change log folder? (or maybe that a good thing to catch that)

source/features/link-to-changelog-file.tsx
yakov116: The issue is that the [Releases/Tags] have a unique padding

source/features/link-to-changelog-file.tsx
yakov116:
```graphql
{
    defaultBranchRef {
      target {
        ... on Tree {
          entries {
            name
          }
        }
      }
    }
  }
}

```

Cant get this to work

yakov116:
Tested

source/features/link-to-changelog-file.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+import features from '.';
+import * as api from '../github-helpers/api';
+import {buildRepoURL, getRepo} from '../github-helpers';
+
+const getCacheKey = (): string => `changelog:${getRepo()!.nameWithOwner}`;
+
+async function parseFromDom(): Promise<false> {
+	await cache.delete(getCacheKey());
+	void cache.set(getCacheKey(), select('.js-navigation-item [title^="changelog." i]')?.textContent ?? false);
```
yakov116, line 16:
```suggestion
	void cache.set(getCacheKey(), select('.js-navigation-item [title^="changelog" i]')?.textContent ?? false);
```

source/features/link-to-changelog-file.tsx
fregante: Refer to other features, we use it elsewhere

yakov116: I know. But its not working (its an issue on github's end)

source/features/link-to-changelog-file.tsx
```diff
@@ -11,11 +11,13 @@ import {buildRepoURL, getRepo} from '../github-helpers';
 
 const getCacheKey = (): string => `changelog:${getRepo()!.nameWithOwner}`;
 
-function parseFromDom(): string | false {
-	return select('.js-navigation-item [title^="changelog." i]')?.textContent ?? false;
+async function parseFromDom(): Promise<false> {
+	await cache.delete(getCacheKey());
```
fregante, line 15: I don’t think this should be necessary. .set already overwrites it. 

If you copied this exact piece of code from other features, they should probably be updated too, but I think those used cache.delete because they called cache.function afterwards, not cache.set

yakov116: > but I think those used cache.delete because they called cache.function afterwards, not cache.set

correct

source/features/link-to-changelog-file.tsx
yakov116: https://github.community/t/on-tree-fails-on-defaultbranchref/161041

fregante: It confused me a bit too, but this made it clear:

```gql
defaultBranchRef {
	target {
		__typename
	}
}
```

The type of `target` is `Commit`, so `Tree` just does not match. To get the tree from `Commit` you use:

```gql
{
    defaultBranchRef {
      target {
        ... on Commit {
          tree {
            entries {
              name
            }
          }
        }
      }
    }
  }
}
```

But probably there's an even shorter way, avoiding `defaultBranchRef.target` altogether. 🤔

fregante: I'm not sure my example is worth it given the length.

yakov116: Should we get the default branch `await defaultBranch` so we dont have issues with branches/N/slashes?

fregante: This wouldn’t be any better than just using HEAD:

fregante (approved):
Maybe? 😅

641210 (commented):
****

PR #3968, opened 2021-02-11
Force `init` for some features

### LINKED ISSUE #3945: Copy button doesn't show (and other file-related features)

Esemesek:
Navigating to files doesn't reload the whole page. This causes some of the features to not initialize - a similar situation to `Go back` in the browser history.

Solution:

Added additional flag to force re-init on `pjax:end`, but it does the same thing as `repeatOnBackButton`. According to [pjax docs](https://github.com/defunkt/jquery-pjax#reinitializing-pluginswidget-on-new-page-content) this is something that we should do.

I added an additional flag for this one because the `repeatOnBackButton` name doesn't really fit in this case. I can merge those two with a better name if we agree that this solution is okay.

cc @fregante 

Closes #3945

### DIFF
--- a/source/features/index.tsx
+++ b/source/features/index.tsx
@@ -25,6 +25,9 @@ interface FeatureLoader extends Partial<InternalRunConfig> {
 	/** When pressing the back button, the DOM and listeners are still there, so normally `init` isn’t called again. If this is true, it’s called anyway.  @default false */
 	repeatOnBackButton?: true;
 
+	/** Force re-initialization on pjax:end event @default false */
+	forceReinit?: boolean;
+
 	/** When true, don’t run the `init` on page load but only add the `additionalListeners`. @default false */
 	onlyAdditionalListeners?: true;
 
@@ -211,6 +214,7 @@ const add = async (id: FeatureID, ...loaders: FeatureLoader[]): Promise<void> =>
 			awaitDomReady = true,
 			repeatOnBackButton = false,
 			onlyAdditionalListeners = false,
+			forceReinit = false,
 			additionalListeners = []
 		} = loader;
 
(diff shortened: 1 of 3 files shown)


yakov116:
This will lead to duplicate buttons

Esemesek:
@yakov116 Well, I was testing it multiple times and it never gave me multiple buttons. Additionally, it seems that we already had this 'solution' in place for the `download-folder-button` feature.

yakov116:
Because I added
https://github.com/Esemesek/refined-github/blob/5079d87442b4836752a928eaa81c1e68c78207ad/source/features/download-folder-button.tsx#L24-L25

source/features/index.tsx
```diff
@@ -237,7 +241,7 @@ const add = async (id: FeatureID, ...loaders: FeatureLoader[]): Promise<void> =>
 		}
 
 		document.addEventListener('pjax:end', () => {
-			if (repeatOnBackButton || !select.exists('has-rgh')) {
+			if (repeatOnBackButton || !select.exists('has-rgh') || forceReinit) {
```
fregante, line 244: As this line suggests, this is equivalent to `repeatOnBackButton: true`, just under a different name.

This issue is a little complex. Basically we have 2 solutions:

- rework how the automatic deduplication mechanism work (hard)
- rename `repeatOnBackButton: true` to `iAlreadyHandleDeduplicationThankYouVeryMuch: true` and handle deduplication via the usual `:not(.rgh-etc)` selector like we do for most features.

I think the latter is easier. This way we still keep `has-rgh` as a generic fallback but also have the opportunity to opt-out when it causes problems (like in this case)

Esemesek: This is exactly what I was asking about. Figured out the root of the issue, but I wasn't exactly sure how we want to tackle it - I think that I'll work on this `dedup` mechanism, and create another PR for it then.

Esemesek:
Closing this PR - will open another with a better solution.

PR #3971, opened 2021-02-12
Reimplement deduplication of features

### LINKED ISSUE #3945: Copy button doesn't show (and other file-related features)

Esemesek:
Proposing a solution for a new feature deduplication mechanism. Instead of creating one empty `has-rgh` element, now we create `rgh-feature` element with different attributes. Whenever a feature loads, it adds an attribute to `rgh-feature` tag, preventing it from being loaded again in `pjax:end`.

I was roughly testing it on different pages but didn't notice anything wrong or unusual. Also, I am not sure if we should remove an attribute for features that have `deinit` function.

cc. @fregante 

Closes #3945

### DIFF
--- a/source/features/download-folder-button.tsx
+++ b/source/features/download-folder-button.tsx
@@ -21,8 +21,7 @@ function init(): void {
 		);
 	} else {
 		// "Repository refresh" layout
-		for (const deleteButton of select.all(`form[action^="/${getRepo()!.nameWithOwner}/tree/delete"]:not(.rgh-download-folder)`)) {
-			deleteButton.classList.add('rgh-download-folder'); // TODO: Drop when #3945 is completely fixed
+		for (const deleteButton of select.all(`form[action^="/${getRepo()!.nameWithOwner}/tree/delete"]`)) {
 			deleteButton.before(
 				<a className="dropdown-item" href={downloadUrl.href}>
 					Download directory
@@ -39,6 +38,5 @@ void features.add(__filebasename, {
 	exclude: [
 		pageDetect.isRepoRoot // Already has an native download ZIP button
 	],
-	repeatOnBackButton: true, // TODO: Drop when #3945 is completely fixed
 	init
 });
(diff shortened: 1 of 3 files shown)


fregante:
Sorry this doesn’t solve the issue either. 

has-rgh worked because it marked the current ajaxed page as processed. Every time a new ajax page is loaded, the element is removed by GitHub and the features are run again. 

Currently this is failing on file lists because ajaxed views are nested, so even if we mark the upper view as processed, the file list is loaded *inside* it so the whole view is still marked as processed, because has-rgh hasn’t been replaced. 

Since you’re not changing the position of the element in the DOM, this doesn’t change anything. Try navigating into 3 nested folders, you’ll see that the copy file isn’t being added.

Esemesek:
@fregante I haven't added everything I had locally on the branch. Also, added resetting `rgh-features` on every `init`. This is causing features to reload every time but should fix the issue.

fregante:
If you remove the element at every pjax:end, you're defeating the purpose of even having it there. `init` is only generally called on first load and on pjax:end, so if you remove the the element like you do, it never `exists` when checked.

As the previous boolean option suggests (`repeatOnBackButton`), `has-rgh` was implemented to avoid this:

1. Visit https://github.com/sindresorhus/refined-github/blob/main/safari/app-store-description.txt
2. Click "Safari" (the parent folder)
3. _Go back_ in the browser

This is the current PR:
gif

fregante:
To explain more what's happening, when going back, `pjax:end` is fired again, but the DOM already our new elements (including `has-rgh`). Our code detects `has-rgh` and doesn't call `init` again.

This is identical to per-feature checks like `:not(.rgh-download-button)`, except that it's implemented globally instead of having to take care of it specifically in each feature.

As I suggested in #3968, the solution you're trying to find is hard to find (because of the nested ajaxed views) and it may be easier to just have per-feature checks (exactly like what we have now in `download-folder-button`, except that `repeatOnBackButton` should be renamed to make more sense)

PR #3973, opened 2021-02-14
Update dependencies 

yakov116:
Lint is Thanks to https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1045

I mainly did it to update the bugs octoicon to look like all the other new icons.

I was unable to update `size-plugin` I could not figure out why webpack stopped working.


sindresorhus:
> Lint is Thanks to sindresorhus/eslint-plugin-unicorn#1045

Note that it's not done and that I'm personally skeptical about it. Destructuring can be nice, but sometimes it makes things less readable. I think it's hard to automate when it makes sense.

yakov116:
> Note that it's not done and that I'm personally skeptical about it. Destructuring can be nice, but sometimes it makes things less readable. I think it's hard to automate when it makes sense.

I agree. 

I went through each one of the auto fixes to make sure that it makes sense.

yakov116:
@fregante any of them should be reverted?

fregante:
I agree with Sindre, but this PR looks _ok_ except the `pr.*` part, it was short and readable enough earlier

fregante:
Let's take this opportunity to add `engines.npm` >= 7

https://docs.npmjs.com/cli/v7/configuring-npm/package-json#engines

fregante:
https://github.com/fregante/shorten-repo-url/releases/tag/v2.1.0

fregante:
Weird, my commit didn’t change any code nor lint config

yakov116:
> Weird, my commit didn’t change any code nor lint config

Node 15 enables some features

fregante:
I think most of them are not applicable and used to be eslint-ignored before

webpack.config.ts
```diff
@@ -56,10 +56,10 @@ function parseFeatureDetails(readmeContent: string, id: FeatureID): FeatureMeta
 }
 
 function getFeatures(): FeatureID[] {
-	return Array.from(
-		readFileSync(path.join(__dirname, 'source/refined-github.ts'), 'utf-8').matchAll(/^import '\.\/features\/([^.]+)';/gm),
-		match => match[1] as FeatureID
-	).sort().filter(id => !id.startsWith('rgh-'));
+	return [...readFileSync(path.join(__dirname, 'source/refined-github.ts'), 'utf-8').matchAll(/^import '\.\/features\/([^.]+)';/gm)]
```
fregante, line 59: Let’s move the readFileSync to it’s own constant. Match all can stay on this line probably

source/features/tags-on-commits-list.tsx
```diff
@@ -86,7 +86,7 @@ async function getTags(lastCommit: string, after?: string): Promise<CommitTags>
 			}
 		}
 		`);
-	const nodes: TagNode[] = repository.refs.nodes;
+	const {nodes} = repository.refs;
```
fregante, line 89: This loses the TagNode type making everything that depends on it of type any

webpack.config.ts
yakov116: I have to go now. I can work on it tomorrow or it you want you can commit.

source/features/tags-on-commits-list.tsx
fregante: In fact types are failing

webpack.config.ts
fregante: No rush

yakov116:
> I was unable to update `size-plugin` I could not figure out why webpack stopped working.

https://github.com/GoogleChromeLabs/size-plugin/issues/40

fregante:
> > I was unable to update `size-plugin` I could not figure out why webpack stopped working.
> 
> [GoogleChromeLabs/size-plugin#40](https://github.com/GoogleChromeLabs/size-plugin/issues/40)

Too bad that size-plugin isn't fully maintained. It feels like we always have issues with it, and it's not like we need it to do much. Apparently there isn't any alternative and its output is basically perfect, so we just have to suck it up.

source/features/tags-on-commits-list.tsx
```diff
@@ -86,7 +86,7 @@ async function getTags(lastCommit: string, after?: string): Promise<CommitTags>
 			}
 		}
 		`);
-	const nodes: TagNode[] = repository.refs.nodes;
+	const {nodes}: {nodes: TagNode[]} = repository.refs;
```
fregante, line 89: This is one of those lint rules that in this case just ruin the code. Does this work?
```suggestion
	const nodes = repository.refs.nodes as TagNode[];
```

yakov116: Yes

fregante:
🎉

yakov116:
Please merge before something else has to be updated 😆

fregante:
It's all yours

PR #3982, opened 2021-02-15
Keep diff stats in "Jump to" dropdown when header is fixed

### LINKED ISSUE #3925: Keep diff stats on PR file list when header is fixed
There's a bug in the GitHub UI with the Jump to file list on PR pages.

To be clear, this is not a bug from Refined GitHub. It's a UI glitch which GitHub has and I hope Refined GitHub could improve similar to others.

To reproduce:

1. Open a PR Files tab on GitHub with enough changes - e.g. https://github.com/sindresorhus/refined-github/pull/3922/files
2. Open the "Jump to..." files list.
3. Notice the diff stats on the right of each file.
4. Scroll down the whole page until the header is fixed.
5. Notice the diff stats disappear when the header is fixed.

---

Top of the page, header not fixed:
Screenshot 2021-01-29 at 13 43 54

---

Scrolled down, header fixed:
Screenshot 2021-01-29 at 13 44 02

cheap-glitch:
Thanks for contributing! 🍄

1. LINKED ISSUES: Fixes #3925 

2. TEST URLS: https://github.com/sindresorhus/refined-github/pull/3922/files

3. SCREENSHOT:
<table>
<tr>
	<th>Before
	<td>

image
<tr>
	<th>After
	<td>

image
</table>

The original selector that hides the diff stats is `.is-stuck .diffbar .diffstat`, but since I don't know what else it might affect, I chose something more specific to this particular layout.

Also should this be added to the readme?

### DIFF
--- a/source/refined-github.css
+++ b/source/refined-github.css
@@ -131,3 +131,8 @@ pr-branches
 .WorkflowRunLogsScroll {
 	height: auto !important;
 }
+
+/* Fix #3925 */
+.select-menu-item-text .diffstat {
+	display: inline !important;
+}
--- a/source/refined-github.css
+++ b/source/refined-github.css
@@ -132,7 +132,7 @@ pr-branches
 	height: auto !important;
 }
 
-/* Fix #3925 */
+/* Keep diff stats in "Jump to" dropdown when header is fixed (fix https://github.com/sindresorhus/refined-github/issues/3925) */
 .select-menu-item-text .diffstat {
 	display: inline !important;
 }
--- a/source/refined-github.css
+++ b/source/refined-github.css
@@ -132,7 +132,7 @@ pr-branches
 	height: auto !important;
 }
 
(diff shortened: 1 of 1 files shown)


yakov116:
> Also should this be added to the readme?

Don't think so.

But I think the comment should explain what its fixing and leave the link to the issue

source/refined-github.css
```diff
@@ -131,3 +131,8 @@ pr-branches
 .WorkflowRunLogsScroll {
 	height: auto !important;
 }
+
+/* Keep diff stats in "Jump to" dropdown when header is fixed (fix https://github.com/sindresorhus/refined-github/issues/3925) */
```
yakov116, line 135:
```suggestion
// Keep diff stats in "Jump to" dropdown when the header is fixed #3925
```

Maybe?

yakov116 (approved):

PR #3986, opened 2021-02-16
Add `preview-markdown-result` feature

### LINKED ISSUE #3711: Toggle changes in Markdown preview diff
### Feature
To be able to hide/show the diff in Markdown change previews.

For example, when editing a Markdown file on GitHub.com, you can preview your changes:
https://github.com/sindresorhus/refined-github/edit/master/readme.md

### Use-case
I often edit markdown files in the GitHub UI because it's quick and easy, but when the changes are complex, it becomes hard to see the final output because the diff clutters the preview.

In these cases, I would love to be able to just see the final output instead of what I removed/modified.

Screen Shot 2020-11-10 at 2 22 05 PM

cheap-glitch:
Thanks for contributing! 🍄

1. LINKED ISSUES: Closes #3711 

2. TEST URLS: https://github.com/sindresorhus/refined-github/edit/main/readme.md

3. SCREENSHOT:
preview-markdown-result

Tested in Firefox and Chromium.

### DIFF
--- a/source/features/preview-markdown-result.tsx
+++ b/source/features/preview-markdown-result.tsx
@@ -1,0 +1,32 @@
+import './preview-markdown-result.css';
+import React from 'dom-chef';
+import delegate from 'delegate-it';
+import elementReady from 'element-ready';
+import * as pageDetect from 'github-url-detection';
+
+import features from '.';
+
+function togglePreviewResult({delegateTarget: target}: delegate.Event<MouseEvent, HTMLButtonElement>): void {
+	document.body.classList.toggle('rgh-preview-markdown-result');
+	target.classList.add('selected');
+	(target.previousElementSibling ?? target.nextElementSibling)!.classList.remove('selected');
+}
+
+async function init(): Promise<void> {
+	(await elementReady('.file-header'))!.append(
+		<div className="BtnGroup px-3 pr-md-6 px-lg-2 rgh-preview-toggle">
+			<button className="btn btn-sm BtnGroup-item rgh-preview-button selected" type="button">View diff</button>
+			<button className="btn btn-sm BtnGroup-item rgh-preview-button" type="button">View final result</button>
+		</div>
+	);
+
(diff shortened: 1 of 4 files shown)


fregante:
I like that it’s pretty short!

Does this conflict with our other markdown-source feature?

cheap-glitch:
> Does this conflict with our other markdown-source feature?

No, they don't target the same pages.

source/features/preview-markdown-result.tsx
```diff
@@ -0,0 +1,31 @@
+import './preview-markdown-result.css';
+import React from 'dom-chef';
+import delegate from 'delegate-it';
+import elementReady from 'element-ready';
+import * as pageDetect from 'github-url-detection';
+
+import features from '.';
+
+function togglePreviewResult({delegateTarget: target}: delegate.Event<MouseEvent, HTMLButtonElement>): void {
+	document.body.classList.toggle('rgh-preview-markdown-result');
```
fregante, line 10: Everything looks good. I’m just concerned that the state of this might lose sync with the buttons. I suggest adding a `value=true/false` attribute on the buttons so you can use a safer 
```suggestion
	document.body.classList.toggle('rgh-preview-markdown-result', Boolean(target.value));
```

cheap-glitch: I ended up using `'enabled'`/`'disabled'` because `Boolean('false')` is still `true`.

fregante:
🙌

fregante:
One moment too late… what do you think if we rename this to `hide-markdown-diff`? Technically we’re already on a markdown preview

cheap-glitch:
> One moment too late… what do you think if we rename this to `hide-markdown-diff`? Technically we’re already on a markdown preview

Yeah I don't really like the current name. The description could probably be improved too.

fregante:
Done

641210 (commented):
****

PR #3844, opened 2020-12-25
Fix console errors

yakov116:
- Fix console error in `comments-time-machine-links` - Test on any issue/pr fixes #3845
- Fix console error with `linkify-issues` on gists - Test on https://gist.github.com/yakov116
	After the change to `getCurrentRepo()` getRepo can be undefined

- Fix some empty repo errors
https://github.com/yakov116/fluffy-bassoon

PR #3531, opened 2020-09-05
Improve `cleanup-repo-filelist-actions`’s reliability

kidonng:
1. LINKED ISSUES:
   https://github.com/sindresorhus/refined-github/issues/3527#issuecomment-687519926

2. TEST URLS:
   - Repo root: https://github.com/sindresorhus/refined-github
   - Repo tree: https://github.com/sindresorhus/refined-github/tree/master/source
   - File: https://github.com/sindresorhus/refined-github/blob/master/source/options.css

3. SCREENSHOT:
   None

PR #3988, opened 2021-02-17
Avoid `master` noise in `clean-conversation-headers`

### LINKED ISSUE #3987: `clean-conversation-headers` should also consider `master` as a default
Screen Shot

`master` here refers to the default branch. Since this is a common occurrence, we should just assume it was the _old default branch_

Like https://github.com/sindresorhus/refined-github/issues/3910 and maybe https://github.com/sindresorhus/refined-github/commit/aa6659aad95973fb9e32dd2b7cb4c4273dd100f6

Demo URL: https://github.com/parcel-bundler/parcel/pull/78

yakov116:
1. LINKED ISSUES:
    Closes #3987 

2. TEST URLS:
   https://github.com/parcel-bundler/parcel/pull/78
https://github.com/sindresorhus/refined-github/pull/3844

3. SCREENSHOT:
 None

Threw in a drop of lint hope its ok


source/features/clean-conversation-headers.tsx
```diff
@@ -24,23 +25,22 @@ function initPR(): void {
...
+			for (const node of [...byline.childNodes].slice(isSameAuthor ? 0 : 2, isMergedPR() ? 3 : 5)) {
 				node.remove();
 			}
 
 			if (!isSameAuthor) {
 				byline.prepend('by ');
 			}
 
-			if (isDefaultBranch) {
+			if (isDefaultBranch || baseBranch.textContent!.endsWith(':master')) {
```
fregante, line 43: This only works for the link I posted, but it should work for local PRs too. Don’t just drop the semicolon though, because `not-master` shouldn’t match.

yakov116: https://github.com/sindresorhus/refined-github/pull/3844

yakov116:
My gut feeling is that we should only do this for a merged PR. I can see people having dev as the default branch and master as the release branch.

fregante:
> My gut feeling is that we should only do this for a merged PR

Yeah, the current code and the other feature both already have this logic, right?

fregante:
That made me realize. Exactly like for the previous feature, it's not just merged PRs, but also closed PRs.

Example closed PR with `master`: https://github.com/sindresorhus/refined-github/pull/3531

source/features/clean-conversation-headers.tsx
```diff
@@ -24,23 +25,23 @@ function initPR(): void {
 	observe('.gh-header-meta .flex-auto:not(.rgh-clean-conversation-header)', {
 		add(byline) {
 			byline.classList.add('rgh-clean-conversation-header');
-			const isMerged = select.exists('#partial-discussion-header [title="Status: Merged"]');
 			const isSameAuthor = select('.js-discussion > .TimelineItem:first-child .author')?.textContent === select('.author', byline)!.textContent;
 			const baseBranch = select('.commit-ref:not(.head-ref)', byline)!;
 			const isDefaultBranch = (baseBranch.firstElementChild as HTMLAnchorElement).pathname.split('/').length === 3;
+			const prNotOpen = isClosedPR() || isMergedPR();
```
yakov116, line 31: Don't like the var name here. What would be a better name?

fregante: Merged PR are also closed PRs. Just use a single isClosedPR detection instead of 2

source/features/git-checkout-pr.tsx
```diff
@@ -105,7 +105,7 @@ void features.add(__filebasename, {
 	],
 	exclude: [
 		isMergedPR,
-		() => select.exists('#partial-discussion-header [title="Status: Draft"]')
+		isClosedPR
```
fregante, line 108: …

this is different

yakov116: Look at the full diff. I made a mistake in one of the  commits and changed it to draft

source/features/clean-conversation-headers.tsx
yakov116: But I still need a separate is merged. 

So make a isMerged and isClosed should include is merged right?

fregante:
I think it looks good. It might be worth adding the 2 to url-detection, but `isClosed` should just have a single, joint selector instead of calling `isMerged`

source/features/clean-conversation-headers.tsx
fregante: `title` works

fregante (approved):

PR #3989, opened 2021-02-17
Make `quick-repo-deletion` success message more helpful

### LINKED ISSUE #3984: Improve `quick-repo-deletion`
Screen Shot

On the final information bar:

- [ ] Add link to recovery https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/restoring-a-deleted-repository as suggested by @kidonng 
- [ ] Add link to user’s own links so they can delete more, e.g. https://github.com/fregante?tab=repositories&q=&type=fork&language=

Full example:

```
Repository {url} deleted. You can [restore it](…) or see [your other forks](…).
```

yakov116:
1. LINKED ISSUES:
   Closes #3984 

2. TEST URLS:
None

3. SCREENSHOT:
image

### DIFF
--- a/source/features/quick-repo-deletion.tsx
+++ b/source/features/quick-repo-deletion.tsx
@@ -132,7 +132,7 @@ async function init(): Promise<void | false> {
 			<details className="details-reset details-overlay select-menu rgh-quick-repo-deletion">
 				<summary aria-haspopup="menu" role="button">
 					{/* This extra element is needed to keep the button above the <summary>’s lightbox */}
-					<span className="btn btn-sm btn-danger">Delete repo</span>
+					<span className="btn btn-sm btn-danger">Delete fork</span>
 				</summary>
 			</details>
 		</li>
--- a/source/features/quick-repo-deletion.tsx
+++ b/source/features/quick-repo-deletion.tsx
@@ -88,7 +88,7 @@ async function start(buttonContainer: HTMLDetailsElement): Promise<void> {
 			method: 'DELETE',
 			json: false
 		});
-		addNotice(<span>Repository {nameWithOwner} deleted. You can <a href={`${isOrganizationRepo() ? '/organizations/' + owner : ''}/settings/deleted_repositories`}>restore it</a> or see <a href={`/${owner}?tab=repositories&q=&type=fork`}>your other forks</a></span>, {action: false});
+		addNotice(<span>Repository {nameWithOwner} deleted. You might be able to <a href={`${isOrganizationRepo() ? '/organizations/' + owner : ''}/settings/deleted_repositories`}>restore it</a> or see <a href={`/${owner}?tab=repositories&type=fork`}>your other forks</a></span>, {action: false});
 		select('.application-main')!.remove();
 		if (document.hidden) {
 			// Try closing the tab if in the background. Could fail, so we still update the UI above
--- a/source/features/quick-repo-deletion.tsx
+++ b/source/features/quick-repo-deletion.tsx
@@ -81,14 +81,20 @@ async function start(buttonContainer: HTMLDetailsElement): Promise<void> {
 	}
 
 	select('.btn', buttonContainer)!.textContent = 'Deleting repo…';
-	const isOrganizationRepo = (): boolean => document.querySelector<HTMLFormElement>('[data-owner-scoped-search-url]')!.dataset.ownerScopedSearchUrl!.startsWith('/org');
 	try {
 		const {nameWithOwner, owner} = getRepo()!;
 		await api.v3('/repos/' + nameWithOwner, {
 			method: 'DELETE',
 			json: false
 		});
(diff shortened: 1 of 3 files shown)


source/features/quick-repo-deletion.tsx
```diff
@@ -83,12 +83,12 @@ async function start(buttonContainer: HTMLDetailsElement): Promise<void> {
...
 
 	try {
-		const {nameWithOwner} = getRepo()!;
+		const {nameWithOwner, owner} = getRepo()!;
 		await api.v3('/repos/' + nameWithOwner, {
 			method: 'DELETE',
 			json: false
 		});
-		addNotice(`Repository ${nameWithOwner} deleted`, {action: false});
+		addNotice(<span>Repository {nameWithOwner} deleted. You can <a href="https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/restoring-a-deleted-repository">restore it</a> or see <a href={`/${owner}?tab=repositories&q=&type=fork`}>your other forks</a></span>, {action: false});
```
fregante, line 91: Does `owner` also work for organizations? Also maybe we can provide a direct link for recovery if the link is the same for both users and orgs (via `owner` again)

fregante: I checked now that I'm on the computer. Both URLs vary depending on user vs org:

Deleted repos:

```
https://github.com/settings/deleted_repositories
https://github.com/organizations/GhostText/settings/deleted_repositories
```

Fork list:

```
https://github.com/fregante?tab=repositories&type=fork
https://github.com/GhostText?type=fork
```

The issue is that as far as I know it's not possibly to detect whether a repo belongs to an organization, at least this was the case the last time we needed this piece of information.

For the forks list however, the first actually works for both anyway, `tab` is discarded.

So I think this is ok as is, unless you figure out how to detect an org repo without HTTP calls.

fregante:
> 3\. SCREENSHOT:
>     None

🤷‍♂️

source/features/quick-repo-deletion.tsx
yakov116: $('[data-owner-scoped-search-url]')

fregante: Great! It's worth adding a `isOrganizationRepo` to detection module. The fork list link can still stay as is though.

yakov116: I'm vacationing over the weekend, so either I will do it tonight or next week.

yakov116:
Um um um....
"It may take up to an hour for repositories to be displayed here. You can only restore repositories that have no forks or have not been forked"

You cannot restore it (a fork)! 😠

fregante:
We should change our copy to “You might be able to restore it”

source/features/quick-repo-deletion.tsx
```diff
@@ -81,14 +81,14 @@ async function start(buttonContainer: HTMLDetailsElement): Promise<void> {
...
+	const isOrganizationRepo = (): boolean => document.querySelector<HTMLFormElement>('[data-owner-scoped-search-url]')!.dataset.ownerScopedSearchUrl!.startsWith('/org');
 	try {
-		const {nameWithOwner} = getRepo()!;
+		const {nameWithOwner, owner} = getRepo()!;
 		await api.v3('/repos/' + nameWithOwner, {
 			method: 'DELETE',
 			json: false
 		});
-		addNotice(`Repository ${nameWithOwner} deleted`, {action: false});
+		addNotice(<span>Repository {nameWithOwner} deleted. You can <a href={`${isOrganizationRepo() ? '/organizations/' + owner : ''}/settings/deleted_repositories`}>restore it</a> or see <a href={`/${owner}?tab=repositories&q=&type=fork`}>your other forks</a></span>, {action: false});
```
fregante, line 91:
```suggestion
		addNotice(<span>Repository {nameWithOwner} deleted. You might be able to <a href={`${isOrganizationRepo() ? '/organizations/' + owner : ''}/settings/deleted_repositories`}>restore it</a> or see <a href={`/${owner}?tab=repositories&type=fork`}>your other forks</a></span>, {action: false});
```

source/features/quick-repo-deletion.tsx
```diff
@@ -81,14 +81,14 @@ async function start(buttonContainer: HTMLDetailsElement): Promise<void> {
 	}
 
 	select('.btn', buttonContainer)!.textContent = 'Deleting repo…';
-
+	const isOrganizationRepo = (): boolean => document.querySelector<HTMLFormElement>('[data-owner-scoped-search-url]')!.dataset.ownerScopedSearchUrl!.startsWith('/org');
```
fregante, line 84:
```suggestion

```

Just to keep track: the detection package needs to be updated before merging

fregante (changes requested):

source/features/quick-repo-deletion.tsx
fregante: After both changes, it’s good to go

Issue #3647, opened 2020-10-13
`preview-hidden-comments` no longer works for most types


fregante:
The `preview-hidden-comments` feature isn't working probably because GitHub started hiding some reasons why a comment has been hidden:

<img width="591" alt="" src="https://user-images.githubusercontent.com/1402241/95822633-e85d1c80-0cf1-11eb-8ec2-7f05a70f94bd.png">

I'll add and hide comments below as a live example.

Here's the originally-allowed types for reference:

https://github.com/sindresorhus/refined-github/blob/c6c727ca571c91d1ff426e218ccadb075cdb2af6/source/features/preview-hidden-comments.tsx#L9

_Originally reported in https://github.com/sindresorhus/refined-github/pull/3621#issuecomment-707895439_

fregante:
This comment is off-topic

fregante:
This comment is spam

fregante:
This comment is outdated

fregante:
This comment is resolved

fregante:
This comment is abuse

fregante:
This comment is duplicate

PR #3992, opened 2021-02-18
Hide and auto-expand Similar comments in `hide-useless-comments`

### LINKED ISSUE #3830: Expand but hide "1 similar comment" in `hide-useless-comments`
"Similar" comments are useless comments, so they should be hidden by `hide-useless-comments`.

The feature should run on these comments and, if they are indeed useless, the should be expanded ¹ and hidden.

---

¹ the user shouldn't have to click that comment to expand it again if they already clicked "show useless comments"

Test URL: https://github.com/justinfrankel/licecap/issues/97#issuecomment-737063337

cheap-glitch:
1. LINKED ISSUES: Closes #3830 

2. TEST URLS: https://github.com/justinfrankel/licecap/issues/97

Picked up where #3839 left off. This PR:
 - always hides "similar comment(s)" boxes
 - expand every one of them when the link to show useless comments is clicked

This is not exactly what was described in #3830, rather it's following this review comment: https://github.com/sindresorhus/refined-github/pull/3839#discussion_r548323403

The idea is to consider every similar comment as useless by default, and not expand them to actually check their contents, since there can be a lot of them (I remember seeing issues with 200+ similar comments). Thoughts?

### DIFF
--- a/source/features/hide-useless-comments.tsx
+++ b/source/features/hide-useless-comments.tsx
@@ -12,12 +12,28 @@ function unhide(event: delegate.Event<MouseEvent, HTMLButtonElement>): void {
 		comment.hidden = false;
 	}
 
+	// Expand all "similar comments" boxes
+	for (const similarCommentsExpandButton of select.all('.Details-content--closed > :first-child')) {
+		similarCommentsExpandButton.click();
+	}
+
 	select('.rgh-hidden-comment')!.scrollIntoView();
 	event.delegateTarget.parentElement!.remove();
 }
 
+function hideComment(comment: HTMLElement): void {
+	comment.hidden = true;
+	comment.classList.add('rgh-hidden-comment');
+}
+
 function init(): void {
 	let uselessCount = 0;
+
+	for (const similarCommentsBox of select.all('.Details-element')) {
+		hideComment(similarCommentsBox);
+		uselessCount++;
+	}
+
 	for (const commentText of select.all('.comment-body > p:only-child')) {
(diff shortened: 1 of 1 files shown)


fregante:
Yes the logic in your post is correct

fregante (changes requested):
Both new selectors are rather generic and need to be improved

fregante:
Strangely it hides "Same here" but not "same"

After this change it can be merged :shipit:

fregante:
Screen Shot

Strangely it hides "Same here" but not "same"

After this change it can be merged :shipit:

fregante:
Ouch, I think we got a problem:

Screen Shot

See the error in https://github.com/sindresorhus/refined-github/issues/3647

source/features/hide-useless-comments.tsx
```diff
@@ -12,12 +12,28 @@ function unhide(event: delegate.Event<MouseEvent, HTMLButtonElement>): void {
 		comment.hidden = false;
 	}
 
+	// Expand all "similar comments" boxes
+	for (const similarCommentsExpandButton of select.all('.js-discussion .Details-content--closed > span:only-child')) {
```
fregante, line 16: Maybe

```suggestion
 	for (const similarCommentsExpandButton of select.all('.pagination-loader-container .Details-content--closed > span:only-child')) {
```

fregante: We should fix this soon and I'll release an early version

PR #3993, opened 2021-02-18
Detect more comments in `hide-useless-comments`

cheap-glitch:
1. LINKED ISSUES: -

2. TEST URLS: https://github.com/justinfrankel/licecap/issues/97 (bottom of the page)

3. SCREENSHOT: -

I noticed a couple of useless comments that slipped through the cracks, namely "Dito" and "Followig".


fregante (approved):

fregante:
Screen Shot

Strangely it hides "Same here" but not "same"

After this change it can be merged :shipit:

fregante:
Ouch, I think we got a problem:

Screen Shot

See the error in https://github.com/sindresorhus/refined-github/issues/3647

PR #3996, opened 2021-02-18
Preserve `sticky-sidebar` after a new comment

### LINKED ISSUE #3582: Use `ResizeObserver` in `sticky-sidebar`
Instead of relying on order of executing, we can use this new API to ensure that the sidebar’s stickiness is updated when it changes size. You'd have to combine `ResizeObserver` with `selector-observer`

~~I'm automatically closing this to be picked up after #3084 is resolved.~~

The current feature’s code might be a little hard to adapt/understand, so feel free to write it from scratch; This is probably easier and it's probably only 15 lines of code or less.

cheap-glitch:
1. LINKED ISSUES: Closes #3582 

2. TEST URLS:
 * Conversation sidebar
   * [PR]()
   * [Issue](https://github.com/sindresorhus/refined-github/issues/3582)
 * Repo root
   * [Small sidebar](https://github.com/co1ncidence/mountaineer.vim)
   * [Sidebar too tall for viewport](https://github.com/sindresorhus/refined-github) (disable `clean-repo-sidebar` before)

3. SCREENSHOT: -

I've added types for `ResizeObserver` but I don't know if those are "official" (seems like the TS issue referenced in `on-element-removal` was closed but idk if the changes have been released yet).

### DIFF
--- a/source/features/sticky-sidebar.tsx
+++ b/source/features/sticky-sidebar.tsx
@@ -1,23 +1,26 @@
 import './sticky-sidebar.css';
+import select from 'select-dom';
+import debounce from 'debounce-fn';
 import {observe} from 'selector-observer';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
 
+// The two selectors are present on conversation pages so we need to discriminate
+const sidebarSelector = pageDetect.isRepoRoot() ? '.repository-content .flex-column > :last-child [data-pjax]' : '#partial-discussion-sidebar';
+
+function updateStickiness(): void {
+	const sidebar = select(sidebarSelector)!;
+	const margin = pageDetect.isConversation() ? 60 : 0; // 60 matches sticky header's height
+	sidebar.classList.toggle('rgh-sticky-sidebar', sidebar.offsetHeight + margin < window.innerHeight);
+}
+
+const onResize = debounce(updateStickiness, {wait: 100});
+
 function init(): void {
-	const sidebarSelector = [
-		'#partial-discussion-sidebar', // Conversations
-		'.repository-content .flex-column > :last-child [data-pjax]' // `isRepoRoot`
-	].join();
-
-	const resizeObserver = new ResizeObserver(([{target: sidebar}]) => {
-		const margin = pageDetect.isConversation() ? 60 : 0; // 60 matches sticky header's height
-		sidebar.classList.toggle('rgh-sticky-sidebar', (sidebar as HTMLElement).offsetHeight + margin < window.innerHeight);
-	});
-
 	observe(sidebarSelector, {
 		add(sidebar) {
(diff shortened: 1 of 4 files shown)


source/features/sticky-sidebar.tsx
```diff
@@ -1,27 +1,25 @@
...
 
-function updateStickiness(): void {
-	const sidebar = select(sideBarSelector)!;
-	const margin = pageDetect.isConversation() ? 60 : 0; // 60 matches sticky header's height
-	const sidebarHeight = sidebar.offsetHeight + margin;
-	sidebar.classList.toggle('rgh-sticky-sidebar', sidebarHeight < window.innerHeight);
-}
-
-const onResize = debounce(updateStickiness, {wait: 100});
+	const resizeObserver = new ResizeObserver(([{target: sidebar}]) => {
```
fregante, line 13: I’d keep the debounced listener

source/features/sticky-sidebar.tsx
```diff
@@ -35,9 +36,7 @@ void features.add(__filebasename, {
 	additionalListeners: [
 		() => {
 			window.addEventListener('resize', onResize);
```
fregante, line 37: This should also be dropped

fregante (approved):
The rest seems fine, but untested

source/features/sticky-sidebar.tsx
cheap-glitch: I think this is needed, because the window can get resized and become too small to fit the whole sidebar without the dimensions of the sidebar element actually changing.

fregante: Indeed. You might want to move this to init though. I’m not sure of why I put it in additionalListeners like this

yakov116: I am not at a computer, but if I remember correctly it was put by additional listeners since it needs to run on new comments. (The side bar gets refreshed or something like that)
I can be totally wrong.
Feel free to mark this as resolved.

fregante: > I am not at a computer

Yes but [are you a yellow smiley](https://www.youtube.com/watch?v=lGOofzZOyl8)?

fregante: > it was put by additional listeners since it needs to run on new comments

I didn't check but I think I put it there to avoid creating an init function, which was kinda silly. `additionalListener` is useful because get passed the `init` and then the listener calls it, but in this case neither one does that 🤷‍♂️

Either way the new comments are now handled by the observer

fregante:
Testing

fregante:
Ahh, finally the sidebar doesn't lose its stickiness after a comment

fregante:
However there's a size calculation issue, to be solved separately 
gif

PR #4003, opened 2021-02-20
Improve GitHub Sponsors UX

fregante:
Just a silly feel-good feature.

gif

Named `rgh-` because it's safe so it doesn't need to be documented, but it's not limited to RG.

I'll probably also add an animation on [Sponsor] click

## Test

https://github.com/babel

### DIFF
--- a/source/features/rgh-sponsor-button.tsx
+++ b/source/features/rgh-sponsor-button.tsx
@@ -1,4 +1,5 @@
 import select from 'select-dom';
+import delegate from 'delegate-it';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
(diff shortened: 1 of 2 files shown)


fregante:
gif

source/features/rgh-sponsor-button.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+void features.add(__filebasename, {
+	include: [
+		pageDetect.isIssue,
+		pageDetect.isRepoIssueList
+	],
+	init: handleNewIssue
+});
+
+void features.add(__filebasename, {
```
yakov116, line 73: Why the second features.add?

fregante: D’oh. I forgot we merged them 😂

yakov116:
Not by a computer... Does this check if they are sponsoring already?

fregante:
I don’t think that piece of information is available on the page, right? The button appears as “Sponsoring” on profiles but not on repos

641210 (commented):
****

PR #4009, opened 2021-02-22
Meta: Linkify RG’s own features on this repo via `rgh-linkify-features`

fregante:
Closes https://github.com/sindresorhus/refined-github/issues/3980

This feature is in 2 parts:

- linkifies `code` in comments anywhere via `rgh-linkify-features`
- linkifies `code` in conversation titles via `format-conversation-titles`

## Test

- this very PR, both in the title and in the 2 features mentioned above
- `summary > code` in comment: https://github.com/sindresorhus/refined-github/pull/3783

## Screenshots

Screen Shot

Screen Shot

## TODO

- [ ] the title link appears to be lost after updating the title or posting a new comment
- [x] new comments aren't affected
- [x] Depends on #4010

PR #4010, opened 2021-02-22
Meta: Simplify magic variables and avoid possible duplicate injection

fregante:
Extracting some unrelated changes from #4009 

- Keep a simple array of feature names so they're easy to find
- Instead of magic variables, use `features.list` and `features.meta` where available (don't import `features/index.tsx` in background/options though)


source/features/index.tsx
```diff
@@ -194,7 +194,7 @@ function enforceDefaults(
 const add = async (id: FeatureID, ...loaders: FeatureLoader[]): Promise<void> => {
 	/* Feature filtering and running */
 	const options = await globalReady;
-	if (options[`feature:${id}`] === false) {
+	if (id !== __filebasename && !options[`feature:${id as unknown as string}`]) {
```
fregante, line 197: This wasn't necessary before because the `FeatureConfig` type was `string | boolean` due to some TypeScript restrictions. The current code allows for correct types so XO correctly complained about this.

However `index` now needs its own check because `options.['feature:index']` isn't `false`, but still falsy.

yakov116 (approved):

PR #4011, opened 2021-02-22
Fix console error in `sticky-sidebar`

yakov116:
Fixes a console error in `sticky-sidebar`

Test: Go to conversation tab then go to the files tab. The resize observer was never disconnected so it will error out.


fregante:
Is this executed in order if the sidebar is regenerated? I'd expect `remove` to be called before `add`, but I'm not sure. If it's the opposite, it won't work since `.disconnect()` will forget all the observed elements

yakov116:
test comment

yakov116:
Seems to be working for me.

source/features/sticky-sidebar.tsx
```diff
@@ -16,11 +16,12 @@ function updateStickiness(): void {
 }
 
 const onResize = debounce(updateStickiness, {wait: 100});
+const observer = new ResizeObserver(onResize);
 
 function init(): void {
 	observe(sidebarSelector, {
 		add(sidebar) {
```
yakov116, line 23: Why does this not have a "not selector"

fregante: It’s not making any changes. We’re just listening to it. ResizeObserver already deduplicates identical .observe calls, I suppose

yakov116: ok

fregante:
I’m a bit lost on this feature at the moment but I trust you tested it so it can be merged

fregante:
1. https://github.com/sindresorhus/refined-github
2. Visit Issues
3. Resize window

```
refined-github.js:10773 Uncaught TypeError: Cannot read property 'classList' of undefined
    at updateStickiness (refined-github.js:10773)
    at later (refined-github.js:197)
```

PR #4013, opened 2021-02-23
Fix `hide-useless-comments` selectors

yakov116:
1. LINKED ISSUES:
   #3992 

2. TEST URLS:
   https://github.com/justinfrankel/licecap/issues/97

3. SCREENSHOT:
   Before
image
after
image

### DIFF
--- a/source/features/hide-useless-comments.tsx
+++ b/source/features/hide-useless-comments.tsx
@@ -13,7 +13,7 @@ function unhide(event: delegate.Event<MouseEvent, HTMLButtonElement>): void {
 	}
 
 	// Expand all "similar comments" boxes
-	for (const similarCommentsExpandButton of select.all('.js-discussion .Details-content--closed > span:only-child')) {
+	for (const similarCommentsExpandButton of select.all('.pagination-loader-container .Details-content--closed > span:only-child')) {
 		similarCommentsExpandButton.click();
 	}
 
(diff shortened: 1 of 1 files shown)


fregante:
As long as it was tested on the licecap thread for the similar comment it’s good to be merged

yakov116:
@cheap-glitch can you look this over? Why do we use different selectors for hide and unhide?

source/features/hide-useless-comments.tsx
```diff
@@ -13,7 +13,7 @@ function unhide(event: delegate.Event<MouseEvent, HTMLButtonElement>): void {
 	}
 
 	// Expand all "similar comments" boxes
-	for (const similarCommentsExpandButton of select.all('.js-discussion .Details-content--closed > span:only-child')) {
+	for (const similarCommentsExpandButton of select.all('.pagination-loader-container .Details-content--closed > span:only-child')) {
```
cheap-glitch, line 16: @yakov116 we can indeed simplify this selector to `'.js-discussion .Details-element:not([data-body-version]) > summary'`

yakov116: Thought so but it was waay past my bed time and I wanted to release it. (We don't want broken features)

I put it on my list for my next lint

fregante:
This broke https://github.com/sindresorhus/refined-github/pull/3992

In the linked Licecap thread you can clearly see that the similar comment is hidden, but not expanded later:

Screen Shot

fregante:
Also it might be a good time to also fix that little corner. The rule should be be:

```css
.rgh our class .timeline-comment--caret:before {
	border-right-color: var(--github-red);
}
```

cheap-glitch:
> In the linked Licecap thread you can clearly see that the similar comment is hidden, but not expanded later

Weird, it's working for me (Firefox + `21.2.23`).

PR #4017, opened 2021-02-24
Fix `first-published-tag-for-merged-pr` selector

yakov116:
1. LINKED ISSUES:
   None

2. TEST URLS:
   https://github.com/sindresorhus/refined-github/pull/4013

3. SCREENSHOT:
   None

The feature broke because the selector changed its bound to break again.

I figured its only one API request and we then also have it on all pages. It also allows us not to wait for DOM ready.

### DIFF
--- a/source/features/first-published-tag-for-merged-pr.tsx
+++ b/source/features/first-published-tag-for-merged-pr.tsx
@@ -21,7 +21,7 @@ const getFirstTag = cache.function(async (commit: string): Promise<string | unde
 });
 
 async function init(): Promise<void> {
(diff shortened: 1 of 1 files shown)


fregante:
Now the feature makes 2 HTTP requests on load. Either they should be combined or we should just update the selector.

fregante:
> the selector changed its bound to break again

Such is life

yakov116:
> Either they should be combined or we should just update the selector

I am not sure what you mean. There is no way I can combine it. Also I want the feature to run on all or pages

fregante:
It would have been nice to have it on every page, but not at the cost of 2 HTTP requests. We already have a lot

fregante:
Closing this because it's being handled everywhere in #4021

Issue #3543, opened 2020-09-11
Welcome to Refined GitHub! ✨


fregante:
<div align="center"><img src="https://raw.githubusercontent.com/sindresorhus/refined-github/master/source/icon.png" width="128"> </div>

## Enabling API-based features

Make sure you create and add a **Personal Access Token** in the extension’s options to ensure every feature works correctly.

## Updates

You can follow this repository to know what each new release includes.

## Debugging

If any feature isn't working, [open the web console](https://documentation.concrete5.org/tutorials/how-open-browser-console-view-errors) and look for Refined GitHub-related errors.

## Contributing

Suggestions and pull requests are highly encouraged! Have a look at the [open issues](https://github.com/sindresorhus/refined-github/issues?q=is%3Aissue+is%3Aopen+label%3A%22help+wanted%22+sort%3Areactions-%2B1-desc), especially [the easy ones](https://github.com/sindresorhus/refined-github/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22+sort%3Areactions-%2B1-desc). Check out [contributing.md](https://github.com/sindresorhus/refined-github/blob/master/contributing.md) for more info!

fregante:
Thanks for visiting! This page has been replaced in https://github.com/refined-github/refined-github/pull/7854

PR #4019, opened 2021-02-24
Meta: Cleanup the options page

### LINKED ISSUE #4018: Add feature count in options page
Something really easy, like:

Screen Shot

Probably just:

```js
select('whatever features title').append(' (', features.list.length, ')')
```

fregante:
- [x] Closes #4018

The welcome page points to https://github.com/sindresorhus/refined-github/issues/3543. I never liked having text about GitHub Enterprise in there.

Screen Shot 9


fregante:
Count added: 

Screen Shot 11

distribution/options.html
```diff
@@ -3,6 +3,13 @@
 <title>Refined GitHub options</title>
 <link rel="stylesheet" href="build/options.css">
 <form id="options-form" class="detail-view-container">
+	<p>
+		<strong>Hello!</strong><br>
```
fregante, line 7: Not convinced about this, but also it felt weird as the only section without a title

641210 (commented):
****

PR #4020, opened 2021-02-24
Meta: Move new features to the top on the Options page

### LINKED ISSUE #4012: Move NEW features to the top of the extension options list
Push the NEW ones up in the list, so we'd have:

- disabled
- new
- the rest of them

_Originally posted by @fregante in https://github.com/sindresorhus/refined-github/issues/4004#issuecomment-782935195_

cheap-glitch:
1. LINKED ISSUES: Closes #4012 

2. TEST URLS: the options page

3. SCREENSHOT:
screenshot-1614159444

I chose a method with very little code, but I'm not sure about `await`ing `highlightNewFeatures()`. It causes a small but noticeable content jump when opening the options page.

### DIFF
--- a/source/options.tsx
+++ b/source/options.tsx
@@ -81,8 +81,14 @@ async function validateToken(): Promise<void> {
 	}
 }
 
-function moveDisabledFeaturesToTop(): void {
+function moveNewAndDisabledFeaturesToTop(): void {
 	const container = select('.js-features')!;
+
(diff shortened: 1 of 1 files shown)


fregante:
I think it's worth moving NEW features to the top since:

- they are likely to be few features
- that's what people are likely to be interested about when opening the options _again_

PR #4023, opened 2021-02-26
Add `collapse-markdown-sections` feature

### LINKED ISSUE #3224: Add button to collapse readme/markdown sections
Readmes can become link (e.g. https://github.com/sindresorhus/refined-github) so it would be useful sometimes to collapse/expand all sections to sort of generate a temporary "Table of Contents"

<table><th>Open<th>Collapsed

<tr><td width="50%">

open

<td valign=top>

collapsed

</table>

This is especially useful when looking at API documentation like in https://github.com/fregante/text-field-edit

<table><th>Open<th>Collapsed

<tr><td width="50%">

l

<td valign=top>

s

</table>

```css
/* demo CSS */
.markdown-body h2 ~ p,
.markdown-body h2 ~ ul,
.markdown-body h2 ~ div,
.markdown-body h2 ~ table,
.markdown-body h2 ~ details {
    display: none;
}
```

I'm thinking that this could be implemented as a "click to toggle" on each heading, plus the usual "alt-click to toggle all". Basically no UI.

Alternatively this could be coupled with a "toggle all" button next to `edit-readme`

cheap-glitch:
1. LINKED ISSUES: Closes #3224 

2. TEST URLS:
 * Repo roots
   * https://github.com/fregante/GhostText
   * https://github.com/microsoft/TypeScript
 * Folders with a readme inside:
   * https://github.com/babel/babel/tree/main/packages/babel-cli
 * Markdown files
   * https://github.com/fregante/github-url-detection/blob/master/readme.md
   * https://github.com/ajv-validator/ajv/blob/master/docs/json-schema.md
   * https://github.com/yargs/yargs/blob/master/docs/api.md

3. SCREENSHOT:

collapse-markdown-sections


source/features/collapse-markdown-sections.css
```diff
@@ -0,0 +1,4 @@
+/* Remove the default triangle marker and the padding */
+.rgh-markdown-section > summary {
+	display: block !important;
+}
```
fregante, line 4: Chrome, doesn't work:

Screen Shot 5

This is why I suggested _not_ using details/summary. Since we need the listener for the alt-click, we might as well use the listener to toggle content.

This has also the advantage of not having to touch the DOM at all until the user needs the feature.

fregante (changes requested):

fregante:
Also this only seems to toggle h2s, but I think it should toggle all headers, refer to the "API documentation" example: https://github.com/sindresorhus/refined-github/issues/3224

cheap-glitch:
Rewritten with the following behavior:
 - clicking on a header of any level (except `h1`) toggles the corresponding "section"
 - re-opening a section opens every "sub-section", even if they were hidden when the larger section was toggled off
 - alt-clicking a header toggles all the header of the same level _in the current "super-section" only_

collapse-markdown-sections

fregante (commented):
The logic looks good! However it still differs from the original suggestion, which _never_ hid any headings, leading to a "All collapsed" state that looks like an outline/toc, rather than having to manually still open each section.

I'm not saying it's better yet, but maybe it's nicer to see the whole outline instead. We should test this on long readmes where this actually matters.

cheap-glitch:
> The logic looks good! However it still differs from the original suggestion, which _never_ hid any headings, leading to a "All collapsed" state that looks like an outline/toc, rather than having to manually still open each section.

Yeah that's a much better idea. The outline becomes immediately visible and navigation is a lot easier.

source/features/collapse-markdown-sections.ts
```diff
@@ -0,0 +1,68 @@
...
+				break;
+			}
+
+			if (isSameLevelHeading(sectionHeading, element)) {
+				sectionHeading = element;
+			}
+		}
+	}
+
+	for (let element = sectionHeading; element; element = element.nextElementSibling as HTMLElement) {
```
fregante, line 40:
```suggestion
	let element = sectionHeading;
	while (element) {
```

And then `element = element.nextElementSibling as HTMLElement` at the end.

Same goes for the previous loop

source/features/collapse-markdown-sections.ts
```diff
@@ -0,0 +1,68 @@
...
+	if (event.altKey) {
+		// Set `sectionHeading` as the first same-level header of the larger section
+		for (let element = sectionHeading; element; element = element.previousElementSibling as HTMLElement) {
+			if (!isHeading(element)) {
+				continue;
+			}
+
+			if (isHigherLevelHeading(sectionHeading, element)) {
+				break;
+			}
```
fregante, line 32: `isSameLevelHeading` already has exact comparison, so I think you can drop this

fregante: But really, "alt-clicking" a heading is equivalent to just clicking its parent, so perhaps it shouldn't be a thing.

I'd rather add a "View outline" button next to the "Edit Readme" so that it toggles the whole tree. However I'm not sure of where to add this button on `isSingleFile`

readme.md
```diff
@@ -315,6 +315,7 @@ Thanks for contributing! 🦋🙌
 - [](# "improve-shortcut-help") [Shows all of Refined GitHub’s new keyboard shortcuts in the help modal (<kbd>?</kbd> hotkey).](https://user-images.githubusercontent.com/29176678/36999174-9f07d33e-20bf-11e8-83e3-b3a9908a4b5f.png)
 - [](# "stop-redirecting-in-notification-bar") [Stops redirecting to notification inbox from notification bar actions while holding <kbd>Alt</kbd>.](https://user-images.githubusercontent.com/202916/80318782-c38cef80-880c-11ea-9226-72c585f42a51.png)
 - [](# "hide-zero-packages") [Hides the `Packages` tab in user profiles if it’s empty.](https://user-images.githubusercontent.com/46634000/101151879-9ade9b00-3622-11eb-9755-50ed54099e02.png)
+- [](# "collapse-markdown-sections") [Collapses sections of Markdown files when their corresponding header is clicked (and collapses every sibling sections when `Alt`-clicking)](https://user-images.githubusercontent.com/46634000/109290277-b0312300-7827-11eb-8816-81c0d963057e.gif)
```
yakov116, line 318: @fregante should this go under Repositories? or Code (just like `html-preview-link`)?

source/features/collapse-markdown-sections.ts
cheap-glitch: > However I'm not sure of where to add this button on isSingleFile

How about next to the "Edit" button too?
image

readme.md
fregante: I think both should go under the `file` section, there's also `hide-markdown-diff` in there

source/features/collapse-markdown-sections.ts
fregante: > clicking on a top-level heading (`h1`)

Some documents have multiple h1s where this wouldn't work as well but:

- eh
- you could map `click on h1` to `toggle all` specifically, rather than following the regular logic

source/features/collapse-markdown-sections.ts
```diff
@@ -0,0 +1,39 @@
...
+		} else {
+			element.hidden = !sectionState;
+		}
+
+		element = element.nextElementSibling as HTMLElement;
+	}
+}
+
+function init(): void {
+	delegate(document, '.markdown-body > ' + allHeadingsSelector, 'click', toggleSection);
```
fregante, line 30: Oops! This is merged as `.markdown-body > h1, h2, h3, h4, h5, h6`. It could be changed to

```suggestion
	delegate(document, `.markdown-body > :is(${allHeadingsSelector})`, 'click', toggleSection);
```

Which makes me think: `getScopedSelector` should be dropped and replaced by `:is()`

source/features/collapse-markdown-sections.ts
```diff
@@ -0,0 +1,39 @@
...
+function toggleSection(event: delegate.Event<MouseEvent, HTMLElement>): void {
+	const sectionHeading = (event.target as HTMLElement).closest<HTMLElement>(allHeadingsSelector)!;
+	const sectionState = sectionHeading.classList.contains('rgh-markdown-section-collapsed');
+	sectionHeading.classList.toggle('rgh-markdown-section-collapsed');
+
+	let element = sectionHeading.nextElementSibling as HTMLElement;
+	while (element) {
+		if (element.matches(allHeadingsSelector)) {
+			if (sectionHeading.tagName.localeCompare(element.tagName) >= 0) {
+				break;
```
fregante, line 17: Does this stop the loop? Let's use `return`

source/features/collapse-markdown-sections.ts
```diff
@@ -0,0 +1,39 @@
+import delegate from 'delegate-it';
+import * as pageDetect from 'github-url-detection';
+
+import features from '.';
+
+const allHeadingsSelector = 'h1, h2, h3, h4, h5, h6';
+
+function toggleSection(event: delegate.Event<MouseEvent, HTMLElement>): void {
+	const sectionHeading = (event.target as HTMLElement).closest<HTMLElement>(allHeadingsSelector)!;
```
fregante, line 9:
```suggestion
	const sectionHeading = event.delegateTarget;
```

fregante (approved):
I think other than this it looks fine!

readme.md
fregante: The gif also needs to be updated to show clicking on the repo title to toggle all, instead of an invisible alt-click

source/features/collapse-markdown-sections.ts
cheap-glitch: > you could map click on h1 to toggle all specifically, rather than following the regular logic

:+1:

source/features/collapse-markdown-sections.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+function init(): void {
+	delegate(document, '.markdown-body > h1', 'click', toggleAllTopSections);
+	delegate(document, '.markdown-body > :is(h2, h3, h4, h5, h6)', 'click', toggleSingleSection);
+}
+
+void features.add(__filebasename, {
+	include: [
+		pageDetect.isRepoRoot,
```
fregante, line 44: Not just the root, but any folder that has a readme. Testable in any monorepo like Babel

source/features/collapse-markdown-sections.ts
```diff
@@ -0,0 +1,48 @@
+import select from 'select-dom';
+import delegate from 'delegate-it';
+import * as pageDetect from 'github-url-detection';
+
+import features from '.';
+
+function toggleSection(sectionHeading: HTMLElement): void {
+	const sectionState = sectionHeading.classList.contains('rgh-markdown-section-collapsed');
+	sectionHeading.classList.toggle('rgh-markdown-section-collapsed');
```
fregante, line 9: I think you can merge these 2 lines because toggle also returns a value

source/features/collapse-markdown-sections.ts
```diff
@@ -0,0 +1,48 @@
...
+
+	let element = sectionHeading.nextElementSibling as HTMLElement;
+	while (element) {
+		if (element.matches('h1, h2, h3, h4, h5, h6')) {
```
fregante, line 13: I’m guessing that regex is faster than selector matching
```suggestion
		if (/^h\d$/.test(element.tagName)) {
```

source/features/collapse-markdown-sections.ts
```diff
@@ -0,0 +1,48 @@
...
+			if (sectionHeading.tagName.localeCompare(element.tagName) >= 0) {
```
fregante, line 14:
```suggestion
			if (sectionHeading.tagName[1] >= element.tagName[1]) {
```

I’ve arrived at micro optimizations 😂

source/features/collapse-markdown-sections.ts
```diff
@@ -0,0 +1,46 @@
+import select from 'select-dom';
+import delegate from 'delegate-it';
+import * as pageDetect from 'github-url-detection';
+
+import features from '.';
+
+function toggleSection(sectionHeading: HTMLElement): void {
+	const sectionState = sectionHeading.classList.toggle('rgh-markdown-section-collapsed');
+	let element = sectionHeading.nextElementSibling as HTMLElement;
```
fregante, line 9: If we use this, we don't need `toggleAllTopSections` and the secondary delegate

```suggestion
	if (sectionHeading.tagName === 'H1') {
		  for (const element of sectionHeading.parentElement!.children) {
	 		  if (/^H\d$/.test(element.tagName)) {
	 			  element.classList.toggle('rgh-markdown-section-collapsed', sectionState);
	 		  } else {
	 			  element.hidden = sectionState;
	 		  }
	 	  }
	 	  return;
	}

	let element = sectionHeading.nextElementSibling as HTMLElement;
```

source/features/collapse-markdown-sections.ts
```diff
@@ -0,0 +1,46 @@
+import select from 'select-dom';
+import delegate from 'delegate-it';
+import * as pageDetect from 'github-url-detection';
+
+import features from '.';
+
+function toggleSection(sectionHeading: HTMLElement): void {
+	const sectionState = sectionHeading.classList.toggle('rgh-markdown-section-collapsed');
```
fregante, line 8:
```suggestion
	const isSectionHidden = sectionHeading.classList.toggle('rgh-markdown-section-collapsed');
```

source/features/collapse-markdown-sections.ts
fregante: Better yet, if you set `let element` to pick up the `.firstElementChild` you could use the existing loop, except that you'd also need to skip the `<=` comparison there.

cheap-glitch: > if change the element you could use the existing loop

Yeah I though about that but I wasn't sure making the code messier was warranted by the performance gains (since there's _usually_ just one or two `h1` per Markdown document). But it does makes the feature quite shorter.

fregante (approved):
I haven’t tested again but

fregante:
We got a problem: The section is also collapsed when text is selected

gif

Maybe: https://stackoverflow.com/a/61364802

cheap-glitch:
> We got a problem: The section is also collapsed when text is selected
> 
> Maybe: stackoverflow.com/a/61364802

This mitigates the issue, but it doesn't work when double-clicking to select a word. We could listen to `dblick`, check that there's a selection and if yes re-open the section right after. Worth it?

yakov116:
IMHO return on double click

fregante:
I haven’t tested it, but I assume that on the second click the section reopens. If that’s the case, it’s fine as is. 

The only ways to avoid this are:

- delaying the action until we’re sure there’s no second click (plz no)
- moving this event out of the text altogether, for example on a button on the side

The problem with the latter is that there might be images on the right side of the headers and on small displays there’s no room on the sides.

kidonng:
Too bad didn't see this earlier, I think it makes sense to enable this for wiki as well.

Issue #1757, opened 2019-02-07
New “Projects” tab messes up profile if “gists” tab enabled


yeskunall:
GitHub added a new “Projects” tab. AFAICT, this wasn’t a thing couple of hours ago 🤨

Here’s what it looks like right now:

![image](https://user-images.githubusercontent.com/14703164/52444709-2191c200-2af7-11e9-8992-7664de9d6e9f.png)

For now, my workaround is to simply disable the “gists” tab by adding `profile-gists-link` to my list of features to disable.

![image](https://user-images.githubusercontent.com/14703164/52444939-bb596f00-2af7-11e9-98cd-4eda34c52a96.png)

That fixes it, but I’m sure a lot of people appreciate quick access to their gists, so I’m opening an issue instead. 😅

bfred-it:
The line above isn’t broken. What can be done is hiding the project tab in the existing feature that hides it from the repos

jerone:
A little CSS fixes it:
```css
.UnderlineNav-item {
    margin-right: 8px;
}
```

Then there is enough room for 17 numbers in the counters...

![image](https://user-images.githubusercontent.com/55841/52487688-72202280-2bbe-11e9-8d86-9e7f080e4ba6.png)

bfred-it:
@jerone that's good, we can also limit that to when there are 7+ items, with a quantity query: https://alistapart.com/article/quantity-queries-for-css

PR #4037, opened 2021-02-28
Update dependencies and lint

yakov116:
Test `follow-file-renames` https://github.com/yakov116/refined-github/commits/f23b687b3b89aa95a76193722cdfeff740646670?after=f23b687b3b89aa95a76193722cdfeff740646670+34&path%5B%5D=source&path%5B%5D=features&path%5B%5D=release-download-count.tsx

TODO:
- [x] See why [`2c6b447` (#4037)](https://github.com/sindresorhus/refined-github/pull/4037/commits/2c6b447dc177d26867d24a1f1d63b4d96ad4493d) is needed

@sindresorhus I tried updating `sindresorhus/tsconfig` but `noPropertyAccessFromIndexSignature` conflicts with `@typescript-eslint/dot-notation`

@sindresorhus @fregante should I update the codebase to use `:has`?


sindresorhus:
> @sindresorhus I tried updating sindresorhus/tsconfig but noPropertyAccessFromIndexSignature conflicts with @typescript-eslint/dot-notation

Known problem. Just wait with upgrading that for now.

fregante:
Actually :has doesn’t exist, but it’s :is 

I used :is before and the result isn’t always better, so use it only if the :is() function doesn’t contain long, awkward selectors that force you to wrap (which defeats the purpose)

yakov116:
What about the `forEach`? I think where ever we are using it, it makes sense to keep it.

Thoughts?

fregante:
There’s only one async callback that can’t be changed in on-new-comments. The other async callback needs you to extract that function and call it with `void linkify(button, url)` I think

source/features/follow-file-renames.tsx
```diff
@@ -27,7 +27,7 @@ function init(): false | void {
 		return false;
 	}
 
-	disabledPagination.forEach(async button => {
+	for (const button of disabledPagination) {
```
yakov116, line 30: Am I correct the using `forEach` takes quicker here since it does not respect the await?

source/features/tags-on-commits-list.tsx
```diff
@@ -35,6 +35,7 @@ function mergeTags(oldTags: CommitTags, newTags: CommitTags): CommitTags {
 	const result: CommitTags = {...oldTags};
 	for (const commit in newTags) {
 		if (result[commit]) {
+			// eslint-disable-next-line unicorn/prefer-spread
 			result[commit] = [...new Set(result[commit].concat(newTags[commit]))];
```
yakov116, line 39: Can't figure this one out

source/features/embed-gist-inline.tsx
```diff
@@ -51,9 +51,11 @@ async function embedGist(link: HTMLAnchorElement): Promise<void> {
 }
 
 function init(): void {
-	select.all('.js-comment-body p a:only-child')
-		.filter(item => isGist(item) && isOnlyChild(item))
-		.forEach(embedGist);
+	const gistLinks = select.all('.js-comment-body p a:only-child')
+		.filter(item => isGist(item) && isOnlyChild(item));
```
fregante, line 55: The filter should just be an if inside the loop

source/features/follow-file-renames.tsx
fregante:
```suggestion
	for (const button of disabledPagination) {
	    void linkify(button, url);
	}
```

As I had suggested

source/features/tags-on-commits-list.tsx
fregante: I think this 
```suggestion
			// eslint-disable-next-line unicorn/prefer-spread
			result[commit] = [...new Set([...result[commit], ...newTags[commit]])];
```

source/features/follow-file-renames.tsx
yakov116: Ahhh missed that.

source/features/tags-on-commits-list.tsx
yakov116: I promise I tried that!! Thanks

fregante: If Sindre has a `concat-unique` package I’d use that. The code isn’t very readable here

source/features/highest-rated-comment.tsx
```diff
@@ -68,7 +68,7 @@ function linkBestComment(bestComment: HTMLElement): void {
 	}
 
 	const text = select('.comment-body', bestComment)!.textContent!.slice(0, 100);
-	const {hash} = select('a.js-timestamp', bestComment)!;
+	const {hash} = select<HTMLAnchorElement>('a.js-timestamp', bestComment)!;
```
fregante, line 71: ?

This should be automatic

yakov116: [`2c6b447` (#4037)](https://github.com/sindresorhus/refined-github/pull/4037/commits/2c6b447dc177d26867d24a1f1d63b4d96ad4493d)

I know it was complaining... Let me try to revert

source/features/follow-file-renames.tsx
```diff
@@ -18,49 +18,53 @@ async function findRename(lastCommitOnPage: string): Promise<File[]> {
...
+				);
+			}
+
+			return;
+		}
+	}
+}
+
+function init(): void | false {
+	const disabledPagination = select.all<HTMLButtonElement>('.paginate-container [disabled], .paginate-container .disabled');
```
fregante, line 57: Might be worth checking if .disabled is still needed. If this selector is 1y old, we should simplify it and also use `button` to avoid having to use HTMLButtonElement

yakov116:
@fregante how `:is()` this looking? (pun intended). Should I continue?

source/features/align-issue-labels.css
```diff
@@ -8,13 +8,11 @@
...
 
-.rgh-align-issue-labels .js-issue-row .h4, /* Title */
-.rgh-align-issue-labels .js-issue-row .d-inline-block.mr-1 { /* Build status */
+.rgh-align-issue-labels .js-issue-row :is(.h4 /* Title */, .d-inline-block.mr-1 /* Build status */) {
 	padding-right: 4px; /* Space before the build status */
 }
 
-.rgh-align-issue-labels .js-issue-row .mt-1.text-small.text-gray, /* GHE */
-.rgh-align-issue-labels .js-issue-row .mt-1.text-small.color-text-secondary { /* Issue details line */
+.rgh-align-issue-labels .js-issue-row :is(.mt-1.text-small.text-gray, /* GHE */, .mt-1.text-small.color-text-secondary /* Issue details line */) {
```
fregante, line 15: Yeah this is what I was talking about, the selector is too long and awkward on one line. If you need to wrap it, you might as well not use :is at all

fregante: Plus the GHE one will be dropped at one point so it’s better to just leave it on its own like so the diff will be easy.

source/features/clean-rich-text-editor.css
```diff
@@ -1,8 +1,4 @@
 /* Hide unnecessary comment toolbar items */
-.rgh-clean-rich-text-editor md-mention,
-.rgh-clean-rich-text-editor md-ref,
-.rgh-clean-rich-text-editor md-header,
-.rgh-clean-rich-text-editor md-bold,
-.rgh-clean-rich-text-editor md-italic {
+.rgh-clean-rich-text-editor :is(md-mention, md-ref, md-header, md-bold, md-italic) {
```
fregante, line 2: Here it’s good

source/features/align-issue-labels.css
yakov116: Thought so. 

But `clean-rich-text-editor` made me smile 🎉

source/features/highest-rated-comment.tsx
yakov116: Check the last commit please

fregante: Can you find which dependency update caused this? It worked until now and select-dom hasn’t been updated (its dependency might have, though)

yakov116: I will do it later added a todo to main post

source/features/pr-approvals-count.css
```diff
@@ -1,19 +1,16 @@
 /* Show approvals count in PR list */
-.js-issue-row .text-gray [aria-label*='approval'], /* GHE */
-.js-issue-row .color-text-secondary [aria-label*='approval'] {
+.js-issue-row :is(.text-gray, .color-text-secondary) [aria-label*='approval'] {
```
fregante, line 2: No GHE merging

yakov116: its the .text-gray

source/features/profile-gists-link.css
```diff
@@ -1,5 +1,4 @@
 /* Fix wrapping text on tabs when feature is enabled #1757 */
-.UnderlineNav-item:nth-last-child(n+7),
-.UnderlineNav-item:nth-last-child(n+7) ~ * {
+.UnderlineNav-item:is(:nth-last-child(n+7), :nth-last-child(n+7) ~ *) {
```
fregante, line 2: @sindresorhus thoyghts on the readability of :is() selectors? 

Some are good, some aren’t. Probably it just depends on how simple the selectors inside :is() are. In this case I’d give a 👎

source/features/scrollable-code-and-blockquote.css
```diff
@@ -10,9 +9,6 @@
 	padding-bottom: 0.2em;
 }
 
-.comment-body details blockquote,
-.comment-body details pre,
-.comment-body blockquote blockquote,
-.comment-body blockquote pre {
+.comment-body :is(details blockquote, details pre, blockquote blockquote, blockquote pre) {
```
fregante, line 12: You need 2 is() here
```suggestion
.comment-body :is(details, blockquote) :is(blockquote, pre) {
```

source/features/pr-approvals-count.css
fregante: I explained this in a previous review: GHE-specific selectors are meant to be removed and, like the “repository refresh” changes, it’s easier to keep them on their own instead of intermixing the code. When we finally drop the selectors, we just delete the whole line.

source/features/follow-file-renames.tsx
```diff
@@ -18,49 +18,53 @@ async function findRename(lastCommitOnPage: string): Promise<File[]> {
...
+				);
+			}
+
+			return;
+		}
+	}
+}
+
+function init(): void | false {
+	const disabledPagination = select.all<HTMLButtonElement>('.paginate-container [disabled]');
```
fregante, line 57: If it works
```suggestion
	const disabledPagination = select.all('.paginate-container button[disabled]');
```

source/features/profile-gists-link.css
fregante, line 2: For this selector specifically, it might be worth checking if #1757 still applies after 2 years

fregante (commented):
Looks good so far except the open reviews

source/features/tags-on-commits-list.tsx
fregante: Of course he does: https://github.com/sindresorhus/array-union

yakov116:
@fregante are you ok if I force-push? I want to keep the `css` and the `js` files separate.

Also want to debug  `See why 2c6b447 (#4037) is needed`

fregante:
You mean you want to open a separate PR?

If you want to push to have "a clean history" here it doesn't really help since there are 50 commits. That's only useful if there are a few commits that can be followed. Nobody will check dozens of commits one by one

yakov116:
> You mean you want to open a separate PR?

Much better idea

source/features/profile-gists-link.css
yakov116: I dont have a way to check. I only have 4 tabs

fregante: It doesn't apply only to your own profile. To test it you just have to disable the "hide project tab" feature. My guess is that they added a scrollbar on that tab bar so it no longer applies.

yakov116:
> * See why [`2c6b447` (#4037)](https://github.com/sindresorhus/refined-github/pull/4037/commits/2c6b447dc177d26867d24a1f1d63b4d96ad4493d) is needed

its typescript.

source/features/highest-rated-comment.tsx
fregante, line 71: @g-plane do you know what's happening here? Suddenly `select-dom` can't parse an `'a'` selector into `HTMLAnchorElement`;  It doesn't even return `HTMLElement`, but just an `Element`. `document.querySelector('a')` works correctly.

fregante: Screen Shot

fregante: Must be a TypeScript bug. In the same file, on a different line, it works correctly:

Screen Shot 10

fregante: Apparently TypeScript just does not like that `const {hash}`, because if I change it, it works correctly:

Screen Shot 11

Sorry for bothering you, it's a TypeScript bug.

fregante (approved):

PR #4040, opened 2021-02-28
Clean fork events in `clean-dashboard`

cheap-glitch:
1. LINKED ISSUES: -

2. TEST URLS: the homepage dashboard

3. SCREENSHOT:

<table>
<tr>
	<th> Before
	<td>

Screenshot_2021-02-28 Build software better, together
<tr>
	<th> After
	<td>

Screenshot_2021-02-28 Build software better, together(1)
</table>


fregante:
Hmmm I think they’re meant to be completely hidden, not just cleaned. There’s a specific feature for that

cheap-glitch:
True, this is only useful if `hide-useless-newsfeed-events` is disabled.

fregante:
Yeah we don’t merge features that require other features to be disabled in order to be tested

fregante:
@yakov116 this selector is a good candidate for :is, even if it requires wrapping, because the :is is in the middle of each selector

PR #4041, opened 2021-03-01
Meta: use CSS `:is()`

yakov116:
Cherry-picked out of #4037

I did not see any JS files worth touching now. (maybe after we drop refresh)


source/features/hide-obvious-tooltips.css
```diff
@@ -1,15 +1,12 @@
...
-.notification-indicator::before,
-.notification-indicator::after {
+:is(
+[aria-label^='View the whole file'],
+[aria-label^='Change this file'],
+[aria-label='Delete this file'],
+[aria-label='Edit this file'],
+[aria-label='Edit comment'],
+[aria-label='Edit comment'],
+.notification-indicator
```
fregante, line 9: Indentation?

fregante (approved):

source/features/hide-obvious-tooltips.css
yakov116: Why I put it on a new line? or why did I not add a tab?

yakov116:
Should we wait until after tomorrow's release?

source/features/hide-obvious-tooltips.css
fregante:
```suggestion
	  [aria-label^='View the whole file'],
	  [aria-label^='Change this file'],
	  [aria-label='Delete this file'],
	  [aria-label='Edit this file'],
	  [aria-label='Edit comment'],
	  [aria-label='Edit comment'],
	  .notification-indicator
```

fregante:
@sindresorhus thoughts on the readability of selectors with `:is()`?

sindresorhus:
I think it's a big improvement.

source/features/hide-obvious-tooltips.css
yakov116: Linter does not let

yakov116: See https://github.com/sindresorhus/refined-github/runs/1999791363

fregante: Looks like a stylelint problem. Content inside parens should be indented

fregante: https://github.com/stylelint/stylelint/issues/5181

fregante (approved):

yakov116:
@fregante when we do the next release we should add **Requires Chrome 88 and above**

PR #4044, opened 2021-03-02
Meta: Replace `dom-chef` with `jsx-dom` 

fregante:
Replaces #3904 

[`jsx-dom`](https://github.com/proteriax/jsx-dom) appears to be a better supported and more mature alternative to `dom-chef` (which basically I've ended up maintaining myself for the past 4 years)

It also includes the PR https://github.com/vadimdemedes/dom-chef/pull/75 which was never finalized with correct types.

- [ ] Update code to use function components 
- [ ] Test


fregante:
No time 😢

Issue #3495, opened 2020-08-24
`hide-readme-header` needs more padding


probablykasper:
Not sure if this should be a bug report or feature request, but as you can see here, no feature toggle shows up when searching for `hide-readme-header`:

![image](https://user-images.githubusercontent.com/11315492/90992111-e0181900-e5ad-11ea-909f-dcc35121f597.png)

fregante:
This is such a minor visual adjustment that doesn’t need a toggle. Why do you want to disable it?

probablykasper:
@farnabaz Because I don't like the lack of padding
![image](https://user-images.githubusercontent.com/11315492/90995464-57ed4000-e5bc-11ea-878e-49e4fbbd3c29.png)

fregante:
Can you post a link to that repo? It should not appear so close to the border

probablykasper:
@fregante Turns out a the padding is decreased a little by the [GitHub Dark](https://github.com/StylishThemes/GitHub-Dark). Without it, it's still pretty minimal padding though:

![image](https://user-images.githubusercontent.com/11315492/91003146-dd7aeb00-e5d0-11ea-906e-dacffa08943d.png)

Repo: https://github.com/probablykasper/vidl

fregante:
I think that this can be solved by adding `padding-top: 30px` to `.Box-body`

PR #4046, opened 2021-03-02
Simplify `hide-readme-header`: Fixes "Jump to definition" bug

### LINKED ISSUE #4033: Method definition popover position
All extension are disabled except for Refined GitHub.
https://github.com/no2chem/bigint-buffer

Screen Shot 2021-02-27 at 22 01 02

Clicking on `toBufferBE` shows the method definition popover near the bottom of the window. Disabling Refined GitHub results in the popover appearing near the `toBufferBE` text.

Chrome 88.0.4324.192
Refined GitHub 21.2.23

fregante:
Fixes #4033 

## Test

https://github.com/probablykasper/vidl <-- the spacing should be reasonable, ref: https://github.com/sindresorhus/refined-github/issues/3495
https://github.com/no2chem/bigint-buffer <-- the popover position should be right, ref #4033 

## Screenshots

Screen Shot

Screen Shot

Screen Shot 1

### DIFF
--- a/source/features/hide-readme-header.css
+++ b/source/features/hide-readme-header.css
@@ -1,36 +1,15 @@
-/* Hide the header in README section */
-.repository-content #readme:not(.blob) {
-	position: relative;
+.repository-content #readme:not(.blob) .Box-title {
+	visibility: hidden;
 }
 
 .repository-content #readme:not(.blob) .Box-header {
-	position: absolute;
-	top: 0;
-	right: 0;
-	background-color: transparent !important;
-	padding: 0 !important;
-	border: none;
-}
-
-.repository-content #readme:not(.blob) .Box-header .Box-title {
-	display: none;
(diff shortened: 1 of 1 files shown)


simonkotwicz:
this causes a spacing issue in GitHub Enterprise (2.21.14)

Screen Shot 2021-03-08 at 4 11 08 PM

busches:
Noticed it today, I'll see if I can fix it tonight, otherwise PR welcome.

simonkotwicz:
Well I think it stems from the fact @fregante removed code that is no longer needed in Github but still needed in older Github enterprise versions so I'm not sure what kind of PR would be accepted or if the solution is just to add custom css.

fregante:
Maybe the only change needed here would be to drop the background and border-bottom, which wouldn’t affect github.com

PR #4048, opened 2021-03-02
Add `night-not-found` feature

fregante:
light not found

### DIFF
--- a/source/features/night-not-found.css
+++ b/source/features/night-not-found.css
@@ -54,3 +54,11 @@
 		filter: brightness(0.15) saturate(0.5);
 	}
 }
+
+.js-plaxify[height='230'] {
+	transition: transform 0.1s, filter 1s;
+}
+
+.js-plaxify[height='230']:hover {
+	filter: none;
+}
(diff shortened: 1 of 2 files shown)


fregante:
Because why not

gif

source/features/night-not-found.css
```diff
@@ -0,0 +1,65 @@
...
+	filter: brightness(0.3) saturate(0.5);
+}
+
+[data-color-mode='dark'] .js-plaxify[height='49'] {
+	filter: brightness(0.9) saturate(0.5);
+}
+
+[data-color-mode='dark'] .js-plaxify[height='415'],
+[data-color-mode='dark'] .js-plaxify[height='75'],
+[data-color-mode='dark'] .js-plaxify[height='50'] {
```
yakov116, line 30: is?

busches:
Can we up the contrast on the 404? I find it really hard to see.

fregante:
Try seeing it in the dark.

Jokes aside, `filters` is rather limited and the 404/message block is a single piece, so to make that more visible I can only lighten the landscape, which kinda defeats the purpose. Maybe I'll try adjusting it when nighttime comes again.

fregante:
No luck. Even a 0.05 brightness increase makes it lose that _night_ feeling. The "404" doesn't have great contrast but it's visible enough unless you use the dark theme in direct sunlight. There's text under it with perfect contrast anyway.

PR #4051, opened 2021-03-03
Restore `next-scheduled-github-action` + show if title overflows

### LINKED ISSUE #4050: `next-scheduled-github-action` stopped working
https://github.com/sindresorhus/refined-github/actions

https://github.com/sindresorhus/refined-github/blob/2533d5c32b5c3426b5fcd5c1ecdef1721f95df8a/source/features/next-scheduled-github-action.tsx#L56

`[href*="?query"]` does not exist any more

cheap-glitch:
Fixes #4050 

## Test URLs
* https://github.com/sindresorhus/refined-github/actions
* https://github.com/sindresorhus/refined-github/actions/workflows/release.yml
* https://github.com/microsoft/TypeScript/actions

This PR:
- Fixes the selector
- Adds a tooltip on the list item for situations like this:
image

### DIFF
--- a/source/features/next-scheduled-github-action.tsx
+++ b/source/features/next-scheduled-github-action.tsx
@@ -77,7 +77,7 @@ void features.add(__filebasename, {
 	include: [
 		pageDetect.isRepositoryActions,
 		// Enable feature on the pages of selected workflows
-		() => /actions\/workflows\/\w+\.ya?ml$/.test(location.pathname)
+		() => /actions\/workflows\/.+\.ya?ml$/.test(location.pathname)
 	],
 	awaitDomReady: false,
 	init
--- a/source/features/next-scheduled-github-action.tsx
+++ b/source/features/next-scheduled-github-action.tsx
@@ -53,7 +53,7 @@ async function init(): Promise<false | void> {
 		return false;
 	}
 
-	for (const workflowListItem of select.all('[href*="?query"]', await elementReady('.hx_actions-sidebar'))) {
+	for (const workflowListItem of select.all('.filter-item[href*="/workflows/"]', await elementReady('.hx_actions-sidebar'))) {
 		if (select.exists('.octicon-stop', workflowListItem)) {
 			continue;
 		}
@@ -65,7 +65,10 @@ async function init(): Promise<false | void> {
 
 		const nextTime = parseCron.nextDate(workflows[workflowName]);
 		if (nextTime) {
-			workflowListItem.append(<em>(next <relative-time datetime={nextTime.toString()}/>)</em>);
+			const relativeTime = <relative-time datetime={nextTime.toString()}/>;
+			workflowListItem.append(<em>(next {relativeTime})</em>);
+			workflowListItem.parentElement!.classList.add('tooltipped', 'tooltipped-n');
+			workflowListItem.parentElement!.setAttribute('aria-label', 'Next ' + relativeTime.textContent!.toString());
 		}
 	}
 }
--- a/source/features/next-scheduled-github-action.tsx
+++ b/source/features/next-scheduled-github-action.tsx
@@ -75,9 +75,7 @@ async function init(): Promise<false | void> {
 
 void features.add(__filebasename, {
 	include: [
-		pageDetect.isRepositoryActions,
-		// Enable feature on the pages of selected workflows
-		() => /actions\/workflows\/.+\.ya?ml$/.test(location.pathname)
+		pageDetect.isRepositoryActions
 	],
 	awaitDomReady: false,
 	init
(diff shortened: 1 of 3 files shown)


cheap-glitch:
Side note: the feature doesn't run on pages where a specific workflow is selected, e.g. https://github.com/sindresorhus/refined-github/actions/workflows/labeler.yml. I don't know if this a bug or an intended behavior from `github-url-detection`.

fregante:
In that case I think we just need to add another include: I think it’s called isSingleWorkflow

cheap-glitch:
There's a `isWorkflowFile()` in `go-to-action-from-file`. If we're going to reuse it, maybe it should belong in `github-url-detection`?

source/features/next-scheduled-github-action.tsx
```diff
@@ -65,14 +65,19 @@ async function init(): Promise<false | void> {
...
 		}
 	}
 }
 
 void features.add(__filebasename, {
 	include: [
-		pageDetect.isRepositoryActions
+		pageDetect.isRepositoryActions,
+		// Enable feature on the pages of selected workflows
+		() => /actions\/workflows\/.+\.ya?ml$/.test(location.pathname)
```
fregante, line 80: I think this actually should be included in `isRepositoryActions` like it was before. GitHub just changed the URL. It used to be `/actions?query=` and now it's `/actions/workflows`, so it no longer matches this page.

source/features/next-scheduled-github-action.tsx
```diff
@@ -65,14 +65,19 @@ async function init(): Promise<false | void> {
 
 		const nextTime = parseCron.nextDate(workflows[workflowName]);
 		if (nextTime) {
-			workflowListItem.append(<em>(next <relative-time datetime={nextTime.toString()}/>)</em>);
+			const relativeTime = <relative-time datetime={nextTime.toString()}/>;
+			workflowListItem.append(<em>(next {relativeTime})</em>);
+			workflowListItem.parentElement!.classList.add('tooltipped', 'tooltipped-n');
```
fregante, line 70: Let's show it on the right, like a continuation of the line itself
```suggestion
			workflowListItem.parentElement!.classList.add('tooltipped', 'tooltipped-e');
```

Screen Shot

source/features/next-scheduled-github-action.tsx
fregante, line 80:
```suggestion
		pageDetect.isRepositoryActions
```

fregante (approved):

source/features/next-scheduled-github-action.tsx
fregante: This PR can be merged after this line is dropped from here and `github-url-detection` is updated here as well.

source/features/next-scheduled-github-action.tsx
```diff
@@ -65,7 +65,10 @@ async function init(): Promise<false | void> {
 
 		const nextTime = parseCron.nextDate(workflows[workflowName]);
 		if (nextTime) {
-			workflowListItem.append(<em>(next <relative-time datetime={nextTime.toString()}/>)</em>);
+			const relativeTime = <relative-time datetime={nextTime.toString()}/>;
+			workflowListItem.append(<em>(next {relativeTime})</em>);
+			workflowListItem.parentElement!.classList.add('tooltipped', 'tooltipped-e');
+			workflowListItem.parentElement!.setAttribute('aria-label', 'Next ' + relativeTime.textContent!.toString());
```
cheap-glitch, line 71: It seems this isn't reliable:
image
Is there an easy way to wait for a custom element to have a text content? `MutationObserver` would be an overkill...

fregante: This is due to `relative-time` not setting it right away, so the solution should be simple

Screen Shot

source/features/next-scheduled-github-action.tsx
```diff
@@ -67,8 +67,10 @@ async function init(): Promise<false | void> {
 		if (nextTime) {
 			const relativeTime = <relative-time datetime={nextTime.toString()}/>;
 			workflowListItem.append(<em>(next {relativeTime})</em>);
-			workflowListItem.parentElement!.classList.add('tooltipped', 'tooltipped-e');
-			workflowListItem.parentElement!.setAttribute('aria-label', 'Next ' + relativeTime.textContent!.toString());
+			setTimeout(() => {
```
fregante, line 70: Comment needed

fregante:
```suggestion
			setTimeout(() => { // The content of `relative-time` might not be immediately available
```

PR #4052, opened 2021-03-03
Fix duplicated content and indentation in `hide-markdown-diff`

### LINKED ISSUE #4035: Hide-markdown-diff not hiding all changed
## Description
The `hide-markdown-diff` is not hiding all changes in the final result preview, and effectively shows an inaccurate result.

#### Before
```
> _💁‍♀️ Protip: Use the minify plugin in-place of the loader to transpile your JS_
>
> The `target` option tells _esbuild_ that it can use newer JS syntax to perform better minification. If you're not using TypeScript or any syntax unsupported by Webpack, you can also leverage this as a transpilation step. It will be faster because there's less files to work on and will produce a smaller output because the polyfills will only be bundled once for the entire build instead of per file.
```

#### After
```
#### _💁‍♀️ Protip: Use the minify plugin in-place of the loader to transpile your JS_
The `target` option tells _esbuild_ that it can use newer JS syntax to perform better minification. If you're not using TypeScript or any syntax unsupported by Webpack, you can also leverage this as a transpilation step. It will be faster because there's less files to work on and will produce a smaller output because the polyfills will only be bundled once for the entire build instead of per file.
```

#### Github default preview

Screen Shot 2021-02-27 at 7 31 13 PM

#### With "View final result" toggled
Screen Shot 2021-02-27 at 7 31 01 PM

## Problems
- The "Protip: ..." is duplicated
- Spacing on the first "Protip: ..." is off
- "target" is highlighted

cheap-glitch:
Fixes #4035 

## Test URLs
Example: https://github.com/sindresorhus/refined-github/edit/main/readme.md

For a more thorough testing:
<details>
<summary>Create a Markdown file with the following content:</summary>

````text
paragraph

paragraph

paragraph [REMOVED]

# level 1 heading
# level 1 heading
# level 1 heading [REMOVED]

## level 2 heading
## level 2 heading
## level 2 heading [REMOVED]

### level 3 heading
### level 3 heading
### level 3 heading [REMOVED]

* list item
* list item
* list item [REMOVED]

1. list item
2. list item
3. list item [REMOVED]

> quote
>
> quote
>
> quote [REMOVED]

[link](https://www.example.com)
[link](https://www.example.com)
[link](https://www.example.com)
[link [REMOVED]](https://www.example.com)

`inline code`
`inline code`
`inline code [REMOVED]`

```
code block
```
```
code block
```
```
code block [REMOVED]
```

```javascript
// code block with language
```
```javascript
// code block with language
```
```javascript
// code block with language
```
```javascript
// code block with language [REMOVED]
```

![image](https://fakeimg.pl/400x100/?text=image)
![image](https://fakeimg.pl/400x100/?text=image)
![image](https://fakeimg.pl/400x100/?text=image)
![image](https://fakeimg.pl/400x100/?text=image%20[REMOVED])

paragraph → heading

#### heading → paragraph

* list item → paragraph

paragraph → list item

> quote → paragraph

paragraph → quote

> quote → heading

#### heading → quote

> multi-line quote → paragraph
>
> multi-line quote → paragraph

> multi-line quote → heading
>
> multi-line quote → paragraph
````
</details>

<details>
<summary>Then edit it and replace its content with:</summary>

````text
paragraph

paragraph [CHANGED]



# level 1 heading
# level 1 heading [CHANGED]


## level 2 heading
## level 2 heading [CHANGED]


### level 3 heading
### level 3 heading [CHANGED]


* list item
* list item [CHANGED]


1. list item
2. list item [CHANGED]


> quote
>
> quote [CHANGED]



[link](https://www.example.com)
[link [CHANGED]](https://www.example.com)
[link](https://www.example.com/CHANGED)


`inline code`
`inline code [CHANGED]`


```
code block
```
```
code block [CHANGED]
```




```javascript
// code block with language
```
```javascript
// code block with language [CHANGED]
```
```typescript
// code block with language
```




![image](https://fakeimg.pl/400x100/?text=image)
![image [CHANGED]](https://fakeimg.pl/400x100/?text=image)
![image](https://fakeimg.pl/400x100/?text=image%20[CHANGED])


#### paragraph → heading

heading → paragraph

list item → paragraph

* paragraph → list item

quote → paragraph

> paragraph → quote

#### quote → heading

> heading → quote

multi-line quote → paragraph
multi-line quote → paragraph


## multi-line quote → heading
multi-line quote → paragraph

````
</details>

This PR:
- removes the decorations on added/changed inline code
- fixes wrong text color applied to changed blockquotes
- fixes the GitHub bug where no indentation is applied to the contents of changed quote blocks
- fixes the GitHub bug where a multi-line quote block changed into a paragraph may contain some duplicated content (see https://github.com/sindresorhus/refined-github/pull/4052#issuecomment-790127924 for a minimal reproducible example)

### CODE REVIEW

File: source/features/hide-markdown-diff.tsx [hunk 1]
```diff
@@ -20,6 +35,7 @@ async function init(): Promise<void> {
 		</div>
 	);
 	delegate(document, '.rgh-preview-button:not(.selected)', 'click', togglePreviewResult);
+	new MutationObserver(fixDuplicatedContent).observe((await elementReady('.js-code-editor'))!, {attributeFilter: ['class']});
```
cheap-glitch, line 38: Couldn't find an event corresponding to the preview finishing its loading, so I used a `MutationObserver`.

source/features/hide-markdown-diff.tsx
```diff
@@ -1,11 +1,26 @@
...
+	if (!select.exists('.js-code-editor.show-preview')) {
+		return;
+	}
+
+	for (const possibleDuplicate of select.all('.markdown-body .changed > .changed_tag[data-before-tag="blockquote"] + p')) {
+		// Content is duplicated if already present in a sibling <ins> element
+		if (possibleDuplicate.textContent && select('ins', possibleDuplicate.parentElement!)?.textContent?.includes(possibleDuplicate.textContent)) {
+			possibleDuplicate.classList.add('removed');
+		}
+	}
```
cheap-glitch, line 21: This works fine, but if it's too much complexity we can drop it and show a warning if `'.markdown-body .changed > .changed_tag[data-before-tag="blockquote"]` matches something. 

Although that doesn't necessarily means there's duplicated content, so it'll cause false warnings. Afaik there's no way in this case to detect duplication without actually checking the content of the elements.

fregante: I had in mind something simpler like checking if there’s a yellow diff. I’d rather show the message often/always instead of maintaining this code

cheap-glitch: > I had in mind something simpler like checking if there’s a yellow diff

A yellow diff just means that a block was modified. It could be anything (paragraph, header, image, etc.) but afaik the duplicated content bug only occurs with quote blocks (and only if they're completely removed and turned into something else).

> I’d rather show the message often/always instead of maintaining this code

:+1: :ok:

fregante:
I'm having a hard time seeing the problem with your demo markdown. This is what I see without this PR

Screen Shot 3

The only issue is the green text, but I see no duplicate content, I think

source/features/hide-markdown-diff.tsx
fregante: Let's leave it there for now, I'd like to see how it works first. Refer to my last comment (I can't replicate the original issue with your demo markdown)

cheap-glitch:
It's around the end:

<table>
<tr>
	<th>With diff
	<th>Without diff
<tr>
	<td>

image
	<td>

image
</table>

source/features/hide-markdown-diff.tsx
cheap-glitch: I think I found a pure CSS solution to this \o/
Needs more testing though

source/features/hide-markdown-diff.css
```diff
@@ -5,11 +5,30 @@
...
+	background-color: var(--color-markdown-code-bg) !important;
+	border: none !important;
+}
+
+.rgh-hide-markdown-diff .markdown-body blockquote :is(ins, .added, .changed) {
+	color: var(--color-text-tertiary) !important;
+}
+
+/* Fix duplicated content and missing indentation in changed quote blocks #4035 */
+.rgh-hide-markdown-diff .markdown-body .changed > p:not([class]):not(:empty) {
```
cheap-glitch, line 24: This targets `<p>` elements that should be wrapped in `<del>` elements but aren't.

fregante: Should there be `blockquote` here just to make it as specific to the issue as possible?

source/features/hide-markdown-diff.tsx
```diff
@@ -1,11 +1,19 @@
...
 import elementReady from 'element-ready';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
 
+// Fix missing indentation in changed quote blocks #4035
+function fixMissingIndentation(): void {
+	for (const changedBlockquote of select.all('.show-preview .changed > .changed_tag[data-before-tag="blockquote"]')) {
+		changedBlockquote.parentElement!.classList.add('ml-3');
+	}
```
cheap-glitch, line 14: In the end I had to rely on some JS to fix the missing indentation issue.

At first I applied a margin on any siblings of `.changed_tag[data-before-tag="blockquote"]` but this causes issues with inline `<ins>` elements, so it has to be applied to the diff block wrapper.

fregante:
Needs a little more work, I think.

Right side: this PR

- the duplicate line is still duplicate in this view, it just changed style

Screen Shot

- also the code block changed style in the final result

Screen Shot 1

- The PR title needs to be more specific than "Fix X" unless the feature is completely broken
- The MutationObserver line needs a comment explaining why it's there or what it's waiting for

cheap-glitch:
> the duplicate line is still duplicate in this view, it just changed style

Actually since I switched to the CSS-only solution the style doesn't change anymore. Anyway, I thought it made more sense to hide the duplicated content only when hiding the whole diff — otherwise we're modifying the native GH behavior even when the feature isn't toggled on, which isn't really what's advertised.

fregante:
> * the duplicate line is still duplicate in this view, it just changed style
> 
> Screen Shot

By this I meant that on the right, the line has a red background. Either we drop it or leave it unchanged since there's already a red-text line below it

cheap-glitch:
Since [`d069190` (#4052)](https://github.com/sindresorhus/refined-github/pull/4052/commits/d069190db040d87e3b35e3e239f8b672d7f3b105) the duplicated content isn't marked in the diff anymore. The only change is the missing indentation that's being re-added:

<table>
<tr>
	<th>Vanilla
	<th>RGH
<tr>
	<td>

image
	<td>
image
</table>

fregante (approved):
It seems to work correctly

source/features/hide-markdown-diff.tsx
```diff
@@ -1,11 +1,19 @@
 import './hide-markdown-diff.css';
 import React from 'dom-chef';
+import select from 'select-dom';
```
fregante, line 3: (Testing a feature, ignore)
```suggestion
import select  from 'select-dom';
```

PR #4059, opened 2021-03-04
Don't mistakenly add `remove-label-faster` when creating a label

ahmed2m:
## Issue
Currently the close button appears in the new label dialog too, and when clicked the api returns 404 (because of course it doesn't exist, yet)

## Test URLs
Any PR/issue and attempt to type a new label name in the tooltip to open the new label dialog 

### Before:
before
### After:
after


fregante (approved):

fregante:
Thanks Ahmed!

PR #4065, opened 2021-03-06
Accept the new token format in the options

ckzhang26:
## Issue
In the next couple of weeks GitHub will change the format for new authentication tokens.  

https://github.blog/changelog/2021-03-04-authentication-token-format-updates/

The characters used will change from [a-f0-9] to [A-Za-z0-9_] and new tokens will include a prefix ("gp1_" for Personal Access Tokens).  Additionally, after June 1, 2021 the token length will increase to up to 255 characters.  Currently, the form to submit personal tokens in the options page will not accept the new token format.  

## Changes
This PR changes the pattern of the form input to accept both existing tokens and tokens of the new format. 

Related: https://github.com/fregante/github-issue-link-status/pull/46


fregante (approved):
Thank you!

yakov116 (approved):

Issue #4058, opened 2021-03-04
Use Preact?


fregante:
Would it make sense to drop dom-chef and just go straight to Preact?

Downsides:

- heavier (only +10/15 KB though)
- slower (given the amount of DOM change, how much really?)
- requires rewrite of many parts that aren't just `.append(<SomeJSX/>)`

Upsides:

- no more type issues
- no more DOM spaghetti
- components that require DOM updates will be easier to handle
- ease of contribution since Preact is known and well documented, unlike the alternatives
- we can use `@primer/components` directly, which could reduce the amount of JSX we carry

We can try converting features gradually to test the feasibility and see what changes it brings. First from the simplest feature, then from a feature that would actually benefit from using Preact.

This will likely conflict with our existing JSX types, so we might have to clutter the codebase with `ts-expect-error` or just add `|| true` in the `build:typescript` npm script 😅

busches:
Why do you think it'll be slower? Just the overhead we're adding that we won't use? I initially thought we'd get a speed up for the DOM, but maybe DOM Chef is really speedy.

fregante:
> Why do you think it'll be slower?

I don't think there's a doubt that preact is slower as:

- dom-chef can't do anything less than it does
- preact does a lot more than dom-chef (as demonstrated by the 3x code increase)

Both receive pre-parsed JSX, so it's not like preact could have an advantage optimizing it before it gets to the browser either.

fregante:
I’m more concerned about the third downside since it requires some work + I still don’t know how to manage things like `.before(<div/>)` in Preact since it can only *append* to a root node.

sindresorhus:
I'm a little bit doubtful it's worth the effort. So I remain -0 on this.

fregante:
I think it mostly depends on the code it requires. If it ends up making it longer and more complex, no.

notlmn:
> Our "change of data" currently directly changes the data on the DOM element. Our only overhead in this case is that sometimes we call `select` to get the DOM element again.

If we are to query DOM elements that we created in JSX we already use, then we don't need select for them. `select` is only used for elements that are outside our JSX scope, and that belongs to the DOM generated GitHub. Is'nt that the case?

-----

I really see an upside to DOM diffing instead of `root.innerHTML = <App />` or something like that. But we still have to use `MutationObserver`s all over the place, that's again because most of the DOM is outside our JSX scope. This comes down to how each feature ends up using DOM diffing or other features of Preact.

[...]

PR #4069, opened 2021-03-07
Try preact 

fregante:
#4058


fregante:
Too much work. Too many features.

PR #4070, opened 2021-03-07
Fix `suggest-commit-title-limit` input field size

### LINKED ISSUE #3998: `suggest-commit-title-limit` is sized incorrectly, 1-character short
Tested in Safari. Happens when you're about to merge a PR

gif

The field should match the length limit, but it doesn't anymore. Might have to do with `padding` or `box-sizing`

xtianatilano:
Fixes #3998

## Test URLs
Any PR for merge + This 72-char title:

```
123456789 123456789 123456789 123456789 123456789 123456789 123456789 12
```

## Screenshot

Before:
before

After:
suggest-commit-title

### DIFF
--- a/source/features/monospace-textareas.css
+++ b/source/features/monospace-textareas.css
@@ -1,7 +1,8 @@
 /* Limit width of commit title to 72 characters */
 .rgh-monospace-textareas :is(#merge_title_field, #commit-summary-input) {
-	width: calc(73ch + 18px);
+	width: calc(72ch + 18px);
 	max-width: 100%;
+	box-sizing: content-box;
 }
 
 .rgh-monospace-textareas :is(#merge_title_field, #commit-summary-input, textarea) {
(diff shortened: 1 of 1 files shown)


fregante:
In your gif, as you move the cursor from the end to the start, you can see the text shifting by a pixel. This shows that there’s still overflow. 

The 72ch is the correct size here and should not change. This is more likely caused by `box-sizing: border-box`, which should be changed to `box-sizing: content-box`. Does that fixing?

xtianatilano:
Updated. 

`box-sizing: content-box` does fix it, I initially thought that the extra space was kinda too long. 😅

fregante (approved):
Thank you!

fregante (changes requested):
Tested, actually while it fixes the bug, the final width is longer than 72-chars

gif

fregante (approved):
Thankfully the fix was easy 
gif

Issue #3999, opened 2021-02-19
Support more filenames in `link-to-changelog-file`


jack1142:
While `CHANGELOG` is one way people call changelog files, it is not the only popular one out there.
I've personally seen [`CHANGES`](https://github.com/sphinx-doc/sphinx/), [`NEWS`](https://github.com/pypa/pip), and [`HISTORY`](https://github.com/psf/requests) used often too.

I think that [names supported by dependabot](https://github.com/dependabot/dependabot-core/blob/e7c9f44410c16a8ef3da41bda949d2c110c3ed0e/common/lib/dependabot/metadata_finders/base/changelog_finder.rb#L20) are a good reference on what names are used in the wild.

tooomm:
I assume in the example repo https://github.com/pypa/pip, this feature is supposed to link to https://github.com/pypa/pip/blob/master/NEWS.rst, not https://github.com/pypa/pip/tree/master/news?

jack1142:
Yeah, it should be linking to a changelog file, not the folder with changelog fragments.
It seems that this has been mentioned in https://github.com/sindresorhus/refined-github/pull/4005#issuecomment-782845287, but I'm not sure if there was a reason for this to not be updated.

If detecting folders was intentional, I think the extension should *at least* prioritize the files but I'm not sure what was the original intention here.

yakov116:
> If detecting folders was intentional, I think the extension should _at least_ prioritize the files but I'm not sure what was the original intention here.

👍 

https://github.com/sindresorhus/refined-github/blob/d5b8ecf08b6a5d63127c2e4308da71383ec5c63e/source/features/link-to-changelog-file.tsx#L26

Adding a simple `.reverse()` should work here

https://github.com/sindresorhus/refined-github/blob/d5b8ecf08b6a5d63127c2e4308da71383ec5c63e/source/features/link-to-changelog-file.tsx#L43

It might work ⬆️ too but I am not sure.

fregante:
Unfortunately a reverse would also catch `changelog.3.md` over `changelog.md`

Maybe we should just filter both by type, then only match exactly `*` and `*.*`, never `*.*.*`

yakov116:
Api might be simple to do that. DOM not so much, It would require another select to check the icon

fregante:
Select the icon first, get the filename from there.

PR #4071, opened 2021-03-07
Exclude folders from `link-to-changelog-file`

### LINKED ISSUE #4066: Prioritize files over folders (or don't include folders) in `link-to-changelog-file`
A follow-up issue for [the comment in #3999](https://github.com/sindresorhus/refined-github/issues/3999#issuecomment-784114571) as it doesn't make much sense to track this in an already closed issue.

cheap-glitch:
Closes #4066 

## Test URLs
https://github.com/pypa/pip/releases

Two things to test:
- going directly to the release page
- going to the repo root (with a cleared cache), then to the release page

It was unclear to me what was decided about this, so this PR takes the simplest approach of ignoring folders altogether.

If we want to include them but prioritize files, it would be as simple as separating the list in two instead of filtering it (and maybe also make the selection logic a bit stricter as mentioned in https://github.com/sindresorhus/refined-github/issues/3999#issuecomment-784258402).

### DIFF
--- a/source/features/link-to-changelog-file.tsx
+++ b/source/features/link-to-changelog-file.tsx
@@ -41,7 +41,12 @@ const getChangelogName = cache.function(async (): Promise<string | false> => {
 			}
 		}
 	`);
-	const files: string[] = repository.object.entries.filter((entry: AnyObject) => entry.type === 'blob').map((file: AnyObject) => file.name);
+	const files: string[] = [];
+	for (const entry of repository.object.entries) {
+		if (entry.type === 'blob') {
+			files.push(file.name)
+		}
+	}
 	return findChangelogName(files);
 }, {
 	cacheKey: getCacheKey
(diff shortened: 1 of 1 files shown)


fregante:
> It was unclear to me what was decided about this, so this PR takes the simplest approach of ignoring folders altogether.

This should be fine. Like for `bugs-tab` that only matches the `bugs` label, we only need to match standard/common naming. Repos that aren't standard will _suffer_ the same way they would if they had the readme under `/readme/index.md` instead of `/readme.md`

fregante (approved):
It works, however can you also replace `HEAD` with `getDefaultBranch` in the same PR for a double fix? Not sure why we didn't do it before.

fregante:
Or separate PR is fine

yakov116:
We can't https://github.com/sindresorhus/refined-github/pull/3958#discussion_r574824452 remember?

PR #4072, opened 2021-03-07
Avoid `conversation-activity-filter` duplicates

### LINKED ISSUE #4067: Multiple eye icons
img
https://github.com/harryfei/which-rs/pull/30

This only happens occasionally.

kidonng:
Fix #4067

See https://github.com/sindresorhus/refined-github/issues/4067#issuecomment-792273884 for details

## Test URLs

1. Open issue https://github.com/sindresorhus/refined-github/issues/4067
2. Go to issues tab
3. Open any issue again

## Screenshot

None

P.S. maybe we should rewrite all `onetime` usage for `selector-observer` with `observer.abort()`?

### DIFF
--- a/source/features/conversation-activity-filter.tsx
+++ b/source/features/conversation-activity-filter.tsx
@@ -22,7 +22,7 @@ let currentSetting: State = 'default';
 const dropdownClass = 'rgh-conversation-activity-filter-dropdown';
 const hiddenClassName = 'rgh-conversation-activity-filtered';
 
-let observer: Observer
+let observer: Observer;
 
 function isWholeReviewEssentiallyResolved(review: HTMLElement): boolean {
 	const hasMainComment = select.exists('.js-comment[id^=pullrequestreview] .timeline-comment', review);
--- a/source/features/conversation-activity-filter.tsx
+++ b/source/features/conversation-activity-filter.tsx
@@ -3,7 +3,7 @@ import delay from 'delay';
 import React from 'dom-chef';
 import select from 'select-dom';
 import * as pageDetect from 'github-url-detection';
-import SelectorObserver from 'selector-observer';
+import {observe, Observer} from 'selector-observer';
 import {CheckIcon, EyeClosedIcon, EyeIcon} from '@primer/octicons-react';
 
 import features from '.';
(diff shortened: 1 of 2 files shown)


fregante:
How is this different? The `observe` export is just a pre-bound `new SelectorObserver` from what I saw

kidonng:
https://github.com/josh/selector-observer/blob/b811bb5d13cba9940026430f92a6760d5f9e365c/lib/selector-observer.js#L73

`observer.abort()` will remove the selector from `SelectorObserver`'s ` selectorSet` (yup, that's two different observer object).

fregante:
I opened https://github.com/josh/selector-observer/issues/32

fregante (approved):
The code looks right. I don’t know if it fixes the issue because I wasn’t able to replicate it. 

As for the deinit, yeah that’s not a bad idea

yakov116:
> `observer.abort()` will remove the selector from `SelectorObserver`'s ` selectorSet` (yup, that's two different observer object).

Shouldn't this be changed in all features?

ahmed2m (approved):
Looks like it solved the issue for me.

cheap-glitch:
Can confirm this fixes the bug (tested on Firefox).

> I don’t know if it fixes the issue because I wasn’t able to replicate it.

It happens when navigating between conversations using "linked references". E.g. when clicking on the "Fix 4067" link at the top of this page, then on the PR link at the bottom, and repeating this several times:

Screenshot_2021-03-07 Prevent `conversation-activity-filter` from running multiple times by kidonng · Pull Request #4072 ·

fregante:
> > `observer.abort()` will remove the selector from `SelectorObserver`'s ` selectorSet` (yup, that's two different observer object).
> 
> Shouldn't this be changed in all features?

Most features already have `onetime` and don’t have a deinit, so they aren’t affected by https://github.com/josh/selector-observer/issues/32

But yeah it’s a good idea, as I said previously.

fregante:
Should we disable this feature in the hotfix? I don’t think we’ll do another release just yet, but I don’t want people to be annoyed by it.

yakov116:
> Should we disable this feature in the hotfix branch?

Noooooo. I used this a few times today. 

Its great people can "see" the feature very well 👁️

fregante:
> > Should we disable this feature in the hotfix branch?
> 
> Noooooo. I used this a few times today.

Sideloaded extensions aren't affected; Only store-downloaded ones are.

fregante:
👆 

In this specific case it doesn't matter but, as a general pattern, what do you think if `deinit` is an array of functions to call? This has a few advantages:

- no risk of overriding the single `let observer` variable
- lets you have multiple deinit callbacks without having to defined each separately at the top of the file. This happens only once for `selector-observer` but there's also `delegate`, for which we could also do the same, everywhere

fregante (approved):

PR #4077, opened 2021-03-08
Don't trigger `unfinished-comments` for open comment dropdowns

cheap-glitch:
## Test URLs
1. Go to a PR with reviews, e.g. https://github.com/sindresorhus/refined-github/pull/4023
2. Open the dropdown menu of any review comment
3. Go to a different tab
4. The title of the previous tab should NOT say `(Draft comment)`

### DIFF
--- a/source/features/unfinished-comments.tsx
+++ b/source/features/unfinished-comments.tsx
@@ -7,7 +7,7 @@ let documentTitle: string | undefined;
 
 function hasDraftComments(): boolean {
 	// `[disabled]` excludes the PR description field that `wait-for-build` disables while it waits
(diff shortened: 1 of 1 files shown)


fregante:
Can we have an explicit value check instead? `field.value !== ''`. Empty fields are never something we need to save

cheap-glitch:
> Empty fields are never something we need to save, even if they’ve been changed

True, but we still need the id check (or something equivalent) because those hidden textareas are not empty:
screenshot-1615224951
(Maybe the comment should make that clearer)

fregante (approved):

fregante:
I noticed that the bug happens on any comment, not just reviews

PR #4078, opened 2021-03-08
Meta: Replace `getScopedSelector()` with `:is()`

cheap-glitch:
Follow-up of https://github.com/sindresorhus/refined-github/pull/4023#discussion_r589125189.


yakov116:
Awesome!

fregante:
Indeed! I always found it super awkward but we didn’t have much choice

fregante (approved):

PR #4082, opened 2021-03-10
Fix console errors on some PR pages

### LINKED ISSUE #3932: `onPrMergePanelOpen()` may be broken
1. Go to a PR page (e.g. https://github.com/sindresorhus/refined-github/pull/3886)
2. Type something in the comment form at the bottom
3. Reload the page

Two features will produce an error (sometimes only one of them appear):
```text
clear-pr-merge-commit-message → TypeError: messageField is undefined
suggest-commit-title-limit → TypeError: inputField is undefined
```

Since both of these features make use of `onPrMergePanelOpen`, I'm guessing that's where the problem is.

cheap-glitch:
Fixes #3932 

## Test URLs
1. Type something in the comment box at the bottom (without sending it)
2. Soft-reload the page
3. `clear-pr-merge-commit-message` and `suggest-commit-title-limit` should NOT produce errors in the console

Tested in Firefox & Chromium.

### DIFF
--- a/source/github-events/on-pr-merge-panel-open.ts
+++ b/source/github-events/on-pr-merge-panel-open.ts
@@ -8,10 +8,16 @@ const delegateHandler = mem((callback: EventListener) => (event: delegate.Event)
 	}
 });
 
-const sessionResumeHandler = mem((callback: EventListener) => async (event: Event) => {
+declare global {
+	interface DocumentEventMap {
+		'session:resume': CustomEvent;
+	}
+}
+
+const sessionResumeHandler = mem((callback: EventListener) => async (event: CustomEvent) => {
 	await Promise.resolve(); // The `session:resume` event fires a bit too early
 	// Avoid triggering the callback when a page with a non-empty comment field is reloaded #3932
-	if ((event as CustomEvent).detail.targetId !== 'new_comment_field') {
+	if (event.detail.targetId !== 'new_comment_field') {
 		callback(event);
 	}
 });
--- a/source/github-events/on-pr-merge-panel-open.ts
+++ b/source/github-events/on-pr-merge-panel-open.ts
@@ -8,12 +8,6 @@ const delegateHandler = mem((callback: EventListener) => (event: delegate.Event)
 	}
 });
 
-declare global {
-	interface DocumentEventMap {
-		'session:resume': CustomEvent;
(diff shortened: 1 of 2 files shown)


source/github-events/on-pr-merge-panel-open.ts
```diff
@@ -10,7 +10,10 @@ const delegateHandler = mem((callback: EventListener) => (event: delegate.Event)
 
 const sessionResumeHandler = mem((callback: EventListener) => async (event: Event) => {
 	await Promise.resolve(); // The `session:resume` event fires a bit too early
-	callback(event);
+	// Avoid triggering the callback when a page with a non-empty comment field is reloaded #3932
+	if ((event as CustomEvent).detail.targetId !== 'new_comment_field') {
```
cheap-glitch, line 14: Not sure about the type coercion. Maybe the callback should be defined as taking a `CustomEvent` in the first place?

fregante: Yes if it works

cheap-glitch: It can work without any assertions if we declare `'session:resume'` as a `CustomElement`:
```typescript
declare global {
	interface DocumentEventMap {
		'session:resume': CustomEvent
	}
}
```

source/github-events/on-pr-merge-panel-open.ts
```diff
@@ -8,9 +8,18 @@ const delegateHandler = mem((callback: EventListener) => (event: delegate.Event)
 	}
 });
 
-const sessionResumeHandler = mem((callback: EventListener) => async (event: Event) => {
+declare global {
+	interface DocumentEventMap {
+		'session:resume': CustomEvent;
```
fregante, line 13: global.d.ts has other events too, it should be there, not here

fregante (approved):

fregante:
Thank you for looking into this!

There's also a related behavior that I finally put into an issue: https://github.com/sindresorhus/refined-github/issues/4101

yakov116:
Looks like this broke `clear-pr-merge-commit-message` not sure why. Its very hard to test because once the field is updated GitHub saves what it was updated to.

I reverted this PR locally and then the feature started working correctly

PR #4089, opened 2021-03-12
Add `view-last-pr-deployment` feature

### LINKED ISSUE #2918: Deployment shortcut in pull request headers
Not sure how widespread the use of the GitHub deployments API is, but we use it at my company to report staging deployments for all our pull requests. Problem is, as a pull request timeline grows, it becomes more and more of a pain to dig through to find the link to the latest deployment.

image

I actually built this into my own Chrome extension already, but I think more people could get some use out of it, so I'm opening an issue here.

Demo:
Kapture 2020-03-22 at 15 39 46

Example URL: https://github.com/zeit/next-site/pull/553

yakov116:
Closes #2918 

## Test URLs
https://github.com/yakov116/pr-deployments/pull/1
https://github.com/excalidraw/excalidraw/pull/2993
https://github.com/primer/octicons/pull/453 (a blast from the past ⭐)

## Screenshot
image


source/features/view-last-pr-deployment.tsx
```diff
@@ -0,0 +1,27 @@
...
+	button.classList.add('mr-1');
+	button.textContent = 'View last deployment';
+
+	select('.gh-header-actions')!.prepend(button);
+}
+
+void features.add(__filebasename, {
+	include: [
+		pageDetect.isPRConversation
+	],
```
fregante, line 23: Given that most PRs don't have this, let's not add an observer for them
```suggestion
	],
	exclude: [
		() => !select.exists(deploymentsSelector)
	],
```

yakov116: What if a deployment is triggered for the first time after the person visits the page? Don't worry about it?

fregante: Yeah. In that slim chance, the link will likely be at the bottom of the page, easily accessible anyway.

yakov116:
Do I get your approval?

source/features/view-last-pr-deployment.tsx
```diff
@@ -0,0 +1,27 @@
...
+	exclude: [
+		() => !select.exists(deploymentSelector)
+	],
+	init() {
+		observeElement(select('#partial-discussion-header')!.parentElement!, init);
```
fregante, line 25: If a new deployment happens, the button is stale, so maybe it should also include `onNewComments`

source/features/view-last-pr-deployment.tsx
```diff
@@ -0,0 +1,27 @@
...
+import observeElement from '../helpers/simplified-element-observer';
+
+const deploymentSelector = '.js-timeline-item [data-url$="deployed"] .TimelineItem-body .btn[target="_blank"]';
+
+function init(): void {
+	const lastDeployment = select.last(deploymentSelector)!.cloneNode(true);
+	lastDeployment.classList.add('mr-1');
+	lastDeployment.textContent = 'View last deployment';
+
+	select('.gh-header-actions')!.prepend(lastDeployment);
```
fregante, line 14: I think it's safer to just generate the link via JSX and only copy the URL from the button. It shouldn't be much longer
```suggestion
	const {href} = select.last(deploymentSelector)!;
	select('.gh-header-actions')!.prepend(
		<a className="btn mr-1" href={url} target="_blank">Last deployment</a>
	);
```

fregante: Maybe if you include `a` in the selector it will use HTMLAnchorElement automatically

source/features/view-last-pr-deployment.tsx
yakov116: No need the header will refresh.

I tested it on https://github.com/yakov116/pr-deployments/pull/1

source/features/view-last-pr-deployment.tsx
yakov116: > I think it's safer

Which part?

yakov116: > Maybe if you include `a` in the selector it will use HTMLAnchorElement automatically

no luck

source/features/view-last-pr-deployment.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+import features from '.';
+import observeElement from '../helpers/simplified-element-observer';
+
+const deploymentSelector = '.js-timeline-item [data-url$="deployed"] .TimelineItem-body .btn[target="_blank"]';
+
+function init(): void {
+	const {href} = select.last<HTMLAnchorElement>(deploymentSelector)!;
+	select('.gh-header-actions')!.prepend(
+		<a className="btn btn-sm btn-outline" href={href} target="_blank" rel="noreferrer">Last deployment</a>
```
fregante, line 13: Should we add an "external" icon? Maybe just `Deployment [icon]`

yakov116:
image

How does this look?

fregante:
Tooltip is redundant, deployment should be lowercase (I don’t think there's anything in Title Case on GH), the icon makes it feel like an action, I'd rather have the regular external icon on the right side

yakov116:
> Tooltip is redundant, 
👍 
>deployment should be lowercase (I don’t think there's anything in Title Case on GH), 
👍 
>the icon makes it feel like an action, 
ok
>I'd rather have the regular external icon on the **right side*** as suggested

GitHub style guide always has the icon on the left

yakov116:
image

fregante:
> > I'd rather have the regular external icon on the **right side*** as suggested
> 
> GitHub style guide always has the icon on the left

True

> > deployment should be lowercase (I don’t think there's anything in Title Case on GH),
> 
> 👍

fregante (changes requested):
The comment above 👆 + every test link I opened showed 2 buttons in Chrome, at least the first time

Screen Shot 5

yakov116:
> The comment above 👆 + every test link I opened showed 2 buttons in Chrome, at least the first time

I cannot understand why this is happening! I console.logged it and this observer runs 2 times yet on `first-published-tag-for-merged-pr` it does not 🤷

fregante:
> it runs 2 times yet on [`first-published-tag-for-merged-pr`](/sindresorhus/refined-github/blob/main/source/features/first-published-tag-for-merged-pr.tsx) it does not 🤷

Probably it runs twice there too but then we have

https://github.com/sindresorhus/refined-github/blob/19330b2a4d1715c2d2496e6435a6b43263031c03/source/features/first-published-tag-for-merged-pr.tsx#L33

Maybe we should have an optimized `on-conversation-header-update.ts`?

yakov116:
image

source/features/view-last-pr-deployment.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+import features from '.';
+import onConversationHeaderUpdate from '../github-events/on-conversation-header-update';
+
+const deploymentSelector = '.js-timeline-item [data-url$="deployed"] .TimelineItem-body .btn[target="_blank"]';
+
+function init(): void | false {
+	if (select.exists('.rgh-last-deployment')) {
+		return false;
+	}
```
fregante, line 14: -> `exclude: […]`

fregante (changes requested):
Let's make the button text black

source/features/view-last-pr-deployment.tsx
yakov116: The exists is do to init being called 2 times.

fregante: That's what `exclude` is for. I think:

```js
	exclude: [
 		() => !select.exists(deploymentSelector),
		() => select.exists('.rgh-last-deployment')
 	],
```

Or a single function, if you figure out the correct boolean logic 😃

yakov116: From what I understand exclude only runs once here since we are using `observeElement`, and its not the feature loader is not calling the init.

Please correct me if I am wrong

fregante: 😱 

I was totally sure `additionalListeners` still checked `include`/`exclude`, but they don't 👍

yakov116: Note to self remove the false. (We only add the false for the logger, here it would have ran already right?)

yakov116:
image

fregante (commented):
You can make this change either here or somewhere else. It looks good otherwise. I hope people don’t complain about wrapping in small viewports

source/features/first-published-tag-for-merged-pr.tsx
```diff
@@ -50,7 +50,8 @@ void features.add(__filebasename, {
 	include: [
 		() => pageDetect.isPRConversation() && pageDetect.isMergedPR()
 	],
-	init() {
-		observeElement(select('#partial-discussion-header')!.parentElement!, init);
-	}
+	additionalListeners: [
+		onConversationHeaderUpdate
+	],
+	init
```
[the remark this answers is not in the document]
yakov116: or do additionalListenersOnly

yakov116:
> I hope people don’t complain about wrapping in small viewports

We can hide it...

fregante:
The position was a little too _important_ in small windows
Screen Shot 2

So I hid it for now

dwelle:
Thanks @yakov116 & @fregante! ❤️

fregante:
Screen Shot 5

Bug when the Edit button is visible

yakov116:
> Bug when the Edit button is visible. Can probably be fixed by a single space in the dom

Until the next header update.

I could not always get it to do that. Sometimes yes sometimes no. 🤷‍♂️

PR #4093, opened 2021-03-12
Correctly remove listeners when leaving certain pages

yakov116:
Avoids many console error.

Without this pr go from one issue to the next and the do history.back() there will be a console error. `clean-conversation-headers`
Go from the conversation tab to the files changed there will be a console error. `sticky-sidebar`

If needed I can paste all the console errors

### DIFF
--- a/source/features/sticky-sidebar.tsx
+++ b/source/features/sticky-sidebar.tsx
@@ -38,6 +38,7 @@ void features.add(__filebasename, {
 	],
 	init,
 	deinit: () => {
+		selectObserver.abort();
 		resizeObserver.disconnect();
 		window.removeEventListener('resize', onResize);
 	}
--- a/source/features/sticky-sidebar.tsx
+++ b/source/features/sticky-sidebar.tsx
@@ -22,7 +22,7 @@ let selectObserver: Observer;
 function init(): void {
 	selectObserver = observe(sidebarSelector, {
 		add(sidebar) {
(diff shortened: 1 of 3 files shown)


source/features/sticky-sidebar.tsx
```diff
@@ -38,5 +39,6 @@ void features.add(__filebasename, {
 	init,
 	deinit: () => {
 		observer.disconnect();
+		selectObserver.abort();
 	}
```
fregante, line 43: Both of these functions should be added to a deinit array instead

source/features/sticky-sidebar.tsx
```diff
@@ -17,9 +17,10 @@ function updateStickiness(): void {
 
 const onResize = debounce(updateStickiness, {wait: 100});
 const observer = new ResizeObserver(onResize);
```
fregante, line 19: Can we also rename this appropriately? Too many observers on one page

source/features/sticky-sidebar.tsx
yakov116: I does not work 😢 

I get `Illegal invocation`

fregante (approved):

fregante:
Are these the only ones?

yakov116:
Don't merge. With the current implementation I have errors on `history.back()`. I will double check later today

fregante:
Merge if done

PR #4098, opened 2021-03-14
Add `global-search-filters` feature

### LINKED ISSUE #3613: Show forks by default when the search is sorted by stars
[GitHub doesn't show forks by default when searching](https://docs.github.com/en/free-pro-team@latest/github/searching-for-information-on-github/searching-in-forks), this is fine most of the time, but it often leads to notable forks being ignored.

<details>
<summary>Example</summary>

- https://github.com/search?o=desc&q=deno&s=stars&type=Repositories

image

- https://github.com/search?o=desc&q=deno+fork%3Atrue&s=stars&type=Repositories

image

</details>

Notice in the example, how [denolib/awesome-deno](https://github.com/denolib/awesome-deno) isn't shown by default even it has plenty of stars.

There are existing solutions to this problem such as [Lovely forks](https://github.com/musically-ut/lovely-forks), but we can solve this from another aspect. When a user sorts the search result by stars, we can append `fork:true` automatically so notable forks show up in the search result by default.

This feature should appear on all `https://github.com/search?s=stars&type=Repositories` URL.

kidonng:
Resolve #3613

Better names welcome

## Test URLs

- Sidebar with only categories: https://github.com/search?q=author%3A%40me&type=commits
- Sidebar with categories and languages: https://github.com/search?q=waifu2x-converter-cpp
- Sidebar with  categories, states and languages: https://github.com/search?q=author%3A%40me&type=issues

## Screenshot

image


kidonng:
Actually I have some questions unsolved:

- Should we show a link with `fork: true` if the user doesn't use the dropdown, for example if they clicked a link?
- Should we remove `fork: true` if the user switch to other sorting options?

kidonng:
I think I will go for the alternate option mentioned [here](https://github.com/sindresorhus/refined-github/issues/3613#issuecomment-796695408): add a checkbox to include forks

Take waifu2x-converter-cpp fork:true as an example, you already get good enough results when including forks. You don't have to change sorting criteria.

source/features/show-forks-in-search.css
```diff
@@ -0,0 +1,3 @@
+.rgh-show-forks-in-search-checkbox {
+	vertical-align: middle;
```
fregante, line 2: I think that there's a GitHub class already for this, like `v-align-middle`

source/features/show-forks-in-search.tsx
```diff
@@ -0,0 +1,33 @@
...
+import React from 'dom-chef';
+import select from 'select-dom';
+import * as pageDetect from 'github-url-detection';
+
+import features from '.';
+import SearchQuery from '../github-helpers/search-query';
+
+function init(): void {
+	const includeForks = new SearchQuery(location.search).includes('fork:true');
+	const checkbox = <label className="float-right"><input className="rgh-show-forks-in-search-checkbox mr-1" type="checkbox" checked={includeForks}/>Include forks</label>;
```
fregante, line 11:
```suggestion
	const checkbox = (
		<label className="float-right">
			  <input className="rgh-show-forks-in-search-checkbox mr-1" type="checkbox" checked={includeForks}/>
			  Include forks
		</label>
	);
```

source/features/show-forks-in-search.tsx
```diff
@@ -0,0 +1,33 @@
+import './show-forks-in-search.css';
+import React from 'dom-chef';
+import select from 'select-dom';
+import * as pageDetect from 'github-url-detection';
+
+import features from '.';
+import SearchQuery from '../github-helpers/search-query';
+
+function init(): void {
+	const includeForks = new SearchQuery(location.search).includes('fork:true');
```
fregante, line 10: :tada:

source/features/show-forks-in-search.tsx
```diff
@@ -0,0 +1,33 @@
...
+	const checkbox = <label className="float-right"><input className="rgh-show-forks-in-search-checkbox mr-1" type="checkbox" checked={includeForks}/>Include forks</label>;
+	checkbox.addEventListener('click', () => {
```
fregante, line 12: If the page has PJAX enabled, this listener will be lost on `history.back()`. Try and if it breaks, it needs to use `delegate`

Also it should probably be `change` instead of `click`

source/features/show-forks-in-search.tsx
```diff
@@ -0,0 +1,33 @@
...
+		for (const link of select.all<HTMLAnchorElement>('.codesearch-results .select-menu-item')) {
+			if (includeForks) {
+				new SearchQuery(link).remove('fork:true');
+			} else {
+				new SearchQuery(link).add('fork:true');
+			}
+
+			if (link.getAttribute('aria-checked') === 'true') {
+				link.click();
+			}
```
fregante, line 22: Why change all links in the loop if we're about to lose the dom with `link.click()`?

fregante (changes requested):
If I understand the code correctly, even if you correctly detect `forks:true` on load and mark the checkbox as `checked`, the links in the dropdown aren't actually updated in this case.

Example:

1. Visit the search page
2. Click the checkbox
3. The page reloads, right?
4. Now select another item in the list
5. The page reloads, but `forks:true` is lost

---

There's also a second issue here, I think: It's being shown as a checkbox but clicking it will navigate to the new search page immediately. Checkboxes shouldn't reload pages.

Also, I think this is a bad place for the checkbox since it's a sort menu, not a filter menu.

source/features/show-forks-in-search.tsx
kidonng: I remember linter would yell at me for this before, but it didn't so I passed it for now 😄

kidonng:
> There's also a second issue here, I think: It's being shown as a checkbox but clicking it will navigate to the new search page immediately. Checkboxes shouldn't reload pages.

Well, I just think it's not obvious you have to click the link again to see the new results.

> Also, I think this is a bad place for the checkbox since it's a sort menu, not a filter menu.

Tried to place it next to the dropdown, but the container seems to need a lot of modifications.

source/features/show-forks-in-search.tsx
kidonng: I was being dumb here, possibly I can just get rid of the loop.

source/features/show-forks-in-search.tsx
kidonng: > Also it should probably be `change` instead of `click`

I noticed this, but haven't fixed it yet. Well it says `checkbox` but it is a in fact a label 😅

fregante:
> Tried to place it next to the dropdown, but I'm bad at positioning it (and make it responsive). Suggestions welcome.

Filters are in the sidebar. We can add a whole block like this one, even if it's just one "Forks" link inside it (with an X to remove the filter, just like the Languages box)

Screen Shot

Thoughts?

source/features/show-forks-in-search.tsx
fregante: If you find incomprehensible/difficult build/lint errors you can ask for help, as long as the logic works.

kidonng:
> Filters are in the sidebar.

Wow I totally forget that.

> We can add a whole block like this one, even if it's just one "Forks" link inside it (with an X to remove the filter, just like the Languages box)

Hmm actually this opens new possibilities. Maybe add some other filters as well? At least I found `is:private` useful too.

fregante:
> `user: myself` useful too

You can already start the search with your own user

Screen Shot

> Maybe add some other filters as well

I'm -0 on this.


> Possible design

That's it, as long as it doesn't actually replace the Languages one

kidonng:
> I'm -0 on this.

I won't turn it into another long list of filters, and after reading the list of available filters I end up with only those I mentioned.

> That's it, as long as it doesn't actually replace the Languages one

Why would I do that

fregante:
> > That's it, as long as it doesn't actually replace the Languages one
> 
> Why would I do that

That's what the screenshot looks like 😃

source/features/search-filters.tsx
```diff
@@ -0,0 +1,48 @@
...
+			item.classList.add('selected');
+			item.prepend(<XIcon className="float-right"/>);
+		} else {
+			query.add(filter);
+		}
+
+		return <li>{item}</li>;
+	});
+
+	select('.menu')!.nextElementSibling!.after(
```
fregante, line 31: Selector wait too broad

source/features/search-filters.tsx
```diff
@@ -0,0 +1,48 @@
...
+import {getUsername} from '../github-helpers';
+
+function init(): void {
+	const items = [
+		['Forks', 'fork:true'],
+		['Private', 'is:private'],
+		['Yours', `user:${getUsername()}`]
+	].map(([name, filter]) => {
+		const item = <a className="filter-item" href={location.href}>{name}</a>;
+		// @ts-expect-error
```
fregante, line 17: This should be fixed before merging. Let it fail or else we don't even see the error and help
```suggestion
```

source/features/search-filters.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+import features from '.';
+import SearchQuery from '../github-helpers/search-query';
+import {getUsername} from '../github-helpers';
+
+function init(): void {
+	const items = [
+		['Forks', 'fork:true'],
+		['Private', 'is:private'],
+		['Yours', `user:${getUsername()}`]
```
fregante, line 14:
```suggestion
		['Yours', `user:@me`]
```

source/features/search-filters.tsx
```diff
@@ -0,0 +1,48 @@
...
+	].map(([name, filter]) => {
```
fregante, line 15:
```diff
- .map()
+ for of
```

and then
```diff
- return
+ items.push()
```

source/features/search-filters.tsx
kidonng: `item` is `JSX.Element` while `SearchQuery` expects `HTMLAnchorElement`.

readme.md
```diff
@@ -315,6 +315,7 @@ Thanks for contributing! 🦋🙌
 - [](# "improve-shortcut-help") [Shows all of Refined GitHub’s new keyboard shortcuts in the help modal (<kbd>?</kbd> hotkey).](https://user-images.githubusercontent.com/29176678/36999174-9f07d33e-20bf-11e8-83e3-b3a9908a4b5f.png)
 - [](# "stop-redirecting-in-notification-bar") [Stops redirecting to notification inbox from notification bar actions while holding <kbd>Alt</kbd>.](https://user-images.githubusercontent.com/202916/80318782-c38cef80-880c-11ea-9226-72c585f42a51.png)
 - [](# "hide-zero-packages") [Hides the `Packages` tab in user profiles if it’s empty.](https://user-images.githubusercontent.com/46634000/101151879-9ade9b00-3622-11eb-9755-50ed54099e02.png)
+- [](# "global-search-filters") [Adds useful search filters in search page.](https://user-images.githubusercontent.com/44045911/111149280-a6e5db80-85c7-11eb-9e61-4badb9d3c685.png)
```
fregante, line 318:
```suggestion
- [](# "global-search-filters") [Adds search filters in global search page: forks, private repos, own repos.](https://user-images.githubusercontent.com/44045911/111149280-a6e5db80-85c7-11eb-9e61-4badb9d3c685.png)
```

source/features/global-search-filters.tsx
```diff
@@ -0,0 +1,49 @@
...
+			item.classList.add('selected');
+			item.prepend(<XIcon className="float-right"/>);
+		} else {
+			query.add(filter);
+		}
+
+		items.push(<li>{item}</li>);
+	}
+
+	select('#js-pjax-container .menu')!.nextElementSibling!.after(
```
fregante, line 32: This adds it in place of the languages; it should be added below it to avoid moving the UI too much. Also it's best to use a better selector instead of `nextElementSibling`

fregante (approved):
Untested but it looks correct other than this

source/features/search-filters.tsx
kidonng: For some reason GitHub doesn't show code search results when using `user:@me`:

<details>
<summary><code>user:@me</code></summary>

image

</details>

<details>
<summary><code>user:username</code></summary>

image

</details>

source/features/global-search-filters.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+import features from '.';
+import SearchQuery from '../github-helpers/search-query';
+import {getUsername} from '../github-helpers';
+
+function init(): void {
+	const filters = [
+		['Forks', 'fork:true'],
+		['Private', 'is:private'],
+		['Yours', `user:${getUsername()}`]
```
yakov116, line 14:
```suggestion
		['Yours', 'user:' + getUsername()]
```

source/features/global-search-filters.tsx
```diff
@@ -0,0 +1,49 @@
...
+			item.classList.add('selected');
+			item.prepend(<XIcon className="float-right"/>);
+		} else {
+			query.add(filter);
+		}
+
+		items.push(<li>{item}</li>);
+	}
+
+	select('#js-pjax-container .menu ~ .border')!.after(
```
yakov116, line 32: Could we make this not have to wait for domReady by using element ready?

source/features/global-search-filters.tsx
fregante: Don't let Sindre see this 😃

yakov116:
@kidonng no pressure. But there are just 2 more things to get this merged

kidonng:
Sorry, will visit these today or tomorrow.

kidonng:
Resolved reviews and added more test URLs. I also added a new **Authored** filter `author:@me`, which is useful for e.g. search commits or conversations authored by yourself.

yakov116 (approved):

yakov116:
Thanks @kidonng

641210 (commented):
****

PR #4099, opened 2021-03-14
Add `delete-review-comments-faster` feature

### LINKED ISSUE #3605: Delete review comments while trying to edit them
1. https://github.com/sindresorhus/refined-github/pull/3604/files
2. Go to edit a comment
3. Change your mind, you want to delete it
4. 5 clicks later, you can delete it

I think we can add a red-text "Delete" button to the left to reduce the clicks to just 2 (click + confirm)

This can be done using `selector-observer`

cheap-glitch:
Closes #3605 

## Test URLs
- this PR's discussion thread
- this PR's files

## Screenshot
image


source/features/quick-review-comment-deletion.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'dom-chef';
+import select from 'select-dom';
+import delegate from 'delegate-it';
+import {observe} from 'selector-observer';
+import * as pageDetect from 'github-url-detection';
+
+import features from '.';
+
```
cheap-glitch, line 8: This is a test comment.

cheap-glitch: This is another test comment.

cheap-glitch: Test

source/features/quick-review-comment-deletion.tsx
```diff
@@ -0,0 +1,39 @@
+import React from 'dom-chef';
+import select from 'select-dom';
+import delegate from 'delegate-it';
+import {observe} from 'selector-observer';
+import * as pageDetect from 'github-url-detection';
+
```
cheap-glitch, line 6: This is a review comment.

source/features/delete-review-comments-faster.tsx
```diff
@@ -0,0 +1,52 @@
...
+		});
+		return;
+	}
+
+	deleteComment(comment);
+}
+
+function init(): void {
+	delegate(document, '.rgh-review-comment-delete-button', 'click', onButtonClick);
+	observe('.review-comment > .unminimized-comment .btn-primary[type="submit"]', {
```
cheap-glitch, line 29: For some reason this doesn't work:
```suggestion
	observe('.review-comment > .unminimized-comment .btn-primary[type="submit"]:not(.rhg-delete-button-added)', {
```

source/features/delete-review-comments-faster.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+function deleteComment(comment: HTMLElement): void {
+	select('.dropdown-menu .js-comment-delete > button', comment)!.click();
+}
+
+function onButtonClick(event: delegate.Event): void {
+	const comment = (event.target as HTMLElement).closest<HTMLElement>('.js-comment')!;
+	const dropdownFragment = select('include-fragment.SelectMenu-loading', comment);
+	if (dropdownFragment) {
+		select('.timeline-comment-actions > details:last-of-type', comment)!.dispatchEvent(new Event('mouseover'));
```
fregante, line 17: I think we have a helper function to do exactly this

source/features/delete-review-comments-faster.tsx
```diff
@@ -0,0 +1,52 @@
...
+import * as pageDetect from 'github-url-detection';
+
+import features from '.';
+
+function deleteComment(comment: HTMLElement): void {
+	select('.dropdown-menu .js-comment-delete > button', comment)!.click();
+}
+
+function onButtonClick(event: delegate.Event): void {
+	const comment = (event.target as HTMLElement).closest<HTMLElement>('.js-comment')!;
```
fregante, line 14: When using `delegate`, always use `event.delegateTarget`, it's already the right type

source/features/delete-review-comments-faster.tsx
```diff
@@ -0,0 +1,52 @@
...
+	const dropdownFragment = select('include-fragment.SelectMenu-loading', comment);
+	if (dropdownFragment) {
+		select('.timeline-comment-actions > details:last-of-type', comment)!.dispatchEvent(new Event('mouseover'));
+		dropdownFragment.addEventListener('load', () => {
+			deleteComment(comment);
+		});
+		return;
```
fregante, line 21: If the helper function doesn't already take care of this, then:
```suggestion
		await oneEvent(dropdownFragment, 'load)';
```

source/features/delete-review-comments-faster.tsx
fregante: Try again  😁 I don't see why it wouldn't work, the selector appears to be valid

fregante (commented):
In the PR comment body, you can always attach a simple screenshot/gif, it doesn't have to be final, it's just so the current style can be reviewed without having to check out the PR

source/features/delete-review-comments-faster.tsx
cheap-glitch: I thought so but couldn't find one. Are you referring to this maybe? https://github.com/sindresorhus/refined-github/pull/3749#discussion_r532286271

source/features/delete-review-comments-faster.tsx
cheap-glitch: Got it, it was me who forgot to `onetime()` the `init` :man_facepalming:

fregante:
> image

Oof, I'm not so sure about this anymore. If we add a delete button, we definitely shouldn't move the cancel button to the left side

cheap-glitch:
Yeah I agree that the UX is rather perilous. We could move it all the way to the left of the "Cancel" button + make it smaller (maybe reduce it to a trash icon) so it's harder to hit by mistake.

source/features/delete-review-comments-faster.tsx
fregante: Yes

cheap-glitch: We could extract the logic into a more generic helper and use it in both features.

fregante: If it's similar enough yes. If it requires conditions probably not

fregante:
In what situations do we see a Cancel button? I think:

- Editing a comment
- Editing a review comment

"Cancel" should appear in the same place everywhere, so probably:

```
_________________________________________________
| Editing a review comment                      |
|                                               |
|                                               |
|_______________________________________________|
  [Delete]            [Cancel] [Update comment]
```


```
_________________________________________________
| Editing a regular comment                     |
|                                               |
|                                               |
|_______________________________________________|
                      [Cancel] [Update comment]
```

This partially undoes `safer-destructive-actions`

We also need to consider the "Close" and "Close with comment" buttons that should remain on the left.


```
_________________________________________________
| Sending a new regular comment                 |
|                                               |
|                                               |
|_______________________________________________|
  [Close with comment]                [Comment]
```

cheap-glitch:
I'm -0 on moving the cancel buttons, I think it's better if they stay on the left. Maybe I'm missing something but I don't think we should change every single layout to accommodate for one specific case.

Just to be clear, what I'm proposing is:
```text
_________________________________________________
| Editing a review comment                      |
|                                               |
|                                               |
|_______________________________________________|
  [Delete][Cancel]             [Update comment]
```

This is in keeping with `safer-destructive-actions` since the most destructive action ends up all the way to the left.

FloEdelmann:
What about making the "Delete" button an [icon-only button](https://primer.style/css/components/buttons#button-with-icons) (with the [trash icon](https://octicons-primer.vercel.app/octicons/trash-16))?

```
_________________________________________________
| Editing a review comment                      |
|                                               |
|                                               |
|_______________________________________________|
[Cancel]                    [🗑] [Update comment]
```

Then clicking "Delete" accidentally is more difficult, and it can't be confused (via position or appearance) with the "Cancel" button.

cheap-glitch:
> What about making the "Delete" button an icon-only button (with the trash icon)?

> We could move the "Delete" button all the way to the left of the "Cancel" button, and make it smaller (maybe even reduce it to something like a trash icon) so it's harder to click by mistake.

Great minds... :wink: 

What do you think about:
```text
_________________________________________________
| Editing a review comment                      |
|                                               |
|                                               |
|_______________________________________________|
[🗑][Cancel]                     [Update comment]
```

It's even harder to hit by mistake and it follows the logic of putting the destructive actions as far away from the submit buttons as possible. (And since the delete button is the most destructive action it should be on the far left).

FloEdelmann:
When you click "cancel", you are expecting a confirm dialog, so you'll accept it without reading.

When the delete button and the cancel button are close to each other, it's easier to confuse them, which might lead to some accidentally deleted comments instead of cancelled edits.

cheap-glitch:
> When you click "cancel", you are expecting a confirm dialog, so you'll accept it without reading

Afaik there is no confirm dialog when you cancel editing a comment.

> When the delete button and the cancel button are close to each other, it's easier to confuse them

I agree, but imo having one button reduced to an icon would prevent that. I think it's (slightly) less dangerous than having it near the submit button. It's all just speculation though, I'm not a UX designer and I don't have any real-world data :smiley:

FloEdelmann:
> Afaik there is no confirm dialog when you cancel editing a comment.

There is, but only after actually changing the textarea content.

> It's all just speculation though, I'm not a UX designer and I don't have any real-world data :smiley:

Same for me :sweat_smile:

fregante:
You're both right and I can't think of a position that:

- doesn't move the cancel button: people expect to find it there and end up clicking on delete
- doesn't appear next to Update comment: it goes against the spirit of `safer-destructive-actions`

Maybe I have 2 more suggestions to add to the mix:

- center it between cancel and update, via flex or similar, red **text**
- delete the comment by emptying the field: Clicking "update" after emptying the field doesn't make any changes. It could offer to delete the comment in a generic but somewhat obvious way but:
	- not discoverable
	- requires at least 2 more commands: `cmd+a` + `delete`
	- probably requires changing the Update button to a Delete button, which is a bit of code

source/features/delete-review-comments-faster.tsx
fregante: Alternatively, can you check if we can use _any_ authenticity token to delete it? This way we can just replicate the form without having to wait for it.

Note: The form should still probably be created inline and managed by GitHub so we get:

- native error handling
- native prompt, which I think it's missing altogether here and that's not ok

fregante:
Also this doesn't necessarily need to be restricted to PR comments, however it's probably not as useful outside reviews (which is when one might decide to delete/move comments before submitting a review)

source/features/delete-review-comments-faster.tsx
```diff
@@ -0,0 +1,43 @@
+import React from 'dom-chef';
+import select from 'select-dom';
+import onetime from 'onetime';
+import oneEvent from 'one-event';
+import delegate from 'delegate-it';
+import {observe} from 'selector-observer';
+import * as pageDetect from 'github-url-detection';
+
```
cheap-glitch, line 8:
```suggestion
This is a test suggestion
```

source/features/delete-review-comments-faster.tsx
cheap-glitch: > native prompt, which I think it's missing altogether here and that's not ok

I'm seeing a confirmation dialog when deleting though?

cheap-glitch:
Delete button on the left (but not where "Cancel" is expected to be), reduced to an icon to minimized misclicks:
Screenshot_2021-04-07 Add `delete-review-comments-faster` feature by cheap-glitch · Pull Request #4099 · sindresorhus refin

Delete button in the center, as suggested by @fregante:
Screenshot_2021-04-07 Add `delete-review-comments-faster` feature by cheap-glitch · Pull Request #4099 · sindresorhus refin (1)

I think the first option looks better, but the second one might be the best in terms of UX. @yakov116 @FloEdelmann thoughts?

yakov116:
Option 2

FloEdelmann:
> I think the first option looks better, but the second one might be the best in terms of UX.

Totally agree.

source/features/delete-review-comments-faster.css
```diff
@@ -0,0 +1,3 @@
+.rgh-delete-button-form::before, .rgh-delete-button-form::after {
+	content: none !important;
+}
```
cheap-glitch, line 3: Could not find a GH class to remove the `::before` and `::after`. (They get in the way of the flexbox layout)

source/features/delete-review-comments-faster.tsx
```diff
@@ -0,0 +1,46 @@
...
+		select('.timeline-comment-actions > details:last-of-type', comment)!.dispatchEvent(new Event('mouseover'));
+		await oneEvent(dropdownFragment, 'load');
+	}
+
+	select('.dropdown-menu .js-comment-delete > button', comment)!.click();
+}
+
+function init(): void {
+	delegate(document, '.rgh-review-comment-delete-button', 'click', onButtonClick);
+	observe('.review-comment > .unminimized-comment form:not(.js-single-suggested-change-form) .btn-primary[type="submit"]:not(.rgh-delete-button-added)', {
```
fregante, line 26: Lets use the other pattern for `observe` and `delegate` (the `deinit` array)

source/features/delete-review-comments-faster.tsx
```diff
@@ -0,0 +1,46 @@
...
+		add(submitButton) {
```
fregante, line 27: We should probably start moving these functions to the top level instead of keeping them inline here

source/features/delete-review-comments-faster.tsx
fregante: True. The rest of the review still applies though

source/features/delete-review-comments-faster.tsx
cheap-glitch: So if I understood correctly it should replace the `onetime(init)`?

fregante:
Sorry for the change again, but I think this looks better,  which is important since this appears on a lot of forms

> Screenshot_2021-04-07 Add delete-review-comments-faster feature by cheap-glitch · Pull Request #4099 · sindresorhus refin

source/features/delete-review-comments-faster.tsx
```diff
@@ -0,0 +1,45 @@
+import React from 'dom-chef';
+import select from 'select-dom';
+import delegate from 'delegate-it';
+import {observe} from 'selector-observer';
+import * as pageDetect from 'github-url-detection';
+import {TrashIcon} from '@primer/octicons-react';
```
yakov116, line 6:
```suggestion
import {observe} from 'selector-observer';
import {TrashIcon} from '@primer/octicons-react';
import * as pageDetect from 'github-url-detection';
```

source/features/delete-review-comments-faster.tsx
```diff
@@ -0,0 +1,45 @@
...
+	deinit.push(listener.destroy, observer.abort);
+}
+
+void features.add(__filebasename, {
+	include: [
+		pageDetect.isPRConversation,
+		pageDetect.isPRFiles
+	],
+	awaitDomReady: false,
+	init,
```
yakov116, line 43: Don't we need a `onetime` here?

cheap-glitch: https://github.com/sindresorhus/refined-github/pull/4099#discussion_r614214617

yakov116: I still don't see why it should not need a one time...

On ajax the observe will be added again. Am I wrong?

fregante: Because we have a `deinit` array https://github.com/sindresorhus/refined-github/pull/4072#issuecomment-792402458

source/features/delete-review-comments-faster.tsx
fregante: I checked. The delete form’s authenticity token cannot be replaced with another token, so we can't recreate the form manually.

fregante:
I finally tested this. The delay introduced by the context menu’s delayed loading makes it look like it's not working for a bit:

gif

You can see that I can click multiple times before it starts loading (at my first trial I thought it wasn't loading at all because it took a second or more)

I think we have 2 options:

- add a loading icon in the button (😬)
- use the API to delete the button and then manually update the UI

The first one is still slow and still bound to errors (we should display an alert if we catch an error) and the second one, while easy on the surface, requires a way to correctly update the whole review UI.

It's unfortunate that we need that exact "delete" form to perform the action in the DOM. I tried using a different "authenticity token" and GitHub responded with a 422.

cheap-glitch:
> The delay introduced by the context menu’s delayed loading makes it look like it's not working for a bit

Since this only happens when using the `edit-comments-faster` button that bypasses the dropdown, how about preloading the dropdown when that button is clicked?

fregante:
Good compromise

source/github-events/on-fragment-load.ts
```diff
@@ -0,0 +1,10 @@
+import oneEvent from 'one-event';
+
+export default async function onFragmentLoad(fragment: HTMLElement | undefined, clickZone: HTMLElement): Promise<void> {
+	if (!fragment) {
+		return;
+	}
```
fregante, line 6: `onFragmentLoad` makes it sound like it just waits for a fragment element to load, but actually it triggers its load through a behavior defined by the `details-menu` element.

In short, I'd like this method to be called `loadDetailsMenu` or something  like that and just receive the `details-menu` element. At that point, either the fragment and trigger the load, or just return

fregante (approved):
Works great! There's only one last minor change I'd like to see

source/features/delete-review-comments-faster.tsx
```diff
@@ -0,0 +1,48 @@
...
+import {TrashIcon} from '@primer/octicons-react';
+import * as pageDetect from 'github-url-detection';
+
+import features from '.';
+import onFragmentLoad from '../github-events/on-fragment-load';
+
+const deinit: VoidFunction[] = [];
+
+async function onButtonClick({delegateTarget: button}: delegate.Event): Promise<void> {
+	select('.dropdown-menu .js-comment-delete > button', button.closest<HTMLElement>('.js-comment')!)!.click();
```
fregante, line 14: It's best not to combine `select` with other DOM traversal because the order makes it hard to follow.

Starts from button, then goes right to `.js-comment` and then `select`’s selector is on the left.

Instead:
```suggestion
	button
		  .closest('.js-comment')!
		  .querySelector('.dropdown-menu .js-comment-delete > button')!
		  .click();
```

Also `<HTMLElement>` isn't necessary here and in the other `.closest` call below

source/refined-github.ts
```diff
@@ -213,6 +213,7 @@ import './features/rgh-sponsor-button';
 import './features/rgh-linkify-features';
 import './features/collapse-markdown-sections';
 import './features/conversation-activity-filter';
+import './features/delete-review-comments-faster';
```
yakov116, line 216: Can you move this to the bottom

yakov116:
🎉

bl-ue:
When is this feature going to come out?  I often want to delete un-sent review comments but it's rather tedious 🤔

yakov116:
Tomorrow

PR #1234, opened 2018-03-30
Hide overlapping tooltips

brandon93s:
Hides the tooltip on the following buttons:

image

Keeps the tooltip on the expand / collapse glyph since it contains a helpful tip:

image

---
Closes #1212

PR #4100, opened 2021-03-14
Extend `ci-link` to PR

### LINKED ISSUE #4086: Show `ci-link` for PR next to it's title/number
Similarly to repo CI status:
https://github.com/sindresorhus/refined-github/blob/main/source/features/ci-link.tsx
image

I propose adding a CI status next to PR name/title/number.
Reasons:
1. On long PRs - you need to scroll A LOT to see CI status ([for example](https://github.com/sindresorhus/refined-github/pull/4023) )
2. When you're in another tab - like changes - you need to switch tabs to know CI status. So you need to switch tab to conversations + scroll to bottom.
3. When you're already scrolled in changes view - you need to *lose context* to know CI status. Like half way thru review, or [direct link](https://github.com/sindresorhus/refined-github/pull/3810/files#diff-5a831ea67cf5cf8703b0de46901ab25bd191f56b320053be9332d9a3b0d01d15R224)

kidonng:
Resolve #4086

Meta unchanged though

## Test URLs

#4098

## Screenshot

image

image


kidonng:
Dumb me 😅

source/features/sync-pr-commit-title.tsx
```diff
@@ -16,10 +17,6 @@ function getCommitTitleField(): HTMLInputElement | undefined {
 	return select<HTMLInputElement>(`${mergeFormSelector} #merge_title_field`);
 }
 
-function getPRNumber(): string {
```
yakov116, line 19: I think you should just export it from here.

Also this is not domReady safe

kidonng: > I think you should just export it from here.

Well we may need this in the future. It doesn't hurt.

> this is not domReady safe

What!? This is in the HTML!

image

yakov116: use `getConversationNumber()`

source/features/ci-link.tsx
```diff
@@ -5,31 +5,51 @@ import * as pageDetect from 'github-url-detection';
...
 ));
 
-async function init(): Promise<false | void> {
-	const icon = await getIcon() as HTMLElement | undefined;
+const getPRIcon = onetime(async () => fetchDom(
+	buildRepoURL('pull', getConversationNumber()!, 'commits'), '.js-commits-list-item:last-of-type .commit-build-statuses'
+));
+
+async function initRepo(): Promise<false | void> {
+	const icon = await getRepoIcon() as HTMLElement | undefined;
```
fregante, line 23: Let's try to move these types to the `get*Icon` functions so we avoid assertions

source/features/ci-link.tsx
```diff
@@ -5,31 +5,51 @@ import * as pageDetect from 'github-url-detection';
...
+	}
+
+	icon.classList.add('rgh-ci-link');
+	if (onetime.callCount(getPRIcon) > 1) {
+		icon.style.animation = 'none';
+	}
+
+	// Append to PR title and floating header
+	select('.gh-header-title')!.append(icon);
+	select('.gh-header-number')!.after(icon.cloneNode(true));
```
fregante, line 50: Do the icons stay up there even after new comments, label changes, title changes?

source/features/sync-pr-commit-title.tsx
fregante: @yakov116 I'm confused, why do we have this function at all? `sync-pr-commit-title` runs on PRs too, so the number is in the URL

yakov116: That one needed the `#` so someone added it. I noticed it a while ago, but just left it.

fregante: @yakov116 you stopped the blame too early. That line of code is ancient 😃 https://github.com/sindresorhus/refined-github/commit/ed2d06f0b0c441f16a7b444f3a8fc55e6f986924

source/features/ci-link.tsx
kidonng: Nope... Guess we need selector-observer

source/features/ci-link.tsx
```diff
@@ -1,35 +1,67 @@
...
 }
 
+async function initPR(): Promise<false | void> {
+	const icon = await getPRIcon();
+	if (!icon) {
+		return false;
+	}
+
+	icon.classList.add('rgh-ci-link');
+	if (onetime.callCount(getPRIcon) > 1) {
```
kidonng, line 45: This doesn't work... I have some ideas but none of them looks good :(

source/features/sync-pr-commit-title.tsx
yakov116: Ok I'll excuse you this time 😉

source/features/ci-link.tsx
yakov116:
```js
	init() {
		observeElement(select('#partial-discussion-header')!.parentElement!, init);
	}
```


This should be enough no?

fregante:
Position, size and alignment need to be reviewed in both cases
Screen Shot

Screen Shot 1

source/features/ci-link.tsx
fregante: Hopefully 👆

fregante:
Try placing it in the `#1234` element so they always wrap together and we avoid orphans:

Screen Shot 3

kidonng:
- Updated size to match with the text
- Grouped the icon with PR number

fregante:
> * Updated size to match with the text

Looks good


Screen Shot 2021-04-03 at 22 40 39

Screen Shot 2021-04-03 at 22 40 45


> * Grouped the icon with PR number

Doesn't actually follow its purpose though

Screen Shot 2021-04-03 at 22 41 16


---

- [ ] Readme needs to be updated to mention it also works on PRs

kidonng:
> Doesn't actually follow its purpose though

Works on my machine™️

https://user-images.githubusercontent.com/44045911/113847138-bc66b380-97c9-11eb-83f0-d33b07050eda.mov

source/features/ci-link.tsx
```diff
@@ -1,35 +1,77 @@
...
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
 import fetchDom from '../helpers/fetch-dom';
-import {buildRepoURL} from '../github-helpers';
+import {buildRepoURL, getConversationNumber} from '../github-helpers';
 
 // Look for the CI icon in the latest 2 days of commits #2990
-const getIcon = onetime(async () => fetchDom(
+const getRepoIcon = onetime(async () => fetchDom<HTMLElement>(
```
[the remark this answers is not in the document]
kidonng: This means we can't use `onetime(func).callCount` anymore

source/features/ci-link.tsx
```diff
@@ -1,35 +1,74 @@
...
+	if (onetime.callCount(getRepoIcon) > 1) {
 		icon.style.animation = 'none';
 	}
 
 	// Append to title (aware of forks and private repos)
 	select('[itemprop="name"]')!.parentElement!.append(icon);
 }
 
+async function initPR(): Promise<false | void> {
+	const icon = await getPRIcon() as HTMLElement | undefined;
```
kidonng, line 44: I can't inline this type in `getPRIcon` because eslint thinks `select` returns `HTMLElement` by default while typescript sees it `Element`.

source/features/ci-link.tsx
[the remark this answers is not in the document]
fregante: I forgot about that

source/features/ci-link.tsx
fregante: Must be a TS bug. Fixed by using `<HTMLElement>` on the select itself

fregante:
> Works on my machine™️ (Chrome 89.0.4389.114 on macOS)

Not For Me © 1969-2021 on Firefox (MPL) and Safari (Jony I.V.E.)

source/features/ci-link.tsx
```diff
@@ -23,8 +23,9 @@ const getPRIcon = onetime(async () => {
 		document :
 		await fetchDom(buildRepoURL('pull', getConversationNumber()!, 'commits'));
 
-	// TS bug requires a generic here 🤷‍♂️
-	return select<HTMLElement>('.js-commits-list-item:last-of-type .commit-build-statuses', base);
+	// TS bug requires this 🤷‍♂️
```
yakov116, line 26:
```suggestion
	// TS bug does not allow us to directly return this 🤷‍♂️
```

Maybe?

kidonng:
> Not For Me © 1969-2021 on Firefox (MPL) and Safari (Jony I.V.E.)

I guess they treat flexbox differently. Try changing `.f1-light` from `span` to `div`, it wraps in Safari but not in Chrome.

yakov116:
The X icon still needs some CSS help

image

source/features/ci-link.tsx
```diff
@@ -22,7 +22,7 @@ const getPRIcon = onetime(async () => {
 		await fetchDom(buildRepoURL('pull', getConversationNumber()!, 'commits'));
 
 	// TS bug does not allow us to directly return this 🤷‍♂️
-	const icon = select('.js-commits-list-item:last-of-type .commit-build-statuses', base);
+	const icon = select.last('.js-commits-list-item:last-of-type .commit-build-statuses', base);
```
fregante, line 25: Doesn't :last-of-type already do this? If not, is it needed anymore?

fregante: Also does the new fetchDom type let us use `return select` now?

source/features/ci-link.tsx
```diff
@@ -5,24 +5,57 @@ import * as pageDetect from 'github-url-detection';
...
+		return false;
+	}
+
+	icon.classList.add('rgh-ci-link', 'ml-2');
+	if (onetime.callCount(getPRIcon) > 1) {
+		icon.style.animation = 'none';
+	}
+
+	const headers = select.all(':is(.gh-header-title .f1-light, .js-sticky h1):not(.rgh-ci-link-heading)');
+
```
fregante, line 41:
```suggestion
```

source/features/ci-link.tsx
yakov116: I think its still needed. (Can't check right now)

>The commits can span over a few days, we need the last one on the list

We need the last group and the last one on the list.

source/features/ci-link.tsx
```diff
@@ -38,5 +70,14 @@ void features.add(__filebasename, {
...
-	init
+	init: initRepo
+}, {
+	include: [
+		pageDetect.isPR
+	],
+	additionalListeners: [
+		onConversationHeaderUpdate
+	],
+	awaitDomReady: pageDetect.isPRCommitList(),
```
yakov116, line 81: TS help please

source/features/ci-link.tsx
yakov116: > Also does the new fetchDom type let us use `return select` now?

No it does not

source/features/ci-link.tsx
fregante: Can't be done. `pageDetect.isPRCommitList` would only be called on the first page load, but not on Ajax. You'll have to use `false` here and if `domReady` is needed conditionally, add `if (pageDetect.isPRCommitList()) await domLoaded` in the init function

zarnisoe2353 (approved):

zarnisoe2353 (commented):
4100

zarnisoe2353 (commented):
3826

yakov116:
@zarnisoe2353 STOP

yakov116:
This PR is moot until after https://github.com/sindresorhus/refined-github/issues/4260 is fixed

yakov116:
@kidonng can you finish this up? I would love to see it merged?

fregante:
Closing this PR for now. Feel free to create a new PR after addressing all the reviews and squashing the commits

PR #4105, opened 2021-03-15
Fix alignment of `show-names` in review comments

cheap-glitch:
## Test URLs
Any PR with review comments, e.g. https://github.com/sindresorhus/refined-github/pull/4023

## Screenshots

<table>
<tr>
	<th>Before
	<th>After
<tr>
	<td>
		
	<td>
		
</table>

### DIFF
--- a/source/features/show-names.tsx
+++ b/source/features/show-names.tsx
@@ -56,7 +56,7 @@ async function init(): Promise<false | void> {
 				insertionPoint.after(
 					' (',
 					<bdo className="css-truncate">
-						<span className="css-truncate-target v-align-middle" style={{maxWidth: '200px', transform: 'translateY(1px)'}}>
+						<span className="css-truncate-target" style={{maxWidth: '200px'}}>
 							{names[userKey].name}
 						</span>
 					</bdo>,
--- a/source/features/show-names.tsx
+++ b/source/features/show-names.tsx
@@ -56,7 +56,7 @@ async function init(): Promise<false | void> {
 				insertionPoint.after(
 					' (',
 					<bdo className="css-truncate">
-						<span className="css-truncate-target v-align-middle" style={{maxWidth: '200px'}}>
+						<span className="css-truncate-target v-align-middle" style={{maxWidth: '200px', transform: 'translateY(1px)'}}>
 							{names[userKey].name}
 						</span>
 					</bdo>,
--- a/source/features/show-names.tsx
+++ b/source/features/show-names.tsx
@@ -56,7 +56,7 @@ async function init(): Promise<false | void> {
 				insertionPoint.after(
 					' (',
 					<bdo className="css-truncate">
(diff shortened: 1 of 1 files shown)


fregante (changes requested):
> Timeline event:

This is worse than before. I can't quite tell in the other screenshots since they're cut too close

source/features/show-names.tsx
```diff
@@ -56,7 +56,7 @@ async function init(): Promise<false | void> {
 				insertionPoint.after(
 					' (',
 					<bdo className="css-truncate">
-						<span className="css-truncate-target" style={{maxWidth: '200px'}}>
+						<span className="css-truncate-target v-align-middle" style={{maxWidth: '200px'}}>
```
fregante, line 59: The right alignment for text is `baseline`

cheap-glitch: That's what I thought too, but it's even worse. I tried every `vertical-align` value and `middle` is the ~best~ least worst compromise. We lose a few pixels of alignment in the comment headers and timeline events, but we get much better alignment in review headers.

We could also try to add some padding or a `transform: translate()`

fregante: No, baseline is the correct one and it looks fine where I tried it. What's pulling the element out of the alignment is the avatar. Transform would just be tape to hold it together temporarily

fregante: Baseline on review comment

Screen Shot

fregante: Baseline on review request (zero change, it's the default)

Screen Shot

fregante: Baseline on regular comments (zero change, it's the default)

Screen Shot

yakov116: I see the issue on review comments

image

fregante: My screenshots are with `baseline` applied. I also see the issue on `main`

cheap-glitch: > No, baseline is the correct one and it looks fine where I tried it.

On what element did you apply `vertical-align`? I can't get it to align properly :thinking:

fregante: On this same element. However it works correctly on Safari, but in Chrome it doesn't, `overflow:hidden` throws it off. Investigating

fregante: `display: inline-block` on `bdo` fixes it everywhere in every browser

source/features/show-names.tsx
```diff
@@ -55,7 +55,7 @@ async function init(): Promise<false | void> {
 				const insertionPoint = usernameElement.parentElement!.tagName === 'STRONG' ? usernameElement.parentElement! : usernameElement;
 				insertionPoint.after(
 					' (',
-					<bdo className="css-truncate">
+					<bdo className="css-truncate" style={{display: 'inline-block'}}>
```
fregante, line 58:
```suggestion
					<bdo className="css-truncate d-inline-block">
```

PR #4106, opened 2021-03-15
Fix `preview-hidden-comments` for resolved and outdated comments

### LINKED ISSUE #3647: `preview-hidden-comments` no longer works for most types

cheap-glitch:
Fixes #3647 

## Test URLs
https://github.com/sindresorhus/refined-github/issues/3647

## Screenshot

<table>
<tr>
	<th>Before
	<th>After
<tr>
	<td>
		
	<td>
		
</table>

### DIFF
--- a/source/features/preview-hidden-comments.tsx
+++ b/source/features/preview-hidden-comments.tsx
@@ -6,7 +6,7 @@ import * as pageDetect from 'github-url-detection';
 import features from '.';
 import {upperCaseFirst} from '../github-helpers';
 
-const allowedReasons = new Set(['resolved', 'outdated', 'off-topic']);
+const forbiddenReasons = new Set(['spam', 'duplicate', 'disruptive content']);
 
 const init = (): void => {
 	for (const details of select.all('.minimized-comment:not(.d-none) > details:not(.rgh-preview-hidden-comments)')) {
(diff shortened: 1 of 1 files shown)


source/features/preview-hidden-comments.tsx
```diff
@@ -23,13 +23,13 @@ const init = (): void => {
...
 		const reason = /was marked as ([^.]+)/.exec(header.textContent!)?.[1] ?? '';
-		if (!allowedReasons.has(reason)) {
+		if (forbiddenReasons.has(reason)) {
 			continue;
 		}
 
 		header.append(
 			<span className="Details-content--open">{header.firstChild}</span>,
-			<span className="Details-content--closed">{`${upperCaseFirst(reason)} — ${commentText}`}</span>
+			<span className="Details-content--closed">{(reason.length > 0 ? upperCaseFirst(reason) + ' — ' : '') + commentText}</span>
```
cheap-glitch, line 32: Should be give up on showing the reason altogether? Since only off-topic comments can display it anyway.

fregante:
This is not what the feature was meant to do; It should preview the comments, not add the reason where it's missing. Refer to the existing screenshots

fregante:
I think the logic should be rewritten since the reason can't be parsed anymore. It's as simple as `if message is either off-top or minimized, preview it`

Also the preview should be:


```
Off-topic — This message is blah blah
Hidden — This message is blah blah
```

source/features/preview-hidden-comments.tsx
```diff
@@ -22,14 +22,14 @@ const init = (): void => {
 			'.discussion-item-icon  + div' // Review Comments
 		], details)!;
 
-		const reason = /was marked as ([^.]+)/.exec(header.textContent!)?.[1] ?? '';
-		if (!allowedReasons.has(reason)) {
+		const reason = /was marked as ([^.]+)/.exec(header.textContent!)?.[1] ?? (header.textContent!.trim().endsWith('has been minimized.') ? 'duplicate' : '');
```
fregante, line 25: Probably

```suggestion
		const reason = /minimized|off-topic/.exec(header.textContent!)?.[0];
```

PR #4108, opened 2021-03-16
Abbreviate `bugs-tab`  and `releases-tab` count

### LINKED ISSUE #4107: Use 1k instead of 1000 in `bugs-tab`
The bugs count should format like the issues count.
e

yakov116:
Closes #4107 

## Test URLs
https://github.com/microsoft/vscode

## Screenshot
Before
image

After
image

Question should we replace `.0`?

Without the .0

image

### DIFF
--- a/source/features/bugs-tab.tsx
+++ b/source/features/bugs-tab.tsx
@@ -93,9 +93,7 @@ async function init(): Promise<void | false> {
 
 	// Update bugs count
 	try {
-		bugsCounter.textContent = abbreviateNumber(await countPromise, 1, {padding: false})
-			// Matches GitHub lowercase `k` style
-			.toLowerCase();
+		bugsCounter.textContent = abbreviateNumber(await countPromise, 1, {padding: false}).toLowerCase();
 	} catch (error: unknown) {
 		bugsCounter.remove();
 		features.error(__filebasename, error);
--- a/source/features/bugs-tab.tsx
+++ b/source/features/bugs-tab.tsx
@@ -9,7 +9,7 @@ import features from '.';
 import * as api from '../github-helpers/api';
 import {getRepo} from '../github-helpers';
 import SearchQuery from '../github-helpers/search-query';
-import issueNumberFormat from '../helpers/issue-number-format';
+import abbreviateNumber from '../helpers/abbreviate-number';
 
 const countBugs = cache.function(async (): Promise<number> => {
 	const {search} = await api.v4(`
@@ -93,7 +93,7 @@ async function init(): Promise<void | false> {
 
 	// Update bugs count
 	try {
-		bugsCounter.textContent = issueNumberFormat(await countPromise);
+		bugsCounter.textContent = abbreviateNumber(await countPromise);
 	} catch (error: unknown) {
 		bugsCounter.remove();
 		features.error(__filebasename, error);
--- a/source/features/bugs-tab.tsx
+++ b/source/features/bugs-tab.tsx
@@ -4,12 +4,12 @@ import select from 'select-dom';
 import {BugIcon} from '@primer/octicons-react';
 import elementReady from 'element-ready';
 import * as pageDetect from 'github-url-detection';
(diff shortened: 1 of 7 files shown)


fregante (approved):
Neat!

yakov116:
> Question should we replace `.0`?

?

source/features/bugs-tab.tsx
```diff
@@ -93,7 +93,11 @@ async function init(): Promise<void | false> {
 
 	// Update bugs count
 	try {
-		bugsCounter.textContent = abbreviateNumber(await countPromise);
+		bugsCounter.textContent = abbreviateNumber(await countPromise)
+			// 1.0K to 1k
+			.replace('.0K', 'k')
```
kidonng, line 98: Maybe just `.replace(/(\.0)?K/, 'k')` but this is clearer so fine

fregante:
https://github.com/moh3n9595/js-abbreviation-number#usage

yakov116:
> moh3n9595/js-abbreviation-number#usage

I saw that

Using `abbreviateNumber(number, 0)` will round everything to the nearest 1000.

For example `abbreviateNumber(1500, 0) => 2K` that is not what we want.

fregante:
Ew… we can either find an alternative (micro) module or import it, adjust it, re-export it from a file under the helpers, so the same can be used in the other feature that uses `js-abbreviation-number`

yakov116:
I will cleanup the lock later.

Should should I make a helper called `issueNumberFormat`?

source/features/bugs-tab.tsx
```diff
@@ -93,7 +93,9 @@ async function init(): Promise<void | false> {
 
 	// Update bugs count
 	try {
-		bugsCounter.textContent = numberFormatter.format(await countPromise);
+		bugsCounter.textContent = abbreviateNumber(await countPromise, 1, {padding: false})
+			// Matches GitHub lowercase `k` style
+			.toLowerCase();
```
fregante, line 98:
```suggestion
		bugsCounter.textContent = abbreviateNumber(await countPromise, 1, {
			padding: false,
			symbols: ['', 'k']
		});
```

fregante: Nevermind

fregante:
```suggestion
		bugsCounter.textContent = abbreviateNumber(await countPromise, 1, {padding: false}).toLowerCase();
```

yakov116: DUHH I was fighting with the update for about 35 min before.......

fregante: I can imagine 😅

webpack.config.ts
```diff
@@ -116,7 +116,6 @@ const config: Configuration = {
 			),
 
 			__filebasename: webpack.DefinePlugin.runtimeValue(
-				// @ts-expect-error
```
fregante, line 119: Let's keep unrelated package-lock changes for the next lint. I added some lines in #4111 because they are macOS-specific and always get added by `npm i`, just for macOS users :(

yakov116: I am doing it now. I was just so mad I wanted to just commit and fix it up.

source/helpers/issue-number-format.ts
```diff
@@ -0,0 +1,5 @@
+import {abbreviateNumber} from 'js-abbreviation-number';
+
+export default function issueNumberFormat(count: number): string {
+	return abbreviateNumber(count, 1, {padding: false}).toLowerCase();
```
fregante, line 4:
```suggestion
import {abbreviateNumber as abbreviateNumber_} from 'js-abbreviation-number';

export default function abbreviateNumber(count: number): string {
	return abbreviateNumber_(count, 1, {padding: false}).toLowerCase();
```

yakov116: rename file?

fregante (approved):

PR #4110, opened 2021-03-16
Add `select-all-notifications-shortcut` feature

### LINKED ISSUE #3876: Add a keyboard shortcut to select all notifications
I think that a keyboard shortcut to select all visible notifications on https://github.com/notifications would be a good addition to this project.

There are already built-in keyboard shortcuts to make selected notifications as read (`shift+i`) and done (`e`), but there isn't one to select all notifications. I think it's a fairly common workflow for people to open the notification centre, scroll through and open some notifications in new tabs and then select all and mark as done. It's a bit annoying that one has to click on the checkbox (see below) to select all instead of being able to do that just via a keyboard shortcut.

Screenshot 2021-01-09 at 23 48 28

I'd definitely prefer to type `a` (to select all) followed by `e` to mark all as done once I get to the bottom of the list rather having to scroll back up and click on the checkbox. GitHub doesn't seem to be keen on adding this shortcut, which seems to me like a small but a nice improvement to one's daily workflow.

Is it something that others would also find useful and would a contribution for this feature be welcomed?

cheap-glitch:
Closes #3876 

## Test URLs

## Screenshot
select-all-notifications

Should this be part of `open-all-notifications` instead?


source/features/select-all-notifications.ts
```diff
@@ -0,0 +1,16 @@
+import elementReady from 'element-ready';
+import * as pageDetect from 'github-url-detection';
+
+import features from '.';
+
+async function init(): Promise<void> {
+	(await elementReady<HTMLElement>('.js-notifications-mark-all-prompt'))!.dataset.hotkey = 'a';
```
cheap-glitch, line 7: Seems like the problem mentioned in https://github.com/sindresorhus/refined-github/issues/3876#issuecomment-757399774 (the "Select all" checkbox being removed) doesn't occur anymore, so the feature becomes even simpler.

readme.md
```diff
@@ -316,6 +316,7 @@ Thanks for contributing! 🦋🙌
 - [](# "improve-shortcut-help") [Shows all of Refined GitHub’s new keyboard shortcuts in the help modal (<kbd>?</kbd> hotkey).](https://user-images.githubusercontent.com/29176678/36999174-9f07d33e-20bf-11e8-83e3-b3a9908a4b5f.png)
 - [](# "stop-redirecting-in-notification-bar") [Stops redirecting to notification inbox from notification bar actions while holding <kbd>Alt</kbd>.](https://user-images.githubusercontent.com/202916/80318782-c38cef80-880c-11ea-9226-72c585f42a51.png)
 - [](# "hide-zero-packages") [Hides the `Packages` tab in user profiles if it’s empty.](https://user-images.githubusercontent.com/46634000/101151879-9ade9b00-3622-11eb-9755-50ed54099e02.png)
+- [](# "select-all-notifications") [TODO](TODO)
```
fregante, line 319: No screenshot necessary
```suggestion
- [](# "select-all-notifications") TODO.
```

fregante:
> Should this be part of [`open-all-notifications`](/sindresorhus/refined-github/blob/main/source/features/open-all-notifications.tsx) instead?

It's kinda related, but it's different.

source/features/select-all-notifications.ts
yakov116: What if I don't have any notifications?

image

cheap-glitch: > What if I don't have any notifications?

You mean what if new notifications appear when the page is open? I'd have to check but I'm pretty sure clicking the `x new notification(s)` link fully reloads the page.

yakov116: > Edit: or do you mean we should exclude an empty notification page?

Yes, is there a way to do that domReady safe?

cheap-glitch: I think it's easier to just check the result of `elementReady`?

PR #4115, opened 2021-03-17
Enable `one-key-formatting` on title fields

yakov116:
Closes #3718

## Test URLs
⏫ ⬆️ ⏫ ⬆️ ⏫ ⬆️ ⏫ ⬆️ ⏫ ⬆️ ⏫ ⬆️ ⏫ ⬆️ ⏫ ⬆️ ⏫ ⬆️ ⏫ ⬆️ 

## Screenshot
None

Need some typescript help

Also it was simpler to just enable all format options. Do you want to limit it?

PR #4116, opened 2021-03-17
Update `delegate-it`

fregante:
Trying https://github.com/fregante/delegate-it/pull/25 out. Unfortunately it doesn't help us at all since almost none of our listeners is inlined in the `delegate` call, so we still need to specify the callback parameter types.

~~WIP. Blocked by #4115~~


yakov116:
Unblocked

PR #4118, opened 2021-03-17
Add "Dark dimmed" version to `night-not-found`

fregante:
Context: https://github.blog/changelog/2021-03-16-dimmed-theme-beta-for-github-com/

Tested in both system-sync and single-theme options

github com_404 (2)
github com_404 (1)
github com_404


fregante:
And with this, we close another week and 21.3.17 is shipped.

Issue #3336, opened 2020-07-11
How disable align-repository-header feature?


cipchk:
It conflict with [Gitako](https://github.com/EnixCoda/Gitako) caused misalignment.

png:

![image](https://user-images.githubusercontent.com/2987467/87215155-12efb000-c366-11ea-8ee3-c2b8b9f6bbeb.png)

quangkieu:
I think the cause is on this.
https://github.com/sindresorhus/refined-github/compare/20.7.2...20.7.9#diff-765d4988f08edb45ad447cfeab8a3421R3
Could we have a switch to disable the padding left for those do not want it. And could we change it to something smaller like cal(10%) is what I found reasonable.

fregante:
We can *try* to fix the conflict with it

ZackBoe:
This gets weirder at larger resolutions:
![image](https://user-images.githubusercontent.com/904055/87233473-a391b480-c395-11ea-94c1-800c52ca9d65.png)

fregante:
This is not due to Refined GitHub. It happens even after you disable it

fregante:
Actually, that was incorrect, I had some extra CSS being applied even without Refined GitHub.

However, Gitako here is adding some styles (that do essentially the same thing) but probably shouldn't, because Gitako’s intent is supposedly to add a file tree, not change the layout.

So you can open an issue on their repo asking to drop this unrelated change. 

# Regular (no extensions)

![reg](https://user-images.githubusercontent.com/1402241/87233958-917f3880-c3cc-11ea-83ee-47d4e784527f.png)


# Gitako

Notice the misaligned header and commit content

![gitako](https://user-images.githubusercontent.com/1402241/87233960-9512bf80-c3cc-11ea-9036-6632bf72871f.png)

cipchk:
@fregante I have a suggestion to remove `!important` because `body:not(.full-width):not(.enterprise) .repohead> * {` has sufficient priority.

fregante:
@cipchk That element has a `!important`, so we must also use `!important`

fregante:
This feature has now been dropped in https://github.com/refined-github/refined-github/pull/6247

PR #4119, opened 2021-03-17
Mechanism to find features

### LINKED ISSUE #2586: Mechanism to find which feature is causing issues
> [git-bisect](https://git-scm.com/docs/git-bisect) - Use binary search to find the commit that introduced a bug

In Refined GitHub we could use this mechanism to determine **which feature** causes a specific bug (e.g. https://github.com/sindresorhus/refined-github/issues/2273) rather than _which commit_.

I'm not sure if it's worth it, but perhaps it's not terribly hard if implemented with a `window.prompt()` for yes/no questions running in the options page, as long as the previous settings are restored at the end.

cheap-glitch:
Closes #2586 

## Screenshot

gif


yakov116:
Your on 🔥

source/bisect.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export default async function bisectFeatures(): Promise<void> {
+	const options = perDomainOptions.getOptionsForOrigin();
+	const featuresState = Object.fromEntries(Object.entries(await options.getAll()).filter(([key]) => key.startsWith('feature:')));
+	const savedFeaturesState = {...featuresState};
+
+	// Test with all features disabled first
+	await onlyEnableFeatures(options, featuresState, []);
+	if (await confirmBug()) {
+		window.alert('It’s a native GitHub bug!');
```
busches, line 13: This isn't 100% true as not all features can be disabled.

cheap-glitch: We could change the message to something like
```text
The bug comes from a permanent CSS feature or GitHub itself.
```

source/bisect.ts
```diff
@@ -0,0 +1,54 @@
+import {OptionsSync} from 'webext-options-sync-per-domain';
+
+import {RGHOptions, perDomainOptions} from './options-storage';
+
+export default async function bisectFeatures(): Promise<void> {
+	const options = perDomainOptions.getOptionsForOrigin();
+	const featuresState = Object.fromEntries(Object.entries(await options.getAll()).filter(([key]) => key.startsWith('feature:')));
```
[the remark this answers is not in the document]
fregante: What do you think about moving this whole logic to the content script?

Advantage:

- the user doesn't need to juggle 2 windows (GitHub.com and the options)
- the popup appears wherever the user is and gives exact feedback on whether it's bisecting the current view or not

Disadvantage:

- we have to build the UI instead of relying on alert/confirm

I can help with the style of this UI, you can start with something like this inside `globalReady`:

```jsx
if (await cache.get('bisect')) {
	  document.body.append(
		  <div className="some GH class for background and border" style={{position: fixed, top: 0, left: 0}}>
		    {x} steps remaining. Can you see the bug?
		    <button onClick={onYes}>Yes</button> <!-- or a single {onClick} function -->
		    <button onClick={onNo}>No</button>
		  </div>
	  );
}
```

And then in the options page you just need this:

```js
async function bisectFeaturesHandler() {
	await cache.set('bisect', [], {minutes: 5});
	await browser.tabs.create({url: 'https://github.com'})
}
```

source/bisect.ts
```diff
@@ -0,0 +1,54 @@
...
+	// Test with all features disabled first
+	await onlyEnableFeatures(options, featuresState, []);
+	if (await confirmBug()) {
+		window.alert('It’s a native GitHub bug!');
+	} else {
+		let testedFeatures = Object.keys(featuresState);
+		while (testedFeatures.length > 1) {
+			const half = Math.ceil(testedFeatures.length / 2);
+			const firstHalf = testedFeatures.slice(0, half);
+			const stepsRemaining = Math.ceil(Math.log2(firstHalf.length));
```
fregante, line 19: 🙌

fregante (commented):
# 🤩

fregante:
I changed the title because while it's a meta feature, it's also user-facing and I want it to appear in the changelog

cheap-glitch:
I rewrote the PR as suggested. It's still a bit rough around the edges but it works.

PS: sorry for leaving so many PRs open for so long, I had some serious hardware troubles.

source/features/index.tsx
```diff
@@ -104,12 +105,81 @@ const globalReady: Promise<RGHOptions> = new Promise(async resolve => {
...
+
 	// Create logging function
 	log = options.logging ? console.log : () => {/* No logging */};
 
 	resolve(options);
 });
 
+function createMessageBox(message: string | Element, yesNoButtons = true): void {
+	delegate(document, '.rgh-bisect-button-choice', 'click', onBisectButtonChoiceClick);
+	delegate(document, '.rgh-bisect-button-end', 'click', onBisectButtonEndClick);
```
cheap-glitch, line 142: Had to go with the `delegate` route because `onClick` was giving me serious type headaches. (It demands a `MouseEventHandler<HTMLButton>` and I couldn't fix it with typecasting).

source/features/index.tsx
```diff
@@ -104,12 +105,81 @@ const globalReady: Promise<RGHOptions> = new Promise(async resolve => {
...
+async function onBisectButtonEndClick(): Promise<void> {
+	await cache.delete('bisect');
+	await browser.runtime.sendMessage({reloadTab: true});
+}
+
+function splitArray<T>(array: T[]): [T[], T[]] {
+	const half = Math.ceil(array.length / 2);
+
+	return [array.slice(0, half), array.slice(half)];
+}
```
cheap-glitch, line 181: Should this be moved to a helper module?

source/features/index.tsx
```diff
@@ -104,12 +105,81 @@ const globalReady: Promise<RGHOptions> = new Promise(async resolve => {
...
+
+		const [firstHalf] = testedFeatures;
+		options = {
+			...options,
+			...Object.fromEntries(__features__.map(feature => [`feature:${feature}`, firstHalf.includes(feature)]))
+		};
+
+		if (firstHalf.length === 0) {
+			createMessageBox('Every feature has been disabled. Can you see the bug?');
+			return;
```
cheap-glitch, line 123: Didn't realize this at first but using `return` every time shouldn't work... yet it does :thinking:

source/features/index.tsx
yakov116: Wouldn't be surprised if sindre has a package that does this

fregante: He does. Chunk-array or something like that

source/features/index.tsx
fregante: If you encounter type issues you can let us fix them, as long as the underlying code actually works.

source/features/index.tsx
fregante: It works because it should, but it it's discarding the `log` setting below.

This whole diff should be its own function, like `Object.assign(options, bisectFeatures)`

source/features/index.tsx
```diff
@@ -104,12 +105,81 @@ const globalReady: Promise<RGHOptions> = new Promise(async resolve => {
...
+	if (testedFeatures !== undefined) {
+		if (testedFeatures === false) {
+			createMessageBox('Every feature has been disabled. If you still see the bug, try disabling the whole extension.', false);
+			return;
+		}
+
+		const [firstHalf] = testedFeatures;
+		options = {
+			...options,
+			...Object.fromEntries(__features__.map(feature => [`feature:${feature}`, firstHalf.includes(feature)]))
```
fregante, line 118: Use `features.list` instead of the magic variable

source/features/index.tsx
fregante: Instead of splitting the array, you just need to store the whole list you're testing at the moment, you'll automatically know that half of them are currently disabled.

For example, on the step where 50% of the features are disabled, you store 100% of them and enable only the first half.

Then on the next step, you store 50% of them and enable half of that.

Etc. No need to store split arrays if you already know they're always going to be split in half. We just need to ensure that the math is right and consistent when there's an odd number of features.

cheap-glitch: > We just need to ensure that the math is right and consistent when there's an odd number of features

That's what the `Math.ceil()` is for: it ensures there's always at least one feature in the first half (so if it's empty, it means all features have been disabled).

source/features/index.tsx
```diff
@@ -104,12 +105,74 @@ const globalReady: Promise<RGHOptions> = new Promise(async resolve => {
...
 	resolve(options);
 });
 
+async function bisectFeatures(): Promise<Record<string, boolean>> {
+	const bisectedFeatures: string[] | false | undefined = await cache.get('bisect');
+	if (bisectedFeatures === undefined) {
+		return {};
+	}
+
+	if (bisectedFeatures === false) {
```
fregante, line 122: No need for a new type
```suggestion
	if (bisectedFeatures.length === 0) {
```

source/features/index.tsx
fregante: Let's use onClick

source/features/index.tsx
```diff
@@ -104,12 +105,74 @@ const globalReady: Promise<RGHOptions> = new Promise(async resolve => {
...
+}
+
+function createMessageBox(message: string | Element, yesNoButtons = true): void {
+	delegate(document, '.rgh-bisect-button-choice', 'click', onBisectButtonChoiceClick);
+	delegate(document, '.rgh-bisect-button-end', 'click', onBisectButtonEndClick);
+
+	document.body.append(
+		<div className="Box" style={{position: 'fixed', top: 0, left: 0}}>
+			<p>{message}</p>
+			{yesNoButtons ? <button type="button" className="btn btn-danger rgh-bisect-button-choice" data-answer="no">No</button> : undefined}
```
fregante, line 146:
```suggestion
			{yesNoButtons ? <button type="button" className="btn btn-danger rgh-bisect-button-choice" value="no">No</button> : undefined}
```

source/features/index.tsx
```diff
@@ -104,12 +105,74 @@ const globalReady: Promise<RGHOptions> = new Promise(async resolve => {
...
+			{yesNoButtons ? <button type="button" className="btn btn-primary rgh-bisect-button-choice" data-answer="yes">Yes</button> : undefined}
+			{yesNoButtons ? undefined : <button type="button" className="btn btn-primary rgh-bisect-button-end">OK</button>}
```
fregante, line 148: Probably:

```suggestion
			{yesNoButtons ? undefined : <button type="button" className="btn btn-primary rgh-bisect-button-end">End</button>}
```

source/features/index.tsx
```diff
@@ -104,12 +105,74 @@ const globalReady: Promise<RGHOptions> = new Promise(async resolve => {
...
+
+	if (bisectedFeatures.length === 0) {
+		bisectedFeatures = answer === 'yes' ? false : features.list;
+	} else {
+		const half = Math.ceil(bisectedFeatures.length / 2);
+		bisectedFeatures = answer === 'yes' ? bisectedFeatures.slice(0, half) : bisectedFeatures.slice(half);
+	}
+
+	await cache.set('bisect', bisectedFeatures);
+	await browser.runtime.sendMessage({reloadTab: true});
```
fregante, line 168:
```suggestion
	location.reload();
```

source/features/index.tsx
```diff
@@ -104,12 +105,74 @@ const globalReady: Promise<RGHOptions> = new Promise(async resolve => {
...
 
+	Object.assign(options, await bisectFeatures());
+
 	// Create logging function
 	log = options.logging ? console.log : () => {/* No logging */};
 
 	resolve(options);
 });
 
+async function bisectFeatures(): Promise<Record<string, boolean>> {
```
fregante, line 116: This and related functions should be moved to their own file in `source/helpers` or something

source/features/index.tsx
cheap-glitch: Unless I'm missing something, this is necessary to distinguish between the first step (`Every feature has been disabled. Can you see the bug?`) and the case where the user answered "Yes" to that (`Every feature has been disabled. If you still see the bug, try disabling the whole extension.`)

fregante: If I'm reading this correctly, in the first case we're disabling all features (length === 0) while in the last step `bisect` will have no less than 2 features:

1. one enabled
2. one disabled

So at this point "yes" means 1 is buggy and "no" means 2 is buggy.

distribution/options.html
```diff
@@ -63,6 +63,9 @@
 			<label>
 				<button id="clear-cache">Clear cache</button>
 			</label>
+			<label>
+				<button id="bisect-features">Bisect features</button>
```
fregante, line 67: Let's brand this feature in a way that suggests it can help you find a feature, not just bugs:

```suggestion
				<button id="bisect-features">Find feature</button>
```

source/features/index.tsx
cheap-glitch: Yes, and if there's an odd number of features it could also end up with 2 features enabled and 1 disabled, at which point if the user says no it ends by pointing the single disabled feature as the culprit.

distribution/options.html
fregante: Likewise, mentions of bugs should be removed. This could be the first message:

```
This process will help you identify what Refined GitHub feature is making changes or causing issues on GitHub. 

x steps remaining, do you still see the change or issue?

Note: You can change page here, but don't use multiple tabs.
```

source/helpers/bisect.tsx
```diff
@@ -0,0 +1,61 @@
...
+	return Object.fromEntries(features.list.map(feature => [`feature:${feature}`, firstHalf.includes(feature)]));
+}
+
+function createMessageBox(message: string | Element, yesNoButtons = true): void {
+	document.body.append(
+		<div className="Box" style={{position: 'fixed', top: 0, left: 0}}>
+			<p>{message}</p>
+			{yesNoButtons ? <button type="button" className="btn btn-danger" value="no" onClick={onChoiceButtonClick}>No</button> : undefined}
+			{yesNoButtons ? <button type="button" className="btn btn-primary" value="yes" onClick={onChoiceButtonClick}>Yes</button> : undefined}
+			{yesNoButtons ? undefined : <button type="button" className="btn btn-primary" onClick={onEndButtonClick}>End</button>}
```
fregante, line 35: Sorry I had misunderstood this button, we should probably always show it
```suggestion
			<button type="button" className="btn btn-primary" onClick={onEndButtonClick}>Exit</button>
```

source/helpers/bisect.tsx
```diff
@@ -0,0 +1,61 @@
...
+		</div>
+	);
+}
+
+async function onChoiceButtonClick({target}: React.MouseEvent<HTMLButtonElement>): Promise<void> {
+	const answer = (target as HTMLButtonElement).value;
```
fregante, line 41: `currentTarget` should already have the right type

source/helpers/bisect.tsx
```diff
@@ -0,0 +1,61 @@
...
+	} else {
+		createMessageBox(`${Math.ceil(Math.log2(firstHalf.length))} steps remaining. Can you see the bug?`);
+	}
+
+	return Object.fromEntries(features.list.map(feature => [`feature:${feature}`, firstHalf.includes(feature)]));
+}
+
+function createMessageBox(message: string | Element, yesNoButtons = true): void {
+	document.body.append(
+		<div className="Box" style={{position: 'fixed', top: 0, left: 0}}>
```
fregante, line 31:
```suggestion
		<div className="Box p-3" style={{position: 'fixed', bottom: 10, left: '50%', transform: 'translateX(-50%)'}}>
```

Screen Shot 2021-04-04 at 00 15 27

Probably also the width needs to be limited somehow. Alternative the horizontal centering could be achieved via flexbox + a `pointer-events:none` full-width container

distribution/options.html
cheap-glitch: > Let's brand this feature in a way that suggests it can help you find a feature

How about "Pinpoint a feature" for the button text?

fregante:
I think the conclusion is incorrect. On this page I tried identifying `hide-watch-and-fork-count` but it ended up telling me it was `hide-useless-newsfeed-events`, which is probably the other feature

source/features/index.tsx
fregante: So we agree that `false` should be dropped, right? Just to make sure

source/helpers/bisect.tsx
```diff
@@ -0,0 +1,61 @@
...
+		bisectedFeatures = answer === 'yes' ? bisectedFeatures.slice(0, half) : bisectedFeatures.slice(half);
+	}
+
+	await cache.set('bisect', bisectedFeatures);
+	location.reload();
+}
+
+async function onEndButtonClick(): Promise<void> {
+	await cache.delete('bisect');
+	await browser.runtime.sendMessage({reloadTab: true});
```
fregante, line 60: Same here
```suggestion
	location.reload();
```

source/options.tsx
```diff
@@ -129,6 +129,11 @@ async function clearCacheHandler(event: Event): Promise<void> {
 	}, 2000);
 }
 
+async function bisectFeaturesHandler(): Promise<void> {
+	await cache.set('bisect', [], {minutes: 5});
+	await browser.tabs.create({url: 'https://github.com'});
```
fregante, line 134: Ending up on the homepage is a little confusing because it immediately asks the user if they can see the feature. Instead, in this page we should just instruct the user to visit the page where they're seeing the feature and refresh it to see the new instructions on screen.

Ideally this is done via `alert()` but it might not work on Firefox. In that's the case, let's just add a description of this in the HTML file (a new `<p>`)

distribution/options.html
fregante: I think "Find" is more understandable also for international users.

source/helpers/bisect.tsx
```diff
@@ -0,0 +1,71 @@
...
+		return {};
+	}
+
+	if (bisectedFeatures === false) {
+		createMessageBox(<p>Every feature has been disabled. If you still see the change or issue, try disabling the whole extension.</p>, false);
+		return buildOptionsObject([]);
+	}
+
+	if (bisectedFeatures.length === 0) {
+		createMessageBox(<div><p>This process will help you identify what Refined GitHub feature is making changes or causing issues on GitHub. Do you still see the change or issue?</p><p>Note: You can change page here, but don’t use multiple tabs.</p></div>);
```
fregante, line 18: Probably most of this text can be moved to the options page

source/helpers/bisect.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+function createMessageBox(message: Element, yesNoButtons = true): void {
+	document.body.append(
+		<div className="Box p-3" style={{position: 'fixed', bottom: 10, left: '50%', maxWidth: '600px', transform: 'translateX(-50%)'}}>
+			{message}
+			<div className="d-flex flex-justify-between">
+				<div>
+					<button type="button" className="btn" onClick={onEndButtonClick}>Exit</button>
+				</div>
+				{yesNoButtons ?
```
fregante, line 40: I never remember, does this work?
```suggestion
				{yesNoButtons ??
```

source/features/index.tsx
fregante: I think the main misunderstanding here is that `onChoiceButtonClick` always reloads the page, but instead when the user is on the last yes/no step it should just update the box without reloading.

This avoids having to set `false` altogether, because there are no more actual steps/reloads to follow.

The good part is that we already always have Exit available, so a final yes/no click only needs to:

1. Update the paragraph element
2. Drop the yes/no buttons
3. Clear the cache immediately (so that he user immediately exits the process even by closing the tab)

cheap-glitch: > `false` should be dropped

No, I don't think so?

> 1. Remember the deselected options
> 
> 2. Deselect all
> 
> 3. Ask: can you see the bug?
>        
>   * Yes: Exit, it's not RGH
>   * No: Select half of the feature
> 
> 4. Ask: Can you see the bug?
>        
>   * Yes: Select half of the previous features
>   * No: Select half of the _other_ features
> 
> 5. Repeat 4 until there's only a feature left
> 
> 6. Tell the user which feature it is
> 
> 7. Restore previous options

We need to be able to differentiate between steps `3` and `3` + `Yes`

fregante: I’m not sure I see the issue:

Step 3 will always receive []
- no will set [half of 1]
- yes will exit

Step 4 will always receive [half of 1], if it didn’t exit

cheap-glitch: > shows the final message
>
> displays an Exit button

As the code is written now, this implies a reload of the page. I'll change it so it doesn't and we can drop the `false` then.

source/helpers/bisect.tsx
cheap-glitch: Doesn't seem like it does.

source/helpers/bisect.tsx
```diff
@@ -0,0 +1,73 @@
...
+	const bisectedFeatures: string[] | undefined = await cache.get('bisect');
+	if (bisectedFeatures === undefined) {
+		return {};
+	}
+
+	const enabledFeatures = new Set(bisectedFeatures.slice(0, Math.ceil(bisectedFeatures.length / 2)));
+	if (bisectedFeatures.length === 0) {
+		createMessageBox(<div><p>Every feature has been disabled. Do you still see the change or issue?</p><p>Note: You can change page here, but don’t use multiple tabs.</p></div>);
+	} else if (bisectedFeatures.length === 1) {
+		createMessageBox(<p>The change or issue is caused by <a href={'https://github.com/sindresorhus/refined-github/blob/main/source/features/' + bisectedFeatures[0] + '.tsx'}><code>{bisectedFeatures[0]}</code></a>.</p>, false);
```
fregante, line 17: Ideally this step should also be moved to onChoiceButtonClick, but I can see that it'd mean copying the `slice` there as well.

At a minimum, we can add
```suggestion
	} else if (bisectedFeatures.length === 1) {
		createMessageBox(<p>The change or issue is caused by <a href={'https://github.com/sindresorhus/refined-github/blob/main/source/features/' + bisectedFeatures[0] + '.tsx'}><code>{bisectedFeatures[0]}</code></a>.</p>, false);
		await cache.delete('bisect');
		return;
```

source/helpers/bisect.tsx
```diff
@@ -0,0 +1,73 @@
...
+	} else {
+		createMessageBox(<p>Can you see the change or issue? ({Math.ceil(Math.log2(bisectedFeatures.length))} steps remaining.)</p>);
```
fregante, line 19:
```suggestion
		createMessageBox(<p>Can you see the change or issue? ({Math.ceil(Math.log2(bisectedFeatures.length))} steps remaining)</p>);
```

fregante:
I think we're near! The feature seems to work correctly

source/helpers/bisect.tsx
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	return Object.fromEntries(features.list.map(feature => [`feature:${feature}`, enabledFeatures.has(feature)]));
+}
+
+function createMessageBox(message: Element, yesNoButtons = true): void {
+	document.body.append(
+		<div id="rgh-bisect-dialog" className="Box p-3" style={{position: 'fixed', bottom: 10, left: '50%', maxWidth: '600px', transform: 'translateX(-50%)'}}>
```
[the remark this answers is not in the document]
fregante: Maybe I'd also add a generous red box-shadow so it appears to be on top of everything, because it kind of blends in.

Example `box-shadow: 0 0px 40px var(--color-text-danger);`

Screen Shot 2021-04-05 at 02 19 26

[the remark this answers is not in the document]
fregante: Or a lighter `var(--color-toast-shadow)` (which might be available as a class too)

Screen Shot 2021-04-05 at 02 21 39

source/helpers/bisect.tsx
```diff
@@ -0,0 +1,73 @@
...
+		</div>
+	);
+}
+
+async function onChoiceButtonClick({currentTarget}: React.MouseEvent<HTMLButtonElement>): Promise<void> {
+	const answer = currentTarget.value;
+	const bisectedFeatures: string[] | undefined = await cache.get('bisect');
+	if (!bisectedFeatures) {
+		return;
+	}
```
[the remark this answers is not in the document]
fregante: Rather, if it happens the user opened multiple tabs, so the correct solution would be to _still_ refresh the page and exit
```suggestion
	if (!bisectedFeatures) {
		location.reload();
		return;
	}
```

source/helpers/bisect.tsx
fregante: Also it needs to use `pluralize()`

source/helpers/bisect.tsx
fregante: It works, I meant it should replace the ternary operator, not just this line

cheap-glitch: > I meant it should replace the ternary operator

I understood that, but that's not how nullish coalescing works (unless it has another meaning in JSX?)

fregante:
My review was resolved because I'm working on this, I assigned myself. I'll have to force-push

fregante:
It appears to be working correctly. The only "bug" see is that it might jump from 3 steps to "last step". That's because depending on an answer with an odd number of features, the last steps might have:

1. 1 enabled, 2 disabled
2. 1 enabled, 1 disabled
3. 1 disabled

or

1. 1 enabled, 2 disabled
3. 1 disabled

fregante:
Let me know if you see any issues with the logic and if it works well for you. Try finding different features on the page or even GitHub features (which should result in the "everything was disabled" message)

fregante (approved):

source/helpers/bisect.tsx
```diff
@@ -25,7 +25,19 @@ export default async function bisectFeatures(): Promise<Record<string, boolean>>
...
+			className="Box p-3"
+			style={{
+				position: 'fixed',
+				bottom: '50%',
+				right: '50%',
+				maxWidth: '600px',
+				transform: 'translate(50%, 50%)',
+				boxShadow: 'var(--color-toast-shadow)',
+				zIndex: 2147483647
+			}}
```
fregante, line 64: We can probably move this to `refined-github.css`

source/helpers/bisect.tsx
```diff
@@ -77,6 +77,7 @@ async function onChoiceButtonClick({currentTarget}: React.MouseEvent<HTMLButtonE
 			bisectedFeatures.slice(getMiddleStep(bisectedFeatures))
 		);
 
+		currentTarget.parentElement!.replaceWith(<div className="btn btn-disabled">Reloading…</div>);
```
fregante: `btn-disabled` doesn't seem to work. Does anyone know the correct class without having to use `<button>`?

source/helpers/bisect.tsx
fregante: Also the width can be dropped

source/helpers/bisect.tsx
cheap-glitch: `aria-disabled="true"` seems to work

source/refined-github.css
```diff
@@ -138,3 +138,13 @@ pr-branches
 .select-menu-item-text .diffstat {
 	display: inline !important;
 }
+
+/* Style the "Find feature" dialog that appears on top of the page */
+#rgh-bisect-dialog {
+	position: fixed;
+	bottom: 50%;
```
cheap-glitch, line 145: Why move the dialog to the middle of the page?

fregante: Feels like that'a where dialogs belong. It felt lost at the bottom. Since the rest of the page is scrollable, I think the user can either resize the window or scroll to see under it

distribution/options.html
```diff
@@ -64,6 +64,20 @@
 				<button id="clear-cache">Clear cache</button>
 			</label>
 		</p>
+		<p>
+			<strong>Find a feature</strong>
+		</p>
+		<p>
+			This process will help you identify what Refined GitHub feature is making changes or causing issues on GitHub.
```
sindresorhus, line 71: Should we already here tell the user to first disable Refined GitHub to make sure the issue is with Refined GitHub?

sindresorhus (changes requested):

distribution/options.html
fregante: I had proposed it this way at the beginning but I’m not particularly interested in that, the user can just toggle the extension if they feel this situation is likely. 

If not, they can just follow the procedure as a regular bisect, and that’s just easier to reason and code for us.

sindresorhus: But isn't it worth trying this before wasting time on the bisect?

cheap-glitch:
Tried it with native and non-native features, works great!

Small UX suggestion: how about disabling the "Yes"/"No" buttons until the page has finished loading? A few times I wrongly thought a feature didn't appear and clicked the "No" button by mistake, which means having to restart the whole process.

> The only "bug" see is that it might jump from 3 steps to "last step"

If you're talking about the number of steps left displayed, yeah that might happen because it's a "worst case" prediction. We could reword the message slightly: "(_at most_ x steps left)

Also I agree that it's better to have the "It might not be RGH" message at the end of the process, but in that case a little disclaimer in the options page wouldn't hurt. (Even if it's already in the issue template.)

fregante:
> Small UX suggestion: how about disabling the "Yes"/"No" buttons until the page has finished loading? A few times I wrongly thought a feature didn't appear and clicked the "No" button by mistake, which means having to restart the whole process.

Some features load after the `load` event though, so it won't catch 100% of the scenarios, but it might still be an improvement.



> "(_at most_ x steps left)

heheh I think it's fine the way it is, without complicating things.

distribution/options.html
fregante: Do you personally see yourself using this wizard for that step?

> But isn't it worth trying this before wasting time on the bisect?

Yes and no, fully disabling the extension is more effective than this anyway since we have non-disableable CSS features and code that runs outside `init`, e.g. https://github.com/sindresorhus/refined-github/issues/3336

I know I personally would probably disable it entirely in the first place if I'm unsure whether a bug is caused by RG, e.g. https://github.com/sindresorhus/refined-github/issues/3055#issuecomment-623152965

fregante:
> little disclaimer about disabling the extension in the options page wouldn't hurt

It's there

Screen Shot 2021-04-06 at 09 46 24

source/helpers/bisect.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+	console.log(`Bisecting ${bisectedFeatures.length} features:\n${bisectedFeatures.join('\n')}`);
+	console.log(getMiddleStep(bisectedFeatures));
+
+	const steps = Math.ceil(Math.log2(Math.max(bisectedFeatures.length))) + 1;
+	createMessageBox(
+		`Do you see the change or issue? (${pluralize(steps, 'last step', '$$ steps remaining')})`,
+		<div>
+			<button type="button" className="btn btn-danger mr-2" value="no" aria-disabled="true" onClick={onChoiceButtonClick}>No</button>
+			<button type="button" className="btn btn-primary" value="yes" aria-disabled="true" onClick={onChoiceButtonClick}>Yes</button>
```
fregante, line 78:
```suggestion
			<button type="button" className="btn btn-primary" value="yes" disabled onClick={onChoiceButtonClick}>Yes</button>
```

distribution/options.html
```diff
@@ -64,6 +64,20 @@
...
+		</p>
+		<p>
+			<label>
```
fregante, line 74: No label needed on buttons

distribution/options.html
```diff
@@ -64,6 +64,20 @@
...
+				<button id="find-feature">Find feature…</button>
+			</label>
+		</p>
+		<p id="find-feature-message" hidden>
```
fregante, line 78: I think this should actually go above the button, so when clicking it, the button jumps down. This also helps avoid an issue where the button lines up with the bottom border, so the new message wouldn't be visible.

source/helpers/bisect.tsx
```diff
@@ -0,0 +1,107 @@
...
+		</div>
+	);
+
+	// Enable "Yes"/"No" buttons when the page has done loading
```
fregante, line 82:
```suggestion
	// Enable "Yes"/"No" buttons once the page is done loading
```

source/helpers/bisect.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+	const half = getMiddleStep(bisectedFeatures);
+	const temporaryOptionsPairs = features.list.map(feature => {
+		const index = bisectedFeatures.indexOf(feature);
+		return [
+			`feature:${feature}`,
+			index > -1 && index < half
+		];
+	});
+	const temporaryOptions = Object.fromEntries(temporaryOptionsPairs);
```
fregante, line 104: Functional programming is a lie 🍰 😃 
```suggestion
	const temporaryOptions = {};
	for (const feature of features.list) {
		const index = bisectedFeatures.indexOf(feature);
		temporaryOptions[`feature:${feature}`] = index > -1 && index < half;
	}
```

You might need `as FeatureID` somewhere

source/options.tsx
```diff
@@ -16,6 +16,9 @@ interface Status {
 	scopes?: string[];
 }
 
+// Don't repeat the magic variable, the content will be injected multiple times
```
fregante, line 19:
```suggestion
// Don't repeat the magic variable, or its content will be inlined multiple times
```

fregante (approved):
I haven't tested the latest revision, but after this it should be good to go

fregante:
I'll add a minimum width to avoid this on the last step:

gif

The button moves slightly, and if you're just pressing "no" multiple times in a row without moving your mouse, suddenly the mouse will be on the "yes"

fregante:
# 💯

cheap-glitch:
Now let's see if this is actually going to be useful :smile:

fregante:
I needed it several times when trying to find buggy features already, so I'm confident I will again 😃

PR #4120, opened 2021-03-17
Fix `releases-tab` not showing occasionally

### LINKED ISSUE #3942: `releases-tab` not shown occasionally
Hello, this extension is fantastic, but recently I ran into a trouble that the release button is not shown. It was probably related with the `.latest-tag-button → TypeError: select_dom(...) is undefined` error, which may be linked with the newly added buttons on that list. （Sometimes the Release button shows, but most of the times not, with only this error shown）

Thanks for making this extension and for your time spent on this issue.

(On)Screen Shot 2021-02-04 at 15 18 58
(Off)Screen Shot 2021-02-04 at 15 19 23

On Firefox 85.0 on macOS, using Refined GitHub 21.1.23.

cheap-glitch:
Fixes #3942 

## Test URLs
* [Repo homepage](https://github.com/sindresorhus/refined-github)
* [Releases page](https://github.com/sindresorhus/refined-github/releases)

The bug can be made to appear consistently by throttling the network speed (in Firefox Devtools I use the lowest speed, GPRS).


source/features/releases-tab.tsx
```diff
@@ -15,14 +15,14 @@ import {buildRepoURL, getRepo} from '../github-helpers';
...
+async function parseCountFromDom(): Promise<number> {
 	const releasesCountElement = select('.numbers-summary a[href$="/releases"] .num');
 	if (releasesCountElement) {
 		return looseParseInt(releasesCountElement);
 	}
 
-	// In "Repository refresh" layout, look for the releases link in the sidebar
-	const moreReleasesCountElement = select('[href$="/tags"] strong');
+	// In "Repository refresh" layout, look for the tags link in the header
+	const moreReleasesCountElement = await elementReady('[href$="/tags"] strong');
```
cheap-glitch, line 25: The bug was caused by this selector. It currently targets the `xx tags` link in the homepage header:
image

But since it's lazy-loaded, sometimes it doesn't exist yet and thus `parseCountFromDom()` returns `0`, which in turn causes the feature to give up.

cheap-glitch:
We could also take this opportunity to:
* drop the parts of the feature dealing with the pre-refresh layout (which would probably halve its length)
* make the counter selector (`[href$="/tags"] strong`) more specific

yakov116:
> * drop the parts of the feature dealing with the pre-refresh layout (which would probably halve its length)

We need to wait for june

source/features/releases-tab.tsx
fregante: Keep in mind that [by default](https://github.com/sindresorhus/element-ready#stopondomready) `elementReady` does not work for lazy-loaded elements and the option to do so is unsafe. You'll have to rely on whatever load event they fire or use an observer: simplified, oneMutation or (ideally not) selector-observer

fregante:
> * make the counter selector (`[href$="/tags"] strong`) more specific

Wow, indeed it's pretty generic, it could match any link in the readme

source/features/releases-tab.tsx
```diff
@@ -15,19 +16,22 @@ import {buildRepoURL, getRepo} from '../github-helpers';
...
 	if (releasesCountElement) {
 		return looseParseInt(releasesCountElement);
 	}
 
-	// In "Repository refresh" layout, look for the releases link in the sidebar
-	const moreReleasesCountElement = select('[href$="/tags"] strong');
-	if (moreReleasesCountElement) {
-		return looseParseInt(moreReleasesCountElement);
+	// In "Repository refresh" layout, look for the tags link in the header
+	const repositoryHeader = (await elementReady('.repository-content .file-navigation', {waitForChildren: false}))!;
```
cheap-glitch, line 26: Maybe `{waitForChildren: true}` would circumvent the need for `oneMutation`? Or is this the same pitfall as before?

fregante: I'm confused as to why/when `oneMutation` would be required at all. The value seems to be in the Dom on load, so the only fix in this PR should be to use `elementReady` instead of `select-dom`.

Have you found a situation where the content is loaded later via `include-fragment`? If so, let's target that `include-fragment` directly, like we do in other features.

cheap-glitch: > Have you found a situation where the content is loaded later via include-fragment?

No. I guess I misconstrued your previous comment? ↓

> Keep in mind that by default elementReady does not work for lazy-loaded elements and the option to do so is unsafe. You'll have to rely on whatever load event they fire or use an observer: simplified, oneMutation or selector-observer

I may be wrong and it's not really lazy-loaded. After all `elementReady` seemed to solve the problem when I tested it.

fregante (approved):

scris:
thank you very much for fixing this

PR #4127, opened 2021-03-18
Don't `collapse-markdown-sections` when clicking links

### LINKED ISSUE #4126: `collapse-markdown-sections` should not apply to links
1. View a markdown file with links in headings, for example [this one](https://github.com/kidonng/api)
2. Click the link
3. The section is collapsed but it shouldn't

cheap-glitch:
Fixes #4126 

## Test URLs
- [Links in headings](https://github.com/kidonng/api)
- [Links inside `<code>` elements in headings](https://github.com/cheap-glitch/git-grapnel)

## Screenshot
n/a

Note that the sections can still be toggled by clicking on the "side" of the links (since headings are blocks).

This fix also prevents the toggling of a section when the "anchor links" are clicked (which happens otherwise since they are inside the heading elements).

### DIFF
--- a/source/features/collapse-markdown-sections.ts
+++ b/source/features/collapse-markdown-sections.ts
@@ -3,12 +3,18 @@ import * as pageDetect from 'github-url-detection';
 
 import features from '.';
 
-function onHeadingClick({delegateTarget: sectionHeading}: delegate.Event<MouseEvent, HTMLElement>): void {
+function onHeadingClick(event: delegate.Event<MouseEvent, HTMLElement>): void {
 	// Don't toggle the section if the title text is being selected instead of clicked
 	if (document.getSelection()?.type === 'Range') {
 		return;
 	}
 
+	// Don't toggle the section if a link in the heading is clicked (either the content or the anchor)
+	if ((event.target as HTMLElement).closest('h1, h2, h3, h4, h5, h6, a')!.tagName === 'A') {
+		return;
+	}
+
+	const sectionHeading = event.delegateTarget;
 	const isSectionHidden = sectionHeading.classList.toggle('rgh-markdown-section-collapsed');
 	let element = sectionHeading.tagName === 'H1' ?
 		sectionHeading.parentElement!.firstElementChild as HTMLElement :
--- a/source/features/collapse-markdown-sections.ts
+++ b/source/features/collapse-markdown-sections.ts
@@ -10,7 +10,7 @@ function onHeadingClick(event: delegate.Event<MouseEvent, HTMLElement>): void {
 	}
 
 	// Don't toggle the section if a link in the heading is clicked (either the content or the anchor)
-	if ((event.target as HTMLElement).closest('h1, h2, h3, h4, h5, h6, a')!.tagName === 'A') {
+	if ((event.target as HTMLElement).closest('a')) {
 		return;
 	}
 


source/features/collapse-markdown-sections.ts
```diff
@@ -3,12 +3,18 @@ import * as pageDetect from 'github-url-detection';
...
+function onHeadingClick(event: delegate.Event<MouseEvent, HTMLElement>): void {
 	// Don't toggle the section if the title text is being selected instead of clicked
 	if (document.getSelection()?.type === 'Range') {
 		return;
 	}
 
+	// Don't toggle the section if a link in the heading is clicked (either the content or the anchor)
+	if ((event.target as HTMLElement).closest('h1, h2, h3, h4, h5, h6, a')!.tagName === 'A') {
+		return;
+	}
```
cheap-glitch, line 15: I never know if I should merge the two conditions in that case :sweat_smile:

source/features/collapse-markdown-sections.ts
```diff
@@ -3,12 +3,18 @@ import * as pageDetect from 'github-url-detection';
...
 
-function onHeadingClick({delegateTarget: sectionHeading}: delegate.Event<MouseEvent, HTMLElement>): void {
+function onHeadingClick(event: delegate.Event<MouseEvent, HTMLElement>): void {
 	// Don't toggle the section if the title text is being selected instead of clicked
 	if (document.getSelection()?.type === 'Range') {
 		return;
 	}
 
+	// Don't toggle the section if a link in the heading is clicked (either the content or the anchor)
+	if ((event.target as HTMLElement).closest('h1, h2, h3, h4, h5, h6, a')!.tagName === 'A') {
```
fregante, line 13:
```suggestion
	if ((event.target as HTMLElement).closest('a')) {
```

It doesn't matter which one is the closest, right? If it's a link, then it should be ignored

source/features/collapse-markdown-sections.ts
fregante: Since they both have long comments, they're easier to read this way rather than

```js
if (
	// Don't toggle the section if the title text is being selected instead of clicked
	document.getSelection()?.type === 'Range' ||

 	// Don't toggle the section if a link in the heading is clicked (either the content or the anchor)
 	(event.target as HTMLElement).closest('a')
) {

File: source/features/collapse-markdown-sections.ts (hunk 2 above)
Comment (cheap-glitch) [reply source/features/collapse-markdown-sections.ts:13]: Yes but what if one of the ancestors is a link outside the heading element?

**Review by fregante — APPROVED** [review]
Probably good! Can you think of any other troublesome situations with content in headers?

**Comment by cheap-glitch** [comment]
> Probably good! Can you think of any other troublesome situations with content in headers?

I couldn't find any examples on the spot, but there are some readmes where the `<h1>` is an image, which is automatically linked by GitHub to the image location. In that case one would actually want the link to be "disabled" so that clicking on the image/logo would toggle the whole ToC. Not sure it's worth the effort though :man_shrugging:

**Comment by fregante** [comment]
> but there are some readmes where the `<h1>` is an image, which is automatically linked by GitHub to the image location

Indeed, it's still a regular link though, so it's covered by the current code.

> In that case one would actually want the link to be "disabled"

That kinda gives me an idea for a new feature: disable image auto-links if they point to themselves. This would help me distinguish useless auto-links from custom links 🤔 wanna open a feature request?

**Comment by fregante** [comment]
47 minutes later…


```html
<details><summary><h1>Sup</h1> dude?</summary> <h2>Not<h2> much</details>
```

This probably needs to be excluded as well. `.closest('a, details')`?

fregante:
Also I think we should add a expand cursor via CSS, I'm thinking something like:

```
.markdown-body :is(h1, …) {cursor: expand-y or whatever}
.markdown-body :is(a, details) :is(h1, …) {cursor: initial}

kidonng:
Indeed. Currently it doesn't give any visual clue what has happened if one triggers the feature without knowing it.

PR #4130, opened 2021-03-19
Only expand Checks list on PRs, not in popups

### LINKED ISSUE #4124: Expand only PR merge status list to fit all items
## Summary

There was previously introduced [a CSS rule to show all PR merge status items instead of scrollable div](https://github.com/sindresorhus/refined-github/blob/19330b2a4d1715c2d2496e6435a6b43263031c03/source/refined-github.css#L78-L81). The original intention was using it for PR merge statuses, but this CSS rule also applies to the modals as well (in some cases, it doesn't due to the class `open` not being applied). For repositories with a sizable merge status list ([npm/arborist](https://github.com/npm/arborist) is the example in the screenshot), the height of the list without a scrollbar is quite silly :yum:

I would like to recommend a more modest size (maybe up to 10 items?). I understand the original reason to make it longer, but with no overflow the user will still have to scroll (the page), which defeats the purpose of the original intention of showing all items :sweat_smile:

I would also like to expand on this (in the spirit of this repository), and suggest a future potential enhancement. Maybe this would turn into its own feature. It may be nice to add the ability to resize the merge status list.

This is my first issue in this repository, by the way. Thank you for developing this! Why did I not start using this sooner! :blush: :clap: 

## Screenshot(s)

<details>
  <summary>Long merge status list</summary>
  
  Long merge status list
</details>

## References

Relates to #1004.

jameschensmith:
## Summary

Expands only the PR merge status list (and not the popups/dropdowns) to fit all items. This can be thought of as a fix for the popups/dropdowns.

## Test URL(s)

* [dropdown(s) (main)](https://github.com/npm/arborist)
* [dropdown(s) (commit history)](https://github.com/npm/arborist/commits/main)
* [pull request](https://github.com/npm/arborist/pull/254)

## Screenshot(s)

<details>
  <summary>Long merge status list fixed</summary>

  Long merge status list fixed
</details>

## References

Fixes #4124.

### DIFF
--- a/source/refined-github.css
+++ b/source/refined-github.css
@@ -75,10 +75,9 @@
 	color: #6a737d;
 }
 
-/* Expand merge status lists to fit more items */
-.branch-action-item-simple > .merge-status-list, /* Dropdowns */
-.branch-action-item.open > .merge-status-list { /* Pull requests */
-	max-height: 512px !important;
+/* Expand PR merge status list to fit all items */
+:root .branch-action-item.open:not(.branch-action-item-simple) > .merge-status-list {
+	max-height: none;
 }
 
 /* Make <details> visibly clickable */
--- a/source/refined-github.css
+++ b/source/refined-github.css
@@ -75,9 +75,10 @@
 	color: #6a737d;
 }
 
-/* Expand PR merge status list to fit all items instead of just 5 */
-:root .branch-action-item.open > .merge-status-list {
-	max-height: none;
+/* Expand merge status lists to fit more items */
+.branch-action-item-simple > .merge-status-list, /* Dropdowns */
+.branch-action-item.open > .merge-status-list { /* Pull requests */
+	max-height: 512px !important;
 }
 
 /* Make <details> visibly clickable */


fregante:
On second thought:

- on the PR page, the section is part of the normal flow of the page, so I feel it's perfectly fine to leave it as _infinite._ It's already togglable, so it's doesn't make sense to limit it in anyway.
- the popup’s awkwardly extends past the natural height of the page, so it makes sense to shorten that.

Maybe we can avoid changing the popup’s height altogether, and only leave `max-height: none` on the PR

source/refined-github.css
```diff
@@ -75,9 +75,10 @@
 	color: #6a737d;
 }
 
-/* Expand PR merge status list to fit all items instead of just 5 */
-:root .branch-action-item.open > .merge-status-list {
-	max-height: none;
+/* Expand merge status lists to fit more items */
+.branch-action-item-simple > .merge-status-list, /* Dropdowns */
+.branch-action-item.open > .merge-status-list { /* Pull requests */
```
fregante, line 80:
```suggestion
/* Expand PR merge status list to fit more items */
.branch-action-item.open > .merge-status-list {
```

source/refined-github.css
```diff
@@ -75,9 +75,10 @@
...
+	max-height: 512px !important;
```
fregante, line 81:
```suggestion
/* Expand PR merge status list to fit more items */
:root .branch-action-item.open > .merge-status-list {
	max-height: none;
```

jameschensmith: Your suggested change is what we have right now 😅 We could close this request if this is the desired route.

jameschensmith:
> On second thought:
> 
> * on the PR page, the section is part of the normal flow of the page, so I feel it's perfectly fine to leave it as _infinite._ It's already togglable, so it's doesn't make sense to limit it in anyway.
> * the popup’s awkwardly extends past the natural height of the page, so it makes sense to shorten that.
> 
> Maybe we can avoid changing the popup’s height altogether, and only leave `max-height: none` on the PR

Okay, so it sounds like we should just keep what we have for the PR page, and filter out the dropdown/popup. I understand your request 😊 👍 I'll do my best to figure that out and get the change in!

jameschensmith:
Updated the PR, title, and description to reflect the new request. This now covers addressing the popups/dropdowns, while keeping the PR merge status list infinite.

fregante:
Thank you James!

PR #4136, opened 2021-03-19
Restore `rgh-linkify-features` and `linkify-code` in titles

### LINKED ISSUE #4129: The *feature-linkification* _part_ of `format-conversation-titles` appears to be broken
This part:

https://github.com/sindresorhus/refined-github/blob/19330b2a4d1715c2d2496e6435a6b43263031c03/source/features/format-conversation-titles.tsx#L20-L22

On Chrome:
Screen Shot

URL: https://github.com/sindresorhus/refined-github/issues/4117 and this issue too

That part was already a little broken because it would be lost when the title updated, but it generally worked.

---

Also in the screenshot you can see a _very cool_ one-time `release-tab` bug

yakov116:
Fixes #4129

Test URL's

This PR

Linkify issues:
https://github.com/mozilla/addons-linter/pull/76


source/features/format-conversation-titles.tsx
```diff
@@ -29,10 +18,8 @@ void features.add(__filebasename, {
 		pageDetect.isPR,
 		pageDetect.isIssue
 	],
-	init() {
-		observeElement(select('#partial-discussion-header')!.parentElement!, init, {
-			subtree: true,
-			childList: true
-		});
```
yakov116, line 36: https://github.com/sindresorhus/refined-github/pull/3479 added `subtree` I don't see why it it needed

fregante: Likely for the same reason why `format-conversation-titles` handled parsed the backticks autonomously instead of using the `parse-backticks` feature: The title is changed by changing a single text node, which isn't a direct child of `select('#partial-discussion-header')!.parentElement!`

yakov116: But git hub refreshes' the entire parent element on change. We are using it as a call back. I don't see why subtree is needed.

fregante: If that's the case now, it must have changed: https://github.com/sindresorhus/refined-github/pull/3410#issuecomment-665104765

I still see that exact piece of code though:

Screen Shot

fregante: Is it possible that it only triggers when the title is not markdown? `textContent` would seem to suggest that.

yakov116: I tried almost every event https://github.com/yakov116/pr-deployments/pull/1 and I cannot get it not to work

fregante: Alright let's assume that they changed how the title updates, since https://github.com/sindresorhus/refined-github/issues/4129#issuecomment-802688144, and that this code path is just not activated (it has an existence check in the received JSON)

readme.md
```diff
@@ -318,7 +318,7 @@ Thanks for contributing! 🦋🙌
 - [](# "close-out-of-view-modals") [Automatically closes dropdown menus when they’re no longer visible.](https://user-images.githubusercontent.com/1402241/37022353-531c676e-2155-11e8-96cc-80d934bb22e0.gif)
 - [](# "deprioritize-marketplace-link") Moves the "Marketplace" link from the black header bar to the profile dropdown.
 - [](# "parse-backticks") [Renders `` `text in backticks` `` in issue titles, commit titles and more places.](https://user-images.githubusercontent.com/170270/55060505-31179b00-50a4-11e9-99a9-c3691ba38d66.png)
-- [](# "format-conversation-titles") [Makes issue/PR references in issue/PR titles clickable and renders `` `text in backticks` ``.](https://user-images.githubusercontent.com/22439276/58927232-71ae2780-876b-11e9-941e-bb56a7389123.png)
+- [](# "linkify-issues-in-conversation-titles") [Makes issue/PR references in issue/PR titles clickable.](https://user-images.githubusercontent.com/22439276/58927232-71ae2780-876b-11e9-941e-bb56a7389123.png)
```
fregante, line 321: I wonder if we should just merge this into `linkify-issues` at this point

source/features/rgh-linkify-features.tsx
```diff
@@ -28,4 +34,13 @@ void features.add(__filebasename, {
 		isNotRefinedGitHubRepo
 	],
 	init
+}, {
+	include: [
+		pageDetect.isPR,
+		pageDetect.isIssue
+	],
```
yakov116, line 41:
```suggestion
	],
	exclude: [
		isNotRefinedGitHubRepo
	],
```

fregante (approved):
Untested but LGTM

fregante:
I checked the listener that has been modified in https://github.com/sindresorhus/refined-github/pull/4089 and it appears to work correctly (i.e. it's run the right number of times)

Also I found which part still uses `textContent` until it receives the _markdown_ version of the title:

gif

yakov116:
> Also I found which part still uses `textContent` until it receives the _markdown_ version of the title:

Now I see it, is it an issue? Also we need to anyways wait for GitHub to format it since we need it to in `<code/>`

fregante:
No, it can be merged as is. This is an issue on GitHub's part. We don't deal with titles anymore. Actually it's best to drop that selector from `parse-backticks` as soon as possible.

yakov116:
> Actually it's best to drop that selector from [`parse-backticks`](/sindresorhus/refined-github/blob/main/source/features/parse-backticks.tsx) as soon as possible.

Only `.js-issue-title` needed to be removed

fregante:
Are you sure? That selector only matches the conversation title in lists, but conversation titles appear in all sorts of places with different selectors, like the newsfeed, right? We surely have more than one selector

yakov116:
> Are you sure? That selector only matches the conversation title in lists, but conversation titles appear in all sorts of places with different selectors, like the newsfeed, right? We surely have more than one selector

We sure do. I thought you wanted me to drop only things related to the title.

If i have time I will do it tomorrow

PR #4138, opened 2021-03-19
Update dependencies

yakov116:
No plans on merging before next week.

I may turn this into a lint by doing onConversationHeaderUpdate here, then updating the other prs.

WDYT?


fregante:
Blocked by https://github.com/sindresorhus/refined-github/pull/4116
Blocked by https://github.com/sindresorhus/refined-github/pull/4115

fregante:
> I may turn this into a lint by doing onConversationHeaderUpdate here, then updating the other prs.

👍 

I avoided updating the textarea dependencies because they need to be tested a little bit. `fit-textarea` completely changed mechanism in v3

yakov116:
> I avoided updating the textarea dependencies because they need to be tested a little bit. `fit-textarea` completely changed mechanism in v3

I noticed a bug will open an issue there. (it does not resize when content is changed programmatically)

fregante:
> it does not resize when content is changed programmatically

That can never happen, changing the content programmatically does not trigger any events.

package.json
```diff
@@ -54,19 +54,19 @@
 		"debounce-fn": "^5.0.0",
 		"delay": "^5.0.0",
 		"delegate-it": "^2.0.2",
-		"dom-chef": "^4.2.2",
+		"dom-chef": "^5.0.0",
```
fregante, line 57: https://github.com/vadimdemedes/dom-chef/releases/tag/v5.0.0

We probably need to drop some types as well

webpack.config.ts
```diff
@@ -96,7 +95,6 @@ const config: Configuration = {
 		]
 	},
 	plugins: [
-		new ESBuildPlugin(),
```
fregante, line 99: ?

Is it picked up automatically now? I see the loader still in the dependencies

package.json
fregante: Heh, I see you already did. I wonder how come it works though, the new types in dom-chef don't include SVG

fregante:
I merged #4116 into this. It's now unblocked

webpack.config.ts
yakov116: We still use the loader, just there is no need to do `new ESBuildPlugin()`

yakov116:
@fregante can do a npm install so we can I dont mess you up with fs events?

fregante:
> @fregante can do a npm install so we can I dont mess you up with fs events?

No worries, just do it. And I'll keep either re-adding or ignoring fsevent diffs 😂

yakov116:
Are we good to merge? I want to finish up the other 2 PR's

Issue #4144, opened 2021-03-21
Change the `collapse-wiki-sections` trigger


mathieucaroff:
Consider changing the collapsing behavior trigger to something else than the full title. I believe showing a collapse arrow to the left of the title on hover, next to the anchor icon would be a good idea. Like any other collapse arrow, it would have to change to horizontal direction when the section is collapsed.

Possibly, the arrow would stay visible when the title is collapse, so as to inform the user that some content is hidden.

Mildly related #4126

Thank you for making Refined Github

mathieucaroff:
+ I temporarily disabled `collapse-markdown-sections` because it reacts when I click on the anchor symbol. A behavior hopefully solved by the above linked issue's closing pull request.

mathieucaroff:
+ I realized the PR #4127 has a lot more discussions in it than the issue itself:

> _@fregante in https://github.com/sindresorhus/refined-github/issues/4127#issuecomment-803179552_
>
> Also I think we should add a cursor via CSS to show that clicking does _something_, probably:
>
> ```css
> .markdown-body :is(h1, …) {cursor: expand-y or whatever}
> .markdown-body :is(a, details) :is(h1, …) {cursor: initial}
> ```

> _@kidonng in https://github.com/sindresorhus/refined-github/issues/4127#issuecomment-803193226_
>
> Indeed. Currently it doesn't give any visual clue what has happened if one triggers the feature without knowing it. 
>
> In fact, I only realized this feature is implemented after I accidentally triggered it 😅

kidonng:
> + I temporarily disabled `collapse-markdown-sections` because it reacts when I click on the anchor symbol. A behavior hopefully solved by the above linked issue's closing pull request.

Unfortunately I think we missed that in the mentioned PR.

> I believe showing a collapse arrow to the left of the title on hover, next to the anchor icon would be a good idea.

Ideally we should do that (maybe like `<details>`), but since GitHub is already showing an anchor link for the titles, would an additional arrow looks weird?

fregante:
Buttons could be added to the left of the anchor button, to be shown on hover, only when the window is wide enough. 

However this issue is something that being already taken care of by https://github.com/sindresorhus/refined-github/pull/4127, not yet released

deflock:
Yes, a bit annoying behavior. I click sometimes in random places on the page to make window focused and often these clicks collapse sections :(

fregante:
PR welcome to fix the 2 immediate issues:

[...]

PR #4153, opened 2021-03-25
Show cursor in `collapse-markdown-sections`

yakov116:
#4144 (not sure if we want to close it)

## Test URLs
https://github.com/kidonng/api

I don't know what exactly I am testing for the details. I just followed instructions so we can put out a release

## Screenshot

image


fregante (approved):
Untested. As long as the cursor correctly appears when it’s actually togglable it’s good. 

As for the details, I posted the exact HTML snippet to use in a markdown file. You can test that. 

I’ll be AFK for a while

fregante:
Only merge this PR before releasing, the rest isn’t urgent

fregante:
If https://github.com/sindresorhus/refined-github/issues/4155 is just a class change, can you also make that change before releasing?

yakov116:
> If #4155 is just a class change, can you also make that change before releasing?

ok

yakov116:
image

Blue box has correct cursor red does not. Is that correct?

fregante:
Here's a test URL for the `header in details` issue: https://github.com/sindresorhus/refined-github/blob/Slash/slash/readme.md

fregante:
> Blue box has correct cursor red does not. Is that correct?

No. We should not change that cursor since it's still natively clickable, it's just not managed by us

source/features/collapse-markdown-sections.css
```diff
@@ -0,0 +1,7 @@
+.markdown-body > :is(h1, h2, h3, h4, h5, h6) {
+	cursor: pointer;
+}
+
+.markdown-body :is(a, details) :is(h1, h2, h3, h4, h5, h6) {
```
fregante, line 5: > Blue box has correct cursor red does not. Is that correct?

No. We should not change that cursor since it's still natively clickable, it's just not managed by us

I think we can just drop `details` from here since they always show `pointer` anyway, even if we don't handle it
```suggestion
.markdown-body a :is(h1, h2, h3, h4, h5, h6) {
```

source/features/collapse-markdown-sections.css
```diff
@@ -0,0 +1,7 @@
+.markdown-body > :is(h1, h2, h3, h4, h5, h6) {
```
fregante, line 1: This style will be applied even if the feature is disabled.

After both of these changes are applied, it can be merged

fregante:
:shipit:

source/features/collapse-markdown-sections.tsx
```diff
@@ -36,6 +37,7 @@ function onHeadingClick(event: delegate.Event<MouseEvent, HTMLElement>): void {
 
 function init(): void {
 	delegate(document, '.markdown-body > :is(h1, h2, h3, h4, h5, h6)', 'click', onHeadingClick);
+	document.body.classList.add('.rgh-' + __filebasename);
```
kidonng, line 40:
```suggestion
document.body.classList.add('rgh-' + __filebasename);
```

yakov116: OMG LOL

PR #4168, opened 2021-03-29
Fix `more-dropdown` removing repo tabs container

### LINKED ISSUE #4166: 👋👋👉👉All repository tabs invisible (FIXED)
Screen Shot 2021-03-29 at 9 48 58 AM

This just started happening minutes ago.

```
❌ Refined GitHub → releases-tab → TypeError: Cannot read property 'append' of undefined
    at init (refined-github.js:3835)
    at async runFeature (refined-github.js:1902)
    at async setupPageLoad (refined-github.js:1915) 
Search issue: https://github.com/sindresorhus/refined-github/issues?q=is%3Aissue+Cannot%20read%20property%20'append'%20of%20undefined
```

sharkykh:
Fixes []() #4166 

Looks like GitHub changed their layout:
The `[data-tab-item$="${id}"]` selector now points to the tab anchor/link.

image

### DIFF
--- a/source/features/more-dropdown.tsx
+++ b/source/features/more-dropdown.tsx
@@ -43,7 +43,7 @@ function onlyShowInDropdown(id: string): void {
 		return;
 	}
 
-	tabItem!.parentElement!.remove();
+	tabItem!.remove();
 	const menuItem = select(`[data-menu-item$="${id}"]`)!;
 	menuItem.removeAttribute('data-menu-item');
 	menuItem.hidden = false;
--- a/source/features/more-dropdown.tsx
+++ b/source/features/more-dropdown.tsx
@@ -43,7 +43,7 @@ function onlyShowInDropdown(id: string): void {
 		return;
 	}
 
-	tabItem!.remove();
+	tabItem!.closest('.UnderlineNav-item')!.remove();
 	const menuItem = select(`[data-menu-item$="${id}"]`)!;
 	menuItem.removeAttribute('data-menu-item');
 	menuItem.hidden = false;


source/features/more-dropdown.tsx
```diff
@@ -43,7 +43,7 @@ function onlyShowInDropdown(id: string): void {
 		return;
 	}
 
-	tabItem!.parentElement!.remove();
+	tabItem!.remove();
```
sharkykh, line 46: Seems like the `[data-tab-item$="${id}"]` selector now points to the tab anchor/link.
image

cheap-glitch, line 46: Maybe something like
```suggestion
	tabItem!.closest('[role="tab"]')!.remove();
```
could avoid this next time GH changes its DOM layout.
@fregante thoughts?

sharkykh: I had no issues with what you suggested originally, but since you edited your comment, your suggestion is now wrong.
Selecting the closest `.UnderlineNav-item` will cause the issue we have right now.

cheap-glitch: It does work, I wouldn't post something without having tested it first :)

Both my suggestions are equivalent since the targeted element is the same — and is actually `tabItem` itself, so in this case it's a no-op. It's just in case GH decides to wrap the element bearing `[data-tab-item]` again.

sharkykh: My bad, sorry, got it confused with `UnderlineNav-body` class.

yakov116:
can someone look into why we are seeing double before I merge this?

image

j-f1:
Also seeing that now with more-dropdown disabled

yakov116:
> Also seeing that now with more-dropdown disabled

Thats why its called `more-dropdown`:) When disabled you get more JK

yakov116:
Using https://github.com/sindresorhus/refined-github/pull/4119 looks like its `bugs-tab`

fregante:
Is that due to RG?

yakov116:
#4169 fixed the issue. Merged

yakov116 (approved):
Works for me

fregante:
Thanks for fixing this! Too bad I wasn’t at the computer today 🤷‍♂️ happens every time, it seems.

source/features/more-dropdown.tsx
```diff
@@ -43,7 +43,7 @@ function onlyShowInDropdown(id: string): void {
 		return;
 	}
 
-	tabItem!.parentElement!.remove();
+	tabItem!.closest('.UnderlineNav-item')!.remove();
```
fregante, line 46: They changed it again. Does this work? I'm still AFK. Please someone send a PR before v21.4.1 is released

```suggestion
	tabItem!.closest('.UnderlineNav-item, li')!.remove();
```

PR #4169, opened 2021-03-29
Fix `releases-tab` and `bugs-tab`

cheap-glitch:
## Test URLs
Any repo page

## Screenshot
n/a

GitHub changed the layout of the navbar, now it's no longer a list but instead a simple collection of anchor elements. The anchors act both as links and as tabs, so they bear all the meta attributes directly.

### DIFF
--- a/source/features/bugs-tab.tsx
+++ b/source/features/bugs-tab.tsx
@@ -68,8 +68,8 @@ async function init(): Promise<void | false> {
 
 		// Un-select one of the tabs if necessary
 		const selectedTab = !isBugsPage || pageDetect.isPRList() ? bugsTab : issuesTab;
-		selectedTab?.classList.remove('selected');
-		selectedTab?.removeAttribute('aria-current');
+		selectedTab.classList.remove('selected');
+		selectedTab.removeAttribute('aria-current');
 	} else {
 		// Pre "Repository refresh" layout
 		select('[itemprop="name"]', bugsTab)!.textContent = 'Bugs';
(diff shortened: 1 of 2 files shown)


source/features/bugs-tab.tsx
```diff
@@ -87,7 +86,7 @@ async function init(): Promise<void | false> {
 	bugsCounter.title = '';
 
 	// Update Bugs’ link
-	new SearchQuery(bugsLink).add('label:bug');
+	new SearchQuery(bugsTab as HTMLAnchorElement).add('label:bug');
```
FloEdelmann, line 89: Can you drop this type cast once you change the `issuesTab` selector to `a.UnderlineNav-item[data-hotkey="g i"]` (i.e. including the tag name `a`)?

cheap-glitch: I thought so too, but unfortunately it seems that `cloneNode` loses the element type: https://github.com/Microsoft/TypeScript/issues/283

FloEdelmann: Seems like the workaround added in #2670 doesn't work in this case?

cheap-glitch: No it does actually, I just forgot the return value of `elementReady()` needed to be coerced. My bad.

fregante (approved):

source/features/bugs-tab.tsx
fregante: There’s an issue open on the element-ready repo to implement the selector parser for types, if anyone is interested. Relatively easy PR for that

JourneyOver:
Looks like it's back to being a list once again xD

PR #4180, opened 2021-03-31
Fix `more-dropdown`

yakov116:


source/features/more-dropdown.tsx
```diff
@@ -43,7 +43,7 @@ function onlyShowInDropdown(id: string): void {
 		return;
 	}
 
-	tabItem!.closest('.UnderlineNav-item')!.remove();
+	tabItem!.closest('.UnderlineNav-item,li')!.remove();
```
fregante, line 46:
```suggestion
	tabItem!.closest('.UnderlineNav-item, li')!.remove();
```

fregante:
Can be merged if it works

PR #4181, opened 2021-03-31
Fix features related to repo tabs again

cheap-glitch:
## Test URLs
* `more-dropdown`
  * https://github.com/microsoft/TypeScript
  * https://github.com/eslint/eslint
* `bugs-tab`
  * https://github.com/microsoft/TypeScript
  * [No issue labels](https://github.com/PHPOffice/PhpSpreadsheet)
  * [Archived repo](https://github.com/PHPOffice/PHPExcel)
* `releases-tab`
  * https://github.com/atom-community/sync-settings
  * [Tags instead or releases](https://github.com/tpope/vim-unimpaired)

## Screenshot
n/a

Seems like #4180 wasn't actually needed since the `<a>` is still the one bearing the `.UnderlineNav-item` class:

image

This PR:
  * proposes a hopefully better fix that works for both GitHub and GHE and doesn't leave empty `<li>` in the DOM
  * changes `bugs-tab` and `releases-tab` to follow the new DOM structure

### DIFF
--- a/source/features/more-dropdown.tsx
+++ b/source/features/more-dropdown.tsx
@@ -43,11 +43,11 @@ function onlyShowInDropdown(id: string): void {
 		return;
 	}
 
-	tabItem!.closest('.UnderlineNav-item, li')!.remove();
+	(tabItem!.closest('li') ?? tabItem!.closest('.UnderlineNav-item'))!.remove();
+
 	const menuItem = select(`[data-menu-item$="${id}"]`)!;
 	menuItem.removeAttribute('data-menu-item');
 	menuItem.hidden = false;
-
 	// The item has to be moved somewhere else because the overflow nav is order-dependent
 	select('.js-responsive-underlinenav-overflow ul')!.append(menuItem);
 }
--- a/source/features/bugs-tab.tsx
+++ b/source/features/bugs-tab.tsx
@@ -88,7 +88,12 @@ async function init(): Promise<void | false> {
 	// Update Bugs’ link
 	new SearchQuery(bugsTab).add('label:bug');
 
-	issuesTab.after(bugsTab);
+	// In case GitHub changes its layout again #4166
+	if (issuesTab.parentElement!.tagName === 'LI') {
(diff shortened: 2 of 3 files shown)


cheap-glitch (commented):
```typescript
		appendBefore(
			select('.js-responsive-underlinenav .dropdown-menu ul')!,
			'.dropdown-divider', // Won't exist if `more-dropdown` is disabled
			createDropdownItem('Releases', buildRepoURL('releases'), {
				'data-menu-item': 'rgh-releases-item'
			})
		);
```
Seems like this always add a "Releases" entry in the dropdown. Is that really needed/intended?

source/features/more-dropdown.tsx
```diff
@@ -43,11 +43,11 @@ function onlyShowInDropdown(id: string): void {
 		return;
 	}
 
-	tabItem!.closest('.UnderlineNav-item, li')!.remove();
+	(tabItem!.closest('li') ?? tabItem!.closest('.UnderlineNav-item'))!.remove();
```
fregante, line 46: It didn’t work?

cheap-glitch: If you're talking about #4180: it's not that it didn't work, it's that it's not needed because nothing actually broke in the first place.

This change removes the `<li>` instead of leaving it empty, and it'll work should the devs at GitHub change their minds again.

fregante:
@yakov116 let's test this out so we can merge it together with my PR as soon as possible (tomorrow?)

yakov116 (commented):
> Seems like #4180 wasn't actually needed since the `<a>` is still the one bearing the `.UnderlineNav-item` class:

You are correct. I think what happened was when I tested it I was checking out a branch that was not on master yet (before your other fix) or something like that.

@fregante I tested it, its working great.

I am out for the weekend (will still have some access today), So tested this PR in every way possible. Looks good to me

SekiBetu:
tested, working great.
finally, everything is back.

Issue #4183, opened 2021-03-31
"Search in files" button does not cancel navigation


nao20010128nao:
# "Search in files" button does not cancel navigation
This problem happens in any repository, with URL like: https://github.com/sindresorhus/refined-github

Here's the video showing this behavior:

https://user-images.githubusercontent.com/10355528/113172844-cba6a800-9283-11eb-92ec-dd5782dd02f6.mp4

This also happens with <kbd>t</kbd> key, which is equivalent to push the button.

Thanks.

fregante:
You’re not supposed to click that, it’s already focused. Just start typing immediately

PR #4184, opened 2021-03-31
Highlight `bugs-tab` on bug issues

### LINKED ISSUE #4164: Incorrect tab gets highlighted when opening an issue in the bugs section
## STR:
1. Goto https://github.com/sindresorhus/refined-github/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3Abug.
2. Notice that the tab highlighted on the top is Bugs
3. Open any active issue, the tab highlighted switches to Issues.

Shouldn't the tab highlighted be the bugs tab?

## Video:  

https://user-images.githubusercontent.com/31387795/112757564-e371f800-9007-11eb-98a9-ed645fd6b30b.mov

yakov116:
Fixes #4164 

## Test URLs
Any Bug issue
https://github.com/sindresorhus/refined-github/issues/4183

## Screenshot

image

### DIFF
--- a/source/features/bugs-tab.tsx
+++ b/source/features/bugs-tab.tsx
@@ -11,8 +11,10 @@ import {getRepo} from '../github-helpers';
 import SearchQuery from '../github-helpers/search-query';
 import abbreviateNumber from '../helpers/abbreviate-number';
 
+const cacheKey = (): string => __filebasename + ':' + getRepo()!.nameWithOwner;
+
 async function highlightBugsTabOnIssuePage(): Promise<void | false> {
-	if (!await cache.get<boolean>(__filebasename + ':' + getRepo()!.nameWithOwner)) {
+	if (!await cache.get<boolean>(cacheKey())) {
 		return false;
 	}
 
@@ -38,7 +40,7 @@ const countBugs = cache.function(async (): Promise<number> => {
 }, {
 	maxAge: {minutes: 30},
 	staleWhileRevalidate: {days: 4},
-	cacheKey: (): string => __filebasename + ':' + getRepo()!.nameWithOwner
+	cacheKey: (): string => cacheKey()
 });
 
 async function init(): Promise<void | false> {
--- a/source/features/bugs-tab.tsx
+++ b/source/features/bugs-tab.tsx
@@ -1,6 +1,5 @@
 import React from 'dom-chef';
 import cache from 'webext-storage-cache';
-import delay from 'delay';
 import select from 'select-dom';
 import {BugIcon} from '@primer/octicons-react';
 import elementReady from 'element-ready';
@@ -17,8 +16,7 @@ async function highlightBugsTabOnIssuePage(): Promise<void | false> {
 		return false;
 	}
 
-	await delay(1); // Wait until the bug tab is created
-	const bugsTab = select('.rgh-bug-tab');
+	const bugsTab = await elementReady('.rgh-bug-tab', {stopOnDomReady: false});
 	bugsTab!.classList.add('selected');
 
 	const issuesTab = select('.UnderlineNav-item[data-hotkey="g i"]')!;
--- a/source/features/bugs-tab.tsx
+++ b/source/features/bugs-tab.tsx
@@ -1,5 +1,6 @@
 import React from 'dom-chef';
 import cache from 'webext-storage-cache';
+import delay from 'delay';
 import select from 'select-dom';
 import {BugIcon} from '@primer/octicons-react';
 import elementReady from 'element-ready';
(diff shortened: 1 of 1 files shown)


source/features/bugs-tab.tsx
```diff
@@ -11,6 +11,24 @@ import {getRepo} from '../github-helpers';
...
 import abbreviateNumber from '../helpers/abbreviate-number';
 
+const cacheKey = (): string => __filebasename + ':' + getRepo()!.nameWithOwner;
+
+async function highlightBugsTabOnIssuePage(): Promise<void | false> {
+	if (!await cache.get<boolean>(cacheKey())) {
+		return false;
+	}
+
+	const sideBarLables = await elementReady('#partial-discussion-sidebar .sidebar-labels');
```
fregante, line 21:
```suggestion
	const sidebarLables = await elementReady('#partial-discussion-sidebar .sidebar-labels');
```

source/features/bugs-tab.tsx
```diff
@@ -11,6 +11,24 @@ import {getRepo} from '../github-helpers';
 import SearchQuery from '../github-helpers/search-query';
 import abbreviateNumber from '../helpers/abbreviate-number';
 
+const cacheKey = (): string => __filebasename + ':' + getRepo()!.nameWithOwner;
+
+async function highlightBugsTabOnIssuePage(): Promise<void | false> {
+	if (!await cache.get<boolean>(cacheKey())) {
```
fregante, line 17: Use `countBugs()`, it's a memoized promise so it'll work without race conditions.

source/features/bugs-tab.tsx
```diff
@@ -11,6 +11,24 @@ import {getRepo} from '../github-helpers';
...
+		return false;
+	}
+
+	const sideBarLables = await elementReady('#partial-discussion-sidebar .sidebar-labels');
+	const hasBugLabel = select.exists('.IssueLabel[href$="/bug" i]', sideBarLables);
```
fregante, line 22: You can merge these. Since _change_ is only needed if **it exists**, waiting until dom-ready is fine. 

Likewise, generally I assume that if an issue is a bug _now_, it will still be a bug when you go _back_ to it, so the  `selected` tab doesn't necessarily need to be updated.

Maybe just:
```js
	if (!await elementReady('.IssueLabel[href$="/bug" i]')) {
		return false
	}
```

source/features/bugs-tab.tsx
```diff
@@ -11,6 +11,24 @@ import {getRepo} from '../github-helpers';
...
+
+	const bugsTab = select('.rgh-bug-tab');
+	bugsTab!.classList.toggle('selected', hasBugLabel);
+
+	const issuesTab = select('.UnderlineNav-item[data-hotkey="g i"]')!;
+	issuesTab.classList.toggle('selected', !hasBugLabel);
```
fregante, line 28:
```suggestion
	bugsTab!.classList.add('selected');

	const issuesTab = select('.UnderlineNav-item[data-hotkey="g i"]')!;
	issuesTab.classList.remove('selected');
```

source/features/bugs-tab.tsx
```diff
@@ -11,6 +12,20 @@ import {getRepo} from '../github-helpers';
 import SearchQuery from '../github-helpers/search-query';
 import abbreviateNumber from '../helpers/abbreviate-number';
 
+async function highlightBugsTabOnIssuePage(): Promise<void | false> {
+	if (await countBugs() > 0 && !await elementReady('.sidebar-labels .IssueLabel[href$="/bug" i]')) {
+		return false;
+	}
+
+	await delay(1); // Wait until the bug tab is created
```
fregante, line 20: This is arbitrary. We should use `select(tab) ?? await oneMutation(tabBar, filter)`

yakov116: What about
```suggestion
	await elementReady('.rgh-bug-tab', {waitForDomReady: false})
```

This is really the same thing as oneMutation

fregante: hmmm this might be just about the only case where `waitForDomReady: false` is acceptable, since we handle everything in this file. I mean, not ideal, still, but maybe ok enough.

source/features/bugs-tab.tsx
```diff
@@ -11,6 +11,19 @@ import {getRepo} from '../github-helpers';
 import SearchQuery from '../github-helpers/search-query';
 import abbreviateNumber from '../helpers/abbreviate-number';
 
+async function highlightBugsTabOnIssuePage(): Promise<void | false> {
+	if (await countBugs() > 0 && !await elementReady('.sidebar-labels .IssueLabel[href$="/bug" i]')) {
+		return false;
+	}
+
+	const bugsTab = await elementReady('.rgh-bug-tab', {stopOnDomReady: false});
```
fregante, line 19: Probably


```suggestion
	const bugsTab = await elementReady('.rgh-bug-tab', {stopOnDomReady: false, timeout: 10000});
```

fregante: Can be merged afterwards

Issue #4178, opened 2021-03-30
View File Broken in Firefox


brian-avery:
It seems that the view file link is broken on GitHub in Firefox. For instance:

* Go here: https://github.com/kubernetes/kubernetes/pull/100658/files
* Click on the 3 dots on the first file to open the menu.
* Click view file.

Instead of directing you to the file in the branch, it looks like it's skipping the branch in the URL (github.com/AlexeyPerevalov/kubernetes/blob//pkg/scheduler/testing/wrappers.go)

This results in a 404
![image](https://user-images.githubusercontent.com/8484260/113029308-45db0d80-915a-11eb-985e-b9372a174f17.png)

If I populate the branch in the URL: github.com/AlexeyPerevalov/kubernetes/blob/AddLimitsToReq/pkg/scheduler/testing/wrappers.go then it works.

If I disable refined-github, the link starts working

Looking in the console, I see: 

```
state is null global.js:344
[Exception... "Component returned failure code: 0x80070057 (NS_ERROR_ILLEGAL_VALUE) [nsIDOMWindowUtils.addSheet]"  nsresult: "0x80070057 (NS_ERROR_ILLEGAL_VALUE)"  location: "JS frame :: resource://gre/modules/ExtensionCommon.jsm :: runSafeSyncWithoutClone :: line 74"  data: no] ExtensionCommon.jsm:74:12
TypeError: i.default.detectStore(...) is undefined h1-check.js:1:1301
GEThttps://github.com/AlexeyPerevalov/kubernetes/blob//pkg/scheduler/testing/wrappers.go
[HTTP/2 404 Not Found 226ms]

[Exception... "Component returned failure code: 0x80070057 (NS_ERROR_ILLEGAL_VALUE) [nsIDOMWindowUtils.addSheet]"  nsresult: "0x80070057 (NS_ERROR_ILLEGAL_VALUE)"  location: "JS frame :: resource://gre/modules/ExtensionCommon.jsm :: runSafeSyncWithoutClone :: line 74"  data: no] ExtensionCommon.jsm:74:12
TypeError: i.default.detectStore(...) is undefined h1-check.js:1:1301
XHRHEADhttps://github.com/AlexeyPerevalov/kubernetes/tree//pkg/scheduler/testing
[HTTP/2 404 Not Found 178ms]

XHRHEADhttps://github.com/AlexeyPerevalov/kubernetes/tree//pkg/scheduler
[HTTP/2 404 Not Found 171ms]

[...]

PR #4187, opened 2021-04-01
Further improve branch getter function

### LINKED ISSUE #4196: `follow-file-renames` should handle branch names with slashes, but doesn't
https://github.com/sindresorhus/refined-github/commits/Slash/slash?after=ee8b6422d2ddd1199f557594d03e7ee850e843f0+34&branch=Slash%2Fslash&path%5B%5D=webpack.config.ts

As a consequence of recent changes outlined in #4179 and fixed in #4193, slashed branches are not currently supported in commits pages. 

#4187 might fix it.

Follows #4193

yakov116:
Fixes #4196 
Fixes https://github.com/sindresorhus/refined-github/issues/4271


source/github-helpers/index.ts
```diff
@@ -24,12 +24,13 @@ Example tag content on private repositories https://github.com/private/private/c
 export const getCurrentBranch = (): string | undefined => {
 	// .last needed for #2799
 	const feedLink = select.last('link[type="application/atom+xml"]');
+	const findLink = pageDetect.isPRFiles() ? false : select.last('a[data-hotkey="t"]');
```
yakov116, line 27: .last is needed

fregante:
> @fregante please let me know if we should stop the next release

Good idea, disabled.

> Its not in the head of the page so...

That kills the cat.

Ensure 2 things:

- If the element isn't found, the feature still works, somehow
- If the element isn't found, use `console.warn` to alert us

source/github-helpers/index.ts
```diff
@@ -24,12 +24,13 @@ Example tag content on private repositories https://github.com/private/private/c
 export const getCurrentBranch = (): string | undefined => {
 	// .last needed for #2799
 	const feedLink = select.last('link[type="application/atom+xml"]');
```
fregante, line 26: Feels like this shouldn't be missing. I hope it's temporary

yakov116:
> * If the element isn't found, the feature still works, somehow

This is going to be very hard its used in 15 features.

source/github-helpers/index.ts
```diff
@@ -24,12 +24,13 @@ Example tag content on private repositories https://github.com/private/private/c
...
+	const findLink = pageDetect.isPRFiles() ? false : select.last('a[data-hotkey="t"]');
 	// The feedLink is not available on `isIssue` #3641
-	if (!feedLink) {
+	if (!(feedLink && findLink)) {
 		return;
 	}
 
-	return new URL(feedLink.href)
+	return new URL(feedLink.href ?? decodeURIComponent(findLink.href))
```
fregante, line 33: ??

Both are elements, why would .href be encoded only in one

source/github-helpers/index.ts
yakov116: I wish, but it was removed from so many pages that I think its gone 😢

source/github-helpers/index.ts
fregante: It looks like you should just be able to extend the existing selector, e.g.

```
const linkWithData = select.last('link[type="application/atom+xml"], a[data-hotkey="t"]');
```

source/github-helpers/index.ts
yakov116: `https://github.com/yakov116/TestR/tree/this/branch/has/many/slashes` => https://github.com/yakov116/TestR/find/this%2Fbranch%2Fhas%2Fmany%2Fslashes

source/github-helpers/index.ts
yakov116: We can't do that because we need to return false on `isPrFiles` and I dont want to break GHE

fregante:
Does this PR also fix https://github.com/sindresorhus/refined-github/issues/4178?

yakov116:
No that is `isPrFiles`

yakov116:
@fregante ok I figured out a way to make this DOM safe in most cases. The title has the word ' at ' in it on many pages.

But I dont know regex so I wrote it...

fregante:
One thing to note is that this function only really needs the branch to support branches with slashes. In every other case we really don’t care if it exists or not, it makes no difference (if I remember correctly). So in most cases the branch parser can return undefined and the feature should still work.

source/github-helpers/index.ts
fregante: I’m gonna ping some people on Twitter, hopefully someone answers

yakov116:
> One thing to note is that this function only really needs the branch to support branches with slashes. In every other case we really don’t care if it exists or not, it makes no difference (if I remember correctly). So in most cases the branch parser can return undefined and the feature should still work.

I think you are correct

source/github-helpers/index.ts
fregante: YO, YOU SERIOUS?

This means that the branch name with slashes is FINALLY encoded so we don’t need this whole thing at all?!

source/github-helpers/index.ts
yakov116: Yes

source/github-helpers/index.ts
```diff
@@ -22,14 +22,25 @@ Example tag content on public repositories: https://github.com/sindresorhus/refi
 Example tag content on private repositories https://github.com/private/private/commits/master.atom?token=AEAXKWNRHXA2XJ2ZWCMGUUN44LM62
 */
 export const getCurrentBranch = (): string | undefined => {
+	// Example title: 'tejanium/refined-github:bra/nch' or just 'local-branch`
+	if (pageDetect.isPRFiles()) {
```
fregante, line 26: This weird. The comment above specifically mentions "Subtly incompatible with isPR", but we're still using it there. Maybe `getCurrentBranch` should just _not_ be used there since PRs have 2 branches anyway, this function is too generic.

Anyway, I'm working on this

yakov116: > "Subtly incompatible with isPR"

What that means is that on PR's the current branch can also be a commit.

> Anyway, I'm working on this

👍

fregante:
Probably my PR will have to be integrated with this. I visited https://github.com/eslint/eslint and there's zero mention of `master` in the raw HTML request, which means everything is deferred. 

Generally this is not a problem, but `getCurrentBranch` is used by `getDefaultBranch` so either we await the DOM in that case or just let it do a query.

fregante:
> just let it do a query

That's what #4190 is doing at the moment

source/github-helpers/index.ts
fregante: Apparently they still don't do it in most cases, but I'm hopeful.

Related: https://github.com/fregante/shorten-repo-url/issues/27

fregante:
Parts of this have been added by #4193. Here I'm hoping that we can use the DOM to fill in the details in the remaining case:

- slashed branches in commits pages, like https://github.com/sindresorhus/refined-github/issues/4196

This is because on that page:

- there's no branch selector
- there's no branch mentioned in the title

yakov116:
Oh man this is not going to be a fun merge

fregante:
Sorry! I left it open to avoid closing it altogether, but I think it's best to start over and just copy-paste anything new

source/github-helpers/index.ts
fregante: As an update, the feed is there, but only on iscommits

source/github-helpers/index.ts
```diff
@@ -53,6 +53,10 @@ export const getCurrentBranchFromFeedLink = (): string | undefined => {
 };
 
 export const getCurrentBranchFromFindLink = async (): Promise<string | undefined> => {
+	if (pageDetect.isPRFiles()) {
+		return;
+	}
```
fregante, line 58: Let’s add a comment explaining why. I feel that a lot of code could be explained a little better. Not what it’s doing, but *why* some choices were made

source/github-helpers/index.ts
```diff
@@ -37,6 +37,34 @@ export const getCurrentCommittish = (pathname = location.pathname, title = docum
...
+	return new URL(feedLink.href)
+		.pathname
+		.split('/')
+		.slice(4) // Drops the initial /user/repo/route/ part
+		.join('/')
+		.replace(/\.atom$/, '');
+};
+
+export const getCurrentBranchFromFindLink = async (): Promise<string | undefined> => {
+	if (pageDetect.isPRFiles()) {
```
yakov116, line 56:
```suggestion
	// On `isPRFiles` the `t` shortcut refers to the search file dropdown  
	if (pageDetect.isPRFiles()) {
```

fregante: s/refers to/opens/

fregante: Actually, this link only natively exists on some pages, so we should just target them directly. I think they're isTree and is isSingleFile?

source/github-helpers/index.ts
```diff
@@ -37,6 +37,34 @@ export const getCurrentCommittish = (pathname = location.pathname, title = docum
...
+		return;
+	}
+
+	const feedLink = select.last('a[data-hotkey="t"]') ?? await elementReady('a[data-hotkey="t"]');
```
fregante, line 60: Why last? Why/**when** would there be 2 links with the same keyboard shortcut pointing to 2 different things?

source/github-helpers/index.ts
```diff
@@ -37,6 +37,34 @@ export const getCurrentCommittish = (pathname = location.pathname, title = docum
 	return unslashedCommittish;
 };
 
+export const getCurrentBranchFromFeedLink = (): string | undefined => {
+	// .last needed for #2799
+	const feedLink = select.last('link[type="application/atom+xml"]');
```
fregante, line 42: If you have time, let's verify that #2799 still applies and if not let's add a // GHE #4187

source/github-helpers/index.ts
yakov116: All pages that I have tested

source/github-helpers/index.ts
```diff
@@ -37,6 +37,34 @@ export const getCurrentCommittish = (pathname = location.pathname, title = docum
...
+	if (pageDetect.isPRFiles()) {
+		return;
+	}
+
+	const feedLink = select.last('a[data-hotkey="t"]') ?? await elementReady('a[data-hotkey="t"]');
+	if (!feedLink) {
+		return;
+	}
+
+	return decodeURIComponent(String(feedLink.pathname.split('/').slice(-1)));
```
fregante, line 65:
```suggestion
	return decodeURIComponent(feedLink.pathname.split('/').pop());
```

or 

```suggestion
	return decodeURIComponent(feedLink.pathname.split('/').pop()!);
```

test/fixtures/globals.ts
```diff
@@ -14,3 +14,4 @@ const {window} = new JSDOM('…');
 (global as any).URL = window.URL;
 
 document.head.insertAdjacentHTML('beforeend', '<link href="https://github.com/avajs/ava/commits/master.atom" rel="alternate" title="Recent Commits to ava:master" type="application/atom+xml">');
+document.head.insertAdjacentHTML('beforeend', '<a hidden="" data-hotkey="t" data-pjax="true" href="https://github.com/avajs/ava/find/master"></a>');
```
fregante, line 17: Does any test currently fail without this line? If not, maybe a new test should be added

yakov116: I do not think its needed any more. However leave the test for now I will come back to it later. I need to fully test this PR first. (later tonight)

source/github-helpers/index.ts
yakov116: > Also if .last can be removed, then we only need elementReady since it calls `select` internally immediately anyway

Got it.

source/github-helpers/index.ts
```diff
@@ -37,6 +37,34 @@ export const getCurrentCommittish = (pathname = location.pathname, title = docum
 	return unslashedCommittish;
 };
 
+export const getCurrentBranchFromFeedLink = (): string | undefined => {
```
fregante, line 40: I'm imagining this to be part of `getCurrentCommittish` itself since they do the same thing and they're both dom-ready-safe

source/github-helpers/index.ts
fregante: Actually… isn't this ever only needed on the root? Everything else on those pages is already covered by `document.title` in `getCurrentCommittish`

yakov116: I think you maybe right!

source/github-helpers/index.ts
yakov116: a[data-hotkey="t"] is there first

fregante: Selector order is irrelevant

yakov116: No I mean that elementReady always returned the hotkey first

yakov116:
@fregante can you take a look to make sure I am going in the right direction?

source/github-helpers/github-url.ts
```diff
@@ -37,17 +37,18 @@ export default class GitHubURL {
...
 		const branch = ambiguousReference[0];
+		const currentBranch = getCurrentCommittish();
 		// History pages might use search parameters
+		const branchFromSearch = String(this.searchParams.getAll('branch'));
 		const filePathFromSearch = this.searchParams.getAll('path[]').join('/');
 		if (filePathFromSearch) {
+			this.searchParams.delete('branch');
 			this.searchParams.delete('path[]');
-			return {branch, filePath: filePathFromSearch};
+			return {branch: branchFromSearch ?? currentBranch ?? branch, filePath: filePathFromSearch};
```
fregante, line 47: When would this link have a file path but not a branch? Steps to reproduce please

source/github-helpers/github-url.ts
```diff
@@ -37,17 +37,18 @@ export default class GitHubURL {
 		ambiguousReference: string[]
 	): {branch: string; filePath: string} {
 		const branch = ambiguousReference[0];
+		const currentBranch = getCurrentCommittish();
 		// History pages might use search parameters
+		const branchFromSearch = String(this.searchParams.getAll('branch'));
```
fregante, line 42: Use get instead of getAll. No `String()` either

test/get-current-committish.ts
```diff
@@ -0,0 +1,94 @@
...
+	t.throws(() => getCurrentCommittish(
+		'https://github.com',
+		'github.com'
+	));
+
+	// Root
+	t.is(getCurrentCommittish(
+		'/typescript-eslint/typescript-eslint',
+		'typescript-eslint/typescript-eslint: Monorepo for all the tooling which enables ESLint to support TypeScript'
+	), undefined);
```
fregante, line 22: This is the only case that needs to be changed by the way. You can fix this test by adding the extra element (the find link I think) on line 18 and remove it on line 27. To ensure it works correctly, also use a branch with slashes there

source/github-helpers/index.ts
```diff
@@ -34,9 +34,33 @@ export const getCurrentCommittish = (pathname = location.pathname, title = docum
...
+			.split('/')
+			.slice(4) // Drops the initial /user/repo/route/ part
+			.join('/')
+			.replace(/\.atom$/, '');
+	}
+
 	return unslashedCommittish;
 };
 
+export const getCurrentBranchFromFindLink = async (): Promise<string | undefined> => {
```
fregante, line 51:
```suggestion
export const getCurrentCommittishAnywhere = async (): Promise<string | undefined> => {
    const committish = getCurrentCommittish();
    if (committish) {
        return committish;
    }

```

This way, in places like more-dropdown, you only need to call this function + getCurrentBranch. The only reason for this function to exist at all outside getCurrentCommittish is that that one is domready safe and this one isn’t.

source/github-helpers/github-url.ts
yakov116: https://github.com/yakov116/refined-github/commits/f23b687b3b89aa95a76193722cdfeff740646670?after=f23b687b3b89aa95a76193722cdfeff740646670+34&path%5B%5D=source&path%5B%5D=features&path%5B%5D=release-download-count.tsx

test/get-current-committish.ts
yakov116: How would I add an element here? (I have no idea how tests work)

source/github-helpers/github-url.ts
fregante: Steps to reproduce. That's just a link that anyone can write manually but maybe doesn't exist

test/get-current-committish.ts
fregante: > How would I add an element here?

That's the line we were talking about in this PR:

```
document.head.insertAdjacentHTML('beforeend', '<a hidden="" data-hotkey="t" data-pjax="true" href="https://github.com/avajs/ava/find/master"></a>');
```

Instead of leaving it at the bottom of the page, move it to line 18 and then remove the element on line 27: document.querySelector('[data-hotkey="t"]')`

source/features/more-dropdown.tsx
```diff
@@ -59,9 +58,9 @@ async function init(): Promise<void> {
 		'.UnderlineNav-body'
 	].join());
 
-	const reference = getCurrentCommittish() ?? await getDefaultBranch();
-	const compareUrl = buildRepoURL('compare', reference);
-	const commitsUrl = buildRepoURL('commits', reference);
+	const reference = await getCurrentCommittishAnywhere();
```
fregante, line 61: Default branch is still needed, I mistyped that
```suggestion
	const reference = await getCurrentCommittishAnywhere() ?? await getDefaultBranch();
```

source/github-helpers/index.ts
```diff
@@ -34,9 +34,38 @@ export const getCurrentCommittish = (pathname = location.pathname, title = docum
...
+			.split('/')
+			.slice(4) // Drops the initial /user/repo/route/ part
+			.join('/')
+			.replace(/\.atom$/, '');
+	}
+
 	return unslashedCommittish;
 };
 
+export const getCurrentCommittishAnywhere = async (): Promise<string | undefined> => {
```
fregante, line 51: Can you add a comment explaining when to use this?

source/github-helpers/index.ts
```diff
@@ -34,9 +34,38 @@ export const getCurrentCommittish = (pathname = location.pathname, title = docum
...
+	const committish = getCurrentCommittish();
+	if (committish) {
+		return committish;
+	}
+
+	if (!pageDetect.isRepoHome()) {
+		return;
```
fregante, line 58: Maybe
```suggestion
	if (committish || !pageDetect.isRepoHome()) {
		return committish;
```

source/github-helpers/index.ts
```diff
@@ -34,9 +34,38 @@ export const getCurrentCommittish = (pathname = location.pathname, title = docum
...
+	}
+
+	const feedLink = await elementReady<HTMLAnchorElement>('a[data-hotkey="t"]')!;
```
fregante, line 61:
```suggestion
	const feedLink = await elementReady('a[data-hotkey="t"]')!;
```

source/github-helpers/index.ts
```diff
@@ -34,9 +34,38 @@ export const getCurrentCommittish = (pathname = location.pathname, title = docum
 		return parsedTitle.groups!.branch;
 	}
 
+	// .last needed for #2799 GHE #4187
```
fregante, line 37:
```suggestion
	// TODO [2021-04-12]: .last needed for #2799 // GHE
```

source/features/more-dropdown.tsx
yakov116: Why? I checked even without our feature I was unable to find a page that did not have it.  I ment to open an issue

source/github-helpers/github-url.ts
yakov116: That is the url we were testing until now. The branch in the path is a new thing.

source/github-helpers/index.ts
yakov116: Need to wait for elementReady to be updated first

source/features/more-dropdown.tsx
yakov116: > Why? I checked even without our feature I was unable to find a page that did not have it. I ment to open an issue

Oh my I disabled the wrong feature and I thought GitHub added it LOL

source/github-helpers/github-url.ts
fregante: That URL does not have a `branch` URL parameter

yakov116: https://github.com/sindresorhus/refined-github/commits/Slash/slash?after=ee8b6422d2ddd1199f557594d03e7ee850e843f0+34&branch=Slash%2Fslash&path%5B%5D=webpack.config.ts

yakov116: > That URL does not have a `branch` URL parameter (`branchFromSearch`). Where did you see that?

Oh my I just re read you comment 🤦 

That url is the one we are testing until now. GitHub recently starting adding the branch to the search

yakov116:
@fregante I tried last night again to figure this out, with no luck.

I don't care to close and let someone else attempt. Please do as you see fit.

source/features/latest-tag-button.tsx
```diff
@@ -110,7 +110,7 @@ async function init(): Promise<false | void> {
 	}
 
 	const defaultBranch = await getDefaultBranch();
-	if (currentBranch === defaultBranch) {
+	if (pageDetect.isRepoHome() || currentBranch === defaultBranch) {
```
yakov116, line 113: adsfdsafd

source/github-helpers/github-url.ts
fregante: Once again, I need steps to get that URL. I want a starting URL that isn’t what you’re pasting here. I can also make up a URL and send a comment, that doesn’t mean it exists anywhere else.

fregante:
What’s missing here other than the open reviews?

Also if it’s used by any feature (and if it doesn’t already work), on the file editing page there’s a new place to read the branch name from. https://twitter.com/natfriedman/status/1383169134096093187?s=21

yakov116:
> What’s missing here other than the open reviews?

The tests, I cant figure out how to make it pass.

source/features/more-dropdown.tsx
```diff
@@ -59,9 +58,9 @@ async function init(): Promise<void> {
 		'.UnderlineNav-body'
 	].join());
 
-	const reference = getCurrentCommittish() ?? await getDefaultBranch();
-	const compareUrl = buildRepoURL('compare', reference);
-	const commitsUrl = buildRepoURL('commits', reference);
+	const reference = await getCurrentCommittishAnywhere() ?? await getDefaultBranch();
+	const compareUrl = buildRepoURL('compare', reference!);
+	const commitsUrl = buildRepoURL('commits', reference!);
```
fregante, line 63:
```suggestion
	const compareUrl = buildRepoURL('compare', reference);
	const commitsUrl = buildRepoURL('commits', reference);
```

source/github-helpers/github-url.ts
fregante: This is what I wanted:

1. Visit https://github.com/sindresorhus/refined-github/commits/Slash/slash/webpack.config.ts
2. Click "Older"

source/features/index.tsx
```diff
@@ -78,7 +78,7 @@ const globalReady: Promise<RGHOptions> = new Promise(async resolve => {
 		return;
 	}
 
-	if (document.title === 'Confirm password') {
+	if (document.title === 'Confirm password' || document.title === 'Confirm access') {
```
fregante, line 81: Where does this appear?

source/github-helpers/github-url.ts
```diff
@@ -37,17 +37,18 @@ export default class GitHubURL {
...
 		const branch = ambiguousReference[0];
+		const currentBranch = getCurrentCommittish();
 		// History pages might use search parameters
+		const branchFromSearch = this.searchParams.get('branch');
 		const filePathFromSearch = this.searchParams.getAll('path[]').join('/');
 		if (filePathFromSearch) {
+			this.searchParams.delete('branch');
 			this.searchParams.delete('path[]');
-			return {branch, filePath: filePathFromSearch};
+			return {branch: branchFromSearch ?? currentBranch ?? branch, filePath: filePathFromSearch};
```
fregante, line 47: I don’t think `currentBranch` should be part of this at all. GitHubURL can receive a URL to any page so we can't just change it to use the current branch.

source/github-helpers/github-url.ts
fregante: Anyway, this confirms that there's no way that `filePathFromSearch` is truey while `branchFromSearch` isn't, so it should probably be
```suggestion
			return {branch: branchFromSearch, filePath: filePathFromSearch};
```

source/github-helpers/index.ts
fregante: https://github.com/sindresorhus/element-ready/releases

test/get-current-committish.ts
```diff
@@ -0,0 +1,96 @@
...
+		'https://github.com',
+		'github.com'
+	));
+
+	document.head.insertAdjacentHTML('beforeend', '<a data-hotkey="t" href="https://github.com/avajs/ava/find/master"></a>');
+	// Root
+	t.is(getCurrentCommittish(
+		'/typescript-eslint/typescript-eslint',
+		'typescript-eslint/typescript-eslint: Monorepo for all the tooling which enables ESLint to support TypeScript'
+	), undefined);
```
fregante, line 23: This
```suggestion
	document.head.insertAdjacentHTML('beforeend', '<a data-hotkey="t" href="https://github.com/typescript-eslint/typescript-eslint/find/cool%2Fdefault%2Fbranch"></a>');
	t.is(getCurrentCommittish(
		'/typescript-eslint/typescript-eslint',
		'typescript-eslint/typescript-eslint: Monorepo for all the tooling which enables ESLint to support TypeScript'
	), 'cool/default/branch');
	document.querySelector('[data-hotkey="t"]')!.remove()
```

test/get-current-committish.ts
```diff
@@ -0,0 +1,96 @@
...
+	t.is(getCurrentCommittish(
+		'/',
+		'some page title'
+	), undefined, 'It should never throw with valid input');
+	t.throws(() => getCurrentCommittish(
+		'https://github.com',
+		'github.com'
+	));
+
+	document.head.insertAdjacentHTML('beforeend', '<a data-hotkey="t" href="https://github.com/avajs/ava/find/master"></a>');
```
fregante, line 18:
```suggestion
	document.head.insertAdjacentHTML('beforeend', '<a data-hotkey="t" href="https://github.com/typescript-eslint/typescript-eslint/find/cool%2Fdefault%2Fbranch"></a>');
```

test/get-current-committish.ts
fregante, line 23:
```suggestion
	), 'cool/default/branch');
```

source/features/index.tsx
yakov116: This is the password page. It should not be in this PR. I put it in the other PR.

test/github-url.ts
```diff
@@ -3,6 +3,8 @@ import test from 'ava';
 import './fixtures/globals';
 import GitHubURL from '../source/github-helpers/github-url';
 
+document.head.insertAdjacentHTML('beforeend', '<a hidden="" data-hotkey="t" data-pjax="true" href="https://github.com/avajs/ava/find/master"></a>');
```
fregante, line 6:
```suggestion
document.head.insertAdjacentHTML('beforeend', '<a hidden="" data-hotkey="t" data-pjax="true" href="https://github.com/sindresorhus/refined-github/find/Slash%2Fslash"></a>');
```

fregante: Actually these can't be here. They need to be inside the `test()` function that uses it

source/github-helpers/github-url.ts
yakov116: GHE. The branch part is a new thing (at least it was not there when we made GitHubURL

test/get-current-committish.ts
```diff
@@ -0,0 +1,95 @@
+import test from 'ava';
```
fregante, line 1: It's best to move these tests back to tests/helpers.ts. It's hard to see what's being changed at all

source/features/index.tsx
fregante: "the password page" is 'Confirm password'.

source/github-helpers/github-url.ts
fregante: Did the `pathname` look exactly like now too? Because we can easily extract the slashed branch from there instead.

source/features/index.tsx
yakov116: I found some pages they changed to "Confirm access"

I found it when i tried adding a collaborator to a repo of mine

test/get-current-committish.ts
yakov116: 👍

source/github-helpers/github-url.ts
yakov116: Yes it did.

source/github-helpers/index.ts
```diff
@@ -34,9 +34,35 @@ export const getCurrentCommittish = (pathname = location.pathname, title = docum
 		return parsedTitle.groups!.branch;
 	}
 
+	// TODO [2021-04-12]: .last needed for #2799 // GHE
```
yakov116, line 37:
```suggestion
	// TODO [2021-08-12]: .last needed for #2799 // GHE
```

fregante: Why?

yakov116: https://github.com/sindresorhus/refined-github/runs/2376926124

fregante: Oh god. Hah I meant to write the same day, but in 2022

test/helpers.ts
```diff
@@ -289,7 +291,7 @@ test('getCurrentCommittish', t => {
 	t.is(getCurrentCommittish(
 		'/typescript-eslint/typescript-eslint/commits/chore/lerna-4/docs/getting-started/README.md',
 		'History for docs/getting-started/README.md - typescript-eslint/typescript-eslint'
-	), 'chore'); // Wrong, but
+	), 'master'); // Uses the feedlink
```
fregante, line 294: This only shows that the test is wrong, and thus useless.

For this to be tested correctly, it should have the correct feedLink, added and removed just for this test like we do for  `data-hotkey="t"`
```suggestion
	), 'chore/lerna-4');
	document.querySelector('the feed link')!.remove();
```

**+ the code to add it before the test**

source/github-helpers/index.ts
fregante: Can you also add `unicorn/expiring-todo-comments: warn` in the xo settings? I don't want it to automatically fail lint.

fregante:
I'm closing this PR because extremely long. Please send a new one

yakov116:
> I'm closing this PR because extremely long. Please send a new one if you have time, resolving all the review comments and squashing the commits. Or leave it for someone else because we've gone over the same code several times.

I will leave it for someone else.

PR #4199, opened 2021-04-05
Preserve `select-all-notifications-shortcut` after clicking Done

### LINKED ISSUE #4198: The new `a` shortcut in the Notification page doesn't work with the refreshed notifications after operating `Done`
The recent `select-all-notifications-shortcut` on https://github.com/notifications is fantastic.
But I encountered one instance where the `a` shortcut doesn't work.

Here are how to replicate the phenomenon:

1. (Prerequisite) You have many notifications to be shown in the page.
2. Press the `a` shortcut to check all the notifications and operate `Done` (either via `Done` button or the `e` shortcut).
3. After 2, the remaining notifications will be shown – for these, the `a` shortcut doesn't work.

GIF image:

Kapture 2021-04-05 at 12 19 32

Related issue: #3876 (PR #4110)

cheap-glitch:
Fixes #4198 

## Test URLs
* https://github.com/notifications?query=is%3Adone

## Screenshot
n/a

Steps to reproduce:
1. Make sure you have more than one page of notifications to select
2. Press <kbd>a</kbd> and click "Done"
3. Wait a second or two for the other notifications to load
4. Press <kbd>a</kdb>
5. Nothing happens

### DIFF
--- a/source/features/select-all-notifications-shortcut.tsx
+++ b/source/features/select-all-notifications-shortcut.tsx
@@ -8,7 +8,7 @@ async function init(): Promise<void> {
 	const selectAllNotifications = await elementReady<HTMLElement>('.js-notifications-mark-all-prompt');
 	if (selectAllNotifications) { // Notifications page may be empty
 		selectAllNotifications.dataset.hotkey = 'a';
-		await onElementRemoval(selectAllNotifications);
+		await onElementRemoval(selectAllNotifications); // "Select all" checkbox will be replaced if there's more notifications to load #4199
 		void init();
 	}
 }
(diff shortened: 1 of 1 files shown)


fregante:
Can you add steps on how to replicate this issue? And also add `// #4199` comment

fregante:
It looks like it specifically breaks after clicking Done, that's it. Regardless of select all.

1. Select one notification
2. Click Done
3. Wait a second or two
4. Press <kbd>a</kbd>
5. Nothing happens

cheap-glitch:
It breaks only if clicking "Done" _and there's more notifications to load_. Otherwise the checkbox won't be replaced and the shortcut will still work. This is what tripped me up when testing the previous PR.

fregante:
Gotcha. I never click Done on my notifications so there's always more 😃

PR #4201, opened 2021-04-05
Meta: Clean up webpack config file

fregante:
This extracts code and checks unrelated to the build to its own file.

Ideally, separately, the actual checks would be completely separate from webpack. They'd require this same file, but they wouldn't throw errors.


yakov116:
@fregante can you update webpack too? I tried updating and I was unable to build

package.json
```diff
@@ -120,7 +120,7 @@
 		"type-fest": "^0.21.3",
 		"typed-query-selector": "^2.4.1",
 		"typescript": "^4.2.3",
-		"webpack": "^5.27.1",
+		"webpack": "^5.30.0",
 		"webpack-cli": "^4.5.0",
```
yakov116, line 124: this too

fregante:
It was

build/readme-parser.ts
```diff
@@ -0,0 +1,65 @@
+/// <reference types="../source/globals" />
+import path from 'path';
+import {readFileSync} from 'fs';
+import regexJoin from 'regex-join';
+import {parse as parseMarkdown} from 'markdown-wasm/dist/markdown.node.js';
```
yakov116, line 5:
```suggestion
import path from 'path';
import regexJoin from 'regex-join';
import {readFileSync} from 'fs';
import {parse as parseMarkdown} from 'markdown-wasm/dist/markdown.node.js';
```

build/readme-parser.ts
```diff
@@ -0,0 +1,65 @@
...
+	const markdownDescription = lineMatch[1].replace(/\[(.+?)]\((.+?)\)/g, urlExtracter);
+	return {
+		id,
+		description: parseMarkdown(markdownDescription),
+		screenshot: urls.find(url => /\.(png|gif)$/i.test(url))
+	};
+}
+
+function searchInHighlights(readmeContent: string, id: FeatureID): FeatureMeta | void {
+	// Feature might be highlighted in the readme
```
fregante, line 31:
```suggestion
```

webpack.config.ts
```diff
@@ -1,57 +1,18 @@
 /// <reference types="./source/globals" />
 
 import path from 'path';
-import {readFileSync} from 'fs';
```
yakov116, line 4:
```suggestion
```

fregante: I think actually built-in modules need to be spaced

build/readme-parser.ts
```diff
@@ -0,0 +1,65 @@
+/// <reference types="../source/globals" />
```
yakov116, line 1: Unrelated question. Why is this needed?

yakov116:
Does build need to be added here?

https://github.com/sindresorhus/refined-github/blob/a34191037be2ce5ad44ca6cc7015f71c8c8a7b19/tsconfig.json#L11-L14

If yes `"octicon-svg-loader.ts"` can be removed

build/readme-parser.ts
fregante: Don't ask me!

I think I tried moving `globals` to the root and use it there but it didn't work…

yakov116 (approved):
Untested but looks good.

Want to regenerate the lock?

webpack.config.ts
```diff
@@ -160,9 +106,7 @@ const config: Configuration = {
...
 };
 
-const webpackSetup = (_: string, options: webpack.WebpackOptionsNormalized): Configuration => {
+export default function webpackSetup(_: string, options: webpack.WebpackOptionsNormalized): Configuration {
 	isWatching = Boolean(options.watch);
 	return config;
-};
-
-export default webpackSetup;
+}
```
yakov116, line 112: hey why no `;`?

fregante:
It’s ok for now since only select dev dependencies have been updated.

yakov116:
> It’s ok for now since only select dev dependencies have been updated, we don’t have to do it every time.

I got burned from this so many times. I wish I would know the reason, but since we switched to npm 15 every time I do an update if I don't delete and start over, 💥.

Probably has to do with my machine. 🤷

fregante:
Yeah it does happen but I supposed that if it works on CI it works for everyone. Usually they're peer dependency errors where babel stuff got updated partially and it's not fully compatible across versions.

fregante:
Tested both build and watch

yakov116:
Me too :)

Just a heads up I really learned from this PR. How you put the code together so simply and understandable.

I look at the code and I think to myself "ya maybe one day..."

fregante:
> When do you sleep?!

I recently moved across 12 timezones. I might have normalized my sleep schedule for now 😂

yakov116:
I was wondering!

PR #4206, opened 2021-04-06
Make fork count visible on click

fregante:
Not yet tested


fregante:
The code says

https://github.com/sindresorhus/refined-github/blob/7882e292ee0eec130b542e017d918ab7f06555be/source/features/hide-watch-and-fork-count.css#L2

and indeed it's still right

Issue #4008, opened 2021-02-21
💯  Minor codebase updates and fixes


fregante:
I'll use this issue to add small tasks that could be done, treat it as a wishlist or a good way to start getting acquainted with the codebase.

- [ ] Find fickle selectors and change them with `select(something more specific).closest(the parent)`;

	GitHub has been removing some feature-specific classes like `.js-file-list` which means that we end up with monstrosities like `main > :first-child .mt-1 .btn` which mean nothing and could change at any time. Ideally instead we should find the best selector inside the target element (for example its icon, `select('.octicon-branch[aria-label="Create branch"]')`) and then reach the desired element with `.closest('.btn')`:

  ```diff
  - select('main > :first-child .mt-1 .btn')
  + select('.octicon-branch[aria-label="Create branch"]')!.closest('.btn')
  ```

- Better yet: https://github.com/sindresorhus/refined-github/issues/4456

doc-han:
Hi, I'm taking on this

fregante:
Thank you! They’re all different points, so please do one PR for each so it’s easier to review.

cheap-glitch:
> Use mem.decorator wherever possible sindresorhus/mem#61

Sadly, TypeScript decorators can only be used inside classes: https://www.typescriptlang.org/docs/handbook/decorators.html#decorators

fregante:
- [ ] Change every GHE-related comments to `/* GHE #issue */` so that they're easily trackable and removed once old enough. Never "GitHub Enterprise" but always GHE (example: https://github.com/sindresorhus/refined-github/blob/3aadf2f9141107d7ca92e8753f2a66cbc10ebd9d/source/features/safer-destructive-actions.css#L4)
- [ ] Possibly review and remove such comments if they're older than 1 year

yakov116:
The last 2 I have included in my june refresh PR (un committed). If you want I can commit them now. But I would rather wait for then so we know that we are breaking GHE in one PR 😄

fregante:
- [ ] use `.base-ref` instead of `:not(.head-ref)` 🤷‍♂️ https://github.com/sindresorhus/refined-github/blob/1a0d254f87e235dd7991c1b8f4034326dc822800/source/features/clean-conversation-headers.tsx#L32:L32

yakov116:
Change 
[...]

PR #4207, opened 2021-04-07
Lint

yakov116:
Does most of #4008

- Use table icon in `table-input`
- Use g-plane/typed-query-selector in `fetch-dom` (I have no idea what I did here, I copied what you wrote in select-dom)
- Fix `link-to-changelog-file` book color
- Use `pageDetect` to exclude in `enable-file-links-in-compare-view`
- Restore valid checkmark in options
- `GitHub Enterprise` => `GHE`

Regarding the valid token:

image

Seems like this was removed by mistake when you redid the code


source/features/enable-file-links-in-compare-view.tsx
```diff
@@ -61,7 +61,7 @@ void features.add(__filebasename, {
 	],
 	exclude: [
 		// Only enabled on Open/Draft PRs. Editing files doesn't make sense after a PR is closed/merged.
```
fregante, line 63:
```suggestion
		// Editing files doesn't make sense after a PR is closed/merged
```

source/features/enable-file-links-in-compare-view.tsx
```diff
@@ -61,7 +61,7 @@ void features.add(__filebasename, {
 	],
 	exclude: [
```
fregante, line 62: I just realized that this feature also doesn't make sense when the repo has been deleted but the PR is open. It appears as "unknown repository" and it's worth checking other features relying on "get head branch"

source/features/first-published-tag-for-merged-pr.tsx
```diff
@@ -10,7 +10,7 @@ import {buildRepoURL, getRepo} from '../github-helpers';
 import onConversationHeaderUpdate from '../github-events/on-conversation-header-update';
 
 const getFirstTag = cache.function(async (commit: string): Promise<string | undefined> => {
-	const firstTag = await fetchDom<HTMLAnchorElement>(
+	const firstTag = await fetchDom(
```
fregante, line 13: Looks like the types are working 👍

source/features/link-to-changelog-file.tsx
```diff
@@ -68,7 +68,7 @@ async function init(): Promise<void | false> {
 			style={{padding: '6px 16px'}}
 			role="button"
 		>
-			<BookIcon className="text-blue mr-2"/>
+			<BookIcon className="text-blue color-text-link mr-2"/>
```
fregante, line 71: Good catch. Are there other `text-` classes used anywhere?

source/features/safer-destructive-actions.css
```diff
@@ -1,7 +1,7 @@
 /* Move "close issue" and "cancel" buttons on authoring comments to the left */
 
 /* ...in issue comment form */
-/* the legacy way (kept for GitHub Enterprise) */
+/* the legacy way (kept for GHE) */
```
fregante, line 4: GHE comments should include a date or issue or we should mark ALL of them with the expiring TODOs feature:

```js
// TODO[2021-10-01]: GHE #1234
```

My guess is that this specified piece of code is at least 2 years old and can be dropped though

source/options.tsx
```diff
@@ -22,7 +22,7 @@ function reportStatus({error, text, scopes}: Status): void {
 	if (error) {
 		tokenStatus.dataset.validation = 'invalid';
 	} else {
-		delete tokenStatus.dataset.validation;
+		tokenStatus.dataset.validation = 'valid';
```
fregante, line 25: This is not equivalent. `valid` means valid, the lack of it means _undefined, we don't know yet._ It shows a different icon

fregante: > * Restore valid checkmark in options

This isn't what I suggested

source/features/link-to-changelog-file.tsx
yakov116: Nope!

source/features/safer-destructive-actions.css
yakov116: I am going to drop it when we drop refresh.

Will be much easier to drop it all at once IMO

source/options.tsx
yakov116: Will add.

Wording suggestion?

source/features/safer-destructive-actions.css
fregante: This really is 2 years old though, this rule can be deleted now

Screen Shot 2021-04-07 at 12 30 49

yakov116: What about 

https://github.com/sindresorhus/refined-github/blob/edf6b1a50f1cdb55c3303869537f105b2144afa2/source/features/more-dropdown.css#L1-L7

Also 2 years old

yakov116:
New and updated wording

image


💤 good night will continue tomorrow

source/options.tsx
fregante: No wording changes required. The `p` becomes `li`

distribution/options.html
```diff
@@ -22,6 +22,7 @@
 		The token enables <a href="https://github.com/sindresorhus/refined-github/search?q=github-helpers+api" target="_blank">some features</a> to <strong>read</strong> data from public repositories
 	</p>
 	<ul>
+		<li data-validation data-scope="valid_token">The token is valid and it is 2 AM and I cannot think of better wording 💤 💤 💤
```
fregante, line 25:
```suggestion
		<li data-validation data-scope="valid_token">The token enables <a href="https://github.com/sindresorhus/refined-github/search?q=github-helpers+api" target="_blank">some features</a> to <strong>read</strong> data from public repositories

```

yakov116: One second I dont understand having a token with nothing enabled allows some features to run?

fregante: Yes

distribution/options.html
```diff
@@ -22,6 +22,7 @@
 		The token enables <a href="https://github.com/sindresorhus/refined-github/search?q=github-helpers+api" target="_blank">some features</a> to <strong>read</strong> data from public repositories
```
fregante, line 22: This can be dropped

source/features/cross-deleted-pr-branches.tsx
```diff
@@ -18,9 +18,9 @@ function init(): void | false {
 
 	const deletedBranchName = lastBranchAction.textContent!.trim();
 
-	const headReferenceLink = select('.head-ref a')!;
-	const repoRootUrl = headReferenceLink.href.split('/', 5).join('/');
-	const repoIsDeleted = headReferenceLink.textContent === 'unknown repository';
+	const headReference = select('.head-ref')!;
+	const repoIsDeleted = headReference.textContent === 'unknown repository';
+	const repoRootUrl = select('a', headReference)?.href.split('/', 5).join('/');
```
fregante, line 23:
```suggestion
	const repoRootUrl = select('a', headReference)?.pathname.split('/', 3).join('/');
```

Also I think if `repoRootUrl === undefined` then `repoIsDeleted === true`. In short, we don't need `repoIsDeleted` nor its exact check. Right?

yakov116: I am not sure I think github changed how they handle delete repos.

I will make a GHE comment and remove it when we drop refresh.

fregante: No need to keep dead code around when we’re in doubt. We’re not fully sure this works on GHE either

yakov116: I am going to send this in a separate PR. This is a real fix.

source/features/safer-destructive-actions.css
```diff
@@ -1,12 +1,5 @@
 /* Move "close issue" and "cancel" buttons on authoring comments to the left */
```
yakov116, line 2:
```suggestion

/* ...in issue comment form */
```

yakov116:
@fregante I want to merge this. Are you ok with that?

PR #4210, opened 2021-04-08
Meta: Drop unnecessary web-ext redirect

fregante:
Using `npm run start` instead of `web-ext run --target=chromium` doesn't save a lot of time. Given I have a lot of extensions I just added shorter aliases to my shell and they work everywhere without having to check whether `start` is set. 

For new contributors it also drops the awkward command that appears to be local but requires a global dependency.

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -103,16 +103,14 @@ Then load or reload it into the browser to see the changes.
 
 ## Loading into the browser
 
-Once built, load it in the browser of your choice.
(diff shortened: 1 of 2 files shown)


busches:
I like the updated instructions, but why drop the commands instead of updating them?

fregante:
Because they remain unnecessary redirects that hide the underlying tool. Generally I don’t use “npm run lint” either because I run XO and AVA directly as well

PR #909, opened 2017-12-26
Hide preview of follow events in dashboard

AshikNesin:
Fixes #894 

**Before**
image

**After**

image

PR #872, opened 2017-12-13
Remove Specific Toolbar items

yakov116:
Closes #779
@bfred-it Like this?

PR #576, opened 2017-07-05
(WIP) Bunch of code shortcuts

bfred-it:
Started as a de-jquerification PR like #448, ended up shortening a lot of functions.

The code changes are best read one commit at a time, but can be tried all at once.

Sorry for not pushing all of these separately but I felt that checking each one out separately would be just too much to ask since there's no new functionality added.

PR #271, opened 2016-07-20
Move entire sidebar content down with padding

bfred-it:
This is more flexible than `position: relative` on a single element

PR #4214, opened 2021-04-09
Fix `cross-deleted-pr-branches` on deleted repositories

yakov116:
## Test URLs

- deleted branch: https://github.com/sindresorhus/refined-github/pull/576
- deleted branch (from fork): https://github.com/sindresorhus/refined-github/pull/872
- restored branch (on fork): https://github.com/sindresorhus/refined-github/pull/909
- deleted repository: https://github.com/sindresorhus/refined-github/pull/271
- deleted repository on open PR: https://github.com/yakov116/TestR/pull/7
- deleted repository with no deleted branch https://github.com/jquery/jquery/pull/769

## Screenshot

image

FYI: GitHub used to display 'unknown repository' they no longer do that


source/features/cross-deleted-pr-branches.tsx
```diff
@@ -16,20 +16,19 @@ function init(): void | false {
 		return false;
 	}
 
-	const deletedBranchName = lastBranchAction.textContent!.trim();
-
 	const headReferenceLink = select('.head-ref a')!;
-	const repoRootUrl = headReferenceLink.href.split('/', 5).join('/');
-	const repoIsDeleted = headReferenceLink.textContent === 'unknown repository';
+	if (!headReferenceLink) {
+		return; // Don't return false, if the repo is deleted the CSS will cross it out
```
yakov116, line 21: @fregante can you help me with the wording?

fregante:
More test URLs from  https://github.com/sindresorhus/refined-github/issues/1985

- deleted branch: https://github.com/sindresorhus/refined-github/pull/576
- deleted branch (from fork): https://github.com/sindresorhus/refined-github/pull/872
- restored branch (on fork): https://github.com/sindresorhus/refined-github/pull/909

source/features/cross-deleted-pr-branches.tsx
```diff
@@ -16,20 +16,19 @@ function init(): void | false {
 		return false;
```
fregante, line 16: I think that if you use `lastBranchAction?` we can drop the previous condition

source/features/cross-deleted-pr-branches.tsx
fregante:
```suggestion
		return; // Don't return false; This feature’s CSS already takes care of this
```

fregante: Screen Shot

This doesn't seem to be the case. CSS covers the header, but not the body.

Test: https://github.com/sindresorhus/refined-github/pull/271

source/features/cross-deleted-pr-branches.tsx
yakov116: We need the text of the branch later on

source/features/cross-deleted-pr-branches.tsx
yakov116: Should be fixed now.

source/features/cross-deleted-pr-branches.tsx
```diff
@@ -8,29 +8,29 @@ import features from '.';
 
 function init(): void | false {
 	const lastBranchAction = select.last('.TimelineItem-body .user-select-contain > span:not(.base-ref)');
```
fregante, line 10: Does this work?
```suggestion
	const lastBranchAction = select.last('.TimelineItem-body .user-select-contain > .head-ref');
```

source/features/cross-deleted-pr-branches.tsx
fregante: That's not what I suggested, no variable would be changed. It's just that `?.` would replace the `if`

source/features/cross-deleted-pr-branches.tsx
yakov116: No it will fail here https://github.com/sindresorhus/refined-github/pull/271

Since the deleted branch does not have `.head-ref` after its deleted

source/features/cross-deleted-pr-branches.tsx
yakov116: ~Not sure what you mean here.~ Ohhh I see how misunderstanding. `lastBranchAction.textcontent` = the branch name that we need to delete. That is user later on in the code.

Also the code changed a bit now is it still relevant

source/features/cross-deleted-pr-branches.tsx
fregante: Indeed, how about this?
```suggestion
	const lastBranchAction = select.last('.TimelineItem-body .commit-ref');
```

yakov116: No that was not good. We only want if they deleted, that gives any action with a commit ref.

source/features/cross-deleted-pr-branches.tsx
```diff
@@ -7,30 +7,30 @@ import {wrap} from '../helpers/dom-utils';
 import features from '.';
 
 function init(): void | false {
-	const lastBranchAction = select.last('.TimelineItem-body .user-select-contain > span:not(.base-ref)');
-	if (!lastBranchAction) {
-		return false;
+	const lastBranchAction = select.last('.TimelineItem-body .commit-ref');
```
yakov116, line 10:
```suggestion
	const lastBranchAction = select.last('.TimelineItem-body .user-select-contain > span:not(.base-ref)');
```

source/features/cross-deleted-pr-branches.tsx
fregante: If a branch has been deleted, there are no other commit refs afterwards, I think

yakov116: No but that selector will catch the merged into action

fregante: But I see that .`user-select-contain` here is actually used to find the branch modified event. So
```suggestion
	const lastBranchAction = select.last('.TimelineItem-body .user-select-contain.commit-ref');
```

yakov116: Looks like the old code was to exclude exactly this https://github.com/jquery/jquery/pull/769.

yakov116: > In [jquery/jquery#769](https://github.com/jquery/jquery/pull/769) it selects the base branch but then it's filtered out by the next check (`!includes(' deleted ')`)

Great!

yakov116:
This feature really looks good after https://github.com/sindresorhus/refined-github/pull/4152

@fregante 🚀 ?

source/features/cross-deleted-pr-branches.tsx
yakov116: > In [jquery/jquery#769](https://github.com/jquery/jquery/pull/769) it selects the base branch but then it's filtered out by the next check (`!includes(' deleted ')`)

Just realized that will cause a return false.

If we add :not(.base-ref) it will exclude it.

fregante: > Just realized that will cause a return false.

And? If the last event is not a deletion, then the branch hasn't been deleted. That PR is locked, so even if it's deleted, we can't detect that.

yakov116: Everything is working fine, nothing is broken. The only thing is the feature logger.

As we did the following.
```js
	if (!headReferenceLink && !lastBranchAction) {
		return; // Don't return false, This feature’s CSS already takes care of this
	}
```

yakov116: @fregante if this is a non-issue, lets merge this.

fregante (approved):

Issue #4216, opened 2021-04-10
Safari extension does not appear


L1cardo:
I installed App Store version of the Refined GitHub, and there is nothing in my Safari extension.
If I click “Quit and Open Safari Extensions Preferences”, nothing happened either.
I tried reinstalling and rebooting my Mac, did not work.

Refined GitHub: 14.0.3 (16610.4.3.1.7)
Mac: Mac mini (M1, 2020)
OS: 11.2.3 (20D91)
Safari: 14.0.3 (16610.4.3.1.7)

<img width="1306" alt="image" src="https://user-images.githubusercontent.com/33802186/114262309-dc88b380-9a11-11eb-9666-0dc8f2c8fa45.png">

fregante:
Yep, same problem on my computer starting this week.

fregante:
It works for me in a new macOS user profile, which leads me to believe it's a configuration issue (together with a bug of our app)

sindresorhus:
Strange. It works perfectly fine on my computer. macOS 11.2.3.

sindresorhus:
I would recommend checking logs in the Console app. It usually logs there if there are problems.

fregante:
System.log only:

```
Apr 10 15:01:21 Rico com.apple.xpc.launchd[1]: Coalition Cache Hit: app<application.com.sindresorhus.Refined-GitHub.4638905.4638911(501)> [914]
```

Happens on open

fregante:
```
default	15:03:32.952138+0700	Spotlight	No suitable cardSection builder found for result <SPApplicationQueryResult 0x7fd53e860630> Refined GitHub (4638905-1795162144): score: 0x000000001e0053040000007e8400000b location:/Applications/Refined GitHub.app doNotFold:0 blockId:0 feedbackBlockId:0
default	15:03:32.984915+0700	Spotlight	No suitable cardSection builder found for result <SPApplicationQueryResult 0x7fd53e9addc0> Refined GitHub (4638905-1795162144): score: 0x000000001e0053040000007e8400000b location:/Applications/Refined GitHub.app doNotFold:0 blockId:0 feedbackBlockId:0
[...]

PR #4218, opened 2021-04-10
Improve error handling for the Safari extension app

sindresorhus:
I also slightly improved the look of the wrapper app:

Screen Shot 2021-04-10 at 15 27 41


fregante:
Does this fix or explain #4216? Now that the problem is gone from my computer I can't test it anymore

sindresorhus:
@fregante It doesn't fix it. But if it happens, it shows the user an error dialog at least.

fregante:
Was this included in 21.4.12? 😳 

Anyway, I'm not a Swift dev, so merge as you see fit

sindresorhus:
Ugh, no it was not. I'll do another release in a few days then I guess.

PR #4222, opened 2021-04-10
`fit-textareas`: Improve performance

fregante:
Follows https://github.com/sindresorhus/element-ready/pull/35 and https://github.com/fregante/fit-textarea/pull/15

`fit-textarea` would need some testing to ensure it works correctly, even if it had some automated browser tests.

Hopefully this update fixes the terrible performance this feature has in Safari. I had to disable it in PRs because it would cause each character to appear on screen after 700ms… each.

On my new computer this is less noticeable, but still I can feel it


yakov116:
@fregante can we merge this and see if we have any issues?

fregante:
Tested in Chrome and it seems to behave correctly there at least

PR #4235, opened 2021-04-13
Simplify `show-user-top-repositories`

### LINKED ISSUE #4221: Use sort=stargazers for `show-user-top-repositories`
Now we can [sort repos in user and organization profiles by star count](https://github.blog/changelog/2021-04-09-sort-repos-in-user-and-organization-profiles-by-star-count/), so we can use <https://github.com/sindresorhus?tab=repositories&sort=stargazers> instead of <https://github.com/search?o=desc&q=user%3Asindresorhus&s=stars&type=Repositories>.

Though GitHub provides this feature now, `show-user-top-repositories` is still useful because it needs only one click instead of three.

cheap-glitch:
Closes #4221 

## Test URLs
https://github.com/sindresorhus

## Screenshot
n/a

### DIFF
--- a/source/features/show-user-top-repositories.tsx
+++ b/source/features/show-user-top-repositories.tsx
@@ -3,10 +3,9 @@ import select from 'select-dom';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
-import {getCleanPathname} from '../github-helpers';
 
 function buildUrl(): URL {
-	const url = new URL('/' + getCleanPathname(), location.href);
+	const url = new URL(location.pathname, location.href);
 	url.searchParams.set('tab', 'repositories');
 	url.searchParams.set('sort', 'stargazers');
 	return url;
--- a/source/features/show-user-top-repositories.tsx
+++ b/source/features/show-user-top-repositories.tsx
@@ -5,12 +5,10 @@ import * as pageDetect from 'github-url-detection';
 import features from '.';
 import {getCleanPathname} from '../github-helpers';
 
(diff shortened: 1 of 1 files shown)


yakov116 (approved):
Awesome!

source/features/show-user-top-repositories.tsx
```diff
@@ -5,20 +5,18 @@ import * as pageDetect from 'github-url-detection';
...
 import {getCleanPathname} from '../github-helpers';
 
-function buildUrl(queryField: string): URL {
-	const url = new URL('/search', location.href);
-	url.searchParams.set('o', 'desc');
-	url.searchParams.set('q', `user:${getCleanPathname()}`);
-	url.searchParams.set('s', queryField);
-	url.searchParams.set('type', 'Repositories');
+function buildUrl(): URL {
+	const url = new URL('/' + getCleanPathname(), location.href);
```
yakov116, line 9: I think I may like the search page, since searching for a language and other filters are simpler.

WDYT?

cheap-glitch: I have no opinion about this. @fregante?

fregante: Since we’re on the profile page, we should stay on the profile page

fregante, line 9: This should be enough
```suggestion
	const url = new URL(location.pathname, location.href);
```

fregante (approved):

PR #4245, opened 2021-04-15
Fix `clean-conversation-sidebar` for triage access collaborator

### LINKED ISSUE #4238: `clean-conversation-sidebar` is broken
Extension enabled:
Screenshot_20210414_080246
Extension disabled:
Screenshot_20210414_080311

Extension version: 21.4.13
FF version: 78.9.0esr

Note: Sorry if I am missing something here :sweat_smile:

yakov116:
Fixes #4238

## Test URLs
I tested on https://github.com/download-directory/download-directory.github.io/pull/37

## Screenshot
image

### DIFF
--- a/source/features/clean-conversation-sidebar.tsx
+++ b/source/features/clean-conversation-sidebar.tsx
@@ -101,7 +101,6 @@ async function clean(): Promise<void> {
 
 	// Linked issues/PRs
 	select('[aria-label="Link issues"] p')!.remove(); // "Successfully merging a pull request may close this issue."
-	select('[aria-label="Link issues"] p')?.remove(); // "Successfully merging a pull request may close this issue." Not shown if issues are disabled.
 	cleanSection('[aria-label="Link issues"]');
 
 	// Projects
--- a/source/features/clean-conversation-sidebar.tsx
+++ b/source/features/clean-conversation-sidebar.tsx
@@ -65,8 +65,9 @@ async function clean(): Promise<void> {
 	select('#partial-discussion-sidebar')!.classList.add('rgh-clean-sidebar');
 
 	// Assignees
-	const assignees = select('.js-issue-assignees, .sidebar-assignee')!; // `.sidebar-assignee` is for collaborators with only "triage access"
+	const assignees = select('.js-issue-assignees')!;
 	if (assignees.children.length === 0) {
+	if (assignees.children.length === 0 && !select('#assignees-select-menu')) {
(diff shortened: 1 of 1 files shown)


yakov116:
Hmm this breaks other collaborator types

yakov116:
OK from my finding, they have a unique DOM. Its a regular user with some added select dropdowns.

source/features/clean-conversation-sidebar.tsx
```diff
@@ -101,6 +101,7 @@ async function clean(): Promise<void> {
 
 	// Linked issues/PRs
 	select('[aria-label="Link issues"] p')!.remove(); // "Successfully merging a pull request may close this issue."
+	select('[aria-label="Link issues"] p')?.remove(); // "Successfully merging a pull request may close this issue." Not shown if issues are disabled.
```
fregante, line 104: There are 2? The use a loop

yakov116: No This issue is really unrelated. I just threw it into the PR.

I can revert if you want

yakov116: Reverted

fregante:
I think the issue here is specifically because the “Assign yourself” button is missing from the Triager’s sidebar and therefore the “children count” is zero. Since the next condition selects this  element, we can use *that* instead of this additional select. This also allows for a slight improvement of the whole if/else

yakov116:
I lost you here

yakov116:
I am going to close this as this feature needs a to be fixed anyways

source/features/clean-conversation-sidebar.tsx
```diff
@@ -66,7 +66,8 @@ async function clean(): Promise<void> {
 
 	// Assignees
 	const assignees = select('.js-issue-assignees')!;
-	if (assignees.children.length === 0) {
+	// `#assignees-select-menu` check is for collaborators with only "triage access" #4238
+	if (assignees.children.length === 0 && !select('#assignees-select-menu')) {
```
fregante, line 70: What I meant in my previous comment is that `#assignees-select-menu` covers both cases, so `assignees.children.length === 0` is a redundant check.

The same exact thing happened in https://github.com/sindresorhus/refined-github/pull/4336#discussion_r628903336

Here's a comparison for history purposes:

## Full access

Screen Shot 1

## Triage access

114912044-8a6ad800-9ded-11eb-8540-fdef5d868a88

Issue #4197, opened 2021-04-03
What features should be removed and why?


fregante:
There are 2 types of features I'm expecting to hear:

- controversial features could be dropped or made less controversial
- rare/broken/unnecessary features that either don't belong to Refined GitHub or aren't worth supporting

Not promising that the changes will happen, but it's a good time to discuss since we have 200+ features and some changes need to edit/test _many features_ before going live.

dorian-marchal:
I always disable these features:
 - __Automatically deletes the branch right after merging a PR, if possible.__
  I don't want to delete my colleagues branches when I merge them.
- __Uses the PR’s title as the default squash commit title and updates the PR’s title to match the commit title, if changed.__
  My team relies on the default behavior to identify Merge commits.
  I see that this refers to "squash commit" now, so maybe I could reenable that.
- __Hides forks and watchers counters.__
  That's valuable information I don't want to hide.

kidonng:
@dorian-marchal The first two may be addressed in #3964

fregante:
Thanks for the comment. 

How is the fork and watchers count useful to you that the Star count doesn’t already tell?

dorian-marchal:
I feel like watching a repo shows more involvement than simply starring it.
However after thinking about it, the main reason I disabled this option was that it removed the direct links to watchers/forks. It seems like the watchers link has since been added back.
The forks link is still missing though, and there is a "Find useful forks" button on this page that I like.

fregante:
Indeed, but do they really tell you what you hope? 

- This repo has 189 watchers but I only ever see about 10 people active in comments;
- As for the forks, if you want to know the actual number of contributors, it appears on the repo home tab. People (mis-) use forks to bookmark and backup repositories, so the fork count is way off too.

This watchers count is visible and clickable when you open the watchers dropdown, by the way. This wasn’t possible for forks #4206 

However I feel you about the **link** to the forks list. We turned that from 1-click to 3 clicks (more dropdown, insights, forks)

cheap-glitch:
> However I feel you about the link to the forks list. We turned that from 1-click to 3 clicks (more dropdown, insights, forks)

We could re-add it as a dropdown item in `more-dropdown`.

fregante:
> > However I feel you about the link to the forks list. We turned that from 1-click to 3 clicks (more dropdown, insights, forks)
> 
[...]

PR #4249, opened 2021-04-16
Hide the fork count only on smaller screens

yakov116:
Re https://github.com/sindresorhus/refined-github/issues/4197#issuecomment-816628465

## Test URLs

## Screenshot
Video_2021-04-16_102042


fregante (approved):
cc @sindresorhus

yakov116:
@sindresorhus ?

sindresorhus (approved):
I don't find the fork count useful at all, but I also don't feel strongly about hiding it, so this is fine by me.

fregante:
Yeah the count is useless, but the link points to the useful fork network. It was suggested to add a link to the network in `more-dropdown` but at that point we'd just be adding further code for little benefit, so I just suggested we restore this count.

PR #4250, opened 2021-04-16
Align `comments-time-machine-links` banner to the site header

### LINKED ISSUE #4227: `comment-time-machine-links` is misaligned
Either we align it with the black header or we align it with the repo. I think the header is easier.

Screen Shot

Live URL:  https://github.com/sindresorhus/refined-github/blob/main/media/promo.png

GitHub’s information bar is aligned to the black bar:

```js
// https://github.com/sindresorhus/refined-github/blob/master/media/promo.png
```

Screen Shot 1

yakov116:
Fixes #4227

## Test URLs
https://github.com/sindresorhus/refined-github/blob/main/media/promo.png?rgh-link-date=2017-04-12T07%3A11%3A49Z

## Screenshot
image

### DIFF
--- a/source/github-widgets/notice-bar.tsx
+++ b/source/github-widgets/notice-bar.tsx
@@ -21,8 +21,11 @@ export default function addNotice(
 ): void {
 	select('#start-of-content')!.after(
 		<div className={`flash flash-full flash-${type}`}>
-			<div className="container-lg px-3 d-flex flex-items-center flex-justify-between">
-				<div>{message}</div> {action}
+			<div className="px-2">
+				{action}
+				<div>
+					{message}
+				</div>
 			</div>
 		</div>
 	);
--- a/source/github-widgets/notice-bar.tsx
+++ b/source/github-widgets/notice-bar.tsx
@@ -20,12 +20,10 @@ export default function addNotice(
 	}: Options = {}
 ): void {
 	select('#start-of-content')!.after(
-		<div className={`flash flash-full flash-${type}`}>
-			<div className="px-2">
-				{action}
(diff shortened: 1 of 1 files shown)


fregante:
Hard to see the alignment in the screenshot if the window is 20px larger than the body width 😃

source/github-widgets/notice-bar.tsx
```diff
@@ -21,8 +21,11 @@ export default function addNotice(
 ): void {
 	select('#start-of-content')!.after(
 		<div className={`flash flash-full flash-${type}`}>
-			<div className="container-lg px-3 d-flex flex-items-center flex-justify-between">
-				<div>{message}</div> {action}
+			<div className="px-2">
```
fregante, line 24: Can this element be dropped? `px-2` can probably go on the parent

yakov116: I copied the DOM GitHub uses when marking a PR ready for review.

Can't test now.

fregante (changes requested):
Screen Shot

This is what I see without the flex classes

yakov116:
Isn't that what you wanted?

> **Either we align it with the black header** or we align it with the repo.

fregante:
Look on the right side. It's not aligned

fregante:
Ah sorry, I recreated the change in my developer tools but I didn't notice you also flipped the position of the button in the DOM. It works correctly and I also dropped the wrapper now. I had resolved the review because my next comment made me think that `d-flex` was going to be needed anyway.

fregante (approved):

PR #4274, opened 2021-04-24
Meta: Improve issue template chooser screen

fregante:
I think screen has too much text and an unexpected order. Unfortunately GitHub requires 2 lines so I added some emojis. Alternatives welcome

## Before

Screen Shot

## After

Screen Shot 3

### DIFF
--- a/.github/ISSUE_TEMPLATE/4_change_request.md
+++ b/.github/ISSUE_TEMPLATE/4_change_request.md
@@ -1,0 +1,8 @@
+---
+name: Complain
+about: Is there something you don't like? Let's talk
+labels: change request
+---
+
+<!-- If you find any related threads, link them here -->
+<!-- Duplicate issues are ok, but they may be closed pretty quickly -->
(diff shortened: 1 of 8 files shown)


sindresorhus:
> Unfortunately GitHub requires 2 lines so I added some emojis. Alternatives welcome

How about zero width space?

fregante:
I tried all of these spaces at once 3 times each: https://codepen.io/rubikzube/pen/wolgB

No luck. I suppose I could also use. `———`

sindresorhus:
> I suppose I could also use. ——— (minimum 3 chars)

👍

.github/ISSUE_TEMPLATE/4_change_request.md
```diff
@@ -1,5 +1,5 @@
 ---
-name: Change requests
+name: Complain
 about: Is there something you don't like? Let's talk
 labels: change request
```
fregante, line 4: All good here? Feels like the verb is correct 😃

sindresorhus: 👍

PR #3093, opened 2020-05-13
Add `first-published-tag-for-merged-pr` feature

yakov116:
1. LINKED ISSUES:
   Closes #3088

2. TEST URLS:

	https://github.com/sindresorhus/refined-github/pull/3062
	https://github.com/sindresorhus/refined-github/pull/3079
	https://github.com/ovity/octotree/pull/1
	https://github.com/sindresorhus/refined-github/pull/837

3. SCREENSHOT:

image

Sticky Bar.
image

Issue #4275, opened 2021-04-25
`fit-textareas` causing major performance issue when typing in PR description textarea with Chrome


with-heart:
With the `fit-textarea` feature enabled in `refined-github@21.4.23`, I'm experiencing a severe performance degradation when typing in a PR description textarea. Each letter appears to be taking ~500ms to be handled, making the feature unusable as it takes multiple seconds for a few words to appear.

Performance profile shows this line taking 1.5s per call:
![image](https://user-images.githubusercontent.com/1954752/115977908-18f7fb80-a54a-11eb-998f-02e0ed55facb.png)

Disabling the `fit-textarea` feature removes the delay.

I've uploaded the Chrome performance profile json file here: https://gist.githubusercontent.com/with-heart/f0d3b0914d2c25e6f635d82f91404e10/raw/af6e016d11ddf07e73a4459a6f40b1cc4ad58885/refined-github-performance-profile.json

This issue appears to be related to #4266.

fregante:
Is this something that changed in the last release? Can you post a link to an affected PR? I see this
<img width="615" alt="Screen Shot" src="https://user-images.githubusercontent.com/1402241/115984801-1f7e8580-a5d3-11eb-9a02-ee668f20a035.png">

SimenB:
I'm seeing the same thing. E.g. in https://github.com/facebook/jest/pull/11033, but it happens in every PR for me (although the lag/delay gets worse the larger the diff is)

fregante:
I'll revert this change in the next release, let's see how it will be.

The strange part is that `.scrollHeight` was actually in the previous version as well. This version was supposed to improve performance since it no longer resets the whole textarea every time: https://github.com/fregante/fit-textarea/pull/15

I reopened https://github.com/fregante/fit-textarea/issues/14

kylemh:
I'm also seeing it a PR review comment textarea. It seems very heavy onFocus, onBlur, and on key stroke.

![Screen Shot 2021-04-26 at 8 50 24 AM](https://user-images.githubusercontent.com/9523719/116112496-784e2c80-a66c-11eb-9fd1-49c7795bdb64.png)

yakov116:
We are going to revert the change in the next release.

fregante:
Feature disabled via hotfix for the time being

PR #3678, opened 2020-10-22
Meta: Use the Readme as the only source of feature descriptions

FloEdelmann:
Closes #3672.

~I've left a settings object (that is empty in most features) in `features.add`, because both `disabled` and `shortcuts` need to be defined there.~

The first two commits actually change some of the descriptions/screenshots, so that the later commits don't introduce any changes that are not visible in any diff. I've made those changes manually (with a helper function that generates two files that I could diff).

### Implemented in this PR:

1. Parse features' description and screenshot from readme.
2. Disable features by commenting them out in `refined-github.ts`.
3. Move `shortcuts` from `FeatureMeta` to `FeatureLoader`.
4. Drop the `FeatureMeta` parameter from `features.add`.
5. Render `<i>` and `<kbd>` tags in addon options:
  
6. Update `parseBackticks`, so that ``` `` ` `` ``` and ``` `` `bla` `` ``` are also parsed correctly (and add some tests for it).
7. Improve some feature descriptions.
8. Add and use a `concatRegex` helper (suggested by @fregante in https://github.com/sindresorhus/refined-github/pull/3678#discussion_r510277945).

PR #4276, opened 2021-04-25
Lint

yakov116:
Disabled "@typescript-eslint/no-non-null-assertion": not compatible with DOM
Disabled "@typescript-eslint/no-non-null-asserted-optional-chain"  See https://github.com/sindresorhus/refined-github/pull/3093#discussion_r425338237 and https://github.com/sindresorhus/refined-github/pull/3168#discussion_r434555667

Need typescript help
`@typescript-eslint/no-unsafe-argument` All but 2 are this issue. If you show me how to fix one I can do the rest.


source/features/no-duplicate-list-update-time.tsx
```diff
@@ -9,7 +9,7 @@ function parseTime(element: HTMLElement): number {
 
 function init(): void {
 	for (const issue of select.all('.js-navigation-item[id^="issue_"]')) {
-		const [stateChangeTime, updateTime] = select.all('relative-time', issue);
+		const [stateChangeTime, updateTime] = select.all<'relative-time'>('relative-time', issue);
```
fregante, line 12: ???

source/options.tsx
```diff
@@ -53,8 +54,8 @@ async function getTokenScopes(personalToken: string): Promise<string[]> {
 	});
 
 	if (!response.ok) {
-		const details = await response.json();
-		throw new Error(details.message);
+		const details: JsonObject = await response.json();
+		throw new Error(String(details.message));
```
fregante, line 58: Does this fix a specific error? Seems unnecessary if we're just going to force it into a string

source/features/file-finder-buffer.tsx
```diff
@@ -20,7 +20,7 @@ const getBufferField = onetime((): HTMLInputElement => (
 ));
 
 function pjaxStartHandler(event: CustomEvent): void {
-	const destinationURL = event.detail?.url;
+	const destinationURL: string = event.detail?.url;
```
fregante, line 23: Necessary? This should have `| undefined`

.xo-config.json
```diff
@@ -16,6 +16,8 @@
 		"no-void": "off",
 		"react/jsx-key": "off",
 		"unicorn/no-array-callback-reference": "off",
+		"@typescript-eslint/no-non-null-assertion": "off",
+		"@typescript-eslint/no-non-null-asserted-optional-chain": "off",
```
fregante, line 20: What's the problem again? It worked until now and https://github.com/typescript-eslint/typescript-eslint/issues/1976 was resolved

fregante (commented):
What fixed `.closest`’s types?

fregante:
> Need typescript help

There are a million type issues, can't open a PR and ask me to fix them. Just don't break them

.xo-config.json
yakov116: Same issue

yakov116:
> There are a million type issues, can't open a PR and ask me to fix them. Just leave things unchanged

`@typescript-eslint/no-unsafe-argument` All but 2 are this issue. If you show me how to fix one I can do the rest.

Sorry I wanted to update the first post I did not think you would review so fast 😄

source/options.tsx
yakov116: I did not mean to commit this

source/features/no-duplicate-list-update-time.tsx
yakov116: `@typescript-eslint/no-unsafe-argument`

yakov116:
> What fixed `.closest`’s types

Its not fixed.

source/features/no-duplicate-list-update-time.tsx
fregante: Read the rule description:

> This rule disallows calling a function with `any` in its arguments

Either the rule is faulty or our types are faulty. This piece of code has no meaning and should be undone. If it also breaks `querySelector` it might be worth reporting it on https://github.com/g-plane/typed-query-selector, who could know better what's happening

fregante:
> > What fixed `.closest`’s types
> 
> Its not fixed.

I see that `<HTMLElement>` is being dropped from all. Why? If it works without, that's what I mean by "fixed"

yakov116:
> I see that `<HTMLElement>` is being dropped from all. Why? If it works without, that's what I mean by "fixed"

I am going to work on it. That part is not ready yet.

.xo-config.json
fregante: If Sindre re-enabled them they're probably working now, so maybe our code is wrong.

The only reason I can think of to disable these are if the changes required just clutter the code without actually improving it

fregante:
If this PR makes it for May 1st, can you revert #4222 in here due to https://github.com/sindresorhus/refined-github/issues/4275?

yakov116:
> If this PR makes it for May 1st, can you revert #4222 in here due to #4275?

Sure

.xo-config.json
fregante: `@typescript-eslint/no-non-null-assertion` is correct in being disabled, we need those errors.

I'm checking the other one

fregante: `@typescript-eslint/no-non-null-asserted-optional-chain` should probably not be disabled.

However I'm not liking most of the type changes in this PR so I'd rather avoid all of them until a TS developer can review them properly. 

If you're going to update XO:

- disable `no-unsafe-argument` + a TODO to review later
- disable `no-non-null-asserted-optional-chain` + same TODO
- disable `no-non-null-assertion` with a comment that we need those runtime errors

yakov116: Ok

Can I leave the `interface` ones?

fregante:
I wish could could change some global types, but adding this to global.d.ts doesn't work 😩 

```ts
interface Object {
	keysa<T, K extends keyof T>(o: T): K[];
}

interface JSON {
	parse(text: string, reviver?: (this: any, key: string, value: any) => any): JsonValue;
}

// and also Response#json()
```

source/features/show-associated-branch-prs-on-fork.tsx
```diff
@@ -42,11 +45,11 @@ const getPullRequestsAssociatedWithBranch = cache.function(async (): Promise<Rec
 
 	const pullRequests: Record<string, PullRequest> = {};
 	for (const {name, associatedPullRequests} of repository.refs.nodes) {
-		const [prInfo] = associatedPullRequests.nodes;
+		const [prInfo] = associatedPullRequests.nodes as PullRequest[];
 		// Check if the ref was deleted, since the result includes pr's that are not in fact related to this branch but rather to the branch name.
 		const headRefWasDeleted = prInfo?.timelineItems.nodes[0]?.__typename === 'HeadRefDeletedEvent';
 		if (prInfo && !headRefWasDeleted) {
-			prInfo.state = prInfo.isDraft && prInfo.state === 'OPEN' ? 'Draft' : upperCaseFirst(prInfo.state);
+			prInfo.state = prInfo.isDraft && upperCaseFirst(prInfo.state) === 'Open' ? 'Draft' : upperCaseFirst(prInfo.state) as keyof typeof stateClass;
```
yakov116, line 52: Is there another way to do this without adding the extra function?

fregante: I don't understand why add the function in the first place, it's just changing the case

yakov116: Since 'OPEN' is not `keyof typeof stateClass` only 'Open'

fregante: That string is only used for comparison, it’s not meant to be a specific type. The condition could have been 1===1 and it should have still passed. 

 The rule is wrong here and probably its confusion can be solved with parens in the right place

source/features/stop-redirecting-in-notification-bar.tsx
```diff
@@ -3,9 +3,13 @@ import delegate from 'delegate-it';
...
 import features from '.';
 
+interface SessionStorage extends Storage {
+	notification_shelf: string;
+}
+
 const hasNotificationBar = (): boolean =>
 	location.search.startsWith('?notification_referrer_id=') ||
-	JSON.parse(sessionStorage.notification_shelf ?? '{}').pathname === location.pathname;
+	JSON.parse((sessionStorage as SessionStorage).notification_shelf ?? '{}').pathname === location.pathname;
```
fregante, line 12: Use .getItem instead if this method doesn’t work as well anymore. The only reason why some code is written in a certain way is to make it shorter or more readable. Having to use assertions defeats the purpose of using property access

.xo-config.json
yakov116: We are using a JSON file which makes it not possible to leave a comment.

fregante: Indeed XO does not support JSON with comments. Can you convert the config to .yml?

source/features/release-download-count.tsx
```diff
@@ -31,15 +36,15 @@ async function getAssetsForTag(tags: string[]): Promise<Tag> {
 	`);
 
 	const assets: Tag = {};
-	for (const [tag, release] of Object.entries(repository)) {
-		assets[tag] = (release as AnyObject).releaseAssets.nodes;
+	for (const [tag, release] of Object.entries(repository as ReleaseAssets[])) {
```
fregante, line 39: This isn't right, this is why I don't like typing the API result too early. Repository is an object, not an array.

source/features/show-associated-branch-prs-on-fork.tsx
```diff
@@ -42,11 +45,11 @@ const getPullRequestsAssociatedWithBranch = cache.function(async (): Promise<Rec
 
 	const pullRequests: Record<string, PullRequest> = {};
 	for (const {name, associatedPullRequests} of repository.refs.nodes) {
-		const [prInfo] = associatedPullRequests.nodes;
+		const [prInfo] = associatedPullRequests.nodes as PullRequest[];
 		// Check if the ref was deleted, since the result includes pr's that are not in fact related to this branch but rather to the branch name.
 		const headRefWasDeleted = prInfo?.timelineItems.nodes[0]?.__typename === 'HeadRefDeletedEvent';
 		if (prInfo && !headRefWasDeleted) {
-			prInfo.state = prInfo.isDraft && prInfo.state === 'OPEN' ? 'Draft' : upperCaseFirst(prInfo.state);
+			prInfo.state = prInfo.isDraft && prInfo.state === ('OPEN' as 'Open') ? 'Draft' : upperCaseFirst(prInfo.state) as keyof typeof stateClass;
```
fregante, line 52: I see now why it requires it this way: `prInfo` is now typed 👌

fregante: I think it's easier if we don't use `upperCaseFirst` at all on this line
```suggestion
			prInfo.state = prInfo.isDraft && prInfo.state === 'OPEN' ? 'DRAFT' : prInfo.state;
```

source/features/vertical-front-matter.tsx
```diff
@@ -24,7 +24,7 @@ function init(): void | false {
 		return false;
 	}
 
-	const values = [...rows[0].children];
+	const values = [...rows[0].children as HTMLCollectionOf<HTMLElement>];
```
fregante, line 27: We don't need this array. `{values[index]}` below can be changed to `{rows[0].children[index]}`

source/helpers/fetch-dom.ts
```diff
@@ -1,9 +1,10 @@
+/* eslint-disable no-redeclare */
```
fregante, line 1: Why was this enabled? 🤔 

It's disabled here

https://github.com/xojs/eslint-config-xo-typescript/blob/a9481df0e9b382287ce94fd3ec394bf9d59dae51/index.js#L374

fregante:
Can you update this link? https://github.com/sindresorhus/refined-github/blob/86c0a731a3bb123b7a48fb0e2e21f4192df4fb46/source/features/index.tsx#L70

The file was renamed

fregante:
> If this PR makes it for May 1st, can you revert #4222 in here due to #4275?

Done in ae020ab5. No rush to merge this PR

yakov116:
> Done in [ae020ab](https://github.com/sindresorhus/refined-github/commit/ae020ab5d6c851045b39cd794e384e7aa4c91f1e). No rush to merge this PR

Thanks. I got very busy with work this week.

.xo-config.json
yakov116: Do you have any sample repo that I can follow? I have never seen a yml xo file.

yakov116: @fregante in case you missed this comment

fregante: Find a JSON to YML converter

.xo-config.json
```diff
@@ -56,4 +60,4 @@
 			"version": "16.13"
 		}
 	}
-}
+}
```
fregante, line 63: 👆

.xo-config.json
yakov116: That did not work.

I am going to revert the Type's and just update dep

fregante: You're right, I assumed it worked since ESLint supports it: https://github.com/xojs/xo/issues/536

source/features/vertical-front-matter.tsx
yakov116: I tried this, I did not fully understand what I was doing here.

source/features/release-download-count.tsx
```diff
@@ -31,15 +36,15 @@ async function getAssetsForTag(tags: string[]): Promise<Tag> {
 	`);
 
 	const assets: Tag = {};
-	for (const [tag, release] of Object.entries(repository)) {
-		assets[tag] = (release as AnyObject).releaseAssets.nodes;
+	for (const [tag, release] of Object.entries(repository as ReleaseAssets)) {
```
fregante, line 39: I think this type is still wrong. Repository does not look like

```js
{
 	releaseAssets: {
 		nodes: {...}
 	}
 }
```

`ReleaseAssets` seems to be the type of the `release` constant. Please double check, I already commented on this.

source/globals.d.ts
```diff
@@ -74,3 +74,11 @@ declare module 'react' {
...
 	cloneNode(deep?: boolean): this;
 }
+
+interface HTMLCollectionOf<T extends HTMLElement> {
+	[Symbol.iterator](): Iterator<T>;
+}
+
+interface NodeListOf<T extends ChildNode> {
+	[Symbol.iterator](): Iterator<T>;
+}
```
fregante, line 84: I assume that the first one defaults `HTMLCollectionOf` to `HTMLElement`, which allows us to avoid having to specify `collection as HTMLElement[]` 👍 

But what's the second one for?

fregante (commented):
Other than the outstanding reviews, it looks good so far, except I haven't tested it

source/globals.d.ts
yakov116: It was for something I reverted, This was only needed if XO was updated.

.xo-config.json
sindresorhus: > We are using a JSON file which makes it not possible to leave a comment.

That's not entirely true: https://github.com/sindresorhus/sindre-playground/blob/565ba9c61dbf4b13d587bcae7047120952f924c5/package.json#L26-L31

source/features/list-prs-for-file.tsx
```diff
@@ -61,7 +61,7 @@ function getSingleButton(prNumber: number, _?: number, prs?: number[]): HTMLElem
 const getPrsByFile = cache.function(async (): Promise<Record<string, number[]>> => {
 	const {repository} = await api.v4(`
 		repository() {
-			pullRequests(
+			pullRequests (
```
sindresorhus, line 64: ^

yakov116: I added on purpose it looked wrong without a space and the arguments on a new line.

I can revert if you feel it wrong

source/globals.d.ts
fregante: Let's undo this as well then, otherwise the reason it exists is lost

yakov116:
@fregante can we disable `clean-conversation-sidebar` in this PR?

fregante:
> @fregante can we disable [`clean-conversation-sidebar`](/sindresorhus/refined-github/blob/main/source/features/clean-conversation-sidebar.tsx) in this PR?

Why?

source/features/release-download-count.tsx
```diff
@@ -32,14 +38,14 @@ async function getAssetsForTag(tags: string[]): Promise<Tag> {
 
 	const assets: Tag = {};
```
fregante, line 39:
```suggestion
	const assets = {};
```

fregante (commented):
Should be good to go after this

source/features/release-download-count.tsx
yakov116: That does not work :(

yakov116:
> @fregante can we disable [`clean-conversation-sidebar`](/sindresorhus/refined-github/blob/main/source/features/clean-conversation-sidebar.tsx) in this PR?

Hey looks like the "disabled" feature was removed. https://github.com/sindresorhus/refined-github/pull/3678/files#diff-cb1dafea7a9c68c5c2ebb2d56f749b3506c13f7d26feeaf56c86521b25c60358 🤔

fregante:
Second bullet point in that PR https://github.com/sindresorhus/refined-github/pull/3678

Just comment out the `import clean-conversation-sidebar` line in refined-github.ts

source/features/list-prs-for-file.tsx
fregante: We don't do this anywhere in the code and also doesn't match the function call style in XO. Reverted

PR #4281, opened 2021-04-26
Improve spacing of PR header buttons

### LINKED ISSUE #4277: `view-last-pr-deployment` is not spaced correctly on editable PRs
Screen Shot 4

In my case the PR was https://github.com/npmgraph/npmgraph/pull/75

It's only missing a single space character

kidonng:
Fix #4277

## Test URLs

See #4089

## Screenshot

image

image

### DIFF
--- a/source/features/view-last-pr-deployment.tsx
+++ b/source/features/view-last-pr-deployment.tsx
@@ -14,9 +14,10 @@ function init(): void {
 	}
 
 	const {href} = select.last<HTMLAnchorElement>(deploymentSelector)!;
-	select('.gh-header-actions')!.prepend(
+	const headerActions = select('.gh-header-actions')!;
+	headerActions.prepend(
 		<a
-			className="rgh-last-deployment btn btn-sm d-none d-md-block"
+			className={`rgh-last-deployment btn btn-sm d-none d-md-block ${select.exists('[aria-label="Edit Pull Request title"]', headerActions) ? 'mr-1' : ''}`}
 			href={href}
 			target="_blank"
 			rel="noreferrer"
--- a/source/features/view-last-pr-deployment.tsx
+++ b/source/features/view-last-pr-deployment.tsx
@@ -14,10 +14,9 @@ function init(): void {
 	}
 
 	const {href} = select.last<HTMLAnchorElement>(deploymentSelector)!;
-	const headerActions = select('.gh-header-actions')!;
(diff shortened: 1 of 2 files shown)


fregante:
We just need to add a single space character, unconditionally. Mr-1 is not exactly the same I think

kidonng:
Does it work? I tried and not luck 😅

source/features/view-last-pr-deployment.tsx
```diff
@@ -14,9 +14,10 @@ function init(): void {
 	}
 
 	const {href} = select.last<HTMLAnchorElement>(deploymentSelector)!;
-	select('.gh-header-actions')!.prepend(
+	const headerActions = select('.gh-header-actions')!;
+	headerActions.prepend(
 		<a
-			className="rgh-last-deployment btn btn-sm d-none d-md-block"
+			className={`rgh-last-deployment btn btn-sm d-none d-md-block ${select.exists('[aria-label="Edit Pull Request title"]', headerActions) ? 'mr-1' : ''}`}
```
fregante, line 20: I don't mind having a little extra space to save all of this code here.

This diff should just include an unconditional `mr-1`

kidonng: What do you think about https://github.com/sindresorhus/refined-github/issues/4277#issuecomment-826358707?

Perhaps add a 4px left margin to `Open With` in `refined-github.css` so it looks better and consistent.

---

Also I find that on the issue page the spacing is 8px:

image

vs PR page:

image

kidonng:
I added new screenshots in PR description.

fregante:
If you’re going that route, we only need a CSS rule, no extra classes. I think left:8px and right:4px

kidonng:
That won't work on non-editable PR, the spacing will be 4px from "Open With".

source/refined-github.css
```diff
@@ -149,3 +149,9 @@ pr-branches
 	box-shadow: var(--color-toast-shadow);
 	z-index: 2147483647;
 }
+
+/* Correct spacing between "Edit" and "Open with" button on PR page */
```
fregante, line 153:
```suggestion
/* Space out "Edit" and "Open with" buttons on PR page, also for `view-last-pr-deployment` #4281 */
```

fregante:
# 💯

PR #4282, opened 2021-04-26
Restore `vertical-front-matter`

### LINKED ISSUE #4273: `vertical-front-matter` is broken
https://github.com/sindresorhus/refined-github/blob/main/.github/ISSUE_TEMPLATE/bug_report.md

No errors in the console. Tested in Safari

Screen Shot

kidonng:
Fix #4273

There isn't a selector which tells us it's a front matter or not, so we are on our own.

I come up with some rules that should handle most cases:

- The table should be the first element
- The table should only have one row in the body
- The head row and the body row should have the same number of cells

Suggestions welcome. Honestly I don't want to drop this feature just because we don't have a nice way to support it :(

## Test URLs

- With front matter: https://github.com/gatsbyjs/gatsby/blob/HEAD@%7B2020-07-08T14:12:43Z%7D/docs/blog/2020-07-07-wordpress-source-beta/index.mdx
image
- Invalid table as first element: https://github.com/kidonng/refined-github/blob/8bec852e016d10a574b1a456174139d6bb9e6c99/test.md
image

- Valid table, but not as first element: https://github.com/kidonng/refined-github/blob/798efec7e082531a8ed99aae5d47d06c31f8cbee/test.md

image

### DIFF
--- a/source/features/vertical-front-matter.tsx
+++ b/source/features/vertical-front-matter.tsx
@@ -9,7 +9,17 @@ import features from '.';
 const hasFrontMatter = (): boolean => pageDetect.isSingleFile() && /\.(mdx?|mkdn?|mdwn|mdown|markdown|litcoffee)$/.test(location.pathname);
 
 function init(): void | false {
-	const table = select('[data-table-type="yaml-metadata"]')!;
+	const table = select('.markdown-body > table:first-child');
+	if (!table) {
+		return false;
+	}
+
+	const headRow = select(':scope > thead > tr', table)!;
+	const bodyRows = select.all(':scope > tbody > tr', table);
+	if (bodyRows.length !== 1 || headRow.childElementCount !== bodyRows[0].childElementCount) {
+		return false;
+	}
+
 	const headers = select.all(':scope > thead th', table);
 	if (headers.length <= 4) {
 		return false;
@@ -17,7 +27,7 @@ function init(): void | false {
 
 	const values = select.all(':scope > tbody > tr > td', table);
 	table.replaceWith(
-		<table className="rgh-vertical-front-matter-table" data-table-type="yaml-metadata">
+		<table className="rgh-vertical-front-matter-table">
 			<tbody>
 				{headers.map((cell, index) => (
 					<tr>
--- a/source/features/vertical-front-matter.tsx
+++ b/source/features/vertical-front-matter.tsx
@@ -20,11 +20,11 @@ function init(): void | false {
 	}
 
 	const rows = select.all(':scope > tbody > tr', table);
-	const values = [...rows[0]?.children];
-	if (rows.length !== 1 || headers.length !== values.length) {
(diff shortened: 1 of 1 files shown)


fregante:
There are now 4 `select` here, `headRow` and `headers` are essentially doing the same thing. Any other improvement possible?

kidonng:
I managed to cut down all extra selects 🎉

source/features/vertical-front-matter.tsx
```diff
@@ -9,15 +9,24 @@ import features from '.';
 const hasFrontMatter = (): boolean => pageDetect.isSingleFile() && /\.(mdx?|mkdn?|mdwn|mdown|markdown|litcoffee)$/.test(location.pathname);
 
 function init(): void | false {
-	const table = select('[data-table-type="yaml-metadata"]')!;
+	const table = select('.markdown-body > table:first-child');
```
fregante, line 12: This will create issues, but hopefully first-child 2-rows tables are rare enough 😬

source/features/vertical-front-matter.tsx
```diff
@@ -9,15 +9,24 @@ import features from '.';
...
+	}
+
 	const headers = select.all(':scope > thead th', table);
 	if (headers.length <= 4) {
 		return false;
 	}
 
-	const values = select.all(':scope > tbody > tr > td', table);
+	const rows = select.all(':scope > tbody > tr', table);
+	if (rows.length !== 1 || headers.length !== rows[0].childElementCount) {
```
kidonng, line 23: I think I missed one point: all rows always have the same number of cells, so the second check is necessary.

@yakov116 Could you add this into #4276?

```suggestion
	if (rows.length !== 1) {
```

source/features/vertical-front-matter.tsx
kidonng: That would still be something like key-value pairs, so the direction change won't be confusing.

source/features/vertical-front-matter.tsx
yakov116: Yes

fregante: I was going to say this, but technically tables could have spanning headers:


<table>
<tr>
	<th colspan="3"> title
<tr>
	<td>1
	<td>2
	<td>3
</table>

I thought it was an extra check to ensure we got exact tables

source/features/vertical-front-matter.tsx
fregante: True, but if someone expect a horizontal table and sees a vertical table it can still cause confusion

source/features/vertical-front-matter.tsx
kidonng: 🤦You are right, I forgot there's `colspan`. So the check is indeed necessary (not possible through Markdown though).

@yakov116 sorry for the trouble but we better revert the change 😅

PR #4267, opened 2021-04-22
Restore `ci-link` feature

yakov116:
Fixes #4260
## Test URLs
Here and I tested on private

## Screenshot

image

PR #4294, opened 2021-04-29
Restore `ci-link` on Enterprise

fregante:
Follows https://github.com/sindresorhus/refined-github/pull/4267
Fixes https://github.com/sindresorhus/refined-github/pull/4267#issuecomment-826728022

cc @gustavkj 

Untested because I'm not a GHE user


gustavkj:
@fregante Thanks for the PR! I've tested the PR in Chrome on our GHE instance and can confirm that it works again! 👍

PR #4297, opened 2021-04-29
Restore `faster-pr-diff-options` feature on PRs

### LINKED ISSUE #2597: Restore `faster-pr-diff-options` feature on PRs
Originally disabled on PRs because of https://github.com/sindresorhus/refined-github/issues/2291

This feature was applied to two places:

- PR file header: now disabled because of #2291
- commit header: it still works

## Suggested solution

1. Restore the buttons right next to the gear icon, replacing it
2. On small screens (via `.d-md-block` classes) we can still show the native dropdown

<details><summary>Old suggestion</summary>

I had suggested some alternatives to replacing the original "PR diff options" dropdown:

> - dropping its "apply and reload" button (saves one click)
> - perhaps opening the dropdown on hover (saves one more click)
> 
> There's already a keyboard shortcut to toggle the whitespace.
> _Originally posted by @fregante in https://github.com/sindresorhus/refined-github/issues/2291#issuecomment-517539452_

Or replacing the gear icon with 2 small icons as described in https://github.com/sindresorhus/refined-github/issues/2597#issuecomment-794294349

<details>

cheap-glitch:
Closes #2597 

## Test URLs
* [PR files](https://github.com/sindresorhus/refined-github/pull/4297/files)
* [Filtered PR files](https://github.com/sindresorhus/refined-github/pull/4276/files?file-filters%5B%5D=.json)
* [Single PR commit](https://github.com/sindresorhus/refined-github/pull/4297/commits/2a045b1e566d238885a93ad9b32727777b541ad9)
* [Single PR commit #2](https://github.com/sindresorhus/refined-github/pull/4297/files/532f870c4b490c0b76c6c368d5bde6ffaff3465d..87445f7eb9ef94e7147249957b8e5c6ef88779aa)
* [Commit page](https://github.com/sindresorhus/refined-github/commit/727d4d2ed14899ace206d6937c3d3ab16eb7b6af)
* [Compare page](https://github.com/sindresorhus/refined-github/compare/21.4.13...21.4.23)

## Screenshots

This PR shows the native settings dropdown until 1012px, where it switches to the quick buttons. (This happens on PR files pages only).

<table>
<tr>
	<td>Large screen (≥1012px)
<tr>
	<td>

screenshot-1619699274
<tr>
	<td>Large screen scrolled down
<tr>
	<td>

screenshot-1619699288
<tr>
	<td>Small screen (<1012px)
<tr>
	<td>

screenshot-1619699306
<tr>
	<td>Small screen scrolled down
<tr>
	<td>

screenshot-1619699327
</table>


source/features/faster-pr-diff-options.tsx
```diff
@@ -5,6 +5,8 @@ import {BookIcon, CheckIcon, DiffIcon} from '@primer/octicons-react';
 
 import features from '.';
 
+const prUIBreakpointClassses = pageDetect.isPRFiles() ? 'hide-sm hide-md hide-lg' : '';
```
fregante, line 8: Global constants aren't updated when the page changes via AJAX

fregante: Also I think it should be the opposite: hide always, then `d-block-whatever the visible size is`

fregante:
That's too big, it should have the same style, not `btn`

cheap-glitch:
> That's too big, it should have the same style as the existing gear icon, not `btn`.

Do you mean the buttons should always have that style and be on the left, even at a larger width?

fregante:
It should always be part of that line. On small screens it should just revert to the native dropdown

cheap-glitch:
Changed the style of the buttons:
screenshot-1619881943
image


and with the sticky header:
screenshot-1619881955

Also, noticed this bug where the wrong diff style button is selected when "No whitespace" is enabled (but only on compare and commit pages), e.g. : https://github.com/sindresorhus/refined-github/compare/21.4.13...21.4.23?w=1. @kidonng is this taken care of by #4304?

fregante:
Looks about right! The style could be better but I can't think of how to improve it. Maybe a `•` before the icons would at least visually separate it from the other dropdowns

> Do we keep the old style on compare/commit pages?

I think so, the new style doesn't fit there.

cheap-glitch:
> Maybe a • before the icons would at least visually separate it from the other dropdowns

imo it just adds visual noise:
image
image


How about simply increasing the margin?
image
image

FloEdelmann:
Maybe the currently active code style button could be hidden? So that there is only a "switch to side-by-side style" button when I'm in inline style mode?

cheap-glitch:
> Maybe the currently active code style button could be hidden? So that there is only a "switch to side-by-side style" button when I'm in inline style mode?

I thought about that too, but it's the classic "UX toggle problem": if we show a single state, it's unclear at first glance whether it's _the current one_ or _the one that will be selected upon clicking_.

fregante:
⚠️  Uhm… there is no winning. Try enabling a file filter, scroll down and resize the window. Zero space again.

Screen Shot 1

I think these buttons can only be safely shown on large viewports.

fregante:
I hate this feature. 😅 Maybe we should go back to the original suggestion:

- just keep the dropdown, but drop the "Apply and reload" button, just automatically apply it on click

cheap-glitch:
What if we:
 * show a single diff icon as suggested by FloEdelmann
 * replace the "No whitespace" link by an icon (arguably not gonna be obvious at first glance what it does, but the tooltip helps)
 * remove some of this weird extra spacing around "Clear filters"

?

fregante:
That may work. We're gonna need some actually-explanatory `tooltipped` on those icons though

source/features/faster-pr-diff-options.tsx
```diff
@@ -47,41 +55,41 @@ function createWhitespaceButton(): HTMLElement {
 		<a
 			href={url.href}
 			data-hotkey="d w"
-			className={`btn btn-sm btn-outline tooltipped tooltipped-s ${isHidingWhitespace ? 'bg-gray-light text-gray-light color-text-tertiary' : ''}`}
+			className={'tooltipped tooltipped-s ' + (isPRPage() ? `d-none d-lg-block color-icon-secondary ${isHidingWhitespace ? '' : 'color-icon-info'}` : `btn btn-sm btn-outline tooltipped ${isHidingWhitespace ? 'bg-gray-light text-gray-light color-text-tertiary' : ''}`)}
 			aria-label={`${isHidingWhitespace ? 'Show' : 'Hide'} whitespace in diffs`}
```
fregante, line 59: Let's match GitHub’s wording here

```suggestion
			aria-label={`${isHidingWhitespace ? 'Show' : 'Hide'} whitespace changes`}
```

source/features/faster-pr-diff-options.tsx
```diff
@@ -1,13 +1,17 @@
...
 import features from '.';
 
+function isPRPage(): boolean {
+	return !pageDetect.isSingleCommit() && !pageDetect.isCompare();
+}
+
 function createDiffStyleToggle(): DocumentFragment {
 	const url = new URL(location.href);
-	const isUnified = select.exists([
+	const isUnified = url.searchParams.get('diff') === 'unified' || select.exists([
```
fregante, line 14: This seems redundant. Does it work without this check? The next line already has a selector for "form in PR"

source/features/faster-pr-diff-options.tsx
```diff
@@ -1,13 +1,17 @@
 import React from 'dom-chef';
 import select from 'select-dom';
 import * as pageDetect from 'github-url-detection';
-import {BookIcon, CheckIcon, DiffIcon} from '@primer/octicons-react';
+import {BookIcon, CheckIcon, DiffIcon, DiffModifiedIcon} from '@primer/octicons-react';
 
 import features from '.';
 
+function isPRPage(): boolean {
+	return !pageDetect.isSingleCommit() && !pageDetect.isCompare();
```
fregante, line 9: Can you use PR-related checks instead of exclusions? These 2 alone don't ensure it's a PR page

source/features/faster-pr-diff-options.tsx
```diff
@@ -16,15 +20,19 @@ function createDiffStyleToggle(): DocumentFragment {
...
+				aria-label={`Switch to the ${type} diff view`}
 				href={url.href}
 			>
 				{icon}
 			</a>
 		);
 	}
 
+	if (isPRPage()) {
+		return <>{makeLink(isUnified ? 'split' : 'unified', isUnified ? <BookIcon/> : <DiffIcon/>, false)}</>;
```
fregante, line 33:
```suggestion
		return isUnified ?
			makeLink('split', <BookIcon/>, false) :
			makeLink('unified', <DiffIcon/>, false);
```

source/features/faster-pr-diff-options.tsx
```diff
@@ -47,41 +55,41 @@ function createWhitespaceButton(): HTMLElement {
 		<a
 			href={url.href}
 			data-hotkey="d w"
-			className={`btn btn-sm btn-outline tooltipped tooltipped-s ${isHidingWhitespace ? 'bg-gray-light text-gray-light color-text-tertiary' : ''}`}
+			className={'tooltipped tooltipped-s ' + (isPRPage() ? `d-none d-lg-block color-icon-secondary ${isHidingWhitespace ? '' : 'color-icon-info'}` : `btn btn-sm btn-outline tooltipped ${isHidingWhitespace ? 'bg-gray-light text-gray-light color-text-tertiary' : ''}`)}
```
fregante, line 58: Nested ternary operators and infinite-length line. Can be improved. I'd rather see whole strings instead of nested ones. e.g.

```js
const classes = sPRPage() ?
	'tooltipped tooltipped-s blah blah' :
	'tooltipped tooltipped-s foo foo';
```

And then

```suggestion
			className={classes}
```

source/features/faster-pr-diff-options.tsx
```diff
@@ -47,41 +55,41 @@ function createWhitespaceButton(): HTMLElement {
...
+		</div>
+	);
 }
 
 function init(): false | void {
-	const container = select([
-		'#toc', // In single commit view
-		'.pr-review-tools' // In review view
-	]);
+	const container = isPRPage() ? select('.js-file-filter')?.closest('.flex-auto') : select('#toc');
```
fregante, line 79:
```suggestion
	const container = isPRPage() ?
		select('.js-file-filter')?.closest('.flex-auto') :
		select('#toc');
```

source/features/faster-pr-diff-options.tsx
```diff
@@ -107,6 +113,13 @@ function init(): false | void {
 		// Make space for the new button by removing "Changes from" #655
```
fregante, line 113: This no longer works, can you update it?

source/features/faster-pr-diff-options.tsx
```diff
@@ -107,6 +113,13 @@ function init(): false | void {
...
 		select('[data-hotkey="c"]')!.firstChild!.remove();
 	}
+
+	// Remove extraneous padding on "Clear filters" button
+	const clearFiltersButton = select('.subset-files-tab');
+	if (clearFiltersButton) {
+		clearFiltersButton.classList.remove('px-sm-3');
+		clearFiltersButton.classList.add('ml-2');
```
fregante, line 121:
```suggestion
		clearFiltersButton.classList.replace('px-sm-3', 'ml-2');
```

or 
```js
select('.subset-files-tab')?.classList.replace('px-sm-3', 'ml-2');
```

source/features/faster-pr-diff-options.tsx
cheap-glitch: This fixed a bug I noticed while working on this PR:

> Also, noticed this bug where the wrong diff style button is selected when "No whitespace" is enabled (but only on compare and commit pages), e.g. : https://github.com/sindresorhus/refined-github/compare/21.4.13...21.4.23?w=1.

I'll look into it to see if this can be simplified.

source/features/faster-pr-diff-options.tsx
fregante: Not for me, anywhere:

Screen Shot

Example URL: https://github.com/sindresorhus/refined-github/pull/4297/files/532f870c4b490c0b76c6c368d5bde6ffaff3465d..87445f7eb9ef94e7147249957b8e5c6ef88779aa

source/features/faster-pr-diff-options.tsx
cheap-glitch: The second selector needed fixing (the `diff` parameter in the URL doesn't necessarily occurs at the end).

source/features/faster-pr-diff-options.tsx
```diff
@@ -43,44 +53,27 @@ function createWhitespaceButton(): HTMLElement {
...
-	}
-
-	container.prepend(
-		wrap(
-			createDiffStyleToggle(),
-			createWhitespaceButton()
-		)
+function initPR(): false | void {
+	const container = select('.js-file-filter')!.closest('.flex-auto')!;
+	container.append(
```
fregante, line 74: Variable not necessary
```suggestion
	select('.js-file-filter')!.closest('.flex-auto')!.append(
```

source/features/faster-pr-diff-options.tsx
```diff
@@ -90,31 +83,43 @@ function init(): false | void {
...
+	// Remove extraneous padding around "Clear filters" button
+	select('.subset-files-tab')?.classList.replace('px-sm-3', 'ml-2');
+}
 
-	const prUI = select('.js-diff-settings');
-	if (prUI) {
-		prUI.closest('details')!.remove();
+function initCommitAndCompare(): false | void {
+	const container = select('#toc')!;
+	container.prepend(
```
fregante, line 96: Variable not necessary

source/features/faster-pr-diff-options.tsx
```diff
@@ -90,31 +83,43 @@ function init(): false | void {
...
-	const singleCommitUI = select('[data-ga-load^="Diff, view"]');
-
-	if (singleCommitUI) {
-		singleCommitUI.remove();
-		return;
-	}
+	// Only show the native dropdown on medium and small screens #2597
+	select('.js-diff-settings')!.closest('details')!.classList.add('d-lg-none');
+	// Make space for the new button by removing "Changes from" #655
+	(select('.js-commits-filtered') ?? select('[data-hotkey="c"]'))!.firstChild!.remove();
```
fregante, line 89: Why not merge the selectors? `'.js-commits-filtered,[data-hotkey="c"]'`

source/features/faster-pr-diff-options.tsx
```diff
@@ -90,31 +83,43 @@ function init(): false | void {
...
+	// Remove extraneous padding around "Clear filters" button
+	select('.subset-files-tab')?.classList.replace('px-sm-3', 'ml-2');
```
fregante, line 91: Is `?.` still needed since we split the init?

Also there should be a space above each comment in this block to improve readability

source/features/faster-pr-diff-options.tsx
cheap-glitch: > Why not merge the selectors?

When no commits are selected, the text is directly in the `<summary>` element, but when some commits are selected it's wrapped in a `<span>`:
```html
<summary class="btn-link Link--muted select-menu-button" data-hotkey="c" aria-haspopup="menu" role="button">
      <span class="color-text-link js-commits-filtered">
        Changes from <strong>3 commits</strong>
      </span>
</summary>
```

So `'.js-commits-filtered, [data-hotkey="c"]'` will always select the `<summary>` which is not what we want. There may be a better way to do this though, but I didn't find any.

source/features/faster-pr-diff-options.tsx
cheap-glitch: > Is ?. still needed since we split the init?

Yes because "Clear filters" is only there when there are some filters enabled.

> Also there should be a space above each comment in this block to improve readability

:+1:

source/features/faster-pr-diff-options.tsx
fregante: What about `[data-hotkey="c"] strong` + `.previousSibling`?

source/features/faster-pr-diff-options.tsx
```diff
@@ -90,31 +82,44 @@ function init(): false | void {
...
-		return;
-	}
+	// Make space for the new button by removing "Changes from" #655
+	(select('.js-commits-filtered') ?? select('[data-hotkey="c"]'))!.firstChild!.remove();
 
-	const prUI = select('.js-diff-settings');
-	if (prUI) {
-		prUI.closest('details')!.remove();
+	// Remove extraneous padding around "Clear filters" button
+	select('.subset-files-tab')?.classList.replace('px-sm-3', 'ml-2');
```
fregante, line 92: We need to preserve the media query, I'm not sure if this is right:

```suggestion
	select('.subset-files-tab')?.classList.replace('px-sm-3', 'ml-sm-2');
```

Without this, we're _adding_ margin on windows <544px

fregante (approved):
After the 2 reviews I think this is finally ready 😁

fregante:
By the way, they just added yet another widget on that line 😃 

https://twitter.com/GHchangelog/status/1397299450712821763

fregante:
Thanks for working on this by the way, it was an ooooold issue

PR #4300, opened 2021-04-30
Added additional compatible Chromium browsers

Oliver-Saer:
Added due to also being popular Chromium browser choices.

- Vivaldi (https://vivaldi.com)
- Brave (https://brave.com)
- Chromium (https://www.chromium.org)

## Test URLs
N/A

## Screenshot
N/A


yakov116:
I would think that its understandable if its compatible with chrome its compatible with Chromium.

Not only that the link you are giving is the same link we are giving for chrome.

Oliver-Saer:
Yes, I deliberately used the same link because that's the precedent which was set for the existing Edge and Opera entries.

fregante:
Thanks for the PR, but those are _fringe_ browsers at best. We don't need to list all of the compatible browsers, users should already know that their browser is compatible with Chrome if they land on this repository.

Oliver-Saer:
No problem, thank you for considering my PR and for creating this extension - it works great!

PR #4316, opened 2021-05-04
Load features faster

### LINKED ISSUE #4312: Custom CSS is injected too late
This line is waiting for the tab to be visible before injecting the CSS (via `elementReady('body')`)

https://github.com/sindresorhus/refined-github/blob/1d9dc2fd45c8ea13d898ffd70e7bc509feaa2e01/source/features/index.tsx#L131

It should probably be moved to the top of the function, without waiting for any conditions.

Also I'd replace `elementReady('body')` with a custom awaiter like:

```js
while (!document.body) {
	await Promise.race([new Promise(setTimeout), new Promise(requestAnimationFrame)])
}
```

yakov116:
Closes #4312 

From 280ms to 39.458984375 ms

## Test URLs
All

## Screenshot

None

### DIFF
--- a/source/features/index.tsx
+++ b/source/features/index.tsx
@@ -1,13 +1,14 @@
 import React from 'dom-chef';
 import cache from 'webext-storage-cache';
+import delay from 'delay';
 import select from 'select-dom';
 import domLoaded from 'dom-loaded';
 import stripIndent from 'strip-indent';
 import {Promisable} from 'type-fest';
-import elementReady from 'element-ready';
 import compareVersions from 'tiny-version-compare';
 import * as pageDetect from 'github-url-detection';
 
+import {frame} from '../helpers/dom-utils';
 import onNewComments from '../github-events/on-new-comments';
 import bisectFeatures from '../helpers/bisect';
 import optionsStorage, {RGHOptions} from '../options-storage';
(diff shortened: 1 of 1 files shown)


source/features/index.tsx
```diff
@@ -82,7 +83,16 @@ let logError = (id: FeatureID, error: unknown): void => {
 
 // eslint-disable-next-line no-async-promise-executor -- Rule assumes we don't want to leave it pending
 const globalReady: Promise<RGHOptions> = new Promise(async resolve => {
-	await elementReady('body', {waitForChildren: false});
+	while (!document.body) {
+		// eslint-disable-next-line no-await-in-loop
+		await Promise.race([delay(10), frame()]);
+	}
```
fregante, line 89: If frame resolves, we definitely have body too.

```suggestion
	const oneFrame = frame();
	while (!document.body) {
		// eslint-disable-next-line no-await-in-loop
		await Promise.race([delay(10), oneFrame]);
	}
```

Maybe there's a "better" way to write it though


```suggestion
	await Promise.race([frame(), pWaitFor(() => document.body)]);
```

but unless we already have the dependency, I'd rather not

https://www.npmjs.com/package/p-wait-for

yakov116: > but unless we already have the dependency

its not

source/features/index.tsx
```diff
@@ -82,20 +83,36 @@ let logError = (id: FeatureID, error: unknown): void => {
...
+	log = options.logging ? console.log : () => {/* No logging */};
+
+	// Wait for the document to be at least partially available
+	const oneFrame = frame();
+	while (!document.body) {
+		// eslint-disable-next-line no-await-in-loop
+		await Promise.race([delay(10), oneFrame]);
+	}
+
+	if (pageDetect.is500() || document.title === 'Confirm password' || document.title === 'Confirm access') {
```
fregante, line 114: Next: https://github.com/fregante/github-url-detection/issues/76

fregante (approved):

yakov116:
Title good?

PR #4321, opened 2021-05-06
Don't hide tagged commits in `dim-bots`

### LINKED ISSUE #4318: `dim-bots` hides GitHub Actions commits
https://github.com/fregante/github-url-detection/commits/master?after=e58ce98ffd6a45822d035598eda8f699e1839f06+1&branch=master

Screen Shot

Notice the 4.11.0 commit being dimmed.

cheap-glitch:
Fixes #4318 

## Test URLs
https://github.com/fregante/github-url-detection/commits/master?after=e58ce98ffd6a45822d035598eda8f699e1839f06+1&branch=master

## Screenshot

<table>
<tr>
	<th>Before
	<td>

image
<tr>
	<th>After
	<td>

image
</table>

The tags are lazy-loaded. Not sure using `oneMutation` is the best approach though.

### DIFF
--- a/source/features/dim-bots.tsx
+++ b/source/features/dim-bots.tsx
@@ -33,7 +33,7 @@ async function init(): Promise<void> {
 	}
 
 	for (const tag of select.all('.js-navigation-container .octicon-tag')) {
-		tag.closest('.commit, .Box-row')!.classList.remove('rgh-dim-bot');
+		tag.closest('.rgh-dim-bot')?.classList.add('tagged');
 	}
 }
 
--- a/source/features/dim-bots.tsx
+++ b/source/features/dim-bots.tsx
@@ -1,6 +1,5 @@
 import './dim-bots.css';
 import select from 'select-dom';
-import oneMutation from 'one-mutation';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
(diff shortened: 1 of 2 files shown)


source/features/dim-bots.tsx
```diff
@@ -15,13 +16,25 @@ const botSelector = [
...
+async function init(): Promise<void> {
 	for (const bot of select.all(botSelector)) {
 		// Exclude co-authored commits
 		if (select.all('a', bot.parentElement!).every(link => link.matches(botSelector))) {
 			bot.closest('.commit, .Box-row')!.classList.add('rgh-dim-bot');
 		}
 	}
+
+	if (!tagsLoaded()) {
+		await oneMutation(document.body, {subtree: true, childList: true, filter: tagsLoaded});
```
cheap-glitch, line 32: Maybe we need some kind of timeout here? Since the promise can hang forever on commit lists with no tags.

fregante:
No, here you just need to add a class to the .rgh-dim-bot element and then just handle it via CSS (:not(tagged))

We should almost never need to wait for mutations since, at worst, we can fire custom events and listen to them. One example is the “markdown source” feature

cheap-glitch:
> We should almost never need to wait for mutations

I'm confused :thinking: Since the tags are lazy-loaded, don't we need some kind of listener? Or should this be a simple `setTimeout`?

cheap-glitch:
Closing this as I won't have time to work on it.

fregante:
> Since the tags are lazy-loaded

The tags are not lazy loaded: Refined GitHub is adding the tags. The difference is that since we add them, we also control the code that adds them. In short, the `.rgh-tagged` class should be added by `tags-on-commits-list`, not here.

PR #4327, opened 2021-05-07
Fix console error in `forked-to`

### LINKED ISSUE #4315: `forked-to`: Cannot set property 'href' of undefined
URL: https://github.com/sindresorhus/refined-github/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc

```js
TypeError: Cannot set property 'href' of undefined
  at init (chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/build/refined-github.js:6898:92)
  at async runFeature (chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/build/refined-github.js:1946:18)
  at async setupPageLoad (chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/build/refined-github.js:1959:34)
```
	

When clicking history.back()

yakov116:
Fixes #4315

## How to test

Have 2 forks and do history.back()

## Screenshot

None


fregante:
Forked-details already contains a forked-button. I added that class on both for this specific reason, why isn’t it matched?

yakov116:
Good question I dont know 😕. Maybe it fires too early for it to be caught?

fregante:
Thing is, if that isn't matched, than this shouldn't match either.

yakov116:
But it does I dont get an error after this change!

fregante:
Screen Shot

This PR is adding a `rgh-forked-details` class immediately before the existing `rgh-forked-button` element; It's just duplicating existing functionality. I'd rather see a PR that figures out why the existing check does not work.

yakov116:
Figured it out. Drop down will have the leading slash.

fregante (approved):
🎉

fregante:
Actually, I’m confused. They’re created with the same `createLink` function, how can they be different?

fregante:
> Dropdown does not have the leading slash

Rather than `dropdown` it's that the URLs are `/user/repo` on the root and `/user/repo/blob/file` on other pages. So sometimes it matched, sometimes it didn't. It sounds like it was not related to the dropdown at all

yakov116:
> Rather than `dropdown` it's that the URLs are `/user/repo` on the root and `/user/repo/blob/file` on other pages. So sometimes it matched, sometimes it didn't. It sounds like it was not related to the dropdown at all

Correct

PR #4330, opened 2021-05-07
Don't run `edit-readme` if no readme exists

yakov116:
## Test URLs
https://github.com/sindresorhus/refined-github/tree/main/build

## Screenshot
None


fregante:
Does this solve an issue? Select already checks for undefined in the baseElement and doesn’t run at all

fregante (approved):

Issue #4324, opened 2021-05-07
Add per-feature “unaffected” version in the hotfix mechanism


fregante:
Currently our hotfix system only lets us specify only one version at which ALL of the listed features are fixed. I never anticipated that we’d have to use hotfixes so often 😅 

I suggest using a new format that is also easy to write: CSV. GitHub might also render a table for this

```csv
fit-textareas,21.5.1
avoid-accidental-submissions
```

Notice how the second line, doesn’t have a second value, so it’s not yet fixed. Also it doesn’t have to start with `feature:` since this new file needs to be parsed and not just used as-is

PR #4341, opened 2021-05-10
Meta: Add per-feature “unaffected” version in the hotfix manager

fregante:
Fixes https://github.com/sindresorhus/refined-github/issues/4324

To test this PR, the manifest version needs to be manually changed. I tested it and the function correctly returns the expected option settings.

### DIFF
--- a/source/features/index.tsx
+++ b/source/features/index.tsx
@@ -10,7 +10,7 @@ import {frame} from '../helpers/dom-utils';
 import onNewComments from '../github-events/on-new-comments';
 import bisectFeatures from '../helpers/bisect';
 import optionsStorage, {RGHOptions} from '../options-storage';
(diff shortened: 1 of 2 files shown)


fregante:
To test this, the manifest version needs to be manually changed. I tested it and the function correctly returns the expected option settings.

fregante:
Will wait for https://github.com/sindresorhus/refined-github/pull/4316#issuecomment-838423309

yakov116:
I should have commented earlier.

Maybe we should be able to put in an issue number. Like we used to have and let the users know so we can get some help on the issue?

fregante:
Yeah that would be nice, it can be shown as a notice in the options page:

```
 ------------------------------------------------------
| `dim-bots` has been temporarily disabled due to #123 |
| `bugs-tab` has been temporarily disabled due to #456 |
 ------------------------------------------------------
```

PR #4346, opened 2021-05-10
Remove style on label as no longer needed

### LINKED ISSUE #4339: `remove-label-faster` has the wrong hover color
1. Visit https://github.com/sindresorhus/refined-github/issues/4144 (need repo write access)
2. Hover gray label

gif

MaxElkins:
## Related issue

Closes #4339 

## What I did

- I made use of an existing css variable (--lightness-switch) which is used to determine the text color for labels.
- I inverted what this variable returns so the fill on the X will be opposite to the background color of the icon (determined by --lightness-switch). 

## Screenshot
Screenshot from 2021-05-10 17-50-10

## Note
This is my first pull request so I'm hoping I'm doing this right, I've read all the guidelines I could find but please let me know if there something I should do differently. Thanks! :D

### DIFF
--- a/source/features/remove-label-faster.css
+++ b/source/features/remove-label-faster.css
@@ -16,7 +16,7 @@
 
 .rgh-remove-label-faster:is(:focus, :hover) svg {
 	background-color: currentColor;
-	fill: var(--rgh-remove-label-faster-color);
+	fill: hsl(0, 0%, calc((1 - var(--lightness-switch)) * 100%));
 }
 
 :root:root .rgh-remove-label-faster-already-added {
--- a/source/features/remove-label-faster.tsx
+++ b/source/features/remove-label-faster.tsx
@@ -1,71 +1,82 @@
-import './remove-label-faster.css';
-import React from 'dom-chef';
-import {XIcon} from '@primer/octicons-react';
-import select from 'select-dom';
-import onetime from 'onetime';
-import delegate from 'delegate-it';
-import {observe} from 'selector-observer';
-import * as pageDetect from 'github-url-detection';
+import "./remove-label-faster.css";
+import React from "dom-chef";
+import { XIcon } from "@primer/octicons-react";
+import select from "select-dom";
+import onetime from "onetime";
+import delegate from "delegate-it";
+import { observe } from "selector-observer";
+import * as pageDetect from "github-url-detection";
 
-import features from '.';
-import * as api from '../github-helpers/api';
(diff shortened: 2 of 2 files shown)


MaxElkins:
> Please run `npm run lint` there are many unrelated changes.

I think that's sorted now!

How would I find the formatting config for this project so I could change my auto formatting?

Thanks!

source/features/remove-label-faster.css
```diff
@@ -16,7 +16,7 @@
 
 .rgh-remove-label-faster:is(:focus, :hover) svg {
 	background-color: currentColor;
-	fill: var(--rgh-remove-label-faster-color);
+	fill: hsl(0deg, 0%, calc((1 - var(--lightness-switch)) * 100%));
```
fregante, line 19: Thank you for the PR! Can you check if there’s a GitHub-native variable we can use instead of HSL? They have a lot of dynamic variables that I’m sure are useful for  this

fregante: Can you find out why the previous piece of code is now undefined? I think it’s best to keep using that since it used to work best (reusing the actual color of the label instead of just black/white).

MaxElkins:
The X color now reflects the label color

Screenshot 2021-05-11 at 11 32 58
Screenshot 2021-05-11 at 11 33 03

However I don't think this is an ideal solution.

Previously setting the variable as `'--rgh-remove-label-faster-color': label.style.backgroundColor` was coming out as undefined.

I'm getting around this by setting the variable 'manually' to the label's background 
color. `rgb(var(--label-r),var(--label-g),var(--label-b))` which uses githubs own css variables.

This is being defined in-line in the tsx however it could just be moved to css. I kept it there as reference for now.

I understand this is probably not as futureproof but I'm struggling to figure out why `'--rgh-remove-label-faster-color': label.style.backgroundColor` is coming out as undefined, probably as my typescript skills are not very strong yet! I've changed the PR to draft for now.

Hope this helps in some way!

fregante:
> This is being defined in-line in the tsx however it could just be moved to css

Yep that's better. They started using inline variables instead of setting `background-color` inline, that's why it broke.

MaxElkins:
> They started using inline variables instead of setting `background-color` inline, that's why it broke.

I see! Thanks for explaining. In that case, as we can reference the in-line variables, this solution works?

> Yep that's better. 

Sorry which one is better? haha 😅 Keep it in typescript or move to css?

yakov116:
Thanks @MaxElkins

Issue #4343, opened 2021-05-10
Window/Tab freeze in Firefox (conflict with other extensions)


azdanov:
**
**Fixed in https://github.com/sindresorhus/refined-github/releases/tag/21.5.10**
**

After updating to the latest version https://github.com/sindresorhus/refined-github/releases/tag/21.5.9 I couldn't load GitHub website (infinite load, unresponsive window/tab and browser offering to stop the Refined GitHub extension).

<img width="824" alt="Screenshot 2021-05-10 at 11 02 22" src="https://user-images.githubusercontent.com/6123841/117626022-5500a280-b17f-11eb-9900-284116ca9ae2.png">

I'm using Firefox Developer Edition `89.0b10 (64-bit)` on macOS Big Sur `11.3.1 (20E241)`

~Couldn't reproduce on Windows with same browser.~ Started to happen on Windows 10 also.

Disabling `reactions-avatars` helps sometimes, but not always.

marcphilipp:
This is also happening to me with the regular version of Firefox after upgrading to 21.5.9.

azdanov:
It happens even here: https://github.com/sindresorhus/refined-github/issues/4343
As for "Find feature", I can't do it since the github tab is frozen and doesn't refresh.

gallais:
I am experiencing the same issue on Ubuntu and firefox is similarly offering to temporarily suspend the extension.

* Firefox 88.0
* Refined github 21.5.9

I ran 'find feature' on two different pages and the verdict was 'warning-for-disallow-edits' both times.

fregante:
Thank you! `warning-for-disallow-edits` isn't run on this page so that's likely not the case.

Are other extensions enabled? 

Does it happens spontaneously on load/reload or do you have to click/do something?

Do you see anything on this page that might have triggered it? https://github.com/sindresorhus/refined-github/releases/tag/21.5.9

fregante:
What happens when you click on "Debug script"? Can you take a screenshot of the code it shows?

s-KaiNet:
I have the same problem - version 21.5.9, FF - 88.0.1

s-KaiNet:
> 
> 
> I have the same problem - version 21.5.9, FF - 88.0.1, Win10
> 
> However, I cannot reproduce it all the time. For example, it freezes even on the home page for gists, ie. `https://gist.github.com/`, after few restarts (I also used "clear cache" and "Find feature" from the add-in options), now I can open gists without a problem.

[...]

PR #4352, opened 2021-05-12
Load features faster /2

yakov116:
Tested takes between **0.710205078125** ms and 18.29931640625 ms 

Follows https://github.com/sindresorhus/refined-github/pull/4316#issuecomment-838423309
Fixes https://github.com/sindresorhus/refined-github/issues/4343

### DIFF
--- a/source/features/index.tsx
+++ b/source/features/index.tsx
@@ -5,7 +5,6 @@ import select from 'select-dom';
 import domLoaded from 'dom-loaded';
 import stripIndent from 'strip-indent';
 import {Promisable} from 'type-fest';
-import elementReady from 'element-ready';
 import compareVersions from 'tiny-version-compare';
 import * as pageDetect from 'github-url-detection';
 
@@ -104,17 +103,10 @@ const globalReady: Promise<RGHOptions> = new Promise(async resolve => {
 	// Create logging function
 	log = options.logging ? console.log : () => {/* No logging */};
 
-	await Promise.race([
-		// With comments
-		elementReady('body', {waitForChildren: false}),
-
-		// Maybe written better?
-		(async () => {
-			while (!document.body) {
-				await delay(10);
-			}
-		})()
-	]);
+	while (!document.body) {
+		// eslint-disable-next-line no-await-in-loop
+		await delay(10);
+	}
 
 	if (pageDetect.is500() || pageDetect.isPasswordConfirmation()) {
 		return;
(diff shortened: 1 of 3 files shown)


source/features/index.tsx
```diff
@@ -82,20 +83,41 @@ let logError = (id: FeatureID, error: unknown): void => {
...
-	if (document.body.classList.contains('logged-out')) {
-		console.warn('Refined GitHub is only expected to work when you’re logged in to GitHub. Errors will not be shown.');
-		features.error = () => {/* No logging */};
-		logError = () => {/* No logging */};
+	// Create logging function
+	log = options.logging ? console.log : () => {/* No logging */};
+
+	await Promise.race([
+		// With comments
+		elementReady('body', {waitForChildren: false}),
```
yakov116, line 109: @fregante why do we need this at all?

fregante: rAF might still be called earlier than setTimeout, in my mind. 

I wonder if we should just use oneMutation on `document.documentElement` to wait for the body. This would avoid any timers

yakov116: No freeze on firefox but takes 30 - 50ms

yakov116: No freeze on firefox but takes 30 - 50ms

yakov116:
@fregante I did some real heavy testing.

Using just the while works perfectly.

I tried using oneMutation but it fires waay to early in chrome (0.001ms).

I Tested every which way I can in firefox (large issues, large PR's) I can say that this is working very well.

fregante:
> I tried using oneMutation but it fires waay to early in chrome

Did you put a function filter? oneMutation will fire early because there are a lot of changes in `head` while it loads, so it needs a check on `document.body?.firstChild` probably. I'm not sure if `document.body` is immediately available in every browser or not.

yakov116:
> I just hope we don't fall into the infinite loop that caused #434

What? Did you mean to link another issue?

fregante:
Link fixed

cheap-glitch:
> I Tested every which way I can in firefox (large issues, large PR's) I can say that this is working very well.

Did you test it with other extensions enabled at the same time? I remember the freeze issue disappearing for me when disabling [uBlock](https://ublockorigin.com).

yakov116:
> Did you test it with other extensions enabled at the same time? I remember the freeze issue disappearing for me when disabling [uBlock](https://ublockorigin.com)

Yes. Can you try to test this?

cheap-glitch:
> Can you try to test this?

Tested with and without other extensions on Firefox 88.0.1, I didn't see any problems :tada: (Then again the freeze was very sporadic for me to begin with)

yakov116:
@fregante feel free to merge I will be out until Wednesday

Issue #4135, opened 2021-03-19
Clean up `parse-backticks` selectors


jack1142:
`parse-backticks` rendering in title is now native!
https://twitter.com/natfriedman/status/1372917225993080849

yakov116:
Thanks #4129

jack1142:
Ah, sorry then, I didn't notice 😬 
Closing...

yakov116:
No worries

kidonng:
`parse-backticks` also native for commits, post here so it doesn't get lost:

![image](https://user-images.githubusercontent.com/44045911/116532280-a5236d80-a912-11eb-9c16-e04c3f14857f.png)

yakov116:
I don't like the way it looks and its buggy

fregante:
We can extend `parse-backticks` to detect them and style them correctly. Ideally just via CSS

kidonng:
> its buggy

If you refer to this:

![image](https://user-images.githubusercontent.com/44045911/116716572-7ee0f900-aa0a-11eb-8f57-a19df9a532c6.png)

(The issue is that if the commit title has a reference (issue/PR), the markup becomes `code > a` instead of the styled `a > code`)

fregante:
However it happens, we can fix it via CSS. Hopefully it will be a temporary issue from GitHub

fregante:
I opened a more specific issue

PR #4361, opened 2021-05-14
Lint 

yakov116:
Replace Octoface icon with toast icon - Icon was dropped
Clean up `parse-backticks`


yakov116:
Note: This is the last lint before dropping refresh. 🎉

fregante:
> Clean up [`parse-backticks`](/sindresorhus/refined-github/blob/main/source/features/parse-backticks.tsx)

Does that close #4135?

source/features/bugs-tab.tsx
```diff
@@ -29,7 +35,7 @@ const countBugs = cache.function(async (): Promise<number> => {
 		search(type: ISSUE, query: "label:bug is:open is:issue repo:${getRepo()!.nameWithOwner}") {
 			issueCount
 		}
-	`);
+	`) as Search;
```
fregante, line 38: Can we not?

source/features/conversation-activity-filter.tsx
```diff
@@ -147,17 +146,22 @@ function addWidget(position: Element): void {
...
 
-function init(): void {
-	const observer = observe(':is(.gh-header-sticky .meta, .gh-header-meta .flex-auto):not(.rgh-conversation-activity-filter)', {
-		add: addWidget
-	});
-	deinit.push(observer.abort);
+async function init(): Promise<void> {
+	const headerSeectors = ':is(.gh-header-sticky .meta, .gh-header-meta .flex-auto):not(.rgh-conversation-activity-filter)';
+	await elementReady(headerSeectors);
+	const headers = select.all(headerSeectors);
```
fregante, line 152: These 2 aren't compatible. `element-ready` resolves on the first one, so that means `.all` won't match the second one on some page loads

source/features/parse-backticks.tsx
```diff
@@ -20,16 +20,12 @@ function init(): void {
 		'.js-details-container .pr-1 > code', // `isCompare` pushed commits
 		'.Box-row .mb-1 a', // `isCompare` open Pull Request title
 		'.blame-commit-message', // `isBlame` commit message
-		'a[id^="issue_"]', // `isConversationList` issue and PR title
-		'.TimelineItem-body > del', // `isIssue`, `isPRConversation` title edits
-		'.TimelineItem-body > ins', // `isIssue`, `isPRConversation` title edits
 		'[id^=ref-issue-]', // `isIssue` issue and PR references
 		'[id^=ref-pullrequest-]', // `isPRConversation` issue and PR references
```
fregante, line 24: I think more can be dropped here. I think for example these 2 select the commits that appear in conversations (https://github.com/sindresorhus/refined-github/pull/3812) and those are already formatted.

fregante:
> Make `conversation-activity-filter` faster

Does this fix https://github.com/sindresorhus/refined-github/issues/4365?

source/features/parse-backticks.tsx
yakov116: I did not finish this. I started doing it and then decided to wait until we drop GHE. There is just to much there and I want to wait. (2 more weeks)

source/features/bugs-tab.tsx
yakov116: K I had this from last lint and I forgot to commit it. I will revert.

yakov116:
> Does this fix #4365?

Will check.

fregante:
Another lint can be changing repeated `:not(a):not(b)` to `:not(a,b)`, increasing the minimum browser version to what's specified in this table for "selector list" https://developer.mozilla.org/en-US/docs/Web/CSS/:not#specifications

It seems to work in Safari so it might be worth checking in other browsers as well

yakov116:
> It seems to work in Safari so it might be worth checking in other browsers as well

Does not work in firefox yet

source/features/conversation-activity-filter.tsx
yakov116: > These 2 aren't compatible. `element-ready` resolves on the first one, so that means `.all` won't match the second one on some page loads.

It works since I am not doing `waitForChildren:false`

cheap-glitch:
> > It seems to work in Safari so it might be worth checking in other browsers as well
> 
> Does not work in firefox yet

It does work in all major desktop browsers according to https://caniuse.com/css-not-sel-list.

yakov116:
My bad looks like I was on an old version of firefox 😨

source/features/quick-mention.tsx
```diff
@@ -29,7 +29,7 @@ function mentionUser({delegateTarget: button}: delegate.Event): void {
 function init(): void {
 	// `:first-child` avoids app badges #2630
 	// The hovercard attribute avoids `highest-rated-comment`
-	for (const avatar of select.all(`.TimelineItem-avatar > [data-hovercard-type="user"]:first-child:not([href="/${getUsername()}"]):not(.rgh-quick-mention)`)) {
+	for (const avatar of select.all(`.TimelineItem-avatar > [data-hovercard-type="user"]:first-child:not([href="/${getUsername()}"], .rgh-quick-mention)`)) {
```
fregante, line 32:
```suggestion
	const avatars = select.all(`.TimelineItem-avatar > [data-hovercard-type="user"]:first-child:not([href="/${getUsername()}"], .rgh-quick-mention)`);
	for (const avatar of avatars) {
```

fregante (approved):

fregante (changes requested):
Actually one previous review still unsolved

source/features/conversation-activity-filter.tsx
fregante: I think you're confused about how that works. `waitForChildren` has no effect on this. If anything, it's worse because `false` means it resolves as soon as the opening tag is found.

fregante: Given:

```html
<mango>
	<banana/>
</mango>
<pear></pear>
<mango></mango>
```

1. `elementReady('mango')` will resolve when line 4 is parsed. `select.all('mango')` will not include the second `<mango>`
1. `elementReady('mango', {waitForChildren:false})` will resolve when line 1 is parsed. `select.all('mango')` will definitely not include the second `<mango>`

yakov116: So I should just wait for the sticky, that is second?

distribution/manifest.json
```diff
@@ -8,7 +8,7 @@
 	"applications": {
 		"gecko": {
 			"id": "{a4c4eda4-fb84-4a84-b4a1-f7c1cbf2a1ad}",
-			"strict_min_version": "78.0"
+			"strict_min_version": "88.0"
```
yakov116, line 11: I could do 85. I just did 88, move it back?

source/features/conversation-activity-filter.tsx
fregante: If you know they’re exactly too, you can just use 2 elementReady, one customized for each

distribution/manifest.json
fregante: It looks like it’s 84

source/features/conversation-activity-filter.tsx
```diff
@@ -147,17 +146,21 @@ function addWidget(position: Element): void {
...
-	const observer = observe(':is(.gh-header-sticky .meta, .gh-header-meta .flex-auto):not(.rgh-conversation-activity-filter)', {
-		add: addWidget
-	});
-	deinit.push(observer.abort);
+async function init(): Promise<void> {
+	const headerSelectors = ':is(.gh-header-sticky .meta, .sticky-content .gh-header-meta .flex-auto):not(.rgh-conversation-activity-filter)';
+	await elementReady(headerSelectors);
+	for (const header of select.all(headerSelectors)) {
+		addWidget(header);
+	}
```
yakov116, line 154:
```suggestion
	await elementReady('.gh-header-sticky .meta');
	await elementReady('.sticky-content .gh-header-meta .flex-auto');
	for (const header of select.all(':is(.gh-header-sticky .meta, .sticky-content .gh-header-meta .flex-auto):not(.rgh-conversation-activity-filter)')) {
		addWidget(header);
	}
```

source/features/conversation-activity-filter.tsx
```diff
@@ -147,17 +146,25 @@ function addWidget(position: Element): void {
...
+async function init(): Promise<void> {
+	const mainHeader = await elementReady('.gh-header-meta .flex-auto:not([id])');
+	const stickyHeader = await elementReady('.gh-header-sticky .meta');
+	for (const header of [mainHeader, stickyHeader]) {
+		if (select.exists('.rgh-conversation-activity-filter', header)) {
+			continue;
+		}
+
+		addWidget(header!);
+	}
```
fregante, line 158: More like

```js
function init() {
  await addWidget(selector 1)
  await addWidget(selector 2)
}

Comment (yakov116) [reply source/features/conversation-activity-filter.tsx:158]: I aint done yet 😠. 

I had a better idea.

Comment (fregante) [reply source/features/conversation-activity-filter.tsx]: How's

```js
 	void addWidget(elementReady(selector 1);
 	void addWidget(elementReady(selector 2);
```

better than

```js
 	await addWidget(selector 1);
 	await addWidget(selector 2);
```

?

Also `await` is advised because that's how we correctly catch errors and link them to the feature name

yakov116: ahh did not think of that

fregante (approved):
Untested but it looks good. Only that `.closest('div')` looks a little loose, but it might be ok

Issue #4371, opened 2021-05-18
`pr-approvals-count` should hide the duplicate tooltip


fregante:
URL: https://github.com/sindresorhus/refined-github/pulls

<img width="149" alt="Screen Shot" src="https://user-images.githubusercontent.com/1402241/118632504-af8ea400-b7fa-11eb-82d2-0aa8fc64cb3d.png">

The feature already shows the tooltip inline. We should disable it. Something like:

```css
the right selector:is(:before, :after) {
	content: none;
}
```

PR #4374, opened 2021-05-20
Hide duplicate tooltip in `pr-approvals-count`

BrewedCoffee:
Resolves https://github.com/sindresorhus/refined-github/issues/4371

With this CSS change, the redundant tooltip is no longer present.

Tooltip for requested changes doesn't seem to be redundant as the count isn't shown inline, so this only targets 'approved' tooltips.

## Test URLs
URL: https://github.com/sindresorhus/refined-github/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc


source/features/pr-approvals-count.css
```diff
@@ -17,3 +17,7 @@
 .js-issue-row .color-text-secondary [aria-label*='changes'] {
 	color: var(--github-red) !important;
 }
+
+[aria-label*='approval']:after {
```
fregante, line 21: Let's match the existing selector and add a comment

```suggestion
/* Disable now-duplicate tooltip */
.js-issue-row .color-text-secondary [aria-label*='approval']:after {
```

fregante (approved):

fregante:
Thank you for your PR Jeff!

BrewedCoffee:
How can I run lint checks locally?

cheap-glitch:
> How can I run lint checks locally?

`npm run lint:js` and `npm run lint:css` (or `npm run lint` for both at the same time). All scripts are in package.json

BrewedCoffee:
Sorry to re-request, just need a maintainer to approve workflow tests.

fregante:
Thank you for the PR Jeff! :)

PR #4384, opened 2021-05-22
Fix duplication and cropped hovercard in `user-local-time`

### LINKED ISSUE #3634: `user-local-time` occasionally appears twice
I don't really know what's causing it and how to replicate it consistently, but the time appears twice. First as "Guessing…" then the time is updated at the same time.

### LINKED ISSUE #4345: Profile hover info is cut off
1. Navigate to https://github.com/hexalellogram/iPhoneBatteryClient
2. Hover over the profile image for the latest commit

Screen Shot 2021-05-08 at 22-1 57 51

It looks like adding the time makes the element too high for the window.

Refined GitHub v21.5.9

cheap-glitch:
Fixes #3634 
Fixes #4345 

## Test URLs

Steps to reproduce the bug:
1. Clear the extension cache
2. Go to a page with a lot of usernames, e.g. https://github.com/sindresorhus/refined-github/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc
3. Hover over the various usernames until the bug appears (it may take a while)

For #4345: https://github.com/hexalellogram/iPhoneBatteryClient

## Screenshot

https://user-images.githubusercontent.com/46634000/120510278-8f014500-c3c9-11eb-8473-c11fe6a42449.mp4

For the comment in the CSS:
image


source/features/user-local-time.tsx
```diff
@@ -80,9 +80,9 @@ function parseOffset(date: string): number {
 function init(): void {
 	const hovercard = select('.js-hovercard-content > .Popover-message')!;
 
-	observeElement(hovercard, async () => {
+	observeElement(hovercard, async ([mutation]) => {
 		if (
-			select.exists('.rgh-local-user-time', hovercard) || // Time already added
```
cheap-glitch, line 85: Dropped this condition since it's always false (the contents of the hovercard element are erased every time it disappears).

source/features/user-local-time.tsx
```diff
@@ -80,9 +80,9 @@ function parseOffset(date: string): number {
...
+			mutation?.addedNodes?.length !== 1 || // Avoid duplicated time element issue #3634
```
fregante, line 85: This condition is too broad. There are a number of situations where this condition could be false and repeated. 

I can think of 2 ways to solve this issue, either
- mark the container with a class as soon as the API call starts, or 
- use `mem()` so that it deduplicates call to the container (this is too hard to explain, but we have other examples of deduplication via mem)


In both cases, the “container” is what we select in: `select('div.d-flex.mt-3 > div.overflow-hidden.ml-3', hovercard)!.append(container)`

fregante: But actually we should just use observeSelector probably 🤔
We already know how to deduplicate that

cheap-glitch: > We already know how to deduplicate that

You mean adding a class and doing `:not(.this-class)` in the selector? I guess that could work if we add the class at the very beginning of the mutation handler.

fregante: Yes. Keep in mind that the handler will need to fire multiple times on the same hovercard element since only its content is updated

cheap-glitch:
Seems like #4345 is caused by `user-local-time`. Should I fix it in this PR while I'm at it?

yakov116:
Yes

yakov116:
Maybe its just me. But do you get the feeling the the init function should be split into 2?

source/features/user-local-time.tsx
```diff
@@ -78,67 +79,68 @@ function parseOffset(date: string): number {
...
-	observeElement(hovercard, async () => {
-		if (
-			select.exists('.rgh-local-user-time', hovercard) || // Time already added
-			!select.exists('[data-hydro-view*="user-hovercard-hover"]', hovercard) // It's not the hovercard type we expect
-		) {
-			return;
-		}
+	observe('.js-hovercard-content .Popover-message div.d-flex.mt-3 > div.overflow-hidden.ml-3:not(.rgh-user-local-time-container-added)', {
+		add: async hovercardContainer => {
+			const hovercard = hovercardContainer.closest<HTMLElement>('.Popover-message')!;
```
yakov116, line 84:
```suggestion
			const hovercard = hovercardContainer.closest('div.Popover-message')!;
```

does this work?

cheap-glitch:
> Maybe its just me. But do you get the feeling the the init function should be split into 2?

Yes I'll hoist the `add()` callback to the top level :+1:

yakov116:
LGTM

yakov116:
@cheap-glitch thanks for all your great work!

fregante:
🙌 2 bugs in 1!

PR #4385, opened 2021-05-22
Fix branch switcher conflict with `default-branch-button`

### LINKED ISSUE #4240: `default-branch-button` feature breaks the branch switcher
Here's a GIF:

And a link to the place where this can be reproduced:
https://github.com/File-New-Project/EarTrumpet/blob/2.1.8.0/README.md

cheap-glitch:
Fixes #4240 

## Test URLs

Go to https://github.com/sindresorhus/refined-github/tree/hotfix, click on the branch switcher and select an option above the readme header.

## Screenshot

https://user-images.githubusercontent.com/46634000/119224239-c89a9c00-bafd-11eb-9f2b-3a97987426e5.mp4

<p>&nbsp;</p>

The culprit:
image

### DIFF
--- a/source/features/default-branch-button.tsx
+++ b/source/features/default-branch-button.tsx
@@ -49,12 +49,9 @@ async function init(): Promise<false | void> {
 	}
 
 	branchSelector.parentElement!.before(defaultLink);
+	branchSelector.parentElement!.style.zIndex = 'auto'; // For #4240
 	groupButtons([defaultLink, branchSelector.parentElement!]).classList.add('d-flex');
 	branchSelector.style.float = 'none'; // Pre "Repository refresh" layout
-
-	// Fix #4240
-	branchSelector.style.zIndex = 'auto';
-	branchSelector.parentElement!.style.zIndex = 'auto';
 }
 
 void features.add(__filebasename, {
--- a/source/features/default-branch-button.tsx
+++ b/source/features/default-branch-button.tsx
@@ -1,4 +1,3 @@
-import './default-branch-button.css';
 import React from 'dom-chef';
 import elementReady from 'element-ready';
 import * as pageDetect from 'github-url-detection';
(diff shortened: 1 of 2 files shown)


fregante:
Ohhhh wow. That makes sense. But now I wonder why we’re not seeing this anywhere else on native buttons. Maybe they avoid it with a custom class?

fregante:
> why we’re not seeing this anywhere else on native buttons

I looked around the website a bit and `.BtnGroup` elements are rather rare, especially the ones with a dropdown inside, and this is probably the only situation where the z-index change causes this issue anyway.

fregante:
Can you try adding an inline CSS to the group?

`groupButtons` returns the element, so it's probably:

```ts
const groupedButtons = groupButtons();
groupedButtons.style.zIndex = '…' // comment

**Comment by cheap-glitch** [comment]
> Can you try adding an inline CSS to the group?

It doesn't work, the z-index of the wrapper doesn't affect the buttons since it's overridden by the focus selector. But we can do:
```js
branchSelector.style.zIndex = 'auto';
branchSelector.parentElement!.style.zIndex = 'auto';

```

source/features/default-branch-button.tsx
```diff
@@ -51,6 +51,10 @@ async function init(): Promise<false | void> {
 	branchSelector.parentElement!.before(defaultLink);
 	groupButtons([defaultLink, branchSelector.parentElement!]).classList.add('d-flex');
 	branchSelector.style.float = 'none'; // Pre "Repository refresh" layout
+
+	// Fix #4240
+	branchSelector.style.zIndex = 'auto';
+	branchSelector.parentElement!.style.zIndex = 'auto';
```
fregante, line 57: The first line doesn't seem necessary. Also this order makes aligns the two `branchSelector.parentElement!.` calls

```suggestion
	branchSelector.parentElement!.style.zIndex = 'auto'; // For #4240
	groupButtons([defaultLink, branchSelector.parentElement!]).classList.add('d-flex');
	branchSelector.style.float = 'none'; // Pre "Repository refresh" layout
```

fregante:
Thank you!

PR #4386, opened 2021-05-22
Meta: Fix `lint:css` error on Windows

BrewedCoffee:
When attempting to run lint checks from the top directory, I encountered the following error:

```
C:\Repos\refined-github>npm run lint:css

> lint:css
> stylelint 'source/**/*.css'

Error: No files matching the pattern "'source/**/*.css'" were found.
    at C:\Repos\refined-github\node_modules\stylelint\lib\standalone.js:212:12
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

After this fix, lint checks run as expected without errors.

## Note for reviewers

It's very possible that this is an error that only affects me, seeing as though the code hasn't changed a year. Please let me know if the error exists on your end!

Edit: Sorry for the spam title changing. Forgot that gets added to the log.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
 		"build:webpack": "webpack --mode=production",
 		"fix": "run-p 'lint:* -- --fix'",
 		"lint": "run-p lint:* --continue-on-error",
-		"lint:css": "stylelint 'source/**/*.css'",
+		"lint:css": "stylelint \"source/**/*.css\"",
 		"lint:js": "xo",
 		"pack:safari": "xcodebuild -project 'safari/Refined GitHub.xcodeproj'",
 		"start:safari": "open 'safari/build/Release/Refined GitHub.app'",


fregante:
cmd.exe is not a great terminal to use in 2021, I'd suggest switching to PowerShell, Cygwin, or WSL

darkred:
@fregante


I've just tried `PowerShell` version 5.1s (the windows built-in version) as well as latest 7.1.3
and in both the error occurs. (I use win10 21h1) 

I also tried latest `Cygwin` `echo "${BASH_VERSION}"` --> `4.4.12(3)-release` ​)
as well as `Git Bash` that comes with latest Git 2.31.1 installation (MINGW64, `echo "${BASH_VERSION}"` --> `4.4.23(1)-release` ​)
and in all the error unfortunately occurs.

So, currently, the only alternative solution I see for windows users would be to install WSL (which I haven't tried).

Maybe you'd like to reconsider merging this PR?

fregante:
If you restore the branch I can

darkred:
cc OP @BrewedCoffee

PR #4387, opened 2021-05-22
Prevent discussion link loss

FloEdelmann:
Closes #4357.

## Test URLs

Any page with a rich text editor, e.g. this one.

Try copying the following links into the comment field:

```
https://github.com/streetcomplete/StreetComplete/discussions/2789?sort=top#discussioncomment-646707
https://github.com/streetcomplete/StreetComplete/discussions/2789?sort=top
https://github.com/streetcomplete/StreetComplete/discussions/2789#discussioncomment-646707
https://github.com/streetcomplete/StreetComplete/discussions/2789
```

PR #4394, opened 2021-05-24
Meta: Show features disabled via hotfix in options page

### LINKED ISSUE #4389: Allow issues reference for hotfixes
Yeah that would be nice, it can be shown as a notice in the options page:

```
 ------------------------------------------------------
| `dim-bots` has been temporarily disabled due to #123 |
| `bugs-tab` has been temporarily disabled due to #456 |
 ------------------------------------------------------
```

CSV:

```csv
clean-sidebar,21.2.2,#789
dim-bots,,#123
bugs-tab,,#456
```

_Originally posted by @fregante in https://github.com/sindresorhus/refined-github/issues/4341#issuecomment-846377676_

cheap-glitch:
Closes #4389 

## Test URLs
n/a

## Screenshot
image

With (fake) issue numbers:
image


source/options.tsx
```diff
@@ -173,10 +174,36 @@ async function highlightNewFeatures(): Promise<void> {
...
 }
 
+async function getFeaturesDisabledViaHotfix(): Promise<HTMLElement[]> {
+	const hotfixes = await cache.get<string[][]>('hotfixes');
+	if (!hotfixes) {
+		return [];
+	}
+
+	const {version} = browser.runtime.getManifest();
+	const featureNames: Set<string> = new Set(features.map(({id}) => id));
```
cheap-glitch, line 184: Used a `Set` here only because the linter wanted one.

fregante: Alternatively you can `features.find({id} => id === featureName)` as a condition directly

source/options.tsx
```diff
@@ -173,10 +174,37 @@ async function highlightNewFeatures(): Promise<void> {
...
+	const disabledFeatures = [];
+	for (const [feature, unaffectedVersion] of hotfixes) {
+		const relatedIssue = '#1234';
+		if (featureNames.has(feature) && (version === '0.0.0' || !unaffectedVersion || compareVersions(unaffectedVersion, version) > 0)) {
```
fregante, line 188: `version === '0.0.0'` is constant, it mean it needs to be checked once at the top of this function

source/options.tsx
```diff
@@ -173,10 +174,37 @@ async function highlightNewFeatures(): Promise<void> {
...
+	const hotfixes = await cache.get<string[][]>('hotfixes');
+	if (!hotfixes) {
+		return [];
+	}
+
+	const {version} = browser.runtime.getManifest();
+	const featureNames: Set<string> = new Set(features.map(({id}) => id));
+	const disabledFeatures = [];
+	for (const [feature, unaffectedVersion] of hotfixes) {
+		const relatedIssue = '#1234';
```
fregante, line 187: This this PR incomplete/draft?

source/options.tsx
```diff
@@ -173,10 +174,37 @@ async function highlightNewFeatures(): Promise<void> {
...
+		if (featureNames.has(feature) && (version === '0.0.0' || !unaffectedVersion || compareVersions(unaffectedVersion, version) > 0)) {
+			disabledFeatures.push(
+				<p>
+					<code>{feature}</code> has been temporarily disabled
+					{relatedIssue ? <> due to <a href={'https://github.com/sindresorhus/refined-github/issues/' + relatedIssue.slice(1)}>{relatedIssue}</a></> : false}.
```
fregante, line 192: We control the data, let's never use `.slice(1)` to drop `#` after _we_ add it. 
```suggestion
					{relatedIssue ? <> due to <a href={'https://github.com/sindresorhus/refined-github/issues/' + relatedIssue}>#{relatedIssue}</a></> : false}.
```

Also perhaps we need a global utility function to generate issue links to our repo

source/options.tsx
cheap-glitch: That was for the demo screenshot, forgot to remove it :sweat:

source/options.tsx
cheap-glitch: > Also perhaps we need a global utility function to generate issue links to our repo

Same PR?

source/options.tsx
fregante: What I meant is that it should be an early return. If this, then return []

cheap-glitch: Oh so the notice shouldn't be shown at all while in dev mode? I think it could be useful though, especially with the links to the issues

fregante: Oh I'm reading the condition now. You're _showing_ the notice, not hiding it. In this case it would be wrong because `0.0.0` does not disable any feature, so it would say something incorrect

cheap-glitch: Fixed in [`22e99ee` (#4394)](https://github.com/sindresorhus/refined-github/pull/4394/commits/22e99ee9a7302b182a3a16c967a6dcc51dbb288f)

source/helpers/get-rgh-repo-issue-link.ts
```diff
@@ -0,0 +1,3 @@
+export default function getRghRepoIssueLink(issueId: number | string): string {
```
FloEdelmann, line 1: Maybe rename this to `getRghRepoIssueUrl` to avoid confusion? I used (nearly) the same function name in https://github.com/sindresorhus/refined-github/pull/4387/files#diff-32553a6819bdc0a556cfcf616e0677aa6ddf6aca02bf4c6b302f3f6e9d6a72a5R27 and that function returns an `HTMLAnchorElement` instead.

yakov116: shouldnt yours be called `createRGHIssueLink`?

FloEdelmann: Indeed. But making a clear distinction between *link* and *URL* is a good idea anyway IMO.

fregante:
It appears correct, but it needs testing before merging since it's so important, and I don't have a lot of time right now. 😅

yakov116:
I can do the testing. However can we edit the csv file? Will it mess anything up?

fregante:
Of course we can’t edit the CSV file since it affects everyone. Testing needs to be done locally by replacing the fetch with an inlined CSV for example

fregante:
Screen Shot 4

I reduced the spacing a bit

source/helpers/get-rgh-issue-url.ts
```diff
@@ -0,0 +1,3 @@
+export default function getRghIssueUrl(issueId: number | string): string {
+	return 'https://github.com/sindresorhus/refined-github/issues/' + issueId.toString();
+}
```
fregante, line 3: Can you merge this into the helper file added by https://github.com/sindresorhus/refined-github/pull/4387? 

This function can be used in background.ts, and the **link** function can be used in options.tsx

cheap-glitch (commented):
Otherwise `:not(:empty)` doesn't apply and we get an empty box:

image

fregante:
Good catch!

fregante:
Tested in Chrome and Firefox. Also ensure that the hot fixes still work.

Screen Shot 2
Screen Shot 1

fregante:
Thank you for working on this! 😁 

I wanted to merge this as is, but I was wondering if we should go back to how we used to display them (inline instead of in a notice):

https://github.com/sindresorhus/refined-github/blob/05de5a5c328deb988026b629c2098192c615b9d7/source/options.tsx#L51

cheap-glitch:
> I was wondering if we should go back to how we used to display them (inline instead of in a notice)

I don't have a strong opinion about this, but I think having a list at the top is slightly more convenient, both for users and maintainers.

fregante:
> having a list at the top is slightly more convenient

We can still move the features to the top of the list *and* disable the checkbox.

Issue #4246, opened 2021-04-15
Wiki: Show number of pages.


jtagcat:
There are many repos that have wiki enabled, yet they don't have anything in there.

Ideally, I'd prefer the wiki index below everything else, in the main repo page.  
![image](https://user-images.githubusercontent.com/38327267/114880875-fe0ee400-9df1-11eb-9c3a-5ebd7a74d540.png)

***

Though, here I'm proposing a number counter for the wiki.  
ex URL: https://github.com/koalaman/shellcheck

Before:  
![image](https://user-images.githubusercontent.com/38327267/114881125-3dd5cb80-9df2-11eb-8ba1-ab8b3d4fdcd6.png)

After:  
![image](https://user-images.githubusercontent.com/38327267/114881349-72e21e00-9df2-11eb-9a1c-d0a311ce47e2.png)

kidonng:
I don't think this is useful its own, but perhaps we can add the number or hide the Wiki Tab if it's empty, just like `hide-zero-packages`.

jtagcat:
I wrote a nice detailed thing and the tab crashed with oom. :^)

The number is useful on it's own, showing whether you should bother with the wiki (is the project likely to have a 4-question FAQ or full documentation in the wiki).

> Ideally, I'd prefer the wiki index below everything else, in the main repo page.

I did a mockup.

![image](https://user-images.githubusercontent.com/38327267/116051484-2efbde00-a668-11eb-878a-170fecee568b.png)

The problem is that the wiki is likely to break the scroll-along of the other content. Giving up on aesthetics, it could be placed on the left side:

![image](https://user-images.githubusercontent.com/38327267/116051990-bcd7c900-a668-11eb-8df5-cf73d7891df5.png)

On screens wide enough, it could be an additional column:

![image](https://user-images.githubusercontent.com/38327267/116051875-9ade4680-a668-11eb-9a85-5373bf15dbff.png)

yakov116:
FYI this is only possible via fetchDom there is no API for it

fregante:
> Ideally, I'd prefer the wiki index below everything else, in the main repo page.

That's a bit much, we can't add that

Issue #309, opened 2016-09-30
Remove Projects (0) from repo navigation


hkdobrev:
Since GitHub has introduced projects there is this Projects tab on every repo. I am yet to see a repository using them. So I'd suggest removing the tab if there are 0 projects. It would reduce clutter and you will only see it when a repository is actually using projects.

That said I understand why GitHub did it like that - showing it for every repository even though nobody is using it at first - they want people to know about the feature, to open the tab and see the intro copy there. However I think users of Refined GitHub are heavy GitHub users and already know about it.

Implementation should be fairly easy as GitHub already show the number of projects in the tab button.

sindresorhus:
👍

paulmolluzzo:
If we take out the navigation then how else will people discover Projects? We'd be kneecapping the discovery of a new feature that people worked hard on. 😢

sindresorhus:
@paulmolluzzo We don't have to do it right now, but eventually I'd like to see it gone when unused.

paulmolluzzo:
> @paulmolluzzo We don't have to do it right now, but eventually I'd like to see it gone when unused.

OK, I'm just offering a counterpoint. Again, I don't want to take away the opportunity for all the GH people that worked on this feature to have their work gain traction.

sindresorhus:
@paulmolluzzo It's a good point. I just assumed everyone knew about it from the blog post, but I'm sure that's only a small portion of users.

hkdobrev:
@paulmolluzzo I get your argument, but I don't think we're actually preventing people to know about it. Here are the two use cases I think about:

1. **People using the extension already**: They are certainly heavy users of GitHub, they might have even read the blog post or watched the GitHub Universe keynote. They certainly look at GitHub repositories at least a few days a week. I assume they already know about the projects feature, because they are used to how the tabs are ordered there. They saw that Refined GitHub has removed Pulse and added Releases. So they have also noticed the Projects tab and clicked it to look around. If they have started using it, this change would not affect the repositories where they already have projects.
[...]

PR #4395, opened 2021-05-24
Add `remove-actions-tabs` feature

shinenelson:
GitHub enabled Actions and the corresponding UI tab for all repositories
even when they do not have any workflows configured. This makes it
confusing when navigating through different repositories often
misleading to think that the repository has configured Actions
workflows. This feature removes the Actions tab from view for
repositories that do not have any Actions workflows configured.

## Test URLs

- https://github.com/shinenelson/refined-github
- https://github.com/mozilla/tabzilla

## Screenshot

I am not sure this feature requires a screenshot, but I can create one
if it is required.

### DIFF
--- a/source/features/remove-actions-tab.tsx
+++ b/source/features/remove-actions-tab.tsx
@@ -1,17 +1,26 @@
+import cache from 'webext-storage-cache';
 import elementReady from 'element-ready';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
+import {getRepo} from '../github-helpers';
 import * as api from '../github-helpers/api';
+
+const getWorkflowsCount = cache.function(async (): Promise<number> => {
+	const {workflows} = await api.v3('actions/workflows');
+	return workflows.length;
+}, {
+	maxAge: {days: 1},
+	cacheKey: (): string => __filebasename + ':' + getRepo()!.nameWithOwner
+});
 
 async function removeActionsTab(): Promise<void | false> {
 	const actionsTab = await elementReady('[data-hotkey="g a"]');
-	const {workflows} = await api.v3('actions/workflows');
 
 	if (
 		!actionsTab || // Actions Tab does not exist 🎉
 		actionsTab.matches('.selected') || // User is on Actions tab 👀
-		workflows.length > 0 // Actions Workflows are configured for the repository
+		await getWorkflowsCount() > 0 // Actions Workflows are configured for the repository
 	) {
 		return false;
 	}
--- a/source/features/remove-actions-tab.tsx
+++ b/source/features/remove-actions-tab.tsx
@@ -32,6 +32,10 @@ void features.add(__filebasename, {
 	include: [
 		pageDetect.isRepo
 	],
+	exclude: [
+		// Repo owners should see the tab. If they don't need it, they should disable actions altogether
+		pageDetect.canUserEditRepo
+	],
 	awaitDomReady: false,
 	init: removeActionsTab
 });
--- a/source/features/remove-actions-tab.tsx
+++ b/source/features/remove-actions-tab.tsx
@@ -1,0 +1,28 @@
+import elementReady from 'element-ready';
+import * as pageDetect from 'github-url-detection';
+
+import features from '.';
+import * as api from '../github-helpers/api';
+
+async function removeActionsTab(): Promise<void | false> {
+	const actionsTab = await elementReady('[data-hotkey="g a"]');
+	const {workflows} = await api.v3('actions/workflows');
+
+	if (
+		!actionsTab || // Actions Tab does not exist 🎉
+		actionsTab.matches('.selected') || // User is on Actions tab 👀
+		workflows.length > 0 // Actions Workflows are configured for the repository
+	) {
+		return false;
+	}
+
+	actionsTab.remove();
+}
+
+void features.add(__filebasename, {
+	include: [
+		pageDetect.isRepo
+	],
+	awaitDomReady: false,
+	init: removeActionsTab
+});
--- a/readme.md
+++ b/readme.md
@@ -142,6 +142,7 @@ Thanks for contributing! 🦋🙌
 - [](# "next-scheduled-github-action") [Shows the next scheduled time of relevant GitHub Actions in the workflows sidebar.](https://user-images.githubusercontent.com/46634000/94690232-2476a180-0330-11eb-99d7-e174bb762cea.png)
 - [](# "quick-repo-deletion") [Lets you delete your repos in a click, if they have no stars, issues, or PRs.](https://user-images.githubusercontent.com/1402241/99716945-54a80a00-2a6e-11eb-9107-f3517a6ab1bc.gif)
 - [](# "useful-forks") [Helps you find the most active forks](https://user-images.githubusercontent.com/38117856/107463541-542e8500-6b2c-11eb-8b25-082f344c1587.png), via https://useful-forks.github.io.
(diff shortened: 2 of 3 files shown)


kidonng:
Not that I'm in favor of or against this, just want to say the request probably should be cached.

shinenelson:
> the request probably should be cached.

You are right. I made the change hastily to scratch an itch. I did not
think it through enough. I'll commit the change to cache the request in
a bit.

> This means this feature could be confusing, because afaik there would be no difference (visually) between repos with no workflows but Actions still enabled, and repos with Actions disabled.

You are right, but then I am of the opinion that the tab being there for
repositories with no workflows is useless.

For example, look at the tabzilla repository that I linked to in the
description ( it is the repository that annoyed me enough to open this
pull request ). It is an archived repository that has its last commit in
2016. No one is going to disable Actions for an archived repository and
that repository definitely does not have any Actions workflows
configured. What good is that tab being there?
I had been tripped off many times when I had gone to look at the
workflows on some repositories only to be thrown the 'Automate your
workflow from idea to production' page which is annoying. But this
archived repository is what really annoyed me enough to create the pull
request.

We had a similar discussion when the Projects feature was introduced and
enabled globally for all repositories - #309.

shinenelson:
Oops, that was my email client @yakov116. I'll keep it in mind while replying to comments here.

yakov116:
> No one is going to disable Actions for an archived repository and that repository definitely does not have any Actions workflows configured

I don't think anything should run on archived repo's. I would have disabled all API features it a long time ago if it was possible. We have an issue the we cannot know that the repo is archived until the page is loaded.

@fregante can we have your say?

source/features/remove-actions-tab.tsx
```diff
@@ -0,0 +1,37 @@
...
+	}
+
+	actionsTab.remove();
+}
+
+void features.add(__filebasename, {
+	include: [
+		pageDetect.isRepo
+	],
+	awaitDomReady: false,
```
yakov116, line 35:
```suggestion
	exclude: [
		// Repo/Organization owners should see the tab. If they don't need it, they should disable actions altogether
		pageDetect.canUserEditRepo,
		pageDetect.canUserEditOrganization
	],
	awaitDomReady: false,
```

yakov116 (changes requested):
Just like in the projects tab, we do the same.

source/features/remove-actions-tab.tsx
shinenelson: I don't think we need to enable it on organization pages because Actions is a repository-only thing, isn't it?

shinenelson:
> We have an issue the we cannot know that the repo is archived until the page is loaded.

The archive repository was only an example of one tangent of annoyance. I wasn't implying that we should add this feature only for archived repositories. Like the removal of the Projects tab, my point was to remove the tab for any project that does not have any Workflows configured in the repository.

The most common one is, like I described in my previous comment, when you are 'roaming' around projects and you open their Actions to take a 'peek' into their Workflows to understand how they do their CI. And then you get thrown the default 'Automate your workflow from idea to production' landing page. I have been annoyed by that page so much to the point that I now navigate to the workflow definitions and then make a mental model of their CI runs.

source/features/remove-actions-tab.tsx
```diff
@@ -0,0 +1,41 @@
...
+	actionsTab.remove();
+}
+
+void features.add(__filebasename, {
+	include: [
+		pageDetect.isRepo
+	],
+	exclude: [
+		// Repo owners should see the tab. If they don't need it, they should disable actions altogether
+		pageDetect.canUserEditRepo
```
fregante, line 37: You can't disable actions. Please ensure that every line still applies when copy-pasting code.

The exclusion is fine, but the comment is not applicable

source/features/remove-actions-tab.tsx
```diff
@@ -0,0 +1,41 @@
+import cache from 'webext-storage-cache';
+import elementReady from 'element-ready';
+import * as pageDetect from 'github-url-detection';
+
+import features from '.';
+import {getRepo} from '../github-helpers';
+import * as api from '../github-helpers/api';
+
+const getWorkflowsCount = cache.function(async (): Promise<number> => {
+	const {workflows} = await api.v3('actions/workflows');
```
fregante, line 10: Can you use API v4? It's a smaller response

fregante (commented):
Thanks for the PR! Generally it's best to open an issue to discuss it first, and then specify you're willing to open a PR, since some requests are rejected

source/features/remove-actions-tab.tsx
yakov116: https://github.com/yakov116/TestR

image

I suggested that code.

source/features/remove-actions-tab.tsx
shinenelson: That'd make the feature behind the API token requirement. I was hoping that this feature could also benefit users who do not apply API tokens. But I can change it if you feel strongly about it.

yakov116: Use V4

cheap-glitch:
Just an idea: instead of outright removing the tab, the feature could display a counter next to "Actions" (like the "Issues" or "Releases" tabs) showing the number of workflows in the current repo.

A count of `0` would still communicate that no workflows are configured for the repo, while making the feature less "destructive".

kidonng:
> Just an idea: instead of outright removing the tab, the feature could display a counter next to "Actions" (like the "Issues" or "Releases" tabs) showing the number of workflows in the current repo.

Reminds me of `profile-gists-link` and #4246. I don't hate the idea of heading towards Refined Counters™️ but does them introduce too much requests and information?

yakov116:
Can't you see from `ci-link` if they have actions enabled?

source/features/remove-actions-tab.tsx
fregante: TIL

fregante:
Accepted. It makes sense: We also hide projects in the same fashion. I'm just not _psyched_ about yet another API call.

I'm thinking that we should merge this into `remove-projects-tab` and make a single API call for both. Feasible? It could be named `clean-repo-navigation` or something

M1cha:
Well if you remove the button completely you wouldn't be able to go there to create a new workflow through GitHubs UI.
I dunno how many people would actually use the UI for that but it's something to keep in mind.

source/features/remove-actions-tab.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+	actionsTab.remove();
+}
+
+void features.add(__filebasename, {
+	include: [
+		pageDetect.isRepo
+	],
+	exclude: [
+		// Repo owners should see the tab. If they don't need it, they should disable actions altogether
```
fregante, line 36: @M1cha refer to 👆

M1cha: thanks, next time I'll take a closer look at the code :yum:

fregante: No worries! Thanks for commenting

yakov116:
@shinenelson do you plan on continuing?

shinenelson:
> @shinenelson do you plan on continuing? 

I'd like to, but unfortunately, I've gotten busy with other things. I like Fredrico's idea of `hide-unused-features`.

I am _willing_ to put in the work, but I can't find the time for it. It'd be selfish to keep it blocked for me. I'd be happy if someone else wanted to take over. I think I can find some time next week. If no one else claims it by next week, I'll take it.

fregante:
Closing for cleanup. Open a new PR if you decide to pick it up

PR #4398, opened 2021-05-25
Use system mono font stack

florianbuerger:
SF Mono isn't available in Safari (or any other user installed font for
that matter). I've been using this new font[1] stack in my custom css for a
while now on macOS Safari and it works perfectly.

I am not too familiar with the code base so maybe I've overlooked something.

[1] https://qwtel.com/posts/software/the-monospaced-system-ui-css-font-stack/

## Screenshot

Before 

ScreenShot 2021-05-25 at 10 30 15

After

ScreenShot 2021-05-25 at 10 29 29

### DIFF
--- a/source/features/monospace-textareas.css
+++ b/source/features/monospace-textareas.css
@@ -6,7 +6,7 @@
 }
 
 .rgh-monospace-textareas :is(#merge_title_field, #commit-summary-input, textarea) {
-    font-family: ui-monospace, SFMono-Regular, Consolas, 'Liberation Mono', Menlo, Courier, monospace !important;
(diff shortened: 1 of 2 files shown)


fregante:
The _correct_ font stack for us is what GitHub uses, because we want to match its style exactly, unless we plan on _changing_ the style of every pre/code block (which we don't, currently)

I just checked and GitHub still uses `SFMono-Regular,Consolas,Liberation Mono,Menlo,monospace` exactly

florianbuerger:
Oh good point. I know a couple of folks at GitHub, I'll ping them maybe; this is something that can be done soon-ish.

fregante:
Closing  this for now. The second option could be discussed in a new issue if interested, or we just wait

PR #4382, opened 2021-05-21
Extend `separate-draft-pr-button` to `one-click-pr-or-gist`

darkred:
Closes #4366.
Follows https://github.com/sindresorhus/refined-github/issues/4366#issuecomment-846079038

## Test URLs
https://gist.github.com/

PR #4296, opened 2021-04-29
Drop `link-to-file-in-file-history` feature

shinenelson:
The feature appears to have been natively implemented by GitHub now.

Screenshot_20210429_174007

closes #4271 ( because the feature is being dropped altogether )

PR #4399, opened 2021-05-25
Add `hide-empty-split-diff-side` feature

### LINKED ISSUE #4397: Always use single-column diffs if there are no deletions *or* no additions
Screen Shot 2

Sometimes split diffs are useless because the diffs appear on only one column, like in the files of these PRs:

- https://github.com/sindresorhus/refined-github/pull/4296/files?diff=split (every file)
- https://github.com/sindresorhus/refined-github/pull/3637/files?diff=split (every file)
- https://github.com/sindresorhus/refined-github/pull/4382/files?diff=split (only 2 files of 5)

Showing the diff on 2 columns just limits the available space for code so I think that, if easily possible, we should hide one column in the files that don't benefit from split views.

cheap-glitch:
Closes #4397 

## Test URLs

 * [Right side only](https://github.com/sindresorhus/refined-github/pull/4296/files?diff=split)
 * [Left side only](https://github.com/sindresorhus/refined-github/pull/3637/files?diff=split)
 * [Single-side split diffs & regular split diffs](https://github.com/sindresorhus/refined-github/pull/4382/files?diff=split)
 * [Compare page](https://github.com/sindresorhus/refined-github/compare/main...cheap-glitch:add-hide-empty-split-diff-side?diff=split)
 * [Single commit](https://github.com/sindresorhus/refined-github/commit/3b1359ea465ff5c4d3f0f79e2d6781c7ce9a8283?diff=split)

#### New files
_Before:_
image

_After:_
image

#### Modified files
_Before:_
image

_After:_
image

#### Demo screenshot
hide-empty-split-diff-side


source/features/hide-empty-split-diff-side.tsx
```diff
@@ -0,0 +1,23 @@
...
+				diffTable.classList.add('rgh-empty-split-' + side);
+				break;
+			}
+		}
+	}
+}
+
+void features.add(__filebasename, {
+	include: [
+		() => pageDetect.isPRFiles() && select.exists('[value="split"][checked]')
```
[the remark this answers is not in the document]
fregante: > Maybe this could be applied to everywhere split diffs exists, not just PRs: commit pages, compare pages, new PR creation, etc.?

Yes of course

> maybe `pageDetect` could have `isSplitDiffView()`

That's a little specific, it's not something that generally belongs there.

Also:
```suggestion
		pageDetect.isPRFiles
	],
	exclude: [
		() => !select.exists('[value="split"][checked]')
```

Related: https://github.com/sindresorhus/refined-github/issues/4008#issuecomment-812843149

fregante:
Can you include a screenshot of how it looks in files that aren't completely new? Like the readme.md in this PR

cheap-glitch:
> Can you include a screenshot of how it looks in files that aren't completely new? Like the readme.md in this PR

Done. I didn't make an "official" demo screenshot yet because I still need to fix the weird blank space issue.

fregante:
Keep in mind that both line numbers should be shown, like the native version

cheap-glitch:
> Keep in mind that both line numbers should be shown, like the native version

Do they have to be adjoined though? When the left side is hidden it happens naturally, but when the right side is hidden they will be on both side of the code:
image

The only fixes I can think of is to use `display: flex` and `order` (tried it for bit, it creates a whole slew of new problems) or reorder the `<td>` elements with JS :thinking:

fregante:
Ahh you're right. Yeah sorting can't be done in a clean/safe way, so we'll have to keep the lines on the right side in that case. Let's see if they make sense or if people complain.

fregante:
> I didn't make an "official" demo screenshot yet

Yeah that should come at last

yakov116:
@cheap-glitch whats the status of this just waiting for approval?

yakov116 (approved):
Looks very good! Wow!

source/features/hide-empty-split-diff-side.tsx
```diff
@@ -0,0 +1,31 @@
...
+void features.add(__filebasename, {
+	include: [
+		pageDetect.isCommit,
+		pageDetect.isCompare,
+		pageDetect.isPRFiles
+	],
+	exclude: [
+		() => !select.exists([
+			'[value="split"][checked]', // Form in PR
+			'.table-of-contents .selected[href*="diff=split"]' // Link in single commit
```
yakov116, line 27: What is this excluding its not clear from the comments

cheap-glitch: The pages where there is no "Split diff" button selected. Taken from here:
https://github.com/sindresorhus/refined-github/blob/e948429b30fc204e98f1ba5389c561d86d109287/source/features/faster-pr-diff-options.tsx#L28-L29
(with `split` instead of `unified`)

yakov116: What do you this about adding a comment `The following pages there is no "Split diff" button.`?

readme.md
```diff
@@ -258,6 +258,7 @@ Thanks for contributing! 🦋🙌
 - [](# "git-checkout-pr") [Adds copy-pastable git commands to checkout a PR.](https://user-images.githubusercontent.com/16872793/96938908-8e024f80-1499-11eb-8976-0caf95175dd6.png)
 - [](# "jump-to-change-requested-comment") [Adds a link to jump to the latest changed requested comment.](https://user-images.githubusercontent.com/19198931/98718312-418b9f00-23c9-11eb-8da2-dfb616e95eb6.gif)
 - [](# "view-last-pr-deployment") [Adds a link to open the last deployment from the header of a PR.](https://user-images.githubusercontent.com/16872793/111904355-ef185a00-8a1c-11eb-97dd-aa0272547e73.png)
+- [](# "hide-empty-split-diff-side") [Hides one column in the files that don't benefit from split views.](https://user-images.githubusercontent.com/46634000/120284838-faa9bc00-c2bc-11eb-87b2-fce1bcd71f47.png)
```
fregante, line 261: Rather than "hiding", we're just changing view. So it could be "Always uses single-column diffs on files where split diffs aren't useful" or something like that.

Also what's that "View full" button? If it's not native it should not be part of the screenshot.

The name should also reflect this. Suggestions? I don't want to use the word "force" but "force-unified-view" would almost be correct. "no-useless-unified-diff"? Alternatives?

source/features/hide-empty-split-diff-side.tsx
fregante: Rather, it's excluding pages that are _not_ in split diff. It's best to move this into a function at the top level:

```js
function isUnifiedDiff() {
	return exists(unified:checked, etc)
}
```

source/features/hide-empty-split-diff-side.tsx
```diff
@@ -0,0 +1,31 @@
...
+import select from 'select-dom';
+import * as pageDetect from 'github-url-detection';
+
+import features from '.';
+
+function init(): void {
+	for (const diffTable of select.all('.js-diff-table')) {
+		for (const side of ['left', 'right']) {
+			if (!select.exists(`[data-split-side="${side}"]:is(.blob-code-addition, .blob-code-deletion)`, diffTable)) {
+				diffTable.classList.add('rgh-empty-split-' + side);
```
fregante, line 11: Use attributes
```suggestion
				diffTable.setAttribute('data-rgh-always-unified-diff', side);
```

This way both can be selected with just `[data-rgh-always-unified-diff]

readme.md
cheap-glitch: `no-useless-split-diff-view` sounds good to me :+1:

source/features/hide-empty-split-diff-side.tsx
```diff
@@ -0,0 +1,35 @@
...
+	return select.exists([
+		'[value="unified"][checked]', // Form in PR
+		'.table-of-contents .selected[href*="diff=unified"]' // Link in single commit
+	]);
+}
+
+function init(): void {
+	for (const diffTable of select.all('.js-diff-table')) {
+		for (const side of ['left', 'right']) {
+			if (!select.exists(`[data-split-side="${side}"]:is(.blob-code-addition, .blob-code-deletion)`, diffTable)) {
```
fregante, line 17: Instead of this inner loop and condition, what if you select the data-split-side element and access the side? It doesn’t make sense to query them both specifically

cheap-glitch: But if I select all the `<td>` elements bearing `[data-split-side]`, wouldn't I need to loop over them anyway to check every value of the attribute? :thinking:

fregante: You're right! I was thinking this piece of code was doing something else.

danielsamuels:
For me this is hiding diffs which are actually useful. As an example, here's how a recent diff looks with this feature enabled:

chrome_2021-06-17T090052

What do you think changed in this commit? Well it looks like that line was added - but it wasn't. The normal split diff looks like this:

chrome_2021-06-17T090115

You can see there's a subtle difference between the commits which is actually very important (there was a bug!)

cheap-glitch:
@danielsamuels can you provide an URL where this happens? Thanks

fregante:
Strange, can't replicate.

Diff that's correctly unified: https://github.com/sindresorhus/refined-github/commit/f60cb71e0bc2708ebda622d6c684489241fd367a

Reportedly problematic diff: https://github.com/sindresorhus/refined-github/commit/26e778fd85f7db6c3d5bc6f508e23c77f3ed4a20

danielsamuels:
This is on a GitHub Enterprise instance running 2.22.10, perhaps it's an incompatibility with that version?

fregante:
I think I know what's happening. It checks the left column first, it doesn't find the expected element (`[data-split-side="${side}"]:is(.blob-code-addition, .blob-code-deletion)`) and therefore it assumes it's empty.

I'll open an issue and disable this rule via hotfix for now

PR #4421, opened 2021-05-28
Fix console error in `clean-conversation-headers`

yakov116:
`clean-conversation-headers`: Cannot read property 'textContent' of undefined

URL: https://github.com/yakov116/TestR/pull/10

```
TypeError: Cannot read property 'textContent' of undefined
  at initPR (chrome-extension://pplhppbihehddgkdedhpbfpbbbefeihl/build/refined-github.js:4320:103)
  at async MutationObserver.runFeature (chrome-extension://pplhppbihehddgkdedhpbfpbbbefeihl/build/refined-github.js:8781:11)
```

Sometimes it fires too early.

### DIFF
--- a/source/features/clean-conversation-headers.tsx
+++ b/source/features/clean-conversation-headers.tsx
@@ -27,11 +27,12 @@ async function initIssue(): Promise<void> {
 }
 
 async function initPR(): Promise<void> {
-	const byline = await elementReady('.gh-header-meta .flex-auto:not(.rgh-clean-conversation-header)');
-	const author = select('.author', byline);
-	if (!byline || !author) {
+	const author = await elementReady('.gh-header-meta .flex-auto:not(.rgh-clean-conversation-header) .author');
+	if (!author) {
 		return;
 	}
+
+	const byline = author.closest<HTMLElement>('.flex-auto')!;
 
 	byline.classList.add('rgh-clean-conversation-header');
 
--- a/source/features/clean-conversation-headers.tsx
+++ b/source/features/clean-conversation-headers.tsx
@@ -28,13 +28,13 @@ async function initIssue(): Promise<void> {
 
 async function initPR(): Promise<void> {
 	const byline = await elementReady('.gh-header-meta .flex-auto:not(.rgh-clean-conversation-header)');
-	if (!byline) {
+	const author = select('.author', byline);
+	if (!byline || !author) {
 		return;
 	}
 
 	byline.classList.add('rgh-clean-conversation-header');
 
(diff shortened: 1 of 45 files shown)


fregante (dismissed):
Impossible. 

`.author` is inside `byline`, which is awaited in full: https://github.com/sindresorhus/element-ready/pull/30

If it's not there, it's lazy-loaded, and `elementReady` is not the right tool for lazy-loaded elements.

fregante:
The error suggests that this is happening as a mutation rather than on first load, so it's possible that the feature is running but shouldn't (because .author was already removed)

yakov116:
🤦 I did element ready on the wrong one.

fregante:
It covers all children. It resolves when it finds a sibling of the element you’re querying, which appears only after its end tag.

yakov116:
> so it's possible that the feature is running but shouldn't (because .author was already removed)

Bingo! Its running 2 times. (I cannot figure out when, but it very rare). Should I add a debounce?

fregante:
My guess is that the "header update" is being triggered again, but the byline hasn't changed.

The easy solution is to use `.author` as a way to deduplicate the run instead of what's there now, because it isn't doing its job well enough.

fregante:
I mean use element ready on the author itself and then get the byline after. The current code is just checking for duplication in 2 places.

source/features/clean-conversation-headers.tsx
```diff
@@ -27,14 +27,15 @@ async function initIssue(): Promise<void> {
...
 
 async function initPR(): Promise<void> {
-	const byline = await elementReady('.gh-header-meta .flex-auto:not(.rgh-clean-conversation-header)');
-	if (!byline) {
+	const author = await elementReady('.gh-header-meta .flex-auto:not(.rgh-clean-conversation-header) .author');
+	if (!author) {
 		return;
 	}
 
+	const byline = author.closest<HTMLElement>('.flex-auto')!;
```
fregante, line 35: Didn't we fix this at some point? Or did it never work?
```suggestion
	const byline = author.closest('.flex-auto')!;
```

Anyway the code LGTM

yakov116: The default is `Element` so unless I do `div.flex-auto`... I dont want to do that.

fregante: https://github.com/sindresorhus/refined-github/issues/4008#issuecomment-823814526

yakov116: https://github.com/g-plane/typed-query-selector/issues/16

Issue #4424, opened 2021-05-29
Disappear layout if "repository details" - "Website" contains string "releases".


VergeDX:
At any repo of Github, we can give the repository some details, they will show at the right sidebar.
If we fill in the "Website" section and the link contains the string "release", the details will be missing like this:
![image](https://user-images.githubusercontent.com/25173827/120056510-4535f900-c06f-11eb-9acd-28cbd79e4637.png)

<b>If I turn off the refined extension, it looks like:</b>
![image](https://user-images.githubusercontent.com/25173827/120056527-6dbdf300-c06f-11eb-921b-e9ec9ade98a6.png)

Example repo: [Vanilla-s-Lab/docper](https://github.com/Vanilla-s-Lab/docper), please noticed the right sidebar.
Browser & System: 91.0.4472.77 + macOS 11.4 (20F71)

VergeDX:
After debugging, I confirm this is a bug of clean-repo-sidebar.
Maybe I can find out why.

VergeDX:
At file [source/features/clean-repo-sidebar.tsx#L23](https://github.com/sindresorhus/refined-github/blob/main/source/features/clean-repo-sidebar.tsx#L23), it grab element which is a tag, href end with "/releases",
but if repo detail - website end with "/releases", it will select to wrong element (Website a tag is above Release a tag).

VergeDX:
The current CSS selector of "Release" is ".BorderGrid-cell a[href$="/releases"]",
to avoid choose the "Website", simply add a ":not([role]" can solve this issues.
If the change is appropriate, I can make a PR to finish it. Waiting for reply.

fregante:
Sounds reasonable. If there isn’t any better selector, make sure to include a comment to explain what the “not” is for

VergeDX:
For example, the right element looks like:
``````html
<a href="/sindresorhus/refined-github/releases" data-view-component="true" class="Link--primary no-underline">
    Releases
      <span title="246" data-view-component="true" class="Counter">246</span>
</a>
``````
But if there is a bug, it will select:
``````html
<a title="https://github.com/Vanilla-s-Lab/docper/releases" role="link" target="_blank" class="text-bold" rel="noopener noreferrer" href="https://github.com/Vanilla-s-Lab/docper/releases">github.com/vanilla-s-lab/docper/releases</a>
``````

A simple difference is  `role` attribute, so just find a link without `role` can solve this.
I will make a PR later : )

PR #4428, opened 2021-05-30
`clean-repo-sidebar` fix: Section disappears if URL includes “releases”

VergeDX:
From #4424

### DIFF
--- a/source/features/clean-repo-sidebar.tsx
+++ b/source/features/clean-repo-sidebar.tsx
@@ -20,7 +20,7 @@ async function cleanLicenseText(): Promise<void> {
 }
 
 async function cleanReleases(): Promise<void> {
-	const sidebarReleases = await elementReady('.BorderGrid-cell a[href$="/releases"]', {waitForChildren: false});
+	const sidebarReleases = await elementReady('.BorderGrid-cell h2 a[href$="/releases"]', {waitForChildren: false});
 	if (!sidebarReleases) {
 		return;
 	}


yakov116:
@VergeDX Thanks for the PR.

Can you try not to include issue numbers in commits as creates unnecessary noise

image

source/features/clean-repo-sidebar.tsx
```diff
@@ -20,7 +20,10 @@ async function cleanLicenseText(): Promise<void> {
 }
 
 async function cleanReleases(): Promise<void> {
-	const sidebarReleases = await elementReady('.BorderGrid-cell a[href$="/releases"]', {waitForChildren: false});
+	// Intent: select "Releases" section in right sidebar.
+	// But if "repository details" - "Website" has a link end with "/releases", it will be wrongly selected.
+	// @see refined-github/issues/4424 for more details.
+	const sidebarReleases = await elementReady('.BorderGrid-cell a[href$="/releases"]:not([role])', {waitForChildren: false});
```
yakov116, line 26:
```suggestion
	const sidebarReleases = await elementReady('.BorderGrid-cell a[href$="/releases"]:not([role])', {waitForChildren: false}); // :not('[role]') excludes website links that end with "/releases" #4424
```

I did not test the code, however I think the above comment is enough.

kidonng, line 26: Perhaps

```suggestion
	const sidebarReleases = await elementReady('.BorderGrid-cell h2 a[href$="/releases"]', {waitForChildren: false});
```

fregante: Yeah this would work best  since we’re specifically looking for a header link

VergeDX:
Changes performed, no need any comment...?

fregante:
謝謝～

PR #4431, opened 2021-05-30
Fix: Extension may fail to load if custom CSS is specified

### LINKED ISSUE #4376: Extension sometimes fails to run when custom CSS is injected
I've recently added some custom CSS to hide the "Explore repositories" sidebar on the dashboard. This makes the extension fail regularly with the following error (Firefox 88.0.1, RGH 21.5.19):
```text
TypeError: document.head is null refined-github.js:1913:30
```

The line in question:
https://github.com/sindresorhus/refined-github/blob/83f233677157d21fb6ae11958fa45aa7ee99d65a/source/features/index.tsx#L91

Since this only happens (afaik) when opening new tabs (regardless of the URL), I'm guessing the extension tries to inject the CSS before the `<head>` exists and thus execution stops. This was probably introduced by #4352.

Edit: can confirm the bug goes away when removing the custom CSS.

fregante:
Fixes #4376

This is hard to test since it's a rare occurrence.

1. Specify CSS in the options
2. Reload the page a million times

### DIFF
--- a/source/features/index.tsx
+++ b/source/features/index.tsx
@@ -1,4 +1,3 @@
-import delay from 'delay';
 import React from 'dom-chef';
 import select from 'select-dom';
 import domLoaded from 'dom-loaded';
@@ -6,6 +5,7 @@ import stripIndent from 'strip-indent';
 import {Promisable} from 'type-fest';
 import * as pageDetect from 'github-url-detection';
 
+import waitFor from '../helpers/wait-for';
 import onNewComments from '../github-events/on-new-comments';
 import bisectFeatures from '../helpers/bisect';
 import optionsStorage, {RGHOptions} from '../options-storage';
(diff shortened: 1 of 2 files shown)


fregante:
I looked into contentScript.register, but my polyfill doesn't support `code` yet: https://github.com/fregante/content-scripts-register-polyfill/issues/16

jdreesen:
For me, the notification page is broken most of the time, every other page seems to work 99% of the time.

fregante:
Is it due to the custom CSS? Can you drop the CSS and ensure it doesn’t happen anymore?

jdreesen:
Yes, it is. If I remove the custom CSS, RGH seems to work reliably on the notifications page.

jdreesen:
Just installed the new version, and it's working again. Thank you so much!

fregante:
Great! Thank you for the update

PR #4433, opened 2021-05-31
`embed-gist-inline` fix: Exclude gist user links on Enterprise

fregante:
Tested on non-GHE, still works 

### Embedded gist

https://gist.github.com/fregante/cf4d0415c4802e0c81812ae9bbb821d8

### Big file

https://gist.github.com/fregante/afa017a7c81dd5f5eeefc963593088b4

### User link

https://gist.github.com/fregante

## Screenshot

Screen Shot

### DIFF
--- a/source/features/embed-gist-inline.tsx
+++ b/source/features/embed-gist-inline.tsx
@@ -4,13 +4,26 @@ import select from 'select-dom';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
+import {getCleanPathname} from '../github-helpers';
 
-const isGist = (link: HTMLAnchorElement): boolean =>
-	!link.pathname.includes('.') && // Exclude links to embed files
-	(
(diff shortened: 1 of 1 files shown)


source/features/embed-gist-inline.tsx
```diff
@@ -4,13 +4,26 @@ import select from 'select-dom';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
+import {getCleanPathname} from '../github-helpers';
 
-const isGist = (link: HTMLAnchorElement): boolean =>
-	!link.pathname.includes('.') && // Exclude links to embed files
-	(
-		(link.hostname.startsWith('gist.') && link.pathname.includes('/', 1)) || // Exclude user links
-		link.pathname.startsWith('gist/')
```
fregante, line 12: We were only excluding user URLs from gist.GitHub.com, but not from GHE

PR #4435, opened 2021-05-31
Fix `swap-branches-on-compare` is not operated in direct comparison

### LINKED ISSUE #4390: `swap-branches-on-compare` should not be run on 2 dot notation
URL: https://github.com/sindresorhus/modern-normalize/compare/af8df39ff9abd290cb5ee22b7d1cc3451367f8fe..f9a9045502138c23478c2d4760dfa28e01d53e7b

```js
TypeError: Cannot read property 'parentElement' of undefined
  at init (chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/build/refined-github.js:5848:14)
  at runFeature (chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/build/refined-github.js:1924:24)
  at setupPageLoad (chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/build/refined-github.js:1937:40)
  at chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/build/refined-github.js:1958:35
```

KSH-code:
Closes #4390

The double dotted delimiter is used with direct comparison.

So need to check if path includes how many there is dot.

And then reuse the delimiter at replacement and also find element by `.octicon-arrow-both`. because, in direct comparison, arrow is both directional.

## Test URLs
- https://github.com/sindresorhus/modern-normalize/compare/af8df39ff9abd290cb5ee22b7d1cc3451367f8fe..f9a9045502138c23478c2d4760dfa28e01d53e7b

### DIFF
--- a/source/features/swap-branches-on-compare.tsx
+++ b/source/features/swap-branches-on-compare.tsx
@@ -7,10 +7,13 @@ import features from '.';
 import {buildRepoURL, getRepo} from '../github-helpers';
 
 function init(): void {
-	const references = getRepo()!
+	const referencePath = getRepo()!
 		.path
-		.replace('compare/', '')
-		.split('...')
+		.replace('compare/', '');
+	const isDirectComparision: boolean = referencePath.includes('..');
+	const branchDelimiter: string = isDirectComparision ? '..' : '...';
+	const references: string[] = referencePath
+		.split(branchDelimiter)
 		.reverse();
 
 	// Compares against the "base" branch if the URL only has one reference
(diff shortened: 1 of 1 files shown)


fregante:
If it shouldn’t run, then the only need a function in the `exclude` array or a new specific 3-dot detection in `github-url-detection`. 

The init function should likely not be affected.

KSH-code:
@fregante 
ooops, title is invalid I'll edit.
this is just fixed about not operation with direction comparison.

fregante:
I think 2-dot comparisons can’t be switched, they show the same thing regardless of order. That’s why the feature should not run at all.

KSH-code:
I completely understand and agree with your opinion.
I'm going to close this. then I rework at github-url-detection.

PR #4441, opened 2021-06-03
Enable `file-finder-buffer` everywhere

### LINKED ISSUE #4159: `file-finder-buffer` should work on the same pages as `repo-wide-file-finder`
Would be nice to have the `file-finder-buffer` feature apply when trying to launch it from another repo page.

Not sure if `change-request` or `enhancement`.

cheap-glitch:
Fixes #4159 

## Test URLs
Any repo page, including this one.

## Screenshot
https://user-images.githubusercontent.com/46634000/120622625-45fbd000-c45f-11eb-9d7f-1625257620e9.mp4


source/features/file-finder-buffer.tsx
```diff
@@ -35,7 +35,7 @@ function pjaxStartHandler(event: CustomEvent): void {
 		<span className="flex-self-stretch mr-2">{bufferField}</span>
 	);
 	bufferField.focus();
-	for (const element of select.all('.pagehead-actions, .rgh-ci-link, .octotree-bookmark-btn')) {
+	for (const element of select.all('.pagehead-actions, .rgh-ci-link details, .octotree-bookmark-btn')) {
```
cheap-glitch, line 38: Since `.rgh-ci-link` is applied to the wrapper around the title, this was effectively removing the whole thing (including the inserted buffer).

source/features/repo-wide-file-finder.tsx
```diff
@@ -11,7 +11,7 @@ async function init(): Promise<void> {
 		<a
 			hidden
 			data-hotkey="t"
-			data-pjax="true"
+			data-pjax="#js-repo-pjax-container"
```
cheap-glitch, line 14: Setting the `data-pjax` attribute to `true` caused the hotkey to trigger a "real" navigation instead of an AJAX one, so the handlers added by `file-finder-buffer` weren't called.

This is probably because the element is inserted outside the main "container" (as opposed of the native hotkey which is located on the "Go to file" button).

Fixed it by changing the value of the attribute, mimicking the native link in the repo name heading:
```html
<a data-pjax="#js-repo-pjax-container" href="/sindresorhus/refined-github">refined-github</a>
```

But this could also probably be fixed by inserting the link element directly inside the container.

PR #4446, opened 2021-06-05
`repo-wide-file-finder`: Use current branch/tag when possible

mrdziuban:
The repo wide file finder was always loading the finder for the repo's default branch, even if a different branch/tag was selected. This updates the logic to try to get the current branch from the URL before falling back to the default.

## Test URLs

https://github.com/optics-dev/Monocle/tree/v3.0.0-RC2
https://github.com/optics-dev/monocle/commits/v3.0.0-RC2

### Before

Note `master` in the find URL

<details>
  <summary>expand GIF</summary>
  
  before

</details>

### After

Note `v3.0.0-RC2`, the selected tag, in the find URL

<details>
  <summary>expand GIF</summary>
  
  after

</details>

### DIFF
--- a/source/features/repo-wide-file-finder.tsx
+++ b/source/features/repo-wide-file-finder.tsx
@@ -3,8 +3,8 @@ import select from 'select-dom';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
-import {buildRepoURL, getCurrentCommittish} from '../github-helpers';
 import getDefaultBranch from '../github-helpers/get-default-branch';
+import {buildRepoURL, getCurrentCommittish} from '../github-helpers';
 
 async function init(): Promise<void> {
 	document.body.append(
--- a/source/features/repo-wide-file-finder.tsx
+++ b/source/features/repo-wide-file-finder.tsx
@@ -3,7 +3,7 @@ import select from 'select-dom';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
-import {buildRepoURL} from '../github-helpers';
+import {buildRepoURL, getCurrentCommittish} from '../github-helpers';
 import getDefaultBranch from '../github-helpers/get-default-branch';
 
 async function init(): Promise<void> {
@@ -12,7 +12,7 @@ async function init(): Promise<void> {
 			hidden
 			data-hotkey="t"
 			data-pjax="true"
-			href={buildRepoURL('find', await getDefaultBranch())}
+			href={buildRepoURL('find', await (getCurrentCommittish() ?? getDefaultBranch()))}
 		/>
 	);
 }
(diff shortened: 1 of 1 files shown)


source/features/repo-wide-file-finder.tsx
```diff
@@ -5,14 +5,23 @@ import * as pageDetect from 'github-url-detection';
...
 import getDefaultBranch from '../github-helpers/get-default-branch';
+import GitHubURL from '../github-helpers/github-url';
+
+async function getBranchForFind(): Promise<string> {
+	try {
+		return (new GitHubURL(location.href)).branch;
+	} catch {
+		return getDefaultBranch();
+	}
+}
```
fregante, line 16: I don’t think we need this, there’s already a “get current branch” function somewhere

fregante (commented):
Thank you for the PR! That’s a good point. 

Keep in mind that this feature only extends the native one, which means it does not run on your first example link. Can you add more example links where this change would make a difference? I think they would be file link, a file blame link, etc

mrdziuban:
@fregante I thought that was the case too but as the GIFs I posted demonstrate, this change does affect the first URL. Is there a possibility this feature is running on pages it shouldn’t?

fregante:
Oh you may be right. From your gifs I can see that the tag page does not have the “Find files” button, so it’s possible that GitHub only adds that shortcut on the main branch. What a weird limitation.

source/features/repo-wide-file-finder.tsx
yakov116: getCurrentCommitsh

mrdziuban: Thanks! Updated to use `getCurrentCommitish`

source/features/repo-wide-file-finder.tsx
```diff
@@ -12,7 +12,7 @@ async function init(): Promise<void> {
 			hidden
 			data-hotkey="t"
 			data-pjax="true"
-			href={buildRepoURL('find', await getDefaultBranch())}
+			href={buildRepoURL('find', await (getCurrentCommittish() ?? getDefaultBranch()))}
```
yakov116, line 15:
```suggestion
			href={buildRepoURL('find', getCurrentCommittish() ?? await getDefaultBranch())}
```

source/features/repo-wide-file-finder.tsx
```diff
@@ -3,7 +3,7 @@ import select from 'select-dom';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
-import {buildRepoURL} from '../github-helpers';
+import {buildRepoURL, getCurrentCommittish} from '../github-helpers';
 import getDefaultBranch from '../github-helpers/get-default-branch';
```
yakov116, line 7:
```suggestion
import getDefaultBranch from '../github-helpers/get-default-branch';
import {buildRepoURL, getCurrentCommittish} from '../github-helpers';
```

yakov116 (approved):
Untested but looks good to me

Issue #4266, opened 2021-04-22
fit-textarea causing significant slow down when adding PR review comments


gmaclennan:
Browser: Safari 14.0.3
Example URL: https://github.com/digidem/mapeo-mobile/pull/566/files#

When reviewing a PR and adding a code comment with multiple lines, refined github causes a slowdown that results in a typing appearing on the screen a second or so after keys are pressed.

I ran the profiler on the page and identified that each keypress is calling the `fitTextArea()` function and takes ~5ms.

Turning off the fit-textarea feature resolves the problem and typing is not delayed any more.

yakov116:
Will be fixed in the next release

fregante:
Confirmed and hopefully it will be fixed in the next release. I don’t know when that’ll happen in Safari because releases aren’t automated there yet #3686

fregante:
For the time being, yes, disabling the feature is highly advised. I had to do the same for the past few months.

fregante:
We had to re-enable it due to https://github.com/sindresorhus/refined-github/issues/4275 unfortunately

fregante:
Closing because of  #4447

I'd basically need to rewrite or replace `fit-textareas` to enable the feature everywhere:

- https://github.com/fregante/fit-textarea/issues/21

PR #4447, opened 2021-06-06
Disable `fit-textarea` in Safari

fregante:
[fit-textarea](https://github.com/fregante/fit-textarea) v2 is slow on Safari (https://github.com/sindresorhus/refined-github/issues/4266) and v3 is slow on Chrome (https://github.com/sindresorhus/refined-github/issues/4275)

To re-enable v3, I need to figure out why it's so slow on Chrome

### DIFF
--- a/source/features/fit-textareas.tsx
+++ b/source/features/fit-textareas.tsx
@@ -3,6 +3,7 @@ import select from 'select-dom';
 import delegate from 'delegate-it';
 import fitTextarea from 'fit-textarea';
 import * as pageDetect from 'github-url-detection';
+import isSafari from '../helpers/browser-detection';
 
 import features from '.';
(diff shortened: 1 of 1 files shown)


source/features/fit-textareas.tsx
```diff
@@ -3,6 +3,7 @@ import select from 'select-dom';
 import delegate from 'delegate-it';
 import fitTextarea from 'fit-textarea';
 import * as pageDetect from 'github-url-detection';
+import isSafari from '../helpers/browser-detection';
```
yakov116, line 6:
```suggestion
import isSafari from '../helpers/browser-detection';
import delegate from 'delegate-it';
import fitTextarea from 'fit-textarea';
import * as pageDetect from 'github-url-detection';
```

yakov116 (approved):

PR #4448, opened 2021-06-08
Add `remove-unused-repo-tabs` feature

yakov116:
Replaces and closes #4395

Test on
https://github.com/sindresorhus/eslint-plugin-unicorn 
https://github.com/pixiebrix/pixiebrix-extension
https://github.com/twschiller/dslfin-www/projects
https://github.com/twschiller/dslfin-www

image
image

WIP


yakov116:
@fregante I did some research it looks like the best thing we can do here is display the counter always.

For example https://github.com/twschiller/dslfin-www has wiki enabled but only collaborators can edit. Any non collaborator gets redirected to the main repo page.

Hiding in such cases would not be good.

Can we change this feature to be `add-repo-tab-counters`?

fregante:
> only collaborators can edit

The applies to every other tab we hide: projects, actions, wikis all can be edited by collaborators. 

For projects we added an item to the "+" dropdown; Should we do the same for the other 2?

yakov116:
> For projects we added an item to the "+" dropdown; Should we do the same for the other 2?

I am 👎 on that, we are not "adding an action or wiki".

Should we make an API call to see if they are collaborators and share that across the features?

yakov116:
What if we moved it to the "more drop down" when there are no actions/wiki?

fregante:
> What if we moved it to the "more drop down" when there are no actions/wiki?

Not a bad idea. If it exists, append it there. If it doesn't, 🤷‍♂️.

yakov116:
> > What if we moved it to the "more drop down" when there are no actions/wiki?
> 
> Not a bad idea. If it exists, append it there. If it doesn't, 🤷‍♂️.

@fregante what if they have `more-dropdown` disabled?

shinenelson:
I came back to complete #4395 and this is already done \o/

shinenelson (commented):
Should we, perhaps, consider renaming the `remove-projects-tab` to `add-projects-to-new-drop-down`?

Also, +1 to adding Actions to the `+` drop-down. Then we can re-purpose the existing `remove-actions-tab` to what it is currently doing `add-features-to-new-dropdown`.

yakov116:
@fregante ?

fregante:
> > If it doesn't, 🤷‍♂️.
> 
> @fregante what if they have `more-dropdown` disabled?

🤷‍♂️ => I don't care about them

fregante:
> Should we, perhaps, consider renaming the `remove-projects-tab` to `add-projects-to-new-drop-down`?

That feature doesn't make sense anymore, it should be part of this one. The only reason to add a link there is because the tab has been removed.

yakov116:
> That feature doesn't make sense anymore, it should be part of this one. The only reason to add a link there is because the tab has been removed.

Sorry I was out for a few days, I want to finish this up.

I have a few questions.

1. Are we keeping the "new project dropdown"?
2. If yes are we adding new actions or new wiki to it.

From what I understand we are not going to "remove" the tabs we are just going to move it to the `more-dropdown`

source/features/remove-unused-repo-tabs.tsx
```diff
@@ -10,12 +10,16 @@ import looseParseInt from '../helpers/loose-parse-int';
...
+import {onlyShowInDropdown} from './more-dropdown';
 import {buildRepoURL, getRepo} from '../github-helpers';
 
-function tabCannotBeRemoved(tab: HTMLElement | undefined): boolean {
+async function tabCannotBeHidden(tab: HTMLElement | undefined): Promise<boolean> {
 	if (
 		!tab || // Tab disabled 🎉
-		tab.matches('.selected') // User is on tab 👀
+		tab.matches('.selected') ||// User is on tab 👀
+		// Repo/Organization owners should see the tab. If they don't need it, they should disable the tab altogether
```
shinenelson, line 20:
```suggestion
		// Repo/Organization owners should see the tab. If they don't need it, they should disable the feature altogether
```

yakov116:
@fregante can you give this a basic review? (As in  what this feature should include and how it should be split)

Also what should the screenshot be of?

fregante:
Closing for cleanup. Open a new PR if you decide to pick it up

yakov116:
I am waiting for your feedback...

yakov116:
its fine I will open a new PR with all my questions

PR #4453, opened 2021-06-09
Disable `swap-branches-on-compare` in direct comparison

### LINKED ISSUE #4390: `swap-branches-on-compare` should not be run on 2 dot notation

KSH-code:
Closes #4390

In direct comparison, `swap-branches-on-compare` feature is not needed because of the same result after swap.

To fix this bug, `swap-branches-on-compare` should be disabled.
So check if only 2-dot exists.

## Test URLs
- https://github.com/sindresorhus/modern-normalize/compare/af8df39ff9abd290cb5ee22b7d1cc3451367f8fe..f9a9045502138c23478c2d4760dfa28e01d53e7b

### DIFF
--- a/source/features/swap-branches-on-compare.tsx
+++ b/source/features/swap-branches-on-compare.tsx
@@ -28,7 +28,7 @@ void features.add(__filebasename, {
 		pageDetect.isCompare
 	],
 	exclude: [
-		() => select.exists('.range-editor .octicon-arrow-both')
+		() => location.pathname.includes('..') && !location.pathname.includes('...')
 	],
 	init
 });
--- a/source/features/swap-branches-on-compare.tsx
+++ b/source/features/swap-branches-on-compare.tsx
@@ -28,7 +28,7 @@ void features.add(__filebasename, {
 		pageDetect.isCompare
 	],
 	exclude: [
-		() => location.pathname.includes('..') && !location.pathname.includes('...')
+		() => /\.\.+/.exec(location.pathname)?.[0]!.length === 2
 	],
 	init
 });
--- a/source/features/swap-branches-on-compare.tsx
+++ b/source/features/swap-branches-on-compare.tsx
@@ -27,5 +27,8 @@ void features.add(__filebasename, {
 	include: [
 		pageDetect.isCompare
 	],
+	exclude: [
+		() => select.exists('.range-editor .octicon-arrow-both')
+	],
 	init
 });


yakov116:
It has to work on the following links

https://github.com/sindresorhus/refined-github/compare/master...no-releases
https://github.com/sindresorhus/refined-github/compare/no-releases...master
https://github.com/sindresorhus/refined-github/compare/no-releases
https://github.com/sindresorhus/refined-github/compare/master...busches:fix-console-errors
https://github.com/busches/refined-github/compare/fix-console-errors...sindresorhus:master

Notice https://github.com/sindresorhus/refined-github/compare/no-releases does not have '...'

KSH-code:
@yakov116 
Thanks! 
to exclude only double dots, I uses `#split` and `===` to its length.
Since I expected that only excluding path has double dot.

yakov116:
Nice but...

What about

https://github.com/sindresorhus/refined-github/compare/h.l.o.o

KSH-code:
Good to point out. I had known that branch name cannot include `.`.
I'll fix then change status of PR

source/features/swap-branches-on-compare.tsx
```diff
@@ -27,5 +27,8 @@ void features.add(__filebasename, {
 	include: [
 		pageDetect.isCompare
 	],
+	exclude: [
+		() => select.exists('.range-editor .octicon-arrow-both')
```
yakov116, line 31: I dont like this since if the selector changes we will never know that it broke.

@fregante ?

KSH-code: I agree with that.
By the way, the init function uses the same way.
so I think this file assumes icon always exists.

fregante:
Why are you not using the suggested solution? https://github.com/sindresorhus/refined-github/issues/4390#issuecomment-853775398

Also force pushed and mentioning the original issue **in the commit** are noisy and not advised, they’re noisy

source/features/swap-branches-on-compare.tsx
fregante: @yakov116 we use this pattern in a lot of features. Here the only reason not to use it is that DOM checks are not needed.

KSH-code:
@fregante 
I completely understand.
Finished push new commit you suggested without force option

source/features/swap-branches-on-compare.tsx
```diff
@@ -27,5 +27,8 @@ void features.add(__filebasename, {
 	include: [
 		pageDetect.isCompare
 	],
+	exclude: [
+		() => location.pathname.includes('..') && !location.pathname.includes('...')
```
fregante, line 31: Maybe? This makes it slightly more exact because we're specifically excluding 2 dots. That's all. This avoids the awkward `exclude: negative includes`

```suggestion
		() => /\.\.+/.exec(location.pathname)?.[0]!.length === 2
```

KSH-code: The reason why I did not use regular expressions is because `includes` method represents intention 
 well. If using regular expression represents intention as well, I agree with your opinion!
Thanks for suggestion.

KSH-code: before I change, change PR status to draft.

yakov116: If the suggestion works then its fine. Did you test it?

KSH-code: @yakov116 @fregante 
I've just tested, but it was not worked

KSH-code: I've just tested. it was worked

fregante (approved):

yakov116:
title good?

KSH-code:
@yakov116 did you mean PR title?
Disable `swap-branches-on-compare` in direct comparison seems better to me.

PR #4475, opened 2021-06-14
Fix `prevent-link-loss` warning message copy

waldyrious:
## Test URLs

Try adding this URL to a new comment (e.g. in this PR):

    https://github.com/github/choosealicense.com/compare/gh-pages...Pukimaa:gh-pages#diff-c0311c1df17d5d714db46ca1401ad374a35de861a461d211c6c947f1b2194034R8

## Screenshot

Screenshot 2021-06-14 at 18 16 38

After

image

### DIFF
--- a/source/features/prevent-link-loss.tsx
+++ b/source/features/prevent-link-loss.tsx
@@ -28,7 +28,7 @@ function handleButtonClick({delegateTarget: fixButton}: delegate.Event<MouseEven
 function getUI(field: HTMLTextAreaElement): HTMLElement {
 	return select('.rgh-prevent-link-loss-container', field.form!) ?? (
 		<div className="flash flash-warn mb-2 rgh-prevent-link-loss-container">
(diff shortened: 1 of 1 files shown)


yakov116:
@waldyrious can you post a screenshot of it working?

yakov116 (approved):

waldyrious:
I didn't think that would be needed for a 1-character change 😅 but sure — can you show me how to do it? Would I need to build the extension locally and install it? Or is it a simpler process?

yakov116:
You could download the zip from the actions artifacts and test. If you can't I can try later.

waldyrious:
If you could try it, I'd appreciate it. I'm not entirely clear on how find the zip, and I'm afraid I wouldn't know what to do with it anyway in order to test the extension.

yakov116:
https://github.com/sindresorhus/refined-github/suites/2991340755/artifacts/67565822

PR #645, opened 2017-07-26
Improve avatar style during loading and for transparent avatars

bfred-it:
# Bug

Transparent avatars:

> hole-y avatars  <-- before this PR

As seen on this comment: https://github.com/sindresorhus/refined-github/pull/571#issuecomment-313707428

# Enhancement 

Avatars' gray placeholders appear before they load:

> 

Testable on https://github.com/babel/babel/pull/3646

PR #4476, opened 2021-06-14
Fix very small images in `reaction-avatars`

FloEdelmann:
There are users with a very small profile image:

grafik

GitHub scales up profile images to fill avatar circles in comments though:

grafik

So these two avatar circles look like two different people. Changing `max-width` to `width` in `reaction-avatars` fixes it:

grafik

## Test URLs

- user with a very small profile image: `https://github.com/Strubbl`
- that user's profile image: `https://avatars.githubusercontent.com/u/97055?v=4`
- comment with that user's reaction: `https://github.com/streetcomplete/StreetComplete/issues/2909#issuecomment-860959344`
- comment by that user: `https://github.com/streetcomplete/StreetComplete/issues/1176#issuecomment-860973960`

### DIFF
--- a/source/features/reactions-avatars.css
+++ b/source/features/reactions-avatars.css
@@ -39,6 +39,6 @@ button.reaction-summary-item { /* `button` excludes the "Add reaction" icon */
 
 /* This image will start at height:0 and will expand once loaded, covering the gray placeholder */
 .reaction-summary-item a img { /* `a` required for #3237 */
-	max-width: 100%;
+	width: 100%;
 	background-color: var(--color-bg-info, #fff);
 }


fregante:
Does this change the loading behavior though? Check the comment and the PR that included that code. Normally the avatars appear gray and then the image fills up. I think it you drop `max-` it might stop that.

FloEdelmann:
The CSS rule was introduced in #645, already with `max-width`. I just tried this change in the test URL linked there (https://github.com/babel/babel/pull/3646) with simulated 3G speed and disabled cache, and the gray circles appear before the images are loaded. So it works the same.

fregante:
Thank you!

PR #3573, opened 2020-09-18
Restore `hide-zero-packages`

kidonng:
Thanks for contributing! 🍄

1. LINKED ISSUES:
   Close #3572

   > [Just to mention that if we want to hide “Packages” tab on organization, we need to query the API.](https://github.com/sindresorhus/refined-github/pull/3297#issuecomment-650671415)

2. TEST URLS:
   https://github.com/sindresorhus/refined-github (no packages)
   https://github.com/yarnpkg/berry (with packages)
   https://github.com/sindresorhus (no packages)
   https://github.com/kidonng (with packages)

3. SCREENSHOT:

PR #4481, opened 2021-06-16
Make `hide-zero-packages` more reliable

### LINKED ISSUE #4472: `hide-zero-packages` never runs
Test on https://github.com/sindresorhus

This is because the detection is not dom ready safe.
https://github.com/sindresorhus/refined-github/blob/6f2b3731cfcbc4af1915dac0cb35526594df141f/source/features/hide-zero-packages.tsx#L17-L24

yakov116:
Close #4472 

## Test URLs
   https://github.com/sindresorhus (no packages)
I cannot find any user with packages. Anyone know of a test URL?

## Screenshot
image

### DIFF
--- a/source/features/hide-zero-packages.tsx
+++ b/source/features/hide-zero-packages.tsx
@@ -1,11 +1,11 @@
-import elementReady from 'element-ready';
+import select from 'select-dom';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
 import getTabCount from './remove-projects-tab';
 
 async function init(): Promise<void | false> {
-	const packagesTab = await elementReady('.UnderlineNav-item[href$="?tab=packages"]:not(.selected)');
+	const packagesTab = select('.UnderlineNav-item[href$="?tab=packages"]:not(.selected)');
 	if (!packagesTab || await getTabCount(packagesTab) > 0) {
 		return false;
 	}
@@ -21,6 +21,5 @@ void features.add(__filebasename, {
 		// Keep it visible on your own profile due to #3737
 		pageDetect.isOwnUserProfile
 	],
-	awaitDomReady: false,
 	init
 });
--- a/source/features/hide-zero-packages.tsx
+++ b/source/features/hide-zero-packages.tsx
@@ -10,7 +10,7 @@ async function init(): Promise<void | false> {
 		return false;
 	}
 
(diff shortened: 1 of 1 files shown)


fregante:
> I cannot find any user with packages. Anyone know of a test URL?

I've probably never seen anyone using them. Maybe some GitHub developers? A twitter search might get you some results

yakov116:
> I've probably never seen anyone using them. Maybe some GitHub developers? A twitter search might get you some results

No luck checked 235 accounts! @kidonng how did you test?

kidonng:
I used to have published some packages on GitHub, but I've removed them for, well, obvious reasons.
I guess the situation is that, GitHub packages is more likely to be used a. in private or b. under organization accounts. Perhaps the only way to test is give some fake data 😅

fregante:
Here's a package https://github.com/primer/components/packages/10922

yakov116:
@fregante those are on a repo, which is not handled by this feature.

I need on a user.

fregante:
The first is on an org: https://github.com/orgs/primer/packages
Still not applicable?

yakov116:
Nope

```
include: [
		pageDetect.isUserProfile
	],
	exclude: [
		// Keep it visible on your own profile due to #3737
		pageDetect.isOwnUserProfile
	],
```

fregante:
Why not? It also has a package tab

fregante:
Anyway, a Google search turned up https://github.com/srepollock?tab=packages

yakov116:
And... The only way is a query as the count is not displayed.

Drop Feature?

yakov116:
> Why not? It also has a package tab

I think because the count was not displayed. I can blame it

fregante:
The screenshot in your post was taken from the readme. That makes zero sense

yakov116:
FYI it was not I took it from https://github.com/sindresorhus/refined-github/pull/3573

Besides
> No luck checked 235 accounts!

The point is to know it is working, I could have made a new one. Whatever.

fregante:
> The point is to know it is working, I could have made a new one. Whatever.

How could it be working if the count is totally missing everywhere? The feature just couldn't work at all.

> FYI it was not I took it from #3573

The screenshot in PRs is meant to demonstrate the issue or the bugfix. Taking random existing screenshots doesn't help and it's not what the template asks.

> I don’t know why I have to explain how to follow a template in every PR you open.

PR #4494, opened 2021-06-18
Fix `no-useless-split-diff-view` doesn't work newly loaded files

### LINKED ISSUE #4488: `no-useless-split-diff-view` doesn't work for diffs that are not loaded in time
1.  When the diff is shown after clicking a button, for example, "Large diffs are not rendered by default", `no-useless-split-diff-view` doesn't work.

    Example: <https://github.com/cpeditor/cpeditor/pull/223/files#diff-0f46ff9666ae4dc3789ffaed0f2c4a93fbd4469ab486af80f1b262cb44b7a038>

2.  When the diff is too long and the diff of the files at the bottom are lazy-loaded, `no-useless-split-diff-view` only works for files at the top.

    Example: https://github.com/cpeditor/cpeditor.github.io/commit/b349da5cefab537691e1988f74252759aff9c300

ouuan:
This fixes #4488.

Maybe it's off-topic, but I think the listener's name should be something like `onDiffFileLoad` instead of `onPrFileLoad`. It made me wonder whether it works on non-PR pages.

## Test URLs

I've tested this at:

-   [button](https://github.com/cpeditor/cpeditor/pull/223/files?diff=split#diff-0f46ff9666ae4dc3789ffaed0f2c4a93fbd4469ab486af80f1b262cb44b7a038)

-   [long](https://github.com/cpeditor/cpeditor.github.io/commit/b349da5cefab537691e1988f74252759aff9c300?diff=split)

### DIFF
--- a/source/features/hidden-review-comments-indicator.tsx
+++ b/source/features/hidden-review-comments-indicator.tsx
@@ -7,7 +7,7 @@ import {CommentIcon} from '@primer/octicons-react';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
-import onPrFileLoad from '../github-events/on-pr-file-load';
+import onDiffFileLoad from '../github-events/on-diff-file-load';
 import preserveScroll from '../helpers/preserve-scroll';
 
 // When an indicator is clicked, this will show comments on the current file
@@ -65,7 +65,7 @@ function observeFiles(): void {
 
 function init(): void {
 	observeFiles();
-	onPrFileLoad(observeFiles);
+	onDiffFileLoad(observeFiles);
 	delegate(document, '.rgh-comments-indicator', 'click', handleIndicatorClick);
 }
 
--- a/source/features/hidden-review-comments-indicator.tsx
+++ b/source/features/hidden-review-comments-indicator.tsx
@@ -7,8 +7,8 @@ import {CommentIcon} from '@primer/octicons-react';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
-import onDiffFileLoad from '../github-events/on-diff-file-load';
 import preserveScroll from '../helpers/preserve-scroll';
+import onDiffFileLoad from '../github-events/on-diff-file-load';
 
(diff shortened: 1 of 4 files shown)


fregante:
> Maybe it's off-topic, but I think the listener's name should be something like `onDiffFileLoad` instead of `onPrFileLoad`. It made me wonder whether it works on non-PR pages.

👍

ouuan:
Should I rename the function?

cheap-glitch:
Since the feature might run several times on the same page, I think the already visited diffs should be marked and excluded:
```ts
for (const diffTable of select.all('.js-diff-table:not(.rgh-diff-visited'))
```
then add the class in the loop (class name is just an example, it could be better)

source/features/hidden-review-comments-indicator.tsx
```diff
@@ -7,7 +7,7 @@ import {CommentIcon} from '@primer/octicons-react';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
-import onPrFileLoad from '../github-events/on-pr-file-load';
+import onDiffFileLoad from '../github-events/on-diff-file-load';
 import preserveScroll from '../helpers/preserve-scroll';
```
fregante, line 11:
```suggestion
import preserveScroll from '../helpers/preserve-scroll';
import onDiffFileLoad from '../github-events/on-diff-file-load';
```

source/features/show-whitespace.tsx
```diff
@@ -5,7 +5,7 @@ import * as pageDetect from 'github-url-detection';
 
 import features from '.';
 import getTextNodes from '../helpers/get-text-nodes';
-import onPrFileLoad from '../github-events/on-pr-file-load';
+import onDiffFileLoad from '../github-events/on-diff-file-load';
 import onNewComments from '../github-events/on-new-comments';
```
fregante, line 9:
```suggestion
import onNewComments from '../github-events/on-new-comments';
import onDiffFileLoad from '../github-events/on-diff-file-load';
```

Issue #4493, opened 2021-06-18
Add GitHub Actions workflow to ensure new features are correctly named and documented


fregante:
Add workflow to ensure that PRs with titles like `add xyz feature`, include a file called `xyz.tsx` and add a line with `xyz` in `readme.md`. If they aren't found in the diff, it should fail. 

This would have prevented this: https://github.com/sindresorhus/refined-github/issues/4483#issuecomment-863889718

And also fixes: https://github.com/sindresorhus/refined-github/issues/4320#issuecomment-833301366

Possible starting point, pseudo-code:

```yml
name: New feature checker

on:
  pull_request:
    types: [opened, edited]

jobs:
  checker:
    runs-on: ubuntu-latest
    if: contains(github.event.pr.title, 'Add') && contains(github.event.pr.title, 'feature')
    steps:
      - use: actions/checkout@v2
      - name: Find feature name
        id: feature 
        uses: AsasInnab/regex-action@v1
        with:
          regex_pattern: (?<=Add `)\w+(?=` feature)'
          search_string: github.event.pr.title
      - name: Has ${{ steps.feature.outputs.first_match }}.tsx been created?
        uses: mudlabs/simple-diff@v1.2.0
        with:
          path: source/features/${{ steps.feature.outputs.first_match }}.tsx 
      - name: Has the file been imported
        run: grep ${{ steps.feature.outputs.first_match }} source/refined-github.ts -q
      - name: Has the feature been documented
        run: grep ${{ steps.feature.outputs.first_match }} readme.md -q
```

Issue #4320, opened 2021-05-06
Write Webpack loader for readme.md 


fregante:
I think we’re parsing readme.md the wrong way. Instead of asking webpack to replace magic variables we parse and handle manually, we should be able to write `import featuresInfo from './readme.md'` and have a loader parse and keep track of the file as part of the dependency graph (as originally discussed in https://github.com/sindresorhus/refined-github/pull/3678)

We still need to “parse” refined-GitHub.ts to get the list of active features, but maybe there’s a solution to that too (separately)

fregante:
Docs: https://webpack.js.org/contribute/writing-a-loader/#guidelines

Probably the `/build/readme.loader.ts` file will look something like:


```js
export default function (source) {
  const list = /* extract feature list from `source` */
  return `export default ${JSON.stringify(list)}`;
}
```

The difference with [the current extraction code](https://github.com/sindresorhus/refined-github/blob/main/build/readme-parser.ts) is that that looks for each specific feature by name, while this loader should just detect and parse all lines that include a feature name/description/screenshot.

fregante:
The problem however is that this will no longer cause the build to fail since the build no longer looks for specific features in the file. Suggestions welcome.

fregante:
> The problem however is that this will no longer cause the build to fail since the build no longer looks for specific features in the file. Suggestions welcome.

Found a solution: Move this check to GHA https://github.com/sindresorhus/refined-github/issues/4008#issuecomment-855429007

fregante:
This would allow us to switch to esbuild instead 😃 

It also has a loader-like feature: https://esbuild.github.io/plugins/#load-callbacks

kidonng:
I actually prefer the old way of specifying feature meta. For single source of truth we can generate the README via a script, which is a lot more easier to deal with. Apparent downsides would be duplicate README template content and you will have to remember to generate it whenever you update the info...

kidonng:
Not am I asking to revert the whole efforts of parsing info from README, I just thought this separation feels unnatural for me, and exotic for casual contributors.

fregante:
> Not am I asking to revert the whole efforts of parsing info from README, I just thought this separation feels unnatural for me, and exotic for casual contributors.

Not sure what's exotic about writing documentation in Markdown. If anything, following some specific (parseable) format in a JS comment is exotic. 

The only exotic part is importing the magic variable `from '../readme.md'`, which happens exclusively in the options and background page maybe.

[...]

PR #4495, opened 2021-06-18
Meta: Add workflow that verifies new feature PRs

fregante:
Fixes https://github.com/sindresorhus/refined-github/issues/4493
Unblocks https://github.com/sindresorhus/refined-github/issues/4320

We have a few checks for features as part of the local build. This has led to some confusion, so I think it's best to move the checks to CI. This also allows the build improvements described in https://github.com/sindresorhus/refined-github/issues/4320

### DIFF
--- a/.github/workflows/features.yml
+++ b/.github/workflows/features.yml
@@ -1,4 +1,4 @@
-name: New feature
+name: Feature
 
 on:
   pull_request:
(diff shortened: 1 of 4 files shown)


fregante:
Merging this to see it in action, it can be improved later, I'm not super happy about it

yakov116:
Can you please make it pass instead of skip?

image

Super annoying since it does not auto collapse the checks

fregante:
Ooh that explains it.

I think this needs to be rewritten to check all the files all the time, to avoid https://github.com/sindresorhus/refined-github/pull/4694#issuecomment-903211294

I'll open a new issue

PR #4507, opened 2021-06-23
Support `conversation-activity-filter` in Files Changed tab

### LINKED ISSUE #4506: Files tab supports `conversation-activity-filter`
1. Visit https://github.com/sindresorhus/refined-github/pull/4395/files
2. Too many resolved comments are visible

I do not want to see resolved comments.
it is an obstacle to review.

KSH-code:
closes #4506

@fregante said
> It would be nice, but the feature is too complex to be adjusted to a page where it only hides something that it’s already small and usually disappears once outdated.
GitHub also added a Conversation dropdown that makes it easier to see what’s still unresolved

It is too complicated. 

However, I think that implementation with supporting in files is quite easy. so I make a draft PR.

Can you check this task? @fregante 

That issue I made, it was so lack of description.

## Test URLs
- https://github.com/sindresorhus/refined-github/pull/4395/files

## Screenshot
**Normal state**스크린샷 2021-06-23 오후 11 03 27

**Dropdown**
스크린샷 2021-06-23 오후 11 03 34

**Only unresolved comments**
스크린샷 2021-06-23 오후 11 03 37

**Show comments**
스크린샷 2021-06-23 오후 11 03 45

### DIFF
--- a/source/features/conversation-activity-filter.tsx
+++ b/source/features/conversation-activity-filter.tsx
@@ -109,8 +109,10 @@ async function handleSelection({target}: Event): Promise<void> {
 	);
 
 	// Update the state of the other dropdown
-	select(`.${dropdownClass} [aria-checked="false"][data-value="${currentSetting}"]`)?.setAttribute('aria-checked', 'true');
-	select(`.${dropdownClass} [aria-checked="true"]:not([data-value="${currentSetting}"])`)?.setAttribute('aria-checked', 'false');
+	if (!pageDetect.isPRFiles()) {
+		select(`.${dropdownClass} [aria-checked="false"][data-value="${currentSetting}"]`)!.setAttribute('aria-checked', 'true');
+		select(`.${dropdownClass} [aria-checked="true"]:not([data-value="${currentSetting}"])`)!.setAttribute('aria-checked', 'false');
+	}
 }
 
 function createRadio(filterSettings: State): JSX.Element {
--- a/source/features/conversation-activity-filter.tsx
+++ b/source/features/conversation-activity-filter.tsx
@@ -14,7 +14,8 @@ import onConversationHeaderUpdate from '../github-events/on-conversation-header-
 const states = {
 	default: '',
 	showOnlyComments: 'Only show comments',
-	showOnlyUnresolvedComments: 'Only show unresolved comments'
+	showOnlyUnresolvedComments: 'Only show unresolved comments',
+	hideResolvedComments: 'Hide resolved comments'
 };
 
 type State = keyof typeof states;
@@ -109,7 +110,7 @@ async function handleSelection({target}: Event): Promise<void> {
 	);
 
 	// Update the state of the other dropdown
-	if (!pageDetect.isPRFiles()) {
+	if (pageDetect.isConversation()) {
 		select(`.${dropdownClass} [aria-checked="false"][data-value="${currentSetting}"]`)!.setAttribute('aria-checked', 'true');
 		select(`.${dropdownClass} [aria-checked="true"]:not([data-value="${currentSetting}"])`)!.setAttribute('aria-checked', 'false');
 	}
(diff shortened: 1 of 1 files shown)


source/features/conversation-activity-filter.tsx
```diff
@@ -69,8 +77,10 @@ function processPage(): void {
 		return;
 	}
 
-	for (const item of select.all('.js-timeline-item')) {
-		if (select.exists('.js-comment[id^=pullrequestreview]', item)) {
+	for (const item of select.all('.js-timeline-item, .js-inline-comments-container')) {
```
KSH-code, line 80: `.js-inline-comments-container` is also included  in conversation.

source/features/conversation-activity-filter.tsx
```diff
@@ -99,8 +109,8 @@ async function handleSelection({target}: Event): Promise<void> {
 	);
 
 	// Update the state of the other dropdown
-	select(`.${dropdownClass} [aria-checked="false"][data-value="${currentSetting}"]`)!.setAttribute('aria-checked', 'true');
-	select(`.${dropdownClass} [aria-checked="true"]:not([data-value="${currentSetting}"])`)!.setAttribute('aria-checked', 'false');
+	select(`.${dropdownClass} [aria-checked="false"][data-value="${currentSetting}"]`)?.setAttribute('aria-checked', 'true');
+	select(`.${dropdownClass} [aria-checked="true"]:not([data-value="${currentSetting}"])`)?.setAttribute('aria-checked', 'false');
```
KSH-code, line 113: in Files Changed tab, sticky header is not supported

fregante: This is why I don’t want to make this change. We use `!.` because if GitHub changes some classes we will get some errors in the console. `?.` makes these errors silent

fregante: If the changes required are actually just these few lines, maybe we can just wrap this in a an if !isPRFiles

KSH-code: Yes, I' change into it wrapped as if
Thanks!

source/features/conversation-activity-filter.tsx
```diff
@@ -99,8 +109,10 @@ async function handleSelection({target}: Event): Promise<void> {
 	);
 
 	// Update the state of the other dropdown
```
fregante, line 111:
```suggestion
	// Update the state of the dropdown in the sticky header
```

fregante (approved):

fregante (changes requested):
Actually there are a couple of issues:

- "Only show comments" doesn't make sense here, it should be hidden
- The text should probably be customized to:
	- Show all
	- Hide resolved comments

source/features/conversation-activity-filter.tsx
```diff
@@ -99,8 +110,10 @@ async function handleSelection({target}: Event): Promise<void> {
...
-	select(`.${dropdownClass} [aria-checked="false"][data-value="${currentSetting}"]`)!.setAttribute('aria-checked', 'true');
-	select(`.${dropdownClass} [aria-checked="true"]:not([data-value="${currentSetting}"])`)!.setAttribute('aria-checked', 'false');
+	if (pageDetect.isConversation()) {
```
KSH-code, line 113: flipped the condition you suggested is because `if not condition` is less readable than now.

source/features/conversation-activity-filter.tsx
```diff
@@ -153,12 +170,15 @@ async function addWidget(header: string): Promise<void> {
 
 async function init(): Promise<void> {
 	await addWidget('#partial-discussion-header .gh-header-meta :is(clipboard-copy, .flex-auto)');
-	await addWidget('#partial-discussion-header .gh-header-sticky :is(clipboard-copy, relative-time)');
+	if (pageDetect.isConversation()) {
```
KSH-code, line 173: this is added so that fixing the error from Files tab

KSH-code:
I was not able to think the radio options! I thought only about support feature 
Again I looked into options, it was weird 😂 
Thanks!

KSH-code:
@fregante 
can you please review here again?
If it cannot be accepted as feature, I'll close.

fregante:
The feature is reasonable, I just haven't had time to review it

source/features/conversation-activity-filter.tsx
fregante, line 113: Why does this condition exist? The PR Files tab also has a sticky bar and it should also have the dropdown

firefox

fregante (changes requested):

KSH-code:
to put element in sticky header is hard for  me so it'll be finished. but not anytime soon.

fregante:
Feel free to open a new PR when done

PR #4514, opened 2021-06-24
Fix `no-useless-split-diff-view` on GitHub Enterprise

### LINKED ISSUE #4483: `no-useless-split-diff-view` is broken on GitHub Enterprise
> here's how a recent diff looks with this feature enabled:
> 
> chrome_2021-06-17T090052
> 
> What do you think changed in this commit? Well it looks like that line was added - but it wasn't. The normal split diff looks like this:
> 
> chrome_2021-06-17T090115
> 
> You can see there's a subtle difference between the commits which is actually very important (there was a bug!)
> 

_Originally posted by @danielsamuels in https://github.com/sindresorhus/refined-github/issues/4399#issuecomment-863023847_

> Strange, can't replicate.
> 
> Diff that's correctly unified: https://github.com/sindresorhus/refined-github/commit/f60cb71e0bc2708ebda622d6c684489241fd367a
> 
> Reportedly problematic diff: https://github.com/sindresorhus/refined-github/commit/26e778fd85f7db6c3d5bc6f508e23c77f3ed4a20

_Originally posted by @fregante in https://github.com/sindresorhus/refined-github/issues/4399#issuecomment-863172312_

> This is on a GitHub Enterprise instance running 2.22.10, perhaps it's an incompatibility with that version?

_Originally posted by @danielsamuels in https://github.com/sindresorhus/refined-github/pull/4399#issuecomment-863173685_

> I think I know what's happening. It checks the left column first, it doesn't find the expected element (`[data-split-side="${side}"]:is(.blob-code-addition, .blob-code-deletion)`) and therefore it assumes it's empty.

_Originally posted by @fregante in https://github.com/sindresorhus/refined-github/pull/4399#issuecomment-863175358_

cheap-glitch:
Fixes #4483 

## Test URLs
* [Diff should be collapsed](https://github.com/sindresorhus/refined-github/commit/f60cb71e0bc2708ebda622d6c684489241fd367a?diff=split)
* [Diff should be untouched](https://github.com/sindresorhus/refined-github/commit/26e778fd85f7db6c3d5bc6f508e23c77f3ed4a20?diff=split)

## Screenshot
n/a

This needs to be tested on GHE 2.22.10 or lower (both diffs should be left untouched).

### DIFF
--- a/source/features/no-useless-split-diff-view.tsx
+++ b/source/features/no-useless-split-diff-view.tsx
@@ -15,6 +15,10 @@ function isUnifiedDiff(): boolean {
 function init(): void {
 	for (const diffTable of select.all('.js-diff-table:not(.rgh-no-useless-split-diff-view-visited)')) {
 		diffTable.classList.add('rgh-no-useless-split-diff-view-visited');
+		if (!select.exists('.blob-code-addition, .blob-code-deletion', diffTable)) {
+			continue;
+		}
+
 		for (const side of ['left', 'right']) {
 			if (!select.exists(`[data-split-side="${side}"]:is(.blob-code-addition, .blob-code-deletion)`, diffTable)) {
 				// eslint-disable-next-line unicorn/prefer-dom-node-dataset -- CSS file has the same selector, this can be grepped
--- a/source/features/no-useless-split-diff-view.tsx
+++ b/source/features/no-useless-split-diff-view.tsx
@@ -15,10 +15,6 @@ function isUnifiedDiff(): boolean {
 function init(): void {
 	for (const diffTable of select.all('.js-diff-table:not(.rgh-no-useless-split-diff-view-visited)')) {
 		diffTable.classList.add('rgh-no-useless-split-diff-view-visited');
-		if (!select.exists('.blob-code-addition, .blob-code-deletion', diffTable)) {
-			continue;
-		}
-
 		for (const side of ['left', 'right']) {
 			if (!select.exists(`[data-split-side="${side}"]:is(.blob-code-addition, .blob-code-deletion)`, diffTable)) {
 				// eslint-disable-next-line unicorn/prefer-dom-node-dataset -- CSS file has the same selector, this can be grepped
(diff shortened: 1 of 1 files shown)


source/features/no-useless-split-diff-view.tsx
```diff
@@ -15,6 +15,10 @@ function isUnifiedDiff(): boolean {
 function init(): void {
 	for (const diffTable of select.all('.js-diff-table:not(.rgh-no-useless-split-diff-view-visited)')) {
 		diffTable.classList.add('rgh-no-useless-split-diff-view-visited');
+		if (!select.exists('.blob-code-addition, .blob-code-deletion', diffTable)) {
+			continue;
+		}
```
cheap-glitch, line 20: This seems like the simplest solution to me, but I'm not sure if the check needs to happen for every diff table. Maybe once at the top of `init()` would be enough?

fregante: Good point! However this doesn’t cover the data attribute. The selector should include that too, ideally using :is(data etc left, data etc right):is(addition, deletion) 

Since this feature does not run before dom ready, you can move this into an `exclude` so that it can be disabled completely.

fregante: Also it needs a comment in the code to explain the selector and ensure that the condition in the code will make you change this condition too

cheap-glitch: > you can move this into an exclude so that it can be disabled completely

This is a nitpick but isn't the usual procedure in this case to return `false` from the `init()` function? Since the feature *could* run on the page, but decides not to so it doesn't break things.

fregante: > Since the feature _could_ run on the page, but decides not to so it doesn't break things.

That's contradictory. It's not compatible with the page, so it _cannot_ run. 

After https://github.com/sindresorhus/refined-github/pull/3105 we started preferring `exclude` regardless of the reason, in practice they're equivalent, but moving them to `exclude` is just clearer.

Ideally we would show "feature skipped because incompatible  etc etc" in the console and options page, but we don't have such a system in place. There are a couple of features that are disabled the same way in Firefox and Safari: They _could_ run, but they don't because the browser is incompatible.

PR #4444, opened 2021-06-04
Small bug fixes

yakov116:
Fixes #3945

Make features faster closes #3673

- Make `list-prs-for-file` faster
- Make `default-branch-button` faster
- Move addAfterBranchSelector to its own file
- Make `convert-release-to-draft` faster
- Fix `comment-time-machine-links` date check - https://github.com/sindresorhus/refined-github/blob/main/source/features/action-used-by-link.tsx?rgh-link-date=2021-04-28T07%3A12%3A39Z
- Fix duplication in show-whitespace
- Remove outdated comment
- Correct `show-associated-branch-prs-on-fork` display text
- Fix console error in clean-conversation-filters
- Fix types
- Fix console error in faster-pr-diff-options - https://github.com/pixiebrix/pixiebrix-extension/pull/462/files/1e4b6cd476a9bd7f4d5754a86b09087130991fc6..9377c3db676776e997a3317556bd97d5ef62635f
- Sort `show-associated-branch-prs-on-fork` - Will get the last 100
- Exclude issue links in `mark-merge-commits-in-list` - https://github.com/pixiebrix/pixiebrix-extension/commits/main

PR #4516, opened 2021-06-24
Hide long hunks on only one side for `no-useless-split-diff-view`

ouuan:
Now for hunks that are only non-empty on a single side and contains more than 5 lines, the empty side will be hidden. The line number on the hidden side is also hidden.

The threshold "5" could be changed.

## Test URLs

[Right only](https://github.com/sindresorhus/refined-github/pull/4444/files#diff-053150b640a7ce75eff69d1a22cae7f0f94ad64ce9a855db544dda0929316519R713)

[Left only](https://github.com/sindresorhus/refined-github/pull/4444/files#diff-053150b640a7ce75eff69d1a22cae7f0f94ad64ce9a855db544dda0929316519L1675)

[Short left-only](https://github.com/sindresorhus/refined-github/pull/4444/files#diff-26e829c321ccba564beb4bb33adb6e5a71106315a1e35b430d5dd1d7b383aa05L11)

[Short right-only](https://github.com/sindresorhus/refined-github/pull/4444/files#diff-26e829c321ccba564beb4bb33adb6e5a71106315a1e35b430d5dd1d7b383aa05R10)

[Both sides](https://github.com/sindresorhus/refined-github/pull/4444/files#diff-053150b640a7ce75eff69d1a22cae7f0f94ad64ce9a855db544dda0929316519L6018)

## Screenshots

Right only

Left only

Short hunks

Both sides

### DIFF
--- a/source/features/no-useless-split-diff-view.css
+++ b/source/features/no-useless-split-diff-view.css
@@ -2,8 +2,11 @@
 	table-layout: auto !important;
 }
 
+[data-rgh-hide-empty-split-diff-side='left'] :is([data-hunk], .blob-expanded) td:nth-child(1),
 [data-rgh-hide-empty-split-diff-side='left'] :is([data-hunk], .blob-expanded) td:nth-child(2),
-[data-rgh-hide-empty-split-diff-side='right'] :is([data-hunk], .blob-expanded) td:nth-child(4) {
+[data-rgh-hide-empty-split-diff-side='right'] :is([data-hunk], .blob-expanded) td:nth-child(3),
+[data-rgh-hide-empty-split-diff-side='right'] :is([data-hunk], .blob-expanded) td:nth-child(4),
+[data-rgh-no-useless-split-diff-view-hide-hunk] .empty-cell {
 	display: none !important;
 	width: 0 !important;
 }
--- a/source/features/no-useless-split-diff-view.tsx
+++ b/source/features/no-useless-split-diff-view.tsx
@@ -57,7 +57,7 @@ function init(): void {
 			}
 
 			if (tds[1].classList.contains('blob-code-context')) {
-				// not in a hunk
+				// Not in a hunk
 				handleHunk(currentHunkEmpty, currentHunk);
 				currentHunkEmpty = [true, true];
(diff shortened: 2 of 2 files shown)


fregante:
Thank you for the PR! Unfortunately I think this may be a bit confusing especially on short lines. The right-hand column will appear as if there’s a hole in the code similarly as if the code was removed, even if the code has been added

ouuan:
> Unfortunately I think this may be a bit confusing especially on short lines. The right-hand column will appear as if there’s a hole in the code similarly as if the code was removed, even if the code has been added

I think the background color clearly indicates whether it's added or removed :thinking:

ouuan:
Maybe we can separate this from `no-useless-split-diff-view` so that it can be disabled alone.

fregante:
> > Unfortunately I think this may be a bit confusing especially on short lines. The right-hand column will appear as if there’s a hole in the code similarly as if the code was removed, even if the code has been added
> 
> I think the background color clearly indicates whether it's added or removed 🤔

Yes but it takes more cognitive effort, especially because this formatting would not be common and it might look broken (some hunks are split, some aren’t, even if they’re all adding or all deleting)

> Maybe we can separate this from `no-useless-split-diff-view` so that it can be disabled alone.

We only add features that make sense for most people anyway. Having it disableable doesn’t mean we can add confusing features.

ouuan:
> We only add features that make sense for most people anyway. Having it disableable doesn’t mean we can add confusing features.

Yeah, I thought this was confusing for only a few people :thinking:

ouuan:
Can we ask for more people's opinions?

ouuan:
> Yes but it takes more cognitive effort

I think it does take more cognitive effort in short hunks, so I only hid long hunks. Will you think it's better if the threshold is bigger?

The basic idea is that long hunks on only one side are a big waste of space. And when the hunk is big, you don't need to tell whether it's an addition or deletion very often.

ouuan:
Any reply to https://github.com/sindresorhus/refined-github/pull/4516#issuecomment-868975372? Do you think it's confusing even when the hunk is bigger, e.g. takes a whole screen?

PR #4123, opened 2021-03-17
Add `confirm-short-comments` feature

darkred:
Closes #3911 

## Test URLs

Any new/existing issue page: e.g.
https://github.com/sindresorhus/refined-github/issues/new
https://github.com/sindresorhus/refined-github/issues/3911

## Screenshot

111637477-23fc9500-8802-11eb-8c1b-4023b7b1c371

PR #4520, opened 2021-06-26
Add `comment-on-draft-pr-indicator` feature

### LINKED ISSUE #3857: Indicator that you're about to comment on a Draft PR
I often mix up whether a PR is a draft and final as I do not always pay attention to the subtle difference in the icons.

The latest incident motivating this feature request is the following PR comment.

---
draft-background-color

---

For our team the PR is treated completely differently if it is a draft. Reviewing a draft as a final PR is annoying for the author and a waste of time and miss of target for the reviewer.

My proposal is to draw the entire background on draft PRs in another color, _or_ to add some other visual clue that is similarly impossible to not see.

I expect to see this change on https://github.com/sindresorhus/refined-github/pull/3841 and all sub-pages like https://github.com/sindresorhus/refined-github/pull/3841/files if https://github.com/sindresorhus/refined-github/pull/3841 is a draft.

darkred:
Closes #3857 

## Test URLs

Any draft PR, e.g.
https://github.com/darkred/test/pull/6

## Note to reviewers:

~~The selector for the "Draft" label, `#partial-discussion-header span[title="Status: Draft"]` ,
matches twice in draft PR pages (but none in non-draft ones, of course).
Do I have to make it more specific, nevertheless?~~


yakov116:
> The selector for the "Draft" label, `#partial-discussion-header span[title="Status: Draft"]` ,

There is a detection for isDraft do 
```
exclude: [
	() => !isDraft
]
```

fregante:
Screen Shot 12

It should also cover the Files tab, where comments on temporary code are more likely to happen.

source/features/comment-to-draft-pr-button.tsx
```diff
@@ -0,0 +1,22 @@
+import select from 'select-dom';
+import * as pageDetect from 'github-url-detection';
+
+import features from '.';
+import onNewComments from '../github-events/on-new-comments';
+
+function init(): void {
+	if (select.exists('#partial-discussion-header span[title="Status: Draft"]')) {
+		const commentButton = select('#partial-new-comment-form-actions > div > div.color-bg-secondary.ml-1 > button');
```
fregante, line 9: Selectors should be shorter and smarter than this. The button has a `.btn-primary` class and `type=submit` that are way more specific and don't need almost any other specifiers. An example could be `#partial-new-comment-form-actions .btn-primary`

darkred: Thank you both for the suggestions. I followed them all in the new commit + updated the screenshot.

source/features/comment-to-draft-pr-button.tsx
```diff
@@ -1,19 +1,42 @@
...
+	}
+}
+
 function init(): void {
-	if (select.exists('#partial-discussion-header span[title="Status: Draft"]')) {
-		const commentButton = select('#partial-new-comment-form-actions > div > div.color-bg-secondary.ml-1 > button');
+	let commentButton;
+	if (pageDetect.isPRConversation()) {
+		commentButton = select('#partial-new-comment-form-actions .btn-primary');
 		commentButton!.innerHTML = 'Comment to Draft PR';
```
yakov116:
```suggestion

	if (pageDetect.isPRConversation()) {
		select('#partial-new-comment-form-actions .btn-primary')!.innerHTML = 'Comment to Draft PR';
```

Can we use textContent?

darkred: yes, but, e.g. for the Conversation tab, the 'Comment' button textContent is 
```
\n  \n  \n        Comment\n\n  \n\n
```
and so I'd have to use `.replace('Comment, Comment to Draft PR')`
instead of putting fixed string.

Are you ok with that?

yakov116: Its fine don't worry about the '\n'. I just tried it it will work correctly

darkred: I have just tried it (in the Conversation tab), and unfortunately, the name change occurs multiple times, 
if I submit a comment(or delete comments), focus the text area and start typing:

<details>

<summary>Here is the code with <code>replace</code>+<code>textContent</code> instead of fixed </code>innerHTML</
code></summary>

```js

import features from '.';

function handleClicks(): void {
	const buttonsSingleComment = select.all([
		'table:not([style="display: none"]) .review-simple-reply-button',
		'table:not([style="display: none"]) .add-comment-label' // For when adding a single comment after having started a review
	]);
	for (const element of buttonsSingleComment) {
		element.textContent = element.textContent!.replace('Add single comment', 'Add single comment to Draft PR');
	}

	const buttonsStartReview = select.all('table:not([style="display: none"]) .start-review-label');
	for (const element of buttonsStartReview) {
		element.textContent = element.textContent!.replace('Start a review', 'Start a review to Draft PR');
	}
}

function init(): void {
	let commentButton;
	if (pageDetect.isPRConversation()) {
		commentButton = select('#partial-new-comment-form-actions .btn-primary');
		commentButton!.textContent = commentButton!.textContent!.replace('Comment', 'Comment to Draft PR');
	} else if (pageDetect.isPRFiles()) {
		delegate('.diff-table', 'button', 'click', handleClicks);
	}
}

	include: [
		pageDetect.isPRConversation,
		pageDetect.isPRFiles
	],
	exclude: [
		() => !pageDetect.isDraftPR
	],
	additionalListeners: [
		onNewComments
	],
	init
});

```

</details>

With the previous version this doesn't happen.

---

The only solution I see is to use something like this, which is apparently too easy to break:

```js
.replace('\n  \n  \n        Comment\n\n  \n\n', '\n  \n  \n        Comment to Draft PR \n\n  \n\n')
```

darkred: Or use a regex with replace() with capturing groups around the whitespace characters.

fregante: Yakov is not suggesting to use replace. Just use `=` and it works. The whitespace is useless

source/features/comment-to-draft-pr-button.tsx
```diff
@@ -0,0 +1,45 @@
...
+	}
+}
+
+function init(): void {
+	let commentButton;
+	if (pageDetect.isPRConversation()) {
+		commentButton = select('#partial-new-comment-form-actions .btn-primary');
+		commentButton!.innerHTML = 'Comment to Draft PR';
+	} else if (pageDetect.isPRFiles()) {
+		delegate('.diff-table', 'button', 'click', handleClicks);
```
fregante, line 29: I don't think that's the right way. We add custom buttons like `table-input` on the comment forms on the Files tab. You can copy what they use

source/features/comment-to-draft-pr-button.tsx
darkred: Thanks, I just did. 

---

Btw, I feel once again embarrassed being confused over the same things over and over again..
Could I ask please to add some extra notes/warnings in https://github.com/sindresorhus/refined-github/blob/main/contributing.md to make sure people like me always follow them?

What comes to mind:

- what is the preferable way to sort imports for this repo
- never use `innerHTML` because it's flagged by the extension stores. Just use `textContent`.  
- use `select()`, not `querySelector`
- with `select()`, when having multiple selectors, split them to multiple lines
- Selectors should be short and smart/concise. Use those class and attributes combinations that are as much specific as possible and don't need almost any other specifiers.
- Use ternary `if` when possible.

Do you think these "tips" can be useful to contributors, or they are too self-explanatory/newbie level?

*I could provide a PR with these of course, but I need your opinion first.*

source/features/comment-to-draft-pr-button.tsx
darkred: Sorry, but I don't quite understand.

With this feature, the purpose is to simply change the name of the buttons, 
not (remove the existing ones and) add new buttons with a different name.
Are you suggesting to use `remove()` and `.after()`? 
https://github.com/sindresorhus/refined-github/blob/ae07d212f364ba0193958b95729e4e99d4653b90/source/features/table-input.tsx#L40-L68

source/features/comment-to-draft-pr-button.tsx
fregante: Don't worry too much about these small mistakes. If the PR is salvageable you can learn about the project’s code style over time.

Some of those rules _could_ be part of the contributor guide, but I'd rather have eslint check for them. It's weird that `innerHTML` isn't already caught by eslint, e.g. 

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-dom-node-text-content.md
https://github.com/salesforce/eslint-plugin-lwc/blob/master/docs/rules/no-inner-html.md

source/features/comment-to-draft-pr-button.tsx
fregante: It looks like a plain loop is all that should be needed and GitHub will preserve the changes, no need for onNewComments at all:

https://github.com/sindresorhus/refined-github/blob/ae07d212f364ba0193958b95729e4e99d4653b90/source/features/table-input.tsx#L39-L40

darkred:
1

source/features/comment-to-draft-pr-button.tsx
darkred: Regarding `onNewComments`:  I tested removing it, but then, in 'Conversations' tab, the 'Comment' button name no longer change after submitting a new comment. Therefore it seems necessary.

And regarding my use of `delegate`:
in 'Files 'tab, e.g. https://github.com/darkred/test/pull/6/files
initially, there's no comment textarea,
and no `table:not([style="display: none"]) .review-simple-reply-button/.start-review-label` button elements:
these elements only exist *after* you click in a row ([capture](https://user-images.githubusercontent.com/723651/123557151-fc6d7f00-d797-11eb-89e1-03b0a892464b.gif))
therefore I think it's necessary to delegate clicks on the .diff-table element.
You can't loop on them, before they exist.

Am I not right? Have I got something wrong?

orrc:
Just a small note as a native English speaker… 😃

I would suggest that the prepositions need to be changed ("to" versus "on"), and I would argue that "**draft**" rather than "**Draft**" should be used.

"Draft" is being used as an adjective, it looks a bit nicer as lowercase within a sentence (IMO), and that's how [GitHub uses it](https://docs.github.com/en/github/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request#converting-a-pull-request-to-a-draft) in their docs and UI for sentences, e.g. "Convert to draft", or "… mark a draft pull request as ready …", plus also [elsewhere](https://github.com/sindresorhus/refined-github/blob/2c6cbdf9/source/features/one-click-pr-or-gist.tsx#L25) in Refined GitHub.

| Current text in this PR | Proposed text |
|-|-|
| Add single comment to Draft PR | Add single comment to draft PR |
| Start a review to Draft PR | Start a review on draft PR |
| Comment to Draft PR | Comment on draft PR |

darkred:
Thanks for your comment. 

Regarding `on` --> `to`, thanks for noting, yes, as verb it's "comment on/upon/about", there's no "comment to" . I'll change it.

Regarding `draft PR` --> `Draft PR`, indeed it looks nicer as lowercase, 
but I'd argue that, since the gist of this feature is whether the PR is **draft**, 
having it in uppercase, i.e. `Draft`, would add emphasis.
I'd like to ask for the reviewers opinion, please.

source/features/comment-to-draft-pr-button.tsx
fregante: Where in `table-input` is there `:not([style="display: none"])`? It's not there because it's not needed. As I tried saying before, there's a template that GitHub copies, including all the changes we make to it. Your selector explicitly excludes it.

Same goes for `onNewComments` in this feature, it's not needed.

source/features/comment-to-draft-pr-button.tsx
```diff
@@ -0,0 +1,45 @@
...
+	}
+}
+
+void features.add(__filebasename, {
+	include: [
+		pageDetect.isPRConversation,
+		pageDetect.isPRFiles
+	],
+	exclude: [
+		() => !pageDetect.isDraftPR
```
fregante, line 39:
```suggestion
		() => !pageDetect.isDraftPR()
```

source/features/comment-to-draft-pr-button.tsx
yakov116: Maybe it needs dontDeduplicate. (I forgot what we called it)

darkred: For the 'Files' tab, `onNewComments` is indeed not needed::
I had used the `:not([style="display: none"])` selector part
because I wrongly thought that these hidden elements shouldn't also match my selector.
It indeed works ok after removing that selector part + `onNewComments`.
I just removed the selector part.

BUT, on the Conversations tab, I still believe the `onNewComments` is still needed:
My relevant selector is: `#partial-new-comment-form-actions .btn-primary`. 
There's no `:not([style="display: none"])` in it, and it only matches 1 element, the Comment textarea to the bottom.
I can't find any hidden element that is a "template comment box element" 
that can also be selected with a variation of my selector above.
Therefore I can't seem to make use of a `for (const el of __)` loop.

darkred:
With the new commits, I: 
- added the suggestions by @orrc ,
- renamed the feature + updated the screenshot, and 
- removed the unnecessary `table:not([style="display: none"])` selector part.

source/features/comment-to-draft-pr-button.tsx
darkred: Also, I tried adding the following inside `void features.add(` but it didn't help:

```js
	deduplicate: 'false',
```

darkred: Regarding the Conversations tab, maybe this hidden 'template comment box element' contains code only for the editor's toolbar buttons (a button of which, `md-task-list`, is used as reference in [table-input](https://github.com/sindresorhus/refined-github/blob/main/source/features/table-input.tsx)), 
but not for the 'Comment' button (or the 'Close pull request' one) too?

I couldn't find any such hidden additional button, as I describe in my previous comment.

darkred:
@fregante 
I see you closed this PR 😞 Isn't it salvageable??
It's in functioning state - what I cannot find (and I think it doesn't exist) is a hidden 'template' for the submit buttons, 
that would make `onNewComments` unnecessary.

fregante:
You need to follow the reviews if you want to open PRs here, otherwise we're just wasting time.

What I suggested works exactly like it works for `table-input`. There's one situation where it doesn't work for either one and that doesn't even matter because it only happens after a review started and the user is leaving a comment on the same thread.

gif

darkred:
I followed all reviews, with the only exception of managing to remove the need for `onNewComments`.

The way your presented the alternative approach, appeared to me as a probability, not a certainty:

> It *looks like* a plain loop is all that should be needed and GitHub will preserve the changes, no need for onNewComments at all:

> This is also why `onNewComments` is *very likely* not needed.

and, after my various efforts (with my limited experience compared to you), I came to the result that it wasn't possible.

I see that with your new PR you proved the opposite.
I'm sorry I couldn't make this effort to the end.

darkred:
@fregante 
Could you please help me clarify one last thing, about how to properly contribute to the repo?

&nbsp;&nbsp;
You said above that:

> You need to follow the reviews if you want to open PRs here, otherwise we're just wasting time.

Also, in your new PR you have this commit message  https://github.com/sindresorhus/refined-github/pull/4541/commits/b385a8473a9c3eba90dbcb8538e00e3777b9541e :

> Do what the reviewers say

In that commit, the only significant change is that you removed the `onNewComments` listener.

&nbsp;
I would never disregard/ignore/argue with the reviewers.
You are way more experienced than me, your reviews can only be helpful and instructional.

With your last review, you suggested that use `onNewComments` can be avoided.
Unfortunately, despite my efforts, although I tried as much as I could, I couldn't make it work (the button wouldn't stay after updating). With my comments [1](https://github.com/sindresorhus/refined-github/pull/4520#discussion_r659367108), [2](https://github.com/sindresorhus/refined-github/pull/4520#discussion_r659996890), [3](https://github.com/sindresorhus/refined-github/pull/4520#discussion_r663997924) I just wanted to describe my failed efforts.

Yes, I admit my comments are sometimes too verbose/detailed.
Comparing them to your terse style of commenting (yours and the other collaborators),
I understand that they might appear tedious to read them through, to such experienced people like you,
but that's my way of expressing, I want to provide as much details as possible.

&nbsp;
So, I just wanted to ask you this:
Did you close the PR because you thought I disregard the reviewers (your review)?
Would you prefer me, instead of writing all these comments explaining my failed efforts,
to just immediately remove the `onNewComments` listener via a new commit,
and briefly say the following:

> without `onNewComments` the button won't stay after updating,
> and setting `deduplicate: false` didn't help.
> I can't find a way to make it work without `onNewComments`. Please help/suggest ideas.

But, wouldn't we end up to the same dead-end like in this previous closed PR of mine https://github.com/sindresorhus/refined-github/pull/4123#issuecomment-803487198 
i.e. that you would still have to close the PR:

> If you can't finish the PR autonomously you're asking us to figure it out with you, which means that the PR isn't useful to us, sorry.

I see that you made a lot of improvements from your commit https://github.com/sindresorhus/refined-github/pull/4541/commits/6a30be8f1d7672aa939ad9edaaea1ee9ce1ec531 and on, 
making use of a lightweight listener instead of `onNewComments` to catch dynamic "Add comment", 
something I couldn't achieve myself (I was unaware of `onReplacedElement`).

&nbsp;
I am asking you all this, because I want to have clear in my mind how to properly contribute to your repo,
I'd hate to be a burden, or waste any of your time - I only want to contribute.

fregante:
This was mostly because in this PR a few things had to be repeated several times before being applied, with your focus going on the wrong part of our suggestion. For example you used replace() when Yakov gave you an exact code suggestion that would have just worked, and the table-input suggestion that should have resulted in the thought: “table-input works, why doesn’t it work for the button?” instead of “I doesn’t work for the button, I can’t use it.” 

Perhaps the easiest way to go about it is to make the requested changes and then show how they *don't* work in a gif. This way we can see whether the suggestion was attempted correctly. For example if you implemented the loop with the broken selector my suggestion would have been to just drop the selector and it would have worked most of the time. 

Occasionally suggestions will be wrong though, but that’s the problem with commenting on code without running it. The other problem is that once I have to deal with the code myself, by running it and figure it out, it’s easier to just send a PR myself. And that’s ok sometimes, we can’t merge every PR.

Regarding my “probability” suggestion: 😃 

BE2E3BE4-8FBA-4B87-B4DF-9D0E52608153

Also it would personally help me read your comments if you didn’t add a line break after commas 😂 as that is not customary in English. But that’s just my opinion, man 😜

darkred:
Thanks for replying.

&nbsp;
Yes, my English is far from perfect, and, yes, there can be misunderstanding in a discussion when one is not a native speaker. But I think that most of the time the other understands the gist of what I'm trying to say.

And about my linebreaks after commas: I know it's not necessary, but I don't even realize that I put commas. The point for me is to have linebreaks, as it seems to me that it makes the text easier to read. I guess I'm wrong, I'll try to avoid it.

fregante:
> my English is far from perfect

Don't say that at all! Your english is 💯 !

> it seems to me that it makes the text easier to read

Yeah it's common among programmers, generally I prefer flowing text since the browser will pick the best place to wrap lines.

fregante:
Related meme

image

fregante:
By the way if you don't mind me asking, are you from Greece?

darkred:
Yes, I am from Greece. Neighbor of Italy 🙂

PR #4522, opened 2021-06-28
Fix: `sticky-conversation-sidebar` overlaps "Open with" dropdown

### LINKED ISSUE #4490: "Open with" PR dropdown overlapping `sticky-conversation-sidebar`
On any PR, such as https://github.com/sindresorhus/refined-github/pull/4395.

This is not an easy issue because it's due to z-index. A solution must be found by looking at the previous commits that affected the z-index of the sidebar and avoid regressions.

## Sticky

Screen Shot 19

## Not sticky

Screen Shot 20

cheap-glitch:
Fixes #4490 
Related: #3754, #4364 

## Test URLs
- On this page:
  * click the "Open with" button → the dropdown should be over the sidebar
  * click the "Customize" button next to "Notifications" → the modal's background should cover everything (except the sticky header)
- On a repo root where you can edit the sidebar
  * click the gear icon → the modal's background should cover everything

## Screenshot
image

### DIFF
--- a/source/features/sticky-sidebar.css
+++ b/source/features/sticky-sidebar.css
@@ -4,9 +4,6 @@
 
 .rgh-sticky-sidebar {
 	position: sticky;
-}
-
-:not(#partial-discussion-sidebar).rgh-sticky-sidebar {
 	top: 0;
 
 	/* `sticky` will trap the `fixed` dialog’s z-index in some sidebars #3732 */
--- a/source/features/sticky-sidebar.css
+++ b/source/features/sticky-sidebar.css
@@ -23,3 +23,8 @@
 .rgh-sticky-sidebar + .border-top {
 	display: none;
 }
+
+.gh-header-actions tab-container {
+	/* Needed so the "Open with" PR dropdown is over the sidebar #4490 */
+	z-index: 91 !important;
+}
--- a/source/features/sticky-sidebar.css
+++ b/source/features/sticky-sidebar.css
@@ -18,8 +18,7 @@
 	top: 76px !important; /* Sticky header's height + sidebar item's margin */
 
 	/* Needed so the backdrop of notification modal is over the other elements in the conversation */
-	/* Lower than "Open with" PR dropdown #4490 */
-	z-index: 80;
+	z-index: 1;
 }
 
 /* Hides the last divider (on pull requests) to avoid https://user-images.githubusercontent.com/10238474/62282128-af6fb980-b457-11e9-8b18-c29687b88da1.gif */
--- a/source/features/sticky-sidebar.css
+++ b/source/features/sticky-sidebar.css
@@ -7,6 +7,10 @@
 	top: 0;
 
 	/* `sticky` will trap the `fixed` dialog’s z-index in some sidebars #3732 */
+	z-index: 1;
+}
+
+:not(#partial-discussion-sidebar).rgh-sticky-sidebar {
 	/* Higher than sticky readme header #4192 */
 	z-index: 90;
 }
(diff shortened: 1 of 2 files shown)


fregante:
I'd rather not have to start affecting every other element’s z-index if possible. It appears that if we move the sidebar index to 80 it works correctly. Can you find what made me set it to 92 and can we change that instead?

fregante:
Rather, reading the inline comment, we should probably just limit that 90 index to the readme pages because _that's_ where the problem is.

source/features/sticky-sidebar.tsx
```diff
@@ -8,7 +8,7 @@ import features from '.';
 
 const deinit: VoidFunction[] = [];
 // Both selectors are present on conversation pages so we need to discriminate
-const sidebarSelector = pageDetect.isRepoRoot() ? '.repository-content .flex-column > :last-child [data-pjax]' : '#partial-discussion-sidebar';
+const sidebarSelector = pageDetect.isRepoRoot() ? '.repository-content .flex-column > .flex-shrink-0 > [data-pjax]' : '#partial-discussion-sidebar';
```
cheap-glitch, line 11: The selector for the repo root page was broken on latest Firefox. Fixed it but it's still brittle, should be use the `select().closest()` strategy?

fregante: Yes please if possible

source/features/sticky-sidebar.css
```diff
@@ -4,6 +4,9 @@
 
 .rgh-sticky-sidebar {
 	position: sticky;
+}
+
+:not(#partial-discussion-sidebar).rgh-sticky-sidebar {
```
fregante, line 9: Is there anything better than :not? Like `.readme .sticky`

cheap-glitch: Not that I've found. We could do `.BorderGrid--spacious.rgh-sticky-sidebar` but that's not much better

source/features/sticky-sidebar.tsx
fregante: Ah with selector-observer it's a little more complicated.

I think in that case the best option would be to select the sidebar with the shortest selector and then exclude items in its listener. Maybe.

source/features/sticky-sidebar.css
```diff
@@ -15,8 +18,7 @@
 	top: 76px !important; /* Sticky header's height + sidebar item's margin */
 
 	/* Needed so the backdrop of notification modal is over the other elements in the conversation */
-	/* Higher than sticky readme header #4192 */
-	z-index: 90;
+	z-index: 1;
```
fregante, line 21: It seems this is the lowest we can go. 90 was only needed on the readme. Last change was in https://github.com/sindresorhus/refined-github/pull/4364

source/features/sticky-sidebar.css
fregante: Is this change needed at all? It looks like it will just be overridden by the next selector

fregante, line 9: Is this hunk needed at all? It looks like it will just be overridden by the next selector.

```suggestion
```

I think this is ready to merge after this change

fregante (approved):

source/features/sticky-sidebar.css
cheap-glitch: Yeah I split it that way for clarity's sake, but the selector with `:not()` is iffy

cheap-glitch:
Tested everything again on Firefox with the latest changes, it's :100: :ok:

PR #4526, opened 2021-06-29
Add `select-notifications` feature

### LINKED ISSUE #3229: Notifications "select all" by status/type
Hello!

I've been a user of Refined Github (RG) for a while now and it's become a must-have for my development and project management workflow. The features offered by RG pretty much satisfy all my needs, except in the case of Github Notifications management. My notification feed usually contains thousands of items from tens of different repositories, issues and PRs mixed in. Github offers a **Select all** feature which allows you to grab items in bulk as well as a very limited set of **filters** that can be used to coerce lists of notifications into a set that you can "select all" on and bulk update / review manually. However, there is no set of filters that allow you to select items by status. For example, the flow of "I want to mark all notifications pertaining to a merged PR as read" has no filter shortcut and can be quite tedious.

## Proposed solution

I suggest a dropdown button component not unlike the "..." button that hides the "Mark as read/unread", but for selecting in bulk based on item type and status. With some tinkering, I've put together a mockup of what it could look like:

image

This would sit on the top bar of the [notifications screen](https://github.com/notifications), alongside the "Select all" checkbox so that users can quickly bulk-select items with some granularity. I've also put together a crude [CodePen](https://codepen.io/marccataford/pen/LYGZRMe) based off whatever styles and components are already available on the page to give an idea of what the menu would feel like in the UI.

The filtering would only operate on elements that are already in the DOM and would not require to do any sort of network request; each row is already aware of the information that would be used as filter (based on other information displayed in each row already).

I think that this feature would be tremendously helpful for folks that, like me, are flooded by notifications and want to quickly triage through their queue. It could easily be made accessible and even be associated with hotkeys if there's desire to (depending on what the stance of RG is on tacking hotkeys onto the UI via an extension).

I would be happy to take on some of that work and put a PR together if there's any interest in this! It's certainly something that I would use every day and I'm sure others would too. I'd be more than happy to polish this further if this proposal is missing bits and pieces.

cheap-glitch:
Fixes #3229 

## Test URLs

## Screenshot
https://user-images.githubusercontent.com/46634000/124293072-8df13e00-db56-11eb-952a-5c9e1bd11b24.mp4


source/features/notifications-filter.tsx
```diff
@@ -0,0 +1,140 @@
+import React from 'dom-chef';
+import select from 'select-dom';
+import * as pageDetect from 'github-url-detection';
+import {CheckIcon, CheckCircleIcon, XCircleIcon, IssueOpenedIcon, GitPullRequestIcon, GitPullRequestDraftIcon, GitMergeIcon, DotIcon, DotFillIcon} from '@primer/octicons-react';
+
+import features from '.';
+
+const filters = {
+	PR: '[href*="/pull/"]',
+	Issue: '[href*="/issues/"]',
```
cheap-glitch, line 10: Not sure if those selectors are robust enough. Is it possible for `/pull/` or `/issues/` to appear in a URL that isn't an issue/PR?

source/features/notifications-filter.tsx
```diff
@@ -0,0 +1,140 @@
...
+				aria-haspopup="menu"
+				role="button"
+			>
+				Filter by <span className="dropdown-caret ml-1"/>
+			</summary>
+			<details-menu
+				className="SelectMenu left-0"
+				aria-label="Filter by"
+				role="menu"
+				on-details-menu-selected={handleSelection}
```
cheap-glitch, line 114: Using `on-details-menu-selected` instead of `on-details-menu-select` means we don't have to do `await delay(1)` in the event handler (this event is sent _after_ the dropdown DOM is updated).

I think this could be applied to `conversation-activity-filter` (untested).

source/features/notifications-filter.tsx
```diff
@@ -0,0 +1,140 @@
...
+					{createDropdownList('Read', ['Read', 'Unread'])}
+				</div>
+			</details-menu>
+		</details>
+	);
+}
+
+function init(): false | void {
+	const notificationFilterWrapper = select('notification-filter')!.parentElement!;
+	notificationFilterWrapper.nextElementSibling?.classList.replace('ml-3', 'ml-2');
```
cheap-glitch, line 128: This reduces the spacing between the inserted button and the "Group by date/repo" button (which is only here on the "Inbox" subpage).

source/features/notifications-filter.tsx
```diff
@@ -0,0 +1,140 @@
...
+function createDropdown(): JSX.Element {
+	return (
+		<details className="details-reset details-overlay position-relative d-none d-md-block">
+			<summary
+				className="btn ml-3"
+				data-hotkey="f"
+				aria-haspopup="menu"
+				role="button"
+			>
+				Filter by <span className="dropdown-caret ml-1"/>
```
cheap-glitch, line 108: Not sure about the dropdown button text (and the feature name). "Filter" tends to imply that _all_ notifications are filtered and then displayed, but here it's only the already displayed notifications that are filtered.

This is particularly apparent in the "Done" subpage, maybe the feature simply should run there?

fregante:
😱 the video looks great! Does “select all” select only the visible items? That’s the expectation for me

source/features/notifications-filter.tsx
fregante: Yes it’s possible: 
GitHub.com/fregante/issues (isRepoHome)
GitHub.com/fregante/issues/pulls (isPRList)

cheap-glitch:
> Does “select all” select only the visible items?

No, forgot to take that into account. Will look into it.

Other possible improvements, not yet implemented because I don't know how useful they'd really be:
- a reset button to clear the current filters
- preserve the filters when navigating between the current subpages (Inbox, Saved & Done)
- more filters in "Type" category: Commit, CI checks, Mention, others?

source/features/notifications-filter.tsx
cheap-glitch: Both these URLs lead to a 404 page for me (tried with my username too) :thinking:

readme.md
```diff
@@ -355,6 +355,7 @@ Thanks for contributing! 🦋🙌
 - [](# "stop-redirecting-in-notification-bar") [Stops redirecting to notification inbox from notification bar actions while holding <kbd>Alt</kbd>.](https://user-images.githubusercontent.com/202916/80318782-c38cef80-880c-11ea-9226-72c585f42a51.png)
 - [](# "select-all-notifications-shortcut") Adds a shortcut to select all visible notifications: <kbd>a</kbd>.
 - [](# "global-search-filters") [Adds search filters in global search page: forks, private repos, own repos, authored commits.](https://user-images.githubusercontent.com/44045911/112261285-88a66c80-8ca6-11eb-82cb-933b72c57abd.png)
+- [](# "notifications-filter") [Filter displayed notifications by type and status.](TODO)
```
cheap-glitch, line 358: Should we create a "Notifications" category? There's already several features that could fit under it.

source/features/notifications-filter.tsx
```diff
@@ -0,0 +1,146 @@
...
+			</summary>
+			<details-menu
+				className="SelectMenu left-0"
+				aria-label="Filter by"
+				role="menu"
+				on-details-menu-selected={handleSelection}
+			>
+				<div className="SelectMenu-modal">
+					{createDropdownList('Type', ['PR', 'Issue'])}
```
kidonng, line 123: First of all, this feature is awesome 🔥

However, these item texts should be more inline with GitHub (`Pull requests` and `Issues`) I think? You can compare it to the existing discussions (issues/PR) filter.

source/features/notifications-filter.tsx
fregante: heheh they were examples. I created the repo now so you can see

readme.md
fregante: Yes please! Every time a new category is possible (4/5+ features) we should probably create it. Global and Repository are quite big categories

fregante:
This feature is probably going to be a bit long/difficult. Good luck! 😃 

The feature request was about selecting notifications rather than filtering, which to be fair is less convenient than filtering, but also easier to implement. I think this feature currently does not support notifications grouped by Repository

Screen Shot 6

The drawback of this nice feature is that people will expect/hope that it supports multiple pages of notifications, especially when only a couple of notifications remain visible after filtering. 👈 just a note, I don’t think we can do much about this other than implementing infinite scroll (whole 'nother feature, which could be discussed separately, but I'm not excited about)

fregante:
> Other possible improvements, not yet implemented because I don't know how useful they'd really be:

While we should keep them in mind to maybe find better ways to do it, it's probably best to avoid them for now because you don't want to make it infinitely-complex.

source/features/notifications-filter.tsx
```diff
@@ -0,0 +1,146 @@
...
+		<div className="SelectMenu-list">
+			<header className="SelectMenu-header">
+				<span className="SelectMenu-title">{category}</span>
+			</header>
+			{filters.map(filter => (
+				<label
+					className="SelectMenu-item text-normal"
+					role="menuitemcheckbox"
+					aria-checked="false"
+					tabIndex={0}
```
fregante, line 89: I think this is the default?

source/features/notifications-filter.tsx
```diff
@@ -0,0 +1,146 @@
+import React from 'dom-chef';
+import select from 'select-dom';
+import * as pageDetect from 'github-url-detection';
+import {CheckIcon, CheckCircleIcon, XCircleIcon, IssueOpenedIcon, GitPullRequestIcon, GitPullRequestDraftIcon, GitMergeIcon, DotIcon, DotFillIcon} from '@primer/octicons-react';
```
fregante, line 4:
```suggestion
import {
	CheckIcon,
	CheckCircleIcon, 
	etc
```

source/features/notifications-filter.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+const filters = {
+	PR: '[href*="/pull/"]',
+	Issue: '[href*="/issues/"]',
+	Open: ':is(.octicon-issue-opened, .octicon-git-pull-request)',
+	Closed: ':is(.octicon-issue-closed, .octicon-git-pull-request-closed)',
+	Draft: '.octicon-git-pull-request-draft',
+	Merged: '.octicon-git-merge',
+	Read: 'notification-read',
+	Unread: 'notification-unread'
```
fregante, line 16: Are these selectors or just class names? Instead of using `.classList.contains()` just use `.matches()` so you can turn this into regular selectors like the rest of them.

source/features/notifications-filter.tsx
```diff
@@ -0,0 +1,146 @@
...
+};
+
+type Filter = keyof typeof filters;
+type Category = 'Type' | 'Status' | 'Read';
+
+function handleSelection({target}: Event): void {
+	const activeFilters: Record<Category, string[]> = {
```
fregante, line 23: Do you need the category here? `is:read` should be identical to `is:pr` internally, right? They're all just matched against the selectors in `filters`.

Could this whole function be shortened to something like:

```js
// Merge all the filters into a single selector
const filteredNotificationSelector = select
	.all('[aria-checked="true"]', target as Element)
	.map(el => el.value)
	.join(''); 


for (const notification of select.all('.notifications-list-item')) {
	const hidden = notification.matches(filteredNotificationSelector);
	notification.hidden = hidden;
	select('.js-notification-bulk-action-check-item').toggleAttribute('data-check-all-item', hidden)
}

// Hide empty notifications groups
for (const group of select.all('.js-notifications-group')) {
	group.hidden = !select.exists('.notifications-list-item:not([hidden])', group);
}
```

source/features/notifications-filter.tsx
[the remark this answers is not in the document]
cheap-glitch: It's needed to allow proper keyboard support (otherwise the selection will not enter the dropdown when pressing <kbd>Tab</kbd>).

source/features/notifications-filter.tsx
```diff
@@ -1,7 +1,17 @@
 import React from 'dom-chef';
 import select from 'select-dom';
 import * as pageDetect from 'github-url-detection';
-import {CheckIcon, CheckCircleIcon, XCircleIcon, IssueOpenedIcon, GitPullRequestIcon, GitPullRequestDraftIcon, GitMergeIcon, DotIcon, DotFillIcon} from '@primer/octicons-react';
+import {
+	CheckIcon,
```
yakov116, line 5: Nitpick can  you sort it from a-z. I think the list will be easier to read. @fregante?

source/features/notifications-filter.tsx
```diff
@@ -0,0 +1,146 @@
...
+		}
+	}
+
+	// Hide empty notifications groups
+	for (const group of select.all('.js-notifications-group')) {
+		group.hidden = !select.exists('.notifications-list-item:not([hidden])', group);
+	}
+}
+
+const icons: Record<Filter, JSX.Element> = {
```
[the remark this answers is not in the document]
cheap-glitch: > Alternative solutions welcome

Defining the type of `icons` as `{ [Key in Filter]: JSX.Element }` should do the trick

cheap-glitch:
> The feature request was about selecting notifications rather than filtering, which to be fair is less convenient than filtering, but also easier to implement.

Don't know why I thought this was about filtering, when the original issue is clearly talking about selection :expressionless:
 
> The drawback of this nice feature is that people will expect/hope that it supports multiple pages of notifications, especially when only a couple of notifications remain visible after filtering.

Yep that's what I thought too, and I don't have a solution

> point_left just a note, I don’t think we can do much about this other than implementing infinite scroll (whole 'nother feature, which could be discussed separately, but I'm not excited about)

:+1: We could also alleviate the possible confusion with a little message at the bottom of the dropdown, and maybe by limiting the feature to the "Inbox" sub-page.

But really I think the best "solution" here is to give up on filtering and just implement the selection part — especially since that's what was aksed for in the first place. (And I would have done that had I read the issue properly :roll_eyes:)

source/features/notifications-filter.tsx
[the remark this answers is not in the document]
fregante: That's one-way (i.e. if you add more filters it will still pass without an icon) but I think it's good enough

fregante:
> > point_left just a note, I don’t think we can do much about this other than implementing infinite scroll (whole 'nother feature, which could be discussed separately, but I'm not excited about)
> 
> 👍 We could also alleviate the possible confusion with a little message at the bottom of the dropdown, and maybe by limiting the feature to the "Inbox" sub-page.

Maybe the pagination at the bottom of the page makes it clear that there are multiple pages. Is the filter preserved when navigating to those extra pages? 👈 to be clear it should not be preserved indefinitely, but just for the current tab.

Otherwise, yeah, using this as a "select" instead of filter does away with a lot of issues, especially if we make the select dropdown immediately "reset" when closed — i.e. keep the notifications selected, but start from "zero filters" every time you open the dropdown.

source/features/notifications-filter.tsx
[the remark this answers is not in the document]
cheap-glitch: > That's one-way

It does work both ways, that's the beauty of mapped types :wink:

source/features/notifications-filter.tsx
cheap-glitch: Dammit, thought I could get away with a terse selector :disappointed:\
Guess I have to use a long and ugly `:is()` list now

source/features/notifications-filter.tsx
cheap-glitch: The filtering logic I've chosen to implement is as follow:
- filters in the same category are combined with OR
- categories are combined with AND

I don't think this can be expressed with a single selector. (Plus I've got to use `.matches()` for the Read/Unread filters since `querySelector()` doesn't look at the root.)

cheap-glitch:
> Maybe the pagination at the bottom of the page makes it clear that there are multiple pages.

True, it does, but that makes it even more frustrating, because imo it's not the expected behavior for a filter. And no, its state is not preserved because the whole page is reloaded when moving between tabs & pages.

:point_right: **I think I'll change this to select instead of filter.** If people end up wanting a filter too I can always re-add it.

> especially if we make the select dropdown immediately "reset" when closed 

:+1:

source/features/notifications-filter.tsx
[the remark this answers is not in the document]
fregante: Oh, indeed. In some cases it works as "any key in B is allowed, not expected".

source/features/select-notifications.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+function init(): false | void {
+	select('.js-notifications-mark-all-prompt')!
+		.closest('label')!
+		.after(createDropdown());
+}
+
+void features.add(__filebasename, {
+	shortcuts: {
+		S: 'Open the "Select by" dropdown'
```
cheap-glitch, line 164: Not sure about this shortcut (can't use <kbd>s</kbd> because it's for the header searchbar). Alternatively we could go with <kbd>n</kbd> maybe.

source/features/select-notifications.tsx
```diff
@@ -0,0 +1,176 @@
...
+			// Make excluded notifications unselectable
+			if (!isNotificationSelected) {
+				select('.js-notification-bulk-action-check-item', notification)!.removeAttribute('data-check-all-item');
+			}
+		}
+
+		// If at least one notification is selectable, trigger the "Select all" checkbox
+		if (select.exists('.js-notification-bulk-action-check-item[data-check-all-item]')) {
+			selectAllCheckbox.dispatchEvent(new MouseEvent('click'));
+		}
```
cheap-glitch, line 84: I used this roundabout way of selecting the notifications because:
1. We can't just set the `checked` attribute of their checkboxes (it doesn't update the "Select all" count).
2. Dispatching a `MouseEvent` to every selected notification causes some noticeable lag, when a lot of notifications are selected at once.

readme.md
```diff
@@ -340,6 +340,16 @@ Thanks for contributing! 🦋🙌
 
 <!-- Refer to style guide above. Keep this message between sections. -->
 
+### Notifications
+
+- [](# "open-all-notifications") [Adds a button to open all your unread notifications at once.](https://user-images.githubusercontent.com/1402241/80861295-fbad8b80-8c6d-11ea-87a4-8025fbc3a3f4.png)
+- [](# "select-all-notifications-shortcut") Adds a shortcut to select all visible notifications: <kbd>a</kbd>.
+- [](# "linkify-notification-repository-header") [Linkifies the header of each notification group (when grouped by repository).](https://user-images.githubusercontent.com/1402241/80849887-81531c00-8c19-11ea-8777-7294ce318630.png)
+- [](# "stop-redirecting-in-notification-bar") [Stops redirecting to notification inbox from notification bar actions while holding <kbd>Alt</kbd>.](https://user-images.githubusercontent.com/202916/80318782-c38cef80-880c-11ea-9226-72c585f42a51.png)
+- [](# "select-notifications") [Select notifications by type and status.](https://user-images.githubusercontent.com/46634000/124293072-8df13e00-db56-11eb-952a-5c9e1bd11b24.mp4)
```
cheap-glitch, line 349: :fire: ?

source/features/select-notifications.tsx
```diff
@@ -0,0 +1,176 @@
...
+					isNotificationSelected = false;
+					break;
+				}
+			}
+
+			// Make excluded notifications unselectable
+			if (!isNotificationSelected) {
+				select('.js-notification-bulk-action-check-item', notification)!.removeAttribute('data-check-all-item');
+			}
+		}
```
fregante, line 79: I still think this loop can/should be simplified.

Since this is ignoring unselected items, just query the items you're looking for and loop them specifically. I think that even if they're 3 separate loops they're more readable and less error-prone than several nested loops and conditions.

Improvable pseudo-code:

```js
const types = new FormData(formElement).getAll('Type').map(type => filters[type]).join('') // '' means "AND"
for (const type of types) {
	// select it
}

const status = new FormData(formElement).getAll('Status').map(type => filters[type]).join(',') // ',' means "OR"
for (const status of status) {
	// select it
}

// etc
```

source/features/select-notifications.tsx
```diff
@@ -0,0 +1,176 @@
...
+};
+
+type Filter = keyof typeof filters;
+type Category = 'Type' | 'Status' | 'Read';
+
+function resetFilters({target}: Event): void {
+	for (const checkbox of select.all('input[type="checkbox"]:checked', target as Element)) {
+		checkbox.checked = false;
+		checkbox.parentElement!.parentElement!.setAttribute('aria-checked', 'false');
+	}
```
fregante, line 36: Is there a "native" way to do this? I think that the GitHub component would try to keep these 2 attributes in sync. Maybe calling `form.reset()`?

source/features/select-notifications.tsx
fregante: Would it be possible to set the `checked` property and then trigger the count once at the end?

source/features/select-notifications.tsx
```diff
@@ -0,0 +1,176 @@
...
+				Select by <span className="dropdown-caret ml-1"/>
+			</summary>
+			<details-menu
+				className="SelectMenu left-0"
+				aria-label="Select by"
+				role="menu"
+				on-details-menu-selected={handleSelection}
+			>
+				<div className="SelectMenu-modal">
+					{createDropdownList('Type', ['Pull requests', 'Issues'])}
```
fregante, line 153: Do we need to support discussions too?

source/features/select-notifications.tsx
cheap-glitch: That was my first thought, sadly it doesn't work (every checkbox ends up selected regardless).

source/features/select-notifications.tsx
cheap-glitch: I don't know, I've never used them. Do they generate a lot of noise?

There are more types we could support too, like Commits, CI results, Security warnings... I think we can leave it like this and see what people think. If they ask support for more categories it'll be easy enough to add.

source/features/select-notifications.tsx
cheap-glitch: > I think that the GitHub component would try to keep these 2 attributes in sync

Looking at the custom element's code, this only happens when a checkbox is checked/unchecked. I think it's faster to call `reset()` on the form and reset all the `aria-checked` attributes.

source/features/select-notifications.tsx
fregante: > there's no way to express that in a selector.

Why not? Isn't technically each filter applied on top of the previous selection?

1. Select PRs
2. Out of the selected PRs, select closed or merged
3. Out of the selected closed or merged PRs, select unread

fregante: The current code looks good though@

source/features/select-notifications.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+	if (select.exists(':checked', target as Element)) {
+		const formData = new FormData(select('form#rgh-select-notifications-form'));
+		const types = formData.getAll('Type').map(type => filters[type as Filter]);
+		const statuses = formData.getAll('Status').map(status => filters[status as Filter]);
+		const readStatus = formData.getAll('Read').map(read => filters[read as Filter]);
+
+		for (const notification of select.all('.notifications-list-item')) {
+			if (
+				(types.length > 0 && !select.exists(types, notification)) ||
```
fregante, line 54: Let's dry it up a bit

```js
function getSelector(formData, type): string {
	return formData.getAll(type).map(type => filters[type as Filter]).join(',')
}
```

```suggestion
		const readStatus = getSelector(formData, 'Read');

		for (const notification of select.all('.notifications-list-item')) {
			if (
				(types && !select.exists(types, notification)) ||
```

fregante (approved):
I think after that it looks good! I trust it works 😃

cheap-glitch:
> I trust it works

Tested on Firefox with the latest changes :+1: 

One last thing:

> Not sure about the shortcut (can't use `s` because it's for the header searchbar). Alternatively we could go with `n` maybe.

fregante:
> Not sure about the shortcut (can't use `s` because it's for the header searchbar). Alternatively we could go with `n` maybe.

🤷‍♂️ no preference

leobenkel:
when will that be available ?

zot:
Was this feature removed?
image

codezninja:
Based on this [comment](https://github.com/refined-github/refined-github/issues/3229#issuecomment-1217605325) I seem to have lost this feature recently and not sure when it broke.

yakov116:
@codezninja  its working fine for me.

If its not working for you make a new issue WITH pictures of the feature missing

PR #4528, opened 2021-06-29
Disable `avoid-accidental-submissions` when editing the title of existing PRs

### LINKED ISSUE #4525: `avoid-accidental-submissions` breaks PR title changes with Enter
1. Visit ~~an issue~~ _a PR_ you can edit, like https://github.com/sindresorhus/refined-github/pull/4528
2. Edit the title
3. Press enter
4. The changes are not saved, nothing happens

This is what the feature is supposed to do, but I think we should exclude issue/pr titles.

demo

darkred:
Closes #4525 

## Test URLs

- Existing issue: https://github.com/sindresorhus/refined-github/pull/4525
- Existing PRs (of any state): https://github.com/sindresorhus/refined-github/pull/4528

## Screenshot

None

### DIFF
--- a/source/features/avoid-accidental-submissions.tsx
+++ b/source/features/avoid-accidental-submissions.tsx
@@ -46,7 +46,6 @@ void features.add(__filebasename, {
 		pageDetect.isPRConversation
 	],
 	exclude: [
-		pageDetect.isIssue,
 		pageDetect.isOpenPR,
 		pageDetect.isDraftPR,
 		pageDetect.isClosedPR
--- a/source/features/avoid-accidental-submissions.tsx
+++ b/source/features/avoid-accidental-submissions.tsx
@@ -29,17 +29,12 @@ function onKeyDown(event: delegate.Event<KeyboardEvent, HTMLInputElement>): void
 }
 
 function init(): void {
-	const elementsToDelegateKeydown = [
+	delegate(document, [
+		'form.new_issue input#issue_title',
 		'input#pull_request_title',
 		'input#commit-summary-input',
 		'#merge_title_field'
-	];
-
-	if (!pageDetect.isPRConversation()) { // In all cases other than isPRConversation, disallow title edit via Enter
-		elementsToDelegateKeydown.push('input#issue_title');
-	}
-
-	delegate(document, elementsToDelegateKeydown.join(), 'keydown', onKeyDown);
+	].join(), 'keydown', onKeyDown);
 }
 
 void features.add(__filebasename, {
--- a/source/features/avoid-accidental-submissions.tsx
+++ b/source/features/avoid-accidental-submissions.tsx
@@ -29,12 +29,17 @@ function onKeyDown(event: delegate.Event<KeyboardEvent, HTMLInputElement>): void
 }
 
 function init(): void {
-	delegate(document, [
-		'input#issue_title',
+	const elementsToDelegateKeydown = [
 		'input#pull_request_title',
 		'input#commit-summary-input',
 		'#merge_title_field'
-	].join(), 'keydown', onKeyDown);
+	];
+
+	if (!pageDetect.isPRConversation()) { // In all cases other than isPRConversation, disallow title edit via Enter
+		elementsToDelegateKeydown.push('input#issue_title');
+	}
+
+	delegate(document, elementsToDelegateKeydown.join(), 'keydown', onKeyDown);
 }
 
 void features.add(__filebasename, {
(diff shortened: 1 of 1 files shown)


yakov116:
> I additionally removed `pageDetect.isEditingFile` as it's probably not necessary.

Why not? It works in the submission field

yakov116:
I also think that what @fregante meant was the title field not the whole feature

darkred:
I thought that the idea of #4525 was not getting in the way of modifying *anything existing*,
e.g. when renaming/modyfing an existing file.

yakov116:
No the title says "breaks issue title changes with Enter"

darkred:
Yes, I noticed, of course, it's just that I thought to expand based on it. 
Anyway, I restored it.

yakov116:
I still think this breaks the whole feature.

darkred:
So you are suggesting that I just should close this PR?

darkred:
I'm closing this.
Sorry for the noise - I thought this was simple to fix, and really hoped I could contribute,
especially because the addition of the `avoid-accidental-submissions` feature was my suggestion.

darkred:
I believe that the #4525 report is not correct regarding Issues:
the way `avoid-accidental-submissions` currently works 
*(tested in a fresh Chrome profile with only RG having login to GitHub, in my own test repo)*,
it breaks title changes with Enter **only for PRs**, not for Issues too. For issues it works ok (see [screen capture](https://user-images.githubusercontent.com/723651/123872067-3461f700-d93d-11eb-8fb4-e9865d1be94b.gif)).
That's because, as you can see the current `include` array, it doesn't match existing Issues, only new ones:
https://github.com/sindresorhus/refined-github/blob/55a5ddaff84e13a4d7ffce444ef5028958550ec7/source/features/avoid-accidental-submissions.tsx#L41-L47

---

In my PR I suggested adding this

```js
	exclude: [
		pageDetect.isIssue,
		pageDetect.isOpenPR,
		pageDetect.isDraftPR,
		pageDetect.isClosedPR
	],
```

of which only the `pageDetect.isIssue` has no meaning, because `isNewIssue` matches don't include `isIssue`:

<details>
<summary><code>fregante/github-url-detection</code> results screenshot:</summary>


</details>

I believe that the rest 3 items (`isOpenPR, isDraftPR, isClosed`) are still needed
in order to avoid running on existing PRs, and thus allow editing via Enter.

---

I still believe this PR fixes this bug.
Am I totally wrong? 
If yes, could you please elaborate on why my PR breaks this feature? And, what the proper solution to this bug would be?

darkred:
I think I finally got it right now:

with this PR we want to allow title edits to *existing* PRs.

The issue was with the `input#issue_title` selector: that's the title selector for existing PRs (and new issues).
*In contrast, for new PRs (via compare branch e.g. https://github.com/darkred/test/compare/main...t2?expand=1
the title selector is `input#pull_request_title`.*

So, I turned the selectors array to delegate into its own `elementsToDelegateKeydown` variable,
and I add the `input#issue_title` element in it (to disallow editing via Enter) conditionally, i.e. only for non-PRconversations pages.
So, I removed the unneeded `exclude` array.

fregante:
Yep that’s about right!

However, instead of an explicit page check, I’d rather just have a more specific selector. What is that `#issuetitle` exactly? Let’s target it more specifically, like “form.etc.cetera #issue_title`

darkred:
Thanks, much more elegant solution!
So, I changed the selector `input#issue_title` into `form.new_issue input#issue_title`
and now it applies works only in new issues - not when editing the title of existing PRs.

yakov116:
@fregante this looks good to me but I am unable to test. Think we can still ship it in the next 18 min?

fregante:
Too late! Maybe for v21.8.1 😁

fregante (approved):
Untested but also LGTM. 

Needs testing to ensure it blocks Enter on the new issue page and that it allows Enter when changing the titles in PRs

yakov116:
Tested! Works!

darkred:
> Needs testing to ensure

> Tested! Works!

Just wanted to clarify that I tested it before committing -- although I understand of course that you need to doublecheck before merging.

---

Off-topic, but could you please me clarify something?

I initially stepped in with making this PR, wanting to help, but my initial approach was irrelevant.
Then you helped me understand the problem, and so I came up with a working, but unnecessary complicated approach. 
Finally, you suggested the optimal approach: just make the relevant existing selector more specific, and that would be all.
That's what I did, and so the resulting PR (5 commits later..) is 1-line change, merely the addition of a few characters, 
instead of changes spanning to multiple lines.

So, my question is: is it ok with you that I started this PR with the *belief/hope* that (probably with your help) I could eventually make it complete (based on that the issue seemed simple enough), even though my initial approach was too simplistic and proved wrong?
Or, should I shouldn't start this PR at all, and let a more experienced contributor solve it with just a 1-commit PR ?

fregante:
Regarding the off-topic, reading the issue and understanding the proposed solution correctly helps. Sometimes the issues I open are rather terse so they might be understood if one is not fully familiar with the codebase.

PR #4534, opened 2021-07-01
Add `remove-unused-repo-tabs` feature

### LINKED ISSUE #4246: Wiki: Show number of pages.

yakov116:
Closes #4246
What this feature does it if there are actions/wiki's/projects it leaves the tabs. if not it moves it to the `more-dropdown`.

it also adds the tab count (actions the workflow count and wiki's the page count)

Test on
### Has actions
https://github.com/sindresorhus/eslint-plugin-unicorn 

### Has actions and wiki
https://github.com/pixiebrix/pixiebrix-extension

### has Projects
https://github.com/twbs/bootstrap

### Has no projects/actions/wiki's
https://github.com/twschiller/dslfin-www

### when going to the tab it does not get hidden
https://github.com/twschiller/dslfin-www/projects

### Org
https://github.com/tc39 //No projects
https://github.com/microsoft // has projects

image


source/features/next-scheduled-github-action.tsx
```diff
@@ -28,6 +29,19 @@ const getScheduledWorkflows = cache.function(async (): Promise<Record<string, st
 
 	const workflows = workflowFiles?.entries;
 	if (!workflows) {
+		return [];
+	}
+
+	return workflows;
```
fregante, line 35: I think

```suggestion
	return workflowFiles?.entries ?? [];
```

source/features/remove-unused-repo-tabs.tsx
```diff
@@ -0,0 +1,111 @@
...
+	tabCounter.title = count > 999 ? String(count) : '';
+}
+
+const getWikiPageCount = cache.function(async (): Promise<number> => {
+	const wikiPages = await fetchDom(buildRepoURL('wiki'), '#wiki-pages-box .Counter');
+	if (!wikiPages) {
+		return 0;
+	}
+
+	return looseParseInt(wikiPages);
```
fregante, line 42: If it doesn't already, `looseParseInt` should just return `0` when the input is undefined, so you can drop the condition above as well
```suggestion
	return looseParseInt(wikiPages);
```

source/features/remove-unused-repo-tabs.tsx
```diff
@@ -0,0 +1,111 @@
...
+}, {
+	maxAge: {hours: 1},
+	staleWhileRevalidate: {days: 5},
+	cacheKey: () => __filebasename + 'wiki:' + getRepo()!.nameWithOwner
```
fregante, line 46:
```suggestion
	cacheKey: () => 'wiki-page-count:' + getRepo()!.nameWithOwner
```

source/features/remove-unused-repo-tabs.tsx
```diff
@@ -0,0 +1,111 @@
...
+		!tab || // Tab disabled 🎉
+		tab.matches('.selected') ||// User is on tab 👀
+		// Repo/Organization owners should see the tab. If they don't need it, they should disable the feature altogether
+		pageDetect.canUserEditRepo() ||
+		pageDetect.canUserEditOrganization()
+	) {
+		return true;
+	}
+
+	return false;
```
fregante, line 27:
```suggestion
	return (
		!tab || // Tab disabled 🎉
		tab.matches('.selected') ||// User is on tab 👀
		// Repo/Organization owners should see the tab. If they don't need it, they should disable the feature altogether
		pageDetect.canUserEditRepo() ||
		pageDetect.canUserEditOrganization()
	);
```

fregante (commented):
> Are we keeping the "new project dropdown"?

Probably not, let's drop it. We have `onlyShowInDropdown` now.



> If yes are we adding new actions or new wiki to it.

Nope, `onlyShowInDropdown` is a good solution.



> What should the screenshot be of?

Before: show the tabs of a repo that has 0 projects and 0 packages, but maybe some wikis
After: hide projects and packages, ad count to wiki, show open dropdown menu with projects and packages

When taking the screenshot of the open dropdown:
- drop other unrelated items to make it easier to focus on the 2 items
- add margin below the tab bar so it doesn't overlap over content, like (old screenshot):

source/features/remove-unused-repo-tabs.tsx
```diff
@@ -0,0 +1,113 @@
...
+		// https://github.com/orgs/USER/projects/new
+		const path = '/orgs' + location.pathname.split('/', 3).join('/');
+		select('.Header [href="/new"]')!.parentElement!.append(
+			<a className="dropdown-item" href={path}>
+				New project
+			</a>
+		);
+		projectsTab!.remove();
+		return;
+	}
```
yakov116, line 88: Should we just drop this altogether?

fregante: I just answered thay

yakov116: Sorry I was not clear. My question was should we drop the whole feature on `pageDetect.isOrganizationProfile()`

fregante: I think we should only hide the tabs if we're not owners/members, so this dropdown item should not exist in the first place.

Ownership is easy to detect through the Settings tab. Membership can be detected via the "New X" button on some pages:

Screen Shot 7


Screen Shot 8


The logic should be something like:

```
if we're owner or members
	return
else
	hide tabs
```

This exact logic means that if we can't determine whether we're owners/members from the DOM, the tab will be hidden. This is fine.

yakov116:
###Before

### After
image

Good?

yakov116:
In case you want to know what repo I used https://github.com/harvester/harvester

fregante:
The second screenshot might be a good candidate for `more-dropdown` as is. 👍 That screenshot was old

yakov116:
Migration needed?

source/features/remove-unused-repo-tabs.tsx
```diff
@@ -0,0 +1,109 @@
...
+		setTabCounter(actionsTab, actionsCount);
+	} else {
+		onlyShowInDropdown('actions-tab');
+	}
+}
+
+async function initProjects(): Promise<void | false> {
+	const projectsTab = await elementReady([
+		'[data-hotkey="g b"]', // In organizations and repos
+		'[aria-label="User profile"] [href$="?tab=projects"]' // In user profiles
```
yakov116, line 71: // In user profiles

I forgot to check this one...

yakov116: Have any users to test on?

yakov116: Droped it, it is not DomReady safe

source/features/next-scheduled-github-action.tsx
```diff
@@ -26,8 +27,16 @@ const getScheduledWorkflows = cache.function(async (): Promise<Record<string, st
...
+	maxAge: {days: 1},
+	staleWhileRevalidate: {days: 10},
+	cacheKey: () => 'workflows:' + getRepo()!.nameWithOwner
+});
+
+const getScheduledWorkflows = async (): Promise<Record<string, string> | false> => {
+	const workflows = await getWorkflows();
+	if (workflows.length === 0) {
 		return false;
 	}
```
fregante, line 41: You can drop this condition an just return `{}` instead of `false`. The only problem is that it complicates the later check from `!workflows` to `Object.keys(workflows).length === 0` or something like that. Thoughts?

```suggestion
```

source/features/remove-unused-repo-tabs.tsx
```diff
@@ -0,0 +1,105 @@
...
+import features from '.';
+import fetchDom from '../helpers/fetch-dom';
+import getTabCount from '../github-helpers/get-tab-count';
+import looseParseInt from '../helpers/loose-parse-int';
+import {getWorkflows} from './next-scheduled-github-action';
+import abbreviateNumber from '../helpers/abbreviate-number';
+import {onlyShowInDropdown} from './more-dropdown';
+import {buildRepoURL, getRepo} from '../github-helpers';
+
+function tabCannotBeHidden(tab: HTMLElement | undefined): boolean {
```
fregante, line 15: Let's _not_ use negative items. `true` here would mean `false`
```suggestion
function canTabBeHidden(tab: HTMLElement | undefined): boolean {
```

or
```suggestion
function mustKeepTab(tab: HTMLElement | undefined): boolean {
```

source/features/remove-unused-repo-tabs.tsx
```diff
@@ -0,0 +1,105 @@
...
+});
+
+async function initWiki(): Promise<void | false> {
+	const wikiTab = await elementReady('[data-hotkey="g w"]');
+	if (!wikiTab) {
+		return false;
+	}
+
+	const wikiPageCount = await getWikiPageCount();
+	if (tabCannotBeHidden(wikiTab) || wikiPageCount > 0) {
```
fregante, line 47: `> 0` is faster on my eyes and on the CPU
```suggestion
	if (wikiPageCount > 0 || tabCannotBeHidden(wikiTab)) {
```

source/features/remove-unused-repo-tabs.tsx
```diff
@@ -0,0 +1,105 @@
...
+	if (pageDetect.isOrganizationProfile()) {
+		if (await elementReady('a.btn-primary[href$="repositories/new"]')) {
+			return;
+		}
+
+		projectsTab!.remove();
+		return;
+	}
+
+	onlyShowInDropdown('projects-tab');
```
fregante, line 83: Can we flip this?
```suggestion
	if (pageDetect.is other page()) {
		onlyShowInDropdown('projects-tab');
		return;
	}
	
	if (await elementReady('a.btn-primary[href$="repositories/new"]')) {
		return;
	}

	projectsTab!.remove();
```

source/features/next-scheduled-github-action.tsx
yakov116: Exactly why I changed it. If you look at the last PR you would see it caused me issues.

source/features/remove-unused-repo-tabs.tsx
yakov116: How is it now?

readme.md
```diff
@@ -152,6 +150,7 @@ Thanks for contributing! 🦋🙌
 - [](# "next-scheduled-github-action") [Shows the next scheduled time of relevant GitHub Actions in the workflows sidebar.](https://user-images.githubusercontent.com/46634000/94690232-2476a180-0330-11eb-99d7-e174bb762cea.png)
 - [](# "quick-repo-deletion") [Lets you delete your repos in a click, if they have no stars, issues, or PRs.](https://user-images.githubusercontent.com/1402241/99716945-54a80a00-2a6e-11eb-9107-f3517a6ab1bc.gif)
 - [](# "useful-forks") [Helps you find the most active forks](https://user-images.githubusercontent.com/38117856/107463541-542e8500-6b2c-11eb-8b25-082f344c1587.png), via https://useful-forks.github.io.
+- [](# "remove-unused-repo-tabs") [Moves projects, actions and wiki's tabs that are enabled but are unused to the repository navigation dropdown.](https://user-images.githubusercontent.com/16872793/124681343-4a6c3c00-de96-11eb-9055-a8fc551e6eb8.png)
```
fregante, line 153:
```suggestion
- [](# "remove-unused-repo-tabs") [Moves projects, actions and wiki's tabs to the repository navigation dropdown if they're empty/unused.](https://user-images.githubusercontent.com/16872793/124681343-4a6c3c00-de96-11eb-9055-a8fc551e6eb8.png)
```

fregante (approved):
Untested but it looks pretty straightforward now 👌

source/features/remove-unused-repo-tabs.tsx
```diff
@@ -0,0 +1,106 @@
...
+		return false;
+	}
+
+	if (pageDetect.isRepo()) {
+		onlyShowInDropdown('projects-tab');
+		return;
+	}
+
+	// .isOrganizationProfile
+	if (await elementReady('a.btn-primary[href$="repositories/new"]')) {
```
fregante, line 80: I'd explain what this is though, the comment doesn't help me understand what it does and why we have it

yakov116:
```suggestion
	if (await elementReady('a.btn-primary[href$="repositories/new"]')) { // Check if the user is able to create a new repository
```

Good?

fregante: Yes but why are we checking that? (This need to be in the comment, not here)

fregante: In short, something like: “only remove the tab if you’re not a collaborator”

It may be easier to extract this into its own local function: canUserEditWhateverThisPageIs()

And then comment on its use as “Leave tab visible to allow the current user to create Projects”

fregante: Also there’s one issue specific to negative usage of elementReady: When used this way, you’re basically just waiting for domReady and then removing the tab. It’s still ok but it’d be best to use:

```js 
elementReady(element that always exists)?.querySelector(the button)
``` 

The difference is that this will resolve sooner when the button doesn’t exist

fregante: uhmmm there's already a `pageDetect.canUserEditOrganization`, so what is this line for exactly?

yakov116:
It's not DOM ready safe

source/features/remove-unused-repo-tabs.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+	if (pageDetect.isRepo()) {
+		onlyShowInDropdown('projects-tab');
+		return;
+	}
+
+	if (await canUserEditOrganization()) {
+		// Leave Project tab visible to those who can create a new project
+		return;
+	}
```
yakov116, line 85: Hmm I think we don't need this `elementReady('[data-hotkey="g b"]')` waited long enough to not have to wait for DOM ready.

However the `pageDetect.canUserEditOrganization` selector needs to be updated

yakov116:
Blocked by https://github.com/fregante/github-url-detection/pull/86

And then 🚀! ❤️ Please

fregante:
Merging this, but `canUserEditOrganization` is not the exclusion we need. The previous one has a different meaning from `GitHub-url-detection`:

```js
 async function canUserEditOrganization(): Promise<boolean> {
 	return Boolean(await elementReady('.btn-primary[href$="repositories/new"]'));
 }
```

This exists for any collaborator while the settings tab just points to the org owner, as far as I know. Actually I think `canUserEditOrganization` should be renamed to `isUserTheOrganizationOwner` in that package and here we need to restore the async function

More context in https://github.com/sindresorhus/refined-github/pull/4534#discussion_r665867480

Issue #4244, opened 2021-04-15
Extend "markdown" features to more applicable pages


fregante:
As discovered in https://github.com/sindresorhus/refined-github/pull/4151, we have some features that are applied only to Markdown files but could be applied to a whole host of file types like CSV and SVG.

Please analyze what makes a feature applicable to other filetypes and don't just blindly include random extensions.

Some of these features are:

- `link-to-changelog-file`
- `vertical-front-matter`
- `view-markdown-source` (we may want to revert #4535 to support more types of files)
- `hide-markdown-diff` -> if compatible with other files, it needs to be reimplemented because it was added by GitHub… only on markdown files.
- ~~`collapse-markdown-sections` should be changed to **only** apply to wikis, as discussed in https://github.com/sindresorhus/refined-github/issues/4008#issuecomment-820515283~~ done in #4335 

Some useful links:

- [GitHub Docs: Working with non-code files](https://docs.github.com/repositories/working-with-files/using-files/working-with-non-code-files) (you can find more types of file here)
- [GitHub Markup](https://github.com/github/markup/blob/46908504f5099b15a88030749f0aeb52cd2f6799/README.md#markups) (like the above link 👆 but more technical)
- [Test files](https://github.com/github/markup/tree/master/test/markups)
- [CSV Test files](https://github.com/maxogden/csv-spectrum/tree/master/csvs)

yakov116:
Are we able to use https://github.com/sindresorhus/is-md?

fregante:
The point is to enable these features on pages other than markdown, so no.

fregante:
Is this still applicable?

PR #4535, opened 2021-07-02
Use native "view source" URL in `view-markdown-source`

### LINKED ISSUE #4532: Replace `view-markdown-source` with a simple link
See https://github.blog/changelog/2021-06-30-parameter-to-disable-markdown-rendering/

I would keep one thing, if the URL has a hash, redirect to `?plain=true`. 

I am open to other options, since currently you don't have any change page.

mahdyar:
According to GitHub's Change log, Now there is a [Parameter to disable markdown rendering](https://github.blog/changelog/2021-06-30-parameter-to-disable-markdown-rendering/), so It'd be a better practice to use instead of extracting the source from blame.
Closes #4532 
## Test URLs
https://github.com/mahdyar/mahdyar/blob/main/README.md
https://github.com/mahdyar/mahdyar/blob/main/README.md?plain=1

## Screenshots

preview


source/features/view-markdown-source.tsx
```diff
@@ -50,6 +50,10 @@ This acts as an auto-discarded cache without globals, timers, etc.
 It should also work clicks on buttons sooner than the page loads.
 */
 async function showSource(): Promise<void> {
+	const newURL = window.location.protocol + "//" + window.location.host + window.location.pathname + '?plain=1' + window.location.hash;
```
sindresorhus, line 53: It would be better to use `URL` or URL manipulation.

sindresorhus:
Always ensure running `npm test` locally passes before pushing changes.

yakov116:
I don't think we need to get the blame anymore.  I think it can now be a simple link. We rendered the markdown because there was no other way.

source/features/view-markdown-source.tsx
mahdyar: > It would be better to use `URL` or URL manipulation.

So is it okay to add it to the dependencies? 🤔

mahdyar: > It would be better to use `URL` or URL manipulation.

On it! :)

fregante:
Indeed the whole feature should now be a single plain button linking between the 2 versions, I think. No fetches necessary

source/features/view-markdown-source.tsx
```diff
@@ -1,129 +1,30 @@
...
-	dispatchEvent(sourceButton, 'rgh:view-markdown-rendered');
-
-	(await lineActions()).remove();
-}
 
 async function init(): Promise<void> {
-	delegate(document, '.rgh-md-source:not(.selected)', 'click', showSource);
-	delegate(document, '.rgh-md-rendered:not(.selected)', 'click', showRendered);
-
+	const URLWithoutSearchParameters = window.location.href.split('?')[0];
```
fregante, line 9: This should use `.searchParams.remove('plain')` instead of dropping all search parameters. And the other one should use `.add('plain', '1')`

fregante (commented):
Great! We're almost there. However there are 2 problems:

- one of the buttons points to the current page, I think we should disable it, either via the `disabled` attribute or `btn-disabled` class (the attribute would be best). If not disabled, it should be "selected", I don't remember the class for that.
- Did this feature work with files other than markdown? Does it work after this change? Related: https://github.com/sindresorhus/refined-github/issues/4244

mahdyar:
> Great! We're almost there. However there are 2 problems:
> 
>     * one of the buttons points to the current page, I think we should disable it, either via the `disabled` attribute or `btn-disabled` class (the attribute would be best). If not disabled, it should be "selected", I don't remember the class for that. @yakov116?
> 
>     * Did this feature work with files other than markdown? Does it work after this change? Related: [Extend "markdown" features to more applicable pages #4244](https://github.com/sindresorhus/refined-github/issues/4244)

It's actually an `a` tag, so there is no `disabled` attribute for it. However I added `selected` class to the button which points to the current page. 
And, no, it's just working with markdown.

source/features/view-markdown-source.tsx
```diff
@@ -1,130 +1,33 @@
...
+	select('rgh-view-markdown-source').children[isPlain ? 1 : 0].classList.add('selected');
 }
 
 void features.add(__filebasename, {
 	include: [
 		pageDetect.isSingleFile
 	],
 	exclude: [
-		() => !select.exists('.blob .markdown-body')
+		() => !select.exists('.blob .markdown-body, .type-markdown')
```
fregante, line 29: If it only works with markdown files for now, we should just use a single isMarkdownFile `include`. We have it somewhere in our code

source/features/view-markdown-source.tsx
```diff
@@ -1,130 +1,33 @@
...
-	if (location.hash.startsWith('#L')) {
-		await showSource();
-
-		// Enable selected line highlight
-		window.dispatchEvent(new HashChangeEvent('hashchange', {
-			oldURL: location.href,
-			newURL: location.href
-		}));
-	}
+	select('rgh-view-markdown-source').children[isPlain ? 1 : 0].classList.add('selected');
```
fregante, line 21: Untested. Can you also include a screenshot of what this looks like?

fregante (commented):
I made some changes to point you in the right direction. Can you ensure it works?

source/features/view-markdown-source.tsx
fregante, line 21:
```suggestion
	select('.rgh-view-markdown-source')!.children[isPlain ? 1 : 0].classList.add('selected');
```

source/features/view-markdown-source.tsx
mahdyar: I couldn't find `isMarkdownFile`, but how about this?
```tsx
  include: [
  () => location.pathname.endsWith('.md')
  ],
```

source/features/view-markdown-source.tsx
```diff
@@ -1,130 +1,33 @@
...
-	dispatchEvent(sourceButton, 'rgh:view-markdown-rendered');
-
-	(await lineActions()).remove();
-}
 
 async function init(): Promise<void> {
-	delegate(document, '.rgh-md-source:not(.selected)', 'click', showSource);
-	delegate(document, '.rgh-md-rendered:not(.selected)', 'click', showRendered);
-
+	const isPlain = new URLSearchParams(location.hash).get('plain') === '1';
```
mahdyar: Here `location.hash` wouldn't work, since it's a query not a hash (e.g: `#L5`).

mahdyar:
And Here's what it looks like:
preview

source/features/view-markdown-source.tsx
fregante: Indeed I meant to write `.search` instead. The rest of the line did not need any changes

source/features/view-markdown-source.tsx
fregante: That’s not enough, markdown has plenty of extensions. You can copy/adapt the regex from this file. Please don’t change that file: 

https://github.com/sindresorhus/refined-github/blob/44f4ddc09618b2dadea8ff1a7aa19380708f2e1a/source/features/vertical-front-matter.tsx#L9

mahdyar: I mean that's what I copied from here:
https://github.com/sindresorhus/refined-github/blob/44f4ddc09618b2dadea8ff1a7aa19380708f2e1a/source/features/hide-markdown-diff.tsx#L37
No problem though, I'm on it! :)

fregante:
Can you solve the conflicts?

mahdyar:
> Can you solve the conflicts?

Done!

fregante:
Tested, works great, thank you!

PR #4538, opened 2021-07-05
Meta: Enable npm cache for actions

yakov116:
Re https://github.com/sindresorhus/refined-github/issues/4008#issuecomment-873594509

See https://github.blog/changelog/2021-07-02-github-actions-setup-node-now-supports-dependency-caching

## Test URLs
NA

## Screenshot
NA

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,6 +23,7 @@ jobs:
       - uses: actions/setup-node@v1
         with:
           node-version: 14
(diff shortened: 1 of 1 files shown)


fregante:
This is only supported by v2 of the action as far as I know

fregante:
Also we should bump the Node version to 16 everywhere

yakov116:
Will do

PR #4543, opened 2021-07-07
Meta: Require Node.js 16

yakov116:
Re https://github.com/sindresorhus/refined-github/pull/4538#issuecomment-874910614


fregante:
We also have .nvmrc I think

fregante (approved):

PR #4544, opened 2021-07-07
Lint

fregante:
Hurray for XO v0.41

Depends on:

- https://github.com/sindresorhus/refined-github/pull/4543#pullrequestreview-700880592


yakov116:
I did not update since it break the VS code extension 😢. Which makes linting so much harder (must to npx xo before every commit)

.xo-config.json
```diff
@@ -11,6 +11,12 @@
 		"__filebasename"
 	],
 	"rules": {
+		"@typescript-eslint/no-unsafe-assignment": "warn",
+		"@typescript-eslint/no-unsafe-member-access": "warn",
+		"@typescript-eslint/no-unsafe-return": "warn",
+		"@typescript-eslint/no-unsafe-call": "warn",
```
fregante, line 17: We have `any` in a lot of places, especially but not only due to API calls. I don't know if we can fix these without making API processing verbose. We specifically use `any` in there to avoid having to type the API. Any code we add won't make the code any safer

fregante:
Is there an open issue about it? In my case I just had to update the global XO

package.json
```diff
@@ -94,6 +94,7 @@
 		"@types/firefox-webext-browser": "^82.0.0",
 		"@types/jsdom": "^16.2.10",
 		"@types/mini-css-extract-plugin": "^1.4.3",
+		"@types/node": "^16.0.0",
```
fregante, line 97: This shouldn't be necessary, but… https://github.com/xojs/xo/issues/552#issuecomment-873406603

fregante:
Closing this in favor of a comma-dangle-only PR

.xo-config.json
```diff
@@ -32,9 +31,7 @@
 				"allowConciseArrowFunctionExpressionsStartingWithVoid": false
 			}
 		],
-		"import/first": "error",
 		"import/newline-after-import": "error",
-		"import/no-commonjs": "error",
```
fregante, line 37: Some rules are now the default in XO or have been fixed by this PR

source/features/avoid-accidental-submissions.tsx
```diff
@@ -10,11 +10,11 @@ function addQuickSubmit(): void {
 
 function onKeyDown(event: delegate.Event<KeyboardEvent, HTMLInputElement>): void {
 	if (
-		event.key !== 'Enter' ||
-		event.ctrlKey ||
-		event.metaKey ||
-		event.isComposing || // #4323
-		select.exists('.suggester', event.delegateTarget.form!) // GitHub’s autocomplete dropdown
+		event.key !== 'Enter'
+		|| event.ctrlKey
```
sindresorhus, line 14: I wish we could enforce this to be:
```suggestion
			|| event.ctrlKey
```

source/resolve-conflicts.ts
```diff
@@ -1,5 +1,7 @@
+// eslint-disable-next-line @typescript-eslint/triple-slash-reference -- TODO: Is any other way supported?
 /// <reference types="codemirror" />
```
sindresorhus, line 2: Not if it's global, but you could try getting them to add an export too and import `codemirror` instead.

source/features/table-input.tsx
```diff
@@ -14,12 +14,12 @@ function addTable({delegateTarget: square}: delegate.Event<MouseEvent, HTMLButto
...
 
 	field.focus();
-	const table =
-		'<table>\n' +
-			('<tr>\n' +
-				'\t<td>\n'.repeat(Number(square.dataset.x))
-			).repeat(Number(square.dataset.y)) +
-		'</table>';
+	const table
+		= '<table>\n'
```
fregante, line 18: Not sure I like _this_ change 😳

source/github-events/on-diff-file-load.ts
```diff
@@ -4,13 +4,11 @@ import delegate from 'delegate-it';
...
 	'include-fragment.js-diff-entry-loader', // File diff loader on clicking "Load Diff"
-].join();
+].join(',');
 
 // This lets you call `onDiffFileLoad` multiple times with the same callback but only ever a `load` listener is registered
-const getDeduplicatedHandler = mem((callback: EventListener): delegate.EventHandler => {
-	return (event: delegate.Event) => {
-		event.delegateTarget.addEventListener('load', callback);
-	};
+const getDeduplicatedHandler = mem((callback: EventListener): delegate.EventHandler => (event: delegate.Event) => {
```
fregante, line 10: This is a little hard to see on one line. Should it at least be this?
```suggestion
const getDeduplicatedHandler = mem((callback: EventListener): delegate.EventHandler =>
	(event: delegate.Event) => {
```

Or even what Prettier probably suggests:
```suggestion
const getDeduplicatedHandler = mem(
	(callback: EventListener): delegate.EventHandler =>
		(event: delegate.Event) => {
```

source/helpers/dom-utils.ts
```diff
@@ -53,11 +53,9 @@ export const wrapAll = (targets: Array<Element | ChildNode>, wrapper: Element):
...
 };
 
-export const isEditable = (node: unknown): boolean => {
-	return node instanceof HTMLTextAreaElement ||
-		node instanceof HTMLInputElement ||
-		(node instanceof HTMLElement && node.isContentEditable);
-};
+export const isEditable = (node: unknown): boolean => node instanceof HTMLTextAreaElement
+		|| node instanceof HTMLInputElement
+		|| (node instanceof HTMLElement && node.isContentEditable);
```
fregante, line 58: Probably

```suggestion
export const isEditable = (node: unknown): boolean => (
	node instanceof HTMLTextAreaElement
	|| node instanceof HTMLInputElement
	|| (node instanceof HTMLElement && node.isContentEditable);
)
```

yakov116:
> Is there an open issue about it? In my case I just had to update the global XO
I am talking about ⬇️ 
https://github.com/SamVerschueren/vscode-linter-xo/issues/91

source/features/bypass-checks.tsx
```diff
@@ -23,7 +23,7 @@ function init(): void {
 	const thirdPartyApps = [
 		`a:not([href="/apps/github-actions"]) ~ div .status-actions[href^="${location.origin}"]:not(.rgh-bypass-link)`, // Hovercard status checks
 		'a:not([href="/apps/github-actions"]) ~ div .status-actions[href^="/"]:not(.rgh-bypass-link)',
-	].join();
+	].join(',');
```
yakov116, line 26: I really don't like this rule JMHO

PR #4552, opened 2021-07-09
Fix for highest rated comment header

### LINKED ISSUE #4529: `highest-rated-comment` has a red background in comment header
While looking at issues on GitHub, I noticed that some comments has a bright red background header, where it was dark green before. 
E.g. this comment
https://github.com/vercel/swr/issues/93#issuecomment-551110903
image

Refined GitHub seems to somehow change the css variables definition
image

When I disable Refined GitHub, it is back to being the default black header background.

4086606:
Closes #4529.

## Test URLs

https://github.com/isaacs/github/issues/402#issuecomment-239936197
https://github.com/vercel/swr/issues/93#issuecomment-551110903

### Before
bugfix before

### After
bugfix after

### DIFF
--- a/source/features/highest-rated-comment.css
+++ b/source/features/highest-rated-comment.css
@@ -18,7 +18,6 @@ a.rgh-highest-rated-comment .avatar {
 .rgh-highest-rated-comment::after {
 	--color-bg-info-inverse: var(--color-text-success);
 	--color-bg-tertiary: var(--color-text-success);
-	left: initial;
 }
 
 .rgh-highest-rated-comment .timeline-comment-header {
--- a/source/features/highest-rated-comment.css
+++ b/source/features/highest-rated-comment.css
@@ -19,3 +19,22 @@ a.rgh-highest-rated-comment .avatar {
 	--color-text-secondary: var(--color-text-success);
 	--color-text-tertiary: var(--color-text-success);
 }
+
+.rgh-highest-comment .timeline-comment-header-text::after {
+    background-color: var(--color-text-success);
+    content: '';
+    display: inline-block;
+    height: 18px;
+    -webkit-mask-image: url('https://raw.githubusercontent.com/primer/octicons/main/icons/check-16.svg');
+    -webkit-mask-position: center;
+    -webkit-mask-repeat: no-repeat;
+    -webkit-mask-size: 16px;
+    mask-image: url('https://raw.githubusercontent.com/primer/octicons/main/icons/check-16.svg');
+    mask-position: center;
+    mask-repeat: no-repeat;
+    mask-size: 16px;
+    position: relative;
+    top: -1px;
+    width: 18px;
+    vertical-align: middle;
+}
--- a/source/features/highest-rated-comment.css
+++ b/source/features/highest-rated-comment.css
@@ -14,8 +14,8 @@ a.rgh-highest-rated-comment .avatar {
 	left: -55px;
 }
 
-.rgh-highest-rated-comment:before,
-.rgh-highest-rated-comment:after {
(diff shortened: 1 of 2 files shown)


source/features/highest-rated-comment.css
```diff
@@ -14,12 +14,16 @@ a.rgh-highest-rated-comment .avatar {
...
 
-.rgh-highest-rated-comment {
-	--color-auto-gray-1: var(--color-auto-green-1);
-	--color-auto-blue-4: var(--color-auto-green-4);
-	--color-auto-blue-2: var(--color-auto-green-2);
+.rgh-highest-rated-comment:before,
+.rgh-highest-rated-comment:after {
+	--color-bg-info-inverse: var(--color-text-success, #3fb950);
+	--color-bg-tertiary: var(--color-text-success, #3fb950);
+	left: initial;
```
fregante, line 21: This should not be necessary I think

source/features/highest-rated-comment.css
```diff
@@ -14,12 +14,16 @@ a.rgh-highest-rated-comment .avatar {
...
 }
 
-.rgh-highest-rated-comment {
-	--color-auto-gray-1: var(--color-auto-green-1);
-	--color-auto-blue-4: var(--color-auto-green-4);
-	--color-auto-blue-2: var(--color-auto-green-2);
+.rgh-highest-rated-comment:before,
+.rgh-highest-rated-comment:after {
+	--color-bg-info-inverse: var(--color-text-success, #3fb950);
+	--color-bg-tertiary: var(--color-text-success, #3fb950);
```
fregante, line 20: GitHub has hundreds of color variables, we should never need to use hardcoded variables.

fregante (changes requested):
I think that we should perhaps try to match the discussion styles, by just changing the border color and then adding that round checkmark icon

Example: https://github.com/vercel/next.js/discussions/19409

source/features/highest-rated-comment.css
4086606: They are fallbacks

source/features/highest-rated-comment.css
4086606: It unfortunately is, the green caret appears black otherwise

source/features/highest-rated-comment.css
4086606: I've decided to remove them, let's hope Github doesn't redesign the site anytime soon

4086606: Resolved in b959d25.

source/features/highest-rated-comment.css
4086606: I realise you might be speaking of `left: initial` which fixes the caret positioning

fregante:
> I think that we should perhaps try to match the discussion styles, by just changing the border color and then adding that round checkmark icon

4086606:
Text color as well? Good idea, one moment

source/features/highest-rated-comment.css
fregante:
```suggestion
```

fregante: This is a color change. If GitHub can style the arrow correctly it means its position is correct.


Screen Shot

Also, it's not being applied at all:

Screen Shot 7

4086606:
@fregante I did as you asked and also added a 👍 indicator. Should I change the `mask-image` implementation into a JS icon injection?

4086606:
All done for you @fregante:

highest-rated-comment-bugfix

fregante:
> Don't open a PR that requires 4 reviews; 

That does not match the style of discussions at all. Sorry, closing.

PR #4556, opened 2021-07-11
`avoid-accidental-submissions`: Keep focus and inform the user

### LINKED ISSUE #4462: In `avoid-accidental-submissions`, focus primary button instead of next field
I think `avoid-accidental-submissions` feels a little strange, like slipping on a banana peel. What if we focus the submit button instead of the text area?

Screen Shot 9

Pros:

- submission is possible with an additional plain [Enter] press, without using ctrl/cmd
- maybe it feels less broken

Cons:

- if it _was_ accidental, going back now means pressing shift-tab several times instead of once

---

An alternative would be to add a temporary "confirm" button right below it and focus it.

Screen Shot

The message would still be welcome everywhere.

darkred:
Closes #4462 (based on https://github.com/sindresorhus/refined-github/issues/4462#issuecomment-876987450 and https://github.com/sindresorhus/refined-github/issues/4462#issuecomment-877253535)
The feature now:
- Prevents first <kbd>enter</kbd>, no focus change.
- Shows message ~above~ below the title/commit field.
- Allows a second enter on the field, currently any time later.

## Test URLs

- New issue: https://github.com/sindresorhus/refined-github/issues/new
- New PR: https://github.com/sindresorhus/refined-github/compare/main...darkred:avoid-accidental-submissions-keep-focus-show-message?expand=1
- Merge PR commit: any non-draft PR
- 'create new file' commit: https://github.com/sindresorhus/refined-github/new/main

## Screenshots

<details>
<summary>New issue:</summary>

Animation
</details>

<details>
<summary>New PR:</summary>

Animation2
</details>

<details>
<summary>Merge PR commit:</summary>

Animation3
</details>

<details>
<summary>'create new file' commit:</summary>

Animation4
</details>

## Notes to reviewers:

- In the last case, the message is displayed on the very top, above the filename textbox. I think it might be better if it was displayed lower, right above the commit message textbox. What do you say?
- Following fregante's comment, the message now has blue background color, instead of yellow.
- Would it be better if I used in the message `<code>` tags for the keys  ?

- With this PR, in the new Issue/PR forms, if you press Enter while the title is empty, the message is still displayed, even though the title can't be empty. I believe it would be better if it was only displayed when there's some typed text in the textbox. *Maybe this would better be as part of another PR ?*
- The gif for the feature in `readme.md` should be updated accordingly in the end.


fregante:
On second thought, this feels a bit the like the opposite of what it was: The prevention was silent and now it's too loud. 😅 

Possible improvements:

- use blue instead of yellow (definitely)
- add a background-less smaller message immediately below the field, as previously suggested? (without the button)
	
	121675762-0a57aa80-cade-11eb-8c24-b428807d8f44


- or with the button

	Screen Shot 

- use a temporary overlay exactly like Chrome

	gif



I think the second bullet _might_ be best because:

- it's less noisy
- it doesn't move the focus (if it _was_ accidental, the user definitely wants to stay in the field and continue typing)

Open to more suggestions. 

@darkred Either way wait until we decide before updating the readme

yakov116:
I am leaning towards option 2. 
I think this change would only be necessary for issues. I _think_ commits and merges the old way was ok, but I am not sure.

I will try out the branch later today and let you know my final thoughts.

fregante:
@yakov116 if this message makes sense, I think it should be shown everywhere the feature runs. It wouldn’t make sense to be visible in some places and just change the behavior silently in others.

yakov116:
I agree. My thinking is that when merging and commiting the submit button is just below the textArea. So another button confirm right below is unnecessary.
However that is not the case with new Issues/PR's. (Its a big textArea)
I agree that we should be consistent. 

I want to try out the branch to checkout the annoyance level and will give you feedback.

yakov116 (changes requested):
Option 2 or 3.
- add a background-less smaller message immediately below the field, as previously suggested? (without the button)
	
	121675762-0a57aa80-cade-11eb-8c24-b428807d8f44


- or with the button

	Screen Shot 

But I think it has to take into account if an issue con be summited too.
Currently if I create a blank issue I get the message.

darkred:
I think I have followed all your suggestions:

- I followed Option 2 (I used an arbitrary classname for the message element)
- Tested it in all cases: Issues, PRs compare, PR merge, new file, edit file
- Now the message in new Issues only appears when the title is not empty:
Animation5

fregante:
@darkred instead of checking isIssue and empty fields, what matters to us is whether the form can be submitted. There might be other logic hidden in the form and it’s best to just verify whether the submit button is disabled. Selecting the right button on every page we cover could be troublesome but [type=submit] should bring you almost all the way there.

source/features/avoid-accidental-submissions.tsx
```diff
@@ -14,27 +15,37 @@ function onKeyDown(event: delegate.Event<KeyboardEvent, HTMLInputElement>): void
...
+		|| select.exists([
+			'.suggester', // GitHub’s autocomplete dropdown
+			'.toRemove',
+		],
+		event.delegateTarget.form!)
 	) {
 		return;
 	}
 
+	const message = <p className="toRemove">A submission via <kbd>enter</kbd> has been prevented. You press <kbd>enter</kbd> again or use <kbd>ctrl</kbd>-<kbd>enter</kbd> next time</p>;
```
fregante, line 27: Use classes that start with “rgh-“ and are lowercase so we always know they’re ours.

fregante: Also this likely needs `.mt-1`

darkred:
I know have a main `if`, to check whether the submit button is enabled (not disabled):

```diff
+ if (select.exists('.btn-primary[type="submit"]:not([disabled])')) {
		if (pageDetect.isNewFile() || pageDetect.isEditingFile() || pageDetect.isPRConversation()) {
			focusedInput.after(message);
		} else {
			focusedInput.parentElement!.append(message);
		}
+ }
```

&nbsp;
Note that in PR compare pages, (e.g. `https://github.com/darkred/test/compare/main...t1`) for some strange reason, when having RG enable, even with the 'avoid-accidental-submissions` feature disabled, the "Create pull request" button is initially enabled, even with empty title:

But, if I disable RG, then the button is initially disabled with empty title, as expected:

fregante:
> **instead of** checking isIssue

fregante:
Can you open an issue for that bug?

darkred:
> Can you open an issue for that bug?

Yes, here: #4565

darkred:
> **instead of** checking isIssue

I don't quite understand what you mean:  
after my last changes, there's no longer check for `isIssue` (nor checking for empty fields ) - there's now only an `if` check whether the page's main submit button is active.

The contained `if` :

```js
	if (pageDetect.isNewFile() || pageDetect.isEditingFile() || pageDetect.isPRConversation()) {
	 // ..
	} else {
	 // ..
```
is still needed, it's for a different reason: to create the message, with `.after()` or `parentElement.append()` as needed.

&nbsp;
What more is needed for me to do?

source/features/avoid-accidental-submissions.tsx
```diff
@@ -14,27 +15,39 @@ function onKeyDown(event: delegate.Event<KeyboardEvent, HTMLInputElement>): void
...
+		event.delegateTarget.form!)
 	) {
 		return;
 	}
 
+	const message = <p className="rgh-avoid-accidental-submissions my-1">A submission via <kbd>enter</kbd> has been prevented. You press <kbd>enter</kbd> again or use <kbd>ctrl</kbd>-<kbd>enter</kbd> next time</p>;
+
+	const focusedInput = select(inputElements, event.delegateTarget.form!)!;
+
+	if (select.exists('.btn-primary[type="submit"]:not([disabled])')) {
```
fregante, line 31: This is only preventing the message, but it’s still stopping the submission. The message should always be shown if the submission has been prevented 
```suggestion
	if (!select.exists('.btn-primary[type="submit"]:not([disabled])')) {
	   return
	}
```

Also the selector just looks for that button anywhere on the page, but it should be limited to the form instead.

darkred: Thank you. So I changed it into:

```js
	if (!(select.exists('.btn-primary[type="submit"]:not([disabled])'), event.delegateTarget.form!)) {
		return;
	}
```

source/features/avoid-accidental-submissions.tsx
```diff
@@ -14,27 +15,41 @@ function onKeyDown(event: delegate.Event<KeyboardEvent, HTMLInputElement>): void
...
+		event.delegateTarget.form!)
 	) {
 		return;
 	}
 
+	const message = <p className="rgh-avoid-accidental-submissions my-1">A submission via <kbd>enter</kbd> has been prevented. You press <kbd>enter</kbd> again or use <kbd>ctrl</kbd>-<kbd>enter</kbd> next time</p>;
+
+	const focusedInput = select(inputElements, event.delegateTarget.form!)!;
+
+	if (!(select.exists('.btn-primary[type="submit"]:not([disabled])'), event.delegateTarget.form!)) {
```
fregante:
```suggestion
	if (!select.exists('.btn-primary[type="submit"]:not([disabled])'), event.delegateTarget.form!) {
```

source/features/avoid-accidental-submissions.tsx
```diff
@@ -9,32 +10,48 @@ function addQuickSubmit(): void {
...
-		|| select.exists('.suggester', event.delegateTarget.form!) // GitHub’s autocomplete dropdown
+		|| select.exists([
+			'.suggester', // GitHub’s autocomplete dropdown
+			'.rgh-avoid-accidental-submissions.my-1',
+		], form)
 	) {
 		return;
 	}
 
+	if (!select.exists('.btn-primary[type="submit"]:not(:disabled)', form)) {
```
fregante, line 27: Does this also work? It needs to be tested in every form, because it's not exactly equivalent.
```suggestion
	if (select.exists('.btn-primary[type="submit"]:disabled', form)) {
```

source/features/avoid-accidental-submissions.tsx
```diff
@@ -9,32 +10,48 @@ function addQuickSubmit(): void {
...
+		return;
+	}
+
+	const focusedInput = select(inputElements, form)!;
```
fregante, line 31: I think you just need to use `event.delegateTarget`, no need to reselect the item. The changes around `select().focus()` need to be undone

source/features/avoid-accidental-submissions.tsx
darkred: Yes, I tested it and it works in every form.

source/features/avoid-accidental-submissions.tsx
darkred: I have followed your suggestions in the new commit.
Tested it in all cases and it works.

source/features/avoid-accidental-submissions.tsx
```diff
@@ -9,32 +10,48 @@ function addQuickSubmit(): void {
...
 	) {
 		return;
 	}
 
+	if (select.exists('.btn-primary[type="submit"]:disabled', form)) {
+		return;
+	}
+
+	const message = (
+		<p className="rgh-avoid-accidental-submissions my-1">
```
fregante, line 33: This spacing looks great on `isEditingFile`:

Screen Shot 1

But not so much on `isNewIssue`:

Screen Shot 2

`mt-2 mb-0` would work better:

Screen Shot 1

Let's also make sure the spacing is right in every page this feature runs on. Can you fix this and post some screenshots?

source/features/avoid-accidental-submissions.tsx
```diff
@@ -9,32 +10,48 @@ function addQuickSubmit(): void {
...
+			A submission via <kbd>enter</kbd> has been prevented. You can press <kbd>enter</kbd> again or use <kbd>ctrl</kbd>-<kbd>enter</kbd>.
```
fregante, line 34: Alternative text suggestions welcome.

Also we need a Mac detection in `GitHub-helpers/index.ts`, like `navigator.userAgent.includes('Macintosh')`

```suggestion
			A submission via <kbd>enter</kbd> has been prevented. You can press <kbd>enter</kbd> again or use <kbd>{isMac ? 'cmd' : 'ctrl'}</kbd><kbd>enter</kbd>.
```

fregante (commented):
Small changes required, but no gif update necessary.

source/features/avoid-accidental-submissions.tsx
darkred: Yes, I have fixed this following your suggestion.
Here are screenshots:

<details>
<summary>New Issue</summary>


</details>

<details>
<summary>New PR/branch compare</summary>


</details>

<details>
<summary>PR merge</summary>


</details>

<details>
<summary>New file commit</summary>


</details>

<details>
<summary>Edit file commit</summary>


</details>

source/features/avoid-accidental-submissions.tsx
```diff
@@ -1,40 +1,61 @@
+import React from 'dom-chef';
 import select from 'select-dom';
 import delegate from 'delegate-it';
 import * as pageDetect from 'github-url-detection';
 
+import {isMac} from '../github-helpers';
+
```
yakov116, line 7: Extra Space

darkred: Ok, I swapped its place with the import after it, and removed the extra line.

source/features/avoid-accidental-submissions.tsx
```diff
@@ -1,40 +1,60 @@
+import React from 'dom-chef';
 import select from 'select-dom';
 import delegate from 'delegate-it';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
+import {isMac} from '../github-helpers';
```
yakov116, line 7:
```suggestion
import {isMac} from '../github-helpers';
import features from '.';
```

darkred: I got the order from:
https://github.com/sindresorhus/refined-github/blob/afa8da1a27e80f6364a3e68ab4b4c5a7c1916d2e/source/features/reactions-avatars.tsx#L7-L9

yakov116: We always sort the word "from" as a tree

darkred: I didn't know that. Thanks.

source/features/avoid-accidental-submissions.tsx
```diff
@@ -1,40 +1,60 @@
...
+		], form)
 	) {
 		return;
 	}
 
+	if (select.exists('.btn-primary[type="submit"]:disabled', form)) {
+		return;
+	}
+
+	const preSelector = pageDetect.isNewFile() || pageDetect.isEditingFile() ? 'my-1' : 'mt-2 mb-n1';
```
fregante, line 33:
```suggestion
	const spacingClasses = pageDetect.isNewFile() || pageDetect.isEditingFile() ? 'my-1' : 'mt-2 mb-n1';
```

source/github-helpers/index.ts
```diff
@@ -38,6 +38,7 @@ export const getCurrentCommittish = (pathname = location.pathname, title = docum
 };
 
 export const isFirefox = navigator.userAgent.includes('Firefox/');
+export const isMac = navigator.userAgent.includes('Macintosh/');
```
fregante, line 41: Not the same
```suggestion
export const isMac = navigator.userAgent.includes('Macintosh');
```

fregante (approved):
Looks great! Thanks for the many screenshots ☺️

yakov116 (approved):
@fregante don't co-author by me when merging, I only added a lint.

fregante:
Thanks @darkred!

fregante:
I've been seen it today. It looks pleasant 😌

PR #4557, opened 2021-07-11
Fix: `set-default-repositories-type-to-sources` changes unrelated URLs

yakov116:
Currently all tabs get the `?type=source`

## Test URLs
https://github.com/microsoft

## Screenshot

None


fregante:
Let's rename `set-default-repositories-type-to-sources` to something shorter at some point 😅

PR #4560, opened 2021-07-12
Make `follow-file-renames` more noticeable by adding icons

### LINKED ISSUE #4477: Make `follow-file-renames` more noticeable
I've tried installing the extension and after doing this branch - https://github.com/analogdevicesinc/scopy/tree/try_rename_CI, which just changes the name of the folder for
https://github.com/analogdevicesinc/scopy/commits/try_rename_CI/ci/appveyor/build_appveyor_flatpak.sh . I can still see blame. Is there something I'm missing or I'm misinterpreting the follow-rename feature ?

yakov116:
Closes #4477 

## Test URLs
https://github.com/sindresorhus/refined-github/commits/main?after=285eaec360f21431757236c6c39cdbe6f303f6e2+34&branch=main&path%5B%5D=source&path%5B%5D=background.ts

## Screenshot

# Before
image
<hr>

# After
image

### DIFF
--- a/source/features/follow-file-renames.tsx
+++ b/source/features/follow-file-renames.tsx
@@ -1,7 +1,7 @@
 import React from 'dom-chef';
 import select from 'select-dom';
 import * as pageDetect from 'github-url-detection';
-import {ArrowLeftIcon, ArrowRightIcon} from '@primer/octicons-react';
+import {DiffRenamedIcon} from '@primer/octicons-react';
 
 import features from '.';
 import * as api from '../github-helpers/api';
@@ -41,7 +41,7 @@ async function linkify(button: HTMLButtonElement, url: GitHubURL): Promise<void 
 						aria-label={`Renamed ${isNewer ? 'to' : 'from'} ${file[toKey]}`}
 						className="btn btn-outline BtnGroup-item tooltipped tooltipped-n tooltipped-no-delay"
 					>
-						{isNewer && <ArrowLeftIcon className="mr-1"/>}{button.textContent}{!isNewer && <ArrowRightIcon className="ml-1"/>}
+						{isNewer && <DiffRenamedIcon className="mr-1" transform="rotate(180)"/>}{button.textContent}{!isNewer && <DiffRenamedIcon className="ml-1"/>}
 					</a>,
 				);
 			}
--- a/source/features/follow-file-renames.tsx
+++ b/source/features/follow-file-renames.tsx
@@ -1,6 +1,7 @@
 import React from 'dom-chef';
 import select from 'select-dom';
 import * as pageDetect from 'github-url-detection';
+import {ArrowLeftIcon, ArrowRightIcon} from '@primer/octicons-react';
 
 import features from '.';
 import * as api from '../github-helpers/api';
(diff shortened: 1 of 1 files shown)


fregante:
The arrow alone doesn’t yell “renamed” to me. Maybe the diff-specific renamed icon works better? Not sure it’s particularly better either way though

https://primer.style/octicons/diff-renamed-16

Maybe it should be coupled with the file icon? The submodule and sym-link icons are nice too but they’d be misused here

yakov116:
> The arrow alone doesn’t yell “renamed” to me. Maybe the diff-specific renamed icon works better? Not sure it’s particularly better either way though
> 
> [primer.style/octicons/diff-renamed-16](https://primer.style/octicons/diff-renamed-16)

I don't think the diff icon is any better. I chose this icon since it has an arrow going the other direction.



> Maybe it should be coupled with the file icon? The submodule and sym-link icons are nice too but they’d be misused here

I am looking at the issues as follows: The user cannot tell if the file was renamed, so lets make there something different about the button. I dont think adding a file icon will make it anymore noticeable.

WDYT?

fregante:
I think it _is_ better because the icon is more specific on GitHub; What I don't know if it's the best we can do.

yakov116:
Does this look better to you?

image

image

yakov116:
🚀 ?

source/features/follow-file-renames.tsx
```diff
@@ -15,7 +16,7 @@ interface File {
 async function findRename(lastCommitOnPage: string): Promise<File[]> {
 	// API v4 doesn't support it: https://github.community/t/what-is-the-corresponding-object-in-graphql-api-v4-for-patch-which-is-available-in-rest-api-v3/13590
 	const {files} = await api.v3(`commits/${lastCommitOnPage}`);
-	return files;
+	return files as File[];
```
fregante, line 19: Redundant

source/features/follow-file-renames.tsx
```diff
@@ -40,7 +41,7 @@ async function linkify(button: HTMLButtonElement, url: GitHubURL): Promise<void
 						aria-label={`Renamed ${isNewer ? 'to' : 'from'} ${file[toKey]}`}
 						className="btn btn-outline BtnGroup-item tooltipped tooltipped-n tooltipped-no-delay"
 					>
-						{button.textContent}
+						{isNewer && <DiffRenamedIcon className="mr-1" transform="rotate(180)"/>}{button.textContent}{!isNewer && <DiffRenamedIcon className="ml-1"/>}
```
fregante, line 44: Should be split on 3 lines

fregante (approved):
It’s fine for me but let’s see it there are other opinions. No rush to merge, the next release will be in 7/10/15 days 😃

source/features/follow-file-renames.tsx
yakov116: Takes away a xo warning

fregante: I think we'll have to disable those rules altogether, I didn't notice we have 170 warnings and they're all around API calls (which we won't type)
```suggestion
	return files;
```

fregante:
I think `transform` doesn't work on `<svg>` in Safari:

Screen Shot 2

Test: https://github.com/ForkAwesome/Fork-Awesome/commits/master/source/_plugins/site.rb

PR #4568, opened 2021-07-14
Rename `remove-unused-repo-tabs` to `clean-repo-tabs`

### LINKED ISSUE #4562: `remove-unused-repo-tabs` doesn't move tabs into "more" if `more-dropdown` is disabled
### Please ensure:

1. Enable `remove-unused-repo-tabs` and disable `more-dropdown`.
2. Go to a repo that you can't edit and has unused tabs.
3. The unused repo tabs disappears instead of being moved to "more".

### My suggestion

1. Add a feature that moves "security" and "insights" to the dropdown menu.
2. Add another feature that adds "Compare", "Dependencies", "Commits" and "Branches" in the dropdown menu.
3. Make `more-dropdown` not a "feature", so that any feature can use it, and if there are any features using it, the dropdown menu appears.

BTW, in [the current issue form](https://github.com/sindresorhus/refined-github/blob/9d03cc27e7ff17aaeb3cd56f3ce658141e6a5731/.github/ISSUE_TEMPLATE/1_bug_report.yml), there's no section like "Anything elses?" or "Your suggestions", so I don't know where to properly provide my suggestions.

img

### Console errors, if any

_No response_

### Example URL

https://github.com/trustwallet/assets

### Browser(s) used

Firefox 89.0.2

yakov116:
Closes #4562

## Test URLs

## Screenshot
(Click to enlarge)

<table>
<tr>
	<td>Only <code>clean-repo-tabs</code> enabled
	<td>Only <code>more-dropdown</code> enabled
	<td>Both enabled
<tr>
	<td>
	<td>
	<td>
</table>

**Note:** `d-none` works perfectly and the user should be able to unhide via css.


source/features/more-dropdown.tsx
```diff
@@ -33,17 +33,19 @@ export function onlyShowInDropdown(id: string): void {
 	select('.js-responsive-underlinenav-overflow ul')!.append(menuItem);
 }
 
-async function init(): Promise<void> {
+export async function unhideOverflowDropdown(): Promise<void> {
```
yakov116, line 36: @cheap-glitch can you think of a better function name?

cheap-glitch: Maybe `alwaysShowDropdown` or `showOverflowDropdown`? (but I think it's fine as is :+1:)

source/features/clean-repo-tabs.tsx
```diff
@@ -9,8 +9,12 @@ import getTabCount from '../github-helpers/get-tab-count';
 import looseParseInt from '../helpers/loose-parse-int';
 import {getWorkflows} from './next-scheduled-github-action';
 import abbreviateNumber from '../helpers/abbreviate-number';
-import {onlyShowInDropdown} from './more-dropdown';
 import {buildRepoURL, getRepo} from '../github-helpers';
+import {onlyShowInDropdown, unhideOverflowDropdown} from './more-dropdown';
+
+async function canUserEditOrganization(): Promise<boolean> {
+	return Boolean(await elementReady('.btn-primary[href$="repositories/new"]'));
```
fregante, line 16: This only works on the Repositories tab. Refer to *both* the screenshots I posted or visit https://github.com/rollup and look for items in each tab. There probably isn't just one for everyone and maybe on some tabs there's nothing at all, so let's just try to match as many as possible.

I'm looking for something like:
```suggestion
	return Boolean(await elementReady([
		'.btn-primary[href$="repositories/new"]', // repo tab, or isOrgRepos if it exists
		'other button', // other tab,
		'etc
```

source/features/more-dropdown.tsx
```diff
@@ -24,7 +24,7 @@ export function onlyShowInDropdown(id: string): void {
 		return;
 	}
 
-	(tabItem!.closest('li') ?? tabItem!.closest('.UnderlineNav-item'))!.remove();
+	(tabItem!.closest('li') ?? tabItem!.closest('.UnderlineNav-item'))!.classList.add('d-none');
```
fregante, line 27: Can also be done with `.hidden = true`, which I generally prefer since it's native, but it's _ok_ either way.

source/options-storage.ts
```diff
@@ -23,6 +23,7 @@ const migrations = [
 	featureWasRenamed('separate-draft-pr-button', 'one-click-pr-or-gist'), // Merged in May
 	featureWasRenamed('prevent-pr-commit-link-loss', 'prevent-link-loss'), // Merged in May
 	featureWasRenamed('remove-projects-tab', 'remove-unused-repo-tabs'), // Merged in July
+	featureWasRenamed('remove-unused-repo-tabs', 'clean-repo-tabs'), // Merged in July
```
fregante, line 26: We should also probably add a migration from `more-dropdown` to `clean-repo-tabs` because people have been disabling the feature to restore security and insights.

fregante (approved):
Untested but it looks pretty straightforward! 🎂 

One thing I'd do is rename the ancient `more-dropdown` to `more-dropdown-links` and make sure this is searched and replaced across the repo.

source/features/clean-repo-tabs.tsx
yakov116: I think you missed my post in the other PR.

There are no counts on any other pages. So we cannot hide anything.

yakov116:
> One thing I'd do is rename the ancient [`more-dropdown`](/sindresorhus/refined-github/blob/main/source/features/more-dropdown.tsx) to `more-dropdown-links` **and make sure this is searched and replaced across the repo.**

Super complicated 😄

source/features/more-dropdown.tsx
yakov116: Will require `as HtmlElelemnt` worth it?

fregante:
> Maybe we should just keep it as is?

It hasn't made sense in a while. `more-dropdown` literally added a dropdown. Now the dropdown already exists and is native.

source/features/more-dropdown.tsx
fregante: Nah, but still need https://github.com/sindresorhus/refined-github/issues/4008#issuecomment-823814526

source/features/clean-repo-tabs.tsx
fregante: Which makes me wonder, why are we naming this `clean-REPO-tabs` when it also covers organizations? And why doesn't it cover regular users?

fregante: Anyway yes I missed that message. And yuck every tab is different on there 😂 

gif

yakov116: Its only one tab in organizations. We could drop it altogether.



> And why doesn't it cover regular users?

Not sure what you mean here?

source/refined-github.ts
```diff
@@ -220,4 +220,4 @@ import './features/list-pr-for-branch';
 import './features/cancel-wiki-edit-button';
 import './features/comment-on-draft-pr-indicator';
 import './features/select-notifications';
-import './features/remove-unused-repo-tabs';
+import './features/clean-repo-tabs';
```
fregante, line 223: This also covers organizations so it should be named… `clean-tabs`? 🤔

yakov116: Its only one tab in organizations. We could drop it altogether. Or make it a separate feature

fregante (approved):
Good to go! You may or may not want to extract the `more-dropdown-links`-related changes to their own PR

yakov116:
> One thing I'd do is rename the ancient [`more-dropdown`](/sindresorhus/refined-github/blob/main/source/features/more-dropdown.tsx) to [`more-dropdown-links`](/sindresorhus/refined-github/blob/main/source/features/more-dropdown-links.tsx) **and make sure this is searched and replaced across the repo.**

You would like me change titles of issues? Can you give me guidance?

PR #4571, opened 2021-07-16
Fix: Features are not being applied on page change

### LINKED ISSUE #4567: Features are not being applied on page change
### Please ensure:

When changing pages unless I do a full refresh none of the CSS features are showing. ~I am not sure if this is an issue for all features.~ its all

https://user-images.githubusercontent.com/16872793/125619918-b62ad23f-2bd6-473c-a388-eb4910dff436.mp4

### Console errors, if any

_No response_

### Example URL

### Browser(s) used

Chrome

cheap-glitch:
Fixes #4567 

## Test URLs

Navigate using PJAX, e.g.:
- Go from repo root to issues list
- Go from the issues list to an issue page
- Go from one issue to another using references (e.g. #4567)
- Go back on a page with the back button

All features should work properly, and features affecting the repo header (e.g `ci-link`, `quick-repo-deletion`) should be properly deduplicated.

## Screenshot

https://user-images.githubusercontent.com/46634000/125924186-566a7bee-65b0-4e89-96ab-4e2a409c239d.mp4

### DIFF
--- a/source/features/ci-link.tsx
+++ b/source/features/ci-link.tsx
@@ -35,5 +35,6 @@ void features.add(__filebasename, {
 		pageDetect.isEmptyRepo,
 	],
 	awaitDomReady: false,
+	deduplicate: 'has-rgh',
 	init,
 });
--- a/source/features/ci-link.tsx
+++ b/source/features/ci-link.tsx
@@ -35,6 +35,5 @@ void features.add(__filebasename, {
 		pageDetect.isEmptyRepo,
 	],
 	awaitDomReady: false,
-	deduplicate: 'has-rgh',
 	init,
 });
--- a/source/features/index.tsx
+++ b/source/features/index.tsx
@@ -24,7 +24,7 @@ interface FeatureLoader extends Partial<InternalRunConfig> {
 
 	/** When pressing the back button, DOM changes and listeners are still there, so normally `init` isn’t called again thanks to an automatic duplicate detection.
 	This detection however might cause problems or not work correctly in some cases #3945, so it can be disabled with `false` or by passing a custom selector to use as duplication check
-	@default undefined */
+	@default true */
 	deduplicate?: false | string;
 
 	/** When true, don’t run the `init` on page load but only add the `additionalListeners`. @default false */
@@ -217,8 +217,8 @@ const add = async (id: FeatureID, ...loaders: FeatureLoader[]): Promise<void> =>
 			exclude = [], // Default: nothing
 			init,
 			deinit,
-			deduplicate,
 			awaitDomReady = true,
+			deduplicate = 'has-rgh',
 			onlyAdditionalListeners = false,
 			additionalListeners = [],
 		} = loader;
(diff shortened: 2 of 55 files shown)


source/features/index.tsx
```diff
@@ -246,7 +246,19 @@ const add = async (id: FeatureID, ...loaders: FeatureLoader[]): Promise<void> =>
 		}
 
 		document.addEventListener('pjax:end', () => {
-			if (!deduplicate || !select.exists(deduplicate)) {
+			if (deduplicate === undefined) {
```
fregante, line 249: Let's flip these conditions. I think:

```
if is string && exists()
	load
if is false
	return
then the code you have below, unchanged:

if (select.exists('#repo-content-pjax-container')) {
if (!select.exists('has-rgh-inner')) {
	void setupPageLoad(id, details);
}
} else if (!select.exists('has-rgh')) {
void setupPageLoad(id, details);
}

File: source/features/ci-link.tsx [hunk 2]
```diff
@@ -35,5 +35,6 @@ void features.add(__filebasename, {
 		pageDetect.isEmptyRepo,
 	],
 	awaitDomReady: false,
+	deduplicate: 'has-rgh',
```
fregante, line 38: Why? This should be the default.

fregante (commented):
Awesome! Thanks to this we can also drop many of the custom `deduplicate` selectors we added for #3945

cheap-glitch:
> we can also drop many of the custom `deduplicate` selectors we added for #3945

Same PR?

source/features/ci-link.tsx
cheap-glitch: Since this feature affects the DOM outside of `#repo-content-pjax-container`, we want it to look at `has-rgh` only. Otherwise when `has-rgh-inner` is removed it will run again, even though the repo header hasn't changed.

fregante: Then my suggested logic doesn't work. The intent was to create logic that works for every feature, but if as you say features need to specify which deduplicator to use, then the logic is no longer sound.

fregante (changes requested):
I think everything here needs to be undone. 

In short the problem is that we receive a "page updated" event but we don't know which part updated and which elements are still on the page.

I can't merge this PR as is because it changes the default behavior for **all** the features. That needs testing for every feature and it's not practical.

For now we can take the same exact approach we took for https://github.com/sindresorhus/refined-github/issues/3945: Marking each currently-broken feature with a custom selector.

**However** this new selector can probably be `has-rgh-inner` instead of using one per feature.

> > we can also drop many of the custom `deduplicate` selectors we added for #3945
> 
> Same PR?

You're right that's not urgent. Later. We'll clean this all up after we fix Refined GitHub.

source/features/index.tsx
```diff
@@ -274,6 +291,7 @@ void add(__filebasename, {
 		// `await` kicks it to the next tick, after the other features have checked for 'has-rgh', so they can run once.
 		await Promise.resolve();
 		select('#js-repo-pjax-container, #js-pjax-container')?.append(<has-rgh/>);
+		select('#repo-content-pjax-container')?.append(<has-rgh-inner/>); // #4567
```
fregante, line 294: This line is probably the only one that needs to stay among these changes.

fregante:
Thoughts? Alternative solutions?

cheap-glitch:
It does work well as far as I've seen, but I do agree that changing the default for every feature is a risky move. I still thinks it makes sense as a default though.

For now I'll change this PR to add `deduplicate: 'has-rgh-inner'` to all the features that need it, so we can merge this knowing for sure we fixed more than we broke :smiley:

cheap-glitch:
Okay, I've combed through the output of
```shell
rg --files-without-match 'deduplicate:' source/features/*.tsx | sort
```
and quickly tested every feature I added the new deduplication to.

There might be a few tricky cases where this doesn't cut it entirely, e.g. I've noticed `bugs-tab` is partially broken (I'll open an issue about it). But otherwise this fixes most broken features (for now). Also, PR-related features don't seem affected by this new navigation (yet).

fregante:
Excellent! Version coming! Let’s see how it works in the field. Thanks for working on this!

PR #4005, opened 2021-02-21
Support more filenames in `link-to-changelog-file`

cheap-glitch:
1. LINKED ISSUES: Closes #3999 

2. TEST URLS:
 * [`CHANGELOG.md`](https://github.com/nodeca/js-yaml/releases/tag/4.0.0)
 * [`CHANGELOG.rst`](https://github.com/pyca/cryptography/releases)
 * [`CHANGES`](https://github.com/sphinx-doc/sphinx/releases)
 * [`news`](https://github.com/pypa/pip/releases)
 * [`HISTORY.md`](https://github.com/psf/requests/releases)
 * note: for every URL, there are two things to test:
   1. going directly to the release page
   2. going to the repo root (with a cleared cache), _then_ to the release page

PR #4578, opened 2021-07-17
`link-to-changelog-file`: Exclude non-textual files

### LINKED ISSUE #4577: False positive in `link-to-changelog-file`, linking to a `.js` file which is not the changelog
### Please ensure:

RGH adds the `CHANGELOG` button on the top of the [releases page of hyper](https://github.com/vercel/hyper/releases), with a link to the https://github.com/vercel/hyper/blob/HEAD/release.js page, which is not the changelog (I'm actually not sure whether there even is one, though...)

grafik

### Console errors, if any

_No response_

### Example URL

https://github.com/vercel/hyper/releases

### Browser(s) used

Firefox 90.0 (64-Bit)

darkred:
Closes #4577 (following https://github.com/sindresorhus/refined-github/issues/4577#issuecomment-881951763)

## Test URLs

Match: (as before)
https://github.com/pyca/cryptography/releases (`CHANGELOG.rst`)
https://github.com/nodeca/js-yaml/releases/tag/4.0.0  (`CHANGELOG.md`)

Not match: 
https://github.com/vercel/hyper/releases  (`release.js`)

## Screenshot

(none)

## Notes to reviewers

 - The markdown extension regex is from: [`vertical-front-matter`](https://github.com/sindresorhus/refined-github/blob/31a04dd8b43d4fe121bf5a467dbc0cdada568756/source/features/vertical-front-matter.tsx#L9) / [`view-markdown-source`](https://github.com/sindresorhus/refined-github/blob/31a04dd8b43d4fe121bf5a467dbc0cdada568756/source/features/view-markdown-source.tsx#L34)
    ```regex
    /\.(mdx?|mkdn?|mdwn|mdown|markdown|litcoffee)$/
    ```
    I added to it: `|txt|rst` to maintain matches from the previous related issue examples.

- I noticed that `findChangelogName()` was initially returning string or false, that's why the lengthy:

    ```js
    function findChangelogName(files: string[]): string | false {
    	const filename = files.find(name => regexChangelogMarkdownTextFiles.test(name))!;
    	const result = (typeof filename === 'undefined') ? false : filename;

    	return result;
    }
    ```

    Otherwise, if returning `false` can be omitted, I could replace it with:

    ```js
    function findChangelogName(files: string[]): string{
    	return files.find(name => regexChangelogMarkdownTextFiles.test(name))!;
    }
    ```

### DIFF
--- a/source/features/link-to-changelog-file.tsx
+++ b/source/features/link-to-changelog-file.tsx
@@ -16,14 +16,12 @@ interface FileType {
 
 const getCacheKey = (): string => `changelog:${getRepo()!.nameWithOwner}`;
 
-const regexChangelogMarkdownTextFiles = /^(changelog|news|changes|history|release|whatsnew)\.(mdx?|mkdn?|mdwn|mdown|markdown|litcoffee|txt|rst)$/i
+const regexChangelogMarkdownTextFiles = /^(changelog|news|changes|history|release|whatsnew)\.(mdx?|mkdn?|mdwn|mdown|markdown|litcoffee|txt|rst)$/i;
 function findChangelogName(files: string[]): string | false {
 	const filename = files.find(name => regexChangelogMarkdownTextFiles.test(name))!;
-	if (typeof filename !== 'undefined') {
-		return filename;
-	} else {
-		return false;
-	}
+	const result = (typeof filename === 'undefined') ? false : filename;
+
+	return result;
 }
 
 function parseFromDom(): false {
--- a/source/features/link-to-changelog-file.tsx
+++ b/source/features/link-to-changelog-file.tsx
@@ -16,12 +16,10 @@ interface FileType {
 
 const getCacheKey = (): string => `changelog:${getRepo()!.nameWithOwner}`;
 
-const regexChangelogMarkdownTextFiles = /^(changelog|news|changes|history|release|whatsnew)(\.(mdx?|mkdn?|mdwn|mdown|markdown|litcoffee|txt|rst))?$/i;
+const changelogFiles = /^(changelog|news|changes|history|release|whatsnew)(\.(mdx?|mkdn?|mdwn|mdown|markdown|litcoffee|txt|rst))?$/i;
 function findChangelogName(files: string[]): string | false {
-	const filename = files.find(name => regexChangelogMarkdownTextFiles.test(name));
+	const filename = files.find(name => changelogFiles.test(name));
 	return filename ?? false;
-
-	return result;
 }
 
 function parseFromDom(): false {
--- a/source/features/link-to-changelog-file.tsx
+++ b/source/features/link-to-changelog-file.tsx
@@ -17,13 +17,13 @@ interface FileType {
 const getCacheKey = (): string => `changelog:${getRepo()!.nameWithOwner}`;
 
 const changelogFiles = /^(changelog|news|changes|history|release|whatsnew)(\.(mdx?|mkdn?|mdwn|mdown|markdown|litcoffee|txt|rst))?$/i;
-function findChangelogName(files: string[]): string | undefined {
-	return files.find(name => changelogFiles.test(name));
+function findChangelogName(files: string[]): string | false {
+	return files.find(name => changelogFiles.test(name)) ?? false;
 }
 
 function parseFromDom(): false {
 	const files = select.all('[aria-labelledby="files"] .js-navigation-open[href*="/blob/"').map(file => file.title);
-	void cache.set(getCacheKey(), findChangelogName(files) ?? '');
+	void cache.set(getCacheKey(), findChangelogName(files));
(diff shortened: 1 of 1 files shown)


source/features/link-to-changelog-file.tsx
```diff
@@ -15,16 +15,13 @@ interface FileType {
...
 
+const regexChangelogMarkdownTextFiles = /^(changelog|news|changes|history|release|whatsnew)\.(mdx?|mkdn?|mdwn|mdown|markdown|litcoffee|txt|rst)$/i;
 function findChangelogName(files: string[]): string | false {
-	for (const file of files) {
-		if (changelogNames.has(file.toLowerCase().split('.', 1)[0])) {
-			return file;
-		}
-	}
+	const filename = files.find(name => regexChangelogMarkdownTextFiles.test(name))!;
+	const result = (typeof filename === 'undefined') ? false : filename;
```
cheap-glitch, line 22:
```suggestion
	return filename ?? false;
```

source/features/link-to-changelog-file.tsx
```diff
@@ -15,16 +15,13 @@ interface FileType {
...
-const changelogNames = new Set(['changelog', 'news', 'changes', 'history', 'release', 'whatsnew']);
 
+const regexChangelogMarkdownTextFiles = /^(changelog|news|changes|history|release|whatsnew)\.(mdx?|mkdn?|mdwn|mdown|markdown|litcoffee|txt|rst)$/i;
 function findChangelogName(files: string[]): string | false {
-	for (const file of files) {
-		if (changelogNames.has(file.toLowerCase().split('.', 1)[0])) {
-			return file;
-		}
-	}
+	const filename = files.find(name => regexChangelogMarkdownTextFiles.test(name))!;
```
cheap-glitch, line 21: I don't think you need the assertion here:
```suggestion
	const filename = files.find(name => regexChangelogMarkdownTextFiles.test(name));
```

source/features/link-to-changelog-file.tsx
```diff
@@ -15,16 +15,13 @@ interface FileType {
 }
 
 const getCacheKey = (): string => `changelog:${getRepo()!.nameWithOwner}`;
-const changelogNames = new Set(['changelog', 'news', 'changes', 'history', 'release', 'whatsnew']);
 
+const regexChangelogMarkdownTextFiles = /^(changelog|news|changes|history|release|whatsnew)\.(mdx?|mkdn?|mdwn|mdown|markdown|litcoffee|txt|rst)$/i;
```
cheap-glitch, line 19: There are changelogs with no extension (e.g. https://github.com/sphinx-doc/sphinx/blob/4.x/CHANGES)
```suggestion
const regexChangelogMarkdownTextFiles = /^(changelog|news|changes|history|release|whatsnew)(\.(mdx?|mkdn?|mdwn|mdown|markdown|litcoffee|txt|rst))?$/i;
```

Also the name could be shorter, it's not always Markdown and we know it's a regex by looking at it. Maybe just `changelogFiles`?

cheap-glitch (changes requested):

source/features/link-to-changelog-file.tsx
darkred: > it's not always Markdown

You've read my mind! I was going to ask about files that have no extension 👍 

> Maybe just `changelogFiles`?

Yes, 👍

cheap-glitch (approved):
Tested this with all the links from #4005 and #4071 :+1: :100:

source/features/link-to-changelog-file.tsx
```diff
@@ -15,16 +15,11 @@ interface FileType {
...
 function findChangelogName(files: string[]): string | false {
-	for (const file of files) {
-		if (changelogNames.has(file.toLowerCase().split('.', 1)[0])) {
-			return file;
-		}
-	}
-
-	return false;
+	const filename = files.find(name => changelogFiles.test(name));
+	return filename ?? false;
```
fregante, line 22: Let’s just change the function’s signature to return `string | undefined` so you can skip this intermediary variable and extra line

fregante (approved):

source/features/link-to-changelog-file.tsx
darkred: Done in the new commit.
I also provided an empty string as fallback value `?? ''` in the two function calls.
It should be ok.

yakov116 (approved):
LGTM

source/features/link-to-changelog-file.tsx
cheap-glitch: Nitpick but the fallback value should probably be `false` to keep the logic clear.

fregante:
Thanks @darkred! I reverted it to `false` because `undefined` clears the cache and `''` isn't a good falsey value.

PR #4590, opened 2021-07-20
Try CodeQL out

fregante:
No idea what this does, let's see…


fregante:
Apparently nothing! Our code is already purrfect

PR #4592, opened 2021-07-20
Include other bug labels in `bugs-tab`

### LINKED ISSUE #4452: Add support for labels other than "bug" in `bugs-tab`
I love `bugs-tab` but it doesn't show up https://github.com/nodejs/node, which I work on, because we use the [<kbd>confirmed-bug</kbd>](https://github.com/nodejs/node/issues?q=is%3Aissue+label%3Aconfirmed-bug+is%3Aopen) label for bugs, rather than `bug`.

Could the label name be configurable?  Or, could we list the labels and use ones that match `/\bbug\b/`? 🤔

yakov116:
Closes #4452 

## Test URLs
https://github.com/nodejs/node/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3Aconfirmed-bug
https://github.com/sindresorhus/refined-github/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3A%22bug%22
https://github.com/MithrilJS/mithril.js/issues?q=is%3Aissue+is%3Aopen+label%3A%22Type%3A+Bug%22+sort%3Aupdated-desc
https://github.com/yakov116/TestR/issues?q=is%3Aissue+is%3Aopen+label%3A%22%3Abug%3A+bug%22+sort%3Aupdated-desc

## Screenshot
image


fregante:
gif

yakov116:
OMG!! WOW

source/features/bugs-tab.tsx
```diff
@@ -24,16 +24,40 @@ async function highlightBugsTabOnIssuePage(): Promise<void | false> {
...
+		`);
+
+		const {name: bugLabel, issues} = [...repository.labels.nodes].find(label => [':bug: bug', 'bug', 'confirmed-bug', 'type: bug'].includes(label.name.toLowerCase()));
+		void cache.set(cacheKey(), bugLabel.includes(':') ? `"${bugLabel}"` : bugLabel ?? false);
+		return issues?.totalCount ?? 0;
+	}
+
+	const {repository} = await api.v4(`
+		repository() {
+			label(name: "${(await cache.get(cacheKey())).replaceAll('"', '')}") {
```
fregante, line 50: I like this api more than the search() API, it's shorter and more correct. Are they equivalent? If so, let's use `label()` instead of `search()`

yakov116: Only thing is that its case-sensitive. So if someone changes from "bug" to "Bug" it will not work

yakov116: > And definitely not 2 separate API calls

Why should we query for all labels every time.

Each time we will only use ONE api call.

The first time will load all labels the other times it will ONLY load a single one.

source/features/bugs-tab.tsx
```diff
@@ -24,16 +24,40 @@ async function highlightBugsTabOnIssuePage(): Promise<void | false> {
 	issuesTab.removeAttribute('aria-current');
 }
 
+const cacheKey = (): string => 'bugs-label' + ':' + getRepo()!.nameWithOwner;
 const countBugs = cache.function(async (): Promise<number> => {
-	const {search} = await api.v4(`
-		search(type: ISSUE, query: "label:bug is:open is:issue repo:${getRepo()!.nameWithOwner}") {
-			issueCount
+	if (!await cache.get(cacheKey())) {
```
yakov116, line 29: Not sure if you missed this line

fregante: What this line does is cause 2 API calls for every single repo and that's exactly why that issue was closed the first time it was asked.

yakov116: > What this line does is cause 2 API calls for every single repo

Why?

source/features/bugs-tab.tsx
fregante: Looking good! Proceed 😁

yakov116: Ok the only issue I am having is the `isBugPage` encoding with `"` But I will try to work on it myself

source/features/bugs-tab.tsx
```diff
@@ -24,16 +24,40 @@ async function highlightBugsTabOnIssuePage(): Promise<void | false> {
...
+						issues(states: OPEN) {
+							totalCount
+						}
+					}
+				}
+			}
+		`);
+
+		const {name: bugLabel, issues} = [...repository.labels.nodes].find(label => [':bug: bug', 'bug', 'confirmed-bug', 'type: bug'].includes(label.name.toLowerCase()));
+		void cache.set(cacheKey(), bugLabel.includes(':') ? `"${bugLabel}"` : bugLabel ?? false);
```
fregante, line 44: _I know this is a draft PR_ 😁 

Let's always wrap it in quotes, but not here, we should store the pure label name so it can be potentially reused in different contexts (where `"` don't apply, for example)

Also I think `?? false` is mispositioned here? If it's not found, `find` will return `undefined` and the destructuring on the same line will fail.

source/features/bugs-tab.tsx
fregante: > Only thing is that its case-sensitive. So if someone changes from "bug" to "Bug" it will not work


That's the case for all of our cached content. If the content changes, they'll not see the right data for a while.

source/features/bugs-tab.tsx
yakov116: I know 👍.

The issue is that when searching for a bug page it will only match when using RG bugs-tab.

So we need to match GitHub search.

The code needs a real cleanup, so I will work on it.

fregante:
This feature keeps getting longer and longer…. 😅

yakov116:
Buggy code 😏

source/features/bugs-tab.tsx
```diff
@@ -24,14 +31,45 @@ async function highlightBugsTabOnIssuePage(): Promise<void | false> {
 	issuesTab.removeAttribute('aria-current');
 }
 
+const cacheKey = (): string => 'bugs-label:' + getRepo()!.nameWithOwner;
```
fregante, line 34: Cache key of what? This needs to be more specific

source/features/bugs-tab.tsx
```diff
@@ -24,14 +31,45 @@ async function highlightBugsTabOnIssuePage(): Promise<void | false> {
...
+
 const countBugs = cache.function(async (): Promise<number> => {
-	const {search} = await api.v4(`
-		search(type: ISSUE, query: "label:bug is:open is:issue repo:${getRepo()!.nameWithOwner}") {
-			issueCount
+	if (!await cache.get(cacheKey())) {
```
fregante, line 37: You're getting this value several times in this function. Extract it to its own variable. Also you can use <string> to specify its type instead of String()

source/features/bugs-tab.tsx
```diff
@@ -24,14 +31,45 @@ async function highlightBugsTabOnIssuePage(): Promise<void | false> {
...
+
+		const bugTypes = new Set([':bug: bug', 'bug', 'confirmed-bug', 'type: bug']);
+		const {name: bugLabel, issues} = nodes.find((label: LabelInfo) => bugTypes.has(label.name.toLowerCase())) ?? {};
+		void cache.set(cacheKey(), bugLabel ?? false);
+		return issues?.totalCount ?? 0;
+	}
+
+	const {repository} = await api.v4(`
+		repository() {
+			label(name: "${String((await cache.get(cacheKey())))}") {
```
fregante, line 64: Needs a few more parens ☺️

source/features/bugs-tab.tsx
```diff
@@ -47,7 +85,9 @@ async function init(): Promise<void | false> {
 	// - update the count later
 	// On other pages:
 	// - only show the tab if needed
-	const isBugsPage = new SearchQuery(location.search).includes('label:bug');
+	const bugLabelKey = String(await cache.get(cacheKey()))!;
+	const bugLabel = bugLabelKey.includes(':') ? `"${bugLabelKey}"` : bugLabelKey;
```
fregante, line 89: I'd like to see this turned into a helper inside SearchQuery itself, maybe a static method. Example:

``` JavaScript 
SearchQuery.fromEntries([
  ['label', label],
]) === 'label:"the bug"'
```

This would also work with multiple entries of course. It needs to use entries instead of an object because a query can have multiple `is` queries and could just be 

``` JavaScript 
SearchQuery.fromEntries([
  ['label', label],
  ['my other keyword']
]) === 'label:"the bug" my other keyword'
```

source/features/bugs-tab.tsx
```diff
@@ -11,6 +11,13 @@ import {getRepo} from '../github-helpers';
 import SearchQuery from '../github-helpers/search-query';
 import abbreviateNumber from '../helpers/abbreviate-number';
 
+interface LabelInfo {
```
fregante, line 14: Please stop typing API responses. This is another example of why it should not be done: it's wrong.

source/features/bugs-tab.tsx
yakov116: I need a drop more guidance of how to implement.

fregante: probably:

```js
entries.map(([base, value]) => {
  if (value === undefined) {
    return base
  }

  const escapedValue = /[:\s]/.test(value) ? `"${value}"` : value;
  return base + ':' + escapedValue
})
return entries.join(' ')

File: source/features/bugs-tab.tsx [hunk 9]
```diff
@@ -24,14 +24,46 @@ async function highlightBugsTabOnIssuePage(): Promise<void | false> {
...
 }
 
+const bugLabelCacheKey = (): string => 'bugs-label:' + getRepo()!.nameWithOwner;
+const bugLabel = async (): Promise<string | undefined> => cache.get<string>(bugLabelCacheKey());
+
 const countBugs = cache.function(async (): Promise<number> => {
-	const {search} = await api.v4(`
-		search(type: ISSUE, query: "label:bug is:open is:issue repo:${getRepo()!.nameWithOwner}") {
-			issueCount
+	if (!await bugLabel()) {
```
fregante, line 31: No?

```suggestion
	const label = cache.get<string>(bugLabelCacheKey());
	if (!label) {
```

yakov116: I need it for the next function

source/features/bugs-tab.tsx
yakov116: Did I do it right?

source/features/bugs-tab.tsx
```diff
@@ -47,7 +79,7 @@ async function init(): Promise<void | false> {
 	// - update the count later
 	// On other pages:
 	// - only show the tab if needed
-	const isBugsPage = new SearchQuery(location.search).includes('label:bug');
+	const isBugsPage = new SearchQuery(location.search).includes(`label:${await bugLabel()}`);
```
yakov116, line 82: we have an issue here that bugLabel might not be ready yet

yakov116:
@fregante I think this maybe over me. (I don't want to throw it on you). Can I just add just the ones without a space and make a new issue for the other ones?

fregante:
I'll look into it when I have time

source/features/bugs-tab.tsx
fregante: Not sure I follow. From what I see, if it doesn't exist in the cache, we enter a block that sets both the bug count and bug label into the cache and then it's done. 

If it *does* exist, it's just used as part of the label-specific API call. 

I also don't get when it could be "not ready" as you mentioned in the other comment

Milo123459:
2 requests:
`🐛bug`

`🐞bug`

Milo123459:
^ Oh and include spaces versions, like `🐞bug` and `🐞 bug`

source/features/bugs-tab.tsx
```diff
@@ -24,14 +24,46 @@ async function highlightBugsTabOnIssuePage(): Promise<void | false> {
...
+			}
+		`);
+
+		const {nodes}: AnyObject = repository.labels;
+		if (nodes.length === 0) {
+			return 0;
+		}
+
+		const bugTypes = new Set([':bug: bug', 'bug', 'confirmed-bug', 'type: bug']);
+		const {name, issues} = nodes.find((label: AnyObject) => bugTypes.has(label.name.toLowerCase())) ?? {};
```
fregante, line 51:
```suggestion
		const bugTypes = new Set([':bug:bug', 'bug', 'confirmed-bug', 'type:bug']);
		const {name, issues} = nodes.find((label: AnyObject) => bugTypes.has(label.name.toLowerCase().replace(/\s/g, ''))) ?? {};	
```

Or change it to a regex: `/^(:bug:bug|bug|confirmed-bug|type:bug)$/` or something long those lines, we don't need to make it too complex

source/features/bugs-tab.tsx
yakov116: https://github.com/sindresorhus/refined-github/blob/517cc5d0e27d9b7587f01aa7a26fd871daf49c62/source/features/bugs-tab.tsx#L77-L82

This line runs as soon as the page loads, but before `countPromise` is done.

Also that is the line that I need it again

source/features/bugs-tab.tsx
yakov116: > we don't need to make it too complex

Maybe we do a regex to include `:`word`:`(optinalSpace)`bug` like this we can include all emoji  bug labels?

fregante: The `.replace(/\s/g, '')` part already does away with all the spaces, that's why my suggestions don't have it.

As for the `:\w+:` emoji, sure

fregante:
The linked repos are not exactly what I meant by "the most popular repos." 😃 Please make sure that we support the top 10/25 GitHub repos (no need to link them here), unless they can't be supported

source/features/bugs-tab.tsx
fregante:
```suggestion
	const label = await bugLabel();
	if (!label) {
```

source/features/bugs-tab.tsx
```diff
@@ -24,14 +24,46 @@ async function highlightBugsTabOnIssuePage(): Promise<void | false> {
...
+		void cache.set(bugLabelCacheKey(), name ?? false);
+		return issues?.totalCount ?? 0;
+	}
+
+	const {repository} = await api.v4(`
+		repository() {
+			label(name: "${await bugLabel()!}") {
```
fregante, line 58: And then here
```suggestion
			label(name: "${label}") {
```

source/features/bugs-tab.tsx
```diff
@@ -24,14 +24,46 @@ async function highlightBugsTabOnIssuePage(): Promise<void | false> {
 	issuesTab.removeAttribute('aria-current');
 }
 
+const bugLabelCacheKey = (): string => 'bugs-label:' + getRepo()!.nameWithOwner;
+const bugLabel = async (): Promise<string | undefined> => cache.get<string>(bugLabelCacheKey());
```
fregante, line 28: Let's make it clear that these are getters

```suggestion
const getBugLabelCacheKey = (): string => 'bugs-label:' + getRepo()!.nameWithOwner;
const getBugLabel = async (): Promise<string | undefined> => cache.get<string>(getBugLabelCacheKey());
```

yakov116:
> I'll look into it when I have time. But yet maybe you can revert the change where you're having trouble

@fregante ready.

Right now the only thing that is broken is the `.add` it needs `"`

source/github-helpers/search-query.ts
```diff
@@ -120,6 +120,6 @@ export default class SearchQuery {
 	}
 
 	includes(...searchStrings: string[]): boolean {
-		return this.getQueryParts().some(queryPart => searchStrings.includes(queryPart));
+		return this.getQueryParts().some(queryPart => searchStrings.includes(queryPart) || searchStrings.includes(queryPart.replace(/"/g, '')));
```
fregante, line 123: This is the wrong behavior and the test below should be dropped too. This suggestion was not meant to be part of SearchQuery, but honestly I can't think of another way to do this:

```ts
const isBugsPage = new SearchQuery(location.search).includes(`label:${await getBugLabel() ?? 'bug'}`);
```

without:

- adding support for quoted properties in `SearchQuery`
- adding a "quoter" helper to easily wrap `getBugLabel`, like:
	```ts
	`label:${SearchQuery.escapeValue(await getBugLabel() ?? 'bug'))}`
	```

`SearchQuery.escapeValue` is really simple:

```ts
static escapeValue(value: string) {
	return /:\s/.test(value) ? `"${value}"` : value;
}
```

This PR cannot be merged as is. It's probably best to take care of this in separate PR and then  we can come back here

fregante (changes requested):

source/github-helpers/search-query.ts
yakov116: Curious why does it have to be part of SearchQuery?

fregante: They both affect the search query, right?

yakov116: Sorry I was not clear.

My question was why can it not just be an export from that file just like we do with the `api.escapeKey`?


https://github.com/sindresorhus/refined-github/blob/f908677216e84f4a238a69f0573a257945ea6a25/source/features/release-download-count.tsx#L27

fregante: In practice there's no difference since there's no tree-shaking. Since SearchQuery is structured as a class I think it's easier to have a `SearchQuery.escapeValue` method instead of exporting `escapeValue` on its own, which wouldn't make sense and would have to be named `escapeSearchQueryValue` 🤷‍♂️ 😅

source/features/bugs-tab.tsx
```diff
@@ -24,14 +29,49 @@ async function highlightBugsTabOnIssuePage(): Promise<void | false> {
...
+					issues(states: OPEN) {
+						totalCount
+					}
+				}
+			}
+		}
+	`);
+
+	const label: AnyObject | undefined = repository.labels.nodes
+		.find((label: AnyObject) => isBugLabel(label?.name));
```
fregante, line 47: 👍 

But I assume `label` exists, right? I don’t think `nodes` would be an array of `undefined`s
```suggestion
		.find((label: AnyObject) => isBugLabel(label.name));
```

Other than this the commit look good, it's missing the second part (which would solve this TODO):

https://github.com/sindresorhus/refined-github/blob/ae247fb90a4c4b20524d76c823e97af3881326fb/source/github-helpers/search-query.ts#L88

The regex to fix is this, but it might be a little difficult 😅 

https://github.com/sindresorhus/refined-github/blob/ae247fb90a4c4b20524d76c823e97af3881326fb/source/github-helpers/search-query.ts#L5-L8

source/github-helpers/search-query.ts
yakov116: Got it

source/features/bugs-tab.tsx
yakov116: > But I assume `label` exists, right? I don’t think `nodes` would be an array of `undefined`s

No what if there are no bug labels?
it will return
```graphQL
    "repository": {
      "labels": {
        "nodes": []
      },
```

yakov116: > Other than this the commit look good, it's missing the second part (which would solve this TODO):
> 
> https://github.com/sindresorhus/refined-github/blob/ae247fb90a4c4b20524d76c823e97af3881326fb/source/github-helpers/search-query.ts#L88
> 
> The regex to fix is this, but it might be a little difficult 😅

Lost you here

fregante: > No what if there are no bug labels?

That's an empty array, it means it will never call the `.find` callback:

```js
[].find(console.log)
```

fregante: > Lost you here

I'm talking about this: > adding support for quoted properties in SearchQuery

i.e. fixing that TODO comment

fregante: In short this does not work now:

```js
new SearchQuery('label:"my name"').includes('label:"my name"')
```


because:

```js
'label:"my name"'
// is split into
['label:"my', 'name"'`]
// so
['label:"my', 'name"'`].includes('label:"my name"') === false

Comment (fregante) [reply source/features/bugs-tab.tsx:47]: This might be the solution, but we also need to support quote-less values: https://stackoverflow.com/q/16261635/288906

Comment (yakov116) [reply source/features/bugs-tab.tsx:47]: > is split into
> 
> ```js
> ['label:"my', 'name"'`]
> ```

Why do you say so?

https://regex101.com/r/FBLn1f/1 looks fine to me.

I also tested the feature and its working perfectly.

The only issue I see is with unquoted searches `label:my name`. But in theory that should never happen except on a manual search. But anyways GitHub search will not work if the value is not quoted.

File: source/github-helpers/search-query.ts [hunk 16]
```diff
@@ -54,6 +54,10 @@ export default class SearchQuery {
 		this.set(this.get());
 	}
 
+	static escapeValue(value: string): string {
+		return /\s|^:/.test(value) ? `"${value}"` : value;
```
yakov116, line 58: Changed is since GitHub will add quotes if it starts with a `:` or if it contains a space you can test it https://github.com/yakov116/TestR/labels

fregante: Doesn't look accurate: https://github.com/fregante/sandbox/issues?q=is%3Aopen+sort%3Aupdated-desc+label%3A%3Aa%3A

I think it only happens for spaces, but there might be other characters as well. It might be worth creating new labels to test this out. You can do so in a test repo from the issue sidebar

yakov116: I did you want more tests?

source/features/bugs-tab.tsx
fregante: I say so because the comment in the code says so but it looks like https://github.com/sindresorhus/refined-github/pull/3837 fixed it 🎉 

so perhaps this PR is ready once you delete that comment 

https://github.com/sindresorhus/refined-github/blob/ae247fb90a4c4b20524d76c823e97af3881326fb/source/github-helpers/search-query.ts#L88

yakov116: Should I add a test?

source/github-helpers/search-query.ts
fregante: I mean manual tests on GitHub.com to determine what characters triggers the quotes. From what I see, currently this should be:
```suggestion
		return value.includes(' ') ? `"${value}"` : value;
```

yakov116: But what about

https://github.com/yakov116/TestR/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3A%22t+%3A%2B1%3A%22 ??

yakov116: maybe if it has an emojie?

yakov116: Great idea lets include!

https://github.com/markdown-it/markdown-it-emoji/blob/339abf72f84a02e66fe9a4d682a7217d1f660e6a/dist/markdown-it-emoji.js#L9-L1848

fregante: … that label has a space

fregante: Also emojis are fine: https://github.com/fregante/sandbox/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3A😃m

fregante:
Uhhhh this could have been useful 2 weeks ago 😂 https://github.blog/changelog/2021-08-02-search-issues-by-label-using-logical-or/

Might be worth updating the feature? The current situation is still better for the user though since:

- only one label appears in the query
- we can support any label with regex

So we're good without it anyway 😁

yakov116:
Yup

mcornella:
This doesn't seem to support uppercase characters in the label for me. Shouldn't the regex end with `/i`?

yakov116:
> This doesn't seem to support uppercase characters in the label for me. Shouldn't the regex end with `/i`?

Can you give a real url?

mcornella:
[This one from the original post](https://github.com/MithrilJS/mithril.js/issues?q=is%3Aissue+is%3Aopen+label%3A%22Type%3A+Bug%22+sort%3Aupdated-desc) doesn't seem to work for me:

Captura de pantalla de 2021-09-23 19-14-46

But this repo works:

Captura de pantalla de 2021-09-23 19-15-30

yakov116:
Thanks
Can you open a new issue?

mcornella:
I've opened #4811. I tried the `/i` in `supportedLabels` and it looks like it solves it.

yakov116:
You can send a PR

PR #4608, opened 2021-07-25
Stop linking avatars in `reactions-avatars`

fregante:
I think that having links in here doesn't make much sense and it's just problematic (the browser will _occasionally(!)_ follow the link when clicking the button) so let's just drop it.

~I should probably also replace the A with a SPAN, that requires changes in the CSS file as well~ _Done_

## Test URLs

This PR

### Before feature
Screen Shot 6

### Loading 
Screen Shot 4

### Loaded
Screen Shot 7

### DIFF
--- a/source/features/reactions-avatars.tsx
+++ b/source/features/reactions-avatars.tsx
@@ -59,9 +59,9 @@ async function showAvatarsOn(commentReactions: Element): Promise<void> {
 
 	for (const {container, imageUrl} of flatParticipants) {
 		container.append(
-			<a className="rounded-1 avatar-user">
+			<span className="rounded-1 avatar-user">
 				<img src={imageUrl} className="avatar-user rounded-1"/>
-			</a>,
+			</span>,
 		);
 	}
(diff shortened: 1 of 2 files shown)


source/features/reactions-avatars.tsx
```diff
@@ -58,12 +57,11 @@ async function showAvatarsOn(commentReactions: Element): Promise<void> {
 		.map(getParticipants);
 	const flatParticipants = flatZip(participantByReaction, avatarLimit);
 
-	for (const {container, username, imageUrl} of flatParticipants) {
+	for (const {container, imageUrl} of flatParticipants) {
 		container.append(
-			// Without this, Firefox will follow the link instead of submitting the reaction button
-			<a href={isFirefox ? undefined : `/${username}`} className="rounded-1 avatar-user">
+			<span className="rounded-1 avatar-user">
```
kidonng, line 62: Can we add a `title={username}` here? Guess it won't hurt

Issue #3686, opened 2020-10-25
Setup auto-publishing for Safari version


fregante:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%24200.01%20Funded-%2300A156.svg" />](https://issuehunt.io/r/sindresorhus/refined-github/issues/3686)

<details> <summary>Outdated intro</summary>

Refined GitHub is now also compatible with Safari (https://github.com/sindresorhus/refined-github/pull/3677), but it's not available on the App Store nor it's signed.

## Safari users 👋

If you're looking to use Refined GitHub for Safari, for the time being it's best to use https://github.com/lautis/refined-github-safari

</details>

## Xcode developers 👋

We need your help to:

- [ ] Configure Fastlane: https://docs.fastlane.tools/getting-started/ios/appstore-deployment/
- [ ] Setup the FastLane deployment action: https://github.com/marketplace/actions/fastlane-action

###  How to build

1. Install XCode 12 and Safari 14
2. In Safari, enable **Allow Unsigned Extensions** in the **Develop** menu.
3. Run:

	```sh
    npm install
	npm run build
	npm run pack:safari
	npm run start:safari
	```

4. In Safari, click on the extension icon while on GitHub.com to enable it.

You could also sign it locally following the instructions below: https://github.com/sindresorhus/refined-github/issues/3686#issuecomment-747455258

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $200.01)

- [<img src='https://avatars2.githubusercontent.com/u/6824881?v=4' alt='darthshadow' width=24 height=24> darthshadow](https://issuehunt.io/u/darthshadow) ($10.00)
- [<img src='https://avatars.githubusercontent.com/u/1402241?v=4' alt='fregante' width=24 height=24> fregante](https://issuehunt.io/u/fregante) ($50.01)
- [<img src='https://avatars.githubusercontent.com/u/170270?v=4' alt='sindresorhus' width=24 height=24> sindresorhus](https://issuehunt.io/u/sindresorhus) ($100.00)
- $40.00 have been anonymously funded.

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/sindresorhus/refined-github/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

fregante:
[...]

PR #4609, opened 2021-07-25
Disable `quick-mention` on hidden comments

### LINKED ISSUE #4603: `quick-mention` should be disabled on hidden comments
### What I don't like

 
Screen Shot 3

URL: https://github.com/sindresorhus/refined-github/issues/3686#ref-issue-864796824

It's ok if we only avoid adding the button on first page load

kidonng:
Resolve #4603

## Test URLs

https://github.com/sindresorhus/refined-github/issues/3686#issuecomment-734784717

## Screenshot

Before:

image

After:

image

### DIFF
--- a/source/features/quick-mention.tsx
+++ b/source/features/quick-mention.tsx
@@ -31,6 +31,10 @@ function init(): void {
 	// The hovercard attribute avoids `highest-rated-comment`
 	const avatars = select.all(`.TimelineItem-avatar > [data-hovercard-type="user"]:first-child:not([href="/${getUsername()!}"], .rgh-quick-mention)`);
 	for (const avatar of avatars) {
+		if (select.exists('.minimized-comment', avatar.parentElement!.nextElementSibling!)) {
+			continue;
+		}
+
 		const userMention = select('img', avatar)!.alt;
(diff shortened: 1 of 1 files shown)


source/features/quick-mention.tsx
```diff
@@ -31,6 +31,10 @@ function init(): void {
 	// The hovercard attribute avoids `highest-rated-comment`
 	const avatars = select.all(`.TimelineItem-avatar > [data-hovercard-type="user"]:first-child:not([href="/${getUsername()!}"], .rgh-quick-mention)`);
 	for (const avatar of avatars) {
+		if (select.exists('.minimized-comment', avatar.parentElement!.nextElementSibling!)) {
```
fregante, line 34: Let's just query the common ancestor, I think:

```suggestion
		if (avatar.closest('.TimelineItem')!.querySelector('.minimized-comment')) {
```

👆 also I try to avoid `select()` when doing traversing because it has an unnatural reading order:


```
1.                       avatar
2.                              .parentElement
3.                                             .nextElementSibling
4. '.minimized-comment'

fregante:
Eggcellent! Thank you

PR #546, opened 2017-06-27
Markdown copy improvements

bfred-it:
Mostly fixes #522. There's only [one left](https://github.com/sindresorhus/refined-github/issues/522#issuecomment-310547556) and that's hard to fix.

Edit: **Automated tests have been [added](https://github.com/sindresorhus/refined-github/pull/546/commits/75ad7c36a28fdc7827ee67e32c95a923f227e168) ⚡️**

## Manual tests

✅ Drop `g-emoji` elements: 🔥🔥🔥
✅ Drop tasks from lists
 - [ ] try me out
 - [x] test across lines

✅ Drop autolinks around images: 
✅ Keep `` if it has width, height or align: copy
✅ Unwrap commit/issue autolinks: #522 833d5984fffb18a44b83d965b397f82e0ff3085e
✅ Unwrap **some** shortened links: https://www.npmjs.com https://twitter.com/bfred_it https://github.com/

✅ Wrap orphaned `li`s in their original parent:

- try selecting text from ➡️ here
- to ⬅️ here
- before this, the copied text would have been "➡️here*   to⬅️" without linebreaks

✅ Add support for `ol` lists that start after 1:

99. big lists
100. deserve big numbers

✅ Keep the original number of orphaned `li`s in numbered lists:

1. If you start copying
2. some text from ➡️here
3. to ⬅️ here
4. the copied text should start with `2. here` not, `1. here`

PR #4613, opened 2021-07-26
Prevent `linkify-code` from running multiple times

### LINKED ISSUE #3226: `linkify-code` runs multiple times on the same URL

kidonng:
Fix #3226

## Test URLs

From #2066:

> URLs in diffs
> [Homebrew/homebrew-core#18015 (files)](https://github.com/Homebrew/homebrew-core/pull/18015/files)
> 
> URLs in regular files
> [/source/features/copy-markdown.js@`3f5fc48`#L45](https://github.com/sindresorhus/refined-github/blob/3f5fc489e417d4f4a14da5ea423775e9ca9246fd/source/features/copy-markdown.js#L45)
> 
> URLs in PR files
> [#546 (files)](https://github.com/sindresorhus/refined-github/pull/546/files#diff-36280fbf56b6409696e3d46d6b30118fR51)
> 
> URLs in embedded code
> https://github.com/sindresorhus/refined-github/blob/3f5fc489e417d4f4a14da5ea423775e9ca9246fd/source/features/copy-markdown.js#L45
> 
> Issue reference in file
> 
> [/source/features/sort-issues-by-update-time.tsx@`2ade9a8`#L18](https://github.com/sindresorhus/refined-github/blob/2ade9a84b1894c7879fab9d3e2d045e876f941a6/source/features/sort-issues-by-update-time.tsx#L18)
> 
> Issue reference in embedded code
> 
> https://github.com/sindresorhus/refined-github/blob/2ade9a84b1894c7879fab9d3e2d045e876f941a6/source/features/sort-issues-by-update-time.tsx#L18

---

From #3226:

- Review comment on Conversation tab: https://github.com/sindresorhus/refined-github/pull/3225#discussion_r439535538
- Review comment on Files changed tab: https://github.com/sindresorhus/refined-github/pull/3589/files#r493534691

## Screenshot

Before:
image

After:
image

### DIFF
--- a/source/features/linkify-code.tsx
+++ b/source/features/linkify-code.tsx
@@ -17,8 +17,8 @@ function initTitle(): void {
 
 function init(): void {
 	const selectors = [
-		'.js-blob-wrapper tr:not(.inline-comments)', // `isPRFiles`, `isSingleFile`
-		'.blob-wrapper tr', // `hasComments`, `isPRFiles` review suggestion
+		'.js-blob-wrapper tr:not(.inline-comments)', // file blocks in pages like `isPRFiles`, `isSingleFile`
+		'.blob-wrapper tr', // every other code blocks
 	].map(selector => selector + `:not(.${linkifiedURLClass})`).join(',');
 
 	observe(selectors, {
--- a/source/features/linkify-code.tsx
+++ b/source/features/linkify-code.tsx
@@ -17,9 +17,8 @@ function initTitle(): void {
 
 function init(): void {
 	const selectors = [
-		'.blob-wrapper',
-		'.comment-body.d-block',
-		'.blob-expanded',
+		'.js-blob-wrapper tr:not(.inline-comments)',
+		'.blob-wrapper tr',
 	].map(selector => selector + `:not(.${linkifiedURLClass})`).join(',');
 
 	observe(selectors, {
--- a/source/features/linkify-code.tsx
+++ b/source/features/linkify-code.tsx
@@ -17,8 +17,8 @@ function initTitle(): void {
 
 function init(): void {
 	const selectors = [
-		'.js-blob-wrapper tr:not(.inline-comments)',
-		'.blob-wrapper tr',
+		'.js-blob-wrapper tr:not(.inline-comments)', // `isPRFiles`, `isSingleFile`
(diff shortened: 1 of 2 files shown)


source/features/linkify-code.tsx
```diff
@@ -17,7 +17,6 @@ function initTitle(): void {
 
 function init(): void {
 	const selectors = [
-		'.js-blob-wrapper',
```
kidonng, line 20: This is added in https://github.com/sindresorhus/refined-github/pull/2066/files#diff-49baf2e54681813b16cfb0a3b15c468590e5f5ee7c584af83fecd0beb5856238R47 but honestly I don't understand the reason. Is there any `.js-blob-wrapper` that isn't included in `.blob-wrapper`?

fregante: There might be 2 reasons for that change (I don't remember):

- backwards compatibility, maybe one selector was left in the code but it wasn't needed
- to support one of the many situations mentioned in that post, so if you remove it make sure the feature still works in every one of those

kidonng:
Explaining the selector changes:

- `.comment-body` is dropped because all code inside should also be included in `.blob-wrapper`
  
  The `.d-block` part is from a previous fix but we don't need it anymore

- `.blob-expanded` is dropped because it is covered by `.blob-wrapper tr`
- We are now using more specific selectors to select `tr` (`.js-blob-wrapper tr` and `.blob-wrapper tr`), because code blocks inside comments are also inside file blocks, but `linkifiedURLClass` will prevent the linkifying from running twice.
- We still need the `:not(.inline-comments)` part for `.js-blob-wrapper tr` because code blocks inside `.inline-comments` is already handled by `.blob-wrapper tr`.

fregante:
Thanks! Could you add a few comments explaining where each selector runs, like "isSingleFile, isPRFiles review, etc"

kidonng:
I don't think it make much sense to list every page there so keep in mind that `.js-blob-wrapper` is for file blocks, `.blob-wrapper` for anything else and that's it.

fregante:
Thanks!

image

Issue #4596, opened 2021-07-21
Summer feature freeze


fregante:
We keep adding new features but some existing features are partially or fully broken. Let’s favor bugfixes over new features for the rest of the summer. 

- [bugs (label)](https://github.com/sindresorhus/refined-github/issues?q=is%3Aissue+is%3Aopen+label%3Abug+sort%3Acreated-asc)
- [please ♥ (label)](https://github.com/sindresorhus/refined-github/issues?q=is%3Aissue+is%3Aopen+label%3A%22Please%21+♥%EF%B8%8E%22+sort%3Areactions-%2B1-desc)
- [meta (label)](https://github.com/sindresorhus/refined-github/issues?q=is%3Aopen+label%3Ameta+sort%3Areactions-%2B1-desc)
- issues meant to **slightly** extend/improve/refactor existing features might also be ok

![Stop, summer time](https://user-images.githubusercontent.com/1402241/126462612-f280eaf0-400a-4eab-8fe3-1f434ed1b71f.jpg)

We can go back to adding new features when we bring the number of bugs to a more reasonable level ☀️

kidonng:
From `2021-07-22T00:00:00.000Z` to the post time of this comment, **172 issues** created during this time has been closed and **144 pull requests** created during this time has been merged.

We have been **closing issues 1.56 times faster** `(172 / 3) / (2504 / 68)` and **merging pull requests 1.67 times faster** `(144 / 3) / (1951 / 68)`!

<details>
<summary>Data source</summary>

Query:

```graphql
{
  repository(owner: "refined-github", name: "refined-github") {
    issues(last: 100, states: [CLOSED], filterBy: {since: "2021-07-22T00:00:00.000Z"}) {
      nodes {
        createdAt
      }
    }
    pullRequests(last: 100, states: [MERGED]) {
      nodes {
        createdAt
      }
    }
  }
}
```

Filter:

```js
start = new Date('2021-07-22T00:00:00.000Z').getTime()
res.data.repository.issues.nodes.filter(i => start < new Date(i.createdAt).getTime()).length
res.data.repository.pullRequests.nodes.filter(i => start < new Date(i.createdAt).getTime()).length
```

</details>

fregante:
It was a long and productive summer my friends. Refined GitHub has been under attack from GitHub’s continuous breaking changes 😅 

We can continue adding features, but the bar to entry should be a bit higher since supporting many _complex_ features will just kill the extension altogether in the long run.

[...]

PR #4624, opened 2021-07-30
Feat/2673 load rich diff for images

### LINKED ISSUE #2673: Default to rich diffs for SVG files
The feature existed but was broken. More info found in https://github.com/sindresorhus/refined-github/issues/2041

Removed in https://github.com/sindresorhus/refined-github/commit/e88a05df70ba76134a9b7e2331b18b999291927b, you can probably revert the commit and rewrite the loading behavior as suggested. 

Originally asked in https://github.com/sindresorhus/refined-github/issues/1599

> When SVG files are edited, usually the rich diff (rendered image) is more useful/desirable than textual diff.
> 
> This is particularly useful when such files are edited in an editor rather than by hand, since the changes are usually extensive enough to go over the size limit for showing the diff automatically, so neither the code nor the rich diff are shown by default (the "Load diff" link is shown instead). See for example pretty much any PR at icon repos, such as [elementary/icons/pulls](https://github.com/elementary/icons/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc).

Perni1984:
**1. Does this PR close/fix an existing issue? Write something like `Closes #10`**

Closes #2673  - even more it is configured to load rich diffs for all image types supported by github as stated in https://docs.github.com/en/github/managing-files-in-a-repository/working-with-non-code-files/rendering-and-diffing-images

**2. What pages does this PR affect? Include some REAL URLs where you tested the code**

This PR automatically loads the first 20 (configurable) rich diffs for images in the PR-Files and commit tab. If there are more files a load more button is introduced after the last loaded rich diff. I built this solution because I experienced some loading problems when loading more than 20 rich-diffs simultaneously in my PRs.

Test-Urls:
https://github.com/pcode-at/pcode-web-next/pull/153/files
https://github.com/xcp-ng/xcp-ng-release/pull/7/files

**3. Open the PR as draft and review it yourself first. Fix what you find and explain weird code, if necessary.**

I was having some timing problems with large PRs and include_fragments (I think so) so I just made a hacky "solution" with window.setTimeout(run, 1000); - I would need some guidance what would be a more approriate so everything needed is loaded and ready. cc @fregante 

## Test URLs
https://github.com/xcp-ng/xcp-ng-release/pull/7/files
https://github.com/pcode-at/pcode-web-next/pull/153/files

## Screenshot

image

### DIFF
--- a/source/features/rich-diff-for-images.tsx
+++ b/source/features/rich-diff-for-images.tsx
@@ -1,0 +1,133 @@
+import * as pageDetect from "github-url-detection";
+
+import features from ".";
+import fetchDom from "../helpers/fetch-dom";
+import delegate from "delegate-it";
+
+const MAX_IMAGES_TO_BE_FETCHED_CONCURRENTLY = 20;
+// @see https://docs.github.com/en/github/managing-files-in-a-repository/working-with-non-code-files/rendering-and-diffing-images
+const IMAGE_FILE_TYPES = [".jpg", ".png", ".gif", ".psd", ".svg"];
+
+void features.add(__filebasename, {
+	include: [pageDetect.isCommit, pageDetect.isPRFiles],
+	init,
+});
+
+async function init(): Promise<void> {
+	window.setTimeout(run, 1000);
+
+	delegate(
+		document.body,
+		"include-fragment.diff-progressive-loader",
+		"load",
+		run
+	);
+}
+
+const run = () => {
+	const imageFiletypeCssSelector = IMAGE_FILE_TYPES.map((imageFileType) => {
+		return `.show-inline-notes[data-file-type="${imageFileType}"][data-file-deleted="false"] .js-rendered`;
+	}).join(", ");
+
+	const richDiffButtons = Array.from(
+		document.querySelectorAll(imageFiletypeCssSelector)
+	);
+
+	const unselectedRichDiffButtons = richDiffButtons.filter(
+		(button) => !button.classList.contains("selected")
+	);
+
+	if (unselectedRichDiffButtons.length === 0) return;
+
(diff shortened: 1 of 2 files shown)


fregante:
Thank you for the PR Cristoph! As per the [pinned issue](https://github.com/sindresorhus/refined-github/issues/4596), we're currently not accepting PRs to add new features, and this PR seems that it will require a few review rounds.

I will reopen this in a few months 😅 

Just a few notes though:

- no "load more" button. If the loading is intensive, we should just use IntersectionObserver to load them 600px before they come into view
- if elements are _expected_ to be there (i.e. "if this function runs, then we definitely have this button too"), use `!` to silence TypeScript errors about `| undefined` instead of `if a === null, return`
- use `select`/`select.all` instead of query selector
- use JSX to create elements (the contributing.md guide has some info)

With this and a few more changes the feature will be half as long or shorter… but we won't know for a while.

PRs for bugfixes are welcome though 😌

Perni1984:
@fregante actually #2673 is a feature that has the `please` tag on it, because it has existed before but has been disabled. So I thought a solution to it would be eventually accepted also during the "summer feature freeze" - at least this was my understanding from https://github.com/sindresorhus/refined-github/issues/4596.

Anyways thanks for your valuable feedback / thoughts on the PR - I will try my luck with resubmitting it considering your feedback if that is ok for you.

fregante:
Indeed! It's probably best to acquaintance yourself with the codebase and its style by taking some smaller issue/bugfix first since 130 lines is quite a mouthful for a start. 😁 It'll make the reviewing easier as well. Also make sure it passes linting before any reviews are done.

PR #4629, opened 2021-08-02
Fix some features appearing multiple times

### LINKED ISSUE #4622: `repo-age` appears duplicate
### Please ensure:

### Description

1. Visit https://github.com/vutran/twas/tree/v2.1.2
2. Go to the Master branch **from the branch dropdown** (not with our `default-branch-button`)
3. Go back in history
4. Go forward in history

It might be worth checking if other features are affected.

### Screenshot

https://user-images.githubusercontent.com/1402241/127603617-4eaba574-c1d3-4ba0-bf32-138ca88eef6e.mov

### Console errors

_No response_

### Example URL

https://github.com/vutran/twas/tree/v2.1.2

### Browser(s) used

Safari

dertieran:
Fixes #4622

The issue described in #4622 was caused because the `<has-rgh-inner>` element was removed on the first navigation.
So all features that checked it ran again, but the loader that would add `<has-rgh-inner>` again didn't run because the `<has-rgh>` was still present.

This PR splits them up into two loaders, I did not yet unify the code.
We could e.g. refactor it into a factory function, happy to do this if wanted.

### DIFF
--- a/source/features/index.tsx
+++ b/source/features/index.tsx
@@ -270,24 +270,15 @@ This means that the old features will still be on the page and don't need to re-
 
 This marks each as "processed"
 */
-void add(
-	__filebasename,
-	{
-		deduplicate: 'has-rgh',
-		init: async () => {
-			// `await` kicks it to the next tick, after the other features have checked for 'has-rgh', so they can run once.
-			await Promise.resolve();
-			select('#js-repo-pjax-container, #js-pjax-container')?.append(<has-rgh/>);
-		},
+void add(__filebasename, {
+	deduplicate: false,
+	init: async () => {
+		// `await` kicks it to the next tick, after the other features have checked for 'has-rgh', so they can run once.
+		await Promise.resolve();
+		select('#js-repo-pjax-container, #js-pjax-container')?.append(<has-rgh/>);
+		select('#repo-content-pjax-container')?.append(<has-rgh-inner/>); // #456
 	},
-	{
-		deduplicate: 'has-rgh-inner',
-		init: async () => {
-			await Promise.resolve();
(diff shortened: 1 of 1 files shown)


fregante:
uhmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmm this looks weird. The code you added tries to deduplicate itself, but in practice there's no difference, from what I can tell.

The only thing this is deduplicating is the `has-rgh/inner` elements, which does not need to be deduplicated as it's not seen. If one exists, it acts as a deduplicator; if 2 exist, the second one is ignored; GitHub removes both anyway because the parent of each is removed.

fregante:
Oh, wait… I think I got what's happening.

This function should not have a `deduplicate` logic at all. This can either be done by adding `deduplicate: false` on the original code, if I remember correctly, or by fixing the deduplicator code to detect and disable it for this `init` function. The first option sounds quicker.

dertieran:
This would also be an option, but I thought we would still need to check if we actually need to add the *deduplication* element so we don't add it twice.
My idea was to just reuse the already existing `deduplicate` logic that does this check.
But I'm also fine doing it in `init` and setting `deduplicate: false`.

fregante:
Duplicate deduplicators are fine, as my hidden comment mentioned. I wouldn't worry about that

fregante (approved):

fregante:
Does this also fix https://github.com/sindresorhus/refined-github/issues/4622#issuecomment-893160161?

dertieran:
Yes it does

https://user-images.githubusercontent.com/9264728/128326184-886c5a1c-a1fb-4117-9c44-7182eb0c075d.mp4

PR #4634, opened 2021-08-02
Add username tooltip to `reactions-avatars`

kidonng:
Address https://github.com/sindresorhus/refined-github/pull/4608#discussion_r678439504

## Test URLs

https://github.com/sindresorhus/refined-github/pull/4608

## Screenshot

image


fregante:
By looking at the screenshot, this doesn't seem useful. There's already a tooltip that shows the same names in order. Matching the avatar to the username isn't particularly useful imho

kidonng:
Just realized that, fair point!

Issue #3978, opened 2021-02-14
Enable `reactions-avatars` on discussions  and releases


yakov116:
Such as https://github.com/tophf/mpiv/discussions/50

yakov116:
Its not so easy as its lazy loaded. (The element gets replaced on load)

fregante:
Yeah I'm afraid that it would require substantial rewrite of the feature just to apply to a completely different dom and events 🤔

fregante:
If anyone picks this up, it’d have to be considered brand-new feature. It can be merged later, but likely all the JS needs to be different.

kidonng:
Maybe also on releases: https://github.com/microsoft/winget-cli/releases/tag/v-0.3.11102-preview

![image](https://user-images.githubusercontent.com/44045911/117522707-61ea8e00-afe7-11eb-8b8a-6d19ec07ca2a.png)

fregante:
I think it's less useful on releases

fregante:
> I think it's less useful on releases

I changed my mind

PR #4635, opened 2021-08-02
Fix style conflict caused by `reactions-avatars` on other pages

### LINKED ISSUE #4630: `reactions-avatars` style applied to not implemented pages
### Please ensure:

### Description

The reaction count number in a discussion's or a release's reaction gets a gray background in a square around it.

Not only that, those counts also seem to get slightly misaligned and smaller in comparison to without Refined Github.

Find Feature can't find it, so it may be a problem in a global stylesheet.

### Screenshot

With Refined Github:
imagen

Without Refined Github:
imagen

### Console errors

_No response_

### Example URL

https://github.com/github/feedback/discussions/4291
https://github.com/sindresorhus/refined-github/releases/tag/21.8.1

### Browser(s) used

Firefox 90.0.2

kidonng:
Fix #4630

## Test URLs

- Discussions: https://github.com/github/feedback/discussions/4291
- Releases: https://github.com/sindresorhus/refined-github/releases/tag/21.8.1

### DIFF
--- a/source/features/reactions-avatars.css
+++ b/source/features/reactions-avatars.css
@@ -1,9 +1,9 @@
-.reaction-summary-item {
+.reaction-summary-item:not(.js-reaction-group-button) {
 	padding-right: 0.7em !important;
 	padding-left: 0.7em !important;
 }
 
-button.reaction-summary-item { /* `button` excludes the "Add reaction" icon */
+button.reaction-summary-item:not(.js-reaction-group-button) { /* `button` excludes the "Add reaction" icon */
 	border-top: 1px solid var(--github-border-color); /* Required when the second line is longer than the first line */
 	border-bottom: 1px solid var(--github-border-color); /* Required when the first line is longer than the second line */
 	margin-bottom: -1px; /* Makes up for `order-bottom` */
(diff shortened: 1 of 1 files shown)


fregante:
Why not use a more specific selector instead? If GitHub adds reactions to more places we have to add more exceptions. Is there a comments class that doesn't appear on other pages?

altrisi:
Another solution for most of those selectors would be to add a custom class to reaction avatars so they can be referenced directly instead of relying on them being the only `span` inside the reaction button.

fregante:
Good idea, we can add a class on the `button` and then target the contents via span and img

fregante:
Merging because it’s a fix but hoping for a better fix later. Tagging it as such 👇

kidonng:
I kinda forgot about this PR. IMO it's specific enough because it seems the discussion and release reactions uses the same "newer" markup and I'd expect any (potential) future additional will be the same.

fregante:
Discussions tracked here: https://github.com/sindresorhus/refined-github/issues/3978#issuecomment-835082262

fregante:
Screen Shot

I'm see this on every page now

PR #4639, opened 2021-08-04
Make some feature-generated links load via AJAX

kidonng:
Motivation: https://github.com/sindresorhus/refined-github/pull/4631#issuecomment-891201750

This PR will continue for a few days in case anyone find something new, but hopefully searching for `<a` will give not too much results to check.

Ajaxed features and Test URLS:

- `comments-time-machine-links`: https://github.com/sindresorhus/refined-github/blob/main/.gitattributes?rgh-link-date=2021-08-01
- `linkify-branch-references`: 🤷‍♀️ Something like https://github.com/sindresorhus/refined-github/compare/main...test?quick_pull=1
- `linkify-symbolic-links`: https://github.com/lxqt/lxqt/blob/master/CONTRIBUTING
- `list-prs-for-file`: https://github.com/NixOS/nixpkgs/blob/master/pkgs/top-level/all-packages.nix
- `rgh-feature-descriptions`: https://github.com/sindresorhus/refined-github/blob/main/source/features/ci-link.tsx
- `swap-branches-on-compare`: https://github.com/sindresorhus/refined-github/compare/hotfix...main
- `view-markdown-source`: https://github.com/sindresorhus/refined-github/blob/main/readme.md

PR #4641, opened 2021-08-04
Fix AJAX navigation for compare link in `latest-tag-button` and add it to `view-markdown-source`

dertieran:
As pointed out by @kidonng in https://github.com/sindresorhus/refined-github/pull/4631#discussion_r682548229
The `#repo-content-pjax-container` did not work correctly for the compare link.

As a reparation I added the AJAX navigation to the `view-markdown-source` feature. 😄
I hope this will not conflict with #4639

## Test URLs

- Compare Link: https://github.com/sindresorhus/got
- `view-markdown-source`: https://github.com/sindresorhus/refined-github/blob/main/readme.md

## Screenshot

https://user-images.githubusercontent.com/9264728/128202068-c22975f2-f423-43dd-b0dd-0aafa2bee955.mp4

https://user-images.githubusercontent.com/9264728/128202284-cdce99f0-8345-4c95-b4bf-3829f7ba5519.mp4

### DIFF
--- a/source/features/latest-tag-button.tsx
+++ b/source/features/latest-tag-button.tsx
@@ -136,7 +136,7 @@ async function init(): Promise<false | void> {
 				<a
 					className="btn btn-sm btn-outline tooltipped tooltipped-ne"
 					href={buildRepoURL(`compare/${latestTag}...${defaultBranch}`)}
-					data-pjax="#repo-content-pjax-container"
+					data-pjax="#js-repo-pjax-container"
 					aria-label={`Compare ${latestTag}...${defaultBranch}`}
 				>
 					<DiffIcon className="v-align-middle"/>
--- a/source/features/view-markdown-source.tsx
+++ b/source/features/view-markdown-source.tsx
@@ -11,6 +11,7 @@ async function init(): Promise<void> {
 		<div className="BtnGroup rgh-view-markdown-source mr-1">
 			<a
 				href="?plain=1"
+				data-pjax="#repo-content-pjax-container"
 				className="btn btn-sm BtnGroup-item tooltipped tooltipped-nw"
 				aria-label="Display the source"
 			>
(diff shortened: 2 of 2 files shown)


source/features/latest-tag-button.tsx
```diff
@@ -136,7 +136,7 @@ async function init(): Promise<false | void> {
 				<a
 					className="btn btn-sm btn-outline tooltipped tooltipped-ne"
 					href={buildRepoURL(`compare/${latestTag}...${defaultBranch}`)}
-					data-pjax="#repo-content-pjax-container"
+					data-pjax="#js-repo-pjax-container"
```
kidonng, line 139: No, this won't work either

https://user-images.githubusercontent.com/44045911/128203855-b5f5b3dd-437c-4fc8-a905-ac4976f91517.mov

dertieran: Okay maybe I shouldn't do quick fixes for this 😅
But interesting that the next navigation is the issue in that case.

dertieran:
Okay 👍🏻 
I guess we should still fix the bug asap even if it is just removing the attribute from that link.
So the new version doesn't get published with a known bug, thoughts @fregante?

fregante:
No rush to publish, the bugs are not severe. I'd like to fix them before the next schedule release on September 1st though 😃

kidonng:
Interestingly, the AJAX requests for compare page now returns document fragment correctly. I wonder if something changes in GitHub or it's just my previous test is flawed.

fregante:
Has this been tested? Do you want to merge it as is?

kidonng:
@fregante The fix is not needed anymore

kidonng:
Oh, the `view-markdown-source` link improvement is still useful... But as discussed above I'm probably gonna merge it in my PR.

PR #4646, opened 2021-08-07
Add more details to `useful-not-found-page` 

### LINKED ISSUE #4617: Add more information to `useful-not-found-page`
### Description

Renamed files turn into 404s, but we add a link to the commit history:

Screen Shot 9

What we if were more explicit about what happened to the file and link to the last known version?

Example:

- the file was deleted by `bab4ba` 3 months ago
- the file was renamed to `a.jsx`
- the branch doesn't exist
- see the last version

Arguably these links are more useful than the breadcrumbs resolution and can be appended to the page as the API calls resolve

### Example URL

https://github.com/gorhill/uBlock/blob/master/assets/ublock/resources.txt

gamemaker1:
## What this PR adds

This PR adds a couple of things mentioned in #4617:

- If the branch does not exist and the user is trying to view a file on that branch, tell the user that the branch was deleted.
	* In addition, check if the file exists on the default branch (existing behaviour)
- If the file was deleted/moved, show who did it, when and with which commit. Also link to the last version of the file before the commit took place.
- If the file was renamed, show who did it, when and with which commit. Also link to the renamed file.

All of the above information is added as the API calls resolve.

## Related issues

Closes #4617 (if merged).

## Example URLs

- Renamed file: https://github.com/gamemaker1/test/blob/main/rename-test/readme.txt
- Moved/deleted file: https://github.com/gorhill/uBlock/blob/master/assets/ublock/resources.txt
- Non existent branch (and file does not exist): https://github.com/gorhill/uBlock/blob/non-existent-branch/assets/ublock/resources.txt
- Non existent branch (but file exists on default branch): https://github.com/gorhill/uBlock/blob/non-existent-branch/assets/resources/scriptlets.js

## Screenshots

<details>
<summary>Renamed file</summary>

refined-github-renamed-file

</details>

<details>
<summary>Moved/deleted file</summary>

refined-github-moved-or-deleted-file

</details>

<details>
<summary>Non existent branch (and file does not exist)</summary>

refined-github-non-existent-branch-and-no-file

</details>

<details>
<summary>Non existent branch (but file exists on default branch)</summary>

refined-github-non-existent-branch-and-file-on-default-branch

</details>

---

Please do let me know if there are any improvements that can be made to the changes made in this PR.

Thanks!

### DIFF
--- a/source/features/useful-not-found-page.tsx
+++ b/source/features/useful-not-found-page.tsx
@@ -1,36 +1,37 @@
 import React from 'dom-chef';
-import onetime from 'onetime';
 import select from 'select-dom';
+import onetime from 'onetime';
 import elementReady from 'element-ready';
-
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
-import * as api from '../github-helpers/api';
-import GitHubURL from '../github-helpers/github-url';
 import getDefaultBranch from '../github-helpers/get-default-branch';
 import {getCleanPathname} from '../github-helpers';
-
-function parseCurrentURL(): string[] {
-	const parts = getCleanPathname().split('/');
-	if (parts[2] === 'blob') {
-		parts[2] = 'tree';
-	}
-
-	return parts;
-}
+import * as api from '../github-helpers/api';
+import GitHubURL from '../github-helpers/github-url';
 
 async function is404(url: string): Promise<boolean> {
-	const {status} = await fetch(url, {
-		method: 'head',
-	});
+	const {status} = await fetch(url, {method: 'head'});
 	return status === 404;
+}
+
+function getStrikeThrough(text: string): HTMLElement {
+	return <del className="color-text-tertiary">{text}</del>; /* GHE #4121 */
 }
 
 async function checkIfPartExists(element: HTMLAnchorElement): Promise<void> {
 	if (await is404(element.href)) {
-		element.replaceWith(<del className="color-text-tertiary">{element.textContent}</del>);
+		element.replaceWith(getStrikeThrough(element.textContent!));
 	}
+}
+
+function parseCurrentURL(): string[] {
+	const parts = getCleanPathname().split('/');
+	if (parts[2] === 'blob') { // Blob URLs are never useful
+		parts[2] = 'tree';
+	}
+
+	return parts;
 }
 
 async function showMissingPart(bar: Element): Promise<void> {
@@ -38,12 +39,13 @@ async function showMissingPart(bar: Element): Promise<void> {
 
 	for (const [i, part] of parts.entries()) {
 		if (i === 2 && ['tree', 'blob', 'edit'].includes(part)) {
+			// Exclude parts that don't exist as standalones
 			continue;
 		}
 
-		// Last part caused the 404; we know that it does not exist for sure
+		// The last part of the URL is a known 404
 		if (i === parts.length - 1) {
-			bar.append(' / ', <del className="color-text-tertiary">{part}</del>);
+			bar.append(' / ', getStrikeThrough(part));
 		} else {
 			const pathname = '/' + parts.slice(0, i + 1).join('/');
 			bar.append(i ? ' / ' : '', <a href={pathname}>{part}</a>);
@@ -52,6 +54,7 @@ async function showMissingPart(bar: Element): Promise<void> {
 
 	select('main > :first-child, #parallax_illustration')!.after(bar);
 
+	// Check parts from right to left; skip the last part
 	for (let i = bar.children.length - 2; i >= 0; i--) {
 		void checkIfPartExists(bar.children[i] as HTMLAnchorElement);
 	}
@@ -184,36 +187,27 @@ function initUsefulNotFoundPage(): void {
 	void showHelpfulLinks(bar);
 }
 
-async function initPRCommitNotFoundPage(): Promise<void | false> {
+async function initPRCommit(): Promise<void | false> {
 	const commitUrl = location.href.replace(/pull\/\d+\/commits/, 'commit');
 	if (await is404(commitUrl)) {
 		return false;
 	}
 
-	const blankSlateParagraph = await elementReady('.blankslate p', {
-		waitForChildren: false,
-	});
-
+	const blankSlateParagraph = await elementReady('.blankslate p', {waitForChildren: false});
 	blankSlateParagraph!.after(
-		<p>
-			You can also try to <a href={commitUrl}>view the detached standalone commit</a>.
-		</p>,
+		<p>You can also try to <a href={commitUrl}>view the detached standalone commit</a>.</p>,
 	);
 }
 
-void features.add(
-	__filebasename,
-	{
-		include: [
-			pageDetect.is404,
-		],
-		init: onetime(initUsefulNotFoundPage),
-	},
-	{
-		include: [
-			pageDetect.isPRCommit404,
-		],
-		awaitDomReady: false,
-		init: onetime(initPRCommitNotFoundPage),
-	},
-);
+void features.add(__filebasename, {
+	include: [
+		pageDetect.is404,
+	],
+	init: onetime(initUsefulNotFoundPage),
+}, {
+	include: [
+		pageDetect.isPRCommit404,
+	],
+	awaitDomReady: false,
+	init: onetime(initPRCommit),
+});
--- a/source/features/useful-not-found-page.tsx
+++ b/source/features/useful-not-found-page.tsx
@@ -38,15 +38,26 @@ function parseCurrentURL(): string[] {
 	return parts;
 }
 
-async function displayObjectStatus(bar: Element): Promise<void> {
-	// Object might have been deleted/moved
-	if (!await addObjectStatusInfo(bar)) {
-		// Or 410 Gone
-		await addCommitHistoryLink(bar);
+async function findAndDisplayMissingPart(bar: Element): Promise<void> {
+	// The branch might have been deleted
+	const branchDeleted = await addBranchStatus(bar);
+	// Or the file exists, but on a different branch
+	const branchDoesNotHaveFile = await addDefaultBranchLink(bar);
+	// If either of the above are true, we don't need to go into the file's commit history.
+	if (branchDeleted || branchDoesNotHaveFile) {
+		return;
 	}
+
+	// The object might have been deleted/moved
+	if (await addObjectStatus(bar)) {
+		return;
+	}
+
+	// Or 410 Gone
+	await addCommitHistoryLink(bar);
 }
 
-async function addObjectStatusInfo(bar: Element): Promise<boolean> {
+async function addObjectStatus(bar: Element): Promise<boolean> {
 	// Get the file path from the parts
 	const parts = parseCurrentURL();
 	const filePath = parts.slice(4).join('/');
@@ -72,15 +83,19 @@ async function addObjectStatusInfo(bar: Element): Promise<boolean> {
 		});
 
 		const commitAuthor = lastCommitInfo.author.login;
+		const commitTime = new Date(lastCommitInfo.commit.committer.date);
+		const commitSha = lastCommitInfo.sha;
+		const shortenedCommitSha = commitSha.slice(0, 8);
+		const commitUrl = lastCommitInfo.html_url;
+
 		const urlToCommitAuthorProfile = lastCommitInfo.author.html_url;
 		const urlToLastBlob = fileInfo.blob_url;
-		const commitTime = new Date(lastCommitInfo.commit.committer.date);
 
 		// If it was removed, tell the user
 		if (fileInfo.status === 'removed') {
 			bar.after(
 				<p className="container mt-4 text-center">
-					The file you are looking for was deleted/moved by <a href={urlToCommitAuthorProfile}>{commitAuthor}</a> { twas(commitTime.getTime()) }.
+					The file you are looking for was deleted/moved by <a href={urlToCommitAuthorProfile}>{commitAuthor}</a> {twas(commitTime.getTime())} with commit <a href={commitUrl}>{shortenedCommitSha}</a>.
 					<br/>
 					You can view the last version of the file <a href={urlToLastBlob}>here</a>.
 				</p>,
@@ -93,7 +108,7 @@ async function addObjectStatusInfo(bar: Element): Promise<boolean> {
 		if (fileInfo.status === 'renamed') {
 			bar.after(
 				<p className="container mt-4 text-center">
-					The file you are looking for was renamed by <a href={urlToCommitAuthorProfile}>{commitAuthor}</a> {twas(commitTime.getTime())}.
+					The file you are looking for was renamed by <a href={urlToCommitAuthorProfile}>{commitAuthor}</a> {twas(commitTime.getTime())} with commit <a href={commitUrl}>{shortenedCommitSha}</a>.
 					<br/>
 					You can find the renamed file <a href={urlToLastBlob}>here</a>.
 				</p>,
@@ -107,12 +122,12 @@ async function addObjectStatusInfo(bar: Element): Promise<boolean> {
 }
 
 // If the object was deleted, link to the commit history
-async function addCommitHistoryLink(bar: Element): Promise<void> {
+async function addCommitHistoryLink(bar: Element): Promise<boolean> {
 	const parts = parseCurrentURL();
 	parts[2] = 'commits';
 	const url = '/' + parts.join('/');
 	if (await is404(location.origin + url)) {
-		return;
+		return false;
 	}
 
 	bar.after(
@@ -120,9 +135,11 @@ async function addCommitHistoryLink(bar: Element): Promise<void> {
 			See also the {getType()}’s <a href={url}>commit history</a>.
 		</p>,
 	);
+
+	return true;
 }
 
-async function addBranchStatusInfo(bar: Element): Promise<void> {
+async function addBranchStatus(bar: Element): Promise<boolean> {
 	// Get the current branch
 	const parts = parseCurrentURL();
 	const currentBranch = parts[3];
@@ -138,26 +155,30 @@ async function addBranchStatusInfo(bar: Element): Promise<void> {
 				The branch you are trying to view does not exist.
 			</p>,
 		);
+
+		return true;
 	}
+
+	return false;
 }
 
 // If the object exists in the default branch, link to it
-async function addDefaultBranchLink(bar: Element): Promise<void> {
+async function addDefaultBranchLink(bar: Element): Promise<boolean> {
 	const parts = getCleanPathname().split('/');
 	const branch = parts[3];
 	if (!branch) {
-		return;
+		return false;
 	}
 
 	const defaultBranch = await getDefaultBranch();
 	if (branch === defaultBranch) {
-		return;
+		return false;
 	}
 
 	parts[3] = defaultBranch; // Change branch
 	const url = '/' + parts.join('/');
 	if (await is404(location.origin + url)) {
-		return;
+		return false;
 	}
 
 	bar.after(
@@ -165,6 +186,8 @@ async function addDefaultBranchLink(bar: Element): Promise<void> {
 			The {getType()} exists on the <a href={url}>default branch</a>.
 		</p>,
 	);
+
+	return true;
 }
 
 function init(): false | void {
@@ -197,16 +220,12 @@ function init(): false | void {
 		void checkAnchor(bar.children[i] as HTMLAnchorElement);
 	}
 
-	if (['tree', 'blob'].includes(parts[2])) {
-		void displayObjectStatus(bar);
-	}
-
+	// Check for what exactly is missing in the following order:
+	// - check if the branch exists
+	// - check if the file exists on the default branch
+	// - check if the file was deleted/moved/renamed
 	if (['tree', 'blob', 'edit'].includes(parts[2])) {
-		// File might not be available on the current branch
-		// GitHub already redirects /tree/ and /blob/ natively
-		void addDefaultBranchLink(bar);
-		// Or the branch itself may not exist
-		void addBranchStatusInfo(bar);
+		void findAndDisplayMissingPart(bar);
 	}
 }
 
--- a/source/features/useful-not-found-page.tsx
+++ b/source/features/useful-not-found-page.tsx
@@ -117,7 +117,7 @@ async function showMissingPart(): Promise<void> {
 async function showHelpfulLinks(): Promise<void> {
 	const urlToFileOnDefaultBranch = await getUrlToFileOnDefaultBranch();
 	if (urlToFileOnDefaultBranch) {
-		select('h2')!.after(
+		select('main > .container-lg')!.before(
 			<p className="container mt-4 text-center">
 				<a href={urlToFileOnDefaultBranch}>This {getType()}</a> exists on the default branch.
 			</p>,
@@ -132,7 +132,7 @@ async function showHelpfulLinks(): Promise<void> {
 	const url = new GitHubURL(location.href);
 
 	url.assign({route: 'commits'});
-	const commitHistory = <a href={url.toString()}>Commit history</a>;
+	const commitHistory = <a href={url.toString()}>see commit history</a>;
 	url.assign({route: 'blob', branch: commitInfo.parents[0].sha, filePath: url.filePath});
 	const lastVersionUrl = fileInfo.status === 'removed' ? fileInfo.blob_url : url.toString();
 	const lastVersion = <a href={lastVersionUrl}>This {getType()}</a>;
@@ -141,7 +141,7 @@ async function showHelpfulLinks(): Promise<void> {
 		? 'deleted'
 		: <a href={fileInfo.blob_url}>moved</a>;
 
-	select('h2')!.after(
+	select('main > .container-lg')!.before(
 		<p className="container mt-4 text-center">
 			{lastVersion} was {verb} ({permalink}) - {commitHistory}.
 		</p>,
--- a/source/features/useful-not-found-page.tsx
+++ b/source/features/useful-not-found-page.tsx
@@ -1,20 +1,16 @@
-import React from 'dom-chef';
-import select from 'select-dom';
-import onetime from 'onetime';
-import elementReady from 'element-ready';
-import * as pageDetect from 'github-url-detection';
+import React from "dom-chef";
+import select from "select-dom";
+import onetime from "onetime";
+import elementReady from "element-ready";
+import * as pageDetect from "github-url-detection";
 
-import features from '.';
-import * as api from '../github-helpers/api';
-import getDefaultBranch from '../github-helpers/get-default-branch';
-import {getCleanPathname} from '../github-helpers';
-
-function getType(): string {
-	return location.pathname.split('/').pop()!.includes('.') ? 'file' : 'object';
-}
+import features from ".";
+import * as api from "../github-helpers/api";
+import getDefaultBranch from "../github-helpers/get-default-branch";
+import { getCleanPathname } from "../github-helpers";
 
 async function is404(url: string): Promise<boolean> {
-	const {status} = await fetch(url, {method: 'head'});
+	const { status } = await fetch(url, { method: "head" });
 	return status === 404;
 }
 
@@ -29,42 +25,52 @@ async function checkAnchor(anchor: HTMLAnchorElement): Promise<void> {
 }
 
 function parseCurrentURL(): string[] {
-	const parts = getCleanPathname().split('/');
-	if (parts[2] === 'blob') { // Blob URLs are never useful
-		parts[2] = 'tree';
+	const parts = getCleanPathname().split("/");
+	// Blob URLs are never useful
+	if (parts[2] === "blob") {
+		parts[2] = "tree";
 	}
 
 	return parts;
 }
 
-async function findAndDisplayMissingPart(bar: Element): Promise<void> {
-	// The branch might have been deleted
-	const branchDeleted = await addBranchStatus(bar);
-	// Or the file exists, but on a different branch
-	const branchDoesNotHaveFile = await addDefaultBranchLink(bar);
-	// If either of the above are true, we don't need to go into the file's commit history.
-	if (branchDeleted || branchDoesNotHaveFile) {
+async function getUrlToFileOnDefaultBranch(): Promise<string | undefined> {
+	// Get the current branch
+	const parts = parseCurrentURL();
+	const currentBranch = parts[3];
+
+	if (!currentBranch) {
 		return;
 	}
 
-	// The object might have been deleted/moved
-	if (await addObjectStatus(bar)) {
+	const defaultBranch = await getDefaultBranch();
+	if (currentBranch === defaultBranch) {
 		return;
 	}
 
-	// Or 410 Gone
-	await addCommitHistoryLink(bar);
+	// Change branch
+	parts[3] = defaultBranch;
+	const url = "/" + parts.join("/");
+	// Check if that path exists
+	if (await is404(location.origin + url)) {
+		return;
+	}
+
+	// If it does, return it
+	return url;
 }
 
-async function addObjectStatus(bar: Element): Promise<boolean> {
+async function getLastCommitForFile(
+	branch?: string
+): Promise<Record<string, string | undefined> | undefined> {
 	// Get the file path from the parts
 	const parts = parseCurrentURL();
-	const filePath = parts.slice(4).join('/');
+	let filePath = parts.slice(4).join("/");
 
 	// Get the last 2 commits that include the file
 	const commitsForFileResponse = await api.v4(`
 		repository() {
-			ref(qualifiedName: "${parts[3]}") {
+			ref(qualifiedName: "${branch ?? parts[3]}") {
 				target {
 					... on Commit {
 						history(first: 2, path: "${filePath}") {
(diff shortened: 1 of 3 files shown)


source/features/useful-not-found-page.tsx
```diff
@@ -36,46 +38,156 @@ function parseCurrentURL(): string[] {
 	return parts;
 }
 
-// If the resource was deleted, link to the commit history
-async function addCommitHistoryLink(bar: Element): Promise<void> {
+async function findAndDisplayMissingPart(bar: Element): Promise<void> {
+	// The branch might have been deleted
+	const branchDeleted = await addBranchStatus(bar);
+	// Or the file exists, but on a different branch
+	const branchDoesNotHaveFile = await addDefaultBranchLink(bar);
```
gamemaker1, line 45: When both of these conditions are true (see the [example screenshot](https://user-images.githubusercontent.com/34235681/128588754-d000609d-389f-45e8-b671-8ad957e6a7e4.png) for non existent branch but the file exists on default branch), the `addDefaultBranchLink` function adds the link to the file on the default branch, above the text that says the branch does not exist (added by the `addBranchStatus` function). UI-wise, the order of the text looks wrong to me.

This could be happening because of two reasons:

- We pass the same `bar` instance to both functions, and `bar.after()` keeps adding the elements to the top.
- The `addDefaultBranchLink` function finishes before the `addBranchStatus` function, thus adding its text to the `bar` first.

Possible solutions:

- Combining the two functions and adding the text as one `p` element.
- Returning the bar instance from the `addBranchStatus` function and passing that to the `addDefaultBranchLink` function (might not work if this error happens due to the first reason)
- Not showing the branch does not exist error if the file does exist on another branch.

It would be great if someone could advise me on how to fix this. Thanks!

source/features/useful-not-found-page.tsx
```diff
@@ -36,46 +38,156 @@ function parseCurrentURL(): string[] {
...
 	bar.after(
 		<p className="container mt-4 text-center">
 			See also the {getType()}’s <a href={url}>commit history</a>.
 		</p>,
 	);
+
+	return true;
+}
+
+async function addBranchStatus(bar: Element): Promise<boolean> {
```
fregante, line 142: I just realized that this piece of information is already on screen and we don't need to fetch it again.

source/features/useful-not-found-page.tsx
```diff
@@ -36,46 +38,156 @@ function parseCurrentURL(): string[] {
...
+	// Get the file path from the parts
+	const parts = parseCurrentURL();
+	const filePath = parts.slice(4).join('/');
+
+	// Get the last 2 commits that include the file
+	const previousCommitsIncludingFile = await api.v3(`commits?path=${filePath}&per_page=2`);
+	// The latest commit will be the first object in the array
+	if (previousCommitsIncludingFile[0]) {
+		// Get a list of changes that happened in the repo with this commit
+		const lastCommitInfo = await api.v3(`commits/${previousCommitsIncludingFile[0].sha as string}`);
```
fregante, line 70: I forgot to mention we should use v4. We only use v3 if there's no alternative, which is once or twice.

source/features/useful-not-found-page.tsx
```diff
@@ -36,46 +38,156 @@ function parseCurrentURL(): string[] {
...
+		const commitUrl = lastCommitInfo.html_url;
+
+		const urlToCommitAuthorProfile = lastCommitInfo.author.html_url;
+		const urlToLastBlob = fileInfo.blob_url;
+
+		// If it was removed, tell the user
+		if (fileInfo.status === 'removed') {
+			bar.after(
+				<p className="container mt-4 text-center">
+					The file you are looking for was deleted/moved by <a href={urlToCommitAuthorProfile}>{commitAuthor}</a> {twas(commitTime.getTime())} with commit <a href={commitUrl}>{shortenedCommitSha}</a>.
```
fregante, line 98: We should use the `<relative-time>` element instead of twas. Refer to its usage from other parts of the extension. No imports are necessary.

fregante (commented):
Thank you for the PR! Once settled we should also find a less verbose way to show all of this content. For example just:

- the file was (deleted|renamed) by (commit) (relative-time) - [See commit history](url)

source/features/useful-not-found-page.tsx
gamemaker1: Sure, I'll look up v4 and change all the calls to use that.

source/features/useful-not-found-page.tsx
gamemaker1: I could change this to use the `is404` method instead of calling the API and comparing the current branch with existing branches. Does that sound ok?

source/features/useful-not-found-page.tsx
gamemaker1: Will do, thanks.

source/features/useful-not-found-page.tsx
gamemaker1: I changed the call on line 65 (retrieving commits for the file that wasn't found) to use the v4 API.

On line 70 though (now line 87), it seems there is [no way to get a list of files from the GraphQL API](https://github.community/t/graphql-api-get-list-of-files-related-to-commit/14047/2). I looked up the [API reference](https://docs.github.com/en/graphql/reference/objects#commit) too, and the `history` and `blame` properties on the `Commit` object do not have any fields that will allow us to access the status of each file. I have kept it to use v3 for now. 

Please let me know if I missed something, and it is possible to get the status of each file in a commit using v4.

Thanks!

gamemaker1:
> A less verbose way to show all of this content. For example just:
> 
>     * the file was (deleted|renamed) by (commit) (relative-time) - [See commit history](url)



I changed the message to be a bit shorter:

<details>
<summary>Renamed file</summary>

refined-github-renamed-file-shorter-message

</details>

<details>
<summary>Details</summary>

refined-github-moved-or-deleted-file-shorter-message

</details>

Please let me know if it requires any further changes.

Thanks

source/features/useful-not-found-page.tsx
fregante: The info is already there, it just does not need to exist separately.

fregante: I think it should piggyback on the existing calls to determine what happened to the current URL and display one of:

- branch was deleted
- file was deleted
- file was renamed/moved

source/features/useful-not-found-page.tsx
fregante: Indeed: https://github.com/sindresorhus/refined-github/blob/9b8ac9771b55c0e5f3d6ba53a4da2a3b77556717/source/features/follow-file-renames.tsx#L17

fregante:
- "The file" can be the link to the last known version
- Changes are "renamed/moved" and "deleted". We can just stick to "moved" and "deleted" as verbs

source/features/useful-not-found-page.tsx
fregante: I think we have 2 parallel requests to make:

- does the file exist on default branch, currently
- does the file exist in history

If both are false

- does the file exist in default’s history

If that's also false, the file never existed.

> - does the file exist on default branch, currently

We show "The file only exists on the default branch" or "The branch has been deleted, see the file on the default branch"

> - does the file exist in history

We show "The file was deleted" or "The file was moved"

> - does the file exist in default’s history

Same as above, but change the links to point to default (we don't need to say anything about the missing branch)

source/features/useful-not-found-page.tsx
```diff
@@ -36,46 +37,173 @@ function parseCurrentURL(): string[] {
...
+			);
+
+			return true;
+		}
+
+		// If it was renamed, tell the user
+		if (fileInfo.status === 'renamed') {
+			bar.after(
+				<p className="container mt-4 text-center">
+					This {getType()} was renamed by <a href={commitUrl}>{shortenedCommitSha}</a> (<relative-time datetime={commitTime}/>) - see the {getType()}&apos;s <a href={commitHistoryUrl}>commit history</a> or view the <a href={urlToLastBlob}>renamed file</a>.
```
fregante, line 129: Only files can be renamed

source/features/useful-not-found-page.tsx
fregante: For this reason, the file should be reorganized in a way to have a single "What happened to this file" function (other than the breadcrumbs, which should not be affected). And then use it as something like:

```js
// Random names, to be improved, pseudo-code
const outcome = await whatHappened(location.href);
if (outcome.type === 'renamed') {
	page.append(<a href={outcome.oldFile}>File</a>, 'was renamed etc etc')
} else if (/* deleted */) {
	page.append(<a href={outcome.oldFile}>File</a>, 'was deleted etc etc')
}

File: source/features/useful-not-found-page.tsx (hunk 2 above)
Comment (fregante) [reply source/features/useful-not-found-page.tsx:158]: Refer to my other larger comment for a broader solution.

**Comment by gamemaker1** [comment]
> I think we have 2 parallel requests to make:
> 
> * does the file exist on default branch, currently
> * does the file exist in history
> 
> If both are false
> 
> * does the file exist in default’s history
> 
> 
> If that's also false, the file never existed, the URL is just wrong and we don't add anything else.
>
> * does the file exist on default branch, currently
> 
> We show "The file only exists on the default branch" or "The branch has been deleted, see the file on the default branch"
> 
> * does the file exist in history
> 
> We show "The file was deleted" or "The file was moved"
> 
> * does the file exist in default’s history
> 
> Same as above, but change the links to point to default (we don't need to say anything about the missing branch)

Thanks a lot, this really helped!

## Changes made

I have incorporated all of the changes you suggested, along with some of my own refactoring:

- Moved the rendering of the file path and striking-through into a seperate function (`showWhatIsMissingInPath`).

- Added a function (`showAdditionalInformation`) which asks what happened (by calling the `whatHappenedToTheFile` function) and renders something based on the response.

- Added the function `whatHappenedToTheFile`, which performs three checks, one after the other:

	* First, it checks if the file exists on the default branch (calls `getUrlToFileOnDefaultBranch`).

	* Else, it checks if the file exists in commit history on the current branch (calls `getLastCommitForFile`).

	* Else, it check if the file exists in the default branch's history (calls `getLinkToCommitHistoryOnDefaultBranch`).

- The `whatHappenedToTheFile` function returns an object with a type property. The type can be `removed`, `renamed`, `isOnDefaultBranch` or `wasOnDefaultBranch`. Based on the type, other properties are also returned:

	* `renamed`: 
		- `filePath`
		- `newFilePath`
		- `commitSha`
		- `commitDate`
		- `linkToCommit`
		- `linkToLastVersion`
		- `linkToNewVersion`
		- `linkToCommitHistory`

	* `removed`: 
		- `filePath`
		- `commitSha`
		- `commitDate`
		- `linkToCommit`
		- `linkToLastVersion`
		- `linkToCommitHistory`

	* `isOnDefaultBranch`:
		- `linkToLastVersion`

	* `wasOnDefaultBranch`:
		- `linkToCommitHistory`

- Once the `whatHappenedToTheFile` returns an event object, `showAdditionalInformation` goes ahead and renders info based on the type.

- I have also kept `renamed` and `deleted` as the verbs instead of `renamed` and `moved/deleted`

- Removed the `getType` function. I couldn't find a URL that would result in `object` as the type, so I put in the word file everywhere instead. Let me know if there are any edge cases I might have missed, in which case I will revert this.

- Largely replaced API calls with some URL manipulation and `is404` calls.

## Caveats/Problems

- When a file is renamed, the Github API does not return a link to the previous file (before renaming). I put in a bit of a hack to get a URL to the previous file: I take the sha of the last to last commit (before it was renamed) and generate a blob URL. Please let me know if there is a better way of doing this. Perhaps I should add a `is404` check there if we keep this hack.
- I could not find a test case for the `wasOnDefaultBranch` case. In my test repo, when I deleted the file and then created a new branch, the commit history was also copied over to the new branch - resulting in the deletion event existing in the history for the current branch as well. So the 404 page shows that the file was deleted instead of it is only on the default branch (see the last screenshot).

## Screenshots


<details>
<summary>Renamed file</summary>

refined-github-renamed-file

</details>

<details>
<summary>Deleted file</summary>

refined-github-moved-or-deleted-file

</details>

<details>
<summary>Non existent branch (and file does not exist)</summary>

refined-github-non-existent-branch-and-no-file

</details>

<details>
<summary>Non existent branch (but file exists on default branch)</summary>

refined-github-non-existent-branch-and-file-on-default-branch

</details>

<details>
<summary>`wasOnDefaultBranch` case (repo - https://github.com/gamemaker1/test)</summary>

refined-github-was-on-default-branch

</details>

---

Also, sorry for combining so many changes into one commit. I should have split them, it would have been easier to review.

**Comment by fregante** [comment]
> Also, sorry for combining so many changes into one commit. I should have split them, it would have been easier to review.

I commit the whole code at once, don't worry about splitting the commits

File: source/features/useful-not-found-page.tsx [hunk 6]
```diff
@@ -28,119 +25,360 @@ async function checkAnchor(anchor: HTMLAnchorElement): Promise<void> {
...
+	{
+		include: [pageDetect.is404],
+		init: onetime(init),
+	},
+	{
+		include: [pageDetect.isPRCommit404],
+		awaitDomReady: false,
+		init: onetime(initPRCommit),
+	}
+);
```
fregante, line 384: Please don't run Prettier on the code, use the included linter. You can run `npm run fix` and undo unrelated changes like this one

source/features/useful-not-found-page.tsx
```diff
@@ -28,119 +25,360 @@ async function checkAnchor(anchor: HTMLAnchorElement): Promise<void> {
...
+		// Exclude parts that don't exist as standalones
+		if (i === 2 && ["tree", "blob", "edit"].includes(part)) {
 			continue;
 		}
 
+		// The last part of the URL is a known 404
 		if (i === parts.length - 1) {
-			// The last part of the URL is a known 404
-			bar.append(' / ', getStrikeThrough(part));
+			// Put a strikthrough through it by default
```
fregante, line 313: Other unrelated change. Since this is a huge PR, refer from changing code unnecessarily

source/features/useful-not-found-page.tsx
```diff
@@ -28,119 +25,360 @@ async function checkAnchor(anchor: HTMLAnchorElement): Promise<void> {
...
+
+	if (eventThatHappened.type === "removed") {
+		bar.after(
+			<p className="container mt-4 text-center">
+				<a href={eventThatHappened.linkToLastVersion}>This file</a> was deleted
+				by{" "}
+				<a href={eventThatHappened.linkToCommit}>
+					{eventThatHappened.commitSha}
+				</a>{" "}
+				(<relative-time datetime={new Date(eventThatHappened.commitDate!)} />) -
```
fregante, line 274: Let's skip the commit number, we can just linkify the date (everywhere)
```ts
				<a href={eventThatHappened.linkToCommit}>
					<relative-time datetime={new Date(eventThatHappened.commitDate!)} />
				</a> -
```

source/features/useful-not-found-page.tsx
```diff
@@ -28,119 +25,360 @@ async function checkAnchor(anchor: HTMLAnchorElement): Promise<void> {
...
+	// If it does, return it
+	return url;
+}
+
+async function getLastCommitForFile(
+	branch?: string
+): Promise<Record<string, string | undefined> | undefined> {
+	// Get the file path from the parts
+	const parts = parseCurrentURL();
+	let filePath = parts.slice(4).join("/");
```
fregante, line 68: Refer to `GitHubURL` to parse the URL, I think it works here as well (I might be wrong)

source/features/useful-not-found-page.tsx
```diff
@@ -28,119 +25,360 @@ async function checkAnchor(anchor: HTMLAnchorElement): Promise<void> {
...
+	}
+
+	if (fileInfo.status === "renamed") {
+		filePath = fileInfo.previous_filename || filePath;
+		const newFilePath = fileInfo.previous_filename ? filePath : undefined;
+
+		const linkToNewVersion = fileInfo.blob_url;
+		let linkToLastVersion = "";
+
+		const lastToLastCommit =
```
fregante, line 146: You can get the previous commit from the current one by using one of its parents. Docs: https://docs.github.com/en/graphql/reference/objects#commit

source/features/useful-not-found-page.tsx
```diff
@@ -28,119 +25,360 @@ async function checkAnchor(anchor: HTMLAnchorElement): Promise<void> {
...
+			commitsForFileResponse.repository.ref.target.history.nodes[1];
+		if (lastToLastCommit) {
+			// HACK: Generate a blob URL for the file at the time of the previous commit
```
fregante, line 149: Try using GitHubURL to generate/change the URL instead. Starting from the current page, you should be able to just set 2 parts (I don't remember if these are the correct props):

```js
url.assign({
	ref: lastToLastCommit.oid,
	filePath
})
```

source/features/useful-not-found-page.tsx
```diff
@@ -28,119 +25,360 @@ async function checkAnchor(anchor: HTMLAnchorElement): Promise<void> {
...
+		if (file.filename === filePath) {
+			return true;
+		}
+
+		if (file.status === "renamed" && file.previous_filename === filePath) {
+			return true;
+		}
+
+		return false;
+	});
```
fregante, line 112: Rather than checking, this is just looking for the file
```suggestion
	const file = lastCommitInfo.files.find((file: AnyObject) => [file.filename, file.previous_filename].includes(filePath));
```

source/features/useful-not-found-page.tsx
```diff
@@ -28,119 +25,360 @@ async function checkAnchor(anchor: HTMLAnchorElement): Promise<void> {
...
+			commitDate,
+			linkToCommit,
+			linkToLastVersion,
+			linkToCommitHistory,
+		};
+	}
+
+	if (fileInfo.status === "renamed") {
+		filePath = fileInfo.previous_filename || filePath;
+		const newFilePath = fileInfo.previous_filename ? filePath : undefined;
```
fregante, line 141: If it was renamed, `fileInfo.previous_filename` must exist, right?

source/features/useful-not-found-page.tsx
gamemaker1: Oh sorry about that, I use the prettier extension in VSCode. I'll revert it right away.

source/features/useful-not-found-page.tsx
gamemaker1: Sure, I'll revert those comments.

source/features/useful-not-found-page.tsx
gamemaker1: Sure

source/features/useful-not-found-page.tsx
gamemaker1: Thanks! I'll use that

source/features/useful-not-found-page.tsx
gamemaker1: Sure, will do.

source/features/useful-not-found-page.tsx
gamemaker1: Yes, that is a copy paste error. I'll change that, thanks

gamemaker1:
> What I'm more concerned about is that 300 lines have been added and we'll have to rethink this to at least halve it.

It's about 250 lines now (after incorporating your suggestions/changes), I'll try making it smaller.

> Refer to GitHubURL to parse the URL, I think it works here as well (I might be wrong)

That worked perfectly, I've replaced most of the `parseCurrentUrl` calls with `GithubURL`. Thanks!

source/features/useful-not-found-page.tsx
```diff
@@ -29,71 +27,247 @@ async function checkAnchor(anchor: HTMLAnchorElement): Promise<void> {
...
+	}
+
+	if (eventThatHappened.type === 'wasOnDefaultBranch') {
+		bar.after(
+			<p className="container mt-4 text-center">
+				This file used to exist on the default branch - see <a href={eventThatHappened.linkToCommitHistory}>commit history</a>.
+			</p>,
+		);
+		return;
+	}
```
fregante, line 238: This situation should just be covered by the deleted/renamed one, without mentioning it. Details in https://github.com/sindresorhus/refined-github/pull/4646#discussion_r684608765

source/features/useful-not-found-page.tsx
```diff
@@ -29,71 +27,247 @@ async function checkAnchor(anchor: HTMLAnchorElement): Promise<void> {
...
+	// https://github.community/t/graphql-api-get-list-of-files-related-to-commit/14047/2
+	const lastCommitInfo = await api.v3(
+		`commits/${previousCommitsIncludingFile.oid as string}`,
 	);
+
+	// Check what happened to this particular file
+	const fileInfo = lastCommitInfo.files.find((file: AnyObject) => [file.filename, file.previous_filename].includes(filePath));
+	if (!fileInfo) {
+		return;
+	}
```
fregante, line 108: All of this can be replaced by a single v3 call: https://stackoverflow.com/a/15838387/288906

source/features/useful-not-found-page.tsx
```diff
@@ -29,71 +27,247 @@ async function checkAnchor(anchor: HTMLAnchorElement): Promise<void> {
...
+	// Check if that path exists
+	if (await is404(url.toString())) {
+		return;
+	}
+
+	// If it does, return it
+	return url.toString();
+}
+
+async function getLastCommitForFile(
```
fregante, line 65: The name does not match the return value. Ensure that functions do one thing and their names match what they return

source/features/useful-not-found-page.tsx
```diff
@@ -29,71 +27,247 @@ async function checkAnchor(anchor: HTMLAnchorElement): Promise<void> {
...
+			commitSha,
+			commitDate,
+			linkToCommit,
+			linkToLastVersion,
+			linkToCommitHistory,
+		};
+	}
+
+	if (fileInfo.status === 'renamed') {
+		const newFilePath = fileInfo.filename;
```
fregante, line 135: If variables are used once, just declare them where they're used or skip them. For example this one can just be in the returned object:

```diff
 		return {
 			type: 'renamed',
 			filePath,
+ 			newFilePath: fileInfo.filename,
 			commitSha,
 			commitDate,
 			linkToCommit,
 			linkToLastVersion,
 			linkToNewVersion,
 			linkToCommitHistory,
 		};
```

This applies to many other variables

source/features/useful-not-found-page.tsx
```diff
@@ -29,71 +27,247 @@ async function checkAnchor(anchor: HTMLAnchorElement): Promise<void> {
...
-	const bar = <h2 className="container mt-4 text-center"/>;
+	// Else, check if the file exists in commit history
+	const lastCommitIncludingFile = await getLastCommitForFile();
+	if (lastCommitIncludingFile) {
+		return lastCommitIncludingFile;
+	}
+
+	// Else, check if the file exists in the default branch's history
+	const linkToCommitHistoryOnDefaultBranch
+		= await getLinkToCommitHistoryOnDefaultBranch();
```
fregante, line 205: This seems identical to `getLastCommitForFile`, in practice, except that it's for a different branch. We just need to verify whether it ever existed (via the single v3 call I mentioned in the other comment). I think the whole `getLinkToCommitHistoryOnDefaultBranch` can be dropped.

source/features/useful-not-found-page.tsx
fregante: This chunk, like others, is still affected by the PR. Almost every comment has been moved up by a line

source/features/useful-not-found-page.tsx
```diff
@@ -29,71 +27,247 @@ async function checkAnchor(anchor: HTMLAnchorElement): Promise<void> {
...
+			<p className="container mt-4 text-center">
+				<a href={eventThatHappened.linkToLastVersion}>This file</a> was renamed to <a href={eventThatHappened.linkToNewVersion}>{eventThatHappened.newFilePath}</a> (<a href={eventThatHappened.linkToCommit}><relative-time datetime={new Date(eventThatHappened.commitDate!)}/></a>) - see <a href={eventThatHappened.linkToCommitHistory}>commit history</a>.
+			</p>,
+		);
+	}
+}
+
+async function showWhatIsMissingInPath(bar: Element): Promise<void> {
+	// Get the current URL
+	const parts = parseCurrentURL();
```
fregante, line 260: If you add comments ensure that they are clear, _correct_, and that aren't just rephrasing the code. This one is rephrasing the code (parseCurrentURL === "get the current URL"), others are wrong or vague ("Parse that and do/show something"). If they don't add anything useful, they should not exist

fregante (commented):
A few more comments, we'll get through this eventually ☺️

source/features/useful-not-found-page.tsx
gamemaker1: Yes; I had used the v3 call in the beginning, replaced it with v4 as per your suggestion later. I'll revert that.

source/features/useful-not-found-page.tsx
gamemaker1: Ok, will do

source/features/useful-not-found-page.tsx
gamemaker1: Sure, I'll go through the code and remove them if needed

gamemaker1:
Hi there!

I kind of rewrote the whole file again, keeping all your comments in mind. It's down to ~220 lines (74 lines more than the [original file](https://github.com/sindresorhus/refined-github/blob/main/source/features/useful-not-found-page.tsx)). I have removed some unneccessary comments, used v3 for the two API calls made and bundled most of the logic into one `showHelpfulLinks` function.

Thanks for being patient! Hopefully this PR will be ready for final review soon!

fregante (commented):
There are still unrelated changes and **some pre-existing comments have now been removed**. You can open the files tab yourself to see where these unnecessary changes have been made. They just slow reviews down because now I have to also check them unnecessarily.

gamemaker1:
Sorry about that. I took the original file and copy-pasted the changes over to it; there should be a bit less of unnecessary stuff now.

source/features/useful-not-found-page.tsx
```diff
@@ -107,17 +62,132 @@ function init(): false | void {
...
+				branch: commitInfo.parents[0].sha,
+				filePath,
+			});
+
+			return {
+				type: 'renamed',
+				commitDetails: {
+					filePath: fileInfo.previous_filename, newFilePath: fileInfo.filename,
+					commitSha, commitDate,
+					linkToCommit, linkToCommitHistory,
```
fregante, line 155: Don't use tricks just to save lines, that defeats the purpose. Lines of code are just a metric to quantify the amount of code, this just hides code from the metric.

gamemaker1: I think I got a bit carried away with the lines thing 😅 . Will change that

source/features/useful-not-found-page.tsx
```diff
@@ -107,17 +62,132 @@ function init(): false | void {
...
+
+	const change = await (async (): Promise<Record<string, any> | void> => {
+		const url = new GitHubURL(location.href);
+		const currentBranch = url.branch;
+		const {filePath} = url;
+		if (!currentBranch || !filePath) {
+			return;
+		}
+
+		const commits = await api.v3(`commits?path=${filePath}&sha=${currentBranch}&per_page=2`);
```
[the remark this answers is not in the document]
gamemaker1: Yes, will change

source/features/useful-not-found-page.tsx
```diff
@@ -107,17 +62,132 @@ function init(): false | void {
...
+		if (!commits[0]) {
+			return;
+		}
+
+		const commitInfo = await api.v3(`commits/${commits[0].sha as string}`);
+		const fileInfo = commitInfo.files.find((file: AnyObject) => [file.filename, file.previous_filename].includes(filePath));
+		if (!fileInfo) {
+			return;
```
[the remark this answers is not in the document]
gamemaker1: Ok, sure

source/features/useful-not-found-page.tsx
```diff
@@ -107,17 +62,133 @@ function init(): false | void {
...
+function init(): void {
+	if (
+		parseCurrentURL().length <= 1 || !select.exists('[alt*="This is not the web page you are looking for"]')
+	) {
+		return;
 	}
+
+	const bar = <h2 className="container mt-4 text-center"/>;
+	void showMissingPart(bar);
+	void showHelpfulLinks(bar);
```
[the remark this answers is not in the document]
gamemaker1: I think it is easier to `select` the `#parallax_illustration` element and keep appending using `bar.after`. Otherwise we could give an id (maybe `missing-parts-bar`) to our bar element and and add append elements below that. What do you think?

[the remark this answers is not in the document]
fregante: Nah, forget `bar`, it should just be used in one place. Every time we add an element we just use: `select('main').append(x)` or something like that. I only happens twice in the whole file anyway.

source/features/useful-not-found-page.tsx
```diff
@@ -36,55 +38,112 @@ function parseCurrentURL(): string[] {
...
+								oid
+							}
+						}
+					}
+				}
+			}
+		}
+	`);
+	const commits =
+		commitsResponseObject.repository.ref.target.history.nodes[0];
```
fregante, line 65: Disable Prettier entirely on this project if you haven't

```suggestion
	const commits = commitsResponseObject.repository.ref.target.history.nodes[0];
```

source/features/useful-not-found-page.tsx
```diff
@@ -36,55 +38,112 @@ function parseCurrentURL(): string[] {
...
-// If the resource was deleted, link to the commit history
-async function addCommitHistoryLink(bar: Element): Promise<void> {
-	const parts = parseCurrentURL();
-	parts[2] = 'commits';
-	const url = '/' + parts.join('/');
-	if (await is404(location.origin + url)) {
+async function getLatestChangeToFile(): Promise<Record<string, any> | void> {
+	const url = new GitHubURL(location.href);
+	const currentBranch = url.branch;
+	const {filePath} = url;
```
fregante, line 44:
```suggestion
	const {branch: currentBranch, filePath} = url;
```

or even just this, they're all "current"
```suggestion
	const {branch, filePath} = url;
```

source/features/useful-not-found-page.tsx
```diff
@@ -36,55 +38,112 @@ function parseCurrentURL(): string[] {
...
+				filePath: fileInfo.previous_filename,
+				newFilePath: fileInfo.filename,
+				commitSha,
+				commitDate,
+				linkToCommit,
+				linkToCommitHistory,
+				linkToLastVersion: url.toString(), linkToNewVersion: fileInfo.blob_url,
+			},
+		};
+	}
```
fregante, line 120: This style isn't working, perhaps it should just return `{fileInfo, commitInfo}` and then create the DOM directly in `showHelpfulLinks`. 

One thing that will help in there is to generate the shared DOM chunk only once:

```jsx
const lastVersion = <a href={change.commitDetails.linkToLastVersion}>The file</a>;
const permalink = (
	  <a href={change.commitDetails.linkToCommit}>
		  <relative-time datetime={change.commitDetails.commitDate}/>
	  </a>
);

const commitHistory = <a href={change.commitDetails.linkToCommitHistory}>Commit history</a>;

const verb = status === 'removed'
	? deleted
	: <a href={change.commitDetails.linkToNewVersion}>moved</a>

select(…).append(
	<p className="container mt-4 text-center">
		${lastVersion} was {verb} (${permalink}) - ${commitHistory}.
	</p>		
)
```

1. 👆 no extra text before "Commit history".
2. Don't create the intermediary `change` object, my code is only a demo.

source/features/useful-not-found-page.tsx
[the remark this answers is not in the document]
gamemaker1: Done! I used `select('h2')!.after(...)` instead of `select('main')!.after(...)` to keep the text right below the bar.

gamemaker1:
Here are a couple screenshots of the 404 page after all the changes:

<details>
<summary> Deleted </summary>

refined-github-moved-or-deleted-file

</details>

<details>
<summary> Renamed </summary>

refined-github-renamed-file

</details>

<details>
<summary> File on default branch </summary>

refined-github-non-existent-branch-and-file-on-default-branch

</details>

Please let me know if there are any more changes/improvements to make.

Thanks!

source/features/useful-not-found-page.tsx
```diff
@@ -107,19 +123,55 @@ function init(): false | void {
...
+		const url = new GitHubURL(location.href);
+
+		url.assign({route: 'commits'});
+		const commitHistory = <a href={url.toString()}>Commit history</a>;
+		url.assign({route: 'blob', branch: commitInfo.parents[0].sha, filePath: url.filePath});
+		const lastVersion = <a href={fileInfo.status === 'removed' ? fileInfo.blob_url : url.toString()}>This {getType()}</a>;
+		const permalink = <a href={commitInfo.html_url}><relative-time datetime={commitInfo.commit.committer.date}/></a>;
+		const verb = fileInfo.status === 'removed'
+			? 'deleted'
+			: <a href={fileInfo.status === 'renamed' ? fileInfo.blob_url : url.toString()}>moved</a>;
```
fregante, line 146: What is the status other than removed or renamed? Is this a wrong copy-paste from the `lastVersion` line?

source/features/useful-not-found-page.tsx
```diff
@@ -107,19 +123,55 @@ function init(): false | void {
...
-		void addDefaultBranchLink(bar);
+	const change = await getLatestChangeToFile();
+	if (change) {
+		const {fileInfo, commitInfo} = change;
+		const url = new GitHubURL(location.href);
+
+		url.assign({route: 'commits'});
+		const commitHistory = <a href={url.toString()}>Commit history</a>;
+		url.assign({route: 'blob', branch: commitInfo.parents[0].sha, filePath: url.filePath});
+		const lastVersion = <a href={fileInfo.status === 'removed' ? fileInfo.blob_url : url.toString()}>This {getType()}</a>;
```
fregante, line 142:
```suggestion
		const lastVersionUrl = fileInfo.status === 'removed' ? fileInfo.blob_url : url.toString();
		const lastVersion = <a href={lastVersionUrl}>The {getType()}</a>;
```

I generally don't like single-use variables but here it's better than having to wrap the JSX (which is a little long)

source/features/useful-not-found-page.tsx
```diff
@@ -107,19 +123,55 @@ function init(): false | void {
...
+}
 
-	if (['tree', 'blob'].includes(parts[2])) {
-		// Object might be 410 Gone
-		void addCommitHistoryLink(bar);
+async function showHelpfulLinks(): Promise<void> {
+	// What we should try to do to be helpful:
+	//  1. Check if the file is on the default branch
+	//  2. Check the current branch's history for the file (was it deleted or renamed)
+
```
fregante, line 123:
```suggestion
```

source/features/useful-not-found-page.tsx
```diff
@@ -107,19 +123,55 @@ function init(): false | void {
...
+		);
+		return;
 	}
 
-	if (['tree', 'blob', 'edit'].includes(parts[2])) {
-		// File might not be available on the current branch
-		// GitHub already redirects /tree/ and /blob/ natively
-		void addDefaultBranchLink(bar);
+	const change = await getLatestChangeToFile();
+	if (change) {
```
fregante, line 135: Prefer early returns to reduce indentation
```suggestion
	if (!change) {
		return;
	}

```

source/features/useful-not-found-page.tsx
```diff
@@ -36,54 +38,59 @@ function parseCurrentURL(): string[] {
...
+	const commitInfo = await api.v3(`commits/${commits[0].oid as string}`);
+	const fileInfo = commitInfo.files.find((file: AnyObject) => [file.filename, file.previous_filename].includes(filePath));
+	if (!fileInfo) {
 		return;
 	}
 
-	parts[3] = defaultBranch; // Change branch
-	const url = '/' + parts.join('/');
-	if (await is404(location.origin + url)) {
+	return {fileInfo, commitInfo};
```
fregante, line 76:
```js
if (fileInfo) return {fileInfo, commitInfo}
```

I'd probably just drop the `.find` as well and just use something like:

```js
for (fileInfo of commitInfo.files) {
	if ([file.filename, file.previous_filename].includes(filePath)) {
		return {fileInfo, commitInfo}
	}
}
```

gamemaker1:
Tested it just now, it works!

source/features/useful-not-found-page.tsx
```diff
@@ -36,54 +38,59 @@ function parseCurrentURL(): string[] {
...
 
-// If the resource was deleted, link to the commit history
-async function addCommitHistoryLink(bar: Element): Promise<void> {
-	const parts = parseCurrentURL();
-	parts[2] = 'commits';
-	const url = '/' + parts.join('/');
-	if (await is404(location.origin + url)) {
+async function getLatestChangeToFile(): Promise<Record<string, any> | void> {
+	const url = new GitHubURL(location.href);
+	const {branch, filePath} = url;
```
fregante, line 43: I think it's not used anywhere else
```suggestion
	const {branch, filePath} = new GitHubURL(location.href);
```

source/features/useful-not-found-page.tsx
```diff
@@ -36,54 +38,59 @@ function parseCurrentURL(): string[] {
...
-async function addDefaultBranchLink(bar: Element): Promise<void> {
-	const parts = getCleanPathname().split('/');
-	const branch = parts[3];
-	if (!branch) {
+	const commitsResponseObject = await api.v4(`
+		repository() {
+			ref(qualifiedName: "${branch}") {
+				target {
+					... on Commit {
+						history(first: 1, path: "${filePath}") {
```
fregante, line 53: Every time I see this I wonder if it can can done in fewer steps. Can you look for a better solution? It probably doesn't exist, but…

fregante (commented):
Looking great!

source/features/useful-not-found-page.tsx
gamemaker1: Yes it's another copy paste error

source/features/useful-not-found-page.tsx
gamemaker1: Ok, sure

source/features/useful-not-found-page.tsx
gamemaker1: Sure, I'll try my best

source/features/useful-not-found-page.tsx
```diff
@@ -107,17 +114,53 @@ function init(): false | void {
...
-		// GitHub already redirects /tree/ and /blob/ natively
-		void addDefaultBranchLink(bar);
+function init(): false | void {
+	const parts = parseCurrentURL();
+	if (parts.length <= 1 || !select.exists('[alt*="This is not the web page you are looking for"]')) {
+		return false;
 	}
+
+	void showMissingPart();
+	void showHelpfulLinks();
```
fregante, line 163: To speed things up, let's just split `showHelpfulLinks` into 2 unrelated functions. Even if they end up adding 2 lines, that's fine, it's extra information, it's all good. However we'll also need to **replace every `.after()` call** with something that `append`s  to an element so that we never risk moving existing elements. I think this works:

```js
select('main > .container-lg').before()
```

gamemaker1: Sure. Something like `showDefaultBranchLink` and `showAlternateFileLink`?

source/features/useful-not-found-page.tsx
```diff
@@ -107,17 +112,50 @@ function init(): false | void {
...
+	const change = await getLatestChangeToFile();
+	if (!change) return;
+
+	const {fileInfo, commitInfo} = change;
+	const url = new GitHubURL(location.href);
+
+	url.assign({route: 'commits'});
+	const commitHistory = <a href={url.toString()}>Commit history</a>;
+	url.assign({route: 'blob', branch: commitInfo.parents[0].sha, filePath: url.filePath});
+	const lastVersionUrl = fileInfo.status === 'removed' ? fileInfo.blob_url : url.toString();
```
fregante, line 137: I haven't tested this, but isn't `blob_url` also a 404 here?

gamemaker1: `blob_url` will be a link to the new file in case the file was renamed, but it will be the link to the last known version of the file if it has been deleted. In any case, I don't think it will be a 404.

source/features/useful-not-found-page.tsx
```diff
@@ -132,7 +132,7 @@ async function showHelpfulLinks(): Promise<void> {
 	const url = new GitHubURL(location.href);
 
 	url.assign({route: 'commits'});
-	const commitHistory = <a href={url.toString()}>Commit history</a>;
+	const commitHistory = <a href={url.toString()}>see commit history</a>;
```
fregante, line 137: ?

gamemaker1: Whoops ctrl+z

source/features/useful-not-found-page.tsx
gamemaker1: Done!

source/features/useful-not-found-page.tsx
gamemaker1: I've been messing around with the API and I wasn not able retrieve a list of changed files unless I specifically retrieved a commit using its `sha`/`oid`. I could ask on SO, if someone has a better solution.

source/features/useful-not-found-page.tsx
```diff
@@ -36,54 +38,57 @@ function parseCurrentURL(): string[] {
...
+								oid
+							}
+						}
+					}
+				}
+			}
+		}
+	`);
+	const commits = commitsResponseObject.repository.ref.target.history.nodes;
+	if (!commits[0]) {
```
yakov116, line 63:
```suggestion
	const [commits] = commitsResponseObject.repository.ref.target.history.nodes;
	if (!commits) {
```

I think this should work

source/features/useful-not-found-page.tsx
```diff
@@ -36,54 +38,57 @@ function parseCurrentURL(): string[] {
...
 		return;
 	}
 
-	const defaultBranch = await getDefaultBranch();
-	if (branch === defaultBranch) {
-		return;
+	// API v4 doesn't support retrieving a list of changed files for a commit:
+	// https://github.community/t/graphql-api-get-list-of-files-related-to-commit/14047/2
+	const commitInfo = await api.v3(`commits/${commits[0].oid as string}`);
```
yakov116, line 69:
```suggestion
	const commitInfo = await api.v3(`commits/${commits.oid as string}`);
```

source/features/useful-not-found-page.tsx
```diff
@@ -7,6 +7,8 @@ import * as pageDetect from 'github-url-detection';
 import features from '.';
 import getDefaultBranch from '../github-helpers/get-default-branch';
 import {getCleanPathname} from '../github-helpers';
+import * as api from '../github-helpers/api';
+import GitHubURL from '../github-helpers/github-url';
```
yakov116, line 11:
```suggestion
import features from '.';
import * as api from '../github-helpers/api';
import GitHubURL from '../github-helpers/github-url';
import getDefaultBranch from '../github-helpers/get-default-branch';
import {getCleanPathname} from '../github-helpers';
```

We sort the from like a tree

yakov116 (changes requested):
Few nit picks

source/features/useful-not-found-page.tsx
gamemaker1: It did, thanks

source/features/useful-not-found-page.tsx
gamemaker1: Done

source/features/useful-not-found-page.tsx
```diff
@@ -36,54 +38,57 @@ function parseCurrentURL(): string[] {
...
-		</p>,
-	);
-}
-
-// If the resource exists in the default branch, link to it
-async function addDefaultBranchLink(bar: Element): Promise<void> {
-	const parts = getCleanPathname().split('/');
-	const branch = parts[3];
-	if (!branch) {
+	const commitsResponseObject = await api.v4(`
```
yakov116, line 47:
```suggestion
	const {repository: commitsResponseObject} = await api.v4(`
```

@fregante any reason were not doing this?

fregante: No reason. It can just use the `repository` name though, otherwise we save no characters

source/features/useful-not-found-page.tsx
fregante: This would now be a single commit, not commits. The variable must be renamed

source/features/useful-not-found-page.tsx
fregante: This would now be a single commit, not commits. The variable must be renamed

gamemaker1: Done

source/features/useful-not-found-page.tsx
gamemaker1: Done; changed it to `const {repository} = ...`.

source/features/useful-not-found-page.tsx
fregante: It may be possibly by using `object()` instead of `ref().target`

https://stackoverflow.com/a/55509589/288906

gamemaker1: It worked, thanks!

gamemaker1:
Added all the changes requested above (including the nitpicks as well as `ref.target` => `object` change in the GraphQL query). Let me know if there are any more!

Thanks!

source/features/useful-not-found-page.tsx
```diff
@@ -36,54 +38,55 @@ function parseCurrentURL(): string[] {
...
+	if (!commit) {
 		return;
 	}
 
-	const defaultBranch = await getDefaultBranch();
-	if (branch === defaultBranch) {
-		return;
+	// API v4 doesn't support retrieving a list of changed files for a commit:
+	// https://github.community/t/graphql-api-get-list-of-files-related-to-commit/14047/2
+	const commitInfo = await api.v3(`commits/${commit.oid as string}`);
```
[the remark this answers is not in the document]
gamemaker1: We could, but we require some info (commit date and time, url to last known version of file, parent sha in case of rename, url to view commit) from the `commitInfo` object as well.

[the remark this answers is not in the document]
yakov116: If you are talking about the types, you can add what ever types are missing to it.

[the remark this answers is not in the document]
gamemaker1: Sorry, I didn't understand. Should I replace the API call and comment I'm using with the function, and modify the function to return the `commitInfo` as well?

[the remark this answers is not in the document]
yakov116: > Sorry, I didn't understand. Should I replace the API call and comment I'm using with the function

Yes



> and modify the function to return the `commitInfo` as well?

Not sure I am understanding this comment.

The API is returning the same as you are here, its the same call. 
If you need more info from the type, just add it to the interface there.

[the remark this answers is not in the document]
gamemaker1: I think we could rename the current function to `getLatestCommitForFile` and return only `commitInfo`. The caller can parse the object to get the exact change to the file (`fileInfo`), by calling another function `getChangeToFileInCommit`. Does that sound ok?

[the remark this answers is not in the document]
yakov116: I am lost here.

All I am asking it to rename and export

https://github.com/sindresorhus/refined-github/blob/68abae5ea744ff82ea9682c7d9cefc1949805aee/source/features/follow-file-renames.tsx#L16-L20

And replace
https://github.com/gamemaker1/refined-github/blob/cf6ab76a30be695a43681cfc9936b8d50dd544ca/source/features/useful-not-found-page.tsx#L65-L67

With `const commitInfo = await whatEverYouNamedTheFunction()`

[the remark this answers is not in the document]
gamemaker1: Sorry, let me try explaining it more clearly:

To display the improved 404 page, we need the following data:
	R1. list of files that was changed by the commit
	R2. datetime of the commit
	R3. URL to the commit page
	R4. parent commit sha

The function `findRename` from `follow-file-renames.tsx` provides us only R1. 

To get R2-4 as well, we could either:
O1. Create a separate utility function 'getCommitInfo' in 'github-helpers' (code below), OR
O2. Make the api.v3(...) call in the useful-not-found-page itself.

```ts
async function getCommitInfo(commitSha: string): Promise<Commit> { 
	// API v4 doesn't support it: https://github.community/t/what-is-the-corresponding-object-in-graphql-api-v4-for-patch-which-is-available-in-rest-api-v3/13590 
	const commit = await api.v3(`commits/${commitSha}`);
	return commit;
} 
```

What option would you suggest we go with?

Thanks!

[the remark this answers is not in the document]
gamemaker1: Thanks for pushing the change! Let me know if there is anything else to do before merging.

yakov116 (approved):
LGTM

I tested and it works great!

@fregante any comments?

fregante:
Types can be improved but let’s get this in for now.

fregante:
I’m tagging PRs with “extra wishes” with the Please label, just in case we want to circle back to them in the near future 😬

fregante:
Thanks @gamemaker1!

gamemaker1:
Thanks a lot for your help and guidance on this PR @fregante @yakov116!

source/features/useful-not-found-page.tsx
```diff
@@ -36,54 +39,53 @@ function parseCurrentURL(): string[] {
...
 
-	const defaultBranch = await getDefaultBranch();
-	if (branch === defaultBranch) {
-		return;
+	const commitInfo = await getCommitInfo(commit.oid as string);
+	for (const fileInfo of commitInfo.files) {
+		if ([fileInfo.filename, fileInfo.previous_filename].includes(filePath)) {
+			return {fileInfo, commitInfo};
+		}
 	}
```
fregante, line 71: Can you move this into `getCommitInfo` itself? Both usages look the files array to find one of the files. If it's not found, it should return `undefined`

```ts
function getRenamedFile(commitHash, filePath): Record<string, any> | undefined
```

Also ideally we should return a more specific object than `Record<string, any>`

gamemaker1: Moved that part into the `getRenamedFile` function, renamed that function to `getChangesToFileInCommit(sha, filePath)` and added a `CommitInfo` type in [my fork](https://github.com/gamemaker1/refined-github/commit/68654239c21cf3c583b5e7ae006323c6790fd383), should I open a new PR for this?

fregante: Yes

PR #4650, opened 2021-08-10
Support `kind:bug` label in `bugs-tab`

omBratteng:
Related to #4592

Adds the label "kind:bug" to the `bugs-tab`, as kind is another form of writing type

## Test URLs
https://github.com/vercel/next.js/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3A%22kind%3A+bug%22


fregante (approved):

yakov116 (approved):

yakov116:
Thanks @omBratteng

kidonng:
It's interesting the next.js repo uses both "type:foo" and "kind:bar", I wonder if it's just a remaining mistake (which means we may encounter more in the future?)

PR #4651, opened 2021-08-10
Fix `quick-repo-deletion` not aborting on Firefox

### LINKED ISSUE #4599: Aborting after clicking "Delete fork" not possible
### Please ensure:

### Description

It's not possible to stop the deletion process after clicking the "Delete Fork" button in the top right corner.

### Screenshot

screenshot

### Console errors

_No response_

### Example URL

https://github.com/edisim/outline/tree/master

### Browser(s) used

Firefox Developer Edition 91.0b3 (64-Bit) on Windows 10

cheap-glitch:
Fixes #4599 

## Test URLs
A fork you own or an empty repo with no stars

## Screenshot
https://user-images.githubusercontent.com/46634000/128853508-fdde8f28-4786-465f-ac45-6f71d6178b5e.mp4

Turns out the z-index was not the problem, rather it was the controller that wasn't aborted.\
**Tested on Firefox only**

### DIFF
--- a/source/features/quick-repo-deletion.tsx
+++ b/source/features/quick-repo-deletion.tsx
@@ -58,7 +58,9 @@ async function verifyScopesWhileWaiting(abortController: AbortController): Promi
 async function buttonTimeout(buttonContainer: HTMLDetailsElement): Promise<boolean> {
 	// Sync AbortController and DOM state
 	const abortController = new AbortController();
-	buttonContainer.addEventListener('toggle', abortController.abort, {once: true});
+	buttonContainer.addEventListener('toggle', () => {
+		abortController.abort();
+	}, {once: true});
 	abortController.signal.addEventListener('abort', () => {
 		buttonContainer.open = false;
 	}, {once: true});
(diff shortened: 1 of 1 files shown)


fregante:
Can we also just **change** it to a global click handler as I suggested in the issue? This should just never happen even if in the future z-index *is* the issue.

cheap-glitch:
> Can we also just **change** it to a global click handler as I suggested in the issue?

:+1: 

> because in the #4599’s screenshot it definitely is.

No, it's just what happens when the `<details>` element is closed but the controller isn't aborted (because the button style isn't reset properly).

https://user-images.githubusercontent.com/46634000/129029657-a0437bbb-9c66-4e88-b049-c1d9eeaeb0e8.mp4

fregante:
I think we should preserve the previous toggle/abort state sync. The "close on click anywhere" should be on top of that, but also it should be excluded via "if (is not aborted), preventDefault"

In short now I see this problem:

- if the user lets it complete, the first click after completion will be ignored, even if they click on the menu
- if the overlay is closed in any way other than click (Enter, esc? Whatever closes the detail), the click might not be caught

The code might not end up super clean since we'll end up having several listeners all two-way-linked to an AbortController, but this feature is something we need to take extra care of.

fregante:
This probably still works better than before. We can publish this in the upcoming version and improve it afterwards.

PR #4656, opened 2021-08-11
Meta: Log API calls and other fetches

yakov116:
image

Added to:

- api.v3
- api.v4
- fetchDom

I added as a new option since having it on by default is going to be annoying. Including in the other logging option does not make any sense.

### DIFF
--- a/source/github-helpers/api.ts
+++ b/source/github-helpers/api.ts
@@ -48,6 +48,13 @@ interface RestResponse extends AnyObject {
 	ok: boolean;
 }
 
+async function log(message: string) {
+	const {logAPI} = (await optionsStorage.getAll());
+	if (logAPI) {
+		console.log(message);
+	}
+}
+
 export const escapeKey = (value: string | number): string => '_' + String(value).replace(/[ ./-]/g, '_');
 
 export class RefinedGitHubAPIError extends Error {
@@ -173,6 +180,10 @@ export const v4 = mem(async (
 	}
 
 	query = query.replace('repository() {', () => `repository(owner: "${getRepo()!.owner}", name: "${getRepo()!.name}") {`);
+
+	void log(`{
+		${query}
+	}`);
 
 	const response = await fetch(api4, {
 		headers: {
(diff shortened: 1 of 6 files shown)


yakov116:
@fregante this logger is scaring me. I did not realize how many API requests we make!

fregante:
Instead of a public option I was suggesting it to be enabled only for local installs (so only for contributors) and with console.debug, which can be hidden in the console

yakov116:
> console.debug calls can be hidden in the console so since it's enabled only for local installs it shouldn't be much a problem.> 

But I don't like that idea, for the same reason. That means that I have more things in my console that I do not need. (I need to turn my console setting to verbose)


> > I did not realize how many API requests we make!
> 
> Wait till we do the same for every fetch!

I think I added it all over, do you see anything missing?

fregante:
> I think I added it all over

"Wait until …!" is not an invitation 😅

yakov116:
@fregante any comments before we merge?

source/github-helpers/api.ts
```diff
@@ -48,6 +48,13 @@ interface RestResponse extends AnyObject {
 	ok: boolean;
 }
 
+export async function log(message: string): Promise<void> {
```
fregante, line 51: We should:

1. move this out of here and into features/index
2. conform its style to the existing logging function — i.e. it should query the options once, not in every call

source/github-helpers/api.ts
```diff
@@ -174,6 +182,10 @@ export const v4 = mem(async (
 
 	query = query.replace('repository() {', () => `repository(owner: "${getRepo()!.owner}", name: "${getRepo()!.name}") {`);
 
+	void log(`{
```
fregante, line 185: Let’s also add the `query` word to this so it can actually be copy-pasted

source/options-storage.ts
```diff
@@ -15,6 +15,7 @@ const defaults = Object.assign({
 	customCSS: '',
 	personalToken: '',
 	logging: false,
+	logAPI: false,
```
fregante, line 18: More like logHTTP

source/github-helpers/api.ts
yakov116: Where is that required? In the explorer it not.

fregante:
I disabled the automatic release because there isn't much to release. We'll do it manually when worth it

PR #4667, opened 2021-08-14
Meta: Make `Log API calls in the console` option checkbox visible

### LINKED ISSUE #4666: No checkbox next to the new `Log API calls in the console` option label
### Please ensure:

### Description

In today's release 21.8.14, PR #4656 was merged.

Unfortunately, as you open the extension's Options, you'll notice there's no checkbox next to its "Log API calls in the console" label (see screenshot below). 

I also tested it in a fresh Chrome/Firefox profile, with only RG (w/ token).
There's no error in console.

### Console errors

```shell
```

### Example URL

Any page, e.g. this one

### Browser(s) used

Chrome 92, Firefox 91

darkred:
Closes #4666 

## Test URLs

Any page

## Note to reviewers:

Was it `hidden` on purpose? But, having a label without a checkbox seems unfinished/broken.

### DIFF
--- a/source/features/index.tsx
+++ b/source/features/index.tsx
@@ -292,8 +292,8 @@ const features = {
 export default features;
 
 export async function logQuery(message: string): Promise<void> {
-	const {logAPI} = (await optionsStorage.getAll());
-	if (logAPI) {
+	const {logHTTP} = (await optionsStorage.getAll());
+	if (logHTTP) {
 		console.log(message);
 	}
 }
--- a/source/features/index.tsx
+++ b/source/features/index.tsx
@@ -43,8 +43,6 @@ interface InternalRunConfig {
 	onlyAdditionalListeners: boolean;
 }
 
-let log: typeof console.log;
-export let logHTTP: typeof console.log;
 const {version} = browser.runtime.getManifest();
 
 let logError = (id: FeatureID, error: unknown): void => {
@@ -101,10 +99,6 @@ const globalReady: Promise<RGHOptions> = new Promise(async resolve => {
 		Object.assign(options, localHotfixes);
 	}
 
-	// Create logging function
-	log = options.logging ? console.log : () => {/* No logging */};
-	logHTTP = options.logHTTP ? console.log : () => {/* No logging */};
-
 	await waitFor(() => document.body);
 
 	if (pageDetect.is500() || pageDetect.isPasswordConfirmation()) {
@@ -134,6 +128,12 @@ const globalReady: Promise<RGHOptions> = new Promise(async resolve => {
 	resolve(options);
 });
 
+const log = {
+	info: console.log,
+	http: console.log,
+	error: logError,
+};
+
 const setupPageLoad = async (id: FeatureID, config: InternalRunConfig): Promise<void> => {
 	const {include, exclude, init, deinit, additionalListeners, onlyAdditionalListeners} = config;
 
@@ -147,7 +147,7 @@ const setupPageLoad = async (id: FeatureID, config: InternalRunConfig): Promise<
 			// Features can return `false` when they decide not to run on the current page
 			// Also the condition avoids logging the fake feature added for `has-rgh`
 			if (await init() !== false && id !== __filebasename) {
-				log('✅', id);
+				log.info('✅', id);
 			}
 		} catch (error: unknown) {
 			logError(id, error);
@@ -207,7 +207,7 @@ const add = async (id: FeatureID, ...loaders: FeatureLoader[]): Promise<void> =>
 	const options = await globalReady;
 	// Skip disabled features, unless the "feature" is the fake feature in this file
 	if (!options[`feature:${id}`] && id as string !== __filebasename) {
-		log('↩️', 'Skipping', id);
+		log.info('↩️', 'Skipping', id);
 		return;
 	}
 
@@ -286,6 +286,7 @@ const features = {
 	add,
 	addCssFeature,
 	error: logError,
+	log,
 	shortcutMap,
 	list: __features__,
 	meta: __featuresMeta__,
--- a/source/features/index.tsx
+++ b/source/features/index.tsx
@@ -44,6 +44,7 @@ interface InternalRunConfig {
 }
 
 let log: typeof console.log;
+export let logHTTP: typeof console.log;
 const {version} = browser.runtime.getManifest();
 
 let logError = (id: FeatureID, error: unknown): void => {
@@ -102,6 +103,7 @@ const globalReady: Promise<RGHOptions> = new Promise(async resolve => {
 
 	// Create logging function
 	log = options.logging ? console.log : () => {/* No logging */};
+	logHTTP = options.logHTTP ? console.log : () => {/* No logging */};
 
 	await waitFor(() => document.body);
 
@@ -290,11 +292,3 @@ const features = {
 };
 
 export default features;
-
-const settings = optionsStorage.getAll();
-export async function logQuery(message: string): Promise<void> {
-	const {logHTTP} = await settings;
-	if (logHTTP) {
-		console.log(message);
-	}
-}
--- a/source/features/index.tsx
+++ b/source/features/index.tsx
@@ -129,7 +129,7 @@ const globalReady: Promise<RGHOptions> = new Promise(async resolve => {
 
 	if (select.exists('body.logged-out')) {
 		console.warn('Refined GitHub is only expected to work when you’re logged in to GitHub. Errors will not be shown.');
-		features.error = () => {/* No logging */};
+		features.log.error = () => {/* No logging */};
 		logError = () => {/* No logging */};
 	}
 
@@ -289,7 +289,6 @@ void add(__filebasename, {
 const features = {
 	add,
 	addCssFeature,
-	error: logError,
 	log,
 	shortcutMap,
 	list: __features__,
--- a/source/features/index.tsx
+++ b/source/features/index.tsx
@@ -291,8 +291,9 @@ const features = {
 
 export default features;
 
+const settings = optionsStorage.getAll();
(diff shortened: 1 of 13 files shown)


fregante (approved):

fregante (changes requested):
It looks like the intent was to make it visible only during development, so the attribute should be moved to the P element and a line should be updated in options.tsx

darkred:
I think it should be ok now, I tested it and it works.

yakov116:
@darkred can you follow the rest of the feedback in https://github.com/sindresorhus/refined-github/pull/4656

Thanks

darkred:
I think I have followed the following suggestions correctly:

* [x] move this out of here and into features/index
* [x] conform its style to the existing logging function — i.e. it should query the options once, not in every call
* [x] Let’s also add the `query` word to this so it can actually be copy-pasted
* [x] More like logHTTP

&nbsp;
---

Regarding this one:
* [ ] make the type signature conform to console.log, not just a single string parameter

for reference, per https://developer.mozilla.org/en-US/docs/Web/API/console/log, the `console.log` syntax is:

```
console.log(obj1 [, obj2, ..., objN]);
console.log(msg [, subst1, ..., substN]);
```

The referenced line is:
```ts
export async function log(message: string): Promise<void> {
```

If I replace the type `string` with `...string` I'm getting: 
> Rest parameter 'string' implicitly has an 'any[]' type` error

And if I change it into `string | object` I'm getting: 
> Don't use `object` as a type. The `object` type is hard to use. Use `Record<string, unknown>` instead. See: https://github.com/typescript-eslint/typescript-eslint/pull/848 (@typescript-eslint/ban-types)

&nbsp;
Is `string | unknown` the correct type? 
*(see [The <code>unknown</code> Type in TypeScript](https://mariusschulz.com/blog/the-unknown-type-in-typescript))*

source/github-helpers/api.ts
```diff
@@ -29,6 +29,8 @@ import mem from 'mem';
 import * as pageDetect from 'github-url-detection';
 import {JsonObject, AsyncReturnType} from 'type-fest';
 
+import {logQuery} from '../features/index';
+
```
yakov116, line 33: Nit. Extra spance and don't forget to resort the from

yakov116:
I am not so sure how to do it.

Look in the api.ts file to see how @fregante did it there.
 

> * Let’s also add the `query` word to this so it can actually be copy-pasted

This was not done, But I do not think it should be. Since I the explorer does not need it, I don't know why we should add it. @fregante?

yakov116:
> * conform its style to the existing logging function — i.e. it should query the options once, not in every call

This was not done.

I have an idea how to do it, I will push to your branch later tonight.

source/github-helpers/api.ts
darkred: Ok, done.

darkred:
> > * [x] conform its style to the existing logging function — i.e. it should query the options once, not in every call
> 
> This was not done.
> 
> I have an idea how to do it, I will push to your branch later tonight.

I thought I did it by changing:

```ts
export async function logQuery(message: string): Promise<void> {
	const {logHTTP} = (await optionsStorage.getAll());
```

into:

```ts
const settings = optionsStorage.getAll();
export async function logQuery(message: string): Promise<void> {
	const {logHTTP} = await settings;
```

based on https://github.com/sindresorhus/refined-github/blob/ac8de6bd0ef0a6016732a69556e4eb9c014cfd38/source/github-helpers/api.ts#L67-L69

I guess I was wrong?

yakov116:
> I guess I was wrong?

Maybe I am wrong. I was thinking of a oneTime function.

Leave it as is 👍

darkred:
> > * Let’s also add the `query` word to this so it can actually be copy-pasted
> 
> This was not done, But I do not think it should be. Since I the explorer does not need it, I don't know why we should add it. @fregante?

I renamed the function `log` into `logQuery`. Wasn't this what was wanted?

yakov116:
> I renamed the function `log` into `logQuery`. Wasn't this what was wanted?

What I he meant was for API v4 add the word `query` to the console log.

I see that he marked the comment as resolved to skip it.

source/features/index.tsx
```diff
@@ -290,3 +290,11 @@ const features = {
 };
 
 export default features;
+
```
yakov116, line 293: I think it should be moved to https://github.com/sindresorhus/refined-github/blob/main/source/github-helpers/index.ts

source/github-helpers/api.ts
```diff
@@ -125,7 +118,7 @@ export const v3 = mem(async (
 	}
 
 	const url = new URL(query, api3);
-	void log(String(url));
+	void logQuery(String(url));
```
fregante, line 121: Same for all others
```suggestion
	logQuery(url);
```

source/github-helpers/index.ts
```diff
@@ -116,3 +117,11 @@ export async function isPermalink(): Promise<boolean> {
 export function isNotRefinedGitHubRepo(): boolean {
 	return !location.pathname.startsWith('/sindresorhus/refined-github/');
 }
+
+const settings = optionsStorage.getAll();
+export async function logQuery(message: string): Promise<void> {
```
fregante, line 122: I meant this function 

https://github.com/sindresorhus/refined-github/blob/ac8de6bd0ef0a6016732a69556e4eb9c014cfd38/source/features/index.tsx#L46


> 1. move this out of here and **into features/index**
> 2. conform its style to the **existing logging function**
> 3.  **make the type signature conform to console.log**, not just a single string parameter

fregante: Also if the option is logHTTP then the function should not be logQuery

fregante: Also https://github.com/sindresorhus/refined-github/blob/ac8de6bd0ef0a6016732a69556e4eb9c014cfd38/source/features/index.tsx#L103-L104

source/features/index.tsx
```diff
@@ -44,6 +44,7 @@ interface InternalRunConfig {
 }
 
 let log: typeof console.log;
+export let logHTTP: typeof console.log;
```
yakov116, line 47: I tried for over an hour. I do not know how to do this any other way.

XO does not like this. Any suggestions?

fregante:
```suggestion
let logHTTP: typeof console.log;

export {logHTTP}
```

Or just silence the error.

Also it would be a good opportunity to use a default:
```ts
let log = console.log;
```

yakov116: > Or instead of exporting it here it should be exported on the `features` object, so it can be used as `features.logHTTP()`

It wont allow a `let` to be exported

fregante: Better yet, we can have a single `log` object, like:

```ts
const log = {
	info: console.log,
	http: console.log,
	error: logError,
}
```

And then it can be overridden easily everywhere these methods are used. Could be a separate PR if you prefer.

darkred: @yakov116 I have followed all fregante's suggestions. Is it ok if I push here?

yakov116: Yes

darkred: Here. I tested it and it works ok.

yakov116: Where does it respect the user settings?

darkred: If I click "Clear cache" and then ticking "Log API calls in the console" then logging becomes enabled.
Likewise, clicking "Clear cache" and unticking it, makes logging disabled.
This is how I tested it.

Do you mean something else? Have I got it wrong?

yakov116: What if you dont clear? (You should have to)

darkred: You are right, thanks for noting.
Clearing cache is irrelevant. I had mistakenly removed two lines. 
I restored them and now it respects the setting again.

source/features/index.tsx
```diff
@@ -284,6 +290,7 @@ const features = {
 	add,
 	addCssFeature,
 	error: logError,
```
yakov116, line 292: I think what @fregante meant was to remove the `error` here and update where ever we use `features.error` to be `features.log.error`.

darkred: Thank you for the guidance. Done.

yakov116: Your welcome.
You forgot to remove the export.

darkred: Which export are you referring to?

In `features/index.tsx` there's only the initial one:

https://github.com/sindresorhus/refined-github/blob/3262b37f77f12f404e57f5434c2324f1ed41ffad/source/features/index.tsx#L292

which cannot be removed, and XO suggests to keep it default, i.e. not make it non-default.

yakov116: My bad everything is good, I misread the diff.

🎉

source/features/index.tsx
```diff
@@ -123,7 +129,7 @@ const globalReady: Promise<RGHOptions> = new Promise(async resolve => {
 
 	if (select.exists('body.logged-out')) {
 		console.warn('Refined GitHub is only expected to work when you’re logged in to GitHub. Errors will not be shown.');
-		features.error = () => {/* No logging */};
+		features.log.error = () => {/* No logging */};
 		logError = () => {/* No logging */};
```
fregante, line 133: If we use `features.log.error` exclusively, we no longer need to override `logError` too
```suggestion
		features.log.error = () => {/* No logging */};
```

fregante (approved):
Untested but this looks like exactly what I had in mind 🙌

source/features/index.tsx
darkred: I couldn't commit your suggestion because it shows "This suggestion is invalid because no changes were made to the code", so I did it manually. Thank you.

yakov116 (approved):
LGTM

fregante:
Thanks!

PR #4677, opened 2021-08-18
Fix multiple renames bug in `follow-file-renames` 

shinenelson:
When a file is renamed too often and there are not many commits in between, both buttons for `follow-file-renames` is supposed to be active, ideally.

However, in the current implementation, the global URL object from `init` is being passed and mutated in `linkify` which would mislead the linking logic in the subsequent runs of the loop for the `disabledPagination` buttons.

This change modifies `linkify` to add a new `GitHubURL` object within `linkify` context which is used to linkify the `disabledPagination` buttons. The global URL object is retained since it carries the URL / name of the current file which is required to match the file in comparison.

## Test URLs

The following URLs point to the life-cycle of a single file called `site.rb` in https://github.com/ForkAwesome/Fork-Awesome.

- https://github.com/ForkAwesome/Fork-Awesome/commits/master/src/doc/_plugins/site.rb
- https://github.com/ForkAwesome/Fork-Awesome/commits/master/src/_plugins/site.rb
- https://github.com/ForkAwesome/Fork-Awesome/commits/master/source/_plugins/site.rb
- https://github.com/ForkAwesome/Fork-Awesome/commits/master/build/_plugins/site.rb

The first URL should have only the **Older** button linkified, the second and third URLs should have both buttons linkified and the last URL should have only the **Newer** button linkified.


source/features/follow-file-renames.tsx
```diff
@@ -30,13 +30,14 @@ async function linkify(button: HTMLButtonElement, url: GitHubURL): Promise<void
 	for (const file of (files as File[])) {
 		if (file[fromKey] === url.filePath) {
```
fregante, line 31: Now `url` is only used for this, so `linkify` should probably just accept `filePath`

Also line 59 is lost so it should probably be moved to line ~33

fregante:
heh I tried reverting those 3 changes but I found another bug: `restore-file` does not work with force-pushes. 😅 

Can you rebase this PR or exclude those changes? If too difficult it's best to just open a new PR from `main`

fregante:
Thanks @yakov116

fregante (approved):
Looks good! 

Unrelated to this PR: One thing I'm not seeing is that we should probably skip the feature when we are on _folder_ commit list. Folders can't be renamed and don't appear in commits so I think the feature will never work there.

shinenelson:
> we should probably skip the feature when we are on folder commit list

I agree. However, I could not find a [detection](https://github.com/fregante/github-url-detection/blob/master/index.ts) that is specific for files ( or folders ). I created https://github.com/fregante/github-url-detection/issues/89 for it.

source/features/follow-file-renames.tsx
```diff
@@ -28,19 +28,22 @@ async function linkify(button: HTMLButtonElement, url: GitHubURL): Promise<void
...
 	for (const file of (files as File[])) {
-		if (file[fromKey] === url.filePath) {
+		if (file[fromKey] === filePath) {
 			if (file.status === 'renamed') {
-				url.assign({
+				const linkifiedURL = new GitHubURL(location.href);
+				linkifiedURL.assign({
 					route: 'commits',
 					filePath: file[toKey],
+					// Clear the search from the URL
```
yakov116, line 37: Don't think this comment is needed

yakov116 (approved):
Untested, LGTM

source/features/follow-file-renames.tsx
shinenelson: I thought the second part was obvious because the field was already inside of the link that was being passed to rename.

yakov116: The whole point of the comment is for the second part.

In other words "Why are we clearing the search".

yakov116:
@fregante merged since it is blocking another PR

source/features/follow-file-renames.tsx
```diff
@@ -34,7 +34,7 @@ async function linkify(button: HTMLButtonElement, filePath: string): Promise<voi
 				linkifiedURL.assign({
 					route: 'commits',
 					filePath: file[toKey],
-					// Clear the search from the URL
+					// Clear the search from the url, so it does not get passed to the rename link
```
fregante, line 37: why url tho

fregante:
Tested and it works 🎉 thanks Shiiiine

PR #4686, opened 2021-08-19
Improve progress notification style in `restore-file`

### LINKED ISSUE #4113: Use the new `toast` component in `restore-file`
`restore-file` changes the menu item inline

But instead of changing the menu, we should use the new toast introduced by https://github.com/sindresorhus/refined-github/pull/4016

gif

Some changes will be needed in the toast to show possible errors. Refer to https://primer.style/css/components/toasts for classnames and icons.

jayeshmann:
Closes #4113 
Added more toast states.
This change affects `https://github.com/[USER]/[REPO]/pull/[PR#]/files`. 

## Test URLs
https://github.com/jayeshmann/first_push/pull/1/files

## Screenshot
restore-toast


source/features/restore-file.tsx
```diff
@@ -49,11 +44,23 @@ async function commitFileContent(menuItem: Element, content: string, filePath: s
...
 		pathname = `/${nameWithOwner}/new/${headBranch}?filename=${filePath}`;
 	} else {
-		menuItem.textContent = 'Committing…';
+		void showToast(async () => {
+			// Fake a task
+		}, {
+			message: 'Committing changes…',
+			doneMessage: 'Changes committed!',
+			onDone: ToastOnDoneState.success,
+		});
```
fregante, line 63: This new interface and the fake task defeats the purpose of passing a task to showToast. Refer to my implementation suggestion to see how I'd expect this to work.

The problem is that `showTask` currently doesn't support show progress, so we can pass a callback function to the task to allow doing that.

For example:

```ts
async function restoreFile(progress: (message: string) => void) {
	await api.v4(…);
	progress('Committing…')
	await pushForm();
}
```

and the regular:

```ts
await showToast(restoreFile, {
	message: 'Restoring…',
	doneMessage: 'Restored'
})

File: source/features/restore-file.tsx [hunk 2]
```diff
@@ -79,32 +86,46 @@ async function handleRestoreFileClick(event: delegate.Event<MouseEvent, HTMLButt
 
 	filesRestored.add(menuItem);
 
-	menuItem.textContent = 'Restoring…';
 	event.preventDefault();
 	event.stopPropagation();
```
fregante, line 90: These should now be dropped because they're keeping the menu open

source/features/restore-file.tsx
fregante: Also any progress message should keep the loading gif, not change it to a "done" state, which should only appear when the task is done.

source/features/restore-file.tsx
jayeshmann: Alright, will remove both L#89 & L#90.

source/features/restore-file.tsx
jayeshmann: Thanks!

source/features/batch-mark-files-as-viewed.tsx
```diff
@@ -59,6 +59,7 @@ function onAltClick(event: delegate.Event<MouseEvent, HTMLInputElement>): void {
 			? 'Marking visible files as unviewed'
 			: 'Marking visible files as viewed',
 		doneMessage: 'Marking files completed',
+		onDone: ToastOnDoneState.success,
```
fregante, line 62: This should still be automatic, no change.  There should not be any onDone call back nor ondoneState

source/features/restore-file.tsx
```diff
@@ -79,32 +74,47 @@ async function handleRestoreFileClick(event: delegate.Event<MouseEvent, HTMLButt
 
 	filesRestored.add(menuItem);
 
-	menuItem.textContent = 'Restoring…';
-	event.preventDefault();
-	event.stopPropagation();
-
-	try {
-		const filePath = menuItem.closest<HTMLDivElement>('[data-path]')!.dataset.path!;
+	async function restoreFile(filePath: string, progress: (message: string) => void): Promise<void> {
```
fregante, line 77: Can you move this function to the top level? It should probably be merged with commitFileContent

source/features/restore-file.tsx
```diff
@@ -79,32 +74,47 @@ async function handleRestoreFileClick(event: delegate.Event<MouseEvent, HTMLButt
...
 	} catch (error: unknown) {
-		showError(menuItem, 'Restore failed. See console for details');
+		menuItem.disabled = true;
+		menuItem.style.background = 'none'; // Disables hover background color
+		void showToast(async () => {
+			// Fake a task to show the error message
+		}, {
+			doneMessage: 'Restore failed. See console for details',
+			onDone: ToastOnDoneState.error,
+		});
```
fregante, line 117: This is probably automatically handled by the toast. There should not be any empty task functions

source/features/restore-file.tsx
```diff
@@ -79,32 +74,47 @@ async function handleRestoreFileClick(event: delegate.Event<MouseEvent, HTMLButt
...
 		}
 
-		await commitFileContent(menuItem, file.text, filePath);
+		await commitFileContent(menuItem, file.text, filePath, progress);
+	}
 
+	try {
+		const filePath = menuItem.closest<HTMLDivElement>('[data-path]')!.dataset.path!;
+		// Show toast while restoring
+		const progress = async (message: string): Promise<void> => {
```
fregante, line 97: Not what I meant, refer to my previous example exactly. `progress` is a callback provided by showToast itself to allow its callback to change the text. 

See the line `return await task()`? It should be

```JavaScript 
return await task(message => {
   messageWrapper.textContent = message
})
```

fregante (changes requested):

source/features/batch-mark-files-as-viewed.tsx
jayeshmann: I added States for better flexibility in the future. Info State is the default.

source/features/restore-file.tsx
jayeshmann: Just for clarification, making one top level function and removing both of these?

source/features/restore-file.tsx
jayeshmann: Could you please elaborate further on this?

source/features/restore-file.tsx
jayeshmann: As far as I know, It is not automatically handled by the toast.

source/features/batch-mark-files-as-viewed.tsx
fregante: There's no need

source/features/restore-file.tsx
fregante: Both what?

source/features/restore-file.tsx
fregante: You're right, it's missing a catch after the `return await task()`

source/features/restore-file.tsx
fregante: I don't know what else to say. I gave the exact code to paste there and how to use it in my previous comment.

source/features/batch-mark-files-as-viewed.tsx
jayeshmann: Finally will be run even after catch, right?

jayeshmann: And I show Done in finally block

fregante: Then drop the `finally` and keep its code after the catch:

```js
try
  return await task()
catch (error) {
  messageWrapper.textContent = error.message
  throw error // Throw it again so it can be caught again when awaiting `showToast`
}

toast.classList.replace('Toast--loading', 'Toast--success');
messageWrapper.textContent = doneMessage;
iconWrapper.firstChild!.replaceWith(<CheckIcon/>);
```

jayeshmann: Okay, dropping `finally{ }`.

fregante: I pushed some changes to the toast file. It should probably not receive further changes other than fixing its types.

source/features/restore-file.tsx
jayeshmann: Trying to make this work, Unsuccessful so far.

jayeshmann:
It's not ready for review yet.

fregante:
Sorry I meant to click "Approve and run" 😅

source/features/restore-file.tsx
```diff
@@ -74,47 +87,23 @@ async function handleRestoreFileClick(event: delegate.Event<MouseEvent, HTMLButt
...
-			});
-		};
-
-		const task = restoreFile(filePath, progress);
+		// const task = restoreFile(menuItem, filePath);
 		// Hide file from view
-		await task;
+		// await task;
+		// Show toast while restoring
+		// await showToast(async () => restoreFile() , {message: 'Restoring…',doneMessage: 'Restored!'});
```
jayeshmann, line 96: This doesn't work as 
```js
await showToast(restoreFile, {message: 'Restoring…',doneMessage: 'Restored!'});
```

jayeshmann (commented):
@fregante Could you look into this?

jayeshmann:
Working expected: 
restore-toast1

This behavior was previously found in https://github.com/sindresorhus/refined-github/pull/4686/commits/d1d4b42e4b9664f55bad9ca7e8e438e3e7f4dc3b

fregante:
> Could you look into this?
> 
> ```js
> await showToast(async (progress) => restoreFile(progress,menuItem,filePath) , {message: 'Restoring…',doneMessage: 'Restored!'});
> ```

Yes that code looks about right since now `restoreFile` is at the top level 👍

source/features/restore-file.tsx
```diff
@@ -9,12 +9,7 @@ import features from '.';
 import * as api from '../github-helpers/api';
 import fetchDom from '../helpers/fetch-dom';
 import {getConversationNumber} from '../github-helpers';
-
-function showError(menuItem: HTMLButtonElement, error: string): void {
-	menuItem.disabled = true;
-	menuItem.style.background = 'none'; // Disables hover background color
-	menuItem.textContent = error;
-}
+import showToast from '../github-helpers/toast';
```
yakov116, line 12: Lint
Can you move this line one up?

jayeshmann:
> > Could you look into this?
> > ```js
> > await showToast(async (progress) => restoreFile(progress,menuItem,filePath) , {message: 'Restoring…',doneMessage: 'Restored!'});
> > ```
> 
> Yes that code looks about right since now `restoreFile` is at the top level 👍
> 
> I think that's just about the only thing missing from the current code ☺️

This didn't work, types are possibly wrong?

source/features/restore-file.tsx
jayeshmann: Interchange L#12 with L#11?

yakov116: Yes

jayeshmann:
Finally done!

source/features/restore-file.tsx
```diff
@@ -79,32 +87,19 @@ async function handleRestoreFileClick(event: delegate.Event<MouseEvent, HTMLButt
...
-		}
-
-		if (file.isTruncated) {
-			showError(menuItem, 'Restore failed: File too big');
-			return;
-		}
-
-		await commitFileContent(menuItem, file.text, filePath);
+		// Show toast while restoring
+		void showToast(async progress => restoreFile(progress, menuItem, filePath), {
```
fregante, line 93: We need to await this
```suggestion
		await showToast(async progress => restoreFile(progress, menuItem, filePath), {
```

source/github-helpers/toast.tsx
```diff
@@ -15,7 +15,7 @@ export function ToastSpinner(): JSX.Element {
 
 type Task = () => Promise<unknown>;
 export default async function showToast<TTask extends Task>(
-	task: TTask,
+	task: ((progress: (message: string) => void) => Promise<void>) | TTask,
```
fregante, line 18: Instead of changing it here, `TTask` should be updated. Probably:

```ts
type ProgressCallback = (message: string) => void;
type Task = (progress: ProgressCallback) => Promise<void>;
```

jayeshmann: Lint fails.
on 
```js
const result = await task((message: string): void => {
```

fregante: Let's wait for TypeScript help 😃 

I think that line doesn't need types though because they're inferred from `task()`  itself:

```ts
const result = await task(message => {
```

source/features/restore-file.tsx
```diff
@@ -79,32 +87,19 @@ async function handleRestoreFileClick(event: delegate.Event<MouseEvent, HTMLButt
...
+			message: 'Restoring…',
+			doneMessage: 'Restored!',
+		});
 
 		// Hide file from view
 		menuItem.closest('.file')!.remove();
 	} catch (error: unknown) {
-		showError(menuItem, 'Restore failed. See console for details');
+		menuItem.disabled = true;
+		menuItem.style.background = 'none'; // Disables hover background color
```
fregante, line 102: I think we can leave them alone now
```suggestion
```

For example it would allow the user to update the token in the options and retry without refreshing the page

jayeshmann: Alright!

source/github-helpers/toast.tsx
jayeshmann: Yes, it doesn't need. Where is TypeScript help?

jayeshmann:
Lint passed.

fregante:
gif

yakov116:
Shouldn't it have a min width? I don't like the size jump (just asking)

fregante:
Error works too:

gif

fregante:
> Shouldn't it have a min width?

Nah. https://primer.style/css/components/toasts

fregante (approved):

jayeshmann:
> Shouldn't it have a min width? I don't like the size jump (just asking)

Size jumping will catch the attention of the user. Better for notifying someone?

fregante:
धन्यवाद जयेश 🙏

PR #4689, opened 2021-08-20
Cleanup `profile-hotkey`

shinenelson:
There is no use for the hotkey when the user is already on their own profile.

### DIFF
--- a/source/features/profile-hotkey.tsx
+++ b/source/features/profile-hotkey.tsx
@@ -12,7 +12,7 @@ function init(): void {
 
 void features.add(__filebasename, {
 	shortcuts: {
-		'g m': 'Go to Profile',
+		'g m': 'Go to my profile',
 	},
 	exclude: [ isOwnUserProfile ],
 	awaitDomReady: false,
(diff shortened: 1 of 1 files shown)


fregante:
Thanks for the PR, but that's not how most of GitHub works, for example try `g c` multiple times in a row. Specifically it's not worth adding an extra DOM check just to avoid this.

If you want to improve this function though you can change the link generation to drop the origin altogether since `'/' + username` should work just as well.

source/features/profile-hotkey.tsx
```diff
@@ -14,6 +14,7 @@ void features.add(__filebasename, {
 	shortcuts: {
 		'g m': 'Go to Profile',
```
fregante, line 15: And also we can change
```suggestion
		'g m': 'Go to my profile',
```

shinenelson:
> but that's not how most of GitHub works, for example try g c multiple times in a row

eek! I did not realize that :see_no_evil: 

> If you want to improve this function though you can change the link generation to drop the origin altogether since '/' + username should work just as well.

sounds okay, but I think that would involve some testing. I noticed this as I was working on something else and thought it was an easy change.

I will change this pull request to a draft and modify it later.

source/features/profile-hotkey.tsx
```diff
@@ -12,8 +12,9 @@ function init(): void {
 
 void features.add(__filebasename, {
 	shortcuts: {
-		'g m': 'Go to Profile',
```
fregante, line 15: Let's just send this single line in a new PR if other copy-related changes are included

PR #4690, opened 2021-08-20
Drop `collapse-wiki-sections` feature

### LINKED ISSUE #4144: Change the `collapse-wiki-sections` trigger

shinenelson:
Decision reference : https://github.com/sindresorhus/refined-github/issues/4144#issuecomment-902202288

closes #4144


fregante (approved):

yakov116 (approved):

fregante:
Not really, they just added a feature that made this less useful

PR #4707, opened 2021-08-23
Exclude directories from follow-file-renames

### LINKED ISSUE #4683: Exclude directories from `follow-file-renames`
### Description

following from https://github.com/sindresorhus/refined-github/pull/4677#pullrequestreview-733587222

> we should probably skip the feature when we are on folder commit list.

quoting from https://github.com/fregante/github-url-detection/issues/89#issuecomment-901863779

> It can be determined from the DOM, ideally from some meta tags in the HEAD, otherwise we just need to look for specific /blob/ links on the page

### Example URL

- file commit page : https://github.com/ForkAwesome/Fork-Awesome/commits/master/src/doc/_plugins/site.rb
- directory commit page : https://github.com/ForkAwesome/Fork-Awesome/commits/master/src/doc/_plugins/

shinenelson:
resolves #4683

We cannot simply lookup the blob URL from the DOM because the `route` field for trees would still be 'blob' and we would have to make a `fetch` request and look for a redirect to determine whether the URL is a 'tree' or a 'blob'.

There is another 'tree' URL in the DOM that points to 'Browse the repository at the point in history' which is also unusable.

PS : I am sorry I took too long with this.

## Test URLs

from #4683

- file commit page : https://github.com/ForkAwesome/Fork-Awesome/commits/master/src/doc/_plugins/site.rb
- directory commit page : https://github.com/ForkAwesome/Fork-Awesome/commits/master/src/doc/_plugins/

## Screenshot

https://github.com/ForkAwesome/Fork-Awesome/commits/master/src/doc/_plugins/site.rb : 
image

https://github.com/ForkAwesome/Fork-Awesome/commits/master/src/doc/_plugins/ : 
image

### DIFF
--- a/source/features/follow-file-renames.tsx
+++ b/source/features/follow-file-renames.tsx
@@ -73,11 +73,15 @@ async function linkify(button: HTMLButtonElement, filePath: string): Promise<voi
 async function init(): Promise<void | false> {
 	const disabledPagination = select.all('.paginate-container button[disabled]');
 	const url = new GitHubURL(location.href);
-	const blobURL = `${location.protocol}//${location.hostname}${select('a[aria-label="View at this point in the history"]')!.getAttribute('href')}`;
-	const isFile = await fetch(blobURL, {method: 'HEAD'})
-		.then(response => new GitHubURL(response.url).route === 'blob' ? true : false);
 
-	if (disabledPagination.length === 0 || !url.filePath || !isFile ) {
+	const isFile = await fetch(
+		// eslint-disable-next-line restrict-template-expressions // https://github.com/Microsoft/TypeScript/issues/30239
+		`${location.protocol}//${location.hostname}${select('a[aria-label="View at this point in the history"]')!.getAttribute('href')}`,
+		{method: 'HEAD'},
+	)
+		.then(response => !!(new GitHubURL(response.url).route === 'blob'));
+
+	if (disabledPagination.length === 0 || !url.filePath || !isFile) {
 		return false;
 	}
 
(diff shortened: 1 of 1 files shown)


yakov116:
I don't like this extra call.

shinenelson:
> I don't like this extra call.

Neither do I. I am not happy with this solution. I was hoping that it would be straight-forward to simply lookup the blob URL from the DOM and check `GitHubURL(location.href).route`.

But apparently, GitHub uses the same method. But it makes sense for them since they are going through a network request cycle.

Is there a better method to detect this from the DOM?

fregante:
There's no point in using an HTTP request to determine whether to make another HTTP request. Either the information is available in the DOM or we just let it through. 

Directories are not blob URLs so I don't know what you're referring to

fregante:
I have a sense that if GitHub is pointing to a blob page even on directories they themselves don’t know what it is when generating the page.

shinenelson:
> I have a sense that if GitHub is pointing to a blob page even on directories they themselves don’t know what it is when generating the page.

They do not have to bother about it on commits page. And when the user does navigate to the URL, they will probably detect it on the network layer which is what I leveraged too.

> Either the information is available in the DOM or we just let it through.

I could not find a URL in the DOM that could reliably point to the right kind of page.

Unless there is some other method to detect this reliably without making a new HTTP request, this pull request and its [corresponding issue](https://github.com/sindresorhus/refined-github/issues/4683) can be closed.

fregante:
Thanks for looking into this!

PR #4710, opened 2021-08-24
Restore `linkify-code` for markdown code blocks

### LINKED ISSUE #4688: `linkify-code` is broken for code blocks in markdown
### Please ensure:

### Description

This used to be linkified:

```js
// https://github.com/fregante
```
```html
<a href="https://github.com/fregante">
```

### Screenshot

_No response_

### Console errors

_No response_

### Example URL

https://github.com/sindresorhus/refined-github/issues/4687

### Browser(s) used

safari

gamemaker1:
## Related issues

Closes #4688 if merged

## What this PR does

Adds the `.comment-body` selector back to linkify URLs in markdown code blocks.

This fixes it temporarily - this does not change the way markup is modified (as mentioned in the issue)


source/features/linkify-code.tsx
```diff
@@ -18,7 +18,8 @@ function initTitle(): void {
 function init(): void {
 	const selectors = [
 		'.js-blob-wrapper tr:not(.inline-comments)', // File blocks in pages like `isPRFiles`, `isSingleFile`
-		'.blob-wrapper tr', // Every other code blocks
+		'.blob-wrapper tr', // File blocks in blob URLs
+		'.comment-body', // Markdown code blocks
```
kidonng, line 22: You need to read my comment and related PR/issues carefully, this will cause issues with [GitHub Issue Link Status](https://github.com/fregante/github-issue-link-status) extension.

The markup for review comments ([example](https://togithub.com/sindresorhus/refined-github/pull/3225#discussion_r439535538)) is roughly `.comment-body > .blob-wrapper tr`, so if you directly include `.comment-body` here, `.blob-wrapper tr'` will be linkified twice.

gamemaker1: Thanks for the detailed reply - I will definitely look at the related issues and PR and fix it better

fregante: I can think of 2 possible "easy" solutions:

- Inside the linkifiers, avoid linkification if they're already linkified (not sure if easy/possible)
- Observe each line specifically (inner-blob, pre, etc)


I've avoided the latter until now because there might be hundreds of such elements and therefore hundreds of "addClass" calls. 😰

gamemaker1: > Inside the linkifiers, avoid linkification if they're already linkified (not sure if easy/possible)

Isn't that already done by:

https://github.com/sindresorhus/refined-github/blob/d36430023a7973239d71ca65e2dd9781c80b4c94/source/features/linkify-code.tsx#L22

fregante: Not directly. Here's what happens:

```js
for (const wrapper of select.all('.comment-body, .js-blob-wrapper')) {
  linkify(wrapper.children)
}
```

```html
<div class="comment-body">
	<div class="js-blob-wrapper">
		<span>http://example.com<span>
	</div>
</div>
```

1. Finds `.comment-body`
2. Linkifies `span`
3. Finds `.js-blob-wrapper`, which is inside `.comment-body`
4. `span` is linkified again

This happens in the same loop, _after_ the elements have been selected

fregante: Line 22 only applies to successive calls of the init function itself, like when a new comment appears on the page.

gamemaker1: > Line 22 only applies to successive calls of the init function itself, like when a new comment appears on the page.

Aah ok

> 4. span is linkified again

The link part will be converted to a `<a class="rgh-linkified-code" href="...">...</a>` right? So if the `linkify` function finds an `a` element with the class `rgh-linkified-code`, it should technically ignore it and not surround the existing `a` with another `a` again. Is this something we should add, will it help us?

gamemaker1: I did some experimenting with the selectors - and I found that `['.highlight', '.diff-table']` linkifies coded everywhere except a suggested change in the files tab of a PR.

fregante: Make sure you're testing with https://github.com/fregante/github-issue-link-status so that any errors become visible

gamemaker1: Ok, will try that, thanks!

gamemaker1: Tested it with the `issue-link-status` extension and didn't see any duplicates with `['.highlight', '.diff-table']` - however, it still doesn't work on a suggested change in the files tab of a PR.

fregante:
Closing this because it's not the right solution

PR #4725, opened 2021-08-26
Meta: Catch orphaned CSS files on CI

rmacklin:
This will make CI fail if we remove a feature but forget to remove its CSS file.

Note: There are currently two orphaned CSS files which are being removed in #4715, but https://github.com/sindresorhus/refined-github/pull/4715#discussion_r696328315 suggested extracting this to a separate PR to demonstrate the script failing when we expect it to.

[Example failure](https://github.com/sindresorhus/refined-github/runs/3437113085?check_suite_focus=true):

image

### DIFF
--- a/build/verify-features.ts
+++ b/build/verify-features.ts
@@ -22,6 +22,10 @@ function findError(filename: string): string | void {
 				return `ERR: \`${filename}\` should be imported by \`${correspondingTsxFile}\``;
 			}
 
+			if (entryPointSource.includes(`import './features/${filename}';`)) {
+				return `ERR: \`${filename}\` should only be imported by \`${correspondingTsxFile}\`, not by \`${entryPoint}\``;
+			}
+
 			return;
 		}
 
--- a/build/verify-features.ts
+++ b/build/verify-features.ts
@@ -1,17 +1,45 @@
 #!/usr/bin/env -S TS_NODE_COMPILER_OPTIONS='{"module":"es2020"}' node --loader ts-node/esm
 
-import {readdirSync} from 'node:fs';
+import {existsSync, readdirSync, readFileSync} from 'node:fs';
 
 import {getFeatures, getFeaturesMeta} from './readme-parser.js'; // Must import as `.js`
 
 const featuresDirContents = readdirSync('source/features/');
+const entryPoint = 'source/refined-github.ts';
+const entryPointSource = readFileSync(entryPoint);
 const importedFeatures = getFeatures();
 const featuresInReadme = getFeaturesMeta();
 
 const errors: string[] = [];
 
+function checkIfCssFileIsImported(fileName: string): string | null {
+	if (!entryPointSource.includes(`import './features/${fileName}';`)) {
+		const correspondingTsxFile = `source/features/${fileName.replace(/.css$/, '.tsx')}`;
+
+		if (existsSync(correspondingTsxFile)) {
+			if (!readFileSync(correspondingTsxFile).includes(`import './${fileName}';`)) {
+				return `ERR: The file \`${fileName}\` is not imported \`${correspondingTsxFile}\``;
+			}
+		} else {
+			return `ERR: The file \`${fileName}\` is not imported in \`${entryPoint}\``;
+		}
+	}
+
+	return null;
+}
+
 for (const fileName of featuresDirContents) {
-	if (fileName === 'index.tsx' || fileName.endsWith('.css')) {
+	if (fileName === 'index.tsx') {
+		continue;
+	}
+
+	if (fileName.endsWith('.css')) {
+		const error = checkIfCssFileIsImported(fileName);
+
+		if (error) {
+			errors.push(error);
+		}
+
 		continue;
 	}
 
@@ -22,7 +50,7 @@ for (const fileName of featuresDirContents) {
 
 	const featureId = fileName.replace('.tsx', '');
 	if (!importedFeatures.includes(featureId as FeatureID)) {
-		errors.push(`ERR: ${featureId} should be imported by \`/sources/refined-github.ts\``);
+		errors.push(`ERR: ${featureId} should be imported by \`${entryPoint}\``);
 	}
 
 	if (fileName.startsWith('rgh-')) {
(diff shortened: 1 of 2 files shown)


fregante:
I thought it could also catch wrong migrations in `options-storage.ts`. That is:

- for every line like `featureWasRenamed('FROM', 'TO')`, ensure that `TO` exists in `getFeatures` (or `getFeaturesMeta`, I don't remember)

But it's not necessary to merge this PR, it's just an extra thought.

yakov116:
> * for every line like `featureWasRenamed('FROM', 'TO')`, ensure that `TO` exists in `getFeatures` (or `getFeaturesMeta`, I don't remember)

That would fail if a feature was renamed 2 times and we want to keep both for some time.

fregante:
> > * for every line like `featureWasRenamed('FROM', 'TO')`, ensure that `TO` exists in `getFeatures` (or `getFeaturesMeta`, I don't remember)
> 
> That would fail if a feature was renamed 2 times and we want to keep both for some time.

No, the TO must exist:

```js
featureWasRenamed('initial', 'current')
featureWasRenamed('secondary', 'current')
```

rmacklin:
Sorry for the delay in getting back to this.

> I made some changes to the format and how the CSS features are handled, specifically it must check whether it's a TSX-based feature first and then check for the import in the main file.
> 
> Some types and lint may be wrong

I fixed the type and lint errors in 3b6f20e4a8234d4c7cfb79e2c6d85a551aa746a7 and fixed a logic bug in eca4f468be5e0f93791b8495185430fa7c81f4cf. Also updated an error message in c27c6899ad53ec16e13700d50f2bee8bdcc57254 and added another check in 2efc9ee501489699bcd05c8ac5e18bba3e366bca

The check is [currently failing](https://github.com/sindresorhus/refined-github/runs/3514973458?check_suite_focus=true) due to what seems to be a mistake in a PR that was merged a few hours ago - see https://github.com/sindresorhus/refined-github/pull/4745#discussion_r702330589

I can open a separate PR to fix that.

As for [checking the migrations in `options-storage.ts`](https://github.com/sindresorhus/refined-github/pull/4725#issuecomment-908556335), that sounds like a good idea but I agree we can handle that in a follow-up PR.

build/verify-features.ts
```diff
@@ -1,45 +1,56 @@
...
+const entryPoint = 'source/refined-github.ts';
+const entryPointSource = readFileSync(entryPoint);
 const importedFeatures = getFeatures();
 const featuresInReadme = getFeaturesMeta();
 
-const errors: string[] = [];
-
-for (const fileName of featuresDirContents) {
-	if (fileName === 'index.tsx' || fileName.endsWith('.css')) {
+function findError(fileName): string | void {
```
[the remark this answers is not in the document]
rmacklin: Sure, renamed in 9fdc8e143106fb12fde80b6b1d5d2be9ed568c19

fregante:
Thank you! Good catch! That means that this PR is working. Can you include that change in this PR instead so we can merge both at once?

rmacklin:
Done: 5dbf4dae1cab1cc264458969c2aa7159791e1bcb

build/verify-features.ts
```diff
@@ -1,45 +1,64 @@
...
+	}
+
+	if (filename.endsWith('.css')) {
+		const correspondingTsxFile = `source/features/${filename.replace(/.css$/, '.tsx')}`;
+		if (existsSync(correspondingTsxFile)) {
+			if (!readFileSync(correspondingTsxFile).includes(`import './${filename}';`)) {
+				return `ERR: \`${filename}\` should be imported by \`${correspondingTsxFile}\``;
+			}
+
+			if (entryPointSource.includes(`import './features/${filename}';`)) {
```
fregante, line 25: Can you extract this into a Boolean since it's repeated below?

For example on line 19 add:

```js
const isImportedByEntrypoint = etc
```

build/verify-features.ts
```diff
@@ -1,45 +1,64 @@
...
 const importedFeatures = getFeatures();
 const featuresInReadme = getFeaturesMeta();
 
-const errors: string[] = [];
+function findError(filename: string): string | void {
+	if (filename === 'index.tsx') {
+		return;
+	}
+
+	if (filename.endsWith('.css')) {
```
fregante, line 18: Can you extract this block of checks to its own function?

```suggestion
	if (filename.endsWith('.css')) {
      return findCssFileError(filename)
   }
```

fregante (approved):
I think this is done, there are just a couple of nitpicks

build/verify-features.ts
rmacklin: Addressed in 68bcd6ef6de3c342861027b56fc618d593342de4

build/verify-features.ts
rmacklin: Addressed in 403512a2681737ae137a13c15fdf398a37dff591

yakov116:
@rmacklin can
https://github.com/sindresorhus/refined-github/blob/403512a2681737ae137a13c15fdf398a37dff591/build/verify-features.ts#L51-L53

Be moved to 
https://github.com/sindresorhus/refined-github/blob/403512a2681737ae137a13c15fdf398a37dff591/build/verify-features.ts#L34-L36

Or de we need the other checks to happen? (Just asking)

rmacklin:
There would be a few consequences:
- files prefixed with `rgh-` would not be subject to the restriction of only allowing `.css` and `.tsx` files
- if we ever add any `.css` files with the `rgh-` prefix, we wouldn't be enforcing that they are imported correctly

build/verify-features.ts
```diff
@@ -1,45 +1,67 @@
...
+
+	const featureId = filename.replace('.tsx', '');
 	if (!importedFeatures.includes(featureId as FeatureID)) {
-		errors.push(`ERR: ${featureId} should be imported by \`/sources/refined-github.ts\``);
+		return `ERR: ${featureId} should be imported by \`${entryPoint}\``;
 	}
 
-	if (fileName.startsWith('rgh-')) {
-		continue;
+	if (filename.startsWith('rgh-')) {
```
fregante, line 51:
```suggestion
	// The previous checks apply to RGH features, but the next ones don't
	if (filename.startsWith('rgh-')) {
```

PR #4728, opened 2021-08-27
Restore `emphasize-draft-pr-label` feature

fregante:
- Reverts sindresorhus/refined-github#4450

On second thought… the contrast between the two is still too low.

Screen Shot 4 Screen Shot 5Screen Shot

### DIFF
--- a/source/features/emphasize-draft-pr-label.css
+++ b/source/features/emphasize-draft-pr-label.css
@@ -1,7 +1,11 @@
-.js-issue-row [aria-label='Open draft pull request'] svg {
-	stroke: var(--color-text-secondary, #586069);
-	stroke-width: 1.2px;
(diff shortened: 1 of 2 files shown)


fregante:
The reasons why it was implemented years ago still apply today IMHO. The new icons only slightly improved the situation by making the icon slightly different (but still overall the same shape with very little color contrast between the two)

I just disabled it on non-retina displays because it just sucks there.

kidonng:
Just dropping some styles to revert this:

```css
@media (-webkit-min-device-pixel-ratio: 2) {
    .js-issue-row [aria-label='Open draft pull request'] svg {
        stroke: none;
        color: var(--color-fg-muted) !important;
    }
}
```

PR #4699, opened 2021-08-22
Meta: Verify features meta on CI

gamemaker1:
## Related Issues

Closes #4698 if merged.

## What this PR does

It ensures:
- Only .tsx and .css files exist in `/source/features`.
- Each feature is mentioned in the readme.
- Each file is imported in `refined-github.tsx`.

PR #4729, opened 2021-08-27
Meta: Make `npm run test` run on Windows

darkred:
It's a change regarding the latter line in:
https://github.com/sindresorhus/refined-github/blob/b12bde4a44bb8395c3b2e293ab85451c83cd405b/package.json#L15-L16

*I noticed this while trying to troubleshoot this job failure: https://github.com/sindresorhus/refined-github/actions/runs/1172134227*

This change makes (`npm run test` and) `npm run test:features` work on Windows.
Previously they were resulting in:

```
'.' is not recognized as an internal or external command,
operable program or batch file.
ERROR: "test:features" exited with 1.
```
and

```
> test:features
> ./build/verify-features.ts

'.' is not recognized as an internal or external command,
operable program or batch file.
```
Using Node 16.8.0 , Typescript 4.4.2 on win 10.

---

After submitting this PR I see that this change unfortunately makes the "Features / Verify" job to fail.. See [here](https://github.com/sindresorhus/refined-github/runs/3447416738)
But it's the only way I found to make `npm run test` in PowerShell/Command Prompt/Cygwin.

(related PR: #4699, and for the escaped doublequotes: #4705 and #4386)
cc @gamemaker1


gamemaker1:
Good catch - this is because we use a shebang in `build/verify-features.ts` script. Shebangs won't be recognised by Windows.

To make the script run on Windows too, we'll have to:

1. Replace `./build/verify-features.ts` in the `test: features` script package.json with the command from the shebang, i.e., `node --loader ts-node/esm \"build/verify-features.ts\"`.
2. Figure out how to set the environment variable `TS_NODE_COMPILER_OPTIONS='{"module":"es2015"}` before running the script on both Windows as well as Linux/MacOS

I think the second one will be a built difficult, we might need to restore the `build/tsconfig.json` file.

fregante:
No restoring of the second config file, this can all be done inline

gamemaker1:
Maybe we can use `cross-env` (if it is fine to add another dev dependency) and replace the current `test:features` script with the following:

```
cross-env TS_NODE_COMPILER_OPTIONS='{\"module\":\"es2015\"}' node --loader ts-node/esm \"build/verify-features.ts\"
```

Tested it on my machine and it works

darkred:
> As for the current code, tsc does not run ts files.

Yes, it was irrelevant (documentation).

---

So, I followed your suggestions and moved the shebang out of the file and inside `package.json` making use of [cross-env](https://www.npmjs.com/package/cross-env) package.

Note: regarding the triple backslashes, see the above link and [this StackOverflow comment](https://stackoverflow.com/questions/25112510/how-to-set-environment-variables-from-within-package-json#comment85335014_37356976)

kidonng:
> this is because we use a shebang in the `build/verify-features.ts` script

I thought @yakov116 mostly uses Windows and should catch that? Shells scripts/shebang are in contrary to the cross-platform feature of Node. Generally they should be replaced by Node scripts and `node path/to/script` in `package.json`.

darkred:
@gamemaker1 Why the laugh emoji? Did I do something wrong?

gamemaker1:
> I thought @yakov116 mostly uses Windows and should catch that? Shells scripts/shebang are in contrary to the cross-platform feature of Node. Generally they should be replaced by Node scripts and `node path/to/script` in `package.json`.

Yes, I would have run it using `node path/to/script` but I had set an environment variable - a shebang was the first thing that came to mind

package.json
```diff
@@ -13,7 +13,7 @@
 		"pack:safari": "xcodebuild -project 'safari/Refined GitHub.xcodeproj'",
 		"start:safari": "open 'safari/build/Release/Refined GitHub.app'",
 		"test": "run-p ava lint:* build:* test:features",
-		"test:features": "./build/verify-features.ts",
+		"test:features": "cross-env TS_NODE_COMPILER_OPTIONS={\\\"module\\\":\\\"es2020\\\"} node --loader ts-node/esm \"build/verify-features.ts\"",
```
kidonng, line 16: Maybe just use single quotes:

```suggestion
		"test:features": "cross-env TS_NODE_COMPILER_OPTIONS={\\\"module\\\":\\\"es2020\\\"} node --loader ts-node/esm 'build/verify-features.ts'",
```

gamemaker1:
> @gamemaker1 Why the laugh emoji? Did I do something wrong? 

Oh no sorry I wanted to express my appreciation for  your PR that fixed something I should have thought through earlier when creating the script

package.json
gamemaker1: Yes I think so

darkred: Yes, unfortunately it breaks on Windows, see #4705 and #4386.

yakov116:
> > this is because we use a shebang in the `build/verify-features.ts` script
> 
> I thought @yakov116 mostly uses Windows and should catch that? Shells scripts/shebang are in contrary to the cross-platform feature of Node. Generally they should be replaced by Node scripts and `node path/to/script` in `package.json`.
> 

I was/am away from a personal computer for the last 8 weeks. So I haven't been having any issues 😀.

fregante:
It might need a little more work but it might be easier to just turn the file into CJS. Use a .CJS extension and then use `import()` instead of `import` if required, no extra dependencies needed.

darkred:
@fregante I have followed your suggestions. Would it be ok to push them now? I think it would be useful to do so,
because the `verify-features.cjs` I made it into, is rather different, and because the 3 errors that xojs raises in it cannot be disabled inline it seems, I mean via:  
```
/* eslint-disable @typescript-eslint/consistent-type-definitions, @typescript-eslint/explicit-function-return-type, import/extensions */ 
```

Maybe if you check it out, you might change your mind on the approach?

fregante:
Push and let's see

darkred:
Ok, I just did.

Initially, after only renaming `verify-features.ts` to `.cjs` I was getting these XO errors:


So, I made the changes in the last commit https://github.com/sindresorhus/refined-github/pull/4729/commits/094b3ea087012046554aae89b01a953e8b77155a and now I'm getting these XO errors:


which errors, as I mentioned above, I can't disable these errors inline via:

```
/* eslint-disable @typescript-eslint/consistent-type-definitions, @typescript-eslint/explicit-function-return-type, import/extensions */ 
```

fregante:
Ahhh what a pain is ESM+TS. We can't use .cjs because we're dealing with TypeScript, so the last commit must likely be reverted. I suggest not running this `test` locally but just run the lint part for now

fregante:
If 75b0a81 worked let's stick with that

darkred:
I think you meant to revert two more commits before, to the 2nd commit, https://github.com/sindresorhus/refined-github/pull/4729/commits/38dac12e4b4eea9bf281cb47efdbc6b4129eb1e1, i.e. before renaming verify-features.ts --> .cjs, and still maintaining `cross-env` ?

&nbsp;
Otherwise, if I revert to https://github.com/sindresorhus/refined-github/pull/4729/commits/75b0a81dd4807c81530b2da90291d588a1b2d89c then unfortunately I'm getting this error on lint:


which refers to:
https://github.com/sindresorhus/refined-github/blob/5bccdd68c84b95b040fb1e3a3c2f1e2f5a3f1e60/build/verify-features.ts#L11

fregante:
Ah yes I mentioned the wrong commit

darkred:
Ok, done.

fregante (approved):

fregante:
@darkred last one, can you fix the conflict?

Issue #220, opened 2016-05-17
Verified commits UI is a bit broken


hkdobrev:
Reproducible on https://github.com/mathiasbynens/dotfiles/commit/adc5b5971649017f4de5b28074e51fcf67ac5b0d

![screen shot 2016-05-17 at 12 52 16](https://cloud.githubusercontent.com/assets/506129/15318318/3c22afec-1c2e-11e6-8f98-aa574e79b12c.png)

The space is too narrow for the Verified badge.

I guess we could restyle that a bit or probably remove the `.patch` and `.diff` links if they keep creating problems in that area.

paulmolluzzo:
> I guess we could restyle that a bit or probably remove the .patch and .diff links if they keep creating problems in that area.

We might have discussed this before, but is there some icon we can use instead to save space? IIRC, we went through the process of shrinking padding to make more room there, so I don't know what else we can do except removing content or rearranging the whole area - like a two row layout.

I'm not keen on either scenario, but if anything gets removed it's the patch/diff links since they're added by RG.

Another idea: We can use keyboard shortcuts to switch to patch/diff of the commit and get rid of the links entirely. <kbd>+</kbd> for patch, and <kbd>=</kbd> for diff?

hkdobrev:
I'm open to either shortcuts or icons. I really don't want some of the more obscure things which RG adds to break the default UI. Thanks for the PR, will test it out a bit later today.

paulmolluzzo:
@hkdobrev if you can figure out some icon that works well, I'm for that as a solution too. @sindresorhus brought up the point that we might not remember the shortcuts, which is a valid concern. A link is better for something used infrequently.

sindresorhus:
I'm still not sold on #224.

I managed to fix the UI by decrementing the font-size by one pixel for each element, the following:

- `.full-commit`
- `.full-commit .sha-block`
- `.commit .sha-block, .commit .sha`

<img width="1012" alt="screen shot 2016-05-31 at 15 13 13" src="https://cloud.githubusercontent.com/assets/170270/15667781/39cbdbf0-2742-11e6-86fc-e681d84ce650.png">

paulmolluzzo:
I'm cool with @sindresorhus's solution if someone wants to make a PR.

PR #4735, opened 2021-08-28
Drop `patch-diff-links` font size modification

### LINKED ISSUE #4734: Font size of "Verified" badge and commit hash gets smaller even with `patch-diff-links` disabled
### Please ensure:

### Description

The font size became 10 after enabling the extension, from stylesheets it seems it's `patch-diff-links` feature causing it, but it didn't change if I disable the feature. I also tried on Chrome 92.0.4515.159 and the text didn't get smaller even with the feature enabled.

Also it'd be good if I can enable the feautre without making the text smaller.

### Screenshot

With extension disabled:

Snipaste_2021-08-28_21-29-10

With extension enabled and find features:

Snipaste_2021-08-28_21-29-57

### Console errors

_No response_

### Example URL

https://github.com/sindresorhus/refined-github/commit/7e96d94b06d8f522fdbc5baeb9277f7c2b3da2fc

### Browser(s) used

Firefox 91.0.2

kidonng:
Fix #4734

Instead of adding a `.rgh-patch-diff-links` class to make disabling the styles possible, I just dropped the font size modification.

The reason is that they are introduced in eed152441b60bbe8e0e1babc088931f2ba380212 (5 years ago) and seems to be making space for the verified badge (#220), but the problem no longer exists as of this PR:

https://user-images.githubusercontent.com/44045911/131222243-e392932b-580d-47c4-8014-6ecf65c10be5.mov

And we have to remember that decreasing the font size only _ease_ the "problem" but does not _fix_ it, if one adjust the browser to be very narrow, the multiline commit info will still appear, **even without `patch-diff-links`**.

Below is a screenshot of the second test URL from below (with a fake verified badge for demo purpose), on a maximummed Chrome browser, with 1440 x 810 screen size:

image

It takes almost all the space, but do remember that both a very long left column and 1440x810 screen size are uncommon enough to ignore.

## Test URLs

- Commit with a "Verified" badge https://github.com/sindresorhus/refined-github/commit/7e96d94b06d8f522fdbc5baeb9277f7c2b3da2fc
- Commit with a reasonably long enough left column: https://github.com/mathiasbynens/dotfiles/commit/8794573a49cc2e838ba31e8921aaf81f3ffdce5b

## Screenshot

See above


fregante:
GitHub used to be non-responsive so these changes were easy. Good times 😃

fregante (approved):

PR #4739, opened 2021-08-28
Link to exact file in PR in `list-prs-for-file` (Chrome only)

kidonng:
Via [Scroll To Text Fragment](https://github.com/WICG/scroll-to-text-fragment) we can direct the browser to the corresponding file in `list-prs-for-file`'s links, without fetching the page to get a hash.

This is not 100% accurate but should be fine most of the type, and it does nothing on a unsupported browser.

## Test URLs

https://github.com/NixOS/nixpkgs/blob/master/pkgs/top-level/all-packages.nix

## Screenshot

https://user-images.githubusercontent.com/44045911/131881041-a93830e7-bd3b-4b22-86a6-baa3cd8c41a8.mov

### DIFF
--- a/source/features/list-prs-for-file.tsx
+++ b/source/features/list-prs-for-file.tsx
@@ -10,8 +10,10 @@ import getDefaultBranch from '../github-helpers/get-default-branch';
 import addAfterBranchSelector from '../helpers/add-after-branch-selector';
 import {buildRepoURL, getRepo} from '../github-helpers';
 
+let path: string;
+
 function getPRUrl(prNumber: number): string {
-	return buildRepoURL('pull', prNumber, 'files');
+	return buildRepoURL('pull', prNumber, 'files') + `#:~:text=${path}`;
 }
 
 function getDropdown(prs: number[]): HTMLElement {
@@ -102,7 +104,7 @@ const getPrsByFile = cache.function(async (): Promise<Record<string, number[]>> 
 
 async function init(): Promise<void> {
 	// `clipboard-copy` on blob page, `#blob-edit-path` on edit page
-	const path = (await elementReady('clipboard-copy, #blob-edit-path'))!.getAttribute('value')!;
+	path = (await elementReady('clipboard-copy, #blob-edit-path'))!.getAttribute('value')!;
 	let {[path]: prs} = await getPrsByFile();
 
 	if (!prs) {


fregante:
Ohhhhhhh cool use even though no luck for Safari users. 

What happens when the file is loaded later in large PRs?

kidonng:
> What happens when the file is loaded later in large PRs?

The browser will only try to find the text **once** when the page finish loading (either normally or via AJAX).

fregante (approved):

source/features/list-prs-for-file.tsx
```diff
@@ -10,8 +10,10 @@ import getDefaultBranch from '../github-helpers/get-default-branch';
 import addAfterBranchSelector from '../helpers/add-after-branch-selector';
 import {buildRepoURL, getRepo} from '../github-helpers';
 
+let path: string;
```
fregante, line 13: Not a fan of the global but it's _ok_

fregante:
Can you add a demo video/gif to the first post?

Are you on Mac?

source/features/list-prs-for-file.tsx
kidonng: I tested it and there should not be any issue. Not a fan of having to write `select('clipboard-copy, #blob-edit-path')!.getAttribute('value')` twice either

kidonng:
> Can you add a demo video/gif to the first post?

Done

> Are you on Mac?

Yes but does it matter?

fregante:
> > Are you on Mac?
> 
> Yes but does it matter?

I was going to suggest the feature you used to create the video, but I see you know it already 😁

darkred:
Sorry, but does this work for you?
I am asking because I tried multiple times with the test [URL](https://github.com/NixOS/nixpkgs/blob/master/pkgs/top-level/all-packages.nix?rgh-link-date=2021-08-28T17%3A50%3A57Z) or the OP, as well as in an empty profile (Chrome 93 stable/96 canary) and unfortunately it never scrolls to the the specific highlighted section.
The scrolling only works if I manually enter the URL with the scroll-to-text-fragment.
Should I file a separate issue for this?

Here is a screen capture of RG 21.9.18 with in a fresh profile of Chrome 93 :

<details>
<summary>screen capture:</summary>




</details>

fregante:
This only works if the file is within the first few files that appear in the Files tab. If GitHub defers the loading of some files, they won’t be linkable

darkred:
Thanks for clarifying.

Issue #4742, opened 2021-08-29
modify `navigate-pages-with-arrow-keys` to match GitHub's `n` / `p` keys


shinenelson:
### Description

from https://github.com/sindresorhus/refined-github/pull/4691#issuecomment-907578663

> I think n/p has existed for a while actually. It might make more sense to change our feature to match n/p in order to extend it to more pages. Arrows feel wrong accessibility-wise.

While I agree with the change, I did a quick test to check whether the change was feasible. `p` is used as a hotkey for the **Project** filter on issues and pull request lists. I feel that the hotkey is most used on the issues and pull request lists.

Here are 2 options that I could think of ( both after modifying the feature to match GitHub's choice of `n` / `p` across all other supported pages ) : 
1. Maintain the arrow keys functionality for issues and pull request lists.
2. Drop the feature for issues and pull request lists

Either way, there is going to be a feature rename and migration that would be required.

### Example URLs

- Issue list : https://github.com/sindresorhus/refined-github/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc
- Pull request list : https://github.com/sindresorhus/refined-github/pulls?q=is%3Apr+sort%3Aupdated-desc+is%3Aclosed
- Search page : https://github.com/sindresorhus/refined-github/search?q=navigat
- Commits Page :
  - https://github.com/sindresorhus/refined-github/commits/main/source/features/navigate-pages-with-arrow-keys.tsx
  - https://github.com/sindresorhus/refined-github/commits/main
- Releases Page : https://github.com/sindresorhus/refined-github/releases
- Notifications Page : https://github.com/notifications
- Pull Request Commits : 
  - [`d9621aa` (#4699)](https://github.com/sindresorhus/refined-github/pull/4699/commits/d9621aaf54aae286b45b8805d34d9354224412ca)
  - [`dc04365` (#4699)](https://github.com/sindresorhus/refined-github/pull/4699/commits/dc0436589007560a77db22290427d7b7898657ee)

kidonng:
Arrow key fans here (😄), can't we just support both?

shinenelson:
> can't we just support both?

I thought about that too after I made https://github.com/sindresorhus/refined-github/pull/4691#issuecomment-907690478; but then I realized that it would not make sense ( and probably not work ) with 2 `data-hotkey` attributes for the same element.

shinenelson:
[...]

Issue #4695, opened 2021-08-21
⚠️ Many features don't work until you refresh the page


fregante:
### Please ensure:

- [X] The bug is caused by Refined GitHub. It doesn't happen if I disable the extension.

### Description

One of the recent AJAX PRs might have broken the loader altogether. One example I see is in the PR Files tab.

1. Open https://github.com/sindresorhus/refined-github/pull/4685
2. Click the files tab
3. Notice that almost every feature is missing:
- `remove-diff-signs`
- `pr-jump-to-first-non-viewed-file`
- `faster-reviews`
- `collapsible-button` (but `table-input` is working, these 2 should be identical)
- `batch-mark-files-as-viewed`
- etc etc etc
4. Refresh the page, everything works

cc @kidonng @dertieran @cheap-glitch @yakov116 as you've recently worked on the ajax part

### Screenshot

<img width="225" alt="Screen Shot 2" src="https://user-images.githubusercontent.com/1402241/130322058-16b2cd94-7339-4b0b-8397-daac71b5a394.png">

### Example URL

https://github.com/sindresorhus/refined-github/pull/4685 -> then click the files tab

### Browser(s) used

Safari

williammck:
Seeing this as well.

Trying to run `bisect` to find the issue doesn't work, https://github.com/sindresorhus/refined-github/blob/d8cadc830723728ab5c32f36cea31ea3f6aee814/source/helpers/bisect.tsx#L54 `document.body` is somehow null.

`highlight-deleted-and-added-files-in-diffs` is breaking due to https://github.com/sindresorhus/refined-github/blob/d8cadc830723728ab5c32f36cea31ea3f6aee814/source/features/highlight-deleted-and-added-files-in-diffs.tsx#L43 trying to call `cloneNode` on undefined.

Sometimes refreshing the page can get it to work. May try to look into this sometime this week as I'm really missing these features.

williammck:
Dug around for a bit, this is due to `deduplicate` not being set to `has-rgh-inner` on some features after https://github.com/sindresorhus/refined-github/pull/4571.

e.g. if I apply the following patch, `[faster-reviews](https://github.com/sindresorhus/refined-github/blob/main/source/features/faster-reviews.tsx)` now starts functioning properly even after multiple tab clicks.
```diff
diff --git a/source/features/faster-reviews.tsx b/source/features/faster-reviews.tsx
index 1b11bd54..7cc4e23c 100644
[...]

PR #4743, opened 2021-08-29
modify page navigation hotkeys

shinenelson:
fixes https://github.com/sindresorhus/refined-github/issues/4742

one thing that still needs fixing is :
> [it does not work] when `follow-file-renames` [~was~]is in effect because it is loaded after `navigate-pages-with-arrow-keys`.

I tried re-ordering the lines in `source/refined-github.ts` but it did not seem to have any effect. Maybe I am missing something.

## Test URLs

- Issue list : https://github.com/sindresorhus/refined-github/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc
- Pull request list : https://github.com/sindresorhus/refined-github/pulls?q=is%3Apr+sort%3Aupdated-desc+is%3Aclosed
- Search page : https://github.com/sindresorhus/refined-github/search?q=navigat
- Commits Page :
  - https://github.com/sindresorhus/refined-github/commits/main/source/features/navigate-pages-with-arrow-keys.tsx
  - https://github.com/sindresorhus/refined-github/commits/main
- Releases Page : https://github.com/sindresorhus/refined-github/releases
- Notifications Page : https://github.com/notifications
- Pull Request Commits : 
  - [`d9621aa` (#4699)](https://github.com/sindresorhus/refined-github/pull/4699/commits/d9621aaf54aae286b45b8805d34d9354224412ca)
  - [`dc04365` (#4699)](https://github.com/sindresorhus/refined-github/pull/4699/commits/dc0436589007560a77db22290427d7b7898657ee)

### DIFF
--- a/source/features/pagination-hotkey.tsx
+++ b/source/features/pagination-hotkey.tsx
@@ -1,5 +1,5 @@
 import select from 'select-dom';
-import {isGlobalConversationList,isPRCommit} from 'github-url-detection';
+import {isRepoConversationList,isPRCommit} from 'github-url-detection';
 
 import features from '.';
 
@@ -20,14 +20,26 @@ const previousPageButtonSelectors = [
 ];
 
 function init(): void {
-	const createNextPageButton = select(nextPageButtonSelectors);
-	if (createNextPageButton) {
-		createNextPageButton.dataset.hotkey = 'n';
-	}
+	if (isRepoConversationList()) {
+		const createPreviousPageButton = select('a.previous_page');
+		if (createPreviousPageButton) {
+			createPreviousPageButton.dataset.hotkey = 'ArrowLeft';
+		}
 
-	const createPreviousPageButton = select(previousPageButtonSelectors);
-	if (createPreviousPageButton) {
-		createPreviousPageButton.dataset.hotkey = 'p';
+		const createNextPageButton = select('a.next_page');
+		if (createNextPageButton) {
+			createNextPageButton.dataset.hotkey = 'ArrowRight';
+		}
+	} else {
+		const createNextPageButton = select(nextPageButtonSelectors);
+		if (createNextPageButton) {
+			createNextPageButton.dataset.hotkey = 'n';
+		}
+
+		const createPreviousPageButton = select(previousPageButtonSelectors);
+		if (createPreviousPageButton) {
+			createPreviousPageButton.dataset.hotkey = 'p';
+		}
 	}
 }
 
@@ -41,8 +53,6 @@ void features.add(__filebasename, {
 		() => select.exists('.paginate-container'),
 	],
 	exclude: [
-		// exclude on issue and pull request lists because the `p` hotkey conflicts with the Projects filter
-		isGlobalConversationList,
 		// exclude on pull request commit pages because GitHub already supports it natively
 		isPRCommit,
 	],
--- a/source/features/pagination-hotkey.tsx
+++ b/source/features/pagination-hotkey.tsx
@@ -1,5 +1,5 @@
 import select from 'select-dom';
-import {isPRCommit} from 'github-url-detection';
+import {isGlobalConversationList,isPRCommit} from 'github-url-detection';
 
 import features from '.';
 
@@ -37,6 +37,8 @@ void features.add(__filebasename, {
 		'p': 'Go to the previous page',
 	},
 	exclude: [
+		// exclude on issue and pull request lists because the `p` hotkey conflicts with the Projects filter
+		isGlobalConversationList,
 		// exclude on pull request commit pages because GitHub already supports it natively
 		isPRCommit,
 	],
--- a/source/features/pagination-hotkey.tsx
+++ b/source/features/pagination-hotkey.tsx
@@ -1,9 +1,10 @@
 import select from 'select-dom';
+import {isPRCommit} from 'github-url-detection';
 
 import features from '.';
 
 const nextPageButtonSelectors = [
-	'a.next_page', // Issue/PR list, Search
+	'a.next_page', // Search
 	'.paginate-container > .BtnGroup > :last-child', // Commits
 	'.paginate-container > .pagination > :last-child', // Releases
 	'.js-notifications-list-paginator-buttons > :last-child', // Notifications
@@ -11,7 +12,7 @@ const nextPageButtonSelectors = [
 ];
 
 const previousPageButtonSelectors = [
-	'a.previous_page', // Issue/PR list, Search
+	'a.previous_page', // Search
 	'.paginate-container > .BtnGroup > :first-child', // Commits
 	'.paginate-container > .pagination > :first-child', // Releases
 	'.js-notifications-list-paginator-buttons > :first-child', // Notifications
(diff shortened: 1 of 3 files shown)


source/features/pagination-hotkey.tsx
```diff
@@ -1,4 +1,5 @@
 import select from 'select-dom';
+import {isRepoConversationList, isPRCommit} from 'github-url-detection';
```
yakov116, line 2: Look at other features how this is done. ` import *` etc

source/features/pagination-hotkey.tsx
```diff
@@ -21,19 +22,32 @@ const previousPageButtonSelectors = [
...
+		? {
+			'←': 'Go to the previous page',
+			'→': 'Go to the next page',
+		}
+		: {
+			n: 'Go to the next page',
+			p: 'Go to the previous page',
+		},
+	include: [
+		// Activate only on pages with pagination
```
yakov116, line 45: Comment not needed

source/features/pagination-hotkey.tsx
```diff
@@ -21,19 +22,32 @@ const previousPageButtonSelectors = [
 function init(): void {
 	const createNextPageButton = select(nextPageButtonSelectors);
 	if (createNextPageButton) {
-		createNextPageButton.dataset.hotkey = 'ArrowRight';
+		createNextPageButton.dataset.hotkey = isRepoConversationList() ? 'ArrowRight' : 'n';
```
fregante, line 25: Pick one

source/refined-github.ts
```diff
@@ -62,7 +62,7 @@ import './features/download-folder-button';
 import './features/linkify-branch-references';
 import './features/open-all-conversations';
 import './features/hide-useless-comments';
-import './features/navigate-pages-with-arrow-keys';
+import './features/pagination-hotkey';
```
fregante, line 65: I like it

source/features/pagination-hotkey.tsx
```diff
@@ -21,19 +22,32 @@ const previousPageButtonSelectors = [
...
+			'←': 'Go to the previous page',
+			'→': 'Go to the next page',
+		}
+		: {
+			n: 'Go to the next page',
+			p: 'Go to the previous page',
+		},
+	include: [
+		// Activate only on pages with pagination
+		() => select.exists('.paginate-container'),
```
fregante, line 46: I think we should just pick some pages in `github-url-detection` and specify them here

shinenelson: I was thinking that the feature would automatically work on any pages that GitHub would add pagination in the future.

We do have good selector coverage ( which reminds me, there was another selector class choice for the notifications page that I wanted to suggest, but forgot ) for most cases, but then GitHub could beat us to it too though.

source/features/pagination-hotkey.tsx
shinenelson: I favor `n` / `p` that is closer to GitHub's preference.

But then, that would mean excluding the feature on `isRepoConversationList` - 3018387

shinenelson: or otherwise, like you said in https://github.com/sindresorhus/refined-github/issues/4742#issuecomment-907791088

> don't feel strongly to make the change yet, unless there are already other pages using n/p on GitHub.

we could drop this pull request for now and get back to it when we are ready to make the switch.

source/features/pagination-hotkey.tsx
```diff
@@ -21,19 +22,34 @@ const previousPageButtonSelectors = [
...
+			'→': 'Go to the next page',
+		}
+		: {
+			n: 'Go to the next page',
+			p: 'Go to the previous page',
+		},
+	include: [
+		pageDetect.isConversationList,
+		pageDetect.isLabelList,
+		pageDetect.isNotifications,
```
shinenelson, line 47: there is `.paginate-container.BtnGroup > :child` that would work for Notifications as well. It seems to align visually with the other selectors, except the `a.next_page` selector.

source/features/pagination-hotkey.tsx
```diff
@@ -21,19 +22,34 @@ const previousPageButtonSelectors = [
...
+		pageDetect.isRepoCommitList,
+	],
```
shinenelson, line 49: `isDiscussionsList` also has pagination, but the elements there are weird, there are multiple `.pagination` elements that is breaking the existing selectors ( though ideally, it should have worked ).

source/features/pagination-hotkey.tsx
fregante: Ahhh that's a problem. Indeed than maybe we can leave as was, except the feature name 😰

shinenelson: I do not feel strongly about simply renaming a feature. I think it can go along with the update to the feature _if it happens when it happens_:tm:.

... unless we are doing a mass-rename of many features together to follow https://github.com/sindresorhus/refined-github/pull/4745#discussion_r698044155

rkarendt (approved):

fregante:
Let's keep

- the rename
- the includes

The exclude isn't necessary because it already isn't covered

source/features/pagination-hotkey.tsx
fregante: Let's add a `// TODO:` comment with this information, shortened

fregante (approved):

source/features/pagination-hotkey.tsx
shinenelson: I could test and add the extra selectors; but then it would no longer be a **Meta** change though. 
Should I add the extra selectors to this pull request? Or should I create a separate pull request after this is merged in?

fregante: You can make all the changes here

shinenelson: Yes, that is exactly [how I feel](https://github.com/sindresorhus/refined-github/pull/4743#discussion_r698768658) too. Getting to work on it.

shinenelson:
Enabling this feature for `isDiscussionList` has turned out to be challenging : 
1. There is a nested `.pagination` class that makes a new overlapping class requirement.
   I worked around that by using `select.last`.
2. Newer pages are loaded via `pjax` but, apparently, does not properly activate after loading the new page.
   I am not sure if this a GitHub bug or a Refined GitHub bug.

I would like some help / feedback as to how to test / fix this.

( Or the other option would be to [leave it as a **TODO**](https://github.com/sindresorhus/refined-github/pull/4743/#discussion_r703416240) for someone who can fix it. I would like to fix it if I knew how to )

fregante:
> * I worked around that by using `select.last`.

Might be good enough

> 2\. Newer pages are loaded via `pjax`

We do have big issues with Ajax, not sure if it's due to this #4695

shinenelson:
> We do have big issues with Ajax, not sure if it's due to this #4695

The symptoms described in #4695 is what I was experiencing too. I am leaving it a **TODO** until #4695 is fixed.

fregante:
How's this PR doing?

PR #4756, opened 2021-09-03
Add missing space to `profile-gists-link`

### LINKED ISSUE #4751: Gist profile tab count lacks margin
There's little space on the left side of the Gist count

Screen Shot 5

The same issue might apply to other tabs we add, we should also check them (if we have any)

### Example URL

https://github.com/xt0rted

### Browser(s) used

Safari

kidonng:
Fix #4751 

## Test URLs

https://github.com/xt0rted

## Screenshot

Before: 

image

After:

image

### DIFF
--- a/source/features/profile-gists-link.tsx
+++ b/source/features/profile-gists-link.tsx
@@ -44,6 +44,5 @@ void features.add(__filebasename, {
 	include: [
 		pageDetect.isUserProfile,
 	],
-	awaitDomReady: false,
 	init,
 });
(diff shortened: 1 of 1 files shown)


fregante (approved):

source/features/profile-gists-link.tsx
```diff
@@ -44,6 +45,5 @@ void features.add(__filebasename, {
 	include: [
 		pageDetect.isUserProfile,
 	],
-	awaitDomReady: false,
```
yakov116, line 47: Why was this removed?

kidonng: https://github.com/sindresorhus/refined-github/issues/4751#issuecomment-912606783

yakov116: So then the elementReady should be removed

PR #4759, opened 2021-09-05
Prevent `shorten-links` from conflicting with OctoLinker

### LINKED ISSUE #4758: `shorten-links` conflicts with OctoLinker
### Please ensure:

### Description

It seems [OctoLinker](https://github.com/OctoLinker/OctoLinker) will linkies the code and `shorten-links` kicks in, potentially making the code invalid.

Worth noting that OctoLinker also linkies code blocks in (Markdown) documents (e.g. https://github.com/oakserver/oak#application-middleware-and-context), but Refined GitHub doesn't process those links.

### Screenshot

image

Note the `https://` prefix is stripped

### Console errors

_No response_

### Example URL

https://github.com/oakserver/oak/blob/main/deps.ts

### Browser(s) used

Chrome

kidonng:
Fix #4758

## Test URLs

https://github.com/oakserver/oak/blob/main/deps.ts

### DIFF
--- a/source/features/shorten-links.tsx
+++ b/source/features/shorten-links.tsx
@@ -7,7 +7,7 @@ import features from '.';
 import {linkifiedURLClass} from '../github-helpers/dom-formatters';
 
 function init(): void {
-	observe(`a[href]:not(.${linkifiedURLClass})`, {
+	observe(`a[href]:not(:is(.${linkifiedURLClass}, .octolinker-link))`, {
 		constructor: HTMLAnchorElement,
 		add(link) {
 			applyToLink(link, location.href);


fregante:
Rather than excluding specific links, it should just not be applied to links inside code blocks. Probably you just need to add `if closest(code), return` after adding the class, inside the observer.

source/features/linkify-code.tsx
```diff
@@ -34,9 +33,6 @@ function init(): void {
 			for (const element of select.all('span.pl-c', wrappers)) {
 				linkifyIssues(element);
 			}
-
-			// Mark code block as touched
-			wrappers.classList.add(linkifiedURLClass);
 		},
```
fregante, line 40: 🤔 I have not tested this but this whole file looks wrong? The selector is literally flipped and the class it's looking for is not added here anymore. Also it should not exclude comments

source/github-helpers/dom-formatters.ts
```diff
@@ -7,7 +7,7 @@ import getTextNodes from '../helpers/get-text-nodes';
 import parseBackticksCore from './parse-backticks';
 
 // Shared class necessary to avoid also shortening the links
-export const linkifiedURLClass = 'rgh-linkified-code';
+export const linkifiedURLClass = '.highlight, .js-suggested-changes-blob';
```
fregante, line 10: If this class isn't shared, it should not be in this file, also because it's not a class anymore but it's a selector

source/features/shorten-links.tsx
```diff
@@ -7,9 +7,13 @@ import features from '.';
 import {linkifiedURLClass} from '../github-helpers/dom-formatters';
 
 function init(): void {
-	observe(`a[href]:not(.${linkifiedURLClass})`, {
+	observe('a[href]', {
 		constructor: HTMLAnchorElement,
 		add(link) {
+			if (link.closest(linkifiedURLClass)) {
```
fregante, line 13:
```suggestion
			if (link.closest('code')) {
```

Right? We just don't want to shorten the links that appear in code. The only exception to this is review comments, so you also need to check whether the comment wrapper is "closer" than the code element.

For example the DOM:


```
code
  a -> don't shorten 
  review comment
    a -> shorten
    code suggestion in review comment
      a -> don't shorten

source/features/linkify-code.tsx
fregante: If you drop this line specifically it will just try to parse the whole file every time and add multiple links again like in a recent issue I think

kidonng:
Note to self: this should also include blocks without highlight:

image

fregante:
> Note to self: this should also include blocks without highlight:

Not sure if it's easily doable. Might be best to merge this soon and then extend it later

kidonng:
Will it work just like https://github.com/sindresorhus/refined-github/pull/4856#discussion_r721167511?

PR #4763, opened 2021-09-06
Improve reaction count styling

SNDST00M:
Some work to improve the rendering of reaction counts in the `reaction-avatars` feature.

## Test URLs

https://github.com/microsoft/vscode/issues/124901

## Screenshot

Before:

After:

### DIFF
--- a/source/features/reactions-avatars.css
+++ b/source/features/reactions-avatars.css
@@ -24,8 +24,6 @@
 	box-sizing: content-box;
 	display: inline-block;
 	height: var(--size);
-	margin-left: calc(var(--size) * 0.3); /* Move away from count */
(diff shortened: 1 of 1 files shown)


fregante (changes requested):
I don’t think it's necessary to add a number bubble inside a bubble. The only thing I'd address is the vertical alignment, but that needs a lot of careful work and testing to ensure it works well across browsers and fonts/operating systems

SNDST00M:
Right, that should address the count and emoji alignment in all browsers

fregante:
I appreciate the attempt, but I fear that I won't accept anything closer to this. If the intent is to center the content, the most likely solution is a diff like `+6 -3` rather than 20 additional lines with pixel values

PR #4766, opened 2021-09-07
Fix centering and spacing of `reaction-avatars` count

SNDST00M:
Some work to improve the rendering of reaction counts in the `reaction-avatars` feature.

## Test URLs

https://github.com/microsoft/vscode/issues/124901

## Screenshot

Before:

After:

### DIFF
--- a/source/features/reactions-avatars.css
+++ b/source/features/reactions-avatars.css
@@ -24,8 +24,6 @@
 	box-sizing: content-box;
 	display: inline-block;
 	height: var(--size);
-	margin-left: calc(var(--size) * 0.3); /* Move away from count */
-	margin-right: calc(var(--size) * -0.5);
 	margin: 0;
 	padding: 0;
 	width: var(--size);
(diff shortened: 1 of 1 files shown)


SNDST00M:
@fregante threw out a bunch of lines with only a 1px regression
I did copy code from flash.scss in #4763 but that was a mistake in retrospect

fregante:
When sending PRs it's best to avoid reordering CSS properties or else I have to check all those lines unnecessarily. Most lines don't need to be changed here

SNDST00M:
Sorry, in the habit of sorting alphabetically - almost automatic for me :)

fregante:
The current code is still wrong. Please don't open more PRs on this issue.

### The avatars are no longer overlapped

Screen Shot 10 <-- before
Screen Shot 9 <-- after

### The `-1px` margin just messes with the image

Screen Shot 11

SNDST00M:
While I understand the 4-review limit applies here, it turns out this was a simple matter of:

```css
:root .social-reaction-summary-item .social-button-emoji,
:root .social-reaction-summary-item span:nth-child(2) {
	align-self: center;
	box-sizing: content-box;
}

:root .rgh-reactions span:nth-child(2):not(:last-child) {
	min-width: 16px;
}
```

PR #4775, opened 2021-09-09
Enable `reactions-avatars` on more pages

### LINKED ISSUE #3978: Enable `reactions-avatars` on discussions  and releases

kidonng:
Fix #3978

## Test URLs

- Discussion: https://github.com/tophf/mpiv/discussions/50
- Release: https://github.com/sindresorhus/refined-github/releases/tag/21.9.3

## Screenshot

image

image

### DIFF
--- a/source/features/reactions-avatars.tsx
+++ b/source/features/reactions-avatars.tsx
@@ -84,20 +84,35 @@ const viewportObserver = new IntersectionObserver(changes => {
 	rootMargin: '500px',
 });
 
+const selector = '.has-reactions .comment-reactions-options:not(.rgh-reactions)';
+
+function observeReactions(commentReactions: Element): void {
+	commentReactions.classList.add('rgh-reactions');
+	viewportObserver.observe(commentReactions);
+}
+
 function init(): void {
-	observe('.has-reactions .comment-reactions-options:not(.rgh-reactions)', {
-		add(commentReactions) {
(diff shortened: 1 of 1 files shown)


source/features/reactions-avatars.tsx
```diff
@@ -1,5 +1,7 @@
 import './reactions-avatars.css';
 import React from 'dom-chef';
+import onetime from 'onetime';
+import {observe} from 'selector-observer';
 import select from 'select-dom';
 import {flatZip} from 'flat-zip';
 import * as pageDetect from 'github-url-detection';
```
yakov116, line 7: Sort

fregante:
The problem is that hasComments automatically uses its own listener, you can see it in features/index.tsx

It's probably best to create 2/3 inits. One for hasComments, one with the observer for the discussions, one static for the releases

kidonng:
> The problem is that hasComments automatically uses its own listener, you can see it in features/index.tsx

I forgot that

> It's probably best to create 2/3 inits. One for hasComments, one with the observer for the discussions, one static for the releases

That's what I did initially, I just thought this is simpler

source/features/reactions-avatars.tsx
kidonng: Is there any guidelines for this other than looking at other files? I know some of them but it's not clear.

PR #4777, opened 2021-09-10
Meta: Ensure the body loaded before bisecting

### LINKED ISSUE #4748: Feature bisect doesn't work consistently
https://github.com/sindresorhus/refined-github/blob/d8cadc830723728ab5c32f36cea31ea3f6aee814/source/helpers/bisect.tsx#L54

`document.body` might not be available at this point. It needs an `element-ready` call

## Original report

> Trying to run `bisect` to find the issue doesn't work, https://github.com/sindresorhus/refined-github/blob/d8cadc830723728ab5c32f36cea31ea3f6aee814/source/helpers/bisect.tsx#L54 `document.body` is somehow null.
> 
> _Originally posted by @williammck in https://github.com/sindresorhus/refined-github/issues/4695#issuecomment-908935632_

@williammck if you're seeing the `highlight-deleted-and-added-files-in-diffs` error after refreshing please open a new issue specific to that

yakov116:
Fixes #4748

## Test URLs
None

## Screenshot
None

@fregante if this is not the correct solution you can just close the PR.

### DIFF
--- a/source/helpers/bisect.tsx
+++ b/source/helpers/bisect.tsx
@@ -2,6 +2,7 @@ import React from 'dom-chef';
 import cache from 'webext-storage-cache';
 import select from 'select-dom';
 
+import waitFor from './wait-for';
 import features from '../features';
 import pluralize from './pluralize';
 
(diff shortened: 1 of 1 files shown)


fregante:
Looks right except we can just use elementReady

yakov116:
> Looks right except we can just use elementReady

And I thought I was being cool by using waitfor 

I think it's faster this way. Will see if I have time next week to change it.

fregante:
`waitFor` is only there because:

- https://github.com/sindresorhus/refined-github/issues/4312

fregante (approved):

PR #4787, opened 2021-09-16
Add support for Safari on iOS

### LINKED ISSUE #4580: Refined GitHub for iOS and iPadOS
Are there additional steps that need to be taken care of to publish a version suitable for iPadOS? Visiting the extension‘s AppStore page shows me a message that the app is not suitable for my device.

sindresorhus:
Fixes #4580

Some things that should be improved, but that are outside the scope of this PR and should be done later on:
- The options UI needs some tweaks.
- Some features need to be tweaked to the new screen sizes.

---

iOS Safari extensions compatibility table: https://developer.apple.com/documentation/safariservices/safari_web_extensions/assessing_your_safari_web_extension_s_browser_compatibility

---

Screen Shot 2021-09-16 at 21 04 40
Screen Shot 2021-09-16 at 21 04 50


distribution/manifest.json
```diff
@@ -20,7 +20,8 @@
 		"https://api.github.com/*"
 	],
 	"browser_action": {
-		"default_icon": "icon.png"
+		"default_icon": "icon.png",
+		"default_popup": "options.html"
```
sindresorhus, line 24: This is needed so you can click the extension in Safari to open options.

distribution/manifest.json
```diff
@@ -33,6 +34,7 @@
 		"page": "options.html"
 	},
 	"background": {
+		"persistent": false,
```
sindresorhus, line 37: This is required for iOS Safari.

sindresorhus:
No idea why https://github.com/sindresorhus/refined-github/runs/3622399500?check_suite_focus=true#step:5:16 is failing...

fregante:
> No idea why [sindresorhus/refined-github/runs/3622399500?check_suite_focus=true#step:5:16](https://github.com/sindresorhus/refined-github/runs/3622399500?check_suite_focus=true#step:5:16) is failing...

Because the rule has been enabled globally and we don't have other JS files, so it never came up

https://github.com/sindresorhus/refined-github/blob/32abfaaa4696053fa956f12b0186339e80f19292/.xo-config.json#L22

distribution/manifest.json
fregante: 😬 didn't we remove it recently for something else? Or maybe that's been addressed? I think we discussed this within the last 6 months.

distribution/manifest.json
fregante: But that changes the behavior 😳 

We don't really want the options to appear in yet another view without `options_ui` styles. Is there any other way? Is it time to generate a Safari-specific manifest?

Also the button currently is linked to GitHub itself, to make it slightly less useless (as people don't generally need to change the options to often)

busches: The only time I click the button is to go to options, and it's a bookmark to GitHub which I don't need. 😭

sindresorhus: I couldn't find any other way to show the RG settings in iOS Safari.

Since Safari is the only extension with manual submission, I think it makes more sense to leave it in and remove the property in a script during the automatic Chrome/Firefox submission.

distribution/manifest.json
sindresorhus: I know we talked about it, but I don't remember the outcome. We probably just forgot.

fregante: 👍 

I said this:

> But actually Firefox doesn’t support persistent:false at all so we’re good at this point

fregante (approved):
So much new code… Thanks Apple

PR #4792, opened 2021-09-17
Add support for `rst` files to `view-markdown-source`

ITProKyle:
Related to #4244, this adds support for another common markup format that is rendered by GitHub, reStructuredText.

Should this feature be renamed to `view-markup-source` since it would now support more formats than just markdown? I'll let someone else decide.

## Test URLs

- https://github.com/github/markup/blob/46908504f5/test/markups/README.rst

## Screenshot

image

### DIFF
--- a/source/features/view-markdown-source.tsx
+++ b/source/features/view-markdown-source.tsx
@@ -33,7 +33,7 @@ async function init(): Promise<void> {
 
 void features.add(__filebasename, {
 	include: [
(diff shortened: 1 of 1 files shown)


fregante:
I think it can be renamed to `view-source`, since it could apply to any rendered file, but let's wait for a more comprehensive search of compatible files before we make changes.

PR #4794, opened 2021-09-18
Fix options page on iOS Safari

sindresorhus:


source/options.css
```diff
@@ -4,6 +4,13 @@
 	}
 }
 
+/* For iOS Safari */
+@media only screen and (max-width: 1000px) {
```
fregante, line 8: `width` addresses the viewport, which is ~400px in Chrome’s options (it's an iframe). This seems to be the right one https://css-tricks.com/snippets/css/media-queries-for-standard-devices/

```suggestion
@media only screen and (max-device-width: 1000px) {
```

PR #4749, opened 2021-09-01
Fix: `edit-comments-faster` appears on other users' comments in issues

gamemaker1:
## Related issues

Closes #4712 if merged.

## What this PR does

Does not add the edit button on comments from other users if the current user is not allowed to edit them.

## Test URLs

https://github.com/facebook/jest/issues/6316 (locked issue)
https://github.com/facebook/jest/issues/11791 (regular issue)

PR #4796, opened 2021-09-19
Fix: `edit-comments-faster` appears on every comment

### LINKED ISSUE #4712: `edit-comments-faster` appears on other users' comments in issues
### Please ensure:

### Description

Open an issue, e.g.: https://github.com/facebook/jest/issues/6316 and scroll through the comments in the conversations tab. You will see edit buttons on all comments, including comments from other users. When you click on the edit button, you see an error message: `Sorry, something went wrong`.

### Console errors

A 403 Forbidden when the page attempts to retrieve the edit form for that comment:

```
XHR GET https://github.com/facebook/jest/issue_comments/.../edit_form?textarea_id=issuecomment-...-body&comment_context= [HTTP/2 403 Forbidden 321 ms]
```

### Example URL

https://github.com/facebook/jest/issues/6316 (locked issue)
https://github.com/facebook/jest/issues/11791 (regular issue)

### Browser(s) used

Firefox

gamemaker1:
## Related issues

Closes #4712 if merged.

This is an up-to-date version of #4749.

## What this PR does

Does not add the edit button on comments from other users if the current user is not allowed to edit them.

## Test URLs

https://github.com/facebook/jest/issues/6316 (locked issue)
https://github.com/facebook/jest/issues/11791 (regular issue)

### DIFF
--- a/source/features/quick-comment-edit.tsx
+++ b/source/features/quick-comment-edit.tsx
@@ -13,7 +13,7 @@ function canEditEveryComment(): boolean {
 		'[aria-label^="You are the owner"]',
 		'[title^="You are a maintainer"]',
 		'[title^="You are a collaborator"]',
-	]) && select.exists('div.js-previewable-comment-form');
+	]) || select.exists('#settings-tab');
 }
 
 function init(): void {
--- a/source/features/quick-comment-edit.tsx
+++ b/source/features/quick-comment-edit.tsx
@@ -12,7 +12,7 @@ function canEditEveryComment(): boolean {
 		'[aria-label^="You have been invited to collaborate"]',
 		'[aria-label^="You are the owner"]',
 		'[title^="You are a maintainer"]',
-		'[title^="You are a collaborator"]'
+		'[title^="You are a collaborator"]',
 	]) && select.exists('div.js-previewable-comment-form');
 }
 
--- a/source/features/quick-comment-edit.tsx
+++ b/source/features/quick-comment-edit.tsx
@@ -8,7 +8,12 @@ import * as pageDetect from 'github-url-detection';
 import features from '.';
 
 function canEditEveryComment(): boolean {
-	return select.exists('[aria-label^="You have been invited to collaborate"], [aria-label^="You are the owner"], [title^="You are a maintainer"], [title^="You are a collaborator"]') && select.exists('div.js-previewable-comment-form');
+	return select.exists([
+		'[aria-label^="You have been invited to collaborate"]',
+		'[aria-label^="You are the owner"]',
+		'[title^="You are a maintainer"]',
+		'[title^="You are a collaborator"]']
+	) && select.exists('div.js-previewable-comment-form');
 }
 
 function init(): void {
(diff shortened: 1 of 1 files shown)


source/features/quick-comment-edit.tsx
```diff
@@ -7,13 +7,13 @@ import * as pageDetect from 'github-url-detection';
 
 import features from '.';
 
-function canEditDiscussion(): boolean {
-	return pageDetect.isDiscussion() && select.exists('[title^="You are a maintainer"], [title^="You are a collaborator"]');
+function canEditEveryComment(): boolean {
+	return select.exists('[aria-label^="You have been invited to collaborate"], [aria-label^="You are the owner"], [title^="You are a maintainer"], [title^="You are a collaborator"]') && select.exists('div.js-previewable-comment-form');
```
fregante, line 11: Multiline, also no need for `^=` if we're matching the whole string, `=` is enough

```js
select.exists([
	'a',
	'b',
	'etc,
]) && select.exists('')
```

gamemaker1: I think the `You have been invited to collaborate` string has a suffix with the repo name, i.e., `You have been invited to collaborate on the ... repository`.

source/features/quick-comment-edit.tsx
```diff
@@ -7,13 +7,18 @@ import * as pageDetect from 'github-url-detection';
...
 
-function canEditDiscussion(): boolean {
-	return pageDetect.isDiscussion() && select.exists('[title^="You are a maintainer"], [title^="You are a collaborator"]');
+function canEditEveryComment(): boolean {
+	return select.exists([
+		'[aria-label^="You have been invited to collaborate"]',
+		'[aria-label^="You are the owner"]',
+		'[title^="You are a maintainer"]',
+		'[title^="You are a collaborator"]',
+	]) && select.exists('div.js-previewable-comment-form');
```
fregante, line 16: Is this actually related to `canEditEveryComment`? This looks like a generic test to check whether there are any editable comments at all

gamemaker1: The `select.exists('div.js-previewable-comment-form')` ensures they can edit the first comment, and the rest checks if they are one of `owner`, `collaborator`, or `invited-to-collaborate`.

I'm not sure why `select.exists('div.js-previewable-comment-form')` exists, since if you are `owner`, `collaborator`, or `invited-to-collaborate`, you should be able to edit all, but it works well with this condition now.

gamemaker1: Hmmm you're right, the `select.exists('div.js-previewable-comment-form')` is redundant. I added it as a fix to https://github.com/sindresorhus/refined-github/pull/4749#issuecomment-915767672, but now I realise it doesn't even fix the issue. A better fix would be detecting the settings tab.

gamemaker1: I changed it now, let me know what you think

fregante (approved):
I think this is fine now, it fixes the original issue. However there's a new bug (which should be opened and handled separately) that does the opposite:

- `edit-comments-faster` only appears if you've already left a comment

gamemaker1:
> - `edit-comments-faster` only appears if you've already left a comment

I think that should be fixed by `'#settings-tab'` right?

fregante:
Only for the owner, but not for collaborators

Pingu85:
Hei

Pingu85:
`FreH2506

Issue #1856, opened 2019-03-13
Advanced Search page detected as 404


LeoColomb:
Not a big issue, but the [useful-not-found-page](https://github.com/sindresorhus/refined-github/blob/0f3a4ae5cfc2315b30fac76bc1f405f5c13d798d/source/features/useful-not-found-page.tsx) feature is active on [advanced search page](https://github.com/search/advanced).
Seems related to #1849 & #1837.

---
![image](https://user-images.githubusercontent.com/846943/54307374-d3774f00-45cb-11e9-83b7-aa758ac8c086.png)

PR #4797, opened 2021-09-19
`useful-not-found-page`: Show alternate links only if applicable

### LINKED ISSUE #4790: `useful-not-found-page` shows 'This object exists on the default branch' when viewing a deleted issue
### Please ensure:

### Description

Problem breakdown:

- The `getUrlToFileOnDefaultBranch` ([called by `showDefaultBranchLink`](https://github.com/sindresorhus/refined-github/blob/b504a42cd783836b979688731a43a0a07be65bf8/source/features/useful-not-found-page.tsx#L101)) function checks if the URL is a non-404 when the 'branch' part is replaced with the default branch:

https://github.com/sindresorhus/refined-github/blob/b504a42cd783836b979688731a43a0a07be65bf8/source/features/useful-not-found-page.tsx#L60-L64

- In this case, the 'branch' part is in fact the issue number. So it changes https://github.com/octokit/rest.js/issues/558 => https://github.com/octokit/rest.js/issues/master.
- When we visit an issue page with the issue number set to a string, (e.g.: `https://github.com/octokit/rest.js/issues/master`), Github searches for issues with the string as the `author`.
- This means that the `getUrlToFileOnDefaultBranch` function does not receive a 404, and it returns https://github.com/octokit/rest.js/issues/master as a valid issue link.
- The same does not occur on discussions as it does not support this type of query.

To prevent this bug from occurring, I think we should fix the `GithubURL` class, and make sure it does not set the `branch` variable when it gets an issue link as follows:

```diff
diff --git a/source/github-helpers/github-url.ts b/source/github-helpers/github-url.ts
index e8acd411..ec4e8c53 100644
--- a/source/github-helpers/github-url.ts
+++ b/source/github-helpers/github-url.ts
@@ -77,8 +77,12 @@ export default class GitHubURL {
 
   set pathname(pathname: string) {
     const [user, repository, route, ...ambiguousReference] = pathname.replace(/^\/|\/$/g, '').split('/');
-    const {branch, filePath} = this.disambiguateReference(ambiguousReference);
-    this.assign({user, repository, route, branch, filePath});
+    if (route === 'blob' || route === 'tree') {
+      let { branch, filePath } = this.disambiguateReference(ambiguousReference);
+      this.assign({ user, repository, route, branch, filePath });
+    } else {
+      this.assign({ user, repository, route, branch: undefined, filePath: undefined });
+    }
   }
 
   get href(): string {
```

This will trigger the condition on https://github.com/sindresorhus/refined-github/blob/b504a42cd783836b979688731a43a0a07be65bf8/source/features/useful-not-found-page.tsx#L62 and the function will return, thus not showing the incorrect link.

Let me know if this sounds like an acceptable solution. If so, I'll be happy to send in a PR.

Thanks!

### Screenshot

image

Clicking on [`This object`](https://github.com/octokit/rest.js/issues/master) takes you to https://github.com/octokit/rest.js/issues/master.

### Console errors

```shell
None
```

### Example URL

https://github.com/octokit/rest.js/issues/558

### Browser(s) used

Firefox

gamemaker1:
## Related issues

Fixes #4790 if merged.

## What this PR does

Restores a check from https://github.com/sindresorhus/refined-github/commit/6047ca569d61638025a26f29a31ab9069b0dd108 that does not run the `showDefaultBranchLink` and `showAlternateLink` functions if the URL is not a `blob`/`tree`/`edit` one.

## Test URLs

From the original issue:

https://github.com/octokit/rest.js/issues/558

### DIFF
--- a/source/features/useful-not-found-page.tsx
+++ b/source/features/useful-not-found-page.tsx
@@ -145,7 +145,7 @@ async function showAlternateLink(): Promise<void> {
 
 function init(): false | void {
 	const parts = parseCurrentURL();
-	if (parts.length <= 1 || !select.exists('[alt*="This is not the web page you are looking for"]')) {
+	if (parts.length < 2) {
 		return false;
 	}
 
--- a/source/features/useful-not-found-page.tsx
+++ b/source/features/useful-not-found-page.tsx
@@ -98,6 +98,11 @@ async function showMissingPart(): Promise<void> {
 }
 
 async function showDefaultBranchLink(): Promise<void> {
+	const parts = parseCurrentURL();
+	if (!['tree', 'blob', 'edit'].includes(parts[2])) {
+		return
+	}
+
 	const urlToFileOnDefaultBranch = await getUrlToFileOnDefaultBranch();
 	if (!urlToFileOnDefaultBranch) {
 		return;
@@ -111,6 +116,11 @@ async function showDefaultBranchLink(): Promise<void> {
 }
 
 async function showAlternateLink(): Promise<void> {
+	const parts = parseCurrentURL();
+	if (!['tree', 'blob', 'edit'].includes(parts[2])) {
+		return
+	}
+
 	const url = new GitHubURL(location.href);
 	if (!url.branch || !url.filePath) {
 		return;
--- a/source/features/useful-not-found-page.tsx
+++ b/source/features/useful-not-found-page.tsx
@@ -100,7 +100,7 @@ async function showMissingPart(): Promise<void> {
 async function showDefaultBranchLink(): Promise<void> {
 	const parts = parseCurrentURL();
 	if (!['tree', 'blob', 'edit'].includes(parts[2])) {
-		return
+		return;
 	}
 
 	const urlToFileOnDefaultBranch = await getUrlToFileOnDefaultBranch();
@@ -118,7 +118,7 @@ async function showDefaultBranchLink(): Promise<void> {
 async function showAlternateLink(): Promise<void> {
 	const parts = parseCurrentURL();
 	if (!['tree', 'blob', 'edit'].includes(parts[2])) {
-		return
+		return;
 	}
 
 	const url = new GitHubURL(location.href);
--- a/source/features/useful-not-found-page.tsx
+++ b/source/features/useful-not-found-page.tsx
@@ -98,11 +98,6 @@ async function showMissingPart(): Promise<void> {
 }
 
 async function showDefaultBranchLink(): Promise<void> {
-	const parts = parseCurrentURL();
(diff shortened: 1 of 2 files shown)


source/features/useful-not-found-page.tsx
```diff
@@ -145,13 +145,16 @@ async function showAlternateLink(): Promise<void> {
...
-	if (parts.length <= 1 || !select.exists('[alt*="This is not the web page you are looking for"]')) {
+	if (parts.length < 2) {
 		return false;
 	}
 
 	void showMissingPart();
-	void showDefaultBranchLink();
-	void showAlternateLink();
+
+	if (['tree', 'blob', 'edit'].includes(parts[2])) {
```
fregante, line 154: If we had [a `required` field](https://github.com/sindresorhus/refined-github/issues/4008#issuecomment-812843149), I'd have rather used the regular `isSingleFile`/etc as suggested my comment

```js
	{
		include: [is404],
		require: [() => parseCurrentURL().length > 1],
		init: showMissingPart
	},
	{
		include: [is404],
		require: [
			isSingleFile,
			isTree,
			isEditingFile,
		],
		init: () => {showDefaultBranchLink(); showAlternateLink() }
	},
```

source/features/useful-not-found-page.tsx
```diff
@@ -145,13 +145,16 @@ async function showAlternateLink(): Promise<void> {
 
 function init(): false | void {
 	const parts = parseCurrentURL();
-	if (parts.length <= 1 || !select.exists('[alt*="This is not the web page you are looking for"]')) {
```
yakov116, line 148: If I remember correctly we needed this to detect if it was a 404

yakov116: Yes
https://github.com/sindresorhus/refined-github/commit/bfdad4b4ea240c93ab0c51ff4e2a71a5b3849c94
https://github.com/sindresorhus/refined-github/issues/1856

gamemaker1: I tried it out with the new code, and it does not trigger the `useful-not-found-page` on the advanced search page. I think it's because the page does not count as a 404 page anymore (`pageDetect.is404`).

fregante: > If I remember correctly we needed this to detect if it was a 404

Yeah, and that seems wrong. We already have a `is404` that should that. If it doesn't work, _that_ should be fixed. But by the looks of it, `is404` already does not include the linked page.

fregante: https://github.com/sindresorhus/refined-github/issues/4008#issuecomment-894979392

PR #4798, opened 2021-09-19
Fix: "Done" click is ignored when `select-notifications` is open

### LINKED ISSUE #4753: `select-notifications` should allow clicking on Done and others even if the dropdown is open
This is because the dropdown is a modal, which is fine, but this makes clicking on the actions a bit annoying since that's exactly what will come after selecting some notifications.

Notice how the first click on **Done** is "ignored"

https://user-images.githubusercontent.com/1402241/131730861-cf07a794-7169-4caa-b239-8885ede0f192.mov

cc @cheap-glitch 

## URL

## Browser

Safari

cheap-glitch:
Fixes #4753 

## Test URLs

## Screenshot

https://user-images.githubusercontent.com/46634000/133922795-289fb9b2-1ea9-4b03-b3b7-b3cac7152e36.mp4

### DIFF
--- a/source/features/select-notifications.tsx
+++ b/source/features/select-notifications.tsx
@@ -153,7 +153,7 @@ function createDropdown(): JSX.Element {
 }
 
 function closeDropdown(): void {
-	select('.rgh-select-notifications[open] summary')?.dispatchEvent(new MouseEvent('click'));
+	select('.rgh-select-notifications')?.removeAttribute('open');
 }
 
 function init(): false | void {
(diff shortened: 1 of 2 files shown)


source/features/select-notifications.tsx
```diff
@@ -150,10 +152,17 @@ function createDropdown(): JSX.Element {
 	);
 }
 
+function closeDropdown(): void {
+	select('.rgh-select-notifications[open] summary')?.dispatchEvent(new MouseEvent('click'));
```
fregante, line 156: More compact, no events, simple selector

```suggestion
	select('.rgh-select-notifications')?.removeAttribute('open');
```

fregante (approved):

PR #4799, opened 2021-09-19
Fix: `select-notifications` disappears after clicking "Done"

### LINKED ISSUE #4752: `select-notifications` disappears after clicking Done
https://user-images.githubusercontent.com/1402241/131730615-64308775-edb1-4f75-836c-93c91faf08d4.mov

cc @cheap-glitch as you might know why

## URL

## Browser

Safari

cheap-glitch:
Fixes #4752 

## Test URLs
https://github.com/notifications?query=repo%3Asindresorhus%2Frefined-github

## Screenshot

https://user-images.githubusercontent.com/46634000/133923470-46a61400-7abe-4a28-bd68-915dbb8c063d.mp4

GitHub fires some events when "Done" is clicked (notably two `socket:message` events from `.js-updatable` elements) but they're not correlated to the removal of the dropdown: sometimes it happens before, sometimes after.\
So in the end I went with a simple `selector-observer`.

### DIFF
--- a/source/features/select-notifications.tsx
+++ b/source/features/select-notifications.tsx
@@ -1,5 +1,6 @@
 import React from 'dom-chef';
 import select from 'select-dom';
+import {observe} from 'selector-observer';
 import * as pageDetect from 'github-url-detection';
 import {
 	CheckCircleIcon,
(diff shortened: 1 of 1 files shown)


fregante:
`selector-observer` should be the only thing called in `init`: Use to add the first dropdown and it will automatically add it when the element you're tracking is removed. That's how we use it everywhere

fregante (approved):
Untested but it looks about right!

fregante:
It seems to work!

PR #4800, opened 2021-09-19
Meta: Add an `asLongAs` option to `features.add()`

gamemaker1:
## What this PR does

Adds the `required` variable from https://github.com/sindresorhus/refined-github/issues/4008#issuecomment-812843149 as the `always` variable. Also move negative `exclude`s to their own list of `always` tests.

> https://github.com/sindresorhus/refined-github/blob/936eedbb0bb37a30fc44a2cc7e5f3af31f24a3fc/source/features/fork-source-link-same-view.tsx#L24-L32
> 
> Could be
> 
> ```js
> 	required: [ // name TBD
>		pageDetect.isForkedRepo
>	],
>	include: [
>		pageDetect.isSingleFile,
>		pageDetect.isRepoTree,
>		pageDetect.isEditingFile
>	],
>	exclude: [
>		pageDetect.isRepoRoot
>	],
>```
>
> Side note: https://github.com/sindresorhus/refined-github/pull/4797/files#r711719028

---

I will be changing all the instances of negative exclusions to their own list of `always` tests, until then, the PR will stay a draft. Please do review the changes I made to `source/features/index.ts` while adding the `always` test though.

### DIFF
--- a/source/features/index.tsx
+++ b/source/features/index.tsx
@@ -141,7 +141,7 @@ const globalReady: Promise<RGHOptions> = new Promise(async resolve => {
 const setupPageLoad = async (id: FeatureID, config: InternalRunConfig): Promise<void> => {
 	const {asLongAs, include, exclude, init, deinit, additionalListeners, onlyAdditionalListeners} = config;
 
-	// If any asLongAs is false, every `include` is false or no `exclude` is true, don’t run the feature
+	// Features are enabled if every `asLongAs` matches, at least one `include` matches, none of `exclude` matches
 	if (!asLongAs.every(c => c()) || include.every(c => !c()) || exclude.some(c => c())) {
 		return;
 	}
--- a/source/features/index.tsx
+++ b/source/features/index.tsx
@@ -34,7 +34,7 @@ interface FeatureLoader extends Partial<InternalRunConfig> {
 }
 
 interface InternalRunConfig {
-	always: BooleanFunction[];
+	asLongAs: BooleanFunction[];
 	include: BooleanFunction[];
 	exclude: BooleanFunction[];
 	init: FeatureInit;
@@ -139,10 +139,10 @@ const globalReady: Promise<RGHOptions> = new Promise(async resolve => {
 });
 
 const setupPageLoad = async (id: FeatureID, config: InternalRunConfig): Promise<void> => {
-	const {always, include, exclude, init, deinit, additionalListeners, onlyAdditionalListeners} = config;
+	const {asLongAs, include, exclude, init, deinit, additionalListeners, onlyAdditionalListeners} = config;
 
-	// If any always is false, every `include` is false or no `exclude` is true, don’t run the feature
-	if (!always.every(c => c()) || include.every(c => !c()) || exclude.some(c => c())) {
+	// If any asLongAs is false, every `include` is false or no `exclude` is true, don’t run the feature
+	if (!asLongAs.every(c => c()) || include.every(c => !c()) || exclude.some(c => c())) {
 		return;
 	}
 
@@ -219,7 +219,7 @@ const add = async (id: FeatureID, ...loaders: FeatureLoader[]): Promise<void> =>
 		// Input defaults and validation
 		const {
 			shortcuts = {},
-			always = [], // Default: nothing
+			asLongAs = [], // Default: nothing
 			include = [() => true], // Default: every page
 			exclude = [], // Default: nothing
 			init,
@@ -236,13 +236,13 @@ const add = async (id: FeatureID, ...loaders: FeatureLoader[]): Promise<void> =>
 		}
 
 		// 404 pages should only run 404-only features
-		if (pageDetect.is404() && !include.includes(pageDetect.is404) && !always.includes(pageDetect.is404)) {
+		if (pageDetect.is404() && !include.includes(pageDetect.is404) && !asLongAs.includes(pageDetect.is404)) {
 			continue;
 		}
 
 		enforceDefaults(id, include, additionalListeners);
 
-		const details = {always, include, exclude, init, deinit, additionalListeners, onlyAdditionalListeners};
+		const details = {asLongAs, include, exclude, init, deinit, additionalListeners, onlyAdditionalListeners};
 		if (awaitDomReady) {
 			(async () => {
 				await domLoaded;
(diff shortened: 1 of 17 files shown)


yakov116:
@gamemaker1 please lint before commiting thanks

gamemaker1:
Sorry about that I usually don't use semicolons, so the linter keeps failing. Maybe I should set a pre-commit hook for that 😅

source/features/useful-not-found-page.tsx
```diff
@@ -169,11 +155,21 @@ async function initPRCommit(): Promise<void | false> {
...
+	always: [pageDetect.is404],
 	include: [
-		pageDetect.is404,
+		pageDetect.isSingleFile,
+		pageDetect.isRepoTree,
+		pageDetect.isEditingFile,
 	],
-	init: onetime(init),
+	init: () => {
+		showDefaultBranchLink(); showAlternateLink();
```
yakov116, line 171: This should be put into a regular init function.

source/features/useful-not-found-page.tsx
```diff
@@ -169,11 +155,21 @@ async function initPRCommit(): Promise<void | false> {
 	);
 }
 
-void features.add(__filebasename, {
+void features.add(__filebasename, 	{
+	always: [pageDetect.is404],
```
yakov116, line 159:
```suggestion
	always: [
		pageDetect.is404
	],
```

In all places

cheap-glitch:
Thanks for the PR! Imo `required` is clearer than `always`, so I'd prefer the first one. Just my opinion though.

source/features/useful-not-found-page.tsx
yakov116: > In all places

gamemaker1:
> Thanks for the PR! Imo `required` is clearer than `always`, so I'd prefer the first one. Just my opinion though.

Ok, sure I'll change it to `required`.

fregante:
> Thanks for the PR! Imo `required` is clearer than `always`, so I'd prefer the first one. Just my opinion though.

Agreed, but is it clear that _every_ filter must be true? Is there anything better? In essence we have:

- every
- any – include 
- none – exclude

But I don't know if they make sense on their own

fregante:
Also keep in mind that we have a special handling of is404 in features/index.tsx, so make sure that's updated if necessary.

yakov116:
> asLongAs

👍

source/features/useful-not-found-page.tsx
```diff
@@ -169,10 +160,20 @@ async function initPRCommit(): Promise<void | false> {
 	);
 }
 
-void features.add(__filebasename, {
-	include: [
+void features.add(__filebasename, 	{
+	always: [
 		pageDetect.is404,
 	],
+	include: [() => parseCurrentURL().length > 1],
```
fregante, line 167: Since both are required

```suggestion
	always: [
		pageDetect.is404,
		() => parseCurrentURL().length > 1,
	],
```

gamemaker1:
Changed `always` to `asLongAs`

gamemaker1:
I think I got all of them, let me know if I missed any!

source/features/index.tsx
```diff
@@ -138,10 +139,10 @@ const globalReady: Promise<RGHOptions> = new Promise(async resolve => {
 });
 
 const setupPageLoad = async (id: FeatureID, config: InternalRunConfig): Promise<void> => {
-	const {include, exclude, init, deinit, additionalListeners, onlyAdditionalListeners} = config;
+	const {asLongAs, include, exclude, init, deinit, additionalListeners, onlyAdditionalListeners} = config;
 
-	// If every `include` is false and no `exclude` is true, don’t run the feature
-	if (include.every(c => !c()) || exclude.some(c => c())) {
+	// If any asLongAs is false, every `include` is false or no `exclude` is true, don’t run the feature
```
fregante, line 144:
```suggestion
	// Features are enabled if every `asLongAs` matches, at least one `include` matches, none of `exclude` matches
```

We just need to make sure that works as expected, we should probably extract this condition into a `shouldFeatureRun({include, exclude, asLongAs})` (in helpers.ts) and add tests for it.

source/features/link-to-github-io.tsx
```diff
@@ -42,8 +42,8 @@ async function initRepo(): Promise<void> {
 }
 
 void features.add(__filebasename, {
-	exclude: [
-		() => !getRepo()?.name.endsWith('.github.io'),
+	asLongAs: [
+		() => getRepo()?.name.endsWith('.github.io') ?? false,
```
fregante, line 46:
```suggestion
		() => Boolean(getRepo()?.name.endsWith('.github.io')),
```

gamemaker1:
Added a couple of tests, let me know if and which more tests I should add.

fregante:
Let's move these defaults to the new helper function:

https://github.com/sindresorhus/refined-github/pull/4800/files#diff-cb1dafea7a9c68c5c2ebb2d56f749b3506c13f7d26feeaf56c86521b25c60358R222-R224

Once you remove the `= []` from there, I think you'll only need to fix the 404 check line a bit later, from `include.includes` to `include?.includes`

gamemaker1:
Hmmm.. you can't specify a type and a default value in an object

gamemaker1:
Yay, checks pass! Thanks @fregante!

Any more changes needed?

fregante (approved):

source/features/index.tsx
```diff
@@ -10,6 +10,7 @@ import onNewComments from '../github-events/on-new-comments';
 import bisectFeatures from '../helpers/bisect';
 import optionsStorage, {RGHOptions} from '../options-storage';
 import {getLocalHotfixesAsOptions, updateHotfixes} from '../helpers/hotfix';
+import {shouldFeatureRun} from '../github-helpers';
```
yakov116, line 13: Should be moved up 2 lines

yakov116 (changes requested):
LGTM

Please also add this to the contribution.md

source/features/index.tsx
gamemaker1: Sure

gamemaker1:
> Please also add this to the contributing.md

Done!

yakov116 (approved):
Untested but LGTM

fregante:
Thank you Vedant! It works!

Issue #4679, opened 2021-08-18
`restore-file` occasionally restores the wrong commit


fregante:
### Please ensure:

- [X] The bug is caused by Refined GitHub. It doesn't happen if I disable the extension.

### Description

Rare occurrence but:

1. Create `X` branch from `main`
2. Force-push `main` to something else
3. Create PR from branch `X`

Now you'll have a PR with the changes that the force-push made.

4. Click "Restore file" on one of the files in the PR

You'll notice that the feature "succeeds" but the commits are empty. This is a live example: [`419c37c` (#4677)](https://github.com/sindresorhus/refined-github/pull/4677/commits/419c37c29ed4f4961d48495fadc951140c249f99)

### Example URL

https://github.com/sindresorhus/refined-github/pull/4677, before it was rebased

### Browser(s) used

Safari

fregante:
Maybe it's not even due to force pushes. There were no force pushes in https://github.com/npmhub/npmhub/pull/141, but this "Restore" commit is incorrect: [`7485182` (#141)](https://github.com/npmhub/npmhub/pull/141/commits/74851825ea433e27f411eb1733d3c6cdddd3c82c)

The issue specifically is that it "correctly" fetched the file from the base branch (`main`), but GitHub still shows the file as a "change" in the PR (even though when merged nothing will actually change): https://github.com/npmhub/npmhub/pull/141/files#diff-719f723e7f36f1f0192a75b8750569dcb1319113ea7e038109894895d2ea7d74

You can try restoring the same file multiple times but that's what you get.

In short, I think we should fetch the file not from the base _branch_ but from the base _commit_.

I think the repro is:

1. Create PR with file X
2. Change the same file X from the PR directly on the base branch
3. Restore file X in the PR

fregante:
"merge base" is what we're looking for: https://stackoverflow.com/q/1549146/288906

v3 seems to have `merge_base_commit`, which is likely what we want: https://docs.github.com/en/rest/commits/commits?apiVersion=2022-11-28#compare-two-commits

It seems that it also exists on v4 as `baseTarget`, I'll try it

```gql
{
  repository(owner: "refined-github", name: "sandbox") {
    pullRequest(number: 68) {
      baseRefOid
      mergeable
      viewerCanEditFiles
      headRef {
        compare(headRef: "default-a") {
          status
          aheadBy
          baseTarget {
            oid
          }
        }
      }
    }
  }
}
```

PR #4803, opened 2021-09-21
Use safer saving mechanism in `restore-file`

### LINKED ISSUE #4781: Look into using `createcommitonbranch` for `restore-file`
https://github.blog/changelog/2021-09-13-a-simpler-api-for-authoring-commits/
https://docs.github.com/en/graphql/reference/mutations#createcommitonbranch

gamemaker1:
## Related Issues

Closes #4781 

## Test URLs

https://github.com/gamemaker1/test/pull/1/files

### DIFF
--- a/source/features/restore-file.tsx
+++ b/source/features/restore-file.tsx
@@ -63,13 +63,7 @@ async function restoreFile(progress: (message: string) => void, menuItem: Elemen
 	}
 
 	const [nameWithOwner, prBranch] = select('.head-ref')!.title.split(':');
-	if (menuItem.closest('[data-file-deleted="true"]')) {
-		// The file was deleted by the PR, restore it from the base commit
-		progress('Undeleting…');
-	} else {
-		// The file has been modified, restore it to its original state
-		progress('Committing…');
-	}
+	progress(menuItem.closest('[data-file-deleted="true"]') ? 'Undeleting…' : 'Committing…');
 
 	const content = file.text;
 	await api.v4(`mutation {
--- a/source/features/restore-file.tsx
+++ b/source/features/restore-file.tsx
@@ -83,7 +83,7 @@ async function restoreFile(progress: (message: string) => void, menuItem: Elemen
 	}
 
 	const content = file.text;
-	await api.v4mutation(`mutation {
+	await api.v4(`mutation {
 		createCommitOnBranch(input: {
 			branch: {
 				repositoryNameWithOwner: "${owner}/${repo}",
--- a/source/features/restore-file.tsx
+++ b/source/features/restore-file.tsx
@@ -1,13 +1,11 @@
 import React from 'dom-chef';
 import select from 'select-dom';
 import onetime from 'onetime';
-import pushForm from 'push-form';
 import delegate from 'delegate-it';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
 import * as api from '../github-helpers/api';
-import fetchDom from '../helpers/fetch-dom';
 import showToast from '../github-helpers/toast';
 import {getConversationNumber} from '../github-helpers';
 
@@ -25,6 +23,17 @@ const getBaseReference = onetime(async (): Promise<string> => {
 	`);
 	return repository.pullRequest.baseRefOid;
 });
+const getHeadReference = async (): Promise<string> => {
+	// Get the sha of the latest commit to the PR, required to create a new commit
+	const {repository} = await api.v4(`
+		repository() {
+			pullRequest(number: ${getConversationNumber()!}) {
+				headRefOid
+			}
+		}
+	`);
+	return repository.pullRequest.headRefOid;
+};
 
 async function getFile(filePath: string): Promise<{isTruncated: boolean; text: string} | undefined> {
 	const {repository} = await api.v4(`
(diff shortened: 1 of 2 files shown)


source/features/restore-file.tsx
```diff
@@ -25,6 +23,17 @@ const getBaseReference = onetime(async (): Promise<string> => {
 	`);
 	return repository.pullRequest.baseRefOid;
 });
+const getHeadReference = async (): Promise<string> => {
```
gamemaker1, line 26: This is cached by the `mem` function in `api.ts`, so when you try to restore more than one file one after the other, it fails to get the latest commit after the first restore.

source/features/restore-file.tsx
```diff
@@ -53,27 +62,51 @@ async function restoreFile(progress: (message: string) => void, menuItem: Elemen
...
+					}
+				]
+			},
+			message: {
+				headline: "Restore ${filePath}"
+			}
+		}) {
+			commit {
+				oid
+			}
```
gamemaker1, line 107: Maybe I can remove this, we don't need anything from the server in response

source/features/restore-file.tsx
fregante: You're right, but I think it doesn't immediately affect this feature since each commit only affects one file… which is then gone and doesn't need to be restored again.

source/features/restore-file.tsx
fregante: We just need to confirm whether it was successful. How are errors reported back? Does `v4mutation` throw in that case?

source/github-helpers/api.ts
```diff
@@ -209,6 +209,45 @@ export const v4 = mem(async (
 	cacheKey: JSON.stringify,
 });
 
+export const v4mutation = mem(async (
```
fregante, line 212: Mutations should not be memoized. But also this specific method doesn't look necessary, why did you create it?

```suggestion
export const v4mutation = async (
```

fregante: But actually I think it's best to merge this into the regular v4, perhaps by detecting and allowing the starting `mutation` keyword

source/features/restore-file.tsx
```diff
@@ -53,27 +62,51 @@ async function restoreFile(progress: (message: string) => void, menuItem: Elemen
...
 	if (menuItem.closest('[data-file-deleted="true"]')) {
+		// The file was deleted by the PR, restore it from the base commit
 		progress('Undeleting…');
-		const [nameWithOwner, headBranch] = select('.head-ref')!.title.split(':');
-		pathname = `/${nameWithOwner}/new/${headBranch}?filename=${filePath}`;
+		// The `Edit` button doesn't exist if the file is deleted
+		const [nameWithOwner, prBranch] = select('.head-ref')!.title.split(':');
+		owner = nameWithOwner.split('/')[0];
+		repo = nameWithOwner.split('/')[1];
+		branch = prBranch;
```
fregante, line 78: I think this is unnecessary changes since owner and repo are being merged again later. Wasn't `nameWithOwner` already exactly that?

source/features/restore-file.tsx
```diff
@@ -53,27 +62,51 @@ async function restoreFile(progress: (message: string) => void, menuItem: Elemen
...
-	if (!response.ok) {
-		throw new Error(response.statusText);
-	}
+	await api.v4mutation(`mutation {
+		createCommitOnBranch(input: {
+			branch: {
+				repositoryNameWithOwner: "${owner}/${repo}",
+				branchName: "${branch}"
+			},
+			expectedHeadOid: "${await getHeadReference()}",
```
fregante, line 92: If this key isn't necessary, we can drop it

source/features/restore-file.tsx
```diff
@@ -53,27 +62,51 @@ async function restoreFile(progress: (message: string) => void, menuItem: Elemen
 		throw new Error('Restore failed: File too big');
 	}
 
-	let {pathname} = menuItem.previousElementSibling as HTMLAnchorElement;
-	// Check if file was deleted by PR
+	// Attempt to get details from the `Edit` button in the menu
+	const {pathname: blobOrEditPath} = menuItem.previousElementSibling as HTMLAnchorElement;
+	// eslint-disable-next-line @typescript-eslint/no-unused-vars -- Error comes up even though route and pathToFile are prefixed with a _
+	let [owner, repo, _route, branch, ..._pathToFile] = blobOrEditPath.startsWith('/')
```
fregante, line 68: If it's not being used, why create it? That's what the rule is about

```suggestion
	let [owner, repo, _route, branch] = blobOrEditPath.startsWith('/')
```

source/features/restore-file.tsx
fregante: Also you can move this line outside the `if` and avoid extracting the same information from `blobOrEditPath`

```js
const [nameWithOwner, prBranch] = select('.head-ref')!.title.split(':');
```

fregante (commented):
Sorry for the early reviews 😅

source/features/restore-file.tsx
gamemaker1: `v4mutation` is a copy of `v4`, so it does throw an error. I'll try merging the two functions...

gamemaker1:
> Sorry for the early reviews

No problem, they're helpful to see if I'm going in the right direction, thanks!

source/features/restore-file.tsx
gamemaker1: Ahh good point, will do

source/features/restore-file.tsx
gamemaker1: It is compulsory, otherwise we would have been saved from making one API call

source/github-helpers/api.ts
gamemaker1: Merged the two functions

source/features/restore-file.tsx
gamemaker1: Seems like [GraphQL does not allow returning 'nothing at all'](https://github.com/apollographql/graphql-tools/issues/277), but you can return a value that might be null. So I'm keeping this `commit { oid }` part.

source/features/restore-file.tsx
gamemaker1: Done!

source/features/restore-file.tsx
gamemaker1: This still remains a problem... is there a way to specify when calling the function, that we don't want a cached result from `mem`?

fregante: The closest solution I can think of is to use a cache buster 😬 
```js
 		repository() { # Cache buster ${Math.random()}
 			pullRequest(number: ${getConversationNumber()!}) {
 				headRefOid
 			}
 		}
 	`);
```

A full fix might be too involved (e.g. replacing `mem` with a custom deduplicator in order to allow skipping the cache). This might be useful at some point, but let's not do it now.

fregante: This cache busting might also be a solution to https://github.com/sindresorhus/refined-github/issues/4679 in `getBaseReference` (but can be handled in a separate PR)

gamemaker1: Added and it works!

source/github-helpers/api.ts
```diff
@@ -186,7 +186,7 @@ export const v4 = mem(async (
 			Authorization: `bearer ${personalToken}`,
 		},
 		method: 'POST',
-		body: JSON.stringify({query: `{${query}}`}),
+		body: JSON.stringify({query: (/^(mutation )?{/.test(query.trimStart())) ? query : `{${query}}`}),
```
fregante, line 189: Right?
```suggestion
		body: JSON.stringify({query: query.trimStart().startsWith('mutation') ? query : `{${query}}`}),
```

source/features/restore-file.tsx
```diff
@@ -53,27 +62,40 @@ async function restoreFile(progress: (message: string) => void, menuItem: Elemen
...
-	let {pathname} = menuItem.previousElementSibling as HTMLAnchorElement;
-	// Check if file was deleted by PR
+	const [nameWithOwner, prBranch] = select('.head-ref')!.title.split(':');
 	if (menuItem.closest('[data-file-deleted="true"]')) {
+		// The file was deleted by the PR, restore it from the base commit
 		progress('Undeleting…');
-		const [nameWithOwner, headBranch] = select('.head-ref')!.title.split(':');
-		pathname = `/${nameWithOwner}/new/${headBranch}?filename=${filePath}`;
 	} else {
+		// The file has been modified, restore it to its original state
```
fregante, line 70: I think these comments are redundant. Perhaps the line can be changed to

```js
progress(menuItem.closest('[data-file-deleted="true"]') ? 'Undeleting…' : 'Committing…');
```

fregante (approved):
Not yet tested, but it looks otherwise good

source/github-helpers/api.ts
gamemaker1: Yes, I just wanted to keep it consistent with the check on line 173:

https://github.com/gamemaker1/refined-github/blob/1cd99818a6564a5c56dd56b205a35e18195bf0be/source/github-helpers/api.ts#L173

fregante:
Confirmed to fix the issue

## Before

`restore-file` appended a EOF line break to the restored file, therefore not fully restoring it

https://user-images.githubusercontent.com/1402241/134794406-e8f46368-7720-4f0f-a06d-8ef09391d979.mov

## After

`restore-file` preserves the lack of EOF line break, therefore correctly restoring the file


https://user-images.githubusercontent.com/1402241/134794415-fa5f52a3-d09e-4cc5-8a63-e9b216fd7bb9.mov

source/features/restore-file.tsx
```diff
@@ -53,27 +62,34 @@ async function restoreFile(progress: (message: string) => void, menuItem: Elemen
...
-	// Check if file was deleted by PR
-	if (menuItem.closest('[data-file-deleted="true"]')) {
-		progress('Undeleting…');
-		const [nameWithOwner, headBranch] = select('.head-ref')!.title.split(':');
-		pathname = `/${nameWithOwner}/new/${headBranch}?filename=${filePath}`;
-	} else {
-		progress('Committing…');
-	}
+	const [nameWithOwner, prBranch] = select('.head-ref')!.title.split(':');
+	progress(menuItem.closest('[data-file-deleted="true"]') ? 'Undeleting…' : 'Committing…');
```
kidonng, line 66: A little confused why it's called "Undelete" instead of "Restore", "Revert", "Undo" or whatever

gamemaker1: It sounds funny :P

But otherwise maybe restore is more in line with the feature name...

PR #4804, opened 2021-09-21
Meta: Drop "Feature history" dropdown

fregante:
The dropdown is a little inconvenient because it closes every time you click on a commit, making it difficult to open more than one at a time. I think it's easier to just open the History page.

One possible meta feature in that case would be to dim every meta/lint commit in the regular commit list of Refined GitHub, using the styles of `dim-bots`


kidonng:
> One possible meta feature in that case would be to dim every meta/lint commit in the regular commit list of Refined GitHub, using the styles of [`dim-bots`](/sindresorhus/refined-github/blob/main/source/features/dim-bots.tsx)

I think if we were to differentiate those commits, it's better to highlight everything else other than dimming those (because they are still authored by real human).

PR #4816, opened 2021-09-25
Meta: Improve `conversation-activity-filter` labels

### LINKED ISSUE #4063: Hide/rename “Only show unresolved comments” on issues
I’m using `conversation-activity-filter` in the recently released Safari extension.

It shows an option for “Only show unresolved comments” ~even though there’s no concept of “resolved comments” on issues (as far as I know)~

Correction: this choice also hides hidden comments — maybe the text should change to clarify that (perhaps only for issues)?

kidonng:
Resolve #4063

## Test URLs

Here

## Screenshot

Before:

image

Now:
image


fregante (approved):
Sounds reasonable 👍

PR #4818, opened 2021-09-25
Meta: Fix `rgh-feature-descriptions` feature

### LINKED ISSUE #4815: `rgh-feature-descriptions` is broken
### Please ensure:

### Description

Again! It seems Github removed both `.hx_commit-tease` and `include-fragment.commit-loader`. The closest other one I could find is `#blob_contributors_box`.

### Screenshot

image

### Console errors

```shell
❌ rgh-feature-descriptions 21.9.18 → TypeError: undefined has no properties
    init rgh-feature-descriptions.tsx:98
index.tsx:57:9
```

### Example URL

https://github.com/sindresorhus/refined-github/blob/main/source/features/batch-mark-files-as-viewed.tsx

### Browser(s) used

Firefox

gamemaker1:
Changes the selector to select the commit box to `.Box-header.Details`, as suggested in https://github.com/sindresorhus/refined-github/issues/4815#issuecomment-927072107.

Closes #4815 if merged.

### DIFF
--- a/source/features/rgh-feature-descriptions.tsx
+++ b/source/features/rgh-feature-descriptions.tsx
@@ -13,13 +13,13 @@ async function init(): Promise<void | false> {
 
 	const conversationsUrl = '/sindresorhus/refined-github/issues?q=' + encodeURIComponent(`"${feature.id}" sort:updated-desc`);
 
-	const commitInfoBox = (await elementReady([
-		'.hx_commit-tease', // Selector for commit message
-		'.Box-header.Details', // Generic selector for entire commit box
-		'include-fragment.commit-loader', // Required according to 2683e30
-	].join(',')))!.parentElement!;
-	commitInfoBox.classList.add('width-fit', 'min-width-0', 'flex-auto', 'mb-lg-0', 'mr-lg-3');
-	commitInfoBox.classList.remove('flex-shrink-0');
+	const commit = await elementReady([
+		'.Box-header.Details', // Already loaded
+		'include-fragment.commit-loader', // Deferred loading
+	].join(','));
+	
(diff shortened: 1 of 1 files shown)


source/features/rgh-feature-descriptions.tsx
```diff
@@ -13,7 +13,7 @@ async function init(): Promise<void | false> {
 
 	const conversationsUrl = '/sindresorhus/refined-github/issues?q=' + encodeURIComponent(`"${feature.id}" sort:updated-desc`);
 
-	const commitInfoBox = (await elementReady('.hx_commit-tease, include-fragment.commit-loader'))!.parentElement!;
+	const commitInfoBox = (await elementReady('.Box-header.Details, include-fragment.commit-loader'))!.parentElement!;
```
gamemaker1, line 16: Should we drop `include-fragment.commit-loader`?

kidonng: > `include-fragment.commit-loader` is needed according to [`2683e30` (#3852)](https://github.com/sindresorhus/refined-github/pull/3852/commits/2683e300976cacbbb71eb219347cb82aa964616b).

fregante:
If they're cross-compatible, let's keep the 3 selectors so we don't have to go back and forth for now. Also it's best to use a selector array with a comment after each selector

gamemaker1:
> ... selector array with a comment after each selector

`elementReady` doesn't seem to accept an array. Anything I'm missing?

fregante:
> > ... selector array with a comment after each selector
> 
> `elementReady` doesn't seem to accept an array. Anything I'm missing?

['a', 'b'].join(',')

fregante:
I went back to look at the old PR: The selector was only broken back then because they had dropped the `--blue` class, but was otherwise valid as a generic "Box" selector. In short, the `hx` class was never needed.

PR #4821, opened 2021-09-25
Fix `bugs-tab` highlighting

### LINKED ISSUE #4782: `bugs-tab` removes highlight from "Issues" tab after ajax navigation
### Please ensure:

### Description

1. Click Issues tab
2. Click Issues tab

One click may be enough.

This doesn't happen if `bugs-tab` is disabled

### Screenshot

https://user-images.githubusercontent.com/1402241/133208841-a2c0bbc3-8fa3-4da4-9c76-a94647d00450.mov

### Console errors

_No response_

### Example URL

https://github.com/sindresorhus/refined-github/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc

### Browser(s) used

Safari 14

kidonng:
Fix #4782

## Test URLs

Here


source/features/bugs-tab.tsx
```diff
@@ -17,7 +17,10 @@ const getBugLabel = async (): Promise<string | undefined> => cache.get<string>(g
 const isBugLabel = (label: string): boolean => supportedLabels.test(label.replace(/\s/g, ''));
 
 async function highlightBugsTabOnIssuePage(): Promise<void | false> {
-	if (await countBugs() === 0 || !await elementReady('#partial-discussion-sidebar .IssueLabel[href$="/bug" i]')) {
+	if (await countBugs() === 0
+	|| (pageDetect.isIssue()
+		? !await elementReady('#partial-discussion-sidebar .IssueLabel[href$="/bug" i]')
+		: !new SearchQuery(location.search).includes(`label:${SearchQuery.escapeValue(await getBugLabel() ?? 'bug')}`))) {
```
cheap-glitch, line 23: Maybe split this into (at least) two `if`? It's not very readable

source/features/bugs-tab.tsx
```diff
@@ -17,7 +17,15 @@ const getBugLabel = async (): Promise<string | undefined> => cache.get<string>(g
...
-	if (await countBugs() === 0 || !await elementReady('#partial-discussion-sidebar .IssueLabel[href$="/bug" i]')) {
+	if (await countBugs() === 0) {
+		return false;
+	}
+
+	if (pageDetect.isIssue() && !await elementReady('#partial-discussion-sidebar .IssueLabel[href$="/bug" i]')) {
+		return false;
+	}
+
+	if (!pageDetect.isIssue() && !new SearchQuery(location.search).includes(`label:${SearchQuery.escapeValue(await getBugLabel() ?? 'bug')}`)) {
```
fregante, line 28: Maybe

```suggestion
	if (pageDetect.isRepoConversationList() && !new SearchQuery(location.search).includes(`label:${SearchQuery.escapeValue(await getBugLabel() ?? 'bug')}`)) {
```

source/features/bugs-tab.tsx
```diff
@@ -111,7 +119,7 @@ async function init(): Promise<void | false> {
 	// Disable unwanted behavior #3001
 	bugsTab.removeAttribute('data-hotkey');
 	bugsTab.removeAttribute('data-selected-links');
-	issuesTab.removeAttribute('data-selected-links');
+	bugsTab.removeAttribute('id');
```
fregante, line 122: Ensure that this bug isn't reintroduced: https://github.com/sindresorhus/refined-github/issues/3001

source/features/bugs-tab.tsx
```diff
@@ -158,7 +166,7 @@ void features.add(__filebasename, {
 	init,
 }, {
 	include: [
-		pageDetect.isIssue,
+		pageDetect.isRepo,
```
fregante, line 169: This changes `highlightBugsTabOnIssuePage` from being just for single issues to acting on every page. This was not necessary before so this is likely currently duplicating existing functionality

source/features/bugs-tab.tsx
fregante: Confirmed it's fine at this moment 👍

source/features/bugs-tab.tsx
kidonng: The problem is that we don't highlight the "Bugs" label again when navigating through `isRepo` pages, because `init` only runs once. I tried to make it work with `deduplicate` (see first commit of this PR) but it doesn't work correctly. Hence I think we should just run `highlightBugsTabOnIssuePage` all the time.

fregante: > Hence I think we should just run `highlightBugsTabOnIssuePage` all the time.

Yeah that's fine, but we probably need to drop/adapt the code that took care of this logic in the main `init` function so that the functionality does not overlap.

Maybe that was already taken care of though?

```diff
- 	issuesTab.removeAttribute('data-selected-links');
```

fregante:
Let me know if this is good to go or if the unresolved reviews need extra fixes

fregante:
If you can finish this this weekend I'll release another version soon since I missed a few bugfixes.

kidonng:
Sure, I was reinstalling and setting up my system, or this could have been done 2 days ago.

kidonng:
Just push your code 😄

PR #4824, opened 2021-09-27
Meta: show screenshots in options page

### LINKED ISSUE #4810: Hover on the `screenshot` to preview
### Description

How about adding support for hovering on the `screenshot` to preview?

Or use an icon or emoji instead of the text for hovering.

Example:

Untitled-1

### Screenshot

_No response_

### Example URL

_No response_

kidonng:
Resolve #4810

## Test URLs

chrome://extensions/?options=balgliakcbamepammkbcjdepogihccmj (😄)

### DIFF
--- a/source/options.css
+++ b/source/options.css
@@ -155,3 +155,14 @@ li[data-validation] {
 #debugging p:last-child {
 	margin-bottom: 0;
 }
+
+.show-screenshots-label {
+	float: right;
+}
+
+img {
+	max-width: 100%;
+	margin-top: 0.5em;
+	border: 1px solid #d1d5da;
+	border-radius: 0.2em;
+}
--- a/source/options.tsx
+++ b/source/options.tsx
@@ -107,6 +107,10 @@ function buildFeatureCheckbox({id, description, screenshot}: FeatureMeta): HTMLE
 	const descriptionElement = domify.one(description)!;
 	descriptionElement.className = 'description';
 
+	if (screenshot) {
+		descriptionElement.append(<img hidden data-src={screenshot} loading="lazy"/>);
+	}
+
 	return (
 		<div className="feature" data-text={`${id} ${description}`.toLowerCase()}>
 			<input type="checkbox" name={`feature:${id}`} id={id}/>
@@ -147,6 +151,16 @@ async function findFeatureHandler(event: Event): Promise<void> {
 	}, 10_000);
 
 	select('#find-feature-message')!.hidden = false;
+}
+
+function showScreenshotsHandler(): void {
+	for (const screenshot of select.all('img')) {
+		if (!screenshot.hasAttribute('src')) {
+			screenshot.setAttribute('src', screenshot.getAttribute('data-src')!);
+		}
+
+		screenshot.hidden = !screenshot.hidden;
+	}
 }
 
 function featuresFilterHandler(event: Event): void {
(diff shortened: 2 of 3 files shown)


source/options.css
```diff
@@ -155,3 +155,14 @@ li[data-validation] {
...
 }
+
+.show-screenshots-label {
+	float: right;
+}
+
+img {
+	max-width: 100%;
+	margin-top: 0.5em;
+	border: 1px solid #d1d5da;
```
kidonng, line 166: This uses the same border color as `<kbd>`, it works great on both Chrome's light and dark mode. As for the `border-radius` I choose the same number as `.feature-name`,

cheap-glitch:
Just tested on Firefox, it doesn't work (the screenshots don't appear at all). No errors in the "Inspect" console.

Also the checkbox is misaligned (the black input is a known bug):
image

kidonng:
What Firefox version are you using? Works fine for me with 92.0.1 on macOS:

https://user-images.githubusercontent.com/44045911/134947660-08772f8f-6450-484a-b0ca-91454ebc9ea0.mov

yakov116:
What do you think of adding a link to view a single screenshot inline? 
Would that be duplicate ad unnecessary?

kidonng:
I still prefer using `<details>`, albeit some code are needed to load the image later:

https://user-images.githubusercontent.com/44045911/134953333-5fa2d3f6-f1f3-4e10-bf4d-7c79611ca9da.mov

fregante:
I still don't prefer it for the same reason as before: it saves no clicks and it **only** shows the screenshot in a small area

kidonng:
> it saves no clicks

It's not for saving clicks, it just keeps the context. I believe that's the intention in the original issue.

> and it **only** shows the screenshot in a small area

Unfortunately we can't do much about this, however we show the screenshots.

Do you want me to revert to the first implementation (with a single checkbox)?

fregante:
> Do you want me to revert to the first implementation (with a single checkbox)?

Yes please

cheap-glitch:
> I still don't prefer it for the same reason as before: it saves no clicks and it **only** shows the screenshot in a small area

Showing all the screenshots at once makes the list unreadable imho. (Plus in Firefox the options container gets an annoying scrollbar that interferes with the page scroll.)

Why not make the thumbnails inside each `<details>` element an external link pointing to the full-size image? That does mean two clicks instead of one to get to it, but in most cases the thumbnail should be clear enough on its own.

fregante:
> Showing all the screenshots at once makes the list unreadable imho. 

Which is why it would not be a default. 

> in Firefox the options container gets an annoying scrollbar that interferes with the page scroll.

I'm not sure of what you refer to. Is this a problem during loading of the many images or does it just appear when many images have been loaded? In the latter case it probably appears with this solution too. 

> Why not make the thumbnails inside each `<details>` element an external link pointing to the full-size image?

Ok sounds good. Let's keep the `details`. I wonder if should still have a "toggle all" behavior

kidonng:
> I wonder if should still have a "toggle all" behavior

`toggle-everything-with-alt`?

source/features/batch-mark-files-as-viewed.tsx
```diff
@@ -4,7 +4,7 @@ import * as pageDetect from 'github-url-detection';
 
 import features from '.';
 import showToast from '../github-helpers/toast';
-import {clickAll} from './toggle-everything-with-alt';
+import clickAll from '../helpers/click-all';
```
yakov116, line 7: Lint

source/options.tsx
```diff
@@ -11,6 +11,7 @@ import * as indentTextarea from 'indent-textarea';
 import {getLocalHotfixes} from './helpers/hotfix';
 import {perDomainOptions} from './options-storage';
 import {createRghIssueLink} from './helpers/rgh-issue-link';
+import clickAll from './helpers/click-all';
```
yakov116, line 14: lint

fregante:
I didn't like the style that summary offered, so I hid it behind the old link, which remains clickable via cmd+click.



Missing: 

- [ ] Restore alt-click

PR #4826, opened 2021-09-27
Fix `unfinished-comments` feature

### LINKED ISSUE #4737: `unfinished-comments` not working if user hasn't filled all required fields
### Please ensure:

### Description

1. Open the [example URL](https://github.com/sindresorhus/refined-github/issues/new?assignees=&labels=under+discussion&template=3_discussion.md)
2. Write something in comment, leaving the title field empty
3. Switch to another tab, "(Draft comment)" doesn't show in tab title
4. Write something in title, switch to another tab
5. "(Draft comment)" shows in tab title

I guess it's because when the title field is empty, "Submit new issue" button is disabled because title is required, however RGH thinks the issue is being submitted according the code here:

https://github.com/sindresorhus/refined-github/blob/7e96d94b06d8f522fdbc5baeb9277f7c2b3da2fc/source/features/unfinished-comments.tsx#L13

### Screenshot

_No response_

### Console errors

_No response_

### Example URL

https://github.com/sindresorhus/refined-github/issues/new?assignees=&labels=under+discussion&template=3_discussion.md

### Browser(s) used

Firefox 91.0.2 & Chrome 92.0.4515.159

arielllin:
fix #4737 

## Test URLs
https://github.com/sindresorhus/refined-github/issues/new?assignees=&labels=under+discussion&template=3_discussion.md

## Screenshot
before:
before

after:
after

### DIFF
--- a/source/features/unfinished-comments.tsx
+++ b/source/features/unfinished-comments.tsx
@@ -13,10 +13,7 @@ function hasDraftComments(): boolean {
 		&& (
       !select.exists('.btn-primary[disabled]', textarea.form!) // Exclude forms being submitted
       || 
-      (
-        select.exists('input[aria-label="Title"]', textarea.form!)
-        && !(select('input[aria-label="Title"]', textarea.form!))?.value // Exclude forms doesn't fill in required title but have draft comments
-      )
+      select.exists('input[aria-label="Title"]', textarea.form!) // Exclude forms doesn't fill in required title but have draft comments
     )
 	);
 }
--- a/source/features/unfinished-comments.tsx
+++ b/source/features/unfinished-comments.tsx
@@ -10,7 +10,14 @@ function hasDraftComments(): boolean {
 	// `[id^="convert-to-issue-body"]` excludes the hidden pre-filled textareas created when opening the dropdown menu of review comments
 	return select.all('textarea:not([disabled], [id^="convert-to-issue-body"])').some(textarea =>
 		textarea.value !== textarea.textContent // Exclude comments being edited but not yet changed (and empty comment fields)
-		&& !select.exists('.btn-primary[disabled]', textarea.form!), // Exclude forms being submitted
+		&& (
+      !select.exists('.btn-primary[disabled]', textarea.form!) // Exclude forms being submitted
+      || 
+      (
+        select.exists('input[aria-label="Title"]', textarea.form!)
+        && !(select('input[aria-label="Title"]', textarea.form!))?.value // Exclude forms doesn't fill in required title but have draft comments
+      )
+    )
 	);
 }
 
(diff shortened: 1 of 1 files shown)


fregante:
Thank you for the PR! Can you also test it on the bug report form?

source/features/unfinished-comments.tsx
```diff
@@ -10,7 +10,10 @@ function hasDraftComments(): boolean {
 	// `[id^="convert-to-issue-body"]` excludes the hidden pre-filled textareas created when opening the dropdown menu of review comments
 	return select.all('textarea:not([disabled], [id^="convert-to-issue-body"])').some(textarea =>
 		textarea.value !== textarea.textContent // Exclude comments being edited but not yet changed (and empty comment fields)
-		&& !select.exists('.btn-primary[disabled]', textarea.form!), // Exclude forms being submitted
+		&& (
+			!select.exists('.btn-primary[disabled]', textarea.form!) // Exclude forms being submitted
+			|| select.exists('input[aria-label="Title"]', textarea.form!) // Exclude forms doesn't fill in required title but have draft comments
```
fregante, line 15: cc @cheap-glitch 

I'm stumped here. The `disabled` check was added in https://github.com/sindresorhus/refined-github/pull/3707, apparently to avoid a "click submit and immediately change tab" situation.

I tried looking into it and the same exact mechanism is used to disable the button when the form is invalid (the former via `button[data-disable-with]` and the latter with `button[data-disable-invalid]`), both affecting the same attribute. The submission does not affect other attributes.

I think the solution is to _do_ track the `submit` event (via `capture: true`, probably), and then just disable the feature for ~2 seconds after, so this filter does not have to run at all.

Something like:

```ts
function updateDocumentTitle(): void {
	if (submitting) {
		return;
	}

	// the rest
}
function disableOnSubmit() {
	clearTimeout(submitting);
	submitting = setTimeout(() => {
		submitting = undefined;
	});
}
function init(): void {
	document.addEventListener('visibilitychange', updateDocumentTitle);
	delegate(document, 'submit', disableOnSubmit)
}
```

fregante (changes requested):

source/features/unfinished-comments.tsx
fregante: In short, this would replace/drop `select.exists('.btn-primary[disabled]', textarea.form)` and thus solve the issue.

cheap-glitch: So if I understood correctly, there's no way to distinguish between an form that's invalid and one that's being submitted… :sweat: \
In that case I can't think of a better solution than to listen to the `submit` event. Should probably be a new PR though.

fregante:
I'll close this PR because the desired solution is different, but feel free to open a new one if you want to apply the changes suggested in https://github.com/sindresorhus/refined-github/pull/4826#discussion_r720666321

PR #4842, opened 2021-10-01
Fix `clean-dashboard`’s background color

### LINKED ISSUE #4841: White github.com due to `clean-dashboard`
### Please ensure:

### Description

Screen Shot

### Screenshot

_No response_

### Console errors

_No response_

### Example URL

https://github.com/

### Browser(s) used

safari

fregante:
- Fixes #4841

## Test URLs

https://github.com/

## Before
Screen Shot
## After
Screen Shot 1

### DIFF
--- a/source/features/clean-dashboard.css
+++ b/source/features/clean-dashboard.css
@@ -16,7 +16,7 @@
 
 /* Add background and border Whole list */
 .rgh-clean-dashboard .news > .f4 + div {
-	background-color: var(--color-bg-canvas, #fff);
+	background-color: var(--color-canvas-default, var(--color-bg-canvas));
 	border: 1px solid var(--github-border-color);
 	border-radius: 6px;
 	margin-top: 8px !important;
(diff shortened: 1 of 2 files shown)


source/refined-github.css
```diff
@@ -2,7 +2,7 @@
 	/* 2020-12-15: these variables only exist for GHE #3798 */
 	--github-green: var(--color-text-success, #28a745);
 	--github-red: var(--color-text-danger, #cb2431);
-	--github-border-color: var(--color-border-primary, #e1e4e8);
+	--github-border-color: var(--color-border-muted, #e1e4e8);
```
fregante, line 5: This is also affecting other features. Here's the before/after

Screen Shot 9 Screen Shot 7


Screen Shot 8 Screen Shot 4

source/features/clean-dashboard.css
```diff
@@ -16,7 +16,7 @@
 
 /* Add background and border Whole list */
 .rgh-clean-dashboard .news > .f4 + div {
-	background-color: var(--color-bg-canvas, #fff);
+	background-color: var(--color-canvas-default, var(--color-bg-canvas));
```
fregante, line 19: Also drops the default hardcoded white. Now it will default to transparent instead, which is almost good anyway.

source/refined-github.css
fregante: Like https://github.com/sindresorhus/refined-github/pull/3806

PR #4848, opened 2021-10-02
Fix `quick-repo-deletion` red background

MrBartusek:
In the case when token without `delete_repo` scope is provided the Delete Fork button will just result in `Could not delete the repository` message. The problem is that this doesn't actually close the details of this button and the whole website just stays with a red cover until user closes it manually.

main

This PR fixes it

new

### DIFF
--- a/source/features/quick-repo-deletion.tsx
+++ b/source/features/quick-repo-deletion.tsx
@@ -36,11 +36,10 @@ function handleToggle(event: delegate.Event<Event, HTMLDetailsElement>): void {
 	setTimeout(start, 1, event.delegateTarget);
 }
 
-async function verifyScopesWhileWaiting(abortController: AbortController): Promise<void> {
+async function verifyScopes(): Promise<boolean> {
 	try {
 		await api.expectTokenScope('delete_repo');
 	} catch (error: unknown) {
-		abortController.abort();
 		addNotice([
 			'Could not delete the repository. ',
 			parseBackticks((error as Error).message),
@@ -52,7 +51,9 @@ async function verifyScopesWhileWaiting(abortController: AbortController): Promi
 				</a>
 			),
 		});
+		return false
 	}
+	return true
 }
 
 async function buttonTimeout(buttonContainer: HTMLDetailsElement): Promise<boolean> {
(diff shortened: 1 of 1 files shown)


source/features/quick-repo-deletion.tsx
```diff
@@ -64,7 +65,10 @@ async function buttonTimeout(buttonContainer: HTMLDetailsElement): Promise<boole
 		buttonContainer.open = false;
 	}, {once: true});
 
-	void verifyScopesWhileWaiting(abortController);
+	if(!await verifyScopes()) {
+		document.body.click()
```
MrBartusek, line 69: This trips `The modal was closed too quickly. Disabling close-out-of-view-modals for this session.` and I am not sure how to fix it

source/features/quick-repo-deletion.tsx
```diff
@@ -64,7 +65,10 @@ async function buttonTimeout(buttonContainer: HTMLDetailsElement): Promise<boole
 		buttonContainer.open = false;
 	}, {once: true});
 
-	void verifyScopesWhileWaiting(abortController);
```
kidonng, line 67: This change is unrelated (and should be reverted). You are making the feature wait for scope verification, which is currently asynchronous.

source/features/quick-repo-deletion.tsx
kidonng: I haven't look into the cause but this looks like a hack more than a fix.

source/features/quick-repo-deletion.tsx
MrBartusek: The idea behind this change is so the button won't change intro `Deleting repo in` so there is less visual mess going on when token doesn't have the scope

fregante: The mess is fine as it's only supposed to happen once, if misconfigured. This just makes the operation slower every time instead.

source/features/quick-repo-deletion.tsx
fregante: The overlay is because the button is in a "open" state. The fix should just ensure that the dropdown is closed when the promise is aborted.

fregante:
Thanks for the PR Bartosz! Feel free to open a new PR if you decide to follow the suggest solution

PR #4851, opened 2021-10-02
Fix inconsistent `show-names` color

### LINKED ISSUE #4847: `show-names` text color is inconsistent
### Please ensure:

### Description

Notice how the regular names are gray (just like time) but not in review comments:

image

This is because color from `.review-comment` is overriding the color from `.TimelineItem-body` (P.S. `.timeline-comment-header` also uses gray color).

### Screenshot

_No response_

### Console errors

_No response_

### Example URL

https://github.com/sindresorhus/refined-github/pull/4821#discussion_r716093511

### Browser(s) used

Chrome

MrBartusek:
Fixes #4847

## Test URLs

https://github.com/sindresorhus/refined-github/pull/4821#discussion_r716093511

## Screenshot

Before
image

After
image

### DIFF
--- a/source/features/show-names.css
+++ b/source/features/show-names.css
@@ -2,7 +2,3 @@
 .timeline-comment-header-text bdo {
 	display: inline-block;
 }
-
-.review-comment h4 {
-	color: var(--color-fg-muted, #8b949e);
-}
--- a/source/features/show-names.css
+++ b/source/features/show-names.css
@@ -2,3 +2,7 @@
 .timeline-comment-header-text bdo {
 	display: inline-block;
 }
+
+.review-comment h4 {
+	color: var(--color-fg-muted, #8b949e);
+}
--- a/source/features/show-names.tsx
+++ b/source/features/show-names.tsx
@@ -46,19 +46,20 @@ async function init(): Promise<false | void> {
 		const userKey = api.escapeKey(username);
 
 		// For the currently logged in user, `names[userKey]` would not be present.
-		if (names[userKey]?.name) {
+		const {name} = names[userKey] ?? {};
+
+		if (name) {
 			// If it's a regular comment author, add it outside <strong>
 			// otherwise it's something like "User added some commits"
-			if (compareNames(username, names[userKey].name)) {
-				usernameElement.textContent = names[userKey].name;
+			if (compareNames(username, name)) {
+				usernameElement.textContent = name;
 			} else {
-				const insertionPoint = usernameElement.parentElement!.tagName === 'STRONG' ? usernameElement.parentElement! : usernameElement;
+				const insertionPoint = usernameElement.parentElement!.tagName === 'STRONG'
+					? usernameElement.parentElement!
+					: usernameElement;
 				insertionPoint.after(
-					// That obnoxious part is to comfort the linter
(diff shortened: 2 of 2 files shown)


source/features/show-names.css
```diff
@@ -2,3 +2,7 @@
 .timeline-comment-header-text bdo {
 	display: inline-block;
 }
+
+.review-comment h4 {
```
kidonng, line 6: No need to write a new rule, just use `.color-text-secondary`.

kidonng:
You can change the following lines:

https://github.com/sindresorhus/refined-github/blob/1d858670461a256b29dda730112a0e965b48f6f2/source/features/show-names.tsx#L56-L64

To:

```tsx
				insertionPoint.after(
					// That obnoxious part is to comfort the linter
					<span className="color-text-secondary">{`${'('}`}
						<bdo className="css-truncate css-truncate-target" style={{maxWidth: '200px'}}>
							{names[userKey].name}
						</bdo>)
					</span>,
				);
```

source/features/show-names.tsx
```diff
@@ -54,13 +54,12 @@ async function init(): Promise<false | void> {
 			} else {
 				const insertionPoint = usernameElement.parentElement!.tagName === 'STRONG' ? usernameElement.parentElement! : usernameElement;
 				insertionPoint.after(
-					' (',
-					<bdo className="css-truncate d-inline-block">
-						<span className="css-truncate-target" style={{maxWidth: '200px'}}>
+					// That obnoxious part is to comfort the linter
+					<span className="color-text-secondary">{`${'('}`}
```
fregante, line 58: Lol what a mess here, you're right. I'm gonna open an issue about this on the linter’s repo

source/features/show-names.tsx
```diff
@@ -54,13 +54,12 @@ async function init(): Promise<false | void> {
...
+						<bdo className="css-truncate css-truncate-target" style={{maxWidth: '200px'}}>
```
fregante, line 59: The css-truncate class should be on the parent, or else I don't remember if it affects css-truncate-target

source/features/show-names.tsx
```diff
@@ -46,21 +46,21 @@ async function init(): Promise<false | void> {
...
 				insertionPoint.after(
-					' (',
-					<bdo className="css-truncate d-inline-block">
-						<span className="css-truncate-target" style={{maxWidth: '200px'}}>
-							{names[userKey].name}
-						</span>
-					</bdo>,
-					') ',
+					<span className="color-text-secondary css-truncate">
+						(<bdo className="css-truncate-target" style={{maxWidth: '200px'}}>{name}</bdo>)
```
fregante, line 62: It'd be great to check whether the `maxWidth` and the `insertionPoint` logic are still needed as well, but I'll merge this as is for now

fregante (approved):

fregante:
Dziękuję za pomoc Bartosz ☺️

source/features/show-names.tsx
kidonng: That CSS rule also works if the two classes are in the same element, however I agree we should keep the original form. An oversight when I was tweaking elements and linter😃

yakov116:
After this PR

image

PR #4856, opened 2021-10-02
Run `show-whitespace` on code blocks but not diff headers

### LINKED ISSUE #4843: `show-whitespace` should not run on diff/patch headers
Not really a bug, but it's not useful here

Screen Shot 10

### Example URL

https://github.com/sindresorhus/refined-github/commit/e8ec186217ffc42f351983b22187ee14efb9251b#diff-6a8348ff857b16e455a52ddd651b49f31cd05895459c37301051a0e30bcf077c

### Browser(s) used

safari

kidonng:
Fix #4843

Also runs `show-whitespace` on code blocks now:

```
	A
	Block
	Without
	Highlight
```

```html
	<A Block>
		With Highlight
	</A>
```

## Test URLs

https://github.com/sindresorhus/refined-github/commit/e8ec186217ffc42f351983b22187ee14efb9251b#diff-6a8348ff857b16e455a52ddd651b49f31cd05895459c37301051a0e30bcf077c

## Screenshot

Before:

image

image

After:

image

image

### DIFF
--- a/source/features/show-whitespace.tsx
+++ b/source/features/show-whitespace.tsx
@@ -62,7 +62,13 @@ const viewportObserver = new IntersectionObserver(changes => {
 });
 
 function init(): void {
-	for (const line of select.all('.blob-code-inner:not(.rgh-observing-whitespace)')) {
+	const selectors = [
+		'.blob-code-inner', // Code lines
+		'.highlight pre', // Highlighted code blocks in comments
+		'.snippet-clipboard-content pre', // Not highlighted code blocks in comments
+	].join(',');
+
+	for (const line of select.all(`:is(${selectors}):not(.rgh-observing-whitespace, .blob-code-hunk)`)) {
 		line.classList.add('rgh-observing-whitespace');
 		viewportObserver.observe(line);
 	}


fregante:
Great!

I think the multiple selector will actually cause a similar issue we had with the issue linkification (which I think is still open). In short, review comments in files might have their own code blocks. Those selectors might match them twice. 

And no, the rgh- class won't prevent that because the selector picks them all at once *and then* the class is added inside the loop, at which point it's too late.

kidonng:
These features do share some characteristics, I will try to align the PR for `linkify-code` and `shorten-links` with this.

source/features/show-whitespace.tsx
```diff
@@ -62,7 +62,13 @@ const viewportObserver = new IntersectionObserver(changes => {
...
 
 function init(): void {
-	for (const line of select.all('.blob-code-inner:not(.rgh-observing-whitespace)')) {
+	const selectors = [
+		'.blob-code-inner', // Code lines
+		'.highlight pre', // Highlighted code blocks in comments
+		'.snippet-clipboard-content pre', // Not highlighted code blocks in comments
+	].join(',');
+
+	for (const line of select.all(`:is(${selectors}):not(.rgh-observing-whitespace, .blob-code-hunk)`)) {
```
fregante, line 71: Are you saying that PR isn't affected? Let's try 

```suggestion
	for (demo) {
```

To check, you'll have to include a `.prepend(1)` to the function that parses the spaces. If the suggestion has two 1s, then it's affected.

fregante: Interesting, it does seem to work correctly:

<table>
<tr>
	<td>Screen Shot 8

</table>

fregante (approved):

kidonng:
Wow, my 100th commit 🎉 thanks for all these years @fregante!

image

source/features/show-whitespace.tsx
kidonng: Here's my conclusion:

- `.blob-code-inner` selects lines in _files_ (that is, file content, review suggestions, etc.)
- `.highlight > pre` selects highlighted code blocks in _comments_ (yes, it should have a `>`, since the file content also has `.highlight`, but luckily it is caught by `.rgh-observing-whitespace`).
- `.snippet-clipboard-content > pre` selects _not_ highlighted code blocks in _comments_ (`>` is optional here but I feel it is safer.)

yakov116:
Your welcome!

PR #4857, opened 2021-10-02
Add loading icon to `quick-comment-hiding`

### LINKED ISSUE #4846: `quick-comment-hiding`’s menu does not close immediately after clicking
### Please ensure:

### Description

This is not a big issue, but it might show that something isn't working right. Maybe there's an extra `preventDefault` that should be dropped.

### Screenshot

https://user-images.githubusercontent.com/1402241/135714254-2a919d87-0150-4b7f-9c3d-ebb8e1a865bb.mov

### Console errors

_No response_

### Example URL

https://github.com/sindresorhus/refined-github/pull/4821

### Browser(s) used

Safari

kidonng:
Resolve #4846

Also disabled other buttons when user has clicked one, to avoid race conditions.

## Test URLs

Here?

## Screenshot

https://user-images.githubusercontent.com/44045911/135730930-eaad1c6e-5170-45f3-a719-ab22f135e63d.mov

### DIFF
--- a/source/features/quick-comment-hiding.tsx
+++ b/source/features/quick-comment-hiding.tsx
@@ -69,10 +69,30 @@ function showSubmenu(event: delegate.Event): void {
 	event.preventDefault();
 }
 
+function addLoadingIcon(event: delegate.Event): void {
+	if (event.delegateTarget.firstElementChild) {
+		return;
+	}
+
+	event.delegateTarget.append(
+		<svg width="16" height="16" viewBox="0 0 16 16" className="anim-rotate ml-2 v-align-text-bottom">
+			<circle cx="8" cy="8" r="7" stroke="currentColor" {...{'stroke-opacity': 0.25, 'stroke-width': 2, 'vector-effect': 'non-scaling-stroke'}}/>
+			<path d="M15 8a7.002 7.002 0 00-7-7" stroke="currentColor" {...{'stroke-width': 2, 'stroke-inecap': 'round', 'vector-effect': 'non-scaling-stroke'}}/>
+		</svg>,
+	);
+
+	for (const button of event.delegateTarget.parentElement!.querySelectorAll<HTMLButtonElement>('.dropdown-item')) {
+		if (button !== event.delegateTarget) {
+			button.disabled = true;
+		}
+	}
+}
+
 function init(): void {
 	// `useCapture` required to be fired before GitHub's handlers
 	delegate(document, '.js-comment-hide-button', 'click', showSubmenu, true);
 	delegate(document, '.rgh-quick-comment-hiding-details', 'toggle', resetDropdowns, true);
+	delegate(document, '.rgh-quick-comment-hiding-details .dropdown-item', 'click', addLoadingIcon);
 }
 
 void features.add(__filebasename, {


source/features/quick-comment-hiding.tsx
```diff
@@ -69,10 +69,30 @@ function showSubmenu(event: delegate.Event): void {
 	event.preventDefault();
 }
 
+function addLoadingIcon(event: delegate.Event): void {
+	if (event.delegateTarget.firstElementChild) {
+		return;
+	}
+
+	event.delegateTarget.append(
+		<svg width="16" height="16" viewBox="0 0 16 16" className="anim-rotate ml-2 v-align-text-bottom">
```
kidonng, line 78: This icon is not present in Octicons.

Most markup can be `cloneNode` from somewhere, but I'm not sure about that since it will still need modifications.

source/features/quick-comment-hiding.tsx
```diff
@@ -69,10 +69,30 @@ function showSubmenu(event: delegate.Event): void {
...
+			<circle cx="8" cy="8" r="7" stroke="currentColor" {...{'stroke-opacity': 0.25, 'stroke-width': 2, 'vector-effect': 'non-scaling-stroke'}}/>
+			<path d="M15 8a7.002 7.002 0 00-7-7" stroke="currentColor" {...{'stroke-width': 2, 'stroke-inecap': 'round', 'vector-effect': 'non-scaling-stroke'}}/>
+		</svg>,
+	);
+
+	for (const button of event.delegateTarget.parentElement!.querySelectorAll<HTMLButtonElement>('.dropdown-item')) {
+		if (button !== event.delegateTarget) {
```
kidonng, line 85: Can't disable the triggering button because it will prevent the form from submitting. Better suggestions welcome.

fregante:
Interesting solution. We already have a loading icon component that can be reused and maybe updated

kidonng:
> We already have a loading icon component that can be reused and maybe updated 

I will find it if it lies in `utils` instead of some random feature😃

yakov116:
https://github.com/sindresorhus/refined-github/blob/main/source/github-helpers/toast.tsx

fregante:
> [https://github.com/sindresorhus/refined-github/blob/main/source/github-helpers/toast.tsx](https://github.com/sindresorhus/refined-github/blob/main/source/github-helpers/toast.tsx?rgh-link-date=2021-10-03T14%3A17%3A56Z)

It's a bit much, the wait is just in GitHub’s UI, we're not handling anything other than clicking a button.

fregante:
> EDIT: found it in [`source/github-helpers/icon-loading.tsx`](https://github.com/sindresorhus/refined-github/blob/f9d8b7a5a402e19b0cb16fd937e0f19c43a6e699/source/github-helpers/icon-loading.tsx), but it is just an image?

👇 

> and maybe updated



You can replace it with your animation

fregante:
I like both this and https://github.com/sindresorhus/refined-github/pull/4853. The latter feels more native; This one is a bit more refined, but makes you wait (the modal won't close until it's done or until you click anyway)

kidonng:
#4853 feels wrong for me, because _nothing_ happened after the drop-down is closed. That is even worse than just keep the drop-down opening.

fregante:
But that's standard for GitHub and generally the internet. You click and then wait. If we want to see something happening we can make the comment transparent while we wait for GitHub to replace it. This kinda happens when editing comments.

fregante:
I merged https://github.com/refined-github/refined-github/pull/4853 for the time being. the dimming can be added separately as an enhancement to the feature.

Feel free to also replace loading-icon.tsx

PR #4864, opened 2021-10-04
Drop fallback code for old GitHub Enterprise versions

### LINKED ISSUE #4863: CI is failing
WIP

fregante:
- Closes #4863

### DIFF
--- a/source/features/center-reactions-popup.css
+++ b/source/features/center-reactions-popup.css
@@ -1,5 +1,4 @@
 /* Hide reaction popover text */
-.js-add-reaction-popover .text-gray, /* GHE #4021 */
 .js-add-reaction-popover .color-text-secondary,
 .js-add-reaction-popover .dropdown-divider {
 	display: none;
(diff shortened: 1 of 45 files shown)


source/features/center-reactions-popup.css
```diff
@@ -55,5 +54,5 @@ td.js-line-comments:last-child .timeline-comment-actions .add-reaction-popover::
 /* Keep colored button when reaction popup is open */
 .dropdown-details[open] .timeline-comment-action {
 	opacity: 100%;
-	color: var(--color-mktg-btn-bg, #4078c0); /* GHE #4121 */
```
fregante, line 58: I also dropped every such comment. The idea was that fallback colors were just for GHE, but in reality GH keeps changing the class names…

cc @kidonng the class checker might not be too difficult to create:

1. Find `var([^,]+` in our repo
2. Fetch GitHub’s CSS file (if static)
3. Look for each variable

yakov116 (approved):
Untested but LGTM. I have a local branch with some more stuff but don't have any time to do it.

yakov116:
I would not mark this as meta since old GHE users are affected.

fregante:
Good idea

Issue #4801, opened 2021-09-20
Renamed features appear as "New" in the options


fregante:
I just noticed that renamed features are being shown as NEW because the renames are just stored as migrations. We should move these renames to an object somewhere and then use the object in the migrations, in the options, `rgh-feature-descriptions`, etc — permanently.

<img width="774" alt="Screen Shot 2" src="https://user-images.githubusercontent.com/1402241/134031117-3a372b8d-8181-4ced-bd91-00dc51913428.png">

PR #4865, opened 2021-10-04
Rename `tag-changelog-link` to `tag-changes-link`

### LINKED ISSUE #4860: Rename `tag-changelog-link` to `tag-changes-link`
### Description

First, technically this feature doesn't provide a "changelog" (which IMO should be more like a list of changes, simpler than the compare page).

Second, it looks confusing when `link-to-changelog-file` is also present:

Screen Shot 2021-10-03 at 21 24 08

I suggest rename "changelog" part to "changes", which suits the feature more.

---

For posterity: `link-to-changelog-file` comes almost two years after, I guess it's an oversight when naming `tag-changelog-link`.

### Screenshot

_No response_

### Example URL

https://github.com/kovidgoyal/kitty/releases

kidonng:
Resolve #4860

## Test URLs

https://github.com/kovidgoyal/kitty/releases

## Screenshot
Screen Shot 2021-10-03 at 21 32 27

Screen Shot 2021-10-03 at 21 32 15


cheap-glitch:
Your screenshot shows the old button text :wink:

kidonng:
This is not changing the button text, which is correct. It's that "Changes" text. I suppose I will just post old screenshots and circle them for clarity...

fregante:
I'm hesitant to rename more features due to #4801 😅

kidonng:
That issue _should_ be easy to fix, I will try it in this PR.

kidonng:
@fregante What do you think the migration fix?

fregante:
It might be ok for now, but ideally it would be the opposite:

- have a regular, static object/map with the renames
- use the object to create all migrations at once
- use the object for the feature linkifier

PR #4875, opened 2021-10-06
`highlight-deleted-and-added-files-in-diffs`: Fix extra underline

### LINKED ISSUE #4871: `highlight-deleted-and-added-files-in-diffs` causes copy button to show underline on hover
### Please ensure:

### Description

The copy button after filenames is a link (`.Link--onHover`), but somehow it doesn't show an underline on hover (I can't find the corresponding CSS). However, if you put anything _after_ it (like `highlight-deleted-and-added-files-in-diffs`), it will show a ugly underline on hover:

image

### Screenshot

_No response_

### Console errors

_No response_

### Example URL

https://github.com/refined-github/refined-github/commit/4d7db0859baa6a79ebf3821e8974a100ab5e237d

### Browser(s) used

Chrome

ritikbheda:
Resolve #4871

## Test URLs

[4d7db](https://github.com/refined-github/refined-github/commit/4d7db0859baa6a79ebf3821e8974a100ab5e237d)

## Screenshot
hover is now working as expected
screenshot

### DIFF
--- a/source/refined-github.css
+++ b/source/refined-github.css
@@ -160,4 +160,4 @@ pr-branches
 /* Remove the underline on PR filename copy button hover #4871 */
 .file-header .file-info clipboard-copy {
 	display: inline-block;
-} 
+}


source/refined-github.css
```diff
@@ -156,3 +156,8 @@ pr-branches
 	margin-left: 4px !important;
 	margin-right: 4px !important;
 }
+
+/* removes the underline for whitespace on hover, working for issue #4871*/
+.Link--onHover:hover{
```
fregante, line 161: The selector needs to be made more specific to the item we're targeting. I'm not sure if it's possible to target it exactly. Maybe with sibling selectors.

ritikbheda: yeah, I thought so too. 
I think it can be achieve by going to exact that className from its parent tag's classes. 
A potential solution is this
```
.file-header .file-info .Link--onHover:hover {
	text-decoration: none;
}
```

kidonng: It's better to target `clipboard-copy` instead of more generic `.Link--onHover`.

ritikbheda:
Hi everyone, I would really appreciate if you please add the `hacktoberfest` and/or `hacktoberfest-accepted` labels to this PR.

I am not even sure if you can add label after closing the PR or no.

PR #4876, opened 2021-10-06
Fix missing spaces in `show-names`

cheap-glitch:
Fix bug introduced by #4851 
Also fixes the feature on the dashboard

## Test URLs
* [Conversation](https://github.com/refined-github/refined-github/pull/4851)
* [Dashboard](https://github.com)

## Screenshot
_Conversation (before)_
image

_Conversation (after)_
image

_Dashboard_
https://user-images.githubusercontent.com/46634000/136196560-6db174a4-07b8-4d4d-94f3-d1a29c4b1ba3.mp4

### DIFF
--- a/source/features/show-names.tsx
+++ b/source/features/show-names.tsx
@@ -1,6 +1,7 @@
 import './show-names.css';
 import React from 'dom-chef';
 import select from 'select-dom';
+import onetime from 'onetime';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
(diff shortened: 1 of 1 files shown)


source/features/show-names.tsx
```diff
@@ -74,7 +73,8 @@ void features.add(__filebasename, {
 	additionalListeners: [
 		onNewsfeedLoad,
 	],
-	init: onetime(init),
+	onlyAdditionalListeners: true,
```
cheap-glitch, line 76: Seems the events on the dashboard are always loaded asynchronously now.

fregante: Yeah maybe we can skip this change for now. Maybe add `// TODO: Use onlyAdditionalListeners in February 2022 #4876 `

PR #4886, opened 2021-10-07
Drop `cancel-wiki-edit-button` feature

fregante:
- Context: https://github.com/refined-github/refined-github/issues/4881#issuecomment-937837852
- Related: https://github.com/refined-github/refined-github/issues/4460


fregante:
Comments also have a "Close with comment" button, is that confusing? Should we add a "Cancel" here too? I think not.

Cancel makes sense when here's a process active, for example you're editing a comment and the only way to "go back" is to either save the edits or discard them.

PR #4896, opened 2021-10-09
Restore `restore-file` on forks

fregante:
Reverts refined-github/refined-github#4803 in whole

It doesn't work in forks 🤦‍♂️ 🤦‍♂️ 🤦‍♂️ 🤦‍♂️ 🤦‍♂️ 🤦‍♂️ 

Yet another half-assed feature by GitHub. Sad.


yakov116:
This is so sad. This is why V4 gets me mad! They do a half a job and them never finish it.

PR #4918, opened 2021-10-14
Meta: Refactor custom tab highlighting

cheap-glitch:
Follow-up of https://github.com/refined-github/refined-github/pull/4914#discussion_r727301780 and https://github.com/refined-github/refined-github/pull/4914#discussion_r727882502

## Test URLs
Navigating between tabs in any repo

## Screenshot
https://user-images.githubusercontent.com/46634000/137315048-fe71dad5-f0bb-4034-810f-1556bf445c9d.mp4


fregante:
This could be simplified, for example instead of using events, you can just call a function to re-update the nav bar. 

We could use an `init` with `deduplicate: false` on both features that looks like:

```js
function init () {
	if (shouldBugsTabHaveFocus) {
		// Remove focus from other tab
	}

	// Ensure that the bugs tab is consistent whether it needs it or not
	bugsTab.classList.toggle('current', shouldBugsTabHaveFocus);
	bugsTab.attributes.toggle('aria-current', shouldBugsTabHaveFocus); // Or whatever
}
```

`shouldBugsTabHaveFocus` would be calculated every time `init` is run and/or it could be memoized via `mem` to allow it to be called from the other `init` as well without doing calculations twice.

```ts
const shouldBugsTabHaveFocus = mem((url: string) => {
	///logic
})
```

fregante:
> the problem is ensuring that the `bugsTab` element has actually been added by the "main" `init` before trying to (un)highlight it.


```js
append(bugsTab);
secondInit() // Call it manually, even if it's duplicate right now
```

> A simpler solution could be to have a single un-deduplicated init that would:

That'd also work. As long as the pattern is preserved for `releases-tab`

fregante:
Up to you

cheap-glitch:
@refined-github/maintainers can someone else test this too, ideally in a browser other than Firefox? Thanks!

yakov116:
I can Later tonight

yakov116:
Something is going wrong with my webpack...

```
The system cannot find the path specified.
node:events:368
 event
      ^

Error: spawn ts-node\esm ENOENT     
    at notFoundError (C:\Users\yakov\Documents\GitHub\refined-github\node_modules\cross-spawn\lib\enoent.js:6:26)
    at verifyENOENT (C:\Users\yakov\Documents\GitHub\refined-github\node_modules\cross-spawn\lib\enoent.js:40:16)
    at ChildProcess.cp.emit (C:\Users\yakov\Documents\GitHub\refined-github\node_modules\cross-spawn\lib\enoent.js:27:25)
    at Process.ChildProcess._handle.onexit (node:internal/child_process:290:12)
Emitted 'error' event on ChildProcess instance at:
    at ChildProcess.cp.emit (C:\Users\yakov\Documents\GitHub\refined-github\node_modules\cross-spawn\lib\enoent.js:30:37)
    at Process.ChildProcess._handle.onexit (node:internal/child_process:290:12) {
  code: 'ENOENT',
  errno: 'ENOENT',
  syscall: 'spawn ts-node\\esm',    
  path: 'ts-node\\esm',
  spawnargs: [ 'webpack', '--mode=development', '--watch' ]
}
[11:04:04 PM] Starting compilation in watch mode...

[11:04:17 PM] Found 0 errors. Watching for file changes.
```

fregante:
> Something is going wrong with my webpack...

One more reason for #5014  😜 

Maybe a reinstall will do

yakov116:
I tried, maybe I will have some time later tonight

PR #4928, opened 2021-10-16
Lint

fregante:


cheap-glitch:
Could we also update `webext-base-css` while we're at it? I'd really like to get rid of the black inputs in Firefox's options page.

PR #4939, opened 2021-10-18
Fix console error in `quick-label-removal`

### LINKED ISSUE #4934: Console error in `quick-label-removal`
### Please ensure:

### Include in this issue:

- [X] Console errors, if any

### Description

Removing a label using the feature produces the following error in the console:
```
TypeError: deferredContentWrapper is undefined quick-label-removal.tsx:31:15
```

### Example URL

https://github.com/refined-github/github-sandbox/issues/3

### Browser(s) used

Firefox

cheap-glitch:
Fixes #4934 

## Test URLs
https://github.com/refined-github/github-sandbox/issues/3

## Screenshot
n/a

### DIFF
--- a/source/features/quick-label-removal.tsx
+++ b/source/features/quick-label-removal.tsx
@@ -24,15 +24,6 @@ async function removeLabelButtonClickHandler(event: delegate.Event<MouseEvent, H
 	});
 
 	removeLabelButton.closest('a')!.remove();
-
-	// Force update of label selector if necessary
-	const deferredContentWrapper = select('.label-select-menu [src] .hx_rsm-content');
-	if (deferredContentWrapper) {
-		const menu = deferredContentWrapper.closest('[src]')!;
-		deferredContentWrapper.replaceChildren(
-			<include-fragment src={menu.getAttribute('src')!}/>,
-		);
-	}
 }
 
 async function init(): Promise<void> {


fregante:
> it seems GitHub does update the labels dropdown correctly on its own

Let's drop it if that's the case. 



> Maybe keep it for GHE?

Not a fundamental feature, so we can drop it

source/features/quick-label-removal.tsx
```diff
@@ -24,14 +24,6 @@ async function removeLabelButtonClickHandler(event: delegate.Event<MouseEvent, H
...
 
 	removeLabelButton.closest('a')!.remove();
-
-	// Force update of label selector if necessary
-	if (!select.exists('.sidebar-labels include-fragment')) {
-		const deferredContentWrapper = select('.sidebar-labels .hx_rsm-content')!;
-		const menu = deferredContentWrapper.closest('[src]')!;
-		deferredContentWrapper.textContent = '';
-		deferredContentWrapper.append(<include-fragment src={menu.getAttribute('src')!}/>);
-	}
```
fregante, line 34: I think this is still beneficial.

In https://github.com/refined-github/refined-github/issues/4953 I removed a label and then opened the menu to add one. The label was still selected in there and it added the label again.

Maybe a "simpler" or safer alternative would be to look for the specific label in the dropdown and disable it there as well, to avoid this specific issue

cheap-glitch: I noticed that bug too.

The fix is tricky, because immediately replacing the dropdown content with an include fragment doesn't always work, especially when removing several labels in quick succession and then opening the dropdown right away.

Maybe we should add a listener on the dropdown button? The logic would be:
 1. When a label is removed, mark the dropdown as "outdated" but don't actually force it to update
 2. Listen to the opening of the dropdown. If it's outdated and GitHub hasn't already updated its contents (i.e. no include fragments is already there), replace them with an `<include-fragment>`.

> Maybe a "simpler" or safer alternative would be to look for the specific label in the dropdown and disable it there as well, to avoid this specific issue

Changing the `aria-selected` attribute of a label in the dropdown has zero effect, it's purely cosmetic.

fregante: I don't remember how it works, but doesn't GitHub already use a "data-url" attribute that is then loaded as a fragment when the user clicks it? If it is, then we just need to set it again and empty the dropdown. GitHub will then take care of loading it as usual. 

I don't want something super complicated though 😬 I think this would be good if it works

PR #4946, opened 2021-10-20
Show toast when API fails

### LINKED ISSUE #4828: API actions might fail silently
### Please ensure:

### Description

On private repositories that are part of a private GitHub org using SAML auth, some features do not work if the token has not been authorized.

This is expected, as there is no workaround other than giving the token access. However, ideally the error would be gracefully handled and feedback given to the user.

For example, when attempting to use the `x` to quickly remove a label from an issue, the button simply appears to do nothing. Opening the Console reveals that the

```http
DELETE /repos/ORG/REPO/issues/NUMBER/labels/LABEL
```

request failed with an unhandled `403` response:

```
Uncaught (in promise) Error: Unable to fetch.
Ensure that your token has access to this repo.
{
	"message": "Resource protected by organization SAML enforcement. You must grant your Personal Access token access to this organization.",
	"documentation_url": "https://docs.github.com/articles/authenticating-to-a-github-organization-with-saml-single-sign-on/"
}
```

Ideally,

- either the features that wouldn't work would just not load
  - and maybe a message would be shown explaining why
- or the user would be notified of the error in some way, with steps to resolve it (as provided by GitHub's response)

### Screenshot

_No response_

### Console errors

```shell
VM256 refined-github.js:2930 DELETE https://api.github.com/repos/ORG/REPO/issues/12345/labels/LABEL 403
mem.cacheKey @ VM256 refined-github.js:2930
async function (async)
mem.cacheKey @ VM256 refined-github.js:2926
(anonymous) @ VM256 refined-github.js:2857
removeLabelButtonClickHandler @ VM256 refined-github.js:8507
listenerFn @ VM256 refined-github.js:1682
VM256 refined-github.js:2906 Uncaught (in promise) Error: Unable to fetch.
Ensure that your token has access to this repo.
{
	"message": "Resource protected by organization SAML enforcement. You must grant your Personal Access token access to this organization.",
	"documentation_url": "https://docs.github.com/articles/authenticating-to-a-github-organization-with-saml-single-sign-on/"
}
    at getError (VM256 refined-github.js:2977)
    at async mem.cacheKey (VM256 refined-github.js:2949)
    at async HTMLHtmlElement.removeLabelButtonClickHandler (VM256 refined-github.js:8507)
```

### Example URL

https://github.com/ORG/REPO/issues/12345

### Browser(s) used

Chrome on macOS

yakov116:
Closes #4828

## Test URLs
I made my token invalid 

## Screenshot
Video_2021-10-20_000006


source/features/index.tsx
```diff
@@ -51,6 +52,7 @@ const logError = (id: FeatureID, error: unknown): void => {
 	const message = error instanceof Error ? error.message : String(error);
 
 	if (message.includes('token')) {
+		void showToast(async () => Promise.reject(new Error(message)));
```
fregante, line 55: Not necessary, but it'd be nice to use it as:

```suggestion
		void showToast(new Error(message));
```

and also the regular
```ts
		void showToast("My DONE notification");
```

in addition to the current situation

yakov116:
I will continue this at the end of the week. (Almost done moving!)

fregante (changes requested):
Like the original issue’s title suggest, the toast should only appear on _actions_, not every _call_.

We generally prefer features to fail silently rather than show annoying popups. The toast should only ever be used to inform the user of the progress of an action that they requested, not regular issues, otherwise navigating GitHub might just get you a bunch of toasts even if you're not hungry.

In short we should just wrap these specific functions in a regular `showToast`. In this case:


```diff
+	await showToast(() =>
		api.v3(`issues/${getConversationNumber()!}/labels/${removeLabelButton.dataset.name!}`, {
			method: 'DELETE',
		})
+	);
```

PR #4961, opened 2021-10-21
Meta: Add support for style hotfixes

### LINKED ISSUE #4675: Add hotfix.css
https://github.com/sindresorhus/refined-github/issues/4674 finally pushed me to open this request. I've been thinking about it again just this week.

I feel that occasionally we have breaking styles which might not be fixable by just disabling a feature or that could be _hotfixed_ without forcing another release.

I'm thinking about adding a version-specific CSS file that can be injected before the user stylesheet, living at _`hotfix`/style/21.8.14.css._

cheap-glitch:
Resolves #4675

This is a tentative implementation, feel free to close this PR if it's not the right direction.

### DIFF
--- a/style/0.0.1.css
+++ b/style/0.0.1.css
@@ -1,4 +1,0 @@
-* {
-	color: gold !important;
-	background-color: gold !important;
-}


source/helpers/hotfix.ts
```diff
@@ -22,6 +24,21 @@ export const updateHotfixes = cache.function(async (): Promise<string[][]> => {
...
+	const {content} = await request.json();
+
+	if (!content) {
+		return '';
+	}
+
+	return atob(content).trim();
+}, {
+	maxAge: {hours: 6},
+	cacheKey: ([version]) => 'style-hotfixes:' + version,
```
fregante, line 39: No need to use the version, please match the existing hotfix usage. Also the style hotfix should be cleared when a new version is installed (like the regular hotfixes)

PR #4969, opened 2021-10-23
Meta: Update `ts-node`

cheap-glitch:
Follow-up of https://github.com/refined-github/refined-github/commit/fb7726a2335204d1cd33721b32b6e7b284c7397b#commitcomment-58518052


yakov116:
I tried about 10 times.

The only way I got it to work was to remove module from the package.json

But then AVA failed..

PR #4987, opened 2021-10-25
Meta: Disable checkboxes of hotfixed features

### LINKED ISSUE #4974: Hotfix-disabled features should display as disabled
### Description

Screen Shot 2021-10-24 at 23 41 03

The checkbox is checked and enabled, despite the notice above says it is disabled. It doesn't reflect its actual state.

We should:

- _Display_ the checkbox as **unchecked** (it doesn't affect the _actual_ option state)
- Display the checkbox as **disabled** (since changing the state doesn't matter)
  We used to do this but it was lost in https://github.com/refined-github/refined-github/pull/3678/files#diff-a24f014d7cd2bf18b76e05a5879d46687e5ebdebee94fb2632d724b63ae41464L45.

### Screenshot

_No response_

### Example URL

chrome://extensions/?options=hlepfoohegkhhmjieoechaddaejaokhf

kidonng:
Resolve #4974

## Test URLs

Note: to test it you need to change `version` to `21.10.18` in `manifest.json`.

chrome://extensions/?options=balgliakcbamepammkbcjdepogihccmj

## Screenshot

Before:

image

After:

image

### DIFF
--- a/source/options.tsx
+++ b/source/options.tsx
@@ -208,6 +208,10 @@ async function getLocalHotfixesAsNotice(): Promise<HTMLElement> {
 			disabledFeatures.append(
 				<p><code>{feature}</code> has been temporarily disabled due to {createRghIssueLink(relatedIssue)}.</p>,
 			);
+
+			const input = select<HTMLInputElement>(`#${feature}`)!;
+			input.disabled = true;
+			input.checked = false;
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


yakov116:
> I mean _we_ don't have a reason to. Just because the feature is disabled it doesn't mean we should also stop the user from toggling it for when it's fixed.

The only situation I can see a user needing to toggle is if they are transferring their setting from one computer to the next.

I happen to not see an issue with the whole line being disabled.

JMHO

kidonng:
If the user scrolls down and didn't notice/remebr the hotfix message, they may start to wonder why toggling those features doesn't work.

Yeah I know this is some hypothetical situation but AFAICT the checkboxes had been disabled for the longest time. It's just the behavior isn't keep in the refactor PR.

fregante (approved):

fregante:
Let's release a new version on the 10th. Feel free to merge any approved PRs or review others’

fregante (changes requested):
Changed my mind again. All your arguments still apply to the current situation, mainly:

> If the user scrolls down and doesn't notice/remember the hotfix message, they may wonder why toggling those features doesn't work.

Same exact thing applies now; additionally the checkbox does not respond to clicks at all, i.e. the checkbox itself broken.

I see 2 improvements possible, if you want to pick this up:

- dim the whole line as suggested earlier, which looks nicer than a single  unclickable checkbox
- add the reason why it's dimmed inline, like it was before

fregante:
Closing for now, it can be opened when there are commits

PR #4996, opened 2021-10-27
Drop `monospace-textareas` feature

### LINKED ISSUE #4917: Maybe remove `monospace-textareas`?
### Description

GitHub has added a setting that does this:

https://github.blog/changelog/2021-10-12-fixed-width-font-support-in-markdown-enabled-fields/

### Screenshot

_No response_

### Example URL

Anywhere textareas are shown

cheap-glitch:
Closes #4917 

## Test URLs
This page

## Screenshot

It seems GitHub's default font size is larger than RGH's. I personally don't mind since my screen is both small and hi-res, but others might disagree.

**Before**
screenshot-1635358264

**After**
screenshot-1635358234


source/refined-github.css
```diff
@@ -156,3 +156,10 @@ pr-branches
 .file-header .file-info clipboard-copy {
 	display: inline-block;
 }
+
+/* Limit width of commit title to 72 characters */
+.rgh-monospace-textareas :is(#merge_title_field, #commit-summary-input) {
```
kidonng, line 161: My bad, I didn't notice the review request.

This leading `.rgh-monospace-textareas` should be removed since no feature is adding the class anymore.

levrik:
I used to have this feature disabled in my settings. Since the removal of the toggle I'm seeing monospace font being used for the commit message title with no way of turning off anymore. I wonder why the feature toggle was removed? There are people that don't want monospace font 😅

cheap-glitch:
> I personally also like this feature but not the monospace font so coupling this functionality with the monospace font would still require the custom CSS.

We can't support each and everyone's preferences, so that's (in part) what the custom CSS field is here for :smiley:

levrik:
@cheap-glitch I understand this but the monospace font should at least be bound to the `suggest-commit-title-limit` feature. Because currently this extension introduces a change which can't be turned off via options at all.

cheap-glitch:
> the monospace font should at least be bound to the suggest-commit-title-limit feature

I agree that it makes sense, and I don't mind either way. @fregante wdyt?

> Because currently this extension introduces a change which can't be turned off via options at all.

There are a lot of those actually, see https://github.com/refined-github/refined-github/blob/main/source/refined-github.css and https://github.com/refined-github/refined-github/blob/main/source/features/github-bugs.css. All of which can be "turned off" via custom CSS — granted, some effort and source-reading are needed.

levrik:
@cheap-glitch I see. This was the first change I've noticed which doesn't have an option. Thought everything else is having an option.

kidonng:
> I wonder why the feature toggle was removed but the CSS stayed?

The CSS didn't "stay", it was "restored". I personally also feel the monospace font looks out of place in the commit message field, however `suggest-commit-title-limit` is more valuable to me.

PR #4998, opened 2021-10-27
Add indicator that a workflow can be triggered manually

### LINKED ISSUE #4558: Add indicator that a workflow can be triggered manually in sidebar
Example: https://github.com/fregante/github-url-detection/actions/workflows/npm-publish.yml

Screen Shot 9

We could show a rocket or play icon in the sidebar to show that a workflow is actionable.

`next-scheduled-github-action` acts on exactly the same spot, so it _might_ make sense to rename the feature and extend it, also to avoid possible race conditions.

cheap-glitch:
Resolves #4558 
 
## Test URLs
 * https://github.com/fregante/github-url-detection/actions/workflows/npm-publish.yml
 * https://github.com/refined-github/refined-github/actions

## Screenshots

image
-----
image


fregante:
I just noticed that this is storing the whole workflow. That is unnecessary. It should parse the workflows and only store the cronjob and dispatch flag

fregante:
Looks great! 🤩 

Screen Shot 1

PR #5016, opened 2021-10-30
Split `hide-own-stars` feature into forks and stars

jmir1:
The hide-own-stars feature currently hides "starred" AND "forked" events for your own repos.

This PR splits the hide-own-stars feature into two settings for star and fork events.

This is beneficial if you are interested in who forks your repos because they might be contributing something useful in their forks and you want to check out their progress. (just an example)

Also, the feature description currently doesn't mention that it also hides "forked" events.

## Test URLs

affects the main newsfeed (https://github.com/)


fregante:
Since our intent is to hide both, there's no reason to split the feature and then keep hiding both. The only issue I see is that the name/description doesn't match the feature, so _that_ can be updated.

fregante:
> This is beneficial if you are interested in who forks your repos because they might be contributing something useful in their forks and you want to check out their progress. (just an example)

That's what PRs are for, you're already about to get a notification. Many people will fork your repos just to keep a copy, so the intent is often the same as starring.

PR #5017, opened 2021-10-30
Improve usability of `quick-review-buttons`

### LINKED ISSUE #5010: The default submit action of PR review form isn't "Approve"
### Please ensure:

### Include in this issue:

Screenshot from 2021-10-29 15-18-34

- [ ] Console errors, if any

### Description

The green "Approve" button in the PR review dropdown form visually looks like the primary call to action. However, if I submit the form with ctrl+enter it uses the "Comment" button instead.

It happened to me so many times that I made a "Comment" instead of "Approve" that I think the UI is just wrong.

I can imagine two solutions:
- Change the visual look of the buttons, so that "Comment" looks like the primary call to action
- Change the form default submit action (used when ctrl+enter) to be "Approve"

### Example URL

### Browser(s) used

Google Chrome Version 95.0.4638.54 (Official Build) (64-bit)

fregante:
- Fixes #5010 

## Test URL

This PR or any open PR that you _did not_ open

## Before
Screen Shot 4

## After
Screen Shot 5

### DIFF
--- a/source/features/quick-review-buttons.tsx
+++ b/source/features/quick-review-buttons.tsx
@@ -53,11 +53,9 @@ function init(): false | void {
 			</button>
 		);
 
-		if (radio.value === 'approve') {
+		if (!radio.disabled && radio.value === 'approve') {
 			button.prepend(<CheckIcon className="color-fg-success"/>);
-		}
-
-		if (radio.value === 'reject') {
+		} else if (!radio.disabled && radio.value === 'reject') {
 			button.prepend(<FileDiffIcon className="color-fg-danger"/>);
 		}
 
--- a/source/features/quick-review-buttons.tsx
+++ b/source/features/quick-review-buttons.tsx
@@ -2,6 +2,7 @@ import React from 'dom-chef';
 import select from 'select-dom';
 import delegate from 'delegate-it';
 import * as pageDetect from 'github-url-detection';
+import {CheckIcon, FileDiffIcon} from '@primer/octicons-react';
 
 import features from '.';
 import looseParseInt from '../helpers/loose-parse-int';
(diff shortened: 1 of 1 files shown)


fregante:
cc @slaweet

yakov116 (approved):
I happen to like this!

I don't see the point of removing the red from the "requested changes"

cheap-glitch:
> I don't see the point of removing the red from the "requested changes"

It's to make it seems less like a "destructive" action, which it is not.

Although I quite liked the clear color separation between the three buttons, maybe we could keep just a red border around "Request changes"?

kidonng:
> Although I quite liked the clear color separation between the three buttons, maybe we could keep just a red border around "Request changes"?

I prefer not, because it doesn't fit in with other buttons and more importantly, it could still imply wrong info.

slaweet:
> Screen Shot 2021-10-31 at 16 23 16

I like this :point_up: :+1: 

One further change I would consider, is to leave only the "Comment" button on the right and move the other two buttons to the left. So that it would be similar to PR comment form:
Screenshot from 2021-10-31 13-41-31

kidonng:
How it currently looks:

Screen Shot 2021-10-31 at 21 37 23

cheap-glitch:
> One further change I would consider, is to leave only the "Comment" button on the right and move the other two buttons to the left. So that it would be similar to PR comment form

I'm -1 on this, it doesn't make much sense to me. Buttons on the left are either for "closing" or "cancelling" actions, none of which are present here.

kidonng:
I changed my mind as well. Note that there can be other stuff on the left:

Screen Shot 2021-10-31 at 21 59 36

kidonng:
Oops, I didn't know other people are editing titles as well 😅

fregante:
Last commit:

Screen Shot 15
Screen Shot 16

fregante (commented):
Good to got me

fregante (commented):
Test

Edit: I found a bug in Safari… cmd+enter does not submit it. It works in Chrome (this review). Unrelated to this PR though

kidonng:
We can keep the icon, just don't add color classes:
Screen Shot 2021-10-31 at 22 29 00

yakov116 (approved):
Untested but I like it

fregante:
I'm not a super fan of the icons already, I think it's easier to skip them if the buttons aren't clickable, since the visual differentiation isn't useful.

kidonng (approved):

PR #5021, opened 2021-10-31
Use new background color variables

### LINKED ISSUE #5018: `clean-dashboard` makes background transparent
### Please ensure:

### Include in this issue:

- [ ] Console errors, if any

### Description

image

### Example URL

https://github.com/

### Browser(s) used

Chrome

fregante:
Let's merge this by 21.11.1

- Fixes #5018 

## Test URLs

https://github.com/

## Screenshot

Screen Shot 6

### DIFF
--- a/source/refined-github.css
+++ b/source/refined-github.css
@@ -2,7 +2,7 @@
 	--rgh-green: var(--color-fg-success, #1a7f37);
 	--rgh-red: var(--color-fg-danger, #cf222e);
 	--rgh-border-color: var(--color-border-muted, #d8dee4);
-	--rgh-background: var(--color-canvas-default, var(--color-bg-default, var(--color-bg-canvas, #fff)));
+	--rgh-background: var(--color-canvas-default, var(--color-bg-canvas, #fff));
 }
 
 /* Make tab indented code more readable on GitHub and Gist */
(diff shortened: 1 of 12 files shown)


source/refined-github.css
```diff
@@ -2,6 +2,7 @@
 	--github-green: var(--color-fg-success, #1a7f37);
 	--github-red: var(--color-fg-danger, #cf222e);
 	--github-border-color: var(--color-border-muted, #d8dee4);
+	--rgh-background: var(--color-canvas-default, var(--color-bg-default, var(--color-bg-canvas, #fff)));
```
fregante, line 5: We should rename the other variables in a future lint PR to use `rgh` instead

source/refined-github.css
```diff
@@ -1,7 +1,8 @@
 :root {
-	--github-green: var(--color-fg-success, #1a7f37);
-	--github-red: var(--color-fg-danger, #cf222e);
-	--github-border-color: var(--color-border-muted, #d8dee4);
+	--rgh-green: var(--color-fg-success, #1a7f37);
+	--rgh-red: var(--color-fg-danger, #cf222e);
+	--rgh-border-color: var(--color-border-muted, #d8dee4);
+	--rgh-background: var(--color-canvas-default, var(--color-bg-default, var(--color-bg-canvas, #fff)));
```
cheap-glitch, line 5: Actually we could just drop `--color-bg-default` because this var probably never existed.

It seems some variables match the new class names, but some don't, e.g. `.color-fg-success` but `--color-success-fg` :man_shrugging:

cheap-glitch (approved):

PR #5028, opened 2021-10-31
Prevent `quick-file-edit` links from being AJAXed

### LINKED ISSUE #4638: `quick-file-edit` links should not be AJAXed
### Please ensure:

### Description

Some links are AJAXed but GitHub returns the whole page, so the page will break immediately/once you make another navigation:

- ~~`latest-tag-button`: https://github.com/sindresorhus/refined-github/pull/4631#discussion_r682548229~~

  **EDIT**: no longer a bug, see https://github.com/sindresorhus/refined-github/pull/4641#issuecomment-901225714

- `edit-files-faster`: 

  https://user-images.githubusercontent.com/44045911/128181902-0a128898-53f5-4890-8f43-a2081501667b.mov

  This is harder to fix because we need to find a way to disable the AJAX behavior

### Screenshot

_No response_

### Console errors

_No response_

### Example URL

### Browser(s) used

Chrome

kidonng:
Fix #4638

## Test URLs

https://github.com/refined-github/github-sandbox

## Screenshot

N/A

### DIFF
--- a/source/features/quick-file-edit.tsx
+++ b/source/features/quick-file-edit.tsx
@@ -8,17 +8,8 @@ import {wrap} from '../helpers/dom-utils';
 import features from '.';
 import GitHubURL from '../github-helpers/github-url';
 import {isPermalink} from '../github-helpers';
-import isModifierKeys from '../helpers/is-modifier-keys';
 import getDefaultBranch from '../github-helpers/get-default-branch';
 import onFileListUpdate from '../github-events/on-file-list-update';
-
-function onClick(event: React.MouseEvent): void {
-	if (isModifierKeys(event)) {
-		return;
-	}
-
-	(event.target as HTMLElement).closest('.js-navigation-container')!.removeAttribute('data-pjax');
-}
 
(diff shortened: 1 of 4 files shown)


fregante:
This doesn't look right, there must be a "no Ajax" attribute

yakov116 (approved):
Looks Right

kidonng:
> This doesn't look right, there **must** be a "no Ajax" attribute

Hmm this is just an idea I came out when I was doing other stuff, I didn't dig into GitHub's code. But here it is:

```ts
on('click', '[data-pjax] a, a[data-pjax]', function (event) {
  const link = event.currentTarget
  if (link instanceof HTMLAnchorElement) {
    if (link.getAttribute('data-skip-pjax') != null) {
      return
    }
    if (link.getAttribute('data-remote') != null) {
      return
    }

    const container = detectContainer(link)
    if (container) {
      click(event, {
        container,
        scrollTo: preserveScrollTo(link)
      })
    }
  }
})
```

PR #5031, opened 2021-10-31
Hide duplicate `quick-pr-diff-options` buttons on commit and compare pages

kidonng:
Tackle https://github.com/refined-github/refined-github/issues/5009#issuecomment-955777413

## Test URLs

- [`isSingleCommit`](https://github.com/refined-github/refined-github/commit/49fd6261e21d69bd62094bf0b38783a7e06020d1)
- [`isCompare`](https://github.com/refined-github/refined-github/compare/main...add-multiple-prs-dropdown)

## Screenshot

Before:

image

image

After:

image

image

### DIFF
--- a/source/features/quick-pr-diff-options.tsx
+++ b/source/features/quick-pr-diff-options.tsx
@@ -73,7 +73,7 @@ function createWhitespaceButton(): HTMLElement {
 	);
 }
 
-function initPR(): false | void {
+function init(): false | void {
 	// TODO [2022-05-01]: Remove `.js-diff-settings` from the selector (kept for GHE)
 	const originalToggle = select('.js-diff-settings, [aria-label="Diff settings"]')!.closest('details')!.parentElement!;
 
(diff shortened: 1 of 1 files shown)


fregante:
I hope you're not rushing to add bugfixes at the last minute 😃 
Sometimes they're simple and can be merged at the last minute; Others it's better to wait and test after merge for a few days before publishing them.

fregante:
In this case I don’t think it's worth to replace their buttons. I suggest just excluding this page from the feature and leave the native ones.

kidonng:
> merged at the last minute

Actually I'm not asking for this 😅

> In this case I don’t think it's worth to replace their buttons. I suggest just excluding this page from the feature and leave the native ones, that would let us simplify it a little more.

That's what I wanted to do at first, but I suppose ours are still _quicker_ and holds its value.

fregante:
> leave the native ones

I was referring to the unified/split buttons only, not the whitespace button too. There's no native whitespace button there yet

kidonng:
> > leave the native ones
> 
> I was referring to the unified/split buttons only, not the whitespace button too. There's no native whitespace button there yet

_What happens when you make PR late at night_

fregante (approved):
# 🛻

PR #5036, opened 2021-11-01
Lint

kidonng:
Another dump of small codebase improvements I encountered when poking around.

### DIFF
--- a/source/features/esc-to-deselect-line.tsx
+++ b/source/features/esc-to-deselect-line.tsx
@@ -15,7 +15,7 @@ function isLineSelected(): boolean {
 function listener({key, target}: KeyboardEvent): void {
 	if (key === 'Escape' && isLineSelected() && !isEditable(target)) {
 		location.hash = '#no-line'; // Update UI, without `scroll-to-top` behavior
-		history.replaceState({}, document.title, location.pathname); // Drop remaining # from url
+		history.replaceState(null, document.title, location.pathname); // Drop remaining # from url
 	}
 }
 
--- a/source/features/esc-to-deselect-line.tsx
+++ b/source/features/esc-to-deselect-line.tsx
@@ -15,7 +15,7 @@ function isLineSelected(): boolean {
 function listener({key, target}: KeyboardEvent): void {
 	if (key === 'Escape' && isLineSelected() && !isEditable(target)) {
 		location.hash = '#no-line'; // Update UI, without `scroll-to-top` behavior
-		history.replaceState(null, document.title, location.pathname); // Drop remaining # from url
+		history.replaceState(undefined, document.title, location.pathname); // Drop remaining # from url
 	}
 }
 
--- a/source/features/preview-hidden-comments.tsx
+++ b/source/features/preview-hidden-comments.tsx
@@ -29,9 +29,9 @@ function init(): void {
 		}
 
 		// Hidden review comments that have been preloaded have their header text wrapped in an extra <div>
-		const headerTextWrapper = header.tagName === 'H3' ? select(':scope > .d-inline-block:last-child', header) : undefined;
+		const headerTextWrapper = header.tagName === 'H3' ? select(':scope > .d-inline-block:last-child', header) : header.firstChild;
 		header.append(
-			<span className="Details-content--open">{headerTextWrapper ?? header.firstChild}</span>,
+			<span className="Details-content--open">{headerTextWrapper}</span>,
 			<span className="Details-content--closed">{`${upperCaseFirst(reason)} — ${commentText.slice(0, 100)}`}</span>,
 		);
 	}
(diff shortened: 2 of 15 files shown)


source/features/tag-changes-link.tsx
```diff
@@ -110,11 +110,12 @@ async function init(): Promise<void> {
 			'.list-style-none > .d-inline-block:last-child', // Link to source tarball under release tag
 		], container.element);
 		for (const lastLink of lastLinks) {
+			const currentTag = allTags[index].tag;
 			const compareLink = (
 				<a
 					className="Link--muted tooltipped tooltipped-n"
-					aria-label={'See changes since ' + decodeURIComponent(previousTag)}
```
kidonng, line 116: IMO this "See changes since" is really confusing, to me it sounds like comparing `previousTag` with `HEAD`.

This commit just make it more straightforward. I choose `currentTag` instead of "this tag" because the tooltip happens to cover part of the tag name, and then you have to figure it out what it refers to:

Screen Shot 2021-11-02 at 02 53 38

source/features/esc-to-deselect-line.tsx
```diff
@@ -15,7 +15,7 @@ function isLineSelected(): boolean {
 function listener({key, target}: KeyboardEvent): void {
 	if (key === 'Escape' && isLineSelected() && !isEditable(target)) {
 		location.hash = '#no-line'; // Update UI, without `scroll-to-top` behavior
-		history.replaceState({}, document.title, location.pathname); // Drop remaining # from url
+		history.replaceState(null, document.title, location.pathname); // Drop remaining # from url
```
fregante, line 18: 🤔 this shouldn't pass lint
It should be undefined

source/features/preview-hidden-comments.tsx
```diff
@@ -29,9 +29,9 @@ function init(): void {
 		}
 
 		// Hidden review comments that have been preloaded have their header text wrapped in an extra <div>
-		const headerTextWrapper = header.tagName === 'H3' ? select(':scope > .d-inline-block:last-child', header) : undefined;
+		const headerTextWrapper = header.tagName === 'H3' ? select(':scope > .d-inline-block:last-child', header) : header.firstChild;
 		header.append(
-			<span className="Details-content--open">{headerTextWrapper ?? header.firstChild}</span>,
+			<span className="Details-content--open">{headerTextWrapper}</span>,
```
fregante, line 34: Note: This isn't equivalent. The previous code worked even when the select was empty. 

Not sure if it's still fine

source/features/show-open-prs-of-forks.tsx
```diff
@@ -60,7 +60,7 @@ async function getPRs(): Promise<[prCount: number, url: string] | []> {
 		return [count, `/${forkedRepo}/pull/${firstPr!}`];
 	}
 
-	return [count, `/${forkedRepo}/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc+author%3A${getUsername()!}`];
+	return [count, `/${forkedRepo}/pulls?q=${encodeURIComponent('is:pr is:open sort:updated-desc author:@me')}`];
```
fregante, line 63: Might as well change it to use URLSearchParams or SearchQuery 👍

kidonng: There's only one parameter so `URLSearchParams` doesn't make much difference and `SearchQuery` is `URLSearchParams` under the hood.

source/features/preview-hidden-comments.tsx
cheap-glitch: Yes I remember doing this on purpose. Though this could make clearer, at the code of a little redundancy:
```ts
const headerTextWrapper = header.tagName === 'H3' ? select(':scope > .d-inline-block:last-child', header) ?? header.firstChild : header.firstChild
```

source/features/show-open-prs-of-forks.tsx
fregante: If a "lint" change is made to this line, then it should be linted fully
```suggestion
	const url = new URL(`/${forkedRepo}/pulls`);
	url.searchParams.set('q', 'is:pr is:open sort:updated-desc author:@me');
	return [count, url.href];
```

Building search parameters with strings is generally not advised on the repo

source/features/esc-to-deselect-line.tsx
kidonng: The spec says the type can be `any` so `undefined` is OK.

fregante (approved):
LGTM so far except the open review

kidonng:
> LGTM so far except the open review

The commit changing that part is already reverted, I just left it open to see if I can come up with an answer.

kidonng:
I do have some pending refactor changes but they are more suited for a new PR. So @refined-github/maintainers review and 🚢

cheap-glitch (approved):

PR #5041, opened 2021-11-03
Add images for VIvaldi and Brave

wasi-master:
They are also supported because they're also based on chromium

I've have been using the extension for a long time on vivaldi and it works perfectly, I also recently tried it out on brave and it also seems to be working pretty well

## Test URLs

[The project readme](https://github.com/refined-github/refined-github#readme)

### **Before:**

Image showing what it looked like before

### **After:**

Image showing what it looks like after

PR #5045, opened 2021-11-05
Added brave icon

jerskisnow:
This PR only contains an extra image in the README indicating that the chrome extension is also supported for Brave, because brave is worthy of being mentioned alongside opera and edge.


cheap-glitch:
Duplicate of #4300 and #5041.

kidonng:
I don't know if this happened to be created just after #5041 but maybe revising the README would reduce confusion:

```diff
- also compatible with <Edge icon> <Opera icon>
+ also compatible with Chromium-based browsers
```

@refined-github/maintainers thoughts?

cheap-glitch:
Maybe
```
also compatible with <Edge icon> <Opera icon> and other Chromium-based browsers
```

fregante:
> would reduce confusion

I don't think it would reduce the amount of PRs nor the requests to "add support for Edge"

kidonng:
The issue I see is not that people know no background about their browsers, it's that we list _some_ Chromium-based ones. So I suggest we just list none.

fregante:
> I suggest we just list Chrome only.

I'd prefer that honestly, I just noticed that those logos stopped the requests/PRs for Edge/Opera. There are other browsers but their usage/recognition is minimal. I mean, heck, Brave is at 0.05%, which is 2% of Opera’s usage, which itself is 2% of the global. https://kinsta.com/browser-market-share/

PR #5049, opened 2021-11-05
Meta: drop magic globals in favor of `import`ing the readme

fregante:
- Closes https://github.com/refined-github/refined-github/issues/4320

## Pros

- Readme.md becomes part of the webpack dependency tree, which means it's watched just like other files 
- No magic/global variables that need special handling (e.g. "Don't repeat the magic variable, it will be inlined multiple times")
- More portable codebase, the loader can be more easily ported to esbuild [or something](https://github.com/refined-github/refined-github/pull/5014)
- It makes it more clear that the data is coming from the readme - https://github.com/refined-github/refined-github/pull/3678

## Cons

- A lil hacky because it doesn't use the correct webpack APIs

### DIFF
--- a/build/readme.loader.cts
+++ b/build/readme.loader.cts
@@ -2,7 +2,7 @@
 module.exports = async function ReadmeLoader () {
 	const {getFeatures, getFeaturesMeta} = await import("./readme-parser.js");
 	return `
-		export const features = ${JSON.stringify(getFeatures())};
+		export const featureList = ${JSON.stringify(getFeatures())};
 		export const featuresMeta = ${JSON.stringify(getFeaturesMeta())};
 	`;
 }
--- a/build/readme.loader.cts
+++ b/build/readme.loader.cts
@@ -1,0 +1,8 @@
+// Can't use modules because this is a bizarropackscript world
+module.exports = async function ReadmeLoader () {
+	const { getFeatures, getFeaturesMeta } = await import("./readme-parser.js");
+  return `
+		export const features = ${JSON.stringify(getFeatures())}
+		export const featuresMeta = [${JSON.stringify(getFeaturesMeta())}]
+  `;
+}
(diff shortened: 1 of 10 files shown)


webpack.config.ts
```diff
@@ -29,6 +27,10 @@ const config: Configuration = {
 	},
 	module: {
 		rules: [
+			{
+				test: /readme\.md?$/,
+				loader: './build/readme.loader.cts',
```
fregante, line 32: I'm using .cts here because .ts produces .js, which is interpreted as ESM due to `type:module`, but webpack’s ts-node `require`s it. This instead produces .cjs

build/readme.loader.cts
```diff
@@ -0,0 +1,8 @@
+// Can't use modules because this is a bizarropackscript world
+module.exports = async function ReadmeLoader () {
```
fregante, line 2: In reality we get the readme’s source here already, so it could use it instead of reading it from disk again

build/readme.loader.cts
```diff
@@ -0,0 +1,8 @@
...
+	const {getFeatures, getFeaturesMeta} = await import("./readme-parser.js");
+	return `
+		export const features = ${JSON.stringify(getFeatures())};
```
fregante, line 5: This in reality reads the list of enabled features from the `refined-github.ts`. It _could_ be replaced by a hook that reads the files directly from webpack’s tree… but this function is also used by the feature checker…

webpack.config.ts
```diff
@@ -49,8 +51,6 @@ const config: Configuration = {
 	plugins: [
 		new webpack.DefinePlugin({
 			// Passing `true` as the second argument makes these values dynamic — so every file change will update their value.
-			__features__: webpack.DefinePlugin.runtimeValue(() => JSON.stringify(getFeatures()), true),
-			__featuresMeta__: webpack.DefinePlugin.runtimeValue(() => JSON.stringify(getFeaturesMeta()), true),
 			__filebasename: webpack.DefinePlugin.runtimeValue(info => JSON.stringify(path.parse(info.module.resource).name)),
```
fregante, line 54: PR welcome to drop this too

> Tiny function like `getFeatureName(import.meta.url)` to replace `__basefilename`

PR #5054, opened 2021-11-07
Fix order of actions in `quick-review-buttons`

### LINKED ISSUE #5010: The default submit action of PR review form isn't "Approve"

cheap-glitch:
Fixes #5010 

## Test URLs

 * https://togithub.com/refined-github/sandbox/pull/4/files

## Screenshots

<table>
<tr>
	<th colspan="2" align="center">Open PR
<tr>
	<td>Before
	<td>After
<tr>
	<td>

screenshot-1636294365
	<td>

screenshot-1636297425
<tr>
	<th colspan="2" align="center">Own open PR
<tr>
	<td>Before
	<td>After
<tr>
	<td>

screenshot-1636294415
	<td>

screenshot-1636297479
</table>

### DIFF
--- a/source/features/quick-review-buttons.tsx
+++ b/source/features/quick-review-buttons.tsx
@@ -23,7 +23,8 @@ function init(): false | void {
 			<input
 				type="hidden"
 				name="pull_request_review[event]"
-				value="comment"/>,
+				value="comment"
+			/>,
 		);
 	}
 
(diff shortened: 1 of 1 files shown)


fregante:
Approve should stay on the right to preserve _our_ original order. Also because the positive action is usually on the right

fregante (approved):
Screen Shot

"Cancel review" button also verified

fregante:
I kinda changed my mind on this change in the meanwhile, but let's go forward because the hotfix has been there for a week already. We can review (undo) this in a few months and move the "Comment" button to the right.

The issue was specifically changing the right button’s action without changing its style. This PR still helps do this change in two phases.

PR #5055, opened 2021-11-07
Meta: Ensure features are described only once in the readme

gamemaker1:
## Related Issues

Mentioned in https://github.com/refined-github/refined-github/pull/4699#issuecomment-962593291.

## What this PR does

It ensures that all features are described only once in [`readme.md`](readme.md).

### DIFF
--- a/build/verify-features.ts
+++ b/build/verify-features.ts
@@ -55,7 +55,7 @@ function findError(filename: string): string | void {
 
 	const featureMeta = featuresInReadme.find(feature => feature.id === featureId);
 	if (!featureMeta) {
-		return `ERR: The feature ${featureId} should be described in the readme`;
+		return `ERR: ${featureId} should be described in the readme`;
 	}
 
 	if (featureMeta.description.length < 20) {
@@ -70,7 +70,13 @@ function findError(filename: string): string | void {
 		(lineMatch && lineMatch.length > 1) // If the description occurs more than once in the large list of features
 		|| (imageMatch && lineMatch) // If the description appears in both the feature list and the highlighted features section
 	) {
-		return `ERR: ${featureId} should be described only once in the readme`;
+		const matches = readmeContent.split(/\r?\n/).map((lineContent: string, lineNumber: number) =>
+			lineRegex.test(lineContent)
+				|| regexJoin(`<p><a title="${featureId}"></a> `).test(lineContent)
+				? lineNumber + 1 : -1,
+		).filter(lineNumber => lineNumber > 0);
+
+		return `ERR: ${featureId} is described more than once on lines ${matches.join(', ')}`;
 	}
 }
 
--- a/build/verify-features.ts
+++ b/build/verify-features.ts
@@ -1,7 +1,6 @@
-import regexJoin from 'regex-join';
 import { existsSync, readdirSync, readFileSync } from 'node:fs';
 
-import {findFeatureRegex, findHighlightedFeatureRegex, getFeatures, getFeaturesMeta} from './readme-parser.js'; // Must import as `.js`
+import {findFeatureRegex, getFeatures, getFeaturesMeta} from './readme-parser.js'; // Must import as `.js`
 
 const featuresDirContents = readdirSync('source/features/');
 const entryPoint = 'source/refined-github.ts';
@@ -43,8 +42,8 @@ function findError(filename: string): string | void {
 		return `ERR: The \`/source/features\` folder should only contain .css and .tsx files. File \`${filename}\` violates that rule`;
 	}
 
-	const featureId = filename.replace('.tsx', '');
-	if (!importedFeatures.includes(featureId as FeatureID)) {
+	const featureId = filename.replace('.tsx', '') as FeatureID;
+	if (!importedFeatures.includes(featureId)) {
 		return `ERR: ${featureId} should be imported by \`${entryPoint}\``;
 	}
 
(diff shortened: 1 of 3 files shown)


build/verify-features.ts
```diff
@@ -59,6 +61,14 @@ function findError(filename: string): string | void {
...
 		return `ERR: ${featureId} should be described better in the readme (at least 20 characters)`;
 	}
+
+	const lineRegex = regexJoin(/^/, `- [](# "${featureId}")`, /(?: 🔥)? (.+)$/gm);
+	const lineMatch = readmeContent.match(lineRegex);
+	// `lineMatch` might be null, but in that case the feature might be described
+	// as a highlight. Also, line 56-59 checks if all features are described already
+	if (lineMatch && lineMatch.length > 1) {
+		return `ERR: ${featureId} should be described only once in the readme`;
+	}
```
gamemaker1, line 71: I'll try adding support for parsing the 'highlight'ed features as well

gamemaker1: done!

build/verify-features.ts
```diff
@@ -1,10 +1,12 @@
 import {existsSync, readdirSync, readFileSync} from 'node:fs';
+import regexJoin from 'regex-join';
```
yakov116, line 2:
```suggestion
import regexJoin from 'regex-join';
import {existsSync, readdirSync, readFileSync} from 'node:fs';
```

Lint :)

yakov116:
@gamemaker1 are you able to give the line number in the console? (Only if very simple no need to work hard on it)

gamemaker1:
Will try it out

gamemaker1:
> Line numbers in the console

Done!

build/verify-features.ts
```diff
@@ -53,12 +55,29 @@ function findError(filename: string): string | void {
...
-		return `ERR: The feature ${featureId} should be described in the readme`;
+		return `ERR: ${featureId} should be described in the readme`;
 	}
 
 	if (featureMeta.description.length < 20) {
 		return `ERR: ${featureId} should be described better in the readme (at least 20 characters)`;
 	}
+
+	const lineRegex = regexJoin(/^/, `- [](# "${featureId}")`, /(?: 🔥)? (.+)$/gm);
+	const imageRegex = regexJoin(`<p><a title="${featureId}"></a> `, /(.+?)\n\t+<p><img src="(.+?)">/);
```
fregante, line 66: We already have the regexes somewhere, we should reuse them.

build/verify-features.ts
```diff
@@ -53,12 +55,29 @@ function findError(filename: string): string | void {
...
+	const imageMatch = readmeContent.match(imageRegex);
+	if (
+		(lineMatch && lineMatch.length > 1) // If the description occurs more than once in the large list of features
+		|| (imageMatch && lineMatch) // If the description appears in both the feature list and the highlighted features section
+	) {
+		const matches = readmeContent.split(/\r?\n/).map((lineContent: string, lineNumber: number) =>
+			lineRegex.test(lineContent)
+				|| regexJoin(`<p><a title="${featureId}"></a> `).test(lineContent)
+				? lineNumber + 1 : -1,
+		).filter(lineNumber => lineNumber > 0);
```
fregante, line 77: I don’t think we need the like number, that's easy to find once you know there's a duplicate

gamemaker1: Commented it out for now, let me know if I should remove it completely or keep it.

IMO it's a good feature, saves the effort of a manual ctrl/cmd+f.

build/verify-features.ts
gamemaker1: Done

gamemaker1:
Seems good to go... any other changes to be made?

build/verify-features.ts
```diff
@@ -53,12 +54,31 @@ function findError(filename: string): string | void {
...
+
+	// eslint-disable-next-line @typescript-eslint/prefer-regexp-exec
+	const featureMatch = readmeContent.match(findFeatureRegex(featureId as FeatureID));
+	// eslint-disable-next-line @typescript-eslint/prefer-regexp-exec
+	const highlightedFeatureMatch = readmeContent.match(findHighlightedFeatureRegex(featureId as FeatureID));
+	if (
+		(featureMatch && featureMatch.length > 1) // If the description occurs more than once in the large list of features
+		|| (highlightedFeatureMatch && featureMatch) // If the description appears in both the feature list and the highlighted features section
+	) {
+		// Uncomment the following to print out line numbers
```
cheap-glitch, line 72: I think unused code should be removed. If we change our minds, we can always get it back from this PR's commit.

gamemaker1: I just commented it out for now as I was waiting on the decision to include line numbers. If we decide to keep it, I will uncomment that part, else I will delete it completely.

build/verify-features.ts
```diff
@@ -53,12 +54,31 @@ function findError(filename: string): string | void {
...
+		return `ERR: ${featureId} should be described in the readme`;
 	}
 
 	if (featureMeta.description.length < 20) {
 		return `ERR: ${featureId} should be described better in the readme (at least 20 characters)`;
 	}
+
+	// eslint-disable-next-line @typescript-eslint/prefer-regexp-exec
+	const featureMatch = readmeContent.match(findFeatureRegex(featureId as FeatureID));
+	// eslint-disable-next-line @typescript-eslint/prefer-regexp-exec
```
fregante, line 66: Why?

build/verify-features.ts
```diff
@@ -53,12 +54,31 @@ function findError(filename: string): string | void {
...
+	const highlightedFeatureMatch = readmeContent.match(findHighlightedFeatureRegex(featureId as FeatureID));
+	if (
+		(featureMatch && featureMatch.length > 1) // If the description occurs more than once in the large list of features
+		|| (highlightedFeatureMatch && featureMatch) // If the description appears in both the feature list and the highlighted features section
+	) {
```
fregante, line 71:
```suggestion
	if (featureMatch?.length + highlightedFeatureMatch?.length > 1) {
```

If TS complains, add a couple of `Number()` calls

fregante (approved):
It can be merged after the 3 open reviews are fixed

build/verify-features.ts
gamemaker1: `match` is more concise in this case, as we just need to call it once to get all the matches. `exec` will need us to use a while loop to get all the matches.

build/verify-features.ts
gamemaker1: Used `(featureMatch?.length ?? 0) + (highlightedFeatureMatch?.length ?? 0) > 1`

build/verify-features.ts
gamemaker1: `exec` seems to be more efficient when I'm counting line numbers too... so I'm using `exec` now.

build/verify-features.ts
gamemaker1: I've kept the line number functionality, see my comment below.

gamemaker1:
I rewrote it to use `regexp.exec` instead of `string.match` - the linter was correct after all ;) - and it was much cleaner than using `string.match`. Line numbers for the position of descriptions are also printed in case of duplicates. The following cases are checked:

- If a feature is described twice in the feature list
- If a feature is part of the highlighted features and part of the feature list

The following case, however, is not checked:

- If a feature is described twice within the highlights section

I am trying to add that right now.

build/verify-features.ts
```diff
@@ -67,8 +67,21 @@ function findError(filename: string): string | void {
...
+	do {
+		highlightedFeatureMatch = highlightedFeatureRegex.exec(readmeContent);
+		if (highlightedFeatureMatch) {
+			lines.push(readmeContent.slice(0, highlightedFeatureMatch.index).split(/\r?\n/).length);
+		}
+	} while (highlightedFeatureMatch);
+
+	if (lines.length > 0) {
+		return `ERR: ${featureId} should be described only once in the readme, but it is also described on line(s) ${lines.join(', ')}`;
+	}
```
gamemaker1, line 81: For some reason this regex returns matches only if the feature has been described more than once in the highlights section... not sure why that is happening

gamemaker1: The other regex (for the list of features) is matching all the descriptions

fregante (changes requested):
> I've kept the line number functionality, see my comment below.

I had already reviewed and rejected this part

gamemaker1:
Removed the line number functionality, please let me know if there are any other changes I should make.

fregante:
Thank you!

PR #5059, opened 2021-11-08
Meta: Don't store entire workflow files in cache

cheap-glitch:
Follow-up of https://github.com/refined-github/refined-github/pull/4998#issuecomment-958757002

## Test URLs

## Screenshot

image

### DIFF
--- a/source/features/clean-repo-tabs.tsx
+++ b/source/features/clean-repo-tabs.tsx
@@ -5,9 +5,9 @@ import * as pageDetect from 'github-url-detection';
 
 import features from '.';
 import fetchDom from '../helpers/fetch-dom';
+import * as api from '../github-helpers/api';
 import getTabCount from '../github-helpers/get-tab-count';
 import looseParseInt from '../helpers/loose-parse-int';
-import {getWorkflows} from './scheduled-and-manual-workflow-indicators';
 import abbreviateNumber from '../helpers/abbreviate-number';
 import {buildRepoURL, getRepo} from '../github-helpers';
 import {onlyShowInDropdown, unhideOverflowDropdown} from './more-dropdown-links';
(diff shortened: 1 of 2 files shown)


source/features/clean-repo-tabs.tsx
```diff
@@ -5,9 +5,9 @@ import * as pageDetect from 'github-url-detection';
 
 import features from '.';
 import fetchDom from '../helpers/fetch-dom';
+import * as api from '../github-helpers/api';
 import getTabCount from '../github-helpers/get-tab-count';
 import looseParseInt from '../helpers/loose-parse-int';
-import {getWorkflows} from './scheduled-and-manual-workflow-indicators';
```
cheap-glitch, line 10: The raw workflow files were also used here, but since only their count matters I've uncoupled the two API calls completely.

source/features/clean-repo-tabs.tsx
```diff
@@ -40,6 +40,22 @@ const getWikiPageCount = cache.function(async (): Promise<number> => {
 	cacheKey: () => 'wiki-page-count:' + getRepo()!.nameWithOwner,
 });
 
+const getWorkflowsCount = cache.function(async (): Promise<number> => {
+	const {repository: {workflowFiles}} = await api.v4(`
+		repository() {
+			workflowFiles: object(expression: "HEAD:.github/workflows") {
+				... on Tree { entries { oid } }
```
cheap-glitch, line 47: Seems weird we have to fetch something just to count the entries, there has to be a better way :thinking:

fregante: I assume it's
```suggestion
				totalCount
```

fregante (commented):
If possible we should also match workflows by filename rather than the name specified by the workflow content. The query should include `object.text` and I think `object.name`

fregante (approved):
It can be merged afterwards

source/features/clean-repo-tabs.tsx
```diff
@@ -40,6 +40,22 @@ const getWikiPageCount = cache.function(async (): Promise<number> => {
 	cacheKey: () => 'wiki-page-count:' + getRepo()!.nameWithOwner,
 });
 
+const getWorkflowsCount = cache.function(async (): Promise<number> => {
```
fregante, line 43: This fixes part of https://github.com/refined-github/refined-github/issues/5040

source/features/clean-repo-tabs.tsx
cheap-glitch: I thought so but it doesn't work. The docs don't mention any `length` or `count` fields on `Tree` or `entries`, so I think we're stuck with this :grimacing:

fregante:
It can be merged if this last change was tested

PR #5060, opened 2021-11-08
Fix implementation of hotfixes

### LINKED ISSUE #4861: The hotfix expires too soon
### Please ensure:

### Description

GitHub feed showing weird white-colored panel in the middle. I'm using Brave browser with dark mode enabled.

### Screenshot

Without extension (default): 

image

With extension: 

image

### Console errors

_No response_

### Example URL

https://github.com/

### Browser(s) used

Brave (Chromium-based)

cheap-glitch:
Follow-up of https://github.com/refined-github/refined-github/pull/4961#pullrequestreview-796250499
Fixes #4861 

## Test URLs
n/a

## Screenshot
screenshot-1636376145

### DIFF
--- a/source/features/index.tsx
+++ b/source/features/index.tsx
@@ -109,7 +109,7 @@ const globalReady: Promise<RGHOptions> = new Promise(async resolve => {
 		Object.assign(options, bisectedFeatures);
 	} else {
 		// If features are remotely marked as "seriously breaking" by the maintainers, disable them without having to wait for proper updates to propagate #3529
-		void updateHotfixes();
+		void updateHotfixes(version);
 		Object.assign(options, localHotfixes);
 	}
 
--- a/source/helpers/hotfix.ts
+++ b/source/helpers/hotfix.ts
@@ -38,6 +38,7 @@ export const updateHotfixes = cache.function(async (version: string): Promise<Ho
 	return storage;
 }, {
 	maxAge: {hours: 6},
+	staleWhileRevalidate: {days: 3},
 	cacheKey: () => 'hotfixes',
 });
 
@@ -53,6 +54,7 @@ export const updateStyleHotfixes = cache.function(async (version: string): Promi
 	return atob(content).trim();
 }, {
 	maxAge: {hours: 6},
+	staleWhileRevalidate: {days: 3},
 	cacheKey: () => 'style-hotfixes',
 });
 
--- a/source/helpers/hotfix.ts
+++ b/source/helpers/hotfix.ts
@@ -4,7 +4,9 @@ import compareVersions from 'tiny-version-compare';
 
 import {RGHOptions} from '../options-storage';
 
-export const updateHotfixes = cache.function(async (): Promise<string[][]> => {
+export type HotfixStorage = Array<[FeatureID, string]>;
+
+export const updateHotfixes = cache.function(async (version: string): Promise<HotfixStorage> => {
 	// The explicit endpoint is necessary because it shouldn't change on GHE
 	// We can't use `https://raw.githubusercontent.com` because of permission issues https://github.com/refined-github/refined-github/pull/3530#issuecomment-691595925
 	const request = await fetch('https://api.github.com/repos/refined-github/refined-github/contents/hotfix.csv?ref=hotfix');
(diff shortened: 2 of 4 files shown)


source/helpers/hotfix.ts
```diff
@@ -15,10 +17,16 @@ export const updateHotfixes = cache.function(async (): Promise<string[][]> => {
...
 	}
 
-	return atob(content)
-		.trim()
-		.split('\n')
-		.map(line => line.split(','));
+	const storage: HotfixStorage = [];
+	const lines = atob(content).trim().split('\n');
+	for (const line of lines) {
+		const [featureID, unaffectedVersion, relatedIssue] = line.split(',');
```
cheap-glitch, line 23: Maybe the line itself should be trimmed too just in case?

source/helpers/hotfix.ts
```diff
@@ -15,10 +17,16 @@ export const updateHotfixes = cache.function(async (): Promise<string[][]> => {
...
+		if (!unaffectedVersion || compareVersions(unaffectedVersion, version) > 0) {
```
cheap-glitch, line 24: We could also check that `featureID` and `relatedIssue` are not empty of `undefined` (that would avoid storing the example line).

source/features/index.tsx
```diff
@@ -109,7 +109,7 @@ const globalReady: Promise<RGHOptions> = new Promise(async resolve => {
 		Object.assign(options, bisectedFeatures);
 	} else {
 		// If features are remotely marked as "seriously breaking" by the maintainers, disable them without having to wait for proper updates to propagate #3529
-		void updateHotfixes();
+		void updateHotfixes(version);
```
fregante, line 112: ?



The cache is already updated when the version changes, there's only ever one cache per version. This applies to both JS and CSS fixes

source/helpers/hotfix.ts
fregante: Maybe
```suggestion
		const [featureID, unaffectedVersion, relatedIssue] = line.split(/\s*,\s*/);
```

fregante (commented):
<3

source/features/index.tsx
cheap-glitch: We need it to compare to the unaffected versions when filtering. Also for `updateStyleHotfixes()` it's used to construct the file URL.

Though I guess it could be grabbed directly in `hotfix.ts`.

fregante:
Related, if you want to look into it https://github.com/refined-github/refined-github/issues/4861

source/helpers/hotfix.ts
```diff
@@ -15,12 +28,17 @@ export const updateHotfixes = cache.function(async (): Promise<string[][]> => {
...
+	for (const [featureID, unaffectedVersion, relatedIssue] of parseCsv(atob(content))) {
+		if (featureID && relatedIssue && (!unaffectedVersion || compareVersions(unaffectedVersion, version) > 0)) {
+			storage.push([featureID as FeatureID, relatedIssue]);
+		}
+	}
+
+	return storage;
 }, {
 	maxAge: {hours: 6},
+	staleWhileRevalidate: {days: 3},
```
cheap-glitch, line 41: Hopefully this fixes the situation described in https://github.com/refined-github/refined-github/issues/4861#issuecomment-933148944.

cheap-glitch: @fregante :rocket: ?

source/helpers/hotfix.ts
```diff
@@ -15,12 +28,17 @@ export const updateHotfixes = cache.function(async (): Promise<string[][]> => {
...
 	}
 
-	return atob(content)
-		.trim()
-		.split('\n')
-		.map(line => line.split(','));
+	const storage: HotfixStorage = [];
+	for (const [featureID, unaffectedVersion, relatedIssue] of parseCsv(atob(content))) {
+		if (featureID && relatedIssue && (!unaffectedVersion || compareVersions(unaffectedVersion, version) > 0)) {
+			storage.push([featureID as FeatureID, relatedIssue]);
```
fregante, line 34: Is this not working? All of them are still appearing locally even if I have v21.11.1. Safari

Screen Shot 5

Issue #4891, opened 2021-10-07
Drop `sort-milestones-by-closest-due-date`


kidonng:
### Please ensure:

- [X] The bug is caused by Refined GitHub. It doesn't happen if I disable the extension.

### Description

`sort-milestones-by-closest-due-date` is a little "default search params" feature added more than 4 years ago in #457.

It has a bug, [introduced in the original implementation PR](https://github.com/refined-github/refined-github/pull/457/commits/d2d83fa6b0310c6917c66d820dadcbdbc544495d#diff-1f93cd4783614fe92bbb9d7fba4cd2b31ba1a47c0c2e9b00be36bd3246e5af7fR477). Now I dare you to find it... just kidding.

To reproduce, just go to [some random repo's issue page](https://github.com/microsoft/vscode/issues) and click the "Milestones" link. Now try changing the sort criteria to the default "Recently updated". It won't work because all "milestones" link without explicit criteria are affected.

---

Now you may think, "not a big thing, let's just fix it". Instead, I don't see this feature useful for most people.

- First, the bug has been there for more than 4 years and no one noticed/questioned about it. That alone proves "Milestones" isn't that frequently* used.
  \* To be objective, it is still used more than "Projects", but still.
- Second, GitHub has a sane default in this case. Remember we have a `sort-conversations-by-update-time` feature for this?
- Finally, if you argue (quoting from the [implementation PR](https://github.com/refined-github/refined-github/pull/457#issue-123211247)): "Milestones are used to manage releases; where as the closest one is often more important than the last updated one."

  Well... that's not how things work in reality. Reusing [the random repo](https://github.com/microsoft/vscode/milestones?state=open&direction=asc&sort=due_date) example, you will find it just sorts milestone by their ID/create time, from old to new. If that doesn't sound like an issue to you, take a look at the [closed](https://github.com/microsoft/vscode/milestones?state=closed&direction=asc&sort=due_date) milestones.

  Think about it again, VSCode is maintained by a company, with a fixed release cycle. That's not how most open source projects work. Let's try again with [another random but real™️ open source project](https://github.com/fish-shell/fish-shell/milestones?direction=asc&sort=due_date). Wait, the results looks very similar...

  OK, so you want a project from someone familiar, [here's one](https://github.com/sindresorhus/got/milestones?direction=asc&sort=due_date).

[...]

PR #5078, opened 2021-11-16
Fix `conversation-filters` showing multiple checked marks

### LINKED ISSUE #4589: `conversation-filters` shows the wrong "current filter" checkbox
### Please ensure:

"Everything mentioning you" is cloned as a whole and the checkbox is preserved, but it shouldn't be

Screen Shot

### Console errors, if any

_No response_

### Example URL

```js
// https://github.com/sindresorhus/refined-github/issues?q=is%3Aopen+mentions%3A%40me
```

### Browser(s) used

safari

cheap-glitch:
Fixes #4589 

## Test URLs
```js
// https://github.com/refined-github/refined-github/issues?q=is%3Aopen+mentions%3A%40me
```

## Screenshot

image

### DIFF
--- a/source/features/conversation-filters.tsx
+++ b/source/features/conversation-filters.tsx
@@ -12,12 +12,14 @@ function init(): void {
 	commentsLink.lastChild!.textContent = 'Everything commented by you';
 	commentsLink.removeAttribute('target');
 	new SearchQuery(commentsLink).set('is:open commenter:@me');
+	commentsLink.setAttribute('aria-checked', String(commentsLink.href === location.href)); // #4589
 
 	sourceItem.after(commentsLink);
 
 	// Add "Everything you subscribed to" link
 	const subscriptionsLink = commentsLink.cloneNode(true);
 	subscriptionsLink.lastChild!.textContent = 'Everything you subscribed to';
+	subscriptionsLink.setAttribute('aria-checked', 'false'); // #4589
 
 	const subscriptionsUrl = new URL('https://github.com/notifications/subscriptions');
 	const repositoryId = select('input[name="repository_id"]')!.value;


source/features/conversation-filters.tsx
```diff
@@ -12,12 +12,14 @@ function init(): void {
 	commentsLink.lastChild!.textContent = 'Everything commented by you';
 	commentsLink.removeAttribute('target');
 	new SearchQuery(commentsLink).set('is:open commenter:@me');
+	commentsLink.setAttribute('aria-checked', String(commentsLink.href === location.href)); // #4589
```
cheap-glitch, line 15: This allow marking the filter as selected, just like the native ones.

image

It follows the behavior of the native filters, that are only shown selected when the current search query matches exactly with the one they represent.

source/features/conversation-filters.tsx
```diff
@@ -12,12 +12,14 @@ function init(): void {
...
 
 	sourceItem.after(commentsLink);
 
 	// Add "Everything you subscribed to" link
 	const subscriptionsLink = commentsLink.cloneNode(true);
 	subscriptionsLink.lastChild!.textContent = 'Everything you subscribed to';
+	subscriptionsLink.setAttribute('aria-checked', 'false'); // #4589
```
cheap-glitch, line 22: We never want this to be selected since it's just a link to another page.

cheap-glitch:
I've noticed an inconsistent behavior due to `sort-conversations-by-update-time`: when first selecting a filter (native or not), the results are going to be ordered by update time, which means the query will not be exactly the one of the filter and thus it will not show up as selected.

What's even more confusing, when selecting the filter again, this time it will work as expected because the sorting feature wasn't reapplied due to the AJAX navigation.

https://user-images.githubusercontent.com/46634000/141992334-904349e2-84a3-40b8-82a7-25f844d53bb0.mp4

So imo `sort-conversations-by-update-time` should not apply to conversation filters at all.

kidonng:
> I've noticed an inconsistent behavior due to `sort-conversations-by-update-time`: when first selecting a filter (native or not), the results are going to be ordered by update time, which means the query will not be exactly the one of the filter and thus it will not show up as selected.

A similar bug is reported in #4891. While I think sorting by update time is helpful, it should not _override_ what the filters should do.

fregante:
> While I think sorting by update time is helpful, it should not _override_ what the filters should do.

It should. That feature is meant to change the default of the website everywhere, so unless you intend to change the sorting, then _every link_ should be affected.

I suppose that the drawback of this are widgets that have a strict "current" implementation. For example on the global PR list page `(/pulls)` the filters continue working even with the extra `sort:`

fregante:
> the sorting feature wasn't reapplied due to the AJAX navigation

That's a bug (of the sorting feature)



> I've noticed an inconsistent behavior due to [`sort-conversations-by-update-time`](/refined-github/refined-github/blob/main/source/features/sort-conversations-by-update-time.tsx)

I suppose that that feature would have to also deal with the "current" implementation 😬 which I wouldn't be a fan of… so hopefully there's an elegant solution.

cheap-glitch:
Opened #5090 to fix the issues with `sort-conversations-by-update-time`.

fregante (approved):
Untested but TMNT

PR #5090, opened 2021-11-20
Fix `sort-conversations-by-update-time` on conversation filters

cheap-glitch:
Follow-up of https://github.com/refined-github/refined-github/pull/5078#issuecomment-971465633

## Test URLs
 * https://github.com/refined-github/refined-github/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc

## Screenshot

https://user-images.githubusercontent.com/46634000/142731563-0a0fa9bb-d683-4016-9632-872778534c60.mp4


source/features/sort-conversations-by-update-time.tsx
```diff
@@ -31,5 +36,12 @@ function init(): void {
 }
 
 void features.add(import.meta.url, {
+	deduplicate: 'has-rgh-inner',
```
cheap-glitch, line 39: Fixes deduplication issue (e.g. when selecting several filters one after the other).

source/features/sort-conversations-by-update-time.tsx
```diff
@@ -4,6 +4,11 @@ import * as pageDetect from 'github-url-detection';
 import features from '.';
 import SearchQuery from '../github-helpers/search-query';
 
+function selectCurrentConversationFilter(): void {
+	const currentSearchURL = location.href.replace('/pulls?', '/issues?'); // Replacement needed to make up for the redirection of "Your pull requests" link
+	select(`#filters-select-menu a.SelectMenu-item[href="${currentSearchURL}"]`)?.setAttribute('aria-checked', 'true');
```
cheap-glitch, line 9: Marks the correct current filter as selected, just like in native GitHub.

fregante: I suppose this would work. My concern is that I don't see a `aria-checked = false` anywhere here, so potentially we're leaving pre-selected options unchanged.

Maybe it should look like:

```js
const currentItem = select(etc)

if (currentItem)
	select(aria-selected=true).setAttribute(false)
	currentItem.setAttribute(true)

fregante (approved):
I'm 50/90 on this

PR #5091, opened 2021-11-20
Fix broken "Details" links in PR checks due to `bypass-checks`

### LINKED ISSUE #3938: `bypass-checks` breaks codecov link on PR
On page like `https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1088`, the codecov link is `https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1088/checks?check_run_id=1797833452`, but enables `bypass-checks`, the link will be `https://codecov.io/`, should link to the page for this RR on codecov.io

cheap-glitch:
Fix #3938 

## Test URLs

 * "Details" link should by bypassed
   * https://togithub.com/Kocal/semantic-release-preset/pull/7 (Travis checks)
 * "Details" link should NOT be bypassed
   * https://togithub.com/sindresorhus/eslint-plugin-unicorn/pull/1088#commits-pushed-d34fa26 (commit `f800f82`, "codecov/project")
   * https://togithub.com/mergeability/mergeable/pull/519 ("Mergeable")
   * https://togithub.com/InsightSoftwareConsortium/ITK/pull/2881 ("ghostflow-check-master")

## Screenshot

image

### DIFF
--- a/source/features/bypass-checks.tsx
+++ b/source/features/bypass-checks.tsx
@@ -15,9 +15,16 @@ async function bypass(detailsLink: HTMLAnchorElement): Promise<void> {
 	}
 
 	const {details_url: detailsUrl} = await api.v3(`check-runs/${runId}`);
-	if (detailsUrl && !detailsUrl.startsWith('https://github.com/') && new URL(detailsUrl).pathname !== '/') { // Ignore links to GitHub repos or static product pages #3938
-		detailsLink.href = detailsUrl;
+	if (!detailsUrl) {
+		return;
 	}
+
+	const {pathname, search: queryString} = new URL(detailsUrl);
+	if (detailsUrl.startsWith('https://github.com/') || (pathname === '/' && queryString === '')) { // Ignore links to GitHub repos or static product pages #3938
+		return;
+	}
+
+	detailsLink.href = detailsUrl;
 }
 
 function init(): void {
--- a/source/features/bypass-checks.tsx
+++ b/source/features/bypass-checks.tsx
@@ -5,6 +5,13 @@ import * as pageDetect from 'github-url-detection';
 import features from '.';
 import * as api from '../github-helpers/api';
 
+const uselessDetailsLinks = new Set([
+	'https://codecov.io', // https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1088
+	'https://github.com/jusx/mergeable', // https://github.com/mergeability/mergeable/pull/519
+	'https://github.com/marketplace/wip', // https://github.com/InsightSoftwareConsortium/ITK/pull/2881
+	'https://github.com/apps/kwrobot-v1',
+]);
+
 async function bypass(detailsLink: HTMLAnchorElement): Promise<void> {
 	const runId = pageDetect.isActionJobRun(detailsLink)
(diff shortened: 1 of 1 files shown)


fregante:
Thanks, I'd rather not keep a list though. I think we should add a "external link" icon instead of replacing Details. This way the user the choose.

We can also avoid adding such icon when it points to an obviously unrelated link, like a `https://raw.domain.com/` without any relevance to the build itself.

cheap-glitch:
> I think we should add a "external link" icon (for example) instead of replacing Details. This way the user can choose.

That's what you said in the issue discussion, but then you changed your mind in favor of an "allowlist". (Just checking, I don't mind either way).

fregante:
It's groundhog day

fregante:
Ok, but let's include more apps. For example all the URLs starting with GitHub.com/apps are useless. Same goes with URLs that only have a hostname and no URL parameters

source/features/bypass-checks.tsx
```diff
@@ -14,8 +14,10 @@ async function bypass(detailsLink: HTMLAnchorElement): Promise<void> {
 		return;
 	}
 
-	const directLink = await api.v3(`check-runs/${runId}`);
-	detailsLink.href = directLink.details_url;
+	const {details_url: detailsUrl} = await api.v3(`check-runs/${runId}`);
+	if (detailsUrl && !detailsUrl.startsWith('https://github.com/') && new URL(detailsUrl).pathname !== '/') { // Ignore links to GitHub repos or static product pages #3938
```
cheap-glitch, line 18: I think we can safely ignore all "details" links pointing to GitHub, because those mean an app without its own web interface.

fregante, line 18: I mentioned search params because they could use "http://app.code.com/?run_id=58383" and fall into this exception, but I think it's fine as a false negative, users would just be excluded from this feature rather than receive a broken link.

fregante (approved):

source/features/bypass-checks.tsx
cheap-glitch: I've added a check for the query string and made the conditions more readable.

source/features/bypass-checks.tsx
```diff
@@ -15,9 +15,16 @@ async function bypass(detailsLink: HTMLAnchorElement): Promise<void> {
 	}
 
 	const {details_url: detailsUrl} = await api.v3(`check-runs/${runId}`);
```
fregante, line 17: Have they implemented a v4 in the meanwhile?

fregante (approved):

source/features/bypass-checks.tsx
cheap-glitch: Yes, but we have to fetch the runs through each commit, it's quite cumbersome:

```graphql
{
  repository(name: "refined-github", owner: "refined-github") {
    object(expression: "869d9a7d818e029828a0ff2db0b6d3d4b389c44a") {
      ... on Commit {
        checkSuites(first: 10) {
          nodes {
            checkRuns(first: 10) {
              nodes {
                detailsUrl
              }
            }
          }
        }
      }
    }
  }
}
```

PR #5094, opened 2021-11-21
Fix `highlight-deleted-and-added-files-in-diffs` on compare page

### LINKED ISSUE #5093: `highlight-deleted-and-added-files-in-diffs` doesn't work on compare pages
### Please ensure:

### Include in this issue:

- [ ] Console errors, if any

### Description

The file list on compare pages is lazyloaded so `highlight-deleted-and-added-files-in-diffs` doesn't apply:

image

### Example URL

https://github.com/kidonng/xuann.wang/compare/576490d6a60a52601fbc33f31d5d9afcc3840bde...34b0d598eae7633d64f1ecbeb927254e74737905#diff-05ed1c7f99e45b485708115502a1e0f28c8547e9a9a29c1b8664f79103cf7873

### Browser(s) used

Chrome

kidonng:
Fix #5093

## Test URLs

[`576490d6a60a52601fbc33f31d5d9afcc3840bde...34b0d598eae7633d64f1ecbeb927254e74737905`#diff-05ed1c7f99](https://github.com/kidonng/xuann.wang/compare/576490d6a60a52601fbc33f31d5d9afcc3840bde...34b0d598eae7633d64f1ecbeb927254e74737905#diff-05ed1c7f99e45b485708115502a1e0f28c8547e9a9a29c1b8664f79103cf7873)

## Screenshot

Before:

image

After:

image

### DIFF
--- a/source/features/highlight-deleted-and-added-files-in-diffs.tsx
+++ b/source/features/highlight-deleted-and-added-files-in-diffs.tsx
@@ -57,10 +57,6 @@ async function init(): Promise<void | false> {
 	deinit.push(observer.abort);
 }
 
-function initCompare(): void {
-	onDiffFileLoad(init);
-}
-
 void features.add(import.meta.url, {
 	include: [
 		pageDetect.isPRFiles,
@@ -81,5 +77,9 @@ void features.add(import.meta.url, {
 	exclude: [
 		() => select.exists('.blankslate:not(.blankslate-large)'),
 	],
-	init: initCompare,
+	additionalListeners: [
+		onDiffFileLoad,
+	],
+	onlyAdditionalListeners: true,
+	init,
 });
--- a/source/features/highlight-deleted-and-added-files-in-diffs.tsx
+++ b/source/features/highlight-deleted-and-added-files-in-diffs.tsx
@@ -22,7 +22,7 @@ async function init(): Promise<void | false> {
 	const fileList = await elementReady([
 		'.toc-select details-menu[src*="/show_toc?"]', // `isPR`
 		'.toc-diff-stats + .content', // `isSingleCommit` and `isCompare`
-	].join(','));
+	].join(','), {stopOnDomReady: false});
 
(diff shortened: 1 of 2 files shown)


cheap-glitch:
This should probably be fixed by listening for the `load` event of the `include-fragment` instead:

image

This is likely the same solution as in #5066. Also there may be other features affected by this, so a reusable `onCompareDiffLoad` event listener might be useful.

source/features/highlight-deleted-and-added-files-in-diffs.tsx
```diff
@@ -22,7 +22,7 @@ async function init(): Promise<void | false> {
 	const fileList = await elementReady([
 		'.toc-select details-menu[src*="/show_toc?"]', // `isPR`
 		'.toc-diff-stats + .content', // `isSingleCommit` and `isCompare`
-	].join(','));
+	].join(','), {stopOnDomReady: false});
```
cheap-glitch, line 25: This should probably be fixed by listening for the `load` event of the `include-fragment` instead:

image

This is likely the same solution as in #5066. Also there may be other features affected by this, so a reusable `onCompareDiffLoad` event listener might be useful.

source/features/highlight-deleted-and-added-files-in-diffs.tsx
```diff
@@ -61,11 +57,14 @@ async function init(): Promise<void | false> {
 	deinit.push(observer.abort);
 }
 
+function initCompare(): void {
+	onDiffFileLoad(init);
+}
```
cheap-glitch, line 62: Maybe this could be replaced by `additionalListeners` + `additionalListenersOnly` in the `feature.add()`?

kidonng: Good to know that!

cheap-glitch (approved):
Tested on Firefox :100:

PR #5095, opened 2021-11-21
Fix some missing editor features when editing comments

### LINKED ISSUE #4542: Some features don't work on lazy-loaded comment forms
Also the upload button resurfaced by `minimize-upload-bar` is not showing.

1. Open any PR, like https://github.com/sindresorhus/refined-github/pull/4507
2. Start a review
3. Try to comment on the same thread

They appear when starting on a new thread, but not when continuing on an existing thread.

Screen Shot 15

cheap-glitch:
Extracted from #5066 
Partially fixes #4542 

## Test URLs

 * Edit a regular comment
 * Edit a review comment

## Screenshot

https://user-images.githubusercontent.com/46634000/141314939-c863a86a-e4b1-46f7-8cbd-e421b9a8c6e6.mp4

### DIFF
--- a/source/features/collapsible-content-button.tsx
+++ b/source/features/collapsible-content-button.tsx
@@ -6,8 +6,8 @@ import * as pageDetect from 'github-url-detection';
 import * as textFieldEdit from 'text-field-edit';
 
 import features from '.';
-import onCommentEdit from '../github-events/on-comment-edit';
 import smartBlockWrap from '../helpers/smart-block-wrap';
+import {onCommentEdit} from '../github-events/on-fragment-load';
 
 function addContentToDetails({delegateTarget}: delegate.Event<MouseEvent, HTMLButtonElement>): void {
 	const field = delegateTarget.form!.querySelector('textarea')!;
--- a/source/features/collapsible-content-button.tsx
+++ b/source/features/collapsible-content-button.tsx
@@ -6,6 +6,7 @@ import * as pageDetect from 'github-url-detection';
 import * as textFieldEdit from 'text-field-edit';
 
 import features from '.';
+import onCommentEdit from '../github-events/on-comment-edit';
 import smartBlockWrap from '../helpers/smart-block-wrap';
 
 function addContentToDetails({delegateTarget}: delegate.Event<MouseEvent, HTMLButtonElement>): void {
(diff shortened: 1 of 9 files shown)


cheap-glitch (commented):
This PR:

 * Extracts a generic function to create `<include-fragment>` listeners from `on-diff-file-load.ts` (might also be useful to fix #5093)

 * Enables `table-input` and `collapsible-content-button` _when editing comments_
 
 * Restores the upload bar _when editing comments_

source/features/minimize-upload-bar.css
```diff
@@ -1,9 +1,10 @@
 /* `.js-upload-markdown-image` excludes non-comment attachment drop areas. See #2086 */
-.rgh-minimize-upload-bar .js-upload-markdown-image.is-default .drag-and-drop {
+/* `:not(.js-comment-update)` excludes comment edition forms, as they lack the upload image button. See #4542 */
+.rgh-minimize-upload-bar form:not(.js-comment-update) .js-upload-markdown-image.is-default .drag-and-drop {
 	display: none !important;
 }
 
-.rgh-minimize-upload-bar .js-upload-markdown-image.is-default textarea {
+.rgh-minimize-upload-bar form:not(.js-comment-update) .js-upload-markdown-image.is-default textarea {
```
fregante, line 7: This is a GitHub bug by the way, if you'd like, can you report it? "Upload button not showing on mobile when editing comments"

source/features/collapsible-content-button.tsx
```diff
@@ -44,6 +45,12 @@ function init(): void {
 	}
 }
 
+function init(): void {
+	delegate(document, '.rgh-collapsible-content-btn', 'click', addContentToDetails);
+	addButtons();
+	onCommentEdit(addButtons);
```
fregante, line 51: As far as I know, GitHub duplicates a form that already exists on the page (this was the case for new reviews at least) so that means we only had to edit one form once and then GitHub would just clone it. 

Is that no longer the case? Or maybe here GitHub started loading the whole edit field when clicking Edit? If that's what's happening, then this is correct.

source/features/table-input.tsx
```diff
@@ -70,6 +69,13 @@ function init(): void {
 	}
 }
 
+function init(): void {
+	delegate(document, '.rgh-table-input-cell', 'click', addTable);
+	delegate(document, '.rgh-table-input-cell', 'mouseenter', highlightSquares, {capture: true});
```
fregante, line 74: I'm still disappointed we need this delegated mousenter event because it triggers hundreds of times for every element when the mouse just moves around. Delegate then uses element.match() to avoid calling this callback. 

I remember we tried a CSS-only solution but it didn't work maybe. Can you think of anything else? How often would an onmouseenter inline listener break?

source/github-events/on-diff-file-load.ts
```diff
@@ -11,9 +11,13 @@ const getDeduplicatedHandler = mem((callback: EventListener): delegate.EventHand
 	event.delegateTarget.addEventListener('load', callback);
 });
 
-export default function onDiffFileLoad(callback: EventListener): delegate.Subscription {
+export function createFragmentLoadListener(fragmentSelector: string, callback: EventListener): delegate.Subscription {
```
fregante, line 14: Ohhh there it was. Glad you found it. The file needs to be renamed though. It can export both functions without a default.

fregante (approved):

source/features/collapsible-content-button.tsx
cheap-glitch: > Is that no longer the case?

Doesn't seem like it:
 * the first comment of a conversation is created with a complete edit form on page load
 * all the other comments (regular & reviews) have an associated form with an `<include-fragment>` in it, which is loaded as soon as the user hovers on the "three dots" pop-up menu

source/features/table-input.tsx
cheap-glitch: I'll try something in a separate PR.

PR #5096, opened 2021-11-21
Skip `useful-not-found-page` on pages that never exist

kidonng:
The `commit` variable may be undefined, resulting in:

```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'oid')
    at getLatestCommitToFile (refined-github.js:15213)
    at async showAlternateLink (refined-github.js:15265)
```

## Test URLs

https://github.com/refined-github/refined-github/blob/d77b5ef46be8e935cf032184da94152e1dc660fa/source/features/rgh-linkify-features.ts

## Screenshot

N/A

### DIFF
--- a/source/features/useful-not-found-page.tsx
+++ b/source/features/useful-not-found-page.tsx
@@ -54,7 +54,7 @@ async function getLatestCommitToFile(branch: string, filePath: string): Promise<
 		}
 	`);
 	const commit = repository.object?.history.nodes[0];
-	return commit.oid;
+	return commit?.oid;
 }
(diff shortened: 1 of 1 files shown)


fregante:
When is that the case?

kidonng:
Non-existing file

fregante:
Like, never existed? I suppose that the commit would exist only if the file existed at some point. 


If that's the case, this is good to go

kidonng:
> Like, never existed?

Right.

PR #5097, opened 2021-11-21
Fix `show-whitespace` hanging on very long lines

### LINKED ISSUE #5092: `show-whitespace` stucks with huge (but visible) file
### Please ensure:

### Include in this issue:

- [ ] Console errors, if any

### Description
A huge json, with browser showing "Refined Github is slowing down".

In this case, `element.textContent.length === 478130 && element.textContent.split(' ').length === 47609`.  That means `show-whitespace` will make (at least) 47609 * 2 = 95218 `<span>` and text nodes.

### Example URL
https://github.com/github/copilot.vim/blob/03a11aded57a78975effc87770ade4059c058dc7/copilot/dist/tokenizer.json

Note: Example URL can make your browser tab stuck, so you might want to double-check `show-whitespace` is disabled or not.

### Browser(s) used
Firefox Developer Edition 95 on Arch Linux x64

cheap-glitch:
Fixes #5092 

## Test URLs

https://github.com/github/copilot.vim/blob/03a11aded57a78975effc87770ade4059c058dc7/copilot/dist/tokenizer.json

## Screenshot

n/a

### DIFF
--- a/source/features/show-whitespace.tsx
+++ b/source/features/show-whitespace.tsx
@@ -15,7 +15,7 @@ function showWhiteSpacesOn(line: Element): void {
 	for (const [nodeIndex, textNode] of textNodesOnThisLine.entries()) {
 		// `textContent` reads must be cached #2737
 		let text = textNode.textContent!;
-		if (text.length > 10_000) { // #5092
+		if (text.length > 1000) { // #5092
 			continue;
 		}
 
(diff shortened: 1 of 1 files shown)


source/features/show-whitespace.tsx
```diff
@@ -15,6 +15,9 @@ function showWhiteSpacesOn(line: Element): void {
 	for (const [nodeIndex, textNode] of textNodesOnThisLine.entries()) {
 		// `textContent` reads must be cached #2737
 		let text = textNode.textContent!;
+		if (text.length > 10_000) { // #5092
+			continue;
+		}
```
cheap-glitch, line 20: This limit is very much arbitrary and probably too high.

Should we also check the number of text nodes on the line? Can there be a situation with a huge amount of very short text nodes, or is this impossible?

kidonng: 1000 would be enough IMO

fregante (approved):
After changing to 1000

PR #5098, opened 2021-11-21
Fix alignment of "Code of conduct" link in `clean-repo-sidebar`

cheap-glitch:
## Test URLs

https://github.com/sindresorhus/got

## Screenshot

**Before**

screenshot-1637498668

**After**

screenshot-1637498654

### DIFF
--- a/source/features/clean-repo-sidebar.tsx
+++ b/source/features/clean-repo-sidebar.tsx
@@ -19,6 +19,13 @@ async function cleanLicenseText(): Promise<void> {
 	}
 }
 
+async function cleanCodeOfConductText(): Promise<void> {
+	const codeOfConductLink = await elementReady('.Layout-sidebar .octicon-checklist');
+	if (codeOfConductLink) {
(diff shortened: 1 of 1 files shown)


cheap-glitch (commented):
Alternatively, we could stop removing the extra whitespace before the license link, and _add_ some to the repo age (easy) and the latest release tag (tricky).

fregante:
This looks ok. What we want is the path that requires the least code while being consistent (i.e. no hacks like "let's add a negative margin of 0.3em because that's the width of a space on my computer")

fregante:
Honestly I'd just remove it. The CoC is basically just a "don't be a dick" and that's something people do or don't do regardless of CoCs

cheap-glitch:
> Honestly I'd just remove it. The CoC is basically just a "don't be a dick" and that's something people do or don't do regardless of CoCs

I don't mind removing the new "Code of conduct" link in the repo sidebar, but some might be against it.

cc @sindresorhus  @refined-github/maintainers any thoughts?

fregante (approved):

fregante:
Ooops, I had approved my changes, which I was extracting into their own PR.

PR #5100, opened 2021-11-21
Fix `bugs-tab` where the emoji contains a dash

### LINKED ISSUE #5068: `bugs-tab` does not appear if the bug label emoji has a `-` in it
### Please ensure:

### Include in this issue:

- [ ] Console errors, if any

### Description

Hi and thanks a ton for making this addon!

I manage a few repositories where I renamed the default bug label and can't find a way to make the "Bugs" tab appear in this circumstance.

Here's the thing: I like emojis and therefore like to name this category `🐞 bug`
This may be easily addressable by extending the `supportedLabels` in `source/features/bugs-tab.tsx`

Hope you guys consider this type of request a bug and not a feature request, I wasn't too sure about the issue category for this one.

### Example URL

https://github.com/zdharma-continuum/zinit

### Browser(s) used

Firefox

yakov116:
Fixes #5068

## Test URLs
https://github.com/zdharma-continuum/zinit

## Screenshot
image

### DIFF
--- a/source/features/bugs-tab.tsx
+++ b/source/features/bugs-tab.tsx
@@ -12,7 +12,7 @@ import SearchQuery from '../github-helpers/search-query';
 import abbreviateNumber from '../helpers/abbreviate-number';
 import {highlightTab, unhighlightTab} from '../helpers/dom-utils';
 
-const supportedLabels = /^(bug|confirmed-bug|type:bug|kind:bug|:[\w_-]+:\s?bug)$/i;
+const supportedLabels = /^(bug|confirmed-bug|type:bug|kind:bug|(:[\w_-]+:|\p{Emoji})\s?bug)$/iu;
 const getBugLabelCacheKey = (): string => 'bugs-label:' + getRepo()!.nameWithOwner;
 const getBugLabel = async (): Promise<string | undefined> => cache.get<string>(getBugLabelCacheKey());
 const isBugLabel = (label: string): boolean => supportedLabels.test(label.replace(/\s/g, ''));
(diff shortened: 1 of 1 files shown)


yakov116:
@refined-github/maintainers we have another issue.

Looks like GitHub is responding the an actual emoji 😢 

```graphql
{"data":
	{"repository":
		{"labels":
			{"nodes":
				[{"name":"🐞 bug","issues":{"totalCount":2}}]
			}
		}
	}
}
```

source/features/bugs-tab.tsx
```diff
@@ -12,7 +12,7 @@ import SearchQuery from '../github-helpers/search-query';
 import abbreviateNumber from '../helpers/abbreviate-number';
 import {highlightTab, unhighlightTab} from '../helpers/dom-utils';
 
-const supportedLabels = /^(bug|confirmed-bug|type:bug|kind:bug|:\w+:bug)$/i;
+const supportedLabels = /^(bug|confirmed-bug|type:bug|kind:bug|(:[\w_-]+:|\p{Emoji})\s?bug)$/iu;
```
cheap-glitch, line 15:
```suggestion
const supportedLabels = /^(bug|confirmed-bug|type:bug|kind:bug|(:[\w-]+:|\p{Emoji})bug)$/iu;
```


 * `_` is already included in `\w`
 * we don't need to acount for a space, they're removed before testing

cheap-glitch (approved):
Successfully tested on Firefox :sparkles: (and the suggested changes too)

fregante (approved):
Ohhh yes. But _ is missing?

yakov116:
@fregante as per https://github.com/refined-github/refined-github/pull/5100#discussion_r753836190
> _ is already included in \w

fregante:
> Looks like GitHub is responding the an actual emoji 😢

Can you open an issue? If GitHub responds with an emoji when the label is :something:, does it also work the other way? That is: "label:😃" when in reality it should be `label:":smile:"`

yakov116:
? my regex fixed the issue

yakov116:
Unless your saying the github will sometimes respond as an emoji and sometimes not?

My theory is that if there is a space after the emoji they return the emoji.

PR #5072, opened 2021-11-15
Fix: `conversation-activity-filter` doesn't reset its state when navigating

cheap-glitch:
Fixes #4997 

## Test URLs

https://togithub.com/refined-github/refined-github/issues/4997

## Screenshot

https://user-images.githubusercontent.com/46634000/141775695-a611f3e4-bbf7-493a-b805-56f787db7fdd.mp4

PR #5102, opened 2021-11-22
Fix `conversation-activity-filter` sometimes failing

cheap-glitch:
Fixes bug introduced by #5072

### DIFF
--- a/source/features/conversation-activity-filter.tsx
+++ b/source/features/conversation-activity-filter.tsx
@@ -152,9 +152,9 @@ async function addWidget(header: string): Promise<void> {
 }
 
 async function init(): Promise<void> {
-	// Reset the dropdown state #4997
+	// Reset dropdowns state #4997
 	currentSetting = 'default';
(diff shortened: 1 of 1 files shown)


source/features/conversation-activity-filter.tsx
```diff
@@ -152,9 +152,9 @@ async function addWidget(header: string): Promise<void> {
 }
 
 async function init(): Promise<void> {
-	// Reset the dropdown state #4997
+	// Reset dropdowns state #4997
 	currentSetting = 'default';
-	select('.repository-content')!.classList.remove('rgh-conversation-activity-is-filtered');
+	(await elementReady('.repository-content'))!.classList.remove('rgh-conversation-activity-is-filtered');
```
cheap-glitch, line 157: Because of the `awaitDomReady: false`, the `select` was failing sometimes.

Alternatively the select could be placed at the end of the init (but the icon might flash in red before being reset). Or we could use a `deinit`.

fregante (approved):

PR #5084, opened 2021-11-18
Meta: Improve readme parsing

fregante:
Continues https://github.com/refined-github/refined-github/pull/5055

PR #5103, opened 2021-11-22
Fix the feature list in the options page

cheap-glitch:
Fixes bugs introduced by #5084 

## Test URLs

Options page

## Screenshot

**Before**

image

**After**

image

### DIFF
--- a/build/readme-parser.ts
+++ b/build/readme-parser.ts
@@ -42,12 +42,12 @@ function extractDataFromMatch(match: RegExpMatchArray): FeatureMeta {
 
 export function getFeaturesMeta(): FeatureMeta[] {
 	const readmeContent = readFileSync('readme.md', 'utf-8');
-	return [...readmeContent.matchAll(featureRegex)].map(match => extractDataFromMatch(match));
+	return [...readmeContent.matchAll(featureRegex)]
+		.map(match => extractDataFromMatch(match))
+		.sort((firstFeature, secondFeature) => firstFeature.id.localeCompare(secondFeature.id));
 }
 
-export function getFeatures(): FeatureID[] {
+export function getImportedFeatures(): FeatureID[] {
 	const contents = readFileSync('source/refined-github.ts', 'utf-8');
-	return [...contents.matchAll(/^import '\.\/features\/([^.]+)';/gm)]
-		.map(match => match[1] as FeatureID)
-		.sort();
+	return [...contents.matchAll(/^import '\.\/features\/([^.]+)';/gm)].map(match => match[1] as FeatureID);
 }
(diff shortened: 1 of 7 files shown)


source/options.tsx
```diff
@@ -238,8 +240,8 @@ async function generateDom(): Promise<void> {
 		select('#logHTTP-line')!.hidden = false;
 	}
 
-	// Add feature count. CSS-only features are added approximately
-	select('.features-header')!.append(` (${featuresMeta.length + 25})`);
+	// Add feature count
+	select('.features-header')!.append(` (${featuresMeta.length})`);
```
cheap-glitch, line 244: We have a more precise count now that CSS features are also parsed from the readme.

Maybe a `+ 10` or something is still warranted to account for the small features in `refined-github.css`?

build/readme-parser.ts
```diff
@@ -42,12 +42,12 @@ function extractDataFromMatch(match: RegExpMatchArray): FeatureMeta {
...
+		.sort((firstFeature, secondFeature) => firstFeature.id.localeCompare(secondFeature.id));
 }
 
-export function getFeatures(): FeatureID[] {
+export function getImportedFeatures(): FeatureID[] {
 	const contents = readFileSync('source/refined-github.ts', 'utf-8');
-	return [...contents.matchAll(/^import '\.\/features\/([^.]+)';/gm)]
-		.map(match => match[1] as FeatureID)
-		.sort();
+	return [...contents.matchAll(/^import '\.\/features\/([^.]+)';/gm)].map(match => match[1] as FeatureID);
```
cheap-glitch, line 52: The bugs were caused by the decoupling of `getFeatures()` and `getFeaturesMeta()`. Now the first one doesn't really need to be sorted.

cheap-glitch (commented):
* Sort the features in the options page
 * Don't display CSS-only features in the options page
 * Rename `getFeatures` and `featureList` to `importedFeatures` for clarity

source/options.tsx
fregante: plz no https://github.com/refined-github/refined-github/issues/4951#issuecomment-953289103

build/readme-parser.ts
fregante: Ohhh _that's_ why they were coupled, 👍 I suppose this is happening because we added some CSS features to the readme? I'm confused about `refined-github.css` appearing in the options though 🤔

fregante: Let's preserve the sorting in both functions to avoid issues

fregante (approved):
Untested but the changes make sense, except I'd restore `.sort` in the `importedFeatures` as well

build/readme-parser.ts
cheap-glitch: > I'm confused about `refined-github.css` appearing in the options though

It's in the readme: https://github.com/refined-github/refined-github/blob/047a6c43b7fae828b95a818194332835bfb71194/readme.md?plain=1#L204

PR #5104, opened 2021-11-22
Disable the checkbox of hotfixed features

### LINKED ISSUE #4974: Hotfix-disabled features should display as disabled

cheap-glitch:
Continues #4987
Resolves #4974 

## Test URLs

Options page

## Screenshot

screenshot-1637604493

### DIFF
--- a/source/options.css
+++ b/source/options.css
@@ -133,6 +133,10 @@ li[data-validation] {
 	color: var(--rgh-red);
 }
 
+.feature-checkbox:disabled + .info > *:not(.hotfix-notice) {
+	opacity: 0.5;
+}
+
 .sr-only {
 	display: none;
 }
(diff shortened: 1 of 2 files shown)


fregante (approved):
🚀 

Maybe one day we can use something more sane to build this list instead of using `select` 🥲 (e.g. preact/svelte)

PR #5106, opened 2021-11-22
Fix `show-whitespace` on newly sent suggestions

### LINKED ISSUE #2852: `show-whitespace` doesn’t work on newly sent suggestions

cheap-glitch:
Fixes #2852 

## Test URLs

https://togithub.com/refined-github/refined-github/pull/5103/files

## Screenshot

https://user-images.githubusercontent.com/46634000/142915220-243a89cb-9f40-411b-a5da-59119c3287ef.mp4


source/github-events/on-new-review.ts
```diff
@@ -0,0 +1,13 @@
...
+function callbackWrapper(callback: VoidFunction): (event: CustomEvent) => void {
+	return function (event) {
+		if (/^pull request #\d+ updated$/.test(event.detail.data.reason)) {
+			callback();
+		}
+	};
+}
+
+export default function onNewReviewComment(callback: VoidFunction): void {
+	delegate(document, '.js-pull-refresh-on-pjax', 'socket:message', callbackWrapper(callback), {capture: true});
```
cheap-glitch, line 12: This triggers on other events besides new reviews, e.g. when editing or deleting reviews.

In the first case it's actually a good thing I guess, because the contents of the embedded snippets might have been changed and/or added.

source/github-events/on-new-review.ts
```diff
@@ -0,0 +1,13 @@
+import delegate from 'delegate-it';
+
+function callbackWrapper(callback: VoidFunction): (event: CustomEvent) => void {
+	return function (event) {
+		if (/^pull request #\d+ updated$/.test(event.detail.data.reason)) {
```
cheap-glitch, line 5: Not sure if the test is really needed, it seems all the messages are the same (with just the id changing).

cheap-glitch (commented):
Attempt at implementing a listener for new review comments.

source/github-events/on-new-review.ts
```diff
@@ -0,0 +1,13 @@
+import delegate from 'delegate-it';
+
+function callbackWrapper(callback: VoidFunction): (event: CustomEvent) => void {
```
fregante, line 3: If you don't memoize this call, then `delegate` will add a new listener every time `onNewReviewComment` is called. Event listeners are natively deduplicated if the same exact function is passed to it. Something like:
```suggestion
const callbackWrapper = mem((callback: VoidFunction): (event: CustomEvent) => void => {
```

source/github-events/on-new-review.ts
fregante: `socket:message` can't be used I think. The message doesn't always equal DOM change, which sometimes happens later. We had this issue a few years ago when implementing `onNewComments`, not sure if it applies here.

If it works… it's fine for now.

source/github-events/on-new-review.ts
fregante: The issue is that this event might also be dispatched when a regular comment comes in, so it would effectively duplicate `onNewComment` as well.

cheap-glitch: > The issue is that this event might also be dispatched when a regular comment comes in

It is, but not from the same source, so the delegation ignores it.

The issue is that it is dispatched three times when sending a quick review (with "Add single review comment" button), and afaict the three events are virtually indistinguishable :grimacing:

fregante: That doesn't sound right. What's happening though is that we have a simple loop with deduplication in `init`. So even if it's called twice, no duplication happens.

It may be ok, but I'm still wary of this event and I'd probably merge it back into this file so we know it's not safe for general use. A comment like this should also be inlined.

fregante (approved):

kidonng:
Thanks for solving an oldest bug 🔥

PR #5112, opened 2021-11-27
Show message during API call

fregante:
- Fixes https://github.com/refined-github/refined-github/issues/4828
- Replaces https://github.com/refined-github/refined-github/pull/4946

## Details

I looked at all the features that used a v3 call with custom `method`. Many of them already handle errors inline or have an inline loading icon. Only some of them could be replaced to show a toast only when the action was not successful. Even here it's kind of a stretch.

https://user-images.githubusercontent.com/1402241/143673554-bc38445c-06fa-488f-ae6e-68679fc8ebc0.mov

## Test URLs

- any PR or issue with labels. Or this PR

## Future

- [ ] Adjust `showToast` to also accept a raw error: `showToast(new Error('Something’s up'))`
- [ ] Drop mentions of "bulk action" in showToast’s defaults

## Rejected

I tried converting `convert-release-to-draft` but it makes less sense than before. 

https://user-images.githubusercontent.com/1402241/143673445-77dbd45e-9f14-4c15-b296-a60ed8569a7b.mov

### DIFF
--- a/source/features/quick-label-removal.tsx
+++ b/source/features/quick-label-removal.tsx
@@ -9,8 +9,8 @@ import * as pageDetect from 'github-url-detection';
 
 import features from '.';
 import * as api from '../github-helpers/api';
-import {getConversationNumber} from '../github-helpers';
 import showToast from '../github-helpers/toast';
+import {getConversationNumber} from '../github-helpers';
 
 const canNotEditLabels = onetime((): boolean => !select.exists('.label-select-menu .octicon-gear'));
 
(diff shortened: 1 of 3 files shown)


source/features/quick-label-removal.tsx
```diff
@@ -10,6 +10,7 @@ import * as pageDetect from 'github-url-detection';
 import features from '.';
 import * as api from '../github-helpers/api';
 import {getConversationNumber} from '../github-helpers';
+import showToast from '../github-helpers/toast';
```
yakov116, line 13: Lint

fregante:
I don’t like it. It should hide the label immediately and only call `showToast(new Error)` if it fails

PR #5115, opened 2021-11-29
Meta: Add link checker

fregante:
- Blocked by https://github.com/lycheeverse/lychee-action/issues/60

### DIFF
--- a/.github/workflows/maintenance.yml
+++ b/.github/workflows/maintenance.yml
@@ -21,7 +21,7 @@ jobs:
             --no-progress
             --exclude-file "safari/Shared (App)/Base.lproj/Main.html"
             --exclude-file "distribution/options.html"
-             **/*.md **/*.html **/*.css **/*.ts **/*.tsx
+            -- '**/*.md' '**/*.html' '**/*.css' '**/*.ts' '**/*.tsx'
             
         env:
           GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
--- a/.github/workflows/maintenance.yml
+++ b/.github/workflows/maintenance.yml
@@ -16,7 +16,13 @@ jobs:
       - name: Link Checker
         uses: lycheeverse/lychee-action@v1.1.1
         with:
-          args: --verbose --no-progress **/*.md **/*.html
+          args: >-
+            --verbose
+            --no-progress
+            --exclude-file "safari/Shared (App)/Base.lproj/Main.html"
+            --exclude-file "distribution/options.html"
+             **/*.md **/*.html **/*.css **/*.ts **/*.tsx
+            
         env:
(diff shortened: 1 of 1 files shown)


kidonng:
Most links in this repository points to GitHub which probably lives longer than me, but welcoming though. (Also, nice choice of lychee)

fregante:
Yeah in reality we probably don't need this. It'd be good if it also threw errors in case of redirects and we added it to JS and CSS files. I haven't added those because I was testing https://github.com/lycheeverse/lychee-action/issues/60

mre:
> Error: Compiled regex exceeds size limit of 10485760 bytes.

But why?

fregante:
I suppose it's the many globs? I don't know what's inside the tool

mre:
Ah okay I see what's going on. It's a little funny. 😆 
You use `--exclude-file`, which is not what you think it means.
It's meant as a newline-separated list of regular expressions for filtering out links.
So lychee interpreted that file as one huge regex. 😆 
You meant to exclude the file from checking completely.

I think the wording is a bit misleading. 
Your interpretation makes a lot of sense to me so I created 
https://github.com/lycheeverse/lychee/issues/418 
to handle that in the future.

In the meantime there is no way to exclude the file -- just the links within with a regex.
This will be possible with either https://github.com/lycheeverse/lychee/issues/418 or https://github.com/lycheeverse/lychee/issues/284.

mre:
On top of that your repo also triggers https://github.com/seanmonstar/reqwest/issues/668 in reqwest. 😆 

```
refined-github ❯❯❯ lychee --verbose --no-progress -- '**/*.md' '**/*.html' '**/*.css' '**/*.ts' '**/*.tsx'
thread 'tokio-runtime-worker' panicked at 'a parsed Url should always be a valid Uri: InvalidUri(InvalidUriChar)', /Users/mendler/.cargo/registry/src/github.com-1ecc6299db9ec823/reqwest-0.11.7/src/into_url.rs:70:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'main' panicked at 'Cannot send request: SendError(Ok(Request { uri: Uri { url: Url { scheme: "https", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("${repository.textcontent!.trim()}")), port: None, path: "/", query: None, fragment: None } }, source: FsPath("source/features/link-to-github-io.tsx") }))', lychee-bin/src/commands/check.rs:80:14
```

Created a PR to fix that: https://github.com/seanmonstar/reqwest/pull/1399. 🤞 
It helps find bugs when testing lychee on bigger repos, so thanks for that. 😊

fregante:
> In the meantime there is no way to exclude the file

Gotcha, thanks! I think I'll pass for now then.

PR #5124, opened 2021-11-29
Drop `hide-obvious-tooltips`

kidonng:
First, this is currently broken by #4041 since [:is() does not select pseudo-elements](https://developer.mozilla.org/en-US/docs/Web/CSS/:is#is_does_not_select_pseudo-elements). But rather than fixing it, I think maybe it's not worth keeping anymore.

- This has stopped working for almost 9 months, yet no complaint. Not as long as #4891 but still pretty lengthy. You may argue it's because it's so subtle that we don't know it, but well, if it's subtle enough to not have it, then probably we don't need it.
- Most items in the list don't make sense anymore, or it's just I don't know where are they:
  
  ```css
  [aria-label^='View the whole file'], /* ❓ If I guessed correctly, this seems to be in submenu now and doesn't have a tooltip */
  [aria-label^='Change this file'], /* ❓ Dunno */
  [aria-label='Delete this file'],  /* 🤔 Valid, but most of the time you just click the button and the tooltip doesn't even have enough time to display */
  [aria-label='Edit this file'],  /* 🤔 Same as above */
  [aria-label='Edit comment'], /* ❓ If this refers to the `quick-comment-edit` button, it does not have the tooltip class */
  .notification-indicator /* 🤔 Same as above */
  ```


fregante:
I don't feel strongly about this

kidonng:
It would be hard to decide what is "obvious". For example, the trash icon of deletion button _is_ a little bit too abstract; or, many other tooltips in #1087 (which added this feature) may also be considered obvious. As it currently stands it does not provide much value. 

This is a feature from "the ancient times" (early 2018) and IMO it doesn't age well. We have too many features to add/maintain and this is the least thing to take up space in the codebase.

PR #5125, opened 2021-11-29
Hide badges in repo headers and pinned repo cards

### LINKED ISSUE #4767: Hide the "Public" repo badge
### Description

Hide the "Public" repo badge in profile page, repo list and repo page. It seems like it's recently added to GitHub but I think it's a little distractive and I prefer old behaviour, which is only shows private and archived badge.

### Screenshot

- [Profile page](https://github.com/sindresorhus)

- [Repo list](https://github.com/sindresorhus?tab=repositories&type=source)

- [Repo page](https://github.com/sindresorhus/refined-github)

### Example URL

- Profile page: https://github.com/sindresorhus
- Repo list: https://github.com/sindresorhus?tab=repositories&type=source
- Repo page: https://github.com/sindresorhus/refined-github

cheap-glitch:
Resolves #4767 

## Test URLs

 * [Repo home](https://github.com/refined-github/refined-github)
 * [User pinned repos](https://github.com/fregante)
 * [User repo list](https://github.com/fregante?tab=repositories)
 * [Organization pinned repos & repo list](https://github.com/refined-github?type=source)

## Screenshot

<table>
<tr>
	<td>
	<td>Before
	<td>After
<tr>
	<td>Repo title
	<td>

screenshot-1638202650
	<td>

screenshot-1638202662
<tr>
	<td>Pinned repos
	<td>

screenshot-1638202618
	<td>

screenshot-1638202551
</table>

### DIFF
--- a/source/features/ci-link.css
+++ b/source/features/ci-link.css
@@ -5,8 +5,3 @@
 .rgh-ci-link summary .octicon {
 	vertical-align: 0;
 }
-
-/* Adds space between `ci-link` and "Private" label, if present */
-.rgh-ci-link .Label + .commit-build-statuses {
-	margin-left: 8px;
-}
--- a/source/refined-github.css
+++ b/source/refined-github.css
@@ -166,7 +166,7 @@ pr-branches
 	max-width: 100%;
 }
 
-/* Hide repo badges ("Public", "Template", etc.) on repo pages and pinned repo cards #4767 */
+/* Hide repo badges ("Public", "Template", etc.) in the repo header and pinned repo cards #4767 */
 :is(#repository-container-header h1, .pinned-item-list-item-content) .Label {
 	display: none !important;
 }
--- a/source/refined-github.css
+++ b/source/refined-github.css
@@ -165,3 +165,9 @@ pr-branches
 	width: calc(72ch + 26px);
 	max-width: 100%;
 }
+
+/* Hide all repo badges ("Public", "Template", etc.) #4767 */
+:is(#repository-container-header h1, .pinned-item-list-item-content) .Label,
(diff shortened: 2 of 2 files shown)


fregante:
My arguments only applied to the repository title. Public/private/fork does not appear anywhere else in the repo list for example. I think we can't drop _those_

kidonng:
GitHub has these classes for repositories in a list: `.public` / `.private`, `.source` / `.fork`, `.archived`. Template repositories don't have a corresponding class though.

kidonng:
I just had an idea... what if we turn the label in list items into icons, just like the one in repo header? It's more consistent and cleaner than just doing nothing.

fregante:
I don’t think it's worth the extra code that that'd require. Textual labels aren't *that* bad, it's just that in some cases they show duplicate information.

fregante (approved):
_okaaay let's [goooo](https://www.tiktok.com/@xkypher_21/video/7032357929550368046)_

WEGFan:
Will this have a feature switch instead of applied globally?

cheap-glitch:
> Will this have a feature switch instead of applied globally?

We can easily make it togglable, I personally don't mind

nick-f:
I'd be keen for a toggle as well. There's not an easy way to distinguish between private and internal now.

I had a good couple of minutes of panic when that pill was missing and a repo I thought was meant to be private was public. Thankfully it was private, as intended, and I was able to learn about the repo icon next to the title 😄

cheap-glitch:
> > Will this have a feature switch instead of applied globally?
> 
> We can easily make it togglable, I personally don't mind

@fregante wdyt?

PR #5113, opened 2021-11-28
Lint

kidonng:
A long-overdue thanksgiving gift 🦃️

## Screenshot

- Fix `file-finder-buffer` left margin

Before:

屏幕录制2021-11-29 20 25 06-0001

After:

屏幕录制2021-11-29 20 25 35-0002

Issue #451, opened 2017-05-28
Remove "Marketplace" nav item


jgierer12:
I think its position in the top nav is way too prominent. I'd move it into the profile drop-down, underneath "Explore"

sindresorhus:
I was thinking the same thing.

hkdobrev:
👍 Though, I'd recommend waiting a bit so people could explore it a bit.

sindresorhus:
Sure

bfred-it:
It showed up again for me. I'm just gonna add a uBlock filter:

```
github.com##.js-marketplace-callout-container
```

pablopunk:
You guys think it's time for this? I would like this to be implemented, there are too many buttons in the nav bar

hkdobrev:
@pablopunk Go for it!

Issue #153, opened 2016-04-04
Show "Trending" in the top menu


sindresorhus:
From https://github.com/isaacs/github/issues/245.

Currently the link to your starred repos is hidden in the avatar dropdown. Might be nice to add it to the top menu. Not entirely sure about this, so use "Reactions" to vote whether people want this or not.

<img width="743" alt="napkin 04-04-16 3 47 15 pm" src="https://cloud.githubusercontent.com/assets/170270/14242908/a06c21f0-fa7c-11e5-975c-a6b62539e8c3.png">

sindresorhus:
// @necolas

paulmolluzzo:
This isn't a lot of work, so why not?

sindresorhus:
@paulmolluzzo It's not about the amount of work, rather, is this something you would use? Is it worth adding more stuff to the UI for this?

paulmolluzzo:
@sindresorhus I don't often go to my starred repos, but having the link would be helpful. For others it might be more valuable.

unluisco:
the same for Explore?

sindresorhus:
@unluisco What tab do you want in Explore? Trending?

unluisco:
@sindresorhus yes, always trending

kevva:
Personally, I rarely go to my starred repos either (besides wanting to revisit something I forgot but I know starred). Or at least, not enough for having it in my top menu. I could see the value in having Explore though.

sindresorhus:
Ok, so maybe we add `Trending` instead? You can easily go to your Stars from there.

sindresorhus:
Alright. PR welcome for adding `Trending`.

unluisco:
What happened with Trending? 😢

PR #5126, opened 2021-11-29
`deprioritize-marketplace-link` to footer

kidonng:
Extracted from #5113

I never get his feature, since I use dropdown _way more_ than the header. The extra link suddenly appears in the dropdown, which looks like this feature is making it more noticable. Not to mention the link feels out of place in the dropdown (where should we put it?)

So I just find another nice place to put it: the footer. It's always on the bottom right corner for convenience. 

~~The only problem is that `infinite-scroll` effectively makes it inaccessible, however that's another issue.~~  Fixed by moving the footer to the right.

## Test URLs

- https://github.com/
- Here

## Screenshot

On dashboard:

image

On other pages:

image

### DIFF
--- a/source/features/deprioritize-marketplace-link.tsx
+++ b/source/features/deprioritize-marketplace-link.tsx
@@ -1,30 +1,28 @@
 import React from 'dom-chef';
 import select from 'select-dom';
-import onetime from 'onetime';
-import elementReady from 'element-ready';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
-import onProfileDropdownLoad from '../github-events/on-profile-dropdown-load';
 
 async function init(): Promise<void> {
-	const marketplaceLink = await elementReady('.Header-link[href="/marketplace"]', {waitForChildren: false});
-	if (marketplaceLink) { // On GHE it can be disabled #3725
-		// The link seems to have an additional wrapper that other links don't have https://i.imgur.com/KV9rtSq.png
-		marketplaceLink.closest('.border-top, .mr-3')!.remove();
-
-		await onProfileDropdownLoad();
-		select.last('.header-nav-current-user ~ .dropdown-divider')!.before(
-			<div className="dropdown-divider"/>,
-			<a role="menuitem" className="dropdown-item" href="/marketplace">Marketplace</a>,
-		);
+	const marketplaceLink = select('.Header-link[href="/marketplace"]');
+	// On GHE it can be disabled #3725
+	if (!marketplaceLink) {
+		return;
 	}
+
+	// The link seems to have an additional wrapper that other links don't have https://i.imgur.com/KV9rtSq.png
+	marketplaceLink.closest('.border-top, .mr-3')!.remove();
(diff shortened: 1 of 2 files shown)


source/features/deprioritize-marketplace-link.tsx
```diff
@@ -1,30 +1,33 @@
...
 	}
+
+	select.last('.footer ul')!.append(item);
 }
 
 void features.add(import.meta.url, {
 	exclude: [
 		pageDetect.isGist,
 	],
-	awaitDomReady: false,
```
fregante, line 28: It must still use this, or else it might disappear seconds later, especially on long pages.

You can `await dowloaded` after you `.remove()` it

source/features/deprioritize-marketplace-link.tsx
```diff
@@ -1,30 +1,33 @@
...
-	init: onetime(init),
+	init,
```
fregante, line 32: This is right because the footer needs to be updated every time, but the header remains unchanged across ajaxed loads, so after the first run, `init` will just return on line 11 I think

fregante:
My previous review is still open. Also perhaps I'd just drop the feature if it doesn't cause responsiveness issues. I use the marketplace link to find new actions

source/features/deprioritize-marketplace-link.tsx
kidonng: Do you mean `onetime` should be kept?

kidonng:
> if it doesn't cause responsiveness issues.

Not sure what you mean, does it mean this feature exists to keep `trending-menu-item` from being jumpy?

> I use the marketplace link to find new actions and rarely use the explore 😅 

I don't really use both of them, so I don't have a say. But it sounds interesting, so I found the issues that asks for these features (which, to my surprise, are all from _ancient_ times).

#153 asked for a link to `/stars` in the header first, but the conclusion was to add a trending item instead, which becomes `trending-menu-item` now. At that time there were three links in the header: PR, Issues, Gist. Explore link was in profile dropdown.

#451 asked to remove the marketplace link, and the link has been moved to the dropdown, underneath Explore. This becomes `deprioritize-marketplace-link`. At that time there were four links in the header: PR, Issues, Marketplace, Gist.

Now the link to Gist has been taken place by the explore. The Trending link points to a subsection of Explore.


If we are to drop `deprioritize-marketplace-link`, I suggest change the Explore link to Trending, instead of adding a new one. That way we avoided the responsiveness issue while keep the feature going.

fregante:
> > if it doesn't cause responsiveness issues.
> 
> Not sure what you mean, does it mean this feature exists to keep [`trending-menu-item`](/refined-github/refined-github/blob/main/source/features/trending-menu-item.tsx) from being jumpy?

There are 4 links now so it's fine, but dropping this feature would mean showing 5 links. If you resize the window to around 600px (example), can all the _5_ links stay without wrapping? 



> I don't really use both of them, so I don't have a say

I used to use explore more, now not anymore. I know Sindre wanted that. Let's wait a bit

PR #879, opened 2017-12-16
Clean up the global footer

sindresorhus:
## Before

screen shot 2017-12-17 at 00 47 39

## After

screen shot 2017-12-17 at 00 47 09

PR #5128, opened 2021-11-29
Move footer to right side when enabling `infinite-scroll`

kidonng:
As mentioned in #5126, `infinite-scroll` effectively makes footer inaccessible, not only affecting that PR but also making us lose access to lots of useful links.

## Test URLs

https://github.com/

## Screenshot

<details>

image

(Similiar to Twitter sidebar:)

image

</details

### DIFF
--- a/source/features/clean-footer.css
+++ b/source/features/clean-footer.css
@@ -1,12 +1,12 @@
-body > .footer {
+.footer {
 	transition: opacity 0.2s ease;
 }
 
-body > .footer:not(:hover) {
+.footer:not(:hover) {
 	opacity: 30%;
 }
 
-body > .footer > div > ul:first-of-type li:first-of-type { /* The `© 2017 GitHub, Inc.` text */
+.footer > div > ul:first-of-type li:first-of-type { /* The `© 2017 GitHub, Inc.` text */
 	display: none;
 }
 
@@ -14,7 +14,7 @@ body > .footer > div > ul:first-of-type li:first-of-type { /* The `© 2017 GitHu
 	display: none !important;
 }
 
-body > .footer li a,
-body > .footer li .btn-link {
+.footer li a,
+.footer li .btn-link {
 	color: var(--color-text-disabled, #666);
 }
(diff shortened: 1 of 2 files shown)


yakov116:
GIF?

kidonng:
What do you mean? This does not make the sidebar sticky.

yakov116:
Can you show a GIF or video of this in action?

kidonng:
https://user-images.githubusercontent.com/44045911/143940376-a0d3a599-0afc-4a9f-bedd-2c4918fc50b3.mov

source/features/clean-footer.css
```diff
@@ -1,20 +1,20 @@
-body > .footer {
+.footer {
```
fregante, line 1: If looking at the blame there was no reason for `body` to exist, then it's ok. Basically the concern should be whether `.footer` is used elsewhere, for unrelated widgets. This is especially a problem because this CSS is global.

fregante (approved):

source/features/clean-footer.css
kidonng: `body >` has been there since it was added in #879. I highly doubt GitHub will use this class for anything else.

source/features/infinite-scroll.tsx
```diff
@@ -33,6 +34,22 @@ function init(): void {
 			inView.observe(button);
 		},
 	});
+
+	// Use cloneNode to keep the original ones for responsive layout
```
kidonng, line 38: Turns out this is a wise choice 😁 it works with #5126 flawlessly without changing a single line.

fregante:
Lol who added "Did you know?" https://github.com/refined-github/refined-github/releases/tag/21.12.1 🙌 

I didn't know!

kidonng:
> Lol who added "Did you know?" https://github.com/refined-github/refined-github/releases/tag/21.12.1 🙌 
> 
> I didn't know! 

It's me 😂 I don't want to keep my avatar in the screenshot so...

PR #5129, opened 2021-11-29
Fix `quick-pr-diff-options` on compare pages

cheap-glitch:
## Test URLs

https://github.com/refined-github/refined-github/compare/add-multiple-prs-dropdown

## Screenshot

image

### DIFF
--- a/source/features/quick-pr-diff-options.tsx
+++ b/source/features/quick-pr-diff-options.tsx
@@ -112,11 +112,12 @@ function initCommitAndCompare(): false | void {
 	);
 }
 
-const shortcuts: {
+const shortcuts = {
 	'd w': 'Show/hide whitespaces in diffs',
 };
 
 void features.add(import.meta.url, {
+	shortcuts,
 	include: [
 		pageDetect.isPRFiles,
 		pageDetect.isPRCommit,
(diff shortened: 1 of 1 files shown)


source/features/quick-pr-diff-options.tsx
```diff
@@ -127,8 +128,19 @@ void features.add(import.meta.url, {
...
+	},
+	init: initCommitAndCompare,
+}, {
+	include: [
 		pageDetect.isCompare,
 	],
+	additionalListeners: [
+		onDiffFileLoad,
+	],
+	onlyAdditionalListeners: true,
```
cheap-glitch, line 143: The diff summary and its "toc" are lazy-loaded now.

source/features/quick-pr-diff-options.tsx
```diff
@@ -127,8 +128,19 @@ void features.add(import.meta.url, {
 }, {
 	include: [
 		pageDetect.isSingleCommit,
+	],
+	shortcuts: {
+		'd w': 'Show/hide whitespaces in diffs',
+	},
```
cheap-glitch, line 134: Do we have to copy the shortcut info in every init object? Seems like having it once would be enough.

fregante: The issue is that `shortcuts` is only considered if the feature applies to the current page. This key is used to generate the `?` view

fregante (approved):

PR #5132, opened 2021-11-30
Show error toast when `quick-label-removal` fails

### LINKED ISSUE #4828: API actions might fail silently

cheap-glitch:
Fixes #4828 

## Test URLs

Revoke your API token and try to remove a label on this page.

## Screenshot

https://user-images.githubusercontent.com/46634000/144035109-e9dc3fc6-d729-48b7-bae6-621cf4630cbb.mp4

### DIFF
--- a/source/features/quick-label-removal.tsx
+++ b/source/features/quick-label-removal.tsx
@@ -9,6 +9,7 @@ import * as pageDetect from 'github-url-detection';
 
 import features from '.';
 import * as api from '../github-helpers/api';
+import showToast from '../github-helpers/toast';
 import {getConversationNumber} from '../github-helpers';
 
 const canNotEditLabels = onetime((): boolean => !select.exists('.label-select-menu .octicon-gear'));
@@ -17,13 +18,23 @@ async function removeLabelButtonClickHandler(event: delegate.Event<MouseEvent, H
 	event.preventDefault();
 
 	const removeLabelButton = event.delegateTarget;
-
 	removeLabelButton.disabled = true;
-	await api.v3(`issues/${getConversationNumber()!}/labels/${removeLabelButton.dataset.name!}`, {
-		method: 'DELETE',
-	});
 
-	removeLabelButton.closest('a')!.remove();
+	const label = removeLabelButton.closest('a')!;
+	label.hidden = true;
+	try {
+		await api.v3(`issues/${getConversationNumber()!}/labels/${removeLabelButton.dataset.name!}`, {
+			method: 'DELETE',
+		});
+	} catch (error: unknown) {
+		void showToast(error as Error);
+		removeLabelButton.disabled = false;
+		removeLabelButton.blur();
+		label.hidden = false;
+		return;
+	}
+
+	label.remove();
 }
 
 async function init(): Promise<void> {
--- a/source/features/quick-label-removal.tsx
+++ b/source/features/quick-label-removal.tsx
@@ -18,9 +18,8 @@ async function removeLabelButtonClickHandler(event: delegate.Event<MouseEvent, H
 	event.preventDefault();
 
 	const removeLabelButton = event.delegateTarget;
-	removeLabelButton.disabled = true;
-
 	const label = removeLabelButton.closest('a')!;
+
 	label.hidden = true;
 	try {
 		await api.v3(`issues/${getConversationNumber()!}/labels/${removeLabelButton.dataset.name!}`, {
@@ -28,7 +27,6 @@ async function removeLabelButtonClickHandler(event: delegate.Event<MouseEvent, H
 		});
 	} catch (error: unknown) {
 		void showToast(error as Error);
-		removeLabelButton.disabled = false;
 		removeLabelButton.blur();
 		label.hidden = false;
 		return;
--- a/source/github-helpers/toast.tsx
+++ b/source/github-helpers/toast.tsx
@@ -12,14 +12,14 @@ export function ToastSpinner(): JSX.Element {
 }
 
 type ProgressCallback = (message: string) => void;
-type Task<T> = (progress?: ProgressCallback) => Promise<T>;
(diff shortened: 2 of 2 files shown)


source/github-helpers/toast.tsx
```diff
@@ -12,14 +12,14 @@ export function ToastSpinner(): JSX.Element {
...
-	task: TTask,
+type Task<T> = (progress?: ProgressCallback) => Promise<T>;
+export default async function showToast<TaskOrError extends Task<any> | Error>(
+	task: TaskOrError,
 	{
 		message = 'Bulk actions currently being processed.',
 		doneMessage = 'Bulk action processing complete.',
 	} = {},
-): Promise<ReturnType<TTask>> {
+): Promise<[TaskOrError] extends [Task<infer R>] ? R : void> {
```
cheap-glitch, line 22: Since we never use the return value of `showToast`, all those types could be simplified.

source/github-helpers/toast.tsx
```diff
@@ -40,6 +40,10 @@ export default async function showToast<TTask extends Task>(
 	await delay(30); // Without this, the Toast doesn't appear in time
 
 	try {
+		if (task instanceof Error) {
+			throw task;
+		}
+
```
cheap-glitch, line 46: Not sure if that's acceptable, but it's short :sweat_smile:

source/features/quick-label-removal.tsx
```diff
@@ -17,13 +18,23 @@ async function removeLabelButtonClickHandler(event: delegate.Event<MouseEvent, H
 	event.preventDefault();
 
 	const removeLabelButton = event.delegateTarget;
-
 	removeLabelButton.disabled = true;
```
fregante, line 21: We're now removing the whole label, so no need to disable it

source/github-helpers/toast.tsx
```diff
@@ -12,14 +12,14 @@ export function ToastSpinner(): JSX.Element {
 }
 
 type ProgressCallback = (message: string) => void;
-type Task = (progress?: ProgressCallback) => Promise<unknown>;
-export default async function showToast<TTask extends Task>(
-	task: TTask,
+type Task<T> = (progress?: ProgressCallback) => Promise<T>;
+export default async function showToast<TaskOrError extends Task<any> | Error>(
+	task: TaskOrError,
```
fregante, line 17: I think it should look like this:

```suggestion
type Task = (progress?: ProgressCallback) => Promise<unknown>;
export default async function showToast<TTask extends Task>(
	task: TTask | Error,
```

And then the return `void` as you mentioned.

fregante (approved):

fregante (approved):
Name it, read it, tune it, print it, scan it, send it, fax, rename it
Touch it, bring it, pay it, watch it, turn it, leave it, stop, format it

source/github-helpers/toast.tsx
```diff
@@ -40,11 +40,15 @@ export default async function showToast<TTask extends Task>(
...
+		if (task instanceof Error) {
+			throw task;
+		}
+
+		await task(updateToast);
 		toast.classList.replace('Toast--loading', 'Toast--success');
 		updateToast(doneMessage);
 		iconWrapper.firstChild!.replaceWith(<CheckIcon/>);
-		return result;
+		return;
```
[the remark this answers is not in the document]
cheap-glitch: I left the `return` because I assumed it was there to "avoid" the `finally` block. Is it executed regardless?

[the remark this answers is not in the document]
busches: If you're in a try block then finally will be executed no matter how you leave it.

cheap-glitch:
We've finally dropped to 40 bugs!  (๑•̀ㅂ•́)و✧

Let's see how long that lasts…

Issue #4952, opened 2021-10-21
Dim more bots


kidonng:
### Description

Extend `dim-bots` to include more bots:

- @actions-user: belongs to GitHub. it seems to be used by GitHub's example initially, and now a [widely used](https://github.com/search?o=desc&q=author%3Aactions-user&s=committer-date&type=Commits) placeholder besides `github-actions[bot]`.
- @web-flow: belongs to GitHub. See its bio
- @ImgBotApp: [usage](https://github.com/search?o=desc&q=author%3AImgBotApp&s=committer-date&type=Commits)
- @bors (@rust-lang bot user) and [bors](https://github.com/apps/bors) (GitHub app). Note these are "merge bots" in the sense that the duty of them is to merge commits, so they are already dimmed out by `mark-merge-commits-in-list`. But I think it makes sense enough to fold them via `dim-bots`.
- @Octomerger: as can be seen in https://github.com/github/docs

Adding more popular bots from https://github.com/marketplace is welcome.

### Example URL

N/A

cheap-glitch:
Maybe @snyk-bot from https://snyk.io? (It's the successor of Greenkeeper).

kidonng:
Reopening due to https://github.com/refined-github/refined-github/pull/5192#discussion_r767273718

fregante:
Just push a commit to make that single change 😃

PR #5139, opened 2021-12-02
Add more bots to `dim-bots` commit author list

fluteds:
Adds more bots to [`dim-bots.tsx`](https://github.com/refined-github/refined-github/blob/main/source/features/dim-bots.tsx). Contributes to issue #4952.

PR targets the following bots:

- @actions-user
- @web-flow
- Rust Bots 
	- @bors 
	- @rust-lang
	- @rust-highfive
- @snyk-bot
- @imgbot
- @octomerger

I can add more bots to the list if they aren't already dimmed.


source/features/dim-bots.tsx
```diff
@@ -9,6 +9,14 @@ const commitSelector = [
...
 	'.commit-author[href$="renovate-bot"]',
 	'.commit-author[href$="scala-steward"]',
+	'.commit-author[href$="actions-user"]',
+	'.commit-author[href$="web-flow"]',
+	'.commit-author[href$="bors"]',
+	'.commit-author[href$="rust-lang"]',
+	'.commit-author[href$="rust-highfive"]',
+	'.commit-author[href$="snyk-bot"]',
+	'.commit-author[href$="imgbot"]',
+	'.commit-author[href$="octomerger"]',
```
kidonng, line 19: This will be better as a map like <code>['bot1', 'bot2'].map(bot => \`.commit-author[href$="${bot}"]\`)</code>

source/features/dim-bots.tsx
```diff
@@ -9,6 +9,14 @@ const commitSelector = [
 	'.commit-author[href$="%5Bbot%5D"]',
 	'.commit-author[href$="renovate-bot"]',
 	'.commit-author[href$="scala-steward"]',
+	'.commit-author[href$="actions-user"]',
+	'.commit-author[href$="web-flow"]',
+	'.commit-author[href$="bors"]',
```
yakov116, line 14: This does not seem safe to me

source/features/dim-bots.tsx
fluteds: I'll change to use a map instead 👍

kidonng: You should keep them separated in lines for diff's sake

source/features/dim-bots.tsx
```diff
@@ -6,9 +6,8 @@ import features from '.';
 
 const commitSelector = [
```
cheap-glitch, line 7: The arrays should not be nested, this should be a flat array that is mapped _and then_ joined. (`prSelector` should probably be joined too for consistency.)

Also having a single line for each name would be much more readable.

cheap-glitch (changes requested):

source/features/dim-bots.tsx
yakov116: This was not fixed

source/features/dim-bots.tsx
cheap-glitch: Using `flatMap` on a flat array is pointless :wink: 
What I meant was to remove the outer array.

fluteds: My bad! I'll get the outer array sorted

fregante (approved):
Untested. Is this good to go?

yakov116:
I think after 'bors' is dropped

kidonng:
> I think after 'bors' is dropped

Can we match the specific bot?

cheap-glitch:
> > I think after 'bors' is dropped
> 
> Can we match the specific bot?

This is what should happen already with the inclusion of the slash at the start. It should still be tested on real URLs though.

fregante:
> I think after 'bors' is dropped

I addressed that already by matching `/bors`

cheap-glitch:
My bad, I messed up. Don't really know what happened here :thinking:

kidonng:
> My bad, I messed up. Don't really know what happened here 🤔

It's a "feature" of GitHub. If the PR head ref is identical to the target branch, GitHub would consider the PR meaningless and close it. I've seen it several times, the most recent one is https://github.com/refined-github/refined-github/pull/4770#issuecomment-916126549.

fregante:
@fluteds can you open a new PR with the content from this file?

https://github.com/refined-github/refined-github/blob/e892193da88f89e76ef9f162312f9fe192910bbf/source/features/dim-bots.tsx

So you can appear as a contributor

cheap-glitch:
> So you can appear as a contributor

No need, I'll add him as a co-contributor when merging the PR :smiley: (I could have done it in the first commit but forgot)

fregante:
I'm not sure if that triggers the "Contributor" label. If it does, sure.

kidonng:
> I'm not sure if that triggers the "Contributor" label on their comments. If it does, sure. 

I can confirm what really counts is the commit, not merged PR. I have no merged PR but one commit in https://github.com/kovidgoyal/kitty yet I got the contributor label.

fregante:
Screen Shot

🥳 Thanks for your contribution @fluteds!

fregante:
@kidonng unfortunately that didn't add `fluted` here. I made the change manually though

Screen Shot 6

cheap-glitch:
> unfortunately that didn't add fluted here

Yeah since it's a squash merge, I think it only works if the resulting commit itself is co-authored. Adding the correct footer in the "description" textarea should do the trick.

PR #5141, opened 2021-12-02
Remove issue "May be fixed by" in `clean-conversation-headers`

kidonng:
This info is available in the sidebar:

截屏2021-12-02 21 56 53

## Test URLs

https://togithub.com/refined-github/refined-github/issues/4952

## Screenshot
Before:
image

After:

image

### DIFF
--- a/source/features/clean-conversation-headers.tsx
+++ b/source/features/clean-conversation-headers.tsx
@@ -24,6 +24,9 @@ async function initIssue(): Promise<void> {
 
 	// Removes: octocat opened this issue on 1 Jan [·] 1 comments
 	byline.firstChild!.textContent = byline.firstChild!.textContent!.replace('·', '');
+
+	// Removes: octocat opened this issue on 1 Jan · 1 comments [· May be fixed by #2]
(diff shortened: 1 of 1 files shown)


fregante:
I'm not sure I like deprioritize this piece of information, which is pretty important when looking at an issue. The header is *always* visible

kidonng:
I have conflicted thoughts about this too. I forgot to mention another thing that made me do this PR: the PR header doesn't have this "May be fixed by" link, only "Linked issues".

I would like to align issues with PR, since the latter cannot be changed I just removed the former from issues. How about we add something like "May fix #number" to PR header instead?

cheap-glitch:
> How about we add something like "May fix #number" to PR header instead?

Feels redundant to me, there's a good chance it's mentioned in the first post with a `Fixes #1234`, and it's in the sidebar too.

fregante:
> How about we add something like "May fix #number" to PR header instead?

We had a feature like this before GH added "may be fixed by". I'm -0 on this now though

fregante:
I'd rather not

PR #5143, opened 2021-12-02
URL `include` cleanup

kidonng:
Extracting stuff from #5113

Note to self: this PR mainly does two things:

- Check `isRepoSearch` and friends since [they are excluded from `isRepo`](https://github.com/fregante/github-url-detection/blob/34456a32fe230a1c71012e4e8c574ce16aff2d61/index.ts#L320-L325)
- https://github.com/fregante/github-url-detection/pull/105

## Test URLs

- `linkify-code`
	- https://github.com/search?q=github.com&type=code
	- https://github.com/refined-github/refined-github/search?q=github.com&
	- https://github.com/refined-github/refined-github/search?q=github.com&type=code
- `show-user-top-repositories`: https://gist.github.com/kidonng


source/features/linkify-code.tsx
```diff
@@ -35,6 +35,8 @@ function init(): void {
 void features.add(import.meta.url, {
 	include: [
 		pageDetect.hasCode,
+		() => pageDetect.isGlobalSearchResults() && new URLSearchParams(location.search).get('type') === 'code',
+		() => pageDetect.isRepoSearch() && (new URLSearchParams(location.search).get('type') === 'code' || !new URLSearchParams(location.search).get('type')),
```
cheap-glitch, line 39: Maybe have a line before the init with
```ts
const searchType = new URLSearchParams(location.search).get('type');
```
and then use it in the includes.

source/features/linkify-code.tsx
```diff
@@ -32,9 +32,13 @@ function init(): void {
 	});
 }
 
+const searchType = (): string | null => new URLSearchParams(location.search).get('type');
+
 void features.add(import.meta.url, {
 	include: [
 		pageDetect.hasCode,
+		() => pageDetect.isGlobalSearchResults() && searchType() === 'code',
+		() => pageDetect.isRepoSearch() && (searchType() ?? 'code') === 'code',
```
kidonng, line 41: Note to self: this should be backported into `github-url-detection`'s `hasCode`

source/features/show-user-top-repositories.tsx
```diff
@@ -23,5 +23,8 @@ void features.add(import.meta.url, {
 	include: [
 		pageDetect.isUserProfileMainTab,
 	],
+	exclude: [
+		pageDetect.isGist,
+	],
```
fregante, line 28: It's ok here, but it feels like we should treat `isGist` similarly to what we do with is404: _If the includes/asLongAs have isGist, then enable the feature there, if not don't._ There might only be a handful of situation where features can still apply to isGist, like hasCode, but maybe it's better to edit those instead, e.g.

```js
include: [
	hasCode,
	isGist,
]

kidonng:
Closing until I got the time to finish it.

fregante:
Closing as there's no usable code here (as mentioned in the review).  Avoid keeping PRs open for 2 months 😃

kidonng:
> Closing as there's no usable code here (as mentioned in the review).

It would be better as an issue instead, I just prefer actual code (and hate writing similar stuff twice). Also the feedback would be more related (like in this PR. Thanks!)

> Avoid keeping PRs open for 2 months 😃

Sorry but that's probably just I being myself. [I don't have an issue with that though.](https://github.com//refined-github/refined-github/pull/5111). Does it really matter if external changes aren't that much?

PR #5150, opened 2021-12-04
Fix `reactions-avatars` on releases and review comments

### LINKED ISSUE #5190: reaction avatars not working on review comment
### Please ensure:

### Include in this issue:

- [x] Screenshots/video/gif demonstrating the bug, if it’s visual
- [ ] Console errors, if any

### Description

There are no avatars for the reactions on review comments

Bildschirmfoto 2021-12-11 um 10 12 32

### Example URL

https://github.com/parcel-bundler/parcel/pull/7418#pullrequestreview-828068246

### Browser(s) used

Chrome 96.0.4664.55

cheap-glitch:
Fixes #5190 

## Test URLs

 * [Release list](https://github.com/refined-github/refined-github/releases)
 * [Single release](https://github.com/refined-github/refined-github/releases/tag/21.11.23)
 * [Issue with reactions](https://togithub.com/refined-github/refined-github/issues/2673)

## Screenshot

**Before**

screenshot-1638622978

**After**

screenshot-1638622876

### DIFF
--- a/source/features/reactions-avatars.tsx
+++ b/source/features/reactions-avatars.tsx
@@ -21,8 +21,7 @@ interface Participant {
 
 function getParticipants(button: HTMLButtonElement): Participant[] {
 	// Reaction buttons on releases and review comments have the list of people in their `title` attribute instead of `aria-label` #5150
-	const buttonTitle = button.getAttribute('title')!;
-	const users = (buttonTitle.length > 0 ? buttonTitle : button.getAttribute('aria-label')!)
+	const users = (button.getAttribute('title')! || button.getAttribute('aria-label')!)
 		.replace(/ reacted with.*/, '')
 		.replace(/,? and /, ', ')
 		.replace(/, \d+ more/, '')
--- a/source/features/reactions-avatars.tsx
+++ b/source/features/reactions-avatars.tsx
@@ -21,7 +21,7 @@ interface Participant {
 
 function getParticipants(button: HTMLButtonElement): Participant[] {
 	const currentUser = getUsername();
-	const users = button.getAttribute('aria-label')!
+	const users = (pageDetect.isReleasesOrTags() ? button.getAttribute('title') : button.getAttribute('aria-label'))!
 		.replace(/ reacted with.*/, '')
 		.replace(/,? and /, ', ')
 		.replace(/, \d+ more/, '')
@@ -56,8 +56,8 @@ async function showAvatarsOn(commentReactions: Element): Promise<void> {
 	const avatarLimit = arbitraryAvatarLimit - (commentReactions.children.length * approximateHeaderLength);
 
 	const participantByReaction = select
-		.all(':scope > button[aria-label$="emoji"]', commentReactions)
-		.map(button => getParticipants(button));
+		.all(':scope > button:is([aria-label$="emoji"], [title$="emoji"])', commentReactions) // `aria-label` is for PR/issue comments, `title` for releases
+		.map(button => getParticipants(button as HTMLButtonElement));
 	const flatParticipants = flatZip(participantByReaction, avatarLimit);
 
 	for (const {button, username, imageUrl} of flatParticipants) {
--- a/source/features/reactions-avatars.tsx
+++ b/source/features/reactions-avatars.tsx
@@ -56,7 +56,7 @@ async function showAvatarsOn(commentReactions: Element): Promise<void> {
 	const avatarLimit = arbitraryAvatarLimit - (commentReactions.children.length * approximateHeaderLength);
 
 	const participantByReaction = select
-		.all(':scope > button:is([aria-label$="emoji"], [title$="emoji"])', commentReactions) // `aria-label` is for PR/issue comments, `title` for releases
+		.all(':scope > button.social-reaction-summary-item', commentReactions) // `aria-label` is for PR/issue comments, `title` for releases
 		.map(button => getParticipants(button as HTMLButtonElement));
 	const flatParticipants = flatZip(participantByReaction, avatarLimit);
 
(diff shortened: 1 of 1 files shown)


source/features/reactions-avatars.tsx
```diff
@@ -56,8 +56,8 @@ async function showAvatarsOn(commentReactions: Element): Promise<void> {
 	const avatarLimit = arbitraryAvatarLimit - (commentReactions.children.length * approximateHeaderLength);
 
 	const participantByReaction = select
-		.all(':scope > button[aria-label$="emoji"]', commentReactions)
-		.map(button => getParticipants(button));
+		.all(':scope > button:is([aria-label$="emoji"], [title$="emoji"])', commentReactions) // `aria-label` is for PR/issue comments, `title` for releases
+		.map(button => getParticipants(button as HTMLButtonElement));
```
cheap-glitch, line 60: Alternatively the selector could simply be
```css
button.social-reaction-summary-item
```
It's valid on PR/issues as well and doesn't necessitate the type coercion (which I guess is due to a bug/limitation of `typed-query-selector`?)

source/features/reactions-avatars.tsx
```diff
@@ -21,7 +21,7 @@ interface Participant {
 
 function getParticipants(button: HTMLButtonElement): Participant[] {
 	const currentUser = getUsername();
-	const users = button.getAttribute('aria-label')!
+	const users = (pageDetect.isReleasesOrTags() ? button.getAttribute('title') : button.getAttribute('aria-label'))!
```
cheap-glitch, line 24: The list of people who reacted is now in the `title` attribute (which makes sense 'cause you can hover on the button to see it).
…But only for releases pages, apparently.

cheap-glitch (commented):
The fact that the reaction buttons are stuck together like that is NOT due to RGH btw.

source/features/reactions-avatars.tsx
kidonng: Does GHE has reactions? Maybe it would be better to `button.getAttribute('title') ?? button.getAttribute('aria-label')`

cheap-glitch: > Maybe it would be better to `button.getAttribute('title') ?? button.getAttribute('aria-label')`

Can't, because the buttons on PR/issue comments also have a title attribute (but it's empty). And the buttons on releases also have an `aria-label` — but it doesn't contain the right data for the feature.

kidonng: Aren't the code correct in those conditions?

- PR/issue comments: `title` is empty so it uses `aria-label` (correct)
- Releases: `title` is preffered

cheap-glitch: > `title` is empty so it uses `aria-label`

No, `??` (or its compiled equivalent) triggers only if the left value is `null` or `undefined`, it won't coerce an empty string into a falsy value like `||` would. We could use that last one, but I think the tertiary is clearer about why having both attributes is needed.

kidonng: Right, I missed the difference between `||` and `??` (which I don't really use outside of RGH due to the fact it's still a shiny new thing) 😅

cheap-glitch: > Also let's wait and see if GitHub changes this somewhere else. It's also strange they only do this on one page.

Seems like you were right, it changed for review comments too! But not regular comments… :grimacing: 

> which I don't really use outside of RGH due to the fact it's still a shiny new thing

Not so much, it's supported by all major browsers now, and by Node since v14.

fregante, line 24: Saves one character 😃 
```suggestion
	const users = pageDetect.isReleasesOrTags() ? button.getAttribute('title')! : button.getAttribute('aria-label')!
```

source/features/reactions-avatars.tsx
fregante: Let's do this if it works

```suggestion
		.all(':scope > button.social-reaction-summary-item', commentReactions) // `aria-label` is for PR/issue comments, `title` for releases
		.map(button => getParticipants(button as HTMLButtonElement));
```

cheap-glitch:
Marking as draft, because this needs to be fixed for reactions on review comments too (but not for regular issue/PR comments ¯\\_(ツ)_/¯)

fregante:
We can merge this if it fixes releases, so we can release a new version tomorrow

source/features/reactions-avatars.tsx
cheap-glitch: I just hope this doesn't break things on GHE

source/features/reactions-avatars.tsx
```diff
@@ -20,13 +20,15 @@ interface Participant {
 }
 
 function getParticipants(button: HTMLButtonElement): Participant[] {
-	const currentUser = getUsername();
-	const users = button.getAttribute('aria-label')!
+	// Reaction buttons on releases and review comments have the list of people in their `title` attribute instead of `aria-label` #5150
+	const buttonTitle = button.getAttribute('title')!;
+	const users = (buttonTitle.length > 0 ? buttonTitle : button.getAttribute('aria-label')!)
```
fregante, line 25: If it passes lint, this is equivalent
```suggestion
	const users = (buttonTitle || button.getAttribute('aria-label')!)
```

fregante (approved):
🛳✅🛳✅🛳✅🛳✅🛳
✅🛳✅🛳✅🛳✅🛳✅
🛳✅🛳✅🛳✅🛳✅🛳
✅🛳✅🛳✅🛳✅🛳✅
🛳✅🛳✅🛳✅🛳✅🛳
✅🛳✅🛳✅🛳✅🛳✅
🛳✅🛳✅🛳✅🛳✅🛳
✅🛳✅🛳✅🛳✅🛳✅
🛳✅🛳✅🛳✅🛳✅🛳
✅🛳✅🛳✅🛳✅🛳✅

kidonng:
> The fact that the reaction buttons are stuck together like that is NOT due to RGH btw.

I now see it, maybe it's worth fixing 🤔

PR #5154, opened 2021-12-05
Drop `trending-menu-item` feature

fregante:
- Related to https://github.com/refined-github/refined-github/pull/5153

Already discussed with Sindre, which was the proponent of this and `deprioritize-marketplace-link` if I remember correctly.

The core problem is that neither of these are important enough to warrant the effort/code and especially a keyboard shortcut.

For the record, the Trending page is a subpage of Explore. You could also create a… browser bookmark.


yakov116:
> Already discussed with Sindre, which was the proponent of this and [`deprioritize-marketplace-link`](https://github.com/refined-github/refined-github/blob/main/source/features/deprioritize-marketplace-link.tsx?rgh-link-date=2021-12-05T16%3A38%3A36Z) if I remember correctly.

Yup goes waaaay back https://github.com/refined-github/refined-github/issues/153

PR #5156, opened 2021-12-05
Use emoji in `unfinished-comments`

fregante:
I feel that the text isn't visible enough. The only concern is OSes where emojis aren't available. 🤷‍♂️ 😬 

## Test URLs

This PR, type a comment

## Before

Screen Shot 4

## After

Screen Shot 5

### DIFF
--- a/source/features/unfinished-comments.tsx
+++ b/source/features/unfinished-comments.tsx
@@ -29,7 +29,7 @@ function updateDocumentTitle(): void {
 
 	if (document.visibilityState === 'hidden' && hasDraftComments()) {
 		documentTitle = document.title;
-		document.title = '(Draft comment) ' + document.title;
+		document.title = '✏️ (unsent) ' + document.title;
 	} else if (documentTitle) {
 		document.title = documentTitle;
 		documentTitle = undefined;
--- a/source/features/unfinished-comments.tsx
+++ b/source/features/unfinished-comments.tsx
@@ -29,7 +29,7 @@ function updateDocumentTitle(): void {
 
 	if (document.visibilityState === 'hidden' && hasDraftComments()) {
 		documentTitle = document.title;
-		document.title = '✏️ (unsent) ' + document.title;
+		document.title = '✏️ (draft) ' + document.title;
 	} else if (documentTitle) {
 		document.title = documentTitle;
 		documentTitle = undefined;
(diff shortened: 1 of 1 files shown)


source/features/unfinished-comments.tsx
```diff
@@ -29,7 +29,7 @@ function updateDocumentTitle(): void {
 
 	if (document.visibilityState === 'hidden' && hasDraftComments()) {
 		documentTitle = document.title;
-		document.title = '(Draft comment) ' + document.title;
+		document.title = '✏️ (unsent) ' + document.title;
```
kidonng, line 32:
```suggestion
		document.title = '(✏️ Draft) ' + document.title;
```

kidonng:
> The only concern is OSes where emojis aren't available. 🤷‍♂️ 😬 

Those OSes are probably not able to run RGH anyway: those https://www.unicompat.com/270F-FE0F

source/features/unfinished-comments.tsx
fregante: it's not technically a draft, it's just not yet sent

fregante:
> Those OSes are probably not able to run RGH anyway: [unicompat.com/270F-FE0F](https://www.unicompat.com/270F-FE0F)

That site lacks Linux 🥲

kidonng:
> > Those OSes are probably not able to run RGH anyway: [unicompat.com/270F-FE0F](https://www.unicompat.com/270F-FE0F)
> 
> That site lacks Linux 🥲

https://twitter.com/PixelAmbacht/status/1346751946007392257

(Actually it's answering another question but also applies here.)

source/features/unfinished-comments.tsx
kidonng: > it's not technically a draft as intended in emails, it's just not yet sent


FYI some messager like Telegram also phrase it as "Draft". I would say "unsent" is pretty obscure here.

IMG_20211207_110134

fregante:
```suggestion
		document.title = '✏️ (draft) ' + document.title;
```

source/features/unfinished-comments.tsx
```diff
@@ -29,7 +29,7 @@ function updateDocumentTitle(): void {
 
 	if (document.visibilityState === 'hidden' && hasDraftComments()) {
 		documentTitle = document.title;
-		document.title = '(Draft comment) ' + document.title;
+		document.title = '✏️ (draft) ' + document.title;
```
fregante, line 32: Maybe even just

```suggestion
		document.title = '✏️ Draft - ' + document.title;
```

I'd avoid `Draft:` because what follows isn't the actual content

PR #5166, opened 2021-12-10
Fix option saving in Safari

fregante:
- Fixes https://github.com/refined-github/refined-github/issues/4823

Every time I want to change the options in Safari, I have to manually clear the cache due to this issue. So why not do it automatically?


kidonng:
This is not right, every Chromium's UserAgent contains `Safari/537.36`.

fregante:
I had even run that line in Chrome’s dev tools but forgot to check the result 😅

PR #5175, opened 2021-12-10
Meta: Automatically label commits in releases

fregante:
Untested 

Docs: https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes#configuration-options

Mentioned in

- https://github.com/refined-github/refined-github/issues/4008#issuecomment-942723376


darkred:
Hi @fregante. I think that it would more preferable to reorder the items as follows:
- New 
- Changes
- Bugfixes

I believe 'New' entries deserve to be listed first, followed by 'Changes' and 'Bugfixes' to be in the end.
If you agree, I'd like to make a PR.

PR #5177, opened 2021-12-10
Meta: Fix option saving only in Safari

kidonng:
Reverts refined-github/refined-github#5166

Perhaps it should be "if UserAgent includes `Safari` but not `Chrome` then it's Safari".


fregante:
Well don't revert it, fix it if anything

PR #1392, opened 2018-06-27
Show stargazers you know

Hermanya:
image
This is what it normally looks like.

---

image
This is what it looks like on hover.

---

When there is no people to show, nothing is shown.

PR #5187, opened 2021-12-10
Meta: remove obsolete related projects

kidonng:
- https://github.com/DrewML/OctoEdit: has not seen activity since May 2016
- https://github.com/sindresorhus/refined-twitter: archived

To be honest https://github.com/Hermanya/friendly-github should be dropped as well since a. it is (mostly) a unmaintained fork b. its corresponding browser extensions has not been updated since Sep 2018. As much as I appreciate the author's contributions like #1392, it is mostly a malfunctioning extension we probably don't want to recommend anymore.


fregante:
> has not been updated since Sep 2018

Yeah that's a shame. If it's updated again we can restore it. It's not really a fork but it uses our setup to offer additional features, which is welcome.

fregante:
Thanks for the cleanup!

PR #5194, opened 2021-12-11
Enable `sort-conversations-by-update-time` on label links

### LINKED ISSUE #5176: `sort-conversations-by-update-time` not being applied to label links in issues/PRs
### Please ensure:

### Include in this issue:

- [ ] Console errors, if any

### Description

The "bug" link in the sidebar of this issue points to the unsorted `/labels/` link.

#### Current

https://github.com/refined-github/refined-github/labels/bug

#### Expected

https://github.com/refined-github/refined-github/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3Abug

### Example URL

This issue

### Browser(s) used

Safari

cheap-glitch:
Fix #5176 

## Test URLs

 * This pages's sidebar
 * [An issue's sidebar](https://togithub.com/refined-github/refined-github/issues/5176)
 * [The repo's labels list](https://github.com/refined-github/refined-github/labels)
 * [The global conversation list](https://github.com/issues?q=is%3Aissue+is%3Aopen+author%3A%40me+archived%3Afalse+sort%3Aupdated-desc)

## Screenshot

n/a


cheap-glitch:
My bad, opened this too soon. I didn't realize label links are redirected, and thus their URL must be completely replaced if we want to change the sorting order.

fregante:
This should probably be done in GitHubURL. I don't remember. We have code somewhere that assumes default queries based on their URL. For example "/pulls" implies "is:pull is:open". Same goes for label URLs, except that "implying" this also changes the pathname. It will be difficult to fit it in today's release.

source/github-helpers/search-query.ts
```diff
@@ -66,6 +66,11 @@ export default class SearchQuery {
 			return '';
 		}
 
+		// Parse label links #5176
+		if (this.link.pathname.includes('/labels/')) {
```
cheap-glitch, line 70: Worth using `pageDetect` here?

cheap-glitch (commented):
Should I add a test case in `test/search-query.ts` for this?

source/github-helpers/search-query.ts
yakov116: I think we have one, I remember having some other issue with that page. (Unless that was maybe to exclude this page?) Dont Remember.

fregante: Definitely. This line matches /fregante/label

source/github-helpers/search-query.ts
```diff
@@ -66,6 +66,11 @@ export default class SearchQuery {
...
+			return `is:open label:"${decodeURIComponent(/\/labels\/(.+)$/.exec(this.link.pathname)![1])}"`;
```
fregante, line 71: There's a lot going on here, let's create at least one variable. Also you can use [^/]+ instead of .+, unless slashes are not escaped by GitHub.

source/github-helpers/search-query.ts
```diff
@@ -95,6 +100,10 @@ export default class SearchQuery {
 		const parts = splitQueryString(query);
 		const cleaned = cleanQueryParts(parts);
 		this.searchParams.set('q', cleaned.join(' '));
+		if (this.link?.pathname.includes('/labels/')) {
+			// Avoid a redirection to the conversation list that would erase the search query #5176
+			this.link.pathname = this.link.pathname.replace(/\/labels\/.+$/, '/issues');
```
fregante, line 105: SearchQuery is a mess, sorry about that. We try to parse, clean and update elements on creation. That’s a lot of side effects for a class. 

For this PR this line can be merged, but know that just calling this class on a /label/ link will automatically drop this /label/ before even attempting any changes. 

The idea was to make it easy to update existing links, but perhaps we should avoid changing the original link until .commit() or .updateOriginal() is called. Alternative suggestions welcome.

fregante (changes requested):

source/github-helpers/search-query.ts
cheap-glitch: There's nothing in `pageDetect` to target a label link, so for now I just extracted the regex and used it everywhere.

fregante: > There's nothing in pageDetect to target a label link

And that's fine 👍 Thanks for checking

source/github-helpers/search-query.ts
```diff
@@ -24,6 +24,8 @@ function cleanQueryParts(parts: string[]): string[] {
 	return deduplicateKeywords(parts, 'is:issue', 'is:pr');
 }
 
+const labelLinkRegex = /\/labels\/([^/]+)$/;
```
fregante, line 27: Still matches /fregante/labels/issues. It should also include the the beginning, like `^[/][^/]+[/][^/]+[/]labels[/][^/]+[/]?$`

fregante (approved):

PR #5197, opened 2021-12-13
Restore `hide-watch-and-fork-count`

fregante:
- Fixes https://github.com/refined-github/refined-github/issues/5161

I had hotfixed this in https://github.com/refined-github/refined-github/commit/88c3effcdf64cea8bf5dbdbf3084dce75af9cfb5

The old logic no longer applies, the links are now accessible in the sidebar: https://github.com/refined-github/refined-github/issues/5179

I'd argue we don't even need to keep this disableable since it's so simple and there's no loss of information (it appears in the sidebar), but maybe that's harder to sell since it's been possible to disable it already. Thoughts? Should we make it a CSS-only feature?

## Test URLs

- This page
- https://github.com/refined-github/shorten-repo-url

## Screenshot

Screen Shot

### DIFF
--- a/source/features/hide-watch-and-fork-count.css
+++ b/source/features/hide-watch-and-fork-count.css
@@ -1,25 +1,6 @@
-/* The Watchers count visible (and clickable) when the dropdown is open */
-/* This isn't possible for the Forks count because of an incompatible DOM structure and #2368 */
-.rgh-hide-watch-and-fork-count details:not([open]) .btn[data-ga-click^='Repository, click Watch'] {
-	border-radius: 6px !important; /* Matches GitHub's default */
-}
-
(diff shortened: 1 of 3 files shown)


cheap-glitch:
> Should we make it a CSS-only feature?

Let's, and if people complain we can just restore the `.tsx` file.

kidonng:
> Thoughts? Should we make it a CSS-only feature?

👍 Can't agree more

fregante:
If anyone wants to disable this feature, here's the CSS to paste in the "Custom CSS" box:

```css
.pagehead-actions :is(#repo-network-counter, #repo-notifications-counter) {
	display: inline-block;
}
```

PR #5198, opened 2021-12-13
Delay `dim-bots` after hover

fregante:
- Closes https://github.com/refined-github/refined-github/issues/5158

## Test URLs

https://github.com/pixiebrix/pixiebrix-extension/pulls?q=is%3Apr+sort%3Aupdated-desc+is%3Aclosed+bump+

## PR list

https://user-images.githubusercontent.com/1402241/145775347-7b773183-6efe-475b-9b83-b2753ed221fa.mov

## Commit list

https://user-images.githubusercontent.com/1402241/145978531-ae3d3b9f-8825-45f1-81ff-381aac8c1b70.mov

### DIFF
--- a/source/features/dim-bots.css
+++ b/source/features/dim-bots.css
@@ -3,6 +3,14 @@ READ BEFORE EDITING. Some selectors match the same items ("focused" selector vs 
 Commits that have a long commit message will not be dimmed after the user uncollapses the `<details>` element.
 */
 
+:root {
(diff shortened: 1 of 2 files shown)


fregante:
I looked into dimming the text lightly after the hover, but that means adding a whole other block to add this intermediary state and this file already has a warning in the header, so I'd rather not make it any more complex to edit.

PR #5203, opened 2021-12-15
Meta: Allow setting local dev team

### LINKED ISSUE #5202: Safari test should work with signing certificate
Previously, there was no signing certificate set in the Xcode project, but it's annoying for me to change it to the correct certificate each time. It's better to force `xcodebuild` command to run without a certificate instead.

https://github.com/refined-github/refined-github/runs/4524894735?check_suite_focus=true

https://github.com/refined-github/refined-github/blob/a79f49bdb7e571bd8402f3cddba8154220bd0e0c/package.json#L13

---

I have disabled the test until this is resolved: https://github.com/refined-github/refined-github/commit/a79f49bdb7e571bd8402f3cddba8154220bd0e0c#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R79

fregante:
- Fixes #5202 
- Partially undoes https://github.com/refined-github/refined-github/commit/a79f49bdb7e571bd8402f3cddba8154220bd0e0c

You should now be able to create the file `safari/LocalOverrides.xcconfig` with the content:

```xcconfig
CODE_SIGN_IDENTITY = "Apple Development";
DEVELOPMENT_TEAM = YG56YK5RN5
```

Untested except on CI, does it work for you?

### DIFF
--- a/safari/Config.xcconfig
+++ b/safari/Config.xcconfig
@@ -1,2 +1,4 @@
 MARKETING_VERSION = 21.12.12
 CURRENT_PROJECT_VERSION = 16
+
+#include? "LocalOverrides.xcconfig"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -76,4 +76,4 @@ jobs:
       - uses: maxim-lobanov/setup-xcode@v1
         with:
           xcode-version: latest
-      # - run: npm run pack:safari
+      - run: npm run pack:safari
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@ test/web-ext-profile
 !test/web-ext-profile/.gitkeep
 .cache
 .parcel-cache
+
+LocalOverrides.xcconfig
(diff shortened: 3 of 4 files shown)


safari/Config.xcconfig
```diff
@@ -1,2 +1,4 @@
 MARKETING_VERSION = 21.12.12
 CURRENT_PROJECT_VERSION = 16
```
fregante, line 2: I wonder if we should keep these in the repo too.

- `MARKETING_VERSION` can be set by a custom `version` script inside an additional  `GeneratedOverrides.xcconfig` file
- `CURRENT_PROJECT_VERSION` I think Xcode takes care of this? I think I've seen it automatically bump this number before

sindresorhus:
> MARKETING_VERSION can be set by a custom version script inside an additional GeneratedOverrides.xcconfig file

Yes, it would be nice to not have to duplicate this.

> CURRENT_PROJECT_VERSION I think Xcode takes care of this? I think I've seen it automatically bump this number before

Not as far as I know. Xcode has a CLI tool to bump it.

We could maybe set the build version based on commit count or tag count (+ padding to be higher than existing number).

https://gist.github.com/dimitribouniol/24f0406e0878e5880f3a64dea3c528b6

sindresorhus:
This works, but not quotes on the code sign identity:

```
CODE_SIGN_IDENTITY = Apple Development
DEVELOPMENT_TEAM = YG56YK5RN5
```

PR #5205, opened 2021-12-15
Fix `reactions-avatars` size

fregante:
- Fixes https://github.com/refined-github/refined-github/issues/5204

This is happening because they added a CSS rule that targets `span`s in that button, which we're using.

## Test URLs

https://togithub.com/xojs/xo/issues/261

## Screenshot

Screen Shot


fregante:
> This is happening because they added a CSS rule that targets `span`s in that button, which we're using.

Screen Shot 5

fregante:
I love CSS hotfixes, nobody probably will even see this issue

PR #5210, opened 2021-12-15
Fix `conversation-activity-filter` hiding unresolved comments

### LINKED ISSUE #5000: `conversation-activity-filter` also hides unresolved review comments
### Description

`conversation-activity-filter` seems to be intended to reduce clutter when looking at a pull request and comments. I haven't found a good combination of settings where I see unresolved comments still though. I thought this was originally handled by https://github.com/refined-github/refined-github/issues/3966 

However, this hides all comments that are pulled from in code files. 

Version Used: `21.10.25`

### Screenshot

Screenshot of "hide events and collapsed comments", where comments from files are hidden. 

image

### Example URL

https://github.com/dotnet/roslyn/pull/56957

cheap-glitch:
Fixes #5000 

## Test URLs

https://togithub.com/refined-github/sandbox/pull/10

## Screenshot

https://user-images.githubusercontent.com/46634000/146204750-afac9ede-956c-438f-b4db-fbb2cba43944.mp4

### DIFF
--- a/source/features/conversation-activity-filter.tsx
+++ b/source/features/conversation-activity-filter.tsx
@@ -41,7 +41,7 @@ function processSimpleComment(item: HTMLElement): void {
 	}
 
 	// Hide comments marked as resolved/hidden
-	if (select.exists('.minimized-comment:not(.d-none) > details', item)) {
+	if (select.exists('.minimized-comment > details', item)) {
 		item.classList.add(hiddenClassName);
 	}
 }
--- a/source/features/conversation-activity-filter.tsx
+++ b/source/features/conversation-activity-filter.tsx
@@ -30,8 +30,8 @@ function isWholeReviewEssentiallyResolved(review: HTMLElement): boolean {
 	}
 
 	// Don't combine the selectors or use early returns without understanding what a thread or thread comment is
-	const hasUnresolvedThread = select.exists('details.js-resolvable-timeline-thread-container', review);
-	const hasUnresolvedThreadComment = select.exists('.js-minimize-comment.d-none', review);
+	const hasUnresolvedThread = select.exists('.js-resolvable-timeline-thread-container[data-resolved="false"]', review);
+	const hasUnresolvedThreadComment = select.exists('.timeline-comment-group:not(.minimized-comment)', review);
 	return !hasUnresolvedThread || !hasUnresolvedThreadComment;
 }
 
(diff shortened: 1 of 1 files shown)


source/features/conversation-activity-filter.tsx
```diff
@@ -41,7 +41,7 @@ function processSimpleComment(item: HTMLElement): void {
 	}
 
 	// Hide comments marked as resolved/hidden
-	if (select.exists('.minimized-comment:not(.d-none) > details', item)) {
+	if (select.exists('.minimized-comment > details', item)) {
```
cheap-glitch, line 44: Also fixes the feature for regular comments that have been minimized.

source/features/conversation-activity-filter.tsx
```diff
@@ -51,8 +51,7 @@ function processReview(review: HTMLElement): void {
 		return;
 	}
 
-	const shouldHideWholeReview = isWholeReviewEssentiallyResolved(review);
-	if (shouldHideWholeReview) {
+	if (isWholeReviewEssentiallyResolved(review)) {
 		review.classList.add(hiddenClassName);
 		return;
 	}
```
cheap-glitch, line 57: Additionally, [hiding resolved threads](https://github.com/refined-github/refined-github/pull/5210/files#diff-2db094a8bff1492679cf3b912175ece902270acd21c77d1cfdd991328eb25db7R59-R61) :point_down: could be done via CSS only.

cheap-glitch (commented):
Should we hide the minimized review comments in review threads too? I'm talking about those:

image


It could be done via pure CSS.

fregante:
Refer to the original discussion for context. Maybe it was attempted

source/features/conversation-activity-filter.tsx
fregante: I think the class excludes non-minimized comments. 

In short, GitHub (used to?) has 2 "comment" elements for each comment. One is the regular version and one is the minimized version: They use that class to hide one of them, so effectively we use it to find the comments that are actually minimized (as opposed to DOM elements that aren't shown)

fregante:
Titles should try to be a bit more descriptive. I'm not sure if it's correct. Often just `Fix: <issue title>` is enough if it includes the feature name.

source/features/conversation-activity-filter.tsx
cheap-glitch: Apparently that's no longer the case, because only comments that are actually minimized will have that class now. In fact the `> details` part is not even really needed anymore.

cheap-glitch:
> Should we hide the minimized review comments in review threads too? I'm talking about those:
> 
> It could be done via pure CSS.

I'll try this in another PR since that's technically an enhancement.

PR #5212, opened 2021-12-15
Meta: Clarify feature names

### LINKED ISSUE #4992: Make features with similar names distinctive
### Description

We have two features with similar names: `conversation-filters` and `conversation-activity-filter`. It's not clear from the names they do very different things, and I think it's best to distinguish them via a different name.

I suppose rename `conversation-filters` to `more-conversation-filters`, since that's what the feature does actually. It's not adding a whole feature of conversation filters.

### Screenshot

_No response_

### Example URL

https://github.com/refined-github/refined-github/issues

cheap-glitch:
Resolves #4992 

* `conversation-filters` → `more-conversation-filters`
* `quick-pr-diff-options` → `quick-diff-options`
* `quick-review-buttons` → `review-type-buttons`


fregante:
> * `conversation-filters` → `more-conversation-filters`

👍 

> * `quick-pr-diff-options` → `quick-diff-options`

`one-click-diff-options`?



> * [`quick-review-buttons`](https://github.com/refined-github/refined-github/blob/main/source/features/quick-review-buttons.tsx?rgh-link-date=2021-12-15T19%3A38%3A12Z) → `review-type-buttons`

`one-click-review-submission`?

cheap-glitch:
Actually, most `quick-*` features could be renamed `one-click-*`.

fregante:
> Actually, most `quick-*` features could be renamed `one-click-*`.

Let's keep this in mind for the future. I'm still wary of renaming features because it's harder to find them in discussions etc.

kidonng:
We don't need to go for `one-click-`. Some people don't/can't use mouse. The current way is more inclusive.

PR #5213, opened 2021-12-15
Meta: Replace `deinit` option with `init` return value

FloEdelmann:
Implements the second checkbox from https://github.com/refined-github/refined-github/issues/4008#issuecomment-812885528.

## Test URLs

* `batch-mark-files-as-viewed`: this PR's files
* `copy-on-y`: e.g. https://github.com/refined-github/refined-github/blob/081adcdabbd22d5d59721a1aceffed81fb322c5f/source/features/index.tsx
* `highlight-deleted-and-added-files-in-diffs`: https://github.com/sindresorhus/refined-github/pull/3062/files
* `quick-review-comment-deletion`: this PR's discussion thread or this PR's files
* `releases-tab`: basically all repo pages
* `select-notifications`: https://github.com/notifications
* `sticky-sidebar`: this PR or this repo's homepage
* `wait-for-build`: this PR

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -17,11 +17,11 @@ Suggestions and pull requests are highly encouraged! Have a look at the [open is
 
 The simplest usage of `feature.add` is the following. This will be run instantly on all page-loads:
 
-```js
+```ts
 import * as pageDetect from 'github-url-detection';
 import features from '.';
 
-function init() {
+function init(): void {
 	console.log('✨');
 }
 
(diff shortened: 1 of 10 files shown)


fregante (approved):
Thanks Flo! LGTM and I want to see this merged as soon as possible. 

There are a number of improvements that can come *after merge*, this PR only introduces the pattern to preexisting code. For example we can:

- return *more* Deinit functions, like the result of every single delegate() call in every feature
- Make that VoidFunction a more specific value like:
   `type Deinit = VoidFunction | {abort: VoidFunction} | AbortController, etc` and then set the return value to `Deinit | Deinit[]`
   this will let us return values easily without having to locate the abort function of each
- alternatively I think it’d be great to simply create a new abortSignal at every load and pass it to each function. This might be a little difficult to pass around at times, but it would be great to stop any async activity when unloading. For example addEventListener natively supports this signal so those features would not have to return anything.

PR #5223, opened 2021-12-17
Meta: Automatically generate release again

fregante:
For https://github.com/refined-github/refined-github/issues/4008#issuecomment-942723376

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,6 +30,15 @@ jobs:
       - uses: fregante/daily-version-action@v1
         name: Create tag if necessary
         id: daily-version
+      - name: Create Release Notes
+        uses: actions/github-script@v4.0.2
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            await github.request(`POST /repos/${{ github.repository }}/releases`, {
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -30,6 +30,15 @@ jobs:
       - uses: fregante/daily-version-action@v1
         name: Create tag if necessary
         id: daily-version
+      - name: Create release
```
FloEdelmann, line 33:
```suggestion
      - if: steps.daily-version.outputs.created
        name: Create release
```

.github/workflows/release.yml
```diff
@@ -30,6 +30,15 @@ jobs:
...
+        uses: actions/github-script@v4.0.2
```
FloEdelmann, line 34: Let's use the newest version directly.

```suggestion
        uses: actions/github-script@v5.0.0
```

FloEdelmann (changes requested):

fregante:
This did not work: https://github.com/refined-github/refined-github/actions/runs/1607743677

FloEdelmann:
It's definitely the latter:

grafik

FloEdelmann:
Ahh, of course `steps.version.outputs.version` is empty, if the step is called `daily-version`...

PR #5224, opened 2021-12-17
Meta: Correctly display "token missing" errors in `bugs-tab`

fregante:
Fixes https://github.com/refined-github/refined-github/issues/5181

## Test URLs

https://github.com/refined-github/refined-github but without a token stored

## Screenshot

Not yet tested

### DIFF
--- a/source/features/bugs-tab.tsx
+++ b/source/features/bugs-tab.tsx
@@ -131,7 +131,7 @@ async function addBugsTab(): Promise<void | false> {
 		bugsCounter.title = bugCount > 999 ? String(bugCount) : '';
 	} catch (error: unknown) {
 		bugsCounter.remove();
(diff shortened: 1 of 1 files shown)


cheap-glitch (approved):
Tested on Firefox with a token made invalid.

This correctly suppresses the errors specific to `bugs-tab`, that said, I still get these error messages in the console:

```
Error: The token seems to be incorrect or expired. Update it in the options. refined-github.js:16335:5
```

So it seems like some `RefinedGitHubAPIError` are somehow slipping through?

fregante:
> So it seems like some `RefinedGitHubAPIError` are somehow slipping through?

Incorrect/expired tokens are not handled the same way. We only catch missing token errors in this case because the call isn't made at all so we detect it synchronously and log the message before the feature even runs.

PR #5231, opened 2021-12-20
Fix: The welcome page isn't opened on Safari

### LINKED ISSUE #5228: The welcome page isn't opened on Safari
### Please ensure:

### Include in this issue:

- [ ] Console errors, if any

### Description

We noticed this at the beginning,  but we never investigated. Now I see this error in the background page when toggling the extension:

```
TypeError: undefined is not an object (evaluating 'browser.management.getSelf')
```

The API isn't supported:

https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/management/getSelf

I think we can change this line with a 0.0.0 version check that we're using elsewhere. We can move that function in the helpers.

### Example URL

background page

### Browser(s) used

Safari 15.1

cheap-glitch:
Fixes #5228

### DIFF
--- a/source/background.ts
+++ b/source/background.ts
@@ -4,6 +4,7 @@ import addDomainPermissionToggle from 'webext-domain-permission-toggle';
 
 import './options-storage';
 import {getRghIssueUrl} from './helpers/rgh-issue-link';
+import isDevelopmentVersion from './helpers/is-development-version';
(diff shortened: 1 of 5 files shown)


fregante (approved):
Fun-tastic

PR #5238, opened 2021-12-22
Remove `quick-review` additional listener when navigating

### LINKED ISSUE #5171: `addSidebarReviewButton` TypeError: Cannot read properties of undefined (reading 'append')
### Description

1. Visit https://github.com/refined-github/refined-github/pull/5150#ref-issue-1076541009
2. Click the link to issue `#5169`
3. See error in console

### Browser(s) used

Chrome

cheap-glitch:
Fixes #5171 

## Test URLs

Go to https://togithub.com/refined-github/sandbox/pull/7 and click the linked issue in the sidebar

## Screenshot

n/a


source/features/index.tsx
```diff
@@ -155,6 +155,9 @@ const setupPageLoad = async (id: FeatureID, config: InternalRunConfig): Promise<
 		return;
 	}
 
+	const controller = new AbortController();
+	document.addEventListener('pjax:start', () => { controller.abort(); }, {once: true});
```
cheap-glitch, line 159: For some reason passing `controller.abort` directly doesn't work ¯\\\_(ツ)\_/¯

source/helpers/on-element-removal.ts
```diff
@@ -1,14 +1,22 @@
 import mem from 'mem';
 
 const onElementRemoval = mem(
-	async (element: Element): Promise<void> => (
+	async (element: Element, {signal}: {signal?: AbortSignal} = {}): Promise<void> => (
```
cheap-glitch, line 4: The signal is passed in an "options object" to mimic `addEventListener`.

cheap-glitch (commented):
This is a first attempt at solving this, there might be a simpler solution :thinking:

source/helpers/on-element-removal.ts
fregante: I think we can avoid it in some places like this one since it's not a shared npm module.

```suggestion
	async (element: Element, signal?: AbortSignal): Promise<void> => (
```

source/helpers/on-replaced-element.ts
```diff
@@ -25,7 +25,11 @@ export default async function onReplacedElement(
 
 	while (trackedElement) {
 		// eslint-disable-next-line no-await-in-loop
-		await onElementRemoval(trackedElement);
+		await onElementRemoval(trackedElement, signal);
+		if (signal?.aborted) {
+			return;
+		}
```
cheap-glitch, line 31: This part feels really awkward, because what if the signal is aborted before the `while` loop? Or just before the callback is called?

But short of putting `if` blocks everywhere, I don't really know how to improve it :thinking:

fregante: I think it's ok, `onElementRemoval` is not a poller

fregante (approved):

kidonng:
Problem: page stuck when navigating from `isPRConversation` to other pages via PJAX.

fregante:
For the record: This addressed part of:

- https://github.com/refined-github/refined-github/issues/5222#issuecomment-996625242

PR #5239, opened 2021-12-22
Meta: Refactor `SearchQuery`

cheap-glitch:
Related to https://github.com/refined-github/refined-github/issues/5222#issuecomment-996884047

### DIFF
--- a/source/features/bugs-tab.tsx
+++ b/source/features/bugs-tab.tsx
@@ -71,7 +71,7 @@ async function getSearchQueryBugLabel(): Promise<string> {
 }
 
 async function isBugsListing(): Promise<boolean> {
-	return new SearchQuery(location.search).includes(await getSearchQueryBugLabel());
+	return new SearchQuery(location.href).includes(await getSearchQueryBugLabel());
 }
 
 async function addBugsTab(): Promise<void | false> {
@@ -115,7 +115,7 @@ async function addBugsTab(): Promise<void | false> {
 	bugsCounter.title = '';
 
 	// Update Bugs’ link
-	new SearchQuery(bugsTab).add(await getSearchQueryBugLabel());
+	bugsTab.href = SearchQuery.from(bugsTab).add(await getSearchQueryBugLabel()).href;
 
 	// In case GitHub changes its layout again #4166
 	if (issuesTab.parentElement instanceof HTMLLIElement) {
--- a/source/features/bugs-tab.tsx
+++ b/source/features/bugs-tab.tsx
@@ -71,7 +71,7 @@ async function getSearchQueryBugLabel(): Promise<string> {
 }
 
 async function isBugsListing(): Promise<boolean> {
-	return new SearchQuery(location.href).includes(await getSearchQueryBugLabel());
+	return SearchQuery.from(location).includes(await getSearchQueryBugLabel());
 }
 
 async function addBugsTab(): Promise<void | false> {
--- a/source/features/extend-conversation-status-filters.tsx
+++ b/source/features/extend-conversation-status-filters.tsx
@@ -33,7 +33,7 @@ function addMergeLink(): void {
 		// In this case, `lastLink` is expected to be a "Closed" link
 		const mergeLink = lastLink.cloneNode(true);
 		mergeLink.textContent = 'Merged';
-		mergeLink.classList.toggle('selected', new SearchQuery(location.href).includes('is:merged'));
+		mergeLink.classList.toggle('selected', SearchQuery.from(location).includes('is:merged'));
 		mergeLink.href = SearchQuery.from(mergeLink).replace('is:closed', 'is:merged').href;
 		lastLink.after(' ', mergeLink);
 	}
--- a/source/features/extend-conversation-status-filters.tsx
+++ b/source/features/extend-conversation-status-filters.tsx
@@ -34,7 +34,7 @@ function addMergeLink(): void {
 		const mergeLink = lastLink.cloneNode(true);
 		mergeLink.textContent = 'Merged';
 		mergeLink.classList.toggle('selected', new SearchQuery(location.search).includes('is:merged'));
-		new SearchQuery(mergeLink).replace('is:closed', 'is:merged');
+		new SearchQuery(mergeLink).replace('is:closed', 'is:merged').applyChanges();
 		lastLink.after(' ', mergeLink);
 	}
 }
@@ -49,7 +49,7 @@ function togglableFilters(): void {
 				'is:closed',
 				'is:merged',
 				'is:unmerged',
-			);
+			).applyChanges();
 		}
 	}
 }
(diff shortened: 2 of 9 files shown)


cheap-glitch (commented):
Changes:

- the query parts are stored to avoid re-splitting and re-joining the query string every time it's modified
- the link's URL is only updated when calling the new `.applyChanges()` method
- `.edit()` now operates on the query parts array
- added a test case for label links

source/features/extend-conversation-status-filters.tsx
```diff
@@ -34,7 +34,7 @@ function addMergeLink(): void {
 		const mergeLink = lastLink.cloneNode(true);
 		mergeLink.textContent = 'Merged';
 		mergeLink.classList.toggle('selected', new SearchQuery(location.search).includes('is:merged'));
-		new SearchQuery(mergeLink).replace('is:closed', 'is:merged');
+		new SearchQuery(mergeLink).replace('is:closed', 'is:merged').applyChanges();
```
fregante, line 37: Thank you! I was thinking about dropping the internal `link` altogether:


```suggestion
		mergeLink.href = SearchQuery.from(mergeLink)
			.replace('is:closed', 'is:merged')
			.href;
```

Internally:

`SearchQuery.from` is `() => return new SearchQuery(link.href)`

source/github-helpers/search-query.ts
```diff
@@ -84,56 +78,58 @@ export default class SearchQuery {
...
-		if (this.link && labelLinkRegex.test(this.link.pathname)) {
-			// Avoid a redirection to the conversation list that would erase the search query #5176
-			this.link.pathname = this.link.pathname.replace(/\/labels\/.+$/, '/issues');
+	applyChanges(): void {
+		this.searchParams.set('q', this.get());
+		if (this.link) {
+			this.link.search = String(this.searchParams);
+			if (labelLinkRegex.test(this.link.pathname)) {
+				// Avoid a redirection to the conversation list that would drop the search query #5176
+				this.link.pathname = this.link.pathname.replace(/\/labels\/.+$/, '/issues');
```
fregante, line 107: Since `SearchQuery` doesn't change the ink automatically anymore, you can probably just move this line to the `set` method. This way no changes are made to the url until they're requested (and also since `.set` isn't automatically called in the constructor anymore)

source/features/extend-conversation-status-filters.tsx
cheap-glitch: > I was thinking about dropping the internal `link` altogether

:+1:

> `SearchQuery.from` is `() => return new SearchQuery(link.href)`

We'd still need to pass the link element, or at least its `.pathname`, since some methods need to take a look at it.

source/github-helpers/search-query.ts
cheap-glitch: If we drop `this.link` then this will have to be extracted out of the class regardless.

fregante: No, we still have to deal with the whole URLs inside, it's just that we don't store/expose the original link

source/features/extend-conversation-status-filters.tsx
fregante: `link.href` uses the whole URL

source/github-helpers/search-query.ts
fregante: Firefox bug: https://github.com/refined-github/refined-github/pull/3208

cheap-glitch (commented):
Changes:

- `SearchQuery` now extends `URL`
- `SearchQuery.from()` is used to create a new instance from a link element
- `SearchQuery.fromSearchParams()` is used to create a new instance from a record object
- modifications to the `pathname` are done in the `.href` getter

cheap-glitch (commented):
Right, turns out we still can't extend `URL` in Firefox :grimacing: 

(Even though it seems to work when testing it directly in the web console…)

source/features/bugs-tab.tsx
```diff
@@ -71,7 +71,7 @@ async function getSearchQueryBugLabel(): Promise<string> {
 }
 
 async function isBugsListing(): Promise<boolean> {
-	return new SearchQuery(location.search).includes(await getSearchQueryBugLabel());
+	return new SearchQuery(location.href).includes(await getSearchQueryBugLabel());
```
fregante, line 74:
```suggestion
	return SearchQuery.from(location).includes(await getSearchQueryBugLabel());
```

source/features/extend-conversation-status-filters.tsx
```diff
@@ -33,8 +33,8 @@ function addMergeLink(): void {
 		// In this case, `lastLink` is expected to be a "Closed" link
 		const mergeLink = lastLink.cloneNode(true);
 		mergeLink.textContent = 'Merged';
-		mergeLink.classList.toggle('selected', new SearchQuery(location.search).includes('is:merged'));
-		new SearchQuery(mergeLink).replace('is:closed', 'is:merged');
+		mergeLink.classList.toggle('selected', new SearchQuery(location.href).includes('is:merged'));
```
fregante, line 36:
```suggestion
		mergeLink.classList.toggle('selected', SearchQuery.from(location).includes('is:merged'));
```

source/features/global-conversation-list-filters.tsx
```diff
@@ -25,15 +25,15 @@ function init(): void {
 		url.searchParams.set('q', `${typeQuery} ${defaultQuery} ${query}`);
 		const link = <a href={url.href} title={title} className="subnav-item">{label}</a>;
 
-		const isCurrentPage = new SearchQuery(location.search).includes(query);
+		const isCurrentPage = new SearchQuery(location.href).includes(query);
```
fregante, line 28:
```suggestion
		const isCurrentPage = SearchQuery.from(location).includes(query);
```

source/features/split-issue-pr-search-results.tsx
```diff
@@ -38,7 +35,7 @@ function init(): void {
 
 	const title = select('.codesearch-results h3')!.firstChild!;
 
-	const searchQuery = new SearchQuery(location.search);
+	const searchQuery = new SearchQuery(location.href);
```
fregante, line 38:
```suggestion
	const searchQuery = SearchQuery.from(location);
```

source/github-helpers/search-query.ts
```diff
@@ -35,102 +33,102 @@ export default class SearchQuery {
...
-			const nativeSet = this.searchParams.set;
-			this.searchParams.set = (name, value) => {
-				nativeSet.call(this.searchParams, name, value);
-				link.search = String(this.searchParams);
-			};
-		} else if (link instanceof URL) {
-			this.searchParams = link.searchParams;
-		} else {
-			this.searchParams = new URLSearchParams(link);
+	static from(link: HTMLAnchorElement): SearchQuery {
```
fregante, line 36: This can accept `Location` too, and maybe more.
```suggestion
	static from(link: Source): SearchQuery {
```

Perhaps you can merge `fromSearchParams` back into it

fregante (approved):

PR #5245, opened 2021-12-23
Meta: Move small bug report fields higher

fregante:
I always miss these 2 fields both when looking at an issue and when opening one. I feel that they're fundamental enough to be shown at the top.

Any other suggestions?


cheap-glitch (approved):
Maybe also move the two checkboxes (screenshot & console error) along with the textarea, since they're related to it?

fregante:
It's a-live

Screen Shot

kidonng:
> It's a-live. I also turned the URLs input into a textarea to allow multiple examples, but it's a bit too big and prominent now 😬 
> 
> Screen Shot
> 
> 

Maybe we can apply some styles to make it shorter, only on that page?

PR #5211, opened 2021-12-15
Fix handling of slashed branches in some commits list features

cheap-glitch:
Fixes #4492
Fixes #4196

## Test URLs

 * [Branch commits list](https://togithub.com/refined-github/sandbox/commits/branch/with/slashes)
 * [Renamed file history](https://togithub.com/refined-github/sandbox/commits/branch/with/slashes/renamed.md)

## Screenshot

`follow-file-renames`

image

`list-prs-for-branch`

image

PR #5246, opened 2021-12-26
Fix branch handling on commit list without a branch

### LINKED ISSUE #5240: `default-branch-button` does not work with long branch names that are truncated in the UI
### Please ensure:

### Include in this issue:

- [ ] Console errors, if any

### The Problem

Press the `default-branch-button` in the example url, and it will redirect you to [`https://github.com/gamemaker1/express-rate-limit/commits/opiniated-impr...`](https://github.com/gamemaker1/express-rate-limit/commits/opiniated-impr...). This is because the branch name is truncated in the `#branch-select-menu` div.

### Possible Fix

Change the selector in the `getDefaultBranch` helper function that retrieves the branch name from the UI:

https://github.com/refined-github/refined-github/blob/main/source/github-helpers/get-default-branch.ts#L26

From:

```ts
select('#branch-select-menu [data-menu-button]');
```

To:

```ts
select('#branch-select-menu [data-menu-button]:not(.css-truncate-target)');
```

So that it makes an API call and gets the default branch instead of using the truncated name from the UI (an alternative would be getting the branch from somewhere else in the DOM).

Not tested, but I think this should work.

### Example URL

https://github.com/gamemaker1/test/tree/very-very-long-long-long-long-branch-name

### Browser(s) used

Firefox

kidonng:
This PR actually comes in 3 parts:

1. `getDefaultBranch()` seems to always fallback to using API as the following code fails due to DOM not ready

   https://github.com/refined-github/refined-github/blob/a665157c410da81e9fdd83dcad945cd7085fafa4/source/github-helpers/get-default-branch.ts#L25-L30

   This PR fixes it with `elementReady()`.

2. The above method may return a truncated branch name.
   This PR uses a better method. Fix #5240.

3. `getDefaultBranch()` and `getCurrentCommittish()` doesn't work on commit list without a branch, namely `/repo/user/commits` page.
   This PR fixes it by restoring and using `getCurrentBranchFromFeed()` (renamed), which parses the current branch from the feed link. Reasons to do so:
   - I'm aware of #5211, but this is safe since we only do it on commit lists which always has a predictable feed link.
   - Using DOM is better than calling API.
   - The trick used in 1. does not suit `getCurrentCommittish()` since the latter is a synchronous function.

## Test URLs

1. This can only be tested after you added some indicators whether `getDefaultBranch()` uses API or not (e.g. `console.log()`). After that, test it on:

   - https://github.com/refined-github/refined-github/commits

2. With 1. tested, visit https://github.com/gamemaker1/test, then check if `default-branch-button` appears on https://github.com/gamemaker1/test/tree/very-very-long-long-long-long-branch-name or not.
3. Check if `default-branch-button` appears on https://github.com/refined-github/refined-github/commits.

### DIFF
--- a/source/github-helpers/get-default-branch.ts
+++ b/source/github-helpers/get-default-branch.ts
@@ -1,5 +1,6 @@
 import cache from 'webext-storage-cache';
 import select from 'select-dom';
+import elementReady from 'element-ready';
 import * as pageDetect from 'github-url-detection';
 
 import * as api from './api';
@@ -12,6 +13,16 @@ import {getRepo} from '.';
 // "This branch is 1 commit ahead, 27 commits behind master."
 const branchInfoRegex = /([^ ]+)\.$/;
 
+export function getCurrentBranchFromFeed(): string {
+	const feedLink = select('link[type="application/atom+xml"]')!;
+	return new URL(feedLink.href)
+		.pathname
+		.split('/')
+		.slice(4) // Drops the initial /user/repo/route/ part
+		.join('/')
+		.replace(/\.atom$/, '');
+}
+
 const getDefaultBranch = cache.function(async function (repository?: pageDetect.RepositoryInfo): Promise<string> {
 	if (arguments.length === 0) {
 		repository = getRepo();
@@ -23,10 +34,14 @@ const getDefaultBranch = cache.function(async function (repository?: pageDetect.
 
 	if (arguments.length === 0 || JSON.stringify(repository) === JSON.stringify(getRepo())) {
 		if (pageDetect.isRepoHome()) {
-			const currentBranch = select('#branch-select-menu [data-menu-button]');
+			const currentBranch = await elementReady('[data-hotkey="w"]');
 			if (currentBranch) { // If missing, it'll default to the API call
-				return currentBranch.textContent!;
+				return currentBranch.title;
 			}
+		}
+
+		if (pageDetect.utils.getRepositoryInfo()!.path === 'commits') {
+			return getCurrentBranchFromFeed();
 		}
 
 		if (!pageDetect.isForkedRepo()) {
--- a/source/github-helpers/get-default-branch.ts
+++ b/source/github-helpers/get-default-branch.ts
@@ -30,8 +30,9 @@ const getDefaultBranch = cache.function(async function (repository?: pageDetect.
 			}
 		}
 
-		if (pageDetect.utils.getRepositoryInfo()!.path === 'commits') {
-			return getCurrentBranchFromFeed();
+		const defaultBranch = getCurrentBranchFromFeed();
+		if (defaultBranch) {
+			return defaultBranch;
 		}
 
 		if (!pageDetect.isForkedRepo()) {
--- a/source/github-helpers/get-default-branch.ts
+++ b/source/github-helpers/get-default-branch.ts
@@ -4,7 +4,7 @@ import elementReady from 'element-ready';
 import * as pageDetect from 'github-url-detection';
 
 import * as api from './api';
-import {getRepo} from '.';
+import {getRepo, getCurrentBranchFromFeed} from '.';
 
 // This regex should match all of these combinations:
 // "This branch is even with master."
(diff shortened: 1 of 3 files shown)


kidonng:
What is this any way 🤷

```
source/github-helpers/api.ts(46,32): error TS2304: Cannot find name 'AnyObject'.
source/github-helpers/api.ts(55,12): error TS2304: Cannot find name 'AnyObject'.
 source/github-helpers/api.ts(170,12): error TS2304: Cannot find name 'AnyObject'.
```

cheap-glitch:
> What is this any way :shrug:
> 
> ```
> source/github-helpers/api.ts(46,32): error TS2304: Cannot find name 'AnyObject'.
> source/github-helpers/api.ts(55,12): error TS2304: Cannot find name 'AnyObject'.
>  source/github-helpers/api.ts(170,12): error TS2304: Cannot find name 'AnyObject'.
> ```

I think this happens because you added an import from `get-default-branch.ts` in `./github-helpers/index.ts`, which in turn imports `api.ts` — so now the tests have to import it too, and AVA doesn't seem to be aware of `globals.d.ts`.

source/github-helpers/get-default-branch.ts
```diff
@@ -23,12 +24,16 @@ const getDefaultBranch = cache.function(async function (repository?: pageDetect.
 
 	if (arguments.length === 0 || JSON.stringify(repository) === JSON.stringify(getRepo())) {
 		if (pageDetect.isRepoHome()) {
-			const currentBranch = select('#branch-select-menu [data-menu-button]');
+			const currentBranch = await elementReady('[data-hotkey="w"]');
 			if (currentBranch) { // If missing, it'll default to the API call
-				return currentBranch.textContent!;
+				return currentBranch.title;
```
fregante, line 29: It feels like we changed this in the past in the opposite direction, I don't know why. I may be wrong. @yakov116 remember anything?

source/github-helpers/get-default-branch.ts
```diff
@@ -23,12 +24,16 @@ const getDefaultBranch = cache.function(async function (repository?: pageDetect.
...
 			}
 		}
 
+		if (pageDetect.utils.getRepositoryInfo()!.path === 'commits') {
```
fregante, line 33: isRepoCommitList?

source/github-helpers/index.ts
```diff
@@ -16,6 +16,16 @@ export const getConversationNumber = (): string | undefined => {
 	return undefined;
 };
 
+export function getCurrentBranchFromFeed(): string {
```
fregante, line 19: Don't we have this function elsewhere? If not, did we drop it and why?

source/github-helpers/get-default-branch.ts
kidonng: No, we can only ensure we are on the default branch if it is exactly `/owner/repo/commits`. `/owner/repo/commits/branch` is not deterministic.

source/github-helpers/index.ts
kidonng: See description of part 3

source/github-helpers/get-default-branch.ts
yakov116: I don't recall anything about this other then using the feed link (which broke)

source/github-helpers/index.ts
fregante: I don’t think that answers my question 😜 

I found this:

- it was dropped in https://github.com/refined-github/refined-github/pull/4193, seemingly without a specific reason (other than not being useful anymore)

So restoring should be fine.  👍 

I'm hesitant to modify this function because we touched it _a lot_ over the years.

Some possible safety improvements:

- move the isRepoCommitList check to this function

source/github-helpers/get-default-branch.ts
fregante: Oh, yeah, specifically we can only know the default branch’s name if we _are_ on the default branch. Let's add this note to the code (inside `getCurrentBranchFromFeed` though)
```suggestion
		// Not `isRepoCommitList` because this works exclusively on the default branch
		if (pageDetect.utils.getRepositoryInfo()!.path === 'commits') {
```

source/github-helpers/index.ts
fregante: Nothing, because it does not know it. Void | string

source/github-helpers/get-default-branch.ts
```diff
@@ -23,12 +24,17 @@ const getDefaultBranch = cache.function(async function (repository?: pageDetect.
 
 	if (arguments.length === 0 || JSON.stringify(repository) === JSON.stringify(getRepo())) {
 		if (pageDetect.isRepoHome()) {
-			const currentBranch = select('#branch-select-menu [data-menu-button]');
+			const currentBranch = await elementReady('[data-hotkey="w"]');
 			if (currentBranch) { // If missing, it'll default to the API call
```
fregante, line 28: Does this ever happen? It feels like this should be an error. It's possible that I left it here as a fallback mechanism since it would break a lot of features, but maybe we should add a `console.error` after this condition just to make the error visible.

I'm focusing on this one because `if ! elementReady` is an antipattern which delays all the code after it until the `domready` event

fregante (approved):
It seems GTG other than the comment

Issue #4731, opened 2021-08-27
The user’s CSS is injected before Refined GitHub’s


fregante:
The CSS field in the options is supposed to be used as overrides to Refined GitHub's styles but it's actually injected earlier than RG's, so RG ends up overriding it instead. 

We need to find alternative solutions for the injection. For example:

- how does Stylus inject them?
- can it be registered via contentScript.register()? Does the behavior differ in Chrome's polyfill?
- is there any new way to register stylesheets? For example some CSSOM api

eqvinox:
>     * how does Stylus inject them?

Ostensibly in the same way — because it has exactly the same result, i.e. doesn't work either.  Screenshot:

![css2](https://user-images.githubusercontent.com/1715611/131217631-db6fe2f7-f6ce-4706-8faf-3f5e85179f0a.png)

fregante:
Wow. Thanks for checking and for the screenshot!

cheap-glitch:
> ...is injecting duplicate styles when navigating to different pages via AJAX:

Can't reproduce this on Firefox.

> can it be registered via contentScript.register()?

Yes we can register CSS using `contentScript.register()`, or use [`tab.insertCSS()`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/insertCSS) directly like your polyfill does.

> is there any new way to register stylesheets? For example some CSSOM api

I haven't found an API to register a whole stylesheet, but we could use [`insertRule()`](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet/insertRule) on the main RGH stylesheet to append the user-CSS to it. (Although that might mean manually splitting the CSS into several rules before injecting it.)

fregante:
> > ...is injecting duplicate styles when navigating to different pages via AJAX:
> 
> Can't reproduce this on Firefox.

I assume this is https://github.com/refined-github/refined-github/issues/4827#issuecomment-956768588

fregante:
> or use [`tab.insertCSS()`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/insertCSS) directly like your polyfill does.

That'd mean keeping track of each load from the background page. That API does it automatically (in Firefox)

> I haven't found an API to register a whole stylesheet

https://developers.google.com/web/updates/2019/02/constructable-stylesheets

Not sure if it's cross-browser yet

eqvinox:
[...]

PR #5258, opened 2022-01-02
Mute visited conversation links on conversation lists

kidonng:
When CSS resets became popular, `:visited` was one of the useful browser's native styles that got doomed.

This PR brings back `:visited` style to conversation links on conversation lists by using a muted color. It helps people recognize whether you have went through a conversation or not.

Note: `:visited` style can already be seen on Discussions (e.g. https://github.com/fish-shell/fish-shell/discussions). Not sure if it's a feature or GitHub just forgot to style them though.

## Test URLs

https://github.com/fish-shell/fish-shell/pulls

## Screenshot

<table>
<tr>
<tr>
	<td>
	<td>
</table>


kidonng:
Let me further explain the motivation behind this.

This PR is inspired by an feature seen on v2ex.com, where links to posts have the count of comments as their hashes (e.g. a post with 42 comments will have `#reply42` as the hash of its link). The site also uses a dimmed color for `:visited` links.

The result is that when you visit a post, the link will be dimmed. But if there is any new comment, the comment number will change, so does the hash. The `:visited` selector (or to say browser) considers the hash as well, so the color of the link becomes normal again.

@cheap-glitch I see you voted down this. Are you concerned that more and more links will be muted as the user visits the conversations, making the feature less useful and more annoying? My intention would be that the different colors will serve as a indicator of whether you have read a conversation or not, and whether there's any new activity after you read them, without the hassle of subscribtions.

I will implement the exact feature later, but here's a gist:

```js
for (const row of document.querySelectorAll('.js-issue-row')) {
    const count = row.querySelector('.octicon-comment + .text-small')?.textContent

    if (count) {
        row.querySelector('.h4').hash = 'rgh-comment-count-' + count
    }
}
```

fregante:
Duplicate of https://github.com/refined-github/refined-github/pull/1206 😃

yakov116:
@kidonng you know your on the right track if the king tried this too 😛

cheap-glitch:
> @cheap-glitch I see you voted down this. Are you concerned that more and more links will be muted as the user visits the conversations, making the feature less useful

Yes basically. But with the hash technique it might be interesting, although it still won't react to events other than comments (e.g. an approving review). Still, I'm curious to try it and see how it plays in practice.

kidonng:
> Duplicate of https://github.com/refined-github/refined-github/pull/1206 😃 

🤯 I DIDN'T EXPECT THAT! I feel so lashed out. But the idea is more than some simple CSS declarations, see my previous comment. 

The implementation will be only a few lines at most so it's not something that does little with much efforts.

> Yes basically. But with the hash technique it might be interesting, although it still won't react to events other than comments (e.g. an approving review). Still, I'm curious to try it and see how it plays in practice.

Technically we could put anything into the hash, including open/close state, approvals etc.

fregante:
> @kidonng you know your on the right track if the king tried this too 😛

lol slow down, it was rejected by the land owner

> I will implement the exact feature later, but here's a gist:

I don’t think it's worth it, after all you did _not_ open that specific/important linked comment, so it's worth to follow the browser’s logic here.

This PR is ok for me obviously, but let's see 😅

kidonng:
> after all you did _not_ open that specific/important linked comment, so it's worth to follow the browser’s logic here.

That's an artificial hash, not the actual hash to some comment (and there's no way to get that info on the conversation list, which is the only place the feature would run).

fregante:
I still don't think the extra JS is worth it, especially because that's not the default browser behavior.

fregante (changes requested):
I still don't think the extra JS is worth it, especially because that's not the default browser behavior.

kidonng:
> I still don't think the extra JS is worth it, especially because that's not the default browser behavior.

Fine, I will revert but the implementation will stay in this PR's history, in case someone it's interested in the future.

cheap-glitch:
If it stays CSS-only then I'm -1 on this. I don't really see the point, it doesn't convey any meaningful information apart from "I opened this discussion at least once at some point". I may be missing something but I can't see why that'd be useful to know, except maybe when going through old conversations that aren't updated anymore.

And yes, I know it can easily be disabled via custom CSS, but imho it doesn't make sense as a default in the first place.

kidonng:
So... we both have our stands. I have been using the styles for some time, as a person who often digs into many conversations, and find it helpful while unlikely to be annoying.

Don't get me wrong, I still love hash idea and will try making it a user script.

It seems there's going to be a release recently so let's merge this and see. After all it's just a single CSS rule, very easy to override locally or via hotfix.

cheap-glitch:
> After all it's just a single CSS rule, very easy to override locally or via hotfix

Note: because of #4731, the override selector needs to be something like:

```css
.js-issue-row .h4.h4:visited:not(:hover) {
	color: var(--color-fg-default) !important;
}
```

PR #5278, opened 2022-01-08
Change `hide-repo-badges` into a feature that can be disabled

### LINKED ISSUE #5268: Make "repo badges hiding" optional
### Description

Provide capability to keep on showing the default GitHub repo visibility badge which is safer to always show instead of just an icon.

Having this as an opt-in is fine for me.

### Screenshot

image

### Example URL

N/A

cheap-glitch:
Fixes #5268 

## Test URLs

 * https://togithub.com/refined-github/refined-github
 * https://togithub.com/fregante

## Screenshot

image

**Feature screenshot**

hide-repo-badges

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -178,6 +178,7 @@ Thanks for contributing! 🦋🙌
 - [](# "quick-repo-deletion") [Lets you delete your repos in a click, if they have no stars, issues, or PRs.](https://user-images.githubusercontent.com/1402241/99716945-54a80a00-2a6e-11eb-9107-f3517a6ab1bc.gif)
 - [](# "useful-forks") [Helps you find the most active forks](https://user-images.githubusercontent.com/38117856/107463541-542e8500-6b2c-11eb-8b25-082f344c1587.png), via https://useful-forks.github.io.
 - [](# "clean-repo-tabs") [Moves the "Security" and "Insights"  to the repository navigation dropdown. Also moves the "Projects", "Actions" and "Wiki" tabs if they're empty/unused.](https://user-images.githubusercontent.com/16872793/124681343-4a6c3c00-de96-11eb-9055-a8fc551e6eb8.png)
+- [](# "hide-repo-badges") [Hide repo badges ("Public", "Template", etc.) in the repo header and pinned repo cards.](TODO)
 
 <!-- Refer to style guide above. Keep this message between sections. -->
 
--- a/readme.md
+++ b/readme.md
@@ -178,7 +178,7 @@ Thanks for contributing! 🦋🙌
 - [](# "quick-repo-deletion") [Lets you delete your repos in a click, if they have no stars, issues, or PRs.](https://user-images.githubusercontent.com/1402241/99716945-54a80a00-2a6e-11eb-9107-f3517a6ab1bc.gif)
 - [](# "useful-forks") [Helps you find the most active forks](https://user-images.githubusercontent.com/38117856/107463541-542e8500-6b2c-11eb-8b25-082f344c1587.png), via https://useful-forks.github.io.
 - [](# "clean-repo-tabs") [Moves the "Security" and "Insights"  to the repository navigation dropdown. Also moves the "Projects", "Actions" and "Wiki" tabs if they're empty/unused.](https://user-images.githubusercontent.com/16872793/124681343-4a6c3c00-de96-11eb-9055-a8fc551e6eb8.png)
-- [](# "hide-repo-badges") [Hide repo badges ("Public", "Template", etc.) in the repo header and pinned repo cards.](TODO)
+- [](# "hide-repo-badges") [Hide repo badges ("Public", "Template", etc.) in the repo header and pinned repo cards.](https://user-images.githubusercontent.com/46634000/148645393-d3fa4ca9-6df4-4bb2-b810-7d73cc057772.png)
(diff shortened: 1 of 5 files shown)


tomkerkhove:
Thank you!

fregante (approved):

RussKie:
Can we please have this option opt-in instead of opt-out (i.e. unchecked by default)? I work across public and private repos, and https://github.com/refined-github/refined-github/issues/4767 caused my serious distress.

cheap-glitch:
> Can we please have this option opt-in instead of opt-out (i.e. unchecked by default)?

Sorry for the stress it caused you, but we can't really do that. If a feature is disabled by default, most users won't enable it and it'll be essentially pointless. Not to mention that everyone will have different opinion on what the defaults should be, and we'll never see the end of it.

There's a comprehensive list of "togglable" features in the readme and detailed notes for each release, so it's not like we don't provide any information on what the extension will do if you choose to install it.

That said, I understand it's a lot to go through, and you're not the first to raise this kind of concern. What we could do, maybe:
- add a notice in the "Install" section of the readme, encouraging first-time user to go through the list of features at least once and disable what they don't like
- mark the "controversial" or "dangerous" features with a little emoji, like :fire: does for the highlights (I think that what already proposed elsewhere?)

RussKie:
The issue here is that the hidden "private" label can lead to a security event such as accidental private information disclosure. This is why I ask to reconsider and make it opt-in feature, so that users who do not want to see "public" label can assume all responsibility for hiding the label.

> What we could do, maybe:
> 
> * add a notice in the "Install" section of the readme, encouraging first-time user to go through the list of features at least once and disable what they don't like
> * mark the "controversial" or "dangerous" features with a little emoji, like 🔥 does for the highlights (I think that what already proposed elsewhere?)

I can't really see how this would help. I'm a long time user of the extension, and it gets auto-updated. As long as everything works as I've learned to expect - I have no need (nor time, to be honest) to go and randomly check if there are any new features or toggles that I need to tweak.

cheap-glitch:
> I'm a long time user of the extension, and it gets auto-updated.

This can be disabled on a per-extension basis (at least in Firefox).

> I have no need (nor time, to be honest) to go and randomly check if there are any new features or toggles that I need to tweak.

You can watch the repo for new releases only, and get notified when we do one. We release quite infrequently (once every 2 or 3 weeks on average) and the changelogs are relatively short ­— so it'll take a few minutes twice a month to keep up to date. I don't think that's too much to ask.

It's not like I don't understand where you're coming from, I do a lot of "blind updates" too. But it's a gamble on the user's part, and imo this isn't the devs' responsibility to shoulder — as long as they communicated properly. Which I think we do, but I'm sure we can always improve on that part.

fregante:
> I have no need (nor time, to be honest) to go and randomly check if there are any new features or toggles that I need to tweak.

Such is life on the internet. We wish we could just leave Refined GitHub alone but GitHub itself changes daily. Even without new features, we often make substantial changes to existing features to make them work better or adapt to GitHub’s changes.

As for this feature specifically, what's the issue? We're only hiding the label when it's duplicate information: https://github.com/refined-github/refined-github/issues/4767#issuecomment-964361783

RussKie:
> As for this feature specifically, what's the issue? We're only hiding the label when it's duplicate information: [#4767 (comment)](https://github.com/refined-github/refined-github/issues/4767#issuecomment-964361783)

The fact that the icons between public and private repos aren't very different (especially if viewed on 4K+ monitors), and a user is unable to tell the visibility of a repo. I'm pretty sure it was called out in one of the related threads. 
Besides the security aspect (now it is unclear what context a user is in), there is an accessibility issue as well. I'm sure these two are the reason why GitHub displays the text label and the icon. 
This change is the extension (great otherwise) essentially made the overall developer experience out of the box both less accessible and less secure.

fregante:
I disagree. The text inside the label is smaller than the icon. The lock icon is visually very different from the repo and bigger than the few pixels between <sub>"public"</sub> and <sub>"private"</sub>. The only thing you could complain about is that the marker *you* were using to tell private repos apart is gone, and that I agree with, but the market remains duplicate.

cheap-glitch:
> This change is the extension (great otherwise) essentially made the overall developer experience out of the box both less accessible

I wonder if we could detect via CSS if some a11y features are enabled, and disable the feature (and maybe others?) in that case.

kidonng:
> I wonder if we could detect via CSS if some a11y features are enabled, and disable the feature (and maybe others?) in that case.

We can replicate the behavior of [`.sr-only`](https://primer.style/css/utilities/layout#screen-reader-only):

```css
.sr-only {
    position: absolute;
    width: 1px;
    height: 1px;
    padding: 0;
    overflow: hidden;
    clip: rect(0, 0, 0, 0);
    word-wrap: normal;
    border: 0
}
```

Julian-Dumitrascu:
> notified

To let people subscribe to smaller topics, we are creating a repository per topic.
It seems useful to let people subscribe to notifications about an issue with certain labels.
I've found no such feature in GitHub or in your extension.
Who wants to discuss such a goal, so that we agree on how we're going to achieve it?

fregante:
Please do not comment on random issues regarding unrelated requests, disregarding the banner that invites you to open a new issues.

PR #5284, opened 2022-01-09
Fix `ci-link` on Discussions and get from page if present

### LINKED ISSUE #5283: The list of CI checks in the repo header is broken on Discussion page
### Please ensure:

### Example URLs

https://github.com/cli/cli/discussions

### Browser(s) used

Firefox

### Description

image

Note that this doesn't happen when navigating using pjax, like when switching tabs in a repo.

### Include in this issue:

- [ ] Console errors, if any

kidonng:
This and #5246 both reduce requests when opening a new repo.
Fixes #5283 

## Test URLs

- `isRepoHome`: https://github.com/refined-github/refined-github
- `isRepoCommitList`
  - https://github.com/refined-github/refined-github/commits/main
  - https://github.com/refined-github/refined-github/commits (after #5246 gets merged)
- Fallback: https://github.com/refined-github/refined-github/issues
- Discussions: https://github.com/cli/cli/discussions (reload the page to see the bug)

## Screenshot

image

### DIFF
--- a/source/features/ci-link.tsx
+++ b/source/features/ci-link.tsx
@@ -5,7 +5,6 @@ import * as pageDetect from 'github-url-detection';
 import features from '.';
 import fetchDom from '../helpers/fetch-dom';
 import getDefaultBranch from '../github-helpers/get-default-branch';
-import {onRepoHomeCiDetailsLoad} from '../github-events/on-fragment-load';
 import {buildRepoURL, getCurrentCommittish} from '../github-helpers';
 
 // Look for the CI details dropdown in the latest 2 days of commits #2990
@@ -15,6 +14,11 @@ const ciDetailsSelector = [
 ].join(',');
 
 async function getCiDetails(): Promise<HTMLElement | undefined> {
+	if (pageDetect.isRepoHome()) {
+		// Clone the CI details if they're already loaded, otherwise clone the include-fragment
+		return select('.file-navigation + .Box :is(.commit-build-statuses, .js-details-container include-fragment[src*="/rollup?"])')!.cloneNode(true);
+	}
+
 	if (pageDetect.isRepoCommitList() && getCurrentCommittish() === await getDefaultBranch()) {
 		return select(ciDetailsSelector)!.cloneNode(true);
 	}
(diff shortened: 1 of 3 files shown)


kidonng:
Fixed styles on Discussion page in 716ca6457dc0871c7b0633a8626ebccfa8cf2c37. That can be in a separate PR, but would depend on this eventually (due to the huge refactoring), so I just put them together.

fregante:
I'm not sure I like the extra complexity, we have this option often but I generally avoid it unless it's super simple. 

In this case the repo doesn't have checks, the proposed code queries the dom up to 600 times until it times out, then it does an HTTP request. 

elementReady should only be used for elements already on the page. If the element is going to be deferred it's best we look for it's parent and then listen to the specific fragment load event.

source/features/ci-link.tsx
```diff
@@ -1,39 +1,80 @@
...
+	}
+
+	appendIconToRepoHeader(icon);
+}
+
+function initRepoHome(): void | false {
+	const icon = select('.file-navigation + .Box .commit-build-statuses');
+	if (!icon) {
+		return false;
+	}
```
cheap-glitch, line 52: I think it makes sense to have a separate init for the repo homepage, since the logic and selector are different.

That said, I wonder if we could just clone the `include-fragment` and let GitHub do the rest?

cheap-glitch (commented):
Refactored the PR to drop `elementReady` and clarify things a bit.

source/features/ci-link.tsx
cheap-glitch: > That said, I wonder if we could just clone the include-fragment and let GitHub do the rest?

Apparently yes

source/features/ci-link.tsx
```diff
@@ -1,37 +1,60 @@
...
+
+	return ciDetails;
+}
+
+function appendCiDetailsToRepoTitle(ciDetails: HTMLElement): void {
 	// Append to title (aware of forks and private repos)
 	const repoNameHeader = select('[itemprop="name"]')!.parentElement!;
-	repoNameHeader.append(icon);
+	repoNameHeader.append(ciDetails);
```
cheap-glitch, line 39: Took the opportunity to do some renaming, since we're not really dealing with an icon in this case.

While we're at it, should we rename the feature `ci-details`? Or even `repo-header-ci-details`?

kidonng:
Well done @cheap-glitch, looks much cleaner, especially the trick of cloning `<include-fragment>`.

source/features/ci-link.tsx
kidonng: `repo-header-` prefix is fine but I feel "details" could be misunderstood that we are providing more details to CI status, which is not true. "Icon" is more obvious here, just my opinion.

source/features/ci-link.tsx
kidonng: Are there any other features that can make use of this trick?

cheap-glitch: > Are there any other features that can make use of this trick?

I wonder, but I doubt it. Usually we need to select/modify something in the `include-fragment`, so we have no choice but to wait for it to load.

source/features/ci-link.tsx
```diff
@@ -1,29 +1,47 @@
...
+// Look for the CI details dropdown in the latest 2 days of commits #2990
+const ciDetailsSelector = [
+	'.TimelineItem--condensed:nth-of-type(-n+2) .commit-build-statuses', // TODO[2022-04-29]: GHE #4294
+	'.TimelineItem--condensed:nth-of-type(-n+2) batch-deferred-content[data-url$="checks-statuses-rollups"]',
+].join(',');
+
+async function getCiDetails(): Promise<HTMLElement | undefined> {
+	if (pageDetect.isRepoHome()) {
+		// Clone the CI details if they're already loaded, otherwise clone the include-fragment
+		return select('.file-navigation + .Box :is(.commit-build-statuses, .js-details-container include-fragment[src*="/rollup?"])')!.cloneNode(true);
```
fregante, line 19: A commented multi-line array would be good

fregante (approved):

PR #5289, opened 2022-01-10
Add `same-branch-author-commits` feature

kidonng:
All author names link to `/owner/repo/commits?author=<author>`, that is, the branch is not considered at all. Seems like a GitHub bug, but very easy to fix.

~~The feature can exclude `getRepo()!.path === 'commits'` but there's no need.~~ Now more useful with PJAX!

Note that despite the name is inspired by `same-page-definition-jump`, it can not use `delegate` because the links are actually different.

## Test URLs

https://github.com/refined-github/refined-github/commits/hotfix

## Screenshot

Screen Shot 2022-01-10 at 20 07 55


fregante (approved):
Totally makes sense!

PR #5296, opened 2022-01-13
Add `prevent-pr-merge-panel-opening` feature

### LINKED ISSUE #4101: PRs merge box automatically open (after opening them once)
1. Open an PR you can merge
2. Click "Merge PR" to open the merge form, to look like this:

Screen Shot

From now on, every time you visit that PR, the form will be automatically open. This behavior probably stops once a new commit comes in or maybe when the session ends.

This isn't strictly caused by RG, but RG likely triggers some internal auto-opening behavior of GitHub. It might be due to `clear-pr-merge-commit-message`

cheap-glitch:
Resolves #4101 

## Test URLs

https://github.com/refined-github/sandbox/pull/4

Open the merge box and reload the page (or click on a tab and come back) to see the behavior.

## Screenshot

https://user-images.githubusercontent.com/46634000/149317428-338673be-c42a-4895-993e-728a38716c86.mp4

### DIFF
--- a/source/features/prevent-pr-merge-panel-opening.tsx
+++ b/source/features/prevent-pr-merge-panel-opening.tsx
@@ -1,0 +1,36 @@
+import select from 'select-dom';
+import * as pageDetect from 'github-url-detection';
+
+import features from '.';
+
+let controller: AbortController;
+
+async function sessionResumeHandler(): Promise<void> {
+	await Promise.resolve(); // The `session:resume` event fires a bit too early
+	const cancelMergeButton = select('.merge-branch-form .js-details-target');
+	if (cancelMergeButton) {
+		cancelMergeButton.click();
+		controller.abort();
+	}
+}
+
+function init(): void {
+	controller = new AbortController();
+	document.addEventListener('session:resume', sessionResumeHandler, {signal: controller.signal});
+}
+
+void features.add(import.meta.url, {
+	asLongAs: [
+		// The user is a maintainer, so they can probably merge the PR
+		() => select.exists('.discussion-sidebar-item .octicon-lock'),
+	],
+	include: [
+		pageDetect.isPRConversation,
+	],
+	exclude: [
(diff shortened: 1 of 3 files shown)


source/features/prevent-pr-merge-panel-opening.tsx
```diff
@@ -0,0 +1,36 @@
+import select from 'select-dom';
+import * as pageDetect from 'github-url-detection';
+
+import features from '.';
+
+let controller: AbortController;
+
+async function sessionResumeHandler(): Promise<void> {
+	await Promise.resolve(); // The `session:resume` event fires a bit too early
```
cheap-glitch, line 9: Taken from `on-pr-merge-panel-open.ts`

source/features/prevent-pr-merge-panel-opening.tsx
```diff
@@ -0,0 +1,36 @@
...
+	const cancelMergeButton = select('.merge-branch-form .js-details-target');
+	if (cancelMergeButton) {
+		cancelMergeButton.click();
+		controller.abort();
+	}
```
cheap-glitch, line 14: Several `session:resume` events are fired upon page load, each with the same metadata, so (afaict) they're virtually indistinguishable.

From what I can tell the merge panel DOM will only be injected after the second one, hence the check and the abort controller (since we can't just use `{once: true}` in `addEventListener`).

cheap-glitch (commented):
Afaict this behavior is not caused by RGH, it's indeed a native GitHub bug.

source/features/prevent-pr-merge-panel-opening.tsx
```diff
@@ -0,0 +1,36 @@
...
+}
+
+function init(): void {
+	controller = new AbortController();
+	document.addEventListener('session:resume', sessionResumeHandler, {signal: controller.signal});
```
fregante, line 19: It's easier to remove the handler via removeEventHandler than having a global controller

fregante (approved):

PR #5297, opened 2022-01-13
Fix default branch getter

kidonng:
It seems that the branch selector's `title` will only include the full branch name **only** when the displayed branch name is truncated. Otherwise, it becomes "Switch branches or tags". This causes the default branch getter to return a wrong result most of the time.

When making #5246 I may have forgotten to clear extension cache on https://github.com/refined-github/refined-github, so somehow I didn't notice this.

This PR also addresses https://github.com/refined-github/refined-github/pull/5246#discussion_r781721667.

## Test URLs

1. Clear extension cache
2. Go to https://github.com/refined-github/refined-github
3. Click "<number> commits"
4. The `default-branch-button` should not be there

## Screenshot

<table>
<tr>
<tr>
	<td>
	<td>
</table>

### DIFF
--- a/source/github-helpers/get-default-branch.ts
+++ b/source/github-helpers/get-default-branch.ts
@@ -30,8 +30,6 @@ const getDefaultBranch = cache.function(async function (repository?: pageDetect.
 					? branchSelector.textContent!.trim()
 					: branchSelector.title;
 			}
-
-			console.error(`Could not find the branch selector. Calling API to get the default branch of ${repository.nameWithOwner}.`);
 		}
 
 		const defaultBranch = getCurrentBranchFromFeed();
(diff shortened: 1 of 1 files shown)


kidonng:
@fregante Sorry for causing the trouble, but this is a pretty severe bug. Maybe release a new version after this got merged?

source/github-helpers/get-default-branch.ts
```diff
@@ -24,10 +24,14 @@ const getDefaultBranch = cache.function(async function (repository?: pageDetect.
...
-			if (currentBranch) { // If missing, it'll default to the API call
-				return currentBranch.title;
+			const branchSelector = await elementReady('[data-hotkey="w"]');
+			if (branchSelector) {
+				return branchSelector.title === 'Switch branches or tags'
+					? branchSelector.textContent!.trim()
+					: branchSelector.title;
 			}
+
+			console.error(`Could not find the branch selector. Calling API to get the default branch of ${repository.nameWithOwner}.`);
```
yakov116, line 34: Why do we need the user to see this? If you need it for debugging we have a logger.

kidonng: Maybe I misunderstood what @fregante means 🤷 the reason I do this is more about unwilling to import `features` here.

But perhaps I should just export and import `logError` here...

yakov116 (approved):

yakov116:
@kidonng feel free to merge. Lets wait another 2 hours before shipping to see if we want to include anything else.

yakov116:
@fregante can we please release this?

fregante:
Done

PR #881, opened 2017-12-17
Document the option for disabling individual features

sindresorhus:
Follow-up to #877

PR #5298, opened 2022-01-13
Meta: require cleared cache and extension version in bug reports

kidonng:
These are routine questions we probably should ask in the template.

## Test URLs

https://github.com/kidonng/refined-github/issues/new?assignees=&labels=bug&template=1_bug_report.yml

## Screenshot

image

### DIFF
--- a/.github/ISSUE_TEMPLATE/1_bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/1_bug_report.yml
@@ -14,11 +14,19 @@ body:
       options:
         - label: The bug is caused by Refined GitHub. It doesn't happen if I disable the extension.
           required: true
+      options:
+        - label: The bug is not caused by stale cache. It still happens after clearing extension cache.
+          required: true
   - type: textarea
     attributes:
       label: Example URLs
       description: A REAL URL where the bug appears. If it happens on a private repo, find an equivalent public URL, even if it doesn't happen there.
       placeholder: https://github.com/refined-github/refined-github/issues
+    validations:
+      required: true
+  - type: input
+    attributes:
+      label: Refined GitHub version
     validations:
       required: true
   - type: input
--- a/.github/ISSUE_TEMPLATE/1_bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/1_bug_report.yml
@@ -7,14 +7,14 @@ body:
       value: |
         # Thanks for reporting a bug! ⛰
 
-        Help us replicate the issue by filling in this form. You can use the "Find feature" button in the option to find which feature is causing the issue.
+        Help us replicate the issue by filling in this form. Use the "Find feature" button in the options page to find which feature is causing the issue.
   - type: checkboxes
     attributes:
       label: 'Please ensure:'
       options:
-        - label: The bug is caused by Refined GitHub. It doesn't happen if I disable the extension.
+        - label: The bug is caused by Refined GitHub. It doesn't happen if the extension is disabled.
           required: true
-        - label: The bug is not caused by stale cache. It still happens after clearing extension cache.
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/1_bug_report.yml
```diff
@@ -14,13 +14,20 @@ body:
       options:
         - label: The bug is caused by Refined GitHub. It doesn't happen if I disable the extension.
           required: true
+        - label: The bug is not caused by stale cache. It still happens after clearing extension cache.
+          required: true
```
cheap-glitch, line 18: Maybe it should mention where to find the "Clear cache" button?

kidonng: I'm not good at wording. Do you have any precise example?

cheap-glitch, line 18:
```suggestion
        - label: The bug is not caused by a stale cache. It still happens after clearing the extension cache.
          description: You can find the "Clear cache" button at the bottom of the options page.
          required: true
```

cheap-glitch:
```suggestion
        - label: The bug is not caused by a stale cache. It still happens after clearing the extension cache.
          description: You can find the "Clear cache" button at the bottom of the options page.
          required: true
```

readme.md
```diff
@@ -418,7 +418,7 @@ Thanks for contributing! 🦋🙌
 
 ## Customization
 
-Most features [can be disabled](https://github.com/refined-github/refined-github/pull/877) if they are JavaScript-based *(Experimental)* and you can override our CSS with your own in [the extension options.](https://github.com/refined-github/refined-github/pull/1193)
+Most features [can be disabled](https://github.com/refined-github/refined-github/pull/877) if they are JavaScript-based and you can override our CSS with your own in [the extension options.](https://github.com/refined-github/refined-github/pull/1193)
```
kidonng, line 421: Throwing in a small change here. "Experimental" has been here since #881, I assume it no longer applies after more than 4 years.

.github/ISSUE_TEMPLATE/1_bug_report.yml
kidonng: > Not sure if `description` works with checkboxes…

It doesn't work, so I put your words in the label directly.

cheap-glitch (approved):

source/features/index.tsx
```diff
@@ -70,6 +70,7 @@ const logError = (url: string, error: unknown): void => {
 	newIssueUrl.searchParams.set('template', '1_bug_report.yml');
 	newIssueUrl.searchParams.set('title', `\`${id}\`: ${message}`);
 	newIssueUrl.searchParams.set('example_url', location.href);
```
cheap-glitch, line 72: Shouldn't this be `example_urls` to match with the label?

fregante (approved):
Good idea! As for the version filling I hope we can automate it via JS on the actual page rather than only fill that link in. Most of the issues are opened on GitHub.com

Another change I'd like to make is in the enhancement template: we're no longer in a feature freeze. However I'd like to keep the mention of "good first issues"

cheap-glitch:
> I hope we can automate it via JS on the actual page rather than only fill that link in

> Heck we could even add a "clear cache" button on the New Issue page on our repo

Golden thumbs up

source/features/index.tsx
kidonng: This was once correct, but https://github.com/refined-github/refined-github/pull/5245/files#diff-31bc308e7c7d86220613f3cbc05182706b71911a5d654362c32ea919622764d2R19 didn't change this.

PR #5305, opened 2022-01-14
Make `dim-visited-conversations` a CSS feature

### LINKED ISSUE #5299: Issue titles in issues tab are dimmed / gray
### Please ensure:

### Example URLs

https://github.com/makeworld-the-better-one/amfora/issues

### Browser(s) used

Firefox

### Description

Issue titles are grayed out, only in the issues tab:

image

### Include in this issue:

- [ ] Console errors, if any

kidonng:
Proposed to fix #5299

Make #5258 a standalone **CSS** feature, ~~and disable it on repos the user has write access to~~.

## Test URLs

https://github.com/refined-github/refined-github/issues ~~or any repo you can edit~~

## Screenshot

N/A

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -267,6 +267,7 @@ Thanks for contributing! 🦋🙌
 - [](# "dim-bots") [Dims commits and PRs by bots to reduce noise.](https://user-images.githubusercontent.com/1402241/65263190-44c52b00-db36-11e9-9b33-d275d3c8479d.gif)
 - [](# "esc-to-cancel") [Adds a shortcut to cancel editing a conversation title: <kbd>esc</kbd>.](https://user-images.githubusercontent.com/35100156/98303086-d81d2200-1fbd-11eb-8529-70d48d889bcf.gif)
 - [](# "no-duplicate-list-update-time") [Hides the update time of conversations in lists when it matches the open/closed/merged time.](https://user-images.githubusercontent.com/1402241/111357166-ac3a3900-864e-11eb-884a-d6d6da88f7e2.png)
+- [](# "dim-visited-conversations") [Dims titles of visited conversations, unless you have write access to the repo.](https://user-images.githubusercontent.com/44045911/148573545-39a72dab-a37a-491d-83f3-d6fd015f5dcf.mov)
 
 <!-- Refer to style guide above. Keep this message between sections. -->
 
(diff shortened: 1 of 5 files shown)


cheap-glitch:
I still feel like this is too opinionated to be a RGH feature, even if it's restricted to repos you don't own ­— which is the majority of them.

But let's see what people think :smiley:

kidonng:
> I still feel like this is too opinionated to be a RGH feature

I don't want to see it goes down like this either, but remember I intended to make it more useful, however that was deemed unworthy in https://github.com/refined-github/refined-github/pull/5258#pullrequestreview-847215858.

fregante:
Sorry this doesn't make sense. :visited is a native browser feature. I don't see the need to customize it any further.

kidonng:
This is only a proposed solution, not the final decision. The need to make it a standalone option (or drop it) is still there.

readme.md
```diff
@@ -267,7 +267,7 @@ Thanks for contributing! 🦋🙌
 - [](# "dim-bots") [Dims commits and PRs by bots to reduce noise.](https://user-images.githubusercontent.com/1402241/65263190-44c52b00-db36-11e9-9b33-d275d3c8479d.gif)
 - [](# "esc-to-cancel") [Adds a shortcut to cancel editing a conversation title: <kbd>esc</kbd>.](https://user-images.githubusercontent.com/35100156/98303086-d81d2200-1fbd-11eb-8529-70d48d889bcf.gif)
 - [](# "no-duplicate-list-update-time") [Hides the update time of conversations in lists when it matches the open/closed/merged time.](https://user-images.githubusercontent.com/1402241/111357166-ac3a3900-864e-11eb-884a-d6d6da88f7e2.png)
-- [](# "dim-visited-conversations") [Dims the title of visited conversations, unless you have write access to the repo.](https://user-images.githubusercontent.com/44045911/148573545-39a72dab-a37a-491d-83f3-d6fd015f5dcf.mov)
+- [](# "dim-visited-conversations") [Dims the title of visited conversations.](https://user-images.githubusercontent.com/44045911/149634307-1ed0e554-6957-4afa-8458-9c5bb4f80ebb.mp4)
```
fregante, line 270: The video does not load in Safari for whatever reason. I think it's best to stick with gifs and revert the previous update. I don't want to have to deal with codec issues, gifs just work

Screen Shot

fregante (changes requested):

fregante:
Merging this but the video should still be undone in the lint PR or a PR to revert the whole video feature.

readme.md
kidonng: Previous video featured hash changes which is no longer true, so I rerecorded the video.

And I took the chance to convert it into `.mp4` with `ffmpeg` since Chrome doesn't open `.mov` files if opened directly (or in our case, click the link in README). It's fine if they are embedded on GitHub though.

I'm surprised that Safari doesn't support the converted video because macOS does. Maybe there's something wrong with how I convert it, I will take a look later today.

fregante: It's not too much because, again, I don't want to have debug/review this in the future. Gifs made with licecap are super lightweight too

PR #5310, opened 2022-01-14
Clear `forked-to` cache after `quick-repo-deletion`

kidonng:
I usually try not to correlate features together, but this is a convenient and harmless improvement we can do.

## Test URLs

1. Fork https://github.com/octocat/Spoon-Knife
2. Visit your fork
3. Delete the fork using `quick-repo-deletion`
4. Return to https://github.com/octocat/Spoon-Knife, `forked-to` should not be present

## Screenshot

N/A

### DIFF
--- a/source/features/quick-repo-deletion.tsx
+++ b/source/features/quick-repo-deletion.tsx
@@ -1,6 +1,7 @@
 import './quick-repo-deletion.css';
 import delay from 'delay';
 import React from 'dom-chef';
+import cache from 'webext-storage-cache';
 import select from 'select-dom';
 import delegate from 'delegate-it';
 import elementReady from 'element-ready';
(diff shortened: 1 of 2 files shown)


source/features/quick-repo-deletion.tsx
```diff
@@ -103,6 +105,7 @@ async function start(buttonContainer: HTMLDetailsElement): Promise<void> {
 			{action: false},
 		);
 		select('.application-main')!.remove();
+		await cache.delete(getCacheKey());
```
cheap-glitch, line 108: Could this throw an error if the key doesn't exist in the cache? The user might have disabled `forked-to` for example.

cheap-glitch (commented):
> I usually try to correlate features

I think you mean "decorrelate"? :smiley: 

So if I understand correctly, the point of this is to avoid showing outdated info in `forked-to`?

kidonng:
> I think you mean "decorrelate"? :smiley: 
> 

Good catch, I missed one `not` there.

> So if I understand correctly, the point of this is to avoid showing outdated info in `forked-to`?

Right, even though the expire time is pretty short, it still looks like a glitch that should have been handled.

source/features/quick-repo-deletion.tsx
kidonng: I've tested this beforehand and it does nothing if there's no such key.

fregante (approved):

Issue #5052, opened 2021-11-06
Enable hovercards on links in markdown files


kidonng:
### Description

Inspiration: https://github.com/github/feedback/discussions/7188

GitHub _provides_ native hovercards for most types of links, however they only _enable_ some of them in a rather conservative and inconsistent way.

- Issue/PR links in markup files (README, etc.) doesn't have hovercards despite their large quantities
- Commit links like https://github.com/refined-github/refined-github/commit/4d7d591013bd51d9c93549ddc1e523d4713ae8d1 has hovercard, but if it has a hash then nope: https://github.com/refined-github/refined-github/commit/4d7d591013bd51d9c93549ddc1e523d4713ae8d1#diff-f2360271df152f491c871232f520c9cdcb89e2a93441aeb6d80a609e5d5109d5

Prior art:

- Refined GitHub already does this for [some features](https://github.com/refined-github/refined-github/blob/4d7d591013bd51d9c93549ddc1e523d4713ae8d1/source/features/show-associated-branch-prs-on-fork.tsx#L84-L91).
- OctoLinker also does this for [package source links](https://github.com/OctoLinker/OctoLinker/blob/2b2c34147fc15099f8059fec0d3953eb8ef0cfa2/packages/core/loader.js#L104-L109).

This feature request comes in two parts:

- Enable more hovercards for our features, in particular, `linkify-code`.
- Add a new feature two enable more hovercards on existing links, like in the markup files.

### Example URL

https://github.com/npm/npm

fregante:
Is this done or do you plan on adding more?

kidonng:
Part two (a new feature enabling hovercards on markup files etc.) isn't done yet

fregante:
To clarify, it sounds like you're asking to add the `data-hovercard-url` attribute to [`hasRenderedText`](https://github.com/refined-github/github-url-detection/issues/148) pages. Sounds like a 3-line feature, please send a PR so this issue can be closed 😃

PR #4619, opened 2021-07-28
Fix every instance of `unicorn/no-array-callback-reference`

darkred:
Re https://github.com/sindresorhus/refined-github/issues/4008#issuecomment-888167968

Based on the [<code>unicorn/no-array-callback-reference</code> documentation](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-array-callback-reference.md#pass) , 1st example:

```
Fail: const foo = array.map(callback);
Pass: const foo = array.map(element => callback(element));
```

## Test URLs

N/A

## Screenshot

N/A

PR #5312, opened 2022-01-15
Refactor `list-prs-for-file`

kidonng:
You are suggested to review each individual commit.

- "Improvement: use more specific path selector": the `clipboard-copy` selector is too generic
- "Fix: don't add PJAX buttons on `isEditingFile`": or else you may get duplicate editors:

  

- "Improvement: only append Text Fragment on `isEditingFile` and Chrome": only Chrome supports it now, and it doesn't work with PJAX which resets hash. Also solve https://github.com/refined-github/refined-github/pull/4739#discussion_r700517498
- Refactor commits: separate different logic on `isSingleFile` and `isEditingFile`
- "Fix: add back `.BtnGroup-item` class": this is initially broken by #4619 then some PRs from me 😅
- "Improvement: add hovercards": see #5052

## Test URLs

https://github.com/refined-github/refined-github/edit/lint/source/features/index.tsx

## Screenshot

For "Fix: add back .BtnGroup-item class":

<table>
<tr>
<tr>
	<td>Screen Shot 2022-01-16 at 01 05 11
	<td>Screen Shot 2022-01-16 at 01 39 02
</table>

### DIFF
--- a/source/features/list-prs-for-file.tsx
+++ b/source/features/list-prs-for-file.tsx
@@ -112,7 +112,7 @@ async function getCurrentPath(): Promise<string> {
 	return element!.getAttribute('value')!;
 }
 
-function getPrsForCurrentPath(): Promise<number[]> {
+async function getPrsForCurrentPath(): Promise<number[]> {
 	const [path, prsByFile] = await Promise.all([
 		getCurrentPath(),
 		getPrsByFile(),
--- a/source/features/list-prs-for-file.tsx
+++ b/source/features/list-prs-for-file.tsx
@@ -53,7 +53,6 @@ function getSingleButton(prNumber: number): HTMLElement {
 		<a
 			href={getPRUrl(prNumber)}
 			className="btn btn-sm flex-self-center"
-			data-pjax="#js-repo-pjax-container"
 		>
 			<GitPullRequestIcon className="v-align-middle"/>
 			<span className="v-align-middle"> #{prNumber}</span>
@@ -154,6 +153,7 @@ async function init(): Promise<void> {
 	const link = getSingleButton(prNumber);
 	link.classList.add('tooltipped', 'tooltipped-ne');
 	link.setAttribute('aria-label', `This file is touched by PR #${prNumber}`);
+	link.dataset.pjax = '#js-repo-pjax-container';
 	await addAfterBranchSelector(link);
 }
 
--- a/source/features/list-prs-for-file.tsx
+++ b/source/features/list-prs-for-file.tsx
@@ -104,9 +104,14 @@ const getPrsByFile = cache.function(async (): Promise<Record<string, number[]>> 
 	cacheKey: () => 'files-with-prs:' + getRepo()!.nameWithOwner,
 });
 
-async function init(): Promise<void> {
+async function getCurrentPath(): Promise<string> {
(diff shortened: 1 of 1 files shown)


source/features/list-prs-for-file.tsx
```diff
@@ -106,9 +106,33 @@ const getPrsByFile = cache.function(async (): Promise<Record<string, number[]>>
...
+	// `[aria-label="Copy path"]` on blob page, `#blob-edit-path` on edit page
+	const element = await elementReady('[aria-label="Copy path"], #blob-edit-path');
+	return element!.getAttribute('value')!;
+}
+
 async function init(): Promise<void> {
-	// `clipboard-copy` on blob page, `#blob-edit-path` on edit page
-	path = (await elementReady('clipboard-copy, #blob-edit-path'))!.getAttribute('value')!;
+	const path = await getCurrentPath();
+	const {[path]: prs} = await getPrsByFile();
```
fregante, line 117: If you decouple these you can make them in parallel

source/features/list-prs-for-file.tsx
```diff
@@ -10,10 +11,12 @@ import getDefaultBranch from '../github-helpers/get-default-branch';
...
 function getPRUrl(prNumber: number): string {
-	return buildRepoURL('pull', prNumber, 'files') + `#:~:text=${path}`;
+	return buildRepoURL('pull', prNumber, 'files');
+}
+
+function getHovercardUrl(prNumber: number): string {
+	return buildRepoURL('pull', prNumber, 'hovercard');
 }
 
 function getDropdown(prs: number[]): HTMLElement {
```
fregante, line 22: No longer a dropdown, the name should be updated

fregante (approved):

source/features/list-prs-for-file.tsx
```diff
@@ -31,17 +34,14 @@ function getDropdown(prs: number[]): HTMLElement {
 					File touched by PRs
 				</div>
 				{prs.map(prNumber => (
-					<li
```
fregante, line 34: If you drop the `li`, then it's no longer valid because there's an `ul` around it. It must be fixed either way

fregante (changes requested):

source/features/list-prs-for-file.tsx
```diff
@@ -112,13 +112,16 @@ async function getCurrentPath(): Promise<string> {
 	return element!.getAttribute('value')!;
 }
 
-async function init(): Promise<void> {
+function getPrsForCurrentPath(): Promise<number[]> {
```
yakov116, line 115: Wouldn't this also be undefined? (As per line 24)

source/features/list-prs-for-file.tsx
```diff
@@ -135,11 +138,7 @@ async function init(): Promise<void> {
 }
 
 async function initEditing(): Promise<void> {
-	const [path, prsByFile] = await Promise.all([
-		getCurrentPath(),
-		getPrsByFile(),
-	]);
-	let prs = prsByFile[path];
+	let prs = await getPrsForCurrentPath();
```
kidonng, line 141: `path` is used below so this can not be made into a standalone function.

This is not duplicating a lot of code so I think it's fine to keep it as is.

fregante (approved):

PR #5319, opened 2022-01-19
Clean up and stick sidebar footer in `infinite-scroll`

cheap-glitch:
This PR:
 - cleans up the footer in the sidebar (the notice & subscription link are still available in the original footer at the bottom of the page)
 - makes the footer sticks at the top when the feed is scrolled

## Test URLs

https://github.com

## Screenshot

<table>
<tr>
	<th>Before</th>
	<th>After</th>
</tr>
<tr>
	<td align="top">
		
	</td>
	<td valign="top">
		
	</td>
</tr>
</table>

### DIFF
--- a/source/features/infinite-scroll.css
+++ b/source/features/infinite-scroll.css
@@ -1,6 +1,6 @@
 aside[aria-label='Explore'] > .footer {
 	position: sticky !important;
-	top: 20px;
+	top: 24px;
 }
 
 aside[aria-label='Explore'] > .footer > :first-child {
--- a/source/features/infinite-scroll.css
+++ b/source/features/infinite-scroll.css
@@ -1,0 +1,8 @@
+aside[aria-label='Explore'] > .footer {
+	position: sticky !important;
+	top: 20px;
+}
+
+aside[aria-label='Explore'] > .footer > :first-child {
+	display: none !important;
+}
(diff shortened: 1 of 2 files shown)


kidonng:
Let's keep the feed link. The protip can be hidden since those three links are also available elsewhere.

cheap-glitch:
> Let's keep the feed link. The protip can be hidden since those three links are also available elsewhere.

The thing is, it's not possible to hide the protip and keep the feed link in pure CSS, because the footer contains some unwrapped text nodes.

The original link is still available at the bottom anyway, and it's not like `infinite-scroll` is actually infinite, so it's still reachable even if it takes a little more scrolling.

kidonng:
> The thing is, it's not possible to hide the protip and keep the feed link in pure CSS, because the footer contains some unwrapped text nodes.

We can just clone the feed link over.

kidonng:
Worth making the entire explore section sticky? The sidebar is still a lil empty.

If we do this we should probably put it into a standalone CSS file since it's not about `infinite-scroll` sticky speaking.

fregante (approved):

cheap-glitch:
> Worth making the entire explore section sticky?

Why not, I have no opinion about it since I hide it with my custom CSS anyway.

kidonng:
@fregante gave the thumbs-down so I thought about it again. Making the Explore section sticky would turn the dashboard into Twitter feed, of which I hide most stuff in the right side. So I already feel the pain.

PR #4152, opened 2021-03-24
Refactor `clean-conversation-headers`

kidonng:
Resolve #4145

Code readability 5x improved (possibly)

<table>

<tr>
	<td><strong>Test URL</strong>
	<td><strong>Before</strong>
	<td><strong>After</strong>

<tr>
	<td><a href="https://github.com/sindresorhus/refined-github/pull/4151">Open PR (default branch)</a>
	<td>
	<td>

<tr>
	<td><a href="https://github.com/Kenshin/simpread/pull/698">Open PR (non-default branch)</a>
	<td>
	<td>

<tr>
	<td><a href="https://github.com/sindresorhus/refined-github/pull/3402">Merged PR (same author)</a>
	<td>
	<td>

<tr>
	<td><a href="https://github.com/sindresorhus/refined-github/pull/3227">Merged PR (different author)</a>
	<td>
	<td>

<tr>
	<td><a href="https://github.com/sindresorhus/refined-github/pull/4141">Closed PR</a>
	<td>
	<td>

</table>

Extra test URL: https://github.com/parcel-bundler/parcel/pull/78 (for #3987)

image

Also updated screenshot:

image

PR #5322, opened 2022-01-19
Improve reliability of `clean-conversation-headers`

### LINKED ISSUE #5321: `clean-conversation-headers` displays incorrectly
### Please ensure:

- [X] The bug happens after clearing extension cache. The "Clear cache" button can be found at the bottom of the options page.

### Example URLs

https://github.com/peopledoc/septentrion/pull/136/files/346da19f7cd57956281751f33665c52d128de588

### Extension version

22.1.18

### Browser(s) used

Firefox

### Description

Steps:
- From the notifications page, click on a notification on a PR when the PR has been updated since the notification, leading you to a "We went looking everywhere, but couldn’t find those commits."
- From there, click on the Conversation PR tab
- The conversation header appears incorrectly

I have tried with and without a token configured.

Error in the console: 
```typescript
TypeError: select_dom(...) is undefined @ update-pr-from-base-branch.tsx:18:8
  function getBranches(): {base: string; head: string} {
	  return {
>   	  base: select('.base-ref')!.textContent!.trim(),
		  head: select('.head-ref')!.textContent!.trim(),
	  };
  }
```
Note: video also shows that even reloading the page doesn't fully fix the issue, the PR author still doesn't appear (though at the time of the video, I had not configured my token yet so maybe that part was just my token being over-used)

https://user-images.githubusercontent.com/1457576/150123039-fa1cca85-9351-4b25-b0a7-fe023b1fd688.mp4

Screenshot from 2022-01-19 12-37-03

### Include in this issue:

- [X] Console errors, if any

### LINKED ISSUE #5573: `clean-conversation-headers` is unreliable
### Please ensure:

### Example URLs

https://github.com/TauCetiStation/TauCetiClassic/pull/7772

### Extension version

22.4.1

### Browser(s) used

Yandex.Browser

### Description

~~The line under the PR title is sometimes (every second or third reload) not compressed, and the conversation filter button does not appear because of that~~

After every second or third reload the line under the PR title just becomes normal after the whole page loads (it disappears right after the "Bugs" button is added)

Should be: https://i.imgur.com/eQosD7q.png

Is: https://i.imgur.com/SVpDCCm.png

### Include in this issue:

- [X] Console errors, if any

kidonng:
Fixes #5573 

Goals for this PR:

- Replace most DOM modifications with CSS, which should make the feature idempotent... 
- ...and in turn fixing #5321, which is caused by `include` and `additionalListeners` both triggering.

## Test URLs

- You can reuse all test URLs in #4152 except for "Open PR (default branch)", which can be replaced by this PR 😉
- For #5321: https://github.com/peopledoc/septentrion/pull/136/files/346da19f7cd57956281751f33665c52d128de588

## Screenshot

Before:

image

After:

image


source/features/first-published-tag-for-merged-pr.tsx
```diff
@@ -33,15 +33,16 @@ async function init(): Promise<void> {
 		discussionHeader.classList.add('rgh-first-tag');
 
 		discussionHeader.parentElement!.append(
-			' • ',
```
yakov116, line 36: Why remove this?

source/features/clean-conversation-headers.css
```diff
@@ -5,3 +5,36 @@
...
+	content: 'by ';
+	font-size: 14px;
+}
+
+.rgh-clean-conversation-header .author {
+	margin-right: 4px;
+}
+
+.rgh-clean-conversation-headers-hide-author .author {
+	display: none !important;
```
fregante, line 27: Can we avoid important with a more specific selector or are they using important too?

source/features/clean-conversation-headers.css
```diff
@@ -5,3 +5,36 @@
...
+}
+
+.rgh-clean-conversation-header .commit-ref {
+	font-size: 11.9px;
```
fregante, line 31: Even if not identical, I'd just use 12

source/features/clean-conversation-headers.css
```diff
@@ -5,3 +5,36 @@
 .rgh-clean-conversation-headers-non-default-branch a {
 	color: var(--color-state-hover-primary-text, #fff);
 }
+
+.rgh-clean-conversation-header {
```
fregante, line 9: Like in the now-removed comment, it'd be good to know which words this will end up hiding. This applies to most of the code here where the comments existed.

source/features/clean-conversation-headers.tsx
```diff
@@ -15,49 +15,30 @@ async function initIssue(): Promise<void> {
 		return;
 	}
 
-	byline.classList.add('rgh-clean-conversation-header');
-
-	// Removes: [octocat opened this issue on 1 Jan] · 1 comments
-	for (let i = 0; i < 4; i++) {
-		byline.firstChild!.remove();
-	}
+	byline.classList.add('rgh-clean-conversation-header', 'rgh-clean-conversation-headers-hide-author');
```
fregante, line 18: Why two classes? One should be enough

fregante (commented):
Good stuff if the result is equivalent! 👌

source/features/first-published-tag-for-merged-pr.tsx
kidonng: If you look closer, you will notice GitHub uses ` • ` as a plain text separator.

In this case, the icon itself is already a strong manifestation of division. There's no need to put the dot separator here.

Also notice how GitHub arrange the entire  `byLine` as a single sentence. There's no huge gaps between. This is extremely obvious when you consider a small area of the line: _insert screenshot here_*

\*: TODO since I'm on mobile

kidonng:
> Good stuff if the result is equivalent! 👌 

Not exactly since we are comparing literal whitespaces with margin etc. But the difference should be pretty minimal.

source/features/clean-conversation-headers.tsx
kidonng: Because the `-hide-author` class is reused. It should always apply to issues but only for PRs with same author and meger.

source/features/first-published-tag-for-merged-pr.tsx
cheap-glitch: I think it's fine to break small GH conventions when it improves readability. Just my 2¢

kidonng: I had an alternative design: 8px gap in the header but stay 4px in the sticky header. How's that sound to you?

cheap-glitch: I don't think I'd even notice a 4px difference honestly :smile:

fregante (approved):

kidonng:
Wait for #5327 which changes `tags-on-commits-list` as well.

fregante:
https://github.com/refined-github/refined-github/pull/5327 was merged

javier808 (approved):
I’m a spammer

fregante:
Related PRs have been merged

source/features/conversation-activity-filter.tsx
```diff
@@ -125,10 +125,10 @@ async function addWidget(header: string, state: State): Promise<void> {
 	position.classList.add('rgh-conversation-activity-filter');
 	position.after(
 		<details
-			className={`details-reset details-overlay d-inline-block position-relative ${dropdownClass}`}
+			className={`details-reset details-overlay d-inline-block ml-2 position-relative ${dropdownClass}`}
```
cheap-glitch, line 128: Fixed the spacing of the elements added by RGH.

Before:

before

After:

after

source/features/clean-conversation-headers.tsx
```diff
@@ -83,7 +58,19 @@ void features.add(import.meta.url, {
 	additionalListeners: [
 		onConversationHeaderUpdate,
 	],
-	awaitDomReady: false,
 	deduplicate: 'has-rgh-inner',
 	init: initPR,
 });
+
+/* Test URLs
```
cheap-glitch, line 65: Added some test URLs in the feature file.

source/features/clean-conversation-headers.tsx
```diff
@@ -73,7 +49,6 @@ void features.add(import.meta.url, {
 	additionalListeners: [
 		onConversationHeaderUpdate,
 	],
-	awaitDomReady: false,
```
cheap-glitch, line 76: This makes the feature very unreliable, so I dropped it (and the calls to `elementReady` as well).

fregante: But this makes the feature slower, we don't really want to show these elements until the whole conversation has loaded

fregante (changes requested):
We should keep `awaitDomReady: false`

fregante (commented):
yolo-deadpool

kidonng:
Thanks for finishing this @cheap-glitch! However the long merge window seems to caused some confusion and reverted parts of #5416 (specifically regarding `first-published-tag-for-merged-pr`).

fregante:
I think you're specifically referring to the blue background of 

Screen Shot

kidonng:
Yup, I was supposed to submit a PR right after commenting 😅

PR #5328, opened 2022-01-21
Fix page getting stuck when leaving PR conversation

kidonng:
Fix https://github.com/refined-github/refined-github/pull/5238#issuecomment-1018551405

The problem is `clean-conversation-sidebar` doesn't get updated along with `quick-review`, resulting in looping `while` in `onReplacedElement`.

## Test URLs

From here, visit any PJAXed page

## Screenshot

N/A

### DIFF
--- a/source/features/clean-conversation-sidebar.tsx
+++ b/source/features/clean-conversation-sidebar.tsx
@@ -6,7 +6,7 @@ import * as pageDetect from 'github-url-detection';
 
 import features from '.';
 import onElementRemoval from '../helpers/on-element-removal';
-import onReplacedElement from '../helpers/on-replaced-element';
+import onDiscussionSidebarUpdate from '../github-events/on-discussion-sidebar-update';
 
 const canEditSidebar = onetime((): boolean => select.exists('.discussion-sidebar-item [data-hotkey="l"]'));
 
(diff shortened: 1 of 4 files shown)


cheap-glitch:
Wait, why didn't we notice this before? My PR didn't add the while loop, so it may have created a bug in `onElementRemoval` instead.

kidonng:
> Wait, why didn't we notice this before?

Because `onElementRemoval` is `mem`ed?

fregante (approved):

fregante:
> 6\. **`abort` listener in `onElementRemoval()` fires and resolves the promise**

Ahh this might be a problem we'll see again. Promises that were expected to resolve "when it made sense", now they resolve to `undefined`. We need to be careful when implementing the abort signal again in the future.

cheap-glitch (approved):
> Ahh this might be a problem we'll see again. Promises that were expected to resolve "when it made sense", now they resolve to `undefined`. We need to be careful when implementing the abort signal again in the future.

Yeah, I'm starting to wonder if it was really worth the extra complexity, all to avoid a console error that didn't cause any user-facing bugs in the first place. :thinking:

fregante:
I think generally it makes sense to stop any listeners when changing page, because they will pile up. We just need to keep this in mind

PR #5329, opened 2022-01-22
Disable deduplication mechanism for all CSS features

cheap-glitch:
## Test URLs

To reproduce the bug:
1. Go to https://github.com/refined-github/refined-github/issues and reload the page
2. Click on a conversation
3. The text editor at the bottom isn't cleaned

https://user-images.githubusercontent.com/46634000/150654034-492a0f79-658f-4d3d-9f45-67196e238a72.mp4

### DIFF
--- a/source/features/align-issue-labels.tsx
+++ b/source/features/align-issue-labels.tsx
@@ -3,4 +3,4 @@ import * as pageDetect from 'github-url-detection';
 
 import features from '.';
 
-void features.addCssFeature(import.meta.url, [pageDetect.isConversationList], 'has-rgh-inner');
+void features.addCssFeature(import.meta.url, [pageDetect.isConversationList]);
--- a/source/features/clean-pinned-issues.tsx
+++ b/source/features/clean-pinned-issues.tsx
@@ -3,4 +3,4 @@ import * as pageDetect from 'github-url-detection';
 
 import features from '.';
 
-void features.addCssFeature(import.meta.url, [pageDetect.isRepoIssueList], 'has-rgh-inner');
+void features.addCssFeature(import.meta.url, [pageDetect.isRepoIssueList]);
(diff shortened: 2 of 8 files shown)


fregante:
Uhm What? CSS features cannot be duplicated. If they don't work correctly, the loading logic must be fixed.

cheap-glitch:
> Uhm What? CSS features cannot be duplicated.

It's not, the problem is that all features (including CSS ones) are de-duplicated by looking at `has-rgh` by default. So what happens is:

1. The user opens the conversation list. It doesn't match `hasRichTextEditor` so the feature doesn't run.
2. The user clicks on a conversation, which triggers pjax navigation. But since `has-rgh` is still there, the feature is (wrongly) de-duplicated.

We could just set `deduplicate` to `false` by default for all CSS features. That would avoid these kinds of problems, at the cost of having a few repeated classes on `body`.

Or better we could look at the specific `rgh-` class directly?

fregante:
> We could just set deduplicate to false by default for all CSS features. 

That's the answer

> having a few repeated classes on body.

That does not happen with `classList.add()`. It's zero-cost either way. I'd argue we barely even need the pageDetect since CSS features are not supposed to cause issues on other pages. This is how non-disableable features work: *CSS is global.*

cheap-glitch:
> That's the answer

:rocket: :waning_gibbous_moon: 

> I'd argue we barely even need the pageDetect since CSS features are not supposed to cause issues on other pages.

I don't think we want the headaches of dealing with potentially leaking styles. There's only two places where we put our "really global" CSS, and imo it's best to keep it that way.

Not to mention it might force us to come up with more specific selectors, i.e. more brittle ones.

yakov116:
> > That's the answer
> 
> :rocket: :waning_gibbous_moon: 
> 
> > I'd argue we barely even need the pageDetect since CSS features are not supposed to cause issues on other pages.
> 
> I don't think we want the headaches of dealing with potentially leaking styles. There's only two places where we put our "really global" CSS, and imo it's best to keep it that way.
> 
> Not to mention it might force us to come up with more specific selectors, i.e. more brittle ones.

If I'm not mistaken that is why we added it in the first place

fregante (approved):

PR #5333, opened 2022-01-24
Fix duplicated `releases-tab` in repo header and dropdown

### LINKED ISSUE #4254: Duplicate releases in tabs and `more-dropdown`.
https://github.com/sindresorhus/refined-github/issues/new?assignees=&labels=change+request&template=change_request.md

image

Is this intended, why?

cheap-glitch:
Fixes #4254 
Also fixes the order of the dropdown items ("Releases" should be before "Security" but after "Actions")

## Test URLs

## Screenshot

https://user-images.githubusercontent.com/46634000/150748215-0de1fea8-033d-435e-83ee-a753bd7c95ed.mp4

### DIFF
--- a/source/features/releases-tab.tsx
+++ b/source/features/releases-tab.tsx
@@ -12,7 +12,7 @@ import looseParseInt from '../helpers/loose-parse-int';
 import abbreviateNumber from '../helpers/abbreviate-number';
 import {createDropdownItem} from './more-dropdown-links';
 import {buildRepoURL, getRepo} from '../github-helpers';
-import {appendBefore, highlightTab, unhighlightTab} from '../helpers/dom-utils';
+import {highlightTab, unhighlightTab} from '../helpers/dom-utils';
 
 const getCacheKey = (): string => `releases-count:${getRepo()!.nameWithOwner}`;
 
@@ -57,7 +57,7 @@ async function addReleasesTab(): Promise<false | void> {
 	// Wait for the tab bar to be loaded
 	const repoNavigationBar = (await elementReady('.UnderlineNav-body'))!;
 	const releasesTab = (
-		<li className="d-flex">
+		<li className="d-inline-flex">
 			<a
 				href={buildRepoURL('releases')}
 				className="js-selected-navigation-item UnderlineNav-item hx_underlinenav-item no-wrap js-responsive-underlinenav-item rgh-releases-tab"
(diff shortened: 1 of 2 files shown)


source/features/releases-tab.tsx
```diff
@@ -69,18 +69,15 @@ async function addReleasesTab(): Promise<false | void> {
 				<span data-content="Releases">Releases</span>
 				{count && <span className="Counter" title={count > 999 ? String(count) : ''}>{abbreviateNumber(count)}</span>}
 			</a>
-		</li>
+		</li>,
 	);
-	repoNavigationBar.append(releasesTab);
-
-	// This re-triggers the overflow listener forcing it to also hide this tab if necessary #3347
-	repoNavigationBar.replaceWith(repoNavigationBar);
```
cheap-glitch, line 77: This doesn't seem to have any effect anymore (if it ever had).

cheap-glitch (commented):
Notice a weird bug I couldn't totally get rid of: when navigating from "Code" to "Releases", the dropdown item is always shown. This doesn't happen if the Releases page is opened directly.

It seems like the tab overflow mechanism isn't triggered in this case. The best I could do was to always hide the item dropdown on creation, so at least it's not duplicated. But there's still the overflow problem:

https://user-images.githubusercontent.com/46634000/150754363-1f759607-c78f-4d68-8708-864f544f6fe4.mp4

source/features/releases-tab.tsx
cheap-glitch: I'm guessing the window `resize` event, but as for the initial page load, I haven't seen any custom events being fired so I don't really know.

What I don't get is why only the "Releases" tab is affected by this — and only when navigating with pjax.

fregante: This line was supposed to trigger the component to update the list of children to toggle on the next resize, so `resize` isn't it.

What happens is, if I remember correctly:

1. page loads 10 items in nav bar
1. GitHub detects 10 items
1. RGH adds 2 items
1. GitHub will not track these 2 new items

steps 2 and 3 might happen in the alternate order and GitHub might handle all 12 items, but it's basically a race condition. This line was supposed to avoid that.

cheap-glitch: > it's basically a race condition

Yeah, replacing the element with itself actually works… as long as it's done late enough. Since afaik there's no event we could listen to, the only fix I can think of is

```ts
await delay(50);
```

cheap-glitch: This may not be a race condition, but just a GitHub bug/crappy implementation.

The reflow of the tabs actually happens *seconds* after the page is loaded — if anything, we make it less obvious by hiding some tabs very early during page load.

Since afaik there's no event we could listen to, unless we add a three-second delay in the feature code, any DOM trick we try to pull off is going to happen way too soon and not affect anything.

Possible solution: add a timeout of ~10 seconds (maybe less), and at the end of it check if the "Release" tab is still visible and overlapping with the dropdown (cf. video above). If yes, hide it and bring the dropdown item back.

PR #5337, opened 2022-01-27
Extend `hide-low-quality-comments` with more keywords

### LINKED ISSUE #5257: Extend `hide-low-quality-comments`  with more keywords
### Description

See the attached URL. Instead of matching exact comments [like we're doing](https://github.com/refined-github/refined-github/blob/6f300c941cfa7639a6b32333105f20cf1f0ba4e4/source/helpers/is-low-quality-comment.ts), we can drop every useless word and see what's left after it: _me, too, thank, thanks, you, issue, same, any, updates, solution, following, …_ and we should drop uncommon misspelling (like `followig`)

Also we should ensure we don't hide comments that have links because `[here](##)` could still be useful.

### Screenshot

Screen Shot 6

### Example URL

https://github.com/stephencookdev/speed-measure-webpack-plugin/issues/167#issuecomment-813856455

supercoww:
Closes #5257 

## Test URLs

https://github.com/stephencookdev/speed-measure-webpack-plugin/issues/167

## Screenshot

| Before | After |
|---|---|
| image|image|
|image|image|

### DIFF
--- a/source/helpers/is-low-quality-comment.ts
+++ b/source/helpers/is-low-quality-comment.ts
@@ -1,4 +1,4 @@
 export default function isLowQualityComment(text: string): boolean {
 	// Note: the unicode range targets skin color modifiers for the hand emojis
-	return text.replace(/[\s,.!?👍👎👌🙏]+|[\u{1F3FB}-\u{1F3FF}]|[+-]\d+|⬆️|ditt?o|me|too|here|there|on|same|this|issue?s?|please|pl[sz]|any|updates?|question|solution|th[ae]nks?|thnx|you|followin?g/gui, '') === '';
+	return text.replace(/[\s,.!?👍👎👌🙏]+|[\u{1F3FB}-\u{1F3FF}]|[+-]\d+|⬆️|ditt?o|me|too|here|there|on|same|this|issue?s?|please|pl[sz]|any|updates?|bump|question|solution|th[ae]nks?|thnx|you|followin?g/gui, '') === '';
 }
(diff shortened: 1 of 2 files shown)


yakov116:
I'm -1 on "bump" and "any updates here"

fregante:
I think bumps and "any updates?" are useless when reading the conversation, but maybe we should always exclude the last comment. This is similar to the other open request for this feature. 

However "thanks" should not be considered useless, even though sometimes adding _"any updates here? Thanks"_ would exclude the comment from this filter.

source/helpers/is-low-quality-comment.ts
```diff
@@ -1,4 +1,4 @@
 export default function isLowQualityComment(text: string): boolean {
 	// Note: the unicode range targets skin color modifiers for the hand emojis
-	return text.replace(/[\s,.!?👍👎👌🙏]+|[\u{1F3FB}-\u{1F3FF}]|[+-]\d+|⬆️|ditt?o|me too|same here|same|please update|thanks?( you)?|followin?g/gui, '') === '';
+	return text.replace(/[\s,.!?👍👎👌🙏]+|[\u{1F3FB}-\u{1F3FF}]|[+-]\d+|⬆️|ditt?o|me|too|here|there|on|same|this|issue?s?|please|pl[sz]|any|updates?|bump|question|solution|th[ae]nks?|thnx|you|followin?g/gui, '') === '';
```
fregante, line 3:
```suggestion
	return text.replace(/[\s,.!?👍👎👌🙏]+|[\u{1F3FB}-\u{1F3FF}]|[+-]\d+|⬆️|ditt?o|me|too|t?here|on|same|this|issues?|please|pl[sz]|any|updates?|bump|question|solution|following/gui, '') === '';
```

fregante (approved):

cheap-glitch:
@supercoww Thanks for the PR! If you're up for it, there's also this (from https://github.com/refined-github/refined-github/issues/5257#issue-1091939044):

> Also we should ensure we don't hide comments that have links because `[here](##)` could still be useful.

This should be a new PR ofc.

supercoww:
@cheap-glitch Looks like all comments with images and links are already excluded here
https://github.com/refined-github/refined-github/blob/b3a204285619644413d407e8d7680a8a0b6ef5fd/source/features/hide-low-quality-comments.tsx#L46-L47

cheap-glitch:
> Looks like all comments with images and links are already excluded here

Actually this only checks for images _inside of_ links (probably because GitHub wraps every uploaded image in one).

PR #5339, opened 2022-01-27
Prevent `patch-diff-links` from being AJAXed

kidonng:
PJAX should never enable on these links, yet it does on PR commits page.

## Test URLs

https://github.com//refined-github/refined-github/pull/5238/commits/ebf3f1da71aea11782b8586501ec92fbb002316d

## Screenshot

N/A

### DIFF
--- a/source/features/patch-diff-links.tsx
+++ b/source/features/patch-diff-links.tsx
@@ -12,7 +12,7 @@ function init(): void {
 	}
 
 	select('.commit-meta > :last-child')!.append(
-		<span className="sha-block patch-diff-links">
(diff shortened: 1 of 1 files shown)


source/features/patch-diff-links.tsx
```diff
@@ -12,10 +12,10 @@ function init(): void {
 	}
 
 	select('.commit-meta > :last-child')!.append(
-		<span className="sha-block patch-diff-links">
```
kidonng, line 15: Forgotten in #4735

fregante (approved):

PR #5343, opened 2022-01-29
Fix missing `download-folder-button` on commit tree pages

cheap-glitch:
Fixes https://github.com/refined-github/refined-github/issues/5335

## Test URLs
https://github.com/dotnet/runtime/tree/09ff1acdad2e7789908b5db9bb89896144c13042/src/libraries/Microsoft.VisualBasic.Core

### DIFF
--- a/source/features/download-folder-button.tsx
+++ b/source/features/download-folder-button.tsx
@@ -1,20 +1,40 @@
 import React from 'dom-chef';
 import select from 'select-dom';
+import {DownloadIcon} from '@primer/octicons-react';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
 import {getRepo} from '../github-helpers';
 
+function getDropdownItem(downloadUrl: URL): JSX.Element {
+	return (
+		<a className="dropdown-item rgh-download-folder" href={downloadUrl.href}>
+			Download directory
+		</a>
+	);
+}
+
 function init(): void {
 	const downloadUrl = new URL('https://download-directory.github.io/');
 	downloadUrl.searchParams.set('url', location.href);
 
-	for (const deleteButton of select.all(`form[action^="/${getRepo()!.nameWithOwner}/tree/delete"]`)) {
-		deleteButton.before(
-			<a className="dropdown-item rgh-download-folder" href={downloadUrl.href}>
-				Download directory
+	const deleteButtons = select.all(`form[action^="/${getRepo()!.nameWithOwner}/tree/delete"]`);
+	if (deleteButtons.length === 0) { // There are no buttons to delete the folder on "commit tree" pages #5335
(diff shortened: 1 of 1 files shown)


source/features/download-folder-button.tsx
```diff
@@ -1,20 +1,40 @@
...
+	const deleteButtons = select.all(`form[action^="/${getRepo()!.nameWithOwner}/tree/delete"]`);
+	if (deleteButtons.length === 0) { // There are no buttons to delete the folder on "commit tree" pages #5335
+		select('a.btn[data-hotkey="t"]')!.after(
+			<a
+				className="btn d-none d-md-block tooltipped tooltipped-nw rgh-download-folder"
+				aria-label="Download directory"
+				href={downloadUrl.href}
+			>
+				<DownloadIcon/>
 			</a>,
```
kidonng, line 30: Why not use `getDropdownItem()`?

fregante: Different classes, different labels, different content. It wouldn't DRY it, it would just make it more complex.

source/features/download-folder-button.tsx
```diff
@@ -1,20 +1,40 @@
...
 	const downloadUrl = new URL('https://download-directory.github.io/');
 	downloadUrl.searchParams.set('url', location.href);
 
-	for (const deleteButton of select.all(`form[action^="/${getRepo()!.nameWithOwner}/tree/delete"]`)) {
-		deleteButton.before(
-			<a className="dropdown-item rgh-download-folder" href={downloadUrl.href}>
-				Download directory
+	const deleteButtons = select.all(`form[action^="/${getRepo()!.nameWithOwner}/tree/delete"]`);
+	if (deleteButtons.length === 0) { // There are no buttons to delete the folder on "commit tree" pages #5335
+		select('a.btn[data-hotkey="t"]')!.after(
```
fregante, line 23: This operation is the same in both cases. Can you always target this item instead of looking for the deletion button?

It's possible that you can use:

```
button = select( this button)
button.after(
  button is dropdown?
  <dropdown button> :
  <regular button>
)
```

cheap-glitch: > Can you always target this item instead of looking for the deletion button?

Not really, the layouts aren't exactly the same.

But how about checking the URL? The bug only happens with URLs in the form of
```
/<owner>/<repo>/tree/<commit hash>/...
```

There doesn't seem to be a `pageDetect` method for this though

fregante: Ah sorry I had misread the code. Can you move the loop in a condition instead? Currently  it's essentially [no-negated-condition](https://eslint.org/docs/rules/no-negated-condition)

PR #5344, opened 2022-01-29
Add option to set URL of toolbar icon

### LINKED ISSUE #5342: Configurable Refined-Github start page
### Description

Clicking the refined addon button opens `https://github.com/`. It would be nice, if that could be configurable to e.g. `https://github.com/notifications`.

### Screenshot

_No response_

### Example URL

cheap-glitch:
Resolves #5342 

## Test URLs
n/a

## Screenshot

image

https://user-images.githubusercontent.com/46634000/151666203-5f69a08d-8942-4b38-8af4-96f17b32baf7.mp4

### DIFF
--- a/distribution/options.html
+++ b/distribution/options.html
@@ -30,7 +30,8 @@
 
 	<p>
 		<label>
-			<strong>Action button URL</strong><br>
+			<strong>Refined GitHub link</strong><br>
+			This page will open when clicking the toolbar icon
 			<input type="url" name="actionUrl" placeholder="https://github.com" spellcheck="false" autocomplete="off" autocapitalize="off">
 		</label>
 	</p>
--- a/distribution/options.html
+++ b/distribution/options.html
@@ -30,6 +30,13 @@
 
 	<p>
 		<label>
+			<strong>Action button URL</strong><br>
+			<input type="url" name="actionUrl" placeholder="https://github.com" spellcheck="false" autocomplete="off" autocapitalize="off">
+		</label>
+	</p>
+
+	<p>
+		<label>
 			<strong>Custom CSS</strong> (useful to undo unwanted styles)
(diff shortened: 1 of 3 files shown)


kidonng:
I would prefer we just drop the click action.

That said, if we are going to add this, maybe consider implementing https://github.com/refined-github/refined-github/issues/5342#issuecomment-1024907021 with `<datalist>`.

fregante (approved):
LGTM. I wouldn't be opposed to adding the datalist with 2/3 options, but I always find its usability a bit weird for this purpose because no dropdown will be shown if you picked /issues since it acts as a filter.

distribution/options.html
```diff
@@ -28,6 +28,13 @@
 
 	<hr>
 
+	<p>
+		<label>
+			<strong>Action button URL</strong><br>
```
fregante, line 33: I think we need a better label here, "action" is a webext term. Maybe:

``` html
<strong>Refined GitHub icon link</strong>
```

cheap-glitch: How about: "Open this URL with the toolbar button" ?

kidonng: "Extension icon link" would be enough.

fregante: I think "Extension" makes it a little less clear. Florent’s suggestion could work well as an additional label, e.g.:


```html
<strong>Refined GitHub icon link</strong> this page will open when clicking the toolbar icon
```

fregante: @cheap-glitch Just pick one that makes sense to you and merge as you see fit, we don't need to spend too much time perfecting this 😁

cheap-glitch:
@fregante A little blurb for the next release :point_down: 

## _Did you know?_

Previously, the Refined GitHub icon () in the toolbar always opened the GitHub dashboard when clicked.

You can now change this to a custom page in the options :tada:

image

Issue #5341, opened 2022-01-27
`easy-toggle-commit-messages`: click anywhere to toggle commit messages


kidonng:
### Description

Basically, `easy-toggle-files` for commit messages.

When you work with repos with meaningful commit messages, find, focus and click that tiny toggle becomes a pain.

### Screenshot

<img width="519" alt="Screen Shot 2022-01-28 at 05 56 13" src="https://user-images.githubusercontent.com/44045911/151450379-504ac370-54bc-43e3-bee0-c31e63424626.png">

### Example URL

https://github.com/torvalds/linux/commits/master

PR #5345, opened 2022-01-29
Extend `toggle-everything-with-alt` to commit messages

kidonng:
Inspired by #5341

## Test URLs

- `isPRConversation`: https://github.com/refined-github/refined-github/pull//5324
- `isCommitList`: https://github.com/torvalds/linux/commits/master

### DIFF
--- a/source/features/toggle-everything-with-alt.tsx
+++ b/source/features/toggle-everything-with-alt.tsx
@@ -28,6 +28,8 @@ function collapseSelector(): string {
 	return '.js-file .js-collapse-diff';
 }
 
+const commitMessageSelector = '.TimelineItem .ellipsis-expander';
+
 function init(): void {
 	// Collapsed comments in PR conversations and files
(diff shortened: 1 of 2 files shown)


source/features/toggle-everything-with-alt.tsx
```diff
@@ -4,10 +4,6 @@ import * as pageDetect from 'github-url-detection';
 import features from '.';
 import clickAll from '../helpers/click-all';
 
-function diffsSelector(): string {
```
fregante, line 7: If you move this function inline it will no longer be deduplicated/memoized

fregante (changes requested):

fregante (approved):
LGTM. If you want to use strings in some cases, you can probably change the `clickAll` itself to also accept a string. The function is only _required_ when the selector is dynamic, which appear to be the minority.

PR #5351, opened 2022-01-29
Extend `forked-to` and `fork-source-link-same-view` to entire repo

kidonng:
Most pages of forks and parents are interchangeable, except for issues and PRs.

Enforcing too many limits is unnecessary and `useful-not-found-page` would help if something does go wrong.

## Test URLs

- `forked-to`: https://github.com/refined-github/refined-github/tags
- `fork-source-link-same-view`: https://github.com/kidonng/refined-github/tags

### DIFF
--- a/source/features/fork-source-link-same-view.tsx
+++ b/source/features/fork-source-link-same-view.tsx
@@ -22,6 +22,14 @@ async function init(): Promise<void> {
 		sameViewUrl.branch = await getDefaultBranch(forkedRepository);
 	}
 
+	if (pageDetect.isIssue() || pageDetect.isPR()) {
+		sameViewUrl.assign({
+			route: '',
+			branch: '',
+			filePath: '',
+		});
+	}
+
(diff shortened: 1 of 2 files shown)


source/features/fork-source-link-same-view.tsx
```diff
@@ -7,15 +7,30 @@ import doesFileExist from '../github-helpers/does-file-exist';
...
+
+	if (pageDetect.isIssue() || pageDetect.isPR()) {
+		sameViewUrl.assign({
+			route: '',
+			branch: '',
+			filePath: '',
+		});
+	}
+
+	if (!isFilePath() || await doesFileExist(sameViewUrl)) {
```
fregante, line 33: Doesn't this trigger an extra request for every fork page we visit? I'm surprised we went with this solution.

yakov116: We did since it's a tiny request and we did not want 404s

fregante (approved):

PR #5357, opened 2022-01-31
Improve `highest-rated-comment`

kidonng:
- Align `highest-rated-comment` style with GitHub
- Redesign `highest-rated-comment` jump to comment button (like hidden comments)

  `preview-hidden-comments` is also changed slightly to align with it

- Always add `highest-rated-comment` link unless the fist timeline item is the best comment

  Previous logic is to check if the first **3** timeline items contain the best comment.

  However chances are the first 3 timeline items could be _many_ events like references, label and assignment changes, etc. (e.g. https://github.com//ytdl-org/youtube-dl/issues/30568, previously the link to highest-rated comment isn't created).

## Changes

<table>
<tr>
	<td><strong>Test URL</strong>
	<td><strong>Before</strong>
	<td><strong>After</strong>
<tr>
	<td><a href="https://github.com//vercel/swr/issues/93"><code>highest-rated-comment</code> button</a>
	<td>
	<td>
<tr>
	<td><a href="https://github.com//vercel/swr/issues/93"><code>highest-rated-comment</code> comment</a>
	<td>
	<td>
<tr>
	<td><a href="https://github.com//refined-github/refined-github/issues/5326#event-5924166637"><code>preview-hidden-comments</code> comment</a>
	<td>
	<td>
</table>

### DIFF
--- a/source/features/highest-rated-comment.css
+++ b/source/features/highest-rated-comment.css
@@ -15,13 +15,6 @@ a.rgh-highest-rated-comment {
 	border: 2px solid var(--color-success-emphasis) !important;
 }
 
-a.rgh-highest-rated-comment .timeline-comment-header-text {
-	text-overflow: ellipsis;
-	white-space: nowrap;
-	overflow: hidden;
-	max-width: none;
-}
-
 a.rgh-highest-rated-comment .avatar {
 	position: absolute;
 	left: -58px;
--- a/source/features/highest-rated-comment.css
+++ b/source/features/highest-rated-comment.css
@@ -24,5 +24,5 @@ a.rgh-highest-rated-comment .timeline-comment-header-text {
 
 a.rgh-highest-rated-comment .avatar {
 	position: absolute;
-	left: -55px;
+	left: -58px;
 }
--- a/source/features/highest-rated-comment.css
+++ b/source/features/highest-rated-comment.css
@@ -1,9 +1,10 @@
 .rgh-highest-rated-comment.timeline-comment {
-	border: 2px solid var(--rgh-green);
+	/* Same as GitHub's `.timeline-chosen-answer` */
+	border: 2px solid var(--color-success-emphasis);
 }
 
 .rgh-highest-rated-comment.timeline-comment--caret::before {
-	background-color: var(--rgh-green);
+	background-color: var(--color-success-emphasis);
 }
 
 .rgh-highest-rated-comment.timeline-comment--caret::after {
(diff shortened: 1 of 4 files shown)


kidonng:
Alternative design using labels:

Screen Shot 2022-02-01 at 06 18 31

Screen Shot 2022-02-01 at 04 13 53

fregante:
# 💯 

Labels look good

source/features/highest-rated-comment.tsx
```diff
@@ -60,10 +60,9 @@ function highlightBestComment(bestComment: Element): void {
 }
 
 function linkBestComment(bestComment: HTMLElement): void {
-	// Find position of comment in thread
-	const position = select.all(commentSelector).indexOf(bestComment);
+	const firstTimelineItem = select('#js-timeline-progressive-loader')!.nextElementSibling!;
 	// Only link to it if it doesn't already appear at the top of the conversation
-	if (position < 3) {
+	if (firstTimelineItem === bestComment) {
```
fregante, line 65: I think this brings the limit much higher. The "position" starts at the second \*comment\* I think, while this is the first comment, right?

source/features/highest-rated-comment.css
```diff
@@ -23,5 +24,5 @@ a.rgh-highest-rated-comment .timeline-comment-header-text {
 
 a.rgh-highest-rated-comment .avatar {
 	position: absolute;
-	left: -55px;
+	left: -58px;
```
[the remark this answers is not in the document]
kidonng: - GitHub itself is using absolute position for avatars 😬
- The DOM of an real comment is complex
- And that means things break easily if GitHub changes something in between
- We don't want other feature to accidentally target the link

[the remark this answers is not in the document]
fregante: Ugh yeah their DOM is pretty bad. Just look at this alignment:

Screen Shot 4

kidonng:
Switched to labels. I kept using "Highest-rated" instead of "Best" which looks more opinionated than needed. GitHub itself uses two-word labels as well (e.g. "Public archive").

source/features/highest-rated-comment.tsx
kidonng: This is explained in the PR description.

The problem is the first comment doesn't necessarily appear at the "top of the conversation".

fregante: The logic remains, whatever counts as position (events, etc) this will effectively show a link to the second comment in a thread, potentially only 100px lower, won't it? That's not a useful link.

fregante:
Let's YOLO it if it works well for you

PR #5359, opened 2022-02-02
Align all file navigation buttons

### LINKED ISSUE #5347: `latest-tag-button` is vertically misaligned when window width < 800px
### Please ensure:

- [X] The bug happens after clearing extension cache. The "Clear cache" button can be found at the bottom of the options page.

### Example URLs

https://github.com/dotnet/runtime/tree/09ff1acdad2e7789908b5db9bb89896144c13042/src/libraries/Microsoft.VisualBasic.Core

### Extension version

22.1.18

### Browser(s) used

Firefox

### Description

image

### Include in this issue:

- [ ] Console errors, if any

fregante:
- Close #5347

## Test URLs

https://github.com/dotnet/runtime/tree/09ff1acdad2e7789908b5db9bb89896144c13042/src/libraries/Microsoft.VisualBasic.Core

## Before

Screen Shot 2

## After

Screen Shot

### DIFF
--- a/source/features/latest-tag-button.tsx
+++ b/source/features/latest-tag-button.tsx
@@ -142,7 +142,7 @@ async function init(): Promise<false | void> {
 					<DiffIcon className="v-align-middle"/>
 				</a>
 			);
-			groupButtons([link, compareLink]).classList.add('flex-self-center', 'd-flex');
+			groupButtons([link, compareLink]).classList.add('d-flex');
 		}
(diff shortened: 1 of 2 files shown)


fregante:
This works and should avoid this issue forever and for all other extensions, but ideally I'd want to see https://github.com/refined-github/refined-github/pull/4236 again because it's also avoids overflows in some cases.

PR #5360, opened 2022-02-02
Revert "Allow using videos as screenshots"

fregante:
Neither of the 2 videos work on Safari, we can't deal with codec issues. I used Gifski.app to convert one video and recreated the rest of the screenshots.

Reverts refined-github/refined-github#5271

### DIFF
--- a/build/readme-parser.ts
+++ b/build/readme-parser.ts
@@ -36,7 +36,7 @@ function extractDataFromMatch(match: RegExpMatchArray): FeatureMeta {
 	return {
 		id: simpleId as FeatureID,
 		description: parseMarkdown(linkLessMarkdownDescription),
-		screenshot: urls.find(url => /\.(png|gif|jpg|mp4|mov)$/.test(url)),
+		screenshot: urls.find(url => /\.(png|gif)$/i.test(url)),
 	};
 }
 
--- a/readme.md
+++ b/readme.md
@@ -230,7 +230,7 @@ Thanks for contributing! 🦋🙌
 
 ### Reading comments
 
-- [](# "embed-gist-inline") [Embeds linked gists.](https://user-images.githubusercontent.com/6978877/33911900-c62ee968-df8b-11e7-8685-506ffafc60b4.PNG)
+- [](# "embed-gist-inline") [Embeds short gists when linked in comments on their own lines.](https://user-images.githubusercontent.com/1402241/152117903-80d784d5-4f43-4786-bc4c-d4993aec5c79.png)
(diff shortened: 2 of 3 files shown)


build/readme-parser.ts
```diff
@@ -36,7 +36,7 @@ function extractDataFromMatch(match: RegExpMatchArray): FeatureMeta {
 	return {
 		id: simpleId as FeatureID,
 		description: parseMarkdown(linkLessMarkdownDescription),
-		screenshot: urls.find(url => /\.(png|gif|jpg|mp4|mov)$/.test(url)),
+		screenshot: urls.find(url => /\.(png|gif)$/i.test(url)),
```
cheap-glitch, line 39: How about
```suggestion
		screenshot: urls.find(url => url.endsWith('.png') || url.endsWith('.gif')),
```
?

cheap-glitch (approved):

kidonng:
> Neither of the 2 videos work on Safari, we can't deal with codec issues.

It's Safari's strict sandboxing instead of codec issues:

```
[Error] Blocked script execution in 'https://user-images.githubusercontent.com/44045911/149634307-1ed0e554-6957-4afa-8458-9c5bb4f80ebb.mp4' because the document's frame is sandboxed and the 'allow-scripts' permission is not set.
[Error] Failed to load resource: Plug-in handled load (149634307-1ed0e554-6957-4afa-8458-9c5bb4f80ebb.mp4, line 0)
[Error] Blocked script execution in 'https://user-images.githubusercontent.com/44045911/149634307-1ed0e554-6957-4afa-8458-9c5bb4f80ebb.mp4' because the document's frame is sandboxed and the 'allow-scripts' permission is not set. (x32)
```

cheap-glitch:
> To keep video support, there are some options

It just seems to me like too much hassle, for something we don't really need.

And I don't see GIFs disappearing any time soon either, without them the Internet would just collapse in on itself ;)

kidonng:
> It just seems to me like too much hassle, for something we don't really need.

With the latter option all things needed is to prefix a URL. Is that counted as "too much hassle"?

> And I don't see GIFs disappearing any time soon either, without them the Internet would just collapse in on itself ;)

This is about GIF as a format, not as a name. "GIF" you see everywhere are in fact videos and `<video>`.

fregante:
> It just seems to me like too much hassle, for something we don't really need.

build/readme-parser.ts
fregante: I think the plain regex works well here

kidonng:
> It just seems to me like too much hassle

So you are bother converting videos to GIFs instead of simply prefixing a URL to the videos?

> for something we don't really need.

So is Safari being able to directly open one or two videos, which would play just fine in the Safari's extension options and other browsers, something we _really_ need?

**This PR and what this PR is reverting doesn't even make it into the latest Safari release.**

build/readme-parser.ts
kidonng: We need to either add `jpg` back or replace the JPG screenshots.

fregante:
I really don't understand why you're spending so much energy on this non-issue.  PNG and GIF work out of the box and do not need any "solution" whatsoever. I don't know how else to put it: Zero code needed, zero bikeshedding needed.

PR #5362, opened 2022-02-02
Meta: Add test for feature screenshot

### LINKED ISSUE #5361: Ensure every feature has a compatible screenshot in the readme
The first commit in https://github.com/refined-github/refined-github/pull/5360 passes CI even though some links in the readme are not compatible with the image detection function. We should use the same function in the verifier.

Follows:

- https://github.com/refined-github/refined-github/issues/4698
- https://github.com/refined-github/refined-github/pull/5360

cc @gamemaker1  if you're interested

gamemaker1:
## Related Issues

Closes #5361 if merged

## What This PR Does

Adds a check to the `verify-features` script to ensure every feature has a valid screenshot.

### DIFF
--- a/build/verify-features.ts
+++ b/build/verify-features.ts
@@ -60,8 +60,8 @@ function findError(filename: string): string | void {
 		return `ERR: ${featureId} should be described better in the readme (at least 20 characters)`;
 	}
 
-	if (!featureMeta.screenshot) {
-		return `ERR: ${featureId} should have a screenshot in the readme`;
+	if (featureMeta.screenshot && !/\.(png|jpg|gif)$/.test(featureMeta.screenshot)) {
+		return `ERR: ${featureId} should have a screenshot (png/jpg/gif) in the readme`;
 	}
 
 	if (duplicate) {
--- a/build/verify-features.ts
+++ b/build/verify-features.ts
@@ -60,6 +60,10 @@ function findError(filename: string): string | void {
 		return `ERR: ${featureId} should be described better in the readme (at least 20 characters)`;
 	}
 
+	if (!featureMeta.screenshot) {
+		return `ERR: ${featureId} should have a screenshot in the readme`;
+	}
+
 	if (duplicate) {
 		return `ERR: ${featureId} should be described only once in the readme`;
 	}
--- a/build/verify-features.ts
+++ b/build/verify-features.ts
@@ -60,8 +60,8 @@ function findError(filename: string): string | void {
 		return `ERR: ${featureId} should be described better in the readme (at least 20 characters)`;
 	}
 
-	if (featureMeta.screenshot && !/\.(png|jpg|gif)$/.test(featureMeta.screenshot)) {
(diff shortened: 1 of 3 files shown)


gamemaker1:
I'm not sure if this is what you want, but this change means both:

1. features that do not have a screenshot, and
2. features that have a screenshot but the url does not match the regex given [here](https://github.com/gamemaker1/refined-github/blob/meta/verify-feature-screenshot/build/readme-parser.ts#L39)

will be flagged as errors

cheap-glitch:
> I'm not sure if this is what you want, but this change means both

We should just check the screenshot URL against the regex. Some features don't have an associated screenshot because they're hard to illustrate.

fregante:
Indeed, but how should we go about it? The parser returns undefined whether the URL is missing or it's not detected as an image.

gamemaker1:
I thought we could write another function that returns the raw markdown from the readme, and then check for the screenshot - but I realised we don't have a list of features that must have a screenshot - so we cannot really check if the right features have valid screenshots.

What we could do instead is create a list of not-allowed file extensions to urls and compare the urls against them - so `.mov` and `.mp4` are not allowed, and any urls with those extensions will be flagged as errors.

What do you think?

fregante:
You're right. There's effectively no difference between an unknown link and an invalid media link. In some cases (for example in hotkeys features) we do have a link, but back to the original issue/PR, not to a screenshot.

I think I found a way:

1. the parser returns the "image URL" that looks like a file, like `/[.][^.]+$/`
2. the checker ensures that `[^.]+$` is either `png` or `gif`.

gamemaker1:
@fregante that worked, thanks!

gamemaker1:
CI will fail on two features that have .mp4 files though

build/verify-features.ts
```diff
@@ -60,6 +60,10 @@ function findError(filename: string): string | void {
 		return `ERR: ${featureId} should be described better in the readme (at least 20 characters)`;
 	}
 
+	if (featureMeta.screenshot && !/\.(png|jpg|gif)$/.test(featureMeta.screenshot)) {
```
fregante, line 63: Screenshots should not be jpegs

fregante (commented):
Looks good! Just one change

build/verify-features.ts
gamemaker1: Done, but the following three features have a .jpg image:

```
ERR: comment-on-draft-pr-indicator should have a screenshot (png/gif) in the readme
ERR: link-to-github-io should have a screenshot (png/gif) in the readme
ERR: one-click-pr-or-gist should have a screenshot (png/gif) in the readme
```

yakov116: Can you download them and reload them and just change the extension before uploading it.

gamemaker1: Changed the extension, but the files are still jpegs

gamemaker1: Converted the files to pngs 🎉

fregante (approved):

fregante:
Thanks again @gamemaker1!

PR #5363, opened 2022-02-02
Exclude explicitly linked comments in `hide-low-quality-comments`

### LINKED ISSUE #5334: Exclude explicitly linked comments in `hide-low-quality-comments`
### Description

I just got a notification for a new comment in an [issue](https://togithub.com/streetcomplete/StreetComplete/issues/3667#issuecomment-1020144434) I subscribed to. The comment turned out to be just "Thanks!", so it was hidden by `hide-low-quality-comments`. Initially I was confused what I got the notification for, until I clicked the "unhide" link that revealed the new comment:

grafik

grafik

I suggest to exclude comments that were explicitly linked, i.e. those with an ID matching the one in the URL `#issuecomment-1020144434`.

### Screenshot

_No response_

### Example URL

https://togithub.com/streetcomplete/StreetComplete/issues/3667#issuecomment-1020144434

cheap-glitch:
Resolves #5334 

## Test URLs

- [All 41 useless comments should be hidden](https://github.com/stephencookdev/speed-measure-webpack-plugin/issues/167)
- [Linked comment should be visible (40 hidden comments)](https://github.com/stephencookdev/speed-measure-webpack-plugin/issues/167#issuecomment-821212185)

## Screenshots

image

---

image

### DIFF
--- a/source/features/hide-low-quality-comments.tsx
+++ b/source/features/hide-low-quality-comments.tsx
@@ -37,7 +37,11 @@ function init(): void {
 		lowQualityCount++;
 	}
 
-	for (const commentText of select.all('.comment-body > p:only-child')) {
+	const notLinkedCommentSelector = window.location.hash.startsWith('#issuecomment-')
+		? `.timeline-comment-group:not(${window.location.hash}) .comment-body > p:only-child`
+		: '.comment-body > p:only-child';
+
+	for (const commentText of select.all(notLinkedCommentSelector)) {
 		if (!isLowQualityComment(commentText.textContent!)) {
 			continue;
 		}
--- a/test/is-low-quality-comment.ts
+++ b/test/is-low-quality-comment.ts
@@ -25,4 +25,5 @@ test('isLowQualityComment', t => {
 	t.false(isLowQualityComment('Same here. <some useful information>'));
 	t.false(isLowQualityComment('Same here, please update, thanks'));
 	t.false(isLowQualityComment('Same here! Please update, thank you.'));
+	t.false(isLowQualityComment('[here](https://www.example.com)'));
 });


test/is-low-quality-comment.ts
```diff
@@ -25,4 +25,5 @@ test('isLowQualityComment', t => {
 	t.false(isLowQualityComment('Same here. <some useful information>'));
 	t.false(isLowQualityComment('Same here, please update, thanks'));
 	t.false(isLowQualityComment('Same here! Please update, thank you.'));
+	t.false(isLowQualityComment('[here](https://www.example.com)'));
```
cheap-glitch, line 28: Added a test case to cover https://github.com/refined-github/refined-github/issues/5257#issue-1091939044:

> Also we should ensure we don't hide comments that have links because `[here](##)` could still be useful.

source/features/hide-low-quality-comments.tsx
```diff
@@ -37,7 +37,11 @@ function init(): void {
 		lowQualityCount++;
 	}
 
-	for (const commentText of select.all('.comment-body > p:only-child')) {
+	const notLinkedCommentSelector = location.hash.startsWith('#issuecomment-')
+		? `.timeline-comment-group:not(${location.hash}) .comment-body > p:only-child`
+		: '.comment-body > p:only-child';
```
kidonng, line 42: This can also be written as something like:

```js
const target = document.querySelector(':target')

for (const comment of comments) {
	if (comment === target) {
		return
	}
}
```

fregante: Yeah let's go with that filter instead of making the selector extra verbose

Issue #4292, opened 2021-04-28
Meta feature: Automatically fill in browser+extension information on issue reports


fregante:
On
- https://github.com/sindresorhus/refined-github/issues/new?assignees=&labels=bug&template=1_bug_report.md
- [https://github.com/sindresorhus/refined-github/issues/new?labels=bug&template=1_bug_report.md&title=%60clean-conversation-sidebar%60%3A+Cannot+re(...)%0A%09](https://github.com/sindresorhus/refined-github/issues/new?labels=bug&template=1_bug_report.md&title=%60clean-conversation-sidebar%60%3A+Cannot+read+property+%27remove%27+of+undefined&body=%0AURL%3A+https%3A%2F%2Fgithub.com%2Fsindresorhus%2Frefined-github%2Fissues%2F4288%0A%0A%60%60%60%0ATypeError%3A+Cannot+read+property+%27remove%27+of+undefined%0ATypeError%3A+Cannot+read+property+%27remove%27+of+undefined%0A++at+clean+%28chrome-extension%3A%2F%2Fmcpkcdopehdnjeollbocpopkbdcahhej%2Fbuild%2Frefined-github.js%3A4352%3A106%29%0A++at+runFeature+%28chrome-extension%3A%2F%2Fmcpkcdopehdnjeollbocpopkbdcahhej%2Fbuild%2Frefined-github.js%3A8785%3A17%29%0A++at+setupPageLoad+%28chrome-extension%3A%2F%2Fmcpkcdopehdnjeollbocpopkbdcahhej%2Fbuild%2Frefined-github.js%3A8803%3A11%29%0A++at+chrome-extension%3A%2F%2Fmcpkcdopehdnjeollbocpopkbdcahhej%2Fbuild%2Frefined-github.js%3A8867%3A15%0A%60%60%60%0A%0A%3C%21--+Please+also+include+a+screenshot+if+the+issue+is+visible+--%3E%0A%09) (pre-filled URL)

We can automatically append this information to the text:

```
Refined GitHub: {version} {getAdditionalPermissions().hosts.length > 0 ? 'has GHE' : ''}
Browser: {navigator.userAgent}
```

fregante:
Forget about the userAgent 😅 

https://github.com/refined-github/refined-github/pull/5113#discussion_r761665067

kidonng:
> Let the user enter the browser, which may be reported as "chrome" when in reality it's Vivaldi.

[...]

PR #5364, opened 2022-02-02
Meta: Improve new issue form on RGH repo

cheap-glitch:
Follow-up of https://github.com/refined-github/refined-github/pull/5298#pullrequestreview-853544750

## Test URLs
https://github.com/refined-github/refined-github/issues/new?assignees=&labels=bug&template=1_bug_report.yml

## Screenshot

https://user-images.githubusercontent.com/46634000/152186938-cb08524e-c046-4c8a-97d7-0c5fd297432d.mp4

### DIFF
--- a/source/features/rgh-improve-new-issue-form.tsx
+++ b/source/features/rgh-improve-new-issue-form.tsx
@@ -1,0 +1,36 @@
+import React from 'dom-chef';
+import select from 'select-dom';
+import * as pageDetect from 'github-url-detection';
+
+import features from '.';
+import clearCacheHandler from '../helpers/clear-cache-handler';
+import {isRefinedGitHubRepo} from '../github-helpers';
+
+function init(): false | void {
+	if (new URL(location.href).searchParams.get('template') !== '1_bug_report.yml') {
+		return false;
+	}
+
+	const {version} = browser.runtime.getManifest();
+	select('input#issue_form_extension_version')!.value = version;
+	select('input[id*="extension_cache"]')!.parentElement!.after(
+		<button
+			className="btn"
+			type="button"
+			onClick={clearCacheHandler as unknown as React.MouseEventHandler}
+		>
+			Clear cache
+		</button>,
+	);
+}
+
+void features.add(import.meta.url, {
+	asLongAs: [
+		isRefinedGitHubRepo,
+	],
+	include: [
+		pageDetect.isNewIssue,
+	],
+	deduplicate: 'has-rgh-inner',
+	init,
+});
--- a/source/features/rgh-improve-new-issue-form.tsx
+++ b/source/features/rgh-improve-new-issue-form.tsx
@@ -6,11 +6,7 @@ import features from '.';
 import clearCacheHandler from '../helpers/clear-cache-handler';
 import {isRefinedGitHubRepo} from '../github-helpers';
 
-function init(): false | void {
-	if (new URL(location.href).searchParams.get('template') !== '1_bug_report.yml') {
-		return false;
-	}
-
+function init(): void {
 	const {version} = browser.runtime.getManifest();
 	select('input#issue_form_extension_version')!.value = version;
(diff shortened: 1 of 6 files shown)


source/features/rgh-improve-new-issue-form.tsx
```diff
@@ -0,0 +1,36 @@
...
+import clearCacheHandler from '../helpers/clear-cache-handler';
+import {isRefinedGitHubRepo} from '../github-helpers';
+
+function init(): false | void {
+	if (new URL(location.href).searchParams.get('template') !== '1_bug_report.yml') {
+		return false;
+	}
+
+	const {version} = browser.runtime.getManifest();
+	select('input#issue_form_extension_version')!.value = version;
```
kidonng, line 15: We probably want to remove this line:

https://github.com/refined-github/refined-github/blob/6c6e2799cd1499f6e629cf9764e0e23b3fd19ce5/source/features/index.tsx#L73

source/features/rgh-improve-new-issue-form.tsx
```diff
@@ -0,0 +1,36 @@
...
+	select('input[id*="extension_cache"]')!.parentElement!.after(
+		<button
+			className="btn"
+			type="button"
+			onClick={clearCacheHandler as unknown as React.MouseEventHandler}
+		>
+			Clear cache
+		</button>,
+	);
```
kidonng, line 24: Let's also reword [the template](https://github.com/refined-github/refined-github/blame/6c6e2799cd1499f6e629cf9764e0e23b3fd19ce5/.github/ISSUE_TEMPLATE/1_bug_report.yml#L17) a bit:

```diff
- The "Clear cache" button can be found at the bottom of the options page.
+ The "Clear cache" button below can also be found at the bottom of the options page.
```

source/features/rgh-improve-new-issue-form.tsx
cheap-glitch: We could keep it, in case the user disables the extension in between the time they see the error and open the link. (It's pretty unlikely though.)

source/features/rgh-improve-new-issue-form.tsx
cheap-glitch: How about dropping this line altogether?

kidonng: That could work, but will the user think the button is only available on the bug report form? 😬

kidonng:
Some more stuff to work on potentially:

- [ ] https://github.com/refined-github/refined-github/pull/5245#issuecomment-1004778498
- [ ] #4292
- [ ] [Maybe also put an actual "Find feature" button here](https://github.com/refined-github/refined-github/issues/new?assignees=&labels=change+request&template=4_change_request.yml)
  
  image

source/features/rgh-improve-new-issue-form.tsx
fregante: By the way we should bring that debugging box higher and drop the "warning" styles probably, so more people will be aware of it.

source/features/rgh-improve-new-issue-form.tsx
```diff
@@ -0,0 +1,36 @@
...
+}
+
+void features.add(import.meta.url, {
+	asLongAs: [
+		isRefinedGitHubRepo,
+	],
+	include: [
+		pageDetect.isNewIssue,
+	],
```
fregante, line 33: I think these two should be replaced by a single check for the specific template instead of adding it inside `init`

fregante (commented):
Looks good!

fregante:
@kidonng The "find feature" button requires a lot of explanation, we should probably only link "the options page" like I did on the welcome issue.

fregante (approved):

fregante (approved):

fregante (changes requested):
[]()

fregante (approved):

PR #3199, opened 2020-06-08
Use api to get latest-tag-button ahead count

yakov116:
Closes #2861
Closes #2927
Get only the last 20

Test https://github.com/sindresorhus/refined-github

Version 1
image
Version 2
image

PR #5365, opened 2022-02-02
Only disable `latest-tag-button` when actually on the latest tag

kidonng:
Bug introduced in [`e7f0d3e` (#3199)](https://github.com/refined-github/refined-github/pull/3199/commits/e7f0d3e099edc355aa0f38497cc4ea24e719365f#diff-865e9997a2004e222760004e23714c43b27194f59ec8aae2483c934935ed66b4R110), it only checks if the default branch is not ahead of the latest tag, but doesn't check if you are actually on the default branch.

Also changed the text from "release" to "tag". "Release" is a GitHub term and what we are really working with is Git tags.

## Test URLs

https://github.com/refined-github/refined-github/tree/22.1.18

## Screenshot

<table>
<tr>
	<td><strong>Before</strong>
	<td><strong>After</strong>
<tr>
	<td>
	<td>

### DIFF
--- a/source/features/latest-tag-button.tsx
+++ b/source/features/latest-tag-button.tsx
@@ -118,7 +118,7 @@ async function init(): Promise<false | void> {
 	const defaultBranch = await getDefaultBranch();
 
 	if (currentBranch === latestTag || (currentBranch === defaultBranch && aheadBy === 0)) {
-		link.setAttribute('aria-label', 'You’re on the latest tag');
+		link.setAttribute('aria-label', 'You’re on the latest version');
 		link.classList.add('disabled', 'tooltipped', 'tooltipped-ne');
 		return;
 	}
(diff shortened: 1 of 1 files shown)


fregante:
This is intentional because the feature was introduced to make you aware you’re potentially looking at a pre-release version of the documentation — hence the term “release” rather than the more specific “tag”

kidonng:
> hence the term “release” rather than the more specific “tag”.

But we should be accurate here. It's called `latest-tag-button` for a reason.

People should be aware of the relationship of tag and release and if they do they will get what the tip is conveying

fregante:
“You’re on the latest version”

fregante (approved):

fregante:
Screen Shot

Even after clearing the cache

fregante:
I don't remember our decisions around it but I think that it doesn't need to be async. @yakov116 might have some ideas on why this is happening

yakov116:
We need `getCurrentCommittish` not to depend on the DOM since we have features the depend on it before the DOM loaded and those functions cannot become async.

The issue is that `getCurrentCommittish` depends on the title. However `isRepoHome` does not have a title, so the only way is through the DOM. (Or `getDefaultBranch` since that is the branch you are on)

fregante:
Right, if this hasn't just changed behavior, I think the current situation is fine. In features that run on the repo home we just need to use getDefaultBranch. 

I also saw the mention in https://github.com/refined-github/refined-github/issues/5222#issuecomment-1047967271

PR #5378, opened 2022-02-07
Add link to upstream repo in `quick-repo-deletion`

### LINKED ISSUE #5369: Add link to upstream repo in `quick-repo-deletion`
After deleting a repo with `quick-repo-deletion`, only a header bar remains:
grafik

If the deleted repo was a fork, I often want to go to the upstream repo afterwards. So I suggest to add a link to that in the header bar:
grafik

### Example URL

https://github.com/FloEdelmann/streetcomplete-mapstyle (doesn't really make sense, since it is deleted now), forked from https://github.com/streetcomplete/streetcomplete-mapstyle

yakov116:
Closes #5369

## Test URLs
Any fork

## Screenshot
image

I need help figuring out the exact wording.

Currently I dropped the go to your other forks since I can't see why someone would want to that.

But if someone can figure out the wording. I do not like the idea of "or you can" "or you can.

### DIFF
--- a/source/features/quick-repo-deletion.tsx
+++ b/source/features/quick-repo-deletion.tsx
@@ -9,12 +9,13 @@ import * as pageDetect from 'github-url-detection';
 
 import features from '.';
 import * as api from '../github-helpers/api';
-import {getRepo} from '../github-helpers';
 import pluralize from '../helpers/pluralize';
 import addNotice from '../github-widgets/notice-bar';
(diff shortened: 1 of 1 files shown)


fregante:
Maybe it can just be a list of actions:

- Restore it. Visit source repo. List other forks.

fregante:
Ping

PR #5381, opened 2022-02-08
Add feature `pr-merge-commit-reviewers`

ForrestH:
This feature doesn't have an existing issue. It's just something that I want to use. It affects the pull request page, specifically the merge panel on that page. I tested the changes on a private GitHub enterprise instance.

## Test URL
https://github.com/ForrestH/HTMLClock/pull/1

### DIFF
--- a/source/features/pr-merge-commit-reviewers.tsx
+++ b/source/features/pr-merge-commit-reviewers.tsx
@@ -1,0 +1,46 @@
+import select from 'select-dom';
+import * as pageDetect from 'github-url-detection';
+
+import features from '.';
+import onPrMergePanelOpen from '../github-events/on-pr-merge-panel-open';
+
+function init(): void {
+	const messageField = select('textarea#merge_message_field')!;
+
+	if (messageField.value.includes('Approved by')) {
+		return;
+	}
+
+	const reviewers = select.all('span.reviewers-status-icon');
+	const message = ['Approved by:'];
+
+	for (const reviewer of reviewers) {
+		if (!reviewer.hasAttribute('aria-label')) {
+			continue;
+		}
+
+		const label = reviewer.getAttribute('aria-label');
+		if (label!.includes('approved these changes')) {
+			const name = label!.split(' ')[0];
+			message.push(name);
+		}
+	}
+
+	if (messageField.value.length > 0) {
+		messageField.value += '\n';
+	}
+
+	messageField.value += message.join(' ');
+}
+
+void features.add(import.meta.url, {
+	include: [
+		pageDetect.isPRConversation,
+	],
+	additionalListeners: [
+		onPrMergePanelOpen,
+	],
+	onlyAdditionalListeners: true,
+	deduplicate: 'has-rgh-inner',
+	init,
+});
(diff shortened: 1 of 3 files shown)


source/features/pr-merge-commit-reviewers.tsx
```diff
@@ -0,0 +1,39 @@
...
+import * as pageDetect from 'github-url-detection';
+
+import features from '.';
+import onPrMergePanelOpen from '../github-events/on-pr-merge-panel-open';
+
+function init(): void {
+	const messageField = select('textarea#merge_message_field')!;
+	const reviewers = select.all('span.reviewers-status-icon');
+
+	if (reviewers !== null && !messageField.value.includes('Approved by')) {
```
cheap-glitch, line 11: `select.all` always returns an array, no need to check for `null`. Also try to return early whenever possible instead of nesting blocks.

source/features/pr-merge-commit-reviewers.tsx
```diff
@@ -0,0 +1,39 @@
...
+		let message = 'Approved by: ';
+
+		for (const reviewer of reviewers) {
+			const attrs = reviewer.attributes;
+			for (let i = attrs.length - 1; i >= 0; i--) {
+				if (attrs[i].name === 'aria-label'
+				&& attrs[i].value.includes('approved these changes')) {
```
cheap-glitch, line 18: Just use [`getAttribute()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/getAttribute)

source/features/pr-merge-commit-reviewers.tsx
```diff
@@ -0,0 +1,39 @@
...
+					const name = attrs[i].value.split(' ')[0];
+					message += name + ' ';
```
cheap-glitch, line 20: Make it an array and use `join(' ')`

source/refined-github.ts
```diff
@@ -155,6 +155,7 @@ import './features/pr-commit-lines-changed';
 import './features/show-open-prs-of-forks';
 import './features/deep-reblame';
 import './features/clear-pr-merge-commit-message';
+import './features/pr-merge-commit-reviewers';
```
cheap-glitch, line 158: Add it to the end of the list please

cheap-glitch (changes requested):
Thanks for the PR!

> This feature doesn't have an existing issue. It's just something that I want to use.

Please open an issue next time in order to discuss things first.

It's basic repo etiquette, especially in a project with a lot of users and a high bar for features such as this one. Plus you'll avoid wasting time implementing it if it ends up being rejected.

>  I tested the changes on a private GitHub enterprise instance.

Please don't ignore the template. We need a real, working public URL to test it too.

ForrestH:
> Please open an issue next time in order to discuss things first.
Understood. Will do. 

> Please don't ignore the template. We need a real, working public URL to test it too.
I wasn't ignoring the template. I just figured any pull request that is ready for merge would suffice for testing and it seemed silly to point to a specific one, especially since the one I point to could change so that the merge panel is no longer accessible.

fregante:
Interesting workflow but it's uncommon and thus it can't be the default in Refined GitHub.

PR #5399, opened 2022-02-11
Lint `github-bugs.css`

cheap-glitch:
https://github.com/refined-github/refined-github/issues/5222#issuecomment-1027678780

## Test URLs

See review comments

## Screenshot

See review comments

### DIFF
--- a/source/features/github-bugs.css
+++ b/source/features/github-bugs.css
@@ -13,6 +13,11 @@
 	margin-left: 8px;
 }
 
+/* Avoid scroll bar on single-line `pre` in Markdown #4899 */
+.zeroclipboard-container clipboard-copy {
+	margin-bottom: 0 !important;
+}
+
 /* Bold username in conversation lists #4899 */
 :is(.js-issue-row, .js-pinned-issue-list-item) [data-hovercard-type='user'] {
 	font-weight: 600;
--- a/source/features/github-bugs.css
+++ b/source/features/github-bugs.css
@@ -1,22 +1,3 @@
-/* Place Releases/Tags selector on the same line as `Draft new release` on releases page #2325 */
-.subnav .d-flex.flex-md-row.flex-justify-between.flex-md-items-center {
-	float: left;
-}
-
-/* Align fork popup's header #3352 */
-[aria-label^='Fork'] .Box-title {
-	position: static; /* The title wrongly overlaps the Close button, making it impossible to click */
-	min-height: 0;
-	margin: -5px 0;
-}
-
-/* Add missing border radius to collapsed PR files #3444 */
-.js-blob-wrapper,
-.file:not(.open) .file-header.file-header--expandable {
-	border-bottom-left-radius: 5px;
-	border-bottom-right-radius: 5px;
-}
-
 /* Style backticked code in links #4817, matches GitHub’s `.markdown-title code` rule */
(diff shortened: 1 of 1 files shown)


source/features/github-bugs.css
```diff
@@ -1,22 +1,3 @@
-/* Place Releases/Tags selector on the same line as `Draft new release` on releases page #2325 */
-.subnav .d-flex.flex-md-row.flex-justify-between.flex-md-items-center {
-	float: left;
-}
```
cheap-glitch, line 4: Test URL: https://github.com/refined-github/refined-github/releases

Outdated selector, and the issue is gone:

image

That said, there's some real Picasso-style layout breakage going on, at around 550px wide:

image

This is caused by our "Select tag" dropdown, so it might be worth fixing.

source/features/github-bugs.css
```diff
@@ -1,22 +1,3 @@
...
-
-/* Align fork popup's header #3352 */
-[aria-label^='Fork'] .Box-title {
-	position: static; /* The title wrongly overlaps the Close button, making it impossible to click */
-	min-height: 0;
-	margin: -5px 0;
-}
```
cheap-glitch, line 11: Test URL: the "Fork to" popup

Fixed by GitHub:

image

source/features/github-bugs.css
```diff
@@ -1,22 +1,3 @@
...
-
-/* Add missing border radius to collapsed PR files #3444 */
-.js-blob-wrapper,
-.file:not(.open) .file-header.file-header--expandable {
-	border-bottom-left-radius: 5px;
-	border-bottom-right-radius: 5px;
-}
```
cheap-glitch, line 18: Test URL: https://togithub.com/refined-github/refined-github/pull/5304/files

Fixed by GitHub:

image

source/features/github-bugs.css
```diff
@@ -32,11 +13,6 @@
 	margin-left: 8px;
 }
 
-/* Avoid scroll bar on single-line `pre` in Markdown #4899 */
-.zeroclipboard-container clipboard-copy {
-	margin-bottom: 0 !important;
-}
```
cheap-glitch, line 38: Test URL: https://github.com/VSCodium/vscodium#install-with-brew-mac

I'm not entirely sure about this one. The bug doesn't appear on Firefox, but seeing as scrollbars are also a native thing, this might not be the case on other browsers.

cheap-glitch (commented):
Unsurprisingly, it's mostly the oldest fixes that can be safely dropped.

All the other bugs are still present.

source/features/github-bugs.css
fregante: This one happens more often when there's an emoji in a code block in a comment. It happens more often probably due to our own "limit code block height" feature

kidonng: > The bug doesn't appear on Firefox, but seeing as scrollbars are also a native thing, this might not be the case on other browsers.

I can still reproduce it. It does no harm so we can keep it for everyone.

> This one happens more often when there's an emoji in a code block in a comment.

I do find it is font-related. The issue appears when using the default Menlo but not JetBrains Mono.

> It happens more often probably due to our own "limit code block height" feature

It appears without Refined GitHub though.

fregante: Ah I confused this with another bug. Disregard my comment.

PR #5400, opened 2022-02-11
Hide "Unpin" button in repo header

cheap-glitch:
https://github.com/refined-github/refined-github/issues/5398#issuecomment-1036112206

## Test URLs

Any page of a repo pinned to your profile

## Screenshot

**Before**

before

**After**

after

### DIFF
--- a/source/refined-github.css
+++ b/source/refined-github.css
@@ -161,3 +161,8 @@ pr-branches
 .file-navigation.flex-items-start {
 	align-items: center !important;
 }
+
+/* Hide "Unpin" button in repo header https://github.com/refined-github/refined-github/issues/5398#issuecomment-1036112206 */
+.pagehead-actions [data-test-selector='pin-repo-button'] {
+	display: none !important;
+}


fregante (approved):

kidonng:
Note this hides the button completely, not just when it shows "Unpin".

cheap-glitch:
> Note this hides the button completely, not just when it shows "Unpin".

Can it show something else? If you click it, you get redirected to your profile page anyway.

kidonng:
> Can it show something else?

Let me rephrase a bit 🥲 _it also hides the "Pin" button_

cheap-glitch:
> > Can it show something else?
> 
> Let me rephrase a bit smiling_face_with_tear _it also hides the "Pin" button_

Aaah ok, because my pins are always full I didn't realize there was a complementary "Pin" button :facepalm: 

Dropping some CSS to revert this just in case:

```css
.pagehead-actions.pagehead-actions [data-test-selector="pin-repo-button"] {
	display: inline-block !important;
}
```

PR #5401, opened 2022-02-12
Fix `tags-dropdown` layout

cheap-glitch:
Follow-up of https://github.com/refined-github/refined-github/pull/5399#discussion_r804887489

## Test URLs

- [Releases](https://github.com/refined-github/refined-github/releases)
- [Tags](https://github.com/refined-github/refined-github/tags)

## Screenshots

**Before (~550px wide)**

before

**After**

_Desktop_

image

_Mobile_

https://user-images.githubusercontent.com/46634000/153719068-abc6e8ce-77ef-42df-a88e-683c7dbbe902.mp4

### DIFF
--- a/source/features/tags-dropdown.css
+++ b/source/features/tags-dropdown.css
@@ -10,7 +10,19 @@
 	max-width: 300px;
 }
 
-/* Space out buttons on Releases page */
-.subnav > .float-right {
-	margin-left: 20px;
+/* Limit dropdown height */
+:root .rgh-tags-dropdown .SelectMenu-list {
+	max-height: 300px !important;
+}
+
+/* Fix layout breakage on smaller screens */
+@media (min-width: 544px) and (max-width: 566px) {
+	nav[aria-label='Releases and Tags navigation buttons'] {
+		flex-basis: auto !important;
+		flex-shrink: 0 !important;
+	}
+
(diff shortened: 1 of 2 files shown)


source/features/tags-dropdown.css
```diff
@@ -10,7 +10,19 @@
 	max-width: 300px;
 }
 
-/* Space out buttons on Releases page */
-.subnav > .float-right {
-	margin-left: 20px;
```
cheap-glitch, line 15: Outdated selector.

source/features/tags-dropdown.css
```diff
@@ -10,7 +10,19 @@
...
+/* Limit dropdown height */
+:root .rgh-tags-dropdown .SelectMenu-list {
+	max-height: 300px !important;
+}
```
cheap-glitch, line 16: Avoids this:

<details>
<summary>long_tags.png</summary>

image

</details>

By the way, it looks like GitHub devs are even more confused about their own markup than we are :laughing: 

image

cheap-glitch (commented):
Forgot to mention that this PR changes the position of the dropdown, it is now right next to the searchbar.

1. It makes more sense (imo) that the search tools are close together
2. It's the only way I found to avoid breaking the layout on mobile

fregante (approved):

PR #5404, opened 2022-02-12
Disable `align-repository-header` on small screens

kidonng:
Because it doesn't need to work there.

## Test URLs

## Screenshot

<table>
<tr>
	<td><strong>Before</strong>
	<td><strong>After</strong>
<tr>
	<td>
	<td>


kidonng:
This is just a small tweak and probably not worth much reviewing 🤔

PR #5411, opened 2022-02-15
Fix `clean-repo-sidebar` hiding some readme links

cheap-glitch:
Fixes https://github.com/refined-github/refined-github/issues/5410

## Test URLs

https://github.com/panva/jose

## Screenshot

<table>
<tr>
	<th>Before
	<th>After
<tr>
	<td>

before
	<td>

after
</table>

### DIFF
--- a/source/features/clean-repo-sidebar.css
+++ b/source/features/clean-repo-sidebar.css
@@ -9,13 +9,13 @@
 }
 
 /* Hide "Readme" link made unnecessary by `toggle-files-button` #3580 */
-.rgh-clean-repo-sidebar [href$='#readme'] {
+.rgh-clean-repo-sidebar .Layout-sidebar [href$='#readme'] {
 	display: none;
 }
 
 /* Hide "Releases" header */
 @media (min-width: 768px) {
-	.rgh-clean-repo-sidebar h2 [href$='/releases'] {
+	.rgh-clean-repo-sidebar .Layout-sidebar h2 [href$='/releases'] {
 		display: none;
 	}
 }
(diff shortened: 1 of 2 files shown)


source/features/clean-repo-sidebar.css
```diff
@@ -9,13 +9,13 @@
...
 /* Hide "Readme" link made unnecessary by `toggle-files-button` #3580 */
-.rgh-clean-repo-sidebar [href$='#readme'] {
+.rgh-clean-repo-sidebar .Layout-sidebar [href$='#readme'] {
 	display: none;
 }
 
 /* Hide "Releases" header */
 @media (min-width: 768px) {
-	.rgh-clean-repo-sidebar h2 [href$='/releases'] {
+	.rgh-clean-repo-sidebar .Layout-sidebar h2 [href$='/releases'] {
```
cheap-glitch, line 18: Probably not necessary, but just in case.

yakov116 (approved):
LGTM

kidonng (approved):
😅 I didn't realize `.rgh-clean-repo-sidebar` is applied to `body` instead of only the sidebar

kidonng (changes requested):
Instead of selecting `.Layout-sidebar` manually, we should just add `.rgh-clean-repo-sidebar` to it.

That's what `sticky-sidebar` and `clean-conversation-sidebar` do and also the reason why I thought it that way

cheap-glitch:
> Instead of selecting `.Layout-sidebar` manually, we should just add `.rgh-clean-repo-sidebar` to it, not on the `body`.

I think we couldn't do it at the time because there was no simple selector for the sidebar. But now that we have `.Layout-sidebar`, I agree it makes more sense.

kidonng:
Also check `clean-repo-sidebar.tsx`:

https://github.com/refined-github/refined-github/blob/80e42b31dc33338d3ae12092d67fa8bc6f6d5161/source/features/clean-repo-sidebar.tsx#L59

fregante (approved):
Thanks for the PR and hotfix!

PR #3510, opened 2020-08-28
Faster `edit-readme`

yakov116:
#2671

Test on
https://github.com/yakov116/refined-github

PR #5413, opened 2022-02-16
Make `edit-readme` wait for DOM ready

kidonng:
Fix https://github.com/refined-github/refined-github/pull/5355#issuecomment-1041170072
Reverts #3510

So here comes a reminder: `isArchivedRepo` cannot coexist with `awaitDomReady`
It makes sense as `isArchivedRepo` is primarily dealing with features related to editing or changing stuff, which you don't hurry to do even if the DOM is not ready.

## Test URLs

### DIFF
--- a/source/features/edit-readme.tsx
+++ b/source/features/edit-readme.tsx
@@ -1,7 +1,6 @@
 import React from 'dom-chef';
 import select from 'select-dom';
 import {PencilIcon} from '@primer/octicons-react';
-import elementReady from 'element-ready';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
@@ -11,7 +10,7 @@ import isArchivedRepo from '../helpers/is-archived-repo';
 import getDefaultBranch from '../github-helpers/get-default-branch';
 
 async function init(): Promise<void | false> {
-	const readmeHeader = await elementReady('#readme :is(.Box-header, .js-sticky)');
+	const readmeHeader = select('#readme :is(.Box-header, .js-sticky)');
 
 	// The button already exists on repos you can push to
 	if (!readmeHeader || select.exists('[aria-label="Edit this file"]', readmeHeader)) {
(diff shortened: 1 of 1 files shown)


yakov116 (approved):

fregante:
This is slowing down a feature just so we avoid a console error on random archived repos. That's not an improvement. 

Some repos have really long readmes so it makes sense to preserve the fast behavior

kidonng:
> Some repos have really long readmes so it makes sense to preserve the fast behavior.

Any real world scenario where you need to hurry to edit the readme even if the DOM is not ready, on a repo you don't have write access to?

fregante:
Think of you wanting to edit the readme but having to wait for it. As I said that's more important than *not* showing an button that, _if_ pressed, leads to a 404.

kidonng:
> Think of you wanting to edit the readme but having to wait for it.

Does making it faster meets any **real needs** instead of imaginary ones?

https://github.com/refined-github/refined-github/blob/c913d5663e56175c3a4eefaf93774cb16ba46f22/source/features/edit-readme.tsx#L16

`edit-readme` is **only active** on two scenarios:

- On repos you don't have write access to. How often do you need to edit other people's README __when the DOM is not even ready__, which is done in a blink most of the time?
- On past commits of a different branch. Again, How often do you need to edit possibly outdated README __when the DOM is not even ready__, which is done in a blink most of the time?

fregante:
Now you’re just arguing against the feature itself and against faster (non-shifting) features in general. 

If it’s not useful, we drop the feature. 

If we don’t want our UI to appear faster, we stop using elementReady. 


Since:

- neither is true
- this “fix” only affects a small number of users, as you yourself are describing it

We don’t need it

fregante:
By the way it happens in a blink on fast connections, but that’s not good enough

fregante:
If you really want this change, we can have our cake and eat it too: Move the isArchived after the elementReady. Done. 

if element is missing and is archived, return false
else continue as usual with an exclamation mark.

fregante:
Interestingly, GitHub's app allowed me to merge this when it was closed 😂 

It's fine as is

kidonng:
# Thanks so much for your understanding!

I'm sincerely sorry for my previous inappropriate tone.

PR #5414, opened 2022-02-16
Enable native hovercards when linkifying issues

kidonng:
Part of #5052

## Test URLs

https://github.com//refined-github/refined-github/blob/80e42b31dc33338d3ae12092d67fa8bc6f6d5161/source/features/github-bugs.css#L6

## Screenshot

<table>
<tr>
	<td><strong>Before</strong>
	<td><strong>After</strong>
<tr>
	<td>
	<td>

### DIFF
--- a/source/github-helpers/dom-formatters.ts
+++ b/source/github-helpers/dom-formatters.ts
@@ -24,7 +24,6 @@ export function linkifyIssues(
 		baseUrl: '',
 		...options,
 		attributes: {
-			rel: 'noreferrer noopener',
 			class: linkifiedURLClass, // Necessary to avoid also shortening the links
 			...options.attributes,
 		},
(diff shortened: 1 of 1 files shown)


source/github-helpers/dom-formatters.ts
```diff
@@ -36,11 +35,13 @@ export function linkifyIssues(
...
 		const issue = link.href.split('/').pop();
-		link.setAttribute('class', 'issue-link js-issue-link tooltipped tooltipped-ne');
-		link.dataset.errorText = 'Failed to load issue title';
-		link.dataset.permissionText = 'Issue title is private';
+		link.setAttribute('class', 'issue-link js-issue-link');
+		link.dataset.errorText = 'Failed to load title';
+		link.dataset.permissionText = 'Title is private';
 		link.dataset.url = link.href;
 		link.dataset.id = `rgh-issue-${issue!}`;
+		link.dataset.hovercardType = 'issue';
```
kidonng, line 43: A small note: if `[data-hovercard-type]` is omitted (or if it also has `[data-insert-type-here-hovercards-enabled]`), it will unconditionally active.

Using `issue` is fine since it works and we can't know what type the conversation is without making a request.

source/github-helpers/dom-formatters.ts
```diff
@@ -36,11 +35,13 @@ export function linkifyIssues(
...
+		link.dataset.hovercardUrl = `${link.pathname}/hovercard`;
```
kidonng, line 44: I've said this before but: GitHub's routing is _super-duper_ smart so it works even if the path has a trailing slash.

cheap-glitch (approved):
Successfully tested on Firefox

fregante (approved):

PR #5415, opened 2022-02-16
Refactor `conversation-activity-filter`

kidonng:
TODO

### DIFF
--- a/source/features/conversation-activity-filter.tsx
+++ b/source/features/conversation-activity-filter.tsx
@@ -139,10 +139,14 @@ async function addWidget(header: string): Promise<void> {
 			</summary>
 			<details-menu
 				className="SelectMenu right-0"
-				role="menu"
 				on-details-menu-select={handleSelection}
 			>
 				<div className="SelectMenu-modal">
+					<div className="SelectMenu-header">
+						<h3 className="SelectMenu-title color-fg-default">
+							Filter Conversation Activities
+						</h3>
+					</div>
 					<div className="SelectMenu-list">
 						{createRadio('default')}
 						{createRadio('hideEvents')}
--- a/source/features/conversation-activity-filter.tsx
+++ b/source/features/conversation-activity-filter.tsx
@@ -128,11 +128,11 @@ async function addWidget(header: string): Promise<void> {
 		return;
 	}
 
-	wrap(position, <div className="d-flex flex-items-baseline"/>);
+	wrap(position, <div className="rgh-conversation-activity-filter-wrapper"/>);
 	position.classList.add('rgh-conversation-activity-filter');
 	position.after(
-		<details className={`details-reset details-overlay d-inline-block ml-2 position-relative ${dropdownClass}`}>
-			<summary aria-haspopup="true">
+		<details className={`details-reset details-overlay d-inline-block position-relative ${dropdownClass}`}>
+			<summary className="ml-2">
 				<EyeIcon className="color-text-secondary color-fg-muted"/>
 				<EyeClosedIcon className="color-icon-danger color-fg-danger"/>
 				<div className="dropdown-caret ml-1"/>
--- a/source/features/conversation-activity-filter.tsx
+++ b/source/features/conversation-activity-filter.tsx
@@ -110,14 +110,13 @@ function createRadio(filterSettings: State): JSX.Element {
 	const label = states[filterSettings];
 	return (
 		<div
-			className="select-menu-item"
+			className="SelectMenu-item"
 			role="menuitemradio"
-			tabIndex={0}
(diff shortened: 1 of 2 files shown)


source/features/conversation-activity-filter.tsx
```diff
@@ -37,38 +38,24 @@ function isWholeReviewEssentiallyResolved(review: HTMLElement): boolean {
...
 }
 
-function processPage(): void {
-	removeClassFromAll(hiddenClassName);
-
-	if (currentSetting === 'default') {
-		return;
-	}
-
+const processPage = onetime(() => {
```
cheap-glitch, line 58: I don't think we want `onetime` here, because it won't be called again when a new comment appears.

source/features/conversation-activity-filter.tsx
```diff
@@ -20,9 +21,9 @@ const states = {
 
 type State = keyof typeof states;
 
-let currentSetting: State = 'default';
```
cheap-glitch, line 23: If we're dropping the global "tracking" variable, let's rename the local instances to just `setting`. Also `applyCurrentSetting()` → `applySetting(setting)`.

kidonng: I've planned to rename the variables, in a separate commit to keep the noise out.

source/features/conversation-activity-filter.tsx
```diff
@@ -85,68 +71,73 @@ async function handleSelection({target}: Event): Promise<void> {
...
 	await delay(1);
 
-	currentSetting = select('[aria-checked="true"]', target as Element)!.dataset.value as State;
-	applyCurrentSetting();
+	const state = select('[aria-checked="true"]', target as Element)!.dataset.value as State;
+	apply(state);
 }
 
-function applyCurrentSetting(): void {
+function apply(state: State): void {
```
yakov116, line 78: JMHO this function is not descriptive enough

source/features/conversation-activity-filter.tsx
```diff
@@ -85,92 +71,104 @@ async function handleSelection({target}: Event): Promise<void> {
...
 	getRghIssueUrl(5222),
 	getRghIssueUrl(4008),
-]);
+];
 
 async function init(): Promise<void> {
-	// Reset dropdowns state #4997
-	currentSetting = 'default';
-	(await elementReady('.repository-content'))!.classList.remove('rgh-conversation-activity-is-filtered');
+	const state = minorFixesIssuePages.map(page => location.href.startsWith(page))
```
cheap-glitch, line 163:
```suggestion
	const state = minorFixesIssuePages.any(url => location.href.startsWith(url))
```

kidonng: 🙈 OMG what have I done

It was this way but then I reverted and rewrote it 🥲

kidonng:
No good 🥲 with the shortcut being added we probably have to keep the global state...

...somewhere. Maybe on the DOM.

cheap-glitch:
> Maybe on the DOM.

It's already there ;)

source/features/conversation-activity-filter.tsx
```diff
@@ -85,108 +71,119 @@ async function handleSelection({target}: Event): Promise<void> {
...
-function applyCurrentSetting(): void {
+function applyState(state: State): void {
 	// `onNewComments` registers the selectors only once
 	onNewComments(processPage);
-
 	// Actually process it right now
 	processPage();
 
-	select('.repository-content')!.classList.toggle(
+	const container = select('.js-issues-results')!;
```
yakov116, line 84:
```suggestion
	const containerClassList = select('.js-issues-results').classList!;
```

Does this work?

fregante (approved):
Let's merge this if it works as expected, including when new comments pop in. It's worth testing with that `wow` "debugging-only" classes I added back then

fregante:
Can you also add a `white-space: nowrap` on `summary` to avoid this? 

Screen Shot 11

cheap-glitch:
> Let's merge this if it works as expected, including when new comments pop in.

The feature correctly responds to newly resolved comments, but not to resolved reviews (the page needs to be reloaded).

https://user-images.githubusercontent.com/46634000/158339148-7389ac1f-e070-45f8-910b-68f601c3eb67.mp4

cheap-glitch:
> comments that are loaded by clicking "Load more"

Yes those are hidden properly. Actually, I was thinking of extending the feature to automatically click "Load more" buttons when a filter is activated, since they end taking more space than they're saving when every resolved comment is hidden.

kidonng:
> Can you also add a `white-space: nowrap` on `summary` to avoid this on small windows?

I already fixed that, no need for `white-space: nowrap`.

PR #5327, opened 2022-01-21
Improve `tags-on-commits-list` and `first-published-tag-for-merged-pr` style

kidonng:
Adopt to GitHub's commit ref style, as seen on PRs headers.

## Test URLs

- `tags-on-commits-list`: https://github.com/refined-github/refined-github/commits/22.1.18
- `first-published-tag-for-merged-pr`: https://togithub.com/refined-github/refined-github/pull/5305

## Screenshot

<table>
<tr>
	<td align="center"><strong>Feature</strong>
	<td align="center"><strong>Before</strong>
	<td align="center"><strong>After</strong>
<tr>
	<td><code>tags-on-commits-list</code>
	<td>
	<td>
<tr>
	<td><code>first-published-tag-for-merged-pr</code>
	<td>
	<td>
</table>

PR #5416, opened 2022-02-16
Restyle two tag-related features

kidonng:
First attempt: #5327

I made a mistake last time: the style is referenced from branch names, but tags are not actually branch names. This makes the blue background of branch names (in contrast to the usual gray of `<code>` blocks) less distinctive. The color also makes it appear more prominent than necessary, especially for `tags-on-commits-list`.

So this time I went to take a look at how [release pages](https://github.com/refined-github/refined-github/releases/tag/22.2.13) show the tag, which also shows meta info just like the feature does.

## Screenshot

<table>
<tr>
	<td align="center"><strong>Test URL</strong>
<tr>
	<td><a href="https://github.com/refined-github/refined-github/commits/22.1.18"><code>tags-on-commits-list</code></a>
	<td>
	<td>
<tr>
	<td><a href="https://github.com/preactjs/next-plugin-preact/commits/cf378c33e929efb26b8736c115cff0116e31193d">Multiple <code>tags-on-commits-list</code></a>
	<td>
	<td>
<tr>
	<td><a href="https://github.com//refined-github/refined-github/pull/5305"><code>first-published-tag-for-merged-pr</code></a>
	<td>
	<td>
</table>

## Alternative design

Wrapping the version in `<code>`, which keeps it easily recognizable while being less prominent.

image

### DIFF
--- a/source/features/first-published-tag-for-merged-pr.tsx
+++ b/source/features/first-published-tag-for-merged-pr.tsx
@@ -34,16 +34,13 @@ async function init(): Promise<void> {
 
 		discussionHeader.parentElement!.append(
 			' • ',
-			<TagIcon className="mx-1 color-text-secondary color-fg-muted"/>,
-			<span className="commit-ref">
-				<a
-					href={buildRepoURL('releases/tag', tagName)}
-					className="no-underline"
-					title={`${tagName} was the first Git tag to include this PR`}
-				>
-					{tagName}
-				</a>
-			</span>,
+			<a
+				href={buildRepoURL('releases/tag', tagName)}
+				className="Link--muted"
+				title={`${tagName} was the first Git tag to include this PR`}
+			>
+				<TagIcon/> {tagName}
+			</a>,
 		);
 	}
 }
--- a/source/features/first-published-tag-for-merged-pr.tsx
+++ b/source/features/first-published-tag-for-merged-pr.tsx
@@ -34,12 +34,14 @@ async function init(): Promise<void> {
 
 		discussionHeader.parentElement!.append(
 			' • ',
+			<TagIcon/>,
+			' ',
 			<a
 				href={buildRepoURL('releases/tag', tagName)}
 				className="Link--muted markdown-title"
 				title={`${tagName} was the first Git tag to include this PR`}
 			>
-				<TagIcon/> <code>{tagName}</code>
+				<code>{tagName}</code>
 			</a>,
 		);
 	}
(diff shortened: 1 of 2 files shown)


cheap-glitch:
I'd rather have a background under the tag name, it's usually short so it doesn't naturally stand out otherwise. I'm okay with either blue or gray.

kidonng:
> I'd rather have a background under the tag name, it's usually short so it doesn't naturally stand out otherwise.

That's also one of the reason I had in the first attempt. However I think `first-published-tag-for-merged-pr` doesn't benefit that much from the background:

- (As said above,) it makes branch names next to it less distinctive
- With `clean-conversation-headers` there's really not much stuff there, so it can be easily spotted anyway

fregante:
It kinda makes sense but really I think this piece of information is important enough to deserve visibility. For example in PRs it lets me know whether a PR has been released and in the commit list it acts as an instant "release notes" between versions.

Making it less visible I think lessens the impact of this feature, so its 👎 for me

kidonng:
> Making it less visible I think lessens the impact of this feature, so its 👎 for me

Alternative design then 😅

kidonng:
> the style is referenced from branch names, but tags are not actually branch names. This makes the blue background of branch names (in contrast to the usual gray of `<code>` blocks) less distinctive.

fregante:
> In short: the blue background is misleading. GitHub only uses that for branches, so should we.

I don’t think the color itself has any meaning for the user since this style only seems to appear in PRs headers, so claiming that it misleads the user is like saying all red cars look like Ferraris. 

If you want to make a change to how `first-published-tag-for-merged-pr` looks I'd go in the opposite direction and make it even more noticeable somehow.

My arguments don't apply to `tags-on-commits-list` though so that part can be merged.

kidonng:
> I don’t think the color itself has any meaning for the user since this style only seems to appear in PRs headers

That's not true, it is literally also in the timeline:

chrome_screenshot_1646379046423

And where `linkify-branch-references` apply:





These are just a few examples, I'm pretty sure there's more.

> so claiming that it misleads the user is like saying all red cars look like Ferraris.

That style is from `.commit-ref` and it's called "commit ref" for a reason.

cheap-glitch:
I think it's fine to make small deviations from GitHub conventions when it makes sense. I agree with fregante that the highlight is part of the feature, it's small and when you're actively looking for it the color makes a difference — especially on cramped hi-def screens.

fregante:
Let's try

PR #5418, opened 2022-02-16
Refactor `quick-repo-deletion`

kidonng:
- Adjust the button style to be more consistent with others
- Rewrite animation with pure CSS since it's cooler and no more need to handle exceptions

  It looks the almost the same except for some minor text changes

- Previously the catch-all click listener will prevent your first click after deletion, now the listener will be removed properly
- Also add a listener to abort deletion on "any" keypress (actually doesn't work with many keys like `Esc`, `S` or `Space` due to other listeners)
- Properly deinit

  For example, pressing `t` to search files while the deletion countdown is activate, then go back, the button is no longer in a weird state

## Test URLs

Visit your forks at `https://github.com/<username>?tab=repositories&q=&type=fork`

## Screenshot

<table>
<tr>
	<td><strong>Before</strong>
	<td><strong>After</strong>
<tr>
	<td>
	<td>
</table>

https://user-images.githubusercontent.com/44045911/154350458-aa729b29-4ff9-49d5-afbb-5c4373eafe16.mov

### DIFF
--- a/source/features/quick-repo-deletion.css
+++ b/source/features/quick-repo-deletion.css
@@ -9,3 +9,68 @@
 	opacity: 30%;
 	z-index: 91; /* Higher than sticky readme header and `sticky-sidebar` #4192 */
 }
+
+.rgh-quick-repo-deletion:not([open]) .btn::after {
+	content: 'Delete';
+}
+
+.rgh-quick-repo-deletion[open] .btn {
+	animation: 6s step-end forwards rgh-quick-repo-deletion-scale;
+}
+
+.rgh-quick-repo-deletion[open] .btn::after {
+	animation: 6s step-end forwards rgh-quick-repo-deletion-countdown;
+	content: '';
+}
+
+@keyframes rgh-quick-repo-deletion-scale {
+	0% {
+		transform: scale(1.2);
+	}
+
+	20% {
+		transform: scale(1.5);
+	}
+
+	40% {
+		transform: scale(1.8);
+	}
+
+	60% {
+		transform: scale(2.1);
+	}
+
+	80% {
+		transform: scale(2.4);
+	}
+
+	100% {
+		transform: scale(2.4);
+	}
+}
+
+@keyframes rgh-quick-repo-deletion-countdown {
+	0% {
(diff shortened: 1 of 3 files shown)


yakov116:
Once your at it do https://github.com/refined-github/refined-github/pull/5378

source/features/quick-repo-deletion.tsx
```diff
@@ -23,17 +23,12 @@ function handleToggle(event: delegate.Event<Event, HTMLDetailsElement>): void {
 		'.rgh-open-prs-of-forks', // PRs opened in the source repo
 	]);
 
-	if (hasContent && !confirm('This repo has open issues/PRs, are you sure you want to delete everything?')) {
+	if (hasContent && !confirm('This repository has open issues or pull requests, are you absolutely sure?')) {
```
fregante, line 26: Not an upgrade, I want to preserve "delete everything"

source/features/quick-repo-deletion.css
```diff
@@ -9,3 +9,68 @@
...
+.rgh-quick-repo-deletion[open] .btn {
+	animation: 6s step-end forwards rgh-quick-repo-deletion-scale;
+}
+
+.rgh-quick-repo-deletion[open] .btn::after {
+	animation: 6s step-end forwards rgh-quick-repo-deletion-countdown;
+	content: '';
+}
+
+@keyframes rgh-quick-repo-deletion-scale {
```
fregante, line 26: Not an upgrade. More code and less guarantee that the CSS will sync up (hint: it won't). This CSS file must be restored in its entirety

fregante (changes requested):

source/features/quick-repo-deletion.css
kidonng: Just an experiment 😅 was planning to revert to using JS

source/features/quick-repo-deletion.tsx
kidonng: This was taken from GitHub's popup title

fregante:
Feb 17 -> May 3

PR #5420, opened 2022-02-17
Meta: Change `web-ext` start page to RGH organization dashboard

cheap-glitch:
I think it gives a better overview of the current state of the repo. Thumb down if you don't want to change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -125,7 +125,7 @@
 			"firefoxProfile": "./test/web-ext-profile",
 			"chromiumProfile": "./test/web-ext-profile",
 			"startUrl": [
-				"https://github.com/refined-github/refined-github/issues"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -125,7 +125,7 @@
 			"firefoxProfile": "./test/web-ext-profile",
 			"chromiumProfile": "./test/web-ext-profile",
 			"startUrl": [
-				"https://github.com/refined-github/refined-github/issues"
+				"https://github.com/orgs/refined-github/dashboard"
```
jdreesen, line 128: This redirects to `https://github.com/` for me...

cheap-glitch: Right, forgot this only works for people in org, my bad :sweat_smile:

cheap-glitch:
This can be done with a command-line alias

PR #5422, opened 2022-02-17
Meta: Correctly deinit all features

cheap-glitch:
https://github.com/refined-github/refined-github/issues/5222#issuecomment-996625242

This PR:
 * adds the global `Deinit` type and use it in every relevant `init` return type
 * makes sure that every feature is correctly "deinitified"
 * deinitifies the `additionalListeners` too
 * passes an `AbortSignal` to init functions and uses it in as many `addEventListener` as possible
 * drops every `onetime(init)` where it makes sense


fregante:
😱 
3332f6ef40a0ae9dd34d6d5eaa7e7524656a1f77

source/features/index.tsx
```diff
@@ -148,23 +148,48 @@ const globalReady: Promise<RGHOptions> = new Promise(async resolve => {
 	resolve(options);
 });
 
+function setupDeinit(deinit: Deinit): void {
```
fregante, line 151: Can we change this function to this since `addEventListener` is repeated 4 times here?

```suggestion
function setupDeinit(deinit: Deinit): void {
	document.addEventListener('pjax:start', getDeinitHandler(deinit), {once: true});
}
```

source/features/rgh-sponsor-button.tsx
```diff
@@ -82,19 +82,20 @@ async function suchLove({delegateTarget}: delegate.Event): Promise<void> {
 	love.remove(); // 💔
 }
 
-async function handleNewIssue(): Promise<false> {
+async function handleNewIssue(signal: AbortSignal): Promise<false> {
 	if (getRepo()!.owner !== getUsername() && !await cache.get('did-it-wiggle')) {
 		select('.btn-primary[href$="/issues/new/choose"], .btn-primary[href$="/issues/new"]')
 			?.addEventListener('mouseenter', wiggleWiggleWiggle, {
 				once: true,
+				signal,
```
fregante, line 90: Oops, incompatible with Safari.

I think it won't be a huge issue, it's just that Safari users won't get this enhancement in this case. We just need to determine whether this causes duplicate calls a case-by-case basis.

However I think this shouldn't be covered anyway for the same reason highlighted in my first review: single elements are removed on unload regardless (unless they're outside the ajaxed container)

So let's drop this single line as well and everything like it (if I missed any)

source/features/selection-in-new-tab.tsx
```diff
@@ -16,14 +15,14 @@ function openInNewTab({key, target}: KeyboardEvent): void {
...
+	document.addEventListener('keypress', openInNewTab, {signal});
 }
 
 void features.add(import.meta.url, {
 	shortcuts: {
 		'shift o': 'Open selection in new tab',
 	},
 	awaitDomReady: false,
-	init: onetime(init),
+	init,
```
fregante, line 27: This however is one of those situations where listeners will just pile up in Safari, and that's a big bug. :(

Potential solution: our own addEventListener wrapper:

```JavaScript 
function onEvent(element, name, signal, handler, options) {
	element.addEventListener(name, handler, options)
	signal.addEventListener('abort', () => {	
		element.removeEventListener(name, handler, options)
	})
}
```
😬

source/features/sticky-sidebar.tsx
```diff
@@ -17,22 +17,21 @@ function updateStickiness(): void {
...
 
-	return () => {
-		selectObserver.abort();
-		resizeObserver.disconnect();
-		window.removeEventListener('resize', onResize);
-	};
+	return [
+		selectObserver.abort,
+		resizeObserver.disconnect,
+	];
```
fregante, line 34: Right
```suggestion
	return [
		selectObserver,
		resizeObserver,
	];
```

source/features/sync-pr-commit-title.tsx
```diff
@@ -83,14 +83,17 @@ function disableSubmission(): void {
...
 
-let listeners: delegate.Subscription[];
-function init(): void {
-	listeners = [
-		onPrMergePanelOpen(updateCommitTitle),
+const listeners: delegate.Subscription[] = [];
+
+function init(signal: AbortSignal): Deinit[] {
+	listeners.push(
+		onPrMergePanelOpen(updateCommitTitle, signal),
```
fregante, line 90: Either we pass the `signal` or we return a deinit function, not both.

source/github-events/on-new-comments.ts
```diff
@@ -20,23 +20,13 @@ function paginationSubmitHandler({delegateTarget: form}: delegate.Event): void {
...
-	delegates.clear();
-	handlers.clear();
-	observer.disconnect();
-}
-
-function getFragmentLoadHandler(callback: EventListener): delegate.EventHandler {
+function getFragmentLoadHandler(callback: EventListener, signal: AbortSignal): delegate.EventHandler {
 	return ({delegateTarget}) => {
-		delegateTarget.addEventListener('load', callback);
+		delegateTarget.addEventListener('load', callback, {signal});
```
fregante, line 25: As before: Because this is on a single, removable element, this isn't necessary

source/github-events/on-new-comments.ts
```diff
@@ -57,10 +44,25 @@ function addListeners(): void {
...
-	addListeners();
+export default function onNewComments(callback: VoidFunction, signal: AbortSignal): void {
+	if (signal.aborted) {
+		return;
+	}
+
+	addListeners(signal);
 	handlers.add(callback);
+	signal.addEventListener('abort', removeListeners, {once: true});
 }
```
fregante, line 68: This file is awkward. Can it just return a local `delegates` array instead of calling it in its own `removeListeners`

The array would probably also have to include a custom `() => handlers.clear()`

fregante (changes requested):
oof, big PR.

Some changes are required for Safari compatibility. If you feel that they're too much, I think maybe those specific features can be excluded from this PR and maybe addressed separately

source/features/batch-mark-files-as-viewed.tsx
```diff
@@ -62,15 +62,16 @@ function onAltClick(event: delegate.Event<MouseEvent, HTMLInputElement>): void {
...
-	};
+function init(): Deinit[] {
+	return [
+		// `mousedown` required to avoid mouse selection on shift-click
+		delegate(document, '.js-reviewed-toggle', 'mousedown', batchToggle),
+		delegate(document, '.js-toggle-user-reviewed-file-form', 'submit', remember),
+		delegate(document, '.js-reviewed-toggle', 'click', onAltClick),
+		() => {
+			previousFile = undefined;
+		},
```
kidonng, line 73: Unrelated to this PR but it would be better if we don't need to deal with this global state. Maybe put it in some container's `[data-previous-file]` attribute and let PJAX destroy the container for us.

source/features/copy-on-y.tsx
```diff
@@ -11,12 +11,8 @@ const handler = ({key, target}: KeyboardEvent): void => {
...
 };
 
-function init(): VoidFunction {
-	window.addEventListener('keyup', handler);
-
-	return () => {
-		window.removeEventListener('keyup', handler);
-	};
+function init(signal: AbortSignal): void {
+	window.addEventListener('keyup', handler, {signal});
```
kidonng, line 15: Should this be using `delegate`?

source/features/esc-to-deselect-line.tsx
```diff
@@ -19,8 +18,8 @@ function listener({key, target}: KeyboardEvent): void {
 	}
 }
 
-function init(): void {
-	document.body.addEventListener('keyup', listener);
+function init(signal: AbortSignal): void {
+	document.body.addEventListener('keyup', listener, {signal});
```
kidonng, line 22: Maybe `delegate`

source/features/sync-pr-commit-title.tsx
fregante: Oh, also this needs to be changed from `listeners.push(` to `return [` since this feature has a custom `deinit` function we no longer need after this change.

source/features/batch-mark-files-as-viewed.tsx
fregante: I don't see a problem in this specific instance, the state is correctly erased on page change.

source/features/copy-on-y.tsx
fregante: `delegate` is `addEventListener` + `closest`. Here we're only using an `isEditable` check to _exclude_ specific elements, which doesn't work the same as as `delegate`

source/features/index.tsx
```diff
@@ -148,23 +148,48 @@ const globalReady: Promise<RGHOptions> = new Promise(async resolve => {
...
+		});
+		return;
+	}
+
+	if ('abort' in deinit) { // Selector observer
+		document.addEventListener('pjax:start', deinit.abort, {once: true});
+		return;
+	}
+
+	if ('destroy' in deinit) { // Delegate subscription
```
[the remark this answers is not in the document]
cheap-glitch: Sadly those aren't classes but plain objects:

https://github.com/fregante/delegate-it/blob/main/index.ts#L7-L9
```ts
	export type Subscription = {
		destroy: VoidFunction;
	};
```

https://github.com/josh/selector-observer/blob/b811bb5d13cba9940026430f92a6760d5f9e365c/lib/index.d.ts#L40-L42
```ts
type Observer = {
  abort(): void
}
```

source/features/conversation-activity-filter.tsx
```diff
@@ -140,7 +140,7 @@ async function addWidget(header: string): Promise<void> {
 			<details-menu
 				className="SelectMenu right-0"
 				role="menu"
-				on-details-menu-select={handleSelection}
```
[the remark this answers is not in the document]
cheap-glitch: Right, didn't realize listeners are removed along with the element they've been attached to :facepalm:

source/features/selection-in-new-tab.tsx
cheap-glitch: > This however is one of those situations where listeners will just pile up in Safari

Aren't listeners natively deduplicated if they're used with named functions? Unless Safari doesn't do it

source/features/sync-pr-commit-title.tsx
cheap-glitch: > Either we pass the `signal` or we return a deinit function, not both.

Actually, I'm thinking of dropping the global `AbortSignal` entirely. Having two deinit mechanisms just seems redundant, not to mention Safari doesn't support it in `addEventListener` which was the main reason it was suggested in the first place.

We can keep using `AbortController` in specific features, when it really adds value. Otherwise let's just return `Deinit` everywhere.

source/features/batch-mark-files-as-viewed.tsx
kidonng: It's not a _problem_, I just think [less global state](https://github.com/refined-github/refined-github/pull/5415/commits/5c773a941dbc16eb8f3185fafb3b16709706c013) is better generally. By requiring passing local state you avoid mistakenly relying on the global state when it doesn't actually reflect the current situation.

source/features/conversation-activity-filter.tsx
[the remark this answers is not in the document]
fregante: Technically they're still attached, but the user can't select them. If we're lucky they're then garbage-collected.

source/features/selection-in-new-tab.tsx
fregante: Yes but:

- we’re  passing a new signal every time, which should break the memoization
- I wonder if Safari ignores the signal in the memoization since it doesn’t support it
- This is just one situation, let’s see if there are other `onetime` usages that use other observers.

source/features/sync-pr-commit-title.tsx
fregante: Ideally all of our listeners use the signal so we never have to return anything.

For example if you drop the signal on event handlers you always have to return a custom `() => el.removeEventListener(repeat the arguments)` whereas with a signal ideally you only have to add `, {signal}`

cheap-glitch: True, but I don't think we have much of choice until Safari adds support for `{signal}`.

Keeping the old pattern of `onetime(init)` + global event listeners is probably the simplest for now.

fregante: > Keeping the old pattern of onetime(init) + global event listeners is probably the simplest for now.

Sure, in cases where we only use `addEventListener` that's ok. Creating the handler I suggested is not too difficult though.

source/features/batch-mark-files-as-viewed.tsx
fregante: If there's an alternative solution, sure, but if it doubles or triples the amount of code I'd rather not

source/features/improve-shortcut-help.tsx
```diff
@@ -48,11 +47,18 @@ function observeShortcutModal({key, target}: KeyboardEvent): void {
...
+		() => {
+			document.removeEventListener('keypress', observeShortcutModal);
+		},
+	];
 }
 
 void features.add(import.meta.url, {
 	awaitDomReady: false,
-	init: onetime(init),
+	init,
```
cheap-glitch, line 63: I wonder if it's worth de-initifying features that act on every page and only register a global event listener.

Maybe in this case `onetime(init)` is better, since it won't mean destroying and recreating the listener on every pjax navigation.

source/features/reactions-avatars.tsx
```diff
@@ -56,7 +54,26 @@ function getParticipants(button: HTMLButtonElement): Participant[] {
...
+	// Start loading a little before they become visible
+	rootMargin: '500px',
+});
+
+const resizeObserver = new ResizeObserver(([{target}], observer) => {
+	if (!target.isConnected) {
+		observer.unobserve(target);
+		observeReactions();
+	}
+});
```
cheap-glitch, line 74: Replaced the usage of `onElementRemoval` on every reaction element, with a global `ResizeObserver`.

This makes the feature much easier to deinit, and should make it lighter too.

cheap-glitch (commented):
Current state of this PR:

- dropped the global signal in every `addEventListener` call, and instead made the `init` return a wrapper(s) around the `removeEventListener`(s). This concerns the following features:
    - [`copy-on-y`](https://github.com/refined-github/refined-github/blob/aa3ee7eabac8fbf4f92663c0c2b1736c602732ff/source/features/copy-on-y.tsx)
    - [`esc-to-deselect-line`](https://github.com/refined-github/refined-github/blob/aa3ee7eabac8fbf4f92663c0c2b1736c602732ff/source/features/esc-to-deselect-line.tsx)
    - [`file-finder-buffer`](https://github.com/refined-github/refined-github/blob/aa3ee7eabac8fbf4f92663c0c2b1736c602732ff/source/features/file-finder-buffer.tsx)
    - [`hide-navigation-hover-highlight`](https://github.com/refined-github/refined-github/blob/aa3ee7eabac8fbf4f92663c0c2b1736c602732ff/source/features/hide-navigation-hover-highlight.tsx)
    - [`improve-shortcut-help`](https://github.com/refined-github/refined-github/blob/aa3ee7eabac8fbf4f92663c0c2b1736c602732ff/source/features/improve-shortcut-help.tsx)
    - [`keyboard-navigation`](https://github.com/refined-github/refined-github/blob/aa3ee7eabac8fbf4f92663c0c2b1736c602732ff/source/features/keyboard-navigation.tsx)
    - [`prevent-pr-merge-panel-opening`](https://github.com/refined-github/refined-github/blob/aa3ee7eabac8fbf4f92663c0c2b1736c602732ff/source/features/prevent-pr-merge-panel-opening.tsx)
    - [`selection-in-new-tab`](https://github.com/refined-github/refined-github/blob/aa3ee7eabac8fbf4f92663c0c2b1736c602732ff/source/features/selection-in-new-tab.tsx)
    - [`sticky-sidebar`](https://github.com/refined-github/refined-github/blob/aa3ee7eabac8fbf4f92663c0c2b1736c602732ff/source/features/sticky-sidebar.tsx)
    - [`stop-pjax-loading-with-esc`](https://github.com/refined-github/refined-github/blob/aa3ee7eabac8fbf4f92663c0c2b1736c602732ff/source/features/stop-pjax-loading-with-esc.tsx)
    - [`unfinished-comments`](https://github.com/refined-github/refined-github/blob/aa3ee7eabac8fbf4f92663c0c2b1736c602732ff/source/features/unfinished-comments.tsx)
    - [`wait-for-attachments`](https://github.com/refined-github/refined-github/blob/aa3ee7eabac8fbf4f92663c0c2b1736c602732ff/source/features/wait-for-attachments.tsx)
    - [`wait-for-checks`](https://github.com/refined-github/refined-github/blob/aa3ee7eabac8fbf4f92663c0c2b1736c602732ff/source/features/wait-for-checks.tsx)
    - plus [`on-pr-merge-panel-open`](https://github.com/refined-github/refined-github/blob/aa3ee7eabac8fbf4f92663c0c2b1736c602732ff/source/features/on-pr-merge-panel-open.tsx)

- kept the `AbortSignal` in the `init` of those three features, to use with `onElementRemoval` and `onReplacedElement`:
   - [`clean-conversation-sidebar`](https://github.com/refined-github/refined-github/blob/aa3ee7eabac8fbf4f92663c0c2b1736c602732ff/source/features/clean-conversation-sidebar.tsx)
  - [`comment-on-draft-pr-indicator`](https://github.com/refined-github/refined-github/blob/aa3ee7eabac8fbf4f92663c0c2b1736c602732ff/source/features/comment-on-draft-pr-indicator.tsx)
  - [`select-all-notifications-shortcut`](https://github.com/refined-github/refined-github/blob/aa3ee7eabac8fbf4f92663c0c2b1736c602732ff/source/features/select-all-notifications-shortcut.tsx)

- used `setupDeinit` directly in those two features, because I couldn't find a better way to deinit in nested callbacks:
    - [`pr-branch-auto-delete`](https://github.com/refined-github/refined-github/blob/aa3ee7eabac8fbf4f92663c0c2b1736c602732ff/source/features/pr-branch-auto-delete.tsx)
    - [`wait-for-checks`](https://github.com/refined-github/refined-github/blob/aa3ee7eabac8fbf4f92663c0c2b1736c602732ff/source/features/wait-for-checks.tsx)

fregante:
@refined-github/maintainers 
We're going to need help reviewing and merging this soon

source/features/fit-textareas.tsx
```diff
@@ -29,12 +29,15 @@ function fitPrCommitMessageBox(): void {
...
-	delegate(document, 'textarea:not(#pull_request_review_body)', 'focusin', focusListener);
+function init(): Deinit {
 	for (const textArea of select.all('textarea')) {
 		watchTextarea(textArea);
 	}
+
+	// Exclude PR review box because it's in a `position:fixed` container; The scroll HAS to appear within the fixed element.
+	return delegate(document, 'textarea:not(#pull_request_review_body)', 'focusin', event => {
+		focusListener(event);
+	});
```
fregante, line 40: Right?
```suggestion
	return delegate(document, 'textarea:not(#pull_request_review_body)', 'focusin', focusListener);
```

source/features/improve-shortcut-help.tsx
fregante: Agreed

source/github-events/on-pr-merge-panel-open.ts
```diff
@@ -16,26 +16,13 @@ const sessionResumeHandler = mem((callback: EventListener) => async (event: Cust
...
-				'session:resume',
-				sessionResumeHandler(callback),
-			);
+	return [
+		() => {
+			document.removeEventListener('session:resume', sessionResumeHandler(callback));
 		},
-	};
+		delegate(document, '.js-merge-pr:not(.is-rebasing)', 'details:toggled', delegateHandler(callback)),
+	];
```
fregante, line 27: This file is the biggest winner

source/features/wait-for-attachments.tsx
```diff
@@ -31,11 +31,18 @@ function toggleSubmitButtons({target, type}: Event): void {
 	}
 }
 
-function init(): void {
-	document.addEventListener('upload:setup', toggleSubmitButtons, true);
+function init(): Deinit {
+	document.addEventListener('upload:setup', toggleSubmitButtons, {capture: true});
```
fregante, line 35: 🙌

source/features/wait-for-attachments.tsx
```diff
@@ -31,11 +31,18 @@ function toggleSubmitButtons({target, type}: Event): void {
...
 	document.addEventListener('upload:complete', toggleSubmitButtons);
 	document.addEventListener('upload:error', toggleSubmitButtons);
 	document.addEventListener('upload:invalid', toggleSubmitButtons);
+
+	return () => {
+		document.removeEventListener('upload:setup', toggleSubmitButtons, {capture: true});
+		document.removeEventListener('upload:complete', toggleSubmitButtons);
+		document.removeEventListener('upload:error', toggleSubmitButtons);
+		document.removeEventListener('upload:invalid', toggleSubmitButtons);
```
fregante, line 44: In a later cleanup PR, I think we should just create the `signal` helper I mentioned before. This is so error-prone and I don't wanna see it.

The code would instead look like:

```js
import on from '../heh'

 	on(document, 'upload:setup', toggleSubmitButtons, {signal, capture: true});
	on(document, 'upload:complete', toggleSubmitButtons, {signal});
 	on(document, 'upload:error', toggleSubmitButtons, {signal});
 	on(document, 'upload:invalid', toggleSubmitButtons, {signal});
}
```

and it would actually kinda mirror `delegate` (which would also benefit from https://github.com/fregante/delegate-it/issues/26)

source/features/one-click-review-submission.tsx
```diff
@@ -91,6 +84,13 @@ function init(): false | void {
 			}
 		});
 	});
+
+	// This will prevent submission when clicking "Comment" and "Request changes" without entering a comment and no other review comments are pending
+	return delegate(form, 'button', 'click', ({delegateTarget: {value}}) => {
```
fregante, line 89: oof, this function should probably be moved out. O think it's only jere because of `form`, which is actually accessible via `delegateTarget.form`

source/features/pr-branch-auto-delete.tsx
```diff
@@ -2,22 +2,25 @@ import select from 'select-dom';
...
 import observeElement from '../helpers/simplified-element-observer';
+import features, {setupDeinit} from '.';
 
-function init(): void {
+function init(): Deinit {
 	const subscription = delegate(document, '.js-merge-commit-button', 'click', () => {
 		subscription.destroy();
-
-		observeElement('.discussion-timeline-actions', (_, observer) => {
+		setupDeinit(observeElement('.discussion-timeline-actions', (_, observer) => {
```
fregante, line 11: Why `setupDeinit` being used outside the manager? This doesn't look right. This looks like a prime target for `signal` instead. Alternatively it should just be turned into a regular `MutationObserver` in the top level, which can then be regularly returned, unchanged, in the `init` return as `[observer, subscription]`

source/features/reactions-avatars.tsx
```diff
@@ -118,5 +133,5 @@ void features.add(import.meta.url, {
 	include: [
 		pageDetect.isDiscussion,
 	],
-	init: onetime(discussionInit),
+	init: discussionInit,
```
fregante, line 136: I need some extra eyes on this file. helpppp review

source/features/sync-pr-commit-title.tsx
```diff
@@ -79,28 +79,18 @@ async function updateCommitTitle(event: Event): Promise<void> {
 }
 
 function disableSubmission(): void {
-	deinit();
```
fregante, line 82: Doesn't look right? We intentionally drop every listener because we "cancel" the feature entirely. 

If there's no simple solution to this, please revert this file.

fregante (approved):
I reviewed everything with one exception, a couple of nits, a few things to fix later and just one thing that needs to be reverted.

Let's leave it open for a couple of days to allow some extra reviews (plz) then merge as you see fit

yakov116:
I will do some testing tonight

yakov116:
Click on a PR then go to releases tab

```js
refined-github.js:14298 Uncaught TypeError: Cannot read properties of undefined (reading 'classList')
    at ResizeObserver.updateStickiness (refined-github.js:14298:11)
    at later (refined-github.js:23228:28)
```

yakov116:
```js
<html>
<body>
<!--StartFragment--><div aria-expanded="true" tabindex="-1" style="box-sizing: border-box; min-width: 0px; min-height: 0px; flex: 0 0 auto; color: rgb(232, 234, 237); font-family: consolas, &quot;lucida console&quot;, &quot;courier new&quot;, monospace; font-size: 12px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(41, 50, 61); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><span class="source-code" style="box-sizing: border-box; min-width: 0px; min-height: 0px; font-family: var(--source-code-font-family); white-space: pre-wrap; font-size: var(--source-code-font-size) !important; line-height: 1.2;"><span class="console-message-anchor" style="box-sizing: border-box; min-width: 0px; min-height: 0px; float: right; text-align: right; max-width: 100%; margin-left: 4px;"><span class="devtools-link" role="link" tabindex="-1" title="chrome-extension://pplhppbihehddgkdedhpbfpbbbefeihl/build/refined-github.js:18382" style="box-sizing: border-box; min-width: 0px; min-height: 0px; color: var(--color-text-secondary); text-decoration: underline; cursor: pointer; word-break: break-all;">refined-github.js:18382</span> </span><span class="console-message-text" style="box-sizing: border-box; min-width: 0px; min-height: 0px; color: var(--override-error-text-color) !important;">Uncaught (in promise) <span class="object-value-undefined source-code" style="box-sizing: border-box; min-width: 0px; min-height: 0px; font-family: var(--source-code-font-family); white-space: pre-wrap; font-size: var(--source-code-font-size) !important; color: var(--color-text-disabled); line-height: 1.2;">undefined</span></span></span></div><div class="" role="group" style="box-sizing: border-box; min-width: 0px; min-height: 0px; flex: 0 0 auto; color: rgb(232, 234, 237); font-family: consolas, &quot;lucida console&quot;, &quot;courier new&quot;, monospace; font-size: 12px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(41, 50, 61); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><span class="monospace" style="display: inline-block; --monospace-font-size:12px; --monospace-font-family:consolas,lucida console,courier new,monospace; --source-code-font-size:12px; --source-code-font-family:consolas,lucida console,courier new,monospace; box-sizing: border-box; min-width: 0px; min-height: 0px; font-family: var(--monospace-font-family); font-size: var(--monospace-font-size) !important;">

  | onReplacedElement | @ | refined-github.js:18382
-- | -- | -- | --
  | await in onReplacedElement (async) |   |  
  | onDiscussionSidebarUpdate | @ | refined-github.js:16387
  | setupPageLoad | @ | refined-github.js:8220
  | await in setupPageLoad (async) |   |  
  | (anonymous) | @ | refined-github.js:8280
  | dispatch | @ | pjax.ts:97
  | pjaxRequest | @ | pjax.ts:298
  | await in pjaxRequest (async) |   |  
  | click | @ | navigation.ts:70
  | (anonymous) | @ | navigation.ts:18
  | dispatch | @ | index.js:96

</span></div><!--EndFragment-->
</body>
</html>
```

on any page  navigation

source/features/improve-shortcut-help.tsx
cheap-glitch: I restored the `onetime(init)` for all relevant unconditional features, including the ones which create an observer (for the same reason as stated above).

I also added a `onetime` to two features which I think were missing one:
 - `linkify-user-location`
 - `reload-failed-proxied-images`

source/features/wait-for-attachments.tsx
cheap-glitch: Why not make that helper return a wrapper function containing a `removeEventListener` instead? It'd be more consistent with the way `delegate` works now imo.

source/features/pr-branch-auto-delete.tsx
cheap-glitch: > This looks like a prime target for `signal` instead

Sadly this is just a wrapper around a `MutationObserver` which doesn't support `AbortSignal`, so that would mean setting up a event listener on the signal.

> Alternatively it should just be turned into a regular `MutationObserver` in the top level

+1

source/features/sync-pr-commit-title.tsx
cheap-glitch: > We intentionally drop every listener because we "cancel" the feature entirely.

Couldn't the same effect be achieved by adding `{once: true}` to every delegated listener?

source/features/infinite-scroll.tsx
```diff
@@ -52,11 +51,17 @@ function init(): void {
 			{footer}
 		</div>,
 	);
+
+	return [
+		loadMore.cancel,
+		selectorObserver.abort,
+		inView.disconnect,
+	];
```
fregante, line 59:
```suggestion
	return [loadMore, selectorObserver, inView];
```

source/features/sticky-sidebar.tsx
```diff
@@ -17,7 +17,9 @@ function updateStickiness(): void {
 
 const onResize = debounce(updateStickiness, {wait: 100});
 
-function init(): VoidFunction {
+function init(): Deinit[] {
+	document.body.classList.add('rgh-sticky-sidebar-enabled');
```
fregante, line 21: What for?

source/features/sync-pr-commit-title.tsx
fregante: No because they all need to be removed at once, they should not be listened at all after the feature is disabled

fregante (changes requested):
`sync-pr-commit-title` review

source/features/reactions-avatars.tsx
cheap-glitch: @fregante Should I move those changes to a separate PR?

fregante (approved):

fregante:
Let's merge this _after_ the next release, so we have time to review

source/features/infinite-scroll.tsx
cheap-glitch: We do need to keep `loadMore.cancel` though (`loadMore` is just the de-bounced version)

fregante:
v22.3.17 is out, plz merge

cheap-glitch:
### (╯°□°)╯︵ O˥O⅄

fregante:
Hey, by the way, I got good news: `signal` is now supported in Safari 15.4, soooooo, no more `() => el.removeEventListener()` needed 🎉

fregante:
Welp, now MDN says it's been supported since 🥲 should have tested it myself.

https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#browser_compatibility

cheap-glitch:
> Hey, by the way, I got good news: `signal` is now supported in Safari 15.4, soooooo, no more `() => el.removeEventListener()` needed



> Welp, now MDN says it's been supported since v15 :smiling_face_with_tear: should have tested it myself.

Weird, I remember looking at the compat table and it was red. Maybe they only consider the major version because that's how the other browsers are doing it :man_shrugging:

fregante:
Yeah same, they must have updated it now, we were looking at an incorrect table

cheap-glitch:
> I think this broke a few things, I get this error consistently now

Weird, I never get these on Firefox

> and since 22.4.1 I'm now seeing this

+1

source/helpers/get-deinit-handler.ts
```diff
@@ -0,0 +1,15 @@
+export default function getDeinitHandler(deinit: Deinit): VoidFunction {
+	if (deinit instanceof MutationObserver || deinit instanceof ResizeObserver || deinit instanceof IntersectionObserver) {
+		return deinit.disconnect;
```
fregante, line 3:
```suggestion
		return () => deinit.disconnect();

cheap-glitch: I see, I had this bug in Firefox too but only with `controller.abort`.

PR #5425, opened 2022-02-20
Fix `onNewComments` listener

cheap-glitch:
I noticed `onNewComments` is being wrongly triggered in the following cases:
- when adding or removing a reaction on the first comment of a discussion
- when hovering on the dropdown button of a review thread comment
- when editing or removing a regular comment
- when a new event appeared in the timeline (label added/removed, commit pushed, issue linked, etc.)

While it doesn't cause any visible bugs, it still means unnecessary calls and duplicate observers/listeners being created (every feature that acts on `hasComments` pages implicitly listens to `onNewComments`!)

This PR fixes that.

### DIFF
--- a/source/github-events/on-new-comments.ts
+++ b/source/github-events/on-new-comments.ts
@@ -4,7 +4,17 @@ import delegate from 'delegate-it';
 const discussionsWithListeners = new WeakSet();
 const handlers = new Set<VoidFunction>();
 const delegates = new Set<delegate.Subscription>();
-const observer = new MutationObserver(run);
+
+let commentsCount = 0;
+const observer = new MutationObserver(() => {
+	const commentsNewCount = select.all('.js-comment').length;
+	if (commentsNewCount > commentsCount) {
+		commentsCount = commentsNewCount; // Update the count ASAP to avoid duplicate calls
+		run();
+	}
+
+	commentsCount = commentsNewCount;
+});
 
 function run(): void {
 	// Run all callbacks without letting an error stop the loop and without silencing it
@@ -49,6 +59,7 @@ function addListeners(): void {
 	document.addEventListener('pjax:beforeReplace', removeListeners);
 
 	// When new comments come in via AJAX
+	commentsCount = select.all('.js-comment').length;
 	observer.observe(discussion, {
 		childList: true,
 	});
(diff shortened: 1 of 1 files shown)


source/github-events/on-new-comments.ts
```diff
@@ -57,7 +68,7 @@ function addListeners(): void {
 	delegates.add(delegate(document, '.js-ajax-pagination', 'submit', paginationSubmitHandler));
 
 	// Collapsed comments are loaded later using an include-fragment element
-	delegates.add(delegate(document, 'details.js-comment-container include-fragment', 'loadstart', getFragmentLoadHandler(run), true));
+	delegates.add(delegate(document, 'details.js-comment-container include-fragment:not([class])', 'loadstart', getFragmentLoadHandler(run), true));
```
cheap-glitch, line 71: Collapsed review comments aren't loaded with their dropdown anymore.

Instead, they're created with two `include-fragment` that will be loaded when hovering on the dropdown button (the three dots). This selector excludes those.

source/github-events/on-new-comments.ts
```diff
@@ -4,7 +4,17 @@ import delegate from 'delegate-it';
...
 const delegates = new Set<delegate.Subscription>();
-const observer = new MutationObserver(run);
+
+let commentsCount = 0;
+const observer = new MutationObserver(() => {
+	const commentsNewCount = select.all('.js-comment').length;
+	if (commentsNewCount > commentsCount) {
+		commentsCount = commentsNewCount; // Update the count ASAP to avoid duplicate calls
+		run();
+	}
```
cheap-glitch, line 14: The mutation observer is triggered by all kinds of things: new events, comments edited/deleted, reactions added/removed on the body, etc.

Since the mutation records are identical every time, the simplest way I've found to check for the addition of new comments is to keep track of their current number.

Tested with regular as well as review comments.

source/github-events/on-new-comments.ts
```diff
@@ -4,7 +4,17 @@ import delegate from 'delegate-it';
 const discussionsWithListeners = new WeakSet();
 const handlers = new Set<VoidFunction>();
 const delegates = new Set<delegate.Subscription>();
-const observer = new MutationObserver(run);
+
+let commentsCount = 0;
+const observer = new MutationObserver(() => {
+	const commentsNewCount = select.all('.js-comment').length;
```
fregante, line 10: Why not move this inside the `run` function?

fregante:
Let's wait past 22.2.22 before merging this so it can be tested before the next release

source/github-events/on-new-comments.ts
cheap-glitch: Because the other delegates calling `run` can't be triggered by unrelated events.

That said, it might be worth keeping the comment count up-to-date in `run()`, to handle the following situation:
1. The user clicks on "Load more" to load hidden comments
2. This is handled by the first delegate so `commentsCount` isn't updated.
3. Later, the MutationObserver is triggered by a random event (e.g. adding a label) and sees the comment count has changed, so it calls `run` unnecessarily.

PR #5432, opened 2022-02-23
Add `command-palette-navigation-shortcuts` feature

knpwrs:
This PR adds `ctrl`-`n` and `ctrl`-`p` keybindings to the command palette (`cmd`+`k`). These keybindings are common in other developer tools such as vim, vscode, atom, shells, etc.

Relevant feedback post: https://github.com/github/feedback/discussions/11670

## Test URLs

This affects all pages on Github where `cmd+k` opens the command palette (seemingly everywhere).

## Video

https://user-images.githubusercontent.com/174864/155415449-995a71be-4611-4982-aa26-8de881fa12ff.mov


fregante:
I'm _really_ confused.

There's a filter field here, so pressing `p` is currently _filtering_ the list, so you can skip to "Pull request" in one keystroke. **This PR breaks that.**

source/features/command-palette-vim-keybindings.tsx
```diff
@@ -0,0 +1,30 @@
...
+	if (!ctrlKey || (key !== 'n' && key !== 'p')) {
+		return;
+	}
+
+	const commandPalette = select('command-palette');
+
+	if (commandPalette) {
+		const targetKey = key === 'n' ? 'ArrowDown' : 'ArrowUp';
+		commandPalette.dispatchEvent(new KeyboardEvent('keydown', {bubbles: true, key: targetKey, code: targetKey}));
+	}
```
kidonng, line 16: Prefer early return:

```suggestion
	if (!commandPalette) {
		return;
	}

	const targetKey = key === 'n' ? 'ArrowDown' : 'ArrowUp';
	commandPalette.dispatchEvent(new KeyboardEvent('keydown', {bubbles: true, key: targetKey, code: targetKey}));
```

source/features/command-palette-vim-keybindings.tsx
```diff
@@ -0,0 +1,30 @@
...
+}
+
+function init(): void {
+	document.addEventListener('keydown', commandPaletteKeydown);
+}
+
+void features.add(import.meta.url, {
+	shortcuts: {
+		n: 'Select next item in command palette',
+		p: 'Select previous item in command palette',
```
kidonng, line 26:
```suggestion
		'ctrl n': 'Select next item in command palette',
		'ctrl p': 'Select previous item in command palette',
```

kidonng:
Rename the feature to something like `command-palette-navigation-shortcuts`. `vim` does not ring the bell for all users.

source/features/command-palette-vim-keybindings.tsx
knpwrs: I may be misunderstanding how `init` works. I saw that you can check for the existence of things in `init`, but does that work for dynamic elements that can appear after a keystroke as is the case with the command palette?

source/features/command-palette-vim-keybindings.tsx
knpwrs: I actually tried this first and what wound up happening was whenever I hit `ctrl` the next item would be automatically selected (rather than having to press `ctrl-n`). After holding `ctrl`, `n` and `p` would work, but I didn't want `ctrl` to automatically select the next item.

knpwrs:
Feature renamed in 074e2da

source/features/command-palette-vim-keybindings.tsx
knpwrs: Here is a video showing what I am talking about:

https://user-images.githubusercontent.com/174864/155544311-f4b69e63-9304-41c5-a7cc-7b9f68afdc19.mov

That is with this patch applied, and the only key I am pressing is `ctrl`:

```diff
diff --git a/source/features/command-palette-navigation-shortcuts.tsx b/source/features/command-palette-navigation-shortcuts.tsx
index cd81891e..bb824673 100644
--- a/source/features/command-palette-navigation-shortcuts.tsx
+++ b/source/features/command-palette-navigation-shortcuts.tsx
@@ -3,11 +3,7 @@ import onetime from 'onetime';
 
 import features from '.';
 
-function commandPaletteKeydown({key, ctrlKey}: KeyboardEvent): void {
-	if (!ctrlKey || (key !== 'n' && key !== 'p')) {
-		return;
-	}
-
+function commandPaletteKeydown({key}: KeyboardEvent): void {
 	const commandPalette = select('command-palette');
 
 	if (!commandPalette) {
@@ -24,8 +20,8 @@ function init(): void {
 
 void features.add(import.meta.url, {
 	shortcuts: {
-		n: 'Select next item in command palette',
-		p: 'Select previous item in command palette',
+		'ctrl n': 'Select next item in command palette',
+		'ctrl p': 'Select previous item in command palette',
 	},
 	awaitDomReady: false,
 	init: onetime(init),
```

cheap-glitch: This only affects the text added to the shortcut help window (opened with `?`), not the actual implementation.

knpwrs: Oh! I really should have seen that, I had to call `document.addEventListener` in `init` 🤦🏻.

source/features/command-palette-vim-keybindings.tsx
knpwrs: So it looks like `command-palette` is always in the dom, 6cb42b9 implements a change so we only listen for `keydown` on the `command-palette` element.

knpwrs:
I believe that should address all the feedback so far.

source/features/command-palette-navigation-shortcuts.tsx
```diff
@@ -0,0 +1,32 @@
...
+}
+
+function init(): void {
+	const commandPalette = select('command-palette');
+
+	if (!commandPalette) {
+		return;
+	}
+
+	commandPalette.addEventListener('keydown', commandPaletteKeydown);
```
cheap-glitch, line 22: This is the perfect candidate for delegation:

```suggestion
	delegate(document, 'command-palette', 'keydown', commandPaletteKeydown);
```

source/features/command-palette-navigation-shortcuts.tsx
```diff
@@ -0,0 +1,32 @@
+import select from 'select-dom';
+import onetime from 'onetime';
+
+import features from '.';
+
+function commandPaletteKeydown({key, ctrlKey, currentTarget}: KeyboardEvent): void {
```
cheap-glitch, line 6: The type becomes `delegate.Event<KeyboardEvent>` here.

source/features/command-palette-navigation-shortcuts.tsx
```diff
@@ -0,0 +1,32 @@
...
+	if (!currentTarget || !ctrlKey || (key !== 'n' && key !== 'p')) {
+		return;
+	}
+
+	const targetKey = key === 'n' ? 'ArrowDown' : 'ArrowUp';
+	currentTarget.dispatchEvent(new KeyboardEvent('keydown', {bubbles: true, key: targetKey, code: targetKey}));
```
cheap-glitch, line 12: We probably need to call `preventDefault()` somewhere in the listener because of native browsers shortcuts (e.g. Ctrl+P opens up the print dialog in Firefox).

cheap-glitch (changes requested):
I'm a little hesitant about this because it opens up the door to all kinds of possible shortcuts being proposed — when really it should be added by GitHub in the first place.

But since it's pretty low-cost (a delegate + some simple logic), I guess it's fine for now.

That said, Ctrl+N simply doesn't work with Firefox. It's the shortcut to open a new window, and it doesn't look like it can be overwritten (probably for good reasons): https://discourse.mozilla.org/t/unable-to-suppress-browser-shortcut-keys-ctrl-n-ctrl-s-from-webextension-popup/21255

source/features/command-palette-navigation-shortcuts.tsx
knpwrs: I'll add `preventDefault()`. `ctrl-n/p` are working for me in Firefox on macOS, fwiw.

cheap-glitch:
> but calling `preventDefault()` stops the shortcut from working

Just to be clear I was talking about `event.preventDefault()` :smiley: 

> ctrl-n/p is working for me on Firefox on macOS

That's probably because native MacOs shortcuts are not based on Ctrl, right? And Firefox aligns with that.

> Maybe we could also listen for alt for Windows and Linux users?

Let's not over-complicate things, and pick shortcuts that work for everyone. It's too bad it won't align with Vim but that's life.

And let's keep `preventDefault` as a safety measure, it can't hurt anyway.

knpwrs:
`ctrl`-`j`/`k` probably work just as well. How would you feel about listening for all four of `n`/`p`/`j`/`k` but only documenting `j`/`k` in the shortcuts dialog?

fregante:
Before we go further, why add a combination shortcut when the single-key up/down arrows already worm well? p/n are also on the right side so this isn't even "left-hand-only" argument.

kidonng:
I'm mostly an arrow keys user but I can still see this being useful. As mentioned above, moving your hand to touch the arrow keys breaks the flow visibly.

Generally I hold the attitude "if GitHub already does something similar (in this case the keyboard navigation in e.g. commits list), especially if we also do something similar (e.g. `keyboard-navigation`), adding another suits the gist". 

But IMO to blend better (and avoid "one-hand only" arguments) j/k would be better candidates.

fregante:
> especially if we also do something similar (e.g. [`keyboard-navigation`](https://github.com/refined-github/refined-github/blob/main/source/features/keyboard-navigation.tsx)

> avoid conflicts and "one-hand only" arguments, j/k would be better candidates.

That's for j/k, which isn't possible here since we have an actual input field. ctrl-j/k doesn't make much sense I think because that's not used anywhere.

knpwrs:
> Before we go further, why add a combination shortcut when the single-key up/down arrows already work well?

> moving your hand to touch the arrow keys breaks the flow visibly.

Exactly this. That's why ctrl-p/n are so commonly mapped in developer tools.

> But IMO to blend better, avoid conflicts and "one-hand only" arguments, j/k would be better candidates.

My only issue with j/k is familiarity. n/p are mapped in all kinds of tools that keyboard-centric developers already use every day. I used to map j/k in my vim and shell configuration but I wound up switching to n/p because they are so commonly the default mapping for next/previous.

> For this reason alone we're left with ctrl-n and ctrl-p, which are already used on Windows.

I'm still curious if n/p work on Codesandbox for any Windows users who are able to test since Codesandbox is a fairly popular web app.

fregante:
It looks like ctrl-p/n is also implemented in the File finder, so perhaps we implement those. `j/k` don't apply here because it's a field and `ctrl-j/k` isn't a thing, as far as I know.

So:

1. Let's only implement ctrl-p/n
2. Let's ensure that these don't break the regular Windows behavior (both Chrome and Firefox) even if they don't work there
	- Perhaps let's check how GitHub is handling this discrepancy in the File Finder

knpwrs:
I found the portion of Github's code where they are handling ctrl-n/p:

image

source/features/command-palette-navigation-shortcuts.tsx
```diff
@@ -20,11 +20,15 @@ function init(): void {
...
 
-void features.add(import.meta.url, {
-	shortcuts: {
-		'ctrl n': 'Select next item in command palette',
-		'ctrl p': 'Select previous item in command palette',
-	},
-	awaitDomReady: false,
-	init: onetime(init),
-});
+const onMacOs = !!navigator.userAgent.match(/Macintosh/);
```
fregante, line 23: You can use this

https://github.com/refined-github/refined-github/blob/c9d133547f45321c28fb7ddb35ffc08189c9082e/source/github-helpers/index.ts#L73

source/features/command-palette-navigation-shortcuts.tsx
```diff
@@ -0,0 +1,34 @@
...
+}
+
+function init(): void {
+	delegate(document, 'command-palette', 'keydown', commandPaletteKeydown);
+}
+
+const onMacOs = !!navigator.userAgent.match(/Macintosh/);
+
+if (onMacOs) {
+	void features.add(import.meta.url, {
```
fregante, line 26: We should use this instead:
```suggestion
	void features.add(import.meta.url, {
		include: [
			isMac
		],
```

knpwrs:
@fregante feedback implemented in 7a996ed, except that I had to wrap `isMac` in a function.

readme.md
```diff
@@ -412,6 +412,7 @@ Thanks for contributing! 🦋🙌
 - [](# "convert-pr-to-draft-improvements") [Moves the "Convert PR to Draft" button to the mergeability box and adds visual feedback to its confirm button.](https://user-images.githubusercontent.com/1402241/95644892-885f3f80-0a7f-11eb-8428-8e0fb0c8dfa5.gif)
 - [](# "clean-header-search-field") [Clears duplicate queries in the header search field.](https://user-images.githubusercontent.com/1402241/114177338-7c890300-9966-11eb-83a3-a711a76fae99.png)
 - [](# "prevent-pr-merge-panel-opening") Prevents the merge panel from automatically opening on every page load after it’s been opened once.
+- [](# "command-palette-navigation-shortcuts") Adds keyboard shortcuts to select items in command palette using <kbd>ctrl</kbd> <kbd>n</kbd> and <kbd>ctrl</kbd> <kbd>p</kbd>.
```
cheap-glitch, line 415: It should probably be mentioned here that this is a Mac-only feature.

source/features/command-palette-navigation-shortcuts.tsx
```diff
@@ -0,0 +1,34 @@
...
+	const targetKey = key === 'n' ? 'ArrowDown' : 'ArrowUp';
+	target.dispatchEvent(new KeyboardEvent('keydown', {bubbles: true, key: targetKey, code: targetKey}));
+}
+
+function init(): void {
+	delegate(document, 'command-palette', 'keydown', commandPaletteKeydown);
+}
+
+void features.add(import.meta.url, {
+	include: [
```
cheap-glitch, line 25: Just for semantics:

```suggestion
	asLongAs: [
```

cheap-glitch (approved):
LGTM, just a couple of nitpicks :)

yakov116:
Since when do we add `Mac OS` features only?

cheap-glitch:
> Since when do we add `Mac OS` features only?

I think it's fine, as long as it stays very light and simple.

kidonng:
> Since when do we add `Mac OS` features only?

#14

PR #1612, opened 2018-11-07
Add "star/unstar repo" hotkey

adamlacombe:
Resolves #1479

PR #5436, opened 2022-02-24
Drop `star-repo-hotkey`

kidonng:
This feature has been broken from who-knows-when since the "Security" tab occupies the same <kbd>g</kbd> <kbd>s</kbd> shortcut. I indented to drop it (🤫), hence the branch name.

<details>
<summary>Outdated</summary>

Buuuuuuuuuut that shortcut is a waste for the security tab, so here is a fix instead.

Note this does **not** really give you the ability to "toggle" starring status (but the button will _appear_ to be toggled if you press the shortcut multiple times), since nobody is going to do that and it would need a more complex solution to check the current state.

</details>

P.S. if not counting meta/lint/tree-wide refactors, this feature is literally unmodified since added in #1612 😄

## Test URLs

Here


source/features/star-repo-hotkey.tsx
```diff
@@ -4,15 +4,17 @@ import * as pageDetect from 'github-url-detection';
...
-	for (const button of select.all('.js-social-form > button')) {
-		button.dataset.hotkey = 'g s';
-	}
+	// Remove "Security" tab button shortcut
+	select('[data-hotkey="g s"]')!.remove();
+	select(`:is(${[
+		'.starring-container .unstarred', // Star button on a unstarred repo
+		'.starring-container.on .starred', // Unstar button on a starred repo
+	].join(',')}) .btn`)!.dataset.hotkey = 'g s';
 }
```
cheap-glitch, line 13: I think it's okay to duplicate `.btn` and drop the template sting.

fregante:
Wellllllllllllllllllllllllllllllllllllll this means our shortcut changes a native one.

> This feature has been broken from who-knows-when

Goes to show that probably this isn't a very useful/used feature to begin with. I'd say _go forth with your original intent_

kidonng:
> Wellllllllllllllllllllllllllllllllllllll this means our shortcut changes a native one.
>

But I doubt any people would go to Security page that often, unlike our more useful feature.

> > This feature has been broken from who-knows-when
> 
> Goes to show that probably this isn't a very useful/used feature to begin with. I'd say _go forth with your original intent_

It's fine to me personally, but too bad this mnemonic is wasted on some infrequent action for most people (or perhaps GitHub just want every tab to have a shortcut assigned 🥲

fregante (approved):

PR #735, opened 2017-10-18
Make modal backdrop visible

bfred-it:
GitHub's modals have an invisible layer which lets you _click anywhere to close._ This is invisible, which means that if you open a dropdown and then scroll it out of the view you will lose your first click and won't even know why.

# Before

invisible

# After

visible

# PR test areas

- Global: Top right, **profile** dropdown
- Repo: **Watch/Unwatch** dropdown
- Repo: **More** dropdown
- PRs: Files changed: **Review** button dropdown
- PRs: **change PR base** dropdown
- Issues: sidebar **reviewers/labels** dropdowns
- Issues: new comment: **Insert a reply** dropdown
- Issues list: **author/labels/milestones** filters
- Issues list: **filter** button near search box
- Homepage: **organization** dropdown
- Own profile: **contribution settings** + **jump to** dropdown
- Commits list: **verified** commits dropdown

PR #5441, opened 2022-02-27
Lint `refined-github.css`

kidonng:
<h1 align="center">Mega Linter II: The Return of #5399</h1>

Changes to-do

### DIFF
--- a/source/refined-github.css
+++ b/source/refined-github.css
@@ -35,12 +35,6 @@
 	display: none !important;
 }
 
-@keyframes fade-in {
-	from {
-		opacity: 0%;
-	}
-}
-
 /* Smoothen mouse animation on 404 pages */
 .js-plaxify {
 	transition: transform 0.1s;
--- a/source/refined-github.css
+++ b/source/refined-github.css
@@ -35,11 +35,6 @@
 	display: none !important;
 }
 
-/* Hide `New pull request button` near file list */
-.file-navigation .new-pull-request-btn {
-	display: none;
-}
-
 @keyframes fade-in {
 	from {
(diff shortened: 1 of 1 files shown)


source/refined-github.css
```diff
@@ -35,11 +35,6 @@
 	display: none !important;
 }
 
-/* Hide `New pull request button` near file list */
-.file-navigation .new-pull-request-btn {
-	display: none;
-}
-
```
kidonng, line 42: Cannot attach screenshots to commits so here's a review comment instead.

The "New pull request button" used to be like:



Now:

Screen Shot 2022-02-27 at 13 32 56

No need to hide it to make space for other features anymore

source/refined-github.css
```diff
@@ -35,17 +35,6 @@
...
-@keyframes fade-in {
-	from {
-		opacity: 0%;
-	}
-}
```
kidonng, line 47: Added in #735 but no longer used (I didn't bother to find when it was removed)

And we are actually overriding the native one:

```css
/* https://github.githubassets.com/assets/node_modules/@primer/css/utilities/animations.scss */
@keyframes fade-in {
  0% {
    opacity: 0;
  }

  100% {
    opacity: 1;
  }
}
```

fregante:
If you want to submit small changes, feel free to push them to a branch and then open a PR when ready for discussion or close to merge.

fregante:
One more lint would be to move the tests out of test and next to the related helpers, where possible, e.g.

- test/github-url.ts -> source/github-helpers/github-url.**test.**ts

Issue #2713, opened 2020-01-20
`sync-pr-commit-title` renames PR to "Merge pull request #XXX from <branch>"


Blaisorblade:
Thanks to #1743, PRs are renamed to the merge commit title. However, that's typically an uninformative message, such as "Merge pull request #XXX from <branch>" — the PRs then shouldn't be renamed, but are, per the screenshot. The automated rename is
> Blaisorblade changed the title ~~Complete supporting guarded pDOT: Allow paths as value definitions, try 2~~ Merge pull request #155 from Blaisorblade/path-defs-rebase-2

(I later renamed this back by hand).

Disabling `sync-pr-commit-title` fixes the problem. Unfortunately, I'm only merging PRs in a private repo, so I can't (yet) give a public link (but I should be able to, soon).

![image](https://user-images.githubusercontent.com/289960/72699201-ba262400-3b47-11ea-89e7-90b0bd082245.png)

fregante:
`sync-pr-commit-title` lets you disable the feature before merge:

![](https://user-images.githubusercontent.com/1402241/51669708-9a712400-1ff7-11e9-913a-ac1ea1050975.png)

You can just click `cancel`

Blaisorblade:
Thanks for the answer; the real bug is the default title, but that isn't due to this extension.
GitHub proposes `Merge pull request #XXX from ...` as commit title, but only on some repos :-(.

fregante:
Actually `sync-pr-commit-title` is made of two parts:

1. First it copies the PR title to the PR commit title (even for merge commits, it should replace the default `Merges...`)
2. Then if the user make changes to this title, it should also update the PR title

So in this case it appears that the first part failed (so it remained `Merges...`) while the second part succeeded (so it set the PR title to `Merges...`)

If this behavior is reproducible, this is a bug somewhere here:

https://github.com/sindresorhus/refined-github/blob/22764b8ba0475cef2a87adf134acb828852daaa8/source/features/sync-pr-commit-title.tsx#L87-L89

jsumners:
@Blaisorblade thank you for filing this issue. It helped me figure out what to disable to get rid of this annoyance.

Blaisorblade:
@jsumners A request: has it occurred on a *public* repo? Can you post a link? All my examples are on a private repo 😕.

jsumners:
@Blaisorblade sorry, I am only seeing this in my private repos at the moment.

thibaudcolas:
[...]

PR #5448, opened 2022-03-02
Change: PR title sync to affect squash *and* merge

binury:
I didn't find any means to configure a feature within the extension options UI or the extension architecture itself… I imagine this change to `sync-pr-commit-title` would be appreciated generally but, of course, not by any supermajority especially for existing users. (Changes `sync-pr-commit-title` to apply to _both_ squash *and* merge commit messages.)

I considered _cloning the feature in lieu of configuration_, so that one would continue to apply only to squashes and the other to merge commits, but in addition to being redundant this _wouldn't really work since all features come enabled by default_.

This PR is then meant as an inquiry foremost:
### **Is there any** (ideally, productive) **alternative to feature configuration that isn't as drastic as hard forking the extension?**
 and perhaps for discussion, regarding the ability -in general- to configure features.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -297,7 +297,7 @@ Thanks for contributing! 🦋🙌
 
 ### Editing pull requests
 
-- [](# "sync-pr-commit-title") 🔥 [Uses the PR’s title as the default squash commit title](https://github.com/refined-github/refined-github/issues/276) and [updates the PR’s title to match the commit title, if changed.](https://user-images.githubusercontent.com/1402241/51669708-9a712400-1ff7-11e9-913a-ac1ea1050975.png)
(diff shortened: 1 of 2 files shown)


busches:
The opinions on configuration are laid out here: https://github.com/refined-github/refined-github/issues/2960

In regards to this change, I recall it used to be this way, but maybe I'm misremembering, git history should help you answer it.

binury:
> The opinions on configuration are laid out here: #2960
> 
> In regards to this change, I recall it used to be this way, but maybe I'm misremembering, git history should help you answer it.

It did at one point apply to both but then… 

1. [changed as a workaround](https://github.com/refined-github/refined-github/issues/2713#issuecomment-579097234) 
2. #2738 
3. #2768 was closed but never resolved



Maybe if it is no longer needed and PR titles aren't being accidentally updated/"sync'd" to the default merge commit message, reverting through this PR would make sense after all. I'll take a look and try to replicate #2713

fregante:
That issue was closed, so this changed was already rejected multiple times

PR #5457, opened 2022-03-04
Meta: Warn for missing or invalid token in bug report form

cheap-glitch:
Might help avoiding this kind of situations in the future: https://github.com/refined-github/refined-github/issues/5403

### DIFF
--- a/source/features/rgh-improve-new-issue-form.tsx
+++ b/source/features/rgh-improve-new-issue-form.tsx
@@ -3,29 +3,10 @@ import select from 'select-dom';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
+import openOptions from '../helpers/open-options';
 import clearCacheHandler from '../helpers/clear-cache-handler';
+import {expectTokenScope} from '../github-helpers/api';
 import {isRefinedGitHubRepo} from '../github-helpers';
-import {expectToken, expectTokenScope} from '../github-helpers/api';
-
-async function validateToken(): Promise<void> {
-	try {
-		await expectToken();
-	} catch {
-		throw new Error('Refined GitHub needs a Personal Access Token to work properly.');
-	}
-
-	try {
-		await expectTokenScope('repo');
-	} catch {
-		throw new Error('Your Personal Access Token is valid but missing the `repo` scope. Many features will not work.');
-	}
-
-	try {
-		await expectTokenScope('delete_repo');
-	} catch {
-		throw new Error('Your Personal Access Token is valid but missing the `delete_repo` scope. The `quick-repo-deletion` feature will not work.');
-	}
-}
 
 function init(): void {
 	const {version} = browser.runtime.getManifest();
@@ -39,10 +20,11 @@ function init(): void {
 			Clear cache
 		</button>,
 	);
-	void validateToken().catch(error => {
+	void expectTokenScope('repo').catch(() => {
 		select('#issue_body_template_name')!.before(
 			<div className="flash flash-warn m-2">
-				{error.message}
+				Your Personal Access Token is either missing, incorrect or expired. Some features will not work without it.<br/>
+				You can update it <a href="#" onClick={openOptions as unknown as React.MouseEventHandler}>in the options</a>.
 			</div>,
 		);
 	});
--- a/source/features/rgh-improve-new-issue-form.tsx
+++ b/source/features/rgh-improve-new-issue-form.tsx
@@ -43,7 +43,7 @@ function init(): void {
 		select('#issue_body_template_name')!.before(
 			<div className="flash flash-warn m-2">
 				{error.message}
-			</div>
+			</div>,
 		);
 	});
 }
--- a/source/features/rgh-improve-new-issue-form.tsx
+++ b/source/features/rgh-improve-new-issue-form.tsx
@@ -23,7 +23,7 @@ function init(): void {
 	void expectTokenScope('repo').catch(() => {
 		select('#issue_body_template_name')!.before(
 			<div className="flash flash-warn m-2">
-				Your Personal Access Token is either missing, incorrect or expired. Some features will not work without it.<br/>
+				Your Personal Access Token is either missing, incorrect or expired. Some features of Refined GitHub will not work without it.<br/>
 				You can update it <a href="#" onClick={openOptions as unknown as React.MouseEventHandler}>in the options</a>.
 			</div>,
 		);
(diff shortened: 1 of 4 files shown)


kidonng:
Wonder if we can check (no pun intended) it automatically.

fregante:
Indeed I'd probably just add a "Your token is not set/valid, so [some features](#link) are won't work. Don't open an issue regarding those features."

We can also automatically mark such issues by injecting a secret character in the form, e.g. use version "22.3.3nt"

source/features/rgh-improve-new-issue-form.tsx
```diff
@@ -5,6 +5,27 @@ import * as pageDetect from 'github-url-detection';
...
+		await expectTokenScope('repo');
+	} catch {
+		throw new Error('Your Personal Access Token is valid but missing the `repo` scope. Many features will not work.');
+	}
+
+	try {
+		await expectTokenScope('delete_repo');
+	} catch {
+		throw new Error('Your Personal Access Token is valid but missing the `delete_repo` scope. The `quick-repo-deletion` feature will not work.');
+	}
```
cheap-glitch, line 27: Not sure if it's worth having that precise check

source/features/rgh-improve-new-issue-form.tsx
```diff
@@ -5,6 +5,27 @@ import * as pageDetect from 'github-url-detection';
 import features from '.';
 import clearCacheHandler from '../helpers/clear-cache-handler';
 import {isRefinedGitHubRepo} from '../github-helpers';
+import {expectToken, expectTokenScope} from '../github-helpers/api';
+
+async function validateToken(): Promise<void> {
+	try {
+		await expectToken();
+	} catch {
+		throw new Error('Refined GitHub needs a Personal Access Token to work properly.');
```
cheap-glitch, line 14: A link to the options page here would be nice.

Actually, it'd be nice to be able to drop a special link/placeholder to create a link to the options page (and maybe even a "Clear cache" button), in any issue conversation on the RGH repo.

cheap-glitch (commented):
Dropped the checkbox in favor of a warning box.

source/features/rgh-improve-new-issue-form.tsx
fregante: > it'd be nice to be able to drop a special link/placeholder

We're still in the extension context, it's easy to create this:


```js
function OptionsButton(): JSX.Element {
	return <button onClick={() => browser.runtime.openOptionsPage()}>Options la la la</a>
}
```

and then just use `select('yo').append(<OptionsButton/>)

source/features/rgh-improve-new-issue-form.tsx
fregante: I don’t think so, I think the regular `expectTokenScope` error is clear enough. We can just append "[Some features](#the search link we have in the options page) require it"

source/features/rgh-improve-new-issue-form.tsx
cheap-glitch: Actually we already kind of have this, in `rgh-welcome-issue`. It just needs to be generalized to any issue conversation.

source/features/rgh-improve-new-issue-form.tsx
fregante: I don't think we need this `quick-repo-deletion` special message either, the feature should already be clear about this when failing.

source/features/rgh-improve-new-issue-form.tsx
fregante: That's a bit different because the welcome issue exists without the extension so that piece of code is just _extending_ the page. Here instead we're adding a whole block of text.

Technically we can do the same for the "Clear cache" button though:


```js
// rgh-improve-new-issue-form
select('input[id*="extension_cache"]')!.parentElement!.after(
		<ClearCacheButton/>,
	);
```


```
//options.tsx
select('p#clear-cache')!.append(
		<ClearCacheButton/>,
	);
```

cheap-glitch (commented):
Simplified the checks and added a link to the options page in the warning box:

image

source/features/rgh-improve-new-issue-form.tsx
```diff
@@ -18,6 +20,14 @@ function init(): void {
 			Clear cache
 		</button>,
 	);
+	void expectTokenScope('repo').catch(() => {
+		select('#issue_body_template_name')!.before(
+			<div className="flash flash-warn m-2">
+				Your Personal Access Token is either missing, incorrect or expired. Some features will not work without it.<br/>
+				You can update it <a href="#" onClick={openOptions as unknown as React.MouseEventHandler}>in the options</a>.
```
yakov116, line 27: Maybe "you can update it in the extension options"

Not sure if it's needed.

cheap-glitch: I added a mention to RGH to clarify what the notice is about.

source/features/rgh-improve-new-issue-form.tsx
```diff
@@ -18,6 +20,14 @@ function init(): void {
 			Clear cache
 		</button>,
 	);
+	void expectTokenScope('repo').catch(() => {
+		select('#issue_body_template_name')!.before(
+			<div className="flash flash-warn m-2">
+				Your Personal Access Token is either missing, incorrect or expired. Some features of Refined GitHub will not work without it.<br/>
```
yakov116, line 26: Some Refined GitHub features...

yakov116 (approved):

source/features/rgh-improve-new-issue-form.tsx
```diff
@@ -18,6 +20,14 @@ function init(): void {
 			Clear cache
 		</button>,
 	);
+	void expectTokenScope('repo').catch(() => {
+		select('#issue_body_template_name')!.before(
+			<div className="flash flash-warn m-2">
+				Your Personal Access Token is either missing, incorrect or expired. Some Refined GitHub features will not work without it.<br/>
```
fregante, line 26: We can be more specific here, maybe with a custom API call to check, in order:

- existence (no call at all)
- validity
- expiration
- lack of expected scope

source/features/rgh-improve-new-issue-form.tsx
```diff
@@ -18,6 +20,14 @@ function init(): void {
...
+				You can update it <a href="#" onClick={openOptions as unknown as React.MouseEventHandler}>in the options</a>.
```
fregante, line 27: I forgot we can use `<button className="btn-link">`

fregante:
Merging for now, will suggested changes to meta issue

PR #5464, opened 2022-03-04
Refactor `shorten-links` code

dnicolson:
This pull request fixes tree links that do not shorten unless in preview mode. The solution is to call the `applyToLink` function on page load as well as on DOM changes.

I am not certain about removing the `.js-suggested-changes-blob` selector so it's entirely possible this breaks other things.

## Test URLs
https://github.com/refined-github/refined-github/pull/4759
https://github.com/dnicolson/x/issues/1

## Screenshot
Before:
bad (1)

After:
good (1)

### DIFF
--- a/source/features/shorten-links.tsx
+++ b/source/features/shorten-links.tsx
@@ -15,12 +15,6 @@ function init(): void {
 			}
 		},
 	});
-
-	for (const link of document.querySelectorAll('.comment-body a[href*="/tree/"]')) {
-		if (!isLinkInCode(link)) {
-			applyToLink(link, location.href);
-		}
-	}
 }
 
 function isLinkInCode(link: HTMLAnchorElement): boolean {
--- a/source/features/shorten-links.tsx
+++ b/source/features/shorten-links.tsx
@@ -10,16 +10,24 @@ function init(): void {
 	observe(`a[href]:not(.${linkifiedURLClass})`, {
 		constructor: HTMLAnchorElement,
 		add(link) {
-			// Don't shorten links in code or code suggestions (but shorten them in review comments)
-			// Code explanation: Exclude links if the closest element found is not `.comment-body`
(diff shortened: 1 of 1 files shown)


kidonng:
Did `shorten-links` break recently? Because I'm seeing #4758 again.

fregante:
> Because I'm seeing #4758 again.

I'm not on Safari, but I'm not on 22.3.3

Screen Shot 20

fregante:
> Shorten tree links outside of preview mode

It works for me on your demo URL:

Screen Shot 22


An issue may exist when a comment is updated and we don't detect it. This piece of code has a long history (of bugs) so we need to be super careful when editing it.

source/features/shorten-links.tsx
```diff
@@ -10,16 +10,24 @@ function init(): void {
...
-				return;
+			if (!isLinkInCode(link)) {
+				applyToLink(link, location.href);
 			}
-
-			applyToLink(link, location.href);
 		},
 	});
+
+	for (const link of document.querySelectorAll('.comment-body a[href*="/tree/"]')) {
```
fregante, line 19: This already happens because `observe` automatically calls `add` on load, on every `a[href]`, which includes these.

fregante (approved):
I can merge this `isLinkInCode` extraction though

fregante:
Actually it needs to be updated preserving changes from https://github.com/refined-github/refined-github/pull/5472

cheap-glitch (changes requested):
The bug is not caused by `shorten-links` but by `comments-time-machine-links`, which adds a query parameter to tree URLs, making this condition fail in `applyToLink`:

https://github.com/refined-github/shorten-repo-url/blob/a7c81316f438c38ce71b13309b974f0eb8c6838a/index.js#L239

Example here: https://github.com/refined-github/refined-github/issues/5479#issue-1166000758

PR #5465, opened 2022-03-05
Fix `wait-for-checks` on PR branch with no checks

### LINKED ISSUE #5449: `wait-for-checks` feature hangs forever if the PR doesn't have any checks at all (but `main` does)
### What I don't like

I have one repository that only runs certain CI job on the master branch, so whenever I opened a new PR, there will be no commit status for those commits. However, if I clicked the `Merge` button, I would see the checkbox `wait for builds`, which does not make sense. The checkbox got displayed because it detected one CI commit status icon, which is the one near the repository name (the CI I made for master branch only). And if the checkbox is enabled, the merge will never finish, as the `wait-for-builds` feature will wait forever. 

The source code caused this is https://github.com/refined-github/refined-github/blob/main/source/features/wait-for-checks.tsx#L43

image

cheap-glitch:
Fixes #5449 

## Test URLs

n/a

## Screenshot

TODO

### DIFF
--- a/source/features/wait-for-checks.tsx
+++ b/source/features/wait-for-checks.tsx
@@ -33,15 +33,23 @@ function getCheckbox(): HTMLInputElement | undefined {
 	return select('input[name="rgh-pr-check-waiter"]');
 }
 
+function ciIconPresentOnPage(): boolean {
+	const commitsCount = select.all(prCiStatus.commitSelector).length;
+	if (commitsCount > 1) {
+		// If more than one commit was pushed, only check for a status icon in the conversation feed
+		return select.exists(`#discussion_bucket ${prCiStatus.commitStatusIconSelector}`);
+	}
+
+	// Otherwise, search on the whole page (including `ci-link`)
+	return select.exists(prCiStatus.commitStatusIconSelector);
+}
+
 // Only show the checkbox if the last commit doesn't have a green or red CI icon
 function showCheckboxIfNecessary(): void {
 	const checkbox = getCheckbox();
 	const lastCommitStatus = prCiStatus.getLastCommitStatus();
 
-	const isNecessary = lastCommitStatus === prCiStatus.PENDING
-		// If the latest commit is missing an icon, add the checkbox as long as there's at least one CI icon on the page (including `ci-link`)
-		|| (lastCommitStatus === false && select.exists(prCiStatus.commitStatusIconSelector));
-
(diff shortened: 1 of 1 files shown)


cheap-glitch (commented):
**This has not yet been properly tested!**

This PR:
- improves the DOM checks that decide if the checkbox is necessary, following the suggestions of https://github.com/refined-github/refined-github/issues/5449#issuecomment-1057748188
- aborts waiting if the commit still has no checks running on the second attempt (which basically acts as a 5~10 seconds delay, in case the CI is slow to start)

fregante:
This feature is probably the largest and most fickle feature and it's not even our fault 🥲

cheap-glitch:
> This feature is probably the largest and most fickle feature and it's not even our fault

I'm wondering if we could replace the various DOM checks with a simple API call. We could even drop the observer, and just check if a new commit has been posted in the `p-retry` callback.

cheap-glitch:
> I'm wondering if we could replace the various DOM checks with a simple API call

Just noticed that committing a suggestion from the conversation tab of a PR, doesn't create a new commit "event" in the timeline about 2 out of 3 times :man_facepalming: Which means that unless you reload the page, the checkbox won't appear.

We definitely shouldn't rely on the DOM _at all_ for this feature. It should just talk directly with the API for (almost) everything.

fregante:
This was partially my suggestion in https://github.com/refined-github/refined-github/issues/5336#issuecomment-1024108876

How about this, to streamline it further:

- When the user opens the merge box, start polling the API
	- If there's a pending/inprogress check
		- show/hide the checkbox
		- hold the submission until:
			- all green: submit
			- some failed: cancel submission and close merge ability box

I expect most of the feature to be dropped with the exception of the mergeability box DOM changes (checkbox, submission holder)

cheap-glitch:
@fregante I'm working on a new PR to replace this one. How about adding a very basic CI with some `sleep` commands to https://github.com/refined-github/sandbox so we can have a proper testing ground?

Ideally I'd like to be able to create "failing commits" on purpose, e.g. have the CI fail if the commit message is `FAIL_CI` or something like that. Any pointers?

fregante:
You can commit anything to that repo, but just for temporary testing, don't necessarily expect it to stay there forever

fregante:
See https://github.com/refined-github/refined-github/issues/5449#issuecomment-1115597517

PR #5472, opened 2022-03-08
Don't apply `shorten-links` to links in code blocks

### LINKED ISSUE #4758: `shorten-links` conflicts with OctoLinker

cheap-glitch:
Fixes #4758 
Follow-up of #4759 

## Test URLs

- [Highlighted and non-highlighted code blocks in comment](https://togithub.com//denosaurs/mod.land/issues/55)
- [Code block in readme](https://togithub.com/oakserver/oak#application-middleware-and-context)
- [Review suggestion](https://togithub.com/refined-github/refined-github/pull/4759#discussion_r738167140)

## Screenshot

**Before**

before

**After**

after

### DIFF
--- a/source/features/shorten-links.tsx
+++ b/source/features/shorten-links.tsx
@@ -10,14 +10,12 @@ function init(): void {
 	observe(`a[href]:not(.${linkifiedURLClass})`, {
 		constructor: HTMLAnchorElement,
 		add(link) {
-			// Don't shorten links in code or code suggestions (but shorten them in review comments)
-			// Code explanation: Exclude links if the closest element found is not `.comment-body`
+			// Exclude the link if the closest element found is not `.comment-body`
+			// This avoids shortening links in code and code suggestions, but still shortens them in review comments
 			// https://github.com/refined-github/refined-github/pull/4759#discussion_r702460890
-			if (link.closest('.blob-code, .comment-body, .js-suggested-changes-blob')?.matches('.blob-code, .js-suggested-changes-blob')) {
-				return;
+			if (link.closest('pre, .blob-code, .js-suggested-changes-blob, .comment-body')?.classList.contains('comment-body')) {
+				applyToLink(link, location.href);
 			}
-
-			applyToLink(link, location.href);
 		},
 	});
 }
--- a/source/features/shorten-links.tsx
+++ b/source/features/shorten-links.tsx
@@ -5,6 +5,7 @@ import * as pageDetect from 'github-url-detection';
 
 import features from '.';
 import {linkifiedURLClass} from '../github-helpers/dom-formatters';
+import {codeElementsSelectors} from './show-whitespace';
 
 function init(): void {
 	observe(`a[href]:not(.${linkifiedURLClass})`, {
@@ -13,7 +14,7 @@ function init(): void {
 			// Exclude the link if the closest element found is not `.comment-body`
 			// This avoids shortening links in code and code suggestions, but still shortens them in review comments
 			// https://github.com/refined-github/refined-github/pull/4759#discussion_r702460890
-			if (link.closest('pre, .blob-code, .js-suggested-changes-blob, .comment-body')?.classList.contains('comment-body')) {
+			if (link.closest(`${codeElementsSelectors}, .js-suggested-changes-blob, .comment-body`)?.classList.contains('comment-body')) {
 				applyToLink(link, location.href);
 			}
 		},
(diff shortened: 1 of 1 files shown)


source/features/shorten-links.tsx
```diff
@@ -10,14 +10,12 @@ function init(): void {
...
 		constructor: HTMLAnchorElement,
 		add(link) {
-			// Don't shorten links in code or code suggestions (but shorten them in review comments)
-			// Code explanation: Exclude links if the closest element found is not `.comment-body`
+			// Exclude the link if the closest element found is not `.comment-body`
+			// This avoids shortening links in code and code suggestions, but still shortens them in review comments
 			// https://github.com/refined-github/refined-github/pull/4759#discussion_r702460890
-			if (link.closest('.blob-code, .comment-body, .js-suggested-changes-blob')?.matches('.blob-code, .js-suggested-changes-blob')) {
-				return;
+			if (link.closest('pre, .blob-code, .js-suggested-changes-blob, .comment-body')?.classList.contains('comment-body')) {
```
cheap-glitch, line 16: The only real change is the addition of `pre` in the selector. Everything else is just refactoring.

source/features/shorten-links.tsx
```diff
@@ -5,19 +5,18 @@ import * as pageDetect from 'github-url-detection';
...
 		constructor: HTMLAnchorElement,
 		add(link) {
-			// Don't shorten links in code or code suggestions (but shorten them in review comments)
-			// Code explanation: Exclude links if the closest element found is not `.comment-body`
+			// Exclude the link if the closest element found is not `.comment-body`
+			// This avoids shortening links in code and code suggestions, but still shortens them in review comments
 			// https://github.com/refined-github/refined-github/pull/4759#discussion_r702460890
-			if (link.closest('.blob-code, .comment-body, .js-suggested-changes-blob')?.matches('.blob-code, .js-suggested-changes-blob')) {
-				return;
+			if (link.closest(`${codeElementsSelectors}, .js-suggested-changes-blob, .comment-body`)?.classList.contains('comment-body')) {
```
cheap-glitch, line 17: The only real change is the addition of the `codeElementSelectors`. Everything else is just refactoring.

fregante: Can you move this selector to its own array (plus .join())? Maybe with comments.

fregante (approved):
We can merge this as is, but would you be able to add some tests for this function? It should accept a `Document | Element` root and then we can use snapshots to ensure that the DOM looks like what we expect (it should probably also be extracted to the helpers folder or else AVA will fail)

source/features/shorten-links.tsx
cheap-glitch: I've dropped `.js-suggested-changes-blob` which was unnecessary. Now it's just `codeElementSelectors` which is documented in `show-whitespace`.

cheap-glitch:
> It should accept a `Document | Element` root

What should we use as testing subjects? I see two options:

1. A copy of the markup of a GitHub page, stored in a file. Problem: it will become obsolete as soon as GitHub changes its DOM.
2. Th URL of a GitHub page (via `JSDOM.fromURL()`). Problem: the snapshots will become obsolete when GitHub changes its DOM and the tests will fail.

fregante:
> What should we use as testing subjects? I see two options:

I think a barebones inline HTML snippet is enough, we just want to make sure that we don't keep causing regressions on this piece of code. The snippet can often be something like:

```js
<pre class="code-blob">
	<div class="the-comment-class-we-expect">
		<div class="markdown-body">
			<pre class="code-blob"
				<a>https://google.com</a>

source/features/shorten-links.tsx
kidonng: > The only real change is the addition of the `codeElementSelectors`.

I didn't quite follow #4759 but it seems this should have been done in that PR 😅

> I've dropped `.js-suggested-changes-blob` which was unnecessary (because superseded by `.blob-code-inner`). Now it's just `codeElementSelectors` which is documented in `show-whitespace`.

Ha! That is essentially reverting [`930f860` (#4759)](https://github.com/refined-github/refined-github/pull/4759/commits/930f860f7d76152d6f944b9c1e66cb980eabcd7f).

PR #5478, opened 2022-03-10
Make Options page scrollbar dark in Chrome

rakleed:
Make a dark scrollbar in Chromium-based browsers in a dark theme.

## Test URLs

Refined GitHub options

## Screenshot

<details><summary>Before</summary>

Screenshot 2022-03-10 at 16 55 53

</details>

<details><summary>After</summary>

Screenshot 2022-03-10 at 16 56 23

</details>


cheap-glitch:
Thanks for the PR! This change would benefit https://github.com/fregante/webext-base-css in general, so it should probably happen there.

fregante:
Indeed I'm not sure what I meant by that comment 🤔 

@cheap-glitch What do you think about this change? I think the only concern is that this also affects fields so I need to be careful across browsers. That's probably what I was alluding to: "it works" as in "I already correctly color the fields for dark mode" — but didn't think of the scrollbars. 

@rakleed What does this look like in Firefox?

rakleed:
> @rakleed What does this look like in Firefox?

Since in Firefox the extension settings are displayed in a tab, and not in a pop-up window like in Chrome, in Firefox I can see the scrollbar only in the Custom CSS field. But in this field, it is so dark even without my PR, and nothing changes with my PR.

rakleed:
> Can you send a PR to my module?

Done: https://github.com/fregante/webext-base-css/pull/9

fregante:
Can you update the module here so we can merge this PR?

rakleed:
@fregante I've updated the dependency.

fregante (approved):

fregante:
T h a n k y o u P a v e l !

PR #5487, opened 2022-03-13
Fix GitHub Safari bug: Horizontal scrollbars on Action logs

fregante:
- Closes https://github.com/refined-github/refined-github/issues/5152

## Test URL: 

https://github.com/pixiebrix/pixiebrix-extension/runs/5523238774?check_suite_focus=true

## Before

Screen Shot 7

## After

Screen Shot 6

### DIFF
--- a/source/features/github-bugs.css
+++ b/source/features/github-bugs.css
@@ -49,3 +49,8 @@
 #toc .toc-diff-stats .octicon-file-diff {
 	margin-right: 4px !important;
 }
+
+/* Avoid horizontal scrollbars on Action logs in Safari #5152 */
+:root .CheckStep-line .CheckStep-line-content {
+	overflow: hidden;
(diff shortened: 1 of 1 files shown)


source/features/github-bugs.css
```diff
@@ -49,3 +49,8 @@
 #toc .toc-diff-stats .octicon-file-diff {
 	margin-right: 4px !important;
 }
+
+/* Avoid horizontal scrollbars on Action logs in Safari #5152 */
+:root .CheckStep-line .CheckStep-line-content {
+	overflow: hidden;
```
fregante, line 55: YOLO

this however is better than my suggested "visible" because it does not cause a viewport scrollbar

cheap-glitch (approved):
Looks OK on Firefox

PR #5488, opened 2022-03-13
Fix Safari first install detection

fregante:
- Fixes https://github.com/refined-github/refined-github/issues/5412

### DIFF
--- a/source/background.ts
+++ b/source/background.ts
@@ -1,5 +1,6 @@
 import 'webext-dynamic-content-scripts';
 import cache from 'webext-storage-cache'; // Also needed to regularly clear the cache
+import {isSafari} from 'webext-detect-page';
 import addDomainPermissionToggle from 'webext-domain-permission-toggle';
 
 import optionsStorage from './options-storage';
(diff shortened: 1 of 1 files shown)


source/background.ts
```diff
@@ -51,9 +52,30 @@ browser.browserAction.onClicked.addListener(async () => {
...
+		await browser.storage.sync.getBytesInUse() > 0
+		// Note: Not available in Firefox https://bugzilla.mozilla.org/show_bug.cgi?id=1385832
+		|| Number(await browser.storage.local.getBytesInUse?.()) > 0
+	);
+}
+
+async function isFirstInstall(suggestedReason: string): Promise<boolean> {
+	return (
+		// Always exclude local installs from the welcome screen
+		!isDevelopmentVersion()
```
fregante, line 66: I tested this in Chrome with and without this line and it seems that the logic is still sound at least there

source/background.ts
```diff
@@ -51,9 +52,30 @@ browser.browserAction.onClicked.addListener(async () => {
...
+
+		// Only if the reason is explicitly "install"
+		&& suggestedReason === 'install'
+
+		// Safari reports "install" even on updates #5412
+		&& (isSafari() ? !(await hasUsedStorage()) : true)
```
cheap-glitch, line 72:
```suggestion
		&& !(isSafari() && await hasUsedStorage())
```

fregante: Sounds correct. Let me nap on it.

PR #5491, opened 2022-03-13
Move static files out of distribution + add prepare script for Safari

fregante:
- Closes https://github.com/refined-github/refined-github/issues/5250

@sindresorhus hopefully this means we'll see Safari updates more often 🙂


safari/Config.xcconfig
```diff
@@ -1,4 +1,4 @@
-MARKETING_VERSION = 22.2.13
-CURRENT_PROJECT_VERSION = 18
+MARKETING_VERSION = 0.0.0
+CURRENT_PROJECT_VERSION = 1
```
fregante, line 2: These are just placeholders just for Xcode and are not actually enough for the script to run. You'll have to copy them to your `LocalOverrides` file

build/prepare-safari-release.sh
```diff
@@ -0,0 +1,29 @@
...
+
+CONFIG_FILE=./safari/LocalOverrides.xcconfig
+
+TAG=$(git describe --tags --abbrev=0)
+
+if [[ $(git describe --tags) != "$TAG" ]]; then
+	echo You’re ahead of the latest tag. Run:
+	echo git checkout "$TAG"
+	exit 1
+fi
```
fregante, line 14: This is to ensure we don't get mismatched releases (e.g. "it's 22.3.3 _but_ there are 3 more bugfixes")

If you need to make small changes, I think you'll be able to:

1. check out the tag
2. make the changes
3. run this script
4. _then_ commit them

package.json
```diff
@@ -12,6 +12,7 @@
 		"lint:js": "xo",
 		"pack:safari": "xcodebuild -project 'safari/Refined GitHub.xcodeproj' -scheme 'Refined GitHub (macOS)'",
 		"start:safari": "open 'safari/build/Release/Refined GitHub.app'",
```
fregante, line 14: This command is also broken since the iOS PR probably, we should drop it if there's no fix
```suggestion
```

sindresorhus: Yeah, it's better to just build and run in Xcode.

fregante: The point is that I try to avoid opening XCode at all because I shouldn't have to deal with it. That's why there's a build:safari script as well

fregante:
I tested this with two other extensions and it works flawlessly. I'll tag a new release today so @sindresorhus please try it soon. The other change I requested can come later as it's just a DX thing for local testing

PR #5493, opened 2022-03-13
Ensure custom CSS is added *after* the extension CSS

fregante:
- Closes https://github.com/refined-github/refined-github/issues/4731#issuecomment-1066059907
- Replaces #5492 

Not yet tested

### DIFF
--- a/source/features/index.tsx
+++ b/source/features/index.tsx
@@ -98,14 +98,14 @@ const globalReady: Promise<RGHOptions> = new Promise(async resolve => {
 		bisectFeatures(),
 	]);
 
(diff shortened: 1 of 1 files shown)


fregante:
Indeed this seems to work:

- `<style>` is the custom style I added
- _injected stylesheet_ is `refined-github.css`

Screen Shot 8

cheap-glitch (approved):
Can confirm it's working on Firefox too:

```css
.rgh-hide-diff-signs .blob-code-inner::before,
.rgh-hide-diff-signs .blob-code-marker-cell::before {
	visibility: visible !important;
}
```

image

PR #5496, opened 2022-03-13
Include but skip inert URL segments in `useful-not-found-page`

fregante:
- Closes https://github.com/refined-github/refined-github/issues/5483

## Test URLs

https://github.com/orgs/date-fns/teams/i18n-id

## Before
Screen Shot 9

## After
Screen Shot 10

### DIFF
--- a/source/features/useful-not-found-page.tsx
+++ b/source/features/useful-not-found-page.tsx
@@ -72,32 +72,29 @@ async function getUrlToFileOnDefaultBranch(): Promise<string | void> {
 
 async function showMissingPart(): Promise<void> {
 	const pathParts = parseCurrentURL();
-	const breadcrumbs: HTMLElement[] = [];
+	const breadcrumbs = [...pathParts.entries()]
+		.reverse() // Checks the anchors right to left
+		.map(([i, part]) => {
+			if (
+				// Exclude parts that don't exist as standalones
+				(i === 0 && part === 'orgs') // #5483
+				|| (i === 2 && ['tree', 'blob', 'edit'].includes(part))
+				|| i === pathParts.length - 1 // The last part is a known 404
+			) {
+				return getStrikeThrough(part);
+			}
 
-	for (const [i, part] of pathParts.entries()) {
-		// Exclude parts that don't exist as standalones
-		if (
-			(i === 0 && part === 'orgs')
-			|| (i === 2 && ['tree', 'blob', 'edit'].includes(part))
-			|| i === pathParts.length - 1 // The last part is a known 404
-		) {
-			breadcrumbs.push(getStrikeThrough(part));
-		} else {
 			const pathname = '/' + pathParts.slice(0, i + 1).join('/');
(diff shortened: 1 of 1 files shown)


source/features/useful-not-found-page.tsx
```diff
@@ -71,29 +71,37 @@ async function getUrlToFileOnDefaultBranch(): Promise<string | void> {
...
 		}
 
-		if (i === parts.length - 1) {
+		if (i === pathParts.length - 1) {
 			// The last part of the URL is a known 404
-			bar.append(' / ', getStrikeThrough(part));
+			breadcrumbs.push(getStrikeThrough(part));
 		} else {
-			const pathname = '/' + parts.slice(0, i + 1).join('/');
-			bar.append(i ? ' / ' : '', <a href={pathname}>{part}</a>);
```
fregante, line 88: I had to refactor this because `i` was 0 but `0` was skipped, so the breadcrumbs would start with `/`, which we don't want

kidonng:
I would still show `orgs`, but neither linkified nor deleted. People _could_ have a repo called "teams" and we need to differentiate with that. Also there are other pages under `/orgs`.

There's still potentially improvement here. The `teams` part appears to work but in fact it just redirects and says "You’re not a member of any teams in this organization." I wonder if we should include any link that gives a redirection.

fregante:
Good point, it makes me wonder if we should also *not* drop the other parts of the URL like blob etc, but instead keep them dimmed

fregante:
> we should also _not_ drop the other parts of the URL

Done:


Screen Shot

Test: 

https://github.com/refined-github/refined-github/tree/main/nofile

source/features/useful-not-found-page.tsx
```diff
@@ -71,30 +71,30 @@ async function getUrlToFileOnDefaultBranch(): Promise<string | void> {
...
-	}
+	const pathParts = parseCurrentURL();
+	const breadcrumbs = [...pathParts.entries()]
+		.reverse() // Checks the anchors right to left
+		.map(([i, part]) => {
+			if (
+				// Exclude parts that don't exist as standalones
+				(i === 0 && part === 'orgs') // #5483
+				|| (i === 2 && ['tree', 'blob', 'edit'].includes(part))
+				|| i === pathParts.length - 1 // The last part is a known 404
```
fregante, line 82: All the excluded parts are still shown, but are immediately crossed and not checked for 404s. Do you know of any more parts that need to be excluded?

kidonng:
> > we should also _not_ drop the other parts of the URL
> 
> Done:
> 
> Screen Shot

Nice, we can skip those meaningless parts we are sure e.g. orgs, tree, blob.

Also, could we try to should each part unstyled first, then linkify/delete them based on the response? Theoretically they can lead to _another_ 404 if you try hard enough.

source/features/useful-not-found-page.tsx
kidonng: `blame`, `raw`? I think this part is useless most of the time, except for `commits`.

PR #5497, opened 2022-03-13
Exclude disabled features from "Find feature" wizard

fregante:
- Closes https://github.com/refined-github/refined-github/issues/5442
cc @karfau

## Before

Screen Shot 12

## After

However… still 9 steps required 🙂 maybe this wasn't worth it after all

Screen Shot 11

### DIFF
--- a/source/options.tsx
+++ b/source/options.tsx
@@ -131,7 +131,10 @@ function buildFeatureCheckbox({id, description, screenshot}: FeatureMeta): HTMLE
 }
 
 async function findFeatureHandler(event: Event): Promise<void> {
-	await cache.set<FeatureID[]>('bisect', importedFeatures, {minutes: 5});
+	// TODO: Add support for GHE
+	const options = await perDomainOptions.getOptionsForOrigin().getAll();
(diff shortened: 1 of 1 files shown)


cheap-glitch:
> However… still 9 steps required maybe this wasn't worth it after all

True, I guess you'd need at least half of the features disabled to reduce it to 8 steps :sweat_smile: 

But that was not the main reason for this change anyway so that's fine.

karfau:
> maybe this wasn't worth it after all

Yes, even before the change it was 8 steps for me.
But the main point was to not be surprised by checking features that were not enabled in the first place.

Or put differently: this limits the feature finder to those that are enabled.

Or did you implement some mechanism to search in all features?

(Still not able to understand the code/changes.)

fregante:
> Still not able to understand the code/changes

I changed it from "start checking all the features" to "start checking from the enabled features"

karfau:
> I changed it from "start checking all the features" to "start checking from the enabled features"

Do I understand correctly: If the feature/expected behavior is not "found"/confirmed in the first step(s) it will expand the search to all features?

fregante:
No, if the feature was disabled in the first place, then you know already what it is

PR #5499, opened 2022-03-13
Fix repo tab menu overflow due to `releases-tab` and `bugs-tab`

fregante:
- Fixes https://github.com/refined-github/refined-github/issues/4254
- Replaces and closes https://github.com/refined-github/refined-github/pull/5333

## Repro

1. Resize the window to have it just barely hide the Releases tab2. 
2. Empty the cache (makes the `bugs-tab` appear later)3. 
3. Refresh the page

## Demo

Shows that triggering a fake `resize` event makes the tab disappear

https://user-images.githubusercontent.com/1402241/158060307-b299c3ce-9bba-4150-b242-5fe6517d054b.mov

### DIFF
--- a/source/features/bugs-tab.tsx
+++ b/source/features/bugs-tab.tsx
@@ -124,6 +124,9 @@ async function addBugsTab(): Promise<void | false> {
 		issuesTab.after(bugsTab);
 	}
 
+	// Trigger a reflow to push the right-most tab into the overflow dropdown
+	window.dispatchEvent(new Event('resize'));
+
 	// Update bugs count
 	try {
 		const bugCount = await countPromise;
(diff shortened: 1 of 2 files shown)


source/features/releases-tab.tsx
```diff
@@ -76,6 +76,9 @@ async function addReleasesTab(): Promise<false | void> {
 	// This re-triggers the overflow listener forcing it to also hide this tab if necessary #3347
 	repoNavigationBar.replaceWith(repoNavigationBar);
```
fregante, line 77: I haven't verified whether this works or is necessary, but it's a tiny operation so let's leave it

source/features/bugs-tab.tsx
```diff
@@ -124,6 +124,9 @@ async function addBugsTab(): Promise<void | false> {
 		issuesTab.after(bugsTab);
```
fregante, line 124: @kidonng can we drop this check? #4166 is old now

fregante:
Maybe I can also include https://github.com/refined-github/refined-github/issues/4008#issuecomment-828173667

source/features/releases-tab.tsx
cheap-glitch: We can drop it, afaict it has no effect on the reflow whatsoever

PR #5503, opened 2022-03-14
Update all dependencies + lint some

fregante:


.xo-config.json
```diff
@@ -10,6 +10,7 @@
 	"rules": {
 		"no-alert": "off",
 		"no-warning-comments": "off",
+		"unicorn/prefer-dom-node-dataset": "off",
```
fregante, line 13: This rule breaks greppability because selectors in JS and CSS can't follow it. Change my view.

I still let it autofix our code where the attribute was not referenced anywhere else.

package.json
```diff
@@ -68,11 +68,11 @@
 		"selector-observer": "^2.1.6",
 		"shorten-repo-url": "^2.3.0",
 		"strip-indent": "^4.0.0",
-		"text-field-edit": "^3.1.1",
-		"tiny-version-compare": "^3.0.1",
+		"text-field-edit": "^3.1.9001",
+		"tiny-version-compare": "^4.0.0",
 		"twas": "^2.1.2",
 		"webext-base-css": "^1.3.2",
```
fregante, line 74: I'm leaving this out for #5478

rakleed:
@fregante why didn't you update `axios`? It has a dependency `follow-redirects` which has vulnerabilities in version 1.14.6: https://github.com/advisories/GHSA-74fj-2j2h-c42q, https://github.com/advisories/GHSA-pw2r-vq6v-hr8c

PR #5462, opened 2022-03-04
Fix `latest-tag-button` at repo home on tagged default branch

cheap-glitch:
Fixes #5446 

## Test URLs

- [Latest tag](https://github.com/refined-github/refined-github/tree/22.3.3)
- [Default branch, tagged](https://github.com/kas-elvirov/gloc)
- [Default branch, not tagged](https://github.com/refined-github/github-url-detection)
- [Other branches](https://github.com/refined-github/refined-github/tree/hotfix)

## Screenshot

**Before**
before

**After**
after

PR #5505, opened 2022-03-15
Fix `latest-tag-button` on branch ahead by more than 20 commits

### LINKED ISSUE #5504: `latest-tag-button` always shows "you're on the latest version"
### Example URLs

### Extension version

latest on main today

### Browser(s) used

Chrome 99, Safari 15.4

### Description

@cheap-glitch @kidonng  definitely not working for me even in Chrome with a recent build and clean cache:

Screen Shot 1

Hopefully this is fixed before the next release or I'll have to revert:

- https://github.com/refined-github/refined-github/pull/5462
- https://github.com/refined-github/refined-github/pull/5365

### Include in this issue:

- [ ] Console errors, if any

cheap-glitch:
Fixes #5504 
Follow-up of #5462

## Test URLs

- [Latest tag](https://github.com/refined-github/refined-github/tree/22.3.3)
- [Default branch, tagged](https://github.com/kas-elvirov/gloc)
- [Default branch, not tagged](https://github.com/refined-github/github-url-detection)
- [Default branch, not tagged (20+ commits ahead)](https://github.com/refined-github/refined-github)
- [Other branches](https://github.com/refined-github/refined-github/tree/hotfix)

## Screenshot

image

### DIFF
--- a/source/features/latest-tag-button.tsx
+++ b/source/features/latest-tag-button.tsx
@@ -88,7 +88,7 @@ const getRepoPublishState = cache.function(async (): Promise<RepoPublishState> =
 });
 
 async function init(): Promise<false | void> {
-	const {latestTag, aheadBy = 0} = await getRepoPublishState();
+	const {latestTag, aheadBy} = await getRepoPublishState();
 	if (!latestTag) {
 		return false;
 	}
@@ -118,7 +118,7 @@ async function init(): Promise<false | void> {
 	const defaultBranch = await getDefaultBranch();
 	const onLatestTag = currentBranch === latestTag;
 	const onDefaultBranch = !currentBranch || currentBranch === defaultBranch; // `getCurrentCommittish` returns `undefined` when at the repo root on the default branch #5446
-	const isAhead = aheadBy > 0;
+	const isAhead = aheadBy === undefined || aheadBy > 0;
 
 	if (onLatestTag || (onDefaultBranch && !isAhead)) {
 		link.setAttribute('aria-label', 'You’re on the latest version');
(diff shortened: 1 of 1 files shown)


source/features/latest-tag-button.tsx
```diff
@@ -131,7 +131,7 @@ async function init(): Promise<false | void> {
 		link.setAttribute(
 			'aria-label',
 			isAhead
-				? `${defaultBranch} is ${pluralize(aheadBy, '1 commit', '$$ commits')} ahead of the latest version`
+				? `${defaultBranch} is ${aheadBy === undefined ? 'more than 20 commits' : pluralize(aheadBy, '1 commit', '$$ commits')} ahead of the latest version`
```
cheap-glitch, line 134: Should we also put adapt what's displayed in that case after the `+` in the tag button? Currently there's no number at all, which might be confusing.

source/features/latest-tag-button.tsx
```diff
@@ -88,7 +88,7 @@ const getRepoPublishState = cache.function(async (): Promise<RepoPublishState> =
 });
 
 async function init(): Promise<false | void> {
-	const {latestTag, aheadBy = 0} = await getRepoPublishState();
+	const {latestTag, aheadBy} = await getRepoPublishState();
```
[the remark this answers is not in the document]
fregante: We need to document this or make it more explicit in `getRepoPublishState`. How about returning `Number.MAX_SAFE_INTEGER` or any other **serializable** value?

Maybe save it as something like:

```js
const undeterminableAheadBy = Number.MAX_SAFE_INTEGER
```

source/features/latest-tag-button.tsx
fregante: How about an asterisk? Or <sup>`+?`</sup> or <sup>`+🤷‍♂️`</sup>

cheap-glitch: Let's go with the asterisk. We could also use `++` instead if it's not visible enough.

fregante (approved):

fregante (approved):
🤩

fregante (approved):
🎂

PR #5506, opened 2022-03-15
GitHub bug: Avoid overflow caused by the diff signs

fregante:
- Fixes https://github.com/refined-github/refined-github/issues/5468


fregante:
> Sounds like our GitHub bugs CSS file will receive a +6 diff then 😌

It's a +5 diff after all!

cheap-glitch (approved):

PR #5507, opened 2022-03-15
Update `follow-redirects` dependency

rakleed:
Per https://github.com/refined-github/refined-github/pull/5503#issuecomment-1067622461


fregante:
We don't use it, it's just part of `size-plugin` and it does not use authentication. We can update it on the next round

PR #5518, opened 2022-03-17
Add `has()` selector helper

### LINKED ISSUE #4456: Replace `select().closest()` calls with `sizzle`
The `:has()` CSS selector is nowhere to be seen yet and given the recent de-class-ification of GitHub’s DOM we found ourself needing code like this very often:

```js
select('.some-octicon').closest('details')
```

<details>
<summary>"Custom API" suggestion, discarded. Sizzle is best</summary>

so I wonder how we can "polyfill" the selector without have to load Sizzle. Example:

```js
select2('details').has('.some-octicon')
```

Example:

```ts
function select2(selector: string) {
	return {
		has(children: string) {
			return select(children)?.closest(selector);
		}
	}
}

select2.all = (selector: string) => {
	return {
		has(children: string) {
			return select.all(children) :filter/map: .closest(selector);
		}
	}
}
```

Thoughts? Worth it at all?

</details>

fregante:
- Closes #4456 
- Related: https://github.com/g-plane/typed-query-selector/issues/21

4 fun

It can be used to replace:

```diff
- select('child').closest('parent');
+ selectHas('parent:has(child)')
```

```diff
- select('child').closest('parent').querySelector('family');
+ selectHas('parent:has(child) family')
```

but not:

```diff
existingElement.closest('parent')
```

Since this isn't native, it can't be used:

- in selector-observer
- in element-ready

as well as:

- [ ] no select.all yet, which would be useful in `deep-reblame` and `dim-bots` for example
- [x] no types yet https://github.com/g-plane/typed-query-selector/issues/21

Side note:

- [x] _Fixed:_ it's probably less efficient because instead of using `select + closest` it uses `select.all + select multipletimes` 😅 but how bad can it be? Maybe I can fix this though
- [ ] Untested

### DIFF
--- a/source/features/action-used-by-link.tsx
+++ b/source/features/action-used-by-link.tsx
@@ -4,7 +4,7 @@ import {SearchIcon} from '@primer/octicons-react';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
-import selectHas from '../helpers/selectHas';
+import selectHas from '../helpers/select-has';
 
 function init(): void {
 	const actionRepo = selectHas<HTMLAnchorElement>('a:has(aside .octicon-repo)')!
--- a/source/features/action-used-by-link.tsx
+++ b/source/features/action-used-by-link.tsx
@@ -7,7 +7,7 @@ import features from '.';
 import selectHas from '../helpers/select-has';
 
 function init(): void {
-	const actionRepo = selectHas<HTMLAnchorElement>('a:has(aside .octicon-repo)')!
+	const actionRepo = selectHas('a:has(aside .octicon-repo)')!
 		.pathname
 		.slice(1);
 
--- a/source/features/action-used-by-link.tsx
+++ b/source/features/action-used-by-link.tsx
@@ -4,10 +4,10 @@ import {SearchIcon} from '@primer/octicons-react';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
+import selectHas from '../helpers/selectHas';
 
 function init(): void {
-	const actionRepo = select('aside .octicon-repo')!
-		.closest('a')!
+	const actionRepo = selectHas<HTMLAnchorElement>('a:has(aside .octicon-repo)')!
 		.pathname
 		.slice(1);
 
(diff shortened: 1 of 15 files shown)


fregante:
Tested.

Almost every replacement I made was incorrect: `x.closest(div)` is not equivalent to `div:has(x)`

- If `x` has many `div` parents, the selector will pick the "farthest" instead of the "closest"
- If `x` is a composite selector like `.a .b`, the final selector could look like `.a div:has(.b)` instead of `div:has(.a .b)`

I'll merge this PR to enable more usage in the future. Also ideally we'd add `select.all` support because that's the most useful/common one.

source/features/action-used-by-link.tsx
```diff
@@ -4,10 +4,10 @@ import {SearchIcon} from '@primer/octicons-react';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
+import selectHas from '../helpers/select-has';
 
 function init(): void {
-	const actionRepo = select('aside .octicon-repo')!
-		.closest('a')!
+	const actionRepo = selectHas('a:has(aside .octicon-repo)')!
```
fregante, line 10: These selectors are real, valid selectors and can be tested natively in Safari. In the future we'll be able to just replace `selectHas` with `select`

fregante:
Looking forward to seeing more uses in our features!

source/features/action-used-by-link.tsx
kidonng: I'm tempted to `#enable-experimental-web-platform-features` now

Or maybe install Chrome Dev which just added `:has()` support in 105. But it is scheduled to release on Aug 30 according to [Chromium Dash](https://chromiumdash.appspot.com/schedule), so I'll just wait for two months 😆

fregante: Firefox will delay this a while longer. No work in progress at all. https://bugzilla.mozilla.org/show_bug.cgi?id=418039

I think we can’t try the native implementation until we’re ready to ship or else we need to test each selector for both native and polyfilled version. Keep in mind that the polyfill is very limited so we can’t trust it to match the native implementation yet.

PR #5519, opened 2022-03-17
Meta: Fix changelog creation

fregante:
- Closes https://github.com/refined-github/refined-github/issues/5516

Docs: https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes#configuration-options

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -10,6 +10,9 @@ changelog:
     - title: Changes
       labels:
         - "*"
+      exclude:
+        - bug
+        - meta
     - title: Bugfixes
       labels:
         - bug
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -11,8 +11,9 @@ changelog:
       labels:
         - "*"
       exclude:
-        - bug
-        - meta
+        labels:
+          - bug
+          - meta
     - title: Bugfixes
       labels:
         - bug


.github/release.yml
```diff
@@ -10,6 +10,10 @@ changelog:
     - title: Changes
       labels:
         - "*"
+      exclude:
+        labels:
+          - bug
+          - meta
```
FloEdelmann, line 16: Doesn't `enhancement` need to be excluded here, too?

cheap-glitch: No, it should be "captured" by the "New" category before (cf. the [example config](https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes#example-configuration)).

FloEdelmann: Ah, so `bug` is not "captured" because the "Bugfixes" category comes after the "Changes" category?

And `meta` needs to be repeated because a category's `excludes` overrides the global `excludes`?

fregante: `meta` is here because we don't want to include such changes

fregante: Fixed in https://github.com/refined-github/refined-github/pull/5522, thanks for the comment

PR #5521, opened 2022-03-18
Reliability fixes for `fork-source-link-same-view` /2

fregante:
Due to https://github.com/refined-github/refined-github/pull/5494#discussion_r829019629

cc @kidonng

### DIFF
--- a/source/features/fork-source-link-same-view.tsx
+++ b/source/features/fork-source-link-same-view.tsx
@@ -15,7 +15,8 @@ async function getEquivalentURL(): Promise<string> {
 	const forkedRepository = getRepo(getForkedRepo())!;
 	const defaultUrl = '/' + forkedRepository.nameWithOwner;
 
-	if (pageDetect.isConversation() || pageDetect.isRepoRoot()) {
+	// Do not use `isConversation` https://github.com/refined-github/refined-github/pull/5494#discussion_r829019629
+	if (pageDetect.isIssue() || pageDetect.isPR() || pageDetect.isRepoRoot()) {
 		// We must reset the link because the header is outside the ajaxed area
 		return defaultUrl;
 	}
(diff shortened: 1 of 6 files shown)


kidonng:
Thanks for the PR! I have prepared this in a separate branch (beec305498ef1b70c9c99f76817d4e7973b8eb5f) but this is also fine.

source/features/fork-source-link-same-view.tsx
```diff
@@ -37,7 +38,7 @@ async function getEquivalentURL(): Promise<string> {
 
 async function init(): Promise<void> {
 	// The link must always be updated/reset. This pattern ensures that the link is always updated and never fails through some conditions.
-	select<HTMLAnchorElement>(`[data-hovercard-url="/${getForkedRepo()!}/hovercard"]`)!.href = await getEquivalentURL();
+	select(`a[data-hovercard-url="/${getForkedRepo()!}/hovercard"]`)!.href = await getEquivalentURL();
```
fregante, line 41: Huh they added support for concatenated strings? 😮 This change can probably be done to more places

kidonng: Hmm I don't know of this is only supported recently, I just thought this looks better.

yakov116: I think we did it where we were able to already

fregante: There are more places. For some reason sometimes it works, others it doesn't 🤷‍♂️ 

```js
❯ rg select\<
source/options.tsx
212:                    const input = select<HTMLInputElement>('#' + feature)!;

source/features/suggest-commit-title-limit.tsx
15:     const inputField = select<HTMLInputElement>(fieldSelector)!;

source/features/highest-rated-comment.tsx
70:     const {hash} = select<HTMLAnchorElement>('a.js-timestamp', bestComment)!; // TODO: Drop generic once TypeScript sorts it out, it shouldn't be needed
```

PR #5523, opened 2022-03-18
Deinit event listeners in features using `AbortSignal`

cheap-glitch:
Follow-up #5422 

Let's do https://github.com/fregante/delegate-it/issues/26 while we're at it :fire:

### DIFF
--- a/source/features/prevent-pr-merge-panel-opening.tsx
+++ b/source/features/prevent-pr-merge-panel-opening.tsx
@@ -12,12 +12,8 @@ async function sessionResumeHandler(): Promise<void> {
 	}
 }
 
-function init(): Deinit {
-	document.addEventListener('session:resume', sessionResumeHandler);
-
-	return () => {
-		document.removeEventListener('session:resume', sessionResumeHandler);
-	};
+function init(signal: AbortSignal): void {
+	document.addEventListener('session:resume', sessionResumeHandler, {signal});
 }
 
 void features.add(import.meta.url, {
--- a/source/features/sync-pr-commit-title.tsx
+++ b/source/features/sync-pr-commit-title.tsx
@@ -6,7 +6,6 @@ import * as pageDetect from 'github-url-detection';
 import * as textFieldEdit from 'text-field-edit';
 
 import features from '.';
-import getDeinitHandler from '../helpers/get-deinit-handler';
 import onPrMergePanelOpen from '../github-events/on-pr-merge-panel-open';
 import {getConversationNumber} from '../github-helpers';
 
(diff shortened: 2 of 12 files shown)


source/features/sync-pr-commit-title.tsx
```diff
@@ -84,11 +83,11 @@ function disableSubmission(): void {
...
 
-function init(): Deinit {
-	listeners.push(
-		...onPrMergePanelOpen(updateCommitTitle),
+function init(signal: AbortSignal): Deinit {
+	subscriptions.push(
+		onPrMergePanelOpen(updateCommitTitle, signal),
 		delegate(document, '#merge_title_field', 'input', updateUI),
 		delegate(document, 'form.js-merge-pull-request', 'submit', updatePRTitle),
 		delegate(document, '.rgh-sync-pr-commit-title', 'click', disableSubmission),
```
cheap-glitch, line 93: Worth it to pass the signal to `disableSubmission` so we can replace the custom deinit with a simple `signal.abort()`?

source/features/prevent-pr-merge-panel-opening.tsx
```diff
@@ -12,12 +12,8 @@ async function sessionResumeHandler(): Promise<void> {
 	}
```
cheap-glitch, line 12: Same here, we could pass the signal to `sessionResumeHandler`

source/features/sync-pr-commit-title.tsx
fregante: Up to you. The alternative could also be to use `subscriptions.push(signal)` with the global `signal` so we'd still have the custom deinit but it would be simpler

source/features/sync-pr-commit-title.tsx
```diff
@@ -84,11 +83,11 @@ function disableSubmission(): void {
...
 
-const listeners: Deinit[] = [];
+const subscriptions: delegate.Subscription[] = [];
 
-function init(): Deinit {
-	listeners.push(
-		...onPrMergePanelOpen(updateCommitTitle),
+function init(signal: AbortSignal): Deinit {
+	subscriptions.push(
+		onPrMergePanelOpen(updateCommitTitle, signal),
```
fregante, line 90: This can't work because the signal is managed by the global manager but this array is local and we must be able to deinit it here.

fregante: We have 2 quick solutions here:

- drop the changes made to `onPrMergePanelOpen` and to this file
- create a local abortcontroller so that `onPrMergePanelOpen` can keep its improvement

For the latter, probably:

```js
function init() {
	// Local controller needed because the feature can be deinited locally
	const controller = new AbortControler()
	
	subscriptions.push(
		onPrMergePanelOpen(updateCommitTitle, signal),
		() => controller.abort(),
		etc
```

Other, more involved solutions:

- _only_ use a local controller and have `deinit` call that. `init` would return `controller.abort`
- add an API that allows a feature to deinit itself on demand, like:
	```js
	function deinit() {
		features.deinit(import.meta.url)
	}
	```
	This would free us from having to deal with the local `listeners/subscriptions` arrays

cheap-glitch:
Let's merge this when it passes review, we can include https://github.com/fregante/delegate-it/issues/26 later when it's done.

source/features/index.tsx
```diff
@@ -144,6 +143,26 @@ const globalReady: Promise<RGHOptions> = new Promise(async resolve => {
...
+
+	if ('destroy' in deinit) { // Delegate subscription
+		return deinit.destroy;
+	}
+
+	return deinit;
+}
+
 function setupDeinit(deinit: Deinit): void {
 	document.addEventListener('pjax:start', getDeinitHandler(deinit), {once: true});
```
fregante, line 167: `setupPageLoad` already attaches an ajaxed event listener, so we should change this function to something like:

```ts
 function setupDeinit(signal, deinit: Deinit): void {
   	signal.addEventListener('abort', getDeinitHandler(deinit));

PR #5527, opened 2022-03-20
Adds `C-bug` to supported labels

Milo123459:
## Test URLs

https://github.com/rust-lang/rust

### DIFF
--- a/source/features/bugs-tab.tsx
+++ b/source/features/bugs-tab.tsx
@@ -12,7 +12,7 @@ import SearchQuery from '../github-helpers/search-query';
 import abbreviateNumber from '../helpers/abbreviate-number';
 import {highlightTab, unhighlightTab} from '../helpers/dom-utils';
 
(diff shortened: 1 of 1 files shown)


busches:
At this point should we just make it a regex where anything `-bug` would match?

Also, are there other projects that use this label? Or is C short for confirmed or C the language?

Milo123459:
Rust-analyser uses it, along with every repo under the rust-lang org.

fregante (changes requested):
We're definitely not adding niche labels or else everyone will want its own. 

We didn't initially add just -bug because there might be not-a-bug labels

yakov116:
I think we support enough edge cases. This is sort of custom already.

kidonng:
> We didn't initially add just -bug because there might be not-a-bug labels

I think this should be the other around: we probably want to include every label that matches "bug" _then_ exclude false positives (e.g. `not-a-bug`).

Milo123459:
This label is used a lot, but I understand where you are coming from.

fregante:
@kidonng can do

PR #5542, opened 2022-03-22
Drop `dim-visited-conversations` feature

fregante:
On repos you visit often, this just leads to links having lower contrast.

- See https://github.com/refined-github/refined-github/pull/1206#issuecomment-386051034

Being 3 lines of CSS, it can be added to one’s own CSS field

## Before

Screen Shot 14


cheap-glitch (approved):

kidonng:
Probably by far my biggest mistake, should have not enforced this feature added in its  weakest form. Though I still want to see the original proposal reconsidered.

Issue #5222, opened 2021-12-17
🕊 Minor codebase updates and fixes


fregante:
Continues from https://github.com/refined-github/refined-github/pull/5213

This issue is not for generic bug reports but to suggest changes/refactoring to the extension itself.

fregante:
- [ ] Automatically hide resolved comments on this issue and [#4008](https://github.com/refined-github/refined-github/issues/4008)

Both URLs can be hard-coded in the feature

<img width="644" alt="Screen Shot" src="https://user-images.githubusercontent.com/1402241/146532721-f6a52833-09a1-4adb-979a-dda1d4c9c5f2.png">

fregante:
Moved from https://github.com/refined-github/refined-github/pull/5213#pullrequestreview-833833784 https://github.com/refined-github/refined-github/issues/4008#issuecomment-812885528

- [ ] Maybe rewrite most `selector-observer` usage to use `deinit` instead of `onetime` as described in https://github.com/sindresorhus/refined-github/pull/4072#issuecomment-792402458
- [ ] add `deinit` to more features. Most `delegate` calls should result in a `deinit` on unload
- [ ] Make `init` return a specific type called `Deinit`, like:
   `type Deinit = VoidFunction | {abort: VoidFunction} | AbortController, etc` and then set the return value to `Deinit | Deinit[]`
   this will let us return values easily without having to locate the abort function of each
- [ ] Create a new `abortSignal` at every load and pass it to each `init`. This might be a little difficult to pass around at times, but it would be great to stop any async activity when unloading. For example addEventListener natively supports this signal so those features would not have to return anything:
	```js
	function init(signal: AbortSignal): void {
		document.body.addEventListener('click', doStuff, {signal})
	}
	```
	This case isn't better though:
	```js
	function init(signal: AbortSignal): void {
		const listener = delegate(..etc);
		signal.addEventListener('abort', listener.disconnect); 
	}
	```
	so it should just return `listener.disconnect` instead, as it's already possible
	```js
	function init(): Deinit {
		return delegate(..etc).disconnect;
	}
	```

fregante:
- [ ] Automatically hide resolved comments on this issue and [#4008](https://github.com/refined-github/refined-github/issues/4008)

Both URLs can be hard-coded in the feature

<img width="644" alt="Screen Shot" src="https://user-images.githubusercontent.com/1402241/146532721-f6a52833-09a1-4adb-979a-dda1d4c9c5f2.png">

[...]

PR #5544, opened 2022-03-23
Prevent `dim-bots` error on empty milestone list

kidonng:
Without this I get:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'classList')
    at refined-github.js:8087:78
```

## Test URLs

https://github.com/fish-shell/fish-shell/milestone/30


source/features/dim-bots.tsx
```diff
@@ -41,7 +41,14 @@ function init(): void {
 
 	// Delay collapsing, but only after they're collapsed on load #5158
 	requestAnimationFrame(() => {
-		select('#repo-content-pjax-container .js-navigation-container')!.classList.add('rgh-dim-bots--after-hover');
```
fregante, line 44: Why not just `?.`?
```js
		select('#repo-content-pjax-container .js-navigation-container')?.classList.add('rgh-dim-bots--after-hover');
Write
```

kidonng: I remember somewhere you said "it just silences error".

cheap-glitch: > I remember somewhere you said "it just silences error".

But that's also what that change does, just in a more verbose way. It should probably be
```ts
if (pageDetect.isMilestoneList()) {
	return;
}

requestAnimationFrame(…)
```

kidonng: > ```ts
> if (pageDetect.isMilestoneList()) {
> 	return;
> }
> ```

This disables the feature on `isMilestoneList` completely. However the error only appears on empty lists.

yakov116: It should be if !container && pageDetect....

cheap-glitch: > However the error only appears on empty lists.

Then how about adding `() => select.exists('.blankslate')` to the feature's `exclude`?

kidonng: > I remember somewhere you said "it just silences error".

[I found that quote](https://github.com//refined-github/refined-github/issues/5300#issuecomment-1013645436), though not by @fregante. Sorry!

cheap-glitch: > [I found that quote](https://github.com//refined-github/refined-github/issues/5300#issuecomment-1013645436), though not by fregante. Sorry!

To be fair, I was only repeating what I learned from similar reviews ;)

> Then how about adding () => select.exists('.blankslate') to the feature's exclude?

Might as well do https://github.com/refined-github/github-url-detection/issues/100. @kidonng do you have time for it? Otherwise I can take it

fregante: Re: "it just silences the error"

If we *expect* an element to exist, the we must not use `?.` or a conditional check. 

If there situations in which the element can be missing, then we *expect* that, so if/?. are the right solution. 

However, sometimes there are better solutions, like specifically handling or excluding the pages that don't have the item. Blank state is one such page and in #5222 (or the previous one) I suggested a global solution to that.

cheap-glitch: > I suggested a global solution to that

It's here: https://github.com/refined-github/refined-github/issues/4008#issuecomment-822942520

But how can we tell in `runFeature` which errors come from missing elements and which come from broken or outdated selectors?

fregante: What I suggested there was this logic:

- when an error is thrown, if the page is blank or the repo is archived, ignore the error (or log it as "Page is blank/archived, feature 124 will not work"

fregante (approved):

PR #5548, opened 2022-03-25
fix: make `hide-watch-and-fork-count` feature disableable

kiprasmel:
really don't feel like [pasting in the custom css](https://github.com/refined-github/refined-github/pull/5197#issuecomment-993390710) in each of my accounts in different browsers & laptops..
would really appreciate if we could keep this feature disableable.

partly reverts pr: https://github.com/refined-github/refined-github/pull/5197
or commit: 7551cccbd72ba5fdca3a4a5635c31df488b451ad

### DIFF
--- a/source/features/hide-watch-and-fork-count.css
+++ b/source/features/hide-watch-and-fork-count.css
@@ -1,3 +1,6 @@
-.pagehead-actions :is(#repo-network-counter, #repo-notifications-counter) {
+.rgh-hide-watch-and-fork-count .pagehead-actions :is(
+#repo-network-counter,
+#repo-notifications-counter
+) {
 	display: none;
 }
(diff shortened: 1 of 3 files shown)


source/features/hide-watch-and-fork-count.css
```diff
@@ -1,3 +1,6 @@
-.pagehead-actions :is(#repo-network-counter, #repo-notifications-counter) {
+.rgh-hide-watch-and-fork-count .pagehead-actions :is(
```
kiprasmel, line 1: i don't know if the `.pagehead-actions` is needed here. it wasn't there before the [PR](https://github.com/refined-github/refined-github/pull/5197) that both fixed the issue because of github's redesign, and made the feature non-disableable.

fregante:
This is the easiest feature to undo. It's 3 lines of CSS. Plus the settings already sync across instances of the same browser/account.

kiprasmel:
ah, okay.

tried the custom css snippet, fair enough. just needed to an an `!important` for it to work.

```css
.pagehead-actions :is(#repo-network-counter, #repo-notifications-counter) {
	display: inline-block !important;
}
```

fregante:
Since #6333 (version 23.2.11) you can disable the feature in the options page.
It's now called `clean-repo-header`

PR #5553, opened 2022-03-29
Add filters for deployment requests and failed runs in `select-notifications`

ThiagoSilvaSilveira:
Added a filter for GH Actions deployment notifications.
- New type `Deployments` added.
- New status `Run failed` added.

Deployment review requests can be filtered with type `Deployments` and status `Open`.
Failed deployments can be filtered with type `Deployments` and status `Run failed`.

## Test URLs
[github.com/notifications](https://github.com/notifications)

## Screenshot
image

image

### DIFF
--- a/source/features/select-notifications.tsx
+++ b/source/features/select-notifications.tsx
@@ -14,7 +14,9 @@ import {
 	GitPullRequestDraftIcon,
 	GitPullRequestIcon,
 	IssueOpenedIcon,
+	RocketIcon,
 	XCircleIcon,
+	XIcon,
 } from '@primer/octicons-react';
 
 import features from '.';
@@ -22,10 +24,12 @@ import features from '.';
 const filters = {
 	'Pull requests': ':is(.octicon-git-pull-request, .octicon-git-pull-request-closed, .octicon-git-pull-request-draft, .octicon-git-merge)',
 	Issues: ':is(.octicon-issue-opened, .octicon-issue-closed)',
-	Open: ':is(.octicon-issue-opened, .octicon-git-pull-request)',
+	Deployments: ':is(.octicon-rocket, .octicon-x)',
+	Open: ':is(.octicon-issue-opened, .octicon-git-pull-request, .octicon-rocket)',
 	Closed: ':is(.octicon-issue-closed, .octicon-git-pull-request-closed)',
 	Draft: '.octicon-git-pull-request-draft',
 	Merged: '.octicon-git-merge',
+	'Run failed': '.octicon-x',
 	Read: '.notification-read',
 	Unread: '.notification-unread',
 };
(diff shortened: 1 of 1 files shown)


source/features/select-notifications.tsx
```diff
@@ -14,18 +14,22 @@ import {
...
+	XIcon,
 } from '@primer/octicons-react';
 
 import features from '.';
 
 const filters = {
 	'Pull requests': ':is(.octicon-git-pull-request, .octicon-git-pull-request-closed, .octicon-git-pull-request-draft, .octicon-git-merge)',
 	Issues: ':is(.octicon-issue-opened, .octicon-issue-closed)',
-	Open: ':is(.octicon-issue-opened, .octicon-git-pull-request)',
+	Deployments: ':is(.octicon-rocket, .octicon-x)',
```
cheap-glitch, line 27: `.octicon-x` will match every notification for failed CI jobs, not just deployments, right?

While I like the "Run failed" filter, this one is probably too broad. Is there an easy way to limit it to "failed deployments" only?

source/features/select-notifications.tsx
```diff
@@ -14,18 +14,22 @@ import {
...
+	Open: ':is(.octicon-issue-opened, .octicon-git-pull-request, .octicon-rocket)',
 	Closed: ':is(.octicon-issue-closed, .octicon-git-pull-request-closed)',
 	Draft: '.octicon-git-pull-request-draft',
 	Merged: '.octicon-git-merge',
+	'Run failed': '.octicon-x',
```
cheap-glitch, line 32: Is "Run failed" the exact wording GitHub uses? If not, we should try to match it.

cheap-glitch (changes requested):

source/features/select-notifications.tsx
ThiagoSilvaSilveira: It seems like GH doesn't really distinguishes `Deployments` from any other `CI activity` triggered by a workflow in general as `.octicon-x` does indeed match every notification for failed CI. I am just trying to figure out if it will match any other type, but for any of my internal tests I could not find it.

I am going to propose, instead of trying to handle specifics of failed runs for `Deployments`, renaming it to `CI activity`. It is a somewhat broad filter, but do you think it loses its value?

fregante:
This seems to be a rather obscure feature to be shown here

fregante:
> > This seems to be a rather obscure feature to be shown here
> 
> Can you elaborate on that?

I'm hesitant to add more selectors to this dropdown unless they're common. There's a lot that can be added so if we're not _selective_ we'll end up with too many options and we have to rethink the whole feature.

I'm currently on -1 for this change

ThiagoSilvaSilveira:
Ok then. I am closing it for now.

PR #5560, opened 2022-04-03
Meta: Check out tag automatically in Safari script

sindresorhus:


fregante (approved):

fregante:
Hmmm in reality though this shouldn't be necessary.

The intention of the code was to only warn you in case code was committed after the tag, but you don't need to check out the actual tag. 

It should amount to: "is HEAD tagged?" Or "is HEAD the latest tag?"

sindresorhus:
> but you don't need to check out the actual tag.

But you still want me to release what's in the latest tag, not in `main`, right? So I don't really see a problem with the current logic.

fregante:
Sure, but the `git checkout` doesn't make any changes to the code, so it's meaningless/useless. Just making sure we're on the same page. This script is for exclusively for you so you can customize it as you please.

PR #5562, opened 2022-04-04
Show explicitly linked comment in `conversation-activity-filter`

### LINKED ISSUE #5390: Exclude explicitly linked comments in `conversation-activity-filter`
### Description

Like #5334, but for `conversation-activity-filter`. E.g. in https://github.com/refined-github/refined-github/issues/4008#issuecomment-996659891, I linked an [earlier comment](https://github.com/refined-github/refined-github/issues/4008#issuecomment-824729016) that is now hidden. While the filter is active, I have no idea what I linked :sweat_smile: 

With #5562, explicitly linked comments are excluded on page load, so clicking this link shows the comment: https://github.com/refined-github/refined-github/issues/4008#issuecomment-824729016
However, clicking an in-page link doesn't work.

### Reproduction

1. Go to https://github.com/refined-github/refined-github/issues/4008
2. Load hidden timeline items
  grafik
3. Append this hash `#issuecomment-824729016` to the URL

… or …

1. Go to https://github.com/refined-github/refined-github/issues/4008#issuecomment-996659891
2. Load hidden timeline items
  grafik
3. Scroll down to my comment and click this link there (it's currently hidden as well :slightly_frowning_face:):
  grafik

jopemachine:
Fixes #5390.

## Test URLs

https://github.com/refined-github/refined-github/issues/4008#issuecomment-996659891

### DIFF
--- a/source/features/conversation-activity-filter.tsx
+++ b/source/features/conversation-activity-filter.tsx
@@ -56,10 +56,10 @@ function processReview(review: HTMLElement): void {
 
 function processPage(): void {
 	for (const item of select.all(`.js-timeline-item:not(.${hiddenClassName}, .${collapsedClassName})`)) {
-    // Exclude explicitly linked comment
-    if (location.hash.startsWith('#issuecomment-') && select.exists(location.hash, item)) {
-      continue;
-    }
+		// Exclude deep-linked comment
+		if (location.hash.startsWith('#issuecomment-') && select.exists(location.hash, item)) {
+			continue;
+		}
 
 		if (select.exists('.js-comment[id^=pullrequestreview]', item)) {
 			processReview(item);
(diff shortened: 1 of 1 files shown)


fregante:
Thanks for the PR! I think this can be part of the for/of line, inside the :not selector, something like `:not(id="${location.hash}")`

jopemachine:
> Thanks for the PR! I think this can be part of the for/of line, inside the :not selector, something like `:not(id="${location.hash}")`

Thanks for your feedback.
The selector like `.js-timeline-item:not(#issuecomment-824729016):not(.rgh-conversation-activity-filtered, .rgh-conversation-activity-collapsed)` seems not working because the div with the id is child of selecting items.

For example at the bottom picture, the selector selects the blue line, not having the id.

So I think handling this with one selector seems not possible.

fregante:
Makes sense. Then I'd just move the condition higher, so it's the first thing inside the loop

fregante (approved):

cheap-glitch (approved):

cheap-glitch:
Thanks @jopemachine!

PR #5574, opened 2022-04-06
Meta: Add regression tests for `shorten-links`

cheap-glitch:
Follow-up of https://github.com/refined-github/refined-github/pull/5472#issuecomment-1065227616

This PR adds a `shortenLink` helper in `helpers/dom-formatters` and some regression tests.


fregante (approved):
💙 

If you feel like it, it'd be great to include a script to be able to keep these snippets up to date. Here are three ways:

1. Just fetch the page from the test and test it live (flaky and may break the tests suddenly)
2. Fetch the page the same way, but also include a static HTML. If the static HTML is different, use `console.warn` but keep the tests working. The warning would have to output the new snippet. The comparison can be done without whitespace. If the fetch fails, it shouldn't break the test.
3. Create an external downloader script like `wget page1.html page2.html etc` and then load these pages in `test/shorten-link.ts`

PR #5575, opened 2022-04-06
Fix clean-up of observers used by features

cheap-glitch:
Follow-up of https://github.com/refined-github/refined-github/pull/5422#discussion_r841821958
Fixes https://github.com/refined-github/refined-github/issues/5552 (hopefully)

### DIFF
--- a/source/helpers/get-deinit-handler.ts
+++ b/source/helpers/get-deinit-handler.ts
@@ -1,10 +1,14 @@
 export default function getDeinitHandler(deinit: Deinit): VoidFunction {
 	if (deinit instanceof MutationObserver || deinit instanceof ResizeObserver || deinit instanceof IntersectionObserver) {
-		return deinit.disconnect;
+		return () => {
+			deinit.disconnect();
+		};
 	}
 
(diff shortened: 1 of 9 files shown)


source/helpers/get-deinit-handler.ts
```diff
@@ -1,10 +1,14 @@
...
 	}
 
 	if ('abort' in deinit) { // Selector observer
-		return deinit.abort;
+		return () => {
+			deinit.abort();
+		};
 	}
 
 	if ('destroy' in deinit) { // Delegate subscription
```
cheap-glitch, line 14: @fregante Do you think `delegation.destroy` needs the same treatment?

fregante: Not required in either:

https://github.com/fregante/delegate-it/blob/ec81f4400a6bc1d48e75705a6a8089f6cf6207c8/index.ts#L165-L170

https://github.com/sindresorhus/debounce-fn/blob/7234eb37aa74198dfca2ab49a889cb24555360dc/index.js#L69-L79

fregante (approved):

PR #5587, opened 2022-04-11
Allow `batch-mark-files-as-viewed` to mark all files from first

### LINKED ISSUE #5484: `batch-mark-files-as-viewed` shouldn't require a first click
### Description

https://github.com/refined-github/refined-github/blob/b9c951ae7e151164940662e178e83ce68e0c4553/source/features/batch-mark-files-as-viewed.tsx#L21

Instead of checking whether there was a `previousFile`, it should:

- use `previousFile?.isConnected` to check whether the box is still in the same document
- default to the first item

This is because I scroll down the page, then I remember I haven't clicked "Viewed" yet and have to go all the way up to start the batch \

### Screenshot

_No response_

### Example URL

https://github.com/refined-github/refined-github/pull/5472

jopemachine:
Fixes #5484.

## Test URLs

https://togithub.com/refined-github/refined-github/pull/5422/files

## Screenshot

Apr-11-2022 14-39-08

### DIFF
--- a/source/features/batch-mark-files-as-viewed.tsx
+++ b/source/features/batch-mark-files-as-viewed.tsx
@@ -1,7 +1,6 @@
 import select from 'select-dom';
 import delegate from 'delegate-it';
 import * as pageDetect from 'github-url-detection';
-import elementReady from 'element-ready';
 
 import features from '.';
 import clickAll from '../helpers/click-all';
@@ -19,6 +18,10 @@ function isChecked(file: HTMLElement): boolean {
 }
 
 function batchToggle(event: delegate.Event<MouseEvent, HTMLFormElement>): void {
+	if (!previousFile) {
+		previousFile = select('.js-file');
+	}
+
 	if (!event.shiftKey || !previousFile?.isConnected) {
 		return;
 	}
@@ -64,11 +67,6 @@ function onAltClick(event: delegate.Event<MouseEvent, HTMLInputElement>): void {
 }
 
 function init(): Deinit[] {
-	// eslint-disable-next-line @typescript-eslint/no-floating-promises
-	elementReady('.js-file').then(file => {
-		previousFile = file;
-	});
-
 	return [
 		// `mousedown` required to avoid mouse selection on shift-click
 		delegate(document, '.js-reviewed-toggle', 'mousedown', batchToggle),
--- a/source/features/batch-mark-files-as-viewed.tsx
+++ b/source/features/batch-mark-files-as-viewed.tsx
@@ -34,7 +34,7 @@ function batchToggle(event: delegate.Event<MouseEvent, HTMLFormElement>): void {
 	const isThisBeingFileChecked = !isChecked(thisFile); // Flip it because the value hasn't changed yet
 
 	runningBatch = true;
-	const selectedFiles = getItemsBetween(files, previousFile ?? files[0], thisFile);
+	const selectedFiles = getItemsBetween(files, previousFile, thisFile);
 	for (const file of selectedFiles) {
 		if (file !== thisFile && isChecked(file) !== isThisBeingFileChecked) {
 			select('.js-reviewed-checkbox', file)!.click();
--- a/source/features/batch-mark-files-as-viewed.tsx
+++ b/source/features/batch-mark-files-as-viewed.tsx
@@ -23,22 +23,16 @@ function batchToggle(event: delegate.Event<MouseEvent, HTMLFormElement>): void {
 		return;
 	}
 
-	previousFile ??= select('.js-file'); // #5484
-
 	event.preventDefault();
 	event.stopImmediatePropagation();
 
-	const thisFile = event.delegateTarget.closest('.js-file')!;
 	const files = select.all('.js-file');
+	const thisFile = event.delegateTarget.closest('.js-file')!;
+	const isThisFileChecked = isChecked(thisFile);
 
-	const selectedFiles = getItemsBetween<HTMLElement>(files, previousFile, thisFile);
-
-	if (files.indexOf(previousFile!) > files.indexOf(thisFile)) {
-		selectedFiles.reverse();
-	}
-
+	const selectedFiles = getItemsBetween(files, previousFile ?? files[0], thisFile);
 	for (const file of selectedFiles) {
-		if (isChecked(file) === isChecked(thisFile)) {
+		if (isChecked(file) === isThisFileChecked) {
 			select('.js-reviewed-checkbox', file)!.click();
 		}
 	}
--- a/source/features/batch-mark-files-as-viewed.tsx
+++ b/source/features/batch-mark-files-as-viewed.tsx
@@ -5,6 +5,7 @@ import * as pageDetect from 'github-url-detection';
 import features from '.';
 import clickAll from '../helpers/click-all';
 import showToast from '../github-helpers/toast';
+import getItemsBetween from '../helpers/get-items-between';
 
 let previousFile: HTMLElement | undefined;
 
@@ -36,14 +37,7 @@ function batchToggle(event: delegate.Event<MouseEvent, HTMLFormElement>): void {
 	const files = select.all('.js-file');
 	const previousFileState = isFirstToggle ? !isChecked(thisFile) : isChecked(previousFile);
 
-	const selections = [
-		files.indexOf(previousFile),
-		files.indexOf(thisFile),
-	].sort((a, b) => a - b);
-
-	selections[1] += 1;
-
-	const selectedFiles = files.slice(...selections);
+	const selectedFiles = getItemsBetween<HTMLElement>(files, previousFile, thisFile);
 
 	for (const file of selectedFiles) {
 		if (isChecked(file) !== previousFileState) {
--- a/source/features/batch-mark-files-as-viewed.tsx
+++ b/source/features/batch-mark-files-as-viewed.tsx
@@ -26,16 +26,13 @@ function batchToggle(event: delegate.Event<MouseEvent, HTMLFormElement>): void {
 	const isFirstToggle = previousFile === undefined;
 
 	previousFile ??= select('.js-file'); // #5484
-	if (!previousFile?.isConnected) {
-		return;
-	}
 
 	event.preventDefault();
 	event.stopImmediatePropagation();
 
 	const thisFile = event.delegateTarget.closest('.js-file')!;
 	const files = select.all('.js-file');
-	const previousFileState = isFirstToggle ? !isChecked(thisFile) : isChecked(previousFile);
+	const previousFileState = isFirstToggle ? !isChecked(thisFile) : isChecked(previousFile!);
 
 	const selectedFiles = getItemsBetween<HTMLElement>(files, previousFile, thisFile);
 
--- a/source/features/batch-mark-files-as-viewed.tsx
+++ b/source/features/batch-mark-files-as-viewed.tsx
@@ -22,6 +22,8 @@ function batchToggle(event: delegate.Event<MouseEvent, HTMLFormElement>): void {
 		return;
 	}
 
+	const isFirstToggle = previousFile === undefined;
+
 	previousFile ??= select('.js-file'); // #5484
 	if (!previousFile?.isConnected) {
 		return;
@@ -32,6 +34,7 @@ function batchToggle(event: delegate.Event<MouseEvent, HTMLFormElement>): void {
 
 	const thisFile = event.delegateTarget.closest('.js-file')!;
 	const files = select.all('.js-file');
+	const previousFileState = isFirstToggle ? !isChecked(thisFile) : isChecked(previousFile);
 
 	const selectedFiles = files.slice(...[
 		files.indexOf(previousFile),
@@ -39,7 +42,7 @@ function batchToggle(event: delegate.Event<MouseEvent, HTMLFormElement>): void {
 	].sort((a, b) => a - b));
 
 	for (const file of selectedFiles) {
-		if (!isChecked(file)) {
+		if (isChecked(file) !== previousFileState) {
 			select('.js-reviewed-checkbox', file)!.click();
 		}
 	}
--- a/source/features/batch-mark-files-as-viewed.tsx
+++ b/source/features/batch-mark-files-as-viewed.tsx
@@ -23,8 +23,6 @@ function batchToggle(event: delegate.Event<MouseEvent, HTMLFormElement>): void {
 		return;
 	}
 
-	const isFirstToggle = previousFile === undefined;
-
 	previousFile ??= select('.js-file'); // #5484
 
 	event.preventDefault();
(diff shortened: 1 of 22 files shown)


source/features/batch-mark-files-as-viewed.tsx
```diff
@@ -63,6 +64,11 @@ function onAltClick(event: delegate.Event<MouseEvent, HTMLInputElement>): void {
 }
 
 function init(): Deinit[] {
+	// eslint-disable-next-line @typescript-eslint/no-floating-promises
+	elementReady('.js-file').then(file => {
+		previousFile = file;
+	});
```
cheap-glitch, line 70: Let's not use `elementReady` when we don't need it.

This can be implemented in a simpler way: in `batchToggle`, check if `previousFile` is undefined. If it is, select the first `.js-file` element.

cheap-glitch (changes requested):

source/features/batch-mark-files-as-viewed.tsx
```diff
@@ -18,23 +18,28 @@ function isChecked(file: HTMLElement): boolean {
...
+
 	const selectedFiles = files.slice(...[
-		files.indexOf(previousFile) + 1,
+		files.indexOf(previousFile),
 		files.indexOf(thisFile) + 1,
 	].sort((a, b) => a - b));
 
 	for (const file of selectedFiles) {
-		if (isChecked(file) !== previousFileState) {
+		if (!isChecked(file)) {
```
cheap-glitch, line 42: This is _not_ equivalent. The feature should work the same whether checking or unchecking multiple files.

cheap-glitch (changes requested):

source/features/batch-mark-files-as-viewed.tsx
jopemachine: > This is _not_ equivalent. The feature should work the same whether checking or unchecking multiple files.

Thanks for pointing this out!
I just updated the codes.

source/features/batch-mark-files-as-viewed.tsx
```diff
@@ -18,18 +18,26 @@ function isChecked(file: HTMLElement): boolean {
...
 
 function batchToggle(event: delegate.Event<MouseEvent, HTMLFormElement>): void {
-	if (!event.shiftKey || !previousFile) {
+	if (!event.shiftKey) {
+		return;
+	}
+
+	const isFirstToggle = previousFile === undefined;
+
+	previousFile ??= select('.js-file'); // #5484
```
yakov116, line 27: Wow I have never seen `??=` used in real life.
I had to double check its supported for the browsers we support!
https://caniuse.com/?search=%3F%3F%3D

cheap-glitch: > Wow I have never seen `??=` used in real life.

I think it's my first time using it actually. I feel so swanky :tophat: 

> I had to double check its supported for the browsers we support!

Actually I just assumed TypeScript would transpile it

source/features/batch-mark-files-as-viewed.tsx
```diff
@@ -18,18 +18,26 @@ function isChecked(file: HTMLElement): boolean {
...
 		return;
 	}
 
 	event.preventDefault();
 	event.stopImmediatePropagation();
 
-	const previousFileState = isChecked(previousFile);
 	const thisFile = event.delegateTarget.closest('.js-file')!;
 	const files = select.all('.js-file');
+	const previousFileState = isFirstToggle ? !isChecked(thisFile) : isChecked(previousFile);
```
fregante, line 37: Our logic was a little silly and we're carrying it over here. Let me explain:

- if I click an **unchecked toggle**, I want it to become **checked**
- if I alt-click an **unchecked toggle**, I want _all_ the (related) toggles to become **checked**

What we were doing on line 28 was discarding the current toggle and just using the _previous_’ toggle state. Example:

1. Start with:

	- [ ] a
	- [ ] b
	- [x] c

2. Click on `a`
3. Alt-click on `c`

You'll see this weird behavior (on `main`, not on this PR):


https://user-images.githubusercontent.com/1402241/163171226-9b00a756-c925-4f11-a067-f8ab6e9092ad.mov



This condition should probably be `isChecked(thisFile)` (possibly negated, depending on whether the listener is running before or after the state is changed by the native click)

jopemachine: @fregante

Thanks for your detailed descriptions :)
I organized my thoughts here.

> This condition should probably be isChecked(thisFile)

I agree with this point.
I think maybe the below code using `isChecked(thisFile)` instead of `previousFileState` should work.

```ts
function batchToggle(event: delegate.Event<MouseEvent, HTMLFormElement>): void {
  if (!event.shiftKey) {
    return;
  }

  previousFile ??= select('.js-file'); // #5484
  if (!previousFile?.isConnected) {
    return;
  }

  event.preventDefault();
  event.stopImmediatePropagation();

  const thisFile = event.delegateTarget.closest('.js-file')!;
  const files = select.all('.js-file');

  const selections = [
    files.indexOf(previousFile),
    files.indexOf(thisFile),
  ].sort((a, b) => a - b);

  selections[1] += 1;

  const selectedFiles = files.slice(...selections);

  for (const file of selectedFiles) {
    if (isChecked(file) === isChecked(thisFile)) {
      select('.js-reviewed-checkbox', file)!.click();
    }
  }
}
```

> it doesn't break/change the regular behavior.

I think the changed code seems to affect only `isFirstToggle` is true.
So, it works as existing code (as expected) when `isFirstToggle` is false.

> this new behavior works

Yes, new behavior works as expected.

fregante: > breaks existing behaviors

If it works correctly, that's ok. I just meant that the behavior I described is unrelated to this PR and doesn't _have to_ be fixed here. If you want to do that, feel free to.

fregante: I kinda wish this code was testable. If you decide to fix both, can you extract this logic into a file under `source/helpers`?

Maybe something like this, which should not consider `isChecked` at all:

```ts
/** Get the items between previous and current, both ends included. If `previous` is missing, start from 0 */
function getItemsBetween<T>(items: T[], previous: T | undefined, current: T): T[] {

}
```

I think once you extract this, the rest of the function should be nearly unchanged. There's already `isChecked` in the last loop that takes care of the skipping unrelated items.

The core feature has 3 parts:

1. Determining the final state: set it to `thisFile`, unconditionally
2. Getting the items between previous and current
3. Setting all the items to the final state we already know, skipping those that are already that

fregante: 👆 `getItemsBetween` is the only logic that needs to be tested, making sure that it works even when previous is undefined or if its index comes after `current` (that's already taken care by `sort`)

jopemachine: > If you decide to fix both, can you extract this logic into a file under source/helpers?

I'd like to fix the other part if possible (in another PR).
I just refactored the code using `getItemsBetween`.

test/helpers.ts
```diff
@@ -183,3 +184,12 @@ test('shouldFeatureRun', t => {
 		exclude: yesNo,
 	}), 'If any `exclude` is true, then it should not run, regardless of `asLongAs` and `include`');
 });
+
+test('getItemsBetween', t => {
+	t.deepEqual(getItemsBetween<number>([1, 10, 100], 10, 100), [10, 100]);
```
fregante, line 189: Let's ensure that it works in the middle of the array + alt-clicking the same element

```suggestion
	t.deepEqual(getItemsBetween<number>([1, 10, 100, 1000], 10, 100), [10, 100]);
	t.deepEqual(getItemsBetween<number>([1, 10, 100, 1000], 10, 10), [10]);
```

Can you also replace 1/10/100/1000 with something more easily distinguishable? Colored emojis would be great. Let's make the initial array static too:

```js
const list = ['❤️', '💛', '💚', '💙']
```

source/features/batch-mark-files-as-viewed.tsx
```diff
@@ -18,20 +19,25 @@ function isChecked(file: HTMLElement): boolean {
...
 
 function batchToggle(event: delegate.Event<MouseEvent, HTMLFormElement>): void {
-	if (!event.shiftKey || !previousFile) {
+	if (!event.shiftKey) {
+		return;
+	}
+
+	const isFirstToggle = previousFile === undefined;
+
+	previousFile ??= select('.js-file'); // #5484
```
fregante, line 28: This should now be unnecessary. This "if missing" is already part of `getItemsBetween`

`previousFileState` should be dropped as suggested, we should only follow the current file state. It may be best to fix the whole behavior in this PR I think

source/features/batch-mark-files-as-viewed.tsx
```diff
@@ -18,20 +19,25 @@ function isChecked(file: HTMLElement): boolean {
...
+	if (!previousFile?.isConnected) {
```
fregante, line 29: Why `isConnect`? This is impossible. `batchToggle` is called when the user clicks inside a `.js-file`, so there's no chance that it doesn't exist or that is not connected.

source/features/batch-mark-files-as-viewed.tsx
fregante: > I'd like to fix the other part if possible (in another PR).

I think that's as easy as using the current file’s state instead of the `previous` file. We just need to ensure that `isChecked` works on the current file. By "works" I mean:

- if I'm **checking** the file, then it should return `true` immediately

jopemachine:
> It may be best to fix the whole behavior in this PR I think

Understood. I thought I need to create one more PR. I will try to update the whole behavior in this PR directly.

test/helpers.ts
```diff
@@ -183,3 +184,15 @@ test('shouldFeatureRun', t => {
 		exclude: yesNo,
 	}), 'If any `exclude` is true, then it should not run, regardless of `asLongAs` and `include`');
 });
+
+test('getItemsBetween', t => {
```
cheap-glitch, line 188: Thanks for writing some tests! Could you move them to a separate `get-items-between.test.ts` file and place it next to the helper? I think that's what causing the conflict

source/helpers/get-items-between.ts
```diff
@@ -3,11 +3,9 @@ Get the items between previous and current, both ends included. If `previous` is
 */
 export default function getItemsBetween<T>(items: T[], previous: T | undefined, current: T): T[] {
 	const selections = [
-		previous ? items.indexOf(previous) : 0,
-		items.indexOf(current),
+		items.indexOf(previous) + 1,
+		items.indexOf(current) + 1,
```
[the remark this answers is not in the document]
fregante: Ah right, I tried restoring the previous behavior, forgetting we need to include the previous checkbox now, in order to change it if necessary

jopemachine:
I just updated this PR.
It seems like it works as expected now.
Here are some manual tests I've tried.

### `#1`
demo1

### `#2`
demo2

source/features/batch-mark-files-as-viewed.tsx
```diff
@@ -23,22 +23,16 @@ function batchToggle(event: delegate.Event<MouseEvent, HTMLFormElement>): void {
...
-	const thisFile = event.delegateTarget.closest('.js-file')!;
 	const files = select.all('.js-file');
+	const thisFile = event.delegateTarget.closest('.js-file')!;
+	const isThisFileChecked = isChecked(thisFile);
 
-	const selectedFiles = getItemsBetween<HTMLElement>(files, previousFile, thisFile);
-
-	if (files.indexOf(previousFile!) > files.indexOf(thisFile)) {
-		selectedFiles.reverse();
-	}
```
cheap-glitch: Doesn't really matter in which order the checkboxes are clicked

source/helpers/get-items-between.ts
```diff
@@ -2,12 +2,8 @@
...
-		previous ? items.indexOf(previous) : 0,
-		items.indexOf(current),
-	].sort((a, b) => a - b);
+	const start = previous ? items.indexOf(previous) : 0;
+	const end = items.indexOf(current);
 
-	selections[1] += 1;
-
-	return items.slice(...selections);
+	return items.slice(Math.min(start, end), Math.max(start, end) + 1);
```
cheap-glitch, line 8: Keep It Simple™

cheap-glitch (approved):
Did some manual testing too and it seems to work as expected :100: 

Thanks for your hard work @jopemachine!

source/features/batch-mark-files-as-viewed.tsx
jopemachine: > Doesn't really matter in which order the checkboxes are clicked

This code is related to the call orders of `remember` when `toggleBatch` executed.

You can reproduce the below bug.

The third `batchToggle` is not working as expected in the gif.

bug1

This is because `remember` is always called as the array (selectedFiles)'s order.

For example,

```ts
// Expected result: called in order remember(0), remember(1), remember(2)
// This is same with actual result.
// Works as expected.

const selectedFiles = getItemsBetween(files, 0th_item, 3rd_item);

for (const file of selectedFiles) {
	if (isChecked(file) === isThisFileChecked) {
		select('.js-reviewed-checkbox', file)!.click();
	}
}

```

```ts
// Expected result: called in order remember(2), remember(1), remember(0)
// Actual result: called in order remember(0), remember(1), remember(2)
// Because last called `remember` saves the `previousFile`, previousFile will be files[2]. (And it cause the bug.)

const selectedFiles = getItemsBetween(files, 3rd_item, 0th_item);

for (const file of selectedFiles) {
	if (isChecked(file) === isThisFileChecked) {
		select('.js-reviewed-checkbox', file)!.click();
	}
}
```

source/features/batch-mark-files-as-viewed.tsx
```diff
@@ -18,23 +19,20 @@ function isChecked(file: HTMLElement): boolean {
...
-	const thisFile = event.delegateTarget.closest('.js-file')!;
 	const files = select.all('.js-file');
-	const selectedFiles = files.slice(...[
-		files.indexOf(previousFile) + 1,
-		files.indexOf(thisFile) + 1,
-	].sort((a, b) => a - b));
+	const thisFile = event.delegateTarget.closest('.js-file')!;
+	const isThisFileChecked = isChecked(thisFile);
 
+	const selectedFiles = getItemsBetween(files, previousFile ?? files[0], thisFile);
```
fregante, line 33: If you're going to always provide the previous file, then we don't need to duplicate the logic in `getItemsBetween` with `| undefined` and the ternary operator

source/features/batch-mark-files-as-viewed.tsx
fregante: The logic seems faulty:

- if current is checked like `thisFile`, click it

It should be the opposite:

- if it's different, then click it so it matches the current file.

This suggests that `isChecked` might not be working in time so `isThisFileChecked` is actually `wasThisFileCheckedBeforeTheClick`, which is a long condition to keep in mind.

fregante (changes requested):
I suggest using `table {display: none !important;}` when testing the feature so the files never expand. Unfortunately indeed the last 2 commits broke it:

g2if


while before it was:

gif

source/features/batch-mark-files-as-viewed.tsx
fregante: > Doesn't really matter in which order the checkboxes are clicked

This should be right, but indeed something is not working in this last commit

fregante: Ohhhh I finally understood your comment:

> Because last called `remember` saves the `previousFile`

I think we can avoid this by using:

```js
function remember(event: delegate.Event): void {
	if (event.isTrusted) {
		previousFile = event.delegateTarget.closest('.js-file')!;
	}
}
```

This specifically targets manual clicks and excludes everything else

fregante (approved):
It works and it's correct, but at what cost? 🥲 

gif


(there's a single non-alt click on the first checkbox towards the end of the gif)

cheap-glitch (approved):

mgol:
Not sure if that's related but since the today's release I can no longer mark all files as viewed. No matter what files I select manually, once I click on any checkbox with Alt pressed, all my viewed files get marked as not viewed again and only the single one I Alt-clicked is now viewed. 😢

fregante:
Very likely related, what browser?

mgol:
This was on Chrome 100.0.4896.127 on macOS 12.3.1 (21E258). The PR was quite large.

PR #5588, opened 2022-04-11
Fix slowdown caused by `select-all-notifications-shortcut`

### LINKED ISSUE #5569: Extension slow down GitHub Notifications search
### Please ensure:

### Example URLs

https://github.com/notifications/beta?query=is%3Adone&query=is%3Adone+blabla

### Extension version

22.4.1

### Browser(s) used

Firefox 98.0.2

### Description

I have hundreds subscriptions in the Done page.
If try to search and type in the filter field a random word and press enter, then it freezes with the Firefox popup:

```
"Refined GitHub" is slowing down Firefox. To speed up your browser, stop that extension. ...
```

If press stop, then it continues to search results,

### Include in this issue:

- [ ] Console errors, if any

cheap-glitch:
Fixes #5569 

## Test URLs

## Screenshot

https://user-images.githubusercontent.com/46634000/162682068-4d85ff52-89b5-47e6-ad81-641a8b7dd6df.mp4

### DIFF
--- a/source/features/select-all-notifications-shortcut.tsx
+++ b/source/features/select-all-notifications-shortcut.tsx
@@ -1,16 +1,22 @@
-import elementReady from 'element-ready';
+import select from 'select-dom';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
-import onElementRemoval from '../helpers/on-element-removal';
 
-async function init(signal: AbortSignal): Promise<void> {
-	const selectAllNotifications = await elementReady('.js-notifications-mark-all-prompt');
-	if (selectAllNotifications) { // Notifications page may be empty
-		selectAllNotifications.dataset.hotkey = 'a';
-		await onElementRemoval(selectAllNotifications, signal); // "Select all" checkbox will be replaced if there's more notifications to load #4199
-		void init(signal);
+function runShortcut(event: KeyboardEvent): void {
+	if (
+		event.key === 'a'
+		&& !event.ctrlKey
+		&& !event.metaKey
+		&& !event.isComposing
+	) {
+		select('.js-notifications-mark-all-prompt')!.click();
 	}
+}
+
+function init(signal: AbortSignal): void {
+	// Listen to the shortcut ourselves instead of attaching it to the checkbox #5569
+	document.body.addEventListener('keypress', runShortcut, {signal});
 }
 
 void features.add(import.meta.url, {
--- a/source/features/select-all-notifications-shortcut.tsx
+++ b/source/features/select-all-notifications-shortcut.tsx
@@ -20,6 +20,9 @@ function init(signal: AbortSignal): void {
 }
 
 void features.add(import.meta.url, {
+	shortcuts: {
+		a: 'Select all notifications',
+	},
 	include: [
 		pageDetect.isNotifications,
 	],
--- a/source/features/select-all-notifications-shortcut.tsx
+++ b/source/features/select-all-notifications-shortcut.tsx
@@ -23,6 +23,9 @@ void features.add(import.meta.url, {
 	include: [
 		pageDetect.isNotifications,
 	],
+	exclude: [
+		() => select.exists('img[src$="notifications/inbox-zero.svg"]'), // Don't run on empty inbox page
+	],
 	awaitDomReady: false,
 	init,
 });


source/features/select-all-notifications-shortcut.tsx
```diff
@@ -1,18 +1,24 @@
...
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
-import onElementRemoval from '../helpers/on-element-removal';
 
-async function init(signal: AbortSignal): Promise<void> {
-	const selectAllNotifications = await elementReady('.js-notifications-mark-all-prompt');
-	if (selectAllNotifications) { // Notifications page may be empty
-		selectAllNotifications.dataset.hotkey = 'a';
-		await onElementRemoval(selectAllNotifications, signal); // "Select all" checkbox will be replaced if there's more notifications to load #4199
```
cheap-glitch, line 11: @fregante There's also another feature which looks at the "select all" checkbox, and might cause a slowdown too:

https://github.com/refined-github/refined-github/blob/bd48d98f400569c02f0a4ac774b1554dde57e710/source/features/select-notifications.tsx#L160

cheap-glitch:
@yakov116 Could you take a quick look at this if you have the time? I'd like to merge it before today's release

source/features/select-all-notifications-shortcut.tsx
```diff
@@ -1,22 +1,34 @@
...
+
 void features.add(import.meta.url, {
+	shortcuts: {
+		a: 'Select all notifications',
+	},
 	include: [
 		pageDetect.isNotifications,
 	],
+	exclude: [
+		() => select.exists('img[src$="notifications/inbox-zero.svg"]'), // Don't run on empty inbox page
```
yakov116, line 30: Nit pick
On empty if you click `a` when switching tabs (From participating to Assigned) and its empty

```
Uncaught TypeError: Cannot read properties of undefined (reading 'click')
    at HTMLBodyElement.select_all_notifications_shortcut_runShortcut
```

yakov116 (approved):

cheap-glitch:
> The saved tab

Nice catch, but this isn't the relevant feature :wink:

source/features/select-all-notifications-shortcut.tsx
fregante: If configured correctly, selector-observer already deals with it the right way, probably

source/features/select-all-notifications-shortcut.tsx
fregante: I suppose it would also work if:

1. Select all
2. Mark as done
3. The list is empty
4. Press `a`
5. It throws

But yeah, if that's the procedure, it's not worth fixing

source/features/select-all-notifications-shortcut.tsx
```diff
@@ -1,22 +1,34 @@
...
+		&& !event.metaKey
+		&& !event.isComposing
+	) {
+		select('.js-notifications-mark-all-prompt')!.click();
 	}
 }
 
+function init(signal: AbortSignal): void {
+	// Listen to the shortcut ourselves instead of attaching it to the checkbox #5569
+	document.body.addEventListener('keypress', runShortcut, {signal});
```
fregante, line 19: We could potentially also have a helper that registers a callback with the native `data-hotkey`, e.g.


```js
function registerHotkey(hotkey, handler) {
	document.body.append(<button data-hotkey={hotkey} onClick={handler}>)
	return () => select(`[data-hotkey="${hotkey}"`)?.remove()
}
```

source/features/select-all-notifications-shortcut.tsx
cheap-glitch: > It throws

It doesn't, the "Select all" checkbox stays in the DOM until the page is reloaded

yakov116: I always get it on https://github.com/notifications?query=reason%3Aassign I have nothing there I refreshed a whole bunch of times

cheap-glitch: > I always get it on [github.com/notifications?query=reason%3Aassign](https://github.com/notifications?query=reason%3Aassign) I have nothing there I refreshed a whole bunch of times

Weird, I have no error on Firefox :thinking:

Though I see the feature is running on this tab even when it's empty, because the background image has a slightly different name and thus the `exclude` selector isn't triggered.

We have several notification-related features now, so it's probably worth adding some detections like `isEmptyNotificationPage` and `isNotificationsInboxPage`, that kind of thing.

PR #5591, opened 2022-04-11
Meta: Move some test modules into `source` folder

cheap-glitch:
https://github.com/refined-github/refined-github/issues/5222#issuecomment-1070525355

### DIFF
--- a/source/github-helpers/github-url.test.ts
+++ b/source/github-helpers/github-url.test.ts
@@ -1,6 +1,6 @@
 import test from 'ava';
 
-import '../../fixtures/globals';
+import '../../test/fixtures/globals';
 import GitHubURL from './github-url';
 
 test('branch', t => {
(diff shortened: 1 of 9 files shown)


source/github-helpers/github-url.test.ts
```diff
@@ -1,7 +1,7 @@
 import test from 'ava';
 
-import './fixtures/globals';
-import GitHubURL from '../source/github-helpers/github-url';
+import '../../fixtures/globals';
```
cheap-glitch, line 3: Worth setting up a "path shortcut" in `tsconfig.json` (e.g. `~`) to avoid this?

fregante (approved):

source/github-helpers/github-url.test.ts
fregante: I'd rather not, Maybe there's a way to have AVA automatically load this as an env/setup file

Issue #5435, opened 2022-02-24
`forked-to`: Cannot set properties of undefined (setting 'href')


kidonng:
### Please ensure:

- [X] The bug is caused by Refined GitHub. It doesn't happen if I disable the extension.
- [X] The bug happens after clearing extension cache. The "Clear cache" button below can also be found at the bottom of the options page.

### Example URLs

Just browse around https://github.com/refined-github/refined-github and you will see it in the console

### Extension version

22.2.22

### Browser(s) used

Chrome

### Description

```
TypeError: Cannot set properties of undefined (setting 'href')
    at init (chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/build/refined-github.js:7691:91)
    at async runFeature (chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/build/refined-github.js:3045:20)
    at async setupPageLoad (chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/build/refined-github.js:3055:34)
```

Seems related to #5417. Actual feature is not broken, just some weird error.

### Include in this issue:

- [ ] Screenshots/video/gif demonstrating the bug, if it’s visual
- [X] Console errors, if any

fregante:
Fixed in my PR.

The issue was that we add **and expect** a relative `href` attribute (`/user/repo/page`) but `sort-conversations-by-update-time` sets the `href` to a full URL.

https://github.com/refined-github/refined-github/blob/f32792489d8ce1f7c404346916f8697d882421e5/source/features/sort-conversations-by-update-time.tsx#L28

PR #5597, opened 2022-04-14
Restore `forked-to` feature and enable it for every user

### LINKED ISSUE #5596: `forked-to` seems broken
### Please ensure:

### Example URLs

### Extension version

22.4.11

### Browser(s) used

google chrome  100.0.4896.88

### Description

link to your fork button disappears.
* I have cleared cache
* I have my personal token in refined-github
* I have forked refined-github https://github.com/yzqdev/refined-github

link to your fork button disappears.
I guess maybe it's github changed the fork procedure that causes this. Two days before, when I clicked fork button , a dialog jumped, while now when i click fork button, it'll turn to a new fork page.

### after
image

### Include in this issue:

- [ ] Console errors, if any

cheap-glitch:
Fixes #5596 

## Test URLs

Fork a repo :fork_and_knife:  

## Screenshot

image

### DIFF
--- a/source/features/forked-to.tsx
+++ b/source/features/forked-to.tsx
@@ -45,7 +45,7 @@ async function updateUI(forks: string[]): Promise<void> {
 		return;
 	}
 
-	const forkButton = select('[aria-label^="Fork your own copy of"]')!;
+	const forkButton = select('.pagehead-actions [aria-label^="Fork your own copy of"]')!;
 	forkButton.classList.add('rounded-left-2', 'BtnGroup-item');
 
 	if (forks.length === 1) {
@@ -117,7 +117,7 @@ async function init(): Promise<void | false> {
 	}
 
 	// This feature only applies to users that have multiple organizations, because that makes a fork picker modal appear when clicking on "Fork"
-	const hasOrganizations = select.exists('[data-ga-click*="show fork modal"]');
+	const hasOrganizations = select.exists('.pagehead-actions [data-ga-click*="show fork modal"]');
 
 	// Only fetch/update forks when we see a fork (on the current page or in the cache).
 	// This avoids having to `updateCache` for every single repo you visit.
--- a/source/features/forked-to.tsx
+++ b/source/features/forked-to.tsx
@@ -1,7 +1,6 @@
 import React from 'dom-chef';
 import cache from 'webext-storage-cache';
 import select from 'select-dom';
-import elementReady from 'element-ready';
 import * as pageDetect from 'github-url-detection';
 import {CheckIcon, ChevronRightIcon, TriangleDownIcon, XIcon} from '@primer/octicons-react';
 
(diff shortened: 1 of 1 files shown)


source/features/forked-to.tsx
```diff
@@ -123,7 +117,7 @@ async function init(): Promise<void | false> {
 	}
 
 	// This feature only applies to users that have multiple organizations, because that makes a fork picker modal appear when clicking on "Fork"
-	const hasOrganizations = await elementReady('details-dialog[src*="/fork"]', {waitForChildren: false});
+	const hasOrganizations = select.exists('[data-ga-click*="show fork modal"]');
```
cheap-glitch, line 120: This targets an attribute on the "Fork" button itself, so we shouldn't need the `elementReady`. Not sure if this is really limited to people in organizations though.

cheap-glitch (commented):
@refined-github/maintainers Could someone who's in several organizations test this with multiple forks? Thanks :heart_decoration:

fregante:
GitHub really broke the flow here and I expect them to change this once again soon.

The fork button now **always** points to the new `/fork` page, even if I have a fork.

Big deal breaker: there is NO link to my existing fork. At least I don't see it.

`forked-to` just became _mandatory_ for every GitHub user. 😃

fregante:
Separately perhaps I'd also replace the `?fragment` fetch with an API call because that fragment is no longer shown on GitHub and I expect them to drop it soon (unless they restore it)

source/features/forked-to.tsx
```diff
@@ -123,7 +117,7 @@ async function init(): Promise<void | false> {
 	}
 
 	// This feature only applies to users that have multiple organizations, because that makes a fork picker modal appear when clicking on "Fork"
-	const hasOrganizations = await elementReady('details-dialog[src*="/fork"]', {waitForChildren: false});
+	const hasOrganizations = select.exists('.pagehead-actions [data-ga-click*="show fork modal"]');
```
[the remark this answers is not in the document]
cheap-glitch: So just to be clear we can drop the `hasOrganizations` check and just use `fork || isForkedRepo()`, right?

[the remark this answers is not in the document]
fregante: Probably! There's no `details` anywhere so anything related to that should be gone.

fregante:
I suppose this will also close https://github.com/refined-github/refined-github/issues/5435

cheap-glitch:
> I suppose this will also close #5435

I don't know, I've never seen this error in the first place

source/features/forked-to.tsx
```diff
@@ -122,12 +116,9 @@ async function init(): Promise<void | false> {
 		await updateUI(forks);
 	}
 
-	// This feature only applies to users that have multiple organizations, because that makes a fork picker modal appear when clicking on "Fork"
-	const hasOrganizations = await elementReady('details-dialog[src*="/fork"]', {waitForChildren: false});
-
 	// Only fetch/update forks when we see a fork (on the current page or in the cache).
 	// This avoids having to `updateCache` for every single repo you visit.
-	if (forks || (hasOrganizations && pageDetect.isForkedRepo())) {
+	if (forks || pageDetect.isForkedRepo()) {
```
fregante, line 121: Can you flip this condition to allow an early return? I think

```suggestion
	if (!forks && !pageDetect.isForkedRepo()) {
		return false;
	}
```

fregante (approved):
Feel free to merge and release a new version today

source/features/forked-to.tsx
fregante, line 121: Can you flip this condition to allow an early return? I think

```suggestion
	if (!forks && !pageDetect.isForkedRepo()) {
		return false;
	}
```

cheap-glitch: Thought about it too but in this case, since the first block is only one line long, I think it just makes the logic less clear while providing no benefits at runtime.

kidonng:
> GitHub really broke the flow here and I expect them to change this once again soon.

[_The repository fork button now displays existing forks_](https://github.blog/changelog/2022-06-29-the-repository-fork-button-now-displays-existing-forks/).

fregante:
Called it!

Screen Shot 8


> Is this @fregante's work? Impressive 🎉

Screen Shot

First requested in https://github.com/refined-github/refined-github/issues/1129 and implemented in https://github.com/refined-github/refined-github/pull/2153

PR #5604, opened 2022-04-17
Meta: Use `registerHotkey` everywhere applicable

cheap-glitch:
Also lint some hotkey listeners

### DIFF
--- a/source/features/conversation-activity-filter.tsx
+++ b/source/features/conversation-activity-filter.tsx
@@ -6,10 +6,11 @@ import elementReady from 'element-ready';
 import * as pageDetect from 'github-url-detection';
 import {CheckIcon, EyeClosedIcon, EyeIcon, XIcon} from '@primer/octicons-react';
 
+import {wrap} from '../helpers/dom-utils';
 import features from '.';
 import onNewComments from '../github-events/on-new-comments';
+import registerHotkey from '../github-helpers/register-hotkey';
 import {getRghIssueUrl} from '../helpers/rgh-issue-link';
-import {wrap, isEditable} from '../helpers/dom-utils';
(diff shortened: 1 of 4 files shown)


source/features/conversation-activity-filter.tsx
```diff
@@ -221,7 +218,7 @@ async function init(): Promise<void> {
 		applyState(state);
 	}
 
-	document.body.addEventListener('keypress', runShortcut);
+	return registerHotkey('h', switchToNextFilter);
```
cheap-glitch, line 221: https://user-images.githubusercontent.com/46634000/163721770-9602d35c-da84-4e1e-8f2c-a674c4d3f76f.mp4

source/features/selection-in-new-tab.tsx
```diff
@@ -2,22 +2,24 @@ import select from 'select-dom';
...
+		openUrls: [selected.href],
+	});
+
+	// Get the list element that contains the unread class and mark it as read.
+	selected.closest('.unread')?.classList.replace('unread', 'read');
 }
 
 function init(): void {
-	document.addEventListener('keypress', openInNewTab);
+	registerHotkey('O', openInNewTab);
```
cheap-glitch, line 22: https://user-images.githubusercontent.com/46634000/163721843-122ed089-e69e-4105-87a0-4473b65c0ae8.mp4

fregante: I think it would be shift-o

cheap-glitch: > I think it would be shift-o

No, GitHub understand `O` just fine. We already have `S` for the "Select notification" dropdown, for example.

fregante (approved):

PR #5605, opened 2022-04-20
Remove `avoid-accidental-submissions` message on the second submission

fregante:
The message stays visible even while the submission is going through. This will hide the message on the second `enter`

## Test URLs

wip


source/features/avoid-accidental-submissions.tsx
```diff
@@ -54,7 +59,10 @@ const inputElements = [
 ];
 
 function init(): Deinit {
-	return delegate(document, inputElements.join(','), 'keydown', onKeyDown);
+	return [
+		delegate(document, inputElements.join(','), 'keydown', onKeyDown),
+		delegate(document, ['TODO'].join(','), 'submit', onSubmission),
```
fregante, line 64: Initially I thought I could just listen to events on the fields, but actually we should listen in on submissions on the form itself, so it catches all submissions. No time right now to find the form selector 🤭

PR #5607, opened 2022-04-20
Extract `submission-via-ctrl-enter-everywhere` from others

darkred:
This PR follows the suggestion by @fregante  in https://github.com/refined-github/refined-github/pull/4252#discussion_r853839752 i.e. extracts the `addQuickSubmit()` function *(which adds ability to submit new/edited files via ctrl-enter)* as a separate feature: `allow-submission-via-ctrl-enter`.

> https://github.com/refined-github/refined-github/blob/c2526b02d9d7da4aec7ea8085526884f0d8e6a20/source/features/avoid-accidental-submissions.tsx#L9-L11
> Not sure why we added this here. Doesn't this **allow** submission via ctrl-enter? It should be extracted to its own feature (and the shortcut should mention `cmd` on Mac )

## Notes to reviewers:

1. For reference, I originally got the above initial function from the [create-release-shortcut](https://github.com/refined-github/refined-github/blob/main/source/features/avoid-accidental-submissions.tsx) feature, following yakov's suggestion in https://github.com/refined-github/refined-github/pull/4252#discussion_r618586679 :
https://github.com/refined-github/refined-github/blob/c2526b02d9d7da4aec7ea8085526884f0d8e6a20/source/features/create-release-shortcut.tsx#L6-L8
&nbsp;
2. Maybe I should use `isMac` (with a ternary `if`) instead of `'ctrl/cmd enter': 'Publish a new/edited file',` ?
Can it be used when defining keyboard shortcuts ? 
&nbsp;
3. Maybe the feature should be renamed to something more descriptive, e.g. 
`allow-submission-with-ctrl-enter-when-publishing-files`


fregante:
> 2\. Maybe I should use `isMac`

Yes, it can be used there

> 3\. Maybe the feature should be renamed to something more descriptive, e.g.

The feature should not be necessarily limited to files, we're just ensuring that GitHub’s own behavior is consistent. Maybe `submission-via-ctrl-enter-everywere`? There are more places where this can be enabled too, I think it doesn't work when editing releases.

readme.md
```diff
@@ -327,6 +327,7 @@ Thanks for contributing! 🦋🙌
 - [](# "latest-tag-button") [Adds a link to the latest version tag on directory listings and files.](https://user-images.githubusercontent.com/44045911/155496122-6267c45d-21d4-45c9-adf7-94c9e41cc652.png)
 - [](# "convert-release-to-draft") [Adds a button to convert a release to draft.](https://user-images.githubusercontent.com/46634000/139236979-44533bfd-5c17-457d-bdc1-f9ec395f6a3a.png)
 - [](# "link-to-changelog-file") [Adds a button to view the changelog file from the releases page.](https://user-images.githubusercontent.com/46634000/139236982-a1bce2a2-f3aa-40a9-bca4-8756bc941210.png)
+- [](# "allow-submission-via-ctrl-enter") Adds ability to submit new/edited files via Ctrl+Enter .
```
fregante, line 330: This should also be updated to reflect that we're not adding it, but making it work where GitHub forgot, so it should also be moved to the GitHub Shortcomings section

darkred: Ok, so I updated the description and moved it to the GitHub Shortcomings section. Thanks.

source/features/submission-via-ctrl-enter-everywere.tsx
```diff
@@ -0,0 +1,20 @@
...
+import {isMac} from '../github-helpers';
+import features from '.';
+
+function addQuickSubmit(): void {
+	select('input#commit-summary-input')!.classList.add('js-quick-submit');
+}
+
+void features.add(import.meta.url, {
+	shortcuts: {
+		[isMac ? 'cmd enter' : 'ctrl enter']: 'Publish a new/edited file',
```
darkred, line 13: I tried to test it by changing the user agent in Chrome but unfortunately it doesn't change accordingly (it should change to <kbd>cmd</kbd><kbd>enter</kbd>) :

darkred:
> The feature should not be necessarily limited to files, we're just ensuring that GitHub’s own behavior is consistent. Maybe `submission-via-ctrl-enter-everywere`? There are more places where this can be enabled too, I think it doesn't work when editing releases.

Regarding the above, the currently used text for the created shortcut is "'Publish a new/edited file'".
As I noted in the OP, the `addQuickSubmit()` function also exists in the [`create-release-shortcut`](https://github.com/refined-github/refined-github/blob/main/source/features/avoid-accidental-submissions.tsx) feature.
I don't know/haven't found in which other places to also enable this new feature.

fregante:
> Regarding the above, the currently used text for the created shortcut is "'Publish a new/edited file'".

We should drop it. GitHub does not show that shortcut anywhere else

darkred:
> We should drop it. GitHub does not show that shortcut anywhere else

Ok, I dropped the shortcut.
 
> Good find! Let's move those selectors here too since it's identical.

I moved the selectors.

readme.md
```diff
@@ -399,6 +399,7 @@ Thanks for contributing! 🦋🙌
 - [](# "clean-header-search-field") [Clears duplicate queries in the header search field.](https://user-images.githubusercontent.com/1402241/114177338-7c890300-9966-11eb-83a3-a711a76fae99.png)
 - [](# "prevent-pr-merge-panel-opening") Prevents the merge panel from automatically opening on every page load after it’s been opened once.
 - [](# "command-palette-navigation-shortcuts") Adds keyboard shortcuts to select items in command palette using <kbd>ctrl</kbd> <kbd>n</kbd> and <kbd>ctrl</kbd> <kbd>p</kbd> (macOS only).
+- [](# "submission-via-ctrl-enter-everywhere") Enables submission via <kbd>ctrl</kbd><kbd>enter</kbd> on every page possible.
```
FloEdelmann, line 402:
```suggestion
- [](# "submission-via-ctrl-enter-everywhere") Enables submission via <kbd>ctrl</kbd> <kbd>enter</kbd> on every page possible.
```

source/features/create-release-shortcut.tsx
```diff
@@ -1,12 +1,9 @@
 import select from 'select-dom';
 import * as pageDetect from 'github-url-detection';
 
+import addQuickSubmit from './submission-via-ctrl-enter-everywhere';
```
yakov116, line 4: Lint

darkred: Done, thanks.

fregante (approved):

PR #5612, opened 2022-04-28
Hide text of "Edit Pins" dropdown button in organization repos

cheap-glitch:
The "Pin"/"Unpin" buttons got turned into a dropdown, apparently.

## Test URLs

Any repo page.

## Screenshot

**Before**

before

**After**

after


fregante:
I'm not sure we can hide the dropdown, there appears to be no way to set pins as private in the org profile

cheap-glitch:
> I'm not sure we can hide the dropdown, there appears to be no way to set pins as private in the org profile

In that case, let's just drop the old CSS since it targets a button that doesn't exist anymore.

fregante:
Next up would be:

- https://github.com/refined-github/refined-github/issues/5398

fregante:
We could alternatively just drop the dropdown label (just that one)

cheap-glitch:
I'll fix the spacing of the icon in a separate PR. I noticed the icons of the header buttons all have a different spacing, this can probably be fixed with a single rule in `github-bugs.css`.

PR #5622, opened 2022-05-06
Improve wrapping of repo header buttons on smaller screens

### LINKED ISSUE #5398: Fix repo title wrapping on small screens
### Description

No words honestly. We reject features and drop content from the header to avoid this stuff and they keep adding buttons. Let's limit the buttons to 50% of the screen and then have them wrap

### Screenshot

Screen Shot 10

### Example URL

cheap-glitch:
Fixes #5398 

## Test URLs

[refined-github/refined-github](https://github.com/refined-github/refined-github)
[fregante/webext-dynamic-content-scripts](https://github.com/fregante/webext-dynamic-content-scripts)

## Screenshot

**Before** :-1:

before

**After** :+1:

after

### DIFF
--- a/source/refined-github.css
+++ b/source/refined-github.css
@@ -179,32 +179,20 @@ pr-branches
 }
 
 /* Improve wrapping of repo header buttons on smaller screens #5398 */
-.pagehead-actions {
-	display: flex !important;
-	flex-wrap: wrap !important;
-	gap: 8px;
-}
-
-.pagehead-actions li {
-	margin-right: 0 !important;
-}
-
-@media (max-width: 768px) {
-	/* Put the buttons under the repo title */
-	#repository-container-header > :first-child {
-		flex-direction: column;
-		gap: 8px;
-	}
-}
-
 @media (min-width: 768px) {
-	/* Force the buttons container to wrap instead of the title */
 	#repository-container-header > :first-child > :first-child {
 		flex-shrink: 0 !important;
 	}
 
 	.pagehead-actions {
+		display: flex !important;
 		flex-shrink: 1 !important;
+		flex-wrap: wrap !important;
 		justify-content: flex-end;
+		gap: 8px;
+	}
+
+	.pagehead-actions li {
+		margin-right: 0 !important;
 	}
 }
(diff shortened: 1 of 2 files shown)


cheap-glitch (commented):
Note: as shown in the video above, this PR also restores the buttons on mobile — where they're normally hidden by GitHub.

Not sure if this is a welcomed change, so I don't mind dropping it.

source/refined-github.css
```diff
@@ -177,3 +177,22 @@ pr-branches
...
+		display: flex !important;
+		flex-shrink: 1 !important;
+		flex-wrap: wrap !important;
+		justify-content: flex-end;
+		gap: 8px;
+	}
+
+	.pagehead-actions li {
+		margin-right: 0 !important;
+	}
```
cheap-glitch, line 197: Needed because this is replaced by `gap` on the parent.

cheap-glitch (commented):
I think I've reduced the CSS to the minimum amount necessary. Unfortunately it still requires a bit more than 4 lines.

source/refined-github.css
```diff
@@ -177,3 +177,22 @@ pr-branches
 	min-height: 4em;
 	height: 4em;
 }
+
+/* Improve wrapping of repo header buttons on smaller screens #5398 */
+@media (min-width: 768px) {
+	#repository-container-header > :first-child > :first-child {
+		flex-shrink: 0 !important;
+	}
```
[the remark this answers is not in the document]
fregante: We also have to consider long repository names. `flex-shrink: 3` or higher gives priority to the title without causing the opposite problem.

> not shrink at all

That's the exact reason why this bug exist. This PR causes the opposite problem:

Screen Shot 2

We have to allow wrapping, this is too much code to achieve that

source/refined-github.css
fregante: I'm not sure why but I'm seeing this on Chrome

Screen Shot

cheap-glitch: That's likely because the `display: flex` didn't apply, the selector just needs to be made more specific.

cheap-glitch:
Changed to the "shrink more strategy" and moved the fix to `github-bugs.css`. Tested on Chromium too:

image

cheap-glitch:
Closing in favor of #5622

PR #5623, opened 2022-05-07
Fix GitHub bug: Avoid excessive repo title wrap

fregante:
Fixes https://github.com/refined-github/refined-github/issues/5398
Replaces #5622

## Test URLs

https://github.com/fregante/webext-dynamic-content-scripts

## Before
Screen Shot 9

## After

Screen Shot 10

---

Unfortunately the parent’s flexbox is still pretty bad at sharing the space, but it's unlikely that we can do much about it: Once flexbox has to wrap something, it will drop space **from both**: 1/4 from the title and 3/4 from the actions. This can cause both sides to wrap internally even even the sum of the available space is available from at least one more piece from either side:

Screen Shot 7

### DIFF
--- a/source/features/github-bugs.css
+++ b/source/features/github-bugs.css
@@ -92,6 +92,6 @@
 	}
 
 	.pagehead-actions > li > * {
-		margin: 0 8px 8px 0;
+		margin: 8px 8px 0 0;
 	}
 }
--- a/source/features/github-bugs.css
+++ b/source/features/github-bugs.css
@@ -79,17 +79,19 @@
 }
 
 /* Avoid excessive repo title wrap by deprioritizing the repo header actions #5398 */
-:root .pagehead-actions {
-	flex-shrink: 3 !important;
-	display: inline-flex !important;
-	flex-wrap: wrap;
-	justify-content: flex-end;
-}
+@media (min-width: 768px) { /* Match .d-md-inline-flex */
+	:root .pagehead-actions {
+		flex-shrink: 3 !important;
+		display: inline-flex !important;
+		flex-wrap: wrap;
+		justify-content: flex-end;
+	}
 
-:root .pagehead-actions > li {
-	margin: 0;
-}
+	:root .pagehead-actions > li {
+		margin: 0;
+	}
 
-.pagehead-actions > li > * {
-	margin: 0 8px 8px 0;
+	.pagehead-actions > li > * {
+		margin: 0 8px 8px 0;
+	}
 }
--- a/source/features/github-bugs.css
+++ b/source/features/github-bugs.css
@@ -77,3 +77,16 @@
 .pagehead-actions :is(.btn, summary) .octicon {
 	margin-right: 4px !important;
 }
+
+/* Avoid excessive repo title wrap by deprioritizing the repo header actions #5398 */
+:root .pagehead-actions {
+	flex-shrink: 3 !important;
+}
+
+:root .pagehead-actions > li {
+	margin: 0;
+}
+
+.pagehead-actions > li > * {
+	margin: 0 8px 8px 0;
+}
--- a/source/features/github-bugs.css
+++ b/source/features/github-bugs.css
@@ -81,6 +81,10 @@
 /* Avoid excessive repo title wrap by deprioritizing the repo header actions #5398 */
 :root .pagehead-actions {
 	flex-shrink: 3 !important;
+
+	display: inline-flex !important;
+	flex-wrap: wrap;
+	justify-content: flex-end;
 }
 
 :root .pagehead-actions > li {
--- a/source/features/github-bugs.css
+++ b/source/features/github-bugs.css
@@ -81,7 +81,6 @@
 /* Avoid excessive repo title wrap by deprioritizing the repo header actions #5398 */
 :root .pagehead-actions {
 	flex-shrink: 3 !important;
-
 	display: inline-flex !important;
 	flex-wrap: wrap;
 	justify-content: flex-end;


fregante:
Using flex box on the actions allows us to align them to the right after they wrap:

Screen Shot 5


Unfortunately the parent’s flexbox is still pretty bad at sharing the space, but it's unlikely that we can do much about it: Once flexbox has to wrap something, it will drop space from both: 1/4 from the title and 3/4 from the actions.

source/features/github-bugs.css
```diff
@@ -77,3 +77,19 @@
...
+
+/* Avoid excessive repo title wrap by deprioritizing the repo header actions #5398 */
+:root .pagehead-actions {
+	flex-shrink: 3 !important;
+	display: inline-flex !important;
+	flex-wrap: wrap;
+	justify-content: flex-end;
+}
+
+:root .pagehead-actions > li {
```
fregante, line 89: On non-personal repos there's an empty element (Edit pins) that starts taking up space once we add `margin-bottom`

source/features/github-bugs.css
```diff
@@ -77,3 +77,19 @@
 .pagehead-actions :is(.btn, summary) .octicon {
 	margin-right: 4px !important;
 }
+
+/* Avoid excessive repo title wrap by deprioritizing the repo header actions #5398 */
+:root .pagehead-actions {
```
cheap-glitch, line 82: We still need a media query to keep the buttons hidden on mobile:

```suggestion
/* Avoid excessive repo title wrap by deprioritizing the repo header actions #5398 */
@media (min-width: 768px) {
	:root .pagehead-actions {
```

source/features/github-bugs.css
```diff
@@ -77,3 +77,19 @@
...
+	justify-content: flex-end;
+}
+
+:root .pagehead-actions > li {
+	margin: 0;
+}
+
+.pagehead-actions > li > * {
+	margin: 0 8px 8px 0;
+}
```
cheap-glitch, line 95: Optional but I think using `gap` is a better solution, the intent is clearer and it doesn't create unwanted padding inside the wrapper.

cheap-glitch (changes requested):

cheap-glitch:
Side note: we could probably wrap `ci-link` with the second part of the repo title to avoid this:

source/features/github-bugs.css
fregante: Gap is incompatible with the markup, refer to my review

fregante:
> we could probably wrap ci-link with the second part of the repo title to avoid this:

It has to wrap. Think of the icon as a word of the repo name. Either the icon or the repo name has to wrap. 

What happens if it doesn't wrap? Should the text wrap and the icon stays on the right (middle)? Then there will be space and it won't look better.

If neither wraps, we're back to square one.

cheap-glitch:
> The ideal behavior would be to have it wrap together with a word

Yes that's what I meant, sorry for the confusing wording.

> impossible

Why? We get to decide where it's placed, so we could put it inside the repo name wrapper, instead of after:

```html
<strong itemprop="name" class="mr-2 flex-self-stretch">
    <a data-pjax="#repo-content-pjax-container" href="/refined-github/refined-github">refined-github</a>
	<!-- here -->
</strong>
```

And then make it so that the `<strong>` element never wrap its content.

> it would increase the "seemingly-available space" even more

I'd rather have more blank space than a dangling icon, but I guess that's matter of preference.

source/features/github-bugs.css
cheap-glitch: > Gap is incompatible with the markup, refer to my review

Seems to me like it's perfectly compatible? If you're talking about it not working in Chrome, it was just a matter of increasing the selector's specificity: https://github.com/refined-github/refined-github/pull/5622#issuecomment-1120209012

fregante:
> Edit: it works, but the dropdown button is not clickable due to being inside the link 🥲

Yep, that's why it's "impossible". The alternative is to move the last word out of the link, wrap it with the icon, link it again (just the last word), but then we'd have two separate links for the same repo name. Still not a good compromise.

source/features/github-bugs.css
fregante: ## Your current PR

Screen Shot 13

## My PR

Screen Shot 14


---

This is due to this empty element:

Screen Shot 15

```html
<li>
  <div class="float-left" data-test-selector="pin-repo-button">
    <form
      data-turbo="false"
      action="/fregante/webext-dynamic-content-scripts/profile_pin"
      accept-charset="UTF-8"
      method="post"
    >
      <input
        type="hidden"
        name="authenticity_token"
        autocomplete="off"
      />
      <button
        title="No pin slots remaining in your profile"
        disabled="disabled"
        type="submit"
        data-view-component="true"
        class="btn-sm btn"
      >
        <svg
          aria-hidden="true"
          height="16"
          viewBox="0 0 16 16"
          version="1.1"
          width="16"
          data-view-component="true"
          class="octicon octicon-pin mr-2"
        >
          <path
            fill-rule="evenodd"
            d="M4.456.734a1.75 1.75 0 012.826.504l.613 1.327a3.081 3.081 0 002.084 1.707l2.454.584c1.332.317 1.8 1.972.832 2.94L11.06 10l3.72 3.72a.75.75 0 11-1.061 1.06L10 11.06l-2.204 2.205c-.968.968-2.623.5-2.94-.832l-.584-2.454a3.081 3.081 0 00-1.707-2.084l-1.327-.613a1.75 1.75 0 01-.504-2.826L4.456.734zM5.92 1.866a.25.25 0 00-.404-.072L1.794 5.516a.25.25 0 00.072.404l1.328.613A4.582 4.582 0 015.73 9.63l.584 2.454a.25.25 0 00.42.12l5.47-5.47a.25.25 0 00-.12-.42L9.63 5.73a4.581 4.581 0 01-3.098-2.537L5.92 1.866z"
          ></path></svg
        >Pin
      </button>
    </form>
  </div>
</li>


```

cheap-glitch: Wait, I thought the "Pin"/"Unpin" button was replaced with a dropdown everywhere? But I guess it's only on organization repos then.

That means #5612 needs to be amended.

source/features/github-bugs.css
```diff
@@ -77,3 +77,21 @@
...
 	margin-right: 4px !important;
 }
+
+/* Avoid excessive repo title wrap by deprioritizing the repo header actions #5398 */
+@media (min-width: 768px) { /* Match .d-md-inline-flex */
+	:root .pagehead-actions {
+		flex-shrink: 3 !important;
+		display: inline-flex !important;
+		flex-wrap: wrap;
+		justify-content: flex-end;
```
cheap-glitch, line 87:
```suggestion
		justify-content: flex-end;
		margin: 0 -8px -8px 0; /* Offset the unwanted padding caused by the margins on children elements */
```

cheap-glitch (approved):

source/features/github-bugs.css
fregante: What for? I don't see any unwanted padding

fregante: Ah I'll try it

fregante: This directly overrides the margin we set above, CSS should ideally never undo itself.

I'll use `margin: 8px 8px 0 0;` to replace the existing value

yakov116:
Great Team work!

PR #5625, opened 2022-05-11
Restore `expand-all-hidden-comments` feature

### LINKED ISSUE #4921: `expand-all-hidden-comments` doesn't work for PR conversations
### Please ensure:

### Include in this issue:

- [ ] Console errors, if any

### Description

I would expect that the `expand-all-hideen-comments` feature to work on expanding large sections of hidden conversations on PRs. But it does not, it only expands 20 at a time.

### Example URL

https://github.com/lensapp/lens/pull/3910

### Browser(s) used

firefox

cheap-glitch:
Fixes #4921 

## Test URLs

1. Go to https://togithub.com/lensapp/lens/pull/3910 and scroll down to the "131 hidden items / Load more" button.
2. Alt-click it. Every comment in the main thread should load.
3. Scroll down until you see a review thread with a "44 hidden comments" button.
4. Alt-click it. Every review comment in the thread should load.

## Screenshot

https://user-images.githubusercontent.com/46634000/168293483-7931ccdb-f632-4e9e-8882-f8b08538dc48.mp4

### DIFF
--- a/source/features/expand-all-hidden-comments.tsx
+++ b/source/features/expand-all-hidden-comments.tsx
@@ -1,4 +1,5 @@
 import select from 'select-dom';
+import oneEvent from 'one-event';
 import delegate from 'delegate-it';
 import * as pageDetect from 'github-url-detection';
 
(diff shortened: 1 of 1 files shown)


source/features/expand-all-hidden-comments.tsx
```diff
@@ -1,32 +1,48 @@
...
+
+function submitNewPaginationForm(wrapper: Element): void {
+	const wrapperSelector = wrapper.classList.contains('TimelineItem-body') ? ':scope' : ':scope > #js-progressive-timeline-item-container:last-child';
+	const paginationButton = select(`${wrapperSelector} > ${paginationButtonSelector}`, wrapper);
+	if (!paginationButton) {
+		return;
+	}
+
+	const subWrapper = paginationButton.form!.parentElement!;
+	onPaginationFormLoad(paginationButton.form!, () => {
```
fregante, line 21: I'd use `await oneEvent()` in a loop instead. All of these listeners will stay attached and pile up.

cheap-glitch: > All of these listeners will stay attached and pile up.

They're attached to forms that then get replaced with the loaded content, so it should be fine I think?

That said a loop is usually easier to understand than recursive calls, so I'll change it to that.

source/features/expand-all-hidden-comments.tsx
```diff
@@ -1,32 +1,36 @@
...
-	form.action = url.href;
+	while (paginationButton) {
+		// eslint-disable-next-line no-await-in-loop
+		await oneEvent(paginationButton.form!, 'page:loaded');
+		if (isExpandingMainThread) {
+			// Pagination forms in the main thread load their content in a nested wrapper
+			wrapper = wrapper.lastElementChild!;
+		}
+
+		paginationButton = select(`:scope > ${paginationButtonSelector}`, wrapper);
```
fregante, line 27: `:scope` is likely not needed here. It only changes behavior with a Dom like this:


```
.outer
	.inner
     button
```

`select('.outer button', inner)` selects the button because on its own the button _does_ match the selector.


`select(':scope .outer button', inner)` does not select the button anymore.

Is there a situation where `.ajax-pagination-form` is a parent of `:scope` on this line?

fregante (approved):
Looks great!

source/features/expand-all-hidden-comments.tsx
cheap-glitch: `:scope` is only there to allow using the "direct child combinator" (`>`). The code assumes in several places that the new pagination form is a direct descendant of the current wrapper, so I think the selector should reflect that.

Nokel81:
Woot! Looking forward to the release with this.

PR #5631, opened 2022-05-18
Drop `hide-tips` feature

fregante:
- Closes https://github.com/refined-github/refined-github/issues/4574

I don’t think it's worth hiding these tips. GitHub’s interface is way more complex than when this feature was added, so hiding "all tips" to save space is a uphill battle that does not help new (and old) GitHub users.


yakov116:
If I'm not mistaken this was from our first features!

fregante:
Indeed this comes straight from 1999

PR #5641, opened 2022-05-20
Ensure `restore-file` appears after navigating to the Files tab

fregante:
- Closes https://github.com/refined-github/refined-github/issues/5606

## Test URLs

1. Open any PRs in https://github.com/refined-github/refined-github/pulls
2. Go to Files tab
3. Look for "Restore file"

## Video with before and after

https://user-images.githubusercontent.com/1402241/169457069-b05f0a28-3350-4a19-9f0d-5903808effd9.mov

### DIFF
--- a/source/features/linkify-code.tsx
+++ b/source/features/linkify-code.tsx
@@ -43,6 +43,7 @@ void features.add(import.meta.url, {
 	exclude: [
 		pageDetect.isGist,
 	],
+	deduplicate: 'has-rgh-inner',
 	init,
 }, {
 	include: [
--- a/source/features/linkify-code.tsx
+++ b/source/features/linkify-code.tsx
@@ -43,7 +43,6 @@ void features.add(import.meta.url, {
 	exclude: [
 		pageDetect.isGist,
 	],
-	deduplicate: 'has-rgh-inner',
 	init,
 }, {
 	include: [
--- a/source/features/restore-file.tsx
+++ b/source/features/restore-file.tsx
@@ -139,5 +139,6 @@ void features.add(import.meta.url, {
 		pageDetect.isPRFiles,
 		pageDetect.isPRCommit,
 	],
+	deduplicate: 'has-rgh-inner',
 	init,
 });


source/features/linkify-code.tsx
```diff
@@ -43,6 +43,7 @@ void features.add(import.meta.url, {
 	exclude: [
 		pageDetect.isGist,
 	],
+	deduplicate: 'has-rgh-inner',
```
cheap-glitch, line 46: This fixes a similar bug with `linkify-code`, but since it runs on more pages than `restore-file` it might cause duplication bugs elsewhere.

Should we split the init and use `has-rgh-inner` only with `isPRFiles` or is that unnecessary?

cheap-glitch (approved):

source/features/linkify-code.tsx
fregante: It's worth testing other pages. If this breaks the feature elsewhere it's not worth the change as is

fregante: I dropped it for now, I think the fix there would be to move the deduplication logic out of the manager and into the feature itself. I think it'll just have to use `deduplication:false` since we already use `:not(linkifiedURLClass)`

fregante:
I'll leave this PR open until we 22.5.22 just in case we see new features that need the same fix

cheap-glitch:
> I'll leave this PR open until we 22.5.22 just in case we see new features that need the same fix

Too late but `extend-diff-expander` has the same problem. Might be worth checking all features acting on `isPRFiles`.

fregante:
> Too late

Sorry I had itchy hands 😂

fregante:
> `extend-diff-expander` has the same problem

If it has the same problem, it doesn't;t have the same solution since the feature already uses `has-rgh-inner`

https://github.com/refined-github/refined-github/blob/3194f0dcc22b687c7f0a1ba2f2e25fb5e46aab3a/source/features/extend-diff-expander.tsx#L25

PR #5642, opened 2022-05-20
Don't link to first comment in `highest-rated-comment`

fregante:
- Fixes https://github.com/refined-github/refined-github/issues/5577
- Restores unwanted change from https://github.com/refined-github/refined-github/pull/5357#discussion_r796514659

The feature seems to be broken? Can't test at all

## Test URLs

- https://togithub.com/vercel/next.js/issues/31054

<table><caption>Both screenshots include a fix that restores the feature altogether</caption>
<tr>
	<th>Before
	<th>After
<tr>
	<td>Screen Shot 16
	<td>Screen Shot 15
</table>

### DIFF
--- a/source/features/highest-rated-comment.tsx
+++ b/source/features/highest-rated-comment.tsx
@@ -62,9 +62,11 @@ function highlightBestComment(bestComment: Element): void {
 }
 
 function linkBestComment(bestComment: HTMLElement): void {
-	const firstTimelineItem = select('#js-timeline-progressive-loader')!.nextElementSibling!;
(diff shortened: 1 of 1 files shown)


cheap-glitch:
> The feature seems to be broken?

I think it's just the reaction selectors that need updating. Seems like GitHub moved the list of people who reacted to a separate pop-up element, like on release pages.

Changing `reacted` to `react` in `[aria-label*="reacted with thumbs up"]` et al should work. The `*` isn't even really needed anymore.

PR #5648, opened 2022-05-22
Don't run Refined GitHub on login pages

fregante:
- Fixes https://github.com/refined-github/refined-github/issues/5386

@kidonng @cheap-glitch can you think of any more static URLs that we can exclude this way?

## Test URLs

```
// https://github.com/login <-- nothing runs
// https://github.com/logins <-- works as usual
```

## Screenshot

Screen Shot 13

Screen Shot

### DIFF
--- a/source/manifest.json
+++ b/source/manifest.json
@@ -45,6 +45,9 @@
 				"https://github.com/*",
 				"https://gist.github.com/*"
 			],
+			"exclude_matches": [
+				"https://*/login/*"
+			],
 			"css": [
 				"refined-github.css"
 			],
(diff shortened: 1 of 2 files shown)


source/manifest.json
```diff
@@ -45,6 +45,9 @@
 				"https://github.com/*",
 				"https://gist.github.com/*"
 			],
+			"exclude_matches": [
+				"https://*/login/*"
+			],
```
fregante, line 50: I initially thought about add this pageDetect inside `features.add`, but this is better for such static URLs

```js
const isLogin = (): boolean => location.pathname.split('/')[1] === 'login' || pageDetect.isPasswordConfirmation();
```

cheap-glitch:
> can you think of any more static URLs that we can exclude this way?

Maybe `https://github.com/about` and `https://github.com/pricing`?

fregante:
> > can you think of any more static URLs that we can exclude this way?
> 
> Maybe `https://github.com/about` and `https://github.com/pricing`?

They have the site header so we still have some applicable features

PR #5652, opened 2022-05-22
Meta: Add link to TSX file when looking at CSS feature 

fregante:
## Test URLs

- Yes: https://github.com/refined-github/refined-github/blob/main/source/features/show-names.css
- No (CSS only): https://github.com/refined-github/refined-github/blob/main/source/features/clean-commit-form.css

## Screenshot

<table>
<tr>
	<td>Screen Shot 20
<td>🔗➡️
	<td>Screen Shot 21

</table>

### DIFF
--- a/source/features/rgh-feature-descriptions.tsx
+++ b/source/features/rgh-feature-descriptions.tsx
@@ -49,6 +49,11 @@ async function init(): Promise<void | false> {
 					<div dangerouslySetInnerHTML={{__html: feature.description}} className="text-bold"/>
 					<div className="no-wrap">
 						<a href={conversationsUrl.href} data-pjax="#repo-content-pjax-container">Conversations</a>
+						{
+							location.pathname.endsWith('css')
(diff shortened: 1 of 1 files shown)


source/features/rgh-feature-descriptions.tsx
```diff
@@ -49,6 +49,11 @@ async function init(): Promise<void | false> {
 					<div dangerouslySetInnerHTML={{__html: feature.description}} className="text-bold"/>
 					<div className="no-wrap">
 						<a href={conversationsUrl.href} data-pjax="#repo-content-pjax-container">Conversations</a>
+						{
+							location.pathname.endsWith('css')
```
cheap-glitch, line 53:
```suggestion
							location.pathname.endsWith('.css')
```

cheap-glitch (approved):

source/features/rgh-feature-descriptions.tsx
fregante: This already applies to features exclusively, so unless we have an extensionless feature named `hellocss` this change isn't necessary

PR #5653, opened 2022-05-22
Make `toggle-files-button` unforgettable

fregante:
- Fixes https://github.com/refined-github/refined-github/issues/5524
- Fixes https://github.com/refined-github/refined-github/issues/4402

cc @sideshowbarker because this issue was reported in https://github.com/refined-github/refined-github/issues/4402

## Test URLs

1. Go to https://github.com/refined-github/refined-github
2. Hide files
3. Refresh the page

## Screenshot

Screen Shot 23

## Video

- The notice only appears when the list is hidden automatically so the user doesn't forget about it
- The notice is gone after the first interaction
- The notice is only clickable because the user might click it, but we don't want to actually show a hand because they should prefer clicking the actual button instead

https://user-images.githubusercontent.com/1402241/169680332-6b02c852-be43-44b9-a01a-a7c634aa7e02.mov

### DIFF
--- a/source/features/toggle-files-button.tsx
+++ b/source/features/toggle-files-button.tsx
@@ -44,8 +44,12 @@ async function init(): Promise<Deinit[]> {
 	if (await cache.get<boolean>(cacheKey)) {
 		repoContent.classList.add('rgh-files-hidden');
 		select('#files')!.after(
-			<div className="py-1 text-right text-small rgh-files-hidden-notice" style={{paddingRight: '19px'}}>
-				Files hidden <ArrowUpIcon className="v-align-middle"/>
+			// 19px align this icon with the <UnfoldIcon/> above it
+			<div
+				className="py-1 text-right text-small color-fg-muted rgh-files-hidden-notice"
+				style={{paddingRight: '19px'}}
+			>
+				The file list was collapsed via Refined GitHub <ArrowUpIcon className="v-align-middle"/>
 			</div>,
 		);
 	}
--- a/source/features/toggle-files-button.tsx
+++ b/source/features/toggle-files-button.tsx
@@ -43,10 +43,10 @@ async function init(): Promise<Deinit[]> {
 
 	if (await cache.get<boolean>(cacheKey)) {
 		repoContent.classList.add('rgh-files-hidden');
-		select('#files')!.after(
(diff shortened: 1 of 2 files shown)


source/features/toggle-files-button.tsx
```diff
@@ -33,17 +33,26 @@ function addButton(): void {
...
+	select('.rgh-files-hidden-notice')?.remove();
 }
 
 async function init(): Promise<Deinit[]> {
 	const repoContent = (await elementReady('.repository-content'))!;
+
 	if (await cache.get<boolean>(cacheKey)) {
 		repoContent.classList.add('rgh-files-hidden');
+		select('#files')!.after(
+			<div className="py-1 text-right text-small rgh-files-hidden-notice" style={{paddingRight: '19px'}}>
```
cheap-glitch, line 47: Why 19 pixels exactly? Could this be replaced e.g. by `.pr-3` (which would be 16px)?

fregante:
```suggestion
			// 19px align this icon with the UnfoldIcon above it
			<div
				className="py-1 text-right text-small rgh-files-hidden-notice"
				style={{paddingRight: '19px'}}
			>
```

source/features/toggle-files-button.tsx
```diff
@@ -33,17 +33,26 @@ function addButton(): void {
...
+				Files hidden <ArrowUpIcon className="v-align-middle"/>
```
yakov116, line 48: Maybe write, `unhide files`?

fregante: This is meant to be a notice of the current state of the UI, not a call to action. Maybe it can be extended to "The file list was collapsed via Refined GitHub" and make it in gray text.

cheap-glitch: +1 for a longer and gray text

fregante:
Screen Shot 6
Screen Shot 7

fregante:
`wait-for-build` is basically broken for me. 🥲 Should we hotfix it?

panva:
I find this change distracting, plus I happen to run (but did not figure out the steps to reproduce) into a case where upon navigation the note is added twice (two identical lines).

PR #5656, opened 2022-05-22
Add `collapse-duplicated-reviews` feature

### LINKED ISSUE #5394: Fold duplicate review threads
### Description

When you comment on a existing thread while creating a review, the comment appears in both the thread and your reviews (as a separate thread). The latter one become duplicate and potentially confusing since it loses its context.

We can try to fold these threads by default, to reduce the clutter.

Implementation-wise, we just need to look for review comments with the same `[id]`.

### Screenshot

Screen Shot 2022-02-10 at 22 20 38

### Example URL

https://github.com//refined-github/refined-github/pull/5393#discussion_r803596694

cheap-glitch:
Fixes #5394 

## Test URLs

https://github.com/refined-github/sandbox/pull/10#pullrequestreview-958560681

## Screenshot

https://user-images.githubusercontent.com/46634000/169702068-8bfe4f63-09b6-45c1-b2a1-0dee4d717f86.mp4


cheap-glitch (commented):
Turns out we don't even need a click listener, since review threads are already `<details>` elements.

The style of the "label" replicates the native "Show/Hide resolved" link-button.

fregante:
I'd rather collapse the whole thread into the title (so it's visible) and link the title to the real comment. 

With the current PR, _reacting_ to a comment means uncollapsing it _and then_ clicking the permalink. Let's skip one step.

cheap-glitch:
> I'd rather collapse the whole thread into the title

By "title", do you mean the text next to the "review event"? But what happens if the review is made up of two duplicated comments, each responding to a different review thread?

fregante:
> By "title", do you mean the text next to the "review event"?

I mean this:

Screen Shot 10

instead of this:

Screen Shot 11

The latter is not actionable and it's out of context. That comment should not appear there, but the review should still keep the link to the real review comment, as shown in the first screenshot.

cheap-glitch:
> By title I mean filename

This is already what happens, see video above. We could avoid showing the "Show"/"Hide" text & icon, but since we hide something GitHub shows by default we should probably make it obvious it can be expanded.

Unless you want to drop the review thread completely?

> and link the title to the real comment

+1

source/features/collapse-duplicated-reviews.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+		const details = comment.closest('details')!;
+		details.open = false;
+		details.classList.add('rgh-duplicated-review');
+
+		// Change the link in the thread header to point to the original comment instead of the file diff
+		const commentUrl = new URL(location.href);
+		commentUrl.hash = comment.id;
+		const threadHeaderLink = select('summary a', details)!;
+		threadHeaderLink.href = commentUrl.toString();
```
fregante, line 22:
```suggestion
		threadHeaderLink.href = commentUrl.href;
```

source/features/collapse-duplicated-reviews.css
```diff
@@ -0,0 +1,33 @@
...
+	right: 94px;
+}
+
+.rgh-duplicated-review[open] summary > .d-flex::before {
+	/* https://primer.style/octicons/fold-16 */
+	content: url('data:image/svg+xml;utf8,%3Csvg%20xmlns%3D%22http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%22%20viewBox%3D%220%200%2016%2016%22%20width%3D%2216%22%20height%3D%2216%22%3E%3Cpath%20fill%3D%22%2357606a%22%20d%3D%22M10.896%202H8.75V.75a.75.75%200%2000-1.5%200V2H5.104a.25.25%200%2000-.177.427l2.896%202.896a.25.25%200%2000.354%200l2.896-2.896A.25.25%200%200010.896%202zM8.75%2015.25a.75.75%200%2001-1.5%200V14H5.104a.25.25%200%2001-.177-.427l2.896-2.896a.25.25%200%2001.354%200l2.896%202.896a.25.25%200%2001-.177.427H8.75v1.25zm-6.5-6.5a.75.75%200%20000-1.5h-.5a.75.75%200%20000%201.5h.5zM6%208a.75.75%200%2001-.75.75h-.5a.75.75%200%20010-1.5h.5A.75.75%200%20016%208zm2.25.75a.75.75%200%20000-1.5h-.5a.75.75%200%20000%201.5h.5zM12%208a.75.75%200%2001-.75.75h-.5a.75.75%200%20010-1.5h.5A.75.75%200%200112%208zm2.25.75a.75.75%200%20000-1.5h-.5a.75.75%200%20000%201.5h.5z%22%3E%3C%2Fpath%3E%3C%2Fsvg%3E');
```
fregante, line 21: Not important due to my review, but: I'm not sure I like maintaining these two separately from all the others, especially because the code is already altering the specific DOM so we're 2 lines away from just using `.append(<FoldIcon/>)`

source/features/collapse-duplicated-reviews.tsx
```diff
@@ -0,0 +1,32 @@
...
+	}
+}
+
+void features.add(import.meta.url, {
+	include: [
+		pageDetect.isPRConversation,
+	],
+	deduplicate: 'has-rgh-inner',
+	init,
+});
```
fregante, line 32: Can we add some test URLs at the bottom of the file? They don't need to follow a specific style

fregante (changes requested):
> we should probably make it obvious it can be expanded.

At what pro? We're hiding it because it's duplicate content on the page. What's the point of adding a button to show something we deem unworthy? Even GitHub adding a `details` there is super pointless/OCD to begin with.

This is what I think should look like. I achieved this with something like:

```js
const link = select('a', details)
details.replaceWith(link)
link.classList.add(...details.classList, …otherClasses)
```

cheap-glitch:
@fregante wdyt about this:

image

Instead of using the file name we make it clear this is a link to another comment (plus it's one less `select` because we don't need to grab the `<summary>`).

source/features/collapse-duplicated-reviews.tsx
```diff
@@ -0,0 +1,35 @@
...
+		// Change the <details> element to a link to the original comment
+		const details = comment.closest('details')!;
+		const commentUrl = new URL(location.href);
+		commentUrl.hash = comment.id;
+		details.replaceWith(
+			<a
+				className="d-block py-2 px-3 mb-3 border rounded-2 color-bg-subtle text-mono text-small"
+				href={commentUrl.href}
+			>
+				#{comment.id}
```
fregante, line 23: I'd rather show the file name, this doesn't make sense to a user

source/features/collapse-duplicated-reviews.tsx
```diff
@@ -0,0 +1,35 @@
...
+	const singleReviewThreadComments = select.all('.review-comment.unminimized-comment:only-child');
+	for (const comment of singleReviewThreadComments) {
+		if (!select.exists(`.review-comment#${comment.id}:not(:only-child)`)) { // The review comment is not a duplicate
+			continue;
+		}
+
+		// Change the <details> element to a link to the original comment
+		const details = comment.closest('details')!;
+		const commentUrl = new URL(location.href);
+		commentUrl.hash = comment.id;
```
fregante, line 17: What happens if the comment appears on an old/hidden review thread? 

Edit: answering myself: GitHub detects the hashchange and loads the hidden thread 👍

source/features/collapse-duplicated-reviews.tsx
```diff
@@ -0,0 +1,35 @@
...
+		details.replaceWith(
+			<a
+				className="d-block py-2 px-3 mb-3 border rounded-2 color-bg-subtle text-mono text-small"
+				href={commentUrl.href}
```
fregante, line 21:
```suggestion
				href={'#' + comment.id}
```

source/features/collapse-duplicated-reviews.tsx
cheap-glitch: How about an informative text instead, like "Duplicated comment. Click to jump to original thread"?

I just think showing a file name when this isn't a link to a file is a bit weird :thinking:

fregante: Let we do in `short-repo-url`:

> file.tsx (comment)

Let's not overthink this 😁

fregante:
Closing until you have time to pick it up. This feature is just:

```js
observe('collapsedReview', review => {
	if (isDuplicate(review)) {
	    const link = select('a', review)
	    details.replaceWith(link)
	    link.classList.add(...details.classList, ...otherClasses)
	}
})
```

PR #5658, opened 2022-05-23
Make title change events more readable by stacking them

cheap-glitch:
https://github.com/refined-github/refined-github/issues/5657#issuecomment-1134077169

## Test URLs

```
https://togithub.com/refined-github/refined-github/pull/5626#event-6609835934
```

## Screenshot

**Before**

image

**After**

image

### DIFF
--- a/source/refined-github.css
+++ b/source/refined-github.css
@@ -179,6 +179,6 @@ pr-branches
 }
 
 /* Align vertically the old and new titles in title edition events https://github.com/refined-github/refined-github/issues/5657#issuecomment-1134077169 */
-.TimelineItem[id^="event-"] .TimelineItem-body :is(ins, del) {
+.TimelineItem[id^='event-'] .TimelineItem-body :is(ins, del) {
 	display: block !important;
 }
--- a/source/refined-github.css
+++ b/source/refined-github.css
@@ -178,7 +178,7 @@ pr-branches
 	height: 4em;
 }
 
-/* Align vertically the old and new titles in title edition events https://github.com/refined-github/refined-github/issues/5657#issuecomment-1134077169 */
+/* Make title change events more readable by stacking them #5658 */
 .TimelineItem[id^='event-'] .TimelineItem-body {
 	display: flex;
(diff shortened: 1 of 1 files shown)


FloEdelmann:
Is it possible (and worth the effort) to move the date into the first line to save some vertical space?

grafik

fregante:
> Is it possible (and worth the effort) to move the date into the first line to save some vertical space?

Should be feasible with flex:

1. set to column
2. set min width to 100% for the titles
3. change order of time to 1, I think 
4. ???
5. Merge PR

fregante (approved):

cheap-glitch:
Done!

image

Couldn't use `flex-order: column` because of the DOM, but `flex-wrap: wrap` + `flex: 1 0 100%` does the trick.

ajmeese7 (approved):
Great stuff, looks good to me!

fregante:
Looks good! 

Screen Shot

fregante (changes requested):
This breaks other events. Test on https://github.com/refined-github/refined-github/pull/4030
We should also inline this URL for future reference

Before

Screen Shot 1
Screen Shot 6

After


Screen Shot 2
Screen Shot 5

cheap-glitch:
> This breaks other events

Fixed it by excluding the branch-related events and correcting the labels:

image

image

(The vertical alignment of the label is not exactly the same, but if anything I find them better aligned now).

cheap-glitch:
> Can we be sure we're not messing up the order of any event?

Only `<ins>` and `<del>` elements have they order changed. Afaict the "title change" events are the only ones with those.

The problem is that we have to make `.TimelineItem-body` a flex container regardless, and this is what breaks some other events.

fregante:
> not the reordering

`order` is not what's breaking it indeed, but it's the reordering that requires that we use `flex`. Without it, it's a simple `event (ins, del) {display: block}`, right?

cheap-glitch:
Closing in favor of #5703

PR #5683, opened 2022-06-05
Fix `forked-to` spacing

shaunbharat:
## Does this PR close/fix an existing issue?

There is a gap between the fork button of a GitHub repository and the forked-to button added to the page from refined-github. This PR removes that gap, making it look neater.

## Comparison (Images)

Before
before

After
after

## Test URLs

I've tested this on the [refined-github](https://github.com/refined-github/refined-github) repository after making a fork.


source/features/forked-to.tsx
```diff
@@ -61,6 +61,7 @@ async function updateUI(forks: string[]): Promise<void> {
 				<ChevronRightIcon className="v-align-text-top"/>
 			</a>,
 		);
+		forkButton.style.cssText = 'margin-right:0px !important';
```
shaunbharat, line 64: I used `style.cssText` to be able to set the styling as important, to override the other CSS rules. Setting `cssText` will overwrite the current element's inline styling - but GitHub uses stylesheets, meaning that there's no CSS that will be overwritten.

shaunbharat (commented):
The fork button's style of `margin-right` is set to `0px !important` to remove the gap between the fork button and the forked-to button. It overrides the element's styling, if any.

fregante:
What version and browser are you using? I didn't see that issue. Is it possible that my hotfix for the next version is breaking it? Either way this should not happen to begin with, other buttons behave correctly even if they're "BtnGroup"

shaunbharat:
I'm using Chrome with the latest version on Windows 10. I don't think I would be having anything client-side interfering with the layout, but I could check my currently installed extensions.

I apologize, should have probably opened a new issue first and ask if other people had this too. I'll mark this PR as a draft for now.

fregante:
This appears to be due to https://github.com/refined-github/refined-github/pull/5665, which has been hot-fixed to 22.5.22

The fix for this is to use `wrapButtons` because for some reason it's missing the `.BtnGroup` element

Screen Shot 3
 

I'll temporarily hotfix this issue too, but we need to fix it via JS too

shaunbharat:
> The fix for this is to use wrapButtons because for some reason it's missing the .BtnGroup element

Do you mean `groupButtons` from `source/github-helpers/group-buttons.tsx`?

Replacing the line I added, with `groupSiblings(forkButton);` works with fixing the space between the fork buttons. Unless there's a better fix, or it belongs somewhere else.

fregante:
Yes that's it, but the btnGroup classes we currently add manually need to be dropped since that's already taken care by groupButtons. We can probably even  drop the  .before() call because groupButtons already append all the parameters to a single parent

shaunbharat (commented):
Uses `groupButtons` to group the 'fork' and the 'forked-to' button.

The classes `BtnGroup-parent` and `BtnGroup-item` that are currently added manually, are removed since `groupButtons` already does that.

The forked-to button can end up being either a single button that links to the fork, or a dropdown that lists the different forks. In either case, `groupButtons` groups the fork button and the button with `rgh-forked-button` (which can be a single button or the dropdown).

The class `rgh-forked-button` is switched from the child of the new element's dropdown `<summary>` to the new element's parent `<details>`.

shaunbharat:
Let me know if these changes aren't adequate or needs improvement/other fixes.

source/features/forked-to.tsx
```diff
@@ -101,6 +102,8 @@ async function updateUI(forks: string[]): Promise<void> {
 			</details>,
 		);
 	}
+
+	groupButtons([forkButton, select('.rgh-forked-button')!]);
```
fregante, line 106: I think you can alternatively just use:

```suggestion
	groupSiblings(forkButton);
```

fregante (approved):
If it looks correctly for you, it's the right fix 👌 🎉

fregante (approved):
duplicate review ftw

source/features/forked-to.tsx
shaunbharat: I had tested that before - it only worked in the case of the single button. With the dropdown, it looked really messy. The button ended up being on a new line entirely. `groupSiblings` didn't add the dropdown element to the `BtnGroup`, but it worked with the single button.

Using `groupButtons` with the fork button and the class, works for both the single button and the dropdown though

source/features/forked-to.tsx
```diff
@@ -9,6 +9,7 @@ import features from '.';
 import fetchDom from '../helpers/fetch-dom';
 import GitHubURL from '../github-helpers/github-url';
 import {getUsername, getForkedRepo, getRepo} from '../github-helpers';
+import {groupButtons} from '../github-helpers/group-buttons';
```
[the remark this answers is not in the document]
shaunbharat: Thanks, fixed it

yakov116:
@shaunbharat no need to force push we squash and merge

shaunbharat:
Hey, sorry forgot about that haha! I think you can tell I'm a bit new to contributing using git.

source/features/forked-to.tsx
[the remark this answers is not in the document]
yakov116: This was not fixed

shaunbharat:
Thanks for the last commit, was having trouble updating the pull request from my fork

yakov116:
@shaunbharat because are using the "main" branch
When making a pr dont use the main branch

fregante (changes requested):
@shaunbharat 



I'm trying to push some changes but I think you disabled it

shaunbharat:
@fregante I've had that enabled from the start. I'm looking for the checkbox now, and I can't find it at all. Sorry this pull request was messy, do you think I should open a new one? I put all the changes on my fork on a new branch like @yakov116 said. My changes in my fork's main branch isn't synced with this pull request either

shaunbharat:
New pull request #5691

PR #5684, opened 2022-06-06
Reduce `mark-private-orgs` cache validity

ncfavier:
I was a bit surprised to see an org still marked as private on my profile a week after I set it to public. Apparently this was missed by https://github.com/refined-github/refined-github/issues/3302. One day seems reasonable.

Also it would be nice to have a way to manually invalidate the cache for a given feature.


fregante:
Thanks! We should alternatively use staleWhileRevalidate. You can look at the other features to see what values are expected for both

fregante (approved):

fregante:
@ncfavier thank you!

PR #5654, opened 2022-05-22
Lint

fregante:
# 1
- Close https://github.com/refined-github/refined-github/issues/5367
- Test on https://github.com/refined-github/refined-github/tree/main/source

PR #5685, opened 2022-06-06
Improve `Deinit` type

cheap-glitch:
Follow-up of #5654 

1. Make `Deinit` actually useful
2. Allow it to be an array
3. Enable `no-redundant-type-constituents` lint rule


fregante (approved):

fregante:
yolo-default

PR #5706, opened 2022-06-19
Avoid duplication in `toggle-files-button`

fregante:
- Closes https://github.com/refined-github/refined-github/issues/5692
- Closes https://github.com/refined-github/refined-github/issues/5693

URL: https://github.com/refined-github/refined-github

https://user-images.githubusercontent.com/1402241/174477063-52f36daa-e552-4552-ada6-a51fffe5718a.mov

### DIFF
--- a/source/helpers/attach-element.ts
+++ b/source/helpers/attach-element.ts
@@ -1,0 +1,30 @@
+import select from 'select-dom';
+
+type Position = 'append' | 'prepend' | 'before' | 'after';
+
+interface Attachment {
+	anchor: Element | string | undefined; // `undefined` allows `select(nonExistent)` without requiring `!`
+	getNewElement: () => Element;
+	className: string;
+	position: Position;
+}
+
+export default function attach({
+	anchor,
+	getNewElement,
+	className,
+	position,
+}: Attachment): void {
+	anchor = typeof anchor === 'string' ? select(anchor) : anchor;
+	if (!anchor) {
+		throw new Error('Element not found');
+	}
+
+	if (select.exists('.' + className, anchor.parentElement ?? anchor)) {
+		return;
+	}
+
+	const ship = getNewElement();
+	ship.classList.add(className);
+	anchor[position](ship);
+}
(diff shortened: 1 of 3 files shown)


source/helpers/attach-element.ts
```diff
@@ -0,0 +1,45 @@
...
+	position: Position;
+	allowMissingAnchor?: boolean;
+}
+
+// Get a unique position in the code
+function getSnapshotUUID(index = 3): string {
+	return hashString(new Error('Get stack').stack!.split('\n')[index]);
+}
+
+export default function attach<NewElement extends Element>({
```
yakov116, line 21: Can we call this function `attachElement`?

source/helpers/attach-element.ts
```diff
@@ -0,0 +1,45 @@
+import select from 'select-dom';
+
```
yakov116, line 2:
```suggestion
```

source/helpers/attach-element.ts
```diff
@@ -0,0 +1,45 @@
...
+	// eslint-disable-next-line @typescript-eslint/ban-types --  Allows dom traversing without requiring `!`
+	anchor: Element | string | undefined | null;
+	getNewElement: () => NewElement;
+	className?: string;
+	position: Position;
+	allowMissingAnchor?: boolean;
+}
+
+// Get a unique position in the code
+function getSnapshotUUID(index = 3): string {
```
yakov116, line 17: I am not 100% sure what you are doing here. But can we use https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID ?

fregante: It's not random, it's an automatic identifier/key of the call. It's the same every time it's called (even across sessions), but it's different for every caller.

The index defines which call point is considered (from the stack).

Maybe it should just be named `callerID` 🙂 


This lets us avoid custom classes unless necessary (in this case they're necessary though)

yakov116: This was a really good lesson something I never knew.
Thanks

PR #5709, opened 2022-06-20
Don't show single non-boundary spaces in `show-whitespace`

fregante:
- Closes https://github.com/refined-github/refined-github/issues/4937
- Based on https://github.com/refined-github/refined-github/pull/5710
  - that should be merged first

## Test URLs

- https://github.com/refined-github/sandbox/pull/18

## Before

Screen Shot 3

## After

Screen Shot 2

PR #5710, opened 2022-06-20
Meta: Add tests for `show-whitespace`

fregante:
- Preparing for #5709

### DIFF
--- a/source/features/show-whitespace.tsx
+++ b/source/features/show-whitespace.tsx
@@ -1,67 +1,20 @@
 import './show-whitespace.css';
-import React from 'dom-chef';
 import select from 'select-dom';
 import delegate from 'delegate-it';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
-import getTextNodes from '../helpers/get-text-nodes';
 import onNewComments from '../github-events/on-new-comments';
 import {onDiffFileLoad} from '../github-events/on-fragment-load';
 import {codeElementsSelector} from '../github-helpers/dom-formatters';
-
-// `splitText` is used before and after each whitespace group so a new whitespace-only text node is created. This new node is then wrapped in a <span>
-function showWhiteSpacesOn(line: Element): void {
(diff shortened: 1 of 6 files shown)


source/helpers/show-whitespace-on-line.test.ts
```diff
@@ -0,0 +1,224 @@
...
+function assertHighlighted(t: ExecutionContext, actual: string, expected: string): void {
+	t.is(
+		show(actual, false),
+		expected,
+	);
+}
+
+test('showWhiteSpacesOnLine raw', t => {
+	assert(
+		t,
```
fregante, line 58: I can't find a way to use `t.is` without passing it here. I can't find a way to create custom global assertions either, like:

```js
t.transformsTo(actual, expected)
```

https://jestjs.io/docs/expect#expectextendmatchers

source/helpers/show-whitespace-on-line.test.ts
```diff
@@ -0,0 +1,224 @@
...
+		'',
+		'',
+	);
+	assert(
+		t,
+		' ',
+		'•',
```
fregante, line 65: They're specifically set vertically to help visually evaluate the substitutions too

source/features/show-whitespace.tsx
```diff
@@ -1,67 +1,20 @@
...
 
 import features from '.';
-import getTextNodes from '../helpers/get-text-nodes';
 import onNewComments from '../github-events/on-new-comments';
 import {onDiffFileLoad} from '../github-events/on-fragment-load';
 import {codeElementsSelector} from '../github-helpers/dom-formatters';
-
-// `splitText` is used before and after each whitespace group so a new whitespace-only text node is created. This new node is then wrapped in a <span>
-function showWhiteSpacesOn(line: Element): void {
-	const shouldAvoidSurroundingSpaces = Boolean(line.closest('.blob-wrapper-embedded')); // #2285
```
fregante, line 15: `showWhiteSpacesOn` was moved without change, except for extracting `shouldAvoidSurroundingSpaces`

PR #5711, opened 2022-06-20
Update dependencies

fregante:


source/manifest.json
```diff
@@ -4,11 +4,11 @@
 	"description": "Simplifies the GitHub interface and adds useful features",
 	"homepage_url": "https://github.com/refined-github/refined-github",
 	"manifest_version": 2,
-	"minimum_chrome_version": "88",
+	"minimum_chrome_version": "91",
 	"applications": {
 		"gecko": {
 			"id": "{a4c4eda4-fb84-4a84-b4a1-f7c1cbf2a1ad}",
-			"strict_min_version": "84.0"
+			"strict_min_version": "91.0"
```
fregante, line 11: Only notable change due to the increased `lib` in TypeScript.

tsconfig.json
```diff
@@ -1,7 +1,12 @@
 {
-	"extends": "@sindresorhus/tsconfig/tsconfig.json",
+	"extends": "@sindresorhus/tsconfig",
 	"compilerOptions": {
-		"target": "es2020",
+		"target": "es2021",
```
fregante, line 4: We could target es2022 if it wasn't for Safari

PR #5714, opened 2022-06-21
Fix `release-download-count` on releases with @ in their names

### LINKED ISSUE #5713: `release-download-count`: GraphQL:Parse error on ":" (COLON) at [4, 23]
### Please ensure:

### Example URLs

https://github.com/apollographql/federation-rs/releases/tag/supergraph%40v2.0.5

### Extension version

22.5.22

### Browser(s) used

Microsoft Edge

### Description

```
Error: GraphQL:
Parse error on ":" (COLON) at [4, 23]
    at mem.cacheKey (chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/refined-github.js:4069:60)
    at async chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/refined-github.js:8319:32
    at async init (chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/refined-github.js:8318:24)
    at async runFeature (chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/refined-github.js:3913:20)
    at async setupPageLoad (chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/refined-github.js:3921:34)
    at async chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/refined-github.js:3946:29
```

### Include in this issue:

- [X] Console errors, if any

yakov116:
Closes #5713

## Test URLs
https://github.com/apollographql/federation-rs/releases/tag/supergraph%40v2.0.5

## Screenshot
NA

### DIFF
--- a/source/github-helpers/api.ts
+++ b/source/github-helpers/api.ts
@@ -49,7 +49,7 @@ interface RestResponse extends AnyObject {
 	ok: boolean;
 }
 
-export const escapeKey = (value: string | number): string => '_' + String(value).replace(/[ ./-]/g, '_');
+export const escapeKey = (value: string | number): string => '_' + String(value).replace(/[ ./-@]/g, '_');
 
 export class RefinedGitHubAPIError extends Error {
 	response: AnyObject = {};


fregante (approved):

fregante:
I’m surprised it took us years to find this issue; releases with the @ symbol are not uncommon at all

yakov116:
Yup

source/github-helpers/api.ts
```diff
@@ -49,7 +49,7 @@ interface RestResponse extends AnyObject {
 	ok: boolean;
 }
 
-export const escapeKey = (value: string | number): string => '_' + String(value).replace(/[ ./-]/g, '_');
+export const escapeKey = (value: string | number): string => '_' + String(value).replace(/[ ./-@]/g, '_');
```
fregante, line 52: We should replace this with `source/helpers/hash-string.ts` instead. No point in risking this again, the key is for internal use only

yakov116: Will do

yakov116: Having second thoughts. Would this make debugging harder? Not sure why it would, but this gives a random ID instead of a readable name.

yakov116: But its possible to hit duplicates, we needed a way to garentee no duplicate names.

fregante: > But its possible to hit duplicates

That's not wrong, but highly unlikely. These two already produce the same key, but the chances of that happening in the real world are so low:

- refined/github-repo
- refined-github/repo

A hash-based approach would avoid that too.

PR #5720, opened 2022-06-24
Fix ajax navigation detection

### LINKED ISSUE #5719: Nothing works until you refresh the page (Ajax issue)
### Please ensure:

### Example URLs

Everywhere

### Extension version

22.5.22

### Browser(s) used

Safari

### Description

I keep noticing more and more features are missing until I refresh the page. Example:

2. Open a folder
3. Notice lack of `ci-link`, `rgh-linkify-features`, `quick-file-edit`,  `clean-repo-tabs`, etc

https://user-images.githubusercontent.com/1402241/175481173-960b33f1-f928-43d7-ab5e-2a594c84e261.mov

Interestingly I found the `New Issue` textareas to not turn into rich text editors sometimes (GitHub issue)

before

### Include in this issue:

- [ ] Console errors, if any

fregante:
- Closes #5719 

@refined-github/maintainers please test because I'd like to release a new version within a few hours

https://user-images.githubusercontent.com/1402241/175499367-6c01e4c3-7317-4cbf-8d57-dcf9c7e003a1.mov

### DIFF
--- a/source/features/file-finder-buffer.tsx
+++ b/source/features/file-finder-buffer.tsx
@@ -58,8 +58,8 @@ function pjaxCompleteHandler(): void {
 }
 
 function init(): void {
-	window.addEventListener('pjax:start', pjaxStartHandler);
-	window.addEventListener('pjax:complete', pjaxCompleteHandler);
+	window.addEventListener('turbo:visit', pjaxStartHandler);
+	window.addEventListener('turbo:load', pjaxCompleteHandler);
 }
 
 void features.add(import.meta.url, {
--- a/source/features/index.tsx
+++ b/source/features/index.tsx
@@ -140,8 +140,31 @@ const globalReady: Promise<RGHOptions> = new Promise(async resolve => {
 
 	document.documentElement.classList.add('refined-github');
 
+	if (pageDetect.isEnterprise()) {
+		polyfillTurboEvents();
+	}
+
 	resolve(options);
 });
+
+const turboPolyfillController = new AbortController();
+const nativeTurboDetectorController = new AbortController();
+
+function dispatchTurboEvent(event: Event): void {
+	nativeTurboDetectorController.abort();
+	const turboEvent = event.type === 'pjax:start' ? 'turbo:visit' : 'turbo:load';
+	document.dispatchEvent(new CustomEvent(turboEvent));
+}
+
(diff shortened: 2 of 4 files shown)


fregante:
For debugging, run this in the console:

```
monitorEvents(document.documentElement, ['turbo:before-cache','turbo:before-fetch-request','turbo:before-fetch-response','turbo:before-render','turbo:before-stream-render','turbo:before-visit','turbo:click','turbo:frame-load','turbo:frame-render','turbo:load','turbo:reload','turbo:render','turbo:submit-end','turbo:submit-start','turbo:visit'])
```

## Click example

Screen Shot 2

## Back button example
Screen Shot 3

FloEdelmann:
The old event names are probably still used in GHE, so we should keep supporting them for a while, shouldn't we?

cheap-glitch:
Seems like a lot of features are still affected by this, e.g.
- `highlight-collaborators-and-own-conversations`
- `clean-rich-text-editor`
- `clean-pinned-issues`
- and probably others.

It's like the de-duplication mechanism (with the `has-rgh` and `has-rgh-inner`) needs to be rethought a bit

(This is on `22.6.24` btw.)

fregante:
All of them work for me:

https://user-images.githubusercontent.com/1402241/175771063-d2421acd-af42-49ec-a218-32991cd415b8.mov

cheap-glitch:
Okay, so there must be some kind of tomfoolery goin' on here, because I still see the old event names on Firefox:

image

So basically https://github.com/refined-github/refined-github/commit/f11c40645726a5553f7ff49a05cc22550bbffc17 broke everything for me :disappointed: 

Could it be GitHub doing some A/B testing or something?

fregante:
Likely AB testing. I made the change you suggested and released yet another hotfix release 😁 
We haven't had to do this in years!

cheap-glitch:
> I made the change you suggested and released yet another hotfix release

Working great now, thanks!

cheap-glitch:
> Stopped working for me now...

Same, it broke sometimes yesterday. Looks like everyone has the new events now:

image

But that also means the polyfill actually breaks ajax detection…

fregante:
#5738 is the opposite of what you're describing: runs twice vs runs never

That issue is fixed in https://github.com/refined-github/refined-github/pull/5742

PR #5721, opened 2022-06-24
Restore `user-local-time`

### LINKED ISSUE #5699: `user-local-time` not working
### Please ensure:

### Example URLs

### Extension version

22.5.22

### Browser(s) used

Brave Version 1.39.122 Chromium: 102.0.5005.115 (Official Build) (arm64)

### Description

The user local timezone isn't shown anymore in the hover card.
Not sure if it was broken before, I only noticed it now.

From what I could see the feature hasn't been removed.

CleanShot 2022-06-17 at 13 30 19@2x

### Include in this issue:

- [ ] Console errors, if any

mangkoran:
- Closes #5699

## Test URLs

(In case its you that viewing the PR, I put several accounts here)
- @fregante (sorry for the ping)
- @yakov116 (sorry for the ping)
- @mangkoran

## Screenshot
image

### DIFF
--- a/source/features/user-local-time.tsx
+++ b/source/features/user-local-time.tsx
@@ -142,8 +142,12 @@ async function insertUserLocalTime(hovercardContainer: Element): Promise<void> {
 	container.title = `Timezone guessed from their last commit: ${date}`;
 }
 
+const selector = [
+	'.js-hovercard-content .Popover-message div.d-flex.mt-3.overflow-hidden > div.d-flex:not(.rgh-user-local-time)',
+	'.js-hovercard-content .Popover-message div.d-flex.mt-3 > div.overflow-hidden.ml-3:not(.rgh-user-local-time)', // GHE 2022/06/24
+].join(',');
 function init(): void {
-	observe('.js-hovercard-content .Popover-message div.d-flex.overflow-hidden.mt-3 > div.d-flex:not(.rgh-user-local-time)', {
+	observe(selector, {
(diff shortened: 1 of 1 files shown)


mangkoran:
A bit out of context, but how this feature works? Sometimes I can see the local time, sometimes not (disregarding this issue)?

fregante (approved):
## Mantap!

fregante:
I confirm it works:

Screen Shot



> how this feature works

Commits include your timezone, so we fetch the user’s "github event" until we find the latest commit. If a user hasn't pushed in a while or hasn't pushed at all, the feature won't work.

yakov116:
Will this break GHE?

mangkoran:
Thank you for amending. I have no idea it will break GHE (I have no idea that GHE is different either). Glad to be able to contribute to RGH :D

cheap-glitch:
> I have no idea that GHE is different either

GHE = GitHub Enterprise. If I'm not wrong it's basically separate versions of GitHub that run on private company servers, which means they're always lagging behind in term of updates. So we also have to support (up to a point) "older versions" of GitHub.

> Glad to be able to contribute to RGH

Thank you for your help! We have a bunch of easy issues if you feel like contributing again :wink:

https://github.com/refined-github/refined-github/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3A%22good+first+issue%22

PR #5726, opened 2022-06-25
Meta: Deinit all feature delegates using `AbortSignal`

cheap-glitch:
rocket


source/features/sync-pr-commit-title.tsx
```diff
@@ -77,31 +77,24 @@ async function updateCommitTitle(): Promise<void> {
...
+
+	onPrCommitMessageRestore(updateUI, signal);
+	onPrMergePanelOpen(updateCommitTitle, signal);
+
+	delegate(document, '#merge_title_field', 'input', updateUI, {signal});
+	delegate(document, 'form.js-merge-pull-request', 'submit', updatePRTitle, {signal});
+	delegate(document, '.rgh-sync-pr-commit-title', 'click', () => {
+		internalController.abort();
+		getUI().remove();
+	}, {signal});
```
cheap-glitch, line 97: This code is kind of awkward, there's probably a better solution :thinking:

source/globals.d.ts
```diff
@@ -1,7 +1,7 @@
 /* eslint-disable @typescript-eslint/consistent-indexed-object-style */
 
 type AnyObject = Record<string, any>;
-type DeinitHandle = MutationObserver | ResizeObserver | IntersectionObserver | {destroy: VoidFunction} | {abort: VoidFunction} | VoidFunction;
+type DeinitHandle = MutationObserver | ResizeObserver | IntersectionObserver | AbortController | {abort: VoidFunction} | VoidFunction;
```
cheap-glitch, line 4: Actually we don't need to include `AbortController` here, will drop it

cheap-glitch (commented):
I guess the next step is to add support for `AbortSignal` into `selector-observer` so we can get rid of `Deinit` completely :slightly_smiling_face:

source/features/prevent-link-loss.tsx
```diff
@@ -63,12 +63,11 @@ const updateUI = debounceFn(({delegateTarget: field}: delegate.Event<Event, HTML
...
-	return [
-		updateUI.cancel,
-		delegate(document, 'form:is(#new_issue, #new_release) textarea, form.js-new-comment-form textarea, textarea.comment-form-textarea', 'input', updateUI),
-		delegate(document, '.rgh-prevent-link-loss', 'click', handleButtonClick),
-	];
+function init(signal: AbortSignal): Deinit {
+	delegate(document, 'form:is(#new_issue, #new_release) textarea, form.js-new-comment-form textarea, textarea.comment-form-textarea', 'input', updateUI, {signal});
+	delegate(document, '.rgh-prevent-link-loss', 'click', handleButtonClick, {signal});
+
+	return updateUI.cancel;
```
fregante, line 70: Thoughts?

```suggestion
	signal.addEventListener('abort', updateUI.cancel);
```

I'd love to have an `onAbort(emitter: signal, listener: function | controller)` helper for this kind of situations:

```suggestion
	onAbort(signal, updateUI.cancel)
```

and also when we have multiple controllers and we want to link them together, like:


```js
function foo(otherSignal) {
	const controller = new AbortController
	onAbort(otherSignal, controller)
	return controller;
}
```

In this case, `controller` would also be aborted when `otherSignal` is aborted.

I'm not sure it's useful in this specific case though, `return fn` will always be more convenient.

source/features/quick-label-removal.tsx
```diff
@@ -50,17 +50,15 @@ function addRemoveLabelButton(label: HTMLElement): void {
...
-			add: addRemoveLabelButton,
-		}),
+	delegate(document, '.rgh-quick-label-removal:not([disabled])', 'click', removeLabelButtonClickHandler, {signal});
 
-		delegate(document, '.rgh-quick-label-removal:not([disabled])', 'click', removeLabelButtonClickHandler),
-	];
+	return observe('.js-issue-labels .IssueLabel:not(.rgh-quick-label-removal-already-added)', {
+		constructor: HTMLElement,
+		add: addRemoveLabelButton,
+	});
```
fregante, line 61: I propose wrapping `observe` to enable:

- signal support
- selector parsing for the types
- array of selectors
- simpler listener (`observe(selector, callback)` instead of `observe(selector, {add: callback})`)

and potentially, separately:
- automatic class addition via `getSnapshotUUID`: if we have CSS relying on certain classes we can still add those manually, on top of the automatic class

Can be done in 1-2 additional PRs or added to #5222

source/features/quick-mention.tsx
```diff
@@ -33,7 +33,9 @@ function mentionUser({delegateTarget: button}: delegate.Event): void {
 	textFieldEdit.insert(newComment, `${spacer}${prefixUserMention(userMention)} `);
 }
 
-function init(): Deinit {
+function init(signal: AbortSignal): void {
+	delegate(document, 'button.rgh-quick-mention', 'click', mentionUser, {signal});
+
 	// `:first-child` avoids app badges #2630
 	// The hovercard attribute avoids `highest-rated-comment`
 	// Avatars next to review events aren't wrapped in a <div> #4844
```
fregante, line 41: The following selector can be turned into an array in the lint PR

#5727

source/features/sync-pr-commit-title.tsx
```diff
@@ -77,31 +77,24 @@ async function updateCommitTitle(): Promise<void> {
...
-	for (const subscription of subscriptions) {
-		subscription.destroy();
-	}
-
-	subscriptions.length = 0;
+function init(deinitSignal: AbortSignal): void {
+	const internalController = new AbortController();
+	const {signal} = internalController;
+	deinitSignal.addEventListener('abort', () => {
+		if (!signal.aborted) {
```
fregante, line 84: Like promises, AbortControllers can only be thrown once. Successive calls are ignored, no need to check for this

source/features/sync-pr-commit-title.tsx
```diff
@@ -77,31 +77,24 @@ async function updateCommitTitle(): Promise<void> {
...
-function deinit(): void {
-	for (const subscription of subscriptions) {
-		subscription.destroy();
-	}
-
-	subscriptions.length = 0;
+function init(deinitSignal: AbortSignal): void {
+	const internalController = new AbortController();
+	const {signal} = internalController;
+	deinitSignal.addEventListener('abort', () => {
```
fregante, line 83: The nearby code can be greatly simplified by the `onAbort` helper I mentioned earlier

source/features/sync-pr-commit-title.tsx
fregante: If Safari wasn't a party-pooper (it doesn't bind `.abort`) we could use

```js
 	delegate(document, '.rgh-sync-pr-commit-title', 'click', internalController.abort, {signal});
	onAbort(internalController, () => {
		getUI().remove()
	})
```

Either way not a great improvement

source/features/sync-pr-commit-title.tsx
```diff
@@ -77,31 +77,24 @@ async function updateCommitTitle(): Promise<void> {
...
-}
-
-function deinit(): void {
-	for (const subscription of subscriptions) {
-		subscription.destroy();
-	}
-
-	subscriptions.length = 0;
+function init(deinitSignal: AbortSignal): void {
+	const internalController = new AbortController();
```
fregante, line 81:
```suggestion
	const localController = new AbortController();
```

source/github-events/on-new-comments.ts
```diff
@@ -50,20 +50,18 @@ function addListeners(): delegate.Subscription[] {
...
-export default function onNewComments(callback: VoidFunction): Deinit {
+export default function onNewComments(callback: VoidFunction, signal: AbortSignal): Deinit {
+	addListeners(signal);
 	handlers.add(callback);
 
 	return [
-		...addListeners(),
 		handlers.clear,
 		observer,
 	];
```
fregante, line 67: If it takes a signal, it should use it for everything 

```js
onAbort(signal, handlers, observer);// It could accept multiple objects, like the `getDeinit` function can handle it all
```

source/github-events/on-new-comments.ts
```diff
@@ -50,20 +50,18 @@ function addListeners(): delegate.Subscription[] {
...
 }
 
-export default function onNewComments(callback: VoidFunction): Deinit {
+export default function onNewComments(callback: VoidFunction, signal: AbortSignal): Deinit {
+	addListeners(signal);
 	handlers.add(callback);
 
 	return [
-		...addListeners(),
 		handlers.clear,
```
fregante, line 65: I don’t think this ever worked? `Set#clear` isn't bound

Screen Shot

fregante (approved):
6eg

source/features/prevent-link-loss.tsx
fregante: For this specific case we should just drop this. There's no chance that the user changes page within 300ms of the last keypress

source/features/prevent-link-loss.tsx
```diff
@@ -63,12 +63,11 @@ const updateUI = debounceFn(({delegateTarget: field}: delegate.Event<Event, HTML
...
 });
 
-function init(): Deinit {
-	return [
-		updateUI.cancel,
-		delegate(document, 'form:is(#new_issue, #new_release) textarea, form.js-new-comment-form textarea, textarea.comment-form-textarea', 'input', updateUI),
-		delegate(document, '.rgh-prevent-link-loss', 'click', handleButtonClick),
-	];
+function init(signal: AbortSignal): Deinit {
+	delegate(document, 'form:is(#new_issue, #new_release) textarea, form.js-new-comment-form textarea, textarea.comment-form-textarea', 'input', updateUI, {signal});
```
fregante, line 67: Multiline selector?

source/github-events/on-new-comments.ts
fregante: Fixed in ff282b95

source/features/sync-pr-commit-title.tsx
fregante: We have this now:

- #5791

fregante:
Ping on this if you have time, the onAbort helper was added

fregante:
you only live once

PR #5728, opened 2022-06-25
Fix spacing between team member statuses and footer on dashboard

cheap-glitch:
## Test URLs

https://github.com/orgs/refined-github/dashboard

## Screenshot

**Before**

before

**After**

after


fregante:
I think the footer should have margin-top instead so we don't have to play whack-a-mole with these elements

fregante:
Feel free to reopen when you have time

PR #5752, opened 2022-06-29
Fix CSS features disappearing after clicking on a link

### LINKED ISSUE #5719: Nothing works until you refresh the page (Ajax issue)

fregante:
- Maybe closes #5719 ?
- Potentially also mentioned in https://github.com/refined-github/refined-github/pull/5720#issuecomment-1167966280

## Test URLs

1. Open https://github.com/refined-github/refined-github/issues
3. Click on Issues tab

## Before

https://user-images.githubusercontent.com/1402241/176489932-20874707-d01b-4234-aa1f-fee4da581eea.mov

## After

https://user-images.githubusercontent.com/1402241/176490087-90b92331-a8f3-4f01-b167-017334003e32.mov

### DIFF
--- a/source/features/clean-repo-sidebar.tsx
+++ b/source/features/clean-repo-sidebar.tsx
@@ -61,7 +61,7 @@ async function hideEmptyMeta(): Promise<void> {
 }
 
 async function init(): Promise<void> {
-	document.body.classList.add('rgh-clean-repo-sidebar');
+	document.documentElement.classList.add('rgh-clean-repo-sidebar');
 
 	await Promise.all([
 		cleanReleases(),
--- a/source/features/file-finder-buffer.tsx
+++ b/source/features/file-finder-buffer.tsx
@@ -36,7 +36,7 @@ function pjaxStartHandler(event: CustomEvent): void {
 	bufferField.focus();
 
 	// Hide the header elements instead of removing them so they can be restored #4999
-	document.body.classList.add('rgh-file-finder-buffer');
+	document.documentElement.classList.add('rgh-file-finder-buffer');
 }
 
 function pjaxCompleteHandler(): void {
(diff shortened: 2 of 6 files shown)


yakov116:
Working for me but has a noticeable jump

fregante:
I've been noticing that jump too, but only sometimes (in a Safari window I always get a jump on reload, but on the other window I do not 😰).

Either way it's unrelated to this specific issue and we should probably open a new issue for that if it persists. The only reason why it could be happening is in `globalReady` (or due to an unavoidable browser restriction/bug)

fregante:
ha!

yakov116:
RAAA LOL

fregante:
> ```js
>  await waitFor(() => document.body); 
> ```

That's potentially the issue, but it's not really something new or anything that GitHub or us can affect, that's the weird part.

We can replace it with an `observeElement(document.documentElement)`, assuming that it _always_ exists (can we assume that?)

yakov116:
I forgot already how did we land up with document.body? I would want to check in history why we did not use observe element.

fregante:
> I forgot already how did we land up with document.body?

The content scripts are injected on `document_start`, which may happen before the document is parsed. This means that the script might find an empty page so we have to wait for it. We use to have `requestAnimationFrame` to catch the first draw, but that means we can't do anything until the tab becomes visible, even if it loaded a minute ago.

Maybe we can combine the two in a race:


```js
await Promise.race([
	frame(),
	waitFor(() => document.body)
])
```

yakov116:
If I am not mistaken, requesting the frame broke Firefox. (GitHub froze)

cheap-glitch (approved):
Working on Firefox, but it is somewhat slower than before.

cheap-glitch:
Btw there's still some `document.body.classList` calls in https://github.com/refined-github/refined-github/blob/main/safari/Shared%20(App)/Resources/Script.js, not sure if those should be updated too.

fregante:
That’s a web view in the native app.

PR #5754, opened 2022-06-29
Fit large images in the window when viewing single files

### LINKED ISSUE #5740: Max image size for image previews
### Description

For very large images it makes more sense to resize it to fit the screen, rather than crop it (current behavior). A simple `max-height` and `max-width` for image previews would make these pages easier to see.

### Screenshot

**Current:**

github com_KatsuteDev_Mal4J_blob_main_assets_banner-sq png

**Proposed:**

github com_KatsuteDev_Mal4J_blob_main_assets_banner-sq png (1)

### Example URLs

https://github.com/KatsuteDev/Mal4J/blob/main/assets/banner-sq.png

Katsute:
Closes #5740

Adds the following css to restrict image blob sizes:
```css
div[data-target='readme-toc.content'] div.blob-wrapper img {
	max-width: 100%;
}
```

## Test URLs

https://github.com/KatsuteDev/Mal4J/blob/main/assets/banner-sq.png

## Screenshot

github com_KatsuteDev_Mal4J_blob_main_assets_banner-sq png (2)

### DIFF
--- a/source/refined-github.css
+++ b/source/refined-github.css
@@ -187,7 +187,7 @@ pr-branches
 	display: block !important;
 }
 
-/* Resize large images for blob preview */
+/* Fit large images in the window when viewing single files */
 div[data-target='readme-toc.content'] div.blob-wrapper img {
 	max-width: 100%;
 	max-height: 100%;
--- a/source/refined-github.css
+++ b/source/refined-github.css
@@ -188,7 +188,7 @@ pr-branches
 }
 
 /* Resize large images for blob preview */
-div[data-target="readme-toc.content"] div.blob-wrapper img {
-    max-width: 100%;
-    max-height: 100%;
+div[data-target='readme-toc.content'] div.blob-wrapper img {
+	max-width: 100%;
+	max-height: 100%;
 }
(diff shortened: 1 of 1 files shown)


fregante:
Thank you! 

This doesn't affect images in commits, which is good 👍 

https://github.com/refined-github/refined-github/commit/4218456a172742457c8439728c54ad5a5c7ed887#diff-f44b12b0416bb268cf2491ea46a5f8deae4823068722a5bfcc61d0f87d371169

Katsute:
It only limits the max size of images, it shouldn't affect any images that are smaller.

source/refined-github.css
```diff
@@ -186,3 +186,9 @@ pr-branches
 .TimelineItem[id^='event-'] .TimelineItem-body :is(ins, del) {
 	display: block !important;
 }
+
+/* Resize large images for blob preview */
```
fregante, line 190:
```suggestion
/* Fit large images in the window when view single files */
```

source/refined-github.css
```diff
@@ -186,3 +186,9 @@ pr-branches
...
+div[data-target='readme-toc.content'] div.blob-wrapper img {
+	max-width: 100%;
+	max-height: 100%;
```
fregante, line 193: I don’t think the height matters, `*-height: 100%` is ignored when the container doesn't have a set height

fregante (commented):
Related link with potentially more cases to test or ignore https://docs.github.com/en/repositories/working-with-files/using-files/working-with-non-code-files

cheap-glitch (approved):

PR #5767, opened 2022-07-03
Add `useful-forks` to archived repos

fregante:
- Closes https://github.com/refined-github/refined-github/issues/5443

## Test URLs

https://github.com/reactjs/react-autocomplete

## Screenshot

Screen Shot 4

### DIFF
--- a/source/features/useful-forks.tsx
+++ b/source/features/useful-forks.tsx
@@ -3,9 +3,19 @@ import elementReady from 'element-ready';
 import * as pageDetect from 'github-url-detection';
 import {RepoForkedIcon} from '@primer/octicons-react';
 
+import select from 'select-dom';
+
 import features from '.';
 import {getRepo} from '../github-helpers';
 import looseParseInt from '../helpers/loose-parse-int';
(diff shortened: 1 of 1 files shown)


payne911:
I would have thought it would be nicer to maybe have it closer to

"You can find [Useful Forks](https://useful-forks.github.io/)"

fregante:
This is not my first choice, but it's a compromise for:

- keeping it short so it doesn't push the content down
- giving a heads-up to tell that the user is about to go to a third party
- avoiding an `elementReady` that would negatively affect every single page load (to detect isArchivedRepo) while only being useful on a fraction of them

I suppose "You can find [useful forks [link-icon]](https://useful-forks.github.io/)" could also work, but it's not as clear that it's a third party. This is particularly a pain point when the users find themselves on an unfamiliar site: https://github.com/useful-forks/useful-forks.github.io/issues/42

payne911:
I think it would generally be assumed that the "[link-icon](https://fontawesome.com/v4/icon/external-link)" would open up an external website.

PR #5773, opened 2022-07-04
Polish `latest-tag-button`

fregante:
1. Use compare icon, not commit/diff: https://github.com/pixiebrix/pixiebrix-extension/compare/main
2. Drop useless indicator: if we show the compare link, then it's "behind"; Saying "I don't know how behind" isn't necessary

## Test URLs

- 12 commits: https://github.com/webpack/webpack
- Undeterminable: https://github.com/microsoft/typescript

## Before

Screen Shot 2Screen Shot 3

## After

Screen ShotScreen Shot


fregante:
?

We had the indicator since the beginning. That PR only changed it

kidonng:
I just mentioned when the asterisk is added, because myself haven't really seen (or noticed) it.

Issue #2125, opened 2019-06-06
Images next to comments reactions are broken for bots


webknjaz:
Example comment: https://github.com/aio-libs/aiohttp/pull/3809#issuecomment-498809698

bfred-it:
Probably because it has `[bot]` in the name but shouldn’t. PR welcome to drop it

bfred-it:
Apparently the image URL is different for bots. Users have

https://github.com/bfred-it.png

Bots don't:

https://github.com/dependabot-preview.png

notlmn:
> just drop the avatar from the list.

This, reactions from bots don't matter for regular conversations, ignoring bots makes sense.

webknjaz:
@bfred-it I've got some experience with writing bots. So here's some info about them and common mistakes of integration writers as well as GitHub itself.
1) Bots are GitHub Apps (https://developer.github.com/v3/apps/)
2) When bot posts some content it has a special label in the UI
3) GitHub tries to find the best picture to show:
   - priority goes to the App picture if the developer has uploaded it. For
     example, at https://github.com/apps/ansiwatch I've specified the picture
     so it'll be used.
   - if there's no picture, GitHub will use the developer's avatar.
     - if it's a user account, it'll take it from the profile: https://github.com/webknjaz/.me/issues/1 (issue created by bot) / (app page image: https://github.com/identicons/app/app/dobby-ans) for https://github.com/apps/dobby-ans created by https://github.com/webknjaz (user)
     - if it's an organization, it'll user the org avatar: https://github.com/apps/diactoros (app page reveals that there's no image) but in the event GitHub broadcasts it'll point to https://avatars0.githubusercontent.com/u/40318227?v=4 which is the avatar of https://github.com/sanitizers ‒ org under which this App is created. Here's an excerpt of such event: https://gist.github.com/webknjaz/8590b205a8c14d888385dace75672938
4) Interestingly, when GitHub renders the App name as text it'll append `[bot]` in the end, like `dobby-ans[bot]`. It appears in a lot of places and there are even bugs where they render wrong URLs.
   Some places create links like https://github.com/dobby-ans[bot] which are obviously wrong and the right way to do this is to render them like https://github.com/apps/dobby-ans
   BTW refined-github does the same mistake in reactions: in that comment I linked above it produces https://github.com/dependabot-preview link which should be https://github.com/apps/dependabot-preview

webknjaz:
[...]

PR #5778, opened 2022-07-04
Add `repo-avatars` feature

### LINKED ISSUE #5546: Show repository avatars in header
### Description

You probably have seen it in many places: dashboard, mobile apps, and the new hot "For you" feed:

image

It's not really _repository_ avatars like on e.g. GitLab, just the avatar of the user/organization the repo belongs to. But still, a nice thing to see more.

I propose we also put an avatar in the repository header (which is available now [via my userscript](https://github.com/kidonng/cherry/tree/cc877cce26eb469e132dab37ff6c5482b9387c21/scripts#github-repository-avatars)):

Feeling familiar? Because you've probably seen similar design [elsewhere](https://gist.github.com/sindresorhus/a4706e81fbac24b5380a0d6d6bfb584e):

image

### FAQ

- **Where would this take effect?**

  Public repositories, that is not template/fork/etc.

  In other words, we only replace the vanilla repo icon with a more awesome avatar, which should benefit most repos. We don't want to take away other unique icons that distinguish the status of the repository, especially since we already hide the label.

  Alternatively, we can _prepend_ the avatar to the icon, which will make it suitable for all repositories, though it would not look great.

- **Why not propose to GitHub?**

  _cough_ we are more flexible than their roadmap. _cough_ And seeing this only applied to some places (a.k.a. inconsistencies), I don't see how will GitHub implement this.

- **You know [the first rule](https://github.com/refined-github/refined-github/issues/2960#issuecomment-606764895) right?**

  Because I already _planned_ to to propose it to Refined GitHub. This applies to most of my other user scripts/styles posted here.

### Screenshot

_No response_

### Example URL

kidonng:
Resolve #5546

Code loosely adapted from https://github.com/kidonng/cherry/blob/b7dfa1b01802e708a40c0bef82a93b1f5a6ca3eb/scripts/github-repository-avatars.user.tsx

## Test URLs

- User: https://github.com/sindresorhus/awesome
- Org: Here

## Screenshot

<table>
<tr>
	<td>User
	<td>Org
<tr>
	<td>image
	<td>image
</table>

### DIFF
--- a/source/features/reactions-avatars.tsx
+++ b/source/features/reactions-avatars.tsx
@@ -6,7 +6,7 @@ import {flatZip} from 'flat-zip';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
-import {getUsername} from '../github-helpers';
+import {getUsername, getUserAvatar} from '../github-helpers';
 
 const arbitraryAvatarLimit = 36;
 const approximateHeaderLength = 3; // Each button header takes about as much as 3 avatars
@@ -44,7 +44,7 @@ function getParticipants(button: HTMLButtonElement): Participant[] {
 
 		// If it's not a bot, use a shortcut URL #2125
 		if (cleanName === username) {
-			const imageUrl = (pageDetect.isEnterprise() ? `/${username}.png` : `https://avatars.githubusercontent.com/${username}`) + '?size=32';
+			const imageUrl = getUserAvatar(username, 16);
 			participants.push({button, username, imageUrl});
 		}
 	}
--- a/source/features/reactions-avatars.tsx
+++ b/source/features/reactions-avatars.tsx
@@ -6,7 +6,8 @@ import {flatZip} from 'flat-zip';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
-import {getUsername, getUserAvatar} from '../github-helpers';
+import {getUsername} from '../github-helpers';
+import getUserAvatar from '../github-helpers/get-user-avatar';
 
 const arbitraryAvatarLimit = 36;
 const approximateHeaderLength = 3; // Each button header takes about as much as 3 avatars
@@ -33,18 +34,8 @@ function getParticipants(button: HTMLButtonElement): Participant[] {
 			continue;
 		}
 
-		const cleanName = username.replace('[bot]', '');
-
-		// Find image on page. Saves a request and a redirect + add support for bots
-		const existingAvatar = select(`img[alt="@${cleanName}"]`);
-		if (existingAvatar) {
-			participants.push({button, username, imageUrl: existingAvatar.src});
-			continue;
-		}
-
-		// If it's not a bot, use a shortcut URL #2125
-		if (cleanName === username) {
-			const imageUrl = getUserAvatar(username, 16);
+		const imageUrl = getUserAvatar(username, 16);
+		if (imageUrl) {
 			participants.push({button, username, imageUrl});
 		}
 	}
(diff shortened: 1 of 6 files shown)


source/features/repo-avatars.tsx
```diff
@@ -0,0 +1,46 @@
...
+		return;
+	}
+
+	const author = select('#repository-container-header [rel=author]')!;
+	const username = author.textContent!;
+	const alt = `@${username}`;
+	const fallbackSrc = (pageDetect.isEnterprise()
+		? `/${username}.png`
+		: `https://avatars.githubusercontent.com/${username}`) + '?size=48';
+	const src = select(`img[alt="${alt}"]`)?.src ?? fallbackSrc;
```
fregante, line 21: We have a nearly-identical piece of code in `reaction-avatars`, which also takes display density into account.

source/features/repo-avatars.tsx
```diff
@@ -0,0 +1,46 @@
...
+		<img
+			className="avatar mr-2"
+			src={src}
+			width="24"
+			height="24"
+			alt={alt}
+		/>
+	);
+
+	if (author.dataset.hovercardType === 'user') {
```
fregante, line 33: If you use getRepo, this can be replaced by a select.exists, or maybe we can add this check to url-detections as isOrganizationRepo

fregante (approved):

source/features/repo-avatars.tsx
kidonng: We already have `isOrganizationRepo()` 😃!

source/features/repo-avatars.tsx
kidonng: This is indeed adapted from `reations-avatars`, just haven't thought of making it into a function.

source/features/repo-avatars.tsx
```diff
@@ -0,0 +1,43 @@
...
+async function init(): Promise<void> {
+	// Icon for public but not template/fork/etc. repos
+	const icon = await elementReady('#repository-container-header .octicon-repo');
+	if (!icon) {
+		return;
+	}
+
+	const username = getRepo()!.owner;
+	const alt = `@${username}`;
+	const src = select(`img[alt="${alt}"]`)?.src ?? getUserAvatar(username, 24);
```
fregante, line 18: I was thinking that this select could be part of the helper too

fregante (approved):

source/features/reactions-avatars.tsx
```diff
@@ -33,18 +34,8 @@ function getParticipants(button: HTMLButtonElement): Participant[] {
...
-		const cleanName = username.replace('[bot]', '');
-
-		// Find image on page. Saves a request and a redirect + add support for bots
-		const existingAvatar = select(`img[alt="@${cleanName}"]`);
-		if (existingAvatar) {
-			participants.push({button, username, imageUrl: existingAvatar.src});
-			continue;
-		}
-
-		// If it's not a bot, use a shortcut URL #2125
```
fregante, line 45: Does our current solution cover #2125?

kidonng:
@fregante 🚀?

source/github-helpers/get-user-avatar.ts
```diff
@@ -0,0 +1,21 @@
+import select from 'select-dom';
+import * as pageDetect from 'github-url-detection';
+
+export default function getUserAvatar(username: string, size: number): string | void {
```
yakov116, line 4: Well I am going to review. 🤣 

Can you please change this to an object {username, size}?
So `getUserAvatar(username, 16);` has a description.

kidonng: The variable names are obvious enough I think. Or just look at the function signature.

PR #5779, opened 2022-07-04
Extend `user-local-time` to user profiles

### LINKED ISSUE #5775: Extend `user-local-time` to user profiles
### Description

Quote https://github.com/refined-github/refined-github/pull/2549#issuecomment-559750761:

> * Also display it on profiles like [`linkify-user-location`](https://github.com/refined-github/refined-github/blob/main/source/features/linkify-user-location.tsx)

### Screenshot

image

### Example URLs

https://github.com/fregante

kidonng:
Resolve #5775

A lot of dup code, get this thing out first

## Test URLs

https://github.com/fregante

### DIFF
--- a/source/features/user-local-time.tsx
+++ b/source/features/user-local-time.tsx
@@ -79,6 +79,35 @@ function parseOffset(date: string): number {
 	return (hours * 60) + (hours < 0 ? -minutes : minutes);
 }
 
+async function handleDatePromise({
+	datePromise,
+	placeholder,
+	container,
+}: {
+	datePromise: Promise<string | false>;
+	placeholder: JSX.Element;
+	container: JSX.Element;
+}): Promise<void> {
+	const date = await datePromise;
+	if (!date) {
+		placeholder.textContent = 'Timezone unknown';
+		container.title = 'Timezone couldn’t be determined from their last commits';
+		return;
+	}
+
+	const userTime = new Date();
+	userTime.setMinutes(parseOffset(date) + userTime.getTimezoneOffset() + userTime.getMinutes());
+
+	const timeFormatter = new Intl.DateTimeFormat(undefined, {
+		hour: 'numeric',
+		minute: 'numeric',
+		weekday: userTime.getDay() === new Date().getDay() ? undefined : 'long',
+	});
+
+	placeholder.textContent = timeFormatter.format(userTime);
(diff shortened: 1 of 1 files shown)


source/features/user-local-time.tsx
```diff
@@ -79,6 +79,35 @@ function parseOffset(date: string): number {
 	return (hours * 60) + (hours < 0 ? -minutes : minutes);
 }
 
+async function handleDatePromise({
```
yakov116, line 82: How about naming it something with display? Isn't that what this function does?

kidonng: It's just a random name I come up with, these code are so intertwined I have no idea what I'm doing.

fregante:
The structure looks about right. If we had something like Preact it’d be presumably easier to handle this sort of content. 

I wonder if attachElement would make this somewhat easier, where you have an intermediary function which calls attachElement with the time-getter logic inside the getNewElement callback. 

```py 
onHovercardOpen
  append(user, hovercardLocation)

onProfileLoad
  append(user, profileLocation)

append
  attachElement(
    append: location
    getNewElement
      all creation and update logic

kidonng:
> If we had something like Preact it’d be presumably easier to handle this sort of content.

Or SolidJS, which uses native DOM that maybe interesting for our use case. (Not trying to stir up another UI library war, just mentioning 😂)

fregante (approved):
TL;DReview

fregante:
The problem is that we can’t use JSX (due to the issues described in my last attempt PR) and Solid.js has some footguns when used without it: https://www.solidjs.com/guides/getting-started#buildless-options

Preact doesn’t appear to have this issue. 

I’m not attached to any library but anything that makes it easy, painfree and is under 5kb would be great.

fregante:
I think this would be the correct solution: https://lit.dev/

Unlike most other libraries, these elements are regular dom elements that can be created and appended even with our JSX setup because they’re triggered by a basic `body.append(document.createElement('rgh-time'))` and it automatically detects when the element is removed. 

Any custom-element-based library should work though.

kidonng:
Do we have a list of (relatively) highly dynamic features that will benefit from using a UI library?

fregante:
It’s probably not a whole lot, that’s why library size is important. But if it’s available we’ll surely come up with more. The options page would also definitely benefit from dynamic components, we have a lot of dynamic code in it

PR #5787, opened 2022-07-06
Update README.md to clarify Mac availability

eduwass:
Just a minor nitpick on README. 

From how it's written I thought it was only for iOS but it also supports Mac OS (desktop Safari).

So I just changed the README from saying `also available on iOS!` to `also available on iOS/Mac!` so it's a bit clearer.


fregante:
I disagree, the text is the same for every line:

- [browser] Also on _mobile_

Your suggestion is akin to:

- iPhone: Also on iPhone

fregante:
Will update soon in https://github.com/refined-github/refined-github/pull/5902

Screen Shot

PR #5791, opened 2022-07-07
Add `feature.unload(import.meta.url)`

fregante:
- Complementary to #5726 

In a couple of places we deinit a specific feature after we use it, but in order to do so, we have to reimplement the deinit logic inside the feature itself. This means we rarely do this as it's impractical.

This PR aims to 

- centralize the lifecycle of features into `AbortController`s
- simplify the usage of controllers via `onAbort`
- allow unloading a single feature on demand

The only limitation is that `feature.unload` will disable the entire feature, even if it has multiple `init`s.

## Next

- [x] Use `feature.unload` in at least one feature
- [x] Test further


source/features/bugs-tab.tsx
```diff
@@ -178,6 +178,9 @@ async function init(): Promise<void | false> {
 	}
 
 	await updateBugsTagHighlighting();
+	return () => {
+		console.trace('bugs-tab unloaded')
+	}
```
fregante, line 183: This is a demo `deinit` handler to ensure that the feature is being unloaded when navigating away.

fregante:
Tested, it works, let's see if you have any issues before the next release in august

PR #5793, opened 2022-07-07
Drop `hide-repo-badges` feature

fregante:
As mentioned in https://github.com/refined-github/refined-github/issues/5546, it doesn't work and it became counterproductive with `repo-avatars` in #5778


fregante:
Love the feature verifier 😌

PR #5794, opened 2022-07-07
Add `jump-to-conversation-close-event` feature

### LINKED ISSUE #4773: Jump to Issue/PR closed events
### Description

Comments near closing the issue are often useful.

In the link mentioned, I couldn't find an issue closed event at all.

Unsure on how to handle multiple close-open events.

### Screenshot

_No response_

### Example URL

https://github.com/moby/moby/issues/20316

yakov116:
My first add feature PR in a year! Last one was https://github.com/refined-github/refined-github/pull/4534. 🎉 

Closes #4773

## Test URLs
Closed Issue: https://github.com/refined-github/refined-github/issues/5745
Merged PR: https://github.com/refined-github/refined-github/pull/5752
Closed PR: https://github.com/refined-github/refined-github/pull/5787

## Screenshot
Video_2022-07-07_095050

### DIFF
--- a/source/features/jump-to-conversation-close-event.tsx
+++ b/source/features/jump-to-conversation-close-event.tsx
@@ -1,0 +1,36 @@
+import React from 'dom-chef';
+import select from 'select-dom';
+import {observe} from 'selector-observer';
+import * as pageDetect from 'github-url-detection';
+
+import {wrap} from '../helpers/dom-utils';
+import features from '.';
+
+function init(): Deinit {
+	return observe('#partial-discussion-header :is([title="Status: Closed"], [title="Status: Merged"]):not(.rgh-jump-to-conversation-close-event)', {
+		constructor: HTMLSpanElement,
+		add(messageContainer) {
+			messageContainer.classList.add('rgh-jump-to-conversation-close-event');
+			// Hide the native title
+			messageContainer.style.pointerEvents = 'none';
+			messageContainer.style.cursor = 'default';
+
+			const lastCloseEvent = select.last('.TimelineItem-badge :is(.octicon-issue-closed, .octicon-git-merge, .octicon-git-pull-request-closed)')!.closest('.TimelineItem')!;
+			wrap(messageContainer,
+				<a
+					aria-label="Jump to the last close event."
+					className="tooltipped tooltipped-s"
+					href={location.origin + location.pathname + '#' + lastCloseEvent.id}
+				/>,
+			);
+		},
+	});
+}
+
+void features.add(import.meta.url, {
+	include: [
+		pageDetect.isClosedPR,
+	],
+	deduplicate: 'has-rgh-inner',
+	init,
+});
--- a/readme.md
+++ b/readme.md
@@ -255,6 +255,7 @@ Thanks for contributing! 🦋🙌
 - [](# "esc-to-cancel") [Adds a shortcut to cancel editing a conversation title: <kbd>esc</kbd>.](https://user-images.githubusercontent.com/35100156/98303086-d81d2200-1fbd-11eb-8529-70d48d889bcf.gif)
 - [](# "no-duplicate-list-update-time") [Hides the update time of conversations in lists when it matches the open/closed/merged time.](https://user-images.githubusercontent.com/1402241/111357166-ac3a3900-864e-11eb-884a-d6d6da88f7e2.png)
 - [](# "linkify-user-labels") [Links the "Contributor" and "Member" labels on comments to the author’s commits on the repo.](https://user-images.githubusercontent.com/1402241/177033344-4d4eea63-e075-4096-b2d4-f4b879f1df31.png)
+- [](# "jump-to-conversation-close-event") [Adds a link to jump to the latest close event of a conversation.](https://user-images.githubusercontent.com/19198931/98718312-418b9f00-23c9-11eb-8da2-dfb616e95eb6.gif)
 
 <!-- Refer to style guide above. Keep this message between sections. -->
 
--- a/source/refined-github.ts
+++ b/source/refined-github.ts
@@ -215,3 +215,4 @@ import './features/hide-low-quality-comments';
 import './features/submission-via-ctrl-enter-everywhere';
 import './features/linkify-user-labels';
 import './features/repo-avatars';
+import './features/jump-to-conversation-close-event';


source/features/jump-to-conversation-close-event.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+import {wrap} from '../helpers/dom-utils';
+import features from '.';
+
+function init(): Deinit {
+	return observe('#partial-discussion-header :is([title="Status: Closed"], [title="Status: Merged"]):not(.rgh-jump-to-conversation-close-event)', {
+		constructor: HTMLSpanElement,
+		add(messageContainer) {
+			messageContainer.classList.add('rgh-jump-to-conversation-close-event');
+			// Hide the native title
```
yakov116, line 14: @fregante is this a good idea? We cannot modify the title since that is out detection, but I dont want the title tooltip to show.

yakov116:
> Welcome back!

Thanks! I moved, my computer broke and things just did not settle down until now. 
Hope to be more active now.

source/features/jump-to-conversation-close-event.tsx
```diff
@@ -0,0 +1,36 @@
...
+			messageContainer.style.pointerEvents = 'none';
+			messageContainer.style.cursor = 'default';
+
+			const lastCloseEvent = select.last('.TimelineItem-badge :is(.octicon-issue-closed, .octicon-git-merge, .octicon-git-pull-request-closed)')!.closest('.TimelineItem')!;
+			wrap(messageContainer,
+				<a
+					aria-label="Jump to the last close event."
```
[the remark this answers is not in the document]
yakov116: But this works on Closed issues and Merged PR's

kidonng:
Would love to see a style similar to Discussions:

image

yakov116:
@kidonng we used to have a similar feature
see this

https://github.com/refined-github/refined-github/blob/681bf2c20a14e0212e0724ab1b8cd67784c3fa03/source/features/extend-status-labels.tsx

kidonng:
@yakov116 That feature is dropped not because of style but because GitHub implemented it. This is different.

@fregante we can do it for PR only, which already has the info. And is there any different between who closed an issue or merged a PR?

source/features/jump-to-conversation-close-event.tsx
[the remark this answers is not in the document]
FloEdelmann:
```suggestion
					aria-label="Scroll to most recent close event."
```

fregante:
> Note this will fail on big PR such as #1986

Can you include the test URL at the bottom of the file? Examples

https://github.com/refined-github/refined-github/blob/088ca6f618a80745443c90e663364ea9710abf8c/source/features/suggest-commit-title-limit.tsx#L43
https://github.com/refined-github/refined-github/blob/088ca6f618a80745443c90e663364ea9710abf8c/source/features/suggest-commit-title-limit.tsx#L43
https://github.com/refined-github/refined-github/blob/327747a5f1aff200955580ea8598a74cf17f7ee2/source/features/linkify-user-labels.tsx#L38

No strict formatting rules for this section. It's a useful block of code that can be clicked or copy-pasted in new PRs

source/features/jump-to-conversation-close-event.tsx
fregante: I agree that it's not particularly useful, but let's not mess with it

source/features/jump-to-conversation-close-event.tsx
```diff
@@ -0,0 +1,36 @@
...
+function init(): Deinit {
+	return observe('#partial-discussion-header :is([title="Status: Closed"], [title="Status: Merged"], [title="Status: Closed as not planned"]):not(.rgh-jump-to-conversation-close-event)', {
+		constructor: HTMLSpanElement,
+		add(discussionHeader) {
+			discussionHeader.classList.add('rgh-jump-to-conversation-close-event');
+			// Hide the native title
+			discussionHeader.style.pointerEvents = 'none';
+			discussionHeader.style.cursor = 'default';
+
+			const lastCloseEvent = select.last('.TimelineItem-badge :is(.octicon-issue-closed, .octicon-git-merge, .octicon-git-pull-request-closed, .octicon-skip)')!.closest('.TimelineItem')!;
```
fregante, line 18: :is/:has surely makes selectors unreadable 😰 

Can these be replaced by `color-bg-closed-emphasis` and the purple equivalent?

Also if you'd like to extend our new `selectHas` helper with `selectHas.last` it'd be great. I think it _just_ needs an `if (last) array.reverse()` somewhere in it

source/features/jump-to-conversation-close-event.tsx
```diff
@@ -0,0 +1,36 @@
+import React from 'dom-chef';
+import select from 'select-dom';
+import {observe} from 'selector-observer';
+import * as pageDetect from 'github-url-detection';
+
+import {wrap} from '../helpers/dom-utils';
+import features from '.';
+
+function init(): Deinit {
+	return observe('#partial-discussion-header :is([title="Status: Closed"], [title="Status: Merged"], [title="Status: Closed as not planned"]):not(.rgh-jump-to-conversation-close-event)', {
```
fregante, line 10: We don't need to make this dynamic. Let's just use `select`

source/features/jump-to-conversation-close-event.tsx
```diff
@@ -0,0 +1,36 @@
...
+		constructor: HTMLSpanElement,
```
fregante, line 11: This shouldn't be necessary

source/features/jump-to-conversation-close-event.tsx
```diff
@@ -0,0 +1,36 @@
...
+		add(discussionHeader) {
+			discussionHeader.classList.add('rgh-jump-to-conversation-close-event');
+			// Hide the native title
+			discussionHeader.style.pointerEvents = 'none';
+			discussionHeader.style.cursor = 'default';
```
fregante, line 16: If there are no events, the cursor won't change anyway

fregante (changes requested):

source/features/jump-to-conversation-close-event.tsx
yakov116: The header refreshes on any title change etc. Plus there are 2, one header and one sticky header.

source/features/jump-to-conversation-close-event.tsx
yakov116: I  am changing the style which will give an error on type element

fregante: `HTMLElement`

source/features/jump-to-conversation-close-event.tsx
fregante: ahh you're right. I think we have a listener for that? If not, that's ok, but let's move the callback out of `init`

source/features/jump-to-conversation-close-event.tsx
yakov116: > Can these be replaced by color-bg-closed-emphasis and the purple equivalent?

Would be the same amount of selectors (They each have their own) and I think it would be more fragile.

source/features/jump-to-conversation-close-event.tsx
yakov116: It gets in the way of the other tooltip

fregante: Can you extend the comment explaining why we need to leave the title attribute, why we can’t modify it, and why it needs to be hidden at all?

The current comment only explain what it’s doing but not why

source/features/jump-to-conversation-close-event.tsx
```diff
@@ -0,0 +1,45 @@
...
+import select from 'select-dom';
+import {observe} from 'selector-observer';
+import * as pageDetect from 'github-url-detection';
+
+import {wrap} from '../helpers/dom-utils';
+import features from '.';
+
+function addToConversation(discussionHeader: HTMLElement): void {
+	discussionHeader.classList.add('rgh-jump-to-conversation-close-event');
+	// By hiding the native title we allow our own title (tooltip) to be shown and notify the user of the feature on hover.
```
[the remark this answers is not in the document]
fregante: Now it's not clear how "hiding  native title" is connected to disabling pointerEvents 😅 
```suggestion
	// Avoid native `title` by disabling pointer events, we have our own `aria-label`. We can't drop the `title` attribute because some features depend on it.
```

fregante (approved):

PR #5795, opened 2022-07-07
Improve `readable-title-change-events` order

fregante:
- As suggested in https://github.com/refined-github/refined-github/pull/5703#issuecomment-1176568500

## Test URLs

This page

## Screenshot

Screen Shot 10

### DIFF
--- a/source/features/readable-title-change-events.css
+++ b/source/features/readable-title-change-events.css
@@ -8,7 +8,7 @@
 
 .TimelineItem-body :is(del, ins).markdown-title {
 	flex-basis: 100%;
- 	order: 99;
+	order: 99;
 }
 
 /* TODO: Drop this whole rule when :has() is crossbrowser */
--- a/source/features/readable-title-change-events.css
+++ b/source/features/readable-title-change-events.css
@@ -18,5 +18,5 @@
 
 /*
 Test URLs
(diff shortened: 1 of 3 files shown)


source/features/readable-title-change-events.css
```diff
@@ -0,0 +1,22 @@
...
+}
+
+/* TODO: Drop this whole rule when :has() is crossbrowser */
+.TimelineItem[id^='event-'] .TimelineItem-body :is(ins, del) {
+	display: block !important;
+}
+
+/*
+Test URLs
+https://github.com/refined-github/refined-github/pull/5793#event-6952233065
```
cheap-glitch, line 21:
```suggestion
https://github.com/refined-github/refined-github/pull/5793#event-6952233065
https://github.com/refined-github/refined-github/pull/4030
```

cheap-glitch (approved):
This changes nothing in Firefox, but doesn't seem to break anything either.

source/features/readable-title-change-events.css
fregante: Correct. This is Safari-only at the moment as the label suggests.

fregante: Good suggestion. That PR is enough

```suggestion
https://github.com/refined-github/refined-github/pull/4030#event-4385348325
```

PR #5014, opened 2021-10-30
POC: Use Parcel

fregante:
Annual attempt at migrating to Parcel

- https://github.com/refined-github/refined-github/pull/2983
- https://github.com/refined-github/refined-github/pull/3620

Blocked by:

- https://github.com/parcel-bundler/parcel/issues/4155
- https://github.com/parcel-bundler/parcel/issues/7255

---

- Depends on and closes https://github.com/refined-github/refined-github/issues/4320

Missing (help wanted)

- [ ] Parcel transformer to be able to import the readme
	- discussed in https://github.com/refined-github/refined-github/issues/4320
	- documentation in https://parceljs.org/plugin-system/transformer/
	- usage in https://github.com/parcel-bundler/parcel/issues/4105
- [ ] Tiny function like `getFeatureName(import.meta.url)` to replace `__basefilename`
- [ ] `verify-features.ts` needs to be updated to look for CSS features in `refined-github.css` instead

PR #5804, opened 2022-07-08
Parcel fifteenth time the charm

fregante:
- Follows #5014 

The build works, the readme is part of the bundle, but `import * from './readme.md'` is actually just an empty `{__esModules: true}` object. This suggests that my transformer needs some light fixes.

We're reaaaally close here


fregante:
New issues:

- in build mode `dom-chef` is somehow undefined 🎉 

Side issues:

- Xcode wants a fixed list of files and Parcel renames them every time
  - no native solution for this https://github.com/parcel-bundler/parcel/issues/4553
  - there's a plugin but https://github.com/gxy5202/parcel-namer-hashless/issues/1 🥲 

See you in 2023

fregante:
It's fun to try it every once in a while but honestly the issues I've had trying to port RG to Parcel over the years are… uhm… _entertaining._ Having issues show up only when watching or only when build makes me lose trust in the tool. We can't ship something that only works during development.

fregante:
Once #5801 is in I suppose I could give Vite a try. ESbuilt is still stuck on https://github.com/evanw/esbuild/issues/1492

PR #5806, opened 2022-07-08
Don't add a workflow count to the actions tab in `clean-repo-tabs`

fregante:
I was going to suggest a feature to drop this count when I realized we're adding it.

I don't find it particularly useful and it appears to be noise. 

Drop?

### DIFF
--- a/source/features/clean-repo-tabs.tsx
+++ b/source/features/clean-repo-tabs.tsx
@@ -93,10 +93,10 @@ async function initActions(): Promise<void | false> {
 
 	const actionsCount = await getWorkflowsCount();
 	if (actionsCount > 0 || mustKeepTab(actionsTab)) {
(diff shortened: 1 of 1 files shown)


yakov116:
I use it all the time to download from actions the latest zip or artifact.

Also useful for people that have manual actions.

Don't think we should hide it.

fregante:
This removes the count, not the tab

fregante:
I'd honestly also remove the wiki page count, but that's slightly more useful. We already hide the tab when empty so technically `clean-repo-tabs` would do its dirty job even without the counter.

cheap-glitch:
I'm -0 on this. I think it's nice to have, but I agree it's not that useful since workflows can be very simple just like they can be extremely complex, so that number doesn't really convey much information.

PR #5807, opened 2022-07-09
Lint

fregante:
Leave your tiny changes here

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,4 @@
 * text=auto eol=lf
-*.ai binary
+*.snap linguist-generated=true
 readme.md merge=union
 source/refined-github.css merge=union
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,8 +8,6 @@ env:
 name: Release
 on:
   workflow_dispatch: null
-  schedule:
-    - cron: 42 17 1 * *
 jobs:
   Version:
     outputs:


.gitattributes
```diff
@@ -1,4 +1,4 @@
 * text=auto eol=lf
-*.ai binary
+*.snap linguist-generated=true
```
kidonng, line 2: Side note: `=true` can be omitted ~~like in HTML~~

fregante:
```suggestion
*.snap linguist-generated
```

source/features/first-published-tag-for-merged-pr.tsx
```diff
@@ -77,14 +77,17 @@ function addExistingTagLink(tagName: string): void {
 }
 
 function addLinkToCreateRelease(text = 'Now you can release this change'): void {
+	const url = isRefinedGitHubRepo()
+		? 'https://github.com/refined-github/refined-github/actions/workflows/release.yml'
```
kidonng, line 81: Neat! Now if we have something like `buildRefinedGitHubRepoURL`...

fregante: We do have `getRghIssueUrl` 😃 but no generic version since they're otherwise rare

source/features/first-published-tag-for-merged-pr.tsx
```diff
@@ -11,7 +11,7 @@ import createBanner from '../github-helpers/banner';
 import TimelineItem from '../github-helpers/timeline-item';
 import attachElement from '../helpers/attach-element';
 import {canEditEveryComment} from './quick-comment-edit';
-import {buildRepoURL, getRepo} from '../github-helpers';
+import {buildRepoURL, getRepo, isRefinedGitHubRepo} from '../github-helpers';
 import onConversationHeaderUpdate from '../github-events/on-conversation-header-update';
```
yakov116, line 15:
```suggestion
import onConversationHeaderUpdate from '../github-events/on-conversation-header-update';
import {buildRepoURL, getRepo, isRefinedGitHubRepo} from '../github-helpers';
```

fregante: I think we should stop sorting imports manually 😅 ❌ 🌲

fregante:
Let's keep this rolling in new weekly PRs 🚢 
Keep in mind that lint PRs are for small refactorings and meta changes only. User-facing bug-fixes would be best shown in the release log

Issue #5356, opened 2022-01-31
Many features should not run on certain types of pages


kidonng:
## Completely different

- [x] [404 pages](https://github.com/refined-github/refined-github/blob/76efab19b76ccea4f653a0ef332de98a1593e8c8/source/features/index.tsx#L254-L257)
- [ ] [`gist.github.com`](https://github.com/refined-github/refined-github/pull/5143#discussion_r761667291)

## Insufficient permission

- [ ] Archived repo: #5355
- [ ] Actions that require you to be a collaborator (e.g. editing labels, locking conversations)
- [ ] Actions that require you to be able to change repo settings

### Example URL

N/A

fregante:
I think this was mentioned in https://github.com/refined-github/refined-github/pull/5143/files#r761667291

kidonng:
It is linked as `gist.github.com` above. This issue is about a broader topic.

PR #5814, opened 2022-07-11
Skip some features on private user profiles

kidonng:
Part of #5356 (do I even remember that)

First thing to know: query strings (e.g. `sort=stargazers`) have no effect on private profiles.

- `rgh-sponsor-button`: I'm not sure if users in the program can make their profile private, but there probably won't be a sponsor button there
- `set-default-repositories-type-to-sources`: relies on query strings, so it doesn't work for the repositories tab. There's no organization listed either.
- `show-user-top-repositories`: relies on query strings; "popular repositories" element doesn't exist.
- `user-local-time`: can't get their events/commits
- `user-profile-follower-badge`: can't get their followers

## Test URLs

https://github.com/Centril

## Screenshot

N/A

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,7 +18,7 @@
 				"filter-altered-clicks": "^1.0.1",
 				"fit-textarea": "^2.0.0",
 				"flat-zip": "^1.0.1",
-				"github-url-detection": "^5.12.1",
+				"github-url-detection": "^5.13.0",
 				"image-promise": "^7.0.1",
 				"indent-textarea": "^2.1.0",
 				"js-abbreviation-number": "^1.4.0",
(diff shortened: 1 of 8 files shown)


fregante (commented):
Looks good, but as I had suggested in #5222, I wonder if we should have a global “ignore errors on empty pages” instead of having manually exclude each feature. 

In the runner, we’d have:

``` JavaScript
try {
  init()
  log()
} catch {
    if is empty page or private user {
      log(“feature disabled”) // only if feature logging is active
    } else {
      error()
    }
}

kidonng:
> Looks good, but as I had suggested in #5222, I wonder if we should have a global “ignore errors on empty pages” instead of having manually exclude each feature.

That would better be a catch-all solution, instead of _replacing_ explicit excludes.

PR #5142, opened 2021-12-02
Extend `linkify-user-location` to search results

kidonng:
Extracting stuff from #5113

Also match the style with GitHub on profile page and in hovercards.

## Test URLs

- https://github.com/search?q=sindre&type=users
- https://github.com/somebee

## Screenshot

https://user-images.githubusercontent.com/44045911/144449131-f6883dba-3411-42d7-87ee-5536ca9e7fc8.mov

PR #5816, opened 2022-07-11
Restore `linkify-user-location` on ajax navigation

kidonng:
Well, technically speaking not really a bug, since this type of navigation used to not exist.

Also included style fixes from [`08f6678` (#5113)](https://github.com/refined-github/refined-github/pull/5113/commits/08f667824863fffd935bd62befb5beb8d8e95217) (or dare I say, #5142)

## Test URLs

https://github.com/cheap-glitch

## Screenshot

Before:

https://user-images.githubusercontent.com/44045911/178342110-ed9cc8ce-ab7c-4cbb-bca7-74b8190e6164.mov

After:

https://user-images.githubusercontent.com/44045911/178342127-bdf33db9-2cf6-4440-a46c-e146eef61c44.mov

### DIFF
--- a/source/features/linkify-user-location.tsx
+++ b/source/features/linkify-user-location.tsx
@@ -43,7 +43,7 @@ function hovercardInit(): void {
 }
 
 void features.add(import.meta.url, {
-	init
+	init,
 }, {
 	init: onetime(hovercardInit),
 });
--- a/source/features/linkify-user-location.tsx
+++ b/source/features/linkify-user-location.tsx
@@ -1,6 +1,7 @@
 import React from 'dom-chef';
 import select from 'select-dom';
 import onetime from 'onetime';
+import * as pageDetect from 'github-url-detection';
 
 import features from '.';
 import {wrap, isEditable} from '../helpers/dom-utils';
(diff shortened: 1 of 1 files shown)


source/features/linkify-user-location.tsx
```diff
@@ -26,13 +33,17 @@ const hovercardObserver = new MutationObserver(([mutation]) => {
...
+function hovercardInit(): void {
 	const hovercardContainer = select('.js-hovercard-content > .Popover-message');
 	if (hovercardContainer) {
 		hovercardObserver.observe(hovercardContainer, {childList: true});
 	}
 }
 
 void features.add(import.meta.url, {
-	init: onetime(init),
+	init,
```
kidonng, line 46: I think we should have a `include` here, but where does this feature actually works (excluding the hovercard part)? I can think of user profiles and user followers/following, is there anything else?

source/features/linkify-user-location.tsx
```diff
@@ -26,13 +34,20 @@ const hovercardObserver = new MutationObserver(([mutation]) => {
...
+	include: [
+		pageDetect.isProfile,
```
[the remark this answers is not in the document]
kidonng: The location info is available even for private profile

source/features/linkify-user-location.tsx
fregante: Sounds right. I think this is specifically for profiles and hovercards.

fregante (approved):

PR #5830, opened 2022-07-15
Lint

fregante:
Includes:

- https://github.com/refined-github/refined-github/issues/5222#issuecomment-1184408176


source/features/rgh-linkify-features.tsx
```diff
@@ -64,6 +64,7 @@ void features.add(import.meta.url, {
 		pageDetect.isReleasesOrTags,
 		pageDetect.isCommitList,
 		pageDetect.isSingleCommit,
+		pageDetect.isRepoWiki,
```
kidonng, line 67: 💖 I was going to ask for this! (Since a few days ago when wiki was opened, that is)

source/features/conversation-activity-filter.tsx
```diff
@@ -197,8 +196,8 @@ async function addWidget(header: string, state: State): Promise<void> {
 }
 
 const minorFixesIssuePages = [
-	getRghIssueUrl(5222),
-	getRghIssueUrl(4008),
+	'https://github.com/refined-github/refined-github/issues/5222',
```
kidonng, line 199: It is possible to make this a little nicer, unfortunately still not clickable😅

image

readme.md
```diff
@@ -4,7 +4,7 @@
 
 We use GitHub a lot and notice many annoyances we'd like to fix. So here be dragons.
 
-Our hope is that GitHub will notice and implement some of these much-needed improvements. So if you like any of these improvements, please open a discussion on [GitHub feedback](https://github.com/github-community/community/discussions) or email [GitHub support](https://support.github.com/contact) about doing it.
+Our hope is that GitHub will notice and implement some of these much-needed improvements. So if you like any of these improvements, please open a discussion on [GitHub feedback](https://github.com/github-community/community/discussions) or contact [GitHub support](https://support.github.com/request) about doing it.
```
fregante, line 7: Didn't we just update this URL?

kidonng: It was GitHub feedback URL. I did a more thorough read of README this time

kidonng:
I think we should mention Safari here?

https://github.com/refined-github/refined-github/blob/0741fc984903b14ae1c3bceabbbae0ea05f22d8d/contributing.md?plain=1#L10

fregante:
> `:has()` this September (coincidentally, this is also when Chrome 105 with `:has()` support will be released; not counting Firefox, that is).

10% of our users are on Safari and I'm afraid we won't be able to use a native `has` for a loooong time. They mentioned they have to rewrite their invalidation logic for the selector to work, and they're not even sure it's worth the drawback. 😰 



> I'm not sure about our stance on _latest_ version of Safari.

Almost the same as other browsers. We generally don't require the very last version, there's rarely a new feature that forces us to think about this.

kidonng:
> They mentioned they have to rewrite their invalidation logic for the selector to work, and they're not even sure it's worth the drawback. 😰

At least there's some work being done: https://bugzilla.mozilla.org/show_bug.cgi?id=1771896

> Almost the same as other browsers. We generally don't require the very last version, there's rarely a new feature that forces us to think about this.

Basically I'm just saying Safari is not as "evergreen" as other browsers, in the way new versions are delivered. We have to draw the line _somewhere_.

PR #5831, opened 2022-07-16
Fix `clean-repo-tabs` wiki count

kidonng:
The code assumes the page counter always exists in a wiki, which is not the case if a page named "Home" doesn't exist (and thus failing back to a simple list view).

## Test URLs

- Wiki with "Home" page: https://github.com/pixiebrix/pixiebrix-extension/wiki
- Wiki without "Home" page: https://github.com/refined-github/refined-github/wiki

## Screenshot

image

### DIFF
--- a/source/features/clean-repo-tabs.tsx
+++ b/source/features/clean-repo-tabs.tsx
@@ -50,6 +50,7 @@ const getWikiPageCount = cache.function(async (): Promise<number> => {
 	const dom = await fetchDom(buildRepoURL('wiki'));
 	const counter = dom.querySelector('#wiki-pages-box .Counter');
 
+	// "Home" page may not exist #5831
 	if (counter) {
 		return looseParseInt(counter);
 	}
--- a/source/features/clean-repo-tabs.tsx
+++ b/source/features/clean-repo-tabs.tsx
@@ -47,8 +47,15 @@ function onlyShowInDropdown(id: string): void {
 }
 
 const getWikiPageCount = cache.function(async (): Promise<number> => {
-	const wikiPages = await fetchDom(buildRepoURL('wiki'), '#wiki-pages-box .Counter');
-	return looseParseInt(wikiPages);
+	const dom = await fetchDom(buildRepoURL('wiki'));
+	const counter = dom.querySelector('#wiki-pages-box .Counter');
+
+	if (counter) {
+		return looseParseInt(counter);
+	}
+
+	const wikiPages = dom.querySelector('#wiki-content > .Box > ul');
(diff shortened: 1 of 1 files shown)


fregante:
*Don’t mind me I’m just passively QAing Refined GitHub* 😌

kidonng:
I would say that's expected. Writing our wiki means we'll be dogfooding our features there more, so it's expected [these](https://github.com/refined-github/refined-github/pull/5831) [stuff](https://github.com//refined-github/refined-github/issues/5809) comes up

fregante:
Let's never use Projects then.

source/features/clean-repo-tabs.tsx
```diff
@@ -47,8 +47,16 @@ function onlyShowInDropdown(id: string): void {
...
+	const dom = await fetchDom(buildRepoURL('wiki'));
+	const counter = dom.querySelector('#wiki-pages-box .Counter');
+
+	// "Home" page may not exist #5831
+	if (counter) {
+		return looseParseInt(counter);
+	}
+
+	const wikiPages = dom.querySelector('#wiki-content > .Box > ul');
+	return wikiPages?.childElementCount ?? 0;
```
fregante, line 59: I’d just select the links a return .length

source/features/clean-repo-tabs.tsx
```diff
@@ -47,8 +47,16 @@ function onlyShowInDropdown(id: string): void {
 }
 
 const getWikiPageCount = cache.function(async (): Promise<number> => {
-	const wikiPages = await fetchDom(buildRepoURL('wiki'), '#wiki-pages-box .Counter');
```
fregante, line 50: Unrelated but can you check if there’s an API for this? It looks like the search API has a wikiCount https://docs.github.com/en/graphql/reference/objects#searchresultitemconnection

That’s just about the only wiki reference in the whole GraphQL API (other than “has wiki enabled”)

source/features/clean-repo-tabs.tsx
```diff
@@ -48,7 +47,7 @@ function onlyShowInDropdown(id: string): void {
 
 const getWikiPageCount = cache.function(async (): Promise<number> => {
 	const wikiPages = await fetchDom(buildRepoURL('wiki'), '#wiki-content .Box');
-	return wikiPages.querySelectorAll('.Box-row').length;
+	return wikiPages?.querySelectorAll('.Box-row')?.length ?? 0;
```
fregante: 😂

fregante (approved):
Untested but LGTM

PR #5836, opened 2022-07-19
Make `ci-link` more reliable and avoid duplicates

### LINKED ISSUE #5696: `ci-link` occasionally loads the wrong indicator
1. Visit https://github.com/fregante/content-scripts-register-polyfill/blob/main/ponyfill.ts
2. Click on "Releases" tag

This does not happen if you open the releases in a new tab. It might be due to a mis-ajaxed `releases-tab` button, but I'd like to know _why_ it's getting the second element in the first place.

This bug is only clearly visible when another commit is failing (maybe the second-last? or third?)

https://user-images.githubusercontent.com/1402241/174085713-dd775912-2d12-459a-97b7-b9f7179b9fe3.mov

### Extension version

22.5.22

### Browser(s) used

Safari

### LINKED ISSUE #5755: Duplicate `ci-link`
### Please ensure:

### Example URLs

https://github.com/serenity-rs/serenity

### Extension version

22.6.25

### Browser(s) used

Google Chrome

### Description

image

The bugs tab appears twice on certain repos.

### Include in this issue:

- [x] Screenshots/video/gif demonstrating the bug, if it’s visual
- [ ] Console errors, if any

fregante:
- Fixes #5696 
- Fixes #5755 

`attachElement` to the rescue

## Test URLs

CI:

No CI:
https://github.com/fregante/.github

### DIFF
--- a/source/features/ci-link.tsx
+++ b/source/features/ci-link.tsx
@@ -1,53 +1,51 @@
 import './ci-link.css';
+import React from 'dom-chef';
 import select from 'select-dom';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
-import fetchDom from '../helpers/fetch-dom';
-import getDefaultBranch from '../github-helpers/get-default-branch';
-import {buildRepoURL, getCurrentCommittish} from '../github-helpers';
+import * as api from '../github-helpers/api';
+import {getRepo} from '../github-helpers';
+import attachElement from '../helpers/attach-element';
 
-// Look for the CI details dropdown in the latest 2 days of commits #2990
-const ciDetailsSelector = '.TimelineItem--condensed:nth-of-type(-n+2) batch-deferred-content[data-url$="checks-statuses-rollups"]';
-
-async function getCiDetails(): Promise<HTMLElement | undefined> {
-	if (pageDetect.isRepoHome()) {
-		const ciDetails = select([
-			'.file-navigation + .Box .commit-build-statuses', // Select the CI details if they're already loaded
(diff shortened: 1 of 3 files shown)


source/features/ci-link.tsx
```diff
@@ -1,53 +1,51 @@
...
 import features from '.';
-import fetchDom from '../helpers/fetch-dom';
-import getDefaultBranch from '../github-helpers/get-default-branch';
-import {buildRepoURL, getCurrentCommittish} from '../github-helpers';
-
-// Look for the CI details dropdown in the latest 2 days of commits #2990
-const ciDetailsSelector = '.TimelineItem--condensed:nth-of-type(-n+2) batch-deferred-content[data-url$="checks-statuses-rollups"]';
-
-async function getCiDetails(): Promise<HTMLElement | undefined> {
-	if (pageDetect.isRepoHome()) {
```
fregante, line 14: We should stop having multiple ways to fetch the same content, it makes it really difficult to debug it. We do this in a few places. Maybe the only place where it's worth it is the "current committish" situation but I'm not entirely sure about that either.

source/features/ci-link.tsx
```diff
@@ -1,53 +1,51 @@
...
+import {getRepo} from '../github-helpers';
+import attachElement from '../helpers/attach-element';
 
-	if (pageDetect.isRepoCommitList() && getCurrentCommittish() === await getDefaultBranch()) {
-		const ciDetails = select(ciDetailsSelector);
-		if (ciDetails) {
-			return ciDetails.cloneNode(true);
+async function getHead(): Promise<string> {
+	const {repository} = await api.v4(`
+		repository() {
```
FloEdelmann, line 13: Remember to change that in #5839.

PR #5839, opened 2022-07-19
Meta: Ensure API calls vary depending on the repo (WIP)

fregante:
- Probably solution for https://github.com/refined-github/refined-github/issues/5821

The fix for the issue above is rather easy: change `api.v4`’s  `cacheKey` to include the repo as well as its parameters.

However I didn't think about that until I was half-way through this, and thought it was still worth the simplification.

Every single v4 call using `repository()` (except one) _only_ queried `repository()`, so it made sense to wrap that into a specific helper instead of having a replaceable `repository() {` string. This has the nice side effect that we skip some awkward template strings we had to use. Example below

Untested, quick search and replace. Must be reviewed without whitespace.

### DIFF
--- a/source/features/bugs-tab.tsx
+++ b/source/features/bugs-tab.tsx
@@ -18,20 +18,18 @@ const getBugLabel = async (): Promise<string | undefined> => cache.get<string>(g
 const isBugLabel = (label: string): boolean => supportedLabels.test(label.replace(/\s/g, ''));
 
 async function countBugsWithUnknownLabel(): Promise<number> {
-	const {repository} = await api.v4(`
-		repository() {
-			labels(query: "bug", first: 10) {
-				nodes {
-					name
-					issues(states: OPEN) {
-						totalCount
-					}
+	const labels = await api.v4repository(`
+		labels(query: "bug", first: 10) {
+			nodes {
+				name
+				issues(states: OPEN) {
+					totalCount
 				}
 			}
 		}
 	`);
 
-	const label: AnyObject | undefined = repository.labels.nodes
+	const label: AnyObject | undefined = labels.nodes
 		.find((label: AnyObject) => isBugLabel(label.name));
 	if (!label) {
 		return 0;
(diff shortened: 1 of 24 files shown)


source/features/bugs-tab.tsx
```diff
@@ -18,20 +18,18 @@ const getBugLabel = async (): Promise<string | undefined> => cache.get<string>(g
...
 async function countBugsWithUnknownLabel(): Promise<number> {
-	const {repository} = await api.v4(`
-		repository() {
-			labels(query: "bug", first: 10) {
-				nodes {
-					name
-					issues(states: OPEN) {
-						totalCount
-					}
+	const labels = await api.v4repository(`
```
fregante, line 21: First feature first bug found. 😅  I made the same mistake in many places.

Have not reviewed yet.
```suggestion
	const {labels} = await api.v4repository(`
```

source/features/clean-conversation-filters.tsx
```diff
@@ -8,8 +8,9 @@ import * as api from '../github-helpers/api';
 import {getRepo} from '../github-helpers';
 
 const hasAnyProjects = cache.function(async (): Promise<boolean> => {
+	const {owner, name} = getRepo()!;
 	const {repository, organization} = await api.v4(`
-		repository() {
+		repository(owner: "${owner}", name: "${name}") {
```
fregante, line 13: This is the only exception. The solution is not bad.

source/features/highlight-non-default-base-branch.tsx
```diff
@@ -18,16 +18,12 @@ function isClosed(prLink: HTMLElement): boolean {
 }
 
 function buildQuery(issueIds: string[]): string {
-	return `
-		repository() {
-			${issueIds.map(id => `
+	return issueIds.map(id => `
```
fregante, line 21: This is the nice side effect: no more awkward `array map inside template literal`

fregante:
Closing in favor of a more proper solution https://github.com/refined-github/refined-github/issues/5848

PR #5841, opened 2022-07-20
Fix feature cache across ajaxed page loads

fregante:
Partial and perhaps temporary solution to:

- https://github.com/refined-github/refined-github/issues/5821

We have two separate issues:

- all of our "current repo" API calls use global state
- plenty of our "cached functions" use global state

The first one is easily fixable by including the state in the key itself, manually, when `repository({global state})` is requested. https://github.com/refined-github/refined-github/pull/5839 offers an alternative approach but it requires a lot of testing. I just want to fix most of these issues in today’s release.

The second one is particularly an issue because I think `webext-storage-cache` does not correctly use the provided `cacheKey` for in-memory cache, just for in-storage one. As you can see the `microMemoize` call only uses the parameters as key: https://github.com/fregante/webext-storage-cache/blob/d49454ae720d64f17d3c984734bdf2af22e7e873/index.ts#L160

## Test

0. Clear cache
1. Open https://github.com/refined-github/sandbox/pulls (it uses a non-standard default branch name)
2. Open the notifications
3. Open any other repo
4. Visit the Pull requests page on that repo
5. See `highlight-non-default-base-branch` showing `main` even if it's the default for the current repo

### DIFF
--- a/source/github-helpers/get-default-branch.ts
+++ b/source/github-helpers/get-default-branch.ts
@@ -13,15 +13,9 @@ import {getRepo, getCurrentBranchFromFeed} from '.';
 // "This branch is 1 commit ahead, 27 commits behind master."
 const branchInfoRegex = /([^ ]+)\.$/;
 
-const getDefaultBranch = cache.function(async function (repository?: pageDetect.RepositoryInfo): Promise<string> {
-	if (arguments.length === 0) {
-		repository = getRepo();
-	}
-
-	if (!repository) {
-		throw new Error('getDefaultBranch was called on a non-repository page');
-	}
-
+// DO NOT use optional arguments/defaults in "cached functions" because they can't be memoized effectively
+// https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1864
+const _getDefaultBranch = cache.function(async function (repository: pageDetect.RepositoryInfo): Promise<string> {
 	if (arguments.length === 0 || JSON.stringify(repository) === JSON.stringify(getRepo())) {
 		if (pageDetect.isRepoHome()) {
 			const branchSelector = await elementReady('[data-hotkey="w"]');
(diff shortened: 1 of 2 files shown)


source/github-helpers/get-default-branch.ts
```diff
@@ -59,7 +53,13 @@ const getDefaultBranch = cache.function(async function (repository?: pageDetect.
...
 });
 
-export default getDefaultBranch;
+export default async function getDefaultBranch(repository: pageDetect.RepositoryInfo | undefined = getRepo()): Promise<string> {
+	if (!repository) {
+		throw new Error('getDefaultBranch was called on a non-repository page');
+	}
+
+	return _getDefaultBranch(repository);
+}
```
fregante, line 65: The changes to this file can be reverted once webext-storage-cache is fixed:

- https://github.com/fregante/webext-storage-cache/issues/34

source/github-helpers/get-default-branch.ts
```diff
@@ -13,15 +13,9 @@ import {getRepo, getCurrentBranchFromFeed} from '.';
...
-	if (arguments.length === 0) {
-		repository = getRepo();
-	}
-
-	if (!repository) {
-		throw new Error('getDefaultBranch was called on a non-repository page');
-	}
-
+// DO NOT use optional arguments/defaults in "cached functions" because they can't be memoized effectively
+// https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1864
```
fregante, line 17: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1864 is only an issue because `cacheKey` just isn't followed internally.

fregante:
probably breaking in 251 scenarios while fixing 1

funny-chicken-regret-nothing


release in 3 2 1

kidonng:
I don't think this is actually working

image

fregante:
https://github.com/refined-github/refined-github/issues/5821?

fregante:
This PR only covered the default branch and API calls, so definitely not.

PR #5845, opened 2022-07-21
Lint

fregante:
1. Use https://github.com/fregante/code-tag
2. consolidate hotkey code


source/features/restore-file.tsx
```diff
@@ -76,30 +76,19 @@ async function restoreFile(progress: (message: string) => void, menuItem: Elemen
 	}
 }
 
-const filesRestored = new WeakSet<HTMLButtonElement>();
```
fregante, line 79: This was needed because we kept the dropdown open, I think. Since the dropdown now closes on click, the user cannot double click. By the time they manage to open the menu and click again, it's probably done and the file is gone.

source/features/restore-file.tsx
```diff
@@ -76,30 +76,19 @@ async function restoreFile(progress: (message: string) => void, menuItem: Elemen
...
 
-	// Only allow one click
-	if (filesRestored.has(menuItem)) {
-		return;
-	}
-
-	filesRestored.add(menuItem);
-
-	try {
+	await showToast(async progress => {
```
fregante, line 82: `showToast` already acts as its own try/catch, with errors being shown in the toast itself. No need for an extra try/catch as long as everything is inside.

source/github-helpers/hotkey.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'dom-chef';
+
+export function registerHotkey(hotkey: string, action: VoidFunction | string): Deinit {
+	const element = typeof action === 'string'
+		? <a hidden href={action} data-hotkey={hotkey}/>
```
fregante, line 5: I also added support for strings, but haven't found any use for it where it would simplify the code.

kidonng: What if we treat this string as a selector and do `select(action).click()`?

To emulate link behavior, just pass an action of `() => location.assign(link)`.

fregante: > What if we treat this string as a selector and do `select(action).click()`?

addHotkey already does that:
```js
registerHotkey('ctrl-a', 'a[blah]') // Your suggestion
addHotkey(select('a[blah]'), 'ctrl-a') // Already-possible solution
```

> To emulate link behavior, just pass an action of `() => location.assign(link)`.

There's no substitute for real links (which can be handled by Turbo as well).

I do see two possible improvements here:

- make the two functions consistent (always accept "hotkey" first), or 
- merge the two functions so `registerHotkey` accepts an element passed as "action"

fregante:
Will merge within an hour

fregante:
Screen Shot 18

Screen Shot 19

I let you all down and I apologize.

PR #5853, opened 2022-07-26
Only run `useful-forks` on public repositories

kidonng:
Waiting for https://github.com/refined-github/github-url-detection/pull/140

~~At times like this I want to `npm install refined-github/github-url-detection#isPublicRepo` and chill~~

## Test URLs

https://github.com/username?tab=repositories&type=private


fregante:
I mean… where did this bother you? Why archive a private repo in the first place? 😅 I’d place this in the lint PR if you reeeeally want to make this change.

kidonng:
> I mean… where did this bother you? Why archive a private repo in the first place?

But they are for different usages. Private means "I don't want to show it" and Archived means "Don't bother with it anymore". If I run a script or install a integration for all my repositories, surely I will want to skip archived repos instead of manually excluding them every time.

> I’d place this in the lint PR if you reeeeally want to make this change.

Then consider this pull request "_Skip some features on private repositories_". I will merge it to the lint PR if I don't find anything else worth changing.

tyrann0us:
> I mean… where did this bother you? Why archive a private repo in the first place? 😅 I’d place this in the lint PR if you reeeeally want to make this change.

We are a web (WordPress) agency and there are multiple reasons why we need to archive a private repository:
1. Discourage usage of reusable, but highly outdated (in terms of code quality) packages
2. Discourage usage of reusable packages which no longer have a purpose, e.g. fix has landed upstream (WordPress)
3. If a client cancells SLA, we offboard them and archive all client-specific repositories (which are private)

To give you some numbers, we currently have ~1400 repositories.
* Public: ~60
* Private ~1300
* Archived: ~450
* _Private archived_: ~430 of the ~450

So **TL;DR**, for us it would make much sense to not show the useful forks link for private archived repositories.
Thanks!

kidonng:
Since bumping `github-url-detection` here breaks other features, I will do it in the lint PR.

@tyrann0us thanks for sharing your experience! I will admit I haven't do active searching of what other features may not be useful on private repos yet 😴 (I don't have that many, to begin with). If you noticed any please let us know.

PR #5854, opened 2022-07-26
Less attention grabbing hint to draft a new release

lode:
Fixes https://github.com/refined-github/refined-github/issues/5851.

On merged PRs which are not (yet) part of a release a banner is shown to hint to create a release. This PR changes it from a banner & button, to regular text & link.

I also changed the wording a little, but if this is not wished it can be reversed of course. I did it can be confused for other forms of 'releasing', like deploying. I didn't remove the word 'release', just used other language to make it more understood as a noun.

I'm unsure about the conventions of html elements & classes. I borrowed a class from GitHub to make the text color a little more muted. Is that okay or should a RG class be made & used?

## Test URLs

It can be seen on any merged PR which is not part of a release. _Note that due to https://github.com/refined-github/refined-github/pull/5850, the banner/text isn't shown when the repository doesn't use releases at all._

Current example: https://github.com/refined-github/refined-github/pull/5850.

## Screenshots

From current situation:
afbeelding

To new situation:
afbeelding

### DIFF
--- a/source/features/first-published-tag-for-merged-pr.tsx
+++ b/source/features/first-published-tag-for-merged-pr.tsx
@@ -91,7 +91,7 @@ async function addLinkToCreateRelease(text = 'Now you can release this change.')
 				<p className="TimelineItem-body">
 					{text}
 					<a href={url}>
-						<TagIcon/> Draft a new release
+						Draft a new release
 					</a>
 				</p>
 			</TimelineItem>
--- a/source/features/first-published-tag-for-merged-pr.tsx
+++ b/source/features/first-published-tag-for-merged-pr.tsx
@@ -36,7 +36,7 @@ async function init(): Promise<void> {
 	if (tagName) {
 		addExistingTagLink(tagName);
 	} else if (canCreateRelease()) {
-		void addLinkToCreateRelease('This pull request seems to be unreleased');
+		void addLinkToCreateRelease('This pull request seems to be unreleased.');
 	}
 }
 
@@ -76,7 +76,7 @@ function addExistingTagLink(tagName: string): void {
 	});
 }
 
-async function addLinkToCreateRelease(text = 'Now you can release this change'): Promise<void> {
+async function addLinkToCreateRelease(text = 'Now you can release this change.'): Promise<void> {
 	if (await getReleaseCount() > 0) {
 		return;
 	}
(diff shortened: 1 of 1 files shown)


fregante:
It looks good, but I’m not sure the icon works there. Should we use .Btn-link on the A tag?

lode:
I can agree, the icon is a little weird for an inline link. I will remove.

Adding `.btn-link` causes a background-color when hovering, that's not really nice I think.

fregante:
Not a super fan of how it looks/aligns to be honest 😅 

Screen Shot 8

_I added some spacing next to the link in my last commit_

fregante (changes requested):
I think we should use the same style/alignment as the other two events visible in my screenshot, including the button (but no horizontal rule/line)

lode:
Yeah, good points. The inconsistency is also not nice, it becomes more busy if we add yet another style. Since it is not an event the banner styling is more correct. If I understand you correctly, we can just keep the styling as is, but remove the background color? (Like in the 'Pull request closed' banner, but then with the button and right-floating as currently.)

fregante:
I think we can reuse the whole markup I pasted. It's essentially the same concept:

- PR merged, delete fork?
- PR merged, create release?

Generally these two won't appear at the same time because people who send PRs from temporary forks aren't the same who can create a release. 

The same bubble style could also be ok for every usage in the feature though.

lode:
I'm not sure I understand. The screenshot you attached, looks the same as the current situation, doesn't it?

fregante:
> I'm not sure I understand. The screenshot you attached, looks the same as the current situation, doesn't it?

I'm talking about the "Pull request closed" style. It's true that it's similarly-sized, but it's not colored and it's what GitHub uses for a very similar intent: _You can do something now that the PR has been merged._

fregante:
Or like this, which I mentioned as an alternative:

Screen Shot

lode:
Ah, I understand.

Some thoughts I have about these:

- The PR closed, delete the fork style: it does worry me a little that there is a chance both banners would be shown at the same time, and then the speech bubble is a bit weird.
- The alternative deleted branch style: it doesn't feel correct, since that is about an event placed on the timeline, which this is not.

I'm more inclined to keep the same situation as now, but with a transparent instead of a blue background. This:

- keeps the styling similar to how it is now, not changing too much for current users
- keeps the styling in sync with the first-appeared-in banner, which is nice I think
- gives it less attention, just like GH decided upon for the delete fork banner with similar intent

Although, thinking about the last point. Deleting a fork is less important than drafting a release. So maybe the extra attention is worth it. I'm also fine closing this PR. I think it was good that the banner was removed if a repo doesn't have releases at all, with that this might be good as is.

fregante:
> both banners would be shown at the same time, 

Super rare. People who can release a new versions don't generally use forks. If they do, that's their own issue 😃 

> that is about an event placed on the timeline, which this is not.

I agree, that's why the bubble works, exactly like GitHub is doing.

fregante:
Feel free to send a new PR

PR #5886, opened 2022-08-03
Replace `selector-observer`

### LINKED ISSUE #5874: Wrap `selector-observer` to add `AbortSignal`, better types, etc
I recently mentioned this in a comment but I can't find it anymore. @cheap-glitch also mentioned it in https://github.com/refined-github/refined-github/pull/5726#pullrequestreview-1019332862

Here's part of what I had in mind for a such a wrapper:

- accept `signal`
- use https://github.com/g-plane/typed-query-selector to automatically get types instead of using `constructor`
- better API (`observeElement('a', cb)` instead of `observe('a',{add: cb})`)
- more I can't remember 🥲

fregante:
- Closes #5874 

Not tested.

If the WeakSet doesn't match existing elements, the function will also need to attach a temporary class because I removed our custom `:not(rgh-etc)` selectors

### DIFF
--- a/source/helpers/selector-observer.tsx
+++ b/source/helpers/selector-observer.tsx
@@ -7,10 +7,9 @@ import {ParseSelector} from 'typed-query-selector/parser';
 import {getSnapshotUUID} from './attach-element';
 
 const animation = 'rgh-selector-observer';
-const seen = new Map<string, WeakSet<EventTarget>>();
 const getListener = mem(<ExpectedElement extends HTMLElement>(seenMark: string, selector: string, callback: (element: ExpectedElement) => void) => function (event: AnimationEvent) {
 	const target = event.target as ExpectedElement;
-	if (seen.get(selector)?.has(target) || !target.matches(selector)) {
+	if (!target.matches(selector)) {
 		return;
 	}
 
@@ -20,8 +19,6 @@ const getListener = mem(<ExpectedElement extends HTMLElement>(seenMark: string, 
 	if (!/shortenLink|bypass/.test(callback.toString())) {
 		console.log('selector-observer', target, callback);
 	}
-
-	seen.get(selector)!.add(target);
 
 	callback(target);
 });
@@ -51,7 +48,6 @@ export default function observe<
 			animation: 1ms ${animation};
 		}
 	`);
-	seen.set(String(selector), new WeakSet());
 	getTag().append(rule);
 	signal?.addEventListener('abort', () => {
 		rule.remove();
--- a/source/helpers/selector-observer.tsx
+++ b/source/helpers/selector-observer.tsx
@@ -16,10 +16,6 @@ const getListener = mem(<ExpectedElement extends HTMLElement>(seenMark: string, 
 	// Removes this specific selector’s animation once it was seen
 	target.classList.add(seenMark);
 
-	if (!/shortenLink|bypass/.test(callback.toString())) {
-		console.log('selector-observer', target, callback);
-	}
-
 	callback(target);
 });
 
@@ -42,7 +38,7 @@ export default function observe<
 	}
 
 	const selector = String(selectors); // Array#toString() creates a comma-separated string
-	const seenMark = 'rgh-' + getSnapshotUUID();
+	const seenMark = 'rgh-seen-' + getSnapshotUUID();
 	const rule = new Text(css`
 		:where(${String(selector)}):not(.${seenMark}) {
 			animation: 1ms ${animation};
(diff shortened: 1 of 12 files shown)


fregante:
Already found two pretty serious limitations:

- it won't detect hidden elements (which can be an advantage _sometimes_)
- you can't observe the same element twice (e.g. `shorten-links` applies to every markdown link, so that breaks a lot of features)

I think the "freedom" that an extremely-lightweight solution grants us is completely taken away from the fact that any overlapping selectors may at any time just break.

We _could_ be careful about this, but there's no way we'd actually be successful at this. A 1-character bugfix could break another feature and we wouldn't know until much later.

source/helpers/selector-observer.tsx
```diff
@@ -0,0 +1,54 @@
...
+const tracked = new Map<string, WeakSet<EventTarget>>();
+const getListener = mem(<ExpectedElement extends HTMLElement>(selector: string, callback: (element: ExpectedElement) => void) => function (event: AnimationEvent) {
+	const target = event.target as ExpectedElement;
+	if (tracked.get(selector)?.has(target) || !target.matches(selector)) {
+		return;
+	}
+
+	if (!/shortenLink|bypass/.test(callback.toString())) {
+		console.log('selector-observer', target, callback);
+	}
```
fregante, line 17: Test/log code, will remove

fregante (commented):
I found a solution to the multiple-observers problem and also the compatibility with any other native animations:

- it doesn't matter what animation is triggered, the observer just needs _any_ animation to start in order to find the element

However… I just realized that there's still another problem with this solution: 

1. an animation starts on an element
2. the selector isn't matched yet
3. a class is added
4. the selector is matched
5. the animation doesn't start again

I don’t think there's a solution here. However I slice it, there can only be one animation/trigger per element

fregante:
Ok I found another solution for the issue I described, which also covers the possible issue I mentioned earlier:

> If the WeakSet doesn't match existing elements, the function will also need to attach a temporary class because I removed our custom `:not(rgh-etc)` selectors

In short, like before, once the element is seen, it's marked as seen via `class` for the specific callback. So if a future selector is matched, the `animation` is started again.

The only situation where this workaround fails is on elements that have a permanent/native `animation`, which GitHub uses sparingly. I think that's an acceptable risk at this point.

fregante:
I will merge this PR this weekend, reviews welcome before then.

source/helpers/selector-observer.tsx
```diff
@@ -0,0 +1,52 @@
...
+	document.head.append(style);
+	return style;
+});
+
+export default function observe<
+	Selector extends string,
+	ExpectedElement extends HTMLElement = ParseSelector<Selector, HTMLElement>,
+>(
+	selectors: Selector | readonly Selector[],
+	listener: (element: ExpectedElement) => void,
```
fregante, line 33: Known issue: the type is suggested here:

```js
observe('a', (a /* It's HTMLAnchorElement */) => {
	console.log(a);
});
```

But it's not enforced:

```js
observe('a', (a: HTMLParagraphElement /* This is also accepted */) => {
	console.log(a);
});
```

fregante: never mind, fixed

source/helpers/selector-observer.tsx
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	const selector = String(selectors); // Array#toString() creates a comma-separated string
+	const seenMark = 'rgh-seen-' + getSnapshotUUID();
+	const rule = new Text(css`
+		:where(${String(selector)}):not(.${seenMark}) {
+			animation: 1ms ${animation};
+		}
+	`);
+	getTag().append(rule);
```
fregante, line 46: Here's the `style` tag update in action, as I navigate GitHub:

https://user-images.githubusercontent.com/1402241/183035922-d7fc14a3-4d6f-471b-af90-be4d03de859c.mov

kidonng:
> * it doesn't matter what animation is triggered, the observer just needs _any_ animation to start in order to find the element

Brilliant! I think it's an acceptable cost having to do `matches` and `classList.add` ourselves, for such a simple solution. Native animations not added by the observer are also sparse so not much performance is wasted on checking them.

> The only situation where this workaround fails is on elements that have a permanent/native `animation`, which GitHub uses sparingly. I think that's an acceptable risk at this point.

Agreed, I don't think any of our targets are using animations. In the future we can also use `@layer` to avoid breaking existing animations, if any.

PR #5902, opened 2022-08-06
Lint

fregante:


readme.md
```diff
@@ -408,28 +408,11 @@ Thanks for contributing! 🦋🙌
...
-- [Notifier for GitHub](https://github.com/sindresorhus/notifier-for-github-chrome) - Shows your notification unread count
-- [Mottie's Userscripts](https://github.com/Mottie/GitHub-userscripts) - Countless features to improve GitHub granularly
-- [Contributors on GitHub](https://github.com/hzoo/contributors-on-github) - Shows stats about contributors
-- [Hide Files on GitHub](https://github.com/sindresorhus/hide-files-on-github) - Hides dotfiles from the file browser
-- [GitHub Issue Link Status](https://github.com/fregante/github-issue-link-status) - Colorize issue and PR links to see their status (open, closed, merged)
-- [Notifications Preview for GitHub](https://github.com/tanmayrajani/notifications-preview-github) - See your notifications on hover on all pages
-- [Show All GitHub Issues](https://github.com/sindresorhus/show-all-github-issues) - Shows both Issues and Pull Requests in the Issues tab
-- [Sourcegraph](https://github.com/sourcegraph/sourcegraph/tree/main/client/browser) - Code search and navigation tool
-- [PR Monitor](https://github.com/fwouts/prmonitor) - Notifies you about incoming and outgoing PRs
```
kidonng, line 435: Neat! I always find this part a little bit weird, we are not talking about two or three but (almost) a dozen 😅 .

fregante (commented):
@[]()anyone: feel free to merge at any time

source/features/jump-to-conversation-close-event.tsx
```diff
@@ -40,7 +40,8 @@ void features.add(import.meta.url, {
 		pageDetect.isConversation,
 	],
 	include: [
-		pageDetect.isClosedConversation,
+		pageDetect.isClosedIssue,
+		pageDetect.isClosedPR,
```
kidonng, line 44: Bumped `github-url-detection` and resolved all breaking changes. I'm separating `isClosedConversation` changes here since this is not a simple search & replace, but reverting [`9e8d7a8` (#5807)](https://github.com/refined-github/refined-github/pull/5807/commits/9e8d7a8128d44ef77c103c2f88f0377ee119c2f1). As mentioned in https://github.com/refined-github/github-url-detection/pull/136#issuecomment-1193295489, the only valid use case is in this feature (`jump-to-conversation-close-event`).

source/features/clean-repo-tabs.tsx
```diff
@@ -116,7 +112,7 @@ async function initProjects(): Promise<void | false> {
 		return;
 	}
 
-	if (await canUserEditOrganization()) {
+	if (pageDetect.isUserTheOrganizationOwner()) {
```
fregante, line 115: `canUserEditOrganization` was renamed in the package because it was incorrect, but this is not the same as `isUserTheOrganizationOwner`

- https://github.com/refined-github/github-url-detection/pull/87
- https://github.com/refined-github/refined-github/pull/4534#issuecomment-878059232

If anything, we need to add `canUserEditOrganization () => select.exists('.btn-primary[href$="repositories/new"]')`. We didn't back then because we needed elementReady, which isn't yet available:

- https://github.com/refined-github/github-url-detection/issues/85

kidonng: Thanks for explaining, I'm confused as I think if a user can edit organization, they are certainly a owner. Now that I take a look at the actual selector, I think the name should be something else. Being able to create a repo only means they have org-wide write access. Also, the existing `canUserEditRepo` is checking for a settings tab, which only exists for those who can manage the repo, not just write access.

fregante:
This PR is ready to go 🚀 

Feel free to open a new one as needed

PR #5912, opened 2022-08-16
Fix `repo-avatars` alignment

kidonng:
A small bug introduced in #5885.

## Test URLs

Here

## Screenshot

<table>
<tr>
	<td>Before
	<td>image
<tr>
	<td>After
	<td>image
</table>


fregante (approved):

fregante:
Can you also hotfix it for 22.8.8? It'll probably be a while until we release a feature given the lack of PRs 😅

PR #5913, opened 2022-08-16
Meta: Cleanup readme footer

fregante:


fregante:
I forgot we had a lint PR open. I merged it into that.

PR #5919, opened 2022-08-19
Support "For you" feed in `infinite-scroll`

### LINKED ISSUE #5716: Support “For you” in `infinite-scroll`
### Please ensure:

### Example URLs

https://github.com/

### Extension version

22.5.22

### Browser(s) used

Chrome

### Description

1. Go to the "Following" tab on your GitHub dashboard
2. Reload
3. Notice that the infinite scroll is working here
4. Go to the "For You" tab on your GitHub dashboard
5. Note that the infinite scroll isn't working here
6. Reload
7. Note that the infinite scroll is working now (sometimes it doesn't work for infinite scroll either)

### Include in this issue:

- [ ] Console errors, if any

kidonng:
Resolve #5716

## Test URLs

https://github.com/

## Screenshot

https://user-images.githubusercontent.com/44045911/185581428-71a60c63-9803-4b7c-8b2b-acdc14311120.mov

### DIFF
--- a/source/features/infinite-scroll.tsx
+++ b/source/features/infinite-scroll.tsx
@@ -2,10 +2,10 @@ import './infinite-scroll.css';
 import React from 'dom-chef';
 import select from 'select-dom';
 import debounce from 'debounce-fn';
-import {observe} from 'selector-observer';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
+import observe from '../helpers/selector-observer';
 
 const loadMore = debounce(() => {
 	const button = select('[role="tabpanel"]:not([hidden]) button.ajax-pagination-btn')!;
@@ -29,10 +29,8 @@ const inView = new IntersectionObserver(([{isIntersecting}]) => {
 });
 
 function init(): Deinit {
-	const selectorObserver = observe('.ajax-pagination-btn', {
-		add(button) {
-			inView.observe(button);
-		},
+	observe('.ajax-pagination-btn', button => {
+		inView.observe(button);
 	});
 
 	// Use cloneNode to keep the original ones for responsive layout
@@ -54,7 +52,6 @@ function init(): Deinit {
 
 	return [
 		inView,
-		selectorObserver,
 	];
 }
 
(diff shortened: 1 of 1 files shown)


fregante (approved):
Does it work well when switching between tabs? Can you make sure that listeners aren't piling up?

kidonng:
It works. The two buttons are always in the DOM so listeners are only added once.

source/features/infinite-scroll.tsx
```diff
@@ -29,10 +29,8 @@ const inView = new IntersectionObserver(([{isIntersecting}]) => {
 });
 
 function init(): Deinit {
-	const selectorObserver = observe('.ajax-pagination-btn', {
-		add(button) {
-			inView.observe(button);
-		},
+	observe('.ajax-pagination-btn', button => {
+		inView.observe(button);
 	});
```
kidonng, line 34: Actually this can be two `element-ready` calls, but our selector-observer is so lightweight 😆

source/features/infinite-scroll.tsx
```diff
@@ -29,10 +29,8 @@ const inView = new IntersectionObserver(([{isIntersecting}]) => {
 });
 
 function init(): Deinit {
-	const selectorObserver = observe('.ajax-pagination-btn', {
-		add(button) {
-			inView.observe(button);
-		},
+	observe('.ajax-pagination-btn', button => {
```
fregante, line 32: I think this should use the `signal` passed to `init`

source/features/infinite-scroll.tsx
fregante: Just realized yet another limitation of animations: they don't start when the tab is hidden.

This is the same rAF limitation described here:

- https://github.com/refined-github/refined-github/issues/4312#issuecomment-831007823

kidonng: You have stated it here: https://github.com/refined-github/refined-github/pull/5886#issuecomment-1204363045

It's fine in this case, we don't need to load a hidden tab.

source/features/infinite-scroll.tsx
kidonng: That's new, I will use it

PR #5923, opened 2022-08-23
Fix: Repo tabs are hidden in Actions on Firefox

fregante:
This rule was conflicting with GitHub’s own overflow-nav logic, which ended up hiding every tab.

- Fixes https://github.com/refined-github/refined-github/issues/5921
- Reverts the ancient https://github.com/refined-github/refined-github/issues/3548

## Test URLs

- https://github.com/jerone/eslint-formatter-checklist/runs/7950750517?check_suite_focus=true

## Before

Screen Shot 3

## After

Screen Shot

### DIFF
--- a/source/features/more-dropdown-links.css
+++ b/source/features/more-dropdown-links.css
@@ -3,7 +3,6 @@
 	visibility: visible !important;
 }
 
-
 /* Only show the divider if there are other items above it */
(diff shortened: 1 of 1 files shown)


fregante:
cc @jerone

jerone:
✅ If I disable this style via Dev Tools, the issue indeed disappears.

PR #5930, opened 2022-08-25
Fix position of sticky file headers after opening notification

cheap-glitch:
Fixes https://github.com/refined-github/refined-github/issues/5914

## Test URLs

1. Open a notification pointing to a PR (or PR files)
2. Go to the Files tab
3. This still happens if you go back to the "Conversation" tab, reload the page, and go to "Files" again

## Screenshot

**Before**

01-before

**After**

02-after

### DIFF
--- a/source/features/sticky-file-header.css
+++ b/source/features/sticky-file-header.css
@@ -5,7 +5,7 @@
 	top: 0;
 }
 
-.notification-shelf ~ .application-main .file-header {
+.notification-shelf ~ .application-main .file-header:not(.sticky-file-header) {
 	top: 129px;
 }
 


source/features/sticky-file-header.css
```diff
@@ -5,7 +5,7 @@
 	top: 0;
 }
 
-.notification-shelf ~ .application-main .file-header {
+.notification-shelf ~ .application-main .file-header:not(.sticky-file-header) {
 	top: 129px;
 }
```
cheap-glitch, line 10: Apparently GitHub uses JS to compute and inject the correct value for `top` in each of the file header elements' `style` attribute :grimacing: 

This rule was superseded by those inline styles but still messed with their computations, resulting in a value of `198px` instead of the correct `129px`.

fregante:
Thank god

fregante (approved):
I broke this probably a month ago 🥲 


Would you be able to push a hotfix for 22.8.1 and 22.8.23? I think it looks like:

```css
.notification-shelf ~ .application-main .file-header.sticky-file-header {
 	top: 198px;
}
```

cheap-glitch:
> I think it looks like:

No, a bigger value for `top` will only make it worse :smile: 

Just tested the rule but with `60px` (height of the sticky PR header) and it seems to work fine. Will push the hotfix asap.

cheap-glitch:
Hmm, weird I pushed some hotfixes but even after clearing the cache they don't get downloaded :thinking: 

@fregante Are you seeing them? Should be present for the last three versions

fregante:
> Are you seeing them?

Are you using the dev version? They're not applied 0.0.0

cheap-glitch:
> Are you using the dev version? They're not applied to 0.0.0

I know, which is why I tried in my regular browser profile but no luck. And I know the CSS is valid because adding it to my custom CSS fixes things :man_shrugging: 

Might be an issue with the download/storage of the hotfixes?

fregante:
> Wait, did the hotfixes move to the new `yolo` repo?

Yes, we'll drop the hotfixed branch soon.


> Edit n°2: Added the hotfix to `yolo` but still don't see it 😢

I see it on Safari 22.8.1:

Screen Shot


Run anywhere to find out if the cache was downloaded (keep in mind that you need to open GitHub.com after clearing your cache)

```js
chrome.storage.local.get(console.log)
```

fregante:
The hotfix works for me:

- red: correct position
- blue: hotfix, in `style` tag
- purple: our previous style from refined-github.css

Screen Shot 1

fregante:
The only related bug I'm experiencing is https://github.com/refined-github/refined-github/issues/5857, because the its bug fix was in 22.8.8 but Safari is on 22.8.1 🥲

PR #5873, opened 2022-07-30
Don't deduplicate event-only features

fregante:
- Partial fix for #5871 

Drops deduplication from:

- delegate-only features
- onlyAdditionalListener features

Mixed "change dom + delegate" features are still broken as per #5871, they will need changes like:

- https://github.com/refined-github/refined-github/pull/5872
- custom deduplication detection

I think basically we need to move away from `has-rgh` and its affordances

PR #5931, opened 2022-08-25
Restore `show-names` feature

### LINKED ISSUE #5927: `show-names` is broken
### Description

Names aren't shown anywhere anymore
Screen Shot

- https://github.com/
- This issue

### Extension version

21.8.1, 21.8.8

### Browser(s) used

Safari, Chrome

cheap-glitch:
Fixes #5927 

## Test URLs

- This page
- Your newsfeed

## Screenshot

**Conversation**

image

**Newsfeed**

image

### DIFF
--- a/source/features/show-names.tsx
+++ b/source/features/show-names.tsx
@@ -84,7 +84,6 @@ void features.add(import.meta.url, {
 	include: [
 		pageDetect.hasComments,
 	],
-	onlyAdditionalListeners: true,
 	deduplicate: false,
 	init,
 });
--- a/source/github-events/on-newsfeed-load.ts
+++ b/source/github-events/on-newsfeed-load.ts
@@ -1,17 +1,17 @@
 import select from 'select-dom';
 
-export default async function onNewsfeedLoad(callback: VoidFunction): Promise<void> {
-	const observer = new MutationObserver((([{addedNodes}]) => {
-		callback();
+import {createFragmentLoadListener} from './on-fragment-load';
 
-		// If the newly-loaded fragments allows further loading, observe them
-		for (const node of addedNodes) {
-			if (node instanceof Element && select.exists('.ajax-pagination-form', node)) {
-				observer.observe(node, {childList: true});
-			}
-		}
-	}));
+function onNewsfeedPageLoad(event: Event, callback: EventListener, signal: AbortSignal): void {
+	select('.news .ajax-pagination-form[action^="/dashboard-feed"]')?.addEventListener('page:loaded', event => { // Older newsfeed events are paginated through a <form> element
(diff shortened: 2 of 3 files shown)


source/features/show-names.tsx
```diff
@@ -84,7 +84,6 @@ void features.add(import.meta.url, {
 	include: [
 		pageDetect.hasComments,
 	],
-	onlyAdditionalListeners: true,
```
cheap-glitch, line 87: The feature was inadvertently disabled by #5873 (https://github.com/refined-github/refined-github/pull/5873/files#diff-bb287cee7084ff807d23e0ac0cdee19825f4c215ce976f3ab5f608448c4bc8eaR87)

source/github-events/on-newsfeed-load.ts
```diff
@@ -1,17 +1,17 @@
...
-	// Start from the main container
-	observer.observe(select('.news')!, {childList: true});
+	callback(event);
+}
+
+export default function onNewsfeedLoad(callback: EventListener, signal: AbortSignal): void {
+	createFragmentLoadListener('.news include-fragment[src="/dashboard-feed"]', event => {
+		onNewsfeedPageLoad(event, callback, signal);
+	}, signal);
 }
```
cheap-glitch, line 17: Rewrote the broken `onNewsfeedLoad` listener (only used by `show-names`) to use GitHub events instead of a `MutationObserver`.

fregante: Was this broken too?

We should use the new selector observer wherever possible and move away from other listeners, especially when nested like this one.

But if this already fixes an issue and it works, it's fine

cheap-glitch: > Was this broken too?

Yes, since GitHub changed the DOM a lot when introducing the "For you" tab, and now there's a lot more nested elements that are progressively loaded and don't have very specific classes/attributes to target.

This could also have been fixed by adding `subtree: true` to the listener, but listening to events and running the feature only when needed is probably a lot more efficient.

fregante: > This could also have been fixed by adding `subtree: true` to the listener

See #5886. We don't have to think about that anymore 🎉 

The new observer is not strictly equivalent however, for example the elements must be "not hidden" or else the observer can't see them.

fregante (approved):
It'd be great to see the new observer instead, but merge at will.

cheap-glitch:
I merged it for now, will open a meta PR when I have a bit more time.

This will be a good way of getting my feet wet with the new observer :slightly_smiling_face:

fregante:
> It'd be great to see the new observer instead, but merge at will.

Done in https://github.com/refined-github/refined-github/pull/5991

PR #5933, opened 2022-08-25
Restore `center-reactions-popup` feature

cheap-glitch:
Fix https://github.com/refined-github/refined-github/issues/5881

## Test URLs

- This page
- [Split view with review comments](https://github.com/refined-github/sandbox/pull/38/files?diff=split)

## Screenshots

**Regular comment**

image

**Review comment in split diff view**

image

### DIFF
--- a/source/features/center-reactions-popup.css
+++ b/source/features/center-reactions-popup.css
@@ -3,9 +3,6 @@
 	left: 50% !important;
 	right: auto !important;
 	transform: translate(-50%) !important;
-}
-
-/* Avoid clipping popup of comments on the right side of split file views - https://github.com/refined-github/sandbox/pull/38/files?diff=split */
-.line-comments[data-split-side='right'] .dropdown-menu-reactions {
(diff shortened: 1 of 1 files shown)


cheap-glitch (commented):
Dropped a lot of code since the feature is 3~4 years old and most of it is outdated.

fregante:
Could you also add a width on the popup so they wrap like they used to? I feel that the single line is the most annoying part. 

I fully expect them to add an emoji picker this year though, it feels like that's the reason why they changed the layout now. But… we can hotfix it if/when that happens.

cheap-glitch (commented):
Completely dropped the positioning fix for right-side comments in split diff view, since the two-line layout makes the popup smaller and it never gets clipped, even on smaller screens.

fregante:
It's being overridden by `.d-flex`

Screen Shot

cheap-glitch:
Fixed and tested on Chromium:

image

fregante (approved):

PR #5949, opened 2022-09-07
Fix `set-default-repositories-type-to-sources` on project views

### LINKED ISSUE #5940: `set-default-repositories-type-to-sources`: Failed to execute 'observe' on 'MutationObserver': parameter 1 is not of type 'Node'.
### Description

New GH Projects pages trigger this error:

```
TypeError: Failed to execute 'observe' on 'MutationObserver': parameter 1 is not of type 'Node'.
    at chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/refined-github.js:7536:13
    at new Promise (<anonymous>)
    at oneMutation (chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/refined-github.js:7532:14)
    at chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/refined-github.js:7587:51
    at async chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/refined-github.js:7585:9
    at async runFeature (chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/refined-github.js:3851:20)
    at async setupPageLoad (chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/refined-github.js:3857:34)
    at async chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/refined-github.js:3882:29
```

Just open https://github.com/users/janpio/projects/3/views/1 and observe Console

### Extension version

22.8.23

### Browser(s) used

Version 104.0.5112.81 (Official Build) (64-bit)

chcamiloam:
Closes #5940 

## Test URLs

https://github.com/users/janpio/projects/3/views/1

## Screenshot

Screen Shot 2022-09-07 at 14 50 35

## Description
The error was produces by a selector trying to find an item that doesn't exists in the page.

## Changes
The method was doing reference to the user's profile dropdown using the `:last-child` selector, looks like later a new item was added turning the user's profile item in the second last item. This PR change the selector to select the profile dropdown again.

Screen Shot 2022-09-07 at 15 58 14


fregante:
Thank you for the PR! Doesn't this break it on every other page where the current selector works?

fregante:
I think it doesn't need a per-page selector. We can fix it by using a different selector that doesn't involve nth-child.

chcamiloam:
Let me check how it works, I'll try with the observer, also I was checking other options, looks like the item we are looking for have this unique mix of clases `.position-relative.Header-item.mr-0` at it works perfectly, but let me try the observer

fregante:
Sure. Once you get it working, you can refer to how it's generally used. For example:

- it should accept a `{signal}` parameter (see how that works in other features the use `selector-observer.ts`)
- it should not use `onetime` in the definition (which it currently does)

source/features/set-default-repositories-type-to-sources.tsx
```diff
@@ -1,19 +1,20 @@
...
 }
 
-async function profileDropdown(): Promise<void> {
-	await onProfileDropdownLoad();
-	addSourceTypeToLink(select('.header-nav-current-user ~ a[href$="tab=repositories"]')!); // "Your repositories" in header dropdown
+
+async function profileDropdown(signal: AbortSignal): Promise<void> {
+	observe(`
+		.header-nav-current-user ~ a[href$="tab=repositories"]
+	`, addSourceTypeToLink, {signal});
```
fregante, line 17:
```suggestion
	observe('.header-nav-current-user ~ a[href$="tab=repositories"]', addSourceTypeToLink, {signal});
```

fregante (approved):
Fantastic!

fregante:
> We could add some pre-commits here 😅

Linting and testing is slow, post-commits are better in this case. 😜 You wouldn't want to run the whole suite every time you make a single commit

fregante:
Thank you for your contribution!

janpio:
Thank you indeed!

Issue #5897, opened 2022-08-05
`hidden-review-comments-indicator` is missing the indicator line


fregante:
### Description

![text](https://user-images.githubusercontent.com/1402241/183145510-3a59804c-891d-4b52-a8ee-1628c46fc66f.gif)

### How to replicate the issue

1. Open https://github.com/refined-github/sandbox/pull/18/files
2. Press the i key
3. See no indicator

### Extension version

22.8.1

### Browser(s) used

Safari

fregante:
It is what it is

PR #5950, opened 2022-09-08
Restore `hidden-review-comments-indicator` 

chcamiloam:
- Part of #5897 

## Test URLs
https://github.com/refined-github/sandbox/pull/18/files

## Screenshot
Screen Shot 2022-09-08 at 11 07 41

## Description

Class `.blob-num` was overwriting plugin's css, added `!important` to fix the issue.
Counter was not working, it was using the wrong selector

### DIFF
--- a/source/features/hidden-review-comments-indicator.css
+++ b/source/features/hidden-review-comments-indicator.css
@@ -1,12 +1,20 @@
+.rgh-comments-indicator {
+	position: absolute;
+}
+
 .show-inline-notes .rgh-comments-indicator {
 	display: none;
 }
 
-.rgh-comments-indicator {
-	border: none !important; /* Needs to override .blob-num’s */
-	position: unset !important; /* Needs to override .blob-num’s */
-	border-top: 1px solid !important; /* Needs to override .blob-num’s */
+.rgh-comments-indicator:is(::before, ::after) {
+	content: '' !important; /* Needs to override .blob-num’s */
+	position: absolute;
+	top: -1px;
 	left: 0;
+	width: 200%;
+	border-top: solid 1px;
+	opacity: 40%;
+	transform: translateY(-50%);
 }
 
 /* This is only used to improve the clickability */
@@ -22,7 +30,7 @@
 
 .rgh-comments-indicator button {
 	position: absolute;
-	left: -40px;
+	right: 100%;
 	transform: translateY(-50%);
 	padding: 5px 8px;
 	color: inherit;
@@ -41,5 +49,4 @@
 
 .file-diff-split .rgh-comments-indicator button {
 	padding-right: 1px;
-	left: -30px;
 }
--- a/source/features/hidden-review-comments-indicator.css
+++ b/source/features/hidden-review-comments-indicator.css
@@ -8,7 +8,7 @@
 
 .rgh-comments-indicator:is(::before, ::after) {
 	content: '' !important; /* Needs to override .blob-num’s */
-	position: absolute;
+	position: absolute !important;/* Needs to override .blob-num’s */
 	top: -1px;
 	left: 0;
 	width: 200%;
@@ -30,7 +30,7 @@
 
 .rgh-comments-indicator button {
 	position: absolute;
-	right: 100%;
+	right: 90%;
 	transform: translateY(-50%);
 	padding: 5px 8px;
 	color: inherit;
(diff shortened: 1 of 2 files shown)


source/features/hidden-review-comments-indicator.css
```diff
@@ -49,4 +49,4 @@
 
 .file-diff-split .rgh-comments-indicator button {
 	padding-right: 1px;
-}
+}
```
fregante, line 52: Strangely this was not caught by style lint. Every file must end with a line break

source/features/hidden-review-comments-indicator.css
```diff
@@ -30,7 +30,7 @@
 
 .rgh-comments-indicator button {
 	position: absolute;
-	right: 100%;
+	right: 90%;
```
fregante, line 33: Why? It seems too close now

source/features/hidden-review-comments-indicator.css
```diff
@@ -8,7 +8,7 @@
 
 .rgh-comments-indicator:is(::before, ::after) {
 	content: '' !important; /* Needs to override .blob-num’s */
-	position: absolute;
+	position: absolute !important;/* Needs to override .blob-num’s */
```
fregante, line 11:
```suggestion
	position: absolute !important; /* Needs to override .blob-num’s */
```

source/features/hidden-review-comments-indicator.css
```diff
@@ -8,7 +8,7 @@
 
 .rgh-comments-indicator:is(::before, ::after) {
 	content: '' !important; /* Needs to override .blob-num’s */
-	position: absolute;
+	position: absolute !important; /* Needs to override .blob-num’s */
```
fregante, line 11: Is this selector doing anything? I don't see the line added by `border-top` in the attached screenshot

source/features/hidden-review-comments-indicator.css
fregante: Maybe it's ok

chcamiloam: The Comments icon is cut in 1800px with `100%` but we can try :)

source/features/hidden-review-comments-indicator.css
chcamiloam: The class is working, but I'm not so sure about the selector, let me do some testing :)

chcamiloam:
Removed the css, I took one by one the changes and if doesn't affect anything (the absolute is still the fix of the main issue).
I tried with `right:100%` and this is the result
Screen Shot 2022-09-08 at 13 11 44

I changed it to 95% this time, it should allow to not have issues with 2digits numbers

chcamiloam:
Changed the structure to handle the icon with less css. Everything looks the same but the markup structure is different, add the line again.
Screen Shot 2022-09-08 at 14 46 25

chcamiloam:
Added new selector to change markup in split-view, this may seem unnecessary but I really wanted the line in bot numbers.
The space in the left is lower in split view, so it's not possible to have the number.

image

fregante:
We try to keep the amount of code to a minimum. The previous "complex CSS" worked without changes to the DOM. I'd prefer to keep it that way

chcamiloam:
I can't see how to draw the line in both number columns without changing the markup, can I get a hand here? this is how it looks right now.

image

source/features/hidden-review-comments-indicator.css
```diff
@@ -1,4 +1,4 @@
-.rgh-comments-indicator {
+:root .rgh-comments-indicator {
```
fregante, line 1: This was the only fix required to make the button visible again.

The reason why the line isn't being shown is because GitHub now uses `::before` to show the actual line, so our code fighting with GitHub’s own style.

An easier solution would be to either:

- only use ::after, which doesn't seem to be used by GitHub
- ensure that all the CSS actually applies OR figure out why it doesn't

I undid the previous changes because I don’t think they were going in the right direction for the most part. I'd be ok merging this as is for now and the line can be restored later.

fregante (approved):

source/features/hidden-review-comments-indicator.css
```diff
@@ -29,7 +22,7 @@
 
 .rgh-comments-indicator button {
 	position: absolute;
-	right: 95%;
+	left: -40px;
```
fregante: `left: -40px` only works for the specific icon size, font size and character count.
`right: 100%` is the correct alignment and it works with any amount of text in this button.

The only problem lies in the button’s content itself, which seems to have excessive padding (which exists to expand its clickable area, but it should be counteracted by negative margins)

PR #5956, opened 2022-09-09
Meta: Cleanup DOM tests

fregante:
<s>

I thought I'd give this smaller/newer dependency a try.

It works well, other than a couple of minor issues:

- https://github.com/WebReflection/linkedom/issues/156
- https://github.com/WebReflection/linkedom/issues/159
- https://github.com/WebReflection/linkedom/issues/160

</s>


fregante (commented):
Never mind. JSDOM is still being installed by third party dependencies and it's not a speed bottleneck. Linkedom on the other hand requires a few workarounds.

PR #5959, opened 2022-09-11
Fix reaction popup at bottom of comments

### LINKED ISSUE #5957: Reactions popup shown without a background
### Description

The popup on the top right works well, but at the bottom it does not. Both on issues and releases

## Native

Screen Shot 4

## Current

Screen Shot 3

cc @cheap-glitch I think this is due to:

- #5933 

1. On this issue or on https://github.com/refined-github/refined-github/releases
2. Try to react (bottom left area)

### Extension version

22.9.9

### Browser(s) used

Chrome

cheap-glitch:
Fixes #5957 

## Test URLs

 - #5957 (issue conversation)
 - this page (PR conversation)

## Screenshots

*Before*

image

*After*

image

### DIFF
--- a/source/features/center-reactions-popup.css
+++ b/source/features/center-reactions-popup.css
@@ -3,7 +3,7 @@
 	left: 50%;
 	right: auto;
 	bottom: auto; /* #5957 */
-	transform: translateX(-50%) !important;
+	translate: -50%;
 	display: grid !important;
 	grid-template: auto / repeat(4, auto); /* Restore two-line layout */
 }
--- a/source/features/center-reactions-popup.css
+++ b/source/features/center-reactions-popup.css
@@ -1,9 +1,18 @@
-.dropdown-menu-reactions.dropdown-menu-reactions { /* Double selector to override .d-flex on Chrome */
-	top: 100%;
-	left: 50%;
-	right: auto;
-	bottom: auto; /* #5957 */
-	translate: -50%;
+/* Restore two-line layout */
+:root .dropdown-menu-reactions {
 	display: grid !important;
-	grid-template: auto / repeat(4, auto); /* Restore two-line layout */
+	grid-template: auto / repeat(4, auto);
+}
+
+/* Move the popups */
+/* Align the popups to the border + add 10px to make them overflow a bit */
+/* Button on the top right of comments */
+.dropdown-menu-reactions.dropdown-menu-w {
+	/* Depends on the presence of `quick-comment-edit` */
+	translate: 107px -22px;
+}
+
+/* Button on the bottom left of comments */
+.dropdown-menu-reactions.dropdown-menu-ne {
+	translate: -27px 61px;
 }
--- a/source/features/center-reactions-popup.css
+++ b/source/features/center-reactions-popup.css
@@ -2,12 +2,8 @@
 	top: 100% !important;
 	left: 50% !important;
 	right: auto !important;
+	bottom: auto !important; /* #5957 */
 	transform: translateX(-50%) !important;
 	display: grid !important;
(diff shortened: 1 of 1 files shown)


cheap-glitch:
Just noticed a slight difference between the bottom popup of issue comments and the one of PR comments: the second one wasn't broken (at least for me on Firefox), and appears _under_ the reaction button because the feature applies correctly to it.

But for issues my fix instead restores the "native direction" and makes the popup appear _above_ the button.

We should definitely unify this but what direction should we choose? cc @fregante

source/features/center-reactions-popup.css
```diff
@@ -2,7 +2,12 @@
 	top: 100% !important;
 	left: 50% !important;
 	right: auto !important;
```
fregante, line 4: It looks like we only need `bottom: auto` here so we make the position consistent. There's no particular reason to keep the popup above or below, so we can just stick to "below"

fregante: Also none of these `important`s seem to be necessary except the transform.

Also if you just replace the transform with `translate: -50%` you preserve the popping animation too.

FloEdelmann: I think the bottom emoji selector dropdown should open to the top (like GitHub's native one). When you scroll just so that the emoji selector is the last visible thing on the page, you'd otherwise have to scroll again to click the dropdown.

cheap-glitch: > Also none of these importants seem to be necessary except the transform

True, except for `display: grid` on Firefox

> I think the bottom emoji selector dropdown should open to the top (like GitHub's native one)

+0

source/features/center-reactions-popup.css
```diff
@@ -1,8 +1,9 @@
 .dropdown-menu-reactions.dropdown-menu-reactions { /* Double selector to override .d-flex on Chrome */
-	top: 100% !important;
-	left: 50% !important;
-	right: auto !important;
-	transform: translate(-50%) !important;
+	top: 100%;
+	left: 50%;
+	right: auto;
+	bottom: auto; /* #5957 */
+	translate: -50%;
```
fregante, line 6: 💙

fregante (approved):

source/features/center-reactions-popup.css
cheap-glitch: @fregante wdyt about (@)FloEdelmann's remark?

> I think the bottom emoji selector dropdown should open to the top (like GitHub's native one). When you scroll just so that the emoji selector is the last visible thing on the page, you'd otherwise have to scroll again to click the dropdown.

source/features/center-reactions-popup.css
fregante: @FloEdelmann GitHub’s positioning does indeed make some sense, they try to keep the dropdown inside the comment area, which is necessary to avoid this on tight viewports (this PR, right now)

Screen Shot 7


I would not mind going back to the positioning and try this instead, covering the toggle altogether and keeping it aligned:

Screen Shot 8

It does require further CSS

fregante (changes requested):
As suggested by Flo, the positioning isn't quite right after all:

Screen Shot 7

source/features/center-reactions-popup.css
fregante: @cheap-glitch perhaps you can get away with using the standard `top: 50, left: 50, translate: -50, -50` technique and then adjusting it where necessary

fregante:
# Top

Screen Shot 10

# Bottom

Screen Shot 11

# Bottom, mobile

Screen Shot 12

fregante (approved):

PR #5961, opened 2022-09-11
Don't change `sticky-sidebar` state while interacting with it

fregante:
- Closes https://github.com/refined-github/refined-github/issues/5939

## Test URLs

- this PR
- the repo home

To actually test it, you'd have to try on a repo with projects, as shown in the original issue

### DIFF
--- a/source/features/sticky-sidebar.css
+++ b/source/features/sticky-sidebar.css
@@ -3,8 +3,9 @@
 }
 
 .rgh-sticky-sidebar {
-	position: sticky;
 	--rgh-header-offset: 0;
+
+	position: sticky;
 	top: var(--rgh-header-offset) !important;
 
 	/* `sticky` will trap the `fixed` dialog’s z-index in some sidebars #3732 */
--- a/source/features/sticky-sidebar.css
+++ b/source/features/sticky-sidebar.css
@@ -4,7 +4,8 @@
 
 .rgh-sticky-sidebar {
 	position: sticky;
-	top: 0;
+	--rgh-header-offset: 0;
+	top: var(--rgh-header-offset) !important;
 
 	/* `sticky` will trap the `fixed` dialog’s z-index in some sidebars #3732 */
 	/* Higher than sticky readme header #4192 */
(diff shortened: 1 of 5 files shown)


fregante:
cc @janpio

janpio:
(Is there a simple way to test from a PR? Happy to help out testing this.)

fregante:
I don't know how simple it is, but you can look on contributing.md or in the wiki for instructions on how to run it

yakov116:
@fregante did we remove the artifact? I tried looking for it to give to @janpio but was unable to find it

fregante:
Yes I did in https://github.com/refined-github/refined-github/pull/5799/files because I didn't think about this possibility 😅 it can be restored

janpio:
That was exactly what I spent a few minutes looking for. Checking out and building locally is usually too much hassle.

fregante:
You can now find it on the Actions/Test page:

https://github.com/refined-github/refined-github/actions/runs/3044209088

Or for the direct zip link: https://github.com/refined-github/refined-github/suites/8264307360/artifacts/361603861

janpio:
I can confirm that when clicking "+7 more" in the sidebar on https://github.com/janpio/refined-github-sticky-sidebar-projects-bug/issues/1 does indeed not instantly move the expanded sidebar to the top out of my viewport 👍 

It still does happen as soon as I move my mouse outside of the sidebar area though - not sure if that is expected. (If I keep my mouse on the sidebar, scrolling to top and bottom of the issue behave pretty ok to me.)

janpio:
Uh, I just noticed that the "cog" icon next to "Milestone" or "Projects" etc does not work with the version I tested ("Labels" and "Assignees" still works). No error message in console or similar. Is that a possible side effect? Or something else broken in `main`?

fregante:
It's most likely https://github.com/refined-github/refined-github/issues/5843. We already disabled the feature via [yolo](https://github.com/refined-github/yolo), but it doesn't apply to developer versions (v0.0.0)

janpio:
Thanks for that context.

PR #5964, opened 2022-09-11
Allow replacing quotes in `one-key-formatting`

### LINKED ISSUE #5960: Allow replacing quotes in `one-key-formatting`
### What I don't like

When I want to change `"` to `'` or vice-versa in a comment, I need to first delete the character and then type the other one. When selecting it and typing the other one, `one-key-formatting` wraps it (e.g. `"` → `'"'`).

My suggestion: When a single quote character (`'` or `"`) is selected and a quote character is typed (`'` or `"`), skip `one-key-formatting`. This behavior matches VS Code (which also allows one-key formatting in Markdown).

FloEdelmann:
Closes #5960.

## Test URLs

Any page with a Markdown textarea, e.g. the "new comment" field on this page.

### DIFF
--- a/source/features/one-key-formatting.tsx
+++ b/source/features/one-key-formatting.tsx
@@ -24,7 +24,7 @@ function eventHandler(event: DelegateEvent<KeyboardEvent, HTMLTextAreaElement | 
 		return;
 	}
 
-	// Allow replacing quotes
+	// Allow replacing quotes #5960
 	if (quoteCharacters.has(formattingChar) && end - start === 1 && quoteCharacters.has(field.value.at(start)!)) {
 		return;
 	}
--- a/source/features/one-key-formatting.tsx
+++ b/source/features/one-key-formatting.tsx
@@ -7,24 +7,32 @@ import {onCommentFieldKeydown, onConversationTitleFieldKeydown, onCommitTitleFie
 
 const formattingCharacters = ['`', '\'', '"', '[', '(', '{', '*', '_', '~', '“', '‘'];
 const matchingCharacters = ['`', '\'', '"', ']', ')', '}', '*', '_', '~', '”', '’'];
+const quoteCharacters = new Set(['`', '\'', '"']);
 
 function eventHandler(event: DelegateEvent<KeyboardEvent, HTMLTextAreaElement | HTMLInputElement>): void {
 	const field = event.delegateTarget;
+	const formattingChar = event.key;
 
-	if (!formattingCharacters.includes(event.key)) {
+	if (!formattingCharacters.includes(formattingChar)) {
 		return;
 	}
 
 	const [start, end] = [field.selectionStart, field.selectionEnd];
 
 	// If `start` and `end` of selection are the same, then no text is selected
-	if (start === end) {
+	if (start === end || start === null || end === null) {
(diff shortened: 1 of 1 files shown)


source/features/one-key-formatting.tsx
```diff
@@ -7,24 +7,32 @@ import {onCommentFieldKeydown, onConversationTitleFieldKeydown, onCommitTitleFie
...
 
 	// If `start` and `end` of selection are the same, then no text is selected
-	if (start === end) {
+	if (start === end || start === null || end === null) {
+		return;
+	}
+
+	const selectedText = field.value.slice(start, end);
+
+	// Allow replacing quotes
```
fregante, line 29:
```suggestion
	// Allow replacing quotes #5960
```

source/features/one-key-formatting.tsx
```diff
@@ -7,24 +7,32 @@ import {onCommentFieldKeydown, onConversationTitleFieldKeydown, onCommitTitleFie
...
-	if (!formattingCharacters.includes(event.key)) {
+	if (!formattingCharacters.includes(formattingChar)) {
 		return;
 	}
 
 	const [start, end] = [field.selectionStart, field.selectionEnd];
 
 	// If `start` and `end` of selection are the same, then no text is selected
-	if (start === end) {
+	if (start === end || start === null || end === null) {
```
fregante, line 23: Did you encounter `null` when testing or is it just a type issue? You can use `start!` to silence it.

source/features/one-key-formatting.tsx
```diff
@@ -7,24 +7,32 @@ import {onCommentFieldKeydown, onConversationTitleFieldKeydown, onCommitTitleFie
...
+		return;
+	}
+
+	const selectedText = field.value.slice(start, end);
+
+	// Allow replacing quotes
+	if (quoteCharacters.has(formattingChar) && quoteCharacters.has(selectedText)) {
```
fregante, line 30: Since `.slice` can be a heavy operation on long fields, you can avoid running it altogether in most cases by using this:

```suggestion
	if (quoteCharacters.has(formattingChar) && quoteCharacters.has(textFieldEdit.getSelection(field))) {
```

But in reality it's even cheaper to use a single character check via `end - start === 1 && field.value.at(start)` because this should only work on a single character anyway.

source/features/one-key-formatting.tsx
FloEdelmann: It was only typing. But always silencing it seemed wrong, I mean the type includes `null` for a reason I guess?

source/features/one-key-formatting.tsx
FloEdelmann: Not too readable, but if it's better for performance, let's use that.

source/features/one-key-formatting.tsx
fregante: DOM types consider more than the usual. Even .textContent is string | null, but that only applies to image tags, so silencing them is correct in many cases. 

Also in extensions things change under your feet so we prefer throwing errors when things don't match our expectations instead of just silencing everything like you would in a self-contained application.

fregante (approved):
謝謝

Issue #4902, opened 2021-10-10
Releases UI Refresh


kidonng:
### Please ensure:

- [X] The bug is caused by Refined GitHub. It doesn't happen if I disable the extension.

### Include in this issue:

- [ ] Screenshots/video/gif demonstrating the bug, if it’s visual
- [ ] Console errors, if any

### Description

[GitHub Blog: A new public beta of GitHub Releases: How we’re improving the release experience](https://github.blog/2021-10-04-beta-github-releases-improving-release-experience/)

I just noticed there is a new Feature preview, this issue is for tracking related bugs.

Known broken issues:

- `tags-dropdown`

   ```
   ❌ tags-dropdown 0.0.0 → TypeError: Cannot read properties of undefined (reading 'append')
    at init (VM158 refined-github.js:14439)
    at runFeature (VM158 refined-github.js:8004)
    at setupPageLoad (VM158 refined-github.js:8022)
    at VM158 refined-github.js:8075
   ```

- `link-to-changelog-file`

  ```
  ❌ link-to-changelog-file 0.0.0 → TypeError: Cannot read properties of undefined (reading 'after')
    at init (VM158 refined-github.js:8485)
    at async runFeature (VM158 refined-github.js:8004)
    at async setupPageLoad (VM158 refined-github.js:8022)
  ```

### Example URL

https://github.com/fish-shell/fish-shell/releases

### Browser(s) used

Chrome

yakov116:
![image](https://user-images.githubusercontent.com/16872793/136704411-4e2b8022-7754-43d4-8bc5-97f9607c3189.png)

PR #5966, opened 2022-09-11
Meta: Drop code specific to "Releases UI refresh"

FloEdelmann:
The "Releases UI Refresh" (#4902) was nearly one year ago, so let's drop code still targets the old layout.

## Test URLs

* https://github.com/facebook/react/releases
* https://github.com/facebook/react/tags
* https://github.com/facebook/react/releases/tag/v17.0.2


fregante (approved):

fregante:
Дякую тобі

PR #5972, opened 2022-09-13
Meta: Restore artifacts on CI

yakov116:
Reverts refined-github/refined-github#5799

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -64,6 +64,11 @@ jobs:
           cache: npm
       - run: npm ci
       - run: npm run build:webpack -- --json webpack-stats.json
+      - uses: actions/upload-artifact@v2
+        if: matrix.os == 'ubuntu-latest'
+        with:
+          name: refined-github
+          path: distribution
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -72,11 +77,14 @@ jobs:
           key: ${{ secrets.RELATIVE_CI_KEY }}
 
   Safari:
-    runs-on: macos-12
```
yakov116, line 75: Need to fix this

.github/workflows/test.yml
```diff
@@ -64,6 +64,11 @@ jobs:
           cache: npm
       - run: npm ci
       - run: npm run build:webpack -- --json webpack-stats.json
+      - uses: actions/upload-artifact@v2
+        if: matrix.os == 'ubuntu-latest'
+        with:
+          name: refined-github
+          path: distribution
```
fregante, line 71: Only this, not the Safari part 👍

fregante (approved):

PR #5976, opened 2022-09-13
Fix: `select-notifications` causes button overlap on mobile

### LINKED ISSUE #5796: `select-notifications` z-index bug on mobile
### Please ensure:

### Example URLs

Notifications page 

### Extension version

22.7.1

### Browser(s) used

Safari iOS 

### Description

We push these buttons above the overlay so that they’re clickable when the dropdown is open. This is no longer feasible when the dropdown turns into a modal. 

We should move the z-index change to a media query that matches this UI change. 

B563F813-F391-4E1A-B061-CCF9BD33DBCE

### Include in this issue:

- [ ] Console errors, if any

fregante:
- Fix #5796

## Test URLs

## Before

Screen Shot

## After

Screen Shot 1

### DIFF
--- a/source/features/select-notifications.css
+++ b/source/features/select-notifications.css
@@ -1,5 +1,9 @@
-.js-notifications-mark-selected-actions > *,
-.rgh-open-selected-button {
-	/* Allow clicking on "Done" and other toolbar buttons even when the dropdown (z-index of 99) is open #4753 */
-	z-index: 100;
+/* Only when the meny is shown as a dropdown, not as a modal (mobile) #5796 */
+/* The media query matches the native one exactly, inverted via `not` */
+@media screen and not (max-width: 543px) {
+	.js-notifications-mark-selected-actions > *,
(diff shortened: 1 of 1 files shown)


fregante:
I forgot to test whether this still works where it should though 🥲

source/features/select-notifications.css
```diff
@@ -1,5 +1,9 @@
-.js-notifications-mark-selected-actions > *,
-.rgh-open-selected-button {
-	/* Allow clicking on "Done" and other toolbar buttons even when the dropdown (z-index of 99) is open #4753 */
-	z-index: 100;
+/* Only when the meny is shown as a dropdown, not as a modal (mobile) #5796 */
```
yakov116, line 1: Meny?

fregante: https://meny.no 😂

fregante: When you think _it's just a one-line change_ but it has 5 bugs.

PR #5986, opened 2022-09-15
Avoid triggering `pr-branch-auto-delete` when simply watching a PR

fregante:
- Closes https://github.com/refined-github/refined-github/issues/5868

## Test

Any PR:

- that you can merge
- on a repo without native branch auto-deletion

### Test bugfix

1. Open the PR in two browser, one with RG, one without
2. Merge the PR in the browser without RG
3. The branch should not be deleted 

## Test feature

1. Merge the PR
3. The branch should be deleted

### DIFF
--- a/source/features/first-published-tag-for-merged-pr.tsx
+++ b/source/features/first-published-tag-for-merged-pr.tsx
@@ -6,7 +6,6 @@ import * as pageDetect from 'github-url-detection';
 
 import features from '.';
 import fetchDom from '../helpers/fetch-dom';
-import onPrMerge from '../github-events/on-pr-merge';
 import createBanner from '../github-helpers/banner';
 import TimelineItem from '../github-helpers/timeline-item';
 import attachElement from '../helpers/attach-element';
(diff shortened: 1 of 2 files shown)


source/features/first-published-tag-for-merged-pr.tsx
```diff
@@ -118,14 +122,8 @@ void features.add(import.meta.url, {
...
 		canCreateRelease,
 	],
-	additionalListeners: [
-		onPrMerge,
-	],
-	onlyAdditionalListeners: true,
 	deduplicate: false,
-	init() {
-		void addReleaseBanner();
-	},
```
fregante, line 128: I thought this feature was affected because it relied on the old behavior, but not really.

However it would greatly benefit from being rewritten with the new observer instead of using additional listeners, also because we have two separate UI pieces, one of which needs to be applied multiple times.

Related:

- https://github.com/refined-github/refined-github/issues/5895
- https://github.com/refined-github/refined-github/issues/5222#issuecomment-1242931510

fregante: Also this would help rewrite `pr-branch-auto-delete`:

- https://github.com/fregante/delegate-it/issues/34

PR #5996, opened 2022-09-18
Fix duplication of `profile-gists-link` and `show-user-top-repositories`

### LINKED ISSUE #5995: `show-user-top-repositories` is duplicate
### Description

https://user-images.githubusercontent.com/20497787/190883835-61bcc54c-6d04-4726-bbf7-c23292f9b47d.mp4

1. Go to a user page. e.g, https://github.com/minaminao
2. Click a link in the profile. 
3. Back to the user page.

### Extension version

22.9.15

### Browser(s) used

Chrome

Semiroundpizza8:
Hoping it's okay I open this - I was poking around the code and saw this issue was open and figured I'd take a crack at it

Closes #5995

## Test URLs

I'm assuming this affects any users profile page when using the extension. Tested against this URL along with some other private profiles:
- https://github.com/minaminao

## Screenshot

Untitled_ Sep 18, 2022 4_46 AM

### DIFF
--- a/source/features/profile-gists-link.tsx
+++ b/source/features/profile-gists-link.tsx
@@ -9,6 +9,7 @@ import features from '.';
 import * as api from '../github-helpers/api';
 import {getCleanPathname} from '../github-helpers';
 import createDropdownItem from '../github-helpers/create-dropdown-item';
+import attachElement from '../helpers/attach-element';
 
 const getGistCount = cache.function(async (username: string): Promise<number> => {
 	const {user} = await api.v4(`
@@ -50,7 +51,12 @@ async function init(): Promise<void> {
 	}
 
 	// Add gist counter to profile
-	navigationBar.append(link);
+	attachElement({
+		anchor: navigationBar,
+		append: () => link,
+	});
+
+	// This triggers a refresh logic for GitHub’s overlow menu
 	navigationBar.replaceWith(navigationBar);
 
 	select('.js-responsive-underlinenav .dropdown-menu ul')!.append(
@@ -58,8 +64,12 @@ async function init(): Promise<void> {
 	);
 
 	const count = await getGistCount(username);
+
 	if (count > 0) {
-		link.append(<span className="Counter">{count}</span>);
+		attachElement({
+			anchor: link,
+			append: () => <span className="Counter">{count}</span>,
+		});
 	}
 }
 
--- a/source/features/profile-gists-link.tsx
+++ b/source/features/profile-gists-link.tsx
@@ -44,8 +44,10 @@ async function init(): Promise<void> {
 	const navigationBar = (await elementReady('.UnderlineNav-body'))!;
 
 	// If the button already exists, skip adding it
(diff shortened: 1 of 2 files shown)


source/features/profile-gists-link.tsx
```diff
@@ -42,6 +42,12 @@ async function init(): Promise<void> {
 	);
 
 	const navigationBar = (await elementReady('.UnderlineNav-body'))!;
+
+	// If the button already exists, skip adding it
+	const isGistButtonAlreadyAdded = !!select(".UnderlineNav-body [data-tab-item='rgh-gists-item']");
+	if(isGistButtonAlreadyAdded) return;
```
Semiroundpizza8, line 48: Both these fixes get the job done but they're just band-aid early returns right now - I have a feeling the real issue is that the event handlers that these init functions are adding aren't being torn down when navigating away causing them to build up over time

source/features/show-user-top-repositories.tsx
```diff
@@ -12,8 +12,12 @@ function init(): void {
 		sort: 'stargazers',
 	}).toString();
 
-	// Showcase title
-	select('.js-pinned-items-reorder-container .text-normal')!.firstChild!.after(
+	// If already added, skip
+	const pinnedText = select('.js-pinned-items-reorder-container .text-normal')!;
+	if(pinnedText.innerHTML.includes("Top repositories")) return;
```
Semiroundpizza8, line 17: This was the cleanest solution I could come up with, inner text wasn't working out since the elements added after the <p> tag but I can tweak it if theres a more comfortable workaround here. I figured it'd either be parsing the innerHTML or adding a data attribute and making another select to the page which would be cleaner but very very very slightly less performant lol

Semiroundpizza8 (commented):
Did a quick review before undrafting

fregante:
Thank you! However we already have a helper that automatically avoids duplicates: `attachElement`

source/features/profile-gists-link.tsx
```diff
@@ -42,6 +42,14 @@ async function init(): Promise<void> {
...
 	const navigationBar = (await elementReady('.UnderlineNav-body'))!;
+
+	// If the button already exists, skip adding it
+	const isGistButtonAlreadyAdded = Boolean(select('.UnderlineNav-body [data-tab-item=\'rgh-gists-item\']'));
+	if (isGistButtonAlreadyAdded) {
+		return;
+	}
+
+	// Add gist counter to profile
 	navigationBar.append(link);
```
fregante, line 53: This would become something like:

```js
attachElement({
  anchor: navigationBar,
  append: () => <a/> // etc, all the logic to create the link should be here
})
navigationBar.replaceWith(navigationBar) // This triggers a refresh logic for GitHub’s overlow menu
```


The same goes for the `.append` calls below

Semiroundpizza8:
@fregante Sounds good, thanks for the quick feedback! I added those in and it worked like magic - we should be good to go here!

fregante:
Almost there! There's another .append in the file that you missed. 

Also the entire link generation should happen inside the `append` function, it should not reference it like `() => link`

Refer to other usages of attachElement in the extension for examples

PR #5997, opened 2022-09-19
Restore `clean-conversation-sidebar`

fregante:
- Fixes https://github.com/refined-github/refined-github/issues/5843

Not the most optimal solution, but the fastest one. 

Note: No Firefox support, sorry.

### Future improvements

- [ ] Rewrite using `:has()` selectors that encapsulate more logic (the only reason why this can't be a CSS-only feature is because most hideable parts are text nodes 😰)
- [ ] Reduce/fix spacing. The sidebar’s height slightly increases when the milestones dropdown is opened

## Testing

The actions change depending on:

- issue / PR
- editable / non editable

Tested URLs:

- This PR
- https://github.com/refined-github/refined-github/issues/5843
- https://github.com/facebook/react/issues/13991

## Screenshot

Screen Shot 1

Screen Shot

### DIFF
--- a/source/features/clean-conversation-sidebar.tsx
+++ b/source/features/clean-conversation-sidebar.tsx
@@ -5,7 +5,6 @@ import onetime from 'onetime';
 import * as pageDetect from 'github-url-detection';
 
 import features from '.';
-import selectHas from '../helpers/select-has';
 import onElementRemoval from '../helpers/on-element-removal';
 import onDiscussionSidebarUpdate from '../github-events/on-discussion-sidebar-update';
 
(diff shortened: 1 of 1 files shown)


source/features/clean-conversation-sidebar.tsx
```diff
@@ -46,7 +45,10 @@ function cleanSection(selector: string): boolean {
 		'[aria-label="Select projects"] .Link--primary',
 	];
 
-	const heading = select('.discussion-sidebar-heading', container)!;
+	const heading = select([
+		'details:has(> .discussion-sidebar-heading)', // Can edit sidebar, has a dropdown
+		'.discussion-sidebar-heading', // Cannot editor sidebar, has a plain heading
+	], container)!;
```
fregante, line 51: I couldn't use `selectHas` because it doesn't support selector lists and it's not worth working on that anymore. Firefox support will follow eventually.

PR #6006, opened 2022-09-22
Warp speed 🚀

fregante:
<details>
<summary>Previous details. Already addressed</summary>

I don't know how much I'll change in this PR, but I'm opening it to keep track of what needs to be reviewed before merging or after:

- if `attachElement` are used, then it should:
	- not use `deduplicate`
- if `observe` is used, then it should:
	- use `awaitDomReady: false`
	- not use `el.classList.add('rgh-some-class')` unless it has other uses
	- not use `additionalListeners`
	- be used on elements that are eventually visible, not on meta tags for example
- if `init` has `signal`, then it should:
	- never use `onetime(init)`
- if `deduplicate` was dropped, then it should:
	- use `attachElement`, OR
	- use another deduplication mechanism, like a classic `:not(.rgh-some-class)`

And more generically

- if a feature uses DOM-based detections, it should not use `awaitDomReady: false`
- if the feature lacks `include`, `asLongAs`, and `exclude`, it should probably not use `signal`

All the rules in ripgrep form, which returns a list of "invalid" files: (WIP)

```
rg -l 'attachElement' (rg -l 'deduplicate')
rg -l '\tobserve' (rg --files-without-match 'awaitDomReady')
rg -l '\tobserve' (rg -l 'classList.add')
rg -l '\tobserve' (rg -l 'additionalListeners')
rg -l '\(signal:' (rg -l 'init: onetime') 
```

Commands:
- append `| xargs code` to open the files in vscode instead of just listing them`
- Prepend `$` to the parens if you're not using Fishell or PowerShell

</details>

Features already broken:

- [ ] `jump-to-change-requested-comment`


fregante (commented):
Me, as I progressively delete code and hacks thanks to the new observer

fregante (commented):
Needs testing. 🥲 

Primarily we need to understand whether the selector observer alone can be used as a deduplicator across page loads

fregante (commented):
Time to merge! All features were more or less tested over the past days.

PR #6019, opened 2022-09-25
Add notice to `clear-pr-merge-commit-message`

fregante:
- Replaces and closes https://github.com/refined-github/refined-github/pull/5967
- Part of https://github.com/refined-github/refined-github/issues/5155

This is the minimum we can do to make the feature louder since it overrides the user’s settings.

## Test URLs

This PR

## Screenshot

Screen Shot 4


fregante:
cc @yakov116 @JelteF @karfau

yakov116:
Good (better now then never 😀)

PR #6023, opened 2022-09-26
Make `no-unnecessary-split-diff-view` lighter and immediate 

fregante:
Works like magic 🤯

But `features.addCssFeature` isn't working? 🤔 

### PR files

https://github.com/refined-github/sandbox/pull/50/files?diff=split

Screen Shot 2

### Compare page

https://github.com/refined-github/sandbox/compare/no-unnecessary-split-diff-view?expand=1&diff=split

Screen Shot 3

### Single commit

https://github.com/refined-github/sandbox/commit/c28cc8e5271452b5b4c347d46a63f717c29417d6?diff=split

Screen Shot 4

### DIFF
--- a/source/features/no-unnecessary-split-diff-view.css
+++ b/source/features/no-unnecessary-split-diff-view.css
@@ -1,27 +1,30 @@
+/* The selector looks for diff tables WITHOUT changes on the left XOR on the right */
+/* Instead of duplicating this selector for each rule, we set a variable and pick it up where needed */
 .rgh-no-unnecessary-split-diff-view .js-diff-table:not(
 :has([data-split-side='left']:is(.blob-code-addition, .blob-code-deletion))
-) :is([data-hunk], .blob-expanded) td:nth-child(2),
-.rgh-no-unnecessary-split-diff-view .js-diff-table:not(
-:has([data-split-side='right']:is(.blob-code-addition, .blob-code-deletion))
-) :is([data-hunk], .blob-expanded) td:nth-child(4) {
-	display: none !important;
-	width: 0 !important;
-}
-
-.rgh-no-unnecessary-split-diff-view .js-diff-table:not(
-:has([data-split-side='left']:is(.blob-code-addition, .blob-code-deletion))
-),
-.rgh-no-unnecessary-split-diff-view .js-diff-table:not(
-:has([data-split-side='right']:is(.blob-code-addition, .blob-code-deletion))
 ) {
+	--rgh-only-additions: none;
 	table-layout: auto !important;
 }
 
 .rgh-no-unnecessary-split-diff-view .js-diff-table:not(
-:has([data-split-side='left']:is(.blob-code-addition, .blob-code-deletion))
-) :is(.inline-comments, .js-inline-annotations) .empty-cell:not(.blob-num),
-.rgh-no-unnecessary-split-diff-view .js-diff-table:not(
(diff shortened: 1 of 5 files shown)


source/features/no-unnecessary-split-diff-view.css
```diff
@@ -1,13 +1,29 @@
...
+	table-layout: auto !important;
+}
+
+.rgh-no-unnecessary-split-diff-view
+	.js-diff-table:not(:has([data-split-side="left"]:is(.blob-code-addition, .blob-code-deletion)))
+	:is(.inline-comments, .js-inline-annotations)
+	.empty-cell:not(.blob-num),
+.rgh-no-unnecessary-split-diff-view
+	.js-diff-table:not(:has([data-split-side="right"]:is(.blob-code-addition, .blob-code-deletion)))
+	:is(.inline-comments, .js-inline-annotations)
```
fregante, line 26: No nesting, no custom selectors available. It would be much more readable 🥲 

There's likely room for improvement though, I just copy-pasted the previous selectors and tweaked it until it worked.

FloEdelmann:
@fregante in https://github.com/refined-github/refined-github/issues/5797#issuecomment-1238431504:

> We're not going to retroactively break features

@fregante in #6023: But it's faster :slightly_frowning_face:

fregante:
The problem is that many of our features greatly benefit from the selector. This week I found so many roadblocks and had to leave a bunch of related TODO comments; I could have just used the new selector and trimmed all of the hacky code instead. 

Here I adjusted the feature and then tried using the new selector. So much easier and instant.

FloEdelmann:
I don't really mind, I have enabled `about:config` rule. But maybe we should advertise it a bit?

fregante:
I'll go ahead and merge this as is because:

> this isn't a substantial QoL feature like, say, "update base branch in PR", so breaking it for a few months probably won't be felt by many.

and I'll add:

If you're reading this, you can set `layout.css.has-selector.enabled` to re-enable the feature on Firefox

yakov116:
> If you're reading this, you can set `layout.css.has-selector.enabled` to re-enable the feature on Firefox

This needs to be in the release notes

yakov116:
> set up some sort of automated live testing of CSS selectors

Should not be too difficult, cant make a selector check on load of the feature?

fregante:
> > set up some sort of automated live testing of CSS selectors
> 
> Should not be too difficult, cant make a selector check on load of the feature?

1. Given the complex logic, we need to test the whole selector for the tests to be useful
2. Testing the selectors "live" means we have to import them into the JS side and run them at some point
3. Given the complex logic, these selectors will not appear on every page, so we can't actually _assert_ that a selector _must_ match an element on any given page the user loads.


The proper testing harness I'm thinking of would look like:

1. In node
1. Loads `no-unnecessary-split-diff-view.css`
2. Extracts the selectors and any related URLs, to be defined like:
	```css
	/** @expect-on http://github.com/some/dead-repo*/
	.repo-bar .tab:has(.counter[label=0]) {
		display:none;
	}
	```
4. Fetches the URL and looks for the selector on the page

FloEdelmann:
@fregante Reminder about https://github.com/refined-github/refined-github/pull/6023#issuecomment-1270419090:

> This needs to be in the release notes

fregante:
Done https://github.com/refined-github/refined-github/releases/tag/22.10.10

Issue #5992, opened 2022-09-16
Drop most custom event listeners


fregante:
These handlers probably have to go, at least some of them, and be replaced by a simple `selector-observer`:

- https://github.com/refined-github/refined-github/tree/main/source/github-events

Reasons:

- deduplication logic is making things difficult
- a simple selector observer is usually much easier to reason about

Related:

- https://github.com/refined-github/refined-github/issues/5895
- https://github.com/refined-github/refined-github/issues/5871

fregante:
We're good now! There are only a couple left, both of which are deprecated and only used by 2 features. 🎉 

Then there's this one for which we don't have custom listers, we just have raw `delegate` listeners everywhere:

- https://github.com/refined-github/refined-github/issues/6560

Issue #5871, opened 2022-07-30
Every feature using `delegate` and `deduplicate` is broken on history navigation


fregante:
### Description

Here's what happens:

1. We create the DOM elements and attach the delegated listeners
2. The user navigates away
3. We remove the listeners
4. The user goes back
5. `init` is not run again because it's "deduplicated"

At this point, the elements still exist, but they have no listeners.

### How to replicate the issue

1. From this issue
2. Click the "Issues" tab
3. Go back
4. Click the `collapsible-content-button`

I'm fixing this feature specifically, but likely all of them are.

Related:

- #5869 

### Extension version

main

### Browser(s) used

Chrome

fregante:
https://github.com/refined-github/refined-github/pull/5873 took care of delegate-only features. Features with "change dom + delegate" in their `init` are still broken, they will need changes like:

- https://github.com/refined-github/refined-github/pull/5872
- custom deduplication detection

This is now a sub-issue of #[]()5895, but just for features that use `delegate`:

- https://github.com/refined-github/refined-github/issues/5895

fregante:
Starter:

```
rg --files-without-match "deduplicate: false" (rg -l "delegate\(" source/features/*.tsx) | sort -R | xargs code
```

Finds every delegate feature  without `deduplicate:false` and opens it in vscode in random order.

I'll start on some

fregante:
Not many features left to close this issue ☺️ 

```sh
❯ rg -l deduplicate $(rg -l delegate source/features/*.tsx) | sort -R
source/features/copy-file.tsx
source/features/one-click-diff-options.tsx
source/features/one-click-review-submission.tsx
source/features/convert-release-to-draft.tsx
source/features/open-all-conversations.tsx
source/features/quick-mention.tsx
source/features/rgh-sponsor-button.tsx
source/features/rgh-welcome-issue.tsx
source/features/hide-low-quality-comments.tsx
source/features/deep-reblame.tsx
source/features/esc-to-cancel.tsx
source/features/comment-fields-keyboard-shortcuts.tsx
source/features/new-repo-disable-projects-and-wikis.tsx
source/features/pr-filters.tsx
```

PR #6024, opened 2022-09-26
Bugfixes and `deduplicate` cleanup

fregante:
Related:

- #5871 
- #5992


source/features/create-release-shortcut.tsx
```diff
@@ -17,14 +17,11 @@ void features.add(import.meta.url, {
 	include: [
 		pageDetect.isReleasesOrTags,
 	],
-	deduplicate: 'has-rgh',
 	init,
 }, {
 	include: [
-		pageDetect.isReleasesOrTags, // If the release couldn't be published, GitHub changes the url to /releases while still being on the "New release" page
```
fregante, line 24: This causes an error on every `isReleasesOrTags` page. I'd rather have a feature that uses `replaceState` to `location.href` instead

source/features/bypass-checks.tsx
```diff
@@ -46,10 +47,6 @@ void features.add(import.meta.url, {
 	include: [
 		pageDetect.isRepo,
 	],
-	exclude: [
-		pageDetect.isEmptyRepo,
```
fregante, line 50: It doesn't work with `awaitDomReady: false` and it's not needed with the observer

source/features/prevent-pr-merge-panel-opening.tsx
```diff
@@ -17,17 +17,9 @@ function init(signal: AbortSignal): void {
...
-	asLongAs: [
-		// The user is a maintainer, so they can probably merge the PR
-		() => select.exists('.discussion-sidebar-item .octicon-lock'),
-	],
 	include: [
 		pageDetect.isPRConversation,
 	],
-	exclude: [
-		() => select.exists('#partial-discussion-header [title="Status: Draft"]'),
-	],
```
fregante, line 29: It doesn't work with `awaitDomReady: false` and it's not needed with the observer

PR #6035, opened 2022-10-02
Fix back-forth navigation delay caused by some features

fregante:
- Probably fixes https://github.com/refined-github/refined-github/issues/6028

Here's what I think happens:

- Turbo caches the style elements by `outerHTML`
- We alter the content of the `style` element
- Turbo finds "a new style element" (same element, but different `outerHTML`)
- No load/error events are fired because the element is the same

I assume this issue exists in every browser so it makes sense to use a fix that works in every browser. I haven't experienced it because I'm on the LTS version (Safari 🥲)

### DIFF
--- a/source/helpers/selector-observer.tsx
+++ b/source/helpers/selector-observer.tsx
@@ -1,4 +1,3 @@
-import mem from 'mem';
 import React from 'dom-chef';
 import {css} from 'code-tag';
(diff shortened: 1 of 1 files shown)


fregante:
@kaste @probablykasper

kaste:
Works for me.

PR #6037, opened 2022-10-03
Avoid duplicate `user-local-time`

### LINKED ISSUE #6013: Duplicate on `user-local-time` (official feature now)
### Description

Local timezones were just added to GitHub; optional feature that users can set, only visible on profiles who actually set up a local timezone.

See https://github.blog/changelog/2022-09-23-local-timezones-available-on-profiles/

image

https://github.com/Katsute

### Extension version

22.9.21

### Browser(s) used

Google Chrome 105.0.5195.127

fregante:
- Closes #6013 

## Test URLs

- Right here, hover this: @Katsute 

## Screenshot

Screen Shot

### DIFF
--- a/source/features/user-local-time.tsx
+++ b/source/features/user-local-time.tsx
@@ -6,14 +6,11 @@ import cache from 'webext-storage-cache';
 import delay from 'delay';
 import select from 'select-dom';
 import {ClockIcon} from '@primer/octicons-react';
-import * as pageDetect from 'github-url-detection';
 
 import features from '../feature-manager';
 import observe from '../helpers/selector-observer';
 import * as api from '../github-helpers/api';
-import {getUsername, getCleanPathname} from '../github-helpers';
-import attachElement from '../helpers/attach-element';
-import elementReady from 'element-ready';
+import {getUsername} from '../github-helpers';
 
 type Commit = {
 	url: string;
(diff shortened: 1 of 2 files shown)


source/features/user-local-time.tsx
```diff
@@ -162,49 +165,7 @@ function init(signal: AbortSignal): void {
...
-function createTimeElement(datePromise: Promise<string | false>): JSX.Element {
-	const placeholder = <span className="v-align-middle">Guessing local time…</span>;
-	const container = (
-		<li className="vcard-detail pt-1 css-truncate css-truncate-target">
-			<ClockIcon/> {placeholder}
-		</li>
-	);
-
-	void display({datePromise, placeholder, container});
-	return container;
```
fregante, line 194: @kidonng I reverted https://github.com/refined-github/refined-github/pull/5779 because the _temporary_ duplicate code ended up being permanent and the feature was unmaintainable as is.

fregante (commented):
The feature has a lot of awkward logic due to the hovercard behavior. It can probably be avoided with some tactical CSS that adds the required padding automatically to all user hovercards so that the initial calculation by GitHub takes it into account. Example:

```css
.Popover-message:has([aria-label="user avatar"]):not(:has(.octicon)) {
  padding-bottom: 26px;
}
```

Plus now better selectors are available thanks to the `aria-label` attributes. The feature would look like:

```js
observe('user Details Container without time', container => {
  if (cached time is `false`) {
    return
  }
  const user = container.user ?? globalUser
  container.append(<DynamicUserTime {user}>)
})

Issue #5820, opened 2022-07-12
`one-click-diff-options` removes the built-in diff options


nicolasff:
### Description

I noticed today that the built-in GitHub diff options on the "Files" tab of a pull request were no longer visible when `one-click-diff-options` was enabled.

I opened a PR that looked like this (some info redacted):
![image](https://user-images.githubusercontent.com/7276/178568945-a58cfbd2-d788-4a56-8b98-5a5ca3b4698d.png)

Notice the "book" icon and the `[•]` icon, these are from the `one-click-diff-options` feature. But usually there are also a few more options from GitHub itself, such as "Changes from all commits…" (or a specific commit), "File filter…", and "Jump to…". These are now missing.

I used Refined GitHub's "Start process to identify a feature" to pinpoint which feature was causing this issue. I went through all the steps and a pop-up said it was due to `one-click-diff-options`. When I disabled this feature, the built-in diff options came back.

With the feature disabled, the same page now looks like this, with the expected diff options:
![image](https://user-images.githubusercontent.com/7276/178571689-833530c8-6720-4908-91d8-51dde0194668.png)

Regardless of whether the feature is enabled, I consistently see an error being logged (twice) in the console on the Files page where I noticed this issue. Both log messages are the same:

```
refined-github.js:5963 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'textContent')
    at isArchivedRepo (refined-github.js:5963:75)
    at refined-github.js:3941:95
    at Array.every (<anonymous>)
    at refined-github.js:3941:82
    at setupPageLoad (refined-github.js:3942:8)
    at refined-github.js:3987:35
```

This is preceded by two errors reported by _different_ features (`clean-repo-tabs` and `more-dropdown-links`), so I'm not sure it's relevant here but I'm including it just in case.

I tried the "Clear cache" suggestion, but it didn't help. This happened with GitHub Enterprise Server.

### How to replicate the issue

I am unable to provide a link, but I tried to reproduce the issue on GitHub.com and was unable to.

### Extension version

22.7.11

### Browser(s) used

Google Chrome version 104.0.5112.39 (Official Build) beta (x86_64)

nicolasff:
@busches sorry, I should have mentioned it. It's 3.1.19.

fregante:
https://github.com/refined-github/refined-github/blob/749b028ca9591155cdcfce5226fbb615796cc625/source/features/one-click-diff-options.tsx#L100-L101

[...]

PR #6040, opened 2022-10-03
Disable broken `one-click-diff-options` on GitHub Enterprise

fregante:
- For #5820

The feature actively breaks GitHub’s UI on Enterprise, so I had to disable it there. An actual bugfix would be ideal.

@ELLIOTTCABLE @nicolasff @busches

### DIFF
--- a/source/features/one-click-diff-options.tsx
+++ b/source/features/one-click-diff-options.tsx
@@ -163,6 +163,7 @@ void features.add(import.meta.url, {
 	],
 	exclude: [
 		pageDetect.isPRFile404,
+		pageDetect.isEnterprise, // #5820
 	],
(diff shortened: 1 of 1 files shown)


busches:
What version is it broken on? It was working on Friday for the latest GHE.

fregante:
See https://github.com/refined-github/refined-github/issues/5820#issuecomment-1201897111

It felt like too much of a breaking to leave it unfixed. It can be restored on GHE when it's past our support window, or when the selector is updated by someone on GHE.

busches:
I'm on 3.6 series and it's working fine there. 3.3 series is likely supported through December, unsure of when they changed the markdown to the new layout.

PR #6041, opened 2022-10-03
Restore file titles in global search

fregante:
I don't remember if I already fixed this. Found an old commit. WIP


fregante:
Yep. I did

PR #6046, opened 2022-10-04
Update App Store icon

fregante:
@sindresorhus does a palette inversion quench your refresh thirst? 😃 

Before | After
--- | ---
 | 

Making the icon completely flat white could also work, similar style to GitHub Desktop’s:


sindresorhus:
Not a big fan. The gradient clashes with the artwork.

fregante:
(inner shadow would be better probably)

Before | After
--- | ---
 | AppIcon

fregante:
Inner shadow

Before | After
--- | ---
 | AppIcon2

sindresorhus:
I'm fine with just keeping the existing icon.

PR #6051, opened 2022-10-08
Fix `clean-dashboard` style for multiple commits

bastimeyer:
## Test URLs

Dashboard: https://github.com/
Hovercards are broken when more than one commit gets listed in a single hovercard.

The reason for this is that some hovercards now appear to have the `classes` attribute instead of the `class` attribute, which breaks the existing CSS selector in the `clean-dashboard`:

image

## Screenshot

Before:
image

After:
image


fregante:
That's likely a temporary GitHub bug, so we can't merge this. Also we don't generally add code that is only useful when a feature is disabled, in this case `hide-newsfeed-events-noise`

fregante:
So it turns out this affected 2 of our features, so `hide-newsfeed-noise` was not disabled as I thought.


I hotfixed 22.9.9 and 20.10.3

## Before


Screen Shot 4

## After

Screen Shot 5

PR #6068, opened 2022-10-11
Also `hide-low-quality-comments` containing "thanks"

FloEdelmann:
Otherwise low quality comments that contain "thanks" or "thank you for your work" are still low quality.
Comments that only contain "thanks" or similar are still shown.

## Test URLs
https://github.com/Etar-Group/Etar-Calendar/issues/62#issuecomment-513282994

### DIFF
--- a/source/helpers/is-low-quality-comment.test.ts
+++ b/source/helpers/is-low-quality-comment.test.ts
@@ -27,7 +27,9 @@ describe('isLowQualityComment', () => {
 		'+1 and thank you for your work',
 		'Same here, please update, thanks',
 		'Same here! Please update, thank you.',
-	])('%s', text => assert.isTrue(isLowQualityComment(text)));
+	])('%s', text => {
+		assert.isTrue(isLowQualityComment(text));
+	});
 
 	test.each([
 		'+1\n<some useful information>',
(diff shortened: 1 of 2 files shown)


source/helpers/is-low-quality-comment.test.ts
```diff
@@ -1,28 +1,43 @@
...
-	assert.isFalse(isLowQualityComment('+1\n<some useful information>'));
-	assert.isFalse(isLowQualityComment('Same here. <some useful information>'));
-	assert.isFalse(isLowQualityComment('Same here, please update, thanks'));
-	assert.isFalse(isLowQualityComment('Same here! Please update, thank you.'));
+	test.each([
+		'+1\n<some useful information>',
+		'Same here. <some useful information>',
+		'Same here on v1.2',
+		'Thanks!',
+		'Thank you 👍',
```
fregante, line 39: Why would these not be considered low quality?

fregante: Isn't the intent of this PR to hide "Thanks"?

FloEdelmann: Only saying "thanks" while pushing a bug report or feature request sounds like "thanks if this gets implemented", which is rather rude / demanding and thus low quality.

Saying "thanks" on its own, e.g. when a bug is fixed, is much more rewarding and positive.

fregante: > when a bug is fixed

I think that's what matters though, not the exact wording. I'm not sure it's worth implementing a check to exclude those though

fregante: @FloEdelmann should we block "thanks" like we do other words or can we close this PR?

FloEdelmann: I believe we should not block "thanks". So maybe closing is better in this case.

PR #6079, opened 2022-10-12
Preserve `toggle-files-button` after navigation

fregante:
- Fixes https://github.com/refined-github/refined-github/issues/6077

## Test URLs

1. From this issue
1. Open https://github.com/refined-github/refined-github

## Demo

gif

### DIFF
--- a/source/helpers/selector-observer.tsx
+++ b/source/helpers/selector-observer.tsx
@@ -4,6 +4,7 @@ import onetime from 'onetime';
 import {ParseSelector} from 'typed-query-selector/parser';
 
 import getCallerID from './caller-id';
+import isDevelopmentVersion from './is-development-version';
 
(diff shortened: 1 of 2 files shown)


source/helpers/selector-observer.tsx
```diff
@@ -40,12 +41,17 @@ export default function observe<
...
+		rule.setAttribute('s', selector);
+	}
+
 	rule.textContent = css`
 		:where(${String(selector)}):not(.${seenMark}) {
 			animation: 1ms ${animation};
 		}
 	`;
-	document.head.append(rule);
+	document.body.prepend(rule);
```
fregante, line 54: Turbo still messes with our style tags, so we just can't added to `head` or anywhere near.

PR #6082, opened 2022-10-13
Open RGH links in current tab unless we're on a form

fregante:
GitHub always opens links in the current tab. The only exception should be for situations where following a link results in loss of data (comments, WIP commits, etc)

### DIFF
--- a/source/features/link-to-github-io.tsx
+++ b/source/features/link-to-github-io.tsx
@@ -1,6 +1,6 @@
 import React from 'dom-chef';
 import * as pageDetect from 'github-url-detection';
-import {LinkExternalIcon} from '@primer/octicons-react';
+import {GlobeIcon} from '@primer/octicons-react';
 
 import features from '../feature-manager';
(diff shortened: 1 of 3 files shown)


source/features/link-to-github-io.tsx
```diff
@@ -10,8 +10,6 @@ function getLinkToGitHubIo(repoTitle: HTMLElement, className?: string): JSX.Elem
 	return (
 		<a
 			href={`https://${repoTitle.textContent!.trim()}`}
-			target="_blank"
-			rel="noopener noreferrer"
 			className={className}
 		>
 			<LinkExternalIcon className="v-align-middle"/>
```
fregante, line 15: @yakov116 thoughts on this? Should we change it to https://primer.style/octicons/globe-16 or https://primer.style/octicons/link-16?

yakov116: > [primer.style/octicons/globe-16](https://primer.style/octicons/globe-16)

Issue #5797, opened 2022-07-07
:has() selector discussion


fregante:
Feel free to leave a comment (or more) if you think of any features that can be rewritten or enhanced using the `has` selector. 

Live examples:

- https://github.com/refined-github/refined-github/pull/5795
- https://github.com/refined-github/refined-github/issues/5757#issuecomment-1177997611

But also, maybe:

- [ ] hiding tabs that have a zero (the number is often in an attribute as well, so it can be targeted)
- [ ] hiding events from the newsfeed
- [ ] showing our “update branch” button if the native button doesn’t exist, dynamically, without events! We append the link and just use `body:has(native button) .our-button {hide}` 🤯 

I swear this selector is a game changer for extensions. The only thing it can’t do is *alert us* of such events (maybe 😏 via empty `animationend` on these elements + JS listener)

fregante:
Chrome will see this on August 30, version 105:

- https://chromestatus.com/feature/5794378545102848
- https://chromiumdash.appspot.com/schedule

Also new `has`-only feature coming:

- https://github.com/refined-github/refined-github/pull/5858

fregante:
It is August 30th.

All new features and "restored features" can use `:has()`

Firefox will follow at some point. Tracking issues:

- https://bugzilla.mozilla.org/show_bug.cgi?id=418039
- it seems that it's been implemented, but still buggy and off by default https://news.ycombinator.com/item?id=32533720

jerone:
Please wait until Firefox has full support for `has`.
There are already [3 features](https://github.com/refined-github/refined-github/search?l=CSS&q=has) that use `has` in CSS. 
One of which I wanted to propose, only to found out it [already exists](https://github.com/refined-github/refined-github/blob/3996cc453ca530c441bda87235b4dda383b46cbb/source/features/readable-title-change-events.css#L3), but just doesn't work on Firefox.

fregante:
As mentioned, new features and bugfixes *can* already use the new selector. Firefox will get them whenever they'll get them. We're not going to retroactively break features, so it's ok.

fregante:
Possible new selector usage:


- [ ] CSS-only `dim-bots`, except for a possible ["expand bots" JS listener](https://github.com/refined-github/refined-github/pull/5946#pullrequestreview-1100448376)
- [ ] New hide repeated issue events https://github.com/refined-github/refined-github/issues/615#issuecomment-507524375

FloEdelmann:
[...]

PR #6088, opened 2022-10-16
Reduce `table-input` CPU usage thanks to :has selector

fregante:
- Related to #5797 

`mouseenter` events are triggered hundreds of times just by moving the mouse across the page, whether you’re using `table-input` or not; `delegate` will call `querySelector` in each event to determine whether to call our callback on the specific events we want.

Thankfully `:has` again comes to the rescue, even though the code is awful. I implemented this in a way to preserve backwards compatibility.

## Test URLs

This page

## Screenshot

gif

### DIFF
--- a/source/features/table-input.css
+++ b/source/features/table-input.css
@@ -6,6 +6,10 @@
 	grid-template: repeat(5, 20px) / repeat(5, 20px);
 }
 
+.rgh-table-input .sentinel { /* 🤷‍♂️ */
+	display: none;
+}
+
 :root .rgh-tic {
 	padding: 1px;
 }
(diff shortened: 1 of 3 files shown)


source/features/table-input.css
```diff
@@ -6,18 +6,33 @@
 	grid-template: repeat(5, 20px) / repeat(5, 20px);
 }
 
-:root .rgh-table-input-cell {
+.rgh-table-input .sentinel { /* 🤷‍♂️ */
+	display: none;
```
fregante, line 10: Did they add this element literally just now while I was writing this PR? I think this effectively forces me to push another release soon because the feature looks  wonky 🤔

fregante: They dropped it. What a coincidence.

fregante:
Feature verified both in Chrome and Firefox

PR #6089, opened 2022-10-17
bump follow-redirects package to 1.15

santhosh-chinnasamy:
Updated the version of `follow-redirects` npm package.

addresses https://github.com/advisories/GHSA-pw2r-vq6v-hr8c and https://github.com/advisories/GHSA-74fj-2j2h-c42q


fregante:
It's not used here

Issue #5081, opened 2021-11-17
Color gradient for files age


slavaGanzin:
### Description

Hi

I'm constantly looking for new GitHub projects and one thing that I care the most is maintenance activity (e.g. date of last commit).
So, I created a little extension, which runs only on GitHub pages. So fresh commits would become green and old — red. 
```javascript
Array.from(document.querySelectorAll('[datetime]')).map(x => 
  x.innerHTML = `<span style='filter: sepia(1) saturate(80) hue-rotate(${100 - 25 * (new Date() - new Date(x.getAttribute('datetime')))/(365 * 24 * 60 * 60 * 1000)}deg);'>${x.innerText}</span>`
) 
```
You can paste this in console on any [repository page](https://github.com/refined-github/refined-github)

Maybe it would be interesting for you to merge if I would create a PR?

### Screenshot

![2021-11-17_15-49](https://user-images.githubusercontent.com/1011721/142213102-99265c18-cc33-485f-a806-10697074eb0c.png)

### Example URL

https://github.com/refined-github/refined-github

fregante:
The feature is reasonable and it would match the blame page. The only issue I see is ensuring the colors are accessible and native-looking. We can try using the colors from the blame page itself. Mockup welcome.

slavaGanzin:
@fregante I was misguided by blame colors: red is a color of problem, while green is a color of change. It's natural: like fresh crops, and autumn leaves. Maybe, it would be better to make fresh changes green, and old - pale yellow/red? I think it's not a problem that `.editorconfig` is old, so it shouldn't be red. 
![2021-11-18_11-38](https://user-images.githubusercontent.com/1011721/142390025-b330f7cd-66af-4aa0-89c4-7e1103bfd7d0.png)

fregante:
Red and green are not good for this purpose. Try blame colors as suggested https://github.com/refined-github/refined-github/blame/main/contributing.md

slavaGanzin:
I dont think this way

kidonng:
> The feature is reasonable and it would match the blame page.

Thanks for the reminder, I remember there is something similar on GitHub already but cannot find it.

kidonng:
Reopening since the feature request itself is valid. We can go through implementation details later in the PR, not to mention colors can be easily overridden with custom CSS.

yakov116:
@fregante I tried this out I am not sure I like it.

Please let me know if I should continue.

[...]

PR #6095, opened 2022-10-21
Apply heatmap coloring to `release-download-count`

### LINKED ISSUE #5637: Apply heatmap coloring (like in blame view) to `release-download-count`
### Description

The blame view immediately tells you which lines are the most recent:

https://github.com/refined-github/refined-github/blame/main/package.json

We can use the same pattern/coloring in other areas of GitHub. There's already a feature request for this:

- https://github.com/refined-github/refined-github/issues/5081

But we can start laying down the foundation for a generic "pick heat based on min/max values of list" utility function in `release-download-count`

### Screenshot

Screen Shot 8

### Example URL

https://github.com/notepad-plus-plus/notepad-plus-plus/releases

foolmoron:
Closes #5637

- Added math helpers
  - utils for lerping
  - generic heatmap func that works with any list of values
- Added example of heatmap coloring for release download #s
  - Colors copied from github blame view's CSS: https://github.githubassets.com/assets/code-d21b20a03009.css

## Test URLs
https://github.com/notepad-plus-plus/notepad-plus-plus/releases
https://github.com/notepad-plus-plus/notepad-plus-plus/releases/tag/v8.4.6

## Screenshot
image


fregante:
I don't see any heat change in your screenshot, they're all almost the same color. The "no heat" color should be more gray, matching what GitHub does

foolmoron:
The numbers there are all very small other than 1.1M and 460k, so they will share the same heat. I have an idea for how to give more control to the gradient func. I'll try to update the colors too.

fregante:
In this case for it to be useful the map should match the minimum and maximum rather than use hardcoded values of what's hot and what's not. 

Also we can drop the previous logic to make the hotter download bold.

fregante:
Oh, also if they're all low-heat except 1-2 downloads (which is expected), it's find that they're mostly gray

source/helpers/math.ts
```diff
@@ -0,0 +1,36 @@
...
+	let max = values[0];
+	for (let i = 1; i < values.length; i++) {
+		if (values[i] < min) {
+			min = values[i];
+		} else if (values[i] > max) {
+			max = values[i];
+		}
+	}
+
+	return {min, max};
```
fregante, line 20:
```suggestion
	return {
		min: Math.min(...values),
		max: Math.max(...values),
	};
```

foolmoron: That method is pretty but sucks for medium/large arrays. GH has a lot of large lists so it's better to use the efficient one if this is gonna be a general util. My method is also 1 pass instead of 2.
https://medium.com/coding-at-dawn/the-fastest-way-to-find-minimum-and-maximum-values-in-an-array-in-javascript-2511115f8621

fregante: There are no such list of any size on GitHub close to making this "inefficient". With numbers always less than 100, we prefer readability over microoptimizations

foolmoron:
> I don't see any heat change in your screenshot, they're all almost the same color. The "no heat" color should be more gray, matching what GitHub does. See [#5081 (comment)](https://github.com/refined-github/refined-github/issues/5081#issuecomment-1180431673) to find out how to use GitHub’s own colors, we shouldn't generate our own.

Hm looks like heat colors file isn't loaded on all pages, so we can't use it directly. I can load the file , or copy/paste it with a comment explaining where it comes from.
https://github.githubassets.com/assets/code-d21b20a03009.css

imo copy/paste gives more control and is not much less maintainable.

source/helpers/math.ts
foolmoron: I was thinking of line numbers in files and diffs. But sure, will change.

fregante:
> imo copy/paste gives more control and is not much less maintainable.

Sounds fine to me. My only concern was around theming, but it doesn't seem that they're affected by this.

You can just copy the rules that include `data-heat` like:

```css
.blame-commit[data-heat="1"]{border-right:2px solid #f66a0a}.blame-commit[data-heat="2"]{border-right:2px solid rgba(246,106,10,.9)}.blame-commit[data-heat="3"]{border-right:2px solid rgba(246,106,10,.8)}.blame-commit[data-heat="4"]{border-right:2px solid rgba(246,106,10,.7)}.blame-commit[data-heat="5"]{border-right:2px solid rgba(246,106,10,.6)}.blame-commit[data-heat="6"]{border-right:2px solid rgba(246,106,10,.5)}.blame-commit[data-heat="7"]{border-right:2px solid rgba(246,106,10,.4)}.blame-commit[data-heat="8"]{border-right:2px solid rgba(246,106,10,.3)}.blame-commit[data-heat="9"]{border-right:2px solid rgba(246,106,10,.2)}.blame-commit[data-heat="10"]{border-right:2px solid rgba(246,106,10,.1)}.heat[data-heat="1"]{background:#f66a0a}.heat[data-heat="2"]{background:rgba(246,106,10,.9)}.heat[data-heat="3"]{background:rgba(246,106,10,.8)}.heat[data-heat="4"]{background:rgba(246,106,10,.7)}.heat[data-heat="5"]{background:rgba(246,106,10,.6)}.heat[data-heat="6"]{background:rgba(246,106,10,.5)}.heat[data-heat="7"]{background:rgba(246,106,10,.4)}.heat[data-heat="8"]{background:rgba(246,106,10,.3)}.heat[data-heat="9"]{background:rgba(246,106,10,.2)}.heat[data-heat="10"]{background:rgba(246,106,10,.1)}.blame-commit-date{font-size:12px;line-height:25px;flex-shrink:0}.blame-commit-date[data-heat="1"]{color:#c24e00}.blame-commit-date[data-heat="2"]{color:#ac571f}.blame-commit-date[data-heat="3"]{color:#a35b2c}.blame-commit-date[data-heat="4"]{color:#9a5f38}.blame-commit-date[data-heat="5"]{color:#926245}.blame-commit-date[data-heat="6"]{color:#896651}.blame-commit-date[data-heat="7"]{color:#806a5e}.blame-commit-date[data-heat="8"]{color:#776d6a}.blame-commit-date[data-heat="9"]{color:#6e7177}.blame-commit-date[data-heat="10"]{color:#6a737d}
```

But the classes need to be changed to `.rgh-blah-blah` to make sure we don't override _their_ CSS.

foolmoron:
@fregante Alright, updated in latest. Also updated the description and added a new screenshot. wdyt? Looks pretty nice imo

source/refined-github.css
```diff
@@ -191,3 +191,85 @@ div[data-target='readme-toc.content'] div.blob-wrapper img {
 [data-selected-links^='repo_milestones ']:has([title='0']) svg {
 	margin-right: 20px; /* Push the text farther away from the text so it's cropped out */
 }
+
+/* Heatmap colors copied from GH, because GH does not load them on every page #5637 */
```
fregante, line 195: These should be in their own `heat-map.css` file, with a comment that includes the source of this code so it can be updated again in the future.

source/helpers/math.ts
```diff
@@ -0,0 +1,31 @@
...
+}
+
+export function calculateMinMax(values: number[]): {min: number; max: number} {
+	return {
+		min: Math.min(...values),
+		max: Math.max(...values),
+	};
+}
+
+export function createHeatIndexFunc(values: number[]): (value: number) => number {
```
yakov116, line 16: Maybe `findHeatIndex`?

source/helpers/math.ts
```diff
@@ -0,0 +1,31 @@
+export function lerp(min: number, max: number, t: number): number {
```
yakov116, line 1: Use real words instead of letters. What is `t` times?

source/features/release-download-count.tsx
```diff
@@ -59,9 +60,10 @@ async function addCounts(assetsList: HTMLElement): Promise<void> {
 	const releases = [[releaseName, assetsList]] as const;
 	for (const [name, release] of releases) {
 		const sortedDownloads = assets[api.escapeKey(name)].sort((a, b) => b.downloadCount - a.downloadCount);
```
fregante, line 62: We're no longer using this sorted array. At a minimum we should drop the sorting
```suggestion
		const downloads = assets[api.escapeKey(name)];
```

Better yet if we avoid a loop below by making this a map:

```suggestion
		const downloads = new Map(assets[api.escapeKey(name)].map(asset => [asset.name, asset.downloadCount]))
```

And then later replace the `sortedDownloads` loop with:

```js
const downloadCount = downloads.get(assetName.textContent!)
if (!downloadCount) {
	continue;
}

**Comment by fregante** [comment]
Starting to look good! I'd probably color the related icon. I hope it's still all readable. I've been using the colors on the file list but contrast definitely takes a hit. We'll see how people respond.

**Comment by foolmoron** [comment]
Yeah, one option I was thinking is to restrict the heatmap to just the top 3-5 values. So the contrast is largely maintained at the cost of granularity.

Here's the various options of icon vs text coloration. Take your pick!

Just text
image

Just icon
image

Both
image

File: source/features/release-download-count.tsx (hunk 5 above)
Comment (foolmoron) [reply source/features/release-download-count.tsx:62]: Nice, updated in latest.

File: source/refined-github.css (hunk 2 above)
Comment (foolmoron) [reply source/refined-github.css:195]: Updated in latest. I put the file in the root for now since it can be used across many features.

File: source/helpers/math.ts (hunk 4 above)
Comment (foolmoron) [reply source/helpers/math.ts]: Those are the standard names for interp functions. Tried to be more explict in latest, wdyt?

File: source/helpers/math.ts (hunk 3 above)
Comment (foolmoron) [reply source/helpers/math.ts]: It's a little weird since returns a function, not the head index directly

File: source/refined-github.ts [hunk 6]
```diff
@@ -1,6 +1,7 @@
 import './helpers/types.d.ts';
 
 import './refined-github.css';
+import './heat-map.css';
```
yakov116, line 4: I think this should go in the bottom

source/helpers/math.ts
```diff
@@ -0,0 +1,53 @@
...
+ * The list is used to find the min and max of the range. The function can then
+ * calculate a heat index for a given value based on that range.
+ *
+ * GitHub has 10 heat values, and uses a lower heat index for a higher value. So the
+ * highest heat index that will be returned is 1, and the lowest is 10.
+ *
+ * @param numbers values to use to calculate the min and max of the heat range
+ * @returns function to calculate heat index of a given number
+ */
+export function createHeatIndexFunc(numbers: number[]): (value: number) => number {
```
yakov116, line 38: Maybe I am not understanding what this does. But why cant this return a map?

foolmoron: What would the map look like? The input value can be anything within the min/max range. It's not a limited set of possibilities.

yakov116: Ok I _think_ I understand what you are doing.
Your assigning the heatmap relative to the download count of the page?

So if I have only 2 releases with them both having 5k Downloads what will happen?

foolmoron: That's right. In that case both releases would have the lowest heat since it floors the heat value, so they would both show up as gray.

Oh you made me realize that line 24 would have a divide by zero in that case, let me fix that.

yakov116: I think that the heat map should be a set number not relative to whats on the page.

500k should be dark red regardless if there are any other downloads on the page.

source/refined-github.ts
foolmoron: In latest

source/helpers/math.ts
foolmoron: Totally doable by passing in hardcoded min/max values to the func. I think @fregante wanted it to be dynamic though.

fregante: @yakov116 the problem with hardcoded values is that most repos will appear just black. Making it dynamic would let us use the heatmap in place of "making the most downloaded file bolder"

yakov116: Ok 👍

foolmoron: Any other revisions? Would like to get this in for hacktoberfest :D

source/helpers/math.ts
yakov116: @fregante do you have a better name for this function?

fregante: I think it's a reasonable name, other than "Func" which can just be Function. I'm not sure why eslint didn't catch it https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prevent-abbreviations.md

fregante: the "lerp" functions should be extended though. Use "linear interpolation" or something else. calculateMinMax can just be inlined as:

```js
const min = Math.min(...values);
const max = Math.max(...values);
```

foolmoron: Made the changes in latest!

source/helpers/math.ts
```diff
@@ -0,0 +1,57 @@
+/**
+ * Linear interpolation. Given a range and a value, returns the
+ * percentage position in the range for the value.
+ *
+ * @param min Lowest value in the range
+ * @param max Highest value in the range
+ * @param interpolation Interpolation amount from 0.0 to 1.0 (or beyond to get a value outside the range [min, max])
+ * @returns Interpolated value
+ */
+export function linearInterpolation(min: number, max: number, interpolation: number): number {
```
yakov116, line 10: Why is this function exported?

source/helpers/math.ts
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Inverse operation of linear interpolation. Given a range and a
+ * value, returns the interpolation amount to get that value.
+ *
+ * @param min Lowest value in the range
+ * @param max Highest value in the range
+ * @param value Value in the range
+ * @returns Interpolation amount from 0.0 to 1.0 (or beyond if `x` is outside the range [min, max])
+ */
+export function inverseLinearInterpolation(min: number, max: number, value: number): number {
```
yakov116, line 23: Why is this function exported?

foolmoron:
Both are useful functions for general use

fregante:
I'll work on it 👍

source/helpers/math.ts
foolmoron: Good catch, I deleted the code that used this. Fixed in latest.

source/helpers/math.ts
foolmoron: Fixed in latest

foolmoron:
@fregante I think it should be good now if there's no other comments!

fregante:
It was working until I refreshed the page and then all the downloads got a heat of 10 🤔 I'm confused.

fregante:
Ok fixed. 

.values() returned an iterable which was then empty after the first iteration 😅

foolmoron:
Nice!

fregante:
Updating the screenshot in the readme too:



Screen Shot 1

Also mobile screenshot for future reference 

Screen Shot 2

source/helpers/math.ts
fregante: I restored the export because the file is named "math"

fregante:
Thank you for the PR, Momin!

PR #6103, opened 2022-10-23
Meta: Use new APIs in `update-pr-from-base-branch`

yakov116:
## Test URLs
https://github.com/refined-github/sandbox/pull/9 (Conflict)
https://github.com/refined-github/sandbox/pull/11

## Screenshot
NA

### DIFF
--- a/source/features/restore-file.tsx
+++ b/source/features/restore-file.tsx
@@ -1,6 +1,5 @@
 import React from 'dom-chef';
 import select from 'select-dom';
-import onetime from 'onetime';
 import pushForm from 'push-form';
 import * as pageDetect from 'github-url-detection';
 import delegate, {DelegateEvent} from 'delegate-it';
@@ -15,7 +14,8 @@ import {getConversationNumber} from '../github-helpers';
 Get the current base commit of this PR. It should change after rebases and merges in this PR.
 This value is not consistently available on the page (appears in `/files` but not when only 1 commit is selected)
 */
-const getBaseReference = onetime(async (): Promise<string> => {
+// TODO: Replace this with `get-pr-info` when GHE supports it
+const getBaseReference = async (): Promise<string> => {
 	const {repository} = await api.v4(`
 		repository() {
 			pullRequest(number: ${getConversationNumber()!}) {
@@ -24,7 +24,7 @@ const getBaseReference = onetime(async (): Promise<string> => {
 		}
 	`);
 	return repository.pullRequest.baseRefOid;
-});
+};
 
 async function getFile(filePath: string): Promise<{isTruncated: boolean; text: string} | undefined> {
 	const {repository} = await api.v4(`
--- a/source/features/restore-file.tsx
+++ b/source/features/restore-file.tsx
@@ -9,15 +9,21 @@ import features from '../feature-manager';
 import * as api from '../github-helpers/api';
 import fetchDom from '../helpers/fetch-dom';
 import showToast from '../github-helpers/toast';
-import {getPrInfo} from '../github-helpers/get-pr-info';
+import {getConversationNumber} from '../github-helpers';
 
 /**
 Get the current base commit of this PR. It should change after rebases and merges in this PR.
 This value is not consistently available on the page (appears in `/files` but not when only 1 commit is selected)
 */
 const getBaseReference = onetime(async (): Promise<string> => {
-	const {baseRefOid} = await getPrInfo();
-	return baseRefOid;
+	const {repository} = await api.v4(`
+		repository() {
+			pullRequest(number: ${getConversationNumber()!}) {
+				baseRefOid
+			}
+		}
+	`);
+	return repository.pullRequest.baseRefOid;
 });
 
 async function getFile(filePath: string): Promise<{isTruncated: boolean; text: string} | undefined> {
--- a/source/features/restore-file.tsx
+++ b/source/features/restore-file.tsx
@@ -15,7 +15,7 @@ Get the current base commit of this PR. It should change after rebases and merge
 This value is not consistently available on the page (appears in `/files` but not when only 1 commit is selected)
 */
 // TODO: Replace this with `get-pr-info` when GHE supports it
-const getBaseReference = async (): Promise<string> => {
+async function getBaseReference(): Promise<string> {
 	const {repository} = await api.v4(`
 		repository() {
 			pullRequest(number: ${getConversationNumber()!}) {
@@ -24,7 +24,7 @@ const getBaseReference = async (): Promise<string> => {
 		}
 	`);
 	return repository.pullRequest.baseRefOid;
-};
+}
 
 async function getFile(filePath: string): Promise<{isTruncated: boolean; text: string} | undefined> {
 	const {repository} = await api.v4(`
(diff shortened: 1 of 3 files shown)


source/github-helpers/get-pr-info.ts
```diff
@@ -8,15 +8,27 @@ type PullRequestInfo = {
...
 
-export default async function getPrInfo(number = getConversationNumber()!): Promise<PullRequestInfo> {
+export default async function getPrInfo(head: string, number = getConversationNumber()!): Promise<PullRequestInfo> {
 	const {repository} = await api.v4(`
 		repository() {
 			pullRequest(number: ${number}) {
 				baseRefOid
 				mergeable
 				viewerCanEditFiles
+				headRef {
```
yakov116, line 25: @fregante any reason I should make a new function? This function is not used anywhere else.

source/github-helpers/get-pr-info.ts
```diff
@@ -8,15 +8,27 @@ type PullRequestInfo = {
...
+					compare(headRef: "${base}") {
+						status
+						behindBy
+						aheadBy
```
yakov116, line 29: I put this here in case you wanted to do https://github.com/refined-github/refined-github/issues/3863#issuecomment-1192756619

yakov116:
@busches will this affect GHE?

busches:
@yakov116 you can check the GHE docs to see what, if any, versions support it yet. Guessing it's not in any version yet.

fregante:
@yakov116 we have probably never waited for GHE to use updated APIs

yakov116:
This will break the entire feature, since the api call will fail.

I don't remember im the past ever having a case where an API was not available on GHE.

However I do remember holding off on v4 until GHE had normal support.

busches:
I do use this feature all the time, so I appreciate the thought. 😃

source/features/restore-file.tsx
```diff
@@ -9,21 +9,15 @@ import features from '../feature-manager';
...
 import fetchDom from '../helpers/fetch-dom';
 import showToast from '../github-helpers/toast';
-import {getConversationNumber} from '../github-helpers';
+import {getPrInfo} from '../github-helpers/get-pr-info';
 
 /**
 Get the current base commit of this PR. It should change after rebases and merges in this PR.
 This value is not consistently available on the page (appears in `/files` but not when only 1 commit is selected)
 */
 const getBaseReference = onetime(async (): Promise<string> => {
```
fregante, line 18: We can drop `onetime` while we're at it. All API calls are memoized.

source/features/update-pr-from-base-branch.tsx
```diff
@@ -50,13 +50,13 @@ async function addButton(position: Element): Promise<void> {
...
 		getPrInfo(),
-
-		// TODO: Find how to determine whether the branch needs to be updated via v4
-		// `page=10000` avoids fetching any commit information, which is heavy
-		api.v3(`compare/${base}...${head}?page=10000`),
+		// TODO: Drop v3 when GHE supports this via v4
+		pageDetect.isEnterprise()
+			? api.v3(`compare/${base}...${head}?page=10000`) // `page=10000` avoids fetching any commit information, which is heavy
+			: getPrBranchAheadStatus(base),
 	]);
```
fregante, line 57: We should make obsolete code self-contained so that when the time comes we can just delete lines and have a negative diff, without having to delve into the API.

Since this now requires only one call, this piece of code should just be:

```js
const pr = await getPrInfo();
```

In it, we can have the GHE condition that makes two calls if necessary, like:

```js
function getPrInfo() {
	if(GHE) {
		return {
			other: await api0(),
			compare: await api1()
		}
	}

	return api.v4(our new query)
}

In the current code, everyone still gets two requests, so this PR becomes of little utility for GitHub.com users at this point.

Comment (yakov116) [reply source/features/update-pr-from-base-branch.tsx:57]: > In the current code, everyone still gets two requests, so this PR becomes of little utility for GitHub.com users at this point.

The problem is the I want to use this utility for `restore-file` but if I combine the requests then I will not be able to use it.

Your call

File: source/github-helpers/get-pr-info.ts [hunk 5]
```diff
@@ -31,7 +31,7 @@ export default async function getPrInfo(base: string, head: string, number = get
 
 		return {
 			prInfo: repository.pullRequest,
-			comparison: compare.status.toUpperCase(),
```
yakov116: I need type help

source/github-helpers/get-pr-info.ts
```diff
@@ -31,7 +31,7 @@ export default async function getPrInfo(base: string, head: string, number = get
...
+			comparison: compare,
```
yakov116, line 34: I need type help

source/features/update-pr-from-base-branch.tsx
fregante: It would be nice but not worth it. Your reversal looks good

source/github-helpers/get-pr-info.ts
fregante: Will try later this week

source/github-helpers/get-pr-info.ts
fregante: Indentation needs some fixing

source/github-helpers/get-pr-info.ts
```diff
@@ -1,24 +1,59 @@
...
+		// TODO: Use this for `restore-file` when GHE supports `compare`
+		baseRefOid: string;
+		// https://docs.github.com/en/graphql/reference/enums#mergeablestate
+		mergeable: 'CONFLICTING' | 'MERGEABLE' | 'UNKNOWN';
+		viewerCanEditFiles: boolean;
+	};
+	comparison: {
+		status: 'BEHIND' | 'DIVERGED' | 'AHEAD' | 'IDENTICAL';
+	};
 };
```
fregante, line 17: You can merge these into flat object, no need to have prInfo and comparison keys. While we're at it it can be simplified further, like setting `diverged: true` instead of matching strings outside

source/github-helpers/get-pr-info.ts
yakov116: ?
What are you referring to?

fregante: Ah sorry I must have commented on the wrong thing: look for the `repository()` line that is indented one time too many.

fregante: image

source/github-helpers/get-pr-info.ts
yakov116: > While we're at it it can be simplified further, like setting `diverged: true` instead of matching strings outside

Can you explain this one more time I am not sure I understood what you said

source/github-helpers/get-pr-info.ts
```diff
@@ -1,24 +1,55 @@
...
+				}
 			}
 		}
 	`);
+
+	if (repository.pullRequest.headRef.compare.status !== 'DIVERGED') {
+		return;
+	}
+
 	return repository.pullRequest;
```
fregante, line 54: This just mixes the logic of a specific feature with this generic `get pr info` function, which could return the PR info regardless of the need to update the PR.

I suggested returning `diverged: true` for both API calls, i.e. something like

```suggestion
	return {
		...repository.pullRequest,
		diverged: repository.pullRequest.headRef.compare.status === 'DIVERGED'
	};
```

And the logic just checks `pr.diverged`, or better yet `pr.needsUpdate` as a combination of `mergeable` and `diverged`


---

I suppose it's ok as is for now if it's tested, but it'll need to be changed eventually.  👍

fregante (approved):

PR #6112, opened 2022-10-26
Add tags to  `select-notifications`

alk-sdavid:
Hi, this PR adds Tags type to the select notifications dropdown.

## Test URLs

## Screenshot

Screenshot 2022-10-26 at 11 11 01

### DIFF
--- a/source/features/select-notifications.tsx
+++ b/source/features/select-notifications.tsx
@@ -13,6 +13,7 @@ import {
 	GitPullRequestDraftIcon,
 	GitPullRequestIcon,
 	IssueOpenedIcon,
+	TagIcon,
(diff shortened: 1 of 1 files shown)


fregante:
Thank you for the PR. I'm not sure we should add many more filters here. There's a lot of "potential" for infinite filters and the dropdown will just get too long. See also:

- https://github.com/refined-github/refined-github/pull/5553

alk-sdavid:
ok

fregante:
We could potentially add an "everything else" filter though.

alk-sdavid:
yes it's what I was thinking about 🤔 I might do that soon 🤓

PR #6113, opened 2022-10-27
`preview-hidden-comments` even if reason is missing

### LINKED ISSUE #6098: `preview-hidden-comments` show comment preview even if reason is unknown
### Description

As mentioned in https://github.com/refined-github/refined-github/issues/6017 some hidden comments don't have the reason included, meaning the comment would remain hidden. 

Could this be changed so it would show the comment in all cases, even with no reason?

It seems like this could be implemented by removing:

https://github.com/refined-github/refined-github/blob/ebfb5f62b9f92a677402ffe9d61d9c1eee27824c/source/features/preview-hidden-comments.tsx#L21-L23

and only adding the reason label if it is not null:

https://github.com/refined-github/refined-github/blob/ebfb5f62b9f92a677402ffe9d61d9c1eee27824c/source/features/preview-hidden-comments.tsx#L29

### Screenshot

_No response_

### Example URLs

https://github.com/refined-github/refined-github/issues/3647

Katsute:
Closes #6098

* Write label only if reason exists
* Write comment in all cases (previously would only write if a reason existed)
* Keep comment hidden if disruptive or spam (only if reason exists)

## Test URLs

https://github.com/refined-github/refined-github/issues/3647

https://github.com/refined-github/sandbox/pull/47

## Screenshot

Comment with missing reason:

|Before|After|
|---|---|
|image|image|

Comment with reason:

|Before|After|
|---|---|
|image|image|

### DIFF
--- a/source/features/preview-hidden-comments.tsx
+++ b/source/features/preview-hidden-comments.tsx
@@ -16,7 +16,7 @@ function preview(hiddenCommentHeader: HTMLElement): void {
 	if (commentText.length === 0) {
 		return;
 	}
-	
+
 	const commentHeader = hiddenCommentHeader.textContent!;
 	if (/disruptive|spam/.test(commentHeader)) {
 		return;
--- a/source/features/preview-hidden-comments.tsx
+++ b/source/features/preview-hidden-comments.tsx
@@ -16,13 +16,14 @@ function preview(hiddenCommentHeader: HTMLElement): void {
 	if (commentText.length === 0) {
 		return;
 	}
-
-	const reason = /duplicate|outdated|off-topic|hidden/.exec(hiddenCommentHeader.textContent!)?.[0];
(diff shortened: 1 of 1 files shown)


fregante:
Thank you for the PR! The condition should be kept because we still don't want to show the message when *we know* it's spam. So it should be something like:

```
const header = h.textContent;
const reason = currentRegex.exec(header)
if (!reason || !header)
  return
```

Katsute:
Restored behavior where spam and abuse would be hidden (only if reason exists).

```ts
	const hide = /disruptive|spam/.exec(hiddenCommentHeader.textContent!)?.[0];
	if (hide) {
		return;
	}
```

I wasn't able to hide spam and abuse in the first case because there is no reason and no other way to differentiate between spam and non-spam hidden comments. 😥

fregante (approved):

fregante (approved):

PR #6120, opened 2022-10-30
Add `file-age-color` feature

### LINKED ISSUE #5081: Color gradient for files age

yakov116:
Closes #5081

## Test URLs
https://github.com/refined-github/refined-github/tree/main/source/features

## Screenshot
image

Dark Mode looks much better

image


readme.md
```diff
@@ -185,6 +185,7 @@ Thanks for contributing! 🦋🙌
 - [](# "file-finder-buffer") [Lets you start typing your search immediately after invoking the File Finder (<kbd>t</kbd>), instead of having you wait for it to load first.](https://user-images.githubusercontent.com/1402241/75542106-1c811700-5a5a-11ea-8aa5-bea0472c59e2.gif)
 - [](# "show-associated-branch-prs-on-fork") [Shows the associated pull requests on branches for forked repositories.](https://user-images.githubusercontent.com/16872793/81504659-7e5ec800-92b8-11ea-9ee6-924110e8cca1.png)
 - [](# "html-preview-link") [Adds a link to preview HTML files.](https://user-images.githubusercontent.com/44045911/67634792-48995980-f8fb-11e9-8b6a-7b57d5b12a2f.png)
+- [](# "file-age-colors") [Add wording that needs to be fixed](https://user-images.githubusercontent.com/16872793/178278788-277724fd-530b-420f-a3d0-d3597d7acfe8.png)
```
yakov116, line 188: Any suggestions for a better name?

fregante: We could be more specific using "heatmap" but colors is easier to understand and find. Suggestions still welcome

yakov116: I think we should leave it as is and change it when we add it to other places.

@refined-github/maintainers any one have any suggestions?

fregante: > I think we should leave it as is and change it when we add it to other places.

Renaming features has a cost (it breaks [the search](https://github.com/refined-github/refined-github/blob/844d8742c6af99ced0e6ae6a1b21618273b13bee/source/features/rgh-feature-descriptions.tsx#L20-L21) after every rename) so I've been trying not to.

I thought about it and I'm ok with having multiple features that do similar things in different areas, depending on how much they differ. We already share specific functions between features.

source/features/file-age-color.tsx
```diff
@@ -0,0 +1,47 @@
...
+	return diffDays;
+}
+
+function addHeatIndex(timeAgo: HTMLElement): void {
+	const dateDiff = daysDiff(new Date(timeAgo.attributes.datetime.value));
+	let heatIndex = '0';
+	for (const [days, heat] of heatMap) {
+		if (dateDiff <= Number(days)) {
+			heatIndex = heat;
+			return;
```
fregante, line 31: If you return here, it's not going to be set in the attribute.

Also using `.find()` might be more readable (except you'll need `[...heatMap.entries()].find()`)

source/features/file-age-color.tsx
```diff
@@ -0,0 +1,47 @@
...
+	['3', '1'],
+	['6', '2'],
+	['15', '3'],
+	['24', '4'],
+	['40', '5'],
+	['70', '6'],
+	['100', '7'],
+	['160', '8'],
+	['275', '9'],
+	['30000', '10'],
```
fregante, line 16: We don't need these hard-coded numbers nor do we need to find a difference in "days" exactly (because JS makes that awkward to calculate).

Could you turn this into a mathematical function? This is what it should look like:

```js
function addHeatIndex(timeAgo: HTMLElement): void {
	const diff = Date.now() - new Date(timeAgo.attributes.datetime.value).getTime();
	timeAgo.dataset.rghHeat = Math.max(0, Math.min(10, Math.sqrt(diff / 1000)))
}
```

That exact "1000" number needs to be figured out. Other math functions are also welcome if they're easier to write for you. The square root looks similar to the type of curve I'm imagining.

fregante (changes requested):

source/features/file-age-color.tsx
fregante: There are some "mathematical function finders" online where you can plug in those numbers in a x/y table and you get an approximated function you can paste here. It's probably easier than figuring out that srqt value.

yakov116: I pushed a change using your idea.

It supports up to 125 days it that better?

fregante:
Once again, I wrote a 2-line code suggestion that works nearly out of the box and ignored half of the comment 🤷‍♂️

yakov116:
> Once again, I wrote a 2-line code suggestion that works nearly out of the box and ignored half of the comment 🤷‍♂️

Does not work returns 10 each and everytime

Forget it

fregante:
> Does not work returns 10 each and everytime

The comment mentioned that `1000` needed to be replaced. `400000000` is closer to what we're going for, although it's not exact:

```js

 const heatMap = new Map([
 	['3', '1'],
 	['6', '2'],
 	['15', '3'],
 	['24', '4'],
 	['40', '5'],
 	['70', '6'],
 	['100', '7'],
 	['160', '8'],
 	['275', '9'],
 	['30000', '10'],
 ]);
[...heatMap.keys()].map(days => Math.round(Math.max(0, Math.min(10, Math.sqrt((days * 24 * 60 * 60 * 1000) / 400000000)))))
// [1, 1, 2, 2, 3, 4, 5, 6, 8, 10]
// this only demoes the output in order to compare it to the previous map
// and it's not meant to be used this way
```


The second suggestion was to use a function generator

PR #6124, opened 2022-11-01
Add behind count to `update-pr-from-base-branch`

### LINKED ISSUE #3863: Show base commit in PRs (and not just the base branch)
PRs show the base branch, but they don't tell how long ago the current branch was updated. Example:

```
master         A-B-C-D-E  
                    \     
feature-branch       C-Q-Z    
```

In this case, this feature would tell us that `C` is the real base

I've wanted to know this in the past, but at the moment I'm only thinking of why we _shouldn't_ implement this: 😅

- Our `update-pr-from-base-branch` feature already tells us _whether_ it's up to date and it allows us to update it

Discussion welcome. Would this info be useful? When?

yakov116:
Closes #3863

## Test URLs
https://github.com/refined-github/sandbox/pull/53

## Screenshot
image

### DIFF
--- a/source/features/update-pr-from-base-branch.tsx
+++ b/source/features/update-pr-from-base-branch.tsx
@@ -58,8 +58,7 @@ async function addButton(position: Element): Promise<void> {
 	position.append(' ', (
 		<span className="status-meta d-inline-block">
 			{select('.head-ref')!.cloneNode(true)} is {pluralize(prInfo.headRef.compare.behindBy, '$$ commit', '$$ commits')} behind {select('.base-ref')!.cloneNode(true)}
-			{' ('}<a className="btn-link" href={buildRepoURL('commits/' + prInfo.baseRefOid)}>{prInfo.baseRefOid.slice(0, 8)}</a>)
-			<button type="button" className="btn-link rgh-update-pr-from-base-branch">update branch</button>.
+			{' ('}<a className="btn-link" href={buildRepoURL('commits/' + prInfo.baseRefOid)}>{prInfo.baseRefOid.slice(0, 8)}</a>)<button type="button" className="btn-link rgh-update-pr-from-base-branch">update branch</button>.
 		</span>
 	));
 }
--- a/source/features/update-pr-from-base-branch.tsx
+++ b/source/features/update-pr-from-base-branch.tsx
@@ -8,7 +8,8 @@ import features from '../feature-manager';
 import observe from '../helpers/selector-observer';
 import * as api from '../github-helpers/api';
 import getPrInfo from '../github-helpers/get-pr-info';
-import {getConversationNumber} from '../github-helpers';
+import pluralize from '../helpers/pluralize';
+import {buildRepoURL, getConversationNumber} from '../github-helpers';
 
 const selectorForPushablePRNotice = '.merge-pr > .color-fg-muted:first-child';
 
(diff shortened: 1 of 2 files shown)


source/features/update-pr-from-base-branch.tsx
```diff
@@ -49,16 +50,23 @@ async function handler({delegateTarget}: DelegateEvent): Promise<void> {
...
+		if (pageDetect.isEnterprise()) {
+			position.append(' ', (
+				<span className="status-meta d-inline-block rgh-update-pr-from-base-branch">
+					You can <button type="button" className="btn-link">update the base branch</button>.
+				</span>
+			));
+		} else {
+			position.append(' ', (
+				<span className="status-meta d-inline-block">
+					{select('.head-ref')!.cloneNode(true)} is {pluralize(prInfo.headRef.compare.behindBy, '$$ commit', '$$ commits')} behind {select('.base-ref')!.cloneNode(true)}
```
fregante, line 64: The head is available just a few words before this node, we shouldn't repeat it. If anything we can replace/reword it. 

Before doing further work, let's see how else we can word it or what information we actually want here.

source/features/update-pr-from-base-branch.tsx
```diff
@@ -49,16 +50,23 @@ async function handler({delegateTarget}: DelegateEvent): Promise<void> {
...
+					{' ('}<a className="btn-link" href={buildRepoURL('commits/' + prInfo.baseRefOid)}>{prInfo.baseRefOid.slice(0, 8)}</a>)
+					[<button type="button" className="btn-link rgh-update-pr-from-base-branch">update branch</button>].
```
fregante, line 66: GitHub doesn't use this style anywhere, brackets can be dropped

source/features/update-pr-from-base-branch.tsx
```diff
@@ -49,16 +50,23 @@ async function handler({delegateTarget}: DelegateEvent): Promise<void> {
...
-	}
 
-	if (prInfo.viewerCanEditFiles && prInfo.mergeable !== 'CONFLICTING') {
-		position.append(' ', (
-			<span className="status-meta d-inline-block rgh-update-pr-from-base-branch">
-				You can <button type="button" className="btn-link">update the base branch</button>.
-			</span>
-		));
+	if (prInfo.needsUpdate) {
+		if (pageDetect.isEnterprise()) {
```
fregante, line 55: Why?

source/features/update-pr-from-base-branch.tsx
```diff
@@ -49,16 +50,23 @@ async function handler({delegateTarget}: DelegateEvent): Promise<void> {
...
-		return;
-	}
 
-	if (prInfo.viewerCanEditFiles && prInfo.mergeable !== 'CONFLICTING') {
-		position.append(' ', (
-			<span className="status-meta d-inline-block rgh-update-pr-from-base-branch">
-				You can <button type="button" className="btn-link">update the base branch</button>.
-			</span>
-		));
+	if (prInfo.needsUpdate) {
```
fregante, line 54:
```suggestion
	if (!prInfo.needsUpdate) {
    return
   }
```

source/features/update-pr-from-base-branch.tsx
yakov116: Because they do not support `compare` yet

fregante: We still have this information via v3 API, right?

yakov116: Your correct I did not realize that this information exists

source/features/update-pr-from-base-branch.tsx
fregante: There's also the problem that this feature only appears when the native button doesn't (or at least that's the intention), so we probably shouldn't bundle this information with this feature. 

What if we just add it to the PR header next to the branches? Like:

> (Open) main <- feature (-5)

The number of "+" commits is already visible on the PR Commits tab. The number could show further information on in a tooltip. 

I'm not totally convinced by this however as the commit could be useful to be copied. I'm open to pushing a minimal version and then iterate on it as we start to rely on this piece of information. 

I'd also perhaps just append a simple version of this to both "Update branch" buttons (ours and the native one), and make it easier. Those areas at the bottom have plenty of space

yakov116: > I'd also perhaps just append a simple version of this to both "Update branch" buttons (ours and the native one), and make it easier. Those areas at the bottom have plenty of space

Lost you on this line

yakov116: > What if we just add it to the PR header next to the branches? Like:
> 
> > (Open) main <- feature (-5)

Like I as going to try to do this

yakov116: Should it be part of `clean-conversation-headers`

fregante: > Lost you on this line

# Native

Screen Shot 5

# Ours

Screen Shot 6

# Change requested 

Show "base commit" near both of these. This feature isn't necessarily related to any other. It's not cleaning and it's not necessarily updating the branch.

fregante: > > What if we just add it to the PR header next to the branches? Like:
> >
> > (Open) main <- feature (-5)
> Like I as going to try to do this, where are we putting the base commit link?

👇 

> The number could show further information in a tooltip.



With "further information" being the base commit. I followed that with a comment regarding the need to copy or see this value.

yakov116: In other words we should change it to?

image

PR #6128, opened 2022-11-02
Lint

yakov116:
I hope to do some https://github.com/refined-github/refined-github/issues/5222 too


yakov116:
Don't merge this the icons are messed up. I need to look into it

source/features/sticky-csv-header.css
```diff
@@ -13,8 +13,8 @@
 
 .blob-wrapper.type-csv,
 .Box:has(.csv-data) {
-	border-bottom-right-radius: 0;
-	border-bottom-left-radius: 0;
+	border-bottom-right-radius: none;
+	border-bottom-left-radius: none;
```
FloEdelmann, line 17: This is not allowed, see https://developer.mozilla.org/en-US/docs/Web/CSS/border-radius

Also see https://github.com/xojs/stylelint-config-xo/pull/17, probably you have an old version locally.

FloEdelmann (changes requested):

source/features/sticky-csv-header.css
fregante: That's explain why CI wasn't failing. 👌 

Feel free to "restore the file" once you accept the membership 😉

FloEdelmann: Both done. Thanks for inviting me :)

PR #6131, opened 2022-11-03
Fix `releases-tab` count

### LINKED ISSUE #6130: `releases-tab` count is wrong (includes tags)
### Description

We count the tags, but GitHub does not. See comparison between our tab and their count in the sidebar (where it's only visible after you disable the extension)
Screen Shot 1

Screen Shot 2

2. Look at `releases-tab`
3. Disable RGH
4. Refresh the page (look at the sidebar)

### Extension version

22.10.26

### Browser(s) used

Chrome

yakov116:
Fix #6130 

~I did not blame this yet. There must have been a reason we did it like this~

Assuming this did not exist in the past
## Test URLs

### DIFF
--- a/source/features/releases-tab.tsx
+++ b/source/features/releases-tab.tsx
@@ -28,7 +28,7 @@ async function parseCountFromDom(): Promise<number> {
 async function fetchFromApi(): Promise<number> {
 	const {repository} = await api.v4(`
 		repository() {
-			refs(refPrefix: "refs/tags/") {
+			releases {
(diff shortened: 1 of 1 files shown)


fregante (approved):

yakov116:
Looks like I missed the fetch from dom part.

That it's looking at the tags.

Can I know why we even are taking from the dom this is a tiny request and we do not do this anywhere else if I am not mistaken

PR #6133, opened 2022-11-03
Add `last-notification-page-button` feature

### LINKED ISSUE #6126: Way to go to "Last page" for notifications
### Description

When I have a lot of notifications for a repository, I often want to go to the _last_ page of the notification view, "Open all unread" there so I can go through them in the order they came in or were active (instead of starting with the _newest_ ones).

It would be amazing if Refined GitHub could add a `Last page` or `Last` next to "Prev | Next" below the list of notifications.

### Screenshot

image

### Example URLs

For me e.g. https://github.com/notifications?query=repo%3Aprisma%2Fprisma+is%3Aunread

yakov116:
Closes #6126

## Test URLs
https://github.com/notifications?query=is%3Adone

## Screenshot
image


source/features/linkify-notification-page-numbers.tsx
```diff
@@ -0,0 +1,32 @@
...
+void features.add(import.meta.url, {
+	asLongAs: [
+		() => select.exists('[data-hotkey="ArrowRight"]:not([disabled])'),
+	],
+	include: [
+		pageDetect.isNotifications,
+	],
+	exclude: [
+		pageDetect.isBlank, // Empty notification list
+	],
```
yakov116, line 29:
```suggestion
```
I don't think I need this right?

source/features/linkify-notification-page-numbers.tsx
```diff
@@ -0,0 +1,29 @@
...
+}
+
+void features.add(import.meta.url, {
+	asLongAs: [
+		() => select.exists('[data-hotkey="ArrowRight"]:not([disabled])'),
+	],
+	include: [
+		pageDetect.isNotifications,
+	],
+	deduplicate: 'has-rgh',
```
fregante, line 27: This is deprecated. Every feature needs its own deduplication logic in the init itself, perhaps via observe-selector or attachElement

readme.md
```diff
@@ -366,6 +366,7 @@ Thanks for contributing! 🦋🙌
 - [](# "select-all-notifications-shortcut") Adds a shortcut to select all visible notifications: <kbd>a</kbd>.
 - [](# "linkify-notification-repository-header") [Linkifies the header of each notification group (when grouped by repository).](https://user-images.githubusercontent.com/1402241/80849887-81531c00-8c19-11ea-8777-7294ce318630.png)
 - [](# "stop-redirecting-in-notification-bar") [Stops redirecting to notification inbox from notification bar actions while holding <kbd>Alt</kbd>.](https://user-images.githubusercontent.com/202916/80318782-c38cef80-880c-11ea-9226-72c585f42a51.png)
+- [](# "linkify-notification-page-numbers") [Linkifies the page number of notifications to jump to the last page.](https://user-images.githubusercontent.com/16872793/199788308-0f4d8fa2-9a80-4049-9f43-60a9f79c3b80.png)
```
fregante, line 369: The feature is "linking to the last page;" linkification is just how we achieve that. 

`last-notification-page-button` is more helpful, but long

fregante (changes requested):

source/features/linkify-notification-page-numbers.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+function init(): void {
+	const nextButtonURL = new URL(select('a[data-hotkey="ArrowRight"]')!.href);
+	const nextButtonSearch = new URLSearchParams(nextButtonURL.search);
+	const lastNotificationPage = select('.js-notifications-list-paginator-counts')!.lastChild!;
+	const lastNotificationPageNumber = looseParseInt(lastNotificationPage);
+	const lastCursor = Math.floor(lastNotificationPageNumber / 50) * 50;
+	nextButtonSearch.set('after', btoa('cursor:' + String(lastCursor)));
+	nextButtonURL.search = nextButtonSearch.toString();
+	wrap(lastNotificationPage, <a href={nextButtonURL.href}/>);
```
fregante, line 17: Wrapping the whole node with "of" and whitespace looks ugly, only the number should be linkified. Also the underling should always be visible

yakov116: >Also the underling should always be visible
?

fregante: Is the underline always visible or only on hover. 

It should always be visible. I don't know how else to phrase it.

yakov116: Done I meant to write `underling ?` but look like I forgot that word.

Changed

source/features/last-notification-page-button.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+function init(): void {
+	const nextButtonURL = new URL(select('a[data-hotkey="ArrowRight"]')!.href);
+	const nextButtonSearch = new URLSearchParams(nextButtonURL.search);
+	const lastNotificationPageNode = select('.js-notifications-list-paginator-counts')!.lastChild!;
+	const lastNotificationPageNumber = looseParseInt(lastNotificationPageNode);
+	const lastCursor = Math.floor(lastNotificationPageNumber / 50) * 50;
+	nextButtonSearch.set('after', btoa('cursor:' + String(lastCursor)));
+	nextButtonURL.search = nextButtonSearch.toString();
+	lastNotificationPageNode.replaceWith(<> of <a style={{textDecoration: 'underline'}} className="rgh-last-notification-page-button" href={nextButtonURL.href}>{lastNotificationPageNumber}</a></>);
```
fregante, line 16: I think this API accepts multiple parameters so you can wrap this properly, without `<>`

```js
replaceWith(
  " of ",
  <a
    className
    etc
  >
    {number}
  </a>

File: source/features/last-notification-page-button.tsx [hunk 6]
```diff
@@ -0,0 +1,30 @@
...
+}
+
+void features.add(import.meta.url, {
+	asLongAs: [
+		() => select.exists('[data-hotkey="ArrowRight"]:not([disabled])'),
```
fregante, line 29: This same query is in init already, let's just use `return false` in init. When they button is disabled, it's not a `a`, so select is already returning undefined there.

source/features/last-notification-page-button.tsx
```diff
@@ -0,0 +1,38 @@
...
+import looseParseInt from '../helpers/loose-parse-int';
+
+function init(): void {
+	const nextButtonURL = new URL(select('a[data-hotkey="ArrowRight"]')!.href);
+	const nextButtonSearch = new URLSearchParams(nextButtonURL.search);
+	const lastNotificationPageNode = select('.js-notifications-list-paginator-counts')!.lastChild!;
+	const lastNotificationPageNumber = looseParseInt(lastNotificationPageNode);
+	const lastCursor = Math.floor(lastNotificationPageNumber / 50) * 50;
+	nextButtonSearch.set('after', btoa('cursor:' + String(lastCursor)));
+	nextButtonURL.search = nextButtonSearch.toString();
```
fregante, line 15: I think you can set the raw search string directly onto the anchor, without using nextButtonURL at all ever:
```js
const nextButtonSearch = new URLSearchParams(nextButton.search)

// etc

<a href={String(nextButtonSearch)}>

File: source/features/last-notification-page-button.tsx [hunk 8]
```diff
@@ -0,0 +1,39 @@
...
+		return false;
+	}
+
+	const lastNotificationPageNode = select('.js-notifications-list-paginator-counts')!.lastChild!;
+	const lastNotificationPageNumber = looseParseInt(lastNotificationPageNode);
+	const lastCursor = Math.floor(lastNotificationPageNumber / 50) * 50;
+	const nextButtonSearch = new URLSearchParams(nextButton.search);
+	nextButtonSearch.set('after', btoa('cursor:' + String(lastCursor)));
+	nextButton.search = nextButtonSearch.toString();
+	lastNotificationPageNode.replaceWith(' of ',
```
fregante, line 20:
```suggestion
	lastNotificationPageNode.replaceWith(
		' of ',
```

source/features/last-notification-page-button.tsx
```diff
@@ -0,0 +1,39 @@
...
+	if (!nextButton) {
+		return false;
+	}
+
+	const lastNotificationPageNode = select('.js-notifications-list-paginator-counts')!.lastChild!;
+	const lastNotificationPageNumber = looseParseInt(lastNotificationPageNode);
+	const lastCursor = Math.floor(lastNotificationPageNumber / 50) * 50;
+	const nextButtonSearch = new URLSearchParams(nextButton.search);
+	nextButtonSearch.set('after', btoa('cursor:' + String(lastCursor)));
+	nextButton.search = nextButtonSearch.toString();
```
fregante, line 19: This changes the next button too. Please check my suggestions carefully, I wrote some [specific code](https://github.com/refined-github/refined-github/pull/6133#discussion_r1013518601) that isn't this. Let's not do 40-comments reviews again :(

```suggestion
```

source/features/last-notification-page-button.tsx
```diff
@@ -0,0 +1,39 @@
...
+	lastNotificationPageNode.replaceWith(' of ',
+		<a
+			style={{textDecoration: 'underline'}}
+			className="rgh-last-notification-page-button"
```
fregante, line 23: GitHub has a class for this:
```suggestion
			className="text-underline rgh-last-notification-page-button"
```

I thought there was a class specific to links but I didn't find it

source/features/last-notification-page-button.tsx
```diff
@@ -0,0 +1,39 @@
...
+import features from '../feature-manager';
+import looseParseInt from '../helpers/loose-parse-int';
+
+function init(): void | false {
+	const nextButton = select('a[data-hotkey="ArrowRight"]');
+	if (!nextButton) {
+		return false;
+	}
+
+	const lastNotificationPageNode = select('.js-notifications-list-paginator-counts')!.lastChild!;
```
fregante, line 14: Let's use a [content assertion](https://github.com/refined-github/refined-github/pull/6076) here, I think:

```suggestion
	const lastNotificationPageNode = select('.js-notifications-list-paginator-counts')!.lastChild!;
	assertNodeContent(lastNotificationPageNode, /^of \d+$/)
```

source/features/last-notification-page-button.tsx
yakov116: This will not work `{String(nextButtonSearch) => after=Y3Vyc29yOjEwNTA%3D&query=is%3Adone`

source/features/last-notification-page-button.tsx
yakov116: https://github.com/refined-github/refined-github/pull/6133#discussion_r1013878505

source/features/last-notification-page-button.tsx
```diff
@@ -0,0 +1,41 @@
...
+import {assertNodeContent} from '../helpers/dom-utils';
+
+function init(): void | false {
+	const nextButton = select('a[data-hotkey="ArrowRight"]')?.cloneNode();
+	if (!nextButton) {
+		return false;
+	}
+
+	const lastNotificationPageNode = select('.js-notifications-list-paginator-counts')!.lastChild!;
+	assertNodeContent(lastNotificationPageNode, ' of ');
```
fregante, line 16: Assertions are done after trimming

```suggestions
	assertNodeContent(lastNotificationPageNode, 'of');
```

source/features/last-notification-page-button.tsx
fregante: I fail to see how that would not work. `String(nextButtonSearch)` is exactly the same as `nextButtonSearch.toString()` and the search is exactly what we need.

yakov116: Now I understand what you are saying. 👍

yakov116: Your original suggestion was to remove the nextButtonURL, which I did.

If I don't include this line then the URL will only be the search, `https://github.com/after=Y3Vyc29yOjEwNTA%3D&query=is%3Adone` missing the notifications URL.

I feel bad we are going though so many comments.

If you feel that I am missing something, I have no issue with you closing and allowing someone else to pick this up. (No hard feelings I promise)

fregante: Gotcha, the only misunderstanding here is that `?` is missing when using `String()` whereas it exists in `url.search`. We just need to prepend the question mark so that the attribute will be exactly `"?after=Y3Vyc29yOjEwNTA%3D&query=is%3Adone"`, which is relative to the current pathname.

Possible ways, depending on which one TS likes the most

```js
 			href={'?' + String(nextButtonSearch)}
```
```js
 			href={`?${nextButtonSearch}`}
```

source/features/last-notification-page-button.tsx
```diff
@@ -0,0 +1,40 @@
...
+import {assertNodeContent} from '../helpers/dom-utils';
+
+function init(): void | false {
+	const nextButton = select('a[data-hotkey="ArrowRight"]');
+	if (!nextButton) {
+		return false;
+	}
+
+	const lastNotificationPageNode = select('.js-notifications-list-paginator-counts')!.lastChild!;
+	assertNodeContent(lastNotificationPageNode, new RegExp(/^of \d+$/));
```
fregante, line 16: Why `new RegExp()`

source/features/last-notification-page-button.tsx
```diff
@@ -0,0 +1,40 @@
...
+	const lastNotificationPageNumber = looseParseInt(lastNotificationPageNode);
+	const lastCursor = Math.floor(lastNotificationPageNumber / 50) * 50;
+	const nextButtonSearch = new URLSearchParams(nextButton.search);
+	nextButtonSearch.set('after', btoa('cursor:' + String(lastCursor)));
```
fregante, line 20:
```suggestion
	nextButtonSearch.set('after', btoa(`cursor:${lastCursor}`));
```

fregante (commented):
I just remembered this is a dynamic page so I think you'll need to use `observe` rather than a regular `select` I think. The listener function should be at the top level (outside `init`)

To trigger this, mark 10+ notifications as "Done" and wait for the page to update.

source/features/last-notification-page-button.tsx
```diff
@@ -0,0 +1,40 @@
+import React from 'dom-chef';
+import select from 'select-dom';
+import * as pageDetect from 'github-url-detection';
+
+import features from '../feature-manager';
+import looseParseInt from '../helpers/loose-parse-int';
+import {assertNodeContent} from '../helpers/dom-utils';
+
+function init(): void | false {
+	const nextButton = select('a[data-hotkey="ArrowRight"]');
```
fregante, line 10: This attribute is added by us https://github.com/refined-github/refined-github/blob/4939c6a2862f9ddcb138874efc627e9b1b8d97ff/source/features/pagination-hotkey.tsx#L24

fregante (changes requested):

janpio:
Thank you so much for implementing this @yakov116 (and for the reviews @fregante).

PR #6135, opened 2022-11-04
Fix `releases-tab` count /3

yakov116:

### DIFF
--- a/source/features/releases-tab.tsx
+++ b/source/features/releases-tab.tsx
@@ -12,12 +12,13 @@ import looseParseInt from '../helpers/loose-parse-int';
 import abbreviateNumber from '../helpers/abbreviate-number';
 import createDropdownItem from '../github-helpers/create-dropdown-item';
 import {buildRepoURL, getRepo} from '../github-helpers';
+import {releasesSidebarSelector} from './clean-repo-sidebar';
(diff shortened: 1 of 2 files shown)


source/features/releases-tab.tsx
```diff
@@ -17,7 +17,7 @@ import {appendBefore, highlightTab, unhighlightTab} from '../helpers/dom-utils';
 const getCacheKey = (): string => `releases-count:${getRepo()!.nameWithOwner}`;
 
 async function parseCountFromDom(): Promise<number> {
-	const moreReleasesCountElement = await elementReady('.file-navigation .octicon-tag + strong');
+	const moreReleasesCountElement = await elementReady('.Layout-sidebar .BorderGrid-cell h2 a[href$="/releases"] .Counter');
```
fregante, line 20: I don’t think it needs to be so specific

yakov116: I took it from here

https://github.com/refined-github/refined-github/blob/9b60bfe4049e4f02a75d800784b9ce8784723394/source/features/clean-repo-sidebar.tsx#L10

fregante: We can fix both

yakov116: Looks like we went though a few bugs with this selector already I will make a comment code https://github.com/refined-github/refined-github/pull/4428

PR #6166, opened 2022-11-17
Improve `select-notifications` style

fregante:
The button style feels more appropriate for actions in this case, rather than for selections. This is more clear when a notification is selected and you can tell the two types apart (select vs act)

## Test URLs

## Before

Screen Shot 2

## After

Screen Shot 3

### After, open

Screen Shot 4


FloEdelmann (approved):
Looks good and makes sense to me :ok_hand:

PR #6169, opened 2022-11-17
Restore `quick-file-edit` on new directory views

fregante:
- Part of https://github.com/refined-github/refined-github/issues/6152

## Test URLs

- https://github.com/refined-github/refined-github/tree/lint/safari

## Screenshot

gif23

### DIFF
--- a/source/features/quick-file-edit.tsx
+++ b/source/features/quick-file-edit.tsx
@@ -24,7 +24,6 @@ const cachedGetDefaultBranch = mem(getDefaultBranch, cachePerPage);
 
 async function linkifyIcon(fileIcon: Element): Promise<void> {
 	const isPermalink_ = await cachedIsPermalink();
-	debugger;
 	const fileLink = fileIcon.closest('.js-navigation-item, .react-directory-filename-column')!.querySelector('a.js-navigation-open, a.Link--primary')!;
 	const url = new GitHubURL(fileLink.href).assign({
 		route: 'edit',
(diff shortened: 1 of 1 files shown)


source/features/quick-file-edit.tsx
```diff
@@ -37,7 +39,10 @@ async function linkifyIcon(fileIcon: Element): Promise<void> {
 }
 
 function init(signal: AbortSignal): void {
-	observe('.js-navigation-container :not(a) > .octicon-file', linkifyIcon, {signal});
+	observe([
+		'.react-directory-filename-column svg.color-fg-muted',
+		'.js-navigation-container .octicon-file',
```
FloEdelmann, line 44: Can `:not(a)` be dropped or was it necessary to exclude symbolic links?

fregante: I dropped it because:

- `svg` is a direct descendant of `.react-directory-filename-column` (`X > SVG`), but `:not(a)` expects the `X > * > SVG` structure
- with the new observer we don't really need these deduplication mechanisms

fregante: This selector is not very specific though and I expect errors to crop up with rare icons (git modules, mirrors, etc) so it can be retouched later.

FloEdelmann: It's fine in the new UI. Still, you also changed the selector in the old UI.

fregante: Oops

PR #6182, opened 2022-12-04
Add `unhide-reactions-on-mobile` feature

jtagcat:
Closes https://github.com/refined-github/refined-github/issues/6161

## Test URLs
https://github.com/refined-github/refined-github/issues/6161

Firefox users have to set `layout.css.has-selector.enabled` to true in `about:config` (https://developer.mozilla.org/en-US/docs/Web/CSS/:has#browser_compatibility).

## Screenshot
demo

## Tested
On Google Play Store version of Firefox Nightly for Developers, version `109.0a1`.
With `npx web-ext run -t firefox-android --adb-device <use adb devices> --firefox-apk org.mozilla.fenix`

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -258,6 +258,7 @@ Thanks for contributing! 🦋🙌
 - [](# "no-duplicate-list-update-time") [Hides the update time of conversations in lists when it matches the open/closed/merged time.](https://user-images.githubusercontent.com/1402241/111357166-ac3a3900-864e-11eb-884a-d6d6da88f7e2.png)
 - [](# "linkify-user-labels") [Links the "Contributor" and "Member" labels on comments to the author’s commits on the repo.](https://user-images.githubusercontent.com/1402241/177033344-4d4eea63-e075-4096-b2d4-f4b879f1df31.png)
 - [](# "jump-to-conversation-close-event") [Adds a link to jump to the latest close event of a conversation.](https://user-images.githubusercontent.com/16872793/177792713-64219754-f8df-4629-a9ec-33259307cfe7.gif)
+- [](# "unhide-reactions-on-mobile") [Unhides reactions button on mobile.](https://user-images.githubusercontent.com/38327267/205493629-14563105-eaa1-4885-9f07-8b3ee9214c86.png)
 
 <!-- Refer to style guide above. Keep this message between sections. -->
 
--- a/source/features/unhide-reactions-on-mobile.css
+++ b/source/features/unhide-reactions-on-mobile.css
@@ -1,0 +1,3 @@
+.timeline-comment-actions > .d-none:has(.js-comment-header-reaction-button) {
+	display: inline-block !important;
+}
--- a/source/refined-github.ts
+++ b/source/refined-github.ts
@@ -25,6 +25,7 @@ import './features/sticky-file-header.css';
 import './features/readable-title-change-events.css';
 import './features/clean-checks-list.css';
 import './features/sticky-csv-header.css';
+import './features/unhide-reactions-on-mobile.css';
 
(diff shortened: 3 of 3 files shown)


readme.md
```diff
@@ -258,6 +258,8 @@ Thanks for contributing! 🦋🙌
 - [](# "no-duplicate-list-update-time") [Hides the update time of conversations in lists when it matches the open/closed/merged time.](https://user-images.githubusercontent.com/1402241/111357166-ac3a3900-864e-11eb-884a-d6d6da88f7e2.png)
 - [](# "linkify-user-labels") [Links the "Contributor" and "Member" labels on comments to the author’s commits on the repo.](https://user-images.githubusercontent.com/1402241/177033344-4d4eea63-e075-4096-b2d4-f4b879f1df31.png)
 - [](# "jump-to-conversation-close-event") [Adds a link to jump to the latest close event of a conversation.](https://user-images.githubusercontent.com/16872793/177792713-64219754-f8df-4629-a9ec-33259307cfe7.gif)
+- [](# "unhide-reactions-on-mobile") [Unhides reactions button on mobile.](https://user-images.githubusercontent.com/38327267/205493629-14563105-eaa1-4885-9f07-8b3ee9214c86.png)
+  - Firefox requires `about:config` → `layout.css.has-selector.enabled`: `true`
```
fregante, line 262: No need here. We can mention it in the wiki and/or release

source/features/unhide-reactions-on-mobile.css
```diff
@@ -0,0 +1,4 @@
+/* :has() does not work on Firefox, unless `about:config` → `layout.css.has-selector.enabled`: `true` */
+.d-none:has(.js-comment-header-reaction-button) {
```
fregante, line 2: This can match any large item that is temporarily hidden. My selector worker. Also the Firefox-related comment isn’t necessary

fregante (approved):
Thank you 💚

source/features/unhide-reactions-on-mobile.css
jtagcat: > This can match any large item that is temporarily hidden. My selector worker.

Not sure what you mean by that. This is `.d-none:has(.js-comment-header-reaction-button)`? What is any large item, and your selector worker?

jtagcat: Yep, my mistake. I had it in there initially, but didn't get any results (was using stock Firefox). I tried reducing it to bare minimum components and combining them step by step.

fregante: Sorry it was as a typo. I updated the comment. 

Your selector can affect even the outer element, making it visible when it shouldn't:

``` html
<div class="d-none comment">
  This whole comment is not meant to be visible, ever

  <div class="d-none">
    <details class="js-comment-header-reaction-button">
  </div>
</div>

fregante (changes requested):
> I added a `h4 { color red; }` as an initial test. With `isIssueOrPRList`, it should not affect the login page..?

I just remembered why. All CSS is always imported. For `addCssFeature` to work, you have to use `.rgh-*` in every selector. Refer to other features that use `addCssFeature`

But in this case we don't need this at all, it's a small CSS bugfix that does not need to be "disableable".

Changes required:

- [x] Delete .tsx file
- [ ] Import .css in refined-github.ts and move it to the right place
- [ ] Probably revert the changes to the readme as well if CI fails

fregante (approved):

jtagcat:
Can confirm it still working on a mobile device.

fregante:
Love your screenshot by the way.

Are there other features hidden on mobile that you find useful? I think I always need something else but never open an issue when I do 😳

Issue #6192, opened 2022-12-12
`hide-own-stars` is a bit broken


fregante:
### Native 

<img width="661" alt="Screen Shot 1" src="https://user-images.githubusercontent.com/1402241/206991140-e0a82422-fa72-4562-b77a-939673386bf9.png">

### RG

The single events are hidden, but the group isn't

![gif23](https://user-images.githubusercontent.com/1402241/206991204-ccc489df-01c2-49db-a0d1-3331dac650d4.gif)

### How to replicate the issue + URL

1. https://github.com
2. Look for people starring your own repos

### Extension version

22.11.21

### Browser(s) used

Safari, Chrome

### Related

- #6188

Issue #6188, opened 2022-12-09
`clean-dashboard` is a little broken


fregante:
### Description

<img width="957" alt="Screen Shot" src="https://user-images.githubusercontent.com/1402241/206649882-16d57586-6c7f-4eb7-aa0a-8b7240c38a68.png">

### How to replicate the issue + URL

https://github.com/

### Extension version

22.11.21

### Browser(s) used

Safari

cooljeanius:
I only just started using this and was wondering about the changes to the news feed; how is it *supposed* to look? i.e., which changes made to the news feed by this extension are intentional, vs. which are examples of this bug?

fregante:
> how is it _supposed_ to look?

See https://github.com/refined-github/refined-github/pull/6136

PR #6194, opened 2022-12-12
Fix `hide-newsfeed-noise`

fregante:
## Test URLs

https://github.com/

## Before

Screen Shot 5

## After
Screen Shot 6

## Related

- #6192 
- #6188

### DIFF
--- a/source/features/highest-rated-comment.tsx
+++ b/source/features/highest-rated-comment.tsx
@@ -116,7 +116,6 @@ void features.add(import.meta.url, {
 	init,
 });
 
-
 /*
 Test URLs:
 


source/features/hide-newsfeed-noise.css
```diff
@@ -9,15 +9,14 @@ published a release
...
-.rgh-hide-newsfeed-noise .news :is(
-[classes~='push'],
-[classes~='fork'],
-[classes~='delete'],
-[classes~='follow'],
-[classes~='release'],
-[classes~='gollum']
-),
+/* TODO: Drop old classes, no longer used */
 .rgh-hide-newsfeed-noise .news :is(.push, .fork, .delete, .follow, .release, .gollum) {
```
fregante, line 13: It seems that some classes like `watch_started` and `follow` are still used, but I don't have other events.

PR #6230, opened 2023-01-17
Fix pagination in `last-notification-page-buttons`

### LINKED ISSUE #6218: `last-notification-page-button`: Link to last page is sometimes broken or wrong
### Description

There seems to be a off by one problem going on somehow when the # of issues is 150 (and hits exactly when pagination would happen):
image
=>
image
The "Prev" button on that page unfortunately also does not work (but that is a GH bug).

Oh, even with 149 something fishy is going on:
image
=>
image
Here clicking "Next" brings me to the _actual_ last page then though.

(cc @yakov116)

https://github.com/notifications?query=repo%3Aprisma%2Fprisma+is%3Aunread

### Extension version

22.12.15

### Browser(s) used

Chrome

janpio:
fixes #6218

Tested via `npx web-ext run --target=chromium` and https://github.com/notifications?query=repo%3Aprisma%2Fprisma+is%3Aunread where the problem manifested for me.

### DIFF
--- a/source/features/last-notification-page-button.tsx
+++ b/source/features/last-notification-page-button.tsx
@@ -7,7 +7,7 @@ import looseParseInt from '../helpers/loose-parse-int';
 import {assertNodeContent} from '../helpers/dom-utils';
 import observe from '../helpers/selector-observer';
 
-let itemsPerNotificationsPage = 25;
+const itemsPerNotificationsPage = 25;
 
 function linkify(nextButton: HTMLAnchorElement): void {
 	const lastNotificationPageNode = select('.js-notifications-list-paginator-counts')!.lastChild!;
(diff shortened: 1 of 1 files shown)


fregante (approved):
Congratulations on your first PR! There's only one lint issue

source/features/last-notification-page-button.tsx
```diff
@@ -7,11 +7,13 @@ import looseParseInt from '../helpers/loose-parse-int';
 import {assertNodeContent} from '../helpers/dom-utils';
 import observe from '../helpers/selector-observer';
 
+let itemsPerNotificationsPage = 25;
```
janpio, line 10:
```suggestion
const itemsPerNotificationsPage = 25;
```

fregante:
I finally got around to fixing that. Thanks again

PR #6247, opened 2023-01-30
Drop `align-repository-header` feature

### LINKED ISSUE #6229: `align-repository-header` should be disabled on new file/folder views
### Description

Screenshot 1

https://github.com/refined-github/refined-github/blob/main/source/features/align-repository-header.css

### Extension version

22.11.21

### Browser(s) used

Safari

fregante:
- Closes #6229 

Why: 

- Maybe it's not such a bad idea to align it left
- The site header is already aligned left
- We still aligned left on full-width pages, so RG was making the header switch between the two positions on navigation
- It seems to be hard to target at the moment, I don't see the `full-width` class anymore
- Aligning it to the left means that the repo nav now has more space to show its tabs, which means _less nav overflow_

cc @szmarczak due to https://github.com/refined-github/refined-github/issues/3280

## Test URLs

See previous URLs on https://github.com/refined-github/refined-github/pull/3322


yakov116:
Wow I really miss this 😢

IceSentry:
I don't understand why this change was done. It _is_ a bad idea to align it to the left. What's wrong with things being in the center where the entire content is?

Is there any way to restore this feature? As someone with an ultrawide, having those headers all the way to the left is really annoying to use now.

image

fregante:
I also liked it before but… see https://github.com/refined-github/refined-github/issues/6229 for the reason why. Also I wrote a lot in the first comment here.

Issue #6152, opened 2022-11-10
Review rules around global search


fregante:
https://twitter.com/MaximeDeGreve/status/1590415616921698305

The UI has been updated and at least one of our features can now be dropped

waldyrious:
Commit message length checking ([suggest-commit-title-limit](https://github.com/refined-github/refined-github/blob/main/source/features/suggest-commit-title-limit.tsx)) also doesn't work anymore :(

fregante:
cc @refined-github/maintainers what features do you seem missing? Feels like half the extension is broken again 😅 

- [ ]  `quick-file-edit`

FloEdelmann:
- [ ] `hide-diff-signs` on https://github.com/refined-github/refined-github/edit/main/source/features/hide-diff-signs.css (after editing and clicking on preview)

fregante:
- [ ] `copy-on-y` on https://github.com/refined-github/refined-github/edit/main/source/features/hide-diff-signs.css
- [ ] `rgh-feature-descriptions` on https://github.com/refined-github/refined-github/edit/main/source/features/default-branch-button.tsx
- [ ] `default-branch-button` after visiting https://github.com/refined-github/refined-github/tree/lint/test and selecting a file (it should stay on the file, but it goes back to the /test folder)

kema-dev:
When editing a file, the top bar “Edit File / Preview” does not stick to the top of the screen any more. I tried to figure out which feature was responsible for this behaviour but I couldn't, it looks like it is enabled even when all features are disabled in options. :raised_eyebrow: I would happily give a hand if you can give me a little tip !

fregante:
So we can't even rely on the URL to always show the same UI. Here for example, if you visit https://github.com/refined-github/refined-github/tree/lint directly you see the old UI, but if you open a folder and then go back, you'll end up on the new UI

![ko](https://user-images.githubusercontent.com/1402241/215427932-047d75a8-8812-4081-8f95-c93e9947488e.gif)

fregante:
Please report(in new issues) any recently-broken features that have been annoying you every day. I might address them before the next version.

I know a lot is broken at the moment but often the fixes appear to be simple.

PR #6249, opened 2023-01-30
Restore `default-branch-button`

fregante:
- Part of #6152 

## Test URLs

Branch, subfolder: https://github.com/refined-github/refined-github/tree/better-ajax-detection/safari

Screenshot 14

Commit, home page: https://github.com/refined-github/refined-github/tree/2ce3317eba7d01d0c728605ad7730b4ee067835d

Screenshot 13

### DIFF
--- a/source/refined-github.css
+++ b/source/refined-github.css
@@ -192,7 +192,7 @@ div[data-target='readme-toc.content'] div.blob-wrapper img {
 	margin-right: 20px; /* Push the text farther away from the text so it's cropped out */
 }
 
-/* Add override for CSS-in-JSS components that don't use regular .btn styles */
+/* Add override for CSS-in-JS components that don't use regular .btn styles */
 .BtnGroup-item:last-child {
 	border-top-left-radius: 0;
 	border-bottom-left-radius: 0;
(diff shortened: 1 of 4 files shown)


fregante:
Also untouched on the main branch, both on the home and in a subfolder. They have completely different markup

Screenshot 15

Screenshot 16

fregante:
Note that in some views, GitHub may be display the same button twice. I'm not gonna address this now as I think they might _fix it_ eventually. 

Screenshot 22



Anyway the feature is applied to both places 🤷‍♂️ 

Screenshot 21

source/refined-github.css
```diff
@@ -191,3 +191,9 @@ div[data-target='readme-toc.content'] div.blob-wrapper img {
 [data-selected-links^='repo_milestones ']:has([title='0']) svg {
 	margin-right: 20px; /* Push the text farther away from the text so it's cropped out */
 }
+
+/* Add override for CSS-in-JSS components that don't use regular .btn styles */
```
FloEdelmann, line 195:
```suggestion
/* Add override for CSS-in-JS components that don't use regular .btn styles */
```

fregante:
```suggestion
/* Add override for CSS-in-JSSS components that don't use regular .btn styles */
```

https://en.wikipedia.org/wiki/JavaScript_Style_Sheets

FloEdelmann:
```suggestion
/* Add override for CS-in-JSS components that don't use regular .btn styles */
```

https://en.wikipedia.org/wiki/Computer_science

fregante:
```suggestion
/* Add override for JAXA-in-ISS components that don't use regular .btn styles */
```

https://en.wikipedia.org/wiki/International_Space_Station

PR #6252, opened 2023-01-30
Restore `copy-on-y` and enable it on every page

fregante:
- Fixes https://github.com/refined-github/refined-github/issues/6210
- Part of #6152

Over the years I've come to expect "copy on y" on any page so I think it wouldn't be too bad to actually enable it site-wide.

Thoughts?

download

￼

## Test URLs

Literally any page, including:

https://github.com/refined-github/refined-github/blob/main/source/features/hide-diff-signs.css

### DIFF
--- a/source/features/copy-on-y.tsx
+++ b/source/features/copy-on-y.tsx
@@ -1,29 +1,18 @@
-import select from 'select-dom';
-import * as pageDetect from 'github-url-detection';
-
 import features from '../feature-manager';
 import {isEditable} from '../helpers/dom-utils';
 
-const handler = ({key, target}: KeyboardEvent): void => {
+async function handler({key, target}: KeyboardEvent): Promise<void> {
 	if (key === 'y' && !isEditable(target)) {
-		const permalink = select('a.js-permalink-shortcut')!.href;
-		void navigator.clipboard.writeText(permalink + location.hash);
+		await navigator.clipboard.writeText(location.href);
+		console.log('Copied URL to the clipboard', location.href);
 	}
-};
+}
 
 function init(signal: AbortSignal): void {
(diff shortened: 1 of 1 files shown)


source/features/copy-on-y.tsx
```diff
@@ -1,29 +1,20 @@
...
 
-const handler = ({key, target}: KeyboardEvent): void => {
+async function handler({key, target}: KeyboardEvent): Promise<void> {
 	if (key === 'y' && !isEditable(target)) {
-		const permalink = select('a.js-permalink-shortcut')!.href;
-		void navigator.clipboard.writeText(permalink + location.hash);
+		const url = location.href;
+		await navigator.clipboard.writeText(url);
+		// Log to ensure we're coping the new URL
+		console.log('Copied URL to the clipboard', url);
```
yakov116, line 9: Your keeping this here?

yakov116 (approved):
We added the filters at a time when we were trying to limit the features not to run on every page.

source/features/copy-on-y.tsx
fregante: Yes for now:

- could be helpful if suddenly this starts copying the non-permalinked URL
- we're not giving any feedback that this feature is doing anything
- the log only appears in rare situations, i.e. when the user does tap Y

yakov116: Why not use the logger function?

fregante: Laziness to check how that worked 😂 

Will do 🚀

fregante: We don't seem to use that anywhere except "show features that are run on the page", it doesn't seem strictly related to this type of logging. Given this is user feedback, I'd still keep it here unconditionally

PR #6254, opened 2023-01-30
Update links in `sticky-file-header`

fregante:
This updates the links following a report in https://github.com/refined-github/refined-github/issues/6152#issuecomment-1376541166

Everything seems to still work, with the exception that GitHub now does this natively almost everywhere.

@kema-dev where do you see the issue? Please include URLs, your browser and the version you're using

### DIFF
--- a/source/features/hide-diff-signs.css
+++ b/source/features/hide-diff-signs.css
@@ -2,13 +2,3 @@
 .rgh-hide-diff-signs .blob-code-marker-cell::before { /* Commit page */
 	visibility: hidden;
 }
-
-/*
-
-File URLs
-
-Any commits on https://github.com/sindresorhus/refined-github/commits/main
-Any reviewed block on https://github.com/sindresorhus/refined-github/pull/1783
-Full diff on https://github.com/sindresorhus/refined-github/pull/1783/files
-
-*/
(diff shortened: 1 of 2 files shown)


source/features/hide-diff-signs.css
```diff
@@ -2,3 +2,13 @@
...
+
+/*
+
+File URLs
+
+Any commits on https://github.com/sindresorhus/refined-github/commits/main
+Any reviewed block on https://github.com/sindresorhus/refined-github/pull/1783
+Full diff on https://github.com/sindresorhus/refined-github/pull/1783/files
+
+*/
```
fregante, line 14: Same goes for `hide-diff-signs` cc @FloEdelmann, hidden everywhere. 
Screenshot 27

Screenshot 29


Screenshot 28

fregante: @FloEdelmann never mind, I reread your report. Will fix separately.

FloEdelmann: 1. Go to https://github.com/refined-github/refined-github/edit/main/package.json
2. edit anything
3. click on preview

Diff signs are not hidden there:

grafik

fregante: See https://github.com/refined-github/refined-github/pull/6255

fregante:
Indeed that’s the only page where it doesn’t work, but given the very-broken double-scrollbar they currently have, I don’t want to mess with it yet. Let’s try again in a few months, but ideally you can report it to GitHub.

PR #6262, opened 2023-01-30
Restore `download-folder-button`

fregante:
- Finally fixes https://github.com/refined-github/refined-github/issues/6154

## Test URLs

- Own repo: https://github.com/refined-github/refined-github/tree/main/.github
- Archived repo: https://github.com/fregante/object-fit-images/tree/master/demo

## Screenshot

Seen in both desktop and mobile resolutions.

Screenshot 4

Screenshot 5

### DIFF
--- a/source/features/download-folder-button.tsx
+++ b/source/features/download-folder-button.tsx
@@ -1,42 +1,28 @@
 import React from 'dom-chef';
-import select from 'select-dom';
 import {DownloadIcon} from '@primer/octicons-react';
 import * as pageDetect from 'github-url-detection';
 
 import features from '../feature-manager';
-import {getRepo} from '../github-helpers';
+import observe from '../helpers/selector-observer';
 
-function getDropdownItem(downloadUrl: URL): JSX.Element {
-	return (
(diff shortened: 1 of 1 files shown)


source/features/download-folder-button.tsx
```diff
@@ -1,36 +1,18 @@
 import React from 'dom-chef';
-import select from 'select-dom';
 import {DownloadIcon} from '@primer/octicons-react';
 import * as pageDetect from 'github-url-detection';
 
 import features from '../feature-manager';
-import {getRepo} from '../github-helpers';
+import observe from '../helpers/selector-observer';
 
-function getDropdownItem(downloadUrl: URL): JSX.Element {
```
fregante, line 9: This is no longer practical. There are no icons, classes, titles, aria-label, nothing in the dropdown. It's completely unselectable. This is the kind of issues we'll find more and more as GitHub switches more of its interface to React and CSS-in-JS random classes. 

Here the solution is to just move the button back out of the dropdown. 

Simplification of the features is key.

PR #6267, opened 2023-01-31
Reduce content shift on navigation

### LINKED ISSUE #5762: FOUC after Ajax navigation
1. Visit https://github.com/refined-github/refined-github/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc
2. Click "Closed" filter
2. Notice features appearing a fraction of a second later

before

In some cases this is unavoidable, but for these instant features this wasn't the case before. AJAX navigation loads the whole document at once and then immediately calls the "turbo:render" event, which lets us swoop in and make the changes before it's shown to the user.

GitHub recently changed the Ajax library so this might be causing issues. We're already using the earliest event for this so I don't know what we can do, exploration is necessary:

- https://turbo.hotwired.dev

Here's the sequence again:

Screen Shot

Generated via

```js
['turbo:before-cache',
'turbo:before-fetch-request',
'turbo:before-fetch-response',
'turbo:before-render',
'turbo:before-stream-render',
'turbo:before-visit',
'turbo:click',
'turbo:frame-load',
'turbo:frame-render',
'turbo:load',
'turbo:reload',
'turbo:render',
'turbo:submit-end',
'turbo:submit-start',
'turbo:visit',
'pjax:start',
'pjax:end',
'pjax:complete',
'pjax:error',].map(x => document.addEventListener(x, e => console.log(x)))
```

## Notes

- `turbo:before-render` appears to be too early and some features won't work. 
- do we have to use `before-render` and pass around the new `body` before it appears? https://turbo.hotwired.dev/handbook/drive#pausing-rendering

## Related

- https://github.com/refined-github/refined-github/issues/5719
- https://github.com/refined-github/refined-github/issues/5751
- https://github.com/refined-github/refined-github/issues?q=+label%3Aajax

### LINKED ISSUE #6001: `clean-dashboard` lost after going back
### Description

https://user-images.githubusercontent.com/1402241/191226237-1d27c940-e882-46cf-98e0-8a0cebc5a986.mov

1. Open https://github.com/
2. Visit any link on the page
3. Click the browser’s Back button

### Extension version

22.9.9

### Browser(s) used

Safari

### Related

- #5895

fregante:
Probably:

- Fixes #6001 
- Fixes #5762

## Demo

https://user-images.githubusercontent.com/1402241/215728496-84f7bba7-ae29-445c-93ac-3302838b6a51.mov

### DIFF
--- a/source/feature-manager.tsx
+++ b/source/feature-manager.tsx
@@ -301,7 +301,7 @@ const unload = (featureUrl: string): void => {
 	}
 };
 
-document.addEventListener('turbo:visit', () => {
+document.addEventListener('turbo:render', () => {
 	for (const feature of currentFeatureControllers.values()) {
 		for (const controller of feature) {
 			controller.abort();


source/feature-manager.tsx
```diff
@@ -301,7 +301,7 @@ const unload = (featureUrl: string): void => {
 	}
 };
 
-document.addEventListener('turbo:visit', () => {
+document.addEventListener('turbo:render', () => {
```
fregante, line 304: This is the same event used to run features and it's currently attached/run _before_ our "on load" listeners. I think this is helping avoid any flashes due to momentarily-missing classes. For example if a feature removes a class on unload, the the "on load" listener will restore it immediately after, without a render/display of the page.

PR #6286, opened 2023-02-02
Update `webext-domain-permission-toggle`

fregante:
- Includes https://github.com/fregante/webext-domain-permission-toggle/pull/25
- Required for https://github.com/refined-github/refined-github/issues/5531

I don’t think anything changed, but it's worth a run here first


fregante:
The item isn't updating early enough. I'm not sure if this is new behavior

fregante:
Nope. I need to revisit the module, the behavior exists in previous versions too

PR #6296, opened 2023-02-04
Drop `file-finder-buffer` feature

fregante:
Another one bites the dust

ko

Follow-up:

- https://github.com/refined-github/refined-github/issues/6142


fregante:
There's a change that the bug was caused by an earlier PR, but the feature was already on life support (and broken on Firefox and Safari)

- https://github.com/refined-github/refined-github/pull/6267
- https://github.com/refined-github/refined-github/issues/6108
- https://github.com/refined-github/refined-github/issues/3846

Issue #6293, opened 2023-02-04
`release-tab` is added too early, out of order, sometimes


fregante:
### Description

<img width="645" alt="Screenshot 1" src="https://user-images.githubusercontent.com/1402241/216769596-da0bb3ec-72d9-4f21-960e-fd2b3af58cc0.png">

### How to replicate the issue + URL

It happens occasionally, but often enough

https://github.com/refined-github/refined-github

### Extension version

23.2.1

### Browser(s) used

Chrome

innocenzi:
It also sometimes doesn't appear at all and I can't figure out why. 

Sometimes it's because the repository owner unchecked "releases" to the "show on the home page" setting, but sometimes the tab is not there even though releases exist.

Maybe these issues could be fixed together

fregante:
This is strange, we do have a specific waiter for the whole bar:

https://github.com/refined-github/refined-github/blob/60b7aa240d0902ed34904129a0b83631f88870f6/source/features/releases-tab.tsx#L62-L63

PR #6298, opened 2023-02-04
Improve `releases-tab` reliability

innocenzi:
This pull request fixes an issue with `releases-tab` where the tab would not be displayed if Refined GitHub could not find the release count from the DOM, which would be the case if a user disabled "releases" from the "show in the home page" setting.

CleanShot 2023-02-04 at 15 43 35

Example repositories:
- https://github.com/hybridly/hybridly
- https://github.com/vueuse/motion

Note: this change doesn't affect the fact that the release tab won't be displayed if there's no release at all.

### DIFF
--- a/source/features/releases-tab.tsx
+++ b/source/features/releases-tab.tsx
@@ -38,7 +38,7 @@ async function fetchFromApi(): Promise<number> {
 	return repository.releases.totalCount;
 }
 
-export const getReleaseCount = cache.function(async () => pageDetect.isRepoRoot() ? parseCountFromDom() : fetchFromApi(), {
+export const getReleaseCount = cache.function(async () => await parseCountFromDom() || await fetchFromApi(), {
 	maxAge: {hours: 1},
 	staleWhileRevalidate: {days: 3},
 	cacheKey: getCacheKey,
--- a/source/features/releases-tab.tsx
+++ b/source/features/releases-tab.tsx
@@ -42,7 +42,7 @@ async function fetchFromApi(): Promise<number> {
 // - It is disabled by repository owner on the home page (release DOM element won't be there)
 // - It only contains pre-releases (count badge won't be shown)
 // For this reason, if we can't find a count from the DOM, we ask the API instead (see #6298)
-export const getReleaseCount = cache.function(async () => await parseCountFromDom() || await fetchFromApi(), {
+export const getReleaseCount = cache.function(async () => await parseCountFromDom() || fetchFromApi(), {
 	maxAge: {hours: 1},
 	staleWhileRevalidate: {days: 3},
 	cacheKey: getCacheKey,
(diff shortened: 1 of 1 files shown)


fregante:
We can't fix it this way, if it doesn't exist then the code is wrong. 

The problem here is that we're using `awaitDomReady` false and then using `select` everywhere, which nothing is found because `init` is run before the page exists. 

The select calls need to be replaced by elementReady, but please don't just search and replace without understanding the logic in the file

fregante:
By the way this issue isn't strictly related to https://github.com/refined-github/refined-github/issues/6293, but using `elementReady` correctly would definitely fix that issue as well.

innocenzi:
I found out a common point between these two repositories: they don't have a `.Counter` because all their releases are pre-releases, so GitHub doesn't count them there.

| Normal releases | Pre-releases only |
|-|-|
| CleanShot 2023-02-04 at 17 52 13 | CleanShot 2023-02-04 at 17 53 08 |

The current code works fine with repositories with normal releases, except when "releases" is unchecked — the "isRepoRoot" URL check assumes it will be. That's why I don't see a better way than simply asking the API if we can't find anything in the DOM. What do you think?

innocenzi:
Let me know if the change I just made would fix the issue. I couldn't reproduce #6293 myself.

source/features/releases-tab.tsx
```diff
@@ -80,8 +80,13 @@ async function addReleasesTab(): Promise<false | void> {
 	// Trigger a reflow to push the right-most tab into the overflow dropdown (second attempt #4254)
 	window.dispatchEvent(new Event('resize'));
 
+	const dropdownMenu = await elementReady('.js-responsive-underlinenav .dropdown-menu ul', {waitForChildren: false})
```
fregante, line 83: We do need to wait for children or else "append" won't work

source/features/releases-tab.tsx
```diff
@@ -80,8 +80,13 @@ async function addReleasesTab(): Promise<false | void> {
...
+	if (!dropdownMenu) {
+		return
```
fregante, line 85: We expect the dropdown to exist; if it doesn't, we want errors to tell us we should update the code.

You can use `dropdownMenu!` inside `appendBefore` to silence the typescript issue

fregante (approved):
That makes perfect sense. So yes this would be a good fix in this case. #6293 would need a different fix and would also cover other race conditions.

Can you fix the lint and add a comment to justify the `||` in `getReleaseCount`?

source/features/releases-tab.tsx
innocenzi: So I remove the guard and let the underlying `parent.append` trigger an error right?

fregante (approved):

PR #6301, opened 2023-02-05
Add automated selector testing

### LINKED ISSUE #6060: Set up automated live testing of some CSS selectors
_Originally posted in https://github.com/refined-github/refined-github/issues/6023#issuecomment-1270439879_
      
---

> > set up some sort of automated live testing of CSS selectors
> 
> Should not be too difficult, cant make a selector check on load of the feature?

1. Given the complex logic, we need to test the whole selector for the tests to be useful
2. Testing the selectors "live" means we have to import them into the JS side and run them at some point
3. Given the complex logic, these selectors will not appear on every page, so we can't actually _assert_ that a selector _must_ match an element on any given page the user loads.

The proper testing harness I'm thinking of would look like:

1. In node
1. Loads `no-unnecessary-split-diff-view.css`
2. Extracts the selectors and any related URLs, to be defined like:
	```css
	/** @expect-on http://github.com/some/dead-repo*/
	.repo-bar .tab:has(.counter[label=0]) {
		display:none;
	}
	```
4. Fetches the URL and looks for the selector on the page

fregante:
- Closes #6060

## Next

- [ ] Add more selectors, at least 10 to begin
- [ ] CSS parsing?
	- Thought of Lightning CSS, but no comments support https://github.com/parcel-bundler/lightningcss/issues/366

and

- [ ] CI: Exclude tests unless the selectors file has been changed?
	- Maybe only if duration becomes a problem
	- "Vitest watch" already does this natively locally 😍


christian-bromann:
@fregante I recommend use WebdriverIO for this task. It can load the actual page in a (headless) browser and do all necessary assertions. Happy to take a stab at this if you like.

fregante:
Thanks! It would be interesting to see how much code is necessary to test one of our features, compared to just adding URLs as suggested here. 

You could test that a "Bugs" tab is being added to the URL https://github.com/refined-github/sandbox

"Cleanup" features are probably not testable with this method since the test `.selector doesn't exist` would pass even if the `.selector` element never existed.

kidonng:
[happy-dom](https://github.com/capricorn86/happy-dom) should be a little faster than JSDOM (I've used it for scraping).

I think we only need headless browser if we want to test logged-in only features.

fregante:
Merge now, extend later 🚀

PR #6311, opened 2023-02-08
Restore `clean-header-search-field` for some users

fregante:
- Fixes https://github.com/refined-github/refined-github/issues/6310

## Test URLs

https://github.com/fregante/GhostText/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc

## Screenshot

Untested, depends on the user

cc @kaste

### DIFF
--- a/source/features/clean-header-search-field.tsx
+++ b/source/features/clean-header-search-field.tsx
@@ -5,7 +5,14 @@ import select from 'select-dom';
 import features from '../feature-manager';
 
 async function init(): Promise<void> {
-	const headerSearchButton = await elementReady('button.header-search-button');
+	const headerSearchButton = await elementReady('button.header-search-button, input.header-search-input');
+	if (headerSearchButton instanceof HTMLInputElement) {
+		// Previous version #6310
+		// TODO: Remove in mid 2023
+		headerSearchButton.value = '';
+		return;
+	}
+
 	headerSearchButton!.classList.add('placeholder');
 	select('[data-target="search-input.inputButtonText"]', headerSearchButton)!.textContent = headerSearchButton!.getAttribute('placeholder');
 }


source/features/clean-header-search-field.tsx
```diff
@@ -5,7 +5,14 @@ import select from 'select-dom';
 import features from '../feature-manager';
 
 async function init(): Promise<void> {
-	const headerSearchButton = await elementReady('button.header-search-button');
+	const headerSearchButton = await elementReady('button.header-search-button, input.header-search-input');
+	if (headerSearchButton instanceof HTMLInputElement) {
+		// Previous version #6310
+		// TODO: Remove in mid 2023
```
yakov116, line 11: Maybe put a real date so it shows up as a warning

fregante: I know we've been doing that, but what does that change in practice? We don't want CI to start failing and force us to immediately take action 🤔

auscompgeek:
> Untested, depends on the user

It should be toggleable in the feature preview settings:

image

source/features/clean-header-search-field.tsx
yakov116: If I am not mistaken we set it to warn

fregante: We did, but due to: 

- https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1540

I think `expiring-todo-comments` only makes sense:

- if it never breaks (on a date) and forces anyone to act immediately
- if someone is actively looking for and fixing TODO comments 
- if there are no other TODOs that are _also_ waiting for a fix anyway, for months

Essentially I think we can do away with them. Moving this discussion upstream in https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1541

PR #6324, opened 2023-02-10
Make `create-release-shortcut` more resilient

fregante:
It seems that all shortcuts (even native ones) are broken at the moment 🤷‍♂️  but it should fix the first part of:

- https://github.com/refined-github/refined-github/issues/5884

## Test URLs

https://github.com/refined-github/sandbox/releases/new
https://github.com/refined-github/sandbox/releases/tag/cool

## Screenshot

Unable to test since it's natively broken


fregante:
`c` is already in use, dropping the feature instead:

Screenshot 5
Screenshot 4


I'm not sure why this feature was added since creating releases is such a relatively rare action 🤷‍♂️

source/features/submission-via-ctrl-enter-everywhere.tsx
```diff
@@ -3,7 +3,7 @@ import * as pageDetect from 'github-url-detection';
 
 import features from '../feature-manager';
 
-export default function addQuickSubmit(): void {
+function addQuickSubmit(): void {
```
fregante, line 6: The fun part was that this part of the feature was being exported from `submission-via-ctrl-enter-everywhere`, duplicating it entirely

PR #6341, opened 2023-02-13
Use `batched-function` module

Richienb:
https://github.com/sindresorhus/project-ideas/issues/136#issuecomment-1416652879

https://github.com/Richienb/batched-function


fregante (approved):
Thanks! There are a couple more places where this could be used, would you like to apply it there? One is

https://github.com/refined-github/refined-github/blob/af339717863d8b2ef11080db1b7c4db3b89254e2/source/features/release-download-count.tsx#L67

PR #6345, opened 2023-02-15
Fix overlap in `rgh-feature-descriptions`

### LINKED ISSUE #6337: Feature description overlaps first line of code on new file viewer
### Description

image

Happens on any [feature source code](https://github.com/refined-github/refined-github/tree/main/source/features) file. For example: https://github.com/refined-github/refined-github/blob/main/source/features/action-used-by-link.tsx

This does **not** happen when source file is not an feature with an description. For example: https://github.com/refined-github/refined-github/blob/main/build/prepare-safari-release.sh

### Extension version

23.2.11

### Browser(s) used

Firefox

amanmalh:
Fixes #6337 

## Test URLs
https://github.com/refined-github/refined-github/blob/main/source/features/action-used-by-link.tsx

## Screenshot
feature-description-overlap

### DIFF
--- a/source/features/rgh-feature-descriptions.tsx
+++ b/source/features/rgh-feature-descriptions.tsx
@@ -22,6 +22,7 @@ async function add(infoBanner: HTMLElement): Promise<void> {
 	conversationsUrl.searchParams.set('q', `sort:updated-desc "${feature.id}"`);
 
 	infoBanner.before(
+		{/* Block and width classes required to avoid margin collapse */}
 		<div className="Box mb-3 d-inline-block width-full">
 			<div className="Box-row d-flex gap-3 flex-wrap">
 				<div className="rgh-feature-description">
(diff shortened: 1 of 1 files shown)


source/features/rgh-feature-descriptions.tsx
```diff
@@ -22,7 +22,7 @@ async function add(infoBanner: HTMLElement): Promise<void> {
 	conversationsUrl.searchParams.set('q', `sort:updated-desc "${feature.id}"`);
 
 	infoBanner.before(
-		<div className="Box mb-3">
+		<div className="Box mb-3 d-inline-block width-full">
```
fregante, line 25: Thank you! Can you also add a comment? It's not immediately clear why they're there
```suggestion
		{/* Block and width classes required to avoid margin collapse */}
		<div className="Box mb-3 d-inline-block width-full">
```

fregante (approved):

PR #6346, opened 2023-02-15
Properly clear field in `clear-pr-merge-commit-message`

vdimir:

### DIFF
--- a/source/features/clear-pr-merge-commit-message.tsx
+++ b/source/features/clear-pr-merge-commit-message.tsx
@@ -24,7 +24,7 @@ async function init(): Promise<void | false> {
 		return false;
 	}
 
-	set(messageField, cleanedMessage + '\n');
+	set(messageField, cleanedMessage ? cleanedMessage + '\n' : '');
 	attachElement(messageField, {
 		after: () => (
 			<div>


fregante (changes requested):
What is this about? I need more information. In what scenario does `cleanCommitMessage` ever return a whitespace-only string?

vdimir:
It could be connected with any other extension. I tried to disable user-css and user-js, still reproduces. Which additional information may help to investigate?

`Firefox Developer Edition 111.0b1 (64-bit) on macOS`

https://user-images.githubusercontent.com/7023786/219027615-9a881c44-c0bc-491c-9091-872d7e7bd0f2.mov

fregante:
I see, the change makes sense but the solution can be simplified. The function does not return whitespace-only strings so the issue is exclusively adding the final line break after an empty string. 

The solution can be simplified to:

``` js 
set(field, message ? message + '\n' : '')
``` 

Also could you make sure that the linting step passes?

fregante (approved):

fregante:
Thank you Vladimir!

ssbarnea:
I wonder if that bug is related to the one we recently faced, where pressing the merge (squash) button created a very ugly and unexpected commit message: https://github.com/ansible/vscode-ansible/pull/782

It happened several times for at least two different users in the last two weeks and both of us are users of refined-github. Still, we were not able to manually reproduce the issues and the moment we see it is, usually too late to action on it.

RussKie:
> I wonder if that bug is related to the one we recently faced, where pressing the merge (squash) button created a very ugly and unexpected commit message: [ansible/vscode-ansible#782](https://github.com/ansible/vscode-ansible/pull/782)
> 
> It happened several times for at least two different users in the last two weeks and both of us are users of refined-github. Still, we were not able to manually reproduce the issues and the moment we see it is, usually too late to action on it.

Yes, that's precisely one of the issues I described in https://github.com/refined-github/refined-github/issues/6547.

PR #6354, opened 2023-02-18
Drop `copy-file` feature

fregante:
👋 

Its time has come, GitHub has implemented it a long time ago and we only kept it around because it was missing on some files. The new file views break the feature so it's time to drop it.

This was one of the first features

Native feature appearing even on mobile

image

Related:

- https://github.com/refined-github/refined-github/issues/4802


yakov116 (approved):

NasaGeek:
It doesn't appear that Github ever implemented support for this in Gists (or it's since broken). Is there any appetite in reintroducing this feature for Gists specifically?

PR #6357, opened 2023-02-18
Run 50 more features before the whole page has loaded

fregante:
By default, features no longer wait for DOM-ready. This makes slower features more visible/painful in code


source/features/reload-failed-proxied-images.tsx
```diff
@@ -6,6 +6,8 @@ import delegate, {DelegateEvent} from 'delegate-it';
 import features from '../feature-manager';
 
 async function handleErroredImage({delegateTarget}: DelegateEvent<ErrorEvent, HTMLImageElement>): Promise<void> {
+	console.log('Refined GitHub: image failed loading, will retry', delegateTarget.src);
```
fregante, line 9: I'm adding some logging to see if this feature is ever needed anymore

source/features/git-checkout-pr.tsx
```diff
@@ -125,5 +125,6 @@ void features.add(import.meta.url, {
 	exclude: [
 		pageDetect.isClosedPR,
 	],
+	awaitDomReady: false,
```
fregante, line 128: There are a couple of features that have DOM-based filters AND don't await dom-ready. I only kept those that aren't fundamental, like this exclusion. I prefer a faster feature to a console without errors. This can be fixed in a better way (e.g. move this check inside the event listener)

Same goes for features that use `isBlank` as an exclusion: as mentioned in #5222, the manager should instead catch the error and silence it on blank pages.

source/features/linkify-commit-sha.tsx
```diff
@@ -6,7 +6,7 @@ import {wrap} from '../helpers/dom-utils';
 import features from '../feature-manager';
 
 function init(): void {
-	const element = select('.sha.user-select-contain');
+	const element = select('.sha.user-select-contain:not(a *)');
```
fregante, line 9: Automatic "don't rewrap links" deduplication logic

fregante:
Will merge this tomorrow and create a new release probably

PR #6358, opened 2023-02-18
typo in options.html

xBiei:
smol typo 🤏.


busches:
It's correct, you can follow or sponsor their work.

PR #6374, opened 2023-02-22
Improve UX of `dim-bots` by disabling the feature on click

### LINKED ISSUE #5993: Disable `dim-bots` entirely (for the current load) after interacting with a PR
Extracted from https://github.com/refined-github/refined-github/pull/5946#pullrequestreview-1100448376

cc @gaetanmaisse 

---

As I found myself dealing with bots more often, I've found our UI a bit clunky to work with.

What are your thoughts on this alternative behavior?

- dim and collapse them (same as now)
- un-dim them on hover (50% of what we do now)
- add "Expand bots" text on hover (even just via `.bot:after {content: "expand"}` 🆕
- expand and undim all bots when clicking on any non-link on a bot line (which would include this fake "expand" link) 🆕

This would keep the bots dimmed until you start interacting with them, at which point the feature is effectively disabled until a reload.

We only need to ensure to update the scroll position after the click (like we do with `toggle-all-with-alt`)

--- 

If you'd like to try, the easiest way would be something like:

```js
function handler (event) {
  if (event.delegateTarget.closest('a, button, summary')) {
		return
	}
	document.body.classList.add('rgh-no-mo-dim-bot')
}
delegate(document, 'click', '.rgh-dim-bot', handler, {signal})
```

We don't need to actually add the "Expand bots" label at this point. I tried it and it's a little awkward to position with CSS only. Maybe I'd like a robot icon, but still difficult to make it follow the text color while living in the CSS file as well.

fregante:
- Closes #5993 
- Better solution to https://github.com/refined-github/refined-github/issues/5158
	- than https://github.com/refined-github/refined-github/pull/5198

Probably the next improvements to this feature is to make it CSS-only via `:has()` to avoid the delay after load

## Test URLs

- PRs: https://github.com/OctoLinker/OctoLinker/pulls?q=is%3Apr+is%3Aclosed
- Commits: https://github.com/typed-ember/ember-cli-typescript/commits/master?after=5ff0c078a4274aeccaf83382c0d6b46323f57397+174

## PRs

ko

## Commits (`dim-bots` and `mark-merge-commits-in-list`)

The latter now also has a fade and it matches the opacity

<table>
<tr>
	<th>Before
	<th>After
<tr>
<td> 

https://user-images.githubusercontent.com/1402241/220601266-12a8a1a8-9f8e-46b6-8213-7338f15e7313.mov

<td>

https://user-images.githubusercontent.com/1402241/220601280-a459847f-115d-4153-ac02-2b32156d8bac.mov

</table>

### DIFF
--- a/source/features/dim-bots.tsx
+++ b/source/features/dim-bots.tsx
@@ -76,9 +76,7 @@ void features.add(import.meta.url, {
 
 Test URLs
 
-
-https://github.com/typed-ember/ember-cli-typescript/commits/master
-https://github.com/OctoLinker/OctoLinker/commits/master
-https://github.com/OctoLinker/OctoLinker/pulls?q=is%3Apr+is%3Aclosed
+- Commits: https://github.com/typed-ember/ember-cli-typescript/commits/master?after=5ff0c078a4274aeccaf83382c0d6b46323f57397+174
+- PRs: https://github.com/OctoLinker/OctoLinker/pulls?q=is%3Apr+is%3Aclosed
 
 */
--- a/source/features/dim-bots.tsx
+++ b/source/features/dim-bots.tsx
@@ -1,6 +1,7 @@
 import './dim-bots.css';
 import select from 'select-dom';
 import * as pageDetect from 'github-url-detection';
+import delegate from 'delegate-it';
 
 import features from '../feature-manager';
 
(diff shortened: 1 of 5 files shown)


source/features/dim-bots.tsx
```diff
@@ -1,8 +1,10 @@
 import './dim-bots.css';
 import select from 'select-dom';
 import * as pageDetect from 'github-url-detection';
+import delegate from 'delegate-it';
 
 import features from '../feature-manager';
+import preserveScroll from '../helpers/preserve-scroll';
```
fregante, line 7: I love this little util

PR #6376, opened 2023-02-22
Meta: Update `webext-storage-cache`

fregante:
- Includes https://github.com/fregante/webext-storage-cache/pull/39

It looks pretty smooth, no great improvement though since many functions silently depend on a global (`user/repo`). This should help on that front:

- https://github.com/refined-github/refined-github/issues/5848


fregante:
### You Only Live As Many Times As You're Allowed To Merge Without Consequences

PR #6385, opened 2023-02-23
Sort Refined GitHub shortcuts in Shortcut Help dialog

fregante:
- Closes https://github.com/refined-github/refined-github/issues/5439

## Test URLs

This page, press the `?` key

## Screenshot

<table>
<tr>
	<th>Before
	<th>After
<tr>
	<td>Screenshot 2
	<td>Screenshot 1
</table>


fregante:
I might reject this because this has the unwanted side effect of losing the `g 1 2 3 4` sort 😰

Thoughts? @kidonng

PR #6392, opened 2023-02-26
Fix `one-click-pr-or-gist` private free repos pr creation

### LINKED ISSUE #6391: `one-click-pr-or-gist` is broken in free private repos
### Description

As you can see, in free private repo, the button has changed to do the promotion of Draft PR that offers paid plans.
> 

This rule can not be respected because GitHub just changed the design for PRIVATE repositories (with free plan only).

Public repo
> 

Any private repo (in free plan)
> 

### Steps

1. Go to any private repo with a free plan
2. Go to pull requests
3. Click on the "New Pull Request" green button
4. Select a mergeable branch
5. Click on the "New Pull Request" green button

### Extension version

23.2.20

### Browser(s) used

Chrome

DraftProducts:
Closes #6391

This page affects PR creation in private repos

## Screenshot

Without the fix
> 

With the fix
>

### DIFF
--- a/source/features/one-click-pr-or-gist.tsx
+++ b/source/features/one-click-pr-or-gist.tsx
@@ -17,12 +17,7 @@ function init(): void | false {
 	for (const dropdownItem of select.all('.select-menu-item', initialGroupedButtons)) {
 		let title = select('.select-menu-item-heading', dropdownItem)!.textContent!.trim();
 		const description = select('.description', dropdownItem)!.textContent!.trim();
-		const radioButton = select('input[type=radio]', dropdownItem);
-
-		if (!radioButton) {
-			continue;
-		}
-
+		const radioButton = select('input[type=radio]', dropdownItem)!;
 		const classList = ['btn', 'ml-2', 'tooltipped', 'tooltipped-s'];
 
 		if (/\bdraft\b/i.test(title)) {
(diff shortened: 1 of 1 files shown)


yakov116:
It would be better to check if the "unable to create draft" button exists. Since now if the dom changes at any point we won't see any errors

fregante:
I think that we just want to leave the drop-down untouched in that case. 

Can you craft a selector for that "teams" label and add it as a function in the `exclude` array? Find "exclude" on https://github.com/refined-github/refined-github/blob/main/contributing.md#featuresadd for more info

DraftProducts:
Hi @fregante 👋 
I'd like to propose an alternative rather than disabling this feature.
If the goal is to keep promoting GitHub for its paid plan, couldn't we just recreate a button next to it that would continue to open the promotion modal? Adding the attribute `data-show-dialog-id="drafts-upgrade-dialog"` to any button would open it.

fregante:
No, there's no point in keeping the feature enabled for a button that doesn't work.

We can't hide it (as in your original screenshot) and we should not hide the inline documentation/explanation. So leaving it alone is the easiest thing to do (also because there will be no one around to fix it when it breaks)

fregante (approved):
Thank you, excellent! 💚

CyberFlameGO (approved):
LGTM!

fregante (approved):
Perfect it works, thank you!

PR #6394, opened 2023-03-01
Disable `quick-comment-edit` on non-editable issues

fregante:
- Fixes https://github.com/refined-github/refined-github/issues/6390

Note this fix is `has` selector-based: #5797 

## Test URLs

A conversation:

- you commented on in the past
- you can't edit or comment on anymore (locked or user blocked you)

## Screenshot

Screenshot

### DIFF
--- a/source/features/quick-comment-edit.tsx
+++ b/source/features/quick-comment-edit.tsx
@@ -1,3 +1,4 @@
+import './quick-comment-edit.css';
 import React from 'dom-chef';
 import select from 'select-dom';
 import {PencilIcon} from '@primer/octicons-react';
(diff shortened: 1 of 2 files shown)


source/features/quick-comment-edit.tsx
```diff
@@ -59,5 +60,7 @@ void features.add(import.meta.url, {
 	include: [
 		pageDetect.hasComments,
 	],
+	// The feature is "disabled" via CSS selector when the conversation is locked.
+	// We want the edit buttons to appear while the conversation is loading, but we only know it's locked when the page has finished.
```
fregante, line 64: I did not see any changes in the comment’s DOM, they're all marked as "editable".

If anyone sees any differences, please leave a comment or send a PR to move this check from CSS to the `addQuickEditButton` function (which should be able to tell whether the current comment is editable)

It would be even greater if this could be merged into a single "can comment be edited by current user", replacing the whole selector-based logic we have

fregante: WIP: It might be possible via `.comment .js-update-minimized-content` 🤔

Issue #6405, opened 2023-03-04
`useful-forks` does not show on the new forks page


Aholicknight:
### Description

![msedge_20230315_zZ0evV1kkd](https://user-images.githubusercontent.com/7843719/222928848-1b399959-d42b-4ebe-a8e0-da06c897a78e.png)

Find useful forks button no longer appears on the new fork page.

### How to replicate the issue + URL

https://github.com/refined-github/refined-github/forks

### Extension version

23.3.2

### Browser(s) used

Microsoft Edge

fregante:
Mentioned in https://github.com/refined-github/refined-github/issues/6397

I'll leave this issue open so it can be fixed in the meanwhile

KaKi87:
In the meantime, [here](https://git.kaki87.net/KaKi87/config/src/commit/2c72652c09764d29acf7f26cf86334ae63bd8f44/ocgpenflpmgnfapjedencafcfakcekcd/Redirector.json#L161)'s a [Redirector](https://github.com/einaregilsson/Redirector) configuration that redirects the new forks page to the old one.

fregante:
You can just disable the beta, as far as I know.

KaKi87:
I keep clicking *switch back to the tree view* but it still gets me to the new page.

PR #6406, opened 2023-03-05
Fix "show useful forks" button on new forks page

Aholicknight:
This fixes the problem #6405 (hopefully)

I will check to see if this problem is fixed on my end with this new code change, which is why this is a draft pull request.

### DIFF
--- a/source/features/useful-forks.tsx
+++ b/source/features/useful-forks.tsx
@@ -18,12 +18,14 @@ function getUrl(): string {
 
 async function init(): Promise<void | false> {
 	const forkCount = await elementReady('#Layout-main');
+        const forkCount = await elementReady('#isRepoForksList'); // keep both
 	if (looseParseInt(forkCount) === 0) {
 		return false;
 	}
 
 	const selector = [
 		'#Layout-main', // `isRepoForksList`
+		'#isRepoForksList' // keep both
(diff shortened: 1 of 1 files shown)


source/features/useful-forks.tsx
```diff
@@ -17,13 +17,13 @@ function getUrl(): string {
...
 async function init(): Promise<void | false> {
-	const forkCount = await elementReady('#repo-network-counter');
+	const forkCount = await elementReady('#Layout-main');
 	if (looseParseInt(forkCount) === 0) {
 		return false;
 	}
 
 	const selector = [
-		'#network', // `isRepoForksList`
+		'#Layout-main', // `isRepoForksList`
```
fregante, line 26: Thanks for the PR! We should keep both selectors here so that it keeps working on the old view until it's finalized. 

The old selector can be commented with `// `isRepoForksList` before March 2023

Aholicknight (commented):
Thank you!

source/features/useful-forks.tsx
Aholicknight: @fregante So do I change the `Layout-main` back to `isRepoForksList`?

fregante: Both should stay. So just *add* the new one

PR #6407, opened 2023-03-06
Remove style override that is no longer needed

mskelton:
The custom style which applies margin-right to the header button icons is no longer needed as GitHub has these styles corrected itself, and moreover the custom style is slightly incorrect as it was applying both to the icon before the text as well as the dropdown icon which doesn't need the margin.

## Test URLs

## Screenshot

**Before**

image

**After**

SCR-20230306-agp

### DIFF
--- a/source/features/github-bugs.css
+++ b/source/features/github-bugs.css
@@ -53,11 +53,6 @@
 	vertical-align: baseline !important;
 }
 
-/* Fix spacing of repo header button icons #5620 */
-.pagehead-actions :is(.btn, summary) .octicon {
-	margin-right: 4px !important;
-}
-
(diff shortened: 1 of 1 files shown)


yakov116 (approved):

mskelton:
@fregante Is it possible to get this merged. Looks like it has a couple approvals.

fregante:
I didn't merge it beacause I didn't test it, this PR undoes https://github.com/refined-github/refined-github/pull/5620 without replacing the fix. See before/after:

ko

mskelton:
@fregante Ah, sorry I missed that.  Added back in my other PR:

https://github.com/refined-github/refined-github/pull/6415/files#diff-24ef5929102e80386c178f9c0e79536d11501971d9dc7e7a3490b48bf565f174

PR #6421, opened 2023-03-18
Stop highlighting NEW features

fregante:
Closes https://github.com/refined-github/refined-github/issues/5539

~~I'm not gonna drop the code for now, but it's been broken for longer than a year. I'll likely drop it sooner or later if not fixed,~~ we do have releases now.

### DIFF
--- a/source/options.tsx
+++ b/source/options.tsx
@@ -244,8 +244,13 @@ async function generateDom(): Promise<void> {
 	await perDomainOptions.syncForm('form');
 
 	// Decorate list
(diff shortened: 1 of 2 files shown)


source/options.tsx
```diff
@@ -188,27 +183,6 @@ function featuresFilterHandler(event: Event): void {
 	}
 }
 
-async function highlightNewFeatures(): Promise<void> {
```
fregante, line 191: On second thought: 👋

PR #6427, opened 2023-03-21
Add `netiquette` feature

fregante:
- Part of #5934 

## Test URLs

- Old issue: https://github.com/refined-github/refined-github/issues/3076
- Old PR: https://github.com/refined-github/refined-github/pull/159

## Screenshot

Screenshot 10

Screenshot 11

Issue #5934, opened 2022-08-30
Discourage commenting on old, closed, or super-active threads


fregante:
### Description

Every so often someone leaves an incomplete bug report, a feature request, or a “me too” comment on closed issues and PRs. I do this too occasionally.

We should add a banner above the comment field to encourage people to open a new issue or do nothing instead. 

Messaging examples:

- This issue/PR is 3 years old. Consider opening a new issue instead
- This issue/PR is super active. Only leave a comment if you read the whole conversation
- Leaving a comment will notify about 30 people. Plz stop

### Example URLs

- https://github.com/refined-github/refined-github/issues/3076
- https://github.com/refined-github/refined-github/issues/3229

cheap-glitch:
+100 for the first one, not sure about the other two

fregante:
> * This issue/PR is super active. Only leave a comment if you read the whole conversation

StackOverflow [does it](https://stackoverflow.com/q/23667086/288906)

<img width="557" alt="Screen Shot 3" src="https://user-images.githubusercontent.com/1402241/187418889-d9569845-24f4-4faf-ad04-8034c6f3b297.png">


> * Leaving a comment will notify about 30 people. Plz stop

Receiving random notifications from threads you commented on 3 years ago is annoying, it's why I've been locking more and more threads lately, especially the most commented ones. I'd rather not do that honestly, but people keep commenting there.

fregante:
Related:

- https://github.com/refined-github/refined-github/issues/5815#issuecomment-1181290574

fregante:
Heheh yeah it's more as a reminder. This also occasionally happens to me when I land on an issue from Google and comment only to realize it's 4 years old. 

Adding a banner as a reminder of etiquette is welcome to me too and I hope some GitHub developer notices it and makes it a default.

fregante:
So, next step

- add banner to super-active threads like [this one](https://togithub.com/refined-github/refined-github/issues/3686)

What should the logic be?

<table>
<tr>
	<th>Participants
	<th><img width="316" alt="Screenshot 8" src="https://user-images.githubusercontent.com/1402241/230735414-95c99607-5ef4-4792-961a-a0f98756c3cc.png">
<tr>
	<td>Comments
	<td><img width="103" alt="Screenshot 9" src="https://user-images.githubusercontent.com/1402241/230735412-8cc93b1c-3d84-46d4-8e07-a190240da8e1.png">
</table>

[...]

PR #6432, opened 2023-03-26
Add `rgh-netiquette` feature

fregante:
- Continues from #6427 
- Part of #5934 

## Why

I don't want to lock issues, but also I don't want comments on years-old issues.

I think the only comments allowed there would be "GitHub implemented this" or "I extracted this code into its own extension". That is: actually-useful comments, not just disagreements or bug reports.

## Possible additional changes

- [ ] Completely block comments for old issues (>6mo)
- [ ] Keep comment box visible to us maintainers

👆 maybe not, because of the aforementioned _useful_ comments.

## Test URLs

- Old issue: https://github.com/refined-github/refined-github/issues/3076
- Old PR: https://github.com/refined-github/refined-github/pull/159

## Demo

fixed

### DIFF
--- a/source/features/netiquette.tsx
+++ b/source/features/netiquette.tsx
@@ -49,7 +49,7 @@ export function getNoticeText(): JSX.Element {
 function addConversationBanner(newCommentField: HTMLElement): void {
 	newCommentField.before(
 		createBanner({
-			icon: <InfoIcon/>,
+			icon: <InfoIcon className="m-0"/>,
 			classes: 'p-2 m-2 text-small color-fg-muted border-0'.split(' '),
 			text: getNoticeText(),
 		}),
--- a/source/features/netiquette.tsx
+++ b/source/features/netiquette.tsx
@@ -8,7 +8,7 @@ import {InfoIcon} from '@primer/octicons-react';
 import createBanner from '../github-helpers/banner';
 import features from '../feature-manager';
 import observe from '../helpers/selector-observer';
-import {buildRepoURL} from '../github-helpers';
+import {buildRepoURL, isAnyRefinedGitHubRepo} from '../github-helpers';
 import {closedOrMergedMarkerSelector, getLastCloseEvent} from './jump-to-conversation-close-event';
 
 const isClosedOrMerged = (): boolean => select.exists(closedOrMergedMarkerSelector);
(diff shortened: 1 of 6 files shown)


source/features/netiquette.tsx
```diff
@@ -26,40 +26,49 @@ function getCloseDate(): Date {
...
-			text: (
-				<div className="d-flex flex-items-center gap-2">
-					<InfoIcon className="m-0"/>
-					{/* TODO: Drop any after https://github.com/frenic/csstype/issues/177 */}
-					<span style={{textWrap: 'balance'} as any}>
-						This {pageDetect.isPR() ? 'PR' : 'issue'} was closed {ago}. Please consider opening a {newIssue} instead of leaving a comment here.
-					</span>
-				</div>
-			),
+			text: getNoticeText(),
```
fregante, line 54: To avoid duplicating a lot of logic and to avoid awkward if/else in this feature, I tried making multiple components the best I could. Not super great 😂 but reasonable dry

source/features/netiquette.tsx
```diff
@@ -71,7 +80,7 @@ void features.add(import.meta.url, {
 
 Test URLs
 
-- Old issue: https://github.com/refined-github/refined-github/issues/3076
-- Old PR: https://github.com/refined-github/refined-github/pull/159
+- Old issue: https://github.com/facebook/react/issues/227
```
fregante, line 83: Screenshot 7

source/features/netiquette.tsx
```diff
@@ -71,7 +80,7 @@ void features.add(import.meta.url, {
...
+- Old PR: https://github.com/facebook/react/pull/209
```
fregante, line 84: Screenshot 8

source/features/rgh-netiquette.tsx
```diff
@@ -0,0 +1,63 @@
...
+	],
+	awaitDomReady: true, // We're specifically looking for the last event
+	init,
+});
+
+/*
+
+Test URLs
+
+- Old issue: https://github.com/refined-github/refined-github/issues/3076
```
fregante, line 60: Screenshot 9

source/features/rgh-netiquette.tsx
```diff
@@ -0,0 +1,63 @@
...
+- Old PR: https://github.com/refined-github/refined-github/pull/159
```
fregante, line 61: Screenshot 10

source/features/netiquette.tsx
fregante: Mobile is a bit broken at the moment:

Screenshot 11

fregante: Spacing fixed

## Desktop


Screenshot 2


## Mobile

Screenshot 1

source/features/netiquette.tsx
```diff
@@ -26,40 +26,49 @@ function getCloseDate(): Date {
...
+	return timeAgo(closingDate) > threeMonths;
+}
 
+export function getNoticeText(): JSX.Element {
+	const closingDate = getCloseDate();
 	const ago = <strong>{twas(closingDate.getTime())}</strong>;
 	const newIssue = <a href={buildRepoURL('issues/new/choose')}>new issue</a>;
+	return (
+		<>
+			This {pageDetect.isPR() ? 'PR' : 'issue'} was closed {ago}. Please consider opening a {newIssue} instead of leaving a comment here.
```
SuperSandro2000, line 44: Not sure if showing this by default is a good idea. There are projects which actually do the opposite to reduce the amount of created issues and relevant people get a notification.

kidonng: We do what makes sense for most projects (personally I don't mind either way), and it is possible to bring back the comment field.

Also you are kinda commenting on the wrong place, as this pull request is for this repo (https://github.com/refined-github/refined-github) only.

SuperSandro2000: > Also you are kinda commenting on the wrong place, as this pull request is for this repo ([refined-github/refined-github](https://github.com/refined-github/refined-github)) only.

Than this part is borked, since this also shows up in other places like https://github.com/NixOS/nixpkgs/pull/205391

image

fregante: @SuperSandro2000 to clarify:

### `netiquette` (everywhere, can be disabled)

It just adds a notice

Screenshot 8

### `rgh-netiquette` (just the `refined-github` organization, cannot be disabled)

It hides the field completely

Screenshot 10

SuperSandro2000: > There are 2 features, one that applies to all repositories and one that applies to only refined-github.
> 
> This PR is for the refined-github.

Yeah, I got this, that's not my point. I was thinking if this is not better disabled by default, but the more I think about it, the more I am divided and the more I think the unintrusive reminder is probably a good idea.

PR #6435, opened 2023-03-28
Drop `unhide-reactions-on-mobile` feature

fregante:
It's here

👇


fregante:
## Screenshot

> IMG_6759

PR #6442, opened 2023-03-31
Add `workflow-log-copy-paste` feature

fregante:
- Close https://github.com/refined-github/refined-github/issues/6317

So painful. Finally I did this.

## Test URLs

1. Open https://github.com/refined-github/refined-github/actions
2. Navigate to a log (they expire eventually)


fregante:
You gotta be kidding me. I wrote this feature while I was postponing a system updated that finally added support for `user-select: none` to Safari 😂 

Neat 📸

PR #6457, opened 2023-04-06
Restore `swap-branches-on-compare`

### LINKED ISSUE #6361: `swap-branches-on-compare` broken
### Description

No link is added here

```
undefined is not an object (evaluating 'icon.parentElement.attributes["aria-label"].value')
```

Screenshot

https://github.com/refined-github/refined-github/compare/main...23.2.1

### Extension version

23.2.1

### Browser(s) used

safari

kidonng:
Fix #6361

## Test URLs

https://github.com/refined-github/refined-github/compare/23.2.1...main

## Screenshot

截屏2023-04-06 19 52 55

### DIFF
--- a/source/features/swap-branches-on-compare.tsx
+++ b/source/features/swap-branches-on-compare.tsx
@@ -32,6 +32,8 @@ void features.add(import.meta.url, {
 	],
 	exclude: [
 		() => /\.\.+/.exec(location.pathname)?.[0]!.length === 2,
+		// Prefer the native "switch base" button if it is available
+		// Do not use `pageDetect.isBlank` which gives false positives
 		() => select.exists('.range-editor + .blankslate'),
 	],
 	awaitDomReady: true,
--- a/source/features/swap-branches-on-compare.tsx
+++ b/source/features/swap-branches-on-compare.tsx
@@ -17,11 +17,10 @@ function init(): void {
 		references.unshift(select('.branch span')!.textContent!);
 	}
 
-	const link = select('.js-toggle-range-editor-cross-repo')!;
-	link.after(
-		' or ',
-		<a href={buildRepoURL('compare/' + references.join('...'))}>
-			switch the base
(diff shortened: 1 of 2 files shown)


fregante:
Thanks for the PR!

I noticed you sent this before my last comment. How do you feel about this?

Screenshot 10

Or maybe `.btn-octicon` to match the other icon colors:

Screenshot 12

source/features/swap-branches-on-compare.tsx
```diff
@@ -18,19 +17,34 @@ function init(): void {
...
 }
 
 void features.add(import.meta.url, {
 	include: [
 		pageDetect.isCompare,
 	],
 	exclude: [
+		// Disable on Two-dot Git diff comparison #4453
 		() => /\.\.+/.exec(location.pathname)?.[0]!.length === 2,
+		// Prefer the native "switch base" button if it is available
```
fregante, line 36: Is there a native one? When does it appear? Maybe we should just reuse its style

kidonng: Oops, not button but link (it's the one in my first screenshot)

fregante: Oh right

kidonng:
Yeah a button works too, in that case there's no need to hide it.

fregante (approved):
LGTM, I'm just afraid that it steals too much attention with a regular button style. Don't like `btn-octicon`?

kidonng:
Because it is... not an icon? Looks too weird and not something clickable to me.

grahamperrin:
Hey, thanks! 

A few times, before this feature reappeared, I clicked the three dots beneath the arrow (between the two things to be compared), imagining that a menu would appear with _Swap_ amongst its options. Oops.

kidonng:
The three dots link is a nice touch nonetheless, it just _happened_ to break our feature.

PR #6462, opened 2023-04-06
Update CSS comment

WebSnke:


fregante:
Could you also update the selector? It's not actually working

WebSnke:
> Could you also update the selector? It's not actually working 

That's weird since I only updated the comment.  It shouldn't affect the code at all.

fregante:
I know but I'm saying it's altering a comment on a piece of code that doesn't work. Feels strange 😃

WebSnke:
> Could you also update the selector? It's not actually working

How do you know that it is not working? The newest version of the Add-on works in Firefox for me.

fregante:
I think you're misunderstanding me. I'm saying you're making unnecessary changes to a feature that is currently not working. The line after the comment you edited is not working, the year is not being hidden:

Screenshot

I'm not saying that you broke it, I'm saying that I'm not going to accept menial PRs that alter comments around broken code.

kidonng:
> Could you also update the selector? It's not actually working

It's working for me 🤔 I checked it's hidden by RGH

fregante:
Indeed it was updated in https://github.com/refined-github/refined-github/pull/6415 and it's still broken in Safari 👍

PR #6465, opened 2023-04-07
 `sticky-notifications-actions` - New feature

### LINKED ISSUE #6176: Notification page : keep the action bar sticky
### Description

In the notification page, an action toolbar is present when selecting at least one notification. 

This notification bar disappear with the scroll. 

Once you selected every notification that you want to resolve, you need to scroll on the top of the page to click on "done".

This proposal suggest to keep the action bar sticky on the top of the page once you start scrolling.

### Screenshot

A tab to show the problem, and a tab with a CSS POC that show the expected behavior.

https://user-images.githubusercontent.com/1398469/203935704-c9f97d0f-e9c9-4798-b541-96a2b9d2da04.mp4

### Example URLs

jdeniau:
Closes #6176 

<details>
<summary>Outdated demos</summary>

## Test URLs

## Screenshot

image

https://user-images.githubusercontent.com/1398469/203935704-c9f97d0f-e9c9-4798-b541-96a2b9d2da04.mp4

</details>


source/features/sticky-notifications-actions.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+function init(): void {
+	const filterBox = selectHas(FILTER_BAR_SELECTOR);
+
+	if (filterBox) {
+		filterBox.classList.add(CLASS_TO_ADD);
+	}
+}
+
+void features.add(import.meta.url, {
```
fregante, line 18: Thank you for your PR! 

This feature can be CSS-only, you can drop this entire file. I think the feature looks something like:

```css
.js-notifications-container .Box:has(.js-select-all-text) {
	position: sticky;
	top: 0;
	z-index: 1;
}
```

It's so minor it doesn't need to be disabled. Please ensure it works on small screens too (or at least that it doesn't break anything)

source/features/sticky-notifications-actions.css
```diff
@@ -0,0 +1,10 @@
+.rgh-sticky-notifications-actions {
+	position: sticky;
+	top: 0;
+	z-index: 1;
```
kidonng, line 4:
```suggestion
	z-index: 10;
```

Or it may get covered

source/features/sticky-notifications-actions.tsx
fregante: > Please ensure it works on small screens too

Also ensure that "select by" and the other dropdown also works correctly, after setting `z-index: 10`

kidonng:
Hah! Guess I [should have just submitted this](https://github.com/kidonng/cherry/commit/ebbe026d6bc6451cf6d7935abfff238eead13c74) to RGH first 😆

source/features/sticky-notifications-actions.tsx
jdeniau: I started like that, but the `:has` selector [does not work on firefox for now](https://caniuse.com/css-has), so the `selectHas` js helper is really useful here.

The fact that there is no selector to target this action bar makes me want to limit the side effect too.

source/features/sticky-notifications-actions.css
jdeniau: I had no issue with `z-index: 1`. Do you have a case in mind where it would be covered ? I will re-test this but I would prefer setting the lowest possible z-index, what do you think ?

kidonng: image

source/features/sticky-notifications-actions.tsx
fregante: Unfortunately we're not accepting new code that exists only to work around Firefox bugs. 95% of the users don't need it and maintainers will have to rewrite the feature in the future. It's easier to copy-paste my suggested CSS file than review the current PR’s JS code

fregante: I should add that Firefox supports some `has` selectors via flag. See:

- #5797

jdeniau: OK no problem. I think this will be in firefox soon enough though.

source/features/sticky-notifications-actions.css
```diff
@@ -0,0 +1,10 @@
+.js-notifications-container .Box:has(.js-select-all-text) {
```
kidonng, line 1:
```suggestion
.notifications-list > .Box-header {
```

I have been using this for a long time, should not need `:has`

source/features/sticky-notifications-actions.tsx
fregante: Thanks for the screenshots! I think this piece of code should be inside a media query, like `@media (min-height: 500px)`, to ensure we don't cause issues on small viewports

jdeniau:
@fregante I handled all the review and the CI, and rebased on `main` .

I kept the `:has` selector because otherwise all `.Box-header` were targeted, so the code did not work. Furthermore, firefox did enabled the feature behind a feature-flag and should be available in firefox 121 according to caniuse that should be released on [19 December 2023](https://whattrainisitnow.com/calendar/)

readme.md
```diff
@@ -347,6 +347,7 @@ Thanks for contributing! 🦋🙌
 - [](# "linkify-notification-repository-header") [Linkifies the header of each notification group (when grouped by repository).](https://user-images.githubusercontent.com/1402241/80849887-81531c00-8c19-11ea-8777-7294ce318630.png)
 - [](# "stop-redirecting-in-notification-bar") [Stops redirecting to notification inbox from notification bar actions while holding <kbd>Alt</kbd>.](https://user-images.githubusercontent.com/202916/80318782-c38cef80-880c-11ea-9226-72c585f42a51.png)
 - [](# "last-notification-page-button") [Adds a link to the last page of notifications.](https://user-images.githubusercontent.com/16872793/199828181-3ff2cef3-8740-4efa-8122-8f2f222bd657.png)
+- [](# "sticky-notifications-actions") [Make the notifications action bar sticky.](https://user-images.githubusercontent.com/1398469/230582441-aed8bb24-4180-488e-b1d5-4e4a45a5d31c.png)
```
fregante, line 350: It's not clear from the screenshot, can you make a very short gif? Also don't include the browser UI

fregante (approved):
thank you!

fregante (changes requested):
It doesn't work correctly

gif2

jdeniau:
> It doesn't work correctly
> 
> gif2 [ gif2 ](https://user-images.githubusercontent.com/1402241/285101784-161318c0-72ec-4a1c-8320-bc59622c8ad7.gif) [ ](https://user-images.githubusercontent.com/1402241/285101784-161318c0-72ec-4a1c-8320-bc59622c8ad7.gif)

I am investigating, but for the record, it does work fine for me in the "default" notification view, but does not work when I filter on a repository.

https://github.com/refined-github/refined-github/assets/1398469/01501e1d-9be1-4d8f-8684-6d90d2cf689e

jdeniau:
OK it's because sometime the "select all" box is above the first bloc:

image

But when you have one block only, the "select all" box is "grouped" with the notifications:

image

jdeniau:
@fregante I did updated the CSS to handle the two use-cases and uploaded the following GIF

https://user-images.githubusercontent.com/1398469/285497350-95ead695-a023-4aeb-b3bc-7c815157b439.gif

I tested on chrome and firefox with `:has` selector enabled. It looks OK to me on both.

fregante:
<table>
<tr>
 <th>By date
 <th>By repo
<tr>
 <td>
 <td>
</table>


<table>
<tr><th>Dark mode + wide screen demo
<tr><td>Screenshot 1
</table>

PR #6466, opened 2023-04-07
Enable `hide-user-forks` on more links

### LINKED ISSUE #6464: `hide-user-forks` missing from some links
### Description

Both missing in profile nav and profile menu with "Global navigation updates":

image

---

image

https://github.com/kidonng?tab=repositories

### Extension version

23.3.31

### Browser(s) used

Chrome

kidonng:
Resolve #6464

Draft pending "Global navigation update" preview

## Test URLs

- User profile: https://github.com/fregante
- Organization profile: https://github.com/refined-github
- User content
	- User profile: https://github.com/kidonng?tab=repositories
	- Organization repositories: https://github.com/orgs/refined-github/repositories

## Screenshot

N/A

### DIFF
--- a/source/features/set-default-repositories-type-to-sources.tsx
+++ b/source/features/set-default-repositories-type-to-sources.tsx
@@ -10,9 +10,11 @@ function addSourceTypeToLink(link: HTMLAnchorElement): void {
 // No `include`, no `signal` necessary
 function init(): void {
 	observe([
-		'.header-nav-current-user ~ a[href$="tab=repositories"]', // "Your repositories" in the header profile dropdown
-		'[aria-label="User profile"] a[href$="tab=repositories"]', // "Repositories" tab on user profile
+		`a:is([href^="/"], [href^="${location.origin}"])[href$="?tab=repositories"]`, // "Repositories" tab on user profile
+		'[aria-label="Organization"] [data-tab-item="org-header-overview-tab"] a', // "Overview" tab on organization profile
 		'[aria-label="Organization"] [data-tab-item="org-header-repositories-tab"] a', // "Repositories" tab on organization profile
+		'[aria-label="Organization"] a[data-tab-item="i0overview-tab"]', // "Overview" tab on organization profile (Global navigation update)
+		'[aria-label="Organization"] a[data-tab-item="i1repositories-tab"]', // "Repositories" tab on organization profile (Global navigation update)
 		'a[data-hovercard-type="organization"]', // Organization name on repo header + organization list on user profile
 	], addSourceTypeToLink);
 }
--- a/source/features/set-default-repositories-type-to-sources.tsx
+++ b/source/features/set-default-repositories-type-to-sources.tsx
@@ -11,10 +11,9 @@ function addSourceTypeToLink(link: HTMLAnchorElement): void {
 function init(): void {
 	observe([
 		`a:is([href^="/"], [href^="${location.origin}"])[href$="?tab=repositories"]`, // "Repositories" tab on user profile, only matches relative or same-origin links
+		`a:is([href^="/orgs/"], [href^="${location.origin}/orgs/"])[href$="/repositories"]`, // "Repositories" tab on organization profile, only matches relative or same-origin links
 		'[aria-label="Organization"] [data-tab-item="org-header-overview-tab"] a', // "Overview" tab on organization profile
-		'[aria-label="Organization"] [data-tab-item="org-header-repositories-tab"] a', // "Repositories" tab on organization profile
 		'[aria-label="Organization"] a[data-tab-item="i0overview-tab"]', // "Overview" tab on organization profile (Global navigation update)
-		'[aria-label="Organization"] a[data-tab-item="i1repositories-tab"]', // "Repositories" tab on organization profile (Global navigation update)
 		'a[data-hovercard-type="organization"]', // Organization name on repo header + organization list on user profile
 	], addSourceTypeToLink);
 }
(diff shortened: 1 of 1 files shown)


source/features/set-default-repositories-type-to-sources.tsx
```diff
@@ -10,9 +10,11 @@ function addSourceTypeToLink(link: HTMLAnchorElement): void {
 // No `include`, no `signal` necessary
 function init(): void {
 	observe([
-		'.header-nav-current-user ~ a[href$="tab=repositories"]', // "Your repositories" in the header profile dropdown
-		'[aria-label="User profile"] a[href$="tab=repositories"]', // "Repositories" tab on user profile
+		`a:is([href^="/"], [href^="${location.origin}"])[href$="?tab=repositories"]`, // "Repositories" tab on user profile
```
fregante, line 13: This is the only one needed

fregante (changes requested):

source/features/set-default-repositories-type-to-sources.tsx
kidonng: Other links doesn't match `?tab=repositories` though?

fregante: Nice selector, it would be good to explain the `is` part

fregante: Right, there should be another generic one then, or else this link won't be covered either I think:

- https://github.com/orgs/refined-github/repositories

```
https://github.com/orgs/refined-github/repositories
```

fregante: By the way CSS Nesting is coming soon 😍 and it would simplify a lot of our selectors: https://caniuse.com/css-nesting

- Chrome has it
- Safari 16.5 has it (maybe June)
- Firefox eh https://bugzilla.mozilla.org/show_bug.cgi?id=1648037

kidonng:
Now it still misses direct links to organization profiles, i.e.

```
https://github.com/refined-github
```

But it's not worth it to check every link `isProfile` just for this.

source/features/set-default-repositories-type-to-sources.tsx
fregante: This wasn't addressed. We should not depend on any classes.


> Now it still misses direct links to organization profiles, i.e.
> 
> ```
> https://github.com/refined-github
> ```

This feature does not target those links, it should target every:

- `/*?tab=repositories`
- `/orgs/*/repositories/` 

otherwise we're back to square one (i.e. "some links don't work")

Also I'd change the name of the feature to something more reasonable because such long features are impossible to mention in commit titles. I'd go for `hide-user-forks` maybe

source/features/set-default-repositories-type-to-sources.tsx
```diff
@@ -10,13 +10,23 @@ function addSourceTypeToLink(link: HTMLAnchorElement): void {
...
 function init(): void {
 	observe([
-		'.header-nav-current-user ~ a[href$="tab=repositories"]', // "Your repositories" in the header profile dropdown
-		'[aria-label="User profile"] a[href$="tab=repositories"]', // "Repositories" tab on user profile
-		'[aria-label="Organization"] [data-tab-item="org-header-repositories-tab"] a', // "Repositories" tab on organization profile
+		`a:is([href^="/"], [href^="${location.origin}"])[href$="?tab=repositories"]`, // "Repositories" tab on user profile, only matches relative or same-origin links
+		`a:is([href^="/orgs/"], [href^="${location.origin}/orgs/"])[href$="/repositories"]`, // "Repositories" tab on organization profile, only matches relative or same-origin links
+		'[aria-label="Organization"] [data-tab-item="org-header-overview-tab"] a', // "Overview" tab on organization profile
+		'[aria-label="Organization"] a[data-tab-item="i0overview-tab"]', // "Overview" tab on organization profile (Global navigation update)
 		'a[data-hovercard-type="organization"]', // Organization name on repo header + organization list on user profile
```
fregante, line 17: I think the solution is:

```suggestion
		// User repos
		`a:is([href^="/"], [href^="${location.origin}"])[href$="?tab=repositories"]`,

		// Organization repos
		`a:is([href^="/orgs/"], [href^="${location.origin}/orgs/"])[href$="/repositories"]`,
```

source/features/set-default-repositories-type-to-sources.tsx
```diff
@@ -10,13 +10,23 @@ function addSourceTypeToLink(link: HTMLAnchorElement): void {
...
 void features.add(import.meta.url, {
 	init,
 });
+
+/*
+
+Test URLs:
+
+- User profile: https://github.com/fregante
+- Organization profile: https://github.com/refined-github
```
fregante, line 30: To avoid confusion: These aren't the only URLs on which the feature runs, it runs everywhere. Instead of "test URLs" we need a list of examples to test, e.g.

```suggestion
Live test:
- https://github.com/fregante?tab=repositories
- https://github.com/orgs/refined-github/repositories
- The "Your repositories" link in the user dropdown in the header
- The "Repositories" tab in
	- https://github.com/fregante
	- https://github.com/refined-github
```

source/features/set-default-repositories-type-to-sources.tsx
fregante, line 1: I'd also change the name of the feature to something more reasonable because such long features are impossible to mention in commit titles. I'd go for `hide-user-forks` maybe

source/features/set-default-repositories-type-to-sources.tsx
kidonng: Did you mean we drop support for these?
Screenshot 2023-04-16 at 20 16 39
Screenshot 2023-04-16 at 20 16 57

PR #6472, opened 2023-04-08
Drop `clean-mergeability-box` feature

fregante:
Should we drop this feature? 29 lines of complex selectors only to play spot the differences:

<table>
<tr>
<td>

Screenshot 1

<td>

Screenshot 5

</table>

The original screenshot looks good in theory, but doesn't seem worth it, especially because GitHub blesses us with this kind of CSS to work on:

Screenshot

## Related

- Partially replaces https://github.com/refined-github/refined-github/pull/6468


kidonng (approved):
Basically what you said in #5631.

Issue #6336, opened 2023-02-11
`linkify-code` broken


fregante:
### Description

<table>
<tr>
	<td>Screenshot
<tr>
	<td><img width="568" alt="Screenshot" src="https://user-images.githubusercontent.com/1402241/218266444-51461ba7-5fdd-451d-9a10-38345bb17a8c.png">
</table>

### How to replicate the issue + URL

https://github.com/refined-github/refined-github/blob/be756e68dd4bf24d53e802e1fec126506979fa80/test/linkify-urls-in-code.js#L32-L51

### Extension version

23.2.1

### Browser(s) used

Safari

amanmalh:
Hi @fregante, I need more clarification on this. I tested this on chrome, as well as safari. The URLs under 'findURLs' test seem to be working fine for me. The issue reference in the comment is working fine as well. What exactly is broken?

fregante:
Thank you for looking into this! It's related to the new code view, you'll have to enable it to see the issue: https://docs.github.com/en/search-github/github-code-search/about-github-code-search#enabling-and-disabling-the-new-code-search-and-code-view-beta

amanmalh:
I've requested join the waitlist for the new code view, I'll look into it once I have access. Meantime, I'll look into other issues.

fregante:
Easier said than done 😅 

Given they just enabled the beta for everyone, I think the opposite should be suggested. 

1. We fix the feature on the new view
2. User notices it's broken
3. We suggest switching to the new view

kidonng:
This seems tricky because there's a `<textarea>` overlay above the actual code. We can linkify the code but no one is able to click them 😅

fregante:
🤔 what's that textarea for? Is the code inside it? How is the user able to select it?

kidonng:
Presumably for getting cursor position and acting on it, as it is easier than dealing with text nodes.

The real bummer is that the highlighted code is now pseudo-elements:

```css
.react-file-line [data-code-text]::before {
    content: attr(data-code-text);
}
```

Which means in addition to dealing with clicks, we need to construct the DOM back ourselves. And a11y will suffer.

I can't think of an efficient way to do all these without breaking GitHub functionality.

fregante:
> there's a `<textarea>` overlay above the actual code

I'm not seeing that 🤔 

Tested on a random file and element, it's able to linkify it as usual:


https://github.com/refined-github/refined-github/blob/main/test/get-current-committish.test.ts

[...]

PR #6476, opened 2023-04-08
Add `status-subscription` feature (WIP)

fregante:
- If finalized, closes https://github.com/refined-github/refined-github/issues/3005

Findings:

- The DOM doesn't include information regarding the exact current subscription state, but it does mention "custom state" (which could be either one or both "closed" and "reopened")

	Screenshot 11

- GraphQL doesn't include information regarding custom subscriptions states nor allows setting it: 
	- https://docs.github.com/en/graphql/reference/enums#subscriptionstate
	- https://docs.github.com/en/graphql/reference/mutations#updatesubscription
- I don't see any way to get or set subscription state to a specific issue via Rest API
	- https://docs.github.com/en/rest/issues
- Both the "unsubscribe" and "custom" forms point to `/notifications/thread` 👍 

That means:

- we can use that "reason" paragraph to detect the current state
- we we have to submit GitHub’s own form to make changes

UI-wise, none of the options mentioned so far work that well together with the current "subscribe/unsubscribe" button, because it acts as a "single action that also shows the (partial) status" and it's hard to combine it with "subscribe, but only to status events"

This is the best I could come up with: replacing the native button and sometimes hiding the reason:

Screenshot 14

here's how it compares to the native one:

Screenshot

### DIFF
--- a/source/features/status-subscription.tsx
+++ b/source/features/status-subscription.tsx
@@ -1,0 +1,57 @@
+import React from 'dom-chef';
+import * as pageDetect from 'github-url-detection';
+import {BellIcon, BellSlashIcon, IssueReopenedIcon} from '@primer/octicons-react';
+
+import features from '../feature-manager';
+import observe from '../helpers/selector-observer';
+
+function addButton(subscriptionButton: HTMLButtonElement): void {
+	subscriptionButton.before(
+		<div className="BtnGroup d-flex width-full">
+			<button
+				type="button"
+				className="btn btn-sm flex-grow-1 BtnGroup-item tooltipped tooltipped-s"
+				aria-label="Unsubscribe"
+			>
+				<BellSlashIcon/> None
+			</button>
+			<button
+				aria-selected
+				type="button"
+				className="btn btn-sm flex-grow-1 BtnGroup-item tooltipped tooltipped-s selected"
+				aria-label="Subscribe"
+			>
+				<BellIcon/> All
+			</button>
+			<button
+				type="button"
+				className="btn btn-sm flex-grow-1 BtnGroup-item tooltipped tooltipped-s"
+				aria-label="Subscribe just to close event"
+			>
+				<IssueReopenedIcon/> Status
+			</button>
+		</div>,
+	);
+	subscriptionButton.hidden = true;
+}
+
+function init(signal: AbortSignal): void | false {
+	observe('button[data-thread-subscribe-button]', addButton, {signal});
+}
+
+void features.add(import.meta.url, {
+	include: [
+		pageDetect.isConversation,
+	],
+	awaitDomReady: true, // The sidebar is at the end of the page
+	init,
+});
+
+/*
+
+Test URLs
+
+- Issue: https://github.com/refined-github/sandbox/issues/3
+- PR: https://github.com/refined-github/sandbox/pull/4
+
+*/
--- a/source/refined-github.css
+++ b/source/refined-github.css
@@ -188,3 +188,7 @@ div[data-target='readme-toc.content'] div.blob-wrapper img {
 	border-top-left-radius: 0;
 	border-bottom-left-radius: 0;
 }
+
+.flex-grow-1 {
+	flex-grow: 1 !important;
+}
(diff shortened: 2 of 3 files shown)


source/features/status-subscription.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+function addButton(subscriptionButton: HTMLButtonElement): void {
+	const status = getCurrentStatus(subscriptionButton);
+	// Save first
+	const originalId = subscriptionButton.form!.elements.id;
+
+	subscriptionButton.after(
+		<div className="BtnGroup d-flex width-full">
+			<SubButton
+				// @ts-expect-error I don't remember how to fix this
```
fregante, line 54: Probably just a consequence of dom-chef’s "bad types"

source/features/status-subscription.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+function getCurrentStatus(subscriptionButton: HTMLButtonElement): 'none' | 'all' | 'status' {
+	const reason = subscriptionButton
+		.closest('.thread-subscription-status')!
+		.querySelector('.reason')!
+		.textContent!;
+
+	// You’re receiving notifications because you chose custom settings for this thread.
+	if (reason.includes('custom settings')) {
+		return 'status';
```
fregante, line 35: The "Status" button will be highlighted even if the just one event was selected, there's no "partially-selected" status. Not something I'm interested in fixing, just click "None" and then "Status" to get the full status subscription. "Customize" shows the correct status anyway.

fregante:
This works beautifully without any events. Let's enjoy this until the Reactmageddon comes wipe Refined GitHub off the face of GitHub

https://github.com/refined-github/refined-github/assets/1402241/3937789f-0778-4c67-bdd1-7157095b023e

fregante:
Screenshot 16

readme.md
```diff
@@ -102,8 +102,8 @@ GitHub Enterprise is also supported: [How to enable it](https://fregante.github.
 			<p><img src="https://user-images.githubusercontent.com/1402241/109592127-5f922200-7ad4-11eb-8dfa-1d80fb28e70e.png">
 	<tr>
 		<th width="50%">
-			<p><a title="linkify-code"></a> Linkifies issue/PR references and URLs in code and conversation titles
-			<p><img src="https://cloud.githubusercontent.com/assets/170270/25370217/61718820-29b3-11e7-89c5-2959eaf8cac8.png">
+			<p><a title="linkify-code"></a> Lets you subscribe to opening/opening closing events of issues in one click.
+			<p><img src="https://github.com/refined-github/refined-github/assets/1402241/cbc98b51-d173-40c6-b21e-5f0bae3d800c">
```
fregante, line 106: Old features lose the highlight… especially because they're breaking 🥲 

- #6336

fregante: So this URL redirects to `https://github-production-user-asset-6210df.s3.amazonaws.com/1402241/238186901-cbc98b51-d173-40c6-b21e-5f0bae3d800c.png`

Should we just use the final URL? I think that as long as we have the original image uploaded to the issue, we can use the final .png URL (which is why the feature verification is failing)

yakov116:
Can't wait for this!!

source/refined-github.css
```diff
@@ -189,6 +189,10 @@ div[data-target='readme-toc.content'] div.blob-wrapper img {
 	border-bottom-left-radius: 0;
 }
 
+.flex-grow-1 {
+	flex-grow: 1 !important;
```
FloEdelmann, line 193: I guess the `.flex-1` or `.flex-auto` classes in Primer (https://primer.style/css/utilities/flexbox#flex) are not sufficient? See also https://developer.mozilla.org/en-US/docs/Web/CSS/flex#formal_definition.

fregante:
Finally! 🎉

yakov116:
This is amazing 👏

fregante:
I'm kind of focusing on _visible_ features that improve the workflow on GitHub rather than small refinements lately, I think these are really good:

- `previous-version`
- `netiquette`
- `status-subscription`
- `closing-remarks`
- `unreleased-commits`
- `rgh-feature-descriptions`

No as mere "click-saving" features but also as "the information you need, where you need it." Also kind of loving the new super-clear label in `restore-file`

PR #6480, opened 2023-04-09
Refactor `tab-size` to reduce conflicts

### LINKED ISSUE #6461: `tab-size` broken in new code view
### Description

When the GitHub tab size preference is set to anything but 4, the tab display is broken under the new code view preview (the old one, without directory listing on the left side, is not affected).

Screenshot

Upon inspection, it seems like GitHub added a read-only textarea element behind the existing code display, which is overriden by the catch-all rules in `tab-size`.

https://github.com/refined-github/refined-github/blob/c470d8103ebc9aa02735d4c0cb72d543333c5c06/source/features/tab-size.css#L3-L5

https://github.com/refined-github/refined-github/blob/c470d8103ebc9aa02735d4c0cb72d543333c5c06/source/features/tab-size.css#L11-L15

Screenshot 2023-04-06 at 9 51 13 PM

URL in the screenshot: https://github.com/refined-github/refined-github/blob/c470d8103ebc9aa02735d4c0cb72d543333c5c06/source/features/sync-pr-commit-title.tsx

"New Code Search and Code View" must be turned on under Feature preview.

### Extension version

23.3.31

### Browser(s) used

Chrome 111.0.5563.146 / macOS 13.3

### LINKED ISSUE #6477: File previews broken due to lack of respect for user's `tab-size`
### Description

On XML based languages, Refined GitHub renders the code in weird ways:
Seemingly, it puts in more spacing than is usual, so there is a remain of the code in it's 'usual' state.
Not so sure about this one.
###### 📝 Note that this only occurs when selecting text.
Second of all, the text selector can only go to the end of the 'original' black text and not of the white one, so I cannot visually select the whole text.

**I have a video of this, but could not upload it to GitHub.**
image
See:
https://riversidevalley-my.sharepoint.com/:v:/g/personal/lamparter_theriversi_de/EdcPsBJFcslMlVmICHSqgK0B_k0xd72Ep1hu9Tk35qFZoA?e=Dieyoo
This video is of https://github.com/files-community/Website/blob/cfcfa9098860261e4433eecc77ae80b2c0e8d281/src/i18n/locales/en-US.json

Seemingly occurs in HTML/XML based languages, such as Svelte, XAML and of course, HTML, as well as JSON (a lot of web languages! but not Ruby); I have used the Refined GitHub feature identifier and it did not identify the problem, however disabling the Refined GitHub extension does fix the problem.

Example: https://github.com/RiversideValley/Website/blob/main/src/app.html
image

### Extension version

23.3.31

### Browser(s) used

Microsoft Edge Insider Dev

kidonng:
Fix #6461
Fix #6477

## Test URLs

- Rendered documents: https://github.com/refined-github/refined-github/blob/154c0f8c86b23e52d1e1ee947b90bc36bc6188ba/contributing.md#featuresadd
- Code blocks and embeds in conversations: https://github.com/refined-github/sandbox/pull/19
- The comment field everywhere

### Unaffected URLs

- isFile: https://github.com/refined-github/refined-github/blob/6979e88e6a7b3273338a2651ac4b533609ef97df/source/options.html
- isEditingFile: https://github.com/refined-github/refined-github/edit/main/source/options.html
- isGist: https://gist.github.com/kidonng/3ab6b5779e1a5a46984c14eecc0fc3a0

## Screenshot

image


source/features/tab-size.css
```diff
@@ -1,6 +1,6 @@
 /* Make tab-indented code more readable on GitHub and Gist */
 
-* {
+.markdown-body {
```
fregante, line 3: In this case I don't think the rest of the file is needed, it needs to be reviewed

fregante: Maybe more specifically `code` because I don’t think tabs are rendered anywhere else in markdown
```suggestion
.markdown-body code {
```

kidonng: > In this case I don't think the rest of the file is needed, it needs to be reviewed

It's for `show-whitespaces`

> Maybe more specifically `code` because I don’t think tabs are rendered anywhere else in markdown

`code` is for inline code. `pre` isn't right either, we need to use [`codeElementsSelector`](https://github.com/refined-github/refined-github/blob/1c04d21e0f2391c9249157919b4db62847422a3b/source/github-helpers/dom-formatters.ts#L14), though you can see how this is unnecessary

fregante: `Code` isn't for inline code, every `pre` usually *also* wraps a `code`. Either way pre or code should be good enough here. 

> It's for show-whitespaces

If it only affects that now, we should move it there (and also extend it because there are 12 steps, not just 2,4,8,10,13)

fregante: Anyway not a big deal, it can stay without `code`

kidonng: > `Code` isn't for inline code, every `pre` usually _also_ wraps a `code`

I'm talking about how GitHub uses it, not HTML semantics

> If it only affects that now, we should move it there (and also extend it because there are 12 steps, not just 2,4,8,10,13)

You just read my mind 😄

fregante: > I'm talking about how GitHub uses it

Yeah, same:

```
here
```

Screenshot 7

> `code` is for inline code

So, again, `code` is _not_ for "inline code", it's just for "code"

source/features/show-whitespace.css
```diff
@@ -16,3 +16,53 @@
...
 }
+
+/* Set a sensible default for user content */
+.markdown-body {
+	--tab-size: 4;
+}
+
+[data-tab-size='1'] {
+	--tab-size: 1;
+}
```
kidonng, line 27: 99.9% people will either use 2, 4 or 8 and 0% will use this

```suggestion
```

yakov116: Maybe my memory is failing me, but I think we had a conflict with another extension so we added all of these

I could be waaay off, just mentioning it.

fregante: I think you're referring to the extension that allowed the user to customize the tab size, which we contributed in order to correctly pick up the right tab size in `show-whitespace`

If that's it, then we're good

fregante: These rules are used to carry the `tab-size` value into `show-whitespace` CSS variable. This is because we can't use `calculated(tab-size)` but only `var(--tab-size)`

source/features/tab-size.css
```diff
@@ -1,41 +1,15 @@
...
 
-* {
-	tab-size: var(--tab-size) !important;
-}
-
-[data-tab-size='2'] {
-	--tab-size: 2;
-}
-
 :root, /* Changes the browser default */
```
fregante, line 3:
```suggestion
:root, /* Changes the browser default, it applies to textareas too */
```

source/features/tab-size.css
```diff
@@ -1,41 +1,15 @@
...
-}
-
-[data-tab-size='8'] {
-	--tab-size: 8;
-}
-
-[data-tab-size='10'] {
-	--tab-size: 10;
+.comment-body .tab-size[data-tab-size='8'] /* User setting is ignored in comments #4833 */ {
+	tab-size: 4;
```
fregante, line 5: The solution here is to preserve that `tab-size: var(--tab-size)` to never have these inconsistencies:

```suggestion
	tab-size: var(--tab-size);
```

And also keep this only in the root to allow a global override:

```css
:root {
	--tab-size: 4;
}
```

fregante (changes requested):
See the "test" section in https://github.com/refined-github/refined-github/pull/4833

I see at least one issue currently:

- `tab-size.css` is setting `tab-size` but not `--tab-size`, which means `show-whitespace` will be broken

I think this only affects inlined code: Screenshot from https://github.com/refined-github/refined-github/pull/4833
Screenshot 10

source/features/tab-size.css
fregante: Also you can follow your own review here 😁 

https://github.com/refined-github/refined-github/pull/4833#discussion_r733884651

fregante:
I hope to merge this by tomorrow so that I can release 23.4.10

source/features/tab-size.css
```diff
@@ -1,41 +1,16 @@
...
-}
-
-:root, /* Changes the browser default */
-.comment-body [data-tab-size='8'], /* User setting is ignored in comments #4833 */
-[data-tab-size='4'] {
+:root, /* Changes the browser default, it applies to textareas too */
+.comment-body .tab-size[data-tab-size='8'] /* User setting is ignored in file blocks in comments #4833 */ {
 	--tab-size: 4;
+	tab-size: var(--tab-size);
 }
```
kidonng, line 7: Instead of this (setting `--tab-size` again inside comments), it may be better to have

```css
[data-tab-size='8']:not(.highlight) {
	--tab-size: 8;
}
```

Thoughts?

fregante: That looks good as long as the final result is 4 characters in that situation, which is the point of `tab-size`

kidonng: Right, this just makes it easier to override `--tab-size`:

```diff
- .comment-body .tab-size[data-tab-size='8'],
  :root {
  	tab-size: var(--tab-size);
  }
```

source/features/show-whitespace.css
```diff
@@ -16,3 +16,49 @@
...
+	--tab-size: 5;
+}
+
+[data-tab-size='6'] {
+	--tab-size: 6;
+}
+
+/* File blocks in comments always have [data-tab-size=8] #4833 */
+[data-tab-size='8']:not(.highlight) {
+	--tab-size: 8;
```
fregante, line 46: Oh I see what you meant. As long as this doesn't exclude other pieces in our Tests it looks great.

source/features/tab-size.css
```diff
@@ -1,41 +1,19 @@
 /* Make tab-indented code more readable on GitHub and Gist */
```
fregante, line 1: Let's update this description to mention this is only meant to fill some gaps

fregante (approved):
All aboard! Thanks for looking into this!

source/features/show-whitespace.css
fregante: `--tab-size` must always match `tab-size`. If `tab-size.css` wants to override this, it should take care of all of it, otherwise we risk losing `--tab-size` sync or selector sync

source/features/tab-size.css
```diff
@@ -1,41 +1,30 @@
...
+*/
 
-:root, /* Changes the browser default */
-.comment-body [data-tab-size='8'], /* User setting is ignored in comments #4833 */
-[data-tab-size='4'] {
+:root,
+.tab-size[data-tab-size='8'] {
 	--tab-size: 4;
+	tab-size: var(--tab-size);
 }
```
fregante, line 14: The `tab-size` feature overrides the tab size. This one single selector overrides it. I think this is the most consistent situation we got so far

fregante (approved):
I think we're good now, there was an issue with the inconsistency between `--tab-size` and `tab-size` in some situations, so I think this should make sure it doesn't happen anymore.

`* {tab-size: var(--tab-size)}` would probably be the only way to ensure this, but I don’t think it's necessary.

### Note

GitHub currently has a bug where the first tab in code embeds is trimmed. This is unrelated to RG (see the tabs below)

Screenshot 11

fregante:
phew finally 😅

source/features/show-whitespace.css
kidonng: > `--tab-size` must always match `tab-size`. If `tab-size.css` wants to override this, it should take care of all of it, otherwise we risk losing `--tab-size` sync or selector sync

That's one way to see it, and the reason why [I asked for comments above](https://github.com/refined-github/refined-github/pull/6480#discussion_r1161304645)

But the situation is they depend on each use unless `tab-size` doesn't set or use `var(--tab-size)`

source/features/show-whitespace.css
```diff
@@ -41,15 +41,26 @@
 	--tab-size: 6;
 }
 
-/* File blocks in comments always have [data-tab-size=8] #4833 */
-[data-tab-size='8']:not(.highlight) {
+[data-tab-size='7'] {
+	--tab-size: 7;
+}
```
kidonng, line 46: Does this exist? Preferences don't have it

source/features/tab-size.css
kidonng: Well, you could have just reveted [`eb6f729` (#6480)](https://github.com/refined-github/refined-github/pull/6480/commits/eb6f72935f3cf62db5c3917044fe3e2c153d6a41)

Now this selector isn't specific enough🤔

source/features/show-whitespace.css
fregante: You're right, this was more like (unnecessary) future-proofing 😂 

Screenshot

source/features/tab-size.css
fregante: Heh you're right I didn't even notice that I just undid your commit 😅

kidonng: It's fine, I guess I'm just bad at expressing my intent (and tend to complicate things) 😂

source/features/show-whitespace.css
kidonng: I will add this to lint (to-do)

PR #6486, opened 2023-04-10
Restore `quick-new-issue` on Global navigation update

### LINKED ISSUE #6455: `quick-new-issue` broken on "Global navigation update"
### Description

Screenshot 2

```
refined-github.js:3727 📕 0.0.0 → TypeError: Cannot read properties of null (reading 'textContent')
    at Module.isArchivedRepo (refined-github.js:26112:90)
    at isArchivedRepoAsync (refined-github.js:18131:60)
    at async init (refined-github.js:12525:7)
    at async runFeature (refined-github.js:3797:16)
    at async setupPageLoad (refined-github.js:3809:5)
```

Every page, including this issue

### Extension version

23.3.2, 23.3.31

### Browser(s) used

Safari, Chrome

fregante:
- Fixes #6455 

## Test URL

here

## Demo

Screenshot 27

## Missing

- [ ] https://github.com/refined-github/github-url-detection/pull/167

### DIFF
--- a/source/github-helpers/index.ts
+++ b/source/github-helpers/index.ts
@@ -177,7 +177,7 @@ export function shouldFeatureRun({
 
 export async function isArchivedRepoAsync(): Promise<boolean> {
 	// Load the bare minimum for `isArchivedRepo` to work
-	await elementReady('#repository-container-header');
+	await elementReady('main > div');
 
 	// DOM-based detection, we want awaitDomReady: false, so it needs to be here
 	return pageDetect.isArchivedRepo();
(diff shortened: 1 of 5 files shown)


source/github-helpers/index.ts
```diff
@@ -177,7 +177,7 @@ export function shouldFeatureRun({
 
 export async function isArchivedRepoAsync(): Promise<boolean> {
 	// Load the bare minimum for `isArchivedRepo` to work
-	await elementReady('#repository-container-header');
+	await elementReady('main > div');
```
fregante, line 180: This should work for both. The first `div` in `main` appears to always be the flash banner itself or some other empty element if the banner is missing.

I can't use `main` because that would wait for the whole page before ever resolving, whether the banner exists or not

PR #6487, opened 2023-04-10
Less eye-catching `rgh-netiquette`

### LINKED ISSUE #6484: Make `rgh-netiquette` less eye-catching
I think this is a bit much:

Screenshot 18

It should roughly match the visual weight of the comment box. Maybe it should be gray or just not a banner at all.

### Test URL

https://github.com/refined-github/sandbox/pull/17
https://github.com/refined-github/sandbox/pull/1

fregante:
- fixes #6484 

## Test URLs

https://github.com/refined-github/sandbox/pull/17
https://github.com/refined-github/sandbox/pull/1

## Before

Screenshot 25

## After

Screenshot 24

### Dark

Screenshot 26

### DIFF
--- a/source/features/rgh-netiquette.tsx
+++ b/source/features/rgh-netiquette.tsx
@@ -24,6 +24,7 @@ function addConversationBanner(newCommentBox: HTMLElement): void {
 	const banner = (
 		<TimelineItem>
 			{createBanner({
+				classes: ['rgh-bg-transparent'],
 				icon: <InfoIcon className="mr-1"/>,
 				text: <>{getNoticeText()} If it must really be here, you can {button}.</>,
 			})}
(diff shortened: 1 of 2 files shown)


fregante:
I think this does its job well:

- makes the issue look like it's locked
- anyone looking for the comment box, will find a reasonably-visible and large message
- anyone just visiting won't be bothered just body text with a thin light border

kidonng:
> * anyone looking for the comment box, will find a reasonably-visible and large message

Isn't that enough to leave out the border? Looks weird to me.

fregante:
It needs to fit in with the other boxes, also notice how GitHub does the same (border-ony) with their first box

Screenshot 28

The only inconsistency here is the info icon 😳

fregante:
Merging this for now before the release, we can discuss further improvements separately

PR #6488, opened 2023-04-10
Lint

kidonng:


source/features/highlight-collaborators-and-own-conversations.tsx
```diff
@@ -42,7 +42,7 @@ void features.add(import.meta.url, {
 		pageDetect.isRepoIssueOrPRList,
 	],
 	exclude: [
-		() => select.exists('.blankslate'),
+		pageDetect.isBlank,
```
kidonng, line 45: Regarding `isBlank`, I want to mention that `.blankslate` always exists on some pages (e.g. https://github.com/refined-github/refined-github/compare/main...esbuild-2), maybe we need to review usage of it to fix false positives

fregante: Refer to https://github.com/refined-github/refined-github/issues/5222#issuecomment-1168304082 before making large changes

source/features/archive-forks-link.tsx
```diff
@@ -23,6 +21,10 @@ void features.add(import.meta.url, {
 	include: [
 		pageDetect.hasRepoHeader,
 	],
+	asLongAs: [
+		pageDetect.isPublicRepo,
+		pageDetect.isArchivedRepo,
```
fregante, line 26: isArchivedRepo can't be used this way, that's why it was inlined. And also it can't have domReady:true because it's way up in the header. 

isPublicRepo can stay here though, it's meta tag based.

fregante (changes requested):

source/feature-manager.tsx
```diff
@@ -47,8 +47,11 @@ type FeatureLoader = {
 } & Partial<InternalRunConfig>;
 
 type InternalRunConfig = {
+	/** Every condition must be true */
 	asLongAs: BooleanFunction[] | undefined;
+	/** At least one condition must be true */
 	include: BooleanFunction[] | undefined;
+	/** No conditions must be true */
 	exclude: BooleanFunction[] | undefined;
```
fregante, line 55: Probably more readable as "Every condition must be false", both here and in the place you copied this from

PR #6489, opened 2023-04-10
Change `closing-remarks` style

### LINKED ISSUE #6482: `closing-remarks` has wrong style
### Description

Icon has unnecessary margin

image

https://github.com//refined-github/refined-github/pull/6427#ref-pullrequest-1640816674

### Extension version

23.3.31

### Browser(s) used

Chrome

kidonng:
Resolve #6482

- Chose blue icon because other colors are associated with their own meanings
- Unfortunately no alternative tag icon
- Write a helper to create timeline item?
- Our `conversation-activity-filter` won't filter it

## Test URLs

- "See release": https://github.com//refined-github/refined-github/pull/6427#ref-pullrequest-1640816674
- "Draft a new release": 🤷‍♂️

## Screenshot

image

### DIFF
--- a/source/features/closing-remarks.tsx
+++ b/source/features/closing-remarks.tsx
@@ -7,8 +7,6 @@ import * as pageDetect from 'github-url-detection';
 import features from '../feature-manager';
 import fetchDom from '../helpers/fetch-dom';
 import onPrMerge from '../github-events/on-pr-merge';
-import createBanner from '../github-helpers/banner';
-import TimelineItem from '../github-helpers/timeline-item';
 import attachElement from '../helpers/attach-element';
 import {canEditEveryComment} from './quick-comment-edit';
 import onConversationHeaderUpdate from '../github-events/on-conversation-header-update';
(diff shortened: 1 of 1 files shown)


fregante:
It looks good but this doesn't belong on the timeline: 

- we don't show the tag on the right spot (and we shouldn't, because our intent is to make this visible)
- non-events don't belong on the timeline ("this event didn't happen, make it happen?")

fregante:
I think it could work if you keep it below the horizontal rule as in my mockup. 

The other issue would be that the right-hand button might be too too far from the text itself since the text is so short, but this is probably relatively minor

kidonng:
Closing in favor of #6489

fregante:
Sorry, it looked good, but _free-flowing text_ always looks a little out of place, so I think a light border would fit better. Like the tips below:

Screenshot 10

kidonng:
My main point is that we should not make them look like alerts when they are not, which your simpler proposal (just remove the solid background) does achieve. Yeah it's a good middle ground.

PR #6492, opened 2023-04-10
Add `small-user-avatars` feature

### LINKED ISSUE #4906: Add `small-user-avatars` feature
### Description

Goal: be able to **identify which user has created the PRs directly from the PR list**

Current behaviour: 
- the GitHub userid is displayed, but when that userID is made of random numbers because of your company policy, it's _not that useful_...
- you can hover on the user id of a PR to see user details (including full name and avatar) but it's less useful that having the info displayed automatically for each PR without having to hover on each one

There are several ways to do that:
- add user avatar for graphic recognition
- add full username after the userid (like on PR comments)
- replace userid by username (reduce cluttering but would probably not be used by teams that work with meaningful userids)
- ....

### Screenshot

_No response_

### Example URL

https://github.com/refined-github/refined-github/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc

kidonng:
Resolve #4906

## Test URLs

- https://github.com/refined-github/refined-github/issues
- With pinned issue: https://github.com/dependabot/dependabot-core/issues

## Screenshot

image

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -162,7 +162,7 @@ Thanks for contributing! 🦋🙌
 - [](# "clean-repo-tabs") [Moves the "Security" and "Insights"  to the repository navigation dropdown. Also moves the "Projects", "Actions" and "Wiki" tabs if they're empty/unused.](https://user-images.githubusercontent.com/16872793/124681343-4a6c3c00-de96-11eb-9055-a8fc551e6eb8.png)
 - [](# "repo-avatars") [Adds the profile picture to the header of public repositories.](https://user-images.githubusercontent.com/44045911/177211845-c9b0fa37-c157-4449-890e-af2602c312e3.png)
 - [](# "quick-new-issue") [Adds a link to create issues from anywhere in a repository.](https://user-images.githubusercontent.com/1402241/218251057-b94b62dd-a944-4763-b78a-fc233f7c9fd3.png)
-- [](# "small-user-avatars") [Shows a small avatar next to the username in conversation lists.](https://user-images.githubusercontent.com/44045911/230923502-de8fdbf8-96eb-4dfc-b947-6c411ecd80dc.png)
+- [](# "small-user-avatars") [Shows a small avatar next to the username in conversation lists.](https://user-images.githubusercontent.com/44045911/230930235-62272e3c-edae-408a-a260-5b63b6edb77c.png)
 
 <!-- Refer to style guide above. Keep this message between sections. -->
 
--- a/readme.md
+++ b/readme.md
@@ -162,6 +162,7 @@ Thanks for contributing! 🦋🙌
 - [](# "clean-repo-tabs") [Moves the "Security" and "Insights"  to the repository navigation dropdown. Also moves the "Projects", "Actions" and "Wiki" tabs if they're empty/unused.](https://user-images.githubusercontent.com/16872793/124681343-4a6c3c00-de96-11eb-9055-a8fc551e6eb8.png)
 - [](# "repo-avatars") [Adds the profile picture to the header of public repositories.](https://user-images.githubusercontent.com/44045911/177211845-c9b0fa37-c157-4449-890e-af2602c312e3.png)
 - [](# "quick-new-issue") [Adds a link to create issues from anywhere in a repository.](https://user-images.githubusercontent.com/1402241/218251057-b94b62dd-a944-4763-b78a-fc233f7c9fd3.png)
+- [](# "small-user-avatars") [Shows a small avatar next to the username in conversation lists.](https://user-images.githubusercontent.com/44045911/230923502-de8fdbf8-96eb-4dfc-b947-6c411ecd80dc.png)
 
 <!-- Refer to style guide above. Keep this message between sections. -->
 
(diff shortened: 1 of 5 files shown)


source/features/small-user-avatars.tsx
```diff
@@ -0,0 +1,35 @@
...
+		const username = link.textContent!;
+		const size = 16;
+
+		link.prepend(
+			<img
+				className="avatar avatar-user mr-1"
+				src={getUserAvatarURL(username, size)!}
+				width={size}
+				height={size}
+				alt={`@${username}`}
```
fregante, line 21: The name is already there, this should not repeat it, or else it would read as:

> Image "@fregante", fregante

source/features/small-user-avatars.tsx
```diff
@@ -0,0 +1,35 @@
...
+function init(): void {
+	for (const link of select.all(
+		':is(.js-issue-row, .js-pinned-issue-list-item) [data-hovercard-type="user"]', // Excludes bots
+	)) {
+		const username = link.textContent!;
+		const size = 16;
+
+		link.prepend(
+			<img
+				className="avatar avatar-user mr-1"
```
fregante, line 17: The avatar has extra space on the left when inside a rgh-collaborator bubble. It should have the same spacing too and bottom (a flexbox might be necessary)

fregante (commented):
Looks good except a couple small issues

source/features/small-user-avatars.tsx
kidonng: Oops, this is perhaps an artifact back when I made `repo-avatars`

source/features/small-user-avatars.tsx
kidonng: I have already added some extra padding. I tried flexbox and it will wreck everything in that line

We don't need perfect aligning when it's that small

fregante: But this looks *unrefined* 🥲 

image

With inline-flex this shouldn't be excessively difficult to fix. I can look into it tomorrow

kidonng: Currently the vertical padding is smaller

Having it the same as horizontal margin will look like this:

image

This place is just too small... or you don't mind some extra margin?

fregante:
That looks better, but the alignment is off, see how nicely the name was wrapped before and how it's off-center now:

fixed

Also the height and `awaitDomReady: true` is causing content shift:

fixed


Maybe we should reduce the avatar and padding around it

source/features/highlight-collaborators-and-own-conversations.css
```diff
@@ -1,5 +1,11 @@
...
+	display: inline-block;
 	border: 1px solid var(--color-border-default);
 	border-radius: 2em;
-	padding: 2px 5px;
+	padding: 0 5px;
+}
+
+/* Extra padding for `small-user-avatars` feature */
+.rgh-collaborator:has(.avatar) {
+	padding-bottom: 2px;
```
fregante, line 10: I don't think any of this CSS should be used, it's causing the misalignment and layout shift. It's best to align the image directly

fregante: Here's a zero-shift image that can be inlined there (rough values, just show it's possible)


```html
<img class="from-avatar avatar-user" src="https://avatars.githubusercontent.com/u/44045911?s=40&amp;v=4" width="14" height="14" alt="@kidonng" style="
    margin-right: 3px;
    vertical-align: -2px;
    margin-left: -2.6px;
">
```

fixed

Tested in Chrome only, may need `inline-flex` or `em` to avoid font weirdness

kidonng: I really want to stick to 16px as that's what we use for `reaction-avatars`, but yeah 14px is fine, this place is just too small.

This is how it looks like now, without any magic numbers.

image

fregante: Let's not bring these PRs into loops. I already commented on that, it's not aligned and it should be fixed

kidonng: This is the best I can do:

image

Perhaps I'm really bad at CSS, I just don't know how to

- Not change `highlight-collaborators-and-own-conversations`
- Not causing content shifts
- And have all these places aligned

image

So I don't get which places you want it aligned.

fregante: I mean I want it as it appears in the gif I provided and the magic-numbered CSS. It's ok, I can look into a proper solution later if there's no easy way to achieve it.  Dealing with text alignment is always a pain in CSS

fregante (approved):
fixed

Looking good!

kidonng:
Oops, this would be better using `selector-observer` right?

fregante:
Judging by the behavior on load, it doesn't need to: GitHub's own `relative-time` handler runs on domready apparently. 

It's ok either way, the "deduplicate" logic is essentially deprecated now because we can't really trust it, it can change at any time and break everything again.

readme.md
```diff
@@ -162,6 +162,7 @@ Thanks for contributing! 🦋🙌
 - [](# "clean-repo-tabs") [Moves the "Security" and "Insights"  to the repository navigation dropdown. Also moves the "Projects", "Actions" and "Wiki" tabs if they're empty/unused.](https://user-images.githubusercontent.com/16872793/124681343-4a6c3c00-de96-11eb-9055-a8fc551e6eb8.png)
 - [](# "repo-avatars") [Adds the profile picture to the header of public repositories.](https://user-images.githubusercontent.com/44045911/177211845-c9b0fa37-c157-4449-890e-af2602c312e3.png)
 - [](# "quick-new-issue") [Adds a link to create issues from anywhere in a repository.](https://user-images.githubusercontent.com/1402241/218251057-b94b62dd-a944-4763-b78a-fc233f7c9fd3.png)
+- [](# "small-user-avatars") [Shows a small avatar next to the username in conversation lists.](https://user-images.githubusercontent.com/44045911/230960291-721f42cc-e1ac-4fdc-83ea-2430b062f9ce.png)
```
fregante, line 165: Can you update this in your lint PR? All screenshots should be in light mode and with an arrow ideally. Also it's best to disable nearby features to avoid confusion

Screenshot 6

Vendicated:
I love this, thank you!!

fregante:
Misaligned on iOS, this is why I was suggesting flexbox 🥲 

IMG_7612

fregante:
Also it makes the list a little heavy/busy 🥲 it's probably worth it though, let's see how the feature is received.

## Before

Screenshot 3


## After

Screenshot 2

PR #6502, opened 2023-04-12
Fix `more-conversation-filters` subscription link

### LINKED ISSUE #6501: `more-conversation-filters`: "Subscribed" link broken
### Description

Confirmed to be unrelated to "New Code Search and Code View" or "Global navigation update".

```
TypeError: Cannot read properties of undefined (reading 'value')
    at init (chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/refined-github.js:6108:149)
    at runFeature (chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/refined-github.js:3695:26)
    at setupPageLoad (chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/refined-github.js:3701:40)
    at chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/refined-github.js:3722:35
```

https://github.com/refined-github/refined-github/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc

### Extension version

23.4.10

### Browser(s) used

Chrome

kidonng:
Fix #6501

## Test URLs

https://github.com/refined-github/refined-github/issues

## Screenshot

I took the chance to make some slight visual changes:

image

Because previously when pressing back button you see this:

image

### DIFF
--- a/source/features/more-conversation-filters.tsx
+++ b/source/features/more-conversation-filters.tsx
@@ -16,8 +16,9 @@ function init(): void {
 
 	sourceItem.after(commentsLink);
 
-	// "Subscribed" link
-	const subscriptionsLink = select('#filters-select-menu a:last-child')!.cloneNode(true);
+	// "Subscribed" external link
+	const searchSyntaxLink = select('#filters-select-menu a:last-child')!;
+	const subscriptionsLink = searchSyntaxLink.cloneNode(true);
 	subscriptionsLink.lastElementChild!.textContent = 'Everything you subscribed to';
 
 	const subscriptionsUrl = new URL('https://github.com/notifications/subscriptions');


source/features/more-conversation-filters.tsx
```diff
@@ -17,12 +17,11 @@ function init(): void {
...
 	// "Subscribed" link
-	const subscriptionsLink = commentsLink.cloneNode(true);
-	subscriptionsLink.lastChild!.textContent = 'Everything you subscribed to';
-	subscriptionsLink.setAttribute('aria-checked', 'false'); // #4589
+	const subscriptionsLink = select('#filters-select-menu a:last-child')!.cloneNode(true);
+	subscriptionsLink.lastElementChild!.textContent = 'Everything you subscribed to';
 
 	const subscriptionsUrl = new URL('https://github.com/notifications/subscriptions');
-	const repositoryId = select('input[name="repository_id"]')!.value;
+	const repositoryId = select('meta[name="octolytics-dimension-repository_id"]')!.content;
```
kidonng, line 24: Worth preserving previous logic for GHE?

fregante: Sure if you feel like it. I've been keeping both selectors wherever simple, but if the logic is too different either we keep two versions or we drop GHE.

In this case I think you would need to write:

``` 
select(meta)?.content ?? select(input)!.value
```
This is probably the most straightforward way and the easiest one to drop in the future (which should always be our priority when doing this kind of code forks)

source/features/more-conversation-filters.tsx
```diff
@@ -17,12 +17,11 @@ function init(): void {
 	sourceItem.after(commentsLink);
 
 	// "Subscribed" link
-	const subscriptionsLink = commentsLink.cloneNode(true);
-	subscriptionsLink.lastChild!.textContent = 'Everything you subscribed to';
-	subscriptionsLink.setAttribute('aria-checked', 'false'); // #4589
+	const subscriptionsLink = select('#filters-select-menu a:last-child')!.cloneNode(true);
```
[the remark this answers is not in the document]
fregante: Love it. Worth a comment though. I'd perhaps even just extract the source link to its own variable to clarify what it's targeting:

```js
const searchSyntaxLink = select()
const subBlahBlah = searchSyntaxLink.clone(true)

PR #6506, opened 2023-04-13
Add `releases-dropdown` feature

fregante:
Follows:
- https://github.com/refined-github/refined-github/issues/6257
- https://github.com/refined-github/refined-github/pull/6438

Super easy, barely an inconvenience. The only issue is that it's technically misusing the `datalist` for navigation (just like mobile websites have done with `select` from 2007 until recently)

## Test URLs

## Screenshot

Screenshot

### DIFF
--- a/source/features/tags-dropdown.tsx
+++ b/source/features/tags-dropdown.tsx
@@ -1,0 +1,41 @@
+import React from 'dom-chef';
+import * as pageDetect from 'github-url-detection';
+
+import select from 'select-dom';
+
+import * as api from '../github-helpers/api';
+
+import features from '../feature-manager';
+
+const gql = `
+	repository() {
+		refs(refPrefix: "refs/tags/", last: 100) {
+			nodes {
+				name
+			}
+		}
+	}
+`;
+
+async function init(): Promise<false | void> {
+	const {repository} = await api.v4(gql);
+	const tags = repository.refs.nodes as Array<{name: string}>;
+	if (tags.length === 0) {
+		return false;
+	}
+
+	const searchField = select('input#release-filter')!;
+	searchField.after(
+		<datalist id="rgh-tags-dropdown">
+			{tags.reverse().map(tag => <option value={tag.name}/>)}
+		</datalist>,
+	);
+	searchField.setAttribute('list', 'rgh-tags-dropdown');
+}
+
+void features.add(import.meta.url, {
+	include: [
+		pageDetect.isReleasesOrTags,
+	],
+	init,
+});
--- a/source/features/tags-dropdown.tsx
+++ b/source/features/tags-dropdown.tsx
@@ -59,3 +59,12 @@ void features.add(import.meta.url, {
 	],
 	init,
 });
+
+/*
+
+## Test URLs
+
+https://github.com/refined-github/refined-github/tags
+https://github.com/refined-github/sandbox/releases
+
+*/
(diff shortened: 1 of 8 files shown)


source/features/tags-dropdown.tsx
```diff
@@ -0,0 +1,41 @@
...
+		<datalist id="rgh-tags-dropdown">
+			{tags.reverse().map(tag => <option value={tag.name}/>)}
+		</datalist>,
+	);
+	searchField.setAttribute('list', 'rgh-tags-dropdown');
+}
+
+void features.add(import.meta.url, {
+	include: [
+		pageDetect.isReleasesOrTags,
```
fregante, line 58: Related:

- https://github.com/refined-github/github-url-detection/issues/168

source/features/tags-dropdown.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+	// Save globally
+	latestTags = nodes.map(({name}) => name);
+
+	searchField.after(
+		<datalist id="rgh-tags-dropdown">
+			{latestTags.reverse().map(tag => <option value={tag}/>)}
+		</datalist>,
+	);
+	searchField.setAttribute('list', 'rgh-tags-dropdown');
```
fregante, line 47: Unreal… `searchField.list` is read-only 🤷‍♂️

fregante:
Ready. Readme screenshot:

Screenshot 4

https://user-images.githubusercontent.com/1402241/231678551-f4f11778-be43-4276-9ec5-628f3987e49f.mov

source/features/tags-dropdown.tsx
```diff
@@ -0,0 +1,70 @@
...
+		}
+	}
+`;
+
+// `datalist` selections don't have an `inputType`
+function selectionHandler(event: DelegateEvent<Event, HTMLInputElement>): void {
+	const field = event.delegateTarget;
+	const selectedTag = field.value;
+	if (!('inputType' in event) && latestTags!.includes(selectedTag)) {
+		location.href = buildRepoURL('releases/tag', selectedTag);
```
fregante, line 27: As expected, it's broken on unescaped tag names:

Screenshot 5

fregante: All good:

- https://github.com/refined-github/sandbox/releases/tag/JavaScript%25250Aalert.(1)''%22%22%60%60%2526apos%3B)%3C!--%3E%3CTitle%3CStyle%3CScript%3CtextArea%3CiFrame%3CnoScript%3E%3Ck%3CKcontentEditableautoFocusOnFocus%3D%24%7B%3B%7B(alert)(1)%7D%3E%3CBaseHref%3DX55.is%3E--%3E

Also cool UI:

- https://github.com/refined-github/sandbox/pull/40


Screenshot 7

source/features/tags-dropdown.tsx
kidonng: Then this should be `isReleases`? Because the search box only exists there

kidonng:
The search icon and dropdown arrow is unaligned. Can't unsee. Let's add these CSS:

```css
.subnav-search-icon {
    top: 8px;
}

#release-filter::-webkit-calendar-picker-indicator {
    position: relative;
    top: -2px;
}
```

image

source/features/tags-dropdown.tsx
fregante: > Then this should use isReleases? Because the search box only exists there


Done

> Also previous tags-dropdown works on tags page, should this one?

The issue is that the native field SEARCHES the releases notes, which means it does nothing on tags. See https://github.com/python/cpython/releases?q=v3&expanded=true which is empty even if there are plenty "v3" tags

So if we were to add this feature there, we'd have to reinvent the feature to _not_ use `datalist`. Not worth it in this PR (or maybe ever)

fregante:
Added both to `GitHub-bugs`. Let's always include test URLs there because it's hard to keep those bugfixes up to date.

Screenshot 5

PR #6448, opened 2023-04-04
Add options export/import

fregante:
- Fixes https://github.com/refined-github/refined-github/issues/3981
- Includes https://github.com/fregante/webext-options-sync/pull/65

Screenshot 7

Issue #3981, opened 2021-02-15
Import/export refined github settings


0xdevalias:
This extension has a LOT of features, and it took me quite a while to go through and toggle all of the ones off/on to suit my needs/desires. It would be good if I was able to export my settings as a `.json` file or similar, so that I can back them up, and/or to be able to share them with my friends/colleagues.

Then there would also need to be a way for me to import this `.json` of settings again too.

fregante:
Duplicate of #2726

tupaschoal:
I definitely vouch for that, specially because I want to use the same settings on different machines that I own, so it would be very handy.

kidonng:
For the moment, you can use the following method to sync settings manually:

```js
// 1. Open Console from DevTools of the options page
// 2. Run the snippet to copy the list of disabled features
copy($$('.feature > input:not(:checked)').map(i => i.id))
// 3. On another browser, open console and input disabled features
disabledFeatures = // paste here
// 4. Run the snippet to sync states
$$('.feature > input').forEach(i => i.checked === disabledFeatures.includes(i.id) && i.click())
```

---

@tupaschoal If you own the machines, you can just login to the same account and the browser will sync them.

tupaschoal:
Thanks for the console commands. I don't want to login because some machines are not owned by me (i.e. work machines), but used on a daily basis.

noroadsleft:
I'm immensely grateful for this extension - it's become a vital part of my toolkit here on GitHub - so I am greatly dismayed to become the bearer of potentially bad news:

I'm attempting to use this feature in Firefox, and it does not appear to work at all. Clicking Export or Import results in the following:

![image](https://user-images.githubusercontent.com/18669334/230980397-fc0cca98-fbae-47cb-a011-a41240c04db2.png)

This is with Refined GitHub 23.4.10 in [Firefox 111.0 running on Linux Mint 21.1 MATE 64-bit](https://user-images.githubusercontent.com/18669334/230980732-dfbdc03c-df40-44a5-afe1-afa74a35fc09.png), but I received the same error attempting to export the settings from my Firefox 111.0.1 on Windows 8.1 64-bit.

I'm unclear as to whether this requires https://github.com/fregante/webext-options-sync/pull/65 to be merged in order to work, so if that's the case please let me know. Otherwise I'll just wait patiently.

Thank you for a wonderfully powerful extension.

kidonng:
[...]

PR #6518, opened 2023-04-14
Add support for Firefox and Safari in options import/export

fregante:
Follows:

-  #3981 
-  #6448 

Includes:

- https://github.com/fregante/webext-options-sync/pull/65/files/06fafa8a6f002908e2ef43683a4d8181baf612da..7cf1c25080528a177192d8563b68c5a68afb5f46

Working at first try 😍  I also tested importing a file from Chrome

https://user-images.githubusercontent.com/1402241/232036339-58584155-e802-4d0d-8d9c-4def47bf39f5.mov


fregante:
Of course Safari is in its own league. The file won't be saved but it's just shown and it's up to the user to save it. Even FileSaver.js fails here because it's uses Blob internally.

The closest I got to saving was by using `application/octet-stream` but it was saved as `Unknown.txt`, which is worse, because it requires a manual rename.

PR #6532, opened 2023-04-16
Use correct `view-last-pr-deployment` icon

kidonng:
GitHub represents deployments with rocket icon, I think it is better than a general external link icon.

image

Also changed the copy from "last" to "latest" which is more conventional.

## Test URLs

https://github.com//vuejs/core/pull/6160

## Screenshot

Screenshot 2023-04-16 at 21 00 48


yakov116 (approved):
LGTM

fregante (approved):

PR #6542, opened 2023-04-18
Meta: Refine bisect message

janpio:
Nice feature, but the wording could use some refinement.


busches:
Seems like we could drop `at all` now too.

janpio:
True, that still sounds a bit passive-aggressive when read wrong. Removed it.

busches (approved):

fregante (approved):

PR #6546, opened 2023-04-20
Fix `list-prs-for-file` (hopefully)

### LINKED ISSUE #6299: `list-prs-for-file`: Cannot read properties of undefined (reading 'getAttribute')
### Description

```
 0.0.0 → TypeError: Cannot read properties of undefined (reading 'getAttribute')
    at getCurrentPath (refined-github.js:9747:18)
    at async init (refined-github.js:9750:29)
    at async runFeature (refined-github.js:3797:16)
    at async setupPageLoad (refined-github.js:3809:5)
```

1. https://github.com/refined-github/yolo/blob/main/broken-features.csv

### Extension version

23.2.1

### Browser(s) used

Chrome

Pwuts:
Try pasting this in the console:
```js
JSON.parse(document.querySelector('[data-target="react-app.embeddedData"]').innerText.trim()).payload.path;
```

Test URL: https://github.com/Significant-Gravitas/Auto-GPT/blob/master/Dockerfile

My fix is based on this.
Not sure how to test, as Local Overrides don't work for extensions.

Closes #6299

### DIFF
--- a/source/features/list-prs-for-file.tsx
+++ b/source/features/list-prs-for-file.tsx
@@ -106,8 +106,8 @@ const getPrsByFile = cache.function('files-with-prs', async (): Promise<Record<s
 });
 
 async function getCurrentPath(): Promise<string> {
-	const viewData = JSON.parse((await elementReady('[data-target="react-app.embeddedData"]'))!.textContent.trim());
-	return viewData.payload.path;
+	const viewDataJson = (await elementReady('[data-target="react-app.embeddedData"]'))!.textContent.trim();
+	return JSON.parse(viewDataJson).payload.path;
 }
 
 async function init(): Promise<void> {
--- a/source/features/list-prs-for-file.tsx
+++ b/source/features/list-prs-for-file.tsx
@@ -106,7 +106,7 @@ const getPrsByFile = cache.function('files-with-prs', async (): Promise<Record<s
 });
 
 async function getCurrentPath(): Promise<string> {
-	const viewData = JSON.parse((await elementReady('[data-target="react-app.embeddedData"]'))!.innerText.trim());
+	const viewData = JSON.parse((await elementReady('[data-target="react-app.embeddedData"]'))!.textContent.trim());
 	return viewData.payload.path;
 }
 
--- a/source/features/list-prs-for-file.tsx
+++ b/source/features/list-prs-for-file.tsx
@@ -106,9 +106,9 @@ const getPrsByFile = cache.function('files-with-prs', async (): Promise<Record<s
 });
 
 async function getCurrentPath(): Promise<string> {
-	const data = (await elementReady('[data-target="react-app.embeddedData"]')).innerText.trim();
-	data = JSON.parse(data);
-	return data.payload.path;
+	const codeViewJson = (await elementReady('[data-target="react-app.embeddedData"]'))!.innerText.trim();
+	payload = JSON.parse(data).payload;
+	return payload.path;
 }
 
 async function init(): Promise<void> {
--- a/source/features/list-prs-for-file.tsx
+++ b/source/features/list-prs-for-file.tsx
@@ -106,9 +106,8 @@ const getPrsByFile = cache.function('files-with-prs', async (): Promise<Record<s
 });
 
 async function getCurrentPath(): Promise<string> {
-	const codeViewJson = (await elementReady('[data-target="react-app.embeddedData"]'))!.innerText.trim();
-	payload = JSON.parse(data).payload;
-	return payload.path;
+	const viewData = JSON.parse((await elementReady('[data-target="react-app.embeddedData"]'))!.innerText.trim());
+	return viewData.payload.path;
 }
 
 async function init(): Promise<void> {
--- a/source/features/list-prs-for-file.tsx
+++ b/source/features/list-prs-for-file.tsx
@@ -106,9 +106,9 @@ const getPrsByFile = cache.function('files-with-prs', async (): Promise<Record<s
 });
 
 async function getCurrentPath(): Promise<string> {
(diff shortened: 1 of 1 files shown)


source/features/list-prs-for-file.tsx
```diff
@@ -106,9 +106,8 @@ const getPrsByFile = cache.function('files-with-prs', async (): Promise<Record<s
 });
 
 async function getCurrentPath(): Promise<string> {
-	// `[aria-label="Copy path"]` on blob page, `#blob-edit-path` on edit page
-	const element = await elementReady('[aria-label="Copy path"], #blob-edit-path');
-	return element!.getAttribute('value')!;
+	const viewData = JSON.parse((await elementReady('[data-target="react-app.embeddedData"]'))!.innerText.trim());
+	return viewData.payload.path;
```
yakov116, line 110: For readability I would put the JSON.parse here

```suggestion
	return JSON.parse(viewData).payload.path;
```

yakov116:
Please include test URLs in your first post

Pwuts:
Sorry, editing from my browser as it seemed such a simple fix..

source/features/list-prs-for-file.tsx
```diff
@@ -106,9 +106,8 @@ const getPrsByFile = cache.function('files-with-prs', async (): Promise<Record<s
 });
 
 async function getCurrentPath(): Promise<string> {
-	// `[aria-label="Copy path"]` on blob page, `#blob-edit-path` on edit page
-	const element = await elementReady('[aria-label="Copy path"], #blob-edit-path');
-	return element!.getAttribute('value')!;
+	const viewDataJson = (await elementReady('[data-target="react-app.embeddedData"]'))!.textContent!.trim();
+	return JSON.parse(viewDataJson).payload.path;
```
fregante, line 110: @yakov116 Do you remember why we're not using location.href instead?

fregante (commented):
Thanks for the PR! I believe you're referring to Xcode/Safari, which is a pain to test. I suggest testing this in Chrome using the instructions in contributing.md

My guess is that this isn't enough to fix the feature because it also needs to be re-positioned in the page. This just happens to be the *first* failure point.

Pwuts:
I actually specifically downloaded the extension for this feature. It would be really useful in sorting out the PR tsunami in https://github.com/Significant-Gravitas/Auto-GPT :')

Pwuts:
> I believe you're referring to Xcode/Safari, which is a pain to test.

I was trying to test in Chrome using a Local Override to edit `refined-github.js`, but that didn't work.

fregante:
What's a Local Override?

Pwuts:
A way to edit source files of a website and preserve the edits through page reloads.

source/features/list-prs-for-file.tsx
yakov116: Its here since the feature was created https://github.com/refined-github/refined-github/pull/2197 🤷

fregante:
Oh you're talking about the chrome dev tools feature 

https://www.debugbear.com/blog/devtools-local-overrides


Yeah you'd need to build the extension locally instead

source/features/list-prs-for-file.tsx
fregante: Indeed it's present in the very first commit and we never reviewed that. 

@Pwuts let's use `new GitHubURL(location.href).filePath` instead of this getter

fregante (changes requested):
Indeed this now fails later in the file. Please build and load the extension locally and ensure it works, you can include a screenshot in the first PR "comment" after this is done

```
 0.0.0 → TypeError: Cannot read properties of undefined (reading 'closest')
    at addAfterBranchSelector (refined-github.js:18830:48)
    at async init (refined-github.js:10006:3)
    at async runFeature (refined-github.js:3797:16)
    at async setupPageLoad (refined-github.js:3809:5)
```

Pwuts:
FYI I don't have the time to fix this feature in the extension right now, so I'm going to fix this for myself by writing a simple shell script that uses the GraphQL API.

fregante:
No worries, thanks for trying

Pwuts:
Result:
```bash
#!/bin/bash
if [ -z "$1" ]; then
  echo "Usage: list-prs-for-path PATH [branch]";
  echo "Lists all PRs touching files starting with PATH";
  exit 1;
fi

branch=${2:-$(gh repo view --json defaultBranchRef --jq '.defaultBranchRef.name')}

echo "Querying PRs to branch $branch touching paths starting with '$1'";

response=$(gh api graphql -f query="query {
  repository($(gh repo view --json 'name,owner' --template 'owner: "{{.owner.login}}", name: "{{.name}}"')) {
    pullRequests(
      first: 100,
      states: OPEN,
      baseRefName: \"$branch\",
      orderBy: {
        field: UPDATED_AT,
        direction: DESC
      }
    ) {
      nodes {
        number author {
          login
        }
        title updatedAt url
        files(first: 100) {
          nodes {
            path
          }
        }
  } } }
}")

errors=$(jq -r '.errors' <<< "$response")

if [ ! "$errors" = "null" ]; then
  echo "GitHub API Query returned errors: $errors" >&2;
  exit 1;
fi

# # uncomment for debugging:
# jq -r '.' <<<"$response"

jq -r --arg path "$1" '.data.repository.pullRequests.nodes[] | select(.files.nodes | map(select(.path | startswith($path))) | length > 0) | {
  number: .number, title: .title, author: .author.login,
  updatedAt: .updatedAt,
  url: .url,
  filesChanged: .files.nodes | map(.path)
}' <<< "$response"
```

fregante:
- Fixed in https://github.com/refined-github/refined-github/pull/6578

It required a near-total rewrite because GitHub is making large changes to the UI

PR #6550, opened 2023-04-23
Add `previous-version` feature

### LINKED ISSUE #5942: View earlier version of file
### Description

You're looking at a file, how many clicks do you need to go to the previous version?

"The previous version" is at the core of how git works, yet this operation requires 2 full page loads.

We need to add a button to the previous version

### Screenshot

Example:

Screen Shot

### Example URLs

https://github.com/refined-github/refined-github/blob/main/.gitignore

134130:
## Description
- Closes #5942 
- Add "Previous" button to switch the file to a previous version
- Warn : ~~The header bar is different between not signed in and signed in~~ Not supporting not signed in
- Blaming is Welcome!

### Signed in
refined-github-signed-in

## Test URLs
https://github.com/refined-github/refined-github/blob/main/contributing.md

## Screenshot
image


yakov116:
There is no need to support the feature when not logged in

source/features/previous-version.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+import cache from 'webext-storage-cache';
+
+import features from '../feature-manager';
+import * as api from '../github-helpers/api';
+import observe from '../helpers/selector-observer';
+
+import GitHubURL from '../github-helpers/github-url';
+
+const getHistoryOids = cache.function('file-history', async (branch: string, filePath: string): Promise<string[] | false> => {
```
yakov116, line 13: Not by a computer but take a look at the initial implementation of this feature for how the api code should be done.

https://github.com/refined-github/refined-github/commit/b01263379d091df018286360968f48cadfe43a86

134130: Thank you!
I searched https://github.com/fregante/webext-storage-cache and now know how to use it correctly.

I also find the API calling always based on `refined-github/refined-github` and fixed.

The function needs `owner`, `repository`, `branch` and `filePath`.
So I make the cache key is combining of these values.

source/features/previous-version.tsx
```diff
@@ -0,0 +1,120 @@
...
+async function init(signal: AbortSignal): Promise<false | void> {
+	const historyOids = await getHistoryOids(new GitHubURL(location.href));
+
+	if (!historyOids) {
+		return false;
+	}
+
+	observe([
+		'#repos-sticky-header .react-blob-header-edit-and-raw-actions',	// For signed in
+		'readme-toc .Box-header.js-blob-header div:has(.BtnGroup)',			// For not signed in
```
fregante: Refined GitHub doesn't support logged out states

source/features/previous-version.tsx
```diff
@@ -0,0 +1,117 @@
...
+								const item = (
+									<div className="SelectMenu-item no-wrap text-normal f5">
+										<div className="d-flex width-full gap-4">
+											<div className="color-fg-default flex-auto">{i + 2} commits ago</div>
+											<div className="color-fg-muted flex-shrink-0">{element.slice(0, 7)}</div>
+										</div>
+									</div>
+								);
+
+								item.addEventListener('click', () => {
```
fregante, line 84: Events need to be attached via delegate() due to how GitHub handles AJAX navigation

source/features/previous-version.tsx
```diff
@@ -0,0 +1,117 @@
...
+					});
+
+					return button;
+				})()}
+			</div>
+
+			<details className="details-reset details-overlay select-menu BtnGroup-parent d-inline-block position-relative" open={false}>
+				<summary
+					data-disable-invalid="" data-disable-with=""
+					data-dropdown-tracking={'{"type":"blob_edit_dropdown.more_options_click","context":{"repository_id":51769689,"actor_id":null,"github_dev_enabled":false,"edit_enabled":false,"small_screen":false}'}
```
fregante, line 67: This is not for us, it should be dropped

source/features/previous-version.tsx
```diff
@@ -0,0 +1,117 @@
...
+						location.href = url.toString();
+					});
+
+					return button;
+				})()}
+			</div>
+
+			<details className="details-reset details-overlay select-menu BtnGroup-parent d-inline-block position-relative" open={false}>
+				<summary
+					data-disable-invalid="" data-disable-with=""
```
fregante, line 66: You should drop any attributes and classes you don't understand

source/features/previous-version.tsx
```diff
@@ -0,0 +1,117 @@
...
+import React from 'dom-chef';
+import cache from 'webext-storage-cache';
+
+import features from '../feature-manager';
+import * as api from '../github-helpers/api';
+import observe from '../helpers/selector-observer';
+
+import GitHubURL from '../github-helpers/github-url';
+
+const getHistoryOids = cache.function('previous-version', async (githubUrl: GitHubURL): Promise<string[] | false> => {
```
fregante, line 11: It's best to accept location.href here and then call GitHubURL inside. This also makes it easier to handle cacheKey. 

The name could be `getPastCommits`

source/features/previous-version.tsx
```diff
@@ -0,0 +1,117 @@
...
+		}
+	`);
+
+	const nodes = history.nodes as any[];
+
+	if (nodes.length === 0) {
+		return false;
+	}
+
+	return nodes.map<string>(n => n.oid);
```
fregante, line 30: This "string" here sets the type of `n`, which isn't right since it's an object

source/features/previous-version.tsx
```diff
@@ -0,0 +1,117 @@
...
+import cache from 'webext-storage-cache';
+
+import features from '../feature-manager';
+import * as api from '../github-helpers/api';
+import observe from '../helpers/selector-observer';
+
+import GitHubURL from '../github-helpers/github-url';
+
+const getHistoryOids = cache.function('previous-version', async (githubUrl: GitHubURL): Promise<string[] | false> => {
+	const {resource: {history}} = await api.v4(`
```
yakov116, line 12: You can do `repository()` and then use expression

No need to specify the URL in full

source/features/previous-version.tsx
fregante: I think Yakov is referring to the lack of "cache name" here. They key is being set to the URL, but it's missing `previous-versions:`

fregante: I just noticed that it's been fixed in your last commit

source/features/previous-version.tsx
```diff
@@ -0,0 +1,117 @@
...
+		resource(url: "/${githubUrl.user}/${githubUrl.repository}/commit/${githubUrl.branch}") {
+			... on Commit {
+				history(path: "${githubUrl.filePath}") {
```
fregante, line 15: Let's limit this to 4. I don’t think it's particularly useful to go to a random "15 commits ago"

I'm not sure we even need more than "previous version." Can we think of a reason to just jump to "two unnamed commits ago"?

yakov116: I can, but not more than 4.

source/features/previous-version.tsx
fregante: I'm not sure that helps, also due to future GraphQL improvements I mentioned before

source/features/previous-version.tsx
134130: Hmm.. The generic of Array.map means return type, not input type
But I just changed this to `nodes.map(node => node.oid as string);`

source/features/previous-version.tsx
134130: Yes, But in this case I need to use `resource` query not `repository` query.
So I'll just preserve it.

source/features/previous-version.tsx
134130: Yeah, not providing like Jetbrains's Git history. I think no need to go back over 4 commits.
If we show "commit message", then showing over 4 commits would be usefull

134130:
I changed your suggestions. Thanks for your supports!
And then I added some feature (fix bugs)
- Tooltips for Buttons
- Disable showing button when no history
- Disable showing expand button when history is only one

Please check these too. Thanks

source/features/previous-version.tsx
```diff
@@ -0,0 +1,128 @@
...
+import React from 'dom-chef';
+import cache from 'webext-storage-cache';
+
+import delegate from 'delegate-it';
+
+import features from '../feature-manager';
+import * as api from '../github-helpers/api';
+import observe from '../helpers/selector-observer';
+
+import GitHubURL from '../github-helpers/github-url';
```
yakov116, line 11:
```suggestion
import cache from 'webext-storage-cache';
import React from 'dom-chef';
import delegate from 'delegate-it';
import * as pageDetect from 'github-url-detection';

import features from '../feature-manager';
import observe from '../helpers/selector-observer';
import * as api from '../github-helpers/api';
import GitHubURL from '../github-helpers/github-url'; 
```

Nit pick just fixing the spacing

134130: Checked lint but GibHub Direct commit adds space on EOF.. 😞

source/features/previous-version.tsx
```diff
@@ -0,0 +1,126 @@
...
+					const button = (
+						<div className="btn-sm BtnGroup-item btn">
+							Previous
+						</div>
+					);
+
+					delegate(button, '*', 'click', () => {
+						const url = new GitHubURL(location.href);
+						url.branch = pastCommits[1];
+						location.href = url.toString();
```
fregante, line 97: I just realized you’re using a listener to create a fake link. Please just use `<a href` instead, never use buttons create links. 

Also for next time check how delegate is used in other features because this is also incorrect (it should be in init and it should start with `delegate(document, 'the actual selector'`

source/features/previous-version.tsx
```diff
@@ -0,0 +1,126 @@
...
+};
+
+async function init(signal: AbortSignal): Promise<false | void> {
+	const historyOids = await getPastCommits(location.href);
+
+	if (!historyOids) {
+		return false;
+	}
+
+	observe(['#repos-sticky-header .react-blob-header-edit-and-raw-actions'], add, {signal});
```
fregante, line 118:
```suggestion
	observe('#repos-sticky-header .react-blob-header-edit-and-raw-actions', add, {signal});
```

source/features/previous-version.tsx
```diff
@@ -0,0 +1,126 @@
...
+	actionButtons.prepend(button);
+};
+
+async function init(signal: AbortSignal): Promise<false | void> {
+	const historyOids = await getPastCommits(location.href);
+
+	if (!historyOids) {
+		return false;
+	}
+
```
fregante, line 117:
```suggestion
```

This is already called in the observer, we don't need to add this optimization

source/features/previous-version.tsx
```diff
@@ -0,0 +1,126 @@
...
+				}
+			}
+		}
+	`);
+
+	const nodes = history.nodes as any[];
+
+	if (nodes.length === 0) {
+		return false;
+	}
```
fregante, line 29: No need for a specific check and `false`. "No commits" is already identified by an empty array. This way you can exclusively use `if (commits.length === 0)` later on

source/features/previous-version.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+	return nodes.map(node => node.oid as string);
+}, {
+	maxAge: {hours: 1},
+	staleWhileRevalidate: {days: 1},
+	cacheKey: ([currHref]) => currHref,
+});
+
+const createDetailsButton = (pastCommits: string[]): Element | void => {
+	if (pastCommits.length <= 2) {
```
fregante, line 39: There's already an identical check outside this function, we don't need to repeat it

source/features/previous-version.tsx
```diff
@@ -0,0 +1,126 @@
...
+						const url = new GitHubURL(location.href);
+						url.branch = pastCommits[1];
+						location.href = url.toString();
+					});
+
+					return button;
+				})()}
+			</div>
+
+			{pastCommits.length > 2 && createDetailsButton(pastCommits)}
```
fregante, line 104:
```suggestion
			{pastCommits.length > 1 && createDetailsButton(pastCommits.slice(1))}
```


Just pass the relevant data to functions

source/features/previous-version.tsx
```diff
@@ -0,0 +1,126 @@
...
+		}
+	`);
+
+	const nodes = history.nodes as any[];
+
+	if (nodes.length === 0) {
+		return false;
+	}
+
+	return nodes.map(node => node.oid as string);
```
fregante, line 31:
```suggestion
	return nodes.slice(1).map(node => node.oid as string);
```


If the first commit is always the current commit, just slice it here so you don't have to use [1] and .slice(2) later on

fregante (commented):
Sorry we have a few more comments coming 🥲 😅

source/features/previous-version.tsx
fregante: I don’t think that's a problem, we always use it so don't worry about it

source/features/previous-version.tsx
fregante: > I can

Go on? 😅 what's the reason to go blindly back two commits?

134130:
I think my code is not refined 😅😅
Thank you for your review!
Maybe the changing will be slow since the weekend is over

source/features/previous-version.tsx
134130: I called this before observer for loading the PastCommits before page loaded.
Do we not need to cache?

source/features/previous-version.tsx
yakov116: On code that I wrote I usually remember what I did.

But your right, on a random repo more than 2 is sort of blind.

source/feature-manager.tsx
```diff
@@ -274,11 +274,13 @@ const add = async (url: string, ...loaders: FeatureLoader[]): Promise<void> => {
...
-				void setupPageLoad(id, details);
-			}
-		});
+		for (const eventType of ['turbo:render', 'soft-nav:render']) {
+			document.addEventListener(eventType, () => {
+				if (!deduplicate || !select.exists(deduplicate)) {
+					void setupPageLoad(id, details);
+				}
+			});
+		}
```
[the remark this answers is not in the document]
fregante: Don't do this here, the feature should not listen to global events on its own. You can leave the bug here because it must be fixed globally.

source/features/previous-version.tsx
```diff
@@ -0,0 +1,104 @@
+import cache from 'webext-storage-cache';
+import React from 'dom-chef';
+import * as pageDetect from 'github-url-detection';
+
+import features from '../feature-manager';
+import observe from '../helpers/selector-observer';
+import * as api from '../github-helpers/api';
+import GitHubURL from '../github-helpers/github-url';
+
+const getPastCommits = cache.function('previous-version', async (currHref: string): Promise<string[] | false> => {
```
fregante, line 10:
```suggestion
const getPastCommits = cache.function('previous-version', async (currHref: string): Promise<string[]> => {
```

source/feature-manager.tsx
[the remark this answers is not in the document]
134130: Let me know how do you planning to resolve this.
Without this change, the feature is not working correctly. (It works just only on turbo event)
Maybe The PR should be converted to draft.

fregante:
I updated the code style and removed the dropdown:

Screenshot

source/features/previous-version.tsx
```diff
@@ -7,12 +6,12 @@ import observe from '../helpers/selector-observer';
 import * as api from '../github-helpers/api';
 import GitHubURL from '../github-helpers/github-url';
 
-const getPastCommits = cache.function('previous-version', async (pathname: string): Promise<string[]> => {
```
fregante: I dropped the cache because this would create a new cache item for every page visited, recreating the entire data every time you click on "previous" anyway.

fregante:
@134130 feel free to revert 4abec33b26cba49814ecb2c0ddf03678883d431b  and apply the suggested changes. If you don't have time or are not interested, we can merge this as is.

134130:
@fregante I also agree there is no need to expaning button. And It is more refined 😎
Never mind! Thank you!

fregante:
I suppose the last question is whether it makes sense to position it there, especially just labelled as "Previous"

I'm thinking it should still:
- be placed next to the "History" button
- have a "versions" icon instead of the label, after all this does exactly what the same icon does in the blame view

https://primer.style/design/foundations/icons/versions-16



@yakov116 thoughts?

readme.md
```diff
@@ -176,6 +176,7 @@ Thanks for contributing! 🦋🙌
 - [](# "show-associated-branch-prs-on-fork") [Shows the associated pull requests on branches for forked repositories.](https://user-images.githubusercontent.com/16872793/81504659-7e5ec800-92b8-11ea-9ee6-924110e8cca1.png)
 - [](# "html-preview-link") [Adds a link to preview HTML files.](https://user-images.githubusercontent.com/44045911/67634792-48995980-f8fb-11e9-8b6a-7b57d5b12a2f.png)
 - [](# "file-age-color") [Highlights the most-recently-modified items in file lists.](https://user-images.githubusercontent.com/1402241/218314631-1442cc89-3616-40fc-abe2-9ba3d3697b6a.png)
+- [](# "previous-version") [Adds a button to switch the file to a previous version.](https://user-images.githubusercontent.com/50487467/233849156-07e41e5b-90cc-44f2-8bf3-371914290100.png)
```
fregante, line 179: This also needs to be updated **after** we finalize the UI/icon. Make sure to:

- use the light theme
- use an arrow instead of a rectangle

source/features/previous-version.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+import features from '../feature-manager';
+import observe from '../helpers/selector-observer';
+import * as api from '../github-helpers/api';
+import GitHubURL from '../github-helpers/github-url';
+
+async function getPreviousCommitForFile(pathname: string): Promise<string> {
+	const {user, repository, branch, filePath} = new GitHubURL(pathname);
+	const {resource} = await api.v4(`
+		resource(url: "/${user}/${repository}/commit/${branch}") {
```
fregante, line 12: How did you find this out? I'm trying to use v4 for another issue and can't find out to use it, if it's even possible:

https://github.com/refined-github/refined-github/blob/9d80ba25d83cf329d368a4210ed7261e24ad9657/source/features/bypass-checks.tsx#L16-L17

Example check_run_id: https://github.com/microsoft/TypeScript/pull/54006/checks?check_run_id=13000796765

fregante: Doesn't work:

```gql
{
  resource(url: "/microsoft/TypeScript/runs/13000796765") {
    resourcePath
    __typename
    ... on CheckRun {
      detailsUrl
      name
      status
    }
  }
}
```

Some random code I found online, it seems that the "ID" that GraphQL uses is the databaseId, which is different from `check_run_id`

```gql
{
  node(id: "MDg6Q2hlY2tSdW4zNDU2NjQ5MDk4") {
    __typename
    ... on CheckRun {
      resourcePath
      permalink
      externalId
      databaseId
      id
      url
    }
  }
}
```

134130: @fregante Sorry, I draw a blank. 😭 
I figured out reversly. Tried same way. but not works.
```js
{
  resource(url: "/microsoft/TypeScript/pull/54006") {
    ... on PullRequest {
      commits(first: 10) {
        nodes {
          commit {
            author {
              name
            }
            checkSuites(first: 20) {
              nodes {
                checkRuns(first: 10) {
                  nodes {
                    name
                    externalId
                    resourcePath
                  }
                }
                workflowRun {
                  runNumber
                  resourcePath
                }
              }
            }
          }
        }
      }
    }
  }
}
```

You can get resourcePath with this query.
And with the `resourcePath` should return the `CheckRun` but it doesn't.
When I call v3 API with `resourcePath`'s ID, It responses well.
```sh
gh api \
  -H "Accept: application/vnd.github+json" \
  -H "X-GitHub-Api-Version: 2022-11-28" \
  /repos/microsoft/TypeScript/check-runs/13004627613
```

Maybe the GitHub's bug. 😢

yakov116: Known issue i tried very hard too, v3 is much better with the runs

fregante: Comment updated in [`1bf6952` (#6555)](https://github.com/refined-github/refined-github/pull/6555/commits/1bf69520dd766cced310ed5f97d9d73266111def)

/OT

fregante:
@134130 by the way after this PR it's better to address small changes/bugfixes instead of adding new features, there are 50+ bugs open 😅

134130:
@fregante 😂 I really want to fix Ajax problem.
I dont have idea which don't breaks existing features.

fregante:
> @fregante 😂 I really want to fix Ajax problem.

It's a big issue and might require restructuring of the whole extension, I wouldn't worry about it now.

fregante:
Related feature:

- https://github.com/refined-github/refined-github/issues/3289

fregante (commented):
I'd love to see this merged

134130:
UI fixed.
I found same button style from primer, [ActionList](https://primer.style/react/ActionList), But I failed to find same css from not-react

source/features/previous-version.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+	if (!previousCommit) {
+		return;
+	}
+
+	const url = new GitHubURL(location.href)
+		.assign({branch: previousCommit});
+
+	const historyButtonContainer = historyButton.parentElement!;
+	historyButtonContainer.insertBefore(
```
fregante, line 40: You can use the new .before() API instead

source/features/previous-version.tsx
```diff
@@ -0,0 +1,57 @@
...
+		<a href={url.href} className="UnderlineNav-item no-wrap tooltipped tooltipped-n ml-2" aria-label="View prior to this change">
+			<VersionsIcon className="UnderlineNav-octicon" style={{marginRight: 0}}/>
+		</a>,
+		historyButton,
+	);
+}
+
+async function init(signal: AbortSignal): Promise<void> {
+	observe('a[aria-label=History].react-last-commit-history-group', add, {signal});
```
fregante, line 49:
```suggestion
	observe('a[aria-label="History"].react-last-commit-history-group', add, {signal});
```

source/features/previous-version.tsx
```diff
@@ -0,0 +1,57 @@
...
+		return;
+	}
+
+	const url = new GitHubURL(location.href)
+		.assign({branch: previousCommit});
+
+	const historyButtonContainer = historyButton.parentElement!;
+	historyButtonContainer.insertBefore(
+		<a href={url.href} className="UnderlineNav-item no-wrap tooltipped tooltipped-n ml-2" aria-label="View prior to this change">
+			<VersionsIcon className="UnderlineNav-octicon" style={{marginRight: 0}}/>
```
fregante, line 42:
```suggestion
			<VersionsIcon className="UnderlineNav-octicon mr-0"/>
```

source/features/previous-version.tsx
```diff
@@ -0,0 +1,57 @@
...
+	if (!previousCommit) {
+		return;
+	}
+
+	const url = new GitHubURL(location.href)
+		.assign({branch: previousCommit});
+
+	const historyButtonContainer = historyButton.parentElement!;
+	historyButtonContainer.insertBefore(
+		<a href={url.href} className="UnderlineNav-item no-wrap tooltipped tooltipped-n ml-2" aria-label="View prior to this change">
```
fregante, line 41: I don’t think “no-wrap” does anything here. 

The label should probably be:

```suggestion
		<a href={url.href} className="UnderlineNav-item tooltipped tooltipped-n ml-2" aria-label="View previous version">
```

fregante (approved):

fregante:
고마워지환

PR #6552, opened 2023-04-24
Fix navigation issues in file explorer

### LINKED ISSUE #6414: `download-folder-button` duplicated
### Steps to reproduce:

1. Open folder
2. Open file
3. Open folder again

ezgif-1-c1e5a66abb

The bug happens after clearing extension cache. 

https://github.com/refined-github/refined-github/tree/main/source/features

### Extension version

23.3.2

### Browser(s) used

Firefox 110

134130:
## Description
- Closes #6414
- Reproducing route
	- Enter a file tree of repository.
	- Enter some file of the tree.
	- Pop to back page (I don't know the correct word of this)
	- Enter some file of the tree.

## Test URLs
https://github.com/refined-github/refined-github/tree/main/source/features

### Before
The button is duplicating and get pushed out to left
image

### After
Remove previous button and stay the button right position.
Also, create button when accessing not only `repo-tree` but also `single-file`
image

### DIFF
--- a/source/feature-manager.tsx
+++ b/source/feature-manager.tsx
@@ -274,11 +274,13 @@ const add = async (url: string, ...loaders: FeatureLoader[]): Promise<void> => {
 			void setupPageLoad(id, details);
 		}
 
-		document.addEventListener('turbo:render', () => {
-			if (!deduplicate || !select.exists(deduplicate)) {
-				void setupPageLoad(id, details);
-			}
-		});
+		for (const eventType of ['turbo:render', 'soft-nav:render']) {
+			document.addEventListener(eventType, () => {
+				if (!deduplicate || !select.exists(deduplicate)) {
+					void setupPageLoad(id, details);
+				}
+			});
+		}
 	}
 };
 
--- a/source/features/download-folder-button.tsx
+++ b/source/features/download-folder-button.tsx
@@ -8,7 +8,7 @@ import features from '../feature-manager';
 import observe from '../helpers/selector-observer';
 
 function add(folderDropdown: HTMLElement): void {
-	folderDropdown.parentElement!.querySelector('a.rgh-download-folder-button')?.remove()
+	folderDropdown.parentElement!.querySelector('a.rgh-download-folder-button')?.remove();
 
 	const downloadUrl = new URL('https://download-directory.github.io/');
 	downloadUrl.searchParams.set('url', location.href);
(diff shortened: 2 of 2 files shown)


source/feature-manager.tsx
```diff
@@ -274,11 +274,13 @@ const add = async (url: string, ...loaders: FeatureLoader[]): Promise<void> => {
 			void setupPageLoad(id, details);
 		}
 
-		document.addEventListener('turbo:render', () => {
-			if (!deduplicate || !select.exists(deduplicate)) {
-				void setupPageLoad(id, details);
-			}
-		});
+		for (const eventType of ['turbo:render', 'soft-nav:render']) {
```
fregante, line 277: Both of these events are called when clicking one of the repos tabs, for example when you click on Issues. This means you're running every feature twice on every page load.

Also, neither one is called when you:

1. Open https://github.com/refined-github/refined-github/tree/main/.github/workflows
2. Go up one folder
3. Go back in browser history ❌ 
4. Go forward in browser history ❌

source/features/download-folder-button.tsx
```diff
@@ -24,14 +26,16 @@ function add(folderDropdown: HTMLElement): void {
...
+		'[title="More options"]',	// PageDetect.isRepoTree
+		'[title="More file actions"]', // PageDetect.isSingleFile
 		'[aria-label="Add file"] + details', // TODO: Drop in mid 2023. Old file view #6154
 	], add, {signal});
 }
 
 void features.add(import.meta.url, {
 	include: [
 		pageDetect.isRepoTree,
+		pageDetect.isSingleFile,
```
fregante, line 38: This is a hack, a feature should not be enabled on random pages just to cleanup after itself

fregante (changes requested):
Unfortunately it's not that easy. The issue is not strictly related to `download-folder-button` so, if you want to look into it, you'd have to make sure that the logic in the feature manager is correct.

Another feature affected is:

- https://github.com/refined-github/refined-github/issues/6528#issuecomment-1509866866

source/feature-manager.tsx
fregante: > I might push a fix specific to this feature

I "fixed the duplication" by using `[title="More options"]:not(:has(+ .rgh-download-folder-button))`, but this means that the old button is left in place, pointing to the old folder. 🥲 

Let's move the discussion to:

- https://github.com/refined-github/refined-github/issues/6554

fregante:
I'm closing this PR because nothing can be used here unfortunately. If you'd like, maybe you can verify whether we can safely **replace** `turbo:render` with `soft-nav:render` **everywhere**` in our extension

source/features/download-folder-button.tsx
134130: Yes, but add singleFile is needed.
When we access blob direct link, Download button will not be rendered.

fregante: This feature must appear exclusively on directories because it _downloads directories_ and the `include` array tells me exactly that.

It's only "needed" here because the feature manager is broken, but it's still a hack that we should not use

fregante:
I'm reopening this in case you want to keep experimenting or proposing new AJAX solutions

fregante:
Closing for now. If you have any suggestions let me know

Issue #5932, opened 2022-08-25
`git-checkout-pr` giving wrong code


Sparticuz:
### Description

The code given on the `Checkout with Git` http and ssh boxes is incorrect. The third line should read `git switch --track upstream/zb140/refactor-compose-files`. Upstream instead of the repo owners name because in line one we added upstream as the name of the remote. (Or line one should be changed to be the repo owners name instead of upstream)
```
git remote add upstream git@github.com:games-on-whales/gow.git
git fetch upstream zb140/refactor-compose-files
git switch --track games-on-whales/zb140/refactor-compose-files
```

![image](https://user-images.githubusercontent.com/161913/186670472-85651837-8828-4f9b-a7c5-4fa852642d2c.png)

### How to replicate the issue

Open a PR and click the button to pop out the `git-checkout-pr` box

### Extension version

22.8.23

### Browser(s) used

Chrome 104

fregante:
I don't remember the logic exactly, but basically it depends on a few things:

- whether the PR is yours 
- whether it's on a fork

There's a whole matrix of expectations for this. Please report the exact URL you found this problem in, as suggested by the issue template.

yakov116:
@Sparticuz if I am not mistaken there is reason we wrote it like this. Since if upstream exists the first like will error so we need to write it again on the second line.

Unless I am not understanding your issue. Please post a url as stated in the issue template.

Sparticuz:
Right, what I'm saying is the third line fails because, in this case, there is not a remote named `games-on-whales`, the remote was named `upstream` in line 1

Sparticuz:
```
kyle@bently 10:38:33 ~ $ git clone git@github.com:Sparticuz/gow.git gow-test
Cloning into 'gow-test'...
remote: Enumerating objects: 974, done.
remote: Counting objects: 100% (188/188), done.
remote: Compressing objects: 100% (119/119), done.
remote: Total 974 (delta 82), reused 129 (delta 65), pack-reused 786
Receiving objects: 100% (974/974), 11.14 MiB | 12.12 MiB/s, done.
Resolving deltas: 100% (446/446), done.
kyle@bently 10:38:43 ~ $ cd gow
gow/      gow_home/ gow.old/  gow-test/ 
kyle@bently 10:38:43 ~ $ cd gow-test/
kyle@bently 10:38:49 ~/gow-test on master $ git remote add upstream git@github.com:games-on-whales/gow.git
[...]

PR #6561, opened 2023-04-25
Drop `git-checkout-pr` feature

fregante:
Dropping it because:

- the logic behind it is [complex](https://github.com/refined-github/refined-github/pull/3596#issuecomment-720910840) and it depends on the user’s local setup
-  #5932
- it doesn't work in some situations: https://github.com/refined-github/refined-github/issues/4701
- GitHub already offers a better solution to it, via `gh-cli`
- GitHub already fixed the reason why it was added (cross-repo PRs checked out via `gh` now have push access via successive `git` commands)

closes https://github.com/refined-github/refined-github/issues/5932
closes https://github.com/refined-github/refined-github/issues/4701


yakov116 (approved):

gengjiawen:
It's so usefully when I debug. Can we add it back and not turn it on by default. Gh not available in most cloud dev env, and you can't use it without a token.

In my use case, I am a Node.js collaborator, run into different compile issues on different platform. I use this every time when debuging.

PR #6565, opened 2023-04-25
Enable `fit-textareas` sooner, don't wait for focus

fregante:
Testable here. It works.

Ideal testing:

1. Find a tall message you wrote
2. Click edit

### DIFF
--- a/source/features/fit-textareas.tsx
+++ b/source/features/fit-textareas.tsx
@@ -1,41 +1,29 @@
 import './fit-textareas.css';
-import select from 'select-dom';
-import delegate, {DelegateEvent} from 'delegate-it';
 import {isSafari} from 'webext-detect-page';
 import fitTextarea from 'fit-textarea';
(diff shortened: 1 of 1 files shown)


source/features/fit-textareas.tsx
```diff
@@ -1,41 +1,27 @@
...
-}
-
 function init(signal: AbortSignal): void {
 	// Exclude PR review box because it's in a `position:fixed` container; The scroll HAS to appear within the fixed element.
-	delegate('textarea:not(#pull_request_review_body)', 'focusin', focusListener, {signal});
-
-	for (const textArea of select.all('textarea')) {
-		watchTextarea(textArea);
-	}
+	observe('textarea:not(#pull_request_review_body)', watchTextarea, {signal});
```
fregante, line 24: Instead of enabling textareas on focus, we do so when they become visible. No weird listeners necessary

PR #6572, opened 2023-04-26
Highlight RGH features on the page

fregante:
I thought about this for a while, this can help us see through Refined GitHub a bit. It's not perfect but I don’t think we should focus too much on ensuring every generated/altered element has an `rgh` class.

This would be hidden behind a "debugging" toggle.

Screenshot 9

### DIFF
--- a/source/refined-github.css
+++ b/source/refined-github.css
@@ -192,3 +192,13 @@ div[data-target='readme-toc.content'] div.blob-wrapper img {
 .rgh-bg-none {
 	background: none !important;
 }
+
+[class*="rgh-"]:not([class*="rgh-seen"]) {
+	outline: solid 3px violet;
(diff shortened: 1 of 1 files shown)


fregante:
Could be too much already 😂 

Screenshot 10

Screenshot 11

fregante:
Not worth the extra logic. Just paste this CSS in your options:

```css
/* Highlight RGH features */
[class*="rgh-"]:not([class*="rgh-seen"]) {
	outline: solid 3px var(--color-scale-red-3);
	outline-offset: 3px;
}

/* Highlight watched elements (not advised, noisy) */
[class*="rgh-seen"] {
	outline: solid 3px var(--color-scale-orange-1);
	outline-offset: 3px;
}
```

PR #6575, opened 2023-04-26
Drop `latest-tag-button` feature

fregante:
- Applies https://github.com/refined-github/refined-github/issues/6250
- Close https://github.com/refined-github/refined-github/issues/6553
- Close https://github.com/refined-github/refined-github/issues/6494
- Close https://github.com/refined-github/refined-github/issues/6571
- Gives way for https://github.com/refined-github/refined-github/issues/6499
- Followed by #6576

## Note

Refer to the discussion in https://github.com/refined-github/refined-github/issues/6250 for the extended reasoning. In short:

- `latest-tag-button` provides some information that isn't necessarily useful "on every page" and allows an action that probably isn't too common (visiting the latest published version of the current view)

Part of the feature is still available natively via the branch selector and via the upcoming `unreleased-commits` feature (#6576), which inherits much of the logic, but is only shown on the default branch.


fregante:
If anyone “wants this feature back,” please take a minute to explain what you’re missing from it exactly and how you use it in your workflow.

PR #6583, opened 2023-04-26
Disable `ci-link` on new UI

fregante:
- Fixes https://github.com/refined-github/refined-github/issues/6453

The feature currently only works on the repo root, which is exactly where it isn't needed. See screenshot.

## Test URLs

## Screenshot

Screenshot 13


fregante:
Old UI not affected

Screenshot

PR #6592, opened 2023-04-29
Automatically set popup to contain options on iOS

### LINKED ISSUE #6587: Inconsistent options popup code in Safari
### Description

The "browser action" in iOS opens the options page. This change is done manually in the manifest before every deployment, but it can be automated in background.js:

- if mobile Safari is detected, call [browser.browserAction.setPopup()](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/action/setPopup)

There’s a chance that we’ll also have to support more browser in this condition:

- https://github.com/refined-github/refined-github/issues/6591

Using the same logic, this section should be hidden in the options page since it’s no longer applicable:

IMG_7861 2

fregante:
- Includes https://github.com/fregante/webext-detect-page/releases/tag/v4.1.0
- Will fix #6587

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -43,7 +43,7 @@
 				"twas": "^2.1.3",
 				"webext-base-css": "^1.4.3",
 				"webext-detect-page": "^4.1.0",
-				"webext-domain-permission-toggle": "^4.0.0-0",
+				"webext-domain-permission-toggle": "^4.0.1",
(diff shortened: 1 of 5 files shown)


fregante:
Confirmed, the "Button link" section no longer appears:

Screenshot

PR #6596, opened 2023-05-01
Fix `restore-file`: restores based on user's current diff

### LINKED ISSUE #4679: `restore-file` occasionally restores the wrong commit

134130:
## Description
- Closes #4679
- Previously, `restore-file` restores file to the base of the Pull Request.
	- So when the Base's HEAD is changed after PR has been opened, The restoring looks oddly
- But GitHub's UI shows file's diff from the commit, commit-range and pull-request-totally (all-commits).
- And we usually uses `git restore` for restoring a file to "previous status", not "untouched status".
- Let's restore file to diff's original.

## Test URLs
- Full files diff: https://github.com/134130/refined-github-test/pull/1/files
- Range commit diff: https://github.com/134130/refined-github-test/pull/1/files/15f7fb6..5deabc1
- Single commit diff: /134130/refined-github-test/pull/1/commits/7cd9c13

Maybe the above url's cannot work. Check `Range commit diff` and `Single commit diff` with this screenshot
image

## Screenshot
Same as before.


source/features/restore-file.tsx
```diff
@@ -7,21 +8,45 @@ import * as pageDetect from 'github-url-detection';
...
-			pullRequest(number: ${getConversationNumber()!}) {
-				baseRefOid
-			}
-		}
-	`);
-	return repository.pullRequest.baseRefOid;
+	const {base} = getBranches();
+	const {baseRefOid} = await getPrInfo(base.relative);
+	return baseRefOid;
 });
```
134130, line 19: I didn't used `onetime` on `headRefOid` of the Pull Request since `headRefOid` keeps changing when we restoring files continuously.

fregante:
Maybe I'm confused, but "Restore file" should not depend on the current URL.

The intent of the button is to completely remove the file from the PR, does this happen with this code?

Examples to test:

- change A.js in a PR, change A.js on the `main` branch, click "Update branch", click restore
- The same, but use a "Rebase merge" in the branch update

Live example:

- https://github.com/refined-github/sandbox/pull/29/files

134130:
Oops if the intend of the feature is "completely remove the file from the PR", This PR is not correct.
I though the "restore" means "git restore" and We need to restore file relative on commits not relative on a full pull request.

You can remove file from the PR as you did before, on `Full files diff` page.

I intended restore works like git.

Example situation: 
1. The PR has 5 commits.
2. first commit is full refactoring commit, and the last commit is adding some notes.
3. You thinks the refactoring is correct, but the adding notes is not needed.
4. So you want to drop adding notes commit.

Also, when we use restore button on this page, How do you expect the restore to behave?
image

Before: restore to just 2 lines state. since removing the file from PR
On this PR: 6 and 7 lines are added. restore to 5 lines state

---

Examples to test:
- change A.js in a PR, change A.js on the main branch, click "Update branch", click restore
  - The whole file will be restored to where the PR branch is created.
  - So the change A.js on the main branch will not be reflected
- The same, but use a "Rebase merge" in the branch update
  - The whole file will be restored to where the PR branch is created.
  - So the change A.js on the main branch will be reflected

yakov116:
I did not read your entire post, this feature restores the as it was prior to the PR was submitted.

134130:
@yakov116 Then I think we need to hide restore button on showing commit's different page since the UX is different.
We need to change the button only shows on full file changes since the file is restored to before PR submitted.

134130:
I'm not arguing for it, just for a slightly better UX. If that was the original intent, I'm willing to fix it, and it's very easy.

fregante:
> I think we need to hide restore button on showing commit's different page

All the other buttons are still available there and refer to “the file in the PR”, not “the file in the state shown in this URL.” Even the Edit button is shown.

fregante:
Maybe we can make this more clear by changing the label to “Drop from PR” or “Discard changes” like GitHub Desktop calls it

134130:
Decided on the wording, I'll rework it to drop.

source/features/restore-file.tsx
```diff
@@ -9,18 +9,14 @@ import * as api from '../github-helpers/api';
 import showToast from '../github-helpers/toast';
 import {getConversationNumber} from '../github-helpers';
 import {getBranches} from '../github-helpers/pr-branches';
+import getPrInfo from '../github-helpers/get-pr-info';
 
 // Get the current base commit of this PR. It should change after rebases and merges in this PR.
 // This value is not consistently available on the page (appears in `/files` but not when only 1 commit is selected)
 const getBaseReference = onetime(async (): Promise<string> => {
```
fregante, line 16: We can drop `onetime` and the `This value is not consistently…` line

fregante (approved):
The changes look good on paper. Just make sure it works and update the PR title to reflect the changes (it's no longer a bugfix I think)

source/features/restore-file.tsx
```diff
@@ -127,7 +123,7 @@ function handleMenuOpening({delegateTarget: dropdown}: DelegateEvent): void {
 			role="menuitem"
 			type="button"
 		>
-			Restore file
+			Drop from PR
```
yakov116, line 126: I am not sure i like this wording, we "restoring the file" to how it was prior to the pr.

fregante: Yeah I suggested a change because "Restore" isn't really clear what we're restoring to.

134130:
@fregante 
"Revert file to before PR" looks pretty good.
But It seems a little long for feature's title.
What do you think of it?

While working with `drop-file-from-pr`, I found "Adds a button to revert all the changes to a file in a PR" description for the feature.

fregante:
Yes we used to call it "Revert file" but that was just as ambiguous. That label is also too long.

source/github-helpers/get-pr-info.ts
```diff
@@ -2,7 +2,6 @@ import * as api from './api';
 import {getConversationNumber} from '.';
 
 export type PullRequestInfo = {
-	// TODO: Use this for `restore-file`
```
yakov116, line 5: When we added this line of code we needed to wait for something on GHE. @fregante are we good to drop this now?

fregante:
@134130 no need to rename features, renaming is a last resort that breaks our search. Just change the label here

readme.md
```diff
@@ -109,8 +109,8 @@ GitHub Enterprise is also supported: [How to enable it](https://fregante.github.
 			<p><img src="https://user-images.githubusercontent.com/1402241/35192861-3f4a1bf6-fecc-11e7-8b9f-35ee019c6cdf.gif">
 	<tr>
 		<th width="50%">
-			<p><a title="restore-file"></a> Adds a button to revert all the changes to a file in a PR
-			<p><img src="https://user-images.githubusercontent.com/1402241/130660479-083e91e6-0778-446a-9aaf-b9b3e7214281.gif">
+			<p><a title="drop-file-from-pr"></a> Adds a button to revert all the changes to a file in a PR
+			<p><img src="https://user-images.githubusercontent.com/50487467/236245182-afc97200-2634-4f82-9e9c-1f605fb04565.gif">
```
fregante, line 113: Thanks for the gif update!

source/features/restore-file.tsx
```diff
@@ -48,17 +42,17 @@ async function getFile(filePath: string): Promise<{isTruncated: boolean; text: s
...
 
-async function restoreFile(progress: (message: string) => void, menuItem: Element, filePath: string): Promise<void> {
+async function dropFile(progress: (message: string) => void, menuItem: Element, filePath: string): Promise<void> {
 	const file = await getFile(filePath);
 
 	if (!file) {
 		// The file was created by this PR.
 		// This code won’t be reached if `highlight-deleted-and-added-files-in-diffs` works.
-		throw new Error('Nothing to restore. Delete file instead');
+		throw new Error('Nothing to drop. Delete file instead');
```
FloEdelmann, line 51: That error message seems wrong now: There is something to drop (the whole new file), but nothing to restore. When does this error occur?

Suggestions for better wording:

* File was created in this PR. Delete it instead.
* Can't drop created file. Delete it instead.

FloEdelmann (changes requested):

134130:
@fregante It also fixes #4679 since we change getBaseRef function.
When the base branch is changed after the PR is opened, We've got the reference of base branch not the PR 's head commit.
Now, We can get the PR's head commit correctlly

fregante:
I does not. The underlying API is unchanged from what I see

PR #6597, opened 2023-05-01
Use Parcel instead of Webpack

fregante:
It builds.

Screenshot 4
Screenshot 5
Screenshot 6


fregante:
Next:

- [ ] Update `verify-features`
- [ ] Try [`link:` in package.json,](https://parceljs.org/features/plugins/#the-link%3A-protocol) 
- [ ] Allow `link:` in CI Security

source/refined-github.css
```diff
@@ -1,3 +1,26 @@
+@import './refined-github.css';
```
fregante, line 1: Oops. Recursive import

I also need to think of a way to extract the raw CSS below into another file because this change changes the cascade order.

fregante:
This isn't a big deal to Parcel devs, but it renders our builds unreadable. The solution is to either disable scope hosting, which balloons the  bundle to 1.2MB or to mangle the names, which means our clean `addConversationBanner` becomes `a`, an issue also visible in error stacks. We can't ship sourcemaps.

- https://github.com/parcel-bundler/parcel/issues/6092

I can try again in 2024, but at this point the output is subpar. I'll extract some improvements from this PR though.

PR #6612, opened 2023-05-04
Meta: Use dynamic file list in Xcode

fregante:
Xcode has static project file lists so if our files are dynamically created one needs to open Xcode and update the project.

This seems to add a link to the parent `distribution` folder so that only _one_ item needs to be static. 

Inspired by:

- https://github.com/refined-github/refined-github/pull/6597


sindresorhus:
I think we already tried symlink before. The problem is that it copies the whole `distribution` folder into `Resources` instead of the contents of the folder. So you end up with:

```
Refined GitHub Extension.appex/Contents/Resources/distribution/manifest.json
```

I think `manifest.json` has to be directly inside `Resources`.

fregante:
Can you try building and running it in Safari macOS? It seems to work correctly in iOS (I only manage to run the dev version on macOS if I uninstall RGH completely)

fregante:
Well, this works on iOS somehow but it crashes Safari macOS.

I'll move the assets to `Resources/assets` instead so that just `asset` can be a link. However I'd still ask you to try to make `Resources` a symlink to `distribution` instead

sindresorhus:
I don't think `Resources` can be a symlink. Then we would have done that last time. I think the only other option is to go back to the script again (which has downsides I forgot): https://github.com/refined-github/refined-github/commit/9036cceec3455f275a5bd3cca8064d73309195dd#diff-f6596608b09878fdc971f0f15ca2f2b4fc694cf71f1908ddc5564994f9528706R242

fregante:
The downsides of the scripts are that it doesn’t watch the FS so developing the extension in Safari means having to build it manually every time. Extension testing seems already broken compared to the default setup since Safari can’t see the extension until I uninstall the App Store version, usually the dev version automatically replaces it immediately.

safari/Refined GitHub.xcodeproj/project.pbxproj
```diff
@@ -7,6 +7,8 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		1F9945322A07DB5E005BC236 /* resources in Resources */ = {isa = PBXBuildFile; fileRef = 1F9945312A07DB5E005BC236 /* resources */; };
+		1F9945332A07DB5E005BC236 /* resources in Resources */ = {isa = PBXBuildFile; fileRef = 1F9945312A07DB5E005BC236 /* resources */; };
```
fregante, line 11: Since `distribution` can't be changed, I added `distribution/resources` as a symlink:

Screenshot 3

fregante (commented):
Tested:

- content script
- options page
- background page (via `open-all-notifications`)
- `resolve-conflicts` on https://github.com/refined-github/sandbox/pull/14/conflicts

However today iOS decided to fail loading here and on main 🤷‍♂️ 

Screenshot 4

sindresorhus:
> However today iOS decided to fail building here and on main 🤷‍♂️

I'm getting this too. If you look at the full log, this is the problem:

> /Users/sindresorhus/Library/Developer/Xcode/DerivedData/Refined_GitHub-dshkgxrkdsfhtxedicouzyfghibo/Build/Products/Debug-iphonesimulator/Refined GitHub Extension.appex: code object is not signed at all
> In subcomponent: /Users/sindresorhus/Library/Developer/Xcode/DerivedData/Refined_GitHub-dshkgxrkdsfhtxedicouzyfghibo/Build/Products/Debug-iphonesimulator/Refined GitHub Extension.appex/icon.png

sindresorhus:
Fixed in `main`.

fregante:
codesign worked for me eventually, I had to reset my keys in the keychain etc etc, hundreds of people complaining about that issue for years on Stackoverflow.

One also mentioned the possible incompatibility with folders called `resources`, I wonder if this PR triggered that.

sindresorhus:
Hmm, actually no. Now it complains about the manifest file (`main` builds fine though):

> /Users/sindresorhus/Library/Developer/Xcode/DerivedData/Refined_GitHub-dshkgxrkdsfhtxedicouzyfghibo/Build/Products/Debug-iphonesimulator/Refined GitHub Extension.appex: code object is not signed at all
> In subcomponent: /Users/sindresorhus/Library/Developer/Xcode/DerivedData/Refined_GitHub-dshkgxrkdsfhtxedicouzyfghibo/Build/Products/Debug-iphonesimulator/Refined GitHub Extension.appex/manifest.json
Command CodeSign failed with a nonzero exit code

sindresorhus:
The extension contains:

```
resources
Info.plist
manifest.json
Refined GitHub Extension
```

However, a proper extension has all the resources in a `Resources` directory. `manifest.json` cannot be at the top-level. It should also be `Resources`, not `resources` (unclear if that matters though).

fregante:
Double-check, this PR only moves the assets one level deep. Manifest.json is unchanged in Resources/manifest.json

I stupidly named this subfolder `resources` though so you should be seeing `Resources/resources/background.js`

safari/Refined GitHub.xcodeproj/project.pbxproj
```diff
@@ -31,26 +33,8 @@
 		E35EB4F226F36E0900DDEA60 /* Assets.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = E35EB4B026F36E0900DDEA60 /* Assets.xcassets */; };
 		E35EB4F326F36E0900DDEA60 /* SafariWebExtensionHandler.swift in Sources */ = {isa = PBXBuildFile; fileRef = E35EB4B226F36E0900DDEA60 /* SafariWebExtensionHandler.swift */; };
 		E35EB4F426F36E0900DDEA60 /* SafariWebExtensionHandler.swift in Sources */ = {isa = PBXBuildFile; fileRef = E35EB4B226F36E0900DDEA60 /* SafariWebExtensionHandler.swift */; };
-		E3919AF727E61192009C4956 /* refined-github.js.LICENSE.txt in Resources */ = {isa = PBXBuildFile; fileRef = E3919AE627E61192009C4956 /* refined-github.js.LICENSE.txt */; };
-		E3919AF827E61192009C4956 /* refined-github.js.LICENSE.txt in Resources */ = {isa = PBXBuildFile; fileRef = E3919AE627E61192009C4956 /* refined-github.js.LICENSE.txt */; };
 		E3919AF927E61192009C4956 /* manifest.json in Resources */ = {isa = PBXBuildFile; fileRef = E3919AE727E61192009C4956 /* manifest.json */; };
 		E3919AFA27E61192009C4956 /* manifest.json in Resources */ = {isa = PBXBuildFile; fileRef = E3919AE727E61192009C4956 /* manifest.json */; };
```
fregante, line 37: Unchanged

safari/Refined GitHub.xcodeproj/project.pbxproj
```diff
@@ -264,16 +240,8 @@
 		E3919AC127E6107E009C4956 /* Resources */ = {
 			isa = PBXGroup;
 			children = (
-				E3919AEA27E61192009C4956 /* background.js */,
-				E3919AED27E61192009C4956 /* icon.png */,
+				1F9945312A07DB5E005BC236 /* resources */,
 				E3919AE727E61192009C4956 /* manifest.json */,
```
fregante, line 244: - Resources/resources/*
- Resources/manifest.json

safari/Refined GitHub.xcodeproj/project.pbxproj
```diff
@@ -264,16 +240,8 @@
 		E3919AC127E6107E009C4956 /* Resources */ = {
 			isa = PBXGroup;
 			children = (
-				E3919AEA27E61192009C4956 /* background.js */,
-				E3919AED27E61192009C4956 /* icon.png */,
+				1F9945312A07DB5E005BC236 /* assets */,
```
fregante, line 243: 1. I checked out this PR
2. Got `codesign` error again
3. Renamed this subfolder to `assets`
4. It builds now, both iOS and macOS

Screenshot 7

sindresorhus:
It builds now and the extension seems to work, but the options page is blank (tested on device and simulator):

Simulator Screenshot - iPhone 14 Pro Max - 2023-05-14 at 16 40 24

fregante:
Thanks for checking! Fixed

PR #6614, opened 2023-05-04
Lint (devDependencies)

fregante:
Extracting some changes from

- https://github.com/refined-github/refined-github/pull/6597

Continues:

- https://github.com/refined-github/refined-github/pull/6613


yakov116:
If you dropped ts-node I don't think we need cross-env

PR #6621, opened 2023-05-06
`restore-file` UX improvements

fregante:
- Continues #6596 

## Changes

- Allow deletion via API
- "Drop" was a bit awkward, I replaced with with "Remove from PR"

## Test URLs

https://github.com/refined-github/sandbox/pull/16/files

### GitHub Desktop

Screenshot 9

### VS Code

Screenshot 10

cc @134130

### DIFF
--- a/source/features/restore-file.tsx
+++ b/source/features/restore-file.tsx
@@ -115,7 +115,7 @@ function handleMenuOpening({delegateTarget: dropdown}: DelegateEvent): void {
 	}
 
 	if (editFile.closest('.file-header')!.querySelector('[aria-label="File added"]')) {
-		// The file is new. "Removing from PR" it means deleting it, which is already possible.
+		// The file is new. "Discarding changes" means deleting it, which is already possible.
 		// Depends on `highlight-deleted-and-added-files-in-diffs`.
 		return;
 	}
--- a/source/features/restore-file.tsx
+++ b/source/features/restore-file.tsx
@@ -42,23 +42,34 @@ async function getFile(filePath: string): Promise<{isTruncated: boolean; text: s
 	return repository.file;
 }
 
-async function dropFile(progress: (message: string) => void, menuItem: Element, filePath: string): Promise<void> {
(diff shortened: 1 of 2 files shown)


source/features/restore-file.tsx
```diff
@@ -42,23 +42,34 @@ async function getFile(filePath: string): Promise<{isTruncated: boolean; text: s
...
 }
 
-async function dropFile(progress: (message: string) => void, menuItem: Element, filePath: string): Promise<void> {
+async function discardChanges(progress: (message: string) => void, filePath: string): Promise<void> {
 	const file = await getFile(filePath);
 
-	if (!file) {
-		// The file was created by this PR.
-		// This code won’t be reached if `highlight-deleted-and-added-files-in-diffs` works.
-		throw new Error('Nothing to drop. Delete file instead');
```
fregante, line 51: This is now possible. We had only dropped this part because we used to load the native "delete" page and act on that. Now it's much easier via API

source/features/restore-file.tsx
```diff
@@ -42,23 +42,34 @@ async function getFile(filePath: string): Promise<{isTruncated: boolean; text: s
...
+	if (file?.isTruncated) {
+		throw new Error('File too big, you’ll have to use git');
```
fregante, line 49: Improved error message, including solution

source/features/restore-file.tsx
```diff
@@ -42,23 +42,34 @@ async function getFile(filePath: string): Promise<{isTruncated: boolean; text: s
...
+			{
+				path: "${filePath}",
+				contents: "${btoa(unescape(encodeURIComponent(file.text)))}"
+			}
+		]
+	`;
 
 	const {nameWithOwner, branch: prBranch} = getBranches().head;
-	progress(menuItem.closest('[data-file-deleted="true"]') ? 'Undeleting…' : 'Committing…');
+	progress('Committing…');
```
fregante, line 71: It's not necessary to differentiate the status here. In every case we're "Committing…"

yakov116:
Much better

PR #6623, opened 2023-05-06
Fix `open-all-notifications` error display

fregante:
- Fixes https://github.com/refined-github/refined-github/issues/6590

I need to verify that it actually throws an error, because I think it doesn't

### DIFF
--- a/source/helpers/open-tabs.ts
+++ b/source/helpers/open-tabs.ts
@@ -1,10 +1,18 @@
+import showToast from '../github-helpers/toast.js';
+import pluralize from '../helpers/pluralize.js';
+
 export default async function openTabs(urls: string[]): Promise<boolean> {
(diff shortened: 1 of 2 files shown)


source/helpers/open-tabs.ts
```diff
@@ -1,11 +1,19 @@
...
 
-	await browser.runtime.sendMessage({
+	const response = browser.runtime.sendMessage({
 		openUrls: urls,
 	});
 
+	await showToast(response, {
+		message: 'Opening…',
+		doneMessage: pluralize(urls.length, '$$ tab') + ' opened',
+	});
```
fregante, line 16: The toast is now shown only if the user confirms the opening.

Also it acts as a notification on mobile, where opening tabs isn't as visible

https://github.com/refined-github/refined-github/assets/1402241/7406c973-6d51-43ad-baf1-08df23611d2f

PR #6646, opened 2023-05-12
Add pr-files-commit-range-not-found feature

mheidinger:
This adds a new feature that automatically redirects the user from a "Commit range not found" PR files page to the latest changes.
To check for whether no commit range was found, the existence of a link in a `blankslate` div is checked. If it exists, it is clicked to redirect to the latest changes of this PR.

Relates to the following suggestion: https://github.com/refined-github/refined-github/issues/4648#issuecomment-921957276

## Test URLs
URL for which a redirect should happen: https://github.com/refined-github/refined-github/pull/6646/files/abc..def
"Good" URLs which should not redirect:
- https://github.com/refined-github/refined-github/pull/6646/files
- https://github.com/refined-github/refined-github/pull/6488/files/274865d4a0ecea7039a57383337b0e9a3bc72bca..5be67c6d9d698f59c79d513ed0e1acfdae0459be

## Screenshot
pr-files-commit-range-not-found

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -275,6 +275,7 @@ Thanks for contributing! 🦋🙌
 - [](# "view-last-pr-deployment") [Adds a link to open the latest deployment from the header of a PR.](https://user-images.githubusercontent.com/44045911/232313171-b54ac9cc-ebb1-43ef-bd41-5d81ec9f9588.png)
 - [](# "no-unnecessary-split-diff-view") [Always uses unified diffs on files where split diffs aren’t useful.](https://user-images.githubusercontent.com/46634000/121495005-89af8600-c9d9-11eb-822d-77e0b987e3b1.png)
 - [](# "emphasize-draft-pr-label") [Makes it easier to distinguish draft PR in lists.](https://user-images.githubusercontent.com/1402241/218252438-062a1ab3-4437-436d-9140-87bee23aaefb.png)
+- [](# "pr-files-commit-range-not-found") [Redirect to all changes if commit range not found.](https://user-images.githubusercontent.com/8899636/237991799-2fec2f38-7311-4e7d-9050-9449f75a5317.gif)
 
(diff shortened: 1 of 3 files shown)


fregante:
I’ll be honest, I’d probably prefer redirecting PR notifications to the Conversation tab instead, before the user even lands there. 

The current setup and it causes a “strange automatic reload,” which isn’t great UX, and breaks the back button after the redirect.

mheidinger:
> I’ll be honest, I’d probably prefer redirecting PR notifications to the Conversation tab instead, before the user even lands there.
> 
> The current setup causes a “strange automatic reload,” which isn’t great UX, and breaks the back button after the redirect.

I agree on the not so great UX but unfortunately I think my use-case can only be fixed like that: Because bot PRs spam my notifications (unrelated problem), I only use the GitHub emails where I have filters setup to sort out everything which does not require my attention. When I click a link from an email I can only catch this scenario on page load.

> Also don’t force-push here, commits are squashed on PR merge. Force pushes just make it difficult to review changes, among other issues such as the one you’re trying to fix here.

Guess here lies the original "problem" with people using Git differently and us relying a lot on a clean history which requires force pushes. Because of this I often run into the problem that I try to fix with this PR.

fregante:
The good news is that this behavior can be implemented via userscript in just a line (and appropriate URL filter)

```js
document.querySelector('.blankslate > p > a')?.click();
```

The bad news is that breaking the back button and causing awkward reloads aren’t a good fit for Refined GitHub, so I’ll have to close this PR.

fregante:
I’ve been hating this behavior of GitHub since this PR. I’d accept a PR to simply drop `/files` from the links on the notifications page so they always automatically point to the Conversation tab. Every time a notification leads me to the Files tab (for a single commit) I never know what to do with that information, I always have to see the whole Files or recent conversation.

PR #6653, opened 2023-05-15
Meta: Ensure new or edited features have Test URLs

fregante:
Most of our features lack test URLs. 

There's no chance of adding them all, but we can try to by reminding ourselves to add them every time we create and change (and _therefore **test**_) a feature.

This will help us progressively add more test URLs

### DIFF
--- a/build/verify-test-urls.sh
+++ b/build/verify-test-urls.sh
@@ -1,0 +1,43 @@
+#! /bin/bash
+
+# Automatically exit on error
+set -e
+
+EXIT_CODE=0
+
+# Keep short because GitHub doesn't wrap it
+ANNOTATION="Test URLs missing.
+The section must be appended to each .tsx file:
+
+/*
+
+Test URLs:
+
+https://github.com/a/REAL/url/here
+https://github.com/another/REAL/url/here
+
+*/
+"
+
+# Replace line breaks with "%0A"
+# https://github.com/actions/toolkit/issues/193#issuecomment-605394935
+ANNOTATION=${ANNOTATION//$'\n'/%0A}
+
+echo WILL VERIFY:
+echo
+
+for FILE in "$@"; do
+	if grep -q "test url" -i "$FILE"; then
+		echo ✅ "$FILE"
+	else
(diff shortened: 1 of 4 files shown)


build/verify-test-urls.sh
```diff
@@ -0,0 +1,43 @@
...
+		echo ❌ "$FILE"
+		echo "::error file=$FILE,line=1::$ANNOTATION"
+	fi
+done
+
+if [ $EXIT_CODE -eq 1 ]; then
+	echo
+	echo VERIFICATION FAILED, "Test URLs" MISSING
+	exit $EXIT_CODE
+fi
```
fregante, line 43: Screenshot 3

build/verify-test-urls.sh
```diff
@@ -0,0 +1,43 @@
...
+
+https://github.com/a/REAL/url/here
+https://github.com/another/REAL/url/here
+
+*/
+"
+
+# Replace line breaks with "%0A"
+# https://github.com/actions/toolkit/issues/193#issuecomment-605394935
+ANNOTATION=${ANNOTATION//$'\n'/%0A}
```
fregante, line 24: Screenshot 4

build/verify-test-urls.sh
```diff
@@ -0,0 +1,43 @@
+#! /bin/bash
+
+# Automatically exit on error
+set -e
+
+EXIT_CODE=0
+
+# Keep short because GitHub doesn't wrap it
+ANNOTATION="Test URLs missing.
```
yakov116, line 9: Maybe its just me, the wording sounds flipped
```suggestion
ANNOTATION="Missing Test URLs.
```

fregante: No, it’s like VIRUS DETECTED, right? 

“Missing X” would be followed by a list of X, like:

```
Missing keys:

- A
- Q
``` 

but here it works like:

```
# A.tsx

Keys missing (from this file)
```

yakov116: Ok

PR #6655, opened 2023-05-16
Fix notification action bar style

fregante:
- Partially reverts https://github.com/refined-github/refined-github/pull/5707

## Test URL

- https://github.com/notifications (sort by date OR repo)

## Native

before

## Refined GitHub  (causes jump)

after

## This PR

afterr

### DIFF
--- a/source/features/open-all-notifications.tsx
+++ b/source/features/open-all-notifications.tsx
@@ -128,3 +128,13 @@ void features.add(import.meta.url, {
 	],
 	init,
 });
+
+/*
+
+Test URLs:
+	
+https://github.com/notifications (Grouped by date)
+https://github.com/notifications (Grouped by repo)
+https://github.com/notifications?query=reason%3Acomment (which is an unsaved filter)
+
+*/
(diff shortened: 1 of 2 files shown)


source/refined-github.css
```diff
@@ -27,13 +27,7 @@ https://github.com/notifications (Grouped by repo)
 https://github.com/notifications?query=reason%3Acomment (which is an unsaved filter)
 */
 .js-check-all-container .js-bulk-action-toasts ~ div .Box-header {
-	height: auto !important;
 	flex-wrap: wrap;
```
fregante, line 30: I'm considering dropping this and restore the old width on mobile too. The same issue still happens on mobile because the buttons actually wraps into two lines.

I think the dropdown was an elegant solution to the issue, but I prefer that the header just scrolls instead, so this issue can be solved without wrapping:

- https://github.com/refined-github/refined-github/issues/5704

source/features/open-all-notifications.tsx
```diff
@@ -128,3 +128,13 @@ void features.add(import.meta.url, {
...
+
+/*
+
+Test URLs:
+	
+https://github.com/notifications (Grouped by date)
+https://github.com/notifications (Grouped by repo)
+https://github.com/notifications?query=reason%3Acomment (which is an unsaved filter)
+
+*/
```
fregante, line 140: Without this:

Screenshot


Thanks to:

- https://github.com/refined-github/refined-github/pull/6653

Issue #6589, opened 2023-04-28
`clean-rich-text-editor` no longer hides "please fill out this field" tooltips


Luke18033:
### Description

A while ago in #2349 I requested a feature that would hide "please fill out this field" tooltips on text boxes which are useless and do nothing but obscure what you're typing. This feature was implemented as part of `clean-rich-text-editor`, but at some point this stopped working and the tooltips now appear again.
![image](https://user-images.githubusercontent.com/19228201/235058983-9d5f936b-7d2a-451d-bf37-776195d60a88.png)

### How to replicate the issue + URL

Any page with required text inputs, such as [the create issue page](https://github.com/refined-github/refined-github/issues/new?template=1_bug_report.yml), [an issue](https://github.com/refined-github/refined-github/issues/6408), or [a discussion](https://github.com/StrataSource/Portal-2-Community-Edition/discussions/706).

### Extension version

23.4.26

### Browser(s) used

Firefox

AmalChandru:
@fregante Can I pick this one?

fregante:
You can just send a PR anytime 👌

fregante:
I ended up deciding that this isn't worth the hassle

PR #6658, opened 2023-05-17
Meta: Add test URLs to `clean-rich-text-editor`

AmalChandru:
Closes [#6589](https://github.com/refined-github/refined-github/issues/6589)

## Test URLs
Test on any page with required text inputs, such as [the create issue page](https://github.com/refined-github/refined-github/issues/new?template=1_bug_report.yml), [an issue](https://github.com/refined-github/refined-github/issues/6408), or [a discussion](https://github.com/StrataSource/Portal-2-Community-Edition/discussions/706)

### DIFF
--- a/source/features/clean-rich-text-editor.tsx
+++ b/source/features/clean-rich-text-editor.tsx
@@ -1,31 +1,9 @@
-import select from 'select-dom';
 import './clean-rich-text-editor.css';
 import * as pageDetect from 'github-url-detection';
 
 import features from '../feature-manager.js';
 
-const textInputSelectors = [
-	'#issue_title',
-	'#issue_form_repo',
-	'#issue_form_description',
-	'#issue_form_browser', // Id's for text input fields inside issues
-	'#new_comment_field', // Id for text input field in comments
-];
-
-async function init(): Promise<void> {
-	document.documentElement.classList.add('rgh-clean-rich-text-editor');
-	// Hide unnecessary tooltips
-	for (const textarea of select.all(textInputSelectors)) {
-		textarea.title = '';
-	}
-}
-
-void features.add(import.meta.url, {
-	include: [
-		pageDetect.hasRichTextEditor,
-	],
-	init,
-});
+void features.addCssFeature(import.meta.url, [pageDetect.hasRichTextEditor]);
 
 /*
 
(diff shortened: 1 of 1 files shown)


source/features/clean-rich-text-editor.tsx
```diff
@@ -2,5 +2,25 @@ import './clean-rich-text-editor.css';
...
+        textarea[id="issue_form_repro"], 
+        textarea[id="issue_form_description"], 
+        input[id="issue_form_browser", 
+        textarea[id="new_comment_field"]
+    `)) {
+        textarea.title = '';
+    }
+}
 
 void features.addCssFeature(import.meta.url, [pageDetect.hasRichTextEditor]);
```
fregante, line 19: This should be removed. In the new init function, it should just add the same class to the body/html. Refer to the source of addCssFeature to see what it does

source/features/clean-rich-text-editor.tsx
```diff
@@ -2,5 +2,25 @@ import './clean-rich-text-editor.css';
 import * as pageDetect from 'github-url-detection';
 
 import features from '../feature-manager.js';
+import select from 'select-dom';
+
+function hideToolTip(): void {
```
fregante, line 7: This can be named `init`

source/features/clean-rich-text-editor.tsx
```diff
@@ -2,5 +2,25 @@ import './clean-rich-text-editor.css';
...
+    for (const textarea of select.all(`
+        input[id="issue_title"],
```
fregante, line 9: The selector for id starts with #

Please replace all of these to look like `#issue_title` or `input#issue_title` (I don’t think the tag name is necessary)

Also select.all() accepts an array of strings. You can use that to improve wrapping and allow comments. You can refer to the other features to see how it can be done

fregante (commented):
Thank you for your PR. It seems that this was removed by @kidonng in https://github.com/refined-github/refined-github/pull/5113#discussion_r758575965

fregante:
After seeing all those new selectors, I realized that this minor change is not worth being implemented and maintained.

I altered the PR in order to still accept the Test URLs that you added, but the original issue was closed

fregante:
Thank you for the PR, @AmalChandru!

PR #6672, opened 2023-05-20
Open options in a new tab on iOS Safari

fregante:
- Closes https://github.com/refined-github/refined-github/issues/5455
- Follows https://github.com/refined-github/refined-github/pull/6592

Popups are awkward. Let's try this

IMG_8582


fregante:
Good news! This also fixes the “Export” button in iOS. I suppose Safari didn’t like showing a “Save as” dialog from inside the popup.

PR #6674, opened 2023-05-20
Meta: Use variables in GraphQL

fregante:
- Extracted from https://github.com/refined-github/refined-github/pull/6671
- Closes https://github.com/refined-github/refined-github/issues/5848

This PR

-  ✅  Adds support for variables
- ❌  Extracts GraphQL expressions into their own .gql files (https://www.apollographql.com/docs/react/integrations/webpack/)

This is harder than anticipated: 

- you can't include variable data in the JSON and not use them (i.e. no default user/repo)
- you can't just _use_ variables; each query must ALSO declare them…

```diff
- query {
+ query ($owner: String!, $name: String!) {
    repository(owner: $owner, name: $name) {
```

## Test

Extensions suggested: 
- https://chrome.google.com/webstore/detail/graphql-network-inspector/ndlbedplllcgconngcnfmkadhokfaaln
- https://marketplace.visualstudio.com/items?itemName=GraphQL.vscode-graphql-syntax

### DIFF
--- a/source/features/list-prs-for-file.gql
+++ b/source/features/list-prs-for-file.gql
@@ -1,0 +1,22 @@
+query getPrsByFile($owner: String!, $name: String!, $defaultBranch: String!) {
+	repository(owner: $owner, name: $name) {
+		pullRequests(
+			first: 25,
+			states: OPEN,
+			baseRefName: $defaultBranch,
+			orderBy: {
+				field: UPDATED_AT,
+				direction: DESC
+			}
+		) {
+			nodes {
+				number
+				files(first: 100) {
+					nodes {
+						path
+					}
+				}
+			}
+		}
+	}
+}
--- a/source/features/list-prs-for-file.tsx
+++ b/source/features/list-prs-for-file.tsx
@@ -10,6 +10,7 @@ import getDefaultBranch from '../github-helpers/get-default-branch.js';
 import {buildRepoURL, cacheByRepo} from '../github-helpers/index.js';
 import GitHubURL from '../github-helpers/github-url.js';
 import observe from '../helpers/selector-observer.js';
+import listPrsForFileQuery from './list-prs-for-file.gql';
 
 function getPRUrl(prNumber: number): string {
 	// https://caniuse.com/url-scroll-to-text-fragment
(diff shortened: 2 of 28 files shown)


yakov116:
JMHO, on any non simple query I think this lowers readability

fregante:
It’s definitely more verbose but there’s no alternative if we want to extract the queries: 

- https://github.com/refined-github/refined-github/issues/5848

Once extracted we’ll have syntax highlighting and potentially some validation, as well as proper types for v4 calls. I think it’s worth it, even if the query ends up being in a different file.

fregante (commented):
Everything here was tested. Some features were even restored, some tickets were created.

In a future PR:

- [ ] Extract to .gql files
- [ ] Update _every_ query (I ignored the simple ones without additional variables, although they're still affected by this change)

source/features/list-prs-for-file.gql
```diff
@@ -0,0 +1,22 @@
...
+	repository(owner: $owner, name: $name) {
+		pullRequests(
+			first: 25,
+			states: OPEN,
+			baseRefName: $defaultBranch,
+			orderBy: {
+				field: UPDATED_AT,
+				direction: DESC
+			}
+		) {
```
fregante, line 11: Demo extracted GQL file

source/features/list-prs-for-file.tsx
```diff
@@ -57,28 +58,12 @@ function getDropdown(prs: number[]): HTMLElement {
...
-						nodes {
-							path
-						}
-					}
-				}
-			}
-		}
-	`);
+	debugger;
+	const {repository} = await api.v4(listPrsForFileQuery, {
```
fregante, line 62: Demo feature with extracted query

fregante (commented):
You can see what the usage of `.gql` files will look like in `list-prs-for-file` 

Missing: 

- [ ] TypeScript integration that will help us use the results in the file
- [ ] GraphQL autocomplete in the IDE thanks to the schema
- [ ] Typed `variables` that make TypeScript complain if one is missing or unused
- [ ] Prettier/formatter for GQL support (we have a .prettierignore file)

source/features/list-prs-for-file.tsx
yakov116: Nice! 

Can we put this in a new pr?

yakov116: Never mind just saw your post

PR #6677, opened 2023-05-21
 Better wrap: `clean-pinned-issues`/`pinned-issues-update-time`

fregante:
## Test URLs

- https://github.com/refined-github/refined-github/issues

### Before
Screenshot 4

### After
Screenshot 2

### Before

Screenshot 5

### After

Screenshot 6


source/features/clean-pinned-issues.css
```diff
@@ -3,6 +3,15 @@
...
 }
 
+.rgh-clean-pinned-issues .js-pinned-issues-reorder-list {
+	gap: 16px;
+}
+.rgh-clean-pinned-issues .pinned-issue-item {
+	width: auto !important;
+	flex: 1;
+	margin-inline: 0 !important;
+}
```
fregante, line 13: This fixes the default "3-column" layout, making it fit to 1 or 2 issues.

The single-column on the phone is unaffected:

Screenshot

source/features/clean-pinned-issues.css
```diff
@@ -26,11 +35,13 @@
 		display: table-cell !important;
 		padding: 6px 12px;
 		vertical-align: middle;
+		white-space: nowrap;
```
fregante, line 38: This avoids some awkward wrapping in certain cases

Screenshot

source/features/pinned-issues-update-time.tsx
```diff
@@ -43,12 +44,13 @@ async function init(): Promise<void | false> {
...
-		select('.pinned-item-desc', pinnedIssue)!.append(
-			' • ',
-			<span className="color-fg-muted d-inline-block">
+		select('.pinned-item-desc', pinnedIssue)!.after(
+			<span className="text-small color-fg-muted">
 				updated <relative-time datetime={updatedAt}/>
 			</span>,
 		);
+
+		select('.pinned-item-desc', pinnedIssue)!.hidden = true;
```
fregante, line 53: This changes the behavior of the feature. There's too much content up there and it causes the issue list to jump down in some cases.

I initially thought about dropping the feature, but I think it's useful more so than the native "user X opened on DATE" information.

The element is hidden, not removed, so it can be restored via CSS by the user.

PR #6684, opened 2023-05-21
Hide reason text in `status-subscription`

fregante:
Extracted from a discussion in https://github.com/refined-github/refined-github/pull/6667 cc @FloEdelmann 

This text doesn't add anything more to the state shown by the feature, except when "all" is selected, because it can be one of:

- You’re receiving notifications because you’re watching this repository.
- You’re receiving notifications because you authored the thread.
- You’re receiving notifications because you commented.
- You’re receiving notifications because you were mentioned.
- _who knows what else_

## URL

Here

### DIFF
--- a/source/features/status-subscription.tsx
+++ b/source/features/status-subscription.tsx
@@ -92,6 +92,7 @@ function addButton(subscriptionButton: HTMLButtonElement): void {
 	originalId.remove();
 	subscriptionButton.hidden = true;
 
+	// 'all' can have many reasons, but the other two don't add further information #6684
 	if (status !== 'all') {
 		getReasonElement(subscriptionButton).hidden = true;
 	}
--- a/source/features/status-subscription.tsx
+++ b/source/features/status-subscription.tsx
@@ -24,11 +24,14 @@ function SubButton(): JSX.Element {
 	);
 }
 
-function getCurrentStatus(subscriptionButton: HTMLButtonElement): 'none' | 'all' | 'status' {
-	const reason = subscriptionButton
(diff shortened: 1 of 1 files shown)


fregante:
Also verified:

Screenshot 2

source/features/status-subscription.tsx
```diff
@@ -88,6 +91,10 @@ function addButton(subscriptionButton: HTMLButtonElement): void {
 	// Remove it only if the form was successfully added
 	originalId.remove();
 	subscriptionButton.hidden = true;
+
+	if (status !== 'all') {
```
FloEdelmann, line 95:
```suggestion
	// Different reasons lead to the "all notifications enabled" state #6684
	if (status !== 'all') {
```

FloEdelmann (approved):
Makes sense to me :+1:

source/features/status-subscription.tsx
```diff
@@ -98,7 +98,7 @@ function addButton(subscriptionButton: HTMLButtonElement): void {
 }
 
 function init(signal: AbortSignal): void | false {
-	observe('button[data-thread-subscribe-button]', addButton, {signal});
+	observe('button[data-thread-subscribe-button]:not([disabled])', addButton, {signal});
```
FloEdelmann, line 101: Maybe mention this in the "Test URLs" section?

source/features/status-subscription.tsx
fregante: How about

```suggestion
	// 'all' can have many reasons, but the other two don't add further information #6684
	if (status !== 'all') {
```

FloEdelmann (approved):
Again :wink:

PR #6690, opened 2023-05-27
Fix `previous-version` throws error on brand-new files

### LINKED ISSUE #6681: `previous-version` throws error on brand-new files
### Description

New files don't have a previous version, the API returns one commit instead of 2.

Quick solution: if there's only one "node" in the response, return `false`

cc @134130

1. https://github.com/refined-github/sandbox/blob/6619/6619

### Extension version

23.5.10

### Browser(s) used

Chrome

134130:
## Description
- Closes #6681
- > New files don't have a previous version, the API returns one commit instead of 2.
Quick solution: if there's only one "node" in the response, return false

## Test URLs
1. [refined-github/sandbox@6619/6619](https://github.com/refined-github/sandbox/blob/6619/6619?rgh-link-date=2023-05-21T13%3A33%3A12Z)

## Screenshot
Before
- image
After
- No Error

### DIFF
--- a/source/features/previous-version.tsx
+++ b/source/features/previous-version.tsx
@@ -8,7 +8,7 @@ import observe from '../helpers/selector-observer.js';
 import api from '../github-helpers/api.js';
 import GitHubURL from '../github-helpers/github-url.js';
 
-async function getPreviousCommitForFile(pathname: string): Promise<string> {
+async function getPreviousCommitForFile(pathname: string): Promise<string | false> {
 	const {user, repository, branch, filePath} = new GitHubURL(pathname);
 	const {resource} = await api.v4(`
 		query getPreviousCommitForFile($resource: URI!, $filePath: String!) {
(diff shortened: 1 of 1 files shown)


fregante (approved):
Thank you!

134130:
@fregante New changes
- Use optional chaining
- Use `.gql` file

source/features/previous-version.tsx
```diff
@@ -1,41 +1,28 @@
...
 import {VersionsIcon} from '@primer/octicons-react';
 
 import features from '../feature-manager.js';
 import observe from '../helpers/selector-observer.js';
 import api from '../github-helpers/api.js';
 import GitHubURL from '../github-helpers/github-url.js';
+import previousVersionQuery from './previous-version.gql';
 
-async function getPreviousCommitForFile(pathname: string): Promise<string> {
+async function getPreviousCommitForFile(pathname: string): Promise<string | undefined> {
```
fregante, line 11: Actually we only needed `false` if the `webext-storage-cache` was involved, because `undefined === do not cache`

In this case there's no caching, so `undefined` is fine

fregante (approved):

PR #6692, opened 2023-05-27
Fix page number calculation in `last-notification-page-button`

janpio:
Main change here is the `-1` that is subtracted from the number of total notifications to make the calculation return the correct value for the cursor.

image

Other changes are renaming variables to make them more descriptive waht they contain.

Closes https://github.com/refined-github/refined-github/issues/6691

### DIFF
--- a/source/features/last-notification-page-button.tsx
+++ b/source/features/last-notification-page-button.tsx
@@ -12,7 +12,7 @@ const itemsPerNotificationsPage = 25;
 function linkify(nextButton: HTMLAnchorElement): void {
 	const totalNotificationsNode = select('.js-notifications-list-paginator-counts')!.lastChild!;
 	assertNodeContent(totalNotificationsNode, /^of \d+$/);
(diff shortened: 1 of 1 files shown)


janpio:
Not sure how to deal with the test URL in this case, as it is user specific.
I could put the same here that I used in the issue, but those will not work for anyone else but me.

fregante:
> Not sure how to deal with the test URL in this case, as it is user specific.

The test URL is `https://github.com/notifications`. While the bug is user-specific, the feature works for everyone on every `/notification` subpage.

fregante (approved):
Thank you for the PR!

janpio:
All fixed 👍

fregante:
🤞 hopefully this is the last math fix in this feature 😅

janpio:
I'll know when I use it 😆

janpio:
Seems to work now 👍

fregante:
Thanks for the update

PR #6693, opened 2023-05-27
Adjust wording of `status-subscription`’s tooltip

waldyrious:
The button group for but the current label ("Subscribe just to the close event") is a little misleading because that subscription level also includes other status change events (e.g. merging a PR or reopening an issue). I slightly reworded the tooltip to be a bit more accurate.

## Test URLs

This PR, for instance.

## Screenshot

Before:

Screenshot from 2023-05-27 18-01-51

After:

Screenshot from 2023-05-27 18-04-33

### DIFF
--- a/source/features/status-subscription.tsx
+++ b/source/features/status-subscription.tsx
@@ -74,7 +74,7 @@ function addButton(subscriptionButton: HTMLButtonElement): void {
 			<SubButton
 				// @ts-expect-error I don't remember how to fix this
 				value="subscribe_to_custom_notifications"
-				aria-label="Subscribe just to status changes (closing, reopening)"
+				aria-label="Subscribe just to status changes&#10;(closing, reopening, merging)"
 				{...(status === 'status' && disableAttrs)}
 			>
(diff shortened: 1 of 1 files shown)


fregante:
This was intentional, it subscribes to up to three events and I’m not sure I want to make it too verbose. I suppose that it could be both short and correct though:

> Subscribe to closing, reopening, merging events

Does this make the tooltip too long? Does it cause issues on mobile? I think we could alternatively wrap it manually via `%A0` character (if I remember correctly) to make sure it’s never longer than the column.

waldyrious:
> I think we could alternatively wrap it manually via %A0 character (if I remember correctly) to make sure it’s never longer than the column.

Ah, nice idea. U+00A0 is actually the non-breaking space; newline is U+000A. I believe that in this case, since it's part of an HTML attribute, we need to encode it as an HTML entity (`&#x0a;`, or `&#10;` in decimal notation), not as URL-encoded (`%0A`). I'll make the change.

waldyrious:
@fregante here's what it should look like now:

image

PR #6694, opened 2023-05-28
Add `action-pr-link` feature

### LINKED ISSUE #6206: When an action has spawned from a PR, provide a link back to the PR
### Description

On a github PR if you follow the pathway below:

Screenshot 2022-12-21 at 10 22 13

You end up on a screen with no breadcrumb back to the original PR.

Neither here:
Screenshot 2022-12-21 at 10 22 39

Or here:
Screenshot 2022-12-21 at 10 24 32

Only by clicking summary and then this tiny link can you get back to the original PR you've navigated from.
Screenshot 2022-12-21 at 10 29 47

What we want, is simply a link at the top of the page that links back to a PR, if it exists. e.g.

Screenshot 2022-12-21 at 10 32 54

### Screenshot

(Attached with context in description)

### Example URLs

https://github.com/refined-github/refined-github/actions/runs/3732055304/jobs/6331000824

134130:
## Description
- Closes #6206

## Test URLs
- https://github.com/refined-github/refined-github/actions

## Screenshot
image


134130:
When the action is defined as check run, `matchingPullRequests` is missing and `checkSuite > checkRuns` contains some values.
Need to find the way of figure out related PR

yakov116:
I don't understand why can you not there the pr number from the issue title?

134130:
@yakov116 That is not the issue number, The WorkflowRun Number

source/features/linkify-action.tsx
```diff
@@ -0,0 +1,81 @@
...
+	if (resource.event !== 'pull_request') {
+		return false;
+	}
+
+	const {edges} = resource.checkSuite.matchingPullRequests;
+	if (edges.length === 0) {
+		return false;
+	}
+
+	return edges[0].node.number;
```
fregante, line 44: You can use this format from now on, perhaps also changing your previous PR (which uses a similar .length check, if I remember correctly). That change can also be done in this PR

```suggestion
	return resource.checkSuite.matchingPullRequests.edges[0]?.node.number ?? false;
```

yakov116:
Aha, use api v3

https://github.com/refined-github/refined-github/blob/3eb69778b29b87b02a4ab2ddda4d8a302307f759/source/features/bypass-checks.tsx#L16

source/features/linkify-action.tsx
```diff
@@ -0,0 +1,81 @@
...
+	const paths = url.pathname.split('/').slice(1);
+	const prNumber = await getPullRequestNumber(paths[0], paths[1], paths[4]);
+
+	if (!prNumber) {
+		return;
+	}
+
+	contextBar.classList.add('flex-column', 'flex-items-start');
+	contextBar.append(
+		<span className="PageHeader-parentLink">
```
fregante, line 62: Instead of adding a new line, we should add it next to an existing text. The current setup pushes the whole page down when loaded, which we should avoid. See this issue I recently fixed:

- #6677 

A possible position is either next to "ci" or next to the run number:

```
{title………} #123 (PR #6666)
```

source/refined-github.ts
```diff
@@ -216,3 +216,4 @@ import './features/pr-base-commit.js';
 import './features/unreleased-commits.js';
 import './features/previous-version.js';
 import './features/status-subscription.js';
+import './features/linkify-action.js';
```
fregante, line 219: Probably:

- action-pr-link
- action-back-to-pr

(names should be short)

source/features/linkify-action.tsx
```diff
@@ -0,0 +1,81 @@
...
+	if (edges.length === 0) {
+		return false;
+	}
+
+	return edges[0].node.number;
+}
+
+async function add(contextBar: HTMLDivElement): Promise<void> {
+	const url = new URL(location.href);
+	if (typeof url.searchParams.get('pr') === 'string') {
```
fregante, line 49: This is a static check that can be part of the `exclude` array in `features.add()`

```js
exclude: [
	 new URLSearchParams(location.search).has('param');
]
```

However… why are these being excluded? Can you provide an example?

source/features/linkify-action.tsx
```diff
@@ -0,0 +1,81 @@
+import React from 'dom-chef';
+
+import * as pageDetect from 'github-url-detection';
+
+import {ArrowLeftIcon} from '@primer/octicons-react';
```
fregante, line 5:
```suggestion
import React from 'dom-chef';
import * as pageDetect from 'github-url-detection';
import {ArrowLeftIcon} from '@primer/octicons-react';
```

readme.md
```diff
@@ -164,6 +164,7 @@ Thanks for contributing! 🦋🙌
 - [](# "quick-new-issue") [Adds a link to create issues from anywhere in a repository.](https://user-images.githubusercontent.com/1402241/218251057-b94b62dd-a944-4763-b78a-fc233f7c9fd3.png)
 - [](# "small-user-avatars") [Shows a small avatar next to the username in conversation lists.](https://user-images.githubusercontent.com/44045911/230960291-721f42cc-e1ac-4fdc-83ea-2430b062f9ce.png)
 - [](# "unreleased-commits") 🔥 [Tells you whether you're looking at the latest version of a repository, or if there are any unreleased commits.](https://user-images.githubusercontent.com/1402241/234576563-1a0ca255-4c0d-45ae-883d-2b1aa2d7f4c1.png)
+- [](# "linkify-action") 🔥 [Adds a link back to the PR.](https://github.com/refined-github/refined-github/assets/50487467/ba5bd616-7a48-4495-be99-c9e315f8629f)
```
fregante, line 167: "Adds a link back to the PR" isn't descriptive enough

fregante (commented):
Thank you for the PR!

source/features/linkify-action.tsx
134130: I'm on mobile.
Please check this comment! https://github.com/refined-github/refined-github/issues/6206#issuecomment-1566090775

fregante: Ohh, that explains the choices made here.

source/features/linkify-action.tsx
fregante: I see that this is the native position and style in:


https://github.com/refined-github/refined-github/actions/runs/5104153489/jobs/9174884558?pr=6694

So now I'm thinking… why don't we change the links in https://github.com/refined-github/refined-github/actions to include this `pr` parameter instead of using the API? This would make the feature simpler (I'm not sure if there are other situations where this "fix" is possible)

134130: I did not added search params despite it is simple since the original back link button will be disappeared.

And how can we observe that? Adding some observing feature on local memory or storage? How about we visit the action page directly not caching actions page? (also pagination)

fregante: > the original back link button will be disappeared

I don't find that button useful and GitHub agrees since they're hiding it. We can discuss this before making a decision though.

cc @kidonng too

134130: I agree original link is not useful

fregante: > And how can we observe that?

The feature I'm suggesting would be similar to `hide-user-forks`. In short:

```js
observe('that pr link I mentioned', prLink => {
  prLink.closest('.row').querySelector('a').search = '?pr=' + prLink.textContent
})
```

134130: Oops my bad english skill.
I understood observe triggered PR number and Add back link with that information on action run page. But you said actions (list) page.
That feature is easy. What should the feature name be

fregante: Yes, I’m suggesting that those 3 lines are the entire feature, no API calls at all.

source/refined-github.ts
134130: The feature name will be `action-pr-link`

source/features/linkify-action.tsx
```diff
@@ -0,0 +1,81 @@
...
+				}
+			}
+		}
+	`, {
+		variables: {
+			resource: `/${user}/${repository}/actions/runs/${runDatabaseId}`,
+		},
+	});
+
+	if (resource.event !== 'pull_request') {
```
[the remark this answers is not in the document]
134130: This feature is linking actions to PR.
Let me know which target is to be linked with "push" event

134130:
I found the pull_request field works correctly only on administrator of repository regardless v4 or v3.

source/features/linkify-action.tsx
134130: @fregante When we append search params, the page is being refreshed (and flickering a time). I don't think this is refined.
How about just replacing original link to new link?

source/features/linkify-action.tsx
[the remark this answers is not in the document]
fregante: This event is part of the current PR, but GitHub doesn’t seem to link that in any way 😣 

Last commit here:

https://github.com/refined-github/sandbox/pull/12#commits-pushed-9802281

I suppose the “observer” would also have to watch the CI links in PRs

fregante:
> I found the pull_request field works correctly only on administrator of repository regardless v4 or v3.


As mentioned before, I don’t think this feature should use the API, but just the observer to alter the links adding `?pr=` wherever possible. We can see how well this works and potentially add the API in a month if it brings substantial benefits.

source/refined-github.ts
fregante: We’ve been inconsistent when naming these features. Sometimes we call them “action”, others “workflows”, others “checks”. Wait before renaming until we figure out where this applies.

source/features/linkify-action.tsx
134130: It seems cool

134130:
Yes I'm working with not using api call :)
Just note for future.

readme.md
134130: Changed to "Add a link back to the PR that ran the action."

source/refined-github.ts
134130: Yes. In this repo, it usually called "action", but I think the formal is "workflow-run".
"workflows" means just recipe file of "workflow-run".
"checks" is more larger concept which can also contains "workflow-run" as "check". The "checks" can be other api jobs not in GitHub.

source/features/linkify-action.tsx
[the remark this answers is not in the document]
134130: I don't know if I understood it correctly, 
But when the pull_request's head is changed, `pull_request.synchronize` is the event.

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request

source/refined-github.ts
fregante: > The "checks" can be other api jobs not in GitHub.

Indeed, those also have a dedicated "checks" page here on GitHub: https://github.com/refined-github/refined-github/issues/1620

But yeah, that page only appears in the PR view, not in the Actions tab.

"action" works well here.

We also have `github-actions-indicators` which is similar but it's yet another slight variation 😅

source/features/action-pr-link.tsx
```diff
@@ -0,0 +1,34 @@
...
+import observe from '../helpers/selector-observer.js';
+
+async function add(row: HTMLDivElement): Promise<void> {
+	const prNumber = select('a[data-hovercard-type="pull_request"]', row)?.textContent?.slice(1);
+
+	if (prNumber === undefined) {
+		return;
+	}
+
+	select('a.Link--primary', row)!.href += `?pr=${prNumber}`;
```
fregante, line 14: I think needs something like this to be safe, just in case these links happen to have additional parameters in the future:

```suggestion
	const runLink = select('a.Link--primary', row)!;
	const parameters = new URLSearchParams(runLink.search);
	parameters.set('pr', prNumber);
	runLink.search = String(parameters);
```

I wish HTMLAnchorElement had a `searchParams` argument to make this easier 😒

source/features/action-pr-link.tsx
```diff
@@ -0,0 +1,34 @@
...
+}
+
+async function init(signal: AbortSignal): Promise<void> {
+	observe('div.Box-row[id^=check_suite_]', add, {signal});
```
fregante, line 18: Just select the `a[data-hovercard-type="pull_request"]` instead, so you don't have to `select` it again and then `return` when undefined.

```suggestion
	observe('div.Box-row[id^=check_suite_] a[data-hovercard-type="pull_request"]', add, {signal});
```

and then probably:

```ts
function add(prLink: HTMLAnchorElement) {
	const runLink = prLink.closest('.Row')!.querySelector('a.Link--primary')!;
	const parameters = new URLSearchParams(runLink.search);
	parameters.set('pr', prNumber);
	runLink.search = String(parameters);
}
```

fregante (approved):
This is great!

One limitation I found is that GitHub does not support issues, not via `pr=` nor via `issue=`

For example this workflow runs on both, but only the PRs can be linkified: https://github.com/refined-github/refined-github/actions/workflows/labeler.yml

readme.md
```diff
@@ -163,6 +163,7 @@ Thanks for contributing! 🦋🙌
 - [](# "quick-new-issue") [Adds a link to create issues from anywhere in a repository.](https://user-images.githubusercontent.com/1402241/218251057-b94b62dd-a944-4763-b78a-fc233f7c9fd3.png)
 - [](# "small-user-avatars") [Shows a small avatar next to the username in conversation lists.](https://user-images.githubusercontent.com/44045911/230960291-721f42cc-e1ac-4fdc-83ea-2430b062f9ce.png)
 - [](# "unreleased-commits") 🔥 [Tells you whether you're looking at the latest version of a repository, or if there are any unreleased commits.](https://user-images.githubusercontent.com/1402241/234576563-1a0ca255-4c0d-45ae-883d-2b1aa2d7f4c1.png)
+- [](# "action-pr-link") 🔥 [Adds a link back to the PR that ran the action.](https://github.com/refined-github/refined-github/assets/50487467/076a0137-36a2-4fd0-a66e-735ef3b3a563)
```
fregante, line 166:
```suggestion
- [](# "action-pr-link") 🔥 [Adds a link back to the PR that ran the workflow.](https://github.com/refined-github/refined-github/assets/50487467/076a0137-36a2-4fd0-a66e-735ef3b3a563)
```

Screenshot

PR #6695, opened 2023-05-28
Fix `quick-review` position due to race condition

fregante:
Old bug. I was sure it was reported before but… nope.

I don't see any other "reasonable" positions so I'll just add a delay

## Test URLs

Any PR

## Issue

Screenshot


yakov116:
Thanks for fixing I think it was in one of these long issues we once had

fregante:
Maybe I just randomly mentioned it somewhere but never opened a dedicated issue. It happens all the time 😅

PR #6702, opened 2023-05-31
Harder better faster stronger `new-or-deleted-file`

fregante:
- Close https://github.com/refined-github/refined-github/issues/6678

## Test URLs

Commit: https://github.com/refined-github/sandbox/commit/a00694ff7370d46b5f6d723a0b39141903dae45a
PR: https://github.com/refined-github/sandbox/pull/71/files
PR with CODEOWNERS: https://github.com/dotnet/winforms/pull/6028/files

Note: Not possible with 1-file commits: https://github.com/refined-github/sandbox/commit/1ed862e3abd13004009927b26355a51a109d6855

## PR

Screenshot 2

## Commit

Screenshot 3


readme.md
```diff
@@ -304,7 +304,7 @@ Thanks for contributing! 🦋🙌
 - [](# "tags-on-commits-list") [Displays the corresponding tags next to commits.](https://user-images.githubusercontent.com/14323370/66400400-64ba7280-e9af-11e9-8d6c-07b35afde91f.png)
 - [](# "mark-merge-commits-in-list") [Marks merge commits in commit lists.](https://user-images.githubusercontent.com/16872793/75561016-457eb900-5a14-11ea-95e1-a89e81ee7390.png)
 - [](# "deep-reblame") [When exploring blames, `Alt`-clicking the “Reblame” buttons will extract the associated PR’s commits first, instead of treating the commit a single change.](https://user-images.githubusercontent.com/16872793/77248541-8e3f2180-6c10-11ea-91d4-221ccc0ecebb.png)
-- [](# "highlight-deleted-and-added-files-in-diffs") [Indicates with an icon whether files in commits and pull requests being added or removed.](https://user-images.githubusercontent.com/1402241/90332474-23262b00-dfb5-11ea-9a03-8fd676ea0fdd.png)
+- [](# "new-or-deleted-file") [Indicates with an icon whether files in commits and pull requests being added or removed.](https://user-images.githubusercontent.com/1402241/90332474-23262b00-dfb5-11ea-9a03-8fd676ea0fdd.png)
```
fregante, line 307: New name ok?

source/features/new-or-deleted-file.tsx
```diff
@@ -0,0 +1,52 @@
...
+	const fileInList = select(`[href="${filename.hash}"]`, list);
+	if (!fileInList) {
+		features.log.error(import.meta.url, 'Could not find file in sidebar, is the sidebar loaded?');
+		features.unload(import.meta.url);
+		return;
+	}
+
+	const icon = select('.octicon-diff-removed, .octicon-diff-added', fileInList)
+		?.cloneNode(true);
+	if (icon) {
```
fregante, line 24: I added a bunch of conditions to have clear errors in case filelists are deferred in big PRs. I can drop the `features.log.error` condition if anyone finds one such big PR

source/features/new-or-deleted-file.tsx
```diff
@@ -0,0 +1,52 @@
+import React from 'dom-chef';
+import select from 'select-dom';
+import * as pageDetect from 'github-url-detection';
+
+import features from '../feature-manager.js';
+import observe from '../helpers/selector-observer.js';
+
+function add(filename: HTMLAnchorElement): void {
+	const list = select('ul[aria-label="File Tree"]');
```
fregante, line 9: This feature used to use this dropdown:

Screenshot 4

but it only appears on PRs _and_ it's actually been relegated to small screens, because the whole dropdown is hidden/duplicate of the sidebar on large screens.

It seems that the sidebar is actually part of the original HTML, so we don't even need to wait for it to load (at least some times)

fregante: I ended up _not_ using the API because I figured we'd still have trouble figuring out the exact changes/api to consider in case of single commits and multi-PR-commits. The sidebar is recent enough, hopefully it will last a while.

readme.md
FloEdelmann: Hmm, I don't find it as clear as the old one. Maybe `new-or-deleted-file-diff-indicator`?

fregante: Feature names should no longer be fully descriptive, I'm only renaming long feature names because they're impossible to mention in commits or to remember. See also the other recent renames in `options-storage`

fregante: See https://github.com/refined-github/refined-github/issues/6707

fregante: I agree that the new name isn't very descriptive, but between descriptiveness and brevity we should prefer the latter, as described in #6707. 

Alternative short names are welcome; names that are only slightly shorter than what we have aren't enough.

FloEdelmann: `file-diff-type-indicator`?

fregante: While that sounds correct, it gives me even less information about what the indicator is about. "diff type" == ?

Maybe `mark-new-or-deleted-file` but that's so long too.

Issue #5531, opened 2022-03-22
Migrating to Manifest V3


kidonng:
### Description

I don't think we mentioned this anywhere so here we are. Nothing to hurry, just in case we don't forget it.

[Manifest V2 will be gone next year](https://developer.chrome.com/docs/extensions/mv3/mv2-sunset/), and we don't want Refined GitHub to stop evolving and rot in GitHub updates.

![](https://wd.imgix.net/image/SHhb2PDKzXTggPGAYpv8JgR81pX2/zXdU3hdkj1K0Ks6tAfB4.png)

_Hope we are not going through the 2to3 pain again like last time the world upgrades a certain reptile_

### Screenshot

![](https://i.imgflip.com/69izzk.jpg)

### Example URL

The universe

fregante:
A few things keeping us from updating:

- no Firefox support: https://bugzilla.mozilla.org/show_bug.cgi?id=1578284
- support for GHE isn't possible at the moment and it won't be until the figure out the `optional_permissions`:
	- https://github.com/w3c/webextensions/issues/119
	- https://github.com/fregante/webext-dynamic-content-scripts/issues/18
	- https://github.com/fregante/webext-domain-permission-toggle/issues/19

Katsute:
If you want to preserve some compatibility between Chrome and Firefox you could have a post build script that modifies the manifest version and produces two distributions: one for Chrome and one for Firefox.

https://github.com/Katsute/GitHub-Red-Issues/blob/3ff79d59722e1851508b5e6d87804a1d335de932/build.ts#L35-L45

The optional perm issue might only be solvable with a separate GHE distribution that requires the perm if they don't added it to MV3.

fregante:
I fully expect both issues to be resolved by January 2023, we really don't need to update _now_. If the deadline is near and issues persist, we'll figure something out then.

FloEdelmann:
FYI: https://blog.mozilla.org/addons/2022/05/18/manifest-v3-in-firefox-recap-next-steps/

fregante:
Let's not waste time debugging their stuff just yet, we'll do it in 5-6 months. 👍 

Support for GHE is completely missing in Chrome MV3 so Firefox is not the only reason to wait.

fregante:
It's probably time to start testing. We can convert the extension to use MV3 and then call a MV3-to-MV2 script here:

- in package.json as `mv2-conversion`
[...]

PR #6710, opened 2023-06-01
Restore full GitHub Enterprise support

fregante:
- Fixes https://github.com/refined-github/refined-github/issues/6025

Context:

- Dynamic injection isn't fully possible in event pages with the current API: https://github.com/fregante/content-scripts-register-polyfill/commit/eac3418eb435e0e3e220721eee9ff5bee1a1804a
- iOS _only_ supports event pages

Solution:

- Ship regular background pages to desktop browsers
- Ship event pages to Safari/iOS

Future:

- #5531


fregante:
Let's give it a go

PR #6722, opened 2023-06-01
Meta: Use GitHub style in options

fregante:
- Closes https://github.com/refined-github/refined-github/issues/6402

Screenshot 7
Screenshot 8

### DIFF
--- a/source/options.css
+++ b/source/options.css
@@ -12,7 +12,17 @@
 }
 
 :root {
-	--rgh-red: #cf222e;
+	--red: #cf222e;
+	--border: #d0d7de;
+	--background: #f6f8fa;
+}
+
+@media (prefers-color-scheme: dark) {
+	:root {
+		--red: #f85149;
+		--border: #3e4348;
+		--background: #3a3d4066;
+	}
 }
 
 html::after {
@@ -23,12 +33,6 @@ html::after {
 	/* webpackIgnore: true */
 	background: url('icon.png') center no-repeat;
 	filter: opacity(0.1) saturate(0) brightness(1.2);
-}
-
-@media (prefers-color-scheme: dark) {
-	:root {
-		--rgh-red: #f85149;
-	}
 }
 
 p {
@@ -46,6 +50,8 @@ li[data-validation] {
 
 details {
 	margin-bottom: 1em;
+	border: solid 1px var(--border);
+	border-radius: 6px;
 }
 
 output {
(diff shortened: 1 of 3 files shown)


source/options.css
```diff
@@ -46,41 +50,35 @@ li[data-validation] {
...
-	padding-bottom: 1px;
-}
-
 summary {
 	--summary-padding: 10px;
-	background: #aaa1;
+	background: var(--background);
 	list-style: none;
 	padding: var(--summary-padding);
-	cursor: pointer;
```
fregante, line 66: No cursor needed on `details`

source/options.css
```diff
@@ -46,41 +50,35 @@ li[data-validation] {
...
+	border-radius: 6px;
 }
 
 summary::-webkit-details-marker {
 	display: none; /* Just for Safari. Chrome uses `list-style` */
 }
 
-summary:hover {
-	background: #aaa4;
-}
```
fregante, line 75: No hover style for now, let's see how it feels. I think it's not too difficult to understand that the sections are togglable

source/options.tsx
```diff
@@ -204,9 +204,11 @@ function featuresFilterHandler(event: Event): void {
...
 }
 
-function focusFirstField(event: DelegateEvent<Event, HTMLDetailsElement>): void {
-	if (event.delegateTarget.open) {
-		const field = select('input, textarea', event.delegateTarget);
+function focusFirstField({delegateTarget: section}: DelegateEvent<Event, HTMLDetailsElement>): void {
+	// @ts-expect-error No Firefox support https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoViewIfNeeded
+	(section.scrollIntoViewIfNeeded ?? section.scrollIntoView).call(section);
+	if (section.open) {
+		const field = select('input, textarea', section);
```
fregante, line 211: https://github.com/refined-github/refined-github/assets/1402241/e009ad5b-ec0d-4a91-9e5d-b10275bd5dd0

source/options.css
FloEdelmann: Maybe add `cursor: pointer` instead?

fregante: See comment above, I just removed it 😃 

https://marvelapp.com/blog/buttons-shouldnt-hand-cursor/

fregante:
I'm not sure I like it more than the current one 😬 which is a bit more modern and fits better with the browser’s _chrome_

See screenshots in https://github.com/refined-github/refined-github/pull/6384#discussion_r1116543903

source/options.css
FloEdelmann: But wasn't the whole point of this PR to match GitHub's style? GitHub does use the pointer cursor for buttons and collapsible sections.

Also, these section elements do not have the "strong accordance" the linked article is referring to, since the default `<details>` styling is changed.

fregante:
Never mind

PR #6725, opened 2023-06-02
Add `mobile-tabs` feature

### LINKED ISSUE #5805: No-text repo tabs on mobile
We discussed this before but there was never a specific issue, I think.

Can we hide the labels **on mobile**?

Screen Shot 2

```css
@media (max-width:767px) {
	.UnderlineNav-item [data-content] {
		overflow: hidden;
		display: inline-block;
		width: 0; /* Preserve the button height; display: none collapses it*/
	}
	.UnderlineNav-item svg {
		margin-right: 0 !important;
	}
}
```

If the tabs aren't differentiable enough, we can keep the hover state on:

Screen Shot 3

But there's probably a better style for that.

```css
.UnderlineNav-item {
	background: var(--color-action-list-item-default-hover-bg);
}
```

fregante:
- Closes #5805 

## Test URLs

Here

## Screenshot

121495005-89af8600-c9d9-11eb-822d-77e0b987e3b1


yakov116:
I like it!

fregante:
After seeing the last screenshot above, I realized that there's a lot of space and maybe the background isn't necessary (even though it does look nice)

## Background

Screenshot 1

## No background

Screenshot


Also notice the shortening to "Pulls" like GitHub has done for so long in the old site header

yakov116:
I like with background

134130:
When we use background, we need more width to display them.
So I think we need to find some middle one.
How about chrome's browser tabs UI?
image


Also, I think the font size is too small.
It would be nice to follow the smallest text size on GitHub.
image
On this image, i think "owner"'s font size and "updated 5 days ago" is the good size.


I also think We do not need to show label. The Icons already represent the features.

fregante:
> How about chrome's browser tabs UI?

It looks nice but I think that thin line doesn't improve on the background-less style.



> Also, I think the font size is too small.

I think the size is fine on the phone, it's only supposed to be a little more helpful.



> I also think We do not need to show label

I tried with the icons in https://github.com/refined-github/refined-github/issues/5805 but most icons are not clear. Code, issues, bugs, PRs yes… but actions, wiki, projects not really.

I'll try to restore the background but maybe reduce the spacing to 1px between items. Unfortunately I wasn't able to find a way to reduce the padding in the tabs while keeping the [icon + counter] center-aligned.

precha1989 (approved):
> 

- [ ] 
Duplicate of #

PR #6733, opened 2023-06-08
Use new cache API

fregante:
- Includes https://github.com/fregante/webext-storage-cache/pull/41

Some usage has been updated, some raw `cache.get/set` has not, because sometimes it requires a rewrite.

Question: How's the API looking? 

The main visual change is that `getSomething('a', 'b')` becomes `something.get('a', 'b')`, which leads to some less readable changes like `hasAnyProjects()` -> `hasAnyProjects.get()`

I'm thinking of using function-related method names instead, like:

- `getSomething.call('a', 'b')`


yakov116:
Mixed feelings

fregante:
I wish there was a concept of “callable class instances”, so that we could get a `getSomething()` but also `getSomething.fresh()`, without having to go back to the previous pattern. 

Maybe I don’t have a choice though, so it’s either this or:

- new CacheItem() (which is fine)
- cacheFunction() (exactly like cache.function)

I liked that CacheItem and CacheFunction had a similar style

fregante:
I don’t think there's any way to win this honestly, there's both an API to "call a function" and one to "get the results of a previous call, so you'd still have a discrepancy/awkwardness:

- `getSomething('a', 'b')`
- `getSomething.getCached('a', 'b')`

The alternative would be to go the way `mem` went with `clear`:

- `getSomething('a', 'b')`
- `getCached(getSomething, 'a', 'b')`

Which is neat but also not perfect and also hard to type.

The remaining strange part is the `set` method, which looks like:

- `getSomething.set('a', 'b', value)`

I wonder if a database-like API would make sense:

- `something.select('a', 'b')`
- `something.selectCached('a', 'b')`
- `something.selectFresh('a', 'b')`
- `something.delete('a', 'b')`

Overall I think the current get/getCached setup is easy to understand without getting too technical. I just need to update or remove the `.set` method until I figure a name (and need) for it.

yakov116:
I like your last option

source/features/bugs-tab.tsx
fregante, line 1: Currently broken:

https://github.com/axios/axios

fregante (commented):
All good so far except `bugs-tab`

source/features/bugs-tab.tsx
fregante: Due to https://github.com/fregante/webext-storage-cache/blob/9689195139b6a921c4fed03bbba9e3a8af3f47c1/source/legacy.ts#L28

args === 0 so the `cacheByRepo` `cacheKey` is ignored

PR #6772, opened 2023-07-14
Allow certain bots for issues

AaronDewes:
Right now, issues by the "issue-up" are hidden. This bot is used for forwarding bug reports between projects and its issues should not be hidden.

Example usage of the bot: https://github.com/nuxt/nuxt/issues/22133 -> https://github.com/unjs/nitro/issues/1432


fregante:
Thank you for the PR, but it's not a common situation that we can support

PR #6779, opened 2023-07-20
Restore `hide-navigation-hover-highlight` on files/notifications

fregante:
## Test URLs

- Notifications list: https://github.com/notifications
- Issue list: https://github.com/refined-github/refined-github/issues
- React file list: https://github.com/refined-github/refined-github/tree/main/.github
- File list: https://github.com/refined-github/refined-github

## Screenshot

afterr

### DIFF
--- a/source/features/hide-navigation-hover-highlight.tsx
+++ b/source/features/hide-navigation-hover-highlight.tsx
@@ -15,3 +15,14 @@ function init(): void {
 void features.add(import.meta.url, {
 	init: onetime(init),
 });
+
+/*
+
+Test URLs
+
+- Notifications list: https://github.com/notifications
+- Issue list: https://github.com/refined-github/refined-github/issues
+- React file list: https://github.com/refined-github/refined-github/tree/main/.github
+- File list: File list
+
+*/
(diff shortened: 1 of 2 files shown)


source/features/hide-navigation-hover-highlight.tsx
```diff
@@ -4,14 +4,26 @@ import onetime from 'onetime';
...
 
 const className = 'rgh-no-navigation-highlight';
+const html = document.documentElement;
 
 function init(): void {
-	document.documentElement.classList.add(className);
-	document.documentElement.addEventListener('navigation:keydown', () => {
-		document.documentElement.classList.remove(className);
+	html.classList.add(className);
+	html.addEventListener('navigation:focus', () => {
```
fregante, line 11: `navigation:keydown` was the same as `keydown`, so the style would get lost even if just `ctrl` was pressed

source/features/hide-navigation-hover-highlight.tsx
```diff
@@ -1,17 +1,28 @@
...
+		html.classList.remove(className);
 	}, {once: true});
 }
 
 void features.add(import.meta.url, {
-	init: onetime(init),
+	init,
```
fregante, line 16: Turbo drops all HTML and BODY classes, so they should be re-added every time.

PR #6792, opened 2023-07-26
Reverse `open-all-notifications` tab open order (#6755)

### LINKED ISSUE #6755: Reverse tab open order for `open-all-notifications`
### Description

Currently, when using the "Open unread" button on repositories' notifications tabs are opened in the order they appear in the document, i.e. in reverse chronological order. Especially when following only releases of a repo I'd much prefer tabs to be opened chronologically in release order, and personally feel like that would make sense for other notification types as well.

While I see a change of open order for release notifications as a clear improvement I can imagine it may be more controversial for other notifications? Alternatively, a way to sort the notifications manually - at least for the current page - might be an option? I wouldn't consider a new global feature for reversing notification order as reverse chronological order by default makes sense to me for display.

### Example URLs

prettyv:
As proposed in #6755, this PR reverses the order in which new tabs are opened when using the `open-all-notifications` feature to have them opened in chronological order (especially useful when following only releases of a repository). Implemented using a classic for-loop counting from the last index downward instead of the suggested `Array.prototype.reverse()` usage to save an array copy.

Closes #6755

## Test URLs

### DIFF
--- a/source/features/open-all-notifications.tsx
+++ b/source/features/open-all-notifications.tsx
@@ -25,8 +25,10 @@ function getUnreadNotifications(container: ParentNode = document): HTMLElement[]
 
 async function openNotifications(notifications: Element[], markAsDone = false): Promise<void> {
 	const urls: string[] = [];
-	for (const notification of notifications) {
-		urls.push(notification.querySelector('a')!.href);
+
+	// iterate in reverse to open notification tabs in chronological order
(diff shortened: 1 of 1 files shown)


fregante:
Please just use .reverse(). This kind of optimization have no place here 😁

fregante:
Ping

fregante:
Thank you ☺️

PR #6795, opened 2023-07-27
Dim unclickable reaction buttons

fregante:
This has annoyed me for quite a while: locked issues still have hoverable buttons. Notice how the button does not turn into a hand, but the background changes:

before

After:

a

Also tested on own reactions:

after

Test URL:

- https://github.com/refined-github/sandbox/pull/48
- https://togithub.com/RupertBenWiser/Web-Environment-Integrity/issues/28

### DIFF
--- a/source/features/github-bugs.css
+++ b/source/features/github-bugs.css
@@ -98,6 +98,7 @@
 }
 
 /* Disable hover state on unclickable reaction buttons */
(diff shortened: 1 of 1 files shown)


fregante:
Better yet, I think they should be muted altogether. The bright emojis and avatars still look too clickable:

Screenshot 8

So this is at 50% grayscale:

Screenshot 9

PR #6797, opened 2023-07-28
Extend `action-pr-link` to links in PRs

### LINKED ISSUE #6767: Extend `action-pr-link` to links in PRs
### Description

It looks like these links aren't currently covered:

Screenshot

### Example URLs

1. Open https://github.com/refined-github/refined-github/pull/6766
2. Click on the "x" next to the first commit
3. Click on any "Details" link

Expect:

- https://github.com/refined-github/refined-github/actions/runs/5495796892/jobs/10015405756?pr=6766

Current:

- https://github.com/refined-github/refined-github/actions/runs/5495796892/jobs/10015405756

134130:
## Description
- Closes #6767

## Test URLs
- https://github.com/refined-github/refined-github/pull/6794

## Screenshot
image

### DIFF
--- a/source/features/action-pr-link.tsx
+++ b/source/features/action-pr-link.tsx
@@ -50,6 +50,6 @@ https://github.com/refined-github/refined-github/actions
 
 https://github.com/refined-github/refined-github/pull/6794
 
-https://github.com/refined-github/refined-github/pull/6797/commits
+https://github.com/refined-github/refined-github/pull/6794/commits
 
 */
--- a/source/features/action-pr-link.tsx
+++ b/source/features/action-pr-link.tsx
@@ -55,4 +55,6 @@ void features.add(import.meta.url, {
 
 https://github.com/refined-github/refined-github/actions
 
+https://github.com/refined-github/refined-github/pull/6794
+
 */
--- a/source/features/action-pr-link.tsx
+++ b/source/features/action-pr-link.tsx
@@ -1,29 +1,30 @@
 import * as pageDetect from 'github-url-detection';
 
+import select from 'select-dom';
+
 import features from '../feature-manager.js';
 import observe from '../helpers/selector-observer.js';
-import select from "select-dom";
-import GitHubURL from "../github-helpers/github-url.js";
+import GitHubURL from '../github-helpers/github-url.js';
 
-function setSearchParam(anchorElement: HTMLAnchorElement, name: string, value: string) {
-	const params = new URLSearchParams(anchorElement.search);
-	params.set(name, value)
-	anchorElement.search = String(params)
+function setSearchParameter(anchorElement: HTMLAnchorElement, name: string, value: string): void {
+	const parameters = new URLSearchParams(anchorElement.search);
+	parameters.set(name, value);
+	anchorElement.search = String(parameters);
 }
 
 async function addForRepositoryActions(prLink: HTMLAnchorElement): Promise<void> {
 	const prNumber = prLink.textContent!.slice(1);
 
 	const runLink = prLink.closest('.Box-row')!.querySelector('a.Link--primary')!;
-	setSearchParam(runLink, 'pr', prNumber)
+	setSearchParameter(runLink, 'pr', prNumber);
 }
 
 async function addForPR(dropdown: HTMLDivElement): Promise<void> {
 	const {branch: prNumber} = new GitHubURL(location.href);
-	const actions = select.all('div.merge-status-item:has(a[href="/apps/github-actions"]) a.status-actions', dropdown)
+	const actions = select.all('div.merge-status-item:has(a[href="/apps/github-actions"]) a.status-actions', dropdown);
 
-	for(const action of actions) {
-		setSearchParam(action, 'pr', prNumber);
+	for (const action of actions) {
+		setSearchParameter(action, 'pr', prNumber);
 	}
 }
 
@@ -44,7 +45,7 @@ void features.add(import.meta.url, {
 	include: [
 		pageDetect.isPR,
 	],
-	init: initForPRPage
+	init: initForPRPage,
 });
 
 /*
--- a/source/features/action-pr-link.tsx
+++ b/source/features/action-pr-link.tsx
@@ -22,8 +22,6 @@ async function addForPR(dropdown: HTMLDivElement): Promise<void> {
 	const {branch: prNumber} = new GitHubURL(location.href);
 	const actions = select.all('div.merge-status-item:has(a[href="/apps/github-actions"]) a.status-actions', dropdown)
 
-	// TODO: The actions is empty when ajax response is slow
-
 	for(const action of actions) {
 		setSearchParam(action, 'pr', prNumber);
 	}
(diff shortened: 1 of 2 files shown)


source/features/action-pr-link.tsx
```diff
@@ -2,25 +2,49 @@ import * as pageDetect from 'github-url-detection';
...
 	const runLink = prLink.closest('.Box-row')!.querySelector('a.Link--primary')!;
-	const parameters = new URLSearchParams(runLink.search);
-	parameters.set('pr', prNumber);
-	runLink.search = String(parameters);
+	setSearchParam(runLink, 'pr', prNumber)
+}
+
+async function addForPR(dropdown: HTMLDivElement): Promise<void> {
+	const {branch: prNumber} = new GitHubURL(location.href);
+	const actions = select.all('div.merge-status-item:has(a[href="/apps/github-actions"]) a.status-actions', dropdown)
```
fregante, line 23: I suggest selecting the links in the `observe`, not by selecting the dropdown first

source/features/action-pr-link.tsx
```diff
@@ -2,25 +2,44 @@ import * as pageDetect from 'github-url-detection';
...
+	const {branch: prNumber} = new GitHubURL(location.href);
+	setSearchParameter(actionLink, 'pr', prNumber);
+}
+
+async function initForRepositoryActionsPage(signal: AbortSignal): Promise<void> {
+	observe('div.Box-row[id^=check_suite_] a[data-hovercard-type="pull_request"]', addForRepositoryActions, {signal});
+}
+
+async function initForPRPage(signal: AbortSignal): Promise<void> {
+	observe('div.js-timeline-item details.commit-build-statuses div.merge-status-item:has(a[href="/apps/github-actions"]) a.status-actions', addForPR, {signal});
```
fregante, line 30: This is a bit too specific. It should also work on https://github.com/refined-github/refined-github/pull/6797/commits

source/features/action-pr-link.tsx
```diff
@@ -2,25 +2,44 @@ import * as pageDetect from 'github-url-detection';
...
-	parameters.set('pr', prNumber);
-	runLink.search = String(parameters);
+	setSearchParameter(runLink, 'pr', prNumber);
 }
 
-async function init(signal: AbortSignal): Promise<void> {
-	observe('div.Box-row[id^=check_suite_] a[data-hovercard-type="pull_request"]', add, {signal});
+async function addForPR(actionLink: HTMLAnchorElement): Promise<void> {
+	const {branch: prNumber} = new GitHubURL(location.href);
+	setSearchParameter(actionLink, 'pr', prNumber);
```
fregante, line 22: `GitHubURL` is only for for files. Use `getConversationNumber` instead

source/features/action-pr-link.tsx
134130: FYI. Action statuses are only appeared to repository owners
But I'll implement for commits page.

I intentionally detailed before since this one.
image

source/features/action-pr-link.tsx
134130: I think the utilities are little bit huddle for new contributors.
It will be better when the utilities are scoped like `GitHubURL.of(location.href) returns GitHubConversationURL`. and it will make better intellisense. But this is just my opinion as user of strongly typed language like Kotlin and C#.

source/features/action-pr-link.tsx
```diff
@@ -2,25 +2,49 @@ import * as pageDetect from 'github-url-detection';
...
 }
 
-async function init(signal: AbortSignal): Promise<void> {
-	observe('div.Box-row[id^=check_suite_] a[data-hovercard-type="pull_request"]', add, {signal});
+async function initForRepositoryActionsPage(signal: AbortSignal): Promise<void> {
+	observe('div.Box-row[id^=check_suite_] a[data-hovercard-type="pull_request"]', addForRepositoryActions, {signal});
+}
+
+async function initForPRPage(signal: AbortSignal): Promise<void> {
+	observe(':not(span.rgh-ci-link) > details.commit-build-statuses div.merge-status-item:has(a[href="/apps/github-actions"]) a.status-actions', addForPR, {signal});
```
fregante, line 29: Still too specific. The tag name is rarely needed. We usually specify it only when we want to get a specific tag in typescript, for example in this case we specify the selected `a` element.

This works

```suggestion
	// Exclude rgh-link, include isPRCommits
	observe('main [href="/apps/github-actions"] ~ div a.status-actions', addForPR, {signal});
```

source/features/action-pr-link.tsx
```diff
@@ -29,4 +53,8 @@ void features.add(import.meta.url, {
 
 https://github.com/refined-github/refined-github/actions
 
+https://github.com/refined-github/refined-github/pull/6794
+
+https://github.com/refined-github/refined-github/pull/6797/commits
```
fregante, line 52: It doesn't appear on this page. These "test urls" must work.

This works:
```suggestion
https://github.com/refined-github/refined-github/pull/6794/commits
```

source/features/action-pr-link.tsx
fregante: > GitHubURL.of(location.href) returns GitHubConversationURL

This isn't possible in TypeScript. The returned type depends on a runtime string so the compile-time type would be `GitHubFileURL | GitHubConversationURL`, which isn't particularly useful.

`GitHubURL` exists because extracting the branch information from URLs can become complex, for example in case of branches that looks like `feature/something-something`. Extracting the number is instead very easy, it's basically a `href.split('/')[4]` and some checks.

I agree that `GitHubURL` is too generically-named so it should probably be renamed to `GitHubFile` (separate PR welcome)

source/features/action-pr-link.tsx
134130: Looks much better :D

source/features/action-pr-link.tsx
134130: Ooops, sorry for the confusion.
More typescript-ly, I suggested like this one.

```typescript
// Like this
const features = {
	add,
	unload,
	addCssFeature,
	log,
	shortcutMap,
	getFeatureID,
	getIdentifiers,
};

export default features;

// Suggestion (The naming is too java style, but the structure)
function ofFile(): GitHubFile { ... }
function ofConversation(): GitHubConversation { ... }

const GitHubURL = {
    ofFile,
    ofConversation, // ofPullRequest | ofIssue
}

export default GitHubURLs;
```

But I think this is not generic in typescript ecosystem, and it could be because I am not familiar with it.
Never mind! Thank you for the detailed explanation.

fregante: > Also, my suggestion seems to require an over-implementation

Yeah, GitHubURL returns a lot of information for file paths, but:

- other paths are simpler, don't need such specific parsing, or are just not parsed often enough
- GitHub has a _lot_ of different URLs, as can be seen on https://refined-github.github.io/github-url-detection/

While it'd be nice to have a consistent API, it's definitely not worth the effort

Issue #5828, opened 2022-07-14
Show `rgh-` features in the options list?


fregante:
Thoughts on showing our "private" features in the options? We can move them to the bottom.

- `rgh-welcome-issue`
- `rgh-sponsor-button`
- `rgh-linkify-features`
- `rgh-feature-descriptions`
- `rgh-improve-new-issue-form`

The main annoyance here is that it would be difficult to document these since our descriptions live in readme.md, but we can probably wrap a whole `rgh` section in `<!-- -->` so it exists in the file but it isn't rendered in the browser.

### Related

- https://github.com/refined-github/refined-github/issues/4288
- https://github.com/refined-github/refined-github/issues/4163
- https://github.com/refined-github/refined-github/issues/5668
- https://github.com/refined-github/refined-github/wiki/Meta-features

fregante:
Since these features apply to our repos exclusively and are for our own advantage, like `rgh-improve-new-issue-form`, I don't want people to be able to disable them.

The reason why I opened this issue is similar to #5668: _we make changes that can't be disabled._

These features don't affect the rest of the site so it makes even less sense to disable them. We can just extend the bisect message to include this piece of information instead, exactly like for #5668.

Issue #5668, opened 2022-06-01
"Find feature" gives incorrect information for CSS features


fregante:
We can't disable CSS features so our flow will end up reporting the wrong information. 

Perhaps we can start the flow with a step without any features enabled. If the user can see the bug we'll say something like:

- Refined GitHub can't disable this. It may be a CSS-only feature or completely unrelated to Refined GitHub

cheap-glitch:
+1 but:

```diff
-that can't be disabled
+that can (only) be disabled via the "Custom CSS" field on the options page
```

fregante:
👌 We can add a link to https://github.com/refined-github/refined-github/issues/5501

fregante:
Related information:

- https://github.com/refined-github/refined-github/issues/5828

We can add a link to the wiki for extended details that we can change and improve at any time.

> Unable to identify feature. It might be a CSS-only, a meta-feature or unrelated to Refined GitHub. More info on our [wiki](#page-to-be-defined)

PR welcome

PR #6828, opened 2023-08-14
Change the message displayed when bisection fails

LilithHafner:
This changes the text displayed after a failed bisection so that it is less misleading. 

This PR basically implements @cheap-glitch's [suggestion](https://github.com/refined-github/refined-github/issues/5668#issuecomment-1143189820) 
Helps with #5668, but does not close that issue because it would still be nice to start with a step detecting this as suggested [here](https://github.com/refined-github/refined-github/issues/5668#issue-1254939578)
The word "public" and the fact that I say "that cannot be disabled" rather than "CSS feature" are due to #5828

### DIFF
--- a/source/helpers/bisect.tsx
+++ b/source/helpers/bisect.tsx
@@ -33,7 +33,7 @@ async function onChoiceButtonClick({currentTarget: button}: React.MouseEvent<HTM
 
 	// Last step, no JS feature was enabled
 	if (answer === 'yes') {
-		createMessageBox('No features were enabled on this page. Try disabling Refined GitHub to see if the change or issue is caused by the extension.');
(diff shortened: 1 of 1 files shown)


source/helpers/bisect.tsx
```diff
@@ -33,7 +33,7 @@ async function onChoiceButtonClick({currentTarget: button}: React.MouseEvent<HTM
 
 	// Last step, no JS feature was enabled
 	if (answer === 'yes') {
-		createMessageBox('No features were enabled on this page. Try disabling Refined GitHub to see if the change or issue is caused by the extension.');
+		createMessageBox('No public javascript features were enabled on this page. It could be a Refined GitHub feature that cannot be disabled or it could be unrelated to Refined Github. Try disabling Refined GitHub to find out which.');
```
fregante, line 36: The suggested copy was:

> Unable to identify feature. It might be a CSS-only, a meta-feature or unrelated to Refined GitHub. More info on our wiki

fregante (commented):
Thank you for the PR!

PR #6837, opened 2023-08-17
Improve support of New Repository Overview

### LINKED ISSUE #6818: refined github throws error on `New Repository Overview`
### Description

When enabling `New Repository Overview` feature, refined github throws 
```console
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'title')
    at extractCurrentBranchFromBranchPicker (refined-github.js:3455:57)
    at refined-github.js:4646:81
    at async CachedFunction.updater (refined-github.js:4645:121)
    at async getSet (refined-github.js:2344:30)
```

enable `New Repository Overview` and visit any repository 

### Extension version

23.8.3

### Browser(s) used

chrome

134130:
## Describe
- Closes #6818

When visiting some pages which using `getDefaultBranch()` API, like repository's main page with enabling GitHub feature preview: `New Repository Overview`, The below error is occuring.
```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'title')
    at extractCurrentBranchFromBranchPicker (refined-github.js:3455:57)
    at refined-github.js:4646:81
    at async CachedFunction.updater (refined-github.js:4645:121)
    at async getSet (refined-github.js:2344:30)
```

- The reason is `getDefaultBranch()` uses DOM-Selector
  - `getDefaultBranch()` -> `fromDOM()` uses `export const branchSelector = '[data-hotkey="w"]';`
  - We can fix the #6818 with adding option `awaitDomReady: true` to `repo-wide-file-finder`
  - However, the other rgh-features which using `getDefaultBranch()` and doesn't added `awaitDomReady: true` will occur same issue
  - This error is not occurred frequently since `getDefaultBranch()` uses caching.
  - Add awaiting-DOM to `fromDOM()` function to make function it works self

## Test URLs

- Enable GitHub feature preview: `New Repository Overview`
  - Visit some new repositories where you are as owner

## Screenshot
- no screenshot

### DIFF
--- a/source/github-helpers/get-default-branch.ts
+++ b/source/github-helpers/get-default-branch.ts
@@ -11,7 +11,7 @@ const isCurrentRepo = ({nameWithOwner}: RepositoryInfo): boolean => Boolean(getR
 // Do not make this function complicated. We're only optimizing for the repo root.
 async function fromDOM(): Promise<string | undefined> {
 	if (!['', 'commits'].includes(getRepo()!.path)) {
-		return undefined;
+		return;
 	}
 
 	// We're on the default branch, so we can extract it from the current page. This exclusively happens on the exact pages:
@@ -20,7 +20,7 @@ async function fromDOM(): Promise<string | undefined> {
 	const element = await elementReady(branchSelector);
 
 	if (!element) {
-		return undefined;
+		return;
 	}
 
 	return extractCurrentBranchFromBranchPicker(element);
--- a/source/github-helpers/get-default-branch.ts
+++ b/source/github-helpers/get-default-branch.ts
@@ -1,9 +1,7 @@
 import {CachedFunction} from 'webext-storage-cache';
 import elementReady from 'element-ready';
 import {type RepositoryInfo} from 'github-url-detection';
-
 import domLoaded from 'dom-loaded';
-
 import delay from 'delay';
 
 import api from './api.js';
--- a/source/github-helpers/get-default-branch.ts
+++ b/source/github-helpers/get-default-branch.ts
@@ -2,6 +2,10 @@ import {CachedFunction} from 'webext-storage-cache';
 import elementReady from 'element-ready';
 import {type RepositoryInfo} from 'github-url-detection';
 
+import domLoaded from 'dom-loaded';
+
+import delay from 'delay';
+
 import api from './api.js';
 import {extractCurrentBranchFromBranchPicker, getRepo} from './index.js';
 import {branchSelector} from './selectors.js';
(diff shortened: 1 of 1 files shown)


source/github-helpers/get-default-branch.ts
```diff
@@ -14,6 +18,9 @@ async function fromDOM(): Promise<string | undefined> {
 		return undefined;
 	}
 
+	await domLoaded; // DOM-based filter
+	await delay(100);
```
fregante, line 22: `delay` is generally a last resort, but `domLoaded` should not be needed in the first place, we already have `elementReady`.

source/github-helpers/get-default-branch.ts
```diff
@@ -2,6 +2,10 @@ import {CachedFunction} from 'webext-storage-cache';
 import elementReady from 'element-ready';
 import {type RepositoryInfo} from 'github-url-detection';
 
+import domLoaded from 'dom-loaded';
+
+import delay from 'delay';
```
fregante, line 7:
```suggestion
import domLoaded from 'dom-loaded';
import delay from 'delay';
```

fregante (changes requested):

source/github-helpers/get-default-branch.ts
134130: Yeah, I know that elementReady means same works.
But the elementReady or domLoaded not works exactly in this case.

134130: We can resolve this with this change on `element-ready`
https://github.com/sindresorhus/element-ready/blob/036c1085b10ef01313ec60ef183f399dfdd05322/index.js#L7-L8C34

```js
// AS-IS
const isDomReady = target =>
	['interactive', 'complete'].includes((target.ownerDocument || target).readyState);

// TO-BE
const isDomReady = target =>
	['complete'].includes((target.ownerDocument || target).readyState);
```

fregante: There's a misunderstanding in how these two work. If you wait for the whole document to be ready, then you don't need elementReady, because at that point it behaves as a simple `querySelector`/`select`. If the element is not part of the original document, then neither dom-loaded nor element-ready should be used. 

If you're looking for an element/component that's loaded dynamically, it's best to use the selector observer. Or maybe in this case it's easier to use `elementReady(selector, {stopOnDomLoaded: false, timeout: 1000})`

134130: We need to refactor this API 🤔

134130: I fixed just returning `undefined` when elementReady failed to find element

fregante (approved):
If this works, it's perfect!

source/github-helpers/get-default-branch.ts
```diff
@@ -17,7 +17,13 @@ async function fromDOM(): Promise<string | undefined> {
 	// We're on the default branch, so we can extract it from the current page. This exclusively happens on the exact pages:
 	// /user/repo
 	// /user/repo/commits (without further path)
-	return extractCurrentBranchFromBranchPicker((await elementReady(branchSelector))!);
+	const element = await elementReady(branchSelector);
+
+	if (!element) {
+		return undefined;
```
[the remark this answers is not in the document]
134130: Looks nice 😎

PR #6839, opened 2023-08-18
Fix `conflict-marker` duplicated

### LINKED ISSUE #6809: `conflict-marker` duplicated
### Description

`conflict-marker`'s warning button is duplicating when page back.

image

1. https://github.com/pulls
2. Go to any page.
3. Go back to previous page

### Extension version

23.7.25

### Browser(s) used

Chrome 115.0.5790.114 (arm64)

134130:
## Description
- Closes #6809
- Use new patterns
  - element observing
  - seperate gql

## Test URLs
- https://github.com/pulls
- https://github.com/refined-github/refined-github/pulls

## Screenshot
image

### DIFF
--- a/source/features/conflict-marker.tsx
+++ b/source/features/conflict-marker.tsx
@@ -60,6 +60,6 @@ void features.add(import.meta.url, {
 /*
 Test URLs
 https://github.com/pulls
-https://github.com/refined-github/refined-github/pulls
+https://github.com/refined-github/sandbox/issues?q=conflict
 https://github.com/kubernetes/kubernetes/milestone/62
 */
--- a/source/features/conflict-marker.tsx
+++ b/source/features/conflict-marker.tsx
@@ -30,6 +30,10 @@ async function addConflictMarkers(container: HTMLDivElement): Promise<void> {
 		};
 	});
 
+	if (prConfigs.length === 0) {
+		return;
+	}
+
 	const batchQuery = prConfigs.map(prConfig =>
 		createQueryFragment(prConfig.key, prConfig.user, prConfig.repo, prConfig.number),
 	).join('\n');
--- a/source/features/conflict-marker.tsx
+++ b/source/features/conflict-marker.tsx
@@ -1,14 +1,13 @@
 import './conflict-marker.css';
 import React from 'dom-chef';
-import select from 'select-dom';
 import {AlertIcon} from '@primer/octicons-react';
-import oneMutation from 'one-mutation';
 import * as pageDetect from 'github-url-detection';
 
 import features from '../feature-manager.js';
 import api from '../github-helpers/api.js';
 import observe from "../helpers/selector-observer.js";
 
+// TODO: Remove type
 type PRConfig = {
 	number: string;
 	user: string;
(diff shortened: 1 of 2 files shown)


fregante:
This unfortunately triggers 25 HTTP requests every time you open the pull requests page, which isn't acceptable. We need to use `batched-function` to compose the old GraphQL call and inside it add the icons to the DOM. I don’t think we used this exact pattern in the extension yet but we need to start. The .gql file cannot be used in this case because of the composed GraphQL call.

134130:
> This unfortunately triggers 25 HTTP requests every time you open the pull requests page, which isn't acceptable. We need to use `batched-function` to compose the old GraphQL call and inside it add the icons to the DOM. I don’t think we used this exact pattern in the extension yet but we need to start. The .gql file cannot be used in this case because of the composed GraphQL call.

Ok I'll make bunch API call as it was before

source/features/conflict-marker.tsx
```diff
@@ -72,25 +51,22 @@ async function init(): Promise<false | void> {
 	}
 }
 
+function init(signal: AbortSignal): void {
+	observe('#js-issues-toolbar', addConflictMarkers, {signal});
```
fregante, line 55: What is this element? Does it exist on the milestones page? Does it work there now?

source/features/conflict-marker.tsx
```diff
@@ -1,63 +1,42 @@
...
 			}
 		}
 	`;
 }
 
-function buildQuery(prs: PRConfig[]): string {
-	return prs.map(pr => createQueryFragment(pr)).join('\n');
-}
+async function addConflictMarkers(container: HTMLDivElement): Promise<void> {
+	const links = select.all('.js-issue-row:has(.octicon-git-pull-request.color-fg-open) a.js-navigation-open', container);
```
fregante, line 23: The feature will likely fail when there are 0 links

source/features/conflict-marker.tsx
134130: It works ;)
image

source/features/conflict-marker.tsx
134130: Thank you! It was fixed.

source/features/conflict-marker.tsx
fregante: I think this will fail on slower connections because the observer will call the callback as soon as `#js-issues-toolbar` is found, but at that point the long list of PRs might not be done downloading yet. This is why I was suggesting to use batched-function and the link selector

```js
const addIcon = batchedFunction((icons) => {
  // create query, etc
});

observe('.js-issue-row:has(.octicon-git-pull-request.color-fg-open) a.js-navigation-open', addIcon, {signal})
```

134130: 🤔 I've tested a lot but It can detect icons successly. But I'll test more slow network conditions.
And is your suggestion works correctly? As I know, the selector-observe only works as single element, not all elements.

fregante: > As I know, the selector-observe only works as single element, not all elements

No, the observer detects any and all elements that appear on the page until the next page load.

Here's what happens with the suggested code:

1. `observe` finds 3 elements
2. calls `addIcon` once per element
3. batched function then calls the anonymous function with a single array: `[prLink1, prLink2, prLink3]`
4. you can use this array to construct the GQL call
5. if more items are detected later, it repeats from step 1

On fast connections maybe batchedFunction will call the function with 25 items, on slower connections it'll make multiple API calls, that's ok

134130: Exactly, that was I thought.
I'm just wondering is there existing some utility function: batchedFunction, and the batchedFunction's throttle (or debounce) time.

fregante: What do you mean? It does exist and we're already using it. We don't need to change it at this point https://github.com/Richienb/batched-function

fregante:
Nice, that looks pretty simple to implement in more features

fregante (approved):

source/features/conflict-marker.tsx
```diff
@@ -1,96 +1,65 @@
...
-	deduplicate: 'has-rgh',
-	awaitDomReady: true, // TODO: Use observe + batched-function
 	init,
 });
+
+/*
+Test URLs
+https://github.com/pulls
+https://github.com/refined-github/sandbox/issues?q=conflict
+https://github.com/kubernetes/kubernetes/milestone/62
```
fregante, line 64: This section exists to simplify testing so we should try to use URLs that do and always will include the feature. Our sandbox repo is where these scenarios exist and can be created.

source/features/conflict-marker.tsx
```diff
@@ -1,96 +1,65 @@
...
-				</a>,
+					<AlertIcon className="v-align-middle" />
+				</a>
 			);
 		}
 	}
 }
 
+function init(signal: AbortSignal): void {
+	observe('.js-issue-row:has(.octicon-git-pull-request.color-fg-open) a.js-navigation-open', batchedFunction(addIcon, {delay: 100}), {signal});
```
fregante, line 47: Without a `delay` it's too fast even for high-speed internet:

Screenshot

100ms seems to fix it:

Screenshot 1



We have 2 more functions that use `batchedFunction` and likely suffer the same issue. PR welcome in that case too, inlining the `batchedFunction()` call this way

fregante (approved):

PR #6840, opened 2023-08-18
Feat: check default branch

### LINKED ISSUE #6413: `useful-not-found` broken on new 404 repo page
### Description

The page finally looks nice, but it breaks our feature (and maybe even our detection)

Screenshot 3

1. https://github.com/fregante/GhostText/tree/3cacd7df71b097dc525d99c7aa2f54d31b02fcc8/chrome/scripts/InputArea

### Extension version

23.3.2

### Browser(s) used

Safari

134130:
## Description
- Closes #6413

As @fregante said on https://github.com/refined-github/refined-github/issues/6413#issuecomment-1683661277
> - ensure that `default-branch-button` works on this page
> - run checkAnchor on the existing breadcrumbs
> - run showAlternateLink to know what happened to the file
> - add isRepoFile404 to the exclude array of the existing configuration

## Test URLs
- https://github.com/fregante/GhostText/tree/3cacd7df71b097dc525d99c7aa2f54d31b02fcc8/chrome/scripts/InputArea
- https://github.com/refined-github/refined-github/blob/some-non-existent-ref/source/features/bugs-tab.tsx

## Screenshot
image

### DIFF
--- a/source/features/default-branch-button.tsx
+++ b/source/features/default-branch-button.tsx
@@ -12,6 +12,11 @@ import {branchSelector} from '../github-helpers/selectors.js';
 import isDefaultBranch from '../github-helpers/is-default-branch.js';
 import {isRepoCommitListRoot} from '../github-helpers/index.js';
 
+async function is404(url: string): Promise<boolean> {
+	const {status} = await fetch(url, {method: 'head'});
+	return status === 404;
+}
+
 async function add(branchSelector: HTMLElement): Promise<void> {
 	// Don't show the button if we’re already on the default branch
 	// TODO: Move to `asLongAs` when it accepts async detections
@@ -38,6 +43,12 @@ async function add(branchSelector: HTMLElement): Promise<void> {
 			<ChevronLeftIcon/>
 		</a>
 	);
+
+	if (pageDetect.isRepoFile404() && await is404(url.href)) {
+		defaultLink.removeAttribute('href');
+		defaultLink.setAttribute('aria-disabled', 'true');
+		defaultLink.setAttribute('aria-label', 'This view is not exists on the default branch');
+	}
 
 	// The DOM varies between details-based DOM and React-based one
 	const selectorWrapper = branchSelector.tagName === 'SUMMARY'
--- a/source/features/toggle-files-button.tsx
+++ b/source/features/toggle-files-button.tsx
@@ -87,6 +87,9 @@ void features.add(import.meta.url, {
 	include: [
 		pageDetect.isRepoTree,
 	],
+	exclude: [
+		pageDetect.isRepoFile404,
+	],
 	init,
 });
 
--- a/source/features/useful-not-found-page.tsx
+++ b/source/features/useful-not-found-page.tsx
@@ -9,7 +9,7 @@ import api from '../github-helpers/api.js';
 import GitHubURL from '../github-helpers/github-url.js';
 import getDefaultBranch from '../github-helpers/get-default-branch.js';
 import {getCleanPathname} from '../github-helpers/index.js';
-import observe from "../helpers/selector-observer.js";
+import observe from '../helpers/selector-observer.js';
 
 type File = {
 	previous_filename?: string;
@@ -151,7 +151,7 @@ async function showDefaultBranchLink(): Promise<void> {
 	);
 }
 
-async function getAlternateLinkMessage(): Promise<string | undefined> {
+async function getAlternateLink(): Promise<HTMLElement | undefined> {
 	const url = new GitHubURL(location.href);
 	if (!url.branch || !url.filePath) {
 		return;
@@ -162,9 +162,21 @@ async function getAlternateLinkMessage(): Promise<string | undefined> {
 		return;
 	}
 
-	const fileChanges = await getChangesToFileInCommit(commitSha, url.filePath);
+	let fileChanges = await getChangesToFileInCommit(commitSha, url.filePath);
 	if (!fileChanges) {
-		return;
+		fileChanges = {
+			file: {
+				filename: 'filename',
+				status: 'active',
+				blob_url: '1234',
+			},
+			commit: {
+				parentSha: 'parentSha',
+				date: new Date(),
+				url: 'url',
+			},
+		};
+		// Return;
 	}
 
 	url.assign({route: 'commits'});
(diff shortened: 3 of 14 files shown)


source/features/useful-not-found-page.tsx
```diff
@@ -40,6 +41,7 @@ function getStrikeThrough(text: string): HTMLElement {
 }
 
 async function checkAnchor(anchor: HTMLElement): Promise<void> {
+	console.log(anchor);
```
fregante, line 44: Note

source/features/toggle-files-button.tsx
```diff
@@ -87,6 +87,9 @@ void features.add(import.meta.url, {
 	include: [
 		pageDetect.isRepoTree,
 	],
+	exclude: [
+		pageDetect.isRepoFile404,
+	],
```
fregante, line 92: Oof these are a lot of manual exclusions. Ok for now but eventually we'd want to silence errors on 404 and empty pages.

- https://github.com/refined-github/refined-github/issues/6558

source/features/default-branch-button.tsx
```diff
@@ -39,6 +44,12 @@ async function add(branchSelector: HTMLElement): Promise<void> {
 		</a>
 	);
 
+	if (pageDetect.isRepoFile404() && await is404(url.href)) {
```
fregante, line 47: If it's a 404, then we don't add the link at all. This is what we used to do on the old 404 page too

source/features/useful-not-found-page.tsx
```diff
@@ -199,15 +221,19 @@ async function initPRCommit(): Promise<void | false> {
 	);
 }
 
-void features.add(import.meta.url, 	{
+function initRepoFile(signal: AbortSignal): void {
+	observe('main #repos-header-breadcrumb-wide-heading+ol a', checkAnchor, {signal});
+	observe('main div[data-testid="eror-404-description"]', addAlternateLinkForRepoFile, {signal});	// Eror is not rgh's typo
```
fregante, line 226:
```suggestion
	observe('main div[data-testid="eror-404-description"]', addAlternateLinkForRepoFile, {signal});	// "eror" as misspelled by GitHub
```

source/features/useful-not-found-page.tsx
```diff
@@ -199,15 +221,19 @@ async function initPRCommit(): Promise<void | false> {
 	);
 }
 
-void features.add(import.meta.url, 	{
+function initRepoFile(signal: AbortSignal): void {
+	observe('main #repos-header-breadcrumb-wide-heading+ol a', checkAnchor, {signal});
```
fregante, line 225:
```suggestion
	observe('#repos-header-breadcrumb-wide-heading + ol a', checkAnchor, {signal});
```

source/features/useful-not-found-page.tsx
```diff
@@ -223,6 +249,15 @@ void features.add(import.meta.url, 	{
 		pageDetect.isPRCommit404,
 	],
 	init: onetime(initPRCommit),
+}, {
+	asLongAs: [
+		pageDetect.isRepoFile404,
```
fregante, line 254: This already includes isSingleFile and isRepoTree, right? We don't need the extra `include`

source/features/useful-not-found-page.tsx
```diff
@@ -223,6 +249,15 @@ void features.add(import.meta.url, 	{
...
+	],
+	include: [
+		pageDetect.isRepoTree,
+		pageDetect.isSingleFile,
+	],
+	init: onetime(initRepoFile),
```
fregante, line 260:
```suggestion
	init: initRepoFile,
```

The observers are disabled on page unload via the AbortSignal, so we can't and don't need to use `onetime`

source/features/default-branch-button.tsx
```diff
@@ -12,6 +12,11 @@ import {branchSelector} from '../github-helpers/selectors.js';
 import isDefaultBranch from '../github-helpers/is-default-branch.js';
 import {isRepoCommitListRoot} from '../github-helpers/index.js';
 
+async function is404(url: string): Promise<boolean> {
+	const {status} = await fetch(url, {method: 'head'});
+	return status === 404;
```
fregante, line 17: Is this a copy? We can just move the function to the helpers/index.js file

source/features/default-branch-button.tsx
```diff
@@ -39,6 +39,11 @@ async function add(branchSelector: HTMLElement): Promise<void> {
 		</a>
 	);
 
+	// Don't show the button if the file is not exists on the default branch
+	if (await is404Page(url.href)) {
```
fregante, line 43: On second though, I don't want this button to be delayed on every page load, so we should always add it regardless of 404 status. We can drop this condition

source/features/useful-not-found-page.tsx
```diff
@@ -223,6 +243,11 @@ void features.add(import.meta.url, 	{
 		pageDetect.isPRCommit404,
 	],
 	init: onetime(initPRCommit),
+}, {
+	asLongAs: [
```
fregante, line 247:
```suggestion
	include: [
```

`include` is the default. `asLongAs` is only used when every condition must be true

fregante (approved):
LGTM so far. I'll see how it looks

source/features/default-branch-button.tsx
134130: I'll check is404 only on RepoFile404 page.

source/features/default-branch-button.tsx
```diff
@@ -30,7 +30,7 @@ async function add(branchSelector: HTMLElement): Promise<void> {
 
 	const defaultLink = (
 		<a
-			className="btn tooltipped tooltipped-ne px-2"
+			className="btn tooltipped tooltipped-se px-2"
```
fregante, line 33: Before:

Screenshot 6

After:

Screenshot 7

fregante (approved):
Should be good. Review?

source/features/useful-not-found-page.tsx
```diff
@@ -170,26 +166,46 @@ async function showAlternateLink(): Promise<void> {
...
+	if (link) {
+		select('main > .container-lg')!.before(link);
+	}
+}
+
+async function showGitObjectHistoryOnRepo(description: HTMLDivElement): Promise<void> {
+	const link = await getGitObjectHistoryLink();
+	if (link) {
+		link.className = description.className;
+		link.style.display = 'inline';
```
fregante, line 196: Their DOM sucks:

Screenshot 9

They're hiring just about anyone at GitHub. What a dumpster fire. Let's just use utility classes:


```suggestion
		link.className = 'color-fg-muted';
```

134130: 🤔 How can we resolve this CLS issue? using skeleton?
화면 기록 2023-08-27 23 47 12

fregante:
Thank you!

fregante:
> 🤔 How can we avoid this CLS issue? use skeleton?

Skeletons are hard to maintain and the text will still wrap on mobile, causing extra space.

I'd rather just move the text below the green button or drop the button altogether since it's a super-duplicate button

Screenshot 7


(orange slash = duplicate only when already on the main branch)

source/github-helpers/index.ts
```diff
@@ -138,6 +138,11 @@ export const cacheByRepo = (): string => getRepo()!.nameWithOwner;
 // Commit lists for files and folders lack a branch selector
 export const isRepoCommitListRoot = (): boolean => pageDetect.isRepoCommitList() && document.title.startsWith('Commits');
 
+export const isUrlReachable = mem(async (url: string): Promise<boolean> => {
```
jerone, line 141:
```suggestion
export const isUrlUnreachable = mem(async (url: string): Promise<boolean> => {
```

134130: 🥹

fregante: It looks like I made a couple of mistakes when renaming:

- I meant to change the 404 check to just use `response.ok`
- I forgot to flip every condition to actually match "is reachable", to avoid double negations

PR #6872, opened 2023-09-03
Meta: Fix some TODOs

fregante:
From https://github.com/refined-github/refined-github/issues/6000#issuecomment-1521431676

What I didn't touch:

- TODOs dated in the future
- selector-observer TODOs, for the most part
- has-selector-related TODOs
- TODOs dependent on external factors/libraries

### DIFF
--- a/source/features/hide-newsfeed-noise.css
+++ b/source/features/hide-newsfeed-noise.css
@@ -9,7 +9,6 @@ published a release
 edited a gollum (wiki)
 */
 
-/* TODO: Drop old classes, no longer used */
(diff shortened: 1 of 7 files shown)


source/features/hide-newsfeed-noise.css
```diff
@@ -9,7 +9,6 @@ published a release
 edited a gollum (wiki)
 */
 
-/* TODO: Drop old classes, no longer used */
```
fregante, line 12: Not hurting anyone. Some classes are still used.

PR #6874, opened 2023-09-03
Target macOS 13, iOS 16, Safari 16.4

sindresorhus:
I have recreated the Xcode project from scratch, which massively simplifies the container app. It's now just a single SwiftUI app for both macOS and iOS.


sindresorhus:
@fregante Do you remember how you managed to include the "assets" thing in https://github.com/refined-github/refined-github/pull/6612? That's the only thing left. Maybe you could do it again.

sindresorhus:
The window is currently resizable on macOS. We need to target macOS 13 to be able to lock the window size.

fregante:
If you drag the assets folder into the resources window it will add just the folder, not its contents

sindresorhus:
That did it 👍

sindresorhus:
I assume because of the new target iOS version, this is now required: https://github.com/refined-github/refined-github/pull/6874/commits/fa99606fa6648b7835f8cb1ba744007c33a5503b

Screenshot 2023-09-03 at 19 06 27

It works after adding that.

sindresorhus:
@fregante This is now done.

source/manifest.json
```diff
@@ -35,7 +35,8 @@
 	"background": {
 		"scripts": [
 			"assets/background.js"
-		]
+		],
+		"persistent": false
```
fregante, line 39: This has always been a requirement on iOS but we dropped it because it caused issues with GHE if I remember correctly. It has been added via the Safari preparation script, which needs to be updated as part of this PR as well, instead of altering the manifest.

sindresorhus: Ah ok. Reverted: https://github.com/refined-github/refined-github/pull/6874/commits/ab8efd73cde824b5135a2e9cf2df5b97d36a8685

I had not run the prepare script as I was not doing a new version.

fregante:
A couple of issues:

- the app doesn't quit when the window is closed
- the app starts with a wide window
-

sindresorhus:
These things require macOS 13. I'm too lazy to polyfill when we can target macOS 13 in a month. I don't think this should hold back merging.

sindresorhus:
I think this is reason enough to just target macOS 13 now.

fregante:
> These things require macOS 13

Sizing a window requires macOS 13?? What APIs are you using?

fregante:
> I think this is reason enough to just target macOS 13 now.

That was the idea since Sonoma will be released in like 3 weeks. Those changes can be made here and then it can be merged in a month

sindresorhus:
I don't see why it needs to stay open. I don't like long-running PRs, and I also don't plan to do a new Safari version until then anyways.

sindresorhus:
> Sizing a window requires macOS 13?? What APIs are you using?

It's complicated. The window used to always be fixed size if you set a fixed size on the root view, but they changed it in macOS 13 to not respect the root view, but the modifiers to change it back require macOS 13.

fregante:
> I don't like long-running PRs

I mean, I did mention that there was no rush and that we could wait until Sonoma, but you can merge anytime.

sindresorhus:
> I mean, I did mention that there was no rush and that we could wait until Sonoma, but you can merge anytime.

Hehe. I know, but today was the time I had time and motivation.

fregante (approved):
:ship-it:

fregante:
Includes https://github.com/refined-github/refined-github/issues/6000#issuecomment-1486143405

Safari 16.4+

PR #6879, opened 2023-09-04
Meta: Extract GQL queries

fregante:
- Part of https://github.com/refined-github/refined-github/issues/6686

This extracts the remainder of the extractable GQL queries.

~~Untested, not cleaned~~


source/features/repo-age-first-commit.gql
```diff
@@ -0,0 +1,16 @@
+query GetFirstCommit($owner: String!, $name: String!) {
```
fregante, line 1: I'm trying to keep the gql files next to their features so that they're clearly linked. This decision isn't final. 

The issue is that we're importing .gql files as raw text at the moment so we can't have multiple queries in one yet.

This file is the result of both points.

source/features/tags-on-commits-list.gql
```diff
@@ -0,0 +1,39 @@
+query GetTagsOnCommit(
+	$owner: String!
+	$name: String!
+	$commit: String!
+	$after: String
```
fregante, line 5: Needs to be tested, may be undefined

source/features/tags-on-commits-list.tsx
```diff
@@ -49,43 +50,12 @@ function isTagTarget(target: CommonTarget): target is TagTarget {
...
-			object(expression: "${lastCommit}") {
-				... on Commit {
-					committedDate
-				}
-			}
-		}
-		`);
+	const {repository} = await api.v4(GetTagsOnCommit, {
+		variables: {
+			after: after ?? null,
```
fregante, line 55: How is this passing lint? 😩 Unreal


Anyway, it's the result of accepting a Json type

source/features/link-to-changelog-file.tsx
```diff
@@ -9,6 +9,7 @@ import features from '../feature-manager.js';
 import api from '../github-helpers/api.js';
 import {wrapAll} from '../helpers/dom-utils.js';
 import {buildRepoURL, getRepo} from '../github-helpers/index.js';
+import GetFilesOnRoot from './show-associated-branch-prs-on-fork.gql';
```
FloEdelmann, line 12: Shouldn't this be

```suggestion
import GetFilesOnRoot from './link-to-changelog-file.gql';
```

?

source/features/repo-age-first-commit.gql
FloEdelmann: Having one file per feature sounds like the most maintainable solution for me. (And it could be checked automatically in the feature validator, which could prevent mistakes like the one mentioned in my previous comment.)

But having multiple queries and importing them as `import {Query1, Query2} from './feature.gql'` would of course be helpful.

source/features/tags-on-commits-list.tsx
FloEdelmann: Why shouldn't it be passing lint? :thinking:

fregante: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-null.md

source/features/tags-on-commits-list.gql
fregante: Great! It works with `null`, but I can't find a situation where `after` is defined 🤷‍♂️ 


Screenshot 3
Screenshot 2

source/features/link-to-changelog-file.tsx
FloEdelmann: @fregante This was not fixed, right?

fregante: Oops, no

FloEdelmann: So now `link-to-changelog-file` is probably broken in the latest release?

source/features/tags-on-commits-list.tsx
fisker: xo disabled `no-null`. https://github.com/xojs/xo/blob/main/config/plugins.cjs#L163

fregante: Oh I thought I had it enabled here. I've been switching between projects where this was enforced

PR #6880, opened 2023-09-04
Ask for review in the Safari extension app

sindresorhus:
The review is requested on the second launch, and to not be annoying, it's only requested once ever. We kinda have to do it on the second launch as most don't even launch it a second time, even less a third time.

### DIFF
--- a/safari/Refined GitHub/MainScreen.swift
+++ b/safari/Refined GitHub/MainScreen.swift
@@ -1,7 +1,10 @@
 import SwiftUI
 import SafariServices
+import StoreKit
 
 struct MainScreen: View {
+	@Environment(\.requestReview) private var requestReview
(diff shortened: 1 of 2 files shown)


fregante:
If it follows this setting I have no problem with it https://9to5mac.com/2021/12/29/how-to-turn-off-in-app-review-requests-on-iphone-ipad-and-mac/

Can you also attach a screenshot for future reference?

sindresorhus:
> If it follows this setting I have no problem with it https://9to5mac.com/2021/12/29/how-to-turn-off-in-app-review-requests-on-iphone-ipad-and-mac/

It does.

> Can you also attach a screenshot for future reference?

Screenshot 2023-09-04 at 21 10 59

PR #6886, opened 2023-09-07
Drop ability to `quick-file-edit` on permalinks

fregante:
- Follows https://github.com/refined-github/refined-github/pull/6875


akhilappu40131-tech (commented):
Hh
Hot webseries

PR #6893, opened 2023-09-09
Match the behavior of function `isUrlReachable` to its name

134130:
## Description
- `isUrlReachable` function's behavior is not matches it's name.
- Flip it's behavior and usages
- https://github.com/refined-github/refined-github/pull/6840#discussion_r1320079324

cc. @jerone

### DIFF
--- a/source/github-helpers/index.ts
+++ b/source/github-helpers/index.ts
@@ -140,7 +140,7 @@ export const isRepoCommitListRoot = (): boolean => pageDetect.isRepoCommitList()
 
 export const isUrlReachable = mem(async (url: string): Promise<boolean> => {
 	const {status} = await fetch(url, {method: 'head'});
-	return status === 404;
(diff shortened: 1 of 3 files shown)


source/github-helpers/index.ts
```diff
@@ -140,7 +140,7 @@ export const isRepoCommitListRoot = (): boolean => pageDetect.isRepoCommitList()
 
 export const isUrlReachable = mem(async (url: string): Promise<boolean> => {
 	const {status} = await fetch(url, {method: 'head'});
-	return status === 404;
+	return status === 200;
```
fregante, line 143: Use `response.ok`

fregante (approved):
As long as every usage has been updated, it looks good

134130:
All usages are checked with `WebStorm`

PR #6895, opened 2023-09-09
Add support for New Repository Overview in `unreleased-commits`

### LINKED ISSUE #6851: `unreleased-commits` missing from New Repository Overview
### Description

Old:

Screenshot 10

New:

Screenshot 11

### Extension version

23.8.3

### Browser(s) used

Chrome

134130:
## Describe
- Closes #6851 

## Test URLs
- Repo with no tags (no button)
https://github.com/refined-github/yolo

- Repo with too many unreleased commits
https://github.com/refined-github/sandbox

- Repo with some unreleased commits (If no button, check if is latest tag)

## Screenshot
image

### DIFF
--- a/source/features/unreleased-commits.tsx
+++ b/source/features/unreleased-commits.tsx
@@ -12,6 +12,7 @@ import pluralize from '../helpers/pluralize.js';
 import {branchSelector} from '../github-helpers/selectors.js';
 import getPublishRepoState from './unreleased-commits.gql';
 import getDefaultBranch from '../github-helpers/get-default-branch.js';
+import {wrapAll} from '../helpers/dom-utils.js';
 import abbreviateString from '../helpers/abbreviate-string.js';
 
 type RepoPublishState = {
@@ -77,19 +78,19 @@ async function add(branchSelector: HTMLButtonElement): Promise<void> {
 			: pluralize(aheadBy, '$$ unreleased commit');
 	const label = `There are ${commitCount} since ${abbreviateString(latestTag, 30)}`;
 
-	const container = <div className="d-flex gap-2"/>;
-	branchSelector.before(container);
-
-	container.append(
-		branchSelector,
-		<a
-			className="btn px-2 tooltipped tooltipped-se"
-			href={buildRepoURL('compare', `${latestTag}...${await getDefaultBranch()}`)}
-			aria-label={label}
-		>
-			<TagIcon className="v-align-middle"/>
-			{aheadBy === undeterminableAheadBy || <sup className="ml-n2"> +{aheadBy}</sup>}
-		</a>,
+	wrapAll(
+		[
+			branchSelector,
+			<a
+				className="btn px-2 tooltipped tooltipped-se"
+				href={buildRepoURL('compare', `${latestTag}...${await getDefaultBranch()}`)}
+				aria-label={label}
+			>
+				<TagIcon className="v-align-middle"/>
+				{aheadBy === undeterminableAheadBy || <sup className="ml-n2"> +{aheadBy}</sup>}
+			</a>,
+		],
+		<div className="d-flex gap-2"/>,
 	);
 }
 
--- a/source/features/unreleased-commits.tsx
+++ b/source/features/unreleased-commits.tsx
@@ -1,18 +1,18 @@
 import React from 'dom-chef';
-import { CachedFunction } from 'webext-storage-cache';
+import {CachedFunction} from 'webext-storage-cache';
 import * as pageDetect from 'github-url-detection';
-import { TagIcon } from '@primer/octicons-react';
+import {TagIcon} from '@primer/octicons-react';
 
 import features from '../feature-manager.js';
 import observe from '../helpers/selector-observer.js';
 import api from '../github-helpers/api.js';
-import { addAfterBranchSelector, buildRepoURL, cacheByRepo, getLatestVersionTag } from '../github-helpers/index.js';
+import {addAfterBranchSelector, buildRepoURL, cacheByRepo, getLatestVersionTag} from '../github-helpers/index.js';
 import isDefaultBranch from '../github-helpers/is-default-branch.js';
 import pluralize from '../helpers/pluralize.js';
-import { branchSelector, branchSelectorParent } from '../github-helpers/selectors.js';
+import {branchSelector, branchSelectorParent} from '../github-helpers/selectors.js';
 import getPublishRepoState from './unreleased-commits.gql';
 import getDefaultBranch from '../github-helpers/get-default-branch.js';
-import { wrapAll } from '../helpers/dom-utils.js';
+import {wrapAll} from '../helpers/dom-utils.js';
 import abbreviateString from '../helpers/abbreviate-string.js';
 
 type RepoPublishState = {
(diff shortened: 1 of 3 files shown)


source/features/unreleased-commits.tsx
```diff
@@ -6,12 +6,13 @@ import {TagIcon} from '@primer/octicons-react';
 import features from '../feature-manager.js';
 import observe from '../helpers/selector-observer.js';
 import api from '../github-helpers/api.js';
-import {addAfterBranchSelector, buildRepoURL, cacheByRepo, getLatestVersionTag} from '../github-helpers/index.js';
+import {buildRepoURL, cacheByRepo, getLatestVersionTag} from '../github-helpers/index.js';
 import isDefaultBranch from '../github-helpers/is-default-branch.js';
-import getDefaultBranch from '../github-helpers/get-default-branch.js';
 import pluralize from '../helpers/pluralize.js';
-import {branchSelectorParent} from '../github-helpers/selectors.js';
```
134130, line 13: `addAfterBranchSelector` and `branchSelectorParent` is now only used on a  one feature.
Need to consider dropping these from helpers later.

source/features/unreleased-commits.tsx
```diff
@@ -74,12 +75,15 @@ async function add(branchSelectorParent: HTMLDetailsElement): Promise<void> {
 		= aheadBy === undeterminableAheadBy
 			? 'more than 20 unreleased commits'
 			: pluralize(aheadBy, '$$ unreleased commit');
-	const label = `There are ${commitCount} since ${latestTag}`;
+	const label = `There are ${commitCount} since ${abbreviateString(latestTag, 30)}`;
```
fregante, line 78: Instead of manually abbreviating we can use GitHub's own "truncate" class. They used it every time a branch name is used too

source/features/unreleased-commits.tsx
```diff
@@ -74,12 +75,15 @@ async function add(branchSelectorParent: HTMLDetailsElement): Promise<void> {
...
+
+	const container = <div className="d-flex gap-2"/>;
+	branchSelector.before(container);
```
fregante, line 81: You're probably looking for the `wrapAll` utility

source/features/unreleased-commits.tsx
134130: AFAIK, CSS Truncate `text-overflow: ellipsis` only can be applied to `width fixed elements`
GitHub is using `Truncate` `Truncate-text` or `css-truncate` classes on their `inline elements` not tooltips.

fregante: This is the `branch` link in PR headers, you can definitely use it here too. You just need two elements and the class already has a max-width

Screenshot

Screenshot 4

source/features/unreleased-commits.tsx
```diff
@@ -96,7 +101,7 @@ async function init(signal: AbortSignal): Promise<false | void> {
 
 	await api.expectToken();
 
-	observe(branchSelectorParent, add, {signal});
+	observe(branchSelector, add, {signal});
```
fregante, line 104: We generally try to keep compatibility with both versions wherever possible, or else we get issues like:

- #6899

source/features/unreleased-commits.tsx
134130: 🤔 I dont know is this possible.
The tooltip is `pseudo-element` and it effected by parent element's overflow attribute.
When we use `overflow: hidden` with `css-truncate > css-truncate-target` will not appears tooltip.
And the `text-overflow: ellipsis` needs `overflow: hidden`
화면 기록 2023-09-10 20 22 28

source/features/unreleased-commits.tsx
134130: I'll check both of versions 😉

source/features/unreleased-commits.tsx
fregante: Ahhh you're absolutely right.

source/helpers/abbreviate-string.ts
```diff
@@ -0,0 +1,3 @@
+export default function abbreviateString(string_: string, length: number): string {
+	return string_.slice(0, length) + '…';
```
fregante, line 2: This appends `…` whether it's needed or not. There should be a condition

source/features/unreleased-commits.tsx
```diff
@@ -97,6 +122,7 @@ async function init(signal: AbortSignal): Promise<false | void> {
 	await api.expectToken();
 
 	observe(branchSelectorParent, add, {signal});
+	observe(branchSelector, addForNewRepositoryOverview, {signal});
```
fregante, line 125: When adding code that supports multiple versions:

- add a TODO comment with a date 6-8 months from now to remind us to delete the old code
- make it easy to delete the old code without having to rewrite much else
- don't use "new" in the names; the "new" code is actually just "current" in the future

source/features/unreleased-commits.tsx
```diff
@@ -62,30 +64,53 @@ export const repoPublishState = new CachedFunction('tag-ahead-by', {
...
 
-async function add(branchSelectorParent: HTMLDetailsElement): Promise<void> {
-	const {latestTag, aheadBy} = await repoPublishState.get();
-	const isAhead = aheadBy > 0;
-
-	if (!latestTag || !isAhead) {
-		return;
-	}
-
+async function createUnreleasedCommitsTag(latestTag: string, aheadBy: number): Promise<HTMLElement> {
```
fregante, line 67:
```suggestion
async function createLink(latestTag: string, aheadBy: number): Promise<HTMLElement> {
```

source/features/unreleased-commits.tsx
```diff
@@ -62,30 +64,53 @@ export const repoPublishState = new CachedFunction('tag-ahead-by', {
...
 			: pluralize(aheadBy, '$$ unreleased commit');
-	const label = `There are ${commitCount} since ${latestTag}`;
+	const label = `There are ${commitCount} since ${abbreviateString(latestTag, 30)}`;
 
-	addAfterBranchSelector(
-		branchSelectorParent,
+	return (
 		<a
-			className="btn px-2 tooltipped tooltipped-ne"
+			className="btn px-2 tooltipped tooltipped-se"
```
fregante, line 76: Can you also fix `default-branch-button`?

Screenshot 3


Example URL:

https://github.com/refined-github/sandbox/tree/4679-1

134130: 🤔  I think it is already fixed on https://github.com/refined-github/refined-github/commit/aa9d746180b8d5a96307ae311d5a7adeee73feaa#diff-76f0cf976001e57ff88f7efe4ab593cbcf69929b5ae3e5cc5f085cc85de175e6L33

fregante: Ah yeah, Safari has an old version

source/helpers/abbreviate-string.ts
```diff
@@ -0,0 +1,3 @@
+export default function abbreviateString(string_: string, length: number, abbrevMarker = '…'): string {
+	return string_.slice(0, length) + abbrevMarker;
```
fregante, line 2: This is not a condition, the result is unchanged. I meant:

```suggestion
	return string.length < length ? string : string_.slice(0, length) + '…';
```

134130: Sorry my bad 🥲

fregante:
Screenshot 22

fregante (approved):

PR #6897, opened 2023-09-10
Add `mark-private-repos` feature

### LINKED ISSUE #6656: Make the public/private repos more distinguishable
### Description

In one's own repository tab, it is hard to decide at a glance which repos are private and which ones are public. These two labels look very similar. Example:

image

In contrast, in repo pages, while still using the same labels, private repos have this little lock icon to clearly indicate that they are private. Example:

image
image

I wonder if some improvements can be done to the own repository tab.

Some suggestions on the top of my mind: add a lock icon somewhere around the label; make the Private label a different colour (maybe orange); make the Public label a different colour (maybe green).

Thank you very much!

### Example URLs

`https://github.com/<oneself>?tab=repositories&type=source`

thehale:
The official styling in each user's list of personal repositories makes it difficult to distinguish public from private repos at a glance.

This commit colors the "Private" label with GitHub's "attention" color so that it stands out more prominently.

closes #6656

## Test URLs
`https://github.com/YOUR_USERNAME?tab=repositories`

### DIFF
--- a/source/features/mark-private-repos.css
+++ b/source/features/mark-private-repos.css
@@ -1,0 +1,4 @@
+li.private .Label--secondary {
+	border-color: var(--color-attention-muted);
+	color: var(--color-attention-emphasis);
+}
--- a/source/refined-github.ts
+++ b/source/refined-github.ts
@@ -20,6 +20,7 @@ import './features/sticky-file-header.css';
 import './features/readable-title-change-events.css';
 import './features/clean-checks-list.css';
 import './features/sticky-csv-header.css';
+import './features/mark-private-repos.css';
 
 // DO NOT add CSS files here if they are part of a JavaScript feature.
 // Import the `.css` file from the `.tsx` instead.


source/refined-github.ts
```diff
@@ -20,6 +20,7 @@ import './features/sticky-file-header.css';
 import './features/readable-title-change-events.css';
 import './features/clean-checks-list.css';
 import './features/sticky-csv-header.css';
+import './features/mark-private-repos.css';
```
thehale, line 23: While most sections of features are sorted, I noticed that the list of CSS-only features was not sorted. So I just put mine at the end.

I'm happy to push up a commit sorting the list of CSS-only features if desired.

fregante: They're not sorted, we just append features to the list, so this was the right choice.

source/features/mark-private-repos.css
```diff
@@ -0,0 +1,4 @@
+li.private .Label--secondary {
+	border-color: var(--color-attention-muted);
+	color: var(--color-attention-emphasis);
```
fregante, line 3: Looks good to me, but I'd want you to make sure that these are the colors GitHub uses in this case, I know that there's a specific "border" color type as well.

fregante (approved):

source/features/mark-private-repos.css
134130: 😉 
image

134130: And the `Attention Label` is used for `Public archive` repositories.
Is Private repository label colors can be same with these?

134130: FYI, other colors 
image

fregante:
Thank you for the PR, Joseph!

Issue #6473, opened 2023-04-08
Safari 16.4 transparent text bug in `pr-approvals-count`


fregante:
### Expected

<img width="591" alt="Screenshot 6" src="https://user-images.githubusercontent.com/1402241/230712520-0df1c13b-7d27-430d-a355-308e9fb6587c.png">

### Current

<img width="591" alt="Screenshot 4" src="https://user-images.githubusercontent.com/1402241/230712527-ba636a5e-8d7f-43f1-a8fa-34f3ed379870.png">

### How to replicate the issue + URL

https://github.com/refined-github/refined-github/pulls?q=+review%3Aapproved+

### Extension version

23.3.2

### Browser(s) used

Safari 16.4

fregante:
Hotfixed for now. If it survives one or more RGH Safari releases I'll merge the hotfix into `main`

teor2345:
This is still happening in MobileSafari on iPadOS 16.4.1 (a) (20E772520a).

Manually applying the hotfix from https://github.com/refined-github/yolo/blob/main/style/23.4.15.css fixes it.

fregante:
Thank you for the report, I forgot about it. I’ll re-apply the hotfix now

PR #6898, opened 2023-09-10
Add support for "minimum font size" browser setting in `pr-approvals-count`

134130:
## Description
- Restoring workaround for Safari on other browsers.
- Related #6473

## Test URLs
- https://github.com/pulls

- _**Cannot install `refined-github` on my MacOS. Can somebody help for testing this PR?**_

## Screenshot
- Before
  image

- After
  image

### DIFF
--- a/source/features/pr-approvals-count.css
+++ b/source/features/pr-approvals-count.css
@@ -2,9 +2,7 @@
 .js-issue-row
 	.color-fg-muted
 	[href$='#partial-pull-merging'][aria-label*='approval'] {
-	/* font-size: 0; */
-	/* Safari 16.4 color bug #6473 */
-	font-size: 0.6px;
+	font-size: 0;
 	color: transparent !important;
 }
 
@@ -13,7 +11,7 @@
 	[href$='#partial-pull-merging'][aria-label*='approval']::before {
 	all: unset;
 	content: attr(aria-label);
-	display: inline !important;
+	display: inline-block !important;
 	color: var(--rgh-green);
 	font-size: 12px;
 }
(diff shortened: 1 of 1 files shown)


fregante:
Restoring what exactly? It works for me in both Safari and Chrome. Where are you experiencing that bug?

Screenshot 2
Screenshot 1

134130:
Can you check is there some overrides on your browser?
And I am using `chrome: 115.0.5790.170` `MacOS: Ventura`
image

fregante:
Version 116.0.5845.179, Ventura 13.4.1

Screenshot

source/features/pr-approvals-count.css
```diff
@@ -2,10 +2,21 @@
...
+}
+
+.js-issue-row
+	.color-fg-muted
+	[href$='#partial-pull-merging'][aria-label*='approval']
+	@media
+	not
+	all
+	and
+	(min-resolution: 0.001dpcm) {
```
fregante, line 16: This isn't a valid selector, you can't mix media queries and selectors

fregante:
Is it using this font for you?

Screenshot 1

134130:
Yes, rendered with same font on my local.

source/features/pr-approvals-count.css
134130: I'll fix it after finding the reason of this bug

134130:
Sorry for Korean, but I think it is releated with Chrome's settings
chrome://settings/fonts > Min font size is was 10 in default in my case

fregante (approved):
I found an alternative fix for Safari. The text was invisible until I set `inline-block`

134130:
A lot of Thanks

PR #6901, opened 2023-09-10
Restore `releases-tab` when "Global Navigation Bar" is disabled

### LINKED ISSUE #6899: `releases-tab`: not visible when GitHub's Feature preview - "Global Navigation Update" is disabled
### Description

`release-tab` is not visible when GitHub's Feature preview - "Global Navigation Update" is disabled

image

turns out there are multiple `.UnderlineNav-body` on the page, and the last one is the desired target:

image

- Disable "Global Navigation Update" in "Feature preview"
- Open https://github.com/refined-github/refined-github

### Extension version

23.9.7

### Browser(s) used

Chromium 117.0.5938.48

134130:
## Description
- Closes #6899
- `releases-tab` is not working when GitHub's Feature preview: `Global Navigation Update` is disabled.
  - Since the `.UnderlineNav-body` count is 3 when disabled. More specifying with `ul.UnderlineNav-body` selector
  - It works both of disabled and enabled

## Test URLs

## Screenshot
- Disabled
  image

- Enabled
  image

### DIFF
--- a/source/features/releases-tab.tsx
+++ b/source/features/releases-tab.tsx
@@ -14,7 +14,7 @@ import createDropdownItem from '../github-helpers/create-dropdown-item.js';
 import {buildRepoURL, cacheByRepo, getRepo} from '../github-helpers/index.js';
 import {releasesSidebarSelector} from './clean-repo-sidebar.js';
 import {appendBefore, highlightTab, unhighlightTab} from '../helpers/dom-utils.js';
-import {underlineNavDropdownUl} from '../github-helpers/selectors.js';
+import {repoUnderlineNavUrl, repoUnderlineNavDropdownUl} from '../github-helpers/selectors.js';
 import GetReleasesCount from './releases-tab.gql';
 
 async function parseCountFromDom(): Promise<number> {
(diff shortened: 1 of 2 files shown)


source/features/releases-tab.tsx
```diff
@@ -58,7 +58,7 @@ async function addReleasesTab(): Promise<false | void> {
 	}
 
 	// Wait for the tab bar to be loaded
-	const repoNavigationBar = (await elementReady('.UnderlineNav-body'))!;
+	const repoNavigationBar = (await elementReady('ul.UnderlineNav-body'))!;
```
fregante, line 61: We should reuse `underlineNavDropdownUl` if possible

fregante: But even then, why is this feature querying the same element twice (line 85)? It needs to be updated.

134130: Yeah I think this feature is too old and needs some enhancement too.
I'll work this PR again and reopen this PR.

fregante: Hmm I'm not seeing any glaring issues with this feature, I would not rewrite it. If your latest changes work well already, let's just drop line 85

fregante: Actually this was wrong, they're two different elements. I send a new PR for further adjustments

PR #6900, opened 2023-09-10
Add `repo-header-info` feature

134130:
## Description
- Closes #6798
- On original conversation suggested, which is using default `details` component, is too heavy and ugly.
- Just showing count of fork and stargazer count

## Test URLs

- Regular: https://github.com/refined-github/refined-github
- Fork: https://github.com/134130/refined-github
- Fork with native icon: https://github.com/refined-github/fork
- Private: https://github.com/refined-github/private
- Private forks are not allowed on GitHub

## Screenshot
[outdated](https://github.com/refined-github/refined-github/assets/50487467/2e9450e6-e1f2-41d0-a8ed-d7e66e66d375)

PR #6908, opened 2023-09-12
Restore `clean-repo-header` feature

fregante:
Reverts refined-github/refined-github#6759

The header is still long, ugly and it still causes a wrap and [Cumulative Layout Shift](https://web.dev/cls/).

We can restore it without making it disableable because _all_ of that information is visible in the sidebar, on the same page, as well as in the upcoming:

- #6900 

Screenshot 16


fregante:
## Before

Screenshot 20


## After

Screenshot 21

fregante:
I limited this feature to the new header

## Affected



Screenshot 24


## Unaffected (except "pins" label)

Screenshot 23

innocenzi:
This is disruptive when disabling the new `repo-header-info` feature, we don't get star count anywhere. :/

Should the CSS in this PR not only be added when `repo-header-info` is enabled?

fregante:
>we don't get star count anywhere. :/

See the first screenshot, the stars are there

fregante:
- https://github.com/refined-github/refined-github/pull/6941

Update to 23.9.21

PR #6910, opened 2023-09-13
Disable `dim-bots` only when clicking on empty areas

fregante:
This was annoying to do until now: ctrl-click multiple PRs.

This PR makes 2 changes:

- the bots won't undim when clicking on active elements 
- the bots won't undim when pressing _any_ keys

https://github.com/refined-github/refined-github/assets/1402241/5841672a-767f-4226-895a-095456313d81

## Test URL

https://github.com/mozilla/web-ext/pulls

### DIFF
--- a/source/features/dim-bots.tsx
+++ b/source/features/dim-bots.tsx
@@ -1,7 +1,7 @@
 import './dim-bots.css';
 import select from 'select-dom';
 import * as pageDetect from 'github-url-detection';
-import delegate from 'delegate-it';
+import delegate, {DelegateEvent} from 'delegate-it';
 
 import features from '../feature-manager.js';
 import preserveScroll from '../helpers/preserve-scroll.js';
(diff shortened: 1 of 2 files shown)


source/features/dim-bots.tsx
```diff
@@ -58,9 +63,6 @@ function init(signal: AbortSignal): void {
 
 	// Undim on mouse focus
 	delegate(dimBots.selector, 'click', undimBots, {signal});
-
-	// Undim on keyboard focus
-	document.documentElement.addEventListener('navigation:keydown', undimBots, {once: true, signal});
```
fregante, line 63: Because GitHub crippled this event, I just have to drop it. There's no way to distinguish between an actual navigation and the mere pressing of the `ctrl` key. So, nothing, I'm not going to rewrite their logic to detect actual j/k navigation.

source/features/hide-navigation-hover-highlight.tsx
```diff
@@ -7,7 +7,7 @@ const html = document.documentElement;
 
 function init(): void {
 	html.setAttribute(attribute, '');
-	html.addEventListener('navigation:focus', () => {
+	html.addEventListener('navigation:keydown', () => {
```
fregante, line 10: Bonus bugfix: likewise, just hovering an item focuses the item, so I had to change it to at least `keydown` (not perfect, but not as noticeable as in `dim-bots`)

PR #6916, opened 2023-09-16
`show-names` - Exclude invalid elements on new dashboard

### LINKED ISSUE #6915: `show-names` broken on GitHub dashboard's follwed article.
### Description

`show-names` on dashboard page's followed article is broken.

`김정현` is already user's name and user's key is next of it's name with `kimjh11130`.
image
image

Go dashbord (https://github.com/dashboard) and find `followed article`.

### Extension version

23.9.7

### Browser(s) used

Chrome

134130:
## Description
- Closes #6915 
- Fix `show-names` fetches redundant username on `followed article` in `GitHub newsfeed`

## Test URLs
- https://github.com/ (some followed article)

## Screenshot
image

### DIFF
--- a/source/features/show-names.tsx
+++ b/source/features/show-names.tsx
@@ -16,7 +16,8 @@ const batchUpdateLinks = batchedFunction(async (batchedUsernameElements: HTMLAnc
 	const myUsername = getUsername();
 	for (const element of new Set(batchedUsernameElements)) {
 		const username = element.textContent;
-		if (username && username !== myUsername && username !== 'ghost') {
+
+		if (username && username !== myUsername && username !== 'ghost' && element.href.endsWith(`/${username}`)) {
 			usernames.add(element.textContent!);
 		}
 
@@ -82,7 +83,7 @@ const usernameLinksSelector = [
 	// On dashboard
 	// `.Link--primary` excludes avatars
 	// `.color-shadow-medium` excludes links in cards #6530
-	'#dashboard article header a.Link--primary[data-hovercard-type="user"]:not(.color-shadow-medium a)',
+	'#dashboard a.Link--primary[data-hovercard-type="user"]:not(.color-shadow-medium a)',
 ] as const;
 
 function init(signal: AbortSignal): void {
--- a/source/features/show-names.tsx
+++ b/source/features/show-names.tsx
@@ -82,7 +82,7 @@ const usernameLinksSelector = [
 	// On dashboard
 	// `.Link--primary` excludes avatars
 	// `.color-shadow-medium` excludes links in cards #6530
-	'#dashboard a.Link--primary[data-hovercard-type="user"]:not(.color-shadow-medium a)',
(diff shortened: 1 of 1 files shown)


134130:
There is two options for fixing this issue. 
- Only enable `show-names` on article's header user [`5382431` (#6916)](https://github.com/refined-github/refined-github/pull/6916/commits/53824312f5335f834bcd24326f40f7735127b3b3)
- Filter `show-names` with it's link validation [`70356a7` (#6916)](https://github.com/refined-github/refined-github/pull/6916/commits/70356a77a9046c8d92508d74702267d8d27a2c55)

I think 2nd one looks better

134130:
Both are verified

source/features/show-names.tsx
```diff
@@ -16,7 +16,8 @@ const batchUpdateLinks = batchedFunction(async (batchedUsernameElements: HTMLAnc
 	const myUsername = getUsername();
 	for (const element of new Set(batchedUsernameElements)) {
 		const username = element.textContent;
-		if (username && username !== myUsername && username !== 'ghost') {
+
+		if (username && username !== myUsername && username !== 'ghost' && element.href.endsWith(`/${username}`)) {
```
fregante, line 20: When does this not end with /username?

134130: You can see on my description.

In my case the element's textContent is '김정현' and href is 'ki.jh11012'.
'김정현' is user's nickname not github username. So in this feature, no need to query its nickname since it is already showing.

fregante: If the selector is selecting the wrong thing, maybe we should alter it. Is there any simple change that would avoid this selection? For example by appending `:not(.followed *)` if that class exists on a parent

134130: I'll digging is there some way to specifying these with css selector.

134130: I think there is no way to detect `followed` article just with css selector.

fregante: `:not([aria-label="card content"] *)`

fregante: This href-based condition fails when people set the same name, for example (no RGH enabled)

Screenshot

134130: - It have been tried to excluding card content before on this issue.
  - https://github.com/refined-github/refined-github/issues/6530
- Since `:not([aria-label="card content"] *)` can avoid above case also, just using it and discarded old selector: `:not(.color-shadow-medium a)`

fregante: Thanks for the change! I think we can drop this exclusion though, I'd rather tweak the selector over time if necessary

134130: Oops, sorry. I'll drop it

fregante (approved):

fregante:
I feel like we should start naming PRs with the name of the feature first, it will make for a more readable release log as well as commit list

yakov116:
Agree 👍

PR #6931, opened 2023-09-19
Meta: Rename `GitHubURL` to `GitHubFileURL`

fregante:
Since Firefox still fails, we can only rename the class.

- Replaces https://github.com/refined-github/refined-github/pull/6870

### DIFF
--- a/source/features/conflict-marker.tsx
+++ b/source/features/conflict-marker.tsx
@@ -7,6 +7,7 @@ import batchedFunction from 'batched-function';
 import features from '../feature-manager.js';
 import api from '../github-helpers/api.js';
 import observe from '../helpers/selector-observer.js';
+import {isHasSelectorSupported} from '../helpers/select-has.js';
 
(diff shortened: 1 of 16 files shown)


source/features/conflict-marker.tsx
```diff
@@ -48,6 +49,9 @@ function init(signal: AbortSignal): void {
 }
 
 void features.add(import.meta.url, {
+	asLongAs: [
+		isHasSelectorSupported,
+	],
```
fregante, line 54: Also exclude this feature from Firefox (1/2)

source/features/rgh-dim-commits.tsx
```diff
@@ -20,6 +21,7 @@ function init(signal: AbortSignal): void {
 void features.add(import.meta.url, {
 	asLongAs: [
 		isRefinedGitHubRepo,
+		isHasSelectorSupported,
```
fregante, line 24: Also exclude this feature from Firefox (2/2)

PR #6932, opened 2023-09-19
`repo-header-info` - Add link and tooltip

fregante:
- Close https://github.com/refined-github/refined-github/issues/6926
- Address https://github.com/refined-github/refined-github/pull/6900#issuecomment-1724702659

### Normal

Screenshot 6

### Hover

Screenshot 5


Teraskull:
Can you please also format the number with a thousands separator like in the original title: `21,710`? Maybe something like [`Intl.NumberFormat()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat) would fit.

fregante:
Screenshot 3

---

Screenshot 2

PR #6956, opened 2023-09-24
Support fine-grained tokens in the options and update the widget

### LINKED ISSUE #6952: Detect fine-grained tokens in the options and update the widget
This happens when you paste a fine-grained token:

Screenshot

The token works but the UI isn't updated.

We cannot show all the scopes of the token, but we should at least update the UI to say something like:

- Fine-grained token validated, but some features and organizations might not support them [More info](https://github.com/refined-github/refined-github/issues/6092#issuecomment-1732286938)

Hiding everything else.

Extracted from: 

- https://github.com/refined-github/refined-github/issues/6092

134130:
## Description
- Closes #6952

## Test URLs
- Your extension's options page

## Screenshot

https://github.com/refined-github/refined-github/assets/50487467/df98d2d8-c6b0-4e7b-8195-870f72384519

### DIFF
--- a/source/options.tsx
+++ b/source/options.tsx
@@ -25,15 +25,15 @@ import {state as bisectState} from './helpers/bisect.js';
 type TokenType = 'classic' | 'fine_grained';
 
 type Status = {
+	tokenType: TokenType;
 	error?: true;
 	text?: string;
-	tokenType?: TokenType;
 	scopes?: string[];
 };
 
 const {version} = browser.runtime.getManifest();
 
-function reportStatus({error, text, tokenType, scopes}: Status): void {
+function reportStatus({tokenType, error, text , scopes}: Status): void {
 	const tokenStatus = select('#validation')!;
 	tokenStatus.textContent = text ?? '';
 	if (error) {
@@ -43,7 +43,6 @@ function reportStatus({error, text, tokenType, scopes}: Status): void {
 	}
 
 	// Toggle the ulists by token type (default to classic)
-	tokenType = tokenType ?? 'classic';
 	for (const ulist of select.all('[data-token-type]')) {
 		ulist.style.display = ulist.dataset.tokenType === tokenType ? '' : 'none';
 	}
@@ -129,7 +128,7 @@ async function validateToken(): Promise<void> {
 		});
 	} catch (error) {
 		assertError(error);
-		reportStatus({error: true, text: error.message, tokenType});
+		reportStatus({tokenType, error: true, text: error.message});
 		expandTokenSection();
 		throw error;
 	}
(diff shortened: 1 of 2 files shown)


source/options.tsx
```diff
@@ -39,6 +42,12 @@ function reportStatus({error, text, scopes}: Status): void {
 		delete tokenStatus.dataset.validation;
 	}
 
+	// Toggle the ulists by token type (default to classic)
+	tokenType = tokenType ?? 'classic';
+	for (const ulist of select.all('[data-token-type]')) {
+		ulist.style.display = ulist.dataset.tokenType === tokenType ? '' : 'none';
+	}
```
134130, line 49: Used like this since I want to make `tokenType`'s default is 'classic' on status reporting, but below looks more better
```suggestion
	for (const ulist of select.all('[data-token-type]')) {
		ulist.style.display = (ulist.dataset.tokenType === tokenType ?? 'classic') ? '' : 'none';
	}
```

source/options.tsx
```diff
@@ -68,7 +77,8 @@ async function getTokenScopes(personalToken: string): Promise<string[]> {
 		throw new Error(details.message);
 	}
 
-	const scopes = response.headers.get('X-OAuth-Scopes')!.split(', ');
+	// If `X-OAuth-Scopes` is not present, the token is a fine-grained token.
+	const scopes = response.headers.get('X-OAuth-Scopes')?.split(', ') ?? [];
```
134130, line 81:
```suggestion
	// If `X-OAuth-Scopes` is not present, the token may be not a classic token.
	const scopes = response.headers.get('X-OAuth-Scopes')?.split(', ') ?? [];
```

source/options.tsx
```diff
@@ -99,13 +109,9 @@ async function updateStorageUsage(area: 'sync' | 'local'): Promise<void> {
...
 async function validateToken(): Promise<void> {
-	reportStatus({});
 	const tokenField = select('input[name="personalToken"]')!;
-
-	if (tokenField.value.startsWith('github_pat_')) {
-		// Validation not supported yet https://github.com/refined-github/refined-github/issues/6092
-		return;
-	}
+	const tokenType = tokenField.value.startsWith('github_pat_') ? 'fine_grained' : 'classic';
+	reportStatus({tokenType});
```
134130, line 114: Just reporting after tokenType detecting.
If it reported duplicately, UI will being flickering.
Maybe we can drop undefinable attribute on `tokenType` of `reportStatus` param.

source/options.tsx
```diff
@@ -22,15 +22,18 @@ import isDevelopmentVersion from './helpers/is-development-version.js';
 import {doesBrowserActionOpenOptions} from './helpers/feature-utils.js';
 import {state as bisectState} from './helpers/bisect.js';
 
+type TokenType = 'classic' | 'fine_grained';
```
fregante, line 25: Snake case is only used by GitHub's own permissions, but the rest of the form only uses camelCase. Let's stick to that here

fregante (approved):
😍 fantastic

PR #6964, opened 2023-09-27
Restore `one-click-review-submission` feature

### LINKED ISSUE #6963: `one-click-review-submission` seems broken
### Description

Screenshot from 2023-09-27 14-14-57

`one-click-review-submission` is completly broken, as show in the screenshot.

https://github.com/refined-github/refined-github/pull/6954/files

### Extension version

23.9.21

### Browser(s) used

Firefox

RobHannay:
Fixes #6963.

GitHub have changed their review submission modal so that it:
- now has the radio buttons properly nested in a form
- the radios' labels are not adjacent siblings anymore

## Test URLs

[This PR??](https://github.com/refined-github/refined-github/pull/6964)

## Screenshot

Capture d’écran 2023-09-27 à 10 21 11 PM

https://github.com/refined-github/refined-github/assets/609062/be30dd75-a894-49e4-8023-9d7994165440

### Description tooltips

image
image
image

### DIFF
--- a/source/features/one-click-review-submission.tsx
+++ b/source/features/one-click-review-submission.tsx
@@ -9,7 +9,6 @@ import observe from '../helpers/selector-observer.js';
 function replaceCheckboxes(originalSubmitButton: HTMLButtonElement): void {
 	const form = originalSubmitButton.form!;
 	const actionsRow = originalSubmitButton.closest('.form-actions')!;
-	const formAttribute = originalSubmitButton.getAttribute('form')!;
 
 	// Do not use `select.all` because elements can be outside `form`
 	// `RadioNodeList` is dynamic, so we need to make a copy
@@ -43,18 +42,19 @@ function replaceCheckboxes(originalSubmitButton: HTMLButtonElement): void {
 			classes.push('tooltipped tooltipped-nw tooltipped-no-delay');
 		}
 
+		const label = document.querySelector(`label[for="${radio.id}"]`)?.textContent;
+
 		const button = (
 			<button
 				type="submit"
 				name="pull_request_review[event]"
 				// The buttons are no longer inside the form itself; this links the form
-				form={formAttribute}
 				value={radio.value}
 				className={classes.join(' ')}
 				aria-label={tooltip!}
 				disabled={radio.disabled}
 			>
-				{radio.nextSibling}
+				{label}
 			</button>
 		);
 
@@ -68,9 +68,7 @@ function replaceCheckboxes(originalSubmitButton: HTMLButtonElement): void {
 	}
 
 	// Remove original fields at last to avoid leaving a broken form
-	for (const radio of radios) {
-		radio.closest('.form-checkbox')!.remove();
-	}
+	radios[0].closest('fieldset')!.remove();
 
 	originalSubmitButton.remove();
 }
--- a/source/features/one-click-review-submission.tsx
+++ b/source/features/one-click-review-submission.tsx
@@ -5,10 +5,12 @@ import {CheckIcon, FileDiffIcon} from '@primer/octicons-react';
 
 import features from '../feature-manager.js';
 import observe from '../helpers/selector-observer.js';
+import {assertNodeContent} from '../helpers/dom-utils.js';
 
 function replaceCheckboxes(originalSubmitButton: HTMLButtonElement): void {
 	const form = originalSubmitButton.form!;
 	const actionsRow = originalSubmitButton.closest('.form-actions')!;
+	const formAttribute = originalSubmitButton.getAttribute('form')!;
 
 	// Do not use `select.all` because elements can be outside `form`
 	// `RadioNodeList` is dynamic, so we need to make a copy
@@ -29,24 +31,25 @@ function replaceCheckboxes(originalSubmitButton: HTMLButtonElement): void {
 		);
 	}
 
-	const formAttribute = originalSubmitButton.getAttribute('form')!;
-
 	// Generate the new buttons
 	for (const radio of radios) {
-		const labelNode = radio.labels?.[0];
-
-		const tooltip = labelNode?.nextElementSibling?.textContent;
+		const parent = radio.parentElement!;
+		const labelElement = (
+			parent.querySelector('label')
+			?? radio.nextSibling! // TODO: Remove after April 2024
+		);
+		const tooltip = parent.querySelector([
+			'p', // TODO: Remove after April 2024
+			'.FormControl-caption',
+		])!.textContent!.trim().replace(/.$/, '');
+		assertNodeContent(labelElement, /^(Approve|Request changes|Comment)$/);
 
 		const classes = ['btn btn-sm'];
-		if (radio.value === 'comment') {
-			classes.push('btn-primary');
-		}
 
 		if (tooltip) {
 			classes.push('tooltipped tooltipped-nw tooltipped-no-delay');
 		}
 
-		const oldLabel = radio.nextSibling?.textContent?.trim();
 		const button = (
 			<button
 				type="submit"
@@ -55,11 +58,10 @@ function replaceCheckboxes(originalSubmitButton: HTMLButtonElement): void {
 				form={formAttribute}
 				value={radio.value}
 				className={classes.join(' ')}
-				aria-label={tooltip!}
+				aria-label={tooltip}
 				disabled={radio.disabled}
 			>
-				{/* Backwards-compat label content queries. Issue #6963. */}
-				{oldLabel ? oldLabel : labelNode?.textContent}
+				{labelElement.textContent}
 			</button>
 		);
 
--- a/source/features/one-click-review-submission.tsx
+++ b/source/features/one-click-review-submission.tsx
@@ -35,7 +35,7 @@ function replaceCheckboxes(originalSubmitButton: HTMLButtonElement): void {
 	for (const radio of radios) {
 		const labelNode = radio.labels?.[0];
 
-		// new ?? old. Backwards-compat tooltip content queries. Issue #6963
+		// New ?? old. Backwards-compat tooltip content queries. Issue #6963
 		const tooltip = labelNode?.nextElementSibling?.textContent ?? radio.parentElement!.getAttribute('aria-label');
 
 		const classes = ['btn btn-sm'];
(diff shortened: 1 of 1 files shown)


source/features/one-click-review-submission.tsx
```diff
@@ -9,7 +9,6 @@ import observe from '../helpers/selector-observer.js';
 function replaceCheckboxes(originalSubmitButton: HTMLButtonElement): void {
 	const form = originalSubmitButton.form!;
 	const actionsRow = originalSubmitButton.closest('.form-actions')!;
-	const formAttribute = originalSubmitButton.getAttribute('form')!;
```
RobHannay, line 12: Because the submit is now nested, we needn't set the `form` attribute anymore

source/features/one-click-review-submission.tsx
```diff
@@ -43,18 +42,19 @@ function replaceCheckboxes(originalSubmitButton: HTMLButtonElement): void {
 			classes.push('tooltipped tooltipped-nw tooltipped-no-delay');
 		}
 
+		const label = document.querySelector(`label[for="${radio.id}"]`)?.textContent;
```
RobHannay, line 45: The labels are not adjacent siblings anymore, so I'm just grabbing the text from their actual `<label>`s.

Let me know if there's a way to do this that's more in keeping with the repo

source/features/one-click-review-submission.tsx
```diff
@@ -68,9 +68,7 @@ function replaceCheckboxes(originalSubmitButton: HTMLButtonElement): void {
 	}
 
 	// Remove original fields at last to avoid leaving a broken form
-	for (const radio of radios) {
-		radio.closest('.form-checkbox')!.remove();
-	}
+	radios[0].closest('fieldset')!.remove();
```
RobHannay, line 71: The radios are now wrapped in a `fieldset`, which also has padding of its own, so it's cleaner just to remove.

Accessing the first radio and therefore its closest `fieldset` felt like the most straightforward way to achieve without having to find anything more specific about the fieldset (it only has an `m-3` class)

capfei:
Not sure if it is considered better or not but I was able to get the button text by replacing `{radio.nextSibling}` with `{radio.labels[0].innerHTML}`.

RobHannay:
> Not sure if it is considered better or not but I was able to get the button text by replacing `{radio.nextSibling}` with `{radio.labels[0].innerHTML}`.

Thanks @capfei, I like this suggestion

@fregante do you have a preference?

source/features/one-click-review-submission.tsx
```diff
@@ -32,7 +31,9 @@ function replaceCheckboxes(originalSubmitButton: HTMLButtonElement): void {
 
 	// Generate the new buttons
 	for (const radio of radios) {
-		const tooltip = radio.parentElement!.getAttribute('aria-label');
+		const labelNode = radio.labels?.[0];
+
+		const tooltip = labelNode?.nextElementSibling?.textContent;
```
RobHannay, line 36: I've also just amended this so that we get those caption descriptions in the button tooltips.

```html
<div class="FormControl-radio-wrap">
  <input aria-describedby="validation-a44ca7e3-1a4e-40ba-9969-cac56e1cf44b caption-a44ca7e3-1a4e-40ba-9969-cac56e1cf44b" class="FormControl-radio" type="radio" value="comment" checked="checked" name="pull_request_review[event]" id="pull_request_review[event]_comment">
  <span class="FormControl-radio-labelWrap">
    <label class="FormControl-label" for="pull_request_review[event]_comment">
      Comment
    </label>       
     <span class="FormControl-caption" id="caption-a44ca7e3-1a4e-40ba-9969-cac56e1cf44b">Submit general feedback without explicit approval.</span>

  </span>
</div>
```

Again, it would possibly be less brittle to select another way, like using the describedby IDs to get the elements that truly describe them, but that requires us to build another query. WDYT?

RobHannay: image
image
image

fregante:
It's safer to deal with nodes, `innerHTML` triggers extra checks by Mozilla https://stackoverflow.com/q/45579400/288906

Also I'm not sure I see the advantage in your later commit. It's best to use selectors and limit index-based (`[0]`) and repeated DOM traversing (`nextElementSibling`)


Something you could optimize for would be to maintain compatibility with the older version so that we don't break GitHub Enterprise users

RobHannay:
> It's safer to deal with nodes, `innerHTML` triggers extra checks by Mozilla [stackoverflow.com/q/45579400/288906](https://stackoverflow.com/q/45579400/288906)

This is useful, thanks. I still kept `textContent` instead of `innerHTML` - does that mitigate the above?

> Something you could optimize for would be to maintain compatibility with the older version so that we don't break GitHub Enterprise users

Ah yes, what's the usual process with this? When features get yolo-disabled I assume that also disables for GHE users?

source/features/one-click-review-submission.tsx
fregante: Why did the previous version not work?

fregante:
> This is useful, thanks. I still kept `textContent` instead of `innerHTML` - does that mitigate the above?
> 
> >

`textContent` is perfectly fine 👍 

> > GitHub Enterprise users
> 
> Ah yes, what's the usual process with this?

It depends. For less visible features we just make the changes and wait for GHE users to open issues or submit patches, if they're necessary.

This feature is rather visible and quite important, so we can't risk breaking GHE users like #6963.

For GitHub changes that are introduced via Feature Preview, it's easy to toggle back and forth and see both DOM versions. In this case:

- we write code that can work in both cases (e.g. keep the `form` attribute), OR
- we use detection (e.g. `if this element exist, OLD CODE, otherwise NEW CODE`), keeping the changes to a minimum, OR
- we keep a whole new copy of the code in cases where that detection makes the code too messy. See [#](https://github.com/refined-github/refined-github/pull/6927/files)

I suggest trying the first one, then we can tag a GHE user to test it out.

fregante:
> When features get yolo-disabled I assume that also disables for GHE users?

Yes. That also means that if we ship a feature that breaks GHE, we have to yolo-disable it on GitHub.com too

fregante:
@busches can you paste the form’s HTML here so that we can work on a compatible fix?

source/features/one-click-review-submission.tsx
RobHannay: Nothing - the radio's parent (see the `div` above) doesn't have an `aria-label` attribute anymore because it's now properly labelled by the linked `<label>` element

busches:
Here's what it looks like on GHE 3.9.4:
image

Here's the HTML:
[modal.txt](https://github.com/refined-github/refined-github/files/12749744/modal.txt)

source/features/one-click-review-submission.tsx
```diff
@@ -30,9 +29,13 @@ function replaceCheckboxes(originalSubmitButton: HTMLButtonElement): void {
...
 	}
 
+	const formAttribute = originalSubmitButton.getAttribute('form')!;
+
 	// Generate the new buttons
 	for (const radio of radios) {
-		const tooltip = radio.parentElement!.getAttribute('aria-label');
+		const labelNode = radio.labels?.[0];
+
+		const tooltip = labelNode?.nextElementSibling?.textContent;
```
RobHannay, line 38: So, weirdly it looks from the previous DOM structure like there wasn't an `aria-label` on its parent anyway. I suspect that could have broken a while ago. Not sure exactly what the intended behaviour was before, but I've actually re-removed the hybrid approach because I don't believe it was doing anything

source/features/one-click-review-submission.tsx
```diff
@@ -47,14 +50,15 @@ function replaceCheckboxes(originalSubmitButton: HTMLButtonElement): void {
...
+				// Old version of GH don't nest the submit button inside the form, so must be linked manually. Issue #6963.
 				form={formAttribute}
 				value={radio.value}
 				className={classes.join(' ')}
 				aria-label={tooltip!}
 				disabled={radio.disabled}
 			>
-				{radio.nextSibling}
+				{/* Old || new. Backwards-compat label content queries. Issue #6963. */}
+				{radio.nextSibling?.textContent?.trim() || labelNode?.textContent}
```
RobHannay, line 61: So this label fix works by:

- Trying to use the old approach _first_, 
- Failing properly if `nextSibling` doesn't contain any text (in the new structure, it was containing a line break hence the trim)
- Falling back to the new way of doing labelling which only has the single-word labels, where the description is now somewhere else

RobHannay:
I've hit two conflicting lint rules.

I want to use a logic OR for the label fallback; I can't use nullish coalescing operator because the "old" label will be an empty string if it doesn't exist. So I changed to use a ternary op `a ? a : b` which tbf is called out, and is worse imo. 

Should I lint-ignore the nullish operator one or what? Is that only because it doesn't have enough type information?

source/features/one-click-review-submission.tsx
```diff
@@ -30,9 +29,13 @@ function replaceCheckboxes(originalSubmitButton: HTMLButtonElement): void {
 		);
 	}
 
+	const formAttribute = originalSubmitButton.getAttribute('form')!;
```
fregante, line 32: Let's move this back to line 12 to keep the name history

fregante:
I hope to see this completed soon so I can merge it ☺️ if you see any awkward lint errors you can leave them broken and I'll take care of them

fregante:
@busches could you help bring this over to the finish line so I can publish it without having to disable the feature on GHE?

busches:
@fregante I probably won't have time till Friday to take a look. If you have something to test, I get to that tomorrow.

source/features/one-click-review-submission.tsx
busches: This label is grabbing too much on GHE:
```
<label class="d-block">
        <input
          type="radio"
          name="pull_request_review[event]"
          value="comment"
          checked=""
        />
        Comment
        <p class="text-normal color-fg-muted">
          Submit general feedback without explicit approval.
        </p>
      </label>
```
Can do something "fun" like:
```
const label = labelNode?.nextElementSibling;
const nestedDescription = label.getElementsByTaName('p'); // GHE has a nested P with description inside the label
if (nestedDescription.length) {
   label.removeChild(nestedDescription[0]);
}
const tooltip = label.textContent;
```

I'm sure there's a better way to achieve that, but that should get us there.

fregante:
All good now. Testing for the GHE version was as easy as pasting that HTML replacing the original `<form>`, and the selector-observer automatically reapplied the changes.  Thanks @busches!

I also made a change that was long overdue: I dropped the green background from the "comment" button. It was technically correct (green = default action, on GitHub) but it was confusing (green = approval)

Screenshot 6

PR #6986, opened 2023-10-16
Escape `.` in the tooltips of the one-click review radio buttons

Jiralite:
The regular expression `/.$/` did not escape the `.`. Observe:

image

Only one of these ends with a `.` What `/.$/` did was transform them all to reduce one character at the end, leading to this:

image

Whilst this will be right for commenting, it was not right for the rest. Simply escaping the `.` will resolve this issue:

https://github.com/refined-github/refined-github/assets/33201955/81cd80b2-8315-4689-bdc3-e5a8a3ee19ca


fregante:
I'm stupid 🥲 😂

Thank you for the fix, I forgot to escape it

fregante (approved):

Issue #6967, opened 2023-09-28
`repo-header-info` misaligned


fregante:
### Description

I'm just realizing that the text isn't well aligned, it should probably be aligned to the repo name’s baseline. `ci-link` might also need adjustments 

<img width="397" alt="Screenshot 4" src="https://github.com/refined-github/refined-github/assets/1402241/150f5c2f-47cf-4397-ae40-d4d9201bee9a">

### How to replicate the issue + URL

This page

### Extension version

23.9.21

### Browser(s) used

Safari

fregante:
`ci-link` fell down this week so that's probably what's making it more noticeable

Last week in https://github.com/refined-github/refined-github/pull/6932:


<img width="197" alt="Screenshot 6" src="https://github.com/refined-github/refined-github/assets/1402241/dff94829-ec0f-4526-9b71-84cc9c9be305">


Today:

<img width="199" alt="Screenshot 6" src="https://github.com/refined-github/refined-github/assets/1402241/de7931bf-869e-42af-aa7f-8f452e0a3d01">

PR #6987, opened 2023-10-17
`repo-header-info` - Improve alignment 

134130:
## Describe
- Closes: #6967 

## Test URLs

### Original 😱 
original

### `vertical-align: text-top` ✅ 
text-top

### `vertical-align: bottom` 🤔 
vertical-align-bottom

### DIFF
--- a/source/features/repo-header-info.tsx
+++ b/source/features/repo-header-info.tsx
@@ -49,7 +49,7 @@ async function add(repoLink: HTMLAnchorElement): Promise<void> {
 			<a
 				href={buildRepoURL('stargazers')}
 				title={`Repository starred by ${stargazerCount.toLocaleString('us')} people`}
-				className="d-inline-flex flex-items-center flex-justify-center mr-1 gap-1 v-align-bottom color-fg-muted"
+				className="d-flex flex-items-center flex-justify-center mr-1 gap-1 color-fg-muted"
 			>
 				<StarIcon className="ml-1" width={12} height={12}/>
 				<span className="f5">{abbreviateNumber(stargazerCount)}</span>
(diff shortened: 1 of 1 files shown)


source/features/repo-header-info.tsx
```diff
@@ -49,10 +49,12 @@ async function add(repoLink: HTMLAnchorElement): Promise<void> {
 			<a
 				href={buildRepoURL('stargazers')}
 				title={`Repository starred by ${stargazerCount.toLocaleString('us')} people`}
-				className="d-flex flex-items-center flex-justify-center mr-1 gap-1 color-fg-muted"
+				className="d-inline-block"
 			>
-				<StarIcon className="ml-1" width={12} height={12}/>
-				<span className="f6">{abbreviateNumber(stargazerCount)}</span>
+				<div className="d-inline-flex flex-items-center flex-justify-center mr-1 gap-1 v-align-text-top color-fg-muted">
```
FloEdelmann, line 54: Can't you avoid the wrapper `<div>` by applying exactly these classes to `<a>`?

134130: Thank you!
I thought It was not changed when without wrapping because of my old eyes issue.

fregante:
The alignment in your screenshot appears to be correct, but we should also match the font size exactly

Screenshot 2

fregante:
Also I think this isn't actually affecting Safari for some reason. It looks exactly the same. The Star alignment is also bad there 😕 


Screenshot 5

134130:
image

Please can you check it is aligned on safari?

fregante (approved):
The two classes don't change anything in Safari, do they change the alignment in Chrome?

The font size alone actually fixes the alignment though:

Screenshot 2

134130:
These were remnants of trying with wrapper.

PR #6991, opened 2023-10-20
Remind users to be kind when commenting

fregante:
Minor feature, similar in spirit to `netiquette` and the now defunct [`rgh-sponsor-button`](https://github.com/refined-github/refined-github/pull/4003). 

For now I made it a "private" feature because it's innocuous and "bug-free"

## Before

Screenshot

## After

Screenshot 2


fregante:
I'll roll this into `netiquette` actually, the description still matches:

> Adds unobtrusive netiquette reminders.

doc-sheet:
How do I disable it?

alextes:
@doc-sheet I came here looking for the same thing.

You can toggle netiquette off in the extensions options:

Screenshot 2023-11-02 at 10 41 31

To give some context, I actually love kindness and compassion. I'm an avid fan of meditation and have spent many hours doing nothing but reorient my mind towards loving kindness 💗. So why in the world would I want to disable this?! For me, getting reminded to "be kind" feels a little uncomfortable unless I am being unkind. By default, I try to be kind! No doubt, not what the author intended but nevertheless how I feel. If the comment makes you smile instead, or revise something unkind you wrote, great, keep it on! Whether this is a good feature in general, no idea. Depends on if your user-base acts more kind because of it or less so.

doc-sheet:
> You can toggle netiquette off

I'm ok with netiquette it's quite useful. I just don't like this particular feature.

pankgeorg:
I want to thank whoever thought of this (@fregante? 💖). "Commenting" and "being kind" are usually found far from each other. I believe what is best in humanity lies in the intersection and it's easy to forget, especially in challenging times. Thanks for the reminded. Saw this and really moved me, and made my day.

Thanks 💖

zot:
Yes, I saw that today also and it made a really good impression -- THANKS!!!

fregante:
If you find "Be kind" to be passive aggressive, you might want to think about _why_ you see it that way. It's not a reprimand, it's not an order. It's an invitation to be mindful, in a similar spirit to "Be safe" and "Take care"

PR #7004, opened 2023-10-29
`small-user-avatars` - Adjust spacing

### LINKED ISSUE #7002: `small-user-avatars` needs more spacing before a user mention
### Description

The avatar starts significantly closer to the previous word than a normal letter would, which looks especially bad when the user mention is the first word of a new sentence.

Zoomed in using Windows Magnifier at 600%, with a vertical line added:
image

Zoomed at 110% using Firefox:
image

https://github.com/electron/forge/pull/3391#issue-1965984062

### Extension version

23.10.5

### Browser(s) used

Firefox

Katsute:
Fixes #7002 

Adds left margin to user avatars to better match word spacing

## Test URLs

<table><td>
Hello world. @Katsute
<br>
Hello world. Omniscience 
</td></table>

<table><td>
@Katsute
<br>
Ordinary
</td></table>

## Screenshot

image
↓
image

### DIFF
--- a/source/features/highlight-collaborators-and-own-conversations.css
+++ b/source/features/highlight-collaborators-and-own-conversations.css
@@ -3,9 +3,9 @@
 	border-radius: 2em;
 	padding: 2px 5px;
 
-    .rgh-small-user-avatars {
-        margin-left: -3px;
-        margin-right: 2px !important;
-    }
+	.rgh-small-user-avatars {
+		margin-left: -3px;
+		margin-right: 2px !important;
+	}
 
 }
--- a/source/features/highlight-collaborators-and-own-conversations.css
+++ b/source/features/highlight-collaborators-and-own-conversations.css
@@ -7,5 +7,4 @@
 		margin-left: -3px;
 		margin-right: 2px !important;
 	}
-
 }
--- a/source/features/highlight-collaborators-and-own-conversations.css
+++ b/source/features/highlight-collaborators-and-own-conversations.css
@@ -2,4 +2,10 @@
 	border: 1px solid var(--color-border-default);
 	border-radius: 2em;
 	padding: 2px 5px;
+
+    .rgh-small-user-avatars {
+        margin-left: -3px;
+        margin-right: 2px !important;
+    }
+
 }
--- a/source/features/small-user-avatars.css
+++ b/source/features/small-user-avatars.css
@@ -5,5 +5,5 @@
 }
 
 .user-mention .rgh-small-user-avatars {
-	margin-left: 2px;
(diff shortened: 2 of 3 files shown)


NoahAndrews:
While I much prefer this to the status quo, now there's a little too much margin to my eyes (please forgive the pickiness, I'm not trying to demand extra work). 

Screenshot 2023-10-31 143308

Screenshot 2023-10-31 143104

Katsute:
image

source/features/small-user-avatars.css
```diff
@@ -3,3 +3,7 @@
 	margin-left: -3px;
 	margin-right: 2px !important;
 }
+
+.user-mention .rgh-small-user-avatars {
+	margin-left: 2px;
```
NoahAndrews, line 8: Will `px` units scale correctly as the text size changes? `em` may be a better option.

Katsute: Avatar images are set to a fixed size by this feature, the margin would be correct regardless of text size

fregante:
Thanks for looking into this! The issue is due to the image's border not using `border` but `box-shadow`, so it doesn't expand the `img`'s box. The border here is `1px` so we should match that, it seems that when comparing it to the letter O it aligns decently, while 2px seems too much (especially on a new line)

## 2px

Screenshot 13

## 1px

Screenshot 12

fregante (approved):

source/features/small-user-avatars.css
```diff
@@ -3,3 +3,7 @@
 	margin-left: -3px;
 	margin-right: 2px !important;
 }
+
+.user-mention .rgh-small-user-avatars {
+	margin-left: 1px;
+}
```
fregante, line 9: Actually let's move this to the JSX instead. Then we can move the CSS rule above to `highlight-collaborators-and-own-conversations` instead and drop this CSS file altogether. You can:

- https://github.com/refined-github/refined-github/issues/6665

Katsute: 766dba633eeeb7d9269ec0077e2f5ad9a07642b0

source/features/highlight-collaborators-and-own-conversations.css
```diff
@@ -2,4 +2,9 @@
 	border: 1px solid var(--color-border-default);
 	border-radius: 2em;
 	padding: 2px 5px;
+
+	.rgh-small-user-avatars {
+		margin-left: -3px;
+		margin-right: 2px !important;
+	}
```
fregante, line 9: Nested CSS like it's 2023😎

fregante (approved):

PR #7005, opened 2023-10-29
`linkify-code` - Fix issue links on global search pages

### LINKED ISSUE #6996: `linkify-code` breaks on search page
### Description

image

Need to update this part to extract info from URL:

https://github.com/refined-github/refined-github/blob/906da8c189127627f23fd3f0a706c2369a79c9b0/source/github-helpers/dom-formatters.tsx#L31-L38

https://github.com/search?q=repo%3Anotion-enhancer%2Fnotion-repackaged%20sqlite&type=code

### Extension version

23.10.5

### Browser(s) used

Chrome

Katsute:
Fixes #6996

Retrieves the repository using the line url when getRepo doesn't work

image

`/notion-enhancer/notion-repackaged/blob/4eefa99dc1bcde18268d70bb2b7f050786229feb/scripts/extract-src.sh#L50`

↑ pull repo from this

## Test URLs

https://github.com/search?q=repo%3Anotion-enhancer%2Fnotion-repackaged%20sqlite&type=code

https://github.com/search?q=org%3Arefined-github+%2F%23%5Cd%2B%2F&type=code

https://github.com/search?q=%2F%23%5Cd%7B4%2C%7D%2F+language%3Atypescript&type=code

## Screenshot

image

---

image

### DIFF
--- a/source/features/linkify-code.tsx
+++ b/source/features/linkify-code.tsx
@@ -29,7 +29,18 @@ function linkifyContent(wrapper: Element): void {
 	// Linkify issue refs in comments, exclude gists:
 	// https://github.com/refined-github/refined-github/pull/3844#issuecomment-751427568
 	if (!pageDetect.isGist()) {
-		const currentRepo = getRepo() ?? {};
+		let currentRepo: {owner?: string; name?: string} | undefined = getRepo();
+
+		if (!currentRepo) { // Code search
+			// Grab repo from line url
+			const lineUrl = wrapper.parentElement!.querySelector('.blob-num a')!.getAttribute('href')!.split('/');
+
+			currentRepo = {
+				owner: lineUrl[1],
+				name: lineUrl[2],
+			};
+		}
+
 		for (const element of select.all('.pl-c', wrapper)) {
 			linkifyIssues(currentRepo, element);
 		}
@@ -63,6 +74,7 @@ void features.add(import.meta.url, {
 
 - Discussions: https://github.com/File-New-Project/EarTrumpet/discussions/877
 - Code Search: https://github.com/search?q=repo%3AKatsuteDev%2FBackground+marketplace&type=code
+- Code Search: https://github.com/search?q=org%3Arefined-github+%2F%23%5Cd%2B%2F&type=code
 - Comment: https://github.com/sindresorhus/linkify-urls/pull/40#pullrequestreview-1593302757
 
 */
--- a/source/features/linkify-code.tsx
+++ b/source/features/linkify-code.tsx
@@ -31,10 +31,9 @@ function linkifyContent(wrapper: Element): void {
 	if (!pageDetect.isGist()) {
 		let currentRepo: {owner?: string; name?: string} | undefined = getRepo();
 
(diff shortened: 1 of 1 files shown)


source/features/linkify-code.tsx
```diff
@@ -29,7 +29,17 @@ function linkifyContent(wrapper: Element): void {
 	// Linkify issue refs in comments, exclude gists:
 	// https://github.com/refined-github/refined-github/pull/3844#issuecomment-751427568
 	if (!pageDetect.isGist()) {
-		const currentRepo = getRepo() ?? {};
```
fregante, line 32: Let's detect the global search page instead. When there, there's not need to call getRepo

source/features/linkify-code.tsx
```diff
@@ -20,23 +20,27 @@ function initTitle(signal: AbortSignal): void {
...
 
-	// Linkify issue refs in comments, exclude gists:
-	// https://github.com/refined-github/refined-github/pull/3844#issuecomment-751427568
-	if (!pageDetect.isGist()) {
-		const currentRepo = getRepo() ?? {};
-		for (const element of $$('.pl-c', wrapper)) {
-			linkifyIssues(currentRepo, element);
-		}
+	const currentRepo = pageDetect.isGlobalSearchResults()
+		? getRepo(wrapper.parentElement!.querySelector('.blob-num a')!.href)
```
fregante, line 33: Are you sure this works? The `wrapper` seems to be a `td`, so the parentElement is a `tr`, which does not contain any links.

Screenshot 14


The right ancestor is probably `Box-sc-g0xbh4-0 kNKjla`, which is hard to target 😬

source/features/linkify-code.tsx
```diff
@@ -63,6 +67,7 @@ void features.add(import.meta.url, {
 
 - Discussions: https://github.com/File-New-Project/EarTrumpet/discussions/877
 - Code Search: https://github.com/search?q=repo%3AKatsuteDev%2FBackground+marketplace&type=code
+- Code Search: https://github.com/search?q=org%3Arefined-github+%2F%23%5Cd%2B%2F&type=code
```
fregante, line 70: Can you find a URL that has multiple results? We want to make sure this detection works even when there are multiple repos on the page. The wrong selector might pick the same/wrong repo for all results.

source/features/linkify-code.tsx
Katsute: The selector targets the line link, rather than the repo link at the top of each file.

`tr (parent) .blob-num a` seems to work properly, seeing as the hovercard shows (screenshot ^)

image

source/features/linkify-code.tsx
Katsute: https://github.com/search?q=%2F%23%5Cd%7B4%2C%7D%2F+language%3Atypescript&type=code

image

image

source/features/linkify-code.tsx
fregante: Oh neat, I thought it was targeting the file header

Issue #7001, opened 2023-10-25
`releases-tab` appears in dropdown even when there are no releases


fregante:
### Description

We're adding the tab regardless of releases:

 https://github.com/refined-github/refined-github/blob/906da8c189127627f23fd3f0a706c2369a79c9b0/source/features/releases-tab.tsx#L72

The code should be altered to query the API the same way we do it for the actual tab. That code can be copy-pasted into this function

<img width="643" alt="Screenshot" src="https://github.com/refined-github/refined-github/assets/1402241/56697979-c7b1-4203-8714-02b1143c8c71">

_Extracted from https://github.com/refined-github/refined-github/issues/6995_

### How to replicate the issue + URL

1. Visit https://github.com/refined-github/yolo
2. Open the dropdown

### Extension version

23.9.21

### Browser(s) used

Safari

PR #7027, opened 2023-11-02
`releases-tab`: Hide dropdown item when there are no releases

kovsu:
## Test URLs

close: #7001 

## Screenshot

after:
image


fregante:
多谢 ☺️

kovsu:
> 多谢 ☺️

:)

PR #7031, opened 2023-11-03
`releases-tab`: Add support for tags-only repos

fregante:
- Closes https://github.com/refined-github/refined-github/issues/6517

## Test URLs

Releases: https://github.com/refined-github/refined-github
Tags: https://github.com/python/cpython

## Screenshot

Screenshot 5


fregante:
I considered also adding support for repos that use neither, and instead just have a changelog file (which we already link to in `link-to-changelog-file`).

However I think these are rare and it's not worth supporting bad git usage.

fregante:
Fast-merging to release a new bug fix version

PR #7042, opened 2023-11-07
`table-input` and `collapsible-content-button` - Silence Firefox error 

fregante:
See https://github.com/refined-github/refined-github/issues/5797#issuecomment-1796191920

### DIFF
--- a/source/features/table-input.css
+++ b/source/features/table-input.css
@@ -30,12 +30,6 @@
 	}
 }
 
-/** TODO: Drop JS-based version after :has() has full support */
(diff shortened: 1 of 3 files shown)


source/features/table-input.css
```diff
@@ -30,12 +30,6 @@
 	}
 }
 
-/** TODO: Drop JS-based version after :has() has full support */
-.rgh-table-input:hover .selected div {
```
fregante, line 34: No full support, but the feature now _requires_ `:has()`, so I'm dropping the fallback.

PR #7051, opened 2023-11-12
`locked-issue` - New feature

### LINKED ISSUE #6551: show "locked" status of issues
### Description

Refined GitHub should show an indicator when an issue is "locked".  I think it's most important when viewing an issue page, but it would also be helpful when showing a list of issues, if that's possible.

### Example URLs

https://github.com/PyCQA/pycodestyle/issues/703

Katsute:
Closes #6551 🎉

Display a locked label on issues and PRs based on the existence of the react button https://github.com/refined-github/refined-github/issues/6551#issuecomment-1651713075

## Test URLs

* unlocked issue: https://github.com/refined-github/refined-github/issues/6551
* locked issue: https://github.com/refined-github/sandbox/issues/74
* unlocked pr: https://github.com/refined-github/refined-github/pull/7051
* locked pr: https://github.com/refined-github/sandbox/pull/48

any issue or pr from these results:

* https://github.com/search?q=org%3Arefined-github+is%3Alocked&type=issues
* https://github.com/search?q=org%3Arefined-github+is%3Alocked&type=pullrequests

## Screenshot

image

image

image

image

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -250,7 +250,7 @@ Thanks for contributing! 🦋🙌
 - [](# "linkify-user-labels") [Links the "Contributor" and "Member" labels on comments to the author’s commits on the repo.](https://user-images.githubusercontent.com/1402241/177033344-4d4eea63-e075-4096-b2d4-f4b879f1df31.png)
 - [](# "jump-to-conversation-close-event") [Adds a link to jump to the latest close event of a conversation.](https://user-images.githubusercontent.com/16872793/177792713-64219754-f8df-4629-a9ec-33259307cfe7.gif)
 - [](# "close-as-unplanned") [Lets you "close issue as unplanned" in one click instead of three.](https://github-production-user-asset-6210df.s3.amazonaws.com/1402241/279745773-709cde60-c26a-4a0e-89e1-56444d25ebdf.png)
-- [](# "locked-issue") [Show a label on locked issues and pulls.](https://user-images.githubusercontent.com/58778985/282343812-a5811a61-0507-4058-bb3e-299a59bcbc71.png)
+- [](# "locked-issue") [Show a label on locked issues and pulls.](https://github.com/refined-github/refined-github/assets/58778985/aed20e5b-6a8e-45f7-af2d-99928d18875d)
 
 <!-- Refer to style guide above. Keep this message between sections. -->
 
--- a/readme.md
+++ b/readme.md
@@ -250,6 +250,7 @@ Thanks for contributing! 🦋🙌
 - [](# "linkify-user-labels") [Links the "Contributor" and "Member" labels on comments to the author’s commits on the repo.](https://user-images.githubusercontent.com/1402241/177033344-4d4eea63-e075-4096-b2d4-f4b879f1df31.png)
 - [](# "jump-to-conversation-close-event") [Adds a link to jump to the latest close event of a conversation.](https://user-images.githubusercontent.com/16872793/177792713-64219754-f8df-4629-a9ec-33259307cfe7.gif)
 - [](# "close-as-unplanned") [Lets you "close issue as unplanned" in one click instead of three.](https://github-production-user-asset-6210df.s3.amazonaws.com/1402241/279745773-709cde60-c26a-4a0e-89e1-56444d25ebdf.png)
+- [](# "locked-issue-indicator") [Show a label on locked issues and pulls.]()
 
 <!-- Refer to style guide above. Keep this message between sections. -->
 
--- a/readme.md
+++ b/readme.md
@@ -250,7 +250,7 @@ Thanks for contributing! 🦋🙌
 - [](# "linkify-user-labels") [Links the "Contributor" and "Member" labels on comments to the author’s commits on the repo.](https://user-images.githubusercontent.com/1402241/177033344-4d4eea63-e075-4096-b2d4-f4b879f1df31.png)
 - [](# "jump-to-conversation-close-event") [Adds a link to jump to the latest close event of a conversation.](https://user-images.githubusercontent.com/16872793/177792713-64219754-f8df-4629-a9ec-33259307cfe7.gif)
 - [](# "close-as-unplanned") [Lets you "close issue as unplanned" in one click instead of three.](https://github-production-user-asset-6210df.s3.amazonaws.com/1402241/279745773-709cde60-c26a-4a0e-89e1-56444d25ebdf.png)
-- [](# "locked-issue-indicator") [Show a label on locked issues and pulls.](https://user-images.githubusercontent.com/58778985/282315306-7c323583-df1a-4fd5-bc13-79c89cdf4144.png)
+- [](# "locked-issue-indicator") [Show a label on locked issues and pulls.](https://user-images.githubusercontent.com/58778985/282343812-a5811a61-0507-4058-bb3e-299a59bcbc71.png)
 
 <!-- Refer to style guide above. Keep this message between sections. -->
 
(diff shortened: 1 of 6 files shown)


source/features/locked-issue-indicator.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+import onetime from 'onetime';
+
+import {$, elementExists} from 'select-dom';
+
+import features from '../feature-manager.js';
+import observe from '../helpers/selector-observer.js';
+
+function addLock(element: HTMLElement): void {
+	if (!elementExists('reactions-menu', element)) {
```
fregante, line 13: This can be moved to the observed selector, as `:not(:has(reactions-menu))`, I think

source/features/locked-issue-indicator.tsx
```diff
@@ -0,0 +1,57 @@
...
+				</span>
+			</div>,
+		);
+		// Add locked indicator to sticky header
+		$('.gh-header-sticky .flex-row')!.children[0].after(
+			<div className="mr-2 mb-2 flex-shrink-0">
+				<span title="Status: Locked" className="State d-flex flex-items-center">
+					<LockIcon className="flex-items-center mr-1"/>
+					Locked
+				</span>
```
fregante, line 29: This span can be extracted into its own function in this file, e.g. 

```js
const Icon = () => <span>etc</span>
``` 

And then you can use it as `<Icon/>`

source/features/locked-issue-indicator.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+import {$, elementExists} from 'select-dom';
+
+import features from '../feature-manager.js';
+import observe from '../helpers/selector-observer.js';
+
+function addLock(element: HTMLElement): void {
+	if (!elementExists('reactions-menu', element)) {
+		// Add locked indicator to header
+		$('.gh-header-meta')!.children[0].after(
```
fregante, line 15: You're looking for `> :first-child`

fregante (commented):
Thank you! I'm a bit concerned about observing just the reactions; if GitHub updates the DOM via Ajax the indicator is lost. 

I think it should observe the two headers instead, but also include the "locked" logic in the selector itself, via `:has()`. For example:

```js
observe(':has(.pr-review-reactions:not(:has(reactions-menu))) .gh-header-meta', add)
```

Edit: `:has()` cannot be nested 😩 

Edit: there's a way around it:

```css
:has(.js-pick-reaction:first-child) .gh-header-meta
```

If `reactions-menu` exists, then `.js-pick-reaction` is the second child

source/features/locked-issue-indicator.tsx
Katsute: The positioning was copied from the state label:

image

image

Katsute: Discussions also uses this positioning, however they omit both labels in the sticky header

image

image

Katsute:
This feature does indeed break when the locked state is changed while a user is looking at the issue.

There seems to be some kind of timing difference between when the header is updated and when the reaction button is changed.

* The observer fires when the issue is **both** locked and unlocked (even though the reaction menu _should_ not be there)
* Adding a redundant `elementExists` check in the add function doesn't fix this either, it returns true for the existence of the reaction button
* Running a query selector in the console does find that the reaction button does indeed not exist

I'm not sure how to resolve this issue.

Katsute:
Solved by adding the below css to hide the locked state:

```css
:has(reactions-menu) .rgh-locked-issue-indicator {
	display: none;
}
```

https://github.com/refined-github/refined-github/assets/58778985/64b263ff-d7fa-478a-8696-273e773484d6

readme.md
```diff
@@ -250,6 +250,7 @@ Thanks for contributing! 🦋🙌
 - [](# "linkify-user-labels") [Links the "Contributor" and "Member" labels on comments to the author’s commits on the repo.](https://user-images.githubusercontent.com/1402241/177033344-4d4eea63-e075-4096-b2d4-f4b879f1df31.png)
 - [](# "jump-to-conversation-close-event") [Adds a link to jump to the latest close event of a conversation.](https://user-images.githubusercontent.com/16872793/177792713-64219754-f8df-4629-a9ec-33259307cfe7.gif)
 - [](# "close-as-unplanned") [Lets you "close issue as unplanned" in one click instead of three.](https://github-production-user-asset-6210df.s3.amazonaws.com/1402241/279745773-709cde60-c26a-4a0e-89e1-56444d25ebdf.png)
+- [](# "locked-issue-indicator") [Show a label on locked issues and pulls.](https://user-images.githubusercontent.com/58778985/282343812-a5811a61-0507-4058-bb3e-299a59bcbc71.png)
```
fregante, line 253: Can you take a "retina" screenshot? You can zoom the page at 200% and then take a screenshot.

Also we can shorten the name to `locked-issue`

source/features/locked-issue-indicator.css
```diff
@@ -0,0 +1,3 @@
+:has(reactions-menu) .rgh-locked-issue-indicator {
```
fregante, line 1: Let's use the same selector 
```suggestion
:has(.js-pick-reaction:first-child) .rgh-locked-issue-indicator {
```

source/features/locked-issue-indicator.tsx
```diff
@@ -0,0 +1,60 @@
+import './locked-issue-indicator.css';
+
+import React from 'react';
+import {LockIcon} from '@primer/octicons-react';
+import * as pageDetect from 'github-url-detection';
+
+import {$} from 'select-dom';
```
fregante, line 7: The space is only needed between `node_modules` and local imports
```suggestion
import './locked-issue-indicator.css';
import React from 'react';
import {LockIcon} from '@primer/octicons-react';
import * as pageDetect from 'github-url-detection';
import {$} from 'select-dom';
```

source/features/locked-issue-indicator.tsx
```diff
@@ -0,0 +1,60 @@
...
+	$('.gh-header-sticky .flex-row > :first-child')!.after(
+		<div className="mr-2 mb-2 flex-shrink-0 rgh-locked-issue-indicator">
+			<LockedIndicator/>
+		</div>,
+	);
+}
+
+function init(signal: AbortSignal): void {
+	// If reactions-menu exists, then .js-pick-reaction is the second child
+	observe(':has(.js-pick-reaction:first-child) .gh-header-meta', addLock, {signal});
```
fregante, line 39: `observe` acts as a selector, we should not need to also use `$` in `addLock`
```suggestion
	observe(':has(.js-pick-reaction:first-child) .gh-header-meta > :first-child', addLock, {signal});
	observe(':has(.js-pick-reaction:first-child) .gh-header-sticky .flex-row > :first-child', addLockToSticky, {signal});
```

source/features/locked-issue-indicator.tsx
```diff
@@ -0,0 +1,60 @@
...
+}
+
+void features.add(import.meta.url, {
+	asLongAs: [
+		isHasSelectorSupported,
```
fregante, line 44: 💙

source/features/locked-issue-indicator.css
Katsute: The selector is intentional, it needs to re-hide the locked label caused by https://github.com/refined-github/refined-github/pull/7051#issuecomment-1809450140. It should only activate when the reaction button does exist, not when it doesn't.

readme.md
Katsute: image

source/features/locked-issue-indicator.css
fregante: Oh right, so it should be 
```suggestion
:has(.js-pick-reaction:not(:first-child)) .rgh-locked-issue-indicator {
```

The point is that we should try to use the same exact logic or else it might get out of sync. If GitHub changes the element name, the feature works, but not this piece, and we'll never notice.

fregante (approved):
Screenshot 5

fregante:
Thanks @Katsute! We'll forever wonder how we lived without this

tofagerl:
This breaks in the following case:
- Set repo to require X approvals
- Set a team as required approvals
- Reach X approvals where X < group size
- (Reaction button disappears, since it's no longer necessary)
- New user from group opens PR, and gets told that it's locked, even though it isn't.

tofagerl:
This breaks in the following case:
- Set repo to require X approvals
- Set a team as required approvals
- Reach X approvals where X < group size
- (Reaction button disappears, since it's no longer necessary)
- New user from group opens PR, and gets told that it's locked, even though it isn't.

fregante:
> Reaction button disappears, since it's no longer necessary

What's not necessary? Approvals are unrelated to reactions

tofagerl:
Huh. I dug around, and it turns out we just don't have reactions on any PRs. 
Note that this is GHES 3.8.

fregante:
@tofagerl indeed, see https://github.com/refined-github/refined-github/issues/7063

PR #7053, opened 2023-11-13
Fix "Hotfixes are not applied on GitHub Enterprise." error

fregante:
This has been here ever since I added this feature and nobody ever complained 😅 

Screenshot

### DIFF
--- a/source/options.tsx
+++ b/source/options.tsx
@@ -9,7 +9,7 @@ import {assertError} from 'ts-extras';
 import * as indentTextarea from 'indent-textarea';
 import delegate, {DelegateEvent} from 'delegate-it';
 import {isChrome, isFirefox} from 'webext-detect-page';
-import {isEnterprise} from 'github-url-detection';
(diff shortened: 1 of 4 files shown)


source/options.tsx
```diff
@@ -9,7 +9,7 @@ import {assertError} from 'ts-extras';
 import * as indentTextarea from 'indent-textarea';
 import delegate, {DelegateEvent} from 'delegate-it';
 import {isChrome, isFirefox} from 'webext-detect-page';
-import {isEnterprise} from 'github-url-detection';
```
fregante, line 12: `isEnterprise` is always true in the options, because "the location is not GitHub.com" 🤦‍♂️ 

I fixed this detection

busches:
I have been annoyed that hotfixes disable GHE, but never realized that was not the intended behavior :D

fregante:
We don't apply CSS hot fixes to GHE because they're meant to fix brand new changes on GitHub.com. The bug here was that the GHE message was showing for everyone.

PR #7055, opened 2023-11-13
`prevent-comment-loss` - New feature

### LINKED ISSUE #5680: Open links in a new tab while writing comments
### Description

I don't expect to navigate away while I'm writing a comment, but that's what GitHub does.

This feature can probably be limited to links in the preview and in the issue templates. I assume it's something as easy as:

```js
delegate('form a,.preview a,.etc a', 'click', openInNewTab)
```

### Screenshot

https://user-images.githubusercontent.com/1402241/172036862-8a22803b-a2c0-48a1-b0c9-3ff2ed5edda8.mov

### Example URL

- https://github.com/refined-github/refined-github/issues/new?assignees=&labels=enhancement%2C+under+discussion&template=2_feature_request.yml 
- https://github.com/refined-github/shorten-repo-url/issues/new 
- Probably hasRichTextEditor and isEditingFile because both have a preview

ansg191:
New feature will open links inside GitHub's Rich Text editor preview in a new tab instead of navigating away from the page.

Closes #5680 

## Test URLs

https://github.com/refined-github/sandbox/issues/new
https://github.com/refined-github/refined-github/edit/main/readme.md

## Screenshot

editor-link-new-tab


source/features/editor-link-new-tab.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+import features from '../feature-manager.js';
+
+function openInNewTab(event: DelegateEvent<MouseEvent, HTMLLinkElement>): void {
+	event.preventDefault();
+	window.open(event.delegateTarget.href, '_blank');
+}
+
+function init(signal: AbortSignal): void {
+	delegate('div.js-preview-body a, div.html-blob a', 'click', openInNewTab, {signal});
```
fregante, line 12: It should wrap the listener with this https://github.com/fregante/filter-altered-clicks

readme.md
```diff
@@ -390,6 +390,7 @@ Thanks for contributing! 🦋🙌
 - [](# "prevent-pr-merge-panel-opening") Prevents the merge panel from automatically opening on every page load after it’s been opened once.
 - [](# "command-palette-navigation-shortcuts") Adds keyboard shortcuts to select items in command palette using <kbd>ctrl</kbd> <kbd>n</kbd> and <kbd>ctrl</kbd> <kbd>p</kbd> (macOS only).
 - [](# "submission-via-ctrl-enter-everywhere") Enables submission via <kbd>ctrl</kbd> <kbd>enter</kbd> on every page possible.
+- [](# "editor-link-new-tab") [Opens links in rich text editor in new tab instead of navigating away from the page.](https://github.com/refined-github/refined-github/assets/17681399/e18ca488-e055-4e41-943c-6243d9582bb7)
```
fregante, line 393: This is a perfect demo, thank you!

However:

- Only pngs and gifs are allowed
- This type of link requires a redirect, so we should use the final address instead, which includes the file extension.

source/features/editor-link-new-tab.tsx
fregante, line 12:
```suggestion
	delegate([
'div.js-preview-body a', // on page abc
'div.html-blob a', // on page def
], 'click', openInNewTab, {signal});
```

We should describe each selector wherever possible

fregante (approved):

fregante:
@yakov116 @kidonng thoughts on the feature name and scope?

yakov116:
This is good unless you want to add the `preview` to it

FloEdelmann:
Maybe `preview-links-new-tab`?

cooljeanius (approved):

grafst:
If it's not only in comments but allso in issues, bugs, PR's etc then it maybe should be called `prevent-draft-loss`?
or maybe `prevent-leaving-preview`?

fregante:
`prevent-draft-loss` would be good.

`prevent-leaving-preview` is nice, but I think it sounds less important than "prevent loss"

I think "comment" like "issue" can be used as generic word for "text, content, anything you wrote that isn't a file", so `prevent-comment-loss` is fine too.

FloEdelmann:
`draft` in GitHub immediately leads my mind to "Draft pull request", rather than "draft comments"/"draft texts".

fregante:
@FloEdelmann we also use this word in `unfinished-comments`

FloEdelmann:
Maybe we could change that to something like "Unsaved comment"?

fregante:
Tabs are often very short, so anything longer than "draft" would probably be less effective.

fregante:
Looks great! Thank you Anshul

Julian-Dumitrascu:
It's worked for me only when I followed your instructions exactly.
This feature is checked in the list displayed in the Options window of your extension, so a user can expect it to be enabled.
I've restarted the Vivaldi browser and asked your extension to update, but this seems to have made no difference.
I'd like this feature available in any program. 
What should you and I do to enable it in GitHub?

Julian-Dumitrascu:
> expect

I've read again the description of this feature. I want this feature for any link e.g. from an issue.
When you're going to talk again online about your software, we can [choose some software](https://github.com/Sol-Global-Management/computer-software/issues/3) for this. (That page should open in a new tab.)

fregante:
You can write a userscript to do that. This feature and this extension won't.

PR #7058, opened 2023-11-15
`wait-for-attachments` - Drop feature

fregante:
It seems that it now works on new comments too

## Test URLs

Here

## Related

- https://github.com/refined-github/refined-github/issues/7011


fregante:
testing

fregante:
testing rgh

fregante:
The heck?

The button is correctly disabled natively, but neither GitHub nor Refined GitHub actually block submission via cmd-enter. It seems that the feature is still needed, but it would need to be rewritten to actually block submissions.

I'll drop it for now and open an issue

Issue #7063, opened 2023-11-15
`locked-issue` - Showing everything as locked on Enterprise Github


Johannes-Andersen:
### Description

It seems like the new locked issue icon introduced here: https://github.com/refined-github/refined-github/pull/7051 does not work as expected on github enterprise

All issues and PRs show as locked, even when they aren't 

![Screenshot 2023-11-15 at 14 02 34](https://github.com/refined-github/refined-github/assets/57638117/0f904959-c9ca-4db0-b96b-db6a752af914)
![Screenshot 2023-11-15 at 14 05 19](https://github.com/refined-github/refined-github/assets/57638117/8f9f2cb4-2577-45ec-98a2-1f23d8d72141)
(both images have the PR and Issue details hidden to not show internal information)

Server:
GitHub Enterprise Server 3.8.11

### How to replicate the issue + URL

https://github.<enterprise_url>.io/<org>/<repo>/pull/<pr>
https://github.<enterprise_url>.io/<org>/<repo>/issues/<issue>

### Extension version

23.11.15

### Browser(s) used

Chrome - 119.0.6045.159

fregante:
Thank you for the report. You can temporarily disable the feature in your Refined GitHub options, selecting your GHE domain

fregante:
I'll release a new version just for this

fregante:
It will be fixed in `23.11.15.1433`, available soon. Here some live badges with the latest version on each store:



[link-chrome]: https://chrome.google.com/webstore/detail/refined-github/hlepfoohegkhhmjieoechaddaejaokhf 'Version published on Chrome Web Store'
[link-firefox]: https://addons.mozilla.org/firefox/addon/refined-github-/ 'Version published on Mozilla Add-ons'
[link-safari]: https://apps.apple.com/app/id1519867270 'Version published on the Mac App Store'

[<img src="https://raw.githubusercontent.com/alrra/browser-logos/90fdf03c/src/chrome/chrome.svg" width="48" alt="Chrome" valign="middle">][link-chrome] [<img valign="middle" src="https://img.shields.io/chrome-web-store/v/hlepfoohegkhhmjieoechaddaejaokhf.svg?label=%20">][link-chrome] 

[...]

PR #7072, opened 2023-11-16
`locked-issue` - Disable on logged out users due to false positives

fregante:
Every issue is marked as locked when logged out because logged out users cannot react.

cc @Katsute we might need an improved logic in the future, also considering:

- #7063 

## Test URLs

1. Log out
2. Visit any conversation, like https://github.com/refined-github/refined-github/pull/7051

## Before

Screenshot

## After

Screenshot 4

### DIFF
--- a/source/features/locked-issue.tsx
+++ b/source/features/locked-issue.tsx
@@ -30,8 +30,9 @@ function addStickyLock(element: HTMLElement): void {
 
 function init(signal: AbortSignal): void {
 	// If reactions-menu exists, then .js-pick-reaction is the second child
-	observe(':has(.js-pick-reaction:first-child) .gh-header-meta > :first-child', addLock, {signal});
-	observe(':has(.js-pick-reaction:first-child) .gh-header-sticky .flex-row > :first-child', addStickyLock, {signal});
+	// Logged out users never have the menu, so they should be excluded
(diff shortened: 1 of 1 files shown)


Katsute:
**This solution might work:**

For non-collaborators, simply base the lock status on the comment box:

image

For collaborators, simply base the lock status based on the lock/unlock button:
   
image

We also wouldn't need to check if the user is even a collaborator, these items are mutually exclusive; nons will never see the unlock/lock button and collabs won't ever see the locked comment box.

fregante:
Those are good. Pros:

- fewer chances of false positives

Cons:

- the entire page has to load before they appear in the DOM

```css
body:has(
	.js-lock-issue [action$="/unlock"],
	#new_comment_form .blankslate
) .gh-header-meta > :first-child
```

> For non-collaborators, simply base the lock status on the comment box:

This also includes archived repos though, which are technically not "locked": https://github.com/isaacs/github/issues/1710

Katsute:
Might need to disable the feature on archived repos, doesn't seem to have any indicators we can use to guess the lock state. I think its a given that archived repos are _'locked'_ so the feature might not be needed.

Katsute:
Been testing this on my [fork](https://github.com/Katsute/refined-github/tree/locked-issue-selector), everything works except the comment form detection when the locked state changes. The comment form does not update when an issue is locked/unlocked, which is weird because that is one thing that should update when the lock state changes.

PR #7074, opened 2023-11-17
fix(list-prs-for-file): Added css rule to fix stacking order

### LINKED ISSUE #6845: `list-prs-for-file`: drop down covered by code
### Description

List of PRs is being covered by code.

image

image

https://github.com/hashicorp/vault/blob/main/go.mod

Or any file page that has more than 2 open PRs against it.

### Extension version

23.8.3

### Browser(s) used

Firefox

Askath:
This PR should fix #6845. 

This change affects the list-pr-for-file feature. The dropdown menu got overlapped by the editors header which seemed to be a order stack issue. 

I added a small css to set the header to a different zindex which seemed to fix the issue without breaking other features.

## Test URLs

https://github.com/hashicorp/vault/blob/main/go.mod

image


Askath:
found issue regarding other features

fregante:
Thank you for the PR, but negative z-index are never the answer in this case, they always have a risk of going under their own parent and becoming unclickable.

Askath:
> Thank you for the PR, but negative z-index are never the answer in this case, they always have a risk of going under their own parent and becoming unclickable.

Ye I noticed some issues. I am not quite sure how to fix it atm but I'll take a look

PR #7081, opened 2023-11-21
New feature to create a copy of an issue

### LINKED ISSUE #7080: Add an option to create a copy of an existing issue
### Description

Often times I have an issue that I need to use as a temporary template ; I need to create the same issue for every quarter, or maybe for every team member, etc.

An option to take an existing issue and make a copy of it, would speed this up quite a bit! Then I would just make the copy, change anything I need to change, and hit save.

### Example URLs

https://github.com/internetarchive/openlibrary/issues/8544

cdrini:
Closes #7080 . Adds a new feature to create a copy of an existing issue.

## Test URLs

https://github.com/internetarchive/openlibrary/issues/8544

## Screenshot

image
Which then copies over the title, body, labels, assignees, milestones, and sends you to here:
image
to save your new issue.


fregante:
Please wait for the feature to be accepted first

Issue #7003, opened 2023-10-29
`mark-merge-commits-in-list`: No commits found, selector likely out of date


imranbarbhuiya:
### Description

When using the new commit page feature, I'm getting the below error. Since this is a very new feature so this type of bug is expected so I'm reporting it just to inform you

```
Error: No commits found, selector likely out of date
    at init (chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/assets/refined-github.js:7956:45)
    at chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/assets/refined-github.js:4216:28
    at chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/assets/refined-github.js:4212:64
    at Array.map (<anonymous>)
    at chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/assets/refined-github.js:4212:45
    at runFeature (chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/assets/refined-github.js:4213:10)
    at setupPageLoad (chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/assets/refined-github.js:4228:43)
```

### How to replicate the issue + URL

https://github.com/typed-ember/ember-cli-typescript/commits/master?after=5ff0c078a4274aeccaf83382c0d6b46323f57397+174

### Extension version

23.10.5

### Browser(s) used

chrome

PR #7087, opened 2023-11-22
`mark-merge-commits-in-list` - Support new commit list style

kovsu:
close: #7003 

## Test URLs

- isCommitList: https://github.com/typed-ember/ember-cli-typescript/commits/master?after=5ff0c078a4274aeccaf83382c0d6b46323f57397+174


kovsu:
after:
image

fregante:
Thank you! Can you also look at `dim-bots`?

- https://github.com/refined-github/refined-github/issues/7068

fregante (changes requested):
So I think here (and maybe in the other PR) it can be greatly simplified. Most likely you don't need a `isLegacy` check:

- query `'.markdown-title a, a.markdown-title'` in the same function. The right element will be selected in both cases.
- add `.listviewitem` to the `$$()` array

I think that's about it, here. In the other feature `isLegacy` is still needed for the `flex-wrap` addition I think

This change is particularly necessary because the old view isn't really legacy, I just noticed that `isPRCommits` still uses it: https://github.com/refined-github/refined-github/pull/7087/commits

source/features/mark-merge-commits-in-list.tsx
```diff
@@ -38,21 +38,37 @@ export function getCommitHash(commit: HTMLElement): string {
 	return $('a.markdown-title', commit)!.pathname.split('/').pop()!;
```
fregante, line 38:
```suggestion
	return $([
		'a.markdown-title', // Old view style (before November 2023)
		'.markdown-title a',
	], commit)!.pathname.split('/').pop()!;
```

You can use array to comment on each selector

kovsu:
> So I think here ~(and maybe in the other PR)~ it can be greatly simplified. Most likely you don't need a `isLegacy` check:
> 
> * query `'.markdown-title a, a.markdown-title'` in the same function. The right element will be selected in both cases.
> * add `.listviewitem` to the `$$()` array
> 
> I think that's about it, here. In the other feature `isLegacy` is still needed for the `flex-wrap` addition I think
> 
> This change is particularly necessary because the old view isn't really legacy, I just noticed that `isPRCommits` still uses it: [#7087 (commits)](https://github.com/refined-github/refined-github/pull/7087/commits)

I haven't used `select-dom` before. I will check it.

fregante (approved):

kovsu:
Should I update the other PR?

fregante:
Already done 😃

fregante:
## Updated screenshot

Screenshot 3

fregante:
If you'd like, it'd be good to fix the TODO in this file. You can see how it's used in other features

kovsu:
like this? @fregante
image

fregante:
Yes I think so ☺️ 

You can also drop the `duplicate` key

PR #7094, opened 2023-11-23
`list-prs-for-file` - Fix overlap with other components

fregante:
- Fixes https://github.com/refined-github/refined-github/issues/6845
- Closes #7074 

## Test URLs

https://github.com/hashicorp/vault/blob/main/go.mod

## Screenshot

Screenshot


fregante:
Merging this to fix the issue. It would still be good to find out why there's a fight at all  🥲

Issue #7108, opened 2023-11-24
Remove editor header in issues on archived repositories


tyrann0us:
### Description

On issues in archived repositories, GitHub displays the editor header ("Write" and "Preview"), which doesn't make sense.
<img width="983" alt="Bildschirmfoto 2023-11-24 um 08 46 12" src="https://github.com/refined-github/refined-github/assets/8144115/2817a3ac-1b7b-428b-bccf-f3f2d5ed9f6b">

What Refined GitHub does is remove the avatar in front of it. However, it would make more sense to remove the header completely.

### Example URLs

https://github.com/liip/bund_ds/issues/132#event-10996157695

fregante:
PR welcome. This can be a 3-line CSS-only feature: `features/clean-archived-issue.css` I think

a0m0rajab:
Hi @fregante I would like to work on this, but did not find the mentioned file, 
could it be a feature with the same logic as this one: https://github.com/refined-github/refined-github/blob/2fcbc2b28b6ca8c241c0411887e53a61b9e33e1e/source/features/archive-forks-link.tsx#L9 


To get this done we could remove the div with the next classes: `comment-form-head tabnav`

a0m0rajab:
This is the first time for me to contribute here, how would the extension detect the CSS in the file and understand if the issue is archived?

PR #7112, opened 2023-11-26
Hide inactive comment box tabs

a0m0rajab:
This PR adds a file to hide the comment in the archived issues, to have a better view. 
I checked this with PRs and issues that was not archived and did not see and breaking changes. 

close https://github.com/refined-github/refined-github/issues/7108

I could not test this because I was getting an error from nodejs: 

```bash
node: --import is not allowed in NODE_OPTIONS
ERROR: "build:webpack" exited with 9.
```
I will double check it later and see if I could get over that error. 

## Test URLs

https://github.com/liip/bund_ds/issues/132#event-10996157695

## Screenshot

| Before | After |
|--------|--------|
| image | image |

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -250,7 +250,6 @@ Thanks for contributing! 🦋🙌
 - [](# "jump-to-conversation-close-event") [Adds a link to jump to the latest close event of a conversation.](https://user-images.githubusercontent.com/16872793/177792713-64219754-f8df-4629-a9ec-33259307cfe7.gif)
 - [](# "close-as-unplanned") [Lets you "close issue as unplanned" in one click instead of three.](https://github-production-user-asset-6210df.s3.amazonaws.com/1402241/279745773-709cde60-c26a-4a0e-89e1-56444d25ebdf.png)
 - [](# "locked-issue") [Show a label on locked issues and PRs.](https://user-images.githubusercontent.com/1402241/283015579-0a04becc-9bff-4aef-8770-272d6804970b.png)
-- [Hides the inactive comment box tabs.](https://github.com/refined-github/refined-github/issues/7108)
 
 <!-- Refer to style guide above. Keep this message between sections. -->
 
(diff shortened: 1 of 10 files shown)


a0m0rajab:
Note: the lint is unsuccessful due to files I did not touch.

source/features/clean-archived-issue.css
```diff
@@ -0,0 +1,3 @@
+.comment-form-head.tabnav {
```
fregante, line 1: This selects all the navigation, it has no reference to archived issues

source/features/clean-archived-issue.css
```diff
@@ -0,0 +1,3 @@
...
+    visibility: hidden;
```
fregante, line 2: Use display: none

fregante (changes requested):

source/features/clean-archived-issue.css
a0m0rajab: Done.

source/features/clean-archived-issue.css
a0m0rajab: Thanks, I will check the MDN docs: https://developer.mozilla.org/en-US/docs/Web/CSS/:has

tyrann0us (commented):
I tested the CSS selector in different scenarios:
* archived repository -> editor header gone ✅
* locked issue -> editor header gone ✅
* locked issue with write permissions -> editor header present due to different HTML structure ✅

So, the feature works. However, it affects not only archived repositories but also locked issues of _non-archived repositories_. This is actually good, and I didn't have this in mind when creating #7108, but we should name the feature accordingly.

Alternatively, and I actually would prefer this, instead of adding a new CSS file, we could add the rule to https://github.com/refined-github/refined-github/blob/feat-hide-comment-area/source/features/locked-issue.css.

tyrann0us:
> it's easier to just keep separate files and name features.

Ok. So `clean-archived-issue.css` should be renamed to `clean-locked-issue.css` or similar. CC @a0m0rajab

readme.md
```diff
@@ -250,6 +250,7 @@ Thanks for contributing! 🦋🙌
 - [](# "jump-to-conversation-close-event") [Adds a link to jump to the latest close event of a conversation.](https://user-images.githubusercontent.com/16872793/177792713-64219754-f8df-4629-a9ec-33259307cfe7.gif)
 - [](# "close-as-unplanned") [Lets you "close issue as unplanned" in one click instead of three.](https://github-production-user-asset-6210df.s3.amazonaws.com/1402241/279745773-709cde60-c26a-4a0e-89e1-56444d25ebdf.png)
 - [](# "locked-issue") [Show a label on locked issues and PRs.](https://user-images.githubusercontent.com/1402241/283015579-0a04becc-9bff-4aef-8770-272d6804970b.png)
+- [Hides the inactive comment box tabs.](https://github.com/refined-github/refined-github/issues/7108)
```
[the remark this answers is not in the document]
fregante: Actually I don’t think this needs to be documented. It's super minor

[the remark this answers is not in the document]
kidonng: Ah yes, didn't notice this goes into `github-bugs`.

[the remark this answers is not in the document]
a0m0rajab: > Please follow [the description style guide](https://github.com/refined-github/refined-github/blob/461eb5cf467feaf240dcd87d78873c579245473f/readme.md?plain=1#L113-L138) and add a screenshot.

Did not notice that there is a description style guide, thank you for showing me that! 

> Actually I don’t think this needs to be documented. It's super minor

I will remove this part.

PR #7127, opened 2023-11-30
`visit-tag` - Enable on single releases

kidonng:
Mentioned in https://github.com/refined-github/refined-github/issues/7125#issuecomment-1833875365

## Test URLs

https://github.com/refined-github/refined-github/releases/tag/23.11.23

## Screenshot

image


fregante:
It'd be good to add that url to the test urls

PR #7128, opened 2023-11-30
Bold current pull request tab

kidonng:
The current repository tab is made bold, so should the current pull request tab.

image

Based on [my dogfooding](https://github.com/kidonng/cherry/commit/0517b4317825c407962c7c6eadb8bd05f032776b) this has remain unchanged for > 1.5 years :shrug: 

## Test URLs

Here

## Screenshot

<table>
<tr>
 <td>Before
 <td>

image

<tr>
 <td>After
 <td>

image

</table>

### DIFF
--- a/source/features/github-bugs.css
+++ b/source/features/github-bugs.css
@@ -54,3 +54,8 @@
 	pointer-events: none;
 	filter: grayscale(0.5);
 }
+
+/* Bold current pull request tab */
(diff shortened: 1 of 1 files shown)


fregante:
Yeah, also here they already dim the inactive tabs, so if we wanted to match the style, this would't be enough. And given that they're on two different levels, we shouldn't match it either.

Screenshot

PR #7129, opened 2023-12-01
Enable rgh-pr-template for all repos

LudvigHz:
Enable `rgh-pr-template` for all repos and rename to `pr-template`. 

Is there a reason why this feature was locked to refined-github only? It would be very nice if we could get this as a part of rgh on all repos :)

## Test URLs

https://github.com/refined-github/refined-github/compare/main...sandbox/keep-branch?quick_pull=1

## Screenshot

See #6947

### DIFF
--- a/source/features/pr-template.tsx
+++ b/source/features/pr-template.tsx
@@ -24,7 +24,6 @@ function init(signal: AbortSignal): void {
 
 void features.add(import.meta.url, {
 	asLongAs: [
-		isRefinedGitHubRepo,
 		pageDetect.isCompare,
 	],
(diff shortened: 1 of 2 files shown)


fregante:
Sorry, can't do. It doesn't match everyone's expectations. More often than not, that description is meant to be used and filled-in like a templated, not removed.

We tried this before, it's not possible.

fregante:
Some context:

- https://github.com/refined-github/refined-github/issues/3365

That feature actually tried to render the template, just to show that sometimes the requirements are the opposite of this

PR #7130, opened 2023-12-01
Meta: Support new screenshot URLs in readme parser

fregante:
For https://github.com/refined-github/refined-github/issues/6000#issuecomment-1714997576


build/__snapshots__/readme-parser.test.ts.snap
```diff
@@ -0,0 +1,1137 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
```
fregante, line 1: Snapshots are a bit annoying, but I think in the past we had at least one instance of nuking the options page due to some parser mistakes.

This can also help as to double-check what is actually being extracted from the readme. Currently the only way is to check the generated bundles or opening the options page (where screenshot URLs are not immediately visible for example)

On a related note, I wonder if we should export this list to a JSON file into the repo so that it can be used.

https://vitest.dev/api/expect.html#tomatchfilesnapshot

build/__snapshots__/imported-features.json
```diff
@@ -0,0 +1,193 @@
+[
+  "action-pr-link",
```
fregante, line 2: This is great. We now have a static JSON file with a list of features.

This could enable things like #4493 😜 

It can potentially also be transformed to a raw-text list

PR #7131, opened 2023-12-02
Lint

fregante:


source/github-helpers/index.ts
REBELminds249, line 1: Fuck

REBELminds249 (approved):

source/features/embed-gist-inline.tsx
```diff
@@ -87,3 +88,11 @@ void features.add(import.meta.url, {
 	],
 	init,
 });
+
+/*
+
+Test URLs
+
+https://github.com/refined-github/sandbox/issues/77
```
fregante, line 96: The new selector was also tested here

fregante (commented):
I wrote and then reverted a bunch of commits that will alter become their own PRs

PR #7152, opened 2023-12-11
Meta: Update Octicons; Use SVGR for imports

### LINKED ISSUE #6878: Review `@primer/octicons-react` usage, update/replace dependency
Our `@primer/octicons-react` package cannot be updated:

- v17.11 breaks dom-chef
- [v19 uses forwardRef](https://github.com/primer/octicons/releases/tag/v19.0.0)

There were a couple of attempts at fixing this:

- https://github.com/refined-github/refined-github/pull/4044

Probably `jsx-dom` is the best transition we can do right now, but also it's probably not working with v19.

So I think the easiest way forward would be to:

1. import the raw SVG
2. have webpack turn it into JSX-compatible imports ("svgr")

This should be reasonably easy to implement and likely the most stable and lightweight solution (octicons-react has more code than we need, at this point)

fregante:
- Fixes #6878 
- Extracted from #7131

### DIFF
--- a/build/new-feature.sh
+++ b/build/new-feature.sh
@@ -10,7 +10,7 @@ then
 fi
 
 echo "import React from 'react';
-import BugIcon from '@primer/octicons-react/build/svg/bug-16.svg';
+import BugIcon from '@primer/octicons/build/svg/bug-16.svg';
 import * as pageDetect from 'github-url-detection';
 
 import features from '../feature-manager.js';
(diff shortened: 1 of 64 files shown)


fregante:
Blocked by:

- https://github.com/primer/octicons/issues/997

fregante:
I don’t think that bug is going to be resolved anytime soon, so I should probably run through the extension and see if any obvious issue appears, and then just merge it. 

If bothersome I'll just republish it and add an update script. I've been doing this for a few bothersome packages like `web-ext-submit`

PR #7156, opened 2023-12-13
`hide-yarn-berry-cache` - New feature

effx13:
While using a [Yarn berry Zero-installs](https://yarnpkg.com/features/caching#zero-installs) feature, yarn berry creates a huge load of binary zip files.
This feature is very useful for developers, but yarn berry makes it difficult to see and makes me angry when I see PRs.
For instance, if I accidentally delete node_modules and attempts to reinstall, I would have to review hundreds of files again.

Therefore, I created a feature to collapse the `.yarn/cache` folder by default and reduce its size. 

## Test URLs

on PR:
- https://github.com/refined-github/sandbox/pull/78/files

on Compare:
- https://github.com/refined-github/sandbox/compare/default-a...effx13:sandbox:test-yarn-berry

### DIFF
--- a/build/__snapshots__/features-meta.json
+++ b/build/__snapshots__/features-meta.json
@@ -310,6 +310,11 @@
 		"screenshot": "https://github-production-user-asset-6210df.s3.amazonaws.com/1402241/263195425-85cf0951-c6ed-45fe-8cfc-e447e3ed2a25.png"
 	},
 	{
+		"id": "hide-yarn-berry-cache",
+		"description": "Always collapse a lot of <code>.yarn/cache</code> diff in PRs.",
+		"screenshot": "https://github.com/refined-github/refined-github/assets/29156835/9432f9b8-1406-434a-8752-4779ceac8884"
+	},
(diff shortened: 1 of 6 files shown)


fregante:
See https://github.com/refined-github/refined-github/pull/1085

Thank you for the PR, but it's best to discuss features before writing code.

fregante:
By the way that cache folder should not be committed to the repo.

effx13:
@fregante 
[Yarn Berry](https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored) introduces saving cache folder to git as a method, and I don't think this is a wrong way to use it.
However, I should have looked more closely at the previous discussion, but I wrote the code in a hurry.

As in issue https://github.com/refined-github/refined-github/issues/7116, I aware that github has performance issue when there are a large number of PR files.
So, I think it would be nice to have a feature to filter specific files. 
It's already on Github, but it doesn't apply globally. If this feature hasn't been discussed before, can I create a new Issue?

PR #7161, opened 2023-12-15
Use `octicons-plain-react`

### LINKED ISSUE #6878: Review `@primer/octicons-react` usage, update/replace dependency

fregante:
- Replaces and closes https://github.com/refined-github/refined-github/pull/7152
- Works arounds https://github.com/primer/octicons/issues/997
- Fixes #6878 
- Uses https://github.com/fregante/octicons-plain-react

It doesn't look like Octicons is updated that often so it's best not to be blocked by them. It also has [a few annoyances](https://github.com/primer/octicons/issues?q=author%3Afregante) that we can just fix in our repo instead.

This has the nice benefit of moving the SVG->React transformation and complexity away from our build.

## Screenshot

<table>
<tr>
	<th>Before
	<th>After
<tr>
	<td>Screenshot 5
	<td>Screenshot 4
</table>

The new SVGs have the `class` before the `xmlns` attribute due to

https://github.com/fregante/octicons-plain-react/blob/13b7daf92581954247ee68a292f83edffae7b680/scripts/prepack.sh#L11

Screenshot 3


package.json
ZaLiTHkA, line 1: I'm having trouble getting this project running in my local dev environment, with "TS2307: Cannot find module" errors for all the `octicons-plain-react` imports.. @fregante, would this be a repo or dev env issue?

fregante: Are you using the right Node version? I think it should be node 20 now

ZaLiTHkA: just tested with NodeJS 20.13.1, built the project without any issues. sorry I didn't think to try that first, definitely not a repo issue then. thank you. :)

PR #7166, opened 2023-12-16
Meta: Fix `selector` tests

fregante:
While working on https://github.com/refined-github/refined-github/pull/7165 I realized that `selector.test.ts` would _always_ pass. Great.

Now the issue is that these selectors don't actually exist on logged-out pages. _What do?_


fregante:
Fixed in https://github.com/refined-github/refined-github/pull/7165 itself

PR #7180, opened 2023-12-21
`rgh-dim-commits` - Add support for new commit view

### LINKED ISSUE #7179: `rgh-dim-commits` don't work on new commits view
### Description

Feature doesn't work with new commits ui.

image

### Extension version

23.12.17

### Browser(s) used

Google Chrome & Firefox

Katsute:
Fix #7179

## Test URLs

https://github.com/refined-github/refined-github/commits/98fb24e53704b436d68f254429885fb7bef09541/source/features/bypass-checks.tsx

## Screenshot

image

### DIFF
--- a/source/features/rgh-dim-commits.tsx
+++ b/source/features/rgh-dim-commits.tsx
@@ -3,9 +3,17 @@ import * as pageDetect from 'github-url-detection';
 import {isRefinedGitHubRepo} from '../github-helpers/index.js';
 import features from '../feature-manager.js';
 import observe from '../helpers/selector-observer.js';
+import {isHasSelectorSupported} from '../helpers/select-has.js';
 
 // Source: https://github.com/fregante/release-with-changelog/blob/779fd5e658f82e5b11b1c0a352a6838d3bd8f67f/generate-release-notes.js#L6
 const excludePreset = /^bump |^meta|^document|^lint|^refactor|readme|dependencies|^v?\d+\.\d+\.\d+/i;
+
+// TODO: Remove in June 2024
+function dimOld(commitTitle: HTMLElement): void {
+	if (excludePreset.test(commitTitle.textContent.trim())) {
+		commitTitle.closest('.js-commits-list-item')!.style.opacity = '50%';
+	}
+}
 
 function dim(commitTitle: HTMLElement): void {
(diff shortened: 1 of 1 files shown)


source/features/rgh-dim-commits.tsx
```diff
@@ -3,25 +3,23 @@ import * as pageDetect from 'github-url-detection';
...
 function dim(commitTitle: HTMLElement): void {
 	if (excludePreset.test(commitTitle.textContent.trim())) {
-		commitTitle.closest('.js-commits-list-item')!.style.opacity = '50%';
+		commitTitle.closest('[data-testid="commit-row-item"]')!.style.opacity = '50%';
 	}
 }
 
 function init(signal: AbortSignal): void {
-	observe('.js-commits-list-item p:has(> .js-navigation-open)', dim, {signal});
+	observe('[data-testid="commit-row-item"] [data-testid="listview-item-title-container"] span', dim, {signal});
```
fregante, line 17: We can preserve GHE compatibility by keeping both of these selectors
```suggestion
	observe([
		'blah blah', // TODO: Remove in June 2024
		'[data-testid="commit-row-item"] [data-testid="listview-item-title-container"] span'
	], dim, {signal});
```

Also in `[data-testid="commit-row-item"] [data-testid="listview-item-title-container"]` only the latter is necessary, probably.

That final `span` however seems a bit broad. Maybe we can use `span:first-child`?

Katsute: Restored original behavior:

image

Katsute: I think `[data-testid="listview-item-title-container"] .markdown-title span` is sufficient for the span selector

fregante: Note that I suggest to use an array, not to duplicate the observer and handler. We should only do that when the handler differs substantially. In the handler `.closest()` can also take an array of selectors

Katsute: neat. didn't know you could do that

fregante: Technically it just accepts string, but this is JavaScript, sooooooo… I altered the types in global.d.ts 😃

PR #7189, opened 2023-12-27
Automatically fit comment preview height

pgimalac:
Fix https://github.com/refined-github/refined-github/issues/7187.

Rendered markdown div automatically fit their content, but for some reason Github explicitly sets the `min-height` style property to the height of the input textarea.
Simply overriding the `min-height` to `0` fixes this bug.

Affects any div that renders markdown.

## Screenshot

| Input | Original | Fixed |
|--------|--------|--------|
| image | image | image |

### DIFF
--- a/build/verify-test-urls.sh
+++ b/build/verify-test-urls.sh
@@ -36,7 +36,7 @@ echo
 
 for FILE in "$@"; do
 	# skip github-bugs.css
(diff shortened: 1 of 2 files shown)


fregante (approved):
Thank you! I think we can set it to something like 2em to make sure it's never collapsed to 0 (which would probably look broken)

You can ignore the CI failure

pgimalac:
Done !

PR #7194, opened 2023-12-30
`show-associated-branch-prs-on-fork` - Support new page style

Machine-Maker:
Closes https://github.com/refined-github/refined-github/issues/7193

## Test URLs

https://github.com/bfred-it-org/github-sandbox/branches

## Screenshot
With this fix
image
Without
image

### DIFF
--- a/source/features/show-associated-branch-prs-on-fork.css
+++ b/source/features/show-associated-branch-prs-on-fork.css
@@ -1,32 +1,24 @@
+/* Styles copied from React component */
 .rgh-pr-box {
 	display: flex;
-	-mox-box-align: center;
 	align-items: center;
 	height: 32px;
 }
 
 .rgh-pr-link {
 	color: var(--color-fg-muted);
-	/* color: rgb(125 133 144); */
 	text-decoration: none;
 	font-weight: normal;
-	-mox-box-align: center;
 	align-items: center;
 }
 
 .rgh-pr-link .rgh-pr-label {
-	-moz-box-align: center;
 	align-items: center;
-	background-color: transparent;
-	border-width: 1px;
+	border: solid 1px var(--color-border-default);
 	border-radius: 999px;
-	border-style: solid;
 	display: inline-flex;
-	font-weight: 600;
-	font-size: 12px;
-	line-height: 1;
+	font: 600 12px/1 inherit;
 	white-space: nowrap;
-	border-color: var(--color-border-default);
 	height: 20px;
 	padding: 0 7px;
 }
@@ -35,24 +27,7 @@
 	background-color: var(--color-canvas-subtle);
 }
 
-.rgh-pr-link .rgh-pr-label .rgh-pr-open {
-	color: var(--color-open-fg);
-}
-
-.rgh-pr-link .rgh-pr-label .rgh-pr-closed {
-	color: var(--color-closed-fg);
-}
-
-.rgh-pr-link .rgh-pr-label .rgh-pr-draft {
-	color: var(--color-fg-muted);
-}
-
-.rgh-pr-link .rgh-pr-label .rgh-pr-merged {
-	color: var(--color-done-fg);
-}
-
 .rgh-pr-link .rgh-pr-text {
 	margin-left: 4px;
-	/* color: rgb(125 133 144); */
 	color: var(--color-fg-muted);
 }
--- a/source/features/show-associated-branch-prs-on-fork.css
+++ b/source/features/show-associated-branch-prs-on-fork.css
@@ -1,0 +1,52 @@
+.pr-box {
+	display: flex;
+	-mox-box-align: center;
+	align-items: center;
+	height: 32px;
+}
+
+.pr-link {
+	color: rgb(125 133 144);
+	text-decoration: none;
+	font-weight: normal;
+	-mox-box-align: center;
+	align-items: center;
+}
+
+.pr-link .pr-label {
+	-moz-box-align: center;
+	align-items: center;
+	background-color: transparent;
+	border-width: 1px;
+	border-radius: 999px;
+	border-style: solid;
+	display: inline-flex;
+	font-weight: 600;
+	font-size: 12px;
+	line-height: 1;
+	white-space: nowrap;
+	border-color: rgb(48 54 61);
+	height: 20px;
+	padding: 0 7px;
+}
+
+.pr-link .pr-label .pr-open {
+	color: var(--color-open-fg);
+}
+
+.pr-link .pr-label .pr-closed {
+	color: var(--color-closed-fg);
+}
+
+.pr-link .pr-label .pr-draft {
+	color: var(--color-fg-muted);
+}
+
+.pr-link .pr-label .pr-merged {
+	color: var(--color-done-fg);
+}
+
+.pr-link .pr-text {
+	margin-left: 4px;
+	color: rgb(125 133 144);
+}
--- a/source/features/show-associated-branch-prs-on-fork.css
+++ b/source/features/show-associated-branch-prs-on-fork.css
@@ -4,7 +4,16 @@
 	align-items: center;
 	flex-wrap: wrap;
 	gap: 4px;
-	height: auto; /* Must override the hardcoded height */
+}
+
+.rgh-pr-cell {
+	flex-wrap: wrap;
+	gap: 4px;
+
+	&	> [class*=Box] {
+		/* Drop forced height on pre-existing cell */
+		height: auto;
+	}
 }
 
 .rgh-pr-link {
@@ -13,7 +22,7 @@
 	font-weight: normal;
 	align-items: center;
 	border: solid 1px var(--color-border-default);
-	text-decoration: none;
+	text-decoration: none !important;
 	border-radius: 999px;
 	display: inline-flex;
 	font: 600 12px/1 inherit;
@@ -24,9 +33,4 @@
 
 .rgh-pr-link:hover {
 	background-color: var(--color-canvas-subtle);
-}
-
-.rgh-pr-text {
-	margin-left: 4px;
-	color: var(--color-fg-muted);
 }
--- a/source/features/show-associated-branch-prs-on-fork.css
+++ b/source/features/show-associated-branch-prs-on-fork.css
@@ -1,8 +1,10 @@
-/* Styles copied from React component */
-.rgh-pr-box {
+/* Styles mostly copied from React component */
+:root .rgh-pr-box {
 	display: flex;
 	align-items: center;
-	height: 32px;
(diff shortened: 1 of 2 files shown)


Machine-Maker:
The CI failure for test urls... I don't see those in other `.css` files that are attached to a tsx file. Can I just ignore that?

source/features/show-associated-branch-prs-on-fork.css
Machine-Maker, line 1: These replicate the style given by the obfuscated(?) class names? I couldn't just use those class names because sometimes they weren't included in the page's inline styles.

fregante: Yeah it's going to be tough with those styles, but GitHub loads [primer styles](https://primer.style) too, which include a lot of utility classes and other styles. In Chrome's dev tools, I suggest searching something like `align-items: center;` and finding the related utility class for that. You might get lucky and find a pre-existing `.label` class that includes all you need

source/features/show-associated-branch-prs-on-fork.css
```diff
@@ -0,0 +1,52 @@
+.pr-box {
```
fregante, line 1: All of our classes start with `.rgh`

source/features/show-associated-branch-prs-on-fork.css
```diff
@@ -0,0 +1,52 @@
...
+	background-color: transparent;
+	border-width: 1px;
+	border-radius: 999px;
+	border-style: solid;
+	display: inline-flex;
+	font-weight: 600;
+	font-size: 12px;
+	line-height: 1;
+	white-space: nowrap;
+	border-color: rgb(48 54 61);
```
fregante, line 28: GitHub already loads plenty of CSS variables, we should never inline colors, because they depend on the style. In this case the color is probably named something like `--color-border-muted`

source/features/show-associated-branch-prs-on-fork.tsx
```diff
@@ -75,6 +76,23 @@ function addAssociatedPRLabel(branchCompareLink: Element, prInfo: PullRequest):
 	);
 }
 
+function addAssociatedPRLabelNew(parent: Element, prInfo: PullRequest): void {
+	const StateIcon = stateIcon[prInfo.state];
+
+	parent.append(
+		<div className="pr-box">
+			<a href={prInfo.url} target="_blank" data-hovercard-url={prInfo.url + '/hovercard'} aria-label={`Link to the ${prInfo.isDraft ? 'draft ' : ''}pull request #${prInfo.number}`} className="pr-link" rel="noreferrer">
```
fregante, line 84: Long attributes list should be wrapped, like:

```jsx
<a
	attr=1
	name=2
>
	<span/>
</a>
```

source/features/show-associated-branch-prs-on-fork.tsx
```diff
@@ -84,8 +102,22 @@ async function addLink(branchCompareLink: Element): Promise<void> {
 	}
 }
 
+async function addLinkNew(prAnchorElement: HTMLAnchorElement): Promise<void> {
+	const prs = await pullRequestsAssociatedWithBranch.get();
+	const tableRow = prAnchorElement.closest('.TableRow') as HTMLTableRowElement;
+	const branchName = (tableRow.firstChild! as HTMLTableCellElement).querySelector('div[title]')!.getAttribute('title')!;
+	const prInfo = prs[branchName];
```
fregante, line 109: You probably want to select the branch name in `observe` and then get only the `prCell` from there, so that we limit the amount of DOM traversing.

fregante:
> Can I just ignore that?

Yes.

Eventually we'll update that checker:

- https://github.com/refined-github/refined-github/issues/7132#issuecomment-1849218717

Machine-Maker:
So I fixed most of those things. I could not find any such utility classes. I looked at the primer source for Label and it just hardcodes all the style. 
image

I also couldn't find an exact match for rgb(125, 133, 144) which is why its still in the css file, just commented out. The github source shows that it should be something like 
image
but the `--color-fg-muted` does not match the color I saw in the inline style, so I'm not sure what to do there.

source/features/show-associated-branch-prs-on-fork.css
```diff
@@ -0,0 +1,58 @@
+.rgh-pr-box {
```
fregante, line 1: Let's add this header to the file

```css
/* Styles copied from React component */
```

source/features/show-associated-branch-prs-on-fork.css
```diff
@@ -0,0 +1,58 @@
...
+	font-weight: normal;
+	-mox-box-align: center;
+	align-items: center;
+}
+
+.rgh-pr-link .rgh-pr-label {
+	-moz-box-align: center;
+	align-items: center;
+	background-color: transparent;
+	border-width: 1px;
```
fregante, line 21: Let's use shorthand properties, like

```css
	border: solid 1px var(--color-border-default);
	font: 600 12px/1 inherit;
```

source/features/show-associated-branch-prs-on-fork.css
```diff
@@ -0,0 +1,58 @@
...
+	text-decoration: none;
+	font-weight: normal;
+	-mox-box-align: center;
+	align-items: center;
+}
+
+.rgh-pr-link .rgh-pr-label {
+	-moz-box-align: center;
+	align-items: center;
+	background-color: transparent;
```
fregante, line 20: I don’t think we need this?

fregante (changes requested):

source/features/show-associated-branch-prs-on-fork.tsx
```diff
@@ -75,6 +76,29 @@ function addAssociatedPRLabel(branchCompareLink: Element, prInfo: PullRequest):
...
+			<a
+				href={prInfo.url}
+				target="_blank"
+				data-hovercard-url={prInfo.url + '/hovercard'}
+				aria-label={`Link to the ${prInfo.isDraft ? 'draft ' : ''}pull request #${prInfo.number}`}
+				className="rgh-pr-link" rel="noreferrer"
+			>
+				<span className="rgh-pr-label">
+					<div data-testid="draft-pull-request-icon" className="rgh-pr-box">
+						<StateIcon width={14} height={14} className={'rgh-pr-' + prInfo.state.toLowerCase()}/>
```
fregante, line 93: GitHub already has dedicated color classes:

```
.draft.octicon {
    color: var(--fgColor-muted, var(--color-fg-muted))
}

.closed.octicon,.reverted.octicon {
    color: var(--fgColor-closed, var(--color-closed-fg))
}

.open.octicon {
    color: var(--fgColor-open, var(--color-open-fg))
}

.closed.octicon.octicon-issue-closed,.merged.octicon {
    color: var(--fgColor-done, var(--color-done-fg))
}
```

so I think this will already work:

```suggestion
						<StateIcon width={14} height={14} className={prInfo.state.toLowerCase()}/>
```

plus some adjustments for `draft`

source/features/show-associated-branch-prs-on-fork.css
```diff
@@ -0,0 +1,58 @@
...
+
+.rgh-pr-link .rgh-pr-label .rgh-pr-merged {
+	color: var(--color-done-fg);
+}
+
+.rgh-pr-link .rgh-pr-text {
+	margin-left: 4px;
+	/* color: rgb(125 133 144); */
+	color: var(--color-fg-muted);
+}
```
fregante, line 58: Use the classes `color-fg-muted ml-1` on the element itself instead

source/features/show-associated-branch-prs-on-fork.tsx
```diff
@@ -75,6 +76,29 @@ function addAssociatedPRLabel(branchCompareLink: Element, prInfo: PullRequest):
 	);
 }
 
+function addAssociatedPRLabelNew(parent: Element, prInfo: PullRequest): void {
+	const StateIcon = stateIcon[prInfo.state];
+
+	parent.append(
+		<div className="rgh-pr-box">
+			<a
```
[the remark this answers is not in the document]
fregante: I see why it's there, I think the `a` should just be moved to the existing `.Box-` element instead

source/features/show-associated-branch-prs-on-fork.tsx
```diff
@@ -75,6 +76,29 @@ function addAssociatedPRLabel(branchCompareLink: Element, prInfo: PullRequest):
...
+				target="_blank"
+				data-hovercard-url={prInfo.url + '/hovercard'}
+				aria-label={`Link to the ${prInfo.isDraft ? 'draft ' : ''}pull request #${prInfo.number}`}
+				className="rgh-pr-link" rel="noreferrer"
+			>
+				<span className="rgh-pr-label">
+					<div data-testid="draft-pull-request-icon" className="rgh-pr-box">
+						<StateIcon width={14} height={14} className={'rgh-pr-' + prInfo.state.toLowerCase()}/>
+					</div>
+					<span className="rgh-pr-text">#{prInfo.number}</span>
```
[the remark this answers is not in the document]
fregante: The table style is craptastic, it's inflexible:

Screenshot 14

Maybe we can only use an icon instead, like
```suggestion
					<span className="rgh-pr-text"><ForkIcon/>{prInfo.number}</span>
```

fregante:
Sorry for the million notes 🥹 🙏

source/features/show-associated-branch-prs-on-fork.tsx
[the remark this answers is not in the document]
Machine-Maker: I'm not sure I'm following, there isn't an existing Box element, the table cell is empty.

Machine-Maker:
Ok, addressed all the reviews (except the one I commented on) and I updated the image in the PR description of what it looks like now.

source/features/show-associated-branch-prs-on-fork.tsx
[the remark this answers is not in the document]
fregante: There is, but only if there's another PR there too. See my example link

fregante (approved):
Thank you! Can you post a screenshot for future reference?

Machine-Maker:
> Thank you! Can you post a screenshot for future reference?

I did, I updated the screenshot in the pr description.

fregante:
Screenshot


I'm seeing duplicates for some reason, but I think it's due to a malfunctioning of the selector observer, it's detecting two animations for the same element 🤷‍♂️ 

I'll fix it via memoize for now

fregante:
I tried reusing the existing box if available but GitHub's code removes it in some cases:

all

So I just adjusted the style to wrap and space the items the best it can

PR #7198, opened 2023-12-31
`use-first-commit-message-for-new-prs` - Do not replace title on page reload

pgimalac:
Closes https://github.com/refined-github/refined-github/issues/7191.

Do not set the title to the first commit message on page reload.
It avoids overwriting user input.

The feature still works on page `load`, but not on `reload`, github just restores whatever input was before, whether the first commit title or user input.

## Test URLs
https://github.com/pgimalac/refined-github/compare/main...pgimalac:refined-github:pgimalac/fit-rendered-markdown?expand=1

## Screenshot

https://github.com/refined-github/refined-github/assets/23154723/8d70e4e2-560c-4363-bc19-d5bd8d0ca1d3

### DIFF
--- a/source/features/use-first-commit-message-for-new-prs.tsx
+++ b/source/features/use-first-commit-message-for-new-prs.tsx
@@ -37,6 +37,14 @@ function getFirstCommit(): {title: string; body: string | undefined} {
 }
 
 async function init(): Promise<void | false> {
+	if (window.performance.navigation?.type === 1 ||
+		window.performance.getEntriesByType('navigation').some(
+			(nav) => (nav as PerformanceNavigationTiming).type === 'reload')
+		) {
+		// if the page is reloaded, let github restore the previous title
+		return false;
+	}
+
 	const requestedContent = new URL(location.href).searchParams;
 	const commitCountIcon = await elementReady('div.Box.mb-3 .octicon-git-commit');
 	const commitCount = commitCountIcon?.nextElementSibling;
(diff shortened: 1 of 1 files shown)


fregante:
I'd be more comfortable targeting this specific GitHub behavior rather than guess the user action. It seems to be easy at this point, assuming that our code runs before Github's `session:resume` event:

```js
function into () {
  if (sessionStorage.getItem(`session-resume:/${getRepo()?.nameWithOwner}/compare`)) {
	  return false;
  }
  
  // …
}
```

fregante (changes requested):
I'd be more comfortable targeting this specific GitHub behavior rather than guess the user action. It seems to be easy at this point, assuming that our code runs before Github's `session:resume` event:

```js
function into () {
  if (sessionStorage.getItem(`session-resume:/${getRepo()?.nameWithOwner}/compare`)) {
	  return false;
  }
  
  // …
}
```

pgimalac:
Hey @fregante, sorry for the delay, I gave it a try but it doesn't work, the code runs before the field is reloaded (the title is the branch name, which is the default) but there is no `session-resume` in the `sessionStorage`...

fregante:
session:resume is an event on window. Look into this repo to find how to use it

pgimalac:
> session:resume is an event on window. Either way we don't need to do anything about it

I'm not sure I understand, do you want me to do anything about this PR then ?

fregante:
> do you want me to do anything about this PR then ?

I posted the piece of code I'd like to see inside `init`. Does that fix the issue? In my manual tests it worked

pgimalac:
> Hey @fregante, sorry for the delay, I gave it a try but it doesn't work, the code runs before the field is reloaded (the title is the branch name, which is the default) but there is no session-resume in the sessionStorage...

As I said in my previous message I tried replacing the block I added with the one you wrote using `session-resume`, but it didn't work. I printed all `sessionStorage` key/value and none were `session-resume`...
It worked for you ?

PR #7201, opened 2024-01-02
`use-first-commit-message-for-new-prs` - fix for different page layout

pgimalac:
Fix https://github.com/refined-github/refined-github/issues/7200.

Handle different page layout in `use-first-commit-message-for-new-prs`.

## Notes
With that layout, commits are loaded after the page itself, so I had to add a wait for the element.

I went with minimal changes, let me know if you want me to refactor something.

I used a max delay of 3 seconds to wait for the commit list to load as I considered after a few seconds the user would likely have started to write the title.

## Test URLs
https://github.com/refined-github/refined-github/compare/refined-github:refined-github:esbuild-2...pgimalac:refined-github:pgimalac/fit-rendered-markdown?expand=1

## Screenshot

<table>
<tr>
 <td> Before (Github uses branch name)
 <td> After (feature sets first commit message as title)
<tr>
 <td> <video src="https://github.com/refined-github/refined-github/assets/23154723/225ad455-8100-4b16-ac7b-2109af69b8e0">
 <td> <video src="https://github.com/refined-github/refined-github/assets/23154723/2177898f-76a1-4604-8355-c59762399b2c">
</table>

### DIFF
--- a/source/features/use-first-commit-message-for-new-prs.tsx
+++ b/source/features/use-first-commit-message-for-new-prs.tsx
@@ -36,11 +36,31 @@ function getFirstCommit(): {title: string; body: string | undefined} {
 	return {title, body};
 }
 
+async function getCommitCount(): Promise<number> {
+	const commitCountIcon = await elementReady('div.Box.mb-3 .octicon-git-commit');
+	let commitCount = commitCountIcon?.nextElementSibling;
+	if (commitCount) {
+		return looseParseInt(commitCount);
+	}
+
+	const commitCountTabnav = await elementReady('div.tabnav .octicon-git-commit');
+	commitCount = commitCountTabnav?.nextElementSibling;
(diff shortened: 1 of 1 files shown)


fregante (changes requested):
Thank you for the PR! Actually we have an utility that simplifies this greatly. You just need this instead of `elementReady`:

```js
observe([
	'selector commit title in the regular layout',
	'selector commit title in the other layout'
], useCommitTitle, {signal})
```

In reality this selector already works for both though `.js-commits-list-item p`.

Then there are two issues:

- there are many such items on the page, `observe` will find them all, so you probably have to add `.TimelineItem:first-child .js-commits-list-item:first-child`
- you have to select the counter, but this time you can just use `$` instead (`select-dom`)

fregante:
Ping if you have time

PR #7212, opened 2024-01-10
Fix flacky selectors tests

fregante:
The tests are failing at random, only on CI. This makes them optional

Context:

- https://github.com/vitest-dev/vitest/issues/4920


fregante:
Turns out it's not that easy. In other places we still have an uncustomized `npm test` https://github.com/refined-github/refined-github/actions/runs/7472345016/job/20334329858

Hopefully the suggestion gets accepted, but in the meanwhile I'll disable testing just for the release.

PR #7214, opened 2024-01-11
`table-input`/`collapsible-content-button` - Fix overflow

fregante:
- Fixes https://github.com/refined-github/refined-github/pull/7211#issuecomment-1885863318

## Test URLs

Here

## Before

Screenshot

## After

Screenshot 1

### DIFF
--- a/source/features/table-input.tsx
+++ b/source/features/table-input.tsx
@@ -65,7 +65,11 @@ function add(anchor: HTMLElement): void {
 					<TableIcon/>
 				</div>
 			</summary>
-			<details-menu className="select-menu-modal position-absolute left-0 hx_rsm-modal rgh-table-input" role="menu">
+			<details-menu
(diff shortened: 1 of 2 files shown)


source/features/table-input.tsx
```diff
@@ -65,7 +65,11 @@ function add(anchor: HTMLElement): void {
 					<TableIcon/>
 				</div>
 			</summary>
-			<details-menu className="select-menu-modal position-absolute left-0 hx_rsm-modal rgh-table-input" role="menu">
+			<details-menu
+				className="select-menu-modal position-absolute left-0 hx_rsm-modal rgh-table-input"
+				role="menu"
+				data-targets="action-bar.items" // Enables automatic hiding when it doesn't fit
```
ForsakenHarmony, line 71: the `<details-menu />` is always below the first button, this makes it impossible to ever open the popup, it also never hides the button itself

fregante: PR welcome

PR #7247, opened 2024-02-04
`quick-reactions` - New feature

brumm:
Optimistically applies reactions instead of waiting for a slow network request to finish.

This should work anywhere you can react to something on GitHub. Reactions are added, updated and removed immediately. New reactions are inserted in the correct place and will not jump around when the new markup comes in.

## Test URLs

- regular comment https://github.com/brumm/tako/issues/27#issuecomment-617314195
- discussion comment https://github.com/vercel/next.js/discussions/40684#discussioncomment-3689167

## Screenshot

CleanShot 2024-02-04 at 22 28 20


fregante:
Love this. I'm concerned that it will break quickly and painfully, but if it's solid enough I'll just merge it anyway, because we can deal with that via hotfixes. 

The concern at the moment is that Refined GitHub is seeing some serious slowdowns in issues/PRs at the moment so I might not be able to merge this or add new features for a while :(

fregante:
It's not an issue with this PR, but rather with Refined GitHub itself. I can merge it once I figure out the root cause of this:

- https://github.com/refined-github/refined-github/issues/7176

because those features also use the selector observer, so I'm afraid that that's what's causing the slow-downs.

fregante (approved):

build/__snapshots__/features-meta.json
```diff
@@ -669,6 +669,11 @@
 		"description": "Adds a link to create issues from anywhere in a repository.",
 		"screenshot": "https://github-production-user-asset-6210df.s3.amazonaws.com/1402241/274816033-820ec518-049d-4248-9f8a-27b9423e350b.png"
 	},
+	{
+		"id": "quick-reactions",
+		"description": "Shows a reaction optimistically",
+		"screenshot": "https://user-images.githubusercontent.com/1402241/236628453-8b646178-b838-44a3-9541-0a9b5f54a84a.png"
```
bsrz, line 675: I'm just commenting for an issue I'm opening.
Will edit comment with issue number.

Codex-jen:
<link href="https://cdn.jsdelivr.net/npm/tailwindcss@2.2.19/dist/tailwind.min.css" rel="stylesheet"><p>Review the cover letter with the applied color scheme and refine the color scheme as needed.</p>

source/features/quick-reactions.tsx
```diff
@@ -0,0 +1,98 @@
+import delegate, {DelegateEvent} from 'delegate-it';
+import React from 'dom-chef';
+import * as pageDetect from 'github-url-detection';
+
+import features from '../feature-manager.js';
+
+const emojiSortOrder = ['THUMBS_UP', 'THUMBS_DOWN', 'LAUGH', 'HOORAY', 'CONFUSED', 'HEART', 'ROCKET', 'EYES'];
+
+function optimisticReactionFromPost(event: DelegateEvent<MouseEvent>): void {
+	const reaction = event.delegateTarget as HTMLButtonElement;
```
fregante, line 10: IIRC you can just pass the type here, no need to assert it. Refer to its docs if it fails, also fix it elsewhere

```suggestion
function optimisticReactionFromPost(event: DelegateEvent<MouseEvent, HTMLButtonElement>): void {
```

source/features/quick-reactions.tsx
```diff
@@ -0,0 +1,98 @@
...
+	updateReaction(reaction);
+}
+
+function optimisticReactionFromMenu(event: DelegateEvent<MouseEvent>): void {
+	const reactionButton = event.delegateTarget as HTMLButtonElement;
+	const reactionsMenu = reactionButton.closest('reactions-menu')!;
+	const details = reactionsMenu.querySelector('details');
```
fregante, line 17: Can you replace every querySelector usage with `$` from `select-dom`? The only exception is when the line has multiple traversing, e.g. `x.closest().querySelector('a')` is still more readable than `$('a', x.closest())`

source/features/quick-reactions.tsx
```diff
@@ -0,0 +1,98 @@
...
+	details!.open = false; // Close reactions menu immediately
+
+	const reactionsContainer = reactionsMenu.nextElementSibling!.querySelector('.js-comment-reactions-options')!;
```
fregante, line 20: Positional traversing makes the code a little more brittle because a simple `span` in between will break it. If possible, select the parent element instead (e.g. `menu.parentElement.querySelector`) or just change what `reactionsMenu` to select its parent instead

source/features/quick-reactions.tsx
```diff
@@ -0,0 +1,98 @@
...
+	const reactionButtonValue = reactionButton.getAttribute('value')!;
```
fregante, line 21: Are you looking for `reactionButton.value`?

source/features/quick-reactions.tsx
```diff
@@ -0,0 +1,98 @@
...
+			referenceNode,
+		);
+	}
+}
+
+function updateReaction(buttonElement: HTMLButtonElement): void {
+	const countElement = buttonElement.querySelector(
+		'.js-discussion-reaction-group-count',
+	)!;
+	const count = Number.parseInt(countElement.textContent, 10);
```
fregante, line 60: Probably also works

```suggestion
	const count = Number(countElement.textContent);
```

source/features/quick-reactions.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+function optimisticReactionFromMenu(event: DelegateEvent<MouseEvent>): void {
+	const reactionButton = event.delegateTarget as HTMLButtonElement;
+	const reactionsMenu = reactionButton.closest('reactions-menu')!;
+	const details = reactionsMenu.querySelector('details');
+	details!.open = false; // Close reactions menu immediately
+
+	const reactionsContainer = reactionsMenu.nextElementSibling!.querySelector('.js-comment-reactions-options')!;
+	const reactionButtonValue = reactionButton.getAttribute('value')!;
+	const existingReaction = reactionsContainer.querySelector<HTMLButtonElement>(`[value="${reactionButtonValue}"]`);
```
fregante, line 22: If the selector includes the tag name, you probably won't need to specify the type

```suggestion
	const existingReaction = $(`button[value="${reactionButtonValue}"]`, reactionsContainer);
```

source/features/quick-reactions.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+	if (existingReaction) {
+		// The user is updating an existing reaction via the menu, just update the reaction
+		updateReaction(existingReaction);
+	} else {
+		// Reactions have a specific order, so we need to insert the new reaction in the correct position
```
fregante, line 28: If:

- GitHub still updates the UI later
- it simplifies this piece of code

Please just **append** the reaction instead, it will update within a second anyway. I'd rather have a feature that lasts longer than one that looks _absolutely perfect_ for only one month.

source/features/quick-reactions.tsx
```diff
@@ -0,0 +1,98 @@
...
+		const reactionIndex = emojiSortOrder.findIndex(item => reactionButtonValue.startsWith(item));
+		const insertionIndex = [...reactionsContainer.querySelectorAll('button')].findIndex(button => {
+			const index = Number.parseInt(button.getAttribute('data-button-index-position')!, 10);
+			return index > reactionIndex;
+		});
+		const referenceNode = reactionsContainer.querySelectorAll('button').item(insertionIndex);
+		const emojiElement = reactionButton.querySelector('g-emoji')!.cloneNode(true);
+		emojiElement.className = 'social-button-emoji';
+
+		reactionsContainer.insertBefore(
```
fregante, line 38: See https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-modern-dom-apis.md

source/features/quick-reactions.tsx
```diff
@@ -0,0 +1,98 @@
...
+	} else {
+		// Reactions have a specific order, so we need to insert the new reaction in the correct position
+		const reactionIndex = emojiSortOrder.findIndex(item => reactionButtonValue.startsWith(item));
+		const insertionIndex = [...reactionsContainer.querySelectorAll('button')].findIndex(button => {
+			const index = Number.parseInt(button.getAttribute('data-button-index-position')!, 10);
+			return index > reactionIndex;
+		});
+		const referenceNode = reactionsContainer.querySelectorAll('button').item(insertionIndex);
+		const emojiElement = reactionButton.querySelector('g-emoji')!.cloneNode(true);
+		emojiElement.className = 'social-button-emoji';
```
fregante, line 36: What is this class doing and why is it not there already?

fregante (changes requested):
Alright, I'm almost ready to merge this, probably after the next release.

There are a few notes about the code though.

Also, does this behave well/correctly if I double-click the reaction button? Or is that not-possible?

fregante:
Ping me if you're interested in picking this up

Issue #7261, opened 2024-02-13
Token validator fails with no `read:project` permission even if full `project` permission is given.


134130:
### Description

The below error is appearing..
<img width="766" alt="image" src="https://github.com/refined-github/refined-github/assets/50487467/8fbae5fc-9721-435d-90c1-d4bdd7b12534">

When below permission is given, validator fails. (if only `read:project`, validator works)
<img width="424" alt="image" src="https://github.com/refined-github/refined-github/assets/50487467/29f5fce0-a56e-4fac-b340-9248908301b7">

### How to replicate the issue + URL

NOTE: Only occurs on my private repositories (SAML required Organization)
- https://github.com/refined-github/refined-github/pulls

> Also, `pr-approvals-count`, the css only feature is not works on above pulls page, on my private repositories (SAML required also)
In this case; It seems not `refined-github`'s bug, just GitHub not supplies hint for this feature. Just FYI

### Extension version

24.2.8

### Browser(s) used

Chrome 120.0.6099.234 (arm64)

fregante:
Sigh 🥹 

This check needs to be adjusted further:

- https://github.com/refined-github/refined-github/pull/6728

PR #7262, opened 2024-02-14
`clean-conversation-filters` - fix token permission error

134130:
## Description
- Closes: #7261
- Centralize token-management helper
  - Previously, github-token management was separated to "for option" and "for features"

## Test URLs
NOTE: Only occurs on my private repositories (SAML required Organization)

[refined-github/refined-github/pulls](https://github.com/refined-github/refined-github/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc)

## Screenshot
image

### DIFF
--- a/source/features/quick-review.tsx
+++ b/source/features/quick-review.tsx
@@ -11,6 +11,7 @@ import observe from '../helpers/selector-observer.js';
 import showToast from '../github-helpers/toast.js';
 import {getConversationNumber, getUsername} from '../github-helpers/index.js';
 import {randomArrayItem} from '../helpers/math.js';
+import {getToken} from '../github-helpers/github-token.js';
 
 const emojis = [...'🚀✅🐿️⚡️🤌🥳🥰🤩🥸😎🤯🚢🛫🏳️🏁'];
 
@@ -48,7 +49,7 @@ async function addSidebarReviewButton(reviewersSection: Element): Promise<void> 
 
 	// Can't approve own PRs and closed PRs
 	// API required for this action
-	if (getUsername() === $('.author')!.textContent || pageDetect.isClosedPR() || !(await api.getToken())) {
+	if (getUsername() === $('.author')!.textContent || pageDetect.isClosedPR() || !(await getToken())) {
 		return;
 	}
 
--- a/source/github-helpers/github-token.ts
+++ b/source/github-helpers/github-token.ts
@@ -1,0 +1,40 @@
+import { v3 } from './api.js';
+import optionsStorage from '../options-storage.js';
+
+const settings = optionsStorage.getAll();
+
+export async function getToken(): Promise<string | undefined> {
+	const {personalToken} = await settings;
+	return personalToken;
+}
+
+export async function expectToken(): Promise<string> {
+	const personalToken = await getToken();
+	if (!personalToken) {
+		throw new Error('Personal token required for this feature');
+	}
+	return personalToken;
+}
+
+export async function parseTokenScopes(headers: Headers): Promise<string[]> {
+	// If `X-OAuth-Scopes` is not present, the token may be not a classic token.
+	const scopes = headers.get('X-OAuth-Scopes')?.split(', ') ?? [];
+	scopes.push('valid_token');
+	if (scopes.includes('repo')) {
+		scopes.push('public_repo');
+	}
+
+	if (scopes.includes('project')) {
+		scopes.push('read:project');
+	}
+
+	return scopes;
+}
+
+export async function expectTokenScope(scope: string): Promise<void> {
+	const {headers} = await v3('/');
+	const tokenScopes = await parseTokenScopes(headers);
+	if (!tokenScopes.includes(scope))  {
+		throw new Error('The token you provided does not have ' + (tokenScopes ? `the \`${scope}\` scope. It only includes \`${tokenScopes}\`.` : 'any scope. You can change the scope of your token at https://github.com/settings/tokens'));
+	}
+}
--- a/source/github-helpers/github-token.ts
+++ b/source/github-helpers/github-token.ts
@@ -1,4 +1,4 @@
-import { v3 } from './api.js';
+import {v3} from './api.js';
 import optionsStorage from '../options-storage.js';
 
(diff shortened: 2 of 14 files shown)


source/features/quick-review.tsx
```diff
@@ -11,6 +11,7 @@ import observe from '../helpers/selector-observer.js';
 import showToast from '../github-helpers/toast.js';
 import {getConversationNumber, getUsername} from '../github-helpers/index.js';
 import {randomArrayItem} from '../helpers/math.js';
+import {getToken} from '../github-helpers/github-token.js';
```
134130, line 14: Does the refined-github requires any import-order convention?

fregante: Not anymore 👌

source/github-helpers/github-token.ts
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	return personalToken;
+}
+
+export async function parseTokenScopes(headers: Headers): Promise<string[]> {
+	// If `X-OAuth-Scopes` is not present, the token may be not a classic token.
+	const scopesHeader = headers.get('X-OAuth-Scopes');
+	if (!scopesHeader) {
+		return [];
```
fregante, line 24: What does this do in practice? Avoids the check and just runs the feature anyway? Or does it break all features that require a specific token?

source/github-helpers/github-token.ts
```diff
@@ -0,0 +1,46 @@
...
+export async function expectToken(): Promise<string> {
+	const personalToken = await getToken();
+	if (!personalToken) {
+		throw new Error('Personal token required for this feature');
+	}
+
+	return personalToken;
+}
+
+export async function parseTokenScopes(headers: Headers): Promise<string[]> {
```
fregante, line 20: This doesn't need to be asynchronous

fregante (commented):
I'm missing the specific change that fixes the project token support

source/github-helpers/github-token.ts
134130: This function just 'parse' scopes.
So, running the feature or requiring a specific token is not Interest of it.

`options.tsx` can pass fetch resp's headers, to parse available scopes, and using parsed result (checking permissions, ...), is its freedom.

`expectTokenScope()` can use this function to spray scopes. (like - if project scope exists, append read:project additionally)
throwing error or return boolean with checking includes is its freedom too.

134130:
https://github.com/refined-github/refined-github/blob/main/source%2Ffeatures%2Fclean-conversation-filters.tsx#L25

This is not captured in diff 😅

source/github-helpers/github-token.ts
fregante: Ok but in the end what happens when a feature expects a scope that this does not return?

fregante:
But that line only *adds* a check. For that bug to be fixed, I expect a change in the scope validator itself

source/github-helpers/github-token.ts
134130: When we need that, change this function, but not now.
Currently, we are checking is valid token with appending our custom scope 'valid_token', so we can approach same result. But returning null or undefined for future is just making confuses.

Let's make the simple thing it works. Do not consider too further future. IMO, checking if a header exists may not be necessary forever.

134130:
Aha, just checking do the token has any 'project' or 'read:project' permission?

Ros-Cos (approved):

fregante:
So I don't understand how that can fix a situation where the check was already done elsewhere, and it failed

134130:
Unfortunately, I've failed to understand your comment.
So here is my thought-flow. (Would have been better if it was written in the Description first 🥲)

---

Background
- I have wrote this PR to fix the bug in `clean-conversation-filters` feature.
- The feature needs `read:project` permissioned token.
- When user created token with `projects` permission, an error is throwing.
- But `projects` permission also contains `read:project` permission internally.

In refined-github,
- There is two similar function to control token's permission set.
- One is for [options](https://github.com/refined-github/refined-github/pull/7262/files#diff-a24f014d7cd2bf18b76e05a5879d46687e5ebdebee94fb2632d724b63ae41464L81-L92), another one is for [features](https://github.com/refined-github/refined-github/pull/7262/files#diff-317127f4f5c1e5c18ca5c1ab05d9af620ca22ec6cd8e62cad7ed4847c03741deL79-L82).

So I thought token permission handling is separated and duplicated currently. And I also thought this utility function can be use in many cases.

I have created a new function: `parseTokenScopes()` which appends related sub-permissions additionally. For example:
- `repo` permission contains sub-permission `public_repo`
- `project` permission contains sub-permission `read:project`
- https://github.com/refined-github/refined-github/pull/7262/files#diff-bc3cdd5ee48cee7907b51d15d595bcad876234978442024474b7133c4122340aR27-R35

<br>

And I changed `expectTokenScope` from [this](https://github.com/refined-github/refined-github/pull/7262/files#diff-317127f4f5c1e5c18ca5c1ab05d9af620ca22ec6cd8e62cad7ed4847c03741deL77-L83) to [this](https://github.com/refined-github/refined-github/pull/7262/files#diff-bc3cdd5ee48cee7907b51d15d595bcad876234978442024474b7133c4122340aR40-R46)
- So new `expectTokenScope('read:project')` can not throws when `project` permission is given.
- And that fixes `clean-conversation-filters` ([code](https://github.com/refined-github/refined-github/pull/7262/files#diff-3af8a657b0fe4460af1e140b0037e066232a5785b12cb5939ab241d0eacc08edL7-L28))

fregante:
Look, you linked to this line and said that it fixed #7261:

https://github.com/refined-github/refined-github/blob/fc0c70921b64419d5cf3cc5bec694b88a0b4077d/source/features/clean-conversation-filters.tsx#L25

That line ***did not*** fix that issue because that line already exists on `main`. The line that fixed it is this one:

https://github.com/refined-github/refined-github/blob/fc0c70921b64419d5cf3cc5bec694b88a0b4077d/source/features/clean-conversation-filters.tsx#L25

fregante (approved):

PR #7269, opened 2024-02-18
`unreleased-commits` - Avoid wrapping

### LINKED ISSUE #7268: `unreleased-commits`: draft release button wraps below
### Description

After the branch and tags count is loaded, the draft release button wraps down.

expected:
image

actual:
image

Happens on https://github.com/coronasafe/care_fe (with write access)

### Extension version

24.2.8

### Browser(s) used

firefox

sainak:
## Test URLs
https://github.com/coronasafe/care_fe

## Screenshot
image

fixes #7268

### DIFF
--- a/source/features/unreleased-commits.tsx
+++ b/source/features/unreleased-commits.tsx
@@ -115,7 +115,7 @@ async function createLinkGroup(latestTag: string, aheadBy: number): Promise<Elem
 		</a>,
 	]);
 
-	linkGroup.classList.add("rgh-unreleased-commits-home-button-group");
+	linkGroup.classList.add('rgh-unreleased-commits-home-button-group');
 	return linkGroup;
 }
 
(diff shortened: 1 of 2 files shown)


source/features/unreleased-commits.tsx
```diff
@@ -132,7 +132,7 @@ async function addToHome(branchSelector: HTMLButtonElement): Promise<void> {
 		);
 	} else {
 		wrapAll(
-			<div className="d-flex gap-2"/>,
+			<div className="d-flex gap-2" style={{width: "max-content"}}/>,
```
fregante, line 135: It's weird, why is flexbox wrapping? Normally it shouldn't. The right class for this should be `flex-nowrap` or something like that, if they have it, but that's the default for flex box

sainak: The buttons are in a button group which is a normal div, so that forces it to wrap, we should ideally set width on this group element, but I found it simpler to set this on the branch parent element

fregante:
after

sainak:
image

fregante:
Yeah exactly, this issue is because there's no space on that line, so either something wraps or we need to hide something earlier than GitHub does.

Similar bug in the past with the feature that preceded `unreleased-commits`

- https://github.com/refined-github/refined-github/issues/5071

PR #7284, opened 2024-03-09
`restore-file` - fix(restore-file): use conventional commit

skjnldsv:
https://www.conventionalcommits.org


fregante:
Thanks but I'd rather not

skjnldsv:
@fregante why not? :)

fregante:
It's just one of many conventions, I don't think I've ever contribute to more than 3 projects with that conventions ever.

skjnldsv:
Ah, too bad!
Many popular tools now supports it.
Dependabot, renovate... etc

PR #7306, opened 2024-03-20
Fix colors on dark themes 

### LINKED ISSUE #7305: Excessive box shadow in dark mode
### Description

In dark mode on Firefox Release 124, when navigating to https://github.com/notifications, there is a new excessive box shadow added by this style: https://github.com/refined-github/refined-github/blob/5e2f132d99a9ea3825cc5f01d19f0c38d3460c12/source/features/sticky-notifications-actions.css#L2

Screenshot 2024-03-19 at 21 18 49

I just updated from Firefox 123 to 124, I am not sure the add-on has been updated along the way (I doubt it since the add-on was last updated a month ago). Nothing else has changed.

Browse https://github.com/notifications in dark mode on Firefox Release 124.

### Extension version

24.2.8

### Browser(s) used

Firefox Release 124.0

FlaminSarge:
GitHub recently seems to have changed its `--color*` vars to `--fgColor*`/`--bgColor*`. This updates the `--rgh*` color vars to use the new names, as leaving them at defaults causes issues in colorblind/dark themes.

Note that `--fgColor-danger` in Default Light Mode is now `#d1242f` which differs from the previously (still) defined default we have of `#cf222e` (now on `--bgColor-danger-emphasis`, among other vars). I am colorblind so I will need someone else to tell me if this is a significant difference and/or if the default value should also be updated for `--rgh-red`.

Also note that `--borderColor-muted` in Default Light Mode is now `#d0d7deb3` which differs from our default of `#d8dee4`. A var with this color is no longer present in Default Light Mode, but we can consider changing it to `--borderColor-default` instead, which is `#d0d7de` (no alpha).

Fixes #7305 

## Test URLs
https://github.com/notifications, various other pages that rely on rgh* color values.

## Screenshot
Old Light:
Screenshot 2024-03-19 at 5 04 19 PM

New Light:
Screenshot 2024-03-19 at 5 04 59 PM

Old Dark:
Screenshot 2024-03-19 at 5 05 38 PM

New Dark:
Screenshot 2024-03-19 at 5 07 06 PM

### DIFF
--- a/source/refined-github.css
+++ b/source/refined-github.css
@@ -1,7 +1,10 @@
 :root {
 	--rgh-green: var(--fgColor-success, var(--color-success-fg, #1a7f37));
 	--rgh-red: var(--fgColor-danger, var(--color-danger-fg, #cf222e));
-	--rgh-border-color: var(--borderColor-muted, var(--color-border-muted, #d8dee4));
+	--rgh-border-color: var(
+		--borderColor-muted,
+		var(--color-border-muted, #d8dee4)
+	);
 	--rgh-background: var(--bgColor-default, var(--color-canvas-default, #fff));
 }
 
--- a/source/refined-github.css
+++ b/source/refined-github.css
@@ -1,8 +1,8 @@
 :root {
-	--rgh-green: var(--color-success-fg, #1a7f37);
-	--rgh-red: var(--color-danger-fg, #cf222e);
-	--rgh-border-color: var(--color-border-muted, #d8dee4);
-	--rgh-background: var(--color-canvas-default, #fff);
+	--rgh-green: var(--fgColor-success, #1a7f37);
+	--rgh-red: var(--fgColor-danger, #cf222e);
+	--rgh-border-color: var(--borderColor-muted, #d8dee4);
+	--rgh-background: var(--bgColor-default, #fff);
 }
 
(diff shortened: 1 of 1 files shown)


source/refined-github.css
```diff
@@ -1,8 +1,8 @@
 :root {
-	--rgh-green: var(--color-success-fg, #1a7f37);
-	--rgh-red: var(--color-danger-fg, #cf222e);
-	--rgh-border-color: var(--color-border-muted, #d8dee4);
-	--rgh-background: var(--color-canvas-default, #fff);
+	--rgh-green: var(--fgColor-success, #1a7f37);
```
busches, line 2: Can we fallback to the old for GHE?

FlaminSarge: Ah, good point, I was wondering whether GHE would need backcompat on it.
Would that be var(GH, var(GHE, default))?

FlaminSarge: From what I'm reading, it looks like var(var1, fallback1, fallback2) behavior is inconsistent across browsers, so I'm going with nested var.

busches: Yea, I believe you have to do `var(GH, var(GHE, default)`

fregante:
Thank you @FlaminSarge! That fixed it.

I expect more issue to crop up since we have more such variables, but this seems to be the most urgent/visible

https://github.com/search?q=repo:refined-github/refined-github%2520--color-&type=code

Issue #7308, opened 2024-03-20
`file-age` not working anymore?


melMass:
### Description

The file-age feature (#6340) seems to have broken very recently:
![image](https://github.com/refined-github/refined-github/assets/7041726/846b4674-0884-4894-8d8d-0e8eda222c70)

### How to replicate the issue + URL

https://github.com/refined-github/refined-github

### Extension version

24.3.20

### Browser(s) used

Firefox

fregante:
Likely related to:

- https://github.com/refined-github/refined-github/pull/7306

GitHub changed the variable color names so we should replace all of them at once as part of this issue. 

PR welcome 🙏

melMass:
I'm going to try to setup a dev env for it, I couldn't quickly find how to test it locally but I might next week if no one did by then.
It seems like `js-navigation-item` doesn't exist anymore yet the `react-directory-commit-age` does have the proper style if I look from chrome tools but what is targeting `title` is now in a shadow-root:

![image](https://github.com/refined-github/refined-github/assets/7041726/d58bc24b-1204-421d-8755-223b9e5f62da)

fregante:
They probably changed the DOM too

ilike2burnthing:
Oddly it works when not signed in to GitHub (personal token or not, dark mode or light), and I can't see any difference in the HTML that would explain it.

offensive-vk:
Same issue with Brave, Chrome in windows too.

PR #7313, opened 2024-03-22
Use new GitHub color variables

kovsu:
close: #7308
close: #7317

## Screenshot

截屏2024-03-22 16 52 46

### DIFF
--- a/source/features/clean-conversation-headers.css
+++ b/source/features/clean-conversation-headers.css
@@ -1,9 +1,15 @@
 :root .rgh-clean-conversation-headers-non-default-branch {
-	background-color: var(--color-state-hover-primary-bg, #0366d6);
+	background-color: var(
+		--control-checked-bgColor-hover,
+		var(--color-state-hover-primary-bg, #0366d6)
+	);
 }
 
 .rgh-clean-conversation-headers-non-default-branch a {
-	color: var(--color-state-hover-primary-text, #fff);
+	color: var(
+		--control-checked-fgColor-rest,
+		var(--color-state-hover-primary-text, #fff)
+	);
 }
 
 /* Removes: every text (but not icons) */
--- a/source/features/extend-diff-expander.css
+++ b/source/features/extend-diff-expander.css
@@ -5,8 +5,17 @@
 		.blob-num:not(:hover) .directional-expander:first-child,
 		.blob-num:not(:hover) + .blob-code
 	) {
-	color: var(--color-state-hover-primary-text, #fff);
-	background: var(--color-state-hover-primary-bg, #0366d6);
-	border-color: var(--color-state-hover-primary-border, #0366d6);
+	color: var(
+		--control-checked-fgColor-rest,
+		var(--color-state-hover-primary-text, #fff)
+	);
+	background: var(
+		--control-checked-bgColor-hover,
+		var(--color-state-hover-primary-bg, #0366d6)
+	);
+	border-color: var(
+		--control-checked-borderColor-hover,
+		var(--color-state-hover-primary-border, #0366d6)
+	);
 	cursor: pointer;
 }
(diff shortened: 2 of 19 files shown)


kovsu:
截屏2024-03-22 17 09 07

截屏2024-03-22 17 10 18

kovsu:
I don't think the ci fail is caused by my code. 😧

fregante:
It is, we have a prettier script in package.json, just for CSS. 

The `Test URLs` job left an annotation in the Files tab with more info

kovsu:
@fregante I get it. I find the `TestURLs` failed is because the `mark-private-repos` feature, but it doesn't have the feature actually.

image

fregante:
> but it doesn't have the feature actually.

It does, it's CSS only, you're editing the file in this PR, that's why it's complaining.

fregante (approved):
Thank you! Does this cover all of our --color variables?

kovsu:
@fregante Yes, it covers all of --color variables now. 😀

source/features/extend-diff-expander.css
```diff
@@ -5,8 +5,17 @@
 		.blob-num:not(:hover) .directional-expander:first-child,
 		.blob-num:not(:hover) + .blob-code
 	) {
-	color: var(--color-state-hover-primary-text, #fff);
-	background: var(--color-state-hover-primary-bg, #0366d6);
-	border-color: var(--color-state-hover-primary-border, #0366d6);
+	color: var(
+		--control-checked-fgColor-rest,
```
fregante, line 9: Screenshot 1

Some of the colors are not right, this should match the color it extends. It would be great if you could verify each color in the dark theme via screenshots 🥹 🙏 

Here the right one `var(--bgColor-accent-emphasis, var(--color-accent-emphasis))`

fregante: > unless it's an exact match

Here are the files that don't have exact matches:

Screenshot 3

kovsu:
`clean-conversation-header.css`

test urls is `https://github.com/Kenshin/simpread/pull/698`, because I only updated the non-default-branch's variables

截屏2024-03-25 22 51 27

截屏2024-03-25 22 56 47


image

kovsu:
`extend-diff-expander.css`

The invalid variables in this file are the same as those in `clean-conversation-headers.css`.

kovsu:
`mark-private-repos.css`

It has `--bgColor-attention-emphasis` and `--borderColor-attention-emphasis`, but in `mark-private-repos.css` what I need to update is `color` not `background-color` or `border-color`, so I choose `--fgColor-attention`.

kovsu:
`show-associated-branch-prs-on-fork.css`

截屏2024-03-25 23 26 52

image

kovsu:
`refined-github.css`

It just has one variable called `--shadow-floating-large` contain the `large` word.

image

source/features/extend-diff-expander.css
fregante: This is still not resolved, the background should not be white

source/features/clean-conversation-headers.css
```diff
@@ -1,9 +1,15 @@
...
+	background-color: var(
+		--control-checked-bgColor-hover,
+		var(--color-state-hover-primary-bg, #0366d6)
+	);
 }
 
 .rgh-clean-conversation-headers-non-default-branch a {
-	color: var(--color-state-hover-primary-text, #fff);
+	color: var(
+		--control-checked-fgColor-rest,
```
fregante, line 10: I think this should match the `bicolor` state, which is

```suggestion
		--control-checked-fgColor-hover,
```

source/features/extend-diff-expander.css
fregante: Oh yeah it works 🤔 

Screenshot

fregante:
Thank you @kovsu! 🙏 🙏

yakov116:
Thanks!!

PR #7324, opened 2024-04-01
`repo-avatars` - Fix wrong organization avatar

### LINKED ISSUE #7322: `repo-avatars` - Wrong organization icon
### Description

It shows a bot avatar when the bot has the same display name.

See this link below:

- https://github.com/simple-icons/simple-icons/pull/10728.

### Extension version

24.3.20

### Browser(s) used

Safari, Chrome

LitoMore:
Closes #7322

I've added a new optional parameter `ignoreBots`, for the `repo-avatars` scene.

## Test URLs

- https://github.com/simple-icons/simple-icons/pull/10728

### DIFF
--- a/source/features/repo-avatars.tsx
+++ b/source/features/repo-avatars.tsx
@@ -9,7 +9,7 @@ import observe from '../helpers/selector-observer.js';
 async function add(ownerLabel: HTMLElement): Promise<void> {
 	const username = getRepo()!.owner;
 	const size = 16;
-	const src = getUserAvatar(username, size, true)!;
+	const src = getUserAvatar(username, size)!;
 
 	const avatar = (
 		<img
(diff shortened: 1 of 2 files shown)


fregante:
Thank you for the PR! Sorry for the delay, I wanted to think about this for a bit. 

I think that rather than a flag like this, it should just change the way it selects existing avatars in every case. 

There could be a @LitoMore user and a "LitoMore" app. The `alt` text provided by GitHub is actually incorrect, you cannot @-mention apps, you can only @-mention actual user/orgs (sometimes this overlaps, sometimes this doesn't)

So instead of querying `[alt="@simple-icon"]` it should query `[href="/simple-icon" I]`. I think that's what we had at some point, so we had to blame the line and see why it was changed.

fregante:
Indeed! So that selector is new specifically to add support for bots, the `/user` was just a raw request URL.

My suggested selector should work

fregante (approved):
Thank you! Have you tested this change? Does it fix the issue for you?

LitoMore:
@fregante I've tested it on my side. It's working correctly.

PR #7347, opened 2024-05-03
fix: restore previous layout

kovsu:
I remember that the buttons was at the right

## Screenshot
before: 
CleanShot 2024-05-03 at 17 42 27@2x

after:
CleanShot 2024-05-03 at 17 42 10@2x

### DIFF
--- a/source/features/select-notifications.tsx
+++ b/source/features/select-notifications.tsx
@@ -127,7 +127,7 @@ function createDropdownList(category: Category, filters: Filter[]): JSX.Element 
 
 const createDropdown = onetime(() => (
 	<details
-		className="details-reset details-overlay position-relative rgh-select-notifications mx-2"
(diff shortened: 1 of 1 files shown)


fregante:
What does it look like without Refined GitHub? We shouldn't change that

source/features/select-notifications.tsx
```diff
@@ -127,7 +127,7 @@ function createDropdownList(category: Category, filters: Filter[]): JSX.Element
 
 const createDropdown = onetime(() => (
 	<details
-		className="details-reset details-overlay position-relative rgh-select-notifications mx-2"
+		className="details-reset details-overlay position-relative rgh-select-notifications mx-2 mr-auto"
```
fregante, line 130: You probably just need to change the existing class
```suggestion
		className="details-reset details-overlay position-relative rgh-select-notifications ml-2"
```

PR #7349, opened 2024-05-07
`collect-commit-comments` - Show all commit comments in the comment section

Leguan16:
## Notes
This is my first time doing this kind of stuff and I am sure there is a lot of improvements to make. 
I don't think there is an easy way to replicate comments, so I had to create them manually. This is also the reason why it misses several features. The reason why I want to add such a feature is because it's hard to find comments on large commits (See test url). I would like to get feedback about the general idea and maybe some people can give me tips on how to improve things  :)

## Test URLs

https://github.com/PaperMC/DataConverter/commit/31387432cacf86b8e76362c6d7de44c5104afbf1#comments

### Before
image

### After
image

### DIFF
--- a/source/features/collect-commit-comments.css
+++ b/source/features/collect-commit-comments.css
@@ -1,0 +1,17 @@
+.rgh-comment-header {
+	display: flex;
+	align-items: center;
+	padding-left: 16px;
+	padding-top: 8px;
+	padding-bottom: 8px;
+	color: var(--fgColor-muted, var(--color-fg-muted));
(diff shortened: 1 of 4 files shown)


fregante:
Interesting feature but we never try to recreate large pieces of UI in Refined GitHub, we just focus on small improvements. Here each comment DOM is missing a lot of features

Leguan16:
Yeah, I agree it's not ideal but it's the best we can do with what github gives us to work with. It's just annoying to find comments in large commits, so i hope they add such feature to the official website. Thanks for the feedback :)

PR #7355, opened 2024-05-11
Add dependabot

sebastiaanspeck:
This would result in 10 PR’s opened by Dependabot:
https://github.com/sebastiaanspeck/refined-github/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc


fregante:
I'm good thank you

fregante:
That's not what I want, but I know I don't want dependabot

PR #7365, opened 2024-05-20
feat: expand avatar affect & bind username match avatar

btea:
## Test URLs
No.

## Screenshot

https://github.com/refined-github/refined-github/assets/24516654/ebffcb94-4905-42f0-93b6-75882410c6dd


fregante:
No

PR #7374, opened 2024-05-27
Update XO

fregante:
- Extracted from https://github.com/refined-github/refined-github/pull/7367
- Follows https://github.com/refined-github/refined-github/pull/7373


source/features/list-prs-for-file.tsx
```diff
@@ -71,7 +71,7 @@ const getPrsByFile = new CachedFunction('files-with-prs', {
 
 		for (const pr of repository.pullRequests.nodes) {
 			for (const {path} of pr.files.nodes) {
-				files[path] = files[path] ?? [];
+				files[path] ??= [];
```
yakov116, line 74: Just double checking this is supported in all browsers already?

build/features.test.ts
```diff
@@ -58,7 +58,7 @@ class FeatureFile {
 		return existsSync(this.path);
 	}
 
-	// eslint-disable-next-line n/prefer-global/buffer -- Type only
+	// eslint-disable-next-line @typescript-eslint/ban-types -- Just passing it
 	contents(): Buffer {
```
sindresorhus, line 62: `Buffer` is a `Uint8Array` subclass, so you could just use
```suggestion
	contents(): Uint8Array {
```

here.

sindresorhus (changes requested):

build/features.test.ts
fregante: Buffer has an `includes(string)` method but `Uint8Array` doesn't. Since that's what I'm using it for, I'll keep Buffer for now so I can merge this. Happy to change it if there's a similar method that doesn't require conversions. Maybe later I'll just read it as utf-8 instead

PR #7376, opened 2024-05-30
`unreleased-commits` - Avoid internal wrapping

### LINKED ISSUE #7375: "Draft a new release" button is wrapped
### Description

The "Draft a new release" button is being wrapped onto a new line instead of being in-line. I started encountering this a week or two ago.

image

I believe this buttons only show up when you have access to releases on the repo. So unless someone has created a repo giving every GitHub user release permissions, I don't think it's possible for me to find a public URL.

https://github.com/SVGtoPPT/svgtoppt

### Extension version

24.4.9

### Browser(s) used

Arc

g1eny0ung:
Close #7375.

image

Since the branch picker sets the width to 100%, the unreleased commits button group will shrink in a flexbox.

Set `flex-shrink: 0` solves this problem, it will let the width of the button group grow from its min-content.

### DIFF
--- a/source/features/unreleased-commits.tsx
+++ b/source/features/unreleased-commits.tsx
@@ -131,10 +131,14 @@ async function addToHome(branchSelector: HTMLButtonElement): Promise<void> {
 			await createLinkGroup(latestTag, aheadBy) as HTMLElement,
 		);
 	} else {
+		const linkGroup = await createLinkGroup(latestTag, aheadBy) as HTMLElement;
+
+		linkGroup.style.flexShrink = '0';
+
 		wrapAll(
 			<div className="d-flex gap-2"/>,
 			branchSelector,
-			await createLinkGroup(latestTag, aheadBy),
+			linkGroup,
 		);
 	}
 }


fregante:
See https://github.com/refined-github/refined-github/pull/7269

Does this PR have the same issue? Can you take a video showing how it behaves on window resize?

g1eny0ung:
> See #7269
> 
> Does this PR have the same issue? Can you take a video showing how it behaves on window resize?

Not quite the same, #7269 arbitrarily changes the original button group. It turns it into a flexbox, which is unnecessary because the original button group already allows the internal elements to be placed horizontally. Making it a flexbox just makes its children a block, so they don't wrap (because the minimum width (min-content) is the width of the two blocks together).

So I think adding `flex-shrink: 0` is a better way to solve it.

https://github.com/refined-github/refined-github/assets/15034155/d94b48c1-ec3a-467d-998a-b72ce696665d

But I'm not sure if GitHub has updated this part of the style, it doesn't look like the problem I'm finding is all that consistent with #7269, currently it isn't a lack of width issue.

fregante:
> But I'm not sure if GitHub has updated this part of the style

I think they did.


I can confirm that your change fixes the issue 🎉

Untitled

Issue #6000, opened 2022-09-19
🥉 Minor codebase updates and fixes


fregante:
Congrats! I'm gonna use this issues as a placeholder for the next #5222 issue 😌 

plz don't comment or add labels.

If you see this issue, you will have 56 years of good luck.

fregante:
https://github.com/refined-github/refined-github/issues/5222 is starting to hide some events in the middle, so I'll start adding more items here:

- [ ] Update `@primer/octicons-react` (17.11 breaks dom-chef)

https://github.com/refined-github/refined-github/pull/6326/files#r1103554022

fregante:
- [ ] Fix some TODOs 
- [ ] Review and update ancient screenshots, e.g. https://user-images.githubusercontent.com/1402241/54178764-d1c96080-44d1-11e9-889c-734ffd2a602d.png

Gifs may be left unchanged because they're harder to replicate.

fregante:
- [ ] Revert https://github.com/refined-github/refined-github/pull/3208

This works now in Firefox 🎉 

```js
Object.getPrototypeOf(new (class extends URL {method() {}})(location.href)).method
```

Context: https://github.com/refined-github/refined-github/issues/3193#issuecomment-93760

Screenshot: https://github.com/refined-github/refined-github/assets/1402241/91d43f7b-54b8-459a-b851-bd2a3650ca40

cc @kidonng

fregante:
- [ ] Use `SearchQuery` in most places where we build/alter search queries.

Examples: https://github.com/search?q=repo%3Arefined-github%2Frefined-github+%2Fis%3Aissue%2F&type=code

fregante:
- [ ] Use `experiments.css` in webpack: https://webpack.js.org/configuration/experiments/#experimentscss

> CSS extraction: webpack will extract CSS into a separate file. This functionality replaces the need for mini-css-extract-plugin and css-loader, as it provides native support.

fregante:
- [ ] Show the username of the entered token (as part of the scope check UI)

fregante:
- [ ] Correctly support URLs like `https://github.com/refined-github/refined-github/assets/50487467/747b5133-61b8-490e-b961-4e5390ea9482` in the readme-parser and feature validator. Currently they're silently ignored

Seen in https://github.com/refined-github/refined-github/pull/6900#pullrequestreview-1621264680

fregante:
- [ ] Add waitForChildren to the selector observer


[...]

PR #7379, opened 2024-06-01
Update token input box

its-miroma:
Fix part of #6000

## Screenshot

Updated token section


its-miroma:
I'll lint it soon, sorry

fregante:
Thank you for the PR! This somewhat related to https://github.com/refined-github/refined-github/issues/7231, rather than #6000. 

I intentionally left the token visible to not lull the user into a false sense of security since the token is stored in clear text in Chrome's config. Also personally I leave it visible in order to reuse it across browsers/extensions  as needed. 

I'm not sure I want to make this specific change here, but I'm open for discussion

its-miroma:
> This somewhat related to https://github.com/refined-github/refined-github/issues/7231, rather than https://github.com/refined-github/refined-github/issues/6000.

I updated the description, to highlight the comment this PR is related to.

> I intentionally left the token visible to not lull the user into a false sense of security

I think it's wrong to think "Let's not implement X security feature because Y is not implemented", because they are two different problems: if you're recording/sharing the screen, you do not want to accidentally leak the token.

> the token is stored in clear text in Chrome's config

That's a different issue, that you can argue is way more critical, but it's not what I am trying to fix.

> I leave it visible in order to reuse it across browsers/extensions as needed

Of course, that's why I added a checkbox to display the token

fregante:
> you do not want to accidentally leak the token

This is mostly accomplished by the small field. The token is never shown in its entirety, but there are always 6+ characters missing

its-miroma:
Should I create a different PR to only add the username, as suggested in https://github.com/refined-github/refined-github/issues/6000#issuecomment-1556136109?

fregante:
Yes please that would be great

PR #7380, opened 2024-06-02
Show token owner in options page

its-miroma:
Related to <https://github.com/refined-github/refined-github/issues/6000#issuecomment-1556136109>

## Screenshot

why_is_that_token_not_hidden

### DIFF
--- a/source/options.tsx
+++ b/source/options.tsx
@@ -115,7 +115,7 @@ async function validateToken(): Promise<void> {
 
 	try {
 		const scopes = await getTokenScopes(tokenField.value);
-		const text = `👋 Welcome, ${await getNameFromToken(tokenField.value)}!`;
+		const text = `👤 @${await getNameFromToken(tokenField.value)}`;
 		reportStatus({
 			tokenType,
 			text,
@@ -137,14 +137,13 @@ async function getNameFromToken(token: string): Promise<string> {
 			},
 		},
 	);
-	const userData = await response.json();
 
-	let name: string = userData.login;
-	if (name !== userData.name) {
-		name += ` (${userData.name})`;
+	const userData = await response.json();
+	if (!response.ok) {
+		throw new Error(userData.message);
 	}
 
-	return name;
+	return userData.login;
 }
 
 function moveDisabledFeaturesToTop(): void {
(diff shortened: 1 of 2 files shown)


fregante (commented):
Thank you! Can you move it next to the field instead? Where this appears:

Screenshot 2024-06-02 at 14 16 03

Mockup:

Screenshot 2024-06-02 at 14 19 17

its-miroma:
ٍSome longer usernames and names may not fit next to the token input box.

fregante:
I see space for about 35 characters, if someone has a username that long they're seeing it wrapping/clipping/causing issues all over the interface already.
Screenshot 2024-06-02 at 14 34 59

and even then, it's already very close to the limit:

Screenshot 2024-06-02 at 14 36 44

its-miroma:
No, I mean that I had implemented `show-names`, which would show `@its-miroma (Miroma)`

fregante:
Ah yes, that part should be removed. `show-names` doesn't show your own username either for example.

source/options.tsx
```diff
@@ -114,9 +114,12 @@ async function validateToken(): Promise<void> {
 	reportStatus({text: 'Validating…', tokenType});
 
 	try {
+		const scopes = await getTokenScopes(tokenField.value);
+		const text = `👤 @${await getNameFromToken(tokenField.value)}`;
```
fregante, line 118: Better parallelize them:
```suggestion
		const [scopes, user] = await Promise.all([
			getTokenScopes(tokenField.value),
			getNameFromToken(tokenField.value),
		]);
```

The emoji is part of the UI so it should be added near the DOM/HTML generation

its-miroma: > The emoji is part of the UI so it should be added near the DOM/HTML generation

I'm sorry, I'm unsure how to do that, I can't find any place where the HTML is generated. The images seem to get added through CSS

fregante:
I like it!

Untitled

Thank you Miroma!

fregante:
I forgot to mention this: it lacks support for GitHub Enterprise and it will break for GHE users. Can you add it? It looks like this:

https://github.com/refined-github/refined-github/blob/eb1d1af666dfd5de3cf783728b47dfd22fa0e746/source/options.tsx#L62-L66

PR #7383, opened 2024-06-02
`quick-review` - Restore auto-focus

### LINKED ISSUE #7272: `quick-review` doesn't autofocus the field
### Description

This selector doesn't match anything:

https://github.com/refined-github/refined-github/blob/e0eeb515af6fb6931c6c5b2bf991ffff4b775b9e/source/features/quick-review.tsx#L80-L81

First reported in https://github.com/refined-github/refined-github/issues/7256

1. Open https://github.com/refined-github/sandbox/pull/12/files#review-changes-modal
2. See the field not being focused

### Extension version

24.2.8

### Browser(s) used

Safari

134130:
## Description
- Closes #7272

## Explain
- GitHub changed to use [Top Layer](https://developer.mozilla.org/en/docs/Glossary/Top_layer) with [Popover API](https://developer.mozilla.org/en/docs/Web/API/Popover_API) on modal elements.
- And the [docs](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API#interfaces) says popover triggers `ToggleEvent` when they called. but `lib.dom.ts` is typed with `Event` maybe for backward compatibility.
- So I hacked the type like `(event as unknown) as ToggleEvent`

## Test URLs
- Open [refined-github/sandbox#12 (files)](https://github.com/refined-github/sandbox/pull/12/files#review-changes-modal)
See the field not being focused

## Screenshot
https://github.com/refined-github/refined-github/assets/50487467/fd653e1b-e25b-4a5c-8ac5-52eddc5c26c5

### DIFF
--- a/source/features/quick-review.tsx
+++ b/source/features/quick-review.tsx
@@ -70,7 +70,7 @@ async function initSidebarReviewButton(signal: AbortSignal): Promise<void> {
 	delegate('.rgh-quick-approve', 'click', quickApprove, {signal});
 }
 
-function focusReviewTextarea(event: DelegateEvent<Event, HTMLDetailsElement>): void {
+function focusReviewTextarea(event: DelegateEvent<Event, HTMLElement>): void {
 	if (((event as unknown) as ToggleEvent).newState === 'open') {
 		$('textarea', event.delegateTarget)!.focus();
 	}
--- a/source/features/quick-review.tsx
+++ b/source/features/quick-review.tsx
@@ -70,9 +70,8 @@ async function initSidebarReviewButton(signal: AbortSignal): Promise<void> {
 	delegate('.rgh-quick-approve', 'click', quickApprove, {signal});
 }
 
-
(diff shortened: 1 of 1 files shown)


fregante:
> * And the [docs](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API#interfaces) says popover triggers `ToggleEvent` when they called. but `lib.dom.ts` is typed with `Event` maybe for backward compatibility.

See note in https://developer.mozilla.org/en-US/docs/Web/API/ToggleEvent

The type might technically be `Event | ToggleEvent`, which means TypeScript would simplify/broaden it to `Event` anyway.

fregante:
> * Open [refined-github/sandbox#12 (files)](https://github.com/refined-github/sandbox/pull/12/files#review-changes-modal)
>   See the field not being focused

Indeed the field loses the focus when using the deep link, sometimes:


https://github.com/refined-github/refined-github/assets/1402241/a46e97a8-0c4b-4a16-b288-93a67a3962e4



It's good enough for now

PR #7388, opened 2024-06-02
Show token owner in options page: GHE support

its-miroma:
Related: <https://github.com/refined-github/refined-github/pull/7380#issuecomment-2143737791>

### DIFF
--- a/source/options.tsx
+++ b/source/options.tsx
@@ -143,14 +143,12 @@ async function getNameFromToken(token: string): Promise<string> {
 		},
 	);
 
+	const details = await response.json();
 	if (!response.ok) {
-		const details = await response.json();
 		throw new Error(details.message);
 	}
 
-	const userData = await response.json();
-
-	return userData.login;
+	return details.login;
 }
 
 function moveDisabledFeaturesToTop(): void {
--- a/source/options.tsx
+++ b/source/options.tsx
@@ -132,8 +132,13 @@ async function validateToken(): Promise<void> {
 }
 
 async function getNameFromToken(token: string): Promise<string> {
+	const tokenLink = $('a#personal-token-link')!;
+	const url = tokenLink.host === 'github.com'
+		? 'https://api.github.com/'
+		: `${tokenLink.origin}/api/v3/`;
+
 	const response = await fetch(
-		'https://api.github.com/user', {
+		url, {
 			headers: {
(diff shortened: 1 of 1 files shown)


source/options.tsx
```diff
@@ -132,19 +132,26 @@ async function validateToken(): Promise<void> {
 }
 
 async function getNameFromToken(token: string): Promise<string> {
+	const tokenLink = $('a#personal-token-link')!;
+	const url = tokenLink.host === 'github.com'
+		? 'https://api.github.com/'
+		: `${tokenLink.origin}/api/v3/`;
```
fregante, line 138: Could you extract this URL logic to its own function? e.g. `x('user')`

its-miroma:
Some code uses `Url` while some other uses `URL`. I went for the latter.

source/options.tsx
```diff
@@ -133,18 +136,20 @@ async function validateToken(): Promise<void> {
 
 async function getNameFromToken(token: string): Promise<string> {
 	const response = await fetch(
-		'https://api.github.com/user', {
+		getApiURL(), {
```
fregante, line 139: /user?

source/options.tsx
```diff
@@ -59,13 +59,16 @@ function reportStatus({tokenType, error, text, scopes}: Status): void {
 	}
 }
 
-async function getTokenScopes(personalToken: string): Promise<string[]> {
+function getApiURL(): string {
```
fregante, line 62: Let's prefer Url unless the function returns a `new URL` instance (which in this case it doesn't)

fregante (approved):

its-miroma:
Sorry for making many rookie errors, I'm still very much a beginner

source/options.tsx
```diff
@@ -133,18 +136,20 @@ async function validateToken(): Promise<void> {
...
 	const response = await fetch(
-		'https://api.github.com/user', {
+		`${getApiUrl()}/user`, {
 			headers: {
 				Authorization: `Bearer ${token}`,
 			},
 		},
 	);
 
-	const userData = await response.json();
```
fregante, line 143: I prefer this to be outside. You can call it details

fregante (commented):
No worries about the mistakes 🙌

fregante (approved):

PR #7421, opened 2024-06-12
`fit-textareas` - Enable CSS-native feature where available

fregante:
- Closes https://github.com/refined-github/refined-github/issues/7414

After this PR:

- Chrome 123+: uses `field-sizing`
- Firefox and Chrome 122: uses `fit-textarea@2` module
- Safari: enabled only when `field-sizing` will be available
	- Since `fit-textarea` can introduce new performance issues, I'd rather not enabled it

Tracking issues:

- https://bugzilla.mozilla.org/show_bug.cgi?id=1832409
- https://bugs.webkit.org/show_bug.cgi?id=264720

## Test URLs

Here

## Screenshot

See how the field is resized without any changes to the DOM

https://github.com/refined-github/refined-github/assets/1402241/ea600268-3c99-4172-a84d-f363c5ed1e34

### DIFF
--- a/source/features/fit-textareas.tsx
+++ b/source/features/fit-textareas.tsx
@@ -1,9 +1,12 @@
 import './fit-textareas.css';
+import {isSafari} from 'webext-detect-page';
 import fitTextarea from 'fit-textarea';
 import * as pageDetect from 'github-url-detection';
 
 import features from '../feature-manager.js';
 import observe from '../helpers/selector-observer.js';
+
+const nativeFit = 'fieldSizing' in document.body.style;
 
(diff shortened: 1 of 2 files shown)


source/features/fit-textareas.tsx
```diff
@@ -17,14 +19,17 @@ function inputListener({target}: Event): void {
 }
 
 function watchTextarea(textarea: HTMLTextAreaElement, {signal}: SignalAsOptions): void {
+	// Disable constrained native feature
```
fregante, line 22:
```suggestion
	// Disable constrained GitHub feature
```

PR #123, opened 2016-03-30
Move page detection checks to a separate file

hkdobrev:
Closes #120.

In this change I've tried to improve readability and consistency and not focus so much on performance as microoptimisations in these checks are not worth it.

Except for moving the checks to a separate file I've unified them a bit and made some fixes:

- simplified and unified dashboard regular expressions
- made all regex checks nested under a repo consistent
- re-used as much as possible especially for commit checks
- simplified repo root check
- fixed repo root check with `tree` in the regex for repositories like https://github.com/react-component/tree
- tried to use consistent naming and order of the methods
- fixed a few regular expressions not taking trailing slash in the URL into account

PR #7428, opened 2024-06-12
`suggest-commit-title-limit` - Enable on PR title

fregante:
- Closes https://github.com/refined-github/refined-github/issues/5375

## Test URLs

Here

## Screenshot
Screenshot 5


fregante:
cc @karlhorky

karlhorky:
Amazing, thanks!

Doesn't GitHub UI cut off after 50 characters? Would it make sense to make this configurable?

karlhorky:
More research:

On [this PR](https://github.com/refined-github/refined-github/pull/7238) it seems like the GitHub UI is truncating after 67 characters:

Screenshot 2024-06-12 at 15 46 26


```
> 'Revert "use-first-commit-message-for-new-prs - Preserve previous us'.length
67
```

fregante:
> Doesn't GitHub UI cut off after 50 characters?

Never, 50 is just a suggestion for commits titles.

72 is GitHub's actual limit (72 characters are allowed, less than that is clipped to 70 + ellipsis I guess)

The hard limit for PR titles is 256 characters: 

- https://github.com/refined-github/sandbox/issues/37

> Would it make sense to make this configurable?

No: https://github.com/refined-github/refined-github/wiki/%22Can-you-add-this-feature%3F%22#3-it-doesnt-require-options



> 67

That' missing the backticks used to turn part of the title into `title`, so 67 + 2 + `...` = 72.



---

As for the limit to use here, we actually had this issue before and it was the reason why we dropped this feature (we either had it or tried to add it before).

The issue: PR titles are used as commit titles _together with ` (#123)`_, so the limit should probably be `72 - 4 - String(prNumber).length`

fregante:
You can see it live here: https://github.com/refined-github/refined-github/issues/2178#issuecomment-505940703

karlhorky:
> That' missing the backticks used to turn part of the title into `title`, so 67 + 2 + `...` = 72.

Interesting, so that means GitHub is rendering an ellipsis character with the `...` but still counting all 3 characters 🤔 weird.

But good catch on the backticks, which I indeed did not count there.

> The issue: PR titles are used as commit titles _together with ` (#123)`_, so the limit should probably be `72 - 4 - String(prNumber).length`

Given the new information about the 72 characters, makes sense to me 👍

fregante:
> configure the limit

Will never happen, we just don't do that: https://github.com/refined-github/refined-github/wiki/%22Can-you-add-this-feature%3F%22#3-it-doesnt-require-options

I don’t think even GitHub cares enough about that to make it a per-repo limitation. If you want actual limits you'll have to use GitHub Actions and protected branches.

The scope of `suggest-commit-title-limit` is to avoid mistakenly exceeding this invisible 72-character limit that makes commits harder to read in a list when clipped. We had initially avoided limiting PR titles because there's no such limit _in PR titles._ I'm adding it now because PR titles eventually end up as part of the commit title anyway.

fregante:
Calculation now includes the PR number:

Untitled

fregante:
Sorry that it couldn't match the desired length count 🙏

karlhorky:
Nice, no worries - I can understand your logic with it.

Maybe 50-character commit title counts are outdated. 72 is much more roomy, and will probably make it easier to be expressive for complex changes without too much endless re-editing.

I'll consider switching for our team.

karlhorky:
@fregante it appears this isn't applying on the "Files changed" tab in PRs - `/pull/<number>/files` (maybe also not applying on other tabs)

Screenshot 2024-06-24 at 12 01 56

Would it be possible to also apply this to other tabs? I tried looking for a URL / regex in the diff, but I didn't see anything that would match the `/pull/<number>` pattern in this PR...

fregante:
Indeed! The check is called `isPRConversation`. It can be changed to `isPR`. PR welcome. I’m not sure if the other part of the feature will cause issues though, might be worth creating two `init` (`features.add` accepts multiple arguments)

karlhorky:
@fregante ok, maybe this is the one you mean?

- https://github.com/refined-github/refined-github/pull/7484

fregante:
Thank you!

karlhorky:
I just realized today, another place that this is missing is the "create new PR" URL on the compare page:

```html
<div class="discussion-topic-header p-0 mb-3">
  <h3 id="pull_request_title_header" data-view-component="true" class="f4 mb-2">Add a title</h3>
  <text-expander keys=":" data-emoji-url="/autocomplete/emoji" class="position-relative">
    <input class="form-control required title js-session-resumable js-quick-submit FormControl-input FormControl-large valid" placeholder="Title" aria-labelledby="pull_request_title_header" autofocus="autofocus" autocomplete="off" required="required" type="text" value="Remove obsolete Gatsby dev server try/catch" name="pull_request[title]" id="pull_request_title">
  </text-expander>
</div>
```

Would this also just be another item in the `include` array from the `pageDetect` object?

karlhorky:
@fregante created a draft PR here, in case `pageDetect.isCompare` would work here:

- #7689

Issue #7223, opened 2024-01-20
`default-branch-button` - z-index issue when commit is not in repo


fregante:
### Description

<img width="924" alt="Screenshot 22" src="https://github.com/refined-github/refined-github/assets/1402241/b36f8b78-5e07-4731-8e27-486fa8f3a07c">

### How to replicate the issue + URL

https://github.com/pixiebrix/pixiebrix-extension/blob/04df74beaa0a1dcba9abc498924bb83d11669872/src/blocks/renderers/documentView/DocumentViewLazy.tsx

### Extension version

23.12.1

### Browser(s) used

Safari

fregante:
This is the reason, from https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_positioned_layout/Understanding_z-index/Stacking_context

> A stacking context is formed, anywhere in the document, by any element in the following scenarios:
>
> - Element with a [container-type](https://developer.mozilla.org/en-US/docs/Web/CSS/container-type) value size or inline-size set, intended for [container queries](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_containment/Container_queries).


Their own element has the same bug:

![Untitled](https://github.com/refined-github/refined-github/assets/1402241/5ddb92ed-0c98-4078-8ca4-cb6cf83a6fdc)

However they can easily fix it by adding more DOM and `<tool-tip>` elements, which I haven't figured out yet.

This fixes it, but I'm afraid it'll cause more issues (and also the selector is too broad)

```css
.container {
    z-index: 1;
    position: relative;
}
```

<img width="340" alt="Screenshot 2" src="https://github.com/refined-github/refined-github/assets/1402241/646577f4-46a6-4cd3-bf18-781c3af41bd9">

PR #7430, opened 2024-06-12
`default-branch-button` - Fix overlap with other components

kovsu:
close: #7223 

## Screenshot

CleanShot 2024-06-12 at 16 46 38@2x

CleanShot 2024-06-12 at 16 46 32@2x

### DIFF
--- a/source/features/default-branch-button.css
+++ b/source/features/default-branch-button.css
@@ -3,9 +3,7 @@
 	max-width: 80px;
 }
 
-.container:has(> #StickyHeader .react-code-view-header-wrap--narrow):has(
-		#expand-button-file-tree-button
-	) {
+.rgh-default-branch-button-container {
 	position: relative;
 	z-index: 1;
 }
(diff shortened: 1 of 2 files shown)


fregante:
Indeed, I think this won't be possible due to :has(). We've been dealing with performance issues related to it so its usage should be limited.

- https://github.com/refined-github/refined-github/issues/7331#issuecomment-2120464245

kovsu:
So just use `#StickyHeader` ?

kovsu:
image

image

How about doing it this way? And it seems like this style only needs to be added on the `isSingleFile` page.

image

fregante:
I think that's fine 👍

fregante (approved):

fregante (approved):

fregante:
If you'd like, it'd be great to make the same change to `list-prs-for-file` to avoid having to use `:has` as well.

Issue #7424, opened 2024-06-12
`previous-version` broken


fregante:
### Description

See https://github.com/refined-github/refined-github/pull/6550#issuecomment-1537288308

<img width="1311" alt="Screenshot" src="https://github.com/refined-github/refined-github/assets/1402241/88a717bf-53e9-4680-8d3e-022f6832396a">

### How to replicate the issue + URL

It should appear on https://github.com/refined-github/refined-github/blob/main/contributing.md

### Extension version

24.6.9

### Browser(s) used

Chrome

PR #7431, opened 2024-06-12
`previous-version` - Restore feature

kovsu:
close: #7424 

## Screenshot

CleanShot 2024-06-12 at 17 01 17@2x


fregante:
Thank you! Would you be able to change the feature to include a label? I feel that it's nearly invisible there. This could be accomplished by using `historyButton.cloneNode(true)` and then changing the label and icon.

kovsu:
@fregante 

CleanShot 2024-06-12 at 22 26 17@2x

Maybe we just need add the label in the `a` tag?

fregante:
That works too 👍

PR #7433, opened 2024-06-12
`comment-excess` - New feature

fregante:
- Close https://github.com/refined-github/refined-github/issues/5764

## Test URLs

- https://togithub.com/prettier/prettier/issues/7475

## Screenshot

https://github.com/refined-github/refined-github/assets/1402241/66e55008-7e24-4c97-9a4c-effe00709975

### DIFF
--- a/build/__snapshots__/imported-features.txt
+++ b/build/__snapshots__/imported-features.txt
@@ -23,6 +23,7 @@ close-out-of-view-modals
 closing-remarks
 collapsible-content-button
 command-palette-navigation-shortcuts
+comment-excess
 comment-fields-keyboard-shortcuts
 comment-on-draft-pr-indicator
 comments-time-machine-links
--- a/source/features/comment-excess.tsx
+++ b/source/features/comment-excess.tsx
@@ -1,0 +1,49 @@
+import React from 'react';
+import * as pageDetect from 'github-url-detection';
+import elementReady from 'element-ready';
+import {expectElement} from 'select-dom';
+
+import features from '../feature-manager.js';
+import observe from '../helpers/selector-observer.js';
+import {assertNodeContent} from '../helpers/dom-utils.js';
+import {paginationButtonSelector} from '../github-helpers/selectors.js';
+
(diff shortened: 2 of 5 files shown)


fregante:
I left PRs out of this because 
- it's easy to reach a lot of (review) comments
- PRs are more rarely affected by this issue
- we already have performance issues on large PRs

fregante:
cc @jtagcat

jtagcat:
Is it shown after scrolling as well?

image

IMO I'd like to have it more visible, but I already know the button will exist + am here more as a bikeshed than contributing.

fregante:
No I tried to keep it simple for now. 

I also added an auto-scroll feature that will move the collapsed comments into view when the user presses Command-F, so that you will never accidentally search a page that has hidden items. This alone should cover the use case completely. 

I'm not entirely sure I like this piece of interaction, but I think absolutely no one wants to search a page that is incomplete. 

The alternative would be to show a toast instead.

fregante:
Let's see how it goes

monperrus:
Great feature, thanks a lot!

How to automatically execute this and show all hidden comments?

fregante:
We don't do that because most of the time you do not need to read every comment. It would increase load for GitHub unnecessarily

monperrus:
The use case in "open issue" >> CTRL-F to find something. When not finding, it always takes me some minutes to realize that the issue is incomplete due to hidden items.

fregante:
> it always takes me some minutes to realize

Are you using Refined GitHub? This feature will scroll to the hidden items when you press ctrl-f, it's already a solution to that specific issue

jtagcat:
Pardon that I didn't test this earlier, this has turned in to a blunder of a discussion.

Ctrl-f regression: top nav obscures the Load More button:
image

Could the Alt-Click be advertised on the button?

fregante:
What browser are you using?

jtagcat:
> What browser are you using?

FF 127.0.2 (64-bit)

PR #7434, opened 2024-06-12
`open-all-notifications` - Add keyboard shortcut: P

fregante:
- Closes https://github.com/refined-github/refined-github/issues/6273

I picked P because it's near O (the one used to open the focused item). Alt-Shift-O would have been an alternative too, but unnecessary at this point since P seems unused.

## Test URLs

## Screenshot

Screenshot 11

### DIFF
--- a/source/features/open-all-notifications.tsx
+++ b/source/features/open-all-notifications.tsx
@@ -9,6 +9,7 @@ import features from '../feature-manager.js';
 import openTabs from '../helpers/open-tabs.js';
 import {appendBefore} from '../helpers/dom-utils.js';
 import observe from '../helpers/selector-observer.js';
+import {multilineAriaLabel} from '../github-helpers/index.js';
 
 // Selector works on:
 // https://github.com/notifications (Grouped by date)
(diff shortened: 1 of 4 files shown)


source/features/open-all-notifications.tsx
```diff
@@ -69,7 +69,12 @@ function removeOpenUnreadButtons(container: ParentNode = document): void {
 
 function addSelectedButton(selectedActionsGroup: HTMLElement): void {
 	const button = (
-		<button className={'btn btn-sm mr-2 ' + openSelected.class} type="button">
+		<button
+			type="button"
+			className={'btn btn-sm mr-2 tooltipped tooltipped-s ' + openSelected.class}
+			data-hotkey="p"
+			aria-label="Open selected (keyboard shortcut: P)"
```
fregante, line 76: Good tooltip? Other items have the shortcut but no tooltip 🤷‍♂️ but I don't want this shortcut to just exist without anyone knowing. We should try to always make them visible.

fregante:
cc @0xdevalias

source/features/open-all-notifications.tsx
0xdevalias: It might be redundant given the context, but to be explicit, perhaps:

```suggestion
			aria-label="Open selected notification(s) (keyboard shortcut: P)"
```

fregante:
```suggestion
			aria-label="Open selected notifications%A0Hotkey: P"
```

Should appear on two lines

fregante:
Screenshot

PR #7437, opened 2024-06-12
Meta: Allow opening the options page via "Button Link"

fregante:
- Closes https://github.com/refined-github/refined-github/issues/7188

## Screenshot

https://github.com/refined-github/refined-github/assets/1402241/52f4fe48-f864-46d6-abc9-539e9fa8f342

### DIFF
--- a/source/options.html
+++ b/source/options.html
@@ -82,7 +82,8 @@
 	<details id="action">
 		<summary><strong>🔗 Button link</strong></summary>
 		<p>You can pick what page opens when you click the Refined GitHub icon in the browser toolbar</p>
-		<p><input type="url" name="actionUrl" placeholder="https://github.com" spellcheck="false" autocomplete="off" autocapitalize="off"></p>
+		<p><input type="text" name="actionUrl" placeholder="https://github.com" spellcheck="false" autocomplete="off" autocapitalize="off"></p>
+		<p>If you enter <strong>options</strong>, the options page will be opened on click.</p>
 	</details>
 
 	<details id="debugging">
--- a/source/options.html
+++ b/source/options.html
@@ -82,8 +82,8 @@
 	<details id="action">
 		<summary><strong>🔗 Button link</strong></summary>
 		<p>You can pick what page opens when you click the Refined GitHub icon in the browser toolbar</p>
-		<p><input type="text" name="actionUrl" placeholder="https://github.com" spellcheck="false" autocomplete="off" autocapitalize="off"></p>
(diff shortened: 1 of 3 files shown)


fregante:
Cc @venkatasg

source/options.html
```diff
@@ -82,7 +82,8 @@
 	<details id="action">
 		<summary><strong>🔗 Button link</strong></summary>
 		<p>You can pick what page opens when you click the Refined GitHub icon in the browser toolbar</p>
-		<p><input type="url" name="actionUrl" placeholder="https://github.com" spellcheck="false" autocomplete="off" autocapitalize="off"></p>
+		<p><input type="text" name="actionUrl" placeholder="https://github.com" spellcheck="false" autocomplete="off" autocapitalize="off"></p>
+		<p>If you enter <strong>options</strong>, the options page will be opened on click.</p>
```
FloEdelmann, line 86: I think it's a bit awkward to have that special `options` keyword. I'd suggest "Leave empty to open options on click.". `https://github.com` can be entered manually if you want that to be opened. Then the field can also stay `type="url"`.

fregante: The problem is then having to distinguish people who left it empty before and those who fill it after this change.

The options are managed by https://github.com/fregante/webext-options-sync and its migrations are not version-based

FloEdelmann: We could

1. rename the option and migrate users to the new option.
2. just live with that and tell people that the default behavior changed and that they can easily change that back in the options by entering `https://github.com/`.

fregante: Duh, why didn't I think of that 👍

fregante: Never mind, the problem is the same even if I change the name, because:

- the value of the old key will be undefined on the first run and it will be undefined after the migration
- the module provides the default for the new key even if unset


In short I get a `undefined, https://github.com` pair on every run, regardless of migrations or first install. 

I don't want to actually change the default and open the options first, because that's useless to most users. The reason this exists at all is so that users have a reason to keep the extension icon visible; opening GitHub is useful. 

I might just have to introduce an external key and mark the migration as done for the next few months. Let's see.

fregante: Never mind x2, this is a non-issue. 😂 

I think that this works by just setting the default in the schema (rather than in the click handler)

- Unset users will just continue to open GitHub, because the default is effectively unchanged 
- Users with explicit GitHub.com will also be unchanged

The only difference is that the default *field* was empty, but now it's set. That's all. At least that's what it looks like in my tests.

PR #7444, opened 2024-06-13
Accept OAuth tokens in the options

tonglil:
Allow users to use an GitHub OAuth token (`gho_...`). Right now the form validation regex prevent me from even using it.

For example, I am using the one from the GitHub CLI (`gh auth token`). GitHub's OAuth app has as much permissions as one can grant and should be sufficient for refined-github usage.

My organization does not allow PATs (`ghp_...`) to access GH.

Related to:
- https://github.com/refined-github/refined-github/issues/6951
- https://github.com/refined-github/refined-github/issues/6092#issuecomment-1732286938

## Test URLs

n/a

## Screenshot

n/a

### DIFF
--- a/source/options.html
+++ b/source/options.html
@@ -23,7 +23,7 @@
 		<p>
 			<!-- placeholder is set to enable use of :placeholder-shown CSS selector -->
 			<!-- https://gist.github.com/magnetikonline/073afe7909ffdd6f10ef06a00bc3bc88 -->
-			<input type="text" name="personalToken" pattern="[\da-f]{40}|ghp_\w{36,251}|github_pat_[a-zA-Z0-9]{22}_[a-zA-Z0-9]{59}" placeholder=" " spellcheck="false" autocomplete="off" autocapitalize="off">
+			<input type="text" name="personalToken" pattern="[\da-f]{40}|gh[po]_\w{36,251}|github_pat_[a-zA-Z0-9]{22}_[a-zA-Z0-9]{59}" placeholder=" " spellcheck="false" autocomplete="off" autocapitalize="off">
 			<span id="validation"></span>
 		</p>
 		<ul data-token-type="classic">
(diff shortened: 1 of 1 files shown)


fregante:
Interesting. How do you generate this token? Do you have to create your own dedicate GitHub app? I'd want to document this in the Security page of the wiki as well.

fregante:
I dropped the pattern validation altogether, we already verify that the token is valid directly with GitHub.

Can you edit the Wiki to mention this? I don't know if you have access to it

tonglil:
Thanks for merging.

> Interesting. How do you generate this token?

I am using the token outputted by `gh auth token` obtained through `gh auth login`, which by default authenticates via web flow with a GitHub owned OAuth app for the `gh` [cli](https://cli.github.com/) ([docs](https://cli.github.com/manual/gh_auth_login)).

image

> Do you have to create your own dedicate GitHub app?

If someone _wants_ to generate OAuth tokens from their own GitHub OAuth App, they could. They would be in control of their OAuth App's scopes, which would limit the scopes of the tokens generated from the App.

> Can you edit the Wiki to mention this? 

Sure, I can add this below "fine grained tokens":

```
GitHub OAuth App Tokens are supported as long as they have the required scopes and the GitHub OAuth App is granted access to the organizations your want to access.

Using a GitHub OAuth App Token can be an alternative to PATs if your organization restricts PAT usage and approves the GitHub OAuth App. For example, sourcing the token used by the official [GitHub CLI](https://cli.github.com/) with `gh auth token`.
```

What do you think?

fregante:
Try and let me know if it lets you. 👍 I'll have to tweak the page and error messages to clarify that organizations *are* supported with this token.

tonglil:
Actually I don't have permission to write/edit the wiki.

fregante:
I reworded it and [added it](https://github.com/refined-github/refined-github/wiki/Security). Thank you again for this piece of information! I also have to alter the `options.html` page in this repo to mention support.

Issue #7447, opened 2024-06-14
TypeError: document.body is null


Enubia:
### Description

extension errors out on initialization with `TypeError: document.body is null`

### How to replicate the issue + URL

everywhere on github

### Extension version

24.6.13.1731

### Browser(s) used

Firefox 126.0.1

![image](https://github.com/refined-github/refined-github/assets/44695100/d0e96f5e-f425-44c7-8e23-12731895e847)

Issue #7446, opened 2024-06-14
`quick-review` Feature broken?


huehnerlady:
### Description

Since the latest update of my Brave browser and refined github, I do not have the shortcuts for approving a pull request above the requested reviewer, nor does the `v` shortcut work anymore.

![image](https://github.com/refined-github/refined-github/assets/12533979/efa0ee60-c375-4dc0-9edc-49276de6c021)

When I go to the review area manually, I can see the change in the buttons:
![image](https://github.com/refined-github/refined-github/assets/12533979/e40a76f4-48f1-4296-8929-d9f39c4d93f8)

So it seems half of the feature is loaded?

I also tried Chrome Version 126.0.6478.62 (Official Build) (arm64) and I have the same problem there

### How to replicate the issue + URL

open a pull request for review with Brave browser

https://github.com/huehnerlady/Test-refined-github/pull/1

### Extension version

24.6.13.1731

### Browser(s) used

Brave Version 1.67.115 Chromium: 126.0.6478.56 (Official Build) (arm64)

PR #7450, opened 2024-06-14
Restore extension

kovsu:
close: #7447
close: #7446

CleanShot 2024-06-14 at 16 27 15@2x


kovsu:
feature broken:

image

after:

fregante:
Actually there's a better way nowadays: https://developer.mozilla.org/en-US/docs/Web/API/CSS/supports_static

fregante:
Thanks @kovsu 🙏

PR #7451, opened 2024-06-14
`list-pr-for-files` - Improve performance

kovsu:


fregante:
How does it look? I'll merge soon

kovsu:
@fregante I'm sorry for replying so late. It looks fine.

CleanShot 2024-06-16 at 23 26 40@2x

PR #7458, opened 2024-06-14
feat: add open-notification-external feature

kecrily:
After returning from a vacation, we may have a lot of notifications. This time opening the notifications as new tabs will help us to deal with them one by one like a to-do list.

## Test URLs

<https://github.com/notifications>

## Screenshot

N/A


fregante:
Thanks for the PR but features need to be discussed first. Opening links in new tabs is a bad default to begin with https://www.nngroup.com/articles/new-browser-windows-and-tabs/

The good part is that Refined GitHub already has `open-all-notifications` which lets you open all unread notifications in new tabs _or_ any selected notification.

PR #7460, opened 2024-06-14
`rgh-netiquette`: Add "Don't comment" banner

its-miroma:
In some issues, the author wrote "Please do not leave comments", because those issues are not meant for discussion.

This PR adds a banner to issues with such requests in `rgh-netiquette`.

The `/regex/` may not be perfect, please test if it works on other issues

## Test URLs

- https://github.com/refined-github/refined-github/issues/7404
- https://github.com/refined-github/refined-github/issues/7000
- https://github.com/refined-github/refined-github/issues/6000

## Screenshot

A lot of cake

### DIFF
--- a/source/features/netiquette.tsx
+++ b/source/features/netiquette.tsx
@@ -16,7 +16,7 @@ import {canEditEveryComment} from './quick-comment-edit.js';
 // TODO: Not exact, replace with API
 const isCollaborator = canEditEveryComment;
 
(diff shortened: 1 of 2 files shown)


fregante:
Thank you for the interest! However this is extremely rare so it's not worth running this regex on every single page load. On 7000 issues/PRs we only have like 6 issues where I said "don't leave comments"

PR #7462, opened 2024-06-15
`open-with-buttons` - New feature

pl4nty:
The "Code" dropdown in a repository's root has buttons to open the repository in GitHub Desktop or Visual Studio. This feature adds more buttons for other external apps like Visual Studio Code.

To start with, I've added two buttons:
* [Visual Studio Code Dev Containers](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers)
* [Visual Studio Code remote repositories](https://marketplace.visualstudio.com/items?itemName=ms-vscode.remote-repositories)

This feature could lead to requests for other external apps, so an opt-in approach might be useful to avoid overfilling the "Code" dropdown. Not sure of the best way to structure this.

## Test URLs
https://github.com/refined-github/refined-github/tree/main
https://github.com/refined-github/refined-github/tree/ad0054477020c52e7140278200be00fc764cc469

## Screenshot
buttons

### DIFF
--- a/build/__snapshots__/features-meta.json
+++ b/build/__snapshots__/features-meta.json
@@ -530,6 +530,11 @@
 		"screenshot": "https://github-production-user-asset-6210df.s3.amazonaws.com/83146190/261159396-0610574b-ab1f-42fb-813f-ee7310a1e5b6.png"
 	},
 	{
+		"id": "open-with-buttons",
+		"description": "Adds buttons to open a repository in external apps like Visual Studio Code.",
+		"screenshot": "https://github.com/refined-github/refined-github/assets/21111317/13516c22-585f-440e-b4de-53be95995af7"
+	},
+	{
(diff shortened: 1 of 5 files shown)


fregante:
> This feature could lead to requests for other external apps

Correct, there are plenty such tools, with dedicated browser extensions that add these buttons. For this also there are more appropriate extensions already:

https://chromewebstore.google.com/search/GitHub%20vscode

> so an opt-in approach might be useful to avoid overfilling the "Code" dropdown.

Refined GitHub doesn't do opt-in:

https://github.com/refined-github/refined-github/wiki/%22Can-you-add-this-feature%3F%22

pl4nty:
Thanks, I missed that wiki page. I haven't found any extensions that implement these buttons though, particularly the Dev Container option, instead they just redirect to vscode.dev. I'll just write my own extension

Issue #6686, opened 2023-05-23
Extra more GQL queries to their own files


fregante:
Continuing the work from:

- https://github.com/refined-github/refined-github/issues/5848
- https://github.com/refined-github/refined-github/pull/6674

To do, extracted from the last PR:

> - [ ] TypeScript integration that will help us use the results in the file
> - [ ] GraphQL autocomplete in the IDE thanks to the schema
> - [ ] Typed `variables` that make TypeScript complain if one is missing or unused
> - [ ] Prettier/formatter for GQL support (we have a .prettierignore file)

> Extensions suggested: 
> - https://chrome.google.com/webstore/detail/graphql-network-inspector/ndlbedplllcgconngcnfmkadhokfaaln
> - https://marketplace.visualstudio.com/items?itemName=GraphQL.vscode-graphql-syntax

fregante:
Potential setup:

- webpack: https://www.apollographql.com/docs/react/integrations/webpack/ (allows storing multiple queries in one file)
- typescript: https://the-guild.dev/graphql/codegen/plugins/typescript/typescript-graphql-files-modules
- eslint: https://the-guild.dev/graphql/eslint/docs (schema validation)
- schema: https://the-guild.dev/graphql/config/docs

PR #7472, opened 2024-06-19
Replace Webpack with Rollup

### LINKED ISSUE #7471: Build broken on Node 22 Windows
Seen in https://github.com/refined-github/refined-github/pull/7470

It looks like webpack can't find the config file on Windows 🤷‍♂️ 

```
> x:webpack
> cross-env NODE_OPTIONS="--import tsx --no-warnings=ExperimentalWarning" webpack --mode=production webpack-stats.json

[webpack-cli] Invalid configuration object. Webpack has been initialized using a configuration object that does not match the API schema.
 - configuration.entry[0] should be a non-empty string.
   -> A module that is loaded upon startup. Only the last one is exported.
```

fregante:
- Closes #7471 
- Related to https://github.com/refined-github/refined-github/pull/6609
- Ideally followed up by https://github.com/rollup/rollup/issues/5626

I'm interested in Rollup's ability to spit out independent files instead of a single bundle.

~~This is not done yet, it's having issues with the .d.ts import https://github.com/rollup/plugins/issues/992~~


fregante:
```
[!] (plugin readme-plugin) Error: Could not load ./refined-github/readme.md (imported by source/options.tsx): Cannot find module './refined-github/readme-parser-bhiWxXuB.js' imported from ./refined-github/rollup.config-1718791086347.mjs
```

🤷‍♂️ 

The typescript converter spits out the wrong path for the generated file, even if the `import()` for the parser is correct. 🤦‍♂️

build/readme-parser.js
fregante, line 1: I got it to build by turning this TS file into JS. Not ideal but not a big loss.

rollup.config.js
```diff
@@ -0,0 +1,43 @@
...
+	output: {
+		dir: 'distribution/assets',
+		// AssetFileNames: "[name]-[hash][extname]", // for CSS
+	},
+	plugins: [
+		json(), // TODO: Drop after https://github.com/refined-github/shorten-repo-url/issues/47
+		styles(),
+		string({
+			include: '**/*.gql',
+		}),
```
fregante, line 28: This could use a GQL-specific plugin in the future, but for now it's 1:1 for easier upgrade.

- https://github.com/rollup/plugins/tree/master/packages/graphql
- #6686

rollup.config.js
```diff
@@ -0,0 +1,43 @@
...
+		alias({
+			entries: [
+				{find: 'react', replacement: 'dom-chef'},
+				{find: '@cheap-glitch/mi-cron', replacement: '@cheap-glitch/mi-cron/index.min.js'},
```
fregante, line 32: Rollup is picking up the .ts file to read the exported values (ESM) and then using the `main` entry point (CJS) 🤦‍♂️ 🤷‍♂️ 

Could be solved by:

- https://github.com/cheap-glitch/mi-cron/issues/10

.github/workflows/test.yml
```diff
@@ -65,13 +65,6 @@ jobs:
           node-version-file: package.json
           cache: npm
       - run: npm ci
-      - run: npm run build:webpack -- --json webpack-stats.json
-      - name: Send webpack stats to RelativeCI
-        if: matrix.os == 'ubuntu-latest'
-        uses: relative-ci/agent-action@v2
-        with:
-          webpackStatsFile: ./webpack-stats.json
-          key: ${{ secrets.RELATIVE_CI_KEY }}
```
fregante, line 74: Probably:

- https://github.com/relative-ci/rollup-plugin-webpack-stats

fregante (commented):
The extension is completely unbundled now 🎉

Screenshot 2

It looks like I'm missing:

- [ ] CSS extraction
- [x] manifest.json copy to dist
- [ ] Restore RelativeCI stats
- [ ] Ensure it still works

fregante:
So close, yet so far.

CSS files are extracted, but `preserveModules: true` _also_ causes CSS files to not be bundled, but we don't want that (rather, it does _not_ preserve `@import`s to these files, which are now orphaned)

fregante:
Wah… have we finally reached this point? It appears to work 😱 

I've been trying to move off of Webpack for years and we've finally come full circle:

- https://github.com/refined-github/refined-github/pull/577

Complexity-wise, the advantages of Rollup over Webpack are minor or non-existent, but we gain the ability to output separate files. This means we will be able to drop hacks like `id: import.meta.url` and `features.log.error(import.meta.url)`. Also any uncaught errors are no longer attached to `refined-github.js` but to the specific file (or feature) where they are thrown.

These improvements can come later, when our new Rollup build has been validated.

fregante:
A couple of warnings:


```

(!) "this" has been rewritten to "undefined"
https://rollupjs.org/troubleshooting/#error-this-is-undefined
node_modules/webext-options-sync-per-domain/index.js
1: var __decorate = (this && this.__decorate) || function (decorators, target, key, desc) {
                     ^
2:     var c = arguments.length, r = c < 3 ? target : desc === null ? desc = Object.getOwnPropertyDescriptor(target, key...
3:     if (typeof Reflect === "object" && typeof Reflect.decorate === "function") r = Reflect.decorate(decorators, targe...
...and 1 other occurrence
(!) Circular dependency
source/github-helpers/github-token.ts -> source/github-helpers/api.ts -> source/github-helpers/github-token.ts
```

fregante:
@kidonng @yakov116 does anyone of you run a dev version daily? It would be good to test-drive this version as your daily version.

I didn't see issues but I'm nervous to merge it 😅

yakov116:
I can use it for the next few days

fregante:
SHIPPING TIME

PR #7473, opened 2024-06-19
Migrate to MV3

### LINKED ISSUE #5531: Migrating to Manifest V3

fregante:
- Close #5531 

Thanks to all the work I've done supporting my [`webext-*` modules](https://github.com/fregante/webext-fun) for PixieBrix, updating Refined GitHub meant changing only a few keys in the manifest. Now I just need to check what this means for GH Enterprise users.

So this upgrade is **sponsored by [PixieBrix](https://www.pixiebrix.com)** :tada:


source/manifest.json
```diff
@@ -17,23 +18,28 @@
 	},
 	"permissions": [
 		"storage",
+		"scripting",
```
fregante, line 21: Explanation for the Chrome Web Store:

> The scripting API is used to inject the content script into new hosts when granted by the user.

source/manifest.json
```diff
@@ -17,23 +18,28 @@
...
+	"host_permissions": [
 		"https://github.com/*",
 		"https://api.github.com/*"
 	],
-	"browser_action": {
+	"action": {
 		"default_icon": "assets/icon.png"
 	},
-	"optional_permissions": [
+	"optional_host_permissions": [
```
fregante, line 33: Firefox _still_ doesn't support this. The good news is that it should be out on July 9th. I'll wait until then to merge this PR.

https://bugzilla.mozilla.org/show_bug.cgi?id=1766026

source/manifest.json
```diff
@@ -57,6 +63,9 @@
 		}
 	],
 	"web_accessible_resources": [
-		"assets/resolve-conflicts.js"
-	]
+    {
+      "resources": [ "assets/resolve-conflicts.js" ],
+      "matches": [ "https://*/*" ]
```
fregante, line 68: Also verified to work on https://github.com/refined-github/sandbox/pull/82/conflicts

Screenshot

fregante (commented):
Blocked by Firefox. v128 will be out on July 9th.

fregante (commented):
- Added GHE support for Firefox <128 so I don't have to wait. I tested the "invalid" `optional_permissions` array on the Chrome Web Store and it was fine
- Fixed a bug where it would be injected multiple times on GHE as the worker restarts

source/manifest.json
```diff
@@ -17,23 +18,31 @@
 	},
 	"permissions": [
 		"storage",
+		"scripting",
 		"contextMenus",
 		"activeTab",
```
willdurand, line 23: Why is this permission required? There is a side-effect with Firefox MV3: this extension now "needs attention" on every tab that isn't GitHub.

Screenshot 2024-06-24 at 12 26 19

fregante: That's not it, see #7477 

The permission is required in MV3 to add support for GitHub Enterprise

Issue #7477, opened 2024-06-21
Extension demands access on all websites (Firefox)


HanabishiRecca:
### Description

Using Firefox, after 24.6.21 update the extension demands access on all websites. I guess it's the Manifest V3 thing?

![image](https://github.com/refined-github/refined-github/assets/13597663/07e080a3-bbd2-4234-b312-2671c29c883f)
![image](https://github.com/refined-github/refined-github/assets/13597663/605a82e6-6180-40ac-a831-e8ab88a83653)

### How to replicate the issue + URL

1. Open any unrelated website.
2. Observe the extension icon.

### Extension version

24.6.21

### Browser(s) used

Firefox 127.0.1

fregante:
Thank you for the report! Unfortunately this is a Firefox bug so there's nothing we can do here.

I'll keep this issue open just for reference.

Bug described at:

- https://discourse.mozilla.org/t/permission-needed-indicator-on-browser-action/123183


Bug tracked at: (10 months old)

- https://bugzilla.mozilla.org/show_bug.cgi?id=1851083

fregante:
# 24.6.14 (manifest version 2)

<img width="386" alt="Screenshot 3" src="https://github.com/refined-github/refined-github/assets/1402241/1f87f5f0-0b14-4059-be64-70136caefa49">

# 24.6.21 (manifest version 3)

<img width="386" alt="Screenshot 4" src="https://github.com/refined-github/refined-github/assets/1402241/3841b793-64cd-410f-8167-2d5e8bf14711">

fregante:
Until that Firefox bug is resolved, I think we could just ship MV2 to Firefox.

We can add a small Node script to reverse the manifest.json changes from https://github.com/refined-github/refined-github/pull/7473. The rest of the code already works regardless of the manifest version.

That indicator is complete nonsense and I don't want to bother Refined GitHub users unnecessarily.

HanabishiRecca:
Yeah, if it is not too cumbersome to maintain. Unlike Chrome, Firefox is not planning on dropping V2.

INSANEMODE:
> Until that Firefox bug is resolved, I think we could just ship MV2 to Firefox.
> 
> We can add a small Node script to reverse the manifest.json changes from #7473. The rest of the code already works regardless of the manifest version.
> 
> That indicator is complete nonsense and I don't want to bother Refined GitHub users unnecessarily.

[...]

PR #7480, opened 2024-06-23
Improve support for mobile Firefox

INSANEMODE:
1. No relevant issue found
## Description
This change is made to detect if the user is on a mobile platform (ios or android), instead of just looking for mobile Safari, or explicitly checking for user agents of mobile browsers.

Mobile safari, and other mobile browsers like Kiwi and Firefox for Android do not support context menus. 
This change allows android browsers to access the options page.

`getPlatformInfo` seems to have fairly wide browser support, so it should be fine to use, instead of checking the user agent, like `webext-detect-page` does. 

https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/getPlatformInfo

This method is used, because `webext-detect-page` does not seem to have functions to detect mobile browsers other than Safari.

This has only been tested on Firefox for android, and Firefox on windows. Kiwi on android should behave similarly, but i do not have an ios device to test mobile Safari.

## Test URLs
Affects browser action button on mobile browsers, not URLs.

## Screenshot
Screenshot 2024-06-23 044312

### DIFF
--- a/source/background.ts
+++ b/source/background.ts
@@ -15,8 +15,11 @@ import {styleHotfixes} from './helpers/hotfix.js';
 
 const {version} = browser.runtime.getManifest();
 
+if(!doesBrowserActionOpenOptions)
+{
 // GHE support
-addPermissionToggle();
+	addPermissionToggle();
+}
 
 const messageHandlers = {
 	async openUrls(urls: string[], {tab}: Runtime.MessageSender) {
--- a/source/background.ts
+++ b/source/background.ts
@@ -15,11 +15,8 @@ import {styleHotfixes} from './helpers/hotfix.js';
 
 const {version} = browser.runtime.getManifest();
 
-if(!doesBrowserActionOpenOptions)
-{
 // GHE support
-	addPermissionToggle();
-}
+addPermissionToggle();
 
 const messageHandlers = {
 	async openUrls(urls: string[], {tab}: Runtime.MessageSender) {
--- a/package-lock.json
+++ b/package-lock.json
@@ -47,7 +47,7 @@
 				"webext-detect-page": "^5.0.1",
 				"webext-dynamic-content-scripts": "^10.0.3",
 				"webext-options-sync-per-domain": "^4.2.2",
-				"webext-permission-toggle": "^5.0.1",
+				"webext-permission-toggle": "^5.0.2",
 				"webext-storage-cache": "^6.0.1",
 				"zip-text-nodes": "^1.0.0"
 			},
@@ -11013,9 +11013,10 @@
 			}
 		},
 		"node_modules/webext-permission-toggle": {
-			"version": "5.0.1",
-			"resolved": "https://registry.npmjs.org/webext-permission-toggle/-/webext-permission-toggle-5.0.1.tgz",
-			"integrity": "sha512-7sMgpgdmKhKZr6zYsBR+RmXUfL8Ahr2fRDuGEKomZDUQC5ZhpmdrvfNU5ERnPU6F604FMB3kj71Q+scGJ6thzw==",
+			"version": "5.0.2",
+			"resolved": "https://registry.npmjs.org/webext-permission-toggle/-/webext-permission-toggle-5.0.2.tgz",
+			"integrity": "sha512-PQOfyW1qnE+TKjFtByCduO4uKZhrBZpiGSDWEPRWp9+7cSjw70I2P1NbUZI25I5RaNdYR0nQMVVZbZ6Q2LXUng==",
+			"license": "MIT",
 			"dependencies": {
 				"webext-alert": "^1.0.0",
 				"webext-content-scripts": "^2.6.1",
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
 		"webext-detect-page": "^5.0.1",
 		"webext-dynamic-content-scripts": "^10.0.3",
 		"webext-options-sync-per-domain": "^4.2.2",
-		"webext-permission-toggle": "^5.0.1",
+		"webext-permission-toggle": "^5.0.2",
 		"webext-storage-cache": "^6.0.1",
 		"zip-text-nodes": "^1.0.0"
 	},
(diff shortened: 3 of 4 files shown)


source/background.ts
```diff
@@ -15,8 +15,11 @@ import {styleHotfixes} from './helpers/hotfix.js';
 
 const {version} = browser.runtime.getManifest();
 
+if(!doesBrowserActionOpenOptions)
+{
 // GHE support
-addPermissionToggle();
+	addPermissionToggle();
```
fregante, line 21: 👌

fregante (commented):
Good idea. I think the detection should look like "does this device have context menus"? This should probably also cover iPadOS. Ideally the chrome API can answer this question, but further exploration would be needed.

source/background.ts
INSANEMODE: Hm, i don't believe `browser.contextMenus` is defined on browsers that don't support them, so maybe a check to see if it is undefined or not could be useful, if there isn't a chrome api solution available?

fregante: That appears to be accurate: https://github.com/fregante/webext-permission-toggle/pull/26

So you can just drop the mobile/safari check and replace it with `!chrome.contextMenus`

`addPermissionToggle` has its own check inside so this specific change isn't needed here

INSANEMODE: The following error gets thrown on Firefox for android with the `doesBrowserActionOpenOptions` check removed on `addPermissionToggle`


```
00:13:02.267 Uncaught Error: webext-permission-toggle requires the `contextMenus` permission
    addPermissionToggle moz-extension://24fb74ed-aa72-4132-a59d-33a31291dced/assets/background.js:5183
    <anonymous> moz-extension://24fb74ed-aa72-4132-a59d-33a31291dced/assets/background.js:5857
    <anonymous> moz-extension://24fb74ed-aa72-4132-a59d-33a31291dced/assets/background.js:5961
    <anonymous> moz-extension://24fb74ed-aa72-4132-a59d-33a31291dced/assets/background.js:5963
background.js:5183:15
    addPermissionToggle moz-extension://24fb74ed-aa72-4132-a59d-33a31291dced/assets/background.js:5183
    <anonymous> moz-extension://24fb74ed-aa72-4132-a59d-33a31291dced/assets/background.js:5857
    <anonymous> moz-extension://24fb74ed-aa72-4132-a59d-33a31291dced/assets/background.js:5961
    <anonymous> moz-extension://24fb74ed-aa72-4132-a59d-33a31291dced/assets/background.js:5963
```

This prevents the action button from doing anything when pressed.

fregante: Thanks for checking 🙏 

What does the `permissions` array looks like there? The getter should return the manifest as is, is Firefox altering the manifest and removing the permission that ***does exist*** in the manifest?

Can you try `console.log(chrome.runtime.getManifest())`?

fregante: According to [MDN](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/getManifest), the API:

> Returns null for missing values, and removes unsupported keys

This means that it probably removes the `"contextMenus"` permission. This makes it difficult to detect when the author simply forgot to add it. The safest way would be to use `fetch(chrome.runtime.getURL('manifest.json'))` and read it directly, but that's annoying and makes the function async :(

fregante: - I opened https://github.com/fregante/webext-permission-toggle/issues/52

INSANEMODE: on firefox for android, the contextMenu permission isn't a valid permission, so it is just seems to ignore it if it is in the manifest.

Screenshot 2024-06-24 005921


Screenshot 2024-06-24 010001

INSANEMODE: Screenshot 2024-06-24 010125

fregante (approved):
This is good here, I'll send a PR to the other package and then publish it. It can then be installed here and merged

fregante:
Done https://github.com/fregante/webext-permission-toggle/releases/tag/v5.0.2

Does it work in both FirefoxAndroid and Kiwi?

fregante:
I just realized, if `addPermissionToggle` was throwing an error, every feature that depended on the background script, including options migrations, etc, were broken until now. So this PR would actually fix a user-facing bug.

INSANEMODE:
> Done [fregante/webext-permission-toggle@`v5.0.2` (release)](https://github.com/fregante/webext-permission-toggle/releases/tag/v5.0.2)
> 
> Does it work in both FirefoxAndroid and Kiwi?

Yep, it works on firefox for android with that change. 

Attempted to load it on kiwi, and it looks like it can't install the extension at the moment, probably due to this issue. https://github.com/kiwibrowser/src.next/issues/1138

fregante:
Thank you! If you’re installing a local build, you can change the manifest locally and double-check. But yeah they still wouldn’t be able to install 24.6.21 and future updates. 

However… as part of #7477 I can remove the `script` from the manifest in the Chrome build and that’ll restore support for Kiwi.

INSANEMODE:
> Thank you! If you’re installing a local build, you can change the manifest locally and double-check. But yeah they still wouldn’t be able to install 24.6.21 and future updates.
> 
> However… as part of #7477 I can remove the `script` from the manifest in the Chrome build and that’ll restore support for Kiwi.

removing `background.script` did allow me to install the extension on kiwi.

when tapping the action button, it takes me to github, like the desktop version does. I guess the contextMenus check doesn't work as intended on kiwi, but the background script isn't getting killed, at least.

Maybe it supports the contextMenus permission and chrome.contextMenus isn't undefined, but It doesn't seem like long pressing the button or anything triggers it. 

Heading to bed, so i won't be able to narrow down that issue, at the moment.

fregante:
A couple of things:

- Kiwi browser hasn't seen commits, let alone releases in 10 months. That's a long time for browsers: https://github.com/kiwibrowser
- Nobody knows how to trigger context menus, and probably there isn't a way other than to sniffing the user agent: https://www.reddit.com/r/kiwibrowser/comments/11v9ycq
- Nobody ever asked about it on their repo either: https://github.com/search?q=owner%3Akiwibrowser+contextmenu&type=issues

The lack of updates also explains why a v121 feature still isn't available (`background.script` on MV3).

I'm going to keep dropping `background.script` on MV3 for now, but I won't do any further work for a doubly-unsupported browser (unsupported by them, unsupported by us)

PR #7482, opened 2024-06-24
Drop `webextension-polyfill`

fregante:
- Follows #7473

### DIFF
--- a/source/background.ts
+++ b/source/background.ts
@@ -54,6 +54,7 @@ chrome.runtime.onMessage.addListener((message: typeof messageHandlers, sender): 
 	}
 });
 
+// `browserAction` needed for Firefox MV2 https://github.com/refined-github/refined-github/issues/7477
 (chrome.action ?? chrome.browserAction).onClicked.addListener(async tab => {
 	if (doesBrowserActionOpenOptions) {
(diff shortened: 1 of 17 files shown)


source/background.ts
```diff
@@ -40,36 +38,36 @@ const messageHandlers = {
...
+chrome.runtime.onMessage.addListener((message: typeof messageHandlers, sender): Promise<unknown> | void => {
 	for (const id of objectKeys(message)) {
 		if (id in messageHandlers) {
 			return messageHandlers[id](message[id], sender);
 		}
 	}
 });
 
-browser.action.onClicked.addListener(async tab => {
+(chrome.action ?? chrome.browserAction).onClicked.addListener(async tab => {
```
fregante, line 58: This will be needed for MV2 Firefox once this is completed:

- #7477

webpack.config.ts
```diff
@@ -55,9 +55,6 @@ const config: Configuration = {
 	},
 	plugins: [
 		new MiniCssExtractPlugin(),
-		new webpack.ProvidePlugin({
-			browser: 'webextension-polyfill',
-		}),
```
fregante, line 60: This also helps:

- #7472

Issue #7404, opened 2024-06-05
Performance issue on PRs with lots of changes (4th issue)


fregante:
This issue is for **version 24.6.14** only, make sure you're using that before interacting.

If you have performance issues[,](https://github.com/refined-github/refined-github/issues/7316) react with 👍 to browser below. Chrome, Firefox, or Safari.

**Please do not leave "me too" comments**

fregante:
Firefox

fregante:
Firefox and other forks

fregante:
Safari and iOS Safari

fregante:
Chrome and forks

fregante:
Doesn't seem too bad. If you do encounter substantial slowdowns, please leave a comment detailing how to reproduce (URL, browser, Refined GitHub version, etc). Make sure Refined GitHub is actually causing them.

I'll look into the other cause of this slowdown hopefully soon:

- https://github.com/refined-github/refined-github/issues/6437

PR #7487, opened 2024-06-25
`show-whitespace` - Restore feature on Firefox and Safari

fregante:
- Partially reverts https://github.com/refined-github/refined-github/pull/7315
- Related to #7404


cooljeanius:
So if I'd manually toggled this feature previously before it was removed, will it automatically be re-enabled, or will that still require further manual interaction?

fregante:
The way this was disabled was only in its init, unlike the usual hotfixes which appear in the options. In short you could toggle it on and off in the options but it wouldn't run anyway. 

To answer your question: no further action is necessary.

PR #7488, opened 2024-06-25
Drop Node 0.4 dependencies

fregante:
See https://github.com/SukkaW/nolyfill

Let's see how this works


fregante:
It doesn't seem to affect the final bundle (presumably they're all in the dev dependencies) but it brings the total number of installed packages from 802 to 734.

However I'm not a fan of `@latest` 

- https://github.com/SukkaW/nolyfill/issues/73

fregante:
Just doing my part to avoid trash packages on npm. Wish it could be better.

PR #7494, opened 2024-06-26
Larger text size in the options

### LINKED ISSUE #7483: Increase options page font size
### Description

Frist, congratulations on the MV3 upgrade 🎉

The options page now (24.6.21) opens in a new tab in Chrome, and has this miniscule font size that IMO looks weird (in a standalone page) not suitable for our options page packed with dense feature descriptions:

Screenshot 2024-06-24 at 1 51 34 PM

(Compare with browser UI)

Not sure if this is a Chrome only thing, but I'd like we bump it up to at least 14px compare to 75% (12px):

Screenshot 2024-06-24 at 1 47 44 PM

### Example URLs

chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/assets/options.html

kidonng:
Resolve #7483

I only tested this in (desktop) Chrome since the smaller font seems to be a Chrome only thing. ~~In other browsers the change should make no difference.~~

I do anticipate we need to tweak a few other sizes to make it look perfect (see the logo for example).

## Test URLs

chrome-extension://<RGH>/assets/options.html

## Screenshot

Screenshot 2024-06-27 at 3 37 47 PM

### DIFF
--- a/source/options.css
+++ b/source/options.css
@@ -25,6 +25,7 @@ html::after {
 /* Use sensible font size in Chrome #7483 */
 body {
 	font-size: 100%;
+	line-height: 1.5;
 }
 
 p {
--- a/source/options.css
+++ b/source/options.css
@@ -22,6 +22,11 @@ html::after {
 	}
 }
 
+/* Use sensible font size in Chrome #7483 */
+body {
+	font-size: 100%;
+}
+
 p {
 	margin-top: 0;
 }


source/options.css
```diff
@@ -22,6 +22,11 @@ html::after {
 	}
 }
 
+/* Use sensible font size in Chrome #7483 */
+body {
+	font-size: 100%;
```
fregante, line 27: We probably also want a sensible line-height, at least `1.3`. It's too tight as is

kidonng: It defaults to `normal` (which is `1.2` [according to MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#normal)). GitHub's `body`, `.markdown-body` and [Tailwind's preflight](https://github.com/tailwindlabs/tailwindcss/blob/62de02a37946c48c5fa6cdb800173b54198ebb9b/packages/tailwindcss/preflight.css#L30) use `1.5`, I assume it's fine to follow suit.

fregante (approved):

PR #7496, opened 2024-06-27
`list-prs-for-file` - Use new style

### LINKED ISSUE #7468: `list-prs-for-file` dropdown gets hidden behind Symbols panel
### Description

The dropdown to list PRs for the current file is obscured behind the opened Symbols panel.

It's easy enough to close the panel to see the content, but it should try to come up to the top of whatever Z index is there.

I found these related bits, might be helpful:

* https://github.com/refined-github/refined-github/blob/main/source/features/list-prs-for-file.tsx
* https://github.com/refined-github/refined-github/pull/7417
* https://github.com/refined-github/refined-github/blob/ce7d7e71c32c11a68e7fa54ce5e46a7227132ceb/source/features/list-prs-for-file.tsx#L40-L53

* Visit https://github.com/pypi/warehouse/blob/main/warehouse/packaging/models.py
* Open the Symbols panel
* Click on the List Pull Requests dropdown
* Observe that the behavior is hidden behind the Symbols list

Screenshot 2024-06-18 at 12 32 51

### Extension version

24.6.14

### Browser(s) used

Chrome Version 126.0.6478.62 (Official Build) (arm64)

kidonng:
~~Fix~~ #7468

Updated the feature to use new `.Overlay` style and `<anchored-position>` element, fixing the issue along the way.

## Test URLs

https://github.com/pypi/warehouse/blob/main/warehouse/packaging/models.py

## Screenshot

Screenshot 2024-06-27 at 7 36 56 PM

There's a new issue though: hovercards now get covered by the popover - maybe there's someway to tweak the popover or we just disable the hovercard (and maybe revert to tooltips instead - I remember that's how the feature used to work).

Screenshot 2024-06-27 at 7 37 22 PM


source/features/list-prs-for-file.tsx
```diff
@@ -23,35 +23,54 @@ function getHovercardUrl(prNumber: number): string {
...
+
 	return (
-		<details className="dropdown details-reset details-overlay flex-self-center rgh-list-prs-for-file">
-			<summary className="btn btn-sm">
+		<div>
+			<button
+				type="button"
+				className="Button Button--secondary color-fg-muted"
+				id={buttonId}
+				// @ts-expect-error Used by GitHub
```
kidonng, line 41: I tried to add `popovertarget ` to `IntrinsicAttributes` with no luck, TS still complains 🤷

fregante:
> There's a new issue though

That would be a deal breaker since this is a cosmetic change. The new dropdown uses the new popover API and their hovercard doesn't. There's zero way for us to "change the z-index" here unless we hack the hovercard to use the same API — which we shouldn't do. 

I'd say let's wait until they update that component and then we can reopen this PR.

kidonng:
> The new dropdown uses the new popover API and their hovercard doesn't.

Yeah too bad, this also breaks on GitHub's own menu (though the example below is a bit contrived):

Screenshot 2024-06-29 at 10 58 42 AM

> There's zero way for us to "change the z-index" here unless we hack the hovercard to use the same API — which we shouldn't do.

I just discovered there's also this `[data-hovercard-z-index-override]` attribute, but you are right it won't help.

source/features/list-prs-for-file.tsx
kidonng: I forgot this is HTML standard now (thought it was a GitHub thing), I really only read it one time or two.

fregante:
> this also breaks on GitHub's own menu

We can only hope that they will update the hovercard soon. In their case however the user doesn't _need_ to see the link while that popup is open.

Before this PR: The user has to close the symbols sidebar to use it
After this PR: The user cannot see the link they're hovering.

I'll close this PR for now because I found a fix for that specific issue. It'd be great to leave a comment in `list-prs-for-file` to mention this PR for future upgrades, so that we don't forget it

fregante:
@kidonng wanna retry this? I found this class in the source code: `js-hovercard-left`

## Before (below)

Screenshot 2

## After (left)

Screenshot 1

fregante:
Thank you by the way! 🙏

fregante:
Screenshot 1

PR #7510, opened 2024-06-30
`unreleased-commits` - Fix icon alignment

kidonng:
Small cosmetic tweak

## Test URLs

## Screenshot

| Before | After |
|--------|--------|
| Screenshot 2024-06-30 at 1 58 47 PM | Screenshot 2024-06-30 at 2 04 30 PM |

### DIFF
--- a/source/features/unreleased-commits.tsx
+++ b/source/features/unreleased-commits.tsx
@@ -91,8 +91,8 @@ async function createLink(
 			href={buildRepoURL('compare', `${latestTag}...${await getDefaultBranch()}`)}
 			aria-label={label}
 		>
-			<TagIcon className="v-align-middle"/>
-			{aheadBy === undeterminableAheadBy || <sup className="ml-n2"> +{aheadBy}</sup>}
+			<TagIcon/>
+			{aheadBy === undeterminableAheadBy || <sup className="ml-n1"> +{aheadBy}</sup>}
 		</a>
 	);
 }


source/features/unreleased-commits.tsx
```diff
@@ -91,8 +91,8 @@ async function createLink(
 			href={buildRepoURL('compare', `${latestTag}...${await getDefaultBranch()}`)}
 			aria-label={label}
 		>
-			<TagIcon className="v-align-middle"/>
-			{aheadBy === undeterminableAheadBy || <sup className="ml-n2"> +{aheadBy}</sup>}
```
fregante, line 95: n2 was intentionally to bring it closer together, it's not a label but it's "part of the icon"

kidonng: Only remove `v-align-middle`:

Screenshot 2024-06-30 at 2 20 55 PM

Ok-ish but in my opinion we should have a little space here.

fregante (changes requested):
I think it’s fine as is. If you think of the + as part of the icon, it doesn’t look off center. The visual weight of the icon is somewhat balanced. In your last screenshot there’s a lot of space below the label and icon and it looks off balance.

fregante:
Can you resend this?

- align + icon
- align tag icon only if count is missing

Because this does look misaligned:

Screenshot 15

PR #7513, opened 2024-06-30
De-init features earlier

fregante:
- Will fix https://github.com/refined-github/refined-github/issues/6437

_untested_


fregante:
This appears to work without visible side effects 😮 


If https://github.com/refined-github/refined-github/issues/5723 is ever implemented, then we'll have to revisit, but I don't see that happening anytime soon.

kidonng:
> This appears to work without visible side effects 😮

Sadly I don't have much opportunity to test this (and follow all the pjax/turbo changes in recent years), if it works then 🎉

PR #7514, opened 2024-06-30
`comment-on-draft-pr-indicator` - Create `netiquette`-like banner

### LINKED ISSUE #7438: `comment-on-draft-pr-indicator`: Main body comment button not modified
### Description

The comment button's label is unaltered in the main body of a pull request. The screenshot example in the feature shows it works, so either it got removed or the example needs to be updated.

Not entirely sure how it worked before, but the class is not present here:

https://github.com/refined-github/refined-github/blob/6302d1e0c78ee4126187008c42bbf068c819dd8a/source/features/comment-on-draft-pr-indicator.tsx#L11-L17

Adding `'.btn-primary'` modifies the button's label, but I am not sure if this is the proper resolution.

1. Open a draft pull request (https://github.com/refined-github/sandbox/pull/7)
2. See the comment button

### Extension version

24.6.9

### Browser(s) used

Arc, Firefox

kidonng:
Resolve #7438

## Test URLs

https://github.com/refined-github/sandbox/pull/7

## Screenshot

Regular

Screenshot 2024-06-30 at 7 11 03 PM

Inline

Screenshot 2024-06-30 at 7 11 57 PM

### DIFF
--- a/source/features/comment-on-draft-pr-indicator.tsx
+++ b/source/features/comment-on-draft-pr-indicator.tsx
@@ -27,3 +27,11 @@ void features.add(import.meta.url, {
 	awaitDomReady: true,
 	init,
 });
+
+/*
+
+Test URL:
+
+https://github.com/refined-github/sandbox/pull/7
+
+*/
--- a/source/features/comment-on-draft-pr-indicator.tsx
+++ b/source/features/comment-on-draft-pr-indicator.tsx
@@ -1,28 +1,29 @@
+import React from 'dom-chef';
+import InfoIcon from 'octicons-plain-react/Info';
 import * as pageDetect from 'github-url-detection';
 
 import features from '../feature-manager.js';
 import observe from '../helpers/selector-observer.js';
+import createBanner from '../github-helpers/banner.js';
 
-function addIndicator(button: HTMLElement): void {
-	const preposition = button.textContent.includes('Add') ? ' to ' : ' on ';
-	button.textContent += preposition + 'draft PR';
+function addDraftBanner(newCommentField: HTMLElement): void {
+	newCommentField.before(
+		createBanner({
+			icon: <InfoIcon className="m-0"/>,
+			classes: 'p-2 my-2 mx-md-2 text-small color-fg-muted border-0'.split(' '),
+			text: <>This is a <strong>draft PR</strong>. Make sure you would like to comment.</>,
+		}),
+	);
 }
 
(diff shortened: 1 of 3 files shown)


source/features/comment-on-draft-pr-indicator.tsx
```diff
@@ -1,28 +1,37 @@
...
 
-function addIndicator(button: HTMLElement): void {
-	const preposition = button.textContent.includes('Add') ? ' to ' : ' on ';
-	button.textContent += preposition + 'draft PR';
+function addDraftBanner(newCommentField: HTMLElement): void {
+	newCommentField.before(
+		createBanner({
+			icon: <InfoIcon className="m-0"/>,
+			classes: 'p-2 my-2 mx-md-2 text-small color-fg-muted border-0'.split(' '),
+			text: <>This is a <strong>draft PR</strong>. Make sure you would like to comment.</>,
```
kidonng, line 14: Wording suggestions welcome

fregante:
Great! I remember either you or @yakov116 mentioned another situation where such a warning would be useful. Do you remember what that was?

source/features/comment-on-draft-pr-indicator.tsx
fregante:
```suggestion
			text: <>This is a <strong>draft PR</strong>, it might not be ready for review.</>,
```


I'd also replace the info icon with the draft PR icon

fregante (changes requested):

fregante:
It might also be a good time to update the screenshot, after we settle on the wording/style/breadth. The old screenshot is also low-quality and unnecessarily split, we just need a single small retina screenshot.

kidonng:
Would you fancy something like this?

Screenshot 2024-06-30 at 6 23 00 PM


> another situation where such a warning would be useful. Do you remember what that was?

Nope, but `netiquette` and this should pretty much cover all common scenario?

fregante:
It feels a bit much and not something GitHub would use. Commenting on draft PRs is still very common so we shouldn't make it too ugly/noisy. The usual notice should be ok.

fregante:
`npm run fix` next time. It autofixes everything it can autofix

fregante (approved):
LGTM but let's not rush it because to see if there's better wording or if Yakov remembers what we wanted.

fregante:
cc https://github.com/refined-github/refined-github/issues/7541

Issue #6554, opened 2023-04-25
React-based views don't correctly update features


fregante:
I'm opening this issue to track it properly. In short:

1. Open https://github.com/refined-github/refined-github/blob/main/source/features/action-used-by-link.tsx
2. Visit a folder or change file via the sidebar or go back and forth

### Known issues

- `turbo:render` isn't called
- `popstate` events don't trigger any events
- the view isn't generated from scratch, so previous elements stick around 

### Related issues 

- #6414 
- #6528 

### Current status

For the whole existence of Refined GitHub, we've relied on _GitHub_ to delete our elements when the page/URL changes, with one exception (I don't remember which). New React-based views break this assumption. I think we need to:

- listen to `soft-nav:render`
- listen to `popstate` and avoid duplicates calls in case `soft-nav:render` is _also_ fired
- figure out a way to make widgets/buttons dynamic so that they change or disappear when the URL changes
	- this could help, but that alone doesn't fit into our concept of "feature unloading when pages change": #6493 
	- maybe we could add a `onAbort(() => element.remove())`

134130:
I usually used `MutationObserver`.
I think watching popstate of Element is more proper to use It

fregante:
> I think using It is more proper for watching popstate of Element.

Each page/view is still tracked "by URL", so listening to URL changes is definitely the proper way to do it.

Detecting changes isn't the problem, our problem is that we need to respond to those changes.

We can already use our selector observer to watch the page for specific elements without tracking "every change". The only disadvantage is that it doesn't tell us when elements are removed, but we actually have a separate listener specifically for that: `onElementRemoval` and `onElementReplacement` (which haven't really been needed since we added the selector observer)

In Refined GitHub we have a global feature manager that loads and unloads features based on a "page change". There is no easy and reliable way to detect a "page change" with a MutationObserver because some elements are dropped/regenerated by React and that doesn't mean that the page was updated.

fregante:
There might be a way to reliabley catch URL changes:

- https://github.com/fregante/webext-events/issues/6

However I don't know yet:

- are the events fast enough to replace `turbo:render` and avoid FOUC? It seems to me that they're actually early (in the background page) but there could be some delay when passing the message to the content script
- does RGH react well when a "url change" event is received but the page doesn't change? e.g. pressing `y` on files will update just the URL
- could this event be early enough to be used as an "unload" event?
	- #6437  

Unresolved issues:

[...]

PR #7524, opened 2024-07-03
`conversation-links-on-repo-lists` - Improve AJAX support

gxr404:
Closes https://github.com/refined-github/refined-github/issues/7520

### Bug1:  It should not be displayed in search "code" panel 

1. open [github.com/search?q=%22webext%22&type=repositories](https://github.com/search?q=%22webext%22&type=repositories)
2. Click "Code" in the sidebar
3. Notice extra links
4. Refresh the page
5. The links are gone

### Bug2: Refresh the page from  search "code" panel , but it does not display after clicking "repositories"

1. open [github.com/search?q=%22webext%22&type=code](https://github.com/search?q=%22webext%22&type=code)
2. Click "Repositories" in the sidebar
3. The links are gone
4. Refresh the page
5.  links show

## Screenshot

Before:

before 

After: 

after
|

### DIFF
--- a/source/features/conversation-links-on-repo-lists.tsx
+++ b/source/features/conversation-links-on-repo-lists.tsx
@@ -37,6 +37,9 @@ function addConversationLinks(repositoryLink: HTMLAnchorElement): void {
 }
 
 function addSearchConversationLinks(repositoryLink: HTMLAnchorElement): void {
+	// Do not move to `includes` until React AJAX issues are resolved:
+	// https://github.com/refined-github/refined-github/pull/7524#issuecomment-2211692096
+	// https://github.com/refined-github/refined-github/issues/6554
 	if (new URLSearchParams(location.search).get('type') !== 'repositories') {
 		return;
 	}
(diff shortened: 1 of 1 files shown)


fregante:
Thanks for the PR, but I suggested something different in the original issue: I'd rather make the CSS selector stricter so that it doesn't select those items at all. Is it possible?

fregante:
Technically the feature doesn't work at all because the page has changed, so the selector definitely needs to be updated

https://github.com/search?q=x&type=repositories

Screenshot

gxr404:
sorry，Does “CSS selector stricter” refer to the observer there?

Since clicking from "code" page to "repositories" page does not refresh and only changes the dom,  `include` cannot add `repositories` as a judgment criterion

The dom of "repositories" page is similar to the dom of other search results

If solved from the selector level,  `observer`needs to be changed to `observe('[data-testid="results-list"] > div > div > div > h3 .search-title a', ....` in order to accurately select "repositories" page DOM
Is this acceptable？

fregante:
> Technically the feature doesn't work at all

Actually not entirely true, the feature works but only when the user lands on the specific repo search page. This is definitely an AJAX issue:

- #6554 



> sorry，Does “CSS selector stricter” refer to the observer there?

Yes:

https://github.com/refined-github/refined-github/blob/e33a312b6bbbd52536e7a64860a4e82fd984d100/source/features/conversation-links-on-repo-lists.tsx#L77

fregante (approved):
I think your changes actually fix both issues

PR #7530, opened 2024-07-08
`fit-textareas` - Apply to issue template placeholders

karlhorky:
Closes https://github.com/refined-github/refined-github/issues/7522

## Test URLs

- https://github.com/mdx-js/mdx/issues/new?assignees=&labels=&projects=&template=1-bug.yml

## Screenshot

Screenshot 2024-07-08 at 15 45 51

### DIFF
--- a/source/features/fit-textareas.css
+++ b/source/features/fit-textareas.css
@@ -1,4 +1,5 @@
 .rgh-fit-textareas {
+	height: auto !important;
 	max-height: none !important;
 	field-sizing: content;
 }
--- a/source/features/fit-textareas.css
+++ b/source/features/fit-textareas.css
@@ -3,6 +3,6 @@
 	field-sizing: content;
 }
 
-textarea.rgh-fit-textareas.form-control {
+:root textarea.rgh-fit-textareas.form-control {
 	height: auto;
 }
--- a/source/features/fit-textareas.css
+++ b/source/features/fit-textareas.css
@@ -1,5 +1,8 @@
 .rgh-fit-textareas {
-	height: auto !important;
 	max-height: none !important;
 	field-sizing: content;
+}
+
+textarea.rgh-fit-textareas.form-control {
+	height: auto;
 }
--- a/source/features/fit-textareas.tsx
+++ b/source/features/fit-textareas.tsx
@@ -20,7 +20,12 @@ function inputListener({target}: Event): void {
 
 function watchTextarea(textarea: HTMLTextAreaElement, {signal}: SignalAsOptions): void {
 	// Disable constrained GitHub feature
-	textarea.classList.replace('js-size-to-fit', 'rgh-fit-textareas');
+	if (textarea.classList.contains('js-size-to-fit')) {
+		textarea.classList.replace('js-size-to-fit', 'rgh-fit-textareas');
+	} else {
+		textarea.classList.add('rgh-fit-textareas');
+	}
+
 	if (nativeFit) {
 		return;
 	}
--- a/source/features/fit-textareas.tsx
+++ b/source/features/fit-textareas.tsx
@@ -20,11 +20,8 @@ function inputListener({target}: Event): void {
 
 function watchTextarea(textarea: HTMLTextAreaElement, {signal}: SignalAsOptions): void {
 	// Disable constrained GitHub feature
-	if (textarea.classList.contains('js-size-to-fit')) {
-		textarea.classList.replace('js-size-to-fit', 'rgh-fit-textareas');
-	} else {
-		textarea.classList.add('rgh-fit-textareas');
-	}
+	textarea.classList.remove('js-size-to-fit');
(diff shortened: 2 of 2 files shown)


source/features/fit-textareas.tsx
```diff
@@ -20,7 +20,12 @@ function inputListener({target}: Event): void {
 
 function watchTextarea(textarea: HTMLTextAreaElement, {signal}: SignalAsOptions): void {
 	// Disable constrained GitHub feature
-	textarea.classList.replace('js-size-to-fit', 'rgh-fit-textareas');
+	if (textarea.classList.contains('js-size-to-fit')) {
+		textarea.classList.replace('js-size-to-fit', 'rgh-fit-textareas');
+	} else {
+		textarea.classList.add('rgh-fit-textareas');
+	}
```
karlhorky, line 27: also happy to do this differently, if desired:

1. only 1 line of `textarea.classList.add('rgh-fit-textareas');`
2. add CSS to `.rgh-fit-textareas` class in `source/features/fit-textareas.css` to override the `.js-size-to-fit` styles

source/features/fit-textareas.css
```diff
@@ -1,4 +1,5 @@
 .rgh-fit-textareas {
+	height: auto !important;
```
fregante, line 2: I believe this will break the manual resizing (you can drag the bottom right corner to resize it)

source/features/fit-textareas.tsx
fregante: No conditions are necessary, you can just remove the class (whether it exists or not) and add the other. No "replace" call

source/features/fit-textareas.css
karlhorky: indeed it does 🤔 😟

the reason I added it was that:

1. the `height: 100px !important` below
2. `height: <pixel value>` was appearing as an inline style on some of these `textarea`s

```css
.issue-form-textarea {
    height: 100px !important;
    min-height: 100px !important;
}
```

but maybe I should just replace that `.issue-form-textarea` class instead (and probably others, eg. PRs, Discussions, etc, hmm...

if you have any suggestions, also happy to look

source/features/fit-textareas.tsx
karlhorky: sounds good:

- 05b69265182fdc4dd73dbad0b988fcb3cf377506

source/features/fit-textareas.css
karlhorky: in addition to the styles above, there's also another block for the unfocused fields:

```css
.form-group textarea.form-control {
    width: 100%;
    height: 200px;
    min-height: 200px;
}
```

where the `height: 200px` is also causing constraints there

karlhorky: alright, I tried out the first approach:

- remove the first class 84903c6cfa4e35b5ba9253159e9b4205795d9df1
- make selector stronger 8e890d62f73550e1621396712952b77f5e8c8488

karlhorky: if the `height: auto !important` with the `.rgh-fit-textarea.issue-form-textarea` selector is preferable after all, I can switch to that instead

source/features/fit-textareas.css
```diff
@@ -2,3 +2,7 @@
 	max-height: none !important;
 	field-sizing: content;
 }
+
+textarea.rgh-fit-textareas.form-control {
```
fregante, line 6: Don't all textareas have this class?

karlhorky: the focused `textarea`s have this class:

```
js-comment-field js-paste-markdown js-task-list-field js-quick-submit FormControl-textarea CommentBox-input rgh-fit-textareas js-saved-reply-shortcut-comment-field input-monospace rgh-seen--7661699013
```

even if `height: auto` (without `!important`) was applied in future to the focused `textarea` elements, it would still not cause a problem for the manual resizing, which sets an inline `height`, which is higher specificity

fregante: Selector not strong enough for me, it's 200px:

Screenshot 3

```suggestion
:root textarea.rgh-fit-textareas.form-control {
```

Should work

karlhorky: ok sounds good. interesting, wonder why that didn't show up for me - maybe I didn't test correctly

fregante: > wonder why that didn't show up for me

My guess is that the browser isn't consistent in the injection order of the styles.

karlhorky:
Thanks for the review and merge and publish in [`refined-github@24.7.10`](https://github.com/refined-github/refined-github/releases/tag/24.7.10)

I tested the new version just now in Chrome, and it is no longer obscuring the text in the unfocused fields 🎉


https://github.com/user-attachments/assets/7082fd07-be92-4716-a7b7-8bf2c3eb8265

PR #7532, opened 2024-07-09
`clean-repo-sidebar` Wait for package count to load for internal packages

ramblingenzyme:
This affects repositories with only internal packages. So inside a Github Organisation.
This affects the repository overview page where `clean-repo-sidebar` runs.

## Copied HTML
```html
<include-fragment src="/my-org/my-repo/packages_list?current_repository=my-repo" aria-busy="true" aria-label="Loading latest packages">
    <h2 class="h4 mb-3">
  <a href="/orgs/my-org/packages?repo_name=my-repo" data-view-component="true" class="Link--primary no-underline Link d-flex flex-items-center">
    Packages
      <span title="0" hidden="hidden" data-view-component="true" class="Counter ml-1">0</span>
</a></h2>


        <div class="mb-2 d-flex flex-items-center">
          <div class="Skeleton mr-2" style="width:20px;height:20px;"></div>
          <div class="Skeleton Skeleton--text flex-auto">&nbsp;</div>
        </div>
        <div class="mb-2 d-flex flex-items-center">
          <div class="Skeleton mr-2" style="width:20px;height:20px;"></div>
          <div class="Skeleton Skeleton--text flex-auto">&nbsp;</div>
        </div>
        <div class="mb-2 d-flex flex-items-center">
          <div class="Skeleton mr-2" style="width:20px;height:20px;"></div>
          <div class="Skeleton Skeleton--text flex-auto">&nbsp;</div>
        </div>
</include-fragment>
```

## Test URLs
* https://github.com/refined-github/refined-github

### DIFF
--- a/source/features/clean-repo-sidebar.tsx
+++ b/source/features/clean-repo-sidebar.tsx
@@ -30,7 +30,7 @@ async function cleanReleases(): Promise<void> {
 }
 
 async function hideEmptyPackages(): Promise<void> {
-	const packagesCounter = await elementReady('.Layout-sidebar .BorderGrid-cell a[href*="/packages?"] .Counter', {waitForChildren: false});
+	const packagesCounter = await elementReady('.Layout-sidebar .BorderGrid-cell a[href*="/packages?"] .Counter:not([hidden="hidden"])', {waitForChildren: false});
 	if (packagesCounter && packagesCounter.textContent === '0') {
 		packagesCounter.closest('.BorderGrid-row')!.hidden = true;
 	}
--- a/source/features/clean-repo-sidebar.tsx
+++ b/source/features/clean-repo-sidebar.tsx
@@ -32,14 +32,23 @@ async function cleanReleases(): Promise<void> {
 }
 
 async function hideEmptyPackages(): Promise<void> {
-	const packageLoader = await elementReady('include-fragment[src*="packages_list"]', {waitForChildren: false});
-	if (packageLoader) {
-		// 'loadend' captures success and error
-		// https://github.com/github/include-fragment-element/blob/5249243ee1cbdc82ab69c5d7c6318cd61a524b93/src/include-fragment-element.ts#L261
-		await oneEvent(packageLoader, 'loadend');
-	}
+	const packagesCounterPromise = elementReady('.Layout-sidebar .BorderGrid-cell a[href*="/packages?"] .Counter', {waitForChildren: false});
 
-	const packagesCounter = await elementReady('.Layout-sidebar .BorderGrid-cell a[href*="/packages?"] .Counter', {waitForChildren: false});
+	void elementReady('include-fragment[src*="packages_list"]', {waitForChildren: false})
+		.then(async packageLoader => {
+			if (packageLoader) {
+				// 'loadend' captures success and error
+				// https://github.com/github/include-fragment-element/blob/5249243ee1cbdc82ab69c5d7c6318cd61a524b93/src/include-fragment-element.ts#L261
+				await oneEvent(packageLoader, 'loadend');
+
+				const packagesCounter = await packagesCounterPromise;
+				if (packagesCounter && packagesCounter.textContent !== '0') {
+					packagesCounter.closest('.BorderGrid-row')!.hidden = true;
+				}
+			}
+		});
+
+	const packagesCounter = await packagesCounterPromise;
 	if (packagesCounter && packagesCounter.textContent === '0') {
 		packagesCounter.closest('.BorderGrid-row')!.hidden = true;
 	}
--- a/source/features/clean-repo-sidebar.tsx
+++ b/source/features/clean-repo-sidebar.tsx
@@ -5,6 +5,7 @@ import elementReady from 'element-ready';
 import * as pageDetect from 'github-url-detection';
 
 import features from '../feature-manager.js';
+import oneEvent from 'one-event';
 // The h2 is to avoid hiding website links that include '/releases' #4424
 // TODO: It's broken
 const releasesSidebarSelector = '.Layout-sidebar .BorderGrid-cell h2 a[href$="/releases"]';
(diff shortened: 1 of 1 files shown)


source/features/clean-repo-sidebar.tsx
```diff
@@ -31,8 +31,21 @@ async function cleanReleases(): Promise<void> {
...
-	if (packagesCounter && packagesCounter.textContent === '0') {
-		packagesCounter.closest('.BorderGrid-row')!.hidden = true;
+	if (!packagesCounter || packagesCounter.textContent !== '0') {
+		return;
+	}
+
+	packagesCounter.closest('.BorderGrid-row')!.hidden = true;
+
+	// Github seems to fetch internal packages separately, so if a repo only has internal packages, this makes the panel visible after they've loaded
+	const updatedPackagesCounter = await elementReady('.Layout-sidebar .BorderGrid-cell a[href*="/packages?"] .Counter:not([hidden="hidden"])', {
```
ramblingenzyme, line 41: I'm not very sold on this idea tbh but I'm not sure how to make it better, since I think any repo inside an Org can have internal packages?

source/features/clean-repo-sidebar.tsx
```diff
@@ -31,8 +31,21 @@ async function cleanReleases(): Promise<void> {
...
+		waitForChildren: false,
+		stopOnDomReady: false,
+		timeout: 10 * 1000,
```
fregante, line 44: The problem with this, other than the repeated code, is that it affects everyone else negatively. `elementReady` is relatively heavy because it queries the DOM up to 60 times per second, for 10 seconds in this case.

What you want is to target the "deferred" placeholder element that GitHub probably has, and wait for it to trigger the load event.

I think the code would look like:

```js
const el = await elementReady('deferred-content[src*=packages]');
await oneEvent(el, 'load')
hidden = false
```

fregante (changes requested):

source/features/clean-repo-sidebar.tsx
fregante: By the way I don't know what the DOM in that case looks like so it might be good to click "View source" on your repo page and copy the HTML from there, which should show that "deferred content" element (I don't remember the exact name)

ramblingenzyme: Is it `include-fragment`? https://github.com/github/include-fragment-element

fregante: Correct

fregante:
This order won't work when the fragment is missing, rather it will delay the execution of the rest of the code until the page has finished loading. 

I think it's better to restore the order you had before:

1. Hide it when missing
2. *then* look for a fragment
3. And if found and loaded…
4. Unhide if no longer missing

fregante:
This might be a good opportunity to try with CSS selectors and dropping this JS part altogether. 

```css
[rgh-clean-repo-sidebar] .BorderGrid-row:has(a[href*="/packages?repo_name"].Counter[title="0"]) {
	display: none;
}
```

This rule should work, you just need to add the attribute on HTML at the beginning of the `init` in JS:

```js
document.documentElement.setAttribute('rgh-clean-repo-sidebar', '');
```

How does that look? I hope to eventually move most or all of the feature to CSS, so this would be a good first piece to take care of.

ramblingenzyme:
> This order won't work when the fragment is missing, rather it will delay the execution of the rest of the code until the page has finished loading.
> 
> I think it's better to restore the order you had before:
> 
>     1. Hide it when missing
> 
>     2. _then_ look for a fragment
> 
>     3. And if found and loaded…
> 
>     4. Unhide if no longer missing

Is there a reason the fragment would be missing? I didn't find that during my testing, the `if` block is only there because `elementReady` can return `undefined` and Typescript would've complained. 

However, the `include-fragment` element does disappear *after* the content loads, so searching for it afterwards creates a race condition.

ramblingenzyme:
> This might be a good opportunity to try with CSS selectors and dropping this JS part altogether.
> 
> ```css
> [rgh-clean-repo-sidebar] .BorderGrid-row:has(a[href*="/packages?repo_name"].Counter[title="0"]) {
> 	display: none;
> }
> ```
> 
> This rule should work, how does that look? I hope to eventually move most or all of the feature to CSS, so this would be a good first piece to take care of.


🤔 I'll give the CSS selector approach a go and see if it works, though I think it'd be 

```css
 [rgh-clean-repo-sidebar] .BorderGrid-row:has(a[href*="/packages_list?current_repository="].Counter[title="0"]) {
 	display: none;
 }
```


Otherwise I think this would be the JS approach I'd take next.
```js
async function hideEmptyPackages(): Promise<void> {
    const packagesCounterPromise = elementReady('.Layout-sidebar .BorderGrid-cell a[href*="/packages?"] .Counter', {waitForChildren: false});

    elementReady('include-fragment[src*="packages_list"]', {waitForChildren: false}).then(async (packageLoader) => {
        if (packageLoader) {
            // 'loadend' captures success and error
            // https://github.com/github/include-fragment-element/blob/5249243ee1cbdc82ab69c5d7c6318cd61a524b93/src/include-fragment-element.ts#L261
            await oneEvent(packageLoader, 'loadend');

            const packagesCounter = await packagesCounterPromise
            if (packagesCounter && packagesCounter.textContent !== '0') {
                packagesCounter.closest('.BorderGrid-row')!.hidden = false;
            }
        }
    });

    const packagesCounter = await packagesCounterPromise;
    if (packagesCounter && packagesCounter.textContent === '0') {
        packagesCounter.closest('.BorderGrid-row')!.hidden = true;
    }
}
```
> 
> Also please include a test URL in the PR (as the template requested) and at the end of the file.

I'm not sure I understand the point of providing a test URL for this, given it'd be for a private repo inside my workplace's Github Organisation, and internal packages are a Github Enterprise Cloud feature as far as I'm aware anyway.

fregante:
I don't know if the fragment is always there, you haven't included any test URLs yet. 

As for the race condition, it wouldn't matter, because if it's missing then there's no change to make and the polling ends on DOMready anyway. 

At any rate I prefer the CSS-only solution, either in this PR or I'll send a PR later today

ramblingenzyme:
@fregante  I've added a test URL, but it's only going to be useful to check if this hasn't broken anything.

ramblingenzyme:
> As for the race condition, it wouldn't matter, because if it's missing then there's no change to make and the polling ends on DOMready anyway.

The order would be:
```
1. Look for the counter
2. Hide it when missing
3. _then_ look for a fragment
4. Wait for `loadend` event
5. Unhide if the counter no longer says 0
```

Doesn't that means the fragment could trigger the `loadend` event or disappear before we look for it, which means we won't unhide the Packages panel, and the bug isn't fixed?

fregante:
> I'm not sure I understand the point of providing a test URL for this,

The PR template gave you some information that was disregarded:

Screenshot 4

Without a test URL, I can't verify that the issue has been fixed _and_ it will have a higher risk of being undone/broken in the future because no one knows where it happened and how to reproduce it. This is also why we've been adding the test URLs in each file as well.




> I'm getting confused between the `include-fragment[src]` and `a[href]` selectors.

The CSS selector does not need to account for the fragment, it simply looks for the counter, whenever that is set/reset to 1, dynamically.

I opened a PR with the selector (https://github.com/refined-github/refined-github/pull/7552). It already worked, except I forgot a space before `.Counter`

ramblingenzyme:
@fregante so, for future reference, you're saying you actually *do* want people to submit inaccessible test URLs on PRs?

fregante:
As shown by the bug report and in my PR, the issue exists outside private repos, what you're saying does not apply. 

Even in the case of private-only bugs, you can leave a URL template to make sure people have a chance to recreate the issue *somehow*.

At any rate I appreciate the report/PR

Issue #7536, opened 2024-07-10
`unreleased-commits` - Duplicated after system theme update (dark mode)


fregante:
### Description

Same as:

- https://github.com/refined-github/refined-github/issues/7098
- https://github.com/refined-github/refined-github/pull/7403

<img width="328" alt="Screenshot 1" src="https://github.com/refined-github/refined-github/assets/1402241/ee89a0fc-37ba-45bd-857f-07d83e6e06bb">

 

### How to replicate the issue + URL

1. https://github.com/refined-github/refined-github
2. toggle your system color

### Extension version

24.6.14

### Browser(s) used

 Safari

fregante:
The solution like looks like:

https://github.com/refined-github/refined-github/blob/77496780e3604e1ad399d1642aeeb0a2422e5276/source/features/default-branch-button.tsx#L35-L39

PR #7440, opened 2024-06-13
`previous-version` - Make it more visible

kovsu:
## Test URLs

https://github.com/refined-github/refined-github/blob/main/readme.md

## Screenshot

before:

image

after:

CleanShot 2024-06-13 at 08 50 23@2x

PR #7539, opened 2024-07-10
`previous-version` - Various fixes

fregante:
- Part of #6554 
- Follows #7440 
- Also fixes a theme-related issue like #7536 
- Also enables it on folders

## Test URLs

https://github.com/refined-github/refined-github/blob/main/readme.md

## Screenshot

Screenshot 7

### DIFF
--- a/source/github-events/on-react-page-update.tsx
+++ b/source/github-events/on-react-page-update.tsx
@@ -5,6 +5,7 @@ export default function onReactPageUpdate(
 	signal: AbortSignal,
 ): void {
 	document.addEventListener('soft-nav:payload', () => {
+		// @ts-expect-error https://github.com/DefinitelyTyped/DefinitelyTyped/issues/60868
 		const unifiedSignal = AbortSignal.any([
 			signal, // User-provided, likely Turbo page navigation event
 			signalFromEvent(document, 'soft-nav:payload'), // A "React page"-specific page navigation event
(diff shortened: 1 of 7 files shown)


source/github-events/on-react-page-update.tsx
```diff
@@ -0,0 +1,15 @@
+import {signalFromEvent} from 'abort-utils';
+
+export default function onReactPageUpdate(
+	callback: (signal: AbortSignal) => void,
+	signal: AbortSignal,
+): void {
+	document.addEventListener('soft-nav:payload', () => {
```
fregante, line 7: This event appears on both clicks and history navigation, just on React pages.

In the future we might use a more standard https://developer.mozilla.org/en-US/docs/Web/API/Navigation_API, but currently no Firefox nor Safari support.



### Click

Screenshot 8


### History

Screenshot 9

source/manifest.json
```diff
@@ -5,15 +5,15 @@
...
 	"browser_specific_settings": {
 		"gecko": {
 			"id": "{a4c4eda4-fb84-4a84-b4a1-f7c1cbf2a1ad}",
-			"strict_min_version": "121.0"
+			"strict_min_version": "124.0"
 		},
 		"gecko_android": {
 			"id": "{a4c4eda4-fb84-4a84-b4a1-f7c1cbf2a1ad}",
-			"strict_min_version": "121.0"
+			"strict_min_version": "124.0"
```
fregante, line 16: Minimum supported version upped due to AbortSignal.any()

PR #7543, opened 2024-07-11
Meta: Simplify GitHubFileURL

fregante:
Retries https://github.com/refined-github/refined-github/pull/6870


fregante:
Firefox still buggy.

1. Open https://github.com/refined-github/sandbox/tree/locked
2. Notice `default-branch-button` pointing to the wrong page

PR #7569, opened 2024-07-21
`netiquette` - Avoid appearing on wrong places

kidonng:
Follow up https://github.com/refined-github/refined-github/pull/7568#issuecomment-2241694654

## Test URLs

https://github.com//nodejs/node/pull/43689#discussion_r914328969

## Screenshot

| Before | After |
|--------|--------|
| Screenshot 2024-07-21 at 11 53 05 PM | Screenshot 2024-07-22 at 12 01 12 AM |

### DIFF
--- a/source/features/comment-on-draft-pr-indicator.tsx
+++ b/source/features/comment-on-draft-pr-indicator.tsx
@@ -6,7 +6,7 @@ import features from '../feature-manager.js';
 import observe from '../helpers/selector-observer.js';
 import createBanner from '../github-helpers/banner.js';
 import {isOwnConversation} from '../github-helpers/index.js';
-import {newCommentFieldSelector} from '../github-helpers/selectors.js';
+import {newCommentField} from '../github-helpers/selectors.js';
 
 function addDraftBanner(newCommentField: HTMLElement): void {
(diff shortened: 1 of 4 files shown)


source/features/comment-on-draft-pr-indicator.tsx
```diff
@@ -17,19 +17,21 @@ function addDraftBanner(newCommentField: HTMLElement): void {
 	);
 }
 
+export const newCommentFieldSelector = [
+	'[input="fc-new_comment_field"]',
+	'[input^="fc-new_inline_comment_discussion"]',
+];
```
fregante, line 23: Can you move this to selectors.ts? Then add an empty array for the tests, if allowed. Or `[0, 'some url']` to have `0 expected matches`

fregante (commented):
I'm not 100% sure about this. Even if you opened the conversation, you still need to "be kind" and "don't comment on this old issue"

The only change required for this feature would be that it should exclude the "edit comment" field, if `netiquette` appears here.

Other cleanup changes can stay in this PR regardless of this 👍

fregante (approved):

PR #7570, opened 2024-07-21
`comment-on-draft-pr-indicator` rename to `draft-pr-notice`

kidonng:
I forgot in #7514

## Test URLs

N/A

## Screenshot

N/A


fregante (approved):

fregante:
Should we also rename this to `draft-pr-banner`? The feature name is so long 🥲

kidonng:
I also feel like a better name. Perhaps `draft-pr-indicator` or `draft-pr-reminder`? I don't think `-banner` cuts it, it may not use a banner someday.

fregante:
`draft-pr-notice` 

It doesn't have to be exact, just short in order to be referenced easily in commits. 

We've been avoiding renames and just doing them for feature names that are too long. I think this one needs it… but it's not egregiously long so it may be ok.

fregante (approved):

PR #7580, opened 2024-07-23
`release-download-count` - Prevent time overflow

### LINKED ISSUE #7579: `release-download-count` causes layout shift
### Description

~~Seems only reproducible when there are no downloads~~

`<relative-time>` means the date may occupy more space than `Jan 31, 2000`

Screenshot 2024-07-23 at 11 41 39 AM

Note: may not replicate as time changes

https://github.com/kidonng/notion-appimage/releases/tag/3.10.0-1

### Extension version

24.7.10

### Browser(s) used

Chrome

kidonng:
Fix #7579

## Test URLs

https://github.com/cli/cli/releases/tag/v2.30.0

## Screenshot

Note: time is simulated

| Before | After |
|--------|--------|
| Screenshot 2024-07-23 at 12 02 20 PM | Screenshot 2024-07-23 at 12 02 24 PM |

### DIFF
--- a/source/features/release-download-count.css
+++ b/source/features/release-download-count.css
@@ -4,6 +4,11 @@
 	min-width: 6em;
 }
 
+/* <relative-time> can occupy more space #7579 */
+.rgh-release-download-count > span:last-child {
+	min-width: 8em;
+}
+
 /* Makes space for right aligning download count */
 .rgh-release-download-count [data-rgh-heat] {
 	min-width: 5em;
--- a/source/features/release-download-count.css
+++ b/source/features/release-download-count.css
@@ -5,7 +5,7 @@
 }
 
 /* <relative-time> can occupy more space #7579 */
-.rgh-release-download-count > span:last-child {
+.rgh-release-download-count > span:has(relative-time) {
 	min-width: 8em;
 }
 


source/features/release-download-count.css
```diff
@@ -4,6 +4,11 @@
 	min-width: 6em;
 }
 
+/* <relative-time> can occupy more space #7579 */
+.rgh-release-download-count > span:last-child {
```
fregante, line 8: You can use `has(relative-time)`

fregante (approved):

source/features/release-download-count.css
```diff
@@ -4,6 +4,11 @@
 	min-width: 6em;
```
fregante, line 4: Is this the minimum we could have? The columns seem quite wide for the file size

kidonng: Without RGH:

Screenshot 2024-07-24 at 1 08 21 AM

With RGH:

Screenshot 2024-07-24 at 1 26 31 AM

Probably not much gain making it narrower

fregante: The first one is dynamic, our width are fixed. The gain comes when the filename is long and the viewport is wide.

Since we're adding 2em to one column, it'd make sense to reduce the other by 2em, if it fits.

kidonng: > For example it already is broken on 24.7.11

Broken on rgh.not.enabled too 😃 well I guess it's not too hard to fix

Screenshot 2024-07-24 at 1 57 44 AM

fregante: I think grid/subgrid is the only safe solution here. Would you like to explore it? 

I think `display: grid` and `display: contents` might already be enough. `display: subgrid` might be easier to set up.

fregante: For the purpose of this bug fix, I think this is enough:

```suggestion
	min-width: 4em;
```

It can be improved further separately. Let's just test and merge

fregante, line 4: For the purpose of this bug fix, I think this is enough:

```suggestion
	min-width: 4em;
```

It can be improved further separately. Let's just test and merge

fregante (approved):

fregante:
ping

PR #7582, opened 2024-07-23
Try biome migration

fregante:
Full migration may not be possible due to:

- https://github.com/biomejs/biome/discussions/3493#discussion-6964316

We still want ESLint unless we want Prettier-style formatting on JS too 😖


fregante:
- replaced by https://github.com/refined-github/refined-github/pull/7584

ESLint can't be replaced until https://github.com/biomejs/biome/discussions/3493, which doesn't look like it'll ever happen.

PR #7587, opened 2024-07-23
`always-show-branch-delete-buttons` - Drop feature

fregante:
GitHub no longer hides them because the whole table is not responsive anymore

IMG_4132 Medium


fregante:
Old feature 🙂 

- https://github.com/refined-github/refined-github/pull/2466
- https://github.com/refined-github/refined-github/issues/2697

PR #7495, opened 2024-06-27
`tags-on-commits-list` - Restore feature

kidonng:
Fix #7459

I only updated the selector. Further work may include moving them to `selectors.ts` (which is a new thing to me) and also migrate from the deprecated `cache` function.

## Test URLs

https://github.com/refined-github/refined-github/commits/main/

## Screenshot

Screenshot 2024-06-27 at 6 03 18 PM

PR #7593, opened 2024-07-24
`mark-merge-commits-in-list` - Restore feature

kovsu:
same as  #7592


fregante:
For both PRs:

- we should leave both selectors in, with one marked as `TODO: Drop in 2025`

And then: are there more features that use this old class name?

kovsu:
@fregante 

There are no more features that use the old classname

image

fregante (approved):

kidonng:
> And then: are there more features that use this old class name?

For reference:

- #7495

Issue #7549, opened 2024-07-15
`default-branch-button` doesn't appear on `isSingleFile` pages


injust:
### Description

```
refined-github.js:3281 Uncaught (in promise) TypeError: Cannot read properties of null (reading 'classList')
    at fixFileHeaderOverlap (refined-github.js:3281:34)
    at default_branch_button_add (refined-github.js:5480:27)
    at refined-github.js:3561:9
```

The button doesn't appear:
<img width="319" alt="Screenshot 2024-07-15 at 7 07 16 PM" src="https://github.com/user-attachments/assets/edf442f6-200c-47e9-966d-06bd67179c0a">

### How to replicate the issue + URL

https://github.com/refined-github/refined-github/blob/07ecc75/extension/content.js

### Extension version

24.7.10

### Browser(s) used

Chrome 126.0.6478.127 on macOS Sonoma 14.5

fregante:
Indeed, it doesn't work when the window is very wide and the sidebar is pinned open.

PR #7600, opened 2024-07-26
`default-branch-button` - Restore feature in sidebar

kovsu:
close: #7549 

## Screenshot

When `default-branch-button` appears in the header, we can find the closet container element so it works fine!

image

But whetn it appears in the sidebar, the closet container element is null. And I think we don't need to add `rgh-z-index-5` class in this situation.

image

### DIFF
--- a/source/github-helpers/index.ts
+++ b/source/github-helpers/index.ts
@@ -160,7 +160,7 @@ export function triggerConversationUpdate(): void {
 
 // Fix z-index issue https://github.com/refined-github/refined-github/pull/7430
 export function fixFileHeaderOverlap(child: Element): void {
-	// When the default branch button appears in the sidebar, the closet container is null
+	// In the sidebar the container is not present and this fix is not needed
 	child.closest('.container')?.classList.add('rgh-z-index-5');
 }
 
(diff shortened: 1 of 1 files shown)


kovsu:
after:

image

source/github-helpers/index.ts
```diff
@@ -160,7 +160,11 @@ export function triggerConversationUpdate(): void {
 
 // Fix z-index issue https://github.com/refined-github/refined-github/pull/7430
 export function fixFileHeaderOverlap(child: Element): void {
-	child.closest('.container')!.classList.add('rgh-z-index-5');
```
fregante, line 163: You can just turn the exclamation mark into ?.

kovsu: Oh, I forget its ts file. hhhhh

source/github-helpers/index.ts
```diff
@@ -160,7 +160,7 @@ export function triggerConversationUpdate(): void {
...
+	child.closest('.container')?.classList.add('rgh-z-index-5');
```
fregante, line 163: We still want the comment here though, to clarify where it happens

fregante (approved):

source/github-helpers/index.ts
```diff
@@ -160,7 +160,8 @@ export function triggerConversationUpdate(): void {
 
 // Fix z-index issue https://github.com/refined-github/refined-github/pull/7430
 export function fixFileHeaderOverlap(child: Element): void {
-	child.closest('.container')!.classList.add('rgh-z-index-5');
+	// When the default branch button appears in the sidebar, the closet container is null
```
fregante, line 163:
```suggestion
	// In the sidebar the container is not present and this fix is not needed
```

PR #7602, opened 2024-07-26
`rgh-improve-new-issue-form` - Fix disabled button

fregante:
I regularly have issues when reporting bugs on Refined GitHub: the "Submit issue" button remains disabled until I touch every field (rather, the version field)

This triggers the right events on the field rather than silently setting its value. Maybe GitHub notices it now.

I can't repro the bug though, it just happens 

URL: https://github.com/refined-github/refined-github/issues/new?assignees=&labels=bug&projects=&template=1_bug_report.yml


fregante:
No, the field remains focused :( that shouldn't happen. Probably due to

- https://github.com/fregante/text-field-edit/issues/21

PR #7603, opened 2024-07-26
Format feature names in titles

### LINKED ISSUE #7014: Auto-format feature names in titles and comments
Sometimes features are mentioned without using the proper formatting, so linkification fails:

- ✅ `linkify-code`
- ❌ linkify-code
- ❌ linkify code

We can fix this on a repo level via workflow that edits titles and comments like https://github.com/refined-github/refined-github/blob/main/.github/workflows/labeler.yml

Possible solution:

1. Listen to new comments and titles (no edits)
1. Get feature list using https://github.com/refined-github/refined-github/blob/main/build/readme-parser.ts
2. Run a search/replace with the feature list

Ideally this logic is in a standalone .ts file and has its own .test.ts file so that we can ensure we don't break it.

Example scenario where this workflow would have run: https://github.com/refined-github/refined-github/issues/7010#event-10828153656

fregante:
- Close #7014 

I published:

- https://github.com/fregante/title-replacer-action
- https://github.com/marketplace/actions/title-replacer 

I think this action is useful in any repo that formats keywords, for example monorepos that format package names and eslint plugins that have multiple rules.

The action is not done and it needs a way to read from a file list and ideally from a directory.

The action was already tested on this repo 😍 

Screenshot 2


fregante:
Merging as is for now because I would have need it just now 🥹 #7604

PR #7605, opened 2024-07-26
Restore tooltip on reactions in locked conversations

### LINKED ISSUE #7527: Reactions' tooltip box doesn't appear for locked conversations
### Description

Hovering the cursor over a reaction box for locked conversations doesn't show a tooltip box that lists usernames unlike when the extension is disabled.

Example of a locked conversation: https://github.com/golang/go/issues/45729
Screenshot 2024-07-05 at 20 14 14

Example of a non-locked conversation: https://github.com/golang/go/issues/52188
Screenshot 2024-07-05 at 20 13 55

### Extension version

24.7.1

### Browser(s) used

Google Chrome 126.0.6478.127 macOS arm64

zzJinux:
Closes #7527

## Test URLs

https://github.com/golang/go/issues/45729#issue-866319435

## Screenshot
tooltip


zzJinux:
@fregante You mean the public URL?: https://github.com/golang/go/issues/45729#issue-866319435

source/features/github-bugs.css
```diff
@@ -51,10 +51,14 @@
 /* Mute unclickable reaction buttons, disable hover */
 /* Test: https://github.com/refined-github/sandbox/pull/48 */
 .social-reaction-summary-item[disabled] {
-	pointer-events: none;
+	pointer-events: auto;
```
fregante: This is actually the default, so I dropped it

source/features/github-bugs.css
```diff
@@ -51,10 +51,14 @@
...
 	filter: grayscale(0.5);
 }
 
+.social-reaction-summary-item:hover[disabled] {
```
fregante: `:hover` is not necessary I think, it can just be part of the main rule. If the selector is not strong enough, we can use `:root .social-reaction-summary-item[disabled]` for the whole rule.

fregante:
Thank you! The solution appears to work however it doesn't cover your own reactions, they do have a background and it's not `initial`

zzJinux:
Covering own reactions case:

https://github.com/user-attachments/assets/c26da078-a820-4758-a2bd-964464861aa1

---

Style rule changes sum up:

```css
.social-reaction-summary-item[disabled] {
	// pointer-events property is deleted
	filter: grayscale(0.5);

	// A CSS property for disabled reactions without your own.
	// This comes from the ".social-reaction-summary-item" rule.
	background-color: transparent;
}

.social-reaction-summary-item.user-has-reacted[disabled] {
	// CSS properties for disabled reactions _with_ your owns.
	// These come from the ".social-reaction-summary-item.user-has-reacted" rule.

	color: var(--reactionButton-selected-fgColor-rest, var(--color-accent-fg));

	// "!important" is needed to defeat the ".social-reaction-summary-item.user-has-reacted:hover" rule.
	background-color: var(--reactionButton-selected-bgColor-rest, var(--color-accent-subtle)) !important;
}
```

zzJinux:
I intentionally leave the comments in the code as-is.

source/features/github-bugs.css
zzJinux: Indeed it was. There was no selector weakness issue too.

fregante (changes requested):
It doesn't work for me:

c

Maybe the selector isn't strong enough. Try prepending `:root` to line 53 as I suggested.

fregante:
Indeed it works on this PR (now locked). Thank you!

c

Issue #7611, opened 2024-07-28
Feature `fix-no-pr-search` broken


fregante:
### Description

🤷‍♂️ 

### How to replicate the issue + URL

1. Visit https://github.com/refined-github/refined-github/pulls?q=sort%3Aupdated-desc+is%3Apr+is%3Aopen
2. Clear issue search field
3. Press enter
4. No issues appear, it's still on /pulls

### Extension version

24.7.10

### Browser(s) used

Safari

PR #7614, opened 2024-07-29
`fix-no-pr-search` - Restore feature

kovsu:
Fixes: #7611 

## Screenshot

https://github.com/user-attachments/assets/3e1fa0d5-fec4-46a6-b90f-8ba6b4378723

### DIFF
--- a/source/features/fix-no-pr-search.tsx
+++ b/source/features/fix-no-pr-search.tsx
@@ -6,10 +6,19 @@ import SearchQuery from '../github-helpers/search-query.js';
 
 function redirectToIssues(event: DelegateEvent<Event, HTMLFormElement>): void {
 	const form = event.delegateTarget;
-	const searchQuery = new SearchQuery(form.elements['js-issues-search'].value);
+	const searchValue = form.elements['js-issues-search'].value;
+	let redirect = true;
 
-	if (!searchQuery.includes('is:pr')) {
+	if (searchValue) {
+		const searchQuery = new SearchQuery(searchValue);
+		if (searchQuery.includes('is:pr')) {
+			redirect = false
+		}
+	}
+
+	if (redirect) {
 		form.action = form.action.replace(/\/pulls$/, '/issues');
+		form.submit();
 	}
 }
 
--- a/source/features/fix-no-pr-search.tsx
+++ b/source/features/fix-no-pr-search.tsx
@@ -5,6 +5,8 @@ import features from '../feature-manager.js';
 import SearchQuery from '../github-helpers/search-query.js';
 
 function redirectToIssues(event: DelegateEvent<Event, HTMLFormElement>): void {
+	event.preventDefault();
+
 	const form = event.delegateTarget;
 	const searchValue = form.elements['js-issues-search'].value;
 	let redirect = true;
--- a/source/features/fix-no-pr-search.tsx
+++ b/source/features/fix-no-pr-search.tsx
@@ -5,13 +5,14 @@ import features from '../feature-manager.js';
 import SearchQuery from '../github-helpers/search-query.js';
 
 function redirectToIssues(event: DelegateEvent<Event, HTMLFormElement>): void {
-	event.preventDefault();
 	const form = event.delegateTarget;
-	const query = SearchQuery.from(location)
+	const query = SearchQuery.from(location);
 	query.set(form.elements['js-issues-search'].value);
 
 	if (!query.includes('is:pr')) {
 		form.action = form.action.replace(/\/pulls$/, '/issues');
+		// Prevent submission via AJAX and use .submit() to allow the change from /pulls to /issues
+		event.preventDefault();
 		form.submit();
 	}
 }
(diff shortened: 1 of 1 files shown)


source/features/fix-no-pr-search.tsx
```diff
@@ -5,11 +5,22 @@ import features from '../feature-manager.js';
 import SearchQuery from '../github-helpers/search-query.js';
 
 function redirectToIssues(event: DelegateEvent<Event, HTMLFormElement>): void {
+	event.preventDefault();
+
 	const form = event.delegateTarget;
-	const searchQuery = new SearchQuery(form.elements['js-issues-search'].value);
+	const searchValue = form.elements['js-issues-search'].value;
```
fregante, line 11: If this is missing, just return early, before calling preventDefault. No need for `redirect`. I don’t think the previous feature structure need to be changed this much. 

In what case is it missing by the way?

source/features/fix-no-pr-search.tsx
```diff
@@ -5,11 +5,22 @@ import features from '../feature-manager.js';
 import SearchQuery from '../github-helpers/search-query.js';
 
 function redirectToIssues(event: DelegateEvent<Event, HTMLFormElement>): void {
+	event.preventDefault();
```
fregante, line 8: What happens if this is missing? Is GitHub going ahead and loading it via ajax?

source/features/fix-no-pr-search.tsx
kovsu: if `url === ''`, it will show a error like 👇

image

image

source/features/fix-no-pr-search.tsx
kovsu: like this 👇.

https://github.com/user-attachments/assets/19b9086d-42b6-442d-bc97-9f745aa29673

source/features/fix-no-pr-search.tsx
fregante: I'm gonna guess that this feature never worked, SearchQuery accepts a full URL, it shouldn't take random strings from this input.

I think the correct usage is:

```js
const query = SearchQuery.from(location)
query.set(form.elements['js-issues-search'].value);
```

source/features/fix-no-pr-search.tsx
fregante: > Is GitHub going ahead and loading it via ajax?

The answer is yes. They ajaxed the form.

kovsu: If I don't add `submit()` the active tab will not change, it stay in pull requests

kovsu: image

source/features/fix-no-pr-search.tsx
kovsu: done!

source/features/fix-no-pr-search.tsx
fregante: Got it 🙏 thanks!

source/features/fix-no-pr-search.tsx
```diff
@@ -5,11 +5,14 @@ import features from '../feature-manager.js';
...
 	const form = event.delegateTarget;
-	const searchQuery = new SearchQuery(form.elements['js-issues-search'].value);
+	const query = SearchQuery.from(location)
+	query.set(form.elements['js-issues-search'].value);
 
-	if (!searchQuery.includes('is:pr')) {
+	if (!query.includes('is:pr')) {
 		form.action = form.action.replace(/\/pulls$/, '/issues');
+		form.submit();
```
fregante, line 15: You most likely want to move it here

```suggestion
		// Prevent submission via AJAX and use .submit() to allow the change from /pulls to /issues
		event.preventDefault();
		form.submit();
```

kovsu: Learn a lot from you! Appreciate it :P

source/features/fix-no-pr-search.tsx
```diff
@@ -6,10 +6,14 @@ import SearchQuery from '../github-helpers/search-query.js';
...
 function redirectToIssues(event: DelegateEvent<Event, HTMLFormElement>): void {
 	const form = event.delegateTarget;
-	const searchQuery = new SearchQuery(form.elements['js-issues-search'].value);
+	const query = SearchQuery.from(location);
+	query.set(form.elements['js-issues-search'].value);
 
-	if (!searchQuery.includes('is:pr')) {
+	if (!query.includes('is:pr')) {
 		form.action = form.action.replace(/\/pulls$/, '/issues');
+		// Prevent submission via AJAX and use .submit() to allow the change from /pulls to /issues
```
fregante, line 14:
```suggestion
		// Prevent submission via AJAX and use .submit() to allow the change from /pulls to /issues
		// https://github.com/refined-github/refined-github/pull/7614/files#r1694731354
```

fregante:
If you tested the latest version and it works, this is good to go 👍

kovsu:
@fregante It works fine in the latest version

Issue #7618, opened 2024-07-30
`default-branch-button` doesn't appear in sticky header


fregante:
### Description

<img width="403" alt="Screenshot" src="https://github.com/user-attachments/assets/f556b856-c7d3-47b1-babc-f7776252c476">

Related:

- https://github.com/refined-github/refined-github/issues/7549

cc @kovsu if you're interested

### How to replicate the issue + URL

https://github.com/refined-github/refined-github/blob/9803e03037196494aba814912ec720e7f5475758/source/features/tags-on-commits-list.tsx#L155

### Extension version

24.7.10

### Browser(s) used

Safari

fregante:
`attachElement` could be used instead of the `elementExist` since it has its own integrated local check, but I don't know if it can handle the theme change https://github.com/refined-github/refined-github/pull/7623#issuecomment-2264977339

PR #7623, opened 2024-07-31
`default-branch-button` - Restore feature in sticky header

kovsu:
close: #7618 

## Screenshot

image


fregante (approved):

fregante:
There was a reason why it was written that way :(

c

fregante:
You can see how I had to deal with the page having two buttons:

https://github.com/refined-github/refined-github/blob/b28e8335564ec6f4c652084d08cd813d078cc139/source/features/previous-version.tsx#L64

Basically each button needs its own class in order to be deduplicated correctly

Issue #7616, opened 2024-07-30
Extend `fix-no-pr-search` to work on global lists


fregante:
### Description

Removing `is:pr` from here won't work: https://github.com/pulls

### Example URLs

https://github.com/pulls

fregante:
@kovsu

kovsu:
I can't reproduce it. Can you provide me a gif or a video?

fregante:
1. Open https://github.com/pulls
2. Delete `is:pr` from search field
3. Press enter
4. You should end up on a list that contains all your ISSUES and PRs https://github.com/issues?q=sort%3Aupdated-desc+is%3Aopen+author%3A%40me+archived%3Afalse

kovsu:
@fregante Got it!

PR #7625, opened 2024-07-31
`fix-no-pr-search` - Enable on global PR list

kovsu:
close: #7616 

https://github.com/user-attachments/assets/ad7c9370-0a34-49a8-bc96-62357cb6fa56

### DIFF
--- a/source/features/fix-no-pr-search.tsx
+++ b/source/features/fix-no-pr-search.tsx
@@ -28,7 +28,7 @@ function init(signal: AbortSignal): void {
 void features.add(import.meta.url, {
 	include: [
 		pageDetect.isRepoPRList,
-		pageDetect.isPRList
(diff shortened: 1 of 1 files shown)


source/features/fix-no-pr-search.tsx
```diff
@@ -28,6 +28,7 @@ function init(signal: AbortSignal): void {
 void features.add(import.meta.url, {
 	include: [
 		pageDetect.isRepoPRList,
```
fregante, line 30: I think this can be dropped

Can you also add a test URL at the bottom?

fregante (approved):

source/features/fix-no-pr-search.tsx
kovsu: 👌

Issue #7622, opened 2024-07-30
`mark-private-orgs`: Outdated screenshot


injust:
### Description

The [current example screenshot](https://user-images.githubusercontent.com/6775216/44633467-d5dcc900-a959-11e8-9116-e6b0ffef66af.png) is outdated. It shows a padlock icon, but the current icon is the `eye-closed` Octicon.

### How to replicate the issue + URL

https://github.com/refined-github/refined-github/blob/main/source/features/mark-private-orgs.tsx

### Extension version

24.7.23

### Browser(s) used

Chrome on macOS Sonoma 14.5

fregante:
![Screenshot 21](https://github.com/user-attachments/assets/145a7a97-7b8c-4ac4-8288-f72dcb4613ea)

PR #7626, opened 2024-08-01
`mark-private-orgs` - replace eyeclose icon with lock icon

kovsu:
close: #7622


fregante:
The screenshot is out of date, that's why I marked the issue as Meta instead of Bug. The icon change was intentional:

- https://github.com/refined-github/refined-github/pull/2788

PR #7635, opened 2024-08-02
`previous-version` - Show on mobile

fregante:
- Fixes https://github.com/refined-github/refined-github/issues/7540

## Test URLs

https://github.com/refined-github/refined-github/blob/main/source/background.ts

## Screenshot
c


fregante:
The code required is so ugly 🥲👹

PR #7653, opened 2024-08-03
Add required "token" checkbox when reporting bugs

fregante:
Mentioned in https://github.com/refined-github/refined-github/issues/7231

Screenshot


fregante:
Nothing works. 

- https://github.com/refined-github/refined-github/issues/7659#issuecomment-2267489442

PR #7661, opened 2024-08-06
`default-branch-button` - Restore feature in sticky header

kovsu:
close: #7618 

this pr fix two problem
1. #7618 
2. when switch the theme, the default branch button lost its radius style.

image

## Screenshot

👇 here is the video that test all test urls.
[Google Chrome.mp4.zip](https://github.com/user-attachments/files/16504827/Google.Chrome.mp4.zip)

### DIFF
--- a/source/features/default-branch-button.tsx
+++ b/source/features/default-branch-button.tsx
@@ -34,8 +34,8 @@ const getUrl = memoize(async (currentUrl: string): Promise<string> => {
 async function add(branchSelector: HTMLElement): Promise<void> {
 	// The DOM varies between details-based DOM and React-based one
 	const selectorWrapper = branchSelector.tagName === 'SUMMARY'
-			? branchSelector.parentElement!
-			: branchSelector;
+		? branchSelector.parentElement!
+		: branchSelector;
 
 	let defaultLink = $('.rgh-default-branch-button', branchSelector.parentElement!);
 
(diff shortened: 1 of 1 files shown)


fregante:
You can upload the video directly, no need to zip it

source/features/default-branch-button.tsx
```diff
@@ -32,38 +32,37 @@ const getUrl = memoize(async (currentUrl: string): Promise<string> => {
 });
 
 async function add(branchSelector: HTMLElement): Promise<void> {
-	// React issues. Duplicates appear after a color scheme update
-	// https://github.com/refined-github/refined-github/issues/7098
-	if (elementExists('.rgh-default-branch-button')) {
-		return;
-	}
```
fregante, line 39: I'd rather keep this comment *and* this early return. 

To fix the grouping issue, you can copy-paste the groupButtons call here (or make a function)

source/features/default-branch-button.tsx
```diff
@@ -32,38 +32,37 @@ const getUrl = memoize(async (currentUrl: string): Promise<string> => {
...
+				// data-turbo-frame="repo-content-turbo-frame"
+			>
+				<ChevronLeftIcon/>
+			</a>
+		);
+
+		selectorWrapper.before(defaultLink);
+	}
+
 	groupButtons([defaultLink, selectorWrapper]).classList.add('d-flex', 'rgh-default-branch-button-group');
```
fregante, line 66: Can you check why we added this "group" class? It doesn't seem to be used anywhere else. Only the flex box is needed, probably

fregante (approved):

Issue #7646, opened 2024-08-02
`pr-branch-auto-delete` broken


fregante:
### Description

I think `onPrMerge` is broken because `init` is not being reached

### How to replicate the issue + URL

https://github.com/refined-github/sandbox/pull/87, but exactly when merging it

### Extension version

24.8.1

### Browser(s) used

Chrome

PR #7662, opened 2024-08-06
`pr-branch-auto-delete` - Restore feature

kovsu:
close: #7646 

We need all conditions in `asLongAs` are true, so that we can successfully get `runFeature` function. So we need to add `awaitDomReady: true` to make `canCreateRelease` to be `true`.

image

### DIFF
--- a/source/features/pr-branch-auto-delete.tsx
+++ b/source/features/pr-branch-auto-delete.tsx
@@ -43,7 +43,7 @@ void features.add(import.meta.url, {
 	additionalListeners: [
 		onPrMerge,
 	],
-	awaitDomReady: true,
+	awaitDomReady: true, // TODO: Remove after https://github.com/refined-github/refined-github/issues/6566
 	onlyAdditionalListeners: true,
 	init,
 });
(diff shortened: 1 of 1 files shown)


source/features/pr-branch-auto-delete.tsx
```diff
@@ -43,6 +43,16 @@ void features.add(import.meta.url, {
...
 	onlyAdditionalListeners: true,
 	init,
 });
+
+/*
+
+Test URLs:
+
+https://github.com/USERNAME/REPO_NAME/pull/PR_NUMBER
+https://github.com/ORGANIZATION_NAME/REPO_NAME/pull/PR_NUMBER
```
fregante, line 56: Test URLs must be real URLs or easy steps to find them. In this case:

```
1. Open https://github.com/pulls
2. Click on any PRs you can merge (in repositories without native auto-delete)
3. Merge the PR
```

fregante (approved):

source/features/pr-branch-auto-delete.tsx
```diff
@@ -43,6 +43,16 @@ void features.add(import.meta.url, {
 	additionalListeners: [
 		onPrMerge,
 	],
+	awaitDomReady: true,
```
fregante, line 46: Can we add a TODO comment saying that this can be removed after:

- https://github.com/refined-github/refined-github/issues/6566

Because then the check will be async (which asLongAs will await)

Blackman99:
Please **DO NOT** enable this option as default!!

PR #7666, opened 2024-08-09
`show-names` - Fix comment header cleanup

kovsu:
before:

image

image

after:

image

### DIFF
--- a/source/features/show-names.tsx
+++ b/source/features/show-names.tsx
@@ -55,7 +55,11 @@ async function updateLink(batchedUsernameElements: HTMLAnchorElement[]): Promise
 
 		// This change is ideal but should not break the feature if it fails
 		// And it should be done before inserting the name
-		dropExtraCopy(usernameElement);
+		try {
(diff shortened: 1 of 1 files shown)


source/features/show-names.tsx
```diff
@@ -54,6 +52,11 @@ async function updateLink(batchedUsernameElements: HTMLAnchorElement[]): Promise
 		// If it's a regular comment author, add it outside <strong> otherwise it's something like "User added some commits"
 		const {parentElement} = usernameElement;
 		const insertionPoint = parentElement!.tagName === 'STRONG' ? parentElement! : usernameElement;
+
+		// This change is ideal but should not break the feature if it fails
+		// And it should be done before inserting the name
+		dropExtraCopy(usernameElement);
```
fregante, line 58: This chance goes against the previous comment: it should be last specifically to avoid breaking the entire feature when this piece breaks. 

If you prefer, update the comment and wrap this call in a try/catch (+ `feature.log.error` call in case of errors)

Issue #7670, opened 2024-08-12
`warn-pr-from-master`: select_dom_$(...) is undefined


Seregy:
### Description

```
init@moz-extension://08fad806-d697-4a80-bb11-b8095e9d3105/assets/refined-github.js:6487:59
setupPageLoad/runFeature/<@moz-extension://08fad806-d697-4a80-bb11-b8095e9d3105/assets/refined-github.js:3122:28
setupPageLoad/runFeature/<@moz-extension://08fad806-d697-4a80-bb11-b8095e9d3105/assets/refined-github.js:3118:72
setupPageLoad/runFeature/<@moz-extension://08fad806-d697-4a80-bb11-b8095e9d3105/assets/refined-github.js:3118:45
runFeature@moz-extension://08fad806-d697-4a80-bb11-b8095e9d3105/assets/refined-github.js:3119:10
setupPageLoad@moz-extension://08fad806-d697-4a80-bb11-b8095e9d3105/assets/refined-github.js:3134:43
```

Fork repository selector for the cross-fork PRs seems to be no longer valid.
The feature attempts to extract the repository name from the element selected via the `[title^="head: "]`, which worked with the previous structure:
```html
<span class="css-truncate css-truncate-target" data-menu-button="" title="head: username/repo-name">username/repo-name</span>
```
But now the element with the repository name no longer contains the `title` attribute:
```html
<span class="css-truncate css-truncate-target" data-menu-button="">username/repo-name</span>
```

### How to replicate the issue + URL

Try to compare the changes across forks

https://github.com/refined-github/refined-github/compare/sandbox/keep-branch...yakov116:upstream

### Extension version

24.8.3

### Browser(s) used

Firefox, Chrome

### Token ⚠️

- [X] I set a token in the options page. I understand that some features require a token to work.

PR #7683, opened 2024-08-14
`warn-pr-from-master` - Restore feature

kovsu:
Closes: #7670 

## Test URLs

https://github.com/refined-github/refined-github/compare/sandbox/keep-branch...yakov116:upstream

## Screenshot

image


source/features/warn-pr-from-master.tsx
```diff
@@ -38,3 +43,11 @@ void features.add(import.meta.url, {
 	deduplicate: 'has-rgh',
```
fregante, line 43: Both of these can be dropped.

Also, is the banner removed/hidden when the user selects another branch?

fregante (approved):

kovsu:
https://github.com/user-attachments/assets/e0b6c0b7-f14f-46ed-b90a-04769960eaa5

source/features/warn-pr-from-master.tsx
```diff
@@ -27,14 +28,24 @@ async function init(): Promise<false | void> {
...
-	awaitDomReady: true,
-	deduplicate: 'has-rgh',
 	init,
 });
+
+/*
+
+Test URLs
+
+https://github.com/refined-github/refined-github/compare/sandbox/keep-branch...yakov116:upstream
```
fregante, line 49: The `upstream` name is confusing, let's use this one

```suggestion
https://github.com/refined-github/refined-github/compare/main...fregante:refined-github:main
```

source/features/warn-pr-from-master.tsx
```diff
@@ -27,14 +28,24 @@ async function init(): Promise<false | void> {
...
+	observe('.is-cross-repo', warnPR, {signal});
+}
+
 void features.add(import.meta.url, {
 	include: [
 		pageDetect.isCompare,
 	],
 	exclude: [
 		pageDetect.isBlank,
 	],
```
fregante, line 41: This can also be dropped, see https://github.com/refined-github/refined-github/issues/7000#issuecomment-2287762931

```suggestion
```

fregante (approved):

source/features/warn-pr-from-master.tsx
```diff
@@ -1,15 +1,16 @@
...
 import {getRepo} from '../github-helpers/index.js';
+import observe from '../helpers/selector-observer.js';
 
-async function init(): Promise<false | void> {
+async function warnPR(crossRepo: Element): Promise<false | void> {
 	let defaultBranch;
-	if (elementExists('.is-cross-repo')) {
-		const forkedRepository = getRepo($('[title^="head: "]')!.textContent)!;
+	if (crossRepo) {
+		const forkedRepository = getRepo($('.range-cross-repo-pair:has(#head-ref-selector) span[data-menu-button]')!.textContent)!;
```
fregante, line 13: This needs further changes actually. This condition is now always true, so `else` is never called.

But I don’t think it ever made sense, it was added in https://github.com/refined-github/refined-github/pull/3511

So:

- drop the `if(crossRepo)` line
- drop the entire `else` part
- Move the `data-menu-button` selector to the observer

fregante (changes requested):

source/features/warn-pr-from-master.tsx
fregante: I'm surprised that neither TS nor ESLint is complaining that the condition is always truthy 😬

source/features/warn-pr-from-master.tsx
```diff
@@ -1,15 +1,16 @@
...
+import {$} from 'select-dom';
 import * as pageDetect from 'github-url-detection';
 
 import features from '../feature-manager.js';
 import getDefaultBranch, {defaultBranchOfRepo} from '../github-helpers/get-default-branch.js';
 import {getRepo} from '../github-helpers/index.js';
+import observe from '../helpers/selector-observer.js';
 
-async function init(): Promise<false | void> {
+async function warnPR(crossRepo: Element): Promise<false | void> {
```
fregante, line 10: This this is no longer an `init` function, it doesn't need to return `false` anymore`, it can just `return;`

```suggestion
async function warnPR(crossRepo: Element): Promise<void> {
```

source/features/warn-pr-from-master.tsx
fregante: This is weird actually, we don't need to read this data from the DOM, it's already in the URL, it's being checked just later.

So probably:

- keep `observe('.is-cross-repo'`
- read `forkedRepository` from the `...user:repo` URL part instead


Also make sure that renamed repos are supported, test URL:

- https://github.com/refined-github/sandbox/compare/default-a...bfred-it-org:github-sandbox:main?expand=1

fregante:
> I think it reload the page. Is it necessary to use observe?

You're right, it isn't, but it allows us to call the API earlier. Maybe we can drop the observer (which isn't necessary since all the info we need is in the URL) and use elementReady() instead of $()

If it works as is, for both test URLs, it's fine to merge, but ideally the feature should be rewritten:

- extract a function to get the head branch/repo information from the URL 
- add tests
- perhaps move this custom condition to asLongAs instead of the init
- use the new element ready function based on the observer (I forget the name, but it's in the same observer source file

Basically:

```js
async function init() {
  const anchor = await elementReady('the position')
  anchor.before(our banner)
}

async function isCrossRepoCompareFromMaster() {
  const {base, head} = parseCompareUrl(location.pathname)
  return head.repo !== base.repo && head.branch === await getDefaultBranch(head.repo)
}

features.add({
  init,
  asLongAs: [
    isCompare,
    isCrossRepoCompareFromMaster,
  ],
})
```

fregante (approved):

kovsu:
@fregante I will try it after this pr merge. I think first thing we need to do is to get the feature working.

fregante (changes requested):
I think it broke in a recent commit. This is why I was suggesting to just parse the URL instead

Screenshot 1

kovsu:
https://github.com/user-attachments/assets/6cf07101-d965-4c53-bf22-83ebdb32c66b

kovsu:
I mess it up 😂. Open a new pr.

Issue #7682, opened 2024-08-13
`tags-on-commits-list`: Cannot read properties of undefined (reading 'pathname')


eranb-di:
### Description

```
TypeError: Cannot read properties of undefined (reading 'pathname')
    at getCommitHash (chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/assets/refined-github.js:6125:35)
    at init (chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/assets/refined-github.js:7411:134)
    at chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/assets/refined-github.js:3119:28
    at chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/assets/refined-github.js:3115:64
    at Array.map (<anonymous>)
    at chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/assets/refined-github.js:3115:45
    at runFeature (chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/assets/refined-github.js:3116:10)
    at setupPageLoad (chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/assets/refined-github.js:3131:43)
    at async chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/assets/refined-github.js:3156:11
```

### How to replicate the issue + URL

https://github.com/refined-github/refined-github/commits/main/

### Extension version

24.8.13

### Browser(s) used

Google Chrome 127.0.6533.100 (Official Build) (64-bit)

### Token ⚠️

- [X] I set a token in the options page. I understand that some features require a token to work.

eranb-di:
Updated the browser type
Thanks :)

kovsu:
<img width="500" alt="image" src="https://github.com/user-attachments/assets/4ec7b4b3-8601-47ec-b95a-8a03c886f80c">


the `.list-view-item` selector all be affected.
 
https://github.com/search?q=repo%3Arefined-github%2Frefined-github%20.list-view-item&type=code

fregante:
I think we can use `[role="listitem"]`, but a bit more specific to the commit list

kovsu:
Will it be dropped one day if use `[data-testid="commit-row-item"]` ?

fregante:
Nothing is guaranteed. We can try though

PR #7685, opened 2024-08-14
`tags-on-commits-list` + `dim-bots` - Restore features

kovsu:
Closes:  #7682

## Test URLs

https://github.com/typed-ember/ember-cli-typescript/commits/master?after=5ff0c078a4274aeccaf83382c0d6b46323f57397+174

## Screenshot

image

### DIFF
--- a/source/features/dim-bots.tsx
+++ b/source/features/dim-bots.tsx
@@ -27,9 +27,10 @@ function undimBots(event: DelegateEvent): void {
 
 function dim(commit: HTMLElement): void {
 	commit.closest([
-		'[data-testid="commit-row-item"]',
 		'.list-view-item', // Commits
 		'.listviewitem', // TODO: Drop in 2025
+		'[data-testid="commit-row-item"]',
+
 		'.Box-row', // PRs
 	])!.classList.add(dimBots.class);
 }
--- a/source/features/dim-bots.tsx
+++ b/source/features/dim-bots.tsx
@@ -27,6 +27,7 @@ function undimBots(event: DelegateEvent): void {
 
 function dim(commit: HTMLElement): void {
 	commit.closest([
+		'[data-testid="commit-row-item"]',
 		'.list-view-item', // Commits
 		'.listviewitem', // TODO: Drop in 2025
 		'.Box-row', // PRs
--- a/source/features/dim-bots.tsx
+++ b/source/features/dim-bots.tsx
@@ -27,8 +27,6 @@ function undimBots(event: DelegateEvent): void {
 
 function dim(commit: HTMLElement): void {
 	commit.closest([
-		'.list-view-item', // Commits
-		'.listviewitem', // TODO: Drop in 2025
 		'[data-testid="commit-row-item"]',
 
 		'.Box-row', // PRs
--- a/source/features/mark-merge-commits-in-list.tsx
+++ b/source/features/mark-merge-commits-in-list.tsx
@@ -37,7 +37,8 @@ const filterMergeCommits = async (commits: string[]): Promise<string[]> => {
 
 function getCommitLink(commit: HTMLElement): HTMLAnchorElement | undefined {
 	return $([
-		'[data-testid="list-view-item-metadata-item"]:last-child a',
+		'a.markdown-title', // Old view style (before November 2023)
+		'.markdown-title a',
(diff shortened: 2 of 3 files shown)


source/features/mark-merge-commits-in-list.tsx
```diff
@@ -27,7 +27,7 @@ const filterMergeCommits = async (commits: string[]): Promise<string[]> => {
 
 	const mergeCommits = [];
 	for (const [key, commit] of objectEntries(repository)) {
-		if (commit.parents.totalCount >= 2) {
+		if (commit?.parents.totalCount >= 2) {
```
kovsu, line 30: here is to fix below.

CleanShot 2024-08-14 at 11 26 02@2x

source/features/dim-bots.tsx
```diff
@@ -27,6 +27,7 @@ function undimBots(event: DelegateEvent): void {
 
 function dim(commit: HTMLElement): void {
 	commit.closest([
+		'[data-testid="commit-row-item"]',
 		'.list-view-item', // Commits
```
fregante, line 31:
```suggestion
		'[data-testid="commit-row-item"]', // Commits
```

source/features/mark-merge-commits-in-list.tsx
fregante, line 30: How's this possible/necessary all of the sudden?

source/features/mark-merge-commits-in-list.tsx
```diff
@@ -68,6 +68,7 @@ async function markCommits(commits: HTMLElement[]): Promise<void> {
 
 async function init(signal: AbortSignal): Promise<void> {
 	observe([
+		'[data-testid="commit-row-item"]',
 		'.list-view-item', // `isCommitList`
```
fregante, line 72:
```suggestion
		'[data-testid="commit-row-item"]', // `isCommitList`
```

source/features/mark-merge-commits-in-list.tsx
kovsu: image

image

fregante: That looks like the selector isn't good enough and it's causing the extension to request invalid data to the API. It should be fixed instead of altering this line

kovsu: I think reason is here, but what caused the problem.
image

image

fregante: I think it's getCommitLink. It should use a different link, not the title. There's another link to the commit on the side (which is where the commit hash is, that's what we want)

source/features/mark-merge-commits-in-list.tsx
```diff
@@ -37,8 +37,7 @@ const filterMergeCommits = async (commits: string[]): Promise<string[]> => {
 
 function getCommitLink(commit: HTMLElement): HTMLAnchorElement | undefined {
 	return $([
-		'a.markdown-title', // Old view style (before November 2023)
-		'.markdown-title a',
+		'[data-testid="list-view-item-metadata-item"]:last-child a',
 	], commit);
```
fregante, line 41: Since there's only one selector we can remove the array. Also I wonder if there's a way to more precisely select the commit element there, for example by targeting the copy button and reading its attributes

fregante (approved):

source/features/tags-on-commits-list.tsx
```diff
@@ -86,7 +86,7 @@ async function getTags(lastCommit: string, after?: string): Promise<CommitTags>
 async function init(): Promise<void | false> {
 	const cacheKey = `tags:${getRepo()!.nameWithOwner}`;
 
-	const commitsOnPage = $$('.list-view-item');
```
fregante, line 89: Make sure to include test URLs for each feature changed.

e.g. https://github.com/refined-github/refined-github/commits/19.5.21.1921

source/features/mark-merge-commits-in-list.tsx
fregante, line 1: I had to discard changes to this feature because it was a bit wonky and it needed a better fix. The icon changed position (it was next to the title)

Screenshot 2

fregante (approved):

PR #12, opened 2016-02-16
Add `Releases` tab to the repo page

hkdobrev:
Closes #8

I've successfully added a releases tab.
However I have some concerns which I hope I'd be able to address with your help.

- Adding just `<span class="octicon octicon-tag">` does not render an icon. I needed to add a full SVG markup like the rest of the icons I could found on github.com.
- Appending the *Releases* tab and activating it instead of *Code* is rather slow and you could visually see the change after the page loads. Feedback appreciated.
- If you have an idea how to easily count the number of releases (or tags) to put next to the label on the tab, that would be great.

P.S. This also adds a new shortcut `g r` for navigating to the releases page of a repo :smiley_cat:

PR #7689, opened 2024-08-15
`suggest-commit-title-limit` - Enable on compare pages

karlhorky:
Fixes https://github.com/refined-github/refined-github/pull/7428#issuecomment-2290952228

## Test URLs

https://github.com/refined-github/refined-github/compare/main...karlhorky:refined-github:test/bun?expand=true

## Screenshot

### DIFF
--- a/source/features/suggest-commit-title-limit.tsx
+++ b/source/features/suggest-commit-title-limit.tsx
@@ -27,6 +27,7 @@ function init(signal: AbortSignal): void {
 void features.add(import.meta.url, {
 	include: [
 		pageDetect.isEditingFile,
+		pageDetect.isCompare,
 		pageDetect.isPR,
 	],
 	init,
--- a/source/features/suggest-commit-title-limit.tsx
+++ b/source/features/suggest-commit-title-limit.tsx
@@ -21,7 +21,10 @@ function validatePrTitle({delegateTarget: field}: DelegateEvent<Event, HTMLInput
 function init(signal: AbortSignal): void {
 	document.body.classList.add('rgh-suggest-commit-title-limit');
 	onCommitTitleUpdate(validateCommitTitle, signal);
-	delegate('#issue_title, #pull_request_title', 'input', validatePrTitle, {signal, passive: true});
+	delegate([
+		'#issue_title',
+		'#pull_request_title',
+	], 'input', validatePrTitle, {signal, passive: true});
 }
 
 void features.add(import.meta.url, {
--- a/source/features/suggest-commit-title-limit.tsx
+++ b/source/features/suggest-commit-title-limit.tsx
@@ -24,7 +24,7 @@ function init(signal: AbortSignal): void {
 	delegate([
 		'#issue_title',
 		'#pull_request_title',
-	], 'input', validatePrTitle, {signal, passive: true});
+	].join(', '), 'input', validatePrTitle, {signal, passive: true});
 }
 
 void features.add(import.meta.url, {
--- a/source/features/suggest-commit-title-limit.tsx
+++ b/source/features/suggest-commit-title-limit.tsx
@@ -21,7 +21,7 @@ function validatePrTitle({delegateTarget: field}: DelegateEvent<Event, HTMLInput
 function init(signal: AbortSignal): void {
 	document.body.classList.add('rgh-suggest-commit-title-limit');
 	onCommitTitleUpdate(validateCommitTitle, signal);
-	delegate('#issue_title', 'input', validatePrTitle, {signal, passive: true});
+	delegate('#issue_title, #pull_request_title', 'input', validatePrTitle, {signal, passive: true});
 }
 
 void features.add(import.meta.url, {
--- a/source/features/suggest-commit-title-limit.css
+++ b/source/features/suggest-commit-title-limit.css
@@ -18,7 +18,8 @@
 :is(
 		#commit-summary-input,
 		#merge_title_field,
-		#issue_title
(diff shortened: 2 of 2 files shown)


fregante:
Nope, not enough:

Screenshot

karlhorky:
Interesting, the input field appears to have an attribute `id="issue_title"` on a PR page eg. https://github.com/refined-github/refined-github/pull/7689

Screenshot 2024-08-18 at 17 21 08

It is `id="pull_request_title"` on the `compare/...?expand=true` pages eg. https://github.com/refined-github/refined-github/compare/main...karlhorky:refined-github:test/bun?expand=true

Screenshot 2024-08-18 at 17 21 30

So that leads me to believe that the selector needs to be changed:

- 08dd9aefe6cbb4140a0adcb1aa3306481e1cb28b

source/features/suggest-commit-title-limit.tsx
```diff
@@ -21,12 +21,13 @@ function validatePrTitle({delegateTarget: field}: DelegateEvent<Event, HTMLInput
 function init(signal: AbortSignal): void {
 	document.body.classList.add('rgh-suggest-commit-title-limit');
 	onCommitTitleUpdate(validateCommitTitle, signal);
-	delegate('#issue_title', 'input', validatePrTitle, {signal, passive: true});
+	delegate('#issue_title, #pull_request_title', 'input', validatePrTitle, {signal, passive: true});
```
fregante, line 24:
```suggestion
	delegate([
		'#issue_title', 
		'#pull_request_title',
	], 'input', validatePrTitle, {signal, passive: true});
```

source/features/suggest-commit-title-limit.tsx
```diff
@@ -21,12 +21,16 @@ function validatePrTitle({delegateTarget: field}: DelegateEvent<Event, HTMLInput
 function init(signal: AbortSignal): void {
 	document.body.classList.add('rgh-suggest-commit-title-limit');
 	onCommitTitleUpdate(validateCommitTitle, signal);
-	delegate('#issue_title', 'input', validatePrTitle, {signal, passive: true});
+	delegate([
+		'#issue_title',
+		'#pull_request_title',
+	], 'input', validatePrTitle, {signal, passive: true});
```
karlhorky, line 27: Does this need `.join(', ')` ?

```suggestion
	].join(', '), 'input', validatePrTitle, {signal, passive: true});
```

fregante: Sigh, I thought I had implemented that:

- https://github.com/fregante/delegate-it/issues/35

Anyway for now it's just a type error, all arrays are flattened to `string 1,string 2` before use

karlhorky: ah interesting, should I commit the suggestion for now anyway to get rid of the type error?

karlhorky:
@fregante anything else I can do in this PR to make it ready?

source/features/suggest-commit-title-limit.tsx
fregante: Ok

karlhorky:
Ah, I think I found a bug in my implementation:

```js
getConversationNumber('https://github.com/refined-github/refined-github/compare/main...karlhorky:refined-github:test/bun?expand=true')
// => undefined
```

https://github.com/refined-github/refined-github/blob/bfbcec1f7284c88c6421938a8d1744fc53e8c7e0/source/github-helpers/index.ts#L15-L18

https://github.com/refined-github/refined-github/blob/8a57d3bbd2902d34652742316097aa633a5a3575/source/features/sync-pr-commit-title.tsx#L24-L26

This will warn about a PR titles earlier than necessary, because the suffix will be ` (#undefined)` 😮

Maybe Refined GitHub has a way to get the "last" or "highest" PR / issue number? (hopefully without API / token access)

fregante:
> hopefully without requiring a token

👎 Refined GitHub does not try to avoid token usage because token-less usage is not supported: https://github.com/refined-github/refined-github/issues/7231



> This will warn about a PR titles earlier than necessary, because the suffix will be ` (#undefined)`

Good point. We have several options here:

- count a minimum of 6 characters off ` (#12)`
- fetch the latest issue/PR via GraphQL

I think being 2 characters off is not a big deal, so even counting 4-digits would be safe for most projects. No HTTP request needed

karlhorky:
Looks like it's possible like this - not sure if there is code like this already somewhere or whether [`api.v3` or `api.v4`](https://github.com/refined-github/refined-github/blob/bfbcec1f7284c88c6421938a8d1744fc53e8c7e0/source/github-helpers/api.ts#L61-L67) should be used instead:

```js
const response = await fetch(`https://api.github.com/repos/refined-github/refined-github/issues?sort=created&direction=desc&per_page=1`);
const result = await response.json();
console.log(result[0].number); // 7726
```

fregante:
That should look something like:

```js
const issues = await api.v3('issues?sort=created&direction=desc&per_page=1');
return issues[0].number;
```


But does `issues` include PRs as well?

karlhorky:
> But does `issues` include PRs as well?

Yep #7726 is a PR

I'll check whether that's the default sorting and open a PR

karlhorky:
PR open at:

- https://github.com/refined-github/refined-github/pull/7728

karlhorky:
Thanks for the release of [`24.8.26`](https://github.com/refined-github/refined-github/releases/tag/24.8.26), works well!

- https://x.com/karlhorky/status/1828488263868317713

PR #7692, opened 2024-08-17
`linkify-labels-on-dashboard` - Drop feature

fregante:
I don't see any labels on the dashboard. Even if there are, I don’t think it's worth keeping this feature around. https://github.com/refined-github/refined-github/issues/4740 is open to create a more generic linkifier feature that could address this and more, if you see it again.

Historical note: it's not the first time we drop the feature 😃 

1. https://github.com/refined-github/refined-github/pull/6016
1. https://github.com/refined-github/refined-github/pull/6036


fregante:
Noticed when working on

- https://github.com/refined-github/refined-github/issues/7656

PR #7693, opened 2024-08-17
Meta: Add more test URLs

fregante:
- Part of https://github.com/refined-github/refined-github/issues/7656

### DIFF
--- a/source/feature-renames.json
+++ b/source/feature-renames.json
@@ -1,4 +1,5 @@
 {
+	"scrollable-code-and-blockquote": "scrollable-areas",
 	"separate-draft-pr-button": "one-click-pr-or-gist",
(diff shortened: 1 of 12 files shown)


source/feature-renames.json
```diff
@@ -1,4 +1,5 @@
 {
+	"scrollable-code-and-blockquote": "scrollable-areas",
```
fregante, line 2: It would be great to add all previous names in all the rename PRs that came before this rename tracking: https://github.com/refined-github/refined-github/pulls?q=is%3Apr+is%3Aclosed+rename+in%3Atitle+sort%3Acreated-asc

But it's hard to find them, especially because some features have been dropped altogether

PR #7695, opened 2024-08-18
Increase size of profile picture in user hovercard

### LINKED ISSUE #7481: Increase size of profile picture to user hovercard
### Description

I think it would be cool if there was an option to slightly increase the size of the picture in the profile hovercard. It's currently nearly the same size as the hover target.

Here's a quick CSS solution I threw together:

**Before:**

Refined Github profile hover view

**After:**

Modified Refined Github profile hover view

### Example URLs

https://github.com

Andrew6rant:
This PR adds 4 lines of CSS to `refined-github.css`, increasing the size of profile pictures in hovercards by 150%. Closes #7481.

## Test URLs

## Screenshots
Before:
refined-github-before
After:
refined-github-after

Opening this PR as a draft, because in rare cases it can cut off the top of the hovercard (reproduce by hovering over the avatar of a user with a two or more line bio, in the first comment of an issue or PR).

refined-github-cut-off

### DIFF
--- a/source/refined-github.css
+++ b/source/refined-github.css
@@ -174,7 +174,8 @@ div[data-target='readme-toc.content'] div.blob-wrapper img {
 	z-index: 5;
 }
 
-/* Increase size of profile picture in user hovercard by 1.5x (default is 48px)  */
+/* Increase size of profile picture in user hovercard by 1.5x (default is 48px) #7695 */
+/* Test: https://github.com/refined-github/refined-github */
 .user-hovercard-avatar > img.d-block.avatar-user {
 	width: 72px;
 	height: 72px;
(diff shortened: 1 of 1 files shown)


source/refined-github.css
```diff
@@ -173,3 +173,9 @@ div[data-target='readme-toc.content'] div.blob-wrapper img {
 	position: relative;
 	z-index: 5;
 }
+
+/* Increase size of profile picture in user hovercard by 1.5x (default is 48px)  */
```
fregante, line 177:
```suggestion
/* Increase size of profile picture in user hovercard by 1.5x (default is 48px) #7695 */
/* Test: https://github.com/refined-github/refined-github */
```

fregante (approved):
The overflow is fine, it seems to actually work correctly most of the times as long as the styles are loaded before the hovercard is added. Minimal overflow is actually already caused by `user-local-time`

fregante:
Thanks @Andrew6rant!

PR #7698, opened 2024-08-19
`warn-pr-from-master` - Restore feature

kovsu:
Closes: #7670 

## Test URLs

- Simple: https://github.com/refined-github/refined-github/compare/main...fregante:main
- Renamed fork and changed default branch: https://github.com/refined-github/sandbox/compare/default-a...bfred-it-org:github-sandbox:main?expand=1
- Non-standard default name: https://github.com/refined-github/refined-github/compare/sandbox/keep-branch...yakov116:upstream

## Screenshot

https://github.com/user-attachments/assets/c52fb5f3-b264-4781-bb8c-5ec310450ecb


source/features/warn-pr-from-master.tsx
```diff
@@ -1,41 +1,43 @@
...
+	anchor?.before(
 		<div className="flash flash-error my-3">
 			<strong>Note:</strong> Creating a PR from the default branch is an <a href="https://blog.jasonmeridth.com/posts/do-not-issue-pull-requests-from-your-master-branch/" target="_blank" rel="noopener noreferrer">anti-pattern</a>.
 		</div>,
 	);
 }
 
+async function isCrossRepoCompareFromMaster(): Promise<boolean> {
+	const {nameWithOwner, path, name} = getRepo()!;
+	const base = getRepo(nameWithOwner)!;
```
fregante, line 20: Why not just:

```suggestion
	const base = getRepo()!;
```

I'd still like to see a testable `parseCompareUrl` function as suggested in https://github.com/refined-github/refined-github/pull/7683#issuecomment-2288668004

It can live in `github-helpers/parse-compare-url.ts` and be tested in `github-helpers/parse-compare-url.test.ts`. 

Line 27 shows why this is needed: `.split` will fail on the URLs:

- `https://github.com/refined-github/refined-github/compare`
- `https://github.com/refined-github/refined-github/compare/main`
- `https://github.com/refined-github/refined-github/compare/main..fregante:refined-github:main` (notice the two dots, not three)

Also the function can short-circuit when the comparison is not cross-repo, like for:

- `https://github.com/refined-github/refined-github/compare/main...rollup`

```ts
type Comparison {
	head: RepositoryInfo;
	base: RepositoryInfo;
	isCrossRepo: boolean;
}
function parseCompareUrl(pathname: string): Comparison;
```

One more weird URL:

- `https://github.com/refined-github/refined-github/compare/main...refined-github:refined-github:rollup` (it mentions the repo/user, but it's not cross-repo)

kovsu:
@fregante please reveiw. I think this function works fine in my end.

kovsu:
here is some regex result for test urls.

```
compare

compare/main
main

compare/main...test/bun
main
...
test/bun

compare/sandbox/keep-branch...yakov116:upstream
sandbox/keep-branch
...
yakov116:upstream

compare/main...refined-github:refined-github:rollup
main
...
refined-github:refined-github:rollup

compare/main..fregante:refined-github:main
main
..
fregante:refined-github:main
```

fregante:
It looks like `getDefaultBranch` is not used:

- Line 60 _always_ overrides `baseBranch`
- `headBranch` is _always_ overridden in the `switch`

I pushed a commit to drop this, how does it look?

fregante (changes requested):
~~I did some minimal changes to accept a `pathname` parameter as I had initially suggested and added one test. It seems to be parsing the URL incorrectly.~~

_Ignore, it was bad copilot generated input :(_

fregante (approved):
Should be good to go now 🍻

source/github-helpers/parse-compare-url.test.ts
```diff
@@ -0,0 +1,45 @@
...
+const base = {
+	branch: 'main',
+	repo: {
+		nameWithOwner: 'john/recipes',
+	},
+};
+
+test('parseCompareUrl', () => {
+	expect(parseCompareUrl('/john/recipes/compare')).toBe(undefined);
+	expect(parseCompareUrl('/john/recipes/compare/main')).toBe(undefined);
```
fregante, line 14: Behavior change"

Without information of the other branch, there's no comparison, it's always empty. We need to "handle" these URLs by just returning early and returning undefined/false

source/github-helpers/get-default-branch.ts
```diff
@@ -27,11 +26,12 @@ async function fromDOM(): Promise<string | undefined> {
 	return extractCurrentBranchFromBranchPicker(element);
 }
 
-async function fromAPI(repository: RepositoryInfo): Promise<string> {
+async function fromAPI(repository: NameWithOwner): Promise<string> {
+	const [name, owner] = repository.split('/');
```
kovsu, line 30: owner is first. 😂

kovsu: image

fregante:
```
                 TOP
                NOTCH
```

PR #7702, opened 2024-08-19
`same-branch-author-commits` - Restore feature

Seregy:
At the moment the `same-branch-author-commits` doesn't work correctly as the current selector no longer matches the author link element

## Test URLs

https://github.com/refined-github/sandbox/commits/branch/with/slashes/

## Screenshot

same-branch-author-restore

### DIFF
--- a/source/features/same-branch-author-commits.tsx
+++ b/source/features/same-branch-author-commits.tsx
@@ -1,22 +1,21 @@
+import {$$} from 'select-dom';
 import * as pageDetect from 'github-url-detection';
 
 import features from '../feature-manager.js';
-import observe from '../helpers/selector-observer.js';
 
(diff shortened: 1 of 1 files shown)


fregante (commented):
Thanks for the PR!

source/features/same-branch-author-commits.tsx
```diff
@@ -1,19 +1,22 @@
...
+
+function init(signal: AbortSignal): void {
+	observe(authorLinkSelector, changePath, {signal});
 }
 
 void features.add(import.meta.url, {
 	include: [
 		pageDetect.isRepoCommitList,
 	],
-	awaitDomReady: true, // Small page
```
fregante, line 16: I'd prefer keeping this and `$$` though, let's just change the selector

fregante:
Дякую, Сергій!

Issue #7699, opened 2024-08-19
`show-names` is broken on github enterprise 3.11.13


SuperSandro2000:
### Description

I am on a private GitHub Enterprise version 3.11.13 and the show-names features is replacing the names with undefined.

I get the following trace in console:

```
refined-github.js:3397 Uncaught (in promise) Error: GraphQL:
Could not resolve to a User with the login of 'undefined'.
    at Object.v4uncached (refined-github.js:3397:60)
    at async show_names_updateLink (refined-github.js:6239:21)
```

![image](https://github.com/user-attachments/assets/828d2dec-ae78-4e4b-a21e-318d2faf11c5)

I couldn't find the graphql query for this feature in the network tab, only the defaultBranchRef and total count of releases appeared there.

### How to replicate the issue + URL

An equivalent URL on public github where the bug does not appear https://github.com/NuschtOS/nixos-modules/pull/76

### Extension version

24.8.13

### Browser(s) used

Google Chrome 127.0.6533.120

### Token ⚠️

- [X] I set a token in the options page. I understand that some features require a token to work.

fregante:
I'd accept a quick PR that detects `undefined` and throws an error instead so I can merge it in the next release this week.

fregante:
@SuperSandro2000 can you investigate further? How can this be happening? We have a couple of places where `undefined` would fail before DOM replacement happens.

1. We're exclusively selecting `a` elements, so `textContent` _cannot_ be `undefined` there, it has to be a string, even if empty.
	https://github.com/refined-github/refined-github/blob/6e167b9d881bd05db8cb92a5e140144cda7fd873/source/features/show-names.tsx#L80
3. Right before this replacement happens, we call a function that exclusively works on strings:
	https://github.com/refined-github/refined-github/blob/6e167b9d881bd05db8cb92a5e140144cda7fd873/source/features/show-names.tsx#L47


So it seems to me that this is somehow getting an `'undefined'` string. Given the timing, this is very likely a consequence of:

- https://github.com/refined-github/refined-github/pull/7666

SuperSandro2000:
Derp 🤦🏼 I totally forgot that I have a similar addon installed for our internal GitHub which provides a similar feature but translates our internal IDs to readable github names. When I disabled that addon the bug goes away. The feature in Refined GitHub works as expected but only shows the internal user IDs which is not that helpful 😅 

Big Sorry!

fregante:
Thanks for updating us 😅 🙏

PR #7705, opened 2024-08-20
show-names - Handle missing textContent

kovsu:
Fixes: #7699

```tsx
const names = await api.v4(
	[...usernames].map(user =>
		api.escapeKey(user) + `: user(login: "${user}") {name}`,
	).join(','),
);
```

```
refined-github.js:3397 Uncaught (in promise) Error: GraphQL:
Could not resolve to a User with the login of 'undefined'.
    at Object.v4uncached (refined-github.js:3397:60)
    at async show_names_updateLink (refined-github.js:6239:21)
```

The reason cause this error is `user` is `undefined`, it means `usernames` set contains `undefined` data.

### DIFF
--- a/source/features/show-names.tsx
+++ b/source/features/show-names.tsx
@@ -21,7 +21,14 @@ function dropExtraCopy(element: HTMLAnchorElement): void {
 // The selector observer calls this function several times, but we want to batch them into a single GraphQL API call
 async function updateLink(batchedUsernameElements: HTMLAnchorElement[]): Promise<void> {
 	const myUsername = getUsername();
-	batchedUsernameElements = batchedUsernameElements.filter(({textContent: name}) => name !== myUsername && name !== 'ghost');
+	batchedUsernameElements = batchedUsernameElements.filter((element) => {
+		const name = element.textContent;
+		if (!name) {
+			console.warn('Error', element)
+			throw new Error('Failed to get textContent in the element');
+		}
+		return name !== myUsername && name !== 'ghost'
+	});
 	const usernames = new Set(batchedUsernameElements.map(element => element.textContent));
 
 	if (usernames.size === 0) {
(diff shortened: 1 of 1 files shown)


source/features/show-names.tsx
```diff
@@ -21,7 +21,17 @@ function dropExtraCopy(element: HTMLAnchorElement): void {
 // The selector observer calls this function several times, but we want to batch them into a single GraphQL API call
 async function updateLink(batchedUsernameElements: HTMLAnchorElement[]): Promise<void> {
 	const myUsername = getUsername();
-	batchedUsernameElements = batchedUsernameElements.filter(({textContent: name}) => name !== myUsername && name !== 'ghost');
+	batchedUsernameElements = batchedUsernameElements.filter(element => {
+		const name = element.textContent;
+
+		// https://github.com/refined-github/refined-github/issues/7699
+		if (!name) {
+			console.warn('Error', element);
```
fregante, line 29:
```suggestion
		const name = element.textContent;
		if (!name) {
			console.warn('show-names has a bug: ttps://github.com/refined-github/refined-github/issues/7699', element);
```

fregante (approved):
Thanks for the PR!

This is good as a temporary fix, we just want to avoid breaking _all usernames_ on GitHub Enterprise.

Ideally we want to actually fix the underlying issue so that `undefined` does not appear. I'm actually surprised this is a bug since `textContent` should not be undefined on non-empty elements (empty elements being `img`, `input`, etc)

fregante:
@busches it'd be great to have a proper fix at some point. I'm gonna merge this PR this week as a temporary fix.

busches:
I'll take a look ideally tonight, I haven't had any issues, but they upgraded GHE this weekend and maybe now there are. 😂

fregante:
Never mind, the user said it was a conflict with their other extension

PR #7715, opened 2024-08-21
`extend-conversation-status-filters` - Improve reliability

kovsu:
Related with: https://github.com/refined-github/refined-github/issues/5871 

before:

https://github.com/user-attachments/assets/12c51cd9-aadd-4e13-bfa0-c7fd1e1a1e1a

after:

https://github.com/user-attachments/assets/83c43900-51d7-4f5a-af6d-2b781f4bf5a8

### DIFF
--- a/source/features/extend-conversation-status-filters.tsx
+++ b/source/features/extend-conversation-status-filters.tsx
@@ -4,41 +4,42 @@ import features from '../feature-manager.js';
 import SearchQuery from '../github-helpers/search-query.js';
 import observe from '../helpers/selector-observer.js';
 
-function getFilterState(): {isMerged: boolean; isUnmerged: boolean} {
-	const locationQuery = SearchQuery.from(location);
-
-	return {
-		isMerged: locationQuery.includes('is:merged'),
-		isUnmerged: locationQuery.includes('is:unmerged'),
-	};
-}
-
 function addMergeLink(lastLink: HTMLAnchorElement): void {
 	// It's shouldn't be added in issues list
 	if (!pageDetect.isPRList()) {
 		return;
 	}
 
-	const {isMerged, isUnmerged} = getFilterState();
+	const locationQuery = SearchQuery.from(location);
+	const isMerged = locationQuery.includes('is:merged');
+	const isUnmerged = locationQuery.includes('is:unmerged');
 
+	// The links in `.table-list-header-toggle` are either:
+	//   1 Open | 1 Closed
+	//   1 Total            // Apparently appears with is:merged/is:unmerged
 	if (isMerged) {
 		// It's a "Total" link for "is:merged"
 		lastLink.lastChild!.textContent = lastLink.lastChild!.textContent.replace('Total', 'Merged');
+		return;
 	} else if (isUnmerged) {
 		// It's a "Total" link for "is:unmerged"
 		lastLink.lastChild!.textContent = lastLink.lastChild!.textContent.replace('Total', 'Unmerged');
+		return;
 	} else {
 		// In this case, `lastLink` is expected to be a "Closed" link
 		const mergeLink = lastLink.cloneNode(true);
 		mergeLink.textContent = 'Merged';
 		mergeLink.classList.toggle('selected', isMerged);
-		// If link is selected, the filters are already removed
-		mergeLink.href = lastLink.classList.contains('selected') ? SearchQuery.from(mergeLink).append('is:merged').href : SearchQuery.from(mergeLink).replace('is:closed', 'is:merged').href;
+		mergeLink.href = SearchQuery.from(mergeLink).replace('is:closed', 'is:merged').href;
 		lastLink.after(' ', mergeLink);
 	}
 }
 
 function removeAllFilters(link: HTMLAnchorElement): void {
+	if (link === link.parentElement!.lastElementChild) {
+		addMergeLink(link)
+	}
+
 	if (link.classList.contains('selected')) {
 		link.href = SearchQuery
 			.from(link)
@@ -54,7 +55,6 @@ function removeAllFilters(link: HTMLAnchorElement): void {
 
 function init(signal: AbortSignal): void {
 	observe('.table-list-header-toggle.states a', removeAllFilters, {signal});
-	observe('.table-list-header-toggle.states a:last-child', addMergeLink, {signal});
 }
 
 void features.add(import.meta.url, {
--- a/source/features/extend-conversation-status-filters.tsx
+++ b/source/features/extend-conversation-status-filters.tsx
@@ -25,7 +25,7 @@ function addMergeLink(lastLink: HTMLAnchorElement): void {
 		// // It's a "Total" link for "is:merged"
 		lastLink.lastChild!.textContent = lastLink.lastChild!.textContent.replace('Total', 'Merged');
 	} else if (isUnmerged) {
-			// It's a "Total" link for "is:unmerged"
+		// It's a "Total" link for "is:unmerged"
 		lastLink.lastChild!.textContent = lastLink.lastChild!.textContent.replace('Total', 'Unmerged');
 	} else {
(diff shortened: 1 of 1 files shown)


fregante (commented):
We shouldn’t use observe just as a deduplicator. There are two selectors at lines 18 and 43 that should be two separate observe() calls instead. Also there’s an inline pageDetect on line 11, I’m not sure why it’s there

fregante:
Also thanks for linking to https://github.com/refined-github/refined-github/issues/5871, I forgot this was an issue. I’ve had a problem with buttons not working recently but I didn’t understand why. Maybe this was the reason.

fregante:
I'd rather not make the UI too cluttered, I've only been asked that question once before so I assume not many need it anyway.

kovsu:
PR list page:

https://github.com/user-attachments/assets/27301728-1dee-4826-8028-587bbd4b468a

Issue list page:

https://github.com/user-attachments/assets/3d455a24-bf2c-4ffe-8db8-6057ad7ae060

source/features/extend-conversation-status-filters.tsx
```diff
@@ -1,80 +1,67 @@
...
+		mergeLink.classList.toggle('selected', isMerged);
+		// If link is selected, the filters are already removed
+		mergeLink.href = lastLink.classList.contains('selected') ? SearchQuery.from(mergeLink).append('is:merged').href : SearchQuery.from(mergeLink).replace('is:closed', 'is:merged').href;
 		lastLink.after(' ', mergeLink);
 	}
 }
 
-function togglableFilters(): void {
-	for (const link of $$('.table-list-header-toggle.states a')) {
-		$('.octicon', link)?.remove();
```
kovsu, line 44: The reason of removing this line is it casues a bad experience when it's loading. And I think the original icon is not bad we don't need to remove.

fregante: The reason it's there is because we're changing the way the widget works: it's being turned into a radio/checkbox

source/features/extend-conversation-status-filters.tsx
```diff
@@ -1,80 +1,67 @@
...
-import CheckIcon from 'octicons-plain-react/Check';
-import elementReady from 'element-ready';
 import * as pageDetect from 'github-url-detection';
 
 import features from '../feature-manager.js';
 import SearchQuery from '../github-helpers/search-query.js';
+import observe from '../helpers/selector-observer.js';
 
-function addMergeLink(): void {
+function getFilterState(): {isMerged: boolean; isUnmerged: boolean} {
```
fregante, line 7: This can be inlined in `addMergeLink`, like:

```js
const locationQuery = SearchQuery.from(location);
const isMerged = locationQuery.includes('is:merged');
const isUnmerged = locationQuery.includes('is:unmerged');
```

source/features/extend-conversation-status-filters.tsx
```diff
@@ -1,80 +1,67 @@
...
+
+function addMergeLink(lastLink: HTMLAnchorElement): void {
+	// It's shouldn't be added in issues list
 	if (!pageDetect.isPRList()) {
 		return;
 	}
 
-	// The links in `.table-list-header-toggle` are either:
-	//   1 Open | 1 Closed
-	//   1 Total            // Apparently appears with is:merged/is:unmerged
```
fregante, line 17: Make sure to keep useful comments

source/features/extend-conversation-status-filters.tsx
```diff
@@ -1,80 +1,67 @@
...
-			// It's a "Total" link for "is:unmerged"
-			lastLink.lastChild!.textContent = lastLink.lastChild!.textContent.replace('Total', 'Unmerged');
-			continue;
-		}
+	const {isMerged, isUnmerged} = getFilterState();
 
+	if (isMerged) {
+		// It's a "Total" link for "is:merged"
+		lastLink.lastChild!.textContent = lastLink.lastChild!.textContent.replace('Total', 'Merged');
+	} else if (isUnmerged) {
```
fregante, line 27: Prefer early returns

```js
		return;
	}
	
	if (isUnmerged) {
```

source/features/extend-conversation-status-filters.tsx
```diff
@@ -1,80 +1,67 @@
...
+		// It's a "Total" link for "is:unmerged"
+		lastLink.lastChild!.textContent = lastLink.lastChild!.textContent.replace('Total', 'Unmerged');
+	} else {
```
fregante, line 30: Also here
```js
		return;
	}
```

source/features/extend-conversation-status-filters.tsx
```diff
@@ -1,80 +1,67 @@
...
 }
 
-async function init(): Promise<void | false> {
-	await elementReady('.table-list-filters');
-
-	addMergeLink();
-	togglableFilters();
+function init(signal: AbortSignal): void {
+	observe('.table-list-header-toggle.states a', removeAllFilters, {signal});
+	observe('.table-list-header-toggle.states a:last-child', addMergeLink, {signal});
```
fregante, line 57: This is redundant, we can inline this check in `removeAllFilters` like:

```ts
if (link === link.parentElement!.lastElementChild) {
	addMergeLink(link)
}
```

fregante (commented):
Thanks for the cleanup!

source/features/extend-conversation-status-filters.tsx
```diff
@@ -4,41 +4,42 @@ import features from '../feature-manager.js';
...
 		lastLink.lastChild!.textContent = lastLink.lastChild!.textContent.replace('Total', 'Unmerged');
+		return;
 	} else {
 		// In this case, `lastLink` is expected to be a "Closed" link
 		const mergeLink = lastLink.cloneNode(true);
 		mergeLink.textContent = 'Merged';
 		mergeLink.classList.toggle('selected', isMerged);
-		// If link is selected, the filters are already removed
-		mergeLink.href = lastLink.classList.contains('selected') ? SearchQuery.from(mergeLink).append('is:merged').href : SearchQuery.from(mergeLink).replace('is:closed', 'is:merged').href;
+		mergeLink.href = SearchQuery.from(mergeLink).replace('is:closed', 'is:merged').href;
```
kovsu, line 33: Update it because we add link before remove all filters.

source/features/extend-conversation-status-filters.tsx
```diff
@@ -1,80 +1,65 @@
...
-
-		if (lastLinkQuery.includes('is:unmerged')) {
-			// It's a "Total" link for "is:unmerged"
-			lastLink.lastChild!.textContent = lastLink.lastChild!.textContent.replace('Total', 'Unmerged');
-			continue;
-		}
-
+	if (isMerged) {
+		// It's a "Total" link for "is:merged"
+		lastLink.lastChild!.textContent = lastLink.lastChild!.textContent.replace('Total', 'Merged');
```
kovsu, line 22: It removed my `return` after running fix.

fregante: The suggestion included dropping `else`

- https://github.com/sindresorhus/eslint-plugin-unicorn/issues/16

source/features/extend-conversation-status-filters.tsx
```diff
@@ -1,80 +1,69 @@
...
-		lastLink.after(' ', mergeLink);
+function removeAllFilters(link: HTMLAnchorElement): void {
+	if (link === link.parentElement!.lastElementChild) {
+		addMergeLink(link);
 	}
-}
 
-function togglableFilters(): void {
-	for (const link of $$('.table-list-header-toggle.states a')) {
-		$('.octicon', link)?.remove();
```
fregante, line 44: Don't change how the icons work

fregante (approved):

fregante:
Thanks!

PR #7716, opened 2024-08-21
`comment-excess` -  Fix rare bug + improve style

kovsu:
## Test URLs

https://github.com/refined-github/sandbox/issues/3

## Screenshot

before:

image

after:

image


fregante:
I mean “match in style” not in regex. The two • in the second screenshot have different sizes. One of them is added by us and should be changed to match GitHub’s character instead. Or maybe it needs a class?

kovsu:
Oooo, get it.

kovsu:
Maybe just need change the character. `•` -> `·`

fregante (approved):

Issue #7000, opened 2023-10-25
`releases-tab` appears in dropdown even when there are no releases


fregante:
### Description

We're adding the tab regardless of releases:

 https://github.com/refined-github/refined-github/blob/906da8c189127627f23fd3f0a706c2369a79c9b0/source/features/releases-tab.tsx#L72

The code should be altered to query the API the same way we do it for the actual tab. That code can be copy-pasted into this function

<img width="643" alt="Screenshot" src="https://github.com/refined-github/refined-github/assets/1402241/56697979-c7b1-4203-8714-02b1143c8c71">

_Extracted from https://github.com/refined-github/refined-github/issues/6995_

### How to replicate the issue + URL

1. Visit https://github.com/refined-github/yolo
2. Open the dropdown

### Extension version

23.9.21

### Browser(s) used

Safari

fregante:
# Drop `awaitDomReady: true`

This isn't _required_, but generally we should prefer:

- using the selector observer instead of `$$()`
- having async detections via `elementReady`, if they're DOM-based

fregante:
# Replace `$` with `expectElement as $`

We've been using `$(selector)!` as a way to say we _expect_ the item to exist, and if it doesn't, the code should throw an error.

The issue is that the errors is usually not readable, like [x is undefined](https://togithub.com/refined-github/refined-github/issues/7670)

So we should use `expectElement` instead, and map it to `$`. [Example](https://github.com/refined-github/refined-github/commit/2662181319e5d51a6d68e9e59d7350a06edd8397)

fregante:
# Replace `attachElement` with `observer`

`attachElement` is our previous deduplicating method, we should now prefer the selector observer. This isn't always practical, but `attachElement` is to be considered _legacy_

fregante:
# Replace `deduplicate` with `observer`

`deduplicate: 'has-rgh'` is the old way to mark entire views as "processed". These are at risk of breaking (or are already broken) when GitHub changes the way pages are updated.

Using the selector observer is preferrable

fregante:
# Replace `isBlank` detections with `observer`

They were used to avoid `$()` throwing errors on empty pages, but `observer` does not throw errors in those cases, at least until https://github.com/refined-github/refined-github/issues/6096

fregante:
# Replace `element-ready` with `observer.waitForElement`

[...]

PR #7722, opened 2024-08-22
`action-used-by-link` - Prevent duplicates on back button

kovsu:
## Test URLs

https://github.com/marketplace/actions/title-replacer

## Screenshot

https://github.com/user-attachments/assets/cd0cb81a-49de-4024-b5b3-e3999d244dd3

### DIFF
--- a/source/features/action-used-by-link.tsx
+++ b/source/features/action-used-by-link.tsx
@@ -25,7 +25,7 @@ function getActionURL(): URL {
 function addUsageLink(side: HTMLElement): void {
 	const actionURL = getActionURL();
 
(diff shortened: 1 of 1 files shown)


fregante:
As mentioned in #7000:

> There's no need to manually go through and change all the features, but they're worth keeping in mind if we're already working on a specific feature

Does this fix a specific bug? `observe` is always heavier than a plain `$` so if the latter suffices we shouldn't change it.

kovsu:
You can see the bug in the left browser window. And it also relates to https://github.com/refined-github/refined-github/issues/5871

PR #7723, opened 2024-08-22
Update draft icon style

kovsu:
## Screenshot

before:

CleanShot 2024-08-22 at 15 32 49@2x

after:

CleanShot 2024-08-22 at 15 31 15@2x

CleanShot 2024-08-22 at 15 30 33@2x


fregante:
This is intentional, it's the whole point of `emphasize-draft-pr-label`

PR #7726, opened 2024-08-22
`tag-changes-link` - Improve reliability 

kovsu:
## Test URLs

https://github.com/refined-github/refined-github/tags
https://github.com/refined-github/refined-github/tags?after=0.5.0

## Screenshot

Sometimes it's all fail.
image

image

It's may add the `compareLink` to the end after reloading. But it's still fail for the last tag in this page.

image

after:

https://github.com/user-attachments/assets/97694952-2993-43db-b4a0-d18b3bb9443b

### DIFF
--- a/source/features/tag-changes-link.tsx
+++ b/source/features/tag-changes-link.tsx
@@ -36,6 +36,7 @@ async function getNextPage(): Promise<DocumentFragment> {
 
 function parseTags(element: HTMLElement): TagDetails {
 	// DO NOT change this to `pathname` because it's empty when element is from `getNextPage` function
+	// https://github.com/refined-github/refined-github/pull/7726#discussion_r1727135015
 	const tagUrl = expectElement(['a[href*="/tree/"]', 'a[href*="/tag/"]'], element).href;
 	const tag = /\/(?:releases\/tag|tree)\/(.*)/.exec(tagUrl)![1];
 
--- a/source/features/tag-changes-link.tsx
+++ b/source/features/tag-changes-link.tsx
@@ -33,8 +33,15 @@ async function getNextPage(): Promise<DocumentFragment> {
 }
 
 function parseTags(element: HTMLElement): TagDetails {
-	// Safari doesn't correctly parse links if they're loaded via AJAX #3899
-	const {pathname: tagUrl} = new URL($(['a[href*="/tree/"]', 'a[href*="/tag/"]'], element)!.href);
+	let tagUrl = $(['a[href*="/tree/"]', 'a[href*="/tag/"]'], element)!.href;
+
+	// If tagUrl from next page, it is just a pathname not a URL
+	if (tagUrl.startsWith('https://github.com')) {
+		// Safari doesn't correctly parse links if they're loaded via AJAX #3899
+		const {pathname} = new URL(tagUrl);
+		tagUrl = pathname;
+	}
+
 	const tag = /\/(?:releases\/tag|tree)\/(.*)/.exec(tagUrl)![1];
 
 	return {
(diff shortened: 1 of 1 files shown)


source/features/tag-changes-link.tsx
```diff
@@ -33,8 +33,15 @@ async function getNextPage(): Promise<DocumentFragment> {
 }
 
 function parseTags(element: HTMLElement): TagDetails {
-	// Safari doesn't correctly parse links if they're loaded via AJAX #3899
```
fregante, line 36: This is an old bug I can't repro anymore, so the comment and workaround can be dropped

source/features/tag-changes-link.tsx
```diff
@@ -33,8 +33,15 @@ async function getNextPage(): Promise<DocumentFragment> {
...
-	const {pathname: tagUrl} = new URL($(['a[href*="/tree/"]', 'a[href*="/tag/"]'], element)!.href);
+	let tagUrl = $(['a[href*="/tree/"]', 'a[href*="/tag/"]'], element)!.href;
+
+	// If tagUrl from next page, it is just a pathname not a URL
```
fregante, line 38: What does this mean? The `href` property is never relative, it's always the computed full URL; this is not the same as the `href` _attribute_, which _can_ be relative (but we're not using that anywhere)

fregante (approved):

source/features/tag-changes-link.tsx
fregante: Given that the Safari bug was fixed, I think you just want to use `const tagUrl = $().pathname`

kovsu:
I want to know what the difference between `$` and `expectElement`. Why them cause different result. 😥

fregante:
> Why them cause different result. 😥

They shouldn't have a different runtime behavior (if the element is found). `expectElement` just calls `$()` internally and then throws an error if it's `undefined`

source/features/tag-changes-link.tsx
```diff
@@ -33,8 +35,7 @@ async function getNextPage(): Promise<DocumentFragment> {
 }
 
 function parseTags(element: HTMLElement): TagDetails {
-	// Safari doesn't correctly parse links if they're loaded via AJAX #3899
-	const {pathname: tagUrl} = new URL($(['a[href*="/tree/"]', 'a[href*="/tag/"]'], element)!.href);
+	const tagUrl = expectElement(['a[href*="/tree/"]', 'a[href*="/tag/"]'], element).href;
```
fregante, line 38: It still needs to work just on `pathname`

```suggestion
	const {pathname: tagUrl} = expectElement(['a[href*="/tree/"]', 'a[href*="/tag/"]'], element);
```

fregante (approved):

source/features/tag-changes-link.tsx
kovsu: You can see the pathname is empty in the console.
image

kovsu: You can see the last dom is from next page. Because of some reasons its pathname is `''`.
image

```ts
```

kovsu: How about this?
```ts
const tagUrl = expectElement(['a[href*="/tree/"]', 'a[href*="/tag/"]'], element).href;
const tag = tagUrl.split('/').pop()!;
```

fregante: This seems to be the same issue as https://github.com/refined-github/refined-github/issues/3899, except that it's in Chrome too.

It seems that tag names are escaped in URLs: https://github.com/refined-github/sandbox/releases/tag/sandbox%2Freleases%2Fsandbox%2Freleases%2F1

So it's probably fine to just use `.href`, but this needs to be documented:

```
// DO NOT change this to `pathname` because blah blah
```

PR #7728, opened 2024-08-22
`suggest-commit-title-limit` - Fix length limit on new PRs

karlhorky:
Fixes https://github.com/refined-github/refined-github/pull/7689#issuecomment-2304519269

## Test URLs

https://github.com/refined-github/refined-github/compare/main...karlhorky:refined-github:test/bun?expand=true

## Screenshot

### DIFF
--- a/source/features/suggest-commit-title-limit.tsx
+++ b/source/features/suggest-commit-title-limit.tsx
@@ -4,7 +4,8 @@ import * as pageDetect from 'github-url-detection';
 
 import features from '../feature-manager.js';
 import onCommitTitleUpdate from '../github-events/on-commit-title-update.js';
-import {getConversationNumber, getNextConversationNumber} from '../github-helpers/index.js';
+import {getNextConversationNumber} from '../github-helpers/get-next-conversation-number.js';
+import {getConversationNumber} from '../github-helpers/index.js';
 import {formatPrCommitTitle} from './sync-pr-commit-title.js';
 
 // https://github.com/refined-github/refined-github/issues/2178#issuecomment-505940703
--- a/source/features/suggest-commit-title-limit.tsx
+++ b/source/features/suggest-commit-title-limit.tsx
@@ -4,7 +4,7 @@ import * as pageDetect from 'github-url-detection';
 
 import features from '../feature-manager.js';
 import onCommitTitleUpdate from '../github-events/on-commit-title-update.js';
-import {formatPrCommitTitle} from './sync-pr-commit-title.js';
+import {formatPrOrCompareCommitTitle} from './sync-pr-commit-title.js';
 
 // https://github.com/refined-github/refined-github/issues/2178#issuecomment-505940703
 const limit = 72;
@@ -13,9 +13,9 @@ function validateCommitTitle({delegateTarget: field}: DelegateEvent<Event, HTMLI
 	field.classList.toggle('rgh-title-over-limit', field.value.length > limit);
 }
 
-function validatePrTitle({delegateTarget: field}: DelegateEvent<Event, HTMLInputElement>): void {
+async function validatePrTitle({delegateTarget: field}: DelegateEvent<Event, HTMLInputElement>): Promise<void> {
 	// Include the PR number in the title length calculation because it will be added to the commit title
-	field.classList.toggle('rgh-title-over-limit', formatPrCommitTitle(field.value).length > limit);
+	field.classList.toggle('rgh-title-over-limit', (await formatPrOrCompareCommitTitle(field.value)).length > limit);
 }
 
 function init(signal: AbortSignal): void {
--- a/source/features/suggest-commit-title-limit.tsx
+++ b/source/features/suggest-commit-title-limit.tsx
@@ -4,7 +4,7 @@ import * as pageDetect from 'github-url-detection';
 
 import features from '../feature-manager.js';
 import onCommitTitleUpdate from '../github-events/on-commit-title-update.js';
-import {getNextConversationNumber} from '../github-helpers/get-next-conversation-number.js';
+import getNextConversationNumber from '../github-helpers/get-next-conversation-number.js';
 import {getConversationNumber} from '../github-helpers/index.js';
 import {formatPrCommitTitle} from './sync-pr-commit-title.js';
 
--- a/source/features/suggest-commit-title-limit.tsx
+++ b/source/features/suggest-commit-title-limit.tsx
@@ -4,7 +4,8 @@ import * as pageDetect from 'github-url-detection';
 
 import features from '../feature-manager.js';
 import onCommitTitleUpdate from '../github-events/on-commit-title-update.js';
-import {formatPrOrCompareCommitTitle} from './sync-pr-commit-title.js';
+import {getConversationNumber, getNextConversationNumber} from '../github-helpers/index.js';
+import {formatPrCommitTitle} from './sync-pr-commit-title.js';
 
 // https://github.com/refined-github/refined-github/issues/2178#issuecomment-505940703
 const limit = 72;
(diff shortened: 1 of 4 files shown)


source/features/sync-pr-commit-title.tsx
```diff
@@ -25,6 +25,10 @@ export function formatPrCommitTitle(title: string): string {
 	return `${title} (#${getConversationNumber()!})`;
 }
 
+export async function formatPrOrCompareCommitTitle(title: string): Promise<string> {
+	return `${title} (#${getConversationNumber() || await getNextConversationNumber()})`;
```
fregante, line 29: This function is duplicating `formatPrCommitTitle` without actually replacing it. You might as well just create a dedicated `formatCompareCommitTitle` function with only `getNextConversationNumber` and move it out of `sync-pr-commit-title.tsx`

source/features/suggest-commit-title-limit.tsx
```diff
@@ -13,9 +13,10 @@ function validateCommitTitle({delegateTarget: field}: DelegateEvent<Event, HTMLI
 	field.classList.toggle('rgh-title-over-limit', field.value.length > limit);
 }
 
-function validatePrTitle({delegateTarget: field}: DelegateEvent<Event, HTMLInputElement>): void {
+async function validatePrTitle({delegateTarget: field}: DelegateEvent<Event, HTMLInputElement>): Promise<void> {
 	// Include the PR number in the title length calculation because it will be added to the commit title
-	field.classList.toggle('rgh-title-over-limit', formatPrCommitTitle(field.value).length > limit);
+	const prTitle = await formatPrOrCompareCommitTitle(field.value);
```
fregante, line 18: by the way API calls are automatically cached (in RAM), so this should be instant after the first `input` event

source/features/sync-pr-commit-title.tsx
karlhorky: yeah, I opted to duplicate it to make the pattern `${title} (#...)` as visually close together in the code as possible, to avoid the pattern implementations getting out of sync

if you don't think it's a risk, then happy to move this function out, eg. into `source/features/suggest-commit-title-limit.tsx` ... at that point we might even just inline the expression in the `validatePrTitle()` function instead of keeping it in a separate function 🤔

karlhorky: an alternative would be that we could keep the single function and instead pass in the PR title function:

```ts
export function formatPrCommitTitle(
  title: string,
  getPrNumber = () => getConversationNumber()!
): string {
	return `${title} (#${getPrNumber()!})`;
}
```

then the call inside of `source/features/suggest-commit-title-limit.tsx` could "inject" this expression dependency:

```ts
async function validatePrTitle({delegateTarget: field}: DelegateEvent<Event, HTMLInputElement>): Promise<void> {
	// Include the PR number in the title length calculation because it will be added to the commit title
	const prTitle = await formatPrOrCompareCommitTitle(field.value, () => {
      return getConversationNumber() ?? await getNextConversationNumber()
    });
	field.classList.toggle('rgh-title-over-limit', prTitle.length > limit);
}
```

karlhorky: Originally I considered just making `formatPrCommitTitle()` async, but this is used in quite a few places which do not need to call `getNextConversationNumber()`, which would add unnecessary `async/await` all over the place

fregante: I'm ok with duplicating it, but it doesn't belong here anymore

karlhorky: what about alternative 3 of passing in the PR number as an expression? feels cleaner somehow...

fregante: Yeah that looks good too

karlhorky: done in 675ab9511425633e3f6ec92890bf2b7bbf7cea6c

fregante:
hm weird, ripgrep is normally available on GitHub Actions, did they drop it? 🤦‍♂️

fregante:
I opened an issue for `rg`, but I don't know what's up with .then, you might have to google a bit.

- https://github.com/refined-github/refined-github/issues/7729

source/github-helpers/index.ts
```diff
@@ -17,6 +18,11 @@ export function getConversationNumber(): number | undefined {
 	return (type === 'pull' || type === 'issues') && Number(prNumber) ? Number(prNumber) : undefined;
 }
 
+export async function getNextConversationNumber(): Promise<number> {
+	const issues = await api.v3('issues?per_page=1');
+	return issues[0].number + 1;
```
karlhorky, line 23:
```suggestion
	return (issues[0].number as number) + 1;
```

karlhorky:
What looks potentially more interesting are these error messages here from [the failing workflow run](https://github.com/refined-github/refined-github/actions/runs/10509515097/job/29115903313?pr=7728):

```
⎯⎯⎯⎯⎯⎯ Failed Suites 5 ⎯⎯⎯⎯⎯⎯⎯

 FAIL  test/shorten-link.test.ts [ test/shorten-link.test.ts ]
 FAIL  source/github-helpers/get-current-git-ref.test.ts [ source/github-helpers/get-current-git-ref.test.ts ]
 FAIL  source/github-helpers/github-file-url.test.ts [ source/github-helpers/github-file-url.test.ts ]
 FAIL  source/github-helpers/index.test.ts [ source/github-helpers/index.test.ts ]
 FAIL  source/github-helpers/prevent-link-loss.test.ts [ source/github-helpers/prevent-link-loss.test.ts ]
TypeError: Method Promise.prototype.then called on incompatible receiver [object Module]
⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/5]⎯
```

I could try to track down any `.then()` calls, but I was wondering if there's an easier / more understandable way to read this test output. Maybe a different test reporter or something...

karlhorky:
@fregante So the Vitest test failures start as soon as `api` is imported in `source/github-helpers/index.ts`.

Minimal reproduction:

`source/github-helpers/index.ts`

```diff
import {$, elementExists, expectElement} from 'select-dom';
import onetime from 'onetime';
import elementReady from 'element-ready';
import compareVersions from 'tiny-version-compare';
import {RequireAtLeastOne} from 'type-fest';
import * as pageDetect from 'github-url-detection';
import mem from 'memoize';

+import api from './api.js';
import {branchSelector} from './selectors.js';

+console.log(api);
```

karlhorky:
@fregante a possible fix is to switch from static `import` to dynamic `import()`, in case this is ok:

```diff
import {$, elementExists, expectElement} from 'select-dom';
import onetime from 'onetime';
import elementReady from 'element-ready';
import compareVersions from 'tiny-version-compare';
import {RequireAtLeastOne} from 'type-fest';
import * as pageDetect from 'github-url-detection';
import mem from 'memoize';

-import api from './api.js';
import {branchSelector} from './selectors.js';

// This never changes, so it can be cached here
export const getUsername = onetime(pageDetect.utils.getUsername);
export const {getRepositoryInfo: getRepo, getCleanPathname} = pageDetect.utils;

export function getConversationNumber(): number | undefined {
	const [, _owner, _repo, type, prNumber] = location.pathname.split('/');
	return (type === 'pull' || type === 'issues') && Number(prNumber) ? Number(prNumber) : undefined;
}

export async function getNextConversationNumber(): Promise<number> {
+	const api = await import('./api.js');
	const issues = await api.v3('issues?per_page=1');
	return issues[0].number;
}
```

fregante:
So yes the issue is about circular references. The solution would be to move this helper out of index.ts and into its own file. `import()` may cause additional bundles to be generated and we don’t want that.

karlhorky:
Done in 65c64807a18803d98ef81892e5d37b76d793cf23

fregante (approved):
Thank you!

karlhorky:
Thanks for the guidance, review and merge!

Issue #7724, opened 2024-08-22
`rgh-feature-descriptions` - Add link to CSS file


fregante:
Since https://github.com/refined-github/refined-github/pull/6645, when you visit a feature's .css file, you get a link to its .tsx counterpart, like on https://github.com/refined-github/refined-github/blob/main/source/features/ci-link.css

<img width="510" alt="Screenshot" src="https://github.com/user-attachments/assets/326108bb-6987-4f55-ba8f-be10a5df8e7d">

Now I'd like to have the opposite: when you're on a .tsx file with a .css file, point to the .css file. The complication here is that you need to parse the following source code to determine whether the feature loads a stylesheet. `observe.waitForElement()`/`element-ready` might be useful here.

## Test URL

https://github.com/refined-github/refined-github/blob/main/source/features/ci-link.tsx

PR #7730, opened 2024-08-23
`rgh-feature-description` - Add link to css file

kovsu:
Closes: #7724 

## Test URLs

- Regular feature: https://github.com/refined-github/refined-github/blob/main/source/features/sync-pr-commit-title.tsx
- CSS counterpart: https://github.com/refined-github/refined-github/blob/main/source/features/sync-pr-commit-title.css
- RGH feature: https://github.com/refined-github/refined-github/blob/main/source/features/rgh-feature-descriptions.css
- CSS-only feature: https://github.com/refined-github/refined-github/blob/main/source/features/center-reactions-popup.css

## Screenshot

CleanShot 2024-08-23 at 09 13 48@2x

image

### DIFF
--- a/source/features/rgh-feature-descriptions.tsx
+++ b/source/features/rgh-feature-descriptions.tsx
@@ -4,8 +4,6 @@ import AlertIcon from 'octicons-plain-react/Alert';
 import CopyIcon from 'octicons-plain-react/Copy';
 import InfoIcon from 'octicons-plain-react/Info';
 
-import {$} from 'select-dom';
-
 import features from '../feature-manager.js';
 import optionsStorage, {isFeatureDisabled} from '../options-storage.js';
 import {featuresMeta, getNewFeatureName} from '../feature-data.js';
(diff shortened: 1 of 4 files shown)


source/features/rgh-feature-descriptions.tsx
```diff
@@ -16,13 +18,14 @@ import {isFeaturePrivate} from '../helpers/feature-utils.js';
 
 function addDescription(infoBanner: HTMLElement, id: string, meta: FeatureMeta | undefined): void {
 	const isCss = location.pathname.endsWith('.css');
+	const isCssOnly = isCss ? !$(`li[id="source/features/${id}.tsx-item"]`) : false;
```
fregante, line 21: This is in the sidebar, right? I think we can’t expect the list to be loaded because sometimes it’s loaded later. To check this I think you can:

1. close the sidebar
2. change branch
4. visit `view-source:https://github.com/ etc etc` on the current page 

Load the source, don’t look at the Elements panel in the console. 

Anyway, this can be 
```suggestion
	const isCssOnly = isCss ? !elementExists(`li[id="source/features/${id}.tsx-item"]`) : false;
```

Also coming from `select-dom`

kovsu:
Maybe we only can scan the `source/features` folder and generate a json file to record this?

fregante:
We're already doing that as part of `getFeatureMeta`. We can definitely alter the json to include `css: true` when the file is present (and no property when it's absent)

fregante:
I think it would be as easy as:

```js
css: fs.existsSync(`source/features/${featureName}.css`);
```

on:


https://github.com/refined-github/refined-github/blob/09f447d05014f8578af35624dbfe610bba281702/build/readme-parser.ts#L42

kovsu:
Works fine! 😬

https://github.com/user-attachments/assets/c81f7979-7c03-4e2a-94ae-392017f4c7d6

fregante (approved):

source/globals.d.ts
```diff
@@ -11,6 +11,7 @@ interface FeatureMeta {
 	id: FeatureID;
 	description: string;
 	screenshot: string | null; // eslint-disable-line @typescript-eslint/ban-types -- We use `null` in the JSON file
+	css?: boolean;
```
fregante, line 14:
```suggestion
	css?: true;
```

fregante:
🙌

Issue #7681, opened 2024-08-13
`default-branch-button`: Branch name or commit hash is truncated


injust:
### Description

When the button for this feature is added, the branch name or commit hash is truncated, even though visually there is enough room for the full branch name/commit hash to be displayed.

![image](https://github.com/user-attachments/assets/75f8994c-8480-4352-8c9f-26606d454dc4)

### How to replicate the issue + URL

https://github.com/refined-github/refined-github/blob/74d86d37d7301719e0d9bd6536c52e32379e78e8/source/features/default-branch-button.tsx

### Extension version

24.8.13

### Browser(s) used

Chrome

### Token ⚠️

- [X] I set a token in the options page. I understand that some features require a token to work.

fregante:
Indeed this issue can even be seen in my own gif: https://github.com/refined-github/refined-github/pull/7403

Cause: https://github.com/refined-github/refined-github/pull/6249

https://github.com/refined-github/refined-github/blob/51b867bf47a95569a4d13ddaaaff161c288a5910/source/features/default-branch-button.css#L1-L4

PR #7731, opened 2024-08-23
`default-branch-button` - Stop shortening branch name

kovsu:
closes: #7681 

I think it doesn't need to set width. Github seem already handle this situation.

## Screenshot

截屏2024-08-23 10 21 46


fregante:
How does it look on https://github.com/refined-github/sandbox/tree/very-very-long-long-long-long-branch-name?

Make sure to look at it in various window sizes, with/without sidebar, and also scrolling down the page so that the sticky header appears

kovsu:
@fregante Everything works fine!

fregante (approved):

Issue #7737, opened 2024-08-24
`toggle-files-button` More and more toggle buttons


akku1139:
## Folder

1. Open GitHub repository
2. Open folder
![image](https://github.com/user-attachments/assets/e0a7407c-85ac-4294-99f5-f4b78b17fa00)
3. Move to another folder (other than repository root)
4. Two toggle buttons is displayed
![image](https://github.com/user-attachments/assets/169b9dbd-5d91-492e-808b-08ad9bc23262)
5. Move to another folder (other than repository root)
6. Three toggle buttons!
![image](https://github.com/user-attachments/assets/2d033b88-383f-41d1-8648-ce4661816bd6)
7. Open another folder and more buttons will be added

## File

1. Open GitHub repository
2. Open file
![image](https://github.com/user-attachments/assets/ceb6296c-4c59-4040-a739-93651bff5c22)
3. Move to another file
4. Two toggle buttons is displayed
![image](https://github.com/user-attachments/assets/8b31faf6-ef3e-4734-80de-4761b5401770)
5. Move to another file
6. Three toggle buttons!
![image](https://github.com/user-attachments/assets/9056cfc4-496f-4e43-bd80-960b537e4f34)
7. Open another file and more buttons will be added
![image](https://github.com/user-attachments/assets/4fc2d593-7e1c-40e5-998f-ccbd718029a9)

Also, the file toggle button is not visible when opening the file directly
even if you move to another file.
(e.g. paste URL. middle mouse button. open link in new tab.)

### How to replicate the issue + URL

https://github.com/zen-browser/desktop

(Probably all repositories)

### Extension version

24.8.20

### Browser(s) used

Firefox Developer Edition

### Token ⚠️

- [X] I set a token in the options page. I understand that some features require a token to work.

fregante:
Probably going to have to revert https://github.com/refined-github/refined-github/pull/7700

cc @kovsu

kovsu:
```ts
const existingButton = elementExists(`.${toggleButtonClass}`, commitsLink.closest('[class^="Box"]')!);
if (existingButton) {
	return;
}
```
How about this way? I run it in my browser, it works fine.

fregante:
I suppose so, but:

- no need to use the second argument of `elementExists`, there can only be one element on the whole page
- this doesn't solve the fact that it also appear on files

[...]

PR #7740, opened 2024-08-24
`toggle-files-button` - Avoid duplicates

kovsu:
closes: #7737 

## Test URLs

## Screenshot

https://github.com/user-attachments/assets/eb818ee8-12c3-4c57-b1c3-2370b9d1531f

### DIFF
--- a/source/features/toggle-files-button.tsx
+++ b/source/features/toggle-files-button.tsx
@@ -27,6 +27,9 @@ const noticeClass = 'rgh-files-hidden-notice';
 const noticeStyle = {paddingRight: '16px'};
 
 function addButton(commitsLink: HTMLElement): void {
+	// React view support
+	// https://github.com/refined-github/refined-github/issues/6554
+	// https://github.com/refined-github/refined-github/pull/7740#discussion_r1730211546
 	if (pageDetect.isSingleFile()) {
 		return;
 	}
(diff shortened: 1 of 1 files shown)


kovsu:
@fregante Please review!

source/features/toggle-files-button.tsx
```diff
@@ -27,6 +27,16 @@ const noticeClass = 'rgh-files-hidden-notice';
 const noticeStyle = {paddingRight: '16px'};
 
 function addButton(commitsLink: HTMLElement): void {
+	if (pageDetect.isSingleFile()) {
```
fregante, line 30: Why here and not include?

kovsu: Is better to put it in `exclude`?

fregante: No it cannot be moved because we still don't have a "page changed" listener, so the observer is only instantiated once (#6554):

1. open https://github.com/refined-github/refined-github/blob/main/test/setup-file.js
2. visit folder
3. no repotree features

Anyway, this feature makes no sense outside the homepage:

form

And arguably it makes little sense elsewhere.

Merging this PR as is for now but I might drop the feature

fregante: Opened
-  https://github.com/refined-github/refined-github/pull/7741

PR #7741, opened 2024-08-25
`toggle-files-button` - Drop feature

fregante:
This feature makes no sense.

form

The only reason to hide the file list _on a file list view_ is to see what's below it—and guess what, GitHub already provided a button to scroll to the readme (at least on the homepage)

Screenshot

There are two drawbacks to removing this feature:

- GitHub does not provide a link to the readme on non-root views, like: https://github.com/OctoLinker/OctoLinker/tree/main/packages
	- we could just add a feature, but it's so rarely useful
- people who never use the file list, now have to click/scroll to the readme on every load instead 
	- who _never_ uses the file list?


fregante:
The feature has also been one of the most troublesome: https://github.com/refined-github/refined-github/issues?q=sort%3Aupdated-desc+%22toggle-files-button%22+is%3Aclosed+in%3Atitle

x9136:
It was the most useful feature for me.

In the repositories where I need files, it's easier to get to them by searching, because most of the files I need are inside folders. In repositories where I don't need files, I immediately see the description without additional actions, like scrolling or clicking on the readme link.

Thanks for supporting this feature for a while.

PR #7762, opened 2024-08-28
Meta: Improve build speed and avoid redundant type issues

kovsu:


rollup.config.js
```diff
@@ -38,6 +38,7 @@ const rollup = {
 			return chunkInfo.name.replace('build/__snapshots__/', '') + '.js';
 		},
 	},
+	context: 'window',
```
kovsu, line 41: Avoid this warning. https://rollupjs.org/troubleshooting/#error-this-is-undefined
image

rollup.config.js
```diff
@@ -57,11 +58,8 @@ const rollup = {
 				{find: 'react', replacement: 'dom-chef'},
 			],
 		}),
-		typescript({
-			compilerOptions: {
-				noEmitOnError: !process.env.ROLLUP_WATCH,
-				module: 'Node16',
-			},
+		sucrase({
+			transforms: ['typescript', 'jsx'],
```
kovsu, line 62:
```js
exclude: ['node_modules/**'],
transforms: ['typescript', 'jsx'],
```

I prefer the config like this, but the `octicons-plain-react` library has exported `tsx` file, so I can't exclude `node_modules`. If you update the`octicons-plain-react`, I think maybe we can use `rollup-plugin-esbuild` to replace `@rollup/plugin-sucrase`.

fregante (approved):
Wow, this speeds up the build considerably.

rollup.config.js
fregante: We should add a link to https://github.com/sindresorhus/memoize/issues/102 so we will drop this after the change is done. 
```suggestion
   // TODO: blah blah
	context: 'window',
```

rollup.config.js
fregante: Given how fast it seems to be, I think it’s good as is. As soon as Vitest switches to Rolldown or [esbuild supports `preserveModules`](https://github.com/evanw/esbuild/issues/708) we might change bundler again anyway.

kovsu:
Oh, I forget to update that. 😂 Sooooorry

fregante:
Tested, the extension works

Issue #7765, opened 2024-08-28
`rgh-feature-descriptions` hides first line if feature is disabled 


rm-steele:
### Description

When I visit the source code for a disabled feature, the box notifying me it's disabled covers about half of the first line. Applies to both automatically and manually disabled features.
![checkboxes](https://github.com/user-attachments/assets/6954bc33-4675-4db2-b185-4cbe2b7d0bdb)
![example1](https://github.com/user-attachments/assets/0cedb068-3a03-455f-b929-85fa9878549b)
![example2](https://github.com/user-attachments/assets/0ecbbd88-7a9c-4f35-a69b-41cd94d16eed)
![example3](https://github.com/user-attachments/assets/15c882e5-f145-46ec-a40c-a7d06f7edd20)

### How to replicate the issue + URL

1. Disable a feature if none are disabled
2. In the extension settings, find a disabled feature and click the "source" link
3. Wait for the yellow "This feature is disabled" text box to load

As of writing, the easiest place to view the issue would be the [table input source code](https://github.com/refined-github/refined-github/blob/main/source/features/table-input.tsx), because it's forced off.

### Extension version

24.8.26.1

### Browser(s) used

Mozilla Firefox 129.0

### Token ⚠️

- [X] I set a token in the options page. I understand that some features require a token to work.

fregante:
Thanks! We had this same exact issue before, but I don’t remember which feature caused it. I think the yellow bar needs to change container/parent

fregante:
Duplicate of https://github.com/refined-github/refined-github/issues/6337

PR #7768, opened 2024-08-29
`rgh-feature-descriptions` - Fix overlap of "Feature disabled"

kovsu:
Closes: #7765 

## Test URL

https://github.com/refined-github/refined-github/blob/main/source/features/table-input.tsx

## Screenshot

before:

image

after:

image

### DIFF
--- a/source/features/rgh-feature-descriptions.tsx
+++ b/source/features/rgh-feature-descriptions.tsx
@@ -86,7 +86,8 @@ function addDescription(infoBanner: HTMLElement, id: string, meta: FeatureMeta |
 }
 
 async function getDisabledReason(id: string): Promise<JSX.Element | undefined> {
-	const classes = ['mb-3 d-inline-block width-full'];
+	// Block and width classes required to avoid margin collapse
+	const classes = ['mb-3', 'd-inline-block', 'width-full'];
 	// Skip dev check present in `getLocalHotfixes`, we want to see this even when developing
 	const hotfixes = await brokenFeatures.get() ?? [];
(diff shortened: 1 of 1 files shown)


kovsu:
It's interesting. Just swap two banners' order, it works fine. 🤣

fregante:
Don't forget to include test URLs in your PR as the template suggests ☺️

source/features/rgh-feature-descriptions.tsx
```diff
@@ -135,8 +135,8 @@ async function add(infoBanner: HTMLElement): Promise<void> {
 	// This ID exists whether the feature is documented or not
 	const id = meta?.id ?? filename;
 
-	addDescription(infoBanner, id, meta);
 	await addDisabledBanner(infoBanner, id);
+	addDescription(infoBanner, id, meta);
```
fregante, line 139: This change makes it dependent on the disabled banner code, we don't have to do that. Check out the previous solution in

- https://github.com/refined-github/refined-github/pull/6345

source/features/rgh-feature-descriptions.tsx
```diff
@@ -86,7 +86,7 @@ function addDescription(infoBanner: HTMLElement, id: string, meta: FeatureMeta |
 }
 
 async function getDisabledReason(id: string): Promise<JSX.Element | undefined> {
-	const classes = ['mb-3'];
+	const classes = ['mb-3 d-inline-block width-full'];
```
fregante, line 89:
```suggestion
	// Block and width classes required to avoid margin collapse
	const classes = ['mb-3', 'd-inline-block', 'width-full'];
```

Issue #7564, opened 2024-07-18
`repo-header-info` - Yellow star if starred


fregante:
It might get out of date quickly, but I think it might be possible to get this piece of information in the same API call.

### Before

<img width="322" alt="Screenshot 3" src="https://github.com/user-attachments/assets/42f8bf49-a469-4368-aba0-c9b433511456">

### After

<img width="322" alt="Screenshot" src="https://github.com/user-attachments/assets/8d5e3b65-189e-4a86-9508-73e3606b6b1e">

PR #7776, opened 2024-09-02
`repo-header-info` - Shows whether you starred the repo

kovsu:
closes: #7564

## Test URLs

## Screenshot

I upload another screenshot of the repo because `refined-github` has a cache in my browser.

image

here is where `viewerHasStarred` is from. https://docs.github.com/en/graphql/reference/interfaces#starrable


source/features/repo-header-info.tsx
```diff
@@ -53,7 +55,11 @@ async function add(repoLink: HTMLAnchorElement): Promise<void> {
 				title={`Repository starred by ${stargazerCount.toLocaleString('us')} people`}
 				className="d-flex flex-items-center flex-justify-center mr-1 gap-1 color-fg-muted"
 			>
-				<StarIcon className="ml-1" width={12} height={12} />
+				{
+					viewerHasStarred
+						? <StarFillIcon className="ml-1" color="var(--button-star-iconColor)" width={12} height={12} />
```
fregante, line 60: `color`? Shouldn’t it be `fill`?

fregante (approved):

source/features/repo-header-info.tsx
kovsu: image

fregante: Worth adding a comment. 

Also… something is up with our JSX, there are extra props. This is likely due to sucrase and it might need further config.

fregante (approved):

PR #7796, opened 2024-09-08
`conventional-commits` - New feature

### LINKED ISSUE #7795: Display conventional commits as colored labels before the commit message
## Description

Enhance the display of commit messages by rendering SemVer tags as colored badges, improving clarity and readability.

## Visual Example

Current commit view:

Current View

Proposed improvement:

Proposed view

## Detailed Explanation

`semantic-release` prefixes commit messages with conventional labels (e.g., `feat`, `fix`, `chore`), creating a structured naming convention. Currently, commit titles look like this:

Example Commit

GitHub does not visually distinguish SemVer prefixes from the rest of the commit message, making it harder to quickly identify commit types. These prefixes, however, are crucial for tools that automatically generate changelogs and versioning.

By transforming these prefixes into colored badges, we can instantly communicate the purpose of each commit:
- **Red** for `fix`
- **Green** for `feat`
- **Gray** for `chore`
- **Orange/Brown** for `refactor` (and similar types)

The scope (inside parentheses) could be displayed next to the commit type, as shown in the example.

## Customization Options

To increase flexibility, the following options could be added:

- **Badge Position**: Display the badge either *before* or *after* the commit title.
- **Color Customization**: Allow users to customize badge colors for each commit type.
- **Custom Types**: Let users define additional commit types and assign specific colors to them.

## Guideline Compliance:

- **Existing solutions**: No known userscripts or extensions provide this functionality after a thorough search.
- **Universal applicability**: SemVer is widely adopted in libraries, making this feature useful for most users.
- **No options required**: The SemVer specification dictates clear rules, so additional options are unnecessary. Customization (e.g., color changes) could be optional.
- **Minimal UI/code effort**: This feature only replaces the prefix with a colored badge, requiring minimal code changes.
- **No external server**: The badge rendering can happen statically, with no need for server interactions.
- **No increased HTTP traffic**: The solution is entirely client-side and does not introduce additional network requests.
- **Scope of enhancement**: This enhancement targets commit views, not GitHub Projects.

### Example URLs

https://github.com/revanced/revanced-patches/commits (In the list of commits)
https://github.com/ReVanced/revanced-patches/commit/09e3b3044824d0317c408637d157d93d0201272c (Where the commit is shown)
https://github.com/revanced/revanced-patches (Where the last commit is shown)

oSumAtrIX:
Closes #7795.

## Test URLs

https://github.com/ReVanced/revanced-patches/commits/main/

## Screenshot

https://github.com/user-attachments/assets/630f4be4-e1fb-4315-89a6-1cf8d87612f8

# Todo

This PR is still a draft and is awaiting many changes such as:

- [x] Display labels instantly (Help needed from maintainers)
- [x] Render everywhere where commits are shown `pageDetect.isRepoCommitList for some reason doesn't work` (Help needed from maintainers)
- [x] Render when clicking on "Next" or "Previous" in the commit list instead of refreshing (and similar actions) (Help needed from maintainers)
- [x] Implement missing default commit types
- ~~[ ] Add options to specify color, custom commit types , to hide the scope and display the label before or after the commit message~~


fregante:
Let’s flesh out the idea before starting to code. I’m still unsure that it should be part of Refined GitHub, especially at this length. Most features are very short or very useful. Currently this is neither

oSumAtrIX:
Thanks, I understand the concern. I have been verbose with the code and created many readable code blocks and functions to keep it simple. It is possible to half the code size probably at a similar length like other features

fregante:
I only now saw the video. I kind like it actually, better than the original mockup. Potentially this also voids what I said about the non-semrel `fix` keyword since its visual position is preserved on the commit title line. 

You can continue iterating


> * [ ]  Display labels instantly (Help needed from maintainers)

Look into the `observe` function and drop `awaitDomReady`

> * [ ]  Render everywhere where commits are shown `pageDetect.isRepoCommitList for some reason doesn't work` (Help needed from maintainers)

`isCommitList` already includes that. 

Now the next question is: should this only apply to commit lists? How about single commits? How about PRs (this exact page, not the commit list)? I’m afraid that colors and bubbles might be noisy or “ugly” there. 

> * [ ]  Render when clicking on "Next" or "Previous" in the commit list instead of refreshing (and similar actions) (Help needed from maintainers)

The `observe` function suggested above helps.

> * [ ]  Implement missing default commit types

If this is merged, we should only support the official keywords and format as defined by the official documentation. 

> * [ ]  Add options to specify color, custom commit types , to hide the scope and display the label before or after the commit message

No options allowed in Refined GitHub, see:

https://github.com/refined-github/refined-github/wiki/%22Can-you-add-this-feature%3F%22#3-it-doesnt-require-options

fregante:
I want to include “semantic release” in the feature name. How about this one?

source/features/render-semantic-types.tsx
```diff
@@ -0,0 +1,191 @@
...
+			colorCssRules: {
+				'--label-r': 14,
+				'--label-g': 138,
+				'--label-b': 22,
+				'--label-h': 123,
+				'--label-s': 81,
+				'--label-l': 29,
+			},
+		},
+	],
```
fregante, line 81: CSS goes in CSS files. This shortens this file considerably. Each class can be named like `.rgh-semantic-release-feat`

source/features/render-semantic-types.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+function isSemanticCommitTitleElement(commitTitle: string): boolean {
+	// This may not be fully accurate.
+	// Tests, if the title string
+	// starts with a word,
+	// followed by an anything in parentheses,
+	// followed by an optional exclamation mark,
+	// followed by a colon,
+	// followed by anything.
+	return /^\w*(?:\(.*\))?!?:.*/.test(commitTitle);
```
fregante, line 15: We need to extract this regex and add tests. You can create:

- source/helpers/semantic-release.ts
- source/helpers/semantic-release.test.ts

I’d want to use a single regex for parsing and extraction of the data to make sure that these two parts are never out of sync. You should use regex named groups for this.

source/features/render-semantic-types.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+function extractSemanticCommitTitleTypeAndScope(semanticCommitTitle: string): [string, string?] {
+	// Gets the first word from the title.
+	const type = /^(\w*)/.exec(semanticCommitTitle)!.pop()!;
+
+	// Gets the word in parentheses from the title.
+	const scope = /\((\w*)\)/.exec(semanticCommitTitle)?.pop();
+
+	return [type, scope];
+}
```
fregante, line 50: These two functions should be a single regex.exec call for detection and extraction, as suggested above.

source/features/render-semantic-types.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+function removeSemanticCommitTitleAndScopeFromTitleElement(semanticCommitTitleElement: HTMLElement): void {
+	const children = semanticCommitTitleElement.childNodes;
+
+	// Remove all children until the child with a colon is found.
+	// Then remove the colon and everything before it.
+	for (const child of children) {
+		if (child.textContent.includes(':')) {
+			child.textContent = child.textContent.split(':').slice(1).join(':');
```
fregante, line 59: Like said above, we should try to keep the logic in a single place, so ideally we’d still use the regex to remove the same exact part of the title. This however isn’t straightforward across inline elements, it might need to be done via Range API.

I think however this can also be easily solved via zip-text-nodes. We can review it later.

source/features/render-semantic-types.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+	return (
+		<span
+			style={{
+				marginTop: 0,
+				marginLeft: 0,
+				fontSize: 'var(--text-body-size-large)',
+				fontWeight: 500,
+				padding: '2px 0.5rem',
+				...colorCssRules,
```
fregante, line 151: All of this should be moved to the CSS too

source/features/render-semantic-types.tsx
```diff
@@ -0,0 +1,191 @@
...
+	);
+	const defaultSemanticCommitTitleElementsOnPage = semanticCommitTitleElementsOnPage.filter(semanticCommitTitleElement =>
+		isSemanticCommitTitleTypeDefault(semanticCommitTitleElement.textContent),
+	);
+
+	for (const semanticCommitTitleElement of defaultSemanticCommitTitleElementsOnPage) {
+		const [type, scope] = extractSemanticCommitTitleTypeAndScope(semanticCommitTitleElement.textContent);
+
+		removeSemanticCommitTitleAndScopeFromTitleElement(semanticCommitTitleElement);
+		prependLabelToSemanticCommitTitleElement(semanticCommitTitleElement, type, scope);
```
fregante, line 181: Same as before, detecting, extracting and replacing should be done in one piece. Work on the rest for now and I’ll try writing a `zip-text-node`-based remover

fregante (dismissed):

fregante:
By the way let’s merge this feature “as is” before attempting to add more features, like `skip-ci` detection etc.

source/features/render-semantic-types.tsx
fregante: We should include this link also, for documentation: https://github.com/angular/angular/blob/main/CONTRIBUTING.md#-commit-message-format

source/features/render-semantic-types.tsx
fregante: Here’s some code written with Claude:

``` JavaScript 
function removeText(element, searchRegex) {
  const modifiedText = element.textContent.replace(searchRegex, match => `<s>${match}</s>`);
  
  const temp = document.createElement('div');
  temp.innerHTML = modifiedText;

  zipTextNodes(element, temp);

  element.querySelector('s').remove();
}
```

I should do the job.

So the logic would probably be like:

``` JavaScript 
const match = regex.exec(commitTitle.textContent)

if (!match) return

const {type, scope} = match.groups;
prependLabel(commitTitle, type, scope)
removeText(commitTitle, match[0])
```

I think all these changes would make the feature really small actually

source/features/render-semantic-types.tsx
```diff
@@ -0,0 +1,191 @@
...
+				'--label-h': 256,
+				'--label-s': 81,
+				'--label-l': 50,
+			},
+		},
+	],
+]);
+
+function createLabelElement(type: string, scope?: string): JSX.Element {
+	const {label, colorCssRules} = defaultCommitTypesMapping.get(type)!;
```
fregante, line 141: I think you don’t actually need the mapping. Only `feat` is shortened and we could arguably leave it as is. Even “first letter uppercase” might not be necessary since we’re following the style of native GitHub labels—and they’re all lowercase by default. Let’s keep it simple for a first iteration

oSumAtrIX:
> If this is merged, we should only support the official keywords and format as defined by the official documentation.

Semantic release uses the angular commit convention by default, specifying a handful of commit types. These should be the defaults.

> No options allowed in Refined GitHub, see:

It says "required". No options are "required". "required" options mean for example an API key for a feature to function. The options I suggest are not required but good to have and probably should be added because while semantic release uses angular commit conventions by default, people conventionally use "chore" for example. semantic-release itself does it too but its not specified in the angular commit convention. Additionally, semantic release allows using custom commit types.

fregante:
> > No options allowed in Refined GitHub, see:
> 
> It says "required". No options are "required". "required" options mean for example an API key for a feature to function.

I wrote the document, it means no options period. You can check the feature list and you’ll find no per-feature option. This is non-debatable.

oSumAtrIX:
In that case, the default types should be extended to other common types that everyone uses. Remember that if you do not use them in your project, the feature will still work as expected. It will render the types you use, but for those that use the other common types, it will also render. Semantic release itself already uses "chore," which isn't in the angular conventional commits spec.

oSumAtrIX:
image

Here are the current colors

fregante:
Let’s focus on the functionality for now. Colors can actually be customized by the user via class name so they’re minor. I’d expect them to be colorless by default except for the standard bug/enhancement/documentation label colors that GitHub natively offers.

source/features/render-semantic-types.tsx
oSumAtrIX: > I’d want to use a single regex for parsing and extraction of the data to make sure that these two parts are never out of sync. You should use regex named groups for this.

Sometimes I only want to parse the scope, so I can not reuse the regex that captures the type+ optional scope.

fregante: Why you can’t? Regex can do it

oSumAtrIX: I'll try to resolve it.

fregante: Something like this, the scope is also optional:

IMG_5350

source/features/render-semantic-types.tsx
oSumAtrIX: This solution seems less performant. I prefer my current solution because it doesn't rely on libraries or add any temporary div to a DOM. zipTextNodes feels like a hack on it's own as well. I applied your suggestion anyways, but I'd suggest reverting it.

oSumAtrIX:
> By the way let’s merge this feature “as is” before attempting to add more features, like skip-ci detection etc.

This was not part of the issue, I had it in the screenshot accidentally 👍

oSumAtrIX:
> Let’s focus on the functionality for now. Colors can actually be customized by the user via class name so they’re minor. I’d expect them to be colorless by default except for the standard bug/enhancement/documentation label colors that GitHub natively offers.

I am already done with all reviews except one you wanted to take over (I don't know how to implement it anyways).
Here is what it would look like colorless:

image

And here with colors:

image

I think most people that use the extension will not change anything let alone the CSS. They will see the badges and be annoyed they are colored the same way, so I think a default color is better than nothing. Less people will be annoyed, and those who are can change it via the CSS.

source/features/render-semantic-types.tsx
oSumAtrIX: GitHub using lowercase by default for issue labels is a mistake done by GitHub. There is no incentive to use lower case. In my repos I have fixed this mistake:

image

Everywhere else, GitHub uses uppercase:

image
image
image

Arguably, this is better because in English, the first character is written big. The official primer design language also starts with a big letter in labels, furthermore strengthening the fact that GitHub made a mistake and we should use uppercase first letter:

image

I have added the mapping for scaling reasons. Yes, maybe there's currently only "feat" that would be mapped to "Feature" (and also "perf" to "Performance"), but we don't know how we will scale in the future, which is why I think the mapping makes sense here. It is already used for two names that need to be written out. 

Additionally, GitHub uses "full" words and not abbreviated ones:

image
image
image

I believe shortening terms often impairing readability, and given that only a couple of words are lengthened by just a couple of characters currently, not adding any meaningful or annoying length, I believe we should keep the current mapping:

image

source/features/render-semantic-release-commit-types.tsx
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	if (!isSemanticCommitTitleTypeDefault(commitTitleElement.textContent)) {
+		return;
+	}
+
+	const [type, scope] = getSemanticCommitAndScope(commitTitleElement.textContent)!;
+
+	removeSemanticCommitTypeAndScopeFromCommitTitleElement(commitTitleElement);
+	prependLabelToSemanticCommitTitleElement(commitTitleElement, type, scope);
```
fregante, line 86: This function can actually be inlined, it’s a single call even shorter than this function name. 

```suggestion
	commitTitleElement.prepend(createLabelElement(type, scope));
```

source/features/render-semantic-release-commit-types.tsx
```diff
@@ -0,0 +1,112 @@
...
+	renderLabelInCommitTitle(repoCommitTitleElement);
+}
+
+function init(signal: AbortSignal): void {
+	const renderLabelInCommitTitle = pageDetect.isRepoCommitList()
+		? renderLabelInRepoCommitTitle
+		: renderLabelInPRCommitTitle;
+
+	observe(commitTitleInLists, renderLabelInCommitTitle, {signal});
+}
```
fregante, line 105: These three functions should probably be reorganized into 2, something like:

``` JavaScript 
function initPrCommitList() {
  observe(selector + ' a', formatInPr)
}
function initRepoCommitList() {
  observe(selector + ' h4 > span', formatInRepo)
}
``` 

And then the features.add() can take multiple init objects with specific `include` each

source/features/tags-on-commits-list.tsx
```diff
@@ -139,7 +139,7 @@ async function init(): Promise<void | false> {
 
 void features.add(import.meta.url, {
 	include: [
-		pageDetect.isRepoCommitList,
+		pageDetect.isCommitList,
```
fregante: Unrelated change, the DOM for pagination is different and it’s unlikely to find tagged commits in PRs, so not worth it for the vast majority of cases. This feature is not optimized

source/helpers/render-semantic-release-commit-types.ts
```diff
@@ -0,0 +1,28 @@
+import zipTextNodes from 'zip-text-nodes';
+
+// Using https://github.com/angular/angular/blob/main/CONTRIBUTING.md?rgh-link-date=2024-09-09T05%3A51%3A31Z#-commit-message-format as the reference for the seamntic commit message format.
+
+export function getSemanticCommitAndScope(commitTitle: string): [string, string?] | undefined {
+	const match = /^(\w*)(?:\((.+)\))?!?:.*/.exec(commitTitle);
```
fregante, line 6: Did ChatGPT’s regex not work? I had specifically requested named groups as well. Not that .* is meaningless at the end because we’re not capturing that part and the regex matches the first part just as well.

source/helpers/render-semantic-release-commit-types.test.ts
```diff
@@ -0,0 +1,14 @@
...
+
+test('getSemanticCommitAndScope', () => {
+	assert.equal(getSemanticCommitAndScope('Commit message'), undefined);
+	assert.equal(getSemanticCommitAndScope('fix: Commit message'), ['fix']);
+	assert.equal(getSemanticCommitAndScope('feat: Commit message'), ['feat']);
+	assert.equal(getSemanticCommitAndScope('feat(scope): Commit message'), ['feat', 'scope']);
+	assert.equal(getSemanticCommitAndScope('feat(sco pe): Commit message'), ['feat', 'sco pe']);
+	assert.equal(getSemanticCommitAndScope('feat(): Commit message'), ['feat']);
+	assert.equal(getSemanticCommitAndScope('feat: Commit (message)'), ['feat']);
+	assert.equal(getSemanticCommitAndScope('fe at(scope): Commit message) '), [undefined, 'scope']);
```
fregante, line 13: I’d expect this not to match anything. Perhaps the ChatGPT regex already addresses this

source/helpers/render-semantic-release-commit-types.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export function strip(
+	element: HTMLElement,
+	from: string,
+): void {
+	const modified = element.textContent
+		.replace(from, match => `<s>${match}</s>`);
+
+	const temporaryDiv = document.createElement('div');
+	temporaryDiv.innerHTML = modified;
```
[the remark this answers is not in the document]
fregante: As for performance, I agree it’s not the most well-optimized code for this intent, but it’s certainly tested and fully safe. In fact this XSS doesn’t actually make it to the page because the zip-text-nodes package will throw an error.

After merge I’ll check if I can move this strip function into zip-text-node to use its internal utilities directly, without the zipping part (but it’s unlikely, it would be more complex)

source/features/render-semantic-release-commit-types.tsx
```diff
@@ -0,0 +1,112 @@
...
+	['ci', 'CI'],
+	['perf', 'Performance'],
+]);
+
+function removeSemanticCommitTypeAndScopeFromCommitTitleElement(semanticCommitTitleElement: HTMLElement): void {
+	const [type, scope] = getSemanticCommitAndScope(semanticCommitTitleElement.textContent)!;
+
+	const semanticCommitTypeAndScope = scope ? `${type}(${scope}):` : `${type}:`;
+
+	strip(semanticCommitTitleElement, semanticCommitTypeAndScope);
```
fregante, line 40: It’s safer to strip the exact match rather than having to reconstruct the string. So probably `getSemanticCommitAndScope` should return the full matched commit prefix. 

Also note that this is the 4th call of getSemanticCommitAndScope in this file. `removeSemanticCommitTypeAndScopeFromCommitTitleElement` should also be inlined because it’s just `strip(commitTitleElement, prefix)`

source/refined-github.ts
```diff
@@ -104,6 +104,7 @@ import './features/suggest-commit-title-limit.js';
 import './features/highest-rated-comment.js';
 import './features/clean-conversation-filters.js';
 import './features/tags-on-commits-list.js';
+import './features/render-semantic-release-commit-types.js';
```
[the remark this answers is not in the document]
oSumAtrIX: I was thinking about `render-semantic-release-commits` for that reason in the first place, which is 4 words. The feature does render commits based on a syntax the same way HTML is rendered from a syntax. It uses the syntax semantic-release describes so the name  `render-semantic-release-commits` would be fitting. Let me know what you think.

source/helpers/render-semantic-release-commit-types.ts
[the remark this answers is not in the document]
oSumAtrIX: The alternative solution is the one I had earlier, where I was iterating the nodes and contacting the strings until I saw ":" and stripped it away. No XSS & dependencies and easier to understand than an abstract concept like zipping nodes. I'll leave the decision to you.

source/features/render-semantic-release-commit-types.css
```diff
@@ -0,0 +1,85 @@
...
+}
+
+.rgh-render-semantic-release-commit-types-chore-label-colors {
+    --label-r: 100;
+    --label-g: 100;
+    --label-b: 100;
+	--label-h: 0;
+	--label-s: 0;
+	--label-l: 255;
+}
```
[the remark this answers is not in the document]
oSumAtrIX: Was this documented? I am not too familiar with the NPM ecosystem so I was not aware of this listing command.

source/features/tags-on-commits-list.tsx
oSumAtrIX: I am not sure what you mean by that. Can you elaborate the change? This feature handles rendering in repo and PR commits.

source/helpers/render-semantic-release-commit-types.ts
oSumAtrIX: Part of the regex is to detect if it is a valid semantic commit. "feat:" is not a valid semantic release commit and should not be matched by the regex. There may be other quirks to a valid semantic commit that I do not know. For now this regex should pass the tests.

oSumAtrIX:
On a related note, I have remembered all this is based on https://www.conventionalcommits.org/en/v1.0.0/. The feature should be renamed to "render-conventional-commits-labels" for example

source/helpers/render-semantic-release-commit-types.ts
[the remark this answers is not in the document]
oSumAtrIX: https://gist.github.com/marcojahn/482410b728c31b221b70ea6d2c433f0c see this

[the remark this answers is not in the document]
oSumAtrIX: Also it works with the regex I have locally:

image

Consider the link above though.

oSumAtrIX:
@134130 Hi, thanks for your concerns. I have mentioned this in another comment:

> The prefix serves a functional purpose but is not meant to be shown directly to the user. conventional commits and semantic-release are widely adopted conventions for versioning and release automation across many languages and ecosystems:
> image
> While some might opt against using conventional commits, it’s worth noting that it is near-universally adopted (you’ve likely encountered it in the form of versions like "v1.2.3" without realizing it). 

https://github.com/refined-github/refined-github/issues/7795#issuecomment-2336805181

source/helpers/render-semantic-release-commit-types.ts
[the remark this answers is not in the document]
fregante: Question (I’m on mobile): do commit titles with emojis have extra elements in them? GitHub usually uses `<g-emoji>` to render them. If not, we might be able to simplify this a lot.

source/helpers/render-semantic-release-commit-types.test.ts
fregante: Run `npm run` to see the available commands. You can run `npm run test` to run all the unit tests/lint/formatting checkers at once. If you want just the unit tests specifically, it’s better you run `npx vitest` because that enables the watcher and the tests are rerun automatically.

source/helpers/render-semantic-release-commit-types.ts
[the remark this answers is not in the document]
oSumAtrIX: Seems to be just an emoji image

The regex \w does not match an emoji though: 

image

134130:
@oSumAtrIX Thank you for your reply. I've read your comment, but I just commented again.

In open-source, many projects are library, so they use semantic version. But products are may more profitable to use another version system, like [CalVer](https://calver.org/), The Ubuntu, and The JetBrains is using.

Of course, the SemVer is a really popular versioning strategy, so there is the middle way, like [HeadVer](https://github.com/line/headver).

And, see the kubernetes and typescript projects. They contain both conventional-commit and not been. Can we say the feature looks cool on that page also?

oSumAtrIX:
https://github.com/refined-github/refined-github/actions/runs/10795725898/job/29943063240?pr=7796#step:5:23 what is a test URL?

134130:
fregante is a nice maintainer, and I will follow his choice. (I don't have the power to decide)

oSumAtrIX:
I am unsure how to generate a .png link on GitHub. I have added an image link but it fails a test.

build/__snapshots__/features-meta.json
```diff
@@ -748,6 +748,12 @@
 		"description": "Retries downloading images that failed downloading due to GitHub limited proxying.",
 		"screenshot": "https://user-images.githubusercontent.com/14858959/64068746-21991100-cc45-11e9-844e-827f5ac9b51e.png"
 	},
+	{
+		"id": "render-conventional-commit-types",
+		"description": "Renders conventional commit types as labels prefixing the commit message.",
+		"css": true,
+		"screenshot": "https://github.com/user-attachments/assets/2a5ae00e-f440-49ef-b810-aa9a5a2ac269"
```
fregante, line 755: Where did you upload this? The URL looks fine but maybe we need to update the feature checker. 

What I do is drop the files in the PR comments and take the URL as is. If that’s what you did, we’ll have to update the feature checker to also accept this format.

build/__snapshots__/features-meta.json
```diff
@@ -748,6 +748,12 @@
 		"description": "Retries downloading images that failed downloading due to GitHub limited proxying.",
 		"screenshot": "https://user-images.githubusercontent.com/14858959/64068746-21991100-cc45-11e9-844e-827f5ac9b51e.png"
 	},
+	{
+		"id": "render-conventional-commit-types",
```
fregante, line 752: The closest feature we have is called `parse-backticks` so we can name this `parse-conventional-commits` or honestly I just want `conventional-commits`. I don’t expect additional related features due to what @134130 said so we can keep it simple with two words.

readme.md
```diff
@@ -288,6 +288,7 @@ Thanks for contributing! 🦋🙌
 - [](# "same-branch-author-commits") [Preserves current branch and path when viewing all commits by an author.](https://user-images.githubusercontent.com/44045911/148764372-ee443213-e61a-4227-9219-0ee54ed832e8.png)
 - [](# "easy-toggle-commit-messages") [Enables toggling commit messages by clicking on the commit box.](https://user-images.githubusercontent.com/1402241/152121837-ca13bf8a-9b7f-4517-8e8d-b58bb135523b.gif)
 - [](# "link-to-compare-diff") [Linkifies the "X files changed" text on compare pages to allow jumping to the diff.](https://user-images.githubusercontent.com/46634000/157072587-0335357a-18c7-44c4-ae6e-237080fb36b4.png)
+- [](# "render-conventional-commit-types") [Renders conventional commit types as labels prefixing the commit message.](https://user-images.githubusercontent.com/assets/2a5ae00e-f440-49ef-b810-aa9a5a2ac269.png)
```
fregante, line 291: cc @yakov116 if you have wording preferences.

source/features/render-conventional-commit-types.css
```diff
@@ -0,0 +1,85 @@
...
+	--label-g: 25;
+	--label-b: 231;
+	--label-h: 256;
+	--label-s: 81;
+	--label-l: 50;
+}
+
+.rgh-commit-type-label {
+	font-size: var(--body-font-size, 14px) !important;
+	padding: 2px 0.5rem !important;
```
fregante, line 84: What is this for, btw? It might be worth leaving a comment explaining why we don’t just use the default padding.

source/features/render-conventional-commit-types.tsx
```diff
@@ -0,0 +1,82 @@
...
+	removeCommitAndScope(commitTitleElement, match!);
+	commitTitleElement.prepend(createLabelElement(type, scope));
+}
+
+function renderLabelInRepoCommitTitle(commitTitleElement: HTMLElement): void {
+	const repoCommitTitleElement = $('h4 > span', commitTitleElement)!;
+	renderLabelInCommitTitle(repoCommitTitleElement);
+}
+
+function renderLabelInPRCommitTitle(commitTitleElement: HTMLElement): void {
```
fregante, line 50: These two functions should be dropped. The observed selector already points to the final element. 

Pass `renderLabelInCommitTitle` directly to the observer

source/features/render-conventional-commit-types.tsx
```diff
@@ -0,0 +1,82 @@
...
+	const prCommitTitleElement = $('a', commitTitleElement)!;
+	renderLabelInCommitTitle(prCommitTitleElement);
+}
+
+function initRepoCommitList(signal: AbortSignal): void {
+	observe(commitTitleInLists + ' h4 > span', renderLabelInPRCommitTitle, {signal});
```
fregante, line 56: The safe way to merge selectors is `:is(${selector}) h4 > span` as it will work even if `selector` has commas. 

One think I’d like you to check is whether both of these selectors point to a single element or if there are multiple spans/links (which I think is the case, especially if backticks and issue references appear in the commit title)

If that’s the case, the selector probably also needs `:first-child` somewhere

source/helpers/render-conventional-commit-types.ts
```diff
@@ -0,0 +1,31 @@
+import zipTextNodes from 'zip-text-nodes';
+
+// Using https://www.conventionalcommits.org/ as a reference.
+
+export function getConventionalCommitAndScopeMatch(commitTitle: string): RegExpExecArray | undefined {
```
fregante, line 5:
```suggestion
export function parseConventionalCommit(commitTitle: string): RegExpExecArray | undefined {
```

source/helpers/render-conventional-commit-types.ts
```diff
@@ -0,0 +1,31 @@
...
+	const match = /^(?<type>\w+)(?:\((?<scope>.+)\))?: .+$/.exec(commitTitle);
+
+	if (!match) {
+		return;
+	}
+
+	return match;
+}
+
+export function removeCommitAndScope(
```
fregante, line 15: This doesn’t remove the commit

```suggestion
export function cleanConventionalCommitElement(
```

source/helpers/render-conventional-commit-types.ts
```diff
@@ -0,0 +1,31 @@
+import zipTextNodes from 'zip-text-nodes';
+
+// Using https://www.conventionalcommits.org/ as a reference.
+
+export function getConventionalCommitAndScopeMatch(commitTitle: string): RegExpExecArray | undefined {
+	const match = /^(?<type>\w+)(?:\((?<scope>.+)\))?: .+$/.exec(commitTitle);
```
fregante, line 6:
```suggestion
	return /^(?<type>\w+)(?:\((?<scope>.+)\))?: .+$/.exec(commitTitle);
```

If it complains about `null` anywhere you can append `?? undefined` to this line.

source/helpers/render-conventional-commit-types.ts
```diff
@@ -0,0 +1,31 @@
...
+
+	const semanticCommitTypeAndScope = scope ? `${type}(${scope}):` : `${type}:`;
+
+	const modified = element.textContent.replace(semanticCommitTypeAndScope, match => `<s>${match}</s>`);
+
+	const temporaryDiv = document.createElement('div');
+	temporaryDiv.innerHTML = modified;
+
+	zipTextNodes(element, temporaryDiv);
+
```
fregante, line 29: We can simplify this function, but it would drop support for `feat(#123): title`, which is totally fine for me. 

```js 
assertNodeContent(node, theRegexAbove)
node.textContent = node.textContent.replace(theRegexAbove)
```

Notes:

- you will likely need to find the exact text node for this to work safely, probably `node = element.firstChild`
- you can find the assertion function in our helpers 
- the regex from the other function should be moved to the top scope and not copy-pasted

fregante (commented):
I think this will be the last pass 😁

build/__snapshots__/features-meta.json
oSumAtrIX: I dropped it in the comment text box. Also tried in the code editor.

build/__snapshots__/features-meta.json
oSumAtrIX: I guess `conventional-commits` would work, yeah

source/features/render-conventional-commit-types.css
oSumAtrIX: GitHub messed up with the CSS in labels. The text is not centered and it doesn't have proper padding. I try to fix it here. You can remove the rules and see the result. (I'll upload a screenshot in a bit)

source/helpers/render-conventional-commit-types.ts
oSumAtrIX: Sorry it should've said type and scope, not commit and scope

source/helpers/render-conventional-commit-types.ts
oSumAtrIX: Sorry but i don't quite understand what you mean with this review.

build/__snapshots__/features-meta.json
fregante: I think it actually works now. CI is only failing because the snapshots are not up to date.

source/features/render-conventional-commit-types.css
fregante: If you’re trying to align it via baseline and it should use vertical-align:baseline or align-self:baseline depending on the layout model. Hardcoded values might work for some fonts but not others. If you can’t figure it out it may be ok for now, I’ll give it a try when I get my computer back.

source/features/render-conventional-commit-types.tsx
fregante: My “multiple” I mean *per commit”

I know GitHub has some awkward splitting of the links so there might be multiple links and multiple spans. You can see this on any commit that has backticks or PR references (like most merged PR commits on this repo)

source/helpers/render-conventional-commit-types.ts
fregante: Yep, that’s why I suggested this name (with either `clean` or `strip`)

source/helpers/render-conventional-commit-types.ts
fregante: Those two lines are the entirety of the function, plus the likely `node = element.firstChild` line. Forget the zip dependency 🙌 👍

oSumAtrIX: Can you elaborate on what you mean by # in the scope?

source/features/render-conventional-commit-types.css
oSumAtrIX: Sorry, I missunderstood the question. The padding was increased because the font size was increased too:

image

In contrast to the default padding:

image

source/features/render-conventional-commit-types.tsx
oSumAtrIX: > One thing I’d like you to check is whether both of these selectors point to a single element or if there are multiple spans/links (which I think is the case, especially if backticks and issue references appear in the commit title)

There is only one element matching the selector.

source/helpers/render-conventional-commit-types.ts
oSumAtrIX: > Also, I mentioned this before, do not include .+$ in the regex because it’s completely unnecessary.

It was included to match only when a conventional commit is present.

`a:` is not a conventional commit. Let me know if you still want the change.

source/helpers/render-conventional-commit-types.ts
oSumAtrIX: This solution seems to break here:

image

(Likely an issue with `node = element.firstChild`: 
image)

oSumAtrIX:
I can not pass a test, but I am unsure why (It seems like a silly skill issue on my end). Other than that ready for review one more time.

readme.md
oSumAtrIX: I've updated the text to `Display conventional commits as colored labels before the commit message` simplifying the words "Render" and "prefixing".

fregante:
Did you run `npm run fix`

readme.md
```diff
@@ -288,6 +288,7 @@ Thanks for contributing! 🦋🙌
 - [](# "same-branch-author-commits") [Preserves current branch and path when viewing all commits by an author.](https://user-images.githubusercontent.com/44045911/148764372-ee443213-e61a-4227-9219-0ee54ed832e8.png)
 - [](# "easy-toggle-commit-messages") [Enables toggling commit messages by clicking on the commit box.](https://user-images.githubusercontent.com/1402241/152121837-ca13bf8a-9b7f-4517-8e8d-b58bb135523b.gif)
 - [](# "link-to-compare-diff") [Linkifies the "X files changed" text on compare pages to allow jumping to the diff.](https://user-images.githubusercontent.com/46634000/157072587-0335357a-18c7-44c4-ae6e-237080fb36b4.png)
+- [](# "conventional-commits") [Displays conventional commit types as labels before the commit message.](https://user-images.githubusercontent.com/assets/2a5ae00e-f440-49ef-b810-aa9a5a2ac269.png)
```
fregante, line 291: I think you updated the snapshot but left this broken link in here. Make sure to use the right link here. 

Also our screenshot are always with the light theme, can you take another one?

source/helpers/render-conventional-commit-types.ts
fregante: Try creating that # example commit. # is used for issue references, so GitHub will create this DOM:

``` html
<a href=“commit”>
  feat(
</a>
<a href=“issue or PR”>
  #123 
</a>
<a href=“commit”>
  ): title
</a>
``` 

Which is perfectly supported via zip-text-nodes but not with your first solution or this last solution. 

This is likely not a common scenario so that’s why I’m ok with it not being supported.

source/features/conventional-commits.tsx
```diff
@@ -0,0 +1,71 @@
...
+function renderLabelInCommitTitle(commitTitleElement: HTMLElement): void {
+	const match = parseConventionalCommit(commitTitleElement.textContent);
+
+	const {type, scope} = match?.groups ?? {};
+	if (!type || !types.has(type)) {
+		return;
+	}
+
+	removeCommitAndScope(commitTitleElement, match!);
+	commitTitleElement.prepend(createLabelElement(type, scope));
```
fregante, line 41: We should flip these two lines so that if the label creation fails, the commit won’t appear *trimmed*

source/helpers/conventional-commits.test.ts
```diff
@@ -0,0 +1,21 @@
...
+
+import {parseConventionalCommit} from './conventional-commits.js';
+
+function assertRegexDeepStrictEqual(input: string, expected: [string, string | undefined]) {
+	const result = parseConventionalCommit(input);
+	assert.deepStrictEqual(result, [input, ...expected]);
+}
+
+test('parseConventionalCommit', () => {
+	assertRegexDeepStrictEqual('fix: Commit message', ['fix', undefined]);
```
fregante, line 11:
```suggestion
	assertRegexDeepStrictEqual('fix: Commit message', ['fix: ', 'fix', undefined]);
```

this will tix the assertion. 

To answer a previous concern you had: it’s true that this will match `a: ` but then we have the whitelist of types, which `a` will fail. I suppose we can exclude commits that are titled exactly `fix: ` but I reckon that’s an extremely rare occurrence. 

At some point later I’ll move the `types` map to the helpers and have this parser return exactly `{match: 'chore(release): ', label: 'Chore: release'}` but that’s not necessary for now. It’s best to wrap up this PR 😃

readme.md
oSumAtrIX: This is what it looks in light mode: image

oSumAtrIX:
> Did you run `npm run fix`

It removes the screenshot in the snapshot and also throws an error because the image link is not supported by the check: https://github.com/user-attachments/assets/739264d7-b4b2-425d-91bb-d91c42c1ab21. I do not know where to upload the image to on GitHub to pass the check (it needs a .png suffix).

source/helpers/conventional-commits.test.ts
fregante: The argument is returned, so it has to be included in the tests and not ignored. The point of tests is to validate the entire result, not part of it

fregante:
`npm run fix` is not optional, do not edit the snapshot files manually. 

Set that URL in the readme and then I’ll look into supporting that format in the tests.

source/features/conventional-commits.css
```diff
@@ -0,0 +1,85 @@
...
+	--label-l: 50;
+}
+
+.rgh-commit-type-label-refactor-colors {
+	--label-r: 165;
+	--label-g: 138;
+	--label-b: 82;
+	--label-h: 40;
+	--label-s: 33;
+	--label-l: 48;
```
fregante, line 52: Can you pick a better color for this? Black on mustard isn’t very good

source/helpers/conventional-commits.test.ts
```diff
@@ -0,0 +1,21 @@
+import {test, assert} from 'vitest';
+
+import {parseConventionalCommit} from './conventional-commits.js';
+
+function assertRegexDeepStrictEqual(input: string, expected: [string, string | undefined]) {
+	const result = parseConventionalCommit(input);
+	assert.deepStrictEqual(result, [input, ...expected]);
```
fregante: When you adjust the expected arrays, this should also be changed:

```suggestion
	assert.deepStrictEqual(result, expected);
```

fregante (approved):
I think we’re finally done 🤩

source/features/conventional-commits.css
oSumAtrIX: I went with placeholder colors. Someone dedicated may know better and more standard color combinations for GitHub.

oSumAtrIX: image

how about this color

fregante: The previous color looked good in dark mode but not in light mode

oSumAtrIX: image

I've updated the colors to the default palette GitHub suggests

oSumAtrIX: One thing to consider is to switch the ci and chore colors. chore seems to be used more often but has a more annoying color.

fregante: Which is why I’d rather see it in neutral/gray color. Anyway this is good enough for now, this week I’ll be tweaking the feature after merge.

fregante:
Can you post a screenshot of the other page too? I’ve only been seeing the repo commit list page. 

Just attach it to a comment and I’ll merge the PR

fregante:
For the record, after merge these are the changes I’ll want to explore:

- use pastel colors for all
- only place the type in the bubble: `(Chore) release: v1.2.3`

This should help make the feature appear less disruptive/flashy. I’ll also:

- create a commit list with all the types in the sandbox repo so it can be used as a complete test URL 
- apply some of the changes I mentioned earlier (e.g. change the output of the parser function, update the tests to follow Vitest patterns)

oSumAtrIX:
> Can you post a screenshot of the other page too? I’ve only been seeing the repo commit list page.
> 
> Just attach it to a comment and I’ll merge the PR


https://github.com/user-attachments/assets/5c6a5eaf-7062-4bb0-846c-9ffdfab0c2db

> use pastel colors for all

I think using the default GitHub colors makes sense in terms of it blending in with the design style. Pastel colors can look nice and maybe better but would stop blending in with GitHub's design.

> only place the type in the bubble: (Chore) release: v1.2.3

This reminds me of your good argument about not modifying commit messages. Rearranging words deviates further from the original message. Currently, we only remove brackets and colons, which serve syntactic purposes, without significantly altering the core of the commit message.

Your suggestion to move the 'scope' (not the 'type') to the front raises a question: why prioritize something less semantically valuable over something more? Scopes are typically used in large-scale scenarios to categorize commits by areas like frontend, backend, or modules:

image

 However, the commit type addresses the commit message directly.

> create a commit list with all the types in the sandbox repo so it can be used as a complete test URL

I would've liked to be able to specify a mapping from commit type to color in refined-github. This way the user would be able to use their own conventional types. semantic-release allows this for example. Also:

image

But I understand the design of refined-github.

Other than that, this PR has a green light from me too :) thanks for the reviews.

fregante:
> Rearranging words deviates further from the original message.

Does it? I suggested that style because that’s how they appear to be rendered in the release notes. 

```
fix(player): pause on click
``` 

is turned into

### Bugfixes

- player: pause on click

fregante:
So, why do you think that `Fix:` commit was skipped in your case? Do I need to make the regex case-insensitive?

oSumAtrIX:
> > Rearranging words deviates further from the original message.
> 
> Does it? I suggested that style because that’s how they appear to be rendered in the release notes.
> 
> ```
> fix(player): pause on click
> ```
> 
> is turned into
> 
> ### Bugfixes
> * player: pause on click

I'd say it makes sense in the context of a changelog. I misunderstood one thing you said which is to only render the type as a label and keep the scope in the commit message. I'd prefer to keep the scope in the label as the scope is metainfo which GitHub has intended labels for (Primer for example mentions that labels can be suffixed with a colon and additional information which is a perfect use case for the scope). The actual message is then presented as the commit title proceeding the commit type and scope, which is meta information to the commit message.



> So, why do you think that `Fix:` commit was skipped in your case? Do I need to make the regex case-insensitive?

Commit types should be lowercase. I don't remember if there's a specification for that but I've not seen anyone uppercase it. What you saw in the recording was a feature branch that would've been squashed into a single commit, so we didn't pay much attention to the commit messages there and uppercased the commit type. If the spec does not mention casing for the commit type, technically as long as you are consistent with it, all casing combinations for the commit type should work. The spec mentions to be consistent though with casing referring to the beginning of the commit message after the colon.

Issue #6566, opened 2023-04-25
Move user-capability checks to API (they're currently wrong)


fregante:
We have detections like `userCanLikelyMergePR` that have to guess the capabilities of the current user, but I think it would be better to query the API.

https://github.com/search?q=repo%3Arefined-github%2Frefined-github%20userCanLikelyMergePR&type=code

https://github.com/search?q=repo%3Arefined-github%2Frefined-github%20canEditSidebar&type=code

https://github.com/search?q=repo%3Arefined-github%2Frefined-github%20canEditEveryComment&type=code

https://github.com/search?q=repo%3Arefined-github%2Frefined-github%20canCreateRelease&type=code

This can be addressed progressively. Pick a detection and make sure it works in every feature it appears.

fregante:
In some places, we're already doing it:

https://github.com/refined-github/refined-github/blob/6708dfaeeed3b4abe447ae265f525c69fa060707/source/features/update-pr-from-base-branch.tsx#L51

PR #7798, opened 2024-09-10
Improve user capability detection across multiple features

kovsu:
Closes: #6566


kovsu:
@fregante Do you have any advice?

fregante:
Looks great so far! Does it work? 

Include a link to the docs for this specific value in the source so it’s easily verifiable

kovsu:
I think it works on my end, but the actual situation may need you to test.
image

source/features/pr-branch-auto-delete.tsx
```diff
@@ -57,12 +54,11 @@ void features.add(import.meta.url, {
 	asLongAs: [
 		pageDetect.isPRConversation,
 		pageDetect.isOpenPR,
-		canCreateRelease,
+		userHasPushAccess,
```
fregante, line 57: See note.

Technically this should be:
```suggestion
		async () => {
			await elementReady('PR header')
			const pr = getBranches();
			return userHasPushAccess(pr.head.nameWithOwner)
		},
```

Not sure it's worth it though. I'll just leave a note for now

source/features/quick-comment-edit.tsx
```diff
@@ -49,27 +50,13 @@ function addQuickEditButton(commentDropdown: HTMLDetailsElement, {signal}: Signa
...
-		'.lock-toggle-link > .octicon-lock',
-
-		// Some pages like `isPRFiles` does not have a lock button
-		// These elements only exist if you commented on the page
-		'[aria-label^="You have been invited to collaborate"]',
-		'[aria-label^="You are the owner"]',
-		'[title^="You are a maintainer"]',
-		'[title^="You are a collaborator"]',
-	]) || pageDetect.canUserEditRepo();
-}
```
fregante, line 64: My concern is that the check is no longer instantaneous :(

Maybe:

- cache the call
- also include DOM-based detections as a first check, like `elementExists(…) || await getViewerPermission() === 'something'`

source/github-helpers/get-user-permission.ts
```diff
@@ -0,0 +1,43 @@
...
+			viewerPermission
+		}
+	`);
+
+	return repository.viewerPermission;
+}
+
+export async function userIsAdmin(): Promise<boolean> {
+	const repoAccess = await getViewerPermission();
+	return repoAccess === 'ADMIN';
```
fregante, line 30: New. Replaces `canUserEditRepo` where beneficial

source/github-helpers/get-user-permission.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+/** Check if the user has complete write access to the repo (but no access to the repo Settings) */
+export async function userHasPushAccess(): Promise<boolean> {
+	const repoAccess = await getViewerPermission();
+	return repoAccess !== 'READ' && repoAccess !== 'TRIAGE';
```
fregante, line 36: Merged `userCanMergePR` and `userCanRelease` into this. I think the previous check was incorrect for TRIAGE because the can _manage PRs_ but they don't have push access. I assume this means they cannot merge PRs either.


> TRIAGE: Can read and clone this repository. Can also manage issues and pull requests.

source/github-helpers/get-user-permission.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+/** Check if the user can edit all comments and comment on locked issues on the current repo */
+export async function userIsModerator(): Promise<boolean> {
+	const repoAccess = await getViewerPermission();
+	return repoAccess !== 'READ';
```
fregante, line 42: `userCanEditEveryComment` was a bit too specific to the feature that used it.

fregante (approved):
Untested but mostly unchanged. What do you think?

source/features/quick-comment-edit.tsx
kovsu: > Why did you choose v4uncached? The regular v4 should already cache by repo.

I think I just copy this code from `get-pr-info`. But `v4` it's better.

fregante:
Just for the record, this PR shows an error on: https://github.com/fregante?tab=repositories

Screenshot 1

Two issues:

- likely a feature calls "userHasPushAccess" before verifying it's on a repo page
- the feature manager is not correctly catching this error. https://github.com/refined-github/refined-github/pull/7825 might fix this or need extra work

fregante:
Likely feature:

https://github.com/refined-github/refined-github/blob/96c7ffd1631ac5a61a5c0c432c4b3c26ec53f586/source/features/clear-pr-merge-commit-message.tsx#L45-L58

Fix: 

```js
	asLongAs: [
		pageDetect.isPRConversation,
		userHasPushAccess,
	],
```

fregante:
So this change actually is more severe than I thought. The PR broke the extension for all users that lack a token. I made the API call "softer" to ensure that the check never throws. All those features now require the API to work correctly, but we can't use `expectToken` because the check fails before the runner reaches `init`

fregante:
I checked the extension without token and it seems to be fine. There are also some changes coming in https://github.com/refined-github/refined-github/pull/7825 that will address `asLongAs`-related issues.

I also restored a DOM-based check for admins (cached) and for features that only require moderator access (uncached). This speeds them up quite a bit.

kovsu:
Nice work!

Issue #7679, opened 2024-08-13
Highlight the non-default branch in files/folder views


fregante:
### Description

If you visit [this page](https://github.com/refined-github/refined-github/tree/8d4dab1a7741474b23ab499fef8d4ffeecc033d9) it's not immediately clear that it's not the default branch. You might wander a bit until you realize you're looking at an old commit.

<img width="1360" alt="Screenshot 4" src="https://github.com/user-attachments/assets/e66f3449-acb1-4eb3-ac7c-e120181eeda0">

I think we should alter `default-branch-button` or add a new feature like `highlight-non-default-base-branch`.

Note that we already do this in `clean-conversation-headers` in [PRs](https://github.com/refined-github/sandbox/pull/53)

<img width="527" alt="Screenshot" src="https://github.com/user-attachments/assets/986d6c2e-951e-46d5-bbe5-62b2d615c71d">

https://github.com/refined-github/refined-github/blob/7167430953d697d7157bb3cac75d509631887260/source/features/clean-conversation-headers.tsx#L44-L47

Maybe something like this?

<img width="220" alt="Screenshot 7" src="https://github.com/user-attachments/assets/6364ef1d-6555-4901-aee6-80ceaa5ef493">

or something more explicit

<img width="355" alt="Screenshot 8" src="https://github.com/user-attachments/assets/094d1b36-0b8e-4adc-a138-47944c070856">

### Example URLs

https://github.com/refined-github/refined-github/tree/lint
https://github.com/refined-github/refined-github/commits/test/bun

Andrew6rant:
Maybe add the name of the default branch to the `default-branch-button` and highlight the current branch?

Here's how it'll look using the colors of the right sidebar's topics bubbles:
```CSS
background: var(--bgColor-accent-muted, var(--color-accent-subtle));
color: var(--fgColor-accent, var(--color-accent-fg));
```
![refined-github-fork-highlight](https://github.com/user-attachments/assets/e312dba1-811e-4fdc-b5e2-6b7064bf2316)

fregante:
I like that color! But we can’t add the branch name because space is always limited there

Andrew6rant:
Thanks! But I'm not quite sure what you mean with the crossed-out text, the default branch name (or a truncated default branch name) should be able to fit on any screen size.

PR #7804, opened 2024-09-12
`default-branch-button` - Highlight non default branch button

kovsu:
Closes: #7679 

## Test URLs

https://github.com/refined-github/refined-github/tree/sandbox/keep-branch

## Screenshot

image

image

image

image

### DIFF
--- a/source/features/default-branch-button.css
+++ b/source/features/default-branch-button.css
@@ -1,9 +1,9 @@
 .rgh-highlight-non-default-branch {
-	background-color: var(--bgColor-accent-muted, var(--rgh-red)) !important;
-	color: var(--fgColor-accent) !important;
-	border-color: var(--fgColor-accent, var(--rgh-red)) !important;
+	background-color: var(--bgColor-accent-muted, fuchsia) !important;
+	color: var(--fgColor-accent, fuchsia) !important;
+	border-color: var(--fgColor-accent, fuchsia) !important;
 }
 
 .rgh-highlight-non-default-branch svg {
-	color: var(--fgColor-accent) !important;
+	color: var(--fgColor-accent, fuchsia) !important;
 }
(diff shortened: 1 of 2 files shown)


source/features/default-branch-button.css
```diff
@@ -0,0 +1,12 @@
+.rgh-highlight-non-default-branch {
+	background-color: var(
+		--bgColor-accent-muted,
+		var(--color-accent-subtle)
```
fregante, line 4: I think these colors have been removed or deprecated years ago, we don’t need the fallbacks. We should probably remove them from the whole repo (if I’m right) and replace them with `red` so that these bugs are immediately seen, reported and fixed.

fregante: I mean in this PR let’s only change this file. The rest of the repo can follow separately

fregante (approved):
Looks great!

source/features/default-branch-button.css
kovsu: image

fregante:
No I mean literally `red` CSS keyword. Or `fuchsia` since it’s still bright and more easily greppable

kovsu:
@fregante  I think `--rgh-red` is ok.
`--rgh-red: var(--fgColor-danger, var(--color-danger-fg, #cf222e));`

fregante:
The purpose of this fallback color is to make bugs immediately visible, not to make it more beautiful. `fuchsia` will do that.

fregante:
You can see GitHub using the same strategy here in order to catch usage of certain variables:

- https://github.com/refined-github/refined-github/issues/4529

fregante:
I'm not _super_ convinced by the color, but let's try it out. It captures the attention (as desired) and hopefully it doesn't _distract_ the user too much.

PR #7813, opened 2024-09-21
`conventional-commits` - Cleanup

fregante:
- Follows https://github.com/refined-github/refined-github/pull/7796

### Addresses

- support titles that contain HTML
- `parseConventionalCommit` now rejects invalid types
- use CSS attributes for type matching
- avoid hardcoded padding values
- reduce label to type
- add test URLs

### Test URLs

- Repo commits: https://github.com/refined-github/sandbox/commits/conventional-commits/
- PR commits: https://github.com/refined-github/sandbox/pull/91/commits
- Real data: https://github.com/semantic-release/semantic-release/commits
- Repo without conventional commits: https://github.com/refined-github/refined-github/commits

### Demo

<table>
<tr>
	<td>light 
	<td>dark
</table>

### DIFF
--- a/source/features/conventional-commits.tsx
+++ b/source/features/conventional-commits.tsx
@@ -16,16 +16,17 @@ function renderLabelInCommitTitle(commitTitleElement: HTMLElement): void {
 	}
 
 	commitTitleElement.prepend(
-		<span className="IssueLabel hx_IssueLabel mr-1" rgh-conventional-commits={commit.type}>
-			{commit.label}
+		<span className="IssueLabel hx_IssueLabel mr-2" rgh-conventional-commits={commit.rawType}>
+			{commit.type}
 		</span>,
+		commit.scope!,
 	);
 
 	removeCommitAndScope(textNode);
 }
 
 function initRepoCommitList(signal: AbortSignal): void {
-	observe(`:is(${commitTitleInLists}) h4 > span`, renderLabelInCommitTitle, {signal});
+	observe(`:is(${commitTitleInLists}) h4 > span > a:first-child`, renderLabelInCommitTitle, {signal});
 }
 
 function initPrCommitList(signal: AbortSignal): void {
--- a/source/features/conventional-commits.tsx
+++ b/source/features/conventional-commits.tsx
@@ -42,7 +42,7 @@ Test URLs:
 
 - Repo commits: https://github.com/refined-github/sandbox/commits/conventional-commits/
 - PR commits: https://github.com/refined-github/sandbox/pull/91/commits
-- Real data: https://github.com/semantic-release/semantic-release/commits
+- Real data: https://github.com/conventional-changelog/standard-version/commits
 - Repo without conventional commits: https://github.com/refined-github/refined-github/commits
 
 */
--- a/source/features/conventional-commits.tsx
+++ b/source/features/conventional-commits.tsx
@@ -19,7 +19,7 @@ function renderLabelInCommitTitle(commitTitleElement: HTMLElement): void {
 		<span className="IssueLabel hx_IssueLabel mr-2" rgh-conventional-commits={commit.rawType}>
 			{commit.type}
 		</span>,
-		commit.scope!,
+		commit.scope ? <em>{commit.scope}</em> : '',
 	);
 
 	removeCommitAndScope(textNode);
@@ -49,7 +49,8 @@ void features.add(import.meta.url, {
 
 Test URLs:
 
-https://github.com/semantic-release/semantic-release/commits/master/
-https://github.com/ReVanced/revanced-patches/commits/main/
+- Repo commits: https://github.com/refined-github/sandbox/commits/conventional-commits/
+- PR commits: https://github.com/refined-github/sandbox/pull/91/commits
+- Real data: https://github.com/semantic-release/semantic-release/commits
 
 */
--- a/source/features/conventional-commits.tsx
+++ b/source/features/conventional-commits.tsx
@@ -19,7 +19,8 @@ function renderLabelInCommitTitle(commitTitleElement: HTMLElement): void {
 		<span className="IssueLabel hx_IssueLabel mr-2" rgh-conventional-commits={commit.rawType}>
 			{commit.type}
 		</span>,
-		commit.scope ? <em>{commit.scope}</em> : '',
+		// Keep scope outside because that's how they're rendered in release notes as well
+		commit.scope ? <span style={{opacity: 0.7}}>{commit.scope}</span> : '',
 	);
 
 	removeCommitAndScope(textNode);
--- a/source/features/conventional-commits.tsx
+++ b/source/features/conventional-commits.tsx
@@ -7,38 +7,21 @@ import observe from '../helpers/selector-observer.js';
 import {commitTitleInLists} from '../github-helpers/selectors.js';
 import {parseConventionalCommit, removeCommitAndScope} from '../helpers/conventional-commits.js';
 
-const types = new Map([
-	['feat', 'Feature'],
-	['fix', 'Fix'],
-	['chore', 'Chore'],
-	['docs', 'Docs'],
-	['build', 'Build'],
-	['refactor', 'Refactor'],
-	['test', 'Test'],
-	['ci', 'CI'],
-	['perf', 'Performance'],
-]);
-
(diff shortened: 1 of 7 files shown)


source/helpers/conventional-commits.ts
```diff
@@ -1,26 +1,50 @@
...
-	zipTextNodes(element, temporaryDiv);
-
-	element.querySelector('s')!.remove();
+/**
+Remove the raw commit prefix from the single text node.
+Note: It does not support pre-formatted titles like: fix(#1 `x`)
+*/
+export function removeCommitAndScope(textNode: Text | ChildNode): void {
+	assertNodeContent(textNode, conventionalCommitRegex);
+	textNode.textContent = textNode.textContent.replace(conventionalCommitRegex, '');
```
fregante, line 49: <table>
<tr>
	<th>Before
	<th>After
<tr>
	<td>Screenshot 1
	<td>Screenshot 6
</table>

source/features/conventional-commits.tsx
```diff
@@ -7,65 +7,42 @@ import observe from '../helpers/selector-observer.js';
...
 	}
 
-	commitTitleElement.prepend(createLabelElement(type, scope));
-	removeCommitAndScope(commitTitleElement, match!);
-}
+	commitTitleElement.prepend(
+		<span className="IssueLabel hx_IssueLabel mr-2" rgh-conventional-commits={commit.rawType}>
+			{commit.type}
+		</span>,
+		commit.scope ? <em>{commit.scope}</em> : '',
```
fregante, line 22: Badges shouldn't really contain that much information. This matches the size to GitHub's own usage of labels/bubbles.



<table>
<tr>
	<th>Before
	<th>After
<tr>
	<td>
Screenshot 3
	<td>
Screenshot 4

</table>

source/features/conventional-commits.tsx
```diff
@@ -7,65 +7,42 @@ import observe from '../helpers/selector-observer.js';
...
 
-function initRepoCommitList(signal: AbortSignal): void {
-	observe(`:is(${commitTitleInLists}) h4 > span`, renderLabelInCommitTitle, {signal});
+	removeCommitAndScope(textNode);
 }
 
-function initPrCommitList(signal: AbortSignal): void {
-	observe(`:is(${commitTitleInLists}) a`, renderLabelInCommitTitle, {signal});
+function init(signal: AbortSignal): void {
+	observe(`:is(${commitTitleInLists}) h4 > span > a:first-child`, renderLabelInCommitTitle, {signal});
```
fregante, line 29: I don't know why this was split between PR and Repo commit lists, the DOM was identical and this works on both.

source/features/conventional-commits.tsx
```diff
@@ -7,65 +7,42 @@ import observe from '../helpers/selector-observer.js';
...
-	['feat', 'Feature'],
-	['fix', 'Fix'],
-	['chore', 'Chore'],
-	['docs', 'Docs'],
-	['build', 'Build'],
-	['refactor', 'Refactor'],
-	['test', 'Test'],
-	['ci', 'CI'],
-	['perf', 'Performance'],
-]);
```
fregante, line 20: Moved all to the parsing stage.

source/features/conventional-commits.tsx
```diff
@@ -7,65 +7,42 @@ import observe from '../helpers/selector-observer.js';
...
-	init: initPrCommitList,
+	init,
 });
 
 /*
 
 Test URLs:
 
-https://github.com/semantic-release/semantic-release/commits/master/
-https://github.com/ReVanced/revanced-patches/commits/main/
```
fregante, line 69: No PR commit list as I had suggested

source/helpers/conventional-commits.ts
fregante: This indeed breaks support for references and backticks, but also restores the links on those issue references (which is preferable anyway)


<table>
<tr>
	<th>Before
	<th>After
<tr>
	<td>Screenshot 5
	<td>Screenshot 7


</table>

fregante (commented):
cc @oSumAtrIX

source/helpers/conventional-commits.ts
oSumAtrIX: I don't know how I feel about the after. Maybe you missed this https://github.com/refined-github/refined-github/pull/7796#issuecomment-2357085761.

Additionally you add arbitrary cursive formatting to the scope.

 > This indeed breaks support for references and backticks, but also restores the links on those issue references (which is preferable anyway)

Is ther no way to retain references and backticks in the label? Pretty sure you can just wrap all the elements as a child of the label element.

source/features/conventional-commits.tsx
oSumAtrIX: This formatting makes it look like the meta information "scope" belongs to the commit message, which is not the case. The scope is meta information for the commit message that is followed by the colon. If at all, it should be visually and semantically grouped with the commit type. The length you see in the commits is specific to the repository you have screenshotted. Scopes usually are not very long as shown in the conventional commit page: 
image

source/helpers/conventional-commits.ts
fregante: I did not miss those, I just disagreed. 

> arbitrary cursive formatting to the scope.

If you look at a conventional changelog, you’ll see that the scope is bold. Since the whole commit title is bold, I made it cursive. Hardly arbitrary. 

> Is ther no way to retain references and backticks in the label? Pretty sure you can just wrap all the elements as a child of the label element.

There sure is, but it’s hard and it already took days of reviews to pass a barebones version, I don’t want to spend days reviewing code that aims to preserve the DOM without drawbacks.

source/features/conventional-commits.tsx
fregante: Are you sure? Have you looked at a changelog? This is exactly what a conventional changelog looks like, I already mentioned this before.

source/helpers/conventional-commits.ts
oSumAtrIX: > Since the whole commit title is bold, I made it cursive

This is arbitrary. Why not make it normal? Why not make it uppercase? Why cursive?

> There sure is, but it’s hard and it already took days of reviews to pass a barebones version, I don’t want to spend days reviewing code that aims to preserve the DOM without drawbacks.

The only difference that needs to be done is to collect all the elements to an array until the colon , and then add as content of the label.

source/features/conventional-commits.tsx
oSumAtrIX: Yes, I am. You can check many other conventional commit reports, and you'll see them using short scopes.

fregante: I’m not talking about the length but about the scope usage/display:

IMG_5675

oSumAtrIX: The scope of a commit is _meta_ information. If you were to have a repository with conventional commits and had the task to revert using conventional commits, you'd get rid of everything before the colon. The actual information is the commit message after the colon.

source/helpers/conventional-commits.ts
fregante: In typography bold and italics are used to give emphasis. Again, since the whole title is already bold, I used the second style. 

> The only difference that needs to be done is to collect all the elements to an array until the colon , and then add as content of the label.

Easier said than done. I’m done discussing this.

oSumAtrIX: Unfortunately, I am not comfortable with this change due to the reasons I have mentioned, so I'll stop the discussion and move forward with a fork that reflects the changes I have in mind. Thanks regardless.

fregante:
Screenshot updated

Screenshot 4

oSumAtrIX:
One more idea that came up is to add a second label for the scope in a lighter tint (or just grey)

fregante:
I reviewed the spec at https://www.conventionalcommits.org/en/v1.0.0/ and found that:

- scope should "a noun", so technically multiple words shouldn't even be recognized as valid
- this also means <code>feat(\`backticks\`): foo</code> isn't valid either, and it's fine if we don't support it
- the prefix can also include `!` to mark major changes


This to show there were arguments about out-of-spec stuff while also ignoring the actual spec.

The feature is finally ready to be shipped

fregante:
@oSumAtrIX By the way I opened a feedback issue for this feature. Let's refrain from commenting there

PR #7817, opened 2024-09-22
Trying out rsbuild

fregante:
- [ ] Blocked by https://github.com/web-infra-dev/rsbuild/discussions/3529


fregante:
don't like it

PR #7836, opened 2024-09-30
`tags-on-commits` - Restore feature

kovsu:
https://github.com/refined-github/refined-github/issues/7808#issuecomment-2363295100

## Test URLs

## Screenshot

image

image

### DIFF
--- a/source/github-helpers/selectors.ts
+++ b/source/github-helpers/selectors.ts
@@ -136,8 +136,8 @@ export const newCommentField = [
 export const newCommentField_ = [] satisfies UrlMatch[];
 
 export const commitHashLinkInLists = [
-	'[data-testid="commit-row-browse-repo"]',
-	'[aria-label="View commit details"] a.text-mono', // `isCommitList`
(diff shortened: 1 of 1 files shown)


source/github-helpers/selectors.ts
```diff
@@ -136,6 +136,7 @@ export const newCommentField = [
 export const newCommentField_ = [] satisfies UrlMatch[];
 
 export const commitHashLinkInLists = [
+	'[data-testid="commit-row-browse-repo"]',
```
fregante, line 139: We need a comment like the ones below. Where does this apply and how is it different? Which one does this deprecates? We should mark the old one with "TODO: remove in May 2025”

kovsu: I wonder how we know the date.

fregante: We leave the old selectors in for ~6 months to attempt to support GitHub Enterprise. We only do this when it's easy to support it.

PR #7846, opened 2024-10-02
`hide-interaction-limits-warning` - Remove the interaction limits warning

LitoMore:
## Test URLs

Any organization repository homepage https://github.com/refined-github/refined-github

This warning message is only visible to organization maintainers. Need the feature to be enabled on organization level: https://github.com/organizations/refined-github/settings/interaction_limits


fregante:
Why do you want to remove it? It’s important to have it there so you know the limits are enabled

LitoMore:
As a maintainer, this is annoying when I see it every day open my project. I'm aware of this behavior since I enabled it from the organization settings. GitHub doesn't even give an option to hide it.

fregante:
What happens when the time expires and you don’t have any visual element telling you to re-enable it? I don’t think this is a good default. 

You can hide it via CSS though, with something like:

``` css
.flash:has(a[href*='interaction-limits']) {
  display: none
}
``` 

or with your ad blocker

PR #7847, opened 2024-10-02
`conventional-commit` - Add support for type `meta`

WikiRik:
This PR adds support for the type `meta` in `conventional-commit`. 
No need for a special color. The default, used for `chore`, `style` and `revert`, will do just fine imo.

## Test URLs

https://github.com/sequelize/sequelize/commits/main/

## Screenshot

Will add later


fregante:
The list is not customizable. We follow https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-conventional#rules

I'll add the link to the file

WikiRik:
Thanks for the quick response!

PR #7848, opened 2024-10-02
`sidebar-focus-file` - New feature

foriequal0:
GitHub should scroll the file tree to the current file.
But it seems that GitHub couldn't do it properly when you're logged in
(except when you're navigating to a file from the directory view, or to a directory from the file path segments)

I reproduced it with Firefox 132.0b1, and Chrome 129.0, without Refined GitHub extension.

This extension attempts to fix that.

## Test URLs

The current file should be in the center of the view on the following cases:
- pages
  - on the directory page
https://github.com/refined-github/refined-github/tree/main/source/features (on directory)
  - on the file page
https://github.com/refined-github/refined-github/blob/main/source/refined-github.ts (on file)

- viewport width, and file tree mode (no problem before)
  - the viewport is wide and
    - the file tree is expanded by default
    - the file tree is collapsed and expanded again manually
  - the viewport is narrow and
    - the file tree is collapsed by default and expanded manually

- after
  - refreshing on the file page
  - refreshing on the directory page
  - navigating to other files/directories through the file tree (no problem before)
  - navigating to other directories through the file path (no problem before)
  image
  - back to the file/directory page from other tabs (Pull Requests, Issues, etc.)

## Known issue

sometimes it scrolls the entire window a little bit.

## Screenshot

image


fregante:
Thanks for the PR. Generally we'd want to discuss this first in an issue, but the feature should be rather simple.

From what I see, this behavior already works well in most scenarios, the only time it doesn't work is on the first load of the code viewer, which is when you open the link https://github.com/refined-github/refined-github/blob/30114ade84c18aba40454c8dd42dc8268ec80794/source/features/link-to-changelog-file.gql from anywhere other than the repo's **Code** tab.

https://github.com/user-attachments/assets/6defa9b5-ee3c-486f-9b6f-3db45ee57acf

build/features.test.ts
```diff
@@ -30,6 +30,7 @@ const noScreenshotExceptions = new Set([
 	'selection-in-new-tab',
 	'submission-via-ctrl-enter-everywhere',
 	'click-outside-modal',
+	'scroll-to-current-file',
```
fregante, line 33: No exceptions for this one, it needs a screenshot with an arrow pointing to the sidebar

build/__snapshots__/features-meta.json
```diff
@@ -788,6 +788,11 @@
 		"description": "Preserves current branch and path when viewing all commits by an author.",
 		"screenshot": "https://user-images.githubusercontent.com/44045911/148764372-ee443213-e61a-4227-9219-0ee54ed832e8.png"
 	},
+	{
+		"id": "scroll-to-current-file",
+		"description": "Scrolls the file tree to the current file when signed in.",
```
fregante, line 793: We don't care about the signed out experience, so no need to mention "signed in" here

foriequal0:
Thanks for the quick reply!
You're right. I'll create a new discussion on the GitHub community here https://github.com/orgs/community/discussions

source/features/scroll-to-current-file.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+function scrollToCurrentFile(): void {
+	const url = new GitHubFileURL(location.href);
+	const filePath = url.filePath;
+
+	// we need to escape the filePath if we use querySelector.
+	// eslint-disable-next-line unicorn/prefer-query-selector
+	const item = document.getElementById(`${filePath}-item`);
+	const text = item?.querySelector('.PRIVATE_TreeView-item-content-text');
+	text?.scrollIntoView({block: 'center', behavior: 'auto'});
```
fregante, line 14: I think you just need this:

```suggestion
	scrollIntoViewIfNeeded(document.getElementById(`${filePath}-item`));
```

We have a `scrollIntoViewIfNeeded` function

source/features/scroll-to-current-file.tsx
```diff
@@ -0,0 +1,57 @@
...
+	asLongAs: [
+		pageDetect.isLoggedIn,
+	],
+	include: [
+		pageDetect.isRepoTree,
+		pageDetect.isSingleFile,
+	],
+	init,
+	awaitDomReady: true,
+});
```
fregante, line 48: I think you only need this according to what I mentioned in my comment

```suggestion
function init(): void {
	scrollToCurrentFile();
}

void features.add(import.meta.url, {
	include: [
		pageDetect.isRepoTree,
		pageDetect.isSingleFile,
	],
	exclude: [
		pageDetect.isRepoRoot,
	],
	awaitDomReady: true,
	init,
});
```

fregante (changes requested):

fregante:
> I'll create a new discussion on the GitHub community here [github.com/orgs/community/discussions](https://github.com/orgs/community/discussions)

I mean an issue on this repo. That community is different from this repo

build/features.test.ts
foriequal0: Thanks. I was about to remove it from exceptions once I figured out how to get a public link to the image.
The new screenshot attached to the bottom of the PR would be okay?

source/features/scroll-to-current-file.tsx
foriequal0: The Directory item wraps the entire child items, so it needed to select the first text child.
Thanks for the `scrollIntoViewIfNeeded` function.

source/features/scroll-to-current-file.tsx
foriequal0: It seems that GitHub adjusts the layout/scroll multiple times after the DOM is ready. It jiggles on the load when `turbo:load` listener is removed.

https://github.com/user-attachments/assets/49efc45e-e450-4bc4-88b3-f41e6d7ed6bb

It seems that `soft-nav:end` is not needed anymore since the file tree works fine by default. (It used to be problematic)

foriequal0:
> I mean an issue on this repo. That community is different from this repo

Sorry for the misunderstanding.
I'll create a discussion there soon anyway since it is better to fix a bug on the upstream.

build/features.test.ts
fregante: > The new screenshot attached to the bottom of the PR would be okay?

Yes

source/features/scroll-to-current-file.tsx
fregante: `turbo:load` is fired at every page load so we don't want to mess with the native scrolling. You might want to find another event or just use `setTimeout(scrollToCurrentFile, 500)` inside `init`

source/features/scroll-to-current-file.tsx
```diff
@@ -0,0 +1,54 @@
...
+}
+
+function unload(): void {
+	// After the file navigation
+	window.removeEventListener('turbo:load', scrollToCurrentFile);
+}
+
+void features.add(import.meta.url, {
+	asLongAs: [
+		pageDetect.isLoggedIn,
```
fregante, line 37: Check my example code. `asLongAs` should be dropped and `exclude` should be added. `init` must be last in the object

source/features/scroll-to-current-file.tsx
foriequal0: Okay. I was trying to center the directory but it doesn't seem necessary. I'll change that.

source/features/scroll-to-current-file.tsx
foriequal0: "turbo:load" has a similar lifecycle to `init` func.
It isn't fired when you just navigate between files.
It is fired only when you navigate between file/dir <-> Root, Issues, Pull Requests, etc.

Also, I found that the invocation timing of `init` is inconsistent.
It is called before when you open an new link, but it is called after `turbo:load` when you navigate back to the file/directory view.
It might be due to the preceding `async/await` in the `setupPageLoad`.

foriequal0: I could fix the event timing, but it seems messy to play with `Promise`. I'll just stick to `setTimeout`

foriequal0:
I've changed to simple `setTimeout`.

This is what it looks like when `setupPageLoad` is fixed.
https://github.com/foriequal0/refined-github/tree/scroll-to-current-file-2

This is the discussion:
https://github.com/orgs/community/discussions/140389

source/features/scroll-to-current-file.tsx
```diff
@@ -0,0 +1,42 @@
...
+	await delay(500);
+
+	const {filePath} = new GitHubFileURL(location.href);
+
+	// eslint-disable-next-line unicorn/prefer-query-selector -- `querySelector` requires escaping
+	const item = document.getElementById(`${filePath}-item`);
+	if (item) {
+		scrollIntoViewIfNeeded(item);
+	} else {
+		return false;
```
fregante, line 19:
```suggestion
		// The sidebar might be closed
		return false;
```

source/features/scroll-to-current-file.tsx
```diff
@@ -0,0 +1,42 @@
...
+async function init(): Promise<void | false> {
+	// The sidebar is loaded a bit later
+	await delay(500);
+
+	const {filePath} = new GitHubFileURL(location.href);
+
+	// eslint-disable-next-line unicorn/prefer-query-selector -- `querySelector` requires escaping
+	const item = document.getElementById(`${filePath}-item`);
+	if (item) {
+		scrollIntoViewIfNeeded(item);
```
fregante, line 17:
```suggestion
		// This feature is only needed for the very first load of the view.
		// GitHub does it natively after that.
		scrollIntoViewIfNeeded(item);
```

fregante (approved):
I think it's fine, we just need a couple of comments

Issue #7231, opened 2024-01-25
Mark more clearly that the token is "required"


fregante:
Every so often someone is confused as to why some features don't work. Since features vary over time and occasionally start requiring the token, people are further confused as to why they stop working.

This makes no sense and I don't want to support token-less usage anymore. People can continue to use RG without it, but then we don't want to waste time figuring that out.

Why?

- the token can be read-only and limited to public repos
- our permissions already only target GitHub.com

Changes suggested:

- [x] Update wording in [welcome issue](https://togithub.com/refined-github/refined-github/issues/3543)
- [x] https://github.com/refined-github/refined-github/issues/7382
- [x] Improve wording in options page to suggest the above
- [ ] Show a block screen on the bug report page to invite people to use the token before opening an issue. I'm not sure if this screen can be made closeable (to allow people to report real bugs) or not (to block people from reporting token-related non-bugs)

Related:

- https://github.com/refined-github/refined-github/issues?q=+to+set+the+token+in%3Atitle
- https://github.com/refined-github/refined-github/issues/7163#issuecomment-1909923561
- https://github.com/refined-github/refined-github/pull/7221#issuecomment-1894910985

PR #7844, opened 2024-10-01
Fix a few options page bugs

fregante:

PR #7854, opened 2024-10-04
Improve first-install flow to facilitate token input

fregante:
- #7231 

I think people install the extension and don't realize that it asks for a token. This is somewhat understandable, at least until they open the options page (if ever) or look at the console. More likely they will just complain that the UI doesn't work the way they expect it to. 

A second issue is that in Safari they will be shown the Welcome Page but Refined GitHub is not loaded yet because Safari does not automatically grant access on install. This UI merges both needs into one step.

## Safari

https://github.com/user-attachments/assets/a104c286-5203-4865-ab63-0aa8f658b8c8

## Chrome + error demo

https://github.com/user-attachments/assets/dc60b6e8-73fe-4434-b5b0-33c0238942e8

## Safari iOS

https://github.com/user-attachments/assets/217de10e-95d3-4671-8d89-0fb1183e3732


rollup.config.js
```diff
@@ -9,13 +9,16 @@ import json from '@rollup/plugin-json';
 import copy from 'rollup-plugin-copy';
 import del from 'rollup-plugin-delete';
 import webpackStatsPlugin from 'rollup-plugin-webpack-stats';
+import svelte from 'rollup-plugin-svelte';
+import {sveltePreprocess} from 'svelte-preprocess';
```
fregante, line 13: The cool part is that this PR finally adds support for proper dynamic components that can even be used via our native-DOM JSX solution. This will eventually help features on React views, for example:

- https://github.com/refined-github/refined-github/issues/6554

Related:

rollup.config.js
```diff
@@ -46,6 +49,12 @@ const rollup = {
 			targets: ['distribution/assets'],
 			runOnce: true, // `false` would be nice, but it deletes the files too early, causing two extension reloads
 		}),
+		svelte({
+			compilerOptions: {
+				customElement: true,
```
fregante, line 54: Svelte is probably the best solution here:

- no JSX conflicts
- dedicated .svelte file that can live next to the .tsx file as `<feature-name>.svelte`
- custom elements support, so these components can be used in JSX as `<rgh-something count={1}/>`

source/background.ts
```diff
@@ -89,7 +89,7 @@ chrome.runtime.onInstalled.addListener(async ({reason}) => {
 	// Only notify on install
 	if (await isFirstInstall(reason)) {
 		await chrome.tabs.create({
-			url: 'https://github.com/refined-github/refined-github/issues/3543',
+			url: chrome.runtime.getURL('assets/welcome.html'),
```
fregante, line 92: The old "welcome page" is now shown as a second step.

source/features/rgh-options-link.tsx
```diff
@@ -0,0 +1,34 @@
+/* You can have invisible/inactive links to the options page on the wiki. This feature enables them. */
```
fregante, line 1: Two changes:

- it now applies to the entire wiki, and maybe the entire repo in the future
- it simplifies the markup while remaining invisible without extension
  
  ```diff
  - [](#rgh-linkify-welcome-issue)extension’s options[](#rgh-linkify-welcome-issue)
  + <a name="options-page-link">extension’s options</a>
  ```

source/options.css
```diff
@@ -127,7 +127,7 @@ details[open] > :not(summary) {
 }
 
 [name='personalToken'] {
-	width: 30em !important;
+	width: 20em !important; /* https://github.com/refined-github/refined-github/issues/1374#issuecomment-397906701 */
```
fregante, line 130: Unrelated, undoes a change from:

- #7844

source/options.html
```diff
@@ -12,7 +12,7 @@
 	<details id="info" open>
 		<summary><img src="icon.png" alt="" width="24" height="24"> Refined GitHub</summary>
 		<p>
-			Visit the <a href="https://github.com/refined-github/refined-github/issues/3543">welcome page</a> to learn about updates, debugging, and GitHub Enterprise.
+			Visit the <a href="https://github.com/refined-github/refined-github/wiki">welcome page</a> to learn about updates, debugging, and GitHub Enterprise.
```
fregante, line 15: The welcome issue will now live in the wiki

source/welcome.css
```diff
@@ -0,0 +1,86 @@
...
+li::before {
+	content: '';
+	display: inline-block;
+	width: 1em;
+	height: 1em;
+	vertical-align: -0.2em;
+	margin-right: 0.3em;
+	background-image: url('data:image/svg+xml;utf8,<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16"><path fill-rule="evenodd" fill="gray" d="M8 5.5a2.5 2.5 0 100 5 2.5 2.5 0 000-5zM4 8a4 4 0 118 0 4 4 0 01-8 0z"></path></svg>');
+	background-size: contain;
+}
```
fregante, line 58: Styles come from the options page. Eventually I might componentize this.

source/welcome.svelte
```diff
@@ -0,0 +1,126 @@
+<svelte:options customElement='rgh-welcome' />
```
fregante, line 1: This file is a bit messy. Suggestions on how to make it more "svelte-like" very much welcome.

source/welcome.svelte
```diff
@@ -0,0 +1,126 @@
...
+
+<script lang='ts'>
+	import './welcome.css';
```
fregante, line 4: One thing I skipped (for now?) is the use of `<style>` tags because I can't format the CSS in it without also formatting the JS:

- https://github.com/sveltejs/prettier-plugin-svelte/issues/55

JS formatting is taken care of by eslint.

Maybe one day with:

- https://github.com/refined-github/refined-github/pull/7584#issuecomment-2245113330

yakov116:
Looks awesome!!

I would love to test it, I will ty it later and get back to you.

source/background.ts
```diff
@@ -81,15 +80,15 @@ async function isFirstInstall(suggestedReason: string): Promise<boolean> {
...
 
 		// Safari reports "install" even on updates #5412
-		&& !(isSafari() && await hasUsedStorage())
+		&& !(await hasUsedStorage())
 	);
 }
 
 chrome.runtime.onInstalled.addListener(async ({reason}) => {
 	// Only notify on install
 	if (await isFirstInstall(reason)) {
```
fregante, line 89: @yakov116 Note that this needs to be set to `true` to work since we disable in on dev builds

fregante:
## Match GitHub style

chrome-extension___jialghbnladbdlhaibaoejjlnihkmemc_assets_welcome html (1)
chrome-extension___jialghbnladbdlhaibaoejjlnihkmemc_assets_welcome html (2)
chrome-extension___jialghbnladbdlhaibaoejjlnihkmemc_assets_welcome html(iPhone SE) (1)

source/welcome.svelte
```diff
@@ -0,0 +1,132 @@
...
+				Grant
+			{/if}
+			the extension access to github.com
+		</li>
+
+		<!-- svelte-ignore a11y-no-noninteractive-element-interactions -->
+		<!-- svelte-ignore a11y-click-events-have-key-events -->
+		<li class:valid={stepValid >= 2} class:visible={stepVisible >= 2} class='will-show' on:click={showThirdStep}>
+			<a
+				href='https://github.com/settings/tokens/new?description=Refined%20GitHub&scopes=repo,read:project'
```
fregante, line 98: @djhoese thanks for the comment!

The link should already pre-select the minimal scopes we expect, see the link here. Does it not happen for you? What's your browser? This is what you should see:

IMG_6045

djhoese: Ah, that would do it. I didn't click that link. I went to the token page myself. Makes sense.

fregante: I'll see if I can add that information in the wiki as well though, and/or I'll bring out the same UI we have in the options page.

PR #7866, opened 2024-10-07
`quick-repo-deletion` - Two-click deletion with key modifiers

### LINKED ISSUE #7865: `quick-repo-deletion`: Reintroduce old behaviour
### Description

This feature's behaviour was changed in #7826. I understand that this was done to minimize the risk of something going wrong, but it is now way too inconvenient to delete a fork.

* Before: One button click.
* Now:
	* One click to go to the settings with the panel displayed.
	* One click to confirm that I indeed want to delete the repository.
	* One click to confirm that I understand the effects of doing so.
	* One click to submit the form.
	* 2FA authentication; either:
		* One click to send a notice to my phone, then turn it on and tap the notification.
		* Type my password, then one click to submit.

Granted, the last step is only necessary once every two hours, but it is quite rare for me to delete more than one fork in the same day, let alone two hours.

Could the old behaviour be reintroduced somehow? Perhaps via a disabled-by-default feature?

### Example URLs

https://github.com/InSyncWithFoo/some-non-existent-fork

InSyncWithFoo:
Resolves #7865.

I managed to make <kbd>Ctrl</kbd> + <kbd>Alt</kbd> + Clicking work for the normal button, but I'm not sure about the one in the settings page; our event listener doesn't seem to be called at all.

### Test

1. Fork a repo, like https://github.com/left-pad/left-pad
2. Shift-alt-click "Delete fork"


source/features/quick-repo-deletion.tsx
```diff
@@ -22,12 +34,122 @@ async function isRepoUnpopular(): Promise<boolean> {
...
+				Repository <strong>{nameWithOwner}</strong> deleted. <a href={restoreURL}>Restore it</a>, <a href={forkSource}>visit the source repo</a>, or see <a href={otherForksURL}>your other forks.</a>
+			</span>
+		</>,
+		{action: false},
+	);
+	$('.application-main')!.remove();
+
+	if (document.hidden) {
+		// Try closing the tab if in the background. Could fail, so we still update the UI above
+		void messageBackground({closeTab: true});
```
fregante, line 110: We can drop this because we no longer have the timeout. The user has no time to switch tabs.

source/features/quick-repo-deletion.tsx
```diff
@@ -22,12 +34,122 @@ async function isRepoUnpopular(): Promise<boolean> {
...
+async function performDeletion(button: DeleteButton): Promise<void> {
+	if (!(await tokenHasDeleteRepoScope())) {
+		notifyMissingTokenScope();
+		return;
+	}
+
+	const originalButtonText = button.textContent;
+	const {nameWithOwner} = getRepo()!;
+
+	setButtonText(button, 'Deleting repo…');
```
fregante, line 123: Can you use our toast component instead? The same component can also contain the error. We don't need to create a banner/notice anymore. 

The component can be use around the whole `performDeletion`, like:

```js
await showToast(performDeletion, {
  message: 'Deleting repo...',
  doneMessage: 'Repo deleted'
});
``` 

You can customize each error message by throwing it: `throw new Error('You need to add the delete_repo scope to your token')`

`modifyUIAfterSuccessfulDeletion` is fine though

source/features/quick-repo-deletion.tsx
```diff
@@ -22,12 +34,122 @@ async function isRepoUnpopular(): Promise<boolean> {
...
+
+	try {
+		deleteRepository(nameWithOwner);
+	} catch (error) {
+		assertError(error);
+		removeButtonIfAtRepoRoot(button);
```
fregante, line 129: No need because the button still works without modifiers.

source/features/quick-repo-deletion.tsx
```diff
@@ -22,12 +34,122 @@ async function isRepoUnpopular(): Promise<boolean> {
...
+	}
+}
+
 function addButton(header: HTMLElement): void {
 	header.prepend(
 		<li>
 			<a
 				href={buildRepoURL('settings', buttonHashSelector)}
 				className="btn btn-sm btn-danger"
+				onClick={deleteButtonClicked}
```
fregante, line 152: You should use delegate too, that's what we usually do. GitHub generates and regenerates the DOM losing listeners.

source/features/quick-repo-deletion.tsx
```diff
@@ -65,6 +191,17 @@ void features.add(import.meta.url, {
 	init: initSettingsPage,
 });
 
+features.add(import.meta.url, {
```
fregante, line 194: `features.add` accepts multiple objects, like:

`features.add(url, {init}, {init})`

source/features/quick-repo-deletion.tsx
```diff
@@ -65,6 +191,17 @@ void features.add(import.meta.url, {
...
+	asLongAs: [
+		pageDetect.isRepoSettings,
+		pageDetect.isForkedRepo,
+		canUserDeleteRepository,
+		isRepoUnpopular,
```
fregante, line 199: Only leave repo settings. The request is very explicit so if they want to delete other repos in one ctrl-alt-click then let them do it.

fregante:
Try `{capture: true}`

source/features/quick-repo-deletion.tsx
```diff
@@ -56,7 +160,11 @@ void features.add(import.meta.url, {
 		canUserDeleteRepository,
 		isRepoUnpopular,
 	],
-	init,
+	init: initRepoRoot,
+}, {
+	asLongAs: [pageDetect.isRepoSettings],
+	awaitDomReady: true,
```
fregante, line 166: No need
```suggestion
```

source/features/quick-repo-deletion.tsx
```diff
@@ -22,12 +30,103 @@ async function isRepoUnpopular(): Promise<boolean> {
...
+
+	const {nameWithOwner} = getRepo()!;
+
+	try {
+		deleteRepository(nameWithOwner);
+	} catch (error) {
+		assertError(error);
+		notifyDeletionFailure(error);
+
+		throw error;
```
fregante, line 103:
```suggestion
		throw new Error('Could not delete the repository', {cause: error})
```

source/features/quick-repo-deletion.tsx
```diff
@@ -22,12 +30,103 @@ async function isRepoUnpopular(): Promise<boolean> {
...
+async function performDeletion(): Promise<void> {
+	if (!(await tokenHasDeleteRepoScope())) {
+		notifyMissingTokenScope();
+		return;
+	}
+
+	const {nameWithOwner} = getRepo()!;
+
+	try {
+		deleteRepository(nameWithOwner);
```
fregante, line 98: If this is not awaited, errors won't be caught
```suggestion
		await deleteRepository(nameWithOwner);
```

source/features/quick-repo-deletion.tsx
```diff
@@ -22,12 +30,103 @@ async function isRepoUnpopular(): Promise<boolean> {
...
+				Repository <strong>{nameWithOwner}</strong> deleted. <a href={restoreURL}>Restore it</a>, <a href={forkSource}>visit the source repo</a>, or see <a href={otherForksURL}>your other forks.</a>
+			</span>
+		</>,
+		{action: false},
+	);
+	$('.application-main')!.remove();
+}
+
+async function performDeletion(): Promise<void> {
+	if (!(await tokenHasDeleteRepoScope())) {
```
fregante, line 90:
```suggestion
	// TODO: Add support for app tokens 
	if (!(await tokenHasDeleteRepoScope())) {
```

I think app tokens don't have the same concept of scopes, so they would fail even if the token is very capable. See the last paragraph on https://github.com/refined-github/refined-github/wiki/Security#token

source/features/quick-repo-deletion.tsx
fregante: This is weird, we have a rule called `no-floating-promises` for this. I see the same issue on `showToast`

source/features/quick-repo-deletion.tsx
```diff
@@ -95,12 +96,12 @@ async function performDeletion(): Promise<void> {
 	const {nameWithOwner} = getRepo()!;
 
 	try {
-		deleteRepository(nameWithOwner);
+		await deleteRepository(nameWithOwner);
 	} catch (error) {
 		assertError(error);
 		notifyDeletionFailure(error);
```
fregante, line 103: These were part of the code to be dropped, the error replaces the `notice`

fregante:
> @fregante The settings page button's default listener is still triggered after ours, even with all of `.preventDefault()`, `.stopPropagation()` and `.stopImmediatePropagation()`. Any ideas how to snooze it?

The reason why it's failing is because they're also using `capture` on `document`, so I'm not entirely sure we can even get it. We can close it though:

excluded

fregante:
Tested

## No token

token

## No scope

token

## Network error

offline


## Success

success

Issue #7853, opened 2024-10-04
Fix bad sizing of sponsor button on mobile


fregante:
### Description

Can’t believe this has been broken for so long. We just need 5 lines of code in github-bugs.css

![IMG_5877](https://github.com/user-attachments/assets/09ac98c0-46a7-4dc0-bca6-a5f1d1b6de18)

### Example URLs

https://github.com/refined-github/refined-github

PR #7868, opened 2024-10-08
 Fix sponsor button size in mobile view

kovsu:
close: #7853 

## Test URLs

## Screenshot

image

### DIFF
--- a/source/features/github-bugs.css
+++ b/source/features/github-bugs.css
@@ -75,6 +75,7 @@ div.markdown-body {
 }
 
 /* Sponsor button in mobile view #7853 */
+/* Test: https://github.com/refined-github/refined-github */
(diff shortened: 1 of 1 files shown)


source/features/github-bugs.css
```diff
@@ -74,6 +74,13 @@ div.markdown-body {
 	min-height: 2em !important;
 }
 
+/* Sponsor button in mobile view #7853 */
```
fregante, line 77: Thank you! Can you add a test URL? Just our repo

This does not affect the button on large windows?

kovsu: It works fine.

https://github.com/user-attachments/assets/5ee87312-3139-485b-8d5a-aa01c96825ac

fregante (approved):

PR #7871, opened 2024-10-08
Add a Copy Link Issue Feature

akirk:


akirk:
FWIW here is the above code as user script: https://gist.github.com/akirk/e92612a35620eafe32f6964f83d2258b

fregante:
Thank you for posting it here!

Issue #7856, opened 2024-10-04
⚠️ Features broken on new issues (beta)


fregante:
### Description

I'm sure every `isConversationList` feature is currently broken on that page. 

### How to replicate the issue + URL

Test URLs:

- https://github.com/download-directory/download-directory.github.io/issues/141
- https://github.com/eslint/eslint/issues
- https://github.com/tc39/proposal-await.ops/issues

The beta is enabled by organization. These three have it. You can spam issues on https://github.com/download-directory/download-directory.github.io/issues if necessary, I'll delete them once we're done. Or leave comments in https://github.com/download-directory/download-directory.github.io/issues/141, better.

fregante:
By the way, finally we have an autocomplete!

<img width="625" alt="Screenshot 3" src="https://github.com/user-attachments/assets/ca369f13-e87d-437b-8c25-064dbfc94615">

fregante:
More features are broken, like the label alignment

fregante:
Bye `quick-comment-hiding`

<img width="500" alt="Screenshot 28" src="https://github.com/user-attachments/assets/994afdf5-2504-4986-91c8-4e28f0bd1a55">

fregante:
These are features running on `isIssueList` and `isSingleIssue`

I'll check these off after verifying them, or with 👋 if they're no longer needed


### Important features


- [ ] https://github.com/refined-github/refined-github/issues/8226
- [x] `comments-time-machine-links`
- [x] https://github.com/refined-github/refined-github/issues/8586
- [x] https://github.com/refined-github/refined-github/issues/8024
- [x] #7968 


### Done

Edit February 6th: Some of these have been broken again. Please open a new issue in that case.

- [x] `quick-mention`
- [x] `selection-in-new-tab`
- [x] `shorten-links`
- [x] `linkify-user-location`
- [x] `readable-title-change-events`
- [x] `jump-to-conversation-close-event`
- [x] `show-names`
- [x] `small-user-avatars`
- [x] `sticky-sidebar`
- [x] `table-input`
- [x] `reactions-avatars`
- [x] `scrollable-areas`
- [x] `prevent-link-loss`
- [x] `collapsible-content-button`
- [x] `embed-gist-inline`
- [x] `emphasize-draft-pr-label`
- [x] `linkify-user-labels`
- [x] `jump-to-conversation-close-event`
- [ ] `avoid-accidental-submissions` 👋
- [ ] `esc-to-cancel` 👋
[...]

PR #7873, opened 2024-10-09
`more-conversation-filters` - Fix error on new beta view

kovsu:
Closes: #7856 

Because our test page doesn't have `#filters-select-menu`, so it throws an error.

## Test URLs

https://github.com/tc39/proposal-await.ops/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen

## Screenshot

image

image

### DIFF
--- a/source/features/more-conversation-filters.tsx
+++ b/source/features/more-conversation-filters.tsx
@@ -7,9 +7,9 @@ import SearchQuery from '../github-helpers/search-query.js';
 
 function init(): void {
 	const sourceItem = $('#filters-select-menu a:nth-last-child(2)');
-
-	if (!sourceItem)
+	if (!sourceItem) {
 		return;
+	}
 
 	// "Involved" filter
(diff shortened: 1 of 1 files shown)


source/features/more-conversation-filters.tsx
```diff
@@ -32,7 +33,16 @@ function init(): void {
 	commentsLink.after(subscriptionsLink);
 }
 
+// The page below doesn't have the filters menu
+// https://github.com/tc39/proposal-await.ops/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen
+async function hasFilters(): Promise<boolean> {
+	return Boolean(await elementReady('#filters-select-menu', {waitForChildren: false}));
```
fregante, line 39: The feature has `awaitDromReady`, so it doesn't need `elementReady`

But here the fix is to just add `if (!sourceItem) return` on line 10, not this

PR #7876, opened 2024-10-09
`clean-conversation-filters` - Add support for new beta view

kovsu:
Closes: #7856

## Test URLs

https://github.com/tc39/proposal-await.ops/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen

## Screenshot

image

👆 this is the usual page

👇 this is the new view

image

image

### DIFF
--- a/source/features/clean-conversation-filters.tsx
+++ b/source/features/clean-conversation-filters.tsx
@@ -43,25 +43,32 @@ function getCount(element: HTMLElement): number {
 	return Number(element.textContent.trim());
 }
 
-const MILESTONE_SELECTORS = ['#milestones-select-menu', '[data-testid="action-bar-item-milestones"]'];
-const PROJECT_SELECTORS = ['#project-select-menu', '[data-testid="action-bar-item-projects"]'];
+const milestonesSelectors = [
+	'#milestones-select-menu', // TODO: Drop in March 2025
+	'[data-testid="action-bar-item-milestones"]',
+];
+const projectSelectors = [
+	'#project-select-menu', // TODO: Drop in March 2025
+	'[data-testid="action-bar-item-projects"]',
+];
 
 async function hide(container: HTMLElement): Promise<void> {
 	const milestones = $('[data-selected-links^="repo_milestones"] .Counter');
 	if (milestones && getCount(milestones) === 0) {
-		$(MILESTONE_SELECTORS, container)?.remove();
+		$(milestonesSelectors, container)?.remove();
 	}
 
-	if (await hasAnyProjects.get()) {
-		return;
+	if (!await hasAnyProjects.get()) {
+		const projectsDropdown = $(projectSelectors, container);
+		projectsDropdown!.remove();
 	}
-
-	const projectsDropdown = $(PROJECT_SELECTORS, container);
-	projectsDropdown?.remove();
 }
 
 function init(signal: AbortSignal): void {
-	observe(['#js-issues-toolbar', '[data-testid="list-view-metadata"]'], hide, {signal});
+	observe([
+		'#js-issues-toolbar', // TODO: Remove after March 2025
+		'[data-testid="list-view-metadata"]',
+	], hide, {signal});
 }
 
 void features.add(import.meta.url, {
--- a/source/features/clean-conversation-filters.tsx
+++ b/source/features/clean-conversation-filters.tsx
@@ -53,6 +53,8 @@ const projectSelectors = [
 ];
 
 async function hide(container: HTMLElement): Promise<void> {
+	// TODO: Drop in March 2025
+	// The new beta view doesn't have the count selector, maybe we should do it like the projects
 	const milestones = $('[data-selected-links^="repo_milestones"] .Counter');
 	if (milestones && getCount(milestones) === 0) {
 		$(milestonesSelectors, container)?.remove();
--- a/source/features/clean-conversation-filters.tsx
+++ b/source/features/clean-conversation-filters.tsx
@@ -1,5 +1,5 @@
 import {CachedFunction} from 'webext-storage-cache';
-import {elementExists} from 'select-dom';
+import {$, elementExists} from 'select-dom';
 import elementReady from 'element-ready';
 import * as pageDetect from 'github-url-detection';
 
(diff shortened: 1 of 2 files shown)


kovsu:
It doesn't remove the `milestones` in the test URL, because it can't find the count dom.

fregante:
> It doesn't remove the `milestones` in the test URL, because it can't find the count dom.

Indeed it looks like they dropped the counter altogether.

This is a test repo that _does_ have milestones, for reference: https://github.com/tc39/ecma402/issues

source/features/clean-conversation-filters.tsx
```diff
@@ -42,44 +43,33 @@ function getCount(element: HTMLElement): number {
...
-async function hideMilestones(): Promise<void> {
-	const milestones = await elementReady('[data-selected-links^="repo_milestones"] .Counter');
-	if (getCount(milestones!) === 0) {
-		(await elementReady('[data-hotkey="m"]'))!.parentElement!.remove();
+const MILESTONE_SELECTORS = ['#milestones-select-menu', '[data-testid="action-bar-item-milestones"]'];
+const PROJECT_SELECTORS = ['#project-select-menu', '[data-testid="action-bar-item-projects"]'];
+
+async function hide(container: HTMLElement): Promise<void> {
+	const milestones = $('[data-selected-links^="repo_milestones"] .Counter');
+	if (milestones && getCount(milestones) === 0) {
```
fregante, line 51: Since the counter is missing, we should probably set up an API call like we do for the projects.

However since they have an overflow mechanism here I'm not sure it's worth it.

I'd leave a note with this information to also drop this in March 2025

source/features/clean-conversation-filters.tsx
```diff
@@ -42,44 +43,33 @@ function getCount(element: HTMLElement): number {
...
+		$(MILESTONE_SELECTORS, container)?.remove();
+	}
+
+	if (await hasAnyProjects.get()) {
+		return;
 	}
```
fregante, line 57: Easier to write:

```js
if (!await hasAnyProjects.get()) {
	$(projectDropdown, container).remove();
}
```

source/features/clean-conversation-filters.tsx
```diff
@@ -42,44 +43,33 @@ function getCount(element: HTMLElement): number {
...
-}
 
-async function hideProjects(): Promise<void> {
-	const projectsDropdown = await elementReady('[data-hotkey="p"]');
-	projectsDropdown?.parentElement!.remove();
+	const projectsDropdown = $(PROJECT_SELECTORS, container);
+	projectsDropdown?.remove();
```
fregante, line 60: Why `?`

When I encounter `| undefined` on Refined GitHub I have to make decide between:

- I know it exists in this view
- I know it might not exist in this view

In both cases I have to know.

In the first case, either I use `!` or in this case `expectElement` so that when it's missing I get a proper error.

In the latter, I leave a comment explaining in what situation it might be missing. 

When I don't know for sure, I just assume it exists and then wait for the errors to appear or be reported.

Errors help me see when selectors break; `?.` instead just hides errors

source/features/clean-conversation-filters.tsx
```diff
@@ -42,44 +43,33 @@ function getCount(element: HTMLElement): number {
...
 }
 
-// Toolbar is shown only if the repo has ever had an issue/PR
-async function hasConversations(): Promise<boolean> {
-	return Boolean(elementReady('#js-issues-toolbar', {waitForChildren: false}));
+function init(signal: AbortSignal): void {
+	observe(['#js-issues-toolbar', '[data-testid="list-view-metadata"]'], hide, {signal});
```
fregante, line 64:
```suggestion
	observe([
		'#js-issues-toolbar',  // old? add "TODO remove"
		'[data-testid="list-view-metadata"]', the new one doesn't need a comment
	], hide, {signal});
```

source/features/clean-conversation-filters.tsx
kovsu: Oh, I get it. We should get an error if it doesn't exist.

source/features/clean-conversation-filters.tsx
```diff
@@ -42,44 +43,42 @@ function getCount(element: HTMLElement): number {
 	return Number(element.textContent.trim());
 }
 
-async function hideMilestones(): Promise<void> {
-	const milestones = await elementReady('[data-selected-links^="repo_milestones"] .Counter');
-	if (getCount(milestones!) === 0) {
-		(await elementReady('[data-hotkey="m"]'))!.parentElement!.remove();
+const milestonesSelectors = [
+	'#milestones-select-menu', // TODO: Drop in March 2025
+	'[data-testid="action-bar-item-milestones"]',
```
fregante: We can't remove the milestones here, so there's no need to add this code

source/features/clean-conversation-filters.tsx
```diff
@@ -42,44 +43,42 @@ function getCount(element: HTMLElement): number {
...
+		$(milestonesSelectors, container)?.remove();
 	}
-}
 
-async function hideProjects(): Promise<void> {
-	const projectsDropdown = await elementReady('[data-hotkey="p"]');
-	projectsDropdown?.parentElement!.remove();
+	if (!await hasAnyProjects.get()) {
+		const projectsDropdown = $(projectSelectors, container);
+		projectsDropdown!.remove();
```
fregante: As a rule, if you use `$()!`, then you should use `expectElement` instead

source/features/clean-conversation-filters.tsx
```diff
@@ -42,44 +43,42 @@ function getCount(element: HTMLElement): number {
...
+	'#project-select-menu', // TODO: Drop in March 2025
+	'[data-testid="action-bar-item-projects"]',
+];
+
+async function hide(container: HTMLElement): Promise<void> {
+	// TODO: Drop in March 2025
+	// The new beta view doesn't have the count selector, maybe we should do it like the projects
+	const milestones = $('[data-selected-links^="repo_milestones"] .Counter');
+	if (milestones && getCount(milestones) === 0) {
+		$(milestonesSelectors, container)?.remove();
```
fregante: As mentioned before, this should likely not use `?` and therefore I changed it to `expectElement`

fregante (approved):

PR #7879, opened 2024-10-09
`mobile-tabs` - Improve reliability

### LINKED ISSUE #7812: `mobile-tabs` bad wrapping on 1000+ counters
### Description

This is a little difficult to fix in flexbox. If it only happens on `bugs-tab` maybe we just need `min-width`, but that’s just a last resort and not a full fix. 

IMG_5652

https://github.com/oven-sh/bun

### Extension version

24.8.20

### Browser(s) used

Safari

### Token ⚠️

fregante:
- Fixes #7812 

## Test URLs

https://github.com/oven-sh/bun

<details>
<summary>outdated screenshots</summary>

## Before

Screenshot 9

## After (unfinished)

I have to restore some extra padding

Screenshot 10

</details>


fregante:
# Bugfix before/after

Screenshot 8
Screenshot

# Base view before/after

The wide buttons were just a consequence of the flex autolayout; the desired padding has always been 10px.

Screenshot 7
Screenshot 5

build/__snapshots__/imported-features.json
```diff
@@ -89,7 +89,6 @@
 	"locked-issue",
 	"mark-merge-commits-in-list",
 	"mark-private-orgs",
-	"mobile-tabs",
```
fregante, line 92: This feature can no longer be disabled now. The delayed addition of the `rgh-mobile-tabs` attribute sometimes caused the overflow mechanism to run before the size change, leaving just 2 tabs visible.

fregante:
Found a bug, fixed. 

Screenshot 9


Here are all the possible layouts, correctly centered:

Screenshot 13
Screenshot 11
Screenshot 12

Screenshot 10

source/features/mobile-tabs.css
```diff
@@ -1,85 +1,99 @@
...
+	:root .UnderlineNav-item:has(.Counter:empty) {
+		grid-template-areas:
+			'icon icon'
+			'title title';
+
+		svg {
+			justify-self: center;
+		}
 	}
 }
```
fregante, line 91: I wish there was a way to do this natively in CSS Grid.

In short, the icon should _span_ and _center_ when the counter is missing or empty.

If you have any ideas, please try them and open a PR. I already tried all the "what about X" 😅

PR #7898, opened 2024-10-13
`select-notifications` - Support workflow failures

gioragutt:
# Naming

I've called the filter "Workflow Failures" since this is what triggers this for me, but I there might be more cases where it applies (it may be Failures in general).

So I'm open for name suggestions :)

## Test URLs

## Screenshot

* [ ] TBD: add examples when I have some in my notifications page


fregante:
Refer to https://github.com/refined-github/refined-github/issues/6980#issuecomment-1760057868


Fixed by:

- https://github.com/refined-github/refined-github/pull/7088

PR #7899, opened 2024-10-14
Improve console output for features that require a token

fregante:
- Mentioned in https://github.com/refined-github/refined-github/issues/6000#issuecomment-2185499249
- Follows https://github.com/refined-github/refined-github/pull/7798
- Related to https://github.com/refined-github/refined-github/pull/7825

This surfaces the errors earlier in features that use `observe`/`delegate`, even if the callback hasn't been called yet

More error-related work:

- https://github.com/refined-github/refined-github/issues/6096

### DIFF
--- a/build/features.test.ts
+++ b/build/features.test.ts
@@ -149,6 +149,12 @@ function validateTsx(file: FeatureFile): void {
 
 	assert(/test url/i.test(file.contents().toString()), 'Should have test URLs');
 
+	if (file.contents().includes('api.v4') && /observe\(|delegate\(/.test(String(file.contents()))) {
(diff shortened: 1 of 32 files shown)


fregante:
# Before

Screenshot 7

# After

Screenshot 8

fregante:
One thing to note is that a couple of these features did not _strictly require_ the token, but still used it in some situations. This means _all_ those features now _do_ require the token _all_ the time.

This possibility has been outlined in the options (and now [in the wiki](https://github.com/refined-github/refined-github/wiki/Security#token)) for a long time.

PR #7905, opened 2024-10-14
Fix recent token-related issues

fregante:
- Fixes https://github.com/refined-github/refined-github/issues/7897
- GHE token check broken by https://github.com/refined-github/refined-github/pull/7845
- Extracted from #7904

### DIFF
--- a/source/github-helpers/get-user-permission.ts
+++ b/source/github-helpers/get-user-permission.ts
@@ -32,7 +32,7 @@ async function getViewerPermission(): Promise<RepositoryPermission> {
 		return 'ADMIN';
 	}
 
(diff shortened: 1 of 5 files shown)


source/github-helpers/get-user-permission.ts
```diff
@@ -32,7 +32,7 @@ async function getViewerPermission(): Promise<RepositoryPermission> {
 		return 'ADMIN';
 	}
 
-	if (!hasToken()) {
+	if (!await hasToken()) {
```
fregante, line 35: Additional small bug introduced by https://github.com/refined-github/refined-github/pull/7798

Issue #7808, opened 2024-09-19
⚠️ Features broken on new commit view page (beta)


fregante:
### Description

https://github.blog/changelog/2024-09-18-new-commit-details-page-public-beta/

Some features are surely broken here, like our diff/patch links. 

### How to replicate the issue + URL

https://github.com/refined-github/refined-github/commit/2583024c7f8eb01ebd74d339849f1d551af80c70

### Extension version

24.8.20

### Browser(s) used

Safari

### Token ⚠️

- [X] I set a token in the options page. I understand that some features require a token to work.

elmaxe:
The tag preview next to commits also seems to be broken with the beta turned on

https://github.com/orgs/community/discussions/139005#discussioncomment-10694056

jdreesen:
For me, `patch-diff-links` is broken:

> 12:57:46.803 📕 24.8.30 → TypeError: commitMeta is undefined

Browser: Firefox 130

fregante:
PR welcome 🙏

jdreesen:
It's actually more than that:
![grafik](https://github.com/user-attachments/assets/b6c295b6-668a-4572-ad29-1d2cefcc80be)
I'll see if I can create a PR in the next few days.

fregante:
This is the full list of features running on single commit and commit list pages. PR Commit views are tracked separately: https://github.com/refined-github/refined-github/issues/7988

Some of these are a bit harder to verify because we don't have test links, specifically those around "comments on commits". Honestly we can just skip all features related to that view because it's rare, complex and undesirable (#7571) anyway.


For the record, here's a test link: https://github.com/refined-github/sandbox/commit/54c10fe670779eab43b6a18b7fb06e51dd7050af#r148774860

Important or easy:

- [x] `reactions-avatars`
- [x] `keyboard-navigation`
- [x] `pagination-hotkey`
- [x] `toggle-everything-with-alt`

Medium:

- [ ] `no-unnecessary-split-diff-view`
- [ ] `parse-backticks`
- [ ] `preview-hidden-comments`
- [x] `more-file-links`

Difficult:
- [ ] `show-whitespace` - see https://github.com/refined-github/refined-github/issues/7843
- [ ] `quick-comment-edit` - see #7893 
- [ ] `linkify-code` - see #6336 
- [ ] `new-or-deleted-file` - see https://github.com/refined-github/refined-github/issues/6741


No longer relevant:

- [ ] `prevent-comment-loss` 👋
[...]

Issue #7886, opened 2024-10-11
Features broken on new input fields


fregante:
### Description

<img width="499" alt="Screenshot 2" src="https://github.com/user-attachments/assets/f5db7577-4e4f-4894-aebb-8d8b638e3413">

Missing at least:

- `indent-textarea`
- `collapsible-content-button`
- `clean-rich-text-editor`
- `table-input` (already broken)
- `unfinished-comments`

### How to replicate the issue + URL

https://github.com/eslint/eslint/issues/new?template=change.yml
https://github.com/eslint/eslint/issues/18093

### Extension version

24.10.1

### Browser(s) used

Safari

### Token ⚠️

- [X] I set a token in the options page. I understand that some features require a token to work.

fregante:
I think all "reading comments", "writing comments" and "conversations" issues are currently broken on the new views.

Might be time to retire Refined GitHub 🥲🫡

fregante:
I'll check these off after verifying them

- [ ] archive-forks-link
- [ ] avoid-accidental-submissions
- [ ] bugs-tab
- [ ] ci-link
- [ ] clean-conversation-headers
- [ ] clean-conversation-sidebar
- [ ] clean-repo-tabs
- [ ] clean-rich-text-editor
- [ ] close-as-unplanned
- [ ] close-out-of-view-modals
- [ ] collapsible-content-button
- [ ] command-palette-navigation-shortcuts
- [ ] comment-excess
- [ ] comments-time-machine-links
- [ ] conversation-activity-filter
- [ ] copy-on-y
- [ ] embed-gist-inline
- [ ] emphasize-draft-pr-label
- [ ] esc-to-cancel
- [ ] esc-to-deselect-line
- [ ] expand-all-hidden-comments
- [ ] fit-textareas
- [ ] hide-diff-signs
- [ ] hide-low-quality-comments
- [ ] hide-navigation-hover-highlight
- [ ] hide-user-forks
- [ ] improve-shortcut-help
- [ ] keyboard-navigation
- [ ] linkify-branch-references
- [ ] linkify-code
- [ ] linkify-line-numbers
- [ ] linkify-user-edit-history-popup
- [ ] linkify-user-labels
- [ ] linkify-user-location
- [ ] more-dropdown-links
- [ ] netiquette
- [ ] one-key-formatting
- [ ] parse-backticks
- [ ] prevent-comment-loss
- [ ] prevent-link-loss
- [ ] preview-hidden-comments
[...]

PR #7909, opened 2024-10-15
`patch-diff-links` - Support new commit view

kovsu:
- part of #7808 

## Test URLs

- Commit:https://github.com/refined-github/refined-github/commit/132272786fdc058193e089d8c06f2a158844e101
- PR Commit: [`07ddf83` (#7751)](https://github.com/refined-github/refined-github/pull/7751/commits/07ddf838c211075701e9a681ab061a158b05ee79)

## Screenshot

image

image

https://github.com/user-attachments/assets/91528850-34c7-4f0f-be5a-5b663800ccb6

### DIFF
--- a/source/features/patch-diff-links.tsx
+++ b/source/features/patch-diff-links.tsx
@@ -15,7 +15,7 @@ async function addPatchDiffLinks(commitMeta: HTMLElement): Promise<void> {
 
 	commitMeta!.classList.remove('no-wrap'); // #5987
 	commitMeta!.prepend(
-		<span className="sha-block ml-2" data-turbo="false">
+		<span className="sha-block" data-turbo="false">
 			<a href={`${commitUrl}.patch`} className="sha color-fg-default">patch</a>
 			{' '}
 			<a href={`${commitUrl}.diff`} className="sha color-fg-default">diff</a>
@@ -26,7 +26,7 @@ async function addPatchDiffLinks(commitMeta: HTMLElement): Promise<void> {
 
 async function init(signal: AbortSignal): Promise<void> {
 	observe([
-		'.commit-meta > span:last-child', // TODO remove in March 2025
+		'.commit-meta > span:last-child', // `isPRCommit` + old `isSingleCommit`
 		'[class*="commit-header-actions"] + div pre',
 	], addPatchDiffLinks, {signal});
 }
--- a/source/features/patch-diff-links.tsx
+++ b/source/features/patch-diff-links.tsx
@@ -1,11 +1,11 @@
 import React from 'dom-chef';
-import elementReady from 'element-ready';
 import * as pageDetect from 'github-url-detection';
 
 import features from '../feature-manager.js';
 import {getCleanPathname} from '../github-helpers/index.js';
+import observe from '../helpers/selector-observer.js';
 
-async function init(): Promise<void> {
+async function addPatchDiffLinks(commitMeta: HTMLElement): Promise<void> {
 	let commitUrl = '/' + getCleanPathname();
 
 	// Avoids a redirection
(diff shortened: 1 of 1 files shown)


fregante (commented):
One thing to keep in mind is that this is still in opt-in beta, so features must still keep working on both versions, especially because more people are likely using the old version than the beta.

**If it's easier,** feel free to "duplicate" most of the code so that the old code applies to the old version and the new code applies to the new version. This is actually easier to delete next year.

source/features/patch-diff-links.tsx
```diff
@@ -1,37 +1,42 @@
...
-			<a href={`${commitUrl}.diff`} className="sha">diff</a>
+			<a href={`${commitUrl}.diff`} className="sha color-fg-default">diff</a>
 		</span>,
 	);
 }
 
+async function init(signal: AbortSignal): Promise<void> {
+	observe([
+		'.commit-meta', // TODO remove in March 2025
+		'pre:has([data-hovercard-url])',
```
fregante, line 29: This is too generic unfortunately.

Also I think we should just mark the old selector as `isPRFiles` instead of dropping it. Can you ensure it still works since you dropped `lastElementChild` on  [`ef22049` (#7909)](https://github.com/refined-github/refined-github/pull/7909/commits/ef22049cf89d2934640ba350c16d3fd49b2ad97e)? Maybe it the new selector should be `.commit-meta > span:last-child`

kovsu: > Also I think we should just mark the old selector as `isPRFiles` instead of dropping it. Can you ensure it still works since you dropped `lastElementChild` on [`ef22049` (#7909)](https://github.com/refined-github/refined-github/pull/7909/commits/ef22049cf89d2934640ba350c16d3fd49b2ad97e)? Maybe it the new selector should be `.commit-meta > span:last-child`

It's ok.

kovsu: CleanShot 2024-10-20 at 00 12 56@2x

kovsu: > This is too generic unfortunately.

how about `pre:has([aria-label^="Copy full SHA for"])` ?

CleanShot 2024-10-20 at 00 20 41@2x

fregante:
I'd like to see this:

Screenshot 1

So it's better to:

- leave the existing code unchanged
- add one more `observe('[class^="CommitHeader"] + div pre', addLinks, {signal})`

Inserted into the `pre`. The dot is:

```htm
<span class="px-2">·</span>
```

source/features/patch-diff-links.tsx
```diff
@@ -1,37 +1,42 @@
...
 			{' '}
-			<a href={`${commitUrl}.diff`} className="sha">diff</a>
+			<a href={`${commitUrl}.diff`} className="sha color-fg-default">diff</a>
 		</span>,
 	);
 }
 
+async function init(signal: AbortSignal): Promise<void> {
+	observe([
+		'.commit-meta', // TODO remove in March 2025
```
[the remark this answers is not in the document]
fregante: The comment wasn't updated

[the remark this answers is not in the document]
fregante: It's important that we only mark code for removal if we're sure that the new version completely replaces it. 

Out of the open 3 issues (#7856 #7886 #7808), only features that run on _only_ on `isSingleIssue` are guaranteed to be removed. If a feature runs on `isConversation`, then it's here to stay because PRs are not affected for example.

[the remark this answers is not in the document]
kovsu: I think updated it, but it seems I discarded it.

fregante (approved):

fregante:
I think this broke the old view:

Screenshot


while I can see it working on my older 24.10.1 install

PR #7912, opened 2024-10-15
`sticky-sidebar` - Support new issue view

ToBinio:
part of #7856

## Test URLs


source/features/sticky-sidebar.css
```diff
@@ -21,6 +22,11 @@ Exclusively use simple sums and use `0px` instead of `0`
 	--rgh-sticky-notifitication-header-height: 69px;
 }
 
+/* required to make `sticky` work for the issue Sidebar */
+div:has(> div[data-testid='issue-viewer-metadata-pane']) {
+	display: unset;
+}
```
fregante, line 28: I think it's better to set this via JS in trackSidebar. The `:has()` is heavy when its root contains a lot of elements, like in this case the generic `div`

ToBinio: since the `display: unset` doesn't seem to affect the other sidebars
I changed it so a class gets set on all the parent objects and I then just use that class.

fregante:
I made some changes to support the notification bar. But then I noticed a z-index issue with the header:

varpopup

ToBinio:
should be fixed

I also removed the additional spacing on the repo site
image

fregante (approved):

fregante:
Good catch!

If you're interested, I realized that this feature doesn't apply to discussions, so it can be added there, it it's not difficult. 

Test URL:

https://github.com/eslint/eslint/discussions/18304

Issue #7910, opened 2024-10-15
`mobile-tabs` misaligned and with bad spacing on profiles


fregante:
### Description

I did not test profiles in #7812 

<img width="677" alt="Screenshot 4" src="https://github.com/user-attachments/assets/f3c9ccd9-7f86-4778-b1d8-c2b7acfb5472">

For reference, this is what it looked like before the last PR:

<img width="676" alt="Screenshot 5" src="https://github.com/user-attachments/assets/ec1f8b74-4fa5-46fa-8084-d82a03fc5dec">

### How to replicate the issue + URL

https://github.com/refined-github

This is happening because the tab label doesn't have the same markup:

<img width="676" alt="Screenshot 5" src="https://github.com/user-attachments/assets/618f94c4-7f85-47fd-961f-ce6ff98a7be2">

So it doesn't span at the bottom and the label font size isn't scaled down.

<img width="130" alt="Screenshot 9" src="https://github.com/user-attachments/assets/aac10d45-ef4c-4d56-9a0a-a480b1fc8200">

### Extension version

24.10.14

PR #7915, opened 2024-10-16
`mobile-tabs` - Fix layout on profiles

kovsu:
Close: #7910 

## Test URLs

https://github.com/refined-github

## Screenshot

image

image

https://github.com/user-attachments/assets/b42d6736-2672-4adc-b5fe-4e5a0dbcad2d


source/features/mobile-tabs.css
```diff
@@ -67,7 +67,11 @@
 		justify-self: end;
 	}
 
-	:root .UnderlineNav-item .Counter {
+	:root .UnderlineNav-item span {
```
fregante, line 70:
```suggestion
	/* Tab label on on profiles */
	:root .UnderlineNav-item span:not([class]) {	
```

I'd also fix the extra padding and reduce the font size

kovsu:
Updated:

CleanShot 2024-10-16 at 22 32 08@2x

CleanShot 2024-10-16 at 22 32 23@2x

fregante:
It works on orgs but not on profiles 🥲 

https://github.com/fregante

Screenshot 10

kovsu:
Everthing works fine now!

Issue #7809, opened 2024-09-19
Clarify Error Popover Origin


misund:
### Problem:
![image](https://github.com/user-attachments/assets/cd175db3-4ae6-45c8-8b43-0ab7052a62d8)

When an error occurs while using the Refined GitHub browser extension, a popover notification appears in the bottom right corner of the screen. However, after using the extension for an extended period, it becomes difficult to tell whether the error message originates from GitHub natively or from the extension. This is especially true because many of Refined GitHub's features are so well integrated that they feel like part of GitHub's own functionality—an impressive testament to how seamlessly the extension enhances the user experience.

That said, this can sometimes lead to confusion when troubleshooting issues, as it's not immediately clear whether the feedback is coming from GitHub or from the extension.

### Proposed Solution:
To improve clarity and avoid confusion, it would be helpful to label the error popover as originating from Refined GitHub. Prefix error messages with something like "Refined GitHub:" or "Refined GitHub Error:" to differentiate them from native GitHub notifications.

By making it clear that the error is coming from Refined GitHub, users will be able to distinguish between GitHub's native features and the extension's enhancements, making troubleshooting easier and enhancing the overall experience.

If a prefix would feel quirky, an alternative solution could be a heading.

### Additional Context:
The fact that the features added by Refined GitHub integrate so naturally is one of the extension's greatest strengths. However, ensuring that users can easily identify the source of errors or feedback is essential to maintaining an intuitive experience.

### Example URLs

https://github.com/refined-github/refined-github/pull/7804

fregante:
I don’t think we show that message without explicit action from the user. What did you click for that to appear? Was it `discard-changes`?

You’re right though. We need to linkify “the options”  at the very least

fregante:
PR welcome to add a permanent lightweight "Refined GitHub" marker to our Toast component.

Perhaps it should just be a Refined GitHub logo/block on the left of the green/progress area.

```
┌────┌────┐──────────────────────┐
│    │    │                      │
│ RG │ ✅ │  Changes discarded.  │
│    │    │                      │
└────└────┘──────────────────────┘
```

fregante:
![token](https://github.com/user-attachments/assets/a1263d46-b3fc-4316-99f0-bc15eae8bfbc)

PR #7918, opened 2024-10-16
Animate notifications

kovsu:
<del>Close: 7809 </del>

## Screenshot

BTW, as you can we only had a fade-in effect before and I found the toast component has a `Toast--animateIn` class to use. In the end, we have animate-in and animate-out.

Before:

https://github.com/user-attachments/assets/a5faa888-d3e6-4990-8213-aa05c04d953c

After:

https://github.com/user-attachments/assets/b75c34c5-d7d7-4e41-ab27-8834935651fd

### DIFF
--- a/source/github-helpers/toast.tsx
+++ b/source/github-helpers/toast.tsx
@@ -49,6 +49,22 @@ export default async function showToast(
 		messageWrapper.textContent = message;
 	};
 
+	const finalUpdateToast = async (message: string): Promise<void> => {
+		updateToast(message);
+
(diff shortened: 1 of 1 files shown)


fregante:
I'm not loving it to be honest :(

Screenshot 12

The logo is too big and central. I also tried it on the left and I can't make it fit:

Screenshot 16


Screenshot 15

Screenshot 14

and even then, the logo isn't recognizable enough. I think this would be best:

Screenshot 17

I dropped the icon for now. We can merge it as is just for the animation improvement, but to close #7809 we'd have to implement this label style.

PR #7924, opened 2024-10-17
`small-user-avatars` - Support new issue page

ToBinio:
part of #7856

## Test URLs

https://github.com/nuxt/nuxt/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen

## Screenshot

image

> added after comment


fregante:
Please include a screenshot in your PRs so that we can verify how it looks and to keep it for future reference

PR #7930, opened 2024-10-21
`reactions-avatars` - Support new issue view

kovsu:
- part of #7856 

## Test URLs

https://togithub.com/eslint/eslint/issues/3565

## Screenshot

https://github.com/user-attachments/assets/f33689a8-e466-4f7c-a836-5cd7b841cbf5


fregante:
Don't forget to link the original issue, like:

```md
- part of #7856 
```

source/features/reactions-avatars.tsx
```diff
@@ -21,13 +21,31 @@ type Participant = {
...
+		users = button.getAttribute('aria-label')!
+			.replace(/.*including /, '')
+			.replace(/\)/, '')
+			.replace(/,? and /, ', ')
+			.replace(/, \d+ more/, '')
+			.split(', ');
+	}
+
+	// TODO: Remove after March 2024
+	if (button.nextElementSibling) {
```
fregante, line 36: `else if`

source/features/reactions-avatars.tsx
```diff
@@ -80,7 +104,10 @@ function observeCommentReactions(commentReactions: Element): void {
 }
 
 function init(signal: AbortSignal): void {
-	observe('.has-reactions .js-comment-reactions-options', observeCommentReactions, {signal});
+	observe([
+		'.has-reactions .js-comment-reactions-options', // TODO: Remove after March 2024
```
fregante, line 108: It seems that the feature supports discussions too, so shouldn't drop these. 

The second line can be marked with `// React views, isIssue`

fregante (approved):
Code looks good other than this

fregante:
This selector needs to be adjusted:

https://github.com/refined-github/refined-github/blob/f7617783376d3b31f359373cf52d56cdf0ffb2f3/source/features/reactions-avatars.css#L9-L10

Screenshot 7

fregante:
Also I'm seeing an error on discussions: https://github.com/orgs/community/discussions/5841

```
❌ Refined GitHub: reactions-avatars
errors.js:77 📕 0.0.0 → SyntaxError: Failed to execute 'querySelector' on 'Document': '[href="/    🚀
            3
" i] img' is not a valid selector.
    at $ (select-dom.js:8:38)
    at getUserAvatar (get-user-avatar.js:8:25)
    at getParticipants (reactions-avatars.js:53:20)
    at reactions-avatars.js:86:19
    at Array.map (<anonymous>)
    at showAvatarsOn (reactions-avatars.js:86:5)
    at IntersectionObserver.rootMargin (reactions-avatars.js:65:4)
```

Before this PR, the error was there, but different:

```
❌ Refined GitHub: reactions-avatars
errors.js:77 📕 0.0.0 → TypeError: Cannot read properties of null (reading 'textContent')
    at getParticipants (reactions-avatars.js:24:4)
    at reactions-avatars.js:64:19
    at Array.map (<anonymous>)
    at showAvatarsOn (reactions-avatars.js:64:5)
    at IntersectionObserver.rootMargin (reactions-avatars.js:49:4)
```

fregante:
Just to be clear, the new error is specifically because `getParticipants` is returning bad data, so it ends up in the selector

kovsu:
image

image

fregante:
Sorry I had pointed you in the wrong direction, it was not `:last-of-type` but the regular `margin-right` that needed to be increased.

Screenshot 19

fregante (approved):

source/features/reactions-avatars.tsx
```diff
@@ -21,13 +21,33 @@ type Participant = {
...
+			.replace(/\)/, '')
+			.replace(/,? and /, ', ')
+			.replace(/, \d+ more/, '')
+			.split(', ');
+	} else if (button.nextElementSibling) {
+		// TODO: Remove after March 2024
+		// The list of people who commented is in an adjacent `<tool-tip>` element #5698
+		const tagName = button.nextElementSibling.tagName;
+		if (tagName !== 'TOOL-TIP')
+			return [];
```
fregante, line 38: This needs to be cleaned up a bit.

- In what case is it not `tool-tip`? Is the selector somewhere selecting the wrong elements?
- Does the current version work correctly on discussions?

At a minimum the code should be changed to:

```suggestion
		// TODO: Remove after March 2024
		// The list of people who commented is in an adjacent `<tool-tip>` element #5698
	} else if (button.nextElementSibling?.tagName === 'TOOL-TIP') {
```

fregante: How about the questions?

source/github-helpers/get-user-avatar.ts
```diff
@@ -4,6 +4,11 @@ import * as pageDetect from 'github-url-detection';
 export default function getUserAvatar(username: string, size: number): string | void {
 	const cleanName = username.replace('[bot]', '');
 
+	if (!/[^\w-]/.test(cleanName)) {
```
kovsu, line 7: Maybe here is `!/[\w-]/` or `/[^\w-]/`

source/features/reactions-avatars.tsx
kovsu: > Does the current version work correctly on discussions?

It works on discusssions.

CleanShot 2024-10-24 at 08 39 35@2x

kovsu: > In what case is it not tool-tip? Is the selector somewhere selecting the wrong elements?

As you can see, we get the button then we get its `nextElementSibling`.

https://github.com/refined-github/refined-github/blob/704e344b04e8e758637cdbc74a333f94169e9b1a/source/features/reactions-avatars.tsx#L23

I find the reason it's the `button.nextElementSibling` is a button, but the comment says `The list of people who commented is in an adjacent <tool-tip> element`. So I think we should to determine.

image

If `button.nextElementSibling` is a button, this is what's happened.

```
❌ Refined GitHub: reactions-avatars
errors.js:77 📕 0.0.0 → SyntaxError: Failed to execute 'querySelector' on 'Document': '[href="/    🚀
            3
" i] img' is not a valid selector.
    at $ (select-dom.js:8:38)
    at getUserAvatar (get-user-avatar.js:8:25)
    at getParticipants (reactions-avatars.js:53:20)
    at reactions-avatars.js:86:19
    at Array.map (<anonymous>)
    at showAvatarsOn (reactions-avatars.js:86:5)
    at IntersectionObserver.rootMargin (reactions-avatars.js:65:4)
```

kovsu:
```tsx
if (button.getAttribute('disabled') === 'disabled')
		return [];
```

Maybe we just should add this.

fregante: 👆 exactly. If the tooltip doesn't exist when it's disabled, then the observed selector should just be `:not([disabled])`, getParticipants shouldn't be called at all

source/github-helpers/get-user-avatar.ts
fregante: The latter

fregante:
Adjusted, restested, merging…

fregante:
This would also be good to address, it's not in the original list of broken features: https://github.com/refined-github/refined-github/blob/db5e00462bdbe398512f223cabb41961e81fc559/source/features/center-reactions-popup.css#L4

source/features/reactions-avatars.tsx
```diff
@@ -58,18 +69,24 @@ const viewportObserver = new IntersectionObserver(changes => {
...
+	const reactionTypes = $$([
+		'.social-reaction-summary-item',
+		'[class^="Tooltip__TooltipBase"]',	// React views, isIssue
+	], commentReactions).length;
 	const avatarLimit = arbitraryAvatarLimit - (reactionTypes * approximateHeaderLength);
 
 	const participantByReaction
-		= $$(':scope > button.social-reaction-summary-item', commentReactions)
+		= $$([
+			':scope > button.social-reaction-summary-item:enabled',
```
fregante, line 80: `:enabled` broke it on locked issues:

- https://github.com/refined-github/refined-github/issues/8081

PR #7940, opened 2024-10-23
`ci-link` - Add explanatory tooltip

### LINKED ISSUE #7938: `ci-link` status icon needs a tooltip to explain what it is
### Description

I was seeing a mysterious "X" at the end of repository names.  It looked exactly like the thing you would click on to delete something, or close a popup window.  When I moved the mouse over that icon, I saw no explanation text or tooltip.

It turns out that it was the "ci-link" feature, which tells you the status of continuous integration.

But I really would have preferred it if moving the mouse over the icon would have told me that it was related to continuous integration, and not just been a mysterious "X" symbol without any context.  Maybe a message like "Continuous Integration: Failed" or something.

### Example URLs

not applicable

fregante:
- Closes #7938 

## Test URLs

Here

## Screenshot

Screenshot 20

### DIFF
--- a/source/features/ci-link.tsx
+++ b/source/features/ci-link.tsx
@@ -29,7 +29,7 @@ async function add(anchor: HTMLElement): Promise<void> {
 
 	const endpoint = buildRepoURL('commits/checks-statuses-rollups');
 	anchor.parentElement!.append(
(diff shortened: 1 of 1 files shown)


fregante:
cc @Dwedit

PR #7946, opened 2024-10-23
`pr-branch-auto-delete` - Preserve more `release` branches

berniedurfee-renaissance:
We use a single release branch named `release` that I'd like to protect from auto-deletion.

The current functionality is to protect branches named `release/*`. This PR would change that to protect branches named `release*`.


fregante (approved):
Reasonable. Keep in mind that you should just add branch protections if you don't want the branch to be deleted.

PR #7947, opened 2024-10-24
Log broken selectors in the observer

fregante:
- Closes https://github.com/refined-github/refined-github/issues/6096

It's quite noisy because many of these features are "optional", for example on issues often you won't be able to edit _any_ comments, but that doesn't mean that the feature is broken.

It might be a way to detect broken features in 7856 #7886 #7808, but it's probably just noise.

I added it temporarily behind the `logging` flag, let's see how it works.

## Screenshot

Screenshot 23

### DIFF
--- a/source/helpers/event-listener-loop.test.ts
+++ b/source/helpers/event-listener-loop.test.ts
@@ -1,0 +1,74 @@
+import {describe, it, expect} from 'vitest';
+
+import createEventIterator from './event-listener-loop.js';
+
+describe('createEventIterator', () => {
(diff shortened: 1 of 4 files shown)


source/helpers/event-listener-loop.test.ts
```diff
@@ -0,0 +1,74 @@
+import {describe, it, expect} from 'vitest';
+
+import createEventIterator from './event-listener-loop.js';
```
fregante, line 3: Unused for now, keeping it in in case I/Sindre want to use it in the future

PR #7950, opened 2024-10-24
Use nested CSS

fregante:
- Picks up https://github.com/refined-github/refined-github/pull/7553
- Unblocked by #7949


fregante (commented):
All verified in Safari, never crashed

PR #7955, opened 2024-10-25
`prevent-link-loss` - Support new input field

kovsu:
- part of #7886

## Test URLs

## Screenshot

https://github.com/user-attachments/assets/32eac77e-e0c2-4878-8653-7d72a2432e71


fregante (approved):
I think that mx-2 needs to be removed in this case, in order to fix the alignment.

kovsu:
👇
image

image

fregante (approved):

source/features/prevent-link-loss.tsx
```diff
@@ -20,16 +20,25 @@ import createBanner from '../github-helpers/banner.js';
...
+
 	replaceFieldText(field, prCommitUrlRegex, preventPrCommitLinkLoss);
 	replaceFieldText(field, prCompareUrlRegex, preventPrCompareLinkLoss);
 	replaceFieldText(field, discussionUrlRegex, preventDiscussionLinkLoss);
 	fixButton.closest('.flash')!.remove();
 }
 
-function getUI(field: HTMLTextAreaElement): HTMLElement {
-	return $('.rgh-prevent-link-loss-container', field.form!) ?? (createBanner({
+function getUI(field: HTMLElement): HTMLElement {
```
fregante, line 40: It's not a field anymore, it can be `container`

source/features/prevent-link-loss.tsx
```diff
@@ -40,7 +49,7 @@ function getUI(field: HTMLTextAreaElement): HTMLElement {
 				{' by GitHub.'}
 			</>
 		),
-		classes: ['rgh-prevent-link-loss-container', 'flash-warn', 'my-2', 'mx-2'],
+		classes: ['rgh-prevent-link-loss-container', 'flash-warn', 'my-2', field.tagName === 'FORM' ? 'mx-2' : ''],
```
fregante, line 52:
```suggestion
		classes: [
			'rgh-prevent-link-loss-container',
			'flash-warn',
			'my-2',
			field.tagName === 'FORM' ? 'mx-2' : '',
		],
```

source/features/prevent-link-loss.tsx
```diff
@@ -55,7 +64,13 @@ function isVulnerableToLinkLoss(value: string): boolean {
 
 function updateUI({delegateTarget: field}: DelegateEvent<Event, HTMLTextAreaElement>): void {
 	if (isVulnerableToLinkLoss(field.value)) {
-		$('file-attachment .js-write-bucket', field.form!).append(getUI(field));
+		if (field.form) {
+			$('file-attachment .js-write-bucket', field.form).append(getUI(field.form));
+		} else {
+			// New commit input field
```
fregante, line 70: No "new" comments

source/features/prevent-link-loss.tsx
```diff
@@ -66,8 +81,14 @@ const updateUIDebounced = debounceFn(updateUI, {
 });
 
 function init(signal: AbortSignal): void {
-	delegate('textarea.js-comment-field', 'input', updateUIDebounced, {signal});
-	delegate('textarea.js-comment-field', 'focusin', updateUI, {signal});
+	delegate([
+		'textarea.js-comment-field',
+		'textarea[aria-labelledby="comment-composer-heading"]', // New commit input field
+	], 'input', updateUIDebounced, {signal});
```
fregante, line 87: Let's move this array out so it can be reused: `fieldSelector`

source/features/prevent-link-loss.tsx
```diff
@@ -20,16 +20,25 @@ import createBanner from '../github-helpers/banner.js';
...
+
+	if (fixButton.form) {
+		/* There's only one rich-text editor even when multiple fields are visible; the class targets it #4678 */
+		field = fixButton.form!.querySelector('textarea.js-comment-field')!;
+	} else {
+		// New commit input field
+		field = fixButton.closest('[data-testid="markdown-editor-comment-composer"]')!
+			.querySelector('textarea[aria-labelledby="comment-composer-heading"]')!;
+	}
+
```
fregante, line 33: You can use `fixButton.closest(['form', 'test id selector']).querySelector( fieldSelector)`. The form/testid array should also be on multiple lines and commented as usual

fregante (changes requested):

source/features/prevent-link-loss.tsx
kovsu: Why doesn't eslint do that? 😿

fregante: We don't run Prettier on JS code. I don't like how it affects all code even when it's not needed. _Maybe one day_

kovsu:
Update the screenshot. It all works fine.

source/features/prevent-link-loss.tsx
```diff
@@ -17,19 +17,27 @@ import {
 } from '../github-helpers/prevent-link-loss.js';
 import createBanner from '../github-helpers/banner.js';
 
+const fieldSelector = [
+	'textarea.js-comment-field',
+	'textarea[aria-labelledby="comment-composer-heading"]', // Commit input field
```
fregante, line 22: Commit? Or comment? Maybe it can just be 
```suggestion
	'textarea[aria-labelledby="comment-composer-heading"]', // React view
```

source/features/prevent-link-loss.tsx
```diff
@@ -17,19 +17,27 @@ import {
...
+];
```
fregante, line 23: I think if you this you won't need `as HTMLTextAreaElement` later
```suggestion
] as const;
```

fregante (approved):

source/features/prevent-link-loss.tsx
kovsu: If we use `as const`, this type may not be `strings[]`

kovsu: https://github.com/refined-github/refined-github/blob/6c3e648a8fe312f5c45da960ecc025a403a2ff49/source/github-events/on-commit-title-update.ts#L3-L6

Here hasn't use it also.

fregante: It does now :) https://github.com/fregante/delegate-it/releases/tag/v6.2.1

fregante:
Releasing a new version in a minute

PR #7958, opened 2024-10-25
`embed-gist-inline` - Support new view

kovsu:
- part of #7856 

## Test URLs

## Screenshot

https://github.com/user-attachments/assets/2eb5d186-b203-4885-850a-b0a7bb0e5735

### DIFF
--- a/source/github-helpers/selectors.ts
+++ b/source/github-helpers/selectors.ts
@@ -9,12 +9,20 @@ export const repoUnderlineNavUl_ = [
 	[1, 'https://github.com/refined-github/refined-github/releases'],
 ] satisfies UrlMatch[];
 
-export const standaloneGistLinkInMarkdown = css`
-	.js-comment-body p a:only-child:is(
-		[href^="https://gist.github.com/"],
-		[href^="${location.origin}/gist/"]
-	)
-` as 'a'; // TODO: Drop after https://github.com/fregante/code-tag/issues/12
+export const standaloneGistLinkInMarkdown = [
+	css`
+		.js-comment-body p a:only-child:is(
+			[href^="https://gist.github.com/"],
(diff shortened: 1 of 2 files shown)


source/github-helpers/selectors.ts
```diff
@@ -9,12 +9,20 @@ export const repoUnderlineNavUl_ = [
...
 
-export const standaloneGistLinkInMarkdown = css`
-	.js-comment-body p a:only-child:is(
-		[href^="https://gist.github.com/"],
-		[href^="${location.origin}/gist/"]
-	)
-` as 'a'; // TODO: Drop after https://github.com/fregante/code-tag/issues/12
+export const standaloneGistLinkInMarkdown = [
+	css`
+		.js-comment-body p a:only-child:is(
```
fregante, line 14: The only difference here is the first class. You can use `:is(.js-comment-body, .react-issue-comment)` instead of duplicating 5 lines

fregante (changes requested):

source/github-helpers/selectors.ts
```diff
@@ -10,7 +10,7 @@ export const repoUnderlineNavUl_ = [
 ] satisfies UrlMatch[];
 
 export const standaloneGistLinkInMarkdown = css`
-	.js-comment-body p a:only-child:is(
+	:is(.js-comment-body, .react-issue-comment) p a:only-child:is(
```
fregante, line 13: Tip: If you see simple class changes like these, it might be good to search them on the repo and fix multiple features at once. For example there's `rgh-linkify-features` which will need this change eventually.

kovsu: 👍 Other pr.

fregante:
I love it when they just need a selector updated 😍

PR #7982, opened 2024-10-29
`show-names` - Support new issue view

ToBinio:
part of #7856

## Test URLs

## Screenshot

image

### DIFF
--- a/source/github-helpers/selectors.ts
+++ b/source/github-helpers/selectors.ts
@@ -192,27 +192,17 @@ export const botLinksPrSelectors = [
 export const usernameLinksSelector = [
 	// `a` selector needed to skip commits by non-GitHub users
 	// # targets mannequins #6504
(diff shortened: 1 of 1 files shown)


source/github-helpers/selectors.ts
```diff
@@ -205,6 +205,16 @@ export const usernameLinksSelector = [
...
 	// https://github.com/refined-github/refined-github/issues/7232#issuecomment-1910803157
 
+	// On new issue view
+	`.react-issue-comment
+	a[data-testid='avatar-link']:not(
+		[href='#'],
+		[href*='/apps/'],
+		[href*='/marketplace/'],
+		[data-hovercard-type='organization'],
+		[show_full_name='true']
```
ToBinio, line 215: I wasn't able to find anywhere to check if these still apply.

fregante: Yes they most likely apply except for the last one. We can just keep it because we did remove it too early last time.

Issue #7954, opened 2024-10-24
`tags-on-commits-list`: Expected element not found because the base is specified but null


josephharrington:
### Description

```
ElementNotFoundError: Expected element not found because the base is specified but null
    at expectElement (chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/assets/npm/select-dom.js:15:15)
    at getCommitHash (chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/assets/features/mark-merge-commits-in-list.js:40:15)
    at init (chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/assets/features/tags-on-commits-list.js:90:27)
    at async chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/assets/feature-manager.js:119:19
    at async Promise.all (index 0)
    at async asyncForEach (chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/assets/helpers/async-for-each.js:6:2)
    at async runFeature (chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/assets/feature-manager.js:118:3)
    at async setupPageLoad (chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/assets/feature-manager.js:133:3)
```

### How to replicate the issue + URL

Public GitHub URL: https://github.com/microsoft/vscode/commits/release/1.25/
Note issue does not occur on this public GitHub though -- tags will show correctly at this URL. The issue seems to be specific to GitHub Enterprise. I observed this issue on GitHub Enterprise Server 3.13.4.

The problem is that the tags are not being shown next to the commits. From debugging it appears to be a difference between GH and GHE in how the page is loaded. I set a breakpoint in [tags-on-commits-list.js](https://github.com/refined-github/refined-github/blob/main/source/features/tags-on-commits-list.tsx#L91-L93) and I see that `commitsOnPage` is an empty list because the commit list has actually not yet been rendered on the page. This seems to cause the next line to throw an error.

<img width="891" alt="Screenshot 2024-10-24 at 12 38 45" src="https://github.com/user-attachments/assets/0454277a-33c5-4504-b46e-6d0d718b9c0a">

### Extension version

24.10.14

### Browser(s) used

Chrome

### Token ⚠️

- [X] I set a token in the options page. I understand that some features require a token to work.

fregante:
Parent issue:

- #7808 


cc @kovsu since you fixed this last summer

[...]

PR #7987, opened 2024-10-31
`tags-on-commits-list` - Use observe

kovsu:
Closes: #7954

## Test URLs

## Screenshot

image

### DIFF
--- a/source/features/tags-on-commits-list.tsx
+++ b/source/features/tags-on-commits-list.tsx
@@ -1,10 +1,10 @@
 import React from 'dom-chef';
 import cache from 'webext-storage-cache/legacy.js';
-import {$$} from 'select-dom';
 import {$} from 'select-dom/strict.js';
 
 import TagIcon from 'octicons-plain-react/Tag';
 import * as pageDetect from 'github-url-detection';
+import batchedFunction from 'batched-function';
 
 import features from '../feature-manager.js';
 import api from '../github-helpers/api.js';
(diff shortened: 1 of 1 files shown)


kovsu:
> Caching should probably also be applied per commit (difficult) or just removed. https://github.com/refined-github/refined-github/issues/7954#issuecomment-2448905214

What does this means here? Isn't per commit has been cached now?

source/features/tags-on-commits-list.tsx
```diff
@@ -141,12 +139,15 @@ async function init(): Promise<void | false> {
 	await cache.set(cacheKey, cached, {days: 1});
```
fregante, line 139: To avoid race conditions, we need to split this up to cache each commit individually calling `cache.set` as needed, for example replacing `commitsWithNoTags.push(targetCommit);` with `void cache.set('${cacheKey}:${commit}', [])`

fregante: > What does this means here? Isn't per commit has been cached now?

Yeah you're right, it's just that due to the querying it has limited utility.

source/features/tags-on-commits-list.tsx
```diff
@@ -1,17 +1,18 @@
...
 import cache from 'webext-storage-cache/legacy.js';
-import {$$} from 'select-dom';
 import {$} from 'select-dom/strict.js';
 
 import TagIcon from 'octicons-plain-react/Tag';
 import * as pageDetect from 'github-url-detection';
+import batchedFunction from 'batched-function';
 
 import features from '../feature-manager.js';
 import api from '../github-helpers/api.js';
```
fregante, line 10: The api/query setup is complex here, because it relies on the "last commit on the page", which now is not what we expect. 

This is why I never ported this feature to the observer. I wonder if we should just fix the original bug via polling instead of changing this whole file. I never reviewed/touched this feature.

I'd probably be more comfortable with a rewrite from scratch than trying to iterate on it, but since this is a GHE-only but maybe we should just use polling + `$$` somehow. Any ideas?

PR #7990, opened 2024-11-01
Autofix: The pull request description will be empty if we are going to merge a approved PR.

lattaai1:
I have identified the issue in the `clear-pr-merge-commit-message.tsx` file. The problem was that the feature was being unloaded after the first run, preventing it from clearing the commit message field on subsequent openings. I've modified the code to remove the unloading logic, allowing the feature to run every time the merge commit message field is displayed. 
> [!CAUTION]  
> Disclaimer: This fix was created by Latta AI and you should never merge before you check the correctness of generated code!

---
This bug was fixed for free by Latta AI - https://latta.ai/ourmission

### DIFF
--- a/source/features/clear-pr-merge-commit-message.tsx
+++ b/source/features/clear-pr-merge-commit-message.tsx
@@ -13,9 +13,6 @@ import {expectToken} from '../github-helpers/github-token.js';
 const isPrAgainstDefaultBranch = async (): Promise<boolean> => getBranches().base.branch === await getDefaultBranch();
 
 async function clear(messageField: HTMLTextAreaElement): Promise<void | false> {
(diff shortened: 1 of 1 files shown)


fregante:
Thanks, but this is intentional: https://github.com/refined-github/refined-github/pull/6330

so the fix isn't enough

yuqi1129:
> Thanks, but this is intentional: #6330
> 
> so the fix isn't enough

Thank you for your response. So will the problem not be resolved? The behavior from the previous version is more suitable for me. In the latest version, I have to manually write the commit log. It's quite inconvenient.

fregante:
You can disable `clear-pr-merge-commit-message`

Issue #7988, opened 2024-10-31
⚠️ Features broken on new PR commit view page (beta)


fregante:
### Description

There's some overlap with https://github.com/refined-github/refined-github/issues/7808 but this also has more PR features to take care of.

GitHub is also reportedly working on a PR page redesign so there's some overlap with that one too.

<img width="626" alt="Screenshot 4" src="https://github.com/user-attachments/assets/87582103-7749-41ce-883c-fbdec7c9968c">

### How to replicate the issue + URL

- https://github.com/refined-github/refined-github/pull/7893/commits
- [`32352f7` (#7981)](https://github.com/refined-github/refined-github/pull/7981/commits/32352f73b951779f7e07e15be5fecc4ebe62e664)

### Likely affected features

These contain the explicit `isPRCommit` check, it's not a complete list:

- [x] `patch-diff-links`
- [x] `pagination-hotkey`
- [x] `previous-next-commit-buttons`
- [ ] `action-pr-link`
- [x] `useful-not-found-page`
- [x] `linkify-commit-sha`
- [x] `pr-commit-lines-changed`

These have `isCommitList`

- [x] `conventional-commits`
- [x] `dim-bots`
- [x] `easy-toggle-commit-messages`
- [x] `mark-merge-commits-in-list`
- [ ] `parse-backticks`
- [x] `rgh-dim-commits`
- [ ] `rgh-linkify-features`
- [x] `toggle-everything-with-alt`

kovsu:
> https://github.com/refined-github/refined-github/pull/7987/commits

I think this is test url [`32352f7` (#7981)](https://github.com/refined-github/refined-github/pull/7981/commits/32352f73b951779f7e07e15be5fecc4ebe62e664). Like [patch-diff-links](https://github.com/refined-github/refined-github/blob/main/source/features/patch-diff-links.tsx) hasn't applied in  https://github.com/refined-github/refined-github/pull/7987/commits

fregante:
They've both been updated. For example the `clean-conversation-headers` applies to all PR pages as well. I added your test URL above

fregante:
There are a couple that I haven't checked, but I think one is broken: `action-pr-link`

1. Open https://github.com/download-directory/download-directory.github.io/pull/146/commits
2. Click on a commit's checks link
3. Click "details"
4. URL does not have `?pr=146`

[...]

PR #7981, opened 2024-10-29
`shorten-links` - Support new issue view

kovsu:
- For #7856 

## Test URLs

https://github.com/download-directory/download-directory.github.io/issues/141#issuecomment-2443099561

## Screenshot

| Before | After |
| ------ | ------ |
| image | image |
| image | image |

PR #8001, opened 2024-11-03
`clean-converstion-headers` - Support new PR view

kovsu:
Part of #7988

## Test URLs

- https://github.com/refined-github/refined-github/pull/7987/commits
- [`32352f7` (#7981)](https://github.com/refined-github/refined-github/pull/7981/commits/32352f73b951779f7e07e15be5fecc4ebe62e664)

## Screenshot

CleanShot 2024-11-03 at 19 15 26@2x

### DIFF
--- a/source/features/clean-conversation-headers.css
+++ b/source/features/clean-conversation-headers.css
@@ -42,12 +42,12 @@ a.rgh-clean-conversation-headers-non-default-branch {
 /* Removes on PRs: [octocat] merged 1 commit into master from feature */
 .rgh-clean-conversation-headers.rgh-clean-conversation-headers-hide-author .author + .Label, /* #5832 */
 .rgh-clean-conversation-headers.rgh-clean-conversation-headers-hide-author .author,
-.rgh-clean-conversation-headers.rgh-clean-conversation-headers-hide-author a[data-hovercard-url]{
+.rgh-clean-conversation-headers.rgh-clean-conversation-headers-hide-author a[data-hovercard-url] {
 	display: none;
 }
 
 .rgh-clean-conversation-headers .commit-ref,
-.rgh-clean-conversation-headers [class^="BranchName"] {
+.rgh-clean-conversation-headers [class^='BranchName'] {
 	font-size: 12px;
 }
 
--- a/source/features/clean-conversation-headers.css
+++ b/source/features/clean-conversation-headers.css
@@ -15,7 +15,7 @@ a.rgh-clean-conversation-headers-non-default-branch {
 
 /* Removes: every text (but not icons) */
 .rgh-clean-conversation-headers {
-	font-size: 0;
+	font-size: 0 !important;
 }
 
 /* Shows on issues: [octocat] opened this issue on [1 Jan] · 1 comments */
@@ -63,8 +63,4 @@ a.rgh-clean-conversation-headers-non-default-branch {
 /* Fixes vertical alignment of base branch dropdown #5598 */
 .rgh-clean-conversation-headers .commit-ref-dropdown {
 	margin-top: -2px !important;
-}
-
-.d-flex:has(.rgh-clean-conversation-headers) {
-	align-items: center !important;
 }
--- a/source/features/clean-conversation-headers.css
+++ b/source/features/clean-conversation-headers.css
@@ -5,7 +5,8 @@
 	);
 }
 
-.rgh-clean-conversation-headers-non-default-branch a {
+.rgh-clean-conversation-headers-non-default-branch a,
+a.rgh-clean-conversation-headers-non-default-branch {
 	color: var(
 		--control-checked-fgColor-rest,
 		var(--color-state-hover-primary-text, #fff)
@@ -40,11 +41,13 @@
 /* Removes on issues: [octocat] opened this issue on 1 Jan · 1 comments */
 /* Removes on PRs: [octocat] merged 1 commit into master from feature */
 .rgh-clean-conversation-headers.rgh-clean-conversation-headers-hide-author .author + .Label, /* #5832 */
-.rgh-clean-conversation-headers.rgh-clean-conversation-headers-hide-author .author {
+.rgh-clean-conversation-headers.rgh-clean-conversation-headers-hide-author .author,
+.rgh-clean-conversation-headers.rgh-clean-conversation-headers-hide-author a[data-hovercard-url]{
 	display: none;
 }
 
-.rgh-clean-conversation-headers .commit-ref {
+.rgh-clean-conversation-headers .commit-ref,
+.rgh-clean-conversation-headers [class^="BranchName"] {
 	font-size: 12px;
 }
 
--- a/source/features/clean-conversation-headers.css
+++ b/source/features/clean-conversation-headers.css
@@ -64,3 +64,7 @@ a.rgh-clean-conversation-headers-non-default-branch {
 .rgh-clean-conversation-headers .commit-ref-dropdown {
 	margin-top: -2px !important;
 }
+
+.d-flex:has(.rgh-clean-conversation-headers) {
+	align-items: center !important;
+}
--- a/source/features/clean-conversation-headers.tsx
+++ b/source/features/clean-conversation-headers.tsx
@@ -11,6 +11,7 @@ import getDefaultBranch from '../github-helpers/get-default-branch.js';
 import observe from '../helpers/selector-observer.js';
 import {expectToken} from '../github-helpers/github-token.js';
 import {parseReferenceRaw} from '../github-helpers/pr-branches.js';
+import {assertNodeContent} from '../helpers/dom-utils.js';
 
 async function cleanIssueHeader(byline: HTMLElement): Promise<void> {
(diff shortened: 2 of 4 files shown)


kovsu:
| | Conversations | Commits | 
| --------- | --------- | --------- | 
| Open PR (default branch) | image | image |
| Open PR (non-default branch) | image | image |
| Merged PR (same author) | image | image |
|  Merged PR (different author) | image | image |
| Merged PR (different author + first published tag) | image | image |
| Closed PR | image | image |

source/features/clean-conversation-headers.tsx
```diff
@@ -28,24 +28,30 @@ async function cleanPrHeader(byline: HTMLElement): Promise<void> {
...
 	if (baseBranchDropdown) {
 		baseBranchDropdown.after(<span>{arrowIcon}</span>); // #5598
 	} else {
-		base.nextElementSibling!.replaceChildren(arrowIcon);
+		base.after(<span>{arrowIcon}</span>);
 	}
 
-	const baseBranch = base.title.split(':')[1];
+	const baseBranch = base.textContent.split(':')[1];
 	const wasDefaultBranch = pageDetect.isClosedPR() && baseBranch === 'master';
```
kovsu, line 55: Actually I don't konw what's it means here. But it caused below works wrong in PRCommitList.

- Merged PR (same author): https://github.com/sindresorhus/refined-github/pull/3402
- Merged PR (different author): https://github.com/parcel-bundler/parcel/pull/78
- Merged PR (different author + first published tag): https://github.com/sindresorhus/refined-github/pull/3227

kovsu:
`cross-deleted-pr-branches` seems also need to fix.

source/features/clean-conversation-headers.tsx
fregante: When the default branch is changed (as is common from `master` to `main`), every old PR appears as "merged into non-default" and therefore highlighted. Here we added an exception specifically for closed/old PRs so that the feature is "silent" in this very common scenario. 

It might be worth inlining this very comment in the code.

fregante: > But it caused some problems

What problem is it?

source/features/clean-conversation-headers.css
```diff
@@ -40,11 +41,13 @@
...
+.rgh-clean-conversation-headers.rgh-clean-conversation-headers-hide-author .author,
+.rgh-clean-conversation-headers.rgh-clean-conversation-headers-hide-author a[data-hovercard-url] {
 	display: none;
 }
 
-.rgh-clean-conversation-headers .commit-ref {
+.rgh-clean-conversation-headers .commit-ref,
+.rgh-clean-conversation-headers [class^='BranchName'] {
 	font-size: 12px;
 }
```
fregante, line 52: Might be worth nesting selectors here.

```css
.rgh-clean-conversation-headers {
	.rgh-clean-conversation-headers-hide-author {
		.author+.Label,
		.author {
			
		}
	}
	.commit-ref,
	[class^='BranchName'] {
		
	}
}
```

source/features/clean-conversation-headers.tsx
```diff
@@ -23,29 +23,41 @@ async function cleanIssueHeader(byline: HTMLElement): Promise<void> {
...
+	], byline);
 	const baseBranchDropdown = $optional('.commit-ref-dropdown', byline);
 
 	// Shows on PRs: main [←] feature
 	const arrowIcon = <ArrowLeftIcon className="v-align-middle mx-1" />;
 	if (baseBranchDropdown) {
 		baseBranchDropdown.after(<span>{arrowIcon}</span>); // #5598
 	} else {
-		base.nextElementSibling!.replaceChildren(arrowIcon);
+		base.after(<span>{arrowIcon}</span>);
```
fregante, line 56: This changes the behavior for the old version though. `replaceChildren` presumably also removes some elements

source/features/clean-conversation-headers.tsx
```diff
@@ -23,29 +23,41 @@ async function cleanIssueHeader(byline: HTMLElement): Promise<void> {
...
 	}
 
-	const baseBranch = base.title.split(':')[1];
+	const baseBranchContent = base.textContent.split(':');
```
fregante, line 59: This change needs to be reviewed because this part of the UI is rather tricky, especially when introducing long names, escaping, cross-repo PRs, etc. However we already have a function that gets the base and head branch information, so we should probably use that one (as well as ensure that it works as expected on this new view, so that all remaining features benefit from it)

source/features/clean-conversation-headers.tsx
```diff
@@ -60,6 +72,7 @@ async function init(signal: AbortSignal): Promise<void> {
 	observe([
 		'.gh-header-meta > .flex-auto', // Real
 		'.rgh-conversation-activity-filter', // Helper in case it runs first and breaks the `>` selector, because it wraps the .flex-auto element
+		'[class^="StateLabel"] + div > span',
```
fregante, line 75: If this is so generic, we might want to use `:first-child` on `span` to avoid ever matching elements that we don't intend to match.

source/features/clean-conversation-headers.tsx
```diff
@@ -23,29 +23,41 @@ async function cleanIssueHeader(byline: HTMLElement): Promise<void> {
 async function cleanPrHeader(byline: HTMLElement): Promise<void> {
 	byline.classList.add('rgh-clean-conversation-headers');
 
+	const authorSelector = [
```
fregante, line 26:
```js
	const prCreatorSelector = [
```

source/features/clean-conversation-headers.tsx
kovsu: The baseBranch wich is default branch has been marked as `non-default-branch`

source/features/clean-conversation-headers.css
```diff
@@ -61,3 +64,7 @@
 .rgh-clean-conversation-headers .commit-ref-dropdown {
 	margin-top: -2px !important;
 }
+
+.d-flex:has(.rgh-clean-conversation-headers) {
```
fregante, line 68: This is very broad, it can target any flex element wrapping the whole page. Maybe you want `.d-flex:has(> .rgh-clean-conversation-headers)` and/or even better prepend some other class like `.whatever .d-flex:has(.rgh-clean-conversation-headers)`

But actually it's likely easier to just add `.flex-items-center` to the element, like:

```js
byline.closest('.d-flex')!.add('flex-items-center')
```

:has() can be fatal to performance so it needs to be used cautiously:

- https://github.com/refined-github/refined-github/issues/7331#issuecomment-2120464245

source/features/clean-conversation-headers.tsx
kovsu: It's a text node. I try it but it not works.

source/features/clean-conversation-headers.tsx
fregante: Then I think the logic elsewhere is wrong. is `wasDefaultBranch` is true, then the result should be the opposite: `master` should be marked as default even if it currently isn't

source/features/clean-conversation-headers.css
fregante: Not necessary though, it can be done separately since it affects the whole file.

kovsu:
BTW, it works well on sticky.
before:

image

after:

image

source/features/clean-conversation-headers.tsx
kovsu: `pageDetect.isClosedPR` in PR conversation is `true`, but `false` in commit list.

image

image

kovsu: https://github.com/refined-github/github-url-detection/pull/200 @fregante

fregante: Merged and published

kovsu:
Screenshot updated!

source/features/clean-conversation-headers.tsx
kovsu: I'm using `parseReferenceRaw`. It works well.

source/features/clean-conversation-headers.tsx
```diff
@@ -22,35 +24,51 @@ async function cleanIssueHeader(byline: HTMLElement): Promise<void> {
...
 	if (!isDefaultBranch && !wasDefaultBranch) {
 		base.classList.add('rgh-clean-conversation-headers-non-default-branch');
 	}
+
+	// Shows on PRs: main [←] feature
+	const anchor
+		= $optional('.commit-ref-dropdown', byline)?.nextSibling // TODO: Drop old PR layout support
+		?? base.nextSibling?.nextSibling;
+	assertNodeContent(anchor, 'from');
+	anchor!.after(<span><ArrowLeftIcon className="v-align-middle mx-1" /></span>);
```
fregante, line 71: # No branch selector

Screenshot
Screenshot 1

## New view

Screenshot 2
Screenshot 3

# With branch selector

Screenshot 4
Screenshot 5

## New view

Screenshot 6
Screenshot 7

fregante (approved):

PR #8003, opened 2024-11-03
`preview-search-result` open global code search results file in modal preview

ndom91:
The goal here is to recreate a feature I really loved from https://grep.app. See:

out

Allowing you to flip through full files like this let me stay in-context while searching much more easily. Using GitHub and their default search behaviour, one has to jump back n forth between tabs all the time and it feels super annoying.

With this refined-github feature, I've added little icon buttons next to the filename in the global code search results page. When clicked they open the contents of the respective file in a `dialog` element. This dialog element also has a header with the filename and three buttons:
1) copy contents
2) open in background tab
3) close dialog

Anyway, this is my first refined-github feature so I'd be happy about any feedback or any ways to do this more effectively!

One somewhat major thing I haven't implemented, but that I would really love to have working here, is syntax highlighting in the dialog file previews. Any idea how to leverage GitHub's built-in syntax highlighting CSS here? :thinking: 

## Test URLs

https://github.com/search?q=tauri&type=code

## Screenshot

out2

--- 

I was hoping to use this as the screenshot in the features json, however I'm not sure where I can upload it now that GitHub has disabled hotlinking from most places :thinking: 

> swappy-20241103-204226


fregante:
This is an impressive PR given the correct use of a lot of our utilities 🙏 but it's rather "complex" for Refined GitHub's standards and it opens a whole can of worms: content previews.

There's a lot of content that can be previewed a way or another and I've always pushed back on that. If you have to click on something and wait for it to load, I think there's not substantial advantage to just clicking on it and… waiting for it to load in the regular view. 

Given that this also lacks some useful features that users expect (syntax highlighting, line numbers, etc) I think it just opens the door to more work and more complexity down the line. I'd suggest publishing this as a userscript or a standalone extension

ndom91:
Hey so I appreciate that it might be a bit much for refined-github. 

But if I were to get syntax highlighting and line numbers working, would you consider adding it?

fregante:
The rest of the reasons still apply, and also those two pieces would likely also make it even more complex

ndom91:
> The rest of the reasons still apply, and also those two pieces would likely also make it even more complex

Gotcha, fair enough. Just wanted to see if its worth it for me to push this PR further or pivot to my own extension or something

ndom91:
If anyone comes across this, I've added syntax highlighting and line numbers and made it a standalone extension here: https://github.com/ndom91/github-search-preview

PR #8005, opened 2024-11-04
`tags-on-commits` - Improve GitHub Enterprise compatibility

fregante:
- Closes https://github.com/refined-github/refined-github/issues/7954
- Replaces https://github.com/refined-github/refined-github/pull/7987
- Includes https://github.com/fregante/select-dom/pull/31

## Test URLs

Like this, but on GHE https://github.com/microsoft/vscode/commits/release/1.25/

## Screenshot

_Untested_

### DIFF
--- a/source/features/tags-on-commits-list.tsx
+++ b/source/features/tags-on-commits-list.tsx
@@ -1,7 +1,6 @@
 import React from 'dom-chef';
 import cache from 'webext-storage-cache/legacy.js';
-import {$$} from 'select-dom';
-import {$} from 'select-dom/strict.js';
+import {$, $$, $$optional} from 'select-dom/strict.js';
 
 import TagIcon from 'octicons-plain-react/Tag';
 import * as pageDetect from 'github-url-detection';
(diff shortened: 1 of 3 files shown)


source/features/tags-on-commits-list.tsx
```diff
@@ -90,7 +90,13 @@ async function init(): Promise<void | false> {
...
 	const cacheKey = `tags:${getRepo()!.nameWithOwner}`;
 
-	const commitsOnPage = $$('[data-testid="commit-row-item"]');
+	let commitsOnPage = $$optional('[data-testid="commit-row-item"]');
+	if (commitsOnPage.length === 0) {
+		// Try waiting a bit longer
+		// https://github.com/refined-github/refined-github/issues/7954
+		await delay(1000);
+		commitsOnPage = $$('[data-testid="commit-row-item"]');
+	}
```
fregante, line 99: - Normal users: unaffected
- GHE users: the feature retries after 1 second
- Future normal users of broken feature: error appears after 1 second, but is explicitly `Expected elements not found: [data-testid="commit-row-item"]` rather than `Expected element not found because the base is specified but null` on the wrong line

PR #8014, opened 2024-11-06
Meta: Fix build warnings

fregante:


eslint.config.js
```diff
@@ -17,6 +17,13 @@ export default antfu(
 			overrides: {
 				'ts/method-signature-style': 'off', // Disagree and it breaks types https://github.com/typescript-eslint/typescript-eslint/issues/1991
 				'ts/consistent-type-definitions': 'off', // Review later
+				'ts/consistent-type-imports': [
+					'error',
+					{
+						// Preferred style
+						fixStyle: 'inline-type-imports',
+					},
+				],
```
fregante, line 26: > Failed to resolve entry for package "type-fest". The package may have incorrect main/module/exports specified in its package.json.

package-lock.json
fregante, line 1:
```
=============

WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: >=4.7.4 <5.6.0

YOUR TYPESCRIPT VERSION: 5.6.3

Please only submit bug reports when using the officially supported version.

=============
```

PR #8015, opened 2024-11-06
`patch-diff-links` - Restore feature on non-beta issues

kovsu:
https://github.com/refined-github/refined-github/pull/7909#issuecomment-2458740998

Test URL:

https://github.com/refined-github/refined-github/commit/132272786fdc058193e089d8c06f2a158844e101

## Screenshot

New view:

image

image

Old view:

image

### DIFF
--- a/source/features/patch-diff-links.tsx
+++ b/source/features/patch-diff-links.tsx
@@ -26,7 +26,7 @@ async function addPatchDiffLinks(commitMeta: HTMLElement): Promise<void> {
 
 async function init(signal: AbortSignal): Promise<void> {
 	observe([
-		'.commit-meta > span:last-child', // `isPRCommit` + old `isSingleCommit`
+		'.commit-meta > :is(span, div):last-child', // `isPRCommit` + old `isSingleCommit`
 		'[class*="commit-header-actions"] + div pre',
 	], addPatchDiffLinks, {signal});
 }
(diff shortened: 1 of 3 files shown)


source/features/patch-diff-links.tsx
```diff
@@ -14,19 +14,23 @@ async function addPatchDiffLinks(commitMeta: HTMLElement): Promise<void> {
...
 	commitMeta!.classList.remove('no-wrap'); // #5987
-	commitMeta!.prepend(
-		<span className="sha-block" data-turbo="false">
-			<a href={`${commitUrl}.patch`} className="sha color-fg-default">patch</a>
-			{' '}
-			<a href={`${commitUrl}.diff`} className="sha color-fg-default">diff</a>
-		</span>,
-		<span className="px-2">·</span>,
+	// TODO: Improve after April 2025
+	const node = (
```
fregante, line 18: Why?

source/features/patch-diff-links.tsx
```diff
@@ -14,19 +14,23 @@ async function addPatchDiffLinks(commitMeta: HTMLElement): Promise<void> {
...
+		<>
+			<span className="sha-block" data-turbo="false">
+				<a href={`${commitUrl}.patch`} className="sha color-fg-default">patch</a>
+				{' '}
+				<a href={`${commitUrl}.diff`} className="sha color-fg-default">diff</a>
+			</span>
+			{commitMeta.tagName !== 'DIV' && <span className="px-2">·</span>}
```
fregante, line 25: You can use a ternary operator with `=== DIV ? undefined : <span>` instead of using <>

source/features/patch-diff-links.tsx
kovsu: I mean, when the old version is deprecated we just have to go back to the old code.

fregante (approved):

PR #8025, opened 2024-11-08
`conversation-authors` - Fix border partially visible in new view

kovsu:
## Test URLs

https://github.com/download-directory/download-directory.github.io/issues?q=sort%3Aupdated-desc%20is%3Aissue%20state%3Aclosed

## Screenshot

Before:

image

After:

image

### DIFF
--- a/source/features/conversation-authors.css
+++ b/source/features/conversation-authors.css
@@ -9,6 +9,6 @@
 	}
 }
 
-[data-testid="list-view-item-description"] {
+[data-testid='list-view-item-description'] {
 	margin-top: 1px;
 }
--- a/source/features/conversation-authors.css
+++ b/source/features/conversation-authors.css
@@ -9,6 +9,7 @@
 	}
 }
 
+/* Fix border partially visible in new view #8025 */
 [data-testid='list-view-item-description'] {
 	margin-top: 1px;
 }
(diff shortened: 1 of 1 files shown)


source/features/conversation-authors.css
```diff
@@ -8,3 +8,7 @@
 		margin-right: 2px !important;
 	}
 }
+
```
fregante, line 11: We need a description for these "strange" rules. Particularly the PR title and a link to the PR/issue itself would be good.

fregante (approved):

darkred:
@kovsu I tested the new version 24.11.9 with Chrome/Firefox (screen resolution is 1920x1200) and I see that the margin-top value needs to be 1.5px for me:

darkred:
@fregante Is 1px working ok for you? If not, should I make a PR using 1.5px? Or maybe @kovsu prefers to do it? Or leave it as it is?

fregante:
What OS are you on exactly? It seems that the username as an excessive amount of padding above it. Compare it to @kovsu's screenshot.

darkred:
It's Windows 11 24H2. Yes, I agree that the username seems as excessive padding. Nevertheless I noticed that for me the border gets fixed only with 1.5px (1.4px is not enough).

Anyway I took the liberty and made PR #8051 with 1.5px . Feel free to close it not needed.

PR #8028, opened 2024-11-09
Adjust spacing/borders in options

fregante:
## Before

Screenshot 11

## After

Screenshot 10


fregante:
## Mobile before

Screenshot 7

## Mobile after

Screenshot 8

PR #8038, opened 2024-11-10
`conversation-activity-filter` - Change position

fregante:
Mentioned in a previous PR. Not super convinced about this

## Test URLs

Here

## Regular

Screenshot 5

## Sticky

Screenshot 6


fregante:
Not necessary

PR #8041, opened 2024-11-10
Work around Firefox bug that causes tab freezes

### LINKED ISSUE #8039: Firefox hanging on selecting issue/PR labels
### Description

Since yesterday, an issue similar to #7176 occurs again where opening the label dropdown will freeze the whole page. It also seems to makes it unable to open any new github tabs for the next several minutes as they will seem like they are still loading for an eternity.

This issue started happening when I opened my macbook yesterday. Before that I did not notice anything. I was still running 24.11.5 (I did not notice any slowdown with that version in the days before that with that version). I updated to 24.11.9 and it still happened. And I don't believe firefox updated in the meantime my macbook was closed.

It happens for me on every issue or PR in https://github.com/home-assistant/core. We have quite a lot of labels so that might be the cause. I remember that in that last issue you got access to a similar repo with 1000+ labels. I will try to see what I can do to replicate this in a more public environment.

### Extension version

24.11.9

### Browser(s) used

Firefox

### Token ⚠️

joostlek:
Make table-input more performant as @fregante discovered in fixes #8039.

It will ensure that firefox does not freeze when opening the labels dialog when you have a large number of labels (somewhere between 500 and 2241).

### DIFF
--- a/source/features/table-input.css
+++ b/source/features/table-input.css
@@ -72,4 +72,3 @@
 .rgh-table-input .sentinel {
 	display: none;
 }
-
--- a/source/features/table-input.css
+++ b/source/features/table-input.css
@@ -35,23 +35,23 @@
 }
 
 .rgh-tic:hover div,
-.rgh-tic:is(:nth-of-type(5n + 1)):has(~ :hover:nth-of-type(5n + 1)) div,
+.rgh-tic:is(:nth-of-type(5n + 1)):has(~ .rgh-tic:hover:nth-of-type(5n + 1)) div,
 .rgh-tic:is(:nth-of-type(5n + 1), :nth-of-type(5n + 2)):has(
-		~ :hover:nth-of-type(5n + 2)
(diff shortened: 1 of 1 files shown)


fregante:
Can you see any difference between this PR and the complete removal of `source/features/table-input.css`?

fregante (approved):

joostlek:
Without the CSS file and the import at the top of the TSX file, it just seems to work. (Same disclaimer as in the issue, haven' tested the table input, just the labels)

fregante:
Yeah what I'm asking if whether this completely fixes the performance issue or if the new selector still slows you dow (even just a bit).

If you say "works like normal" then I can release this as is, probably tomorrow.

joostlek:
Ah, yes this fixes the performance issue and works like normal!

fregante:
Thanks for looking into this! I was having anxiety flashbacks from earlier this year 😵‍💫

joostlek:
Hah, same, but now I at least had some time to look into this :)

Issue #7806, opened 2024-09-18
`more-dropdown-links` duplicated on security tab


Snoopy1866:
### Description

![image](https://github.com/user-attachments/assets/89c543a6-6a30-4a4a-a088-68f4f3433f72)

Look at the picture, when I click (correctly speaking, switch click) the left tabs for _N_ times, the buttons displayed in the top-right three dots also repeated _N_ times.

### How to replicate the issue + URL

**How to replicate**: same as description
**URL**: all public repo's security page, for example: https://github.com/PyStatPower/PyStatPower/security

### Extension version

24.8.30

### Browser(s) used

Chrome 128.0.6613.138

### Token ⚠️

- [X] I set a token in the options page. I understand that some features require a token to work.

fregante:
Likely due to the console error:
```
undefined is not an object (evaluating 'menuItem.removeAttribute')
onlyShowInDropdown@webkit-masked-url://hidden/:44:10
```

kovsu:
It's a wired thing. It only happen on `security-tab`.

kovsu:
It's a weird thing. It only happens on `security-tab`.

fregante:
Likely reason:

https://github.com/refined-github/refined-github/blob/9a70bc1ea2aded61e311eddb275c6efc24baff86/source/features/clean-repo-tabs.tsx#L21

It should also happen on the Insights tab, but AJAX navigation fails there so our code runs on a fresh document every time.

kovsu:
I'm trying use `observe` in `more-dropdown-links`. It doesn't have the problem. What's your opinion? @fregante 

https://github.com/user-attachments/assets/21d87ded-6006-4b7c-a42d-aea986f0ac5f

fregante:
I think it fails the overflow logic. When you resize the window, GitHub's code will hide the tab and show the item in the dropdown. If the item hasn't been added yet it will fail.

fregante:
Oh sorry I got confused. Maybe it's not relevant here. I'm not sure how it's relevant to this issue

PR #8044, opened 2024-11-11
`more-dropdown-links` - Fix duplicates on security tab

kovsu:
Close: #7806 

## Test URLs

## Screenshot

https://github.com/user-attachments/assets/aa4f0746-027d-40e6-b411-46629850fb38

### DIFF
--- a/source/features/more-dropdown-links.tsx
+++ b/source/features/more-dropdown-links.tsx
@@ -15,6 +15,7 @@ import createDropdownItem from '../github-helpers/create-dropdown-item.js';
 import {buildRepoURL} from '../github-helpers/index.js';
 import getCurrentGitRef from '../github-helpers/get-current-git-ref.js';
 import observe from '../helpers/selector-observer.js';
+import {expectToken} from '../github-helpers/github-token.js';
 
 export async function unhideOverflowDropdown(): Promise<boolean> {
 	// Wait for the tab bar to be loaded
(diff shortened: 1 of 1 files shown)


source/features/more-dropdown-links.tsx
```diff
@@ -75,8 +81,6 @@ void features.add(import.meta.url, {
 		// No dropdown on mobile #5781
 		() => !elementExists('.js-responsive-underlinenav'),
 	],
-	deduplicate: 'has-rgh',
```
fregante, line 78: The exclusions can now also be dropped

source/features/more-dropdown-links.tsx
```diff
@@ -28,16 +30,14 @@ export async function unhideOverflowDropdown(): Promise<boolean> {
...
 }
 
-async function init(): Promise<void> {
+async function addDropdownItems(repoNavigationDropdown: HTMLElement): Promise<void> {
 	const reference = getCurrentGitRef() ?? await getDefaultBranch();
 	const compareUrl = buildRepoURL('compare', reference);
 	const commitsUrl = buildRepoURL('commits', reference);
 	const branchesUrl = buildRepoURL('branches');
 	const dependenciesUrl = buildRepoURL('network/dependencies');
 	await unhideOverflowDropdown();
```
fregante, line 39: If it's hidden, the observer will never fire. The feature is probably only working because `clean-repo-tabs` is enabled. 

I think the fix is to move this before the observer. Please. Verify the bug by disabling the other feature

Issue #7800, opened 2024-09-10
`status-subscription` - Make selected status more visible


kovsu:
### Description

The selected background is not that much visible in dark mode.

before:

<img width="346" alt="image" src="https://github.com/user-attachments/assets/022c4acc-e557-4168-ac81-bf9dd7803d25">

<img width="326" alt="image" src="https://github.com/user-attachments/assets/3b1d8872-4d06-4cac-ab11-0e6962f49136">

after:

<img width="372" alt="image" src="https://github.com/user-attachments/assets/2c0d2370-ff9f-4f6d-ac8d-75b65d51800f">

<img width="335" alt="image" src="https://github.com/user-attachments/assets/9192ed42-9939-43c7-a8e0-1dae368dfa19">

### Example URLs

https://github.com/refined-github/refined-github/pull/7798

fregante:
You’re absolutely right, their “selected” style isn’t visible in dark mode, it looks decent in light mode though. My concern with making it *that* bright is that it might capture too much attention, plus it’s not a standard style.

I’m thinking of two alternatives:

- add some color (custom CSS) just in dark mode, like some sort of glow maybe
- look for similar widgets on GitHub or primer.style
- use the “preview | code | blame” style found on files, which is basically the same component but styled more recently. I think we should go with this, but it might be a React-only component/style


![IMG_5360](https://github.com/user-attachments/assets/c5a7b366-0194-4473-8f52-2f81d264c388)

kovsu:
How about this?

<img width="342" alt="image" src="https://github.com/user-attachments/assets/91abd728-8d53-4ec2-9ce5-86d50d0d8435">

<img width="328" alt="image" src="https://github.com/user-attachments/assets/25428e7e-e15e-4fdb-832a-40768edb0fe9">

fregante:
Is it a custom style or just an existing class? If the “preview code blame” style can be reused here I’d prefer it

fregante:
We can try your custom style. We just need to make sure it works as expected on high-contrast themes too

fregante:
If you still have the style I'd love to get that into the next version

PR #8047, opened 2024-11-12
`status-subscription` - Improve contrast

kovsu:
Close: #7800 

## Test URLs

https://github.com/refined-github/refined-github/issues/7800

## Screenshot

image

image

### DIFF
--- a/source/features/status-subscription.css
+++ b/source/features/status-subscription.css
@@ -1,0 +1,11 @@
+.thread-subscribe-form .btn {
+	color: var(--fgColor-muted, var(--color-fg-muted));
+
+	&.selected {
+		color: var(--fgColor-default, var(--color-fg-default));
+
+		.octicon {
+			color: var(--fgColor-default, var(--color-fg-default));
+		}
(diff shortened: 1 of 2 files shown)


fregante:
How about the border like in your earlier screenshot?

kovsu:
@fregante I think the boreder maybe we don't need now. I think it is distracting if we add the border.

fregante:
I think the contrast here is still too low, especially when the middle one is selected in dark mode. A slightly stronger border would definitely help.

kovsu:
Actually I just want to notice it when I look at the notifications, not when I enter the page.

fregante:
Yes I'm not talking about the blue border but just a brighter/darker one. As is, there's not much improvement over the current one

fregante:
Screenshot 8


Screenshot 9

kovsu:
@fregante I'm just working on it. 😂

image

fregante:
too late haha, I wanted to release a new version 😁

PR #8048, opened 2024-11-12
Improve CI testing experience 

fregante:
I need to double-check the comment logic


fregante:
> I need to double-check the comment logic

yolo, we'll fix it in post-prod

PR #8051, opened 2024-11-12
`conversation-authors` - Avoid border clip on Windows

darkred:
Follow-up to https://github.com/refined-github/refined-github/pull/8025#issuecomment-2469007955
I noticed that the margin-top value needs to be 1.5px instead of 1px.

## Test URLs

https://github.com/download-directory/download-directory.github.io/issues?q=sort%3Aupdated-desc%20is%3Aissue%20state%3Aclosed

## Screenshot

Comparison of 1px vs 1.5px padding:
1px

1.5px

### DIFF
--- a/source/features/conversation-authors.css
+++ b/source/features/conversation-authors.css
@@ -11,5 +11,5 @@
 
 /* Fix border partially visible in new view #8025 */
 [data-testid='list-view-item-description'] {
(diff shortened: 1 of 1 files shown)


fregante (approved):
Approving for now. I'll try to figure out if there's a better way

Issue #8055, opened 2024-11-13
`pr-branch-auto-delete` is broken for me


NoahAndrews:
### Description

Here's a log from me merging a PR whose branch should get auto-deleted: (ndr: deleted, no relevant logs)

### How to replicate the issue + URL

Merge https://github.com/refined-github/sandbox/pull/100 and (hopefully) observe that the branch doesn't auto-close. I do have "Allow edits by maintainers" checked on that PR, and hopefully that's good enough. I can reproduce this even when I have full write access to the software in question.

### Extension version

24.11.12

### Browser(s) used

Chrome 130.0.6723.117

### Token ⚠️

- [X] I set a token in the options page. I understand that some features require a token to work.

fregante:
We have a number of exceptions now, does any of these match your branch?

https://github.com/refined-github/refined-github/blob/5106ef18bcc81cbf213c6e7b5bedb611bd3d67e2/source/features/pr-branch-auto-delete.tsx#L13

fregante:
Merging that test PR won't delete the branch because the branch is not mine, the button won't appear for me, even if you granted *edit* access.

fregante:
I can confirm it's not working

kovsu:
I'm back to https://github.com/refined-github/refined-github/tree/b22f2c9a8c519596a831bceae2f1f571285a18d8, it works. So https://github.com/refined-github/refined-github/tree/d47243ff5752e637c11501925089d3b6115e725f has the problem make it doesn't work.

kovsu:
https://github.com/user-attachments/assets/a13df7c7-491d-4e5d-8b4e-aca28ab26434

fregante:
Glad to hear someone uses and expects the feature, it's probably the most controversial one 😅

NoahAndrews:
Gotta save those seconds!

PR #8056, opened 2024-11-14
`pr-branch-auto-delete` - Restore feature

kovsu:
Close: #8055 

## Test URLs

Any PR you can merge

## Screenshot

https://github.com/user-attachments/assets/8a68f5a6-ef96-4998-9c38-f57abc5dd74d

### DIFF
--- a/source/github-events/on-pr-merge.ts
+++ b/source/github-events/on-pr-merge.ts
@@ -5,5 +5,5 @@ import observe from '../helpers/selector-observer.js';
 // This event ensures that the callback appears exclusively to the person that merged the PR and not anyone who was on the page at the time of the merge
 export default async function onPrMerge(callback: VoidFunction, signal: AbortSignal): Promise<void> {
 	await oneEvent('.js-merge-commit-button', 'click', {signal});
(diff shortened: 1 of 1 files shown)


kovsu:
I have "Rever" N times 😬

source/github-events/on-pr-merge.ts
```diff
@@ -5,5 +5,5 @@ import observe from '../helpers/selector-observer.js';
 // This event ensures that the callback appears exclusively to the person that merged the PR and not anyone who was on the page at the time of the merge
 export default async function onPrMerge(callback: VoidFunction, signal: AbortSignal): Promise<void> {
 	await oneEvent('.js-merge-commit-button', 'click', {signal});
-	observe('.TimelineItem-badge .octicon-git-merge', callback, {signal, once: true});
+	observe('.TimelineItem-badge .octicon-git-merge', callback, {signal});
```
fregante, line 8: Thank you! This means that the selector also matches things we don't expect and that callback is called multiple times on other elements. 

Probably not important at the moment since the PR view beta will come in the following months anyway.

fregante (approved):

NoahAndrews:
Thanks for working to fix this so quickly!

PR #8057, opened 2024-11-14
`rgh-token-user` - Fix token button style

kovsu:
- Follows https://github.com/refined-github/refined-github/pull/8002

## Test URLs

## Screenshot

Before:

image

After:

image


fregante:
🤔 How come it's missing? We use that class in a lot of places, it's being loaded for me:

Screenshot

fregante (approved):
Ah it's not that it's missing, it's just not the right class. I don't know how I made that original screenshot then 🤔

PR #8071, opened 2024-11-17
`unread-anywhere` - Change shortcut to `g u` to avoid conflicts

fregante:
- Fixes https://github.com/refined-github/refined-github/issues/8060

### DIFF
--- a/source/features/unread-anywhere.tsx
+++ b/source/features/unread-anywhere.tsx
@@ -39,8 +39,8 @@ async function openUnreadNotifications(): Promise<void> {
 	});
 }
 
-function init(): void {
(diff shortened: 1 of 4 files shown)


source/features/unread-anywhere.tsx
```diff
@@ -40,13 +40,13 @@ async function openUnreadNotifications(): Promise<void> {
 	});
 }
 
-function init(signal: AbortSignal): void {
-	registerHotkey('Mod+u', openUnreadNotifications, {signal});
```
fregante, line 44: I was mistakenly removing the listener even if `onetime` was active.

PR #8095, opened 2024-11-22
Meta: Nested CSS

fregante:
- Part of https://github.com/refined-github/refined-github/issues/6665
- Follows #7950

### DIFF
--- a/source/features/night-not-found.css
+++ b/source/features/night-not-found.css
@@ -1,145 +1,123 @@
-/* Brighten Octocat on hover 1/2 */
+/* Animate :hover Octocat brightening */
 .font-mktg > .position-relative [height='230'] {
-	transition:
-		transform 0.1s,
-		filter 1s;
+	transition: filter 1s ease-out;
 }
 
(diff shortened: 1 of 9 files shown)


source/features/night-not-found.css
```diff
@@ -1,145 +1,123 @@
-/* Brighten Octocat on hover 1/2 */
+/* Animate :hover Octocat brightening */
```
fregante, line 1: This feature sees huge improvements thanks to nesting

source/features/scrollable-areas.css
```diff
@@ -2,7 +2,8 @@
 	.comment-body,
 	.markdown-body,
 	[data-testid='markdown-body'] {
-		:is(blockquote, pre) {
+		blockquote,
+		pre {
```
fregante, line 6: Something cool about nested selectors is that they can replace `:is()` as well, sometime making it easier to read/format. 

In this case it was already nested, so `:is()` could just be dropped.

Issue #8101, opened 2024-11-24
`clean-conversation-headers`: "fixed by" link dropped


fregante:
### Description

<img width="734" alt="Screenshot" src="https://github.com/user-attachments/assets/1ae3476b-38e4-46a4-b4ef-115fba5de5ae">

Likely related:

- #8001 cc @kovsu 
- https://github.com/refined-github/refined-github/pull/8036/files#diff-272f9298e268b9341e93a968c01d3e243b9ff7d0c15aaa3d9f69b33a01341d5aL43

Cause:

<img width="426" alt="Screenshot 1" src="https://github.com/user-attachments/assets/f51043d6-7cde-4242-ac07-4a6f4d39d6a6">

### How to replicate the issue + URL

https://github.com/sindresorhus/serialize-error/issues/48

### Extension version

24.11.12 (note that it's not the latest)

### Browser(s) used

Safari

### Token ⚠️

- [X] I set a token in the options page. I understand that some features require a token to work.

PR #8102, opened 2024-11-25
`clean-conversation-headers` - Fix hidden fixed link

kovsu:
Close: #8101 

## Test URLs

https://github.com/sindresorhus/serialize-error/issues/48

## Screenshot

image

image

### DIFF
--- a/source/features/clean-conversation-headers.css
+++ b/source/features/clean-conversation-headers.css
@@ -39,14 +39,9 @@
 	&.rgh-hide-author {
 		.author + .Label, /* #5832 */
 		.author,
-		a[data-hovercard-url] {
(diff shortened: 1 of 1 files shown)


fregante (changes requested):
We have code that does the opposite, we should remove/fix that instead. 

It seems that it's targeting the duplicate issue author link on the conversation tab. 

This needs to be tested on issues, PRs, new and old view

fregante:
That conflicting selector was specifically added for the new view, so this fix would definitely need to be verified on that view too. 

`a[data-hovercard-url]` in the preceding CSS rule maybe can be replaced with something more specific to authors, like `a[data-hovercard-type=user]` if it exists.

PR #8126, opened 2024-12-07
Faster CSS hotfixes

fregante:
- Part of https://github.com/refined-github/refined-github/issues/8054

Pros:

- This will avoid the FOUC on Slow browsers (Safari…) because the hotfix no longer depends on the JS content script
- It should fix injection on GitHub Enterprise (not sure how it was broken)
	- https://github.com/refined-github/refined-github/issues/8091#issuecomment-2511308547

Edit: well, it seems that it was intentional:

https://github.com/refined-github/refined-github/blob/aff10c3ec47f0062c908bc089212e235c7cae9ea/source/options.tsx#L76-L78

### DIFF
--- a/source/background.ts
+++ b/source/background.ts
@@ -21,6 +21,10 @@ addPermissionToggle();
 // Firefox/Safari polyfill
 addOptionsContextMenu();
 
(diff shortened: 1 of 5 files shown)


fregante:
Not yet possible. See parent issue.

PR #8135, opened 2024-12-08
Open Welcome page links in new tab in Safari

fregante:
- should fix https://github.com/refined-github/refined-github/issues/8058

Reason: `instanceof HTMLAnchorElement` fails across web components. Neat.


fregante:
YOLO

PR #8140, opened 2024-12-09
Faster disableable CSS features

fregante:
- Closes https://github.com/refined-github/refined-github/issues/7944


fregante:
All verified in both states.

Skippyy96 (commented):
Thanks

Issue #8157, opened 2024-12-13
`show-names` misses space between username and display name on new issue page


sersorrel:
### Description

In the new issues page, `show-names` fails to insert a space between the (existing) username and the (inserted) display name.

### How to replicate the issue + URL

correct: https://github.com/refined-github/refined-github/issues/8147#issuecomment-2534107483

![image](https://github.com/user-attachments/assets/1e6efd01-d88b-4247-901c-05381da19d2d)

broken: https://github.com/keepassxreboot/keepassxc/issues/693#issuecomment-357446532

![image](https://github.com/user-attachments/assets/a78e9cd9-7ea2-4cbe-a653-a146fdb4c9f2)

### Extension version

24.12.10

### Browser(s) used

Firefox 133.0

### Token ⚠️

- [X] I set a token in the options page. I understand that some features require a token to work.

PR #8021, opened 2024-11-07
`show-names` - Extend support of new issue view

kovsu:
- part of https://github.com/refined-github/refined-github/issues/7856

## Test URL

https://github.com/eslint/eslint/issues/18391

## Screenshot

before:

image

after:

image

PR #8159, opened 2024-12-16
`show-names` - Fix spacing on new issues

kovsu:
Closes: #8157 

## Test URLs

- New issue view: https://github.com/keepassxreboot/keepassxc/issues/693#issuecomment-357446532

- New issue view (first comment):  https://github.com/eslint/eslint/issues/18588

## Screenshot

Old: image

New: image


source/features/show-names.tsx
```diff
@@ -24,10 +24,15 @@ async function dropExtraCopy(link: HTMLAnchorElement): Promise<void> {
 function appendName(element: HTMLAnchorElement, fullName: string): void {
 	// If it's a regular comment author, add it outside <strong> otherwise it's something like "User added some commits"
 	const {parentElement} = element;
-	const insertionPoint = parentElement!.tagName === 'STRONG' ? parentElement! : element;
+	const isStrong = parentElement?.tagName === 'STRONG';
 
+	if (isStrong) {
+		parentElement?.classList.add('inline');
```
kovsu, line 30: `strong` take up more width than its child elements.

source/features/show-names.tsx
```diff
@@ -24,10 +24,15 @@ async function dropExtraCopy(link: HTMLAnchorElement): Promise<void> {
...
+	}
+
+	const insertionPoint = isStrong ? parentElement! : element;
 	insertionPoint.after(
-		' ',
+		<>&nbsp;</>,
```
kovsu, line 35: In new issue view the first space we need to use `&nbsp;`, because whitespace is not work here. And the second space in new issue view we don't need to care about it. Its parent have flex attribute. If we do same thing in second space it will have more space than one whitespace.

source/github-helpers/selectors.ts
```diff
@@ -197,6 +197,7 @@ const authorLinks = [
 	'h3 a[data-testid="issue-body-header-author"]', // The first issue comment
 	'.react-issue-comment a[data-testid="avatar-link"]',
 	'[data-testid="comment-header"] a[data-testid="avatar-link"]', // React commit view
+	'.react-issue-body a[data-testid="issue-body-header-author"]', // React issue view first comment
```
kovsu, line 200: #8021 doesn't work now, GitHub has updated the `test-id` here.

source/features/show-names.tsx
fregante:
```suggestion
		parentElement?.classList.add('d-inline-block');
```

GitHub already has a lot of utility classes

source/features/show-names.tsx
```diff
@@ -24,10 +24,15 @@ async function dropExtraCopy(link: HTMLAnchorElement): Promise<void> {
 function appendName(element: HTMLAnchorElement, fullName: string): void {
 	// If it's a regular comment author, add it outside <strong> otherwise it's something like "User added some commits"
 	const {parentElement} = element;
-	const insertionPoint = parentElement!.tagName === 'STRONG' ? parentElement! : element;
+	const isStrong = parentElement?.tagName === 'STRONG';
```
fregante, line 27:
```suggestion
	const isStrong = parentElement!.tagName === 'STRONG';
```

This is never missing, we should always prefer `!.` because `?.` silences future bugs

source/github-helpers/selectors.ts
fregante: If selectors are no longer matching where they used to be, we should drop them. The view is still in beta so the selector will not appear in GHE.

source/features/show-names.tsx
kovsu: Dont't need this.

fregante:
Changing from flex to inline-block might introduce unwanted behavior particularly on mobile, so I replaced it with a conditional spacing

## Screenshot

Screenshot 4

### Old version still working

Screenshot 5

Issue #8123, opened 2024-12-04
Features broken on new PR merge box beta 


fregante:
- https://github.com/orgs/community/discussions/143787
- https://github.blog/changelog/2024-12-03-improved-pull-request-merge-experience-now-in-public-preview/

One thing to note is that `clean-checks-list` is now implemented natively 🎉 

The first step here is to figure out what's broken exactly.

kovsu:
- [x] `clear-pr-merge-commit-message`
	- new bug opened https://github.com/refined-github/refined-github/issues/8246
- [x] `closing-remarks` (specifically after clicking the merge button)
- [x] `cross-deleted-pr-branches`
- [x] `pr-branch-auto-delete`
- [x] `update-pr-from-base-branch`
- [x] `pr-base-commit`
- [x] https://github.com/refined-github/refined-github/issues/8902
- [x] `suggest-commit-title-limit`
- [ ] ~~`convert-pr-to-draft-improvements`~~ 👋

fregante:
I'll also list features that are no longer needed:

- `clean-checks-list`
- `prevent-pr-merge-panel-opening`

jpike88:
pr-branch-auto-delete is the most pressing one, can that be prioritized

fregante:
@jpike88 done.

`update-pr-from-base-branch` would be another one that's super important.

We should just add a whole section to the box instead of trying to extend the existing message _as natively as possible,_ because that solution was unmaintainable:

- https://github.com/refined-github/refined-github/issues/6574

fregante:
@alanconway this repo is for a browser extension. You should report that to GitHub

vintagesucks:
After #8193, I see this in a private repo. Should I open a new issue?

![Image](https://github.com/user-attachments/assets/c30518f9-3db4-458a-8160-4cbffb745385)

Without Refined GitHub, it looks like this:

![Image](https://github.com/user-attachments/assets/e3739644-493d-4c00-bf3f-94e1538ed604)

---

<details>
<summary>DOM without `update-pr-from-base-branch` (25.1.24)</summary>

```html
[...]

PR #8164, opened 2024-12-18
`closing-remarks` - Support new PR merge box

fregante:
- For #8123 

## Test URLs

Any open PR, like this page

## Screenshot

Gif also shows other auto-deletion working, but that's a separate PR

merging 2

### DIFF
--- a/source/github-events/on-pr-merge.ts
+++ b/source/github-events/on-pr-merge.ts
@@ -1,4 +1,4 @@
-import oneEvent from 'one-event';
+import {oneEvent} from 'delegate-it';
 
 import observe from '../helpers/selector-observer.js';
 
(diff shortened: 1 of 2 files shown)


source/helpers/selector-observer.tsx
```diff
@@ -46,7 +48,7 @@ export default function observe<
 	}
 
 	const selector = typeof selectors === 'string' ? selectors : selectors.join(',\n');
-	const seenMark = 'rgh-seen-' + getCallerID();
+	const seenMark = 'rgh-seen-' + getCallerID(ancestor);
```
fregante, line 51: I don’t even know  how this worked until now. The observer's callback was only called once because the generated `rgh-seen` class name was identical and successive callbacks would fail.

The observer should likely be refactored to make sure all listeners are called (probably just needs a `await Promise.resolve()` before removing the class but alas)

source/github-events/on-pr-merge.ts
```diff
@@ -1,9 +1,30 @@
...
+	});
+
+	// TODO: Restore delegate-it after https://github.com/fregante/delegate-it/issues/55
+	// Possible selector + textContent filter: '[aria-label="Checks"] ~ div:has(textarea) button[data-size="medium"][data-variant="primary"]'
+	await oneEvent(document.body, 'click', {signal, filter: ({target}: Event) => {
+		const clicked = target as HTMLElement;
+		// TODO: Drop `js-merge-commit-button` in May 2025
+		return Boolean(clicked.closest('.js-merge-commit-button'))
+			|| /^Confirm .+ merge$/i.test(clicked.textContent ?? '');
+	}});
```
fregante, line 20: The selector I found was not specific enough, it also matched the initial "Merge and squash" button. 

But actually with `:has(textarea)` this appears to be enough.

I'll wait for https://github.com/fregante/delegate-it/issues/55 to maybe restore an extra `textContent` check

PR #8189, opened 2025-01-07
`easy-toggle-files` - Support new commit view

kidonng:
#7808

## Test URLs

https://github.com/refined-github/refined-github/commit/7b12c475a54ee32531ea4f9a548f08aea1081b83

## Screenshot

https://github.com/user-attachments/assets/4cf494ca-a1a8-4f8b-8bce-33816c132623

### DIFF
--- a/source/features/easy-toggle-files.tsx
+++ b/source/features/easy-toggle-files.tsx
@@ -9,9 +9,22 @@ function toggleFile(event: DelegateEvent<MouseEvent>): void {
 	const elementClicked = event.target as HTMLElement;
 	const headerBar = event.delegateTarget;
 
-	// The clicked element is either the bar itself or one of its 2 children
-	if (elementClicked === headerBar || elementClicked.parentElement === headerBar) {
-		$('[aria-label="Toggle diff contents"]', headerBar)
+	if (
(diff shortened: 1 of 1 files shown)


fregante:
PR/commits/released are time-based, they can use "new"

source/features/easy-toggle-files.tsx
```diff
@@ -9,9 +9,22 @@ function toggleFile(event: DelegateEvent<MouseEvent>): void {
...
-		$('[aria-label="Toggle diff contents"]', headerBar)
+	if (
+		// The clicked element is either the bar itself or one of its 2 children
+		elementClicked === headerBar
+		|| elementClicked.parentElement === headerBar
+		// React
+		|| elementClicked.matches([
+			'[class^="DiffFileHeader-module__diff-file-header"] > div',
+			'[class^="DiffFileHeader-module__diff-file-header"] > div > div',
+		])
```
fregante, line 20: Not a fan of this, maybe we can find better. It looks like all the active elements are either `button` or `a` so we should be able to safely exclude them:

```suggestion
		|| !elementClicked.closest('a, button')
```

fregante (approved):
LGTM after that change, if safe

PR #8194, opened 2025-01-07
`show-names` - Add support for RTL names

### LINKED ISSUE #8191: Issue with display user Arabic name
### Description

Arabic names in parentheses are not displayed correctly. Instead, the order of the Arabic text appears reversed

image

The correct shape is `محمد الصوالحي`

as here

image

1. Follow any user that have an Arabic name like me https://github.com/AhmedElTabarani
2. Go to the home page and see my activities, my Arabic name will display in the reversed order

### Extension version

 24.12.20

### Browser(s) used

Google Chrome

### Token ⚠️

kidonng:
Fix #8191

## Test URLs

- RTL: https://github.com/refined-github/refined-github/issues/8191
- Bidi override case 1: https://togithub.com/FortAwesome/Font-Awesome/issues/2465
- Bidi override case 2: https://togithub.com/w3c/webdriver/issues/385#issuecomment-598407238

## Screenshot

| | Before | After |
|--------|--------|--------|
| RTL | CleanShot 2025-01-07 at 18 48 22@2x | CleanShot 2025-01-07 at 18 48 40@2x |
| Bidi override case 1 | CleanShot 2025-01-07 at 19 26 04@2x | CleanShot 2025-01-07 at 19 26 28@2x |
| Bidi override case 2 | CleanShot 2025-01-07 at 19 26 12@2x | CleanShot 2025-01-07 at 19 26 36@2x |

### DIFF
--- a/source/features/show-names.tsx
+++ b/source/features/show-names.tsx
@@ -107,4 +107,10 @@ Test URLs:
 - mannequins: https://togithub.com/python/cpython/issues/67591
 - newsfeed: https://github.com
 
+Special cases:
+
+- RTL: https://github.com/refined-github/refined-github/issues/8191
+- Bidi override case 1: https://togithub.com/FortAwesome/Font-Awesome/issues/2465
+- Bidi override case 2: https://togithub.com/w3c/webdriver/issues/385#issuecomment-598407238
+
 */
--- a/source/features/show-names.tsx
+++ b/source/features/show-names.tsx
@@ -25,7 +25,7 @@ function appendName(element: HTMLAnchorElement, fullName: string): void {
 	const insertionPoint = parentElement!.tagName === 'STRONG' ? parentElement! : element;
 	const nameElement = (
 		<span className="color-fg-muted css-truncate d-inline-block ml-1">
-			{/* .css-truncate-target sets display: inline-block and can confine bidi overrides */}
+			{/* .css-truncate-target sets display: inline-block and confines bidi overrides #8191 */}
 			(<span className="css-truncate-target" style={{maxWidth: '200px'}}>{fullName}</span>)
 		</span>
 	);
--- a/source/features/show-names.tsx
+++ b/source/features/show-names.tsx
@@ -65,6 +65,8 @@ async function updateLinks(found: HTMLAnchorElement[]): Promise<void> {
 
 		for (const element of elements) {
 			if (isUsernameAlreadyFullName(username, fullName)) {
+				// Confine bidi overrides #8191
+				element.classList.add('d-inline-block');
 				element.textContent = fullName;
 			} else {
 				appendName(element, fullName);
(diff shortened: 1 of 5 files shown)


source/features/show-names.tsx
```diff
@@ -27,7 +25,8 @@ function appendName(element: HTMLAnchorElement, fullName: string): void {
 	const insertionPoint = parentElement!.tagName === 'STRONG' ? parentElement! : element;
 	const nameElement = (
 		<span className="color-fg-muted css-truncate d-inline-block ml-1">
-			(<bdo className="css-truncate-target" style={{maxWidth: '200px'}}>{fullName}</bdo>)
+			{/* .css-truncate-target sets display: inline-block and can confine bidi overrides */}
+			(<span className="css-truncate-target" style={{maxWidth: '200px'}}>{fullName}</span>)
```
fregante, line 29: We should fetch the original buggy test URL from a few years ago and add it to the list below.

kidonng: 😅 Oh you are right, I just noticed we also merge names when appropriate, and that case is not fixed yet

fregante: I think you're referring to when we change "georgemichael" to "George Michael" instead of adding a new name next to it. But I don’t think that can ever happen with RTL scripts

kidonng: It's not RTL, it's that bidi override can still break stuff in the merged case (the if branch in following code, this diff is the `appendName` in the else branch)

https://github.com/refined-github/refined-github/blob/76d23a0d6d78f54a34f4e956bf58f8733034f4c0/source/features/show-names.tsx#L68-L72

fregante: Interestingly, it does (why is `isUsernameAlreadyFullName` so broad?) but I'm not seeing that bug in v24.12.20 (Safari) so I think there's nothing to fix (i.e. no need to add `d-inline-block`)

fregante (approved):
Bugs and bugfixes confirmed 🙏

source/features/show-names.tsx
```diff
@@ -66,6 +65,8 @@ async function updateLinks(found: HTMLAnchorElement[]): Promise<void> {
 
 		for (const element of elements) {
 			if (isUsernameAlreadyFullName(username, fullName)) {
+				// Confine bidi overrides #8191
+				element.classList.add('d-inline-block');
```
fregante, line 69: Optionally:


- add the offending string to the tests of `isUsernameAlreadyFullName`
- change `/W` with a more restricted amount of characters, like `[],.:;()-_–— ` in https://github.com/refined-github/refined-github/blob/d6897f4b7959a4b47eccb8d81c1cc8a15d5096c2/source/github-helpers/index.ts#L52

fregante: Done

PR #8201, opened 2025-01-09
`small-user-avatars` - Fix position

kidonng:
## Test URLs

https://github.com/refined-github/refined-github/releases/tag/23.9.21

## Screenshot

| Before | After |
|--------|--------|
| CleanShot 2025-01-09 at 14 25 58@2x | CleanShot 2025-01-09 at 14 26 13@2x |


fregante:
I wonder if this is some browser/OS inconsistency. It looked fine in the original PR

fregante:
I think it looks fine as is, it's a bit a too low without the spacing:

Screenshot 7


Screenshot 8

PR #8208, opened 2025-01-09
GitHub bug fix: remove bad spacing in notifications header

### LINKED ISSUE #8203: Notification page refactoring
### Description

# Filters
| Before | After |
| ------ | ----- |
| grafik | grafik |

Currently, we have (more than) **three** different places for (query) filters:
* Toggle "All vs Unread"
* Built-in "Saved" and "Done"
* Custom filters

# Grouping
**Before**
grafik

**After**
grafik

# Related views
**Before**
grafik

**After**
grafik

Also, since we are already talking about inconsistencies and possible improvements. What's up with the space there?
| Before | After |
| ------ | ----- |
| image | image |

I guess, it is to give new notifications breathing space, but this spacing can be also added to the otherwise hidden element instead of the parent:
| Before | &rarr;  | After |
| ------ | ------- | ----- |
| image | image | image |

### Example URLs

shaedrich:
Closes  #8203

## Test URLs

## Screenshot
**Before**
grafik

**After**
|   |   |
| - | - |
| grafik | grafik |

### DIFF
--- a/source/features/fix-notifications-layout.css
+++ b/source/features/fix-notifications-layout.css
@@ -1,26 +1,4 @@
-body.notifications-v2
-	.js-check-all-container
-	.gap-3:has(.gap-4.pl-3 > div > .js-socket-channel.js-updatable-content) {
-	gap: 0 !important; /* Turns .gap-3 off */
-}
 
-body.notifications-v2
-	.js-check-all-container
-	.gap-4.pl-3:has(div > .js-socket-channel.js-updatable-content) {
-	padding-left: 0 !important; /* Moves .pl-3 to child element that toggles visibility */
-}
-
-body.notifications-v2
-	.js-check-all-container
-	.gap-4.pl-3
-	> div
-	> .js-socket-channel.js-updatable-content
-	> a.h6 {
-	padding-left: var(
-		--base-size-24,
-		24px
-	) !important; /* Turns .pl-3 into .pl-4 */
-}
 
 /*
 
--- a/source/features/fix-notifications-layout.css
+++ b/source/features/fix-notifications-layout.css
@@ -1,9 +1,0 @@
-
-
-/*
-
-Test URLs:
-
-- Notifications: https://github.com/notifications
-
-*/
--- a/source/features/fix-notifications-layout.css
+++ b/source/features/fix-notifications-layout.css
@@ -21,3 +21,11 @@ body.notifications-v2
 		24px
 	) !important; /* Turns .pl-3 into .pl-4 */
 }
+
+/*
+
+Test URLs:
+
+- Notifications: https://github.com/notifications
+
+*/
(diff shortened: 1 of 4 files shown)


source/features/fix-notifications-layout.css
```diff
@@ -0,0 +1,31 @@
...
+	.gap-4.pl-3
+	> div
+	> .js-socket-channel.js-updatable-content
+	> a.h6 {
+	padding-left: var(
+		--base-size-24,
+		24px
+	) !important; /* Turns .pl-3 into .pl-4 */
+}
+
```
fregante, line 24: Thanks for the PR! The selector is unnecessarily specific though. Check https://blog.pixiebrix.com/blog/7-bite-sized-tips-for-reliable-web-automation-and-scraping-selectors for some advice and check other features to see other acceptable selectors. 

In practice I expect this feature to be about 5 lines long, with a selector that looks something like:

`grouped button:last-child`

The ":last-child" would act as a condition to disable the padding only when there's nothing after the it. 

As for the button selector, we already have a feature that acts on it ("unwrap unnecessary dropdowns") so you can copy the selector from there. 

Since the CSS is so small, it should just go in GitHub-bugs.css instead of being a dedicated feature

fregante (changes requested):

source/features/fix-notifications-layout.css
shaedrich: The difference is, that "unwrap unnecessary dropdowns" is not a CSS-only feature and can therefore use `github-url-detection` to make sure, it is only applies to the notification page—that's why I needed to make sure of that via a CSS-only version

shaedrich: The difference is, that "unwrap unnecessary dropdowns" is not a CSS-only feature and can therefore use `github-url-detection` to make sure, it is only applies to the notification page—that's why I needed to make sure of that via a CSS-only version

fregante: That feature use the selector `[action="/notifications/beta/update_view_preference"]`, so it's specific enough.

shaedrich: Ah, thanks for the hint 👍🏻

source/features/github-bugs.css
```diff
@@ -102,6 +102,29 @@ svg.octicon-package + a.Truncate[href*='/releases/download/'] > .Truncate-text {
...
+:has([action='/notifications/beta/update_view_preference'])
+	.gap-4.pl-3
+	> div
+	> .js-socket-channel.js-updatable-content
+	> a.h6 {
+	padding-left: var(
+		--base-size-24,
+		24px
+	) !important; /* Turns .pl-3 into .pl-4 */
+}
```
fregante, line 126:
```suggestion
/* Matches .d-md-block */
@media (min-width: 768px) {
	/* Pulls the alert container to the right when empty */
	.gap-4.pl-3:has(> .d-md-block > [data-url^="/notifications/beta/recent_notifications_alert"]):not(:has(a.h6)) {
		margin-right: calc(0px - var(--base-size-16, 16px) * 2);
	}
}
```


While not as "clean" it's much easier to reason about.

shaedrich: You know what you need better than I do—be my guest 🤷🏻‍♂️

fregante (approved):
I'm a bit hesitant about the :has() selector so I'll re-review later on

PR #8225, opened 2025-01-14
`quick-mention` - Support new issue view

kovsu:
- Part of #7856

## Test URLs

https://github.com/download-directory/download-directory.github.io/issues/127

https://github.com/eslint/eslint/issues/8213

## Screenshot
Normal Issue:
 CleanShot 2025-01-15 at 10 09 24
 
Locked issue:
 CleanShot 2025-01-15 at 10 17 11@2x

### DIFF
--- a/source/features/quick-mention.css
+++ b/source/features/quick-mention.css
@@ -14,6 +14,10 @@
 	padding: 4px 12px !important;
 }
 
+.rgh-quick-mention.react-view {
+	top: 58px;
+}
+
 /* Show icon in black if it's not hovered/focused, instead of the blue color caused by .btn-link */
 .rgh-quick-mention:not(:hover, :focus) {
 	color: inherit;
--- a/source/features/quick-mention.css
+++ b/source/features/quick-mention.css
@@ -1,6 +1,6 @@
 /* Show button when the avatar or the button itself are hovered/focused */
 :not(:hover, :focus) + .rgh-quick-mention:not(:hover, :focus) {
-	opacity: 10%;
+	opacity: 20%;
 }
 
 .rgh-quick-mention {
@@ -15,7 +15,8 @@
 }
 
 .rgh-quick-mention.react-view {
-	top: 58px;
+	position: static !important;
+	translate: none;
 }
 
 /* Show icon in black if it's not hovered/focused, instead of the blue color caused by .btn-link */
--- a/source/features/quick-mention.tsx
+++ b/source/features/quick-mention.tsx
@@ -22,7 +22,7 @@ function prefixUserMention(userMention: string): string {
 
 function mentionUser({delegateTarget: button}: DelegateEvent): void {
 	const userMention = button.parentElement!.querySelector('img')!.alt;
-	const newComment = $(fieldSelector);
+	const newComment = $(['textarea#new_comment_field', '[data-testid="markdown-editor-comment-composer"] textarea']);
 	newComment.focus();
 
 	// If the new comment field has selected text, don’t replace it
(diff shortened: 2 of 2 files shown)


source/features/quick-mention.tsx
```diff
@@ -13,7 +13,10 @@ import features from '../feature-manager.js';
 import {getUsername, isArchivedRepoAsync} from '../github-helpers/index.js';
 import observe from '../helpers/selector-observer.js';
 
-const fieldSelector = 'textarea#new_comment_field';
+const fieldSelector = [
+	'textarea#new_comment_field',
+	'#react-issue-comment-composer textarea',
+];
```
fregante, line 19:
```suggestion
] as const;
```

This will make the $() below detect they're textareas without further assertions.

kovsu: Got it.

fregante:
Please open PRs from local branches, I can't make changes to your PRs otherwise

Screenshot

source/features/quick-mention.css
```diff
@@ -14,6 +14,10 @@
 	padding: 4px 12px !important;
 }
 
+.rgh-quick-mention.react-view {
+	top: 58px;
+}
```
fregante, line 19:
```patch
commit f53968a3f99e9a6532dfa2ed72e46e7f6ce4bb2c
Author: Federico <me@fregante.com>
Date:   Wed Jan 15 17:19:03 2025 +0100

    No magic numbers
---
 source/features/quick-mention.css | 5 +++--
 1 file changed, 3 insertions(+), 2 deletions(-)

diff --git a/source/features/quick-mention.css b/source/features/quick-mention.css
index 62f9a308..127ded13 100644
--- a/source/features/quick-mention.css
+++ b/source/features/quick-mention.css
@@ -1,6 +1,6 @@
 /* Show button when the avatar or the button itself are hovered/focused */
 :not(:hover, :focus) + .rgh-quick-mention:not(:hover, :focus) {
-	opacity: 10%;
+	opacity: 20%;
 }
 
 .rgh-quick-mention {
@@ -15,7 +15,8 @@
 }
 
 .rgh-quick-mention.react-view {
-	top: 58px;
+	position: static !important;
+	translate: none;
 }
 
 /* Show icon in black if it's not hovered/focused, instead of the blue color caused by .btn-link */
```

kovsu:
> Please open PRs from local branches, I can't make changes to your PRs otherwise

It's ok now.

source/features/quick-mention.tsx
```diff
@@ -50,34 +53,51 @@ function add(avatar: HTMLElement): void {
...
+
 	// Wrap avatars next to review events so the inserted button doesn't break the layout #4844
 	if (avatar.classList.contains('TimelineItem-avatar')) {
 		avatar.classList.remove('TimelineItem-avatar');
 		wrap(avatar, <div className="avatar-parent-child TimelineItem-avatar d-none d-md-block" />);
 	}
 
-	const userMention = $('img', avatar).alt;
+	if (!isOldView) {
+		wrap(avatar, <div className="avatar-parent-child d-none d-md-block" />);
```
kovsu, line 91: Old view don't need to be wrapped. It's already wrapped.

source/features/quick-mention.tsx
```diff
@@ -50,34 +53,51 @@ function add(avatar: HTMLElement): void {
...
+		if (!contentItem) {
+			return;
+		}
 	}
 
 	if (debug) {
 		timelineItem.style.border = 'solid 5px green';
 	}
 
+	const isOldView = !avatar.hasAttribute('data-testid');
```
fregante, line 82: Line 58 appears to suggest that only the old view has a specific element, so you can use the same detection here instead of a new detection

```suggestion
	const isOldView = Boolean(timelineItem);
```

I'd also move this line above the first condition so you can delete the comment on line 58

fregante: By the way, if the differences between two views are more than a couple of selectors, it's sometimes easier to copy-paste the whole `init` and customize it properly. When the time comes, I can just delete the old `init` instead of walking through each condition. This method also makes it easier to write the new version. 

Not necessary though in this PR though

source/features/quick-mention.tsx
```diff
@@ -50,34 +53,51 @@ function add(avatar: HTMLElement): void {
...
-	) {
-		return;
+	if (timelineItem) {
+		// Old view
+		if (debug) {
+			timelineItem.style.border = 'solid 5px red';
+		}
+
+		if (
+			// TODO: Rewrite with :has()
```
fregante, line 64:
```suggestion
```

fregante (approved):

PR #8233, opened 2025-01-22
Meta: Cleanup dependencies

fregante:
- Drops unused code
- Extracts utils into https://github.com/fregante/webext-bugs

### Tasks

- [x] Test target-blank polyfill in
	- [x] Safari
- [ ] Test options menu in
	- [ ] Safari
		- I assume it works. Can't test in iOS simulator
	- [x] Firefox
- [ ] Test "Enable on all websites" in 
	- [ ] Safari
		- It seems that "Allow all" doesn't actually grant `*://*/*` but just "all manifest permissions", which is perfectly fine and desirable
	- [x] Firefox


yakov116:
Nice

PR #8242, opened 2025-01-28
`quick-review` - Update the quick review hint to display the key to hold for approving without confirmation on macOS

pavinduLakshan:
$subject

## Test URLs

To be added

## Screenshot

To be added


fregante:
No need, thank you

PR #8265, opened 2025-02-06
`pr-approvals-count` - Highlight "Review Required" in yellow

SierraKomodo:
An addition to the [Shows color-coded review counts in PR lists.](https://github-production-user-asset-6210df.s3.amazonaws.com/83146190/253125143-d10d95df-4a89-4692-b218-5eba5cd79906.png) feature. Highlights the 'Review Required' text in yellow so it pops more easily when skimming PR lists.

I would appreciate some guidance on a few things:
1. Existing CSS vars for `--rgh-green` and `--rgh-red` reference `--color-*` vars that I'm unable to find in refined-github or in github's own CSS files, so I cannot apply an applicable one to the `--rgh-yellow` var I've added.
2. I'm not 100% confident I sourced the default color for `--rgh-yellow` from the correct place. I've made the assumption it's meant to be the hex code that the `--fgColor-*` var being used is assigned by github's CSS files.
3. I'm also not 100% confident `--fgColor-attention` is the correct color var to use for this. It was the closest to yellow/amber I could find that looked intended for text colors.

## Test URLs
https://github.com/Baystation12/Baystation12/pulls?q=sort%3Aupdated-desc+is%3Apr+is%3Aopen

### Before
firefox_2GJiRKlgQV

### After
firefox_Lqavrp3hrx

### DIFF
--- a/source/refined-github.css
+++ b/source/refined-github.css
@@ -1,6 +1,7 @@
 :root {
 	--rgh-green: var(--fgColor-success, var(--color-success-fg, #1a7f37));
 	--rgh-red: var(--fgColor-danger, var(--color-danger-fg, #cf222e));
+	--rgh-yellow: var(--fgColor-attention, #c69026);
 	--rgh-border-color: var(
 		--borderColor-muted,
 		var(--color-border-muted, #d8dee4)
--- a/source/features/pr-approvals-count.css
+++ b/source/features/pr-approvals-count.css
@@ -1,3 +1,5 @@
+/* TODO: Drop coloring in July 2025. New PR view renders this defunct and redundant. */
+
 /* Show approvals count in PR list */
 .js-issue-row
 	.color-fg-muted
--- a/source/features/pr-approvals-count.css
+++ b/source/features/pr-approvals-count.css
@@ -22,6 +22,12 @@
 	color: var(--rgh-red) !important;
 }
 
+.js-issue-row
+	.color-fg-muted
+	[href$='#partial-pull-merging'][aria-label*='required'] {
+	color: var(--rgh-yellow) !important;
+}
+
 /* Disable now-duplicate tooltip */
 .js-issue-row
 	.color-fg-muted


fregante:
I like it. Does your CSS also works on the new view?

Test URL: https://github.com/Baystation12/Baystation12/issues?q=sort%3Aupdated-desc+is%3Apr+is%3Aopen

source/refined-github.css
```diff
@@ -1,6 +1,7 @@
 :root {
 	--rgh-green: var(--fgColor-success, var(--color-success-fg, #1a7f37));
 	--rgh-red: var(--fgColor-danger, var(--color-danger-fg, #cf222e));
+	--rgh-yellow: var(--fgColor-attention, #c69026);
```
fregante, line 4: Yeah this looks correct. Can you also find the second variable name? It might still be used on GitHub enterprise

SierraKomodo: I'm not sure on the second variable name (See beginning of the OP, or the quoted line below). I wasn't able to find any CSS classes in github's served files that matched the naming scheme.

I also don't have access to any enterprise instances to check those.

> 1. Existing CSS vars for `--rgh-green` and `--rgh-red` reference `--color-*` vars that I'm unable to find in refined-github or in github's own CSS files, so I cannot apply an applicable one to the `--rgh-yellow` var I've added.

SierraKomodo:
> I like it. Does your CSS also work on the new view?
> 
> Test URL: [Baystation12/Baystation12/issues (sort:updated-desc is:pr is:open)](https://github.com/Baystation12/Baystation12/issues?q=sort%3Aupdated-desc+is%3Apr+is%3Aopen)

I didn't know this view existed. Pulling it up with web-ext, it looks like the feature needs me to be logged in, so I checked with manually adding the CSS to firefox's debugger instead.

It looks like it has no effect, which is the desired result I believe?

firefox_9FqiIhKoin

fregante:
Oh interesting, they added the icons there, rendering our coloring outdated. 

So it's good as is. Let's add a TODO above these 3 rules saying "Drop coloring in July 2025"

SierraKomodo:
Looks like the entire file will become irrelevant once GitHub's changes go live, so I dropped it at the beginning of the file itself.

fregante:
It's true that it doesn't work as is, but we'd still want the core part of the feature: showing the number of approvals. I don't know if that's going to be possible though

SierraKomodo:
You'd think I'd realize "There's more to this feature than colors" considering the title of it lol.

Do you prefer comments relating to specific lines of code to be on the prior line, or the same line? I.e.:

```
color: var(--rgh-green); /* TODO: Drop coloring in July 2025. New PR view renders this defunct and redundant. */
```

Or

```
/* TODO: Drop coloring in July 2025. New PR view renders this defunct and redundant. */
color: var(--rgh-green);
```

Issue #8255, opened 2025-02-05
`action-used-by-link`: this doesn't seem to work


nazdridoy:
### Description

I don't see any "usages example" button.

![Image](https://github.com/user-attachments/assets/34c731ec-b3db-430d-82bf-e1490577bc54)

### How to replicate the issue + URL

https://github.com/marketplace/actions/urlchecker-action

### Extension version

25.1.24

### Browser(s) used

firefox and zen

### Token ⚠️

- [x] I set a token in the options page. I understand that some features require a token to work.

fregante:
Confirmed on Safari too

PR #8285, opened 2025-02-12
`action-used-by-link` - Restore feature

kovsu:
Close: #8255

## Test URLs

https://github.com/marketplace/actions/urlchecker-action

### DIFF
--- a/source/features/action-used-by-link.tsx
+++ b/source/features/action-used-by-link.tsx
@@ -6,8 +6,8 @@ import * as pageDetect from 'github-url-detection';
 import features from '../feature-manager.js';
 import observe from '../helpers/selector-observer.js';
 
-function getActionURL(): URL {
-	const actionRepo = $('[data-testid="resources"] > ul a:has(.octicon-repo)')
+function getActionURL(side: HTMLElement): URL {
+	const actionRepo = $('a:has(.octicon-repo)', side)
 		.pathname
 		.slice(1);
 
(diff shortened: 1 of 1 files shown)


source/features/action-used-by-link.tsx
```diff
@@ -7,7 +7,7 @@ import features from '../feature-manager.js';
 import observe from '../helpers/selector-observer.js';
 
 function getActionURL(): URL {
-	const actionRepo = $('aside a:has(.octicon-repo)')
+	const actionRepo = $('[data-testid="resources"] > ul a:has(.octicon-repo)')
```
fregante, line 10:
```suggestion
	const actionRepo = $('a:has(.octicon-repo)', side)
```

You can pass the element selected by the observer instead

source/features/action-used-by-link.tsx
```diff
@@ -23,7 +23,7 @@ function getActionURL(): URL {
 }
 
 function addUsageLink(side: HTMLElement): void {
-	const actionURL = getActionURL();
+	const actionURL = getActionURL(side);
 
 	side.after(
```
fregante, line 28: Is this still right? The observer used to select a link and we'd add a link after it.

Now it selects a UL and we add a link *outside* it. It feels like it's missing a `li`, in fact the screenshot is not aligned and doesn't feel native

kovsu: Yeah, but it's a complex dom structure in `li`. To be consistent we need to add extra css or copy the `li` and do replacement many times. Maybe it doesn't worth this.

fregante: I should write a reusable helper to aid this kind of extension, we're gonna need more now that they use random CSS classes…

fregante (approved):

source/features/action-used-by-link.tsx
kovsu: These selectors that we add recently iss so unstable. It's a big reason I won't to try to fix the features. I'm waiting.... 🥲

Issue #8275, opened 2025-02-10
Enter in `Find features` field disables all features


injust:
### Description

If I press <kbd>Enter</kbd> here:

<img width="522" alt="Image" src="https://github.com/user-attachments/assets/a62f292e-002c-439f-aae0-1836aacd7672" />

It disables all features. I don't think this is intended, given how the functionality to do that is hidden away in Debugging.

### How to replicate the issue + URL

chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/assets/options.html

### Extension version

25.1.24

### Browser(s) used

Chrome 133

### Token ⚠️

- [x] I set a token in the options page. I understand that some features require a token to work.

fregante:
oof I love/hate this. I wonder what's triggering this

PR #8292, opened 2025-02-19
Options page: Don't disable all features when pressing `enter`

kovsu:
Close: #8275 

## Screenshot

https://github.com/user-attachments/assets/0951a496-9d02-435b-af08-58c6bdb1be09

### DIFF
--- a/source/options.html
+++ b/source/options.html
@@ -58,6 +58,7 @@
 				If you want to go through and only select a few improvements, you'll miss out on the best parts of Refined GitHub. Also note that new features will still be enabled by default and that some CSS-only refinements cannot be disabled.
 			</p>
 			<p>
+				<button hidden />
 				<button id="disable-all-features">Disable all features</button>
 				<button id="enable-all-features">Enable all features</button>
 			</p>
--- a/source/options.html
+++ b/source/options.html
@@ -20,6 +20,8 @@
 <div id="js-failed">JavaScript failed to load. Your development build failed or your browser has some issue.</div>
 
 <form id="options-form" class="detail-view-container">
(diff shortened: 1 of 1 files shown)


kovsu:
When I prevent the submit event in form, it's also will trigger disable all features button. So the easiest way slove issue is add a hidden button here. When press enter in input, it triggers the hidden button's click event and nothing will happen.

source/options.html
```diff
@@ -58,6 +58,7 @@
 				If you want to go through and only select a few improvements, you'll miss out on the best parts of Refined GitHub. Also note that new features will still be enabled by default and that some CSS-only refinements cannot be disabled.
 			</p>
 			<p>
+				<button hidden />
```
fregante, line 61: Excellent, Konv!

I'd just move this right after `<form>` and add a comment explaining why it's here:

```html
<form id="options-form" class="detail-view-container">
	<button hidden /> <!-- Captures and ignores native enter-to-submit action -->
```

fregante (approved):

fregante:
For the record, this bug was introduced by https://github.com/refined-github/refined-github/pull/8130

source/options.html
kovsu: It's wired. I move the button to top and the page be broken.

CleanShot 2025-02-19 at 14 22 10@2x

fregante: > I think the button can't be self closing tag.


That's right, HTML is not XHTML. `/` does not alter how tags behave. `<input>` is always self-closing and `<button>` is never self-closing

fregante (approved):

Issue #8185, opened 2024-12-30
`default-branch-button`: Button links to repo root


injust:
### Description

The default branch button links to `https://github.com/typescript-eslint/typescript-eslint/blob/main`.

Unsure if related: When I checked the console after encountering this issue, the error in https://github.com/refined-github/refined-github/issues/8184 appeared.

### How to replicate the issue + URL

https://github.com/typescript-eslint/typescript-eslint/blob/86409083167bee4ba1ad3ebba22785b3a4a262ee/packages%2Feslint-plugin%2Fdocs%2Frules%2Fstrict-boolean-expressions.mdx

### Extension version

24.12.20

### Browser(s) used

Chrome 131

### Token ⚠️

- [X] I set a token in the options page. I understand that some features require a token to work.

fregante:
May be related:

- https://github.com/refined-github/refined-github/issues/8180

fregante:
@injust how did you end up on that URL? It's falling into a check that detects `%2F` and treats it as a branch. The URL I reach is:


> https://github.com/typescript-eslint/typescript-eslint/blob/86409083167bee4ba1ad3ebba22785b3a4a262ee/packages/eslint-plugin/docs/rules/strict-boolean-expressions.mdx

not

> https://github.com/typescript-eslint/typescript-eslint/blob/86409083167bee4ba1ad3ebba22785b3a4a262ee/packages%2Feslint-plugin%2Fdocs%2Frules%2Fstrict-boolean-expressions.mdx

injust:
~~I'm not certain, but I suspect it was from a GitHub search result -- there's really no other reason I would be on that specific commit, other than a search resulting landing me on the most recently indexed version of the file.~~

~~I can't find a way to get to that URL anymore, so I'm thinking at best it's a GitHub bug that has since been fixed, and at worse a very tricky edge case.~~

injust:
@fregante Ugh, I know where it came from.

Try the URL https://github.com/encode/httpcore/blob/master/unasync.py. The file was moved, and the link on the word `moved` goes to `https://github.com/encode/httpcore/blob/55181722f97f73fa42ab505b628586171f3fcf80/scripts%2Funasync.py`.

fregante:
Ohhh I see. I remember something related to URL decoding at some point. 

Thank you for letting me know.

PR #8299, opened 2025-02-24
`default-branch-button` - Click the button return to repo root

kovsu:
Close: #8185 

## Test URLs

https://github.com/typescript-eslint/typescript-eslint/blob/86409083167bee4ba1ad3ebba22785b3a4a262ee/packages%2Feslint-plugin%2Fdocs%2Frules%2Fstrict-boolean-expressions.mdx

## Screenshot

https://github.com/user-attachments/assets/2a91507d-2091-4f9a-a576-387c74f1ad52

### DIFF
--- a/source/github-helpers/github-file-url.ts
+++ b/source/github-helpers/github-file-url.ts
@@ -80,7 +80,7 @@ export default class GitHubFileURL {
 				repository,
 				route,
 				branch,
-				filePath: '',
+				filePath: ambiguousReference[1].replaceAll('%2F', '/'),
 			});
 			return;
 		}


source/github-helpers/github-file-url.ts
```diff
@@ -80,7 +80,7 @@ export default class GitHubFileURL {
 				repository,
 				route,
 				branch,
-				filePath: '',
+				filePath: ambiguousReference[1].replaceAll('%2F', '/'),
```
fregante, line 83: The function has worked for years so it's unlikely that this is the wrong part.

fregante: In fact, we're in the `isRepoRoot` condition. The path must be `''`

Given the TODO above the condition, it's likely that the function must be updated in `github-url-detection` instead

fregante (changes requested):
See comment above

Issue #8218, opened 2025-01-10
`small-user-avatars`: `Quote reply` inputs the user avatar image in the comment


lgoettgens:
### Description

When quoting a comment that contains a user handle, the avatar image that gets added by refined-github gets quoted as well. This is only happening on a limited number of repositories for me, and only started very recently.

![2025-01-10_12-12_1](https://github.com/user-attachments/assets/f8ac5977-781f-45e0-b2f2-2ced15d33d8d)
![2025-01-10_12-12](https://github.com/user-attachments/assets/527d7142-aa2a-4bc3-aab1-ecb18e2e74cd)

The user handle used to get quoted just as the user handle, without doing any markdown magic with links or images.

### How to replicate the issue + URL

Go to the comment https://github.com/gap-system/gap/issues/5902#issuecomment-2582415960 which includes `@lgoettgens`, click on the three dots, and then on `Quote Reply`.

### Extension version

24.12.20

### Browser(s) used

Google Chrome Version 131.0.6778.204 (Official Build) (64-bit)

### Token ⚠️

- [X] I set a token in the options page. I understand that some features require a token to work.

fregante:
That's because it's part of the DOM and GitHub's "Quote reply" feature takes the DOM as is and tries to convert it to markdown.

The only solution for us would be to use `::before {content: URL()}` instead of attaching a new `img`.

fregante:
I can't repro via <kbd>r</kbd> in both Chrome and Safari: 

![merging 2](https://github.com/user-attachments/assets/e92bd7d5-a4ba-4c53-8033-6b45d07c0066)

Nor via "quote reply":

![2](https://github.com/user-attachments/assets/ac5c3a9c-834c-4d6f-849e-d53dd4524340)

fregante:
It looks like this is a bug that only applies to the new view, both via <kbd>r</kbd> and "Quote reply" via dropdown:

- #7856

SunsetTechuila:
> The only solution for us would be to use `::before {content: URL()}` instead of attaching a new `img`.

No, we also can set `aria-hidden` to `"true"` to exclude an image element from the quote

<details>
<summary>Source code</summary>

<img width="1030" height="887" alt="Image" src="https://github.com/user-attachments/assets/9994f2a0-dca1-4872-af37-f522c74fbc48" />

</details>

PR #8300, opened 2025-02-25
`small-user-avatars` - Don't quote avatar into comment field

kovsu:
Close: #8218

## Test URLs

#8218

## Screenshot

https://github.com/user-attachments/assets/1899da15-c701-4518-8f60-3b5c09f725ad


fregante (approved):
giphy

PR #8307, opened 2025-02-26
Fix  handling repo file path got an empty path

kovsu:
Maybe handle it this way for now, and resolve the TODO later.

Close: #8185
Close: #8189

## Test URLs

http://github.com/typescript-eslint/typescript-eslint/blob/86409083167bee4ba1ad3ebba22785b3a4a262ee/packages%2Feslint-plugin%2Fdocs%2Frules%2Fstrict-boolean-expressions.mdx

## Screenshot

https://github.com/user-attachments/assets/540c33f2-1857-4928-9de7-4b694325a788

### DIFF
--- a/source/github-helpers/github-file-url.ts
+++ b/source/github-helpers/github-file-url.ts
@@ -1,4 +1,4 @@
-import {isRepoRoot} from 'github-url-detection';
+import {isRepoRoot, isSingleFile} from 'github-url-detection';
 
 import getCurrentGitRef from './get-current-git-ref.js';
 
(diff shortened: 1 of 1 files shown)


kovsu:
Noooope......

source/github-helpers/github-file-url.ts
```diff
@@ -73,7 +73,7 @@ export default class GitHubFileURL {
 	set pathname(pathname: string) {
 		const [user, repository, route, ...ambiguousReference] = pathname.replaceAll(/^\/|\/$/g, '').split('/');
 		// TODO: `isRepoRoot` uses global state https://github.com/refined-github/refined-github/issues/6637
-		if (isRepoRoot() || (ambiguousReference.length === 2 && ambiguousReference[1].includes('%2F'))) {
+		if (!isSingleFile() && (isRepoRoot() || (ambiguousReference.length === 2 && ambiguousReference[1].includes('%2F')))) {
```
fregante, line 76: One mistake that gets made often here is that GitHubURL does *not* work on the "current page", but it's a constructor that accepts a URL. All of its work must depend on the URL received, not on the current page (aka "global state")

The comment means that it should be `isRepoRoot(url)` if I remember correctly. This change just adds one more incorrect piece of code that can break other features. 

The problem specifically with this part of the class is that we're trying to tell `/user/repo/tree/branch/with/slash` from `/user/repo/tree/branch/folder/file`; `isSingleFile` can't tell these two apart either.

fregante: I suggest just fixing what's causing isRepoRoot to think we're on the root. Likely a selector is broken and it's returning "" for the current path, making it think we're on the repo root.

PR #8325, opened 2025-03-05
`collapsible-content-button` - Improve description

swrobel:
It's nice to be able to search the features for a consistent string and see all features that impact it. Since the other 2 features that modify the text editor mention it in their descriptions, I figured this one should as well. Note how it's currently left out of this search:

Screenshot 2025-03-05 at 3 49 29 PM


fregante (approved):
Thanks!

fregante (approved):

Issue #8324, opened 2025-03-05
Perpetual repetition of the "This issue is highly active" banner


agriyakhetarpal:
### Description

Hi there! I'm currently facing an issue where [the `"This issue is highly active"` banner](https://github.com/refined-github/refined-github/blob/d979a7362307e469ca7c77b438d6f5926dc7663c/source/features/netiquette.tsx#L83-L97) is being prepended to the issue comment box multiple times, even if I switch between the write and preview modes. It doesn't seem to stop, and rather continues multiple times.

Here's a screen recording:

https://github.com/user-attachments/assets/04d7c436-f06d-45cf-b5b1-6ea95b711d05

### How to replicate the issue + URL

`https://github.com/numpy/numpy/issues/26191` (enclosed in backticks so as not to link the issue here). This issue has 152 comments at the time of writing, and is an ideal candidate for replicating the bug.

### Extension version

25.2.26

### Browser(s) used

Arc Browser, version 1.83.0 (59042), based on Chromium Engine Version 133.0.6943.127.

### Token ⚠️

- [x] I set a token in the options page. I understand that some features require a token to work.

fregante:
This should have been fixed by https://github.com/refined-github/refined-github/pull/8298 but I think the fix needs to be applied to multiple parts of the feature.

fregante:
Cc @kovsu

agriyakhetarpal:
Thank you so much for the quick fix!

PR #8298, opened 2025-02-24
`netiquette` - Switch write/preview tabs cause multiple banner

kovsu:
Close: #8239 

## Test URLs

https://github.com/fregante/sandbox/issues/8

## Screenshot

https://github.com/user-attachments/assets/5cf33c68-312d-48d2-8b6a-cc3d8cbc8e15

PR #8326, opened 2025-03-06
`netiquette` - Switch write/preview tabs cause multiple banner

kovsu:
Close: #8324

I forget to add it in #8298.

## Test URLs

https://github.com/numpy/numpy/issues/26191


fregante:
How about the draft banner? It's best to deal with all the similar issues at once rather than one line per PR

kovsu:
@fregante It's on pr view not issue.

PR #8334, opened 2025-03-14
`clean-repo-filelist-actions` - Restore feature

### LINKED ISSUE #8259: `clean-repo-filelist-actions`: doesn't seem to have any effect
### Description

Image

https://github.com/refined-github/sandbox

### Extension version

25.1.24

### Browser(s) used

firefox, zen

### Token ⚠️

kovsu:
Close #8259

## Test URLs

## Screenshot

https://github.com/user-attachments/assets/ae9b34c7-cd79-413b-b997-1ddf6e6c8218

### DIFF
--- a/source/features/clean-repo-filelist-actions.css
+++ b/source/features/clean-repo-filelist-actions.css
@@ -1,0 +1,12 @@
+.rgh-clean-repo-filelist-actions-search {
+	/* Override the value that GitHub sets */
+	min-width: unset !important;
+
+	input {
+		width: 0;
+
+		&:focus {
+			width: 100%;
+		}
+	}
+}
--- a/source/features/clean-repo-filelist-actions.css
+++ b/source/features/clean-repo-filelist-actions.css
@@ -2,11 +2,7 @@
 	/* Override the value that GitHub sets */
 	min-width: unset !important;
 
-	input {
+	input:not(:focus) {
 		width: 0;
-
-		&:focus {
-			width: 100%;
-		}
 	}
 }
--- a/source/features/clean-repo-filelist-actions.tsx
+++ b/source/features/clean-repo-filelist-actions.tsx
@@ -1,5 +1,5 @@
 import React from 'dom-chef';
-import {$} from 'select-dom/strict.js';
+import {$, $optional} from 'select-dom/strict.js';
 import {elementExists} from 'select-dom';
 import * as pageDetect from 'github-url-detection';
 import CodeIcon from 'octicons-plain-react/Code';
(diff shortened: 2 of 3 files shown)


kovsu:
Need to think how to hanlde the search input element. 🤔

fregante:
You can reduce its width when not focused

```css
input:not(:focus) {width: 5em}
```

or

```css
.some-parent:not(:focus-within) {width: 5em}
```

source/features/clean-repo-filelist-actions.tsx
```diff
@@ -53,9 +53,27 @@ function cleanFilelistActions(searchButton: Element): void {
...
+	const codeButtonContent = $('[data-component="text"]', codeButton);
+	codeButtonContent.remove();
+	addTooltipToSummary(codeButton, 'Clone, open or download');
+}
+
 function init(signal: AbortSignal): void {
 	// `.btn` selects the desktop version
-	observe('.btn[data-hotkey="t"]', cleanFilelistActions, {signal});
+	observe('.btn[data-hotkey="t"]', oldCleanFilelistActions, {signal}); // TODO: Drop after August 2025
+	observe('.react-directory-remove-file-icon', cleanFilelistActions, {signal});
```
fregante, line 76: For these minor features feel free to just drop the old code when it makes life easier.

kovsu: I'm not sure the enterprise is same here.

kovsu: But we can drop these code quickly.

kovsu:
I don't konw why the ci failed. 😞

source/features/clean-repo-filelist-actions.tsx
fregante: > I'm not sure the enterprise is same here.

Yeah here's the thing: with so many bugs, we have to limit the time spent on enterprise users. So if fixing it on GitHub.com means breaking the feature for them, so be it.

We try to preserve features, particularly those that are useful to everyday usage, but not at the expense of maintainability. So the suggestion is always: if it's too hard to preserve compat, just drop it.

In this case it seems easy enough, particularly because you just split the observer. This will also be easy to drop in the future.

fregante:
> I don't konw why the ci failed. 😞

not sure either, could be an issue on their end. I'll merge the PR manually when you think it's ready

source/features/clean-repo-filelist-actions.css
```diff
@@ -0,0 +1,12 @@
...
+	/* Override the value that GitHub sets */
+	min-width: unset !important;
+
+	input {
+		width: 0;
+
+		&:focus {
+			width: 100%;
+		}
+	}
```
fregante, line 11: I prefer the negated version as I specified

fregante:
Thanks!

PR #8344, opened 2025-03-19
Partially fix #8343

### LINKED ISSUE #8343: GitHub token is not being saved securely, token saved to autofill, etc
### Description

Multiple issues with this:
1. If I click on the text box for the token, it is revealed in PLAIN TEXT. Tokens should not be visible/accessible to the user after they are saved.
2. I understand that tokens are stored in the backend/config as plain text. Please update this implementation. A possible solution would be to use a GitHub OAuth App (which would also make the user experience smoother).
3. Entering a token is causing it to be saved to my browser's autofill. Screenshot below.
4. (Less important) Clicking on the `Generate Token` link populates it with a default expiration of `No Expiration`. This should be set to 365 days.

Image

URLs:
- chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/assets/options.html

Replicating:
1. Generate a token.
2. Paste it into the box in the settings page with browser autofill enabled
3. Save and reload a few times
4. Clear the text box
5. Trigger autofill (usually by clicking the box, or use developer mode for it)

### Extension version

25.2.26

### Browser(s) used

Google Chrome

### Token ⚠️

zumbiepig:
Partially fixes #8343

### DIFF
--- a/source/options.html
+++ b/source/options.html
@@ -31,7 +31,7 @@
 			<p><strong>Token-less usage is not officially supported.</strong></p>
 			<p>
 				<input
-					type="text"
+					type="password"
 					name="personalToken"
 					spellcheck="false"
 					autocomplete="password"
--- a/source/options/token-validation.tsx
+++ b/source/options/token-validation.tsx
@@ -77,12 +77,6 @@ export default async function initTokenValidation(syncedForm: SyncedForm | undef
 	// Listen to events
 	const field = $('input[name="personalToken"]');
 	field.addEventListener('input', validateToken);
-	field.addEventListener('focus', () => {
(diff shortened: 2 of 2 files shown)


zumbiepig:
@fregante

zumbiepig:
@fregante why did you decline this PR? it seems like a reasonable change

fregante:
I explained why 3 times in your issue

zumbiepig:
@fregante 
> I explained why 3 times in your issue

> Thanks for the report.
> [...]
> 3. Yikes, never seen that before. Can be fixed via autofill attribute. PR welcome.
> [...]

This is the requested PR.

PR #8356, opened 2025-03-28
[DRAFT] Add fn to remove target attribute from a tags

### LINKED ISSUE #8349: Remove unnecessary `target="_blank"` across the site
### Description

https://www.paulingraham.com/target-blank-is-evil.html

GitHub was pretty good at this, they never used `target=_blank` until recently. Now I see it where it's completely unnecessary.

Nobody asked for this, GitHub. In fact there are 3 links on that line and only one opens in a new tab. 🤦‍♂️ 

Image

### Example URLs

https://github.com/refined-github/refined-github/issues/8346#event-16947543136

Feature name should be `same-tab-links` and should target specific links on the page by removing the attribute. e.g.

```js
observe('.thatline a', a => a.removeAttribute('target'))
```

AbhinRustagi:
Closes #8349 

## Test URLs

images: [https://github.com/refined-github/refined-github](https://github.com/refined-github/refined-github)
new issue button: [https://github.com/refined-github/refined-github/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen](https://github.com/refined-github/refined-github/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen)

## Screenshot

**Before**
Screenshot 2025-03-28 at 2 34 43 pm

**After**
Screenshot 2025-03-28 at 2 35 10 pm


source/features/same-tab-links.tsx
```diff
@@ -0,0 +1,24 @@
+// ...existing code...
+import features from '../feature-manager.js';
+import observe from '../helpers/selector-observer.js';
+
+function removeTargetBlank(anchor: HTMLAnchorElement): void {
+	anchor.removeAttribute('target');
+}
+
+function initOnce(): void {
```
fregante, line 9: This can be called init

source/features/same-tab-links.tsx
```diff
@@ -0,0 +1,24 @@
...
+	observe('a[target="_blank"]', removeTargetBlank);
+}
+
+void features.add(import.meta.url, {
+	init: initOnce,
```
fregante, line 14: This is actually called at every page change, even without include/rzclude, so you'll need `init: onetime(init)`. You can copy the exact naming from other similar functions

source/features/same-tab-links.tsx
```diff
@@ -0,0 +1,24 @@
+// ...existing code...
+import features from '../feature-manager.js';
+import observe from '../helpers/selector-observer.js';
+
+function removeTargetBlank(anchor: HTMLAnchorElement): void {
+	anchor.removeAttribute('target');
+}
+
+function initOnce(): void {
+	observe('a[target="_blank"]', removeTargetBlank);
```
fregante, line 10: As mentioned in my issue, I'd like to make this more specific, pointing to some specific links, not all links. You'll have to find the correct selector for that 
```suggestion
	observe([
     'the closed event a[target="_blank"]',
   ], removeTargetBlank);
```

source/features/same-tab-links.tsx
```diff
@@ -0,0 +1,24 @@
...
+void features.add(import.meta.url, {
+	init: initOnce,
+});
+
+/*
+
+Test URLs:
+
+images: https://github.com/refined-github/refined-github
+new issue button: https://github.com/refined-github/refined-github/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen
```
fregante, line 22: This was just an example of situation where it's not used and thus completely unnecessary, so we don't need to change it.

The one that needs it is in the "example URLs" section of my issue

fregante (commented):
Thank you for the PR! You can run `npm test` locally to see what needs to be done for CI to pass.

source/features/same-tab-links.tsx
AbhinRustagi: Hm, so I think we'd need a list of those specific ones or a modifiable list for that, yeah?

fregante: Yes, but for now just one is fine, we can extend the list as needed.

fregante:
> You can run `npm test` locally to see what needs to be done for CI to pass

Also see https://github.com/refined-github/refined-github/wiki/Contributing, there are 2 commands that will help you.

source/features/same-tab-links.tsx
AbhinRustagi: Which specific anchor tag are you referring to?
None of these open in a new tab for me :|

AbhinRustagi:
@fregante Couldn't figure out the readme parser issue, would appreciate direction here!

source/features/same-tab-links.tsx
fregante: 1. See "example URLs" section
2. Click the link
3. See link that matches the screenshot right above the "example URLs" section

fregante:
> > You can run `npm test` locally to see what needs to be done for CI to pass
> 
> Also see [Wiki: Contributing](https://github.com/refined-github/refined-github/wiki/Contributing), there are 2 commands that will help you.

It seems that you didn't read my comments, so it's easier for me to write the feature right away. This comment contained the solution to "the readme parser issue"

PR #8369, opened 2025-03-31
`github-actions-indicators` - Open dropdown on icon click

fregante:
- Closes https://github.com/refined-github/refined-github/issues/6443

## Test URLs


source/features/github-actions-indicators.tsx
```diff
@@ -138,12 +151,24 @@ async function init(signal: AbortSignal): Promise<false | void> {
 	observe('a.ActionListContent', addIndicators, {signal});
 }
 
+function openRunWorkflow(): void {
+	removeHashFromUrlBar();
+	const dropdown = $('details[data-deferred-details-content-url*="/actions/manual?workflow="]');
```
fregante, line 156:
```suggestion
	// Note that the attribute is removed after the first opening, so the selector only matches it once
	const dropdown = $('details[data-deferred-details-content-url*="/actions/manual?workflow="]');
```

PR #8370, opened 2025-03-31
`clean-readme-url` - Improve on Safari and Firefox

fregante:
The navigation API is still nowhere to be found, so let's just use interval on non-Chrome browsers

https://developer.mozilla.org/en-US/docs/Web/API/Navigation_API#api.navigation

### DIFF
--- a/source/features/clean-readme-url.tsx
+++ b/source/features/clean-readme-url.tsx
@@ -12,7 +12,17 @@ function maybeCleanUrl(event?: NavigateEvent): void {
 
 function init(signal: AbortSignal): void {
 	maybeCleanUrl();
-	globalThis.navigation?.addEventListener('navigate', maybeCleanUrl, {signal});
+	let interval: NodeJS.Timeout;
+	if (globalThis.navigation) {
+		navigation.addEventListener('navigate', maybeCleanUrl, {signal});
+	} else {
+		interval = setInterval(() => {
+			maybeCleanUrl();
+		}, 1000);
+		signal.addEventListener('abort', () => {
+			clearInterval(interval);
+		});
+	}
 }
 
 void features.add(import.meta.url, {


source/features/clean-readme-url.tsx
```diff
@@ -12,7 +12,17 @@ function maybeCleanUrl(event?: NavigateEvent): void {
...
-	globalThis.navigation?.addEventListener('navigate', maybeCleanUrl, {signal});
+	let interval: NodeJS.Timeout;
+	if (globalThis.navigation) {
+		navigation.addEventListener('navigate', maybeCleanUrl, {signal});
+	} else {
+		interval = setInterval(() => {
+			maybeCleanUrl();
+		}, 1000);
+		signal.addEventListener('abort', () => {
+			clearInterval(interval);
```
kovsu, line 23: Does this signal has been aborted ? The interval is still running.

https://github.com/user-attachments/assets/9f3da681-639e-400a-8224-8ffbacd3a99c

kovsu: BTW I use this code to test in my Chrome.

```ts
	maybeCleanUrl();
	// let interval: NodeJS.Timeout;
	// if (globalThis.navigation) {
	// 	navigation.addEventListener('navigate', maybeCleanUrl, {signal});
	// } else {
	const interval: NodeJS.Timeout = setInterval(() => {
		maybeCleanUrl();
		console.log('interval');
	}, 1000);
	signal.addEventListener('abort', () => {
		clearInterval(interval);
	});
	// }
}
```

fregante: The signal is the navigation signal, when you change page. The URL update doesn't count as a page change. 

The interval is specifically because GitHub keeps adding the URL parameter while you switch readme tabs ("readme", "license", etc) which doesn't qualify as a page change.

Issue #8309, opened 2025-02-27
`quick-mention` causes slowdown


kilianyp:
### Description

In the comment box for a PR, after tagging a user @/xyz, typing becomes very unresponsive, e.g. there's a sudden delay between typing a letter and it appearing.

I used the "identify feature" functionality and it seems to be the quick-mention feature. After disabling it, I could confirm that the issue is gone for me

### How to replicate the issue + URL

https://github.com/huggingface/transformers/pull/33962

- type (should be normal)
- tag any user
- type again (now there's a slowdown)

⚠ I didn't set a token 

### Extension version

25.2.17

### Browser(s) used

Google chrome 132.0.6834.160

### Token ⚠️

- [x] I set a token in the options page. I understand that some features require a token to work.

fregante:
The feature does have two has() selectors, which are known to cause slowdowns in this fashion. 

See https://github.com/refined-github/refined-github/pull/8225/files

Please keep quick-mention disabled and let me know if you still experience slowdowns. 

I had some serious perf issues when the PR sidebar dropdowns open, but I'm on an older version (Safari)

mpsijm:
I can confirm this issue. It also appears to happen after I start typing emoji (e.g. `:smile:`), not only with mentions. Input delay appears to be unaffected until the "mention dropdown" has shown once, and even after the dropdown is closed (because the emoji or mention is complete), there is a significant input delay for every character typed.

I captured a performance profile, and it shows that most of the time is spent in "Recalculating style":

![Image](https://github.com/user-attachments/assets/3a0da9f9-cd51-4cc2-aa3f-19f80ad7ff35)

joshprzybyszewski:
When I pulled a performance profile, it looks similar to @mpsijm 's. The stack trace on the "recalculate style" brought me [here](https://github.com/iansan5653/dom-input-range/blob/main/src/input-style-clone-element.ts#L235-L237). Is there a usage of InputRange in refined github that could be at play?

fregante:
It's probably the has() selector again. The profiler just points you to the latest piece of code that pushes the browser into that selector. The difference is that last time it was Firefox, so I'm not sure about it now.

PR #8373, opened 2025-04-01
`quick-mention` - Restore feature on issues

kovsu:
## Test URLs

Any issue

## Screenshot

https://github.com/user-attachments/assets/17f0c5d1-32d4-4f9c-bd1d-c6ae260324ba

### DIFF
--- a/source/features/quick-mention.tsx
+++ b/source/features/quick-mention.tsx
@@ -75,7 +75,7 @@ function add(avatar: HTMLElement): void {
 		}
 	}
 
-	if (debug && isOldView) {
+	if (debug) {
 		timelineItem.style.border = 'solid 5px green';
 	}
 
--- a/source/features/quick-mention.tsx
+++ b/source/features/quick-mention.tsx
@@ -75,9 +75,9 @@ function add(avatar: HTMLElement): void {
 		}
 	}
 
-	// if (debug) {
-	// 	timelineItem.style.border = 'solid 5px green';
-	// }
+	if (debug && isOldView) {
+		timelineItem.style.border = 'solid 5px green';
+	}
 
 	// Wrap avatars next to review events so the inserted button doesn't break the layout #4844
 	if (avatar.classList.contains('TimelineItem-avatar')) {
--- a/source/features/quick-mention.tsx
+++ b/source/features/quick-mention.tsx
@@ -75,9 +75,9 @@ function add(avatar: HTMLElement): void {
 		}
 	}
 
-	if (debug) {
-		timelineItem.style.border = 'solid 5px green';
-	}
+	// if (debug) {
+	// 	timelineItem.style.border = 'solid 5px green';
+	// }
 
 	// Wrap avatars next to review events so the inserted button doesn't break the layout #4844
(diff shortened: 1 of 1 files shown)


source/features/quick-mention.tsx
```diff
@@ -75,9 +75,9 @@ function add(avatar: HTMLElement): void {
 		}
 	}
 
-	if (debug) {
-		timelineItem.style.border = 'solid 5px green';
-	}
+	// if (debug) {
+	// 	timelineItem.style.border = 'solid 5px green';
+	// }
```
fregante, line 80: Debug is already false

source/features/quick-mention.tsx
```diff
@@ -123,7 +124,7 @@ async function init(signal: AbortSignal): Promise<void> {
 			a.TimelineItem-avatar
 		):not([href="/${getUsername()!}"])
 	`,
-		`[data-testid="issue-viewer-container"]:has(${fieldSelector[1]}) [class^="LayoutHelpers-module__timelineElement"] > img[data-component="Avatar"]:not([alt="${getUsername()!}"])`,
+		`[data-testid="issue-viewer-container"]:has(${fieldSelector[1]}) [class^="LayoutHelpers-module__timelineElement"] a:not([href="/${getUsername()!}"])`,
```
fregante, line 127: Can you also change this selector to drop the `:has()`? `[data-testid="issue-viewer-container"]` is large and :has() causes slowdowns:

- #8309 

Separate PR is also ok

kovsu: I'd like to open a new pr to do it.

source/features/quick-mention.tsx
kovsu: In new view, we don't have the `timelineItem`.
```tsx
const isOldView = Boolean(timelineItem);
```

And we do this for old view. So I think we don't need this code.
```tsx
if (isOldView) {
		if (debug) {
			timelineItem.style.border = 'solid 5px red';
		}
}
```

fregante: The code is already behind a manual condition, it doesn't cause any issues if you don't manually go and enable it already. It's dead code, no need to comment it out.

It can be dropped if we drop the timelineItem variable altogether in the future, but that's not happening in this PR

fregante (approved):
Can be merged after:

- uncommenting code
- making sure it still works on PRs

source/features/quick-mention.tsx
kovsu: This is an error when we enable `debug` I mean. So if you want to keep this code, I will add a condition in `if`.

source/features/quick-mention.tsx
```diff
@@ -75,7 +75,7 @@ function add(avatar: HTMLElement): void {
 		}
 	}
 
-	if (debug) {
+	if (debug && isOldView) {
```
fregante, line 78:
```suggestion
	if (debug) {
```


What's the point of enabling debugging if it silently does nothing? The variable does one thing, you set it to true when loading the old view exclusively. Since it doesn't work on the new view, you either don't use it or fix it so that it does work.

PR #8375, opened 2025-04-01
`quick-mention` - Improve feature to avoid web lagging

kovsu:
Close: #8309 

## Test URLs
https://github.com/huggingface/transformers/pull/33962

## Screenshot

https://github.com/user-attachments/assets/d76c36ab-796e-4207-ac86-2ebe311ce294

### DIFF
--- a/source/features/quick-mention.css
+++ b/source/features/quick-mention.css
@@ -17,6 +17,11 @@
 .rgh-quick-mention.react-view {
 	position: static !important;
 	translate: none;
+
+	&::after {
+		bottom: unset;
+		transform: none;
+	}
 }
 
 /* Show icon in black if it's not hovered/focused, instead of the blue color caused by .btn-link */
--- a/source/features/quick-mention.tsx
+++ b/source/features/quick-mention.tsx
@@ -18,6 +18,24 @@ const fieldSelector = [
 	'#react-issue-comment-composer textarea',
 ] as const;
 
+// Old Issue View and PR View
+// `:first-child` avoids app badges #2630
+// Avatars next to review events aren't wrapped in a <div> #4844
+const prCommentSelector = `
+	.js-quote-selection-container
+	:is(
+		div.TimelineItem-avatar > [data-hovercard-type="user"]:first-child,
+		a.TimelineItem-avatar
+	):not([href="/${getUsername()!}"])
+`;
+
+const issueCommentSelector = [
+	// React Issue View
+	`[data-testid="issue-viewer-comments-container"] [class^="LayoutHelpers-module__timelineElement"] a:not([href="/${getUsername()!}"])`,
+	// React Issue View (first comment)
+	`[data-testid="issue-viewer-issue-container"] a[class^="Avatar-module__avatarLink"]:not([href="/${getUsername()!}"])`,
+];
+
 function prefixUserMention(userMention: string): string {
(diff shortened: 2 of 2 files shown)


fregante:
If you record smaller windows and without the extra border, they can be small enough to be uploaded to GitHub 👍 see the videos in my PRs

source/features/quick-mention.css
```diff
@@ -17,6 +17,11 @@
 .rgh-quick-mention.react-view {
 	position: static !important;
 	translate: none;
+
+	&::after {
+		bottom: unset;
+		transform: none;
+	}
 }
```
fregante, line 25: What is this targeting? It looks like we're undoing GitHub's style

source/features/quick-mention.tsx
```diff
@@ -108,23 +111,27 @@ async function init(signal: AbortSignal): Promise<void> {
 		return;
 	}
 
-	delegate('button.rgh-quick-mention', 'click', mentionUser, {signal});
+	const isExist = elementExists(fieldSelector);
+	if (!isExist) {
+		return;
+	}
```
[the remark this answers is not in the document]
fregante: Yes this solution is not good, the we use the observer specifically to catch elements that load later. This condition is run at the beginning of the document and on slower connections it will always be false. 

It seems we need to start layering observers:

1. First observe the field selector
2. Once found, disable the first observer and start an observer for the avatars


Technically elementReady() could be used to replace this elementExists check, but that function stops at dom-ready (because it's very inefficient to keep running afterwards).

I wrote an observer-based replacement but never finished it and I think it doesn't work. So for now you might need `observe(field, () => observe(avatars, addButton))` (plus the signal to stop the first observer, etc

source/features/quick-mention.tsx
```diff
@@ -110,22 +131,19 @@ async function init(signal: AbortSignal): Promise<void> {
...
-		.js-quote-selection-container:has(${fieldSelector[0]})
-		:is(
-			div.TimelineItem-avatar > [data-hovercard-type="user"]:first-child,
-			a.TimelineItem-avatar
-		):not([href="/${getUsername()!}"])
-	`,
-		`[data-testid="issue-viewer-container"]:has(${fieldSelector[1]}) [class^="LayoutHelpers-module__timelineElement"] a:not([href="/${getUsername()!}"])`,
-	], add, {signal});
+	const filedSignal = new AbortController();
+	observe(fieldSelector, field => {
```
fregante, line 135:
```suggestion
	observe(fieldSelector, onetime(field => {
```
Just to ensure it's never called twice. 

---

Alternatively, you can change this pattern to something like:

```JS
const controller = new AbortController();
const field: HTMLElement | undefined = await new Promise(resolve => {
  observe(fieldSelector, resolve, AbortSignal.any(etc))
  onAbort(controller, () => resolve())
});
controller.abort()

//then continue with the standard observer
```

I think this pattern is more self contained and can be replaced easily once I finish up my element-ready replacement. 

---

But not necessary

fregante (approved):
Should be good, thanks!

If you have a chance, can you review my PRs too? So I can merge them and release a new version.

kovsu:
> Should be good, thanks!
> 
> If you have a chance, can you review my PRs too? So I can merge them and release a new version. 

Ok, I will try it.

PR #8389, opened 2025-04-08
`small-user-avatar` - Support clipped mentions in hovercards

kovsu:
## Test URLs

https://github.com/refined-github/refined-github/pull/8360

## Screenshot

Before:

After:


source/features/small-user-avatars.tsx
```diff
@@ -23,10 +23,10 @@ function addAvatar(link: HTMLElement): void {
 	);
 }
 
-function addMentionAvatar(link: HTMLElement): void {
+function addMentionAvatar(link: HTMLLinkElement): void {
```
fregante, line 26: This is `<link>`

You're looking for HTMLAnchorElement

source/github-helpers/get-user-avatar.ts
```diff
@@ -1,12 +1,18 @@
 import {$optional} from 'select-dom/strict.js';
 import * as pageDetect from 'github-url-detection';
 
-export default function getUserAvatar(username: string, size: number): string | void {
-	const cleanName = username.replace('[bot]', '');
+export default function getUserAvatar(username: string, size: number, href?: string): string | void {
```
fregante, line 4: The function is already receiving the username. If it's wrong, it should be fixed *outside* this function.

fregante (changes requested):

source/features/small-user-avatars.tsx
```diff
@@ -23,10 +31,21 @@ function addAvatar(link: HTMLElement): void {
 	);
 }
 
-function addMentionAvatar(link: HTMLElement): void {
+function addMentionAvatar(link: HTMLAnchorElement): void {
 	assertNodeContent(link.firstChild, /^@/);
 	const username = link.textContent.slice(1);
```
fregante, line 36: Why not just use the URL here?

source/features/small-user-avatars.tsx
```diff
@@ -23,10 +31,21 @@ function addAvatar(link: HTMLElement): void {
...
+	// First use link content, then href
+	let avatarUrl = getAvatarUrlSafely(username, 16);
+
+	if (!avatarUrl) {
+		const username = link.href.split('/').pop()!;
+		avatarUrl = getAvatarUrlSafely(username, 20);
+	}
+
+	if (!avatarUrl) {
+		return;
```
fregante, line 47: There's no way an avatar doesn't exist.

source/features/small-user-avatars.tsx
```diff
@@ -7,6 +7,14 @@ import observe from '../helpers/selector-observer.js';
 import getUserAvatarURL from '../github-helpers/get-user-avatar.js';
 import './small-user-avatars.css';
 
+function getAvatarUrlSafely(username: string, size: number): string {
+	try {
+		return getUserAvatarURL(username, size) || '';
```
fregante, line 12: Why would this throw ? The username is just a string. Any string will product a URL without errors. If it throws, then it should be fixed.

source/features/small-user-avatars.tsx
kovsu: The operation URL might be more complex than text content, I think.

source/features/small-user-avatars.tsx
```diff
@@ -23,9 +22,8 @@ function addAvatar(link: HTMLElement): void {
 	);
 }
 
-function addMentionAvatar(link: HTMLElement): void {
-	assertNodeContent(link.firstChild, /^@/);
-	const username = link.textContent.slice(1);
+function addMentionAvatar(link: HTMLAnchorElement): void {
+	const username = link.href.split('/').pop()!;
```
fregante, line 26: Is every link we select exactly `/$user` or do we get `/apps/$name` too?

fregante (approved):
May be good, some questions though

source/features/small-user-avatars.tsx
kovsu: Can you give me a link? I can't find this situation.

fregante: Mine was a question, I don't know if it exists

kovsu: I don't know but it's fine for test URLs. Maybe we can slove it when we meet it.

source/github-helpers/get-user-avatar.ts
```diff
@@ -5,8 +5,7 @@ export default function getUserAvatar(username: string, size: number): string |
 	const cleanName = username.replace('[bot]', '');
 
 	if (/[^\w-]/.test(cleanName)) {
-		// TODO: December 2024: Turn into TypeError once we're sure it's not breaking anything
-		console.error(`Expected a username, got ${cleanName}`);
+		throw new TypeError (`Expected a username, got ${cleanName}`);
```
fregante, line 8: 🙏

fregante (approved):

source/github-helpers/get-user-avatar.ts
kovsu: Do you mean revert this or what?

fregante: It's a "thank you for addressing the TODO" 😁

kovsu: Oh, get it.

PR #8390, opened 2025-04-08
`quick-mention` - Exclude apps

kidonng:
Motivation is to exclude Copilot but could catch more cases.

I asked Copilot for a review here so we have a live example 😉 

## Test URLs

https://togithub.com/dotnet/maui/pull/28831#pullrequestreview-2747076622

## Screenshot

Before:

Screenshot 2025-04-08 at 5 45 53 PM

After: Copilot is excluded

### DIFF
--- a/source/features/quick-mention.tsx
+++ b/source/features/quick-mention.tsx
@@ -122,7 +122,10 @@ async function init(signal: AbortSignal): Promise<void> {
 		:is(
 			div.TimelineItem-avatar > [data-hovercard-type="user"]:first-child,
 			a.TimelineItem-avatar
-		):not([href="/${getUsername()!}"])
+		):not(
+			[href="/${getUsername()!}"],
+			[href^="/apps/"]
+		)
 	`,
 		`[data-testid="issue-viewer-container"]:has(${fieldSelector[1]}) [class^="LayoutHelpers-module__timelineElement"] a:not([href="/${getUsername()!}"])`,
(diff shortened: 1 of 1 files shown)


fregante:
@Copilot does this do anything?

fregante:
Ok copilot doesn't, but @dependabot does

Screenshot


Screenshot


Technically this is an "exception" since they have both `/apps/dependabot` and the actual user `/dependabot`.

What do we do? I think it's easier to leave it as is and have a mention that does nothing, rather than receive a bug report for "`quick-mention` doesn't appear on bots"

kidonng:
I forgot this kind of thing exists 😅 not worth the effort then.

As a side note `linkify-user-labels` is in the same "uselessly targets bots but fine" boat.

fregante:
I think that feature only affects comments added by bots that committed to the repo. In which case the link does point to a list of real commits.

fregante:
GitHub lies 🥲 

Screenshot

fregante:
While it affects us, I think that we're not wrong for showing the link. For dependabot's for example the link does have commits: https://github.com/dotnet/maui/commits?author=dependabot%5Bbot%5D

Screenshot

I say that this is a GitHub bug (the UI says "previously committed" but it's not true) and it's not worth addressing.

There's also a chance that the bot _did_ commit to the repo, but the commit does not appear on `main`

PR #8407, opened 2025-04-14
Style fill stared icon and code in commit title

kovsu:
## Screenshot

| Before | After |
| ---- | ---- |
|  |  |
|  |  |

### DIFF
--- a/source/refined-github.css
+++ b/source/refined-github.css
@@ -180,11 +180,3 @@ div[data-target='readme-toc.content'] div.blob-wrapper img {
 .octicon-star-fill {
 	color: var(--button-star-iconColor);
 }
-
-[data-testid="commit-row-item"] code  {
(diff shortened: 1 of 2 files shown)


source/refined-github.css
```diff
@@ -176,3 +176,15 @@ div[data-target='readme-toc.content'] div.blob-wrapper img {
...
+	color: var(--button-star-iconColor);
+}
+
+[data-testid="commit-row-item"] code  {
+	padding: 2px var(--base-size-4);
+	font-size: .9em;
+	line-height: 1;
+	background-color: var(--bgColor-neutral-muted);
+	border-radius: var(--borderRadius-medium);
+}
```
kovsu, line 190: Do we have refined-github only css? Maybe we can move this to it.

fregante (approved):

PR #8410, opened 2025-04-17
`update-pr-from-base-branch` - Fix wrong position when check failed

kovsu:
## Test URLs

#8407

## Screenshot

Before: 

CleanShot 2025-04-17 at 15 06 03

After: 

image

### DIFF
--- a/source/features/update-pr-from-base-branch.tsx
+++ b/source/features/update-pr-from-base-branch.tsx
@@ -115,9 +115,11 @@ async function addButton(mergeBar: Element): Promise<void> {
 			</div>,
 		);
 	} else {
+		// We need to create a new row when `Checks` is present
+		const checkFailed = $optional('[aria-label="Checks"]');
 		// Old view draft PRs require a new row to display the button
 		// https://github.com/refined-github/refined-github/pull/8193#discussion_r1908581612
-		mergeBar.before(createMergeabilityRow({
+		(checkFailed ?? mergeBar).before(createMergeabilityRow({
 			className: 'rgh-update-pr-from-base-branch-row',
 			action: createButton(),
 			icon: <CheckIcon />,


source/features/update-pr-from-base-branch.tsx
```diff
@@ -115,9 +115,11 @@ async function addButton(mergeBar: Element): Promise<void> {
 			</div>,
 		);
 	} else {
+		// We need to create a new row when `Checks` is present
+		const checkFailed = $optional('[aria-label="Checks"]');
```
kovsu, line 119: If we add this in `mergeabilityRow`, we will get this. So obviously we need a new row.

image

fregante: This check seems rather arbitrary, the issue is that we use .before() on the wrong element. I think just `mergeBar` needs to be update and then maybe use .prepend()

Also after this PR it would be great to see `pr-commit` working again

fregante: Feel free to drop old code and just work on the new version. The widget is already complex as is, we don't need to be slowed down by GitHub Enterprise.

kovsu: We don't have this `[aria-label="Conflicts"]` if check faild. But if check passed, we will get `[aria-label="Conflicts"]`.

fregante: I'm seeing it in https://github.com/mozilla/addons-linter/pull/5426, I don’t think the two things are related


Screenshot 1


Screenshot 2

kovsu:
| | Check Failed | Check Passed |
| ---- | ----- | ----- |
| No Conflict |  CleanShot 2025-04-18 at 08 38 36 | CleanShot 2025-04-18 at 08 37 50 |

When the PR has conflicts the feature will return early.

source/features/update-pr-from-base-branch.tsx
kovsu: > I don’t think the two things are related

I mean if there has `Conflict` row,  we just need to add a update branch button here. But if we don't have this row, we have to creat a new row here.

fregante: If there's a "resolve conflicts" button, the feature must be disabled. That button is already an "update branch" config

kovsu: https://github.com/user-attachments/assets/8f656e77-e5a1-4dcc-9247-2c99c9c42c23

fregante:
YOLO sorry for the wait 🥹

PR #8421, opened 2025-04-25
`repo-header-info` - Restore feature

kovsu:
Same as #8417 

## Test URLs

Any Repo


fregante:
Can you search across the repo in case there are other features too? It's worth doing this every time a selector needs to be updated

fregante:
Sometimes selectors are not exactly the same, like they use the li:first child but target a different parent

fregante (approved):

PR #8422, opened 2025-04-27
feat(`copy-on-y`): enhance clipboard functionality for URLs

Abdulrahman-Sallam:
Details:
-
Add functionality to copy the current page URL along with its title to the clipboard when the 'y' key is pressed, provided the target element is not editable. This improves user experience by allowing easy sharing of links with context.

Implementation Details:
-
- Checks for clipboard API availability and permission before attempting to copy.
- Supports both plain text and HTML formats for the copied content.

Testing Notes:
-
- Verified functionality by pressing 'y' on various pages to ensure the correct URL and title are copied to the clipboard.

### DIFF
--- a/source/features/copy-on-y.tsx
+++ b/source/features/copy-on-y.tsx
@@ -4,6 +4,38 @@ import {isEditable} from '../helpers/dom-utils.js';
 async function handler({key, target}: KeyboardEvent): Promise<void> {
 	if (key === 'y' && !isEditable(target)) {
 		const url = location.href;
+		// make the URL a permalink with title.text being the text of the link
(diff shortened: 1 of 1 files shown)


fregante:
Did you write this with AI?

Abdulrahman-Sallam:
> Did you write this with AI?
Yes, I used Copilot for most of the code.
Is that a problem?

fregante:
The problem is that this was not discussed or desired. It's also very verbose as the same can be achieved in 3 lines of code considering that `navigator.clipboard.writeText` is already there without permission checking and error catching.

PR #8430, opened 2025-05-03
`bugs-tab` - Support Bug type

### LINKED ISSUE #8227: Support for Github Issue Types in `bugs-tab`
### Description

One of the nice features of refined-github is that it has the cool "Bugs" tab:

Image

which I believe is `bugs-tab`. However, recently GitHub has enabled for GA (or at least many repos) the Issue types:

https://github.blog/changelog/2025-01-13-evolving-github-issues-public-preview/

And one of my orgs/repos was one that got it, and as [you'll see](https://github.com/GEOS-ESM/MAPL/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen) I have replaced our old "Bug" label to use the new "Bug" issue type:

Image

But on that page, I have `bugs-tab` enabled, but no "Bugs" tab is made. I believe this is because `bugs-tab` looks only at labels.

So I was wondering, can we get support for both Bug-as-label and Bug-as-issue-type in `bugs-tab`?

(And I suppose, would there a be a reason to make a `Feature` and `Task` tab? Those are the other types of Issues now available.)

### Example URLs

https://github.com/GEOS-ESM/MAPL/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen

Pixxle:
Closes #8227

Not the best at crafting regexp, so would appreciate suggestions on how to improve it if possible, but all existing tests on queryPartsRegExp are still passing.

## Test URLs
https://github.com/refined-github/refined-github/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen+%28label%3Abug+OR+type%3ABug%29

Unfortunately don't have any public repository available to test the issue type bug on. 

## Screenshot

image

image

### DIFF
--- a/source/features/bugs-tab.tsx
+++ b/source/features/bugs-tab.tsx
@@ -24,22 +24,22 @@ type Bugs = {
 async function countBugs(): Promise<Bugs> {
 	const {repository} = await api.v4(CountBugs);
 
-	const issuesBugCount = repository.issues.totalCount ?? 0;
+	const bugTypeCount = repository.issues.totalCount ?? 0;
 
 	// Prefer native "bug" label
 	for (const label of repository.labels.nodes) {
 		if (label.name === 'bug') {
-			return {label: 'bug', count: (label.issues.totalCount ?? 0) + issuesBugCount};
+			return {label: 'bug', count: (label.issues.totalCount ?? 0) + bugTypeCount};
 		}
 	}
 
 	for (const label of repository.labels.nodes) {
 		if (isBugLabel(label.name)) {
-			return {label: label.name, count: (label.issues.totalCount ?? 0) + issuesBugCount};
+			return {label: label.name, count: (label.issues.totalCount ?? 0) + bugTypeCount};
 		}
 	}
 
-	return {label: '', count: issuesBugCount};
+	return {label: '', count: bugTypeCount};
 }
 
 const bugs = new CachedFunction('bugs', {
--- a/source/features/bugs-tab.tsx
+++ b/source/features/bugs-tab.tsx
@@ -24,24 +24,22 @@ type Bugs = {
 async function countBugs(): Promise<Bugs> {
 	const {repository} = await api.v4(CountBugs);
 
-	if (repository.issues.totalCount > 0) {
-		return {label: 'bug', count: repository.issues.totalCount};
-	}
+	const issuesBugCount = repository.issues.totalCount ?? 0;
 
 	// Prefer native "bug" label
 	for (const label of repository.labels.nodes) {
 		if (label.name === 'bug') {
-			return {label: 'bug', count: label.issues.totalCount ?? 0};
+			return {label: 'bug', count: (label.issues.totalCount ?? 0) + issuesBugCount};
 		}
 	}
 
 	for (const label of repository.labels.nodes) {
 		if (isBugLabel(label.name)) {
-			return {label: label.name, count: label.issues.totalCount ?? 0};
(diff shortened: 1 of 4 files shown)


source/features/bugs-tab.tsx
```diff
@@ -49,7 +51,7 @@ const bugs = new CachedFunction('bugs', {
 
 async function getSearchQueryBugLabel(): Promise<string> {
 	const {label} = await bugs.getCached() ?? {};
-	return 'label:' + SearchQuery.escapeValue(label ?? 'bug');
+	return `(label:${label ?? 'bug'} OR type:Bug)`;
```
fregante, line 54: Don't drop the escape

source/features/bugs-tab.tsx
```diff
@@ -24,20 +24,22 @@ type Bugs = {
 async function countBugs(): Promise<Bugs> {
 	const {repository} = await api.v4(CountBugs);
 
+	const issuesBugCount = repository.issues.totalCount ?? 0;
+
 	// Prefer native "bug" label
```
fregante, line 29: Can you simplify this setup? I'd rather see:

```JS
const label = repository.labels.find( etc ) ?? 'bug'
const count = repository.labels.find( etc ) ?? 0
return {label, count: count + issuesBugCount}
```

source/features/bugs-tab.tsx
```diff
@@ -24,20 +24,22 @@ type Bugs = {
 async function countBugs(): Promise<Bugs> {
 	const {repository} = await api.v4(CountBugs);
 
+	const issuesBugCount = repository.issues.totalCount ?? 0;
```
fregante, line 27:
```suggestion
	const bugTypeCount = repository.issues.totalCount ?? 0;
```

source/github-helpers/search-query.ts
```diff
@@ -1,4 +1,4 @@
-const queryPartsRegExp = /(?:[^\s"]|"[^"]*")+/g;
+const queryPartsRegExp = /[^\s"()]+:[^"\s()]*(?:"[^"]*")?|\([^)]*\)|"[^"]*"|[^\s"():]+/g;
```
fregante, line 1: Oof this is a mouthful. Would be good to add more tests with and without parents to see how it behaves

fregante (commented):
Thank you for the PR!

source/github-helpers/search-query.ts
Pixxle: Yeah, the regex isn't beautiful, I'll add tests for different cases en scenarios!

Pixxle:
> Thank you for the PR!

Of course, I'll take a look at all your suggestions in the coming week!

source/features/bugs-tab.tsx
```diff
@@ -49,7 +57,7 @@ const bugs = new CachedFunction('bugs', {
 
 async function getSearchQueryBugLabel(): Promise<string> {
 	const {label} = await bugs.getCached() ?? {};
-	return 'label:' + SearchQuery.escapeValue(label ?? 'bug');
+	return `(label:${SearchQuery.escapeValue(label ?? 'bug')} OR type:Bug)`;
```
fregante, line 60: My main concern with this is that selecting the `bug` label will no longer activate the "bugs" tab

Screenshot

But that's ok because now that's not enough to show all bugs in the repo. Thanks GitHub.

fregante (approved):
YOLO

fregante:
@Pixxle thank you for working on this?

PR #8435, opened 2025-05-06
`show-names` - treat `github-apps` as a bot

### LINKED ISSUE #8433: `show-names` fails if one username errors
### Description

Show names is hitting a bot user without a login, causing all of show names to fail. Happy to open a PR, but it's not clear to me if we should filter out the GHE bot user or not fail on GraphQL error or both. At a quick glance of show names, I didn't see where we filtered out bot users, outside of `ghost`.

Sample response where `not_a_login` is the bot user without a login. The path to their account is `<ghe_base>/github-apps/not_a_login`. They also have the `bot` tag next to their comment.

`{"data":{"_bob":{"name":"Bob"},"_not_a_login":null},"errors":[{"type":"NOT_FOUND","path":["_not_a_login"],"locations":[{"line":1,"column":84}],"message":"Could not resolve to a User with the login of 'not_a_login'."}]}`

Couldn't find a public URL to test it on, as bots don't follow the same format.

### Extension version

25.4.8

### Browser(s) used

Firefox 

### Token ⚠️

busches:
Closes #8433 

Documentation references to `/github-apps`:
- https://docs.github.com/en/enterprise-server@3.16/apps/using-github-apps/installing-a-github-app-from-a-third-party
- https://docs.github.com/en/enterprise-server@3.16/apps/sharing-github-apps/sharing-your-github-app
- https://docs.github.com/en/enterprise-server@3.16/apps/creating-github-apps/about-creating-github-apps/migrating-oauth-apps-to-github-apps

No test url as it applies to private GHE only.

### DIFF
--- a/source/github-helpers/selectors.ts
+++ b/source/github-helpers/selectors.ts
@@ -173,6 +173,7 @@ const botNames = [
 	'scala-steward',
 	'weblate',
 	'apps', // Matches any `/apps/*` URLs
+	'github-apps', // GHE apps
 ] as const;
 
 const botAttributes = botNames.map(bot => `[href^="/${bot}"]`).join(', ');


fregante (approved):

fregante:
Thank you!

BoomchainLabs (approved):
Lgtm

Issue #8367, opened 2025-03-31
Replace Stylelint with Biome


fregante:
We use Prettier for formatting and Stylelint for linting. I never liked the latter, it's time to drop it.

Luckily Biome can act as a CSS linter so let's set it up as a Stylelint replacement.

---

In the future maybe it can replace Prettier too, but not yet:

- https://github.com/refined-github/refined-github/pull/8082

As a secondary request, maybe also partially replace Prettier as the formatter for everything except Svelte files, but that's more involved as it requires keeping both formatters and configuring Prettier to run on certain files and not others.

PR #8438, opened 2025-05-09
Meta: Replace stylelint with biome

kovsu:
Close: #8367 

## Screenshot

CleanShot 2025-05-09 at 11 07 01

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,10 +6,10 @@
 		"build": "run-p build:* --continue-on-error",
 		"build:typescript": "tsc --noEmit",
 		"build:bundle": "rollup -c",
-		"fix": "run-p \"lint:css -- --fix\" \"lint:js -- --fix\" format \"vitest -- --update\" --continue-on-error",
+		"fix": "run-p \"lint:css --write\" \"lint:js -- --fix\" format \"vitest -- --update\" --continue-on-error",
 		"format": "prettier . --write",
 		"lint": "run-p lint:* --continue-on-error",
-		"lint:css": "stylelint \"source/**/*.css\"",
+		"lint:css": "biome lint ./source/**/*.css",
 		"lint:js": "eslint .",
 		"lint:prettier": "prettier . --check",
 		"new": "bash build/new-feature.sh",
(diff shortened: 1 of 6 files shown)


source/features/emphasize-draft-pr-label.css
```diff
@@ -1,6 +1,6 @@
 /* Explanation for the feature in https://github.com/refined-github/refined-github/pull/4728#issuecomment-910630727 */
 
-@media (resolution >= 2x) {
+@media (resolution: 2x) {
```
fregante, line 3: Not equivalent. Does Biome not support the new one? You might want to try the 2.0.0-beta.2 version

package.json
```diff
@@ -6,10 +6,10 @@
 		"build": "run-p build:* --continue-on-error",
 		"build:typescript": "tsc --noEmit",
 		"build:bundle": "rollup -c",
-		"fix": "run-p \"lint:css -- --fix\" \"lint:js -- --fix\" format \"vitest -- --update\" --continue-on-error",
+		"fix": "run-p \"lint:css --write\" \"lint:js -- --fix\" format \"vitest -- --update\" --continue-on-error",
 		"format": "prettier . --write",
 		"lint": "run-p lint:* --continue-on-error",
-		"lint:css": "stylelint \"source/**/*.css\"",
+		"lint:css": "biome lint ./source/**/*.css",
```
fregante, line 12: Nit: Since the config already includes this, you might be able to drop it

source/features/emphasize-draft-pr-label.css
kovsu: Maybe we should use [min-resolution](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/resolution).

kovsu:
🤬 We have configured '*.json' to be skipped in our settings, yet the package-lock.json file persists in being included.

fregante:
Known issue in v1 too :(

This is what I have in my other project:

```
{
  "$schema": "./node_modules/@biomejs/biome/configuration_schema.json",

  "vcs": {
    "enabled": true,
    "clientKind": "git",
    "useIgnoreFile": true
  },
  "files": {
    "ignore": [
      // npm already handles it
      "**/package.json",
    ]
  },
```

but biome v2 moved `ignore` to `files`


I think v2 still has issues with its scanner, taking more files than asked.

fregante:
- https://github.com/biomejs/biome/discussions/5633#discussioncomment-12811072

fregante:
Oh you mean this

https://github.com/refined-github/refined-github/blob/1d2383a290ee945cbd19aa327c8f83ebdc6d6092/.github/workflows/test.yml#L66-L69

kovsu:
> Oh you mean this

Yeah I mean it.

> I think v2 still has issues with its scanner, taking more files than asked.

I think the scanner is ok in my ide after I configure this.

```json
"files": {
	"includes": [
		"source/**/*.css"
	]
}
```

fregante:
Bug:

- https://github.com/codespell-project/actions-codespell/issues/88

source/features/emphasize-draft-pr-label.css
fregante: https://stylelint.io/user-guide/rules/media-feature-range-notation/ is a newer lint that biome doesn't have yet, but we still want to follow it and have no reason to change it here.

Related:
- https://github.com/biomejs/biome/issues/2511

fregante (approved):

kovsu:
I have one question.Given that we are exclusively linting CSS files here, should we also implement it for formatting?

fregante:
That was the second part of the issue. If you want, you can do it. Biome v2 should be slightly better at it now

kovsu:
Do it in the next step 🤣. The css code formatting by biome is not well.

fregante:
Yeah no rush. We can review when biome 2 is out of beta

PR #8454, opened 2025-05-19
GitHub bug fix: container's width is overing browser viewport

kovsu:
## Test URLs

PR description's input in [Folo](https://github.com/RSSNext/Folo)

## Screenshot

| Before | After |
| ----- | ----- |
| CleanShot 2025-05-19 at 15 13 07 | CleanShot 2025-05-19 at 15 27 46 |

### DIFF
--- a/source/features/github-bugs.css
+++ b/source/features/github-bugs.css
@@ -118,7 +118,9 @@ svg.octicon-package + a.Truncate[href*='/releases/download/'] > .Truncate-text {
 	}
 }
 
-/* Container's width is overing browser viewport #8454 */
+/* Issue picker does not clip unbreakable issue titles in PRs */
+/* Info: https://github.com/refined-github/refined-github/pull/8454 */
+/* Test: type "#113" in https://github.com/refined-github/sandbox/pull/4 */
 .suggester-container li {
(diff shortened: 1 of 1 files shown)


fregante:
I can't repro on https://github.com/refined-github/sandbox/issues/112

Screenshot

fregante:
So yeah this seems to be a bug with the picker in the old view, currently only on PRs

Screenshot 1

source/features/github-bugs.css
```diff
@@ -118,6 +118,14 @@ svg.octicon-package + a.Truncate[href*='/releases/download/'] > .Truncate-text {
 	}
 }
 
+/* Container's width is overing browser viewport #8454 */
```
fregante, line 121:
```suggestion
/* Issue picker does not clip unbreakable issue titles in PRs */
/* Info: https://github.com/refined-github/refined-github/pull/8454 */
/* Test "#113 in https://github.com/refined-github/sandbox/pull/4 */
```

fregante (approved):
I'm not a fan of adding features for views that are about to disappear, but this is small enough. If it causes issues it will be dropped rather than fixed though

Issue #8462, opened 2025-05-25
`sticky-sidebar` causes the README header and site footer to display over the Edit Repository Details overlay


Banaanae:
### Description

When the Edit Repository Details overlay is open, the readme header (the bar above readme used to swap between it and license) displays on top of the overlay

### How to replicate the issue + URL

1. Go to the main page of any repository you own (e.g. https://github.com/Banaanae/ComponentsEditor)
2. On the right side of the screen, click the settings icon next to the repository description
![Image](https://github.com/user-attachments/assets/a1b5ebeb-2c38-440d-8393-e2e9b3e128c1)
3. With this overlay open, scroll down until the readme header is visible, and observe
![Image](https://github.com/user-attachments/assets/1f260aee-50ad-4020-9283-032a297bada6)

### Extension version

25.5.18

### Browser(s) used

Firefox

### Token ⚠️

- [x] I set a token in the options page. I understand that some features require a token to work.

fregante:
Caused by https://github.com/refined-github/refined-github/pull/8447

I thought I checked this, I guess not on the repo home

Banaanae:
Was playing around with https://github.com/refined-github/refined-github/pull/8447/files#diff-9cf23e5c85f4864b29a8fddcf86ee951a259cc88afe241cda71d2314c634308dL41
and found that adding this line back fixes the issue, both keeping at 90 and setting to 2 work.
Didn't see the issue in #8445 with this line either, any reason to not just re-add this line?

fregante:
You can see the rest of the file and the list of bugs that it causes

PR #8477, opened 2025-06-13
`sticky-header` - Fix overlap with repository readme header

kovsu:
Close: #8462

## Test URLs

Any repo you own.

## Screenshot

https://github.com/user-attachments/assets/43c5ca32-9461-48dd-afd1-4a0d6af635fe

### DIFF
--- a/source/features/sticky-sidebar.css
+++ b/source/features/sticky-sidebar.css
@@ -46,6 +46,7 @@ Exclusively use simple sums and use `0px` instead of `0`
 	display: none;
 }
 
+/* Higher than sticky readme header https://github.com/refined-github/refined-github/issues/8462 */
 .rgh-sticky-sidebar:has(.details-reset[open]) {
 	z-index: 90;
 }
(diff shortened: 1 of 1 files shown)


fregante:
Have you verified that the other bugs that were mentioned in the code that I deleted? No point in merging fixes that just cause bugs elsewhere

kovsu:
@fregante Where is the list ?

fregante:
All the issue references in the deleted code in https://github.com/refined-github/refined-github/pull/8447/files

source/features/sticky-sidebar.css
```diff
@@ -35,6 +35,11 @@ Exclusively use simple sums and use `0px` instead of `0`
 .rgh-sticky-sidebar {
 	position: sticky;
 	top: var(--rgh-sticky-sidebar-offset, 0) !important;
+	/*
+		Make sure the container's z-index is higher than the readme header
+		to avoid the README header overlapping the sidebar.
+	 */
+	z-index: 9;
```
fregante: Rather than trigger another z-index fight, I'd either disable the feature entirely on the repo home or disable it when that modal is open. Hopefully this can be done without using `:has`. 

I'd be totally ok with calling `feature.unload` when the popup is opened and just be done with it

fregante:
Verified, thank you!

PR #8480, opened 2025-06-16
`global-conversation-list-filters` - Drop feature

fregante:
It's native now: https://github.com/issues

Screenshot 1


fregante:
Hopefully `more-conversation-filters` is next, they mentioned that they want to add it but haven't figured out how yet. The feature is already broken on repo issue lists

fregante:
Oof, didn't realize that the PR list still has this:

Screenshot 2

However nothing stops you from opening /issues and adding PR views in there, even if temporarily less convenient.

PR #8492, opened 2025-06-24
`list-prs-for-file` - Restore feature

moreal:
This pull request fixes a bug where `list-prs-for-file` does not add the dropdown button to view the PR list as expected. From my understanding, when it was last written, the "More file actions" dropdown button had an `aria-label` attribute, but it has now been changed to use `aria-labelledby`, which causes it not to function properly. So when I checked using the browser’s built-in Inspect tool, the `data-testid` attribute among the "More file actions" dropdown button’s properties appeared to be usable for identification purposes. I applied this to the code, and since it appears to work correctly in my case, I’m submitting this PR.

## Test URLs

- https://github.com/vuejs/vitepress/blob/main/src/client/theme-default/components/VPDocFooterLastUpdated.vue
- https://github.com/vuejs/vitepress/blob/main/src/node/plugins/localSearchPlugin.ts
- https://github.com/refined-github/sandbox/blob/6619/6619

## Screenshot

Screenshot for https://github.com/vuejs/vitepress/blob/main/src/client/theme-default/components/VPDocFooterLastUpdated.vue
image

Screenshot for https://github.com/vuejs/vitepress/blob/main/src/node/plugins/localSearchPlugin.ts
image

Screenshot for https://github.com/refined-github/sandbox/blob/6619/6619
image

Screenshot for https://github.com/refined-github/refined-github/blob/main/source/features/list-prs-for-file.tsx with narrow screen
image

### DIFF
--- a/source/features/list-prs-for-file.tsx
+++ b/source/features/list-prs-for-file.tsx
@@ -137,7 +137,7 @@ async function init(signal: AbortSignal): Promise<void> {
 	await expectToken();
 
 	observe([
-		'[aria-label="More file actions"]', // `isSingleFile`
+		'[data-testid="more-file-actions-button-nav-menu-wide"]', // `isSingleFile`
 		'[data-hotkey="Mod+s"]', // `isEditingFile`
 	], add, {signal});
 }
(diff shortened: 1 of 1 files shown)


moreal:
Hello @fregante, I pushed a new commit to make it show the PRs dropdown button even if the screen is narrow because the `data-testid` attribute is changed by the width of screen!

PR #8505, opened 2025-06-30
`reactions-avatars` - Restore feature for issue view

kovsu:
## Test URLs

https://github.com/refined-github/refined-github/issues/7856

## Screenshot

| Before | After |
| --- | ---- |
| CleanShot 2025-06-30 at 09 35 11 | CleanShot 2025-06-30 at 09 35 38 |

### DIFF
--- a/source/features/reactions-avatars.tsx
+++ b/source/features/reactions-avatars.tsx
@@ -26,11 +26,11 @@ function getParticipants(button: HTMLButtonElement): Participant[] {
 	let users;
 
 	if (button.getAttribute('role') === 'switch') { // [aria-label] alone is not a differentiator
-		users = button.getAttribute('aria-label')!
-			.replace(/.*including /, '')
-			.replace(/\)/, '')
+		users = button.nextElementSibling!
+			.textContent
 			.replace(/,? and /, ', ')
 			.replace(/, \d+ more/, '')
+			.replace(/\)/, '')
 			.split(', ');
 	} else if (button.nextElementSibling?.tagName === 'TOOL-TIP') {
 		// The list of people who commented is in an adjacent `<tool-tip>` element #5698


source/features/reactions-avatars.tsx
```diff
@@ -26,11 +26,11 @@ function getParticipants(button: HTMLButtonElement): Participant[] {
 	let users;
 
 	if (button.getAttribute('role') === 'switch') { // [aria-label] alone is not a differentiator
-		users = button.getAttribute('aria-label')!
-			.replace(/.*including /, '')
-			.replace(/\)/, '')
```
kovsu: Don't need to clean `including`

CleanShot 2025-06-30 at 09 37 16

kovsu:
It seems `locked` issue and  `locked` discussion losed their style.

source/features/reactions-avatars.tsx
fregante: "Including you" appears when you also reacted. Try on this PR

kovsu: Not like you said.

CleanShot 2025-06-30 at 11 00 44

fregante (approved):

fregante:
I'll release a new version tomorrow, so any PRs merged before then will make the cut.

source/features/reactions-avatars.tsx
fregante: Must change changed since. Would be good to extract this title cleaner and add a .test.ts file to document the supported formats.

We'd need to go back in time and see how we got to the current parser, and if those formats are still valid or not.

A good test string would be 
- `foo and 788more` -> `['foo', '788more']`
- `foo and 788 more` -> `['foo']`

PR #8527, opened 2025-07-03
`align-issue-labels` -  Support new issue lists

### LINKED ISSUE #8256: `align-issue-labels`: alignment doesn't work for "Issue"
### Description

In issue/PR lists, doesn't aligns the labels to the left, below each title.

Image

https://github.com/refined-github/refined-github/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen

### Extension version

25.1.24

### Browser(s) used

firefox, zen

### Token ⚠️

SunsetTechuila:
closes #8256

see the issue for more info

## Test URLs

https://github.com/nrwl/nx/issues

## Screenshot

image


source/features/align-issue-labels.tsx
```diff
@@ -1,8 +1,56 @@
...
+			continue;
+		}
+
+		issueTypeTokenContainer.classList.add('rgh-issue-type');
+		badgesContainer.prepend(issueTypeTokenContainer);
+	}
+}
+
+async function init(signal: AbortSignal): Promise<void> {
+	observe(issueRowSelector, batchedFunction(alignBadges, {delay: 100}), {signal});
```
fregante: batchedFunction is just for API calls. Here "observe" should call `alignBadge` directly

source/features/align-issue-labels.tsx
```diff
@@ -1,8 +1,56 @@
...
+
+const issueRowSelector = '[class^="IssueRow-module__row"]';
+const badgesContainerSelector = '[class*="trailingBadgesContainer"]';
+const mainContentInnerSelector = '[class^="MainContent-module__inner"]';
+const descriptionContainerSelector = '[class^="Description-module__container"]';
+const issueTypeContainerSelector = '[class*="issueTypeTokenContainer"]';
+
+function alignBadges(issueRows: HTMLElement[]): void {
+	for (const issueRow of issueRows) {
+		const badgesContainer = issueRow.querySelector(badgesContainerSelector);
```
fregante: Consider shortening variable names to just "badges", "description", "types"

source/features/align-issue-labels.tsx
```diff
@@ -1,8 +1,56 @@
...
+		if (!badgesContainer) {
+			continue;
+		}
+
+		const mainContentInner = issueRow.querySelector(mainContentInnerSelector);
```
fregante: All querySelector usage should be replaced with select-dom. Import the "strict" version and then use `$()`

`$optional()` can also be used if it *already* an element doesn't exist. If some must exist now but doesn't in the future, `$()` throws an error so we know when/where the feature failed.

source/features/align-issue-labels.css
```diff
@@ -1,36 +1,22 @@
-html:not([rgh-OFF-align-issue-labels]) .js-issue-row {
```
fregante, line 1: The old view is still used on PRs so most likely we need to preserve the old CSS too. 

You can keep the `addCssFeature` call in JS alongside the regular `add`

SunsetTechuila: check failed

SunsetTechuila: could you please show where the old css still works?

fregante: https://github.com/refined-github/refined-github/pulls

fregante: You can delete these lines, they're no longer applicable:

https://github.com/refined-github/refined-github/blob/037fa9a507b83d1cf94a71c318d777551c9ca9df/build/features.test.ts#L165-L168

source/features/align-issue-labels.css
```diff
@@ -34,3 +34,28 @@ html:not([rgh-OFF-align-issue-labels]) .js-issue-row {
 		font-size: 11px !important;
 	}
 }
+
+/* New view */
```
fregante:
```suggestion
/* React issue lists */
/* TODO: Remove the old styles in 2026 */
```

source/features/align-issue-labels.tsx
```diff
@@ -1,9 +1,54 @@
 import './align-issue-labels.css';
 
+import * as pageDetect from 'github-url-detection';
+import {$, $optional} from 'select-dom/strict.js';
+
 import features from '../feature-manager.js';
+import observe from '../helpers/selector-observer.js';
 
 void features.addCssFeature(import.meta.url);
```
fregante, line 10:
```suggestion
// TODO: Drop in 2026
void features.addCssFeature(import.meta.url);
```

source/features/align-issue-labels.tsx
```diff
@@ -1,9 +1,54 @@
...
+	mainContent.classList.add('rgh-issue-main-content-inner');
+	mainContent.append(badges);
+
+	if (issueType) {
+		issueType.classList.add('rgh-issue-type');
+		badges.prepend(issueType);
+	}
+
+	const description = $(descriptionSelector, issue);
+	description.classList.add('rgh-issue-description');
```
fregante: If the selectors are used only once, it's preferable to have them inline

```suggestion
	$('[class^="Description-module__container"]', issue).classList.add('rgh-issue-description');
```

source/features/align-issue-labels.tsx
```diff
@@ -1,9 +1,54 @@
...
 
+/* New view */
+
+const issueSelector = '[class^="IssueRow-module__row"]';
+const badgesSelector = '[class*="trailingBadgesContainer"]';
+const mainContentSelector = '[class^="MainContent-module__inner"]';
+const descriptionSelector = '[class^="Description-module__container"]';
+const issueTypeSelector = '[class*="issueTypeTokenContainer"]';
+
+function alignBadges(issue: HTMLElement): void {
```
fregante:
```suggestion
function alignBadge(issue: HTMLElement): void {
```

source/features/align-issue-labels.tsx
```diff
@@ -12,5 +57,6 @@ https://github.com/pulls
 https://github.com/bmish/eslint-doc-generator/pulls
 https://github.com/bmish/eslint-doc-generator/issues
 https://github.com/bmish/eslint-doc-generator/milestone/1
```
fregante: I'm just replacing the URL.

```suggestion
https://github.com/bmish/eslint-doc-generator/milestone/1?closed=1
```

I assume the feature is broken there, I wouldn't worry about it in this PR.

source/features/align-issue-labels.tsx
```diff
@@ -1,9 +1,54 @@
...
+const mainContentSelector = '[class^="MainContent-module__inner"]';
+const descriptionSelector = '[class^="Description-module__container"]';
+const issueTypeSelector = '[class*="issueTypeTokenContainer"]';
+
+function alignBadges(issue: HTMLElement): void {
+	const badges = $(badgesSelector, issue);
+	const mainContent = $(mainContentSelector, issue);
+	const issueType = $optional(issueTypeSelector, issue);
+
+	if (badges.children.length === 0 && !issueType) {
```
fregante: If there are no labels, we shouldn't move anything at all, right?


```suggestion
	if (badges.children.length === 0) {
```

fregante (commented):
A few nits 🙏

source/features/align-issue-labels.tsx
SunsetTechuila: I did that so the layout would be consistent

| current | suggested |
|--------|--------|
| image | image |

fregante: I like consistency, but I prefer the "suggested" version for now since it's more compact and avoids DOM changes

source/features/align-issue-labels.tsx
SunsetTechuila: it does work 

image

source/features/align-issue-labels.tsx
SunsetTechuila: but there could be multiple badges and the function aligns them all 🤔

fregante: Oh yeah 😅

source/features/align-issue-labels.css
```diff
@@ -34,3 +34,29 @@ html:not([rgh-OFF-align-issue-labels]) .js-issue-row {
...
+	margin-top: var(--base-size-4);
+}
+
+.rgh-issue-main-content-inner {
+	gap: var(--base-size-4);
+}
+
+.rgh-issue-description {
+	gap: unset;
+}
```
fregante: Instead of creating custom classes, I think these already exist as utility classes. Maybe they're `mt-1`, `gap-1` and `gap-0`. You'd have to find them in the developer tools, there are CSS files loaded by GitHub named "primer", etc

source/features/align-issue-labels.css
```diff
@@ -34,3 +34,29 @@ html:not([rgh-OFF-align-issue-labels]) .js-issue-row {
...
+
+.rgh-issue-type {
+	display: inline-block;
+	margin-right: var(--base-size-4);
+
+	.rgh-issue-badges:has(:nth-child(2)) &:after {
+		content: '•';
```
fregante: Not necessary, but if you can copy the line as it's in the app. Here's the light version so you can find the color that works for both light and night mode

Image

fregante (approved):
🎈

source/features/align-issue-labels.css
SunsetTechuila: which is better?

| 1 | 2 |
|--------|--------|
| image | image |

source/features/align-issue-labels.css
SunsetTechuila: there is no utility classes for setting `gap` and we still have to create custom classes for badges and issue type to add a separator

source/features/align-issue-labels.css
fregante: 1

source/features/align-issue-labels.tsx
```diff
@@ -24,15 +24,15 @@ function alignBadges(issue: HTMLElement): void {
 		return;
 	}
 
-	badges.classList.add('rgh-issue-badges');
+	badges.classList.add('rgh-issue-badges', 'mt-1');
```
fregante: rgh-issue-badges and rgh-issue-type don't exist anymore, they can be dropped

SunsetTechuila: they do: https://github.com/refined-github/refined-github/commit/a172704293489de5afc48bb87e6a831edc25e4ae#diff-57cfe7f2460a24b5830d888bebaced23a00239bacd74fa83ff654e0e4b27b365R49

source/features/align-issue-labels.css
SunsetTechuila: I will use the color from the second screenshot because the color in the first is barely visible on the dark theme

fregante:
Thank you it's great! I'll test it out later today and merge it tomorrow.

fregante:
It seems that some aren't covered here:

https://github.com/issues?q=wonderful%20facebook%20space%20bug%20tweet%20downloads%20the%20github%20optional

Screenshot 2

This seems to be ok:

https://github.com/pulls?q=%22refined+github%22+is%3Apr+-repo%3Arefined-github+sweet+bug

SunsetTechuila:
fixed

image

source/features/align-issue-labels.css
SunsetTechuila: > gap-1

found in `global.css` with this exact name

source/features/align-issue-labels.css
```diff
@@ -1,3 +1,11 @@
+.rgh-issue-badges:has(> :nth-child(2)) .rgh-issue-type:after {
```
fregante: We should move this condition to the JS since we have JS. Our usage of :has() has to be limited due to perf concerns.

SunsetTechuila: The problem is that labels could be updated in the background

SunsetTechuila: Cool, right?

fregante: Yes but the rest of the logic still lives in the JS anyway, so what happens when they remove the issue type? It's ok if it's slightly inconsistent due to this. I prefer not having "RGH is slow" issues open.

SunsetTechuila: The current implementation handles label changes in the background fine; the suggested one would introduce:

- If all labels were removed in the background:  
image
- If labels were added in the background:  
Screenshot 2025-07-09 193233

> so what happens when they remove the issue type?

- If you change the issue type, the type badge updates without moving
- If you remove the issue type, nothing happens
- If you remove the issue type and then add a type, the moved type badge stays in place, but a new one is added. So you can end up with something like this:  
image  
or even this:  
image

Maybe it would be better not to touch the type badges at all

SunsetTechuila: > Maybe it would be better not to touch the type badges at all

Or we could move the whole issue type badge container instead (`[class^="IssueTypeIndicator-module__container"]`)

SunsetTechuila: > Or we could move the whole issue type badge container instead ([class^="IssueTypeIndicator-module__container"])

Epic fail, this is what happened after removing the issue type in the background:

image

SunsetTechuila: So what should we do now?

BoomchainLabs (approved):

BoomchainLabs (approved):

source/features/align-issue-labels.css
fregante: Does that error happen consistently? Like if you interact with the list (change page, sorting, add labels)

fregante: Does that error happen with the current version or only if you move the logic to the JS?

SunsetTechuila: Labels are fine; the issue type badges are problematic, and this error only happens when the issue type is removed

> Does that error happen with the current version

No, this error is the result of moving the issue type badge container

fregante: > with the current version

I mean the current version of this PR, so the answer appears to be "yes"

SunsetTechuila: No, in the current version of this PR a different element is moved: https://github.com/refined-github/refined-github/pull/8527/files#diff-90ef46e4a41ff744d8fa919d247b49eae4f2a407fadedb8f2cc3addecdb54cc0R25

fregante: Let me rephrase that: is this PR currently causing that error in any situation?

SunsetTechuila: No, this is what it causes:

> * If you remove the issue type, nothing happens
> * If you remove the issue type and then add a type, the moved type badge stays in place, but a new one is added. So you can end up with something like this:
>   image
>   or even this:
>   image

source/features/align-issue-labels.tsx
```diff
@@ -1,16 +1,56 @@
...
+	const mainContent = $('[class^="MainContent-module__inner"]', row);
+	mainContent.classList.add('gap-1');
+	mainContent.append(badges);
+
+	$('[class^="Description-module__container"]', row).style.gap = 'unset';
+
+	const issueType = $optional('[class*="issueTypeTokenContainer"]', row);
+	if (issueType) {
+		issueType.classList.add('rgh-issue-type', 'd-inline-block');
+		badges.prepend(issueType);
```
fregante: So we have to chose between these two:

Screenshot 11

1. either move the type to the last line — and risk an out of date / wrong view when updating the type
2. or leave the type on the second line — and making it look a bit awkward at all times 

I'm leaning towards the ugly bug-free version (2) over the pretty, buggy one (1)

I'm still hoping that GitHub realizes that their mobile app team is just better at design and they follow it so we can drop this feature.

SunsetTechuila: > GitHub realizes that their mobile app team is just better at design

you can't be sure that they moved the badges to the bottom because it makes them easier to read, not because phones don't have much horizontal space)

fregante:
Can you update the readme screenshot with this URL? https://github.com/user-attachments/assets/dca5dc12-7283-4704-a93f-5bfe5f2b1938

fregante (approved):

fregante:
Great! Looking forward to seeing this style again. It was one of the earliest Refined GitHub features

kovsu:
Can we just keep one `bug` label here? It makes me a little bit annoyed.

CleanShot 2025-08-06 at 10 15 20

fregante:
I don’t think we can do that, if people want to use both, it's their fault. 

Maybe open a new feature request to discuss it

Issue #8526, opened 2025-07-03
`center-reactions-popup` broken


fregante:
### Description

<img width="382" alt="Image" src="https://github.com/user-attachments/assets/82ab1a00-7a4f-4f19-bcb5-a103b9a300db" />

See desired alignment in https://github.com/refined-github/refined-github/pull/6329, it's should be outside the box (bottom, left), more centered to the reaction button itself 

### How to replicate the issue + URL

Here

### Extension version

25.5.18

### Browser(s) used

Safari 

### Token ⚠️

- [x] I set a token in the options page. I understand that some features require a token to work.

fregante:
Test comment

fregante:
多谢啦~

PR #8528, opened 2025-07-04
`center-reactions-popup` - Restore feature

kovsu:
Close: #8526 

## Test URLs

Any issue

## Screenshot

CleanShot 2025-07-04 at 09 08 48

### DIFF
--- a/source/features/center-reactions-popup.css
+++ b/source/features/center-reactions-popup.css
@@ -1,5 +1,5 @@
 /* Restore two-line layout */
-:root .dropdown-menu-reactions {
+:root ul[role="menu"]:has([class*="ReactionsMenuItem-module"]) {
 	display: grid !important;
 	grid-template-rows: 1fr 1fr; /* Keep two lines whether there are 6 or 8 reactions */
 	grid-auto-flow: column;


fregante:
It's not centered around the mouse. After you click it, the mouse should be already on top of it.

source/features/center-reactions-popup.css
```diff
@@ -12,12 +12,28 @@
...
 		translate: -27px 61px;
 	}
 }
 
+/* React Issue View */
+:root div[data-visibility-visible]:has([class*='ReactionsMenuItem-module']) {
+	& > ul {
+		display: grid !important;
+		grid-template-rows: 1fr 1fr; /* Keep two lines whether there are 6 or 8 reactions */
+		grid-auto-flow: column;
```
fregante, line 30: We should probably also change the width, it seems too wide.

fregante:
Are other places affected too? Releases, Discussions, etc. 

No need to fix them right here, I'm just wondering.

kovsu:
I have checked them and it works fine here.

kovsu:
> We should probably also change the width, it seems too wide.

Updated.

source/features/center-reactions-popup.css
```diff
@@ -12,12 +12,31 @@
...
+	}
+
+	/* Reset the width */
+	min-width: 120px !important;
```
fregante:
```suggestion
	min-width: 0px !important;
```

fregante (approved):

fregante:
I added some gifs to the PR for future reference

PR #8544, opened 2025-07-07
`sticky-comment-header` - New feature

### LINKED ISSUE #8463: Sticky header for issue / PR / discussion comments
Hi @fregante, hope everything is good with you! 👋

### Description

For long comments on issues, PRs and discussions, I was wondering if there would be any appetite for Refined GitHub to have a sticky header for quick access to common features without scrolling up to the top of the comment (editing the comment, linking to the comment, navigating to the commenter's GitHub profile, anything else in the three dots menu).

It would mean that there would be 2 sticky headers, although the comment header is not very tall.

I would find this personally useful when making many updates to long comments, where I want to visually see the bottom of the comment, but quickly be able to edit the comment.

### Example URLs

- https://github.com/reactjs/react.dev/pull/5487#issuecomment-1409720741
- https://github.com/nodejs/typescript/issues/14

### Mockup

https://github.com/user-attachments/assets/fdb887c4-9328-4fc9-929c-2a3416936b06

SunsetTechuila:
resolves #8463

## Test URLs

https://github.com/refined-github/sandbox/issues/117

https://github.com/refined-github/sandbox/pull/118

### DIFF
--- a/source/features/sticky-comment-header.css
+++ b/source/features/sticky-comment-header.css
@@ -16,9 +16,7 @@ html:not([rgh-OFF-sticky-comment-header]) {
 		z-index: 2;
 	}
 
-	[data-testid='issue-body']
-		[class^='Box']:has(> #issue-body-viewer)
-		> [class^='Box'],
+	[class^="IssueBodyHeader-module__IssueBodyHeaderContainer"],
 	[data-testid='comment-header'] {
 		border-radius: 0;
 	}
--- a/source/features/sticky-comment-header.css
+++ b/source/features/sticky-comment-header.css
@@ -17,7 +17,7 @@ html:not([rgh-OFF-sticky-comment-header]) {
 	[data-testid='issue-body'] [class^='Box']:has(> #issue-body-viewer),
 	[class^='IssueCommentViewer-module__IssueCommentContent'],
 	/* PR description/comment container */
-	.timeline-comment {
+		.timeline-comment {
 		border-radius: var(--borderRadius-medium);
 		overflow: clip;
 	}
--- a/source/features/sticky-comment-header.css
+++ b/source/features/sticky-comment-header.css
@@ -11,7 +11,7 @@ html:not([rgh-OFF-sticky-comment-header]) {
 	}
 
 	/* Added by `show-names` */
-	.rgh-user-own-comment,
+	.rgh-viewer-did-author,
 	.timeline-comment.current-user .timeline-comment-header,
 	[class*="IssueBodyHeader-module__viewerDidAuthor"] {
 		--background: var(--bgColor-accent-muted, fuchsia);
--- a/source/features/sticky-comment-header.css
+++ b/source/features/sticky-comment-header.css
@@ -29,15 +29,20 @@ html:not([rgh-OFF-sticky-comment-header]) {
 	}
 
 	@supports (anchor-name: --test) {
-		[class*='ActivityHeader-module__activityHeader']
-			button[aria-haspopup='true'][aria-expanded='true'] {
-			anchor-name: --header;
-		}
+		[app-name='issues-react'] {
+			[class*='ActivityHeader-module__activityHeader']
+				button[aria-haspopup='true'][aria-expanded='true'] {
+				anchor-name: --header;
+			}
 
-		[class^='prc-Overlay-Overlay'][data-variant='anchored'] {
-			position: fixed !important;
-			position-anchor: --header;
-			top: anchor(bottom) !important;
+			[class^='prc-Overlay-Overlay'][data-variant='anchored']:has(
+					.octicon-quote,
+					[class*='scrollableEditHistoryList']
+				) {
+				position: fixed !important;
+				position-anchor: --header;
+				top: anchor(bottom) !important;
+			}
 		}
 	}
 }
--- a/source/features/sticky-comment-header.css
+++ b/source/features/sticky-comment-header.css
@@ -1,8 +1,6 @@
 html:not([rgh-OFF-sticky-comment-header]) {
 	/* Issue description header */
-	[data-testid='issue-body']
-		[class^='Box']:has(> #issue-body-viewer)
-		> [class^='Box'],
+	[class^="IssueBodyHeader-module__IssueBodyHeaderContainer"],
 	/* Issue comment header */
 		[data-testid='comment-header'],
 	/* PR description/comment header */
--- a/source/features/sticky-comment-header.css
+++ b/source/features/sticky-comment-header.css
@@ -15,7 +15,7 @@ html:not([rgh-OFF-sticky-comment-header]) {
 	.timeline-comment.current-user .timeline-comment-header,
 	[class*="IssueBodyHeader-module__viewerDidAuthor"] {
 		/* Color from GitHub"s dark mode styles */
-		--background: var(--bgColor-accent-muted, var(--color-accent-subtle, rgba(56, 139, 253, 0.1)));
+		--background: var(--bgColor-accent-muted, fuchsia);
 		background: linear-gradient(var(--background)), var(--rgh-background) !important;
 	}
 
--- a/source/features/sticky-comment-header.css
+++ b/source/features/sticky-comment-header.css
@@ -1,9 +1,9 @@
 html:not([rgh-OFF-sticky-comment-header]) {
-	/* Issue description header */
+	/* Issue body header */
 	[class^="IssueBodyHeader-module__IssueBodyHeaderContainer"],
-	/* Issue comment header */
+	/* Issue body header */
 		[data-testid='comment-header'],
-	/* PR description/comment header */
+	/* PR body/comment header */
 		.timeline-comment-header {
 		position: sticky;
 		top: calc(var(--base-sticky-header-height, 0px) + 56px);
@@ -27,7 +27,7 @@ html:not([rgh-OFF-sticky-comment-header]) {
 		border-radius: 0;
 	}
 
-	/* Issue description container */
+	/* Issue body container */
 	[data-testid="issue-body"] [class^="Box"]:has(> #issue-body-viewer),
 	/* Issue comment container */
 		[class^="IssueCommentViewer-module__IssueCommentContent"] {
--- a/source/features/sticky-comment-header.css
+++ b/source/features/sticky-comment-header.css
@@ -6,20 +6,24 @@ html:not([rgh-OFF-sticky-comment-header]) {
 	/* Issue comment header */
 	[data-testid='comment-header'],
 	/* PR description/comment header */
-		.timeline-comment-header {
+	.timeline-comment-header {
 		position: sticky;
 		top: calc(var(--base-sticky-header-height, 0px) + 56px);
 		z-index: 1;
 		backdrop-filter: blur(20px);
+	}
+
+	[data-testid='issue-body']
+		[class^='Box']:has(> #issue-body-viewer)
+		> [class^='Box'],
+	[data-testid='comment-header'] {
 		border-radius: 0;
 	}
 
 	/* Issue description container */
 	[data-testid='issue-body'] [class^='Box']:has(> #issue-body-viewer),
 	/* Issue comment container */
-	[class^='IssueCommentViewer-module__IssueCommentContent'],
-	/* PR description/comment container */
-		.timeline-comment {
+	[class^='IssueCommentViewer-module__IssueCommentContent'] {
 		border-radius: var(--borderRadius-medium);
 		overflow: clip;
 	}
--- a/source/features/sticky-comment-header.css
+++ b/source/features/sticky-comment-header.css
@@ -14,7 +14,6 @@ html:not([rgh-OFF-sticky-comment-header]) {
 	.rgh-user-own-comment,
 	.timeline-comment.current-user .timeline-comment-header,
 	[class*="IssueBodyHeader-module__viewerDidAuthor"] {
-		/* Color from GitHub"s dark mode styles */
 		--background: var(--bgColor-accent-muted, fuchsia);
 		background: linear-gradient(var(--background)), var(--rgh-background) !important;
 	}
--- a/source/features/sticky-comment-header.css
+++ b/source/features/sticky-comment-header.css
@@ -1,0 +1,24 @@
+html:not([rgh-OFF-sticky-comment-header]) {
+	/* Issue header */
+	[data-testid='issue-body']
+		[class^='Box']:has(> #issue-body-viewer)
+		> [class^='Box'],
+	[data-testid='comment-header'],
+	/* PR header */
+		.timeline-comment-header {
+		position: sticky;
+		top: calc(var(--base-sticky-header-height, 0px) + 56px);
+		z-index: 10;
+		backdrop-filter: blur(20px);
+
+		/* &::before {
+			content: '';
+			position: absolute;
+			height: 38px;
+			left: 0;
+			right: 0;
+			background-color: var(--rgh-background);
+			z-index: -1;
+		} */
+	}
+}
--- a/source/features/sticky-comment-header.css
+++ b/source/features/sticky-comment-header.css
@@ -8,37 +8,43 @@ html:not([rgh-OFF-sticky-comment-header]) {
 		position: sticky;
 		top: calc(var(--base-sticky-header-height, 0px) + 56px);
 		z-index: 1;
-		backdrop-filter: blur(20px);
 	}
 
-	/* show menus on top of headers */
+	.timeline-comment.current-user .timeline-comment-header,
+	[class*="IssueBodyHeader-module__viewerDidAuthor"] {
+		/* Color from GitHub"s dark mode styles */
+		--background: var(--bgColor-accent-muted, var(--color-accent-subtle, rgba(56, 139, 253, 0.1)));
+		background: linear-gradient(var(--background)), var(--rgh-background) !important;
+	}
+
+	/* Show menus on top of headers */
 	.js-timeline-item:has(details:open) .timeline-comment-header {
 		z-index: 2;
 	}
 
 	[class^="IssueBodyHeader-module__IssueBodyHeaderContainer"],
-	[data-testid='comment-header'] {
+	[data-testid="comment-header"] {
 		border-radius: 0;
 	}
 
 	/* Issue description container */
-	[data-testid='issue-body'] [class^='Box']:has(> #issue-body-viewer),
+	[data-testid="issue-body"] [class^="Box"]:has(> #issue-body-viewer),
 	/* Issue comment container */
-		[class^='IssueCommentViewer-module__IssueCommentContent'] {
+		[class^="IssueCommentViewer-module__IssueCommentContent"] {
 		border-radius: var(--borderRadius-medium);
 		overflow: clip;
 	}
 
 	@supports (anchor-name: --test) {
-		[app-name='issues-react'] {
-			[class*='ActivityHeader-module__activityHeader']
-				button[aria-haspopup='true'][aria-expanded='true'] {
+		[app-name="issues-react"] {
+			[class*="ActivityHeader-module__activityHeader"]
+				button[aria-haspopup="true"][aria-expanded="true"] {
 				anchor-name: --header;
 			}
 
-			[class^='prc-Overlay-Overlay'][data-variant='anchored']:has(
+			[class^="prc-Overlay-Overlay"][data-variant="anchored"]:has(
 					.octicon-quote,
-					[class*='scrollableEditHistoryList']
+					[class*="scrollableEditHistoryList"]
 				) {
 				position: fixed !important;
 				position-anchor: --header;
--- a/source/features/sticky-comment-header.css
+++ b/source/features/sticky-comment-header.css
@@ -14,8 +14,7 @@ html:not([rgh-OFF-sticky-comment-header]) {
 	.rgh-viewer-did-author,
 	.timeline-comment.current-user .timeline-comment-header,
 	[class*="IssueBodyHeader-module__viewerDidAuthor"] {
-		--background: var(--bgColor-accent-muted, fuchsia);
-		background: linear-gradient(var(--background)), var(--rgh-background) !important;
+		background: linear-gradient(var(--bgColor-accent-muted, fuchsia)), var(--rgh-background) !important;
 	}
 
 	/* Show menus on top of headers */
--- a/source/features/sticky-comment-header.css
+++ b/source/features/sticky-comment-header.css
@@ -1,7 +1,7 @@
 html:not([rgh-OFF-sticky-comment-header]) {
 	/* Issue body header */
 	[class^="IssueBodyHeader-module__IssueBodyHeaderContainer"],
-	/* Issue body header */
+	/* Issue comment header */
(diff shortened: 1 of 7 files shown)


SunsetTechuila:
Please add a screenshot yourself 

Issues:

- ~Menus are not sticky.~ (Fixed on Chromium)
- ~In the dark mode, the headers of your own comments have a semi-transparent background color. Since I couldn’t add a backdrop, I used a blur effect instead. However, this makes the header text difficult to read on brighter backgrounds.~
- Issue descriptions headers start to move a little bit earlier than they should:
  

From https://github.com/refined-github/refined-github/issues/8463#issuecomment-2975150677:

- [x] that bar
- [x] mobile view
- [x] notification bar
- [x] automatically adjust when the notification bar is closed
- [x] should still only be CSS only and avoid broad `:has()` usage (e.g. `body:has()`)

source/features/sticky-comment-header.css
```diff
@@ -17,7 +17,7 @@ html:not([rgh-OFF-sticky-comment-header]) {
 	[data-testid='issue-body'] [class^='Box']:has(> #issue-body-viewer),
 	[class^='IssueCommentViewer-module__IssueCommentContent'],
 	/* PR description/comment container */
-	.timeline-comment {
+		.timeline-comment {
```
SunsetTechuila: what is this formatting

fregante: Biome is a bit weird with CSS comments

source/features/sticky-comment-header.css
```diff
@@ -0,0 +1,24 @@
...
+	[data-testid='issue-body']
+		[class^='Box']:has(> #issue-body-viewer)
+		> [class^='Box'],
+	[data-testid='comment-header'],
+	/* PR description/comment header */
+		.timeline-comment-header {
+		position: sticky;
+		top: calc(var(--base-sticky-header-height, 0px) + 56px);
+		z-index: 1;
+		backdrop-filter: blur(20px);
```
fregante: The thing is we already have performance issues with selectors, we can't add *rendering* slowdowns on top of that. 

Since it's translucent, you might be able to easily add a color below it that matches that comment body. I say "below it" because the comment header has at least 3 colors (normal, your comments, :target comments) so you'd have to add 3 selectors to override them all. 

If there's no additional wrapper, you could add a color via `box-shadow` (without blur), I don't know if it paints under the element though.

source/features/sticky-comment-header.css
```diff
@@ -0,0 +1,24 @@
+html:not([rgh-OFF-sticky-comment-header]) {
+	/* Issue description header */
+	[data-testid='issue-body']
+		[class^='Box']:has(> #issue-body-viewer)
+		> [class^='Box'],
+	[data-testid='comment-header'],
```
fregante: What is this second selector?

source/features/sticky-comment-header.tsx
```diff
@@ -0,0 +1,11 @@
+import './sticky-comment-header.css';
+
+import features from '../feature-manager.js';
+
+void features.addCssFeature(import.meta.url);
+
+/*
+
+# Test URLs
```
fregante: This needs both issues and PR URLs specified. You can find some in the sandbox repo

fregante:
> Issue descriptions headers start to move a little bit earlier that they should:

It's unfortunately one of those things that will cause bug reports and I don't really want any more of those. 

The solution is likely to use :has() but that's why one of the restrictions was specifically "no broad has" 🥲 

> Menus are not sticky.

They open in the right place at least, right?

source/features/sticky-comment-header.css
SunsetTechuila: issue comment header

SunsetTechuila:
> > Menus are not sticky.
> 
> They open in the right place at least, right?

Yes

source/features/sticky-comment-header.css
fregante: It needs a comment so we know, even if (at the moment) the selector itself already suggests what it is.

source/features/sticky-comment-header.css
SunsetTechuila: > the comment header has at least 3 colors

Yes, and the class that define the color of your own comment headers is auto-generated:

image

> I don't know if it paints under the element though

It doesn't

fregante: Worth trying with :before {background: theColorWeWant}

SunsetTechuila: Tried: https://github.com/refined-github/refined-github/pull/8544/commits/52b8f650206522d5dfa56be02468f49533ee593f#diff-78e683a73e20858e09eb458d59f87f8cf6d4f4c172065e375dbd7beef53815e7R14-R22

SunsetTechuila:
> Menus are not sticky.

More or less fixed

Vangelis66:
(Reposted from https://github.com/refined-github/refined-github/issues/8463#issuecomment-3065674630 )

**Browser**: Firefox Nightly v**142.0a1** 32-bit

I've tested the artifact from: 

https://github.com/refined-github/refined-github/actions/runs/16232281217?pr=8544

on below **GH issue**: 

https://github.com/e3kskoy7wqk/Firefox-for-windows-7/issues/69

and there's a bug when I scroll down to **my comment**; e.g. 

https://github.com/e3kskoy7wqk/Firefox-for-windows-7/issues/69#issue-3203311286

is fine: 

Image

... however 

https://github.com/e3kskoy7wqk/Firefox-for-windows-7/issues/69#issuecomment-3040324555

appears bugged (the second fixed header has a **transparent** background) : 

Image

Hopefully an easy fix 😄 ...

SunsetTechuila:
> appears bugged (the second fixed header has a transparent background) :

do you have hw acceleration enabled? `backdrop-filter: blur` doesn't work without it

fregante:
As mentioned, this PR cannot be merged with `filter` so if there's no alternative solution it can be closed.

SunsetTechuila:
I don't see any css-only alternative

SunsetTechuila:
if only there were a way to extract the alpha value from the color, then `color-mix` can be used

Vangelis66:
(Apologies for the tardy response, I had been away from my laptop for several hours 😉 )

> do you have **hw acceleration enabled**?

Actually no 😞 ; this is a quite **old** laptop with a _blacklisted_, vendor customised, gfx driver that can't be further updated; so Firefox's **WebRender** is running in "**Software**" mode; but the overall browsing experience here is quite satisfactory for 99% of my needs (they exclude watching online 1080p video 😉 ) ...

> `backdrop-filter: blur` doesn't work **without it**

... Sad to hear; it's unfortunate when website code **discriminates against older H/W**; and I'm not talking only about _CSS_ here; websites have appeared (in **WASM**) which only load with a SSE4.2+ CPU 😠 ...

FWIW, the bug I reported seems to be more acute (on my system) **when inside an issue tracker**; things look much better when inside a **PR**'s `Conversation` tab: 

https://github.com/refined-github/refined-github/pull/8544#issuecomment-3065792439

sh3

I still think there's merit 👍 in this actual PR, thus sad it can't be implemented in a **CSS-only manner**; is everybody sure no compromise can be reached without `filter` ?

karlhorky:
I see two alternatives to blurring:

1. What I did in my implementation for the mockup in #8463 was to hardcode a non-transparent background color, sampling the resulting dark blue from applying the transparent color on top of the background color (alternatively, using a different existing  non-transparent color from GitHub's CSS variables). If hardcoding values is unwanted, this could also possibly be calculated dynamically, using alpha compositing / blending.
2. Or, if calculating a non-transparent color value seems too brittle, then adding a new element behind the header (exactly sized to the header) with the GitHub background color would be another option

SunsetTechuila:
The problem is how to do it, not what to do: https://github.com/refined-github/refined-github/pull/8544#discussion_r2192402641

karlhorky:
@SunsetTechuila I don't understand what you mean, your CSS selector in the current PR seems to target the element?

```css
[data-testid='issue-body'] [class^='Box']:has(> #issue-body-viewer) > [class^='Box'],
[data-testid='comment-header'],
.timeline-comment-header {}
```

Or do you mean that using `background: <color>` won't work because of specificity or something?

source/features/sticky-comment-header.css
karlhorky: `::before` pseudoelements do look like they work, but needs a `position: relative` on the parent:

```css
[data-testid='issue-body'] [class^='Box']:has(+ #issue-body-viewer) {
  position: relative;
}

[data-testid='issue-body'] [class^='Box']:has(+ #issue-body-viewer)::before {
  content: '';
  position: absolute;
  left: 0;
  right: 0;
  top: 0;
  background: #a1ff04;
  bottom: 0;
  z-index: -1;
}
```

Screenshot 2025-07-13 at 18 08 05

source/features/sticky-comment-header.css
```diff
@@ -0,0 +1,53 @@
+html:not([rgh-OFF-sticky-comment-header]) {
+	/* Issue description header */
+	[data-testid='issue-body']
+		[class^='Box']:has(> #issue-body-viewer)
+		> [class^='Box'],
```
karlhorky: not sure if sibling selectors inside `:has()` are more performant, but this would be an option:

```css
[data-testid='issue-body'] [class^='Box']:has(+ #issue-body-viewer) {}
```

or, more brittle, but no `:has()`:

```css
[data-testid='issue-body'] > [class^='Box'] > [class^='Box'] > [class^='Box'] {}
```

karlhorky: would also be an option to combine the two to fail more gracefully if the triple child combinator fails, not sure if that's better though:

```css
[data-testid='issue-body'] > [class^='Box'] > [class^='Box'] > [class^='Box']:has(+ #issue-body-viewer) {}
```

SunsetTechuila:
> I don't understand what you mean

What I mean is, I don't see any way to draw a solid background under the header, and it's not possible to target the headers of your own comments, which have a semi-transparent background, on the issue page

source/features/sticky-comment-header.css
SunsetTechuila: This replaces background of the header. And we have no way of knowing what the original background was: https://github.com/refined-github/refined-github/pull/8544#discussion_r2192402641

karlhorky:
Ok, I'm understanding the challenges a bit better now, thanks!

> it's also not possible to target your own comments on the issue page

Would it make sense for Refined GitHub to add selectors on comments which are your own? Or is this JS too expensive?

SunsetTechuila:
> Would it make sense for Refined GitHub to add selectors on comments which are your own? Or is this JS too expensive?

That's something @fregante should answer. Initial requirement was to make this feature css-only

karlhorky:
I tried a few things, and the following didn't work:

1. `::before` on the header - background over top of header
2. `box-shadow` - doesn't show up

One other idea I had was to add another sticky sibling element behind it (`::before` on the parent) - which appears to work - but I'm guessing these hardcoded `38px` values unacceptable.

Maybe we can calculate the height somehow...? Maybe just adding a space (content) and then adding the same top + bottom padding...?

```css
[data-testid='issue-body'] [class^='Box']:has(> #issue-body-viewer)::before {
  content: '';
  display: block;
  height: 38px;
  margin-top: -38px;
  position: sticky;
  top: calc(var(--base-sticky-header-height, 0px) + 56px);
  background: var(--bgColor-default, var(--color-canvas-default));
}
```

Screenshot 2025-07-13 at 19 41 29

fregante:
I think nothing is stopping us from changing the `background` color to _not_ be translucent, right? It's only "annoying" because it needs 3/4 rules to cover its various states. Maybe it really only needs two CSS rules:

- regular
- own

fregante:
> it's not possible to target the headers of your own comments

GitHub does it, but only on PRs

Screenshot 12

I'm totally ok on piggybacking on `show-names` to add a class like `rgh-own-comment` so you can target it. That feature already detects and ignores them, so it can add the class and ignore them.

The drawback is that if users disable `show-names` they get the bad background on `sticky-comment-header`. I'm ok with that.

SunsetTechuila:
> The drawback is that if users disable [show-names](https://github.com/refined-github/refined-github/blob/main/source/features/show-names.tsx) they get the bad background on [sticky-comment-header](https://github.com/refined-github/refined-github/blob/main/source/features/sticky-comment-header.tsx). I'm ok with that.

Maybe we should add `dependsOnFeatures([id])` parameter to `features.addCssFeature` and `features.add`?

SunsetTechuila:
> I'm totally ok on piggybacking on [show-names](https://github.com/refined-github/refined-github/blob/main/source/features/show-names.tsx) to add a class like rgh-own-comment so you can target it. That feature already detects and ignores them, so it can add the class and ignore them.

In my opinion, this fits the dictionary definition of "spaghetti code". It might be better to create a helper to mark various elements on different pages.

fregante:
Mamma mia, i like it spaghetti

SunsetTechuila:
Do you dislike the suggested implementation, or the idea of explicitly connecting the two features?

fregante:
Let me explain my reasoning:

- I think all features make sense for everyone (or most people) and I think they should be disabled when broken or causing issues (performance, etc)
- While I want features to be independent, I can consider scenarios where they can help/augment each other
- Since this is very rare, I don't want to set up the whole thing to declare features as dependent on each other: you'd have to parse this from the .tsx files and show it in the options UI, and all the maintenance burden that comes with this

The drawback is just some marinara and own comments that are not as bright as they should be IF you disable `show-names`. The alternative is using listeners on pages that get long and busy, worsening performance. I'll take the spaghetti.

Maintaining the extension is difficult enough given the frequency GitHub changes UIs and the quantity of features, I try to limit "meta" complexity by keeping things as simple as possible. There are already a long list of broken features; I shouldn't really be spending any time adding more of them right now.

source/features/sticky-comment-header.css
SunsetTechuila: not only it has a proper class now, but also 

image

SunsetTechuila:
> I'm totally ok on piggybacking on [`show-names`](https://github.com/refined-github/refined-github/blob/main/source/features/show-names.tsx) to add a class like `rgh-own-comment` so you can target it. That feature already detects and ignores them, so it can add the class and ignore them.

done

Vangelis66:
I've just installed the latest artifact from 
https://github.com/refined-github/refined-github/actions/runs/16363550749?pr=8544
to test how this new feature is being developed and improved 👍 ; same system with **old H/W** as in [my previous comment](https://github.com/refined-github/refined-github/pull/8544#issuecomment-3066188600) here, that doesn't allow for HW compositing/acceleration; some quick results: 

1. Own comment inside a **PR** looks OK now, with a **non-transparent sticky header**; e.g. :
https://github.com/refined-github/refined-github/pull/8544#issuecomment-3066188600

SH-1

2. When on a repo's **issue tracker**, I see _mixed results_; if I am the **OP/issue author**, **the sticky header is now non-transparent**, which is OK in my system 👍 , e.g.: 
https://github.com/Eclipse-Community/r3dfox/issues/318#issue-3184196332

SH-2

... however, when I have commented _on someone else's original issue_, **the sticky header is still transparent**; e.g.: 
https://github.com/Eclipse-Community/r3dfox/issues/325#issuecomment-3046113949

SH-3

Hopefully the geniuses here 😉 could also find a way to tackle this small visual bug...
Many thanks ❤️ for all the hard efforts and time spent by the coders to perfect this invaluable extension 🥇 !

SunsetTechuila:
> when I have commented on someone else's original issue, the sticky header is still fully transparent

Yeah, happens sometimes because of #6554

SunsetTechuila:
generally ready to be re-reviewed

source/features/sticky-comment-header.tsx
```diff
@@ -0,0 +1,14 @@
+import './sticky-comment-header.css';
+
+import features from '../feature-manager.js';
+
+void features.addCssFeature(import.meta.url);
```
fregante, line 5: > > when I have commented on someone else's original issue, the sticky header is still fully transparent
> 
> Yeah, happens sometimes because of #6554

Since this feature is CSS, it's not affected by #6554. The perks of CSS-only features 🙌

Also I asked to depend on `show-names` specifically to avoid `backdrop-filter`, but the line is still there.

source/features/sticky-comment-header.css
```diff
@@ -0,0 +1,55 @@
...
+		position: sticky;
+		top: calc(var(--base-sticky-header-height, 0px) + 56px);
+		z-index: 1;
+	}
+
+	/* Added by `show-names` */
+	.rgh-viewer-did-author,
+	.timeline-comment.current-user .timeline-comment-header,
+	[class*="IssueBodyHeader-module__viewerDidAuthor"] {
+		background: linear-gradient(var(--bgColor-accent-muted, fuchsia)), var(--rgh-background) !important;
```
fregante, line 17: I don’t understand why this is using a gradient and `--rgh-background`

The color is defined by GitHub and it does not use gradients. Please keep things simple and native.
Screenshot

source/features/sticky-comment-header.css
```diff
@@ -0,0 +1,55 @@
+html:not([rgh-OFF-sticky-comment-header]) {
+	/* Issue body header */
+	[class^="IssueBodyHeader-module__IssueBodyHeaderContainer"],
+	/* Issue comment header */
+		[data-testid='comment-header'],
+	/* PR body/comment header */
+		.timeline-comment-header {
+		position: sticky;
+		top: calc(var(--base-sticky-header-height, 0px) + 56px);
```
fregante, line 9: What's base-sticky-header-height? I don't see it used anywhere in the repo.


PS: there are a number of variables defined in `sticky-sidebar` so I'd hope to share the same logic for both, maybe moving them to `refined-github.css` if they're really shared. I don't know what's _required_ for this feature to work though.

source/features/sticky-comment-header.css
```diff
@@ -0,0 +1,55 @@
...
+	.rgh-viewer-did-author,
+	.timeline-comment.current-user .timeline-comment-header,
+	[class*="IssueBodyHeader-module__viewerDidAuthor"] {
+		background: linear-gradient(var(--bgColor-accent-muted, fuchsia)), var(--rgh-background) !important;
+	}
+
+	/* Show menus on top of headers */
+	.js-timeline-item:has(details:open) .timeline-comment-header {
+		z-index: 2;
+	}
```
fregante, line 23: The comment could be clearer. Without this line, it already shows on top of the header, just _not the following comments' headers._

It's useful to attach screenshots to PRs for future reference:

Screenshot

source/features/sticky-comment-header.css
```diff
@@ -0,0 +1,55 @@
...
+
+	[class^="IssueBodyHeader-module__IssueBodyHeaderContainer"],
+	[data-testid="comment-header"] {
+		border-radius: 0;
+	}
+
+	/* Issue body container */
+	[data-testid="issue-body"] [class^="Box"]:has(> #issue-body-viewer),
```
fregante, line 31: @kovsu it feels like we just dropped a selector just like this (box:has) because it was causing perf issues.

fregante (changes requested):

fregante:
Sorry for wasting your time, but this feature is not practical to write and to support given the small advantage it gives. Closing

source/features/sticky-comment-header.tsx
SunsetTechuila: nevertheless, I will answer

> Since this feature is CSS, it's not affected by https://github.com/refined-github/refined-github/issues/6554. The perks of CSS-only features 🙌

`show-names` is affected 

> Also I asked to depend on [show-names](https://github.com/refined-github/refined-github/blob/main/source/features/show-names.tsx) specifically to avoid backdrop-filter, but the line is still there.

`backdrop-filter` was removed

source/features/sticky-comment-header.css
SunsetTechuila: > I don’t understand why this is using a gradient and --rgh-background

because by using a gradient we can draw the `bgColor-accent-muted` color, which can be semi-transparent, on top of the `rgh-background`, which is solid

source/features/sticky-comment-header.css
SunsetTechuila: > What's base-sticky-header-height? I don't see it used anywhere in the repo.

That's a variable defined by GitHub

karlhorky:
Ahh too bad! I understand the point about [the React views being brittle](https://github.com/refined-github/refined-github/issues/8463#issuecomment-3094391041). Hope this changes in future.

@SunsetTechuila @Vangelis66 thanks for your effort!

> @Vangelis66: the sticky header is still fully transparent

Here's my version of the CSS that doesn't attempt to piggyback on `show-names` but rather sets everything to a default dark gray background:

- https://github.com/refined-github/refined-github/issues/8463#issuecomment-3094626175

fregante:
@Vangelis66 thanks for your QA! It has proven very valuable

Vangelis66:
> thanks for your QA! It has proven very valuable

... You're most welcome 😄 ; now back from a brief summer vacation 😜 and with access again to my _old_ (but treasured) laptop... Below is the link to the _last_/most recent **GH artifact** implementing the proposed `sticky-comment-header` feature: 

https://github.com/refined-github/refined-github/actions/runs/16389640058?pr=8544

(GH artifacts **do expire over time**, not sure about the "artifact retention" duration in this repo...)

If I'm allowed to share some last test findings (using above artifact), I'll have to note that @SunsetTechuila's implementation must have used CSS code properly supported _only in fairly recent Firefox versions_; this extension, in theory at least, currently supports the (now _previous_) **FirefoxESR 128 branch** (`128.12.0esr` was the last release; latest ESR is now **v140-based**, `140.1.0esr` has just been released); below comments refer to artifact installation on **FirefoxESR-v128.12.0** (32-bit): 

1. The header of an **own comment** inside a PR `Conversations` tab is black (when non-stickied): 



The same is true for "older" Fx versions such as `131.0.x` and `135.0.x`; in Fx `139.0.x`, the header is non-transparent blue-ish (as posted [above](https://github.com/refined-github/refined-github/pull/8544#issuecomment-3089572616))

2. If I am **the author of an issue** in a tracker, **the top (first) header** is, again, **black**: 



In Fx `139.0.x`, the header is non-transparent blue-ish (as posted [above](https://github.com/refined-github/refined-github/pull/8544#issuecomment-3089572616))

3. The header of an **own comment** inside an **issue tracker**, in **non-sticky** mode, is blue-ish, consistent with recent Fx versions: 



4. **Own-comment headers inside an issue tracker** are **fully transparent** (like glass) when in "**sticky mode**": 



I realise my test findings have little value now, because: 
1. The extension will soon _drop support_ for out-going **FxESR-128** versions (it's up to the devs, ofc)
2. This PR has been now closed 😞 ...

However, I decided to post them for "completeness" purposes 😉 ; apologies if I created unwanted noise 😊 ...

source/features/sticky-comment-header.css
SunsetTechuila: > And we have no way of knowing what the original background was

this might help in the future https://developer.mozilla.org/docs/Web/CSS/CSS_containment/Container_size_and_style_queries#container_style_queries

SunsetTechuila: also https://github.com/w3c/csswg-drafts/issues/5292

PR #8547, opened 2025-07-08
`quick-label-removal` - Restore feature on new issues

### LINKED ISSUE #8226: `quick-label-removal` is not showing on issues
### Description

It seems that the `quick-label-removal` feature broke. I have a private repository in a private GitHub.com organization.

image

As you can see the 'gear icon' is there, so I do have permission to manage labels here, but the 'remove' icon that quick-label-remove is suppose to arrive is missing.
Inspecting the css classes, I can see that the css class is: `octicon octicon-gear`, while in the codebase we seem to be looking for `.label-select-menu .octicon-gear`.

~~The weird thing is that on my open source repository here: https://github.com/Brend-Smits/Brend-Smits/issues/2 it is working just fine. The css selectors are also different.~~ I think this was working due to caching.

CSS Selectors open source repo:
https://gist.github.com/Brend-Smits/ab991dcab0e45c1f5f8dcbfaae0a7487

CSS Selectors private repo in an enterprise org:
https://gist.github.com/Brend-Smits/969762b0508d56a2900be1bb7b107210

Example repo: https://github.com/Brend-Smits/Brend-Smits/issues/2

### Extension version

24.12.20

### Browser(s) used

Microsoft Edge Version 131.0.2903.147 (Official build) (arm64)

### Token ⚠️

SunsetTechuila:
fixes #8226

### DIFF
--- a/source/features/quick-label-removal.tsx
+++ b/source/features/quick-label-removal.tsx
@@ -15,9 +15,11 @@ import {getConversationNumber} from '../github-helpers/index.js';
 import observe from '../helpers/selector-observer.js';
 import {expectToken} from '../github-helpers/github-token.js';
 
+type LabelRemovalEvent = DelegateEvent<MouseEvent, HTMLButtonElement>;
+
 // Don't cache: https://github.com/refined-github/refined-github/issues/7283
 function canEditLabels(): boolean {
-	return elementExists('.label-select-menu .octicon-gear');
+	return elementExists(['[data-testid="sidebar-section"]:has([data-testid="issue-labels"]) .octicon-gear', '.label-select-menu .octicon-gear']);
 }
 
 function getLabelList(): HTMLElement {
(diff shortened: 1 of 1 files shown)


SunsetTechuila:
Works, but: 

- https://github.com/refined-github/refined-github/issues/8226#issuecomment-3050174565
- The feature isn't activated when opening an issue page; you have to reload it
- If you try to edit labels using the list immediately after removing a label, the removed label reappears
- Using a global `isNewUI` variable would be better

fregante:
> * The feature isn't activated when opening an issue page; you have to reload it

- #6554 


> * If you try to edit labels using the list immediately after removing a label, the removed label reappears

This could be difficult to handle, but before it was fixed in the old feature, it was a huge annoyance since often I will remove labels with this feature _and then open the dropdown to **add** some._ So the feature cannot exist if its actions are undone by interacting with the dropdown.

A possible solution could be to add one more observer for the dropdown after the user interacts with the feature and then uncheck the labels that were removed.

This might feel a bit hacky though. A safer way would be to tell the React app that the data has been updated and that it should fetch it again. This would need some reverse-engineering.

SunsetTechuila:
> > * The feature isn't activated when opening an issue page; you have to reload it
> 
> * [⚠️ React-based views don't correctly update features #6554](https://github.com/refined-github/refined-github/issues/6554)

Is there anything I can do to fix this for this particular feature? I’d prefer to get more familiar with GitHub internals and the codebase before diving into something more fundamental.

fregante:
Not easy. Might be best to start from issues tagged [bug][small]

BoomchainLabs (approved):

PR #8549, opened 2025-07-08
Use a simple debouncing to ensure token validation on domain change

### LINKED ISSUE #8548: Options page checks token for github.com, even if domain is changed
### Description

When having configured to GHES, switching the domain selector from default to the custom domain, the token check fails as it still tries to check the token against GitHub Cloud (github.com), despite the token being valid for the GHES

https://github.com/user-attachments/assets/014f834a-3780-469e-95c4-a7e95cc1c753

I mean... extension://hlepfoohegkhhmjieoechaddaejaokhf/assets/options.html, I guess :D

### Extension version

25.7.1

### Browser(s) used

Edge 138.0.3351.77

### Token ⚠️

balcsida:
Created a debounced version of validateToken for domain changes. this prevents race conditions where validateToken might run before the personal-token-link host is updated by the onChange handler in options.tsx

Closes #8548

(I also ran the fix npm script, that is why other lines changed, let me know if I should drop them)

### DIFF
--- a/source/options/token-validation.tsx
+++ b/source/options/token-validation.tsx
@@ -36,6 +36,7 @@ function getApiUrl(): string {
 		? 'https://api.github.com/'
 		: `${tokenLink.origin}/api/v3/`;
 }
+
 function expandTokenSection(): void {
 	$('details#token').open = true;
 }
@@ -45,8 +46,8 @@ async function validateToken(): Promise<void> {
 	reportStatus();
 
 	if (!tokenField.validity.valid || tokenField.value.length === 0) {
-	// The Chrome options iframe auto-sizer causes the "scrollIntoView" function to scroll incorrectly unless you wait a bit
-	// https://github.com/refined-github/refined-github/issues/6807
+		// The Chrome options iframe auto-sizer causes the "scrollIntoView" function to scroll incorrectly unless you wait a bit
+		// https://github.com/refined-github/refined-github/issues/6807
 		setTimeout(expandTokenSection, 100);
 		return;
 	}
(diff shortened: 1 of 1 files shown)


fregante:
Thank you for the bugfix! It seems that this issue was not fixed correctly:

- https://github.com/fregante/webext-options-sync-per-domain/issues/10

Can you find out why it's firing too early? My guess is that `deserializeForm` is asynchronous but it cannot be awaited.

---

The syncing of the form is being awaited here:

https://github.com/fregante/webext-options-sync-per-domain/blob/c60d9eaebb9bb79bfb33654aaae047e4991349bc/index.ts#L173

`syncForm` gets the data from storage here:

https://github.com/fregante/webext-options-sync/blob/ce51ec6ac169873de8a4f843010fb7497370fd52/index.ts#L183

Then the form is updated here, via `deserialize`:

https://github.com/fregante/webext-options-sync/blob/ce51ec6ac169873de8a4f843010fb7497370fd52/index.ts#L369

And only then the "change" event is fired:

https://github.com/fregante/webext-options-sync-per-domain/blob/c60d9eaebb9bb79bfb33654aaae047e4991349bc/index.ts#L174

---

So the fix would be to apply the debouncing in `webext-options-sync-per-domain` just around the `this.#changeEventTarget.dispatchEvent(new Event('change'))` call

BoomchainLabs (approved):

BoomchainLabs (approved):

balcsida:
Hey @fregante,

Sorry for not replying earlier, I finally went on a min-vacation after many years.
I can definitely check why is it firing too early, let me check it and get back to you tomorrow.

Kind regards,
@balcsida

balcsida:
Sorry, I didn't manage to catch why it is firing early

fregante:
That's ok, I reopened https://github.com/fregante/webext-options-sync-per-domain/issues/10

PR #8570, opened 2025-07-20
`restore-file` - Ask for commit title

### LINKED ISSUE #8280: `restore-file` commit title
Is there an option to change commit format of actions that requires commits like discard changes? We would like it to be in conventional commits.

Thank you

fregante:
- Closes #8280 

## Test URLs

https://github.com/refined-github/refined-github/pull/8570/files
## Screenshot

Screenshot 10

### DIFF
--- a/source/features/restore-file.tsx
+++ b/source/features/restore-file.tsx
@@ -38,7 +38,7 @@ async function getFile(filePath: string): Promise<string | undefined> {
 	return textContent;
 }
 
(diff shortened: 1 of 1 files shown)


fregante:
Trying out jules.google.com. So far, it's pretty useless for extensions. Our failing `selectors.test` file additionally sends it into a loop or completely unrelated path instead of just letting me publish the branch.

Also most of the time-consuming part is visually verifying that something works, so I spent time trying to make it fix tasks instead of just fixing it and running it myself, faster. :(

kovsu (approved):

PR #8571, opened 2025-07-20
`sticky-sidebar` - Improve performance

fregante:
- Addresses https://github.com/refined-github/refined-github/issues/8406

Not yet verified  https://github.com/refined-github/refined-github/pull/8570#issuecomment-3093721524


fregante:
Nope. `mouseenter` fires only once, exactly as expected

PR #8572, opened 2025-07-20
Add syntax highlighting to custom CSS field

### LINKED ISSUE #7706: Add syntax highlighting to CSS field in options
We can use this font, if extensions can read it directly from GitHub, so we don’t have to bundle it: https://blog.glyphdrawing.club/font-with-built-in-syntax-highlighting/

It’s not huge though, just 52kb: https://blog.glyphdrawing.club/assets/fonts/FontWithASyntaxHighlighter-Regular.woff2

fregante:
- Closes #7706 

- [ ] Commit the woff file instead

## Test URLs

options page

## Screenshot

Not yet verified  https://github.com/refined-github/refined-github/pull/8570#issuecomment-3093721524


source/options.css
```diff
@@ -2,6 +2,11 @@
 	display: none;
 }
 
+@font-face {
+	font-family: 'Font With A Syntax Highlighter';
+	src: url('https://blog.glyphdrawing.club/assets/fonts/FontWithASyntaxHighlighter-Regular.woff2') format('woff2');
```
SunsetTechuila: maybe 
```suggestion
	src: url('https://github.com/hlotvonen/FontWithASyntaxHighlighter/raw/refs/heads/main/FontWithASyntaxHighlighter-Regular.woff2') format('woff2');
```

fregante: See task in header, the file should be bundled

fregante:
I think the font makes the page a lil more crashy crashy:

Screenshot 19

It would look like:

Screenshot 18

And at 50kb I think it's easier to just load a small editor, at least you get better editing behavior.

I'd accept a small PR to add the lightest possible editor with CSS syntax highlighting available, <10kb though

Issue #8484, opened 2025-06-19
`bugs-tab` - bugs are counted twice if "bug" type is used


fregante:
### Description

It looks like the feature is simply adding `bugsLabel.issue.count + bugsType.issue.count` even if they're most likely overlapping. cc @Pixxle

<img width="259" alt="Image" src="https://github.com/user-attachments/assets/536c42da-9494-46a4-abd2-152dfa9b2a30" />

### How to replicate the issue + URL

It's obvious on this repo because all of their issues are bugs:

https://github.com/nrwl/nx

### Extension version

25.5.18

### Browser(s) used

Safari

### Token ⚠️

- [x] I set a token in the options page. I understand that some features require a token to work.

PR #8578, opened 2025-07-24
`bugs-tab` - Stop counting bugs twice when using issue types

Abhi-Bohora:
Linked Issue: [#8484](https://github.com/refined-github/refined-github/issues/8484)

this PR solves the issue where in the `Bugs` tabs it showed the wrong number of bugs.

fix: assumes that most bug-type issues also have bug labels, so we take the larger of the two counts as our approximation

## Screenshot
screenshot of issues filtered by `(label:"type: bug" OR type:Bug) ` and it also matches `Bugs` tab both are `554`
bug-tab

### DIFF
--- a/source/features/bugs-tab.tsx
+++ b/source/features/bugs-tab.tsx
@@ -38,13 +38,15 @@ type Bugs = {
 async function countBugs(): Promise<Bugs> {
 	const {repository} = await api.v4(CountBugs) as {repository: ApiResponse};
 	const bugTypeCount = repository?.issues?.totalCount ?? 0;
-
+	
 	let label = repository?.labels?.nodes?.find(label => label.name === 'bug');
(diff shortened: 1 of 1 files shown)


fregante (approved):
I think this is a good fix for the time being, even if it's still inaccurate.

- #[]()12 - label: bug
- #[]()34 - label: bug, type: Bug
- #[]()56 - type: Bug

These are 3 bug, but the tab will show 2.

The correct fix would be to count the issues returned by the exact search query we show the user in the end.

The problem is that this further delays the tab because it needs:

1. query to find the exact bugs label used by the repo
2. query to count the issues returned by the search query

I'll merge this and open a new issue

fregante:
धेरै धन्यवाद अभि!

Issue #8546, opened 2025-07-08
`unwrap-unnecessary-dropdowns` - Add support for "re-run jobs"


fregante:
### Description

Most annoying dropdown:

<img width="258" height="144" alt="Image" src="https://github.com/user-attachments/assets/6f9753c8-bd6c-42c0-b481-763751e4801b" />

It should just be this:

```
[♻️ Re-run all] [♻️ Re-run failed]
```

There's so much space on that line but they put two burger menus next to each other 😭 
 

This could be a dedicated feature just like `close-as-unplanned` or be part of `unwrap-unnecessary-dropdowns`

### Example URLs

https://github.com/refined-github/refined-github/actions/runs/16143473286?pr=8544

PR #8580, opened 2025-07-25
`unwrap-unnecessary-dropdowns` - Support re-run jobs

kovsu:
Close: #8546

## Test URLs

https://github.com/refined-github/refined-github/actions/runs/16143473286?pr=8544

## Screenshot

https://github.com/user-attachments/assets/aa8e8702-1487-4177-b088-f10837aa8235

### DIFF
--- a/source/features/unwrap-unnecessary-dropdowns.tsx
+++ b/source/features/unwrap-unnecessary-dropdowns.tsx
@@ -36,12 +36,45 @@ async function unwrapNotifications(): Promise<void | false> {
 	button.textContent = `Group by ${button.textContent.toLowerCase()}`;
 }
 
+function unwrapRerunActions(): void {
+	const menu = $('.PageHeader-actions action-menu');
+
+	const dropdown = $('anchored-position', menu);
+	if (!dropdown) {
+		return;
+	}
+
+	const triggerButton = $('focus-group > button', menu);
+	console.log(triggerButton.textContent.trim());
+
+	// We only need to unwrap the re-run jobs menu
+	if (triggerButton.textContent.trim() !== 'Re-run jobs') {
+		return;
+	}
+
+	const container = menu.parentElement!;
+	container.classList.add('d-flex', 'gap-2');
+
+	const buttons = $$('button.ActionListContent', dropdown);
+	for (const button of buttons) {
+		button.className = 'Button--secondary Button--medium Button';
+		container.append(button.cloneNode(true));
+	}
+
+	menu.remove();
+}
+
 void features.add(import.meta.url, {
 	include: [
 		pageDetect.isNotifications,
 	],
 	deduplicate: 'has-rgh',
 	init: unwrapNotifications,
+}, {
+	include: [
+		pageDetect.isActionRun,
+	],
+	init: unwrapRerunActions,
 });
 
 /*


source/features/unwrap-unnecessary-dropdowns.tsx
```diff
@@ -36,12 +36,44 @@ async function unwrapNotifications(): Promise<void | false> {
 	button.textContent = `Group by ${button.textContent.toLowerCase()}`;
 }
 
+function unwrapRerunActions(): void {
+	const menu = $('.PageHeader-actions action-menu');
```
fregante: If you don't use observe(), this will run too early, especially on slow connection, and it won't find anything.

You can see that the other part of the feature uses elementReady for this.

source/features/unwrap-unnecessary-dropdowns.tsx
```diff
@@ -36,12 +36,44 @@ async function unwrapNotifications(): Promise<void | false> {
...
+	// We only need to unwrap the re-run jobs menu
+	if (triggerButton.textContent.trim() !== 'Re-run jobs') {
+		return;
+	}
+
+	const container = menu.parentElement!;
+	container.classList.add('d-flex', 'gap-2');
+
+	const buttons = $$('button.ActionListContent', dropdown);
+	for (const button of buttons) {
```
fregante:
```suggestion
	for (const button of $$('button.ActionListContent', dropdown)) {
```

I'd suggest using the strict version where possible. We want to move away from the loose one eventually

source/features/unwrap-unnecessary-dropdowns.tsx
```diff
@@ -36,12 +36,44 @@ async function unwrapNotifications(): Promise<void | false> {
...
+		button.className = 'Button--secondary Button--medium Button';
+		container.append(button.cloneNode(true));
+	}
+
+	menu.remove();
```
fregante: Instead of removing it completely, I'd create a new element that only appears on large windows. We do this with the "quick diff buttons" feature on PRs.

Basically it's:

- create new element with `d-none d-md-block`
- add `d-md-none` to the old dropdown

Or something like that

kovsu: We don't need to do this. It already handled bt GitHub.

CleanShot 2025-07-25 at 10 41 42

source/features/unwrap-unnecessary-dropdowns.tsx
```diff
@@ -36,12 +37,45 @@ async function unwrapNotifications(): Promise<void | false> {
 	button.textContent = `Group by ${button.textContent.toLowerCase()}`;
 }
 
+function replaceRerunDropdown(menu: Element): void {
+	const dropdown = $('anchored-position', menu);
+	if (!dropdown) {
```
fregante: Too many if else here. In what case we have an action-menu but not anchored-position?

source/features/unwrap-unnecessary-dropdowns.tsx
fregante: See comment regarding select-dom/strict

kovsu: 😅 I have to wear my glasses.

source/features/unwrap-unnecessary-dropdowns.tsx
```diff
@@ -1,9 +1,9 @@
-import {$$} from 'select-dom';
-import {$} from 'select-dom/strict.js';
+import {$, $$} from 'select-dom/strict.js';
```
fregante: Line 18 suggests that that selector could return 0 items, so it should be updated. Check for other usages as well.

source/features/unwrap-unnecessary-dropdowns.tsx
```diff
@@ -15,9 +15,6 @@ function replaceDropdownInPlace(dropdown: Element, form: Element): void {
 async function unwrapNotifications(): Promise<void | false> {
 	await elementReady('.js-check-all-container > :first-child'); // Ensure the entire dropdown has loaded
 	const forms = $$('[action="/notifications/beta/update_view_preference"]');
-	if (forms.length === 0) {
-		return false;
-	}
```
fregante, line 20: I could be wrong, but I don’t think that's accurate. This page has no dropdown: https://github.com/notifications?query=is%3Asaved

You can use `$$optional` here and maybe look into using `observe` in a separate PR

source/features/unwrap-unnecessary-dropdowns.tsx
```diff
@@ -36,17 +33,47 @@ async function unwrapNotifications(): Promise<void | false> {
...
+	for (const button of $$('button.ActionListContent', menu)) {
+		button.className = 'Button--secondary Button--medium Button';
+		container.append(button.cloneNode(true));
+	}
+
+	menu.classList.add('d-none');
+}
+
+function unwrapRerunActions(signal: AbortSignal): void {
+	observe('.PageHeader-actions action-menu', replaceRerunDropdown, {signal});
```
fregante, line 62: `observe` resolves as soon as the page is rendered with the matched element, it does not await its contents. See `Ensure the entire dropdown has loaded` above for example.

I suppose we can try it as is and fix it later if it's troublesome.

source/features/unwrap-unnecessary-dropdowns.tsx
```diff
@@ -36,17 +33,47 @@ async function unwrapNotifications(): Promise<void | false> {
...
+	}
+
+	const container = menu.parentElement!;
+	container.classList.add('d-flex', 'gap-2');
+
+	for (const button of $$('button.ActionListContent', menu)) {
+		button.className = 'Button--secondary Button--medium Button';
+		container.append(button.cloneNode(true));
+	}
+
```
fregante: Since `$$()` might throw, let's add the classes later:

```suggestion
	for (const button of $$('button.ActionListContent', menu)) {
		button.className = 'Button--secondary Button--medium Button';
		container.append(button.cloneNode(true));
	}

	container.classList.add('d-flex', 'gap-2');
```

source/features/unwrap-unnecessary-dropdowns.tsx
```diff
@@ -14,7 +14,8 @@ function replaceDropdownInPlace(dropdown: Element, form: Element): void {
 
 async function unwrapNotifications(): Promise<void | false> {
 	await elementReady('.js-check-all-container > :first-child'); // Ensure the entire dropdown has loaded
```
fregante: This should be quick to fix the tests, even if not completely accurate. Replace `elementReady` with `observe('.js-check-all-container > :first-child')`

kovsu:
Good to see this again.

CleanShot 2025-07-28 at 09 49 26

PR #8581, opened 2025-07-25
Add hover effects to release assets list

### LINKED ISSUE #8339: [Suggestion] Add background color to the mouse-hovering attachment row on the release page
### Description

Add background color to the attachment row on the release page when the mouse is hovering

Before:

Image

After:

Image

(It would be nice if the background color can be changed by user)

### Example URLs

https://github.com/PowerShell/PowerShell/releases/tag/v7.5.0

SunsetTechuila:
closes #8339

## Test URLs

https://github.com/refined-github/refined-github/releases/tag/25.7.16

## Screenshot

image


source/refined-github.css
```diff
@@ -186,3 +186,13 @@ div[data-target='readme-toc.content'] div.blob-wrapper img {
 .ActionListItem .ActionListItem-label--truncate {
 	white-space: unset !important;
 }
+
+html:not([rgh-no-navigation-highlight])
+	/* Attachment row on the release page */
+	details-toggle
+	.Box-row:has([href*='/releases/download/'], [href*='/archive/refs/']):hover {
```
fregante: Fingers crossed for this :has() 🥹 

Any better selectors?

source/refined-github.css
```diff
@@ -186,3 +186,13 @@ div[data-target='readme-toc.content'] div.blob-wrapper img {
 .ActionListItem .ActionListItem-label--truncate {
 	white-space: unset !important;
 }
+
+html:not([rgh-no-navigation-highlight])
```
fregante: Unrelated feature or it wouldn't belong to `refined-github.css`; `no-navigation-highlight` _removes_ the hover.
```suggestion
```

source/refined-github.css
```diff
@@ -186,3 +186,13 @@ div[data-target='readme-toc.content'] div.blob-wrapper img {
 .ActionListItem .ActionListItem-label--truncate {
 	white-space: unset !important;
 }
+
```
fregante, line 189: We're gonna need a comment header exactly like the others above (`description`, `related issue`, `test URL`, formatted as the others)

fregante (commented):
Thanks for the PR!

source/refined-github.css
SunsetTechuila: alternative could be

```css
turbo-frame[src*="/releases/"] details-toggle .Box-row:hover
```

source/refined-github.css
SunsetTechuila: I assume that if a user doesn't want hover effects in the repo file browser, then they wouldn't want them on the release page

source/refined-github.css
fregante: Much better, if it's consistent. I think `turbo-frame` only appears on reload, not when moving between repo tabs or releases pages. Worth a try

SunsetTechuila: `turbo-frame` is always there, but the `src` attribute doesn't have a value on reload

> Any better selectors?

No

source/refined-github.css
fregante: - This adds a hover; that feature removes it. They are opposites, and we don't have opposing features
- We don't have off-by-default features

The way I see it is that the hover makes sense here but not on issue lists:

- rows are often empty, hard to follow from left to right
- rows are not tall, hard to follow from left to right
- important content is on both sides, meaning you often want to follow from left to right
	- name on the left
	- counts on the right, and counts show which asset is the most wanted)

Whereas issue/PR lists for example:

- often long title names
- taller rows
- important content grouped left, with only the comment count on the right

source/refined-github.css
SunsetTechuila: > [We have a feature on that page already that adds the downloads, you can see the selector used there or piggyback on its logic.](https://github.com/refined-github/refined-github/issues/8339#issuecomment-3055196907)

The selector is not specific and the feature ignores source code archive rows

source/refined-github.css
```diff
@@ -186,3 +186,12 @@ div[data-target='readme-toc.content'] div.blob-wrapper img {
...
+/* Add hover effect to attachment row on the release page #8339 */
+/* Test: https://github.com/refined-github/refined-github/releases/tag/25.7.16 */
+.Box-footer .Box--condensed li:has(.octicon-package, .octicon-file-zip):hover {
```
fregante, line 192: This selector (minus `li:hover`) is already used by `release-download-count`, so let's match it. 

The only difference is that the JS feature is limited to a specific page while this selector works on the whole site.

Since it's relatively innocuous, let's try it and see if it matches unwanted items too.

PR #8584, opened 2025-07-27
`release-download-count` - Avoid conflict with asset hash

### LINKED ISSUE #8583: `release-download-count` - misaligned when hash is present
### Description

The element is supposed to be appended before the attachment size, but when the hash is present it appears before it

Image

1. Go to https://github.com/refined-github/refined-github/releases/tag/25.7.16
2. Observe

### Extension version

25.7.16

### Browser(s) used

Microsoft Edge 139.0.3405.52 (Official build) beta (64-bit)

### Token ⚠️

SunsetTechuila:
fixes #8583 

## Test URLs

https://github.com/refined-github/refined-github/releases/tag/25.7.16

## Screenshot

image


SunsetTechuila:
to make the feature 100% future-proof we can check the element text content against a regex

fregante:
Actually we have a function for that, called assertNodeContent if I remember correctly. Use assetSize.firstChild because the assertion must be on a text node for actual safety.

source/features/release-download-count.tsx
```diff
@@ -51,7 +52,8 @@ async function addCounts(assetsList: HTMLElement): Promise<void> {
 		// Place next to asset size
 		const assetSize = assetLink
 			.closest('.Box-row')!
-			.querySelector(':scope > .flex-justify-end > :first-child')!;
+			.querySelector(':scope > .flex-justify-end > span')!;
+		assertNodeContent(assetSize.firstChild, /^\d+ \w{2}$/);
```
fregante: If eslint lets you, just repeat it
```suggestion
		assertNodeContent(assetSize.firstChild, /^\d+ \w\w$/);
```

But I think this will fail when the size is 1-100 BYTEs. So probably


```suggestion
		assertNodeContent(assetSize.firstChild, /^\d+ \w\w?$/);
```

SunsetTechuila: why not `/^d+ \w{2,5}$/`?

fregante: What's 5? I don’t think it's actually shown as "bytes" if that's what you're matching

SunsetTechuila: image

SunsetTechuila: I suggest to create a release in the sandbox repo for future testing

source/features/release-download-count.tsx
```diff
@@ -51,7 +52,8 @@ async function addCounts(assetsList: HTMLElement): Promise<void> {
 		// Place next to asset size
 		const assetSize = assetLink
 			.closest('.Box-row')!
-			.querySelector(':scope > .flex-justify-end > :first-child')!;
+			.querySelector(':scope > .flex-justify-end > span')!;
```
kovsu, line 55: Why not use `$` here?

fregante: It makes it confusing to read

$(end, start.closest(middle))

Vs

start.closest(middle).query(end)

fregante: I'm thinking of adding "closest" to select-dom/strict but I don't know how to make this better without reinventing jQuery

fregante:
Looks great! Thanks for fixing this view

source/features/release-download-count.tsx
fregante: I think we can avoid this type of code by selecting `.Box-row` in the loop instead of `assetLink`

```js
	for (const asset of $$('.Box-row', assetsList)) {
		const assetLink = $('.octicon-package ~ a');
		// Match the asset in the DOM to the asset in the API response
		const downloadCount = assets[assetLink.pathname.split('/').pop()!] ?? 0;

		// Place next to asset size
		const assetSize = $(':scope > .flex-justify-end > span', asset);
```

but it needs to be verified, .Box-row is too broad.

Issue #8582, opened 2025-07-27
`new-repo-disable-projects-and-wikis` - Misaligned on new UI


fregante:
### Description

Lower priority since it's there, just misaligned. The feature can now be properly integrated in the form.

Since it's so different, it's best to create a dedicated `observe` and "add" functions so that we can just drop the old version at a later point.

<img width="913" height="1056" alt="Image" src="https://github.com/user-attachments/assets/d9245b5e-f100-464e-9a2e-e03ffee0e055" />

### How to replicate the issue + URL

https://github.com/new?use_v2_form=true

### Extension version

all

### Browser(s) used

Safari

### Token ⚠️

- [x] I set a token in the options page. I understand that some features require a token to work.

PR #8587, opened 2025-07-28
`new-repo-disable-projects-and-wikis` - Support new view

kovsu:
For: #8582 

## Test URLs

https://github.com/new?use_v2_form=true

## Screenshot

<table>

<tr>

<td>

Classic

<td>

New

<tr>

<td>

CleanShot 2025-07-28 at 15 44 41

<td>

CleanShot 2025-07-28 at 15 44 34

</table>


fregante:
Not what I had in mind, read the issue. It should be integrated better and make look native. But again this is not something I need right now. Merging without closing the issue

PR #8594, opened 2025-07-31
`update-pr-from-base-branch` - Re-add button after canceling merge

kovsu:
## Test URLs

https://github.com/refined-github/sandbox/pull/13

## Screenshot

<table>

<tr>

<td>

Before

<td>

After

<tr>

<td>

https://github.com/user-attachments/assets/d5767eac-8702-4dad-a315-956301131aa1

<td>

https://github.com/user-attachments/assets/b29b74bd-f277-4340-b214-cc3ded0c843d

</table>

### DIFF
--- a/source/features/update-pr-from-base-branch.tsx
+++ b/source/features/update-pr-from-base-branch.tsx
@@ -80,12 +80,17 @@ function createButton(): JSX.Element {
 	);
 }
 
-async function addButton(mergeBar: Element): Promise<void> {
+async function addButton(): Promise<void> {
 	if (canNativelyUpdate()) {
 		// Ideally the "canNativelyUpdate" observer is fired first and this listener isn't reached, but that is not guaranteed.
 		disableFeatureOnRepo();
 		return;
 	}
+
+	const mergeBar = $([
+		'.mergeability-details > *:last-child',
+		'[class^="MergeBox-module__mergePartialContainer"]',
+	]);
 
 	const {base} = getBranches();
 	const prInfo = await getPrInfo(base.relative);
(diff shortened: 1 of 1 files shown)


source/features/update-pr-from-base-branch.tsx
```diff
@@ -145,7 +150,7 @@ async function init(signal: AbortSignal): Promise<false | void> {
 	delegate('.rgh-update-pr-from-base-branch', 'click', handler, {signal});
 	observe([
 		'.mergeability-details > *:last-child', // Old view - TODO: Drop after June 2025
-		'[class^="MergeBox-module__mergePartialContainer"]',
+		'[aria-label="Conflicts"] [class^="MergeBoxSectionHeader-module__wrapper"] h3 + .fgColor-muted',
```
fregante: I like the idea but this is so overly specific 🥲

fregante: Related issue:
- #6554

kovsu: It's not related to the issue you mentioned. It looks like only part of the merge box was updated, but the whole box we observed wasn't refreshed. That's probably why the Update branch button is missing.

fregante: > only part of the merge box was updated

via React, hence #6554 

Basically that issue is about views that update more precisely, leaving the DOM with `rgh-seen`  classes. A lot of it could be resolved by clearing those classes when the URL changes, but not all.

fregante: I see you copied the selector from `pr-base-commits`. Can you move it to selectors.ts instead? This means future fixes can affect both features.

Only `[aria-label="Conflicts"] [class^="MergeBoxSectionHeader-module__wrapper"] h3 + .fgColor-muted` though

fregante: But actually we don't need this, you can target `MergeBoxSectionHeader`, it's already fully regenerated after cancelling the merge

fregante: I'm working on it 👍

PR #8602, opened 2025-08-04
`small-user-avatars` - Disable on own username

fregante:
@kovsu @yakov116 @busches thoughts on this change?

## Test URLs

- Any page where your username is mentioned
- https://github.com/refined-github/refined-github/issues?q=(involves%3Akovsu%20OR%20involves%3Ayakov116%20OR%20involves%3Abusches)%20is%3Apr%20sort%3Areactions-smile-desc
- https://github.com/refined-github/refined-github/releases/tag/25.7.1

## Screenshot

Screenshot

Screenshot 13

### DIFF
--- a/source/features/small-user-avatars.tsx
+++ b/source/features/small-user-avatars.tsx
@@ -5,10 +5,14 @@ import features from '../feature-manager.js';
 import observe from '../helpers/selector-observer.js';
 import getUserAvatarURL from '../github-helpers/get-user-avatar.js';
 import './small-user-avatars.css';
+import {getUsername} from '../github-helpers/index.js';
 
 function addAvatar(link: HTMLElement): void {
-	const username = link.textContent;
 	const size = 14;
+	const username = link.textContent;
+	if (username === getUsername()) {
+		return;
+	}
 
 	link.classList.add('d-inline-block', 'lh-condensed-ultra');
 	link.prepend(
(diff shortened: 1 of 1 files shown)


fregante:
The reason is that I want to match `rgh-collaborators` to the native mention style (yellow), removing the border and the avatar as well (because they don't look great with the yellow)

Screenshot 9

Screenshot 10

kovsu:
For me, the small avatar makes it much easier for me to immediately recognize that this information is about me.

fregante:
The list of changes (in separate PRs)

- `conversation-authors` - remove border from own conversations, you already know whether you're a collaborator/member of the current repo
- `conversation-authors` - drop `font-style: italics` to mark own conversations
- new `own-conversations` - mark own conversations with yellow (seen above)

I was afraid this wouldn't work that well, but actually it's not too bad:

Screenshot 14
Screenshot 15

fregante:
> For me, the small avatar makes it much easier for me to immediately recognize that this information is about me.

The intent is to match GitHub's own style for that: a yellow highlight.

My reason to keep them would be more centered around consistency. While not showing your own full name in `show-names` makes sense, here maybe makes things look missing/broken/inconsistent/misaligned. So probably won't merge this.

fregante:
Yeah let's not, my last screenshots look good enough

PR #8603, opened 2025-08-04
`conversation-authors` - Match own-user highlight to GitHub's

fregante:
- follows https://github.com/refined-github/refined-github/pull/8602

## Test URLs

https://github.com/issues

## Screenshot

Screenshot 16

Screenshot 17

## Before

<table>
<tr>
	<th>Before
	<th>After
<tr>
	<td>Screenshot 18
	<td>Screenshot 17
</table>

### DIFF
--- a/source/features/conversation-authors.tsx
+++ b/source/features/conversation-authors.tsx
@@ -28,7 +28,8 @@ async function highlightCollaborators(signal: AbortSignal): Promise<void> {
 		'.js-issue-row [data-hovercard-type="user"]',
 		'a[class^="IssueItem-module__authorCreatedLink"]',
 	], author => {
-		if (list.includes(author.textContent.trim())) {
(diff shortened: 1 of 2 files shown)


source/features/conversation-authors.tsx
```diff
@@ -28,7 +28,8 @@ async function highlightCollaborators(signal: AbortSignal): Promise<void> {
 		'.js-issue-row [data-hovercard-type="user"]',
 		'a[class^="IssueItem-module__authorCreatedLink"]',
 	], author => {
-		if (list.includes(author.textContent.trim())) {
+		const name = author.textContent.trim();
+		if (list.includes(name) && name !== getUsername()) {
```
fregante, line 32: Note that this also stops highlighting your own username when it's detected as a collab, because I don’t like how this looked (and also because it seems unnecessary)

Image

PR #8623, opened 2025-08-14
Add support for multiple cron lines

### LINKED ISSUE #8621: `github-actions-indicators` - Support multiple cron lines
### Description

Currently github-actions-indicators only gets the first cron schedule from an action's YAML, leading to inaccurate information.

While I'm not sure whether having multiple cron schedules is a majority use case, certainly tweaking this feature to support multiple lines won't have an effect on people who do only have one, except perhaps a minor latency increase for this feature.

### Example
```yaml
# action.yml
on:
  schedule:
    - cron: '15 9,11 * * 1-5'
    - cron: '15 15 * * 1-4'
```
Image

The time of this screenshot being taken was ~2pm. Ideally it should read "in 1 hour" or something similar.

### Counterpoint
Theoretically I/others could get around this issue by doing something like:
```yaml
on:
  schedule:
    - cron: '15 9,11,14 * * 1-5'
    - cron: '15 14 * * 5'
jobs:
  steps:
    - name: foo
      if: github.event.schedule != '15 14 * * 5'
```
But I feel this is needless indirection, especially since it would only be to make the feature work as expected.

### Example URLs

bcgreenwood-rs:
Fixes #8621.

The test workflow below uses the following configuration:
```yaml
on:
  schedule:
    - cron: '0 0 1 4 *'  # 01 April
    - cron: '0 0 1 12 *' # 01 December
  workflow_dispatch:
```

## Test URLs
https://github.com/bcgreenwood-rs/workflow-schedule-test/actions

## Screenshot
image
image
image

### DIFF
--- a/source/features/github-actions-indicators.tsx
+++ b/source/features/github-actions-indicators.tsx
@@ -19,7 +19,7 @@ type Workflow = {
 };
 
 type WorkflowDetails = {
-	schedule?: string;
+	schedules: string[];
 	manuallyDispatchable: boolean;
 };
 
@@ -60,6 +60,23 @@ async function getFilesInWorkflowPath(): Promise<Record<string, string>> {
 	return result;
 }
 
+function extractCronExpressions(input: string): string[] {
+	// Limit our search to just the first schedule block
+	const startRe = /(?:^|\r?\n) {2}schedule:\s*(?:#.*)?\r?\n/m;
+	const matches = startRe.exec(input);
+	if (!matches) {
+		return [];
+	};
+	const start = matches.index + matches[0].length;
+	const rest = input.slice(start);
+	const nextBoundary = rest.search(/\r?\n {2}[^ ]/);
+	const block = nextBoundary === -1 ? rest : rest.slice(0, nextBoundary);
+
+	// Find all cron lines - supports quotes and trailing comments
+	const cronRe = /^(?: {4}|\t\t)-\s*cron[:\s'"]+([^'"\n]+)/gm;
+	return Array.from(block.matchAll(cronRe), matches => matches[1]);
+};
+
 const workflowDetails = new CachedFunction('workflows-details', {
 	async updater(): Promise<Record<string, Workflow & WorkflowDetails>> {
 		const [workflows, workflowFiles] = await Promise.all([getWorkflows(), getFilesInWorkflowPath()]);
@@ -74,11 +91,10 @@ const workflowDetails = new CachedFunction('workflows-details', {
 				continue;
 			}
 
-			// Single-line regex, allows comments around
-			const cron = /^(?: {4}|\t\t)-\s*cron[:\s'"]+([^'"\n]+)/m.exec(workflowYaml);
+			const crons = extractCronExpressions(workflowYaml);
 			details[workflow.name] = {
 				...workflow,
-				schedule: cron?.[1],
+				schedules: crons,
 				manuallyDispatchable: workflowYaml.includes('workflow_dispatch:'),
 			};
 		}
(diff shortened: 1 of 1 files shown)


source/features/github-actions-indicators.tsx
```diff
@@ -124,14 +140,22 @@ async function addIndicators(workflowListItem: HTMLAnchorElement): Promise<void>
...
 
-	const nextTime = parseCron.nextDate(workflow.schedule);
-	if (!nextTime) {
+	const nextTimes = workflow.schedules
+		.map(schedule => parseCron.nextDate(schedule))
+		.filter(schedule => schedule !== undefined);
+	if (nextTimes.length === 0) {
 		return;
 	}
+	let nextTime: Date | undefined;
```
bcgreenwood-rs, line 130: This kinda looks like it could end up `undefined` at the end, which isn't the case. Perhaps I should start it off at `nextTimes[0]`.

source/features/github-actions-indicators.tsx
```diff
@@ -74,11 +91,10 @@ const workflowDetails = new CachedFunction('workflows-details', {
 				continue;
 			}
 
-			// Single-line regex, allows comments around
-			const cron = /^(?: {4}|\t\t)-\s*cron[:\s'"]+([^'"\n]+)/m.exec(workflowYaml);
```
fregante, line 78: You can just use matchAll and none of the extra function, regex unchanged other than the extra `g` flag:


Screenshot 5


But since we want to extract just the values:

```js
Object.values(Object.fromEntries([...yml.matchAll(/^(?: {4}|\t\t)-\s*cron[:\s'"]+([^'"\n]+)/mg)]))
```

Screenshot 6

source/features/github-actions-indicators.tsx
```diff
@@ -124,14 +140,22 @@ async function addIndicators(workflowListItem: HTMLAnchorElement): Promise<void>
...
+		.filter(schedule => schedule !== undefined);
+	if (nextTimes.length === 0) {
 		return;
 	}
+	let nextTime: Date | undefined;
+	for (const possibleTime of nextTimes) {
+		if (!nextTime || possibleTime.getTime() < nextTime.getTime()) {
+			nextTime = possibleTime;
+		}
+	}
```
fregante, line 136: These many loops could be merged into:

```ts
let nextTime: Date | undefined;

for (const schedule of workflow.schedules) {
  const time = parseCron.nextDate(schedule);
  if (time && (!nextTime || time < nextTime)) {
    nextTime = time;
  }
}
```
or this if `<` fails with Date

```ts
let nextTime: Date | undefined;

for (const schedule of workflow.schedules) {
  const time = parseCron.nextDate(schedule);
  if (time && (!nextTime || time.getTime() < nextTime.getTime())) {
    nextTime = time;
  }
}
```

source/features/github-actions-indicators.tsx
bcgreenwood-rs: I was considering doing just that since it is by far the easier option, but figured the added robustness of not potentially matching cron expressions that aren't actually part of the schedule block may be appreciated. Happy to use the simpler approach if you think that's fine.

fregante: Yes. The regex is good enough. We can tweak it if it ever matches anything other than that block.

The only improvement here would be to alter the regex so it only returns the value, so that you only need `[...x.matchAll(y)]`

Any more complexity than this and it's better to just load a yaml parser (which I still wouldn't do)

bcgreenwood-rs: No worries then; I think your suggestion for improving the regex wouldn't be possible since `matchAll` always returns `[match, group]` pairs so even if you used lookbehinds or something you could at best end up with the match and the group being identical. Unless there's some behaviour there I'm unaware of.

Also, I've opted for the format `[...matchAll].map(v => v[1])` as it benchmarked about 20% faster than your suggestion and also `Array.from`.

fregante: > always returns `[match, group]` pairs

Yes and no:
```ts
const string = "hello 13, say hello 234 again, hello 4567!"
const regex = /(?<=hello\s+)\d+/g
[...string.matchAll(regex)].flat()
// ["13", "234", "4567"]
```



> I've opted for the format `[...matchAll].map(v => v[1])`

Fair

Issue #1426, opened 2018-07-19
Keep notifications icon up to date


borekb:
I often have tens of GitHub tabs open and GitHub somehow isn't good at keeping the notifications icon up to date: I often see the blue marker even if I read all the notifications in some other tab.

Would it be in the scope of this extension to periodically poll GitHub and clear the bell icon when all notifications have been read?

tanmayrajani:
[NPG](https://github.com/tanmayrajani/notifications-preview-github) does that already. It polls for notifications and keeps the number updated. Plus gives same page pop-overs to quickly glance/act on notifications. See if it helps! Cheers :)

yakov116:
@borekb um I have been using it for quite a few months and have not seen the issues you are experiencing. 

Why don't you open an issue there?

borekb:
@yakov116 because I'm not a user of NPG :) The scope of this feature request is much smaller than the ambition of NPG, and personally I want to stay close to the default GitHub experience. I think that the proposed visual fix would fit into Refined GitHub quite nicely, though I'm not sure if it's easy to implement.

yakov116:
@tanmayrajani correct me if I'm wrong. I think the only way to do that would be to make an api request every few seconds. 

Which would mean that this feature would require a token (in order not to max out requests )

I am not 100% sure I am correct

tanmayrajani:
> Which would mean that this feature would require a token

Yes, but not if we `fetch` the `/notifications` page every few seconds (like we do with NPG).

bfred-it:
@borek NPG has pretty much the same maintainers as this extension and technically already does what you ask.

The only thing missing is the option to disable the dropdown entirely, to just leave the number.

borekb:
The number is actually another feature I don't want :)

notlmn:
From what I know and have experienced, GitHub already does this (It does update the notification icon, and removes the blue dot if there are no notifications left to read).

borekb:
@notlmn I suspect it has something to do with having many tabs open and putting the computer to sleep occasionally. GitHub certainly _should_ hide the blue dot as you suggest but in practice, I often click a bell with a blue dot in some "old" tab only to find out that there are no actual notifications to read.

bfred-it:
> putting the computer to sleep occasionally

Yeah that's it; GH closes the websocket once the computer sleeps and the pages no longer update... sometimes.

sindresorhus:
[...]

PR #8638, opened 2025-08-23
`unread-anywhere` - Add button

fregante:
- Closes https://github.com/refined-github/refined-github/issues/8619

The notification buttons is a rather vital part of GitHub so changes here need to be careful. 

I thought about using a larger blue bubble but that is both visually distracting and not really clickable (it's `::before`, which you can't detect in a `click` event)

This style was one of the ones [initially suggested](https://github.com/refined-github/refined-github/issues/7442#issuecomment-2168340282).

Also I left it visible even if there are no notifications, because GitHub's UI is not necessarily up to date (#1426)

## Screenshot

Screenshot 3

Screenshot 6

### DIFF
--- a/source/features/unread-anywhere.tsx
+++ b/source/features/unread-anywhere.tsx
@@ -60,7 +60,7 @@ function addButton(nativeLink: HTMLAnchorElement): void {
 			style={{width: 10}}
 			aria-label="Open unread notifications"
 		/>
-	)
+	);
 	nativeLink.before(button);
 	button.classList.add(
 		...classes,


FloEdelmann:
To me, that style looks like a broken button (or a missing icon). I'll disable the feature anyway (because I have too many unread notifications for it to be useful), so I don't care enough to open an issue about it.

ewjoachim:
I don't plan on using this feature, because to me, keeping unread notifications unread is an important part of making sure I read them at the right time, and I have a mix of personal and professional notifications, so I never want to mix them.
I find the position of the button a bit annoying, as a single pixel can be the difference between the button I use 60 times per day, and the button I must never click lest it will change something that cannot be undone (make notifs as read).
(I wonder if there are accessibility guidelines for how small and close together 2 buttons may be before it starts to be really hard to some users?)

There was a suggestion on making it on alt+click that was set aside as "too hard to discover to be useful". What about changing the tooltip of the button to "You have unread notifications. Use alt+click to open them all"

That said, I've disabled the feature. I think it's the first time I had to do that, and I'm perfectly fine doing that, but I wanted to offer some feedback so that there's a chance to revisit design choices. But, again, I'm perfectly fine if nothing changes.

fregante:
Thanks for the feedback! Keep them coming. 

This is the reasoning:

- the button must exist because I myself forgot I even wrote the feature for several months. Shortcuts are effectively hidden features. 
- the button cannot be too big because it causes content shift on the sidebar (e.g. to try to click the native Notifications link, but suddenly the new button appears)
- the button must be near the native link, not hidden in menus for example


This is the closest I could get:

- native style, it only lacks an icon
- its target is small enough to make the action more difficult 
- the action is non-destructive:
  - up to 10 tabs are opened and in each there's a bar to "mark as unread" again

Improvements are still possible, so I'm open to suggestions. For example:

- add a confirmation every time (annoying, but can be accepted with enter)
- limit to 5 tabs
- change style, but this is the hard part given the limitations

ewjoachim:
> in each there's a bar to "mark as unread" again

there is ?! I never saw it ! (I'm seeing it now)

fregante:
heheh yes! For future reference, this is what I see after opening your notification:

Screenshot

ewjoachim:
A suggestion: a dropdown in the style of the copilot button ?
Screenshot 2025-08-25 at 19 07 02

It means it takes 2 clicks, BUT they're much closer from one another than a confirm dialog

Edit: an additional thought: it's fairly obvious what each button in the top bar does without needing the tooltip. (Assuming you remember that the circle with dot now means "issue", but that ship has sailed). That small button now really is something you need the tooltip to have an idea what it does. It sure means "something something notifications", but it's hard to guess.

fregante:
> A suggestion: a dropdown in the style of the copilot button ?

This violates the second rule: can't be too big.

I think I'd need to go back to playing with the blue indicator, but without a blue center because it makes it too visually distracting

Screenshot 1

- small/native enough
- it doesn't shift content at all

I think there's no way to fulfill these two constraints at the same time:

- looks 100% native
- doesn't shift content

This is basically the only option, but it still looks weird to have a "duplicate" notifications action so far away from the notifications.

Screenshot


> it's hard to guess.

As you said, the "too many icons" ship has sailed. You will find out what it means by hovering it.

fregante:
See https://github.com/refined-github/refined-github/pull/8642

PR #8648, opened 2025-08-27
Fix "token is for a different user" false positive

fregante:
- Fixes https://github.com/refined-github/refined-github/issues/8408


source/github-helpers/index.ts
```diff
@@ -6,12 +6,10 @@ import type {RequireAtLeastOne} from 'type-fest';
 import * as pageDetect from 'github-url-detection';
 import mem from 'memoize';
 
-import onetime from '../helpers/onetime.js';
```
fregante, line 9: The core of the change is this. The function is no longer cached on first call. Whichever feature that used to call it _before it's ready_ is still broken, but we'll fix it once that's found and reported.

PR #8661, opened 2025-09-05
`show-names` - Exclude username in nested item

kovsu:
## Test URLs

https://github.com/dashboard

## Screenshot

<table>
<tr>
 <td>
Before
 <td>
After
<tr>
 <td>
CleanShot 2025-09-05 at 11 01 34@2x
 <td>
CleanShot 2025-09-05 at 10 58 19@2x

</table>

### DIFF
--- a/source/features/show-names.tsx
+++ b/source/features/show-names.tsx
@@ -39,6 +39,7 @@ function appendName(element: HTMLAnchorElement, fullName: string): void {
 
 async function updateLinks(found: HTMLAnchorElement[]): Promise<void> {
 	const users = Map.groupBy(
+		// Exclude nested items https://github.com/refined-github/refined-github/pull/8661
 		found.filter(element => element.textContent.trim() === element.href.split('/').pop()),
 		element => element.textContent.trim(),
 	);


kidonng:
Fix the selector instead, or update it?

https://github.com/refined-github/refined-github/blob/581643d8453220b644df5ff338f59f7aab9f7580/source/github-helpers/selectors.ts#L225-L228

kovsu:
Better do this with js not selector i think.

source/features/show-names.tsx
```diff
@@ -38,7 +38,23 @@ function appendName(element: HTMLAnchorElement, fullName: string): void {
...
+			continue;
+		}
+
+		if (users.has(username)) {
+			users.get(username)!.push(element);
+		} else {
+			users.set(username, [element]);
+		}
+	}
+
```
scarf005:
```suggestion
	const users = Map.groupBy(
	    found.filter(element => element.textContent.trim() === element.href.split('/').pop()),
	    element => element.textContent.trim()
	);
```
how about this approach?

PR #8664, opened 2025-09-05
`conflict-marker` - Remove feature for repo issue list

kovsu:
> Shows which PRs have conflicts in PR lists

As the description of `conflict-marker` says, we don't need the feature in issue list page.

## Test URLs

https://github.com/CherryHQ/cherry-studio/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen

## Screen shot

Before

CleanShot 2025-09-05 at 16 59 19@2x

After

CleanShot 2025-09-05 at 16 58 57@2x

### DIFF
--- a/source/features/conflict-marker.tsx
+++ b/source/features/conflict-marker.tsx
@@ -60,7 +60,7 @@ async function init(signal: AbortSignal): Promise<void> {
 
 void features.add(import.meta.url, {
 	include: [
-		pageDetect.isIssueOrPRList,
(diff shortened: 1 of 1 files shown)


fregante:
The previous selector covers this search:

https://github.com/refined-github/sandbox/issues?q=Conflicts

fregante:
Oh I see the issue now. The selector for the feature likely needs to be adjusted to remove that icon from there. Feel free to reopen but with a selector change rather than an `include` change

PR #8668, opened 2025-09-08
`conflict-marker` - Remove feature for repo issue list

kovsu:
ref: #8664

## Test URLs

https://github.com/CherryHQ/cherry-studio/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen

## Screen shot

Before

CleanShot 2025-09-05 at 16 59 19@2x

After

CleanShot 2025-09-05 at 16 58 57@2x

### DIFF
--- a/source/github-helpers/selectors.ts
+++ b/source/github-helpers/selectors.ts
@@ -78,10 +78,7 @@ export const openPrsListLink = [
 	`,
 	// React view
 	css`
-		li[role="listitem"]:has(
-			.octicon-git-pull-request.color-fg-open,
-			.octicon-git-pull-request-draft
-		) a[data-hovercard-url*="/pull"]
+		li[role="listitem"] h3 a[data-hovercard-url*="/pull"]
 	`,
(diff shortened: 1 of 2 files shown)


source/github-helpers/selectors.ts
```diff
@@ -78,7 +78,10 @@ export const openPrsListLink = [
 	`,
 	// React view
 	css`
-		li[role="listitem"] a[data-hovercard-url*="/pull"]
+		li[role="listitem"]:has(
+			.octicon-git-pull-request.color-fg-open,
+			.octicon-git-pull-request-draft
+		) a[data-hovercard-url*="/pull"]
```
fregante: I think you just need to make the selector more specific without `has`

```suggestion
		li[role="listitem"] h3 a[data-hovercard-url*="/pull"]
```

Or even but just using the new `[data-testid="issue-pr-title-link"]` on its own

fregante (approved):

PR #8684, opened 2025-09-22
`no-modals` - Disable on projects 

### LINKED ISSUE #8681: Split existing `no-modals` Project behavior into a separate feature
### Description

It seems a [semi-recent change](https://github.githistory.xyz/refined-github/refined-github/blob/main/source/features/no-modals.tsx) to the `no-modals` makes it disable Projects' slide in pane feature. I agree with `no-modals` for all other use cases, but the project slide in is useful because it allows quickly inspecting or adjusting the settings of multiple project items without having to wait out page loads. If someone wants to open a project issue in a new tab, ctrl-click or middlemouse is barely more work.

I would appreciate it if `no-modals` for projects could be made a separate feature so I can still use `no-modals` for issues and sub-issues.

https://github.com/refined-github/refined-github/blob/544c415c5ac8a41e36f6df697fb19b4bb63e3f5f/source/features/no-modals.tsx#L12-L16

Side note: I am surprised it works on Projects at all given that ["Can you add this feature?" #7](https://github.com/refined-github/refined-github/wiki/%22Can-you-add-this-feature%3F%22#7-it-doesnt-enhance-projects) says features typically don't target Projects. Guessing it was added simply because it was easy to tack on to the existing Issues behavior.

### Example URLs

Any project works, but here's one. https://github.com/orgs/satisfactorymodding/projects/7

I noticed that the [test links](https://github.com/refined-github/refined-github/blob/main/source/features/no-modals.tsx#L36) lacks a project, feel free to add it there.

### LINKED ISSUE #8689: `no-modals`: prevents modal opening inside project boards
### Description

This settings also prevents opening modals inside the project board when clicking on existing issues on a board or table. I think this is undesired behavior because when you're looking at a board with issues, you want to just peek into the issue using a modal instead of navigating away from the project to the issue. The reason is that you want to go through issues one by one, and navigating to the actual issue instead of a modal causes this process to be more cumbersome.

I used the `Identify feature` to find the `no-modals` setting that was causing this, great feature by the way!

1. Go to https://github.com/orgs/google-gemini/projects/11?query=sort%3Aupdated-desc+is%3Aopen
2. Click on any of the issues
3. Expect a modal instead of navigating directly to the issue

### Extension version

25.8.27

### Browser(s) used

Chrome

### Token ⚠️

- [ ] I set a token in the options page. I understand that some features require a token to work.

budak7273:
Closes #8681
Closes #8689 

## Test URLs

https://github.com/orgs/satisfactorymodding/projects/8

## Screenshot

Image

Image

### DIFF
--- a/source/features/no-modals.tsx
+++ b/source/features/no-modals.tsx
@@ -13,7 +13,6 @@ function init(signal: AbortSignal): void {
 		'a[href$="/issues/new/choose"]', // New issue button
 		'a[class*="SubIssueTitle"]', // Sub-issue links
 		'a[data-testid="issue-pr-title-link"]', // Global issue list links
-		'.board-view-column-card a[href^="https:"]', // Project issue links
 	], 'click', fix, {signal, capture: true});
 }
 
@@ -22,7 +21,6 @@ void features.add(import.meta.url, {
 		pageDetect.isIssue,
 		pageDetect.isRepoIssueList,
 		pageDetect.isGlobalIssueOrPRList,
-		pageDetect.isProject,
 	],
 	init,
 });


fregante (approved):

budak7273:
Do I need to do anything about the [Labeler action fail](https://github.com/refined-github/refined-github/actions/runs/17923578097/job/50964118285?pr=8684)? I assume no

PR #8704, opened 2025-10-10
Preserve new-lines in user bio

### LINKED ISSUE #8631: Render multi-line bio properly
### Description

GitHub doesn't set the style required to properly render multi-line bio, even though it is inputed through a textarea, which to some extent is expected since most websites/apps only do single-line bio.

I learned this early but never associated it to why many users have awkward-reading bio: likely they are unaware of the bug and think it would be fine! (Then we have [people do nbsp](https://web.archive.org/web/20190814140806/https://github.com/scorphus) which is not responsive at all)

Would like to include this in `github-bugs`:

```css
.user-profile-bio,
/* Hovercard */
[aria-label="User bio"] div {
    white-space: pre-wrap;
}
```

I probably will also submit a bug report to GitHub, though it could be deemed too trivial.

### Example URLs

| User | Before | After |
|--------|--------|--------|
| https://github.com/caugner | Image | Image|
| https://github.com/peterbe | Image | Image |
| https://github.com/rs | Image | Image |

kidonng:
Resolve #8631

## Test URLs

- Profile: https://github.com/caugner
- Hovercard: https://github.com/mdn/fred

## Screenshot

| | Profile | Hovercard |
|--------|--------|--------|
| Before | Screenshot 2025-10-10 at 9 48 18 PM | Screenshot 2025-10-10 at 9 48 30 PM |
| After | Screenshot 2025-10-10 at 9 48 40 PM | Screenshot 2025-10-10 at 9 48 53 PM |

### DIFF
--- a/source/features/github-bugs.css
+++ b/source/features/github-bugs.css
@@ -117,8 +117,7 @@ svg.octicon-package + a.Truncate[href*='/releases/download/'] > .Truncate-text {
 /* Test (profile): https://github.com/caugner */
 /* Test (hovercard): https://github.com/mdn/fred */
 .user-profile-bio,
-/* Hovercard */
-[aria-label="User bio"] div {
+[aria-label='User bio'] div /* Hovercard */ {
 	white-space: pre-wrap;
 }
 
--- a/source/features/github-bugs.css
+++ b/source/features/github-bugs.css
@@ -113,6 +113,15 @@ svg.octicon-package + a.Truncate[href*='/releases/download/'] > .Truncate-text {
 	white-space: nowrap;
 }
 
+/* Preserves new-lines in user bio #8631 */
+/* Test (profile): https://github.com/caugner */
+/* Test (hovercard): https://github.com/mdn/fred */
+.user-profile-bio,
+/* Hovercard */
+[aria-label="User bio"] div {
+	white-space: pre-wrap;
+}
+
 /*
 
 Test URLs:


source/features/github-bugs.css
```diff
@@ -113,6 +113,15 @@ svg.octicon-package + a.Truncate[href*='/releases/download/'] > .Truncate-text {
 	white-space: nowrap;
 }
 
+/* Preserves new-lines in user bio #8631 */
+/* Test (profile): https://github.com/caugner */
+/* Test (hovercard): https://github.com/mdn/fred */
+.user-profile-bio,
+/* Hovercard */
+[aria-label="User bio"] div {
+	white-space: pre-wrap;
```
kidonng, line 121: I do not like adding a random max-height limit now. Not until someone abuses it  (seriously?)

source/features/github-bugs.css
```diff
@@ -113,6 +113,14 @@ svg.octicon-package + a.Truncate[href*='/releases/download/'] > .Truncate-text {
 	white-space: nowrap;
 }
 
+/* Preserves new-lines in user bio #8631 */
+/* Test (profile): https://github.com/caugner */
+/* Test (hovercard): https://github.com/mdn/fred */
```
fregante, line 118: I think in hovercard it's a mistake. GitHub also avoids line breaks when previewing issues and PRs

fregante (approved):
Can be merged after it's removed from hovercards

Issue #8092, opened 2024-11-21
Provide a quick "cancel run" button on the workflow lists (Actions)


szalapski:
### Description

Current: 

In any workflow list (in Actions), I have to click the […] button to show a menu, then choose cancel run.
![image](https://github.com/user-attachments/assets/fc123039-ef55-44c4-b67a-ad7f141533c5)

Desired: A small red "Cancel run" button in the place of the […] button.  Maybe some stop glyph, like ⏹ or🛑on the button would be appropriate rather than text, or maybe you have a better icon/glyph library to choose from.

Bonus ask: Ideally, the page wouldn't fully refresh after doing so, as I often want to cancel several runs in a row.  Not sure if this is possible.

Consider replacing all the choices in the […] button to instead be quick single buttons, thus getting rid of the […] button.  Other choices in […] currently might be "View workflow file" or "Delete workflow run".  There never seems to be more than two but I might not be diving deep enough.

### Example URLs

https://github.com/szalapski/szalapski.github.io/actions

fregante:
> Desired: A small red "Cancel run" button

It saves one click and 20px of mouse travel. Not worth it.



> the page wouldn't fully refresh after doing so, as I often want to cancel several runs in a row

This would be more useful. Perhaps it can be done via:

```jsx
delegate(click on "Cancel run", ajaxify);

function ajaxify() {
	dropdown.close()
	await showToast(async () => {
		await pushForm(button.form)
		$(progressIcon).replaceWith(<StoppedIcon />)
	})
}
```

PR welcome. It's not something I do often enough to want to work on it though, so I'll close the request in a couple of months unless a PR is sent.

SunsetTechuila:
@fregante what do you think?

<img width="1295" height="545" alt="Image" src="https://github.com/user-attachments/assets/e954e5e1-b44d-4588-af68-e063dbe49cc5" />

SunsetTechuila:
kinda like `unwrap-unnecessary-dropdowns`

fregante:
> It saves one click and 20px of mouse travel. Not worth it.

and

> I'll close the request in a couple of months unless a PR is sent.

SunsetTechuila:
> It saves one click and 20px of mouse travel. Not worth it.

It's at least 20px of mouse travel and one click each time. And doesn't this apply to most extension features? Sometimes I have to use actions a lot, and in those moments this feature would save me a ton of mouse clicks and travel in total

[...]

PR #8707, opened 2025-10-12
`unwrap-workflow-menu`: add feature

SunsetTechuila:
just a basic implementation to decide if the feature is complicated or not

#8092

## Test URLs

## Screenshot

image


fregante:
> just a basic implementation to decide if the feature is complicated or not

This layout jump is not acceptable, so so far it's more complicated than it seems:

tinymce

SunsetTechuila:
> This layout jump is not acceptable, so so far it's more complicated than it seems:

can be fixed with https://github.com/refined-github/refined-github/issues/8092#issuecomment-3394109935, but that would require creating new elements and identifying menu items

SunsetTechuila:
I've thought about that, and I agree that adding quick cancel and delete buttons is enough

PR #8708, opened 2025-10-12
`actions-run-removal` - New feature

### LINKED ISSUE #8092: Provide a quick "cancel run" button on the workflow lists (Actions)

SunsetTechuila:
resolves #8092

## Test URLs

## Screenshot

image

### DIFF
--- a/source/features/actions-run-removal.tsx
+++ b/source/features/actions-run-removal.tsx
@@ -28,7 +28,7 @@ function addQuickButtons(contextMenuIcon: HTMLElement): void {
 	button.classList = 'timeline-comment-action color-fg-muted btn-link rgh-actions-run-removal p-1';
 	$('summary', contextMenuDetails).classList.add('p-1');
 	const rightControlsContainer = contextMenuDetails.parentElement!;
-	// Prepending so that the cloned dialog shows instead of the one inside the menu, as it is hidden when the menu is closed
+	// Prepending so that the cloned dialog opens instead of the one inside the menu, as it is hidden when the menu is closed
 	rightControlsContainer.classList.add('d-flex', 'flex-column-reverse', 'mt-n2', 'mb-n2');
 	rightControlsContainer.prepend(menuItem);
 }


source/features/quick-workflow-run-removal.css
```diff
@@ -0,0 +1,3 @@
+:is(.rgh-delete-workflow-run, .rgh-cancel-workflow-run):is(:hover, :focus) {
+	color: var(--fgColor-danger, fuchsia) !important;
```
fregante: It feels like there's a class for this already. If there isn't, then it shouldn't behave this way.

build/__snapshots__/features-meta.json
```diff
@@ -678,6 +673,12 @@
 		"description": "Simplifies the UI to hide comments.",
 		"screenshot": "https://user-images.githubusercontent.com/1402241/43039221-1ddc91f6-8d29-11e8-9ed4-93459191a510.gif"
 	},
+	{
+		"id": "quick-delete-cancel-workflow",
+		"description": "Adds \"Cancel\" and \"Delete\" buttons to each workflow run row.",
```
fregante: "Lets you cancel or delete workflows in faster from the workflow list"

source/features/quick-workflow-run-removal.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+function addQuickButtons(actionRunRow: HTMLDivElement): void {
+	const contextMenuDetails = $('details:has(.octicon-kebab-horizontal)', actionRunRow);
+	const rightControlsContainer = contextMenuDetails.parentElement!;
+
+	const deleteWorkflowButton = $optional('button[data-show-dialog-id^="delete-workflow-run"]', contextMenuDetails);
+	if (deleteWorkflowButton) {
+		const dialogHelper = $('dialog-helper', deleteWorkflowButton.parentElement!);
+		dialogHelper.classList.add('text-left');
+		transformIntoIconButton(deleteWorkflowButton, <TrashIcon />);
```
fregante: Like we do for `quick-comment-edit`, clone the button, don't remove it from the dropdown. This also lets the user have a fallback in case this breaks.

`.cloneNode(true)` does this

source/features/quick-delete-cancel-workflow.tsx
```diff
@@ -0,0 +1,68 @@
...
+	const contextMenuDetails = $('details:has(.octicon-kebab-horizontal)', actionRunRow);
+	const rightControlsContainer = contextMenuDetails.parentElement!;
+
+	const deleteWorkflowButton = $optional('button[data-show-dialog-id^="delete-workflow-run"]', contextMenuDetails);
+	if (deleteWorkflowButton) {
+		const dialogHelper = $('dialog-helper', deleteWorkflowButton.parentElement!);
+		dialogHelper.classList.add('text-left');
+		transformIntoIconButton(deleteWorkflowButton, <TrashIcon />);
+		deleteWorkflowButton.classList.add('rgh-delete-workflow-run');
+		rightControlsContainer.prepend(deleteWorkflowButton, dialogHelper);
```
fregante: I'd place it below the `. . .` button, so it doesn't cause _any_ horizontal shift

source/features/quick-delete-cancel-workflow.tsx
```diff
@@ -0,0 +1,68 @@
...
+		cancelWorkflowButton.classList.add('rgh-cancel-workflow-run');
+		removeRedundantDivider(cancelForm.parentElement!);
+		rightControlsContainer.prepend(cancelForm);
+	}
+
+	rightControlsContainer.classList.toggle('d-flex', !!(cancelForm || deleteWorkflowButton));
+}
+
+function init(signal: AbortSignal): void {
+	observe('#partial-actions-workflow-runs .Box-row', addQuickButtons, {signal});
```
fregante: `observe` should just point to the element you're looking for, not its ancestors, otherwise it defeats the purpose. Since the observer won't work on hidden elements, I suppose you'll have to target the closest element (contextMenuDetails?)

Most likely the selector should be `#partial-actions-workflow-runs .Box-row details .octicon-kebab-horizontal` and then you work your way from there. `:has()` has bad performance in the observer, so we try to avoid it.

source/features/quick-delete-cancel-workflow.tsx
```diff
@@ -0,0 +1,68 @@
...
+	const cancelForm = $optional('form[action$="/cancel"]', contextMenuDetails);
+	if (cancelForm) {
+		const cancelWorkflowButton = $('button[type="submit"]', cancelForm);
+		transformIntoIconButton(cancelWorkflowButton, <SquareCircleIcon />);
+		cancelWorkflowButton.classList.add('rgh-cancel-workflow-run');
+		removeRedundantDivider(cancelForm.parentElement!);
+		rightControlsContainer.prepend(cancelForm);
+	}
+
+	rightControlsContainer.classList.toggle('d-flex', !!(cancelForm || deleteWorkflowButton));
```
fregante: Given the changes suggested, you're probably going to use something like:

```js
wrapAll(<div className="d-flex flex-column"/>, contextMenuDetails, cancelForm, deleteWorkflowButton)
```

This way they're be wrapped if they exist. And if they don't, that's likely fine, no visual change with this extra element.

source/features/quick-delete-cancel-workflow.tsx
```diff
@@ -0,0 +1,68 @@
...
+}
+
+function addQuickButtons(actionRunRow: HTMLDivElement): void {
+	const contextMenuDetails = $('details:has(.octicon-kebab-horizontal)', actionRunRow);
+	const rightControlsContainer = contextMenuDetails.parentElement!;
+
+	const deleteWorkflowButton = $optional('button[data-show-dialog-id^="delete-workflow-run"]', contextMenuDetails);
+	if (deleteWorkflowButton) {
+		const dialogHelper = $('dialog-helper', deleteWorkflowButton.parentElement!);
+		dialogHelper.classList.add('text-left');
```
fregante: Do you have to move this? What does it look like without `text-left`?

SunsetTechuila: > Do you have to move this?

Yes

> What does it look like without text-left?

image

source/features/quick-workflow-run-removal.css
fregante: You're probably looking for the "danger" button type. Like this, but "icon only"

Screenshot 20

fregante: They don't use it for GHA secrets for example:


Screenshot

SunsetTechuila: do you insist on removing the hover effect?

source/features/quick-workflow-run-removal.tsx
```diff
@@ -0,0 +1,61 @@
...
+function addQuickButtons(contextMenu: HTMLElement): void {
+	const contextMenuDetails = contextMenu.closest('details')!;
+	const rightControlsContainer = contextMenuDetails.parentElement!;
+
+	const deleteWorkflowMenuItem = $optional('li:has(> button[data-show-dialog-id^="delete-workflow-run"])', contextMenuDetails)?.cloneNode(true);
+	if (deleteWorkflowMenuItem) {
+		const deleteWorkflowButton = $('button', deleteWorkflowMenuItem);
+		transformIntoIconButton(deleteWorkflowButton, <TrashIcon />);
+		deleteWorkflowButton.classList.add('rgh-delete-workflow-run');
+		rightControlsContainer.prepend(deleteWorkflowMenuItem);
```
SunsetTechuila: it is prepended, and not appended, so that the cloned dialog opens instead of the one inside the context menu

source/features/quick-workflow-run-removal.tsx
```diff
@@ -0,0 +1,61 @@
...
+	const cancelForm = $optional('form[action$="/cancel"]', contextMenuDetails)?.cloneNode(true);
+	if (cancelForm) {
+		const cancelWorkflowButton = $('button', cancelForm);
+		transformIntoIconButton(cancelWorkflowButton, <SquareCircleIcon />);
+		cancelWorkflowButton.classList.add('rgh-cancel-workflow-run');
+		rightControlsContainer.prepend(cancelForm);
+	}
+
+	if (cancelForm || deleteWorkflowMenuItem) {
+		rightControlsContainer.classList.add('d-flex', 'flex-column-reverse');
```
SunsetTechuila: `flex-column-reverse` because https://github.com/refined-github/refined-github/pull/8708#discussion_r2426513355

kidonng:
> can't think of a better name

`quick-ci-actions`? Unfortunately GitHub already call the feature "Actions" so we can't just use `quick-<feature>-actions` 😅

But definitely not `-workflow`. You cancel/delete the _workflow runs_, not the workflows themselves.

SunsetTechuila:
Maybe `quick-workflow-run-actions`. It's not very concise, but it's descriptive

fregante:
I think we have another workflow feature. Technically it's deleting a run, so it should include that. 

"quick-run-removal" or "quick-workflow-run-removal" but I prefer the 3-word version.

SunsetTechuila:
> Unfortunately GitHub already call the feature "Actions" so we can't just use `quick-<feature>-actions`

We already have several features named like that

> quick-workflow-run-removal

But what about cancellation?

fregante:
Same same 

Names don't have to be exact. See notes in wiki https://github.com/refined-github/refined-github/wiki/Contributing#feature-name

SunsetTechuila:
renamed

fregante:
Screenshot 6

fregante:
We have `github-actions-indicators` which applies to the same page. "GitHub" is kinda redundant so I dropped it.

source/features/actions-run-removal.tsx
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	const button = $('button', menuItem);
+	button.ariaLabel = button.textContent.trim();
+	button.replaceChildren(menuItem.tagName === 'FORM' ? <SquareCircleIcon /> : <TrashIcon />);
+	button.classList = 'timeline-comment-action color-fg-muted btn-link rgh-actions-run-removal p-1';
+	$('summary', contextMenuDetails).classList.add('p-1');
+	const rightControlsContainer = contextMenuDetails.parentElement!;
+	rightControlsContainer.classList.add('d-flex', 'flex-column', 'mt-n2', 'mb-n2');
+	rightControlsContainer.append(menuItem);
```
SunsetTechuila: https://github.com/refined-github/refined-github/pull/8708#discussion_r2426513355

try clicking a delete button

SunsetTechuila: > the cloned dialog opens instead of **the one inside the context menu**

which is hidden when the menu is closed

fregante:
> `quick-ci-actions`?

I want to move away from "quick" since it's relatively redundant. Most features let you do things faster.

> do you insist on removing the hover effect?

I attempted it but I agree it's better here and it matches the red shown in the dropdown.

fregante:
>  fregante force-pushed 

argh. Since when GitHub Desktop force pushes?! Didn't mean to do that

PR #8712, opened 2025-10-19
`comments-time-machine-links` - Restore feature on issues

### LINKED ISSUE #8698: `comments-time-machine-links`: Cannot read properties of undefined (reading 'value')
### Description

I've notcied that time machine links disappeared

Image

To test it I enabled debug mode 

Image

And I've found a console error - could be the possible cause?
```
TypeError: Cannot read properties of undefined (reading 'value')
    at observe.signal.signal (chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/assets/features/comments-time-machine-links.js:133:80)
    at globalThis.addEventListener.once.once (chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/assets/helpers/selector-observer.js:87:3)
```

https://github.com/refined-github/refined-github/issues/8693

### Extension version

25.10.2

### Browser(s) used

Microsoft Edge Version 141.0.3537.57 (Official build) (64-bit)

### Token ⚠️

SunsetTechuila:
fixes #8698 

## Test URLs

https://github.com/refined-github/refined-github/issues/8698

## Screenshot

image

### DIFF
--- a/source/features/comments-time-machine-links.tsx
+++ b/source/features/comments-time-machine-links.tsx
@@ -1,5 +1,5 @@
 import React from 'dom-chef';
-import {$, $$optional} from 'select-dom/strict.js';
+import {$, $$optional, $optional} from 'select-dom/strict.js';
 import elementReady from 'element-ready';
 import * as pageDetect from 'github-url-detection';
 import delegate, {type DelegateEvent} from 'delegate-it';
@@ -117,19 +117,29 @@ function addDropdownLink(menu: HTMLElement, timestamp: string): void {
 async function addDropdownLinkReact({delegateTarget: delegate}: DelegateEvent): Promise<void> {
 	const timestamp = $('relative-time[datetime]', delegate.closest('[class^="Box"]')!).attributes.datetime.value;
 	const menu = (await elementReady('[class^="prc-ActionList-ActionList"]'))!;
-	const divider = $('[data-component="ActionList.Divider"]', menu).cloneNode();
-	const menuItem = $('[class^="prc-ActionList-ActionListItem"]:has(> a)', menu).cloneNode(true);
+	const divider = $optional('[data-component="ActionList.Divider"]', menu)?.cloneNode();
+	const menuItem = $('[class^="prc-ActionList-ActionListItem"]', menu).cloneNode(true);
 
-	const link = $('a', menuItem);
-	link.href = buildRepoURL(`tree/HEAD@{${timestamp}}`);
-	link.removeAttribute('aria-keyshortcuts');
-	link.removeAttribute('target');
-	link.title = 'Browse repository like it appeared on this day';
+	menuItem.removeAttribute('aria-keyshortcuts');
+	menuItem.role = 'none';
+	const menuItemContentWrapper = $('[class^="prc-ActionList-ActionListContent"]', menuItem);
+	const link = (
+		<a
+			href={buildRepoURL(`tree/HEAD@{${timestamp}}`)}
+			className={menuItemContentWrapper.className + ' ' + linkifiedURLClass}
+			role="menuitem"
+			title="Browse repository like it appeared on this day"
+			aria-keyshortcuts="v"
+		>
+		</a>
+	);
+	link.append(...menuItemContentWrapper.childNodes);
+	menuItemContentWrapper.replaceWith(link);
 	$('[class^="prc-ActionList-ItemLabel"]', menuItem).textContent = 'View repo at this time';
 	$('[class^="prc-ActionList-LeadingVisual"]', menuItem).replaceChildren(<HistoryIcon />);
 
 	menu.append(
-		divider,
+		divider ?? '',
 		menuItem,
 	);
 }
@@ -154,7 +164,7 @@ async function init(signal: AbortSignal): Promise<void> {
 		addDropdownLink(menu, timestamp);
 	}, {signal});
 
-	delegate(`:is(${commentSelector}) button:has(> .octicon-kebab-horizontal)`, 'click', addDropdownLinkReact, {signal});
+	delegate(`:is(${commentSelector}) button[data-component="IconButton"]:has(> .octicon-kebab-horizontal)`, 'click', addDropdownLinkReact, {signal});
 
 	observe(
 		`:is(${commentSelector}) relative-time[datetime]`


source/features/comments-time-machine-links.tsx
```diff
@@ -131,7 +131,8 @@ async function init(signal: AbortSignal): Promise<void> {
 		'div.react-issue-body', // First comment
 		'[data-testid="review-thread"] > div',
 	], comment => {
-		const timestamp = $('relative-time', comment).attributes.datetime.value;
+		const relativeTime = $('relative-time', comment);
+		const timestamp = relativeTime.attributes.datetime?.value ?? new Date(`${relativeTime.textContent} UTC`).toISOString();
```
fregante: So I wonder: does the DOM get updated with this attribute eventually? Or it's just missing in this case?

SunsetTechuila: It does get updated

fregante: Then let's wait! 😃 just update the observed selector to include the attribute

fregante: That's actually not straightforward, let me think about it

fregante: This can actually be done in a more efficient way by turning this observer into a mousedown event listener. We do something similar for some label links for the "sort by updated time" feature. 

I think you'd have to merge the selectors from this observer and the loop inside addInlineLinks. 

Does this make sense?

SunsetTechuila: > Then let's wait! 😃 just update the observed selector to include the attribute


https://github.com/refined-github/refined-github/issues/8698#issuecomment-3398578807:

> 3. https://github.com/SunsetTechuila/refined-github/commit/801d3634522ac7d1a8031fed89dfbb74a88dc6f5

SunsetTechuila: > Does this make sense?

Yes, but I'd rather just add `[datetime]` to selectors

fregante: > > Then let's wait! 😃 just update the observed selector to include the attribute
> 
> [#8698 (comment)](https://github.com/refined-github/refined-github/issues/8698#issuecomment-3398578807):
> 
> > 3. [SunsetTechuila@801d363](https://github.com/SunsetTechuila/refined-github/commit/801d3634522ac7d1a8031fed89dfbb74a88dc6f5)

I don't know what you're asking. But that kind of DRY is excessive here because it makes it difficult to actually read the selectors. If you see the selectors in that file, they're all longer than just a class

fregante: > > Does this make sense?
> 
> Yes, but I'd rather just add `[datetime]` to selectors

Just commit and let's see what it looks like. I don't want to discuss code that lives in our minds without seeing it work.

SunsetTechuila: commited

SunsetTechuila: or we can do `const relativeTime = await elementReady('relative-time[datetime]', comment);`

source/features/comments-time-machine-links.tsx
```diff
@@ -15,6 +15,10 @@ import GetCommitAtDate from './comments-time-machine-links.gql';
 import {expectToken} from '../github-helpers/github-token.js';
 import getDefaultBranch from '../github-helpers/get-default-branch.js';
 
+const reactIssueComment = 'div.react-issue-comment'; // First comment
+const reactIssueBody = 'div.react-issue-body'; // Comments
+const reactCommitView = '[data-testid="review-thread"] > div';
+
```
fregante, line 25: Can be merged into a [].join() string. It can then be used as `:is(${the var}) relative-date[the attribute]`

SunsetTechuila: so you prefer this over https://github.com/refined-github/refined-github/pull/8712#discussion_r2483716370?

fregante: Yes

SunsetTechuila:
> partially _fixes https://github.com/refined-github/refined-github/issues/8698

should I open a separate PR that adapts the feature for the react dropdowns?

fregante:
Why is it a partial fix? What does it take to fix it? It would be great to just fix it in a single PR if possible

SunsetTechuila:
> Why is it a partial fix?

This fixes a bug where the `rgh-link-date` parameter wasn't appended to links from the issue body. However, it doesn't fix the absence of the "View repo at this time" menu item in react context menus

> It would be great to just fix it in a single PR if possible

Ok

SunsetTechuila:
let me know if you have any ideas for how I can integrate my code into the existing one

source/features/comments-time-machine-links.tsx
```diff
@@ -106,6 +114,36 @@ function addDropdownLink(menu: HTMLElement, timestamp: string): void {
...
+			role="menuitem"
+			title="Browse repository like it appeared on this day"
+			aria-keyshortcuts="v"
+		>
+		</a>
+	);
+	link.append(...menuItemContentWrapper.childNodes);
+	menuItemContentWrapper.replaceWith(link);
+	$('[class^="prc-ActionList-ItemLabel"]', menuItem).textContent = 'View repo at this time';
+	$('[class^="prc-ActionList-LeadingVisual"]', menuItem).replaceChildren(<HistoryIcon />);
```
SunsetTechuila, line 139: would it be better to create a new menu item from scratch rather than modifying a cloned one?

source/features/comments-time-machine-links.tsx
```diff
@@ -106,6 +114,36 @@ function addDropdownLink(menu: HTMLElement, timestamp: string): void {
 	);
 }
 
+async function addDropdownLinkReact({delegateTarget: delegate}: DelegateEvent): Promise<void> {
+	const timestamp = $('relative-time[datetime]', delegate.closest('[class^="Box"]')!).attributes.datetime.value;
+	const menu = (await elementReady('[class^="prc-ActionList-ActionList"]'))!;
```
fregante, line 119: Note that `elementReady` acts as a `querySelector` [once the page has already loaded](https://github.com/sindresorhus/element-ready#stopondomready), so it seems to me that `$()` would work the same way here?

Element-ready is to be considered "legacy"/not-ideal. The observer is what we use now to wait for elements. I can see however why this was done: there's no easy way to link the dropdown menu to the comment. So this solution could be ok and required.

In this case it seems that every comment, including my own, has a "report comment" link that references the comment (see screenshot). This also works for the top comment, which has no reference to a comment ID but you can get the "top comment" easily.

Image

Whether changing it to use this is worth it is up to you. My preference goes:

1. performance first (no `:has()` in `observe` if we can avoid it)
2. simple code (one selector in `observe` vs select + select + select + etc)

source/features/comments-time-machine-links.tsx
```diff
@@ -106,6 +114,36 @@ function addDropdownLink(menu: HTMLElement, timestamp: string): void {
 	);
 }
 
+async function addDropdownLinkReact({delegateTarget: delegate}: DelegateEvent): Promise<void> {
+	const timestamp = $('relative-time[datetime]', delegate.closest('[class^="Box"]')!).attributes.datetime.value;
```
fregante, line 118: When traversing is needed, prefer just using querySelector, this is easier to follow:

```js
delegate.closest('[class^="Box"]')!.querySelector('relative-time[datetime]')
``

File: source/features/comments-time-machine-links.tsx (hunk 3 above)
Comment (fregante) [reply source/features/comments-time-machine-links.tsx:139]: Which one is easier? Which one is possible? With generated classes, cloning is becoming the only possible solution.

File: source/features/comments-time-machine-links.tsx [hunk 6]
```diff
@@ -106,6 +114,36 @@ function addDropdownLink(menu: HTMLElement, timestamp: string): void {
...
+	const menu = (await elementReady('[class^="prc-ActionList-ActionList"]'))!;
+	const divider = $optional('[data-component="ActionList.Divider"]', menu)?.cloneNode();
```
fregante, line 120: When is this optional?

source/features/comments-time-machine-links.tsx
SunsetTechuila: > In this case it seems that every comment, including my own, has a "report comment"

I thought that too, but no: https://github.com/refined-github/sandbox/issues/108

image

source/features/comments-time-machine-links.tsx
SunsetTechuila: https://github.com/refined-github/sandbox/issues/108

fregante: I can see it

Screenshot 3

fregante (approved):
Looks good. Any improvements can be a followup PR. I'll release a new RGH version today. There's one small issue that I'm looking into

fregante:
> There's one small issue that I'm looking into

When I hovered over it, this link did not have the URL parameter, then I refreshed the page and it appeared. My guess is that GitHub updated the contents of the comment but did not remove our "rgh-seen" class. You can confirm this by opening https://github.com/refined-github/refined-github/issues/8698 and waiting a few minutes.

Screenshot 5


This is basically the risk we run into when we `observe` the wrong element. The observer can track seen elements and re-run when they're regenerated, but if we're _looking at the wrong element_ then we can't see what changed elsewhere.

This just explains the _theory,_ in practice sometimes we have no choice and/or React still updates the contents/textnodes without removing our rgh-seen class.

source/features/comments-time-machine-links.tsx
SunsetTechuila: This is what I meant by "from scratch":

```ts
	const newMenuItem =
	<li
		className={[...menuItem.classList].find(c => c.startsWith('prc-ActionList-ActionListItem'))}
		role="none"
	>
		<a
			href={buildRepoURL(`tree/HEAD@{${timestamp}}`)}
			className={menuItemContentWrapper.className + ' ' + linkifiedURLClass}
			role="menuitem"
			title="Browse repository like it appeared on this day"
			data-size={menuItemContentWrapper.dataset.size}
			aria-keyshortcuts="v"
		>
			<span className={menuItemIcon.className}>
				<HistoryIcon />
			</span>
			<span className={menuSubContent.className}>
				<span className={menuItemLabel.className}>View repo at this time</span>
			</span>
		</a>
	</li>
```

fregante: To me that doesn't look easier or shorter. You still have to select the original elements to get their classes, so why not just select them and replace their contents?

SunsetTechuila: > cloning is becoming the only possible solution.

We can use the `@primer/react` package, but you would probably say that this will bloat the extension

Another option:

```js
const require = webpackChunk_github_ui_github_ui.push([[Symbol()], {}, re => re]);
const chunks = Object.entries(require.m);
const cache = Object.keys(require.m).map(id => require(id));
const modules = cache
	.filter(module => typeof module === "object")
	.flatMap(module => {
		try {
			return Object.values(module);
		} catch {}
	});
const { LinkItem } = modules.find(m => m?.displayName === "ActionList");
```

SunsetTechuila: > To me that doesn't look easier or shorter.

Yes, I just find it easier to read and follow

SunsetTechuila: I mean that you don't have to see the DOM to understand what the code expects it to be

fregante: I see more disadvantages than advantages. The structure can get out of sync anyway and now we have two points of failure:

- select original elements
- replicate the whole DOM

SunsetTechuila:
> My guess is that GitHub updated the contents of the comment but did not remove our "rgh-seen" class.

No, it's just that `.markdown-body` has been re-rendered

SunsetTechuila:
I swear GitHub starts behaving differently every time I think I've figured out what it does

fregante:
> > updated the contents of the comment but did not remove our "rgh-seen" class.
> 
> No, it's just that `.markdown-body` has been re-rendered

It's the same thing. "re-rendering markdown-body` is the same as "update the (DOM) contents of the comment"

SunsetTechuila:
> React still updates the contents/textnodes without removing our rgh-seen class.

This doesn't happen. New elements are created, and observers work as expected. You can check and see that the caller id changes when the comment body re-renders

fregante:
> the caller id changes

Caller IDs don't change, they're static across browser reloads and are connected to a specific `observe` call. If it "changes", then it's because a different selector matched it, completely different `observe` call or different feature.

> This doesn't happen

I've been working on Refined GitHub for 8 years and I wrote the observer, I know what's happening.

https://github.com/refined-github/refined-github/blob/4be765a6af58a43ceff58b675a688cf09ee2bdf1/source/features/comments-time-machine-links.tsx#L169-L176

This observer tracks the `relative-time` element and then traverses the DOM until it finds the links. If `relative-time` remains unchanged but the contents change, no `rgh-link-date` parameter is added; Refined GitHub cannot see the changes of _elements that are not being observed._

The `rgh-seen` class you see on each link is not added by `comments-time-machine-links`. The links are observed by other features.

Verifying this is pretty simple:

1. disable all features (there's a toggle)
2. enable only `comments-time-machine-links`
3. look at this inline line: https://github.com/refined-github/refined-github/blob/main/source/features/comments-time-machine-links.tsx

You will see the `rgh-seen` class on `relative-time` but not on the link.

So, again: 

> GitHub updated the contents of the comment but did not remove our "rgh-seen" class

> but if we're _looking at the wrong element_ then we can't see what changed elsewhere


The fix is to change that observer to select the individual links in the comments rather than the `relative-time` element that sits outside the updatable area.

SunsetTechuila:
Alright, thank you for the detailed explanation

source/features/comments-time-machine-links.tsx
```diff
@@ -126,14 +164,16 @@ async function init(signal: AbortSignal): Promise<void> {
...
 
-	observe([
-		'div.react-issue-comment', // Comments
-		'div.react-issue-body', // First comment
-		'[data-testid="review-thread"] > div',
-	], comment => {
-		const timestamp = $('relative-time', comment).attributes.datetime.value;
-		addInlineLinks(comment, timestamp);
-	}, {signal});
+	delegate(`:is(${commentSelector}) button[data-component="IconButton"]:has(> .octicon-kebab-horizontal)`, 'click', addDropdownLinkReact, {signal});
```
SunsetTechuila, line 167: they added test id `comment-header-hamburger`

PR #8720, opened 2025-10-24
`convert-release-to-draft` - Fix button size

u01jmg3:
## Test URLs

- When viewing a repo with write permission: https://github.com/:org/:repo/releases/tag/v1.0.0

## Screenshot

Before:

before

After:

after

### DIFF
--- a/source/features/convert-release-to-draft.tsx
+++ b/source/features/convert-release-to-draft.tsx
@@ -52,7 +52,7 @@ function attachButton(editButton: HTMLAnchorElement): void {
 	editButton.before(
 		<button
 			type="button"
-			className="btn btn-sm ml-3 mr-1 rgh-convert-draft"
+			className="Button Button--secondary Button--small ml-3 mr-1 rgh-convert-draft"
 		>
 			Convert to draft
 		</button>,


source/features/convert-release-to-draft.tsx
```diff
@@ -52,7 +52,7 @@ function attachButton(editButton: HTMLAnchorElement): void {
 	editButton.before(
 		<button
 			type="button"
-			className="btn btn-sm ml-3 mr-1 rgh-convert-draft"
+			className="Button Button--secondary Button--small ml-3 mr-1 rgh-convert-draft"
```
fregante, line 55: If they dropped the .btn classes were about to have a bad time across the website 🥲

u01jmg3: I thought the same. I also thought BEM was less common nowadays.

SunsetTechuila: > If they dropped the .btn classes

No, it's just that the container got bigger

image

SunsetTechuila, line 55:
```suggestion
			className="Button Button--secondary Button--small ml-2 rgh-convert-draft"
```

| Before | After |
| --- | --- |
| Image | Image |

PR #8740, opened 2025-11-07
Support Firefox containers when opening tabs

fregante:
- oughta fix https://github.com/refined-github/refined-github/issues/8657

## Test URLs


prplecake:
I think this should have updated the permissions for the extension. I'm getting an error when trying to open multiple notifications at once.

> Uncaught (in promise) Error: No permission for cookieStoreId: firefox-default

I believe the extension now requires "cookies" permission to read the store ID.

PR #8743, opened 2025-11-07
`user-local-time` - Drop feature

fregante:
- Will close https://github.com/refined-github/refined-github/issues/8729

The change by GitHub is most likely intentional. I don't see a way to find the user's timezone without access to their commits.

The feature can be reinstated once/if there's a way to do it.

I will likely also disable the feature via hotfix to avoid noisy `timezone unknown` labels


kovsu (approved):
😢 T.T

PR #8752, opened 2025-11-09
`quick-repo-deletion` - Don't show on undeletable organization forks

fregante:
- oughta fix https://github.com/refined-github/refined-github/issues/8652

## Test URLs

1. join org
2. be assigned a role just before "admin" (you should see the Settings tab)
3. find a fork in that org

### DIFF
--- a/source/github-helpers/get-user-permission.ts
+++ b/source/github-helpers/get-user-permission.ts
@@ -1,5 +1,4 @@
 import {CachedFunction} from 'webext-storage-cache';
-import {canUserEditRepo} from 'github-url-detection';
 import {elementExists} from 'select-dom';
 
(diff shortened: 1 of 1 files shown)


source/github-helpers/get-user-permission.ts
```diff
@@ -26,12 +25,6 @@ async function getViewerPermission(): Promise<RepositoryPermission> {
 		throw new Error('This can only be called on a repository page');
 	}
 
-	// Faster DOM-based check, if the DOM is available.
-	// This can be cached because it's the highest level of access
-	if (canUserEditRepo()) {
```
fregante, line 31: This is because the DOM check simply looked for the Settings tab, which does _not_ mean you're an admin

PR #8754, opened 2025-11-09
`comments-time-machine-links` - Fix date parameter disappearing 

SunsetTechuila:
addresses https://github.com/refined-github/refined-github/pull/8712#issuecomment-3484194515

fixes the issue of the date parameter not reappearing after a comment was edited

addresses some review comments

## Test URLs

https://github.com/refined-github/refined-github/issues/8698

### DIFF
--- a/source/features/comments-time-machine-links.tsx
+++ b/source/features/comments-time-machine-links.tsx
@@ -81,7 +81,7 @@ async function showTimeMachineBar(): Promise<void | false> {
 	);
 }
 
-async function addDateParameterToLink(link: HTMLAnchorElement): Promise<void> {
+function addDateParameterToLink(link: HTMLAnchorElement): void {
 	if (!pageDetect.isRepoGitObject(link)) {
 		return;
 	}


source/features/comments-time-machine-links.tsx
```diff
@@ -160,18 +166,14 @@ async function init(signal: AbortSignal): Promise<void> {
...
 
 	delegate(`:is(${commentSelector}) button[data-component="IconButton"]:has(> .octicon-kebab-horizontal)`, 'click', addDropdownLinkReact, {signal});
 
 	observe(
-		`:is(${commentSelector}) relative-time[datetime]`
-		, relativeTime => {
-			const comment = relativeTime.closest(`:is(${commentSelector})`)!;
-			addInlineLinks(comment, relativeTime.attributes.datetime.value);
-		},
+		`:is(${commentSelector}) a[href^="${location.origin}"]:not(.${linkifiedURLClass})`,
```
SunsetTechuila, line 171: this should be safe to replace with `.markdown-body a[href^="${location.origin}"]:not(.${linkifiedURLClass})`

source/features/comments-time-machine-links.tsx
```diff
@@ -138,8 +143,8 @@ async function addDropdownLinkReact({delegateTarget: delegate}: DelegateEvent):
 	$('[class^="prc-ActionList-ItemLabel"]', menuItem).textContent = 'View repo at this time';
 	$('[class^="prc-ActionList-LeadingVisual"]', menuItem).replaceChildren(<HistoryIcon />);
```
SunsetTechuila, line 141: should I use [`createDropdownItem`](https://github.com/refined-github/refined-github/blob/4be765a6af58a43ceff58b675a688cf09ee2bdf1/source/github-helpers/create-dropdown-item.tsx#L12)?

It looks fine when wrapped with an element that provides inline margin

source/features/comments-time-machine-links.tsx
```diff
@@ -80,24 +82,28 @@ async function showTimeMachineBar(): Promise<void | false> {
...
+	}
 
-		saveOriginalHref(link);
+	await Promise.any([oneEvent(link, 'pointerenter'), oneEvent(link, 'focus')]);
 
-		const searchParameters = new URLSearchParams(link.search);
-		searchParameters.set('rgh-link-date', timestamp);
-		link.search = String(searchParameters);
-	}
+	const comment = link.closest(`:is(${commentSelector})`)!;
```
fregante:
```suggestion
	const comment = link.closest(commentSelector)!;
```

source/features/comments-time-machine-links.tsx
```diff
@@ -80,24 +82,28 @@ async function showTimeMachineBar(): Promise<void | false> {
...
-			continue;
-		}
+	// Skip permalinks
+	const linkParts = link.pathname.split('/');
+	if (/^[\da-f]{40}$/.test(linkParts[4])) {
+		return;
+	}
 
-		saveOriginalHref(link);
+	await Promise.any([oneEvent(link, 'pointerenter'), oneEvent(link, 'focus')]);
```
fregante: Why wait?

fregante (approved):
Looks mostly right 🙏 will review again tomorrow

source/features/comments-time-machine-links.tsx
SunsetTechuila: For efficiency - why update the links that user doesn't interact with?

fregante: Makes sense, but by the time this line is reached, we already:

- added a `style` element
- added a global `animationstart` event
- added an `rgh-seen` class to this same element

Adding two listeners instead of just altering the attribute feels too little too late at this point.

In some cases we do prefer using `delegate` globally and not make _any_ changes until the user interacts with something, but we cannot use `delegate` with mouse/pointer events, they fire way too frequently and each event causes a `event.target.closest(observedSelector)` call, which is not necessarily lightweight.

PR #8760, opened 2025-11-10
`no-unnecessary-split-diff-view` - Restore on new PR & commit pages

### LINKED ISSUE #8559: `no-unnecessary-split-diff-view` broken in new PR Files view
### Description

This is so painful.

Image

I managed to change the width by setting this to `0` via CSS (`col:nth-child(2) {width: 0}` but I can't repro anymore 😳 I think I also used width/min-width on `.left-side` or something

Image

https://github.com/refined-github/refined-github/pull/8544/files#diff-78e683a73e20858e09eb458d59f87f8cf6d4f4c172065e375dbd7beef53815e7

### Extension version

25.7.1

### Browser(s) used

Safari

### Token ⚠️

SunsetTechuila:
fixes #8559

## Test URLs

https://github.com/refined-github/sandbox/pull/50/files?diff=split

https://github.com/fregante/sandbox/pull/30/files

https://github.com/refined-github/sandbox/commit/c28cc8e5271452b5b4c347d46a63f717c29417d6?diff=split

## Screenshot

image

### DIFF
--- a/source/features/no-unnecessary-split-diff-view.css
+++ b/source/features/no-unnecessary-split-diff-view.css
@@ -68,6 +68,11 @@ html:not([rgh-OFF-no-unnecessary-split-diff-view]) {
 		--rgh-left-side-w: 100%;
 	}
 
+	/* Align line number cells with the header first cell */
+	.diff-line-number {
+		min-width: 40px;
+	}
+
 	.diff-line-row {
 		/* The left side (deletions) */
 		/* > to avoid selecting suggested changes */
@@ -86,10 +91,5 @@ html:not([rgh-OFF-no-unnecessary-split-diff-view]) {
 			/* Only additions: expand the right side */
 			width: var(--rgh-right-side-w, auto) !important;
 		}
-	}
-
-	/* Align line number cells with the header first cell */
-	.diff-line-number {
-		min-width: 40px;
 	}
 }
--- a/source/features/no-unnecessary-split-diff-view.tsx
+++ b/source/features/no-unnecessary-split-diff-view.tsx
@@ -1,9 +1,9 @@
+import './no-unnecessary-split-diff-view.css';
 import * as pageDetect from 'github-url-detection';
 import {$, $optional} from 'select-dom/strict.js';
 
 import features from '../feature-manager.js';
 import observe from '../helpers/selector-observer.js';
-import './no-unnecessary-split-diff-view.css';
 
 /* TODO: remove in late 2026 */
 void features.addCssFeature(import.meta.url);
@@ -16,7 +16,8 @@ function manageSplitDiffState(tableBody: HTMLTableSectionElement): void {
 		table.classList.remove('rgh-no-split-diff');
 		return;
 	}
-	// avoid selecting suggested deletions/additions
+
+	// Avoid selecting suggested deletions/additions
 	if (!$optional(':scope > tr > td:nth-child(2) > .deletion', tableBody)) {
 		table.classList.add('rgh-no-split-diff', 'rgh-only-additions');
 	} else if (!$optional(':scope > tr > td:nth-child(4) > .addition', tableBody)) {
--- a/source/features/show-names.tsx
+++ b/source/features/show-names.tsx
@@ -62,8 +62,10 @@ async function updateLinks(found: HTMLAnchorElement[]): Promise<void> {
 			// For `sticky-comment-header`. Use attribute because classes are altered by GitHub
 			currentUserElement.closest('[data-testid="comment-header"]')?.setAttribute('data-rgh-viewer-did-author', '');
 		}
+
 		users.delete(currentUser);
 	}
+
 	users.delete('ghost'); // Consider using `github-reserved-names` if more exclusions are needed
 
 	if (users.size === 0) {
--- a/source/features/unread-anywhere.tsx
+++ b/source/features/unread-anywhere.tsx
@@ -52,7 +52,6 @@ async function openUnreadNotifications(event?: React.MouseEvent): Promise<void> 
 	}, {
 		message: 'Loading notifications…',
 		doneMessage: false,
-	// eslint-disable-next-line promise/prefer-await-to-then -- Nah
 	}).finally(() => {
 		if (event?.target instanceof HTMLButtonElement) {
 			event.target.disabled = false;


SunsetTechuila:
the only problem is https://github.com/fregante/sandbox/pull/30/files

I don't know how to fix this and why it is like that

image

upd: same problem here: [refined-github/sandbox#114 (files)](https://github.com/refined-github/sandbox/pull/114/files)

fregante:
The old version also had a similar problem. IMHO it's still an improvement over a 50% of the screen being empty.

SunsetTechuila:
and clicking the number of added/removed row hides the row(???)

SunsetTechuila:
everything seems to be fixed

SunsetTechuila:
> It's very broken in every view for me. This PR looks like this:

could this be a safari thing? 🤔

image

SunsetTechuila:
please try commenting width changes

fregante:
No I tried it in Chrome, I don't test PRs in Safari

> please try commenting width changes

?

Screenshot 10

fregante:
Here's my DOM, with the selector incorrectly targeting and hiding the `td` containing the added code (blue: selected; gray: hovered and tooltipped)

Screenshot 13

SunsetTechuila:
> image

You are not testing the latest changes. I've removed `--rgh-only-additions` & `--rgh-only-deletions` variables from `[class*='DiffLines-module__tableLayoutFixed']:has(col:nth-child(4))` in https://github.com/refined-github/refined-github/pull/8760/commits/434ddf2bf6e166b13e3ff8b4d06b601787df7e8a

> > please try commenting width changes
> ?

commenting out *

fregante:
> You are not testing the latest changes

You only pushed those changes 14 minutes before I posted my review

source/features/no-unnecessary-split-diff-view.css
```diff
@@ -47,4 +47,46 @@ html:not([rgh-OFF-no-unnecessary-split-diff-view]) {
 	:is(.inline-comments, .js-inline-annotations) .empty-cell:not(.blob-num) {
 		display: var(--rgh-only-additions, var(--rgh-only-deletions, table-cell)) !important;
 	}
+
+	/* TODO: remove everything above in late 2026 */
+	[class*='DiffLines-module__tableLayoutFixed']:has(col:nth-child(4)) {
+		/* No changes on the left side */
+		&:not(:has(td:nth-child(2) > .deletion)) {
```
fregante: I'm afraid of these :has() selectors on already-laggy pages. I'd probably be more inclined to use JS to detect these tables instead. What do you think? It should be a relatively simple JS feature that adds. It should be a pretty simple `observe('diff table')` + `if !addition, add class; else if !deletion, add other class`

SunsetTechuila: layout jumps for a second when you switch diff view modes

source/features/no-unnecessary-split-diff-view.css
```diff
@@ -50,46 +50,44 @@ html:not([rgh-OFF-no-unnecessary-split-diff-view]) {
...
+}
 
-	/* TODO: remove everything above in late 2026 */
-	[class*='DiffLines-module__tableLayoutFixed']:has(col:nth-child(4)) {
-		/* No changes on the left side */
-		&:not(:has(td:nth-child(2) > .deletion)) {
-			--rgh-left-side-d: none;
-			--rgh-right-side-w: 100%;
+/* TODO: remove everything above in late 2026 */
+.rgh-no-split-diff {
```
SunsetTechuila, line 56: this could be `table.rgh-diff:has(> tbody.rgh-no-split-diff)` if you are fine with that

SunsetTechuila: https://github.com/SunsetTechuila/refined-github/commit/ba07f681179e386261235d5754b5944211aa3477

fregante: I suggested using JS specifically to avoid :has() this page

source/features/no-unnecessary-split-diff-view.css
fregante: I'm ok with that given the alternative. The PR files natively performs awfully. If anything we should probably review our other features for bad selectors that apply to this page.

source/features/no-unnecessary-split-diff-view.tsx
```diff
@@ -1,9 +1,42 @@
...
+import './no-unnecessary-split-diff-view.css';
 
+/* TODO: remove in late 2026 */
 void features.addCssFeature(import.meta.url);
 
+function manageSplitDiffState(tableBody: HTMLTableSectionElement): void {
+	const table = tableBody.closest('table')!;
+	const columnsGroup = $('colgroup', table);
+	// Diff view is unified
+	if (columnsGroup.childElementCount !== 4) {
```
fregante, line 15: Is there no selector (other than `has`) that can detect this?

source/features/no-unnecessary-split-diff-view.tsx
```diff
@@ -1,9 +1,42 @@
-import './no-unnecessary-split-diff-view.css';
+import * as pageDetect from 'github-url-detection';
+import {$, $optional} from 'select-dom/strict.js';
 
 import features from '../feature-manager.js';
+import observe from '../helpers/selector-observer.js';
+import './no-unnecessary-split-diff-view.css';
```
fregante: CSS imports must be first, why was this moved?

fregante (approved):

fregante (approved):
All the test URLs look good 🙌

source/features/no-unnecessary-split-diff-view.tsx
SunsetTechuila: I really tried to find one

source/features/no-unnecessary-split-diff-view.tsx
```diff
@@ -1,9 +1,43 @@
...
+	// Diff view is unified
+	if (columnsGroup.childElementCount !== 4) {
+		table.classList.remove('rgh-no-split-diff');
+		return;
+	}
+
+	// Avoid selecting suggested deletions/additions
+	if (!$optional(':scope > tr > td:nth-child(2) > .deletion', tableBody)) {
+		table.classList.add('rgh-no-split-diff', 'rgh-only-additions');
+	} else if (!$optional(':scope > tr > td:nth-child(4) > .addition', tableBody)) {
```
SunsetTechuila, line 23: I should have used `elementExists`

fregante: Would be great to write a `no-restricted-syntax` selector to detect `$optional()` or `$()` calls in if statements and include/exclude arrays. Claude is pretty good at this.

SunsetTechuila: > Claude is pretty good at this.

Too bad I recently lost my free Copilot Pro access. Maybe I'll join some org on GitHub by 2026 to get it back

fregante: I use Claude for free, this is 3 lines of outputted code

PR #8771, opened 2025-11-14
`select-notifications` - Fix UI consistency issues

### LINKED ISSUE #8321: `select-notifications` incorrectly leaves "Select all" checkbox ticked
### Description

On the notifications page in the inbox, I used the "Select by: Merged" feature to select the merged items. It selected the merged items on the first page and added a link "Select all 100 notifications" (with 100 being the number of notifications in the inbox). After clicking that it showed "100 selected" at the top, but the check boxes didn't change, making it easy to miss that all items are now selected.

This is bad because it is easy to miss that all items are selected now. It's easy to misunderstand the link to mean "select all merged items on all pages" instead of "select all items on all pages".

#### Screenshots

before clicking the "Select all 100 notifications" link:

Image

after clicking the "Select all 100 notifications" link:

Image

### Extension version

25.2.26

### Browser(s) used

Firefox

### Token ⚠️

SunsetTechuila:
fixes #8321

fixes the "Select all" checkbox not being in an intermediate state when only some notifications are selected

makes the code easier to read and understand

## Test URLs

## Screenshot

image

### DIFF
--- a/source/features/select-notifications.tsx
+++ b/source/features/select-notifications.tsx
@@ -63,7 +63,7 @@ function handleSelection(): void {
 	for (const notification of notifications) {
 		input = $('input.js-notification-bulk-action-check-item', notification);
 		// Updating the "checked" property does not raise any events
-		input.checked = !deselectAll && selectorGroups.every(selectorGroup => elementExists(selectorGroup, notification));
+		input.checked = !deselectAll && selectorGroups.every(selectors => elementExists(selectors, notification));
 	}
 
 	// Trigger the selection action bar update


fregante:
I don't really have an appetite to rewrite this function, "Select all" was used specifically because we had performance issues with other options. The bug seems to be caused by an extra click on GitHub's "select ALL ALL", so I'd be more inclined in fixing the state of the checkbox itself instead (if possible). 

## Current

Screenshot 3

## Correct


Screenshot 4

SunsetTechuila:
> "Select all" was used specifically because we had performance issues with other options.

https://github.com/refined-github/refined-github/blob/9f9dab2e3d09f9c03ed6ca30de083494290a7f2b/source/features/select-notifications.tsx#L66

No GitHub code is run until this

https://github.com/refined-github/refined-github/blob/9f9dab2e3d09f9c03ed6ca30de083494290a7f2b/source/features/select-notifications.tsx#L75

SunsetTechuila:
> No GitHub code is run until this

Including this handler:

image

image

SunsetTechuila:
These lint errors are confusing

Why does it want this:

image

source/features/select-notifications.tsx
```diff
@@ -49,45 +49,27 @@ function getFiltersSelector(formData: FormData, category: Category): string[] {
...
+		const notification = notifications[index];
+		const input = $('input.js-notification-bulk-action-check-item', notification);
+
+		// Updating the "checked" property does not raise any events
+		input.checked = !deselectAll && selectorGroups.every(selectorGroup => elementExists(selectorGroup, notification));
+
+		// Trigger the selection action bar update
+		if (notifications.length - index === 1) {
+			input.dispatchEvent(new Event('change', {bubbles: true}));
 		}
```
fregante: Why not just place `dispatchEvent` outs the loop? Then you can use a regular for-of, no `index`

fregante (approved):
It looks good assuming it works

source/features/select-notifications.tsx
SunsetTechuila: Forgot to change that when replaced `$$optional('.notifications-list-item')` with `$$('.notifications-list-item')`

SunsetTechuila: Or not, maybe I left it like that to not select the input element again

SunsetTechuila:
> assuming it works

no one told me that was mandatory

PR #8777, opened 2025-11-17
`one-click-review-submission` - Restore on new "Files changed" page

### LINKED ISSUE #8711: `one-click-review-submission` doesn't work
### Description

Hi, I've recently noticed that one click review submission features are not working.

Image

And I'd test that one click review submission features are enabled on console log by debug mode

This issue seems to be caused by a recent change in GitHub's pull request review submission UI. 
This is one of the features I use the most, so I would really appreciate it if you could look into it and fix it.

All url with this patterns 
"https://github.com/*/*/pull/*/files"

### Extension version

25.10.2

### Browser(s) used

Chrome 141.0.7390.77

### Token ⚠️

- [ ] I set a token in the options page. I understand that some features require a token to work.

SunsetTechuila:
fixes #8711

this is **painful**

## Test URLs

https://github.com/refined-github/refined-github/pull/8766/files

## Screenshot

image

### DIFF
--- a/source/features/one-click-review-submission.tsx
+++ b/source/features/one-click-review-submission.tsx
@@ -204,13 +204,8 @@ function replaceCheckboxesReact({delegateTarget}: DelegateEvent): void {
 		const shouldDisableAll = $(cancelButtonSelector, actionRow).disabled;
 		// Start with 1 to exclude the "Comment" button
 		for (let index = 1; index < buttons.length; index++) {
-			const button = buttons[index];
 			const radioButton = choices[index][0];
-			if (shouldDisableAll) {
-				button.disabled = true;
-			} else {
-				button.disabled = radioButton.disabled;
-			}
+			buttons[index].disabled = shouldDisableAll || radioButton.disabled;
 		}
 	}
 


SunsetTechuila:
Maybe it makes sense to use the API at this point

busches:
This is still working on GHE, let me know when this is ready to test and I'll confirm it's still good there.

SunsetTechuila:
> This is still working on GHE, let me know when this is ready to test and I'll confirm it's still good there.

It works on https://github.com as well if you disable the new files changed experience, which is in preview

source/features/one-click-review-submission.tsx
```diff
@@ -102,6 +105,129 @@ function replaceCheckboxes(originalSubmitButton: HTMLButtonElement): void {
...
+	cancelButton.parentElement!.replaceChildren(...buttons.toReversed());
+
+	radioGroup.classList.add('d-none');
+	actionRow.classList.add('rgh-action-row');
+	actionRow.after(rghActionRow);
+	// Fix tooltips getting cut off
+	dialog.style.overflow = 'visible';
+
+	// actionRow re-renders each time the submit button state changes
+	new MutationObserver(syncButtonsDisabledState).observe(actionRow, {childList: true});
```
fregante, line 228: This will pile up, a new observer on every click (replaceCheckboxesReact) and is never cleaned up when the feature unloads.

fregante:
> to save _maybe_ one click

When put under this perspective, I think the feature does not deserve to exist, especially with the risk and complexity involved. If it breaks it's an emergency.

source/features/one-click-review-submission.tsx
kovsu: Can we manually intercept the dialog’s show and hide behavior? Maybe we can intercept its close event and then control its visibility using css.

SunsetTechuila: > This will pile up

I thought about that, but browsers (at least Chromium-based ones) seem smart now - I've inspected the heap snapshot with devtools and didn't find extra observers

SunsetTechuila: > Can we manually intercept the dialog’s show and hide behavior?

I don't think this is possible

> Maybe we can intercept its close event

There are no events, it is all [React state](https://react.dev/reference/react/useState) and conditional returns

source/features/one-click-review-submission.tsx
```diff
@@ -102,6 +105,129 @@ function replaceCheckboxes(originalSubmitButton: HTMLButtonElement): void {
...
+function replaceCheckboxesReact({delegateTarget}: DelegateEvent): void {
+	// Exclude the "Submit comments" button
+	if (delegateTarget.textContent !== 'Submit review') {
+		return;
+	}
+
+	const dialog = $('div[role="dialog"]');
+	// Desktop layout first, then mobile layout
+	const reviewBody = $(
+		['[class^="ReviewMenuButton-module__AnchoredReviewBody"]', '[class^="prc-Dialog-Body"]'],
```
FloEdelmann, line 120: Class order might change without changing functionality, so maybe better use `*=` instead of `^=`? (also in other selectors)

SunsetTechuila:
> I think the feature does not deserve to exist, especially with the risk and complexity involved. If it breaks it's an emergency.

I expected that kind of feedback, and I actually agree with it myself. Especially considering that the new experience is still in preview and keeps changing. I will miss this feature, though

> a lot that can go wrong on this crucial GitHub feature.

However, I wouldn't say this feature has high potential to break the UI. It will just be difficult to maintain

Patching the component may be somewhat more reliable, but this is unlikely [to ever happen](https://github.com/refined-github/refined-github/pull/8793#issuecomment-3566970632)

SunsetTechuila:
Just saying: it still works two months later

fregante:
You can restore the branch and reopen this PR

SunsetTechuila:
I'll keep it as a draft until I _thoroughly_ re-test it

SunsetTechuila:
Did you change your mind?

fregante:
Yes my previous point still stands: too much can go wrong in this fundamental part of the website. It works now but eventually it *will* break.

SunsetTechuila:
I wonder what's changed since you said I could reopen this PR

PR #8786, opened 2025-11-20
Add cookies permissions missed in #8740

### LINKED ISSUE #8657: `unread-anywhere` should inherit the container in Firefox
### Description

`unread-anywhere` does not preserve the Firefox container of the current tab. Instead, it opens new tabs with no container. Since Firefox containers segregate cookies, this can lead to the newly opened tabs missing the session cookies for the current account or containing cookies for an entirely different account.

Ideally, the newly opened tabs would inherit the container, as if the user (<kbd>Ctrl</kbd>-)clicked the notification links manually.

### Example URLs

### LINKED ISSUE #8785: `open-all-notifications` broke
### Description

When I click the “open unread” button, it disappears, but doesn't actually open any tabs.

Go to your [notifications](https://github.com/notifications), click “group by repo” and then “open unread” on any repo group header.

### Extension version

25.11.19

### Browser(s) used

Firefox 145.0.1

### Token ⚠️

prplecake:
Really fixes #8657 this time. #8740 missed adding the required "cookies" permissions.

From https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Work_with_contextual_identities#permissions:

> If an API enables cookies to be modified, you need the "cookies" permission. For example, using cookieStoreId in [tabs.query](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/query) does not require the "cookies" API, as reading the property doesn't affect the cookies in the containers. **However, using [tabs.create](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/create) requires the permission because the opened tab can read and modify cookies in a container.**

Fixes #8785 

## Test URLs

### DIFF
--- a/source/manifest.json
+++ b/source/manifest.json
@@ -20,7 +20,8 @@
 		"scripting",
 		"contextMenus",
 		"activeTab",
-		"alarms"
+		"alarms",
+		"cookies"
 	],
 	"host_permissions": [
 		"https://github.com/*",


fregante:
Refer to https://github.com/refined-github/refined-github/issues/8782#issuecomment-3558569772

Does this introduce new text in the permission dialog? If so, this change would make Firefox disable Refined GitHub updates everywhere.

fregante (changes requested):
From what I can tell it doesn't trigger a new permission dialog in neither Chrome nor Firefox, however:

- the permission requires an explanation in the web store, and it's not needed in Chrome, so it would be hard to justify
- we currently don't have any way to customize builds by browser, which impacts DX for local extension development


In the past we customized the build just before deployment, but it has drawbacks.

- if we leave `cookies` in the manifest.json and remove it in the `chrome` deployment: if this fails silently, a bad manifest will be published to the store
- if we add `cookies` to the manifest.json in the `Firefox` deployment: developers won't be able to use `openTabs` locally

Perhaps:

1. add `cookies` to the manifest.json in the `Firefox` deployment
2. _also check if the `cookies` permission exists before adding `cookieStoreId` in `openTabs`_

For the time being I'll soft-revert the previous PR and release a new Refined GitHub version.

fregante:
Continues in https://github.com/refined-github/refined-github/issues/8657

PR #8787, opened 2025-11-20
`resolve-conflicts` - Drop feature

fregante:
- fixes https://github.com/refined-github/refined-github/issues/8783
- closes https://github.com/refined-github/refined-github/issues/8691

This also drops the feature on GHE.

It's relatively easy to add check to see if GitHub is showing the native feature, but it would need testing since the editor is dynamic and the widget might be out of view(/DOM). So probably easier to just drop it.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -56,6 +56,9 @@ GitHub Enterprise is also supported: [How to enable it](https://fregante.github.
 		<th width="50%">
 			<p><a title="show-whitespace"></a> Makes whitespace characters visible
 			<p><img src="https://user-images.githubusercontent.com/1402241/61187598-f9118380-a6a5-11e9-985a-990a7f798805.png">
+		<th width="50%">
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -56,9 +56,6 @@ GitHub Enterprise is also supported: [How to enable it](https://fregante.github.
 		<th width="50%">
 			<p><a title="show-whitespace"></a> Makes whitespace characters visible
 			<p><img src="https://user-images.githubusercontent.com/1402241/61187598-f9118380-a6a5-11e9-985a-990a7f798805.png">
-		<th width="50%">
-			<p><a title="resolve-conflicts"></a> Adds one-click merge conflict fixers
-			<p><img src="https://user-images.githubusercontent.com/1402241/54978791-45906080-4fdc-11e9-8fe1-45374f8ff636.png">
```
fregante, line 61: TODO: place another feature here in the highlights

PR #8790, opened 2025-11-21
`status-subscription` - Restore feature

SunsetTechuila:
I don't really expect this to be merged

Fixes #7968

## Test URLs

https://github.com/refined-github/refined-github/issues/8785

## Screenshot

image

PR #8793, opened 2025-11-22
add `getReactProps` helper and use it for `quick-comment-edit`

SunsetTechuila:
Please let me know right away if you don't find this approach feasible, so I won't waste my time finalizing it.

This doesn’t work yet because React props are only available in the main world. We have to either move the content script to the main world, or `getReactProps` and call hooks via RPC. Some adblockers access React props too, but I haven’t looked into how they do it yet.

Pull requests that can benefit from this helper: #8777, #8790


SunsetTechuila:
> move the content script to the main world

This opens a path to [reusing](https://github.com/refined-github/refined-github/issues/7856#issuecomment-3497843499) and [patching](https://github.com/refined-github/refined-github/issues/6554#issuecomment-3054178103) GitHub's webpack modules, but it does require some architectural changes

> or `getReactProps` and call hooks via RPC

This is easier to implement but offers fewer possibilities

> if you don't find this approach feasible

However, I don't think we can completely avoid it. As the React rewrite continues, we'll have to drop more and more features without it, and the remaining ones will become increasingly hard to maintain unless they're very basic

fregante:
> unless they're very basic

I'm ok with keeping only small refinements around. New features have been easy to write, but maintenance is a different story. If features no longer fit the bill, then they should go. As always, _Refined GitHub doesn't have to do everything,_ features can live in additional extensions.

> move the content script to the main world

Not going to happen, this would just open a bag of worms when most features wouldn't benefit from this. Unfortunately I've never published a helper to simplify safe/unsafe world communication. https://github.com/fregante/GhostText/blob/63daa079b651babd532af44df21cca5c6b831f9d/source/advanced-editors-messenger.js#L10-L12 has an example though.

My suggestion would be to implement a simple "main-world.js" script that adds a `get` event listener to the `window` object.

When the content script fires a bubbling `get` event on the react element, it "returns" its props. This would enable access to the data, as well as access to its `edit` function which can then be attached as a click listener.

You can see how GhostText "returns" the object, there's no serialization involved:

https://github.com/fregante/GhostText/blob/63daa079b651babd532af44df21cca5c6b831f9d/source/advanced-editors-messenger.js#L26-L28

---

The approach is interesting, especially if it allows to restore important features like the 2 you mentioned. If accepted, this would need some documentation and some debugging tools or else no one can use this new helper.

As for the class, I'm not a fan of it. I expect `props` to be an object, right, so why not pre-parse it and let me use it as `props.whatever.edit()`? The `findBy` methods seem unnecessary.

SunsetTechuila:
> The `findBy` methods seem unnecessary.

These methods are mainly for stability - I think it's reasonable not to depend on exact prop names

> so why not pre-parse it and let me use it as `props.whatever.edit()`?

You can do that. In this case we're looking for two hooks: `startIssueBodyEdit` and `editComment`, so it would be `editHook === props.startIssueBodyEdit ?? props.editComment`

SunsetTechuila:
> My suggestion would be to implement a simple "main-world.js" script that adds a get event listener to the window object.

Would a message manager like GitHub's `socket:message` thing be overengineering?

SunsetTechuila:
You can't transfer anything other than primitives via event.detail

image

image

SunsetTechuila:
Yes, this was predictable since objects are passed by reference, which would not be valid in another isolated memory context. I thought maybe they would be copied from one to another in this case, but of course, they are not

fregante:
> You can't transfer anything other than primitives via event.detail

Indeed, how about a second `call` event that receives the [path](https://github.com/sindresorhus/dot-prop) to the function and calls it?

```js
window.dispatchEvent(new CustomEvent('rgh:call', {detail: 'props.whatever.edit'}))
```

> I think it's reasonable not to depend on exact prop names

We always depend on the exact prop names, see classnames, data-testid, general DOM structure, etc, etc.

> You can do that. In this case we're looking for two hooks: `startIssueBodyEdit` and `editComment`, so it would be `editHook === props.startIssueBodyEdit ?? props.editComment`

Exact method calls seem very reasonable to me. What if they add `editCommentTasks` and `/edit/` matches that? I'd rather have a feature that doesn't work at all than one that does something else entirely.

SunsetTechuila:
> You can't transfer anything other than primitives via event.detail

Not really. You can't transfer functions and class instances, but you can transfer arrays and objects as long as they don't contain them. Very interesting

SunsetTechuila:
I wish I had known that before writing a serialization-deserialization layer

SunsetTechuila:
Waiting for general feedback before continuing

fregante:
This is growing into something that just isn't Refined GitHub anymore. Again if something requires this much complexity, we just don't do it. Closing for now, sorry. I don't want you to spend time on this and I can't review it.

SunsetTechuila:
https://github.com/aidenybai/bippy

PR #8805, opened 2025-11-29
Fix button color contrast when deleting tags

u01jmg3:
Improve button color contrast for unused tag deletion (https://github.com/refined-github/refined-github/issues/8803)

> [!NOTE]  
>
> - `!important` was required as `.color-fg-danger` in GitHub's CSS was already using `!important`.
> - This may break when GitHub switches to BEM as seen on other pages, but I couldn't find an example of BEM being used on a dropdown component.

## Test URLs

https://github.com/refined-github/refined-github/tags

## Screenshot

Before:

After:

### DIFF
--- a/source/features/github-bugs.css
+++ b/source/features/github-bugs.css
@@ -115,10 +115,10 @@ svg.octicon-package + a.Truncate[href*='/releases/download/'] > .Truncate-text {
 
 /* Fix button color contrast when deleting an unused tag */
 /* Info: https://github.com/refined-github/refined-github/issues/8803 */
-/* Test: https://github.com/refined-github/refined-github/tags */
+/* Test: https://github.com/refined-github/sandbox/tags */
 .dropdown-item.btn-link.color-fg-danger:hover {
-	color: var(--fgColor-onEmphasis) !important;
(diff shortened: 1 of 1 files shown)


source/features/github-bugs.css
```diff
@@ -113,6 +113,14 @@ svg.octicon-package + a.Truncate[href*='/releases/download/'] > .Truncate-text {
 	white-space: nowrap;
 }
 
+/* Fix button color contrast when deleting an unused tag */
+/* Info: https://github.com/refined-github/refined-github/issues/8803 */
+/* Test: https://github.com/refined-github/refined-github/tags */
+.dropdown-item.btn-link.color-fg-danger:hover {
+	background-color: var(--button-danger-bgColor-hover);
+	color: var(--button-danger-fgColor-hover) !important;
```
SunsetTechuila:
```suggestion
	color: var(--fgColor-onEmphasis) !important;
	background-color: var(--bgColor-danger-emphasis);
```

Image

fregante (approved):

fregante:
Thank you!

PR #8806, opened 2025-11-29
`conversation-activity-filter` - Restore on issues & sticky headers

### LINKED ISSUE #8586: `conversation-activity-filter` icon/filter missing in issues
### Description

- [x] The bug is caused by Refined GitHub. It doesn't happen if I disable the extension.
- [x] The bug happens after clearing extension cache. The "Clear cache" button can be found in the options page.

 
Thanks for the great extension

1. Visit this url (or any other issue) https://github.com/refined-github/refined-github/issues/7856
2. Look at the status bar for the issue

### Extension version

25.7.16

### Browser(s) used

Arc Version 1.95.0 (62781) Chromium Engine Version 136.0.7103.114

### Token ⚠️

SunsetTechuila:
fixes #8586

## Test URLs

#8586

#8806

## Screenshot

image

image

image

### DIFF
--- a/source/features/conversation-activity-filter.css
+++ b/source/features/conversation-activity-filter.css
@@ -27,6 +27,7 @@
 	display: inline !important;
 }
 
+/* Fix dropdown getting cut off on issue headers */
 [class^='HeaderMetadata-module__titleAndMetadata'] {
 	overflow: visible !important;
 }


source/features/conversation-activity-filter.tsx
SunsetTechuila: `anchored-position` element uses [Popover API](https://developer.mozilla.org/docs/Web/API/Popover_API), so it is not compatible with `details`

SunsetTechuila: should I rewrite the details menu as an action menu? @fregante

fregante: Feels like this is unrelated to the bugfix, so it's best done as a followup, if at all. I like that this PR is rather minimal so once ready it will be easy to review and merge.

SunsetTechuila:
I don't know how to achieve the same for the issue header 

image

SunsetTechuila:
Affected by #6554

source/features/conversation-activity-filter.css
```diff
@@ -25,33 +27,35 @@
...
+	.rgh-conversation-activity-filter-wrapper {
+		display: block;
+		margin-bottom: 8px; /* Preserve centered vertical alignment with status badge */
+	}
+
+	.rgh-conversation-activity-filter {
+		display: inline;
+	}
 
-/* Our `position:relative` overlaps this element https://github.com/refined-github/refined-github/issues/6663 */
```
SunsetTechuila, line 54: Image

source/features/conversation-activity-filter.tsx
```diff
@@ -249,14 +258,17 @@ async function init(signal: AbortSignal): Promise<void> {
...
 
 	registerHotkey('h', switchToNextFilter, {signal});
 }
 
 void features.add(import.meta.url, {
 	include: [
 		pageDetect.isConversation,
+		// Workaround for #6554
+		// TODO: remove once the issue is resolved
+		pageDetect.isIssueOrPRList,
```
fregante, line 271: How does this work around the fact that the page isn't updated? It purposefully runs the feature on the list (where nothing happens) waiting for an issue to be opened?

fregante (approved):
YOLO

source/features/conversation-activity-filter.tsx
SunsetTechuila: > It purposefully runs the feature on the list (where nothing happens) waiting for an issue to be opened?

Yes

PR #8807, opened 2025-11-30
 `releases-dropdown` - Remove mention of tags

Michael21107:
As shown in the screenshot below, the tags page has no search bar, so the description of `releases-dropdown` was corrected to reflect that.

image


kovsu:
`tags-dropdown` has been dropped since https://github.com/refined-github/refined-github/pull/6438

PR #8809, opened 2025-12-01
`sticky-comment-header` - Fix viewer's comment headers color on GHE

### LINKED ISSUE #8801: fuchsia background with `sticky-comment-header` on GitHub Enterprise Server v3.12.17
### Description

There's a fuchsia background on all comments/PR and Issue description headers authored by the user on GitHub Enterprise Server v3.12.17, I think this means that the referenced CSS variable isn't available on GHES?

Image

Not sure how to replicate as this is on a private GitHub Enterprise Server setup

### Extension version

25.11.22

### Browser(s) used

Arc / Chrome

### Token ⚠️

SunsetTechuila:
fixes #8801

### DIFF
--- a/source/features/sticky-comment-header.css
+++ b/source/features/sticky-comment-header.css
@@ -22,6 +22,8 @@ html:not([rgh-OFF-sticky-comment-header]) {
 			var(--base-sticky-header-height, 0px) +
 			var(--rgh-issue-or-pr-sticky-header-height, 0px)
 		);
+		/* Fallback in case `show-names` is disabled - #8726#issuecomment-3606420543 */
+		background: var(--bgColor-muted, var(--color-canvas-subtle, fuchsia));
 	}
 
 	:is([id^='gistcomment'], [id^='commitcomment']) > * {
--- a/source/features/sticky-comment-header.css
+++ b/source/features/sticky-comment-header.css
@@ -1,4 +1,9 @@
 html:not([rgh-OFF-sticky-comment-header]) {
+	--rgh-bg-accent: var(
+		--bgColor-accent-muted,
+		var(--color-accent-subtle, fuchsia)
+	);
+
 	/* biome-ignore format: bug */
 	/* Issue body header */
 	[class^='IssueBodyHeader-module__IssueBodyHeaderContainer'],
@@ -29,10 +34,8 @@ html:not([rgh-OFF-sticky-comment-header]) {
 	[class*='IssueBodyHeader-module__viewerDidAuthor'] {
 		/* Draw a possibly semi-transparent accent color over the opaque background */
 		background:
-			linear-gradient(
-				var(--bgColor-accent-muted, fuchsia),
-				var(--bgColor-accent-muted, fuchsia)
-			), var(--rgh-background);
+			linear-gradient(var(--rgh-bg-accent), var(--rgh-bg-accent)),
+			var(--rgh-background);
 	}
 
 	/* Prevent header of the next comment from overlapping context menu of the sticky header */
--- a/source/features/sticky-comment-header.css
+++ b/source/features/sticky-comment-header.css
@@ -22,8 +22,6 @@ html:not([rgh-OFF-sticky-comment-header]) {
 			var(--base-sticky-header-height, 0px) +
 			var(--rgh-issue-or-pr-sticky-header-height, 0px)
 		);
-		/* Fallback in case `show-names` is disabled - #8726#issuecomment-3606420543 */
-		background: var(--bgColor-muted, var(--color-canvas-subtle, fuchsia));
 	}
 
 	:is([id^='gistcomment'], [id^='commitcomment']) > * {


source/features/sticky-comment-header.css
```diff
@@ -17,6 +22,8 @@ html:not([rgh-OFF-sticky-comment-header]) {
 			var(--base-sticky-header-height, 0px) +
 			var(--rgh-issue-or-pr-sticky-header-height, 0px)
 		);
+		/* Fallback in case `show-names` is disabled - #8726#issuecomment-3606420543 */
+		background: var(--bgColor-muted, var(--color-canvas-subtle, fuchsia));
```
SunsetTechuila: I'm not sure If I like this. I'll let @fregante decide

cc @karlhorky

https://github.com/refined-github/refined-github/pull/8726#issuecomment-3606710723

SunsetTechuila: You can't be more explicit than that

image

SunsetTechuila: That's I don't like for sure

msedge_O7rDNAJFSg

source/features/sticky-comment-header.css
```diff
@@ -17,6 +22,8 @@ html:not([rgh-OFF-sticky-comment-header]) {
 			var(--base-sticky-header-height, 0px) +
 			var(--rgh-issue-or-pr-sticky-header-height, 0px)
 		);
+		/* Fallback in case `show-names` is disabled - #8726#issuecomment-3606420543 */
```
fregante:
```suggestion
		/* Fallback in case `show-names` is disabled */
		/* https://github.com/refined-github/refined-github/pull/8726#issuecomment-3606420543 */
```

Due to https://github.com/refined-github/refined-github/issues/6336 I reduced my reliance on references in code

fregante:
> Screenshot

?

I assume this makes the comment headers just use the standard (non-blue) header color, right?

fregante (approved):

source/features/sticky-comment-header.css
SunsetTechuila: I'm reverting that. Due to #6554, the background of comment headers that should be blue is gray when navigating from issue list to issue page. The header also turns gray after a comment is edited

> I assume this makes the comment headers just use the standard (non-blue) header color, right?

Yes, that's how it worked

SunsetTechuila: Can `html:not([rgh-OFF-show-names])` be used here?

fregante: I don’t think that attribute exists? But I suppose manually adding a class in that feature could also work. 👌

karlhorky:
@fregante @SunsetTechuila thanks for the PR and merge!

karlhorky:
@SunsetTechuila Why was the fallback fixing [the transparent header problem](https://github.com/refined-github/refined-github/pull/8726#issuecomment-3606420543) reverted?

- https://github.com/refined-github/refined-github/pull/8809/changes/917f53157161f4aa5be797b9ac4b1c34d63f12c3

SunsetTechuila:
@karlhorky https://github.com/refined-github/refined-github/pull/8809#discussion_r2585065742 https://github.com/refined-github/refined-github/pull/8809#discussion_r2585076341 https://github.com/refined-github/refined-github/pull/8809#discussion_r2618714434

karlhorky:
> That's I don't like for sure

@SunsetTechuila what do you mean? That it starts out as black and then changes to blue on hydration? Maybe the fallback can be applied differently, to avoid the FOUC? Eg. via JS or something.

karlhorky:
> > Can `html:not([rgh-OFF-show-names])` be used here?
>
> But I suppose manually adding a class in that feature could also work. 👌

Or maybe adding a class, as fregante mentioned?

SunsetTechuila:
> Or maybe adding a class, as fregante mentioned?

That's what I'm planning to do: add this class and use it to disable `sticky-comment-header` if `show-names` is disabled

karlhorky:
Maybe we could do it in such a way that the feature doesn't get disabled, but instead degraded? (graceful degradation). Black/grey background is fine in that case.

Then the notice about the `show-names` could also probably be removed.

SunsetTechuila:
> Maybe we could do it in such a way that the feature doesn't get disabled, but instead degraded? (graceful degradation).

I couldn't. Maybe you can

fregante:
Ok, how about this? `rgh-own-comments` feature that simply adds `.rgh-own-comment` to your own comments. It would be a silent, non-disableable feature that can be used by others. This way we drop the inter-feature dependency (`show-names`) and can be used by more features in the future if helpful.

karlhorky:
@SunsetTechuila what do you think of this approach with the non-disableable `rgh-own-comments` feature? Do you think that you could do this? Or should I give it a shot?

SunsetTechuila:
Maybe there are already several helper features like this in RGH, but if not, then I'd rather stop trying to keep `sticky-comment-header` CSS-only. Keeping it CSS-only has caused way more hassle than ~10 lines of simple JS code ever could

karlhorky:
@fregante is it a requirement to keep it CSS-only? Or would a short JS feature like what @SunsetTechuila is proposing also be acceptable?

fregante:
I already gave my suggestion

karlhorky:
Ok, so [the non-disableable `rgh-own-comments` feature](https://github.com/refined-github/refined-github/pull/8809#issuecomment-3707766715) is the path forward then.

@SunsetTechuila if you think you would do this, then great! Otherwise, I can take a shot at it.

fregante:
For future reference: Note how much this "small feature" had to be discussed and tweaked and bugfixed for months. This should help understand why I'm hesitant to accept certain features and why I might drop this one.

karlhorky:
I empathize with the back and forth of the discussion sapping of focus away from other work - it's unfortunate that GitHub's DOM, tech stack and framework integration makes for a less stable foundation upon which to build a browser extension (on the other hand, luckily, there has been no shifting in the relevant parts of that foundation since this feature was first implemented).

However, aside from the back and forth in discussions and the technical requirements, I'm not sure I agree that the amount of total implementation work that has gone into this feature is overly blown-up (also as I mentioned in [my comment in October](https://github.com/refined-github/refined-github/issues/8463#issuecomment-3448948880) - most of the work was done before then): the first versions and fixes have been developed pretty quickly by @SunsetTechuila (also taking into account browser compatibility).

Maybe with looser requirements like not having the CSS-only rule and not having the dependency on other features it could have gone a bit faster - but probably also not worlds apart.

SunsetTechuila:
> Ok, how about this? `rgh-own-comments` feature that simply adds `.rgh-own-comment` to your own comments. It would be a silent, non-disableable feature that can be used by others.

> Maybe there are already several helper features like this in RGH, but if not, then I'd rather stop trying to keep [`sticky-comment-header`](https://github.com/refined-github/refined-github/blob/main/source/features/sticky-comment-header.tsx?rgh-link-date=2026-01-05T12%3A50%3A05Z) CSS-only.

> @SunsetTechuila if you think you would do this, then great! Otherwise, I can take a shot at it.

@karlhorky I didn't look thoroughly, but I couldn't find any other helper features like that. So if you want to implement it, you'll have to do it yourself.

When you do, please also change `[class^='ReviewThreadComment-module']` to `[class*='ReviewThreadComment']` - GitHub added a similar component with a different name, `InlineReviewThread`. Test url: https://github.com/refined-github/sandbox/commit/7becc26eccef5da79f64003356067e089043d8fb#r174065194

<details>
<summary>Off-topic</summary>

> For future reference: Note how much this "small feature" had to be discussed and tweaked and bugfixed for months. This should help understand why I'm hesitant to accept certain features and why I might drop this one.

The lesson I've learned is that it depends on how broad the feature's effect is. If it affects something very common that appears in different forms (like comments) - expect problems.

</details>

SunsetTechuila:
> Test url: [refined-github/sandbox@7becc26#r174065194](https://github.com/refined-github/sandbox/commit/7becc26eccef5da79f64003356067e089043d8fb#r174065194)

I can't see my comments when logged in 🤔. I thought they had been hidden or deleted, but they haven't. I tried disabling all extensions, so it seems GitHub engineers are doing a great job with the rewrite

karlhorky:
> @karlhorky I didn't look thoroughly, but I couldn't find any other helper features like that. So if you want to implement it, you'll have to do it yourself.

Ok, I'll take a look

karlhorky:
> > @karlhorky I didn't look thoroughly, but I couldn't find any other helper features like that. So if you want to implement it, you'll have to do it yourself.
> 
> Ok, I'll take a look

I just saw that the decoupling from `show-names` was done by @fregante in PR https://github.com/refined-github/refined-github/pull/9905, and released in [`refined-github@26.8.8`](https://github.com/refined-github/refined-github/releases/tag/26.8.8)

I can confirm that it's working:

Screenshot 2026-08-16 at 16 45 05

Thanks!

PR #8830, opened 2025-12-22
`previous-next-commit-buttons` - Restore on new PR commit pages

SunsetTechuila:
related: https://github.com/refined-github/github-url-detection/pull/212

## Test URLs

`https://github.com/refined-github/refined-github/pull/4448/changes/0b8966c918eae11da9fc992368741757088edf08`

## Screenshot

image

### DIFF
--- a/source/features/previous-next-commit-buttons.tsx
+++ b/source/features/previous-next-commit-buttons.tsx
@@ -6,7 +6,7 @@ import features from '../feature-manager.js';
 
 function init(): false | void {
 	const originalPreviousNext = $optional('.commit .float-right.ButtonGroup') // Legacy
-		?? $optional('a[aria-label="Previous commit"]')?.closest('[class^="prc-ButtonGroup-ButtonGroup"]');
+		?? $optional('[class^="prc-ButtonGroup-ButtonGroup"]:has(a[aria-label="Previous commit"])');
 	if (!originalPreviousNext) {
 		return false;
 	}


SunsetTechuila:
I think we should move those buttons to the sticky header. It would be much more convenient this way

image

fregante:
> I think we should move those buttons to the sticky header. It would be much more convenient this way


That bar is too busy already, I don't want to add more items. Open the sidebar, fix the "quick diff buttons" feature, have a moderately-sized monitor and it will overflow. This is not something so important that needs this extra trouble.

source/features/previous-next-commit-buttons.tsx
```diff
@@ -5,13 +5,17 @@ import * as pageDetect from 'github-url-detection';
 import features from '../feature-manager.js';
 
 function init(): false | void {
-	const originalPreviousNext = $optional('.commit .float-right.ButtonGroup');
+	const originalPreviousNext = $optional('.commit .float-right.ButtonGroup') // Legacy
+		?? $optional('a[aria-label="Previous commit"]')?.closest('[class^="prc-ButtonGroup-ButtonGroup"]');
```
fregante:
```suggestion
		?? $optional('[class^="prc-ButtonGroup-ButtonGroup"]:has(a[aria-label="Previous commit"])');
```

fregante (approved):

PR #8832, opened 2025-12-22
Automatically assign labels to PRs

### LINKED ISSUE #8727: Make the `Labeler` workflow automatically assign a label
This workflow creates a lot of noise, requiring routine work. By checking labels of the issues linked to the PR and its name, we can reliably determine the correct label for the PR in most cases.

SunsetTechuila:
resolves #8727

## Test URLs

[refined-github/sandbox#50](https://www.youtube.com/watch?v=WVSaFHk6aVk)

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -12,6 +12,28 @@ on:
       - edited
 
 jobs:
+  Reminder:
+    permissions:
+      pull-requests: write
+    needs:
+      - Title
+      - Labels
+    if: ${{ always() && github.event_name == 'pull_request_target' }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Post comment if unlabeled
+        if: ${{ needs.Labels.outputs.added == 'false' }}
+        uses: marocchino/sticky-pull-request-comment@v2
+        with:
+          header: reminder
+          message: 'To maintainers: Please add labels to this PR'
+      - name: Remove comment if labeled
+        if: ${{ needs.Labels.outputs.added == 'true' || github.event.action == 'labeled' }}
+        uses: marocchino/sticky-pull-request-comment@v2
+        with:
+          header: reminder
+          delete: true
+
   Title:
     permissions:
       pull-requests: write
@@ -80,25 +102,3 @@ jobs:
               labels,
             });
             return true;
-
-  Reminder:
-    permissions:
-      pull-requests: write
-    needs:
-      - Title
-      - Labels
-    if: ${{ always() && github.event_name == 'pull_request_target' }}
-    runs-on: ubuntu-latest
-    steps:
-      - name: Post comment if unlabeled
-        if: ${{ needs.Labels.outputs.added == 'false' }}
-        uses: marocchino/sticky-pull-request-comment@v2
-        with:
-          header: reminder
-          message: 'To maintainers: Please add labels to this PR'
-      - name: Remove comment if labeled
-        if: ${{ needs.Labels.outputs.added == 'true' || github.event.action == 'labeled' }}
-        uses: marocchino/sticky-pull-request-comment@v2
-        with:
-          header: reminder
-          delete: true


.github/workflows/labeler2.yml
```diff
@@ -0,0 +1,113 @@
...
+            const result = await github.graphql(query, { owner, repo, prNumber });
+            const { pullRequest } = result.repository;
+            const linkedIssues = pullRequest.closingIssuesReferences?.nodes;
+            const labelNames = new Set(
+              linkedIssues?.flatMap(
+                linkedIssue => linkedIssue.labels?.nodes.map(label => label.name) ?? []
+              )
+            );
+            isEnhancement = isEnhancement || labelNames.has('enhancement');
+            isBugFix = isBugFix || labelNames.has('bug');
```
SunsetTechuila: I can change this to 

```js
const linkedIssues = pullRequest.closingIssuesReferences?.nodes ?? [];
for (const linkedIssue of linkedIssues) {
  const labelNames = linkedIssue.labels?.nodes.map(label => label.name) ?? [];
  isEnhancement = isEnhancement || labelNames.some(labelName => labelName === 'enhancement');
  isBugFix = isBugFix || labelNames.some(labelName => labelName === 'bug');
}
```

if you want

.github/workflows/conversations.yml
```diff
@@ -0,0 +1,25 @@
...
+    permissions:
+      pull-requests: write
+      issues: write
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: fregante/keyword-formatter-action@v1
+        with:
+          token: ${{ github.token }}
+          keywords-path: source/features
```
SunsetTechuila: I hope this is well-secured

SunsetTechuila: My own code is safe

.github/workflows/labeler.yml
```diff
@@ -2,48 +2,128 @@ name: Labeler
...
+            const result = await github.graphql(query, { owner, repo, prNumber });
+            const { pullRequest } = result.repository;
+            const linkedIssues = pullRequest.closingIssuesReferences?.nodes;
+            const labelNames = new Set(
+              linkedIssues?.flatMap(
+                linkedIssue => linkedIssue.labels?.nodes.map(label => label.name) ?? []
+              )
+            );
+            isEnhancement = isEnhancement || labelNames.has('enhancement');
+            isBugFix = isBugFix || labelNames.has('bug');
```
SunsetTechuila: should I try to copy the labels instead?

.github/workflows/labeler.yml
```diff
@@ -2,48 +2,128 @@ name: Labeler
...
+        uses: actions/github-script@v8
         with:
-          message: Please add labels to this PR
-  Title:
-    if: ${{ github.event_name != 'pull_request_review' && github.event.action != 'closed' }}
+          script: |
+            const prTitle = context.payload.pull_request.title;
+            let isEnhancement = /^`[\w`-]+` - New feature/.test(prTitle) || /^feat:/.test(prTitle);
+            let isBugFix = /^`[\w`-]+` - Fix/.test(prTitle) || /^fix:/.test(prTitle);
+            if (isEnhancement && isBugFix) return ['enhancement', 'bug'];
```
SunsetTechuila: `title-to-labels-action` could be used here if it outputted labels instead of assigning them

SunsetTechuila: I can open a PR if you want

fregante: We already have that action, this means this logic is redundant.

1. use title to label action on PR open
2. add labels when the PR is opened or edited
3. add comment on review just like it used to


These three points don't have to be sequential. By the time 3 fires, then 1 and 2 have already run.

.github/workflows/labeler.yml
SunsetTechuila: No

.github/workflows/labeler.yml
```diff
@@ -2,48 +2,137 @@ name: Labeler
...
-      issues: write
+    needs: check_labels
+    if: ${{ needs.check_labels.outputs.has_labels == 'false' }}
+    runs-on: ubuntu-latest
     steps:
       - uses: fregante/title-to-labels-action@v1
         with:
           token: ${{ github.token }}
-      - uses: actions/checkout@v4
-      - uses: fregante/keyword-formatter-action@v1
```
fregante, line 46: This should be part of the same workflow, it does not have to depend on any `output`, `title-to-labels-action` doesn't care if it already has labels

.github/workflows/labeler.yml
```diff
@@ -2,48 +2,137 @@ name: Labeler
...
-      - name: Post comment if unlabeled
-        if: steps.check_labels.outputs.result == 'true'
-        uses: marocchino/sticky-pull-request-comment@v2
-        with:
-          message: Please add labels to this PR
-  Title:
-    if: ${{ github.event_name != 'pull_request_review' && github.event.action != 'closed' }}
-    runs-on: ubuntu-latest
+  title_to_labels:
+    name: Map PR title to labels
```
fregante: Keep names short. The previous `Labeler | Title` was fine for example

.github/workflows/labeler.yml
```diff
@@ -2,48 +2,137 @@ name: Labeler
...
+              labels: prLabels,
+            });
+            return true;
+
+  add_reminder:
+    name: Remind to add labels
+    permissions:
+      pull-requests: write
+    needs: assign_labels
+    if: ${{ needs.assign_labels.outputs.assigned == 'false' }}
```
fregante: As I said, this does not need to be related. The 3 steps happen in 3 separate events. Please keep it simple

> 1. use title to label action on PR open
> 2. add labels when the PR is opened or edited
> 3. add comment on review just like it used to

Also actionlint is failing

SunsetTechuila: Actionlint is failing because of the other workflow. I don't feel comfortable "fixing" it, since the only way to do so would be to remove code that was intentionally left in place, just disabled. So, it's up to you

.github/workflows/labeler.yml
SunsetTechuila: I've moved it from the Labeler workflow because:

1. It doesn't add labels
2. It runs on issues as well, forcing all other jobs to check the trigger event

SunsetTechuila: > 2. It runs on issues as well, forcing all other jobs to check the trigger event

Posting a comment only after the review is submitted or the PR is merged (giving a chance to do that without labelling it) also forces other jobs to check the trigger event anyway, so whatever

.github/workflows/labeler.yml
```diff
@@ -2,48 +2,99 @@ name: Labeler
...
+            const linkedIssueLabels = linkedIssues?.flatMap(
+              linkedIssue => linkedIssue.labels?.nodes.map(label => label.name) ?? []
+            );
+
+            const labels = LABELS_TO_INHERIT.filter(label => linkedIssueLabels.includes(label));
+            await github.rest.issues.addLabels({
+              owner,
+              repo,
+              issue_number: prNumber,
+              labels,
```
fregante: I think this will fail if the intersection is empty (i.e. no labels to add). A condition might be necessary

.github/workflows/labeler.yml
```diff
@@ -2,48 +2,101 @@ name: Labeler
...
+                owner,
+                repo,
+                issue_number: prNumber,
+                labels,
+              });
+            }
+
+  Reminder:
+    permissions:
+      pull-requests: write
```
SunsetTechuila: This doesn't work for `pull_request_review` because the workflow runs in the context of the forked repository, meaning the token only grants write access to pull requests in the fork, not the base repo

.github/workflows/labeler.yml
SunsetTechuila: > > 3. add comment on review just like it used to

Not possible, see https://github.com/refined-github/refined-github/pull/8832#discussion_r2658332924

SunsetTechuila:
No, 84c971321759d1c884b835e93bec794fcc49a7a1 is not enough, https://github.com/refined-github/refined-github/pull/8832/changes/aabbc0c6cc5a0cdf2fd06dee5faede95936ecb38 and https://github.com/refined-github/refined-github/pull/8832/changes/a99a0bd2671ddbc7138f7a18940814caf7801582 have to be reverted

.github/workflows/labeler.yml
```diff
@@ -2,48 +2,106 @@ name: Labeler
...
+          script: |
+            const { labels } = context.payload.pull_request;
+            return labels.length > 0;
+      - name: Post comment if unlabeled
+        if: ${{ steps.has_labels.outputs.result == 'false' }}
+        uses: marocchino/sticky-pull-request-comment@v2
+        with:
+          header: reminder
+          message: 'To maintainers: Please add labels to this PR'
+      - name: Remove comment if labeled
```
SunsetTechuila: The comment isn't removed if labels were added by previous jobs. This job can't rely on the GitHub API, it should use outputs from the previous jobs instead

.github/workflows/labeler.yml
```diff
@@ -2,48 +2,103 @@ name: Labeler
...
-      github.event_name == 'pull_request_review' ||
-      (github.event_name == 'pull_request_target' && github.event.pull_request.merged == true)
+  Title:
+    permissions:
+      pull-requests: write
+      issues: write
+    if: ${{ github.event.action == 'opened' }}
     runs-on: ubuntu-latest
+    steps:
+      - uses: fregante/title-to-labels-action@v1
```
SunsetTechuila, line 44: This action should output whether it added any labels or not

.github/workflows/labeler.yml
```diff
@@ -2,48 +2,103 @@ name: Labeler
...
-    if: >
-      github.event_name == 'pull_request_review' ||
-      (github.event_name == 'pull_request_target' && github.event.pull_request.merged == true)
-    runs-on: ubuntu-latest
     permissions:
-      contents: read
       pull-requests: write
+    needs:
+      - Title
+      - Labels
```
fregante, line 20: I don't want to repeat things 3 times for something that saves me 2 clicks per PR

fregante (changes requested):

PR #8835, opened 2025-12-25
`quick-mention` - Fix comment header overlap

btea:
## Test URLs
You can view the comments corresponding to the PR to reproduce the issue. https://github.com/refined-github/refined-github/pull/8771

## Screenshot
image

As shown in the image, when the approved comment card appears, the tooltip content is obscured when the mouse hovers over the quick mention icon.

### DIFF
--- a/source/features/quick-mention.css
+++ b/source/features/quick-mention.css
@@ -29,6 +29,7 @@
 	color: inherit;
 }
 
+/* Show tooltip on top of `sticky-comment-header` */
 .TimelineItem-avatar:has(.rgh-quick-mention) {
 	z-index: 4;
 }


btea:
I've adjusted the solution, and it seems to be working correctly so far. However, the `:has` selector is only compatible with Chrome 105+.

fregante (approved):
Better. We should try it in our local CSS field before merging

SunsetTechuila:
Caused by `sticky-comment-header`

Johnqertest:
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2FHw8vYF4DNRCKY%2Fgiphy.gif&token=shipit" data-gitmeme-token="shipit">shipit</a>

source/features/quick-mention.css
```diff
@@ -28,3 +28,7 @@
 .rgh-quick-mention:not(:hover, :focus) {
 	color: inherit;
 }
+
+.TimelineItem:has(.rgh-quick-mention.tooltipped) {
+	z-index: 5;
+}
```
fregante: Instead of pushing the whole block up, you can limit it to `.TimelineItem-avatar`

```suggestion
.TimelineItem-avatar:has(.rgh-quick-mention) {
	z-index: 4;
}
```


This seems to work for me, testing on https://github.com/facebook/react/pull/35363#pullrequestreview-3580274496

Image

PR #8845, opened 2026-01-04
`conventional-commits` - Support uppercase type

btea:
Some commits may have a type with the first letter capitalized; perhaps we can make the type parsing process more compatible.


fregante:
This is intentionally not supported because it's not in the spec

fregante:
Although… it's a pretty easy change 👌

fregante (approved):

PR #8854, opened 2026-01-07
`no-self-reference` - New feature

### LINKED ISSUE #8804: Highlight pr self-closing link
### Description

There is a common mistake in a pull request.

Since user may switch between pr and different issues, then write `Close/Fixes #123`, and paste the wrong issue number, which is the PR self.

Is it possible to point it out when browsing the pr page?

### Example URLs

https://github.com/prettier/prettier/pull/9275

This PR didn't link the right issue

Harrish92:
closes #8804

## Test URLs

https://github.com/refined-github/sandbox/pull/120
https://github.com/refined-github/sandbox/issues/122

## Screenshot

Screenshot 2


source/features/underline-pr-self-reference.tsx
```diff
@@ -0,0 +1,47 @@
...
+}
+
+function init(signal: AbortSignal): void {
+	observe([
+		'.comment-body',
+		'.markdown-body',
+		'.js-comment-body',
+		'.soft-wrap',
+		'.user-select-contain',
+		'.d-block',
```
fregante: Way too broad. Here there should be a single selector for the specific issue link already. I think it's easy to select, something like: `.whatever class in the first PR comment a[href=${location.href}]` (the href needs to be cleaned from #hash though)

Harrish92: I tried doing `location.href.split('#')[0]` , but it doesn't seem to work. So I just opted for a selector that is more specific.

source/features/underline-pr-self-reference.tsx
```diff
@@ -0,0 +1,35 @@
+import * as pageDetect from 'github-url-detection';
+
+import features from '../feature-manager.js';
+import observe from '../helpers/selector-observer.js';
+import {getConversationNumber} from '../github-helpers/index.js';
+
+function underlinePrSelfReference(prConversation: HTMLElement): void {
+	const prNumber = getConversationNumber();
+
+	const prLink = prConversation.textContent;
```
fregante: You'll have to use the URL because the text is rendered in bullet points:

- close #8854

source/features/underline-pr-self-reference.tsx
```diff
@@ -0,0 +1,35 @@
...
+		pageDetect.isPR,
+	],
+	init,
+});
+
+/*
+
+## Test URLs
+
+https://github.com/prettier/prettier/pull/9275
```
fregante:
```suggestion
https://github.com/refined-github/sandbox/pull/120
```

Make sure it follows the markings in that PR (ignore/underline). In short it should underline any links to the current page, unless they point to a specific comment.

Under this optic, I wonder if the feature should be a more general `no-self-reference` and apply to issues too. Linking to the same issue or same PR (including Files tab) is likely an issue.

Effectively it should target the **exact** raw links (no hash, no extra slashes):

- ``buildRepoURL(`issues/${getConversationNumber()}`)``
- ``buildRepoURL(`pull/${getConversationNumber()}`)``

source/features/underline-pr-self-reference.tsx
```diff
@@ -0,0 +1,35 @@
...
+import {getConversationNumber} from '../github-helpers/index.js';
+
+function underlinePrSelfReference(prConversation: HTMLElement): void {
+	const prNumber = getConversationNumber();
+
+	const prLink = prConversation.textContent;
+	const prSelfReference = '#' + prNumber!.toString();
+
+	if (prLink.includes(prSelfReference)) {
+		prConversation.style.textDecoration = 'underline wavy red';
```
fregante: It should also include `tooltipped` to tell the user what's wrong

> This link points to the current PR, it's most likely a mistake

Harrish92: I warn the user by setting the `title` attribute.

source/features/no-self-reference.tsx
```diff
@@ -0,0 +1,30 @@
+import * as pageDetect from 'github-url-detection';
+
+import features from '../feature-manager.js';
+import observe from '../helpers/selector-observer.js';
+
+function underlineSelfReference(link: HTMLElement): void {
+	link.style.textDecoration = 'underline wavy red';
+	link.title = 'Link is a self-reference';
```
SunsetTechuila:
```suggestion
	link.classList.add("tooltipped", "tooltipped-s");
	link.ariaLabel = "This link is a self-reference. It's most likely a mistake";
```

tooltipped:

image

title:

image

source/features/no-self-reference.tsx
```diff
@@ -0,0 +1,39 @@
...
+import observe from '../helpers/selector-observer.js';
+
+function underlineSelfReference(link: HTMLElement): void {
+	link.title = 'Link is a self-reference';
+
+	// Disable link and hovercard
+	link.removeAttribute('href');
+	link.removeAttribute('data-hovercard-url');
+
+	// TODO: Use shorthand `text-decoration` property in 2027 (due to Safari 18)
```
SunsetTechuila, line 13: https://github.com/refined-github/refined-github/pull/8878/changes#diff-afd8f4f4844a5ad7b1b98d25703a5ce8238465fe439a9718083e279b925f44caR18

SunsetTechuila:
this link is clickable -> #8854 if you enter the editing mode and switch to "Preview"

fregante:
Damn 😅 

I guess the easiest think would be to rebuild the element.  @Harrish92 something like

```jsx
link.replaceWith(
	<span style={{etc}} title="etc">{link.textContent}</span>
);
```

Edit: this won't work well with links rendered inside bullet points.

fregante:
That's weird though, are we doing that? If I disable Refined GitHub and remove the `href` attribute, it won't open a new tab.

fregante:
It's good enough for now, it can be updated separately, but honestly it's not worth it.

SunsetTechuila:
> That's weird though, are we doing that?

It's `prevent-comment-loss`

fregante:
How is the feature able to get the href if we remove it? 😅 

I think an extra `a[href]` in its delegate selector might help

PR #8866, opened 2026-01-12
`reactions-avatars`, `small-user-avatars` - Fix Copilot user

### LINKED ISSUE #8864: `reactions-avatars` shows bad icon for copilot
### Description

Image

When you mention copilot, it will leave a temporary reaction. The avatar we show is wrong 

https://github.com/refined-github/shorten-repo-url/pull/70#issuecomment-3734370625

### Extension version

25.11.4

### Browser(s) used

Safari

### Token ⚠️

SunsetTechuila:
resolves #8864

## Test URLs

https://github.com/refined-github/refined-github/issues/7761

https://github.com/aio-libs/aiohttp/pull/3809#issuecomment-498809698

https://github.com/refined-github/refined-github/pull/8848#issuecomment-3732149089

## Screenshot

image

image

image

### DIFF
--- a/source/features/small-user-avatars.tsx
+++ b/source/features/small-user-avatars.tsx
@@ -3,41 +3,45 @@ import React from 'dom-chef';
 import onetime from '../helpers/onetime.js';
 import features from '../feature-manager.js';
 import observe from '../helpers/selector-observer.js';
-import {getUserAvatar} from '../github-helpers/get-user-avatar.js';
+import getUserAvatarURL from '../github-helpers/get-user-avatar.js';
 import './small-user-avatars.css';
 
-function addAvatar(link: HTMLAnchorElement): void {
-	const username = link.classList.contains('user-mention')
-		? link.href.split('/').pop()! // #8389
-		: link.textContent;
-	const avatar = getUserAvatar(username, 14);
+function addAvatar(link: HTMLElement): void {
+	const username = link.textContent;
+	const size = 14;
 
 	link.classList.add('d-inline-block', 'lh-condensed-ultra');
 	link.prepend(
-		<span
+		<img
 			className="avatar avatar-user v-align-text-bottom mr-1 rgh-small-user-avatars"
-			aria-hidden="true" // https://github.com/refined-github/refined-github/issues/8218#issuecomment-3762891075
-		>
-			{typeof avatar === 'string'
-				? <img src={avatar} loading="lazy" />
-				// @ts-expect-error -- createElement should infer the props, but it doesn't
-				: React.createElement(avatar, {width: '75%', height: '75%'})
-			}
-		</span>,
+			src={getUserAvatarURL(username, size)!}
+			width={size}
+			height={size}
+			loading="lazy"
+		/>,
 	);
+}
+
+function addMentionAvatar(link: HTMLAnchorElement): void {
+	// Don't use textContent #8389
+	const username = link.href.split('/').pop()!;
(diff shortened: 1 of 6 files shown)


source/features/small-user-avatars.tsx
```diff
@@ -23,7 +23,7 @@ function addAvatar(link: HTMLElement): void {
 }
 
 function addMentionAvatar(link: HTMLAnchorElement): void {
-	const username = link.href.split('/').pop()!;
+	const username = link.textContent.slice(1);
```
fregante: Feels like we did the opposite a few years ago. Worth a git blame on this line

SunsetTechuila: Nice catch. #8389

SunsetTechuila: Do we really need to support hovercards? Maybe we can just return if a mention is truncated?

fregante: I think it's not too much trouble to drop /apps/ or whatever appears in bots' URLs. The copilot-specific exception is fine too if it's only one. 

Either way it's worth leaving a comment pointing to that PR explaining why we don't use textContent

source/github-helpers/get-user-avatar.ts
```diff
@@ -2,25 +2,25 @@ import {$optional} from 'select-dom/strict.js';
...
-	// Find image on page. Saves a request and a redirect + add support for bots
+	// Find image on page. Saves a request and a redirect
 	const existingAvatar = $optional(`[href="/${cleanName}" i] img`);
-
 	if (existingAvatar) {
 		return existingAvatar.src;
 	}
 
-	// If it's not a bot, use a shortcut URL #2125
-	if (cleanName === username) {
```
SunsetTechuila, line 19: It's technically possible for bots to have a name/avatar different from the user or organization that registered them, so `https://avatars.githubusercontent.com/${cleanName}` won't work in those cases. Still, I consider that scenario rare enough for bots that leave reactions (I couldn't find any examples) that it's probably not worth accounting for it

source/features/small-user-avatars.tsx
```diff
@@ -40,7 +40,7 @@ function initOnce(): void {
 			[data-testid="closed-at"]
 		) a[data-hovercard-url*="/users"]`, // `isIssueList`
 	], addAvatar);
-	observe('.user-mention:not(.commit-author)[data-hovercard-type="user"]', addMentionAvatar);
+	observe('.user-mention:not(.commit-author)', addMentionAvatar);
```
SunsetTechuila: Is this something we want?

image

fregante: Sure I think it makes sense

source/github-helpers/get-user-avatar.ts
```diff
@@ -1,7 +1,11 @@
 import {$optional} from 'select-dom/strict.js';
 import * as pageDetect from 'github-url-detection';
+import CopilotIcon from 'octicons-plain-react/Copilot';
```
fregante: Why? Can't we load the image from GitHub somewhere? I don't want this extra complexity for little improvement

SunsetTechuila: > for little improvement

On my main monitor (27" QHD) raster Copilot avatar is barely readable (the screenshots attached are 150% zoom). But alright, reverted

SunsetTechuila: Good that this change didn't make it into the main branch

>  raster Copilot avatar is barely readable

It's much better now:

image

PR #8868, opened 2026-01-14
`repo-avatars` - Restore feature

kovsu:
## Test URLs

https://github.com/refined-github/refined-github/

## Screenshot

image

### DIFF
--- a/source/features/repo-avatars.tsx
+++ b/source/features/repo-avatars.tsx
@@ -30,7 +30,10 @@ async function add(ownerLabel: HTMLElement): Promise<void> {
 }
 
 function init(signal: AbortSignal): void {
-	observe('header.GlobalNav [data-testid="top-nav-center"] ol > li:first-child span', add, {signal});
+	observe([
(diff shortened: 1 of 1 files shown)


kovsu:
`repo-header-info` is also broken now.

fregante:
> [`repo-header-info`](https://github.com/refined-github/refined-github/blob/main/source/features/repo-header-info.tsx) is also broken now.

See also https://github.com/refined-github/refined-github/issues/8867

source/features/repo-avatars.tsx
```diff
@@ -30,7 +30,7 @@ async function add(ownerLabel: HTMLElement): Promise<void> {
 }
 
 function init(signal: AbortSignal): void {
-	observe('.AppHeader-context-full [role="listitem"]:first-child .AppHeader-context-item-label', add, {signal});
```
fregante, line 33: We need to preserve the old selectors where possible. It seems that they're still doing AB testing. All features still work on my Safari desktop but not Safari iPhone (e.g. `bugs-tab`)

kovsu:
> See also https://github.com/refined-github/refined-github/issues/8867

I saw it. But the main problem is that the header seems lose its dropdown.

PR #8869, opened 2026-01-14
`repo-header-info` - Restore feature

kovsu:
## Test URLs

https://github.com/refined-github/refined-github/

## Screenshot

CleanShot 2026-01-14 at 16 24 56@2x

### DIFF
--- a/source/features/repo-header-info.tsx
+++ b/source/features/repo-header-info.tsx
@@ -7,8 +7,6 @@ import React from 'dom-chef';
 import {elementExists} from 'select-dom';
 import {CachedFunction} from 'webext-storage-cache';
 
-import {$} from 'select-dom/strict.js';
-
 import observe from '../helpers/selector-observer.js';
 import features from '../feature-manager.js';
 import api from '../github-helpers/api.js';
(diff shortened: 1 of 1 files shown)


source/features/repo-header-info.tsx
kovsu, line 51: I don't know if we can drop these code. It seems that GitHub had already supported it.

fregante: We kept it because GitHub was inconsistent at the time. You can open a new PR to drop these for now and we can restore the feature if GitHub still fails

source/features/repo-header-info.tsx
```diff
@@ -77,7 +83,7 @@ async function add(repoLink: HTMLAnchorElement): Promise<void> {
 
 async function init(signal: AbortSignal): Promise<void> {
 	await expectToken();
-	observe('.AppHeader-context-full [role="listitem"]:last-child a.AppHeader-context-item', add, {signal});
+	observe('header.GlobalNav [data-testid="top-nav-center"] ol > li:last-child', add, {signal});
```
fregante: Likewise we need to preserve the old selector

source/features/repo-header-info.tsx
```diff
@@ -33,7 +35,9 @@ const repositoryInfo = new CachedFunction('stargazer-count', {
 	cacheKey: cacheByRepo,
 });
 
-async function add(repoLink: HTMLAnchorElement): Promise<void> {
+async function add(ownerWithNameLabel: HTMLElement): Promise<void> {
+	const repoLink = $('a', ownerWithNameLabel);
```
fregante: This won't work on the old version, might be best to move it to the observer

kovsu: Oh im dumb.

PR #8749, opened 2025-11-08
`restore-file` - Add support for new PR Files

ndom91:
Fixes: https://github.com/refined-github/refined-github/issues/8714

Reopening #8715. Good call, the new file name was available in the dropdown on the first party links

- Unfortunately the since the dropdowns are now portaled we can't get the old filename from the file header in the main diff view. That old filename really doesn't seem to be anywhere in the portaled dropdown unfortunately.
- But it looks like the new react UI props are available in a script tag that we can leverage. Was able to grab the old file name from any changes that are labelled "RENAMED" from there. What do you think of using those generally?

For example: 

<details>
<summary><code>Set of react props from PR page script tag</code></summary>

```json
{
    "payload": {
        "isSingleFileMode": false,
        "aliveChannel": "eyJjIjoicHVsbF9yZXF1ZXN0OjI5OTE1ODAzMDUiLCJ0IjoxNzYyNjI1ODAyfQ==--916b0885b1749d5aa89bf6de209b19f96ba2481b586140d4f928ca8b4c565055",
        "bannersData": {
            "banners": {
                "dependabotAutomatedSecurityUpdates": {
                    "render": false
                },
                "pausedDependabotUpdate": {
                    "render": false
                },
                "hiddenCharacterWarning": {
                    "render": false
                }
            }
        },
        "changeGroupGeneration": {
            "canAccessChangeGroups": false,
            "changeGroupFeedbackOptions": [
                {
                    "label": "Groups contain changes that do not belong together",
                    "value": "NOT_RELEVANT"
                },
                {
                    "label": "Groups are harmful or unsafe",
                    "value": "OFFENSIVE_OR_DISCRIMINATORY"
                },
                {
                    "label": "Groups are poorly formatted",
                    "value": "POORLY_FORMATTED"
                },
                {
                    "label": "Groups are not helpful",
                    "value": "NOT_HELPFUL"
                },
                {
                    "label": "Groups miss changes that should be included",
                    "value": "MISSED_CONTENT"
                },
                {
                    "label": "Other",
                    "value": "OTHER"
                }
            ],
            "changeGroups": [],
            "unavailableReason": null,
            "latestChangeGroupOrchestration": null,
[...]

PR #8870, opened 2026-01-14
`restore-file` - Add support for new PR view

maschwenk:
Fixes

- https://github.com/refined-github/refined-github/issues/8714

Reimplementation of 
- #8749

addressing the maintainer's feedback, trying to sidestep drama. In honesty I don't totally know how to test this.

## Changes

- Added support for the new React-based PR files view while keeping the legacy view handling completely unchanged
- Use `GitHubFileURL` helper for URL parsing instead of manual regex
- Track focused file container via a global variable (`focusedFileContainer`) for removal after discard operation
- Use the octicon library (`GitCompareIcon`) for the menu item icon
- Handle renamed files by checking React props for `RENAMED` status with the suggested conditional expression
- Use `handleMenuOpening` pattern from `more-file-links.tsx` for adding menu items

## Key differences from previous PR

1. **Legacy code unchanged**: The legacy view handling (`addLegacyMenuItem`, `observe`) remains exactly as it was
2. **Global file tracking**: Following the suggestion to save the file container on menu open for later removal
3. **Cleaner URL parsing**: Using `GitHubFileURL` instead of manual string manipulation
4. **Proper getFilenames function**: Lives in `restore-file.tsx` where it belongs, returns `{original, new}` as suggested

## Test URLs

- https://github.com/refined-github/sandbox/pull/16/files
- https://github.com/refined-github/sandbox/pull/29/files

### DIFF
--- a/source/features/restore-file.tsx
+++ b/source/features/restore-file.tsx
@@ -157,8 +157,7 @@ function addLegacyMenuItem(editFile: HTMLAnchorElement): void {
 // New React view handler: track the file container and add menu item
 function handleMenuOpening({delegateTarget}: DelegateEvent<MouseEvent, HTMLElement>): void {
 	// Track the file container for later removal
-	focusedFileContainer = delegateTarget.closest('[class*="DiffFileHeader-module__diff-file-header"]')!
-		.parentElement!;
+	focusedFileContainer = delegateTarget.closest('div[id^="diff-"]')!;
 
 	// Wait for the menu to be rendered
 	requestAnimationFrame(() => {


source/features/restore-file.tsx
```diff
@@ -1,7 +1,9 @@
 import React from 'dom-chef';
+import {$} from 'select-dom/strict.js';
 import delegate, {type DelegateEvent} from 'delegate-it';
 import * as pageDetect from 'github-url-detection';
 import {stringToBase64} from 'uint8array-extras';
+import GitCompareIcon from 'octicons-plain-react/GitCompare';
```
SunsetTechuila: I think `Undo` is more suitable here 

Image

source/features/restore-file.tsx
```diff
@@ -120,10 +171,45 @@ function add(editFile: HTMLAnchorElement): void {
...
+
+	// Wait for the menu to be rendered
+	requestAnimationFrame(() => {
+		const editFile = $('[class^="prc-ActionList-ActionListItem"]:has(.octicon-pencil)');
+
+		// Skip if already added
+		if (editFile.nextElementSibling?.classList.contains('rgh-restore-file')) {
+			return;
+		}
+
```
SunsetTechuila: > if already added

This will never happen 

```suggestion
```

maschwenk: 👍🏼 address here https://github.com/refined-github/refined-github/pull/8870/commits/4d94e9bd5d62164d64ebc6fd0d486d87f1818987

source/features/restore-file.tsx
maschwenk: 👍🏼 addressed here 9c8dcfd

source/features/restore-file.tsx
```diff
@@ -85,30 +91,75 @@ async function discardChanges(progress: (message: string) => void, originalFileN
...
+	const diffContents = reactProps.payload.diffContents.find(
+		(dc: {path: string}) => dc.path === newFileName,
+	);
+
+	const originalFileName = diffContents?.status === 'RENAMED'
+		? diffContents.oldTreeEntry.path
+		: newFileName;
+
+	return {original: originalFileName, new: newFileName};
```
SunsetTechuila: You can do:

```tsx
const [originalFileName, newFileName = originalFileName] = $('.Link--primary span:not(.sr-only)', focusedFileContainer).textContent.split('  ')
```

maschwenk: [94a7d16](https://github.com/refined-github/refined-github/pull/8870/commits/94a7d162b7d55f32fb59f7946f6dfb4cdc15ab09)

done

source/features/restore-file.tsx
```diff
@@ -85,30 +90,60 @@ async function discardChanges(progress: (message: string) => void, originalFileN
 	});
 }
 
+function getFilenames(menuItem: HTMLElement): {original: string; new: string} {
+	// Legacy view: get filenames from the data-path and Link--primary elements
+	// TODO: Drop in June 2026
```
SunsetTechuila: This is probably premature. The policy is to keep support for **removed** UI for 6 months. The React-based view is still in preview and could remain there for a couple more months. The same applies to lines 133, 145, and 181.

source/features/restore-file.tsx
SunsetTechuila: The variant I previously suggested works only for renamed files and doesn't account for invisible LRM chars.

Here is a better one:

```tsx
	const fileNameElement = $('[class^="DiffFileHeader-module__file-name"]', focusedFileContainer);
	const span = $optional('span:not(.sr-only)', fileNameElement);
	const [originalFileName, newFileName = originalFileName] = (span ?? fileNameElement)
		// eslint-disable-next-line unicorn/prefer-string-replace-all -- Invisible char
		.textContent.split('  ').map(text => text.replaceAll(/\u200E/g, ''));
```

source/features/restore-file.tsx
```diff
@@ -120,10 +155,39 @@ function add(editFile: HTMLAnchorElement): void {
 	);
 }
 
+// New React view handler: track the file container and add menu item
+function handleMenuOpening({delegateTarget}: DelegateEvent<MouseEvent, HTMLElement>): void {
+	// Track the file container for later removal
+	focusedFileContainer = delegateTarget.closest('[class*="DiffFileHeader-module__diff-file-header"]')!
+		.parentElement!;
```
SunsetTechuila: You are selecting only a header 

```suggestion
	focusedFileContainer = delegateTarget.closest('div[id^="diff-"]')!;
```

maschwenk:
@SunsetTechuila I was also able to follow the good instructions in https://github.com/refined-github/refined-github/wiki/Contributing to help me actually test this by the end. And I can confirm it's working! Thanks so much!

maschwenk:
@SunsetTechuila @fregante do I need to do anything to make it mergeable? 😆 I want this so badly it's my favorite feature of RG

SunsetTechuila:
> do I need to do anything

Only wait a bit

fregante:
Thanks @maschwenk!

PR #8875, opened 2026-01-18
`repo-header-info` - Use DOM to determine if user starred repo

SunsetTechuila:
reduces user confusion by avoiding the use of potentially outdated data

## Test URLs


SunsetTechuila:
Nvm

fregante:
By the way, we moved away from using more than one source. We used have like 4 sources to get the default branch for example, but since this was cached long term we'd occasionally get the wrong data and not know its origin. Debugging it would mean having to cycle through multiple pages, emptying the cache every time, etc

fregante:
I say this just in case you want to add this with an API call *fallback*

PR #8877, opened 2026-01-18
`quick-comment-hiding` - Drop feature

SunsetTechuila:
https://github.com/refined-github/refined-github/blob/54d1d7ff19ef3a3de94966f993d0b29e248b9b7c/source/features/quick-comment-hiding.tsx#L89-L90


fregante:
Some of those dates were overly optimistic. The feature is still needed for PRs and will stay here until the end of 2026 most likely

SunsetTechuila:
> The feature is still needed for PRs

Indeed🤦‍♂️

PR #8881, opened 2026-01-20
Fix feature loading on soft navigation between React-based pages

SunsetTechuila:
partially _fixes #6554

## Test URLs

- https://github.com/refined-github/refined-github/issues/new/choose -> https://github.com/refined-github/refined-github/issues/new?template=1_bug_report.yml

- https://github.com/refined-github/refined-github/blob/main/source/features/action-used-by-link.tsx -> other file (checkout https://github.com/refined-github/refined-github/tree/rgh-feat-desc)


package.json
```diff
@@ -60,8 +60,8 @@
 		"linkify-urls": "^5.2.0",
 		"memoize": "^10.2.0",
 		"octicons-plain-react": "^19.20.0",
-		"one-event": "^4.3.0",
 		"one-mutation": "^3.0.1",
+		"p-event": "^7.1.0",
```
fregante: Uh, why. p-event is much bigger

SunsetTechuila: Because `one-event` doesn't support multiple target events:

https://github.com/refined-github/refined-github/blob/68809587ce648fdec007d8830311bf30392e9682/source/feature-manager.tsx#L219

https://github.com/refined-github/refined-github/blob/68809587ce648fdec007d8830311bf30392e9682/source/feature-manager.tsx#L154

Should I open a PR to `one-event` instead?

fregante: 🚀

fregante:
```
 one-event 4.4.0 published 🎉
```

source/features/linkify-text.tsx
```diff
@@ -6,12 +6,11 @@ import features from '../feature-manager.js';
 import observe from '../helpers/selector-observer.js';
 import {getRepo} from '../github-helpers/index.js';
 import {linkifyIssues} from '../github-helpers/dom-formatters.js';
-import {logError} from '../helpers/errors.js';
 
 function linkifyIssue(paragraph: HTMLParagraphElement): void {
 	// Already linkified
 	if (elementExists('a', paragraph)) {
-		logError(new Error(`${paragraph.textContent} is already linkified`));
```
fregante, line 14: Why remove this? This change seems to suggest that features will now run multiple times on the same view. 

The PR also seems to add a listener  for the "render" but not for the "unload", causing all features to potentially be `init`ed multiple times. What happens if I call `observe('a');observe('a')` for example? I think it's deduplicated almost correctly, but then does the abort signal work well?

The alternative could be to also unload the features on "soft nav react bye" if it exists, but then most likely we'll get flickering from features coming and going.

I think a quick way to test whether the PR works as-is would be to open the Task Manager and go back and forth between React tabs. If the page doesn't leak and slow down considerably after 50 back and forth, then we should be good.

source/feature-manager.tsx
```diff
@@ -216,7 +216,7 @@ async function add(url: string, ...loaders: FeatureLoader[]): Promise<void> {
 					}
 				}
 			});
-		} while (await oneEvent(document, 'turbo:render'));
+		} while (await oneEvent(document, ['turbo:render', 'soft-nav:react-done']));
```
SunsetTechuila, line 221: unlike `soft-nav:render`, `soft-nav:react-done` is fired when user navigates the session history too

source/features/linkify-text.tsx
SunsetTechuila: > The PR also seems to add a listener for the "render" but not for the "unload", causing all features to potentially be `init`ed multiple times.

No, features are unloaded correctly, you don't need to listen for any new events

https://github.com/refined-github/refined-github/blob/5f6060f8d08ad94f7d1df88462dbbf563258f820/source/feature-manager.tsx#L129-L130

SunsetTechuila: Normally you won't see this error, but it can happen if you try hard enough 

To reproduce:

1. Open a repo's issue list
2. Open an issue
3. Click the Issues navbar tab
4. Hit the browser back button to return to the issue

Timings are important - you've to be fast, but not too fast. Usually takes me 2-3 tries. Сan't say for sure yet what exactly is happening and why

SunsetTechuila: > No, features are unloaded correctly, you don't need to listen for any new events
> 
> https://github.com/refined-github/refined-github/blob/5f6060f8d08ad94f7d1df88462dbbf563258f820/source/feature-manager.tsx#L129-L130

This is no longer true (but certainly was)

SunsetTechuila: I suppose our only option is to unload the features on the navigate "push"/"traverse" events

image

fregante: The Navigation API should now be available cross browser, I'm just afraid it fires too often. Worth a check too

SunsetTechuila: [I mean](https://developer.mozilla.org/docs/Web/API/Navigation/navigate_event#browser_compatibility)

image

SunsetTechuila: I checked to see if we could hook some function to emit custom navigation events, but none of the breakpoints I set on navigation-related functions were hit. I might be missing something, though. If you want to check for yourself, the `ReactAppElement` source is a good starting point.

SunsetTechuila: > but none of the breakpoints I set on navigation-related functions were hit

Managed to hit them - the functions are being used, just not when navigating from the issue list to an issue page

SunsetTechuila: Since the navigation from the issue list to an issue page is no longer a soft one, features are loading and unloading as expected. Other scenarios are covered by https://github.com/refined-github/refined-github/pull/8881/changes/8a551b82582e7c1099409c7887a8ba2588516de3

source/feature-manager.tsx
```diff
@@ -128,7 +128,7 @@ const globalReady = new Promise<RGHOptions>(async resolve => {
 	// https://github.com/refined-github/refined-github/issues/6437#issuecomment-1489921988
 	document.addEventListener('turbo:before-fetch-request', unloadAll); // Clicks
 	document.addEventListener('turbo:visit', unloadAll); // Back/forward button
-	document.addEventListener('soft-nav:progress-bar:start', unloadAll); // Soft navigation
+	document.addEventListener('soft-nav:start', unloadAll); // Soft navigation
```
SunsetTechuila, line 131: This is a bit too early, but other events are fired too late

source/feature-manager.tsx
```diff
@@ -128,6 +128,8 @@ const globalReady = new Promise<RGHOptions>(async resolve => {
 	// https://github.com/refined-github/refined-github/issues/6437#issuecomment-1489921988
 	document.addEventListener('turbo:before-fetch-request', unloadAll); // Clicks
 	document.addEventListener('turbo:visit', unloadAll); // Back/forward button
+	document.addEventListener('soft-nav:start', unloadAll); // Soft navigation
+	addEventListener('popstate', unloadAll); // History navigation
```
fregante, line 132: Do you reckon that we can proceed without these two early events or do they cause too much trouble/noise in the console? 

What does "firing too early" mean? Does the page break visually?

SunsetTechuila: > What does "firing too early" mean? Does the page break visually?

Here is a demonstration:

msedge_prwsHOQhvG

https://github.com/refined-github/refined-github/blob/815661e638b641ccb1674cef8564b43aec4d15db/source/features/rgh-feature-descriptions.tsx#L150-L152

https://github.com/refined-github/refined-github/blob/815661e638b641ccb1674cef8564b43aec4d15db/source/features/rgh-feature-descriptions.tsx#L157-L159

SunsetTechuila: Now I think why not *unload* on `soft-nav:render`? Seems to work well

SunsetTechuila: > Do you reckon that we can proceed without these two early events

We can't. Neither the `turbo:before-fetch-request` nor the `turbo:visit` events are fired in the test cases I provided in the PR description

SunsetTechuila: > Now I think why not _unload_ on `soft-nav:render`?

I've let this idea marinate in my head, and I think it's too risky

fregante:
I'm planning to release 26.3.14, this PR won't make the cut so don't force yourself to finish it

SunsetTechuila:
Current status of this PR:

- It works as-is
- Some features should be updated to take advantage of it (e.g. `rgh-feature-descriptions` - see https://github.com/refined-github/refined-github/tree/rgh-feat-desc), while others don't need changes (e.g. `no-self-reference`)
- Unloading on navigation start (or even earlier) looks too bad with client-side navigation on React-based views. See: https://github.com/refined-github/refined-github/pull/8881#discussion_r2736435748

Next steps:

- Replace at least the `soft-nav:start` listener with `soft-nav:render`. We might also want to remove `turbo:before-fetch-request`
- After that, we need to do **extensive** testing to make sure nothing breaks

fregante:
Judging by https://github.com/refined-github/refined-github/issues/6554#issuecomment-1713339754, this wouldn't be enough, at least for navigations that keep the same layout

> The main issue is that the DOM is no longer actually re-generated:
> 
> 1. our selector observer tags a "seen" element
> 2. after a page change the element is never removed
> 3. no `add` callback is ever fired

It could cover at least some page transitions though, like issue list -> issue

SunsetTechuila:
> Some features should be updated to take advantage of it (e.g. `rgh-feature-descriptions` - see https://github.com/refined-github/refined-github/tree/rgh-feat-desc)

fregante:
Right, but that has always been *the* issue. 

1. User navigates to https://github.com/refined-github/sandbox/issues/74
2. `locked-issue` is appended
3. User navigates to a different issue via direct link 
4. The state label is left behind 

I think this is doesn't happen for just one or two features. Any feature that appends *or even alters* DOM is going to be affected. 

We can still try this PR (or https://github.com/refined-github/refined-github/pull/6905 depending on which one has the closest *timing*) and then address the individual issues as they come up.

Issue #8882, opened 2026-01-20
Some `rgh-*` features are broken


fregante:
### Description

These features are meant to be silent so they're more likely to fail without anyone noticing.  It's worth testing each one of them to see if they're still working in their entirety. 

For example if you visit https://github.com/refined-github/refined-github/issues/new?template=1_bug_report.yml your should:

- have the version autofilled
- the token checkbox checked
- a gif shown if th checkbox is manually checked when the token is not set 

Again this is just one known broken feature but there might be more. 

### How to replicate the issue + URL

See above

### Extension version

25.12

### Browser(s) used

Safari iOS 

### Token ⚠️

- [x] I set a token in the options page. I understand that some features require a token to work.

SunsetTechuila:
Regarding `rgh-improve-new-issue-form`: the `turbo-frame` element is not being replaced when going from https://github.com/refined-github/refined-github/issues/new/choose to https://github.com/refined-github/refined-github/issues/new?template=1_bug_report.yml, but it's child `react-app` is re-rendered
We have to replace `deduplicate` with observers:

https://github.com/refined-github/refined-github/blob/ab35209c089ba1604c03ef3dfae1730b69b0ad1c/source/features/rgh-improve-new-issue-form.tsx#L114

SunsetTechuila:
- `rgh-feature-descriptions` - ✅ works 
- `rgh-linkify-features` - ☑️ fixed by #8911
- `rgh-options-link` - ✅ works 
- `rgh-improve-new-issue-form` - ☑️ fixed by #8884
- `rgh-linkify-yolo` - ✅ works
- `rgh-netiquette` - ☑️ fixed by #8891
- `rgh-dim-commits` - ☑️ fixed by #8893
- `rgh-pr-template` - ✅ works 
- `rgh-token-user` - ✅ works

PR #8884, opened 2026-01-21
`rgh-improve-new-issue-form` - Restore feature

SunsetTechuila:
addresses #8882

The `turbo-frame` element is not replaced when navigating from https://github.com/refined-github/refined-github/issues/new/choose to https://github.com/refined-github/refined-github/issues/new?template=1_bug_report.yml, but it's child `react-app` is re-rendered, so `deduplicate` has to be replaced with observer

## Test URLs

https://github.com/refined-github/refined-github/issues/new/choose -> https://github.com/refined-github/refined-github/issues/new?template=1_bug_report.yml

## Screenshot

image

image

### DIFF
--- a/source/features/rgh-improve-new-issue-form.tsx
+++ b/source/features/rgh-improve-new-issue-form.tsx
@@ -16,7 +16,7 @@ const isSetTheTokenSelector = 'input[type="checkbox"][required]';
 const liesGif = 'https://github.com/user-attachments/assets/f417264f-f230-4156-b020-16e4390562bd';
 
 function addNotice(adjective: JSX.Element | string): void {
-	$('#issue_body_template_name').before(
+	$('[class^="IssueFormElements-module__formElementsContainer"]').prepend(
 		<div className="flash flash-error h3 my-9" style={{animation: 'pulse-in 0.3s 2'}}>
 			<p>
 				Your token is {adjective}. Many Refined GitHub features don't work without it.


source/features/rgh-improve-new-issue-form.tsx
SunsetTechuila: is this where it supposed to be placed?

Image

SunsetTechuila: Or it should be like this?

image

SunsetTechuila: At this point, I would rather disable the submit button than add a GIF and a giant banner

fregante: The position is fine either way. 

The bug could be about the token so it doesn't make sense to disable reporting altogether. The banner and gif are about reminding the user that they're doing something wrong, not Refined GitHub

fregante: I'm still always free to close, lock and block as I see fit for users who don't read any instructions. This setup has already stopped users from reporting bugs for missing tokens, which is what I was trying to do. Probably the welcome screen also helped

SunsetTechuila: found a screenshot https://github.com/refined-github/refined-github/pull/7413#issue-2341695278

PR #8887, opened 2026-01-22
`jump-to-conversation-close-event`, `netiquette`, `rgh-netiquette` - Restore

### LINKED ISSUE #8296: `rgh-netiquette`: Cannot read properties of undefined (reading 'getAttribute')
### Description

Even closing all features, i got:

```
TypeError: Cannot read properties of undefined (reading 'getAttribute')
    at getCloseDate (chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/assets/features/netiquette.js:24:38)
    at wasClosedLongAgo (chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/assets/features/netiquette.js:47:22)
    at init (chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/assets/features/rgh-netiquette.js:52:7)
    at chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/assets/feature-manager.js:167:26
    at chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/assets/helpers/async-for-each.js:6:52
    at Array.map (<anonymous>)
    at asyncForEach (chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/assets/helpers/async-for-each.js:6:34)
    at chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/assets/feature-manager.js:166:9
    at async asyncForEach (chrome-extension://hlepfoohegkhhmjieoechaddaejaokhf/assets/helpers/async-for-each.js:6:2)
```

https://github.com/refined-github/refined-github/issues/4166

### Extension version

25.2.17

### Browser(s) used

Chrome Version 133.0.6943.99 (Official Build) (arm64)

### Token ⚠️

SunsetTechuila:
jeez

addresses #8882

fixes #8296

### DIFF
--- a/source/features/netiquette.tsx
+++ b/source/features/netiquette.tsx
@@ -17,8 +17,6 @@ import {loadedConversationTimeline, newCommentField} from '../github-helpers/sel
 import {userIsModerator} from '../github-helpers/get-user-permission.js';
 import looseParseInt from '../helpers/loose-parse-int.js';
 
-// TODO: REFACTOR!!! At least when the PR page is rewritten in React
-
 /** Returns milliseconds passed since `date` */
 function timeAgo(date: Date): number {
 	return Date.now() - date.getTime();


source/github-helpers/selectors.ts
```diff
@@ -262,3 +262,8 @@ export const deletedHeadRepository_ = [
 	[2, 'https://github.com/refined-github/refined-github/pull/271'],
 	[1, 'https://github.com/refined-github/refined-github/pull/271/files'],
 ];
+
+export const loadedConversationTimeline = [
+	'#issue-timeline:not(:has(> [data-wrapper-timeline-id="load-top"]:last-child))',
```
SunsetTechuila, line 267: Either this or MutationObserver. The solution involving the latter turned out worse

source/features/netiquette.tsx
```diff
@@ -13,10 +13,12 @@ import features from '../feature-manager.js';
 import observe from '../helpers/selector-observer.js';
 import {buildRepoURL, isAnyRefinedGitHubRepo, isOwnConversation} from '../github-helpers/index.js';
 import {getLastCloseEvent} from './jump-to-conversation-close-event.js';
-import {newCommentField} from '../github-helpers/selectors.js';
+import {loadedConversationTimeline, newCommentField} from '../github-helpers/selectors.js';
 import {userIsModerator} from '../github-helpers/get-user-permission.js';
 import looseParseInt from '../helpers/loose-parse-int.js';
 
+// TODO: REFACTOR!!! At least when the PR page is rewritten in React
```
fregante: Refactor what? Any necessary refactor would be self-evident, no need to leave generic comments

source/features/netiquette.tsx
```diff
@@ -163,7 +166,6 @@ void features.add(import.meta.url, {
 	include: [
 		pageDetect.isConversation,
 	],
-	awaitDomReady: true, // We're specifically looking for the last event
```
fregante, line 166: ?

source/features/rgh-netiquette.tsx
```diff
@@ -1,21 +1,28 @@
...
 import createBanner from '../github-helpers/banner.js';
 import features from '../feature-manager.js';
 import observe from '../helpers/selector-observer.js';
 import {isAnyRefinedGitHubRepo} from '../github-helpers/index.js';
 import {getResolvedText, wasClosedLongAgo} from './netiquette.js';
 import {TimelineItem, TimelineItemOld} from '../github-helpers/timeline-item.js';
+import {loadedConversationTimeline} from '../github-helpers/selectors.js';
 
-function addConversationBanner(newCommentBox: HTMLElement): void {
+function addConversationBanner(): void {
```
fregante, line 14: This function accepts a field because there might be multiple fields on the page, e.g. when there are review comments in the PR Conversation tab.

.nvmrc
```diff
@@ -1 +1 @@
-20
+22
```
fregante, line 1: I'd extract this and the dependency bump to its own PR so I can merge them on sight and keep this PR more focused

source/github-helpers/selectors.ts
fregante: Unfortunately we can't use has() on large elements, Firefox just becomes unusable.

source/features/netiquette.tsx
SunsetTechuila: Not needed when `observe(loadedConversationTimeline)` is used

source/github-helpers/selectors.ts
SunsetTechuila: > the latter turned out worse

https://github.com/SunsetTechuila/refined-github/commit/bcfa9301c8d585222162e158342ef9b666512894

fregante: Nothing is worse than making GitHub so slow it's unusable. 

As for the mutation observer, I don't see why we need it, we already have a selector-based observer.

fregante: It looks like you're tracking the removal of the loader, we probably already have a ResizeObserver-based helper for that. 

But, why not just observe all close events and then take the last one of the list? If more elements are found, the link can be updated.

source/features/netiquette.tsx
fregante: I think it can still stay. There's no advantage in watching the page before it finishes loading.

SunsetTechuila:
It probably makes more sense to open a new PR that fixes just `jump-to-conversation-close-event` than to revert all the commits and start from scratch

source/github-helpers/selectors.ts
SunsetTechuila: Do you suggest to call `lastElement(conversationCloseEvent)` every time new `conversationCloseEvent` is observed?

fregante: I haven't looked at the logic, but you generally are not going to find more than 1-3 open/close events per issue, so "every time" is fine.

PR #8891, opened 2026-01-23
`netiquette`, `rgh-netiquette` - Fix banners

SunsetTechuila:
addresses #8882

fixes #8296

fixes #8908

doesn't work without #8889

## Test URLs

https://github.com/refined-github/refined-github/issues/4166

https://github.com/facebook/react/issues/13525

## Screenshot

image

| | |
|--------|--------|
| TimelineItem | image |
| TimelineItemOld  | image |

PR #8892, opened 2026-01-23
`jump-to-conversation-close-event` - Restore on long issues

SunsetTechuila:
addresses https://github.com/refined-github/refined-github/issues/8296#issuecomment-2765270023

depends on #8891

## Test URLs

https://github.com/refined-github/refined-github/issues/4166

### DIFF
--- a/source/features/jump-to-conversation-close-event.tsx
+++ b/source/features/jump-to-conversation-close-event.tsx
@@ -10,7 +10,7 @@ import {conversationCloseEvent} from '../github-helpers/selectors.js';
 import {getIdentifiers} from '../helpers/feature-helpers.js';
 import './jump-to-conversation-close-event.css';
 
-export const statusBadge = [
+export const statusBadgeSelector = [
 	'#partial-discussion-header .State:not(.rgh-locked-issue)',
 	'[data-testid="header-state"]',
 ] as const;
@@ -21,7 +21,7 @@ function updateStatusBadges(): void {
 	// Not processing the element that has been observed because past events may load in the middle of the page
 	const lastCloseEvent = lastElement(conversationCloseEvent);
 	const eventAnchor = $('a[href*="#event-"]', lastCloseEvent);
-	const statusBadges = $$(statusBadge);
+	const statusBadges = $$(statusBadgeSelector);
 
 	for (const statusBadge of statusBadges) {
 		const maybeWrapper = statusBadge.parentElement!;


source/github-helpers/selectors.ts
```diff
@@ -262,3 +262,20 @@ export const deletedHeadRepository_ = [
...
+export const conversationCloseEvent = [
+	// Old view (PRs)
+	`.TimelineItem:has(.TimelineItem-badge :is(
+		.octicon-issue-closed,
+		.octicon-git-merge,
+		.octicon-git-pull-request-closed,
+		.octicon-skip
+	))`,
+	// React view (Issues)
+	'[data-timeline-event-id]:has([data-testid="state-reason-link"])',
```
SunsetTechuila, line 275: Is it what you wanted to move to the selectors?

fregante: LGTM

SunsetTechuila:
I'm going to force push this to clean-up the commit history

source/features/jump-to-conversation-close-event.tsx
```diff
@@ -1,55 +1,51 @@
...
-		'[data-timeline-event-id]',
-	])?.querySelector('relative-time') ?? undefined;
-}
+const featureId = getFeatureID(import.meta.url);
 
-function addToConversation(discussionHeader: HTMLElement): void {
-	// Avoid native `title` by disabling pointer events, we have our own `aria-label`. We can't drop the `title` attribute because some features depend on it.
-	discussionHeader.style.pointerEvents = 'none';
+function updateStatusBadges(): void {
+	const lastCloseEvent = lastElement(conversationCloseEvent);
```
fregante, line 22: If you're using `debounce` correctly (i.e. fire on last call, not first) then `updateStatusBadges` should already receive the last close event, no need to query it again.

Using `observe` and then not using its element it's kind of a smell

source/features/jump-to-conversation-close-event.tsx
```diff
@@ -1,55 +1,51 @@
...
-		`[data-testid="state-reason-link"]:is(
-			[href*="reason%3Acompleted"],
-			[href*="reason%3Anot-planned"]
-		)`,
-	])?.closest([
-		'.TimelineItem', // Old view
-		'[data-timeline-event-id]',
-	])?.querySelector('relative-time') ?? undefined;
-}
+const featureId = getFeatureID(import.meta.url);
```
fregante: FeatureID doesn't contain the `rgh-` prefix we need to use for classes. There's `getIdentifiers` for this purpose

source/features/jump-to-conversation-close-event.tsx
SunsetTechuila: > If you're using `debounce` correctly (i.e. fire on last call, not first) then `updateStatusBadges` should already receive the last close event, no need to query it again.

That's what I initially did, but it's wrong. `updateStatusBadges` would receive the last **loaded** close event, which is not necessary the last **occurred** close event

SunsetTechuila: I left `debounce` to avoid calling `updateStatusBadges` multiple times on the initial load if there are more than one close event on the page

SunsetTechuila: > `updateStatusBadges` receives the last **loaded** close event, which is not necessary the last **occurred** close event

Example: https://github.com/refined-github/sandbox/issues/73

SunsetTechuila: > Using `observe` and then not using its element it's kind of a smell

That's why I asked: https://github.com/refined-github/refined-github/pull/8887#discussion_r2721503313

fregante: As the comment said, I only accepted the performance impact, not the logic

SunsetTechuila: > As the comment said, I only accepted the performance impact, not the logic

That doesn't address my other comments

fregante: Missed that part. Makes sense, please add an inline comment specifying that past events may load in the middle of the page and are not necessarily the last event.

SunsetTechuila:
Doesn't work on https://github.com/refined-github/sandbox/issues/73

source/features/jump-to-conversation-close-event.tsx
```diff
@@ -1,55 +1,52 @@
...
-	wrap(
-		discussionHeader,
-		<a
-			aria-label="Scroll to most recent close event"
-			className="tooltipped tooltipped-e"
-			href={getLastCloseEvent()!.closest('a')!.href}
-		/>,
-	);
+	for (const statusBadge of statusBadges) {
+		const maybeWrapper = statusBadge.parentElement!;
```
fregante, line 27:
```suggestion
		const maybeWrapper = statusBadge.parentElement as HTMLAnchorElement
```

`!` and `as` are both assertions, no need to use both

SunsetTechuila: We can't be sure that it is `HTMLAnchorElement` until we check the class

SunsetTechuila: I just dislike this block, but I can’t think of anything better. I guess it’s not worth caring about so much

image

fregante: Yes but it doesn't really matter, we only use it as an anchor. I'm not a fan of `(obj as T).prop` inline assertions, same as https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-await-expression-member.md but for assertions

PR #8901, opened 2026-01-24
Bump dependencies

fregante:


package.json
```diff
@@ -47,7 +47,7 @@
 		"dom-chef": "^5.1.1",
 		"dom-loaded": "^3.1.0",
 		"doma": "^4.0.0",
-		"element-ready": "^8.0.0",
+		"element-ready": "^9.0.1",
```
fregante: Blocked by

- https://github.com/sindresorhus/element-ready/pull/63

fregante: Unblocked, but needs to be tested

PR #8904, opened 2026-01-25
`sync-pr-commit-title` - Restore feature

### LINKED ISSUE #7096: `sync-pr-commit-title` - "Cancel" requires two clicks
### Description

I don’t think this is caused by us, but probably there's an event listener that cancels propagation so our `delegate()` never receives the event:

https://github.com/refined-github/refined-github/assets/1402241/ff3c3394-5f3b-4c63-b33b-d58347d9d412

It seems to be related with the focus of the field, because if I blur it first, it works with one click. This is also true when blurring via `tab`

https://github.com/refined-github/refined-github/assets/1402241/823d11a3-626e-4810-b0a9-a269eb4a624f

https://github.com/refined-github/sandbox/pull/12

### Extension version

23.11.3, 23.11.15

### Browser(s) used

Chrome, Safari

### LINKED ISSUE #8902: `sync-pr-commit-title` broken on new PR merge box
### Description

Image

1. Checkbox doesn't appear
2. PR title isn't updated after merge

1. any open PR you can merge

### Extension version

26.1.18

### Browser(s) used

Chrome

### Token ⚠️

SunsetTechuila:
fixes #8902 

fixes #7096

## Test URLs

https://github.com/refined-github/sandbox/pull/124

### DIFF
--- a/source/features/sync-pr-commit-title.tsx
+++ b/source/features/sync-pr-commit-title.tsx
@@ -38,7 +38,7 @@ function needsSubmission(): boolean {
 		return false;
 	}
 
-	const currentCommitTitle = getCurrentCommitTitle();
+	const currentCommitTitle = getCurrentCommitTitle()!;
 	return Boolean(currentCommitTitle) && (createCommitTitle() !== currentCommitTitle);
 }
 


source/github-events/on-commit-title-update.ts
```diff
@@ -2,13 +2,10 @@ import delegate, {type DelegateEventHandler} from 'delegate-it';
 
 const fieldSelector = [
 	'#commit-summary-input', // Commit title on edit file page
-	'#merge_title_field', // PR merge message field
+	'[class^="MergeBox-module__mergePartialContainer"] input', // PR merge message field
 ];
 
 export default function onCommitTitleUpdate(callback: DelegateEventHandler<Event, HTMLInputElement>, signal: AbortSignal): void {
-	// GitHub restores the value from the previous session and only triggers this event
-	delegate(fieldSelector, 'change', callback, {signal});
```
SunsetTechuila, line 10: GitHub doesn't restore changes made to commit message anymore

https://github.com/refined-github/refined-github/pull/5549/changes#r836514999

SunsetTechuila: also see https://github.com/refined-github/refined-github/issues/7096#issuecomment-3796582378

source/features/sync-pr-commit-title.tsx
```diff
@@ -9,12 +9,20 @@ import {getConversationNumber, userCanLikelyMergePR} from '../github-helpers/ind
...
 
 function getCurrentCommitTitleField(): HTMLInputElement | undefined {
-	return $optional(commitTitleFieldSelector);
+	const mergeButton = $optional(mergeButtonSelector);
+	if (mergeButton?.textContent === 'Confirm squash and merge') {
+		return $optional(commitTitleFieldSelector);
+	}
+
+	// eslint-disable-next-line no-useless-return -- "Not all code paths return a value. ts(7030)" if removed
+	return;
```
fregante: I think the expected fix would be to turn `| undefined` into `| void`

fregante: But what is the reason for this change? Does the optional selector match anything when  the button is "Confirm squash and merge"?

source/features/sync-pr-commit-title.tsx
```diff
@@ -46,7 +54,7 @@ function getUI(): HTMLElement {
 
 function updateUI(): void {
 	if (needsSubmission()) {
-		getCurrentCommitTitleField()!.after(getUI());
+		getCurrentCommitTitleField()!.parentElement!.after(getUI());
```
fregante, line 56: I assume this PR breaks support for the old merge box. I'm ok with that, as long as the old merge box is now unaffected by this feature. By the looks of it, there are still some observers, the UI is probably ok but the console will be spammed with errors due to this change. 

Is it easier to support the old view or to *not* support it *correctly?*

SunsetTechuila: Let's break down the current `init` function line by line:

1. `observe(commitTitleFieldSelector, updateCommitTitle, {signal});` - `commitTitleFieldSelector` has been updated; the old input field selector was removed
2. `observe('.gh-header-title', updateCommitTitle, {signal});` - `updateCommitTitle` calls `getCurrentCommitTitleField`, which now looks for the new merge button. If the new button isn't found, `updateCommitTitle` does nothing
3. `onCommitTitleUpdate(updateUI, signal);` - The `fieldSelector` used inside `onCommitTitleUpdate` has been updated; the old input field selector was removed
4. `delegate(mergeButtonSelector, 'click', updatePRTitle, {signal});` - `mergeButtonSelector` only targets the new merge button
5. `delegate('.rgh-sync-pr-commit-title', 'click', disableSubmission, {signal});` - `disableSubmission` will never be called on the old merge box (because of the change in point 3)

SunsetTechuila: > but the console will be spammed with errors due to this change.

So, this isn't true. That line won't be reached if the user has the old merge box

source/features/sync-pr-commit-title.tsx
SunsetTechuila: I'm not sure what are you asking about

SunsetTechuila:
I don't know how what's wrong with typescript

source/features/sync-pr-commit-title.tsx
```diff
@@ -9,12 +9,17 @@ import {getConversationNumber, userCanLikelyMergePR} from '../github-helpers/ind
...
-const commitTitleFieldSelector = '.is-squashing form:not([hidden]) input#merge_title_field';
+const commitTitleFieldSelector = '[data-testid="mergebox-partial"] input';
+const mergeButtonSelector = '[data-testid="mergebox-partial"] button[data-variant="primary"]';
 
-function getCurrentCommitTitleField(): HTMLInputElement | undefined {
-	return $optional(commitTitleFieldSelector);
+function getCurrentCommitTitleField(): HTMLInputElement | void {
+	const mergeButton = $optional(mergeButtonSelector);
+	if (mergeButton?.textContent === 'Confirm squash and merge') {
+		return $<HTMLInputElement>(commitTitleFieldSelector);
```
fregante: My question was, what happens if you just return this without any condition? Does the field exist even when it's not "squash and merge"?

It sounds like the condition is in the wrong place. There's already `needsSubmission` to exclude unsupported merge methods. 

Also this function returns void but then right after you're using `!.` after its call. There's some leaky abstraction here most likely, and that's probably related to needsSubmission. But I see that this already existed

source/features/sync-pr-commit-title.tsx
fregante: See new comment above.

source/features/sync-pr-commit-title.tsx
SunsetTechuila: > Does the field exist even when it's not "squash and merge"?

Yes, it also exists when you choose to create a merge commit

PR #8916, opened 2026-01-28
`conversation-activity-filter` - Save per-page setting for current tab

### LINKED ISSUE #8914: `conversation-activity-filter` - Remember settings after page refresh
### Description

I have GitHub Actions pipelines that are extremely chatty (generating many comments we can't prevent or delete), so the "Filter conversation activities" feature is an absolute life-saver!

Unfortunately, GitHub has a bad habit of only updating some parts of the PR page, which requires me to refresh the page regularly. After a refresh, the setting goes back to it's default, which means this button is practically my most interacted with UI element on this page 😅

I think it would be a very useful feature to remember the setting between page loads. 
This could be through cookies or local storage, and you can save it per repo (or per PR if you're feeling generous 😉).

I don't think enabling this by default will bother many users, especially not if you save it per repo (since every PR in a repo will roughly have the same structure, and if you want to hide messages for one, it's likely you want to do that too for others). It at least gives more control over how you like to view your page instead of being forced into the default setting every load

### Example URLs

https://github.com/refined-github/refined-github/pull/8878

SunsetTechuila:
resolves #8914 

## Test URLs

#8914

#8916

## Screenshot

msedge_D2tpoTf57R

### DIFF
--- a/source/features/conversation-activity-filter.tsx
+++ b/source/features/conversation-activity-filter.tsx
@@ -253,8 +253,7 @@ function switchToNextFilter(): void {
 }
 
 async function init(signal: AbortSignal): Promise<void> {
-	const sessionPageSetting = SessionPageSetting.get();
-	const initialState = sessionPageSetting
+	const initialState = SessionPageSetting.get()
 		?? (minorFixesIssuePages.some(url => location.href.startsWith(url))
 			? 'hideEventsAndCollapsedComments' // Automatically hide resolved comments on "Minor codebase updates and fixes" issue pages
 			: 'default');


fregante:
The issue was still `under discussion`, don't jump on new feature requests so fast, 😅 there are many approved issues labeled https://github.com/refined-github/refined-github/labels/small

SunsetTechuila:
> The issue was still `under discussion`

I know - I just really liked that suggestion. You don't have to merge or even review this PR if you think otherwise; I'm fine with that.

fregante:
Let's see how it goes. I think this is the first feature where we save the latest state, we've always pushed back against it.

PR #8918, opened 2026-01-29
Add support for deep-linking of features in extension settings

### LINKED ISSUE #8833: Highlight the target disabled feature on the options page
### Description

Image

It would be convenient if clicking this button not only opened the options page, but also expanded the feature list, scrolled to the target feature, and highlighted it

Image

### Example URLs

https://github.com/refined-github/refined-github/blob/main/source/features/clean-repo-sidebar.tsx

SunsetTechuila:
resolves #8833

## Screenshot

msedge_L7l5WpFNWB

### DIFF
--- a/source/background.ts
+++ b/source/background.ts
@@ -55,7 +55,7 @@ handleMessages({
 		const response = await fetch(url);
 		return response.json();
 	},
-	async openOptionsPage(hash?: string) {
+	async openOptionsPage(hash: string) {
 		return chrome.tabs.create({
 			url: chrome.runtime.getURL(`assets/options.html${hash && `#${hash}`}`),
 		});


SunsetTechuila:
> +57 −11

> [I don't see where the 50-100 LOC would be](https://github.com/refined-github/refined-github/issues/8833#issuecomment-3728525827)

image

source/helpers/open-options.tsx
```diff
@@ -1,13 +1,14 @@
...
 
-export default function openOptions(event: Event | React.UIEvent): void {
-	event.preventDefault();
-	void messageRuntime({openOptionsPage: true});
+export default function openOptions(hash?: string): void {
+	void messageRuntime({openOptionsPage: hash});
 }
 
-export function OptionsLink(): JSX.Element {
+export function OptionsLink({hash}: {hash?: string}): JSX.Element {
```
fregante: Unfortunately dom-chef doesn't support this. Attributes are set onto the returned element, not passed  to this function.

source/options.tsx
```diff
@@ -178,9 +178,31 @@ function addEventListeners(): void {
...
+
+	element.scrollIntoView({
+		block: 'center',
+	});
+}
+
 async function init(): Promise<void> {
 	await generateDom();
 	addEventListeners();
+	scrollTargetIntoView();
```
fregante, line 211: No JS for this please: https://github.com/refined-github/refined-github/issues/8833#issuecomment-3728666182

See also the mention of scroll-margin a couple of comments above.

source/helpers/open-options.tsx
fregante: But token doesn't really need this, it's not a "feature" as the issue requested and it's the first first-level block

fregante:
> > +57 −11
> 
> > [I don't see where the 50-100 LOC would be](https://github.com/refined-github/refined-github/issues/8833#issuecomment-3728525827)
> 
> image

If you implement things that don't need to be implemented, then you're going to write more code than needs to be written 😉

source/options.tsx
SunsetTechuila: > No JS for this please: [#8833 (comment)](https://github.com/refined-github/refined-github/issues/8833#issuecomment-3728666182)

No JS:

msedge_zwcOx7aMSS

fregante: What browser is that? My gif in that comment shows it working in Chrome

SunsetTechuila: > See also the mention of scroll-margin a couple of comments above.

I decided that scrolling to the center was better than setting `scroll-margin-top` to an arbitrary value

fregante: "Better" is arbitrary, all :targets work like this on the web: the linked anchor appears at the top. Here we can also highlight it, which improves it further.

SunsetTechuila: > What browser is that?

MS Edge. But I'm sure you'll get the same result with Chrome if you try this branch with `scrollTargetIntoView` commented

SunsetTechuila: My understanding is that the element is being added to the DOM too late

fregante: It's surprising that it's too slow for Edge considering it's a tiny SPA loaded locally, but that would explain the behavior.

SunsetTechuila: > it's too slow for Edge

It's not too slow for Edge - turns out Edge is too slow for it!

I just checked Helium and Brave, and they can handle this without JS in most cases. As a user, I obviously knew Edge had become terribly slow, but I never realized just how much. It's sad how that once-great browser has been enshittified to oblivion in only 1-1.5 years 😢

SunsetTechuila: Firefox needs JS too

source/options.css
```diff
@@ -140,14 +139,52 @@ summary:hover {
...
 	flex-shrink: 0;
+	scroll-margin-top: calc(
+		var(--outline-width) +
+		var(--padding-block) *
+		2 +
+		/* summary header */
+		1.5rem +
+		var(--summary-padding) +
+		1lh
+	);
```
fregante: I don’t think we need to be this specific here. Can we just use an approximate number? All of this feels like we're recreating the sizing/style of the header, which can change without changing any of these. 

It's easier to have '50px' and add a comment (which isn't necessary)

source/helpers/open-options.tsx
```diff
@@ -1,13 +1,14 @@
 import React from 'dom-chef';
 import {messageRuntime} from 'webext-msg';
 
-export default function openOptions(event: Event | React.UIEvent): void {
-	event.preventDefault();
```
SunsetTechuila, line 5: This breaks `rgh-options-link`

fregante: PreventDefault can be moved into the inline arrow function

source/features/rgh-options-link.tsx
```diff
@@ -16,7 +16,7 @@ function linkify(anchor: HTMLAnchorElement): void {
 
 function init(signal: AbortSignal): void {
 	observe(placeholdersSelector, linkify, {signal});
-	delegate(placeholdersSelector, 'click', openOptions, {signal});
+	delegate(placeholdersSelector, 'click', () => openOptions, {signal});
```
SunsetTechuila:
```suggestion
	delegate(placeholdersSelector, 'click', () => openOptions(), {signal});
```

Actually, this is what broke it + you can't call background script function with an argument of `undefined`

https://github.com/refined-github/refined-github/pull/8918#discussion_r3035881270

source/background.ts
```diff
@@ -55,8 +55,10 @@ handleMessages({
 		const response = await fetch(url);
 		return response.json();
 	},
-	async openOptionsPage() {
-		return chrome.runtime.openOptionsPage();
+	async openOptionsPage(hash?: string) {
+		return chrome.tabs.create({
+			url: chrome.runtime.getURL(`assets/options.html${hash && `#${hash}`}`),
```
fregante, line 60: Does `${false}` produce `''`?

SunsetTechuila: We can't get `${false}` - the type of the `hash` argument is `string`

source/options.tsx
```diff
@@ -176,9 +182,33 @@ function addEventListeners(): void {
 	$('#fetch-hotfixes').addEventListener('click', fetchHotfixes);
 }
 
+function scrollTargetIntoView(): void {
+	const {hash} = location;
+	if (!hash) {
+		return;
+	}
+
+	const element = $optional(hash);
```
fregante, line 191: I'd say this shouldn't be optional. Even if it's user-provided, we might as well throw the error

source/options.tsx
```diff
@@ -176,9 +182,33 @@ function addEventListeners(): void {
...
+	if (!element) {
+		return;
+	}
+
+	const details = element.closest('details');
+	if (details) {
+		details.open = true;
+	}
+
+	element.scrollIntoView({
```
fregante, line 201: I'm ok as long as this doesn't conflict with Chrome's native scrolling (conflict == janky animation)

fregante:
Anyway, merging because it's been long enough. Can be tweaked separately

source/options.tsx
SunsetTechuila: It doesn't

PR #8921, opened 2026-01-31
Meta: extract confirm merge button selector

SunsetTechuila:
This doesn't fix much by itself, I just decided it would be better to open a separate PR

Problems with current selectors:

- `[data-testid="mergebox-partial"] button[data-variant="primary"]` - selects "Select merge method" and merge method buttons

- `div[class^="MergeBox-module"] div[data-has-label] ~ div button[data-size="medium"][data-variant="primary"]` - doesn't select "Confirm auto-merge (squash)" and "Confirm rebase and merge"

## Test URLs

https://github.com/refined-github/sandbox/pull/4

### DIFF
--- a/build/__snapshots__/features-meta.json
+++ b/build/__snapshots__/features-meta.json
@@ -356,6 +356,7 @@
 	{
 		"id": "jump-to-conversation-close-event",
 		"description": "Adds a link to jump to the latest close event of a issue/PR.",
+		"css": true,
 		"screenshot": "https://user-images.githubusercontent.com/16872793/177792713-64219754-f8df-4629-a9ec-33259307cfe7.gif"
 	},
 	{
--- a/source/features/jump-to-conversation-close-event.css
+++ b/source/features/jump-to-conversation-close-event.css
@@ -1,0 +1,6 @@
+html:not([rgh-OFF-jump-to-conversation-close-event]) {
+	/* Fix tooltip getting cut off */
+	[class^='HeaderMetadata-module__metadataContent'] {
+		overflow: visible !important;
+	}
+}
--- a/source/features/jump-to-conversation-close-event.tsx
+++ b/source/features/jump-to-conversation-close-event.tsx
@@ -1,55 +1,52 @@
 import React from 'dom-chef';
-import {lastElement} from 'select-dom';
+import {$, $$, lastElement} from 'select-dom/strict.js';
 import * as pageDetect from 'github-url-detection';
+import debounce from 'debounce-fn';
 
 import {wrap} from '../helpers/dom-utils.js';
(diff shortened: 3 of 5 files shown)


source/features/sync-pr-commit-title.tsx
```diff
@@ -33,7 +33,7 @@ function createCommitTitle(): string {
 }
 
 function needsSubmission(): boolean {
-	const mergeButton = $optional(mergeButtonSelector);
+	const mergeButton = $optional(confirmMergeButton);
 	if (mergeButton?.textContent !== 'Confirm squash and merge') {
```
SunsetTechuila, line 37: should also update this to allow 'Confirm auto-merge (squash)'

source/github-events/on-pr-merge.ts
```diff
@@ -11,14 +12,7 @@ export default async function waitForPrMerge(signal: AbortSignal): Promise<void>
 		observe('.TimelineItem-badge.color-fg-on-emphasis .octicon-git-merge', resolve, {ancestor: 4});
 	});
 
-	await oneEvent([
-		// TODO: Drop in 2026
-		'.js-merge-commit-button',
```
fregante, line 16: When did the beta for this end? Feels like it was a recent feature, maybe we keep this selector a bit longer. There's a number of features that still have code for the old view

SunsetTechuila: > When did the beta for this end?

More than six month ago: https://github.blog/changelog/2025-03-04-improved-pull-request-merge-experience-is-now-generally-available

fregante: Cool! Do you want to remove all the code related to the old mergeability box in this PR? I'll merge it after v26.2.2 is released. It would delete a nice chunk of code I think.

SunsetTechuila: 🤝

SunsetTechuila: I haven't found much. There’s a bit of code (~5 lines in total) in `update-pr-from-base-branch` and `pr-branch-auto-delete`, but both features are broken. It would be better to remove that code in the PRs that fix them

SunsetTechuila: `safer-destructive-actions` could be updated, but that also worth a separate PR:

https://github.com/refined-github/refined-github/blob/a11f855268a6538caff697d76799325c4f487a25/source/features/safer-destructive-actions.css#L11-L25

fregante: `safer-destructive-actions` should just be dropped. It's a very old feature I readded recently but I'm not too sure about anymore. There's also an open issue

Issue #8617, opened 2025-08-09
Add button to delete current branch on `/tree/`


fregante:
### Description

1. You visit a repo's [branches list](https://github.com/refined-github/refined-github/branches)
2. Open a branch to see what's inside
3. Want to delete it because it's useless

How long does it take to go back to the list and locate it without mistakes?

### Example URLs

https://github.com/refined-github/refined-github/tree/test/bun

<img width="903" height="126" alt="Image" src="https://github.com/user-attachments/assets/f42a49c1-b4fb-460d-a6ba-7a0ed117f858" />

kidonng:
This could be implemented in a similar fashion to `quick-repo-deletion`:

1. Get current branch
2. Add a button which links to `/branches/all?query=<branch>#rgh-delete-branch`
3. Check for hash and auto click the delete button (or no automation)

francopetruu:
Hey guys! Thought this feature would be cool for my first contribution. Just posting the screenshot here to get the link.

<img width="2478" height="1305" alt="Image" src="https://github.com/user-attachments/assets/ac4196b0-ac0e-4e21-82eb-d665bf2fb3a6" />

fregante:
The hard part here is actually deciding how to tell the user what happened. The button on `/branches` has little to no visual feedback.

An API call + showToast + clearing the page (like for `quick-repo-deletion`) is the most straightforward for this, except we lose the ability to undo the deletion.

Redirecting the user to `/branches` would mean either visually inviting the user to click a 24x24 button or informing that the action has been completed.

fregante:
Found an easy way:

1. API request
2. redirect to https://github.com/refined-github/refined-github/activity?activity_type=branch_deletion

Existing example:

https://github.com/refined-github/refined-github/blob/7054fa2d590e46174a6270a879f5c4558c2ea859/source/features/deep-reblame.tsx#L63-L70

PR #8922, opened 2026-01-31
Add delete current branch button

francopetruu:
## Summary
- add a “Delete branch” button to the branch info bar on /tree pages
- link directly to the filtered branches list with the delete action
- auto-trigger delete button when opening the filtered branches page
- refs #8617

## Test URLs
- https://github.com/refined-github/refined-github/tree/test/bun

## Screenshot
image


fregante:
Thanks for the PR! 

This a lot longer than I'd like, you can use the API to delete the branch. Search `showToast` and `api.v3` because those are the functions that need to be called here.

Also if this is AI, please review the code before marking it as ready.

SunsetTechuila:
I would also suggest checking how this looks at narrow resolutions. GitHub's new UI isn't mobile-friendly at all, which means we should be extra careful not to make things worse

<details>
<summary>GitHub</summary>

image

image


</details>

SunsetTechuila:
You may want to remove the text when there's not enough space

SunsetTechuila:
@francopetruu Hello! Do you plan to finish this PR?

PR #8928, opened 2026-02-01
`select-notifications` - Fix list scroll on mobile

### LINKED ISSUE #8927: `select-notifications` not scrollable on mobile
### Description

The background scrolls, the modal does not

https://github.com/user-attachments/assets/7590be2d-e641-41d1-a5b9-4f75f044b4a1

Notification page

### Extension version

25.12.14

### Browser(s) used

Safari mobile 26.2

### Token ⚠️

SunsetTechuila:
fixes #8927

I assume you prefer this over [refactoring](https://github.com/refined-github/refined-github/issues/8927#issuecomment-3831325344)

## Test URLs

## Screenshot

msedge_txDq4h5TVQ

### DIFF
--- a/source/features/select-notifications.css
+++ b/source/features/select-notifications.css
@@ -16,7 +16,7 @@
 	}
 	@media (max-width: 543px) {
 		.SelectMenu-item {
-			padding: var(--control-large-paddingBlock, 0.625rem);
+			padding: var(--control-large-paddingBlock, 0);
 		}
 	}
 }


fregante:
Can you reduce the padding too? It seems that they're too spaced out on mobile. See the "Switch inbox" modal for comparison 


<table>
<tr>
 <td>



 <td>



</table>

SunsetTechuila:
done

image

source/features/select-notifications.css
```diff
@@ -7,3 +7,16 @@
...
+
+#rgh-select-notifications-form {
+	&,
+	& .SelectMenu-list {
+		/* Remove scrollbars when they are not needed */
+		margin-bottom: 0;
+	}
+	@media (max-width: 543px) {
+		.SelectMenu-item {
+			padding: var(--control-large-paddingBlock, 0.625rem);
```
fregante: We want to use *bad* fallback values so missing vars become immediately visible and are reported. See `fuchsia`

0 would look bad enough without making it look too broken. 50px would lead to an even faster and urgent report (not sure if that's better)

fregante (approved):

source/features/select-notifications.css
SunsetTechuila: Why should we care if that property is removed?

SunsetTechuila: I mean, that's a variable for a different component anyway. And if `0.625rem` looks good now, it's very likely that it will continue to look good until the `SelectMenu` component is removed

fregante: If we don't care, why use the variable at all? If we use it, we want to make sure it's used, otherwise it's just dead code.

Issue #8504, opened 2025-06-28
⚠️ Features broken on new PRs (beta)


fregante:
Here we go: https://github.com/orgs/community/discussions/163932

<img width="336" alt="Image" src="https://github.com/user-attachments/assets/e5364460-9b34-4ef8-b41c-017d5c17a4dd" />

fregante:
Thank you @tyrann0us. I extended the list with all the features containing `isPRFiles` and `hasFiles`

joemckie:
Unsure if this has been reported yet (I couldn't see it mentioned anywhere), but clicking "Filter files" will point to the first file, which makes the input unusable.

fregante:
> clicking "Filter files" will point to the first file

Clicking the field? 

I'm not sure what you're referring to but I think that's unrelated to Refined GitHub, make sure it only happens when Refined GitHub is enabled.

danielkim-moloco:
For toggle-everything-with-alt, "expand all lines" is quite useful for me.

<img width="1304" height="40" alt="Image" src="https://github.com/user-attachments/assets/b908be43-89dd-46ce-b454-b62a3e5eab9e" />

joemckie:
@fregante clicking the input field, yep.

I already checked that 😄 disabling the extension fixes the issue, and enabling it causes it to break again

fregante:
@joemckie  It doesn't happen here. Please open a **new issue** with a gif/video of it so we can make sure we're talking about the same thing.

joemckie:
No worries, I created an issue [here](https://github.com/refined-github/refined-github/issues/8610). Thanks 😄

rasmusbe:
[pr-jump-to-first-non-viewed-file](https://github.com/refined-github/refined-github/blob/main/source/features/pr-jump-to-first-non-viewed-file.tsx) still causes issues with the new PR Files view. When I try to submit a review and add a comment it unfocuses the field all the time

https://github.com/user-attachments/assets/ca3c34fd-1cc8-4347-b34c-4fb85ad81f12

ThoSap:
@rasmusbe I had the same issue today.

fregante:
Thanks for the report. That bug is tracked in https://github.com/refined-github/refined-github/issues/8610

I just disabled the feature via hotfix for now because it's not worth the trouble. Bugfix still welcome

jimeh:
[...]

PR #8930, opened 2026-02-02
`batch-mark-files-as-viewed` - Restore on new Files changed experience

SunsetTechuila:
#8504

## Test URLs

`https://github.com/refined-github/sandbox/pull/55/changes?new_files_changed=false`

`https://github.com/refined-github/sandbox/pull/55/changes?new_files_changed=true`

### DIFF
--- a/source/features/batch-mark-files-as-viewed.tsx
+++ b/source/features/batch-mark-files-as-viewed.tsx
@@ -64,14 +64,12 @@ function batchToggle(event: DelegateEvent<MouseEvent, HTMLFormElement>): void {
 }
 
 function markAsViewedSelector(file: HTMLElement): string {
+	const fileElement = fileSelector.join(',');
+	const viewedToggle = viewedToggleSelector.join(',');
+	const checkedState = isChecked(file) ? `:not(${checkedSelector})` : checkedSelector;
 	// The `hidden` attribute excludes filtered-out files
 	// https://github.com/refined-github/refined-github/issues/7819
-	return [
-		`:is(${fileSelector.join(',')}):not([hidden])`,
-		' ',
-		`:is(${viewedToggleSelector.join(',')})`,
-		isChecked(file) ? `:not(${checkedSelector})` : checkedSelector,
-	].join('');
+	return `:is(${fileElement}):not([hidden]) :is(${viewedToggle})${checkedState}`;
 }
 
 const markAsViewed = clickAll(markAsViewedSelector);


source/features/batch-mark-files-as-viewed.tsx
```diff
@@ -1,42 +1,52 @@
...
 function isChecked(file: HTMLElement): boolean {
-	return $('input.js-reviewed-checkbox', file).checked;
+	const viewedToggle = $(viewedToggleSelector, file);
+
+	return viewedToggle instanceof HTMLInputElement
+		? viewedToggle.checked
+		: elementExists('.octicon-checkbox-fill', viewedToggle);
 }
 
-// A single click is somehow causing two separate trusted `click` events, so it needs to be debounced
```
SunsetTechuila, line 27: > A single click is somehow causing two separate trusted `click` events

Doesn't happen when the checkbox is targeted instead of the label containing it

SunsetTechuila:
image

I don't think `shorten-links` should hide query params

fregante:
Tell that to GitHub:

https://github.com/refined-github/refined-github/pull/8930?hello=world

I would agree but adding support for it would be annoying at the moment. `shorten-repo-url` does preserve the parameters for external URLs but not for local ones. I think there aren't that many situations that need it.

SunsetTechuila:
I tested this against the latest RGH release and I don't think I changed the behavior

SunsetTechuila:
A lot of non-null assertions. We need `unwrap` analogue for JS

source/features/batch-mark-files-as-viewed.tsx
```diff
@@ -1,42 +1,54 @@
...
 import features from '../feature-manager.js';
 import clickAll from '../helpers/click-all.js';
 import showToast from '../github-helpers/toast.js';
 import getItemsBetween from '../helpers/get-items-between.js';
 
+const viewedToggleSelector = [
+	'button[class*="MarkAsViewedButton"]',
+	// Old view
+	'input.js-reviewed-checkbox',
+];
```
fregante:
```suggestion
] as const;
```

on both arrays

source/features/batch-mark-files-as-viewed.tsx
```diff
@@ -46,54 +58,41 @@ const batchToggle = debounceFn((event: DelegateEvent<MouseEvent, HTMLFormElement
...
-
-function markAsViewedSelector(target: HTMLElement): string {
-	const checked = isChecked(target) ? ':not([checked])' : '[checked]';
+function markAsViewedSelector(file: HTMLElement): string {
 	// The `hidden` attribute excludes filtered-out files
 	// https://github.com/refined-github/refined-github/issues/7819
-	return '.file:not([hidden]) .js-reviewed-checkbox' + checked;
+	return `:is(${fileSelector.join(',')}):not([hidden]) `
+		+ `:is(${viewedToggleSelector.join(',')})`
+		+ (isChecked(file) ? `:not(${checkedSelector})` : checkedSelector);
```
fregante: I'd rather see `[1,2,3].join('')` since it wraps better

SunsetTechuila:
> Do you still have access to the old view?

Ofc, I linked both views. Are you asking because you don't have access?

image

> I tested this against the latest RGH release and I don't think I changed the behavior

Also, how would I compare the behavior without access to the old view

fregante:
I'm computerless for at least the next week, I've been reviewing from the GitHub app 😅

mwarkentin:
Hoping this can land soon, I installed the extension in hopes of preventing carpal tunnel.. I've got a bunch of PRs that have hundreds of files changed like this.. 😓 

image

PR #8933, opened 2026-02-03
`suggest-commit-title-limit` - Restore feature

SunsetTechuila:
#8123

#8946

## Test URLs

https://github.com/refined-github/sandbox/pull/8

`https://github.com/refined-github/refined-github/compare/main...karlhorky:refined-github:test/bun?expand=true`

https://github.com/refined-github/refined-github/edit/main/.nvmrc

## Screenshot

image

image

image

image

### DIFF
--- a/source/features/suggest-commit-title-limit.tsx
+++ b/source/features/suggest-commit-title-limit.tsx
@@ -39,9 +39,9 @@ async function init(signal: AbortSignal): Promise<void> {
 	delegate([
 		'[class^="prc-PageLayout-Header"] input', // `isPR`
 		'input[name="pull_request[title]"]', // `isCompare`
-		'#pull_request_title', // Old `isCompare` - TODO: Remove after August 2026
-		'#issue_title', // Old `isPR` view - TODO: Remove after July 2026
-	], 'input', async ({delegateTarget}) => validatePrTitle(delegateTarget as HTMLInputElement), {signal, passive: true});
+		'input#pull_request_title', // Old `isCompare` - TODO: Remove after August 2026
+		'input#issue_title', // Old `isPR` view - TODO: Remove after July 2026
+	], 'input', async ({delegateTarget}) => validatePrTitle(delegateTarget), {signal, passive: true});
 	// `isCompare` - input is re-rendered when previously entered title is restored
 	observe('input[name="pull_request[title]"]', validatePrTitle, {signal});
 


source/features/suggest-commit-title-limit.css
```diff
@@ -3,7 +3,12 @@
 }
 
 /* Limit width of commit title to 72 characters */
-.rgh-suggest-commit-title-limit :is(#merge_title_field, #commit-summary-input) {
+/* biome-ignore format: bug */
```
fregante: What's the bug?

SunsetTechuila: image

fregante: That's not worth a comment. Prettier doesn't support comments in selectors at all btw https://github.com/prettier/prettier/issues/15356

source/features/suggest-commit-title-limit.css
```diff
@@ -3,7 +3,13 @@
 }
 
 /* Limit width of commit title to 72 characters */
-.rgh-suggest-commit-title-limit :is(#merge_title_field, #commit-summary-input) {
+.rgh-suggest-commit-title-limit
+:is(
+	#commit-summary-input
+		/* TODO: Remove next selector after July 2026 */
+		#commit-message-input,
```
fregante: Why is this `#id #id` selector needed? Do they have more than one message?

SunsetTechuila: This is a mistake

SunsetTechuila:
https://github.com/refined-github/refined-github/pull/8957#issuecomment-3885242914

SunsetTechuila:
@fregante ready to be reviewed

fregante (approved):

SunsetTechuila:
the last change 🙏

fregante:
Please don't ask me to review the same PR 4 times

SunsetTechuila:
But I asked only twice

fregante:
If the PR is "ready for review" that's already asking for it to be reviewed. You pushed commits after my review, should I let the PR simmer for a few more weeks instead? Because that's what I've been doing when these "mismatched expectations" happen.

mschoettle:
The change to the CSS rule caused a regression. Attempt to fix it: #9035

PR #8939, opened 2026-02-04
Fix PR state getting

SunsetTechuila:
affects a lot of features


SunsetTechuila:
Perhaps this doesn't have to be a separate PR. Closing in favor of #8933

PR #8878, opened 2026-01-18
`linkify-code`, `show-whitespace` - Restore on React-based code views and diffs

SunsetTechuila:
fixes #6336

fixes #7843

## Test URLs

https://github.com/refined-github/refined-github/blob/be756e68dd4bf24d53e802e1fec126506979fa80/test/linkify-urls-in-code.js

https://github.com/refined-github/refined-github/pull/546/changes

## Screenshot

image

PR #8940, opened 2026-02-04
`linkify-symbolic-links` - Restore feature

SunsetTechuila:
follow-up to #8878

## Test URLs

https://github.com/wmluke/angular-flash/blob/0.1.14/app/components

## Screenshot

image


fregante (approved):
🙌

SunsetTechuila:
The reminder works

PR #8944, opened 2026-02-04
`ci-link` - Restore feature

### LINKED ISSUE #8943: `ci-link` is broken
### Description

`ci-link` feature stopped working due to now non-existing anchor on a GitHub repo page.

In the code it is looking for now non-existing parts:
```tsx
	await expectToken();

	observe([
		// Desktop
		'.AppHeader-context-item:not([data-hovercard-type])',

		// Mobile. `> *:first-child` avoids finding our own element
		'.AppHeader-context-compact-mainItem > span:first-child',
	], add, {signal});
}
```

- open https://github.com/refined-github/refined-github with `ci-link` feature switched on
- or see this picture:
  Image

### Extension version

 26.2.2

### Browser(s) used

firefox, chrome

### Token ⚠️

blzsaa:
Update the anchor according to the UI changes on GitHub.

closes #8943

## Test URLs

## Screenshot
Without this:
image
With this:
image

### DIFF
--- a/source/features/ci-link.tsx
+++ b/source/features/ci-link.tsx
@@ -55,7 +55,7 @@ async function init(signal: AbortSignal): Promise<void> {
 	await expectToken();
 
 	observe([
-		'div[data-testid="top-nav-center"] li[class^="prc-Breadcrumbs-ItemWrapper"]:last-child a[class*="prc-Breadcrumbs-Item"]',
+		'div[data-testid="top-nav-center"] li:last-child > a[class*="prc-Breadcrumbs-Item"]',
 		// TODO: Remove after July 2026
 		// Desktop
 		'.AppHeader-context-item:not([data-hovercard-type])',
--- a/source/features/repo-header-info.tsx
+++ b/source/features/repo-header-info.tsx
@@ -82,7 +82,7 @@ async function add(repoLink: HTMLAnchorElement): Promise<void> {
 async function init(signal: AbortSignal): Promise<void> {
 	await expectToken();
 	observe([
-		'div[data-testid="top-nav-center"] li[class^="prc-Breadcrumbs-ItemWrapper"]:last-child a[class*="prc-Breadcrumbs-Item"]',
+		'div[data-testid="top-nav-center"] li:last-child > a[class*="prc-Breadcrumbs-Item"]',
 		'.AppHeader-context-full [role="listitem"]:last-child a.AppHeader-context-item', // TODO: Drop after May 2026
 	], add, {signal});
 }


source/features/ci-link.tsx
```diff
@@ -50,13 +50,7 @@ async function add(anchor: HTMLElement): Promise<void> {
...
 	await expectToken();
 
-	observe([
-		// Desktop
-		'.AppHeader-context-item:not([data-hovercard-type])',
-
-		// Mobile. `> *:first-child` avoids finding our own element
-		'.AppHeader-context-compact-mainItem > span:first-child',
-	], add, {signal});
+	observe('[itemprop="name"]', add, {signal});
```
SunsetTechuila: Thank you for the contribution.

Please leave the old selectors for compatibility. Also, the link should appear in the header.

```suggestion
	observe([
		'div[data-testid="top-nav-center"] li[class^="prc-Breadcrumbs-ItemWrapper"]:last-child > a',
		// TODO: Remove after July 2026
		'.AppHeader-context-item:not([data-hovercard-type])', // Desktop
		'.AppHeader-context-compact-mainItem > span:first-child', // Mobile. `> *:first-child` avoids finding our own element
	], add, {signal});
```

SunsetTechuila (changes requested):

source/features/ci-link.tsx
blzsaa: I tried with `div[data-testid="top-nav-center"] li[class^="prc-Breadcrumbs-ItemWrapper"]:last-child > a` but it rendered `ci-link` twice.

As a solution I added `header.GlobalNav [data-testid="top-nav-center"] ol > li:last-child a:first-child` instead, from `repo-header-info.tsx`

source/features/ci-link.css
```diff
@@ -1,3 +1,7 @@
+.rgh-ci-link {
+	align-self: center;
```
blzsaa: Without this:

Image

source/features/ci-link.tsx
SunsetTechuila: > I tried with `div[data-testid="top-nav-center"] li[class^="prc-Breadcrumbs-ItemWrapper"]:last-child > a` but it rendered [`ci-link`](https://github.com/refined-github/refined-github/blob/main/source/features/ci-link.tsx?rgh-link-date=2026-02-05T08%3A31%3A58Z) twice.

Are you sure you tried that exact selector and not one of the ones I initially suggested? I'm not having any issues with it.

> `header.GlobalNav [data-testid="top-nav-center"] ol > li:last-child a:first-child`

I don't like that selector; it's somewhat loose and ambiguous. Also, the `header.GlobalNav` part is redundant. However, @fregante may have a different opinion, given that he approved it previously.

source/features/ci-link.tsx
SunsetTechuila: I don't know how well this would work with the old navbar, but it works better than `align-self: center` with the current one

```suggestion
		<span className="rgh-ci-link ml-1 d-none d-sm-flex flex-items-center flex-justify-center" title="CI status of latest commit">
```

| Before | image |
|--------|--------|
| After | image |

source/features/ci-link.tsx
SunsetTechuila: > I don't like that selector; it's somewhat loose and ambiguous

And causes this:

image

SunsetTechuila:
Please do not force-push without good reason. This creates noise and makes review process harder

source/features/ci-link.tsx
blzsaa: > Are you sure you tried that exact selector and not one of the ones I initially suggested? I'm not having any issues with it.

After some investigation I found out that if `repo-header-info` is switched on `document.querySelectorAll('div[data-testid="top-nav-center"] li[class^="prc-Breadcrumbs-ItemWrapper"]:last-child > a').length` will return 2, otherwise 1.

As a solution to this problem I can change the selector to `'div[data-testid="top-nav-center"] li[class^="prc-Breadcrumbs-ItemWrapper"]:last-child > a:first-child'`, what is your opinion about that?

SunsetTechuila: Sounds good

source/features/ci-link.tsx
SunsetTechuila: My initial suggestion doesn't work with `repo-header-info` enabled. Updated it to copy classes from `repo-header-info` instead

source/features/ci-link.css
SunsetTechuila: Part of https://github.com/refined-github/refined-github/pull/8944/changes#r2768107399

```suggestion
.rgh-ci-link .commit-build-statuses {
	font-size: 0;
```

SunsetTechuila:
I decided it would be faster to commit the changes myself

source/features/ci-link.tsx
```diff
@@ -51,6 +55,8 @@ async function init(signal: AbortSignal): Promise<void> {
 	await expectToken();
 
 	observe([
+		'div[data-testid="top-nav-center"] li[class^="prc-Breadcrumbs-ItemWrapper"]:last-child a[class*="prc-Breadcrumbs-Item"]',
```
fregante: How many breadcrumbs are on the page? The first part of the selector seems redundant. 

Maybe even the second one. How about `li:last-child > a[class*="prc-Breadcrumbs-Item"]`. Note the additional `>` since `li` could be any parent otherwise

SunsetTechuila: > How many breadcrumbs are on the page? 

Technically, these are breadcrumbs too, but yes, you can remove the `div[data-testid="top-nav-center"]` part - I just decided to play it safe

Image

> `li:last-child > a[class*="prc-Breadcrumbs-Item"]`

I like that

SunsetTechuila:
msedge_pDZ5mkc3Yg

source/features/ci-link.tsx
```diff
@@ -55,7 +55,7 @@ async function init(signal: AbortSignal): Promise<void> {
 	await expectToken();
 
 	observe([
-		'div[data-testid="top-nav-center"] li[class^="prc-Breadcrumbs-ItemWrapper"]:last-child a[class*="prc-Breadcrumbs-Item"]',
+		'div[data-testid="top-nav-center"] li:last-child > a[class*="prc-Breadcrumbs-Item"]',
```
fregante, line 58: div[data-testid="top-nav-center"] still needed?

SunsetTechuila: I think I answered in https://github.com/refined-github/refined-github/pull/8944#discussion_r2770249981

PR #8958, opened 2026-02-08
Fix parsing of head branch of deleted repos

SunsetTechuila:
This is not actually enough to fix the issue. The value of `nextElementSibling!.textContent` is `""`, since it's a copy button, not a tooltip - try executing `getBranches().head` in the DevTools console. This applies to both old and new headers. What should we do?

## Test URLs

~~https://github.com/refined-github/refined-github/pull/3227/checks (Old header)~~ Updated

https://github.com/refined-github/refined-github/pull/3227

### DIFF
--- a/source/github-helpers/pr-branches.ts
+++ b/source/github-helpers/pr-branches.ts
@@ -54,7 +54,7 @@ export function parseReferenceRaw(absolute: string, relative: string): PrReferen
 function parseReference(referenceElement: HTMLElement): PrReference {
 	const {title, textContent, nextElementSibling} = referenceElement;
 
-	// In the old React version, we have a `title` attribute but it's used to mark deleted repos instead
+	// In the React version, we have a `title` attribute but it's used to mark deleted repos instead
 	return title && title !== 'This repository has been deleted'
 		? parseReferenceRaw(title, textContent.trim()) // TODO: Remove in June 2026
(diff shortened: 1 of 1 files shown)


SunsetTechuila:
@fregante this one needs your attention

fregante:
Not sure what you're asking. If it fixes something, merge it. If an alternative fix is needed, find it 😅 or leave the issue open. 

> The value of nextElementSibling!.textContent is ""

therefore an alternative should be found, just like for every other piece of code that no longer works.

SunsetTechuila:
> therefore an alternative should be found, just like for every other piece of code that no longer works.

I'm asking if you have any suggestions

fregante:
Not at this moment 😇

SunsetTechuila:
Then I'll create an issue instead

source/github-helpers/pr-branches.ts
```diff
@@ -54,7 +54,7 @@ export function parseReferenceRaw(absolute: string, relative: string): PrReferen
 function parseReference(referenceElement: HTMLElement): PrReference {
 	const {title, textContent, nextElementSibling} = referenceElement;
 
-	// In the old React version, we have a `title` attribute but it's used to mark deleted repos instead
+	// In the React version, we have a `title` attribute but it's used to mark deleted repos instead
```
SunsetTechuila, line 57: test comment

PR #8965, opened 2026-02-10
`rerun-workflow` - New feature

### LINKED ISSUE #8964: Add a keyboard shortcut for re-running a failed GitHub Actions run
### Description

Our CI is unfortunately somewhat flaky. While we are working on fixing it, it is a common thing for us to need to re-run a particular CI execution. It would be really nice if there was a way to do this without having to reach for my mouse.

My proposal would be to bind keyboard shortcuts to "Re-run all jobs", "Re-run failed jobs", and, when focusing a single job, "Re-run this job". This way the entire re-running process can be done from the keyboard:

- Go to notifications (<kbd>G</kbd> <kbd>N</kbd>)
- Select the "run failed" notifications (<kbd>J</kbd> and <kbd>K</kbd>)
- Open the failed run (<kbd>O</kbd> or <kbd>Enter</kbd>)
- Re-run the job (missing keybind, that's where this issue comes in)
- Mark the notification as done (<kbd>E</kbd>)

### Example URLs

https://github.com/refined-github/refined-github/actions/runs/21794152097?pr=8957

costajohnt:
Closes #8964

Adds a `R+F` keyboard shortcut on Actions run pages to re-run failed jobs.

Per @fregante's guidance, only "Re-run failed jobs" gets a shortcut — no shortcut for "Re-run all jobs" to avoid conflicts.

## How it works

- Uses `registerHotkey` to bind `R+F` on `isActionRun` pages
- Finds the "Re-run failed jobs" button in `.PageHeader-actions` and clicks it
- Falls back gracefully if the button isn't found (e.g., no failed jobs)

## Test URL

https://github.com/refined-github/refined-github/actions/runs/16143473286?pr=8544

## Before/After
composite

### DIFF
--- a/build/__snapshots__/features-meta.json
+++ b/build/__snapshots__/features-meta.json
@@ -764,7 +764,7 @@
 	{
 		"id": "rerun-workflow",
 		"description": "Unwraps the \"Re-run jobs\" dropdown into individual buttons and adds a keyboard shortcut to re-run failed jobs: <kbd>r</kbd> <kbd>f</kbd>.",
-		"screenshot": null
+		"screenshot": "https://github.com/user-attachments/assets/727b389d-c4a3-4d21-9932-db474d233cf3"
 	},
 	{
 		"id": "restore-file",


source/features/rerun-workflow.tsx
```diff
@@ -0,0 +1,36 @@
...
+import * as pageDetect from 'github-url-detection';
+
+import features from '../feature-manager.js';
+import {registerHotkey} from '../github-helpers/hotkey.js';
+
+function rerunFailedJobs(): void {
+	// The `unwrap-unnecessary-dropdowns` feature exposes re-run buttons directly
+	// in the page header. If disabled, fall back to buttons inside the dropdown menu.
+	const button = $$('.PageHeader-actions button')
+		.find(button => button.textContent?.trim() === 'Re-run failed jobs');
```
SunsetTechuila:
```suggestion
	const button = $optional('button[data-show-dialog-id="rerun-dialog-failed"]');
```

source/features/rerun-workflow.tsx
```diff
@@ -0,0 +1,36 @@
+import {$$} from 'select-dom/strict.js';
```
SunsetTechuila:
```suggestion
import {$optional} from 'select-dom/strict.js';
```

https://github.com/refined-github/refined-github/pull/8965/changes#r2785690134

source/features/rerun-workflow.tsx
```diff
@@ -0,0 +1,32 @@
...
+import features from '../feature-manager.js';
+import {registerHotkey} from '../github-helpers/hotkey.js';
+
+function rerunFailedJobs(): void {
+	const button = $optional('button[data-show-dialog-id="rerun-dialog-failed"]');
+	button?.click();
+}
+
+function init(signal: AbortSignal): void {
+	registerHotkey('shift r', rerunFailedJobs, {signal});
```
fregante: `r f` might be a good shortcut. But:

- check if the element already has any shortcuts attached
- verify that 'r f' isn't already used on that page
- also use `observe()` to mention the shortcut in the button's tooltip (either use the 'title' attribute or check how to use `tooltipped`)

source/features/rerun-workflow.tsx
```diff
@@ -0,0 +1,32 @@
+import {$optional} from 'select-dom/strict.js';
+import * as pageDetect from 'github-url-detection';
+
+import features from '../feature-manager.js';
+import {registerHotkey} from '../github-helpers/hotkey.js';
+
+function rerunFailedJobs(): void {
+	const button = $optional('button[data-show-dialog-id="rerun-dialog-failed"]');
+	button?.click();
```
fregante: We do want this to throw if missing, so don't use $optional. 

The logic is that if the shortcut was pressed, we shouldn't just ignore it, we should tell the user (via the console error) that the button it was looking for is missing.

fregante:
Thanks for the PR! Please don't force push because it makes it difficult to see what changed between reviews

costajohnt:
Thanks for the review! Noted on the force push — I'll use regular commits going forward.

Changes in the latest commit:
- Switched to throwing `$` so a missing button surfaces in the console
- Changed shortcut to `r f`
- Added `observe()` to set a tooltip on the rerun button with the shortcut hint
- Extracted the selector to a shared `const` since it's now used in two places

source/features/rerun-workflow.tsx
```diff
@@ -0,0 +1,39 @@
...
+import observe from '../helpers/selector-observer.js';
+
+const rerunButtonSelector = 'button[data-show-dialog-id="rerun-dialog-failed"]';
+
+function rerunFailedJobs(): void {
+	$(rerunButtonSelector).click();
+}
+
+function addShortcutHint(button: HTMLButtonElement): void {
+	button.setAttribute('title', 'Re-run failed jobs (r f)');
```
SunsetTechuila: I suggest to use a `tool-tip` element like this one:

image

SunsetTechuila: This would make it possible to use `kbd` elements like in that tooltip 

image

SunsetTechuila: https://primer.style/product/getting-started/rails/components/tooltip/

costajohnt:
Thanks! Good question, I looked into it.

I think they're better as separate features: the dropdown unwrapping is a visual simplification, while the shortcut is about keyboard access. A user might want one without the other, and keeping them separate preserves that choice. The shortcut also works independently since `data-show-dialog-id` targets the native button regardless of whether the dropdown has been unwrapped.

That said, there is a coordination detail worth addressing: the tooltip hint timing with the cloned buttons from `unwrap-unnecessary-dropdowns`. I'll make sure the shortcut hint is applied correctly regardless of feature ordering.

I'll also address @SunsetTechuila's suggestion about using `<tool-tip>` instead of a plain `title` attribute.

Happy to consolidate if you'd prefer it though, just let me know.

fregante:
> there is a coordination detail

That's one reason why they should be one feature

costajohnt:
Consolidated the dropdown unwrapping into this feature as suggested.

Changes in the latest commit:
- Moved `replaceRerunDropdown` from `unwrap-unnecessary-dropdowns` into `rerun-workflow` — both now live in one feature since they operate on the same button
- Replaced the `title` attribute with a `<tool-tip>` element using `<kbd>` for the shortcut hint (per @SunsetTechuila's suggestion)
- Used `$optional` for the guard clause since the observer matches all `action-menu` elements, not just the "Re-run jobs" one
- Clone buttons before restyling to avoid mutating the visible originals
- Added `tool-tip` to JSX `IntrinsicElements` in `globals.d.ts`
- `unwrap-unnecessary-dropdowns` is now notifications-only

source/features/rerun-workflow.tsx
```diff
@@ -0,0 +1,64 @@
...
+		const clone = button.cloneNode(true);
+		clone.className = 'Button--secondary Button--medium Button';
+		clone.firstElementChild!.className = 'Button-label';
+		container.append(clone);
+
+		if (clone.dataset.showDialogId === 'rerun-dialog-failed') {
+			clone.after(
+				<tool-tip data-direction="s" data-type="description" role="tooltip">
+					Re-run failed jobs <kbd>r</kbd> <kbd>f</kbd>
+				</tool-tip>,
```
SunsetTechuila: 1. This doesn't work
2. GitHub uses tooltips only for icon buttons
3. The tooltip will look strange in either of two cases: if it contains only hotkeys or hotkeys and the same text as the button
4. `kbd` elements inside `tool-tip` aren't styled, we have to clone and edit the notifications button tooltip 

My suggestion is to remove the tooltip

SunsetTechuila: > 4\. `kbd` elements inside `tool-tip` aren't styled, we have to clone and edit the notifications button tooltip

I've already written the code for that, though

build/__snapshots__/features-meta.json
```diff
@@ -765,6 +765,11 @@
 		"description": "Enables the File Finder keyboard shortcut (<kbd>t</kbd>) on entire repository.",
 		"screenshot": null
 	},
+	{
+		"id": "rerun-workflow",
+		"description": "Unwraps the \"Re-run jobs\" dropdown into individual buttons and adds a keyboard shortcut to re-run failed jobs: <kbd>r</kbd> <kbd>f</kbd>.",
+		"screenshot": null
```
fregante: This feature can now have a screenshot. Also make sure that the other "unwrap" feature doesn't mention this dropdown anywhere, including the screenshot

SunsetTechuila: @costajohnt can you do that part please?

costajohnt: Here's the feature in action:
Screenshot 2026-02-13 at 6 32 04 PM

SunsetTechuila: I meant a preview for the feature like this:

image

https://github.com/refined-github/refined-github/wiki/Contributing#screenshot

fregante: We copy paste those exact before/after labels from an existing screenshot

costajohnt:
Working on getting the screenshot in the before/after format. I'll add it to the PR description once I have the proper before/after comparison captured following the wiki's contributing guidelines.

Could @SunsetTechuila share the image generation approach used for existing screenshots (e.g., the tool or template used for the "Before Refined GitHub / After Refined GitHub" overlays)?  That'll help me match the exact style.

build/__snapshots__/features-meta.json
costajohnt: Added the before/after screenshot to the PR description. Will update `features-meta.json` with the screenshot URL in the next commit, and check that the `unwrap-unnecessary-dropdowns` screenshot no longer includes the re-run dropdown.

SunsetTechuila:
Okay, now I'm sure there's no human in the loop and what you "autopiloted" is wasting people's time. Great. I never understood why people do this and I never will. I'm not engaging with these "autopiloted" PRs any further

costajohnt:
> Okay, now I'm sure there's no human in the loop and what you "autopiloted" is wasting people's time. Great. I never understood why people do this and I never will. I'm not engaging with these "autopiloted" PRs any further

What’s the problem? I’ve added the screenshot you asked for to the description. Looks like I have some CI to fix, was going to get to it. Feel free to close the PR if you’re not interested in engaging.

source/features/rerun-workflow.tsx
SunsetTechuila: Some text fields have keyboard shortcut icons 

image

PR #8970, opened 2026-02-11
`show-names` - Restore for some elements & fix spacing

mschoettle:
There is margin missing on the inserted name element for timeline items like "User added some commits" or "User force-pushed".

It is only missing on non-comments and those without `.markdown-title` (since they have their own CSS rule to change it to a flex container with gap).

## Test URLs

- This very PR :smile:
- https://github.com/refined-github/refined-github/issues/8504
- https://github.com/orgs/refined-github/dashboard
- https://github.com/feed (depends on https://github.com/refined-github/github-url-detection/pull/235)

## Screenshot

| Before | image |
|--------|--------|
| **After** | image |

| Before | image |
|--------|--------|
| **After** | image |

| Before | image |
|--------|--------|
| **After** | image |

| Before | image |
|--------|--------|
| **After** | image |

### DIFF
--- a/source/features/show-names.tsx
+++ b/source/features/show-names.tsx
@@ -46,7 +46,7 @@ function createElement(element: HTMLAnchorElement, fullName: string): JSX.Elemen
 	} else if (
 		element.matches(
 			// Issue event:
-			//  - https://github.com/refined-github/sandbox/issues/3#event-1142607402
+			//  - https://github.com/refined-github/sandbox/issues/3#event-5474574930
 			//  - https://github.com/refined-github/sandbox/issues/3#event-10265481248
 			// Username in issue events already has left margin
 			// Rename events use `gap` for spacing


source/features/show-names.tsx
```diff
@@ -36,6 +36,12 @@ function createElement(element: HTMLAnchorElement, fullName: string): JSX.Elemen
 		nameElement.classList.add('ml-1');
 	}
 
+	// If it's a comment like "User added some commits" we need to add some margin
+	const {parentElement} = element;
+	if (parentElement && parentElement.tagName !== 'STRONG') {
```
SunsetTechuila:
```suggestion
	if (element.closest('.TimelineItem-body')) {
```

mschoettle: Thanks for your comment. That helped me see that this condition I used was indeed correct (although incomplete) :smile: Otherwise there is extra margin for comments (like this one).

It should be good now.

fregante (approved):

source/features/show-names.tsx
```diff
@@ -36,6 +36,13 @@ function createElement(element: HTMLAnchorElement, fullName: string): JSX.Elemen
 		nameElement.classList.add('ml-1');
 	}
 
+	// If it's a comment like "User added some commits" we need to add some margin
+	// Ensure it does not aply to regular comment author elements and those timeline items with del.markdown-title (already have gaps via CSS rule)
+	const {parentElement} = element;
+	if (parentElement!.tagName !== 'STRONG' && !parentElement?.querySelector('del.markdown-title')) {
```
SunsetTechuila: It's `readable-title-change-events` that adds a gap

fregante: What's the suggested change?

mschoettle: I reworded it slightly and mentioned `readable-title-change-events`. Is this better, @SunsetTechuila?

SunsetTechuila: > What's the suggested change?

I'll just push the changes I think should be made, and then it's up to you to decide whether to accept them or not

mschoettle: I am happy to explore other options if you think there is a better way to do this.

I tried to find a way to do it with CSS like `readable-title-change-events` does but could not find one. It does not work for all cases (those that have a floating button on the right).

fregante:
If would be good to include a comment with the test URL in each block, plus ideally a description of what to look for. Long conditions like these can become hard to parse and test later on.

mschoettle:
I just tried out the latest changes and can confirm that it works. Thanks.

SunsetTechuila:
`.feed-item-content *` and `[data-testid="issue-body-header-author"]` is pretty self-explanatory

SunsetTechuila:
@mschoettle Thank you

PR #8972, opened 2026-02-12
`improve-shortcut-help` - Make dialog taller

SunsetTechuila:
## Test URLs

https://github.com/refined-github/refined-github/pull/8971

## Screenshot

| Before | After |
|--------|--------|
| image | image | 

<details><summary></summary>
<p>

image

</p>
</details>

### DIFF
--- a/source/features/improve-shortcut-help.css
+++ b/source/features/improve-shortcut-help.css
@@ -1,5 +1,5 @@
 html:not([rgh-OFF-improve-shortcut-help]) {
-	#__primerPortalRoot__ div[class*='ShortcutsDialogRoot'] {
+	div[class^='ShortcutsDialog-module__ShortcutsDialogRoot'] {
 		height: auto;
 	}
 }


source/features/improve-shortcut-help.tsx
```diff
@@ -8,6 +8,7 @@ import features from '../feature-manager.js';
 import {isEditable} from '../helpers/dom-utils.js';
 import {shortcutMap} from '../helpers/feature-helpers.js';
 import observe from '../helpers/selector-observer.js';
+import './improve-shortcut-help.css';
```
fregante: CSS as first import

fregante (approved):
do not long the cat

source/features/improve-shortcut-help.tsx
SunsetTechuila: ESLint should throw an error about that

fregante: Biome can do that actually, do you want to give it a go? The entry point for the content script should not be sorted though

PR #8973, opened 2026-02-13
Improve error message for org-blocked PAT tokens

### LINKED ISSUE #8942: Improve API error message on organizations that block PAT token
See #8941. It should just say "The organization requires a specific type of token. Read more" pointing to the last paragraph of the security page (via selection anchor)

Image

I think we have code detecting this already but I guess it failed.

costajohnt:
Closes #8942

When an organization blocks classic PATs, the GitHub API returns "Resource not accessible by personal access token". Previously this fell through to the generic "Unable to fetch. Ensure that your token has access to this repo." message, which doesn't explain the actual problem.

This adds a specific case in `getError()` that shows a clear message directing users to the Security wiki page about token types. The "Read more" link is rendered as a clickable link in the toast notification.

Additionally, the toast system now auto-linkifies both markdown-style `[text](url)` links and bare URLs, so any error message containing a URL will show it as a clickable link.

## Test URLs

Any org repo page where the org has blocked classic PATs (e.g. a feature like `quick-review` that makes API calls).

## Screenshot

BEFORE
before-message

AFTER
Screenshot 2026-03-05 at 7 46 37 PM

### DIFF
--- a/eslint-suppressions.json
+++ b/eslint-suppressions.json
@@ -69,11 +69,6 @@
       "count": 2
     }
   },
-  "source/features/pagination-hotkey.tsx": {
-    "@typescript-eslint/naming-convention": {
-      "count": 2
-    }
-  },
   "source/features/pr-first-commit-title.tsx": {
     "@typescript-eslint/naming-convention": {
       "count": 1
@@ -136,7 +131,7 @@
   },
   "source/github-helpers/api.tsx": {
     "@typescript-eslint/naming-convention": {
-      "count": 8
+      "count": 6
     }
   },
   "source/github-helpers/dom-formatters.tsx": {


source/github-helpers/api.ts
```diff
@@ -271,6 +271,13 @@ async function getError(apiResponse: JsonObject): Promise<RefinedGitHubAPIError>
 		);
 	}
 
+	if ((apiResponse.message as string)?.includes('Resource not accessible by personal access token')) {
+		return new RefinedGitHubAPIError(
+			'The organization requires a specific type of token.',
```
SunsetTechuila: I don't think this is a clear message. Please consider changing it to something like "Your organization forbids usage of personal access tokens. Refined GitHub requires that specific type of token"

SunsetTechuila:
The "T" in "PAT" stands for "token", so "PAT token" is "PersonalAccessToken token" 🤓

source/github-helpers/api.ts
fregante: That's not accurate, Refined GitHub supports oauth tokens too, which is what the "read more" points to.

source/github-helpers/api.ts
```diff
@@ -271,6 +271,13 @@ async function getError(apiResponse: JsonObject): Promise<RefinedGitHubAPIError>
...
+			'Read more: https://github.com/refined-github/refined-github/wiki/Security#token',
```
fregante: "Read more" should be clickable instead

costajohnt:
Updated — the "Read more" text is now rendered as a clickable link in the toast. The toast system auto-linkifies markdown-style `[text](url)` links and bare URLs.

I'll paste a real screenshot into the PR description shortly.

source/github-helpers/toast.tsx
```diff
@@ -43,8 +43,28 @@ export default async function showToast(
 			</span>
 		</div>
 	);
+	let lastRawMessage = message;
 	const updateToast = (message: string): void => {
```
fregante: Instead of accepting a string and then linkifying it at runtime, it should accept `string | JSX.Element`. This way you can pass `<>fragment with <a href="link">read more</a></>`

source/github-helpers/api.ts
SunsetTechuila: The message is still unclear, imo. I wouldn't have understood what it was about until I read the wiki

fregante: I don't want to add a whole paragraph explaining whose fault it is and why in a toast. The message as set by this PR is perfectly accurate, even if incomplete. 

The only thing I'd change is to clarify that it *can* be fixed, like "Fix…" instead of "Read more", and a dedicated section in the wiki.

SunsetTechuila: > I don't want to add a whole paragraph explaining whose fault it is and why in a toast.

"Your organization forbids usage of personal access tokens" would suffice

fregante: If you're just translating the error yes, but from the point of view of a user who clicked a button it's just as incomplete as what I suggested. If anything, it's even less clear that there's an alternative (i.e. the oauth token)

SunsetTechuila: Can we at least agree to replace "The organization" with "Your organization", then?

costajohnt:
Addressed the feedback:

- Changed "The organization" → "Your organization"
- Changed "Read more" → "Fix…" with a clickable link
- Instead of parsing markdown links at runtime, `showToast` now accepts `string | JSX.Element` via a new `ToastMessage` type. The `RefinedGitHubAPIError` class has a `richMessage` property that carries JSX directly — the toast renders it as-is when present, falling back to `error.message` for plain string errors.
- Removed the regex-based link parser from `updateToast` entirely
- Renamed `api.ts` → `api.tsx` for JSX support

source/github-helpers/api.tsx
```diff
@@ -53,6 +54,7 @@ const escapeKey = (...keys: Array<string | number>): string => '_' + String(keys
 
 export class RefinedGitHubAPIError extends Error {
 	response: AnyObject = {};
+	richMessage?: JSX.Element;
```
fregante, line 57: If possible, I'd rather extend the 'message' prop instead of adding a new one. But not required to merge this PR

fregante (approved):
As one last check before merging, can you provide a real screenshot of the toast with this new message using the latest code?

costajohnt:
> As one last check before merging, can you provide a real screenshot of the toast with this new message using the latest code?

Added before and after screenshots to the description

Issue #8946, opened 2026-02-05
Features broken on updated PR page


SunsetTechuila:
### Description

So, the header was updated and is now React-based, which has broken a bunch of features.

Non-exhaustive list:

- [ ] `clean-conversation-headers`
- [ ] `conversation-activity-filter`
- [ ] `jump-to-conversation-close-event`
- [ ] `sync-pr-commit-title`
- [ ] `clear-pr-merge-commit-message`
- [ ] `suggest-commit-title-limit`
- [ ] `locked-issue`

As of right now, I don’t see any other issues reported about these except mine, so this change might be rolling out slowly, which is good for us.

### How to replicate the issue + URL

https://github.com/refined-github/refined-github/pull/2611

### Extension version

26.1.24

### Browser(s) used

Microsoft Edge 145.0.3800.22 (Official build) beta (64-bit)

### Token ⚠️

- [x] I set a token in the options page. I understand that some features require a token to work.

SunsetTechuila:
I don't have the React header anymore. Good opportunity to test all opened PRs on the old page

SunsetTechuila:
It's back again

SunsetTechuila:
`awaitDomReady: true` doesn't always work correctly on issue pages 🤔

fregante:
How so? The feature awaits the DOM ready event or whichever ajaxed "ready" event we listen to. What you're most likely seeing is just due to deferred loading, which `awaitDomReady` never covered or aimed to cover.

SunsetTechuila:
Yeah, the problem is that GitHub stopped using SSR for issues

fregante:
Once those three are covered, this issue can be closed. I'd suggest creating the 3 separate issues and closing it early. 

Any new feature we find broken can be reported independently. We're likely going to have another similar issue once they publish the new PR pages in full.

fregante:
3 issues extracted, new issues can be opened for each new broken feature discovered. 

Thanks for working on this @SunsetTechuila!

PR #8979, opened 2026-02-14
`sticky-sidebar` - Restore feature

SunsetTechuila:
restores feature on PR (#8946) and repo root pages + cleanup

## Test URLs

https://github.com/refined-github/refined-github/pull/755

## Screenshot

image

image

### DIFF
--- a/source/features/sticky-sidebar.css
+++ b/source/features/sticky-sidebar.css
@@ -12,7 +12,9 @@ Exclusively use simple sums and use `0px` instead of `0`
 }
 
 /* `isRepoRoot` */
-div[class^="prc-PageLayout-Pane"]:has(> rails-partial[data-partial-name="codeViewRepoRoute.Sidebar"]),
+div[class^='prc-PageLayout-Pane']:has(
+		> rails-partial[data-partial-name='codeViewRepoRoute.Sidebar']
+	),
 /* Old `isRepoRoot` - Remove after August 2026 */
 .Layout-sidebar .BorderGrid {
 	--rgh-sticky-sidebar-offset: 0;


source/features/sticky-sidebar.css
```diff
@@ -27,26 +20,15 @@ Exclusively use simple sums and use `0px` instead of `0`
 	/* It needs to be on the same element as above for calc() to work */
 	--rgh-sticky-notification-header-height: 69px;
```
SunsetTechuila, line 25: We can use `--base-sticky-header-height` GH variable. `sticky-comment-header` already does

source/features/sticky-sidebar.css
```diff
@@ -1,10 +1,3 @@
-[rgh-sticky-sidebar-enabled] #js-repo-pjax-container ul.pagehead-actions,
-/* https://github.com/refined-github/refined-github/issues/6761 */
-	[rgh-sticky-sidebar-enabled]
-	[id^='my-forks-menu'] {
-	z-index: initial !important; /* This affects the social buttons. It seems to have no effect but it constantly causes trouble with other features. */
-}
-
```
SunsetTechuila, line 7: Outdated

source/features/sticky-sidebar.css
```diff
@@ -27,26 +20,15 @@ Exclusively use simple sums and use `0px` instead of `0`
...
 }
 
-#discussion_bucket #partial-discussion-sidebar.rgh-sticky-sidebar-container {
-	height: min-content;
-}
-
-/* Hides the last divider (on pull requests) to avoid https://user-images.githubusercontent.com/10238474/62282128-af6fb980-b457-11e9-8b18-c29687b88da1.gif */
-.rgh-sticky-sidebar + .border-top {
-	display: none;
-}
```
SunsetTechuila, line 48: Outdated

source/features/sticky-sidebar.css
```diff
@@ -28,25 +21,11 @@ Exclusively use simple sums and use `0px` instead of `0`
 	--rgh-sticky-notification-header-height: 69px;
 }
 
-/* required to make `sticky` work for the issue Sidebar */
-.rgh-sticky-sidebar-container {
-	display: unset !important;
-}
```
SunsetTechuila, line 34: Outdated

source/features/sticky-sidebar.tsx
```diff
@@ -13,9 +13,8 @@ const minimumViewportWidthForSidebar = 768; // Less than this, the layout is sin
 
 const sidebarSelector = [
 	'.Layout-sidebar .BorderGrid', // `isRepoRoot`
-	'.Layout-sidebar #partial-discussion-sidebar', // Old `isConversation`
-	'div[data-testid="issue-viewer-metadata-pane"]', // `isConversation`
-	'#discussion_bucket #partial-discussion-sidebar', // `isDiscussion`
+	'div[data-testid="issue-viewer-metadata-pane"]', // Issue `isConversation`. TODO: Remove after March 2026
```
SunsetTechuila: Issue sidebar is natively sticky for at least 5 months: https://github.com/refined-github/refined-github/pull/8696

SunsetTechuila:
@fregante ready to be reviewed

fregante:
This is one of those I'm not sure about since it's toggling a native class. Why that change? Does the sidebar stick natively or not? I'm confused there

SunsetTechuila:
> it's toggling a native class

`is-stuck` doesn't do anything by itself

> Does the sidebar stick natively or not? 

It doesn't

fregante:
Your previous comment suggests that it's natively enabled on issues and the `includes` still targets issues. That should be updated too. 

Also again it's not clear why you switched from the rgh class to a non rgh class. Why? Can that change be reverted?

SunsetTechuila:
> it's natively enabled on issues and the `includes` still targets issues.

That's right

> That should be updated too.

What should be updated, and in what way?

> Also again it's not clear why you switched from the rgh class to a non rgh class. Why?

For me, adding the `rgh-sticky-sidebar` class to the sidebar and toggling `is-stuck` makes more sense than adding `rgh-sticky-sidebar-container` and toggling `rgh-sticky-sidebar`. Don't you agree?

fregante:
> What should be updated, and in what way?

The feature shouldn't run on issues. 

> Don't you agree?

No. If you use GitHub's classes it's because they do something. If the behavior is entirely controlled by us, then we use rgh- classes. 

It's true that technically we're replicating GitHub's behavior, but especially for that reason we should not reuse their classes in order to avoid future conflicts.

SunsetTechuila:
I expect sidebars to be rewritten soon. Let's see if they become natively sticky. If they do, I'll close this PR; if not, I'll update it

SunsetTechuila:
> I expect sidebars to be rewritten soon.

It seems I overestimated their development speed

PR #9019, opened 2026-02-27
`keyboard-navigation` - Restore feature on PR files

SunsetTechuila:
addresses https://github.com/refined-github/refined-github/issues/8504#issuecomment-3970308551

## Test URLs

https://github.com/refined-github/refined-github/pull/8517/changes

### DIFF
--- a/source/features/keyboard-navigation.tsx
+++ b/source/features/keyboard-navigation.tsx
@@ -22,7 +22,7 @@ function runShortcuts(event: KeyboardEvent): void {
 	const focusedComment = $optional(':target');
 	const items
 		= $$([
-			'.js-targetable-element[id^="diff-"]', // Files in diffs
+			'div[class*="targetable" i][id^="diff-"]', // Files in diffs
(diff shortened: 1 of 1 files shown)


source/features/keyboard-navigation.tsx
```diff
@@ -22,7 +22,7 @@ function runShortcuts(event: KeyboardEvent): void {
 	const focusedComment = $optional(':target');
 	const items
 		= $$([
-			'.js-targetable-element[id^="diff-"]', // Files in diffs
+			'div[class*="targetable" i][id^="diff-"]', // Files in diffs
```
SunsetTechuila, line 25: or `div:is([class*="diffTargetable"], .js-targetable-element)[id^="diff-"]` would be preferable?

SunsetTechuila: or `['div[class^="Diff-module__diffTargetable"]', '.js-targetable-element[id^="diff-"]']`?

fregante: I think this one is specific enough

